Issue #1703, opened 2019-01-09
Should the debug messages of the well model not be printed to the terminal by default?


andlaus:
when I run flow on e.g. Norne, the terminal gets cluttered by quite a few messages like this

```
    Switching control mode for well C-1H from SURFACE_RATE to BHP
    Switching control mode for well C-1H from BHP to SURFACE_RATE
```

Since they are not interesting for anyone who is not developing on the well model, they IMHO these should only go into the DBG and maybe the PRT file. I had a brief look at fixing this, but the `WellSwitchingLogger` class defeated me...

GitPaean:
> they are not interesting for anyone who is not developing on the well model

I am not sure about this. Some users/engineers might be interested in this. 

Anyway, it can be discussed.

andlaus:
> I am not sure about this. Some users/engineers might be interested in this.

sure, but then they can look at the DBG file, just like for FIP numbers. Also, the terminal output is currently pretty useless because it is impossible to say at which iteration it happened.

alfbr:
> Since they are not interesting for anyone who is not developing on the well model

On what basis do you claim that? Are you working with users who have complained? Not saying that you are wrong, just want context.

andlaus:
> On what basis do you claim that?

I claim this on the basis that personally I'm even less interested in these messages than in the FIP numbers and that these messages tend to be annoying if many wells switch their control mode often. That said, I don't think that this is a big issue, so if you insist that cluttering the terminal with these messages without any context, I'm fine with the current state of affairs. (I just think that the current policy what the user sees on the terminal is very inconsistent because more important information is *not* printed.)

GitPaean:
> very inconsistent because more important information is not printed

What do you mean by `more important information` here? Well control mode switching are actually quite important. I do not insist they should be put on screen.

andlaus:
> What do you mean by more important information here? 

for people like me, this is e.g., convergence information of the newton solver during a time step. (I don't complain, I don't have an issue with looking at the DBG file), for reservoir engineers this is e.g., the FIP data.

andlaus:
> Well control mode switching are actually quite important.

I agree that this is the case for people who are primarily concerned with developing the well model.

> I  do not insist they should be put on screen.

that's my point :)

GitPaean:
Maybe for the well control, for the screen (even PR file), we can just tell the converged well control mode at the end of the time step, not giving the back-and-forth switching process.

andlaus:
[...]

PR #1713, opened 2019-01-18
Replace well switching logger

fgfuchs:
Replace well switching logger with the more general deferred logger functionality


andlaus:
does this fix #1703? if yes. +1 from my side!

atgeirr:
> does this fix #1703? if yes. +1 from my side!

It is not intended to make any behaviour changes, as the WellSwitchingLogger is used correctly (if not in the way we may want it to). Next stop for this line of work is making all other logging in the well code correct in the parallel setting, changing behaviour as requested in #1703 will be a separate thing.

atgeirr:
jenkins build this please

atgeirr:

blattms:
benchmark

fgfuchs:
I think I have a couple more changes that are necessary...

atgeirr:
@blattms do you expect that communicating strings rather than just flags will have a measurable performance impact, or are you mostly using this PR to test the benchmarking system?

atgeirr:
Since there were no further comments, discussion or issues I will merge this now. Thanks!

PR #1712, opened 2019-01-18
adapt to the new preparation procedure of ewoms linear solvers

andlaus:
this is required to keep the linear solver of eWoms and opm-simulators identical, cf OPM/ewoms#461.

PR #1718, opened 2019-01-30
Properly adapt to the new linear solver API used by eWoms

andlaus:
I tested #1712 without deriving the `EclFlowProblem` type tag from `FlowIstlSolver`, and it worked because the linear solver was still `Opm::ISTLSolverEbos` as the linear solver is set via the `LinearSolverSplice` a few lines down.

This time, I verified that the `Ewoms::Linear::ParallelBiCGStabSolverBackend` was used if the offending line was commented out. also, Norne worked fine with the default solver as long as the Schur complement for the wells was done explicitly.

Finally, the naming of the eWoms API is a bit inconsistent (`setMatrix()` vs. `setResidual()`). any opinions here? I'm fine with whatever.

### DIFF
--- a/opm/autodiff/BlackoilModelEbos.hpp
+++ b/opm/autodiff/BlackoilModelEbos.hpp
@@ -69,7 +69,7 @@
 
 BEGIN_PROPERTIES
 
-NEW_TYPE_TAG(EclFlowProblem, INHERITS_FROM(BlackOilModel, EclBaseProblem, FlowNonLinearSolver, FlowIstlSolver, FlowModelParameters, FlowTimeSteppingParameters));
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this please

atgeirr:
> Norne worked fine with the default solver as long as the Schur complement for the wells was done explicitly.

What is the default you refer to here?

andlaus:
this one: https://github.com/OPM/ewoms/blob/master/ewoms/linear/parallelbicgstabbackend.hh

atgeirr:
Looks ok, merging. The naming is inconsistent as you say, so let's just agree that this may not be the final API yet...

PR #1721, opened 2019-02-05
Correct Logging for Black Oil Well Model for parallel runs

fgfuchs:
use deferred logging to log correctly in parallel for wells

### DIFF
--- a/opm/autodiff/BlackoilWellModel_impl.hpp
+++ b/opm/autodiff/BlackoilWellModel_impl.hpp
@@ -311,36 +311,35 @@ namespace Opm {
     template<typename TypeTag>
     void
     BlackoilWellModel<TypeTag>::wellTesting(const int timeStepIdx, const double simulationTime) {
-        const auto& wtest_config = schedule().wtestConfig(timeStepIdx);
-        if (wtest_config.size() == 0) { // there is no WTEST request
-            return;
-        }
-
-        // average B factors are required for the convergence checking of well equations
-        // Note: this must be done on all processes, even those with
-        // no wells needing testing, otherwise we will have locking.
-        std::vector< Scalar > B_avg(numComponents(), Scalar() );
-        computeAverageFormationFactor(B_avg);
         Opm::DeferredLogger local_deferredLogger;
+        const auto& wtest_config = schedule().wtestConfig(timeStepIdx);
+        if (wtest_config.size() != 0) { // there is a WTEST request
 
-        const auto& wellsForTesting = wellTestState_.updateWell(wtest_config, simulationTime);
-        for (const auto& testWell : wellsForTesting) {
-            const std::string& well_name = testWell.first;
+            // average B factors are required for the convergence checking of well equations
+            // Note: this must be done on all processes, even those with
+            // no wells needing testing, otherwise we will have locking.
+            std::vector< Scalar > B_avg(numComponents(), Scalar() );
+            computeAverageFormationFactor(B_avg);
 
-            // this is the well we will test
-            WellInterfacePtr well = createWellForWellTest(well_name, timeStepIdx);
+            const auto& wellsForTesting = wellTestState_.updateWell(wtest_config, simulationTime);
+            for (const auto& testWell : wellsForTesting) {
+                const std::string& well_name = testWell.first;
 
-            // some preparation before the well can be used
-            well->init(&phase_usage_, depth_, gravity_, number_of_cells_);
-            const WellNode& well_node = wellCollection().findWellNode(well_name);
-            const double well_efficiency_factor = well_node.getAccumulativeEfficiencyFactor();
-            well->setWellEfficiencyFactor(well_efficiency_factor);
-            well->setVFPProperties(vfp_properties_.get());
+                // this is the well we will test
+                WellInterfacePtr well = createWellForWellTest(well_name, timeStepIdx);
 
-            const WellTestConfig::Reason testing_reason = testWell.second;
+                // some preparation before the well can be used
+                well->init(&phase_usage_, depth_, gravity_, number_of_cells_);
+                const WellNode& well_node = wellCollection().findWellNode(well_name);
+                const double well_efficiency_factor = well_node.getAccumulativeEfficiencyFactor();
+                well->setWellEfficiencyFactor(well_efficiency_factor);
+                well->setVFPProperties(vfp_properties_.get());
 
-            well->wellTesting(ebosSimulator_, B_avg, simulationTime, timeStepIdx,
-                              testing_reason, well_state_, wellTestState_, local_deferredLogger);
+                const WellTestConfig::Reason testing_reason = testWell.second;
+
+                well->wellTesting(ebosSimulator_, B_avg, simulationTime, timeStepIdx,
+                                  testing_reason, well_state_, wellTestState_, local_deferredLogger);
+            }
         }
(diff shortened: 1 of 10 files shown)


opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -876,9 +876,13 @@ namespace Opm {
         } while (it < max_iter);
 
         if (converged) {
-            deferred_logger.debug("Well equation solution gets converged with " + std::to_string(it) + " iterations");
+            if (terminal_output_) {
```
blattms: I guess you need to change the initialization of `terminal_output_` to achieve logging on all processes. Currently it is initialized by
```
terminal_output_ = false;
if (ebosSimulator.gridView().comm().rank() == 0)
  terminal_output_ = EWOMS_GET_PARAM(TypeTag, bool, EnableTerminalOutput);
```

opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -311,36 +311,35 @@ namespace Opm {
     template<typename TypeTag>
     void
     BlackoilWellModel<TypeTag>::wellTesting(const int timeStepIdx, const double simulationTime) {
+        Opm::DeferredLogger local_deferredLogger;
         const auto& wtest_config = schedule().wtestConfig(timeStepIdx);
-        if (wtest_config.size() == 0) { // there is no WTEST request
```
blattms: Did we deadlock previously or what it is this reason for this change?

opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -386,9 +387,15 @@ namespace Opm {
         catch ( std::runtime_error& e )
         {
             const std::string msg = "A zero well potential is returned for output purposes. ";
-            OpmLog::warning("WELL_POTENTIAL_CALCULATION_FAILED", msg);
+            local_deferredLogger.warning("WELL_POTENTIAL_CALCULATION_FAILED", msg);
         }
         previous_well_state_ = well_state_;
+
+        Opm::DeferredLogger global_deferredLogger = gatherDeferredLogger(local_deferredLogger);
+        if (terminal_output_) {
```
blattms: If you change `terminal_output_` as suggested elsewhere this check will be true on all processes. And by design of `gatherDeferredLogger` (which uses MPI_Allgatherv) all processes will have all messages and log them. Hence multiples of all messages.

Maybe gatherDeferredLogger should be refactored to gather the messages only to one process?

opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -641,20 +649,26 @@ namespace Opm {
...
             }
             // TODO: should we update the explicit related here again, or even prepareTimeStep().
             // basically, this is a more updated state from the solveWellEq based on fixed
             // reservoir state, will tihs be a better place to inialize the explict information?
         }
-        assembleWellEq(dt);
+        assembleWellEq(dt, local_deferredLogger);
+
+        Opm::DeferredLogger global_deferredLogger = gatherDeferredLogger(local_deferredLogger);
+        if (terminal_output_) {
```
blattms: ditto

opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -905,11 +919,13 @@ namespace Opm {
         ConvergenceReport report = gatherConvergenceReport(local_report);
 
         // Log debug messages for NaN or too large residuals.
-        for (const auto& f : report.wellFailures()) {
-            if (f.severity() == ConvergenceReport::Severity::NotANumber) {
-                OpmLog::debug("NaN residual found with phase " + std::to_string(f.phase()) + " for well " + f.wellName());
-            } else if (f.severity() == ConvergenceReport::Severity::TooLarge) {
-                OpmLog::debug("Too large residual found with phase " + std::to_string(f.phase()) + " for well " + f.wellName());
+        if (terminal_output_) {
```
blattms: Mind the initialization of `terminal_output_`, see above.

blattms: Well, here the old initialization would be correct: Log only on one process.

fgfuchs:
> We also need to be aware that in case any method call between the initialization of `DeferredLogger` and the call to `gatherDeferredLogger` throws an uncaught exception, we will loose all the log messages. Not sure whether this might happen, though.

Good point, I do not know how to handle that.

fgfuchs:
> BTW: Title of PR should be changed, the commit squashed and commit message should contain a bit more detailed explanation.

Hope this is good enough

opm/autodiff/BlackoilWellModel_impl.hpp
fgfuchs: > I guess we have to have another variable apart from `terminal_output` to make this work as expected and/or refactor `gatherDeferredLogger`

@blattms thank you for the comment. The way parallel logging works, is that all processes get the messages of all processes. terminal_output can be true only on rank 0. Hence, the messages from all processes get logged only once. Please, feel free to correct me, if I am wrong.

opm/autodiff/BlackoilWellModel_impl.hpp
fgfuchs: If I understand correctly, than, before this change, logging happened only from rank 0. Now we use if(wtest_config.size() != 0) to test whether we need to execute well testing, and run gatherDeferredLogger afterwards in any case.

opm/autodiff/BlackoilWellModel_impl.hpp
fgfuchs: see other comments

opm/autodiff/BlackoilWellModel_impl.hpp
fgfuchs: ditto

opm/autodiff/BlackoilWellModel_impl.hpp
fgfuchs: see my other comments

blattms:
Well in a parallel run your PR clearly is an improvement (we did not see any logging despite of rank 0 before).

The question is how does the deferred logger work when there is only one rank? Does ist still only log messages in
the logMessages call? If so, maybe we can change this to use direct logging in case of only rank. Hence the situation
will be the same as before for a serial run. We do not loose anything but win something in a parallel run.


Great work BTW.

atgeirr:
> If so, maybe we can change this to use direct logging in case of only rank

That may be a good idea, and I am open to it. On the other hand it could make it harder to spot cases of user-unfriendly logging (for example, if logging happens too late to make sense of it) if behaviour is different between serial and parallel. Eating the parallel dogfood in serial if you will...

opm/autodiff/BlackoilWellModel_impl.hpp
blattms: I see. Makes sense.

opm/autodiff/BlackoilWellModel_impl.hpp
blattms: You are right. I oversaw that convergence is checked globally.

blattms:
With a good night of sleep and glasses on my nose. This now looks good. Sorry for bogus nitpicks before.

atgeirr:
jenkins build this please

atgeirr:
Thanks @fgfuchs for your work, and @blattms for your review. All is green so I'll merge!

PR #1723, opened 2019-02-05
flow: account for external setup costs

andlaus:
this actually adds the external setup time to the TCPU output in `flow`. It depends on OPM/ewoms#470.

### DIFF
--- a/flow/flow.cpp
+++ b/flow/flow.cpp
@@ -114,6 +114,9 @@ namespace detail
 // ----------------- Main program -----------------
 int main(int argc, char** argv)
 {
+    Dune::Timer externalSetupTimer;
+    externalSetupTimer.start();
+
     detail::handleVersionCmdLine(argc, argv);
     // MPI setup.
 #if HAVE_DUNE_FEM
@@ -221,13 +224,13 @@ int main(int argc, char** argv)
             // oil-gas
             if (phases.active( Opm::Phase::GAS ))
             {
-                Opm::flowEbosGasOilSetDeck(*deck, *eclipseState, *schedule, *summaryConfig);
+                Opm::flowEbosGasOilSetDeck(externalSetupTimer.elapsed(), *deck, *eclipseState, *schedule, *summaryConfig);
                 return Opm::flowEbosGasOilMain(argc, argv);
             }
             // oil-water
             else if ( phases.active( Opm::Phase::WATER ) )
             {
-                Opm::flowEbosOilWaterSetDeck(*deck, *eclipseState, *schedule, *summaryConfig);
+                Opm::flowEbosOilWaterSetDeck(externalSetupTimer.elapsed(), *deck, *eclipseState, *schedule, *summaryConfig);
                 return Opm::flowEbosOilWaterMain(argc, argv);
             }
             else {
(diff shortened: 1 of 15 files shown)


andlaus:
jenkins build this ewoms=470 please

andlaus:
did you forget to merge this, @atgeirr?

andlaus:
Note that the ewoms part has already been merged last Thursday.

bska:
This PR essentially adds an offset to the values of the existing TCPU curve.  Do you have an example of a before/after plot for a run on, say, Norne or Model 2, that will let us assess the impact?

andlaus:
your assessment is correct: the only difference is that after this PR the TCPU curve will be shifted upward a bit. for SPE-1 the impact is barely measurable, for norne it only is about 3.5 seconds on my machine, but some other decks which we're currently looking at (i.e., CO2CASE) it is in the order of a minute. since Eclipse seems to include these costs in TCPU (its the curve starts about 60 seconds away from 0), this PR mainly constitutes a small convenience: it is a bit annoying having to sit in front of the computer with a stopwatch to be able to compare the curves produced by Eclipse and flow ;)

alfbr:
I have not tested this PR, but generally the added time is useful output. Ideally I would like us to separate time spent in user code with wall-clock time, having the first in TCPU and the second in ELAPSED. However, there seems to be no willingness to go down this route now. With that said, this is good to go on our end.

bska:
> With that said, this is good to go on our end.

Thanks for the input.  This is certainly an aspect of the reporting that we can rework/improve over time.  I'll merge into master.  Thanks a lot.

andlaus:
> Ideally I would like us to separate time spent in user code with wall-clock time

out of curiosity: what do you mean with "user time" here? it is probably not the concept that the linux kernel uses for that term?!

alfbr:
Google knows ;-)

https://stackoverflow.com/questions/7335920/what-specifically-are-wall-clock-time-user-cpu-time-and-system-cpu-time-in-uni

andlaus:
ok, this is the linux kernel definition of that concept. preceeding your flow run with `time` will provide you with what you want:

```
time ./bin/flow ~/src/opm-data/norne/NORNE_ATW2013 --output-dir=.
[...]
================    End of simulation     ===============

Total time (seconds):         574.707
Solver time (seconds):        572.552
 Assembly time (seconds):     288.434 (Failed: 9.35329; 3.24278%)
 Linear solve time (seconds): 270.136 (Failed: 6.39468; 2.3672%)
 Update time (seconds):       6.71502 (Failed: 0.226964; 3.37994%)
 Output write time (seconds): 2.67415
Overall Well Iterations:      949 (Failed: 8; 0.842993%)
Overall Linearizations:       1972 (Failed: 63; 3.19473%)
Overall Newton Iterations:    1636 (Failed: 63; 3.85086%)
Overall Linear Iterations:    25270 (Failed: 561; 2.22002%)


real    9m37.648s
user    9m38.366s
sys     0m2.740s
```

(i.e., for Norne the difference is less than half a percent on my machine. note that user+sys!=real because of the async output writing. Also, I don't see why splitting these two is useful, but if you're willing to fund it: The customer is always right ;))

PR #1727, opened 2019-02-13
flow: add a test case using free boundary conditions

andlaus:
this is based on case 1 of SPE-1.


andlaus:
jenkins build this opm-tests=92 please

akva2:
needs to be added in tests/update_reference_data.sh.

tests/update_reference_data.sh
```diff
@@ -37,6 +37,7 @@ tests[spe12p]="flow spe1 SPE1CASE2_2P spe1_2p"
 tests[spe1oilgas]="flow spe1 SPE1CASE2_OILGAS spe1_oilgas"
 tests[spe1nowells]="flow spe1 SPE1CASE2_NOWELLS spe1_nowells"
 tests[spe1thermal]="flow spe1 SPE1CASE2_THERMAL spe1_thermal"
+tests[spe1thermal]="flow spe1 SPE1CASE1_FREEBC spe1_freebc"
```
akva2: unique key please.

andlaus:
> needs to be added in tests/update_reference_data.sh.

thanks. this is not obvious at all. should be fixed, can you confirm?

tests/update_reference_data.sh
andlaus: what do you mean? (I copied the line above and replaced "thermal" by "freebc".)

andlaus: ah, I see. forgot the LHS...

andlaus: should be fixed. it is possibly worth to think about modifying the build system to do this automatically. ( `opm_add_test()` should already get all required information.)

andlaus:
jenkins build this opm-tests=92 please

tests/update_reference_data.sh
akva2: i don't think so. this is not related to the build system, nor is it a task for the build system. this is related to jenkins automation only.

andlaus: yes, but the build system could generate the update script. Anyway, this is (mostly) not my cup of tea...

alfbr: Just want to add that not all results are automatically updated. The cell volume calculations need to be manually updated, and that is deliberate.

totto82:
I indent to add a test for mass flow bc to this test and rename it to spe1_bc. Can I close this and make a new one or should I merge this and make a new PR.

andlaus:
from my POV: great, let's close this one.

PR #1730, opened 2019-02-14
correct exception handling in wells for parallel runs

fgfuchs:
correct exception handling in wells classes for serial and parallel runs

### DIFF
--- a/opm/autodiff/BlackoilWellModel_impl.hpp
+++ b/opm/autodiff/BlackoilWellModel_impl.hpp
@@ -1,5 +1,26 @@
+/*
+ Copyright 2016 - 2019 SINTEF Digital, Mathematics & Cybernetics.
+ Copyright 2016 - 2018 Equinor ASA.
+ Copyright 2017 Dr. Blatt - HPC-Simulation-Software & Services
+ Copyright 2016 - 2018 Norce AS
 
+ This file is part of the Open Porous Media project (OPM).
 
+ OPM is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ OPM is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <opm/simulators/DeferredLoggingErrorHelpers.hpp>
 
 namespace Opm {
     template<typename TypeTag>
@@ -146,17 +167,22 @@ namespace Opm {
     forceShutWellByNameIfPredictionMode(const std::string& wellname,
                                         const double simulation_time)
     {
+        Opm::DeferredLogger local_deferredLogger;
         // Only add the well to the closed list on the
         // process that owns it.
         int well_was_shut = 0;
         for (const auto& well : well_container_) {
             if (well->name() == wellname) {
-                if (well->underPredictionMode()) {
+                if (well->underPredictionMode(local_deferredLogger)) {
                     wellTestState_.addClosedWell(wellname, WellTestConfig::Reason::PHYSICAL, simulation_time);
                     well_was_shut = 1;
                 }
                 break;
             }
+        }
+        Opm::DeferredLogger global_deferredLogger = gatherDeferredLogger(local_deferredLogger);
+        if (terminal_output_) {
+            global_deferredLogger.logMessages();
         }
 
         // Communicate across processes if a well was shut.
@@ -180,6 +206,8 @@ namespace Opm {
     BlackoilWellModel<TypeTag>::
     beginReportStep(const int timeStepIdx)
     {
+        Opm::DeferredLogger local_deferredLogger;
+
         const Grid& grid = ebosSimulator_.vanguard().grid();
         const auto& defunct_well_names = ebosSimulator_.vanguard().defunctWellNames();
         const auto& eclState = ebosSimulator_.vanguard().eclState();
@@ -249,14 +277,19 @@ namespace Opm {
         // Compute reservoir volumes for RESV controls.
         rateConverter_.reset(new RateConverterType (phase_usage_,
                                                     std::vector<int>(number_of_cells_, 0)));
-        computeRESV(timeStepIdx);
+
+        int exception_thrown = 0;
+        try {
+            computeRESV(timeStepIdx, local_deferredLogger);
+        } catch (const std::exception& e){
+            exception_thrown = 1;
+        }
+        logAndCheckForExceptionsAndThrow(local_deferredLogger, exception_thrown, "beginReportStep() failed.", terminal_output_);
 
         // update VFP properties
         vfp_properties_.reset (new VFPProperties<VFPInjProperties,VFPProdProperties> (
                                    schedule().getVFPInjTables(timeStepIdx),
                                    schedule().getVFPProdTables(timeStepIdx)) );
-
-
 
     }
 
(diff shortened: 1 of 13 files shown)


opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -244,13 +251,25 @@ namespace Opm {
         // Compute reservoir volumes for RESV controls.
         rateConverter_.reset(new RateConverterType (phase_usage_,
                                                     std::vector<int>(number_of_cells_, 0)));
-        computeRESV(timeStepIdx);
+
+        int exception_thrown = 0;
+        try {
+            computeRESV(timeStepIdx, local_deferredLogger);
+        } catch (...) {
```
blattms: I do not like this construct. Why the ellipsis? Can't we be more specific about what exceptions we are awaiting?

opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -261,16 +280,25 @@ namespace Opm {
...
+            wellTesting(reportStepIdx, simulationTime, local_deferredLogger);
 
-        // create the well container
-        well_container_ = createWellContainer(reportStepIdx);
+            // create the well container
+            well_container_ = createWellContainer(reportStepIdx, local_deferredLogger);
+        } catch (...) {
+            exception_thrown = 1;
+        }
+        OPM_CHECK_FOR_EXCEPTIONS_AND_LOG_AND_THROW(local_deferredLogger, exception_thrown, "BlackoilWellModel::beginTimeStep() failed.", terminal_output_);
```
blattms: Every call to this macro represents a synchronization point of flow. These should occur very scarcely as the might limit the scalability of it. To there very least there should be quite some amount of computation between them. I doubt that this is the case here.

opm/simulators/DeferredLoggingErrorMacros.hpp
```diff
@@ -0,0 +1,68 @@
...
+// exception class must exhibit a constructor with the signature
+// (const std::string &message). Since this condition is not fulfilled
+// for the std::exception, you should use this macro with some
+// exception class derived from either std::logic_error or
+// std::runtime_error.
+//
+// Usage: OPM_THROW(ExceptionClass, "Error message " << value);
+#define OPM_DEFLOG_THROW(Exception, message, deferred_logger)                             \
+    do {                                                                \
+        std::ostringstream oss__;                                       \
```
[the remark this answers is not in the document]
fgfuchs: probably, but I used the same approach as in ErrorMacros.hpp

opm/simulators/DeferredLoggingErrorMacros.hpp
```diff
@@ -0,0 +1,68 @@
...
+        oss__ << "[" << __FILE__ << ":" << __LINE__ << "] " << message; \
+        deferred_logger.error(oss__.str());                               \
+        throw Exception(oss__.str());                                   \
+    } while (false)
+
+#define OPM_CHECK_FOR_EXCEPTIONS_AND_THROW(exception_thrown, message)\
```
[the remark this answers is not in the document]
fgfuchs: It does not have to be. The advantage is, that one does not have to specify types and include the necessary header files.

opm/simulators/DeferredLoggingErrorMacros.hpp
```diff
@@ -0,0 +1,68 @@
...
+    do {\
+        auto cc = Dune::MPIHelper::getCollectiveCommunication();\
+        if (cc.max(exception_thrown) == 1) {\
+            throw std::logic_error(message);\
+        }\
+    } while (false)
+
+#define OPM_CHECK_FOR_EXCEPTIONS_AND_LOG_AND_THROW(deferred_logger, exception_thrown, message, terminal_output)\
```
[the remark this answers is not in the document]
fgfuchs: It does not have to be. The advantage is, that one does not have to specify types and include the necessary header files.

opm/autodiff/BlackoilWellModel_impl.hpp
fgfuchs: I agree, however, see my remark under.

fgfuchs:
> But from what I read the approach is against at least my principles of efficient scalabale parallel programs which should limit global synchronization points to the bare minimum absolutely needed.

I agree. However, the approach taken here, is to first get the correct behavior. Then, as a second step, optimize. The approach that "this almost never happens" is not a good one, considering that the aim is to improve stability and scaling of the code.

opm/simulators/DeferredLoggingErrorMacros.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+
+  Inspired by ErrorMacros.hpp in opm-common.
```
atgeirr: Move this comment to line 32 or so.

opm/simulators/DeferredLoggingErrorMacros.hpp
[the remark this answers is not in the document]
atgeirr: I agree, make it a function.

opm/autodiff/BlackoilWellModel_impl.hpp
fgfuchs: changed

opm/simulators/DeferredLoggingErrorMacros.hpp
fgfuchs: done

atgeirr:
jenkins build this please

atgeirr:
jenkins build this please

opm/simulators/DeferredLoggingErrorHelpers.hpp
```diff
@@ -0,0 +1,72 @@
...
+// Usage: OPM_DEFLOG_THROW(ExceptionClass, "Error message " << value, DeferredLogger);
+#define OPM_DEFLOG_THROW(Exception, message, deferred_logger)                             \
+    do {                                                                \
+        std::ostringstream oss__;                                       \
+        oss__ << "[" << __FILE__ << ":" << __LINE__ << "] " << message; \
+        deferred_logger.error(oss__.str());                               \
+        throw Exception(oss__.str());                                   \
+    } while (false)
+
+inline void check_for_exceptions_and_throw(const int exception_thrown, const std::string& message)
```
atgeirr: Use camelCase() for functions.

opm/autodiff/WellInterface_impl.hpp
```diff
@@ -1179,16 +1182,37 @@ namespace Opm
...
         do {
-            assembleWellEq(ebosSimulator, dt, well_state, deferred_logger);
+            int exception_thrown = 0;
+            try {
+                assembleWellEq(ebosSimulator, dt, well_state, deferred_logger);
+            } catch (...) {
+                exception_thrown = 1;
+            }
+
+            OPM_CHECK_FOR_EXCEPTIONS_AND_THROW(exception_thrown, "assembleWellEq in WellInterface::solveWellEqUntilConverged() failed.");
```
atgeirr: It is always wrong to communicate within the "single well" classes (WellInterface and its derived classes). Errors here must be caught in the BlackoilWellModel class.

opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -1,5 +1,5 @@
 
-
+#include <opm/simulators/DeferredLoggingErrorHelpers.hpp>
```
atgeirr: Seems this file is missing a copyright header, please add one...

opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -297,7 +325,7 @@ namespace Opm {
         {
             const Grid& grid = ebosSimulator_.vanguard().grid();
             if (PolymerModule::hasPlyshlog() || GET_PROP_VALUE(TypeTag, EnablePolymerMW) ) {
-                computeRepRadiusPerfLength(grid);
+                computeRepRadiusPerfLength(grid, local_deferredLogger);
```
atgeirr: Should be moved inside the try block, since it can throw.

opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -435,6 +460,7 @@ namespace Opm {
                 return well;
             }
         }
+        //this is not thread safe...
```
atgeirr: This comment, in isolation, is wrong. If the calling scope ensures catching and communication all is fine.

opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -643,53 +671,63 @@ namespace Opm {
...
-
     template<typename TypeTag>
     void
     BlackoilWellModel<TypeTag>::
     assembleWellEq(const double dt, Opm::DeferredLogger& deferred_logger)
     {
+        int exception_thrown = 0;
         for (auto& well : well_container_) {
-            well->assembleWellEq(ebosSimulator_, dt, well_state_, deferred_logger);
+            try {
```
atgeirr: Can the try be moved outside the for loop?

opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -757,11 +795,16 @@ namespace Opm {
...
     {
-        if (!localWellsActive())
-            return;
+        Opm::DeferredLogger local_deferredLogger;
 
-        for (auto& well : well_container_) {
-            well->recoverWellSolutionAndUpdateWellState(x, well_state_);
+        if (localWellsActive()) {
+            for (auto& well : well_container_) {
+                well->recoverWellSolutionAndUpdateWellState(x, well_state_, local_deferredLogger);
```
atgeirr: Can throw, need to catch and communicate.

opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -864,12 +907,20 @@ namespace Opm {
...
             ++it;
+            int exception_thrown = 0;
             if( localWellsActive() )
             {
                 for (auto& well : well_container_) {
-                    well->solveEqAndUpdateWellState(well_state_);
+                    try {
+                        well->solveEqAndUpdateWellState(well_state_, deferred_logger);
+                    } catch (...) {
+                        exception_thrown = 1;
```
atgeirr: No break? Also here: try{} the loop as a whole.

opm/simulators/DeferredLoggingErrorHelpers.hpp
fgfuchs: done

opm/autodiff/BlackoilWellModel_impl.hpp
fgfuchs: done

opm/autodiff/WellInterface_impl.hpp
fgfuchs: you are right.

opm/autodiff/BlackoilWellModel_impl.hpp
fgfuchs: done

opm/autodiff/BlackoilWellModel_impl.hpp
fgfuchs: done

opm/autodiff/BlackoilWellModel_impl.hpp
fgfuchs: done

opm/autodiff/BlackoilWellModel_impl.hpp
fgfuchs: done

fgfuchs:
Thank you @atgeirr @blattms for your comments. I have now updated the pull request according to your comments.

atgeirr:
> But from what I read the approach is against at least my principles of efficient scalabale parallel programs
> which should limit global synchronization points to the bare minimum absolutely needed.

I think the modified PR answers this pretty well, we still synchronize but less (higher in the call stack) than the original one. I am therefore ok with merging this now, unless you have further objections @blattms?

atgeirr:
jenkins build this please

atgeirr:
I do not understand the compile error in the failed test, @blattms or @fgfuchs do you?

akva2:
remove the const from the exception_thrown parameter. the max template cannot be instanced for const types.

fgfuchs:
@atgeirr please rebuild this on jenkins

akva2:
u can trigger now.

atgeirr:
jenkins build this please

akva2:
think u forgot to push?

atgeirr:
jenkins build this please

fgfuchs:
> think u forgot to push?

yes, I needed to force push, since I amended my commit.

fgfuchs:
jenkins build this please

fgfuchs:
jenkins build this please

atgeirr:
Thanks for your effort! Merging.

PR #1731, opened 2019-02-15
flow: tell EclProblem the size of our next time step before writing output

andlaus:
if we don't do this, the internal logic of EclProblem is used instead. this is an attempt to fix OPM/ewoms#479. As mentioned by @totto82 there, if the correct time step size is written by the current master it rather a coincidence than expected behavior...

### DIFF
--- a/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
+++ b/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
@@ -322,6 +322,7 @@ namespace Opm {
                         Opm::time::StopWatch perfTimer;
                         perfTimer.start();
 
+                        ebosProblem.setForcedNextStepSize(dtEstimate);
                         ebosProblem.writeOutput(/*isSubStep=*/true);
 
                         report.output_write_time += perfTimer.secsSinceStart();


totto82:
jenkins build this please

totto82:
jenkins build this ewoms=479 please

opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
```diff
@@ -322,6 +322,7 @@ namespace Opm {
                         Opm::time::StopWatch perfTimer;
                         perfTimer.start();
 
+                        ebosProblem.setForcedNextStepSize(dtEstimate);
                         ebosProblem.writeOutput(/*isSubStep=*/true);
```
totto82: This line also needs to be added after
https://github.com/OPM/opm-simulators/blob/657c45be92f23215f9bc78000a0dcad9dfa2c66d/opm/autodiff/SimulatorFullyImplicitBlackoilEbos.hpp#L291

Then Jenkins is happy

andlaus: Oops, thanks for finding this. actually, the machinery for forcing the next time step size was already present but simply not used by the new timestep proposal machinery in ebos. fixed in the ewoms PR. (BTW: this only has only ever worked with flow's adaptive time stepper.)

totto82: Great. That means one hack less.

andlaus:
this is obsolete.

PR #1735, opened 2019-02-25
well model: fix application of Schur complements

andlaus:
using the eWoms API for wells, the Schur compliment was not applied at all. If `BlackOilWellModel::linearize()` was made non-trivial, the Schur complement was applied twice in `flow`. With this patch, we only apply this using the eWoms API (in `BlackOilWellModel::linearize()`). I could not observe an effect on the convergence behaviour of `flow` for the cases which I tested (Norne and realization 5 of Model 2).

### DIFF
--- a/opm/autodiff/BlackoilModelEbos.hpp
+++ b/opm/autodiff/BlackoilModelEbos.hpp
@@ -246,7 +246,7 @@ namespace Opm {
             report.total_linearizations = 1;
 
             try {
-                report += assemble(timer, iteration);
+                report += assembleReservoir(timer, iteration);
                 report.assemble_time += perfTimer.stop();
             }
             catch (...) {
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this please

totto82:
> using the eWoms API for wells, the Schur compliment was not applied at all.

It was applied in solve Jacobian system. 

With this patch the schur complement is again added before the getConvergence method which I don't think is currect.

andlaus:
at least it did not change performance for the two tested non-trivial cases. in principle, I see the point, though: probably `getConvergence()` needs to be called between computation of the "raw" reservoir residual (i.e., including the well source terms) and the inclusion of the Schur complement instead of after. The question is: if it does not matter, why bother?

GitPaean:
> The question is: if it does not matter, why bother?

It it does not matter, we should stay with the correct formulation unless the understanding changes. 

Otherwise, it might cause problem later and difficult to debug.

andlaus:
ok, I added a patch which hopefully does thinks correctly.

andlaus:
jenkins build this please

totto82:
The patch seems fine. Could you check that the residual for instance on norne stays the same. A diff between the .DEBUG file for master and this patch should only have timing differences due to the usual noise.

andlaus:
> Could you check that the residual for instance on norne stays the same. A diff between the .DEBUG file for master and this patch should only have timing differences due to the usual noise.

I just verified that this is the case. Also, the total number of non-linear and linear iterations are exactly the same.

andlaus:
Anything which prevents merging this

totto82:
No, thanks.

PR #1736, opened 2019-02-25
aquifers: add missing serialize() and deserialize() methods

andlaus:
the original purpose of these methods is to provide a checkpoint/restart mechanism using an ad-hoc file format. They might also be useful/required for implementing the adjoint functionality, though.

### DIFF
--- a/opm/autodiff/BlackoilAquiferModel.hpp
+++ b/opm/autodiff/BlackoilAquiferModel.hpp
@@ -60,6 +60,12 @@ namespace Opm {
             void endTimeStep();
             void endEpisode();
 
+            template <class Restarter>
+            void serialize(Restarter& res);
+
+            template <class Restarter>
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this please

GitPaean:
Maybe we should throw when these two functions are called to make it clear that trying to do that with Aquifer model is not supported and not possible at the moment.

andlaus:
> Maybe we should throw when these two functions are called to make it clear that trying to do that with Aquifer model is not supported and not possible at the moment.

fixed, thanks for bringing it up. mind to press Green?

andlaus:
jenkins build this please

GitPaean:
I am not sure why we need these functions now before we figure out whether we need it or how to implement it. But I will merge them as requested.

PR #1734, opened 2019-02-21
move ebos over from eWoms

andlaus:
ebos is continuing to be the more experimental simulator for research, while flow remains to primarily targeted at production.

this PR is the twin of #OPM/ewoms#484.

@akva2: can you do your history transfer magic? Be aware that ebos was located in the directory `applications/ebos` until November 2016.

PR #1737, opened 2019-02-27
Import ebos from eWoms

akva2:
This imports ebos from eWoms with history. @andlaus easiest is probably merging this, then rebasing your branch afterwards.


andlaus:
I agree. This PR probably does not deal with the build system?

akva2:
Pure import. You will have a couple of trivial conflicts in the files you changed when moving things over.

-------- Opprinnelig melding --------
Fra: Andreas Lauser <notifications@github.com>
Dato: 27.02.2019 22:23 (GMT+01:00)
Til: OPM/opm-simulators <opm-simulators@noreply.github.com>
Ko: Arne Morten Kvarving <Arne.Morten.Kvarving@sintef.no>, Author <author@noreply.github.com>
Emne: Re: [OPM/opm-simulators] Import ebos from eWoms (#1737)


I agree. This PR probably does not deal with the build system?

andlaus:
okay, I've made the build system changes of #1734 into a PR for this one: https://github.com/akva2/opm-simulators/pull/5 . as far as I can see, once you've merged this, there is nothing which prevents hitting green on the ebos moval PRs.

akva2:
jenkins build this please

andlaus:
thanks for merging!

bska:
Are the ebos headers supposed to be a part of opm-simulators public interface when installed as a package?  If so, they must be added to [`CMakeLists_files.cmake`](https://github.com/OPM/opm-simulators/blob/master/CMakeLists_files.cmake#L113-L115). At present the headers are not copied to the destination directory during `make install` as evidenced by `ls -1` on my laptop:
```sh
$ ls -1 inst/include/opm/
autodiff
common
core
elasticity
flowdiagnostics
grid
json
material
output
parser
porsol
simulators
upscaling
utility
$
```
To further drive the point home, here's a more directed search
```sh
$ find inst/include/opm/ -name '*.hh'
inst/include/opm/elasticity/materials.hh
inst/include/opm/elasticity/mpc.hh
inst/include/opm/elasticity/boundarygrid.hh
inst/include/opm/elasticity/material.hh
inst/include/opm/grid/cpgrid/dgfparser.hh
inst/include/opm/grid/polyhedralgrid/iterator.hh
inst/include/opm/grid/polyhedralgrid/persistentcontainer.hh
inst/include/opm/grid/polyhedralgrid/entityseed.hh
inst/include/opm/grid/polyhedralgrid/geometry.hh
inst/include/opm/grid/polyhedralgrid/grid.hh
inst/include/opm/grid/polyhedralgrid/dgfparser.hh
inst/include/opm/grid/polyhedralgrid/entitypointer.hh
inst/include/opm/grid/polyhedralgrid/cartesianindexmapper.hh
inst/include/opm/grid/polyhedralgrid/gridhelpers.hh
inst/include/opm/grid/polyhedralgrid/indexset.hh
inst/include/opm/grid/polyhedralgrid/entity.hh
inst/include/opm/grid/polyhedralgrid/intersectioniterator.hh
inst/include/opm/grid/polyhedralgrid/intersection.hh
inst/include/opm/grid/polyhedralgrid/declaration.hh
inst/include/opm/grid/polyhedralgrid/capabilities.hh
inst/include/opm/grid/polyhedralgrid/idset.hh
inst/include/opm/grid/polyhedralgrid/gridview.hh
inst/include/opm/grid/common/p2pcommunicator.hh
inst/include/opm/grid/common/p2pcommunicator_impl.hh
inst/include/opm/grid/polyhedralgrid.hh
$
```

andlaus:
> Are the ebos headers supposed to be a part of opm-simulators public interface when installed as a package?  

I don't know and actually I do not really have an opinion here: Currently in my mind, the purpose of opm-simulators isn't to provide infrastructure for the consumption of downstream modules.

That said, maybe this is a good motivation to switch creating these file lists via glob matching: I'm not aware of any problems due to missing cmake runs in eWoms and opm-material (where they are glob-produced) over the last years, but there is a constant background noise of problems due to forgotten headers in the lists in the other OPM modules.

atgeirr:
I think globbing is acceptable for header files (and I would be willing to switch for them), but for files that should trigger a compile (cpp files) they must be explicitly listed if we do not want the old hassle of forced reconfiguration each time a new compilation unit is added. Or has globbing changed in cmake, so a simple 'make' will re-scan directories?

akva2:
no, globbing is just as error prone as it ever was. i am still opposed to globbing, it makes the build system behave in non-deterministic ways.

andlaus:
> they must be explicitly listed if we do not want the old hassle of forced reconfiguration each time a new compilation unit is added.

I'm fine with that, but I'd like to amend that adding a trivial change to `CMakeLists.txt` serves the same purpose. As far as I can see, the worst which could happen here that we get some github comments about missed touches of `CMakeLists.txt` (which we currently get anyway due to incorrect file lists).

> no,  globbing is just as error prone as it ever was. i am still opposed to  globbing, it makes the build system behave in non-deterministic ways.

as said: I'm pretty indifferent here, because this issue doesn't bother me much one way or the other. I just would like to point out the fact that the operational history over the last ~5 years did not indicate any problems with globbing, while about every second PR which added or removed a file did not adapt the respective list, at least in the initial version.

blattms:
Actually, I totally concur with @akva2 on the matter of globbing. But there might be  a compromise supported by CMake
on all platforms since at least 2.8.9. `INSTALL` supports the options `FILES_MATCHING PATTERN <pattern>` and a REGEX
version of it if one installs a directory. Therefore
```
install(DIRECTORY opm/ DESTININATION ${CMAKE_INSTALL_INCLUDEDIR}/opm FILES_MATCHING PATTERN "*.hpp")
```
might now be an alternative.

PR #1739, opened 2019-03-01
Miscellaneous fixes and improvements

andlaus:
this PR contains some random fixes and improvements that do I think not merit individual PRs and which have accumulated on my local copy of opm-simulators during the last weeks.

### DIFF
--- a/ebos/eclnewtonmethod.hh
+++ b/ebos/eclnewtonmethod.hh
@@ -159,15 +159,24 @@ public:
             }
 
             const auto& r = currentResidual[dofIdx];
-            const double pvValue =
+            Scalar pvValue =
                 this->simulator_.problem().porosity(dofIdx)
                 * this->model().dofTotalVolume(dofIdx);
             sumPv += pvValue;
             bool cnvViolated = false;
 
+            Scalar dofVolume = this->model().dofTotalVolume(dofIdx);
+
             for (unsigned eqIdx = 0; eqIdx < r.size(); ++eqIdx) {
                 Scalar tmpError = r[eqIdx] * dt * this->model().eqWeight(dofIdx, eqIdx) / pvValue;
                 Scalar tmpError2 = r[eqIdx] * this->model().eqWeight(dofIdx, eqIdx);
+
+                // in the case of a volumetric formulation, the residual in the above is
+                // per cubic meter
+                if (GET_PROP_VALUE(TypeTag, UseVolumetricResidual)) {
+                    tmpError *= dofVolume;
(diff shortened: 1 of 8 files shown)


andlaus:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -707,6 +707,18 @@ public:
...
+                boost::posix_time::from_time_t(timeMap.getStartTime(nextEpisodeIdx+1));
+            std::cout << "Report step " << nextEpisodeIdx + 2
+                      << "/" << timeMap.numTimesteps()
+                      << " at day " << timeMap.getTimePassedUntil(nextEpisodeIdx+1)/(24*3600)
+                      << "/" << timeMap.getTotalTime()/(24*3600)
+                      << ", date = " << curDateTime.date()
+                      << "\n ";
+        }
+
         if (nextEpisodeIdx > 0 &&
```
totto82: Does this add output to flow?

ebos/eclnewtonmethod.hh
```diff
@@ -159,16 +159,25 @@ public:
             }
 
             const auto& r = currentResidual[dofIdx];
-            const double pvValue =
+            Scalar pvValue =
```
GitPaean: why not const here?

ebos/eclproblem.hh
andlaus: oops, yes. I "fixed" this by making it conditional on the `EnableExperiments` property. Thanks for noticing.

ebos/eclnewtonmethod.hh
andlaus: because the variable is potentially useful for modification in the function and honestly `const` is IMO slightly ugly while not being useful for by-value objects.

andlaus:
jenkins build this please

ebos/eclnewtonmethod.hh
GitPaean: > because the variable is potentially useful for modification in the function and honestly

when you need to change it, you remove `const`, instead of the other way around.

andlaus: ... or just never use `const` for by-value objects. this is the approach 90% of C++ code seems to use ;)

GitPaean: > this is the approach 90% of C++ code seems to use ;)

I am really not sure whether this statistics exist, while you do not have a good reason to remove the existing `const`s here.

totto82:
Thanks for the fixes. The const discussion belongs in the pub. (i.e. mostly matter of taste IMO.)

PR #1741, opened 2019-03-01
EclProblem: Rename NewtonRawTolerance to NewtonTolerance

andlaus:
this is the downstream of OPM/ewoms#486: this property is actually not used by `flow` because `flow` implements its own Newton method, but it not renaming the property prevents `flow` from building.


andlaus:
jenkins build this ewoms=486 please

PR #1742, opened 2019-03-01
flow: use a specialized linearizer

andlaus:
This patch enables the linearization infrastructure of the research codes in eWoms to diverge from the production ones while guaranteeing stability for the latter. The actual new linearizer class for flow currently corresponds to the one of eWoms master, but with slightly more extensive usage of openMP.

Another purpose of the patch is to provide a blueprint of how diverging requirements of research- and production codes can be handled in the future: In this approach, the affected classes are simply forked off and replace their original by overwriting the respective eWoms property; in the particular case of `flow`, this approach allows to add ECL-specific code in otherwise general classes that cannot be easily changed without severely disturbing the eWoms-based research simulators.

### DIFF
--- a/flow/flowlinearizer.hh
+++ b/flow/flowlinearizer.hh
@@ -1,0 +1,608 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this please

hnil:
This seems like a good idea. But could it not only inherit fvbaselinearizer?

andlaus:
if you want to have the option to modify one without affecting the other (this is the point here, isn't it?), inheriting is not a terribly good idea...

hnil:
I thought the idea was that one could set the linearizer as a tag, but that was already present? Then I do not see a point before we have a need. I think the most likely use will be to reuse some and modify other parts, where inheriting seems natural. Now it is only a copy? A usecase will probably be splitting methods, different linarizations of the combination of the same system will be needed (but also the original linarization)

andlaus:
> I thought the idea was that one could set the linearizer as a tag, but that was already present?

what do you mean with that? there has been a `Linearizer` property tag from day 1 of ewoms. (so far, it was just never changed from the default value.)

> Now it is only a copy?

as of now, almost: it uses openMP more extensively. also, in the past @atgeirr hinted several times that he wants to fork basic eWoms infrastructure in order to amend it with ECL specific functionality. This PR shows how to do this without disturbing the research oriented simulators and models shipped by eWoms.

bska:
I'm closing this as stale now.

PR #1744, opened 2019-03-04
warning instead of throwing for nnc between active cells and inactive cells

GitPaean:
to make the current targeting case to run through. 

Something wired is that, now it runs, while I could not find the warning message anywhere.

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -676,9 +676,14 @@ private:
                 // Silently discard as it is not between active cells
                 continue;
 
-            if (low == -1 || high == -1)
-                OPM_THROW(std::logic_error, "NNC between active and inactive cells (" <<
-                          low << " -> " << high);
+            if (low == -1 || high == -1) {
+                // Discard the NNC if it is between active cell and inactive cell
+                std::ostringstream sstr;
+                sstr << "NNC between active and inactive cells ("
+                     << low << " -> " << high << ")";
+                Opm::OpmLog::warning(sstr.str());
+                continue;
+            }
 
             auto candidate = trans_.find(isId_(low, high));
 


GitPaean:
jenkins build this please

totto82:
Didn't @atgeirr fix this?

GitPaean:
> Didn't @atgeirr fix this?

is there an open PR about his?

GitPaean:
The reason that there is not warning message is that 

We call `setupLogging();` after `setupEbosSimulator();` in FlowMainEbos. 

But if we switch the order of them, we get the following segmentation fault. 
```
Reading deck file 'FULL.DATA'

Program received signal SIGSEGV, Segmentation fault.
0x0000000000ab5729 in std::unique_ptr<Ewoms::EclCpGridVanguard<Ewoms::Properties::TTag::EclFlowProblem>, std::default_delete<Ewoms::EclCpGridVanguard<Ewoms::Properties::TTag::EclFlowProblem> > >::get (this=<optimized out>) at /usr/include/c++/5/bits/unique_ptr.h:308
308	      { return std::get<0>(_M_t); }
(gdb) backtrace 
#0  0x0000000000ab5729 in std::unique_ptr<Ewoms::EclCpGridVanguard<Ewoms::Properties::TTag::EclFlowProblem>, std::default_delete<Ewoms::EclCpGridVanguard<Ewoms::Properties::TTag::EclFlowProblem> > >::get (this=<optimized out>) at /usr/include/c++/5/bits/unique_ptr.h:308
#1  std::unique_ptr<Ewoms::EclCpGridVanguard<Ewoms::Properties::TTag::EclFlowProblem>, std::default_delete<Ewoms::EclCpGridVanguard<Ewoms::Properties::TTag::EclFlowProblem> > >::operator* (this=<optimized out>) at /usr/include/c++/5/bits/unique_ptr.h:294
#2  Ewoms::Simulator<Ewoms::Properties::TTag::EclFlowProblem>::vanguard (this=<optimized out>) at /home/kaib/OPM-master-test/debug/ewoms/ewoms/common/simulator.hh:174
#3  Opm::FlowMainEbos<Ewoms::Properties::TTag::EclFlowProblem>::eclState (this=<optimized out>) at /home/kaib/OPM-master-test/debug/opm-simulators/opm/autodiff/FlowMainEbos.hpp:496
#4  Opm::FlowMainEbos<Ewoms::Properties::TTag::EclFlowProblem>::setupLogging (this=this@entry=0x7fffffffb240) at /home/kaib/OPM-master-test/debug/opm-simulators/opm/autodiff/FlowMainEbos.hpp:334
#5  0x0000000000c0e3a9 in Opm::FlowMainEbos<Ewoms::Properties::TTag::EclFlowProblem>::execute (this=this@entry=0x7fffffffb240, argc=<optimized out>, argv=<optimized out>) at /home/kaib/OPM-master-test/debug/opm-simulators/opm/autodiff/FlowMainEbos.hpp:223
#6  0x0000000000a12c62 in Opm::flowEbosBlackoilMain (argc=6, argv=0x7fffffffdd78) at /home/kaib/OPM-master-test/debug/opm-simulators/flow/flow_ebos_blackoil.cpp:62
#7  0x00000000009212e6 in main (argc=6, argv=0x7fffffffdd78) at /home/kaib/OPM-master-test/debug/opm-simulators/flow/flow.cpp:280
```

totto82:
> is there an open PR about his?

I thought that OPM/opm-grid#355 made sure that no NNC was created between active and non-active cells. But maybe I am wrong.

GitPaean:
> I thought that OPM/opm-grid#355 made sure that no NNC was created between active and non-active cells. But maybe I am wrong.

@atgeirr just pointed out that with the following code in OPM/opm-grid#355 , the situation that the current PR is addressing should not happen, which means there is a bug somewhere. 
```
                const int c1 = global_to_local[nncpair.first];
                global_to_local.resize(cart_size, -1);	                const int c2 = global_to_local[nncpair.second];
                for (size_t idx = 0; idx < global_cell.size(); ++idx) {	                if (c1 < 0 || c2 < 0) {
                    global_to_local[global_cell[idx]] = idx;	                    Opm::OpmLog::warning("nnc_inactive", "NNC connection requested between inactive cells.");
                    continue;
                }
```

blattms:
@atgeirr might wrong on this one. We are looping over the nncs here as provided by the deck, CpGrid will not change these. Therefore there might be nncs specified between active and inactive cells, if that is valid for eclipse.

Seems like this code is from. I guess I should given myself the same advice when writing as later to atgeirr.

atgeirr:
> @atgeirr might be wrong on this one.

You are correct, the code in opm-grid only deals with inserting a virtual (nnc) face in the grid or not, and the transmissibility code here loops over all nncs, regardless of any connection in the grid.

GitPaean:
> You are correct, the code in opm-grid only deals with inserting a virtual (nnc) face in the grid or not, and the transmissibility code here loops over all nncs, regardless of any connection in the grid.

Maybe we should update the nncs in the parser when handling the nnc related things in opm-grid, assuming the grid will not be changed again after opm-grid.

Similar to https://github.com/OPM/ewoms/pull/283 . 

Anyway, this PR provides a temporary solution for now.

blattms:
The parser should be / is  readonly

atgeirr:
Looks fine, will merge as is.

ebos/ecltransmissibility.hh
```diff
@@ -676,9 +676,14 @@ private:
...
                 continue;
 
-            if (low == -1 || high == -1)
-                OPM_THROW(std::logic_error, "NNC between active and inactive cells (" <<
-                          low << " -> " << high);
+            if (low == -1 || high == -1) {
+                // Discard the NNC if it is between active cell and inactive cell
+                std::ostringstream sstr;
+                sstr << "NNC between active and inactive cells ("
+                     << low << " -> " << high << ")";
```
andlaus: the message here is probably not really useful for people who do the deck-wrestling: at least one of `high` and `low` is -1...

andlaus: also: are these variables holding Cartesian indices? As far as I know, people who create decks have no good way to translate compressed indices to the ones which they actually need to use (i.e., Cartesian ones).

GitPaean: Agreed, probably we should return I, J, K in order to help to fix the deck.  

At the moment, we will not get any message yet, since we We call `setupLogging();` after `setupEbosSimulator();` in `FlowMainEbos`.

PR #1746, opened 2019-03-05
ebos: implement partial elimination of systematic mass defects

andlaus:
the idea is to compensate the residual of the final solution of a time step by means of an opposing source term in the next time step. While performance seems to be a few percent worse than the master version when using the same settings, the results often seem to be much better, so we can often get away with more relaxed tolerances.

This patch has been developed as a joint project with @totto82 and @osae.

(Note that `flow` is unaffected by this because drift compensation is an experimental feature and thus disabled within the production simulator.)

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -127,6 +127,10 @@ NEW_PROP_TAG(EnableWriteAllSolutions);
 // The number of time steps skipped between writing two consequtive restart files
 NEW_PROP_TAG(RestartWritingInterval);
 
+// Enable partial compensation of systematic mass losses via the source term of the next time
+// step
+NEW_PROP_TAG(EclEnableDriftCompensation);
+
 // Enable the additional checks even if compiled in debug mode (i.e., with the NDEBUG
 // macro undefined). Next to a slightly better performance, this also eliminates some
 // print statements in debug mode.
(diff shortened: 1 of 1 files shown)


ebos/eclproblem.hh
```diff
@@ -1443,6 +1480,38 @@ public:
...
+            // compute the weighted total drift rate
+            Scalar totalDriftRate = 0.0;
+            for (unsigned eqIdx = 0; eqIdx < numEq; ++ eqIdx)
+                totalDriftRate +=
+                    std::abs(dofDriftRate[eqIdx])*dt*model.eqWeight(globalDofIdx, eqIdx)/poro;
+
+            // make sure that we do not exceed the maximum rate of drift compensation
+            if (totalDriftRate > maxCompensation)
+                dofDriftRate *= maxCompensation/totalDriftRate;
+
```
totto82: The implementation here is slightly different from the earlier prototype.  How well did you test this? How sensitive is the results on the particular choice of maxCompensation?

andlaus: I ran it on Norne and M2R5, and it seemed to work well. more extensive testing is definitely in order, though...

andlaus:
jenkins build this please

ebos/eclproblem.hh
totto82: Ok. I will do some more testing. Since it is within the experimental mode I will merge it as is.

PR #1747, opened 2019-03-05
add some ebos variants

andlaus:
these variants should cover most of the common use cases. For developers, the main advantage of using one of these instead of `flow` is that they have faster round-trip times because they build faster: 53 seconds for `ebos` instead 1 minute 25 for `flow` on my machine with 32 parallel build threads (for machines that have fewer than 9 physical cores, the difference is likely even larger).

That said, there are no plans to provide simulators for combinations of blackoil extensions or a "all-in-one" simulator like `flow`: If someone is interested in e.g., an oil-water simulator with polymer and energy enabled, a separate self-compiled executable should be added locally.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,8 @@ cmake_minimum_required (VERSION 2.8)
 
 option(SIBLING_SEARCH "Search for other modules in sibling directories?" ON)
 set( USE_OPENMP_DEFAULT OFF ) # Use of OpenMP is considered experimental
+option(BUILD_FLOW "Build the production oriented flow simulator?" ON)
+option(BUILD_EBOS "Build the research oriented ebos simulator and its variants?" ON)
 
 if(SIBLING_SEARCH AND NOT opm-common_DIR)
   # guess the sibling dir
@@ -127,7 +129,9 @@ opm_add_test(test_gatherdeferredlogger
     5 ${CMAKE_BINARY_DIR}
 )
 
+include(OpmBashCompletion)
 
+if (BUILD_FLOW)
 # the production oriented general-purpose ECL simulator
 opm_add_test(flow
   ONLY_COMPILE
@@ -145,12 +149,15 @@ opm_add_test(flow
     flow/flow_ebos_oilwater_polymer.cpp
     flow/flow_ebos_oilwater_polymer_injectivity.cpp)
 install(TARGETS flow DESTINATION bin)
+opm_add_bash_completion(flow)
 
 add_test(NAME flow__version
          COMMAND flow --version)
 set_tests_properties(flow__version PROPERTIES
                      PASS_REGULAR_EXPRESSION "${${project}_LABEL}")
+endif()
 
+if (BUILD_EBOS)
 # the research oriented general-purpose ECL simulator ("ebos" == &ecl
 # &black-&oil &simulator)
 opm_add_test(ebos
@@ -208,6 +215,10 @@ install(TARGETS ebos_gasoil DESTINATION bin)
 install(TARGETS ebos_oilwater DESTINATION bin)
 install(TARGETS ebos_thermal DESTINATION bin)
 
-include(OpmBashCompletion)
-opm_add_bash_completion(flow)
 opm_add_bash_completion(ebos)
+opm_add_bash_completion(ebos_solvent)
+opm_add_bash_completion(ebos_polymer)
+opm_add_bash_completion(ebos_gasoil)
+opm_add_bash_completion(ebos_oilwater)
+opm_add_bash_completion(ebos_thermal)
+endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,8 @@ cmake_minimum_required (VERSION 2.8)
 option(SIBLING_SEARCH "Search for other modules in sibling directories?" ON)
 set( USE_OPENMP_DEFAULT OFF ) # Use of OpenMP is considered experimental
 option(BUILD_FLOW "Build the production oriented flow simulator?" ON)
-option(BUILD_EBOS "Build the research oriented ebos simulator and its variants?" ON)
+option(BUILD_EBOS "Build the research oriented ebos simulator?" ON)
+option(BUILD_EBOS_EXTENSIONS "Build the variants for various extensions of ebos by default?" OFF)
 
 if(SIBLING_SEARCH AND NOT opm-common_DIR)
   # guess the sibling dir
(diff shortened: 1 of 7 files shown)


joakim-hove:
Are these varieties compiled unconditionally? If possible I would prefer if they were behind a 
```
if (BUILD_EBOS) 
   ...
endif()
``` 
in `cmake` - as you point out yourself build times are long.

andlaus:
> Are these varieties compiled unconditionally?

so far, yes. note that you can already build selectively using

```
make -j32 ebos ebos_solvent flow
```

andlaus:
jenkins build this please

GitPaean:
The most useful one will be flow_blackoil, which only handles threephase blackoil simulation. it will be very helpful for developers. 

For executable are not supposed to enter release binaries, I think they should be built conditionally with cmake arguments. They are only intended for developers not users.

andlaus:
> The most useful one will be `flow_blackoil`

do you mean `ebos_blackoil`? this is already there and just called `ebos`. (`ebos` is an acronym for "**E**CL **b**lack-**o**il **s**imulator".)

> For executable are not supposed to enter release binaries, I think they should be built conditionally with cmake arguments. They are only intended for developers not users.

why simply not packaging these binaries in the same package as `flow`? most users won't self-compile their binaries anyway and those who do, will not do it often (i.e., build times are rather irrelevant for users). Developers on the other hand can (and should IMO) build selectively already, it is convenient not having to supply any special flags to `cmake` to test out the experimental feature you're currently working on, and also, even if you build everything, on my machine `make -j32 all` takes only 5 seconds longer with this PR than without it...

GitPaean:
> do you mean ebos_blackoil? this is already there and just called ebos. (ebos is an acronym for "ECL black-oil simulator".)

nope. flow currently incorporates several simulators together, it requires quite some of memory and times to compile it, however, most of the times, developers is focusing on the threephase blackoil functionalities the most. With flow_blackoil, probably it is faster and needs less memory to re-compile to develop and test. 

Making these variants to be built conditionally is just trying to reduce the time needs to re-compile when do `make` or `make -j 8` or something similar which is the default option to build that includes all the tests that needs to be tested.

andlaus:
> nope. flow currently incorporates several simulators together

I know, but this PR only concerns `ebos`, not `flow`. As said before: `ebos` is the *research* simulator and hence targeted at a convenient *development* experience, while `flow` is for *productive* use, i.e., it aims at more stability and a good *usage* experience...

> however, most of the times, developers is focusing on the threephase blackoil functionalities the most

IMO, developers should use `ebos` for their day-to-day work nowadays; it should currently work on exactly the same decks as `flow`.

> Making these variants to be built conditionally is just trying to reduce the time needs to re-compile when do make or make -j 8 or something similar which is the default option to build that includes all the tests that needs to be tested.

to only build the stuff which is required for the tests, use `make test_suite`. (maybe this does not work fully with the default build system, though.)

BTW: only compiling the `flow_ebos_blackoil.cpp` file needs 1 minute 5 instead of 1 minute 25 on my machine, i.e. besides confusing users, not too much would be gained by adding a `flow_blackoil` binary.

GitPaean:
> needs 1 minute 5 instead of 1 minute 25

20 seconds is significant, and it probably saves more on my machine.

andlaus:
> 20 seconds is significant, and it probably saves more on my machine.

I agree that 20 seconds is significant, but I hope that you agree that 32 seconds is more significant.

GitPaean:
> 32 seconds is more significant.

Agreed. 

What causes ebos is faster than flow_blackoil in term of compiling? 

In term of functionality, are you sure ebos works with the decks that flow is handling?

joakim-hove:
> In term of functionality, are you sure ebos works with the decks that flow is handling?

This is outside of my day-to-day competence, so please bear with me if I am talking bullshit - but anyways: I see it as extremely important and valuable for the long term performance of flow that it is the normal day-to-day workhorse for the developers; so in my opinion focus should rather be on reducing the compile time of flow.

GitPaean:
> so in my opinion focus should rather be on reducing the compile time of flow.

Yes.

andlaus:
> What causes ebos is faster than flow_blackoil in term of compiling?

I can only speculate, but I guess it is the layers build on top of the `ebos` infrastructure.

> In term of functionality, are you sure ebos works with the decks that flow is handling?

"sure" is too strong a term, but all I've tried so far worked fine...

andlaus:
> so in my opinion focus should rather be on reducing the compile time of flow.

I don't mind this. But this is a quite different matter than the PR at hand?!

GitPaean:
> I don't mind this. But this is a quite different matter than the PR at hand?!

In my opinion we should make the build of these variants conditional.

andlaus:
> In my opinion we should make the build of these variants conditional.

No objections if the default for building the ebos variants is made the same as building the tests in all modules, because both are only relevant for developers and IMO `ebos` is quite a bit more useful than the unit tests...

andlaus:
okay, build switches added.

andlaus:
jenkins build this please

andlaus:
BTW: If you consider the build time of flow important, don't you agree that the highly experimental `flow_ebos_oilwater_polymer_injectivity` specialization should not be build by default?

GitPaean:
> BTW: If you consider the build time of flow important, don't you agree that the highly experimental flow_ebos_oilwater_polymer_injectivity specialization should not be build by default?

Okay, since you bring up this point. The following is my understanding, 

more than 90%  (probably can be higher) of efforts and time are focused on threephase blackoil simulation only, and all the other variants are only needed occasionally.  To me, you need it 2% of the time or 0.5% of the time, it makes no difference. 

As it is said, the problem is not any of these variants or simulators. The problem is the way we are handling it. I bear in mind all the time that the compilation time is a problem now, as a result, I did not add an injectivity simulator for blackoil simulation. But if it is requested, we also need to add it. 

We need to find better solution instead of attacking any of them. With time, we will have more and more variants needed for sure, which is because we need it, not because we want to make the compiling more time-consuming.

andlaus:
> more than 90% (probably can be higher) of efforts and time are focused on threephase blackoil simulation only, and all the other variants are only needed occasionally. To me, you need it 2% of the time or 0.5% of the time, it makes no difference.

this is also my understanding. my question was rather: should these 0.5%-2% of usefulness that is provided by `flow_ebos_oilwater_polymer_injectivity` be allowed to almost double the build time for people building with 4 or 8 threads?

> We need to find better solution instead of attacking any of them. With time, we will have more and more variants needed for sure, which is because we need it, not because we want to make the compiling more time-consuming.

Sorry, I don't really get what you mean here either: When it comes to this, the only difference between `flow` and `ebos` that I can see is that the former links everything into the same binary, while the latter uses specialized executables  which are hence separately targetable by `make`. I also don't know what your plans for adding specializations are, but the way I envision this is that at most one variant per black-oil extension is added to the master branch and that combinations are kept locally on the interested persons' machines. (with exceptions being welcome as long as they are useful for a sufficiently wide audience.)

long term, it would probably be possible to assign the meaning of PDEs in the model dynamically even if their *number* stays compile time specified. This would probably be a multi-year project, though.

andlaus:
jenkins is happy and the requested cmake switch has been added. can this be merged? The discussion already went quite a bit OT IMO and can be continued after merge...

atgeirr:
> on my machine `make -j32 all` takes only 5 seconds longer with this PR than without it...

On my machine this will make the compile take a minute longer (I will check to be precise). It should definitely be conditional.

atgeirr:
> On my machine this will make the compile take a minute longer (I will check to be precise). It should definitely be conditional.

1m 37s difference.

totto82:
Maybe the build switch should be 
option(BUILD_EXTENTIONS "Build the research oriented extentions?" OFF)
and then all the variants of ebos are put under this switch.  
This will also allow us to remove some of the variants currently included in flow and make them only available explicit via flow_foo under this switch. 

Anyhow, I suggest we discuss this further in the next hangout.

andlaus:
@totto82: good idea, I'll have a look but this likely requires a patch against opm-common.
@atgeirr: sounds like your development machine is quite a bit underpowered. how many parallel build processes do you use?

atgeirr:
> sounds like your development machine is quite a bit underpowered. how many parallel build processes do you use?

8 (Intel i7 4 cores with HT). This is the laptop I use for everything... If I have to have a separate beefy desktop, bringing work home would be much less practical.

andlaus:
hm, interesting: I get almost the same results as you've reported if I limit my workstation to 4 cores. what's the result of `make clean && make -j8 ebos` versus `make clean && make -j8 flow` on your machine.

andlaus:
okay, I've excluded the non-default `ebos` from `make all` by default and added the possibility to also do this for `flow` and the default `ebos`. this requires OPM/opm-common#666 as a consequence.

andlaus:
jenkins build this opm-common=666 please

andlaus:
with jenkins being happy and no effect on the build on people who do not take explicit action, I cannot see any arguments why this cannot be merged. Anyone mind to press Green?

joakim-hove:
I am happy to push this through, but I did not really understand the interaction with opm-commom/666?

andlaus:
OPM/opm-common#666 is required if a target ought to be specified that is present if explictly requested, but not build by default. I haven't tested the present PR without it but I guess that without OPM/opm-common#666 it will work but all all ebos variants are simply build by default.

andlaus:
i.e., `make ebos_polymer` will always work even if a plain call to `make` won't build it...

totto82:
Thanks. With the opm-common PR merged and Jenkins happy I will merge this.

PR #1752, opened 2019-03-11
Fix bugs in the TUNING implementation in ebos

totto82:


totto82:
jenkins build this please

andlaus:
thanks. I've tested this and it seems to work as advertised. (with `ebos`, the "tuned" Norne is quite a bit slower than the untuned one, BTW.) I'll merge, since this does not affect the production simulator.

PR #1754, opened 2019-03-11
add an ebos variant which uses completely different numbering for phases and components

andlaus:
for some reason, this yields quite different results for norne than the default variant, e.g. when comparing PRESSURE, we get

```
> compareECL -k PRESSURE -t UNRST ebos/NORNE_ATW2013 ebos_altidx/NORNE_ATW2013 1 1e-4
Comparing 'ebos/NORNE_ATW2013' to 'ebos_altidx/NORNE_ATW2013'.
Comparing PRESSURE...
Occurrence in first file    = 9
Occurrence in second file   = 9
Value index                 = 0
(first value, second value) = (254.195, 253.191)

Program threw an exception: [/home/and/src/opm-common/build-cmake/fake-src/examples/test_util/EclRegressionTest.cpp:161] Deviations exceed tolerances.
The absolute deviation is 1.00311, and the tolerance limit is 1.
The relative deviation is 0.00394624, and the tolerance limit is 0.0001.
```

IMO this is a bug, but the reasons for it are currently mysterious to me.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ set( USE_OPENMP_DEFAULT OFF ) # Use of OpenMP is considered experimental
 option(BUILD_FLOW "Build the production oriented flow simulator?" ON)
 option(BUILD_EBOS "Build the research oriented ebos simulator?" ON)
 option(BUILD_EBOS_EXTENSIONS "Build the variants for various extensions of ebos by default?" OFF)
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this please

andlaus:
any objections to merging this? if no, I'll do so tomorrow.

andlaus:
silence is golden, merging.

PR #1756, opened 2019-03-13
Add support for water induced compaction 

totto82:
TODO: 
- [x] Add ROCK2DTR
- [x] Test results 
- [ ] Check that it doesn't change performance on other cases
- [ ] Clean the commit history

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -507,19 +507,19 @@ public:
                 soMax_[globalDofIdx] = elemCtx.simulator().problem().maxOilSaturation(globalDofIdx);
 
             if (swMax_.size() > 0)
-                swMax_[globalDofIdx] = elemCtx.simulator().problem().maxWaterSaturation(globalDofIdx);
+                swMax_[globalDofIdx] = elemCtx.simulator().problem().maxWaterSaturation(/*timeIdx=*/0, globalDofIdx);
 
             if (minimumPressure_.size() > 0)
-                minimumPressure_[globalDofIdx] = elemCtx.simulator().problem().minimumPressure(globalDofIdx);
+                minimumPressure_[globalDofIdx] = elemCtx.simulator().problem().minimumPressure(/*timeIdx=*/0, globalDofIdx);
 
             if (overburdenPressure_.size() > 0)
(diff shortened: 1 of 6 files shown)


ebos/ecloutputblackoilmodule.hh
```diff
@@ -830,6 +844,14 @@ public:
         if (bubblePointPressure_.size() > 0)
             sol.insert ("PBUB", Opm::UnitSystem::measure::pressure, std::move(bubblePointPressure_), Opm::data::TargetType::RESTART_AUXILIARY);
 
+
+        if (swMax_.size() > 0)
+            sol.insert ("SWMAX", Opm::UnitSystem::measure::identity, std::move(swMax_), Opm::data::TargetType::RESTART_AUXILIARY);
+
+        if (porvMultiplier_.size() > 0)
+            sol.insert ("PORV_MOD", Opm::UnitSystem::measure::identity, std::move(porvMultiplier_), Opm::data::TargetType::RESTART_AUXILIARY);
+
```
totto82: @joakim-hove I expected this to be sufficient to add PORV_MOD to the restart file, but it isn't. Do I need to modify something in opm-common or libecl?

joakim-hove: I'll look into it - what you have done *should* be enough, but ....

totto82: Changing it to RESTART_SOLUTION seems to work.

bska: > Changing [the `target`] to `RESTART_SOLUTION` seems to work.

Are you emitting strictly ECLIPSE-compatible restart files (`--enable-opm-rst-file=false`)?  If so, the vectors marked as `RESTART_AUXILIARY` get [dropped on the floor](https://github.com/OPM/opm-common/blob/7a2c1157a02259cad5deed554db08877f5af5b89/src/opm/output/eclipse/RestartIO.cpp#L499-L501)

totto82: @bska Thanks for the explaination.

andlaus: I guess it would be good if the output code would at least produce a warning in that case?

totto82:
jenkins build this ewoms=491 please

totto82:
jenkins build this ewoms=491 please

totto82:
jenkins build this ewoms=491 please

PR #1760, opened 2019-03-15
replacing StandardWell with StandardWellV, then renaming StandardWellV to be StandardWell.

GitPaean:
This PR is removing the duplication  between StandardWellV and StandardWell.


GitPaean:
benchmark

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 0.951    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.999    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 0.998    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 0.965    |



View result details @ https://www.ytelses.com/opm/?page=result&id=301

GitPaean:
benchmark

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 0.988    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.931    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.028    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 0.947    |



View result details @ https://www.ytelses.com/opm/?page=result&id=303

GitPaean:
It used to be that with `StandardWellV` and `StandardWell`, flow produces the same residual history. 

It looks like something changed later makes the residual history different and there is slightly different iteration behavoir, which makes the evaluation of the performance overhead not very easy because this overhead should only be within a couple of percents.

GitPaean:
benchmark

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.01     |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.971    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 0.997    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1        |



View result details @ https://www.ytelses.com/opm/?page=result&id=304

GitPaean:
benchmark

GitPaean:
jenkins build this please

ytelses:
Benchmark result overview: 




View result details @ https://www.ytelses.com/opm/?page=result&id=313

GitPaean:
Is there some system change in jenkins recently? First time I saw different test results of Jenkins with my old computer.

alfbr:
Nope, we still have some teething issues on the performance benchmarking. Working on it :-) BTW, in order for people to be able to comment about benchmarks here, we have now implemented "benchmark please" as the trigger.

ytelses:
Benchmark result overview: 




View result details @ https://www.ytelses.com/opm/?page=result&id=314

GitPaean:
benchmark please

GitPaean:
jenkins build this please

GitPaean:
I made this PR to be removing the StandardWellV. 

It is possible we should have ignored the performance overhead due to the original `DynamicEvaluation` before https://github.com/OPM/opm-material/pull/327 .  The performance overhead is difficult to be able to tested out with a real field model. With  https://github.com/OPM/opm-material/pull/327, the results can be slightly different from the results with the original `DynamicEvaluation`, possibly due to some compiler / hardware level difference. 

On the other side, there is something fragile with OPM, which makes small changes in float-point calculation be able to change the iteration process and possibly result in convergence failure or simulation failure.

andlaus:
> I made this PR to be removing the StandardWellV.

I'm a bit confused: do you remove `StandardWellV` or do you rename it to `StandardWell` and remove the current implementation? I assume the latter, but I haven't yet looked at the patch..

GitPaean:
> I'm a bit confused: do you remove StandardWellV or do you rename it to StandardWell and remove the current implementation? I assume the latter, but I haven't yet looked at the patch..

It is the latter. Sorry for the confusion.

GitPaean:
jenkins build this please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 0.958    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.981    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.016    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.123    |



View result details @ https://www.ytelses.com/opm/?page=result&id=318

GitPaean:

GitPaean:
jenkins build this opm-tests=103 please

GitPaean:

GitPaean:
jenkins build this opm-tests=103 please

GitPaean:
jenkins build this opm-tests=103 please

GitPaean:
it looks like `update_data` is not sufficient for unclear cause.

bska:
> it looks like `update_data` is not sufficient for unclear cause.

Update_data often does not work for existing PRs that have requested a data update previously.  I've occasionally been able to work around that by closing the source PR (that would be #1760 in this case) for some time&mdash;e.g., an hour&mdash;and then rerunning `update_data`, but there's absolutely no guarantee of that doing what I want.

GitPaean:
> I 've occasionally been able to work around that by closing the source PR (that would be #1760 in this case) for some time—e.g., an hour—and then rerunning update_data

thanks for the suggestion, I am trying.

GitPaean:

GitPaean:
jenkins build this opm-tests=103 please

GitPaean:
@bska , your approach works.  Thanks.

GitPaean:
I think should go in. I am removing the tag [for testing]. It has been a trouble to have two mostly identical classes related to standard wells.

bska:
> I think should go in. I am removing the tag [for testing].

So, just to be clear, you're saying that your original statement

>> this PR is not for merging. It is created for testing the performance overhead due to StandardWellV.

no longer applies?

GitPaean:
> > this PR is not for merging. It is created for testing the performance overhead due to StandardWellV.

> no longer applies?

I should have updated the statement, sorry for the confusion.  From the testing, the performance overhead is very minor for the real models and difficult to test out due to performance fluctuation and also changed residual history, while it did not fail any case we are supposed to run as far as I noticed.

totto82:
I agree with @GitPaean that the performance overhead from the dynamic version of the well model is not significant enough to pay for the maintenance of two versions. What worries my a bit is the ever growing complexity of the well model. But refactoring the well model is for another day. If you can verify that this doesn't significantly effect the convergence on "the prediction case" , this should be merged.

GitPaean:
> What worries my a bit is the ever growing complexity of the well model. But refactoring the well model is for another day. 

Yes. the well model needs to be decoupled to be smaller based on different roles of the different parts. Like some information is about the specification of the well, some is for the calculation or numerical part, like the matrices and vectors, and so on.  Before that, I think the WellState needs to be refactored to handle each individual well first. 

> If you can verify that this doesn't significantly effect the convergence on "the prediction case" , this should be merged.

I will do that.

atgeirr (approved):
I also agree that this should be merged after the final verification.

There is one small detail that bothers me, but that should be adressed separately and not in this PR: the syntax change
```C++
// before
EvalWell well_fraction = 1.0;
// after
EvalWell well_fraction(numWellEq_ + numEq, 1.0);
```
is not very nice, and I should have noticed when we merged the new DynamicEvaluation. First, the arguments are easily swapped by mistake, and I would expect the value argument to be first. Second, when storing a constant we should be able to use an evaluation with no derivatives to represent it, allowing us to construct the constant without also passing in the number of derivatives. This will likely require some fiddling in the DynamicEvaluation class.

GitPaean:
> would expect the value argument to be first. 

I shared the same opinion https://github.com/OPM/opm-material/pull/302#discussion_r198462926 , while obviously people do have different opinions here. 

>  Second, when storing a constant we should be able to use an evaluation with no derivatives to represent it, 

I agree, while for this specific `well_fraction` case, we need to return it with the number of primary variables.  `EvalWell well_fraction = 1.0;` will still generate derivatives, while the number of the derivatives is known during compilation.

GitPaean:
> If you can verify that this doesn't significantly effect the convergence on "the prediction case" , this should be merged.

The testing results are bad. It is broken in the master branch.  Master branch and this PR both stop at the same report step for the same reason.

totto82:
Does your master includes OPM/opm-common#705?

atgeirr:
> The testing results are bad. It is broken in the master branch.

That is worrying, and we need to give this very high priority!

> Does your master includes [OPM/opm-common#705](https://github.com/OPM/opm-common/pull/705)?

Sounds like a good place to start investigating.

andlaus:
>> would expect the value argument to be first.
>
> I shared the same opinion OPM/opm-material#302 (comment) , while obviously people do have different opinions here.

the reason I insisted on this is to make the order the same as for `std::vector`: Even if you can argue that the reverse order is more logical, it is just not worth the confusion...

GitPaean:
> Does your master includes OPM/opm-common#705?

yes.

GitPaean:
> the reason I insisted on this is to make the order the same as for std::vector: Even if you can argue that the reverse order is more logical, it is just not worth the confusion...

`std::vector` is different. All the elements in a `std::vector` are kind of in a equal position. while for `Evaluation`, its first element is the value, the rest is the derivatives.

andlaus:
> `std::vector` is different.

obviously; it is for a different purpose. but the use case of specifying a size plus a value is similar enough for me not wanting to be confused.

GitPaean:
> obviously; it is for a different purpose. but the use case of specifying a size plus a value is similar enough for me not wanting to be confused.

value is for the value, the number is the number of the derivatives.

GitPaean:
> That is worrying, and we need to give this very high priority!

The failure exception is from potential calculation, which makes it less worrying. But we need to check what difference makes it a problem now while not before.

GitPaean:

GitPaean:
jenkins build this opm-tests=103 please

GitPaean:
It is closed temporary to `update_data` later. It looks possible that `update_data` has problems in obtaining most updated source code.

GitPaean:

GitPaean:
jenkins build this opm-tests=103 please

GitPaean:
> The failure exception is from potential calculation, which makes it less worrying. But we need to check what difference makes it a problem now while not before.

By commenting out the potential calculation, one well gets shut while not be able to be re-opened, this changes the later results of a few other wells.  This is unfortunate.

GitPaean:
jenkins build this please

GitPaean:

GitPaean:

GitPaean:
jenkins build this opm-tests=103 please

GitPaean:
jenkins build this opm-tests=103 please

bska:
> jenkins build this opm-tests=103 please

PR OPM/opm-tests#103 was created before we merged OPM/opm-common#707. Therefore the former's `.INIT` files don't account for the updated `TAB` contents from the latter. You will, at minimum, have to re-update the reference solutions in 103.

GitPaean:
> PR OPM/opm-tests#103 was created before we merged OPM/opm-common#707. Therefore the former's .INIT files don't account for the updated TAB contents from the latter. You will, at minimum, have to re-update the reference solutions in 103.

Thanks for the information. I did not notice that.

GitPaean:

GitPaean:
jenkins build this opm-tests=103 please

GitPaean:
jenkins build this opm-tests=103 please

GitPaean:
jenkins build this please

GitPaean:
jenkins build this opm-tests=103 please

GitPaean:

GitPaean:
benchmark please

GitPaean:

GitPaean:
```
{standard input}: Assembler messages:
{standard input}:71214: Warning: end of file not at end of a line; newline inserted
c++: internal compiler error: Killed (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-6/README.Bugs> for instructions.
Recording test results
```

GitPaean:

bska:
> ```
> {standard input}: Assembler messages:
> {standard input}:71214: Warning: end of file not at end of a line; newline inserted
> c++: internal compiler error: Killed (program cc1plus)
> ```

Just a word of caution: I've noticed that when those messages have appeared recently, relaunching the build shortly after would likely run into the same symptoms; especially when there are other build running on the CI system and especially if any of those other builds are also targeting the same module.  If the condition recurs I advice waiting a while before relaunching.  I don't have a plausible explanation for what the underlying issue might be, but recently we've been running into something that at least on the face of it looks like race conditions on the CI system.

GitPaean:
jenkins build this opm-tests=103 please

GitPaean:

GitPaean:
In 4c52c430ed8c3dd0e59b151459b8f92562057495 , the same invertMatrix function is created for 4 X 4 matrix for DynamicMatrix. The reason to do that to make sure for blackoil cases, the new StandardWell model will produce the same result with the master branch.  So it will relieve us from the heavy burden of testing of this PR with all the real blackoil cases we have. 

It creates some duplicated code. Later we need to address whether to keep the overloading of the invertMatrix function or just use the version based on LU decomposition provided by Dune. 

It will be much easier to test and review them in separate PRs. 

Due to possible different invert() implementation for 3X3 matrices for `FieldMatrix` and `DynamicMatrix` in Dune, there will be still some failures related to 2-phase cases.  But they are much less and much easier to check. 

Unless there is something shows up in the jenkins test, this PR is ready for review and merging from my side.

GitPaean:
The failure of `mpi.compareECLFiles_flow+2D_THREEPHASE_POLY_HETER` is a surprise. Will investigate it.

GitPaean:
jenkins build this opm-material=338 please

GitPaean:
> The failure of mpi.compareECLFiles_flow+2D_THREEPHASE_POLY_HETER is a surprise. Will investigate it.

It is fixed through OPM/opm-material#338

GitPaean:
jenkins build this opm-material=338 please

GitPaean:
One question, for the above jenkins request, will jenkins need to re-compile opm-common?  

https://ci.opm-project.org/job/opm-simulators-PR-builder/1570/ is rebuilding opm-common. 

Jenkins has been facing some challenges, maybe this is one symptom related to the problem?

But if the answer is yes, then it is not that relevant.  @akva2 .

GitPaean:
```

Test Name | Duration | Age
-- | -- | --
mpi.compareECLFiles_flow+3D_MSW | 3.4 sec | 2
mpi.compareECLFiles_flow+MSW_MODEL_1 | 1 sec | 2
mpi.compareECLFiles_flow+FAULTS_MODEL_1 | 1 sec | 2
mpi.compareECLFiles_flow+SPE1CASE2_OILGAS | 0.65 sec | 5
mpi.compareECLFiles_flow+2D_OILWATER_POLYMER | 0.97 sec | 5

```

failure related to `3D_MSW, MSW_MODEL_1 and FAULTS_MODEL_1` are not expected and I could not reproduce them locally.

GitPaean:
jenkins build this opm-material=338 opm-tests=124 please

GitPaean:
jenkins build this opm-material=338 opm-tests=124 dune2.4 please

GitPaean:
Some information might be relevant or not.  Locally, with dune 2.6 and dune 2.5.2, all the tests pass without necessity to update opm-tests. 

With dune 2.4.2, opm-tests need OPM/opm-tests#124 to pass. 

The difference I can recall here is the different implementation of invert() function for DynamicMatrix, maybe also for FieldMatrix. 

My compiler version is 5.4.  Having no idea what is the reason why the three tests related to Model_1 fails. It looks like not related to dune version.

andlaus:
>> The failure of mpi.compareECLFiles_flow+2D_THREEPHASE_POLY_HETER is a surprise. Will investigate it.
>
> It is fixed through OPM/opm-material#338

As I stated there, this is IMO rather a small bug in the well model because it fails to initialize a variable explicitly...

GitPaean:
This PR does not depend on https://github.com/OPM/ewoms/pull/517, while it requires https://github.com/OPM/ewoms/pull/517 to make sure to obtain well defined behavoir for the black oil polymer case with shear effects. 

This PR should not change any results of the black oil cases. 

it will change the results of the two phase cases due to the different way of inverting function between FieldMatrix and DynamicMatrix. The behavoir will also be different between different DUNE version due to the change of implementation of the invert function.

GitPaean:
jenkins build this ewoms=517 please

GitPaean:
As expected, only the following two twophase cases failed. 

```
mpi.compareECLFiles_flow+SPE1CASE2_OILGAS
mpi.compareECLFiles_flow+2D_OILWATER_POLYMER
```

GitPaean:

GitPaean:
Local tests shows it does not change the result in anyway for blackoil simulations, which are the cases we care the most.

GitPaean:
jenkins build this opm-tests=124 please

GitPaean:
jenkins build this opm-tests=103 please

GitPaean:
`update_data` did not work properly, so with https://github.com/OPM/opm-tests/pull/103 , `2D_THREEPHASE_POLY_HETER` test failed.

GitPaean:
jenkins build this opm-tests=124 please

atgeirr:
Finally green! And with no changes to blackoil runs. I'll merge.

PR #1763, opened 2019-03-20
added possibility to get well debugging info

fgfuchs:
setting --well-debug-verbosity-level to an integer > 0 will trigger logging debug information on wells in a file called WELLOUTPUT.WDBG.


fgfuchs:
jenkins build this please

OPMUSER:
I would suggest you write the debug output to the standard debug file to avoid confusion and the overwriting exiting files

GitPaean:
> I would suggest you write the debug output to the standard debug file to avoid confusion and the overwriting exiting files

It is something we have discussed. With different level of debugging output, the output can become massive and tedious, which can make the information in the original .DBG not readable. And also, we have not figured out what is the best way to output debug information for well model, which depends on how we use. At least for the moment, not intervening with the existing facility can be a good idea. But investing efforts to improve the debugging facility for well model will be certainly beneficial for later. 

Anyway, it is something people can comment and discuss.

OPMUSER:
Why not use the DEBUG keyword to determine the type of output to the debug file? This would be completely under the developers control (that is nothing to do with the commercial simulator).  I just don't like the idea of creating additional files when there is a file for this purpose.

Just my two cents.

fgfuchs:
@OPMUSER, I see your point. As I understand, the file is only created when --well-debug-verbosity-level is used and set to an integer bigger than 0. So the only time this is activated is when one wants to see exactly the output that is put into the extra file, alleviating grepping an parsing time. In all other cases **no** extra file is created.

alfbr:
It this is a temporary solution, then anything is OK I guess. The problem is if this is meant as a permanent option for output. While adding on an extra file is quick to implement, it is not intuitive for anybody else, and it breaks with the current approach. Hence, if this debug output is intended to be merged into master, made available for all developers in flow, then we should probably invest the time to find a good solution for everybody.

andlaus:
@fgfuchs: you could make this functionality conditional on `enableExperiments`. In this case, `flow` will be completely unaffected so @alfbr and @OPMUSER should be happy. (you will have to use `ebos` to do the development which you need this for, though.)

atgeirr:
> While adding on an extra file is quick to implement, it is not intuitive for anybody else, and it breaks with the current approach

I do not see how it breaks the current approach, as it is only created if explicitly requested, and the information in it is going to be really detailed and developer-only. No user is ever going to see it unless we ask them specifically increase this verbosity level to help with some debugging. I see the point about using the existing DBG file for this, but these files grow into multi-gigabyte monsters with the amount of debugging info we need and produce, which renders them very hard to use. A separate file for dumping the matrices etc. allows the DBG file to be still useful for following the convergence progress etc.

alfbr:
I understand, and an extra file may very well be the practical solution. All I am saying is that this is worth a bit of extra discussions, because there may be better solutions in the larger picture.

OPMUSER:
I still think implementing the DEBUG keyword and using the DBG file is the way to implement developer debug information. Once implemented this would be easy to expand to various other debug information as well as giving more flexibility, for example on could set the debug output at a given time and then switch it off:
```
--
--     SET DEBUG OUTPUT OPTIONS
--
DEBUG                                                                           
       8*0  1  11*0  1  30*0                                                   /     

DATES                                                                           
1  JAN   2020  /                                                                
/                                                                               
                                                                                
RPTSCHED                                                                                                                             
         'NOTHING'                                                             /                                                    
                                           
DATES                                                                                                                                                                                                                          
        1  APR   2020  /                  
        1  JLY   2020  /                  
        1  OCT   2020  /                  
/
--
--     SET DEBUG OUTPUT OPTIONS TO ALL OFF
--
DEBUG                                                                           
/     
--
--       FINAL REPORT AND RESTART AT YEAR END
--
RPTSCHED                                                                                                                             
          'WELLS=2'    'WELSPECS'    'CPU=2'     'FIP=2'                        /
 
RPTRST                                                                           
         'BASIC=2'                                                             /
 
DATES                                                                           
         31 DEC   2020  /                                                                
/
```                                                                                      
Thus, one could reduce the output to the time period of interest, rather than running the whole deck (although the END keyword could be used to terminate the run as well).

In addition, it would eliminate the need of yet another command line parameter, as the command line augments  seems to be growing with each release as developers add more debug/features. I understand why this the preferred way to implement additional debug output as the infrastructure is already in place, but once the DEBUG keyword is implemented then this will also be the case.   

One last point, by using the DEBUG keyword it will result in a more "disciplined" approach to incorporating developer debug information.

Just a suggestion.

GitPaean:
> I still think implementing the DEBUG keyword and using the DBG file is the way to implement developer debug information.

I like the feature that we can turn on and off the output of the debug information for certain report steps  shown in your example. How to organize and use the debugging information will be the key, which we are not sure and having different opinions. 

And also, we should also avoid break the ECL running of the deck. (may be not mandatory?)

bska:
@fgfuchs : Do you wish to revive this work?

atgeirr:
I think that implementing the DEBUG keyword (or maybe something like OPMDBG) is probably the best solution to the general issue of enabling/disabling debug output. However, there are some useful output functions that could be scavenged from this PR, and put in a separate one (for manual/debugger use until we implement DEBUG).

PR #1764, opened 2019-03-21
Move events

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/688

### DIFF
--- a/opm/autodiff/StandardWellV_impl.hpp
+++ b/opm/autodiff/StandardWellV_impl.hpp
@@ -671,6 +671,7 @@ namespace Opm
             // Compute Productivity index if asked for
             const auto& pu = phaseUsage();
             const Opm::SummaryConfig& summaryConfig = ebosSimulator.vanguard().summaryConfig();
+            const Opm::Schedule& schedule = ebosSimulator.vanguard().schedule();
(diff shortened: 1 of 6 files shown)


opm/autodiff/StandardWellV_impl.hpp
```diff
@@ -679,7 +680,8 @@ namespace Opm
                     const unsigned int compIdx = flowPhaseToEbosCompIdx(p);
                     const double drawdown = well_state.perfPress()[first_perf_ + perf] - intQuants.fluidState().pressure(FluidSystem::oilPhaseIdx).value();
                     double productivity_index = cq_s[compIdx].value() / drawdown;
-                    scaleProductivityIndex(perf, productivity_index, deferred_logger);
+                    bool new_well = schedule.hasWellEvent(name(), ScheduleEvents::NEW_WELL, current_step_);
```
atgeirr: Please constify new_well here (also in StandardWellsV), and where used as function argument (apart from correctness issues, that is the style of the existing code).

joakim-hove: OK - done; Jenkins was green - I'will not rerun after the `const` fixup.

Issue #1453, opened 2018-04-10
Flow Crashes in Checked Iterator/Debug Build Running SPE1CASE1


bska:
It's been a while since I tried to run a checked iterator/Debug build of flow so I don't know if this is a new issue or if it simply has not been reported before.

Earlier today I tried to run
```
${flow} deck_filename=${spe1case1} output_dir=./spe/1/case1
```
with `${flow}` pointing to a checked iterator build (`-O0 -ggdb3 -D_GLIBCXX_DEBUG=1 -D_GLIBCXX_DEBUG_PEDANTIC=1`) of `flow` and `${spe1case1}` being the full path to [SPE1CASE1.DATA](https://github.com/OPM/opm-data/blob/master/spe1/SPE1CASE1.DATA) from the opm-data repository.

This simulation run crashed on the second time-step (i.e., "ministep" index 1 within report step 0) with an "index out of bounds" exception on line [1091](https://github.com/OPM/ewoms/blob/master/ebos/eclproblem.hh#L1091) of `ewoms/ebos/eclproblem.hh`.  If I locally disable the portion of `Ewoms::EclProblem<>::endTimeStep()` that [discards the `initialFluidStates_`](https://github.com/OPM/ewoms/blob/master/ebos/eclproblem.hh#L661-L663), then I can run the checked iterator build to completion.  I do however not know if disabling that section is just papering over some other issues so I'd like a little input from parties that know that area of the code better than I do.

atgeirr:
To me this looks like a simple bug: the temperature() method should probably access some potentially time-varying field instead of the initial temperature, that looks like a placeholder from before the thermal stuff was implemented. That said, why is this method called at all when running SPE1?

What do you think @andlaus?

andlaus:
> That said, why is this method called at all when running SPE1?
>
> What do you think @andlaus?

My first hunch is that it is related to output. I'll investigate.

bska:
>> That said, why is this method called at all when running SPE1?
> 
> My first hunch is that it is related to output. I'll investigate.

Just for reference, I uploaded a GDB backtrace to [Gist](https://gist.github.com/bska/9bd81d3a03e3c4866e516cbead138ef7).  Frame 11 stands out as being a call to
```
Ewoms::FvBaseAdLocalLinearizer<Ewoms::Properties::TTag::EclFlowProblem>::linearize
```
That looks a bit like we end up in `temperature()` when assembling the linear operator during a non-linear iteration.

[...]

PR #1765, opened 2019-03-21
Minor cleanups

andlaus:
remove the unused `initialTemperature_` array in `EclProblem` and do not use the uninitialized `isIoRank_` member in `ISTLSolverEbos` anymore.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1321,12 +1321,10 @@ public:
     template <class Context>
     Scalar temperature(const Context& context, unsigned spaceIdx, unsigned timeIdx) const
     {
-        // use the temporally constant temperature, i.e. use the initial temperature of
-        // the DOF
+        // use the initial temperature of the DOF if temperature is not a primary
+        // variable
         unsigned globalDofIdx = context.globalSpaceIndex(spaceIdx, timeIdx);
-        if (!initialFluidStates_.empty())
-            return initialFluidStates_[globalDofIdx].temperature(/*phaseIdx=*/0);
-        return initialTemperature_[globalDofIdx];
+        return initialFluidStates_[globalDofIdx].temperature(/*phaseIdx=*/0);
     }
 
     /*!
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this please

bska:
> remove the unused `initialTemperature_` array in `EclProblem`

Is #1453, which resulted in OPM/ewoms#311, no longer an issue?

andlaus:
I bet a beer that this PR does not change the state of this in any regard: if you look at the changes, you might notice that the `initialTemperature_` array was never initialized anywhere.

bska:
>> Is #1453 [...] no longer an issue?
>
> This PR does not change the state of this in any regard:

That's not what I asked.  I asked if the reason why introduced the data member in the first place still exists.  If the issue is still present, then we need to keep the data member (and initialise it properly).

andlaus:
> I asked if the reason why introduced the data member in the first place still exists.

As stated above: I'm reasonably sure that the answer to that question is "no".

bska:
> > asked if the reason why introduced the data member in the first place still exists.
> 
> As stated above: I'm reasonably sure that the answer to that question is "no".

That's not a good enough answer.  You need to be able to point to a sequence of changes that explicitly solve the problem.

andlaus:
is the statement "the `initialTemperature_`  array was not initialized and used and using it would have resulted in a segmentation fault" a "good enough" answer for you?

bska:
> is the statement "the `initialTemperature_` array was not initialized and used and using it would have resulted in a segmentation fault" a "good enough" answer for you?

It's not.

There are at least two possible reasons for the member to be unused.  One is that it was inadvertently dropped in a refactoring and the other is that it was intentionally dropped because the underlying issue (accessing non-existent temperature data during system assembly of non-thermal problems) was solved differently.  If it is the former then the original, underlying problem still exists and we need to deal with it somehow.  If it is the latter then we can safely remove the data member.

andlaus:
interesting theories. I'd like to request a reversal of the "burden of proof".

joakim-hove:
image

But seriously guys - we have a testsuite?!

bska:
Just for context: The test suite did not catch the original issue, #1453.

joakim-hove:
> Just for context: The test suite did not catch the original issue, #1453.

Well - then that is where focus should be.

totto82:
Jenkins is happy, so I will merge. Thanks for the clean up.

PR #1767, opened 2019-03-21
Version of cpr amg which can reuse setup and also change smoothers of…

hnil:
Added a version of the most robust and fastest method so far. For norne and model2
--system-strategy=trueimpes is better however this fails for CO2CASE.

This version is purely serial. Any help to make it flexible for parallel is wellcome..


GitPaean:
for testing

atgeirr:
Replaced, closing.

PR #1768, opened 2019-03-22
ebos: add some checks to make sure that the simulator produces what the deck specifies

andlaus:
this should significantly reduce surprising behavior for users. since this is an experimental feature, `flow` is unaffected.


andlaus:
jenkins build this please

andlaus:
since Jenkins is happy and this does not affect `flow`, I'll also self-merge this.

PR #1769, opened 2019-03-22
ebos: add a stub implementation for API tracking

andlaus:
the only thing which this does so far is to introduce the respective property and `ebos` will abort the run if the deck requests API tracking.

As usual for experimental features, `flow` is unaffected.


andlaus:
jenkins build this please

PR #1770, opened 2019-03-22
Cater for empty fipnum vectors in output code.

blattms:
Due to loadbalancing this vector might be empty under some circumstances.
Hence the old code would dereference the `end` iterator which is a bug.


bska:
jenkins build this please

bska:
This seems to have fallen by the wayside.  As far as I can tell, this fixes a real problem.  Unless anyone objects I will merge this into master tonight CEST.

totto82:
Looks good. Thanks for the fix.

PR #1772, opened 2019-03-26
ebos: make its core headers self sufficient

andlaus:
this is part of the release maintainance. in this context "core ebos headers" means the headers located underneath the `ebos` directory which do not include the well model headers, and only those which are concerned with non-exotic functionality, e.g., the PolyhedralGrid and ALUGrid vanguards are not changed.


andlaus:
jenkins build this please

PR #1774, opened 2019-03-28
Support THPRESFT

andlaus:
this is the simulator side of OPM/opm-common#693. Since this is an experimental feature, `flow` is unaffected.

### DIFF
--- a/ebos/eclthresholdpressure.hh
+++ b/ebos/eclthresholdpressure.hh
@@ -56,6 +56,7 @@ NEW_PROP_TAG(Scalar);
 NEW_PROP_TAG(Evaluation);
 NEW_PROP_TAG(ElementContext);
 NEW_PROP_TAG(FluidSystem);
+NEW_PROP_TAG(EnableExperiments);
 
 END_PROPERTIES
 
@@ -80,6 +81,7 @@ class EclThresholdPressure
     typedef typename GET_PROP_TYPE(TypeTag, ElementContext) ElementContext;
     typedef typename GET_PROP_TYPE(TypeTag, FluidSystem) FluidSystem;
 
+    enum { enableExperiments = GET_PROP_VALUE(TypeTag, EnableExperiments) };
     enum { numPhases = FluidSystem::numPhases };
 
 public:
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this please

ebos/eclthresholdpressure.hh
```diff
@@ -307,6 +337,45 @@ private:
...
+
+        // extract the multipliers from the deck keyword
+        int numFaults = faults.size();
+        int numCartesianElem = eclState.getInputGrid().getCartesianSize();
+        thpresftValues_.resize(numFaults, -1.0);
+        cartElemFaultIdx_.resize(numCartesianElem, -1);
+        for (size_t recordIdx = 0; recordIdx < thpresftKeyword.size(); ++ recordIdx) {
+            const Opm::DeckRecord& record = thpresftKeyword.getRecord(recordIdx);
+
+            const std::string& faultName = record.getItem("FAULT_NAME").getTrimmedString(0);
```
joakim-hove: Whenever I get the chance I advocate that things like this should be in the `EclipseState` - and that the `Deck` should not be accessed outside of opm-common; the region based `THPRES` already has an implementation in the EclipseState object.

andlaus: okay, how to get to this information only using the `eclipseState`, then? I propose that this stays as-is until the parser properly internalizes THPRESFT. Also, be aware that this is clearly marked as an experimental feature.

joakim-hove: Well - you implement support in the EclipseState :-)

joakim-hove:
>Since this is an experimental feature, flow is unaffected.

Well - I guess the moment it is merged it is suddenly a supported feature? Or how is the in-repository upgrade path from experimental to production code?

andlaus:
> Well - I guess the moment it is merged it is suddenly a supported feature?

it will stay an experimental feature because it will stay disabled in `flow`. If you want to use it, you have to use a research simulator, i.e., `ebos` or one of its variants. in my current understanding, there are no guarantees about the results which `ebos` produces, so I would still count such the features as "unsupported".

andlaus:
> Or how is the in-repository upgrade path from experimental to production code?

basically somebody has to be willing to take the blame if something blows up by taking the code in question out of experimental mode.

ebos/eclthresholdpressure.hh
andlaus: Interesting theory ;)

joakim-hove: > Interesting theory ;)

Yes - a quite good plan; and frankly not a joke - although I see you take that way!

totto82: > Whenever I get the chance I advocate that things like this should be in the EclipseState - and that the Deck should not be accessed outside of opm-common; the region based THPRES already has an implementation in the EclipseState object.

I guess I am guilty in violating the "not access the deck directly outside opm-common" rule several times. For simple keywords I love the way you can just add a JSON description and just use it in the simulator right away. For more complicated stuff when different keywords interact I think the "rule" should be obeyed strictly. 

I am open for merging this as is. But I also think we should have a discussion about the usage of the deck outside opm-common in general.

totto82:
I will merge this now if nobody objects loudly.

PR #1751, opened 2019-03-08
various improvements/corrections to the multisegment well model

GitPaean:
And also trying to activate the multisegment well model by default. 

With multisegment well model, more parameters can be tuned and tested, there might be some small updates base on the testing result. 

Creating this PR to warp up the efforts to improve the multisegment well model for now.

PR #1775, opened 2019-03-28
flow: make sure that the ebosSimulator's episode machinery is up to date

andlaus:
if it isn't using the current episode index to query data from the deck will result in tears...


andlaus:
jenkins build this please

andlaus:
jenkins build this please

andlaus:
jenkins build this please

andlaus:
jenkins build this please

andlaus:
I did a bit of headbanging with the nearest wall and I'm now pretty convinced that the new result is the correct one, i.e. the test failure for SPE5CASE1 is probably because the master is wrong.

@totto82: can you have a look as well?

andlaus:

andlaus:
I finally found the issue (I hope): it is a shenanigan in the adaptive time stepper's `suggestedNextStep()` method.

andlaus:
jenkins build this please

andlaus:
jenkins build this please

andlaus:
this turned out to be a much deeper rabbit hole than I thought: for the last three days and the better part of Friday, I struggled with beating `flow` into shape and the episode size is still not correct because this requires a data update. a patch for this will coming once the present PR is in...

andlaus:
jenkins build this ewoms=500 please

andlaus:
jenkins build this ewoms=500 please

ebos/eclproblem.hh
```diff
@@ -2684,31 +2699,27 @@ private:
...
-        dtNext = std::min(dtNext, remainingEpisodeTime);
+            simulator.episodeStartTime() + simulator.episodeLength()
+            - (simulator.startTime() + simulator.time());
+        assert(remainingEpisodeTime >= 0.0);
 
         // if we would have a small amount of time left over in the current episode, make
         // two equal time steps instead of a big and a small one
-        if (dtNext < remainingEpisodeTime*(1.0 - 1e-5) && dtNext*1.25 > remainingEpisodeTime)
+        if (remainingEpisodeTime/2.0 < dtNext && dtNext < remainingEpisodeTime*(1.0 - 1e-5))
             // note: limiting to the maximum time step size here is probably not strictly
```
totto82: What is the reasoning behind changing this?

ebos/eclproblem.hh
```diff
@@ -2684,31 +2699,27 @@ private:
...
     // this method applies the runtime constraints specified via the deck and/or command
-    // line parameters for the size of the next time step size.
+    // line parameters for the size of the next time step.
     Scalar limitNextTimeStepSize_(Scalar dtNext) const
     {
-        const auto& events = this->simulator().vanguard().schedule().getEvents();
-        int episodeIdx = this->simulator().episodeIndex();
+        if (!enableExperiments)
+            return dtNext;
+
```
totto82: Using !enableExperiments for flow spesific stuff is in my opinion not a good strategy. Or have I misunderstood this line?

ebos/eclproblem.hh
```diff
@@ -2183,7 +2189,7 @@ private:
         initialFluidStates_ = tmpInitialFs;
         // make sure that the stuff which needs to be done at the beginning of an episode
         // is run.
-        this->beginEpisode(/*isOnRestart=*/true);
+        this->beginEpisode();
```
totto82: why did you remove the (/*isOnRestart=*/true)?

ebos/eclproblem.hh
andlaus: the idea is that two identical time steps should be taken earlier if the remaining time of the current episode after the current time step would be too short. note that this change only affects `ebos`. (also, there could be a mistake in the logic here; I changed this back and forth about a gazillion times so a wrong version might have ended up in the final patch.)

ebos/eclproblem.hh
andlaus: the problem is that this is called deeply within `flow` and it causes the `remainingEpisodeTime` to be smaller than 0 at some occasions. Since the result for is ignored for `flow` anyway, I gave up on fixing this. This statement applies for a work-in-progress version of the patch, so maybe it is not required anymore...

ebos/eclproblem.hh
andlaus: for some reason, it is not required and the `beginEpisode()` callback should not have such a parameter in the first place?!

totto82:
I did some testing on this on the big models and this patch both changes the convergence slightly (is that expected?) and significantly slows down the simulator. I only looked at the reported simulation time, could it be that the reported simulation time is messed up in this PR?

andlaus:
> is that expected?

no

> [...] and significantly slows down the simulator.

bugger! is this only for `flow` or also for `ebos`?

> I only looked at the reported simulation time, could it be that the reported simulation time is messed up in this PR?

everything is possible, but I consider it unlikely that the simulation time is incorrect. coul you look at the diff betwen the DBG files?

totto82:
> I did some testing on this on the big models and this patch both changes the convergence slightly (is that expected?) and significantly slows down the simulator. I only looked at the reported simulation time, could it be that the reported simulation time is messed up in this PR?

Sorry, unintentionally different compiler flags where used. With the same flags the performance and also the convergence doesn't change.

GitPaean:
>  also the convergence doesn't change.

If the residuals are the same, then it should be fine.

GitPaean:
> unintentionally different compiler flags

Just out of curiosity, could you also share about the `compiler flags`? I was struggling with `compiler flags` when I changed computer. Finally, I removed `-march=native -mtune=native`, it is much better. Actually now, I manage to have the same testing results with jenkins. 

However, when I was struggling to get the exactly same residuals, it is mostly just different residual history and iteration process changes a little bit and there.  For the case `flow` can run through, it did not impact the performance much. 

I am wondering what is happening here with `compiler flags` which made the performance in a significant way.

totto82:
@GitPaean  Everything is the same as expected. This touches the well model (come unrelated cleanup) should I wait merging it to after #1751?

GitPaean:
> @GitPaean Everything is the same as expected. This touches the well model (come unrelated cleanup) should I wait merging it to after #1751?

#1751 is going in soon. So yes, this PR should wait a little bit after.

totto82:
I added -fsanitize=address for some other testing and forgot to remove it.

andlaus:
#1751 did not conflict with this PR on my machine -- at least I did not have to manually resolve any conflicts...

andlaus:
(I get lots of new compiler warnings due to #1751, though. I guess they should be fixed before it goes in.)

GitPaean:
> (I get lots of new compiler warnings due to #1751, though. I guess they should be fixed before it goes in.)

Thanks for the remind. I will handle them.

GitPaean:
> I added -fsanitize=address for some other testing and forgot to remove it.

I am reading about this, it looks like related to memory error detector. Curious why it changed the results.

andlaus:
okay, with #1751 merged, this should be able go in without potentially breaking stuff for other people. maybe only next week, though?!

akva2:
just fyi: release branches have been cut so that is not stopping anything.

andlaus:
cool, thanks for the info. possibly we want keep the diff of the release branches to master small until the release happened, though? (this would indicate an even longer delay.)

andlaus:
I've rebased this on  the current master. now that the 2019.04 release is basically out of the door, it is IMO time to merge things like this.

andlaus:
jenkins build this please

totto82:
Thanks for the clean-up. The difference in time management between flow and the underlying structure in ewoms has hit me several times.

PR #1777, opened 2019-03-29
Update missing features

joakim-hove:
See: https://github.com/OPM/opm-common/issues/701

The RPTSCEHD, WTEST and EDITNNC are (to best of my knowledge) fully supported all the way through and should be remowed from this list. The ECHO and NOECHO keywords are not supported - but if/when we are to support them that will be in the input layer, so it is really not the simulators business to complain about that.


atgeirr:
jenkins build this please

PR #1778, opened 2019-04-01
Fix cpr oil gas

hnil:
Fixing index for pressure equation used in gasoil systems.


totto82:
jenkins build this please

totto82:
Thanks for the fix. I haven't tested this, but I guess it is reasonable to pick the equation for the least compressible fluid. i.e. oil when you don't have water.

PR #1788, opened 2019-04-10
Amg cpr reusing hierarchy and supporting dune 2.5

blattms:
This is #1787 rebased and fixed for DUNE 2.5

The reste of the description of #1787 applies


blattms:
jenkins build this please

blattms:
jenkins build this please

atgeirr:
Thanks a lot! I'll merge now.

bska:
Just out of curiosity, does the project have a set of declared minimum requirements for third-party libraries?

The reason I ask is that I was not able to get this to compile locally in a configuration that was based on Dune 2.4.1.  The build failure referenced the type [`Dune::DirectSolverSelector<>`](https://github.com/OPM/opm-simulators/blob/628be79cfec3d8ef62855959003dfef8191df36b/opm/autodiff/amgcpr.hh#L487) which seems to be new in Dune 2.5 (I haven't really checked).

As another piece of information, our build system looks for [Boost 1.44 or newer](https://github.com/OPM/opm-simulators/blob/628be79cfec3d8ef62855959003dfef8191df36b/opm-simulators-prereqs.cmake#L24-L25) which probably has not been exercised in a long time.  I personally use Boost 1.58.

bska:
On a second note, this just failed the post-merge build check on Jenkins.  That **probably** means there's a piece of the PR that does not account for `!HAVE_MPI`, because the post-merge check builds both sequential and parallel versions whereas the regular "build this" command only builds the MPI version.  The failing build was `configuration=serial`.

Full build log at https://ci.opm-project.org/job/opm-simulators/704/consoleFull.

blattms:
Thanks for the heads up. Will fix.

blattms:
Will need some further testing but #1789 should nearly go all the way

blattms:
BTW I thought I test sequential, but i was so intelligent as to deactivate building the test which means the cpr stuff was never built. Always worth double checking...

PR #1792, opened 2019-04-12
[for testing]  Testing the development on MSW

GitPaean:
The tolerances are relaxed.


GitPaean:
The second commit should be cherry-picked to the master branch since it makes us run through a real case. In this case, during the iteration, the BHP just keep reducing until negative, which crashes the simulation.  The truth is that the well can not produce the target and it has to switch to BHP control before it goes to negative.

GitPaean:
The important part of this PR is in PR #1832 now. Closing.

PR #1800, opened 2019-04-17
fix the regula falsi solver for capillary pressure inversion

andlaus:
it is generally not a good idea to compare floating point values exactly. In this instance, the equil initializer tripped for Norne for a branch which I'm currently working on...


andlaus:
jenkins build this please

andlaus:
jenkins build this please

atgeirr:
There seems to be a lot of duplication here of the algorithm from opm/common/utility/numeric/RootFinders.hpp, why was that duplicated instead of used? That code already contains the feature included here (without the tolerance hardcoding).

andlaus:
turns out that the root cause of the error I've encountered was something else. closing.

Issue #1607, opened 2018-10-29
Failed in getting converged for the potential calculation for the well


subhis:
Hi. I'm trying to run a case on flow, but have problems with gas lift. With the previous OPM version (2018.04), I was having FLO_TYPE error when specifying the ALQ in WCONPROD section  (#1598), which was solved with latest version of the OPM.  But with 2018.10 version, I get the following error:
```
Report step  0/304 at day 0/15396, date = 06-Nov-1997

Time step 0, stepsize 1 days.

Error: [/build/opm-simulators-J49W6c/opm-simulators-2018.10-rfinal/opm/autodiff/StandardWell_impl.hpp:1887] Failed in getting converged for the potential calculation for well PROD_2
  ```
The error can be reproduced by the attached deck below. Would appreciate the help. 
[Norne_ALQ.zip](https://github.com/OPM/opm-simulators/files/2525916/Norne_ALQ.zip)

GitPaean:
At the moment, the well potential calculation based on THP limit is not supported in a good way. For some cases, you might get a converged result, but for many cases, you will not, mostly depending on the properties of the VFP tables. The reason is that we did not have good understanding how THP/VFP works and the current version was an implementation based on the VFP curves for some specific cases. 

The new development of THP/VFP related, including the well potential calculation based on THP limit, will be updated in the near future (before New Year most likely. )

subhis:
Alright, thanks. Looking forward for the new update.

PR #1802, opened 2019-04-23
Add well potential output to MSW

totto82:
The common part is moved to the well interface class

### DIFF
--- a/opm/autodiff/BlackoilModelParametersEbos.hpp
+++ b/opm/autodiff/BlackoilModelParametersEbos.hpp
@@ -72,7 +72,7 @@ SET_BOOL_PROP(FlowModelParameters, UpdateEquationsScaling, false);
 SET_BOOL_PROP(FlowModelParameters, UseUpdateStabilization, true);
 SET_BOOL_PROP(FlowModelParameters, MatrixAddWellContributions, false);
 SET_SCALAR_PROP(FlowModelParameters, TolerancePressureMsWells, 0.01 *1e5);
-SET_SCALAR_PROP(FlowModelParameters, MaxPressureChangeMsWells, 2.0 *1e5);
+SET_SCALAR_PROP(FlowModelParameters, MaxPressureChangeMsWells, 1e6);
 SET_BOOL_PROP(FlowModelParameters, UseInnerIterationsMsWells, true);
 SET_INT_PROP(FlowModelParameters, MaxInnerIterMsWells, 100);
 
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -306,7 +306,7 @@ namespace Opm {
             wellTesting(reportStepIdx, simulationTime, local_deferredLogger);
 
             // create the well container
-            well_container_ = createWellContainer(reportStepIdx, local_deferredLogger);
+            well_container_ = createWellContainer(reportStepIdx, wells(), /*allow_closing_opening_wells=*/true, local_deferredLogger);
 
             // do the initialization for all the wells
             // TODO: to see whether we can postpone of the intialization of the well containers to
@@ -414,7 +414,6 @@ namespace Opm {
         updateWellTestState(simulationTime, wellTestState_);
 
         // calculate the well potentials for output
-        // TODO: when necessary
         try {
             std::vector<double> well_potentials;
             computeWellPotentials(well_potentials, local_deferredLogger);
@@ -513,7 +512,7 @@ namespace Opm {
     template<typename TypeTag>
     std::vector<typename BlackoilWellModel<TypeTag>::WellInterfacePtr >
     BlackoilWellModel<TypeTag>::
-    createWellContainer(const int time_step, Opm::DeferredLogger& deferred_logger)
+    createWellContainer(const int time_step, const Wells* wells, const bool allow_closing_opening_wells, Opm::DeferredLogger& deferred_logger)
     {
         std::vector<WellInterfacePtr> well_container;
 
@@ -525,7 +524,7 @@ namespace Opm {
             // With the following way, it will have the same order with wells struct
             // Hopefully, it can generate the same residual history with master branch
             for (int w = 0; w < nw; ++w) {
-                const std::string well_name = std::string(wells()->name[w]);
+                const std::string well_name = std::string(wells->name[w]);
 
                 // finding the location of the well in wells_ecl
                 const int nw_wells_ecl = wells_ecl_.size();
@@ -543,58 +542,61 @@ namespace Opm {
 
                 const Well2& well_ecl = wells_ecl_[index_well];
 
-                // A new WCON keywords can re-open a well that was closed/shut due to Physical limit
-                if ( wellTestState_.hasWell(well_name, WellTestConfig::Reason::PHYSICAL ) ) {
-                    // TODO: more checking here, to make sure this standard more specific and complete
-                    // maybe there is some WCON keywords will not open the well
-                    if (well_state_.effectiveEventsOccurred(w)) {
-                        if (wellTestState_.lastTestTime(well_name) == ebosSimulator_.time()) {
-                            // The well was shut this timestep, we are most likely retrying
-                            // a timestep without the well in question, after it caused
-                            // repeated timestep cuts. It should therefore not be opened,
-                            // even if it was new or received new targets this report step.
-                            well_state_.setEffectiveEventsOccurred(w, false);
-                        } else {
-                            wellTestState_.openWell(well_name);
+                if (allow_closing_opening_wells) {
+                    // A new WCON keywords can re-open a well that was closed/shut due to Physical limit
+                    if ( wellTestState_.hasWell(well_name, WellTestConfig::Reason::PHYSICAL ) ) {
+                        // TODO: more checking here, to make sure this standard more specific and complete
+                        // maybe there is some WCON keywords will not open the well
+                        if (well_state_.effectiveEventsOccurred(w)) {
+                            if (wellTestState_.lastTestTime(well_name) == ebosSimulator_.time()) {
+                                // The well was shut this timestep, we are most likely retrying
+                                // a timestep without the well in question, after it caused
+                                // repeated timestep cuts. It should therefore not be opened,
+                                // even if it was new or received new targets this report step.
+                                well_state_.setEffectiveEventsOccurred(w, false);
+                            } else {
+                                wellTestState_.openWell(well_name);
+                            }
                         }
                     }
-                }
 
-                // TODO: should we do this for all kinds of closing reasons?
-                // something like wellTestState_.hasWell(well_name)?
-                if ( wellTestState_.hasWell(well_name, WellTestConfig::Reason::ECONOMIC) ||
-                     wellTestState_.hasWell(well_name, WellTestConfig::Reason::PHYSICAL) ) {
-                    if( well_ecl.getAutomaticShutIn() ) {
-                        // shut wells are not added to the well container
-                        // TODO: make a function from well_state side to handle the following
-                        well_state_.thp()[w] = 0.;
-                        well_state_.bhp()[w] = 0.;
-                        const int np = numPhases();
-                        for (int p = 0; p < np; ++p) {
-                            well_state_.wellRates()[np * w + p] = 0.;
+
+                    // TODO: should we do this for all kinds of closing reasons?
+                    // something like wellTestState_.hasWell(well_name)?
+                    if ( wellTestState_.hasWell(well_name, WellTestConfig::Reason::ECONOMIC) ||
+                         wellTestState_.hasWell(well_name, WellTestConfig::Reason::PHYSICAL) ) {
+                        if( well_ecl.getAutomaticShutIn() ) {
+                            // shut wells are not added to the well container
+                            // TODO: make a function from well_state side to handle the following
+                            well_state_.thp()[w] = 0.;
+                            well_state_.bhp()[w] = 0.;
+                            const int np = numPhases();
+                            for (int p = 0; p < np; ++p) {
+                                well_state_.wellRates()[np * w + p] = 0.;
+                            }
+                            continue;
+                        } else {
+                            // close wells are added to the container but marked as closed
+                            struct WellControls* well_controls = wells->ctrls[w];
+                            well_controls_stop_well(well_controls);
                         }
-                        continue;
-                    } else {
-                        // close wells are added to the container but marked as closed
-                        struct WellControls* well_controls = wells()->ctrls[w];
-                        well_controls_stop_well(well_controls);
                     }
                 }
 
                 // Use the pvtRegionIdx from the top cell
-                const int well_cell_top = wells()->well_cells[wells()->well_connpos[w]];
+                const int well_cell_top = wells->well_cells[wells->well_connpos[w]];
                 const int pvtreg = pvt_region_idx_[well_cell_top];
 
                 if ( !well_ecl.isMultiSegment() || !param_.use_multisegment_well_) {
                     if ( GET_PROP_VALUE(TypeTag, EnablePolymerMW) && well_ecl.isInjector() ) {
-                        well_container.emplace_back(new StandardWellV<TypeTag>(well_ecl, time_step, wells(),
+                        well_container.emplace_back(new StandardWellV<TypeTag>(well_ecl, time_step, wells,
                                                     param_, *rateConverter_, pvtreg, numComponents() ) );
                     } else {
-                        well_container.emplace_back(new StandardWell<TypeTag>(well_ecl, time_step, wells(),
+                        well_container.emplace_back(new StandardWell<TypeTag>(well_ecl, time_step, wells,
                                                     param_, *rateConverter_, pvtreg, numComponents() ) );
                     }
                 } else {
-                    well_container.emplace_back(new MultisegmentWell<TypeTag>(well_ecl, time_step, wells(),
+                    well_container.emplace_back(new MultisegmentWell<TypeTag>(well_ecl, time_step, wells,
                                                 param_, *rateConverter_, pvtreg, numComponents() ) );
                 }
             }
@@ -679,17 +681,17 @@ namespace Opm {
 
         int exception_thrown = 0;
         try {
-            if (iterationIdx == 0) {
-                calculateExplicitQuantities(local_deferredLogger);
-                prepareTimeStep(local_deferredLogger);
-            }
-
-            updateWellControls(local_deferredLogger);
-            // Set the well primary variables based on the value of well solutions
-            initPrimaryVariablesEvaluation();
-
             std::vector< Scalar > B_avg(numComponents(), Scalar() );
             computeAverageFormationFactor(B_avg);
+
+            if (iterationIdx == 0) {
+                calculateExplicitQuantities(local_deferredLogger);
+                prepareTimeStep(B_avg, local_deferredLogger);
+            }
+
+            updateWellControls(B_avg, local_deferredLogger);
+            // Set the well primary variables based on the value of well solutions
+            initPrimaryVariablesEvaluation();
 
             if (param_.solve_welleq_initially_ && iterationIdx == 0) {
                 // solve the well equations as a pre-processing step
@@ -699,7 +701,7 @@ namespace Opm {
                 if (initial_step_) {
                     // update the explicit quantities to get the initial fluid distribution in the well correct.
                     calculateExplicitQuantities(local_deferredLogger);
-                    prepareTimeStep(local_deferredLogger);
+                    prepareTimeStep(B_avg, local_deferredLogger);
                     last_report_ = solveWellEq(B_avg, dt, local_deferredLogger);
                     initial_step_ = false;
                 }
@@ -708,7 +710,7 @@ namespace Opm {
                 // reservoir state, will tihs be a better place to inialize the explict information?
             }
 
-	        assembleWellEq(B_avg, dt, local_deferredLogger);
+            assembleWellEq(B_avg, dt, local_deferredLogger);
 
         } catch (std::exception& e) {
             exception_thrown = 1;
@@ -921,7 +923,7 @@ namespace Opm {
                 // are active wells anywhere in the global domain.
                 if( wellsActive() )
                 {
-                    updateWellControls(deferred_logger);
+                    updateWellControls(B_avg, deferred_logger);
                     initPrimaryVariablesEvaluation();
                 }
             } catch (std::exception& e) {
@@ -1025,7 +1027,8 @@ namespace Opm {
     template<typename TypeTag>
     void
     BlackoilWellModel<TypeTag>::
-    updateWellControls(Opm::DeferredLogger& deferred_logger)
+    updateWellControls(const std::vector<Scalar>& B_avg,
+                       Opm::DeferredLogger& deferred_logger)
     {
         // Even if there are no wells active locally, we cannot
         // return as the DeferredLogger uses global communication.
@@ -1033,10 +1036,10 @@ namespace Opm {
         if( !wellsActive() ) return ;
 
         for (const auto& well : well_container_) {
-            well->updateWellControl(ebosSimulator_, well_state_, deferred_logger);
+            well->updateWellControl(ebosSimulator_, B_avg, well_state_, deferred_logger);
         }
 
-        updateGroupControls(deferred_logger);
+        updateGroupControls(B_avg, deferred_logger);
 
     }
 
@@ -1066,17 +1069,74 @@ namespace Opm {
     BlackoilWellModel<TypeTag>::
     computeWellPotentials(std::vector<double>& well_potentials, Opm::DeferredLogger& deferred_logger)
     {
-        Opm::DeferredLogger local_deferredLogger;
         // number of wells and phases
         const int nw = numWells();
         const int np = numPhases();
         well_potentials.resize(nw * np, 0.0);
 
+
+        const int reportStepIdx = ebosSimulator_.episodeIndex();
+        const double invalid_alq = -1e100;
+        const double invalid_vfp = -2147483647;
+        auto well_state_copy = well_state_;
+        const Wells* local_wells = clone_wells(wells());
+        std::vector<WellInterfacePtr> well_container_copy = createWellContainer(reportStepIdx, local_wells, /*allow_closing_opening_wells=*/ false, deferred_logger);
+
+        // average B factors are required for the convergence checking of well equations
+        // Note: this must be done on all processes, even those with
+        // no wells needing testing, otherwise we will have locking.
+        std::vector< Scalar > B_avg(numComponents(), Scalar() );
+        computeAverageFormationFactor(B_avg);
+
         const Opm::SummaryConfig& summaryConfig = ebosSimulator_.vanguard().summaryConfig();
         int exception_thrown = 0;
         try {
-            for (const auto& well : well_container_) {
+            for (const auto& well : well_container_copy) {
                 // Only compute the well potential when asked for
+                well->init(&phase_usage_, depth_, gravity_, number_of_cells_);
+
+                WellControls* wc = well->wellControls();
+                well_controls_clear(wc);
+                well_controls_assert_number_of_phases( wc , np);
+                if (well->wellType() == INJECTOR) {
+                    const auto& injectionProperties = well->wellEcl()->getInjectionProperties();
+
+                    if (injectionProperties.hasInjectionControl(WellInjector::THP)) {
+                        const double thp_limit  = injectionProperties.THPLimit;
+                        const int    vfp_number = injectionProperties.VFPTableNumber;
+                        well_controls_add_new(THP, thp_limit, invalid_alq, vfp_number, NULL, wc);
+                    }
+
+                    // we always have a bhp limit
+                    const double bhp_limit = injectionProperties.BHPLimit;
+                    well_controls_add_new(BHP, bhp_limit, invalid_alq, invalid_vfp, NULL, wc);
+                } else {
+                    const auto& productionProperties = well->wellEcl()->getProductionProperties();
+                    if (productionProperties.hasProductionControl(WellProducer::THP)) {
+                        const double thp_limit  = productionProperties.THPLimit;
+                        const double alq_value  = productionProperties.ALQValue;
+                        const int    vfp_number = productionProperties.VFPTableNumber;
+                        well_controls_add_new(THP, thp_limit, alq_value, vfp_number, NULL, wc);
+                    }
+
+                    // we always have a bhp limit
+                    const double bhp_limit = productionProperties.BHPLimit;
+                    well_controls_add_new(BHP, bhp_limit, invalid_alq, invalid_vfp, NULL, wc);
+
+                    well->setVFPProperties(vfp_properties_.get());
+
+                }
+
+                if (has_polymer_)
+                {
+                    const Grid& grid = ebosSimulator_.vanguard().grid();
+                    if (PolymerModule::hasPlyshlog() || GET_PROP_VALUE(TypeTag, EnablePolymerMW) ) {
+                        well->computeRepRadiusPerfLength(grid, cartesian_to_compressed_, deferred_logger);
+                    }
+                }
+
+
+
                 const bool needed_for_output = ((summaryConfig.hasSummaryKey( "WWPI:" + well->name()) ||
                                                  summaryConfig.hasSummaryKey( "WOPI:" + well->name()) ||
                                                  summaryConfig.hasSummaryKey( "WGPI:" + well->name())) && well->wellType() == INJECTOR) ||
@@ -1086,7 +1146,7 @@ namespace Opm {
                 if (needed_for_output || wellCollection().requireWellPotentials())
                 {
                     std::vector<double> potentials;
-                    well->computeWellPotentials(ebosSimulator_, well_state_, potentials, deferred_logger);
+                    well->computeWellPotentials(ebosSimulator_, B_avg, well_state_copy, potentials, deferred_logger);
                     // putting the sucessfully calculated potentials to the well_potentials
                     for (int p = 0; p < np; ++p) {
                         well_potentials[well->indexOfWell() * np + p] = std::abs(potentials[p]);
@@ -1111,7 +1171,8 @@ namespace Opm {
     template<typename TypeTag>
     void
     BlackoilWellModel<TypeTag>::
-    prepareTimeStep(Opm::DeferredLogger& deferred_logger)
+    prepareTimeStep(const std::vector<Scalar>& B_avg,
+                    Opm::DeferredLogger& deferred_logger)
     {
 
         if ( wellCollection().havingVREPGroups() ) {
@@ -1132,7 +1193,7 @@ namespace Opm {
         int exception_thrown = 0;
         try {
             for (const auto& well : well_container_) {
-                well->checkWellOperability(ebosSimulator_, well_state_, deferred_logger);
+                well->checkWellOperability(ebosSimulator_, B_avg, well_state_, deferred_logger);
             }
             // since the controls are all updated, we should update well_state accordingly
             for (const auto& well : well_container_) {
@@ -1144,7 +1205,7 @@ namespace Opm {
                 if (!well->isOperable() ) continue;
 
                 if (well_state_.effectiveEventsOccurred(w) ) {
-                        well->updateWellStateWithTarget(ebosSimulator_, well_state_, deferred_logger);
+                        well->updateWellStateWithTarget(ebosSimulator_, B_avg, well_state_, deferred_logger);
                 }
 
                 // there is no new well control change input within a report step,
@@ -1372,7 +1433,8 @@ namespace Opm {
     template<typename TypeTag>
     void
     BlackoilWellModel<TypeTag>::
-    updateGroupControls(Opm::DeferredLogger& deferred_logger)
+    updateGroupControls(const std::vector<Scalar>& B_avg,
+                        Opm::DeferredLogger& deferred_logger)
     {
 
         if (wellCollection().groupControlActive()) {
@@ -1399,7 +1461,7 @@ namespace Opm {
 
             // TODO: we should only do the well is involved in the update group targets
             for (auto& well : well_container_) {
-                well->updateWellStateWithTarget(ebosSimulator_, well_state_, deferred_logger);
+                well->updateWellStateWithTarget(ebosSimulator_, B_avg, well_state_, deferred_logger);
                 well->updatePrimaryVariables(well_state_, deferred_logger);
             }
         }
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -681,17 +681,17 @@ namespace Opm {
 
         int exception_thrown = 0;
         try {
-            std::vector< Scalar > B_avg(numComponents(), Scalar() );
-            computeAverageFormationFactor(B_avg);
-
             if (iterationIdx == 0) {
                 calculateExplicitQuantities(local_deferredLogger);
-                prepareTimeStep(B_avg, local_deferredLogger);
+                prepareTimeStep(local_deferredLogger);
             }
 
-            updateWellControls(B_avg, local_deferredLogger);
+            updateWellControls(local_deferredLogger);
             // Set the well primary variables based on the value of well solutions
             initPrimaryVariablesEvaluation();
+
+            std::vector< Scalar > B_avg(numComponents(), Scalar() );
+            computeAverageFormationFactor(B_avg);
 
             if (param_.solve_welleq_initially_ && iterationIdx == 0) {
                 // solve the well equations as a pre-processing step
@@ -701,7 +701,7 @@ namespace Opm {
                 if (initial_step_) {
                     // update the explicit quantities to get the initial fluid distribution in the well correct.
                     calculateExplicitQuantities(local_deferredLogger);
-                    prepareTimeStep(B_avg, local_deferredLogger);
+                    prepareTimeStep(local_deferredLogger);
                     last_report_ = solveWellEq(B_avg, dt, local_deferredLogger);
                     initial_step_ = false;
                 }
@@ -923,7 +923,7 @@ namespace Opm {
                 // are active wells anywhere in the global domain.
                 if( wellsActive() )
                 {
-                    updateWellControls(B_avg, deferred_logger);
+                    updateWellControls(deferred_logger);
                     initPrimaryVariablesEvaluation();
                 }
             } catch (std::exception& e) {
@@ -1027,8 +1027,7 @@ namespace Opm {
     template<typename TypeTag>
     void
     BlackoilWellModel<TypeTag>::
-    updateWellControls(const std::vector<Scalar>& B_avg,
-                       Opm::DeferredLogger& deferred_logger)
+    updateWellControls(Opm::DeferredLogger& deferred_logger)
     {
         // Even if there are no wells active locally, we cannot
         // return as the DeferredLogger uses global communication.
@@ -1036,10 +1035,10 @@ namespace Opm {
         if( !wellsActive() ) return ;
 
         for (const auto& well : well_container_) {
-            well->updateWellControl(ebosSimulator_, B_avg, well_state_, deferred_logger);
+            well->updateWellControl(ebosSimulator_, well_state_, deferred_logger);
         }
 
-        updateGroupControls(B_avg, deferred_logger);
+        updateGroupControls(deferred_logger);
 
     }
 
@@ -1171,8 +1170,7 @@ namespace Opm {
     template<typename TypeTag>
     void
     BlackoilWellModel<TypeTag>::
-    prepareTimeStep(const std::vector<Scalar>& B_avg,
-                    Opm::DeferredLogger& deferred_logger)
+    prepareTimeStep(Opm::DeferredLogger& deferred_logger)
     {
 
         if ( wellCollection().havingVREPGroups() ) {
@@ -1193,7 +1191,7 @@ namespace Opm {
         int exception_thrown = 0;
         try {
             for (const auto& well : well_container_) {
-                well->checkWellOperability(ebosSimulator_, B_avg, well_state_, deferred_logger);
+                well->checkWellOperability(ebosSimulator_, well_state_, deferred_logger);
             }
             // since the controls are all updated, we should update well_state accordingly
             for (const auto& well : well_container_) {
@@ -1205,7 +1203,7 @@ namespace Opm {
                 if (!well->isOperable() ) continue;
 
                 if (well_state_.effectiveEventsOccurred(w) ) {
-                        well->updateWellStateWithTarget(ebosSimulator_, B_avg, well_state_, deferred_logger);
+                        well->updateWellStateWithTarget(ebosSimulator_, well_state_, deferred_logger);
                 }
 
                 // there is no new well control change input within a report step,
@@ -1433,8 +1431,7 @@ namespace Opm {
     template<typename TypeTag>
     void
     BlackoilWellModel<TypeTag>::
-    updateGroupControls(const std::vector<Scalar>& B_avg,
-                        Opm::DeferredLogger& deferred_logger)
+    updateGroupControls(Opm::DeferredLogger& deferred_logger)
     {
 
         if (wellCollection().groupControlActive()) {
@@ -1461,7 +1458,7 @@ namespace Opm {
 
             // TODO: we should only do the well is involved in the update group targets
             for (auto& well : well_container_) {
-                well->updateWellStateWithTarget(ebosSimulator_, B_avg, well_state_, deferred_logger);
+                well->updateWellStateWithTarget(ebosSimulator_, well_state_, deferred_logger);
                 well->updatePrimaryVariables(well_state_, deferred_logger);
             }
         }
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -260,8 +260,9 @@ namespace Opm
     void
     MultisegmentWell<TypeTag>::
     updateWellStateWithTarget(const Simulator& /* ebos_simulator */,
+                              const std::vector<Scalar>& B_avg,
                               WellState& well_state,
-                              Opm::DeferredLogger& /* deferred_logger */) const
+                              Opm::DeferredLogger& /* deferred_logger */)
     {
         // Updating well state bas on well control
         // Target values are used as initial conditions for BHP, THP, and SURFACE_RATE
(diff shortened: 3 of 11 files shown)


totto82:
jenkins build this please

totto82:
The failure is expected since non-trivial well potentials are outputted.

totto82:

totto82:
jenkins build this opm-tests=110 please

totto82:
Wrong results for the field model. Needs more investigation.

totto82:
I had to add some inner-iterations for the MSW model to account for change in the friction. The results are ok. But some convergence problems due to poor initial guess. Since I am not very familiar with the MSW code, I would vary much like some feedback before I finalize the code.  

I also added output for disgas and vapoil while working on the code. It should be trivial.

GitPaean:
> I had to add some inner-iterations for the MSW model to account for change in the friction. 

I think iterating until converged is not only for the friction. Different from Standard Well model, the density is implicit with MSW. As a result, iteration is always required. 

> I would vary much like some feedback before I finalize the code.
I will try to provide some comments soon.

totto82:
Now also the density in the standard wells are recomputed. 

@GitPaean Great that you will have a look. Maybe we could have a short hangout when you have look at the code. Also there is still some convergence issues when the well potential are calculated for MSW that results in nonphysical potential values. Any hint how to improve it?

GitPaean:
> Also there is still some convergence issues when the well potential are calculated for MSW that results in nonphysical potential values. Any hint how to improve it?

What is the BHP values used to calculate the well potentials? If the BHP value is outside of the range of the PVT tables, it is hard for MSW. It is possible there is some assumption when coming to calculation of well potentials for MSW.

totto82:
> What is the BHP values used to calculate the well potentials?

Its defaulted i.e. 1 bar. So yes it may very well be outside the pvt range.

opm/autodiff/WellInterface_impl.hpp
```diff
@@ -1171,8 +1327,9 @@ namespace Opm
     template<typename TypeTag>
     bool
     WellInterface<TypeTag>::
-    solveWellEqUntilConverged(Simulator& ebosSimulator,
+    solveWellEqUntilConverged(const Simulator& ebosSimulator,
                               const std::vector<double>& B_avg,
```
totto82: Why was the simulator non-const here and several other places in the code?

opm/autodiff/MultisegmentWell.hpp
```diff
@@ -115,13 +115,14 @@ namespace Opm
         virtual void assembleWellEq(const Simulator& ebosSimulator,
                                     const std::vector<Scalar>& B_avg,
                                     const double dt,
+                                    const double strictBHP,
                                     WellState& well_state,
```
totto82: Passing strictBHP like this is quite ugly. Any idea how to improve this?

opm/autodiff/WellInterface_impl.hpp
GitPaean: I think it is because the function `assembleWellEq` might change the Jacobian, which is part of `ebosSimulator`? I did not re-read the code, excuse me it is wrong.

opm/autodiff/MultisegmentWell.hpp
GitPaean: We will discuss the interface things later.

opm/autodiff/StandardWellV_impl.hpp
```diff
@@ -2293,9 +2293,9 @@ namespace Opm
     void
     StandardWellV<TypeTag>::
     computeWellRatesWithBhp(const Simulator& ebosSimulator,
-                            const EvalWell& bhp,
+                            const double& bhp,
                             std::vector<double>& well_flux,
```
totto82: Any reasons for using EvalWell here?

GitPaean: Probably because we can not convert a `Scalar` to `DynamiceEvaluation`?

GitPaean:
> Its defaulted i.e. 1 bar. So yes it may very well be outside the pvt range.

It is not recommended in practice, while we should always be able to handle it in theory.

opm/autodiff/WellInterface_impl.hpp
totto82: That is no-longer the case since the Jacobian part is a separate step. So then it should be changed to const.

GitPaean: okay. If it compiles with `const` there, then `const` should be there.

GitPaean:
The potential calculation under THP limit in the `StandardWell` model is problematic, it remains to be changed to be more advanced techniques, either based on intersection of IPR and VFP curves or a full newton solution. 

Same applies to MSW when coming to well potential calculation under THP limit. 

For the well potential calculation under BHP limit, it is relatively easy to handle. I need a little more time to think through it.

GitPaean:
> Part of the convergence issue was cased by the strict restriction on pressure change.
(the default max-pressure-change-ms-wells ). Is this restriction important for other models?

I believe the value we are using is pretty relaxed.

totto82:
For the well potential calculation the pressure can change from say 100bars to 1bar, then the current default limitation is problematic.

GitPaean:
> current default limitation

Do you mean `--max-pressure-change-ms-wells`? 
Yes, it can be problematic. These parameters needs to be optimized further. I will finish the formulation part soon, then I will try to optimize these parameters.  They can be a little out-dated because the current formulation (code) is much more improved in term of convergence behavior and these parameters were set relatively conservatively to get converged results. 

And also, if the BHP changes so drastically, maybe we should update the WellState so that the initial guess is better. 

We can have a short hangout after May 1st. With this PR, are you trying to complete the well potential calculation for both StandardWell and Multisegment Well, and for both THP and BHP limits, or there is some parts are of higher priority than other parts?

totto82:
> With this PR, are you trying to complete the well potential calculation for both StandardWell and Multisegment Well, and for both THP and BHP limits, or there is some parts are of higher priority than other parts?

I haven't looked at the THP part for the MSW model. The focus is on output.

GitPaean:
> I haven't looked at the THP part for the MSW model. The focus is on output.

okay. If I understand correctly, the key part is still to get the correct values.  Let us have a talk after May 1st.

opm/autodiff/MultisegmentWell_impl.hpp
```diff
@@ -738,6 +840,8 @@ namespace Opm
             average_density /= sum_kr;
 
             cell_perforation_pressure_diffs_[perf] = gravity_ * average_density * cell_perforation_depth_diffs_[perf];
+
+            Base::perf_densities_[perf] = average_density;
```
GitPaean: I do not think MSW should use `perf_densities_`. It should stay within `StandardWell`.  MSW has different way in handling the fluid densities within wellbore.

opm/autodiff/MultisegmentWell_impl.hpp
```diff
@@ -798,7 +902,7 @@ namespace Opm
                 const int sign = dwells[seg][SPres] > 0.? 1 : -1;
                 const double dx_limited = sign * std::min(std::abs(dwells[seg][SPres]), relaxation_factor * max_pressure_change);
                 // const double dx_limited = sign * std::min(std::abs(dwells[seg][SPres]) * relaxation_factor, max_pressure_change);
-                primary_variables_[seg][SPres] = old_primary_variables[seg][SPres] - dx_limited;
+                primary_variables_[seg][SPres] = std::max( old_primary_variables[seg][SPres] - dx_limited, 1e5);
```
GitPaean: For this one, when negative BHP happens, it should switch to BHP control. Not sure we would like that. Similar code exists due to historical reason.

opm/autodiff/MultisegmentWell_impl.hpp
```diff
@@ -638,7 +639,7 @@ namespace Opm
 
         // we need some intput the convergence calculations.
         // we don't need the exact values so we just use something ad-hoc
-        const std::vector<Scalar> B_avg = {800,1000,1};
+        const std::vector<Scalar> B_avg = {1000,800,1};
```
GitPaean: I think the values of `B` are totally off. They should be below 1?

GitPaean:
For non-reasonable BHP limit (1 atm), I believe it is okay to report that our solution is not converged and we did not get a solution for it.

GitPaean:
They are just my comments when reading the PR, while I still do not have very concrete suggestion for how we should do it in a nice way. Let us continue the discussion next week.

Probably we will try to have implementation for StandardWell and MSW in a separate way first, then we try to figure out whether we can unify them. 

You do  not need to do anything to this PR at this point.

opm/autodiff/MultisegmentWell_impl.hpp
totto82: I agree that they are computed differently, but both methods computes as average perforation density.

opm/autodiff/MultisegmentWell_impl.hpp
totto82: Ops. It was meant has a quick fix for testing. I will use the competed values.

GitPaean:
Referring to our discussion previously. We will address only the well potential calculated based on BHP first in this PR.  I will suggest to simply through a warning within function `computeWellPotentialWithTHP()` to say we are not handling it yet now.

GitPaean:
For the calculation of well potentials with BHP limit, at the moment, one possible suggestion is to create a temporary `WellControls` that only contain a BHP limit. Then we solve the well equations based on this BHP limit. After we obtain the well potentials, we recover the old `WellControls`.  All these should only happen in the function `computeWellPotentials`, we re-use other functions within this function without change other functions.  

The other way is to pass a bool flag around to state we are doing well potential calculation. With this way, we might need to change more functions. 

The first option might be easier since it might potentially change less functions and can be easier to have a more unified implementation.

GitPaean:
> The well controls are changed in a non-intuitive way (at least for my intuition) and we really don't want the potential calculation to mess up the real well solution. I plan to test a new solution where I solve the well potential on a complete separate well container set where the WellControls only contains BHP limits.

In my opinion, since you are creating `WellControls` that only contains BHP limits, you can just do it following my first suggestion. Do it within the function `computeWellPotentials`.  You need to recover the well controls, well state and anything changed during this solution system.  To me, this can be the easiest way. 

If you go to the `BlackoilWellModel` level, then you still need to deal with each individual well. The good thing about your approach is that you intervene less with the normal running.

totto82:
@GitPaean Now I have made a version that adjust the well controls instead and removed the hacking flags. Works well for the models I have tested.

totto82:
jenkins build this please

totto82:
jenkins build this please

GitPaean:
I will try to review this PR soon.  What is the status of the main motivation of the PR?

totto82:
The main motivation for this PR is to compute and output the correct well potentials for MSW. It also contains an improvement for the standard wells. The PR is ready from my side. I plan a rebase to clean the commit messages, thats all, but that I will do after addressing your review.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -107,6 +107,7 @@ namespace Opm
         well_efficiency_factor_ = 1.0;
 
         connectionRates_.resize(number_of_perforations_);
+        perf_densities_.resize(number_of_perforations_);
```
GitPaean: This should not be used in MSW at all. So it should not go to WellInterface.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -325,6 +331,161 @@ namespace Opm
     }
 
 
+    template<typename TypeTag>
+    template<class ValueType>
+    ValueType
+    WellInterface<TypeTag>::
+    calculateBhpFromThp(const std::vector<ValueType>& rates,
```
GitPaean: For MSW, the top segment might not have a perforation. Unless we have some strong evidence, we should not assume to do THP related to be the same way with StandardWell.  It should be similar though. 

As we discussed, we should only address the calculation related to `BHP` limit in this PR. If we handle THP related, it will be a much more extensive PR. 

It will be good to just through a warning when handle well potential related to THP limit, including in StandardWell. 

Referring to the discussion in https://github.com/OPM/opm-simulators/issues/1607 , we should address THP related in a different PR.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -325,6 +331,161 @@ namespace Opm
...
+    template<typename TypeTag>
+    std::vector<double>
+    WellInterface<TypeTag>::
+    computeWellPotentialWithTHP(const Simulator& ebosSimulator,
+                                const double initial_bhp, // bhp from BHP constraints
+                                const std::vector<double>& initial_potential,
+                                Opm::DeferredLogger& deferred_logger)
+    {
+        // TODO: pay attention to the situation that finally the potential is calculated based on the bhp control
+        // TODO: should we consider the bhp constraints during the iterative process?
```
GitPaean: Referring to the comments above. This function should thrown a warning and return a vector of zero values.  Including the existing same function in StandardWell. The implementation is outdated.

opm/simulators/wells/WellInterface_impl.hpp
totto82: I have added a warning if THP limit if the user tries to compute well potentials for cases with THP and MSW. (line 593 in MultisegmentWell_impl.hpp) 

Even though some details may be different, I would assume that most of the code will be common for MSW and standard wells regarding VFP. But since you are planning to do the VPF stuff I suggest that things gets merged as is, and you can refactor and change the code according to your need later. It is not easy to guess for me to guess what may or may not be a good code structure.

opm/simulators/wells/WellInterface_impl.hpp
totto82: I don't see why. The perf_densities are well defined also for MSW and computed in computePerfCellPressDiffs.

GitPaean: But it is only temporary in that function, it is used for calculation of something else, then its life is finished there.  So it does not need to be a member at all. 

And also, they have different definition and different way of calculation. 

Unless there are some other evidence to extend its lifespan, we should not do that.

totto82: Currently, it is used in calculateBhpFromThp. If you are so strongly against this, I can revert these changes and move back the cacluateBhpFromThp code. I just tried to remove some code duplication between the standard and MSW code. Any other things you want me to change in the PR?

GitPaean: Please do that. Because we should only introduce working code, remove code not working.  It is not good to introduce code not working. Depending on the plan, it might take some time before we get a chance to re-factor or correct code that not working. 

Please also remove the code (not working) in the function `computeWellPotentialWithTHP()`, instead, you can just throw a warning and return some trivial values, depending on how you want to organize the `computeWellPotentials`.  You can decide whether you want to put `computeWellPotentialWithTHP` to `StandardWell` or keep it in `WellInterface`. 

Those are the big parts with a quick review.  After that, I will provide more detailed review comments. It will be easier when we reduce the code change first.

totto82:
@GitPaean I have now removed the THP stuff from the common code as you suggested.

GitPaean:
> @GitPaean I have now removed the THP stuff from the common code as you suggested.

Thanks. You will hear from me tomorrow.

totto82:
jenkins build this please

totto82:
I notice that  XGRP is changed by this PR. What is XGRP?

bska:
> I notice that XGRP is changed by this PR. What is XGRP?

Double precision quantities per group. Typically actual and potential flow rates and cumulative totals.  Full list of supported items in [`output/eclipse/VectorItems/group.hpp`](https://github.com/OPM/opm-common/blob/4010399f08a61b5194e7ca33cb6a0f03b6bf9217/opm/output/eclipse/VectorItems/group.hpp). Since this PR modifies flow **potentials**, I'd say that's the most likely change although I'd have to look at the specific vectors to be sure.

Structure of `XGRP` is NXGRPZ (180) items for each of `NGMAXZ + 1` groups (`NGMAXZ` is item 3 of WELLDIMS, +1 for "FIELD").  FIELD results stored as last sub-vector (vector ID `NGMAXZ` in zero-based indexing).

totto82:
@bska Thanks.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -689,27 +698,24 @@ namespace Opm {
             // Set the well primary variables based on the value of well solutions
             initPrimaryVariablesEvaluation();
 
-            std::vector< Scalar > B_avg(numComponents(), Scalar() );
-            computeAverageFormationFactor(B_avg);
-
```
GitPaean: This potentially changes result, because it is not updated anymore for each iteration.  Do we have any document about what the correct way should be?

totto82: The B_avg does not change much during simulation. With this change it is only updated every report step i.e. when the well is created, I will be very surprised if this has any noticeable effect on the results.

GitPaean: Sure. I understand it, while it probably depends on the individual case setup. 

There are some other places also changes the iteration, we will evaluate it. 

Most ideally, if this PR only output the well potentials for the cases that request the output without changing any running results, it will be much easier to review and test. 

If the iteration changes, it just adds some testing burden for the cases we have concerns with.

totto82:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
totto82: > There are some other places also changes the iteration, we will evaluate it.

If you send me an email with the name of the models that changes results due to this I could take a look myself.

opm/autodiff/BlackoilModelParametersEbos.hpp
```diff
@@ -72,7 +72,7 @@ SET_BOOL_PROP(FlowModelParameters, UpdateEquationsScaling, false);
 SET_BOOL_PROP(FlowModelParameters, UseUpdateStabilization, true);
 SET_BOOL_PROP(FlowModelParameters, MatrixAddWellContributions, false);
 SET_SCALAR_PROP(FlowModelParameters, TolerancePressureMsWells, 0.01 *1e5);
-SET_SCALAR_PROP(FlowModelParameters, MaxPressureChangeMsWells, 2.0 *1e5);
+SET_SCALAR_PROP(FlowModelParameters, MaxPressureChangeMsWells, 1e6);
 SET_BOOL_PROP(FlowModelParameters, UseInnerIterationsMsWells, true);
```
totto82: @GitPaean This could affect performance, even thought I havn't notice any regression.

GitPaean: I will test it separately. Hopefully it only does good.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -880,25 +886,31 @@ namespace Opm {
     template<typename TypeTag>
     SimulatorReport
     BlackoilWellModel<TypeTag>::
-    solveWellEq(const std::vector<Scalar>& B_avg, const double dt, Opm::DeferredLogger& deferred_logger)
+    solveWellEq(const double dt, Opm::DeferredLogger& deferred_logger)
     {
         WellState well_state0 = well_state_;
 
+        if( wellsActive() )
+        {
```
GitPaean: Did you see some values are not initialized under some circumstances?

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -415,14 +421,17 @@ namespace Opm {
...
 
         // calculate the well potentials for output
-        // TODO: when necessary
         try {
             std::vector<double> well_potentials;
             computeWellPotentials(well_potentials, local_deferredLogger);
         } catch ( std::runtime_error& e ) {
             const std::string msg = "A zero well potential is returned for output purposes. ";
             local_deferredLogger.warning("WELL_POTENTIAL_CALCULATION_FAILED", msg);
+        } catch (std::logic_error& e ) {
```
GitPaean: Where can this exception be triggered?

totto82: If the inverting of D fails it throws a numerical_error that is catched by the Deferred_logger that re-throw a logical_error in logAndCheckForExceptionsAndThrow. 

I don't know the details of the deferred_logger so I just added a fix for my problems. 
Maybe @atgeirr can explain why the exception is handled this way?

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1067,17 +1079,75 @@ namespace Opm {
...
+        auto well_state_copy = well_state_;
+        const Wells* local_wells = clone_wells(wells());
+        std::vector<WellInterfacePtr> well_container_copy = createWellContainer(reportStepIdx, local_wells, deferred_logger);
+
+
+        // average B factors are required for the convergence checking of well equations
+        // Note: this must be done on all processes, even those with
+        // no wells needing testing, otherwise we will have locking.
+        std::vector< Scalar > B_avg(numComponents(), Scalar() );
+        computeAverageFormationFactor(B_avg);
```
GitPaean: Since `B_avg` is re-calculated here, it might not be necessary to make it a member of the `WellInterface`.  If it is seen as  a simplification of the interfaces of the original master branch, maybe it is better to do it in a separate PR so it is easier to review and evaluate?

opm/simulators/wells/BlackoilWellModel_impl.hpp
atgeirr: Thanks for noticing!

The logAndCheckForExceptionsAndThrow() function does two things:
  - Writes all deferred log messages to the output rank.
  - If any rank have thrown an exception (and caught it, set the flag and passed to this function) then we throw logic_error.

I realize that this represents a loss of information compared to before for the serial case (however for the parallel case all we lost was race conditions...). Since the log messages are written, they should hopefully contain enough information to understand what happened, even if the higher level code no longer sees the exception type.

In order to re-throw the same type of exception, we would need to communicate the exception type somehow, and we have not made a facility for this. We could of course do that, it would requires us to define error codes (like in the good old days) to be used for this purpose and to change the catch spots. Perhaps adding an error code to the exception classes used also. It would be a bit of work to ensure it was done consistently across the whole system.

GitPaean: It looks the `runtime_error` and `logic_error` are both related to the calculation of well potentials, I am wondering why this one is a `logic_error`? 

And since you are here, since when we terminate the simulation when we encounter `NumericalIssue`? Maybe it is just better to chop the time step to re-run to see whether the `NumericalIssue` is still there?

opm/simulators/wells/BlackoilWellModel_impl.hpp
totto82: > if it is seen as a simplification of the interfaces of the original master branch, maybe it is better to do it in a separate PR

Is that really necessary. The effect on the results seems to be below the Jenkins tolerance. Since test that don't involve output of well potentials or usage of well potential does not fail. And among the test that fails the reason is the well potential change. If you have another model you are worried about, please tell me. And I can verify that the B_avg change doesn't hurt.

totto82:
jenkins build this please

GitPaean:
Jenkins failed in the following test. 
```
Report step 21/30 at day 210/300, date = 30-Jul-2015

Time step 0, stepsize 10 days.
[opm-deb:16521] *** Process received signal ***
[opm-deb:16521] Signal: Segmentation fault (11)
[opm-deb:16521] Signal code: Address not mapped (1)
[opm-deb:16521] Failing at address: 0x56514b6893a8
[opm-deb:16521] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x110e0)[0x7f4a8be860e0]
[opm-deb:16521] [ 1] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(_ZN3Opm13WellInterfaceIN5Ewoms10Properties4TTag14EclFlowProblemEE17updateWellControlERKNS1_9SimulatorIS4_EERNS_30WellStateFullyImplicitBlackoilERNS_14DeferredLoggerE+0x253)[0x5639b8b1f553]
[opm-deb:16521] [ 2] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(_ZN3Opm13WellInterfaceIN5Ewoms10Properties4TTag14EclFlowProblemEE25solveWellEqUntilConvergedERKNS1_9SimulatorIS4_EERNS_30WellStateFullyImplicitBlackoilERNS_14DeferredLoggerE+0xb6)[0x5639b8b36806]
[opm-deb:16521] [ 3] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(_ZN3Opm12StandardWellIN5Ewoms10Properties4TTag14EclFlowProblemEE23computeWellRatesWithBhpERKNS1_9SimulatorIS4_EERKdbRSt6vectorIdSaIdEERNS_14DeferredLoggerE+0xf0)[0x5639b8c3d230]
[opm-deb:16521] [ 4] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(_ZN3Opm12StandardWellIN5Ewoms10Properties4TTag14EclFlowProblemEE21computeWellPotentialsERKNS1_9SimulatorIS4_EERKNS_30WellStateFullyImplicitBlackoilERSt6vectorIdSaIdEERNS_14DeferredLoggerE+0x235)[0x5639b8c3e5e5]
[opm-deb:16521] [ 5] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(_ZN3Opm17BlackoilWellModelIN5Ewoms10Properties4TTag14EclFlowProblemEE21computeWellPotentialsERSt6vectorIdSaIdEERNS_14DeferredLoggerE+0x40e)[0x5639b8b6db8e]
[opm-deb:16521] [ 6] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(_ZN3Opm17BlackoilWellModelIN5Ewoms10Properties4TTag14EclFlowProblemEE19prepareGroupControlERNS_14DeferredLoggerE+0x159)[0x5639b8b6e399]
[opm-deb:16521] [ 7] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(_ZN3Opm17BlackoilWellModelIN5Ewoms10Properties4TTag14EclFlowProblemEE15prepareTimeStepERNS_14DeferredLoggerE+0x75)[0x5639b8b70715]
[opm-deb:16521] [ 8] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(_ZN3Opm17BlackoilWellModelIN5Ewoms10Properties4TTag14EclFlowProblemEE8assembleEid+0x260)[0x5639b8b760f0]
[opm-deb:16521] [ 9] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(+0x89fe72)[0x5639b8c4de72]
[opm-deb:16521] [10] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(+0x8a145e)[0x5639b8c4f45e]
[opm-deb:16521] [11] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(+0x8a575a)[0x5639b8c5375a]
[opm-deb:16521] [12] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(_ZN3Opm12FlowMainEbosIN5Ewoms10Properties4TTag14EclFlowProblemEE12runSimulatorEv+0x18b)[0x5639b8c55c6b]
[opm-deb:16521] [13] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(_ZN3Opm12FlowMainEbosIN5Ewoms10Properties4TTag14EclFlowProblemEE7executeEiPPc+0x233)[0x5639b8c56083]
[opm-deb:16521] [14] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(_ZN3Opm20flowEbosBlackoilMainEiPPc+0x136)[0x5639b8a452e6]
[opm-deb:16521] [15] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(main+0x9f8)[0x5639b8951d08]
[opm-deb:16521] [16] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf1)[0x7f4a85cfa2e1]
[opm-deb:16521] [17] /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow(_start+0x2a)[0x5639b898ea2a]
[opm-deb:16521] *** End of error message ***
/var/lib/jenkins/workspace/opm-simulators-PR-builder/tests/run-regressionTest.sh: line 19: 16521 Segmentation fault      ${BINPATH}/${EXE_NAME} ${TEST_ARGS} --output-dir=${RESULT_PATH} 
```

totto82:
> Jenkins failed in the following test.

Not anymore. I needed to update the currentControll in the wellState in computWellRatesWithBhp.

opm/simulators/wells/BlackoilWellModel_impl.hpp
GitPaean: > And I can verify that the B_avg change doesn't hurt.

I understand what you mean. But on the other way, does it do any good or we have a good reason to do it?

It will result in a much smaller pull request and we do not need to evaluate the possible effects from it.

totto82:
For reference: The simulation results on SPE9_CP_GROUP (where Jenkins claims results have changed ) is closer Ecl results.

GitPaean:
> For reference: The simulation results on SPE9_CP_GROUP (where Jenkins claims results have changed ) is closer Ecl results.

Good to know. Can you share some plots?

opm/simulators/wells/BlackoilWellModel_impl.hpp
totto82: In my opinion it simplifies the interface. With the improved well potential calculations the B_avg values are needed deep down in the code. I really want to avoid passing around the B_avg values in a lot of files. With 3 wells models and both .hpp and impl_hpp files + the well interface it is significant work to change this. I think we should allow clean-up of the code while adding features. If not, I am afraid that we will never have time to improve the code.

opm/simulators/wells/BlackoilWellModel_impl.hpp
totto82: I did. But now the code seems to work without. I will remove the change.

totto82:
image

GitPaean:
> Random production plot from the case. For the last part of the simulation well potentials are used to distribute the group target among wells.

Thanks for sharing. 

It can be from better well potential calculation since the pressure head drop difference is updated with a solution now. 

It is also possible from the change of the way of `B_avg`, although not very likely. If it does, then it might be a good clue that it is possible we should handle `B_avg` differently.

totto82:
> It is also possible from the change of the way of B_avg, although not very likely. If it does, then it might be a good clue that it is possible we should handle B_avg differently.

Of course this has nothing to do with the B_avg computation and all to do with improved well potential calculations. My point is that results changes only for cases that it should change and it is improved.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -2217,6 +2246,11 @@ namespace Opm
                 well_flux[ebosCompIdxToFlowCompIdx(p)] += cq_s[p].value();
             }
         }
+
+
+        // rest bhp limit
```
GitPaean: rest -> reset

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -2189,12 +2189,41 @@ namespace Opm
...
                             std::vector<double>& well_flux,
-                            Opm::DeferredLogger& deferred_logger) const
+                            Opm::DeferredLogger& deferred_logger)
     {
+
+
         const int np = number_of_phases_;
         well_flux.resize(np, 0.0);
+        WellControls* wc = well_controls_;
+        const double orig_bhp = well_controls_iget_target(wc, BHP);
```
GitPaean: I do not understand the usage `well_controls_iget_target(wc, BHP);`. 

Please excuse me if I misread it. I think it is a wrong usage.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -566,24 +566,112 @@ namespace Opm
...
+        WellState copy = ebosSimulator.problem().wellModel().wellState();
+
+        // try to improve the initial condition.
+        // we adjust the initial seg pressure with half the difference
+        // between the real bhp and the potential bhp.
+        // This is ad-hoc, but seems to work reasonbly well.
+        const double adjustBhp = (bhp - primary_variables_[0][SPres]);
+        double adjustTotalRate = 0.0;
+        for (int p = 0; p < number_of_phases_; ++p) {
+            adjustTotalRate += copy.wellPotentials()[index_of_well_ * number_of_phases_ + p];
```
GitPaean: It is easy to understand how the BHP and segment pressure is adjusted. It is hard to understand how the rate is adjusted. Because the primary variable GTotal is a weighted sum of the well rates. When gas exist, this way of scaling is probably highly problematic. 

And what happens if the wellPotentials from the `copy` is not updated yet? then what happens? Please check whether the update of rates is necessary here. 

and also, since in the function, it is using `wellPoentials()` in a explicit way, maybe it is good to change the function name to be `computeWellPotentialsWithBhpLimit`, since it is not a general way of calculating well rates anymore.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -566,24 +566,112 @@ namespace Opm
...
+        initPrimaryVariablesEvaluation();
+
+        if (iterate) {
+            const double dt = ebosSimulator.timeStepSize();
+            // iterate to get a solution that satisfies the bhp potential.
+            iterateWellEquations(ebosSimulator, dt, copy, deferred_logger);
+        }
+
+        // compute the potential and store in the flux vector.
+        const int np = number_of_phases_;
```
GitPaean: I believe the following code will be equivalent to get the well rates based on `getSegmentRate(0, comp_idx)` . Basically the well rates of the top segment should be same with summing through all the perforation rates. 

Please report if they turned out to be different.

GitPaean:
Please revert the change related to `B_avg`. Then as a result, we will limit the change only within a few functions, and they will be only related to the calculation of well potentials. Then all the failures will be related to the output and use of well potentials. 

I am running some simulation to evaluate the parameter change related to MSW.

totto82:
> Then all the failures will be related to the output and use of well potentials.

All failures are related to the well potential stuff!!!!!

opm/simulators/wells/StandardWell_impl.hpp
totto82: Based on your suggestion I changed the code such that the control is changed rather than explicitly setting the bhp in the equation. Here is just store the original bhp control target value so that I can reset the control later.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: > Please check whether the update of rates is necessary here.

I think it can be removed. It improved the convergence but only slightly.  I can test without it on some of the big models.

GitPaean:
jenkins build this please

GitPaean:
> All failures are related to the well potential stuff!!!!!

The cases in Jenkins are simple. Many might not care about `B_avg`, but some might care. Like the following one

https://ci.opm-project.org/job/opm-simulators-PR-builder/1504/testReport/junit/(root)/mpi/compareECLFiles_flow_3D_WECON/

For bigger cases, like the cases that we care the most, mostly likely `B_avg` will make a difference in term of iteration behavoir, then we have to evaluate.  If the change of `B_avg` is not intentional (having a good reason), we better not deal with it. 

A pull request will be good if it can fulfill its purpose with making minimal change. 

I am here to help. I am not here to make problems.

opm/simulators/wells/StandardWell_impl.hpp
GitPaean: I think `BHP` is a `enum` control mode. I do not think the usage of `well_controls_iget_target(wc, BHP)` is correct.

GitPaean:
By changing the `MaxPressureChangeMsWells` to be `1e+6`, the two cases I ran, one is faster, one is slower.  More cases will be tested.

GitPaean:
> By changing the MaxPressureChangeMsWells to be 1e+6, the two cases I ran, one is faster, one is slower. More cases will be tested.

I conducted running of another 4 examples, 3 of them are slightly slower, while one of them is faster, however, the one faster is faster in a more significant way. 

I think it is okay to change this parameter.

totto82:
> I am here to help. I am not here to make problems.

I know. Thanks!!! I will do what you suggest, but keep in mind that all changes you suggest makes me spend more time on this.

opm/simulators/wells/StandardWell_impl.hpp
totto82: You are right. I havn't been working the the well controls in awhile. I guess I will have to iterate over all controls and pick the one with BHP.

GitPaean:
> but keep in mind that all changes you suggest makes me spend more time on this.

Thanks. I have my concerns too. One is that I do think it cost less time than running through the important cases to evaluate possible impacts.  And also, if it causes later problems, for sure, it will be a bigger effort to fix it. It is good to limit the scope of PRs. 

I can help some patches if needed. Just let me know.

totto82:
I have reverted the B_avg change. I am still looking into the WECON failure.

alfbr:
With this PR I am still getting messages like this:

Warning: Well potential calculation is not supported for multisegment wells 
A well potential of zero is returned for output purposes. 
If you need well potential to set the guide rate for group controled wells 
you will have to change the A-1 well to a standard well

Should not the message be removed when this PR is merged?

bska:
> With this PR I am still getting messages like this:
> 
>> Warning: Well potential calculation is not supported for multisegment wells

Does the message say **exactly**
```
... not supported for multisegment ...
```
or is slightly more refined as
```
... not supported for thp controlled multisegment ...
```
with the difference between the two being "thp controlled"?

If it's the first, then that's likely to be a bug.  If it's the second, then you've "just" bumped into a different limitation (cf. https://github.com/OPM/opm-simulators/blob/1e4ab75716dbcc36637496b3d8651b9fd1dcd383/opm/simulators/wells/MultisegmentWell_impl.hpp#L570-L573)

opm/simulators/wells/StandardWellV_impl.hpp
```diff
@@ -2265,12 +2273,42 @@ namespace Opm
...
+        WellControls* wc = well_controls_;
+        const int bhp_index = Base::getControlIndex(BHP);
+        const double orig_bhp = well_controls_iget_target(wc, bhp_index);
+        const auto orig_current = well_controls_get_current(wc);
+
+        well_controls_iset_target(wc, bhp_index, bhp);
+        well_controls_set_current(wc, bhp_index);
+
+        // iterate to get a more accurate well density
+        if (iterate) {
```
GitPaean: I understand the problem better now.

I do not have access to reference result. It is possible for well potential calculation, it needs to solve the well equation first, to update the perforation pressure drops to be proper first. 

And also, at the same time, for some other purposes, it might not need it, because the perforation pressure drops are treated in an explicit manner.  There are specified perforation pressure drops there already. 

Maybe we should split them `computeWellPotentialsWithBhp` and you call `computeWellRatesWithBhp` within the previous function. 

Will this make the interface change of `B_avg` less, instead of using a boolean input arguments to put them together?

totto82: But now I have done the work. I really don't want to spend another half day changing it all back again.

totto82:
> With this PR I am still getting messages like this:

The messages was accidentally reintroduced during the process of addressing code review. It is long gone now.

alfbr:
@totto82 are you sure, this is from git master built this morning, tagged this morning. I double checked that I had this PR in. I can triple check of course.

> Does the message say **exactly**
> 
> ```
> ... not supported for multisegment ...
> ```

@bska, yes the message is cut'n'paste.

totto82:
@alfbr It was removed 3h ago.

opm/simulators/wells/StandardWellV_impl.hpp
GitPaean: Sure. It is fine. I put it on my list. I will handle it.

alfbr:
Thanks, will give it another go.

totto82:
> Thanks, will give it another go.

I suggest that you wait. After the latest change of code there is some issues with the WECON computations. I.e. a potential but in the PR in its current form.

alfbr:
Sure, just give me a heads-up when it is ready.

totto82:
jenkins build this please

totto82:
Anything more? This Pr needs an update of data. I have verified that all failures are due to improved well potential calculations.

bska:
> This Pr needs an update of data.

Please hold all data updates until we've merged #1848 and OPM/opm-tests#119.

GitPaean:
> Anything more?

Thanks for the patience. I will review it today and tomorrow. It is possible I will crate PR directly towards your branch. 

It is not easy to tell the best way to do the things related as you can see. I will need some functionality in the immediate future. So let us spend a little more efforts before it goes in. 

You do not need to do anything more without my further request.

GitPaean:
> Please hold all data updates until we've merged #1848 and OPM/opm-tests#119.

No problem. We will wait to do it after merging #1848.

GitPaean:
It is good if you can do a rebase. So it might be easier to create PR against your branch.

opm/simulators/wells/StandardWellV_impl.hpp
```diff
@@ -2265,12 +2273,42 @@ namespace Opm
...
+                            const double& bhp,
+                            const bool iterate,
                             std::vector<double>& well_flux,
-                            Opm::DeferredLogger& deferred_logger) const
+                            Opm::DeferredLogger& deferred_logger)
     {
+
         const int np = number_of_phases_;
         well_flux.resize(np, 0.0);
+        WellControls* wc = well_controls_;
```
GitPaean: Without `iterate` to be `true`,  the setting of the `bhp` limit should not happen here.  An immediate change will be put these into the `if (iterate) {`, including the following code
```
        // reset bhp limit
        well_controls_iset_target(wc, bhp_index, orig_bhp);
        well_controls_set_current(wc, orig_current);
    }	

```
Please do not do anything now to see whether we can find a better solution.

alfbr:
I believe a change on this PR during the last 24 hours has hurt numerical convergence/robustness. I have a couple of cases over here that fails to run now, but successfully ran through two days ago.

totto82:
@alfbr Maybe you could contact me directly with some more details.

GitPaean:
> I believe a change on this PR during the last 24 hours has hurt numerical convergence/robustness. I have a couple of cases over here that fails to run now, but successfully ran through two days ago.

Thanks for reporting. That is the most worrying and difficult part. Please provide more details through emails if necessary.

opm/autodiff/BlackoilModelParametersEbos.hpp
```diff
@@ -72,7 +72,7 @@ SET_BOOL_PROP(FlowModelParameters, UpdateEquationsScaling, false);
 SET_BOOL_PROP(FlowModelParameters, UseUpdateStabilization, true);
 SET_BOOL_PROP(FlowModelParameters, MatrixAddWellContributions, false);
 SET_SCALAR_PROP(FlowModelParameters, TolerancePressureMsWells, 0.01 *1e5);
-SET_SCALAR_PROP(FlowModelParameters, MaxPressureChangeMsWells, 2.0 *1e5);
```
GitPaean: @alfbr , @totto82 , it is mostly likely this change makes some cases fail.

GitPaean: By changing it to `2.e5`, the running is recovered, while some slightly different iteration process was shown. One more testing is ongoing and it will be reported back.

GitPaean: Did not figure out why the iteration behavior changed compared with master branch. But with resetting it back to `2.e5`, the running of the obtained problematic case is recovered.

totto82: I hope #1855 will fix the problem. Testing in progress......

totto82: #1855 fixed the problem.

GitPaean: Good! What is the actual problem? Some exceptions we could not catch?

totto82: With #1855 all exceptions should be caught. I have added a partial fix for this in this PR. I will remove it when #1855 is merged.

GitPaean: Good! How the performance looks like? similar?

totto82: I am testing it at the moment. I.e. The same case with `2e5`

opm/simulators/wells/StandardWellV_impl.hpp
```diff
@@ -2265,12 +2273,42 @@ namespace Opm
     void
     StandardWellV<TypeTag>::
     computeWellRatesWithBhp(const Simulator& ebosSimulator,
-                            const EvalWell& bhp,
+                            const std::vector<Scalar>& B_avg,
```
GitPaean: I am trying to rewriting this comment. 

I suggest to split this function to be two functions. 

The first part basically applies to the part that when `iterate` to be `true`. It includes resetting the bhp limit at the beginning, solving the well equations and update the explicit quantities, and the recover the the well controls at the end. Please make a proper name for it. 

The second function will be the original `computeWellRatesWithBhp` unchanged. 

With this two functions, you can make a new function called `computeWellPotentialsWithBhpLimit`.

GitPaean: The resetting of the BHP limit in the current form is not very proper, since it discard the fact that there might be other controls included there at the same time. It did not cause problem is because it is only used when `iterate` to be `true`. 

And also, we have two places resetting the BHP limits. The other places is the `computeWellPotentials` in `BlackoilWellModel`. We should not need to do it twice. 

I will suggest we only do it in one places.  The well control only contains a BHP limit can be created using some function as follows, 
```
    template<typename TypeTag>
    struct WellControls*
    WellInterface<TypeTag>::
    creatingWellControlWithBhpTargetOnly(const double bhp)
    {
        struct WellControls* controls = well_controls_create();
        const double invalid_alq = -1.e100;
        const double invalid_vfp = -1.;
        well_controls_add_new(BHP, bhp, invalid_alq, invalid_vfp, NULL, controls);

        assert(well_controls_get_num(controls) == 1);
        well_controls_set_current(controls, 0);
        return controls;
    }
```

Then we can do the resetting the well control in some manner as follows,
```
        // back up the old control data
        const struct WellControls* original_well_controls = well_controls_;
        const int original_current_control = well_controls_get_current(original_well_controls);
```

then 
```
well_controls_ = this->creatingWellControlWithBhpTargetOnly(bhp);
```

At the end, we recover the old control

```
        // recovering the old well control
        well_controls_destroy(well_controls_);
        well_controls_ = original_well_controls;
        well_controls_set_current(well_controls_, original_current_control);
```

All these should be the first function mentioned above.  We do not need to do it in the `computeWellPotentials` in `BlackoilWellModel`.

GitPaean: In MultisegmentWell, the suggestions will be similar. 

Ideally, we should not need to recreate the `well_container`. I do see that it might needs less cleaning and recovering work by creating a new one.

GitPaean: Great you find that the well potentials can be more correct when updating the explicit quantities after solving the system til converged under the desired BHP target. 

It is possible in the future we should always do that, while let us do it step by step.

opm/autodiff/BlackoilModelParametersEbos.hpp
totto82: The 1e6 speeds up the computations due to less convergence issues.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -544,58 +543,60 @@ namespace Opm {
...
-                    if (well_state_.effectiveEventsOccurred(w)) {
-                        if (wellTestState_.lastTestTime(well_name) == ebosSimulator_.time()) {
-                            // The well was shut this timestep, we are most likely retrying
-                            // a timestep without the well in question, after it caused
-                            // repeated timestep cuts. It should therefore not be opened,
-                            // even if it was new or received new targets this report step.
-                            well_state_.setEffectiveEventsOccurred(w, false);
-                        } else {
-                            wellTestState_.openWell(well_name);
+                if (allow_closing_opening_wells) {
```
GitPaean: if `allow_closing_opening_wells` is `false`, does this mean, even a well is shut due to economic or physical reason, we still need to calculate the well potentials?

maybe it is possible for the wells shut due to economic reasons, while it can be hard for the wells shut due to physical reason. 

I think it is reasonable to assume that we set the well potentials to be zero or not trying to calculate the well potentials for the wells are shut.

opm/autodiff/BlackoilModelParametersEbos.hpp
GitPaean: That is good to know. Is this the one problematic case we were trying to fix?  Can you post the performance information here so that I do not need to run it.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -541,24 +542,73 @@ namespace Opm
...
+    void
+    MultisegmentWell<TypeTag>::
+    computeWellRatesWithBhp(const Simulator& ebosSimulator,
+                            const std::vector<Scalar>& B_avg,
+                            const double& bhp,
+                            const bool iterate,
+                            std::vector<double>& well_flux,
+                            Opm::DeferredLogger& deferred_logger)
+    {
+        // store a copy of the well state, we don't want to update the real well state
```
GitPaean: Suggestion on how to handle this function is similar to the previous comment with StandardWell. The well_controls_ is not reset here, it is reset in `BlackoilWellModel`. It might not be correct since a THP limit can be there at the same time. 

Eventually, if our code can just set the BHP and THP together, then we just solve it, the resulting well rates will be the well potentials, that will be great and that should be our goal, while I do not think we can do that in this PR, at least not for MSW.  We still have to calculate under BHP and THP separately and pick the one gives the smaller well potential value.

GitPaean:
My current comments are done. Please address following the comments. 

This PR is much better now compared with its original form. And it is hard to tell what is good way / correct way to do it. Continuous improvements will be required even after this PR, while it can be easier just try to little by little, which can result better understanding and smaller risk.  

Currently, due to the existence of the explicit quantities, StandardWell has more complicated logic than MSW in some sense.  Maybe we should switch to implicit for StandardWell too. Then the logic can be simpler. We just solve it, not updating a little bit to make the result slightly better. But it is good you found out that one more updating of the explicit quantities can improve the well potential calculation, which is an important clue.

GitPaean:
By `complicated logic` related to StandardWell, I mean we have some explicit quantities, but sometimes we want to update it by solving once, because we found it makes the result slightly better. It is possible by updating it twice, the result will be better slightly further. If we keep updating it until result not change anymore, it is kind of implicit in some sense. Maybe by treating them implicit, we will be able to simplify the logic.

alfbr:
A conflict resolution is highly appreciated, would love to build again and test more :-)

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

bska:
jenkins build this please

totto82:

totto82:
jenkins build this opm-tests=110 please

totto82:
jenkins build this opm-tests=110 please

bska:
The data update PR (OPM/opm-tests#110) was created before we merged OPM/opm-common#707 and its associate data update, OPM/opm-tests#114. Therefore it is likely that OPM/opm-tests#110 is stuck referencing older data. In previous cases like this I've been able to work around that problem by closing the data update PR&mdash;here OPM/opm-tests#110&mdash;for about one hour, before reopening it and rerunning `update_data`.

I would however advice not doing so until all reviewers of this PR has approved of the changes so that the only remaining step is the final data update.

atgeirr:
This PR can now be merged, when green. Thanks for your patience and the changes you have made. The data update things suggested by @bska above are ok to to now.

totto82:

atgeirr:
jenkins build this opm-tests=110 please

PR #1803, opened 2019-04-24
Remove warnings experienced with g++-8 

blattms:
Would be nice if this would be merged into the release, too.

### DIFF
--- a/opm/autodiff/BlackoilWellModel_impl.hpp
+++ b/opm/autodiff/BlackoilWellModel_impl.hpp
@@ -1737,7 +1737,7 @@ namespace Opm {
                 const auto& segments = well.segments;
 
                 // \Note: eventually we need to hanlde the situations that some segments are shut
-                assert(int(segment_set.size()) == segments.size());
+                assert( 1u + segment_set.size() == segments.size());
 
                 for (const auto& segment : segments) {
                     const int segment_index = segment_set.segmentNumberToIndex(segment.first);
--- a/opm/autodiff/BlackoilWellModel_impl.hpp
+++ b/opm/autodiff/BlackoilWellModel_impl.hpp
@@ -38,7 +38,7 @@ namespace Opm {
     template<typename TypeTag>
     void
     BlackoilWellModel<TypeTag>::
-    init(const Opm::EclipseState& eclState, const Opm::Schedule& schedule)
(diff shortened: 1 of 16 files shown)


blattms:
jenkins build this please

bska:
No opinion on the PR itself. I just notice that `BlackoilWellModel_impl.hpp` has some [conflict resolution markers](https://github.com/OPM/opm-simulators/blob/a9d59c0084cbc41e92005f5bdda17ab07fe9491b/opm/autodiff/BlackoilWellModel_impl.hpp#L1740-L1744) that need to be addressed before this will successfully build.

blattms:
Oops. Indeed the rebasing was pretty messed up. Sorry. Fixed now.

blattms:
jenkins build this please

blattms:
jenkins build this please

bska:
I'm going to merge this into master now. I don't see any reason to hold it back anymore. The change-set removes a set of warnings and does not alter simulation results.

opm/autodiff/BlackoilWellModel_impl.hpp
```diff
@@ -1737,7 +1737,7 @@ namespace Opm {
                 const auto& segments = well.segments;
 
                 // \Note: eventually we need to hanlde the situations that some segments are shut
-                assert(int(segment_set.size()) == segments.size());
+                assert( 1u + segment_set.size() == segments.size());
```
GitPaean: Just curious why we need to add 1 here? Is it because other changes somewhere? I am not saying it is wrong, since the tests pass.

GitPaean: I now think it is a mistake. We have not got hit by it is because it only called by a restarting case with multi-segment well, with which we do not have a test case yet.

GitPaean: Will the original form `int(segment_set.size()) == segments.size()` cause warning with g++-8.0?

GitPaean: @blattms

blattms:
```
assert( 1u + segment_set.size() == segments.size())
```

Indeed, that is aserious bug. Thanks for the hint. Should be 0u. Will fix.

`WellSegment::size() return an int and the right hand size a `std::size_t`.

atgeirr:
> Will fix.

Already fixed and merged in #1808.

PR #1804, opened 2019-04-25
BUGFIX. Make sure that the convergece report is stored before throw

totto82:
@akva2 This should be part of the release.


totto82:
jenkins build this please

PR #1806, opened 2019-04-26
[bugfix] Use NNC data with applied EDITNCC for transmissibility.

blattms:
Unfortunately, we first created NNC with applied EDITNNC and then still used the original NNC data to set the transmissibility. Thus we were actually ignoring EDITNNC.

This commit fixes this by using the data structure that has EDITNNC applied.


blattms:
@akva2 Should probably also be in the release?

blattms:
jenkins build this please

PR #1808, opened 2019-04-30
Correcting a small assertion mistake introduce in PR#1803

GitPaean:
And also removing the explicit using of `--use-multisegment-well` in `compareECLFiles` since the multisegment well is activated by default through PR #1807


GitPaean:
jenkins build this please

PR #1814, opened 2019-05-02
updated test scripts to use eclipse compatible restart

tskille:
Reference solutions must be updated since arrays used by opm compatible restart, no longer will be output to restart. It has been verified that remaining arrays still are within tolerances.


akva2:
jenkins build this please

akva2:
this is ready to be merged, but we're having a mexican standoff situation on the opm-tests repo right now, and i don't want to annoy people by jumping ahead in queue..

PR #1815, opened 2019-05-02
Use new well implementation Well2 from opm-common

joakim-hove:
With this PR the simulator is updated to use the new `Well2` class from opm-common. The changes in this PR are mostly mechanical - and include:

1. The well type has been renamed to `Well2`.
2. The previous API made extensive use of naked `Well *` pointers, we now use `Well2` or `Well2&`.
3. Previously the properties - like connection sets and production properties were all accessed through a function with a timestep argument - that has been removed.

Downstream of: https://github.com/OPM/opm-common/pull/731

The work in opm-common should get more testing and scrutiny before merge, but the part here  should not change significantly whatever happens to opm-common - and hence should be ready for review.

### DIFF
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -191,7 +191,7 @@ public:
 
             //distribute the grid and switch to the distributed view.
             {
-                const auto wells = this->schedule().getWells();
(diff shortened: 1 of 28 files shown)


joakim-hove:
jenkins seems stuck on an old ref

PR #1816, opened 2019-05-02
ebos: use the non-multisegmented well model by default

andlaus:
this avoids regressions for decks that use well testing and makes `ebos` work as expected if UMFPACK is not available, but obviously it will not work for decks that use multisegment wells in earnest.

`flow` is unaffected by this because it does not use this type tag.

### DIFF
--- a/ebos/ebos.hh
+++ b/ebos/ebos.hh
@@ -56,6 +56,11 @@ SET_BOOL_PROP(EbosTypeTag, EnableExperiments, true);
 // use flow's well model for now
 SET_TYPE_PROP(EbosTypeTag, EclWellModel, Opm::BlackoilWellModel<TypeTag>);
 
+// currently, ebos uses the non-multisegment well model by default to avoid
+// regressions. the --use-multisegment-well=true|false command line parameter is still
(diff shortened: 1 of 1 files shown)


andlaus:
jenkins build this pleas

andlaus:
jenkins build this please

GitPaean:
Following discussion in https://github.com/OPM/opm-common/pull/278 , UMFPack is intended to be a hard prerequisite, while the implementation might not work as expected following your comment, then we should fix it.

andlaus:
sure, but well testing will not work even with MSW even if UMFPACK is available!? IMO this implies that `ebos` should stick to the standard well model for now to avoid this regression. (note that I do not have an opinion about the question of whether or not `flow` should trade in better support for MSW for well testing...)

GitPaean:
> sure, but well testing will not work with MSW even if UMFPACK is available!?

Sure. If well testing is the important feature you need for you current development, then it is fine. There are a few important features available for StandardWell while not in place for MSW yet.

totto82:
Lets keep MSW out of default ebos while it matures. I will merge this.

PR #1818, opened 2019-05-03
[WIP] Make NNC output more eclipse compliant

blattms:
There were some hickups in the NNC output to the eclipse files due to the way we determine them. We used the NNC as specified by the deck (without EDITNNC scaling) and then iterated through the grid and added an NNC for each face between non-cartesian neighbours with the transmissibilities as used by the simulator.  Note that in these transmissibilities small once (i.e. <1.0e-6)  are substituted by zero during setup of the simulator. This leads to the following deviations from the way eclipse does it or problems:

1. We wrote all NNCs even small ones (specified by the deck) and zero ones (those substituted during the setup). Eclipse ignores them.
2. NNCs that were specified in the deck and not scaled by EDITNNC were output twice (bug!) with the same transmissibility.
3. NNCs that were specified in the deck and scaled by EDITNNC were output twice with the differing transmissibility. Once with the unscaled one and once with the scaled one. These are actually two bugs.
4. If we pinch out cells then we write an NNC between the newly connected vertical neighbors and will not write this transmissibility out in the TRANZ keyword. While does not seem wrong, but is different from Eclipse, which will not write an NNC here but will write the transmissibility via TRANZ.

 Currently this PR addresses 1-3. Hence the WIP tag. How to fix 4) and whether we even need to fix it is up for discussion.


blattms:
jenkins build this please

ebos/nncsorter.hpp
```diff
@@ -0,0 +1,34 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
```
andlaus: this file seems to be rather superfluous. also: please put the code directly into the header for core ebos...

ebos/nncsorter.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include <opm/common/OpmLog/OpmLog.hpp>
+#include <opm/parser/eclipse/EclipseState/Grid/NNC.hpp>
+
+#include <sstream>
+#include <vector>
+#include <algorithm>
+#include <iostream>
+namespace Ewoms
+{
+namespace Detail
```
andlaus: please remove the `Detail` namespace: nested namespaces should be avoided in opm-material, ewoms and core-ebos. (as should deeply nested locations of include files.)

ebos/eclwriter.hh
```diff
@@ -419,18 +458,46 @@ private:
                 // TODO (?): use the cartesian index mapper to make this code work
                 // with grids other than Dune::CpGrid. The problem is that we need
                 // the a mapper for the sequential grid, not for the distributed one.
-                int cc1 = globalGrid_.globalCell()[c1];
-                int cc2 = globalGrid_.globalCell()[c2];
+                std::size_t cc1 = globalGrid_.globalCell()[c1];
```
andlaus: nit: this is probably a case where `auto` would be more readable because if it is decided that the return value of `grid.globalCell()[c1]` should change its signeness, no follow up to get rid of the warning will be needed.

ebos/eclwriter.hh
```diff
@@ -372,9 +376,44 @@ private:
 
     Opm::NNC exportNncStructure_() const
     {
-        Opm::NNC nnc = eclState().getInputNNC();
-        int nx = eclState().getInputGrid().getNX();
-        int ny = eclState().getInputGrid().getNY();
+        std::size_t nx         = eclState().getInputGrid().getNX();
```
andlaus: no ascii art for aligning lines please. (because this makes it basically impossible to search-and-replace variable names without breaking the formatting.)

ebos/eclwriter.hh
```diff
@@ -372,9 +376,44 @@ private:
...
+        std::size_t ny         = eclState().getInputGrid().getNY();
+        auto nncData           = eclState().getInputNNC().nncdata();
+        auto editnncData       = eclState().getInputEDITNNC().data();
+        const auto& unitSystem = simulator_.vanguard().deck().getActiveUnitSystem();
+        auto edited = Detail::sortNncAndApplyEditnnc(nncData, editnncData);
+        std::vector<Opm::NNCdata> outputNnc;
+        std::size_t index = 0;
+
+        for( auto& entry : nncData )
```
andlaus: `const auto&`?

blattms:
When comparing the output for a Norne run we have the following picture:
- Eclipse outputs one NNC that is there when using flow: (14,16,07) -> (13,16,09).
- All NNCs output by flow but not by eclipse seem to be  a result from pinch outs (i.e. the cells the would between the two neighbours in a cartesian grid are not active).

ebos/nncsorter.hpp
blattms: No it is not. It is need to to reduce code duplication, as these functions are at two different places: ecltransmissibility.hh and eclwriter.hh.
Putting everything in the header is also not possible (without abusing the `inline` keyword) as you will get errors during linking because of multiply defined symbols.

ebos/nncsorter.cpp
blattms: Find by me but it exposes the implementation details ebos.
What the heck is core-ebos? This is opm-simulators!

ebos/eclwriter.hh
blattms: Well, this is exactly there to prevent such warnings as `globalGrid_.glglobalCell()` returns a `std::vector<int>` and `eclState().getInputGrid().getNX()` a `std::size_t`. I just reckoned that a negative global index would never appear. That is up for debate of course. But it cannot become `auto` without producing warnings. So it is either the new type or the old. I do not care.

ebos/nncsorter.cpp
andlaus: core-ebos is everyting under the `ebos` directory.

ebos/eclwriter.hh
blattms: I doubt this statement. I recall having done this with emacs search-and-replace-regex. 
On the other hand I do not care and will change it just to make you happy.

ebos/eclwriter.hh
andlaus: okay, seems like that API has a lot of potential for improvement...

ebos/eclwriter.hh
blattms: Nope. I need to change the entry to faciliate searching (cell1<=cell2 needs to hold). See a few lines below.

ebos/nncsorter.cpp
andlaus: also: the fact that it is an implementation detail is -- in my not so humble opinion -- made abundantly clear by the function name...

ebos/eclwriter.hh
blattms: `int` or `std::size_t`? Your call.

ebos/nncsorter.hpp
andlaus: okay if you really insist it can stay. The compile unit must go, though...

ebos/eclwriter.hh
andlaus: I constantly have problems with this and if the author does not notice this, it results in botched-up formatting. I'm pretty sure that you can find quite a few instances of this in the OPM and DUNE code bases...

blattms:
jenkins build this please

ebos/eclwriter.hh
andlaus: that's what I'd call surprising behavior: if some objects in a container needs to be modified, it should be much more easily recognizable, e.g. `container[nncIdx].value = foo();`...

ebos/eclwriter.hh
andlaus: I don't really care. `int` does reduces the code grunt and is the type stored in the container stored `globalCell()`, so this is IMO slightly better.

ebos/nncsorter.hpp
blattms: Sorry, but I do not understand why and how.

ebos/eclwriter.hh
blattms: Mmh, then I need to change `auto cellDiff` to ìnt cellDiff, too. Otherwise I will get
```
dune/opm/opm-simulator/ebos/eclwriter.hh:399:43: warning: comparison of integer expressions of different signedness: ‘long unsigned int’ and ‘int’ [-Wsign-compare]
             if (cellDiff != 1 && cellDiff != nx && cellDiff != nx*ny)
                                  ~~~~~~~~~^~~~~
```
And I would need to do another cast to prevent:
```
src/dune/opm/opm-simulator/ebos/eclwriter.hh:480:76: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare]
                   while ( candidate != nncData.end() && candidate->cell1 == cc1
                                                         ~~~~~~~~~~~~~~~~~^~~~~~
```
I will stick with the way this is and wonder whether this nitpick was really worth investing any time (and time it did cost!).

ebos/nncsorter.hpp
andlaus: opm-material, ewoms and core-ebos are currently all header only. I'll argue that this constitutes a significant improvement compared to the days when you had to recompile several modules to develop most features: e.g., you had to recompile opm-core, opm-grid and opm-autodiff after any changes you made in the transmissibility code and if you forgot some of them it resulted in *very* weird symptoms. now you can do this by only recompiling the simulator of interest. I want core-ebos to stay a "good upstream citizen".

ebos/eclwriter.hh
andlaus: I clearly stated that it is a nit, and I do not insist on fixing nits before merging. If you wasted time because of this, I don't have an idea what else I should do about this...

blattms:
The test failures are expected as we really change the number of entries and in case of EDITNNC even values of the keywords NNC1, NNC2 and TRANNNC. But someone else should recheck whether this makes sense.

ebos/eclwriter.hh
blattms: I did even forget that int = unsigned int is a narrowing conversion.

blattms:
@andlaus Please make proposal in terms of PRs for the rest. That was other can discuss. 

IMHO this PR is as good as it gets in the areas you commented. But I can be wrong and you can prove it.

andlaus:
see blattms/opm-simulators#4. While doing this, I noticed that the sortnnc lacks a doxygen comment.

andlaus:
jenkins build this please

andlaus:
Since you did not merge blattms/opm-simulators#4, I'd like to merge and do this cleanup in a separate PR. Unfortunately, jenkins is not happy about this...

blattms:
The jenkins error is expected as pointed out above:

diff of NNCs for opm-tests/editnnc/EDITNNC.DATA:
```
$ diff master-editnnc branch-editnnc
--- master-editnnc      2019-05-06 14:14:20.825836669 +0200
+++ branch-editnnc      2019-05-06 14:14:54.631047526 +0200
@@ -19,25 +19,10 @@
 (04,05,02) -> (03,05,04)  T:0.167721 P:0
 (04,05,02) -> (03,05,05)  T:0.000239151 P:0
 (04,01,03) -> (03,01,05)  T:0.00071147 P:0
-(04,01,03) -> (03,01,06)  T:0 P:0
 (04,02,03) -> (03,02,05)  T:0.00071147 P:0
-(04,02,03) -> (03,02,06)  T:0 P:0
 (04,03,03) -> (03,03,05)  T:0.000711469 P:0
-(04,03,03) -> (03,03,06)  T:0 P:0
 (04,04,03) -> (03,04,05)  T:0.00071147 P:0
-(04,04,03) -> (03,04,06)  T:0 P:0
 (04,05,03) -> (03,05,05)  T:0.00071147 P:0
-(04,05,03) -> (03,05,06)  T:0 P:0
-(04,01,04) -> (03,01,06)  T:0 P:0
-(04,01,04) -> (03,01,07)  T:0 P:0
-(04,02,04) -> (03,02,06)  T:0 P:0
-(04,02,04) -> (03,02,07)  T:0 P:0
-(04,03,04) -> (03,03,06)  T:0 P:0
-(04,03,04) -> (03,03,07)  T:0 P:0
-(04,04,04) -> (03,04,06)  T:0 P:0
-(04,04,04) -> (03,04,07)  T:0 P:0
-(04,05,04) -> (03,05,06)  T:0 P:0
-(04,05,04) -> (03,05,07)  T:0 P:0
 (04,01,05) -> (03,01,07)  T:7.12146e-06 P:0
 (04,02,05) -> (03,02,07)  T:7.12146e-06 P:0
 (04,03,05) -> (03,03,07)  T:7.12146e-06 P:0
$
```
And for NNC_EDITNNC case:
```
$ diff master-nnc-editnnc branch-editnnc
--- master-nnc-editnnc  2019-05-06 14:14:37.022416306 +0200
+++ branch-nnc-editnnc  2019-05-06 14:15:21.019995459 +0200
@@ -19,29 +19,12 @@
 (04,05,02) -> (03,05,04)  T:0.167721 P:0
 (04,05,02) -> (03,05,05)  T:0.000239151 P:0
 (04,01,03) -> (03,01,05)  T:0.00071147 P:0
-(04,01,03) -> (03,01,06)  T:0 P:0
 (04,02,03) -> (03,02,05)  T:0.00071147 P:0
-(04,02,03) -> (03,02,06)  T:0 P:0
 (04,03,03) -> (03,03,05)  T:0.000711469 P:0
-(04,03,03) -> (03,03,06)  T:0 P:0
 (04,04,03) -> (03,04,05)  T:0.00071147 P:0
-(04,04,03) -> (03,04,06)  T:0 P:0
 (04,05,03) -> (03,05,05)  T:0.00071147 P:0
-(04,05,03) -> (03,05,06)  T:0 P:0
 (03,01,04) -> (04,01,05)  T:9.5 P:0
-(03,01,04) -> (04,01,05)  T:9.5 P:0
-(04,01,04) -> (03,01,06)  T:0 P:0
-(04,01,04) -> (03,01,07)  T:0 P:0
-(03,02,04) -> (04,02,05)  T:7.6 P:0
 (03,02,04) -> (04,02,05)  T:7.6 P:0
-(04,02,04) -> (03,02,06)  T:0 P:0
-(04,02,04) -> (03,02,07)  T:0 P:0
-(04,03,04) -> (03,03,06)  T:0 P:0
-(04,03,04) -> (03,03,07)  T:0 P:0
-(04,04,04) -> (03,04,06)  T:0 P:0
-(04,04,04) -> (03,04,07)  T:0 P:0
-(04,05,04) -> (03,05,06)  T:0 P:0
-(04,05,04) -> (03,05,07)  T:0 P:0
 (04,01,05) -> (03,01,07)  T:7.12146e-06 P:0
 (04,02,05) -> (03,02,07)  T:7.12146e-06 P:0
 (04,03,05) -> (03,03,07)  T:7.12146e-06 P:0
```

The files {master,branch}*-editnnc are generated with a patched version of libecl/python/examples/cmp_nnc.py (T:<transmissibility> P:<is pinch out NNC>). In these tests only the NNCs with small transmissibility are omitted. Same holds for Norne, see attachment 
[master-editncc-branch-diff.txt](https://github.com/OPM/opm-simulators/files/3147928/master-editncc-branch-diff.txt)

blattms:
Just for clarity:

All this nitpicking actually made me look at the code again and discover some deficiencies that were not discovered previously (e.g. debug output). I tried to adhere to @andlaus's coding style changes when editing the history. The const in the for-loop that he was missing (and that I previously said cannot be there) is gone now due to refactoring of the sorting code. Indeed, this is now more elegant.

The only thing I did not address is the wish for header only. IMHO blattms#4 is an abuse of inline and this (helper) code would be compiled thrice for no reason. There is also already a lib in opm-simulators. I do not concur with the outlined superiority of header only. Or maybe I am just not smart enough to get it. If this is really wanted by the majority of the community it can be addresses in another PR (even by me).

blattms:

blattms:
WIP removed as topic 4) will not currently addressed.

blattms:
I guess I have messed up the reference tests a bit here. I thought update would also touch the opm-tests/editnnc stuff. Seems like I am wrong. In that case, how do prevent the automatically created PR from merging?
Can someone more knowledgeable take a look and help me fix this?

akva2:
hmm? it updated the nnc test, but it also updated a lot of other tests, which is not really expected is it?

blattms:
Sorry @akva2. I should have looked at that PR. Now I see that is updates all tests. Only for NORNE_ATW2013 and NNC_AND_EDITNNC the size of the binary files changes. Seems ok, I guess.

I was just expecting the check to succeed.

akva2:
ah no, it's a two step process. we now run against the updated data.

akva2:
jenkins build this opm-tests=112 please

andlaus:
> The only thing I did not address is the wish for header only.

okay, I'll do this myself once the present PR is in. also, please squash before merge: IMO it is not the best idea to have 10 almost trivial commits which have not been tested for bisectability in the master.

blattms:
They have been tested, and present logical steps in an iterative process towards the solution.

BTW: Thanks for calling this work trivial.

andlaus:
Please do not consciously misinterpret me, I didn't say that: I said that most of these patches trivial and i meant this as in "the-usual-back-and-forth-during-development", the overall changes are not trivial and I want the final result in master. That said I'd like to keep the commit history reasonable, i.e., I'm not so much interested in being able to retrace the thinking process that went into a change. If you want to invest the time to split this PR into self-contained commits that fully bisectable, I would appreciate this effort.

bska:
Is there expected to be any more development here? This is partly blocking OPM/opm-common#707, because that PR also needs to run `update_data` and will alter all reference solution `.INIT` files.

If there is no more expected development here, then I would very much appreciate it if we could merge this into master soon. Otherwise, I may just have to merge OPM/opm-common#707 as that work is complete now and there is other worked blocked on 707 becoming available in master. Note that merging 707 before this (#1818) gets in will create conflicts in this PR's associated data update PR (OPM/opm-tests#112).

andlaus:
From my side this is good to go once the commit history has been sanitized (i.e., either all commits have been squashed together or -- preferably -- refactored into a bisectable series of self-contained commits that each contains a functional change that appears in the final result.)

totto82:
> They have been tested, and present logical steps in an iterative process towards the solution.

I guess this already means 

> -- refactored into a bisectable series of self-contained commits 

i.e. this can be merged. It shouldn't even bother Atgeirrs refactoring since it doesn't touch code under opm/simulators

PR #1825, opened 2019-05-08
Adjust to shared_ptr change in dune-istl >= 2.7.

dr-robertk:
This PR fixes compilation issues with flow when using the current DUNE master. In dune-istl the return value of the ConstructionTraits and the Constructor parameters have been changed from raw pointers to std::shared_ptr.

### DIFF
--- a/opm/simulators/linalg/BlackoilAmg.hpp
+++ b/opm/simulators/linalg/BlackoilAmg.hpp
@@ -127,7 +127,7 @@ scaleMatrixDRS(const Operator& op, std::size_t pressureEqnIndex, const Vector& w
         for (auto i = matrix->begin(); i != endi; ++i) {
             const BlockVector& bw = weights[i.index()];
             const auto endj = (*i).end();
-            for (auto j = (*i).begin(); j != endj; ++j) {  
+            for (auto j = (*i).begin(); j != endj; ++j) {
                 Block& block = *j;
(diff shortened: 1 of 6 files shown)


dr-robertk:
@bska: I think some of these are to be shared variables and then some maybe not. But then unique_ptr (which I like better also) is used, then things like copying become very difficult. But I have not been involved in the decision.

atgeirr:
I think shared_ptr communicates clearly how it will be used inside. In dune there are many constructors taking const reference arguments, and that can mean several things:
 - I will use this object only in construction. If you change it in the outside you should construct me again.
 - I will keep a reference, and expect it both to remain valid and that the underlying object is not changed.
 - I will keep a reference, and expect it to remain valid. If you change the underlying object that is fine, and it will change my behaviour.
 - I will take a copy, and you may do whatever you like to the original passed-in object.

With move semantics the last one should always be replaced with pass-by-value, other than that I think the three other options are reasonable enough to require distinguishing between them in class docs or by convention requiring one of them. Using shared_ptr is like the third option above, except that it is guaranteed to remain valid even if the passed-in object goes out of scope.

@blattms There are lots of such reference constructor args in dune-istl, is there a convention for how they are treated (i.e. one of the options above)?

dr-robertk:
jenkins build this please

dr-robertk:
jenkins build this please

atgeirr:
Looks good and is green, merging.

PR #1823, opened 2019-05-08
Make NNC output more eclipse compliant - with bisectability

akva2:
In interest of removing the hold on the work progress;

This is https://github.com/OPM/opm-simulators/pull/1818 with the requested bisectability (tested using make check-commits).

PR #1826, opened 2019-05-08
ebos/flow: make the transmissibility code easily replaceable

andlaus:
it works by introducing a "TransmissibilityCalculator" property at compile time. The default is to use `Ewoms::EclTransmissibility<TypeTag>`.

besides this, this commit fixes a potential output bug: the results of `sortNncAndApplyEditnnc()` were put through `std::sort()` again before written, so if the sorting within `sortNncAndApplyEditnnc()` was changed in any way, this would not have been reflected by the output...


andlaus:
jenkins build this please

blattms:
mmh, so now we are using the property system to make ebos headers only?

tests/test_nncsorter.cpp
```diff
@@ -17,9 +17,11 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+#include "config.h"
+
 #define BOOST_TEST_MODULE NNCSortTest
+#include <ebos/ebos.hh>
```
atgeirr: This is a symptom of overuse of the tag system. Including the full ebos just for a typetag is overkill and should be avoided when possible.

atgeirr:
I still think we should not force this to be header-only.

> this commit fixes a potential output bug

This should be dealt with in any case of course.

tests/test_nncsorter.cpp
andlaus: please suggest a better solution of being able to specify everything in a graph with 150 notes that is potentially fully connected and works for highly performance sensitive stuff.

joakim-hove: > [...] specify everything in a graph with 150 notes that is potentially fully connected

Well - that is the problem, the fact that you need to include the full simulator to be able to *test* a facility for sorting NNC is a disease.

andlaus: > [...] the fact that you need to include the full simulator to be able to test a facility for sorting NNC is a disease.

it is a choice between two evils: either you chose to go with a myriad of template parameters and the resulting code becomes *really* impervious (cf. e.g. dune-pdelab which has template parameters for template parameters for template parameters, and -- to a somewhat lesser extend -- the ECL satfunc implementation in opm-material), or you go with a glorified traits approach/property system. I'll argue that the latter is easier to develop with and to understand beyond a certain complexity like for a simulator but I freely acknowledge that this approach has drawbacks as well, as you point out. Also, as always: PRs for better solutions are appreciated.

atgeirr: While it is true that a complete simulator will have either a myriad of template params or have to use a traits system, it is not true for many sub-parts, such as the code in question here. There is no reason sub-parts cannot use a few template parameters (or as in the current case: none) instead of the full traits/tag, while the simulator is still configured using it (and extracts the template arguments to be used for its sub-parts from it).

andlaus: > it is not true for many sub-parts, such as the code in question here.

Please tell me how the transmissibility calculation code can be made swappable otherwise. an alternative would be to introduce a separate property for the `sortNnc()` function, but that would be quite a bit more ugly than the proposed solution. (also the `sortNnc()` function was there before #1823, in retrospect it was a mistake to let it diffuse out of that class.)

andlaus: (i.e., I should have been more "nitpicky" before merging #1823. I guess that this is my take-home message of that episode.)

atgeirr: I am not arguing against adding a `TransmissibilityCalculator` property (although I think we should not add it until we need it, as it is often a mistake to add a customization point before we have any use of it). I am arguing against making the nnc sorting facility depend on the type tag.

andlaus: > although I think we should not add it until we need it, as it is often a mistake to add a customization point before we have any use of it

you changed your opinion recently? cool!

> I am arguing against making the nnc sorting facility depend on the type tag.

if you don't do this, you cannot change the sorting facility when you set a custom transmissibility calculator!?

blattms:
> Please tell me how the transmissibility calculation code can be made swappable otherwise.

Use dynamic polymorphism. The performance penalty should be nearly zero here.

andlaus:
this does not solve anything: you still need to move the actual code into the class, and the class requires a `TypeTag`.

tests/test_nncsorter.cpp
atgeirr: > you changed your opinion recently? cool!

I have held this opinion for a long time. So is there a case for adding custom transmissibility calculators?

> if you don't do this, you cannot change the sorting facility when you set a custom transmissibility calculator!?

Why do they have to change in lock-step?

andlaus: >>  you changed your opinion recently? cool!
>
> I have held this opinion for a long time. So is there a case for adding custom transmissibility calculators?

two words: interface classes.

>> if you don't do this, you cannot change the sorting facility when you set a custom transmissibility calculator!?
>
> Why do they have to change in lock-step?

because they logically belong together and it is IMO madness to have two different knobs to change change the same thing.

bska:
I'm closing this as it has become stale.  If we wish to do something like this in the future it is better to start from scratch.

PR #1824, opened 2019-05-08
ebos: remove the newly introduced compile unit for 'nncsorter.cpp'

andlaus:
instead, move the code to the respective header file like everything else in opm-material/eWoms/core ebos. IMO, this change makes ebos a better upstream citizen again, because it allows to use the non-flow-wells parts of it without having to worry about libraries.


andlaus:
superseeded by #1826. closing.

Issue #1828, opened 2019-05-09
Simulation Restart Broken in Current Master


bska:
I was first notified of this problem in the context of a real field case, but I have been able to reproduce the symptom on OPM Tests' [`SPE1CASE2`](https://github.com/OPM/opm-tests/blob/7e678982e76cb4baaef0e4bc6f258919e1d2e674/spe1/SPE1CASE2.DATA) model.

From a normal base run of `SPE1CASE2`, define a restart run `SPE1CASE2-RST` by the following change:
```diff
--- ./SPE1CASE2.DATA	2019-05-09 11:20:24.588971166 +0200
+++ ./SPE1CASE2-RST.DATA	2019-05-09 11:23:10.716977454 +0200
 -- 	   - there must be two wells in a group as there are two wells in total
    2 1 1 2 /
 
+UNIFIN
 UNIFOUT
 
 GRID
@@ -250,7 +250,7 @@
 SOLUTION
 -- -------------------------------------------------------------------------
 
-EQUIL
+--EQUIL
 -- Item 1: datum depth (ft)
 -- Item 2: pressure at datum depth (psia)
 -- 	   - Odeh's table 1 says that initial reservoir pressure is 
@@ -268,7 +268,10 @@
 -- Item 9: Set to 0 as this is the only value supported by OPM
 
 -- Item #: 1 2    3    4 5    6 7 8 9
-	8400 4800 8450 0 8300 0 1 0 0 /
+--	8400 4800 8450 0 8300 0 1 0 0 /
+
+RESTART
+  SPE1CASE2 42 /
 
 RSVD
 -- Dissolved GOR is initially constant with depth through the reservoir.
```

Then, when running `SPE1CASE2-RST` you get
```
**********************************************************************
*                                                                    *
*                      This is flow 2019.10-pre                      *
*                                                                    *
* Flow is a simulator for fully implicit three-phase black-oil flow, *
*             including solvent and polymer capabilities.            *
*          For more information, see https://opm-project.org         *
*                                                                    *
**********************************************************************

Reading deck file './SPE1CASE2-RST.DATA'
Program threw an exception: Could not initialize the problem.
```

Backtrace in GDB says:
```
(gdb) bt
#0  0x00007ffff629d1d0 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
[...]

PR #1830, opened 2019-05-09
fixed: use of restartTimeStep() before it has been read

akva2:
it is read in the beginRestart() method, reorder code accordingly.

ref valgrind log
```
==45745== Conditional jump or move depends on uninitialised value(s)
==45745==    at 0x177DC2A: double const& std::min<double>(double const&, double const&) (stl_algobase.h:200)
==45745==    by 0x17FF497: Ewoms::EclProblem<Ewoms::Properties::TTag::EclFlowProblem>::readEclRestartSolution_() (eclproblem.hh:2450)
==45745==    by 0x17C9451: Ewoms::EclProblem<Ewoms::Properties::TTag::EclFlowProblem>::initialSolutionApplied() (eclproblem.hh:1482)
==45745==    by 0x17B4198: Ewoms::FvBaseDiscretization<Ewoms::Properties::TTag::EclFlowProblem>::applyInitialSolution() (fvbasediscretization.hh:599)
==45745==    by 0x179C0CD: Opm::FlowMainEbos<Ewoms::Properties::TTag::EclFlowProblem>::setupEbosSimulator() (FlowMainEbos.hpp:458)
==45745==    by 0x178EE3E: Opm::FlowMainEbos<Ewoms::Properties::TTag::EclFlowProblem>::execute(int, char**) (FlowMainEbos.hpp:228)
==45745==    by 0x17427B4: Opm::flowEbosBlackoilMain(int, char**) (flow_ebos_blackoil.cpp:62)
==45745==    by 0x1640A1D: main (flow.cpp:280)
```

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -2447,10 +2447,10 @@ private:
                                    timeMap.getTimeStepLength(episodeIdx));
         simulator.setEpisodeIndex(episodeIdx);
 
+        eclWriter_->beginRestart();
+
         Scalar dt = std::min(eclWriter_->restartTimeStepSize(), simulator.episodeLength());
         simulator.setTimeStepSize(dt);
-
-        eclWriter_->beginRestart();
 
         size_t numElems = this->model().numGridDof();
         initialFluidStates_.resize(numElems);


akva2:
jenkins build this please

andlaus:
does this work with stuff that requires the initial solution like, e.g. threshold pressures. If yes, does this apply to the E100 compatible restart files or only for extended ones?

akva2:
hmm? this is a logic error, it is wrong in all cases.

andlaus:
agreed. I thought that this was a fix for #1828, but it seems like it isn't?

Since Jenkins is happy, let's merge.

akva2:
not a fix for #1828 no, just something i found while looking into it.

PR #1832, opened 2019-05-10
updating well control during iterateWellEquations in MSW

GitPaean:
it enabled running through a real case with MSW, it should not break anything else. 

It was from https://github.com/OPM/opm-simulators/pull/1792, while I was hoping to improve it to be better form, while we need to lower the priority of it now.


GitPaean:
jenkins build this please

atgeirr:
As this allows some more cases to run through without doing no harm I'll merge. Thanks!

PR #1837, opened 2019-05-13
add a summary state member in eclproblem

totto82:
This just adds a empty container for summary values to be  
used by udq and actionx

To be filled later see #1831

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -44,6 +44,8 @@
 #include <opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
 #include <opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/SummaryState.hpp>
+
 
 #if HAVE_MPI
 #include <mpi.h>
@@ -379,6 +381,21 @@ public:
     static void setExternalSummaryConfig(Opm::SummaryConfig* summaryConfig)
     { externalEclSummaryConfig_ = summaryConfig; }
 
+
+    /*!
+    * \brief Returns the summary state
+    *
+    * The summary state is a small container object for
+    * computed, ready to use summary values. The values will typically be used by
+    * the UDQ, WTEST and ACTIONX calculations.
+    */
+    const Opm::SummaryState& summaryState() const
+    { return summaryState_; }
+
+    Opm::SummaryState& summaryState()
+    { return summaryState_; }
+
+
     /*!
      * \brief Returns the name of the case.
      *
@@ -515,6 +532,9 @@ private:
 
     Opm::Schedule* eclSchedule_;
     Opm::SummaryConfig* eclSummaryConfig_;
+
+    Opm::SummaryState summaryState_;
+
 };
 
 template <class TypeTag>


andlaus:
I believe it is better to put this into the vanguard (i.e., `EclBaseVanguard`) which already manages the other parser objects!?

joakim-hove:
> I believe it is better to put this into the vanguard (i.e., EclBaseVanguard) which already manages the other parser objects!?

I do not hold strong opinions on that - but observe that this is quite far from what I would call a *parser object* - although the implementation and main use is in opm-common.

totto82:
> I believe it is better to put this into the vanguard (i.e., EclBaseVanguard) which already manages the other parser objects!?

Good point. I will move it there.

totto82:
jenkins build this please

ebos/eclbasevanguard.hh
```diff
@@ -513,8 +533,11 @@ private:
     Opm::Deck* deck_;
     Opm::EclipseState* eclState_;
 
-    Opm::Schedule* eclSchedule_;
```
andlaus: nit: trailing spaces

ebos/eclbasevanguard.hh
```diff
@@ -193,6 +195,8 @@ public:
      */
     EclBaseVanguard(Simulator& simulator)
         : ParentType(simulator)
+        , summaryState_()
```
andlaus: nit: if you don't need to explicitly call the default constructor of an attribute.

ebos/eclbasevanguard.hh
```diff
@@ -379,6 +383,22 @@ public:
...
     { externalEclSummaryConfig_ = summaryConfig; }
 
+
+    /*!
+    * \brief Returns the summary state
+    *
+    * The summary state is a small container object for
+    * computed, ready to use summary values. The values will typically be used by
+    * the UDQ, WTEST and ACTIONX calculations. Observe that all value *have been
+    * converted to the correct output units*.
```
andlaus: I think that the comment about the units is better put to the documentation of the `Opm::SummaryState` class within `opm-common`. out of curiosity: does this mean that the `summaryState_` contains its values in FIELD units for field decks? If so, this has *a lot* of potential for confusion...

joakim-hove: > does this mean that the summaryState_ contains its values in FIELD units for field decks? 

Yes.

> If so, this has a lot of potential for confusion...

Hopefully you are wrong .... the object in question is a *state* object; i.e. the state of it is updated during the course of the simulation. It therefor needs to have the same lifetime as the simulator, but all *use* of it is confined to opm-common code; so the ugly detail about units should not really leak out.

andlaus: I also hope that I'm wrong. The reason why I fear that I'm not is that the SI-based simulator needs to fill this object, and the output code also must handle these quantities differently than the normal SI-based ones. IMO it would be thus good if the unit conversion was  done internally by this class, i.e. something like `summaryState_.setQuantitySi("FOO", index, value)`.`summaryState_.setQuantityDeckUnits("FOO", index, value)` plus the corresponding getters.

totto82: This comment is basically a copy from the class documentation. I can remove the stuff about the units Since the usage is internal in opm-common I guess the units are taken care of outside the simulator.

joakim-hove: > is that the SI-based simulator needs to fill this object, 

No - this is already handled in the output layer, and will continue to be so. This class is not new - we have used for ~one year.

andlaus: > No - this is already handled in the output layer

if the simulator does not need to interact with it what is the point in increasing its entropy by directly exposing it to the simulator instead of keeping it within `EclipseState`, then?

joakim-hove: > if the simulator does not need to interact with it what is the point in increasing its entropy by directly exposing it to the simulator instead of keeping it within EclipseState, then?

Well - the `EclipseState` object is immutable (or at least close ...), whereas the `SummaryState` object evolves with the simulation. It is *not* something generated from the deck - as the `EclipseState` is.

andlaus: In the long term, the current approach is IMO a recipe for maintenance hell -- or at least for considerable annoyance -- because each and every of the gazillion dynamic parser objects needs to be handled by the simulator and the simulator is not in that business IMHO. Rather, the schedule-dependent objects should be collected by a single class (`DynamicEclipseState`?) which ideally should also provide a pointer to the immutable data like the deck and the eclState that is created from it.

joakim-hove: > because each and every of the gazillion dynamic parser objects 

Gazillion(?) - such exaggerations makes a meaningful discussion difficult, please refrain from it. We have the following objects implemented and instantiated in the opm-common repository:

**`Deck`**: The first internalization of the input file - should ideally not be used outside of opm-common.

**`EclipseState`**: Everything static assembled from the deck input.

**`Schedule`**: Everything dynamic from the deck input.

**`SummaryConfig`**: An object internalizing all the summary variables requested in the deck.

These objects are instantiated (roughly) as such:
```
Deck <- input_file
EclipseState <- Deck
Schedule <- Deck, EclipseState
SummaryConfig <- Deck, EclipseState, Schedule
```
the `SummaryConfig` class is *a small detail* - and it unfortunate that it is a class that must be managed at top level by the simulator, that is because it requires the `Schedule`object for construction[1]. So currently we have `Gazillion == 4`. 

This PR adds a new[2] class to the simulator. The class in question is implemented, and used, in code which is currently located in opm-common, but it is *not* instantiated there - it does not use any input from the deck, and is not derived from any of the classes in opm-common. Since it needs to have application lifetime, and is a dynamic quantity - it makes sense to manage it from the simulator.

> Rather, the schedule-dependent objects should be collected by a single class 

The only `Schedule` dependent object - apart from `Schedule` itself is the `SummaryConfig`object.

> [....] which ideally should also provide a pointer to the immutable data like the deck and the eclState that is created from it.

I guess you mean "that it is created from"? Yes - that would be possible, but I would consider that an unfortunate coupling in the code.


[1]: Come to think of it - it *could* be internalized in the `Schedule` object - does not really belong there, but that way the object count would go down with one.

[2]: As I have already pointed out the class `SummaryState`is *not* new - it has been used internally in the output code for ~one year.

andlaus: there is also at least the `parseContext` and the `errorGuard` which the simulators need to wrangle with. Anyway, whether you declare a "gazillion" as `> 4` or not is probably a matter of taste: if I'm strict, I'd define "gazillion" as `> 1` in this context.

Anyway, can we please at least agree upon the fact that the number of these objects has only been growing over the last years and that it probably will continue to do so. (That said, I promise to shut up about the current state if you solemnly swear that no new such objects will be introduced after `summaryState` anymore.) 

(note that maybe our main dissonance stems from the fact that you seem to consider the parser as an end to itself, while I see it more as a necessary evil to achieve another goal, i.e., my view of this code is that the file I/O libraries are good if and only if they do not get in the way.)

totto82: Doesn't introducing concepts like ACTIONX and UDQ already violate the strict decoupling between the simulator and I/O. That said, I don't have any strong opinion on where to store the SummaryState object, as long it doesn't  involve too much extra work.

joakim-hove: > Anyway, can we please at least agree upon the fact that the number of these objects has only been growing over the last years 

Yes - that is true; along with `flow` becoming more and more capable.

> [...] . (That said, I promise to shut up about the current state if you solemnly swear that no new such objects will be introduced after summaryState anymore.)

Of course I can not swear to anything like that, but these objects are not introduced for fun - or to piss you off. It is done out of necessity - and in the way I think is the best way - the latter can of course be discussed.

joakim-hove: > that the file I/O libraries are good if and only if they do not get in the way.)

You are of course entitled to that opinion, but without IO there would not be much point in simulating.

andlaus: that's a quite petroleum-centric view, cf. all other ewoms tests/simulators. Also, even for petroleum engineering there would be little point in the I/O code without a simulator. (Seems like we're stuck ;))

andlaus: > [...] but these objects are not introduced for fun - or to piss you off.

I did not imply that. I said that the current design is flawed: these objects should be collected by a class that is tied to the I/O code instead of the simulator. If this class occasionally needs some new data from the simulator, so be it, but currently these objects are kept around at the simulator level for almost no other purpose than to keep the I/O code happy...

joakim-hove:
jenkins build this please

ebos/eclbasevanguard.hh
atgeirr: I actually agree, the core numerical code should not store these objects (EclipseState etc.) at all, they should be created at main() level or equivalent, and passed to the simulator code as necessary. But that design was left behind with the legacy simulator...

Given that we do not at this point refactor this design completely, @totto82's current proposal is fine.

joakim-hove: > they should be created at main() level or equivalent, and passed to the simulator code as necessary

I fully agree; the point is that currently `eclproblem/vanguard/simulator/...` serve as a de-facto main - that is is the *really broken design* around here.

atgeirr:
Since the nits were addressed and Jenkins is green I'll merge. Thanks!

ebos/eclbasevanguard.hh
andlaus: that was not my point, but well. as usual: change the design, I'm happy with anything as long as it only affects flow and nothing else...

PR #1839, opened 2019-05-13
Fix episode index for restart

totto82:
This restores restart for a model where not every episode is stored. 

in  `void readEclRestartSolution_() `
the episodeidx is set `initconfig.getRestartStep() - 1;`
while the output writer in opm-common expects the episode index to be equal to   `initconfig.getRestartStep()`

A better fix may be to pass the episodeidx from initconfig directly?
Any comment @andlaus?

### DIFF
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -266,10 +266,16 @@ public:
         std::vector<Opm::RestartKey> extraKeys = {{"OPMEXTRA", Opm::UnitSystem::measure::identity, false},
                                                   {"THRESHPR", Opm::UnitSystem::measure::pressure, inputThpres.active()}};
 
-        unsigned episodeIdx = simulator_.episodeIndex();
+        // The episodeIndex is rewined one back before beginRestart is called
+        // and can not be used here.
+        // We just ask the initconfig directly to be sure that we use the correct
+        // index.
+        const auto& initconfig = simulator_.vanguard().eclState().getInitConfig();
+        int restartStepIdx = initconfig.getRestartStep();
+
         const auto& gridView = simulator_.vanguard().gridView();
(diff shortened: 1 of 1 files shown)


andlaus:
seems fine. I trust that you have tested that this works?

andlaus:
jenkins build this please

ebos/eclwriter.hh
```diff
@@ -266,7 +266,7 @@ public:
         std::vector<Opm::RestartKey> extraKeys = {{"OPMEXTRA", Opm::UnitSystem::measure::identity, false},
                                                   {"THRESHPR", Opm::UnitSystem::measure::pressure, inputThpres.active()}};
 
-        unsigned episodeIdx = simulator_.episodeIndex();
+        unsigned episodeIdx = simulator_.episodeIndex()+1;
```
andlaus: the variable should probably renamed to something akin `reportStepNum` to avoid confusion...

totto82: Ok. I will grap it directly from the initconfig as well to avoid problems.

totto82:
jenkins build this please

GitPaean:
> jenkins build this please

jenkins is not working at the moment following @akva2 's notice.

totto82:
jenkins build this please

totto82:
This fixes a critical bug and should be merged asap in my opinion

andlaus:
thanks for the patch.

GitPaean:
It is good that we have a test to protect this. Otherwise, it might get broken again without anyone knowing it.  For example, we did not know when it was broken until we need this functionality for development. 

I guess it needs a little non-trivial `RPTRST` setup, something like 

`FREQ=6`, to make sure we only have restart output for some report steps not all the report steps. 

It is just a guess.

PR #1842, opened 2019-05-13
If requested always print version and exit.

blattms:
Previously this was only done if there were no other parameters on the
command line. That approach feels a bit unusual.

PR #1845, opened 2019-05-15
Abort if flow is run with unknown keywords.

blattms:
This change will print the encountered unknown keywords to std::cerr (albeit in the format ThisKeyword and not --this-keyword) and abort before running the simulator.

This is an alternative to #1842 that uses existing ewoms functionality. Namely, `Parameters::printUnused()` which returns a bool (Note: a print function that returns a value...)


blattms:
jenkins build this please

atgeirr:
I'll merge this now, it's useful straight away to many and is green.

PR #1833, opened 2019-05-10
run compareECL once, checking all files in one go

tskille:
update of script `run-regressionTest.sh`. Should be merged after PR OPM/opm-common#765. This PR will enable testing of RFT files if these are found in opm-tests reference folder.
 
Has been tested locally by manually copying RFT files from opm-simulators/build/tests/results folders
```
./flow+msw_model_1/MSW_MODEL_1.RFT   => opm-tests/model1/opm-simulation-reference/flow/MSW_MODEL_1.RFT
./flow+base_model_1/BASE_MODEL_1.RFT  => opm-tests/model1/opm-simulation-reference/flow/BASE_MODEL_1.RFT  
.flow+faults_model_1/FAULTS_MODEL_1.RFT  => opm-tests/model1/opm-simulation-reference/flow/FAULTS_MODEL_1.RFT
```

PR #1846, opened 2019-05-15
Update Data: Make Protocol Aware of RFT Files

bska:
If a simulation run intentionally changes an RFT file, then that change should be reflected in the new reference data.

This is in preparation of merging PRs OPM/opm-common#765 and #1833.

### DIFF
--- a/tests/update_reference_data.sh
+++ b/tests/update_reference_data.sh
@@ -67,7 +67,7 @@ do
       $configuration/build-opm-simulators/tests/results/$binary+$tname/ \
       $OPM_TESTS_ROOT/$dirname/opm-simulation-reference/$binary \
       $casename \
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska:
I intend to merge this to start the process of merging OPM/opm-common#765 and #1833.  Should I hold until the currently running package builds on the CI systems have completed?

akva2:
no need to wait, just go ahead and use jenkins. i'm just double-checking that the my last cleanup got rid of the segfault causing malware but this should not interfere with normal operations.

bska:
> no need to wait, just go ahead and use jenkins.

:+1: 

> i'm just double-checking that the my last cleanup got rid of the segfault causing malware but this should not interfere with normal operations.

Much appreciated!

PR #1847, opened 2019-05-15
Let only master process print information about exception.

blattms:
Previously, each process did this in a parallel run which clutters the terminal output quite a bit in a parallel run.


atgeirr:

atgeirr:
Thanks, merging.

PR #1848, opened 2019-05-16
Write transmissibility between direct vertical neighbors into TRANZ.

blattms:
In the case where two were direct vertical neighbors in the grid but not in the underlying cartesian grid (e.g. because of MINPV or pinch outs), we treated them as NNCs and wrote the transmissibilty
to TRANNC.

With this patch we detect this situation (two neighbor cells with identical i and j and no active cells between them) and do not create an NNC in the eclipse output files but write the transmissibility to TRANZ.

### DIFF
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -43,6 +43,7 @@
 #include <opm/parser/eclipse/Units/UnitSystem.hpp>
 
 #include <opm/grid/GridHelpers.hpp>
+#include <opm/grid/utility/cartesianToCompressed.hpp>
 
 #include <opm/material/common/Valgrind.hpp>
 #include <opm/material/common/Exceptions.hpp>
@@ -68,6 +69,55 @@ class EclWriter;
 
 template <class TypeTag>
 class EclOutputBlackOilModule;
+
+/*!
+ * \brief Detect whether two cells are direct vertical neighbours.
+ *
+ * I.e. have the same i and j index and all cartesian cells between them
+ * along the vertical column are inactive.
+ *
+ * \tparam CM The type of the cartesian index mapper.
+ * \param cartMapper The mapper onto cartesian indices.
+ * \param cartesianToActive The mapping of cartesian indices to active indices.
+ * \param smallGlobalIndex The cartesian cell index of the cell with smaller index
+ * \param largeGlobalIndex The cartesian cell index of the cell with larger index
+ * \return True if the cells have the same i and j indices and all cartesian cells
+ *         between them are inactive.
+ */
+inline
+bool directVerticalNeighbors(const std::array<int, 3>& cartDims,
+                             const std::unordered_map<int,int>& cartesianToActive,
+                             int smallGlobalIndex, int largeGlobalIndex)
+{
+    assert(smallGlobalIndex <= largeGlobalIndex);
+    std::array<int, 3> ijk1, ijk2;
+    auto globalToIjk = [cartDims](int gc) {
+                           std::array<int, 3> ijk;
+                           ijk[0] = gc % cartDims[0];
+                           gc /= cartDims[0];
+                           ijk[1] = gc % cartDims[1];
+                           ijk[2] = gc / cartDims[1];
+                           return ijk;
+                       };
+    ijk1 = globalToIjk(smallGlobalIndex);
+    ijk2 = globalToIjk(largeGlobalIndex);
+    assert(ijk2[2]>=ijk1[2]);
+
+    if ( ijk1[0] == ijk2[0] && ijk1[1] == ijk2[1] && (ijk2[2] - ijk1[2]) > 1)
+    {
+        assert((largeGlobalIndex-smallGlobalIndex)%(cartDims[0]*cartDims[1])==0);
+        for ( int gi = smallGlobalIndex + cartDims[0] * cartDims[1]; gi < largeGlobalIndex;
+              gi += cartDims[0] * cartDims[1] )
+        {
+            if ( cartesianToActive.find( gi ) != cartesianToActive.end() )
+            {
+                return false;
+            }
+        }
+        return true;
+    } else
+        return false;
+}
 
 /*!
  * \ingroup EclBlackOilSimulator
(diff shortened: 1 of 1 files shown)


blattms:
this continues the work of #1818. Now there is only NNC that Eclipse outputs but not flow does not. That is `(14,16,07) -> (13,16,09)`.

I know thank jenkins will report error. But I abuse it to double check how TRANZ changes,

blattms:
jenkins build this please

blattms:
Mmh, this is a bit weired. The NNCs are not written anymore, but it seems like the transmissibilities written
to TRANZ appear in the wrong places. I will need to investigate this a bit more.

alfbr:
Along with this change, I believe it is also necessary to write out PINCH-OUT connections. Is this included in here? If not, is there anything I can do to assist?

blattms:
What do you mean by "write out PINCH-OUT connections", is there a special eclipse keyword?

With change the follwing holds: If cell (i,j,k) is active, (i,j,k+1) gets pinched out and
(i,j,k+2) is active, then TRANZ for (i,j,k) will store the transmissibility between (i,j,k)
and (i,j,k+2). That is what eclipse seems to do.

alfbr:
> What do you mean by "write out PINCH-OUT connections", is there a special eclipse keyword?

I am not aware of any special keyword. Actually, it seems the PINCH-OUT connections are not written out to any result file. The only output is through the PRT file and LOG/STDOUT, where the number of PINCH-OUT connections are reported.

blattms:
Well the information about pinch out is encoded into the INIT and EGRID file. If you have a nonzero TRANZ entry where the cartesian neighbour in k direction is not active then there is a pinch out. python code is
```python
#!/usr/bin/env python
import sys
from operator import itemgetter
from ecl.eclfile import EclFile
from ecl.grid import EclGrid

if __name__ == "__main__":
    case = sys.argv[1]
    init_file = EclFile("%s.INIT" % case)
    grid = EclGrid("%s.EGRID" % case)

    tranz = init_file["TRANZ"][0]
    dims = grid.get_dims()

    for index, t in zip(range(0,len(tranz)), tranz):
        i1,j1,k1 = grid.get_ijk( active_index = index )
        gi = grid.get_global_index(active_index=index)
        kdiff = 2
        if k1 + kdiff < dims[2] and not grid.active(ijk=(i1, j1, k1+1)) and t!=0.0:
            while not grid.active(ijk=(i1, j1, k1+ kdiff)):
                kdiff = kdiff + 1
            print "tranz has entry for (%02d,%02d,%02d) -> (%02d,%02d,%02d)" %(i1,j1,k1,i1,j1,k1+kdiff)
```

And yes eclipse actually does store it exactly like this and not as an NNC as previously assumed.

blattms:
After fighting with TRANZ being stored as a compressed vector in the INIT file (kudos to @joakim-hove for the enlightenment), I have now double checked that the PR does what it should and that Eclipse actually does it the same way.

blattms:

blattms:
jenkins build this please

ebos/eclwriter.hh
```diff
@@ -69,6 +70,58 @@ class EclWriter;
...
+    if ( ijk1[0] == ijk2[0] && ijk1[1] == ijk2[1] && (ijk2[2] - ijk1[2]) > 1)
+    {
+        assert((largeGlobalIndex-smallGlobalIndex)%(cartDims[0]*cartDims[1])==0);
+        for ( int gi = smallGlobalIndex + cartDims[0] * cartDims[1]; gi < largeGlobalIndex;
+              gi += cartDims[0] * cartDims[1] )
+        {
+            auto found = cartesianToActive.find( gi );
+            if ( found != cartesianToActive.end() )
+            {
+                std::array<int, 3> ijk = globalToIjk(found->second);
```
bska: The `ijk` variable appears to be unused.  Is that intentional?

bska:
I have a general question: How close to completion is this PR? Just final review missing, or do you foresee any additional work?

The reason I ask is that due to the previous [data update request](https://github.com/OPM/opm-simulators/pull/1848#issuecomment-494127675), this PR now effectively conflicts with OPM/opm-tests#118 which is needed for OPM/opm-common#765 which in turn is blocking OPM/opm-common#763. I'd like to get the latter into master to proceed with the task of replacing the `libECLRestart` component in opm-common. On the other hand, merging that along with its associate data update will create problems for this PR (will at minimum require re-updating OPM/opm-tests#119, although that might not be sufficient) so if this is close to final we should merge this first.

ebos/eclwriter.hh
blattms: Nope, that was some left over from debugging. I have amended the commit to fix this.

blattms:
IMHO this is finished and tested.

blattms:
jenkins build this opm-test=119 please

bska:
> [...] opm-test=119 [...]

This needs to be -test**S**.  I learned that the hard way.

bska:
jenkins build this opm-tests=119 please

blattms:
Thanks.

ebos/eclwriter.hh
```diff
@@ -361,16 +414,23 @@ private:
...
 
-                if (gc2 - gc1 == cartDims[0])
+                if (gc2 - gc1 == cartDims[0]) {
                     trany.data[gc1] = globalTrans->transmissibility(c1, c2);
+                    continue; // skipt next if clause as it needs some computation
+                }
 
-                if (gc2 - gc1 == cartDims[0]*cartDims[1])
+                if ( gc2 - gc1 == cartDims[0]*cartDims[1] ||
+                     directVerticalNeighbors(Dune::CartesianIndexMapper<Grid>(globalGrid_), cartesianToActive, gc1, gc2))
```
bska: There's one possible optimisation here.  You're forming a new `CartesianIndexMapper`, but the implementation of the `directVerticalNeighbors` function "only" requires the global Cartesian dimensions.  Could you not pass a reference to the [`cartDims`](https://github.com/OPM/opm-simulators/blob/4d31f116051b39ed4f8e9bd0b3ea9974b71668f5/ebos/eclwriter.hh#L364) object that already exists instead of a full mapper object?

blattms: Indeed, Nice catch. I amended the commit accordingly.

blattms:
jenkins build this opm-tests=119 please

bska:
Merging this into master now.  Thanks a lot!

PR #1855, opened 2019-05-22
throw runtime_error instead of logic_error

totto82:
With this change we throw something we actually catch.


GitPaean:
I agree. If the type of exception is missing, `runtime_error` is more sensible than `logic_error`.

GitPaean:
jenkins build this please

PR #1861, opened 2019-05-27
Fix FPRP output

totto82:
Currectly output FPRP values, instead of overwriting the FPR values.


totto82:
jenkins build this please

totto82:
jenkins build this please

PR #1862, opened 2019-05-27
adding function updateUpwindingSegments() for MSW

GitPaean:
to update the upwinding segment for each segment based on flow direction

This is a refactoring efforts to simplify the upwinding treatment.  It does not include the extension of the upwinding to more items. It should not change any results in any way. 

However, for more complicated cases, the iteration behavoir still changes after half-way running for unclear reason.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -44,6 +44,7 @@ namespace Opm
     , segment_viscosities_(numberOfSegments(), 0.0)
     , segment_mass_rates_(numberOfSegments(), 0.0)
     , segment_depth_diffs_(numberOfSegments(), 0.0)
+    , upwinding_segments_(numberOfSegments(), 0)
     {
         // not handling solvent or polymer for now with multisegment well
         if (has_solvent) {
@@ -1303,24 +1304,9 @@ namespace Opm
     typename MultisegmentWell<TypeTag>::EvalWell
     MultisegmentWell<TypeTag>::
     getSegmentRateUpwinding(const int seg,
-                            const int comp_idx,
-                            const bool upwinding,
-                            int& seg_upwind) const
+                            const int comp_idx) const
     {
-        // not considering upwinding for the injectors for now
-        if ((!upwinding) || (well_type_ == INJECTOR) || (primary_variables_evaluation_[seg][GTotal] <= 0.) || (seg == 0)) {
-            seg_upwind = seg; // using the composition from the seg
-            return primary_variables_evaluation_[seg][GTotal] * volumeFractionScaled(seg, comp_idx);
-        }
-
-        // assert( seg != 0); // if top segment flowing towards the wrong direction, we are not handling it
-
-        // basically here, it a producer and flow is in the injecting direction
-        // we will use the compsotion from the outlet segment
-        const int outlet_segment_index = segmentNumberToIndex(segmentSet()[seg].outletSegment());
-        seg_upwind = outlet_segment_index;
-
-        // TODO: we can refactor above code to use the following return
+        const int seg_upwind = upwinding_segments_[seg];
         return primary_variables_evaluation_[seg][GTotal] * volumeFractionScaled(seg_upwind, comp_idx);
     }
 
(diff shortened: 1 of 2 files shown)


opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1303,24 +1304,9 @@ namespace Opm
     typename MultisegmentWell<TypeTag>::EvalWell
     MultisegmentWell<TypeTag>::
     getSegmentRateUpwinding(const int seg,
-                            const int comp_idx,
-                            const bool upwinding,
-                            int& seg_upwind) const
+                            const int comp_idx) const
     {
-        // not considering upwinding for the injectors for now
-        if ((!upwinding) || (well_type_ == INJECTOR) || (primary_variables_evaluation_[seg][GTotal] <= 0.) || (seg == 0)) {
```
GitPaean: The only explanation is `primary_variables_evaluation_` and `primary_variables_` do not have the same values under some circumstances.

GitPaean: There should be an `initPrimaryVariablesEvaluation()` after https://github.com/OPM/opm-simulators/blob/c554fdb5fb016406b5540c69def2fe474e349d12/opm/simulators/wells/BlackoilWellModel_impl.hpp#L948 .

GitPaean: Basically, in the master branch, when solveWellEq failed, it tries to recover the old primary variables. But it does not recover the evaluation version of the primary variables, which are the ones that work when assembling the well equations, so the recovering is not very complete. And some inconsistency is introduced there. 

In the function `assemble`, 
https://github.com/OPM/opm-simulators/blob/c554fdb5fb016406b5540c69def2fe474e349d12/opm/simulators/wells/BlackoilWellModel_impl.hpp#L668

When solveWellEq does not get converged for the first time, maybe we should not enter the second solveWellEq, since the well_state is still the old one, not improved by the first solveWellEq at all.

GitPaean:
jenkins build this please

GitPaean:
This PR changed back to use `primary_variables_evaluation_` for the upwinding determination. 

So it should not change result at all with its latest form. 

The original version reveals that during solveWellEq, there might be some inconsistency when recovering from failing in getting converged in solveWellEq. However,  I do not have an answer about what is the best way to handle this situation, so I decided not to change that part within this PR. 

This PR simplifies some repeated calculations related to upwinding. It was part of the efforts to extend upwinding to all the items that are involved in the MSW well equations, while it does not work well. More careful checking about formulations remains a continuous efforts for later. 

As for now, this PR should not change any results only serving for a cleaning up for previous upwidning implementation in MSW, which hopefully makes the testing and review this PR easier.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1303,24 +1304,9 @@ namespace Opm
...
-
-        // basically here, it a producer and flow is in the injecting direction
-        // we will use the compsotion from the outlet segment
-        const int outlet_segment_index = segmentNumberToIndex(segmentSet()[seg].outletSegment());
-        seg_upwind = outlet_segment_index;
-
-        // TODO: we can refactor above code to use the following return
+        const int seg_upwind = upwinding_segments_[seg];
         return primary_variables_evaluation_[seg][GTotal] * volumeFractionScaled(seg_upwind, comp_idx);
     }
```
totto82: Maybe I am wrong, but is this correct? You are multiplying derivatives of the primary variables wrt to the segment "seg" with ones wrt. to the segment "seg_upwind"  I have not looked very carefully at the code so I may have missed something.

GitPaean: Yes. We use the composition from the upwinding direction. We have some cases in the wellbore of MSW, the flow can go different directions (some flux go up, some flux go down).  Without upwinding, the flux will be with wrong composition, then the mass of component will not be conserved.

GitPaean:
It is confirmed this PR does not change results in anyway compared with master branch. 

And the bug revealed from this PR will be solved when we have a solution for it.

PR #1865, opened 2019-05-29
Use shared summarystate

joakim-hove:
When this is ready and merged the simulator side of UDQ and ACTIONX is complete&trade;

PR #1867, opened 2019-05-30
not trying to update RESV target for prediction producers

### LINKED ISSUE #1866: A bug for assuming RESV can only be history matching wells
If WCONPROD is specified with RESV control,  the following assertion failure will be triggered. 

```
OPM-devel/debug/opm-simulators/opm/autodiff/SimFIBODetails.hpp:91: void Opm::SimFIBODetails::historyRates(const Opm::PhaseUsage&, const Opm::ProductionControls&, std::vector<double>&): Assertion `! p.prediction_mode' failed.
```

It should not be difficult to fix anyway. 

Possibly other related bugs can be found related to injectors.

GitPaean:
this closes https://github.com/OPM/opm-simulators/issues/1866

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1695,6 +1695,7 @@ namespace Opm {
                         well_controls_iset_distr(ctrl, rctrl, & distr[0]);
 
                         // for the WCONHIST wells, we need to calculate the RESV rates since it can not be specified directly
+                        // for the WCONPROD wells, the rates are specified already, it is not necessary to update
                         if (is_producer) {
                             const WellMap::const_iterator i = wmap.find(wells()->name[*rp]);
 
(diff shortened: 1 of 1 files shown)


opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1703,21 +1704,23 @@ namespace Opm {
                             }
                             const auto& wp = i->second;
                             const auto production_controls = wp.productionControls(summaryState);
```
GitPaean: Not sure why `summaryState` is here. It is possible something WIP.

GitPaean:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
bska: > Not sure why `summaryState` is here. It is possible something WIP.

It was introduced in commit 511645d12, as part of PR #1831. It's work in progress in the sense that the ACTIONX/UDQ feature is as yet incomplete, but the simulator side of the work is apparently very close to completion (see PR #1865).

bska (approved):
No real comments. This addresses the immediate problem. On the other hand the `computeRESV` function is becoming very complicated (many nested levels of conditionals) and might be candidate for refactoring at some point.

opm/simulators/wells/BlackoilWellModel_impl.hpp
GitPaean: Thanks for the explanation.

PR #1869, opened 2019-05-31
adding some comments about getSegmentRateUpwinding()

GitPaean:
Some comments are added as requested and a variable renaming for better readability. This should not change any result in any sense.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1391,6 +1391,9 @@ namespace Opm
                             const int comp_idx) const
     {
         const int seg_upwind = upwinding_segments_[seg];
+        // the result will contain the derivative against GTotal in segment seg,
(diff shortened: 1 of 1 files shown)


GitPaean:
jenkins build this please

GitPaean:
Added an assertion to make sure the pressure derivative in getSegmentRateUpwinding() is zero.  Otherwise, the assumption fails.

GitPaean:
jenkins build this please

bska:
Looks okay, just one question: Does "derivative against" mean "derivative with respect to"?

GitPaean:
> Does "derivative against" mean "derivative with respect to"?

Yes. I will change it since it is possibly easier to people to understand.

GitPaean:
jenkins build this please

atgeirr:
Thanks, this makes that code a lot easier to understand (and to avoid making mistakes). Merging.

PR #1870, opened 2019-06-03
Ebos cleanups

andlaus:
this PR contains a bunch of cleanups -- like typo and whitespace fixes, message improvements, etc -- for ebos. It also adds the possibility to disable aquifer handling in ebos, because I had some severe issues with them for a deck I was recently working on. none of them should have any effect on the simulation results of `flow`...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -259,6 +259,14 @@ opm_add_test(ebos_altidx
   DEPENDS "opmsimulators"
   LIBRARIES "opmsimulators")
 
+opm_add_test(ebos_plain
+  ONLY_COMPILE
+  DEFAULT_ENABLE_IF ${EBOS_DEBUG_EXTENSIONS_DEFAULT_ENABLE_IF}
+  SOURCES ebos/ebos_plain.cc
+  EXE_NAME ebos_plain
+  DEPENDS "opmsimulators"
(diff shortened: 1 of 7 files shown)


andlaus:
jenkins build this please

CMakeLists.txt
```diff
@@ -259,6 +259,14 @@ opm_add_test(ebos_altidx
...
   LIBRARIES "opmsimulators")
 
+opm_add_test(ebos_plain
+  ONLY_COMPILE
+  DEFAULT_ENABLE_IF ${EBOS_DEBUG_EXTENSIONS_DEFAULT_ENABLE_IF}
+  SOURCES ebos/ebos_plain.cc
+  EXE_NAME ebos_plain
+  DEPENDS "opmsimulators"
+  LIBRARIES "opmsimulators")
+
```
totto82: I assume this is a preparation for a ebos multiplexer binary since the ebos binary is kept.

totto82:
Thanks for the clean-up.

CMakeLists.txt
andlaus: Not really: I just noticed could only be build with the dune based BS, but not with the default one. (I exclusively use the former, except occasionally for testing.) Anyway, thanks for merging.

PR #1873, opened 2019-06-03
output .INIT and .GRID also on restart

totto82:


totto82:
jenkins build this please

PR #1874, opened 2019-06-04
Call Summary::eval() and ::add_timestep separately 

joakim-hove:
Followup to: https://github.com/OPM/opm-common/pull/795

Again a step in UDQ/ACTIONX sequel

PR #1877, opened 2019-06-10
Summary bcast

joakim-hove:
The last time I wrote MPI code we were searching in the internet with AllTheWeb and Infoseek; feedback welcome. 

As indicated my MPI skills are not razor sharp, and I would like some feedback on that, but apart from that I consider this to be roughly complete and ready for review/comments.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -438,9 +438,11 @@ public:
     * computed, ready to use summary values. The values will typically be used by
     * the UDQ, WTEST and ACTIONX calculations.
     */
-    Opm::SummaryState& summaryState() 
+    Opm::SummaryState& summaryState()
     { return summaryState_; }
 
+    const Opm::SummaryState& summaryState() const
+    { return summaryState_; }
     /*!
      * \brief Returns the name of the case.
      *
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -54,6 +54,10 @@
 #include <utility>
 #include <string>
 
+#ifdef HAVE_MPI
+#include <mpi.h>
+#endif
+
 BEGIN_PROPERTIES
 
 NEW_PROP_TAG(EnableEclOutput);
@@ -207,13 +211,109 @@ public:
     /*!
      * \brief collect and pass data and pass it to eclIO writer
      */
-    void writeOutput(bool isSubStep)
+
+    void evalSummaryState(bool isSubStep)
     {
+        int reportStepNum = simulator_.episodeIndex() + 1;
+        /*
+          The summary data is not evaluated for timestep 0, that is
+          implemented with a:
+
+             if (time_step == 0)
+                 return;
+
+          check somewhere in the summary code. When the summary code was
+          split in separate methods Summary::eval() and
+          Summary::add_timestep() it was necessary to pull this test out
+          here to ensure that the well and group related keywords in the
+          restart file, like XWEL and XGRP were "correct" also in the
+          initial report step.
+
+          "Correct" in this context means unchanged behavior, might very
+          well be more correct to actually remove this if test.
+        */
+        if (reportStepNum == 0)
+            return;
+
+        const auto& summary = eclIO_->summary();
         Scalar curTime = simulator_.time() + simulator_.timeStepSize();
         Scalar totalCpuTime =
             simulator_.executionTimer().realTimeElapsed() +
             simulator_.setupTimer().realTimeElapsed() +
             simulator_.vanguard().externalSetupTime();
+
+        Opm::data::Wells localWellData = simulator_.problem().wellModel().wellData();
+
+        const auto& gridView = simulator_.vanguard().gridView();
+        int numElements = gridView.size(/*codim=*/0);
+        bool log = collectToIORank_.isIORank();
+        eclOutputModule_.allocBuffers(numElements, reportStepNum, isSubStep, log);
+
+        ElementContext elemCtx(simulator_);
+        ElementIterator elemIt = gridView.template begin</*codim=*/0>();
+        const ElementIterator& elemEndIt = gridView.template end</*codim=*/0>();
+        for (; elemIt != elemEndIt; ++elemIt) {
+            const Element& elem = *elemIt;
+            elemCtx.updatePrimaryStencil(elem);
+            elemCtx.updatePrimaryIntensiveQuantities(/*timeIdx=*/0);
+            eclOutputModule_.processElement(elemCtx);
+        }
+
+        if (collectToIORank_.isParallel())
+            collectToIORank_.collect({}, eclOutputModule_.getBlockData(), localWellData);
+
+        std::map<std::string, double> miscSummaryData;
+        std::map<std::string, std::vector<double>> regionData;
+        eclOutputModule_.outputFipLog(miscSummaryData, regionData, isSubStep);
+
+        std::vector<char> buffer;
+        if (collectToIORank_.isIORank()) {
+            const auto& eclState = simulator_.vanguard().eclState();
+
+            // Add TCPU
+            if (totalCpuTime != 0.0)
+                miscSummaryData["TCPU"] = totalCpuTime;
+
+            const Opm::data::Wells& wellData = collectToIORank_.isParallel() ? collectToIORank_.globalWellData() : localWellData;
+
+            const std::map<std::pair<std::string, int>, double>& blockData
+                = collectToIORank_.isParallel()
+                ? collectToIORank_.globalBlockData()
+                : eclOutputModule_.getBlockData();
+
+            summary.eval(summaryState(),
+                         reportStepNum,
+                         curTime,
+                         eclState,
+                         schedule(),
+                         wellData,
+                         miscSummaryData,
+                         regionData,
+                         blockData);
+            buffer = summaryState().serialize();
+        }
+
+        if (collectToIORank_.isParallel()) {
+            unsigned long buffer_size = buffer.size();
+
+#ifdef HAVE_MPI
+            MPI_Bcast(&buffer_size, 1, MPI_UNSIGNED_LONG, collectToIORank_.ioRank, MPI_COMM_WORLD);
+            if (!collectToIORank_.isIORank())
+                buffer.resize( buffer_size );
+
+            MPI_Bcast(buffer.data(), buffer_size, MPI_CHAR, collectToIORank_.ioRank, MPI_COMM_WORLD);
+            if (!collectToIORank_.isIORank()) {
+                Opm::SummaryState& st = summaryState();
+                st.deserialize(buffer);
+            }
+#endif
+        }
+    }
+
+
+    void writeOutput(bool isSubStep)
+    {
+        Scalar curTime = simulator_.time() + simulator_.timeStepSize();
         Scalar nextStepSize = simulator_.problem().nextTimeStepSize();
 
         // output using eclWriter if enabled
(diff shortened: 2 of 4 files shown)


joakim-hove:
jenkins build this please

atgeirr:
I'll try to look at it tomorrow. An aside: I just noticed the changed PR description accidentally, it is better to make a new message to get attention to the fact that it is now ready and should be reviewed. Changing a message does not trigger a new email.

joakim-hove:
> it is better to make a new message to get attention to the fact that it is now ready and should be reviewed. 

Good point :+1:

ebos/eclwriter.hh
```diff
@@ -588,7 +684,8 @@ private:
         // callback to eclIO serial writeTimeStep method
         void run()
         {
-            eclIO_.writeTimeStep(reportStepNum_,
+            eclIO_.writeTimeStep(summaryState,
```
atgeirr: This is the only serious defect I think in this code. I believe at this point you are passing (a reference to) the SummaryState object owned by the vanguard. That will change in the next call to writeOutput(), which may happen before this thread is done. To be safe, only constant objects may be passed by reference to EclWriteTasklet. The EclipseIO object itself is passed by reference, but I believe that is *only* changed inside this call, so the pipelineing of the tasklets prevents this from being a problem.

Either we must find a way to avoid passing the SummaryState, or a copy must be passed to the EclWriteTasklet. What is the state used for inside writeTimeStep().

ebos/eclwriter.hh
```diff
@@ -250,14 +257,77 @@ public:
...
             if (totalCpuTime != 0.0)
                 miscSummaryData["TCPU"] = totalCpuTime;
 
+            const Opm::data::Wells& wellData = collectToIORank_.isParallel() ? collectToIORank_.globalWellData() : localWellData;
+
+            const std::map<std::pair<std::string, int>, double>& blockData
+                = collectToIORank_.isParallel()
+                ? collectToIORank_.globalBlockData()
+                : eclOutputModule_.getBlockData();
+
```
atgeirr: This function is already long, so this part should be refactored into its own function (or multiple). I think a `broadcastSummaryState()` could hide the buffer and MPI calls from the rest of the function in a good way.

I love the long comment explaining the weirdness of the `if (reportStepNum > 0)` part, but I miss comments saying "what is going on here".

Nitpick: please avoid double blank lines inside functions. (Especially in files which have only a single blank between functions.

ebos/eclwriter.hh
joakim-hove: Many good points - I am working on an update.

ebos/eclbasevanguard.hh
```diff
@@ -390,13 +390,9 @@ public:
     * computed, ready to use summary values. The values will typically be used by
     * the UDQ, WTEST and ACTIONX calculations.
     */
-    const Opm::SummaryState& summaryState() const
```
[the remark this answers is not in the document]
joakim-hove: Should probably keep it ..

ebos/eclwriter.hh
```diff
@@ -555,6 +641,9 @@ private:
     struct EclWriteTasklet
         : public TaskletInterface
     {
+        Opm::SummaryState& summaryState;
```
atgeirr: Apparently this code (which I think is suspect as described before) was already added in #1874, I have no idea why it looks new in this PR?!?

bska: > I have no idea why it looks new in this PR?!?

**Probably** a missing rebase.

joakim-hove:
OK; this has been significantly rewritten and should be ready for a new review. Observe that with UDQ/UDA variables the importance of the `SummaryState`class - and the instance owned by the ebos simulator has increased considerably.

Evaluating the `SummaryState` is still part of the ecl output logic, but when the same class/instance is also used to update well controls it is maybe a bit misleading to file it under *output*. I guess it feels slightly klunky that flow coexists with another simulator which is not aware/interested in eclipse idosyncracies?

I think the current PR should work&trade; and is reasonably clean, but I have opted to make the minimal amount of changes, as this is code I am not very familar with.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this opm-common=1888 please

ebos/eclwriter.hh
atgeirr: I appreciate the changes, in particular splitting the long function and moving the eval() out of the tasklet. However I still think this code is buggy (although not introduced in this PR as I thought) for the reasons outlined above?

joakim-hove: Ehhh - I might misunderand, bu I think you are wrong. The tasklet now takes a copy.

atgeirr: Line 664 says it's a reference. Where is the copy?

Also noticed new nitpick: we now have code mixing trailing underscores and no trailing underscores. Should be fixed.

blattms:
Well, I do not feel fit to review this as I do not even understand what  the functional changes are.

ebos/eclwriter.hh
joakim-hove: > we now have code mixing trailing underscores and no trailing underscores. Should be fixed.

OK - now I have been sitting on my hands for about three hours, and body temperatur has fallen below 38&deg; C and I am ready to start sprinkling the code with underscores.

joakim-hove: > Line 664 says it's a reference. Where is the copy?

You are right - it was only in my head I had changed that to a copy; *now* it is a copy though.

joakim-hove:
> I do not feel fit to review this as I do not even understand what the functional changes are.

So much for self documenting code .... The changes are related to the shared `SummaryState`instance owned by the simulator: Initially the `SummaryState` class was only used as a temporary container for the summary variables like `WWCT:OPX` and `FOPT` which are calculated by the `Summary` class in the output code. With the advent of UDA variables the `SummaryState` is gaining more importance, and in particular it is used while initializing well rates.

The calculations are only performed by IO rank, but now when the `SummaryState`is used by the well initialization code, we need an updated (and globally correct) `SummaryState` instance on all PE's. So in summary the changes in this PR are:

1. The `Summary:eval()` and `Summary::add_timestep()` calls have been separated - only the latter happens in a separate IO thread.
2. The `Summary::eval()` function is only called on IO rank, but distributed globally afterwards.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

ebos/eclwriter.hh
atgeirr: > OK - now I have been sitting on my hands for about three hours - at a high hourly rate - body temperature has fallen below 38° C and I am ready to start sprinkling the code with underscores.

It was only a nitpick, it was not my intention to cause bodily harm... But I think you would agree that we should keep consistent within each class.

ebos/eclwriter.hh
```diff
@@ -207,13 +211,111 @@ public:
...
+        eclOutputModule_.allocBuffers(numElements, reportStepNum, isSubStep, log);
+
+        ElementContext elemCtx(simulator_);
+        ElementIterator elemIt = gridView.template begin</*codim=*/0>();
+        const ElementIterator& elemEndIt = gridView.template end</*codim=*/0>();
+        for (; elemIt != elemEndIt; ++elemIt) {
+            const Element& elem = *elemIt;
+            elemCtx.updatePrimaryStencil(elem);
+            elemCtx.updatePrimaryIntensiveQuantities(/*timeIdx=*/0);
+            eclOutputModule_.processElement(elemCtx);
```
atgeirr: I do not see a corresponding call and loop in the master code, why is this necessary?

ebos/eclwriter.hh
joakim-hove: > It was only a nitpick,

Well - either you write a comment and expect it to be adhered to, or you let it be.  

> it was not my intention to cause bodily harm... But I think you would agree that we should keep consistent within each class.

Of course I was trying to make a funny point; I did not sit on my hands while cooling off, instead I went pacing around the house. But seriously - and this you can quote me on: "The argument of consistency is consistently brought up when it aligns with the general preference of the person bringing it up. A bit like business leaders who argue for a *fair* tax regime - when what they really want is zero tax just for them." Of course you are taking more than your fair share of Sintef flak now; anyway underscores have been introduced on the `SummaryState` member.

ebos/eclwriter.hh
```diff
@@ -207,13 +211,111 @@ public:
...
+        eclOutputModule_.outputErrorLog();
+
+        if (collectToIORank_.isParallel())
+            collectToIORank_.collect({}, eclOutputModule_.getBlockData(), localWellData);
+
+        std::map<std::string, double> miscSummaryData;
+        std::map<std::string, std::vector<double>> regionData;
+        eclOutputModule_.outputFipLog(miscSummaryData, regionData, isSubStep);
+
+        // write output on I/O rank
```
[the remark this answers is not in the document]
joakim-hove: I'll look into this later - at least the call to `outputFipLog()` *should* be there, the function is used to assemble properties together for the summary evaluation, I take no blame for the name though.

ebos/eclwriter.hh
joakim-hove: I don't understand this code - but it is in master.

ebos/eclwriter.hh
akva2: don't claim this is a sintef thing! i work in sintef and i completely disagree. trailing underscores are evil eviil things. the only sane thing to do is to let members  have a m_ prefix :trollface:

joakim-hove: I don't know if you are serious now - and I will anyway not spam any further on the topic, but just for the record I consider the `m_` prefix high on the top ten list of my professional blunders. My current rage on the matter is using `this->` consistently #IalsoLikePython

ebos/eclwriter.hh
bska: I've spent some time in this area as part of the work on restart I/O.  Basically [`EclOutputBlackOilModule::processElement()`](https://github.com/OPM/opm-simulators/blob/e114d4a46d910ce0f18df6adff381668bbb1a2b3/ebos/ecloutputblackoilmodule.hh#L371-L375) extracts information that is maintained in the element context (e.g., cell pressure, saturations, gas/oil ratio, relative permeability) into linear buffers that are suitable for output to restart files (i.e., buffers that are used as backing data in subsequent calls to [`EclOutputBlackOilModule::assignToSolution()`](https://github.com/OPM/opm-simulators/blob/e114d4a46d910ce0f18df6adff381668bbb1a2b3/ebos/ecloutputblackoilmodule.hh#L779-L782) and [`EclOutputBlackOilModule::outputFipLog()`](https://github.com/OPM/opm-simulators/blob/e114d4a46d910ce0f18df6adff381668bbb1a2b3/ebos/ecloutputblackoilmodule.hh#L923-L924).)

joakim-hove: Thank you - yes that was my impression to, and since the summary output can potentially need this data for e.g. block pressure and fipxxx output I concluded - without checking very much, that it was needed even though the code in question does not output restart fields.

joakim-hove:
Hehe - this PR will soon get the :fire: label on GitHub overview!

ebos/eclwriter.hh
atgeirr: Ok thanks.

I do not understand how to read this diff on github though: There are green lines with a plus for this code block, but no corresponding red lines with minus showing it was moved (try to search in the browser for processElement for example). However when I check the master, it is clearly there early in the writeOutput method. So there was one code block before, there is one now (net change zero), yet GitHub seems to show addition of the block with no corresponding removal (net change one). How can that be?

joakim-hove: > but no corresponding red lines with minus showing it was moved

Well it was not moved, it was copied.

> So there was one code block before [...]

and now there is two - the original in a code block with no changes.

atgeirr: Thanks, I failed to spot the collapsed code part that contained the original block still. That brings up another question though: should it not be removed from its original spot?

joakim-hove: > That brings up another question though: should it not be removed from its original spot?

I saw that one coming; no it should not be removed from the original callsite. The point is that the code we are discussing now is used to serialize and broadcast the solution fields `PRESSURE, SWAT, ...` for output. The serialized solution fields are required both when evaluating the summary variables - the new callsite, and when storing restart files - the original callsite.

It might be possible to optimize a bit here, but I would like to merge "correct and working" first, and then possibly we can optimize a bit in the future.

ebos/eclwriter.hh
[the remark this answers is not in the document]
joakim-hove: OK - I have removed the call to `outputErrorLog()` and the misleading comment.

joakim-hove:
OK - this is to the best of my effort.

joakim-hove:
jenkins build this please

tests/test_ecl_output.cc
```diff
@@ -251,7 +257,7 @@ int main(int argc, char** argv)
     typedef TTAG(TestEclOutputTypeTag) TypeTag;
     Ewoms::registerAllParameters_<TypeTag>();
     test_summary();
-    test_readWriteWells();
+    //test_readWriteWells();
```
atgeirr: Why has this test been disabled?

joakim-hove: Debug

joakim-hove:
jenkins build this please

PR #1879, opened 2019-06-11
add `mebos`, a multiplexed ebos variant

andlaus:
`mebos` works similarly as `flow`, but in contrast to `flow`, `mebos` only creates the deck in the common code path whilst the 'EclipseState' and the other higher-level parser objects are always created internally by the vanguard. this approach avoids code duplication and the worst effects of parser API creep.

to avoid having to compile non-trivial compile units multiple times, the actual code of the variants is moved into `ebos_$VARIANT.{hh,cc}` files and the respective compile units are each put into a small static library whilst the main function of said libraries are invoked by either the multiplexed or the respective specialized simulator's `main()`. This is also somewhat similar of how `flow` works, with the difference that `mebos` uses the blackoil variant to determine the parameters it needs to know for parsing the deck instead of introducing a "fake" type tag for this. The rationale is to reduce compile time compared to the "fake type tag" approach and -- to a lesser extend -- avoid unnecessary copy-and-pasting of code. In particular, this means that for the vast majority of cases, only one place needs changed in the code for all `ebos` variants if, for example, the parser API requires further objects in the future.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,14 +185,17 @@ endif()
   
 # the research oriented general-purpose ECL simulator ("ebos" == &ecl
 # &black-&oil &simulator)
+get_target_property(ecl_INCLUDE_DIRS ecl INTERFACE_INCLUDE_DIRECTORIES)
+include_directories("${ecl_INCLUDE_DIRS}")
+
+add_library(ebos_libblackoil STATIC EXCLUDE_FROM_ALL ebos/ebos_blackoil.cc)
 opm_add_test(ebos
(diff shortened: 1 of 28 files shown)


andlaus:
jenkins build this please

andlaus:
jenkins build this please

andlaus:
yay, jenkins changed its mind without doing any changes to the PR. always a good sign...

totto82:
This was green earlier, so it should be fine to merge. Doesn't affect Flow, but is a wanted extension to ebos. Thanks.

PR #1882, opened 2019-06-12
[WIP] support GOR limit for WECON

GitPaean:
It is trivial and basically a copy from the water cut limit. 

If we are about to add more ratio limit checking, we need to refactor the ratio checking to reduce the duplication of the code. And we should move it  to a helper class/file instead of keeping it in WellInterface. 

It is not tested yet.

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -668,6 +668,98 @@ namespace Opm
     template<typename TypeTag>
     typename WellInterface<TypeTag>::RatioCheckTuple
     WellInterface<TypeTag>::
+    checkMaxGORLimit(const WellEconProductionLimits& econ_production_limits,
+                     const WellState& well_state) const
+    {
+        bool gor_limit_violated = false;
+        int worst_offending_completion = INVALIDCOMPLETION;
+        double violation_extent = -1.0;
+
+        const int np = number_of_phases_;
+        const Opm::PhaseUsage& pu = phaseUsage();
+
+        assert(FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx));
+        assert(FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx));
+
+        const double oil_rate = well_state.wellRates()[index_of_well_ * np + pu.phase_pos[ Oil ] ];
+        const double gas_rate = well_state.wellRates()[index_of_well_ * np + pu.phase_pos[ Gas ] ];
+        double gas_oil_ratio;
+        if (oil_rate != 0.) {
+            gas_oil_ratio = gas_rate / oil_rate;
+        } else {
+            if (gas_rate != 0.) {
+                gas_oil_ratio = 1.e100; // big value to mark it as violated
+            } else {
+                gas_oil_ratio = 0.0;
+            }
+        }
+
+        const double max_gor_limit = econ_production_limits.maxGasOilRatio();
+        if (gas_oil_ratio > max_gor_limit) {
+            gor_limit_violated = true;
+        }
+
+        if (gor_limit_violated) {
+            // need to handle the worst_offending_connection
+            const int perf_start = first_perf_;
+            const int perf_number = number_of_perforations_;
+
+            std::vector<double> gor_perf(perf_number);
+            for (int perf = 0; perf < perf_number; ++perf) {
+                const int i_perf = perf_start + perf;
+                const double oil_perf_rate = well_state.perfPhaseRates()[i_perf * np + pu.phase_pos[ Oil ] ];
+                const double gas_perf_rate = well_state.perfPhaseRates()[i_perf * np + pu.phase_pos[ Water ] ];
+                if (oil_perf_rate != 0.) {
+                    gor_perf[perf] = gas_perf_rate / oil_perf_rate;
+                } else {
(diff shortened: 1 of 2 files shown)


GitPaean:
jenkins build this please

GitPaean:
The refactoring will not be addressed within this PR, which should be a separate efforts to make the review and testing easier.

GitPaean:
jenkins build this please

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -665,6 +665,98 @@ namespace Opm
...
+        if (gor_limit_violated) {
+            // need to handle the worst_offending_connection
+            const int perf_start = first_perf_;
+            const int perf_number = number_of_perforations_;
+
+            std::vector<double> gor_perf(perf_number);
+            for (int perf = 0; perf < perf_number; ++perf) {
+                const int i_perf = perf_start + perf;
+                const double oil_perf_rate = well_state.perfPhaseRates()[i_perf * np + pu.phase_pos[ Oil ] ];
+                const double gas_perf_rate = well_state.perfPhaseRates()[i_perf * np + pu.phase_pos[ Water ] ];
```
atgeirr: `[ Water ]` here looks like a bug...

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -665,6 +665,98 @@ namespace Opm
...
+            const auto& completions = well_ecl_.getCompletions();
+            const auto& connections = well_ecl_.getConnections();
+
+            int complnumIdx = 0;
+            std::vector<double> gor_in_completions(completions.size(), 0.0);
+            for (const auto& completion : completions) {
+                int complnum = completion.first;
+                for (int perf = 0; perf < perf_number; ++perf) {
+                    if (complnum == connections.get ( perf ).complnum()) {
+                        gor_in_completions[complnumIdx] +=  gor_perf[perf];
```
atgeirr: Adding the gas-oil-ratios seems wrong to me, should we not rather accumulate the rates to compute the ratio?

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -665,6 +665,98 @@ namespace Opm
...
+                }
+            }
+            const auto& completions = well_ecl_.getCompletions();
+            const auto& connections = well_ecl_.getConnections();
+
+            int complnumIdx = 0;
+            std::vector<double> gor_in_completions(completions.size(), 0.0);
+            for (const auto& completion : completions) {
+                int complnum = completion.first;
+                for (int perf = 0; perf < perf_number; ++perf) {
```
atgeirr: This is O(n1 *n2), where n1 is the number of completions and n2 is the number of perforations. Can't we drop the outer loop here somehow?

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -665,6 +665,98 @@ namespace Opm
...
+
+        const int np = number_of_phases_;
+        const Opm::PhaseUsage& pu = phaseUsage();
+
+        assert(FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx));
+        assert(FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx));
+
+        const double oil_rate = well_state.wellRates()[index_of_well_ * np + pu.phase_pos[ Oil ] ];
+        const double gas_rate = well_state.wellRates()[index_of_well_ * np + pu.phase_pos[ Gas ] ];
+        double gas_oil_ratio;
```
atgeirr: The calculation below is repeated further down. Make it a lambda instead:
```
auto gor = [](double gas_rate, double oil_rate) { ... };
```

opm/simulators/wells/WellInterface_impl.hpp
GitPaean: Very good point! We need to address it.

opm/simulators/wells/WellInterface_impl.hpp
GitPaean: sure. Good catch. Basically, we need to do some correction and refactoring from this PR.

GitPaean: I think the correct behavoir is to loop through the completion number only.  Because even when we did not use keyword COMPLUMP, by default all the perforations get a completion number already.  let me check the implementations and it is possible that the concept of completion number should be used under more circumstances.

GitPaean: The original design did not consider the problem of `completion vs connection/perforation`. Some refactoring should be done before trying to implement this PR. It will be a separate PR.

GitPaean: I am coming back to this.  The function `getCompletions` returns ` std::map<int, std::vector<Connection>> `

The Connection does not have a index number, which makes it difficult to refer to the perf number to do the calculations here.

GitPaean:
Some tests shows the existing water cut also not work correctly. Will fix the existing water cut result first before continuing this PR.  New knowledge will be obtained through fixing the existing water cut functionality.

GitPaean:
With the refactoring done in #1898, this PR will be a much smaller PR and will done in a different way. Closing it and will open a new one to avoid confusion.

PR #1883, opened 2019-06-13
Upwinding for MSW

totto82:
This improves both convergence and accuracy of injecting MSW. 

Tested on model 2 realization.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -837,7 +837,7 @@ namespace Opm
             }
 
             // update the total rate // TODO: should we have a limitation of the total rate change?
-            {               
+            {
                 primary_variables_[seg][GTotal] = old_primary_variables[seg][GTotal] - relaxation_factor * dwells[seg][GTotal];
 
                 // make sure that no injector produce and no producer inject
@@ -1495,7 +1495,6 @@ namespace Opm
     assembleControlEq(Opm::DeferredLogger& deferred_logger) const
     {
         EvalWell control_eq(0.0);
-        const std::vector<double> g = {1.0, 1.0, 0.01};
 
         switch (well_controls_get_current_type(well_controls_)) {
             case THP: // not handling this one for now
@@ -1510,47 +1509,48 @@ namespace Opm
             }
             case SURFACE_RATE:
             {
-            // finding if it is a single phase control or combined phase control
-            int number_phases_under_control = 0;
-            const double* distr = well_controls_get_current_distr(well_controls_);
-            for (int phase = 0; phase < number_of_phases_; ++phase) {
-                if (distr[phase] > 0.0) {
-                    ++number_phases_under_control;
-                }
-            }
-            assert(number_phases_under_control > 0);
-            const double target_rate = well_controls_get_current_target(well_controls_);
-
-            if (well_type_ == INJECTOR) {
-                assert(number_phases_under_control == 1);
-                // only support single component injection
+                // finding if it is a single phase control or combined phase control
+                int number_phases_under_control = 0;
+                const double* distr = well_controls_get_current_distr(well_controls_);
                 for (int phase = 0; phase < number_of_phases_; ++phase) {
-                    if (distr[phase] > 0.) { // under the control of this phase
-                        control_eq = getSegmentGTotal(0) - g[phase] * target_rate;
-                        break;
+                    if (distr[phase] > 0.0) {
+                        ++number_phases_under_control;
                     }
                 }
-            } else {
-                // TODO: the two situations below should be able to be merged to be handled as one situation
-                if (number_phases_under_control == 1) { // single phase control
+                assert(number_phases_under_control > 0);
+                const double target_rate = well_controls_get_current_target(well_controls_);
+                const std::vector<double> g = {1.0, 1.0, 0.01};
+
+                if (well_type_ == INJECTOR) {
+                    assert(number_phases_under_control == 1);
+                    // only support single component injection
                     for (int phase = 0; phase < number_of_phases_; ++phase) {
                         if (distr[phase] > 0.) { // under the control of this phase
-                            control_eq = getSegmentGTotal(0) * volumeFraction(0, flowPhaseToEbosCompIdx(phase)) - g[phase] * target_rate;
+                            control_eq = getSegmentGTotal(0) - g[phase] * target_rate;
                             break;
                         }
                     }
-                } else { // multiphase rate control
-                    EvalWell rate_for_control(0.0);
-                    const EvalWell G_total = getSegmentGTotal(0);
-                    for (int phase = 0; phase < number_of_phases_; ++phase) {
-                        if (distr[phase] > 0.) {
-                            rate_for_control += G_total * volumeFractionScaled(0, flowPhaseToEbosCompIdx(phase));
+                } else {
+                    // TODO: the two situations below should be able to be merged to be handled as one situation
+                    if (number_phases_under_control == 1) { // single phase control
+                        for (int phase = 0; phase < number_of_phases_; ++phase) {
+                            if (distr[phase] > 0.) { // under the control of this phase
+                                control_eq = getSegmentGTotal(0) * volumeFraction(0, flowPhaseToEbosCompIdx(phase)) - g[phase] * target_rate;
+                                break;
+                            }
                         }
+                    } else { // multiphase rate control
+                        EvalWell rate_for_control(0.0);
+                        const EvalWell G_total = getSegmentGTotal(0);
+                        for (int phase = 0; phase < number_of_phases_; ++phase) {
+                            if (distr[phase] > 0.) {
+                                rate_for_control += G_total * volumeFractionScaled(0, flowPhaseToEbosCompIdx(phase));
+                            }
+                        }
+                        // TODO: maybe the following equation can be scaled a little bit for gas phase
+                        control_eq = rate_for_control - target_rate;
                     }
-                    // TODO: maybe the following equation can be scaled a little bit for gas phase
-                    control_eq = rate_for_control - target_rate;
                 }
-            }
                 break;
             }
             case RESERVOIR_RATE:
@@ -1569,8 +1569,6 @@ namespace Opm
             default:
                 OPM_DEFLOG_THROW(std::runtime_error, "Unknown well control control types for well " << name(), deferred_logger);
         }
-
-
         // using control_eq to update the matrix and residuals
 
         resWell_[0][SPres] = control_eq.value();
@@ -2455,16 +2453,9 @@ namespace Opm
         for (int seg = 0; seg < numberOfSegments(); ++seg) {
             // special treatment is needed for segment 0
             if (seg == 0) {
-                if ( (well_type_ == INJECTOR)) {
-                    // maybe it is safer to return -1 to make sure that we don't return the wrong segment number
-                    // for injectors
-                    upwinding_segments_[seg] = seg;
-                    assert(primary_variables_evaluation_[seg][GTotal] >= 0.);
-                    continue;
-                }
-
-                // we don't handle injecting producers.
-                assert(primary_variables_evaluation_[seg][GTotal] <= 0.);
+                // we are not supposed to have injecting producers and producing injectors
+                assert( ! (well_type_ == PRODUCER && primary_variables_evaluation_[seg][GTotal] > 0.) );
+                assert( ! (well_type_ == INJECTOR && primary_variables_evaluation_[seg][GTotal] < 0.) );
                 upwinding_segments_[seg] = seg;
                 continue;
             }
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1394,6 +1394,11 @@ namespace Opm
         // the result will contain the derivative with resepct to GTotal in segment seg,
         // and the derivatives with respect to WFrac GFrac in segment seg_upwind.
         // the derivative with respect to SPres should be zero.
+        if (seg == 0 && well_type_ == INJECTOR) {
+            const double* distr = well_controls_get_current_distr(well_controls_);
+            return primary_variables_evaluation_[seg][GTotal] * distr[ebosCompIdxToFlowCompIdx(comp_idx)];
+        }
+
         const EvalWell segment_rate = primary_variables_evaluation_[seg][GTotal] * volumeFractionScaled(seg_upwind, comp_idx);
 
         assert(segment_rate.derivative(SPres + numEq) == 0.);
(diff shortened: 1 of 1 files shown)


GitPaean:
There are two separate parts for this PR. How does each part impact the convergence behavoir separately?

totto82:
You need both. I can squeeze them if you want.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1493,6 +1498,10 @@ namespace Opm
             }
             case SURFACE_RATE:
             {
+            if (well_type_ == INJECTOR) {
+                const double target_rate = well_controls_get_current_target(well_controls_);
+                control_eq = getSegmentGTotal(0) - target_rate;
```
GitPaean: This is wrong for gas injection.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2426,11 +2436,19 @@ namespace Opm
...
+                    // maybe it is safer to return -1 to make sure that we don't return the wrong segment number
+                    // for injectors
+                    upwinding_segments_[seg] = seg;
+                    assert(primary_variables_evaluation_[seg][GTotal] >= 0.);
+                    continue;
+                }
+
+                // we don't handle injecting producers.
+                assert(primary_variables_evaluation_[seg][GTotal] <= 0.);
                 upwinding_segments_[seg] = seg;
```
totto82: This assert is triggered in the 3D_MSW test case. How is injecting producers handled in the rest of the code?

GitPaean:
> You need both. I can squeeze them if you want.

Please do not squeeze them. They both impacts the convergence, it is good to understand their separate effects so we can elaborate them separately.

opm/simulators/wells/MultisegmentWell_impl.hpp
GitPaean: For StandardWell it is easier to understand. It is fine to let a producer inject unless THP is involved. For MSW, I do not have an answer yet.  MSW typically needs to handle more complicated fluid behavoir within wellbore.

GitPaean: Short answer is that we have not cared too much about whether a producer is injecting for how. Other constraint might recover the well to return to producing.  Not sure whether this is the final design we want.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: I completely forgot that we discussed this last week. Hopefully, my fix will work. Testing in progress.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1394,6 +1394,11 @@ namespace Opm
         // the result will contain the derivative with resepct to GTotal in segment seg,
         // and the derivatives with respect to WFrac GFrac in segment seg_upwind.
         // the derivative with respect to SPres should be zero.
+        if (seg == 0 && well_type_ == INJECTOR) {
+            const double* distr = well_controls_get_current_distr(well_controls_);
```
GitPaean: gas rate will be wrong here.

GitPaean: If it is RESV control, then the rate for all the phases will be wrong.

GitPaean:
As you might see, although it is not a lot of code, it is still not easy to get everything in correct form. The upwinding part for injecting is very welcomed, even was known as commented and it was not the focus with the old cases.

while about how to calculate the rate for injectors, I suggest to put it in another PR.  This one is the more tricky one, let us discuss and test it separately. 

The upwinding part should be easy to go in with some testing and discussion.  

Then we focus on the injecting rate part.  More testing and discussion will be needed for this part.

totto82:
As I said earlier they are both related and it doesn't make sense to test them separately IMO. They both fixes the upwinding. One for seg>1 and one for seg=0. Why was the upwinding part not included for injectors in the first place? No wonder we had lot of convergence issues for injectors. I am aware that the well model is fragile. That is something we need to continue to improve. But this fixes an obvious bug that significantly improves the convergence and accuracy of the MSW model. I will test them together. I have asked @abir to provide a test case for MSW injectors. A proper MSW test for injectors would have saved us a lot of time.

GitPaean:
> Why was the upwinding part not included for injectors in the first place? No wonder we had lot of convergence issues for injectors.

Because it was not a problem for the previous tests. But we do know it will be a problem eventually, that is why some comments were left there. Everything went in step by step since it is safer. For examples, several things went in together, something is broken, it will be panic to figure it out. And we never understand how each individual works.  And it will result in bigger PRs, more difficult to provide review with good quality. 

> As I said earlier they are both related and it doesn't make sense to test them separately IMO.

It is doable. I will test them separate locally.  Probably it will take longer than two separate PRs.

> I have asked @abir to provide a test case for MSW injectors. A proper MSW test for injectors would have saved us a lot of time.

That is good. At the beginning, we only have cases with MSW producers and it is difficult for us to edit MSW cases manually. We created a few, while it was time consuming. 


Passing tests is one thing, and we also need to make sure the code is correct, because we will not be able to have tests for all kinds of possibilities.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: I am working on fixing this. I will give you a hint when it is ready. No need for testing from your side before this is fixed. I need to assure that everything is scaled with the same factor. Do you have a small test case for MSW injectors that you could send me?

GitPaean: > Do you have a small test case for MSW injectors that you could send me?

No. On the other hand, smaller test for MSW injectors might not need this PR to fix, since it is difficult to produce the complicated situation need to do the upwinding.  For the producers, we saw one single case gives strong sign that upwinding is required for that report step.

totto82: Fixed.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -837,8 +837,17 @@ namespace Opm
             }
 
             // update the total rate // TODO: should we have a limitation of the total rate change?
-            {
+            {               
                 primary_variables_[seg][GTotal] = old_primary_variables[seg][GTotal] - relaxation_factor * dwells[seg][GTotal];
+
+                // make sure that no injector produce and no producer inject
```
GitPaean: Please do not do this, unless we have a good reason to do it.  Similar to in the pressure update, std::max will be likely to be removed. 

Allowing or disallowing injector to produce or producer to inject will be a design problem. Similar to the trick related to function `openCrossFlowAvoidSingularity()` in StandardWell, which was introduced by you to overpass some singularity problem.

totto82: There is a good reason for avoiding injecting producers since at the moment this is undefined behavior. I don't see any reasons for not introducing this. Please argue why you don't think this is a good solution. It seems to work for the cases I have tested. We already do a lot of chopping, why not here?

GitPaean: We might need this feature to test whether a well is operable or not.  The only places that negative rates are dangerous is when THP is involved. I do not have a answer for it yet.

GitPaean: Anyway, it is another thing needs to be tested separately if we really want it.

totto82: I suggest that we merge this then, since it fix a actual and urgent problem for a model 2 realization and you can remove it whenever you have a better solution.

GitPaean: > I suggest that we merge this then, since it fix a actual and urgent problem for a model 2 realization and you can remove it whenever you have a better solution.

Please indicate which realization you are talking about for this PR.  Can be done through private message.

totto82:
jenkins build this please

totto82:
I have manually inspected the test failure, and there is no changes in the "eyeball" norm.

GitPaean:
Tested with the targeting model, this PR fixes the running of the model. The following is the result. 
```
552274 Total time (seconds):         5128.34
552275 Solver time (seconds):        5128.03
552276  Assembly time (seconds):     1636.58 (Failed: 251.1; 15.3429%)
552277  Linear solve time (seconds): 3328.33 (Failed: 337.618; 10.1438%)
552278  Linear solve setup time (seconds): 76.957 (Failed: 12.3524; 16.051%)
552279  Update time (seconds):       77.1977 (Failed: 11.6887; 15.1412%)
552280  Output write time (seconds): 19.9517
552281 Overall Well Iterations:      1522 (Failed: 166; 10.9067%)
552282 Overall Linearizations:       4620 (Failed: 648; 14.026%)
552283 Overall Newton Iterations:    4049 (Failed: 648; 16.004%)
552284 Overall Linear Iterations:    79298 (Failed: 7132; 8.99392%)
```

The major changes in this PR is the change of the following line
```
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -2430,7 +2430,7 @@ namespace Opm
         // but we should
         // and upwinding segment for top segment is itself
         for (int seg = 0; seg < numberOfSegments(); ++seg) {
-            if ( (well_type_ == INJECTOR) || (seg == 0) ) {
+            if ( seg == 0 ) {
                 upwinding_segments_[seg] = seg;
                 continue;
             }
```

With this single line change, the running is also fixed with the following result
```
554461 Total time (seconds):         4880.01
554462 Solver time (seconds):        4879.72
554463  Assembly time (seconds):     1589.87 (Failed: 258.455; 16.2563%)
554464  Linear solve time (seconds): 3135.25 (Failed: 337.448; 10.763%)
554465  Linear solve setup time (seconds): 76.8549 (Failed: 13.1763; 17.1444%)
554466  Update time (seconds):       74.2945 (Failed: 11.989; 16.1371%)
554467  Output write time (seconds): 18.5609
554468 Overall Well Iterations:      1521 (Failed: 173; 11.3741%)
554469 Overall Linearizations:       4685 (Failed: 714; 15.2401%)
554470 Overall Newton Iterations:    4112 (Failed: 714; 17.3638%)
554471 Overall Linear Iterations:    79769 (Failed: 7776; 9.74815%)
554472 
```

totto82:
> Tested with the targeting model

Does this mean you have compared the results?

GitPaean:
> Does this mean you have compared the results?

We do not have the reference result. Please share your finding.

GitPaean:
Basically, from my side, the change to the function `updateUpwindingSegments` to activate the upwinding of for the injectors is not controversial. 

It needs significantly more work to make the other part correct. If it is worth the efforts, we will do it.

totto82:
> It needs significantly more work to make the other part correct.
What is wrong with my approach?

GitPaean:
> What is wrong with my approach?

Please share your finding, then we will focus on it.  It is tricky to fix it for all kinds of circumstances and it is good to get in things with a good form with a good understanding to avoid later problems.

totto82:
I have tested your fix and it is not sufficient to get good results on the model in question.

totto82:
IMO I have addressed the issues you raised. If you still have issues please be specific so we don't need to delay this important bugfix more than necessary.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1493,20 +1510,28 @@ namespace Opm
             }
             case SURFACE_RATE:
             {
-                // finding if it is a single phase control or combined phase control
```
GitPaean: Here, there should be much less code changing here, while some indentation are changed by accident which makes it looks like much more codes are changed.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2426,11 +2452,19 @@ namespace Opm
...
     updateUpwindingSegments()
     {
-        // not considering upwinding for the injectors for now
-        // but we should
-        // and upwinding segment for top segment is itself
         for (int seg = 0; seg < numberOfSegments(); ++seg) {
-            if ( (well_type_ == INJECTOR) || (seg == 0) ) {
+            // special treatment is needed for segment 0
+            if (seg == 0) {
+                if ( (well_type_ == INJECTOR)) {
```
GitPaean: The only difference inside `if` and outside the `if` is the `assert`.  

You can use the assertion of combined logic to avoid to have if and duplicating the following code
```
 upwinding_segments_[seg] = seg;
 continue;
```


Something like assert( ! (well_type_ == INJECTOR && primary_variables_evaluation_[seg][GTotal] <= 0.));

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -837,8 +837,17 @@ namespace Opm
             }
 
             // update the total rate // TODO: should we have a limitation of the total rate change?
-            {
+            {
```
GitPaean: Please remove the white spaces here.

alfbr:
> We do not have the reference result.

I believe it suffices to run flow with and without this PR on the case, you should see significant differences.

atgeirr:
> I believe it suffices to run flow with and without this PR on the case, you should see significant differences.

Yes, but we want to compare the minimum change suggested by @GitPaean to this PR, and while we see differences in wells it is not obvious which one is the best without comparing to a reference.

alfbr:
> Yes, but we want to compare the minimum change suggested by @GitPaean to this PR, and while we see differences in wells it is not obvious which one is the best without comparing to a reference.

According to @totto82 , the differences are indeed what is observed in the reference. I have started a run now so that you can inspect yourself. It will take some hours, but I hope to have it available to you before I go on travel this afternoon.

atgeirr:
We have discussed this PR in a hangout, and it will likely be merged tomorrow after minor revisions.

GitPaean:
> According to @totto82 , the differences are indeed what is observed in the reference

Thanks for providing the reference.  Basically, when it converges, it should converge at the similar results. It it converges at the different solution, it should only happen when the time step size is so small. It is not a so small time step, and it converges at a different solution, we need to understand it.

totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1481,32 +1497,41 @@ namespace Opm
...
-            case THP: // not handling this one for now
-            {
-                OPM_DEFLOG_THROW(std::runtime_error, "Not handling THP control for Multisegment wells for now", deferred_logger);
-            }
-            case BHP:
-            {
-                const double target_bhp = well_controls_get_current_target(well_controls_);
-                control_eq = getSegmentPressure(0) - target_bhp;
-                break;
+        case THP: // not handling this one for now
```
GitPaean: Really sorry for the nittpicking. It is not the purpose of the original comment.  You can create a different pull request if you want to address the indentation of the switch in the whole file, that is perfectly fine. 

It does not makes the pull request easier to read by changing all the indentations here. It is not the style of the switch for this file either.

totto82: >not the style of the switch for this file either.

The file is not consistent with indentation for switches. I choose one of the styles used. I don't have any plan to make a PR to clean-up the well model for indentations. I tried to fix it the way you wanted it to be. I suggest that you make a PR later to synchronize the style in the file, then you get it the way you want.

totto82: Is it the way you want know?

totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
GitPaean: It is better than the previous version.  The code is pretty complicated now. If you would like to, you can put the surface rate control in one function, even split them to two functions for injectors and producers. Then the code will be easier to read.  You do not have to do it. Just let me know.

totto82:
We need an data update for this. I have manually tested the failed case and it shows no differences in the "eyeballnorm". The change for results and convergence can be explained by not allowing producers to injected even a tiny drop anymore during iterations.

totto82:
jenskins build this update_data please

akva2:

akva2:
update data is broken on new jenkins setup, i'll fix asap.

akva2:

akva2:

akva2:
jenkins build this opm-tests=129 please

GitPaean:
Introducing upwinding for injectors is good. The development of MSW has been always focusing on producers.  This PR makes the injecting functionality for MSW complete and robust.  Enforcing the flowing direction of top segments is not needed for producing correct results for the targeting case, while it is required here for the completeness of the current upwinding logic. As agreed, this PR is going in now.

PR #1884, opened 2019-06-13
changed: simplify buildsystem a little

akva2:
- use loops instead of copy-paste
- use object libraries to avoid some unnecessary archives
- tab scrubbing


akva2:
jenkins build this please

atgeirr:
Looks good to me, though I am an amateur at cmake.

akva2:
ping @blattms

blattms:
Sorry, my main compute system is currently broken and thus this took longer.

Looks good and tested locally. Merging.

blattms:
The only thing that I do not get, but that was already there before this PR. Why are we using opm_add_test to build something that we are actually installing? Seems like this could be break
Indeed installation fails withcmake  -DBUILD_EBOS_EXTENSION=ON -DBUILD_TESTING:BOOL=OFF

PR #1886, opened 2019-06-14
Let the output writer determine what to write in ebos

totto82:
Now Ebos outputs summary data for each substep as it should. This should not affect Flow.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -983,24 +983,10 @@ public:
     }
 
     /*!
-     * \brief Returns true if the current solution should be written
-     *        to disk for visualization.
-     *
-     * For the ECL simulator we only write at the end of
-     * episodes/report steps...
+     * \brief Always returns true. The ecl output writer takes care of the rest
      */
     bool shouldWriteOutput() const
-    {
-        const auto& simulator = this->simulator();
-        if (simulator.timeStepIndex() < 0)
-            // always write the initial solution
-            return true;
-
-        if (EWOMS_GET_PARAM(TypeTag, bool, EnableWriteAllSolutions))
-            return true;
-
-        return simulator.episodeWillBeOver();
-    }
+    { return true; }
 
     /*!
      * \brief Returns true if an eWoms restart file should be written to disk.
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -1021,8 +1007,10 @@ public:
         // write the desired VTK files.
         ParentType::writeOutput(verbose);
 
+        bool isSubStep = EWOMS_GET_PARAM(TypeTag, bool, EnableWriteAllSolutions) || !this->simulator().episodeWillBeOver();
```
bska: I don't think this does what you intended to achieve.

Suppose EnableWriteAllSolutions is true. Then isSubStep is always true. That, in turn, means that EclWriter::writeOutput() will treat every write request as if it is a substep (ministep) which ultimately means that no restart file will be created at all. You'll only get get summary output.

totto82: I know. Just discovered this myself as well. Does the writer support BASIC=6? Then we could call the writer unconditionally and let the eclWriter handle this.

joakim-hove: Basic=6 has not been tested and (most probably) does not work.

totto82: It seems like there is an issue with this. The restarts are written out every timesteps, but it seems like the last timesteps get replaced. Or just resinsight gets confused since the same episodeIndex is passed to the outputwriter.

totto82: There seems to be some work left to support BASIC=6 in the eclWriter. I propose that we merge this fix for opm-simulators and fix opm-common later.

totto82:
jenkins build this please

andlaus:
okay: I agree with @totto82 and since jenkins is happy and this should thus not affect `flow`, I don't see any reason not to merge. Thanks for the patch.

bska:
> this should thus not affect `flow`

Should not or does not?  This is in the middle of Flow's output path.

totto82:
The shouldWriteOutput is not called by Flow. And the writeAllTimeStep option is not  correctly supported before nor after this PR.

bska:
> The `shouldWriteOutput` is not called by Flow. And the `writeAllTimeStep` option is not correctly supported before nor after this PR.

That's not what I asked.  If there truly is no impact on Flow, then I should be able to run with `WriteAllSolutions` being `true` and get the exact same output as if `WriteAllSolutions` were `false`.  Is that the case?

totto82:
> Is that the case?

No, without this PR. The writeAllSolutions is simply ignored. Nothing happens for Flow. With this data is written to the restart file every time step but the data overwrites the previous time step until a new episode is encountered.

PR #1891, opened 2019-06-18
[WIP]: Relaxed tolerances

alfbr:
These tolerances decrease runtime and increase numerical robustness significantly. However, they give significant deviations on some models. Other approaches to reach the same performance without sacrificing accuracy is under way, so this PR is only intended for testing at this point.


alfbr:
These changes are not relevant anymore, since the numerical defaults have changed dramatically. Closing.

PR #1892, opened 2019-06-18
fixed: build with dune 2.6 without MPI

akva2:


akva2:
jenkins build this dune2.6 serial please

akva2:
jenkins build this dune2.6 serial please

atgeirr:
Thanks, will merge. I do not think the configuration this is supposed to fix is actually tested by jenkins, correct? (But the jenkins test ensure other configurations were not broken of course.)

akva2:
i actually changed it now. if you include both dune2.6 and serial it will test a serial dune 2.6 as well.

PR #1895, opened 2019-06-19
Add applyActions() to EclProblem

joakim-hove:
Lo and behold - ACTIONX is coming to town!

The current PR seems to run the [UDQ_ACTIONX](https://github.com/OPM/opm-tests/blob/master/udq_actionx/UDQ_ACTIONX.DATA) testcase - although there is a crash at the end of the simulation (which will be fixed in opm-common).

Observe that I have done things slightly differently from what is the common pattern in ebos based simulators; instead of fetching the required objects through omnipresent back-references I have opted to pass the required arguments explicitly. '

**Update**: In all decency I added WIP label; codewise I actually think this is ready - but quite a lot more testing is needed ....

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -87,6 +87,7 @@
 #include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
 #include <opm/parser/eclipse/EclipseState/Tables/Eqldims.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/Action/ActionContext.hpp>
 #include <opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp>
 #include <opm/parser/eclipse/EclipseState/Tables/RockwnodTable.hpp>
 #include <opm/parser/eclipse/EclipseState/Tables/OverburdTable.hpp>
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

opm/autodiff/SimulatorFullyImplicitBlackoilEbos.hpp
```diff
@@ -295,6 +295,9 @@ class SimulatorFullyImplicitBlackoilEbos
             const double nextstep = adaptiveTimeStepping ? adaptiveTimeStepping->suggestedNextStep() : -1.0;
             ebosSimulator_.problem().setNextTimeStepSize(nextstep);
             ebosSimulator_.problem().writeOutput();
+            ebosSimulator_.problem().applyActions(timer.currentStepNum(),
+                                                  ebosSimulator_.vanguard().schedule(),
+                                                  ebosSimulator_.vanguard().summaryState());
             report.output_write_time += perfTimer.stop();
```
totto82: I would prefer that the applyAction(...) call is moved to endEpisode(...) in eclproblem.hh rather than put here.

joakim-hove: Sure

joakim-hove:
OK: With https://github.com/OPM/opm-common/pull/873 applied this "works" in the sense that the conditions in testcase: https://github.com/OPM/opm-tests/blob/master/udq_actionx/UDQ_ACTIONX.DATA  are evaluated correctly and the well is shut when it should - but from then on the convergences grinds to a halt. Not clear to me where the fault of the convergence problems are. Would be good if someone with more knowledge of group control and so on could take a look.

joakim-hove:
Ping! I have now rebased and updated this branch; the comment from 1.st of July still stands!

joakim-hove:
Ping Pong!

I have removed the WIP on this; I am very much prepared make changes to it - but from my hand it is currently ready.

totto82:
This can be merged when Jenkins is ready.

totto82:
jenkins build this please

PR #1896, opened 2019-06-19
Adding time and date to time step report

alfbr:
When there are multiple time steps between two report steps, it can be difficult to see where in the total simulation time a time step is. Hence, there has been user request for the time steps to include the same day and date information found in report steps. This PR provides that.


atgeirr:
jenkins build this please

atgeirr:
Thanks, merging!

PR #1897, opened 2019-06-20
Moved well test messages from debug to log and prt

alfbr:
Two important wtest messages was only visible in the DBG-file. With this pull request they are now also shown in the terminal/LOG and in the PRT file.


GitPaean:
jenkins build this please

GitPaean:
It looks good. Thanks.

PR #1901, opened 2019-06-20
Make sure rvmax and rsmax is non-negtive in the getSurfaceVolume for MSW

totto82:
Negative values can happen if the pressure is outside the table range. 

This makes the soon to come test case involving MSW injector reasonable well potentials.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -2173,7 +2173,10 @@ namespace Opm
             const unsigned gasCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx);
             if (FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx)) {
                 const unsigned oilCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::oilCompIdx);
-                const EvalWell rvmax = FluidSystem::gasPvt().saturatedOilVaporizationFactor(pvt_region_index, temperature, seg_pressure);
+                EvalWell rvmax = FluidSystem::gasPvt().saturatedOilVaporizationFactor(pvt_region_index, temperature, seg_pressure);
+                if (rvmax < 0.0) { // negative rvmax can happen if the seg_pressure is outside the range of the table
+                    rvmax = 0.0;
+                }
                 if (mix_s[oilCompIdx] > 0.0) {
                     if (mix_s[gasCompIdx] > 0.0) {
                         rv = mix_s[oilCompIdx] / mix_s[gasCompIdx];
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

GitPaean:
I suppose for injectors, there will be a high bhp limit.  Some previous experience shows too big rsmax and rvmax value with big injector bhp, which results in `d = 1. - rs * rv` == 0. , which means all the oil can evaporate into gas and all the gas can dissolve into the oil. 

Any clue, how we get the negative rsmax and rvmax here?

totto82:
It is the rsmax and rvmax that gets negative due to extrapolation issues. For instance if your table range is from 50 bars and you want to compute the rsSat value for 1 bar then you might get negative rsSat values. This PR only make sure that rsmax and rvmax is greater then zero.

GitPaean:
>  For instance if your table range is from 50 bars and you want to compute the rsSat value for 1 bar then you might get negative rsSat values. 

Sure. It is for the producers with defaulted bhp limit.

totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2173,7 +2173,10 @@ namespace Opm
             const unsigned gasCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx);
             if (FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx)) {
                 const unsigned oilCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::oilCompIdx);
-                const EvalWell rvmax = FluidSystem::gasPvt().saturatedOilVaporizationFactor(pvt_region_index, temperature, seg_pressure);
+                EvalWell rvmax = FluidSystem::gasPvt().saturatedOilVaporizationFactor(pvt_region_index, temperature, seg_pressure);
+                if (rvmax < 0.0) { // negative rvmax can happen if the seg_pressure is outside the range of the table
```
GitPaean: It is good to be more specific, to say it is due to pressure lower than the table range.  Higher one will result in different symptom.

GitPaean:
I am okay with the current approach.  I am also thinking it can be done from the table code side, since the function will also be called in multiple places. In any places, negative rsmax and rvmax values are not desirable.  Any suggestions, @andlaus , @atgeirr , @hnil

totto82:
I thought about the "table approach" but I don't think we at the moment want a general fix for this. In IMO tables should be in the range of the values you encounter during simulations. Well potential for defaulted bhp limits is not very valuable and a fix like this just makes sure the calculations go throw.

GitPaean:
>  a fix like this just makes sure the calculations go throw.

You mean, the simulation will throw after chopping the rsmax and rvmax values?

GitPaean:
Sorry. I got it now, you means, `go through`.

GitPaean:
> Well potential for defaulted bhp limits is not very valuable and a fix like this just makes sure the calculations go throw.

In a more general sense, failure in calculating well potentials for outputting purpose should be a warning.

totto82:
@GitPaean Anything holding this back?

GitPaean:
> @GitPaean Anything holding this back?

We are waiting for the input from @hnil . He re-wrote the code and hopefully he might give some good insight.

GitPaean:
Any warning message is given from commercial simulator about this extrapolation?  I mean, it is out of the table range. 

I remember there is some warning when using defaulted BHP limit for well potential calculation.

GitPaean:
We did not manage to get chance to discuss this. People are leaving. 
Let us merge it for now for convenience purpose.  The damaging is limited. 

It might be reverted or changed if other solution is suggested.  I do think even for chopping this value to be zero here, chopping from the table code is more promising solution to prevent us from doing the same fix for multiple times. 

And a warning should be given.  Extrapolation out of table range should always let people know so they need to either update the table or update the bhp limit.

GitPaean:
We discussed about this issue. It is okay or better to fix it through the table code.

PR #1902, opened 2019-06-21
Only apply timestep after event at the beginning of an episode

totto82:
Depends on OPM/ewoms#521


totto82:
jenkins build this ewoms=521 please

andlaus:
Since flow is almost certainly unaffected, jenkins and myself are happy, this seems to fix a severe bug in the `ebos` implementation of TUNING and I trust that you tested this, I don't see any reason to hold off. thanks for the patch!

PR #1907, opened 2019-06-24
Improve time integration

andlaus:
Together with OPM/ewoms#522, this improves the time integration code of `EclProblem`. This piece of code caused problems when I was using `ebos` for one of the confidential decks I need to work with when tuning was enabled and the minimum time step size was undercut. Since `flow` rolls its own time stepping code, it is almost certainly unaffected.

@totto82: It would be good if you could ensure that the confidential cases which I recently lost access to do not regress with tuning enabled. (I still expect the majority of them to be faster without tuning, but if tuning is enabled, `ebos` with these patches ought not to be slower than on the current master.)

### DIFF
--- a/ebos/ebos.hh
+++ b/ebos/ebos.hh
@@ -99,6 +99,12 @@ SET_INT_PROP(EbosTypeTag, EclNewtonStrictIterations, 100);
 // relatively often)
 SET_INT_PROP(EbosTypeTag, NewtonMaxIterations, 8);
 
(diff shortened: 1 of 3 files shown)


andlaus:
jenkins build this ewoms=522 please

andlaus:
jenkins build this ewoms=522 please

andlaus:
jenkins build this ewoms=522 please

PR #1911, opened 2019-06-25
adapting to the upstream change

GitPaean:
it adapts the code following https://github.com/OPM/opm-common/pull/854 . 

And a small bug fix.  It should be `sim_time` while the original code is `0`.


atgeirr:
I'll merge this as the upstream was already merged!

PR #1913, opened 2019-06-25
check if either reference or output dir has file in update script

akva2:
to avoid false positives due to missing rft files


akva2:
included in #1927

PR #1916, opened 2019-06-26
mebos, flow: adapt to the eWoms changes in handling unknown parameters

andlaus:
this is the downstream for OPM/ewoms#524. With this patch, there's no change at all for `flow` compared to the current behavior, `ebos` and its variants will complain when it encounters unused parameters and the `--help` handling is correct for `mebos`.

### DIFF
--- a/opm/simulators/flow/FlowMainEbos.hpp
+++ b/opm/simulators/flow/FlowMainEbos.hpp
@@ -163,7 +163,7 @@ namespace Opm
             EWOMS_END_PARAM_REGISTRATION(TypeTag);
 
             // read in the command line parameters
-            int status = Ewoms::setupParameters_<TypeTag>(argc, const_cast<const char**>(argv), false, true, true);
+            int status = Ewoms::setupParameters_<TypeTag>(argc, const_cast<const char**>(argv), /*doRegistration=*/false, /*allowUnused=*/true, /*handleHelp=*/true);
             if (status == 0) {
                 // deal with --print-properties and --print-parameters
 
--- a/opm/simulators/flow/FlowMainEbos.hpp
+++ b/opm/simulators/flow/FlowMainEbos.hpp
@@ -163,7 +163,7 @@ namespace Opm
             EWOMS_END_PARAM_REGISTRATION(TypeTag);
 
             // read in the command line parameters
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this ewoms=524 please

opm/simulators/flow/FlowMainEbos.hpp
```diff
@@ -163,7 +163,7 @@ namespace Opm
             EWOMS_END_PARAM_REGISTRATION(TypeTag);
 
             // read in the command line parameters
-            int status = Ewoms::setupParameters_<TypeTag>(argc, const_cast<const char**>(argv), /*doRegistration=*/false);
+            int status = Ewoms::setupParameters_<TypeTag>(argc, const_cast<const char**>(argv), false, true, true);
```
blattms: Why do we allow unused parameters? Shouldn't this PR (if it is correct) make #1845 abosolete?

BTW: without your helpful comments about what bool is what, this is rather complicated and unreadable. May I suggest adding `/*doRegistration*/` and the likes? That would really help.

How often are we actually calling setupParameters_ in flow?

Why the trailing underline in the name? Wasn't that used to mark functions considered internal? Yet. this one is used externally. That seems a bit confusing.

andlaus: > Why do we allow unused parameters?

at this point we cannot be sure which parameters are available: extensions like polymer, solvent and thermal might introduce some of their own.

> Shouldn't this PR (if it is correct) make #1845 abosolete?

no, not currently.

> BTW: without your helpful comments about what bool is what, this is rather complicated and unreadable. May I suggest adding /*doRegistration*/ and the likes? That would really help.

almost all places in the opm-simulators (except `ebos/`), opm-common and and opm-grid written by myself do not use this. I thought I adapt my coding style to the surroundings...

> How often are we actually calling setupParameters_ in flow?

`flow`: twice
`mebos`: thrice
`ebos*`: once

> Why the trailing underline in the name? Wasn't that used to mark functions considered internal? Yet. this one is used externally. That seems a bit confusing.

it is supposed to be an internal function only used by `Ewoms::start()`. since `flow` and `mebos` need to do some low-level meddling, they need to use it as well...

totto82:
jenkins build this please

PR #1917, opened 2019-06-26
more fixing for the tests

GitPaean:


GitPaean:
jenkins build this opm-common=858 please

GitPaean:
jenkins build this opm-common=858 please

opm/core/wells/WellsGroup.cpp
```diff
@@ -1485,6 +1485,7 @@ namespace Opm
                 case Phase::POLYMER: OPM_THROW(std::invalid_argument, "Polymer injector is not supported.");
                 case Phase::ENERGY:  OPM_THROW(std::invalid_argument, "Energy injector is not supported.");
                 case Phase::POLYMW:  OPM_THROW(std::invalid_argument, "PolyMW injector is not supported.");
+                case Phase::FOAM:    OPM_THROW(std::invalid_argument, "Foam injector is not supported.");
             }
             OPM_THROW(std::logic_error, "Invalid state." );
```
GitPaean: `"Invalid state."` is a very bad message here.  Probably we should add a default for the switch.  We should not address it in this PR though.

bska: > Probably we should add a default for the switch.

Preferably not.  That will hide a useful compiler diagnostic that some enum values are not handled.

atgeirr (approved):
Fine, will merge.

opm/core/wells/WellsGroup.cpp
GitPaean: Does it need to be a `enum class` or it will work for the usual `enum`?

PR #1918, opened 2019-06-26
Format Specifier '%p' Requires Pointers to Void

bska:
While here, also print both `ctrls1` and `ctrls2` instead of printing `ctrls2` twice.


bska:
jenkins build this please

atgeirr:
Wow, that is so C! Merging.

bska:
> Wow, that is so C!

Yeah... At least it's so "variadic function".

PR #1919, opened 2019-06-26
improving WTEST results.

GitPaean:
It is still working in progress. More testing and some refactoring will be done.  The problem of the original version is the explicit quantities are not calculated. 

For the cases in wtest/wecon_wct_max in opm-tests, it improves a lot. Probably it will also improves the results in other cases, while they are not tested yet. 

More investigation and testing will be done before this PR is considered to be ready.

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -1017,10 +1017,11 @@ namespace Opm
 
         WellState well_state_copy = well_state;
 
+        updateWellStateWithTarget(simulator, well_state_copy, deferred_logger);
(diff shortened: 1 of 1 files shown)


GitPaean:
Although there is something not clear, while the PR represents the minimum change to obtain the desired result. 

In the original version, there is one more round well solution, which ideally should improve the result, while it does not make any difference at all.  As a result, I removed the extra round of well solution procedure. More adjustments might be done in the future with new findings. 

In any sense, this PR is an improvement. In the master branch, the explicit quantities stayed to be zero, which is no way correct.

GitPaean:
jenkins build this please

atgeirr (approved):
Looks fine (after some explanation).

PR #1923, opened 2019-06-28
Add UDQ_WCONPROD model to integration testing

joakim-hove:
When https://github.com/OPM/opm-common/pull/870 is merged I consider the UDQ/UDA features sufficiently mature to enable integration testing. Some random plots shown below - more can be provided.

Although I am probably the most trigger happy self-merger around I will *not* push this one in - but I sincerely hope someone else will do!

PS: When this is in I might go and get really drunk - this has been a long haul.

Figure_1
Figure_2
Figure_3
Figure_4

PR #1926, opened 2019-06-29
Add UDQ_WCONPROD model to integration testing

joakim-hove:
See: #1923 which was accidentaly merged - and then immediately reverted.

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -304,6 +304,13 @@ add_test_compareECLFiles(CASENAME multregt_model2
                          REL_TOL ${rel_tol}
                          DIR model2)
 
+add_test_compareECLFiles(CASENAME udq_wconprod
+                         FILENAME UDQ_WCONPROD
+                         SIMULATOR flow
+                         ABS_TOL ${abs_tol}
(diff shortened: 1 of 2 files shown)


bska:
jenkins build this please

bska:
Jenkins says this is okay and I see no reason **not** to expose this to a wider audience. Enjoy your beverage of choice!

joakim-hove:
> Jenkins says this is okay and I see no reason not to expose this to a wider audience. Enjoy your beverage of choice!

Thank you for merging - I must admit I though that some `update_data` magic would be required at some stage in the process, and missing that I halfways assumed the next PR of anything would fail on the UDQ regression test - but that does not seem to be the case?

Maybe all tests are silently passing because the UDQ reference data is missing? 

Sorry @akva2 - could you give a hint here?

akva2:
i'll punt this right to @tskille - this is an unforeseen consequence of compareECL not failing if the reference case does not have a particular file (in this case it has no files at all and compareECL is still happy).

tskille:
this should be fixed  PR #876 in OPM-common.

PR #1927, opened 2019-07-01
fixed: check for existence of reference data paths

akva2:
if not, compareECL will silently not fail and thus
tests do not fail

PR #1928, opened 2019-07-01
Problems starting with group control?

joakim-hove:
The story goes as follows:

1. #1926 was merged to enable integration testing based on the UDQ_WCONPROD testcase. At the time of merging there was no reference data created, but the test machinery actually does not signal failure in this case. 

2. To ensure that jenkins indeed signals failure in the case of a missing reference solution this: #1927 was opened, and the plan was to use this PR to faciliate the necessary `update_data` to create the initial reference data.

3. While working with #1927 it turned out the simulations with the UDQ_WCONPROD testcase did not give reproducible results - some undefined behavior was clearly at play.

With assistance from @akva2 and @GitPaean I have now debugged this and my conclusion is that the well model, specifically the `WellControls` code does not really handle `GRUP` control - at the very least not for wells which *start off*[1] with `GRUP` control. In that case the `controls->current` member remains at the invalid value `-1` - and when that is later used to look up a value in vector undefined behavior is invoked.

I have not tried to actually solve this, in this PR I have just changed to compile the `WellControls` implementation as C++ and added to `throw std::logic_error()` statements which would have caught the undefined behavior we are currently looking at.

**Lessons learned**

1. Use of invalid values like `-1` to signal special state is very dangerous - it will come back to haunt you at some stage.
2. Use of *quantities* - e.g. controls as in this case - as *indicies* is dangerous - that also will come back to haunt you one day.
3. Assuming my hunch below is correct: Carrying state is dangerous - if at all possible make new objects; if new `WellControls` objects were created the issue would have surfaced at the first encounter of `GRUP`, and not only when the well starts with `GRUP` control. 

@akva2 : You might want to merge your fixes in #1927 - and revert the `UDQ` testcase from the regression testing; I do not think I will be able to fix this in short time, so unless someone else comes with a quick fix this will be with us for some time.

[1]: I have not verified this, but my hunch is that when a well is *changed* to `GRUP` control later in a simulation things "work" - because the previous value sticks around.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -32,7 +32,7 @@ list (APPEND MAIN_SOURCE_FILES
   opm/core/wells/WellCollection.cpp
   opm/core/wells/WellsGroup.cpp
   opm/core/wells/WellsManager.cpp
-  opm/core/wells/well_controls.c
+  opm/core/wells/well_controls.cpp
   opm/core/wells/wells.c
   opm/simulators/flow/MissingFeatures.cpp
   opm/simulators/linalg/ExtractParallelGridInformationToISTL.cpp
--- a/opm/core/wells/well_controls.cpp
+++ b/opm/core/wells/well_controls.cpp
@@ -142,7 +142,7 @@ well_controls_create(void)
 {
     struct WellControls *ctrl;
 
-    ctrl = malloc(1 * sizeof *ctrl);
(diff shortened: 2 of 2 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
This needs attention from someone in the know of group control!

joakim-hove:
This has been fixed :-)

Issue #1930, opened 2019-07-02
COMPORD Issue


OPMUSER:
I'm trying to identify why wells are not producing so I'm using the ecl-strict-parsing=true parameter to see if there are any keyword issues. If I use the COMPORD keyword with this setting:

```
-- ---------- ---------- -                                                                                                                                                                                                                                               
--    WELL    COMPLETION                                                                                                                                                                                                                                                 
--    NAME       ORDER                                                                                                                                                                                                                                                   
-- ---------- ---------- -                                                                                                                                                                                                                                               
COMPORD                                                                                                                                                                                                                                                                  
   'GP05    '   'INPUT'  /                                                                                                                                                                                                                                               
/            
```                                                                                                                                                                                                                                                            
I get a message that COMPORD only DEPTH is supported. If I change the setting to:
```
-- ---------- ---------- -                                                                                                                                                                                                                                               
--    WELL    COMPLETION                                                                                                                                                                                                                                                 
--    NAME       ORDER                                                                                                                                                                                                                                                   
-- ---------- ---------- -                                                                                                                                                                                                                                               
COMPORD                                                                                                                                                                                                                                                                  
   'GP05    '   'DEPTH'  /                                                                                                                                                                                                                                               
/                                                                                                                                                                                                                                                                        
```
I get:
`UNSUPPORTED_COMPORD_TYPE: The COMPORD keyword only handles TRACK or INPUT order.`

So is this an inconsistency with ecl-strict-parsing=true or with OPM Flow?

Like I mentioned above, I'm just trying to find out why the wells are not flowing.

OPMUSER:
Update: The inconsistency is with ecl-strict-parsing=true  - OPM Flow terminates if DEPTH is used.

joakim-hove:
> Update: The inconsistency is with ecl-strict-parsing=true - OPM Flow terminates if DEPTH is used.

Yes - fix here: #1931

OPMUSER:
Tusen takk

PR #1931, opened 2019-07-02
Fix error message for COMPORD partially support keyword

joakim-hove:
See: #1930 - fixed the `COMPORD` keyword in the machinery for partially supported keywords.


joakim-hove:
jenkins build this please

PR #1937, opened 2019-07-04
Header file Group.hpp moved to Group/Group.hpp

joakim-hove:
Downstream of https://github.com/OPM/opm-common/pull/881


joakim-hove:
jenkins build this opm-common=881 please

PR #1939, opened 2019-07-08
Remove calls to Schedule::getGroups()

joakim-hove:
Followup to: https://github.com/OPM/opm-common/pull/888


joakim-hove:
jenkins build this opm-common=888 please

PR #1942, opened 2019-07-10
Edge weights method

andrthu:
Add command line option for edge-weight strategy in CpGrid loadBalance method. The option is:

--edge-weights-method=INTEGER  Choose edge-weighing strategy: 0=uniform, 1=trans, 2=log(trans). Default: 1

These options depend on the opm-grid PR: https://github.com/OPM/opm-grid/pull/373.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -75,12 +75,14 @@ NEW_PROP_TAG(EclStrictParsing);
 NEW_PROP_TAG(EclOutputInterval);
 NEW_PROP_TAG(IgnoreKeywords);
 NEW_PROP_TAG(EnableExperiments);
+NEW_PROP_TAG(EdgeWeightsMethod);
 
 SET_STRING_PROP(EclBaseVanguard, IgnoreKeywords, "");
 SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "");
 SET_INT_PROP(EclBaseVanguard, EclOutputInterval, -1); // use the deck-provided value
 SET_BOOL_PROP(EclBaseVanguard, EnableOpmRstFile, false);
 SET_BOOL_PROP(EclBaseVanguard, EclStrictParsing, false);
+SET_INT_PROP(EclBaseVanguard, EdgeWeightsMethod, 1);
 
 END_PROPERTIES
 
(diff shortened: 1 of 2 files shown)


GitPaean:
Maybe you do can do a rebase here and also OPM/opm-grid#373 so the commit history can be cleaner.

ebos/eclbasevanguard.hh
```diff
@@ -443,6 +448,15 @@ public:
...
     const Opm::SummaryState& summaryState() const
     { return summaryState_; }
+
+    /*!
+     * \brief Parameter deciding the edge-weight strategy of the load balancer.
+     */
+    int edgeWeightsMethod() const
+    { return edgeWeightsMethod_; }
+
+    int edgeWeightsMethod()
```
akva2: unnecessary overload.

ebos/eclbasevanguard.hh
```diff
@@ -386,7 +391,7 @@ public:
 
     Opm::Deck& deck()
     { return *deck_; }
-
+
```
akva2: no thanks.

ebos/eclbasevanguard.hh
```diff
@@ -524,7 +538,6 @@ public:
      */
     std::unordered_set<std::string> defunctWellNames() const
     { return std::unordered_set<std::string>(); }
-
```
akva2: no thanks.

akva2:
jenkins build this opm-grid=373 please

blattms:
Looks good. Once the enum is used in OPM/opm-grid#373, would you mind squashing the the commits into just one and making the commit message more insightful, please. The current one ("Add edgeWeightsMethod argument branch. ") is not selfexplanatory. How about "Allow for different edge weight methods during load balancing"?

blattms:
Thanks for the changes. With the introduced enum is there a special reason why we still use int here? It seems better to use the num. (Maybe you need to convert the using `int i=0; EdgeWeightMethod e = EdgeWeightMethod(i)`.

Please also use the defined enum values in the comparsison instead of 1 and 2. That seems more robust
if someone changes opm-grid.

blattms:
Thanks a lot for the changes

blattms:
jenkins build this please

blattms:
jenkins run norne spe please

blattms:
Everything seems fine. Merging.

PR #1945, opened 2019-07-22
Use class Group2 from opm-common

joakim-hove:
Followup to: https://github.com/OPM/opm-common/pull/900

### DIFF
--- a/opm/core/wells/WellCollection.hpp
+++ b/opm/core/wells/WellCollection.hpp
@@ -36,12 +36,12 @@ namespace Opm
     {
     public:
 
-        void addField(const Group2& fieldGroup, size_t timeStep, const PhaseUsage& phaseUsage);
+        void addField(const Group2& fieldGroup, const PhaseUsage& phaseUsage);
 
-        void addWell(const Well2& wellChild, const SummaryState& summaryState, size_t timeStep, const PhaseUsage& phaseUsage);
+        void addWell(const Well2& wellChild, const SummaryState& summaryState, const PhaseUsage& phaseUsage);
 
         void addGroup(const Group2& groupChild, std::string parent_name,
-                      size_t timeStep, const PhaseUsage& phaseUsage);
+                      const PhaseUsage& phaseUsage);
 
         /// Adds the child to the collection
         /// and appends it to parent's children.
--- a/opm/core/wells/WellCollection.hpp
+++ b/opm/core/wells/WellCollection.hpp
@@ -27,7 +27,7 @@
 #include <opm/core/props/phaseUsageFromDeck.hpp>
 
 #include <opm/parser/eclipse/EclipseState/Schedule/Well/Well2.hpp>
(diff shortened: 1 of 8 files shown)


joakim-hove:
This is ready for a review!

tests/test_wellcollection.cpp
```diff
@@ -50,11 +49,11 @@ BOOST_AUTO_TEST_CASE(AddWellsAndGroupToCollection) {
...
 
     // Add groups to WellCollection
-    const auto& fieldGroup =  sched.getGroup("FIELD");
+    const auto& fieldGroup =  sched.getGroup2("FIELD", 2);
     collection.addField(fieldGroup, 2, pu);
 
-    collection.addGroup( sched.getGroup( "G1" ), fieldGroup.name(), 2, pu);
-    collection.addGroup( sched.getGroup( "G2" ), fieldGroup.name(), 2, pu);
+    collection.addGroup( sched.getGroup2( "G1", 2 ), fieldGroup.name(), 2, pu);
+    collection.addGroup( sched.getGroup2( "G2", 2 ), fieldGroup.name(), 2, pu);
```
GitPaean: I think both the `2` is referred to the report step number. Maybe it is possible to remove one of them in the interface?

The second one should not be needed.

joakim-hove: That is probably right - thank you!

opm/core/wells/WellCollection.hpp
```diff
@@ -36,11 +36,11 @@ namespace Opm
     {
     public:
 
-        void addField(const Group& fieldGroup, size_t timeStep, const PhaseUsage& phaseUsage);
+        void addField(const Group2& fieldGroup, size_t timeStep, const PhaseUsage& phaseUsage);
 
         void addWell(const Well2& wellChild, const SummaryState& summaryState, size_t timeStep, const PhaseUsage& phaseUsage);
```
GitPaean: Similar to the Group2, with Well2, the `timeStep` should not be needed here too, while it is used to give a warning message. 

In some way, with the new `Group2` and `Well2`, we already have the information related to group and well for the specified `report step`, and at the same time, we do not know the information about the `report step` through them.  I am not sure what is the best way to keep both information.

opm/core/wells/WellCollection.hpp
```diff
@@ -36,11 +36,11 @@ namespace Opm
...
 
-        void addGroup(const Group& groupChild, std::string parent_name,
+        void addGroup(const Group2& groupChild, std::string parent_name,
```
GitPaean: Please check whether this `timeStep` is needed.

GitPaean:
To me, it looks like pure interface adaptation, the correctness really relies on the upstream PR. 

I only have the comment about whether we need the `report step` in some interfaces. Otherwise, it looks good.

joakim-hove:
I have removed the `report_step` argument from several call sites - that is an improvement!

PR #1955, opened 2019-07-31
using self-copying when calculating well potentials

GitPaean:
The major benefit is that we can simplify the usage of the function `createWellContainer()`. 

This PR tries not to change the simulation behavoir, while it does due to numerical reasons. More investigation of the results are on-going. 

This PR should be only considered as refactoring. Some corrections will be put in later PRs.

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -1442,4 +1442,35 @@ namespace Opm
     }
 
 
+
+
+
+
+    template<typename TypeTag>
+    WellControls*
+    WellInterface<TypeTag>::
+    createWellControlsWithBHPAndTHP(DeferredLogger& deferred_logger) const
+    {
+        WellControls* wc = well_controls_create();
+        well_controls_assert_number_of_phases(wc, number_of_phases_);
+
+        // a well always has a bhp limit
+        const double invalid_alq = -1e100;
+        const double invalid_vfp = -2147483647;
+        const double bhp_limit = this->mostStrictBhpFromBhpLimits(deferred_logger);
+        well_controls_add_new(BHP, bhp_limit, invalid_alq, invalid_vfp, NULL, wc);
+
+        if (this->wellHasTHPConstraints()) {
+            // it might be better to do it through EclipseState?
+            const double thp_limit = this->getTHPConstraint(deferred_logger);
+            const double thp_control_index = this->getControlIndex(THP);
+            const  int vfp_number = well_controls_iget_vfp(well_controls_, thp_control_index);
+            const double alq = well_controls_iget_alq(well_controls_, thp_control_index);
+            well_controls_add_new(THP, thp_limit, alq, vfp_number, NULL, wc);
+        }
+
+        return wc;
+    }
+
+
 }
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -251,7 +251,7 @@ namespace Opm {
             std::vector<bool> is_cell_perforated_;
 
             // create the well container
-            std::vector<WellInterfacePtr > createWellContainer(const int time_step, const Wells* wells, const bool allow_closing_opening_wells, Opm::DeferredLogger& deferred_logger);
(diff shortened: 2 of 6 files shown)


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
The major change in the simulation behavoir from this PR is as follows, 

In the master branch, all the wells in the Wells structure will try to calculate the well potential calculation. 

Wit this PR, the wells got shut due to economic or physical causes will not try to calculate the well potential values. 

I do not know what is the correct behavoir should be. 

The ultimate goal that related to this part is that we should only createWellContainer() only once during each report step, because within the report step, the schedule is not changed. 

There is some suggestions that we should on createWellContainer() once for the whole simulation. It has its advantages and I believe for the SUMMARY output, you might have to do something similar, which is that you need to check the possible well during the whole schedule. And it is also possible good to do this similar thing to the WellState. There will be also some work needs to be done to do this. I do not have an opinion about this at the moment.

GitPaean:
jenkins build this please

GitPaean:
For the failure related to 2D_OW_CTAQUIFER and  2D_FETKOVICHAQUIFER and 2D_OILWATER_POLYMER., with master branch, the well potential calculation could not get converged, while for this PR, it gets converged. 

The more complicated failure is the 3D_WECON.

GitPaean:
failures related to 

```mpi.compareECLFiles_flow+3D_MSW
mpi.compareECLFiles_flow+MSW_MODEL_1
mpi.compareECLFiles_flow+FAULTS_MODEL_1 ```

are related to the small difference in the obtained results, due to the different explicit quantities.

**Comment by GitPaean** [comment]
With this PR, the well potential calculation will consider the  closing and shutting of the well and connections.

With the master branch, the well potential calculation assumes all the wells and the connections are open. 

It is a design problem. If a well is shut, with this PR, it will obtain zero well potential value, with the master branch, it will assume the well is open and trying to obtain the well potential value based on the open status.

**Comment by GitPaean** [comment]
All the failures are investigated, I am removing the WIP status. 

It is a refactoring effort and simplifying the usage of the function createWellContainer and trying to make the well potential calculation more local and less efforts from the BlackoilWellModel. 

In many ways, the concept and data flow is simpler.

File: opm/simulators/wells/WellInterface_impl.hpp [hunk 1]
```diff
@@ -1442,4 +1442,35 @@ namespace Opm
...
 
+
+
+
+
+    template<typename TypeTag>
+    WellControls*
+    WellInterface<TypeTag>::
+    createWellControlsWithBHPAndTHP(DeferredLogger& deferred_logger) const
+    {
```
GitPaean: This function is in this way to make sure this PR is only a refactoring PR to limit the code that actually changes behavoir.  This function will be changed in the following PRs depending how we will use it.

GitPaean:

GitPaean:
jenkins build this opm-tests=153 please

atgeirr:
> It is a design problem. If a well is shut, with this PR, it will obtain zero well potential value, with the master branch, it will assume the well is open and trying to obtain the well potential value based on the open status.

I may be misunderstanding something, but I assumed certain re-opening decisions were based on the potential being positive, and that we should therefore keep calculating it as if open?

GitPaean:
update_data again since we the reference in the opm-tests was updated recently.

GitPaean:

GitPaean:
jenkins build this opm-tests=153 please

GitPaean:

GitPaean:
jenkins build this opm-tests=153 please

GitPaean:
jenkins build this opm-tests=153 please

atgeirr (approved):
I like this way a lot: duplicating the individual well object to solve for potential instead of duplicating the entire collection of them is a nice improvement. Will merge.

totto82:
@atgeirr Be aware that the well model code is rewritten significantly in #1965. Please don't merge to many changes to the well model before considering #1965.

atgeirr:
> Please don't merge to many changes to the well model before considering #1965.

True, sorry if I caused you extra work by merging it. This refactoring should be useful in any case I hope, but the part relating to well control is different.

Issue #1960, opened 2019-08-07
Attributes is not well defined for the cases that only contains Water


GitPaean:
In the function `defineState` of class `SurfaceToReservoirVoidage`, all the values are based on hydrocarbon average. 

When the region does not contain oil or gas, the hydrocarbon volume is zero. As a result, all the values obtained will be NaN. It will enter the simulation and cause more problem. 

Maybe a solution is to use pore volume average instead of hydrocarbon volume average.

PR #1961, opened 2019-08-07
using pore volume to do the average when hydrocarbon is missing

GitPaean:
to avoid NaN resulting from the averaging.

There are some interests in the single phase simulation, NaN values can cause problem when they are used to obtain values through property tables. 

The current patch is working, while I am not sure it is the good way to do it. 

Please suggest.

### DIFF
--- a/opm/simulators/wells/RateConverter.hpp
+++ b/opm/simulators/wells/RateConverter.hpp
@@ -437,7 +437,7 @@ namespace Opm {
                 // and set all attributes to zero
                 const auto& grid = simulator.vanguard().grid();
                 const unsigned numCells = grid.size(/*codim=*/0);
-                std::vector<int> cell2region(numCells, -1);
+                std::vector<RegionId> cell2region(numCells, -1);
                 for (const auto& reg : rmap_.activeRegions()) {
                     for (const auto& cell : rmap_.cells(reg)) {
                         cell2region[cell] = reg;
@@ -449,6 +449,17 @@ namespace Opm {
                     ra.rv = 0.0;
                     ra.pv = 0.0;
 
+                }
+
+                // quantities for pore volume average
+                std::unordered_map<RegionId, Attributes> atrributes_pv;
+
+                // quantities for hydrocarbon volume average
+                std::unordered_map<RegionId, Attributes> atrributes_hpv;
+
+                for (const auto& reg : rmap_.activeRegions()) {
+                    attributes_pv.insert({reg, Attributes()});
+                    attributes_hpv.insert({reg, Attributes()});
                 }
 
                 ElementContext elemCtx( simulator );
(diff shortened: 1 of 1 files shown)


GitPaean:
jenkins build this please

GitPaean:
this closes #1960 .

totto82:
I don't know if this is still relevant. But the code itself looks good. Guards against non-existing oil-phase can be fixed later.

totto82:
jenkins build this please

GitPaean:
jenkins build this please

opm/simulators/wells/RateConverter.hpp
```diff
@@ -482,44 +493,63 @@ namespace Opm {
...
+                        attr.p += fs.pressure(FluidSystem::oilPhaseIdx).value() * hydrocarbonPV;
+                        attr.rs += fs.Rs().value() * hydrocarbonPV;
+                        attr.rv += fs.Rv().value() * hydrocarbonPV;
+                        attr.T += fs.temperature(FluidSystem::oilPhaseIdx).value() * hydrocarbonPV;
+                    }
+
+                    if (pv_cell > 0.) {
+                        auto& attr = attributes_pv[reg];
+                        attr.pv += pv_cell;
+                        attr.p += fs.pressure(FluidSystem::oilPhaseIdx).value() * pv_cell;
```
GitPaean: Any suggestion about how to handle it when the oil phase is not there? I think we did something similar in the onephase simulator case.   Should we do the same here?

atgeirr: Yes we need the same logic. I think we have a consensus that it should be implemented in the FluidState though (and removed from this and other places). Could you add a method `referencePhasePressure()` to the blackoil fluid state (in opm-material), with logic to determine which should be used (that could be an independent PR)? Using it from here would be a good trial.

opm/simulators/wells/RateConverter.hpp
```diff
@@ -451,6 +451,17 @@ namespace Opm {
 
                 }
 
+                // quantities for pore volume average
+                std::unordered_map<RegionId, Attributes> atrributes_pv;
```
atgeirr: I think this version is much better, and avoids the many unnecessary lookups of the previous iteration. Good!

atgeirr (commented):
Apart from dealing with the oil pressure -> reference phase pressure issue, this looks finished to me.

Have you tested on something nontrivial to make sure you get identical computations?

GitPaean:
jenkins build this please

GitPaean:
There is some works suggesting using `sum of the saturation-weighted fluid phase pressures averaged over the pore volume of the reservoir` as the average reservoir pressure. 

This can also be a way to overcome the problem this PR is addressing.

GitPaean:
Looks like this PR can be closed due to the new opening of the PR https://github.com/OPM/opm-simulators/pull/3075 .

PR #1965, opened 2019-08-09
Reimplementation of the well control code

totto82:
Motivation: Make the control code easier to read, modify and extend. Remove the usage of the well_control code from opm-core and instead use the opm-common code directly. 

Missing in this PR:

- [x] Fix THP
- [x] Fix well potential
- [x] Fix initialization
- [x] Fix well efficiency factor 
- [x] Add groups
- [x] Remove old code 
- [x] Clean up and restructure the code 
- [x] fix udq_prod test
- [x] test field models
- [x] support for STOP
- [x] RESV control for production group

Missing functionality in current master that is not included in this PR
- support for GUIDERAT
- guiderates for groups
-  guiderates with different phases
-  fix well potential for thp
-  thp for MSW

PR #1968, opened 2019-08-12
Well potential improvements

GitPaean:
It is after https://github.com/OPM/opm-simulators/pull/1955 .  https://github.com/OPM/opm-simulators/pull/1955 needs to go in first. 

This is a PR to correct some code related to well potential calculation. 

in the original code, when computing the well potential for BHP limit, it is possible that in the well_controls_ there are THP limit. The solution might be obtained not under BHP limit. 

And also, the resetting of the bhp limit is not necessary and is not correct.

It is still implementation based on the code level. It is not aiming to obtain more correct result for any cases.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell.hpp
+++ b/opm/simulators/wells/MultisegmentWell.hpp
@@ -334,11 +334,10 @@ namespace Opm
                          const int perf,
                          std::vector<EvalWell>& mob) const;
 
-        void computeWellRatesWithBhpPotential(const Simulator& ebosSimulator,
-                                              const std::vector<Scalar>& B_avg,
-                                              const double& bhp,
-                                              std::vector<double>& well_flux,
-                                              Opm::DeferredLogger& deferred_logger);
+        void computeWellPotentialsWithBHPLimit(const Simulator& ebosSimulator,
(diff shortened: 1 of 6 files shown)


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

totto82:
This part of the code is completely rewritten in #1965. @GitPaean Please consider taking a look at the refactoring code. I don't think the changes in the PR is relevant if #1965 is merged.

GitPaean:
I have not looked at #1965.  I am not sure the fix in this PR is covered by that PR. 

I am okay to let this PR to wait for now unless it turns out to be in the way of other development.

GitPaean:
It is very likely the opm-data has not been updated in the jenkins test.  I will trigger the test again just to see the results.

GitPaean:
jenkins build this please

GitPaean:
The testing result is expected. It impacts our only test case that involves THP control constraint, which is the major tings that get fixed through this PR.   Other parts are improvements, like renaming, removing unnecessary actions.

GitPaean:
The relevant code is changed so much. This PR is not relevant anymore.  At the beginning, it was fixing a bug existing in the master branch.

PR #1971, opened 2019-08-13
Change testname for UDQ test

joakim-hove:
Resurrected from: #1927. I have been more-than-usually dim when it comes to the arguments in the `tests`directory - would be good if @akva2 could quality check it.

This should be safe to merge right away, the other parts of #1927 - which have been superseded by  https://github.com/OPM/opm-common/pull/876 - can not be merged before #1928 / #1970 has been fully solved.

### DIFF
--- a/tests/update_reference_data.sh
+++ b/tests/update_reference_data.sh
@@ -29,8 +29,15 @@ copyToReferenceDir () {
 }
 
 declare -A tests
-# binary dirname casename [testname]
-# you only have to specify testname if it differs from dirname
+# The tests are listed in a dictionary mapping name of the test to the commands
+# used to run the test:
+#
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

tests/update_reference_data.sh
```diff
@@ -29,9 +29,15 @@ copyToReferenceDir () {
 }
 
 declare -A tests
-# binary dirname casename [testname]
-# you only have to specify testname if it differs from dirname
-tests[spe1]="flow spe1 SPE1CASE1"
```
atgeirr: Isn't this missing now?

akva2: good catch.

joakim-hove: Yes - that was good; I have fixed that now.

PR #1975, opened 2019-08-15
throw when calling addWellContributions() for MSW

GitPaean:


GitPaean:
jenkins build this please

GitPaean:
With the original code, the MSW will call an empty function without any notifying message. This PR is to give some warning message for the people to test CPR with MS well model. 

It is a small PR and it should not break anything in any unnecessary way.  I am merging it now. 

If it turns out it is the bad way to throw the exception, it needs to refactored along other `OPM_THROW`s.

PR #1976, opened 2019-08-15
[WIP] Prediction continue rebase

GitPaean:
Trying to use a different way to calculate the inflow performance relationships (IPR) and different way to do the physical operability testing. 

A good thing about this approach is that we do not try to obtain physical formulation of the IPR any more, which is very hard for MS wells.  Hopefully, with this way, we can handle the physical limit test with MS wells more easily. 

The bad thing is that when we could not get converged solution, which can be due to various causes, we have to shut the well. 

It is in the WIP state. More work needs to be done for the THP control


GitPaean:
With the big changes in the Well control related code and the major idea of this PR has been incorporated in other PRs in various forms.  This PR is not that relevant anymore even might be revisited later.

PR #1980, opened 2019-08-19
Setup logging before the simulator is initialized

joakim-hove:
[This](https://github.com/OPM/opm-common/pull/933) was created and merged to put messages like:

```
    6 Reading GAS      from: /path/udq_actionx/UDQ_WCONPROD.DATA:45
```
in the `.PRT` file - but it turned out that no messages appeared in the file, and as debugged by @attgeir - that was because the logging system was not initialized when the parsing was performed. This PR shuffles things a bit around, to facilitate initialization of the logging system before the parsing starts. It even works on Atgeirrs machine.

NB: this endevaour has revealed that *no* log messages originating from the opm-common setup process have been logged!

### DIFF
--- a/flow/flow.cpp
+++ b/flow/flow.cpp
@@ -116,12 +116,21 @@ namespace detail
 
 }
 
+enum class FileOutputMode {
+    //! \brief No output to files.
+    OUTPUT_NONE = 0,
+    //! \brief Output only to log files, no eclipse output.
+    OUTPUT_LOG_ONLY = 1,
+    //! \brief Output to all files.
+    OUTPUT_ALL = 3
+};
+
 
 
 
 
 // Setup the OpmLog backends
-void setupLogging(int mpi_rank_, const std::string& deck_filename, const std::string& cmdline_output_dir, const std::string& cmdline_output, bool output_cout_, const std::string& stdout_log_id) {
+FileOutputMode setupLogging(int mpi_rank_, const std::string& deck_filename, const std::string& cmdline_output_dir, const std::string& cmdline_output, bool output_cout_, const std::string& stdout_log_id) {
     // create logFile
     using boost::filesystem::path;
     path fpath(deck_filename);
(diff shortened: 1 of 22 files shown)


atgeirr:
> I have not investigated this further - maybe it is freaking obvious to someone?

It's not obvious, but it should not be too much debugger work I think to figure out what is wrong. I'll take a look at it later today.

atgeirr:
Having looked at this, the cause is obvious (asking for the IOConfig of the EclipseState of the EclVanguard before any of these exist), but the cure is not, and will require some refactoring. One way could be to disentangle the output dir functionality from the IOConfig, as it does not depend on the content of the deck but only its filesystem location.

joakim-hove:
> One way could be to disentangle the output dir functionality from the IOConfig, as it does not depend on the content of the deck but only its filesystem location.

That seems like a natural approach to me.

joakim-hove:
OK - this should be ready for review and merge now!

joakim-hove:
jenkins build this please

opm/simulators/flow/FlowMainEbos.hpp
```diff
@@ -306,101 +299,6 @@ namespace Opm
...
-                output_ = outputModeIt->second;
-            }
-            else {
-                output_ = OUTPUT_ALL;
-                std::cerr << "Value " << outputModeString <<
-                    " is not a recognized output mode. Using \"all\" instead."
-                          << std::endl;
-            }
-
-            output_cout_ = false;
```
atgeirr: The `output_cout_` and `output_to_files_` variables must still be initialized properly in this class.

opm/simulators/flow/FlowMainEbos.hpp
```diff
@@ -70,22 +66,21 @@ SET_INT_PROP(EclFlowProblem, OutputInterval, 1);
 
 END_PROPERTIES
 
+enum FileOutputMode {
```
atgeirr: This is only used in flow.cpp now, so it can be moved there.

flow/flow.cpp
```diff
@@ -112,6 +116,104 @@ namespace detail
...
+        // Added rank to log file for non-zero ranks.
+        // This prevents message loss.
+        debugFileStream << "." << mpi_rank_;
+        // If the following file appears then there is a bug.
+        logFileStream << "." << mpi_rank_;
+    }
+    logFileStream << ".PRT";
+    debugFileStream << ".DBG";
+
+    FileOutputMode output_;
```
atgeirr: Some testing with master reveals that this feature is no longer respected. The intention was to be able to run completely output-less, or only with log files, for profiling or other purposes. Running the current master, the eclipse files are always written. Using `--output-mode=none` only turns off the log files, the terminal is still written to. If the last thing is a bug could be discussed, I think "none" should be really nothing, but perhaps we should then also add a "OUTPUT_TERMINAL_ONLY" mode.

The last change would be good to tack onto this PR, which is why I discuss it here. It is in no way this PR's fault that the current master is not behaving properly... Ensuring we turn off eclipse file output when requested to may be a little more involved, or it may be easy, I don't know right now.

opm/simulators/flow/FlowMainEbos.hpp
joakim-hove: Yes - and that inspired a quite significant reworking of this PR!

opm/simulators/flow/FlowMainEbos.hpp
joakim-hove: OK

opm/simulators/flow/FlowMainEbos.hpp
```diff
@@ -627,11 +525,10 @@ namespace Opm
         int  mpi_rank_ = 0;
         int  mpi_size_ = 1;
         bool output_cout_ = false;
-        FileOutputMode output_ = OUTPUT_ALL;
+        //FileOutputMode output_ = OUTPUT_ALL;
```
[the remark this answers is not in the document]
joakim-hove: ok

joakim-hove:
OK - I have addressed @atgeirr's comments in a rather large fixup commit. The change I have done here is to extract two boolean flags out of the `TypeTag` controlled `FlowMainEbos` and just pass them as regular boolean flags. Personally I think this is a development in the Right Direction&trade; - in the sense that it is an &epsilon; change in the direction of looser coupling and more explicit/transparent code; but I understand that others might disagree - and certainly the amount change is quite significant compared to the gain - I will not fight this change!

flow/flow.cpp
joakim-hove: I will come back to this in a subsequent commit.

flow/flow.cpp
```diff
@@ -220,7 +338,7 @@ int main(int argc, char** argv)
             }
         }
         const auto& phases = Opm::Runspec(*deck).phases();
-
+        bool outputFiles = (outputMode != FileOutputMode::OUTPUT_NONE);
```
atgeirr: I think the direction of this (passing the output bools as arguments) is fine, but on this line outputMode is uninitialized if mpiRank != 0.

joakim-hove: OK - should be fixed now.

atgeirr:
jenkins build this please

joakim-hove:
And what happened here for f... sake?!

joakim-hove:
OK - I have fixed the two warnings; but why the tests fail I do not understand yet. Fails on my machine as well :-(

GitPaean:
It looks like all the cases that failed are either two phase case, or there is some other component or energy equation involved.

joakim-hove:
OK - I have looked briefly into this; seems to require the 101 course on TypeTags++ - which I have not passed yet. At least the polymer related failures terminate with the exception:
```C++
Program threw an exception: Polymer treatment disabled at compile time, but the deck contains the POLYMER keyword
```
When looking at the code it seems the correct `flowEbosxxxxMain()` function is invoked; but possibly the subsequent instantiation gets wrong? Frustrating .... ;-(

flow/flow_ebos_energy.cpp
```diff
@@ -62,8 +62,8 @@ int flowEbosEnergyMain(int argc, char** argv)
     Dune::MPIHelper::instance(argc, argv).rank();
 #endif
 
-    Opm::FlowMainEbos<TTAG(EclFlowEnergyProblem)> mainfunc;
-    return mainfunc.execute(argc, argv);
+    Opm::FlowMainEbos<TTAG(EclFlowProblem)> mainfunc;
```
atgeirr: You changed the typetags here. A little aggressive copy-paste... Also for foam, two-phase etc. I'll check better the next time!

joakim-hove: Ok thanks!

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
OK - this is green now - I will merge. Do note that you will now get *two* listings of the type:
```txt
23: RUNSPEC  /path/to/case/CASE.DATA:67
24: GRID          /path/to/case/grid.grdecl:72 
```
that is because `flow`actually parses everything twice - nothing new with that; but it has become very visible with this PR.

PR #1981, opened 2019-08-21
Remove COMPDAT from list of unsupported keywords

joakim-hove:
The `COMPDAT`keyword had sneaked in on the list of unsupported keywords.


joakim-hove:
jenkins build this please

PR #1983, opened 2019-08-22
Flexible solver 

hnil:
This pull request add:
- trueimpes (this has a little over head if solver is recreated very often but this is not the intension of this solver.) 
- corrects citeria for amg agglomeration
- put resonable hardcoded defaults for cpr solver

it include pullrequest #1979


blattms:
Somehow this does not comile for me (at least with master) and most probably for others neither:
```
/home/mblatt/src/dune/opm/opm-simulator/opm/simulators/linalg/matrixmarket_ewoms.hh:172:10: error: ‘mm_multipliers’ is not a class template
   struct mm_multipliers<BCRSMatrix<Opm::MatrixBlock<B,i,j>,A> >
```

It seems at a state from before the renaming of ewoms and the moving of code to opm-simulators. (e.g.
ewoms/linalg/matrixblock.hh is now opm/simulators/linalg/matrixblock.hh.

I will do a complete review once I can compile. For now I have one suggestion:
matrixmarket_ewoms.hh contains a lot of code which might be a duplicate from somewhere else.
The reason is that we are using `MatrixBlock` instead of `FieldMatrix`. Since the former is just a subclass without additional member, one could remove matrixmarket_ewoms.hh completely, and when
calling any matrixmarket functionality use a `reinterpret_cast`:
```
template<class B, class, A, int m, n>
const BCRSMatrix<FieldMatrix<B,n,m>,A>& reinterpretToFieldMatrix(BCRSMatrix<const Ewoms::MatrixBlock<T,i,j>,A>& mat)
{
    reinterpret_cast<const BCRSMatrix<FieldMatrix<B,n,m>,A>&>(mat);
}
....
Dune::writeMatrixMarket(reinterpretToFieldMatrix(matrix), filem);
```

blattms:
Please forget my comment about the code moving and renaming. I have no idea what I looked at since this addressed in 544e7d3

opm/simulators/linalg/BlackoilAmg.hpp
```diff
@@ -553,6 +554,13 @@ class OneStepAMGCoarseSolverPolicy
     CoarseLevelSolver* createCoarseLevelSolver(LTP& transferPolicy)
     {
         coarseOperator_=transferPolicy.getCoarseLevelOperator();
+	if( param_->cpr_solver_verbose_ >10 ){
```
blattms: indentation level seems wrong.

opm/simulators/linalg/BlackoilAmg.hpp
```diff
@@ -553,6 +554,13 @@ class OneStepAMGCoarseSolverPolicy
...
+	    std::string filename = "course_matrix_istl_blackoil.txt";
+	    std::ofstream filem(filename);
+	    //const auto& matrix = *transferPolicy.getCoarseMatrix();
+	    const auto& matrix = coarseOperator_->getmat(); 
+	    Dune::writeMatrixMarket(matrix, filem);
```
blattms: Like proposed in another comment we could do a reinterpret_cast to `BCRSMatrix<FieldMatrix>` here.

opm/simulators/linalg/FlexibleSolver.hpp
```diff
@@ -130,6 +130,14 @@ class FlexibleSolver : public Dune::InverseOperator<VectorTypeT, VectorTypeT>
                                                               tol, // desired residual reduction factor
                                                               maxiter, // maximum number of iterations
                                                               verbosity));
+	} else if (solver_type == "cg") {
+            linsolver_.reset(new Dune::CGSolver<VectorType>(*linearoperator_,
+							    *scalarproduct_,
```
blattms: indentation seems messed up

opm/simulators/linalg/FlexibleSolver.hpp
```diff
@@ -159,6 +167,9 @@ class FlexibleSolver : public Dune::InverseOperator<VectorTypeT, VectorTypeT>
     {
         initOpPrecSp(matrix, prm, comm);
         initSolver(prm);
+	if(prm.get<int>("verbosity")>10){
```
blattms: indentation.

opm/simulators/linalg/FlowLinearSolverParameters.hpp
```diff
@@ -162,6 +164,7 @@ namespace Opm
         bool   use_cpr_;
         std::string system_strategy_;
         bool scale_linear_system_;
+	std::string linear_solver_configuration_;
```
blattms: This will be my last comment about indentation as it seems to messed up quite a lot. Please try to fix this even if I will not
comment on it any more.

opm/simulators/linalg/ISTLSolverEbosFlexible.hpp
```diff
@@ -128,7 +147,43 @@ class ISTLSolverEbosFlexible
...
+	    prm_.get<std::string>("preconditioner.type") == "cprt"
+	    )
+	{	    
+	    bool transpose = false;
+	    if(prm_.get<std::string>("preconditioner.type") == "cprt"){
+		transpose = true;
+	    }
+	    if(prm_.get<std::string>("preconditioner.weight_type") == "quasiimpes") {
+		if(prm_.get<int>("verbosity") > 10){
+		    std::cout << "Using quasiimpes" << std::endl;
```
blattms: Left over debugging? Otherwise maybe use or logging?

opm/simulators/linalg/ISTLSolverEbosFlexible.hpp
```diff
@@ -200,6 +256,46 @@ class ISTLSolverEbosFlexible
...
+    		}
+    	    }
+    	    BlockVector bweights;
+    	    MatrixBlockType block_transpose = Opm::transposeDenseMatrix(block);
+    	    block_transpose.solve(bweights, rhs);
+    	    bweights /= 1000.0; // given normal densities this scales weights to about 1.
+    	    weights[index] = bweights;
+    	    ++index;
+    	}
+    	return weights;
```
blattms: Does the computation in the above loop need values for cell neighbors? If that is the case, then we are missing communication when running parallel.

opm/simulators/linalg/WriteSystemMatrixHelper.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#ifndef OPM_WRITESYSTEMMATRIXHELPER_HEADER_INCLUDED
+#define OPM_WRITESYSTEMMATRIXHELPER_HEADER_INCLUDED
+//#include <dune/istl/matrixmarket.hh>
+//#include <ewoms/linear/matrixblock.hh>
+#include "matrixmarket_ewoms.hh"
+namespace Opm{
```
blattms: Seems like we might be missing some boost headers here.

opm/simulators/linalg/getTrueImpesWeights.hpp
```diff
@@ -0,0 +1,131 @@
...
+                    }
+                }
+                BlockVector bweights;
+                MatrixBlockType block_transpose = Opm::transposeDenseMatrix(block);
+                block_transpose.solve(bweights, rhs);
+                bweights /= 1000.0; // given normal densities this scales weights to about 1.
+                weights[index] = bweights;
+                ++index;
+            }
+            return weights;
```
blattms: Please double check that the computation in the above loop does not depend on values of the cell neighors. If it does the we might need some communication before returning the weights.

CMakeLists.txt
```diff
@@ -171,6 +171,14 @@ opm_add_test(flow_blackoil_dunecpr
   DEPENDS "opmsimulators"
   LIBRARIES "opmsimulators")
 
+
+opm_add_test(flow_blackoil_dunecpr_noflex
+  ONLY_COMPILE
+  DEFAULT_ENABLE_IF ${FLOW_DEFAULT_ENABLE_IF}
+  SOURCES flow/flow_blackoil_dunecpr_noflex.cpp
+  EXE_NAME flow_blackoil_dunecpr_noflex
+  CONDITION NOT MPI_FOUND)
```
blattms: I have the feeling that some lines went missing during rebasing or merging. Shouldn' there be something like
```
DEPENDS "opmsimulators"
  LIBRARIES "opmsimulators"
```

I am also not 100% sure what `CONDITION NOT MPI_FOUND` means. Does it mean cpr is only available if MPI is not found? That would be a bummer.

opm/simulators/linalg/OwningTwoLevelPreconditioner.hpp
```diff
@@ -130,10 +157,16 @@ class OwningTwoLevelPreconditioner : public Dune::PreconditionerWithUpdate<Vecto
...
     {
-        Opm::Amg::getQuasiImpesWeights<MatrixType, VectorType>(
-            linear_operator_.getmat(), prm_.get<int>("pressure_var_index"), transpose, weights_);
+	
+	// if(prm.get<std::string>("weight_type") == "quasiimpes"){
+	//     Opm::Amg::getQuasiImpesWeights<MatrixType, VectorType>(
+	// 	linear_operator_.getmat(), prm_.get<int>("pressure_var_index"), transpose, weights_);
+	//     weights_ = prm.get<VectorType>("weights");
+	// }else{
+	    weights_ = weights;
```
blattms: Can someone please tell me, why this change make sense?

If I am not mistaking, we assume that even if the matrix is changing so much, that there definitely is a need to recalculate the preconditioner, we, nevertheless assume that the absolute maximum entry of x, solving e.g. `a_{ii} x = [ 0 1 0]^T` for any row `i` stays the same.

Previously, we always recalculated the weights. That seemed to make more sense.

What am I missing?

blattms: Ok. I misinterpreted this. The weights do get recalculate, but this is a bit obfuscated as it is moved to ISTLSolverEbosFlexible. I am not smart enough for this distribution of code.

Now the question is why does this code moval makes sense,

blattms: The reason might be that calculating the true impes weights needs the grid and information from opm-models and that should not be exposed to the constructors.

I do not agree 100%  here because the update method now gets passed information that is specific to some preconditioners. Will need to think, whether there is a better way.

blattms:
Another question:

If for e queried key there is no value set in property tree, then what happens? Would e.g. `prm.get<int>("pre_smooth")` return zero in this case? Or is there a default value set somewhere?

blattms:
Testing seems to reveal that unset properties do not have reasonable defaults. The previously proposed json setting will not work due to this. That means we need to resort to using the get method with a default option.

blattms:
And of course the simulator will do happy time step chopping as it thinks the linear system is at fault...

opm/simulators/linalg/BlackoilAmg.hpp
hnil: I think if this workes it wold be nice. It is just messy and a copy of the code from matrixmarket.hh. But it is very convenient to be able to write out the matrices from OPM.

opm/simulators/linalg/ISTLSolverEbosFlexible.hpp
hnil: Correct logging is the way to go.

opm/simulators/linalg/ISTLSolverEbosFlexible.hpp
hnil: The weights a calculated only based on the diagonal block quasiimpes or using the blockmatrix generated from the accumulation term trueimpes

opm/simulators/linalg/getTrueImpesWeights.hpp
hnil: As long the the diagonal of the matrix is calculated correctly for the owning rows I think it is correct.

opm/simulators/linalg/OwningTwoLevelPreconditioner.hpp
hnil: It some time ago I did this but. To make the liner solver independent of simulators. We need to pass in the new weights alternatively using store a reference to the vector and update it from the simulator. Quasimpes on the other hand can recalculate it from the matrix alone. I changed the update to have  both the weights and the parameters as input at some point to make it possible to do more advanced thing in the update. For example update in a particular way given information from the simulator (of coarse the full strategy for this could be in parameter tree used for the construction). The bad thing is that the update now has an interface with weights which specific for one linear solver.

hnil:
May aim was to not have default values at all. So that anything should be set in the json file.

hnil:
Do we have way of aborting the run of unrecoverable runtime errors in the linear solver.

blattms:
I do not think that such json files (without any defaults) are manageable for mere mortals like me. They are just too large. (one of the best things of this PR are IMHO the preconfigured configurations).

In the long term, I think the properties should also end up in the log files.

Please see #2498

blattms:
I am closing this that i did not miss anything of this one in  #2498

PR #1986, opened 2019-08-27
Adapt to enum changes

joakim-hove:
Followup to: https://github.com/OPM/opm-common/pull/972

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1686,7 +1686,7 @@ public:
         const auto& simulator = this->simulator();
         int epsiodeIdx = std::max(simulator.episodeIndex(), 0);
         const auto& oilVaporizationControl = simulator.vanguard().schedule().getOilVaporizationProperties(epsiodeIdx);
-        return (oilVaporizationControl.getType() == Opm::OilVaporizationEnum::VAPPARS);
+        return (oilVaporizationControl.getType() == Opm::OilVaporizationProperties::OilVaporization::VAPPARS);
     }
 
     bool nonTrivialBoundaryConditions() const
--- a/ebos/ecltracermodel.hh
+++ b/ebos/ecltracermodel.hh
@@ -485,14 +485,14 @@ protected:
         const auto& wells = simulator_.vanguard().schedule().getWells2(episodeIdx);
         for (const auto& well : wells) {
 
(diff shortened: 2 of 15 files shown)


joakim-hove:
jenkins build this opm-common=972 please

joakim-hove:
jenkins build this opm-common=972 please

joakim-hove:
jenkins build this opm-common=972 please

joakim-hove:
jenkins build this opm-common=972 please

bska:
```
opm/simulators/wells/WellInterface.hpp:32:70: fatal error: opm/parser/eclipse/EclipseState/Schedule/ScheduleEnums.hpp: No such file or directory
```
You've only updated `WellInterface_impl.hpp` in this PR.

joakim-hove:
> You've only updated WellInterface_impl.hpp in this PR.

Hmm - yes you are right; I *have* updated the `WellInterface.hpp` as well - but I must have missed it in git.

joakim-hove:
jenkins build this opm-common=972 please

joakim-hove:
Is master all green now?

akva2:
no. jenkins bump to dune 2.6 has left a few regtest broken + a unit test in opm-grid (i have opened a pr for that one).

joakim-hove:
Ok - thank you.

akva2:
the regtests that broke are due to uninitialized data output by libecl. since this code will be replaced soonish, i suggest we just update and move on for now.

GitPaean:
> mpi.compareECLFiles_flow+SPE1CASE2_OILGAS
> mpi.compareECLFiles_flow+2D_OILWATER_POLYMER

The change of DUNE version can break the two phase case due to different algorithms of inverting a small matrix.

akva2:
i see, so there may be two things at play then.

atgeirr:
> The change of DUNE version can break the two phase cases due to different algorithms of inverting a small matrix.

True, but in this case there have also been observed some valgrind anomalies from libecl. Both sources could contribute.

atgeirr:
I think we should just do an update_data now, but not let the valgrind observation die either.

akva2:
https://github.com/equinor/libecl/pull/651 takes care of the valgrind issue.

joakim-hove:
jenkins build this opm-common=972 please

PR #1994, opened 2019-09-06
Keywords w

joakim-hove:
Missing features on W


joakim-hove:
Sorry - that was the wrong button; should have run Jenkins first. But it is a quite trivial change and it works on my machine; so I guess no real harm done.

atgeirr:
> But it is a quite trivial change and it works on my machine; so I guess no real harm done.

The post-build will be the judge of that... :-)

PR #2018, opened 2019-09-20
Pass sim start argument to SummaryState constructor

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/1030


bska:
Merging into master to account for upstream API change.  Thanks a lot!

PR #2019, opened 2019-09-21
Change testcode to use Parser API with dimensions in ParserKeyword

joakim-hove:
Followup to: https://github.com/OPM/opm-common/pull/1021. NB - this can be merged before https://github.com/OPM/opm-common/pull/1021

NB: Observe that *iff* the combination https://github.com/OPM/opm-common/pull/1035 + #2020 is merged first (which I would recommend ....) this PR should be closed and forgotten.

PR #2020, opened 2019-09-21
Use new EquilRecord from opm-common

joakim-hove:
Followup to: https://github.com/OPM/opm-common/pull/1035

See alternative: #2019

### DIFF
--- a/ebos/equil/equilibrationhelpers.hh
+++ b/ebos/equil/equilibrationhelpers.hh
@@ -34,8 +34,6 @@
 #include <opm/material/fluidstates/SimpleModularFluidState.hpp>
 #include <opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp>
 
-// due to a bug in Equil.hpp, cstddef must be included before Equil.hpp
-#include <cstddef>
 #include <opm/parser/eclipse/EclipseState/InitConfig/Equil.hpp>
 
 #include <memory>
--- a/tests/test_equil.cc
+++ b/tests/test_equil.cc
@@ -161,57 +161,7 @@ static Opm::EquilRecord mkEquilRecord( double datd, double datp,
                                        double zwoc, double pcow_woc,
                                        double zgoc, double pcgo_goc )
 {
-    using namespace Opm;
-
-    DeckItem dd( "datdep", double() );
-    dd.push_back( datd  );
-    Opm::Dimension dd_dim( "dddim", 1 );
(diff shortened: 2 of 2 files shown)


joakim-hove:
I would like to merge this and:  OPM/opm-common#1035 - comments?

bska:
> I would like to merge this and: [OPM/opm-common#1035](https://github.com/OPM/opm-common/pull/1035) - comments?

Just one.  The new `EquilRecord` constructor has the signature
```C++
EquilRecord(double, double, double, double, double, double, bool, bool, int);
```
I worry that that will be very hard to use correctly and very easy to use incorrectly.  At minimum, all clients/call-sites will have to refer very closely to the declaration of `EquilRecord` to ensure that object is correctly formed.

How many locations do we expect to create `EquilRecord` objects from?

joakim-hove:
> How many locations do we expect to create EquilRecord objects from?

*Two*; the main location is [here]( https://github.com/OPM/opm-common/blob/master/src/opm/parser/eclipse/EclipseState/InitConfig/Equil.cpp) - that should ideally have been private - and the only `EquilRecord` instantiaton - but in addition there is one instantiation here in this test. This test is somewhat awkward because the test is created directly with SI data (i.e. you can not easily embed a test deck and parse it), and at the same time it uses quite low level structures.  

I agree that the current signature is easy to get wrong, but it is at least not any worse than the current approach based on manually crafted `DeckRecord` instance with "wrong" dimensions.

joakim-hove:
> Maybe we can use this opportunity to actually make it private? E.g.,

Yes - that looks like good solution.

> I think that might even solve the "embed testable dataset" problem.

I don't really see how that will help with the testdata problem; the testdata problem is not particularly large - it is *just* to create a deck with inputs manually. I'll do that.

joakim-hove:
OK - I'll close this now; and instead redirect to #2019 - which is kind of funny, but gets the job done.

joakim-hove:
> OK - I'll close this now; 

I'll reopen this and beg on my knees that this can be merged as is. I agree it is not very nice; I still think it is an improvement of the current state - and something must be done in order to merge OPM/opm-common#1021.

The alternative approach in #2020 failed in mysterious ways; think it was due to multiple instances of the `ebos/vanguard` structure - something I was not exactly thrilled by going into.

atgeirr:
I agree this is still nicer than the master, so I'll merge as requested.

PR #2021, opened 2019-09-24
Work in Progress: Python

jmsargado:
Preliminary work for python bindings to blackoil simulator. Entire pybind11 folder was copied to opm-simulators due to linking problems with the one in opm-common. @akva2 Any tips on how to get it to work with pybind11 from opm-common?


joakim-hove:
As mentioned in [this comment](https://github.com/OPM/opm-simulators/pull/2021#discussion_r331485231) the Python code must be updated due to changes in opm-common. Would be good if you could make the suggested changes and confirm that it works - or not?

I am not the main hitman in this repository, but I would suggest splitting this PR in three:

1. Add the pybind11 stuff - without making any changes to the remaining part of the opm-simulators repo.
2. Make the necessary changes to the opm-simulators C++ code - and the necessary pybind11 bindings.
3. Add the pure python example which uses the Python code from opm-common.

For points 1 & 2 my suggestion is to do things as possible as simple, but in the not-so-distant future it might be beneficial to make changes both to the elements covered by 1 & 2 and the corresponding elements in opm-common to make integration/reuse+++ as smooth as possible, but that is not for now.

jmsargado:
I've updated the simulator bindings as well as the python code to use the current master branches from the core modules. Following the above suggestion, this has now been split into 3 new PRs:
1. Add pybind11 #2126
2. Python blackoil simulator #2127
3. Python example #2129

PR #2022, opened 2019-09-25
Make it possible to test driftCompensation i Flow

totto82:
Default is still to not use it, i.e this does not change anything.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -337,7 +337,8 @@ SET_TYPE_PROP(EclBaseProblem, NewtonMethod, Opm::EclNewtonMethod<TypeTag>);
 SET_INT_PROP(EclBaseProblem, RestartWritingInterval, 0xffffff); // disable
 
 // Drift compensation is an experimental feature, i.e., systematic errors in the
-// conservation quantities are only compensated for if experimental mode is enabled.
+// conservation quantities are only compensated for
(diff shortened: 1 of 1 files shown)


ebos/eclproblem.hh
```diff
@@ -337,7 +337,8 @@ SET_TYPE_PROP(EclBaseProblem, NewtonMethod, Opm::EclNewtonMethod<TypeTag>);
 SET_INT_PROP(EclBaseProblem, RestartWritingInterval, 0xffffff); // disable
 
 // Drift compensation is an experimental feature, i.e., systematic errors in the
-// conservation quantities are only compensated for if experimental mode is enabled.
+// conservation quantities are only compensated for
+// as default if experimental mode is enabled.
```
atgeirr: I think this comment is wrong: EnableExperiments and EclEnableDriftCompensation seem independent now, to me?

atgeirr:
jenkins build this please

ebos/eclproblem.hh
totto82: With this PR Flow defaults EclEnableDriftCompensation to false while Ebos to true. (Since EnableExperiments is true in Ebos) So not completely independent.

atgeirr: I did not notice that the default was still set as before, so you are of course correct.

PR #2023, opened 2019-09-25
Fix solvent restart

totto82:
Ask for SSOLVENT from the restart files if solvent model. 
Without this fix solvent saturation will always be zero after restart.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -2492,12 +2492,14 @@ private:
             eclWriter_->eclOutputModule().initHysteresisParams(simulator, elemIdx);
             eclWriter_->eclOutputModule().assignToFluidState(elemFluidState, elemIdx);
 
-            processRestartSaturations_(elemFluidState);
+            if (enableSolvent)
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
@atgeirr Any reasons for not merging this.

ebos/eclproblem.hh
```diff
@@ -2548,7 +2550,7 @@ private:
         eclWriter_->endRestart();
     }
 
-    void processRestartSaturations_(InitialFluidState& elemFluidState)
+    void processRestartSaturations_(InitialFluidState& elemFluidState, size_t elemIdx)
```
atgeirr: Before, processRestartSaturations_ could be a const function, it only changed the input elemFluidState. Now it also changes the solventSaturation_ member. I think it would be better to take a `double&` reference here and pass `solventSaturation_[elemIdx]` in the calling code rather than passing `elemIdx` (assuming changing InitialFluidState to contain solvent is a non-starter).

totto82:
jenkins build this please

ebos/eclproblem.hh
totto82: Done!

PR #2033, opened 2019-10-01
Avoid using global grid (equilGrid) on non-IO processes.

blattms:
This is needed for OPM/opm-grid#389. With that change the global grid on all processes but the one
with rank 0 (IO process) will be empty and hence they do not have the global grid available anymore (or more precise it is a grid without active cells). Any use of it is most likely will result in an error.  This PR should remove the use of the global grid on all but the IO rank.

Please not that with this Schedule::filterConnections is only called on the IO process since for the other
processes it will remove all well connections. This PR needs to be double checked with all well models,
At least for the traditional one it seems to work.

This PR should work stand alone.

### DIFF
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -163,8 +163,15 @@ public:
             // transmissibilities are relatively expensive to compute, we only do it if
             // more than a single process is involved in the simulation.
             cartesianIndexMapper_ = new CartesianIndexMapper(*grid_);
-            globalTrans_ = new EclTransmissibility<TypeTag>(*this);
-            globalTrans_->update();
+            if (grid_->size(0))
+            {
+                globalTrans_ = new EclTransmissibility<TypeTag>(*this);
+                globalTrans_->update();
+            }
+            else
+            {
+                globalTrans_ = nullptr;
+            }
 
             Dune::EdgeWeightMethod edgeWeightsMethod = this->edgeWeightsMethod();
 
@@ -253,7 +260,10 @@ public:
     { return defunctWellNames_; }
 
     const EclTransmissibility<TypeTag>& globalTransmissibility() const
-    { return *globalTrans_; }
+    {
+        assert( globalTrans_ != nullptr );
+        return *globalTrans_;
+    }
 
     void releaseGlobalTransmissibility()
     {
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -120,9 +120,11 @@ public:
      */
     const EquilGrid& equilGrid() const
     {
+#if HAVE_MPI
         int mpiRank;
         MPI_Comm_rank(MPI_COMM_WORLD, &mpiRank);
         assert(mpiRank == 0);
+#endif
         return *equilGrid_;
     }
 
@@ -238,9 +240,11 @@ public:
      */
     const CartesianIndexMapper& equilCartesianIndexMapper() const
     {
+#if HAVE_MPI
         int mpiRank;
         MPI_Comm_rank(MPI_COMM_WORLD, &mpiRank);
         assert(mpiRank == 0);
+#endif
         return *equilCartesianIndexMapper_;
     }
 
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -119,7 +119,12 @@ public:
      * same as the grid which is used for the actual simulation.
      */
     const EquilGrid& equilGrid() const
-    { return *equilGrid_; }
+    {
+        int mpiRank;
+        MPI_Comm_rank(MPI_COMM_WORLD, &mpiRank);
+        assert(mpiRank == 0);
+        return *equilGrid_;
+    }
 
     /*!
      * \brief Indicates that the initial condition has been computed and the memory used
@@ -232,7 +237,12 @@ public:
      * \brief Returns mapper from compressed to cartesian indices for the EQUIL grid
      */
     const CartesianIndexMapper& equilCartesianIndexMapper() const
-    { return *equilCartesianIndexMapper_; }
+    {
+        int mpiRank;
+        MPI_Comm_rank(MPI_COMM_WORLD, &mpiRank);
+        assert(mpiRank == 0);
+        return *equilCartesianIndexMapper_;
+    }
 
     std::unordered_set<std::string> defunctWellNames() const
     { return defunctWellNames_; }
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -244,6 +244,7 @@ public:
         int mpiRank;
         MPI_Comm_rank(MPI_COMM_WORLD, &mpiRank);
         assert(mpiRank == 0);
+        assert(equilCartesianIndexMapper_);
 #endif
         return *equilCartesianIndexMapper_;
     }
@@ -280,8 +281,16 @@ protected:
         // the initial condition is calculated.
         // After loadbalance grid_ will contain a global and distribute view.
         // equilGrid_being a shallow copy only the global view.
-        equilGrid_ = new Dune::CpGrid(*grid_);
-        equilCartesianIndexMapper_ = new CartesianIndexMapper(*equilGrid_);
+        if (grid_->size(0))
+        {
+            equilGrid_ = new Dune::CpGrid(*grid_);
+            equilCartesianIndexMapper_ = new CartesianIndexMapper(*equilGrid_);
+        }
+        else
+        {
+            equilGrid_ = nullptr;
+            equilCartesianIndexMapper_ = nullptr;
+        }
 
         globalTrans_ = nullptr;
     }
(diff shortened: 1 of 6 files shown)


blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
Removed parallel debugging code (which broke test due to missing MPI_Init) and force pushed.

blattms:
jenkins build this please

blattms:
Ok, seems like there a good portion of more work todo, since I need to refactor how restarting values are loaded if there is no global grid. Should have seen that coming...

The other failed parallel tests mpi.compareParallelSim_flow+SPE1CASE2 and
mpi.compareParallelSim_flow+SPE9_CP_SHORT in 
https://ci.opm-project.org/job/opm-simulators-PR-builder/lastBuild/#showFailuresLink
are related to errors in BGSAT and BPR. Any enlightenment how they could be caused is highly appreciated. Maybe missing filtering of the well-connections on slave processes?

joakim-hove:
> The other failed parallel tests mpi.compareParallelSim_flow+SPE1CASE2 and
mpi.compareParallelSim_flow+SPE9_CP_SHORT in
https://ci.opm-project.org/job/opm-simulators-PR-builder/lastBuild/#showFailuresLink

It seems like the `lastBuild` link now points a more recent build which was succesfull; so I could not see your error messages. However:

> are related to errors in BGSAT and BPR. 

The BGSAT and BPR summary keywords are gas saturations and block pressure *in a cell*; i.e. `BPR:1,5,10` is the pressure in cell `(1,5,10)` - that might give you a clue?

joakim-hove:
> It seems like the lastBuild link now points a more recent build which was succesfull; so I could not see your error messages. [...]

The failing build was still linked: https://ci.opm-project.org/job/opm-simulators-PR-builder/214/

The relative error is 1.00 - which to me indicates that the test compares a reference value with an unitiialized 0.0

blattms:
Turns out that the parallel writer still needs to have the global grid on all processes, since it determines the order of the data send and received by via the traversal order of the global grid. (I guess this might be needed to support other grids apart from CpGrid?)

One way to resolve this is by using the order of the cartesian indices of the distributed grid in the case of the CpGrid. Or use CpGrid::gatherGrid, but in that case the way how the cell data is written to the buffer might need to be adapted.

blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
Fixed some of the failed tests. Restart test are still failing as restarting still assumes the global grid to be available everywhere.

bska:
jenkins build this please

opm/simulators/utils/ParallelRestart.hpp
```diff
@@ -0,0 +1,542 @@
...
+template<class T>
+std::size_t packSize(const std::vector<T>& data, Dune::MPIHelper::MPICommunicator comm)
+{
+    if (std::is_pod<T>::value)
+        return packSize(data.data(), data.size(), comm);
+
+    std::size_t size=0;
+    for(const auto& entry: data)
+        size+=packSize(entry, comm);
+}
```
bska: Did you mean
```C++
return size;
```
here?

blattms:
> +template<class T>
> +std::size_t packSize(const std::vector<T>& data, Dune::MPIHelper::MPICommunicator comm)
> +{
> +    if (std::is_pod<T>::value)
> +        return packSize(data.data(), data.size(), comm);
> +
> +    std::size_t size=0;
> +    for(const auto& entry: data)
> +        size+=packSize(entry, comm);
> +}
> 
> Did you mean
> ```C++
> return size;
> ```
> here?

Indeed, that is what I meant.

(Would have loved to say that i did this on purpose relying on the compiler
returning the result of the last operation as it should. But that was not on
my mind).

Thanks for triggering  a build. But I will need to run some testcases here,
still. Would be pleasantly surprised if everything works, though.
Just wanted to have the work on github.

bska:
>> Did you mean
>>  ```C++
>>  return size;
>>  ```
>>  here?
>
> Indeed, that is what I meant.

Very good.

> (Would have loved to say that i did this on purpose relying on the compiler
> returning the result of the last operation as it should. But that was not on
> my mind).

😉 

> Thanks for triggering  a build. But I will need to run some testcases here,
> still.

Okay.  Do let us know when you think this is ready for review then.

> Just wanted to have the work on github.

Better safe than sorry.  Thanks for the clarifications.

blattms:
jenkins build this opm-common=1094 opm-grid=394 please

blattms:
Yes, finally jenkins is happy.

blattms:
But others might not be.

@dr-robertk: I did quite some extensive changes to file collecttoiorank.hh, to  get rid of the dependency of having the global grid available on all processors. Unfortunately, that must have broken usability with grids other than CpGrid. Is it used with theses currently / does anybody rely on it? In that case: Is there a test?

There are basically two changes:
1. Entries of indexMaps_ previously stored the index that the cell was mapped to in the global grid. This is this the case for all entries other than indexMaps.back(). I guess this is an implementation detail (apart from the old usage to query it in `CollectToIORank::localIdxToGlobalId()`.
2. We set up a mapping from local index to global index via `CpGrid::scatter` to use it in `CollectToIORank::localIdxToGlobalId()` (since `indexMaps_.back()` does not provide this information anymore). This of course breaks usability with other grids.

So the question is: If this is used with other grids, then do they use `CollectToIORank::localIdxToGlobalId()`?

dr-robertk:
@blattms: This CollectToIORank is basically a hacky way of implementing a parallel to serial I/O. Storing the global grid on all cores is of course not a good solution. As far as I remember this was only used with CpGrid. PolyhedralGrid is not parallel and the other grids can't produce ecl output. For those the vtk output works and that is parallel.

blattms:
Thanks. Since it is only used with CpGrid this can go in as is then.

That means this is now ready to be reviewed.

blattms:
Latest [successful run](https://ci.opm-project.org/job/opm-simulators-PR-builder/241/) was before the rebase. So let us try again.

blattms:
jenkins build this please

blattms:
At least the failue is consistent with the one in  OPM/opm-grid#389. Therefore I must have messed up my rebase onto @totto82's PR #2034.

blattms:

blattms:
Can I stop jenkins somehow.

Turns out that I am simply to dumb to do some serious rebasing. Thank god I  do not have to do this all the time as a release manager. ;)

Another rebase is upcoming...

blattms:
The problem was that I did not use the SummarySate from the simulator to load the restart values, but used he old approach of creating with the start time.

Whether this fixed the issue is left for proof by jenkins.

atgeirr:
> Can I stop jenkins somehow.

Yes, on https://ci.opm-project.org, when you are logged in a red x button appears next to the build job in list on the left part of the screen.

blattms:
jenkins build this please

blattms:
*sigh* how hard can it be...

blattms:
jenkins build this please

blattms:
:tada: finally rebasing was successful!

Ready for review, nitpicking and whatever else!

ebos/eclcpgridvanguard.hh
```diff
@@ -83,8 +83,11 @@ private:
 
 public:
     EclCpGridVanguard(Simulator& simulator)
-        : EclBaseVanguard<TypeTag>(simulator)
+        : EclBaseVanguard<TypeTag>(simulator), mpiRank()
```
akva2: i think this is less obvious than an #else and being an int there is no cost.

opm/simulators/utils/ParallelRestart.cpp
```diff
@@ -0,0 +1,637 @@
...
+}
+
+void pack(const char* str, std::vector<char>& buffer, int& position,
+          Dune::MPIHelper::MPICommunicator comm)
+{
+#if HAVE_MPI
+    std::size_t length = strlen(str)+1;
+    MPI_Pack(&length, 1, Dune::MPITraits<std::size_t>::getType(), buffer.data(),
+        buffer.size(), &position, comm);
+MPI_Pack(str, strlen(str)+1, MPI_CHAR, buffer.data(), buffer.size(),
```
akva2: wrong indent.

opm/simulators/utils/ParallelRestart.cpp
```diff
@@ -0,0 +1,637 @@
...
+    MPI_Unpack(buffer.data(), buffer.size(), &position, const_cast<char*>(str), length, MPI_CHAR, comm);
+#endif
+}
+
+void unpack(std::string& str, std::vector<char>& buffer, int& position,
+            Dune::MPIHelper::MPICommunicator comm)
+{
+    std::size_t length=0;
+    unpack(length, buffer, position, comm);
+    char* cStr = new char[length];
```
akva2: just because delete makes me twitch; this could be a vector.

ebos/eclcpgridvanguard.hh
blattms: Even with MPI and valgrind (used without MPI bindings) might complain that mpiRank might be uninitialized. So the alternative with else is
```
    : EclBaseVanguard<TypeTag>(simulator)
{
#if HAVE_MPI
        mpiRank = 0
        MPI_Comm_rank(MPI_COMM_WORLD, &mpiRank);
#else 
        mpiRank=0
#endif
        this->callImplementationInit();
    }
```

Looks worse IMHO

opm/simulators/utils/ParallelRestart.cpp
blattms: Indeed, will fix.

opm/simulators/utils/ParallelRestart.cpp
blattms: I actually agree since I have the tendency to falsely use `delete` instead of `delete[]` for arrays. Will fix.

blattms:
Fixed everything but one comment.

Additional I fixed the unused parameter warnings when compiling without MPI (missed them before).

Concerning the other points:
- The hugest strain on my patience is my own constant inability to write bug-free code even for the simplest things.
- MPI_Pack is a rather old and stable API,should we well tested, does not introduce a new dependency and I have used it before. Cereal I have never tried and looking at its github page I am now a bit worried about how well maintained it is. (Last PR merged in May). Therefore we should test it before using.

blattms:
BTW Thanks a lot for the review.

blattms:
jenkins build this please

blattms:
bacckported to release via #2088

PR #2034, opened 2019-10-01
sync restart

totto82:
Stop reading the restart values twice.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -493,6 +493,8 @@ namespace Opm {
 
         const Wells* wells = wellsmanager.c_wells();
 
+        wells_ecl_ = schedule().getWells2(report_step);
+
         const int nw = wells->number_of_wells;
         if (nw > 0) {
             const auto phaseUsage = phaseUsageFromDeck(eclState());
--- a/ebos/eclwellmanager.hh
+++ b/ebos/eclwellmanager.hh
@@ -39,6 +39,8 @@
 #include <opm/parser/eclipse/EclipseState/Schedule/Well/WellConnections.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/Well/Well2.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/TimeMap.hpp>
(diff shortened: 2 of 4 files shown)


totto82:
jenkins build this please

totto82:
@joakim-hove Is the summaryState sufficiently tested such that I can trust Jenkins that this PR doesn't mess up something. 
@atgeirr I have verified that restart in flow where next time step is passed using EXTRA works.

totto82:
@joakim-hove @atgeirr ping!

atgeirr:
Should be merged. I'll rerun jenkins since it was run a while ago.

atgeirr:
jenkins build this please

blattms:
Nice, need to update #2033 afterwards.

atgeirr:
> Nice, need to update #2033 afterwards.

I assume that means you are ok with merging this first?

blattms:
I basically do not care what comes in first

atgeirr:
> I basically do not care what comes in first

I'll merge this now then.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -493,6 +493,8 @@ namespace Opm {
 
         const Wells* wells = wellsmanager.c_wells();
 
+        wells_ecl_ = schedule().getWells2(report_step);
```
blattms: I really need to find out why this PR broke my branch.
Can you please tell me why this is needed here? Previously, this was only done in beginReportStep and now additionally here in initFromRestartFile

atgeirr: I do not see any obvious flaw here unfortunately. What is the broken behaviour?

totto82: Before beginReportStep was called in the restart process. Now it is not. Maybe adding the following lines to initFromRestartFile helps 

```
wells_active_ = localWellsActive() ? 1 : 0;
wells_active_ = grid.comm().max(wells_active_);
```

blattms:
Makes sense in the release and will there save me from a merged conflict.

blattms:
backported to release via #2087

PR #2036, opened 2019-10-01
Make sure subclass functions are not called before subclass is initia…

### LINKED ISSUE #2035: vanguards are fragile (mixing base and subclass function calls in constructor)
I spent the last few our debugging test_ecl_output with my changes from #2033 and it is/was driving me a bit crazy.  The construction seems to be quite fragile or someone might call it a layering violation.

`EclCpGridVanguard` is a subclass of `ECLBaseVanguard`. It does not have constructor itself but inherits the one of `ECLBaseVanguard`. Inside of that constructor functions of the subclass are called (e.g. `EclCpGridVanguard::createGrids_`). The only reason this can have worked is that all the pointers were initialized by calling `EclCpGridVanguard::createGrids_` and `EclCpGridVanguard::loadbalance` before anything else. Without this careful initialization the class is not fully initialized (it certainly breaks when moving to `unique_ptr` since a member function will be called before the newly introduces constructor, calling the one of the base class. is even finished.

blattms:
…lized.

This at least slightly improves the old design. In that design the
subclass had no own constructor but inherited the one of the base class.
That base class constructor called certain subclass
functions (createGrids_, filterConnections_, updateOutputDir_, and
finalizeInit_)that would initialize raw pointers of the
subclass. Hence subclasses where not allowed to have non-pod members
and those used later (e.g. deleted in the destructor) had to be
initialized in these functions.

The new (still ugly) design introduces constructors into the
subclasses and skips inheriting constructors. Now one must call a base
class function classImplementationInit which will still call the
functions createGrids_, filterConnections_, updateOutputDir_, and
finalizeInit_, but at least at this point the baseclass is fully
constructed and the subclass is constructed as much as
possible/needed (non-pod types will be initialized now.)

Closes #2035.

### DIFF
--- a/ebos/eclalugridvanguard.hh
+++ b/ebos/eclalugridvanguard.hh
@@ -81,10 +81,11 @@ private:
     static const int dimension = Grid::dimension;
 
 public:
-    /*!
-     * \brief Inherit the constructors from the base class.
-     */
-    using EclBaseVanguard<TypeTag>::EclBaseVanguard;
+    EclAluGridVanguard(Simulator& simulator)
+        : EclBaseVanguard<TypeTag>(simulator)
+    {
+        this->callImplementationInit();
+    }
 
     ~EclAluGridVanguard()
     {
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -373,16 +373,6 @@ public:
         int outputInterval = EWOMS_GET_PARAM(TypeTag, int, EclOutputInterval);
         if (outputInterval >= 0)
             eclState_->getRestartConfig().overrideRestartWriteInterval(outputInterval);
-
-        asImp_().createGrids_();
-        asImp_().filterConnections_();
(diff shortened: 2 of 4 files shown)


blattms:
jenkins build this please

blattms:
jenkins run norne spe please

blattms:
jenkins failure
```
/var/lib/jenkins/workspace/opm-simulators-PR-builder/jenkins/run-spe.sh: line 19: /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/bin/flow_legacy: No such file or directory
```
seems unrelated. All tests seem to have succeeded

totto82:
I am fine with the changes itself, even though I don't know exactly where the old code bites since I haven't followed the #2033 deeply. Please merge when green.

totto82:
jenkins build this please

atgeirr:
Thanks, I agree this might not be extremely elegant, but at least it is correct!

PR #2039, opened 2019-10-03
move PRT header to the top

totto82:


totto82:
jenkins build this please

opm/simulators/flow/FlowMainEbos.hpp
```diff
@@ -271,33 +270,9 @@ namespace Opm
             }
         }
 
-    protected:
-        void setupParallelism()
```
atgeirr: This function seems unchanged, why is it moved?

totto82: It is the printPRTHeader that is moved to public.

atgeirr: I see, thanks. Merging!

PR #2043, opened 2019-10-04
Prepare for Making Summary Writing Independent of LibECL

bska:
This commit adds requisite libecl includes that are needed as an intermediate step for enabling new summary writing independent of libecl.


bska:
jenkins build this please

PR #2044, opened 2019-10-04
[WIP] adapting to the interface change in OPM/opm-common#1068

GitPaean:
The interface of Aquancon changes there.

And also, the file BlackoilAquiferModel_impl.hpp is changed to unix format to avoid the `^M` when `git diff`.


GitPaean:
The interface of the upstream PR will be updated. This PR is not relevant anymore.

PR #2046, opened 2019-10-08
Add logging in ebos

totto82:
With this .PRT and .DBG files are created also for ebos.

Does not impact Flow. 

Depends on OPM/opm-models#553

### DIFF
--- a/ebos/ebos_altidx.cc
+++ b/ebos/ebos_altidx.cc
@@ -32,6 +32,7 @@
 #include "config.h"
 
 #include "ebos.hh"
+#include "startEbos.hh"
 
 namespace Opm {
 class EclAlternativeBlackOilIndexTraits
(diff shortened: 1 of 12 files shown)


totto82:
jenkins build this opm-models=553 please

totto82:
jenkins build this opm-models=553 please

totto82:
jenkins build this opm-models=553 please

totto82:
@andlaus Do you want to review this?

andlaus:
as you wish, but in the end the people who are active in OPM need to be happy with it, not me ;)

as a general remark, it would IMO be much better to put this kind of functionality into the vanguard instead of copy-and-pasting the main function, i.e., add a `setupLogging()` static method into the [base vanguard](https://github.com/OPM/opm-models/blob/master/opm/models/io/basevanguard.hh) and call this from `start()`. In most cases this method would just be empty, but complicated beasts like `ebos` can still roll their own stuff...

dr-robertk:
This is a good feature for although the implementation may not be perfect yet. But since feature is needed I will merge it.

blattms:
backported to release with PR #2063

Issue #2059, opened 2019-10-15
Eclipse file reading diagnostics printed multiple times in a parallel run


blattms:
Each ranks reads and parses the file and all of them print the diagonstics like:
```
    0 Reading RUNSPEC  in file /var/lib/jenkins/post-builder/workspace/opm-grid-PR-builder/deps/opm-tests/spe1/SPE1CASE2_ACTNUM.DATA, line 23
    1 Reading TITLE    in file /var/lib/jenkins/post-builder/workspace/opm-grid-PR-builder/deps/opm-tests/spe1/SPE1CASE2_ACTNUM.DATA, line 26
    2 Reading DIMENS   in file /var/lib/jenkins/post-builder/workspace/opm-grid-PR-builder/deps/opm-tests/spe1/SPE1CASE2_ACTNUM.DATA, line 29
    3 Reading EQLDIMS  in file /var/lib/jenkins/post-builder/workspace/opm-grid-PR-builder/deps/opm-tests/spe1/SPE1CASE2_ACTNUM.DATA, line 34
    4 Reading TABDIMS  in file /var/lib/jenkins/post-builder/workspace/opm-grid-PR-builder/deps/opm-tests/spe1/SPE1CASE2_ACTNUM.DATA, line 39
```

This should probably on be done on rank 0?

joakim-hove:
> This should probably on be done on rank 0?

The parsing of the file happens on all ranks; changing that is a large task - I guess partly what you are on to. But I can certainly add a `bool verbose` flag to the parser function and pass `isIOrank()` to that.

blattms:
Yes, I meant the printing.
Your solution was what I was thinking of.

joakim-hove:
See: https://github.com/OPM/opm-simulators/pull/2061 - the issue was less clear cut than originally thought

Issue #1900, opened 2019-06-20
Broken tests - on my machine


joakim-hove:
OK - instead of the venerable "works on my machine" I seem to have a "does not work on my machine" situation; looking at jenkins post-builders it seems that current master is all green in that department, but locally on my machine several tests fail:
```
39/59 Test #39: compareECLFiles_flow+3D_MSW .........................***Failed   24.28 sec
45/59 Test #45: compareECLFiles_flow+MSW_MODEL_1 ....................***Failed    4.00 sec
47/59 Test #47: compareECLFiles_flow+FAULTS_MODEL_1 .................***Failed    5.83 sec
39/59 Test #39: compareECLFiles_flow+3D_MSW .........................***Failed   24.28 sec
45/59 Test #45: compareECLFiles_flow+MSW_MODEL_1 ....................***Failed    4.00 sec
47/59 Test #47: compareECLFiles_flow+FAULTS_MODEL_1 .................***Failed    5.83 sec
```

I will post test output from the individual tests in separate comments.

GitPaean:
Yes. You can just post one of them, like 3D_MSW.   They are related to MSW.

joakim-hove:
**39: compareECLFiles_flow+3D_MSW**
```
Unified restart files, sequence  39
39: 
39: Comparing SEQNUM ...  done.
39: Comparing INTEHEAD ...  done.
39: Comparing LOGIHEAD ...  done.
39: Comparing DOUBHEAD ...  done.
39: Comparing IGRP ...  done.
39: Comparing SGRP ...  done.
39: Comparing XGRP ...  done.
39: Comparing ZGRP ...  done.
39: Comparing ISEG ...  done.
39: Comparing ILBS ...  done.
39: Comparing ILBR ...  done.
39: Comparing RSEG ...  done.
39: Comparing IWEL ...  done.
39: Comparing SWEL ...  done.
39: Comparing XWEL ...  done.
39: Comparing ZWEL ...  done.
39: Comparing ICON ...  done.
39: Comparing SCON ...  done.
39: Comparing XCON ...  done.
39: Comparing STARTSOL ...  done.
39: Comparing PRESSURE ...  done.
39: Comparing RS ...  done.
39: Comparing RV ...  done.
39: Comparing SGAS ...  done.
39: Comparing SWAT ...  done.
39: Comparing ENDSOL ...  done.
39:  
39: 4 keywords exhibit failures
39: 	RSEG: Restart, sequence 34
[...]

PR #2061, opened 2019-10-15
stdout logging only for rank 0

joakim-hove:
This starts with #2059 

While discussing a suitable fix for #2059 with @atgeirr it became clear that there is a divide through the code when it comes to logging and parallell behavior:

**opm-common:** This module is totally unaware of MPI and parallell execution. The opm-common code makes many calls to `OpmLog::xxxx()` - these messages will come repeated for every PE. Up until quite recently there has been a bug in the setup of the logging, which resulted in *no messages from opm-common* - we therefor never saw these repeated messages. This bug was fixed  ~1 month ago, and now log messages from opm-common are repeated for every PE - as pointed out in #2059.

For opm-common the situation is that any logging message should be identical on all PE's, and we are only interested in the messages which come from rank 0. The fix in this PR is a good solution seen from the perspective of opm-common.

**opm-simulators:** Here the different PE's see a different world, they will log different messages and should all be able to log. The current PR might silence legitimate `stdout` messages from PE != 0 in e.g. the well code. The file backed logs will still log from all PE's.

My initial plan was to send a boolean flag down to opm-common, and that is of course still possible, but I would prefer not to do it like that. The reason I dislike this approach is:

1. We have a logging system which can be configured quite flexibly. Passing a `bool verbose` flag around is an extra log configuration mechanism on the side.

2. It is a quite invasive change to protect all logging calls in opm-common with `if (verbose) { ... }` and it also sneaks in a parellell awareness in opm-common. I think it would even constitute a *layering violation*? 

I have added the 2019.10 release tag on this PR; that does not imply that it should necessary be merged in anything close current form - but #2059 should be solved in one way or another.

### DIFF
--- a/flow/flow.cpp
+++ b/flow/flow.cpp
@@ -215,10 +215,11 @@ FileOutputMode setupLogging(int mpi_rank_, const std::string& deck_filename, con
         Opm::OpmLog::addBackend("DEBUGLOG", debugLog);
     }
 
-    std::shared_ptr<Opm::StreamLog> streamLog = std::make_shared<Opm::StreamLog>(std::cout, Opm::Log::StdoutMessageTypes);
-    Opm::OpmLog::addBackend(stdout_log_id, streamLog);
-    streamLog->setMessageFormatter(std::make_shared<Opm::SimpleMessageFormatter>(true));
-
+    if (mpi_rank_ == 0) {
+        std::shared_ptr<Opm::StreamLog> streamLog = std::make_shared<Opm::StreamLog>(std::cout, Opm::Log::StdoutMessageTypes);
+        Opm::OpmLog::addBackend(stdout_log_id, streamLog);
+        streamLog->setMessageFormatter(std::make_shared<Opm::SimpleMessageFormatter>(true));
+    }
     return output;
 }
 
@@ -350,7 +351,9 @@ int main(int argc, char** argv)
             eclipseState.reset( new Opm::EclipseState(*deck, parseContext, errorGuard ));
             schedule.reset(new Opm::Schedule(*deck, *eclipseState, parseContext, errorGuard));
             summaryConfig.reset( new Opm::SummaryConfig(*deck, *schedule, eclipseState->getTableManager(), parseContext, errorGuard));
-            setupMessageLimiter(schedule->getMessageLimits(), "STDOUT_LOGGER");
+            if (mpiRank == 0) {
+                setupMessageLimiter(schedule->getMessageLimits(), "STDOUT_LOGGER");
+            }
 
             Opm::checkConsistentArrayDimensions(*eclipseState, *schedule, parseContext, errorGuard);
 


atgeirr:
Just so I don't forget: while this clears up the standard out log, the PRT and DBG logs would still have repeated messages. The underlying issue is that while the opm-common code has a parallel-ignoring design (on purpose) the opm-simulators code does not, and treating log messages the same way may not be appropriate. (For example ignoring not-io-rank messages is fine for the parser where all ranks give the same messages, but may not be fine for the well code where a message typically originates from a single rank and must be communicated).

joakim-hove:
jenkins build this please

bska:
> I don't quite understand why, but this seems to be hitting timeouts

Actually, on closer inspection, other builds are hitting timeouts too.  This **may** be a local computer issue on the CI system (e.g., disk full as a result of high activity in the past 24 hours?)

joakim-hove:
jenkins build this please

joakim-hove:
> I don't quite understand why, but this seems to be hitting timeouts in the parallel regression tests.

Hmmm - still seems to time out :angry: MPI deadlock?

joakim-hove:
jenkins build this please

atgeirr:
> Hmmm - still seems to time out 😠 MPI deadlock?

Does it test fine locally?

joakim-hove:
> Does it test fine locally?

To be honest several ~10 tests fail on my machine and have for a [long time](https://github.com/OPM/opm-simulators/issues/1900) so testing/debugging stuff in opm-simulators is not very motivating :angry: :rage: As for this particular test - it seems to block on my machine as well (I had not tested .... ) - I will look more into it.

akva2:
```diff
diff --git a/flow/flow.cpp b/flow/flow.cpp
index 573bd00ef..33499ae84 100644
--- a/flow/flow.cpp
+++ b/flow/flow.cpp
@@ -351,7 +351,8 @@ int main(int argc, char** argv)
             eclipseState.reset( new Opm::EclipseState(*deck, parseContext, errorGuard ));
             schedule.reset(new Opm::Schedule(*deck, *eclipseState, parseContext, errorGuard));
             summaryConfig.reset( new Opm::SummaryConfig(*deck, *schedule, eclipseState->getTableManager(), parseContext, errorGuard));
-            setupMessageLimiter(schedule->getMessageLimits(), "STDOUT_LOGGER");
+            if (outputCout)
+              setupMessageLimiter(schedule->getMessageLimits(), "STDOUT_LOGGER");
 
             Opm::checkConsistentArrayDimensions(*eclipseState, *schedule, parseContext, errorGuard);
```

the deadlock was due to get getBackend call in setupMessageLimiter throwin on non-root processes.

joakim-hove:
> the deadlock was due to get getBackend call in setupMessageLimiter throwin on non-root processes.

Thank you - :bouquet:

joakim-hove:
OK - I have updated the patch according to diff from @akva2; now we have *two* conditionals which effectively control output to the console:
```C++

if (mpi_rank == 0) {
   ...
}
```
and 
```C++
if (outputCout) {
   ....
}
```
currently `outputCout` is initialized as `outputCout = (mpi_rank == 0)` - so right now they are equivalent, but that could in principle change in the future. 

Anyways - hopefully jenkins will pull through happily now.

blattms:
Thanks a lot.

blattms:
backported to release via #2084

PR #2067, opened 2019-10-17
Missing features on R

joakim-hove:


joakim-hove:
jenkins build this please

blattms:
This marked for the release, but others with "Missing features ..." or "keyword" are not. Is that intentional?

atgeirr:
I think it is safe to assume that they should be treated identically.

blattms:
backported to release via #2081

PR #2070, opened 2019-10-17
enables rockcomp in flow (backport of PR 2047

blattms:


OPMUSER:
The ROCKCOMP keyword activates rock compaction and defines various rock compaction options for the run.  By default OPM Flow models rock compaction via pore volume compressibility as entered on the ROCK keyword in the PROPS section. This keyword enables pressure dependent pore volume and transmissibility multipliers for rock compaction that are entered in the PROPS section using the ROCKTAB keyword. Currently OPM Flow only supports the default options for rock compaction.

**ROCKTAB is in MissingFeatures**

I doubt we are using ROCKTABH for hysteretic rock compaction or the Palmer-Mansoori option using
ROCKPAMA, ROCKPAME or ROCKPMAT keywords - but I have been wrong before ;-)

totto82:
We don't support ROCKTAB but ROCK2D and friends. I guess adding ROCKTAB should not be a big deal, but I don't think it is on any todo list.

OPMUSER:
@totto82  just trying to ensure the manual is consistent with the release. So Rock2D, ROCKWNOD and ROCK2DTR, are the only active keywords, yes.

OPMUSER:
@totto82  So Rock2D, ROCKWNOD and ROCK2DTR, are the only active keywords, yes.

totto82:
We also support OVERBURD

For output we support
PRESROCC, (minimum oil pressure)
PRES_OVB, (overburden pressure)
PORV_RC (porv multiplier)
TMULT_RC (transmissibility multiplier)
SWMAX (max water saturation)

I notices that this is a mixture of ecl names and self-made names and output values.

PR #2071, opened 2019-10-18
Always allocate buffers when restarting

totto82:


blattms:
So previously, the buffers were only written in `beginRestart` if (accidentally) `restartConfig.getWriteRestartFile(reportStepNum, log)` was false. Now they always are.

Release or not Release?

totto82:
Jostein confirms it works.  Release

alfbr:
This is critical on our side, it restores the ability to restart with flow. Hence, I believe it is a strong candidate for release. We already pull it in, but it may bite others for the half year to come.

totto82:
jenkins build this please

bska:
I'm going to merge this now as it fixes an important use-case for Flow.  Thanks a lot for the quick analysis of the issue and the turn-around on providing a solution.

blattms:
backported to release via #2083

PR #2075, opened 2019-10-19
Use Location class for keyword location

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/1122


joakim-hove:
jenkins build this opm-common=1122 please

PR #2077, opened 2019-10-19
[WIP] in fixing aquifer

GitPaean:
for the situation that an aquifer is defined with keyword AQUFETP, while
there is no valid connection specified.

It is a hacky approach without touching the parser.  When some reference behavoir is obtained, the implementation can be updated. 

It results from the deck shared through https://github.com/OPM/opm-common/issues/1130 . 

There are some debugging output code still there, will be removed when removing WIP.


GitPaean:
This PR is probably not very useful since it is intended to address some flawed deck that commercial simulator can handle while OPM can not.  Closing it for now.

PR #2089, opened 2019-10-21
added: allow running the norne test in parallel

akva2:
This enables triggering a parallel norne run using 'run norne_parallel'. trigger runs with 4 procs right now.

### DIFF
--- a/jenkins/run-norne.sh
+++ b/jenkins/run-norne.sh
@@ -7,7 +7,12 @@ test -z $SIM && SIM=flow
 # Run the norne case
 cd norne
 mkdir $SIM
-$WORKSPACE/$configuration/build-opm-simulators/bin/$SIM --output-dir=$SIM NORNE_ATW2013.DATA
+if test -n "$1"
+then
+  mpirun -np $1 $WORKSPACE/$configuration/build-opm-simulators/bin/$SIM --output-dir=$SIM NORNE_ATW2013.DATA
+else
+  $WORKSPACE/$configuration/build-opm-simulators/bin/$SIM --output-dir=$SIM NORNE_ATW2013.DATA
+fi
 test $? -eq 0 || exit 1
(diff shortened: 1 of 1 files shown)


akva2:

akva2:
this worked, but as observed locally, the new opm-grid code deadlocks for norne with 4 processes. @blattms

blattms:
Not sure I can easily fix this in a fast manner and awfully sorry for my premature merge. I guess I will have to propose reverting that merge.

blattms:
I have a merge request reverting the premature merge in OPM/opm-grid#404

atgeirr:
The parallel tests in opm-simulators work, and locally I assumes this works @blattms? Is the error only showing up on Jenkins?

akva2:
the error shows up with norne using e.g. 4 and 8 processes. it passes with 2 and 6. i'll take the liberty of quoting myself;
[quote]
this worked, but as observed *locally*, the new opm-grid code deadlocks for norne with 4 processes.
[/quote]

akva2:
this pr has value in itself, please consider merging it.

atgeirr:

atgeirr:
This ran through, but the artifacts (well plots) were not generated properly:

```
./plotwells.sh: line 22: /var/lib/jenkins/workspace/opm-simulators-PR-builder@2/mpi/install/bin/summary.x: No such file or directory
```

I guess that is a separate problem not caused by this PR, so this should be merged in any case?

akva2:
yeah, that's a libecl binary we have not yet replaced. not related to this pr.

atgeirr:
> yeah, that's a libecl binary we have not yet replaced

No need to replace the binary for that purpose? It's not like it needs to be updated all the time, if at all?

akva2:
not as such, but it was built as part of the libecl dependency. @tskille said he'd look at implementing a replacement.

atgeirr:
> @tskille said he'd look at implementing a replacement.

Ok, I suggest we keep using the utilities though until such a replacement exists.

akva2:
okay, i'll put int the work to use a system hosted installed version and build one up then.

PR #2092, opened 2019-10-22
Add well operability param

totto82:
For several cases (The latest a slightly modified version of Olympus) the current well operability checking does not work as expected. This adds a easy way to disable the well operability checking. The default is still true, i.e. this does not change the current behavior of the simulator.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -1910,6 +1910,12 @@ namespace Opm
                          const WellState& well_state,
                          Opm::DeferredLogger& deferred_logger)
     {
+
+        const bool checkOperability = EWOMS_GET_PARAM(TypeTag, bool, EnableWellOperabilityCheck);
+        if (!checkOperability) {
+            return;
+        }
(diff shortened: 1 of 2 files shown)


totto82:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -1910,6 +1910,11 @@ namespace Opm
                          const WellState& well_state,
                          Opm::DeferredLogger& deferred_logger)
     {
+
+        bool checkOperability = EWOMS_GET_PARAM(TypeTag, bool, EnableWellOperabilityCheck);
```
GitPaean: const bool here and please add brackets for the `if` below.

totto82:
jenkins build this please

GitPaean:
It is a known issue and is from the imperfect related strategy.

Not sure eventually it is desirable to set it false (which means the param can be removed later). While for now, it is okay to provide some means to help the current flow run through some cases.

blattms:
backported to release via PR #2099

PR #2096, opened 2019-10-22
avoid nan in the density calculation

totto82:
This fix avoids nan due to very small values (1e-20) for mix[oil] and max[gas]. The nan was triggered after an STOPed water injector (with epsilon cross flow) started up again. 

Fixes a run for a proprietary model.


totto82:
jenkins build this please

blattms:
Seems like a bugfix we should have in the relase, shouldn't we? I set marker,  remove at your discretion if you do not agree.

blattms:
backported to release via #2106

PR #2100, opened 2019-10-24
BUGFIX. Scale MSW injector

totto82:
Fix scaling on MSW wells.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1621,10 +1621,36 @@ namespace Opm
         } else if (well.isInjector() ) {
             const Opm::Well2::InjectorCMode& current = well_state.currentInjectionControls()[well_index];
             const auto controls = well.injectionControls(summaryState);
+
+            Well2::InjectorType injectorType = controls.injector_type;
+            double scaling = 1.0;
+
+            const auto& pu = phaseUsage();
+
+            switch (injectorType) {
+            case Well2::InjectorType::WATER:
+            {
+                scaling = scalingFactor(pu.phase_pos[BlackoilPhases::Aqua]);
+                break;
+            }
+            case Well2::InjectorType::OIL:
+            {
+                scaling = scalingFactor(pu.phase_pos[BlackoilPhases::Liquid]);
+                break;
+            }
+            case Well2::InjectorType::GAS:
+            {
+                scaling = scalingFactor(pu.phase_pos[BlackoilPhases::Vapour]);
+                break;
+            }
+            default:
+                throw("Expected WATER, OIL or GAS as type for injectors " + well.name());
+            }
+
             switch(current) {
             case Well2::InjectorCMode::RATE:
             {
-                control_eq = getSegmentGTotal(0) * efficiencyFactor - controls.surface_rate;
+                control_eq = getSegmentGTotal(0) * efficiencyFactor - controls.surface_rate*scaling;
                 break;
             }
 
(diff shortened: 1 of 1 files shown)


totto82:
jenskins build this please

totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1621,20 +1621,44 @@ namespace Opm
...
+            }
+            default:
+                throw("Expected WATER, OIL or GAS as type for injectors " + well.name());
+            }
+
             switch(current) {
             case Well2::InjectorCMode::RATE:
             {
-                control_eq = getSegmentGTotal(0) * efficiencyFactor - controls.surface_rate;
+                control_eq = getSegmentGTotal(0) / scaling * efficiencyFactor - controls.surface_rate;
```
GitPaean: I am not sure the usage of `efficiencyFactor` here is correct. But I did not review the previous PR related to this, there is something we can discuss here. 

And also, for the scaling factor here, it is probably better to use multiplication than division .

totto82: > I am not sure the usage of efficiencyFactor here is correct.

Could you be more specific? What is wrong?

> And also, for the scaling factor here, it is probably better to use multiplication than division .

Do you mean. 

```
scaling = 1 / scalingFactor(pu.phase_pos[BlackoilPhases::Vapour]);

control_eq = getSegmentGTotal(0) * scaling * efficiencyFactor - controls.surface_rate;
```

atgeirr: I agree about the scaling, multiply the control instead. That way the control equations are scaled similarly for all RATE targets.

GitPaean: I will come back for the efficiency factor soon later. My current impression is that something is wrong here, but I can be wrong. I need to address something else first.

totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: OK. I have changed the scaling. If there is something wrong with the efficiency factor that will be another PR.

totto82: The order of the scaling may matter. Now I new test case I got from Torbjørn does not run through due to singular matrix. The field case seems unaffected. I may have messed up something els, so I will double check. But we should not merge this before I confirm that the changes you suggested does not harm anything.

GitPaean: > . Now I new test case I got from Torbjørn does not run through due to singular matrix. 

If the multiplication introduces singular matrix, there must be something else is wrong.

totto82:
Any reason for not merging this? It fixes a critical bug.

GitPaean:
I still feel extremely skeptical about the singular matrix resulting from multiplication of the scaling factor, which is probably the good practice both numerically and physically.

@alf and @atgeirr can merge if the ensemble test results are satisfactory.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: You are right it is actually scaled up. I think we agree that the current scaling of the well equation (both the control and the flow) are not optimal. I still think this can be merged. We can change the placement of the scaling factor later when we find the real issue.

alfbr:
> @alfbr  and @atgeirr can merge if the ensemble test results are satisfactory.

Thanks for the confidence Kai, I take your word for it and merge.

PR #2114, opened 2019-10-28
update redhat packaging

akva2:
new boost::filesystem usage requires a newer boost.


blattms:
backported to release via #2116

PR #2126, opened 2019-11-01
Add pybind11

jmsargado:
Work in progress: Python -- add pybind11 files


joakim-hove:
It is good that you have split things up. I think:

1. You should not need the `pycmake` functionality - all of that is already in pybind11 - i.e the commit [076ea3db](https://github.com/OPM/opm-simulators/pull/2126/commits/076ea3db77c3b9638fa396b7f7be8086c303deba) can be completely removed.

2. The `CMakeLists.txt` file in the `python/` folder can be simplified to only contain:
```
add_subdirectory( pybind11 )
```
With these two changes I will merge this.

jmsargado:
@joakim-hove I've removed the pycmake folder and also modified the CMakeLists accordingly.

PR #2127, opened 2019-11-01
Python blackoil simulator

jmsargado:
Python bindings for blackoil simulator using pybind11 (very much still a work in progress). More stuff to follow later.

### DIFF
--- a/python/simulators/simulators.cpp
+++ b/python/simulators/simulators.cpp
@@ -17,6 +17,8 @@
 #include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
 #include <opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp>
 
+#include <opm/parser/eclipse/EclipseState/Schedule/ArrayDimChecker.hpp>
+
 #if HAVE_DUNE_FEM
 #include <dune/fem/misc/mpimanager.hh>
 #else
@@ -24,6 +26,7 @@
 #endif
 
 #include <pybind11/pybind11.h>
+#include <pybind11/embed.h>
 #include <string>
 #include <iostream>
 
@@ -131,7 +134,6 @@ FileOutputMode setupLogging(int mpi_rank_, const std::string& deck_filename, con
     std::string output_dir = cmdline_output_dir;
     if (output_dir.empty()) {
         output_dir = absolute(path(baseName).parent_path()).string();
-        std::cout << "Output dir = " << output_dir << std::endl;
     }
 
     logFileStream << output_dir << "/" << baseName;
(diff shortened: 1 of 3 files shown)


joakim-hove:
The changes in the commit [Skip processing of filename ...](https://github.com/OPM/opm-simulators/pull/2127/commits/08d38fe492bfe23a8ee3afc63c1146d820271f1b) is independent of Python or not and will affect flow usage - I think that should be extracted to a separate pull request.

joakim-hove:
The bindings part in the last commit seems to create a full new simulator implementation; I would *really* have hoped that it should be possible to bind to one of the existing simulators[1] without code duplication.

[1]: Maybe some changes must be made to the existing simulator code, then that should be done first - in separate small pull requests.

python/simulators/simulators.cpp
```diff
@@ -226,6 +233,16 @@ class BlackOilSimulator
         delete[] argv_;
     }
 
+    void setupCmdLineArgs()
```
joakim-hove: I do not understand what this function is needed/used for. Maybe it can be removed entirely?

joakim-hove:
> I've modified the python simulator constructor to accept either the deck filename
'
That seems like a good idea - but I can not see the overload which uses a filename?

jmsargado:
The first constructor implementation accepts a filename, copies it to argv_[1] and then let's the rest of the internal code in the run method do the work. I've put back the code from flow.cpp that generates the deck, eclipse state and other objects given the deck filename. This is then simply skipped when such objects already exist (which is the case when the other constructor is called).

jmsargado:
The setupCmdLineArgs() is just there to setup the internally stored argc and argv that is passed to other functions the need them. Originally the code was part of the constructor but as there's now more than one constructor, I felt it would be easier to have it separate to avoid having to write the same thing twice.

joakim-hove:
I will continue reviewing this tomorrow, it would be good if you could rebase it in the meantime so the diff is a bit more manageable.

joakim-hove:
I have extracted the pure C++ change in a separate PR: https://github.com/OPM/opm-simulators/pull/2151

joakim-hove:
> it would be good if you could rebase it in the meantime so the diff is a bit more manageable.

You really have to rebase this:
```
bash% git fetch upstream
bash% git checkout python_blackoil_simulator
bash% git rebase upstream/master
bash% git push -f origin python_blackoil_simulator
```
it is too cumbersome to review when all the changes from #2126 are also included in the diff.

joakim-hove:
jenkins build this please

joakim-hove:
I have created this : opm/opm-common#1545 to ensure that the Python extension is indeed built by jenkins.

joakim-hove:
Changes are required to `python/simulators.cpp`:

1. The code makes use of `boost::filesystem` - that is no longer available in opm code, you should use the `std::filesystem / std::experimental::filesystem` wrapper available in opm-common:
```
#include <opm/common/utility/FileSystem.hpp>


using fs = Opm::filesystem;
...

``` 
2. It looks like there is much duplicated code between `python/simulators.cpp` and `flow.cpp` - that must be resolved in some way.

joakim-hove:
jenkins build this please

bska:
Closing this now since the patch no longer applies.

PR #2129, opened 2019-11-01
Python example

jmsargado:
Python scripts for running Norne and SPE1. Uses python code from opm-common.


python/examples/flow_example.py
```diff
@@ -0,0 +1,19 @@
...
+from simulators import simulators
+
+parse_context = ParseContext( [("PARSE_RANDOM_SLASH", opm.io.action.ignore)] )
+deck = Parser().parse( '../../../opm-data/norne/NORNE_ATW2013.DATA' , parse_context)
+state = EclipseState(deck)
+schedule = Schedule(deck, state)
+summary_config = SummaryConfig(deck, state, schedule)
+
+p = simulators.BlackOilSimulator()
+p.setDeck(deck)
```
joakim-hove: This should be simplified, either the filename should be a constructor argument:
```
p = simulators.BlackOilSimulator( "filename.DATA")
p.run( )
```
or alternatively *one* method:
```
p = simulators.BlackOilSimulator()
p.load_deck( "file.DATA")
p.run( )
```

bska:
Closing this now since the patch no longer applies.

PR #2130, opened 2019-11-01
Check THP in the end

totto82:
Move the checking of THP down the line. For non-convergent solution we don't want to switch to THP. This is sufficient to get a good match on Norne1B.


totto82:
jenkins build this please

PR #2131, opened 2019-11-01
Fix BlackoilWellModel::numComponents for parallel runs.

blattms:
In the case with active solvent (BlackoilWellModel::has_solvent_ is
true, e.g. case 2D_OILWATER_POLYMER.DATA) and one process have no
wells while another has some this lead to numComponents returning
different values. Later this lead to truncated message in MPI_Allgather.

The underlying problem was the usage of numWells() which returns just
the number of local wells. We resort to using wellsActive which
returns whether there any well active regardless which process knows it.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1380,7 +1380,7 @@ namespace Opm {
     int
     BlackoilWellModel<TypeTag>::numComponents() const
     {
-        if (numWells() > 0 && numPhases() < 3) {
+        if (wellsActive()  && numPhases() < 3) {
             return numPhases();
         }
         int numComp = FluidSystem::numComponents;
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -620,7 +620,7 @@ namespace Opm {
     {
         std::vector<WellInterfacePtr> well_container;
 
-        const int nw = numWells();
+        const int nw = numLocalWells();
 
         if (nw > 0) {
             well_container.reserve(nw);
@@ -732,7 +732,7 @@ namespace Opm {
         const Well2& well_ecl = wells_ecl_[index_well_ecl];
 
         // Finding the location of the well in wells struct.
-        const int nw = numWells();
+        const int nw = numLocalWells();
         int well_index_wells = -999;
         for (int w = 0; w < nw; ++w) {
             if (well_name == std::string(wells()->name[w])) {
@@ -944,7 +944,7 @@ namespace Opm {
     BlackoilWellModel<TypeTag>::
     localWellsActive() const
     {
-        return numWells() > 0;
+        return numLocalWells() > 0;
     }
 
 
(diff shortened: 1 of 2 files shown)


blattms:
fixed the run of 2D_OILWATER_POLYMER.DATA in parallel.

blattms:
@akva2, can I trigger a parallel run of 2D_OILWATER_POLYMER.DATA?

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1380,7 +1380,7 @@ namespace Opm {
     int
     BlackoilWellModel<TypeTag>::numComponents() const
     {
-        if (numWells() > 0 && numPhases() < 3) {
```
GitPaean: This is the very hacky code without comments, which makes it hard. 

Should we check whether we have wells here at all? 

I mean, why do not we just do `if (numPhases() < 3) {`, which is still hacky while simpler.

GitPaean: I will do some tests.  The PR looks okay. I will just check to answer my own question.

GitPaean:
jenkins build this please

GitPaean:
> BlackoilWellModel::has_solvent_ is true, e.g. case 2D_OILWATER_POLYMER.DATA) 

Not related to the code, just for information. `has_solvent_` should be false for the case `2D_OILWATER_POLYMER.DATA`

GitPaean:
> Not related to the code, just for information. has_solvent_ should be false for the case 2D_OILWATER_POLYMER.DATA

But the numComponents are different from numPhases when polymer is involved. That is the problem for this specific case.

GitPaean:
Thanks for the fixing and renaming.  It fixed the running of the problematic case. The code is going in now.

blattms:
the comment about "hacky" code really got me thinking. Looking at the code again, I think there are more
bugs (e.g. numPhases() depends on local wells being active which seems not Ok.

akva2:
Sorry i am off work until monday. There is no such test, consider adding it.

opm/simulators/wells/BlackoilWellModel_impl.hpp
blattms: maybe the original author should comment on whether `numWells()` is needed and what for.

My impression is that this cannot depend on whether there are wells or not and the current version is still buggy if there are no wells at all. The return value of `numComponents` governs the size of `B_avg` used in [`computeAverageFormationFactor`](/OPM/opm-simulators/blob/f95718a3c5d8254053f4ee5097a340d1b8461fc8/opm/simulators/wells/BlackoilWellModel_impl.hpp#L1324-L1333) and that is done for all cells even if there are no wells anywhere. The call of these functions is in `computeWellPotentials` which is again called in any case (whether there are wells or not).

blattms:
backported to release via #2136

blattms:
@GitPaean is right. This had nothing to do with solvent. The problem occured for cases with less than 3 phases where on process stored and computed none of the wells.

atgeirr:
Thanks a lot guys for finding and fixing, you are terrific!

PR #2132, opened 2019-11-01
Support custom vrep and rein group

totto82:
With this item 11 and item 12 in GCONINJE is supported.


atgeirr:
jenkins build this please

PR #2133, opened 2019-11-01
[testing PR] not checking numWells in numComponents in BlackoilWellModel

GitPaean:

PR #2134, opened 2019-11-01
make BlackoilWellModel::numComponents return consistent values.

blattms:
The return value is used to resize a vector that is used unconditionally
(i.e. whether there are wells or not) elsewhere, e.g. in
computeAverageFormationFactor, called from computeWellPotentials
called from timeStepSucceeded. Hence it has to return the same value
whether on all processes whether there wells (globally or locally) or
not.

This is done by this commit that introduces global_num_phases_ that
is the max of all phases on any process.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1151,6 +1151,10 @@ namespace Opm {
     BlackoilWellModel<TypeTag>::
     computeWellPotentials(std::vector<double>& well_potentials, const int reportStepIdx, Opm::DeferredLogger& deferred_logger)
     {
+        if ( ! wellsActive() ) {
+            return;
+        }
+
         // number of wells and phases
         const int nw = numLocalWells();
         const int np = numPhases();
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please

blattms:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -233,6 +233,10 @@ namespace Opm {
         // one process.
         wells_active_ = localWellsActive() ? 1 : 0;
         wells_active_ = grid.comm().max(wells_active_);
+        // Number of phases can only be determined globally as it needs
+        // an active well and there might be processes without wells.
+        global_num_phases_ = numPhases();
+        global_num_phases_ = grid.comm().max(global_num_phases_);
```
GitPaean: I think `numPhases()` should be same cross all the processes, not depending on whether there is a well there. 

If they are different, then the function is an ill function.

blattms:
It turns out (at least for the models I tested) that the case that this PR is supposed to fix does not occur.
`numPhases` already returns the same number on all processes  even of one does not have any and others have.

I am closing this and would expect that the aproach of #2133 should work

PR #2140, opened 2019-11-05
Fix for hierarchy of production groups. 

totto82:
Fix for hierarchy of production groups.


totto82:
jenkins build this please

PR #2141, opened 2019-11-05
fix group fraction bug

totto82:


totto82:
jenkins build this please

PR #2149, opened 2019-11-07
Delete few thermal keywords from list

goncalvesmachadoc:
@slawekszklarz and I were checking the keywords mentioned in warnings as unhandled when running thermal models. The keywords RTEMP, RTEMPVD and THCONR seem to be working fine.  The WATVISCT table breaks the simulation when extrapolation is needed.


totto82:
Thanks. Will merge. I have made a PR addressing WATVISCT extrapolation OPM/opm-material#372. Great if you could test it!

PR #2151, opened 2019-11-07
Skip processing of filename if external deck already exists.

joakim-hove:
This is a plain C++ commit lifted from #2127 

This will allow the vanguard initialization to proceed even if a deck filename is missing - if the Deck / EclipseState7 / ... objects have been supplied externally.


joakim-hove:
jenkins build this please

blattms:
You are aware of the side effects, though? `deckName()` will now return an empty string if there is an
external Deck used.

atgeirr:
I think that seems reasonable, if I constructed the deck inline for example.

blattms:
IMHO If you run flow, an external Deck will be used.

atgeirr:
Right, I'll need to think about this some more then.

PR #2152, opened 2019-11-08
fixing and improvement of the Aquifer models

GitPaean:
The major fixes include two parts, incorporating the derivatives from pressure difference, and correcting the way to calculate the initial aquifer pressure, which is important when the initial aquifer pressure is not specified. 

With this PR and the several merged PRs from opm-common, flow managed to give very good results in the targeting aquifer case with Fetkovich aquifers.  

There will be some cleaning up related to aquifer models, maybe also some fixes should be applied to CT aquifer too.

### DIFF
--- a/opm/simulators/aquifers/AquiferCarterTracy.hpp
+++ b/opm/simulators/aquifers/AquiferCarterTracy.hpp
@@ -63,6 +63,8 @@ namespace Opm
             // Variables constants
             const AquiferCT::AQUCT_data aquct_data_;
             Scalar beta_; // Influx constant
+            // TODO: it is possible it should be a AD variable
+            Scalar mu_w_; // water viscosity
 
             // This function is used to initialize and calculate the alpha_i for each grid connection to the aquifer
             inline void initializeConnections(const Aquancon::AquanconOutput& connection)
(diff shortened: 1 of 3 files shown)


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
Due to the lack of the working example for Carter Tracy aquifer, the refactoring of the both models will remain to be done later.  I am removing the [WIP] tag now. 

In the previous jenkins test, there is the following test failure, which is totally expected. 
```
mpi.compareECLFiles_flow+2D_FETKOVICHAQUIFER
```

I am preforming the `update_data` now.

GitPaean:

GitPaean:
jenkins build this opm-tests=184 please

GitPaean:
jenkins build this opm-tests=184 please

alfbr:
> Due to the lack of the working example for Carter Tracy aquifer

@GitPaean  please elaborate. I believe there is one 2D example in opm-tests under aquifer-oilwater. Do you miss reference results for that one or do you need another case. If so, do you have specific requirements for a new case?

GitPaean:
> @GitPaean please elaborate. I believe there is one 2D example in opm-tests under aquifer-oilwater.

That is a simple case that the master branch handles well. We need a test case with more complicated setup that the master branch can not handle.

For the Fetkowich aquifer, we have a real case, which is the reason we created this PR and the several PRs in opm-common. 

But we do not have something similar for the Carter Tracy aquifer.

GitPaean:
>  do you need another case. If so, do you have specific requirements for a new case?

We will need an another case, with multiple connections, and default initial aquifer pressure with reference result. 

Sorry for not answering the message all by once. It was a little busy here.

PR #2155, opened 2019-11-11
Print number of MPI/OMP threads at start and end.

### LINKED ISSUE #2146: 2020-04 Print End of Simulation Request
I think it would be good to add the number of processes/threads used to run the model, as at the moment there is no such information in any of the output files. So something like this would be useful:
```
================    End of simulation     ===============

Number of processors:         2
Total time (seconds):         591.617
Solver time (seconds):        591.129
 Assembly time (seconds):     215.487 (Failed: 6.72982; 3.12307%)
 Linear solve time (seconds): 322.295 (Failed: 8.95627; 2.7789%)
 Linear solve setup time (seconds): 9.2053 (Failed: 0.330975; 3.59548%)
 Update time (seconds):       22.2743 (Failed: 0.698941; 3.13788%)
 Output write time (seconds): 19.5431
Overall Well Iterations:      920 (Failed: 8; 0.869565%)
Overall Linearizations:       1995 (Failed: 63; 3.15789%)
Overall Newton Iterations:    1659 (Failed: 63; 3.79747%)
Overall Linear Iterations:    25395 (Failed: 663; 2.61075%)
```
If we are counting from zero, then add one so a single core will be print 1 and not zero.

blattms:
Closes #2146

### DIFF
--- a/opm/simulators/flow/FlowMainEbos.hpp
+++ b/opm/simulators/flow/FlowMainEbos.hpp
@@ -237,7 +237,7 @@ namespace Opm
             MPI_Comm_size(MPI_COMM_WORLD, &mpiSize);
 #endif
 
-            std::cout << "Using "<< mpiSize << " MPI processors with "<< threads <<" OMP threads on each \n\n";
+            std::cout << "Using "<< mpiSize << " MPI processes with "<< threads <<" OMP threads on each \n\n";
         }
 
         /// This is the main function of Flow.  It runs a complete simulation with the
@@ -489,13 +489,13 @@ namespace Opm
                 if (output_cout) {
                     std::ostringstream ss;
                     ss << "\n\n================    End of simulation     ===============\n\n";
-                    ss << "Number of MPI processors: " << std::setw(6) << mpi_size_ << "\n";
+                    ss << "Number of MPI processes: " << std::setw(6) << mpi_size_ << "\n";
 #if _OPENMP
                     int threads = omp_get_max_threads();
 #else
                     int threads = 1;
 #endif
-                    ss << "Threads per MPI processor:  " << std::setw(4) << threads << "\n";
+                    ss << "Threads per MPI process:  " << std::setw(5) << threads << "\n";
                     successReport.reportFullyImplicit(ss, &failureReport);
                     OpmLog::info(ss.str());
                 }
--- a/opm/simulators/flow/FlowMainEbos.hpp
+++ b/opm/simulators/flow/FlowMainEbos.hpp
@@ -219,6 +219,19 @@ namespace Opm
             std::cout << "*          For more information, see https://opm-project.org         *\n";
             std::cout << "*                                                                    *\n";
             std::cout << "**********************************************************************\n\n";
+
+            int threads = 1;
+            int mpiSize = 1;
+
+#ifdef _OPENMP
+            threads = omp_get_max_threads();
+#endif
+
+#if HAVE_MPI
+            MPI_Comm_size(MPI_COMM_WORLD, &mpiSize);
+#endif
+
+            std::cout << "Using "<< mpiSize << " MPI processors with "<< threads <<" OMP threads on each \n\n";
         }
 
         /// This is the main function of Flow.  It runs a complete simulation with the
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

opm/simulators/flow/FlowMainEbos.hpp
```diff
@@ -219,6 +219,18 @@ namespace Opm
             std::cout << "*          For more information, see https://opm-project.org         *\n";
             std::cout << "*                                                                    *\n";
             std::cout << "**********************************************************************\n\n";
+
+#ifdef _OPENMP
+            const int threads = 1;
+#else
+            const int threads = omp_get_max_threads();
+#endif
```
bska: Am I reading this wrong?  This appears to invoke `omp_get_max_threads()` when `_OPENMP` is **not** defined.

atgeirr: I agree that looks flipped, another thing: I do not think we use the maximum number of threads, but instead default to 2 threads, if OPENMP is present or 1 if not. On my OpemMP-less build
```
bin/flow --help | grep threads
```
returns
```
    --threads-per-process=INTEGER                 The maximum number of threads to be instantiated per process ('-1' means 'automatic'). Default: 1
```

akva2: omp_get_max_threads() returns the number of threads that will be used in the next parallel section without a num_threads clause. it isnot the amount of hardware threads. ie, this is correct.

atgeirr: Thanks for the explanation! BTW, what should one use to get the amount of hardware threads then? Also: I assume the call to set the number of threads to be used has been made before the call site?

opm/simulators/flow/FlowMainEbos.hpp
```diff
@@ -219,6 +219,19 @@ namespace Opm
             std::cout << "*          For more information, see https://opm-project.org         *\n";
             std::cout << "*                                                                    *\n";
             std::cout << "**********************************************************************\n\n";
+
+            int threads = 1;
+            int mpiSize = 1;
+
+#ifdef _OPENMP
+            const int threads = omp_get_max_threads();
```
akva2: try not re-declaring.

atgeirr: This will shadow the existing variable, why not just
```
threads = omp_get_max_threads();
```

opm/simulators/flow/FlowMainEbos.hpp
akva2: omp_get_num_procs() will tell you the number of physical/ logical cores. but you really should not use this because OMP_NUM_THREADS env var is the authority.

blattms:
Embarassing. You were reading correctly and I was writing rubbish. Amended with correction
and force pushed.

opm/simulators/flow/FlowMainEbos.hpp
blattms: The number of threads used in any serial section is `1`, that is what `opm_get_num_thread()` returns. `opm_get_max_threads` returns the maximum available. That is the value that we set (default is 2 if there are at least 2 processors).

opm/simulators/flow/FlowMainEbos.hpp
blattms: amended and force pushed.

blattms: amended and force pushed.

blattms:
jenkins build this please

blattms:
I will need to double check this again.

blattms:
jenkins build this please

blattms:
Turns out that the number of OpenMP threads gets set  only after the deck is read. Of course that is a bit late for outputting. Hence we now compute the number of threads just for outputting them after the banner.

With 80fd428 the number of threads is now correct for the first print. Before it was only correct if OMP_NUM_THREADS was set, otherwise the number of cores would have been printed.

blattms:
Already cherry-picked to the release as this is super-handy:

```
**********************************************************************
*                                                                    *
*                      This is flow 2019.10-rc4                      *
*                                                                    *
* Flow is a simulator for fully implicit three-phase black-oil flow, *
*             including solvent and polymer capabilities.            *
*          For more information, see https://opm-project.org         *
*                                                                    *
**********************************************************************

Using 8 MPI processors with 2 OMP threads on each 

....

================    End of simulation     ===============

Number of MPI processors:      8
Threads per MPI processor:     2
Total time (seconds):         1088.82
Solver time (seconds):        1088.69
....
```

alfbr:
Just a nit, maybe it should be "processes" rather than "processors", and "process" rather than "processor"?

blattms:
Nitpicks adresses and rebased on current master.

blattms:
jenkins build  this please

blattms:
mmh, tests have not been run. I should learn how to keep jenkins happy.

blattms:
jenkins build this please

bska:
I'm merging this into master now.  More information of this kind is only beneficial.  Thanks a lot!

PR #2156, opened 2019-11-12
Make sure the perforation pressures is initialized also for STOP wells

joakim-hove:
For a well which starts in the state `STOP` the perforation pressures are not properly initialized, and the `-1e100` value from: [`WellState` initialization](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/WellState.hpp#L177) spreads around creating havoc.

~~The current PR works for me - I am trying to make a proper test for it.~~ Help from @atgeirr in the debugging process is acknowledged.


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

GitPaean:
I think it is very good. A stopped well can still have crossflow and reasonable bhp value, as a result, it should have proper perforation rates and perforation pressures.

atgeirr:
Looks good, merging.

PR #2161, opened 2019-11-14
Only filter in filterConnections_ on root process.

blattms:
When filterConnections_ is called the grid is not load balanced, yet. Currently that means that grid() will also return the unbalanced grid and all processes will see the whole global grid.

We will change semantics of the unbalanced grid soon: Only the root process will see the whole grid and the others will see an empty partition of it. Hence filtering on this partition will remove all connections on all wells in the schedule for non-root processes and produce wrong results.

For non-root process the filtering needs to be done on the load balanced grid. This is accomplished by this commit.

This PR is needed for the upcoming rebase PR OPM/opm-grid#405 to produce correct results.

### DIFF
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -196,6 +196,15 @@ public:
             grid_->switchToDistributedView();
 
             cartesianIndexMapper_.reset();
+
+            if ( ! equilGrid_ )
+            {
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

blattms:
This really should change anything for the current master of opm-grid.

blattms:
jenkins run norne parallel please

PR #2165, opened 2019-11-18
Use EclEpsGridProperties constructor

joakim-hove:
Downstream of: https://github.com/OPM/opm-material/pull/363


joakim-hove:
jenkins build this opm-material=363 please

PR #2174, opened 2019-11-20
added: support rhel8

akva2:


blattms:
I am surely lacking the skills to review this, but I am curious.

Does this change mean that we drop rhel6 (as we drop devtools needed for it) and now just support the newer versions?

akva2:
No, the %{?!el8 means include this when el8 is not defined, without the ! it is included if el8 is defined.

redhat/opm-simulators.spec
```diff
@@ -114,14 +116,15 @@ Requires:       ptscotch-mpich
 %description mpich-bin
 This package contains the applications for opm-simulators
 
+%global debug_package %{nil}
+
 %prep
 %setup -q -n %{name}-release-%{version}-%{tag}
 
 %build
-scl enable devtoolset-6 bash
```
akva2: or was this what you were referring to markus? this line is void and a leftover after prior attempts. each line in the spec executes in a separate shell so it has no effect. this is why i have to pass the compilers manually below.

blattms:
thanks for the clarification.

PR #2175, opened 2019-11-21
fix issue when GRUP is used in WCONPROD and WCONINJE and NONE 

totto82:
Use a bhp control when a well is controlled by a invalided group control

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1999,7 +1999,7 @@ namespace Opm
         }
 
         if (!group.isInjectionGroup() || currentGroupControl == Group::InjectionCMode::NONE) {
-            // use bhp as control eq and let the updateControl code find a vallied control
+            // use bhp as control eq and let the updateControl code find a valid control
             const auto& controls = well.injectionControls(summaryState);
             control_eq = getSegmentPressure(0) - controls.bhp_limit;
             return;
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1998,8 +1998,12 @@ namespace Opm
             return;
         }
 
-        if (!group.isInjectionGroup())
+        if (!group.isInjectionGroup() || currentGroupControl == Group::InjectionCMode::NONE) {
+            // use bhp as control eq and let the updateControl code find a vallied control
+            const auto& controls = well.injectionControls(summaryState);
+            control_eq = getSegmentPressure(0) - controls.bhp_limit;
             return;
+        }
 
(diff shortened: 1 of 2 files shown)


totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2037,7 +2041,8 @@ namespace Opm
         switch(currentGroupControl) {
         case Group::InjectionCMode::NONE:
         {
-            OPM_DEFLOG_THROW(std::runtime_error, "NONE group control not implemented for injectors" , deferred_logger);
+            // The NONE case is handled earlier
+            assert(false);
```
bska: What is this supposed to do?  `Assert(false)` vanishes when `NDEBUG` is defined (i.e., in release mode).

totto82: This code part should never be called since the NONE case is handled just above as a special case. The assert is just to give a snap to the developer if it tries to remove the NONE handling above.

bska: > The assert is just to give a snap to the developer if it tries to remove the NONE handling above.

That's my point.  Developers who typically use release builds wont see `assert(false)` at all.

blattms:
I did look at `ḐeferredLogger` and kind of understand what you are trying to accomplish. You want to be able catch this situation in a debugger where it actually happens. That is not possible with `ḐeferredLogger` as it will not throw directly but only somewhen later. That is indeed cumbersome for debugging.

I guess a clean way to helper developers in debugging would be to add a compile time switch (#define) that tells `DeferredLogger` to throw directly.

If you are 100% sure that getting here is an unrecoverable bug and should never happen, you should probably simply throw an
exception as an intermediate workaround.

opm/simulators/wells/MultisegmentWell_impl.hpp
atgeirr: The assert() is an acceptable way to guard against developer error, even if release mode builds do not see it. I'll let this stand.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1998,8 +1998,12 @@ namespace Opm
             return;
         }
 
-        if (!group.isInjectionGroup())
+        if (!group.isInjectionGroup() || currentGroupControl == Group::InjectionCMode::NONE) {
+            // use bhp as control eq and let the updateControl code find a vallied control
```
atgeirr: Typo: valid
(also line 2128, and in StandardWell_impl.hpp)

totto82: Done

totto82:
jenkins build this please

PR #2177, opened 2019-11-22
Solve well as a preprocessing step at every iteration 

totto82:
This PR includes #2175.

This PR may affect performance and results for cases without groups and should be tested thoroughly.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -801,18 +801,14 @@ namespace Opm {
                 calculateExplicitQuantities(local_deferredLogger);
                 prepareTimeStep(local_deferredLogger);
             }
-            updateWellControls(local_deferredLogger, true);
+            updateWellControls(local_deferredLogger, /*allow for switching to group controls*/true);
 
-            // only check group controls for iterationIdx smaller then nupcol
+            // only update REIN and VREP rates if iterationIdx is smaller than nupcol
             const int reportStepIdx = ebosSimulator_.episodeIndex();
             const int nupcol = schedule().getNupcol(reportStepIdx);
             if (iterationIdx < nupcol) {
                 if( localWellsActive() ) {
                     const Group& fieldGroup = schedule().getGroup("FIELD", reportStepIdx);
-                    std::vector<double> groupTargetReduction(numPhases(), 0.0);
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

totto82:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.013    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.075    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.008    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 0.979    |



View result details @ https://www.ytelses.com/opm/?page=result&id=583

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -826,7 +822,7 @@ namespace Opm {
             std::vector< Scalar > B_avg(numComponents(), Scalar() );
             computeAverageFormationFactor(B_avg);
 
-            if (param_.solve_welleq_initially_ && iterationIdx == 0) {
+            if (param_.solve_welleq_initially_) {
```
atgeirr: As you write, this change should be well tested before merging. I assume it helps or is necessary for the case motivating the work. Have you tested any other complex models?

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1140,16 +1136,26 @@ namespace Opm {
...
         }
 
         for (const auto& well : well_container_) {
             well->updateWellControl(ebosSimulator_, well_state_, deferred_logger);
         }
+
+        // the group target reduction rates needs to be update since wells may have swicthed to/from GRUP control
+        // Currently the group targer reduction does not honor NUPCOL
+        if( localWellsActive() ) {
+            std::vector<double> groupTargetReduction(numPhases(), 0.0);
```
atgeirr: I think I understand why you have moved this block out from the nupcol-related if: to ensure that we change the group target reduction when a well is moved in/out of group control. But keeping the targets fixed for wells under group control should be ok (under the assumption that nupcol is "large enough" to make the targets reasonable). If wells A and B get rates 900 and 800 at the nupcol deadline, then B switches to BHP and produces 780, it is still reasonable that A produces 900 and the acceleration nupcol is supposed to give requires that we keep the target fixed.

totto82:
jenkins build this please

totto82:
This PR is replaced by #2210 and #2211

PR #2179, opened 2019-11-25
support NONE Exceed limit

totto82:
Nothing happens when the control is violated if NONE is specified as exceed limit.


totto82:
jenkins build this please

PR #2182, opened 2019-11-25
ParallelRestart: use references and allow non-default template parame…

blattms:
…ters.

packSize now uses a reference to map/unordered_map to prevent copying.
In addition we always specify all template parameters for the containers to
allow for sending data if non-default ones are used.

### DIFF
--- a/opm/simulators/utils/ParallelRestart.cpp
+++ b/opm/simulators/utils/ParallelRestart.cpp
@@ -90,8 +90,8 @@ std::size_t packSize(const std::pair<T1,T2>& data, Dune::MPIHelper::MPICommunica
     return packSize(data.first, comm) + packSize(data.second, comm);
 }
 
-template<class T>
-std::size_t packSize(const std::vector<T>& data, Dune::MPIHelper::MPICommunicator comm)
+template<class T, class A>
+std::size_t packSize(const std::vector<T,A>& data, Dune::MPIHelper::MPICommunicator comm)
 {
     if (std::is_pod<T>::value)
         // size written automatically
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please

atgeirr:
> I'm worried about all the potential overloading necessary for those additional template parameters, maybe defaulting to whatever the container default them to might make sense

I am not sure that is an issue here. Since the implementations are only in the cpp file, usage there will be the only way more instantiations are triggered. (Actually I do not understand why the declarations are in the hpp file, could they not be moved inside the cpp file?) Please explain if I misunderstood!

Anyway I will merge this now, given your approval and that Jenkins is green. Thanks!

akva2:
it's not intended to be private symbols only used for parallel restart. in fact, it would make sense to split this file in two; the restart related things and the serialization related things. i'm working on reusing the serialization for a lot of things at the moment to avoid having an ecl state on all processes.

atgeirr:
Should this then be the start of something like "opm/common/serialization/...", and provide general serialization features similar to the serialization libraries we have discussed? I did not think we had concluded yet, but if you are hard at work using this code for serialization I guess that is a kind of conclusion...

akva2:
it uses MPI for serialization so i don't think that is wanted?

atgeirr:
> it uses MPI for serialization so i don't think that is wanted?

No, I see that. But we do not have to use MPI for serialization to transport data by MPI messages, the job is the same in any case. Therefore we should work towards a common serialization approach, as any such would also work for MPI, right?

akva2:
yeah, agreed, we do not have to use MPI_Pack.

blattms:
Well, we do not, but currently using it has the following advantages
- no additional dependencies since MPI is already one.
- it is in the MPI standard and not going to change
- most probably well maintained and tested.

PR #2185, opened 2019-11-26
FOR TESTING: Solve well every continue

totto82:
This gathers all group related commits in one PR for easier testing. 
Depends on OPM/opm-common#1252 and OPM/opm-common#1258


totto82:
This PR is replaced by #2210 and #2211

PR #2186, opened 2019-11-26
changed: use a separate output dir in the parallel norne test

akva2:
this way we get curves for both the serial and parallel runs in the well plots.

requires https://github.com/OPM/opm-tests/pull/193

### DIFF
--- a/jenkins/run-norne.sh
+++ b/jenkins/run-norne.sh
@@ -9,7 +9,7 @@ cd norne
 mkdir $SIM
 if test -n "$1"
 then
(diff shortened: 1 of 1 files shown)


akva2:
jenkins run norne norne_parallel opm-tests=193 please

akva2:
jenkins run norne norne_parallel opm-tests=193 please

akva2:
this should be uncontroversial, proved working and is testing related so kinda my turf. shameless self-merge.

PR #2188, opened 2019-11-26
Use EclipseState based Schedule constructor

joakim-hove:
Small step in the 3D refactoring saga

### DIFF
--- a/tests/test_stoppedwells.cpp
+++ b/tests/test_stoppedwells.cpp
@@ -36,11 +36,7 @@ BOOST_AUTO_TEST_CASE(TestStoppedWells)
     Opm::Parser parser;
     Opm::Deck deck(parser.parseFile(filename));
     Opm::EclipseState eclipseState(deck);
-    const auto& grid = eclipseState.getInputGrid();
-    const TableManager table ( deck );
-    const Eclipse3DProperties eclipseProperties ( deck , table, grid);
-    const Opm::Runspec runspec (deck);
-    const Schedule sched(deck, grid, eclipseProperties, runspec);
+    const Schedule sched(deck, eclipseState);
 
     // Both wells are open in the first schedule step
     {


joakim-hove:
jenkins build this please

joakim-hove:
OK - the issue with bogus SATNUM values should be fixed here: #2191 (not tested yet)

But the norne init time is still an issue. Looking into it.

joakim-hove:
The issue with init time is probably related to this: https://github.com/OPM/opm-material/pull/365 looking into it.

joakim-hove:
OK: https://github.com/OPM/opm-material/pull/368 should fix the init time.

joakim-hove:
Where can I find the relperm diagnostic output which @bska posted further up?

akva2:
you have to run norne.

joakim-hove:
> you have to run norne, then inspect console output.

OK - thanks; and that seems to be "No news is good news".

blattms:
Is this constructor supposed to be used elsewhere instead of the one using the deck and the grid, too (i.e. apart from tests)?

joakim-hove:
> Is this constructor supposed to be used elsewhere instead of the one using the deck and the grid, too

That is just a matter of convenience; but if you have an `EclipseState` handy at callsite I would say yes. The reason for this small change is the ongoing 3D refactor; with this PR opm-simulators can be fully shielded from code changes due to that refactor.

blattms:
But the grid we are simulating one might have subtle difference to the one
in the EclipseState. For a sequnetial run just recall the issues we saw when
we tried to remove filterConnections. In a parallel run we are only using a
part of the grid. Will that work?

joakim-hove:
>But the grid we are simulating one might have subtle difference to the one in the EclipseState

Ahhh ... I had forgot. We'll stick to updating tests.

PR #2189, opened 2019-11-27
Add ascii diff of changes to data update PR

akva2:
requested here; https://github.com/OPM/opm-tests/pull/192

Downstream of https://github.com/OPM/opm-common/pull/1263

### DIFF
--- a/tests/update_reference_data.sh
+++ b/tests/update_reference_data.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
 
 OPM_TESTS_ROOT=$1
+TMPDIR=`mktemp -d`
+mkdir $TMPDIR/orig
+mkdir $TMPDIR/new
 
 # Copy results from a test run to refence dir
 # $1 = source directory to copy data from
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this opm-common=1263 update_data please

akva2:
jenkins build this opm-common=1263 update_data please

akva2:
slight road bump. i am using curl to post the comment and it easily exceeds the max command line. i'll have change to some other tool to interact with github, will take some more time.

akva2:
would it be acceptable to instead upload the diff to the CI server? that would be a lot easier. @bska @joakim-hove opinions.

bska:
> would it be acceptable to instead upload the diff to the CI server? that would be a lot easier

That's perfectly fine.  Would you additionally be able to automatically post a link to the diff in a Github comment?

akva2:
no, that would be a chicken-hen problem. but it's easily found, you just go to the details under the check and it will be right there.

joakim-hove:
> would it be acceptable to instead upload the diff to the CI server? that would be a lot easier.

Of course. I guess I should be slightly more careful with suggestions and loose thoughts - you jump fast!

akva2:

akva2:
okay, see under details now. sadly this approach will not pick up all changes. in particular, lots of files were changed but only the forced file appears in the diff since the formatted files are identical.

joakim-hove:
> okay, see under details now

The "data_diff" view? Looks fine - we'll leave it here. I was just voicing a question which I thought could be valuable - not a real need.

PR #2192, opened 2019-11-27
Added cusparseSolver, requires NVIDIA GPU

Tongdongq:
This branch contains a linear solver, implemented in cusparse, which can run on an NVIDIA GPU.
To use it, CUDA_ENABLED must be set to true and the path to the CUDA install must be fixed at the top of CMakeLists.txt. The user must pass
```sh
--matrix-add-well-contributions=true --use-gpu=true
```
to the executable in order to use it. If the gpu is selected, but the well contributions are not added, an error is printed and the program exits. Only NORNE is tested, which generates blocks of 3x3. Other blocksizes do not work, but might in the future.

Some notes:
- based on release/2019.04
- now exits when `matrix-add-well-contribs` is `false` and `use_gpu` is true, should it print a warning instead and fallback to the cpu?
- forward template declarations in `BdaBridge.cpp` are a not very nice
- converting from `BCRSMatrix` to our format and back assumes underlying data is contiguous, testing shows that it is, but is it guaranteed?
- the part from `backend->initialize()` to `backend->solve_system()` should probably be moved to within the backend.
- why `BdaResult`? Because the `cusparseSolverBackend.cu` is compiled using nvcc, adding the `InverseOperatorResult` there causes a compile error:
```
/data/work/tqiu/dune/dune-istl/dune/istl/basearray.hh:101:58: error: ‘typename Dune::Imp::base_array_unmanaged<B, A>::RealIterator’ names ‘template<class B, class A> template<class T> struct Dune::Imp::base_array_unmanaged<B, A>::RealIterator’, which is not a type
       friend class RealIterator<const ValueType>;
```
- timing functions in `cusparseSolverBackend` use `gettimeofday()`
- BdaBridge misuses the std::vectors a bit.
- the CUDA compiler nvcc defines the macro `__NVCC__`, if not defined, the `BdaBridge` will be mostly empty.

---

## Results ##
Intel Xeon E5-2620 v3 @ 2.4GHz, NVIDIA RTX GeForce 2080Ti

### Dune: ###
```
Total time (seconds):         774.473
Solver time (seconds):        767.034
Assembly time (seconds):     298.541 (Failed: 3.21499; 1.0769%)
Linear solve time (seconds): 443.508 (Failed: 5.63787; 1.2712%)
Linear solve setup time (seconds): 14.3272 (Failed: 0.184655; 1.28884%)
Update time (seconds):       11.2785 (Failed: 0.130548; 1.1575%)
Output write time (seconds): 9.36335
Overall Well Iterations:      918 (Failed: 2; 0.217865%)
Overall Linearizations:       1932 (Failed: 21; 1.08696%)
Overall Newton Iterations:    1598 (Failed: 21; 1.31414%)
Overall Linear Iterations:    23982 (Failed: 308; 1.2843%)
```

### cusparseSolver: ###
```
Total time (seconds):         460.21
Solver time (seconds):        452.812
Assembly time (seconds):     293.769 (Failed: 6.40129; 2.17902%)
Linear solve time (seconds): 134.676 (Failed: 4.4542; 3.30735%)
Linear solve setup time (seconds): 14.3 (Failed: 0.37382; 2.61413%)
Update time (seconds):       11.2345 (Failed: 0.259182; 2.30703%)
Output write time (seconds): 9.33874
Overall Well Iterations:      921 (Failed: 4; 0.434311%)
Overall Linearizations:       1940 (Failed: 42; 2.16495%)
Overall Newton Iterations:    1605 (Failed: 42; 2.61682%)
Overall Linear Iterations:    24387 (Failed: 861; 3.53057%)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,8 +14,17 @@
 #                                                                         #
 ###########################################################################
 
+
+set(CUDA_ENABLED false)
+
 # Mandatory call to project
-project(opm-simulators C CXX)
+if(CUDA_ENABLED)
+  project(opm-simulators LANGUAGES CUDA C CXX)
+  # cuda_runtime.h
+  include_directories(/path/to/cuda/include)
+else()
+  project(opm-simulators LANGUAGES C CXX)
+endif()
 
 cmake_minimum_required (VERSION 2.8)
 
@@ -273,3 +282,8 @@ if (BUILD_EBOS_EXTENSIONS)
   install(TARGETS ebos_thermal DESTINATION bin)
 endif()
 
+# must link libraries after target 'flow' has been defined
+if(CUDA_ENABLED)
+  target_link_libraries( flow ${CUDA_cublas_LIBRARY} )
+  target_link_libraries( flow ${CUDA_cusparse_LIBRARY} )
+endif()
--- a/opm/bda/BdaBridge.cpp
+++ b/opm/bda/BdaBridge.cpp
@@ -1,0 +1,240 @@
+
+
+#include <memory>
+
+//#define HAVE_NULLPTR 1 // otherwise dune/common/nullptr.hh will not compile
+
+#include <opm/bda/BdaBridge.hpp>
+#include <opm/bda/BdaResult.hpp>
+
+#define PRINT_TIMERS_BRIDGE_BRIDGE 0
+
+typedef Dune::InverseOperatorResult InverseOperatorResult;
+
+namespace Opm
+{
+
+BdaBridge::BdaBridge(bool use_gpu_, int maxit, double tolerance) : use_gpu(use_gpu_){
+#ifdef __NVCC__
+    if(use_gpu){
+    	backend = new cusparseSolverBackend(maxit, tolerance);
+    }
+#endif
+}
+
+BdaBridge::~BdaBridge(){
+#ifdef __NVCC__
+	if(use_gpu){
+		delete backend;
+	}
+#endif
+}
+
+#ifdef __NVCC__
+template <class BridgeMatrix>
+int checkZeroDiagonal(BridgeMatrix& mat) {
+	static std::vector<int> diag_indices;   // contains offsets of the diagonal nnzs
+	int numZeros = 0;
+	const int dim = 3;
+	const double zero_replace = 1e-15;
+	double *nnzs = &(mat[0][0][0][0]);
+	if(diag_indices.size() == 0){
+		int N = mat.N()*dim;
+		diag_indices.reserve(N);
+		for(typename BridgeMatrix::const_iterator r = mat.begin(); r != mat.end(); ++r){
+			for(auto c = r->begin(); c != r->end(); ++c){
+				if(r.index() == c.index()){
+					for(int rr = 0; rr < dim; ++rr){
+						// pointer arithmetic
+						int offset = (int)((long unsigned)&(mat[r.index()][c.index()][rr][rr]) - (long unsigned)nnzs); // in bytes
+						offset /= sizeof(double);  // convert offset to doubles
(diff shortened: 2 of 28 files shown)


blattms:
Thanks a lot. as you can see, this not mergable. To make us able to merge you need to base you work on current master, e.g. by
```
git checkout master
git pull
git checkout -b new-fancy-branch
git cherry-pick 97c1537 
```
Then push and open a new PR.

CMakeLists.txt
```diff
@@ -14,8 +14,17 @@
...
 
+
+set(CUDA_ENABLED false)
+
 # Mandatory call to project
-project(opm-simulators C CXX)
+if(CUDA_ENABLED)
+  project(opm-simulators LANGUAGES CUDA C CXX)
+  # cuda_runtime.h
+  include_directories(/path/to/cuda/include)
```
blattms: Whether we need a  CUDA_ENABLED switch is maybe a matter of taste. Up to now I was not aware that one could specify CUDA as a language to `project`. If that imlicitly run find_package(CUDA) one coulde use the `-DCMAKE_DISABLE_FIND_PACKAGE_CUDA=ON` switch  to disable cuda.

It seems like CUDA was added as a language from version 3.12 but we are supporting Ubuntu bionic which only ship the 3.10 version. Maybe we should leave the `project` as is and run `find_package(CUDA)` afterwards or even better specify it in the opm-simulators-prereqs.cmake file.

Also the include path used seems to be wrong on most systems.  Maybe we should use the `CUDA_INCLUDE_DIRECTORIES` variable instead?

opm/bda/BdaBridge.cpp
```diff
@@ -0,0 +1,240 @@
...
+#define PRINT_TIMERS_BRIDGE_BRIDGE 0
+
+typedef Dune::InverseOperatorResult InverseOperatorResult;
+
+namespace Opm
+{
+
+BdaBridge::BdaBridge(bool use_gpu_, int maxit, double tolerance) : use_gpu(use_gpu_){
+#ifdef __NVCC__
+    if(use_gpu){
```
blattms: Are you sure this file will be compiled with nvcc and not the c++compiler?

opm/bda/cusparseSolverBackend.hpp
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright 1993-2015 NVIDIA Corporation.  All rights reserved.
+ *
```
blattms: Is this copyright disclaimer correct? Does not look like open source.

Tongdongq: We overlooked this. The header is not correct, it should have read 'This software contains source code provided by NVIDIA Corporation.' because the EULA specifies that we can distribute our derivation, while mentioning that we based it on their samples.
However, digging deeper, it seems that the NVIDIA EULA is not compatible with GPL.

CMakeLists.txt
Tongdongq: I think using `find_package(CUDA)` is much better. I'll try adding a HAVE_CUDA to the opm-simulators-prereqs.cmake file. This will also solve your second point, because .cpp files are indeed not compiled by `nvcc`.

blattms:
I have another question concerning your choice of matrix type:
Is there a special reason to use plain compressed sparse row format (CSR) and not block compressed sparse row format (BSR). The latter is provided by cusparse and would seem more natural for case. Is it not well supported by the preconditioners?

Tongdongq:
I don't think we are using plain CSR. The function convertMatrixBsr() in BdaBridge copies the underlying nonzeroes from the BCRSMatrix, which are assumed to be contiguous. The colIndices are put into a contiguous array using the iterator.
cusparse indeed supports BSR.

PR #2195, opened 2019-11-28
changed: use a macro to declare prototypes in ParallelRestart.hpp

akva2:
more readable, less tedious to add new types.

at least in my opinion. feel free to reject.

### DIFF
--- a/opm/simulators/utils/ParallelRestart.hpp
+++ b/opm/simulators/utils/ParallelRestart.hpp
@@ -32,6 +32,7 @@
 #include <vector>
 #include <map>
 #include <unordered_map>
+
 namespace Opm
 {
 namespace Mpi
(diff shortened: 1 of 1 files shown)


opm/simulators/utils/ParallelRestart.hpp
```diff
@@ -221,32 +164,28 @@ template<class T1, class T2, class H, class P, class A>
...
+ADD_PACK_PROTOTYPES(data::CellData)
+ADD_PACK_PROTOTYPES(data::Connection)
+ADD_PACK_PROTOTYPES(data::Rates)
+ADD_PACK_PROTOTYPES(data::Segment)
+ADD_PACK_PROTOTYPES(data::Solution)
+ADD_PACK_PROTOTYPES(data::Well)
+ADD_PACK_PROTOTYPES(data::WellRates)
+ADD_PACK_PROTOTYPES(RestartKey)
+ADD_PACK_PROTOTYPES(RestartValue)
+ADD_PACK_PROTOTYPES(std::string)
```
blattms: Would have been easier if the old ordering would have been used. Seems like at least `const char*` is missing.

Otherwise I am neutral.

akva2: const char* has another signature so it's not included.

blattms: Indeed. 
From the diff it seemes as the lines 
```
void unpack(char* str, std::size_t length, std::vector<char>& buffer, int& position,
            Dune::MPIHelper::MPICommunicator comm);
```
were removed. But they were added again below . Did not see that due the reordering

blattms:
Slipped through the cracks. Merging.

PR #2196, opened 2019-11-28
Add option for throwing for small time steps

totto82:
Default is zero. -> should not change anything if not provoked.

### DIFF
--- a/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
+++ b/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
@@ -27,6 +27,7 @@ NEW_PROP_TAG(SolverRestartFactor);
 NEW_PROP_TAG(SolverGrowthFactor);
 NEW_PROP_TAG(SolverMaxGrowth);
 NEW_PROP_TAG(SolverMaxTimeStepInDays);
+NEW_PROP_TAG(SolverMinTimeStep);
 NEW_PROP_TAG(SolverMaxRestarts);
 NEW_PROP_TAG(SolverVerbosity);
 NEW_PROP_TAG(TimeStepVerbosity);
@@ -45,6 +46,7 @@ SET_SCALAR_PROP(FlowTimeSteppingParameters, SolverRestartFactor, 0.33);
 SET_SCALAR_PROP(FlowTimeSteppingParameters, SolverGrowthFactor, 2.0);
 SET_SCALAR_PROP(FlowTimeSteppingParameters, SolverMaxGrowth, 3.0);
 SET_SCALAR_PROP(FlowTimeSteppingParameters, SolverMaxTimeStepInDays, 365.0);
+SET_SCALAR_PROP(FlowTimeSteppingParameters, SolverMinTimeStep, 0.0);
 SET_INT_PROP(FlowTimeSteppingParameters, SolverMaxRestarts, 10);
 SET_INT_PROP(FlowTimeSteppingParameters, SolverVerbosity, 1);
 SET_INT_PROP(FlowTimeSteppingParameters, TimeStepVerbosity, 1);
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

joakim-hove:
image


`ParseContext` versus `TypeTags` - something to consider?

opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
```diff
@@ -147,6 +151,8 @@ namespace Opm {
                                  "The maximum factor time steps are elongated after a report step");
             EWOMS_REGISTER_PARAM(TypeTag, double, SolverMaxTimeStepInDays,
                                  "The maximum size of a time step in days");
+            EWOMS_REGISTER_PARAM(TypeTag, double, SolverMinTimeStepInDays,
+                                 "The minimum size of a time step in days, before the simulator throw");
```
atgeirr: Typo: throws

opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
```diff
@@ -551,6 +570,7 @@ namespace Opm {
         double growthFactor_;                //!< factor to multiply time step when solver recovered from failed convergence
         double maxGrowth_;                   //!< factor that limits the maximum growth of a time step
         double maxTimeStep_;                //!< maximal allowed time step size
+        double minTimeStep_;                //!< mainimal allowed time step size before throwing
```
atgeirr: typo: minimal

opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
atgeirr: I changed my mind on this, I do not think we should document this (to the users) as "throwing" at all. I think we should rather say something like: "The minimum time step size in days. If a step cannot converge without getting cut below this step size the simulator will stop."

Also, we should make sure that when chopping the step, we set it to max(current step size * factor, minimum step size), and only terminate after actually trying the minimum step size. So instead of step sizes going (with 1 day minimum) for example 3 days -> 1.5 days -> terminate (as 0.75 days < 1 day) we go 3 days -> 1.5 days -> 1 day -> terminate.

bska: > The minimum time step size in days.

That will be very inconvenient for LAB units (time unit hours).

joakim-hove:
This is a bit frustrating. We do not have very good tradition for discussing technical solutions, and details like whitespace and naming get a disproportionaly large share of the review effort.

For this particular review I tried to initiate a discussion with a bang - but there was certainly a real point behind. The `ParseContext` mechanism is a quite flexible method to achieve user/runtime configurable error handling.

I am not saying it is necessarily a good idea to start using it more - although I think it is something to consider - but I do think it proves my point when at least the initial review comments to this PR was purely esthetic, and that my question is totally ignored.

Not that my question/comment here is particularly important, and Attgeir is getting unfair flak right now (sorry Attgeir) - but I do think that our culture for improving technical solutions in cooperation should improve. It is unfortunate if we are in a situation where everyone is mostly focused on her current task and without attention to the entire codebase. Kjør debatt!

atgeirr:
I only partially understood what you meant by pitting ParseContext vs TypeTag, although I could have guessed you meant how it relates to error handling; your initial comment was a bit too brief to base a discussion on without asking what you meant (and it was getting late...). 

Now that it is clarified, I do have a few arguments against expanding ParseContext usage (apart from the need to give it a new name): a) In its current usage, it is quite centralized, with the key strings being static members of the ParseContext class. This is done in order to avoid the possibility of an unnoticed typo leading to undetected subtly wrong error handling. Using the addKey() method in a more free way would be possible but carry the risk of mismatches. With the parameter system, the change is localized and the mismatch risk is not there. b) The parameter system provides the excellent `--help` option making it easy to discover and understand the option.

I am not saying those two arguments are impossible to address, or that there are no other arguments, or that the parameter system is perfect; for example, it is not possible to have a dynamic hierarchy of options as we would like for linear solvers (and how would we deal with the `--help` feature for such things anyway).

joakim-hove:
> I only partially understood what you meant by pitting ParseContext vs TypeTag, although I could have guessed you meant how it relates to error handling; 

I did not mean very much; my trigger point was that this PR added a user configurable option/error handler - and that the `ParseContext` has handled such things quite well - IMHO; and that I at least have found the existing parameter system quite cumbersome to use - for all it is worth.

But the main point I tried to raise with my second post is that I think that our approach to technical discussions should be improved - in general I think lot's of good things would result if more people joined in on technical discussions. Currently most discussions only involve two persons - what does the silent majority think?

**Error handling**
I did chime in on this PR because I think error handling is important - apart from the actual simulation results that is probably the most important way the users interact with flow. To get a really consistent error handling across the whole application I argue that we would need some application wide framework, and I think that (a suitably renamed and extended) `ParseContext` could be one such alternative - maybe as a singleton like the logging facility. Not really part of this PR - but at least something I would encourage people to think of.

totto82:
The motivation for this PR was to add a trigger to stop simulations that takes ages to finish. This was asked for by a colleague that uses Flow for ensemble runs.  I will fix the minor comments and typos, and switch to seconds.

atgeirr:
On the topic of user-configurable error handling, I agree that ParseContext is a contender. I do not see all that many cases where we would want it in the simulator, which makes it feasible to standardise. A pattern I can see right now is: something "bad" happens, and we can let the user choose to stop outright, cut the timestep, or continue with the current solution iterate ignoring the error. This could be applied to linear solver failures for example.

For the current PR, I'll jenkins it and merge if green.

atgeirr:
jenkins build this please

PR #2197, opened 2019-11-28
Load aquifer data

bska:
This PR adds a new member function,
```C++
AquiferInterface::initFromRestart()
```
that consumes a `vector<data::AquiferData>` constructed from information in the restart file's `XAAQ` vector.  At the moment, we use only the total produced liquid volume and the aquifer pressure.

We implement the interface's member function in terms of the virtual function
```C++
AquiferInterface::assignRestartData()
```
that must be overridden in derived classes.

Implement a trivial such function for Carter-Tracy aquifers, and a function that only stores the aquifer pressure for the Fetkovich aquifer model.

Additionally, record whether or not the aquifer object was initialised from a previous solution.  If so, don't reset total produce liquid volumes or aquifer pressures to their base values from the model input file.

We thread calls to this function through from the restart information loaded in `EclWriter::beginRestart()`.

---

This PR depends on OPM/opm-common#1277.

### DIFF
--- a/ebos/eclbaseaquifermodel.hh
+++ b/ebos/eclbaseaquifermodel.hh
@@ -31,6 +31,8 @@
 
 #include <opm/models/utils/propertysystem.hh>
 
+#include <exception>
+#include <stdexcept>
 #include <vector>
 
 BEGIN_PROPERTIES
@@ -78,7 +80,12 @@ public:
      *        volume from the model's aquifers.
      */
     void initFromRestart(const std::vector<data::AquiferData>& aquiferSoln OPM_UNUSED)
-    { }
+    {
+        throw std::logic_error {
+            "Initialization from restart data not supported "
+            "for base aquifer model"
+        };
+    }
 
     /*!
      * \brief This method is called when a new episode (report step) starts.
--- a/ebos/eclbaseaquifermodel.hh
+++ b/ebos/eclbaseaquifermodel.hh
@@ -27,7 +27,11 @@
 #ifndef EWOMS_ECL_BASE_AQUIFER_MODEL_HH
 #define EWOMS_ECL_BASE_AQUIFER_MODEL_HH
 
+#include <opm/output/data/Aquifer.hpp>
+
 #include <opm/models/utils/propertysystem.hh>
+
+#include <vector>
 
 BEGIN_PROPERTIES
 
@@ -62,6 +66,18 @@ public:
      *        condition has been applied.
      */
     void initialSolutionApplied()
+    { }
+
+    /*!
+     * \brief Called if aquifers are being initialized from values retrieved
+     *        from a restart file.
+     *
+     * \param[in] aquiferSoln Set of aquifer-related initial values, mostly
+     *        pertaining to analytic aquifers.  Contains at minimum the
+     *        aquifer pressure and the base run's total produced liquid
+     *        volume from the model's aquifers.
+     */
+    void initFromRestart(const std::vector<data::AquiferData>& aquiferSoln OPM_UNUSED)
     { }
 
     /*!
--- a/opm/simulators/aquifers/AquiferCarterTracy.hpp
+++ b/opm/simulators/aquifers/AquiferCarterTracy.hpp
@@ -23,6 +23,8 @@
 
 #include <opm/simulators/aquifers/AquiferInterface.hpp>
 
+#include <opm/output/data/Aquifer.hpp>
+
 namespace Opm
 {
 
@@ -144,6 +146,9 @@ namespace Opm
                     : ( connection.influx_multiplier.at(idx) * Base::faceArea_connected_.at(idx) )/denom_face_areas;
                 }
             }
+
+            void assignRestartData(const data::AquiferData& /* xaq */) override
+            {}
 
             inline void getInfluenceTableValues(Scalar& pitd, Scalar& pitd_prime, const Scalar& td)
             {
--- a/opm/simulators/aquifers/AquiferCarterTracy.hpp
+++ b/opm/simulators/aquifers/AquiferCarterTracy.hpp
@@ -25,6 +25,9 @@
 
 #include <opm/output/data/Aquifer.hpp>
 
+#include <exception>
+#include <stdexcept>
+
 namespace Opm
 {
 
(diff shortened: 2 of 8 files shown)


bska:
Marking this as ready for review now.  The "load-from-restart" situation is mostly handled now.  We're missing the part that constructs actual aquifer model objects directly from the restart file, but this is sufficient to get going if the `.DATA` file of a restart model also supplies the aquifer connections in its `SOLUTION` section.

bska:
jenkins build this please

opm/simulators/aquifers/AquiferCarterTracy.hpp
```diff
@@ -145,6 +147,9 @@ namespace Opm
                 }
             }
 
+            void assignRestartData(const data::AquiferData& /* xaq */) override
```
GitPaean: I think we have not finished the restarting implementation for AquiferCarterTracy.  Maybe it is good to throw a message here.

opm/simulators/aquifers/AquiferInterface.hpp
```diff
@@ -74,6 +76,26 @@ namespace Opm
...
 
+    void initFromRestart(const std::vector<data::AquiferData>& aquiferSoln)
+    {
+      auto xaqPos = std::find_if(aquiferSoln.begin(), aquiferSoln.end(),
+        [this](const data::AquiferData& xaq) -> bool
+      {
+        return xaq.aquiferID == this->connection_.aquiferID;
+      });
+
+      if (xaqPos == aquiferSoln.end()) {
```
GitPaean: This is a tricky situation.  By definition, aquifer keywords belong to SOLUTION section and they should not be altered after simulation.  If a restart running can not find the aquifer data in the RESTART file, basically, it means some mistake has happened.  I suggest we throw here.

opm/simulators/aquifers/BlackoilAquiferModel.hpp
```diff
@@ -29,11 +29,17 @@
 #include <opm/parser/eclipse/EclipseState/AquiferCT.hpp>
 #include <opm/parser/eclipse/EclipseState/Aquifetp.hpp>
 #include <opm/parser/eclipse/EclipseState/Aquancon.hpp>
-#include <opm/simulators/timestepping/SimulatorTimer.hpp>
+
+#include <opm/output/data/Aquifer.hpp>
+
 #include <opm/simulators/aquifers/AquiferCarterTracy.hpp>
 #include <opm/simulators/aquifers/AquiferFetkovich.hpp>
+#include <opm/simulators/timestepping/SimulatorTimer.hpp>
```
GitPaean: I could not see where we need this header file. Please confirm that we need it.

ebos/eclbaseaquifermodel.hh
```diff
@@ -64,6 +68,18 @@ public:
...
+     * \brief Called if aquifers are being initialized from values retrieved
+     *        from a restart file.
+     *
+     * \param[in] aquiferSoln Set of aquifer-related initial values, mostly
+     *        pertaining to analytic aquifers.  Contains at minimum the
+     *        aquifer pressure and the base run's total produced liquid
+     *        volume from the model's aquifers.
+     */
+    void initFromRestart(const std::vector<data::AquiferData>& aquiferSoln OPM_UNUSED)
+    { }
```
GitPaean: Maybe we should throw here too to let users know we do not do this yet.

opm/simulators/aquifers/AquiferFetkovich.hpp
```diff
@@ -23,6 +23,11 @@ along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 
 #include <opm/simulators/aquifers/AquiferInterface.hpp>
 
+#include <opm/output/data/Aquifer.hpp>
```
GitPaean: This is really nitpicking. I think it is sufficient to include it from the base class. 

It can be a style thing. You do not have to address this one. 

And similar applies to the two header files below. I think we only need `stdexcept`.

GitPaean:
Thanks for the efforts. It looks good. Only some minor comments.

OPMUSER:
@GitPaean 

> This is a tricky situation. By definition, aquifer keywords belong to SOLUTION section and they should not be altered after simulation. If a restart running can not find the aquifer data in the RESTART file, basically, it means some mistake has happened. I suggest we throw here.

This is not true, in the commercial simulator it is possible to change the aquifer properties in the SCHEDULE section, for example the `AQUFETP – DEFINE FETKOVICH ANALYTICAL AQUIFERS`. Now whether the engineers should be able to do this is another question, but you can bet a dollar that if they are allowed to do it some of them will :-).

GitPaean:
> This is not true, in the commercial simulator it is possible to change the aquifer properties in the SCHEDULE section, for example the AQUFETP – DEFINE FETKOVICH ANALYTICAL AQUIFERS.

Okay.  Then OPM's implementation of AQUFETP might have some flaw in term of making it dynamic.


It is the keyword `AQUANCON` can not be changed. To me, it even creates more complicated situation. 

AQUANCON

opm/simulators/aquifers/AquiferFetkovich.hpp
bska: > This is really nitpicking. I think it is sufficient to include it from the base class.
> 
> It can be a style thing. You do not have to address this one.

It is true, we do get that transitively from `AquiferInterface.hpp`.  On the other hand, I've recently come to prefer "include what you use" and the new implementation definitely uses/refers to  members of the structures declared in `opm/output/data/Aquifer.hpp`.

ebos/eclbaseaquifermodel.hh
bska: > Maybe we should throw here too to let users know we do not do this yet.

I'll update accordingly.

opm/simulators/aquifers/BlackoilAquiferModel.hpp
bska: > I could not see where we need this header file. Please confirm that we need it.

You're right.  We don't use that header anywhere in this file.  I'll remove it.

opm/simulators/aquifers/AquiferCarterTracy.hpp
bska: > think we have not finished the restarting implementation for AquiferCarterTracy

That is correct.

> Maybe it is good to throw a message here.

Okay.  I'll update accordingly.

bska:
I have tried to address the review comments in an update commit.  I am going to rerun the build test now.

bska:
jenkins build this please

GitPaean:
@bska  I do not think you pushed the update.

bska:
> @bska I do not think you pushed the update.

:frowning_face: 

Yes, you're right.  I've pushed the update now.  Sorry.

GitPaean:
jenkins build this please

GitPaean:
Thanks for the work. Merging now.

bska:
> Merging now.

Much appreciated!

PR #2198, opened 2019-11-29
Add PORO to testdata

joakim-hove:
See: https://github.com/OPM/opm-common/issues/1271


joakim-hove:
jenkins build this please

PR #2200, opened 2019-11-29
List the matching wells for Action eval

joakim-hove:
When an action condition has been evaluated the set of wells matching the condition is available; that can be used in place of `'?'` in the `WELOPEN` keyword. This PR will just include the matching wells in the log message.


joakim-hove:
jenkins build this please

joakim-hove:
Needed for debugging - quite trivial. Merging

PR #2203, opened 2019-12-02
Update output from applyActions() to show date

joakim-hove:
Update message from `applyAction()` to show actual date when action is tested.


joakim-hove:
jenkins build this please

joakim-hove:
objections?

atgeirr:
No objections, merging. Although std::put_time() could be considered for date formatting in the future? I guess that would take a little work in order to use consistently though.

PR #2206, opened 2019-12-03
SummaryState is updated at the end of simulator timestep

joakim-hove:
This is needed to ensure that the `SummaryState` is fully updated before calling the `applyAction`code.


joakim-hove:
jenkins build this please

blattms:
So for the initial writeout (before the first timestep he call to `evalSummaryState` is not needed.

PR #2207, opened 2019-12-03
Refactor WellState SHUT

joakim-hove:
- Move some WellState shut code to separate method.
- Add bool flag open_for_output_ to WellState to ensure that shut wells do not
  create output.

Part of the effort to be able to handle ACTIONX in the middle of a report step.

### DIFF
--- a/opm/simulators/wells/WellState.hpp
+++ b/opm/simulators/wells/WellState.hpp
@@ -63,6 +63,7 @@ namespace Opm
                 const int nw = wells_ecl.size();
                 // const int np = wells->number_of_phases;
                 const int np = pu.num_phases;
+                open_for_output_.resize(nw, true);
                 bhp_.resize(nw, 0.0);
                 thp_.resize(nw, 0.0);
                 temperature_.resize(nw, 273.15 + 20); // standard temperature for now
@@ -152,6 +153,17 @@ namespace Opm
             return wellRates().size() / numWells();
         }
 
+
+        virtual void shutWell(int well_index) {
+            this->open_for_output_[well_index] = false;
+            this->thp_[well_index] = 0;
+            this->bhp_[well_index] = 0;
+            const int np = numPhases();
+            for (int p = 0; p < np; ++p)
+                this->wellrates_[np * well_index + p] = 0;
+        }
+
+
         virtual data::Wells report(const PhaseUsage& pu, const int* globalCellIdxMap) const
         {
             using rt = data::Rates::opt;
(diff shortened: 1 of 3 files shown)


joakim-hove:
jenkins build this please

opm/simulators/wells/WellState.hpp
```diff
@@ -152,13 +152,28 @@ namespace Opm
...
 
+
+        void shutWell(int well_index) {
+            this->open_for_output_[well_index] = false;
+            this->thp_[well_index] = 0;
+            this->bhp_[well_index] = 0;
+            const int np = numPhases();
+            for (int p = 0; p < np; ++p) {
+                this->wellrates_[np * well_index + p] = 0;
+                this->perfrates_[np * well_index + p] = 0;
```
bska: What's the size of `perfrates_`?  This looks to be treating it as an `np`-per-well quantity, but from the name I would expect `np`-per-reservoir connection size instead.

joakim-hove: I agree that looks suspicious; I have just copied the code verbatim from the original location in `BlackoilWellModel_impl.hpp` - but it might have been incorrect there as well.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
OK - my understanding of what happens in the failing test is the following:

1. There is logic in the opm-common/output where the the value of `IWEL[8]` is set depending on whether the well is flowing or not. That logic is currently based on the output data structure `opm:output::Wells` where a well is flowing if one of rates is explicitly set - disgregarding the value.

2. In this PR wells which are closed due to WECON are no longer passed to output - previously they were passed with explicit numeric zero. 

The end result is that the logic in opm-common/output comes out differently. I would argue that the new behavior is more correct, and that a update_data is warranted - but others might disagree. See also https://github.com/OPM/opm-common/pull/1288 which independently addresses similar problems from opm-common.

atgeirr:
I have not looked in detail at the failure but have a question that could be related: the open_for_output_ flag gets set to false for a well being shut, but never gets set to true again when a well is reopened (unless at a later report step, then it is all default again), should we not do that?

joakim-hove:
I'll initiate the update_data process - but not merge anything yet.

joakim-hove:

joakim-hove:
> well is reopened (unless at a later report step, then it is all default again), should we not do that?

yes we should - but right now we effectively only have support for *closing* wells; I can add a comment to that effect.

joakim-hove:
jenkins build this opm-tests=200 please

joakim-hove:
jenkins build this opm-tests=200 please

bska:
I think this is fine it its current state and behaves the way it's supposed to.  I am going to merge this into master now.  Thanks a lot.

PR #2212, opened 2019-12-06
WIP Changes to solveWellEq

totto82:


GitPaean:
From the commit message, it looks like it does not try to switch the well control during the solution. Then it is basically iterate the well equations until converged.

Then the suggestion is that this procedure can go to each individual well.   Actually, for MSW, it has already done so, which was extremely important for the current performance.  For each iteration, the well equations need to iterate until converged before the final assembling to form the final matrix. 

For StandardWell, we also tried that, while back then, it did not bring anything useful, then we did not use it. Maybe due to some recent changes, it can be useful now. 

The reason that solveWellEq was called in the level of BlackoilWellModel, it was due to the group control back then. We wanted to get a good initial guess for group control.  If it does not touch group control, then each individual well can do it by itself.  If it does not touch well control even, then it is an even smaller or more incremental change compared with `solveWellEq` for each iteration. It touches less things, it is really good to test in an incremental manner.

totto82:
If I understand you correctly, you suggest to add 
```
        const bool use_inner_iterations = param_.use_inner_iterations_std_wells_;
        if (use_inner_iterations) {
            solveWellEqUntilConverged(....);
        }
```
in assembleWellEq and remove the solveWellEq part completely.  I agree that if the changes suggested in this PR is going into the code it should be done the way you outlined. But especially the "don't update wellControl part" needs some more testing before it can be merged. I will do some more testing during the weekend.

GitPaean:
> But especially the "don't update wellControl part" needs some more testing before it can be merged. I will do some more testing during the weekend.

Sure. We should decide the wanted behavoir first, then the implementation can be detailed.  If group control is not involved in this process, I guess it is probably OKay to not switch the control during the inner iteration.  One argument is that unconverged well solution might cause undesired control switching.  Why do not we wait a little bit until the solution converged. 

I do observe that the commercial simulator switches the well controls less during the solution procedure (converged result can be the same), there is something nice about it.

totto82:
With this I am able to run through the standard norne case without any failures!

GitPaean:
> Also now solveWellEq can be completely removed from the code.

Without considering the group control, one of the difference here is the explicit quantities.  For some cases it does not matter,  for some, it might matter.

totto82:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.001    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.92     |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 0.997    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.001    |



View result details @ https://www.ytelses.com/opm/?page=result&id=600

PR #2213, opened 2019-12-06
Simulate with only active cells

joakim-hove:
**Update:** This is now starting to shape up, and a merge is not so far away.My plan right now is as follows:

1. Regularly rebase this and the upstream PR's in opm-material and opm-common, let it linger for about a week while I periodically bitch about testing.

2. Update the development branches with review comments[1].

3. Wham-bang-merge - around January 10.th.

When the code has been merged, I intend to clean up opm-common, by completely removing the old `Eclipse3DProperties` implementation and also remove all the temporary build flags. Finally there will be a clean up in opm-simulators/opm-material. For a parallell simulation there are three levels of active cells: 

1. It is *all* `nx * ny * nz` cells in the cartesian box - this is what I typically call *global* in code I write.
2. It is the total set of active cells in the model.
3. It is the set of active cells for this particular process.

Previously opm-common created 3D properties of the kind 1 above, and then opm-simulators/opm-material extracted the interesting properties down to the set of active cells *for the active process* - i.e. the transition 1 &rightarrow; 3. The new implementation is on level 2 above, the transition 2 &rightarrow; 3 is quite simple to implement, but to keep the amount of changes down the current code creates a temporary global vector and then subsequently reuses the existing 1 &rightarrow; 3 implementation. The next refactor in opm-simulators/opm-material is to go directly 2 &rightarrow; 3.

[1]: There are several comments and requests for change in this PR which focus on uneccesary copies due to the temporary global vector. I do not intend to address these comments in the initial merge, but they will certainly be addressed in the followup PR which will come quite soon.

When all of this is done I would like to refactor the `namespace keywords {   } ` in opm-common/FieldProps.cpp - my ambition is to generate that static data from the json files - as part of the build process. 

Upstream: https://github.com/OPM/opm-common/pull/1298 https://github.com/OPM/opm-material/pull/380

-----------------

**Warning:** This is so immature it is not even funny; but it at least "compiles" (and then crashes ...) on my machine. Is not yet really ready for testing yet, but that is hopefully quite close. 

To actually test this you need:

https://github.com/OPM/opm-common/pull/1298

And all modules must be compiled with the cmake switch `-DENABLE_3DPROPS_TESTING=ON` and `-DBUILD_TESTING=OFF` (the tests do not compile yet ....).

### DIFF
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -300,7 +300,12 @@ protected:
 
         auto & field_props = this->eclState().fieldProps();
         const_cast<FieldPropsManager&>(field_props).reset_actnum(actnum);
+        int active_cells = 0;
+        for (const int a : actnum)
+            active_cells += a;
+        printf("Calling reset_actnum PE: %d   sum(actnum): %d\n", mpiRank, active_cells);
 #endif
+
     }
 
     // removing some connection located in inactive grid cells
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -258,12 +258,7 @@ public:
 protected:
     void createGrids_()
     {
-#ifdef ENABLE_3DPROPS_TESTING
         const auto& porv = this->eclState().fieldProps().porv(true);
-#else
-        const auto& gridProps = this->eclState().get3DProperties();
-        const std::vector<double>& porv = gridProps.getDoubleGridProperty("PORV").getData();
-#endif
         grid_.reset(new Dune::CpGrid());
         grid_->processEclipseFormat(&(this->eclState().getInputGrid()),
                                     /*isPeriodic=*/false,
@@ -284,7 +279,6 @@ protected:
             equilCartesianIndexMapper_.reset(new CartesianIndexMapper(*equilGrid_));
         }
 
-#ifdef ENABLE_3DPROPS_TESTING
         std::vector<int> actnum;
         unsigned long actnum_size;
         if (mpiRank == 0) {
@@ -300,8 +294,6 @@ protected:
 
         auto & field_props = this->eclState().fieldProps();
         const_cast<FieldPropsManager&>(field_props).reset_actnum(actnum);
-#endif
-
     }
 
     // removing some connection located in inactive grid cells
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -284,7 +284,6 @@ protected:
             equilCartesianIndexMapper_.reset(new CartesianIndexMapper(*equilGrid_));
 
 #ifdef ENABLE_3DPROPS_TESTING
-            grid_->switchToGlobalView();
             auto actnum = Opm::UgGridHelpers::createACTNUM(*grid_);
             auto & field_props = this->eclState().fieldProps();
             const_cast<FieldPropsManager&>(field_props).reset_actnum(actnum);
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -300,8 +300,6 @@ protected:
 
         auto & field_props = this->eclState().fieldProps();
         const_cast<FieldPropsManager&>(field_props).reset_actnum(actnum);
-        auto meminfo = field_props.meminfo();
-        printf("%ld/%ld  int: %ld  double: %ld  size:%ld bytes (%ld MB)\n", meminfo.global_size, meminfo.active_size, meminfo.int_fields, meminfo.double_fields, meminfo.total, meminfo.total / (1024*1024));
 #endif
 
     }
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -282,6 +282,13 @@ protected:
         {
             equilGrid_.reset(new Dune::CpGrid(*grid_));
             equilCartesianIndexMapper_.reset(new CartesianIndexMapper(*equilGrid_));
+
+#ifdef ENABLE_3DPROPS_TESTING
+            grid_->switchToGlobalView();
+            auto actnum = Opm::UgGridHelpers::createACTNUM(*grid_);
+            auto & field_props = this->eclState().fieldProps();
+            const_cast<FieldPropsManager&>(field_props).reset_actnum(actnum);
+#endif
         }
 
 
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -282,14 +282,7 @@ protected:
         {
             equilGrid_.reset(new Dune::CpGrid(*grid_));
             equilCartesianIndexMapper_.reset(new CartesianIndexMapper(*equilGrid_));
-
-#ifdef ENABLE_3DPROPS_TESTING
-            auto actnum = Opm::UgGridHelpers::createACTNUM(*grid_);
-            auto & field_props = this->eclState().fieldProps();
-            const_cast<FieldPropsManager&>(field_props).reset_actnum(actnum);
-#endif
         }
-
 
 #ifdef ENABLE_3DPROPS_TESTING
         std::vector<int> actnum;
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -300,10 +300,8 @@ protected:
 
         auto & field_props = this->eclState().fieldProps();
         const_cast<FieldPropsManager&>(field_props).reset_actnum(actnum);
-        int active_cells = 0;
-        for (const int a : actnum)
-            active_cells += a;
-        printf("Calling reset_actnum PE: %d   sum(actnum): %d\n", mpiRank, active_cells);
+        auto meminfo = field_props.meminfo();
+        printf("%ld/%ld  int: %ld  double: %ld  size:%ld bytes (%ld MB)\n", meminfo.global_size, meminfo.active_size, meminfo.int_fields, meminfo.double_fields, meminfo.total, meminfo.total / (1024*1024));
 #endif
 
     }
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -2835,7 +2835,10 @@ private:
 
 
         if (enableSolvent) {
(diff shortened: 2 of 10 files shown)


ebos/eclcpgridvanguard.hh
```diff
@@ -257,9 +257,13 @@ public:
...
     void createGrids_()
     {
+#ifdef ENABLE_3DPROPS_TESTING
+        const auto& fp = this->eclState().fieldProps();
+        const auto& porv = fp.porv(true);
+#else
         const auto& gridProps = this->eclState().get3DProperties();
         const std::vector<double>& porv = gridProps.getDoubleGridProperty("PORV").getData();
-
+#endif
```
blattms: Does this need a PR in opm-grid? Shouldn't the size of porv differ and call for a different processing in `grid_->processEclipseFormat`?

joakim-hove: It is *very* much WIP; right now I am just trying to get it to compile ... ;-(

joakim-hove:
OK - for all it is worth it seems to run Norne; so it is not vaporware at least.

alfbr:
Please advice if it makes sense to test it at this point.

joakim-hove:
You shouldn't bet your wife on it just yet - but a couple of beers is it worth. I'd say give it a spin.. 😀

alfbr:
Excellent, will do!

ebos/equil/initstateequil.hh
```diff
@@ -945,17 +945,29 @@ public:
...
             swatInit_.resize(nc);
-            const int* gc = Opm::UgGridHelpers::globalCell(grid);
-            for (int c = 0; c < nc; ++c) {
-                const int deckPos = (gc == NULL) ? c : gc[c];
-                swatInit_[c] = swatInitEcl[deckPos];
+
+#ifdef ENABLE_3DPROPS_TESTING
+            if (eclipseState.fieldProps().has<double>("SWATINIT")) {
+                const std::vector<double>& swatInitEcl = eclipseState.fieldProps().get<double>("SWATINIT");
+                for (int c = 0; c < nc; ++c)
```
blattms: Is there a special reason for manual copying instead of
`swatInit_ = eclipseState.fieldProps().get<double>("SWATINIT");

ebos/equil/initstateequil.hh
```diff
@@ -1084,9 +1096,11 @@ private:
     void updateInitialTemperature_(const Opm::EclipseState& eclState)
     {
         // Get the initial temperature data
-        const std::vector<double>& tempiData =
-            eclState.get3DProperties().getDoubleGridProperty("TEMPI").getData();
-
+#ifdef ENABLE_3DPROPS_TESTING
+        std::vector<double> tempiData = eclState.fieldProps().get_global<double>("TEMPI");
```
blattms: Superfluous copy.

ebos/equil/initstateequil.hh
```diff
@@ -1109,8 +1123,11 @@ private:
         std::vector<int> cellPvtRegionIdx(numCompressed);
 
         //Get the PVTNUM data
+#ifdef ENABLE_3DPROPS_TESTING
+        const auto pvtnumData = eclState.fieldProps().get_global<double>("PVTNUM");
```
blattms: Why the copy? Even if a temporary is returned, one can extend its lifetime long enough by
```
const auto& pvtnumData = eclState.fieldProps().get_global<double>("PVTNUM");
```

Judging from the code below the vector has cartesian size like before (in contrast to other properties). That might come as a suprise.

ebos/equil/initstateequil.hh
joakim-hove: A guiding principle for these changes have to a large extent been to minimize the changes; at least until things get through and "work" - but this one was maybe exaggerating it. I have updated to just do `swatInit_ = ...`.

ebos/equil/initstateequil.hh
joakim-hove: For both of these cases I have - initially - opted to use a global vector of cartesian size - just to keep the amount of changes down. The global vectors are by nature copies/temporaries, and then I like to make that clear by capturing the return value in a value object; I would assume that both the value object and the value held alive a with a const reference would have the lifetime - so memory wise it should make no difference?

But I'd be happy to change to `const &`- but maybe even more importantly is that this should eventually be a `get<double>( )` which returns a compressed vector instead of `get_global<double>( )`.

ebos/ecltransmissibility.hh
```diff
@@ -859,22 +899,28 @@ private:
...
         if (!opmfil)
             return;
 
-        const auto& porv = eclState.get3DProperties().getDoubleGridProperty("PORV").getData();
-        const auto& actnum = eclState.get3DProperties().getIntGridProperty("ACTNUM").getData();
         const auto& cartMapper = vanguard_.cartesianIndexMapper();
         const auto& cartDims = cartMapper.cartesianDimensions();
+#ifdef ENABLE_3DPROPS_TESTING
+        const auto& actnum = eclState.fieldProps().actnum();
+        const auto& porv = eclState.fieldProps().porv(true);
```
blattms: Confusing that actnum and porv are vectors of cartesian size. How can a developer know which properties have which size?

ebos/eclcpgridvanguard.hh
```diff
@@ -278,6 +282,13 @@ protected:
         {
             equilGrid_.reset(new Dune::CpGrid(*grid_));
             equilCartesianIndexMapper_.reset(new CartesianIndexMapper(*equilGrid_));
+
+#ifdef ENABLE_3DPROPS_TESTING
+            grid_->switchToGlobalView();
```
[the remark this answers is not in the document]
joakim-hove: OK good; that I did not know. Fixed

ebos/eclproblem.hh
```diff
@@ -2667,29 +2696,47 @@ private:
         const auto& cartSize = simulator.vanguard().cartesianDimensions();
         size_t numCartesianCells = cartSize[0] * cartSize[1] * cartSize[2];
 
-        std::vector<double> waterSaturationData;
+        std::vector<double> waterSaturationData(numCartesianCells, 0.0);
```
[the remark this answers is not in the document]
joakim-hove: Fixed

ebos/ecltransmissibility.hh
joakim-hove: For someone with an unhealthy interest in Eclipse file formats this is very natural, but I see your point. The `ACTNUM` vector is a uncompressed vector of `nx*ny*nz` elements of 0 and 1; where 0 denotes an inactive cell and 1 denotes an active cell - to actually fill this role it must be of full length. I am not arguing this representation is the best way to identify active and inactive cells, but it is way it is done in both the input and output files - and familiar to developers who have worked with these files. 

The `PORV` vector is of course mainly the porevolume of the respective cells, but it also serves as a "floating point ACTNUM vector". The `INIT` file contains many properties like `PERMX, PORO` and `SWL` which are all represented as compressed vectors; and in addition the `INIT` file contains the `PORV` vector as an uncompressed vector. In the uncompressed PORV vector elements with value zero identify inactive cells; so in the output domain inspecting `PORV` from the `INIT` file is an alternative to inspecting `ACTNUM` from the `EGRID` file.

In this code the `ACTNUM` vector is *always* uncompressed; but for `PORV` you can adjust it with the boolean flag.

joakim-hove:
When it comes to the new copies you have pointed out I agree with your point; and they should eventually go, but I would prefer to let them linger because that makes the code changes less invasive. I expect this PR to hang around for some time.

joakim-hove:
OK - this now at least starts simulating Norne with MPI.

ebos/ecltransmissibility.hh
blattms: Maybe we should put this somewhere (e.g. wiki) for reference for people who try to live healthy :laughing:

joakim-hove:
OK - there are excessive amounts of duck-tape chewing gum here; but it actually seems to pull through NORNE with MPI and get identical iteration numbers as master. Lo and behold

joakim-hove:
jenkins build this opm-common=1298 opm-material=380 please

joakim-hove:
I am closing this now - to draw attention to the updated PR description. Will reopen again immediately.

joakim-hove:
jenkins build this opm-common=1298 opm-material=380 please

joakim-hove:
OK; as seen from the PR in opm-common there is now only one failing test (`SPE1CASE2_THERMAL`). That fails with some very small numerical differences which I have not been able to track down. I have no other suggestion than minute differences due to order of computation or similar - cool if someone else takes a look.

I will look a bit further, but currently I am inclined to handle this with an `update_data`.

Again: As summarized in the PR description I am now getting ready to merge this - cool if someone wants to take a look.

atgeirr:
> I will look a bit further, but currently I am inclined to handle this with an `update_data`.

Just asking a few questions regarding this:
 - Are there any other cases that are *different* (just not enough to fail tests)?
 - Are there changed code sites that we know of, that should make us expect (insignificant or not) result changes?

joakim-hove:
> Are there any other cases that are different (just not enough to fail tests)?

I have not checked all the cases - but for the series of `model2/` tests there is bitwise equality.

> Are there changed code sites that we know of, that should make us expect (insignificant or not) result changes?

No - not that I am aware of.

atgeirr:
> > Are there any other cases that are different (just not enough to fail tests)?
> 
> I have not checked all the cases - but for the series of `model2/` tests there is bitwise equality.

Then I think it is probably ok to update_data.

As this is (and should be) a quite significant change we should ensure a bit more than the automated tests run ok. From the discussion I see Norne (with MPI also) is ok, are any more large cases known good with this?

> 
> > Are there changed code sites that we know of, that should make us expect (insignificant or not) result changes?
> 
> No - not that I am aware of.

That is always annoying... It should be bitwise ok, but is not. It is probably not worth the effort to track down if it is only the thermal case, as long as results are fully equivalent.

joakim-hove:
>  [...] are any more large cases known good with this?

No not really - a large Equinor case is verified to run, but that case is still worked and was not in a state for comparisons.

> That is always annoying... It should be bitwise ok, but is not. 

I agree on that - have not entirely given up on tracking it down, but do not have any smoke right now.

atgeirr:
Can be merged at-will.

PR #2216, opened 2019-12-09
Update the internal ACTNUM in the FieldPropsManager

joakim-hove:
When cells have been deactivated we need to call out to the property contanier with the new active cells map.

### DIFF
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -258,9 +258,12 @@ public:
 protected:
     void createGrids_()
     {
+#ifdef ENABLE_3DPROPS_TESTING
+        const auto& porv = this->eclState().fieldProps().porv(true);
+#else
         const auto& gridProps = this->eclState().get3DProperties();
         const std::vector<double>& porv = gridProps.getDoubleGridProperty("PORV").getData();
-
+#endif
         grid_.reset(new Dune::CpGrid());
         grid_->processEclipseFormat(this->eclState().getInputGrid(),
                                     /*isPeriodic=*/false,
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this opm-grid=428 please

blattms:
ok, just for understanding this:
At this stage the fieldprops arrays are already compressed (to the extend that the Grid representation after parsing the deck is)? This second pass is just to catch cells that got inactive because of operations
in the simulator during setup.

joakim-hove:
> At this stage the fieldprops arrays are already compressed (to the extend that the Grid representation after parsing the deck is)? This second pass is just to catch cells that got inactive because of operations
in the simulator during setup.

That is 100% correct!

joakim-hove:
OK - this has received som MPI love and attention. Merge?

joakim-hove:
jenkins build this please

joakim-hove:
I realize the MPI calls are probably not in the way we should do it; my main MPI experience was from the time I was a bachelor - now I am divorced and the kids have gone off to university.

Nevertheless this little piece of code has worked on my machine for some time, I "need" to get it in to reduce the number of in-flight PRs for the 3D refactoring; it is 100% protected by the `#ifdef ENABLE_3DPROPS_TESTING` - so all in all I will put on my asbestos and merge it. When jenkins agrees.

ebos/eclcpgridvanguard.hh
```diff
@@ -280,6 +283,27 @@ protected:
...
+#ifdef ENABLE_3DPROPS_TESTING
+        std::vector<int> actnum;
+        unsigned long actnum_size;
+        if (mpiRank == 0) {
+            actnum = Opm::UgGridHelpers::createACTNUM(*grid_);
+            actnum_size = actnum.size();
+        }
+
+#ifdef HAVE_MPI
+        MPI_Bcast(&actnum_size, 1, MPI_UNSIGNED_LONG, 0, MPI_COMM_WORLD);
```
blattms: please use `grid().comm().bcast(&actnum_size, 1, 0)``then you can skip the ifdef

blattms:
I am wondering why this step is actually needed on all processors, But of course I have not the faintest idea how the size of the properties is in a parallel run. Should it the number of active cell in the global grid or the active cells in the local partition of the distributed grid. At least in the latter case there
is no need to broadcast here if you do the compression only on rank 0.

PR #2220, opened 2019-12-11
[For testing] Enable "owners first" ordering. Altered SpMV, SP and ILU.

andrthu:
This PR require https://github.com/OPM/opm-grid/pull/430 in opm-grid, and in addition to what is contained there, the following is changed:

- Command line argument --owners-first, that turn on/off "owners first" ordering of cells in the parallel grid. Default is off.

- SpMV, inner product and ILU that utilise the "owners first" ordering, and therefore avoids some superfluous computations.

On dual Intel Xeon E5-2670 2.6GHz 8-core CPU nodes this time, some modest improvement in parallel performance was achieved on Norne:

| P | master: time / iter     | owners-first: time / iter  |
|---|--------------------| -----------------------|
| 2  |  426.57 / 25339     |    408.73  /  25179        |
| 4  |     259.07 / 25571    |      245.40 / 25209      |
| 8  |     181.15 / 25026     |     174.39 / 25656       |
| 16 |   129.76 / 25738     |     119.90 / 25227         |
| 32 |    98.59 / 26317     |     87.27 / 26113            |
| 64 |    77.36 / 26769     |     69.73 / 26726          |

The change in iteration count is due to the reordering and ILU being sensitive to that.

This PR was created for testing on models I don't have access to. A more elegant implementation that avoids having to copy the matrix, by altering the assembly part of the code would probably achieve similar performance without having to deal with the "owners first" ordering stuff.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -78,7 +78,7 @@ NEW_PROP_TAG(EclOutputInterval);
 NEW_PROP_TAG(IgnoreKeywords);
 NEW_PROP_TAG(EnableExperiments);
 NEW_PROP_TAG(EdgeWeightsMethod);
-NEW_PROP_TAG(OwnersFirst);
+NEW_PROP_TAG(OwnerCellsFirst);
 
 SET_STRING_PROP(EclBaseVanguard, IgnoreKeywords, "");
 SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "");
@@ -87,7 +87,7 @@ SET_BOOL_PROP(EclBaseVanguard, EnableOpmRstFile, false);
 SET_BOOL_PROP(EclBaseVanguard, EclStrictParsing, false);
 SET_BOOL_PROP(EclBaseVanguard, SchedRestart, true);
 SET_INT_PROP(EclBaseVanguard, EdgeWeightsMethod, 1);
-SET_BOOL_PROP(EclBaseVanguard, OwnersFirst, false);
+SET_BOOL_PROP(EclBaseVanguard, OwnerCellsFirst, false);
 
 END_PROPERTIES
 
@@ -135,7 +135,7 @@ public:
                              "When restarting: should we try to initialize wells and groups from historical SCHEDULE section.");
         EWOMS_REGISTER_PARAM(TypeTag, int, EdgeWeightsMethod,
                              "Choose edge-weighing strategy: 0=uniform, 1=trans, 2=log(trans).");
-        EWOMS_REGISTER_PARAM(TypeTag, bool, OwnersFirst,
+        EWOMS_REGISTER_PARAM(TypeTag, bool, OwnerCellsFirst,
                              "Order cells owned by rank before ghost/overlap cells.");
     }
 
@@ -269,7 +269,7 @@ public:
 
         std::string fileName = EWOMS_GET_PARAM(TypeTag, std::string, EclDeckFileName);
         edgeWeightsMethod_   = Dune::EdgeWeightMethod(EWOMS_GET_PARAM(TypeTag, int, EdgeWeightsMethod));
-        ownersFirst_ = EWOMS_GET_PARAM(TypeTag, bool, OwnersFirst);
+        ownersFirst_ = EWOMS_GET_PARAM(TypeTag, bool, OwnerCellsFirst);
 
         // Skip processing of filename if external deck already exists.
         if (!externalDeck_)
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -78,6 +78,7 @@ NEW_PROP_TAG(EclOutputInterval);
 NEW_PROP_TAG(IgnoreKeywords);
 NEW_PROP_TAG(EnableExperiments);
 NEW_PROP_TAG(EdgeWeightsMethod);
+NEW_PROP_TAG(OwnersFirst);
 
 SET_STRING_PROP(EclBaseVanguard, IgnoreKeywords, "");
 SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "");
@@ -86,6 +87,7 @@ SET_BOOL_PROP(EclBaseVanguard, EnableOpmRstFile, false);
 SET_BOOL_PROP(EclBaseVanguard, EclStrictParsing, false);
 SET_BOOL_PROP(EclBaseVanguard, SchedRestart, true);
 SET_INT_PROP(EclBaseVanguard, EdgeWeightsMethod, 1);
+SET_BOOL_PROP(EclBaseVanguard, OwnersFirst, false);
 
 END_PROPERTIES
 
@@ -133,6 +135,8 @@ public:
                              "When restarting: should we try to initialize wells and groups from historical SCHEDULE section.");
         EWOMS_REGISTER_PARAM(TypeTag, int, EdgeWeightsMethod,
                              "Choose edge-weighing strategy: 0=uniform, 1=trans, 2=log(trans).");
+        EWOMS_REGISTER_PARAM(TypeTag, bool, OwnersFirst,
+                             "Order cells owned by rank before ghost/overlap cells.");
     }
 
     /*!
@@ -265,6 +269,7 @@ public:
 
         std::string fileName = EWOMS_GET_PARAM(TypeTag, std::string, EclDeckFileName);
         edgeWeightsMethod_   = Dune::EdgeWeightMethod(EWOMS_GET_PARAM(TypeTag, int, EdgeWeightsMethod));
+        ownersFirst_ = EWOMS_GET_PARAM(TypeTag, bool, OwnersFirst);
 
         // Skip processing of filename if external deck already exists.
         if (!externalDeck_)
@@ -451,6 +456,13 @@ public:
      */
     Dune::EdgeWeightMethod edgeWeightsMethod() const
     { return edgeWeightsMethod_; }
+
+    /*!
+     * \brief Parameter that decide if cells owned by rank are ordered before ghost cells.
+     */
+    bool ownersFirst() const
+    { return ownersFirst_; }
+    
     /*!
      * \brief Returns the name of the case.
      *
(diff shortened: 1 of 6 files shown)


blattms:
jenkins build this opm-grid=430 please

blattms:
jenkins run norne_parallel opm-grid=430 please

opm/simulators/linalg/ParallelOverlappingILU0.hpp
```diff
@@ -1036,5 +1095,365 @@ class ParallelOverlappingILU0
...
+
+        // iterator types
+        typedef typename Range ::block_type  dblock;
+        typedef typename Domain::block_type  vblock;
+
+        const size_type iEnd = lower_.rows();
+        const size_type lastRow = iEnd - 1;
+        size_type interiorStart = iEnd - interiorSize_;
+        size_type lowerLoopEnd = interiorSize_;
+        if( iEnd != upper_.rows() )
```
blattms: This looks like code duplication. Where is the original code coming from? Could we factor it (the loops) out to function that get an index or iterator as the end and just use the correct one here and the default at other places.

opm/simulators/linalg/findOverlapRowsAndColumns.hpp
```diff
@@ -110,6 +110,28 @@ namespace detail
...
+            return grid.numCells();
+        const auto& gridView = grid.leafGridView();
+        auto elemIt = gridView.template begin<0>();
+        const auto& elemEndIt = gridView.template end<0>();
+
+        // loop over cells in mesh
+        for (; elemIt != elemEndIt; ++elemIt) {
+            const auto& elem = *elemIt;
+
+            // If cell has partition type not equal to interior save row
```
blattms: Comment seems misleading.

opm/simulators/linalg/findOverlapRowsAndColumns.hpp
```diff
@@ -110,6 +110,28 @@ namespace detail
...
+        size_t numInterior = 0;
+        if (!ownerFirst)
+            return grid.numCells();
+        const auto& gridView = grid.leafGridView();
+        auto elemIt = gridView.template begin<0>();
+        const auto& elemEndIt = gridView.template end<0>();
+
+        // loop over cells in mesh
+        for (; elemIt != elemEndIt; ++elemIt) {
+            const auto& elem = *elemIt;
```
blattms: In general (not a CpGrid) this might cost something. Maybe prevent it?

opm/simulators/linalg/findOverlapRowsAndColumns.hpp
```diff
@@ -110,6 +110,28 @@ namespace detail
             }
         }
     }
+    template <class Grid>
+    size_t numInteriorCells(const Grid& grid, bool ownerFirst)
+    {
+        size_t numInterior = 0;
+        if (!ownerFirst)
```
blattms: Add least with the PreconditionerFactory it seems to be possible to construct a ParallelOverlappingILU0 in a serial run. Maybe change to `if (!ownerFirst || grid.com().size()==1)`

ebos/eclthresholdpressure.hh
```diff
@@ -313,7 +313,7 @@ private:
                 const auto& intersection = *isIt;
 
                 // ignore boundary intersections for now (TODO?)
-                if (intersection.boundary())
+                if (!intersection.neighbor())
```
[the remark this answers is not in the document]
blattms: On second thought I would prefer this to be:
```
if(intersection.boundary())
  continue; // ignore boundary intersections for now (TODO?)
else if(!neighbour) //processor boundary but not domain boundary
  continue;
```
This would make clear that the boundary treatment might change, but the processor boundary not. Probably splitting hairs here.

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -240,15 +432,23 @@ class WellModelMatrixAdapter : public Dune::AssembledLinearOperator<M,X,Y>
...
 
-            detail::setWellConnections(gridForConn, wellsForConn, useWellConn, wellConnectionsGraph_);
-            detail::findOverlapAndInterior(gridForConn, overlapRows_, interiorRows_);
-            if (gridForConn.comm().size() > 1) {
-                noGhostAdjacency();
-                setGhostsInNoGhost(*noGhostMat_);
+            ownersFirst_ = EWOMS_GET_PARAM(TypeTag, bool, OwnersFirst);
+            interiorCellNum_ = detail::numInteriorCells(simulator_.vanguard().grid(), ownersFirst_);
+
+            if (!ownersFirst_) {
```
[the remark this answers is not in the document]
andrthu: setWellConnections, findOverlapAndInterior, noGhostAdjacency and setGhostsInNoGhost are only needed for making copying of the Jacobean possible. With --owners-first=true a copy of the Jacobean is not needed, and information about the sparsity pattern is not needed.

flow --owners-first=true --matrix-add-well-connections=true works fine since all well contributions are located in the interior/owner rows.

opm/simulators/linalg/ParallelOverlappingILU0.hpp
```diff
@@ -747,13 +841,15 @@ class ParallelOverlappingILU0
 
         const size_type iEnd = lower_.rows();
         const size_type lastRow = iEnd - 1;
+        size_type interiorStart = iEnd - interiorSize_;
```
blattms: Name is a bit misleading. `upperLoopStart` or `noGhosts`?

opm/simulators/linalg/ParallelOverlappingILU0.hpp
```diff
@@ -1023,7 +1122,7 @@ class ParallelOverlappingILU0
     //! \brief The relaxation factor to use.
     const field_type w_;
     const bool relaxation_;
-
+    size_type interiorSize_;
```
blattms: I do not have a better name, but interiorSize_ for something that might be CpGrid::numCells() sounds wrong.

opm/simulators/linalg/findOverlapRowsAndColumns.hpp
```diff
@@ -112,6 +112,28 @@ namespace detail
             }
         }
     }
+    template <class Grid>
+    size_t numInteriorCells(const Grid& grid, bool ownerFirst)
```
blattms: It is not always the number of interiorCells. At least there should be a comment about this.

andrthu:
SpMV and ILU can utilise the owners first ordering, while CPR (or AMG) can not. Therefore we need to modify the matrix by setting copy rows to 1 on the diagonal when using CPR and AMG, even if --owner-cells-first=true.

ebos/eclbasevanguard.hh
```diff
@@ -77,13 +77,15 @@ NEW_PROP_TAG(EclOutputInterval);
...
 NEW_PROP_TAG(EdgeWeightsMethod);
+NEW_PROP_TAG(OwnersFirst);
 
 SET_STRING_PROP(EclBaseVanguard, IgnoreKeywords, "");
 SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "");
 SET_INT_PROP(EclBaseVanguard, EclOutputInterval, -1); // use the deck-provided value
 SET_BOOL_PROP(EclBaseVanguard, EnableOpmRstFile, false);
 SET_BOOL_PROP(EclBaseVanguard, EclStrictParsing, false);
 SET_INT_PROP(EclBaseVanguard, EdgeWeightsMethod, 1);
+SET_BOOL_PROP(EclBaseVanguard, OwnersFirst, false);
```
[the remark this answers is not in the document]
blattms: Seems like this was changed to OwnerCellsFirst.

blattms:
Would it make sense to add a warning if --owner-cellls-first=true is ignored? Currently one cannot judge this without inside knowledge.

opm/simulators/linalg/ParallelOverlappingILU0.hpp
andrthu: I believe this was resolved by removing the GhostLastParallaelOverlappingILU0 class.

opm/simulators/linalg/findOverlapRowsAndColumns.hpp
andrthu: This is added.

opm/simulators/linalg/findOverlapRowsAndColumns.hpp
andrthu: Just to be clear you men the line: (const auto& elem = *elemIt;)? 

It is now removed.

opm/simulators/linalg/findOverlapRowsAndColumns.hpp
andrthu: name changed to numMatrixRowsToUseInSolver.

opm/simulators/linalg/ParallelOverlappingILU0.hpp
andrthu: Changed to upperLoppStart.

ebos/eclcpgridvanguard.hh
```diff
@@ -191,6 +192,8 @@ public:
             //distribute the grid and switch to the distributed view.
             {
                 const auto wells = this->schedule().getWellsatEnd();
+                auto& eclState = static_cast<ParallelEclipseState&>(this->eclState());
```
blattms: These two lines seem to be left overs from a rebase. Please delete as this in now two lines below in the try-clause

opm/simulators/linalg/ParallelOverlappingILU0.hpp
blattms: Nevermind

ebos/eclcpgridvanguard.hh
andrthu: Now removed.

blattms:
jenkins run norne_parallel opm-grid=430 please

blattms:
Shoot, did not run serial norne.

blattms:
But everything looks reasonable with the default (--owner-cells-first=true) which should not change anything. Merging with upstream.

PR #2222, opened 2019-12-11
Make ebos compile with PolyhedralGrid again.

dr-robertk:
This PR fixes various issues when switching ebos from CpGrid to PolyhedralGrid.

### DIFF
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -148,6 +148,7 @@ class EclWriter
     typedef typename GET_PROP_TYPE(TypeTag, Vanguard) Vanguard;
     typedef typename GET_PROP_TYPE(TypeTag, GridView) GridView;
     typedef typename GET_PROP_TYPE(TypeTag, Grid) Grid;
+    typedef typename GET_PROP_TYPE(TypeTag, EquilGrid) EquilGrid;
     typedef typename GET_PROP_TYPE(TypeTag, Scalar) Scalar;
     typedef typename GET_PROP_TYPE(TypeTag, ElementContext) ElementContext;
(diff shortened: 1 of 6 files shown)


dr-robertk:
@blattms: I removed the renaming of Equil to GlobalIO. It was indeed not entirely necessary.

dr-robertk:
jenkins build this please

ebos/eclwriter.hh
```diff
@@ -476,7 +480,7 @@ private:
 
     Opm::data::Solution computeTrans_(const std::unordered_map<int,int>& cartesianToActive) const
     {
-        const auto& cartMapper = simulator_.vanguard().cartesianIndexMapper();
+        const auto& cartMapper = simulator_.vanguard().equilCartesianIndexMapper();
```
dr-robertk: @totto82: I think this is a potential bug, but I'm not 100% sure.

dr-robertk:
jenkins build this please

blattms:

ebos/eclwriter.hh
blattms: Yes that seems like a bug. The cartesian indices are used to check whether all cells between vertical neighbors are inactive (if not it would be an NNC).

totto82:
```
        const auto& cartMapper = simulator_.vanguard().cartesianIndexMapper();
        const auto& cartDims = cartMapper.cartesianDimensions();
        const int globalSize = cartDims[0]*cartDims[1]*cartDims[2];
```
It is only used to get the globalSize. That is the same for both IndexMappers I think.

blattms:
Seems like this needs a rebase due to the merged DUNE 2.7 PR.

blattms:
jenkins build this please

blattms:
Thanks.

PR #2226, opened 2019-12-14
Use Dune communicator instead of raw MPI call

joakim-hove:
Fixup to #2216


joakim-hove:
jenkins build this please

PR #2229, opened 2019-12-16
check groups from top

totto82:
Check group constraints from the top to allow for groups down in the hierarchy to switch to its own control.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1684,12 +1684,7 @@ namespace Opm {
     BlackoilWellModel<TypeTag>::
     checkGroupConstraints(const Group& group, Opm::DeferredLogger& deferred_logger) {
 
-        // call recursively
         const int reportStepIdx = ebosSimulator_.episodeIndex();
-        for (const std::string& groupName : group.groups()) {
-            checkGroupConstraints( schedule().getGroup(groupName, reportStepIdx), deferred_logger);
-        }
-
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

totto82:
This PR shouldn't change anything other than the order of some calculations. Strange that Jenkins is complaining.

totto82:

totto82:
I have verified the test failure. The udq_wconprod test has two levels so it is affected by this change. The results are the same, but the iterations path is altered.

totto82:
@atgeirr Can this be merged?

atgeirr:
> @atgeirr Can this be merged?

When green, yes.

totto82:
jenkins build this opm-tests=212 please

bska:
> > @atgeirr Can this be merged?
> 
> When green, yes.

The SPE9_CP_GROUP case is **probably** okay, but UDQ_WCONPROD needs more justification before we merge this.  In particular, the `IWEL` differences suggest that the "UP02" and "WIL01" wells at some point in the history were shut but in this PR become open.  Also, the DOUBHEAD differences are massive enough that the timestepping sequence needs more investigation.

totto82:
If you use --solver-max-time-step-in-days=1 the results are between this PR and master is identical.(all residuals are the same) For default timesteps there are slight changes due to convergence differences. Note that you will need to force small timesteps in order to match Eclipse on this case.

totto82:

totto82:
jenkins build this opm-tests=212 please

totto82:
jenkins build this opm-tests=212 please

totto82:
Can this be merged.

PR #2234, opened 2019-12-18
More MPI serialization support

akva2:
Downstream of https://github.com/OPM/opm-common/pull/1353

### DIFF
--- a/tests/test_ParallelRestart.cpp
+++ b/tests/test_ParallelRestart.cpp
@@ -637,6 +637,25 @@ BOOST_AUTO_TEST_CASE(Actdims)
 }
 
 
+BOOST_AUTO_TEST_CASE(Runspec)
+{
+#if HAVE_MPI
+    Opm::Runspec val1(Opm::Phases(true, true, true, false, true, false, true, false),
+                      Opm::Tabdims(1,2,3,4,5,6),
+                      Opm::EndpointScaling(std::bitset<4>(13)),
+                      Opm::Welldims(1,2,3,4),
+                      Opm::WellSegmentDims(1,2,3),
+                      Opm::UDQParams(true, 1, 2.0, 3.0, 4.0),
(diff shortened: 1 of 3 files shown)


bska:
Other than the fact that this looks to be following the pattern established earlier I can't really comment.  I would prefer if the test assertions of the form
```C++
BOOST_CHECK(condition);
```
instead were phrased as
```C++
BOOST_CHECK_MESSAGE(condition, "Contextual failure message");
```
but I suppose we could update those in a future PR.

tests/test_ParallelRestart.cpp
```diff
@@ -571,6 +571,17 @@ BOOST_AUTO_TEST_CASE(Tabdims)
 }
 
 
+BOOST_AUTO_TEST_CASE(EndpointScaling)
+{
+#if HAVE_MPI
+    Opm::EndpointScaling val1(std::bitset<4>(123456));
```
bska: This is an unexpected initial value for a bitset of size 4.  From reference material I know that it's perfectly well defined (and creates an all-ones bitset), but I still had to think twice about it.

akva2: right, i can make it a valid value.

akva2:
i really would prefer to do those afterwards, every change I have to make give me a lot of conflicts to resolve in the remaining patch set of ~ 130 patches..

tests/test_ParallelRestart.cpp
bska: > right, i can make it a valid value.

Cool.  You might even consider constructing the bitset from a string:
```C++
std::bitset<4>("1111")
```

bska:
> i really would prefer to do those afterwards

Very well.  Let's do that.

> every change I have to make give me a lot of conflicts to resolve in the remaining patch set of ~ 130 patches..

Right.  Let's get those in first and tidy up in a separate pass later.

PR #2235, opened 2019-12-18
Fix bug in the well efficiency factor

totto82:
The well efficiency factor should not be in the well control equation.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1806,7 +1806,7 @@ namespace Opm
             switch(current) {
             case Well::InjectorCMode::RATE:
             {
-                control_eq = getSegmentGTotal(0) * efficiencyFactor / scaling - controls.surface_rate;
+                control_eq = getSegmentGTotal(0) / scaling - controls.surface_rate;
                 break;
             }
 
@@ -1838,7 +1838,7 @@ namespace Opm
 
                 }
 
-                control_eq = coeff*getSegmentGTotal(0)*efficiencyFactor / scaling - controls.reservoir_rate;
+                control_eq = coeff*getSegmentGTotal(0) / scaling - controls.reservoir_rate;
                 break;
             }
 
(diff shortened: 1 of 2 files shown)


totto82:
jenkins build this please

atgeirr:
I assumed we had a test for this, but apparently not! Do you have a small testcase we can use (that shows this fix to be correct)?

totto82:
I suggest modifying the SPE9_CP case. I have made a PR OPM/opm-tests#214

atgeirr:
jenkins build this opm-tests=214

totto82:
Why is this Green! Shouldn't adding WEFAC need an update of the reference solution?

joakim-hove:
> Why is this Green!

Never satisfied are you?

atgeirr:
> Shouldn't adding WEFAC need an update of the reference solution?

I agree, this smells a little strange. Did you see changes locally from the case modification?

totto82:
Yes. There are significant changes locally based on WEFAC. WELTARG should not change anything.

totto82:
jenkins build this opm-tests=214 update_data please

akva2:
i posted the explanation in the opm-tests PR.

totto82:
jenkins build this opm-tests=215 please

akva2:
notice the failure for the restart test - that is probably not expected.

totto82:
jenkins build this opm-tests=215 please

totto82:
I have update OPM/opm-tests#215 as you suggested

totto82:
jenkins build this opm-tests=215 update_data please

totto82:
jenkins build this please opm-tests=215 opm-tests=218 please

totto82:
jenkins build this opm-tests=215 opm-tests=218 please

akva2:
that's not going to work :) you have to merge your PR manually.

totto82:
> that's not going to work :)

Jenkins is happy!

akva2:
now that is rather confusing. you cannot specify two pr's for the same repo...

totto82:
Maybe both are ignored

akva2:
haha yes i commented too soon so it saw my comment and ran with master of opm-tests. sigh.

atgeirr:
If it is green with both ignored, we can merge here now and add the case afterwards? Not perfect but workable?

akva2:
jenkins build this opm-tests=219 please

akva2:
let's end this saga.

PR #2236, opened 2019-12-18
at option to set the time step size for shutting problematic wells

totto82:
I think this will be useful for testing and debugging.


totto82:
jenkins build this please

PR #2241, opened 2019-12-18
fixing the non-defaulted influx coefficient for AquiferFetkovich

GitPaean:
to handle the situation when influx coefficient is not defaulted. 

Will check more tomorrow and some correction or improvement can be added.

It is not completely clear from the document. The indentation is not very convenient for this file.

### DIFF
--- a/opm/simulators/aquifers/AquiferFetkovich.hpp
+++ b/opm/simulators/aquifers/AquiferFetkovich.hpp
@@ -105,46 +105,52 @@ protected:
         for (size_t idx = 0; idx < Base::cell_idx_.size(); ++idx) {
             const int cell_index = Base::cartesian_to_compressed_.at(Base::cell_idx_[idx]);
             Base::cellToConnectionIdx_[cell_index] = idx;
-
-            const auto cellFacesRange = cell2Faces[cell_index];
-            for (auto cellFaceIter = cellFacesRange.begin(); cellFaceIter != cellFacesRange.end(); ++cellFaceIter) {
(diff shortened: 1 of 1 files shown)


opm/simulators/aquifers/AquiferFetkovich.hpp
```diff
@@ -105,41 +105,51 @@ namespace Opm
...
+                case 5:
+                    faceDirection = Opm::FaceDir::ZPlus;
+                    break;
+                default:
+                    OPM_THROW(Opm::NumericalIssue,
+                              "Initialization of Aquifer problem. Make sure faceTag is correctly defined");
+                }
+
+                if (faceDirection == connection.reservoir_face_dir.at(idx)) {
+                    Base::faceArea_connected_.at(idx) = Base::getFaceArea(faceCells, ugrid, faceIdx, idx, connection);
```
GitPaean: If understand correctly, the `faceArea_connected_` member can be removed.

GitPaean:
the Fetkovich Aquifer case is expected to fail.

GitPaean:
jenkins build this please

GitPaean:
> the Fetkovich Aquifer case is expected to fail.

The test in OPM-tests have the same size of the face area, so although it ignored the input values, it does not impact the result.

GitPaean:
jenkins build this please

GitPaean:
I think this is a simple fix for a situation we did not support, and it also has very limited scope of impact. 

We should get it in.

PR #2243, opened 2019-12-18
Pass a pointer to the EclipseGrid to CpGrid::processEclipseGrid.

blattms:
... instead of a reference This is needed as we only want to read the
full deck and construct the EclipseGrid only on the master process
with rank 0. Hence all other ranks will pass a nullptr to this
function. This will be possible with this move from a reference to a
pointer.

downstream of PR OPM/opm-grid#433


blattms:
thanks for the approval

PR #2249, opened 2019-12-20
Throw std::runtime_error() instead of std::abort() in test failure

joakim-hove:
+ `std::cerr`instead of `std::cout`for message


joakim-hove:
jenkins build this please

PR #2250, opened 2019-12-20
More MPI serialization support

akva2:
Downstream of https://github.com/OPM/opm-common/pull/1367


bska:
Green from upstream trigger. Merging as promised. Thanks.

PR #2251, opened 2019-12-20
fixed: if not tests failed, do not proceed with update data procedure

akva2:
Avoid non-sense issues in opm-tests.

Downstream of https://github.com/OPM/opm-common/pull/1368


atgeirr:
I guess we should test this by asking for a data update...!

atgeirr:

akva2:
already started from the upstream.

PR #2262, opened 2020-01-03
Avoid global ecl state in setting up additional modules

akva2:
Avoid using deck directly on non-root processes when setting up the solvent, polymer and foam modules.

Downstream of https://github.com/OPM/opm-models/pull/584

### DIFF
--- a/ebos/eclmpiserializer.hh
+++ b/ebos/eclmpiserializer.hh
@@ -1,0 +1,55 @@
+/*
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


totto82:
We have tried to keep as much as possible of the model extension code "hidden" from the standard code and eclproblem.hh is already a very long file. Could you move the (un)pack code to the ...modules.hh files in opm-models if it is not too much work?

akva2:
it depends on opm-simulators code so that is problematic.

akva2:
hopefully you like this one better.

totto82:
Thanks! Please merge when green.

bska:
This appears to [break](https://ci.opm-project.org/job/opm-simulators/215/consoleFull) the sequential/serial build.  How much work will it be to restore that configuration?

bska:
> This appears to break the sequential/serial build.

Just a quick follow-up.  Here are the (unique) diagnostic messages from a local build
```
${ROOT}/opm-simulators/ebos/eclproblem.hh:615:42: error: no matching function for call to ‘Opm::EclMpiSerializer::EclMpiSerializer(const Dune::CollectiveCommunication<Dune::No_Comm>&)’
${ROOT}/opm-simulators/ebos/eclproblem.hh:633:38: error: no matching function for call to ‘Opm::EclMpiSerializer::EclMpiSerializer(const Dune::CollectiveCommunication<Dune::No_Comm>&)’
```

akva2:
Should be fixed by https://github.com/OPM/opm-simulators/pull/2265

bska:
> Should be fixed by #2265

Cool.  Thanks a lot for the quick turn-around.  Merge #2265 when green.

PR #2268, opened 2020-01-08
Avoid using global ecl state setting up SummaryConfig on non-root processes

akva2:
Downstream of https://github.com/OPM/opm-common/pull/1393


akva2:
@bska please don't merge upstreams then leave the downstream hanging :)

PR #2269, opened 2020-01-08
fix for vrep and rein for more then one level of groups

totto82:
Avoid adding the group contribution from children twice. 
This fixes Norne 4B.

### DIFF
--- a/opm/simulators/wells/WellGroupHelpers.hpp
+++ b/opm/simulators/wells/WellGroupHelpers.hpp
@@ -408,31 +408,27 @@ namespace Opm {
     }
 
 
-    inline void updateVREPForGroups(const Group& group, const Schedule& schedule, const int reportStepIdx, const WellStateFullyImplicitBlackoil& wellStateNupcol, WellStateFullyImplicitBlackoil& wellState, double& resv) {
+    inline void updateVREPForGroups(const Group& group, const Schedule& schedule, const int reportStepIdx, const WellStateFullyImplicitBlackoil& wellStateNupcol, WellStateFullyImplicitBlackoil& wellState) {
         for (const std::string& groupName : group.groups()) {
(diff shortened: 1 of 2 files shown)


totto82:
jenkins build this please

totto82:
@atgeirr Review?

atgeirr:
> Review?

Will do, but not sure I can do it this afternoon.

opm/simulators/wells/WellGroupHelpers.hpp
```diff
@@ -408,31 +408,27 @@ namespace Opm {
...
 
 
-    inline void updateVREPForGroups(const Group& group, const Schedule& schedule, const int reportStepIdx, const WellStateFullyImplicitBlackoil& wellStateNupcol, WellStateFullyImplicitBlackoil& wellState, double& resv) {
+    inline void updateVREPForGroups(const Group& group, const Schedule& schedule, const int reportStepIdx, const WellStateFullyImplicitBlackoil& wellStateNupcol, WellStateFullyImplicitBlackoil& wellState) {
         for (const std::string& groupName : group.groups()) {
             const Group& groupTmp = schedule.getGroup(groupName, reportStepIdx);
-            double thisResv = 0.0;
-            updateVREPForGroups(groupTmp, schedule, reportStepIdx, wellStateNupcol, wellState, thisResv);
-            resv += thisResv;
+            updateVREPForGroups(groupTmp, schedule, reportStepIdx, wellStateNupcol, wellState);
```
atgeirr: With this, I first thought that the VREP/REIN numbers stored for each group in the well state would contain only the rates of the direct well children of the group, because I was not aware that sumWellPhaseRates() also does a recursive descent accumulating all subgroups' wells. While I think the code is correct now, perhaps we should rename the function to make it clear, for example `sumAllSubWellPhaseRates`, or preferably something better.

PR #2272, opened 2020-01-09
fixed: use std::make_tuple instead of initializer list

### LINKED ISSUE #2271: test_ParallelRestart fails for gcc 5.4.0
Locally I get the follow error with master
```
/home/thsa/workspace/opm/opm-simulators/tests/test_ParallelRestart.cpp:1353:57: error: converting to ‘std::tuple<double, double, double>’ from initializer list would use explicit constructor ‘constexpr std::tuple< <template-parameter-1-1> >::tuple(_UElements&& ...) [with _UElements = {double, double, double}; <template-parameter-2-2> = void; _Elements = {double, double, double}]’
```
I use gcc 5.4.0 with standard flags including -std=c++14

akva2:
the constructor is explicit with gcc5.

Closes #2271

### DIFF
--- a/tests/test_ParallelRestart.cpp
+++ b/tests/test_ParallelRestart.cpp
@@ -1241,7 +1241,9 @@ BOOST_AUTO_TEST_CASE(UniformXTabulatedTwoDFunction)
 #ifdef HAVE_MPI
     std::vector<double> xPos{1.0, 2.0};
     std::vector<double> yPos{3.0, 4.0};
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

akva2:
@bska @totto82 this should fix it.

bska:
> this should fix it.

Confirmed.  Thanks a lot!  I'll merge into master.

PR #2276, opened 2020-01-13
Fix REIN for multi level groups by passing the top phase

totto82:
Switch all down-stream wells with same phase as the violated group to GRUP control.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1954,9 +1954,9 @@ namespace Opm {
                 well_state.setCurrentProductionGroupControl(group.name(), newControl);
                 ss << "Switching control mode for group "<< group.name() << " to " << Group::ProductionCMode2String(newControl);
             }
-            // Pass a dummy phase for producer groups. The topPhase is only relevant for injector groups
-            const Phase topPhase = Phase::WATER;
-            wellGroupHelpers::setGroupControl(group, schedule(), topPhase, reportStepIdx, false, well_state, ss);
+            // Pass a dummy phase for producer groups. The topUpPhase is only relevant for injector groups
+            const Phase topUpPhase = Phase::WATER;
+            wellGroupHelpers::setGroupControl(group, schedule(), topUpPhase, reportStepIdx, false, well_state, ss);
             break;
         }
         default:
@@ -1994,8 +1994,8 @@ namespace Opm {
             well_state.setCurrentInjectionGroupControl(group.name(), newControl);
         }
         const auto& summaryState = ebosSimulator_.vanguard().summaryState();
-        const Phase& topPhase = group.injectionControls(summaryState).phase;
-        wellGroupHelpers::setGroupControl(group, schedule(), topPhase, reportStepIdx, /*isInjector*/true, well_state, ss);
+        const Phase& topUpPhase = group.injectionControls(summaryState).phase;
+        wellGroupHelpers::setGroupControl(group, schedule(), topUpPhase, reportStepIdx, /*isInjector*/true, well_state, ss);
 
         if (!ss.str().empty())
             deferred_logger.info(ss.str());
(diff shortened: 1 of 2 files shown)


opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1954,7 +1954,9 @@ namespace Opm {
                 well_state.setCurrentProductionGroupControl(group.name(), newControl);
                 ss << "Switching control mode for group "<< group.name() << " to " << Group::ProductionCMode2String(newControl);
             }
-            wellGroupHelpers::setGroupControl(group, schedule(), reportStepIdx, false, well_state, ss);
+            // Pass a dummy phase for producer groups. The topPhase is only relevant for injector groups
+            const Phase topPhase = Phase::WATER;
+            wellGroupHelpers::setGroupControl(group, schedule(), topPhase, reportStepIdx, false, well_state, ss);
             break;
```
totto82: Better to make a separate setInjectionGroupControl() and setProductionGroupControl() to avoid passing a dummy phase?

totto82:
jenkins build this please

totto82:
@atgeirr  review?

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1991,7 +1993,9 @@ namespace Opm {
             }
             well_state.setCurrentInjectionGroupControl(group.name(), newControl);
         }
-        wellGroupHelpers::setGroupControl(group, schedule(), reportStepIdx, /*isInjector*/true, well_state, ss);
+        const auto& summaryState = ebosSimulator_.vanguard().summaryState();
+        const Phase& topPhase = group.injectionControls(summaryState).phase;
```
atgeirr: I think the term "top phase" is inaccurate, is this not the "top up phase"? If so, the variable name should be `topUpPhase` here and in the function called.

opm/simulators/wells/WellGroupHelpers.hpp
```diff
@@ -29,15 +29,17 @@ namespace Opm {
...
 
         for (const std::string& groupName : group.groups()) {
             const Group& groupTmp = schedule.getGroup(groupName, reportStepIdx);
-            setGroupControl(groupTmp, schedule, reportStepIdx, injector, wellState, ss);
-            if (injector)
-                wellState.setCurrentInjectionGroupControl(groupName, Group::InjectionCMode::FLD);
-            else
+            setGroupControl(groupTmp, schedule, topPhase, reportStepIdx, injector, wellState, ss);
+            if (injector) {
+                if (groupTmp.injection_phase() == topPhase || wellState.currentInjectionGroupControl(groupName) == Group::InjectionCMode::NONE) // only switch sub groups with same face or NONE
```
atgeirr: Why do we switch NONE groups?

And: what happens if, in a chain of groups (A parent of B parent of C), if A switches to (say) oil rate control, B is still on gas control, but C is on oil control? Is that possible?

Also a small typo: face -> phase.

opm/simulators/wells/BlackoilWellModel_impl.hpp
totto82: topUpPhase is on its way. See PR OPM/opm-common#1410.

opm/simulators/wells/WellGroupHelpers.hpp
totto82: Hierarchy: That should be fine. Keep in mind these are injectors and injecting oil is maybe not the best example. Say group A switch to water injector control, B keeps injecting gas, then C should change to FLD since it may need to be constraint to honor group As target.  

NONE: If group A with gas rate control is parent of a group B without any control. Group B will default to water phase. But it should be counted in.

opm/simulators/wells/BlackoilWellModel_impl.hpp
totto82: I hoped to get this merged before starting to work on the topUpPhase.

atgeirr: Ok. Will you fix the typo first?

totto82:
jenkins build this please

PR #585, opened 2016-02-16
Account for variable change in 2.4.1

blattms:
Due to the upcomming release the name changes (`j`->`j_`, `nnz`->`nnz_`) already appear
in 2.4.1. This commit now checks for 2.4.1 and newer instead of 3.0 and newer.

PR #433, opened 2015-08-17
optimizing the formInterleavedSystem().

GitPaean:
It avoids the using of `formEllipticSystem()` and `vercatCollapseJacs()`, which take a significant amount of computing time during the non-linear solutions.

The performance tests are conducted with the single option `use_interleaved=true` and all the other options defaulted. 

Testing with SPE9 indicates more than 10% performance enhancement. Testing with Norne only shows about 7% performance enhancement. It should also be a little more than 10% according to the profiling analysis. (My computer is showing some performance instability, which I am looking for causes. Different running with same parameters can show different performance (can be more than 10% different).)

To run through with Norne with `use_interleaved=true`, PR OPM/opm-core#851 is required. 

This PR is done with @atgeirr together.

PR #289, opened 2015-01-29
Include derivatives of vappars

totto82:
The derivatives of the vappars are included in the Jacobian.
To avoid inf derivatives for vap<1, the oil saturation is restricted
from below by sqrt(epsilon).

With this PR, Norne run smoothly past the troublesome parts from 484 days.

PR #2277, opened 2020-01-13
Avoid using deck setting up Schedule on non-root processes

akva2:
Downstream of https://github.com/OPM/opm-common/pull/1407

See description in upstream PR. While the two last commits, marked with [TEMP] should not really make a difference, they do. With the commits, all tests pass, if you drop the two temp commits, the SPE9_CP_GROUP test fails. I cannot figure out why. Please help.

### DIFF
--- a/opm/simulators/utils/ParallelRestart.cpp
+++ b/opm/simulators/utils/ParallelRestart.cpp
@@ -449,7 +449,6 @@ HANDLE_AS_POD(DENSITYRecord)
 HANDLE_AS_POD(Eqldims)
 HANDLE_AS_POD(GuideRateConfig::GroupTarget);
 HANDLE_AS_POD(GuideRateConfig::WellTarget);
-HANDLE_AS_POD(JFunc)
 HANDLE_AS_POD(MLimits)
 HANDLE_AS_POD(PVTWRecord)
 HANDLE_AS_POD(PVCDORecord)
@@ -1913,6 +1912,17 @@ std::size_t packSize(const EclHysterConfig& data,
     return packSize(data.active(), comm) +
            packSize(data.pcHysteresisModel(), comm) +
            packSize(data.krHysteresisModel(), comm);
+}
+
+std::size_t packSize(const JFunc& data,
+                     Dune::MPIHelper::MPICommunicator comm)
+{
+    return packSize(data.flag(), comm) +
+           packSize(data.owSurfaceTension(), comm) +
+           packSize(data.goSurfaceTension(), comm) +
+           packSize(data.alphaFactor(), comm) +
+           packSize(data.betaFactor(), comm) +
+           packSize(data.direction(), comm);
 }
 
 ////// pack routines
@@ -3721,6 +3731,18 @@ void pack(const EclHysterConfig& data,
     pack(data.active(), buffer, position, comm);
     pack(data.pcHysteresisModel(), buffer, position, comm);
     pack(data.krHysteresisModel(), buffer, position, comm);
+}
+
+void pack(const JFunc& data,
+          std::vector<char>& buffer, int& position,
+          Dune::MPIHelper::MPICommunicator comm)
+{
+    pack(data.flag(), buffer, position, comm);
+    pack(data.owSurfaceTension(), buffer, position, comm);
+    pack(data.goSurfaceTension(), buffer, position, comm);
+    pack(data.alphaFactor(), buffer, position, comm);
+    pack(data.betaFactor(), buffer, position, comm);
+    pack(data.direction(), buffer, position, comm);
 }
 
 /// unpack routines
@@ -6321,6 +6343,25 @@ void unpack(EclHysterConfig& data,
     unpack(pcHysteresisModel, buffer, position, comm);
     unpack(krHysteresisModel, buffer, position, comm);
     data = EclHysterConfig(active, pcHysteresisModel, krHysteresisModel);
+}
+
+void unpack(JFunc& data,
+            std::vector<char>& buffer, int& position,
+            Dune::MPIHelper::MPICommunicator comm)
+{
+    JFunc::Flag flag;
+    double owSurfaceTension, goSurfaceTension;
+    double alphaFactor, betaFactor;
+    JFunc::Direction dir;
+
+    unpack(flag, buffer, position, comm);
+    unpack(owSurfaceTension, buffer, position, comm);
+    unpack(goSurfaceTension, buffer, position, comm);
+    unpack(alphaFactor, buffer, position, comm);
+    unpack(betaFactor, buffer, position, comm);
+    unpack(dir, buffer, position, comm);
+    data = JFunc(flag, owSurfaceTension, goSurfaceTension,
+                  alphaFactor, betaFactor, dir);
 }
 
 #define INSTANTIATE_PACK_VECTOR(T) \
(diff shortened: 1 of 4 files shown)


opm/simulators/utils/ParallelRestart.cpp
```diff
@@ -102,6 +102,156 @@
...
+    std::set<Type> unique;
+    for (const auto& w : state.data()) {
+        if (w) {
+            unique.insert(w);
+        }
+    }
+    std::vector<int> idxVec;
+    idxVec.reserve(state.data().size()+1);
+    for (const auto& w : state.data()) {
+        int distance = std::distance(unique.begin(), unique.find(w));
```
blattms: This looks a bit fishy. For `OrderedMap` `T` will be  a `shared_ptr`. If state, for whatever reason, stores a `shared_ptr` which is null, then unique will not have entry for it, hence `distance` will be `unique.size()`.

I guess on the receiving side in `reconstructDynMap` will convert whatever is stored at unique[unique.size()] to a pointer.
In most cases this will not be a `nullptr`

Note that changing the order of packing and unpacking, might have influence in this case. Hopefully, this is the bug your are searching for. HTH

akva2: while indeed this could be an issue, it's not The issue and does not trigger in this case. I repeat it works if i deserialize directly into the variables; the order thing is just so i can do it *after* instancing the Schedule object using the = operator. but nice catch.

joakim-hove:
I'll try to flex my gdb+opm-commen muscles and see if I get some divine inspiration.

akva2:
we all know who's been messing around there, and it ain't me.

akva2:
also i just rebased on master and all is green at my end so i'm fairly certain you screwed up somehow.

finally it's not like i haven't gdb'd this from all angles. neither gdb nor valgrind complains and upon gdb inspection of the variables nothing stands out as broken. example gdb script;

```
set width 80
set height unlimited
set logging overwrite on

# Deserialized GroupMap
break ParallelRestart.cpp:6016
commands
set logging file groups_first.log
set logging on
p groups
end

# Serialized GroupMap
break ParallelRestart.cpp:3530
commands
p data.getGroups()
cont
end

# Serialized Group
break ParallelRestart.cpp:3153
commands
set logging file groups_first.log
set logging on
p data
cont
end

# Deserialized Group
break ParallelRestart.cpp:5436
commands
set logging file groups_second.log
set logging on
p data
cont
end

run SPE9_CP_GROUP
```
line numbers are likely in need of updating. 

execute using
```
mpirun -np 2 konsole -e gdb -command above-text-file /path/to/flow
```

you can then diff groups_first.log and groups_second.log

joakim-hove:
> also i just rebased on master and all is green at my end so i'm fairly certain you screwed up somehow.

I agree that is still a probable explanation; but this time around I have really tried carefully to not f... up. I have cloned new repositories, fetched your branches from opm-common and opm-simulators and then built everything in mode "RelWithDebugInfo" and with MPI on. I still get the same crashes.

I am on Debian10 with gcc: 8.3.0

```txt
opm-common: 657a87525
opm-material: 3d2a2fe
opm-grid: 3d7d0618
opm-models: 62499896d
opm-simulators: 583633a8
```

blattms:
I did not see crashes last night with:

--- calling git for opm-common ---
657a87525 (HEAD -> noecl_flush_schedule, akva2/noecl_flush_schedule) [TEMP] add mutable accessors to wells and groups members in schedule
--- opm-common done ---
--- calling git for opm-material ---
c67d84b (HEAD -> master, upstream/master, upstream/HEAD) Merge pull request #381 from totto82/trinemykk-smartwater
--- opm-material done ---
--- calling git for opm-grid ---
260a49e (HEAD -> master, upstream/master) Merge pull request #433 from blattms/construct-cpgrid-from-eclgrid-pointer
--- opm-grid done ---
--- calling git for opm-models ---
62499896d (HEAD -> master, upstream/master, upstream/HEAD) Merge pull request #585 from totto82/trinemykk-smartwater
--- opm-models done ---
--- calling git for opm-simulators ---
2b43c414f (HEAD -> noecl_flush_schedule, akva/noecl_flush_schedule) [TEMP] unpack using mutable accessors
--- opm-simulators done ---

blattms:
Updated, but still no crash when running SPE9_CP_GROUP, I am on Debian 10.2 with same compiler.
Versions are now

module opm-common: 
657a87525 (HEAD -> noecl_flush_schedule, akva2/noecl_flush_schedule) [TEMP] add mutable accessors to wells and groups members in schedule
module opm-grid: 
260a49e (HEAD -> master, upstream/master) Merge pull request #433 from blattms/construct-cpgrid-from-eclgrid-pointer
module opm-material: 
c67d84b (HEAD -> master, upstream/master, upstream/HEAD) Merge pull request #381 from totto82/trinemykk-smartwater
module opm-simulators: 
583633a87 (HEAD -> noecl_flush_schedule, akva/noecl_flush_schedule) [TEMP] unpack using mutable accessors
module opm-models: 
62499896d (HEAD -> master, upstream/master, upstream/HEAD) Merge pull request #585 from totto82/trinemykk-smartwater
module opm-upscaling: 
4e09ef6 (HEAD -> master, upstream/master, upstream/HEAD) Merge pull request #289 from blattms/construct-cpgrid-from-eclgrid-pointer

blattms:
@joakim-hove You do not seem to use master for opm-material.

blattms:
... and I do not use master of opm-grid.

blattms:
... opm-material and this one breaks it and I see the crash. It must be OPM/opm-material#380

blattms:
I can verify the crash on my machine now even in a serial run. There is no crash if I use master of opm-common and opm-simulators.

gdb says:
```
   at /home/mblatt/src/dune/opm/opm-material/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp:130
130	        if (eclState.fieldProps().has<int>("SATNUM")) {
(gdb) p eclState.fieldProps()
$3 = (const Opm::FieldPropsManager &) @0x55555a93d7d8: {
  fp = std::shared_ptr<Opm::FieldProps> (empty) = {get() = 0x0}}
```

blattms:
It seems like the noecl_flush_schedule branch of opm-common is not rebased onto master.

blattms:
Indeed. Crash is gone after a local rebase of the branches in opm-common and opm-simulators.

blattms:
This is the right PR:

Not sure whether it is this PR or a previous one(maybe be me). But using the mpiwrappers from OpenMPI and 2 processes `LD_PRELOAD=/usr/lib/x86_64-linux-gnu/valgrind/libmpiwrap-amd64-linux.so mpirun -np 2 valgrind ../../opm-simulators/opm-parallel-debug/bin/flow SPE9_CP_GROUP`,  with the stuff marked with `[TEMP]` reverted, shows some potential problems. See
[valgrind.log](https://github.com/OPM/opm-simulators/files/4070811/valgrind.log).

These [wrappers](https://www.valgrind.org/docs/manual/mc-manual.html#mc-manual.mpiwrap) are quite handy. (bummer, that I always forget how to activate them)

HTH

akva2:
i adressed the errors. now i only get errors in the grid setup (false positives?) and first timestep related to residual reduction in the ebos solver.

mystery deepens more though; i can run through valgrind (with mpi wrappers) and test passes. 
if i run a pure release binary, it stalls at report step 10. if i run a release binary with debug symbols, it cuts timestep to pieces at report step 10, then exits. if a run a debug binary, it smashes the stack on startup. hurrah \o/

akva2:
sigh. it's official. i hate myself. 'let's use a set of shared ptrs where we need stable indices'. FFS.

bska:
Merging now.  Thanks a lot.

PR #2278, opened 2020-01-13
add custom target 'opm-simulators_prepare'

akva2:
this acts as a syncronization point for the super-build.
the ebos objects do not depend directly on opm-common, and thus
they are can be built before the generated headers, which are
required, are in place. we can now use this to ensure the
generated headers are in place before proceeding.


bska:
@akva2 : Do you still want to proceed with this PR?

akva2:
no, i have fixed it elsewhere.

PR #2281, opened 2020-01-15
changed: use CpGrid centroids

akva2:
Is the use of the EclipseGrid centroids really crucial here?

I see it causes small failures, but.. I don't want to do the complicated thing unless it actually is important.

Context: No eclipsegrid on non-root processes in parallel.

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -160,11 +160,7 @@ public:
             const auto& elem = *elemIt;
             unsigned elemIdx = elemMapper.index(elem);
 
-            // compute the axis specific "centroids" used for the transmissibilities. for
-            // consistency with the flow simulator, we use the element centers as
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

atgeirr:
Yes this is important, to get the same transmissibilities as Eclipse. Perhaps there are other workarounds?

akva2:
not workarounds, but i'll write the necessary mpi code then. thanks.

atgeirr:
One idea could be to modify the centroids embedded in the grid during grid construction (if desired, which means yes for Flow and no for upscaling codes), so that one can always use the grid centroids directly. Same with volumes. While that would be a significant change, I think it would simplify things in the long run. Would that help you in this context?

akva2:
it would certainly help, but to avoid a long deour, for now i'll just write the ~ 30 lines of code to send from master process.

PR #2282, opened 2020-01-15
changed: avoid usage of eclipseGrid on non-root processes setting up transmissibilities

akva2:
obtain data from root process

@blattms I am sure you do not like the bare MPI calls, please advice which construct to use for non-collective communication.

### DIFF
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -151,7 +151,7 @@ public:
             if (grid_->size(0))
             {
                 globalTrans_.reset(new EclTransmissibility<TypeTag>(*this));
-                globalTrans_->update();
+                globalTrans_->update(false);
             }
 
             Dune::EdgeWeightMethod edgeWeightsMethod = this->edgeWeightsMethod();
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -834,7 +834,7 @@ public:
             eclState.applyModifierDeck(miniDeck);
 
             // re-compute all quantities which may possibly be affected.
-            transmissibilities_.update();
+            transmissibilities_.update(true);
             referencePorosity_[1] = referencePorosity_[0];
             updateReferencePorosity_();
             updatePffDofData_();
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -118,15 +118,16 @@ public:
      * either but at least it seems to be much better.
      */
     void finishInit()
-    { update(); }
+    { update(true); }
 
 
     /*!
      * \brief Compute all transmissibilities
      *
+     * \param global If true, update is called on all processes
      * Also, this updates the "thermal half transmissibilities" if energy is enabled.
      */
-    void update()
+    void update(bool global)
     {
         const auto& gridView = vanguard_.gridView();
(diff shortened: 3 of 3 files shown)


akva2:
jenkins build this please

blattms:
Concerning the MPI calls: Well if we could be sure that this code is
just run with CpGrid we could use `CpGrid::scatterData` for this.
the problem is that we seem to support other grids (in downstream modules,
@drrobertk?) and they only allow distributing data during loadbalance.

Hence the clean way would be to distribute it with loadbalance. But we will
need to distribute other stuff, too. So maybe we should leave this for another
day and open an issue for it?

akva2:
i kinda want to do it properly immediately cause there are a lot of places i have to write code equivalent to this.

blattms:
The proper way to do this is in the grid vanguards. In [eclalugridvanguard.hh#L147](blob/ecfbb0e4919a52b93c97d8c8dfd9164d99bcfada/ebos/eclalugridvanguard.hh#L147) we already used a datahandle for distributing the cartesian index when loadbalancing.

I guess the centroids are only for CpGrid? Then we need to create a Datahandle
```
struct MyHandle{
  using DataType = std::array<double, 3>;
  bool fixedsize(int /* dim */, int /* codim */) const {return true;}
  bool contains (int /* dim /, int codim) const { return codim==0;}
  template<class Entity>
  size_t size (const Entity e) const ( return 1;}
  template<class MessageBufferImp , class EntityType > 
  void gather (MessageBufferImp &buff, const EntityType &e) const
  {
      buffer.write(eclGrid.getCellCenter(elemMapper.index(e)));
  }
  template<class MessageBufferImp , class EntityType > 
  void scatter (MessageBufferImp &buff, const EntityType &e, size_t) const
  {
     buffer.read(centroids[elemMapper.index(e)]);
  }
};
```
and pass that to loadbalance in eclcpgridvanguard. Later we probably need to extend the handle to communicate more.

blattms:
[Datahandle interface](https://www.dune-project.org/doxygen/master/classDune_1_1CommDataHandleIF.html)

blattms:
Not sure whether it is this PR or previous (maybe be me). But using the mpiwrapper from OpenMPI
and 2 processes `LD_PRELOAD=/usr/lib/x86_64-linux-gnu/valgrind/libmpiwrap-amd64-linux.so mpirun -np 2 valgrind ../../opm-simulators/opm-parallel-debug/bin/flow SPE9_CP_GROUP`,  with the stuff marked with `[TEMP]` reverted, shows some problems. See
[valgrind.log](https://github.com/OPM/opm-simulators/files/4070811/valgrind.log).

These [wrappers](https://www.valgrind.org/docs/manual/mc-manual.html#mc-manual.mpiwrap) are quite handy. (bummer, that I always forget how to activate them)

HTH

blattms:
They could be false positives of course.

akva2:
at least that gives me some crumbs to follow. thanks!

blattms:
Arrgh. Wrong PR. Sorry.

blattms:
somehow this is quite different from my approach at akva2/opm-simulators#11. I am a bit surprised that both should accomplish the same.

blattms:

blattms:
jenkins run norne parallel please

blattms:

blattms:
can this be closed?

PR #2283, opened 2020-01-16
make group constrain check conditional

totto82:
As a general rule we don't want to check the current group control. A naive implementation of this did however not work since well controls may change during simulations due to input in the Schedule or closing of wells etc. This PR tries to circumvent this by always checking the first two iterations. An alternative solution is to add an epsilon to avoid switching as has been used in testing branches for groups [#2273] for a long time. The solution suggested in this PR seems more robust.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -306,7 +306,7 @@ namespace Opm {
             // xw to update Well State
             void recoverWellSolutionAndUpdateWellState(const BVector& x);
 
-            void updateWellControls(Opm::DeferredLogger& deferred_logger, const bool checkGroupControl);
+            void updateWellControls(Opm::DeferredLogger& deferred_logger, const bool checkGroupControl, const bool checkCurrentGroupControl);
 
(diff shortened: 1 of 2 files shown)


totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
I have double-checked the failing test. The results are fine.

totto82:
@atgeirr review?

totto82:
Can we merge this?

atgeirr:
I think this can be merged, but Jenkins does not agree...

Also, I am a bit hesitant about the update being decided by report steps and not timesteps in general, is this really thought through? I would think it more appropriate to check every timestep.

totto82:
> is this really thought through

I thought so too, but checking every timestep led to issues for several of the norne models. We still need to iterate on this, but I think this is a move in the right direction. Merging this makes testing easier and hopefully we can improve on this later. It has already gone 3 weeks sine it appeared. I have verified the test failure.  update_data should do the trick.

atgeirr:

atgeirr:
jenkins build this opm-tests=236 please

PR #2285, opened 2020-01-17
changed: avoid near-duplicate code

akva2:
I see no reason why this code is duplicated like this.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -2932,80 +2932,44 @@ private:
         return eclState.fieldProps().get_global_int(prop);
     }
 
-    void updatePvtnum_()
+    template<class T>
+    void updateNum(const std::string& name, std::vector<T>& numbers)
     {
         const auto& simulator = this->simulator();
         const auto& eclState = simulator.vanguard().eclState();
 
-        if (!has_int_prop(eclState, "PVTNUM"))
+        if (!has_int_prop(eclState, name))
             return;
 
-        const auto& pvtnumData = get_int_prop(eclState, "PVTNUM");
+        const auto& numData = get_int_prop(eclState, name);
         const auto& vanguard = simulator.vanguard();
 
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -2932,80 +2932,44 @@ private:
         return eclState.fieldProps().get_global_int(prop);
     }
 
-    void updatePvtnum_()
+    template<class T>
```
bska: Why `template`?  Since this internally uses `has_int_prop()` and `get_int_prop()` the function itself isn't usable for anything other than `T = int`.

akva2: because pvtnum is std::vector<int> and the others are std::vector<unsigned short>

bska: >> Why `template`?
>
> because pvtnum is `std::vector<int>` and the others are `std::vector<unsigned short>`

Okay, but then this probably needs some additional work before it's ready.  In particular since `numData` will unconditionally be `vector<int>`, we should probably make the type conversion explicit when assigning to the result vector (`numbers`).  Finally, it's **possible** that the `*NUM` vectors contain zero values in inactive cells in which case   unconditionally subtracting one will be troublesome for the unsigned representations.

Of course these problems already exist in the functions you're replacing, but I think this would be a good opportunity to correct those oversights.

akva2: i can add the cast but i don't understand what you intend to do about the subtraction. seeing that these cells are inactive, i don't think we would ever read those values later?

bska: > i don't understand what you intend to do about the subtraction.

```C++
numbers[elemIdx] = static_cast<T>(std::max(numData[cartElemIdx], 1) - 1);
```

akva2:
jenkins build this please

bska:
Merging now as the build check was successful.

PR #2287, opened 2020-01-17
avoid instancing a new Runspec from deck to obtain active phases

akva2:
this is already available in the eclipseState.
in particular, this is preparing for the case where only the
root-process has a deck instance.


akva2:
jenkins build this please

bska:
> I'll merge when the build test succeeds.

Build test came back green.  I'll merge into master now.  Thanks a lot.

PR #2290, opened 2020-01-20
Avoid deck usage on non-root processes setting up FluidSystem

akva2:
Downstream of https://github.com/OPM/opm-material/pull/383


bska:
I'll merge into master.  Thanks a lot.

akva2:
(We did have the shared_ptr support before, I just reduced the boilerplating).

PR #2292, opened 2020-01-20
Local actnum prep

joakim-hove:
Extracted two trivial commits from: https://github.com/OPM/opm-simulators/pull/2289

The change to use `std::vector<int>` is to allow for simpler interoperability with the data which comes from opm-common.

The content in the secomd commit is removed because we no longer use properties of global size - I.e it has become irrelevant/wrong.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -3199,11 +3199,10 @@ private:
     EclThresholdPressure<TypeTag> thresholdPressures_;
 
     std::vector<int> pvtnum_;
-    std::vector<unsigned short> satnum_;
-    std::vector<unsigned short> miscnum_;
-    std::vector<unsigned short> plmixnum_;
-
-    std::vector<unsigned short> rockTableIdx_;
+    std::vector<int> satnum_;
+    std::vector<int> miscnum_;
+    std::vector<int> plmixnum_;
+    std::vector<int> rockTableIdx_;
     std::vector<RockParams> rockParams_;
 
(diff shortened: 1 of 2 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

blattms:
Would mind telling us the reasoning for these changes? (Could not find anything in the commit messages...).

blattms:
Stubborn as I am:
- Why the move to int? What was wrong before?
- You have removed an assertion. If the assertion was wrong, why not correct it?

joakim-hove:
> Why the move to int? What was wrong before?

The source data coming from opm-common for the integer vectors like `SATNUM` and  `PVTNUM` is stored in `std::vector<int>` - so with the change to `int` it would be possible to do:
```C++
this->satnumData_ = fp.get<int>("SATNUM");
```
Now based on review comments I have understood that such an assignment is too simple and the assignments must be done on a element by element basis, so then the conversion `int` &rightarrow; `unsigned short/char` can be done implicitly - as is done currently. I would still argue that using `int` consistently is nicer - but it is not important.

> You have removed an assertion. If the assertion was wrong, why not correct it?

Well, now the data to formulate the correct assertion all lies with the `FieldProps` implementation, so that will essentially be a test of that - I do not see that as belonging here. The old assert was based on `sizeof(property) == nx*ny*nz` - which is obviously not correct any longer.

bska:
I'm closing this now since the patch no longer applies.

PR #2293, opened 2020-01-21
More serialization support

akva2:
Downstream of https://github.com/OPM/opm-common/pull/1422

These are the final bits to enable serialization of EclipseState.

As a FYI and motivation; while it's full of hacks, I can now run completely without parsing on non-root processes locally.


joakim-hove:
> I can now run completely without parsing on non-root processes locally.

image

PR #2294, opened 2020-01-22
changed: add helper template to serialize and broadcast static types

akva2:
4 uses, simpler syntax in the problem code. i think this justifies a helper template.

### DIFF
--- a/ebos/eclmpiserializer.hh
+++ b/ebos/eclmpiserializer.hh
@@ -27,7 +27,7 @@ namespace Opm {
 
 class EclMpiSerializer {
 public:
-    EclMpiSerializer(Dune::MPIHelper::MPICommunicator comm) :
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this please

akva2:
okay to merge?

ebos/eclmpiserializer.hh
```diff
@@ -27,7 +27,7 @@ namespace Opm {
 
 class EclMpiSerializer {
 public:
-    EclMpiSerializer(Dune::MPIHelper::MPICommunicator comm) :
+    EclMpiSerializer(Dune::CollectiveCommunication<Dune::MPIHelper::MPICommunicator> comm) :
```
bska: Is there a reason that this constructor is not marked `explicit`?

akva2: no, it should be. i'll include that in upcoming PR that depend on this, so i don't have to rebase this one.

PR #2296, opened 2020-01-22
Avoid deck usage on non-root processes checking for keywords

akva2:
Use EclipseState internalized  variables instead.

Downstream of https://github.com/OPM/opm-common/pull/1424

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -2466,12 +2466,12 @@ private:
     {
         const auto& simulator = this->simulator();
         const auto& vanguard = simulator.vanguard();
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

joakim-hove:
For the gravity and equil flags I think that `bool` flags in `ElipseState::InitConfig` would be a better solution. When it comes to the `SWATINIT` we are both making changes at the same time, so I guess there is no point in being to smart before the dust has settled.

akva2:
there is already an hasEquil in InitConfig, so I'll use that. I'll add the gravity, and drop the swatinit one for now.

PR #2298, opened 2020-01-23
Allow for multiple group injection phases

totto82:
TODO
- [x] fix 2p
- [x] fix mpi
- [x] test on norne 4C

### DIFF
--- a/opm/simulators/wells/WellGroupHelpers.hpp
+++ b/opm/simulators/wells/WellGroupHelpers.hpp
@@ -469,6 +469,7 @@ namespace Opm {
         }
         wellState.setCurrentInjectionVREPRates(group.name(), resv);
     }
+
     inline void updateReservoirRatesInjectionGroups(const Group& group, const Schedule& schedule, const int reportStepIdx, const WellStateFullyImplicitBlackoil& wellStateNupcol, WellStateFullyImplicitBlackoil& wellState) {
         for (const std::string& groupName : group.groups()) {
             const Group& groupTmp = schedule.getGroup(groupName, reportStepIdx);
--- a/opm/simulators/wells/WellGroupHelpers.hpp
+++ b/opm/simulators/wells/WellGroupHelpers.hpp
@@ -469,6 +469,18 @@ namespace Opm {
         }
         wellState.setCurrentInjectionVREPRates(group.name(), resv);
     }
+    inline void updateReservoirRatesInjectionGroups(const Group& group, const Schedule& schedule, const int reportStepIdx, const WellStateFullyImplicitBlackoil& wellStateNupcol, WellStateFullyImplicitBlackoil& wellState) {
+        for (const std::string& groupName : group.groups()) {
+            const Group& groupTmp = schedule.getGroup(groupName, reportStepIdx);
+            updateReservoirRatesInjectionGroups(groupTmp, schedule, reportStepIdx, wellStateNupcol, wellState);
+        }
+        const int np = wellState.numPhases();
+        std::vector<double> resv(np, 0.0);
+        for (int phase = 0; phase < np; ++phase) {
+            resv[phase] = sumWellPhaseRates(wellStateNupcol.wellReservoirRates(), group, schedule, wellState, reportStepIdx, phase, /*isInjector*/ true);
+        }
+        wellState.setCurrentInjectionGroupReservoirRates(group.name(), resv);
+    }
 
     inline void updateREINForGroups(const Group& group, const Schedule& schedule, const int reportStepIdx, const PhaseUsage& pu, const SummaryState& st, const WellStateFullyImplicitBlackoil& wellStateNupcol, WellStateFullyImplicitBlackoil& wellState) {
         const int np = wellState.numPhases();
--- a/opm/simulators/wells/WellGroupHelpers.hpp
+++ b/opm/simulators/wells/WellGroupHelpers.hpp
@@ -262,14 +262,14 @@ namespace Opm {
                 for (Phase phase : all) {
                     const Group::InjectionCMode& currentGroupControl = wellState.currentInjectionGroupControl(phase, groupName);
                     int phasePos;
-                    if (phase == Phase::GAS)
-                        phasePos = pu.phase_pos[ pu.phase_pos[BlackoilPhases::Vapour] ];
-                    else if (phase == Phase::OIL)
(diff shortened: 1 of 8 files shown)


joakim-hove:
The upstream PR https://github.com/OPM/opm-common/pull/1410 is ready to be merged

joakim-hove:
Ping pong! I am ready to merge the upstream PR in opm-common, but this one is a bit over my pay-grade in opm-simulators. Review please!

totto82:
From my point of view this can be merged.

totto82:
@atgeirr review?

opm/simulators/wells/WellGroupHelpers.hpp
```diff
@@ -523,12 +570,22 @@ namespace Opm {
...
         const Group& groupParent = schedule.getGroup(group.parent(), reportStepIdx);
+        int phasePos;
+        if (injectionPhase == Phase::GAS && pu.phase_used[BlackoilPhases::Vapour] )
+            phasePos = pu.phase_pos[ pu.phase_pos[BlackoilPhases::Vapour] ];
+        else if (injectionPhase == Phase::OIL && pu.phase_used[BlackoilPhases::Liquid])
+            phasePos = pu.phase_pos[ pu.phase_pos[BlackoilPhases::Liquid] ];
+        else if (injectionPhase == Phase::WATER && pu.phase_used[BlackoilPhases::Aqua] )
+            phasePos = pu.phase_pos[ pu.phase_pos[BlackoilPhases::Aqua] ];
+        else
+            throw("this should not happen");
```
atgeirr: I prefer assert() in general for such things, but it's ok. But what should actually happen if we specified a gas rate for a well in an oil/water case? Do we terminate when building the Schedule object?

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
```diff
@@ -337,14 +337,14 @@ namespace Opm
...
+            current_injection_group_controls_[std::make_pair(phase, groupName)] = groupControl;
         }
-        const Group::InjectionCMode& currentInjectionGroupControl(const std::string& groupName) const {
-            auto it = current_injection_group_controls_.find(groupName);
+        const Group::InjectionCMode& currentInjectionGroupControl(const Opm::Phase& phase, const std::string& groupName) const {
+            auto it = current_injection_group_controls_.find(std::make_pair(phase, groupName));
 
             if (it == current_injection_group_controls_.end())
-                OPM_THROW(std::logic_error, "Could not find any control for injection group " << groupName);
+                OPM_THROW(std::logic_error, "Could not find any control for " << phase << " injection group " << groupName);
```
atgeirr: The `phase` would be converted to an integer and not make much sense, we should get the corresponding string instead, also add a few words to make the message "Could not find any control for phase GAS for injection group INJ".

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
```diff
@@ -374,6 +374,18 @@ namespace Opm
             return it->second;
         }
 
+        void setCurrentInjectionGroupReservoirRates(const std::string& groupName, const std::vector<double>& target ) {
+            injection_group_reservoir_rates[groupName] = target;
+        }
+        const std::vector<double>& currentInjectionGroupReservoirRates(const std::string& groupName) const {
```
atgeirr: Please add blanks between the functions, especially when there are blanks inside the functions.

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
```diff
@@ -337,14 +337,14 @@ namespace Opm
         }
 
         /// One current control per group.
-        void setCurrentInjectionGroupControl(const std::string& groupName, const Group::InjectionCMode& groupControl ) {
-            current_injection_group_controls_[groupName] = groupControl;
+        void setCurrentInjectionGroupControl(const Opm::Phase& phase, const std::string& groupName, const Group::InjectionCMode& groupControl ) {
```
atgeirr: I would think it more natural to have `groupName` first, then `phase`, since the original call has only `groupName`. Is there a specific reason for preferring the current order?

totto82: It is the logic used in opm-common. I think we should keep the same order as in opm-common

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
totto82: Doesn't 
`std::ostream& operator<<( std::ostream&, const Phase& );` in Runspec.hpp
convert it to string? 

at least std::cout << phase  << std::endl; works perfectly.

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
totto82: done

opm/simulators/wells/WellGroupHelpers.hpp
```diff
@@ -434,6 +469,18 @@ namespace Opm {
         }
         wellState.setCurrentInjectionVREPRates(group.name(), resv);
     }
+    inline void updateReservoirRatesInjectionGroups(const Group& group, const Schedule& schedule, const int reportStepIdx, const WellStateFullyImplicitBlackoil& wellStateNupcol, WellStateFullyImplicitBlackoil& wellState) {
```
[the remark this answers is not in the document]
totto82: done

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
atgeirr: You are right, I did not know of that operator.

atgeirr:
This can be merged when the upstream is done.

joakim-hove:
@totto82 : I tried rebasing this and building it locally, but then I got two test failures:
```
35 - compareECLFiles_flow+SPE9_CP_GROUP (Failed)
48 - compareECLFiles_flow+UDQ_WCONPROD (Failed)
```

When running these tests in the console I get the error message:
```
ERROR: Uncaught std::exception when running tasklet: map::at. Trying to continue.
```
which to me indicates that there is some unchecked access to a phase somewhere. Naive checks on my part did not reveal anything.

PR #2301, opened 2020-01-24
Chase Runspec Member Update

bska:
The Runspec class gained a new data member of type SatFuncControls.
Update the serialization code accordingly.

Downstream from OPM/opm-common#1431.


bska:
jenkins build this opm-common=1431 please

PR #2303, opened 2020-01-27
Serialize rockconfig

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/1433


joakim-hove:
Mergeable?

opm/simulators/utils/ParallelRestart.hpp
```diff
@@ -76,7 +76,7 @@ namespace Action {
 
 class Aqudims;
 class BCConfig;
-class BCConfig::BCFace;
+struct BCConfig::BCFace;
```
akva2: this is useless. 
- you are prototyping an existing structure (ie already included)
- you can never prototype a member of another class like this in any case.

akva2: (I already remove this in https://github.com/OPM/opm-simulators/pull/2305 so my preferred fix is for you not to touch it, so i avoid the rebase)

joakim-hove: OK - done.

PR #2305, opened 2020-01-28
Parallel thermal simulations

akva2:
- some janitoring
- some fixes after brine addition
- avoid deck usage setting up thermal law manager
- add a thermal parallel regression test

I am somewhat worried about the coarse tolerances needed for the test to pass, but in any case this is an improvement.

Downstream of https://github.com/OPM/opm-material/pull/390


joakim-hove:
A complete side note: When working with this thermal code it appears to me to be a significant mismatch between the overall complexity introduced by this code and the amount of use.

bska:
I'm merging this now to keep the APIs in line with the upstream changes in OPM/opm-material#390.

PR #2309, opened 2020-01-29
fixed: run RelpermDiagnostics on the global grid in parallel runs

akva2:
Before it only ran on the part of the grid owned by process 0.

### DIFF
--- a/opm/simulators/flow/FlowMainEbos.hpp
+++ b/opm/simulators/flow/FlowMainEbos.hpp
@@ -449,7 +449,11 @@ namespace Opm
             // Run relperm diagnostics if we have more than one phase.
             if (FluidSystem::numActivePhases() > 1) {
                 RelpermDiagnostics diagnostic;
+                if (mpi_size_ > 1)
+                    this->grid().switchToGlobalView();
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

opm/simulators/flow/FlowMainEbos.hpp
```diff
@@ -449,7 +449,11 @@ namespace Opm
             // Run relperm diagnostics if we have more than one phase.
             if (FluidSystem::numActivePhases() > 1) {
                 RelpermDiagnostics diagnostic;
+                if (mpi_size_ > 1)
```
atgeirr: For ranks > 0, is it allowed to call switchToGlobalView()? I guess if the diagnosis only runs on rank 0 there might be no risk, but I do not see any "only rank 0" (or output_cout) code here?

akva2: see 5 lines further up..

atgeirr: Thanks (sometimes we are blind).

PR #2310, opened 2020-01-29
Preparatory changes for single-process parsing

akva2:
Avoid some deck usage, avoid some dereferencing.

The deck-access avoid is done simple right now, ideally these things would be internalized in EclipseState, but I don't want to be held up by them right now. Furthermore,  since neither tracers nor aquifers currently work in parallel, it's not that big of a deal.

### DIFF
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -258,13 +258,14 @@ public:
 protected:
     void createGrids_()
     {
-        const auto& porv = this->eclState().fieldProps().porv(true);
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

blattms:
Looks good.

blattms:
All these synchronization points will make us (or at least me) really dizzy in the future. I guess they should not stay for long.

akva2:
i agree that it's nasty. they will go away once the need for deck access is gone by the data being internalized in the eclipse state.

PR #2312, opened 2020-01-30
remove minpvFillNtg 

totto82:
The minpvFillNtg code depends on the global Cartesian properties and either needs to be adapted to the new workflow without global Cartesian properties or removed. Since minpv mode OPMFIL is not very much used and if used the ntg averaging is probably not very important for the user, lets removed it.

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -140,8 +140,7 @@ public:
 #endif
 
         // get the ntg values, the ntg values are modified for the cells merged with minpv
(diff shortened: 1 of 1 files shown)


alfbr:
I belive we ended up never using OPMFIL. Removal supported.

totto82:
jenkins build this please

akva2:
there is one more usage of this in eclproblem.hh plus maybe we should kill the enum and stuff in EclipseGrid?

PR #2313, opened 2020-01-30
Lab bc

totto82:
Fix and ads a test for bc


totto82:
jenkins build this opm-tests=232 please

PR #2314, opened 2020-01-30
Add Parallel Serialization Support for CurrentControl

bska:
This commit adds support for serializing `Opm::data::CurrentControl` between parallel processes.

This is downstream of, and must be merged in concert with, PR OPM/opm-common#1446.


bska:
I'm marking this and the upstream PR OPM/opm-common#1446 as "ready for review" since I don't have anything else to add here at this time.

PR #2316, opened 2020-01-31
fixed: add explicit template instantation for blocksize 1

akva2:
fixes linking of flow_onephase with bda/cuda enabled.

no reason to run tests as this code is not used on jenkins.


blattms:
jenkins build this please

PR #2318, opened 2020-02-05
fixed: call initFromDeck for Brine module

akva2:
Downstream of https://github.com/OPM/opm-models/pull/589
Downstream of https://github.com/OPM/opm-material/pull/392


totto82:
Thanks. Tested from below.

PR #2325, opened 2020-02-07
Async output.

dr-robertk:
This PR restores the true asynchronicity for the output and ensures that all output is written before the simulator is finalized.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,7 @@ endif()
 option(SIBLING_SEARCH "Search for other modules in sibling directories?" ON)
 set( USE_OPENMP_DEFAULT OFF ) # Use of OpenMP is considered experimental
 option(BUILD_FLOW "Build the production oriented flow simulator?" ON)
+option(BUILD_FLOW_BLACKOIL_ONLY "Build the production oriented flow simulator only supporting the blackoil model?" OFF)
 option(BUILD_FLOW_VARIANTS "Build the variants for flow by default?" OFF)
 option(BUILD_EBOS "Build the research oriented ebos simulator?" ON)
 option(BUILD_EBOS_EXTENSIONS "Build the variants for various extensions of ebos by default?" OFF)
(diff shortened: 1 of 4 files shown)


dr-robertk:
jenkins build this please

dr-robertk:
@alfbr: This could fix your problem.

alfbr:
Fantastic, will test!

alfbr:
The build fails for me, does it build at your end?

dr-robertk:
jenkins build this please

dr-robertk:
@alfbr: I rebased on top master. Now it should build.

dr-robertk:
jenkins build this please

alfbr:
Tested, this solves the problems. Thank you!

dr-robertk:
@alfbr: Glad I could help. Before this is merged it should be clarified what the barrier was about to make sure this PR does not accidentally break something else.

ebos/eclwriter.hh
```diff
@@ -396,7 +396,7 @@ public:
 
             // then, make sure that the previous I/O request has been completed and the
             // number of incomplete tasklets does not increase between time steps
-            taskletRunner_->barrier();
+            // taskletRunner_->barrier();
```
dr-robertk: @goncalvesmachadoc: Could you comment on why this barrier was introduced?

bska:
The barrier was to ensure that, in the case of output operations taking unusually long&mdash;for instance when writing large amounts of output to an NFS-mounted volume&mdash;the memory requirements of the output thread didn't grow without bounds. In other words, if the simulator runs at normal speed but the output isn't able to keep up.

I know we saw that symptom earlier, but maybe we're in a better position now.

ebos/eclwriter.hh
bska: > [Why was] this barrier was introduced?

Just to avoid confusion: This line has been in place since commit 8f61fe03, and was originally introduced in opm-modules as commit OPM/opm-models@0fb23919 (PR OPM/opm-models#299).

dr-robertk:
jenkins build this please

ebos/eclwriter.hh
dr-robertk: I re-enabled the barrier to avoid problems I might not be aware of.

totto82:
jenkins build this please

GitPaean:
It is probably there is some delay until the new jenkins test can use the newly updated jenkins reference result.

akva2:
i manually ran opm-tests job immediately after merging so it should be fine.

totto82:
Thanks for the fix. The flow blackoil specialization is unreleated, but a wanted feature. It will reduce the compile time for developers.

PR #2326, opened 2020-02-07
fixed: build with dune-fem enabled

akva2:
in that case we do not have an 'instance' of the static mpi helper
class.


akva2:
jenkins build this please

akva2:
this should be non-controversial. merging as it has sat uncommented for 10 days.

PR #2328, opened 2020-02-07
Save and Restore Well's Current Control Target Mode

bska:
This PR uses the new restart file support for active well control target modes to save and restore the ~run's~ wells' active targets.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -213,7 +213,20 @@ namespace Opm {
             }
 
             Opm::data::Wells wellData() const
-            { return well_state_.report(phase_usage_, Opm::UgGridHelpers::globalCell(grid())); }
+            {
+                auto wsrpt = well_state_.report(phase_usage_, Opm::UgGridHelpers::globalCell(grid()));
+
+                for (const auto& well : this->wells_ecl_) {
+                    auto xwPos = wsrpt.find(well.name());
+                    if (xwPos == wsrpt.end()) { // No well results.  Unexpected.
+                        continue;
+                    }
+
+                    xwPos->second.current_control.isProducer = well.isProducer();
+                }
+
+                return wsrpt;
+            }
 
             // substract Binv(D)rw from r;
             void apply( BVector& r) const;
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -543,24 +543,6 @@ namespace Opm {
             wellsToState(restartValues.wells, phaseUsage, handle_ms_well, well_state_);
         }
 
-        // for ecl compatible restart the current controls are not written
-        const auto& ioCfg = eclState().getIOConfig();
-        const auto ecl_compatible_rst = ioCfg.getEclCompatibleRST();        
-        if (true || ecl_compatible_rst) { // always set the control from the schedule
-            for (int w = 0; w <nw; ++w) {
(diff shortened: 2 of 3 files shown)


bska:
Note that this will end up changing all restart files in runs for which the active constraint/target differs from the constraint declared in the input file.

bska:
jenkins build this please

bska:
Note that the sequential tests fail due to #1814 (commit a9c7d112) switching the regression tests to running in ECLIPSE-compatible mode.  The parallel regression tests run with [OPM-extended](https://github.com/OPM/opm-simulators/blob/d4419fedfb5df2f5f10bf83d78224f012c2826ac/tests/run-parallel-regressionTest.sh#L26) restart files.

bska:
I'm marking this as "ready for review" now.  My earlier reservations about updating reference data are less important now in my opinion.

bska:
jenkins build this please

opm/simulators/wells/BlackoilWellModel.hpp
```diff
@@ -184,7 +184,20 @@ namespace Opm {
...
 
             Opm::data::Wells wellData() const
-            { return well_state_.report(phase_usage_, Opm::UgGridHelpers::globalCell(grid())); }
+            {
+                auto wsrpt = well_state_.report(phase_usage_, Opm::UgGridHelpers::globalCell(grid()));
+
+                for (const auto& well : this->wells_ecl_) {
+                    auto xwPos = wsrpt.find(well.name());
+                    if (xwPos == wsrpt.end()) { // No well results.  Unexpected.
+                        continue;
```
totto82: Better to throw? Or can this happen in some corner cases. I thought wellMap and wells_ecl always had the same wells.

bska: > I thought `wellMap` and `wells_ecl` always had the same wells.

I think they do, and I'd be really surprised if this condition actually happened.  As for throwing an exception here, that's a little tenuous.  This function runs in parallel (I think) so exceptions are delicate.  I'd be happy to add an `assert` here instead although that too is troublesome in a parallel context.

bska:
> I haven't look at the opm-common part of this.

Those have been in place since OPM/opm-common#1446.  This PR essentially adds the active data to the output requests that already exist.  Until now we've output the controls specified in the simulation run's input data.

Note that due to the new dynamic data this ends up changing the reference solutions.  We will have to update those in order to merge this PR.

> Will the restart tests include the well control items?

Yes.  The restart regression tests "just" verify that the simulations restarted in the middle of the time series produce (roughly) the same results as the full forward simulation.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this opm-common=1637 please

bska:
Any more comments here?  If not I would like to get this into master now.  I'm going to rerun the build check to verify that the only remaining failure cases are those that relate to active control mode being different from the control mode requested in the pertinent `WCON*` keywords.

bska:
jenkins build this please

bska:
> I'm going to rerun the build check to verify that the only remaining failure cases are those that relate to active control mode being different from the control mode requested in the pertinent `WCON*` keywords.

All remaining test failure are indeed of this type.  That, in turn, requires updating the reference data.

totto82:
> Any more comments here?

Please merge at will, when green after updating the ref. data.

bska:

bska:
jenkins build this opm-tests=271 please

bska:
Merging per prior maintainer approval.

PR #2333, opened 2020-02-11
Remove members from WellSegment class

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/1460

### DIFF
--- a/opm/simulators/utils/ParallelRestart.cpp
+++ b/opm/simulators/utils/ParallelRestart.cpp
@@ -1352,13 +1352,7 @@ std::size_t packSize(const UnitSystem& data,
 std::size_t packSize(const WellSegments& data,
                      Dune::MPIHelper::MPICommunicator comm)
 {
-    return packSize(data.wellName(), comm) +
-           packSize(data.depthTopSegment(), comm) +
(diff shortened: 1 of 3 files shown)


joakim-hove:
Temporary close to help GitHub

GitPaean:
I might misread it and I have to go now. 

In the upstream, you removed the function `WellSegments::MultiPhaseModel WellSegments::multiPhaseModel() const {`, while I did not see the usage is removed here while it compiled. 

Excuse me if I got things wrong here.

joakim-hove:
> while I did not see the usage is removed here while it compiled.

I think you got that one wrong - it is removed here.

GitPaean:
I mean here.  The function `WellSegments::multiPhaseModel() const ` is removed from upstream.

https://github.com/OPM/opm-simulators/blob/0e45f0827ce06445ef1b4f40baaeb8dfcd763a9c/opm/simulators/wells/MultisegmentWell_impl.hpp#L1175

joakim-hove:
> I mean here. The function WellSegments::multiPhaseModel() const is removed from upstream.

OK - removed. Why did it compile/link at all?

PR #2336, opened 2020-02-13
Multiple group inj phases

joakim-hove:
This is just a rebased version of #2298 - I'll merge at the first scent of green.

Upstream: https://github.com/OPM/opm-common/pull/1410


joakim-hove:
See: https://github.com/OPM/opm-simulators/pull/2298#issuecomment-585591655

PR #2340, opened 2020-02-13
Remove serialization support for UDQFunctionTable and UDQFunction

joakim-hove:
The `UDQFunctionTable` can be instantiated without any user data - no need to pass between nodes.


joakim-hove:
jenkins build this please

PR #2342, opened 2020-02-16
Restart flow from restart state

joakim-hove:
Along with opm/opm-common#1436 this can now restart the testcases in opm-test with `--sched-restart=false`. There is is still *much* more work to be done before we can say that the case of restarting only from the restart-file is generally solved - see upstream.


joakim-hove:
jenkins build this opm-common=1436 please

joakim-hove:
jenkins build this opm-common=1436 please

joakim-hove:
jenkins build this opm-common=1436 please

PR #2346, opened 2020-02-17
add well potential to PRT 

goncalvesmachadoc:


atgeirr:
I'll review this today.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -194,15 +194,15 @@ public:
...
     }
 
     /*!
      * \brief Register all run-time parameters for the Vtk output module.
      */
     static void registerParameters()
     {
-        EWOMS_REGISTER_PARAM(TypeTag, bool, ForceDisableFluidInPlaceOutput,
+        EWOMS_REGISTER_PARAM(TypeTag, bool, ForceDisableReportsOutput,
                              "Do not print fluid-in-place values after each report step even if requested by the deck.");
```
atgeirr: This comment is now outdated. The flag applies to all reports, not just fluid-in-place.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1106,7 +1107,7 @@ public:
                     }
                     fipUnitConvert_(tmp);
                     Scalar regHydroCarbonPoreVolumeAveragedPressure = pressureAverage_(regPressurePvHydrocarbon[reg], regPvHydrocarbon[reg], regPressurePv[reg], regionFipValues[FipDataType::PoreVolume][reg], true);
-                    pressureUnitConvert_(regHydroCarbonPoreVolumeAveragedPressure);
+                    pressureUnitConvert_(regHydroCarbonPoreVolumeAveragedPressure);
```
atgeirr: Nitpick: delete the end-of-line whitespace.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1115,13 +1116,13 @@ public:
     }
 
     // write production report to output 
-    void outputProdLog(size_t reportStepNum, const bool substep, bool forceDisableProdOutput) 
+    void outputProdLog(size_t reportStepNum, const bool substep)
```
atgeirr: There is a lot of 8-space indentation here, but fixing that would edit every line... So just leave it like that and we'll eventually do an automatic formatting sweep.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1210,22 +1211,40 @@ public:
...
+                                        case Phase::GAS:   return "Gas";          
+                                        default:
+                                        {
+                                                return "";
+                                        }
+                                        }
+                                };
+                                const auto flowtype = ftype(prefPhase);
+                                Scalar flowPot = 0.0;                                
+                                if(flowtype == "Oil"){ flowPot = tmp_values[2]; }
```
atgeirr: Nitpick: put a space after `if`, and put the code in the blocks (here and the lines below) on separate lines:
```
if (x) {
    // code
} else if (y) {
    // other code
}
```

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1311,45 +1339,41 @@ public:
...
-                                tmp_values[3] = get("WWIR"); //WellInjDataType::WaterRate
-                                tmp_values[4] = get("WGIR"); //WellInjDataType::GasRate 
-                                tmp_values[5] = get("WVIR");//WellInjDataType::FluidResVol                              
-                                tmp_values[6] = get("WBHP"); //WellInjDataType::BHP
-                                tmp_values[7] = get("WTHP"); //WellInjDataType::THP
-                                //tmp_values[8] = 0; //WellInjDataType::SteadyStateII
+                                        flowPot = tmp_values[4];                                        
+                                }                                  
+                                
+                                tmp_values[8] = flowPot/(tmp_values[6] - reservoirAveragePressure_); //WellInjDataType::SteadyStateII
```
atgeirr: I assume this is the correct formula for the (prod/inj)ectivity index, but I am a little puzzled by the average pressure being involved?

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1915,77 +1939,77 @@ private:
...
+                ss << std::right << std::fixed << ":" << std::setw (8) << wellProdNames[WellProdDataType::WellName] << ":" << std::setprecision(0) << std::setw(11) << "" << ":" << std::setw(4) << wellProdNames[WellProdDataType::CTRLMode] << ":" << std::setprecision(1) << std::setw(11) << wellProd[WellProdDataType::OilRate] << ":" << std::setw(11) << wellProd[WellProdDataType::WaterRate] << ":" <<  std::setw(11)<< wellProd[WellProdDataType::GasRate] << ":" <<  std::setw(11) << wellProd[WellProdDataType::FluidResVol] << std::setprecision(3) << ":" <<  std::setw(11) << wellProd[WellProdDataType::WaterCut] << std::setprecision(2) << ":" <<  std::setw(10) << wellProd[WellProdDataType::GasOilRatio] << std::setprecision(4) << ":" <<  std::setw(12) << wellProd[WellProdDataType::WatGasRatio] << std::setprecision(1) << ":" <<  std::setw(8) << "" << ":" <<  std::setw(8) << "" << ":" <<  std::setw(23) << ": \n";
```
atgeirr: These lines are unreadably long (I know, that is not new for this PR). Please break them to make it readable.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1311,45 +1339,41 @@ public:
                 
                                 const auto flowtype = ftype(injType);
                                 const auto flowctl = fctl(ctlMode);
+                                Scalar flowPot = 0.0;
                                 if(flowtype == "Oil") //WellInjDataType::CTRLModeOil
                                 {                    
                                         if (flowctl == "RATE"){ tmp_names[1] = "ORAT"; }
                                         else { tmp_names[1] =  flowctl; }
+                                        flowPot = tmp_values[2];
```
atgeirr: Here and above, this looks like it does not get the well potentials, but the actual flow rates. The calculations further down (line 1362) seem to compute some sort of injectivity index based on the flow rates (same for the productivity index on line 1233).

Is this the quantity sought? Should we not output the actual potentials at least some time? I admit the output string confuses me: "STEADY-STATE PI OR POTN. OF PREF. PH." -- when should I expect to get the potentials?

goncalvesmachadoc: Eclipse outputs the well potential instead of PI only when the drainage radius (DRADIUS) is set to negative in the input deck. I think we could output both. Any suggestion/preference? @jalvestad

ebos/ecloutputblackoilmodule.hh
goncalvesmachadoc: I was/am not sure about what to output.; this eq. is what came out from discussions during the meeting. Should I get the well index and well potential directly from well state?

ebos/ecloutputblackoilmodule.hh
OPMUSER: I would prefer the default behavior to follow the commercial simulator's reporting to avoid confusion between it and OPM Flow, from a users prospective. If we want to deviate from this we can use a switch on the keyword , WELLS=6 for example, to implement an alternative printing option.

PR #2347, opened 2020-02-17
disable json input for boost 1.48

akva2:
Context: Now that we use std::filesystem, the brokenness of boost::property_tree is the only reason why we need to use non-standard boost packages on rhel6/7. Since I doubt anyone would be doing development on these platforms, at least while using standard boost, I deem it better to disable this functionality to make for pain-free installs for end users.

### DIFF
--- a/opm/simulators/linalg/setupPropertyTree.cpp
+++ b/opm/simulators/linalg/setupPropertyTree.cpp
@@ -21,6 +21,7 @@
 
 #include <opm/simulators/linalg/setupPropertyTree.hpp>
 
+#include <boost/version.hpp>
 #include <boost/property_tree/json_parser.hpp>
 
 namespace Opm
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

atgeirr:
We could change this to using the Dune property tree, not sure if there is something as convenient as the json reader for that, though?

akva2:
well. i'm not one to judge if that is warranted. rhel6 is on its way out, i'm just trying to keep it alive until that happens. i considered this simple way to do so, and hopefully in a release or so we don't have the problem any more.

blattms:
Not sure what you mean by as convenient.

For moving to Dune the format needs to changed to something that the [ParameterTree](https://dune-project.org/doxygen/master/classDune_1_1ParameterTree.html) can us, like:
```
[LinearSolver] #or whatever we choose
type = bicgstabsolver
verbose = 1
maxit = 20
reduction = 0.5
[LinearSolver.preconditioner]
type = cpr
[LinearSolver.preconditioner.fineSmoother]
type = ilu0
```

atgeirr:
> something that the [ParameterTree](https://dune-project.org/doxygen/master/classDune_1_1ParameterTree.html) can use

That does not look too bad, where is the format documented?

blattms:
LOL That is a good question.

https://en.wikipedia.org/wiki/INI_file has some ideas.

PR #2345, opened 2020-02-17
Remove using result_type = T

joakim-hove:
Silence a warning

PR #2355, opened 2020-02-19
quell warning about unused local typedef

akva2:
strictly speaking the gcc diagnostics are wrong, the typedef is actually
used once templates are invoked on the structs.
simply use the typedef in the structs itself to avoid the warning
like is done for other variants.

replaces https://github.com/OPM/opm-simulators/pull/2345


akva2:
jenkins build this please

akva2:
merging as green and approved.

PR #2356, opened 2020-02-19
Lessen boost usage

akva2:
- use std::any
- use std::optional
- use internal uppercasing
- avoid lexical_cast
- remove unnecessary includes


akva2:
jenkins build this please

blattms:
Seems like it. In it goes.

Issue #2359, opened 2020-02-21
Wither restart - opinions please!


joakim-hove:
As described here: https://github.com/OPM/opm-common/issues/1396 there is currently work ongoing to be able to restart the simulator in situations where don't have access to the historical schedule file. That capability is currently needed within Equinor.

This issue is about what we should do in situations where we have the historical schedule file, i.e. the only situation we can currently restart from. My initial plan was to change the behavior so that the historical restart section was unconditionally ignored in the case of a restart - that is in accordance with how Eclipse does it, and certainly give many dog-food points. However initializing the `Schedule`object purely from the restart file is complex, and if we just over night change to that approach it is certainly possible we end with a regression - not being able to restart a model which we have previously been able to restart - that would be unfortunate.

So - all in all I am currently leaning towards using the historical schedule file - if we have it. It could be configurable (compile time / run time ??); and also we could *plan* to move in the direction of unconditionally using only the restart file to initialize the `Schedule` object.
 
Opinions?!

OPMUSER:
My preference would be to follow the commercial simulator straight away, combined with testing as many models as possible before the release. Its already confusing having two types of restart files and methods for restarting, the quicker we move to the industry standard the better.

Just my two cents.

joakim-hove:
> My preference would be to follow the commercial simulator straight away

I agree that would be the best - but it is also the most risky alternative.

>  Its already confusing having two types of restart files and methods for restarting

Mind you - this is not about the format for the restart files; it is whether we should use them (irrespective of OPM extensions) to initialize the well and group description, or if we should prefer the historical schedule input - when it is available.

jalvestad:
There are several Schedule data options in E100 that imply dynamic changes to Schedule objects during the simulation and that are not captured by the Schedule data input, but which are stored in the Eclipse-compatible Restart file. Examples of this are: i) Changes to wells etc. due to ACTIONX data (e.g. shutting and opening of wells etc.), ii) Changes to wells using GCONPROD and for procedure on exceeding a maximum rate limit set to WELL (which is used by relevant assets), PLUG - plugging  back worst-offending well, and even CON (or +CON) which  means shutting off connections, and iii) WECON (and GECON - not used in flow yet) which can cause shutting in well or shutting off connections. 

[...]

PR #2365, opened 2020-02-25
Add disabled functionality to init wells and group from restart file

joakim-hove:
This PR contains the changes necessary in opm-simulators to be able to initialize purely from a restart file. Currently the interesting code is guarded by a
```
const bool init_from_restart_file = false;
```
i.e. it is *disabled*. The current PR can and should(?) be merged now, but actually setting the switch to `true` requires that the work in opm-common is completed. 

Eventually the `init_from_restart_file` should be fixed to `true` - but in an interim period it can optionally be a runtime configurable option - based on e.g. a commandline switch, an env variable, the lunar phase, .... ?

See discussion here: #2359 and here: https://github.com/OPM/opm-common/issues/1396

### DIFF
--- a/flow/flow.cpp
+++ b/flow/flow.cpp
@@ -47,6 +47,9 @@
 #include <opm/common/OpmLog/EclipsePRTLog.hpp>
 #include <opm/common/OpmLog/LogUtil.hpp>
 
+#include <opm/io/eclipse/rst/state.hpp>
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
Ping

atgeirr:
Looks like a good way to start. Merging.

PR #2368, opened 2020-02-27
Initialize phase mix with (1/3, 1/3, 1/3) for producer

joakim-hove:
When a well is declared with WELSPECS there is an item to specify the well's preferred[1] phase, this phase is not (yet at least) stored in the restart files written by flow - which implies that it is tricky to get back when initializing from a restart file.

The callsite which is removed in this PR is the only callsite in all of opm; so I suggest this PR.

[1]: What exactly *preferred* implies is less than clear to me.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -2405,9 +2405,9 @@ namespace Opm
                     mix[component] = std::fabs(q_out_perf[perf*num_comp + component]/tot_surf_rate);
                 }
             } else {
-                std::fill(mix.begin(), mix.end(), 0.0);
                 // No flow => use well specified fractions for mix.
                 if (this->isInjector()) {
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

OPMUSER:
Preferred phase means basically the type of well, oil well for example, even though the well may gas out and produce lots of gas and little oil, but is still to be treated as an oil well.

atgeirr:
Changing this initialization will change absolutely every simulation case, and it would have been better to avoid it, or if necessary, to replace it with a "better" initialization. I see the restart problem however. Is it not likely that there is a spot in the restart file for this flag, that we missed? We will need to use the well's preferred phase for certain things related to group controls later in any case, so this is just a temporary workaround.

joakim-hove:
> Changing this initialization will change absolutely every simulation case

yes - I saw that; must admit I was quite surprised by that - an initialization should not change the results so fundamentally - I thought.

Anyway; will try to find in the restart.

PR #2369, opened 2020-02-27
Restricted Additive Schwarz with multiple layers of overlap

andrthu:
Add option to use Restricted Additive Schwarz (RAS) as parallel preconditioner instead of Block-Jacobi (BJ). This is made possible by adding multiple layers of overlap between subdomains after partitioning the grid, and PR depend on: https://github.com/OPM/opm-grid/pull/449. The overlap layers are separated into two types: The Overlap layer(s) and the outer Ghost layer. The matrix entries in the overlap rows are then valuable and can be used in the preconditioner. There is no need for a RAS preconditioner class, since the index set, the comm.project function and comm.copyOwnerToAll function handles everything for us. 

To demonstrate the impact of RAS on the convergence, consider Norne with default partitioning options using 1 and 2 layers of overlap.    

Norne with transmissibility edge-weights: 
Command: mpirun -np N flow Norne.DATA --overlap-layers=L --edge-weights-method=1. 
| N | L=1(BJ) iter | L=2(RAS) iter|
|----|-----------------|-------|
| 1  |  25167 | 25167|
|2  | 25103 | 24753 |
|4 | 25000 | 25328|
|8  | 25328 | 25797|
|16  | 26137 | 25591|
|32  | 26178 | 25925|
|64  | 26505 | 25309|

We observe a slight reduction in iteration count when using RAS over BJ for N=2,16,32 and 64, but higher for n=4 and 8. The main reason for these unimpressive results are that parallel flow on Norne with transmissibility weights produce almost no increase in iterations compared to sequential flow. For different or larger cases running on more processors the benefit could be greater.

To see an example where RAS has a more significant impact, consider Norne again, but now use uniform edge-weights to partition the grid.

Norne with uniform edge-weights:
Command: mpirun -np N flow Norne.DATA --overlap-layers=L --edge-weights-method=0. 
| N | L=1(BJ) iter | L=2(RAS) iter| L=3(RAS) iter |
|----|-----------|---------------|--------------|
| 1    |  25167       |     25167             |   25167 |
|2     | 26533       | 26819                | 26789|
|4     | 27771       | 25276                |25706 |
|8     | 29555      | 26889                | 26844|
|16    | 29695      |28466                |28762|
|32   | 43380      |32572                 |28371|
|64   | 50486      | 34684               | 28006|

Here we see a lower iteration count for all N except 2. We also observe that the benefit of RAS increase with N. 
 
Note that the tables show total number of iterations and not total time. The lower iteration count of RAS comes at the cost of increased parallel overhead.


CMakeLists.txt
```diff
@@ -234,21 +234,23 @@ opm_add_test(flow_blackoil_dunecpr
...
   LIBRARIES "opmsimulators")
 
-opm_add_test(flow_onephase
-  ONLY_COMPILE
-  DEFAULT_ENABLE_IF ${FLOW_VARIANTS_DEFAULT_ENABLE_IF}
-  SOURCES flow/flow_onephase.cpp
-  EXE_NAME flow_onephase
-  DEPENDS "opmsimulators"
-  LIBRARIES "opmsimulators")
+if (FALSE)
```
akva2: nope.

ebos/eclcpgridvanguard.hh
```diff
@@ -201,13 +212,20 @@ public:
                 ActiveGridCells activeCells(grid().logicalCartesianSize(),
                                             grid().globalCell().data(), grid().size(0));
                 this->schedule().filterConnections(activeCells);
-            }
+            }
```
akva2: nope

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -90,6 +90,111 @@ DenseMatrix transposeDenseMatrix(const DenseMatrix& M)
...
+    typedef Y range_type;
+    typedef typename X::field_type field_type;
+
+#if HAVE_MPI
+    typedef Dune::OwnerOverlapCopyCommunication<int,int> communication_type;
+#else
+    typedef Dune::CollectiveCommunication< int > communication_type;
+#endif
+
+#if DUNE_VERSION_NEWER(DUNE_ISTL, 2, 6)
```
akva2: 2.6 is minimum, no need for the ifdefery

blattms:
Please also report timings.

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -391,6 +392,11 @@ class WellModelMatrixAdapter : public Dune::AssembledLinearOperator<M,X,Y>
             // Construct scalar product.
             auto sp = Dune::createScalarProduct<Vector,POrComm>(parallelInformation_arg, category);
 
+            // If overlapLayers_ > 1 we are using Restricted Additive Schwarz and therefore need
+            // the residual at overlap DoFs.
+            if (overlapLayers_ > 1)
+                parallelInformation_arg.copyOwnerToAll(istlb, istlb);
```
blattms: Please explain why this is only needed for more than one overlap layers. Shouldn't the right hand side always be consistent (all procs have the same values?)

blattms: Please explain why this is only needed for more than one overlap layers. Shouldn't the right hand side always be consistent (all procs have the same values?)

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -688,7 +694,7 @@ class WellModelMatrixAdapter : public Dune::AssembledLinearOperator<M,X,Y>
                     pattern[idx].insert(*wc);
 
                 // Add just a single element to ghost rows
-                if (elem.partitionType() != Dune::InteriorEntity)
+                if (elem.partitionType() == Dune::GhostEntity)
```
blattms: This is totally unrelated to this PR but it just struck me when reading the diff:

Should we move the for-loop over the wellConnectionsGraph_ to else branch and explicitly use `noGhostMat_->setrowsize(1);`. That seems clearer and might prevent surprises if we distribute the wells.

GitPaean:
Thanks for the efforts. I will try to test how it works with the model 2 variants.

GitPaean:
Unfortunately,  when tried to run this PR for model 2 in parallel with 4 processes (have not tried other options) with `--overlap-layers=L --edge-weights-method=1 `, 

The following error message was given, 
```
Warning: The completion keywords like: WMCTL are not supported at: /home/kaib/test/model2_new_realisations2/real10/include/omega-ncs-2013.smry, line 274
Program threw an exception: Allocating the simulation vanguard failed: RangeError [get:/home/kaib/local/dune2.6.0/release2.6.0_O3_mpi/include/dune/common/parametertree.hh:181]: Cannot parse value "L" for key ".OverlapLayers"RangeError [parse:/home/kaib/local/dune2.6.0/release2.6.0_O3_mpi/include/dune/common/parametertree.hh:258]:  as a int (on rank 0)

  Process name: [[675,1],1]
  Exit code:    1
```

GitPaean:
Please ignoring my previous message about the parallel running failure with model 2. 
```
 --overlap-layers=L 
```

I did not replace `L` with a number.

GitPaean:
preliminary testings with with two realizations of model2, with `overlap-layers=1` and 4 processes each, it did not provide anything positive in term of the running time and also convergence behaviors, while only slightly worse.

Maybe we need more tests. 

I think it is something good to have for later testing and study, or something we can keep improving.

opm/simulators/linalg/ISTLSolverEbos.hpp
hnil: The right hand side is normally unique, which is correct with only one overlap since this will give Dirichlet condition for the copy (?). I think the problem is that of some reason the residual is only calculated for the owner cells since this is the only needed for the original solvers. With this copying and the additional setting of zero for the copy attribute (i.e. "useless overlap") the right hand side is correct. The correct thing is to assemble for all owner and overlap (I would have prefered all) and then set it to zero (at the same time as the matrix elements is fixed to get correct Dirichlet condition). If amg solver is used I think a change in the redistribution function is needed to get correct behavoir if the coarse scale communicator has overlap attributes.

hnil:
If the only reasonable solver with none empty "overlap attributes"  why do we need an extra option for it?

PR #2372, opened 2020-02-28
Use compressed properties in material law setup

akva2:
Downstream of https://github.com/OPM/opm-material/pull/408


blattms:
puuh. nearly missed merging this. Probably need to get my temperature tested as I am unfamiliarly trigger-happy today.

PR #2373, opened 2020-02-28
fix build without MPI

akva2:
*blush*


akva2:
quick shameful self-merge to restore CI.

PR #2374, opened 2020-02-28
Replace remaining cartesian field property usage with compressed properties

akva2:
This enables the use of distribution field properties everywhere.

PR #2377, opened 2020-03-02
changed: use compressed field properties setting up numbers

akva2:
Let's go slow here. Split out from https://github.com/OPM/opm-simulators/pull/2374

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -2850,32 +2850,22 @@ private:
         }
     }
 
-
-    static bool has_int_prop(const EclipseState& eclState, const std::string& prop) {
-        return eclState.fieldProps().has_int(prop);
-    }
-
-    static std::vector<int> get_int_prop(const EclipseState& eclState, const std::string& prop) {
-        return eclState.fieldProps().get_global_int(prop);
-    }
-
     template<class T>
     void updateNum(const std::string& name, std::vector<T>& numbers)
     {
         const auto& simulator = this->simulator();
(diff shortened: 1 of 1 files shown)


akva2:

akva2:
i hope to use your fever to my advantage ;P

ebos/eclproblem.hh
```diff
@@ -2850,32 +2850,22 @@ private:
...
-        const auto& numData = get_int_prop(eclState, name);
+        const auto& numData = eclState.fieldProps().get_int(name);
         const auto& vanguard = simulator.vanguard();
 
         unsigned numElems = vanguard.gridView().size(/*codim=*/0);
         numbers.resize(numElems);
         for (unsigned elemIdx = 0; elemIdx < numElems; ++elemIdx) {
-            unsigned cartElemIdx = vanguard.cartesianIndex(elemIdx);
-            numbers[elemIdx] = static_cast<T>(std::max(numData[cartElemIdx], 1) - 1);
+            numbers[elemIdx] = numData[elemIdx] -1;
```
blattms: That change I do not get.
If taking `std::max` was necessary before, why isn't now, too. The mapping should element to numData should be the same, no?

Or was it superfluous before, too?

akva2: it was always superfluous. bska was overly concerned with non-active cells, i did not fight it back when i had to introduce it. i can keep it but it serves no role (for valid data).

akva2: that being said i should keep the cast.

PR #2378, opened 2020-03-02
changed: use compressed field properties for blackoil extension init conds

akva2:


akva2:

blattms:
On a side note: In the plots we compare the parallel run with 4 procs with flow_legacy and the commercial simulator. Of course there will be slight discrepancies which hard to judge without an acompanying serial run.

akva2:
okay, i'll make sure to run serial and parallel in future prs.

blattms:
ah, if there has been a serial run then it will be taken into account. Cool.

PR #2379, opened 2020-03-02
use compressed field properties setting up explicit initial conditions

akva2:


akva2:

blattms:
Glad to see that even the jenkins master himself is not understood by his dear butler.

akva2:
can't distinguish serial and parallel in well plots. merging.

PR #2380, opened 2020-03-02
Make Cmake version check for CUDA work for CMake 3.6.

### LINKED ISSUE #2375: Build fails with CMake 3.6
Cmake fails with the following message:

CMake Error at CMakeLists.txt:28 (if):
   if given arguments:
     "NOT" "CMAKE_DISABLE_FIND_PACKAGE_CUDA" "AND" "CMAKE_VERSION" "VERSION_GREATER_EQUAL" "3.8"
 
   Unknown arguments specified

blattms:
It seems like the VERSION_GREATER_EQUAL operator for boolean expressions was introduced after CMake 3.6 and hence the current check whether to activate CUDA or not is broken in version 3.6 and below.

This PR fixes this by using VERSION_GREATER.

Closes #2375.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,8 +25,11 @@ cmake_minimum_required (VERSION 2.8)
 # out. It seems to ignore CMAKE_NVCC_FLAGS and CMAKE. Additionally
 # our way of specifying cuda source files never worked for CMake
 # version < 3.8. Hence we deactivate cuda for these versions.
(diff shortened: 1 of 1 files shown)


blattms:
Will selfmerge after a jenkins test.

Just learned that I cannot compile CMake-3.6 on Debian buster. Luckily they have working tarballs.

blattms:
jenkins build this please

PR #2385, opened 2020-03-02
use compressed field properties setting up rock parameters

akva2:

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -2186,13 +2186,13 @@ private:
         // read the parameters for water-induced rock compaction
         readRockCompactionParameters_();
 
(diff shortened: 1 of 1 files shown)


akva2:

blattms:
Where is this specific to CpGrid?

bska:
> Where is this specific to `CpGrid`?

The statement
```C++
rockTableIdx_[elemIdx] = num[elemIdx] - 1;
```
implicitly assumes that the active element ordering in `EclProblem <>` is the same as the one in the input file (i.e., an IJK-like ordering with I increasing the most rapidly).

akva2:
this would have been remapped earlier on when setting up the distributed properties. granted, not implemented for non-cpgrid yet, but that's the idea.

PR #2386, opened 2020-03-02
use compressed indices setting up transmissibilities

akva2:

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -547,9 +547,9 @@ private:
         ElementMapper elemMapper(gridView, Dune::mcmgElementLayout());
 
         const auto& fp = vanguard_.eclState().fieldProps();
-        const auto& inputTranxData = fp.get_global_double("TRANX");
-        const auto& inputTranyData = fp.get_global_double("TRANY");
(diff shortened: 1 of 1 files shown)


akva2:

ebos/ecltransmissibility.hh
```diff
@@ -575,8 +575,8 @@ private:
 
                 auto isId = isId_(c1, c2);
 
-                int gc1 = std::min(cartMapper.cartesianIndex(c1), cartMapper.cartesianIndex(c2));
-                int gc2 = std::max(cartMapper.cartesianIndex(c1), cartMapper.cartesianIndex(c2));
+                int gc1 = std::min(c1, c2);
+                int gc2 = std::max(c1, c2);
```
bska: This doesn't look right to me, at least not in isolation.  The subsequent code block assumes that the `gc1` and `gc2` quantities are global (i.e., Cartesian) indices.  If they're not then the difference checks don't do what they're supposed to do.  We must either leave `gc1` and `gc2` intact or update the checks.

akva2: yes, this was overzealous. need to keep gc1/gc2, only change indexing below.

akva2:

akva2:
okay to merge?

PR #2390, opened 2020-03-03
tighening tolerance for testing purpose

GitPaean:
brine model is disabled because it can not finish running.


GitPaean:

akva2:
for this to be productive, you shouldn't change the defaults, you should do like https://github.com/OPM/opm-simulators/blob/master/compareECLFiles.cmake#L479 on a per-test basis.

GitPaean:
> for this to be productive, you shouldn't change the defaults, you should do like https://github.com/OPM/opm-simulators/blob/master/compareECLFiles.cmake#L479 on a per-test basis.

You are very right! Thanks for pointing out this.

PR #2398, opened 2020-03-03
tightening the tolerance for polymer cases for testing purpose

GitPaean:

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -263,20 +263,22 @@ add_test_compareECLFiles(CASENAME polymer_oilwater
                          FILENAME 2D_OILWATER_POLYMER
                          SIMULATOR flow
                          ABS_TOL ${abs_tol}
-                         REL_TOL ${rel_tol})
+                         REL_TOL ${rel_tol}
+                         TEST_ARGS --tolerance-mb=1.e-7 )
 
 add_test_compareECLFiles(CASENAME polymer_injectivity
                          FILENAME 2D_POLYMER_INJECTIVITY
                          SIMULATOR flow
                          ABS_TOL ${abs_tol}
-                         REL_TOL ${rel_tol})
+                         REL_TOL ${rel_tol}
+                         TEST_ARGS --tolerance-mb=1.e-7 --max-strict-iter=20 --tolerance-wells=1.e-6)
 
 add_test_compareECLFiles(CASENAME polymer_simple2D
                          FILENAME 2D_THREEPHASE_POLY_HETER
                          SIMULATOR flow
                          ABS_TOL ${abs_tol}
                          REL_TOL ${coarse_rel_tol}
-                         TEST_ARGS --flow-newton-max-iterations=20)
+                         TEST_ARGS --tolerance-mb=1.e-7 --flow-newton-max-iterations=20)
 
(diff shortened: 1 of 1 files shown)


GitPaean:

GitPaean:
With this PR, the testing failure related to polymer cases are recovered from my local machine. @alfbr 
```
The following tests FAILED:
	 37 - compareECLFiles_flow+3D_MSW (Failed)
	 43 - compareECLFiles_flow+MSW_MODEL_1 (Failed)
	 45 - compareECLFiles_flow+FAULTS_MODEL_1 (Failed)
	 57 - compareECLFiles_flow+TEST_WSEGSICD (Failed)
```

GitPaean:
And there is no guarantee this PR does not generate regressions on other people's local test. With tightened tolerance, it is likely to cause time step chopping. It is hard to make sure the chopping behavior are the same.

alfbr:
Excellent! So that means that the stricter tolerances are sufficient. However, are they necessary? I will comment inline on the specific tolerances.

> And there is no guarantee this PR does not generate regressions on other people's local test.

No, but if your machine and Jenkins agree on the solution, that is a good start :-)

compareECLFiles.cmake
```diff
@@ -263,20 +263,22 @@ add_test_compareECLFiles(CASENAME polymer_oilwater
...
+                         REL_TOL ${rel_tol}
+                         TEST_ARGS --tolerance-mb=1.e-7 )
 
 add_test_compareECLFiles(CASENAME polymer_injectivity
                          FILENAME 2D_POLYMER_INJECTIVITY
                          SIMULATOR flow
                          ABS_TOL ${abs_tol}
-                         REL_TOL ${rel_tol})
+                         REL_TOL ${rel_tol}
+                         TEST_ARGS --tolerance-mb=1.e-7 --max-strict-iter=20 --tolerance-wells=1.e-6)
```
alfbr: Do we know that all three are necessary? I.e., if you remove on of them, will the test still pass on your machine?

compareECLFiles.cmake
```diff
@@ -263,20 +263,22 @@ add_test_compareECLFiles(CASENAME polymer_oilwater
...
 
 add_test_compareECLFiles(CASENAME polymer_simple2D
                          FILENAME 2D_THREEPHASE_POLY_HETER
                          SIMULATOR flow
                          ABS_TOL ${abs_tol}
                          REL_TOL ${coarse_rel_tol}
-                         TEST_ARGS --flow-newton-max-iterations=20)
+                         TEST_ARGS --tolerance-mb=1.e-7 --flow-newton-max-iterations=20)
```
alfbr: Same here, if you remove the --flow-newton-max-itrations=20, will that be sufficient for the test to pass on your machine?

GitPaean: `--flow-newton-max-iterations=20` was existing, but I can test it out.

compareECLFiles.cmake
GitPaean: `--tolerance-mb=1.e-7 --max-strict-iter=20 --tolerance-wells=1.e-6` were added in the order as it shows. I can also test it by removing one of the first two.  Only the third one will not work.

PR #2400, opened 2020-03-04
tightening the tolerance for polymer cases for testing purpose

GitPaean:


GitPaean:

GitPaean:
By removing some options, regression test failure related to two polymer cases did not show up
```
The following tests FAILED:
	 37 - compareECLFiles_flow+3D_MSW (Failed)
	 43 - compareECLFiles_flow+MSW_MODEL_1 (Failed)
	 45 - compareECLFiles_flow+FAULTS_MODEL_1 (Failed)
	 57 - compareECLFiles_flow+TEST_WSEGSICD (Failed)
```

compareECLFiles.cmake
```diff
@@ -263,20 +263,22 @@ add_test_compareECLFiles(CASENAME polymer_oilwater
...
+                         REL_TOL ${rel_tol}
+                         TEST_ARGS --tolerance-mb=1.e-7 )
 
 add_test_compareECLFiles(CASENAME polymer_injectivity
                          FILENAME 2D_POLYMER_INJECTIVITY
                          SIMULATOR flow
                          ABS_TOL ${abs_tol}
-                         REL_TOL ${rel_tol})
+                         REL_TOL ${rel_tol}
+                         TEST_ARGS --tolerance-mb=1.e-7 --tolerance-wells=1.e-6)
```
alfbr: Please also test if you can remove the --tolerance-wells

alfbr:
This is good, it simplifies the set-up significantly. Please also see if we can do with only the material balance on both.

compareECLFiles.cmake
GitPaean: from the previous PR, no for this case, because it is the last one added. Anyway, I can create a PR to verify.

alfbr: Sorry, then I misunderstood your previous comment. Please try removing --tolerance-mb=1.e-7 then.

GitPaean: It is not sufficient, which was tested already.

GitPaean: anyway, I will create a test just in case.

alfbr: If I understand correctly, the test will break at your end if one of them is removed, regardless which?

GitPaean: Yes. I believe so. And I could not verify for only `--tolerance-mb=1.e-7 ` for now , due to the following PR did not manage to update the reference results due to compilation error.  But with only `--tolerance-wells=1.e-6` it does not work. 

https://github.com/OPM/opm-simulators/pull/2404

alfbr: I see, yes, please check it out. If the material balance is sufficient there too, we have one common switch for all polymer cases. That is just about as simple as we can get it.

GitPaean: For this case, it needs both `--tolerance-mb=1.e-7` and `--tolerance-wells=1.e-6` on my machine. 

As stated previously, adding these two arguments likely break some other people's local regression test.

alfbr:
> As stated previously, adding these two arguments likely break some other people's local regression test.

Well, unless we disable the tests, we have no other option to make it pass for you. Chances are it will then also pass for all. Ideally, we should re-write the polymer tests to ensure better convergence. Numerically unstable tests are troublesome. But these are the tests we have right now. I would like to merge this. @akva2 do you have any reservations against merging this?

akva2:
Nope. As long as jenkins still pass, worst case we fixed nothing.

alfbr:
@GitPaean please ensure that update_data are up to date with the current tolerances. Once Jenkins is green, you can merge.

atgeirr:
Just a thought on these tests: If we see varying time-stepping behaviour on different machines, leading to test failures, that should mean we are close to the convergence radius of the Newton method, and reducing the time steps taken in the deck itself could resolve that by ensuring everybody takes the same time steps.

alfbr:
True, reducing time steps is one possible solution to make the tests more numerically stable. It will need to be tested though.

GitPaean:
> Well, unless we disable the tests, we have no other option to make it pass for you. 

I do not see it is my problem and the local regression test failure for me does not bother me that much. I do think it will be good if we can improve the test so that most of people can have relatively consistent regression test, which should be our aim. 

>  we should re-write the polymer tests to ensure better convergence.

We should do that. 

>  Once Jenkins is green, you can merge.

I hesitate to do that because I am not sure it fixes more problem or creates more problem.

alfbr:
> I hesitate to do that because I am not sure it fixes more problem or creates more problem.

No problem, I will do it. Please just make sure update_data is up-to-date and Jenkins green.

GitPaean:
> Please just make sure update_data is up-to-date and Jenkins green

It is up to date. I will perform a jenkins build now.

GitPaean:
jenkins build this opm-tests=249 please

alfbr:
All green, merging :-)

PR #2403, opened 2020-03-04
Implements a Python step() method

hakonhagland:
Implements a Python `step()` method that can be used to advance the simulator one report step at a time. This PR builds on the Python bindings introduced in PR #2127 . Since that PR is not merged yet I initially did: 
```
git fetch upstream pull/2127/head
git checkout -b python-step FETCH_HEAD
git rebase master
```
This PR should probably be split into some smaller pull requests if it was possible to submit pull-request that depend on each other. I could not find an easy way to do that, so instead it is split up into a a series of smaller commits. If this PR gets merged, the plan is to submit another PR that implements setting and getting of fluid state and material variables from Python. This can then be used to e.g. run updates on porosity from a chemical mineralization model implemented in Python for each report step.

## Overview of the commits 

- c66b564 : Refactor `run()` into `prepareRun_()`

We would like to add a `step()` method to simulators.cpp, that will let a Python script advance the simulator one report step at a time. Before calling `step()`, the Python script will have to call a `step_init()` method that initializes the simulator, similar to what is done in `run()` now. In order to avoid duplication of code in `run()` and `step_init()`, the `run()` method is here refactored into a `prepareRun_()` function that can be called from both `step_init()` and from `run()`.

- 2a1c1f9 : refactor `flowEbosBlackoilMain()`

Currently, the Python run() calls `flowEbosBlackoilMain()` to start the simulation (which calls `execute()` in class `FlowMainEbos`). The Python `step_init()` method should also do the setup in `flowEbosBlackoilMain()` but it should not call `execute()` directly. In order to avoid code duplication for `run()` and `step_init()`, the initialization part of `flowEbosBlackoilMain()` is isolated into a `flowEbosBlackoilMain_init()` method.

- 77a151b : refactor `execute()`

Currently, execute() calls `runSimulator()` to run the simulation. When the Python `step_init()` is implemented (see a later commit), it will instead call an `executeInitStep()` that will need to do the same initialization as in `execute()` except that it should call a `runSimulatorInit()` instead of
`runSimulator()`. In order to avoid code duplication for `execute()` and `executeStepInit()`, `execute()` is here refactored into an `execute_()` method.

-  313f2be : refactor `runSimulator()`

As discussed for the previous commit 77a151b, `runSimulator()` needs to be refactored to avoid code duplication when `executeStepInit()` is implemented (see later commit). Here, `runSimulator()` is refactored into a `runSimulatorInitOrRun()` that takes a callback function. When `runSimulatorInit()` is implemented it will pass a different callback that only initializes the simulator. Currently, `runSimulator()` passes the callback `runSimulatorRunCallback_()` which runs the whole simulation.  The final output of the simulation summary in `runSimulator()` is also refactored into a
`runSimulatorAfterSim_()` method.

- 460c672 : refactor `run()` in `SimulatorFullyImplicitBlackoilEbos.hpp`

Similar to the previous commit 313f2be, `run()` is here refactored in preparation for the implementation of a Python `step()` function in a later commit. Currently `run()` is called from `runSimulatorInitOrRun()` in `FlowMainEbos.hpp` using the `runSimulatorRunCallback_()`. Later, there will be another callback `runSimulatorInitCallback_()` that will need to call different parts of
the code in `run()`. The `run()` function is thus refactored into `run()`, `runInit()`,
`runStep()`, and `runLastStep()`. Also, some of the local variables in `run()` have to be made persistent between calls to `runStep()`, this applies to variables `report`, `stepReport`, `solverTimer`, `totalTimer`, and `adaptiveTimeStepping`, which are made private class variables.

-  f94f6dc : Implement Python `step_init()` method

The `step_init()` method initializes the simulation. It is required for the Python script to run `step_init()` before calling the `step()` method (which will be implemented in the next commit).  

-  8e4f748 : Implement Python `step()` method.

The Python `step()` method advances the simulator one report step. Before calling `step()` for the first time, `step_init()` must have been called.

### DIFF
--- a/python/simulators/simulators.cpp
+++ b/python/simulators/simulators.cpp
@@ -295,6 +295,14 @@ public:
             return EXIT_FAILURE;
         }
     }
+
+    int step()
+    {
+        if (!hasRunInit_) {
+            throw std::logic_error("step() called before step_init()");
+        }
+        return mainfunc_->executeStep();
+    }
 private:
 
     bool prepareRun_()
@@ -472,5 +480,6 @@ PYBIND11_MODULE(simulators, m)
         .def("set_eclipse_state", &BlackOilSimulator::setEclipseState)
         .def("set_schedule", &BlackOilSimulator::setSchedule)
         .def("set_summary_config", &BlackOilSimulator::setSummaryConfig)
+        .def("step", &BlackOilSimulator::step)
         .def("step_init", &BlackOilSimulator::step_init);
 }
--- a/python/simulators/simulators.cpp
+++ b/python/simulators/simulators.cpp
@@ -268,9 +268,21 @@ public:
 
     int run()
     {
+        if (prepareRun_()) {
+            return Opm::flowEbosBlackoilMain(argc_, argv_, outputCout_, outputFiles_);
+        }
+        else {
+            return EXIT_FAILURE;
+        }
+    }
+
+
+private:
+
+    bool prepareRun_()
+    {
         int argc = argc_;
         char **argv = argv_;
-
         Dune::Timer externalSetupTimer;
         externalSetupTimer.start();
 
@@ -315,9 +327,9 @@ public:
         }
 
         FileOutputMode outputMode = FileOutputMode::OUTPUT_NONE;
-        bool outputCout = false;
+        outputCout_ = false;
         if (mpiRank == 0)
-            outputCout = EWOMS_GET_PARAM(PreTypeTag, bool, EnableTerminalOutput);
+            outputCout_ = EWOMS_GET_PARAM(PreTypeTag, bool, EnableTerminalOutput);
 
         std::string deckFilename;
         std::string outputDir;
@@ -343,7 +355,7 @@ public:
             }
         }
 
-        if (outputCout) {
+        if (outputCout_) {
             Opm::FlowMainEbos<PreTypeTag>::printBanner();
         }
 
@@ -354,7 +366,7 @@ public:
                                     deckFilename,
                                     outputDir,
                                     EWOMS_GET_PARAM(PreTypeTag, std::string, OutputMode),
-                                    outputCout, "STDOUT_LOGGER");
+                                    outputCout_, "STDOUT_LOGGER");
 
             if (mpiRank == 0) {
                 setupMessageLimiter(schedule_->getMessageLimits(), "STDOUT_LOGGER");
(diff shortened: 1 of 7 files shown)


joakim-hove:
> This PR should probably be split into some smaller pull requests if it was possible to submit pull-request that depend on each other.

That is not technically possible; but what is common is to create PR1 with commit1 and then PR2 with commit1 *and* commit2. In the PR description of PR2 you then write that "This PR depends on PR1 which should be merged first" - this works quite well.

To simplify the review process I would recommend doing such a split for this PR. I don't have the full overview, but my suggestion would be that the first PR contains *only* preparatory changes to the C++ code which will affect all of `flow` - i.e. nothing from the `python/` subdirectory; and possibly also those preparations can be split in more than one PR.

joakim-hove:
jenkins build this please

python/simulators/simulators.cpp
```diff
@@ -0,0 +1,485 @@
...
+
+// this is a dummy type tag that is used to setup the parameters before the actual
+// simulator.
+NEW_TYPE_TAG(FlowEarlyBird, INHERITS_FROM(EclFlowProblem));
+
+END_PROPERTIES
+
+namespace detail
+{
+    boost::filesystem::path simulationCaseName( const std::string& casename ) {
```
joakim-hove: After switching to gcc7 as minimum requirement we now use filesystem from `std` and not not boost - this will not compile with a new version of the codebase.
```C++
namespace fs = Opm::filesystem;
...
```

python/simulators/simulators.cpp
```diff
@@ -0,0 +1,485 @@
...
+        namespace fs = boost::filesystem;
+
+        const auto exists = []( const fs::path& f ) -> bool {
+            if( !fs::exists( f ) ) return false;
+
```
joakim-hove: The file `python/simulators/simulators.cpp` *seems* to have much overlap with the existing simulator implementation; but this file should wait for a next PR anyway.

joakim-hove:
```
git fetch upstream pull/2127/head
git checkout -b python-step FETCH_HEAD
git rebase master
```
This code is moving fast - you need to do the 
```
git fetch upstream
git rebase upstream/master
```
dance on a regular basis.

atgeirr:
I strongly support the kind of refactoring you are doing to the C++ code, but do not have the time now to do a proper review. I'd very much like to, though. So what are the time constraints on this work?

hakonhagland:
> That is not technically possible; but what is common is to create PR1 with commit1 and then PR2 with commit1 and commit2. In the PR description of PR2 you then write that "This PR depends on PR1 which should be merged first" - this works quite well.  


@joakim-hove Yes, that sounds like a good idea! I am splitting it up into smaller PRs (each matching a commit in this PR 2403):

- PR #2438  : Refactor run() into prepareRun_(), depends on PR #2127
- PR #2439  : refactor flowEbosBlackoilMain(), depends on PR #2438
- PR #2440  : refactor execute(), depends on PR #2439
- PR #2441  : refactor runSimulator(), depends on PR #2440
- PR #2442  : refactor run() in SimulatorFullyImplicitBlackoilEbos.hpp, depends on PR #2441
- PR #2443  : Implement Python step_init() method, depends on PR #2442
- PR #2444  : Implement Python step() method, depends on PR #2443

> jenkins build this please

@joakim-hove How do I do that?

> I strongly support the kind of refactoring you are doing to the C++ code, but do not have the time now to do a proper review. I'd very much like to, though. So what are the time constraints on this work?

@atgeirr Thanks. I am not aware of any time constraints yet. I will discuss it with the project leader (Tor-Harald).

joakim-hove:
> @joakim-hove How do I do that?

You must be white-listed by @akva2 to be able to initiate a jenkins run.

akva2:
whitelisted for opm-simulators.

hakonhagland:
> You must be white-listed by @akva2 to be able to initiate a jenkins run

Ok I seem to be white-listed now, so what do I do next?

akva2:
https://wiki.opm-project.org/index.php?title=Jenkins_triggers

hakonhagland:
@akva2 Ok then I think I understand. From the wiki page:
> Put the trigger in a separate GitHub comment containing only the trigger line.
> General syntax: `jenkins build this module-name=<module pr number> please`

So the comment "jenkins build this please" was not meant for me, but a way to trigger a build of the current PR?

akva2:
yup.

hakonhagland:
Closing this PR since it was replaced by #2735 after the refactoring of `flow.cpp` in #2516.

PR #2408, opened 2020-03-04
changed: no need to serialize BrineModule any longer

akva2:
Downstream of https://github.com/OPM/opm-models/pull/601


blattms:
This conflict is a bit unfortunate. Will try to resolve.

akva2:
Should be trivial, just changes adjacent lines.. only got my phone here.

PR #2409, opened 2020-03-04
Add commandline switch --sched-restart to control restart init

joakim-hove:
Add a commandline switch `--sched-restart` to flow which can regulate how the schedule information should be initialized when restarting. When using `--sched-restart=true` the simulator will use the historical schedule section to initialize, whereas with `--sched-restart=false` the simulator will initialize from the restart file. The ability to initialize from the restart file is very much work in progress, and this will go through several stages:

1. (This PR) flow can *not* initialize from the restart file - and if you pass `--sched-restart=false` flow will bail with an exception; the default is of course `--sched-restart=true`.

2. When the basic functionality is in place we remove the `throw` and flow will initialize from the restart file if you pass `--sched-restart=false` - but the default will be to use the historical section if it is present (i.e. the current behavior). We keep things like this for a period with testing.

3. We switch the default `--sched-restart=false`- i.e. unless explicitly asked otherwise we will use the restart file to initialize.

4. The switch is completely removed - and restarting from the restart file is the *only* possibility.

1-2-3 - all hold your breath :-)

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -73,6 +73,7 @@ NEW_PROP_TAG(EclDeckFileName);
 NEW_PROP_TAG(OutputDir);
 NEW_PROP_TAG(EnableOpmRstFile);
 NEW_PROP_TAG(EclStrictParsing);
+NEW_PROP_TAG(SchedRestart);
 NEW_PROP_TAG(EclOutputInterval);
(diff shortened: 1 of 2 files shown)


joakim-hove:
jenkins build this please

atgeirr:
Sounds like a good plan! Merging.

PR #2424, opened 2020-03-06
rebased: changed: no need to serialize BrineModule any longer

blattms:
Rebased version of conflicting PR #2408 from @akva2


blattms:
Compared the diffs and the lines added and removed here and in #2408 looke the same. Trusting my eyes and not disturbing jenkins I will merge.

PR #2419, opened 2020-03-06
changed: no need to serialize PolymerModule any longer

akva2:
Downstream of https://github.com/OPM/opm-models/pull/602

PR #2425, opened 2020-03-06
Rebased #2419: changed: no need to serialize PolymerModule any longer

blattms:
Rebased version of @akva2's PR #2419 to remove conflicts.


akva2:
Remove comm and ser

PR #2426, opened 2020-03-08
Shrink EclEquilInitializer::initialFluidStates_.

blattms:
There is not reason to it as a vector of cartesian size as we access its entries by the compressed element index. This should save space and speedup the lookup.

A further step would be to either remove this class or move the initialization into an function of the class that directly return the vector. The class is currently only temporarily initiated by eclproblem to copy out the values of the vector [here](/OPM/opm-simulators/blob/master/ebos/eclproblem.hh#L2408-L2423)


blattms:
jenkins build this please

blattms:

atgeirr:
Code looks fine and well plots too. Merging!

PR #2432, opened 2020-03-09
Add --sched-restart flag when running restart regression tests

joakim-hove:
Adds a flag `--sched-restart` to the test driver for restart tests. For now the value `true` is passed - which corresponds to the current default (and only working alternative). This is a preparation to enable regression testing of the feature: opm/opm-common#1396

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -426,12 +426,14 @@ add_test_compare_restarted_simulation(CASENAME spe1
                                       FILENAME SPE1CASE2_ACTNUM
                                       SIMULATOR flow
                                       ABS_TOL ${abs_tol_restart}
-                                      REL_TOL ${rel_tol_restart})
+                                      REL_TOL ${rel_tol_restart}
+                                      TEST_ARGS --sched-restart=true)
 add_test_compare_restarted_simulation(CASENAME spe9
                                       FILENAME SPE9_CP_SHORT
                                       SIMULATOR flow
                                       ABS_TOL ${abs_tol_restart}
-                                      REL_TOL ${rel_tol_restart})
+                                      REL_TOL ${rel_tol_restart}
+                                      TEST_ARGS --sched-restart=true)
 
 # PORV test
 opm_set_test_driver(${PROJECT_SOURCE_DIR}/tests/run-porv-acceptanceTest.sh "")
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

akva2:
it would work, but seems unnecessary complicated. can't you just pass the extra argument in TEST_ARGS ?

joakim-hove:
> it would work, but seems unnecessary complicated. can't you just pass the extra argument in TEST_ARGS ?

I might misunderstand, but to me it seems like the `${TEST_ARGS}` was supposed to be used in "rest of the arguments in list context" - but the restart driver seems to access it as a scalar root name - so to me it looks I can not "just pass the extra argument in TEST_ARGS". 

Happy to update the PR if I am wrong though.

akva2:
you are wrong. TEST_ARGS is a multiarg, it takes many arguments.

```
TEST_ARGS --sched-restart=true 
```
should(tm) do the job.

joakim-hove:
> TEST_ARGS is a multiarg, it takes many arguments.

yes the `cmake` part is fine - but how about [this](https://github.com/OPM/opm-simulators/blob/master/tests/run-restart-regressionTest.sh#L20) usage in the driver script?

akva2:
from above

```
shift 10
TEST_ARGS="$@"
```

shift drops the first 10 command arguments, then TEST_ARGS is set to $@ (all remaining command line parameters).

joakim-hove:
> shift drops the first 10 command arguments, then TEST_ARGS is set to $@ (all remaining command line parameters).

and then 
```
BASE_NAME=`basename ${TEST_ARGS}_RESTART.DATA`
```
evaluates to?

akva2:
now i see your point. strictly speaking you introduced that line back in 2018, but it was based on bad stuff done by me earlier. this should be fixed. there is no reason for the TEST_ARGS stuff we do in the add_test_compare_restarted_simulation. i'll fix that first.

joakim-hove:
> strictly speaking you introduced that line back in 2018

Ahhh - I was afraid of that.

>  i'll fix that first.

Thank's

joakim-hove:
OK - rebased on top of #2345.

joakim-hove:
jenkins build this please

PR #2438, opened 2020-03-09
Refactor run() into prepareRun_()

hakonhagland:
The first PR in the splitting of PR #2403 into smaller PRs as  [suggested](https://github.com/OPM/opm-simulators/pull/2403#issuecomment-595606972) in the comments. 

NOTE: This PR depends on PR #2127 which should be merged first.

## Description
We would like to add a `step()` method to `simulators.cpp`, that will let a Python script advance the simulator one report step at a time. Before calling `step()`, the Python script will have to call a
`step_init()` method that initializes the simulator, similar to what is done in `run()` now. In order to avoid duplication of code in `run()` and `step_init()`, the `run()` method is here refactored into a `prepareRun_()` function that can be called from both `step_init()` and from `run()`.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -1,2 +1,3 @@
 add_subdirectory( pybind11 )
+add_subdirectory( simulators )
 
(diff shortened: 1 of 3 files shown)


hakonhagland:
Closing this PR since it was reorganized by #2543 which was later superseded by #2555 and #2609.

PR #2439, opened 2020-03-09
refactor flowEbosBlackoilMain()

hakonhagland:
The second PR in the splitting of PR #2403 into smaller PRs.

NOTE: This PR depends on PR #2438 which should be merged first.

## Description
Currently, the Python `run()` calls `flowEbosBlackoilMain()` to start the simulation (which calls execute() in class `FlowMainEbos`). The Python `step_init()` method should also do the setup in `flowEbosBlackoilMain()` but it should not call `execute()` directly. In order to avoid code duplication for `run()` and `step_init()`, the initialization part of `flowEbosBlackoilMain()` is isolated into a `flowEbosBlackoilMain_init()` method.

PR #2440, opened 2020-03-09
refactor execute()

hakonhagland:
The third PR in the splitting of PR #2403 into smaller PRs.

This PR depends on PR #2439 which should be merged first.

## Description
Currently, `execute()` calls `runSimulator()` to run the simulation. When the Python `step_init()` is implemented (see a later commit), it will instead call an `executeInitStep()` that will need to do the same initialization as in `execute()` except that it should call a `runSimulatorInit()` instead of 
`runSimulator()`. In order to avoid code duplication for `execute()` and `executeStepInit()`, `execute()` is here refactored into an `execute_()` method.


hakonhagland:
Closing this PR since it was superseded by #2619.

hakonhagland:
Closing this PR since it was superseded by #2619.

PR #2441, opened 2020-03-09
refactor runSimulator()

hakonhagland:
The 4th PR in the splitting of PR #2403 into smaller PRs.

This PR depends on PR #2440 which should be merged first.

## Description

As discussed in the previous PR #2440,  `runSimulator()` needs to be refactored to avoid code duplication when `executeStepInit()` is implemented (see later commit). Here, `runSimulator()` is refactored into a `runSimulatorInitOrRun()` that takes a callback function. When `runSimulatorInit()` is implemented it will pass a different callback that only initializes the simulator. Currently, `runSimulator()` passes the callback `runSimulatorRunCallback_()` which runs the whole simulation. The final output of the simulation summary in `runSimulator()` is also refactored into a
`runSimulatorAfterSim_()` method.


hakonhagland:
Closing this PR since it was superseded by #2631.

PR #2447, opened 2020-03-09
changed: simplify eclmpiserializer

akva2:
now users only have to implement a single method for
packsize/pack/unpack.

Prerequisites:
https://github.com/OPM/opm-material/pull/412
https://github.com/OPM/opm-simulators/pull/2436

### DIFF
--- a/ebos/eclmpiserializer.hh
+++ b/ebos/eclmpiserializer.hh
@@ -27,23 +27,45 @@ namespace Opm {
 
 class EclMpiSerializer {
 public:
+    enum class Operation {
+        PACKSIZE,
+        PACK,
+        UNPACK
+    };
+
     explicit EclMpiSerializer(Dune::CollectiveCommunication<Dune::MPIHelper::MPICommunicator> comm) :
         m_comm(comm)
     {}
 
     template<class T>
-    std::size_t packSize(const T& data) {
-        return Mpi::packSize(data, m_comm);
+    void operator()(const T& data)
+    {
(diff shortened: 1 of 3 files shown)


atgeirr:
jenkins build this please

ebos/eclmpiserializer.hh
```diff
@@ -27,23 +27,45 @@ namespace Opm {
...
-    std::size_t packSize(const T& data) {
-        return Mpi::packSize(data, m_comm);
+    void operator()(const T& data)
+    {
+        if (m_op == Operation::PACKSIZE)
+            m_packSize += Mpi::packSize(const_cast<const T&>(data), m_comm);
+        else if (m_op == Operation::PACK)
+            Mpi::pack(const_cast<const T&>(data), m_buffer, m_position, m_comm);
+        else if (m_op == Operation::UNPACK)
+            Mpi::unpack(const_cast<T&>(data), m_buffer, m_position, m_comm);
```
bska: Why do you need `const_cast<const T&>` here?  Is the `data` not already a `const T&`?

bska:
The build test failed, but I don't really see why it should.  It's possible this just happened to race with OPM/opm-tests#261.  I'll try to rerun the build test.

bska:
jenkins build this please

ebos/eclmpiserializer.hh
akva2: I don't need them, leftovers from a previous attempt. removed now.

akva2:
jenkins build this please

PR #2446, opened 2020-03-09
Autocreate parallel fieldprops and use compressed ones in initstateequil.hh

blattms:
With this PR `ParallelEclipseState` will be able to autocreate data for somekeywords that have not been explicitly specified in the input file (just like the serial version). For now this still happens by wasting some space, creating global arrays of cartesian size and then compressing them manually. We use this to get rid off using the get_global_* accessors initstateequil.hh except for the temperature stuff which appears to be buggy in parallel (issue #2423) and hence might lead to differing results.

As a drive-by improvement we get rid off (at least) one unneccessary copy (PVTNUM) and on my system this give us  a 10% performance boost for Norne on 4 processors due to faster assembly.

PR #2452, opened 2020-03-10
Use correct region temperature when calculating RS/RV

### LINKED ISSUE #2423: Parallel initstateequil.hh runs on a distributed grid but seems to assume a global one.
I was about to clean up initstateequil.hh as it seemed that getting the global field properties is superfluous. I have to admit that this code leaves me a bit clueless. Is this well tested in parallel?

What strikes me is this:
 [In the constructor of InitialStateComputer](OPM/opm-simulators/blob/a449b2e2e6a2f1bf420a20440f4b70b321d3e64d/ebos/equil/initstateequil.hh#L933) it's `temperature_` member is a vector of the number of elements of the local grid (called after loadbalance). Later in the constructor there is a call to [updateInitialTemperature](/OPM/opm-simulators/blob/a449b2e2e6a2f1bf420a20440f4b70b321d3e64d/ebos/equil/initstateequil.hh#L1086) which overwrites the vector by the global field property with key "TEMPI".  But the grid passed to the constructor is a loadbalanced parallel one (I see this in my gdb session, the backtrace shows that this is called during `problem_finishInit` in the Simulator of opm-models/opm/models/utils/simulator.hh). Later on this array is then used in `Opm::EQUIL::computeRs` and this function seems to assume that the temperature for the i-th cell its stored in `temperature_[i]`, but this array seems to be unchanged from before.

It also seems that all the region stuff cannot work on a distributed grid as it does see the cells of the region that belong to another processor.

I am lost, what am I missing?

blattms:
The `InitialStateComputer::temperature_` array, previously used, was of Cartesian size, but used as if it only contained the values of the cells of a region. With this commit `InitialStateComputer::temperature_` is a compressed local array and we explicitly copy out only the region values when computing RS/RV.

Closes #2423.

Needs  #2446 to be merged first! Won't work otherwise

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -526,8 +526,8 @@ if(MPI_FOUND)
   add_test_compare_parallel_simulation(CASENAME spe1_thermal
                                        FILENAME SPE1CASE2_THERMAL
                                        SIMULATOR flow
-                                       ABS_TOL ${abs_tol_parallel}
-                                       REL_TOL 1e-1
+                                       ABS_TOL ${abs_tol}
+                                       REL_TOL ${rel_tol}
                                        DIR spe1
                                        TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8)
 
--- a/ebos/equil/initstateequil.hh
+++ b/ebos/equil/initstateequil.hh
@@ -1082,7 +1082,7 @@ private:
     void updateInitialTemperature_(const Opm::EclipseState& eclState)
     {
         // Get the initial temperature data
-        std::vector<double> tempiData = eclState.fieldProps().get_global_double("TEMPI");
+        std::vector<double> tempiData = eclState.fieldProps().get_double("TEMPI");
         temperature_ = tempiData;
     }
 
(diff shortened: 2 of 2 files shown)


akva2:
can you tighten the parallel SPE1CASE2_THERMAL tolerances now?

blattms:
Haven't tried. Whenever I try to use the jenkibs scripts locally as described in the [wiki](https://wiki.opm-project.org/index.php?title=Jenkins_triggers#Using_the_scripts_locally), I somehow get stuck. Could use some help.

How would I check that locally?

akva2:
don't have to use jenkins scripts for that.

edit https://github.com/OPM/opm-simulators/blob/master/compareECLFiles.cmake#L530 to something more sane, then run the test using ctest -R Parallel.*THERMAL

blattms:
I could have sworn that the test failed when I tried it on a debug build even for the default settings.

When retrying (release build) it works with tolerances "0.02" "1e-5".

This makes me a bit anxious (even more tighter tolerances seem possible)!

Also please note that my changes, might also change results for serial runs. Even for these we neglected the regions
when querying temperatures for RS/RV. Carefully examination by others is therefore advised.

akva2:
jenkins build this please

blattms:
Thanks for the approval.

I still think that somebody needs to manually verify the results as I am assuming
that for a case with several different equil regions the results of a serial
run should differ from the previous version.

blattms:
Will anybody check this for a case with temperature and multiple regions?

bska:
> Will anybody check this for a case with temperature and multiple regions?

I'll echo this request.  As it happens, I'm doing a bit of work in this area (simulation run initialization) at the moment and I would like to base that work on @blattms's fix here.  While even this PR may not be completely correct, it is my opinion that it is definitely more correct than the current master sources.

blattms:
Cool. Then we should just merge.

bska:
> Cool. Then we should just merge.

Agreed.  I'm merging now and take responsibility if something breaks since I'm working in this area anyway.  Thanks for the fix.

PR #2453, opened 2020-03-10
drop serialization from ParallelEclipseState

akva2:
now done in base class.

Downstream of https://github.com/OPM/opm-common/pull/1552

### DIFF
--- a/opm/simulators/utils/ParallelEclipseState.cpp
+++ b/opm/simulators/utils/ParallelEclipseState.cpp
@@ -136,25 +136,6 @@ ParallelEclipseState::ParallelEclipseState(const Deck& deck)
 }
 
 
-#if HAVE_MPI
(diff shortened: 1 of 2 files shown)


bska:
> now done in base class.

Okay, but where's the call to `EclipseState::serializeOp<>()`?

akva2:
https://github.com/OPM/opm-simulators/blob/master/flow/flow.cpp#L402

bska:
>>> now done in base class.
>> 
>> Okay, but where's the call to `EclipseState::serializeOp<>()`?
>
> [Line 402 of `flow.cpp`](https://github.com/OPM/opm-simulators/blob/5b457cbbd62f90cf8225b04b9d343f335c663a1a/flow/flow.cpp#L402)

I see.  Thanks for the pointer.  I'll merge into master.

PR #2456, opened 2020-03-12
changed: move the serialization template definitions for generic scalars

akva2:
this to avoid having to explicitly instance the templates for
various enums when serialization functions are moved into classes.

@blattms this is a compromise between control of the number of instanced templates
and convenience.


akva2:
jenkins build this please

blattms:
How does this impact compilation time?

akva2:
not noticable as far as i can tell. note this is only used for a very few types, enums, int, double, char, size_t, everything else is overloaded anyways. sum total it will improve compilation time because i avoid a lot of of includes in ParallelRestart.cpp just to get to enums.

PR #2458, opened 2020-03-12
Do not use local id set to index matrices.

blattms:
That the local ids were consecutive and starting from 0 was just a coincidence and they should never be used to access linear systems or vectors. This commit fixes this by using the correct mappers instead.

It is a preparation needed to fix the idsets of CpGrid.

Note the we removed some computations from the constructor of ISTLSolverEbosCpr as it inherits from ISTLSolverEbos and the operations already happnen in constructor of the base class.

### DIFF
--- a/opm/simulators/linalg/ISTLSolverEbos.hpp
+++ b/opm/simulators/linalg/ISTLSolverEbos.hpp
@@ -364,7 +364,13 @@ protected:
 
             if (!ownersFirst_ || parameters_.linear_solver_use_amg_  || parameters_.use_cpr_ ) {
                 detail::setWellConnections(gridForConn, wellsForConn, useWellConn, wellConnectionsGraph_);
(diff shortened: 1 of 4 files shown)


blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
ok I missed the switch about ebos variants. Let's try again.

blattms:
jenkins build this please

blattms:
jenkins run norne norne_parallel

blattms:
Where are my manners?

blattms:

blattms:
norne plots look good before [the rebase](https://ci.opm-project.org/job/opm-simulators-PR-builder/760/). doing only  a simple check now.

blattms:
jenkins build this please

blattms:
shameless self-merge as it turned out that ISTLSolverEbosCpr.hpp is not even used.

PR #2459, opened 2020-03-12
Make available current Production/Injection Group Control parameters for use in  Restart

jalvestad:
This PR has changes that make parameters: currentProductionGroupControl and currentInjectionGroupControl available for writing Eclipse compatible restart files and for use as Summary file data. The content of the restart files will be changed as a consequence of using this information, to be included in a separate PR in opm-common.  

I would like to get this pull request merged soon since it is required to use together with other PRs for further work / improvment of the eclipse compatible restart files.

### DIFF
--- a/ebos/eclwellmanager.hh
+++ b/ebos/eclwellmanager.hh
@@ -42,6 +42,7 @@
 #include <opm/output/eclipse/RestartValue.hpp>
 
 #include <opm/output/data/Wells.hpp>
+#include <opm/output/data/Groups.hpp>
 #include <opm/material/common/Exceptions.hpp>
 
 #include <opm/models/utils/propertysystem.hh>
@@ -565,6 +566,10 @@ public:
         }
 
         return wellDat;
+    }
+
+    Opm::data::Group groupData(const int reportStepIdx, Opm::Schedule& sched) const {
+        return {};
     }
 
     /*!
--- a/ebos/eclwellmanager.hh
+++ b/ebos/eclwellmanager.hh
@@ -568,7 +568,7 @@ public:
         return wellDat;
     }
 
-    Opm::data::Group groupData(const int reportStepIdx, Opm::Schedule& sched) const {
+    Opm::data::Group groupData(const int /* reportStepIdx */, Opm::Schedule& /* sched */) const {
         return {};
     }
 
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -263,10 +263,6 @@ public:
 
         Opm::data::Group localGroupData = simulator_.problem().wellModel().groupData(reportStepNum, simulator_.vanguard().schedule());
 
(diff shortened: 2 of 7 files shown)


bska:
jenkins build this please

bska:
> jenkins build this please (failed)

Right, the PR also depends on OPM/opm-common#1408 and cannot be tested in isolation.

bska:
jenkins build this opm-common=1592 please

ebos/eclwellmanager.hh
```diff
@@ -567,6 +568,10 @@ public:
         return wellDat;
     }
 
+    Opm::data::Group groupData(const int reportStepIdx, Opm::Schedule& sched) const {
```
bska: The parameters here are unused.  I suggest writing the declaration as
```C++
Opm::data::Group groupData(const int /* reportStepIdx */, const Opm::Schedule& /* sched */) const {
```
In other words, without naming the parameters.  Doing so will avoid warnings about unused parameters.

ebos/eclwriter.hh
```diff
@@ -376,6 +385,7 @@ public:
             bool enableDoublePrecisionOutput = EWOMS_GET_PARAM(TypeTag, bool, EclOutputDoublePrecision);
             const Opm::data::Solution& cellData = collectToIORank_.isParallel() ? collectToIORank_.globalCellData() : localCellData;
             const Opm::data::Wells& wellData = collectToIORank_.isParallel() ? collectToIORank_.globalWellData() : localWellData;
+            const Opm::data::Group& groupData = collectToIORank_.isParallel() ? collectToIORank_.globalGroupData() : localGroupData;
```
bska: I don't think this `groupData` object is actually used, since the relevant quantities have already been calculated in `evalSummaryState()` and put into the `summaryState()` object.  The restart writer can retrieve the values from there.

ebos/eclwellmanager.hh
jalvestad: Thank you for the comment, I agree and this change has now been done and pr has been updated.

bska: > Thank you for the comment, I agree and this change has now been done and pr has been updated.

Thank you very much.  Did you want to address my other comment too, or do you consider this PR done now?

ebos/eclwriter.hh
jalvestad: I agree and I have removed this object from the code - thank you for the correction!

bska: > I agree and I have removed this object from the code

Thanks a lot!  I'll launch a data-update now.  I expect we'll be able to merge this later today.

bska:
jenkins build this opm-common=1592 update_data please

bska:
Green from upstream test run.  I'll merge this into master now.  Thanks a lot.

PR #2460, opened 2020-03-12
Remove old serialization for SummaryNode

akva2:
Downstream of https://github.com/OPM/opm-common/pull/1567

Note; the use of a lambda in the vector method might seem overkill but it will make more sense later.

### DIFF
--- a/ebos/eclmpiserializer.hh
+++ b/ebos/eclmpiserializer.hh
@@ -49,6 +49,31 @@ public:
     }
 
     template<class T>
+    void vector(std::vector<T>& data)
+    {
+        static_assert(!std::is_pod<T>::value, "Do not call this for POD vectors");
+        auto handle = [&](auto& d)
+        {
+            for (auto& it : d) {
+                it.serializeOp(*this);
+            }
+        };
+
+        if (m_op == Operation::PACKSIZE) {
+            m_packSize += Mpi::packSize(data.size(), m_comm);
+            handle(data);
+        } else if (m_op == Operation::PACK) {
+            Mpi::pack(data.size(), m_buffer, m_position, m_comm);
+            handle(data);
+        } else if (m_op == Operation::UNPACK) {
+            size_t size;
+            Mpi::unpack(size, m_buffer, m_position, m_comm);
+            data.resize(size);
+            handle(data);
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


ebos/eclmpiserializer.hh
```diff
@@ -48,6 +48,32 @@ public:
             Mpi::unpack(const_cast<T&>(data), m_buffer, m_position, m_comm);
     }
 
+    template<class T>
+    void vector(const std::vector<T>& datac)
```
blattms: Why const in the first place if you cast it away anyway

ebos/eclmpiserializer.hh
```diff
@@ -48,6 +48,32 @@ public:
...
+    {
+        std::vector<T>& data = const_cast<std::vector<T>&>(datac);
+
+        auto handle = [&](auto& d)
+        {
+            for (auto& it : d) {
```
blattms: SO this is never called for a vector of pods? Shouldn't there be at least a check (e.g. `static_assert<!std::is_pod<T>::values, "No no no">()`?

ebos/eclmpiserializer.hh
akva2: well, the "problem" is the the single serialize/deserialize approach in combination with the underlying serializer. I need to cast somewhere since packSize/pack takes consts and unpack take non-const and the compiler  is not smart enough to choose the correct template instance for packSize/pack if we pass a non-const reference. I earlier chose to eat by const ref in operator() to limit casting to one place.

strictly speaking i could only cast for operation == Operation::UNPACK here, but I kept it as-is for simplicity. i can do the limited casting if you prefer.

ebos/eclmpiserializer.hh
akva2: Well I guess that would yield a prettier error message. You'd get no member serializeOp if you try now.

ebos/eclmpiserializer.hh
akva2: Actually, come to think of it I cannot do this. The serializeOp member in the data types are non-const since they do both serialization and deserialization. What I can do, is not eat the parameter as const in the first place. And that makes a lot of sense, I got hung up in the operator() approach.

PR #2463, opened 2020-03-13
Remove obsolete information from the README

atgeirr:


atgeirr:
Merging right away, as this is only updating the README.

blattms:
Is that fixed in master, already?

atgeirr:
> Is that fixed in master, already?

Oh no, I see now that this was applied to the release branch. A user error from me for sure, I used the github online editing facility, chose to do it as a PR to be a bit careful, but still managed to fail... It is still correct for the release branch though, so no real harm done. 

I guess I should cherry-pick it to master.

PR #2464, opened 2020-03-13
Relperm diag no deck

joakim-hove:
Use the [EndpointScaling](https://github.com/OPM/opm-common/blob/master/opm/parser/eclipse/EclipseState/EndpointScaling.hpp) instead of `deck.hasKeyword("SCALESRC")` in the relperm diagnostic. *This reduces deck usage.*

Observe that the semantic of the two is not 100% identical:

Old:
```
if (deck.hasKeyword("SCALESRC")) {
    // do stuff 
}
```

New:
```
if (deck.hasKeyword("SCALESRC")) {
   const auto& keyword = deck.getKeyword("SCALESRC");
   if (keyword.getRecord(0).getItem(0) == "YES") {
        // do stuff
    }
}
```
i.e. the `EndpointSCaling`based implementation checks the value of the `SCALESRC`keyword, which the old implementation ignored.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -552,7 +552,7 @@ protected:
 
         if (enableExperiments) {
             Opm::RelpermDiagnostics relpermDiagnostics;
-            relpermDiagnostics.diagnosis(*eclState_, *deck_, asImp_().grid());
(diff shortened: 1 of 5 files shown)


joakim-hove:
jenkins build this please

PR #2469, opened 2020-03-13
Internal serialization in EclipseState - part 4

akva2:
Sits on top of https://github.com/OPM/opm-simulators/pull/2468

Downstream of https://github.com/OPM/opm-common/pull/1579


ebos/eclmpiserializer.hh
```diff
@@ -75,6 +81,44 @@ public:
...
+    void map(Map<Key, Data>& data)
+    {
+        auto handle = [&](auto& d)
+        {
+            if constexpr (is_vector<Data>::value)
+                vector(d);
+            else if constexpr (is_shared_ptr<Data>::value)
+                shared_ptr(d);
+            else
+                d.serializeOp(*this);
```
blattms: Looks somewhat familiar.

blattms: Aren't there occasions where data is a vector of shared_ptrs? What happens in this case. Or is this gone.

And again I am wondering where the shared_ptrs pointing to the same object are distributed. They are tricky. We definitely, breaking them in our serialization. When serialize and deserialize each shared_ptr will store its own object and have count 1. Will that become a problem?

Hence the question is: why are these shared_ptrs used?

ebos/eclmpiserializer.hh
```diff
@@ -145,6 +209,18 @@ protected:
             const_cast<T2&>(data.second).serializeOp(*this);
     }
 
+    template<class T1>
+    void shared_ptr(const std::shared_ptr<T1>& data)
+    {
+        bool value = data ? true : false;
+        (*this)(value);
+        if (m_op == Operation::UNPACK && value) {
+            const_cast<std::shared_ptr<T1>&>(data) = std::make_shared<T1>();
```
blattms: Does this require a default constructor?

ebos/eclmpiserializer.hh
akva2: Read the code. I have. Carefully. Analyzing every single use of the serialized shared_ptrs. And all of them safe. The actual shared pointers sit in schedule and/or DynamicState<>. Where they are appropriately handled (in the code you pointed out as duplicate). All of this is just like in the existing serialization code.

What happens for vectors of shared_ptr is that we handle them in the 'handle' lambda in the 'vector' member in eclmpiserializer.hh.

ebos/eclmpiserializer.hh
akva2: Yes.

ebos/eclmpiserializer.hh
blattms: I am bowing to the ground before you for that. I have done that myself for the restart stuff.

Good to know that this works.

My fear is that that ther Serializer will one day be treated as an interface, and people will not be aware of the smart pointer trickery, not read the code and happily serialize any kind of shared_ptrs and boom!

But maybe that is for another day to resolve.

blattms:
Also this read the code thing does not scale and quickly becomes expensive.

ebos/eclmpiserializer.hh
akva2: i totally see you fear. But it is very hard to address, I need to serialize what is at hand, and it's almost impossible to handle. We could add a check on the ref count, but that will mostly give a false sense of security.

akva2:
absolutely, i should have stated that otherwise. my point was that i have.

blattms:
I know you have. I was trying to raise general awareness and summon search engines.

PR #2470, opened 2020-03-13
adjust for SummaryNode -> SummaryConfigNode rename

akva2:
Downstream of https://github.com/OPM/opm-common/pull/1582


joakim-hove:
Thank you for the downstream fix :-)

PR #2471, opened 2020-03-13
fixed: do not build ParallelRestart test without MPI

akva2:
Argh, I always forget to check this!


akva2:

akva2:
shameful self-merge to restore serial build.

PR #2457, opened 2020-03-12
EclMpiSerializer for Schedule

akva2:
Downstream of https://github.com/OPM/opm-common/pull/1565

Closes #2467

PR #2474, opened 2020-03-13
Make flow variants use ParallelEclipseState just as flow.

blattms:
Without this patch they would have thrown a bad_cast exception when run with more than 1 process due to the recent changes of having EclipseState only on one process.

I do not claim that I have looked at the results for these.


blattms:
jenkins build this please

akva2:
... would be more productive if https://github.com/OPM/opm-simulators/pull/2457 was processed  first

PR #2483, opened 2020-03-18
Internal serialization in Schedule

akva2:
Downstream of https://github.com/OPM/opm-common/pull/1605


ebos/eclmpiserializer.hh
```diff
@@ -81,17 +85,21 @@ public:
         }
     }
 
-    template<template<class Key, class Data> class Map, class Key, class Data>
+    template<template<class Key, class Data> class Map, class Key, class Data, bool complexType = true>
```
blattms: Even here I do not get it.  Why does this not work:
```
template<class Map, bool complexType = true>
void map(Map& data){
   using Key = typename Map::key_type;
   using Data = typename Map::mapped_type;
  ....
```
Is this for casting?
Comment appreciated

akva2: It's a personal preference to always keep templates template parameters as templates to keep them as generic as possible but i guess in this case agree the cost is high. I have changed it.

akva2:
I also added two final bits; doxy for eclmpiserializer and I split out the serializaton to a separate compile unit to avoid compiling it for every flow variant.

PR #2484, opened 2020-03-19
avoid potentially uninitialized warnings in serial

akva2:
this was actually broken by an upstream change from return-by-reference
to return-by-value (https://github.com/OPM/opm-common/commit/a595e68608830b5c643f19f881422b7d4b80f357), but we were lucky, the return value was
kept in memory until end of block it seems.


akva2:

PR #2485, opened 2020-03-19
fixed: double-parsing of the deck

akva2:
externalDeck = nullptr -> internalDeck was parsed.
i have no idea why we have this external/internal deck in the first place.

now this caught me by surprise! i was doing memory profiling to see why memory usage in parallel was still unexpectedly high in parallel, and boy was i surprised when deck parsing showed up on non-root processes.


akva2:

blattms:
jenkins build this please.

joakim-hove:
> and boy was i surprised when deck parsing showed up on non-root processes.

You can check out any time you like
but you can never leave...

PR #2488, opened 2020-03-20
Fix CUDA version check to allow versions 9.0 or newer.

blattms:


blattms:
jenkins build this please

CMakeLists.txt
```diff
@@ -47,7 +47,12 @@ if(NOT CMAKE_DISABLE_FIND_PACKAGE_CUDA AND
   # only easy way to determine the cublas and cusparse libraries.
   # Hence we call it unconditionally
   # The WellContributions kernel uses __shfl_down_sync, which was introduced in CUDA 9.0
-  find_package(CUDA 9.0)
+  find_package(CUDA)
+  if(CUDA_VERSION VERSION_LESS "9.7")
```
bska: Was this supposed to say `VERSION_LESS "9.0"`?  If not, the diagnostic message should probably refer to version 9.7 instead.

blattms: *flush* This is exactly, why I should not self merge. The changed version number was for testing, only,

PR #2491, opened 2020-03-23
Fixes ILU0 when using flexible solvers.

### LINKED ISSUE #2490: FlexibleSolver with defaults never recalculates ILU pattern.
We end up with
```
Dune::DummyUpdatePreconditioner<Opm::ParallelOverlappingILU0<Dune::BCRSMatrix<Opm::MatrixBlock<double, 3, 3>, std::allocator<Opm::MatrixBlock<double, 3, 3> > >, Dune::BlockVector<Dune::FieldVector<double, 3>, std::allocator<Dune::FieldVector<double, 3> > >, Dune::BlockVector<Dune::FieldVector<double, 3>, std::allocator<Dune::FieldVector<double, 3> > >, Dune::Amg::SequentialInformation> >::update
```
which does nothing and we run into convergence issues for the simplest cases.

blattms:
when running flow_blackoil_dunecpr with the defaults we experienced severs convergence issues with SPE1CASE2, because the decomposition was never recalculated. This is fixed in PR.

Closes #2490

### DIFF
--- a/opm/simulators/linalg/PreconditionerFactory.hpp
+++ b/opm/simulators/linalg/PreconditionerFactory.hpp
@@ -156,26 +156,20 @@ private:
         using C = Comm;
         doAddCreator("ILU0", [](const O& op, const P& prm, const C& comm) {
             const double w = prm.get<double>("relaxation");
-#if DUNE_VERSION_NEWER(DUNE_ISTL, 2, 7)
-            return wrapBlockPreconditioner<DummyUpdatePreconditioner<SeqILU<M, V, V>>>(comm, op.getmat(), w);
-#else
-            return wrapBlockPreconditioner<DummyUpdatePreconditioner<SeqILU0<M, V, V>>>(comm, op.getmat(), w);
-#endif
+            return std::make_shared<Opm::ParallelOverlappingILU0<M, V, V, C>>(
+                op.getmat(), comm, 0, w, Opm::MILU_VARIANT::ILU);
         });
         doAddCreator("ParOverILU0", [](const O& op, const P& prm, const C& comm) {
             const double w = prm.get<double>("relaxation");
             // Already a parallel preconditioner. Need to pass comm, but no need to wrap it in a BlockPreconditioner.
-            return wrapPreconditioner<Opm::ParallelOverlappingILU0<M, V, V, C>>(
+            return std::make_shared<Opm::ParallelOverlappingILU0<M, V, V, C>>(
                 op.getmat(), comm, 0, w, Opm::MILU_VARIANT::ILU);
         });
         doAddCreator("ILUn", [](const O& op, const P& prm, const C& comm) {
             const int n = prm.get<int>("ilulevel");
             const double w = prm.get<double>("relaxation");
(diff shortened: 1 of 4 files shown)


blattms:
Not sure whether this is even built by jenkins.

blattms:
jenkins build this please

blattms:
Needs some more testing before merge.

blattms:
Locally tested. Convergence issues are gone with these changes.  Defaults are still not the same (e.g. --ilu-relaxation=1.0 versus flow using --ilu-relaxation=0.9). Ready for merge,

opm/simulators/linalg/PreconditionerFactory.hpp
```diff
@@ -156,26 +156,20 @@ class PreconditionerFactory
         using C = Comm;
         doAddCreator("ILU0", [](const O& op, const P& prm, const C& comm) {
             const double w = prm.get<double>("relaxation");
-#if DUNE_VERSION_NEWER(DUNE_ISTL, 2, 7)
-            return wrapBlockPreconditioner<DummyUpdatePreconditioner<SeqILU<M, V, V>>>(comm, op.getmat(), w);
-#else
-            return wrapBlockPreconditioner<DummyUpdatePreconditioner<SeqILU0<M, V, V>>>(comm, op.getmat(), w);
-#endif
+            return std::make_shared<Opm::ParallelOverlappingILU0<M, V, V, C>>(
```
atgeirr: This change (together with the "ILUn" change below) seems to eliminate the option to run the Block Jacobi/SeqILU preconditioner combination?

blattms: Well, flow does not use that combination either (see ISTLSolverEbos.hpp). If I am not completely mistaking `ParallelOverlappingILU0` should do the same if the matrix is correctly set up. My aim is to bring the codes paths closer together as they have been different  from the beginning and are diverging more and more...

The previous version using `DummyUpdatePreconditioner` never recalculated the decomposition as its update method is a no-op.

atgeirr: > Well, flow does not use that combination either (see ISTLSolverEbos.hpp). If I am not completely mistaking `ParallelOverlappingILU0` should do the same if the matrix is correctly set up. My aim is to bring the codes paths closer together as they have been different from the beginning and are diverging more and more...

I know regular Flow does not use that combination, just thought it was nice to offer the option (more options to experiment with is better). Which code paths is it you want to bring together?

blattms:
As it was, it was basically broken (admittedly, I did not check the parallel version).

The code paths are the various ISTLSolver*.hpp files which have diverged quite dramatically and will otherwise continue to do so.

blattms:

blattms:
If you insist I will add the ILU0 and ILUn stuff again...

atgeirr:
> If you insist I will add the ILU0 and ILUn stuff again...

No I will not insist. I'll merge now. If we decide we want to re-add it, it would not be hard. It may be a little bit misleading to have both "ILU0" and "ParOverILU0" yield the same preconditioner, but it's not a problem, and it makes the change backwards compatible.

blattms:
Cool, thanks.

PR #2495, opened 2020-03-23
changed: use AquiferConfig when setting up aquifers

akva2:

PR #2497, opened 2020-03-23
Avoid deck usage in simulator

akva2:
- Use unit system from state
- Remove additional call to checkMissingFeatures (already called from flow.cpp / flow_tag.hpp).

With these, https://github.com/OPM/opm-common/pull/1627 and https://github.com/OPM/opm-simulators/pull/2495 I can finally run flow without a deck instance. This potentially yields huge memory savings on root process in large parallel runs. 

I'm not trying to take the convenience vs sanity fight now, but just as a FYI; On one case I am looking at this means we need 4GB insteady of 8GB on root process..


akva2:
jenkins build this please

atgeirr:
Good, merge when green.

PR #2498, opened 2020-03-24
Cleaned up flexible solver improvements.

blattms:
As proposed I have tried to cherry-pick the stuff, that I liked from PR #1983 (skipping the write out parts, basically. Original message:

This pull request add:
-    trueimpes (this has a little over hea../opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.hppd if solver is recreated very often but this is not the intension of this solver.)
-    corrects citeria for amg agglomeration
-    put resonable hardcoded defaults for cpr solver
-  add support for (kamg). I think this is buggy as the hierachy will never be recalculated.

I spent quite some time cleaning it up (whitespace, tabs, etc.) and fixed some bugs in the PR that I noticed during premature testing.

One major change ist the way the weights used in CPR are handled. In the old version they were put into the property tree, and passed to contructor and update method. Now the constructor gets a std::function (re)calculating the weights. See 5789e6d. Admittedly, I thought this would be a bigger improvement than it appears to be in hindsight.

Includes PR #2491 as it is based on it.

This is WIP since I still want to test this a bit.


blattms:
Somewhere in the back of my head there is this faint memory that the flexible CPR solvers are supposed to deliver better performance for some models. What models are this and what parameters do I need to use?

blattms:
I guess with the latest fixes this is even a good improvement to the original PR and ready for review.

The previously proposed json-file (which failed for the original PR) works even better than the preconfigured configurations:
```
{
    "tol": "1e-2",
    "maxiter": "20",
    "verbosity": "0",
    "solver": "bicgstab",
    "preconditioner": {
        "type": "cpr",
        "pressure_var_index": "1",
        "finesmoother": {
            "type": "ParOverILU0",
            "relaxation": "1.0"
        },
        "coarsesolver": {
            "tol": "1e-1",
            "maxiter": "1",
            "verbosity": "0",
            "solver": "loopsolver",
            "preconditioner": {
                "type": "amg",
                "smoother": "ILU0",
                "maxlevel": "5",
                "coarsenTarget": "1000",
                "alpha": "0.2",
                "beta": "0.0001",
                "verbosity": "0",
                "relaxation": "1.0",
                "iterations": "1"
            }
        },
        "verbosity": "0"
    }
}
```

atgeirr:
I have started looking at this, so far I get a local test failure. I'll run jenkins to be sure.

atgeirr:
jenkins build this please

blattms:
I experienced two failed tests (MSW), too. But that did not make sense as the changes
of this PR seem not to be executed by our tests at all.

opm/simulators/linalg/FlexibleSolver.hpp
```diff
@@ -46,16 +46,27 @@ class FlexibleSolver : public Dune::InverseOperator<VectorTypeT, VectorTypeT>
...
     {
-        init(prm, matrix, Dune::Amg::SequentialInformation());
+        init(prm, matrix, weightsCalculator, Dune::Amg::SequentialInformation());
     }
 
     /// Create a parallel solver (if Comm is e.g. OwnerOverlapCommunication).
     template <class Comm>
-    FlexibleSolver(const boost::property_tree::ptree& prm, const MatrixType& matrix, const Comm& comm)
+    FlexibleSolver(const boost::property_tree::ptree& prm,
+                   const typename std::enable_if<std::is_function<Comm>::value,MatrixType>::type& matrix,
```
atgeirr: I think this is too complicated (I did understand it after a while). I realize that we need to distinguish the third argument being a weight calculator vs. it being a communication object. But this is wrong is it not? On line 61, you should pass comm as the third argument here? What about putting the Comm argument first, for the parallel constructor? Then there would be no ambiguity and no need for an extra constructor.

I think encapsulating the optional weights calculator in a std::function object is a good improvement (it is a bit of a pain in any case).

opm/simulators/linalg/FlexibleSolver.hpp
```diff
@@ -89,33 +100,41 @@ class FlexibleSolver : public Dune::InverseOperator<VectorTypeT, VectorTypeT>
...
         scalarproduct_ = std::make_shared<Dune::SeqScalarProduct<VectorType>>();
     }
 
     void initSolver(const boost::property_tree::ptree& prm)
     {
-        const double tol = prm.get<double>("tol");
-        const int maxiter = prm.get<int>("maxiter");
-        const int verbosity = prm.get<int>("verbosity");
-        const std::string solver_type = prm.get<std::string>("solver");
+        const double tol = prm.get<double>("tol", 1e-2);
```
atgeirr: I am a bit doubtful about the defaulting here. I think we should make life easy for users, but if we allow defaulting we must somehow print the actual options applied (default or not) in the log. When we add printing the parameter tree to the log, that would miss out on these defaults.

If this is motivated by the weird behaviour (timestep chop death) currently caused by errors in the json file, I think that we should rather be cautious here and not allow defaults, but fix that weirdness to print proper errors.

opm/simulators/linalg/ISTLSolverEbosFlexible.hpp
```diff
@@ -136,13 +153,49 @@ class ISTLSolverEbosFlexible
             // Never recreate solver.
         }
 
+        std::function<VectorType()> weightsCalculator;
+
+        auto preconditionerType = prm_.get("preconditioner.type", "cpr");
+        if( preconditionerType  == "cpr" ||
```
atgeirr: Nitpick (applies to the whole PR): there are lots of minor formatting issues here. I prefer that you clang-format them away (since they are in new code), but we can do it separately later.

opm/simulators/linalg/OwningTwoLevelPreconditioner.hpp
```diff
@@ -69,51 +86,61 @@ class OwningTwoLevelPreconditioner : public Dune::PreconditionerWithUpdate<Vecto
...
                            coarseSolverPolicy_,
-                           transpose ? 1 : 0,
-                           transpose ? 0 : 1)
+                           prm.get<int>("pre_smooth", transpose? 1 : 0),
+                           prm.get<int>("post_smooth", transpose? 0 : 1))
         , prm_(prm)
     {
-        if (prm.get<int>("verbosity") > 10) {
-            std::ofstream outfile(prm.get<std::string>("weights_filename"));
+        //Opm::Amg::getQuasiImpesWeights<MatrixType, VectorType>(
```
atgeirr: Delete unused.

opm/simulators/linalg/PreconditionerFactory.hpp
```diff
@@ -111,13 +125,15 @@ class PreconditionerFactory
...
-        criterion.setAlpha(prm.get<double>("alpha"));
-        criterion.setBeta(prm.get<double>("beta"));
-        criterion.setMaxLevel(prm.get<int>("maxlevel"));
-        criterion.setSkipIsolated(false);
-        criterion.setDebugLevel(prm.get<int>("verbosity"));
+        criterion.setAlpha(prm.get<double>("alpha", 0.33));
+        criterion.setBeta(prm.get<double>("beta", 1e-5));
+        criterion.setMaxLevel(prm.get<int>("maxlevel", 15));
+        criterion.setSkipIsolated(prm.get<bool>("skip_isolated", false));
+	criterion.setNoPreSmoothSteps(prm.get<int>("pre_smooth", 1));
```
atgeirr: Tabs should be replaced in this file.

atgeirr:
A few more notes:
 - The test failure is present on Jenkins as well. I did not look into the details.
 - I made a branch 'add-prm-printout' where I add the requested printout, but since that branch is based on current master + this branch, there is no clean way to make a PR. Do you want to cherry-pick, or should I try to rebase on top of this to make a PR? If you cherry-pick, there are two commits, the second is very small and just silences a warning.

opm/simulators/linalg/getQuasiImpesWeights.hpp
```diff
@@ -83,6 +83,47 @@ namespace Amg
         return weights;
     }
 
+    template<class Vector, class GridView, class ElementContext, class Model>
+    void getTrueImpesWeights(int pressureVarIndex, Vector& weights, const GridView& gridView,
+                             ElementContext& elemCtx, const Model& model, std::size_t threadId)
+    {
+        using VectorBlockType = typename Vector::block_type;
+        using Matrix = typename std::decay<decltype(model.linearizer().jacobian())>::type;
```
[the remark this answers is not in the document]
bska: Actually, `decay` is hardly ever the right choice. That function converts l-values to r-values and also plays tricks with arrays and pointers. Unless you're absolutely sure you need all of the power of `decay`, it is usually better/safer to use
```c++
remove_cv_t<remove_reference_t<Type>>
```
instead.

opm/simulators/linalg/FlexibleSolver.hpp
blattms: On line 61 `comm` is the fourth argument. Compiler that yes and seems to be right most of the time (which I sometimes regret). 
The thought of putting comm first, did not occur to be. Probably, because I am expecting optional arguments last. And it would result in changes at many places.

Would
```

template<class C>
struct IsComm : std::false_type
{};

template<>
struct IsComm<Dune::Amg::SequentialInformation> : std::true_type
{};

template<class Index>
struct IsComm<Dune::OwnerOverlapCopyCommunication<Index>> : std::true_type
{};
template <class Comm>
FlexibleSolver(const boost::property_tree::ptree& prm,
                       const MatrixType& matrix,
                       const typename std::enable_if<IsComm<Comm>::value, Comm>::type& comm)
```
be better? I will add that.

opm/simulators/linalg/FlexibleSolver.hpp
blattms: I do not agree at all. These json files are so complicated that even the authors of the FlexibleSolvers sometimes do not phrase them correctly. The options are bound to change (e.g. new ones added). For the user writing a self contained json file is an immense error. Let us face it, most of the options that we have will never be changed (and probably should not). Not having defaults simply makes it rather unusable.

It took me a lot of willpower to continue working on this when I experienced the time step chopping. If I would have been a nornal user I would just have abandoned it and as a new user maybe OPM, too. 

When we print the parameter files, we simply set up the defaults at the same time and warn for unused options. But that is for another PR and hopefully different developer. We should not trade usability for preventing developer errors...

opm/simulators/linalg/FlowLinearSolverParameters.hpp
```diff
@@ -220,6 +224,7 @@ namespace Opm
             EWOMS_REGISTER_PARAM(TypeTag, int, CprMaxEllIter, "MaxIterations of the elliptic pressure part of the cpr solver");
             EWOMS_REGISTER_PARAM(TypeTag, int, CprEllSolvetype, "Solver type of elliptic pressure solve (0: bicgstab, 1: cg, 2: only amg preconditioner)");
             EWOMS_REGISTER_PARAM(TypeTag, int, CprReuseSetup, "Reuse Amg Setup");
+            EWOMS_REGISTER_PARAM(TypeTag, std::string, LinearSolverConfiguration, "Configuration of solver valid is: ilu0 (default), cpr_quasiimpes, cpr_trueimpes or file (given by LinserSolverConfigurationJsonFile) ");
```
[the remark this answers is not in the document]
blattms: Will fix

opm/simulators/linalg/ISTLSolverEbosFlexible.hpp
blattms: Someone else should to that later. Removing tabs from the original PR was already quite exausting...

And beware it might change lot of places not touched here and prevent future cherry-picks, reversions, etc.

opm/simulators/linalg/OwningTwoLevelPreconditioner.hpp
blattms: Probably leftover from the original PR. Will remove

opm/simulators/linalg/PreconditionerFactory.hpp
blattms: There you go. Hours spend for tab removal and then I introduce some more.... Will fix.

opm/simulators/linalg/getQuasiImpesWeights.hpp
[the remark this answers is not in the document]
blattms: Will fix

opm/simulators/linalg/setupPropertyTree.cpp
```diff
@@ -36,12 +36,52 @@ boost::property_tree::ptree
 setupPropertyTree(const FlowLinearSolverParameters& p)
 {
     boost::property_tree::ptree prm;
+    if (p.linear_solver_configuration_ == "file") {
```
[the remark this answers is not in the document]
blattms: Fixed the space. I think the rest does not apply anymore. I am against pointing to the json file option. This property should only be used by power users who know what they are doing. I am certainly not one of those,

opm/simulators/linalg/ISTLSolverEbosFlexible.hpp
atgeirr: Ok, note that my suggestion is not to clang-format the files wholesale, but just the new parts (not sure what editor you are using, but I have set up emacs so that `M-x clang-format` modifies the current selection, very useful).

atgeirr:
I think the IsComm solution is fine, and a lot easier to read than the current code. (Although I am slightly in favour of comm-first instead.)

I think the file option really should be mentioned in the option doc and error messages. If we are afraid of users trying and failing, we could write "... file (experimental)" or similar.

About the defaults: If the file option is really only for power users who know what they are doing, then I think it is not a good idea to give them lots of defaults that increase the risk of error. The defaults are only going to come into play (in the case of Flow at least) when combined with the file option, so it is not really giving regular users any benefit?

The "terminate, not chop" behaviour we want to have also needs to be implemented. I can take a shot at it. Did you look at the 'add-prm-printout' branch btw?

opm/simulators/linalg/setupPropertyTree_impl.hpp
```diff
@@ -0,0 +1,112 @@
...
+/// and preconditioner. If the latter, the parameters --linear-solver-reduction,
+/// --linear-solver-maxiter and --linear-solver-verbosity are used, but if reading
+/// from file the data in the JSON file will override any other options.
+template<class TypeTag>
+boost::property_tree::ptree
+setupPropertyTree(const FlowLinearSolverParameters& p)
+{
+    boost::property_tree::ptree prm;
+    if (p.linear_solver_configuration_ == "file") {
+#if BOOST_VERSION / 100 % 1000 > 48
```
atgeirr: Logic still looks wrong here, and error messages should be adjusted. Is there something I don't see?

alfbr:
> About the defaults: If the file option is really only for power users who know what they are doing, then I think it is not a good idea to give them lots of defaults that increase the risk of error. The defaults are only going to come into play (in the case of Flow at least) when combined with the file option, so it is not really giving regular users any benefit?

I probably misunderstand something here, sharing my five cents anyway. On the json file, I really agree with @blattms , I would even consider preventing users from accessing it, as it will do more harm than good. Those power users do not exist, I have at least never seen one. I am afraid it is only for developers or those who want to do research on improving defaults. If we are to put the CPR option in use, we need defaults that works well across a range of problems.

opm/simulators/linalg/setupPropertyTree_impl.hpp
blattms: Now i get it.

blattms: The logic is correct, but the message is wrong.

atgeirr: By "the logic" I mean what happens when the BOOST_VERSION is too low (i.e. the above is false). Then we have
```
    if (p.linear_solver_configuration_ == "file") {
    }
    else
    {
        std::string conf =  p.linear_solver_configuration_;
        ...
```

blattms:
The add_prm_printout is not what we need.
- We need a complete list of all possible parameter names with default values,  Then we need to check whether all parameters passes are in that list and otherwise error out (that is one thing that you wanted and this is the only way to accomplish that).
- I might want to warn about unused ones. 
- As you do not want defaults without barking and do not agree with where my defaults are set, you need to set them here. But only those that are really needed if you want to warn about missing ones.

This has become quite a task with a lot of replicated logic, don't you agree? Seems quite unmaintainable.

atgeirr:
> I probably misunderstand something here, sharing my five cents anyway. On the json file, I really agree with @blattms , I would even consider preventing users from accessing it, as it will do more harm than good. Those power users do not exist, I have at least never seen one. I am afraid it is only for developers or those who want to do research on improving defaults. If we are to put the CPR option in use, we need defaults that works well across a range of problems.

Preventing use of a JSON file would prevent lots of useful experimentation, which is part of what this aimed to accomplish in the first place. I am going to assume that option is for the specialist, and anything useful discovered using it should make its way into more easily accessible options. This PR offers three such options: "ilu0", "cpr_trueimpes" and "cpr_quasiimpes", which should (for now) fulfil user needs well. The discussed defaults do not apply to these "pre-built" options, but only to the case when the user specifies a JSON file. They are therefore in no way affecting (or helping) the regular user.

atgeirr:
> The add_prm_printout is not what we need.
> 
> * We need a complete list of all possible parameter names with default values,  Then we need to check whether all parameters passes are in that list and otherwise error out (that is one thing that you wanted and this is the only way to accomplish that).
> * I might want to warn about unused ones.
> * As you do not want defaults without barking and do not agree with where my defaults are set, you need to set them here. But only those that are really needed if you want to warn about missing ones.
> 
> This has become quite a task with a lot of replicated logic, don't you agree? Seems quite unmaintainable.

It was not intended to be a sufficient solution, only a necessary one. If we do not have defaults, only the second point remains (warn about unused parameters). We had this feature in the old ParameterGroup code, and for sure we could resurrect it, or a variant of it. It is really useful to be able to abort if the user mis-spelled an option. Without defaults, it is not so serious, because if the user mis-spells an option there is no default to make it pass silently.

If you want defaults (to be applied on top of the JSON stuff), there is no way to provide a complete list of all possible keys and default values, since the hierarchy depth is arbitrary at run-time using multi-level preconditioners. I see no way to check this without simply trying to build the solver, and catching the exception on failure. I think that is a reasonable approach. (We could make a separate paramCheck()-type function for everything, but that would only accomplish the same and add extra effort and complexity.)

blattms:
jenkins build this please

blattms:
One testis fixed now. The two still failing on my system are beyond me. They are compareECLFiles_flow+3D_MSW and compareECLFiles_flow+TEST_WSEGSICD.

alfbr:
I guess you could call that an improvement :-) The latest work on tightening up tests left one failing locally. If you have only two failing, it seems we did make some headway. Hope to continue that work soon, so that all integration tests succeed for all developers locally.

blattms:
Well, one of your arguments for not having defaults was to catch typos...

atgeirr:
> Well, one of your arguments for not having defaults was to catch typos...

Yes that is an argument against defaults. With defaults it is much harder. Perhaps we are talking around each other?

blattms:
> The discussed defaults do not apply to these "pre-built" options, but only to the case when the user specifies a JSON file. They are therefore in no way affecting (or helping) the regular user.

I beg to differ. I spent hours testing and trying to make these default reflect the improvements that my mind might have tricked me to hear, previoously. In this work I tried some of the proposed json files in previous PRs and got convergence issues that a non-developer could never have figured out (and I only did because I had to). With sane defaults that would not have happened.

I am in no way arguing against having these properties available. But they need to usable for people besides the authors. Please also be aware that there might still be properties missing. Property files that worked previously need to work in the future IMHO, even if new ones are added.

atgeirr:
> Property files that worked previously need to work in the future IMHO, even if new ones are added.

I'll accept this argument, we'll have defaults! Now how to make it robust, or even to print them?

blattms:
the same two tests fail for master.

atgeirr:
> the same two tests fail for master.

I assume you mean locally? The current test failure is because the new _impl.hpp file is not found somehow.

blattms:
Yep, locally. Did not test installation there.

blattms:
jenkins build this please

opm/simulators/linalg/setupPropertyTree_impl.hpp
blattms: We aree now throwing in this case.

blattms:
> > Property files that worked previously need to work in the future IMHO, even if new ones are added.
> 
> I'll accept this argument, we'll have defaults! Now how to make it robust, or even to print them?

I did not say it is easy. Printing only the valid/used ones without defaults is not easy either.

blattms:
Shall we merge?

atgeirr:
> Shall we merge? Everything besides printing should be done.

Ok. I guess I'll push my tiny printing extension as a separate PR. Then we'll need to figure out something better later.

blattms:
Sounds great.

opm/simulators/linalg/FlexibleSolver.hpp
```diff
@@ -33,6 +33,17 @@
...
+struct IsComm : std::false_type
+{};
+
+template<>
+struct IsComm<Dune::Amg::SequentialInformation> : std::true_type
+{};
+
+template<class Index>
+struct IsComm<Dune::OwnerOverlapCopyCommunication<Index>> : std::true_type
+{};
```
bska, line 46: I'm probably missing something simple, but this construction now generates a build failure for me in a sequential/debug/dune-2.6.0 configuration:
```
In file included from /home/bska/work/opm/src/opm/opm-simulators/tests/test_flexiblesolver.cpp:31:0:
/home/bska/work/opm/src/opm/opm-simulators/opm/simulators/linalg/FlexibleSolver.hpp:45:21: error: ‘OwnerOverlapCopyCommunication’ is not a member of ‘Dune’
 struct IsComm<Dune::OwnerOverlapCopyCommunication<Index>> : std::true_type
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/bska/work/opm/src/opm/opm-simulators/opm/simulators/linalg/FlexibleSolver.hpp:45:21: note: suggested alternative: ‘OwnerOverlapCopyAttributeSet’
 struct IsComm<Dune::OwnerOverlapCopyCommunication<Index>> : std::true_type
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                     OwnerOverlapCopyAttributeSet
/home/bska/work/opm/src/opm/opm-simulators/opm/simulators/linalg/FlexibleSolver.hpp:45:51: error: template argument 1 is invalid
 struct IsComm<Dune::OwnerOverlapCopyCommunication<Index>> : std::true_type
                                                   ^~~~~
/home/bska/work/opm/src/opm/opm-simulators/opm/simulators/linalg/FlexibleSolver.hpp:45:56: error: expected unqualified-id before ‘>’ token
 struct IsComm<Dune::OwnerOverlapCopyCommunication<Index>> : std::true_type
                                                        ^~
/home/bska/work/opm/src/opm/opm-simulators/opm/simulators/linalg/FlexibleSolver.hpp:46:3: warning: extra ‘;’ [-Wpedantic]
 {};
   ^
In file included from /home/bska/work/opm/inst/dune/2.6.0/dbg/seq/normal/include/dune/istl/paamg/construction.hh:11:0,
```
Please let me know if there's any additional information I can provide.

blattms: Soory. See #2515 for a guessed.

blattms: Undermines my impression that compilers a normally right (and not me).

bska: > See #2515 for a guessed fix.

Thanks.  PR #2525 does indeed restore my sequential build.  I appreciate the quick turn-around.

PR #2500, opened 2020-03-25
WIP: Remove internal variables

joakim-hove:
Drive by improvement which went a bit overboard. I know this is work in progress, but I really think we should merge something in this direction - it has &part;<sub>t</sub> S < 0

What I have started on here is essentially to remove the notion of an *internal* deck in `EclVanguard` class - instead the assembly of the input is completed in `flow.cpp` and then the `flowXXXXSetDeck()`function *must* be called.

If this is to fly the parsing must be repeated in all the simulators like `flow_ebos_gin_tonic` - I suggest that we use a simplified parsing for all these alternative simulators - like I have started on in `flow_ebos_foam.cpp`

### DIFF
--- a/flow/flow_ebos_foam.cpp
+++ b/flow/flow_ebos_foam.cpp
@@ -36,13 +36,12 @@ SET_BOOL_PROP(EclFlowFoamProblem, EnableFoam, true);
 }}
 
 namespace Opm {
-void flowEbosFoamSetDeck(double setupTime, Deck *deck, EclipseState& eclState, Schedule& schedule, SummaryConfig& summaryConfig)
+void flowEbosFoamSetDeck(double setupTime, EclipseState& eclState, Schedule& schedule, SummaryConfig& summaryConfig)
 {
     typedef TTAG(EclFlowFoamProblem) TypeTag;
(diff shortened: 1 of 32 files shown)


flow/flow_ebos_foam.cpp
```diff
@@ -64,6 +69,13 @@ int flowEbosFoamMain(int argc, char** argv, bool outputCout, bool outputFiles)
 #endif
 
     Opm::FlowMainEbos<TTAG(EclFlowFoamProblem)> mainfunc;
+    Opm::Parser parser;
+    const auto deck = parser.parseFile( argv[1] );
+    auto ecl_state = Opm::EclipseState(deck);
+    auto schedule  = Opm::Schedule(deck, ecl_state);
+    auto summary_config = Opm::SummaryConfig(deck, schedule, ecl_state.getTableManager());
```
akva2: this would completely kills parallel. see flow.cpp.

joakim-hove: I had misunderstood something here - removing again.

bska:
Closing this now since similar effects have been achieved through other means.

PR #2501, opened 2020-03-25
Add Python argument to Schedule constructor

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/1636

The new PYACTION keyword allows for loading a Python module and then run the `run()` function in that module during the simulation. The module is internalized by the Python interpreter on parse time (from opm-common) - that is essential to enable early error detection.If we do not have Python enabled the Python keywords are ignored in the Schedule construction.

The `python`argument which is added to the `Schedule`constructor is an instance of [this](https://github.com/OPM/opm-common/blob/master/src/opm/parser/eclipse/Python/Python.cpp). Iff this version of opm-common has been compiled with `OPM_ENABLE_EMBEDDED_PYTHON` the `Python`instance holds on to an interpreter - otherwise it is just an empty shell.

Will probably make a constructor argument to the `Python` class so you can can control runtime whether to support Python or not (*iff* it has been enabled compile time).

For now this PR has no consequences whatsover - apart from a warning about an unused parameter in the `Schedule`constructor.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -35,6 +35,7 @@
 #include <opm/grid/cpgrid/GridHelpers.hpp>
 #include <opm/core/props/satfunc/RelpermDiagnostics.hpp>
 
+#include <opm/parser/eclipse/Python/Python.hpp>
 #include <opm/parser/eclipse/Parser/Parser.hpp>
 #include <opm/parser/eclipse/Parser/ParseContext.hpp>
 #include <opm/parser/eclipse/Parser/ErrorGuard.hpp>
@@ -348,7 +349,7 @@ public:
             // create the schedule object. Note that if eclState is supposed to represent
             // the internalized version of the deck, this constitutes a layering
             // violation.
-            internalEclSchedule_.reset(new Opm::Schedule(*deck_, *eclState_, *parseContext_, *errorGuard_));
+            internalEclSchedule_.reset(new Opm::Schedule(*deck_, *eclState_, *parseContext_, *errorGuard_, python));
             eclSchedule_ = internalEclSchedule_.get();
         }
         else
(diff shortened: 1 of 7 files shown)


joakim-hove:
This is the first attempt at making the Python plugin available from flow. Opinions - questions?

ebos/eclbasevanguard.hh
```diff
@@ -629,6 +630,7 @@ private:
     Opm::EclipseState* eclState_;
     Opm::Schedule* eclSchedule_;
     Opm::SummaryConfig* eclSummaryConfig_;
+    Opm::Python python;
```
atgeirr: Is it ok to construct Python twice? I was not able to understand the table in Python.hpp... (Also, in that file, I think the doc has a typo in the example on line 53, I think "with" should be "without".)

Assuming constructing twice is fine, this PR is fine as a whole.

joakim-hove: > Assuming constructing twice is fine, this PR is fine as a whole.

It is safe to do so - but the best is to only construct one instance. The best solution would in my opinion be to get rid of the Internal variables in `eclbasevanguard`- in a fist of hubris I started on that in #2500 - but I don't know if life is long enough to follow that path.

Cool if you comment on the table in the upstream PR - if that was not useful it should be updated.

atgeirr: > It is safe to do so - but the best is to only construct one instance. The best solution would in my opinion be to get rid of the Internal variables in `eclbasevanguard`- in a fist of hubris I started on that in #2500 - but I don't know if life is long enough to follow that path.

I agree. Also with the "maybe not worth it given average life expectancy" part...



> Cool if you comment on the table in the upstream PR - if that was not useful it should be updated.

I added a comment about what confused me.

PR #2504, opened 2020-03-30
fixed: broadcasting an int and receiving as a size_t is a very bad idea

akva2:
pointy-hat: me

### DIFF
--- a/ebos/eclmpiserializer.hh
+++ b/ebos/eclmpiserializer.hh
@@ -179,7 +179,8 @@ public:
 
         if (m_comm.rank() == 0) {
             pack(data);
-            m_comm.broadcast(&m_position, 1, 0);
+            m_packSize = m_position;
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

bska:
Is this the reason that the [nightly-bionic-docker](https://ci.opm-project.org/job/opm-nightly-bionic-docker/) build has been failing lately?

akva2:
very likely.

akva2:
shameful self-merge.

bska:
>> Is this the reason that the nightly-bionic-docker build has been failing lately?
>
> very likely.

Very good.  In any case, this is obviously an improvement, so even if it does not end up fixing the nightly build it should go into master.

> shameful self-merge.

If you didn't merge it yourself I would have.

PR #2506, opened 2020-03-31
Add build time to flow startup banner

joakim-hove:


joakim-hove:
jenkins build this please

blattms:
Please enlighten me. When and how can knowledge of the build time help me?

alfbr:
> Please enlighten me. When and how can knowledge of the build time help me?

It gives an upper time limit to how new the code is I guess.  But don't we already print out the latest commit hash of opm-simulators?

joakim-hove:
> When and how can knowledge of the build time help me?

When debugging an existing simulation - I get a good guesstimate of whether feature 'X' is included.

> But don't we already print out the latest commit hash of opm-simulators?

Yes we do - in many cases that will serve the purpose; but when someone has assembled their own version developer Joe might have that commit at his place.

atgeirr:
I agree that more information is good! But what is BUILD_TIMESTAMP? My google-fu is insufficient on this one, and a tiny "hello build_timestamp" test locally failed to see it?

atgeirr:
> But don't we already print out the latest commit hash of opm-simulators?

We do for release builds, but not for debug builds. I guess the idea is that debug builds are often in flux anyway, with uncommitted minor changes under testing.

akva2:
i'm going to flip the question: what harm would having compile time do?

joakim-hove:
> But what is BUILD_TIMESTAMP?

It is a string generated by the build system: https://github.com/OPM/opm-common/pull/1653 - strictly speaking "CMAKE_CONFIGURE_TIMESTAMP" would be more correct - that could of course be improved.

atgeirr:
> It is a string generated by the build system

Thanks, that explains why it failed standalone...

You should `#include "project-version.h"` though, right?

alfbr:
>i'm going to flip the question: what harm would having compile time do?

Takes up some space in the banner I guess, giving ambiguous information. Personally, I believe it is useful, so I support this. Only developers knows how to track commit hashes, build time is easy for most people to relate to. Hence, they both serve the same purpose, but for different audiences.

joakim-hove:
> You should #include "project-version.h" though, right?

No - the `compileTimestamp()` is a naked function in `moduleVersion.hpp`; which includes `project-version.h`. But you must regenerate the header file based on a grand new opm-common.

blattms:
I guess your workflow is different from mine and if it makes sense there then just go for it. It does not harm at all. I was just trying to understand the need.

For me it does not help if compile time really is when I compiled/cmaked stuff. I usually have multiple checkouts and for me the time I compiled each of them does not mean much as they are on different branches.

joakim-hove:
> For me it does not help if compile time really is when I compiled/cmaked stuff.

This is not about *my* compile time - that I can manage; it is when get a simulation from someone where something has broken.

atgeirr:
All seems ok here, merging.

PR #2510, opened 2020-03-31
[bugfix][ISTLSolverEbos] fix compilation when dune-fem is used in

dr-robertk:
when dune-fem is used in conjunction with opm-simulators.

### DIFF
--- a/opm/simulators/linalg/ISTLSolverEbos.hpp
+++ b/opm/simulators/linalg/ISTLSolverEbos.hpp
@@ -217,7 +217,7 @@ public:
         return overlapping ?
             Dune::SolverCategory::overlapping : Dune::SolverCategory::sequential;
     }
-    
+
     //! constructor: just store a reference to a matrix
     WellModelGhostLastMatrixAdapter (const M& A,
                                      const M& A_for_precond,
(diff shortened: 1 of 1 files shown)


dr-robertk:
jenkins build this please

dr-robertk:
@totto82: Could you merge when the test succeeded?

blattms:
I think this is the wrong fix. One should rather make the vanguards adhere to the same interface (i.e. add a correct gridView() method there).  Otherwise this bound to break again.

A test with another grid would be helpful, too.

dr-robertk:
It's the right fix, trust me. When dune-fem is used then the GridView defined in vanguard is not the same as grid().leafGridView(), that is why the compilation fails. So when an ElementMapper with GridView from vanguard is defined, then gridView from vanguard  should be used and not grid().leafGridView(). It's quite simple. Otherwise Grid::LeafGridView should be used. But then the element numbering might differ. So long story short: It's the right fix!

blattms:
Finally got it. Would probably have made sense to put that description in the commit message.

totto82:
Thanks. Merging.

PR #2512, opened 2020-04-01
Enable owner first ordering by default

alfbr:
I have tested the owner-first ordering on a number of cases and a number of process counts. It has consistently yielded better performance. Hence, I suggest we enable it  by default.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -88,7 +88,7 @@ SET_BOOL_PROP(EclBaseVanguard, EnableOpmRstFile, false);
 SET_BOOL_PROP(EclBaseVanguard, EclStrictParsing, false);
 SET_BOOL_PROP(EclBaseVanguard, SchedRestart, true);
 SET_INT_PROP(EclBaseVanguard, EdgeWeightsMethod, 1);
-SET_BOOL_PROP(EclBaseVanguard, OwnerCellsFirst, false);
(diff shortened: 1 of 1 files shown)


alfbr:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 0.988    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.027    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.044    |



View result details @ https://www.ytelses.com/opm/?page=result&id=700

blattms:
Mmh,  ytelses does not feel that improvement.

atgeirr:
> Mmh, ytelses does not feel that improvement.

2.7% and 4.4% is still well worth it, I think?

atgeirr:
jenkins build this please

alfbr:
Yes, it does. No difference on single process runs as expected. A 2.7% and 4.4% estimated improvement on eight processes. From my testing the improvement increases further at sixteen processes.

hnil:
I think this also is a much safer option for dune-istl, since the matrixredistribution will make such ordering (as far as I understand) This means this ordering is better tested. I also expect that the extra work done in the Andreas Thune's pullrequest is not needed if this ordering is used.

Issue #2508, opened 2020-03-31
filesystem error: cannot make absolute path: Invalid argument []


hakonhagland:
When running `flow` from the master branches:

```
$ flow SPE1CASE1.DATA
**********************************************************************
*                                                                    *
*                      This is flow 2020.04-pre                      *
*                                                                    *
* Flow is a simulator for fully implicit three-phase black-oil flow, *
*             including solvent and polymer capabilities.            *
*          For more information, see https://opm-project.org         *
*                                                                    *
**********************************************************************

Using 1 MPI processes with 2 OMP threads on each 

Reading deck file 'SPE1CASE1.DATA'
terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'
  what():  filesystem error: cannot make absolute path: Invalid argument []
Aborted (core dumped)
```
After running it with gdb, it seems the error is on [line #177](https://github.com/OPM/opm-simulators/blob/202235abe0239276536f6be5f1f38b87c66ed098/flow/flow.cpp#L177) in `flow.cpp`:
```C++
output_dir = absolute(path(baseName).parent_path()).string();
```
here `baseName` is `SPE1CASE1.DATA` and `path(baseName).parent_path()` returns the empty string `""`, finally running `filesystem::path::absolute` on an empty string :
```C++
absolute(path(baseName).parent_path())
```
gives the error.

goncalvesmachadoc:
We have encountered same issue when trying to compile/install flow in the cluster (Cent OS 6):
![OPM_run_fail](https://user-images.githubusercontent.com/54812829/78032759-ca3b5a80-7365-11ea-9721-e1a9b2a3da83.PNG)

blattms:
Are we still supporting RedHat/CentOS 6 in master? I thought we dropped them because of limited c++17 compiler support.

blattms:
But I might be wrong. What compiler is it?

blattms:
Seems like I am wrong. Commit messages indicate that redhat6 should still work. But one needs devtoolset-8.

hakonhagland:
I am using Ubuntu 19.10, gcc version 9.2.1.

joakim-hove:
> flow SPE1CASE1.DATA

Should of course not be necessary, but have have you tried: `flow ./SPE1CASE1.DATA`?

hakonhagland:
@joakim-hove Yes, using `flow ./SPE1CASE1.DATA` gives the same error..

[...]

PR #2514, opened 2020-04-02
changed: avoid calling parent_path on paths that has none

akva2:
causes a throw on gcc-9.

closes https://github.com/OPM/opm-simulators/issues/2508

### DIFF
--- a/flow/flow.cpp
+++ b/flow/flow.cpp
@@ -174,7 +174,9 @@ FileOutputMode setupLogging(int mpi_rank_, const std::string& deck_filename, con
 
     std::string output_dir = cmdline_output_dir;
     if (output_dir.empty()) {
-        output_dir = absolute(path(baseName).parent_path()).string();
+        output_dir = fpath.has_parent_path()
+                   ? absolute(fpath.parent_path()).generic_string()
+                   : Opm::filesystem::current_path().generic_string();
     }
 
     logFileStream << output_dir << "/" << baseName;


bska:
I'm going to abuse this PR to see if I can reproduce my [local symptom](https://github.com/OPM/opm-simulators/pull/2498#discussion_r402360396) on Jenkins too.

bska:

bska:
jenkins build this please

flow/flow.cpp
```diff
@@ -174,7 +174,10 @@ FileOutputMode setupLogging(int mpi_rank_, const std::string& deck_filename, con
 
     std::string output_dir = cmdline_output_dir;
     if (output_dir.empty()) {
-        output_dir = absolute(path(baseName).parent_path()).string();
+        if (path(fpath).has_parent_path())
+          output_dir = absolute(path(fpath).parent_path()).string();
+        else
+          output_dir = Opm::filesystem::current_path().string();
```
bska, line 180: Tiny nit: `fpath` already is a `path` object by construction.  You don't need to construct another `path` object here.  On another note, I think I would probably use the ternary operator:
```C++
output_dir = fpath.has_parent()
    ? absolute(fpath.parent_path()).string()
    : Opm::filesystem::current_path().string();
```
just to underscore the fact that we're assigning a value to `output_dir` in both cases.  I might also consider using `generic_string()` instead of just `.string()`.

akva2: all your nits are belong to you.

bska:
jenkins build this please

PR #2516, opened 2020-04-02
Refactor main() in flow.cpp.

hakonhagland:
A refactoring of `main()` in [`flow.cpp`](https://github.com/OPM/opm-simulators/blob/master/flow/flow.cpp) to address the [comments](https://github.com/OPM/opm-simulators/pull/2127#issuecomment-596529720) on PR #2127 and as preparation for implementing the Python bindings described in PR #2403.

PR #2518, opened 2020-04-02
Python bindings for the blackoil simulator.

hakonhagland:
Python bindings for the blackoil simulator.

NOTE: This PR depends on PR #2516 which should be merged first.

A rewrite of the Python bindings for the blackoil simulator using pybind11 as introduced in PR #2127. The new version uses the refactored `flow.cpp` introduced in PR #2516 and thus avoids duplication of the code in `simulators.cpp`.

This PR will be the starting point for implementing the Python bindings introduced in PR #2403.

### DIFF
--- a/python/simulators/CMakeLists.txt
+++ b/python/simulators/CMakeLists.txt
@@ -1,0 +1,21 @@
+pybind11_add_module(simulators simulators.cpp)
+
+set_target_properties( simulators PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/python/opm2 )
+
+target_sources(simulators
+  PRIVATE
+    ../../flow/flow_ebos_blackoil.cpp
+    ../../flow/flow_ebos_brine.cpp
+    ../../flow/flow_ebos_energy.cpp
+    ../../flow/flow_ebos_foam.cpp
+    ../../flow/flow_ebos_gasoil.cpp
+    ../../flow/flow_ebos_oilwater.cpp
+    ../../flow/flow_ebos_oilwater_polymer.cpp
+    ../../flow/flow_ebos_oilwater_polymer_injectivity.cpp
+    ../../flow/flow_ebos_polymer.cpp
+    ../../flow/flow_ebos_solvent.cpp
+    ../../flow/flow_ebos_blackoil.cpp)
+
+target_link_libraries( simulators PRIVATE opmsimulators )
+
+install(TARGETS simulators DESTINATION ${PYTHON_INSTALL_PREFIX}/simulators)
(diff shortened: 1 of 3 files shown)


python/simulators/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+pybind11_add_module(simulators simulators.cpp)
+
+set_target_properties( simulators PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/python/opm2 )
```
hakonhagland, line 3: Note: I that installed the Python module in `opm2` and not `opm` since there is a conflict with the `opm` module in `opm-common`. If I had called it `opm`, I could not have imported modules from both `opm` in `opm-common` and `opm` in `opm-simulators` in the same Python script

python/simulators/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+
+target_sources(simulators
+  PRIVATE
+    ../../flow/flow_ebos_blackoil.cpp
+    ../../flow/flow_ebos_brine.cpp
```
hakonhagland, line 8: Note: It might be desirable to remove the dependencies on the other source files other than that of the blackoil simulator `flow_ebos_blackoil.cpp` . This will reduce compile time for the Python module, and also make the library smaller. Currently, the Python bindings does not use any of  the sources: `flow_ebos_brine.cpp`, `flow_ebos_energy.cpp`, `flow_ebos_foam.cpp`, `flow_ebos_gasoil.cpp`, `flow_ebos_oilwater.cpp`, `flow_ebos_oilwater_polymer.cpp`, `flow_ebos_oilwater_polymer_injectivity.cpp`, `flow_ebos_polymer.cpp`, `flow_ebos_solvent.cpp`.. But we still need to compile them into the Python library since they are mentioned in the new `Main.hpp`, see PR #2516

hakonhagland:
Updated this PR according to the changes in PR #2516

python/simulators/CMakeLists.txt
atgeirr: In order to do that, I think you may have to define FLOW_BLACKOIL_ONLY in simulators.cpp, otherwise you will fail to link to the other variants.

The C++ parts of this looks fine to me. I am no Python expert, someone else must decide if those parts are good.

hakonhagland: > In order to do that, I think you may have to define FLOW_BLACKOIL_ONLY in simulators.cpp,

Yes, of course! I pushed an update, but for some reason it does not show here. It is available in my fork though: https://github.com/hakonhagland/opm-simulators-1/commit/b2ec7cd9519cf337a07afcde5a25c4d2018c2ffe

hakonhagland:
Closing this PR since it was replaced by #2555.

PR #2520, opened 2020-04-03
add norne and norne_parallel as ExtraTests

akva2:
to execute these you have to use ctest -C ExtraTests
or 'make extra_test'.

As discussed in meeting. Right now these only execute the runs, meant as a POC.

Downstream of https://github.com/OPM/opm-common/pull/1678

### DIFF
--- a/tests/run-test.sh
+++ b/tests/run-test.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# This simply runs a simulator.
+
+INPUT_DATA_PATH="$1"
+RESULT_PATH="$2"
+BINPATH="$3"
+EXE_NAME="$4"
+FILENAME="$5"
+MPI_PROCS="$6"
+shift 7
+TEST_ARGS="$@"
+
+mkdir -p ${RESULT_PATH}
+if (( ${MPI_PROCS} > 1))
+then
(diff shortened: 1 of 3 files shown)


bska:
jenkins build this please

joakim-hove:
> to execute these you have to use ctest -C ExtraTests

Could we have a commandline switch without CamelCase - e.g. `ctest -c extra`?

tests/run-test.sh
```diff
@@ -0,0 +1,23 @@
...
+RESULT_PATH="$2"
+BINPATH="$3"
+EXE_NAME="$4"
+FILENAME="$5"
+MPI_PROCS="$6"
+shift 7
+TEST_ARGS="$@"
+
+mkdir -p ${RESULT_PATH}
+cd ${RESULT_PATH}
```
atgeirr, line 15: Is it not too much to cd into that dir? Using it again as output dir leads to norne/norne/ etc.

akva2: i guess so. this script is not intended for final commit as i (tried to) state above, so i wasn't paying much attention.

akva2:
> > to execute these you have to use ctest -C ExtraTests
> 
> Could we have a commandline switch without CamelCase - e.g. `ctest -c extra`?

configuration renamed.

PR #2521, opened 2020-04-03
Refactor flow.cpp

hakonhagland:
This is the first PR in a splitting up of PR #2516.

According to private discussion with @joakim-hove PR #2516 is splitted into smaller parts to facilitate review and a quicker merge into master.

As mentioned in PR #2516, `main()` in `flow.cpp` is refactored to address the comments on PR #2127 and as preparation for implementing the Python bindings described in PR #2403.


joakim-hove:
jenkins build this please

flow/flow.cpp
```diff
@@ -19,482 +19,12 @@
...
-        if (outputCout) {
-            std::cerr << "Failed to create valid EclipseState object." << std::endl;
-            std::cerr << "Exception caught: " << e.what() << std::endl;
-        }
-        return EXIT_FAILURE;
-    }
-
-    return EXIT_SUCCESS;
+    auto mainObject = Opm::Main(argc, argv);
+    return mainObject.run();
```
hakonhagland, line 28: The `main()` in `flow.cpp` is copied almost verbatim to `Main.hpp`

opm/simulators/flow/Main.hpp
```diff
@@ -0,0 +1,526 @@
...
+    public:
+        Main(int argc, char** argv) : argc_(argc), argv_(argv)  {  }
+
+        int run() {
+            return main_(argc_, argv_);
+        }
+
+    private:
+
+        Opm::filesystem::path simulationCaseName_( const std::string& casename ) {
```
hakonhagland, line 112: Removed the `details::` namespace prefix for `simulationCaseName_()` since it appears it is no longer need?

opm/simulators/flow/Main.hpp
```diff
@@ -0,0 +1,526 @@
...
+
+
+        // This function is an extreme special case, if the program has been invoked
+        // *exactly* as:
+        //
+        //    flow   --version
+        //
+        // the call is intercepted by this function which will print "flow $version"
+        // on stdout and exit(0).
+        void handleVersionCmdLine_(int argc, char** argv) {
```
hakonhagland, line 147: Also removed `details::` prefix for `handleVersionCmdLine_()`

opm/simulators/flow/Main.hpp
```diff
@@ -0,0 +1,526 @@
...
+                                           {Opm::Log::MessageType::Problem,
+                                            msgLimits.getProblemPrintLimit(0)},
+                                           {Opm::Log::MessageType::Bug,
+                                            msgLimits.getBugPrintLimit(0)}};
+            stream_log->setMessageLimiter(std::make_shared<Opm::MessageLimiter>(10, limits));
+        }
+
+
+        // ----------------- Main program -----------------
+        int main_(int argc, char** argv)
```
hakonhagland, line 276: Almost verbatim copy of the original `main()`

joakim-hove:
jenkins build this please

joakim-hove:
As agreed with @atgeirr I'll merge this now!

blattms:
This just increased the maintainance effort of (flow_tag.hpp, which was another copy of flow.cpp) quite a bit. Could we somehow unify these?

joakim-hove:
> This just increased the maintainance effort of (flow_tag.hpp, which was another copy of flow.cpp) quite a bit. 

OK - I'm sorry to hear.

> Could we somehow unify these?

I am all for it; at least for me the situation is that I don't have the surplus understanding (or energy/time for that matter) to do things Right&trade; in this part of the code base. Quite often when I do something in this part of the code I get my immediate problem solved, but I leave with an uneasy feeling that apart from solving my immediate problem I have pushed the code in wrong direction when it comes to being maintainable - I did not have this feeling now though.

I feel inspired to invite for a developer discussion on the topic after easter.

hakonhagland:
@blattms wrote:
> This just increased the maintainance effort of (flow_tag.hpp, which was another copy of flow.cpp) quite a bit. Could we somehow unify these?

What is the purpose of `flow_tag.hpp` or where is it used?

hakonhagland:
It seems it used by [`flow_blackoil_dunecpr.cpp`](https://github.com/OPM/opm-simulators/blob/master/flow/flow_blackoil_dunecpr.cpp), [`flow_onephase.cpp`](https://github.com/OPM/opm-simulators/blob/master/flow/flow_onephase.cpp), and [`flow_onephase_energy.cpp`](https://github.com/OPM/opm-simulators/blob/master/flow/flow_onephase_energy.cpp) to create a template version of `main()` called `mainFlow<TypeTag>` ? So there is duplication of code in `main()` since `main()` in `flow.cpp` is not templetized. The question is then why is not not `main()` in `flow.cpp` using `mainFlow<TypeTag>`()? Then we would could maybe avoid the duplication?

hakonhagland:
I have added two new pull requests (PR #2523, and PR #2524) to address the issue with code duplication.

PR #2523, opened 2020-04-05
Merge flow_tag.hpp into Main.hpp.

hakonhagland:
Addresses the comments in PR #2521 regarding code duplication in `Main.hpp` and `flow_tag.hpp`. This PR merges the code in `flow_tag.hpp` into `Main.hpp` such that `flow_tag.hpp` can be eliminated (which will be done in PR #2524).


opm/simulators/flow/Main.hpp
```diff
@@ -329,17 +376,22 @@ namespace Opm
...
+                    Opm::Parameters::printUsage<PreTypeTag>(PreProblem::helpPreamble(
+                            argc, const_cast<const char**>(argv)), e.what());
+#endif
 #if HAVE_MPI
                 MPI_Finalize();
 #endif
                 return 1;
             }
-
+#ifdef OPM_FLOW_MAIN
```
blattms, line 390: I think it would be an improvement fif we would always print the banner. The only reason not to do this would be compile problems.

opm/simulators/flow/Main.hpp
```diff
@@ -80,6 +93,39 @@ NEW_TYPE_TAG(FlowEarlyBird, INHERITS_FROM(EclFlowProblem));
 
 END_PROPERTIES
 
+#ifndef OPM_FLOW_MAIN
```
blattms, line 96: Shouldn't this compile regardless of the `#ifndef`? In that case could remove it here.

hakonhagland: Yes that code compiles regardless. But if we remove it, wouldn't the code then be included in the built `flow` executable which will never use the code? Adding unnecessary unused code and increasing the executable size (although very little) seems not good, right?

opm/simulators/flow/Main.hpp
hakonhagland: Agree, and there seems to be no compile problems by removing the `#ifdef`..

joakim-hove:
Please - those in the know: give this and #2524 some love and attention; reducing the complexity and repetition in this part of the codebase is a very valuable addition.

opm/simulators/flow/Main.hpp
blattms: As the method has a template parameter it will only be compiled if used.

opm/simulators/flow/Main.hpp
blattms: Ok, then let's change it?

hakonhagland: Done with https://github.com/OPM/opm-simulators/pull/2523/commits/ee84e8bc2b20d75e135d605bc61e04b7c52bddb3

opm/simulators/flow/Main.hpp
hakonhagland: Of course, I did not think about that! Updated the PR with this commit https://github.com/OPM/opm-simulators/pull/2523/commits/125b0f11bc3f74d5e8650d268ad741662b3868c6

blattms:
I apologize for my pickiness and the trouble that this has now caused:

Now there suddenly are a lot of unrelated commits here and 19 files changed, due to merging in master. Please use `git rebase  upstream/master` (assuming that upstream is name of the OPM remote) next time.

Maybe the easiest way to fix this is checking out master and creating a new PR with just the relevant commits cherry-picked?
```
git checkout master
git pull
git checkout -b new-try
git cherry-pick 3ffc5e4 125b0f1 ee84e8b 
```

atgeirr:
> Please - those in the know: give this and #2524 some love and attention; reducing the complexity and repetition in this part of the codebase is a very valuable addition.

I agree, and will prioritize this. I'll start looking at this tomorrow morning at the latest.

hakonhagland:
@blattms wrote:
> Maybe the easiest way to fix this is checking out master and creating a new PR with just the relevant commits cherry-picked?

Good idea, I have submitted a new PR, see https://github.com/OPM/opm-simulators/pull/2535

hakonhagland:
@blattms wrote:
> Please use git rebase upstream/master (assuming that upstream is name of the OPM remote) next time.

Actually I did a `git rebase master`, and I think that was when all the unrelated commits entered into the PR, so it is a little bit creepy. Hope, the same thing will not happen next time :anguished:

akva2:
which means you probably merged in your master. git rebase master rebases on the local master branch.

blattms:
Closing in favor of #2535

PR #2527, opened 2020-04-10
Refactor Equilibration Procedure

bska:
This Pull Request introduces new helper classes,

 * `Opm::EQUIL::Details::PressureTable<>`
 * `Opm::EQUIL::Details::PhaseSaturations<>`

that colectively subsume the responsibility of the existing helper functions

 * `Opm::EQUIL::phasePressures<>()`
 * `Opm::EQUIL::phaseSaturations<>()`

and generalise that functionality to arbitrary depth points within single cells.  This is in preparation of adding support for the `N < 0` case of the initial fluid in place procedure defined in the `EQUIL`
keyword.  The `PhaseSaturations` class consumes an already equlibrated pressure table for the pertinent equilibration region, calculates capillary pressure values and inverts Pc curves to derive saturation values. If the capillary pressure curves are constant within a cell, then a simple depth consideration with respect to the implied sharp phase interface is used to derive saturation values.  We also preserve existing support for SWATINIT-type initialisation of the water saturation field.

Switch `InitialStateComputer<>::calcPressSatRsRv()` over to using the pressure and saturation helper classes instead of the original helper functions since this provides additional control.  Also
remove those helper functions to reduce risk of confusion over which method to use.

Update the unit tests accordingly.

### DIFF
--- a/ebos/equil/initstateequil.hh
+++ b/ebos/equil/initstateequil.hh
@@ -53,8 +53,12 @@
 #include <opm/material/fluidstates/SimpleModularFluidState.hpp>
 #include <opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp>
 
+#include <algorithm>
 #include <array>
 #include <cassert>
+#include <cstddef>
+#include <limits>
+#include <stdexcept>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

atgeirr:
jenkins build this please

PR #2530, opened 2020-04-13
 Equilibration: Add Experimental Support for Horizontal Subdivision

bska:
This Pull Request adds a very early, alpha-quality implementation of the "horizontal subdivision" strategy (N < 0) of the EQUIL directive. This in turn enables more accurate derivations of the initial fluids in place.

Interactions with SWATINIT are completely untested, and the initial Rs/Rv derivations in this context are possibly incomplete.  More work is likely needed in this area, but this does at least enable more widespread testing.

This PR depends on (and includes) #2527 and must not be merged before that PR is in master.

### Example Initialization ###

Below is an example using [MOD4_GRP](https://github.com/OPM/opm-tests/blob/e3e4747457f9d8168909d89afc3657d6ea8523d0/model4/MOD4_GRP.DATA) from OPM-Tests, with the equilibration directive changed as
```diff
 EQUIL
 -- datum    Pres     OWC            Pcow  GOC    Pcog  Rsvd  Rvvd  N
-   1500.0   150      1514.0         0.0   610.0   1*    1     1*   0 /
-   1550.0   150      1547.5         0.0   610.0   1*    1     1*   0 /
+   1500.0   150      1514.0         0.0   610.0   1*    1     1*   1* /
+   1550.0   150      1547.5         0.0   610.0   1*    1     1*   1* /
```
meaning that the 9th item is reset to its default value of `-5` (cells divided into 5+5 horizontal sub-intervals).  At least in this simple case&mdash;a sharp interface between the phases&mdash;we're able to capture a more accurate representation of the O/W contact in the upper compartment.  We also observe that Flow runs more stably in this case and does not cut its step size as frequently or significantly as with the centre-based initialization of current master (86 timesteps in this PR vs. 98 timesteps in current master) when run as
```sh
$ flow --ecl-deck-file-name=${cse} --enable-tuning=true \
    --tolerance-cnv=1.0e-5 --tolerance-cnv-relaxed=5.0e-4
``` 

Simulator |  Initial SOIL
------------ | ---------------
Master | MOD4_GRP_ECL_Flow_Master_3D_View_SOIL_00_30_Jul_2019
This PR | MOD4_GRP_ECL_Flow_Dev_3D_View_SOIL_00_30_Jul_2019
ECLIPSE | MOD4_GRP_ECL_Eclipse_3D_View_SOIL_00_30_Jul_2019

### DIFF
--- a/ebos/equil/equilibrationhelpers.hh
+++ b/ebos/equil/equilibrationhelpers.hh
@@ -675,6 +675,17 @@ public:
      */
     double pcgoGoc() const { return this->rec_.gasOilContactCapillaryPressure(); }
 
+    /**
+     * Accuracy/strategy for initial fluid-in-place calculation.
+     *
+     * \return zero (N=0) for centre-point method, negative (N<0) for the
+     *   horizontal subdivision method with 2*(-N) intervals, and positive
+     *   (N>0) for the tilted subdivision method with 2*N intervals.
+     */
+    int equilibrationAccuracy() const
+    {
+        return this->rec_.initializationTargetAccuracy();
+    }
 
     /**
      * Retrieve dissolved gas-oil ratio calculator of current
@@ -694,7 +705,6 @@ public:
      * Retrieve pvtIdx of the region.
      */
     int pvtIdx() const { return this->pvtIdx_; }
-
 
 private:
     Opm::EquilRecord rec_;     /**< Equilibration data */
--- a/ebos/equil/initstateequil.hh
+++ b/ebos/equil/initstateequil.hh
@@ -1468,12 +1468,10 @@ equilnum(const Opm::EclipseState& eclipseState,
     std::vector<int> eqlnum(grid.size(0), 0);
 
     if (eclipseState.fieldProps().has_int("EQLNUM")) {
-        const int nc = grid.size(/*codim=*/0);
-        eqlnum.resize(nc);
-
         const auto& e = eclipseState.fieldProps().get_int("EQLNUM");
         std::transform(e.begin(), e.end(), eqlnum.begin(), [](int n){ return n - 1;});
     }
+
     return eqlnum;
 }
 
@@ -1514,7 +1512,7 @@ public:
         const Opm::RegionMapping<> eqlmap(equilnum(eclipseState, grid));
         const int invalidRegion = -1;
         regionPvtIdx_.resize(rec.size(), invalidRegion);
-        setRegionPvtIdx(grid, eclipseState, eqlmap);
+        setRegionPvtIdx(eclipseState, eqlmap);
 
         // Create Rs functions.
         rsFunc_.reserve(rec.size());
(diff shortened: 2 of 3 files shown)


atgeirr:
Please rebase this and un-draft if appropriate.

bska:
I have rebased this onto current master and I'll run a build/regression check. Testing by @tskille suggests that this PR generates an initial condition that agrees with ECLIPSE's values&mdash;at least for model 4.

I am however seeing some unsettling differences on Norne (parallel run, 4 MPI processes) between this PR and current master.  In particular, the `PPCW` 3D vector differs from that of master on a subset of the report steps.  I get differences of up to 0.15 % which is a little too big for comfort.  It **may** be a latent issue/race condition that just happens to present as a result of running the loops slightly differently but I'd still like to get to the bottom of it.

Any testing by interested parties would be most welcome.

bska:
jenkins build this please

bska:
jenkins build this please

tskille:
I have tested this PR with model4 is initial water saturation are now calculated correct with item 9 defaulted for Equil. There is a difference in how Eclipse and FLow is handling swatinit together with N < 0. However, this is a rare combination and this is not recommended to do. Hence, I believe this PR should be merged as is.

bska:
> Hence, I believe this PR should be merged as is.

Thank you for the testing&mdash;it is greatly appreciated.  For my part, I'm getting increasingly convinced that the difference I'm seeing for some report steps in the `PPCW` restart vector between this PR and current master is really a latent bug in the way we're handling model initialization in parallel.  We don't distinguish between those cells that are owned by a process and those cells that are in the overlap regions.  Moreover, we presently don't really guarantee that all processes have the same notion of the equilibration regions' vertical extent which does lead to potentially different vertical subdivisions when creating the pressure tables.

Solving those issues is beyond the scope of this PR in my opinion.  I have rebased this onto current master and I'll mark this as "ready for review" and run a build check now.

bska:
jenkins build this please

atgeirr:
jenkins build this please

atgeirr:
Thanks, have read, will merge. I just have a very minor complaint: I would call it vertical subdivision, since it is the vertical direction that is subdivided... (by horizontal slices). But the code is readable and clear so there should be no confusion.

bska:
> I just have a very minor complaint: I would call it vertical subdivision, since it is the vertical direction that is subdivided... (by horizontal slices).

I see the point&mdash;and appreciate it&mdash;but I'm still inclined to refer to this as "horizontal".  The remaining and currently unsupported alternative, `N > 0`, is "tilted" subdivision.  That also subdivides the vertical direction by horizontal slices.  The difference here is that this algorithm (`N < 0`) treats each cell as if the cell **itself** is horizontal whence the cell's vertical extent is defined to be the vertical distance from the centre point of the top surface/connection to the centre point of the bottom surface/Connection.  That range is then split into `2*|N|` intervals of equal vertical size.  In the "tilted" alternative the vertical extent is from the top vertex to the bottom vertex and the range is split into `N` intervals above the cell centre point and `N` intervals below the cell centre point.  The top intervals will generally have a size that differs from the bottom intervals.

PR #2533, opened 2020-04-14
changed: check for pre-existing pybind11 target

akva2:
this is necessary to allow building all modules in a 'super-build' with python enabled.


akva2:
jenkins build this please

akva2:
this should not harm anyone. and it helps me. shameless self-merge.

PR #2536, opened 2020-04-15
Add PYACTION / UDQ testcase

joakim-hove:
This is identical to the existing `UDQ_WCONPROD` but the UDQ values are assigned through Python code instead of through expressions parsed in the .DATA file

In addition to this test it would be advantageous to have a test like 
```
results(UDQ_WCONPROD) == results(PYACTION_WCONPROD)
```

~~opm-test: https://github.com/OPM/opm-tests/pull/283~~
opm-common: https://github.com/OPM/opm-common/pull/1620


akva2:
missing update to update_reference_data.sh

compareECLFiles.cmake
```diff
@@ -399,6 +399,13 @@ add_test_compareECLFiles(CASENAME udq_wconprod
                          REL_TOL ${rel_tol}
                          DIR udq_actionx)
 
+add_test_compareECLFiles(CASENAME udq_pyaction
```
akva2, line 402: this needs to be wrapped in if(OPM_ENABLE_PYTHON)

joakim-hove: Done

joakim-hove: The relevant cmake flag - `OPM_ENABLE_EMBEDDED_PYTHON` is defined and set in opm-common; seems it is not available here? The test is never run.

akva2: right. in that case it needs to be published through the opm-common-config.cmake.

joakim-hove: Ehh - exactly which file is this?

PR #2539, opened 2020-04-15
Removes stale CPR headers that have been superseeded.

blattms:
No need to drag them along and confuse people. There is a new kid in town since quite some time.


blattms:
jenkins build this please

PR #2541, opened 2020-04-16
Check availability for group control separately for inj/prod

joakim-hove:
Dwonstream of: https://github.com/OPM/opm-common/pull/1712


joakim-hove:
As can be seen from the upstream PR this fails *every single regression test* - obviously something I do not understand here ....

joakim-hove:
@atgeirr : It is green from upstream; will review and merge(?) this and upstream?

PR #2543, opened 2020-04-16
Refactor flow_ebos_blackoil.cpp.

hakonhagland:
NOTE: this pull request depends on #2518 which should be merged first.
    
In this PR, we return to the list of PRs presented in Pull #2403, starting with the first two:
- #2438 : *Refactor run() into prepareRun_()*, and
- #2439 : *Refactor flowEbosBlackoilMain()*.
    
The first, #2438, has already been handled by the refactoring of `flow.cpp` in #2516. The second, #2439, has diverged to much from the current master, so I decided to create a new PR (this one) instead of trying to force-adjust it.

### Description
    
We would like to add a `step()` method to `simulators.cpp`, that will let a Python script advance the simulator one report step at a time. Before calling `step()`, the Python script will have to call a `step_init()` method that initializes the simulator, similar to what is done in `run()` now.
    
Currently, the Python `run()` calls `runDynamic()` in `Main.hpp`, which calls `dispatchDynamic_()` which again calls `flowEbosBlackoilMain()` to start the simulation (which calls `execute()` in class `FlowMainEbos`). The Python `step_init()` method should also do the setup in `FlowEbosBlackoilMain()` but it should not call `execute()` directly. In order to avoid code     duplication for `run()` and `step_init()`, the initialization part of `flowEbosBlackoilMain()` is isolated into a `flowEbosBlackoilMainInit()` method.
    
The `step_init()` method (to be implemented in a later PR) will now call `initFlowEbosBlackoil()` in `Main.hpp` to initialize the simulator.

### DIFF
--- a/flow/flow_ebos_blackoil.cpp
+++ b/flow/flow_ebos_blackoil.cpp
@@ -46,8 +46,8 @@ void flowEbosBlackoilSetDeck(double setupTime, Deck *deck, EclipseState& eclStat
     Vanguard::setExternalSummaryConfig(&summaryConfig);
 }
 
-// ----------------- Main program -----------------
-int flowEbosBlackoilMain(int argc, char** argv, bool outputCout, bool outputFiles)
+std::unique_ptr<Opm::FlowMainEbos<TTAG(EclFlowProblem)>>
+flowEbosBlackoilMainInit(int argc, char** argv, bool outputCout, bool outputFiles)
 {
     // we always want to use the default locale, and thus spare us the trouble
     // with incorrect locale settings.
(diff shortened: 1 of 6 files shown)


blattms:
jenkins build this please

hakonhagland:
Seems like [the jenkins build](https://ci.opm-project.org/job/opm-simulators-PR-builder/862/console) failed with:
```
CMake Error at python/simulators/cmake_install.cmake:55 (file):
  file cannot create directory: /simulators.  Maybe need administrative
  privileges.
  python/cmake_install.cmake:43 (include)
  cmake_install.cmake:705 (include)
```

It seems to me that this must come from https://github.com/OPM/opm-simulators/blob/d573f83d64994c32539c320382df827713dbdba2/python/simulators/CMakeLists.txt#L21

Notice that if `PYTHON_INSTALL_PREFIX` is empty the install destination becomes `/simulators`.

hakonhagland:
Closing this PR since it was replaced by #2609.

PR #2555, opened 2020-04-22
Python bindings for the blackoil simulator (simplified).

hakonhagland:
A simplified version of PR #2518 that uses a deck filename to construct the simulator.

After private discussion with @joakim-hove it was decided that the construction of the blackoil simulator from Python using `deck`, `eclipseState`, `schedule`, and `summaryConfig` as constructor arguments (as implemented in #2518) should be replaced by a constructor taking only the deck filename as parameter.

A rewrite of the Python bindings for the blackoil simulator using pybind11 as introduced in #2127. The new version uses the refactored `flow.cpp` introduced in #2516 and thus avoids duplication
of the code in `simulators.cpp`.

This PR will be the starting point for implementing the Python bindings introduced in #2403.

### DIFF
--- a/python/simulators/simulators.cpp
+++ b/python/simulators/simulators.cpp
@@ -1,0 +1,38 @@
+#include "config.h"
+#include <opm/parser/eclipse/Deck/Deck.hpp>
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
+#include <opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp>
+#define FLOW_BLACKOIL_ONLY
+#include <opm/simulators/flow/Main.hpp>
+#include <pybind11/pybind11.h>
+#include <pybind11/embed.h>
+#include <iostream>
+#include <string>
+
+namespace py = pybind11;
+
+class BlackOilSimulator
+{
+public:
+
+    BlackOilSimulator( const std::string &deckFilename) : deckFilename_(deckFilename)
+    {
+    }
+
+    int run()
+    {
+        auto mainObject = Opm::Main( deckFilename_ );
+        return mainObject.runDynamic();
+    }
+    
+private:
+    const std::string deckFilename_;
+};
+
+PYBIND11_MODULE(simulators, m)
+{
+    py::class_<BlackOilSimulator>(m, "BlackOilSimulator")
+        .def(py::init< const std::string& >())
+        .def("run", &BlackOilSimulator::run);
+}
--- a/opm/simulators/flow/Main.hpp
+++ b/opm/simulators/flow/Main.hpp
@@ -64,6 +64,7 @@
 #include <opm/simulators/utils/ParallelSerialization.hpp>
 #endif
 
+#include <string>
 #include <type_traits>
 
 BEGIN_PROPERTIES
@@ -126,6 +127,17 @@ namespace Opm
         };
     public:
         Main(int argc, char** argv) : argc_(argc), argv_(argv)  {  }
+
+        Main(const std::string &filename)
+        {
+            deckFilename_.assign(filename);
+            flowProgName_.assign("flow");
+            argc_ = 2;
+            saveArgs_[0] = const_cast<char *>(flowProgName_.c_str());
+            saveArgs_[1] = const_cast<char *>(deckFilename_.c_str());
+            argv_ = saveArgs_;
+        }
+
         Main(int argc,
              char** argv,
              std::shared_ptr<Opm::Deck> deck,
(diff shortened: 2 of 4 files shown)


joakim-hove:
jenkins build this please

python/simulators/simulators.cpp
```diff
@@ -0,0 +1,38 @@
...
+    {
+        auto mainObject = Opm::Main( deckFilename_ );
+        return mainObject.runDynamic();
+    }
+    
+private:
+    const std::string deckFilename_;
+};
+
+PYBIND11_MODULE(simulators, m)
```
joakim-hove, line 33: Here the `BlackOilSimulator` object is just a glorified `std::string` - I would let the `BlackOilSimulator`object wrap a proper C++ `Opm::Main` object:
```
int run(const Opm::Main * self) {
     return self->runDynamic();
}

py::class_<Opm::Main>(m, "BlackOilSimulator")
        .def(py::init< const std::string& >())
        .def("run", &run);
```

hakonhagland: Interesting idea, but what about future extension of the Python interface. We may get methods like `step()`, `getPrimaryVariable()`, `getPorosity()` and so on. Wouldn't it be better to separate the Python interface from the `Opm::Main` class? So by having a separate class `BlackoilSimulator` for all the Python interface methods, we reduce the complexity of the `Opm::Main` class.

hakonhagland:
jenkins build this please

python/simulators/simulators.cpp
joakim-hove: > but what about future extension of the Python interface?

Well - that is the reason I want to move in the suggested direction; the way it is in the current PR the lifetime of the `Main` object is limited to the `run()` method - so right now it is impossible to create e.g. a `getPorosity()`

> We may get methods like step(), getPrimaryVariable(), getPorosity() 

Yes; and with the exception of `getPrimaryVariable()` where I am uncertain these methods can be based on first getting `EclipseState` and `Schedule` from `Opm::Main` - so the `Opm::Main` class does not need to become very complicated.

hakonhagland: > the Main object is limited to the run()
In the original code, the idea was to store a unique pointer to a `Opm::FlowMainEbos<TTAG(EclFlowProblem)>>` object in the `BlackoilSimulator` object. Then from this pointer one could access fluid state, porosity, cell volumes, and so on...

> does not need to become very complicated

I am not yet convinced if it will be possible to keep it simple. Also, the `Opm::Main` class is used by the flow simulators, and they are not interested in the Python stuff. So why should they be bothered with this extra API to be added to `Opm::Main` that does not apply to them? Wouldn't it be better to try to keep the Python stuff separate? 

Just my thoughts. It is of course OK to try out your idea first and see how it will work out, then we can later refactor if it becomes necessary.

atgeirr: If the Main object is a member of the BlackoilSimulator you would still have separation of interfaces, right? I think that sounds like a good idea (but not a Py expert...)

joakim-hove: > In the original code, the idea was to store a unique pointer to a Opm::FlowMainEbos<TTAG(EclFlowProblem)>>

Well - some like the mother and some like the daughter; personally I find the `TTAG<EclFlowProblem>` hard to work with and would recommend/prefer working through the `EclipseState` and `Schedule` datastructures to the extent possible.

> Also, the Opm::Main class is used by the flow simulators, and they are not interested in the Python stuff.

Well - they *are* interested in increased modularity.

hakonhagland: > would recommend/prefer working through the EclipseState and Schedule

I am not so familiar with `EclipseState` and `Schedule`, I guess you could get the porosity from those, but in the case when the porosity is modified from Python at a report step, it will not be updated in the `EclipseState` right? We need to access the `EclProblem` class object inside the `FlowMainEbos` object in order to update and set the reference porosity dynamically right?

joakim-hove: > We need to access the EclProblem class object inside the FlowMainEbos object in order to update and set the reference porosity dynamically right?

Yes - but I think what you really need to update in that case is the `PORV` property deep in there. I still think it would be smart to get *something* going which does not require full internal details of the `TTAG<EclProblem>` class - at least the `Schedule` class - with all timestepping information and all controls is fully specified on the outside of the `TTAG<>`.

hakonhagland: What about setting the pressure for example? Then we need to access the Blackoil model in the Simulator object within FlowMainEbos? The model has the `solution` vector for each timestep.

hakonhagland:
rebased to current master

joakim-hove:
jenkins build this please

PR #2556, opened 2020-04-23
Fix how the group reduction works

totto82:
The target reduction due to wells not under group control is applied on the group under control level or on levels where the guiderates are explicitly set.


atgeirr:
jenkins build this please

totto82:
jenkins build this please

PR #2564, opened 2020-04-24
Fix GCONSALE

totto82:
1) Only check GCONSALE limits at the end of a timestep. NB. The simulator still throws if some of the limits are violated
2) BUGFIX: don't remove gas consumption twice

Tested on `9_3B_---_STW/MSW` and `9_3C_---_STW/MSW`


totto82:
jenkins build this please

alfbr:
I can confirm that this PR resolves the 9_3B/C cases in the model2 folder.

PR #2565, opened 2020-04-27
use nupcol to avoid oscillation between group controls

totto82:
This is sufficient to run the 9_4[B,C,D] test cases in opm-tests/model2.


totto82:
jenkins build this please

PR #2566, opened 2020-04-27
Remove VISCREF from MissingFeatures

joakim-hove:
The `VISCREF`keyword is supported.


joakim-hove:
jenkins build this please

PR #2569, opened 2020-04-27
Prevent Cmake error for CUDA with incompatible c++ compiler.

blattms:
At least on Debian 10 the standard c++ compiler is g++-8, but CUDA only supports g++-7 and our test for CUDA in cmake did send an error in that case/combination which is quite annoying.

The reason was that check_language(CUDA) did not honor the CMAKE_CUDA_FLAGS variable and always used the default g++7, but enable_language(CUDA) did honor it. As we do set the underlying host compiler the fromer reported that CUDA is available while the latter marked the CUDA compiler as broken.

With this commit we work around this by setting the environment variable ENV{CUDAHOSTCXX} which nvcc will use. Hence now we only try to enable CUDA if it is compatible with the C++ compiler

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,7 +84,7 @@ include ("${project}-prereqs")
 # Make sure we are using the same compiler underneath
 # NVCC as for the rest. In the case that NVCC does not support
 # that compiler it will error out. Unfortunately this will only
-# work CMake >= 3.8. We found no way to make FindCUDA.cmake error
+# work for CMake >= 3.8. We found no way to make FindCUDA.cmake error
 # out. It seems to ignore CMAKE_NVCC_FLAGS and CMAKE. Additionally
 # our way of specifying cuda source files never worked for CMake
 # version < 3.8. Hence we deactivate cuda for these versions.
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

blattms:
Sigh, somehow I am experiencing linking errors now. Might be related to this one.

blattms:
Linker error were due to unclean build. But I fixed a few other problems. Tested with dune 2.7 and master

blattms:
jenkins build this please

blattms:
I would like to hear opinions, whether this approach (non-fatal cmake error and deactivating CUDA if the hostcompiler and OPM comiler are not the same) is the right way. @Tongdongq, @alfbr.
Currently the cmake throws an error and the build stops prematurely. This is annoying if you do not want to test CUDA. If you do that is another story, of course.

CMakeLists.txt
```diff
@@ -93,9 +93,24 @@ include ("${project}-prereqs")
...
+      "is not CUDA compiler if your system's CUDA compiler does not support "
+      "${CMAKE_CXX_COMPILER}.")
+    # set(CMAKE_CUDA_FLAGS "-ccbin ${CMAKE_CXX_COMPILER} ${CMAKE_CUDA_FLAGS}")
+    # check_language does not seem to care about ${CMAKE_CUDA_FLAGS}.
+    # Hence set(CMAKE_CUDA_FLAGS$(_flags_suffix) "-ccbin ${CMAKE_CXX_COMPILER} ${CMAKE_CUDA_FLAGS${_flags_suffix}}")
+    # would have no effect. We need to set the environment variable ENV{CUDAHOSTCXX} such
+    # that cuda picks up the correct compiler during check_language.
+    # In check_language(CUDA) we will get and error if we set and put
+    # "-ccbin ${CMAKE_CXX_COMPILER}" into the CMAKE_CUDA_FLAGS as this results
+    # in "${NVCC} -ccbin=${CMAKE_CXX_COMPILER}   -ccbin ${CMAKE_CXX_COMPILER}"
```
Tongdongq, line 112: I don't understand how `check_language` interacts with `CMAKE_CUDA_FLAGS`. It seems not to care, but if `CMAKE_CUDA_FLAGS` is changed, it does change the compile command.
Possible typo: we will get **and** error if we set and put

CMakeLists.txt
```diff
@@ -105,12 +120,12 @@ if(NOT CMAKE_DISABLE_FIND_PACKAGE_CUDA AND
   check_language(CUDA)
```
Tongdongq, line 120: I cannot comment on lines 118-119: "Since the OPTIONAL flag of enable_language is currently ignored CMake will error out below if the options do not work."
Would this be outdated then?

CMakeLists.txt
blattms: It uses them and if ENV{CUDAHOSTCC} is used in addition we end with two `--ccbin compiler` on the nvcc command line and nvcc will abort. That will make Cmake abort with an error.

blattms:
Starting to share parts of Torvalds view now when it comes to this stuff.

blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
shameless self merge.

blattms:
Probably a bit late for the release. Most people will not have CUDA anyway. Why bother.

If this is not a release, then there should be a known issue about it in the release notes.

PR #2572, opened 2020-04-27
Remove Unused 'currentControl' Variable

bska:
Introduced, but not actually used, in #2564.


bska:
jenkins build this please

PR #2576, opened 2020-04-28
Removes unused member istlSolver_ and method to get it from BlackoilM…

blattms:
…odelEbos

It is not used since quite some time and tends to confuse me.


blattms:
jenkins build this please

PR #2579, opened 2020-04-30
quell deprecation warnings with boost 1.71

akva2:
header was relocated.


tests/test_milu.cpp
```diff
@@ -12,7 +12,12 @@
 #include<opm/simulators/linalg/ParallelOverlappingILU0.hpp>
 
 #include <boost/test/unit_test.hpp>
+#include <boost/version.hpp>
+#if BOOST_VERSION / 100000 == 1 && BOOST_VERSION / 100 % 1000 < 71
```
blattms, line 16: I love these. They are so explicit.

PR #2588, opened 2020-05-04
Implement min/max for GCONSALE

totto82:
Only default action RATE is implemented for violation of maximum sales target

Tested on 9_3D and 9_3E.

PR #2589, opened 2020-05-05
Group tests5

alfbr:
Four more group tests have been checked for correctness. This PR should not be merged before https://github.com/OPM/opm-simulators/pull/2588


totto82:
Last time I checked there seemed to be a too early close down of a well in 9_3E_FOO_MSW due to convergence issues. Is this no longer an issue? Or is the plan to merge this and do a update of the data when that particular issue is solved.

alfbr:
I am not able to reproduce the behaviour. All wells seems to run through as they should. I have tested with default time stepping, and setting the time step to maximum one day. Both run through, and provide similar results. Which well do you observe, and when does it close?

totto82:
I run with enable-tuning=true. Well PROD3 closes 1 March 2019 due to convergence issues. It works fine when I use default time-stepping.

totto82:
The convergence issue seems a bit random. Both --max-pressure-change-ms-wells=2e6  and --max-pressure-change-ms-wells=0.8e6 works. Default is 1.0e6.

alfbr:
Strange, yes I can reproduce when using enable-tuning. However, I use the command line option --solver-max-time-step-in-days=1 instead, and it runs through without issues. It looks like the two command line options are not doing the same thing. I suspect enable-tuning changes more parameters than it is supposed to. I do not think that is related to this PR though, and this fixes a real issue.

totto82:
> I do not think that is related to this PR though, and this fixes a real issue.
Agree, this can be merged after #2588 is merged.

atgeirr:
> Agree, this can be merged after #2588 is merged.

Which is now...

atgeirr:
jenkins build this please

atgeirr:
Only failure is the expected: the new tests lack reference data. Proceeding with a data update.

atgeirr:

bska:
jenkins build this opm-tests=315 please

compareECLFiles.cmake
```diff
@@ -608,6 +608,34 @@ add_test_compareECLFiles(CASENAME 9_3c_grpctl_msw_model2
                          REL_TOL ${rel_tol}
                          DIR model2)
 
+add_test_compareECLFiles(CASENAME 9_3d_grpctl_stw_model2
+                         FILENAME 9_GINJ_GAS_MAX_EXPORT_STW
```
bska, line 612: I think there's a misprint here.  I **believe** this `FILENAME` should be
```
9_3D_GINJ_GAS_MAX_EXPORT_STW
```
instead (with `_3D` inserted between `9` and `_GINJ`.)

compareECLFiles.cmake
```diff
@@ -608,6 +608,34 @@ add_test_compareECLFiles(CASENAME 9_3c_grpctl_msw_model2
...
+                         SIMULATOR flow
+                         ABS_TOL ${abs_tol}
+                         REL_TOL ${rel_tol}
+                         DIR model2)
+
+add_test_compareECLFiles(CASENAME 9_3d_grpctl_msw_model2
+                         FILENAME 9_GINJ_GAS_MAX_EXPORT_MSW
```
bska, line 619: I believe this `FILENAME` should be
```
9_3D_GINJ_GAS_MAX_EXPORT_MSW
```
instead (with `_3D` inserted between `9` and `_GINJ`.)

compareECLFiles.cmake
alfbr: Indeed, thank you!

alfbr:

bska:
jenkins build this opm-tests=315 please

bska:
Looks like we picked up a stale reference to OPM/opm-tests#315.  I'll close that temporarily to (try to) resolve the issue.

bska:
jenkins build this opm-tests=315 please

bska:
Build check is green and the updated reference cases have been merged into master.  I'll merge this now to bring the systems back in sync.  Thanks a lot for expanding our set of regression tests.

PR #2590, opened 2020-05-05
Fix solvent reinjection when different reinjection group is given

totto82:
The solvent model was forgotten when separate re-injection group was implemented.


totto82:
jenkins build this please

atgeirr:
Looks fine, merging.

PR #2592, opened 2020-05-05
Udq test

alfbr:
This PR enables the UDQ_ACTIONX test for model4. It should only be merged after the fix in https://github.com/OPM/opm-common/pull/1772 is merged.

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -699,6 +699,13 @@ add_test_compareECLFiles(CASENAME model4_group
                          REL_TOL ${rel_tol}
                          DIR model4)
 
+add_test_compareECLFiles(CASENAME model4_udq_group
+                         FILENAME MOD4_UDQ_ACTIONX
+                         SIMULATOR flow
+                         ABS_TOL ${abs_tol}
+                         REL_TOL ${rel_tol}
+                         DIR model4)
+
 add_test_compareECLFiles(CASENAME wsegsicd
 			  FILENAME TEST_WSEGSICD
 			  SIMULATOR flow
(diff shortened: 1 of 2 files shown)


joakim-hove:
jenkins build this please

alfbr:
This case needs to be revisited after the latest change in the data set. Running with
flow--solver-max-time-step-in-days=1 MOD4_UDQ_ACTIONX.DATA
The well OPU1 opens at 1 OCT 2019 (as it should)
However, instead of being controlled by LRAT with 150, it switches to GRUP, creating spikes in several summary curves. Using default time stepping, the correct LRAT control is used for the time step.

totto82:
I notice that Flow report the following for OPU1:

> Warning: Failed in getting converged thp based potential calculation for well OPU1. Instead the bhp based value is used

Difference in potentials from this may explain the discrapancy in results. The discrapancy of the results (LRAT) corresponds to the same time as the discrapancy in the well potentials (WOPP) due to the reason given in the warning. @atgeirr The well potential code may need some more tuning.

atgeirr:
The wrong behaviour observed in this test was fixed in https://github.com/OPM/opm-common/pull/1818, so this should be merged now after a data update. I'll start the process.

atgeirr:

atgeirr:
jenkins build this opm-tests=338 please

bska:
jenkins build this opm-tests=338 please

bska:
There were a number of minor solution differences compared to the reference data in OPM/opm-tests#338.  Those were possibly caused by OPM/opm-simulators#2661.  I am nevertheless going to merge this in its current state because OPM/opm-common#1408 will bring a second round of reference data updates which will restore a clean build.

Unfortunately, this means that the UDQ regression test will be broken for a few hours.  Please keep this in mind when reviewing build failures.

bska:
The new reference solution in OPM/opm-tests#338 has been merged into master.  I'm merging this now to activate the regression test.  The reference solution will get a second round of updates as part of merging OPM/opm-common#1408.

PR #2600, opened 2020-05-08
compute initial volume fractions on the well copies

totto82:


totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -778,9 +778,7 @@ namespace Opm
             well_state_copy.currentProductionControls()[index_of_well_] = Well::ProducerCMode::BHP;
         }
         well_state_copy.bhp()[well_copy.index_of_well_] = bhp;
-
-        well_copy.updatePrimaryVariables(well_state_copy, deferred_logger);
-        well_copy.initPrimaryVariablesEvaluation();
+        well_copy.calculateExplicitQuantities(ebosSimulator, well_state_copy, deferred_logger);
```
GitPaean, line 781: It is understandable to calculate the explicit quantities here, which is based on the well_state_copy. Did you evaluate the benefit here or it solves some problems?

And also, why removing the update the primary variables? Because when well state changes, the primary variables need to be updated.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -3006,8 +3000,6 @@ namespace Opm
         updateWellStateWithTarget(ebos_simulator, well_state_copy, deferred_logger);
 
         calculateExplicitQuantities(ebos_simulator, well_state_copy, deferred_logger);
-        updatePrimaryVariables(well_state_copy, deferred_logger);
```
GitPaean, line 3009: Same with here.

opm/simulators/wells/MultisegmentWell_impl.hpp
GitPaean: Okay. I did not see `updatePrimaryVariables` was added to `calculateExplicitQuantities` some time ago. Although it should not be there. 

I take back my comment here. Sorry about that.

totto82:
The test failures are expected since the well potential calculations are affected slightly by this. The changes in results are minimal but it improves convergence for the potential calculations. Tests that uses well potentials are also affected slightly. We need to update the ref. data for this.

totto82:
@atgeirr Any comments? I have some more PRs coming in but I want this merged first. This is a trivial bugfix.

atgeirr:
The code is fine. I looked at some of the test failures and they were indeed unproblematic. I assume you have checked them all?

We should run the update_data, but since there is some ongoing opm-tests merging and data update stuff due to https://github.com/OPM/opm-tests/pull/319 I think perhaps we should wait until the dust settles somewhat.

bska:
> We should run the update_data, but since there is some ongoing opm-tests merging and data update stuff due to [OPM/opm-tests#319](https://github.com/OPM/opm-tests/pull/319) I think perhaps we should wait until the dust settles somewhat.

Actually, the earlier attempt to update the reference data didn't quite succeed so there is currently no active data update request in flight.  If this is ready to go, then we'll merge OPM/opm-tests#319 right away and use this PR to generate new reference data for that update too.

bska:
> If this is ready to go, then we'll merge OPM/opm-tests#319 right away and use this PR to generate new reference data for that update too.

Any objections to this, @totto82, @atgeirr, @alfbr ?

atgeirr:
I assume then the master will be broken for some period, until the data update gets generated and merged? Based on earlier experience I am a bit afraid it could take some hours if we get some stale repo etc. etc.

But: If that's what we have to do then let's do it.

bska:
> I assume then the master will be broken for some period, until the data update gets generated and merged?

That is true.

>  I am a bit afraid it could take some hours if we get some stale repo etc. etc.

Certainly possible, but the risk is low Right Now[tm].  There are no other data updates in flight, for instance.

> If that's what we have to do then let's do it.

Okay, cool.  I'll start the process.

bska:

bska:
jenkins build this opm-tests=321 please

bska:
Build check is green and updated reference data have been merged into master.  I'll merge this now.

PR #2601, opened 2020-05-08
fixed unininitialized value,

hnil:
fixed uninitialized value in the smoother args for make using ParILUO as smoother for amg. 
and added commented specialization for the solver factory

### DIFF
--- a/opm/simulators/linalg/PreconditionerFactory.hpp
+++ b/opm/simulators/linalg/PreconditionerFactory.hpp
@@ -143,12 +143,33 @@ private:
         using SmootherArgs = typename Dune::Amg::SmootherTraits<Smoother>::Arguments;
         SmootherArgs smootherArgs;
         smootherArgs.iterations = prm.get<int>("iterations", 1);
+	//int iluwitdh = prm.get<int>("iluwidth", 0);
+	//MILU_VARIANT milu = convertString2Milu(prm.get<std::string>("milutype",std::string("ilu")));
+	//smootherArgs.setMilu(milu);
+	//smootherArgs.setN(iluwitdh);
         // smootherArgs.overlap=SmootherArgs::vertex;
         // smootherArgs.overlap=SmootherArgs::none;
         // smootherArgs.overlap=SmootherArgs::aggregate;
-        smootherArgs.relaxationFactor = prm.get<double>("relaxation", 0.9);
+        smootherArgs.relaxationFactor = prm.get<double>("relaxation", 1.0);
         return smootherArgs;
     }
+    // template <>
+    // static auto amgSmootherArgs< Opm::ParallelOverlappingILU0< Matrix, Vector, Vector, Comm>  >(const boost::property_tree::ptree& prm)
+    // {
+    // 	using Smoother = Opm::ParallelOverlappingILU0<Matrix, Vector, Vector, Comm>;
+    //     using SmootherArgs = typename Dune::Amg::SmootherTraits<  Smoother >::Arguments;
+    //     SmootherArgs smootherArgs;
+    //     smootherArgs.iterations = prm.get<int>("iterations", 1);
+    // 	int iluwitdh = prm.get<int>("iluwidth", 0);
+    // 	MILU_VARIANT milu = convertString2Milu( prm.get<std::string>("milutype",std::string("ilu") ) );
+    // 	smootherArgs.setMilu(milu);
+    // 	smootherArgs.setN(iluwitdh);
+    //     // smootherArgs.overlap=SmootherArgs::vertex;
+    //     // smootherArgs.overlap=SmootherArgs::none;
+    //     // smootherArgs.overlap=SmootherArgs::aggregate;
+    //     smootherArgs.relaxationFactor = prm.get<double>("relaxation", 1.0);
+    //     return smootherArgs;
+    // }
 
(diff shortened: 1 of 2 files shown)


GitPaean:
Should we have a parallel regression test for `flow_blackoil_dunecpr`?

hnil:
We probably should have a test. I don not know the policy. I would have prefered one with extra tolerances of the linear solve and nonlinear solve. We should also fix a spesialization of:

opm/simulators/linalg/PreconditionerFactory.hpp
```diff
@@ -143,12 +143,33 @@ class PreconditionerFactory
         using SmootherArgs = typename Dune::Amg::SmootherTraits<Smoother>::Arguments;
         SmootherArgs smootherArgs;
         smootherArgs.iterations = prm.get<int>("iterations", 1);
+	//int iluwitdh = prm.get<int>("iluwidth", 0);
+	//MILU_VARIANT milu = convertString2Milu(prm.get<std::string>("milutype",std::string("ilu")));
+	//smootherArgs.setMilu(milu);
+	//smootherArgs.setN(iluwitdh);
```
blattms, line 149: What is up with this? Please remove commented code

opm/simulators/linalg/PreconditionerFactory.hpp
```diff
@@ -143,12 +143,33 @@ class PreconditionerFactory
...
         // smootherArgs.overlap=SmootherArgs::vertex;
         // smootherArgs.overlap=SmootherArgs::none;
         // smootherArgs.overlap=SmootherArgs::aggregate;
-        smootherArgs.relaxationFactor = prm.get<double>("relaxation", 0.9);
+        smootherArgs.relaxationFactor = prm.get<double>("relaxation", 1.0);
```
blattms, line 153: Maybe this deserves to be mentioned in the commit message.

opm/simulators/linalg/PreconditionerFactory.hpp
```diff
@@ -143,12 +143,33 @@ class PreconditionerFactory
...
+    // 	int iluwitdh = prm.get<int>("iluwidth", 0);
+    // 	MILU_VARIANT milu = convertString2Milu( prm.get<std::string>("milutype",std::string("ilu") ) );
+    // 	smootherArgs.setMilu(milu);
+    // 	smootherArgs.setN(iluwitdh);
+    //     // smootherArgs.overlap=SmootherArgs::vertex;
+    //     // smootherArgs.overlap=SmootherArgs::none;
+    //     // smootherArgs.overlap=SmootherArgs::aggregate;
+    //     smootherArgs.relaxationFactor = prm.get<double>("relaxation", 1.0);
+    //     return smootherArgs;
+    // }
```
blattms, line 172: Why the addition a a completely commented out code block?

blattms:
Does this fix the parallel runs with cpr/flexible solvers? I had experienced the opposite  of scalability (more time using more cores) on some very easy problems.

opm/simulators/linalg/PreconditionerFactory.hpp
hnil: Because It should be fixed. But commenting out the code do not compile at least for me.

hnil:
Flexible solver is just a wrapper of dune-istl solvers. So if it is problems with the scaling it need to be related to dune-istl. "Simple cases"/small cases would probably not have good scaling with cpr or rather with amg...

opm/simulators/linalg/PreconditionerFactory.hpp
hnil: Do you have an opinion? The default settings for flow_blackoil_dunecpr is anyway 1.0.

hnil:
I am also unsure if the redistribution feature is compiled correctly with standard settings using OPM. It works when tested standalone.

blattms:
> So if it is problems with the scaling it need to be related to dune-istl
????

Just been hit by this problem in my very own branch (`iluIteration_=32767`) in parallel. Maybe we can agree that even when using wrappers not all problems are upstream? (I might be to blame anyway, though ;)

Will give this a spin and merge if it fixes the problem.

blattms:
jenkins  build this please.

blattms:
With this change it is actually running in parallel. Now we can wonder about scaling...

@hnil Will you clean this up? I.e. remove all no-op changes. The real fix seems to be only one line.

opm/simulators/linalg/PreconditionerFactory.hpp
blattms: Sorry, but that is not an argument. Can't you use git stash?

hnil:
I can clean it up, but need help with the template specialization which I commented out. It is not needed, but it will make the code more complete.

blattms:
Well, me intend was too get the (rather urgent) bugfix merged and leave the
rest for another day. I am not a big fan of Prs including multiple things.
I might do that myself to speed things up.

blattms:
@hnil I created PR #2614 with just the fix. The rest based on that PR can be found in [blattms/opm-simulators:hnil-complete](https://github.com/blattms/opm-simulators/tree/hnil-complete)

atgeirr:
Replaced by #2614 and #2616.

PR #2602, opened 2020-05-08
added absolute minim value for relaxed tolerance

hnil:
Changed relaxed tolerance to absolute minium which make sence. A better choice would be to tighten the normal tolerance to 1e-3 and putt the relaxed to 1e-2. Fully relaxed cnv and no projection of saturation is probably not a good choise for the default options.

### DIFF
--- a/opm/simulators/flow/BlackoilModelParametersEbos.hpp
+++ b/opm/simulators/flow/BlackoilModelParametersEbos.hpp
@@ -61,7 +61,7 @@ SET_SCALAR_PROP(FlowModelParameters, DwellFractionMax, 0.2);
 SET_SCALAR_PROP(FlowModelParameters, MaxResidualAllowed, 1e7);
 SET_SCALAR_PROP(FlowModelParameters, ToleranceMb, 1e-6);
 SET_SCALAR_PROP(FlowModelParameters, ToleranceCnv,1e-2);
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

totto82:
I agree that the current default value sometimes leads to unphysical results in some cells. I often use 1 as relaxed tolerance for my simulations to avoid this. The problem is that changing this could lead to a significant increase in simulation time. i.e. this need some testing before it can be merged.

hnil:
I agree that it will increase time, but having a default which is so relaxed is not ok in long term if we care about results in all cells. My suggestion is 1e-3 as default and 1e-2 as relaxed.

atgeirr:
For now this change has been made less critical by the change in default behaviour (taking pore volume of cells into account). We may need to adjust these tolerances in stricter direction, but we should first see if the new defaults fail on the cases where the old defaults failed. Also, there are now another parameter that could be tweaked, the relative pore volume threshold.

alfbr:
Yes, the default for the relaxed is now 1, and it is by default only used on 3% of the pore volume. I believe it is important to take into account that defaults are supposed to be compromise. In this case a compromise between computational time and accuracy. There will always be corner cases where the solution drifts off regardless of what tolerances we use. @hnil please check if the current defaults work well for your cases, if not, please share.

PR #2606, opened 2020-05-10
Remove 'MAXVALUE' From List of Missing Features

bska:
We've supported that operator for quite some time.


bska:
jenkins build this please

joakim-hove:
Thank you for the fix.

PR #2608, opened 2020-05-11
Make flexible solvers and CPR accessible for flow binary.

blattms:
Previously, one had to call a seperate binary called flow_blackoil_dunecpr. Unforntunately, that was only built if users requested that the tests (`-DBUILD_TESTING=ON`) and the flow variants (`-DBUILD_FLOW_VARIANTS=ON`) should be built. In addition it would use a slightly different nonlinear solver implementation.

With this commit flow can be asked to either use CPR (`--use-cpr=true --matrix-add-well-contributions=true`) or to use the flexible solvers bx setting the `--linear-solver-configuration` option. In all other cases the usual solver implementations are still used.

Note that the flexible solvers still need `--matrix-add-well-contributions=true` and hence cannot cope with multi-segment wells.

### DIFF
--- a/opm/simulators/linalg/ISTLSolverEbos.hpp
+++ b/opm/simulators/linalg/ISTLSolverEbos.hpp
@@ -823,7 +823,7 @@ protected:
 
         bool isParallel() const {
 #if HAVE_MPI
-            return parallelInformation_.type() == typeid(ParallelISTLInformation);
+            return comm_->communicator().size() > 1;
 #else
             return false;
 #endif
--- a/opm/simulators/linalg/ISTLSolverEbos.hpp
+++ b/opm/simulators/linalg/ISTLSolverEbos.hpp
@@ -359,14 +359,13 @@ protected:
             }
 #endif
             extractParallelGridInformationToISTL(simulator_.vanguard().grid(), parallelInformation_);
-            const auto wellsForConn = simulator_.vanguard().schedule().getWellsatEnd();
             const bool useWellConn = EWOMS_GET_PARAM(TypeTag, bool, MatrixAddWellContributions);
 
             ownersFirst_ = EWOMS_GET_PARAM(TypeTag, bool, OwnerCellsFirst);
             interiorCellNum_ = detail::numMatrixRowsToUseInSolver(simulator_.vanguard().grid(), ownersFirst_);
 
             if (!ownersFirst_ || parameters_.linear_solver_use_amg_  || parameters_.use_cpr_ ) {
-                detail::setWellConnections(gridForConn, wellsForConn, useWellConn, wellConnectionsGraph_);
+                detail::setWellConnections(gridForConn, simulator_.vanguard().schedule().getWellsatEnd(), useWellConn, wellConnectionsGraph_);
                 // For some reason simulator_.model().elementMapper() is not initialized at this stage
                 // Hence const auto& elemMapper = simulator_.model().elementMapper(); does not work.
                 // Set it up manually
--- a/opm/simulators/linalg/ISTLSolverEbos.hpp
+++ b/opm/simulators/linalg/ISTLSolverEbos.hpp
@@ -350,7 +350,7 @@ protected:
             ownersFirst_ = EWOMS_GET_PARAM(TypeTag, bool, OwnerCellsFirst);
             interiorCellNum_ = detail::numMatrixRowsToUseInSolver(simulator_.vanguard().grid(), ownersFirst_);
 
-            if (!ownersFirst_ || parameters_.linear_solver_use_amg_  || useFlexible_ ) {
+            if ( isParallel() && (!ownersFirst_ || parameters_.linear_solver_use_amg_  || useFlexible_ ) ) {
                 detail::setWellConnections(gridForConn, simulator_.vanguard().schedule().getWellsatEnd(), useWellConn_, wellConnectionsGraph_);
                 // For some reason simulator_.model().elementMapper() is not initialized at this stage
                 // Hence const auto& elemMapper = simulator_.model().elementMapper(); does not work.
@@ -359,11 +359,8 @@ protected:
                     Dune::MultipleCodimMultipleGeomTypeMapper<GridView>;
                 ElementMapper elemMapper(simulator_.vanguard().gridView(), Dune::mcmgElementLayout());
                 detail::findOverlapAndInterior(gridForConn, elemMapper, overlapRows_, interiorRows_);
-                if (gridForConn.comm().size() > 1) {
-
-                    noGhostAdjacency();
-                    setGhostsInNoGhost(*noGhostMat_);
-                }
+                noGhostAdjacency();
+                setGhostsInNoGhost(*noGhostMat_);
                 if (ownersFirst_)
                     OpmLog::warning("OwnerCellsFirst option is true, but ignored.");
             }
@@ -510,7 +507,7 @@ protected:
                 else {
 
                     typedef WellModelMatrixAdapter< Matrix, Vector, Vector, WellModel, true > Operator;
-
+                    assert (noGhostMat_);
                     copyJacToNoGhost(*matrix_, *noGhostMat_);
                     Operator opA(*noGhostMat_, *noGhostMat_, wellModel,
                                  comm_ );
--- a/opm/simulators/linalg/ISTLSolverEbos.hpp
+++ b/opm/simulators/linalg/ISTLSolverEbos.hpp
@@ -404,13 +404,18 @@ protected:
             {
                 if (firstcall)
                 {
-                    //Only do this once as flexible solvers wil store a reference.
-                    matrix_.reset(new Matrix(M.istlMatrix()));
+                    // ebos will not change the matrix object. Hence simply store a pointer
+                    // to the original one with a deleter that does nothing.
+                    matrix_.reset(const_cast<Matrix*>(&M.istlMatrix()), [](Matrix*){});
                 }
                 else
                 {
-                    // Pointers stays the same for flexible solver
-                    *matrix_ = M.istlMatrix();
+                    // Pointers should not change
+                    if ( &(M.istlMatrix()) != matrix_.get() )
+                    {
+                        OPM_THROW(std::logic_error, "Matrix objects are expected to be reused when reassembling!"
+                                  <<" old pointer was " << matrix_.get() << ", new one is " << (&M.istlMatrix()) );
+                    }
                 }
             }
             rhs_ = &b;
@@ -1133,7 +1138,7 @@ protected:
         mutable bool converged_;
         std::any parallelInformation_;
 
-        std::unique_ptr<Matrix> matrix_;
+        std::shared_ptr<Matrix> matrix_;
         std::unique_ptr<Matrix> noGhostMat_;
         Vector *rhs_;
         std::unique_ptr<Matrix> matrix_for_preconditioner_;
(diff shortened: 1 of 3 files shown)


blattms:
jenkins build this please

hnil:
One could also build it with: make flow_blackoil_dunecpr.

alfbr:
@hnil or @atgeirr , what is the status with respect to multi-segment wells, is it working in  flow_blackoil_dunecpr?

atgeirr:
> @hnil or @atgeirr , what is the status with respect to multi-segment wells, is it working in flow_blackoil_dunecpr?

Not yet. It requires one of two things.
 - Make the flexible solver take the linear operator (which for flow then includes the effect of wells) as argument, and pass the reservoir matrix on to the preconditioner (as is done by the default linear solver setup). I have this implemented in a branch, and working in serial, but in parallel I have some problems (likely related to some of the recent changes in ISTLSolverEbos, but not 100% sure). I hope this can be completed relatively quickly, and it would also be the most flexible and advanced option going forward.
  - Implement the MatrixAddWellContributions option for multi-segment wells. Also should be relatively quick, disadvantage is that it might be a little numerically less robust, as it involves matrix inverses of larger matrices than the standard well case.

Doing both would be ideal, because it would give full freedom to choose the operator/matrix approach that works best for any given case.

blattms:
jenkins build this please

blattms:
The failing test `mpi.compareECLFiles_flow+MOD4_GRP` might be unrelated.

bska:
> The failing test `mpi.compareECLFiles_flow+MOD4_GRP` might be unrelated.

It probably is since the build check was run whilst reference data was being updated.  Still, there's no way to know without rerunning the build check so I'll do that now.

bska:
jenkins build this please

blattms:
I might have introduced a slight overhead. Running cpr via the `flow` binary is about 5% slower than using via `flow_blackoil_dunecpr`. I do copy the matrix which should be preventable when running serial. I will investigate this.

Another funny fact: I think I might have (accidentally) saved us some copies/multiple setup (of parallel index set stuff) when using the legacy appoach. At least for my tests the old default ilu0 is fastest (another 3.x %). Will also re-benchmark this to be sure.

| approach  | Total | Assembly | linear solve | linear setup | Linearizations | Netwon Its | Linear Its |
|---------------|--------|--------------|------------------|-----------------|--------------------|----------------|--------------|
| default ilu0 | 633  | 285           | 302            | 12                |  1979              | 1643          | 24508      |
| blackoilcpr | 657  | 267           | 344            | 87                |   2023             |  1689         |   5113       |
| cpr             | 692  | 278           | 365            | 99                |   2023             |  1689         |   5113       |

blattms:
jenkins build this please

blattms:
> I have this implemented in a branch, and working in serial, but in parallel I have some problems 

In that case an early WIP-PR might have made sense (to e.g. prevent others from doing work that is already done elsewhere)

blattms:
Fixed it cpr via ISTLSolverEbos is now on par with the one without. I also did some further optimizations and now the default solver is even faster. We have moved some setup to be done only once and do not copy the matrix for a serial (parallel with ownerFirst_ being true) anymore. The new numbers for Norne on my machine are (use with a grain of salt, did only run once):

| approach  | Total | Assembly | linear solve | linear setup | Linearizations | Netwon Its | Linear Its |
|---------------|--------|--------------|------------------|-----------------|--------------------|----------------|--------------|
| master ilu0| 612  | 265          |  303             |  12               | 1979               | 1643          | 24508      |
| branch ilu0 | 596  | 253           | 294            | 14                |  1979              | 1643          | 24508      |
| blackoilcpr | 657  | 267           | 344            | 87                |   2023             |  1689         |   5113       |
| branch cpr | 658  | 274           | 337            | 86                |   2023             |  1689         |   5113       |

blattms:
benchmark please

blattms:
Did not really test cpr in parallel, yet. But this should be good enough for merging if jenkins agrees.

blattms:
> One could also build it with: make flow_blackoil_dunecpr.

Indeed. Seems to work without the magical defines. I was not aware of this (might have saved me some hours). But I suspect my brain is to limited to remember that and I would miss manuallly recompiling it very often.

blattms:
jenkins run norne please

hnil:
For reference. The think which make the current version slower than when I tested last spring is that the iteration number of the cpr as gone up:

hnil:
It maybe that for norne trueimpes is better, but for know I think quasiimpes is better as default.

hnil:
As I see the difference may be: trueimpes, post and presmoothing bouth for the cpr(twolevel cpr) and the amg and using SeqILU0 instead of ParILU0. In the above I used 2 smoothings for all tings, but an other run with 1 smoothing gave similar performance with a bit more iterations. 
- Current master do not have the option for post and presmoothing of cpr.
(https://github.com/hnil/opm-simulators/tree/flexible_solver_trueimpes/opm/simulators/linalg branch flexible_solver_trueimpes)

blattms:
What does extension `omp_1` mean? Are you limiting openmp (unlike the default)?

blattms:
Tried your settings on my machine. Indeed it is faster than before but still not fast enough:

| approach  | Total | Assembly | linear solve | linear setup | Linearizations | Netwon Its | Linear Its |
|---------------|--------|--------------|------------------|-----------------|--------------------|----------------|--------------|
| master ilu0| 612  | 265          |  303             |  12               | 1979               | 1643          | 24508      |
| branch ilu0 | 596  | 253           | 294            | 14                |  1979              | 1643          | 24508      |
| blackoilcpr | 657  | 267           | 344            | 87                |   2023             |  1689         |   5113       |
| branch cpr | 658  | 274           | 337            | 86                |   2023             |  1689         |   5113       |
| blackoilcpr hnil | 632 | 247     | 338            | 89                 | 1884               | 1550          |   3750      |

Note that comparing with ILU0 used by the flexible solver might miss some optimizations. Hence I always use our current default for comparison.

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 0.998    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.054    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.041    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.113    |



View result details @ https://www.ytelses.com/opm/?page=result&id=737

alfbr:
The improvements on ILU0 are fantastic, and the results on CPR are very promising. Needless to say, I really look forward to getting the improvements here in!

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -340,7 +311,16 @@ class WellModelGhostLastMatrixAdapter : public Dune::AssembledLinearOperator<M,X
               iterations_( 0 ),
               converged_(false)
         {
+#if HAVE_MPI
+            comm_.reset( new communication_type( simulator_.vanguard().grid().comm() ) );
+#endif
             parameters_.template init<TypeTag>();
+            useFlexible_ = parameters_.use_cpr_ || EWOMS_PARAM_IS_SET(TypeTag, std::string, LinearSolverConfiguration);
```
atgeirr, line 318: I think we should eventually retire the --use-cpr parameter, in favour of --linear-solver-configuration, for now this is fine I think.

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -126,25 +128,9 @@ class WellModelMatrixAdapter : public Dune::AssembledLinearOperator<M,X,Y>
...
-          std::any_cast<const ParallelISTLInformation&>( parallelInformation);
-      comm_.reset( new communication_type( info.communicator() ) );
-    }
-#endif
-  }
-  WellModelMatrixAdapter (const M& A,
-                          const M& A_for_precond,
-                          const WellModel& wellMod,
-                          std::shared_ptr<communication_type> comm )
+                          const std::shared_ptr< communication_type >& comm = std::shared_ptr< communication_type >())
```
atgeirr, line 131: This refactoring does not seem related to the flexible solver? I think it looks better and simpler, so I support the refactoring itself.

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -313,6 +277,7 @@ class WellModelGhostLastMatrixAdapter : public Dune::AssembledLinearOperator<M,X
         typedef typename GET_PROP_TYPE(TypeTag, ThreadManager) ThreadManager;
         typedef typename GridView::template Codim<0>::Entity Element;
         typedef typename GET_PROP_TYPE(TypeTag, ElementContext) ElementContext;
+        using FlexibleSolverType = Dune::FlexibleSolver<Matrix, Vector>;
```
atgeirr, line 280: This is correct, but in the near future I would like to make the FlexibleSolver take the operator type as template argument, at which point we will need to switch to storing a pointer to the abstract base class instead.

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -1037,21 +1138,27 @@ class WellModelGhostLastMatrixAdapter : public Dune::AssembledLinearOperator<M,X
         mutable bool converged_;
         std::any parallelInformation_;
 
-        std::unique_ptr<Matrix> matrix_;
+        std::shared_ptr<Matrix> matrix_;
```
atgeirr, line 1142: This is supposed to be a pointer to the simulator's assembly matrix, so should it not just be a naked pointer? Using unique_ptr or shared_ptr implies that this class participates in ownership.

atgeirr:
Apart from the question about using `shared_ptr<Matrix>` vs `Matrix*` above I think this can be merged.

blattms:
> I think we should eventually retire the --use-cpr parameter, in favour of --linear-solver-configuration, for now this is fine I think.
> 

We agree 100% here.

opm/simulators/linalg/ISTLSolverEbos.hpp
blattms: Indeed. That seems historical baggage. Will fix,

opm/simulators/linalg/ISTLSolverEbos.hpp
blattms: Leaving this for another PR

blattms:
jenkins build this please

atgeirr:
This is good, and should be merged when Jenkins is green.

blattms:
Approved, jenkins green, merging.

blattms:
@hnil you might want to prepare a PR using your defaults in [setupPropertyTree](https://github.com/OPM/opm-simulators/blob/654fbf864e9d9458b4aab2dc4ade70b1ceacd1d1/opm/simulators/linalg/setupPropertyTree_impl.hpp#L60)

atgeirr:
This is not working for me in parallel, I have not investigated why. Seems to work well in serial.

blattms:
> This is not working for me in parallel

What of the myriads of options is not working in parallel?

blattms:
If you are talking about cpr, then that is the next step.

> Did not really test cpr in parallel, yet. But this should be good enough for merging if jenkins agrees.

atgeirr:
> What of the myriads of options is not working in parallel?

Flexible solver with CPR. I see that you wrote that earlier

blattms:
> > What of the myriads of options is not working in parallel?
> 
> Flexible solver with CPR. I see that you wrote that earlier

Please recheck. I think that is working since the merge of PR #2614.

PR #2610, opened 2020-05-11
For DUNE 2.7 and up we need to specialize for Opm:MatrixBlock

blattms:
as there already is a specialization for an arbitrary Block
and that one leads to compile errors because we redefine it
in OPM (and miss some other specializations for Opm:MatrixBlock,
e.g. isNumber).

Fixes fallout from #2599

### DIFF
--- a/opm/simulators/linalg/WriteSystemMatrixHelper.hpp
+++ b/opm/simulators/linalg/WriteSystemMatrixHelper.hpp
@@ -22,41 +22,50 @@
 
 #include <dune/istl/matrixmarket.hh>
 
+namespace Opm
+{
+template<typename T, int i, int j>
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

atgeirr:
I do not quite understand why we cannot use the Dune-provided arbitrary-block version for Dune 2.7?

blattms:
Because that one is for the block being a [number (e.g. double).](https://gitlab.dune-project.org/core/dune-istl/-/blob/ea9f0c4c0a19f31978123fba017302703523b365/dune/istl/matrixmarket.hh#L229)

blattms:
this is stalling me. Will selfmerge to keep going.

PR #2611, opened 2020-05-12
Improve convergence msw

totto82:
The motivation for this PR is to improve the convergence on the 9_4E_..MSW test case in opm-models. The number of convergence failures for the well goes from 7466 in master to 0. (with enable-tuning=true)  Needs testing no bigger models.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -2832,8 +2832,8 @@ namespace Opm
             vol_ratio += mix[comp_idx] / b[comp_idx];
         }
 
-        // segment volume
-        const double volume = segmentSet()[seg_idx].volume();
+        // We increase the segment volume with a factor 10 to stabilize the system.
+        const double volume = segmentSet()[seg_idx].volume() * 10;
 
         return volume / vol_ratio;
     }
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1126,8 +1126,8 @@ namespace Opm
             // update the segment pressure
             {
                 const int sign = dwells[seg][SPres] > 0.? 1 : -1;
-                const double dx_limited = sign * std::min(std::abs(dwells[seg][SPres]), relaxation_factor * max_pressure_change);
-                // const double dx_limited = sign * std::min(std::abs(dwells[seg][SPres]) * relaxation_factor, max_pressure_change);
+                //const double dx_limited = sign * std::min(std::abs(dwells[seg][SPres]), relaxation_factor * max_pressure_change);
+                const double dx_limited = sign * std::min(std::abs(dwells[seg][SPres]) * relaxation_factor, max_pressure_change);
                 primary_variables_[seg][SPres] = std::max( old_primary_variables[seg][SPres] - dx_limited, 1e5);
             }
 
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -578,27 +578,25 @@ namespace Opm
                 const double flux_residual = maximum_residual[eq_idx];
                 // TODO: the report can not handle the segment number yet.
 
-                double relax_factor = 1.0;
-                if (relax_tolerance)
-                    relax_factor = 1.e4;
                 if (std::isnan(flux_residual)) {
                     report.setWellFailed({CR::WellFailure::Type::MassBalance, CR::Severity::NotANumber, eq_idx, name()});
                 } else if (flux_residual > param_.max_residual_allowed_) {
                     report.setWellFailed({CR::WellFailure::Type::MassBalance, CR::Severity::TooLarge, eq_idx, name()});
-                } else if (flux_residual > param_.tolerance_wells_ * relax_factor) {
+                } else if (!relax_tolerance && flux_residual > param_.tolerance_wells_) {
+                    report.setWellFailed({CR::WellFailure::Type::MassBalance, CR::Severity::Normal, eq_idx, name()});
+                } else if (flux_residual > param_.relaxed_inner_tolerance_flow_ms_well_) {
                     report.setWellFailed({CR::WellFailure::Type::MassBalance, CR::Severity::Normal, eq_idx, name()});
                 }
             } else { // pressure equation
                 const double pressure_residual = maximum_residual[eq_idx];
                 const int dummy_component = -1;
-                double relax_factor = 1.0;
-                if (relax_tolerance)
-                    relax_factor = 50;
                 if (std::isnan(pressure_residual)) {
                     report.setWellFailed({CR::WellFailure::Type::Pressure, CR::Severity::NotANumber, dummy_component, name()});
                 } else if (std::isinf(pressure_residual)) {
                     report.setWellFailed({CR::WellFailure::Type::Pressure, CR::Severity::TooLarge, dummy_component, name()});
-                } else if (pressure_residual > param_.tolerance_pressure_ms_wells_ * relax_factor) {
+                } else if (!relax_tolerance && pressure_residual > param_.tolerance_pressure_ms_wells_) {
+                    report.setWellFailed({CR::WellFailure::Type::Pressure, CR::Severity::Normal, dummy_component, name()});
+                } else if (pressure_residual > param_.relaxed_inner_tolerance_pressure_ms_well_) {
                     report.setWellFailed({CR::WellFailure::Type::Pressure, CR::Severity::Normal, dummy_component, name()});
                 }
             }
@@ -2386,7 +2384,7 @@ namespace Opm
 
             const BVectorWell dx_well = mswellhelpers::invDXDirect(duneD_, resWell_);
 
-            if (it > 40)
+            if (it > param_.strict_inner_iter_ms_wells_)
                 relax_convergence = true;
 
             const auto report = getWellConvergence(well_state, B_avg, deferred_logger, relax_convergence);
@@ -2412,11 +2410,11 @@ namespace Opm
                     // Still stagnating, terminate iterations if 5 iterations pass.
                     ++stagnate_count;
                     if (stagnate_count == 6) {
-                        sstr << " well " << name() << " observes sever stagnation and/or oscillation. We relax the tolerance and check for convergence. \n";
+                        sstr << " well " << name() << " observes severe stagnation and/or oscillation. We relax the tolerance and check for convergence. \n";
                         const auto reportStag = getWellConvergence(well_state, B_avg, deferred_logger, true);
                         if (reportStag.converged()) {
                             converged = true;
-                            sstr << " well " << name() << " manage to get converged with relaxed tolerances in " << it << " inner iterations";
+                            sstr << " well " << name() << " manages to get converged with relaxed tolerances in " << it << " inner iterations";
                             deferred_logger.debug(sstr.str());
                             return;
                         }
@@ -2509,9 +2507,14 @@ namespace Opm
             // TODO: without considering the efficiencty factor for now
             {
                 const EvalWell segment_surface_volume = getSegmentSurfaceVolume(ebosSimulator, seg);
+
+                // Add a regularization_factor to increase the accumulation term
+                // This will make the system less stiff and help convergence for 
+                // difficult cases
+                const Scalar regularization_factor =  param_.regularization_factor_ms_wells_;
                 // for each component
                 for (int comp_idx = 0; comp_idx < num_components_; ++comp_idx) {
-                    const EvalWell accumulation_term = (segment_surface_volume * surfaceVolumeFraction(seg, comp_idx)
+                    const EvalWell accumulation_term = regularization_factor * (segment_surface_volume * surfaceVolumeFraction(seg, comp_idx)
                                                      - segment_fluid_initial_[seg][comp_idx]) / dt;
 
                     resWell_[seg][comp_idx] += accumulation_term.value();
@@ -2833,7 +2836,7 @@ namespace Opm
         }
 
         // We increase the segment volume with a factor 10 to stabilize the system.
-        const double volume = segmentSet()[seg_idx].volume() * 10;
+        const double volume = segmentSet()[seg_idx].volume();
 
         return volume / vol_ratio;
     }
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -2445,7 +2445,7 @@ namespace Opm
             std::ostringstream sstr;
             sstr << " well " << name() << " manage to get converged within " << it << " inner iterations \n";
             if (relax_convergence)
-                sstr << "A relaxed tolerance is used after 40 iterations";
+                sstr << "A relaxed tolerance is used after "<< param_.strict_inner_iter_ms_wells_ << " iterations";
             deferred_logger.debug(sstr.str());
         } else {
             std::ostringstream sstr;
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -536,7 +536,7 @@ namespace Opm
     template <typename TypeTag>
     ConvergenceReport
     MultisegmentWell<TypeTag>::
-    getWellConvergence(const WellState& well_state, const std::vector<double>& B_avg, Opm::DeferredLogger& deferred_logger) const
+    getWellConvergence(const WellState& well_state, const std::vector<double>& B_avg, Opm::DeferredLogger& deferred_logger, const bool relax_tolerance) const
     {
         assert(int(B_avg.size()) == num_components_);
 
@@ -577,21 +577,28 @@ namespace Opm
             if (eq_idx < num_components_) { // phase or component mass equations
                 const double flux_residual = maximum_residual[eq_idx];
                 // TODO: the report can not handle the segment number yet.
+
+                double relax_factor = 1.0;
+                if (relax_tolerance)
+                    relax_factor = 1.e4;
                 if (std::isnan(flux_residual)) {
                     report.setWellFailed({CR::WellFailure::Type::MassBalance, CR::Severity::NotANumber, eq_idx, name()});
                 } else if (flux_residual > param_.max_residual_allowed_) {
                     report.setWellFailed({CR::WellFailure::Type::MassBalance, CR::Severity::TooLarge, eq_idx, name()});
-                } else if (flux_residual > param_.tolerance_wells_) {
+                } else if (flux_residual > param_.tolerance_wells_ * relax_factor) {
                     report.setWellFailed({CR::WellFailure::Type::MassBalance, CR::Severity::Normal, eq_idx, name()});
                 }
             } else { // pressure equation
                 const double pressure_residual = maximum_residual[eq_idx];
                 const int dummy_component = -1;
+                double relax_factor = 1.0;
+                if (relax_tolerance)
+                    relax_factor = 50;
                 if (std::isnan(pressure_residual)) {
                     report.setWellFailed({CR::WellFailure::Type::Pressure, CR::Severity::NotANumber, dummy_component, name()});
                 } else if (std::isinf(pressure_residual)) {
                     report.setWellFailed({CR::WellFailure::Type::Pressure, CR::Severity::TooLarge, dummy_component, name()});
-                } else if (pressure_residual > param_.tolerance_pressure_ms_wells_) {
+                } else if (pressure_residual > param_.tolerance_pressure_ms_wells_ * relax_factor) {
                     report.setWellFailed({CR::WellFailure::Type::Pressure, CR::Severity::Normal, dummy_component, name()});
                 }
             }
@@ -2369,17 +2376,20 @@ namespace Opm
         int it = 0;
         // relaxation factor
         double relaxation_factor = 1.;
-        const double min_relaxation_factor = 0.2;
+        const double min_relaxation_factor = 0.6;
         bool converged = false;
         int stagnate_count = 0;
+        bool relax_convergence = false;
         for (; it < max_iter_number; ++it, ++debug_cost_counter_) {
 
             assembleWellEqWithoutIteration(ebosSimulator, dt, inj_controls, prod_controls, well_state, deferred_logger);
 
             const BVectorWell dx_well = mswellhelpers::invDXDirect(duneD_, resWell_);
 
+            if (it > 40)
+                relax_convergence = true;
 
-            const auto report = getWellConvergence(well_state, B_avg, deferred_logger);
+            const auto report = getWellConvergence(well_state, B_avg, deferred_logger, relax_convergence);
             if (report.converged()) {
                 converged = true;
                 break;
@@ -2395,19 +2405,22 @@ namespace Opm
             // TODO: maybe we should have more sophiscated strategy to recover the relaxation factor,
             // for example, to recover it to be bigger
 
-            if (!is_stagnate) {
-                stagnate_count = 0;
-            }
             if (is_oscillate || is_stagnate) {
                 // HACK!
-                if (is_stagnate && relaxation_factor == min_relaxation_factor) {
+                std::ostringstream sstr;
+                if (relaxation_factor == min_relaxation_factor) {
                     // Still stagnating, terminate iterations if 5 iterations pass.
                     ++stagnate_count;
-                    if (stagnate_count == 5) {
-                        // break;
+                    if (stagnate_count == 6) {
+                        sstr << " well " << name() << " observes sever stagnation and/or oscillation. We relax the tolerance and check for convergence. \n";
+                        const auto reportStag = getWellConvergence(well_state, B_avg, deferred_logger, true);
+                        if (reportStag.converged()) {
+                            converged = true;
+                            sstr << " well " << name() << " manage to get converged with relaxed tolerances in " << it << " inner iterations";
+                            deferred_logger.debug(sstr.str());
+                            return;
+                        }
                     }
-                } else {
-                    stagnate_count = 0;
                 }
 
                 // a factor value to reduce the relaxation_factor
@@ -2415,7 +2428,6 @@ namespace Opm
                 relaxation_factor = std::max(relaxation_factor * reduction_mutliplier, min_relaxation_factor);
 
                 // debug output
-                std::ostringstream sstr;
                 if (is_stagnate) {
                     sstr << " well " << name() << " observes stagnation in inner iteration " << it << "\n";
 
@@ -2433,7 +2445,9 @@ namespace Opm
         // TODO: we should decide whether to keep the updated well_state, or recover to use the old well_state
         if (converged) {
             std::ostringstream sstr;
-            sstr << " well " << name() << " manage to get converged within " << it << " inner iterations";
+            sstr << " well " << name() << " manage to get converged within " << it << " inner iterations \n";
+            if (relax_convergence)
+                sstr << "A relaxed tolerance is used after 40 iterations";
             deferred_logger.debug(sstr.str());
         } else {
             std::ostringstream sstr;
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -2088,7 +2088,7 @@ namespace Opm
     getWellConvergence(const WellState& well_state,
                        const std::vector<double>& B_avg,
                        Opm::DeferredLogger& deferred_logger,
-                       const bool relax_tolerance) const
+                       const bool /*relax_tolerance*/) const
     {
         // the following implementation assume that the polymer is always after the w-o-g phases
         // For the polymer, energy and foam cases, there is one more mass balance equations of reservoir than wells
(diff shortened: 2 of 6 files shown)


totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2758,8 +2772,8 @@ namespace Opm
             vol_ratio += mix[comp_idx] / b[comp_idx];
         }
 
-        // segment volume
-        const double volume = segmentSet()[seg_idx].volume();
+        // We increase the segment volume with a factor 10 to stabilize the system.
+        const double volume = segmentSet()[seg_idx].volume() * 10;
```
GitPaean, line 2776: Hi, different from the standard well, the volume for MSW has physical meaning. I am not sure we should change the equation here.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -577,21 +577,28 @@ namespace Opm
             if (eq_idx < num_components_) { // phase or component mass equations
                 const double flux_residual = maximum_residual[eq_idx];
                 // TODO: the report can not handle the segment number yet.
+
+                double relax_factor = 1.0;
+                if (relax_tolerance)
+                    relax_factor = 1.e4;
```
GitPaean, line 583: With this, basically we do not care much about the flux residual. Similar to that we ignore the CNV checking after 8 iterations.

opm/simulators/wells/MultisegmentWell_impl.hpp
hnil: Would it be an idea start volume factor which is adjusted to 1 at convergence?

opm/simulators/wells/MultisegmentWell_impl.hpp
hnil: Does this mean accept anything? If this "tunings" is a good idea, I think they should be paramters forexample --mswells-relaxed-convergence-factor=1e4 --mswells-volume-factor=10, and that the default values are 1.0 for bouth.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: This function is only used for the accumulation term. It really helps signification on the convergence.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: Agree @hnil. We just need to test this on more models first to see if it is a good idea. We may need to change the tolerances a bit after testing. The tolerances here are just tuned to remove all convergence issues for MSW on the 9_4E_...MSW model. If we can accept 10+ convergence issues we can tighten these numbers.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: > Would it be an idea start volume factor which is adjusted to 1 at convergence?

If we see any negative effect on the results with this that could be good idea.

totto82:
I hope this improves the stability of the MSW testing. I therefore plan to run an update_data on this and then pull it locally to see if less tests fails on my machine.

totto82:

opm/simulators/wells/MultisegmentWell_impl.hpp
GitPaean: > Does this mean accept anything? 

It will be tolerance within 86400. m^3/day for the flux tolerance. It is not anything, but almost. The flux residual with this value will not be right.

opm/simulators/wells/MultisegmentWell_impl.hpp
GitPaean: As @hnil  suggests, you might be able to use this as a intermediate term for numerical reason. But the final result should converge with the correct equation, otherwise it might be dangerous for future. When the transition behavoir within the wellbore matters for some cases, this can lead to wrong result.

opm/simulators/wells/MultisegmentWell_impl.hpp
GitPaean: > If this "tunings" is a good idea, I think they should be paramters forexample --mswells-relaxed-convergence-factor=1e4 --mswells-volume-factor=10, and that the default values are 1.0 for both.

This is a good suggestion. Then people can create tests with running time arguments without changing default behavior.

totto82:
> I hope this improves the stability of the MSW testing. I therefore plan to run an update_data on this and then pull it locally to see if less tests fails on my machine.

All tests with MSW still fails on my machine. i.e. we need to continue the hunt.

alfbr:
> All tests with MSW still fails on my machine. i.e. we need to continue the hunt.

Some problems are indeed hard to resolve. I will test a large model over here, compiling now, so should have complete results tomorrow, partial results by end of working day.

bska:
@totto82 said:
> > I hope this improves the stability of the MSW testing. I therefore plan to run an update_data on this and then pull it locally to see if less tests fails on my machine.
> 
> All tests with MSW still fails on my machine. i.e. we need to continue the hunt.

Does that mean that the new candidate reference data in OPM/opm-tests#322 should be discarded, meaning we can/should close the update PR?

totto82:
> Does that mean that the new candidate reference data in OPM/opm-tests#322 should be discarded, meaning we can/should close the update PR?

Based on the comments about this PR probably needs some polishing and a new update_data later. Please feel free to remove it. Or we can just update it when this PR is ready.

GitPaean:
> All tests with MSW still fails on my machine. i.e. we need to continue the hunt.

I believe the problem is in the linear solution stage.  It is good we can investigate with simpler model to check whether we get exact jacobian, residual and linear solutions. 

For some situations, it might not be claimed as failure, but the results are different, but within the comparison tolerance.  This kind of difference will make the comparison a little different among different models, and difficult to predicate. 

The other options are that our regression comparison criterion is much stricter than visual-inspection. Maybe we can relax our regression comparison criterion in some delicate way, so the comparison is good-enough.  With some randomness going on, not sure the currently stricter comparison tolerance will protect us much. 

But for Jenkins, it might be good to keep the strict tolerance because the platform is relatively stable there.

alfbr:
> Does that mean that the new candidate reference data in OPM/opm-tests#322 should be discarded, meaning we can/should close the update PR?

This PR so far seems to be an improvement, so we may want to merge it, and then we need to update the data to get Jenkins green. The tests fail locally, which is the same situation as we have had with master. There is an inherent instability in the multi-segment code. I do not think this PR makes it worse.

GitPaean:
Other things might matter is the upwinding-direciton determination, which makes `-1e-16` and `1e-16` give different upwinding directions, while might change the later calculation significantly.

totto82:
> Other things might matter is the upwinding-direciton determination, which makes -1e-16 and 1e-16 give different upwinding directions, while might change the later calculation significantly.

Good point. I can run some tests to see if the upwind direction oscillates in some of the test cases.

bska:
> > Does that mean that the new candidate reference data in [OPM/opm-tests#322](https://github.com/OPM/opm-tests/pull/322) should be discarded, meaning we can/should close the update PR?
> 
> This PR so far seems to be an improvement, so we may want to merge it, and then we need to update the data to get Jenkins green.

That's fair.  I was mostly asking out of self interest.  I'd like to merge OPM/opm-common#1770 "soon" and that PR will also require an update to the reference data&mdash;mostly to account for the property derivatives with respect to saturation changing as a result of the critical saturations being altered.  Since our current system for handling updates to the reference data always seems to fail in some way and require special case handling when there are multiple data updates in flight I'd like to avoid getting into that situation.

Still, there's no immediate rush merging OPM/opm-common#1770 so if this PR is slated for merging some time this week then that's fine by me.

alfbr:
At this point the update data was to check if this fixed the local test failures. Re-running it is easy, so no reason to hold other PR's.

totto82:
We have now tested this for several field cases and smaller test problems and the results are very good both in terms of accuracy and convergence. I will remove the testing tag and try to get i merged.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: The tolerance relaxation is only in the inner iterations. The strict tolerance is still used when getWellConvergence is called during the main newton loop. The change of tolerance here should therefore not effect the end results only the road towards it. I don't think we can expect the MSW to be able to converge fully early in the newton iterations before the near well solution starts to settle a bit.

alfbr:
I have tested this PR over here, and support merging the functionality.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: I agree that the factor 10 seems very hacky but I can not image the physical meaning of the accumulation term to be significant even for MSW. We usually take timesteps in the order of days and the volume of a well is very small. I have atleast not been able to find a case where this factor changes anything significant on the results. A motivation for this "hack" is to avoid cases where you get epsilon influx in the last segment of an injector. With this hack the well has a bit more room for accumulating this extra fluid instead of pushing it upwards. 

@GitPaean Do you have a case where you might suspect this term to have any effect on the results? 

To facilitate further testing I will make the hard-coded number run-time parameters.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1059,8 +1066,8 @@ namespace Opm
             // update the segment pressure
             {
                 const int sign = dwells[seg][SPres] > 0.? 1 : -1;
-                const double dx_limited = sign * std::min(std::abs(dwells[seg][SPres]), relaxation_factor * max_pressure_change);
-                // const double dx_limited = sign * std::min(std::abs(dwells[seg][SPres]) * relaxation_factor, max_pressure_change);
+                //const double dx_limited = sign * std::min(std::abs(dwells[seg][SPres]), relaxation_factor * max_pressure_change);
+                const double dx_limited = sign * std::min(std::abs(dwells[seg][SPres]) * relaxation_factor, max_pressure_change);
```
atgeirr, line 1070: This looks like a pure bugfix? Is there any reason to keep the out-commented line?

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -2087,7 +2087,8 @@ namespace Opm
     StandardWell<TypeTag>::
     getWellConvergence(const WellState& well_state,
                        const std::vector<double>& B_avg,
-                       Opm::DeferredLogger& deferred_logger) const
+                       Opm::DeferredLogger& deferred_logger,
+                       const bool relax_tolerance) const
```
atgeirr, line 2091: Should be marked as unused argument to avoid warnings. (Or implement the relaxation?)

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2309,17 +2316,20 @@ namespace Opm
...
         bool converged = false;
         int stagnate_count = 0;
+        bool relax_convergence = false;
         for (; it < max_iter_number; ++it, ++debug_cost_counter_) {
 
             assembleWellEqWithoutIteration(ebosSimulator, dt, inj_controls, prod_controls, well_state, deferred_logger);
 
             const BVectorWell dx_well = mswellhelpers::invDXDirect(duneD_, resWell_);
 
+            if (it > 40)
```
atgeirr, line 2329: Nitpick: braces please... (also line 2389).

Alternative (I like it, but not all might):
```
relax_convergence = it > 40;
```

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2335,27 +2345,29 @@ namespace Opm
...
                 // HACK!
-                if (is_stagnate && relaxation_factor == min_relaxation_factor) {
+                std::ostringstream sstr;
+                if (relaxation_factor == min_relaxation_factor) {
                     // Still stagnating, terminate iterations if 5 iterations pass.
                     ++stagnate_count;
-                    if (stagnate_count == 5) {
-                        // break;
+                    if (stagnate_count == 6) {
+                        sstr << " well " << name() << " observes sever stagnation and/or oscillation. We relax the tolerance and check for convergence. \n";
```
atgeirr, line 2355: Typo, "sever" -> "severe".

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2335,27 +2345,29 @@ namespace Opm
...
+                        const auto reportStag = getWellConvergence(well_state, B_avg, deferred_logger, true);
+                        if (reportStag.converged()) {
+                            converged = true;
+                            sstr << " well " << name() << " manage to get converged with relaxed tolerances in " << it << " inner iterations";
```
atgeirr, line 2359: "manage" -> "manages", also on line 2388 (not a new flaw)

opm/simulators/wells/MultisegmentWell_impl.hpp
atgeirr: This makes me nervous, too. I like the suggestion to start off larger, then tighten it, but I don't see how we should approach it: "if this is the last iteration, use the real volume" fails since we never know if it is the last iteration...

opm/simulators/wells/MultisegmentWell_impl.hpp
atgeirr: I think this is done in a good way, and agree that the main convergence check will never see any relaxation.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: No I don't think so. The out-comment lines where kept for the other cases. I can remove them all.

GitPaean:
> This makes me nervous, too. I like the suggestion to start off larger, then tighten it, but I don't see how we should approach it: "if this is the last iteration, use the real volume" fails since we never know if it is the last iteration...

I do not know why I could not reply the comment directly.  How about we using the relax way to get convergence first then switching back to the original way to make sure it is converged with the original equations, also some reasonable tolerances. 

Or, or, at least, we should have the original equations and tolerances honored for the final convergence of the time step.  Otherwise, if we have the equations changed, in the future, we might run into really tedious debugging for different results.

opm/simulators/wells/MultisegmentWell_impl.hpp
GitPaean: It is not a bug. At the old time, first version MSW that ran through model 2, the original code helped us to go through difficult cases.  Now, a lot of things are improved, situation can be different.

totto82:
> I think this should probably be merged, but worry about the segment volume hack. I assume that it does not matter much, as testing has shown ok match, but then I wonder why? If this improves things, why not multiply by 100... There might be some tradeoff here between an easily solvable system, and accuracy, but I do not know how we could find the proper tradeoff (likely varying a lot between cases).
> 
> I have a loose idea: pass a flag to the routine indicating the "volume exaggeration factor", use 10.0 for the first few iterations, reducing it to 1.0 over (say) 5 iterations. Then also require a minimum of 6 iterations to ensure we get to the correct value before terminating.

We have not evaluated the 10 hack separably for the bigger models. It may not be necessary. It is however needed for the 9_4E_..MSW due to the crossflow issue I described earlier. Do we need to split this hack into a separate PR? This will trigger some new testing on the big models to isolate the  effect of this particular change. Or we can just merge it as is and remove/change it if we come up with examples where it fails.

totto82:
> I like the suggestion to start off larger, then tighten it, but I don't see how we should approach it: "if this is the last iteration, use the real volume" fails since we never know if it is the last iteration...

I think such an approach will be difficult. The volume used for the initial accumulation term at the beginning of a timestep and the final one need to be the same. Otherwise mass conservation is not fulfilled. We could move the factor from the volume calculations to the equation to make the regularization more explicit . This may also help to avoid miss-use if we later want to use getSegmentSurfaceVolume() in another place in the code for a different purpose.

atgeirr:
> We could move the factor from the volume calculations to the equation to make the regularization more explicit

I think that is a very good idea.

totto82:
I have added run-time parameters with default values as in the original PR.

totto82:
jenkins build this please

GitPaean:
Just made a test  run for `9_4E_WINJ_GINJ_GUIDERATE_MSW.DATA`.  

Without the factor `10` for segment volume
```
Number of MPI processes:      1
Threads per MPI process:      2
Total time (seconds):         9.9641
Solver time (seconds):        9.95097
 Assembly time (seconds):     1.54421 (Failed: 0; 0%)
 Linear solve time (seconds): 3.73809 (Failed: 0; 0%)
 Linear solve setup time (seconds): 9.543e-05 (Failed: 0; 0%)
 Update time (seconds):       0.116148 (Failed: 0; 0%)
 Output write time (seconds): 0.0374031
Overall Well Iterations:      0 (Failed: 0; -nan%)
Overall Linearizations:       219 (Failed: 0; 0%)
Overall Newton Iterations:    191 (Failed: 0; 0%)
Overall Linear Iterations:    3065 (Failed: 0; 0%)
```

With the factor `10.0` for segment volume
```
Number of MPI processes:      1
Threads per MPI process:      2
Total time (seconds):         9.50761
Solver time (seconds):        9.49395
 Assembly time (seconds):     1.52628 (Failed: 0; 0%)
 Linear solve time (seconds): 3.69769 (Failed: 0; 0%)
 Linear solve setup time (seconds): 9.012e-05 (Failed: 0; 0%)
 Update time (seconds):       0.113016 (Failed: 0; 0%)
 Output write time (seconds): 0.0374894
Overall Well Iterations:      0 (Failed: 0; -nan%)
Overall Linearizations:       217 (Failed: 0; 0%)
Overall Newton Iterations:    189 (Failed: 0; 0%)
Overall Linear Iterations:    3029 (Failed: 0; 0%)
```


Without the factor 10, there is some inner iterations not get converged, but we will end up using the solution anyway. 

With the factor 10, it gets converged but with slightly different equations. 

Not sure whether the measure is overkill considering its potential impacts.

totto82:
> Just made a test run for 9_4E_WINJ_GINJ_GUIDERATE_MSW.DATA.

Please run it with --enable-tuning=true or restrict timesteps to 1 day.

totto82:
I have grouped test failures comes in four categories. 
1) very small differences < 1% ( = ok)
2) differences in WWPR/WWPT in the order of 1% ( = WWPT?)
3) differences due to less number of timesteps.  ( X != Y, X>Y)
4) differences due to more number of timesteps.  ( X != Y, X<Y)

1) Some differences are expeced
2) Manually inspection shows that the differences in water production are very very small. If you close in it seems to agree more with Eclipse. (Could be confidence) 
image

3) Less iterations is fine

4) Since we sometimes solve the inner MSW iteration less accurate it may lead to problems for the Newton solver. I will try to change the tuning a bit to see if I can avoid this. 

Summary of test failures: 

mpi.compareECLFiles_flow+3D_MSW WWPT?
mpi.compareECLFiles_flow+2D_H__ WWPT?
mpi.compareECLFiles_flow+MSW_MODEL_1 WWPT?
mpi.compareECLFiles_flow+FAULTS_MODEL_1 WWPT?
mpi.compareECLFiles_flow+0A1_GRCTRL_LRAT_ORAT_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+0A2_GRCTRL_LRAT_ORAT_GGR_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+0A3_GRCTRL_LRAT_LRAT_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+0A4_GRCTRL_LRAT_LRAT_GGR_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+9_1A_DEPL_MAX_RATE_MIN_BHP_MSW (18 != 17) 
mpi.compareECLFiles_flow+9_1B_DEPL_MAX_RATE_MIN_THP_MSW (29 != 28) 
mpi.compareECLFiles_flow+9_2A_DEPL_GCONPROD_1L_MSW WWPT?
mpi.compareECLFiles_flow+9_2B_DEPL_GCONPROD_2L_MSW (29 != 28)
mpi.compareECLFiles_flow+9_3A_GINJ_REIN-G_MSW (28 != 29)
mpi.compareECLFiles_flow+9_3B_GINJ_GAS_EXPORT_MSW (28 != 29)
mpi.compareECLFiles_flow+9_3E_GAS_MIN_EXPORT_MSW (26 != 24)
mpi.compareECLFiles_flow+9_3D_GINJ_GAS_MAX_EXPORT_MSW WWPT?
mpi.compareECLFiles_flow+9_3C_GINJ_GAS_GCONSUMP_MSW convergence (36 != 37)
mpi.compareECLFiles_flow+9_4A_WINJ_MAXWRATES_MAXBHP_GCONPROD_1L_MSW ok
mpi.compareECLFiles_flow+9_4B_WINJ_VREP-W_MSW (28 != 30)
mpi.compareECLFiles_flow+9_4C_WINJ_GINJ_VREP-W_REIN-G_MSW ok
mpi.compareECLFiles_flow+TEST_WSEGSICD ok
mpi.compareECLFiles_flow+9_4D_WINJ_GINJ_GAS_EXPORT_MSW (29 != 28)
mpi.compareECLFiles_flow+MOD4_GRP (83 != 82)

GitPaean:
> Please run it with --enable-tuning=true or restrict timesteps to 1 day.

I see your point.  The performance summary information is almost the same, so they are not posted here. 

It is interesting, the `INJ1` is having problem in getting converged with smaller time step (1 day) here, not with the bigger time step (16 days). I do not have an explanation for that, which looks like a little random.  And also there is not chopping of time step for this, I guess time step smaller than 1 day is not used anymore?

I saw you added running parameters for the relaxations. Maybe you can use relaxed parameters for this specific case for regression testing. For other cases are not having problems, you do not use the relaxations. 

This can also simplify the procedure of the `update_data`. Because the reference results for other cases are not changed.

totto82:
> I will try to change the tuning a bit to see if I can avoid this. 

Unfortunately, all the three cases where Jenkins fails due to one extra timesteps does not fail locally on my machine. This means that using these tests to improve the tuning is very unpractical. From my side I conclude the testing and will recomend an update data.

totto82:
> It is interesting, the INJ1 is having problem in getting converged with smaller time step (1 day) here, not with the bigger time step (16 days).

From a numerical point of view it may be harder to solve the 1 day case. Large timesteps may smooth things out. In particular for the well model.  

> I saw you added running parameters for the relaxations. Maybe you can use relaxed parameters for this specific case for regression testing. For other cases are not having problems, you do not use the relaxations. 

The motivation for the changes are improvements on several field models not the small test case. The test case 9_4E with 1 day timesteps seemed to point out some issue in the code related to injectors with crossflow.

GitPaean:
> From a numerical point of view it may be harder to solve the 1 day case. Large timesteps may smooth things out. In particular for the well model.

Yes. That was also my guessing, but it can also something imperfect in our design and implementation. The baseline for perfect design is that we should always solve the problem. 

> The motivation for the changes are improvements on several field models not the small test case. The test case 9_4E with 1 day timesteps seemed to point out some issue in the code related to injectors with crossflow.

I do not know about that. From the previous discussion, I thought it is part of the efforts to improve the testing.

totto82:
> Yes. That was also my guessing, but it can also something imperfect in our design and implementation. The baseline for perfect design is that we should always solve the problem.

I have tried another approach that seems to stabilize the system. I scale the pressure variable. With this I am able to run the 9_4E case without closing the injector even if I don't regularize the system. I have not tested this on any of the bigger models but the smaller ones I have tested seems to be fine. I will try to Jenkins it and also update the data to see how this effect the stability of the testing.

totto82:

GitPaean:
> I have tried another approach that seems to stabilize the system. I scale the pressure variable.

Thanks for bringing this up. I know some people have been doing this, also E100. If it is a good idea, maybe we should do it in a systematic way, also including both the well models and also the reservoir equations.

atgeirr:
jenkins build this opm-tests=322 please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2881,14 +2881,14 @@ namespace Opm
...
-            sum += residuals[SPres] / pressure_tolerance;
+        if (residuals[SPres]*scalingFactorPressure > pressure_tolerance) {
+            sum += residuals[SPres] / pressure_tolerance * scalingFactorPressure;
             ++count;
         }
 
         const double control_tolerance = getControlTolerance(well_state, deferred_logger);
-        if (residuals[SPres + 1] > control_tolerance) {
-            sum += residuals[SPres + 1] / control_tolerance;
+        if (residuals[SPres + 1]*scalingFactorPressure > control_tolerance) {
```
atgeirr, line 2890: This part makes me skeptical: It looks like we scale the control equation as if it is a pressure equation, also for the case when it is a rate equation? Maybe getControlTolerance() is a better place to handle that?

totto82: The scaling of the control equation was not intended. I will remove it and update the PR.

totto82:
> If turns out to be a good idea, maybe we should do it in a systematic way, also including both the well models and also the reservoir equations.

I agree. But preferably not in this PR. The reservoar equation already supports some scaling. The scaling of the pressure equation and variables for MSW is crucial so I want this merged first.

totto82:
Currently equation and variable is scaled with the same factor. I have not tried to optimize the scaling factor(s) but +-one orders does not seem to alter the results much. The 1e5 factor seems to be ok. If necessary we can fine tune this later. We have not jet solved the instability of the MSW tests. There may be other issues that needs to be solved first.....

totto82:
jenkins build this please

totto82:
The scaling of the pressures had some unwanted effects on the results for some cases. It will need more investigation. I will make a separate PR for that part.

totto82:
jenkins build this please

GitPaean:
> The scaling of the pressures had some unwanted effects on the results for some cases. It will need more investigation. I will make a separate PR for that part.

Any special features about these cases?

totto82:
jenkins build this please

totto82:
Summary test failures:
mpi.compareECLFiles_flow+0A1_GRCTRL_LRAT_ORAT_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+0A2_GRCTRL_LRAT_ORAT_GGR_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+9_1A_DEPL_MAX_RATE_MIN_BHP_MSW (18 != 16)
mpi.compareECLFiles_flow+9_1B_DEPL_MAX_RATE_MIN_THP_MSW (29 != 28)
mpi.compareECLFiles_flow+9_2A_DEPL_GCONPROD_1L_MSW ok
mpi.compareECLFiles_flow+9_2B_DEPL_GCONPROD_2L_MSW (29 != 28)
mpi.compareECLFiles_flow+9_3A_GINJ_REIN-G_MSW ok
mpi.compareECLFiles_flow+9_3B_GINJ_GAS_EXPORT_MSW ok
mpi.compareECLFiles_flow+9_3E_GAS_MIN_EXPORT_MSW (26 != 24)
mpi.compareECLFiles_flow+9_3C_GINJ_GAS_GCONSUMP_MSW (36 != 34)
mpi.compareECLFiles_flow+9_3D_GINJ_GAS_MAX_EXPORT_MSW ok
mpi.compareECLFiles_flow+9_4A_WINJ_MAXWRATES_MAXBHP_GCONPROD_1L_MSW ok
mpi.compareECLFiles_flow+9_4B_WINJ_VREP-W_MSW (28 != 30)
mpi.compareECLFiles_flow+9_4C_WINJ_GINJ_VREP-W_REIN-G_MSW (29 != 28)
mpi.compareECLFiles_flow+9_4D_WINJ_GINJ_GAS_EXPORT_MSW (29 != 28)

totto82:
> Any special features about these cases?

MOD4_GRP starts osculating between GRUP and BHP for one or two timesteps. It does not change the results much, but it makes me a bit worried that there may either a bug in my implementation or that the scaling factors I use is not optimal.

atgeirr:
> I still think this PR is useful as is since it reduces MSW iterations and makes a parameter of the regularization term.

I agree. So we should do an update_data, then? Or do you think it is not quite mature for merging yet?

totto82:
I have removed the not quite mature part to make it easier to merge. From my part this is ready

GitPaean:
> MOD4_GRP starts osculating between GRUP and BHP for one or two timesteps.

This is probably due to the group control. But your suspicion is also valid.

totto82:

atgeirr:
jenkins build this opm-tests=322 please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2433,7 +2443,9 @@ namespace Opm
         // TODO: we should decide whether to keep the updated well_state, or recover to use the old well_state
         if (converged) {
             std::ostringstream sstr;
-            sstr << " well " << name() << " manage to get converged within " << it << " inner iterations";
+            sstr << " well " << name() << " manage to get converged within " << it << " inner iterations \n";
+            if (relax_convergence)
+                sstr << "A relaxed tolerance is used after 40 iterations";
```
atgeirr, line 2448: This should be "after [StrictInnerIterMsWells] iterations" now I guess?

totto82: I will update it asap

atgeirr:
jenkins build this opm-tests=322 please

PR #2612, opened 2020-05-12
Make only rank zero honor verbosity in flexible solvers.

blattms:
Previously all the ranks print linear solver statistics in verbose mode which cluttered the output in parallel runs. Now only rank 0 will print like it should be.


blattms:
jenkins build this please

atgeirr:
Good, merging.

PR #2613, opened 2020-05-13
Use correct matrix for scaling and get rid of unused preconditioner matrix member

blattms:
There was a time when we supported a different matrix for the preconditioner but that option must have been silently dropped somewhen. Still there were several variables in the code for it present which are dropped with this PR.

In addition I discovered that we sometimes use the wrong matrix (derefernced nullptr) for scaling and fixed.

Based on PR #2608


blattms:
Is there a way to force github to refresh a PR? It shows commits here that are already merged.

atgeirr:
You have to rebase it on top of master.

blattms:
Thanks. Did not think about that and the merge commit of the other PR that was missing here.

blattms:
jenkins build this please

atgeirr:
I think the intention was to be able to pass the matrix-with-well-contributions to the preconditioner, while still using the wells-as-operators approach for the iterative solver. The feature is dead, since all call sites now passed the same matrix for both arguments. I am fine with killing it, we would probably do it slightly different if we needed the feature again.

Can be merged when green.

PR #2617, opened 2020-05-13
Fix Gpu fallout pr 2612

blattms:
Due to the changes in #2613 we experienced linker errors as the matrix passed to the `BdaBridge` was made const there and these functions have not been instantiated. But the bridge needs a mutable matrix as it overwriter zero diagonal values. This PR uses a const cast to pass a mutable matrix to the BdaBridge to fix this. In addition we simplify the explcite template instantiation a bit with a macro.


blattms:
It is a bit surprising that there was no issue when we passed the result of `std::unique_ptr::get()` to the method in  a const function.

blattms:
jenkins build this please

PR #2618, opened 2020-05-13
Use "trueimpes" for weight_type for configuration cpr_trueimpes.

blattms:
Previously, we always used quasiimpes.


blattms:
jenkins build this please

atgeirr:
I am embarrassed that I never noticed... did not try to run with trueimpes apparently.

PR #2619, opened 2020-05-13
Refactor `execute()` in `FlowMainEbos.hpp`

hakonhagland:
A resubmission of commit 77a151b in PR #2403 and PR #2440 to work with the current master.

Currently, `execute()` calls `runSimulator()` to run the simulation. When the Python `step_init()` is implemented (in a later commit), it will instead call an `executeStepInit()` method that will need to do the same initialization as in the current `execute()` except that it should call a `runSimulatorInit()` method instead of `runSimulator()`. In order to avoid code duplication for `execute()` and `executeStepInit()`, `execute()` is here refactored into an `execute_()` method.

### DIFF
--- a/opm/simulators/flow/FlowMainEbos.hpp
+++ b/opm/simulators/flow/FlowMainEbos.hpp
@@ -256,43 +256,8 @@ namespace Opm
         /// input.
         int execute(int argc, char** argv, bool output_cout, bool output_to_files)
         {
-            try {
-                // deal with some administrative boilerplate
-
-                int status = setupParameters_(argc, argv);
-                if (status)
-                    return status;
-
-                setupParallelism();
-                setupEbosSimulator();
-                runDiagnostics(output_cout);
-                createSimulator();
-
-                // do the actual work
-                int retval = runSimulator(output_cout);
-
-                // clean up
-                mergeParallelLogFiles(output_to_files);
-
-                return retval;
-            }
-            catch (const std::exception& e) {
-                std::ostringstream message;
-                message  << "Program threw an exception: " << e.what();
-
-                if (output_cout) {
-                    // in some cases exceptions are thrown before the logging system is set
-                    // up.
-                    if (OpmLog::hasBackend("STREAMLOG")) {
-                        OpmLog::error(message.str());
-                    }
-                    else {
-                        std::cout << message.str() << "\n";
-                    }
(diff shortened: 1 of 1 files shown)


hakonhagland:
jenkins build this please

atgeirr:
I have to admit this looks a bit overcomplicated, what is wrong about splitting into "setup", "run" and "cleanup" functions, then calling them from Python separately?

hakonhagland:
> wrong about splitting into "setup", "run" and "cleanup" functions, then calling them from Python separately?

@atgeirr I was trying to keep the original structure of the `execute()` method:

```
int execute( ... )
{
    try {
        // call methods for setup
        // ...

        int retval = runSimulator();

        // call methods for clean up
        // ....
        
        return retval;
     }
     catch (const std::exception& e) {
         // handle exceptions
     }
}
```
If we try the other approach as you suggest, then we instead write `execute()` llike this (in pseudo code):
```
int execute( ... )
{
    try {
        setup();
        int retval = runSimulator(output_cout);
        cleanup();
        return retval;
    }
    catch (const std::exception& e) {
        exeception_handler();
    }
}
```
Then we need to introduce additional python wrappers:
```
int setup_python()
{
    try {
        setup();
        // Note: We do not want to call runSimulator() here since 
        // we only want to initialize
        runSimulatorInit();  
    }
    catch (const std::exception& e) {
        exception_handler();
    }
}

int step_python()
{
    try {
        // Note: We do not want to call runSimulator() here since 
        // we only want to do a single report step
        runSimulatorStep();
    }
    catch (const std::exception& e) {
        exception_handler();
    }
}

int cleanup_python()
{
    try {
        cleanup();
    }
    catch (const std::exception& e) {
        exception_handler();
    }
}
```

I can try implement try this approach also, but currently I liked my original approach better :wink:

atgeirr:
> I can try implement try this approach also, but currently I liked my original approach better

Let me think a little bit. In general, I want the API used from Python to be so nice that we will use the same API from C++, avoiding two different ways of doing things. If I cannot find a better way quickly, we'll merge this.

atgeirr:
> If I cannot find a better way quickly, we'll merge this.

I'll admit defeat on this one... As long as the python API is not written in stone we can always improve and refactor if we see better ways. I'll merge.

PR #2620, opened 2020-05-14
Opm restart

hnil:
Change to make it simple to restart without file manipulations. This would be usefull at least for developers. It depend on pullrequest from opm-common. The intendend use is
```sh
flow SPE1CASE2.DATA --output-dir=spe1case2restart  \
    --opm-restart=true --opm-restart-step=60 \
    --opm-restart-base=fullpath/spe1_fullfowardrun/SPE1CASE2
```
- this is a draft due to changes in opm-common

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -80,6 +80,9 @@ NEW_PROP_TAG(IgnoreKeywords);
 NEW_PROP_TAG(EnableExperiments);
 NEW_PROP_TAG(EdgeWeightsMethod);
 NEW_PROP_TAG(OwnerCellsFirst);
+NEW_PROP_TAG(OpmRestart);
+NEW_PROP_TAG(OpmRestartBase);
+NEW_PROP_TAG(OpmRestartStep);
 
 SET_STRING_PROP(EclBaseVanguard, IgnoreKeywords, "");
 SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "");
@@ -89,6 +92,10 @@ SET_BOOL_PROP(EclBaseVanguard, EclStrictParsing, false);
 SET_BOOL_PROP(EclBaseVanguard, SchedRestart, true);
 SET_INT_PROP(EclBaseVanguard, EdgeWeightsMethod, 1);
 SET_BOOL_PROP(EclBaseVanguard, OwnerCellsFirst, true);
+
+SET_BOOL_PROP(EclBaseVanguard,OpmRestart,false);
+SET_STRING_PROP(EclBaseVanguard,OpmRestartBase,"none");
+SET_INT_PROP(EclBaseVanguard,OpmRestartStep,-1);
 
 END_PROPERTIES
 
@@ -138,6 +145,10 @@ public:
                              "Choose edge-weighing strategy: 0=uniform, 1=trans, 2=log(trans).");
         EWOMS_REGISTER_PARAM(TypeTag, bool, OwnerCellsFirst,
                              "Order cells owned by rank before ghost/overlap cells.");
+        EWOMS_REGISTER_PARAM(TypeTag, bool, OpmRestart, "Use opm restart feature");
+        EWOMS_REGISTER_PARAM(TypeTag, std::string, OpmRestartBase,"The base of previous run for OPM restart feature");
+        EWOMS_REGISTER_PARAM(TypeTag, int, OpmRestartStep,"Restart step for OPM restart feature");
+
     }
 
     /*!
--- a/opm/simulators/flow/Main.hpp
+++ b/opm/simulators/flow/Main.hpp
@@ -418,15 +418,23 @@ namespace Opm
                           included here as a switch.
                         */
                         const bool init_from_restart_file = !EWOMS_GET_PARAM(PreTypeTag, bool, SchedRestart);
-                        const auto& init_config = eclipseState_->getInitConfig();
+                        auto& init_config = eclipseState_->getInitConfig();
+                        const bool opm_restart = EWOMS_GET_PARAM(PreTypeTag, bool, OpmRestart);
+			if( opm_restart ) {
+			    int report_step = EWOMS_GET_PARAM(PreTypeTag, int, OpmRestartStep);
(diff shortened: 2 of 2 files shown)


joakim-hove:
> -opm-restart=true --opm-restart-step=60 --opm-restart-base=fullpath/spe1_fullfowardrun/SPE1CASE2

Could I suggest the alternative syntax:
```
bash% ...  --opm-restart=fullpath/spe1_fullforward/SPE1CASE2:60
```

hnil:
I can change the syntax, but wait until I know if the opm-common thing is acceptable.

atgeirr:
I like the easy restarting! But I wonder what is supposed to happen with output in the case you describe: will the new output overwrite the `fullpath/spe1_fullforward/SPE1CASE2.*` files, or be placed in the output directory (defaulted or set with `--output-dir`)?

I also like the idea of having a single parameter  named `--restart` as suggested by @joakim-hove, as it means we do not have to deal with sets of multiple parameters that have nontrivial relationships. The syntax suggested is ok.

The upstream PR looks like it's only some details away from merging?

bska:
> But I wonder what is supposed to happen with output in the case you describe: will the new output overwrite the `fullpath/spe1_fullforward/SPE1CASE2.*` files, or be placed in the output directory (defaulted or set with `--output-dir`)?

I can answer that.  The output always goes to the [output directory](https://github.com/OPM/opm-common/blob/3a8ea856bcac466d122432dc422f9e27f906fa68/src/opm/output/eclipse/EclipseIO.cpp#L232-L233).  That, in turn, is what is returned by the run's [`IOConfig::getOutputDirectory()`](https://github.com/OPM/opm-common/blob/3a8ea856bcac466d122432dc422f9e27f906fa68/src/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.cpp#L242-L244) which ultimately is either the directory containing the run's `.DATA` file or a location that's explicitly set using `output-dir`.

As for what happens if you run
```sh
flow SPE1CASE2.DATA --opm-restart-base=SPE1CASE2 --opm-restart-step=60
```
then the new run will overwrite the original run's data from step 60 onward.  The results from the first 59 steps will remain intact.  That said, this particular feature is mostly targeted at developers who should know better than to do that.

atgeirr:
> That said, this particular feature is mostly targeted at developers who should know better than to do that.

Knowing myself, I think we should disallow overwriting, by ensuring we do not restart from and write to the same file. I guess that should not be very hard?

hnil:
I accept anything to get the feature in. It is allowed with overwriting in all other cases. I also like to have the number separate so one do not need to to string parsing and check if the part you search for is int.., but I agree it looks nice. The feature is tested on several cases now and seems to work fine for simple test cases, however a large field case with group control get some wells closed, unsure if this is related to this restart feature. A separate pullrequest is made to make tuning work with restart.

OPMUSER:
Anything that reduces the number of command line parameters gets my vote, as we currently have 138 command line parameters in the 2020-04 release. Everybody always adds command line parameters but nobody removes redundant ones like: --vtk-write-average-molar-masses, --vtk-write-fugacities etc.

So @joakim-hove approach of just using one instead of three gets my vote:

`bash% ...  --restart=fullpath/spe1_fullforward/SPE1CASE2:60`

As it relatively straight forward to use - but I'm not a developer :-)

bska:
jenkins build this opm-common=1786 please

hnil:
Due to the design of the tagsystem paramter options will add even the one which is irrelevent. They can however be hided from flow. We should have make a list or a pullrequest to hide as many as possible. But my opinion is that usefull options should be added, in particular the ones which aim at speed up development or at resonable improvements, which it will take time to get test for all cases.

alfbr:
> Anything that reduces the number of command line parameters gets my vote

I agree, we need to clean it up. I do not think we should prevent new options from coming in though, so I am very supportive of the switches being provided by @hnil 

I do believe we should be able to group the options though, this is something we probably should put effort in before the next release. Also, simplifying the new options like suggested here seems worth while.

hnil:
Some parameter-tree based setup will make gruping much better than the "flat" structure providing by the TypeTags. If we can separate parameters and the templating using Bernts pullrequest this will make this easier. For now we could just collect not used options for flow and hide them ("we are already hiding quite a few")

alfbr:
> parameter-tree

Not sure what you mean by parameter-tree. I am simply thinking of grouping them similar to what users already are used to from other command line options. Check out "ssh -o" or "find -fstype". By allowing the input to the option to be a string, we can group together many options.

hnil:
What I mean with a parameter tree sit that the all the we have a structured tree which definethings and that anything can have different types. This is an example from he linear solvers which set up the default solver (if well contributions is included in the matrix), the importatn thing is that all paramters for the preconditioner is collected in a separate structure, which again could have separate structures:
{
    "tol": "0.01",
    "maxiter": "200",
    "verbosity": "20",
    "solver": "bicgstab",
    "preconditioner": {
        "type": "ParOverILU0",
	"relaxation": "1"
    }
}

alfbr:
I see. Maybe it can replace selected command line parameters, but it would make that functionality far less accessible for users.

hnil:
No. The only change will be to make it possible to group the opptions i.e.
for the example if the above structure is inside the grop linearsolver then
flow --linearsolver.tol=1e-3 --linearsolver.preconditioner.type=SeqILU0  willl set the linearsolver tolerance and the preconditioner type to SeqILU0. Also one could the help message could have default printout of only one level which would make it much cleaner or possibly with a possibility to do flow --help:linearsolver to get the settings for the grupe linearsolver.

alfbr:
> No. The only change will be to make it possible to group the options i.e.
> for the example if the above structure is inside the group linearsolver then
> flow --linearsolver.tol=1e-3 --linearsolver.preconditioner.type=SeqILU0 willl set the linearsolver tolerance and the preconditioner type to SeqILU0. Also the help message could have default printout of only one level which would make it much cleaner or possibly with a possibility to do flow --help:linearsolver to get the settings for the group linearsolver.

I am afraid this moves out of the realm of what most think of as command line options. Moreover, I am afraid such a move would in practice make the options inaccessible to users.

bska:
jenkins build this please

PR #2622, opened 2020-05-14
made constructor consistent with updateTuning

hnil:
Made the construct of timestepping use the same parts of tuning as the updateTuning. This fixed a problem with using tuning in restart runs.


bska:
jenkins build this please

bska:
Exvellent detective work.  I'll merge into master.  Thanks a lot lot!

PR #2625, opened 2020-05-15
Modify cpr defaults (for better performance)

blattms:
Based on the config posted by @hnil in PR #2608, we use trueimpes as default. Still we only use 1 step of pre- and postsmoothing as this is faster for Norne than what was proposed. Added some more cleanups without any effects, tool

Total time
| trueimpes 2 steps | trueimpes 1 step | quasiimpes 1 step |
|---|---|---|
| 640 | 594 | 650 |


blattms:
jenkins build this please

blattms:
Approved, self merging

PR #2593, opened 2020-05-06
Writematrix

hnil:
Added options for writing which the flexible solver solve on. To use it one have to run with optionfile and change the verbosity level in the file. example:
```sh
opm-simulators/bin/flow_blackoil_dunecpr SPE1CASE1.DATA --output-dir=test \
   --linear-solver-configuration-json-file=options.json --linear-solver-configuration=file
```
with `option.json: verbosity>10` will write out matrices
```json
{
    "tol": "0.01",
    "maxiter": "20",
    "verbosity": "20",
    "solver": "bicgstab",
    "preconditioner": {
        "type": "ParOverILU0",
	"relaxation": "1"
    }
}
```

PR #2627, opened 2020-05-15
Write out the linear system when ultra high verbosity is requested.

blattms:
This added to ISTLSolverEbosFlexible in PRs #2593 and #2594  and should be available form ISTLSolverEbos for consistency reasons, too. Done with this commit.

### DIFF
--- a/opm/simulators/linalg/ISTLSolverEbos.hpp
+++ b/opm/simulators/linalg/ISTLSolverEbos.hpp
@@ -31,6 +31,7 @@
 #include <opm/simulators/linalg/findOverlapRowsAndColumns.hpp>
 #include <opm/simulators/linalg/setupPropertyTree.hpp>
 #include <opm/simulators/linalg/FlexibleSolver.hpp>
+#include <opm/simulators/linalg/WriteSystemMatrixHelper.hpp>
 #include <opm/common/Exceptions.hpp>
 #include <opm/simulators/linalg/ParallelIstlInformation.hpp>
 #include <opm/common/utility/platform_dependent/disable_warnings.h>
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

bska:
This ended up creating some really [unbelievable](https://ci.opm-project.org/job/opm-simulators-PR-builder/968/testReport/) regression failures.  Either master is broken now or we picked up a stale reference somewhere.  I'll try to rerun the build check to see if the results change.

bska:
jenkins build this please

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -485,6 +486,8 @@ class WellModelGhostLastMatrixAdapter : public Dune::AssembledLinearOperator<M,X
         }
 
         bool solve(Vector& x) {
+            const int verbosity = prm_.get<int>("verbosity");
```
bska, line 489: Do we know that the `verbosity` parameter always exists?  If not, that could be a possible reason for the very weird [regression failures](https://ci.opm-project.org/job/opm-simulators-PR-builder/969/testReport/) we're getting.

blattms: It should not be the cause, but it could trigger it:
- verbosity might be an arbitrary value, as prm_ is empty for defautl run.
- Assuming it is higher than ten, the (not 100% tested?) `Opm::Helper::writeSystem()` will be called
- If that overwrites something, then it might explain the problems.

This pure speculation as of now, though. I can be wrong and often have been.

blattms: I am able to reproduce locally and will look into this.

blattms: Its me (of course :angry:). This is the throwing version. Hence it throws an exception, the time stepper chops the step (a different bug),  and here we are...

blattms:
jenkins build this please

bska:
Build check is green.  I'll merge this into master.  Thanks a lot.

PR #2628, opened 2020-05-15
cusparseSolver can now handle MultisegmentWells

Tongdongq:
The MultisegmentWells are actually applied on CPU via SuiteSparse/UMFPACK.

some notes:
- The MultisegmentWells are applied on the CPU, this involves copying vectors x and y to the CPU, applying C, applying D using UMFPACK, applying B, copying vector y to the GPU. This is done twice per linear iteration, and incurs huge transfer costs. Applying the MultisegmentWells fully on GPU can be done in a later PR.
- The cusparseSolver object receives a WellContributions object, this object contains the data for StandardWells, plus a vector with pointers to MultisegmentWellContribution objects. StandardWells are applied simultaneously on GPU, thus their data should be in 1 array. MultisegmentWells are applied sequentially on CPU.
- An OpmLog::warning is printed if BlackoilWellModel cannot determine if a well is either StandardWell or MultisegmentWell, then the well is not send to cusparseSolver. This could also be an error.
- The number of Newton Iterations for Dune seem to have decreased between March and May. However, it seems to have increased for the 2080Ti. The K40 has less iterations for the most recent version.
- The MultisegmentWellContribution constructor inserts its data in vectors. Instead of vectors, it could probably use the raw pointers, but the time to insert is quite small anyway.

Intel Xeon E5-2620 v3 @ 2.4 GHz
NVIDIA RTX 2080Ti, NVIDIA Tesla K40c

```
Based on master of March 23 2020 (3d04df97):
NORNE_1A_MSW:
	Dune:
		runtime: 2104 s
		Assembly time: 613 s
		Linear solve time: 1421 s
		Newton Iterations: 4019
		Linear Iterations: 71511
	cusparse + UMFPACK, 2080Ti:
		runtime: 1294 s
		Assembly time: 527 s
		Linear solve time: 701 s
		Newton Iterations: 3346
		Linear Iterations: 51353
		Accum DeviceToHost copy: 170 s, probably wrong measurement
		Accum HostToDevice copy: 98 s
Bigger testcase:
	Dune:
		runtime: 29814 s or 8h16m
		Assembly time: 12267 s
		Linear solve time: 16436 s
		Newton Iterations: 3553
		Linear Iterations: 25905
	cusparse + UMFPACK, 2080Ti:
		runtime: 24864 s or 6h54m
		Assembly time: 13941 s
		Linear solve time: 9591 s
		Newton Iterations: 3515
		Linear Iterations: 23878
		Accum DeviceToHost copy: 3525 s, probably wrong measurement
		Accum HostToDevice copy: 1918 s

Based on master of May 12 2020 (1f177b33):
NORNE_1A_MSW:
	Dune:
		runtime: 1764 s
		Assembly time: 532 s
		Linear solve time: 1165 s
		Newton Iterations: 3381
		Linear Iterations: 58574
	cusparse + UMFPACK, 2080Ti:
		runtime: 2008 s
		Assembly time: 868 s
		Linear solve time: 1047 s
		Newton Iterations: 5586
		Linear Iterations: 80504
		Accum DeviceToHost copy: 546 s, probably wrong measurement
		Accum HostToDevice copy: 152 s
	cusparse + UMFPACK, K40:
		runtime: 1360 s
		Assembly time: 401 s
		Linear solve time: 903 s
		Newton Iterations: 2663
		Linear Iterations: 45255
		Accum DeviceToHost copy: 590 s, probably wrong measurement
		Accum HostToDevice copy: 161 s
```

### DIFF
--- a/opm/simulators/linalg/bda/MultisegmentWellContribution.cpp
+++ b/opm/simulators/linalg/bda/MultisegmentWellContribution.cpp
@@ -19,16 +19,10 @@
 
 
 #include <cstdlib>
-#include <cstring>
 #include <config.h> // CMake
-#include <fstream>
-
-#include <opm/common/OpmLog/OpmLog.hpp>
-#include <opm/common/ErrorMacros.hpp>
 
 #if HAVE_UMFPACK
 #include <dune/istl/umfpack.hh>
-#include <opm/simulators/linalg/bda/cuda_header.hpp>
 #endif // HAVE_UMFPACK
 
 #include <opm/simulators/linalg/bda/MultisegmentWellContribution.hpp>
@@ -63,33 +57,22 @@ namespace Opm
             z1.resize(Mb * dim_wells);
             z2.resize(Mb * dim_wells);
 
-            // allocate pinned memory on host
-            cudaMallocHost(&h_x, sizeof(double) * N);
-            cudaMallocHost(&h_y, sizeof(double) * N);
-
             umfpack_di_symbolic(M, M, Dcols.data(), Drows.data(), Dvals.data(), &UMFPACK_Symbolic, nullptr, nullptr);
             umfpack_di_numeric(Dcols.data(), Drows.data(), Dvals.data(), UMFPACK_Symbolic, &UMFPACK_Numeric, nullptr, nullptr);
         }
 
         MultisegmentWellContribution::~MultisegmentWellContribution()
         {
-            cudaFreeHost(h_x);
-            cudaFreeHost(h_y);
-
             umfpack_di_free_symbolic(&UMFPACK_Symbolic);
             umfpack_di_free_numeric(&UMFPACK_Numeric);
         }
 
 
         // Apply the MultisegmentWellContribution, similar to MultisegmentWell::apply()
+        // h_x and h_y reside on host
         // y -= (C^T * (D^-1 * (B * x)))
-        void MultisegmentWellContribution::apply(double *d_x, double *d_y)
+        void MultisegmentWellContribution::apply(double *h_x, double *h_y)
         {
-            // copy vectors x and y from GPU to CPU
-            cudaMemcpyAsync(h_x, d_x, sizeof(double) * N, cudaMemcpyDeviceToHost, stream);
-            cudaMemcpyAsync(h_y, d_y, sizeof(double) * N, cudaMemcpyDeviceToHost, stream);
-            cudaStreamSynchronize(stream);
-
             // reset z1 and z2
             std::fill(z1.begin(), z1.end(), 0.0);
             std::fill(z2.begin(), z2.end(), 0.0);
@@ -128,10 +111,6 @@ namespace Opm
                     }
                 }
             }
-
-            // copy vector y from CPU to GPU
-            cudaMemcpyAsync(d_y, h_y, sizeof(double) * N, cudaMemcpyHostToDevice, stream);
-            cudaStreamSynchronize(stream);
         }
 
         void MultisegmentWellContribution::setCudaStream(cudaStream_t stream_)
--- a/opm/simulators/linalg/bda/MultisegmentWellContribution.cpp
+++ b/opm/simulators/linalg/bda/MultisegmentWellContribution.cpp
@@ -1,0 +1,138 @@
+/*
+  Copyright 2020 Equinor ASA
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+
+#include <cstdlib>
+#include <cstring>
+#include <config.h> // CMake
+#include <fstream>
+
+#include <opm/common/OpmLog/OpmLog.hpp>
+#include <opm/common/ErrorMacros.hpp>
+
+#if HAVE_UMFPACK
+#include <dune/istl/umfpack.hh>
+#include <opm/simulators/linalg/bda/cuda_header.hpp>
+#endif // HAVE_UMFPACK
+
+#include <opm/simulators/linalg/bda/MultisegmentWellContribution.hpp>
+
+namespace Opm
+{
+
+        MultisegmentWellContribution::MultisegmentWellContribution(unsigned int dim_, unsigned int dim_wells_,
+            unsigned int Nb_, unsigned int Mb_,
+            unsigned int BnumBlocks_, double *Bvalues, unsigned int *BcolIndices, unsigned int *BrowPointers,
+            unsigned int DnumBlocks_, double *Dvalues, int *DcolPointers, int *DrowIndices,
+            double *Cvalues)
+            : dim(dim_), dim_wells(dim_wells_), N(Nb_*dim), Nb(Nb_), M(Mb_*dim_wells), Mb(Mb_), DnumBlocks(DnumBlocks_), BnumBlocks(BnumBlocks_)
+        {
+            Cvals.insert(Cvals.end(), Cvalues, Cvalues + BnumBlocks*dim*dim_wells);
+            Dvals.insert(Dvals.end(), Dvalues, Dvalues + DnumBlocks*dim_wells*dim_wells);
+            Bvals.insert(Bvals.end(), Bvalues, Bvalues + BnumBlocks*dim*dim_wells);
+            Ccols.insert(Ccols.end(), BcolIndices, BcolIndices + BnumBlocks);
+            Bcols.insert(Bcols.end(), BcolIndices, BcolIndices + BnumBlocks);
+            Crows.insert(Crows.end(), BrowPointers, BrowPointers + M + 1);
+            Brows.insert(Brows.end(), BrowPointers, BrowPointers + M + 1);
+
+            Dcols.insert(Dcols.end(), DcolPointers, DcolPointers + M + 1);
+            Drows.insert(Drows.end(), DrowIndices, DrowIndices + DnumBlocks*dim_wells*dim_wells);
+
+            z1.resize(Mb * dim_wells);
+            z2.resize(Mb * dim_wells);
+
+            // allocate pinned memory on host
+            cudaMallocHost(&h_x, sizeof(double) * N);
+            cudaMallocHost(&h_y, sizeof(double) * N);
+
+            umfpack_di_symbolic(M, M, Dcols.data(), Drows.data(), Dvals.data(), &UMFPACK_Symbolic, nullptr, nullptr);
+            umfpack_di_numeric(Dcols.data(), Drows.data(), Dvals.data(), UMFPACK_Symbolic, &UMFPACK_Numeric, nullptr, nullptr);
+        }
+
+        MultisegmentWellContribution::~MultisegmentWellContribution()
+        {
+            cudaFreeHost(h_x);
+            cudaFreeHost(h_y);
+
+            umfpack_di_free_symbolic(&UMFPACK_Symbolic);
+            umfpack_di_free_numeric(&UMFPACK_Numeric);
+        }
+
+
+        // Apply the MultisegmentWellContribution, similar to MultisegmentWell::apply()
+        // y -= (C^T * (D^-1 * (B * x)))
+        void MultisegmentWellContribution::apply(double *d_x, double *d_y)
+        {
+            // copy vectors x and y from GPU to CPU
+            cudaMemcpyAsync(h_x, d_x, sizeof(double) * N, cudaMemcpyDeviceToHost, stream);
+            cudaMemcpyAsync(h_y, d_y, sizeof(double) * N, cudaMemcpyDeviceToHost, stream);
+            cudaStreamSynchronize(stream);
+
+            // reset z1 and z2
+            std::fill(z1.begin(), z1.end(), 0.0);
+            std::fill(z2.begin(), z2.end(), 0.0);
+
+            // z1 = B * x
+            for (unsigned int row = 0; row < Mb; ++row) {
+                // for every block in the row
+                for (unsigned int blockID = Brows[row]; blockID < Brows[row+1]; ++blockID) {
+                    unsigned int colIdx = Bcols[blockID];
+                    for (unsigned int j = 0; j < dim_wells; ++j) {
+                        double temp = 0.0;
+                        for (unsigned int k = 0; k < dim; ++k) {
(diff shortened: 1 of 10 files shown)


blattms:
Thanks a lot. I will review this today.

two questions upfront:
1. What is the rational for using UMFPack on the CPU to solve the multisegment well equations?
2. Is this the (planned) final version or there more PRs based on this one to be expected?

Tongdongq:
1. The normal Dune-based version uses UMFPack to solve the D part of the MultisegmentWells. We could use another instance of cusparseSolver to do that, but it is meant to solve the large linear system and only be instantiated once. It also uses ILU0, which appears to be overkill for the D matrices. They all solve in 0 iterations (probably 0.5), which makes the ILU decomposition relatively expensive. We probably could make the ILU optional in the cusparseSolver, but I don't know how it will affect the result.
2. I'd like to optimize all those vectortransfers away by solving it entirely on the GPU, but we need to get working on the openclSolver.
Another reason for this PR is to discuss the results so far. In the March master, we have speedup for both cases, although for NORNE it is also based on the number of iterations. The May master has less iterations for CPU, but a lot more for the GPU. At least for the 2080Ti, because the K40 has about half the number of iterations. I'm not sure how to interpret these results.

blattms:
jenkins build this please

opm/simulators/linalg/bda/MultisegmentWellContribution.cpp
```diff
@@ -0,0 +1,138 @@
...
+            Cvals.insert(Cvals.end(), Cvalues, Cvalues + BnumBlocks*dim*dim_wells);
+            Dvals.insert(Dvals.end(), Dvalues, Dvalues + DnumBlocks*dim_wells*dim_wells);
+            Bvals.insert(Bvals.end(), Bvalues, Bvalues + BnumBlocks*dim*dim_wells);
+            Ccols.insert(Ccols.end(), BcolIndices, BcolIndices + BnumBlocks);
+            Bcols.insert(Bcols.end(), BcolIndices, BcolIndices + BnumBlocks);
+            Crows.insert(Crows.end(), BrowPointers, BrowPointers + M + 1);
+            Brows.insert(Brows.end(), BrowPointers, BrowPointers + M + 1);
+
+            Dcols.insert(Dcols.end(), DcolPointers, DcolPointers + M + 1);
+            Drows.insert(Drows.end(), DrowIndices, DrowIndices + DnumBlocks*dim_wells*dim_wells);
```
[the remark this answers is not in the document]
atgeirr: I concur with @blattms that this should be in the initializer list, if possible. I prefer that such a list is structured with no more than one member variable initialized on each line, for better readability though.

blattms:
>It also uses ILU0, which appears to be overkill for the D matrices. They all solve in 0 iterations (probably 0.5), which makes the ILU decomposition relatively expensive. 

Sure there is some overhead (one empty matrix traversion?). Isyou comment an assumption or did you actually do tests and have numbers? If there is serious overhead we should also change the CPU code.

>        Another reason for this PR is to discuss the results so far. In the March master, we have speedup for both cases, although for NORNE it is also based on the number of iterations. The May master has less iterations for CPU, but a lot more for the GPU. At least for the 2080Ti, because the K40 has about half the number of iterations. I'm not sure how to interpret these results.

I cannot comment on the numbers as I am missing information: What GPU/system was used for the March numbers? May numbers are the small problem?

opm/simulators/linalg/bda/MultisegmentWellContribution.cpp
```diff
@@ -0,0 +1,138 @@
...
+            cudaMemcpyAsync(h_x, d_x, sizeof(double) * N, cudaMemcpyDeviceToHost, stream);
+            cudaMemcpyAsync(h_y, d_y, sizeof(double) * N, cudaMemcpyDeviceToHost, stream);
+            cudaStreamSynchronize(stream);
+
+            // reset z1 and z2
+            std::fill(z1.begin(), z1.end(), 0.0);
+            std::fill(z2.begin(), z2.end(), 0.0);
+
+            // z1 = B * x
+            for (unsigned int row = 0; row < Mb; ++row) {
```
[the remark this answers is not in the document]
Tongdongq: Yes it should move to the GPU eventually. It should behave similar to the kernel for the StandardWells in WellContributions.cu, with having the z1 and z2 vectors on GPU as well. The application of C and B are easy to port to a kernel, the difficult one is D.

Tongdongq:
Both versions are done on the same machine, with the Xeon E5-2620 v3. I added 2080Ti for the March versions to clarify. What confuses me is the different amount of iterations between March and May, and the 2080Ti and K40 of May. I have no numbers for the K40 of March right now.

The K40 run of May spends about 12s total on solving D (not C or B) using UMFPack for NORNE. I checked a few iterations using another instance of cusparseSolver to solve D. It looks like the ILU decomposition takes about 200-500 us each time, and the ILU application about 60% of that time. It might be worth not doing an ILU solver for D on GPU once we get to that point. I just used it because it was available.
There is no need to change any CPU code, UMFPack uses some kind of backward error method to apply D, no ILU is done there.

blattms:
The segmentation fault was a hickup on my system (I accidentally used master).

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -639,6 +639,66 @@ namespace Opm
...
+                        Bvals.emplace_back((*colB)[i][j]);
+                    }
+                }
+                sizeRow++;
+            }
+            sumBlocks += sizeRow;
+            Brows.emplace_back(sumBlocks);
+        }
+
+        wellContribs.addMultisegmentWellContribution(numEq, numWellEq, Nb, Mb, BnumBlocks, Bvals.data(), Bcols.data(), Brows.data(), DnumBlocks, Dvals, Dcols, Drows, Cvals.data());
```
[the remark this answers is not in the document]
Tongdongq: That is indeed a bit weird. I think we can use `std::move` and use `std::vector` in `MultisegmentWellContribution`, or use the raw pointers. The `Dune::UMFPack` object returns raw pointers for D. If C and B are to be copied to the GPU as well, making those raw pointers as well might make sense, although keeping track of length must be done carefully.
I see that `Ccols` and `Crows` are actually not used, I'll remove those.

blattms:
Mmh, on my system (AMD Ryzen Threadripper 1950X 16-Core Processor, GeForce GTX 1060 6GB, Compute Capability: 6.1) the GPU approach is quite a bit slower than the CPU for NORNE_1A_MSW. On the one hand this is probably due to more time step chopping, on the other the individual solves also appear to be slower. Screenshot meld report step 1 of NORNE_1A_MSW

In total runtime the GPU nearly takes 3 times the time of the CPU:
```
mblatt@smaug:~/src/dune/opm-gpu/opm-tests/norne$ tail -n 15 20200518_1A_MSW__gpu_true.log 

Number of MPI processes:      1
Threads per MPI process:      2
Total time (seconds):         4139.67
Solver time (seconds):        4139.55
 Assembly time (seconds):     695.644 (Failed: 478.069; 68.7233%)
 Linear solve time (seconds): 3369.73 (Failed: 2061.9; 61.1888%)
 Linear solve setup time (seconds): 0.00409871 (Failed: 0.00298142; 72.7404%)
 Update time (seconds):       38.5821 (Failed: 26.0741; 67.5808%)
 Output write time (seconds): 4.82521
Overall Well Iterations:      122 (Failed: 30; 24.5902%)
Overall Linearizations:       4867 (Failed: 3213; 66.016%)
Overall Newton Iterations:    4530 (Failed: 3213; 70.9272%)
Overall Linear Iterations:    66385 (Failed: 38425; 57.8821%)

mblatt@smaug:~/src/dune/opm-gpu/opm-tests/norne$ tail -n 15 20200518_1A_MSW__gpu_false.log 

Number of MPI processes:      1
Threads per MPI process:      2
Total time (seconds):         1411.98
Solver time (seconds):        1411.86
 Assembly time (seconds):     565.486 (Failed: 359.88; 63.6409%)
 Linear solve time (seconds): 776.473 (Failed: 436.097; 56.1639%)
 Linear solve setup time (seconds): 0.00221778 (Failed: 0.0014616; 65.9038%)
 Update time (seconds):       37.7358 (Failed: 23.9204; 63.3893%)
 Output write time (seconds): 4.76881
Overall Well Iterations:      122 (Failed: 32; 26.2295%)
Overall Linearizations:       3798 (Failed: 2332; 61.4007%)
Overall Newton Iterations:    3520 (Failed: 2332; 66.25%)
Overall Linear Iterations:    60505 (Failed: 32858; 54.3063%)
```
Where do I get these accumulated transfer times from?

Tongdongq:
> Where do I get these accumulated transfer times from?

I'm sorry, I didn't include that in the PR.  I did some dirty `static double t_total = 0.0; t_total += t_stop-t_start;`. I could make another commit, but it feels a bit like polluting the history. I can also put it here as a comment, then you can copy-paste it.

opm/simulators/linalg/bda/MultisegmentWellContribution.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include <dune/istl/umfpack.hh>
+#include <opm/simulators/linalg/bda/cuda_header.hpp>
+#endif // HAVE_UMFPACK
+
+#include <opm/simulators/linalg/bda/MultisegmentWellContribution.hpp>
+
+namespace Opm
+{
+
+        MultisegmentWellContribution::MultisegmentWellContribution(unsigned int dim_, unsigned int dim_wells_,
```
blattms, line 39: Sometimes some comments would not hurt. What is `Nb_`? Is that the size of the whole system seen as a scalar matrix?

Tongdongq: There are comments in the header. Now that they all have their own line, shall I duplicate the comments?

opm/simulators/linalg/bda/MultisegmentWellContribution.cpp
```diff
@@ -0,0 +1,138 @@
...
+            umfpack_di_free_numeric(&UMFPACK_Numeric);
+        }
+
+
+        // Apply the MultisegmentWellContribution, similar to MultisegmentWell::apply()
+        // y -= (C^T * (D^-1 * (B * x)))
+        void MultisegmentWellContribution::apply(double *d_x, double *d_y)
+        {
+            // copy vectors x and y from GPU to CPU
+            cudaMemcpyAsync(h_x, d_x, sizeof(double) * N, cudaMemcpyDeviceToHost, stream);
```
blattms, line 89: Assuming that `d_x` and `d_y` are the same for every well, this seems a bit overkill. Wouldn't it suffice to just do:
1. Copy `d_x` and `d_y` to the host once in the outer scope.
2. Solve for each well by calling its `apply`.
3. Copy back once in the outer scope.

Tongdongq: That's pretty good. `WellContributions` can hold that pinned memory instead. I'll try it.

Tongdongq:
With copying done by `WellContributions`:
```
cusparse, umfpack, K40:
	runtime: 860 s
        Assembly time: 376 s
        Linear solve time: 433 s
	Newton Iterations: 2663
	Linear Iterations: 45255
	Accum DeviceToHost copy: 31.3 s
	Accum HostToDevice copy: 17.2 s
```

blattms:
Which means we went down from 1360s to 860s on your system? :tada: I should get a bonus for this... :wink:

I think you should post all numbers again (before this change and after). Scrolling is rather cumbersome.

blattms:
> I think you should post all numbers again (before this change and after). Scrolling is rather cumbersome.

And numbers from the other card, please.

Tongdongq:
I updated the previous comment with more information.
I also noticed this:
```
Dune, NORNE_1A_MSW:
	runtime: 1666 (515+1087) s
	Newton Iterations: 3381
	Linear Iterations: 58574
	Accum MultisegmentWell::apply(): 150.3 s
	Accum mswellhelpers::invDXDirect create Dune::UMFPack: 162.8 s
	Accum mswellhelpers::invDXDirect linsolver.apply(): 19.7 s
```
The creation of a Dune::UMFPack object involves converting a Dune::BCRSMatrix to a CSC format in raw pointers. These raw pointers are used in the UMFPack solve. The invDXDirect() function is called twice per linear iteration per active well, and a few times at other places. Since the D matrix of the MSW does not change during a linear solve, the Dune::UMFPack object, or the CSC format can probably be reused.

blattms:
Do you have an explanation why the number of linear iterations differ so much between K40c and 2080Ti? That is nearly a factor of two?

blattms:
Well, on my system that tiny change (proposed) by me made the runtime go down from 4140 to 1661 seconds (that is still 200seconds more than without GPU)

blattms:
Using UMFPack (which is for unsymmetric sparse matrices) to solve a diagonal block matrix might not be the best choice. One could store all blocks in a vector and use e.g. Lapack to solve. That might be faster.

I am not a top-notch GPU expert. But when moving to the GPU one might want to try to use three big matrices instead of multiple small ones (3 for each well). Or are we already doing that. On second thought this might also be worth a try on the CPU. But that is for another PR.

Will review this again and merge if nothing stands out.

opm/simulators/linalg/bda/MultisegmentWellContribution.cpp
blattms: Indeed

Tongdongq:
I don't know why the  number of iterations is so different. How many iterations did you GPU run take?
For StandardWells, all their data is indeed combined into 3 bigger matrices (C, D and B). They are then solved in parallel. I am also planning this for the MultisegmentWells, but their data is a bit less uniform.

opm/simulators/linalg/bda/MultisegmentWellContribution.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+            // z1 = B * x
+            for (unsigned int row = 0; row < Mb; ++row) {
+                // for every block in the row
+                for (unsigned int blockID = Brows[row]; blockID < Brows[row+1]; ++blockID) {
+                    unsigned int colIdx = Bcols[blockID];
+                    for (unsigned int j = 0; j < dim_wells; ++j) {
+                        double temp = 0.0;
+                        for (unsigned int k = 0; k < dim; ++k) {
+                            temp += Bvals[blockID * dim * dim_wells + j * dim + k] * h_x[colIdx * dim + k];
```
blattms, line 88: Out of curiosity:
Is there a reason for these kind of constructs
```
double temp = 0;
for () temp+= ..;
array[index] += temp;
```
as opposed to
```

auto& val = array[index];
for ()  val+= ..;
```

Tongdongq: I'm not used to using reference variables like that. It might also have had something to do with updating a local variable vs updating an address in the global memory on the GPU. But this is not an issue on the CPU.

opm/simulators/wells/MultisegmentWell_impl.hpp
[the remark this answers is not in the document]
blattms: Is this resolved or will it be?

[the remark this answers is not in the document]
Tongdongq: We didn't really conclude on an implementation. I now pass vector references to be `std::move`d for C and B. D is passed using raw pointers, and initialized in the initializer list.

blattms:
jenkins build this please

PR #2629, opened 2020-05-16
added end_step, usefull for investigating cases for developers

hnil:
Added option to set the end step of a simulation. Together with the restart pull request this make it possible for developers to run and compare sections of previous run. A lot of development aim at fixing issues happening at some small part of the simulation period.

### DIFF
--- a/opm/simulators/flow/FlowMainEbos.hpp
+++ b/opm/simulators/flow/FlowMainEbos.hpp
@@ -54,12 +54,14 @@ NEW_PROP_TAG(EnableDryRun);
 NEW_PROP_TAG(OutputInterval);
 NEW_PROP_TAG(UseAmg);
 NEW_PROP_TAG(EnableLoggingFalloutWarning);
+NEW_PROP_TAG(EndStep);
 
 // TODO: enumeration parameters. we use strings for now.
 SET_STRING_PROP(EclFlowProblem, EnableDryRun, "auto");
 // Do not merge parallel output files or warn about them
 SET_BOOL_PROP(EclFlowProblem, EnableLoggingFalloutWarning, false);
 SET_INT_PROP(EclFlowProblem, OutputInterval, 1);
+SET_INT_PROP(EclFlowProblem, EndStep, std::numeric_limits<int>::max());
 
 END_PROPERTIES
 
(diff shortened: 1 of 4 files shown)


opm/simulators/timestepping/SimulatorTimer.cpp
```diff
@@ -42,14 +43,15 @@ namespace Opm
     {
         const int num_psteps = param.getDefault("num_psteps", 1);
         const double stepsize_days = param.getDefault("stepsize_days", 1.0);
+        const double end_step_ = std::numeric_limits<size_t>::max();
```
bska, line 46: This appears to be an unused local variable.  Is this statement supposed to assign a datamember instead?  The type `double` is also a little questionable given that the right-hand side is `size_t`.

opm/simulators/timestepping/SimulatorTimer.hpp
```diff
@@ -121,6 +121,7 @@ namespace Opm
         std::vector<double> timesteps_;
         int current_step_;
         double current_time_;
+        double end_step_;
```
bska, line 124: Do you really want to have this as `double`?  Is it not supposed to be an integer, maybe even an `int`?

tests/test_timer.cpp
```diff
@@ -47,7 +47,7 @@ BOOST_AUTO_TEST_CASE(CreateTimer)
     boost::gregorian::date defaultStartDate( 2012, 1, 1);
     BOOST_CHECK_EQUAL(  boost::posix_time::ptime(defaultStartDate), simtimer.currentDateTime() );
 
-    simtimer.init(timeMap);
+    simtimer.init(timeMap,0,std::numeric_limits<int>::max());
```
bska, line 50: The function takes a `size_t`, but you're passing an `int` here.  That's a needless type conversion.

opm/simulators/flow/FlowMainEbos.hpp
```diff
@@ -480,10 +484,10 @@ namespace Opm
             const auto& timeMap = schedule.getTimeMap();
             auto& ioConfig = eclState().getIOConfig();
             SimulatorTimer simtimer;
-
+            size_t end_step = EWOMS_GET_PARAM(TypeTag, int, EndStep);
```
bska, line 487: The declared parameter is an `int`, but you're assigning it to a `size_t`.  What is the right parameter type here?

PR #2631, opened 2020-05-19
Refactor runSimulator().

hakonhagland:
A resubmission of commit b25f489 in PR #2403 and PR #2441 to work with the current master.

Continues the work in PR #2619 to refactor `FlowMainEbos.hpp` to work with the Python bindings.

We need to refactor `runSimulator()` to avoid code duplication when `executeStepInit()` is implemented (to be called from the Python interface and implemented in a upcoming PR). Here, `runSimulator()` is refactored into a `runSimulatorInitOrRun()` that takes a callback function. When `runSimulatorInit()` is implemented (upcoming PR) it will pass a different callback that only initializes the simulator. Currently, `runSimulator()` passes the callback `runSimulatorRunCallback_()` which runs the whole simulation. The code outputting the simulation summary in `runSimulator()` is also refactored into a `runSimulatorAfterSim_()` method.

### DIFF
--- a/opm/simulators/flow/FlowMainEbos.hpp
+++ b/opm/simulators/flow/FlowMainEbos.hpp
@@ -488,14 +488,58 @@ namespace Opm
         // Run the simulator.
         int runSimulator(bool output_cout)
         {
+            return runSimulatorInitOrRun_(output_cout, &FlowMainEbos::runSimulatorRunCallback_);
+        }
+
+    private:
+        // Callback that will be called from runSimulatorInitOrRun_().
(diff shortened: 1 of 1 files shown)


joakim-hove:
OK; I must admit I do not fully understand this - and I feel tempted to come with the same comment as @atgeirr did in #2619 - this seems overly complicated; but I also see no other way right now. If jenkins agrees I will merge this, but I can help feeling the current Python <-> C++ interface is more complicated then strictly necessary, so I guess we should revisit this.

joakim-hove:
jenkins build this please

PR #2633, opened 2020-05-20
Mark SuiteSparse as required.

blattms:
UMFPack is used unconditionally in the multisegment well code.
To fully work this also needs OPM/opm-common#1796


blattms:
jenkins build this please

blattms:
Self merging without expressed guru cons

PR #2634, opened 2020-05-21
Default to initialize from restart file in case of restarted run

joakim-hove:
See discussion here: https://github.com/OPM/opm-common/issues/1396

Since some time ago the basic initialize restarted run from the restart file has been merged, and it works for all the restart tests in our test-suite, and also two field cases from Equinor. *But* - it certainly will not work for a lot of models, so when this PR is merged people's restart cases might very well break. They can get the current working behavior back with `--sched-restart=True`. There will be many fallouts which must be fixed, I can certainly do my part in fixing these fallouts - but I it will be a large task and I can not do it alone. 

During the discussion as part of https://github.com/OPM/opm-common/issues/1396 it was agreed that initializing from the restart file was the correct way to go, if we are to eventually get there I see no alternative to merging this PR - but it will probably be painful. There might be reasons to postpone this, or there might be other approaches to get there, but if we do not make an active decision we will not get to the *initialize restarted runs from restart file* situation. 

image

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -86,7 +86,7 @@ SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "");
 SET_INT_PROP(EclBaseVanguard, EclOutputInterval, -1); // use the deck-provided value
 SET_BOOL_PROP(EclBaseVanguard, EnableOpmRstFile, false);
 SET_BOOL_PROP(EclBaseVanguard, EclStrictParsing, false);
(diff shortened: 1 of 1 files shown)


joakim-hove:
ping

alfbr:
From our side, this change is supported. Thanks for the huge effort involved!

alfbr:
This has been open for five days without comments. Unless somebody chimes in, I will merge this at the end of today.

totto82:
I support merging it. If something break we can use the command-line switch as a temporary work-around while fixing the issue.

atgeirr:
I agree, so I'll press green...

PR #2637, opened 2020-05-26
Hide vtk

hnil:
This hide all the vtk output options from flow.  Currently it was not possible to do this dependent on if the enable-vtk option was used.


atgeirr:
I support this, as it will make the "flow --help" experience a lot better. Maybe we need to provide a mechanism to expose all available parameters if requested (unless it already exists), but I think this can be merged. We should wait until after the big de-macro-ing (new word...) of the property system has been merged though.

atgeirr:
jenkins build this please

alfbr:
Hiding less used options is a very good idea. This is exactly what users expect from command line options. A help text only gives a brief description, while the man-page tells the whole story.

hnil:
One could have change the help text for the enable-vtk-output to point to the documentation.

totto82:
What about adding an option that also shows the hidden parameters? Something like --help-all

atgeirr:
After merging the PR changing the property system I think I'll rerun Jenkins to be sure before merging this.

atgeirr:
jenkins build this please

PR #2640, opened 2020-05-28
Rename SpiralICD -> ICD

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/1798

### DIFF
--- a/opm/simulators/utils/ParallelSerialization.cpp
+++ b/opm/simulators/utils/ParallelSerialization.cpp
@@ -22,7 +22,7 @@
 #include <opm/parser/eclipse/EclipseState/Schedule/DynamicState.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/Action/ASTNode.hpp>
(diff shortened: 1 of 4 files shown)


GitPaean:
Hi, how about at least SICD, because there are other ICDs.

GitPaean:
At `ICD.hpp` might have problems with `icd.hpp` in some operating systems, although we do not support them yet.

joakim-hove:
> Hi, how about at least SICD, because there are other ICDs.

Well - the point was that the current `SpiralICD` class seems to be a suitable "base class" for ICD's; see:  OPM/opm-common#1798 - I might be wrong though.

joakim-hove:
> At ICD.hpp might have problems with icd.hpp in some operating systems, although we do not support them yet.

I agree - I will fix that.

atgeirr:
If the "ICD" class is to be (sort of or actually) a base class for ICDs, we could consider giving it the full name "InflowControlDevice", using the ICD abbreviation in the names of subclasses.

joakim-hove:
> we could consider giving it the full name "InflowControlDevice", using the ICD abbreviation in the names of subclasses.

Ohhh man - that is long; I'll start with ICD; and then we will see.

joakim-hove:
Closing to tickle jenkins

PR #2642, opened 2020-05-29
Use std::variant access to Segment::icd

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/1801

~~This code is not yet ready; while refactoring I discovered a method in the MSW implementation which actually mutated the `SpiralICD` instance held by the Schedule object - that was technically allowed in the previous api which had `const std::shared_ptr< mutable >` - but was it intentional - seems suspicious to me!̃~~

Hmmm: pack and unpack of std::variant<> - seems tricky?

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -3627,8 +3627,6 @@ namespace Opm
 
 
 
-
-
     template<typename TypeTag>
     double
     MultisegmentWell<TypeTag>::
@@ -3658,7 +3656,7 @@ namespace Opm
     pressureDropSpiralICD(const int seg) const
     {
         // TODO: We have to consider the upwinding here
-        const SICD& sicd = *segmentSet()[seg].spiralICD();
+        const SICD& sicd = segmentSet()[seg].spiralICD();
 
         const std::vector<EvalWell>& phase_fractions = segment_phase_fractions_[seg];
         const std::vector<EvalWell>& phase_viscosities = segment_phase_viscosities_[seg];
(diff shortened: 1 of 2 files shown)


opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -3638,8 +3638,10 @@ namespace Opm
                     total_connection_length += connection_length;
                 }
 
-                SICD& sicd = *segment.spiralICD();
-                sicd.updateScalingFactor(segment_length, total_connection_length);
```
GitPaean, line 3642: We need to do this, probably because (at least, when the code was written), it was not easy to get the information `segment_length` and `total_connection_length` .  If we can do it during parsing, then we do not need to do it now. 

Not sure when PINCH happens, whether it will change the `connection_length`? which is the tricky part.

GitPaean: One solution can be we do not store the scaling factor as a member. We just call a function  `scalingFactor()`, but we need to calculate the `total_connection_length` each time.

joakim-hove: @GitPaean : I am looking into refactoring this. One question: the current code has the lines:
```C++
const int parental_segment_number = segmentSet()[seg].outletSegment();
const double segment_length = segmentSet().segmentLength(parental_segment_number);
```
Why does the `segment_length` variable use the *outlet_segment*?

GitPaean: yes. That is the definition. The scaling factor can have a few different ways to calculate, one is based on its parental segment, one is based on its total_completion_length. Please refer to the manual for WSEGSICD for more details. I am not familiar with the physics behind.

joakim-hove: Ok thank you - I will stare at the manual, to me it looked a bit suspicious.

joakim-hove:
@akva2: This is *not* particularly general - but it Works for me&trade; - so hopefully this can be merged as a start?

joakim-hove:
jenkins build this opm-common=1801 please

akva2:
yes, if's fine as a start.

PR #2442, opened 2020-03-09
refactor run() in SimulatorFullyImplicitBlackoilEbos.hpp

hakonhagland:
The 5th PR in the splitting of PR #2403 into smaller PRs.

This PR depends on PR #2441 which should be merged first.

## Description

Similar to the previous PR #2441, `run()` is here refactored in preparation for the implementation of a Python `step()` function in a later PR. Currently `run()` is called from `runSimulatorInitOrRun()` in
`FlowMainEbos.hpp` using the `runSimulatorRunCallback_()`. Later, there will be another callback `runSimulatorInitCallback_()` that will need to call different parts of the code in `run()`. The `run()` function is thus refactored into `run()`, `runInit()`, `runStep()`, and `runLastStep()`. Also, some of the local variables in `run()` have to be made persistent between calls to `runStep()`, this applies to
variables `report`, `stepReport`, `solverTimer`, `totalTimer`, and `adaptiveTimeStepping`, which are made private class variables.

PR #2645, opened 2020-05-29
Refactor run() in `SimulatorFullyImplicitBlackoilEbos.hpp`

hakonhagland:
A resubmission of commit bb20804 in PR #2403 and PR #2442 to work with the current master.

Continues the work in #2619 and #2631 to refactor main simulation loop in `flow` to work with the Python bindings.

The `run()` method in `SimulatorFullyImplicitBlackoilEbos.hpp` is refactored in preparation for the implementation of a Python `step()` (and `step_init()`) function (to be implemented in a later commit). Currently `run()` is called from `runSimulatorInitOrRun()` in `FlowMainEbos.hpp` using the `runSimulatorRunCallback_()`. Later, there will be other callbacks like `runSimulatorStepInitCallback_()`, and `runSimulatorStepCallback_()`, that will need to call different parts of the code in `run()`. The `run()` function is thus refactored into `run()`, `runInit()`, `runStep()`, and `runLastStep()`. Also, some of the local variables in `run()` have to be made persistent between calls to `runStep()`, this applies to variables `report`, `solverTimer`, `totalTimer`, and `adaptiveTimeStepping`, which are made private class variables.

### DIFF
--- a/opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.hpp
+++ b/opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.hpp
@@ -129,15 +129,16 @@ public:
     /// \return                    simulation report, with timing data
     SimulatorReport run(SimulatorTimer& timer)
     {
-        runInit(timer);
+        init(timer);
         // Main simulation loop.
         while (!timer.done()) {
-            runStep(timer);
+            bool continue_looping = runStep(timer);
+            if (!continue_looping) break;
         }
-        return runLastStep();
+        return finalize();
     }
 
-    void runInit(SimulatorTimer &timer)
+    void init(SimulatorTimer &timer)
     {
         ebosSimulator_.setEpisodeIndex(-1);
 
@@ -288,7 +289,7 @@ public:
         return true;
     }
 
-    SimulatorReport runLastStep()
+    SimulatorReport finalize()
     {
         // make sure all output is written to disk before run is finished
         {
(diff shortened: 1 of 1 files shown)


hakonhagland:
jenkins build this please

opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.hpp
```diff
@@ -128,170 +128,183 @@ class SimulatorFullyImplicitBlackoilEbos
     /// \param[in,out] state       state of reservoir: pressure, fluxes
     /// \return                    simulation report, with timing data
     SimulatorReport run(SimulatorTimer& timer)
+    {
+        runInit(timer);
```
atgeirr, line 132: I suggest just using the function name `init()` instead of `runInit()`.

opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.hpp
```diff
@@ -128,170 +128,183 @@ class SimulatorFullyImplicitBlackoilEbos
...
+        // Main simulation loop.
+        while (!timer.done()) {
+            runStep(timer);
```
atgeirr, line 135: I think you should check the return value from `runStep()` and return if necessary.

opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.hpp
```diff
@@ -128,170 +128,183 @@ class SimulatorFullyImplicitBlackoilEbos
...
+        }
+        return runLastStep();
```
atgeirr, line 137: The name `runLastStep()` is misleading because it does not run a step. Perhaps `finalize()` is better?

hakonhagland:
jenkins build this please

hakonhagland:
@atgeirr Thanks for the review! Your comment regarding future work on sub step control in `AdaptiveTimeSteppingEbos.hpp` sounds interesting!

PR #2650, opened 2020-06-03
Prevent deadlock after parsing if there were error in parallel

blattms:
Previously only the master process was aware of the error and flow did
deadlock in parallel runs if there were parser errors
encountered. With this commit all processes are made aware of the
problem and flow aborts with an error code.


blattms:
jenkins build this please

blattms:
Shoot. Now `flow help` issues and abort.

blattms:
jenkins build this with opm-models=615

blattms:
jenkins build this with opm-models=615  please

blattms:
Shameless self merge to fix the bug.

PR #2658, opened 2020-06-07
Add restart test of UDQ_WCONPROD

joakim-hove:
Add restart test for case with UDQ / UDA data. Testadata here: https://github.com/OPM/opm-tests/pull/327

~~The restarted case must be run with the non default option `--sched-restart=True` - it was not clear to me how to add that argument driver; last time I tried editing that stuff I spent considerable time - and ended up getting it all wrong. Seems maybe like the `run-restart-regressionTest.sh` script treats treats the arguments a bit haphazardly?~~

My fault: I started with a git ref from last year.

https://github.com/OPM/opm-common/pull/1861

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -926,6 +926,13 @@ add_test_compare_restarted_simulation(CASENAME spe9
                                       REL_TOL ${rel_tol_restart}
                                       TEST_ARGS --sched-restart=false)
 
+add_test_compare_restarted_simulation(CASENAME udq_actionx
+                                      FILENAME UDQ_M1
+                                      SIMULATOR flow
+                                      ABS_TOL ${abs_tol_restart}
+                                      REL_TOL ${rel_tol_restart}
+                                      TEST_ARGS --sched-restart=false)
+
 # The dynamic MSW data is not written to /read from the restart file
 # We therefore accept significant deviation in the results.
 # Note also that we use --sched-restart=true since some necessary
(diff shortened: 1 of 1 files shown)


compareECLFiles.cmake
```diff
@@ -317,6 +317,12 @@ add_test_compare_restarted_simulation(CASENAME spe9
                                       ABS_TOL ${abs_tol_restart}
                                       REL_TOL ${rel_tol_restart})
 
+add_test_compare_restarted_simulation(CASENAME udq_actionx
+                                      FILENAME UDQ_WCONDPROD
+                                      SIMULATOR flow
+                                      ABS_TOL ${abs_tol_restart}
+                                      REL_TOL ${rel_tol_restart})
+
```
alfbr, line 325: > The restarted case must be run with the non default option `--sched-restart=True` - it was not clear to me how to add that argument driver

Try rewriting the last line into the following two lines:
`REL_TOL ${rel_tol_restart}
TEST_ARGS --sched-restart=true)`

joakim-hove: Thank you - that worked; for some reason I had started my work from an ancient git ref.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

jalvestad:
I checked this case today and saw that the restart case produced very different results compared to the base simulation case. Then I ran the same cases with the option --solver-max-time-step-in-days=2 to constrain the time step length. For this time stepping the restart case and the base simulation case seems to be identical! This indicates that the simulations are sensitive to time stepping and perhaps smaller time steps make the UDQ variables vary more smoothly - helping to get the restart values more correct? Still - ideally, the restart case should also be identical with a more flexible time stepping.

joakim-hove:
> I checked this case today [...]

OK - thank you for an interesting analysis; should probably first focus on getting this: #2729 correct and merged?

PR #2660, opened 2020-06-08
Changes in running of restart regression tests

joakim-hove:
- The parallel test gets a designated driver
 - The serial test is run with default adpative timestepping

The main point about these changes is to enable the base run to go with `--enable-adaptive-time-stepping=true` - which is required to get the the base run of: #2658 through


joakim-hove:
jenkins build this please

akva2:
i don't see the change to adaptive time stepping anywhere, and you have to mark the script executable. other than that i support this, makes it clearer.

joakim-hove:
>  i don't see the change to adaptive time stepping anywhere

ehhh - no that was lost in some git foo; now it is there ...

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

PR #2661, opened 2020-06-08
Fix upwinding for friction, acceleration, valve and SIGD for MSW

totto82:
Tested on some relevant test cases and on the big model. 

It mixes pressure derivatives for the density calculations. But since the pressure does not vary to much it should not matter much, but it may introduce some inaccuracy in the matrix.

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -730,6 +730,7 @@ opm_set_test_driver(${PROJECT_SOURCE_DIR}/tests/run-restart-regressionTest.sh ""
 # Cruder tolerances for the restarted tests
 set(abs_tol_restart 2e-1)
 set(rel_tol_restart 4e-5)
+
 add_test_compare_restarted_simulation(CASENAME spe1
                                       FILENAME SPE1CASE2_ACTNUM
                                       SIMULATOR flow
@@ -742,11 +743,19 @@ add_test_compare_restarted_simulation(CASENAME spe9
                                       ABS_TOL ${abs_tol_restart}
                                       REL_TOL ${rel_tol_restart}
                                       TEST_ARGS --sched-restart=false)
+
+# The dynamic MSW data is not written to /read from the restart file
+# We therefore accept significant deviation in the results.
+# Note also that we use --sched-restart=true since some necessary
+# MSW info is still lacking in the restart file.
+set(abs_tol_restart_msw 2e2)
+set(rel_tol_restart_msw 1e-3)
+
 add_test_compare_restarted_simulation(CASENAME msw_3d_hfa
                                       FILENAME 3D_MSW
                                       SIMULATOR flow
-                                      ABS_TOL ${abs_tol_restart}
-                                      REL_TOL ${rel_tol_restart}
+                                      ABS_TOL ${abs_tol_restart_msw}
+                                      REL_TOL ${rel_tol_restart_msw}
                                       TEST_ARGS --enable-adaptive-time-stepping=false --sched-restart=true)
 
 # PORV test
--- a/opm/simulators/wells/MSWellHelpers.hpp
+++ b/opm/simulators/wells/MSWellHelpers.hpp
@@ -206,6 +206,8 @@ namespace mswellhelpers
     template <typename ValueType>
     ValueType velocityHead(const double area, const ValueType& mass_rate, const ValueType& density)
     {
+        // \Note: a factor of 2 is added to the formulation in order to match results from the
+        // reference simulator. This is inline with what is done for the friction loss.
         return (mass_rate * mass_rate / (area * area * density));
     }
 
--- a/opm/simulators/wells/MSWellHelpers.hpp
+++ b/opm/simulators/wells/MSWellHelpers.hpp
@@ -206,7 +206,7 @@ namespace mswellhelpers
     template <typename ValueType>
     ValueType velocityHead(const double area, const ValueType& mass_rate, const ValueType& density)
     {
-        return (0.5 * mass_rate * mass_rate / (area * area * density));
+        return (mass_rate * mass_rate / (area * area * density));
     }
 
 
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1651,13 +1651,9 @@ namespace Opm
             segment_densities_[seg] = density / volrat;
 
             // calculate the mass rates
-            // TODO: for now, we are not considering the upwinding for this amount
-            // since how to address the fact that the derivatives is not trivial for now
-            // and segment_mass_rates_ goes a long way with the frictional pressure loss
-            // and accelerational pressure loss, which needs some work to handle
             segment_mass_rates_[seg] = 0.;
             for (int comp_idx = 0; comp_idx < num_components_; ++comp_idx) {
-                const EvalWell rate = getSegmentRate(seg, comp_idx);
+                const EvalWell rate = getSegmentRateUpwinding(seg, comp_idx);
                 segment_mass_rates_[seg] += rate * surf_dens[comp_idx];
             }
 
@@ -2066,9 +2062,11 @@ namespace Opm
         }
 
         resWell_[seg][SPres] = pressure_equation.value();
-        for (int pv_idx = 0; pv_idx < numWellEq; ++pv_idx) {
-            duneD_[seg][seg][SPres][pv_idx] = pressure_equation.derivative(pv_idx + numEq);
-        }
+        const int seg_upwind = upwinding_segments_[seg];
+        duneD_[seg][seg][SPres][SPres] += pressure_equation.derivative(SPres + numEq);
+        duneD_[seg][seg][SPres][GTotal] += pressure_equation.derivative(GTotal + numEq);
+        duneD_[seg][seg_upwind][SPres][WFrac] += pressure_equation.derivative(WFrac + numEq);
+        duneD_[seg][seg_upwind][SPres][GFrac] += pressure_equation.derivative(GFrac + numEq);
 
         // contribution from the outlet segment
         const int outlet_segment_index = segmentNumberToIndex(segmentSet()[seg].outletSegment());
@@ -2106,8 +2104,16 @@ namespace Opm
     getFrictionPressureLoss(const int seg) const
     {
         const EvalWell mass_rate = segment_mass_rates_[seg];
-        const EvalWell density = segment_densities_[seg];
-        const EvalWell visc = segment_viscosities_[seg];
+        const int seg_upwind = upwinding_segments_[seg];
+        EvalWell density = segment_densities_[seg_upwind];
+        EvalWell visc = segment_viscosities_[seg_upwind];
+        // WARNING
+        // We disregard the derivatives from the upwind density to make sure derivatives
+        // wrt. to different segments dont get mixed.
+        if (seg != seg_upwind) {
+            density.clearDerivatives();
+            visc.clearDerivatives();
+        }
         const int outlet_segment_index = segmentNumberToIndex(segmentSet()[seg].outletSegment());
         const double length = segmentSet()[seg].totalLength() - segmentSet()[outlet_segment_index].totalLength();
         assert(length > 0.);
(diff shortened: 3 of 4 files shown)


totto82:
jenkins build this please

totto82:
The test failure on model2/3D_ does not look good. This PR may need some more care before it can get merged.

totto82:
jenkins build this please

opm/simulators/wells/MSWellHelpers.hpp
```diff
@@ -206,7 +206,8 @@ namespace mswellhelpers
     template <typename ValueType>
     ValueType velocityHead(const double area, const ValueType& mass_rate, const ValueType& density)
     {
-        return (0.5 * mass_rate * mass_rate / (area * area * density));
+        const double sign = mass_rate < 0. ? 1.0 : - 1.0;
+        return (sign*0.5 * mass_rate * mass_rate / (area * area * density));
```
atgeirr, line 210: Are you 100% sure this is correct? I think the head should not depend on the sign here.

If correct, I would prefer to keep this positive, and change the sign in the calling code.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1654,13 +1654,9 @@ namespace Opm
...
             // calculate the mass rates
-            // TODO: for now, we are not considering the upwinding for this amount
-            // since how to address the fact that the derivatives is not trivial for now
-            // and segment_mass_rates_ goes a long way with the frictional pressure loss
-            // and accelerational pressure loss, which needs some work to handle
             segment_mass_rates_[seg] = 0.;
             for (int comp_idx = 0; comp_idx < num_components_; ++comp_idx) {
-                const EvalWell rate = getSegmentRate(seg, comp_idx);
+                const EvalWell rate = getSegmentRateUpwinding(seg, comp_idx);
                 segment_mass_rates_[seg] += rate * surf_dens[comp_idx];
```
atgeirr, line 1660: I think we should add a note to the segment_mass_rates_ data member that its elements have derivatives from mixed sources. Perhaps better, make an explanation of the approach in one place, and refer to it from all other places where it is necessary to have that understanding (such as where we modify duneD_ in the modified code below).

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -3692,49 +3694,55 @@ namespace Opm
...
     {
-        // TODO: We have to consider the upwinding here
         const SICD& sicd = *segmentSet()[seg].spiralICD();
 
-        const std::vector<EvalWell>& phase_fractions = segment_phase_fractions_[seg];
-        const std::vector<EvalWell>& phase_viscosities = segment_phase_viscosities_[seg];
+        const int seg_upwind = upwinding_segments_[seg];
+        // WARNING
+        // We disregard the derivatives from the upwind fractions and viscosities to make sure derivatives
+        // wrt. to different segments dont get mixed.
```
atgeirr, line 3702: Could it be an idea to only discard derivatives if the upwind segment is not the current one?

GitPaean: Then we lose some derivatives. Is that what you are thinking about?

Because the upwinding segment is also in the system.

opm/simulators/wells/MSWellHelpers.hpp
totto82: For an injector the sign of P_seg - P_outlet vil be different and hence the "losses" needs to change sign. From a formulation point of view it would purchase be better to change the sign for the segment pressure and the outlet segment pressure.
In the current kode sign is used for the "losses" instead. I went for the least intrusive change.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: I agree we should go into that direction, but think that should be done in another PR.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: There is already an explanation in getSegmentRateUpwinding. Do you want more?

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2136,33 +2135,33 @@ namespace Opm
...
+        duneD_[seg][seg][SPres][GTotal] -= out_velocity_head.derivative(GTotal + numEq);
+        duneD_[seg][seg_upwind][SPres][WFrac] -= out_velocity_head.derivative(WFrac + numEq);
+        duneD_[seg][seg_upwind][SPres][GFrac] -= out_velocity_head.derivative(GFrac + numEq);
 
         // handling the velocity head of intlet segments
         for (const int inlet : segment_inlets_[seg]) {
-            const EvalWell inlet_density = segment_densities_[inlet];
+            const int seg_upwind_inlet = upwinding_segments_[inlet];
+            const double inlet_area = segmentSet()[inlet].crossArea();
+            const EvalWell inlet_density = segment_densities_[seg_upwind_inlet].value();
```
GitPaean, line 2156: What is the reason we treat the `segment_densties_` and `segment_mass_rates_` differently? They are both associated with the flow compositions, which makes them both related to upwinding.

opm/simulators/wells/MSWellHelpers.hpp
GitPaean: The outlet is defined by the distance to the well head. I mean, the definition of the outlet is not associated with flow direction, at least in the code. Then `seg` and `outlet` will refer to the same two segments. Am I missing something here?

opm/simulators/wells/MultisegmentWell_impl.hpp
atgeirr: No, that is sufficient, thanks for pointing it out. I do think we should refer to it or mark the variables somehow (with comments, not naming) so that we can easily tell which variables are mixed.

opm/simulators/wells/MultisegmentWell_impl.hpp
atgeirr: Ok with me.

opm/simulators/wells/MSWellHelpers.hpp
atgeirr: I think that the sign of (P_seg - P_outlet) will not change depending on flow direction. Typically it will be positive for both injectors and producers for a vertical well, for example.

totto82:
Summary of the test failures. 
Most of them are minor and expected = ok. 
For 9_1B_DEPL_MAX_RATE_MIN_THP_MSW Jenkins throws due to d<0. I am not able to reproduce that locally. 
9_3C_GINJ_GAS_GCONSUMP_MSW (37 != 28) shows a significant improvement in number of problems. 

Note that none of these test cases exposes the A in HFA in WELSEGS. I have also tested several of these models with HFA. Then I get a significant improvement both in terms of results and convergence. 

```
mpi.compareECLFiles_flow+3D_MSW ok
mpi.compareECLFiles_flow+MSW_MODEL_1 ok
mpi.compareECLFiles_flow+FAULTS_MODEL_1 ok
mpi.compareECLFiles_flow+0A1_GRCTRL_LRAT_ORAT_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+0A2_GRCTRL_LRAT_ORAT_GGR_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+0A3_GRCTRL_LRAT_LRAT_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+0A4_GRCTRL_LRAT_LRAT_GGR_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+9_1A_DEPL_MAX_RATE_MIN_BHP_MSW (16 != 17)
mpi.compareECLFiles_flow+9_2A_DEPL_GCONPROD_1L_MSW ok
mpi.compareECLFiles_flow+9_1B_DEPL_MAX_RATE_MIN_THP_MSW (28 != 29) Error: [/var/lib/jenkins/workspace/opm-simulators-PR-builder@2/opm/simulators/wells/MultisegmentWell_impl.hpp:2819] Problematic d value -0 obtained for well PROD1 during convertion to surface volume with rs 580.636, rv 0.00172225 and pressure 7.81824e+07 obtaining d -0

mpi.compareECLFiles_flow+9_2B_DEPL_GCONPROD_2L_MSW ok
mpi.compareECLFiles_flow+9_3A_GINJ_REIN-G_MSW ok
mpi.compareECLFiles_flow+9_3B_GINJ_GAS_EXPORT_MSW ok
mpi.compareECLFiles_flow+9_3C_GINJ_GAS_GCONSUMP_MSW (37 != 28)
mpi.compareECLFiles_flow+9_3E_GAS_MIN_EXPORT_MSW ok
mpi.compareECLFiles_flow+9_3D_GINJ_GAS_MAX_EXPORT_MSW (29 != 28)
mpi.compareECLFiles_flow+9_4A_WINJ_MAXWRATES_MAXBHP_GCONPROD_1L_MSW (28 != 29)
mpi.compareECLFiles_flow+9_4B_WINJ_VREP-W_MSW (30 != 28)
mpi.compareECLFiles_flow+9_4C_WINJ_GINJ_VREP-W_REIN-G_MSW (28 != 29)
mpi.compareECLFiles_flow+TEST_WSEGSICD ok
mpi.compareECLFiles_flow+9_4D_WINJ_GINJ_GAS_EXPORT_MSW (28 != 29)
mpi.compareECLFiles_flow+MOD4_GRP ok
```

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: segment_mass_rates_  contains the derivative wrt. the upwind fractions.

GitPaean: I understand it now. It is really related to how to handle the upwinding for different parts.

totto82:
jenkins build this please

totto82:
New summary of test failures: 
The restart failure needs to be looked further into. The rest are ok. 
```
mpi.compareECLFiles_flow+3D_MSW ok
mpi.compareECLFiles_flow+MSW_MODEL_1 ok
mpi.compareECLFiles_flow+FAULTS_MODEL_1 ok-
mpi.compareECLFiles_flow+0A1_GRCTRL_LRAT_ORAT_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+0A2_GRCTRL_LRAT_ORAT_GGR_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+0A3_GRCTRL_LRAT_LRAT_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+0A4_GRCTRL_LRAT_LRAT_GGR_BASE_MODEL2_MSW ok
mpi.compareECLFiles_flow+9_1A_DEPL_MAX_RATE_MIN_BHP_MSW (16 != 17)
mpi.compareECLFiles_flow+9_1B_DEPL_MAX_RATE_MIN_THP_MSW (28 != 29)
mpi.compareECLFiles_flow+9_2A_DEPL_GCONPROD_1L_MSW ok
mpi.compareECLFiles_flow+9_2B_DEPL_GCONPROD_2L_MSW ok
mpi.compareECLFiles_flow+9_3A_GINJ_REIN-G_MSW ok
mpi.compareECLFiles_flow+9_3B_GINJ_GAS_EXPORT_MSW ok
mpi.compareECLFiles_flow+9_3C_GINJ_GAS_GCONSUMP_MSW (37 != 28)
mpi.compareECLFiles_flow+9_3E_GAS_MIN_EXPORT_MSW ok
mpi.compareECLFiles_flow+9_3D_GINJ_GAS_MAX_EXPORT_MSW (29 != 28)
mpi.compareECLFiles_flow+9_4A_WINJ_MAXWRATES_MAXBHP_GCONPROD_1L_MSW ok
mpi.compareECLFiles_flow+9_4B_WINJ_VREP-W_MSW (30 != 28)
mpi.compareECLFiles_flow+9_4C_WINJ_GINJ_VREP-W_REIN-G_MSW ok
mpi.compareECLFiles_flow+9_4D_WINJ_GINJ_GAS_EXPORT_MSW ok
mpi.compareRestartedSim_flow+3D_MSW ??
mpi.compareECLFiles_flow+MOD4_GRP (92 != 85)
```

GitPaean:
benchmark please

opm/simulators/wells/MSWellHelpers.hpp
```diff
@@ -206,7 +206,7 @@ namespace mswellhelpers
     template <typename ValueType>
     ValueType velocityHead(const double area, const ValueType& mass_rate, const ValueType& density)
     {
-        return (0.5 * mass_rate * mass_rate / (area * area * density));
```
GitPaean, line 209: What is the reason to remove `0.5` here?

ytelses:
Benchmark result overview: 




View result details @ https://www.ytelses.com/opm/?page=result&id=776

alfbr:
The benchmarking service needs some TLC (I am on it). In the mean-time I am manually benchmarking this change  with very positive results. Convergence is significantly improved and computational time somewhat improved. This PR fixes several real issues on predictive cases, and is necessary for correctness. The sooner we can get the improvements here in master, the better. @GitPaean you can check the new cases in opm-tests that outputs the acceleration terms. The results from Flow was wrong. You find them here:
https://github.com/OPM/opm-tests/pull/331

GitPaean:
>  Convergence is significantly improved and computational time somewhat improved. 

That sounds good. As shown in the code comments, it was a known issue. This PR reduces the complexity in implementation by discarding some derivatives, which can be added later when necessary.  Benchmark is just for further checking. 

> The results from Flow was wrong.

Good. Then it is possible the formulations in the manual is buggy. We saw this not very often, but from time to time. The acceleration term is typically too small to verify before.

totto82:
About the restart test failure. 
The restart test does not fail if I add back the schedule info: WELSPECS, WELSEGS etc in 3D_MSW_RESTART.DATA and run with --sched-restart=true. 

@bska Is the HFA information in WELSEGS read from the restart file?

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -3681,26 +3697,39 @@ namespace Opm
...
+            gas_viscosity.clearDerivatives();
+            density.clearDerivatives();
         }
 
         const EvalWell liquid_emulsion_viscosity = mswellhelpers::emulsionViscosity(water_fraction, water_viscosity,
                                                      oil_fraction, oil_viscosity, sicd);
-        const EvalWell mixture_viscosity = (water_fraction + oil_fraction) * liquid_emulsion_viscosity + gas_fraction * gas_viscosities;
+        const EvalWell mixture_viscosity = (water_fraction + oil_fraction) * liquid_emulsion_viscosity + gas_fraction * gas_viscosity;
 
-        const EvalWell& reservoir_rate = segment_reservoir_volume_rates_[seg];
```
atgeirr, line 3695: The code looks good to me. It is a sore point that we have not conclusively agreed on the correct mathematical model, but rather ended up with agreeing on a code that gives good results...

The only thing I'd like changed is: please remove completely the `segment_reservoir_volume_rates_` member, which is not used any more.

atgeirr:
> The restart test does not fail if I add back the schedule info: WELSPECS, WELSEGS etc in 3D_MSW_RESTART.DATA and run with --sched-restart=true.

Then changing the test to do just that is a good idea for now I think. Also, then we have a test that will detect when/if support for `--sched-restart` changes.

opm/simulators/wells/MSWellHelpers.hpp
GitPaean: If you found out that the formulations was wrong. It will be worthwhile to add a comment here for later reference.

totto82: Strange, I thought I added a comment here. I think the reasoning is similar as for  the "2" in the friction term.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: ok

opm/simulators/wells/MSWellHelpers.hpp
GitPaean: Then please test it out with the commercial simulator. We keep the `0.5` for some formulations, like `valveContrictionPressureLoss`.  

Friction and acceleration are different properties of forces coming with two different formulations, not sure we can assume both formulations are wrong at the same time in case it was an accident typo in one of the formulations.

totto82:
> Then changing the test to do just that is a good idea for now I think. Also, then we have a test that will detect when/if support for --sched-restart changes.

OK. I will change the test in a separate PR that should be merged first.

totto82:

opm/simulators/wells/MSWellHelpers.hpp
totto82: > Then please test it out with the commercial simulator. 
That's what I did.

bska:
> @bska Is the HFA information in WELSEGS read from the restart file?

It is not.  Flow does not even **write** that information to the restart file&mdash;mostly because we don't know where to put it and what values to use.

It's **likely** that this information will be represented as a single integer in `IWEL` (one value per well, with regular wells having a trivial/sentinel value).  So far no-one has characterised the requisite items.

opm/simulators/wells/MSWellHelpers.hpp
GitPaean: > That's what I did.

That is very good. I do not have any reference example can check this value, it is very good you can look into it. 

But it comes a little bit surprise, because there is 0.5 \rho v^2 in Bernoulli's equation, not sure why the 0.5 is omitted here.

atgeirr:
jenkins build this opm-tests=340 please

opm/simulators/wells/MSWellHelpers.hpp
GitPaean: One of the possible explanations is that the unit conversion factor (similar to frictional pressure loss, which also has the wrong conversion factor) is wrong, it forces us to match the mistake. 

Since we are using SI unit, it is much easier to program with the correct formulation, since there is no unit conversion involved.

totto82:
jenkins build this opm-tests=340 please

totto82:
This works locally. I will try to update the data once more.

totto82:

totto82:
jenkins build this opm-tests=340 please

totto82:
When I tighten the rel_tol for restart locally slightly I reproduce the test failure locally. Lets see if jenkins is happy if I relax the tolerance a bit for restart with MSW.

totto82:
jenkins build this opm-tests=340 please

totto82:
jenkins build this opm-tests=340 please

compareECLFiles.cmake
```diff
@@ -742,11 +743,19 @@ add_test_compare_restarted_simulation(CASENAME spe9
                                       ABS_TOL ${abs_tol_restart}
                                       REL_TOL ${rel_tol_restart}
                                       TEST_ARGS --sched-restart=false)
+
+# The dynamic MSW data is not written to /read from the restart file
+# We therefore accept significant deviation in the results.
+# Note also that we use --sched-restart=true since some necessary
+# MSW info is still lacking in the restart file.
+set(rel_tol_restart_msw 2e2)
```
atgeirr, line 751: This should be abs_tol I think. A relative tolerance of 20000% is a bit much...

totto82:
jenkins build this opm-tests=340 please

compareECLFiles.cmake
totto82: It worries my that I need to relax the absolute tolerance to 2e2 for this. Locally it runs through without any adjustments. @atgeirr Maybe you could test this (without the last commit with the adjustments) locally on your system. It seems Jenkins does not like MSWs.

atgeirr: As long as the relative tolerance is coarse enough the absolute tolerance can be arbitrary small. I got it through with an absolute tolerance of 2e-12... With that, I tested the boundary between success and failure for the relative tolerance, and got success at 3e-4, failure at 2e-4. So 1e-3 is a bit course, but I think it is not a good idea to go very close to the boundary, so it is ok to keep it at that.

PR #2662, opened 2020-06-09
Serialize variant

joakim-hove:
The `std::variant<>` serialization could obviously be more general - but I started here....


joakim-hove:
jenkins build this please

PR #2663, opened 2020-06-09
fixed: fix parallel builds

akva2:
we need the update-version target to run to generate the
necessary headers.


akva2:
jenkins build this please

PR #2258, opened 2020-01-02
Add inner iterations for standard wells also

totto82:
Testing of #2238 seems to indicate that we sometimes for some reasons need to solve the well equations separately. Since more testing is needed before removing solveWellEq lets take it step-by-step. This PR adds inner iterations to STW similar to MSW. The inner iterations significantly improves the robustness of cases involving group control.

PR #2664, opened 2020-06-09
Adding inner iterations for standard wells. 

GitPaean:
This is following PR  https://github.com/OPM/opm-simulators/pull/2258  with some modifications. 
Separate running time arguments are introduced to handle standard wells instead of using the same with MSW wells. And the function `solveWellEqUntilConverged` from `WellInterface` is moved to `StandardWell`. Unless situation changes, this function should not be used by MS wells. 

Having inner iterations for standard wells is the correct way to go with the current understanding. 

The current PR is created to prepare for group control. 

Due to the small difference in the usage of the functions, this work turned out to be very hard to make it nice. Probably there will more refactoring with time going.


GitPaean:
jenkins build this please

GitPaean:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 0.754    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.724    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.047    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.031    |



View result details @ https://www.ytelses.com/opm/?page=result&id=766

GitPaean:
benchmark please

GitPaean:
> View result details @ https://www.ytelses.com/opm/?page=result&id=766

Is there some way we can check the result details?

alfbr:
> > View result details @ https://www.ytelses.com/opm/?page=result&id=766
> 
> Is there some way we can check the result details?

You mean like iteration counts? No, I am afraid that is not implemented. Currently you only get total run times against master for the four cases. However, this quick check tells you that looking at Norne is not sufficient, you need to investigate the impact on some of the other full field models you have accesss to in order to get a better picture of the impact of inner iterations.

GitPaean:
> You mean like iteration counts? No, I am afraid that is not implemented. Currently you only get total run times against master for the four cases. However, this quick check tells you that looking at Norne is not sufficient, you need to investigate the impact on some of the other full field models you have accesss to in order to get a better picture of the impact of inner iterations.

Sure. I am testing with the original model 2 locally to reproduce the results for flow_mpi_extra - Threads: 8

alfbr:
Remember that they use the relaxed tolerances (it speeds up the bench quite a bit), i.e., 1e-5 for mass balance, and 4 for max-strict-iter.

GitPaean:
>  ```opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.724-- | -- | --```

I reproduced the performance degeneration in this example. which is very unfortunate.

GitPaean:
just using this PR to test serial building.

GitPaean:

GitPaean:
jenkins build this please

GitPaean:
benchmark please

GitPaean:
by default, StandardWell will not use iterative solution process which is proposed in this PR, so it should affect running when run with default setup. 

Other development sees it is good to have option to do the iterative solution process.

atgeirr:
Looks good. I'll run through a local test to verify not a bit has changed in the output and then merge if ok.

PR #2671, opened 2020-06-13
update_data: Move List of Changed Tests to Commit Message Body

bska:
This way we are less at risk of overrunning GitHub's 256 character limit for issue/PR title lengths when a PR happens to affect many regression tests.


bska:
Note: I ran into the 256 character limit when trying to add new reference data for OPM/opm-common#1408.

bska:
jenkins build this please

bska:
Merging per maintainer approval.

PR #2672, opened 2020-06-15
changed: delay deck dereference

akva2:
this was harmless unless enableExperiments was given, since
the deck ref is not used otherwise, but it caused an UBSAN issue
when run in parallel.

### DIFF
--- a/ebos/eclthresholdpressure.hh
+++ b/ebos/eclthresholdpressure.hh
@@ -279,7 +279,6 @@ private:
         const auto& gridView = vanguard.gridView();
         const auto& elementMapper = simulator_.model().elementMapper();
         const auto& eclState = simulator_.vanguard().eclState();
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

blattms:
I would love to learn why there would be undefined behavior. We do not seem to change the deck between the two touched lines, do we?

akva2:
on non-root processes, the deck is a nullptr. This is dereferenced since deck() returns a const ref.

atgeirr:
Good in any case to have the variable in the scope it is used. Merging!

blattms:
??? Does that mean that `enableExperiments` can only be true on rank 0?

akva2:
as such no. enableExperiments things are not something that has been addressed for parallel usage. several of those are in fact still broken in parallel. Nothing changes there, but at least you do not get ubsan issues running without it now.

Issue #2148, opened 2019-11-07
2019-10-RC3 TRANX/TRANY/TRANZ Restart File Bug **Important**


OPMUSER:
There is a problem with the TRANX, TRANY and TRANZ arrays written to the RESTART file. This is a "write" issue as the simulation results are identical with the the previous release.

It looks like TRANY has been set to TRANZ and TRANZ has been set to TRANY.

Test deck attached for reference.
[MOD01-MULT1-OPM1910-R01.zip](https://github.com/OPM/opm-simulators/files/3819097/MOD01-MULT1-OPM1910-R01.zip)

joakim-hove:
I can look at this.

OPMUSER:
The riTranX/Y/Z arrays are correct.

The MULTX/Y/Z arrays are missing as well, as well as the MULTPV array

joakim-hove:
> The riTranX/Y/Z arrays are correct.

Ehhh - this I did not understand?

joakim-hove:
OK - when I look at this I find:

1. `TRANX` different from `TRANY` - I assume they should be ~similar.
2. `TRANZ` == 0

Is that the same as you see?

OPMUSER:
Give me a minute and I'll sent you some pictures.

OPMUSER:
TRANX and TRANY comparison

![MOD01_MULT1_ECL2004_R01_3D_View_TRANX_00_01_Jan_2019](https://user-images.githubusercontent.com/17958927/68387509-6f784b00-0199-11ea-86bd-c650f1be9def.png)

And TRANZ comparison
![MOD01_MULT1_ECL2004_R01_3D_View_TRANX_00_01_Jan_2019](https://user-images.githubusercontent.com/17958927/68387620-ad756f00-0199-11ea-9595-82ed9e0fb7b5.png)
Does that help?

joakim-hove:
> Does that help?

Yes - left column (correct) is from previous release?

OPMUSER:
To clarify, the left column is from the commercial simulator so that is correct. When I restored the 2019-04 run, it appears different. In this case TRANX is correct, but TRANY=TRANZ with TRANZ being correct.

OPMUSER:
This is a 2D cross-section so there is no flow in the y-direction, so TRANY should always be zero.

joakim-hove:
> This is a 2D cross-section so there is no flow in the y-direction, so TRANY should always be zero.

Right!

[...]

PR #2673, opened 2020-06-15
[bugfix] Fix TRANXYZ for problems with less than 3 dimensions.

blattms:
Up to now We always assumed that cardDims[i]>1 holds. which it does for most of the cases. But when e.g. simulating a vertical stack of 5 cells flow would report the transmissibilities in the Z direction in TRANX and output TRANZ as zero. Similar problems should be there for 2D grids.

With this commit we actually check whether there can be neighbors in the X and Y direction to prevent this behavior.

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -530,7 +530,7 @@ private:
                 int gc1 = std::min(cartMapper.cartesianIndex(c1), cartMapper.cartesianIndex(c2));
                 int gc2 = std::max(cartMapper.cartesianIndex(c1), cartMapper.cartesianIndex(c2));
 
-                if (gc2 - gc1 == 1) {
+                if (gc2 - gc1 == 1 && cartDims[0] > 1) {
                     if (tranx_deckAssigned)
                         // set simulator internal transmissibilities to values from inputTranx
                         trans_[isId] = inputTranxData[c1];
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please

blattms:
might also fix parts of #2148

blattms:
All the failing tests
-   mpi.compareECLFiles_flow+2D_OW_CTAQUIFER
-   mpi.compareECLFiles_flow+2D_FETKOVICHAQUIFER
-    mpi.compareECLFiles_flow+2D_OILWATER_POLYMER
-   mpi.compareECLFiles_flow+2D_H__
-    mpi.compareECLFiles_flow+2D_POLYMER_INJECTIVITY
-   mpi.compareECLFiles_flow+2D_THREEPHASE_POLY_HETER
-    mpi.compareECLFiles_flow+BC_LAB

are 2D problems with dimension 1 in the Y-Direction.  I would assum that all of these have some non-negative TRANY before this PR and now TRANY is always zero.

blattms:
Somebody else should double check this please.

bska:
> Somebody else should double check this please.

I'll do it later today.  If we need to update the reference data, then I'm going to hold off on doing so until we've merged OPM/opm-common#1408 and OPM/opm-tests#333.

blattms:
Great. Thanks.

bska:
I'm going to update the pertinent `.INIT` files in the reference data set to account for this data reshuffling.

bska:

bska:
The new reference data candidate contains only those changes we expected.  I'm going to rerun the build check with updated reference data from OPM/opm-tests#335 and merge this if it comes back green.

bska:
jenkins build this opm-tests=335 please

bska:
The build check targeting the updated reference data is was successful and the new data is installed in the opm-tests repository.

I am going to merge this into master now.  Thanks a lot for the effort and analysis!

PR #2675, opened 2020-06-16
change MSW restart test to read the schedule

totto82:
Testing of #2661 indicates that something is not working as expected when restarting MSW without reading the schedule part. This modifies the test to let #2661 through.


totto82:
jenkins build this opm-tests=339 please

totto82:
If I understand @bska correctly. Supporting writing/reading HFA info is a bit ahead in time. I suggest merging this now. Any objections?

atgeirr:
> Any objections?

None, I will merge.

PR #2678, opened 2020-06-19
Add actionState argument also for non MPI branch

joakim-hove:
See: https://github.com/OPM/opm-common/pull/1823

This patch (from @bska) recovers serial build on my machine


joakim-hove:

PR #2679, opened 2020-06-19
make parseSuccess dependent on MPI

akva2:
avoids unused warnings when building without


akva2:
jenkins build this please

atgeirr:
I have to admit the logic there is a bit messy. Maybe we should make some more functions there... But your fix is fine, will merge.

PR #2683, opened 2020-06-23
store polymer rates for output

totto82:


GitPaean:
jenkins build this opm-common=1834 please

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
```diff
@@ -817,6 +824,24 @@ namespace Opm
...
+        /// One rate pr well connection.
+        std::vector<double>& perfRatePolymer() { return perfRatePolymer_; }
+        const std::vector<double>& perfRatePolymer() const { return perfRatePolymer_; }
+
+        /// One rate pr well
+        double polymerWellRate(const int w) const {
+            int connpos = 0;
+            for (int iw = 0; iw < w; ++iw) {
+                connpos += this->well_perf_data_[iw].size();
+            }
```
bska, line 836: If we're going to do this a lot there should really be a precomputed lookup table for the start of the perforations of each well.

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
```diff
@@ -817,6 +824,24 @@ namespace Opm
...
+            double polymer_well_rate = 0.0;
+            const int endperf = connpos + this->well_perf_data_[w].size();
+            for (int perf = connpos; perf < endperf; ++perf ) {
+                polymer_well_rate += perfRatePolymer_[perf];
+            }
+            return polymer_well_rate;
```
bska, line 842:
```C++
const auto endperf = connpos + this->well_perf_data_[w].size();

return std::accumulate(&perfRatePolymer_[0] + connpos, &perfRatePolymer_[0] + endperf, 0.0);
```

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
totto82: There is already a first_perf_index_ vector in BlackoilWellModel that passes the corresponding first_perf to the WellInterface. It is only (I think) used to access the perforation data from the wellState.  Maybe we should make a member in the wellState instead?

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
totto82: OK. I just copied these lines from the solvent code above. I can change this (and the similar lines in the solvent stuff). I assume there is some performance gain from using the std algorithms.

bska: > I just copied these lines from the solvent code above.

I see.

> I can change this (and the similar lines in the solvent stuff).

Yes please.

> I assume there is some performance gain from using the std algorithms.

Not really in this case.  `std::accumulate` is going to instantiate pretty much the same loop that you wrote, but there is substantial maintainability gains from using standard algorithms where possible because it communicates **intent** much more clearly than a raw loop.

atgeirr: For these very simple loops, I do not see much of a read/intent advantage. However, when gradually moving towards making all parts parallel, algorithms are more easily replaced with parallel versions, so it is a bit more future-prepared.

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
bska: >> If we're going to do this a lot there should really be a precomputed lookup table for the start of the perforations of each well.
>
> Maybe we should make [`BlackoilWellModel::first_perf_index_`] a member in the wellState instead?

If that's feasible then I think that would be a good idea, at least for as long as the well state objects maintains the states of all wells as opposed to the state of a single well (cf. #2582).

totto82: I have kept the first_perf in wellInteface, but moved it from BlackoilWellModel to the BlackoilWellState.

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
totto82: The code indeed looks cleaner after @bska suggestions. Thanks.

bska:
jenkins build this opm-common=1834 please

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
```diff
@@ -131,6 +131,8 @@ namespace Opm
             perf_skin_pressure_.resize(nperf, 0.0);
 
             int connpos = 0;
+            first_perf_index_.resize(nw+1, 0);
```
GitPaean, line 134: `connpos` related is from the old `Wells` struct. With the new `first_perf_index_`, the usage of `connpos` can probably be removed.  

And also, unless we use `first_perf_index_` to get the number of connection/perforation for a well, we do not need to make its size `nw+1`. But it really depends how you use it.

totto82:
#include <numeric> is added

totto82:
jenkins build this opm-common=1834 please

bska:
Build check was successful.  I'll merge this into master now.  Thanks a lot.

PR #2684, opened 2020-06-24
use rates from previous time-step in the guiderate conversion 

totto82:
This reduce oscillations on the big model. 
Needs some more testing before it can be merged.


totto82:
jenkins build this please

totto82:
I have investigated the test failures and some variants of these and have the following comments:

1.  This PR does not introduce any significant changes to the results 
2.  This PR does not have any obvious impact on convergence on the failed models. 
3.  The difference in the results from this PR on the failing test are not conclusive wrt. this being the right thing to do.

We are currently looking closer at the effect of this on bigger models.

alfbr:
This PR fixes a real issue, and has been tested thoroughly at our side. Please review/merge.

atgeirr:
I now agree with this PR. It can be merged when green. I'll start a data update.

Long term, we probably want guide rates to be implicit, calculated with full AD. But for now that is not an option.

atgeirr:

atgeirr:
jenkins build this opm-tests=345 please

PR #2689, opened 2020-06-26
Capture the variables argc, argv, outputCout, and outputFiles in Opm::FlowMainEbos

hakonhagland:
Make `Opm::FlowMainEbos` capture the variables `argc`, `argv`, `outputCout`, and `outputFiles`. Passing the variables to the constructor and saving them as class variables in `Opm::FlowMainEbos` makes the implementation of the Python interface simpler. For example, the `step_init()` method (see PR #2690) does not need to ask `Opm::Main` about the values of the variables when it needs to run `execute()` in `FlowMainEbos`.

Another advantage of this refactoring could be that less variables needs to be passed around from `Opm::Main`, to the static flow modules `flow_ebos_xxx.cpp`, and then again to `FlowMainEbos`.

### DIFF
--- a/opm/simulators/flow/FlowMainEbos.hpp
+++ b/opm/simulators/flow/FlowMainEbos.hpp
@@ -79,6 +79,13 @@ namespace Opm
 
         typedef Opm::SimulatorFullyImplicitBlackoilEbos<TypeTag> Simulator;
 
+        FlowMainEbos(int argc, char **argv, bool output_cout, bool output_files )
+            : argc_{argc}, argv_{argv},
+              output_cout_{output_cout}, output_files_{output_files}
+        {
+
+        }
+
         // Read the command line parameters. Throws an exception if something goes wrong.
         static int setupParameters_(int argc, char** argv)
         {
@@ -291,10 +298,9 @@ namespace Opm
         /// This is the main function of Flow.  It runs a complete simulation with the
         /// given grid and simulator classes, based on the user-specified command-line
         /// input.
-        int execute(int argc, char** argv, bool output_cout, bool output_to_files)
+        int execute()
         {
-            return execute_(argc, argv, output_cout, output_to_files,
-                &FlowMainEbos::runSimulator, /*cleanup=*/true);
+            return execute_(&FlowMainEbos::runSimulator, /*cleanup=*/true);
         }
 
         // Print an ASCII-art header to the PRT and DEBUG files.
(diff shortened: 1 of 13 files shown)


bska:
jenkins build this please

opm/simulators/flow/FlowMainEbos.hpp
```diff
@@ -343,33 +349,32 @@ namespace Opm
...
+                    executeCleanup_();
                 }
                 return exitCode;
             }
             catch (const std::exception& e) {
                 std::ostringstream message;
                 message  << "Program threw an exception: " << e.what();
 
-                if (output_cout) {
+                if (output_cout_) {
```
joakim-hove, line 377: Cool if you use `this->output_cout_` to separate better between member variable and argument.

joakim-hove:
OK - one minor comment about use of `this->output_cout_` - then I am ready to merge this.

opm/simulators/flow/FlowMainEbos.hpp
hakonhagland: Thanks for the quick review. Yes it is good to separate between local variables in the method and member variables of the class. I completely agree.  But... what about the underscore? I thought we used a trailing underscore to signify that a variable was a private member variable? So then it is quite clear from the name `output_cout_` that it has to be a member variable, I wonder if it then is too verbose to also add `this` to indicate the same? For a public member variable we would not use a trailing underscore, right? Then it would definitively be cool to add `this` to clarify (since now the trailing underscore is missing).. but is it needed for a private member variable (with trailing underscore) ?

joakim-hove: Well; for good and for bad we do not have any style guides which people really follow in this codebase. Personally I *always* use `this->` in code I write - and like that. For this particular case the variable `output_cout` is used as function/constructor argument in many places - I therefor think concious use of `this->`will increase readability much more than just the trailing underscore `_`.

hakonhagland: > For this particular case the variable output_cout is used as function/constructor argument in many places

I could only see that it is used in the constructor (as an argument). The other places in the file it is used as a member variable. And the usage as an argument in the constructor should not count since the constructor is special and it should not be confusing to use both `output_cout` as an input argument and and then initialize `output_cout_` from `output_cout` since this is commonly what a constructor does?

Anyway, you are the boss so I will follow your suggestion and update the PR. Once I get used to it, I might see that `this->output_cout_` is actually more readable than `output_cout_`

hakonhagland:
jenkins build this please

joakim-hove:
Thanks for updating. Merging

PR #2691, opened 2020-06-27
EclWriter: Refactor Cell Loop out to Helper Function

bska:
That way we can use the same code for both the `writeOutput` and the `evalSummaryState` member functions.  While here, also change the declared types of `wellData` and `groupData` to `auto` in preparation of API updates for collecting group- and well level guiderate data.

Finally, apply `const` where possible.

### DIFF
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -233,7 +233,7 @@ public:
 
     void evalSummaryState(bool isSubStep)
     {
-        int reportStepNum = simulator_.episodeIndex() + 1;
+        const int reportStepNum = simulator_.episodeIndex() + 1;
         /*
           The summary data is not evaluated for timestep 0, that is
           implemented with a:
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska:
jenkins build this please

ebos/eclwriter.hh
```diff
@@ -254,33 +254,21 @@ public:
...
-        ElementContext elemCtx(simulator_);
-        ElementIterator elemIt = gridView.template begin</*codim=*/0>();
-        const ElementIterator& elemEndIt = gridView.template end</*codim=*/0>();
-        for (; elemIt != elemEndIt; ++elemIt) {
-            const Element& elem = *elemIt;
-            elemCtx.updatePrimaryStencil(elem);
-            elemCtx.updatePrimaryIntensiveQuantities(/*timeIdx=*/0);
-            eclOutputModule_.processElement(elemCtx);
-        }
+        this->prepareLocalCellData(isSubStep, reportStepNum);
```
atgeirr, line 267: You can move this call above the creation of localWellData and localGroupData.

ebos/eclwriter.hh
```diff
@@ -743,14 +695,80 @@ private:
...
+
+        // Add suggested next timestep to extra data.
+        if (! isSubStep) {
+            restartValue.addExtra("OPMEXTRA", std::vector<double>(1, nextStepSize));
+        }
+
+        // first, create a tasklet to write the data for the current time
+        // step to disk
+        auto eclWriteTasklet = std::make_shared<EclWriteTasklet>(
+            this->actionState(), this->summaryState(), *this->eclIO_,
```
atgeirr, line 752: Please reformat with one argument per line.

ebos/eclwriter.hh
```diff
@@ -254,33 +254,21 @@ public:
...
-        Scalar curTime = simulator_.time() + simulator_.timeStepSize();
-        Scalar totalCpuTime =
+        const Scalar curTime = simulator_.time() + simulator_.timeStepSize();
+        const Scalar totalCpuTime =
             simulator_.executionTimer().realTimeElapsed() +
             simulator_.setupTimer().realTimeElapsed() +
             simulator_.vanguard().externalSetupTime();
 
-        Opm::data::Wells localWellData = simulator_.problem().wellModel().wellData();
+        const auto localWellData  = simulator_.problem().wellModel().wellData();
```
atgeirr, line 263: Is it necessary to copy these data? Can they be used by const reference? (In writeOutput() you need to move them out, so there I see the need.)

bska: > Is it necessary to copy these data?

Well, not really, but making this `const auto&` is technically "just" lifetime extension in this case.  `wellModel().wellData()` returns an object (of type `data::Wells` which is an alias of [`data::WellRates`](https://github.com/OPM/opm-common/blob/29a946eb9aefaf0f5551d1e713353a2a3de6e39b/opm/output/data/Wells.hpp#L301)) rather than a reference to one.  By making `localWellData` an actual object instead of a reference I'm relying on NRVO to construct the return value in place.

I recently learned that lifetime extension is typically implemented compilers using something close to
```C++
void* __s = __alloca(sizeof(T));  // Magic function with automatic deallocation
const T* __p = new(__s) T(std::move(return_value));
const T& x = *__p;

use(x);

__p->~T();
```
and this has made lifetime extension less appealing to me.

ebos/eclwriter.hh
bska: > You can move this call above the creation of localWellData and localGroupData.

I'll do that the next time I'm touching this file.

ebos/eclwriter.hh
bska: > Please reformat with one argument per line.

I'll do that the next time I'm touching this file.

PR #2693, opened 2020-06-30
Add oil water brine specialization

totto82:
This adds a oil water brine specialization. 

It also change the scaling of the brine residual from b_g to b_w.


totto82:
jenkins build this please

PR #2695, opened 2020-06-30
Add brine to output

totto82:
Depends on OPM/opm-common#1849

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -699,12 +699,16 @@ namespace Opm
             if (has_brine) {
                 // TODO: the application of well efficiency factor has not been tested with an example yet
                 const unsigned waterCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::waterCompIdx);
-                EvalWell cq_s_sm = cq_s[waterCompIdx] * well_efficiency_factor_;
+                EvalWell cq_s_sm = cq_s[waterCompIdx];
                 if (this->isInjector()) {
                     cq_s_sm *= wsalt();
                 } else {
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -699,12 +699,16 @@ namespace Opm
             if (has_brine) {
                 // TODO: the application of well efficiency factor has not been tested with an example yet
                 const unsigned waterCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::waterCompIdx);
-                EvalWell cq_s_sm = cq_s[waterCompIdx] * well_efficiency_factor_;
```
atgeirr, line 702: I think this change looks correct, if we assume that the well_state stores raw rates that do not consider efficiency, and the connectionRates_ contain the exchange with the reservoir including efficiency. But then the same change should be done for the other model options as well, I think.

If you agree, please modify foam and polymer as well. The energy term seems to ignore efficiency. I think the part that gets stored in connectionRates_ should be multiplied with efficiency to be consistent here, but would appreciate if you also could take a look at that.

totto82: The polymer is already fixed. And currently we don't output foam.

atgeirr: You are right about the polymer, I misread that. The foam treatment should still be changed, to be correct when we do want to output it: it may become a hard-to-find bug otherwise.

totto82: I don't think I should change the foam related stuff in this PR. There is a comment about this in the polymer and solvent part right above

atgeirr: Ok, merging.

PR #2696, opened 2020-07-01
Simulator Group Data: Chase Upstream API Update

bska:
This commit makes the 'groupData()' function return a
```C++
map<string, Opm::data::GroupData>
```
object instead of a
```C++
map<string, Opm::data::GroupConstraints>
```
object.  The `GroupData` structure adds a level of indirection to the current per-group summary quantities that are directly assigned by the simulator.  While here, also move the assignment of the current group constraints/control values out to a separate helper to reduce the body of the per-group loop in `groupData()`.

This is in preparation of adding support for reporting group-level production/injection guiderates (`G[OGW][IP]GR`) to the summary file.

---

Note: This PR is the companion to upstream PR OPM/opm-common#1853.  This PR **must not** be merged without also merging the upstream PR lest there be a build break.

### DIFF
--- a/ebos/collecttoiorank.hh
+++ b/ebos/collecttoiorank.hh
@@ -549,41 +549,43 @@ public:
 
     class PackUnPackGroupData : public P2PCommunicatorType::DataHandleInterface
     {
-        const Opm::data::Group& localGroupData_;
-        Opm::data::Group& globalGroupData_;
+        const Opm::data::GroupValues& localGroupData_;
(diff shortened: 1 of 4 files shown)


bska:
jenkins build this opm-common=1853 please

bska:
jenkins build this opm-common=1853 please

bska:
jenkins build this opm-common=1853 please

PR #2698, opened 2020-07-02
Pass the salt concentration to the water PVT

totto82:
This must be merged together with OPM/opm-material#423

I have also added BDENSITY and SALTVD to the missing feature list. They are only partially implemented and never used.


totto82:
jenkins build this opm-material=423 please

totto82:
The test failure is expected. The brine test only inject salt for a small time so the effect is minimal. I have also tested an example where the salt changes the water pvt more and results look reasonable.

totto82:

totto82:
I want to self merge this and the opm-material changes later today if nobody objects. It fixes bug in the brine model and should not affect the rest of the code.

totto82:
jenkins build this opm-tests=346 please

totto82:
jenkins build this opm-material=423 update_data please

bska:
jenkins build this opm-material=423 opm-tests=347 please

totto82:
jenkins build this opm-material=423 opm-tests=347 please

bska:
I'm merging this now.  Thanks a lot.

PR #2699, opened 2020-07-02
Implement SALTVD

totto82:
This PR implements SALTVD

Note it is build on top of #2698 for convenience.  i.e. the two first commits should be review as part of #2698.

The current SPE1_BRINE test has 0 for saltvd. I plan to adapt the test to include non-trivial SALTVD. 

I have tested this. And it gives the expected results.

### DIFF
--- a/ebos/equil/equilibrationhelpers.hh
+++ b/ebos/equil/equilibrationhelpers.hh
@@ -612,6 +612,8 @@ private:
  */
 class EquilReg
 {
+    using TabulatedFunction = typename Opm::Tabulated1DFunction<double>;
+
 public:
     /**
      * Constructor.
@@ -624,10 +626,12 @@ public:
     EquilReg(const Opm::EquilRecord& rec,
              std::shared_ptr<Miscibility::RsFunction> rs,
              std::shared_ptr<Miscibility::RsFunction> rv,
+             const TabulatedFunction& saltVdTable,
              const int pvtIdx)
         : rec_    (rec)
         , rs_     (rs)
         , rv_     (rv)
+        , saltVdTable_ (saltVdTable)
         , pvtIdx_ (pvtIdx)
     {}
 
@@ -701,6 +705,8 @@ public:
     const CalcEvaporation&
     evaporationCalculator() const { return *this->rv_; }
 
+    const TabulatedFunction&
+    saltVdTable() const { return saltVdTable_;}
     /**
      * Retrieve pvtIdx of the region.
      */
@@ -710,6 +716,7 @@ private:
     Opm::EquilRecord rec_;     /**< Equilibration data */
     std::shared_ptr<Miscibility::RsFunction> rs_;      /**< RS calculator */
     std::shared_ptr<Miscibility::RsFunction> rv_;      /**< RV calculator */
+    const TabulatedFunction& saltVdTable_;
     const int pvtIdx_;
 };
 
--- a/ebos/equil/equilibrationhelpers.hh
+++ b/ebos/equil/equilibrationhelpers.hh
@@ -612,7 +612,7 @@ private:
  */
 class EquilReg
 {
-    using TabulatedFunction = typename Opm::Tabulated1DFunction<double>;
+    using TabulatedFunction = Opm::Tabulated1DFunction<double>;
 
 public:
     /**
--- a/ebos/equil/initstateequil.hh
+++ b/ebos/equil/initstateequil.hh
@@ -178,7 +178,8 @@ private:
     density(const double depth,
             const double press) const
     {
-        double saltConcentration = saltVdTable_.eval(depth);
+        // The initializing algorithm can give depths outside the range due to numerical noise i.e. we extrapolate
+        double saltConcentration = saltVdTable_.eval(depth, /*extrapolate=*/true);
         double rho = FluidSystem::waterPvt().inverseFormationVolumeFactor(pvtRegionIdx_, temp_, press, saltConcentration);
         rho *= FluidSystem::referenceDensity(FluidSystem::waterPhaseIdx, pvtRegionIdx_);
         return rho;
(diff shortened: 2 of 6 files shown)


totto82:
I have rebased this on current master.

totto82:
Currently, we don't allow for extrapolation of the SALTVD table (as for RSVD, RVVD). I think it should stay that way. But currently the SALTVD table in the SPE1CASE1_BRINE is outside of the reservoir and needs to be fixed, i.e. this needs to merge together with OPM/opm-tests#348.

totto82:
jenkins build this opm-tests=348 please

ebos/equil/initstateequil.hh
```diff
@@ -1721,10 +1730,30 @@ private:
...
 
+    template <class RMap>
+    void updateInitialSaltConcentration_(const Opm::EclipseState& eclState, const RMap& reg, const Grid& grid)
+    {
+        const auto& tables = eclState.getTableManager();
+        const Opm::TableContainer& saltvdTables = tables.getSaltvdTables();
+        for (size_t i = 0; i < saltvdTables.size(); ++i) {
+            const Opm::SaltvdTable& saltvdTable = saltvdTables.getTable<Opm::SaltvdTable>(i);
+            using TabulatedFunction = typename Opm::Tabulated1DFunction<double>;
+            static TabulatedFunction saltvdTableFunction;
```
bska, line 1741: Just out of curiosity, why is this marked as `static`?  Could you not create the interpolant outside of the outer loop to get the same effect (constructing the interpolant once)?

On another note, I recommend looping the [`activeRegions()`](https://github.com/OPM/opm-grid/blob/fb70c95d5cfc8b1ea62f2f81195557c143a17e47/opm/grid/utility/RegionMapping.hpp#L88) instead of all regions here, e.g.,
```C++
template <class RMap>
void updateInitialSaltConcentration_(const Opm::EclipseState& eclState, const RMap& reg, const Grid& grid)
{
    const auto& tables = eclState.getTableManager();
    const auto& saltvcTables = tables.getSaltvdTables();

    auto saltvdTableFunction = Opm::Tabulated1DFunction<double>{};
    for (const auto& regID : reg.activeRegions()) {
        const auto& saltvdTable = saltvdTables.getTable<Opm::SaltvdTable>(regID);
        saltvdTableFunction.setXYContainers(saltvdTable.getDepthColumn(), saltvdTable.getSaltColumn());

        for (const auto& cell : reg.cells(regID)) {
            const auto depth = UgGridHelpers::cellCenterDepth(grid, cell);
            this->saltConcentration_[cell] = saltvdTableFunction.eval(depth);
        }
    }
}
```

ebos/equil/initstateequil.hh
```diff
@@ -166,14 +169,14 @@ public:
...
+    const double saltConcentration_;
     const int pvtRegionIdx_;
     const double g_;
 
     double
     density(const double press) const
     {
-        double saltConcentration = 0.0; // TODO allow for non-zero initial salt concentration
-        double rho = FluidSystem::waterPvt().inverseFormationVolumeFactor(pvtRegionIdx_, temp_, press, saltConcentration);
+        double rho = FluidSystem::waterPvt().inverseFormationVolumeFactor(pvtRegionIdx_, temp_, press, saltConcentration_);
```
bska, line 179: This will use a constant salt concentration for all cells in an equilibration region.  Is that intended?  If not, do you want to fix it now or at a later time?

totto82: The salt concentration should be pr. cell. I miss read the code. Thanks for noticing. I will update the code accordingly.

totto82:
jenkins build this opm-tests=348 please

totto82:
jenkins build this opm-tests=348 please

ebos/equil/initstateequil.hh
```diff
@@ -1721,10 +1734,42 @@ private:
...
+        const auto& tables = eclState.getTableManager();
+        const Opm::TableContainer& saltvdTables = tables.getSaltvdTables();
+
+        // If no saltvd table is given, we create a trivial table for the density calculations
+        if (saltvdTables.empty()) {
+            std::vector<double> x = {0.0,1.0};
+            std::vector<double> y = {0.0,0.0};
+            for (size_t i = 0; i < numEquilReg; ++i) {
+                saltVdTable_[i].setXYContainers(x,y);
+            }
```
bska, line 1751: You don't actually need the index here&mdash;and as written it produces a "signed-vs-unsigned" comparison [warning](https://ci.opm-project.org/job/opm-simulators-PR-builder/1130/gcc/new/fileName.1081815698/) because `numEquilReg` is `int`&mdash;since you could equally well write this loop as
```C++
for (auto& table : this->saltVdTable_) {
    table.setXYContainers(x, y);
}
```

ebos/equil/initstateequil.hh
```diff
@@ -1721,10 +1734,42 @@ private:
...
+                    const double depth = UgGridHelpers::cellCenterDepth(grid, cell);
+                    this->saltConcentration_[cell] = saltVdTable_[i].eval(depth);
+                }
+            }
+        }
+    }
+
     std::vector< std::shared_ptr<Miscibility::RsFunction> > rsFunc_;
     std::vector< std::shared_ptr<Miscibility::RsFunction> > rvFunc_;
+    using TabulatedFunction = typename Opm::Tabulated1DFunction<double>;
```
bska, line 1768: You don't need `typename` here.

ebos/equil/initstateequil.hh
```diff
@@ -148,31 +149,37 @@ namespace PhasePressODE {
 template <class FluidSystem>
 class Water
 {
+using TabulatedFunction = typename Opm::Tabulated1DFunction<double>;
```
bska, line 152: You don't need `typename` here.

ebos/equil/equilibrationhelpers.hh
```diff
@@ -612,6 +612,8 @@ private:
  */
 class EquilReg
 {
+    using TabulatedFunction = typename Opm::Tabulated1DFunction<double>;
```
bska, line 615: You don't need `typename` here.  There is nothing in the right-hand-side expression that depends on a template parameter.

ebos/equil/initstateequil.hh
```diff
@@ -523,6 +530,7 @@ private:
     double gravity_;
     int    nsample_;
     double temperature_{ 273.15 + 20 };
+    double saltConcentration_{0.0};
```
bska, line 533: Is this datamember still used?  If not it should be removed.

totto82:
Thanks to @bska for review. I have followed your suggestions.

totto82:
jenkins build this opm-tests=348 please

bska:
> Due to [OPM/opm-common#1876](https://github.com/OPM/opm-common/pull/1876) being merged in the interim it is possible that we will have to re-update the reference solutions.

The build check failed, and in the way I thought would happen if the reference data update (OPM/opm-tests#348) were based on a stale reference.  I'm going to close that PR for a while and then re-update the reference solutions based on this PR.

bska:
jenkins build this opm-tests=348 please

PR #1853, opened 2019-05-22
Add keywords to the list of unsupported keywords

joakim-hove:
See: https://github.com/OPM/opm-common/pull/779

PR #2700, opened 2020-07-02
Rename data::Group -> data::Groups

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/1858


bska:
I'd prefer not to merge this. PR #1853 does something similar, but is also more comprehensive.

joakim-hove:
I'll close this - if it comes through green I will merge the remains of OPM/opm-common#1858 - which was unrelated.

PR #2704, opened 2020-07-06
Cell Data: Don't Warn About Supported Output Directives

bska:
This commit ensures that we don't issue misleading warning about output creation directives such as

 * `BASIC`
 * `FREQ`
 * `RESTART`

that could be, and typically are, put into keywords like `RPTRST` and `RPTSCHED`.  Previously, we would interpret these directives as names of cell vectors and we would get very confusing diagnostic messages of the form
```
Keyword 'BASIC' is unhandled for output to file
```
which suggests that `BASIC` is unsupported despite the output system supporting the directive for quite some time.  While here, also apply the same reasoning to the vectors we always output such as `SWAT`, `SGAS`, or `PRESSURE`.

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -221,8 +221,16 @@ public:
         // Only output RESTART_AUXILIARY asked for by the user.
         const Opm::RestartConfig& restartConfig = simulator_.vanguard().schedule().restart();
         std::map<std::string, int> rstKeywords = restartConfig.getRestartKeywords(reportStepNum);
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

atgeirr:
Looks good, merging!

PR #2705, opened 2020-07-09
Add WListManager argument when constructing ActionContext

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/1864


bska:

PR #2707, opened 2020-07-10
Fix some warnings in the OpenCl Solver.

blattms:
PR #2682 introduced some new warnings (shadowing variables, and catching exceptions by value). This PR fixes them.

Would be nice if @Tongdongq would tell his 2 cents on this (especially the shadowed variables).

### DIFF
--- a/opm/simulators/linalg/bda/WellContributions.cpp
+++ b/opm/simulators/linalg/bda/WellContributions.cpp
@@ -92,7 +92,7 @@ void WellContributions::apply(cl::Buffer& d_x, cl::Buffer& d_y) {
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please

Tongdongq:
Looks good.
I also found an 'unused variable' warning at `WellContributions.cpp:95`. The `addMatrix()` takes some arguments. These arguments are passed to `addMatrixGpu()` if CUDA is found, otherwise an exception is thrown. However, when CUDA is not found, the arguments are not used.

blattms:
fixed that also, thanks.

blattms:
jenkins build this please

blattms:
jenkins is happy and in it goes.

PR #2711, opened 2020-07-21
Remove unused variable groups from BlackoilWellModel::wellsToState.

blattms:
Alternatively it could be marked as [[maybe_unused]] if it will be used sometime in the future.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -403,7 +403,6 @@ namespace Opm {
 
             // convert well data from opm-common to well state from opm-core
             void wellsToState( const data::Wells& wells,
-                               const data::GroupValues& groups,
                                const PhaseUsage& phases,
                                const bool handle_ms_well,
                                WellStateFullyImplicitBlackoil& state ) const;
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please

blattms:
@GitPaean would you mind sharing your 2cents about this?

GitPaean:
It looks like something WIP before the summer. 

It is good that @joakim-hove and @bska can comment on it whether they are some current ongoing development about it. 

Otherwise, I think it is okay to go ahead to merge it. They are not difficult to add back anyway.

bska:
> It looks like something WIP before the summer.

It is.  These parameters were added in PR #2703 as a precursor to restoring various group-level quantities upon restart.  That, in turn, is part of the larger effort OPM/opm-common#1396 which aims to remove the `--sched-restart` option at some point in the future&mdash;thereby making Flow's restart model fully compatible with ECLIPSE.  As of PR #2634 we default to **not** reading the `SCHEDULE` data **prior** to the restart point from the `.DATA` file, but we're still able to do so by setting
```
--sched-restart=true
```
on the command line when restarting a simulation.

That said, the parameters mention here are obviously unused right now and they do generate a lot of warnings so I'd be in favour of removing them with the option of reintroducing the support later if actually needed.

GitPaean:
jenkins build this please

GitPaean:
> That said, the parameters mentioned here are obviously unused right now and they do generate a lot of warnings so I'd be in favour of removing them with the option of reintroducing the support later if actually needed.

Thanks for the comments. I am merging this PR now.

PR #2716, opened 2020-07-28
Opencl stdwell v3

ducbueno:
Fixed previous errors and now the add well contributions functionality is working on the OpenCL backend


blattms:
Does this supersede #2712? In that case please close it.

ducbueno:
Yes it supersede #2712. Forgot to close it.

blattms:
When trying this with Norne and my AMD card, I memory errors:
```
Entrei nesta porra
Initializing GPU, matrix size: 133293 blocks, nnzb: 313131
Maxit: 200, tolerance: 1.000000e-02
PlatformID: 0, deviceID: 0

Found 2 OpenCL platforms

Chosen:
Platform name      : AMD Accelerated Parallel Processing
Platform version   : OpenCL 2.0 AMD-APP (3137.0)

Found 1 OpenCL devices

Chosen:
CL_DEVICE_NAME            : gfx803
CL_DEVICE_VERSION         : OpenCL 1.2 

Memory access fault by GPU node-1 (Agent handle: 0x56438fc38d20) on address 0x210180e000. Reason: Page not present or supervisor privilege.
[smaug:13015] *** Process received signal ***
[smaug:13015] Signal: Aborted (6)
[smaug:13015] Signal code:  (-6)
[smaug:13015] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x12730)[0x7f17ce0cf730]
[smaug:13015] [ 1] /lib/x86_64-linux-gnu/libc.so.6(gsignal+0x10b)[0x7f17cac7e7bb]
[smaug:13015] [ 2] /lib/x86_64-linux-gnu/libc.so.6(abort+0x121)[0x7f17cac69535]
[smaug:13015] [ 3] /opt/rocm/lib/../opencl/lib/../../lib/libhsa-runtime64.so.1(+0x55f43)[0x7f17c1dfaf43]
[smaug:13015] [ 4] /opt/rocm/lib/../opencl/lib/../../lib/libhsa-runtime64.so.1(+0x54505)[0x7f17c1df9505]
[smaug:13015] [ 5] /opt/rocm/lib/../opencl/lib/../../lib/libhsa-runtime64.so.1(+0x131c7)[0x7f17c1db81c7]
[smaug:13015] [ 6] /lib/x86_64-linux-gnu/libpthread.so.0(+0x7fa3)[0x7f17ce0c4fa3]
[smaug:13015] [ 7] /lib/x86_64-linux-gnu/libc.so.6(clone+0x3f)[0x7f17cad404cf]
[smaug:13015] *** End of error message ***
```

For my NVIDIA card there seems to be a resource problem:
```
PlatformID: 1, deviceID: 0

Found 2 OpenCL platforms

Chosen:
Platform name      : NVIDIA CUDA
Platform version   : OpenCL 1.2 CUDA 10.1.353

Found 1 OpenCL devices

Chosen:
CL_DEVICE_NAME            : GeForce GTX 1060 6GB
CL_DEVICE_VERSION         : OpenCL 1.2 CUDA

ESC[31;1m
Error: [/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/openclSolverBackend.cpp:679] openclSolverBackend::solve_system error: clEnqueueReadBuffer(-5)
CL_OUT_OF_RESOURCESESC[0m
Program threw an exception: [/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/openclSolverBackend.cpp:679] openclSolverBackend::solve_system error: clEnqueueReadBuffer(-5)
CL_OUT_OF_RESOURCES
```

blattms:
When activating both CUDA and OpenCL I get compilation errors:
```
[  4%] Building CUDA object CMakeFiles/opmsimulators.dir/opm/simulators/linalg/bda/cusparseSolverBackend.cu.o
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.hpp(97): error: "d_Cnnzs" has already been declared in the current scope

/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.hpp(98): error: "d_Dnnzs" has already been declared in the current scope

/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.hpp(99): error: "d_Bnnzs" has already been declared in the current scope

/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.hpp(100): error: "d_Ccols" has already been declared in the current scope

/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.hpp(101): error: "d_Bcols" has already been declared in the current scope

/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.hpp(104): error: "d_val_pointers" has already been declared in the current scope

6 errors detected in the compilation of "/tmp/tmpxft_00003c52_00000000-6_cusparseSolverBackend.cpp1.ii".
make[3]: *** [CMakeFiles/opmsimulators.dir/build.make:310: CMakeFiles/opmsimulators.dir/opm/simulators/linalg/bda/cusparseSolverBackend.cu.o] Fehler 1
make[2]: *** [CMakeFiles/Makefile2:1584: CMakeFiles/opmsimulators.dir/all] Fehler 2
make[1]: *** [CMakeFiles/Makefile2:611: CMakeFiles/flow.dir/rule] Fehler 2
```

Please fix these.

blattms:
Concerning the the norne problem with the NVIDIA (via opencl):
The failed test was for the options `--gpu-mode=opencl --opencl-platform-id=1` (the latter argument is to select the correct card). If I pass `--gpu-mode=opencl --opencl-platform-id=1 --matrix-add-well-contributions=true` it seems to run through.

Tongdongq:
It looks like the user has to choose at compile time between CUDA or OpenCL. Shouldn't it be able to compile for both (if found by CMake) and use the runtime `--gpu-mode` argument?

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -461,7 +461,7 @@ DenseMatrix transposeDenseMatrix(const DenseMatrix& M)
                 bool use_gpu = bdaBridge->getUseGpu();
                 if (use_gpu) {
                     WellContributions wellContribs;
-                    if (!useWellConn_) {
+                    if (useWellConn_) {
```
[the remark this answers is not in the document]
ducbueno: I've removed the negation symbol in order to be able to run the simulation with the well contributions with the flag `--add-well-contributions=true`. It wasn't making much sense to me the addition of the well contributions to take place when the `--add-well-contributions` flag was set to false. But I could (with high probability) be misunderstanding why the flag should be set to false. Anyway, it wouldn't be a problem to change it back to the original version.

opm/simulators/linalg/bda/WellContributions.hpp
```diff
@@ -73,20 +73,14 @@ class WellContributions
     unsigned int num_ms_wells = 0;           // number of MultisegmentWells in this object, must equal multisegments.size()
     unsigned int num_blocks_so_far = 0;      // keep track of where next data is written
     unsigned int num_std_wells_so_far = 0;   // keep track of where next data is written
-    unsigned int *val_pointers = nullptr;    // val_pointers[wellID] == index of first block for this well in Ccols and Bcols
+    unsigned int *val_pointers;              // val_pointers[wellID] == index of first block for this well in Ccols and Bcols
```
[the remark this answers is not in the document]
ducbueno: No specific reason. I was fiddling with arrays and vectors at some point and I forgot to return it to the original. Again, it shouldn't be a problem to return it to the original.

opm/simulators/linalg/bda/WellContributions.cpp
```diff
@@ -83,22 +104,65 @@ void WellContributions::apply(cl::Buffer& d_x, cl::Buffer& d_y) {
...
+    *dim_ = dim;
+    *dim_wells_ = dim_wells;
+}
 
-    // apply StandardWells
-    if (num_std_wells > 0) {
-        OPM_THROW(std::logic_error, "Error StandardWells are not supported by openclSolver");
-    }
+void WellContributions::getData(double **valsC, double **valsD, double **valsB, int **colsC, int **colsB, unsigned int **val_pointers_){
+    std::copy(h_valsC, h_valsC + num_blocks*dim*dim_wells, *valsC);
```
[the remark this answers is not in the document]
ducbueno: I used `std::copy` because I couldn't work a way to concatenate the values passed via `*valsC` to `h_valsC` without it.

ducbueno:
I think I see why these errors occur when both CUDA and OpenCL are present. I think it is because the `#elif` statements. I wasn't getting these errors on the machine I'm currently working on because it doesn't have an NVidia card and, therefore, no CUDA. I think if remove the `#elif`s and allow CUDA and OpenCL to coexist these errors will cease to occur. Already working on it.

opm/simulators/linalg/bda/WellContributions.hpp
```diff
@@ -119,11 +121,19 @@ class WellContributions
...
     /// \param[in] stream           the cudaStream that is used to launch the kernel in
-#if HAVE_CUDA
     void setCudaStream(cudaStream_t stream);
+    
+    /// Apply all Wells in this object
+    /// performs y -= (C^T * (D^-1 * (B*x))) for all Wells
+    /// \param[in] d_x        vector x, must be on GPU
+    /// \param[inout] d_y     vector y, must be on GPU
+    void apply(double *d_x, double *d_y);
+#elif HAVE_OPENCL
```
blattms, line 134: As @Tongdongq already pointed out we can have CUDA and OpenCL. Hence please change to 
```
#endif 
#if HAVE_OPENCL
```
it is also not consistent with the [implementation](/OPM/opm-simulators/blob/be7fed413fc7e82d5639d40e854148c48def4997/opm/simulators/linalg/bda/WellContributions.cpp#L90-L132) and hence produces compile errors:
```
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.cpp:116:136: error: no ‘void Opm::WellContributions::getParams(unsigned int*, unsigned int*, unsigned int*, unsigned int*)’ member function declared in class ‘Opm::WellContributions’
 void WellContributions::getParams(unsigned int *num_blocks_, unsigned int *num_std_wells_, unsigned int *dim_, unsigned int *dim_wells_){
```

ducbueno:
Missed some of the `#elif`s. I think it is OK now.

opm/simulators/linalg/bda/WellContributions.cpp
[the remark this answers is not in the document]
blattms: Thanks, I overlooked that we create one matrix from the multiple ones.

opm/simulators/linalg/bda/WellContributions.hpp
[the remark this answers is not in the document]
blattms: Yes, please.

blattms:
What tests have you run on what systems?

I still have problems running this. With my NVIDIA GPU I get:
```
Found 2 OpenCL platforms

Chosen:
Platform name      : NVIDIA CUDA
Platform version   : OpenCL 1.2 CUDA 10.1.353

Found 1 OpenCL devices

Chosen:
CL_DEVICE_NAME            : GeForce GTX 1060 6GB
CL_DEVICE_VERSION         : OpenCL 1.2 CUDA

ESC[31;1m
Error: [/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/openclSolverBackend.cpp:557] OpenCL Error: clCreateBuffer(-61)
CL_INVALID_BUFFER_SIZEESC[0m
Program threw an exception: [/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/openclSolverBackend.cpp:557] OpenCL Error: clCreateBuffer(-61)
CL_INVALID_BUFFER_SIZE
```

opm/simulators/linalg/bda/WellContributions.hpp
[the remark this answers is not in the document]
ducbueno: I've returned it to the original in the latest commit.

ducbueno:
@blattms I've ran all the SPE cases in an Intel integrated GPU (from an i7 7th generation). Due to the pandemic in Brazil I'm away from the university and, therefore, I still haven't had the opportunity to test on other platforms. However this situation may change in a near future (few weeks), than I'll have access to a machine with a NVidia GPU. 

I really can't see why this `CL_INVALID_BUFFER_SIZE` error is only showing with the NVidia GPU. I tried to maintain the buffer and kernel setup patterns for the functionality I added as similar as possible to the others that were already present.

I've ran the simulation with an AMD card haven't you? I think I saw a comment here somewhere, but I'm not being able to find it...

Anyway I'll review the code.

opm/simulators/linalg/ISTLSolverEbos.hpp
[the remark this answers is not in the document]
blattms: thanks for changing back. Indeed the name of the variable is misleading.

blattms:
Please abstain form merging master into your branches (when it creates merge commits) as it messes up history and will make it nearly impossible to revert merges or rebase branches later. If `git merge --ff-only` (the option is to prevent a merge commit and makes it fail if that is not possible) fails then I would suggest rebasing instead:

If the remote pointing to https://github.com/OPM/opm-simulators (check e.g. with `git remote -v` then
```
git fetch upstream
git rebase upstream/master
```

You might want to read up on rebasing when doing this and getting conflicts.

blattms:
I am sorry  (also for causing this), but your confict resolution was not sound:
```
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.cpp: In destructor ‘Opm::WellContributions::~WellContributions()’:
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.cpp:77:5: error: ‘freeStandardWells’ was not declared in this scope
     freeStandardWells();
     ^~~~~~~~~~~~~~~~~
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.cpp:77:5: note: suggested alternative: ‘allocStandardWells’
     freeStandardWells();
     ^~~~~~~~~~~~~~~~~
```

I changed the method name upstream to `freeCudaMemory`

ducbueno:
Sorry for the merge commit. I'm fairly new to Git and big projects. Hope it didn't mess things up outside of this branch.

blattms:
No problem, we all learn every day.

The freeStandardWells call should just removed as we have freeCudaMemory at the beginning of the destructor.

blattms:
@ducbueno Can you run opm-test/model1/MSW_MODEL_1.DATA with your branch? I keep getting:
```
Error: [/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/cuda_header.hpp:42] invalid argument
BDA error message: "WellContributions::addMatrix() failed"
``

**Comment by blattms** [comment]
Works for you as it is interference with CUDA.

We have to have both CUDA and OpenCL functions available, but we can only call the functions regarding the platform we are using at runtime.

File: opm/simulators/linalg/bda/WellContributions.cpp [hunk 5]
```diff
@@ -82,22 +113,70 @@ void WellContributions::apply(cl::Buffer& d_x, cl::Buffer& d_y) {
...
+void WellContributions::addMatrix(MatrixType type, int *colIndices, double *values, unsigned int val_size)
 {
+    if (!allocated) {
+        OPM_THROW(std::logic_error, "Error cannot add wellcontribution before allocating memory in WellContributions");
+    }
 #if HAVE_CUDA
-        addMatrixGpu(type, colIndices, values, val_size);
-#else
-        OPM_THROW(std::logic_error, "Error cannot add StandardWell matrix on GPU because CUDA was not found by cmake");
+    addMatrixGpu(type, colIndices, values, val_size);
```
blattms, line 143: We need to have a runtime switch here in case both CUDA and OpenCL is suported:
```
#if HAVE_CUDA
  if(running_on_cuda_device) // might be false if HAVE_OPENCL is true
  {
    addMatrixGPU ....
}

File: opm/simulators/linalg/bda/WellContributions.cpp [hunk 6]
```diff
@@ -82,22 +113,70 @@ void WellContributions::apply(cl::Buffer& d_x, cl::Buffer& d_y) {
...
+#endif
+
+#if HAVE_OPENCL
+    switch (type) {
```
blattms, line 147: We need to have a runtime switch here in case both CUDA and OpenCL is suported:
```
#if HAVE_OPENCL
  if(running_on_opencl_device) // might be false if HAVE_CUDA is true
  {
     switch ....
```

blattms:
BTW: there might be more places than the two I just pointed out.

ducbueno:
@blattms `opm-tests/model1/MSW_MODEL_1.DATA` runs fine on my system.

blattms:
Yes, it only happens if both are activated. Looking forward to your patch.

opm/simulators/linalg/bda/WellContributions.hpp
```diff
@@ -138,7 +141,7 @@ class WellContributions
 #endif
 
     /// Create a new WellContributions, implementation is empty
```
Tongdongq, line 143: This comment is outdated now.

ducbueno: Fixed.

opm/simulators/linalg/bda/WellContributions.cpp
```diff
@@ -56,12 +81,30 @@ WellContributions::~WellContributions()
         delete ms;
     }
     multisegments.clear();
-}
 
+#if HAVE_CUDA
+    if(cuda_gpu){
+        freeCudaMemory();
```
blattms, line 87: Now you have reintroduced the bug that I recently fixed in #2720 . Was there a special reason to change the order? This needs to move back to line 46!

opm/simulators/linalg/bda/WellContributions.cpp
```diff
@@ -82,22 +125,74 @@ void WellContributions::apply(cl::Buffer& d_x, cl::Buffer& d_y) {
...
+    std::copy(d_Bnnzs_ocl, d_Bnnzs_ocl + num_blocks*dim*dim_wells, *valsB);
+    std::copy(d_Ccols_ocl, d_Ccols_ocl + num_blocks, *colsC);
+    std::copy(d_Bcols_ocl, d_Bcols_ocl + num_blocks, *colsB);
+    std::copy(val_pointers, val_pointers + num_std_wells + 1, *val_pointers_);
 }
+
 #endif
 
-void WellContributions::addMatrix([[maybe_unused]] MatrixType type, [[maybe_unused]]int *colIndices, [[maybe_unused]] double *values, [[maybe_unused]] unsigned int val_size)
+void WellContributions::addMatrix(MatrixType type, int *colIndices, double *values, unsigned int val_size)
```
blattms, line 149: Why did you remove the `[[maybe_unused]]` specifiers? They still are unused if we do not support the GPU, aren't they?

opm/simulators/linalg/bda/WellContributions.cpp
ducbueno: So the `freeCudaMemory()` function should come before the delete statements which follow `if (h_x) {` ? Is that it?

opm/simulators/linalg/bda/WellContributions.cpp
ducbueno: Again, no specific reason. I think I removed them to run some tests and forgot to put them back. Already fixing it.

opm/simulators/linalg/bda/WellContributions.cpp
blattms: Qutoing from the PR

>With Multisegment wells we allocate WellContributions::hx and hy with CudaMallocHost. Yet we tried to deallocate them with delete[]. This caused segementation faults e.g. for model1/MSW_MODEL_1. Now we use CudaFreeHost for freeing if we used CUDA.

To have `if (h_x)` be false the pointer needs to be freed with `freeCudaMemory` before. Would be nice if you would put  a comment there (which I should probably habve done).

ducbueno: Got it. Fixed it.

blattms:
I am still having troubles to run opm-tests/run model1/MSW_MODEL_1 (aborts with signal 6) and it seems like there are memory issues. Please see the attached [output](https://github.com/OPM/opm-simulators/files/5029072/stdwells_valgrind.log)  from a run with valgrind for more insight

opm/simulators/linalg/bda/openclSolverBackend.cpp
```diff
@@ -212,6 +219,17 @@ void openclSolverBackend<block_size>::gpu_pbicgstab(WellContributions& wellContr
...
 
+    wellContribs.setReordering(toOrder, true);
+    wellContribs.getData(&h_valsC, &h_valsD, &h_valsB, &h_colsC, &h_colsB, &h_val_pointers);
+
+    queue->enqueueWriteBuffer(d_valsC, CL_TRUE, 0, sizeof(double) * num_blocks * dim_ * dim_wells, h_valsC);
+    queue->enqueueWriteBuffer(d_valsD, CL_TRUE, 0, sizeof(double) * num_std_wells * dim_wells * dim_wells, h_valsD);
+    queue->enqueueWriteBuffer(d_valsB, CL_TRUE, 0, sizeof(double) * num_blocks * dim_ * dim_wells, h_valsB);
+    queue->enqueueWriteBuffer(d_colsC, CL_TRUE, 0, sizeof(int) * num_blocks, h_colsC);
+    queue->enqueueWriteBuffer(d_colsB, CL_TRUE, 0, sizeof(int) * num_blocks, h_colsB);
+    queue->enqueueWriteBuffer(d_val_pointers, CL_TRUE, 0, sizeof(unsigned int) * (num_std_wells + 1), h_val_pointers, nullptr, &event);
```
blattms, line 230: Please explain what we are doing here. It seems `getData` is actually performing a superfluous host-to-host copy. Coudn't we just pass the pointer copied from there (d_Cnnzs_ocl, ..., d_Bnnzs_ocl,  d_Ccols_ocl, ...) directly to `queue->enqueueWriteBuffer`?

blattms:
AT least the first valgrind error was there previously and I have a patch coming for that.

opm/simulators/linalg/bda/openclSolverBackend.cpp
ducbueno: Yes, it is a host-to-host copy. I've implemented it like this because the `d_*_ocl` arrays are assembled by parts (through concatenations) in the `addMatrix` function. I couldn't see another way of passing data from the `WellContributions` object to the `openclSolverBackend` object. But I'm completely open to suggestions.

ducbueno:
I think the errors that are happening with MSW_MODEL_1 are due to it using multi-segmented wells. I'm trying to find the bug.

blattms:
I see two possibilities:

1. Hold the data/pointers in OpenCLBackendSolver and pass them to WellContributions::addMatrix for population
2. Add an access method to WellContributions to get the pointers, e.g.
   ```c++
#include<tuple>
class WellContributions{
...
public:
  std::tuple<double*, double*,....> getMatrixData(){
     return std::make_tuple(d_Cnnzs_ocl, ...)
};
...
std::tie(d_valsC, ....) = wellcontribs,getMatrixData();
   ```
   If the pointers might be reallocated during the simulation (which I assume), you need to get them everytime.

opm/simulators/linalg/bda/openclSolverBackend.cpp
Tongdongq: The Opm::Wells are separate objects on the CPU. We want some of their data in a contiguous array on GPU. I only see two options:
- multiple concatenating CPU-to-GPU transfers
- multiple concatenating CPU-to-CPU transfers, preferably to pinned memory. One CPU-to-GPU transfer

ducbueno: I think the last option is what I'm currently doing

ducbueno:
Since this last commit (8ae9b81) the simulation of MSW_MODEL_1 reaches the end and the final report is printed (whats was not happening before), however there is still a core dump. Working on that already. Also checking the well curves that come out of the OpenCL simulations.

Tongdongq:
A core dump after the final summary is printed is probably caused by a
destructor.

ducbueno:
Figured out a way of keeping the data from the `WellContributions` object self-contained. The core dump at the end of the simulation is gone. However, there is still some weird behavior when applying the standard well contributions, since the timesteps get chopped a lot. I think the `add_well_contributions` kernel is working fine. I'll post the link to a repo where I made some tests later.

ducbueno:
I had forgotten to uncomment the commands corresponding to the addition of multi segmented well contributions. By doing it the timestep chopping on simulating the `MSW_MODEL_1.DATA` model got significantly reduced (in some simulations it didn't even happen).

blattms:
Sounds great.

blattms:
Would you mind reporting some numbers for different cases, e.g. the SPE cases with CPU and Opencl (runtime and iterations can be found at the end of the output or *.PRT *.DBG files)?

blattms:
Closing in favor of 2751

Issue #2654, opened 2020-06-04
[CUDA/OpenCL] compile errors for multisegment wells with dune >=2.7


blattms:
With current DUNE master of dune-istl, I get:
```
/home/mblatt/src/dune/opm/opm-simulator/opm/simulators/wells/MultisegmentWell_impl.hpp:677:14: error: cannot convert ‘Dune::ColCompMatrix<Dune::BCRSMatrix<Dune::FieldMatrix<double, 4, 4>, std::allocator<Dune::FieldMatrix<double, 4, 4> > >, long int>::Index* {aka long int*}’ to ‘int*’ in initialization
         int *Dcols = umfpackMatrix.getInternalMatrix().getColStart();
              ^~~~~
/home/mblatt/src/dune/opm/opm-simulator/opm/simulators/wells/MultisegmentWell_impl.hpp:678:14: error: cannot convert ‘Dune::ColCompMatrix<Dune::BCRSMatrix<Dune::FieldMatrix<double, 4, 4>, std::allocator<Dune::FieldMatrix<double, 4, 4> > >, long int>::Index* {aka long int*}’ to ‘int*’ in initialization
         int *Drows = umfpackMatrix.getInternalMatrix().getRowIndex();
              ^~~~~
```
 
Seems like the switch happened last year:
```
commit 1b8df0797afc83f8c758b95090fd862668f39ba4
Author: Dominic Kempf 
Date:   Tue Aug 27 13:44:01 2019 +0200

    [bugfix] Switch UMFPack complex number branch to long size type
    
    THe index type used for UMFPack was recently switched to long int
    by @kilian.weishaupt. However, the specialization of the method
    dispatcher for std::complex was not adapted.

commit 9467fce9114d5afa6246a320bfe249df7dffdf38
Author: Kilian Weishaupt 
Date:   Sat May 4 17:17:07 2019 +0200

    Use long int UMFPack functions to compute larger systems
    
    With the *_dl_* versions instead of the *_di_* versions UMFPACK will
    not have a memory limit of just 2GB.
```

These changes are also included in the 2.7 release and hence I assume that it has the same problem.

blattms:
@Tongdongq could you please take a look?

Tongdongq:
So in Dune 2.6, `UMFPack::UMFPackMatrix` was a `ColCompMatrix`, which had a hardcoded `int` as index type.
[...]

PR #2718, opened 2020-07-31
Reuse umfpack decomposition for multisegment wells.

blattms:
Currently, we decompose the diagonal matrix every time we apply the well equations. With this change we will store the decomoposition and reuse whenever possible. This will ease the fixing of #2654 and save an additional decomposition there.  But already now this speeds up computation by ~3%. 

Iteration counts stay exactly the same with this change.

### DIFF
--- a/opm/simulators/wells/MSWellHelpers.hpp
+++ b/opm/simulators/wells/MSWellHelpers.hpp
@@ -42,8 +42,13 @@ namespace mswellhelpers
     /// Applies umfpack and checks for singularity
     template <typename MatrixType, typename VectorType>
     VectorType
-    applyUMFPack(Dune::UMFPack<MatrixType>& linsolver, VectorType x)
+    applyUMFPack(const MatrixType& D, std::shared_ptr<Dune::UMFPack<MatrixType> >& linsolver, VectorType x)
     {
+        if (!linsolver)
+        {
+            linsolver.reset(new Dune::UMFPack<MatrixType>(D, 0));
+        }
+
         // The copy of x seems mandatory for calling UMFPack!
         VectorType y(x.size());
         y = 0.;
@@ -52,7 +57,7 @@ namespace mswellhelpers
         Dune::InverseOperatorResult res;
 
         // Solve
-        linsolver.apply(y, x, res);
+        linsolver->apply(y, x, res);
 
         // Checking if there is any inf or nan in y
         // it will be the solution before we find a way to catch the singularity of the matrix
@@ -73,8 +78,8 @@ namespace mswellhelpers
     invDXDirect(const MatrixType& D, VectorType x)
     {
 #if HAVE_UMFPACK
-        Dune::UMFPack<MatrixType> linsolver(D, 0);
-        return applyUMFPack(linsolver, x);
+        std::shared_ptr<Dune::UMFPack<MatrixType> > linsolver;
+        return applyUMFPack(D, linsolver, x);
 #else
         // this is not thread safe
         OPM_THROW(std::runtime_error, "Cannot use invDXDirect() without UMFPACK. "
--- a/opm/simulators/wells/MSWellHelpers.hpp
+++ b/opm/simulators/wells/MSWellHelpers.hpp
@@ -64,7 +64,7 @@ namespace mswellhelpers
         for (size_t i_block = 0; i_block < y.size(); ++i_block) {
             for (size_t i_elem = 0; i_elem < y[i_block].size(); ++i_elem) {
                 if (std::isinf(y[i_block][i_elem]) || std::isnan(y[i_block][i_elem]) ) {
-                    OPM_THROW(Opm::NumericalIssue, "nan or inf value found in invDXDirect due to singular matrix");
+                    OPM_THROW(Opm::NumericalIssue, "nan or inf value found after UMFPack solve due to singular matrix");
                 }
             }
         }
--- a/opm/simulators/wells/MSWellHelpers.hpp
+++ b/opm/simulators/wells/MSWellHelpers.hpp
@@ -1,6 +1,7 @@
 /*
   Copyright 2017 SINTEF Digital, Mathematics and Cybernetics.
   Copyright 2017 Statoil ASA.
+  Copyright 2020 Equinor ASA.
 
   This file is part of the Open Porous Media project (OPM).
 
@@ -36,16 +37,16 @@ namespace Opm {
 
 namespace mswellhelpers
 {
-    // obtain y = D^-1 * x with a direct solver
+
+#if HAVE_UMFPACK
+    /// Applies umfpack and checks for singularity
     template <typename MatrixType, typename VectorType>
     VectorType
-    invDXDirect(const MatrixType& D, VectorType x)
+    applyUMFPack(Dune::UMFPack<MatrixType>& linsolver, VectorType x)
     {
-#if HAVE_UMFPACK
+        // The copy of x seems mandatory for calling UMFPack!
         VectorType y(x.size());
         y = 0.;
-
-        Dune::UMFPack<MatrixType> linsolver(D, 0);
 
         // Object storing some statistics about the solving process
         Dune::InverseOperatorResult res;
@@ -62,8 +63,18 @@ namespace mswellhelpers
                 }
             }
         }
-
         return y;
+    }
+#endif
+
+    // obtain y = D^-1 * x with a direct solver
+    template <typename MatrixType, typename VectorType>
+    VectorType
+    invDXDirect(const MatrixType& D, VectorType x)
+    {
+#if HAVE_UMFPACK
+        Dune::UMFPack<MatrixType> linsolver(D, 0);
+        return applyUMFPack(linsolver, x);
 #else
         // this is not thread safe
         OPM_THROW(std::runtime_error, "Cannot use invDXDirect() without UMFPACK. "
--- a/opm/simulators/wells/MSWellHelpers.hpp
+++ b/opm/simulators/wells/MSWellHelpers.hpp
@@ -71,7 +71,7 @@ namespace mswellhelpers
         return y;
 #else
         // this is not thread safe
-        OPM_THROW(std::runtime_error, "Cannot use aplyUMFPack() without UMFPACK. "
+        OPM_THROW(std::runtime_error, "Cannot use applyUMFPack() without UMFPACK. "
                   "Reconfigure opm-simulator with SuiteSparse/UMFPACK support and recompile.");
 #endif // HAVE_UMFPACK
     }
--- a/opm/simulators/wells/MSWellHelpers.hpp
+++ b/opm/simulators/wells/MSWellHelpers.hpp
@@ -38,12 +38,12 @@ namespace Opm {
 namespace mswellhelpers
 {
 
-#if HAVE_UMFPACK
     /// Applies umfpack and checks for singularity
     template <typename MatrixType, typename VectorType>
     VectorType
     applyUMFPack(const MatrixType& D, std::shared_ptr<Dune::UMFPack<MatrixType> >& linsolver, VectorType x)
     {
+#if HAVE_UMFPACK
         if (!linsolver)
         {
             linsolver.reset(new Dune::UMFPack<MatrixType>(D, 0));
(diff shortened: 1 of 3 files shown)


blattms:
jenkins build this please

blattms:
benchmark please

blattms:
@GitPaean would you mind sharing your 2cents about this?

GitPaean:
benchmark please

GitPaean:
It looks like the `benchmark` might not be working. I will do some test on my own machine. Thanks for the patience.

alfbr:
It should be working now. Let's try again.

alfbr:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 0.965    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.007    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.02     |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.008    |



View result details @ https://www.ytelses.com/opm/?page=result&id=851

opm/simulators/wells/MultisegmentWell.hpp
```diff
@@ -234,8 +234,12 @@ namespace Opm
         // two off-diagonal matrices
         mutable OffDiagMatWell duneB_;
         mutable OffDiagMatWell duneC_;
-        // diagonal matrix for the well
+        // "diagonal" matrix for the well. It has offdiagonal entries for inlets and outlets.
```
GitPaean, line 237: Just a comment, not against the change in comment here.

Diagonal here basically means the location of this well matrix in the global matrix, which contains the derivatives of the equations of a well with respect to its own primary variables. 

There are another two matrices associated with this well, which describe the relationship between the well and reservoir. The are on the off diagonal position in the global matrix.

opm/simulators/wells/MSWellHelpers.hpp
```diff
@@ -36,22 +37,27 @@ namespace Opm {
 
 namespace mswellhelpers
 {
-    // obtain y = D^-1 * x with a direct solver
+
+#if HAVE_UMFPACK
+    /// Applies umfpack and checks for singularity
     template <typename MatrixType, typename VectorType>
     VectorType
-    invDXDirect(const MatrixType& D, VectorType x)
```
GitPaean, line 42: I could not comment on the line for some reason, basically the `invDxDirect` in the error message (line 61) should be changed following the change of the function name here?

opm/simulators/wells/MSWellHelpers.hpp
```diff
@@ -62,8 +68,18 @@ namespace mswellhelpers
...
         }
-
         return y;
+    }
+#endif
+
+    // obtain y = D^-1 * x with a direct solver
+    template <typename MatrixType, typename VectorType>
+    VectorType
+    invDXDirect(const MatrixType& D, VectorType x)
```
GitPaean, line 78: I do not think this function is used anymore. It is good to remove it completely to avoid confusion.  

And also, since this function is not used anymore, the checking of the `HAVE_UMFPACK` is also lost.  Please also keep it when removing this function.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1013,7 +1016,8 @@ namespace Opm
     {
         // We assemble the well equations, then we check the convergence,
         // which is why we do not put the assembleWellEq here.
-        const BVectorWell dx_well = mswellhelpers::invDXDirect(duneD_, resWell_);
+        //const BVectorWell dx_well = mswellhelpers::invDXDirect(duneD_, resWell_);
```
GitPaean, line 1019: These commented lines can be removed for clarity of the code.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -3205,6 +3214,7 @@ namespace Opm
             duneD_[seg][outlet_segment_index][SPres][pv_idx] = -outlet_pressure.derivative(pv_idx + numEq);
 
         }
+        duneDSolver_.reset();
```
GitPaean, line 3217: I do not think we need to keep doing this in multiple location. Maybe only do it at the end of function `assembleWellEqWithoutIteration`? It is as far as what I can see now. I can miss something.

opm/simulators/wells/MSWellHelpers.hpp
```diff
@@ -62,8 +68,18 @@ namespace mswellhelpers
                 }
```
GitPaean, line 68: `invDXDirect` in above throwing message be following the function name to be sensible.

opm/simulators/wells/MultisegmentWell_impl.hpp
GitPaean: Or can be easier to understand from code perspective, do the reset after the matrix D is cleared. 
```
2512 
2513         duneD_ = 0.0;
```

opm/simulators/wells/MSWellHelpers.hpp
blattms: Sorry, I do not understand this comment. Can your rephrase it, please.

opm/simulators/wells/MultisegmentWell.hpp
blattms: Understood

opm/simulators/wells/MultisegmentWell_impl.hpp
blattms: Will try.

blattms:
jenkins build this please

opm/simulators/wells/MSWellHelpers.hpp
blattms: Thanks. resolved.

blattms:
jenkins build this please

opm/simulators/wells/MSWellHelpers.hpp
```diff
@@ -36,45 +37,47 @@ namespace Opm {
...
+                    OPM_THROW(Opm::NumericalIssue, "nan or inf value found after UMFPack solve due to singular matrix");
                 }
             }
         }
-
         return y;
 #else
         // this is not thread safe
-        OPM_THROW(std::runtime_error, "Cannot use invDXDirect() without UMFPACK. "
+        OPM_THROW(std::runtime_error, "Cannot use aplyUMFPack() without UMFPACK. "
```
GitPaean, line 74: typo `aplyUMFpack()`.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -640,7 +640,8 @@ namespace Opm
     apply(BVector& r) const
     {
         // invDrw_ = duneD^-1 * resWell_
-        const BVectorWell invDrw = mswellhelpers::invDXDirect(duneD_, resWell_);
+        // const BVectorWell invDrw = mswellhelpers::invDXDirect(duneD_, resWell_);
```
GitPaean, line 643: please remove this commented line.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -999,7 +1000,8 @@ namespace Opm
         // resWell = resWell - B * x
         duneB_.mmv(x, resWell);
         // xw = D^-1 * resWell
-        xw = mswellhelpers::invDXDirect(duneD_, resWell);
+        //xw = mswellhelpers::invDXDirect(duneD_, resWell);
```
GitPaean, line 1003: please remove this commented line.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2399,7 +2398,8 @@ namespace Opm
 
             assembleWellEqWithoutIteration(ebosSimulator, dt, inj_controls, prod_controls, well_state, deferred_logger);
 
-            const BVectorWell dx_well = mswellhelpers::invDXDirect(duneD_, resWell_);
+            // const BVectorWell dx_well = mswellhelpers::invDXDirect(duneD_, resWell_);
```
GitPaean, line 2401: please remove this commented line.

GitPaean (approved):
It looks good. There are a few very minor comments.

opm/simulators/wells/MSWellHelpers.hpp
```diff
@@ -36,45 +37,47 @@ namespace Opm {
 
 namespace mswellhelpers
 {
-    // obtain y = D^-1 * x with a direct solver
+
+    /// Applies umfpack and checks for singularity
     template <typename MatrixType, typename VectorType>
     VectorType
-    invDXDirect(const MatrixType& D, VectorType x)
+    applyUMFPack(const MatrixType& D, std::shared_ptr<Dune::UMFPack<MatrixType> >& linsolver, VectorType x)
```
bska, line 44: Tangential to the topic of this PR I know, but what is the (typical) `VectorType` here?  Is there a reason we take the `x` by value?

blattms: I think I tried changing that, but it seemed like UMFPack (in suitesparse) actually changed `x` when solving.

blattms:
Ok, all comments adressed. I will leave the decision whether to run jenkins or not to the merger. But IMHO cca38dd  and 49ed3d6 should be innocent enough to skip that (I know: famous last words...)

opm/simulators/wells/MSWellHelpers.hpp
GitPaean: Yes. It was tested before this way.

GitPaean:
jenkins build this please

opm/simulators/wells/MSWellHelpers.hpp
bska: > it seemed like UMFPack (in suitesparse) actually changed `x` when solving.

I won't dwell on this and I'm certainly not going to suggest we change it.  I will however remark that if `UMFPack` does change `x` then that's a violation of its declared contract&mdash;at least if `x` is the system right-hand side vector.

The `umfpack_*_solve()` functions are declared as
```C
/* From umfpack_solve.h, using "di" as the prototypical type example */
int umfpack_di_solve
(
    int sys,
    const int Ap [ ],
    const int Ai [ ],
    const double Ax [ ],
    double X [ ],
    const double B [ ],
    void *Numeric,
    const double Control [UMFPACK_CONTROL],
    double Info [UMFPACK_INFO]
);
```
with `B` denoting the right-hand side vector, `Ap`, `Ai`, and `Ax` being the coefficient matrix (CSC format) and `X` being the array in which to output solution to the linear system.

GitPaean:
It is all good. Merging now.

blattms:
thanks!

opm/simulators/wells/MSWellHelpers.hpp
GitPaean: Maybe it is how DUNE wrap it. 

```c++
334       double UMF_Apply_Info[UMFPACK_INFO];
335       Caller::solve(UMFPACK_A,
336                     umfpackMatrix_.getColStart(),
337                     umfpackMatrix_.getRowIndex(),
338                     umfpackMatrix_.getValues(),
339                     reinterpret_cast<double*>(&x[0]),
340                     reinterpret_cast<double*>(&b[0]),
341                     UMF_Numeric,
342                     UMF_Control,
343                     UMF_Apply_Info);
```

Maybe `reinterpret_cast<double*>(&b[0]),` should be `reinterpret_cast<const double*>(&b[0]),`?

bska: > Maybe it is how DUNE wrap it.

That may indeed be part of the reason here, yes.

> Maybe `reinterpret_cast<double*>(&b[0]),` should be `reinterpret_cast<const double*>(&b[0]),`?

I think that would be necessary, but maybe not quite sufficient.  I'd have to look into the details of the wrapper code to be sure.

GitPaean:
One behavoir changed by this PR is that, before without UMFPack, we can build `flow`, but we can not run with multisegment wells cases. 

With this PR, we will not be able to compile `flow` anymore without UMFPack, because the usage of `UMFPack` is unconditional now. 
```
        mutable std::shared_ptr<Dune::UMFPack<DiagMatWell> > duneDSolver_;
```

blattms:
IMHO it has been unconditional since dropping SuperLU support, as AMG needs a
direct solver. It was just not reflected by the buildsystem.

Even if not. If some wells need UMFPACK it should not be conditional as this just produces quite unpleasant surprises at the user site.

GitPaean:
Yes. It was just a note for information. 

What happened to me is that we are building OPM on some HPC clusters and we need to several dependent software by ourselves including SuiteSparse. It took some efforts and it worked at the end. The case we are testing does not include Multisegment well. 

Then today I upgraded OPM to be the latest master, suddenly it did not compile because there is no `Dune::UMFPack` class.  Then I think my configuration related to self-built SuiteSparse is not working. We have been using `Flow` without UMFPack, which was a surprise to me.

PR #2722, opened 2020-08-05
[OpenCL] Fixes out of bounds read in reordering code in ILU0.

blattms:
rowPerColor only has size numColors!

### DIFF
--- a/opm/simulators/linalg/bda/BILU0.cpp
+++ b/opm/simulators/linalg/bda/BILU0.cpp
@@ -119,10 +119,9 @@ namespace bda
         queue->enqueueWriteBuffer(s.invDiagVals, CL_TRUE, 0, mat->Nb * sizeof(double) * bs * bs, invDiagVals);
 
         int *rowsPerColorPrefix = new int[numColors + 1];
-        int prefix = 0;
-        for (int i = 0; i < numColors + 1; ++i) {
-            rowsPerColorPrefix[i] = prefix;
-            prefix += rowsPerColor[i];
+        rowsPerColorPrefix[0] = 0;
+        for (int i = 0; i < numColors; ++i) {
+            rowsPerColorPrefix[i+1] =rowsPerColorPrefix[i] + rowsPerColor[i];
         }
         queue->enqueueWriteBuffer(s.rowsPerColor, CL_TRUE, 0, (numColors + 1) * sizeof(int), rowsPerColorPrefix);
         delete[] rowsPerColorPrefix;
--- a/opm/simulators/linalg/bda/BILU0.cpp
+++ b/opm/simulators/linalg/bda/BILU0.cpp
@@ -121,7 +121,7 @@ namespace bda
         int *rowsPerColorPrefix = new int[numColors + 1];
         rowsPerColorPrefix[0] = 0;
         for (int i = 0; i < numColors; ++i) {
-            rowsPerColorPrefix[i+1] =rowsPerColorPrefix[i] + rowsPerColor[i];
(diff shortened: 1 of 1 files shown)


blattms:
Would be nice if @Tongdongq could double check this.

blattms:
jenkins build this please

Tongdongq:
I like the reordered loop, but there is a `nodesPerColorPrefix[color+1]` read in both `ILU_apply` OpenCL kernels.
All colors are applied sequentially, so I assume that last entry is used as well. The value should be `numColors`.
Also, there is a space missing behind the `=` sign.

blattms:
Not sure what you mean.

Shouldn't the last iteration in the loop set `rowsPerColorPrefix[numColors] = rowsPerColorPrefix[0] + rowsPerColorPrefix[1] + ... + rowsPerColorPrefix[numColors-1]` and `color +1<= numColors` hold?

Tongdongq:
Oops, the `color` should indeed never be equal to `numColors`.
I think the righthand side of your first equation should use `rowsPerColor` instead of `rowsPerColorPrefix`.
The last value shouldn't be equal to `numColors`, it should be `numRows` or something like that.

blattms:
The last values is not `numcolors` it should be the same value as before (`numRows=rowsPerColorPrefix[0] + rowsPerColorPrefix[1] + ... + rowsPerColorPrefix[numColors-1]` should hold). Unless I made a serious mistake.

Tongdongq:
But it's a prefix. So `rowsPerColorPrefix[i] = rowsPerColor[0] + ... + rowsPerColor[i-1]`.
`rowsPerColor[i]` holds the number of rows a particular color contains. So the sum of these should equal the total number of rows.
Both `rowsPerColorPrefix[3]` and `rowsPerColorPrefix[4]` contain `rowsPerColor[1]`, adding them makes no sense.

blattms:
From my point of view the old and new code should behave then same. 

Let us do an example:

numColors = 3, numRows = 8
rowsPerColor = { 1, 3, 4 };
 
Original code:
- prefix = 0
- i = 0:
  + rowsPerColorPrefix[0] = 0
  + prefix = 0 + 1 = 1
- i = 1:
  + rowsPerColorPrefix[1] = 1
  + prefix = 1 + 3 = 4
- i = 2:
  + rowsPerColorPrefix[2] = 4
  + prefix = 4 + 4 = 8
- i = 3:
  + rowsPerColorPrefix[3] = 8
  + prefix = ?? // out of bound read
- result: rowsPerColorPrefix = {0, 1, 4, 8}  

New code:
- rowsPerColorPrefix[0] = 0
- i = 0:
  + rowsPerColorPrefix[1] = 0 + 1 = 1
- i = 1:
  + rowsPerColorPrefix[2] = 1 + 3 = 4
- i = 2:
  + rowsPerColorPrefix[3] = 4 + 4 = 8
- result: rowsPerColorPrefix = {0, 1, 4, 8}

I feel a bit dense. Where is my mistake?

blattms:
Sorry my comment, that we are `numRows=rowsPerColorPrefix[0] + rowsPerColorPrefix[1] + ... + rowsPerColorPrefix[numColors-1]` holds is wrong. It is `numRows=rowsPerColorPrefix[numColors]=rowsPerColor[0]+...+rowsPerColor[numColors-1]`. Sorry for the confusion.

Tongdongq:
No mistake in the code, only in one of the PR comments. You should not
added multiple `prefix` values together. Only add a normal `rowsPerColor`
to the previous prefix.

blattms:
jenkins build this please

PR #2709, opened 2020-07-20
Report Guiderates at Well and Group Levels

bska:
This PR adds support for reporting the simulator's guiderate values at the well and group levels to the output layer through the `wellData()` and `groupData()` member functions.  We add several new member functions that collectively assemble the values and assign them to objects of type `Opm::data::GuideRateValue` (PR OPM/opm-common#1868) for subsequent output to the summary and restart files.

In particular
```
getGuideRateValues(const Well&) const
getGuideRateValues(const Group&) const
```
retrieve the guiderate values for all phases for those individual wells and groups for which the `guideRate_` data member defines a value.  The most complicated function of this PR is
```
calculateAllGroupGuideRates
```
which aggregates those individual contributions from the well (leaf) level up to the root of the group tree (the FIELD group).  This process uses an ancillary array (`up`) to keep track of the parent groups of all wells and all groups, and to ensure that we only visit each parent group once (`sort`+`unique` on subsets of the `up` array).

We do not currently support outputting guiderates for reservoir voidage volume (`GuideRateModel::Target::RES`).

We also add a few helper functions to the `WellGroupHelpers` and `WellStateFullyImplicitBlackoil` in order to simplify the implementation of `calculateAllGroupRates`.

PR #2726, opened 2020-08-17
 Include All Pertinent EFACs In Guiderate Potential Updates

bska:
This PR switches the helper function
```
WellGroupHelpers::updateGuideRateForGroups<>()
```
to include efficiency factors in the potential rates at grouptree levels below a particular group.  We furthermore switch the helper function
```
WellGroupHelpers::updateGuideRatesForWells<>()
```
to not include efficiency factors at all.

The motivation for this change is that efficiency factors always apply to the level we're accumulating rate values into rather than to the rate values themselves.

### DIFF
--- a/opm/simulators/wells/WellGroupHelpers.hpp
+++ b/opm/simulators/wells/WellGroupHelpers.hpp
@@ -110,6 +110,8 @@ namespace WellGroupHelpers
             updateGuideRateForGroups(
                 groupTmp, schedule, pu, reportStepIdx, simTime, isInjector, wellState, comm, guideRate, thisPot);
 
+            const auto gefac = groupTmp.getGroupEfficiencyFactor();
+
             // accumulate group contribution from sub group unconditionally
             if (isInjector) {
                 const Phase all[] = {Phase::WATER, Phase::OIL, Phase::GAS};
@@ -124,7 +126,7 @@ namespace WellGroupHelpers
                     else
                         continue;
 
-                    pot[phasePos] += thisPot[phasePos];
+                    pot[phasePos] += gefac * thisPot[phasePos];
                 }
             } else {
                 const Group::ProductionCMode& currentGroupControl = wellState.currentProductionGroupControl(groupName);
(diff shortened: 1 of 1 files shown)


bska:
Note: This PR depends on PR #2709 and must not be merged before that PR is already installed in master.

bska:
jenkins build this please

bska:
@atgeirr : Would you mind reviewing this?  It is part of the larger task to get guiderate summary output that more closely matches those of ECLIPSE&mdash;at least for a selection of models.

atgeirr:
I think the code looks good. If it does the *right thing* I am not quite so certain of, but I believe it may be so. The difference appears to me to be this (please tell me if I misunderstood): when computing the guide rates of some object, we will no longer multiply its potentials by the efficiency factor of the object itself. Assuming that is the behaviour of eclipse (as tested by you) I am fine with this. I think that a priori it is not clear what is the "right" way, but we should be pragmatic enough to give the same behaviour so users get the experience they expect.

bska:
> The difference appears to me to be this (please tell me if I misunderstood): when computing the guide rates of some object, we will no longer multiply its potentials by the efficiency factor of the object itself.

That is correct.  In this PR we apply efficiency factors only when accumulating an object's (well's or group's) potential rates upwards in the group tree.

> Assuming that is the behaviour of eclipse (as tested by you) I am fine with this.

ECLIPSE's rules for efficiency factors (in the range `[0, 1)`) are as follows:

1. An object (well or group) runs at its full rate (production, injection or potential) as implied by the reservoir and control states (and contributions from any levels below it)
2. The efficiency factor reduces the **time** available for the object to run; this impacts total volumes produced/injected attributable to the object
3. To the grouptree level **above** the object it appears as if the object runs for the full timestep, but with a reduced **rate**

The net outcome of this is that individual rate values are unaffected by the object's efficiency factor, but that the efficiency factor must be included when accumulating the object's contributions into the next level upwards in the group tree.

bska:
Advance notice: Unless anyone objects I intend to generate new reference solutions at noon CEST tomorrow (Tuesday 15th September) and then merge this and the upstream PR (OPM/opm-common#1880) once those new reference solutions have been computed.

bska:
Updated reference data and upstream PRs have been merged into master.  Merging this now per prior approval to complete the feature update.

PR #2727, opened 2020-08-18
fix output unit salt

totto82:
The salt concentration output unit is density.


totto82:
jenkins build this opm-common=1874 please

bska:
New reference data and upstream changes have been merged.  Merging this now to complete the unit update.  Thanks a lot for fixing this problem.

PR #2733, opened 2020-08-23
Serialize variant

joakim-hove:
Another elegant and general `std::variant<...>` serialization implementation :-)

Downstream of: https://github.com/OPM/opm-common/pull/1886

### DIFF
--- a/ebos/eclmpiserializer.hh
+++ b/ebos/eclmpiserializer.hh
@@ -97,7 +97,7 @@ public:
     }
 
 
-    //! \brief Handler for std::variant<> with three types
+    //! \brief Handler for std::variant<> with four types
 
     /*
       The std::variant<> serialization is a first attempt and *not* particularly
@@ -146,6 +146,52 @@ public:
         }
     }
 
+
+    //! \brief Handler for std::variant<> with two fundamental types
+
+    /*
+      This std::variant serialization is highly specialized:
+
+        1. It is hardcoded to take exactly two types T0 and T1.
+
+        2. Both T0 and T1 must be basic types where Mpi::pack(T, ...) overloads
+           must exist.
+
+    */
+    template<class T0, class T1>
+    void variant(std::variant<T0,T1>& data)
+    {
+        auto pack_size = [&](auto& d) {
+                             m_packSize += Mpi::packSize(d, m_comm);
+                         };
+
+        auto pack = [&](auto& d) {
+                          Mpi::pack(d, m_buffer, m_position, m_comm);
+                      };
+
+        if (m_op == Operation::PACKSIZE) {
+            m_packSize += Mpi::packSize(data.index(), m_comm);
+            std::visit( [&] (auto& arg) { pack_size(arg); }, data);
+        } else if (m_op == Operation::PACK) {
+            Mpi::pack(data.index(), m_buffer, m_position, m_comm);
+            std::visit([&](auto& arg) { pack(arg); }, data);
+        } else if (m_op == Operation::UNPACK) {
+            size_t index;
+            Mpi::unpack(index, m_buffer, m_position, m_comm);
+
+            if (index == 0) {
+                T0 t0;
+                Mpi::unpack(t0, m_buffer, m_position, m_comm);
+                data = t0;
+            } else if (index == 1) {
+                T1 t1;
+                Mpi::unpack(t1, m_buffer, m_position, m_comm);
+                data = t1;
+            } else
+                throw std::logic_error("Internal meltdown in std::variant<T0,T1> unpack loaded index=" + std::to_string(index) + " allowed range: [0,1]");
+        }
+
+    }
 
     //! \brief Handler for maps.
     //! \tparam Map map type


joakim-hove:
Comments to this?

ebos/eclmpiserializer.hh
```diff
@@ -147,6 +147,53 @@ public:
...
+            if (index == 0) {
+                T0 t0;
+                Mpi::unpack(t0, m_buffer, m_position, m_comm);
+                data = t0;
+            } else if (index == 1) {
+                T1 t1;
+                Mpi::unpack(t1, m_buffer, m_position, m_comm);
+                data = t1;
+            } else
+                std::logic_error("Internal meltdown in std::variant<T0,T1> unpack");
```
bska, line 192: `throw`?

Also, I think you should include the `index` value in the diagnostic message.

ebos/eclmpiserializer.hh
```diff
@@ -147,6 +147,53 @@ public:
...
+    {
+        auto pack_size = [&](auto& d) {
+                             m_packSize += Mpi::packSize(d, m_comm);
+                         };
+
+        auto pack = [&](auto& d) {
+                          Mpi::pack(d, m_buffer, m_position, m_comm);
+                      };
+
+        std::variant<T0,T1>& data = const_cast<std::variant<T0,T1>&>(_data);
```
bska, line 172: The `variant` member function accepts temporaries (parameter type is reference to `const`) but this will is undefined behaviour if anyone passes a temporary to the `UNPACK` operation.  That problem is, of course, shared by all the other serialisation functions too,  but I'd really prefer if we didn't add more of it.

bska (commented):
Reading this I just realised that there is a serious API bug in this facility.  The member functions accept their arguments as reference-to-const&mdash;which accepts temporaries&mdash;but will `const_cast` that parameter and write to it in the `UNPACK` case.

I'd really prefer if passing temporaries to `EclMpiSerializer::operator()<>()` were disallowed and generated compiler errors.

ebos/eclmpiserializer.hh
joakim-hove: Fixed

joakim-hove:
> Reading this I just realised that there is a serious API bug in this facility. The member functions accept their arguments as reference-to-const—which accepts temporaries—but will const_cast that parameter and write to it in the UNPACK case.

The general solution to this is above my C++ paygrade - but for the new `std::variant<T0,T1>` serialization it is simple to remove the `const` - I have done that.

ebos/eclmpiserializer.hh
```diff
@@ -147,6 +147,52 @@ public:
...
+            if (index == 0) {
+                T0 t0;
+                Mpi::unpack(t0, m_buffer, m_position, m_comm);
+                data = t0;
+            } else if (index == 1) {
+                T1 t1;
+                Mpi::unpack(t1, m_buffer, m_position, m_comm);
+                data = t1;
+            } else
+                std::logic_error("Internal meltdown in std::variant<T0,T1> unpack loaded index=" + std::to_string(index) + " allowed range: [0,1]");
```
bska, line 191: Still missing a `throw` here.  At present this just constructs and destroys a temporary object of type `std::logic_error`, with no transfer of control anywhere.

bska (commented):
Thanks for the updates.  I'll launch a build check.  Meanwhile you may wish to address my one minor comment.

bska:
jenkins build this please

ebos/eclmpiserializer.hh
joakim-hove: Throw added ...

bska: > Throw added ...

Thanks.  I'll merge this once the build check is green

bska:
Build check succeeds.  Merging as promised.

joakim-hove:
Hmmm - turns out the `const_cast< >` was indeed necessary to get the upstream PR: https://github.com/OPM/opm-common/pull/1886 through.

I'll let it linger for a while - but personally I have no other suggestions than bringing back in the `const_cast< >`?

bska:
> personally I have no other suggestions than bringing back in the `const_cast< >`?

To me the biggest problem here is the fact that the templatised [`EclMpiSerializer::operator()`](https://github.com/OPM/opm-simulators/blob/9829bfd61a9ba0cbf43dc0c758ca065c3bd44b91/ebos/eclmpiserializer.hh#L46) takes its parameter as reference-to-const.  This is in direct opposition to the specialisation for [`vector`](https://github.com/OPM/opm-simulators/blob/9829bfd61a9ba0cbf43dc0c758ca065c3bd44b91/ebos/eclmpiserializer.hh#L69) which accepts a reference to a mutable object.

@akva2 : Why does `operator()()` take a `const T&` parameter type?

akva2:
the problem is that we have two serialization operations taking a const param (packsize, pack) and one taking a non-const reference (unpack). either we always pass mutable data, which is kinda meh, or we pass non-mutable data and only make it mutable if we are doing unpack. it's nasty, but it's the cost of having a single interface for all three operations.

bska:
> the problem is that we have two serialization operations taking a const param (packsize, pack) and one taking a non-const reference (unpack). either we always pass mutable data, which is kinda meh, or we pass non-mutable data and only make it mutable if we are doing unpack.

Sure, but right now the following call compiles and it really shouldn't
```C++
serializer(2);
```

PR #2690, opened 2020-06-26
Implements the Python step_init() function

hakonhagland:
Note: This PR depends on #2689 which should be merged first.

A resubmission of commit 11eaa3d7 in PR #2403 and PR #2443 and continues the work in #2555 implementing Python bindings to the flow simulator.

The step_init() method initializes the simulation. It is required for the Python script to run step_init() before calling the step() method (which will be implemented in the next commit).

PR #2444, opened 2020-03-09
Implement Python step() method

hakonhagland:
The 7th PR in the splitting of PR #2403 into smaller PRs.

This PR depends on PR #2443 which should be merged first.

## Description

The Python `step()` method advances the simulator one report step. Before calling `step()` for the first time, `step_init()` must have been called.

PR #2735, opened 2020-08-25
Implements a Python step() method.

hakonhagland:
A resubmission of commit 8e4f748 in PR #2403 and PR #2444 and continues the work in #2690 implementing Python bindings to the flow simulator.

The Python `step()` method advances the simulator one report step. Before calling `step()` for the first time, `step_init()` must have been called.


hakonhagland:
jenkins build this please

PR #2739, opened 2020-08-26
Finally fix selection for relaxed tolerance.

blattms:
Default values are MaxStrictIter=8 and RelaxedPvFraction=1.0.
To activate relaxed pv fraction set MaxStrictIter to 0 and
fraction below 1.


blattms:
jenkins build this please

PR #2745, opened 2020-08-27
Continue replacing macros for typetags et al

akva2:


akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
@berndflemisch you have suggested the removal of default property macros for release 2021.04. Do you have any objections to bump the schedule and remove them for 2020.10? I think all the noise is nasty, and since we (after this PR) can build without them, I would really like to avoid the noise in the release... Others with opinions?

berndflemisch:
I just mirrored the way we did in Dumux where there are probably more downstream modules out of our control and we wanted to give people time to adapt. I certainly don't have any objections if you want to proceed faster here.

berndflemisch:
For the time being, you can set `OPM_ENABLE_OLD_PROPERTY_MACROS` to `0` in your build options to avoid the warnings.

akva2:
Yes I know, and I will do that locally of course. But users won't necessarily so I wanted the discussion on it.

akva2:
anything holding this back? a lot of potential for conflicts, would be good to get it in....

bska:
> anything holding this back?

Are you going to address the fact that there are quite a few new uses of `std::tuple` which currently rely on the `<tuple>` header being brought in transitively?  Possibly other library types too, but I didn't check due to the size of the patch.

akva2:
the property system itself brings in <tuple>, it feels a bit unnecessary to put the include all over. but if you insist..

bska:
> the property system itself brings in `<tuple>`,

I see.

> [...] if you insist..

Tricky.  On the one hand, I like all translation units to explicitly include the headers of all types it needs/uses, on the other there is no tradition in our codebase for doing so.  I think I'll merge this now and make a note of addressing it later if I ever touch these files.

PR #2752, opened 2020-09-01
Pass the UDQState member when evaluating UDQ

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/1887


bska:
Upstream PR and reference data update have both been merged.  Merging this now to complete the update.

PR #2756, opened 2020-09-02
[opencl] Fixes shadowing variable warnings.

blattms:
These seemed familar and I would have sworn that I already fixed them
somewhen. Might have been reintroduced by rebasing.


blattms:
jenkins does not compile these, hence merging. ....famous last words :see_no_evil:

PR #2761, opened 2020-09-03
fixed: use the proper linear operator without wells in matrix in serial

### LINKED ISSUE #2759: Excessive time step choping for Norne when using flexible solvers.
Runing `flow --linear-solver-configuration=ilu0 NORNE_ATW2013` results in excessive time step chopping for current master. Same for CPR. Using the default, i.e. without the flexible solver bx skipping the argument, works as before.

akva2:
Closes #2759 

Thanks to @blattms for pointing it out plus implicitly confirming my fix.

### DIFF
--- a/opm/simulators/linalg/ISTLSolverEbos.hpp
+++ b/opm/simulators/linalg/ISTLSolverEbos.hpp
@@ -768,9 +768,16 @@ DenseMatrix transposeDenseMatrix(const DenseMatrix& M)
                     }
 #endif
                 } else {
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

atgeirr:
Looks correct to me, merge when green. This must be one of the first cases in our code that was correct for parallel and wrong for serial rather than the opposite...?

blattms:
Mine looks the same.

PR #2765, opened 2020-09-04
Add regression testing for MULTZ + PINCH::ALL

joakim-hove:
Enable regression testing of the models with the keyword combination MULTZ + PINCH::ALL

PR #2766, opened 2020-09-04
fixed: insideCartElmIdx and outsideCartElmIdx order in parallel

akva2:
in parallel the intersections point in the opposite
direction. this leads to inside > outside, and then the code in
applyAllZMultipliers_ blows up.

Note I'm not at all sure about this, but this is what happens according to my debugger.
Context is https://github.com/OPM/opm-simulators/pull/2765

It runs with this change and applies the same multipliers. Whether or not this is a workaround for another bug or not is up in the air..

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -344,9 +344,12 @@ public:
                 // apply the full face transmissibility multipliers
                 // for the inside ...
 
-                if (useSmallestMultiplier)
-                    applyAllZMultipliers_(trans, insideFaceIdx, insideCartElemIdx, outsideCartElemIdx, transMult, cartDims);
-                else
+                if (useSmallestMultiplier) {
+                    applyAllZMultipliers_(trans, insideFaceIdx,
+                                          std::min(insideCartElemIdx, outsideCartElemIdx),
+                                          std::max(insideCartElemIdx, outsideCartElemIdx),
+                                          transMult, cartDims);
+                } else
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

blattms:
Took me some time to understand why this should be the opposite in paralllel. As we changed the default ordering at some stage (OwnerCellsFirst=1), the ordering changed for faces between owner and overlap.

Somehow I still think that the fix is not complete or has unwanted side effects. applyAllZMultipliers_ calls applyMultipliers_ for non-Z faces, which means they might now be called with a different cell than before, yet, just below your changes there is a applyMultipliers_ call that uses the old approach. Seems a bit inconsistent and might be fishy.

Maybe the change should move into the  if-branch of [ `applyAllZMultipliers_`](https://github.com/OPM/opm-simulators/blob/ddeb5352f38a785bfd426cb2af1a4594120e43f7/ebos/ecltransmissibility.hh#L479-L492) to leave the call to `applyMultipliers_` for the x and y direction as is?

blattms:
Admittedly, i do not get the applyAllMultipliers_ code and maybe the multipliers. Shouldn't the face direction indicate which of the attached cells has the smaller cartesian index and hence the direction of the loop (increasing for ZMinus or decreasing index for ZPlus)?

blattms:
My alternative fix in #2773 should prevent the side effects mentioned.

akva2:
closing this as a replacement, with more understanding behind it, is in the works.

blattms:
> with more understanding behind it

I am flattered, but I am not sure whether that holds here,

PR #2767, opened 2020-09-07
Started to support summary regions

joakim-hove:
See: https://github.com/OPM/opm-common/pull/1891 for context

Will now support arbitrary regions. Quite significant refactorings. 

Usage from other modules is unhcanged - could maybe also benefit from some cleanup?

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -99,7 +99,7 @@ class EclOutputBlackOilModule
     enum { enableEnergy = getPropValue<TypeTag, Properties::EnableEnergy>() };
 
     typedef std::vector<Scalar> ScalarBuffer;
-    typedef std::vector<std::string> StringBuffer;    
+    typedef std::vector<std::string> StringBuffer;
 
     struct FipDataType
     {
@@ -121,7 +121,7 @@ class EclOutputBlackOilModule
         enum WPId
         {
             WellLocationi = 0, //WLi
-            WellLocationj = 1, //WLj          
+            WellLocationj = 1, //WLj
             OilRate = 2, //OR
             WaterRate = 3, //WR
             GasRate = 4, //GR
@@ -137,7 +137,7 @@ class EclOutputBlackOilModule
         };
         static const int numWPValues = 12;
         static const int numWPNames = 2;
-    };  
+    };
         struct WellInjDataType
     {
         enum WIId
@@ -1144,70 +1144,70 @@ public:
 
     }
 
-    // write production report to output 
-    void outputProdLog(size_t reportStepNum, const bool substep, bool forceDisableProdOutput) 
+    // write production report to output
+    void outputProdLog(size_t reportStepNum, const bool substep, bool forceDisableProdOutput)
     {
                 if (!substep) {
-                        
+
                         ScalarBuffer  tmp_values(WellProdDataType::numWPValues, 0.0);
                         StringBuffer  tmp_names(WellProdDataType::numWPNames, "");
                         outputProductionReport_(tmp_values, tmp_names, forceDisableProdOutput);
-                        
-                        const auto& st = simulator_.vanguard().summaryState(); 
+
+                        const auto& st = simulator_.vanguard().summaryState();
                         const auto& schedule = simulator_.vanguard().schedule();
-            
+
                         for (const auto& gname: schedule.groupNames()) {
-                
-                                auto gName = static_cast<std::string>(gname);                
+
+                                auto gName = static_cast<std::string>(gname);
                                 auto get = [&st, &gName](const std::string& vector)
                                 {
                                         const auto key = vector + ':' + gName;
 
                                         return st.has(key) ? st.get(key) : 0.0;
-                                };      
-                
+                                };
+
                                 tmp_names[0] = gname;
-                
-                                tmp_values[2] = get("GOPR"); //WellProdDataType::OilRate 
+
+                                tmp_values[2] = get("GOPR"); //WellProdDataType::OilRate
                                 tmp_values[3] = get("GWPR"); //WellProdDataType::WaterRate
                                 tmp_values[4] = get("GGPR"); //WellProdDataType::GasRate
                                 tmp_values[5] = get("GVPR"); //WellProdDataType::FluidResVol
                                 tmp_values[6] = get("GWCT"); //WellProdDataType::WaterCut
                                 tmp_values[7] = get("GGOR"); //WellProdDataType::GasOilRatio
-                                tmp_values[8] = get("GWPR")/get("GGPR"); //WellProdDataType::WaterGasRatio      
-                
-                                outputProductionReport_(tmp_values, tmp_names, forceDisableProdOutput); 
+                                tmp_values[8] = get("GWPR")/get("GGPR"); //WellProdDataType::WaterGasRatio
+
+                                outputProductionReport_(tmp_values, tmp_names, forceDisableProdOutput);
                         }
-            
+
                         for (const auto& wname: schedule.wellNames(reportStepNum)) {
-                                
-                                // don't bother with wells not on this process 
-                                const auto& defunctWellNames = simulator_.vanguard().defunctWellNames(); 
+
+                                // don't bother with wells not on this process
+                                const auto& defunctWellNames = simulator_.vanguard().defunctWellNames();
                                 if (defunctWellNames.find(wname) != defunctWellNames.end()) {
                                         continue;
                                 }
 
-                                const auto& well = schedule.getWell(wname, reportStepNum);                     
-                                        
-                                // Ignore injector wells  
-                                if (well.isInjector()){ 
+                                const auto& well = schedule.getWell(wname, reportStepNum);
+
+                                // Ignore injector wells
+                                if (well.isInjector()){
                                         continue;
                                 }
-        
+
                                 tmp_names[0] = wname;//WellProdDataType::WellName
-        
-                                
-                                auto wName = static_cast<std::string>(wname);                
+
+
+                                auto wName = static_cast<std::string>(wname);
                                 auto get = [&st, &wName](const std::string& vector)
                                 {
                                         const auto key = vector + ':' + wName;
 
                                         return st.has(key) ? st.get(key) : 0.0;
                                 };
-                                
+
                                 const auto& controls = well.productionControls(st);
                                 using CMode = ::Opm::Well::ProducerCMode;
-                                
+
                                 auto fctl = [](const auto wmctl) -> std::string
                                 {
                                         switch (wmctl) {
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

bska:

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this opm-common=1916 please

joakim-hove:
jenkins build this opm-common=1916 please

joakim-hove:
jenkins build this opm-common=1916 please

joakim-hove:
This is now ready for review

ebos/ecloutputblackoilmodule.hh
```diff
@@ -256,7 +301,7 @@ public:
             else
                 fip_[i].clear();
         }
-        if (!substep || summaryConfig.hasKeyword("FPR") || summaryConfig.hasKeyword("FPRP") || summaryConfig.hasKeyword("RPR")) {
+        if (!substep || summaryConfig.hasKeyword("FPR") || summaryConfig.hasKeyword("FPRP") || !this->RPRNodes_.empty() > 0) {
```
bska, line 304: What does
```C++
!this->RPRNodes_.empty() > 0
```
mean?

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1044,170 +1089,277 @@ public:
...
-        const auto& comm = simulator_.gridView().comm();
-        auto maxElement = std::max_element(fipnum_.begin(), fipnum_.end());
-        size_t ntFip = 0;
-        if ( maxElement != fipnum_.end() ) {
-            ntFip = *maxElement;
+    int regionMax(const std::vector<int>& region) {
+        int max_value = 0;
+        const auto& comm = this->simulator_.gridView().comm();
+        auto maxElement = std::max_element(region.begin(), region.end());
+        if ( maxElement != region.end() ) {
```
bska, line 1096: This can only fail if `region` is `empty()`.  I'd rather phrase this as
```C++
int regionMax(const std::vector<int>& region)
{
    const auto max_value = region.empty() ? 0
        : *std::max_element(region.begin(), region.end());

    return this->simulator_.gridView().comm().max(max_value);
}
```

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1044,170 +1089,277 @@ public:
...
+            }
+        }
+        return rsum_map;
+    }
+
+    Scalar sum(const ScalarBuffer& v) {
+        Scalar s = 0;
+        for (std::size_t i = 0; i < v.size(); i++)
+            s += v[i];
+        return s;
```
bska, line 1169:
```C++
return std::accumulate(v.begin(), v.end(), Scalar{0});
```

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1044,170 +1089,277 @@ public:
...
+
+
 
+
+    // write Fluid In Place to output log
+    void outputFipLog(std::map<std::string, double>& miscSummaryData,  std::map<std::string, std::vector<double>>& regionData, const bool substep)
+    {
+        auto rsum_map = this->makeRegionSums();
+        if (!isIORank_())
+            return;
```
bska, line 1288: I'll be honest&mdash;I think this code looks weird.  As a reader I wonder why we need to `makeRegionSums()` **before** we check if we're the IO rank when we only use the result on the IO rank.

First: Do we need to do that?  Second, if we do need to `makeRegionSums` on all processes, is there no other way of phrasing this?  I'd prefer if the function were named something along the lines of
```C++
accumulateAndExchangeRegionSums()
```

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1772,13 +1924,9 @@ private:
...
 
-    void createLocalFipnum_()
+    void createLocalRegion_(std::vector<int>& region)
     {
-        const auto& gridView = simulator_.vanguard().gridView();
-        if (simulator_.vanguard().eclState().fieldProps().has_int("FIPNUM"))
-            fipnum_ = simulator_.vanguard().eclState().fieldProps().get_int("FIPNUM");
-        else
-            fipnum_.resize(gridView.size(/*codim=*/0), 1);
+        //const auto& gridView = simulator_.vanguard().gridView();
```
bska, line 1929: Either remove this disabled code or reinstate it.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1044,170 +1089,277 @@ public:
...
+
+    std::unordered_map<std::string, RegionSum> makeRegionSums() {
+        std::unordered_map<std::string, RegionSum> rsum_map;
         const Opm::SummaryConfig summaryConfig = simulator_.vanguard().summaryConfig();
-        if (isIORank_()) {
-            // Field summary output
-            for (int i = 0; i<FipDataType::numFipValues; i++) {
-                std::string key = "F" + fipEnumToString_(i);
+
+        rsum_map.insert(std::make_pair("FIPNUM", makeRegionSum(this->regions_.at("FIPNUM"), true)));
```
bska, line 1154:
```C++
rsum_map.emplace("FIPNUM", makeRegionSum(this->regions_.at("FIPNUM"), true));
```

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1044,170 +1089,277 @@ public:
...
+        for (const auto& fip_region : summaryConfig.fip_regions()) {
+            if (fip_region != "FIPNUM") {
+                const auto& region = this->regions_.at(fip_region);
+                RegionSum rsum = makeRegionSum(region, fip_region == "FIPNUM");
+                rsum_map.insert(std::make_pair(fip_region, std::move(rsum)));
+            }
+        }
```
bska, line 1161:
```C++
if (fip_region == "FIPNUM")
    continue;

rsum_map.emplace(fip_region, makeRegionSum(region, false));
```
This saves a level of indentation and automatically activates move semantics.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -114,14 +119,29 @@ class EclOutputBlackOilModule
...
+            case FipDataType::WaterInPlace: return "WIP";
+            case FipDataType::OilInPlace: return "OIP";
+            case FipDataType::GasInPlace: return "GIP";
+            case FipDataType::OilInPlaceInLiquidPhase: return "OIPL";
+            case FipDataType::OilInPlaceInGasPhase: return "OIPG";
+            case FipDataType::GasInPlaceInLiquidPhase: return "GIPL";
+            case FipDataType::GasInPlaceInGasPhase: return "GIPG";
+            case FipDataType::PoreVolume: return "PV";
+            }
+            throw std::logic_error("fip_type:  " + std::to_string(fip_type) + " not recognized");
```
bska, line 136: Please include `<stdexcept>` for `std::logic_error`.

bska (changes requested):
Overall I think this is a good first step, but I'd like to see some additional changes before I approve of this.  For instance, the function naming leaves a little to be desired and some of the constructs are too verbose.

ebos/ecloutputblackoilmodule.hh
joakim-hove: OK

ebos/ecloutputblackoilmodule.hh
joakim-hove: OK

ebos/ecloutputblackoilmodule.hh
joakim-hove: OK

ebos/ecloutputblackoilmodule.hh
joakim-hove: OK

ebos/ecloutputblackoilmodule.hh
joakim-hove: Does not work with `ScalarBuffer` - or at least not on my machine.

ebos/ecloutputblackoilmodule.hh
joakim-hove: Yes - this is weird; I have only refactored the existing code and not changed semantics / public API at all. The weirdness might stand more out now.

The `makeRegionSums()` accumulates over all processes - so it must be called globally, and then the results is only used on io rank. I'll rename the function to `accumulateRegionsSums()`

ebos/ecloutputblackoilmodule.hh
joakim-hove: Fixed

ebos/ecloutputblackoilmodule.hh
joakim-hove: OK

ebos/ecloutputblackoilmodule.hh
joakim-hove: OK - I tried to initialize with the accumulator with literal `0` - maybe it works with `Scalar{0}`

bska: > Does not work with `ScalarBuffer` - or at least not on my machine.

Weird.  `ScalarBuffer` is a `std::vector<Scalar>` so it definitely has `v.begin()` and `v.end()`.  What diagnostic message do you get?  You **may** need to ìnclude `<numeric>` for the `std::accumulate` function.

joakim-hove: It just calculated wrong - I gave up; I did include `<numeric>`. However - it works now when I initialize with `Scalar{0}`

joakim-hove:
OK - I have addressed all(?) the comments; regarding the weirdness I agree - and it is still weird. What is even more weird is the way this code is invoked from the outside:

```
eclwriter::evalSummaryState() {
    ....
    bool substep;
    outputmodule.outputFipLog(..., substep)
}
```
i.e. in the function `eclwriter::evalSummaryState()` we invoke the function `outputmodule::outputFipLog()`- mainly for it's side effect of calculating region sums used for the summary output, but if `substep == false` we also output `Fluid In Place`reports to the PRT file - *via a function named `evalSummaryState`* - in the interest of keeping the amount of changes down I have not touched this weirdness.

joakim-hove:
jenkins build this please

bska (approved):
Thanks for the updates.  I'm happy with the current state of this work.  I'll just make a note to revisit this component later as there are definitely structural problems here.

bska:
Build check successful.  I'll merge into master.

PR #2771, opened 2020-09-07
Allow processing emtpy grid partitions during parallel output writing.

blattms:
Empty grids result in  empty localIdxToGlobalIdx_ mapping of course. This is needed in parallel where empty partitions might occur. 

Note that we currently throw an exception if load balancing results in an empty partition for one process. This is one of the necessary steps to allow such partitions.

### DIFF
--- a/ebos/collecttoiorank.hh
+++ b/ebos/collecttoiorank.hh
@@ -752,9 +752,6 @@ public:
         if (!isParallel())
             return true;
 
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

blattms:
Let's play it safe,

blattms:
jenkins run norne please

blattms:
Problem already solved with PR #2778. Closing

PR #2772, opened 2020-09-07
Prevent another copy of cell to region mapping in RateConverter.

blattms:
RateConverter itself already copies this and offers the mapping.
With this commit we just use that.


blattms:
jenkins build this please.

blattms:
jenkins build this please

bska (approved):
Thanks. I'll merge into master.

PR #2773, opened 2020-09-07
fix insideCartElmIdx and outsideCartElmIdx order for faces in parallel (2)

blattms:
Alternative fix to #2766 that makes cartesian index usage consistent.

only e80a02cc7 is the fix. The rest is cleanup. the fix simply bases the processing of faces on the cartesian element indices to make each face processed only once. Previous we used the local element indices (the two approaches were consistent until #2220).

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -475,14 +475,11 @@ private:
         if (insideFaceIdx > 3) { // top or or bottom
             Scalar mult = 1e20;
             unsigned cartElemIdx = insideCartElemIdx;
+            assert(insideFaceIdx==5); // as insideCartElemIdx < outsideCartElemIdx holds
             // pick the smallest multiplier while looking down the pillar untill reaching the other end of the connection
             // for the inbetween cells we apply it from both sides
             while (cartElemIdx != outsideCartElemIdx) {
-                if (insideFaceIdx == 4 || cartElemIdx !=insideCartElemIdx )
-                    mult = std::min(mult, transMult.getMultiplier(cartElemIdx, Opm::FaceDir::ZMinus));
-                if (insideFaceIdx == 5 || cartElemIdx !=insideCartElemIdx)
-                    mult = std::min(mult, transMult.getMultiplier(cartElemIdx, Opm::FaceDir::ZPlus));
-
+                mult = std::min(mult, transMult.getMultiplier(cartElemIdx, Opm::FaceDir::ZPlus));
                 cartElemIdx += cartDims[0]*cartDims[1];
             }
             trans *= mult;
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -273,13 +273,13 @@ public:
                         A * (n*d)/(d*d);
                 }
 
-                // we only need to calculate a face's transmissibility
-                // once...
-                if (elemIdx > outsideElemIdx)
-                    continue;
-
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

ebos/ecltransmissibility.hh
```diff
@@ -475,14 +475,11 @@ private:
...
+            assert(insideFaceIdx==5); // as insideCartElemIdx < outsideCartElemIdx holds
             // pick the smallest multiplier while looking down the pillar untill reaching the other end of the connection
             // for the inbetween cells we apply it from both sides
             while (cartElemIdx != outsideCartElemIdx) {
-                if (insideFaceIdx == 4 || cartElemIdx !=insideCartElemIdx )
-                    mult = std::min(mult, transMult.getMultiplier(cartElemIdx, Opm::FaceDir::ZMinus));
-                if (insideFaceIdx == 5 || cartElemIdx !=insideCartElemIdx)
-                    mult = std::min(mult, transMult.getMultiplier(cartElemIdx, Opm::FaceDir::ZPlus));
-
+                mult = std::min(mult, transMult.getMultiplier(cartElemIdx, Opm::FaceDir::ZPlus));
```
bska, line 482: This looks dangerous.  Would you not be missing `MULTZ-` this way?

blattms: Could be (I am able to judge this), but then that was the case also before.

I think the original code gave the impression that each face might be processed for both sides. But that has never been the case. We always processed Z faces only from low to high cartesian indices here, hence ZPLUS.

blattms:
#2765 should of course be merged first.

ebos/ecltransmissibility.hh
blattms: Indeed, looking at [TransMult.cpp#L91-L97](/OPM/opm-common/blob/master/src/opm/parser/eclipse/EclipseState/Grid/TransMult.cpp#L91-L97), I totally agree that this is very dangerous as the multiplier is only stored for one direction and we might miss it. Maybe @joakim-hove can shed some light which direction is chosen and when.

blattms: I correct myself: there might be mutlipiers stored for each direction. Hence you might have found a bug that has been lingering for a long time.

joakim-hove: I don't really know anything about this code - I just accidentally stumbled over a problem here - which I am not familiar with at all.

joakim-hove:
>#2765 should of course be merged first.

It is merged now

blattms:
BTW we are calling applyMultipliers_ for the other direction [here](/OPM/opm-simulators/pull/2773/commits/e80a02cc7d3e420eddeed00803d13b2cdcbddbb4/ebos/ecltransmissibility.hh#L352), just not for all. I guess that should change?

blattms:
Will rebase to pick up the new test.

blattms:
jenkins build this please

blattms:
Closing in favor of #2776

blattms:
jenkins build this please

blattms:
Reopening.
Pinch treatment is of course buggy here like before, but it fixes the cartesian element index bug and simpiifies the buggy code and makes it a tiny bit less opaque.

I guess we should consider merging if jenkins agrees.

blattms:
jenkins build this please

blattms:
I see, one cannot reopen a closed branch that was force pushed to.

PR #2768, opened 2020-09-07
Add pinch testing parallel

joakim-hove:
Add parallell test of the MULTZ + PINCH::ALL situation

Issue #2665, opened 2020-06-10
Does CollectDataToIORank::isGlobalIdxOnThisRank Return True for Ghost Cells?


bska:
The [EBOS Output Module](https://github.com/OPM/opm-simulators/blob/b1adbd45cbe071bdec87a39308703b6e39282301/ebos/ecloutputblackoilmodule.hh) contains the following lines
```C++
// Initialize block output
for (const auto& node: summaryConfig) {
    if (node.category() == SummaryConfigNode::Category::Block) {
        if (collectToIORank.isGlobalIdxOnThisRank(node.number() - 1)) {
            std::pair<std::string, int> key = std::make_pair(node.keyword(), node.number());
            blockData_[key] = 0.0;
        }
    }
}
```
This code&mdash;lines 187 through 195&mdash;creates locations to host block-related summary quantities like
```
BPR
  13 29 8 /
/
```
which are subsequently populated in [`processElement()`](https://github.com/OPM/opm-simulators/blob/b1adbd45cbe071bdec87a39308703b6e39282301/ebos/ecloutputblackoilmodule.hh#L688-L725) and communicated to the I/O rank using [`CollectDataToIORank::collect()`](https://github.com/OPM/opm-simulators/blob/b1adbd45cbe071bdec87a39308703b6e39282301/ebos/eclwriter.hh#L281-L282).

The reason I ask is that I'm trying to analyse a weird symptom in Flow's summary file output in the context of parallel simulation runs.  If I add the summary vectors below to [`NORNE_ATW2013`](https://github.com/OPM/opm-tests/blob/64f2bef4cba614a02442f67de9ceb7c8225c043c/norne/NORNE_ATW2013.DATA) I occasionally get zeros in the summary output from a parallel run when the a sequential run produces non-zero values.
```diff
--- a/norne/NORNE_ATW2013.DATA
+++ b/norne/NORNE_ATW2013.DATA
@@ -449,6 +449,18 @@ INCLUDE
 INCLUDE
  './INCLUDE/SUMMARY/wpave.inc' /

+BPR
+  13 29 8 /
+/
+
+BWSAT
+  13 29 8 /
+/
+
+BGSAT
+  13 29 8 /
+/
+
```
Using the [summary](https://github.com/OPM/opm-common/blob/6a742a20fd6a540a16be343ebb864a5d7619a3cb/test_util/summary.cpp) program as follows
```sh
[...]

PR #2776, opened 2020-09-08
Apply minimum of Z multiplier for PINCH(5)==ALL, Prepare for PINCH(5)==TOP

blattms:
Proposal of fix to potential bug found in discussion of #2773, it includes that PR.
Using just one direction seemed wrong as we use both the smallest multiplier is requested and TransMult might store values for both directions. Hence this commit makes the behaviour consistent in both cases.
    
Note that we take the minimum in one direction and apply that and then in the other directions and apply that. Hence the two minimums might belong to different faces of the cartesian grid.  

One could of course insist that the multipliers belong to the same undirectional face.

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -504,7 +504,7 @@ private:
             }
 
             trans *= mult;
-            //applyMultipliers_(trans, outsideFaceIdx, outsideCartElemIdx, transMult); // Uncommenting makes the tests succeed
+            applyMultipliers_(trans, outsideFaceIdx, outsideCartElemIdx, transMult);
         }
         else
         {
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -345,11 +345,19 @@ public:
                 // for the inside ...
 
                 if (useSmallestMultiplier)
-                    applyAllZMultipliers_(trans, insideFaceIdx, insideCartElemIdx, outsideCartElemIdx, transMult, cartDims);
+                {
+                    // Currently PINCH(4) is never queries and hence  PINCH(4) == TOPBOT is assumed
+                    // and in this branch PINCH(5) == ALL holds
+                    applyAllZMultipliers_(trans, insideFaceIdx, outsideFaceIdx, insideCartElemIdx,
+                                          outsideCartElemIdx, transMult, cartDims,
+                                          /* pinchTop= */ false);
+                }
                 else
+                {
                     applyMultipliers_(trans, insideFaceIdx, insideCartElemIdx, transMult);
-                // ... and outside elements
-                applyMultipliers_(trans, outsideFaceIdx, outsideCartElemIdx, transMult);
+                    // ... and outside elements
+                    applyMultipliers_(trans, outsideFaceIdx, outsideCartElemIdx, transMult);
+                }
 
                 // apply the region multipliers (cf. the MULTREGT keyword)
                 Opm::FaceDir::DirEnum faceDir;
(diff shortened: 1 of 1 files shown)


bska:
Well, actually the final multiplier of a horizontal face&mdash;i.e., one that separates cells in the vertical direction, between layers `K` and `K+1`&mdash;is the product of `MULTZ` in the cell above with `MULTZ-` in the cell below.  In pseudo-code
```C++
const auto mult_z = multz(i, j, k) * multz_(i, j, k + 1);
```
with underscore denoting the `Z-` direction.  Similarly for `I` and `J` directions:
```C++
const auto mult_x = multx(i, j, k) * multx_(i + 1, j, k);
const auto mult_y = multy(i, j, k) * multy_(i, j + 1, k);
```
Of course in **most** case `MULTZ-` is `1.0` for all cells so the practical impact of making a mistake here is usually nil.

blattms:
Then not take any chances and use a minimum per undirectional face. One (or at least I) usually get bitten by them eventually.

bska:
> Then not take any chances and use a minimum per undirectional face.

That may be too large.  If the pertinent `MULTZ` and `MULTZ-` both happen to be 0.5, then their product is 0.25 while the minimum of course is 0.5.

joakim-hove:
Observe that the test which initiated this dicsussion can be found here: #2768

akva2:
if you run the test with this PR, another issue bites
```
115: Keyword: BPR:1,1,18, origin Summary file
115: 
115: Value index                 = 0
115: (first value, second value) = (219.253, 0)
115: 
115: Program threw an exception: [../opm-common/test_util/EclRegressionTest.cpp:195] Deviations exceed tolerances.
115: The absolute deviation is 219.253, and the tolerance limit is 0.02.
115: The relative deviation is -1, and the tolerance limit is 1e-05.
```
so while the parallel run now works, we still have to fix this before the test can be added.

bska:
> ```
> 115: Keyword: BPR:1,1,18, origin Summary file
> 115: 
> 115: Value index                 = 0
> 115: (first value, second value) = (219.253, 0)
> ```

This **may** be related to an/the issue that prompted the discussion in #2665.  There, too, I noticed that we would occasionally get zeros in the output, in a manner that **might** suggest a race condition.

blattms:
last commit is broken, I really should learn how to test before pushing.

blattms:
With dc286d0 we hopefully take the minimum as outlined by @bska. Sequential test at lest succeeds. Parallel still fails.

blattms:
I will take a look at the parallel pinch test as this is a nice small test for debugging. thanks for that one, @joakim-hove

bska:
> With [dc286d0](https://github.com/OPM/opm-simulators/commit/dc286d0a05fc3ed4656475f62ac11fc1a04c40d0) we hopefully take the minimum as outlined by @bska.

Thanks, but on more careful reading of the background material the minimum value should use `MULTZ` only (i.e., `MULTZ-` does not go into this derivation).  Sorry for creating confusion.

That said, I do believe there's still some work to do in this area.  All of this is only supposed to apply to pinched-out connections in the vertical direction (i.e., across layers for which the bottom cell has a layer ID that is **more** than 1 greater than the top cell's layer ID,) and only if the pinch-out option (item 4) is `TOPBOT` at that.

Unless I'm missing something, the [code](https://github.com/OPM/opm-simulators/blob/f2e585bf1fec764ad598f98f45e6f94e52fff762/ebos/ecltransmissibility.hh#L347-L356) in this PR as well as the current master sources applies this treatment to **all** vertical connections, irrespective of pinch-out status, if the "multz" mode happens to be `ALL` (admittedly not the default setting).

Any comments?

blattms:
> Thanks, but on more careful reading of the background material the minimum value should use `MULTZ` only (i.e., `MULTZ-` does not go into this derivation).

Not at all? Not even the normal one as before?

> 
> That said, I do believe there's still some work to do in this area. All of this is only supposed to apply to pinched-out connections in the vertical direction (i.e., across layers for which the bottom cell has a layer ID that is **more** than 1 greater than the top cell's layer ID,) and only if the pinch-out option (item 4) is `TOPBOT` at that.
> 

Apart fromTOPBOT treatment that might already be the case.

bska:
> > Thanks, but on more careful reading of the background material the minimum value should use `MULTZ` only (i.e., `MULTZ-` does not go into this derivation).
> 
> Not at all? Not even the normal one (without taking the minimum and honoring the pinched out cells) as before?

Let's try to clarify

 1. Regular connections in the vertical direction (i.e., between layers `K` and `K + 1`) use the product of `MULTZ` in layer `K` and `MULTZ-` in layer `K + 1`.  Since `MULTZ-` is **typically** 1.0, that usually reduces to just `MULTZ` in layer `K`.

  2. Multiplier handling across pinched-out cells&mdash;vertical connections between layers `K` and `K + k` (`k > 1`)&mdash;depends on items 4 and 5 in the `PINCH` keyword:
      1. If `PINCH(4) == TOPBOT`, then derive handling from `PINCH(5)`.  In both cases (`PINCH(5) == TOP` and `PINCH(5) == ALL`) do we only use `MULTZ` values to derive the final multiplier&mdash;either `MULTZ` from top-most cell, or minimum of `MULTZ` across all pinched-out cells.
      2. Otherwise, i.e., when `PINCH(4) == ALL`, then ignore `PINCH(5)` altogether

> > That said, I do believe there's still some work to do in this area. All of this is only supposed to apply to pinched-out connections in the vertical direction (i.e., across layers for which the bottom cell has a layer ID that is **more** than 1 greater than the top cell's layer ID,) and only if the pinch-out option (item 4) is `TOPBOT` at that.
>
> Apart from `TOPBOT` treatment that might already be the case.

Okay, you're probably right.  It was just not apparent to me whether the `applyAll*` call site (or function) distinguished between pinched-out and regular connections in the vertical direction.

blattms:
Hopefully the new fix is according to your description @bska. It also prepares for PINCH(5)==TOP but that is still deactivated.

blattms:
jenkins build this please

ebos/ecltransmissibility.hh
```diff
@@ -465,30 +472,41 @@ private:
...
-            while (cartElemIdx != outsideCartElemIdx) {
-                if (insideFaceIdx == 4 || cartElemIdx !=insideCartElemIdx )
-                    mult = std::min(mult, transMult.getMultiplier(cartElemIdx, Opm::FaceDir::ZMinus));
-                if (insideFaceIdx == 5 || cartElemIdx !=insideCartElemIdx)
+        if (faceIdx > 3) { // top or or bottom
+            assert(faceIdx==5);
+            assert(highCartElemIdx > lowCartElemIdx);
+            Scalar mult = transMult.getMultiplier( --highCartElemIdx, Opm::FaceDir::ZPlus);
+
+            if ( pinchTop )
```
bska, line 491: Should probably say `! pinchTop` here.

blattms: Indeed.

blattms: rebased

blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
Sorry jenkins if it seems like wasting your time (this will fail)

blattms:
jenkins build this please

blattms:
As expected [build 1232](https://ci.opm-project.org/job/opm-simulators-PR-builder/1232/) failed.

blattms:
jenkins build this please

blattms:
Jenkins succeeded with c2362da. Hence the tests assume that in addition to the minimum of MULTZ+ along the column, the assume that MULTZ- of the cells with the highest cartesian index is also applied.

Now I am sufficiently confused. I also tried to understand what is in the reference manual, but am still not 100% sure. The documentation of PINCH(5) for ALL : "ALL results in the pinch-out transmissibility being calculated from the minimum value of the MULTZ of the active cell at the top of the   pinch-out and  all   the   inactive   cells   in   the   pinch-out   vertical column. "

I guess this in line with bards description that c2362da (multiply by MULTZ- of the bottom cell).
If that isindeed the way forward, then I think somebody needs to double check the failing tests without it.

bska:
> Hence the tests assume that in addition to the minimum of MULTZ+ along the column, the assume that MULTZ- of the cells with the highest cartesian index is also applied.

I agree.  That's what the current set of regression tests assume.  I also don't want to fight this too hard before we have a case that clearly demonstrates the "correct" behaviour.  By including c2362da, it **could** be argued that the entire `MULTZ` treatment is "just" a way of defining an appropriate `MULTZ` value for the top cell (smallest Cartesian index) in the `TOPBOT` transmissibility strategy while the non-neighbouring connection's `MULTZ-` value is always picked from the bottom cell (largest Cartesian index).

Unless you'd like to make any further changes here I suggest we merge this in its current state.

blattms:
Agreed. In it goes

blattms:
As a last safety net:

blattms:
jenkins run norne please

bska:
> jenkins run norne please

Success!  I'll merge this into master now.  Thanks a lot for sticking to this and responding to my changing repsonses.

PR #2781, opened 2020-09-09
fix insideCartElmIdx and outsideCartElmIdx order for faces in parallel (2) 

blattms:
It really fixes only this and tries to make the code simpler and less opaque.

There are still other bugs in the pinch treatment, tough (see #2776). But these will need some update_data and are more risky. One should consider merging this before


blattms:
jenkins build this please

bska (approved):
Cool.  Let's merge this now and deal with the multiplier handling in the dedicated PR #2776.

PR #2782, opened 2020-09-10
Fix oil water for MSW

totto82:
Fixes for oil-water + MSW simulations. A test case will follow.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell.hpp
+++ b/opm/simulators/wells/MultisegmentWell.hpp
@@ -59,15 +59,15 @@ namespace Opm
         // TODO: we need to have order for the primary variables and also the order for the well equations.
         // sometimes, they are similar, while sometimes, they can have very different forms.
 
-        // TODO: the following system looks not rather flexible. Looking into all kinds of possibilities
-        static const bool gasoil = numEq == 2 && (Indices::compositionSwitchIdx >= 0);
-        static const int GTotal = 0;
-        static const int WFrac = gasoil? -1000: 1;
-        //For oilwater system the GFrac is set to 2 (the same as SPres) but it is never used
-        static const int GFrac = gasoil? 1 : 2;
-        static const int SPres = numEq;
+        static const bool has_gas = (Indices::compositionSwitchIdx >= 0);
+        static const bool has_water = (Indices::waterSaturationIdx >= 0);
 
-        ///  the number of well equations // TODO: it should have a more general strategy for it
+        static const int GTotal = 0;
+        static const int WFrac = has_water? 1: -1000;
+        static const int GFrac = has_gas? has_water + 1 : -1000;
+        static const int SPres = has_gas + has_water + 1;
+
+        //  the number of well equations  TODO: it should have a more general strategy for it
         static const int numWellEq = getPropValue<TypeTag, Properties::EnablePolymer>() ? numEq : numEq + 1;
 
         using typename Base::Scalar;
--- a/opm/simulators/wells/MultisegmentWell.hpp
+++ b/opm/simulators/wells/MultisegmentWell.hpp
@@ -60,11 +60,10 @@ namespace Opm
         // sometimes, they are similar, while sometimes, they can have very different forms.
 
         // TODO: the following system looks not rather flexible. Looking into all kinds of possibilities
-        // TODO: gas is always there? how about oil water case?
-        // Is it gas oil two phase case?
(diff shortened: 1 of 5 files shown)


totto82:
jenkins build this please

GitPaean:
Is there a test case for testing running?

totto82:
Not yet, but its on its way. I think you should be able to review this without testing it. It's just add guards for non-existing phases.

GitPaean:
> Not yet, but its on its way. I think you should be able to review this without testing it. It's just add guards for non-existing phases.

Sure. No problem. Without a test case, this PR will not be protected.  I can do a review without testing as soon as possible, later today.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2176,8 +2180,12 @@ namespace Opm
         resWell_[seg][SPres] -= accelerationPressureLoss.value();
         duneD_[seg][seg][SPres][SPres] -= accelerationPressureLoss.derivative(SPres + numEq);
         duneD_[seg][seg][SPres][GTotal] -= accelerationPressureLoss.derivative(GTotal + numEq);
-        duneD_[seg][seg_upwind][SPres][WFrac] -= accelerationPressureLoss.derivative(WFrac + numEq);
-        duneD_[seg][seg_upwind][SPres][GFrac] -= accelerationPressureLoss.derivative(GFrac + numEq);
+        if (FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx)) {
```
GitPaean, line 2183: I think here should be waterPhaseIdx.

opm/simulators/wells/MultisegmentWell.hpp
```diff
@@ -60,13 +60,12 @@ namespace Opm
         // sometimes, they are similar, while sometimes, they can have very different forms.
 
         // TODO: the following system looks not rather flexible. Looking into all kinds of possibilities
-        // TODO: gas is always there? how about oil water case?
-        // Is it gas oil two phase case?
         static const bool gasoil = numEq == 2 && (Indices::compositionSwitchIdx >= 0);
```
GitPaean, line 63: This part has been confusing to read. Let us see whether we can find some good way in doing it. Probably we can do through the `FluidSystem`.

GitPaean:
We can use jenkins to check whether this PR impact other cases we should handle, which makes it mergeable. 

But we still not sure whether this can run a case with its intended purpose. When you get your test case done, please update as soon as possible.

opm/simulators/wells/MultisegmentWell.hpp
GitPaean: One suggestion is as follows, 
```c++
          static const bool has_water = ....;
          static const bool has_gas = ....;
          static const int GTotal = 0;
          static const int WFrac = has_water ? 1 : -1000;
          static const int GFrac = has_gas ? has_water + has_gas : - 1000;
          static const int SPres = has_water + has_gas + 1;
```

The good thing about this way is easier to understand. With `gasoil`, we use some negative logic to tell whether water is there, which is why the code can be confusing to understand or more difficult to modify. 

For `SPres`, it might not be good to relate it directly to `NumEq`, which is the number of the reservoir equations. Maybe it is better to associate with `NumWellEq`. 

You can also name `-1000` as some named variable to indicate that it is just for an invalid index. 

About how to get `has_water` and `has_gas`. At the beginning I was thinking about `FluidSystem`. But it has some problem in compiling, not sure why.  Maybe `Indices`? (`Indices::waterEnabled` and `Indices::gasEnabled`). You understand these much better than me.

totto82:
Now we have a test case. I have verified that it works perfectly. 

https://github.com/OPM/opm-tests/pull/372

totto82:
I will add it as a test case with ref. data

totto82:
jenkins build this opm-tests=375 please

akva2:
this is missing the update_reference_data.sh change

totto82:
> this is missing the update_reference_data.sh change

I hope this works. Now with correct name.

totto82:
jenkins build this opm-tests=376 please

bska:
jenkins build this opm-tests=376 please

totto82:

bska:
jenkins build this opm-tests=377 please

bska:
The updated set of reference data is installed on Jenkins.  I'll merge this now to complete the new fix.  Thanks a lot.

PR #2785, opened 2020-09-14
Improves error handling and reporting in parallel.

### LINKED ISSUE #2764: [MPI] error message is not printed when parser fails if HAVE_MPI is true
I have a deck that abort without any error message due to line 254 in readDeck.cpp

`MPI_Abort(MPI_COMM_WORLD, EXIT_FAILURE);`

In debug mode I can read the failureMessage and it points to the real issue in the deck, but the failureMessage is not printed in the terminal. 

Another way to provoke an error without a failure message is to try to run flow with a deck that doesn't exist. It will abort but not tell you why.

blattms:
As the ErrorGuard also dumps warnings we now always dump it (previously only on error) to get these messages in the
console.

If there are error encountered, we log a meaningful error message (the real cause was missing previously) and do a
graceful exit after MPI_Finalize.

Close #2764

### DIFF
--- a/opm/simulators/utils/readDeck.cpp
+++ b/opm/simulators/utils/readDeck.cpp
@@ -224,6 +224,7 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
             if (!summaryConfig)
                 summaryConfig = std::make_unique<Opm::SummaryConfig>(*deck, *schedule, eclipseState->getTableManager(), *parseContext, *errorGuard);
 
+            Opm::checkConsistentArrayDimensions(*eclipseState, *schedule, *parseContext, *errorGuard);
         }
         catch(const std::exception& e)
         {
@@ -272,7 +273,6 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
 #if HAVE_MPI
         Opm::eclStateBroadcast(*eclipseState, *schedule, *summaryConfig);
 #endif
-        Opm::checkConsistentArrayDimensions(*eclipseState, *schedule, *parseContext, *errorGuard);
     }
     catch(const std::exception& e)
     {
--- a/opm/simulators/utils/readDeck.cpp
+++ b/opm/simulators/utils/readDeck.cpp
@@ -170,9 +170,7 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
         errorGuard = std::make_unique<ErrorGuard>();
     }
 
-#if HAVE_MPI
-    int parseSuccess = 0;
-#endif
+    int parseSuccess = 1; // > 0 is success
     std::string failureMessage;
 
     if (rank==0) {
@@ -223,13 +221,12 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
             }
             if (!summaryConfig)
                 summaryConfig = std::make_unique<Opm::SummaryConfig>(*deck, *schedule, eclipseState->getTableManager(), *parseContext, *errorGuard);
-#if HAVE_MPI
-            parseSuccess = 1;
-#endif
+
         }
         catch(const std::exception& e)
         {
             failureMessage = e.what();
+            parseSuccess = 0;
         }
     }
 #if HAVE_MPI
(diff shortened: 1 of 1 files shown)


blattms:
these warnings from ErrorGuard seem very useful as they contain line numbers. Unfortunately, they do not seem to end up in the PRT files (There are similar warning but with less information).

blattms:
jenkins build this please

bska:
> Unfortunately, they do not seem to end up in the PRT files (There are similar warning but with less information).

Nothing goes into the PRT unless it goes through the logging system.

bska (commented):
Just a couple of questions from me, otherwise no reservations.

opm/simulators/utils/readDeck.cpp
```diff
@@ -241,28 +238,66 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
...
     }
+#endif
+
+    if (*errorGuard) { // errors encountered
+        parseSuccess = 0;
+    }
+
+    // print errors and warnings!
+    errorGuard->dump();
+    errorGuard->clear();
```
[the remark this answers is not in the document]
blattms: Not here, as only rank 0 used errorGuard when creating the eclState, Schedule, etc.

 But below we do when the dimensions are checked on every rank in Opm::checkConsistentArrayDimensions. 

Maybe we should call Opm::checkConsistentArrayDimensions above for the undistributed stuff, too. Then everything below should only point to problems in flow.

opm/simulators/utils/readDeck.cpp
```diff
@@ -241,28 +238,66 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
...
+    if (!parseSuccess)
+    {
+        if (rank == 0)
+        {
+            OpmLog::error(std::string("Unrecoverable errors were encountered while loading input: ")+failureMessage);
+        }
+#if HAVE_MPI
+        MPI_Finalize();
+#endif
+        exit(1);
```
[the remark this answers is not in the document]
blattms: done.

opm/simulators/utils/readDeck.cpp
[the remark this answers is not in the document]
bska: >> Do we risk this getting output by all processes now?
>
> Not here, as only rank 0 used errorGuard when creating the eclState, Schedule, etc.

👍 

> But below we do when the dimensions are checked on every rank in `Opm::checkConsistentArrayDimensions`.
> 
> Maybe we should call `Opm::checkConsistentArrayDimensions` above for the undistributed stuff, too. Then everything below should only point to problems in flow.

That's a good suggestion!

bska (approved):
I like the current state.  Do you intend to put the call to `checkConsistentArrayDimensions()` on rank 0 only as part of this PR or will that be the subject of future work?

opm/simulators/utils/readDeck.cpp
```diff
@@ -241,28 +240,66 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
...
-            errorGuard->clear();
+        if (rank == 0)
+        {
+            OpmLog::error(std::string("Unrecoverable errors were encountered while loading input: ")+failureMessage);
         }
-        MPI_Abort(MPI_COMM_WORLD, EXIT_FAILURE);
+#if HAVE_MPI
+        MPI_Finalize();
+#endif
+        exit(1);
```
bska, line 265: This should be
```C++
std::exit(EXIT_FAILURE);
```
too&mdash;preferably with a `std::` prefix.

blattms:
Just check what checkConsistentArrayDimensions does and seems like it checks
the welldims. that is perfectly suited for rank 0 and will save us output
from other ranks. Will change.

blattms:
Done with some refactoring.

blattms:
jenkins build this please

opm/simulators/utils/readDeck.cpp
```diff
@@ -299,7 +274,7 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
 #if HAVE_MPI
         MPI_Finalize();
 #endif
-        exit(EXIT_FAILURE);
+        exit(1);
```
akva2, line 277: why was it changed back to 1?

opm/simulators/utils/readDeck.cpp
blattms: done

opm/simulators/utils/readDeck.cpp
blattms: There were two of those and the changed one got lost when refactoring. Anyway, fixed.

blattms:
jenkins build this please

bska (approved):
Thanks for the updates. I like this and I'll merge into master now.

PR #2701, opened 2020-07-02
relaxed tolerance for CNV if only small fraction of PV is violated

blattms:
Previously we used relaxed tolerance once a certain number of Newton steps was exceeded. Now we check for all cells violating CNV locally and if their pore volume is less than a certaun fraction (default 3%) we use the relaxed tolerance (default: 1e9)

Original idea originated from Norce / @andlaus / @totto82 and was introduced in ebos.

This decreases runtime 

| model | branch | total time |
|---|---|---|
| Norne | master | 614 |
| Norne | PR  | 552|
| Model2 | master | 3411 |
| Model2 | PR | 2876 |

If somebody has better name than RelaxedMaxPvFraction for the command line parameter then I am happy to change them. Maybe they should be the same as for ebos (EclNewtonRelaxedVolumeFraction)?

PR #2788, opened 2020-09-16
New numerical defaults

alfbr:
Following the implementation in https://github.com/OPM/opm-simulators/pull/2701 it is natural to revisit our default numerical parameters. This is a tedious process, checking numerical performance, scaling and accuracy of solution. Given the many parameters involved, I suggest doing it in a step-wise fashion. This is the first step. I have tested on a large number of models at this point and believe that the changes here consistently improve computational time, both on single- and MPI-simulations. I also believe the accuracy of the solution and robustness of the numerics is improved across the board. Hence, I suggest this is reviewed/merged, while work continues to find more improvements.

### DIFF
--- a/opm/simulators/flow/BlackoilModelParametersEbos.hpp
+++ b/opm/simulators/flow/BlackoilModelParametersEbos.hpp
@@ -164,7 +164,7 @@ struct MaxResidualAllowed<TypeTag, TTag::FlowModelParameters> {
 template<class TypeTag>
 struct RelaxedMaxPvFraction<TypeTag, TTag::FlowModelParameters> {
     using type = GetPropType<TypeTag, Scalar>;
-    static constexpr type value = 1.0;
+    static constexpr type value = 0.03;
 };
 template<class TypeTag>
 struct ToleranceMb<TypeTag, TTag::FlowModelParameters> {
@@ -179,7 +179,7 @@ struct ToleranceCnv<TypeTag, TTag::FlowModelParameters> {
 template<class TypeTag>
 struct ToleranceCnvRelaxed<TypeTag, TTag::FlowModelParameters> {
     using type = GetPropType<TypeTag, Scalar>;
-    static constexpr type value = 1e9;
+    static constexpr type value = 1;
 };
 template<class TypeTag>
 struct ToleranceWells<TypeTag, TTag::FlowModelParameters> {
@@ -206,7 +206,7 @@ struct MaxSinglePrecisionDays<TypeTag, TTag::FlowModelParameters> {
 };
 template<class TypeTag>
 struct MaxStrictIter<TypeTag, TTag::FlowModelParameters> {
-    static constexpr int value = 8;
+    static constexpr int value = 0;
 };
 template<class TypeTag>
 struct SolveWelleqInitially<TypeTag, TTag::FlowModelParameters> {
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -764,7 +764,7 @@ opm_set_test_driver(${PROJECT_SOURCE_DIR}/tests/run-restart-regressionTest.sh ""
 
 # Cruder tolerances for the restarted tests
 set(abs_tol_restart 2e-1)
-set(rel_tol_restart 4e-5)
+set(rel_tol_restart 5e-5)
 
 add_test_compare_restarted_simulation(CASENAME spe1
                                       FILENAME SPE1CASE2_ACTNUM
@@ -829,7 +829,7 @@ if(MPI_FOUND)
 
   # Different tolerances for these tests
   set(abs_tol_parallel 0.02)
-  set(rel_tol_parallel 1e-5)
+  set(rel_tol_parallel 8e-5)
   set(coarse_rel_tol_parallel 1e-2)
 
   add_test_compare_parallel_simulation(CASENAME spe1
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -764,7 +764,7 @@ opm_set_test_driver(${PROJECT_SOURCE_DIR}/tests/run-restart-regressionTest.sh ""
 
 # Cruder tolerances for the restarted tests
 set(abs_tol_restart 2e-1)
-set(rel_tol_restart 5e-5)
+set(rel_tol_restart 4e-4)
 
 add_test_compare_restarted_simulation(CASENAME spe1
                                       FILENAME SPE1CASE2_ACTNUM
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -497,7 +497,8 @@ struct RestartWritingInterval<TypeTag, TTag::EclBaseProblem> {
 // as default if experimental mode is enabled.
 template<class TypeTag>
 struct EclEnableDriftCompensation<TypeTag, TTag::EclBaseProblem> {
-    static constexpr bool value = getPropValue<TypeTag, Properties::EnableExperiments>();
+    static constexpr bool value = true;
+
 };
 
 // By default, we enable the debugging checks if we're compiled in debug mode


alfbr:
benchmark please

alfbr:
jenkins build this please

opm/simulators/flow/BlackoilModelParametersEbos.hpp
```diff
@@ -206,7 +206,7 @@ struct MaxSinglePrecisionDays<TypeTag, TTag::FlowModelParameters> {
 };
 template<class TypeTag>
 struct MaxStrictIter<TypeTag, TTag::FlowModelParameters> {
-    static constexpr int value = 8;
+    static constexpr int value = 0;
```
atgeirr, line 209: This raised my eyebrows... No strict iterations seems very dangerous, even though the relaxed max fraction is low. Your testing makes me relax (ha...) a bit though. Did you try to use a somewhat stricter value for the ToleranceCnvRelaxed? Such as 0.2 or 0.1?

alfbr: I have not looked into the implementation, but basically I was of the impression that combining it with RelaxedMaxPvFraction basically means all iterations are strict. The relaxed tolerance only applies to 3% of the pore volume. Moreover, the relaxed tolerance is tightened from 1e9 to 1. In effect, I believe these are far stricter settings than before. Yes I have played with even lower values for the relaxed tolerance, including 0.1 and 0.4. In some cases it is detrimental to performance (computational time almost doubling).

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.036    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.033    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.125    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.102    |



View result details @ https://www.ytelses.com/opm/?page=result&id=888

opm/simulators/flow/BlackoilModelParametersEbos.hpp
totto82: > I have not looked into the implementation, but I was of the impression that combining it with RelaxedMaxPvFraction basically means all iterations are strict. Specifically, that the relaxed tolerance here only applies to 3% of the pore volume. 

Your understanding is correct. One could envision putting a small number for MaxStriciter and combine it with the RelaxedMaxPvFraction. Then the relaxation of the tolerance for  3%  of the pore volume will only happen when iter>maxStrictIter. I have not yet seen any cases where this is necessary. I can also add that I have tested these and similar options for many simulations during the last year. And it seems to work as long as the relaxed tolerance is ~1. 

@alfbr Have you checked the results for the failing tests?

alfbr: > @alfbr Have you checked the results for the failing tests?

No I have not, I will take  a look.

alfbr:
@totto82 I have now gone through all the test failures. Most if not all are caused by the drift compensation giving a more correct pressure solution. I failed to fully understand two of the failed tests (though I do not think that should hold back merging this). The first is SPE1CASE2_ROCK2DTR, which fails for one Flow-exclusive restart-vector, namely PRESSROCC. Those are small deviations probably pressure solutions that are now slightly more correct. But then one undocumented restart-vector, PORV_RC that shows more significant deviations. I suggest these vectors are given due documentation in the manual, and that somebody familier with them can check at their convenience. My guess is that PORV_RC is very sensitive to pressure, so that a very small change (for the better) in the pressure solution gives significant difference in the PORV_RC vector. Lastly we have 3D_MSW which is particularly interesting given the problems with MSW-tests. Here RS has some deviations beyond what we typically see, but all else is very close to the reference.

totto82:
PORV_RC is the rock compaction adjusted pore volume. The number of failures are drastically reduced  with new defaults. The results here should be fine. I am a bit more worried about the restart and parallel failures. Since the drift compensation is not stored  this may result in different results. I don't know how well the drift compensation is tested in parallel. Maybe we need to split this PR in two. First try to get the CNV relaxed based on PV and than look at the drift compensation for restart and in parallel.

alfbr:
> I am a bit more worried about the restart and parallel failures.

Keep in mind that none of the two failures have any newton iteration counts exceeding eight. Hence, we are in effect relaxing tolerances slightly. In addition, simply upping the relative tolerance from 4e-05 to 5e-05 should have both pass. Not sure storing the drift compensation makes sense here, I would rather alter the tolerance.

alfbr:
I have increased the relative tolerances of the relevant restart and parallel test. Both are below 1e-4, so I believe these are acceptable changes. Let's see what jenkins says.

alfbr:

alfbr:
jenkins build this opm-tests=384 please

bska:
> jenkins build this opm-tests=384 please

Quick question here: Is this an experiment only or do you intend to merge this and the associate reference data updates in OPM/opm-tests#384?  In the latter case we need to coordinate this PR with OPM/opm-common#1882 since they both generate new reference data.  One will have to be merged first and then the other will have to generate a new set of reference data updates to avoid merge conflicts in the opm-tests repository.

alfbr:
> Quick question here: Is this an experiment only or do you intend to merge this and the associate reference data updates in [OPM/opm-tests#384](https://github.com/OPM/opm-tests/pull/384)?

Actually, I was not aware of the other one in flight, thanks for informing me. In any case, this turned out to be an experiment. Jenkins is not green. Restart test need tolerances relaxed further for these defaults. However, there is a failure related to comparing summary file with RSM file, has there been recent activity there that could explain it?

atgeirr:
There is also https://github.com/OPM/opm-simulators/pull/2800 which will require a data update,  OPM/opm-common#1882 should go in first.

alfbr:
Sure, this one is not green anyway, need to figure out why summary output and RSM file has different vectors in one case.

joakim-hove:
jenkins build this op-common=1958 opm-tests=384 please

joakim-hove:

bska:
jenkins build this opm-common=1958 please

joakim-hove:
jenkins build this opm-common=1958 please

akva2:
jenkins build this please

alfbr:
Guys, this needs a data update. And since it is not green, we need a new one for testing.

alfbr:

bska:
> Guys, this needs a data update. And since it is not green, we need a new one for testing.

Updated reference data is not going to help with the RSM vs. UNSMRY failure.  If you're using the new reference data in local testing then that's another matter of course.  We will nevertheless have to do yet another update to the reference data once we merge #2800 and OPM/opm-tests#391.

alfbr:
> > Guys, this needs a data update. And since it is not green, we need a new one for testing.
> 
> Updated reference data is not going to help with the RSM vs. UNSMRY failure.

I know, but there is  another failure to fix too, and I don't remember what it was :-)

alfbr:
jenkins build this opm-tests=384 please

bska:
jenkins build this opm-tests=384 please

alfbr:
I believe we need to wait for a fix to the RSM file checking code in compareECL.

joakim-hove:
jenkins build this opm-common=1965 opm-tests=384 please

joakim-hove:
I'll merge opm-common#1965 - then this should be mergeable

joakim-hove:
jenkins build this opm-tests=384 please

alfbr:
This is now ready for review.

bska:
> This is now ready for review.

It probably needs a new set of reference data update thought,  PR #2800 and its associate data update OPM/opm-tests#391 was merged in the interim and OPM/opm-tests#384 is no longer in a mergeable state.  I recommend closing OPM/opm-tests#384 and creating an entirely new set of reference solutions.

alfbr:

alfbr:
@atgeirr if you have any reservations to the changes here, please test on relevant models and report back.

bska:

bska:
jenkins build this opm-tests=392 please

bska:
jenkins build this opm-tests=392 please

atgeirr:
> @atgeirr if you have any reservations to the changes here, please test on relevant models and report back.

Thanks for asking. I have concluded that we should merge this. In terms of risking "bad solutions", the looser criteria for 8 iterations (compared to the old defaults) are probably less influential than the stricter criteria in later iterations. The risk of accepting a totally wrong solution should be reduced with these changes.

If we need to adjust these parameters in the future, we'll raise the discussion as needed.

PR #2607, opened 2020-05-10
Removed use of explicit regular falsi. Use the standard solver

hnil:
Removed two implementations of regular falsi in the same file.
Changed to use the standard solver also used in the wellmodel handeling.

PR #2791, opened 2020-09-17
increasing maxIter to 1000 for function satFromPc

GitPaean:
it is a temporary fix to make some parallel running easier when forgetting to change it manually.

It is a temporary fix before we can manage to merge https://github.com/OPM/opm-simulators/pull/2607 , but currently we do not have the capacity to do it.


GitPaean:
jenkins build this please

flow/flow.cpp
```diff
@@ -23,8 +23,26 @@
...
+#include <signal.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+void handler(int sig) {
+  void *array[10];
+  size_t size;
+  // get void*'s for all entries on the stack
+  size = backtrace(array, 10);
+  // print out all the frames to stderr
```
GitPaean, line 37: Oh, sorry, this is not intentional. This is some code to help to parallel debugging. I am removing them now.

GitPaean:
jenkins build this please

blattms:
Isn't the a calculation locale to one cell? Why should going parallel change such a computation? If it does it might point to  a bug.

I am not an expert but the regula falsi code looks a bit fishy, shouldn't the setting of `s1` and `f1` in lines [equilibrationhelpers.hh#L857-L858](OPM/opm-simulators/blob/775793676c0bee103ae876312ba3ae4fffa3b6e6/ebos/equil/equilibrationhelpers.hh#L857-L858) be part of the else branch?

GitPaean:
> Isn't the a calculation local to one cell?

I am not that familiar with this issue. But it is the issue we discussed during the past meeting.  With domain decomposition, each domain does not see the global field, which results in some different result during initialization.  It is a bug needs to be addressed. 

This PR and also #2607 are just trying to fix some urgent running issue. They do not address the problem mentioned above. 

#2607 is facing some challenges to get merged. 

This PR does not change anything for the cases that have been successfully running.  It also fixes some running that are not successful because with different processes, sometime the capillary equation needs more iterations to get converge.  Without increasing the maximum iterations, those running just throw and quit.

blattms:
> With domain decomposition, each domain does not see the global field, which results in some different result during initialization.  It is a bug needs to be addressed.

But increasing the maximum iterations here will not fix any faulty initialization code. We will need to find out what is not
correctly initialized and fix that.

GitPaean:
> But increasing the maximum iterations here will not fix any faulty initialization code. We will need to find out what is not
correctly initialized and fix that.

@bska and @hnil  have a good understanding of the issue.  

This PR is more for convenience for the current parallel testing and mostly harmless, and the change will be gone anyway. 

I intend to merge it sometime later today unless someone object.

blattms:
Well, in my experience such changes are not harmless (they might have side effects we do not envision yet) and tend to stay for a long time (often until someone stumbles over them when debugging). Maybe make parallel testing branches would be the better option?

GitPaean:
> Well, in my experience such changes are not harmless (they might have side effects we do not envision yet) 

Yes. You are probably correct. There might be side effects at certain point with very special circumstances, I believe the consequence is limited and not bigger than the current trouble. 

> Maybe making branches for parallel testing would be the better option?

Possibly. It still comes with something extra cost, and when you are developing multiple branches at the same time, it might bit you again and again. 

Sorry for not hearing your voice earlier. But I do believe the potential consequence will be rather limited and this fix will go away in the not long future, we probably will not be able to see it bites us at all.

PR #2794, opened 2020-09-20
Chase Group/Node Summary API Update

bska:
This is in preparation of adding support for outputting the network node pressure quantity, GPR, to the summary file.  In particular, `GroupValues` is renamed to `GroupAndNetworkValues` and has new individual datamembers for the former group-level data and the new node-level data.

Update `BlackoilWellModel::groupData()` and `CollectToIORank` accordingly and bring the parallel restart facility in line with the new layout.

Note this the companion API work that goes along with the upstream PR OPM/opm-common#1940


bska:
@akva2 : Would you mind having a look at the parallel restart/serialization part of this PR?

bska:
jenkins build this opm-common=1940 please

atgeirr:
Looks good! I did check the parallel parts to the best of my understanding, and will merge. If @akva2 wants to do an extra check, that is of course appreciated!

PR #2798, opened 2020-09-21
output temperature when TEMP is set

totto82:
For isothermal simulations (enableEnergy=false) but with non-constant temperature (RTEMPVD) we may want to output the temperature.


totto82:
jenkins build this please

joakim-hove:
jenkins build this please

totto82:
Can we merge this?

bska (approved):
This looks like a reasonable change.  I'll merge into master.

PR #2803, opened 2020-09-22
Compacted time step report

alfbr:
One small step in making output more consistent

### DIFF
--- a/opm/simulators/timestepping/SimulatorReport.cpp
+++ b/opm/simulators/timestepping/SimulatorReport.cpp
@@ -73,15 +73,14 @@ namespace Opm
 
     void SimulatorReportSingle::reportStep(std::ostringstream& ss) const
     {
-        ss << "Time step summary: ";
         if (total_well_iterations != 0) {
-            ss << "well its = " << std::setw(2) << total_well_iterations << ", ";
+            ss << "Well its=" << std::setw(2) << total_well_iterations << ", ";
         }
-        ss << "newton its = " << std::setw(2) << total_newton_iterations << ", "
-           << "linearizations = "  << std::setw(2) << total_linearizations
-           << " ("  << std::fixed << std::setprecision(3) << std::setw(6) << assemble_time << " sec), "
-           << "linear its = " << std::setw(3) << total_linear_iterations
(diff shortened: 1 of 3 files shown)


opm/simulators/timestepping/SimulatorReport.cpp
```diff
@@ -73,15 +73,14 @@ namespace Opm
 
     void SimulatorReportSingle::reportStep(std::ostringstream& ss) const
     {
-        ss << "Time step summary: ";
         if (total_well_iterations != 0) {
-            ss << "well its = " << std::setw(2) << total_well_iterations << ", ";
+            ss << "Well its=" << std::setw(2) << total_well_iterations << ", ";
```
atgeirr, line 77: I must admit that the missing spaces around `=` look ugly to me. Could you post a "before and after" example? I think setw(2) for various times might be too low for somewhat large cases, could we go with setw(3)? I assume the goal is to reduce the total width of the message. To that end I propose dropping the `linearizations` printout here (keeping it only on the final end-of-run printout). That would gain many characters!

alfbr:
Before:
```
Time step 0, stepsize 3 days, at day 299/316, date = 01-Sep-1998
Time step summary: well its =  1, newton its =  2, linearizations =  3 ( 0.912 sec), linear its =  41 ( 1.285 sec)
```
After:
```
Time step 1, stepsize 3 days, at day 1/8, date = 07-Nov-1997
Well its= 1, Newton its= 3, linearizations= 4 (0.3sec), linear its= 43 (0.3sec)

```

After discussion with some of my colleagues it was clear that this line in its completeness is not understood or read by ordinary users no matter what we do. Personally I really us the inforamtion, both timings of linearizations and timing of linear iterations. Removing the timing of the assembly will make the line significantly less informative to me. One option is to move the whole line to the PRT file. I will miss it, but I realize that that is of minor importance.

With respect to your wish of keeping spaces between "its" and "=", I disagree. But if you really feel strongly about it, I will put them back in. Also, please see what cases you have where setw(2) is insufficient. If it only happens very rarely, I would rather have the line a little longer when it happens.

alfbr:
Closed by #2832

PR #2805, opened 2020-09-23
Optional support in eclmpiserializer

akva2:
This adds support for std::optional in the eclmpiserializer class.
This allows for optionals with "complex" types, ie types that themself should be serialized through a serializeOp member.


joakim-hove:
jenkins build this please

ebos/eclmpiserializer.hh
```diff
@@ -194,6 +197,45 @@ public:
 
     }
 
+    //! \brief Handler for std::optional.
+    //! \tparam T Type for data
+    //! \param data The optional to (de-)serialize
+    template<class T>
+    void optional(const std::optional<T>& data)
+    {
+        if (m_op == Operation::PACKSIZE) {
```
joakim-hove, line 206: There is support for similar operations for *non-complex* types already in Parallellrestart.cpp - will that become a conflict?

akva2: it will not be in conflict. they are simply not reachable now, so the optional support can in principle be removed from there now.

PR #2806, opened 2020-09-23
Drop memory used by transmissibility calculator after use

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/1953

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -514,9 +514,9 @@ private:
 
     void updateFromEclState_(bool global)
     {
-        const FieldPropsManager* fp =
-            (global) ? &(vanguard_.eclState().fieldProps()) :
-            &(vanguard_.eclState().globalFieldProps());
+        FieldPropsManager* fp =
+            const_cast<FieldPropsManager*>((global) ? &(vanguard_.eclState().fieldProps()) :
+                                           &(vanguard_.eclState().globalFieldProps()));
 
         std::array<bool,3> is_tran {fp->tran_active("TRANX"),
                                     fp->tran_active("TRANY"),
@@ -535,8 +535,10 @@ private:
 
         for (auto it = trans.begin(); it != trans.end(); ++it, ++key, ++perform)
         {
-            if(perform)
+            if(perform) {
                 fp->apply_tran(*key, *it);
+                fp->drop_tran(*key);
+            }
         }
 
         resetTransmissibilityFromArrays(is_tran, trans);


joakim-hove:
jenkins build this opm-common=1952 please

joakim-hove:
jenkins build this opm-common=1953 please

ebos/ecltransmissibility.hh
```diff
@@ -535,8 +535,10 @@ private:
 
         for (auto it = trans.begin(); it != trans.end(); ++it, ++key, ++perform)
         {
-            if(perform)
+            if(perform) {
                 fp->apply_tran(*key, *it);
+                fp->drop_tran(*key);
```
blattms, line 540: This is run twice. First before loadbalancing and then after. Hence this will clear the  transmissibilities before the loadbalance happens and in parallel we will not see them. Also on rank 0 there are to transmissibility calculators to be cleared.

blattms (changes requested):
I think tests will fail or deadlock.

ebos/ecltransmissibility.hh
```diff
@@ -514,9 +514,9 @@ private:
 
     void updateFromEclState_(bool global)
     {
-        const FieldPropsManager* fp =
-            (global) ? &(vanguard_.eclState().fieldProps()) :
-            &(vanguard_.eclState().globalFieldProps());
+        FieldPropsManager* fp =
+            const_cast<FieldPropsManager*>((global) ? &(vanguard_.eclState().fieldProps()) :
```
blattms, line 518: Hacky.

ebos/ecltransmissibility.hh
joakim-hove: > This is run twice. First before loadbalancing and then after.

OK - that was important information - I did not know. Would you like to suggest a better callsite or should I dig in.

joakim-hove:
> I think tests will fail or deadlock.

For all it is worth the tests seem to go through on my machine (with the normal MSW exceptions); but that might be due to insufficient test data.

blattms:
Well those that were waiting to be put into a PR, which I forgot posting. #2807 hopefully will make something failing.

blattms:
There is no better call site for the hack. But you need to wait for the second
pass (where the boolean is false). and then cleanup both versions (global and
distributed).

If not done already you also need to override the clean up method in
the ParallelFieldPropsManager (not sure whether that is the correct name).

I still think that this change in semantics is dangerous and will cause
additional work in the future.

joakim-hove:
> I still think that this change in semantics is dangerous and will cause
additional work in the future.

I agree - I'll drop the whole thing.

PR #2807, opened 2020-09-23
Run model2/5 tests (TRAN in EDIT)

blattms:


blattms:
Is it safe to run `jenkins build this update_data please` now?

akva2:
You have to update tests/update_reference_data.sh

blattms:
Done, but only once (compareEcl stuff). Do I also need to add something for compare_parallel* stuff for the same cases?

akva2:
No, compareParallel compares results from a serial run to a parallel one. No reference data involved.

akva2:

compareECLFiles.cmake
```diff
@@ -725,6 +741,13 @@ add_test_compareECLFiles(CASENAME model4_udq_group
                          REL_TOL ${rel_tol}
                          DIR model4)
 
+add_test_compareECLFiles(CASENAME maxvalue_in_edit
+                         3_MAXVALUE_TRANZ
```
akva2, line 745: missing FILENAME

compareECLFiles.cmake
```diff
@@ -895,4 +918,19 @@ if(MPI_FOUND)
...
+add_test_compare_parallel_simulation(CASENAME tran_in_edit
+                                     FILENAME 10_TRAN_IN_EDIT
+                                     SIMULATOR flow
+                                     ABS_TOL ${abs_tol}
+                                     REL_TOL ${rel_tol}
+                                     DIR model2)
+
+
+add_test_compare_parallel_simulation(CASENAME maxvalue_in_edit
+                                     3_MAXVALUE_TRANZ
```
akva2, line 931: missing FILENAME

blattms:
Fixed stuff and collapsed commits.

blattms:

blattms:
Can someone look at opm-tests/model5/include/summary.inc I am getting now:
```
Error: Error in keyword "GPR": No such network node: B1[NON-XML-CHAR-0x1B][0m
[NON-XML-CHAR-0x1B][31;1m
Error: Unrecoverable errors were encountered while loading input: SUMMARY_UNKNOWN_NODE: Error in keyword "GPR": No such network node: B1[NON-XML-CHAR-0x1B][0m
```
This error seems new. Probably due to /OPM/opm-common#1944 from @bska.

bska:
> Can someone look at `opm-tests/model5/include/summary.inc` I am getting now:
> ```
> Error in keyword "GPR": No such network node: B1
> ```
> This error seems new. Probably due to /OPM/opm-common#1944 

That is correct.  I added support for writing network node pressure (`GPR`) out to the summary file in OPM/opm-common#1944 and one of the checks I added was to [not configure output for missing network nodes](https://github.com/OPM/opm-common/blob/b9f95089b97ee8ee04396d2416c4bb236a9343dc/src/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp#L493-L499).  In your example, it is the [`handleMissingNode()`](https://github.com/OPM/opm-common/blob/b9f95089b97ee8ee04396d2416c4bb236a9343dc/src/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp#L345-L352) which generates the diagnostic message you're seeing.

As for the underlying problem, your test model uses [`include/summary.inc`](https://github.com/OPM/opm-tests/blob/9666371a5ce48cfbaeffdf1411a5bc9583e17a8b/model5/3_MAXVALUE_TRANZ.DATA#L188-L189) and that, in turn, requests that [`GPR`](https://github.com/OPM/opm-tests/blob/9666371a5ce48cfbaeffdf1411a5bc9583e17a8b/model5/include/summary.inc#L260-L261) be generated for the network nodes `B1`, `C1`, and `PLAT-A`.

I'm not really sure what the right answer here is.  On the one hand I'd like to keep the current behaviour of loudly notifying the user that they've requested summary output which cannot be generated, but on the other hand maybe it should be a warning rather than an error.

I'd very much like to solicit opinions on this, e.g., from yourself, @joakim-hove, @atgeirr, and @alfbr.

blattms:
I actually don't have a take on this one other than that I need to make my test run.
Unfortunately summary.inc is include in all the datafiles of model5 and hence I do
not feel comfortable changing anything. It needs a proper solution by somebody else.

bska:
> Unfortunately summary.inc is include in all the datafiles of model5 and hence I do not feel comfortable changing anything.

I suppose one **possible** approach would be to extract summary keywords unrelated to the network model into a new include file e.g., `include/summary_nonetwork.inc`, and then write the current `summary.inc` as
```
INCLUDE
   'include/summary_nonetwork.inc' /

GPR
  'B1' 'C1' 'PLAT-A' /
```
Then you could use the new `summary_nonetwork.inc` in your model.

I'm certainly open to other suggestions.

atgeirr:
I think making the error a warning instead would be good. Typos etc in the SUMMARY sections are serious enough to tell the user about, but not enough to terminate. The warning will then not be emitted any more once we merge the network option code. (Unless users add nonsensical node names etc.)

bska:
> The warning will then not be emitted any more once we merge the network option code.

I think you'll still get the warning if you request `GPR` in a model which does not use the network option, although I **suppose** we could make the warning conditional on the network option itself and then completely ignore it.  I don't know how hard that would be to implement.

atgeirr:
> you'll still get the warning if you request `GPR` in a model which does not use the network option

That would be fine I think.

joakim-hove:
@blattms: My suggestion would be to set this switch: https://github.com/OPM/opm-common/blob/master/src/opm/parser/eclipse/Parser/ParseContext.cpp#L99 to `InputError::WARN` - at least as a first shot.

bska:
> My suggestion would be to set [this switch](https://github.com/OPM/opm-common/blob/b9f95089b97ee8ee04396d2416c4bb236a9343dc/src/opm/parser/eclipse/Parser/ParseContext.cpp#L99) to `InputError::WARN`&mdash;at least as a first shot.

You may additionally want to add the following patch to the [`keyword_node()`](https://github.com/OPM/opm-common/blob/b9f95089b97ee8ee04396d2416c4bb236a9343dc/src/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp#L469-L474) processor function.  The [default action](https://github.com/OPM/opm-common/blob/b9f95089b97ee8ee04396d2416c4bb236a9343dc/src/opm/parser/eclipse/Parser/ParseContext.cpp#L100) of `SUMMARY_UNHANDLED_KEYWORD` is `InputError::WARN`.

```diff
index d9e1a459e..4a539c658 100644
--- a/src/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
+++ b/src/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp
@@ -472,6 +472,16 @@ void keyword_node( SummaryConfig::keyword_list& list,
                    ErrorGuard& errors,
                    const DeckKeyword& keyword)
 {
+    if (node_names.empty()) {
+        const auto& location = keyword.location();
+        const std::string msg = "The network node keyword " + keyword.name()
+            + " at " + location.filename + ", line " + std::to_string(location.lineno)
+            + " is not supported in runs without networks";
+
+        parseContext.handleError( ParseContext::SUMMARY_UNHANDLED_KEYWORD, msg, errors);
+        return;
+    }
+
     auto param = SummaryConfigNode {
         keyword.name(), SummaryConfigNode::Category::Node, keyword.location()
     }
```

blattms:
jenkins build this update_data opm-common=1970 please

blattms:
I guess having a test for MAX_VALUE based on a model that is not completely tested yet, is a bit unfortunate. I only want to compare transmissibilities. I guess I need to change the test.

blattms:
@akva2 how create a data update locally, to keep some stress from jenkins?

alfbr:
> @akva2 how doI create a data update locally, to keep some stress from jenkins?

Just run Flow from master on all test-cases in opm-tests, and set theoutput directory to overwrite the reference solutions.

blattms:
Needs OPM/opm-tests#396 now.

blattms:
jenkins build this update_data opm-tests=396 please

blattms:
jenkins build this update_data opm-tests=396 please

blattms:
I am giving up. I can hardly cope with my test failing in parallel on jenkins while it works on my machine. But failing tests that I did not even touch are a bit too much.

blattms:
probably shouldn't have rebased this.

PR #2808, opened 2020-09-24
fixed: avoid deadlocks in parallel if serialization throws

akva2:
This happened for instance if serialization was attempted for a type that was not supported.
This then led to jobs hanging on jenkins, until timeout.


akva2:
jenkins build this please

blattms (approved):
Looks good.

May I ask where the exceptions occured?

akva2:
approved and green.

PR #2810, opened 2020-09-25
Pass location to ParseContext.handleError()

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/1946

### DIFF
--- a/opm/simulators/flow/MissingFeatures.cpp
+++ b/opm/simulators/flow/MissingFeatures.cpp
@@ -16,6 +16,7 @@
   You should have received a copy of the GNU General Public License
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
+#include <fmt/format.h>
 
 #include <opm/simulators/flow/MissingFeatures.hpp>
 
@@ -64,10 +65,10 @@ namespace MissingFeatures {
                 else
                     val = it->second.item_value;
 
-                std::string msg = "For keyword '" + it->first + "' only value " + val
-                    + " in item " + it->second.item + " is supported by flow.\n"
-                    + "In file " + location.filename + ", line " + std::to_string(location.lineno) + "\n";
-                parseContext.handleError(ParseContext::SIMULATOR_KEYWORD_ITEM_NOT_SUPPORTED, msg, errorGuard);
+                std::string msg_fmt = fmt::format("Unsupported value for {{keyword}}\n"
+                                                  "In {{file}} line {{line}}\n"
+                                                  "Only the value {} in item {} of {{keyword}} is supported by flow", val, it->second.item);
+                parseContext.handleError(ParseContext::SIMULATOR_KEYWORD_ITEM_NOT_SUPPORTED, msg_fmt, keyword.location(), errorGuard);
             }
         }
     }
@@ -882,9 +883,9 @@ namespace MissingFeatures {
             std::unordered_set<std::string>::const_iterator it;
             it = unsupported_keywords.find(keyword.name());
             if (it != unsupported_keywords.end()) {
-                std::string msg = "Keyword '" + keyword.name() + "' is not supported by flow.\n"
-                    + "In file " + location.filename + ", line " + std::to_string(location.lineno) + "\n";
-                parseContext.handleError(ParseContext::SIMULATOR_KEYWORD_NOT_SUPPORTED, msg, errorGuard);
+                std::string msg_fmt = "Keyword {keyword} is not supported by flow.\n"
+                                      "In {file} line {line}.";
+                parseContext.handleError(ParseContext::SIMULATOR_KEYWORD_NOT_SUPPORTED, msg_fmt, keyword.location(), errorGuard);
             }
             checkOptions<std::string>(keyword, string_options, parseContext, errorGuard);
             checkOptions<int>(keyword, int_options, parseContext, errorGuard);


opm/simulators/flow/Main.hpp
```diff
@@ -433,18 +433,10 @@ namespace Opm
...
-                    OpmLog::info("Reading deck file '" + deckFilename + "'");
-                }
-
                 readDeck(mpiRank, deckFilename, deck_, eclipseState_, schedule_,
                          summaryConfig_, nullptr, python, std::move(parseContext),
                          init_from_restart_file, outputCout_);
 
-                if (outputCout_) {
-                    OpmLog::info("Done reading deck file.");
-                }
```
bska, line 446: Actually, this message is really useful.  It was added in #2760 with express intention that users get some idea of what part of the setup process is going on.

joakim-hove: OK; we are in the process of adding more informative messages which will give essentially the same information. Can of course let this message remain in the interim.

bska: > we are in the process of adding more informative messages which will give essentially the same information

Right, I got that impression from other work being submitted at the moment.

> Can of course let this message remain in the interim.

I guess that depends on the timing.  If the new system is expected to be in place in a few days then there's no need.  Otherwise I'd prefer to keep this message.

joakim-hove: I have resurrected the message

alfbr: Actually, I beg to differ. The message is wrong. The end of parsing is already reported where it actually happens. Keeping this message has as such a negative value. I totally agree that the user needs a fine grained control of what happens after parsing, but this message is not helping that.

atgeirr: > The end of parsing is already reported where it actually happens

I did not think so. Where?

alfbr: All lines parsed are reported, including the last. For Norne it looks like this:
```
 1380 Reading WCONINJE in /home/rustad/Dev/Data/opm-data/norne/INCLUDE/BC0407_HIST01122006.SCH line 9554
 1381 Reading DATES    in /home/rustad/Dev/Data/opm-data/norne/INCLUDE/BC0407_HIST01122006.SCH line 9563

Warning: Unsupported value for ENDSCALE
```

atgeirr: If you know that is the last line, then you are good. But most users do not know that, and if something goes wrong after this point (but before timesteps start) they could blame the deck reader/parser wrongly. That actually happened, which is why we added the extra (user-focused) information that we are done with it now.

alfbr: Seriously, it is always the last line, by design. Just run Flow on some cases and you immediately see that. If something goes wrong after this point we need as I already stated (and which has been discussed and agreed by both @joakim-hove and @bska ) fine grained control on what happens. Basically we need logging of where between the parsing and time step start we are. I repeat, the message in question is not helping that, it is simply wrong.

atgeirr: > Basically we need logging of where between the parsing and time step start we are.

I do not disagree.

> I repeat, the message in question is not helping that, it is simply wrong.

This is the code snippet:
```
                readDeck(mpiRank, deckFilename, deck_, eclipseState_, schedule_,
                         summaryConfig_, nullptr, python, std::move(parseContext),
                         init_from_restart_file, outputCout_);

                if (outputCout_) {
                    OpmLog::info("Done reading deck file.");
                }
```
If this is wrong, then what is perhaps wrong is that readDeck() does more than just reading the deck, it also creates the EclipseState and calls some checks. Is that what you mean? If so, we agree!  I now believe the printout should be moved into readDeck(), after the deck is constructed but before the checks are called (line 192 of readDeck.cpp).

In general, I find that a little bit of information to the user about what phase in the process we have reached is good. Do we agree on that?

atgeirr: > Seriously, it is always the last line, by design. Just run Flow on some cases and you immediately see that

I do not understand what you mean by this?

bska: > Just run Flow on some cases and you immediately see that. If something goes wrong after this point we need

The one thing we **don't** detect if we remove this message is if something goes wrong **before** this point&mdash;e.g., we're reading from a network disk and the controller died meaning the process then just hangs with no additional clue to the problem.  It **could be** a problem constructing the parser output object (`EclipseState`, `EclipseGrid`, `Schedule`, or `SummaryConfig`) or just read failure.

I suppose we could output one final record of the form
```
1234 END (Implied)
```
when we actually get to the end of the input file.  If the input happens to contain an `END` keyword then we don't output that and don't output the "implied" end record.

alfbr: > If this is wrong, then what is perhaps wrong is that readDeck() does more than just reading the deck, it also creates the EclipseState and calls some checks. Is that what you mean? If so, we agree!

Yes, this is exactly what I mean. We need the EclipseState to report what it does when it does it, so we can have a clue as to what goes wrong when it crashes.

> 1234 END (Implied)

That is a possibility, but I am afraid we are solving a problem that does not exist. The user always has the deck available, and will check the deck if Flow crashes during parsing. That was exactly why we needed that keyword-by-keyword reporting, and afaik it does the job.

opm/simulators/flow/Main.hpp
```diff
@@ -433,10 +433,6 @@ namespace Opm
 
                 Opm::FlowMainEbos<PreTypeTag>::printPRTHeader(outputCout_);
 
-                if (outputCout_) {
-                    OpmLog::info("Reading deck file '" + deckFilename + "'");
```
atgeirr, line 437: I now agree with removing this printout from this spot, but would like it to be added at readDeck.cpp line 192 (before the checkKeywords() call).

opm/simulators/utils/readDeck.cpp
```diff
@@ -248,21 +253,19 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
...
 
 #endif
 
     if (*errorGuard) { // errors encountered
         parseSuccess = 0;
+        errorGuard->clear();
     }
 
-    // print errors and warnings!
-    errorGuard->dump();
```
atgeirr, line 264: Since this is no longer dumped, I assume all messages will be printed as-we-go instead of here?

atgeirr (commented):
Looks good, just one minor request and a question from me.

opm/simulators/utils/readDeck.cpp
joakim-hove: The exception part of this PR as been extracted as #2813. See [this comment](https://github.com/OPM/opm-simulators/pull/2813#issuecomment-699825031)there.

PR #2811, opened 2020-09-25
changed: reuse objects if building python inline

akva2:
This avoids rebuilding the objects for the python binding if the binding is built inline with the main buildsystem.

Reduces build time on jenkins a little.

### DIFF
--- a/python/simulators/CMakeLists.txt
+++ b/python/simulators/CMakeLists.txt
@@ -4,8 +4,8 @@ set_target_properties( simulators PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${PROJECT_
 
 target_sources(simulators
   PRIVATE
-    ../../opm/simulators/utils/moduleVersion.cpp
-    ../../flow/flow_ebos_blackoil.cpp)
+  $<TARGET_OBJECTS:moduleVersion>
+  $<TARGET_OBJECTS:flow_libblackoil>)
 
 target_link_libraries( simulators PRIVATE opmsimulators )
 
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this please

python/simulators/CMakeLists.txt
```diff
@@ -2,10 +2,17 @@ pybind11_add_module(simulators simulators.cpp SYSTEM)
 
 set_target_properties( simulators PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/python/opm2 )
 
-target_sources(simulators
-  PRIVATE
+if(TARGET flow_libblackoil)
```
blattms, line 5: I am probably missing something, but it seems like the creation of the target is uncondtional. Hence, is the else branch ever evaluated?

blattms (changes requested):
Apart from the commented part this looks good.

python/simulators/CMakeLists.txt
akva2: Well, you are not wrong as such. But I want to keep the option of building it standalone open. There are however other reasons why that does not work currently, so maybe I'm overthinking it.

akva2: What I mean is that it's currently possible to make a CMakeLists.txt which imports an opmsimulators target from the system, then includes this CMakeLists.txt. That would allow you to build the python bindings separate from opmsimulators lib. But I guess it's not worth it, so I can remove the if'ery.

akva2:
jenkins build this please

akva2:
please hit the green if you are satisfied @blattms

blattms (approved):

PR #2813, opened 2020-09-25
Flow exceptions

joakim-hove:
This is part of the effort to improve error messages. A pattern we have had in flow is that the error message is displayed more than one time - typically with different formatting for a nice visual effect. While working on this I try to enforce the following principle:

1. At the point where we have all relevant context - that might be right when the error is encountered or a couple of frames up - we issue a call to `OpmLog::error()` and throw an exception.

2. If/when we catch the exception at a later stage **we trust that the error message has already reached the user with the previous call to `OpmLog::error()`, and do not output `exc.what()` in any way**. 

Th gist of this PR is that the error messages (i.e. `exception.what()`) is not output again in the `readDeck()` code - we trust that the relevant opm-common code has already logged the error.

### DIFF
--- a/opm/simulators/utils/readDeck.cpp
+++ b/opm/simulators/utils/readDeck.cpp
@@ -42,6 +42,8 @@
 #include <opm/simulators/utils/ParallelEclipseState.hpp>
 #include <opm/simulators/utils/ParallelSerialization.hpp>
 
+#include <fmt/format.h>
+
 #include <cstdlib>
 
 namespace Opm
@@ -54,7 +56,7 @@ void ensureOutputDirExists_(const std::string& cmdline_output_dir)
             Opm::filesystem::create_directories(cmdline_output_dir);
         }
         catch (...) {
-            throw std::runtime_error("Creation of output directory '" + cmdline_output_dir + "' failed\n");
+            throw std::runtime_error(fmt::format("Creation of output directory '{}' failed\n", cmdline_output_dir));
         }
     }
 }
--- a/opm/simulators/utils/readDeck.cpp
+++ b/opm/simulators/utils/readDeck.cpp
@@ -258,6 +258,8 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
     catch(const std::exception& broadcast_error)
     {
         failureMessage = broadcast_error.what();
+        OpmLog::error(fmt::format("Distributing properties to all processes failed\n"
+                                  "Internal error message: {}", broadcast_error.what()));
         parseSuccess = 0;
     }
 
@@ -265,11 +267,10 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
 
     if (*errorGuard) { // errors encountered
         parseSuccess = 0;
+        errorGuard->dump();
+        errorGuard->clear();
     }
 
-    // print errors and warnings!
-    errorGuard->dump();
-    errorGuard->clear();
 
     auto comm = Dune::MPIHelper::getCollectiveCommunication();
     parseSuccess = comm.min(parseSuccess);
@@ -278,7 +279,7 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
     {
         if (rank == 0)
         {
-            OpmLog::error(std::string("Unrecoverable errors were encountered while loading input: ")+failureMessage);
+            OpmLog::error("Unrecoverable errors were encountered while loading input");
         }
 #if HAVE_MPI
         MPI_Finalize();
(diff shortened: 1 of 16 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
A message to stdout has received quite some attention in #2810, but the some changes in exception/message handling which are also part of #2810 - and extracted as a separate PR here - is maybe more important? Cool if you take a look at this.

opm/simulators/utils/readDeck.cpp
```diff
@@ -248,21 +253,19 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
...
 
 #endif
 
     if (*errorGuard) { // errors encountered
         parseSuccess = 0;
+        errorGuard->clear();
     }
 
-    // print errors and warnings!
-    errorGuard->dump();
```
blattms, line 264: Are these guaranteed to be printed elsewhere?

opm/simulators/utils/readDeck.cpp
```diff
@@ -228,9 +229,13 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
...
         }
-        catch(const std::exception& e)
+        catch(const OpmInputError& input_error) {
+            failureMessage = input_error.what();
+            parseSuccess = 0;
+        }
+        catch(const std::exception& std_error)
         {
-            failureMessage = e.what();
+            failureMessage = std_error.what();
```
blattms, line 238: I think this is an unexpected error (maybe programming error), but failureMessage will never be printed.

blattms (commented):
I cannot really judge this. But are you sure that all warnings and errors will be printed. Seems like messages from exceptions apart from OpmInputError will not printed.

Maybe it is safer risking dulicate printing?

joakim-hove:
> But are you sure that all warnings and errors will be printed.

There is no hard guarantee - as in that the compiler guarantees it; the only guarantee is what we provide/guarantee through the implementation/convention.

opm/simulators/utils/readDeck.cpp
joakim-hove: As mentioned - there is no hard guarantee no; but the intention is that the `ParseContext::handleError()` will print these messages - so with the current implementation - yes we are guaranteed that the messages has been printed already.

opm/simulators/utils/readDeck.cpp
joakim-hove: Yes - currently you are right; the `Deck`is created like this:
```
try {
     // parse file; might fail with either std::exception or OpmInputError
} catch (const OpmInputError& opm_error) {
     throw;
} catch (const std::exception& std_error) {
     std::string msg_fmt = fmt::format("Problem with keyword {{keyword}}\n"
                                       "In {{file}} line {{line}}\n",
                                       "Internal error message: {}" , std_error.what());
     OpmLog::error(OpmInputError::format( msg_fmt, keyword.location() );
     throw;
}
``` 
this construction "guarantees" that the error message from the `std::exception` has been printed - with accompanying location information. The intention is to add a similar construction for the construction of `EclipseState, Schedule` and `SummaryConfig` - but you are right; before that has been done we will not see the messages from the construction of `EclipseState, Schedule` and `SummaryConfig`.

I will create a opm-common PR ensuring that.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this opm-common=1967 please

blattms:
The problem is that the last thing the user will see is: "Unrecoverable errors were encountered while loading input". If we are lucky averag Joe will send an error error "Flow does not work, got 'Unrecoverable errors were encountered while loading input'". That report has zero valuable information for the developers, they need to ask back, user might need to run again, ....

Personally, I think it is very useful, if the last outputs of a failing program contain some valuable information about what has failed.

joakim-hove:
> The problem is that the last thing the user will see is: "Unrecoverable errors were encountered while loading input"

Well - that is not entirely true, in "normal failures"[1] the user will see:
```
<red>
Error: Problem handling keyword EQUIL
In /path/to/file/equil.inc line 123
Internal error: Invalid index
</red>

Unrecoverable errors were encountered while loading input
```
I.e. the actual error message will be just before the "Unrecoverable errors ....".

[1]: The exception to this is if the user has requested/used the mode where flow will stubbornly continue as far as possible, reporting errors while going and then finally exiting. Personally I think this is a quite good mode, but we have not really found how to communicate the possibility to the user. However - in this case I agree it makes sense to play through the error messages - I will reinstatiate the `error.dump()` call.

joakim-hove:
jenkins build this please

alfbr:
> Personally, I think it is very useful, if the last outputs of a failing program contain some valuable information about what has failed.

I agree, but I think what is presented here is already miles above what we have. The solution here with logging as we create the different states/configs is what we agreed as the first step. Hence, I suggest this goes in as is, and that we all make sure to bring in improvements on the actual messages in the next round. They need more improvements than what is natural to do in this PR.

joakim-hove:
jenkins build this please

PR #2815, opened 2020-09-25
Remove summary keywords from list of unsupported keywords

joakim-hove:
See comments [here](https://github.com/OPM/opm-common/pull/1951#discussion_r494817955) - might be other summary keywords which should be removed from this list.


joakim-hove:
jenkins build this please

alfbr:
It would be good if somebody with a bit of ownership to this code in opm-simulators can chime in and voice if they have concerns about handling of missing summary keywords being moved to opm-common.

From my end, I have no strong opinions on it. But I do not think this should be merged until we have a well-formatted solution in opm-common (hopefully we can get that in place quickly). Until that is in place, I would rather prefer that missing summary is presented by opm-simulators, not the current solution in opm-common.

PR #2816, opened 2020-09-25
New standard well implementation for the OpenCL backend

ducbueno:
New implementation of the addition of the standard wells contributions in the OpenCL backend.

This implementation pretty much takes care of the time step chopping that was supposed to get fixed in https://github.com/OPM/opm-simulators/pull/2774.

The multisegmented wells contributions still need to be put back into the code. As soon as I get it running I'll push it.

There is still an issue when running Norne: flow throws an exception in the middle of the simulation only when CUDA is also present in the system. By setting the simulation verbosity to 4 the exception seems to be thrown by the preconditioner.


opm/simulators/linalg/bda/openclKernels.hpp
```diff
@@ -366,96 +366,78 @@ namespace bda
...
-        int stride = valsPerBlock;
-        if(wiId < stride){
-            localSum[wiId] += localSum[wiId + stride];
-        }
+            b = wiId/valsPerBlock + val_pointers[wgId];
 
-        barrier(CLK_LOCAL_MEM_FENCE);
+            if(wiId < valsPerBlock){
+                if(c == 0 || c == 2) {localSum[wiId] += localSum[wiId + 2];}
+                if(c == 0 || c == 1) {localSum[wiId] += localSum[wiId + 1];}
```
blattms, line 413: Are we assuming dim<=3? In that case we need to throw somewhere if this does not hold (we might have additional polymer or other).

Also, I do not quite understand how this works. Why do we process `c == 1` in the second if-clause? Doesn't that mean that we are reading (if `c == 0` we read from w+1 for `wid == w`) and writing (`c == 1` holds for `wid = w+1`) for/to the same location? Probably I am missing something here.

opm/simulators/linalg/bda/openclKernels.hpp
```diff
@@ -366,96 +366,78 @@ namespace bda
...
+                                __global const int *Bcols,
+                                __global const double *x,
+                                __global double *y,
+                                __global const int *toOrder,
+                                const unsigned int dim,
+                                const unsigned int dim_wells,
+                                __global const unsigned int *val_pointers,
+                                __local double *localSum,
+                                __local double *z1,
+                                __local double *z2){
```
blattms, line 383: For clarity please document that z1 and z2 might point to the same location.

opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
```diff
@@ -0,0 +1,113 @@
...
+            queue->enqueueWriteBuffer(s.val_pointers, CL_TRUE, 0, sizeof(unsigned int) * wellContribs.h_val_pointers_ocl.size(), wellContribs.h_val_pointers_ocl.data());
+            queue->enqueueWriteBuffer(s.toOrder, CL_TRUE, 0, sizeof(int) * toOrder.size(), toOrder.data(), nullptr, &event);
+            event.wait();
+        }
+    }
+
+    void WellContributionsOCLContainer::update_on_gpu(Opm::WellContributions &wellContribs){
+        if(num_std_wells > 0){
+            cl::Event event;
+            queue->enqueueWriteBuffer(s.Cnnzs, CL_TRUE, 0, sizeof(double) * wellContribs.h_Cnnzs_ocl.size(), wellContribs.h_Cnnzs_ocl.data());
```
blattms, line 74: Won't work in general. We only treat open wells and the number of these might change. Even for each wellsthe number of cells perforated might change (different sparsity pattern). Hence we cannot reuse these buffers as their size might be inadequate and the column information wrong.

blattms (changes requested):
Thanks, I think this is a big step in the right directions. Just a few things need to be fixed.

Did not run this yet.

Can you comment on the exception?

opm/simulators/linalg/bda/openclKernels.hpp
blattms: in the case the we are reading values written by different threads, we might also need synchronization/barriers in-between.

opm/simulators/linalg/bda/openclKernels.hpp
```diff
@@ -366,96 +366,78 @@ namespace bda
...
+            while(b < valSize + val_pointers[wgId]){
+                int colIdx = toOrder[Bcols[b]];
+                localSum[wiId] += Bnnzs[b*dim*dim_wells + r*dim + c]*x[colIdx*dim + c];
                 b += numBlocksPerWarp;
             }
-        }
 
-        barrier(CLK_LOCAL_MEM_FENCE);
+            if(wiId < valsPerBlock){
+                localSum[wiId] += localSum[wiId + valsPerBlock];
```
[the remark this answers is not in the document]
ducbueno: Yes, there is an assumption that `dim` and `dim_wells` will always be 3 and 4, respectively. There is no exception that catches if they are different. I'll look into this.

opm/simulators/linalg/bda/openclKernels.hpp
ducbueno: We assume `dim == 3`. We process `c == 1` because we want to add 3 consecutive values of the `localSum`vector. I've written it like this because it is written in a similar manner in the CUDA kernel and I wanted to test if something was wrong with my previous implementation. I think the previous implementation was correct. I'll run some tests to see if this holds and if the other implementation is faster.

opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
```diff
@@ -0,0 +1,113 @@
...
+            toOrder.insert(toOrder.end(), wellContribs.toOrder, wellContribs.toOrder + Nb);
+
+            cl::Event event;
+            queue->enqueueWriteBuffer(s.Cnnzs, CL_TRUE, 0, sizeof(double) * wellContribs.h_Cnnzs_ocl.size(), wellContribs.h_Cnnzs_ocl.data());
+            queue->enqueueWriteBuffer(s.Dnnzs, CL_TRUE, 0, sizeof(double) * wellContribs.h_Dnnzs_ocl.size(), wellContribs.h_Dnnzs_ocl.data());
+            queue->enqueueWriteBuffer(s.Bnnzs, CL_TRUE, 0, sizeof(double) * wellContribs.h_Bnnzs_ocl.size(), wellContribs.h_Bnnzs_ocl.data());
+            queue->enqueueWriteBuffer(s.Ccols, CL_TRUE, 0, sizeof(int) * wellContribs.h_Ccols_ocl.size(), wellContribs.h_Ccols_ocl.data());
+            queue->enqueueWriteBuffer(s.Bcols, CL_TRUE, 0, sizeof(int) * wellContribs.h_Bcols_ocl.size(), wellContribs.h_Bcols_ocl.data());
+            queue->enqueueWriteBuffer(s.val_pointers, CL_TRUE, 0, sizeof(unsigned int) * wellContribs.h_val_pointers_ocl.size(), wellContribs.h_val_pointers_ocl.data());
+            queue->enqueueWriteBuffer(s.toOrder, CL_TRUE, 0, sizeof(int) * toOrder.size(), toOrder.data(), nullptr, &event);
```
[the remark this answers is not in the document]
blattms: Sorry, to bother you again.

Actually the new changes introduce to many synchronization points. As we do not reuse the host buffers, it should be perfectly fine to use `CL_FALSE` to request non-blocking. How about setting up a `std::vector<cl::Event>` and pass one entry to each `enqueueWriterBuffer` and at the end use [`cl::Event::waitForEvents`](http://github.khronos.org/OpenCL-CLHPP/classcl_1_1_event.html#ae84b90e930f72c1019435dd7cefa2fbd) to wait on all of those.

opm/simulators/linalg/bda/openclKernels.hpp
blattms: Maybe the CUDA code is wrong, too. Or I am (as often) seeing ghosts. This is what I think that we are computing (assuming that in "step 1" we use the original localSum[4], but that value is changed at the same time by the other thread!)

| c | 0 | 1 | 2 | 0 | 1 | 2 |
|---|---|---|---|----|---|----|
| wid | 3 | 4 | 5 | 6 | 7 | 8 |
|---|---|---|---|----|---|----|
| step 0 | localSum[3]+localSum[5] | | localSum[5] + localSum[7] |
| step 1 | localSum[3]+localSum[5]+ localSum[4] | localSum[4]+localSum[5] + localSum[7] |

If that assumption holds we seem to compute the right value for c==0, but the value of c==1 is (hopefully) not used. Where is the flaw in my thinking?

If the same assumptions on dim and dimw are done in the CUDA, please leave changing these for another PR.

opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
```diff
@@ -0,0 +1,131 @@
...
+            event.wait();
+            queue->enqueueWriteBuffer(s.toOrder, CL_TRUE, 0, sizeof(int) * toOrder.size(), toOrder.data(), nullptr, &event);
+            event.wait();
+        }
+    }
+
+    void WellContributionsOCLContainer::update_on_gpu(Opm::WellContributions &wellContribs){
+        if(num_std_wells > 0){
+            cl::Event event;
+            queue->enqueueWriteBuffer(s.Cnnzs, CL_TRUE, 0, sizeof(double) * wellContribs.h_Cnnzs_ocl.size(), wellContribs.h_Cnnzs_ocl.data(), nullptr, &event);
```
blattms, line 80: We are still writing to buffers that have been allocated during `init`. Can you comment on when that function is called and are you sure that the size is correct whenever this function is callled? As mentioned the size of the host buffers in wellContribs make change before each call to solve_system in the backend.

opm/simulators/linalg/bda/openclKernels.hpp
```diff
@@ -408,12 +408,10 @@ namespace bda
 
             b = wiId/valsPerBlock + val_pointers[wgId];
 
-            if(wiId < valsPerBlock){
-                if(c == 0 || c == 2) {localSum[wiId] += localSum[wiId + 2];}
-                if(c == 0 || c == 1) {localSum[wiId] += localSum[wiId + 1];}
-            }
-
             if(c == 0 && wiId < valsPerBlock){
+                for(unsigned int stride = 2; stride > 0; stride >>= 1){
```
blattms, line 412: Interesting notation for division by 2: `stride/=2`. but now it seems correct

opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
blattms: Seems to be still an issue. If you disagree then please point to where the size if the device buffers is changed or explain why this is not needed (I might be missing something).

ducbueno: This is strange. I've implemented a function called `reinit` that recreates the buffers if the vectors change size. I don't know why it is not showing here. 

There is an "outdated" flag appearing in this conversation, maybe this is why.

But the change is commited, I've checked in my repository.

blattms: Must have missed that. It is here, too.

blattms:
jenkins build this please

blattms:
Thanks for your persistence. I have run it on my machine and it works. Great job.

There is still some inconsistency in the linear iterations taken,see #2822.

blattms:
BTW did not see any exception for Norne.

PR #2817, opened 2020-09-28
Add inline fmtlib code in source tree

joakim-hove:


joakim-hove:
jenkins build this please

blattms:
This one I do not understand. Isn't fmtlib already included in opm-common?

joakim-hove:
Yes ut is included om opm-common, but er do not install ut.

atgeirr:
> Yes it is included in opm-common, but we do not install it.

This is getting a bit too much? I guess we cannot install it or get conflicts with non-embedded installation of the lib, but what about putting it in the opm/ tree and using a separate include "OpmFormat.hpp", which then can (eventually) point to the c++20 thing, similar to how we do the filesystem stuff?

joakim-hove:
> This is getting a bit too much? I guess we cannot install it or get conflicts with non-embedded installation of the lib, 

I think the current approach - embedding it in the repos where it is needed - and then (hopefully) forgetting about it is the simplest approach. 

> [....] and using a separate include "OpmFormat.hpp", which then can (eventually) point to the c++20 thing, similar to how we do the filesystem stuff?

We can do that - but I do not really see an advantage - the change when we have std support will be like this:
```C++
#include <fmt/format.h>
...
auto s = fmt::format("Hello {}", "world");
```
```C++
#include <format>
...
auto s  std::format("Hello {}", "world");
```
These should be able to coexists during a transition period.

joakim-hove:
It is my plan to merge this tomorrow - unless someone has a good alternative solution.

joakim-hove:
Hmmm - it was merged a bit before planned ... sorry about that.

blattms:
this is not a good idea.
Seems like you need to use the logging system. The problem is that there are more modules, opm-models, opm-material, opm-upscaling. Maybe eventually we will need it there, too. Also some modules using (like opm-common, opm-material, opm-grid) are used from user modules residing outside of OPM. Are we sure that this use case is covered? I don't think so:
```
mblatt@smaug:~/src/dune/opm/opm-common$ find . -name \*.hpp| xargs grep fmt
./opm/common/utility/OpmInputError.hpp:#include <fmt/format.h>
./opm/common/utility/OpmInputError.hpp:      The message string will be used as format string in the fmt::format()
./opm/common/utility/OpmInputError.hpp:    OpmInputError(const std::string& msg_fmt, const KeywordLocation& loc, const Args ... additionalArguments) :
./opm/common/utility/OpmInputError.hpp:        m_what   { OpmInputError::format(msg_fmt, loc, std::forward<const Args>(additionalArguments)...) },
./opm/common/utility/OpmInputError.hpp:        return fmt::format(msg_format,
./opm/common/utility/OpmInputError.hpp:            fmt::arg("keyword", loc.keyword),
./opm/common/utility/OpmInputError.hpp:            fmt::arg("file", loc.filename),
./opm/common/utility/OpmInputError.hpp:            fmt::arg("line", loc.lineno)
./opm/common/utility/OpmInputError.hpp:    // The location member is here for debugging; depending on the msg_fmt
./opm/parser/eclipse/EclipseState/IOConfig/RestartConfig.hpp:        static std::string getRestartFileName(const std::string& restart_base, int report_step, bool unified, bool fmt_file);
./opm/io/eclipse/OutputStream.hpp:        /// \param[in] fmt Whether or not to create formatted output files.
./opm/io/eclipse/OutputStream.hpp:                      const Formatted& fmt);
./opm/io/eclipse/OutputStream.hpp:        /// \param[in] fmt Whether or not to create formatted output files.
./opm/io/eclipse/OutputStream.hpp:                         const Formatted& fmt,
./opm/io/eclipse/OutputStream.hpp:        /// \param[in] fmt Whether or not to create formatted output files.
./opm/io/eclipse/OutputStream.hpp:                     const Formatted&    fmt,
./opm/io/eclipse/OutputStream.hpp:                                      const Formatted&            fmt,
./opm/io/eclipse/OutputStream.hpp:                      const Formatted& fmt,
mblatt@smaug:~/src/dune/opm/opm-common$ 
```
Whenever these two headers are include we already need the fmt headers. I guess we already need to install it anyway.

joakim-hove:
> The problem is that there are more modules, opm-models, opm-material, opm-upscaling.

Well - my take on it is that if/when that is a problem we add them there as well. 

> Are we sure that this use case is covered? I don't think so:

That point I agree; the use in `OpmInputError.hpp` is problematic - I will look into that. The other finds are false positives.

Issue #2755, opened 2020-09-02
[OpenCL] Number of linear iterations fluctuate between multiple run of same model


blattms:
Running opm-tests/model1/BASE1_MSW_HFA multiple times the number of linear iterations is not the same for every run.

GitPaean:
I tried both parallel and serial, I did not reproduce your problem. 

There are other people also complaining about some random behavior, unfortunately, I did not reproduce either. 

Maybe some linear solution component in the system setup has some components can introduce some randomness in the solution procedure?

blattms:
Probably should have stated this in the text also (not just as a tag in the title).

The problem is only in the OpenCL code (if you explicitly ask for it with --gpu-mode=opencl).

GitPaean:
Okay. I did not notice the `OpenCL` in the title. 

I was thinking it is good to be able to reproduce and find out why about the randomness reported from some colleagues.

blattms:
Reroducing is always highly appreciated. Thanks.

Maybe @ducbueno wants to do that?

blattms:
This problem persist now that both std wells and ms wells are implemented. I get fluctuating number of iterations for SPE9 with current master (std well only) and with #2821 (readds ms wells). It happens both with `--matrix-add-well-contributions=true` and false. For Cusparse there are no fluctuations. I guess we need to check the ILU and bicgstab implementations.

ducbueno:
I believe the fluctuations are due to the OpenCL BiCGStab implementation always using the zero vector an initial guess. There may some other issues with the ILU, but I haven't checked it

Tongdongq:
The cusparseSolver also uses a zero vector as initial guess with
```cudaMemsetAsync(d_x, 0, sizeof(double) * N, stream);```
in `update_system_on_gpu()` and `copy_system_to_gpu()`.

Tongdongq:
@blattms @ducbueno 
I just remembered that the default OpenCL coloring strategy is set to GRAPH_COLORING as oppose to LEVEL_SCHEDULING in `openclSolverBackend.cpp`.
This is generally faster, since it respects dependencies less. This results in more parallelism, but also in more linear iterations needed (up to 2x is normal). What's more, the graph coloring strategy has a **random** output.
We could patch this by replacing
https://github.com/OPM/opm-simulators/blob/ac3004da9deaf8841eade3c4394f5f5c6cffb95b/opm/simulators/linalg/bda/Reorder.cpp#L60
[...]

PR #2821, opened 2020-09-29
Reintroduced multisegment wells to OpenCL backend

ducbueno:
Reintroduced multisegment wells to the new implementation of the well management of OpenCL backend (https://github.com/OPM/opm-simulators/pull/2816)

### DIFF
--- a/opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
+++ b/opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
@@ -167,18 +167,10 @@ namespace bda
     }
 
     WellContributionsOCLContainer::~WellContributionsOCLContainer(){
-        if(num_std_wells > 0){
-            toOrder.clear();
-        }
-
         if(num_ms_wells > 0){
             for (auto ms : multisegments) {
                 delete ms;
             }
-
-            multisegments.clear();
-            x_msw.clear();
-            y_msw.clear();
         }
     }
 } // end namespace bda
--- a/opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
+++ b/opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
@@ -30,18 +30,6 @@ namespace bda
     using Opm::OpmLog;
     using Dune::Timer;
 
-    WellContributionsOCLContainer::WellContributionsOCLContainer(){
-        multisegments.reset(new mswVecT());
-    }
-
-    WellContributionsOCLContainer::~WellContributionsOCLContainer(){
-        if(num_ms_wells > 0){
-            for (auto ms : *multisegments) {
-                delete ms;
-            }
-        }
-    }
-
     void WellContributionsOCLContainer::init(Opm::WellContributions &wellContribs, int N_, int Nb_){
         N = N_;
         Nb = Nb_;
@@ -91,8 +79,8 @@ namespace bda
         }
 
         if(!wellContribs.multisegments.empty()){
-            multisegments = std::move(std::make_unique<mswVecT>(wellContribs.multisegments));
-            num_ms_wells = multisegments->size();
+            multisegments = std::move(wellContribs.multisegments);
+            num_ms_wells = multisegments.size();
             x_msw.reserve(N);
             y_msw.reserve(N);
         }
@@ -116,7 +104,7 @@ namespace bda
         }
 
         if(!wellContribs.multisegments.empty()){
-            multisegments = std::move(std::make_unique<mswVecT>(wellContribs.multisegments));
+            multisegments = std::move(wellContribs.multisegments);
         }
     }
 
@@ -158,11 +146,10 @@ namespace bda
         event.waitForEvents(events);
 
         // actually apply MultisegmentWells
-        for(auto well: *multisegments){
-            well->setReordering(toOrder.data(), reorder);
+        for(Opm::MultisegmentWellContribution *well: multisegments){
+            well->setReordering(toOrder.data(), true);
             well->apply(x_msw.data(), y_msw.data());
         }
-
 
         // copy vector y from CPU to GPU
         queue->enqueueWriteBuffer(y, CL_FALSE, 0, sizeof(double) * N, y_msw.data(), nullptr, &event);
@@ -176,6 +163,14 @@ namespace bda
 
         if(num_ms_wells > 0){
             applyMSWells(x, y);
+        }
+    }
+
+    WellContributionsOCLContainer::~WellContributionsOCLContainer(){
+        if(num_ms_wells > 0){
+            for (auto ms : multisegments) {
+                delete ms;
+            }
         }
     }
 } // end namespace bda
--- a/opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
+++ b/opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
@@ -30,6 +30,18 @@ namespace bda
     using Opm::OpmLog;
     using Dune::Timer;
 
+    WellContributionsOCLContainer::WellContributionsOCLContainer(){
+        multisegments.reset(new mswVecT());
+    }
+
+    WellContributionsOCLContainer::~WellContributionsOCLContainer(){
+        if(num_ms_wells > 0){
+            for (auto ms : *multisegments) {
+                delete ms;
+            }
+        }
+    }
+
     void WellContributionsOCLContainer::init(Opm::WellContributions &wellContribs, int N_, int Nb_){
         N = N_;
         Nb = Nb_;
@@ -79,8 +91,8 @@ namespace bda
         }
 
         if(!wellContribs.multisegments.empty()){
-            multisegments = std::move(wellContribs.multisegments);
-            num_ms_wells = multisegments.size();
+            multisegments = std::move(std::make_unique<mswVecT>(wellContribs.multisegments));
+            num_ms_wells = multisegments->size();
             x_msw.reserve(N);
             y_msw.reserve(N);
         }
@@ -104,7 +116,7 @@ namespace bda
         }
 
         if(!wellContribs.multisegments.empty()){
-            multisegments = std::move(wellContribs.multisegments);
+            multisegments = std::move(std::make_unique<mswVecT>(wellContribs.multisegments));
         }
     }
 
@@ -146,7 +158,7 @@ namespace bda
         event.waitForEvents(events);
 
         // actually apply MultisegmentWells
-        for(Opm::MultisegmentWellContribution *well: multisegments){
+        for(auto well: *multisegments){
             well->setReordering(toOrder.data(), true);
             well->apply(x_msw.data(), y_msw.data());
         }
@@ -163,14 +175,6 @@ namespace bda
 
         if(num_ms_wells > 0){
             applyMSWells(x, y);
-        }
-    }
-
-    WellContributionsOCLContainer::~WellContributionsOCLContainer(){
-        if(num_ms_wells > 0){
-            for (auto ms : multisegments) {
-                delete ms;
-            }
         }
     }
 } // end namespace bda
--- a/opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
+++ b/opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
@@ -75,9 +75,9 @@ namespace bda
     }
 
     void WellContributionsOCLContainer::copy_to_gpu(Opm::WellContributions &wellContribs, int *toOrder_){
-        if(num_std_wells > 0){
-            toOrder.insert(toOrder.end(), toOrder_, toOrder_ + Nb);
+        toOrder.insert(toOrder.end(), toOrder_, toOrder_ + Nb);
 
+        if(num_std_wells > 0){
             cl::Event event;
             std::vector<cl::Event> events(7);
             queue->enqueueWriteBuffer(s.Cnnzs, CL_FALSE, 0, sizeof(double) * wellContribs.h_Cnnzs_ocl.size(), wellContribs.h_Cnnzs_ocl.data(), nullptr, &events[0]);
@@ -159,9 +159,10 @@ namespace bda
 
         // actually apply MultisegmentWells
         for(auto well: *multisegments){
-            well->setReordering(toOrder.data(), true);
+            well->setReordering(toOrder.data(), reorder);
             well->apply(x_msw.data(), y_msw.data());
         }
+
 
         // copy vector y from CPU to GPU
         queue->enqueueWriteBuffer(y, CL_FALSE, 0, sizeof(double) * N, y_msw.data(), nullptr, &event);
--- a/opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
+++ b/opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
@@ -24,14 +24,14 @@
 #include <dune/common/timer.hh>
 
 #include <opm/simulators/linalg/bda/WellContributionsOCLContainer.hpp>
-#include<iostream>
 
 namespace bda
 {
     using Opm::OpmLog;
     using Dune::Timer;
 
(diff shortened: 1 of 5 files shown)


blattms:
Can you please rebase this onto current master. I thought github would automatically detect that Most commits are already in master, but it is still showing all differences since the branch off.

opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
```diff
@@ -127,13 +135,50 @@ namespace bda
...
+
+        if(num_ms_wells > 0){
+            applyMSWells(x, y);
+        }
     }
 
     WellContributionsOCLContainer::~WellContributionsOCLContainer(){
-        toOrder.clear();
+        if(num_std_wells > 0){
+            toOrder.clear();
```
blattms, line 171: I don't think you have to explicitly clear the containers. std::vector will do that in the destructor anyway

blattms (commented):
Only look at the last commit, which looks good.  Will run sum tests.

blattms:
Probably, not related to this particular PR, but when running model1/BASE1_MSW_HFA.DATA I get a segmentation fault:
```
[smaug:27377] *** Process received signal ***
[smaug:27377] Signal: Segmentation fault (11)
[smaug:27377] Signal code: Address not mapped (1)
[smaug:27377] Failing at address: 0xf4
[smaug:27377] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x12730)[0x7f34b2187730]
[smaug:27377] [ 1] ../../opm-simulators/opm-parallel/bin/flow(_ZN3Opm28MultisegmentWellContribution9getColIdxEj+0x12)[0x55c69dbc4422]
[smaug:27377] [ 2] ../../opm-simulators/opm-parallel/bin/flow(_ZN3Opm28MultisegmentWellContribution5applyEPdS1_+0xca)[0x55c69dbc44fa]
[smaug:27377] [ 3] ../../opm-simulators/opm-parallel/bin/flow(_ZN3bda29WellContributionsOCLContainer12applyMSWellsERN2cl6BufferES3_+0x197)[0x55c69dbed1a7]
[smaug:27377] [ 4] ../../opm-simulators/opm-parallel/bin/flow(_ZN3bda19openclSolverBackendILj3EE13gpu_pbicgstabERNS_9BdaResultE+0x572)[0x55c69dbe8bd2]
[smaug:27377] [ 5] ../../opm-simulators/opm-parallel/bin/flow(_ZN3bda19openclSolverBackendILj3EE12solve_systemERNS_9BdaResultE+0x40)[0x55c69dbe9d30]
```
This worked for a run on Sep 2.

opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
atgeirr: Exactly. The only thing this constructor needs to do is to delete the new-ed pointers in the `multisegments` vector. However, the class is missing copy constructor, move constructor, copy assignment operator and move assignment operator. This is the "rule of 5" (used to be the "rule of 3" before move operations came in C++11), if you must implement one, you must implement all of them. As is now, copying this object will cause a crash, as the mentioned pointers will be double-deleted. A better approach is to use smart pointers. If you use unique_ptr instead, you do not have to implement any of these functions, but can safely use the defaults. The copy operations will then not be available since unique_ptr can be moved, not copied.

blattms: Simply delete the copy constructor then:

```
WellContributionsOCLContainer
{
  WellContributionsOCLContainer(const WellContributionsOCLContainer&) = delete;
}
```

ducbueno:
Sorry for pushing all these commits. Forgot to squash first on my local repository. I'm still getting used to git rebase.

blattms:
Sorry to say that, but it seems like, you messed up the rebase. There are now a lot of rebased commits that were already in master.

I assume that 3723d86 is the only relevant commit here. To clean this I would do (upstream pointing to https://github.com/OPM/opm-simulators and origin pointing to https://github.com/ducbueno/opm-simulators):
```
git fetch upstream
git checkout add-mswells
git checkout -b add-mswells-bak # Play it safe with a backup
git checkout add-mswells
git stash # Save uncomitted changes
git reset --hard upstream/master # Now this and master will be the same
git cherry-pick 3723d86
git stash pop # Playback uncommited changes
git push --force origin add-mswells # --force ill overwrite your branch on github.
```

ducbueno:
@blattms sorry for the mess (again). Hope this makes things right.

blattms:
No problem. Everything is cleaned up now.

opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
ducbueno: I tried to add the smart pointers for the move operations in the `WellContributionsOCLContainer` class, however I'm not sure If what I did is what you guys recommended. I still have some doubts of where the destructor for the components of the multisegment wells vector should be (I've put it in the `WellContributionsOCLContainer` class since that is where the vector is moved to). I didn't write the copy constructor, I guess I didn't really understand what that is for.

Anyways, let me know what needs to be fixed.

blattms: Seasoned programmers should not try to impress others in reviews. This time this could just have wasted your time (and mine, too). I will take a look.

atgeirr: > Seasoned programmers should not try to impress others in reviews.

In all seriousness, I just tried to be helpful. I think your suggest to delete the copy constructor is fine as well (but should also delete the copy assignment operator). I do not think it is wasteful to point out bad patterns in code, not sure what ticked you off here?

blattms: It is not 100% clear was @atgeirr was saying or trying to say. I think he should clarify things and maybe should have taken a closer look at the whole code.

Maybe he meant, to use a smartpointer for each entry of the container (and not the container itself)? Then no destructor would be needed and you would not need to add more methods. The problem is that unique_ptr is only movable and not copyable. Hence you cannot use all the functionality of std::vector (e.g. resizing, etc.) and you will have to use push or emplace entries in a special way with `std::move`). That seems like quite a burden to put on people new to the language, and will come as
a surprise to others if they are trying to use std::vector the usual way. Maybe he meant what you did use a std::unique_ptr<std::vector> >?

Otherwise in principle one should implement all five mentioned De- and Constructors to prevent surpises and make the code correct for every use case  is of course very good and should be adhered to. But this is still an internal class and more widespread usage is not foreseeable, yet. I guess there are quite some places in OPM where this is also not done yet, but should, with higher priority.

Indeed the problem that pointers might be deleted twice is gone now (as one cannot copy the class anymore), but the code complexity has increased quite a bit for that.

I am tempted to pass the GPU review torch to @atgeirr.

opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
```diff
@@ -79,8 +91,8 @@ namespace bda
         }
 
         if(!wellContribs.multisegments.empty()){
-            multisegments = std::move(wellContribs.multisegments);
-            num_ms_wells = multisegments.size();
+            multisegments = std::move(std::make_unique<mswVecT>(wellContribs.multisegments));
```
blattms: This is now a change of semantics, not sure whether we really want that.

Previously we moved the vector from `wellContribs` to here. Now we seem to make a copy of the vector in `wellContribs` just to put it in a `std::unique_ptr`.

opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
```diff
@@ -104,7 +116,7 @@ namespace bda
         }
 
         if(!wellContribs.multisegments.empty()){
-            multisegments = std::move(wellContribs.multisegments);
+            multisegments = std::move(std::make_unique<mswVecT>(wellContribs.multisegments));
```
blattms: Same here.

blattms (changes requested):
Personally, I think the unique_ptr is a step in the wrong direction here and I would have deleted the copy and move constructors. But maybe the clarification by @atgeirr might change my view.

opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
blattms: @atgeirr Can you please take a look at the changes you triggered and see whether that is what you suggested?

We now have a `std::unique_ptr<std::vector<MultiSegementWellContribution*>> multisegments` instead of a simple vector and that is initialized by copying ` multisegments = std::move(std::make_unique<mswVecT>(wellContribs.multisegments));`. Destructor is basically unchanged.

opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
ducbueno: I still have the old version saved locally.

I began to study the rule of 5 but I'm having some difficulties in finding examples of it and, consequently, on how to apply it in our case here. If you guys could point me to some examples it would help a lot.

Sorry for not being able to contribute more in this matter.

joakim-hove: Disclaimer: I have not followed this closely myself - and my knowledge of the rule of 5 is mostly theoretical. But still: maybe if you describe your problem with words it would be simpler for us to assist you?

ducbueno: I hadn't heard of the rule of 5 until yesterday and apparently my old implementation (before the last 2 commits) was disrespecting this rule and, therefore, was considered bad practice. So my problem is that I still don't know how and where to put the copy/move constructors and destructors. Should I put them in the `MultisegmentWell` class or in the `WellContributionsOCLContainer` class? Should these destructors and constructors operate in the elements of the vector (which are complex objects by themselves) or should they operate on the vector as whole? These are the questions I'm struggling with. But as I said, I've heard of all this yesterday and only today I'm grasping what it all means. I think I'll create another PR with my attempt at implementing it in this case but I don't know how soon I'll be able to do it

atgeirr: If you have the old version, the simplest may be to go back that, and explicitly delete the copy constructor and copy assignment operator in line with @blattms's suggestion, instead of introducing smart pointers. That would solve the problem of these operations being wrong.

I looked at the modifications done in the "smart pointer" commit, and became somewhat confused. What I intended was that instead of having a vector of pointers, you should have a vector of smart pointer. The change in that commit was something different, it now uses a smart pointer to a vector of pointers, which saves no cleanup. The "problem" is the plain pointers owning the resources they point to. I have now read a bit more of the code, and see that these pointers are not created in this class, but moved from another object, `wellContribs.multisegments`,  (that presumably has the same kind of issue). I think the best solution could be to refactor that class to using a `vector<unique_ptr<MultisegmentWellContribution>>` instead of a vector of plain pointers. Then the rest of the code here would remain unchanged except for that type. I do not know if this is feasible, as I have not looked at that class' code.

If pressed for time, do the simple solution (the first paragraph). In the long run, the second may save you some pain.

blattms: @ducbueno Please don't try the refactoring. If somebody insist on it that can be done later. You have enough on your plate and I am very grateful for your persistent bug hunting (which most of the time are not even due to you).

`wellContribs.multisegments` is used in CUDA code and that does not have the full C++ functionality. Maybe unique-ptr is even not fully supported there. Hence sticking with simple C for its entries might make sense,

atgeirr: > So my problem is that I still don't know how and where to put the copy/move constructors and destructors. Should I put them in the `MultisegmentWell` class or in the `WellContributionsOCLContainer` class? Should these destructors and constructors operate in the elements of the vector (which are complex objects by themselves) or should they operate on the vector as whole?

The goal is to avoid writing any of the 5 functions (including destructors). You can accomplish that by using appropriate smart pointers, because then the compiler-generated default versions of the 5 functions will be correct.

The problem is that when a plain pointer is destroyed, what it points to is not `delete`d. So if you have a vector of plain pointers, when the vector dies it will deallocate the array where the pointers were stored without calling `delete` on them. Therefore, you did the right thing to explicitly call `delete` in the destructors. However, this is error prone, since you have to make the other of the 5 functions for this to be correct (or, explicitly delete those functions).

By replacing the plain pointers of the `multisegment` member in the WellContributions class with std::unique_ptr, you accomplish the following: you do not have to `delete` the pointers anymore, simplifying your WellContributions destructor. (If you had implemented all 5 of the set, you would see similar simplifications in all of them.) Even better, the other class that "steals" these pointers by moving them into its own vector, would also no longer need to delete them. (This was the place that triggered the original comments).

Two more things:
 - In the WellContributions class there are several more members handled by explicit pointers, which prevents you from removing the destructor in that class. I think that all of these can be replaced with std::vector instead. (Using std::vector is better than using a smart pointer to an array, since it has many useful methods, and everyone knows how to use it.) I am not sure what the freeCudaMemory() call does, you may still need to do that in the destructor for all I know.
 - If you use std::unique_ptr in a class, you will not be able to copy objects of that class (assuming you do just let the compiler implement the default versions for the 5 functions). Two of the 5, the copy constructor and copy assignment operator, will not be available. This is because it is not allowed to copy std::unique_ptr. Often, this is exactly what you want! It is possible to move a unique_ptr, so that is still allowed for the class containing it.

ducbueno: @blattms please check the last commit and see if everything is correct. The `BASE1_MSW_HFA.DATA` is now running again.

@atgeirr thank you for the very complete answer, it made things a lot clearer. I'll experiment with the smart pointers in the way you suggested in a new branch for a while. However I may not have the necessary time to refactor things fully, since as @blattms the `wellContribs` object is also used by CUDA and I'll have to be extra careful for not breaking that and I also have to give some attention to https://github.com/OPM/opm-simulators/issues/2755, as it is related to my PhD project.

Sorry for causing all this discussion and thank you again. At least I was able to learn something :)

atgeirr: > At least I was able to learn something :)

Then it was worthwhile, I guess!

opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
```diff
@@ -127,13 +135,42 @@ namespace bda
...
+        std::vector<cl::Event> events(2);
+
+        // copy vectors x and y from GPU to CPU
+        queue->enqueueReadBuffer(x, CL_FALSE, 0, sizeof(double) * N, x_msw.data(), nullptr, &events[0]);
+        queue->enqueueReadBuffer(y, CL_FALSE, 0, sizeof(double) * N, y_msw.data(), nullptr, &events[1]);
+        event.waitForEvents(events);
+
+        // actually apply MultisegmentWells
+        for(Opm::MultisegmentWellContribution *well: multisegments){
+            well->setReordering(toOrder.data(), true);
```
blattms, line 150: Why are we always using true? There was one version where this was reorder.

blattms (commented):
Good, I wlll rerun some tests. The only thing I am wondering about is why we always reorder.

I have posted a PR on this branch that also deletes the assignment operator. Would be nice if you would merge this into here.

opm/simulators/linalg/bda/WellContributionsOCLContainer.cpp
ducbueno: The reorder seem to always occur in the OpenCL backend. The ILU allways produce two reorder vectors and one of these vectors (`toOrder`) is always passed to the `WellContributionsOCLContainer` class. That is why I left the reordering operator set to `true`. It seemed useless to set a variable that never changes in this context.

blattms:
jenkins build this please

blattms (approved):

blattms:
tested by jenkins and locally

PR #2824, opened 2020-09-30
Implements support for gas lift optimization.

hakonhagland:
Depends on https://github.com/OPM/opm-common/pull/1968.

Implements gas lift optimization for a single well. Support for gas lift optimization for multi-segment wells, groups of wells and networks is not implemented yet.

The keywords `LIFTOPT`, `WLIFTOPT`, and `VFPPROD` are used to supply parameters for the optimization. Also adds support for summary output of liftgas injection rate via summary keyword `WGLIR`.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -26,6 +26,7 @@
 
 #include <utility>
 #include <algorithm>
+#include <fmt/format.h>
 
 namespace Opm {
     template<typename TypeTag>
@@ -403,22 +404,10 @@ namespace Opm {
     BlackoilWellModel<TypeTag>::gliftDebug(
         const std::string &msg, Opm::DeferredLogger &deferred_logger) const
     {
-        std::ostringstream ss;
-        ss << msg;
-        gliftDebug(ss, deferred_logger);
-    }
-
-    template<typename TypeTag>
-    void
-    BlackoilWellModel<TypeTag>::gliftDebug(
-        std::ostringstream &ss, Opm::DeferredLogger &deferred_logger) const
-    {
         if (this->glift_debug) {
-            std::string message = ss.str();
-            if (message.empty()) return;
-            std::ostringstream ss2;
-            ss2 << "  GLIFT (DEBUG) : BlackoilWellModel : " << message;
-            deferred_logger.info(ss2.str());
+            const std::string message = fmt::format(
+                "  GLIFT (DEBUG) : BlackoilWellModel : {}", msg);
+            deferred_logger.info(message);
         }
     }
 
@@ -841,9 +830,9 @@ namespace Opm {
 
         Opm::DeferredLogger local_deferredLogger;
         if (this->glift_debug) {
-            std::ostringstream os;
-            os << "assemble() : iteration = " << iterationIdx;
-            gliftDebug(os, local_deferredLogger);
+            const std::string msg = fmt::format(
+                "assemble() : iteration {}" , iterationIdx);
+            gliftDebug(msg, local_deferredLogger);
         }
         last_report_ = SimulatorReportSingle();
         Dune::Timer perfTimer;
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1375,7 +1375,6 @@ namespace Opm {
                     std::vector<double> potentials;
                     well->computeWellPotentials(ebosSimulator_, B_avg, well_state_copy, potentials, deferred_logger);
                     // putting the sucessfully calculated potentials to the well_potentials
-                    well_state_.updateALQ(well_state_copy);
                     for (int p = 0; p < np; ++p) {
                         well_potentials[well->indexOfWell() * np + p] = std::abs(potentials[p]);
                     }
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -330,7 +330,7 @@ namespace Opm {
         Opm::DeferredLogger local_deferredLogger;
 
         well_state_ = previous_well_state_;
-
+        well_state_.disableGliftOptimization();
         const int reportStepIdx = ebosSimulator_.episodeIndex();
         const double simulationTime = ebosSimulator_.time();
 
@@ -393,8 +393,34 @@ namespace Opm {
         //compute well guideRates
         const auto& comm = ebosSimulator_.vanguard().grid().comm();
         WellGroupHelpers::updateGuideRatesForWells(schedule(), phase_usage_, reportStepIdx, simulationTime, well_state_, comm, guideRate_.get());
+        logAndCheckForExceptionsAndThrow(local_deferredLogger,
+            exception_thrown, "beginTimeStep() failed.", terminal_output_);
+
     }
 
+    template<typename TypeTag>
+    void
+    BlackoilWellModel<TypeTag>::gliftDebug(
+        const std::string &msg, Opm::DeferredLogger &deferred_logger) const
+    {
+        std::ostringstream ss;
+        ss << msg;
+        gliftDebug(ss, deferred_logger);
+    }
+
+    template<typename TypeTag>
+    void
+    BlackoilWellModel<TypeTag>::gliftDebug(
+        std::ostringstream &ss, Opm::DeferredLogger &deferred_logger) const
+    {
+        if (this->glift_debug) {
+            std::string message = ss.str();
+            if (message.empty()) return;
+            std::ostringstream ss2;
+            ss2 << "  GLIFT (DEBUG) : BlackoilWellModel : " << message;
+            deferred_logger.info(ss2.str());
+        }
+    }
 
     template<typename TypeTag>
     void
@@ -813,6 +839,12 @@ namespace Opm {
              const double dt)
     {
 
+        Opm::DeferredLogger local_deferredLogger;
+        if (this->glift_debug) {
+            std::ostringstream os;
+            os << "assemble() : iteration = " << iterationIdx;
+            gliftDebug(os, local_deferredLogger);
+        }
         last_report_ = SimulatorReportSingle();
         Dune::Timer perfTimer;
         perfTimer.start();
@@ -821,7 +853,6 @@ namespace Opm {
             return;
         }
 
-        Opm::DeferredLogger local_deferredLogger;
 
         updatePerforationIntensiveQuantities();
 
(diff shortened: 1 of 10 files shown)


hakonhagland:
jenkins build this please

akva2:
that build won't include the upstream, maybe you forgot opm-common=1968 in the trigger?

bska:
> That build won't include the upstream, maybe you forgot opm-common=1968 in the trigger?

In this case it doesn't matter because the upstream change is "only" responsible for outputting values to the summary file.  There's no API change here.

akva2:
I see. I'll shut up then. One question nonetheless: Should this remove keywords from MissingFeatures ?

bska:
>>> That build won't include the upstream, maybe you forgot opm-common=1968 in the trigger?
>> 
>> In this case it doesn't matter because the upstream change is "only" responsible for outputting values to the summary file.
>
> I see. I'll shut up then.

My apologies.  There **was** an API update.  I forgot that the upstream PR also added a new member to the [`Rates`](https://github.com/OPM/opm-common/blob/7db5d2fd5c05584b82945d9d2db17e942a103f5b/opm/output/data/Wells.hpp#L72) object.  So you were right the first time, we do need the upstream PR for this to even build.

> One question nonetheless: Should this remove keywords from `MissingFeatures`?

Yes I believe it should, at least if we think that the features are complete (or nearly so) as of this PR.  I could see us removing at least the following entries:

* [`LIFTOPT`](https://github.com/OPM/opm-simulators/blob/c26aefdd5b523b75753633ee12694dfe4a107e00/opm/simulators/flow/MissingFeatures.cpp#L413)
* [`WLIFTOPT`](https://github.com/OPM/opm-simulators/blob/c26aefdd5b523b75753633ee12694dfe4a107e00/opm/simulators/flow/MissingFeatures.cpp#L823)

Possibly others as well.

bska:
jenkins build this opm-common=1968 please

hakonhagland:
jenkins build this opm-common=1968 please

opm/simulators/wells/StandardWell.hpp
```diff
@@ -233,11 +235,74 @@ namespace Opm
...
+            const Simulator& ebosSimulator,
+            DeferredLogger& deferred_logger
+        ) const;
+
+        virtual void maybeDoGasLiftOptimization (
+            const WellState& well_state,
+            const Simulator& ebosSimulator,
+            DeferredLogger& deferred_logger
+        ) const;
+
```
totto82, line 249: What is the logic of splitting these functions? I hope we can move this ++ function to the interface to make GLIFT work for MSW.

opm/simulators/wells/GasLiftRuntime_impl.hpp
```diff
@@ -0,0 +1,798 @@
...
+    controls_{controls},
+    deferred_logger_{deferred_logger},
+    ebos_simulator_{ebos_simulator},
+    potentials_{potentials},
+    std_well_{std_well},
+    summary_state_{summary_state},
+    well_state_{well_state},
+    debug{true},  // extra debugging output
+    debug_disable{false} // act as glift optimization is disabled
+{
```
totto82, line 51: I think we should not output the debug info as default
Also do we need the posibility to hard-code disabling of the glift optimization anymore

totto82:
Thanks for the contribution. I have done some testing and it works as indented for the tests I have run. I think this can be merged when the comments above are addressed. The test failures are related to the extra output.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -813,6 +839,12 @@ namespace Opm {
              const double dt)
     {
 
+        Opm::DeferredLogger local_deferredLogger;
+        if (this->glift_debug) {
+            std::ostringstream os;
+            os << "assemble() : iteration = " << iterationIdx;
+            gliftDebug(os, local_deferredLogger);
```
joakim-hove, line 846: With the `libfmt`library which I have shoehorned in you could do:
```
#include <fmt/format.h>
...
const std::string msg = fmt::format("assemble() : iteration {}" , iterationIdx);
```
Pick your poison

opm/simulators/wells/GasLiftRuntime_impl.hpp
```diff
@@ -0,0 +1,798 @@
...
+}
+
+template<typename TypeTag>
+void
+Opm::GasLiftRuntime<TypeTag>::
+displayWarning_(std::string msg)
+{
+    std::ostringstream ss;
+    ss << "GAS LIFT OPTIMIZATION, WELL: " << this->well_name_ << " : " << msg;
+    this->deferred_logger_.warning("WARNING", ss.str());
```
joakim-hove, line 196: fmt alternative:
```
const auto& msg = fmt::format("GAS LIFT OPTIMIZATION, WELL: {} : {}", this->well_name_, msg);
```

opm/simulators/wells/StandardWell.hpp
hakonhagland: Do you mean the function `maybeDoGasLiftOptimization()` ? We call it here : https://github.com/OPM/opm-simulators/blob/52f6f819eeebd0e6a6bb53eb8aa59ff0f271664e/opm/simulators/wells/BlackoilWellModel_impl.hpp#L898. Since it is a call through a base class (`WellInterface`) pointer I had to make it a virtual function.

opm/simulators/wells/GasLiftRuntime_impl.hpp
hakonhagland: Thanks for the comment! I have updated the PR.

opm/simulators/wells/BlackoilWellModel_impl.hpp
hakonhagland: Thanks this looks much better! I have updated the code to use `fmt::format()`

hakonhagland:
jenkins build this opm-common=1968 please

opm/simulators/wells/GasLiftRuntime_impl.hpp
```diff
@@ -0,0 +1,791 @@
...
+    std_well_{std_well},
+    summary_state_{summary_state},
+    well_state_{well_state},
+    debug{false}  // extra debugging output
+{
+    int well_index = this->std_well_.indexOfWell();
+    const Well::ProducerCMode& control_mode
+        = well_state_.currentProductionControls()[well_index];
+    assert(control_mode == Well::ProducerCMode::THP );
+    const Opm::Schedule& schedule = this->ebos_simulator_.vanguard().schedule();
```
totto82, line 55: The control_mode is also used for the asset. Try OPM_OPTIM_UNUSED

opm/simulators/wells/GasLiftRuntime_impl.hpp
totto82: Good

totto82:
The test failures are due to new summary output. I plan to make a data update later today and merge this when the last comments are fixed.

opm/simulators/wells/GasLiftRuntime_impl.hpp
hakonhagland: > The `control_mode` is also used for the asset.

Did you mean `assert` ? I am not sure what you are referring to

totto82: Yes, I mean assert. The control_mode is unused if not build in debug mode since the assertion will be skipped. 
```
const Well::ProducerCMode& OPM_OPTIM_UNUSED control_mode
        = well_state_.currentProductionControls()[well_index];
```
would make the compiler not warn against unused variable, I think. (not tested)

joakim-hove: Is there a good argument for `assert( ... ) ` rather than e.g.
```
if (control_mode != Well::ProducerCMode::THP)
     throw std::logic_error("Bug in flow - invalid control mode detected\n");
```

hakonhagland: Thanks for the comments! I have updated the PR.

totto82:
jenkins build this opm-common=1968 update_data please

totto82:
jenkins build this opm-common=1968 opm-tests=397 please

atgeirr:
I see a new warning:
```
/Users/atgeirr/opm/src/opm-simulators/opm/simulators/wells/StandardWell.hpp:245:22: warning: 'maybeDoGasLiftOptimization' overrides a member function but is
      not marked 'override' [-Winconsistent-missing-override]
        virtual void maybeDoGasLiftOptimization (
```
Also in MultiSegmentWell. Will you fix it (add the `override`)?

hakonhagland:
> not marked 'override' [-Winconsistent-missing-override]

@atgeirr Fixed in #2833

OPMUSER:
All the Model 06 tests fail on the the 2020-10 release because the SUMMARY vectors associated with FIPABC keyword do not exist. This is because in all cases except one the FIPABC keyword in REGIONS section is missing. The output for these cases is:

image

Which is good as the simulator catches the error, but not so good for getting the results from the test cases. However, if you run in parallel then the OPM Flow throws an exception:

image

Now for the case that has the FIPABC keyword, `2_WLIFT_MOFEL5` , you get the same, even though SUMMARY vectors are defined as RPR__ABC. This should work.  However, it does work for the Norne model, but here the FIP keyword for Norne is:

`EQUALS
-- FIPs based on geological layers
FIPGL 1    1  46   1 112   1  3   /  Garn 
FIPGL 2    1  46   1 112   4  4   /  Not
FIPGL 3    1  46   1 112   5  5   /  Ile 2.2 
FIPGL 4    1  46   1 112   6  8   /  Ile 2.1
FIPGL 5    1  46   1 112   9  11  /  Ile 1
FIPGL 6    1  46   1 112   12 12  /  Tofte 2.2
FIPGL 7    1  46   1 112   13 15  /  Tofte 2.1
FIPGL 8    1  46   1 112   16 18  /  Tofte 1
FIPGL 9    1  46   1 112   19 22  /  Tilje

-- FIPs based on numerical layers
FIPNL 1    1  46   1 112   1  1   /  Garn 3
FIPNL 2    1  46   1 112   2  2   /  Garn 2
FIPNL 3    1  46   1 112   3  3   /  Garn 1w
FIPNL 4    1  46   1 112   4  4   /  Not   
FIPNL 5    1  46   1 112   5  5   /  Ile 2.2
FIPNL 6    1  46   1 112   6  6   /  Ile 2.1.3      
FIPNL 7    1  46   1 112   7  7   /  Ile 2.1.2
FIPNL 8    1  46   1 112   8  8   /  Ile 2.1.1
FIPNL 9    1  46   1 112   9  9   /  Ile 1.3                  
FIPNL 10   1  46   1 112   10 10  /  Ile 1.2      
FIPNL 11   1  46   1 112   11 11  /  Ile 1.1
FIPNL 12   1  46   1 112   12 12  /  Tofte 2.2            
FIPNL 13   1  46   1 112   13 13  /  Tofte 2.1.3
FIPNL 14   1  46   1 112   14 14  /  Tofte 2.1.2
FIPNL 15   1  46   1 112   15 15  /  Tofte 2.1.1
FIPNL 16   1  46   1 112   16 16  /  Tofte 1.2.2
FIPNL 17   1  46   1 112   17 17  /  Tofte 1.2.1
FIPNL 18   1  46   1 112   18 18  /  Tofte 1.1
FIPNL 19   1  46   1 112   19 19  /  Tilje 4
FIPNL 20   1  46   1 112   20 20  /  Tilje 3
FIPNL 21   1  46   1 112   21 21  /  Tilje 2
FIPNL 22   1  46   1 112   22 22  /  Tilje 1
/`

So it would appear that two characters work but not three for the FIP keyword name.

PR #2826, opened 2020-10-01
fixed: make some indices constexpr

akva2:
also use the bool conditions for conditional deref'ing with the indices.
this helps avoid maybe uninitialized warnings with gcc.

please review carefully, but i think this should be equivalent.


akva2:
jenkins build this please

GitPaean:
Sorry. I did not see your request. I am adding it to my list now.

GitPaean:
jenkins build this please

GitPaean:
I think it is probably okay. In the future, we probably should change all the `FluidSystem::phaseIsActive`  to be consistent. And for better readability, we probably should define `has_water` and `has_gas` based on `FluidSystem`.

PR #2827, opened 2020-10-01
add options for timestepping

totto82:
This PR adds

- a new timestepping options "newtoniterationcount" which increase(decrease) timestep if number of newton iterations is smaller(larger) than the target newton iterations. The increase and decrease is give by TimeStepControlGrowthRate and TimeStepControlDecayRate

- a option to control the damping of the pid+(newton)iteration control
TimeStepControlGrowthDampingFactor
TimeStepControlDecayDampingFactor  
can be used to damp the increase/decrease of the timestep based on iterations counts. 

Defaults stays the same --> does not change any results.

### DIFF
--- a/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
+++ b/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
@@ -93,6 +93,14 @@ struct TimeStepControlGrowthRate {
     using type = UndefinedProperty;
 };
 template<class TypeTag, class MyTypeTag>
+struct TimeStepControlDecayDampingFactor {
+    using type = UndefinedProperty;
+};
+template<class TypeTag, class MyTypeTag>
+struct TimeStepControlGrowthDampingFactor {
+    using type = UndefinedProperty;
+};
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

alfbr:
I have now done quite a bit of testing. These options are really necessary to improve the numerical defaults further. Review/merge is highly appreciated.

atgeirr:
Looks fine, merging.

The parameters may be a little confusing for users, especially the TimeStepControlGrowthRate vs TimeStepControlGrowthDampingFactor (and similar with the decay variants), and the fact that only one of the sets will be used (depending on timestepping algo chosen), even though all parameters can be set. Any ideas for improving this?

alfbr:
> The parameters may be a little confusing for users, especially the TimeStepControlGrowthRate vs TimeStepControlGrowthDampingFactor (and similar with the decay variants), and the fact that only one of the sets will be used (depending on timestepping algo chosen), even though all parameters can be set. Any ideas for improving this?

Agreed. We should probably reorganize the options. At this point in time I believe it is a bit early though. The interplay between the different time stepping algorithms are non-trivial. Thanks for merging!

PR #2830, opened 2020-10-01
Implements a Python step_cleanup() method.

hakonhagland:
Implements a Python `step_cleanup()` method.

Continues the work in #2735 on implementing Python bindings for `flow`. The `step_cleanup()` method should be called after the last `step()` call to finalize the simulation.

### DIFF
--- a/opm/simulators/flow/python/simulators.hpp
+++ b/opm/simulators/flow/python/simulators.hpp
@@ -35,13 +35,16 @@ public:
     BlackOilSimulator( const std::string &deckFilename);
     int run();
     int step();
-    int step_init();
+    int stepInit();
+    int stepCleanup();
 
 private:
     const std::string deckFilename_;
+    bool hasRunInit_;
+    bool hasRunCleanup_;
+
     std::unique_ptr<FlowMainEbosType> mainEbos_;
     std::unique_ptr<Opm::Main> main_;
-    bool hasRunInit_;
 };
 
 } // namespace Opm::Python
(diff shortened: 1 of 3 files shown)


hakonhagland:
jenkins build this please

opm/simulators/flow/python/simulators.hpp
```diff
@@ -35,13 +35,16 @@ class BlackOilSimulator
     BlackOilSimulator( const std::string &deckFilename);
     int run();
     int step();
-    int step_init();
+    int stepInit();
+    int stepCleanup();
 
 private:
     const std::string deckFilename_;
+    bool hasRunInit_;
```
joakim-hove, line 43: These `hasRunxxx` methods should be initialized here with:
 ```
bool hasRunInit_ = false;
bool hasRunCleanup_ = false;
```

joakim-hove:
This can be merged with a minor update about the `hasRunxxx` flags.

opm/simulators/flow/python/simulators.hpp
hakonhagland: OK, as I understand: The reason is that in-class initialization is better than constructor initialization since the first automatically applies to all constructors of the class?  So this works like kind of a default value for the class, and if another value is required for one constructor it can be overridden there, right?

joakim-hove: I don't know if there is a general consensus on this; for this particular case my reasoning is quite different: I do not like the state flags `hasRunInit_` and `hasRunCleanup_` - this functionality should in some way be "guaranteed correct" (i.e. it should be design not be possible to make the errors that these two flags are protecting against). 

So - I therefor want to give the flags "less visibility" and prepare for removal.

hakonhagland:
jenkins build this please

PR #2831, opened 2020-10-02
Delay the check for missing features to *after* state construction

joakim-hove:
This will move the check for missing features to the end of the input processing; the missing keywords is a sort of "ok - here comes the rest which has not processed/supported at all" - in my opinion it makes sense to have that *after* the messages with more context.


joakim-hove:
jenkins build this pleas

alfbr:
My only concern is if usage of missing features is the direct cause of a crash, then seeing the list can help understanding what went wrong. I really do not know to what extent this is relevant though. so from my end we can go with this change and see how it works.

joakim-hove:
> My only concern is if usage of missing features is the direct cause of a crash

That is a legitimate point; to what extent it is relevant I also think it is difficult to assess

PR #2841, opened 2020-10-07
Add Serialization Support for Vector<bool>

bska:
The operator converts the elements to `unsigned char` and uses the regular vector operation with `complexType` set to `false`.

Add unit tests for `vector<bool>` and `vector<int>`.

### DIFF
--- a/ebos/eclmpiserializer.hh
+++ b/ebos/eclmpiserializer.hh
@@ -99,6 +99,30 @@ public:
         }
     }
 
+    void vector(std::vector<bool>& data)
+    {
+        std::vector<unsigned char> cdata;
+        if (m_op != Operation::UNPACK) {
+            cdata.resize(data.size());
+        }
+        if (m_op == Operation::PACKSIZE) {
+            this->vector<unsigned char, false>(cdata);
+        } else if (m_op == Operation::PACK) {
+            const auto n = data.size();
+            for (auto i = 0*n; i < n; ++i) {
+                cdata[i] = static_cast<unsigned char>(data[i]);
+            }
+            this->vector<unsigned char, false>(cdata);
+        } else if (m_op == Operation::UNPACK) {
+            this->vector<unsigned char, false>(cdata);
+            const unsigned char zero = 0;
+            const auto n = cdata.size();
+            data.resize(n);
+            for (auto i = 0*n; i < n; ++i) {
+                data[i] = cdata[i] != zero;
+            }
+        }
+    }
 
     //! \brief Handler for std::variant<> with four types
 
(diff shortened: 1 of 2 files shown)


bska:
Note: We could conceivably use `std::byte` instead of `unsigned char` here.  Furthermore, this&mdash;or something very similar&mdash;is needed to resolve the [build failures](https://ci.opm-project.org/job/opm-common-PR-builder/2604/) of OPM/opm-common#2002.

bska:
jenkins build this please

akva2:
This isn't really necessary. std::vector<bool> is handled in the underlying serializer.
just do
```
serializer(vector_of_bool);
```
instead of
```
serializer.vector(vector_with_bool);
```

bska:
> `std::vector<bool>` is handled in the underlying serializer,

Did you actually test that?  I suppose the `packSize()` might be correct, but the [`pack()`](https://github.com/OPM/opm-simulators/blob/f528adb604bfc1628b1537f601e826b6ef14f415/opm/simulators/utils/ParallelRestart.hpp#L172-L174) and [`unpack()`](https://github.com/OPM/opm-simulators/blob/f528adb604bfc1628b1537f601e826b6ef14f415/opm/simulators/utils/ParallelRestart.hpp#L100-L101) don't look like they'd work out of the box with `vector<bool>` compressed representation and proxy objects.

akva2:
it definitely is/has been in use. why do you think they don't work, they look fine to me?

bska:
> why do you think they don't work, they look fine to me?

This loop is wrong for proxy objects
```C++
for (const auto& x : vector_bool) {
    // use x
}
```
It either needs to be
```C++
for (const bool x : vector_bool) {
    // use x
}
```
or
```C++
for (auto&& x : vector_bool) {
    // use x
}
```

akva2:
i don't get it. it apparently works just fine. according to my debugger the type of x is const bool. i guess const_iterator is implementation defined, but mandated to deref to const_reference (which is a bool) ?

bska:
> according to my debugger the type of x is `const bool`.

Then the debugger is lying to you.  The actual type could be something like
```
const std::_Bit_reference &
```
(GCC/Clang) or
```
const std::_Vb_reference<std::_Wrap_alloc<std::allocator<std::_Vbase>>> &
```
(MSVC).

Anyway, I'll switch to the existing support and close this.

atgeirr:
What about simply outlawing all use of `vector<bool>`...?

blattms:
@bska Because we first copy and the [pack/unpack](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/utils/ParallelRestart.cpp#L391-L399).

@atgeirr I would vote for that. `std:vector<bool>` is a plague.

Issue #2822, opened 2020-09-30
[opencl] linear iterations are inconsistent (standard wells)


blattms:
At least with standard wells the linear iterations when using opencl differ inconsistently from when using Cuda

| Case | OpenCl | Cuda | CPU |
|---|---|---|---|
|   | lin. It | lin It | lin It |
|---|---|---|---|
|SPE1CASE2 | 16275 | 3241 | 3502 |
|SPE9 | 1729 | 1464 | 1619 |
| Norne | 42170 | 22292 | 22545 |

blattms:
Could be a difference in the stopping criterion. At least on SPE1CASE2 we need less nonlinear iterations, which might mean that the linear solver reduces the error more than the one for CUDA.

blattms:
@ducbueno  investigated this a bit and the CUDA implementation reuses the old solution as initial guess for the next while opencl always starts with zero initial guess.

ducbueno:
Markus I think I made a mistake in my statement. CUDA also sets the initial solutions to zero. See lines 301 and 330 of `cusparseSolverBackend.cu`. However I still think that reusing the old solution as initial value can yield advantages. I'll run some tests.

PR #2842, opened 2020-10-07
[GPU] Tweak BiCGStab

ducbueno:
Wrote a small change to the BiCGStab algorithm in the OpenCL backend, in order to tackle issue https://github.com/OPM/opm-simulators/issues/2822.

The initial solution for the algorithm is set to the solution obtained in the previous iteration, instead of being set to the zero vector at every iteration.

Still haven't run extensive tests. However the number of linear iterations for the 'SPE1CASE2.DATA' have reduced to ~13000 (~18% reduction).

### DIFF
--- a/opm/simulators/linalg/bda/openclKernels.hpp
+++ b/opm/simulators/linalg/bda/openclKernels.hpp
@@ -440,6 +440,71 @@ namespace bda
     }
     )";
 
+    inline const char* blockscaleadd_s = R"(
+    __kernel void blockscaleadd(__global const double *vals,
+                                __global const int *cols,
+                                __global const int *rows,
+                                __global const double *x,
+                                __global const double *b,
+                                __global double *ar,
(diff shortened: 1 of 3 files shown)


blattms:
Please document the blockscaleadd method. It is a bit inconvenient if we need to read the source of the openCL kernel. This does y=b-Ax, right? Would not have guessed that from the name.

As discussed this should be tested on several models and reported here.

To merge it, it should be optional.

I might be way off here: Using two kernels (y=A*x and y=b-y), might make the code more explicit and maintainable. No need to change that now, But this should be tested somewhen.

blattms:
@ducbueno somehow this stalled. Should we proceed here?

ducbueno:
At the time this was showing no significant changes neither in simulation time or total linear iterations. Not sure if it is worth keeping this.

blattms:
Closing it then. You may reopen, of course

PR #2844, opened 2020-10-08
Implements access to the porosity from Python.

hakonhagland:
Depends on #2830. 

Implements access (read/write) to the porosity from Python.

### DIFF
--- a/opm/simulators/flow/python/simulators.hpp
+++ b/opm/simulators/flow/python/simulators.hpp
@@ -25,7 +25,6 @@
 #include <opm/models/utils/propertysystem.hh>
 #include <pybind11/pybind11.h>
 #include <pybind11/numpy.h>
-#include <pybind11/embed.h>
 
 namespace py = pybind11;
 
--- a/opm/simulators/flow/python/simulators.hpp
+++ b/opm/simulators/flow/python/simulators.hpp
@@ -35,13 +35,16 @@ public:
     BlackOilSimulator( const std::string &deckFilename);
     int run();
     int step();
-    int step_init();
+    int stepInit();
+    int stepCleanup();
 
 private:
     const std::string deckFilename_;
+    bool hasRunInit_;
+    bool hasRunCleanup_;
+
     std::unique_ptr<FlowMainEbosType> mainEbos_;
     std::unique_ptr<Opm::Main> main_;
-    bool hasRunInit_;
 };
 
 } // namespace Opm::Python
--- a/opm/simulators/flow/python/simulators.hpp
+++ b/opm/simulators/flow/python/simulators.hpp
@@ -40,12 +40,12 @@ public:
 
 private:
     const std::string deckFilename_;
-    bool hasRunInit_;
-    bool hasRunCleanup_;
+    bool hasRunInit_ = false;
+    bool hasRunCleanup_ = false;
 
     std::unique_ptr<FlowMainEbosType> mainEbos_;
     std::unique_ptr<Opm::Main> main_;
 };
 
-} // namespace Opm::Python
+} // namespace Opm::Pybind
 #endif // OPM_SIMULATORS_HEADER_INCLUDED
--- a/opm/simulators/flow/python/simulators.hpp
+++ b/opm/simulators/flow/python/simulators.hpp
@@ -1,8 +1,5 @@
 /*
-  Copyright 2013, 2014, 2015 SINTEF ICT, Applied Mathematics.
-  Copyright 2014 Dr. Blatt - HPC-Simulation-Software & Services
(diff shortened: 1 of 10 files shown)


hakonhagland:
jenkins build this please

opm/simulators/flow/python/simulators.hpp
```diff
@@ -19,21 +16,32 @@
...
+
 #ifndef OPM_SIMULATORS_HEADER_INCLUDED
 #define OPM_SIMULATORS_HEADER_INCLUDED
 
 #include <opm/simulators/flow/Main.hpp>
 #include <opm/simulators/flow/FlowMainEbos.hpp>
+#include <opm/models/utils/propertysystem.hh>
+#include <pybind11/pybind11.h>
+#include <pybind11/numpy.h>
+#include <pybind11/embed.h>
```
joakim-hove, line 28: Should not need the `embed.h` header?

python/test/test_basic.py
```diff
@@ -0,0 +1,64 @@
+import unittest
+from opm2.simulators import BlackOilSimulator
+import os
+
+class TestBasic(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        os.chdir('test_data/SPE1CASE1')
+        # NOTE: Loading the below extension module involves loading a
```
joakim-hove, line 9: Good description of an unfortunate situation

python/test/test_basic.py
```diff
@@ -0,0 +1,64 @@
+import unittest
+from opm2.simulators import BlackOilSimulator
+import os
+
+class TestBasic(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        os.chdir('test_data/SPE1CASE1')
```
joakim-hove, line 8: There are two problems with the `os.chdir('test_data/SPE1CASE1')`:

1. When you hardcode the `test_data/SPE1CASE` that means the code *must* be run with cwd exactly at `python/test`. If you use `test_dir = os.path.join( os.path.dirname(__file__), "test_data/SPE1CASE1")` you are guaranteed to get the correct test directory irrespective of current cwd.

2. To change cwd changes the *state* of the process - that is ugly. With Python I recommend creating a small context:
```
from contextlib import contextmanager

@contextmanager
def pushd(path):
    cwd = os.getcwd()
    if not os.path.isdir(path):
        os.makedirs(path)
    os.chdir(path)

    yield

    os.chdir(cwd)
```
Then you can in your tests temporarily change directory with:
```
with pushd("/some/path/somewhere"):
      do_stuff_in_path()
      more_stuff()
# Now we are back in the original cwd
continue_in_cwd0()
```

python/simulators/CMakeLists.txt
```diff
@@ -13,3 +19,21 @@ set(PYTHON_PACKAGE_PATH "site-packages")
...
+      DESTINATION ${PROJECT_BINARY_DIR}/python)
+file( COPY ${PROJECT_SOURCE_DIR}/python/test_data
+      DESTINATION ${PROJECT_BINARY_DIR}/python)
+
+find_program (BASH_EXECUTABLE bash)
+if (BASH_EXECUTABLE)
+  add_test(NAME python_tests
+    WORKING_DIRECTORY ${PROJECT_BINARY_DIR}/python
+    COMMAND ${CMAKE_COMMAND} -E env PYTHONPATH=${PROJECT_BINARY_DIR}/python
+            ${BASH_EXECUTABLE} run-python-tests.sh )
```
joakim-hove, line 38: Could you not skip the entire `bash` script and use the cmake command:
```
COMMAND ${CMAKE_COMMAND} -E env PYTHONPATH=${PROJECT_BINARY_DIR/python:ENV{PYTHONPATH}
        ${PYTHON_EXECUTABLE} -m unittest
```

python/test/test_basic.py
hakonhagland: Good idea. But it seems the cwd needs to stay at `...test_data/SPE1CASE1` during the whole test, so if I use `with pushd..` in `setUpClass()` the cwd is reset when that finishes, and then `step_init()` throws `Exception received: Cannot find input case 'SPE1CASE1.DATA'.`

joakim-hove: OK - the `pushd` should be in each of the test routines, whereas the `setUpClass` should just store the testpath.

hakonhagland:
jenkins build this please

python/simulators/CMakeLists.txt
hakonhagland: Yes! See the new commit.

hakonhagland:
jenkins build this please

joakim-hove:
Then we merge - patient effort!

PR #2845, opened 2020-10-08
Add Calculator for Connection/Well Productivity Index

bska:
Switches between using the logarithmic and unit scaling factor based on whether or not the well has an explicit, positive drainage radius (`WELSPECS` item 7).  Does presently not include the D factor.

Add a set of unit tests to exercise the facility.

PR #2846, opened 2020-10-09
Unconditionally Calculate PI at End of Timestep 

bska:
This PR ensures that we calculate the well and connection level per-phase steady-state productivity index (PI) at the end of a completed time step (triggered from `endTimeStep()`).

We add a new data member,
```
BlackoilWellModel<>::prod_index_calc_
```
which holds one `WellProdIndexCalculator` for each of the process' local wells and a new interface member function
```
WellInterface::updateProductivityIndex
```
which uses a per-well PI calculator to actually compute the PI values and store those in the `WellState`.  Implement this member function for both `StandardWell` and `MultisegmentWell`.  Were it not for `getMobility` existing only in the derived classes, the two equal implementations could be merged and moved to the interface.

We also add a new data member to the `WellStateFullyImplicitBlackoil` to hold the connection-level PI values.  Finally, remove the conditional PI calculation from `StandardWell`'s well equation assembly routine.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -24,6 +24,10 @@
 #include <opm/simulators/utils/DeferredLoggingErrorHelpers.hpp>
 #include <opm/simulators/linalg/MatrixBlock.hpp>
 
+#include <algorithm>
+#include <functional>
+#include <numeric>
+
 namespace Opm
 {
 
@@ -2288,42 +2292,41 @@ namespace Opm
                             WellState& well_state,
                             DeferredLogger& deferred_logger) const
     {
-        auto recipFVF = [&ebosSimulator, this](const int perf, const int p) -> double
+        auto fluidState = [&ebosSimulator, this](const int perf)
         {
             const auto cell_idx = this->well_cells_[perf];
-            const auto& fs = ebosSimulator.model()
+            return ebosSimulator.model()
                .cachedIntensiveQuantities(cell_idx, /*timeIdx=*/ 0)->fluidState();
-
-            return fs.invB(p).value();
         };
 
-        auto Rs = [&ebosSimulator, this](const int perf) -> double
+        auto isInjecting = [this](const typename StandardWell<TypeTag>::FluidState& fs,
+                                  const int perf) -> bool
         {
-            const auto cell_idx = this->well_cells_[perf];
-            const auto& fs = ebosSimulator.model()
-               .cachedIntensiveQuantities(cell_idx, /*timeIdx=*/ 0)->fluidState();
+            const auto cell_press = this->extendEval(this->getPerfCellPressure(fs));
+            const auto conn_press = this->getBhp() + this->perf_pressure_diffs_[perf];
+            const auto drawdown   = cell_press - conn_press;
 
-            return fs.Rs().value();
+            return drawdown.value() < 0.0;
         };
 
-        auto Rv = [&ebosSimulator, this](const int perf) -> double
+        const int np = this->number_of_phases_;
+        auto setToZero = [np](double* x) -> void
         {
-            const auto cell_idx = this->well_cells_[perf];
-            const auto& fs = ebosSimulator.model()
-               .cachedIntensiveQuantities(cell_idx, /*timeIdx=*/ 0)->fluidState();
-
-            return fs.Rv().value();
+            std::fill_n(x, np, 0.0);
         };
 
-        const auto& pu = this->phaseUsage();
-        const int   np = this->number_of_phases_;
+        auto addVector = [np](const double* src, double* dest) -> void
+        {
+            std::transform(src, src + np, dest, dest, std::plus<>{});
+        };
+
+        const auto allow_xflow = this->getAllowCrossFlow()
+            || this->openCrossFlowAvoidSingularity(ebosSimulator);
 
         auto* wellPI = &well_state.productivityIndex()[this->index_of_well_*np + 0];
         auto* connPI = &well_state.connectionProductivityIndex()[this->first_perf_*np + 0];
 
-        for (int p = 0; p < np; ++p) {
-            wellPI[p] = 0.0;
-        }
+        setToZero(wellPI);
 
         const auto& allConn = this->well_ecl_.getConnections();
         const auto  nPerf   = allConn.size();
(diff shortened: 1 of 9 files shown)


bska:
Note that as this PR depends on #2845 it must not be merged before that PR.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
Marking this as ready for review now that the upstream PR #2845 is merged into master.

bska:
jenkins build this please

blattms:
For `getMobility` the only difference is the `deferredLogger` missing for Multisegment wells. Maybe that should be used there, too?

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
I'm now reasonably confident that we've covered the basic implementation of a complete PI/II calculation and so I'm marking this as "ready for review".

There are nevertheless a couple of caveats here which reviewers should note.  The cross flow implementation is speculative, but mirrors the approach to computing connection-level flow rates.  I should also note that the multisegment well implementation is not very thoroughly tested.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

jalvestad:
I have tested this pullrequest together with pr2866 for a comprehensive set of WELPI data for both production and injection wells. The tests have covered the different ways of using the WELPI keyword. These two pull-requests work correctly for all tests performed and are therefore judged to work according to the specification in the manual - and as Eclipse works!

bska:
jenkins build this please

bska:
> These two pull-requests work correctly for all tests performed and are therefore judged to work according to the specification in the manual - and as Eclipse works!

Thank you so much for taking on the testing task.  This is much appreciated.

bska:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -2263,6 +2240,91 @@ namespace Opm
...
+        {
+            const auto cell_idx = this->well_cells_[perf];
+            return ebosSimulator.model()
+               .cachedIntensiveQuantities(cell_idx, /*timeIdx=*/ 0)->fluidState();
+        };
+
+        auto isInjecting = [this](const typename StandardWell<TypeTag>::FluidState& fs,
+                                  const int perf) -> bool
+        {
+            const auto cell_press = this->extendEval(this->getPerfCellPressure(fs));
```
atgeirr, line 2261: Suggest using `getPerfCellPressure(fs).value()` to do this in doubles instead of dragging around AD objects needlessly here.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -2263,6 +2240,91 @@ namespace Opm
...
+                    this->computeConnLevelInjInd(fs, connPICalc, mob,
+                                                 connPI, deferred_logger);
+                }
+            }
+
+            addVector(connPI, wellPI);
+
+            ++subsetPerfID;
+            connPI += np;
+        }
```
atgeirr, line 2321: I think you should add an assert() that subsetPerfID now has the same value as `number_of_perforations_`. (Make it an `int` and initialize it before the loop.)

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -2263,6 +2240,91 @@ namespace Opm
...
+
+            const auto& fs = fluidState(subsetPerfID);
+
+            auto connPICalc = [&wellPICalc, allPerfID](const EvalWell& mobility) -> double
+            {
+                return wellPICalc.connectionProdIndStandard(allPerfID, mobility.value());
+            };
+
+            setToZero(connPI);
+            if (! isInjecting(fs, subsetPerfID)) {  // Production or zero flow rate
```
atgeirr, line 2305: I missed the `!` on first read, suggest swapping the order of the if and else blocks to make the `isInjecting() == true` clause come first.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -2263,6 +2240,91 @@ namespace Opm
...
+            if (allConn[allPerfID].state() == Connection::State::SHUT) {
+                continue;
+            }
+
+            std::vector<EvalWell> mob(num_components_, {numWellEq_ + numEq, 0.0});
+            getMobility(ebosSimulator, static_cast<int>(subsetPerfID), mob, deferred_logger);
+
+            const auto& fs = fluidState(subsetPerfID);
+
+            auto connPICalc = [&wellPICalc, allPerfID](const EvalWell& mobility) -> double
```
atgeirr, line 2299: Also here, use a double to simplify. Then pass value()s instead of extending the evaluation below such as line 4197.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -2263,6 +2240,91 @@ namespace Opm
...
+
+            setToZero(connPI);
+            if (! isInjecting(fs, subsetPerfID)) {  // Production or zero flow rate
+                if (allow_xflow || this->isProducer()) {
+                    this->computeConnLevelProdInd(fs, connPICalc, mob, connPI);
+                }
+            }
+            else {              // Connection injects into reservoir
+                if (allow_xflow || this->isInjector()) {
+                    this->computeConnLevelInjInd(fs, connPICalc, mob,
```
atgeirr, line 2312: Passing connPICalc as a connIICalc argument, is that ok?

atgeirr (commented):
I have some preliminary comments below, but have one overriding concern: These new functions seem to some extent to repeat the calculations done in computePerfRate(). Would it not be possible to call that function and reuse it? There are some details in that one that is missing here, such as the effect of skin pressure with the polymermw option, that I am not sure how should be treated though.

It seems to me that the code now (especially for injectors) repeats a lot of the logic from there.

I'll pause my review until this is clearer to me.

bska:
> These new functions seem to some extent to repeat the calculations done in computePerfRate(). Would it not be possible to call that function and reuse it?

The PI/II, in this formulation, is pretty much all of `cqs` **except** the pressure drawdown, with a possible logarithmic rescaling term to account for drainage radius being different from the well connection's equivalent radius.  We'd also need to remove the connection transmissibility factor from [wellProdIndexCalculator](https://github.com/OPM/opm-simulators/blob/73b384f8cd33920f4a5452f8ed2e546e5aea5245/opm/simulators/wells/WellProdIndexCalculator.cpp#L72) if we were to use the existing function to compute flow rates, but that's of course possible.

I can picture doing the other way around, though, by having both the `cqs` calculation and the PI/II call a common helper function to compute the PVT-related terms of `cqs` and then doing slightly different transformations to those values.

opm/simulators/wells/StandardWell_impl.hpp
bska: Okay, I'll do that.

opm/simulators/wells/StandardWell_impl.hpp
bska: Very good.  I'll do that.

opm/simulators/wells/StandardWell_impl.hpp
bska: > Passing `connPICalc` as a `connIICalc` argument, is that ok?

I suppose it might be confusing, but yes&mdash;in this context this is okay.  The calculator here is mainly responsible for incorporating logarithmic scaling factors if needed.

bska:
jenkins build this please

bska:
I have pushed an update which addresses most of the initial comments.  I have not, however, refactored the implementation to make PI/II and `cqs` share use a common helper.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
This is has been in the queue for quite time now and is blocking the development of `WELPI`.  May I request final review of the updates here please?

bska:
jenkins build this please

atgeirr (approved):
I think a follow-up should un-duplicate the implementation (currently duplicated between standard and multisegment wells), but other than that this is good. I'll merge.

bska:
> currently duplicated between standard and multisegment wells

It is definitely duplicated to some degree, but there are minute differences relating to constructing objects of types `EvalWell` that currently necessitate having separate implementations.  I'll think about what we can do to merge this as much as possible but I fear we're not going to get very far.

PR #2853, opened 2020-10-12
add python package to debian packaging

akva2:


akva2:
shamless self-merge. want to have it tested by nightlies..

PR #2854, opened 2020-10-12
Prevent global communication when shutting wells in prediction.

blattms:
Previously, we did one global sum for every well name, as we tried to collectively shut each failing well. That is a bit too much. Now we try to shut all failing wells at once locally and then only communicate the outcome afterwards. This might also fix some logging issues.


blattms:
jenkins build this please

blattms:
benchmark please

blattms:
Even though tests are green, I think there still might be  an issue in the code. Let me double check.

opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
```diff
@@ -529,14 +529,8 @@ namespace Opm {
...
-                            int num_shut_wells = 0;
-                            for (const auto& well : failing_wells) {
-                                bool was_shut = solver.model().wellModel().forceShutWellByNameIfPredictionMode(well, substepTimer.simulationTimeElapsed());
-                                if (was_shut) {
-                                    ++num_shut_wells;
-                                }
-                            }
-                            if (num_shut_wells == 0) {
+                            bool wells_shut = solver.model().wellModel().forceShutWellByNameIfPredictionMode(failing_wells, substepTimer.simulationTimeElapsed());
+                            if (wells_shut) {
```
atgeirr, line 533: I think this logic is flipped, we should chop if no wells were shut.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -177,37 +178,89 @@ namespace Opm {
...
                 }
-                break;
             }
+            ++index;
         }
 
         // Communicate across processes if a well was shut.
-        well_was_shut = ebosSimulator_.vanguard().grid().comm().max(well_was_shut);
+        const auto& comm = ebosSimulator_.vanguard().grid().comm();
+        auto no_wells_shut = shut_wells.size();
```
atgeirr, line 221: Suggest `num_wells_shut` to avoid confusion with "no wells were shut".

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.005    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.006    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.011    |



View result details @ https://www.ytelses.com/opm/?page=result&id=926

blattms:
Does anybody know which model/test  I need to run to have an effect of `BlackoilWellModel::forceShutWellByNameIfPredictionMode`?

blattms:
Could not find one (which means there might be skeletons in the closet). Maybe I try to change `MinTimeStepBeforeShuttingProblematicWellsInDays` option (it is so lovely short :wink:).

blattms:
Sorry, cannot manage that. I never seem to end up at [AdaptiveTimeSteppingEbos.hpp:519](/OPM/opm-simulators/blob/c059653fa0327317f82066dae16d79ba3bb9bb80/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp#L532)

atgeirr:
> Does anybody know which model/test I need to run to have an effect of `BlackoilWellModel::forceShutWellByNameIfPredictionMode`?

This requires a prediction case with a failing well. The only cases I know of that have that are real field cases. I guess that means we should create a small such case for the test suite. Tricky, though!

blattms:
Fixed the obvious issues, but this is basically untested. Unless somebody else does that, I would not merge. jenkins is only good to test compilation.

blattms:
jenkins build this please

OPMUSER:
Well why didn't you say you needed at run that does not work, I've got lots of them :-)

Try this one.

[SHALE-06-OPM.zip](https://github.com/OPM/opm-simulators/files/5370855/SHALE-06-OPM.zip)

And yes it works in the commercial simulator.

blattms:
You misunderstood me. I need a run that works with flow and uses the functionality
that I changed. Otherwise I cannot make sure that stuff still runs with my changes.

PR #2855, opened 2020-10-13
Sort local non-shut wells.

blattms:
This will change the order of applying wells. Most tests are unaffected by this (maybe because the wells are sorted anyway), others are (e.g. FAULTS_MODEL_1, and 9_4B_WINJ_VREP-W_STW, probably others, too).

This PR is there to demonstrate the minimal changes needed to change results

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -219,6 +219,11 @@ namespace Opm {
         auto w = schedule().getWells(timeStepIdx);
         globalNumWells = w.size();
         w.erase(std::remove_if(w.begin(), w.end(), is_shut_or_defunct_), w.end());
+        // sort to speed lookup of accompanying parallel_well_info
+        std::sort(w.begin(), w.end(), [](const auto& w1, const auto& w2){
+                                          return w1.name() < w2.name();
+                                      });
+
         return w;
     }
 


blattms:
jenkins build this please

blattms:
Ok, in the well model this reordering should have no effect as we reorder `ecl_wells_` and that seems to be consistently used to set up the rest of the well stuff.

Still, these discrepancies seem a bit too large to just come from a different order of applying the wells. Are we maybe somewhere assuming a certain ordering? I need to check that.

blattms:
If we run in parallel some kind of reordering should always happen. It is a bit surprising that it has no effect on the tests there. Maybe that is due to using other tolerances for comparing.

atgeirr:
> Still, these discrepancies seem a bit too large to just come from a different order of applying the wells. Are we maybe somewhere assuming a certain ordering? I need to check that.

As long as the wells_ecl_ constructed using the method modified here is the one we pass to all other parts of the code, such as well state init(), this should be ok. However, there is a call to schedule().getWellsatEnd() in addNeighbours(). I am not sure if that could cause trouble when the order becomes different?

blattms:
addNeighbours does not care about the ordering, it essentially/finally just uses the cell indices of the perforated cells.

my worries are more about output or how the regression tests work.

blattms:
Too stale now.

PR #2857, opened 2020-10-14
Fixes compilation of CUDA.

blattms:
nvcc exits compilation if the header dune/istl/basearray.hh (form DUNE 2.6) is included as it does not seem to understand the friend declaration there (friend class for a struct).

```
/usr/include/dune/istl/basearray.hh:101:49: error: ‘typename Dune::Imp::base_array_unmanaged<B, A>::RealIterator’ names ‘template<class B, class A> template<class T> struct Dune::Imp::base_array_unmanaged<B, A>::RealIterator’, which is not a type
       friend class RealIterator<const ValueType>;
                                                 ^
```

### DIFF
--- a/opm/simulators/linalg/bda/MultisegmentWellContribution.hpp
+++ b/opm/simulators/linalg/bda/MultisegmentWellContribution.hpp
@@ -26,7 +26,9 @@
 #include <cuda_runtime.h>
 #endif
 
-#include <dune/istl/umfpack.hh>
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please

blattms:
just testing with 2.7 and problem does not depend on DUNE version.

blattms:
Shameless selfmerge as it works for all supported DUNE versions.

PR #2858, opened 2020-10-15
Added cmdline parameter --ilu-reorder-strategy

Tongdongq:
Added commandline parameter to choose ilu reordering strategy for openclSolver.
Valid options are `level_scheduling` and `graph_coloring`. A `std::string` is passed down via `BdaBridge`, `openclSolver` to `BILU0`, where the variable is evaluated.
I could add shorthands like `ls` and `gc` as values, but these might not be clear enough.

The default value is set to `graph_coloring`, which is generally faster, but does introduce random behavior at this point. One way to make it deterministic is to set a constant seed in `Reorder.cpp`.

One thing to note is that this parameter only affects openclSolver. We might want to rename the parameter to reflect that.


atgeirr:
> One thing to note is that this parameter only affects openclSolver. We might want to rename the parameter to reflect that.

I support that. Parameters that are only used under some circumstances are confusing, but proper naming helps.

blattms:
I agree, that the name should be changed to indicate where this applies, How about `OpenclIluReordering`

I would also suggest to do the error checking in `FlowLinearSolverParameters::init` and use an enum class from there instead comparing the strings whenever the ILU0 is constructed.

Tongdongq:
This variable will also be used in a future fpgaSolver, but we can address that then.
`FlowLinearSolverParameters::init` contains no other error checking, it looks a bit out of place there.
An enum makes sense, but earlier it's used, the more includes we'll get.

blattms:
My fear is that throwing during construction of the preconditioner will be
interpreted as the linear solver not converging -> timestep chopping and so on.
But maybe I am wrong.

Please be sure to check what happens if we pass a unsupported value.

I do not understand the argument about the includes. You can e.g. put the
enum into ILU0.hpp and only need to include that somewhere. Or you can use
and int for passing.

BTW I will not insist on using an enum, int or string. (Might just be my
personal taste)

Tongdongq:
```
Error: [/home/tqiu/data/opm/official/opm/opm-simulators/opm/simulators/linalg/bda/BILU0.cpp:46] Error invalid argument for --ilu-reorder-strategy, usage: '--ilu-reorder-strategy=[level_scheduling|graph_coloring]'
Program threw an exception: [/home/tqiu/data/opm/official/opm/opm-simulators/opm/simulators/linalg/bda/BILU0.cpp:46] Error invalid argument for --ilu-reorder-strategy, usage: '--ilu-reorder-strategy=[level_scheduling|graph_coloring]'
```
I thought the enum was going to go into either `FlowLinearSolverParameters.hpp` or `BILU0.hpp`. Including either of those in the chain would include lots of other unnecessary stuff, but if it's put in a single file, this is no problem.

Tongdongq:
Apparently, if a type is used as a parameter, operators << and >> must be defined/overloaded. This means that two files must be defined, since definitions in the header will lead to 'multiple definition' errors. If that's not a problem I'll push it, otherwise I prefer passing a string over an int.

blattms:
If you need a cpp file than add one.

Tongdongq:
If I let `FlowLinearSolverParameters.hpp` use the enum, I have to overload the << and >> operators. I don't know how to do this, my way made flow freeze upon start.

blattms:
@Tongdongq Unfortunately, you need to rebase this on master because of merge conflicts.

blattms:
It does not compile either as the ILUReorder.cpp file is missing.

Tongdongq:
@blattms Can you review this again?

blattms:
THANKS for the reminder. I am sorry for forgetting about this PR. Will review it now.

blattms:
jenkins build this please

blattms (approved):
Looks good and works nicely on my machine. Will merge once jenkins is happy

blattms:
In it goes. thanks a lot for the work.

PR #2861, opened 2020-10-17
Well debug log

joakim-hove:
I need something like this for the restart work; and since I am quite ordinary developer it might be valuable for others as well?

This PR adds a very basic start for logging of well changes/updates. Currently the code is very basic and not by any means complete, but I post it here early so that it can be shot down and buried right away if this is not a wanted feature. Personally I like it of course.

I tried to strike a balance between zero (or veeery low) overhead in the default no-logging case, and not polluting the main code path too much. The intention is that developers working on this type of code should set a boolean flag in the code to `true` and recompile; in addition users can set the environment variable `OPM_DEBUG_WELLS="pattern"` - and the debugging will be turned on for all wells matching `pattern` - I think that could be a valuable tool for remote debugging assistance?

Feel free to slaughter the implementation all the way to minced meat - but fcous on the main question: "Do we want to merge something like this in the future - when the implementation has been fixed?"

### DIFF
--- a/opm/simulators/wells/WellInterface.hpp
+++ b/opm/simulators/wells/WellInterface.hpp
@@ -52,10 +52,14 @@
 #include <opm/material/densead/Math.hpp>
 #include <opm/material/densead/Evaluation.hpp>
 
-#include <string>
-#include <memory>
-#include <vector>
+#include <iostream>
+#include <fstream>
+
 #include <cassert>
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
 
 namespace Opm
 {
@@ -384,6 +388,8 @@ namespace Opm
 
         double wsolvent_;
 
+        mutable std::shared_ptr<std::ofstream> debug_stream_;
+
         const PhaseUsage& phaseUsage() const;
 
         int flowPhaseToEbosCompIdx( const int phaseIdx ) const;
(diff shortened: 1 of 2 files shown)


joakim-hove:
jenkins build this please

atgeirr:
My thoughts on this:
 - Instead of an optional file stream, can't we do this with the regular logging system (add an appropriate logging backend if some condition is fulfilled)?
 - I strongly prefer control by command line argument to control by environment variable.
 - Minor: the message "...changing to..." can be misleading, since the code here does not actually change it, it just detects that a constraint is broken. I suggest changing the message to just state that the constraint is broken.

joakim-hove:
> Instead of an optional file stream, can't we do this with the regular logging system (add an appropriate logging backend if some condition is fulfilled)?

Yes we can - I felt that the different logging levels did not really fit in; the big question is whether to log or not - but I can change to use a dedicated log backend.

joakim-hove:
> I strongly prefer control by command line argument to control by environment variable.

I am inclined to agree; currently I feel the commandline system is quite messy and have a slight wish to just stay out of it. But I am happy to change to use a commandline argument eventually.

joakim-hove:
> [...] I suggest changing the message to just state that the constraint is broken.

I don't object to that - but I would like to focus on the large things first.

PR #2863, opened 2020-10-19
Move 'gridForConn' Into Cuda/OpenCL Block

bska:
It is only used within this context and produces a warning of the form
```
ISTLSolverEbos.hpp:128:25: warning: unused variable ‘gridForConn’
```
unless the build configures accelerator support.


bska:
jenkins build this please

PR #2866, opened 2020-10-20
Initial Implementation of WELPI Feature

bska:
This PR adds an initial implementation of the `WELPI` feature.  At this time there are still issues that need ironing out, especially for injection wells, but I believe we're ready for more widespread testing.

We rely on actual PI values being calculated at the well level at the end of each time step and use those values, in the event of a `WELPI` request, to calculate CTF scaling factors and apply those factors to all subsequent editions of the well provided the connection factors are eligible for `WELPI`-based rescaling.

If we trigger a rescaling event we also reset the `WellState`'s internal copies of the CTFs and reinitialize the Well PI calculators to ensure the rescaling takes effect immediately.  Since we rely on PI values being available at the end of each time step we must also take care to forward those values from the `WellState` of one report step to the `WellState` of the next report step.

This PR depends on #2846 and must not be merged before that PR is merged into master.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -239,49 +239,15 @@ namespace Opm {
         wells_ecl_ = getLocalNonshutWells(timeStepIdx, globalNumWells);
 
         this->initializeWellProdIndCalculators();
-        initializeWellPerfData();
+        // The well state initialize bhp with the cell pressure in the top cell.
+        // We must therefore provide it with updated cell pressures
+        this->initializeWellPerfData();
+        this->initializeWellState(timeStepIdx, globalNumWells, summaryState);
 
         // Wells are active if they are active wells on at least
         // one process.
         wells_active_ = localWellsActive() ? 1 : 0;
         wells_active_ = grid.comm().max(wells_active_);
-
-        // The well state initialize bhp with the cell pressure in the top cell.
-        // We must therefore provide it with updated cell pressures
-        size_t nc = local_num_cells_;
-        std::vector<double> cellPressures(nc, 0.0);
-        ElementContext elemCtx(ebosSimulator_);
-        const auto& gridView = ebosSimulator_.vanguard().gridView();
-        const auto& elemEndIt = gridView.template end</*codim=*/0>();
-        for (auto elemIt = gridView.template begin</*codim=*/0>();
-             elemIt != elemEndIt;
-             ++elemIt)
-        {
-            const auto& elem = *elemIt;
-            if (elem.partitionType() != Dune::InteriorEntity) {
-                continue;
-            }
-            elemCtx.updatePrimaryStencil(elem);
-            elemCtx.updatePrimaryIntensiveQuantities(/*timeIdx=*/0);
-
-            const unsigned cellIdx = elemCtx.globalSpaceIndex(/*spaceIdx=*/0, /*timeIdx=*/0);
-            const auto& intQuants = elemCtx.intensiveQuantities(/*spaceIdx=*/0, /*timeIdx=*/0);
-            const auto& fs = intQuants.fluidState();
-            // copy of get perfpressure in Standard well
-            // exept for value
-            double perf_pressure = 0.0;
-            if (Indices::oilEnabled) {
-                perf_pressure = fs.pressure(FluidSystem::oilPhaseIdx).value();
-            } else {
-                if (Indices::waterEnabled) {
-                    perf_pressure = fs.pressure(FluidSystem::waterPhaseIdx).value();
-                } else {
-                    perf_pressure = fs.pressure(FluidSystem::gasPhaseIdx).value();
-                }
-            }
-            cellPressures[cellIdx] = perf_pressure;
-        }
-        well_state_.init(cellPressures, schedule(), wells_ecl_, timeStepIdx, &previous_well_state_, phase_usage_, well_perf_data_, summaryState, globalNumWells);
 
         // handling MS well related
         if (param_.use_multisegment_well_&& anyMSWellOpenLocal()) { // if we use MultisegmentWell model
@@ -650,6 +616,51 @@ namespace Opm {
             }
             ++well_index;
         }
+    }
+
+
+
+
+
+    template<typename TypeTag>
+    void
+    BlackoilWellModel<TypeTag>::
+    initializeWellState(const int           timeStepIdx,
+                        const int           globalNumWells,
+                        const SummaryState& summaryState)
+    {
+        std::vector<double> cellPressures(this->local_num_cells_, 0.0);
+        ElementContext elemCtx(ebosSimulator_);
+
+        const auto& gridView = ebosSimulator_.vanguard().gridView();
+        const auto& elemEndIt = gridView.template end</*codim=*/0>();
+        for (auto elemIt = gridView.template begin</*codim=*/0>();
+             elemIt != elemEndIt;
+             ++elemIt)
+        {
+            if (elemIt->partitionType() != Dune::InteriorEntity) {
+                continue;
+            }
+
+            elemCtx.updatePrimaryStencil(*elemIt);
+            elemCtx.updatePrimaryIntensiveQuantities(/*timeIdx=*/0);
+
+            const auto& fs = elemCtx.intensiveQuantities(/*spaceIdx=*/0, /*timeIdx=*/0).fluidState();
+
+            // copy of get perfpressure in Standard well except for value
+            double& perf_pressure = cellPressures[elemCtx.globalSpaceIndex(/*spaceIdx=*/0, /*timeIdx=*/0)];
+            if (Indices::oilEnabled) {
+                perf_pressure = fs.pressure(FluidSystem::oilPhaseIdx).value();
+            } else if (Indices::waterEnabled) {
+                perf_pressure = fs.pressure(FluidSystem::waterPhaseIdx).value();
+            } else {
+                perf_pressure = fs.pressure(FluidSystem::gasPhaseIdx).value();
+            }
+        }
+
+        well_state_.init(cellPressures, schedule(), wells_ecl_, timeStepIdx,
+                         &previous_well_state_, phase_usage_, well_perf_data_,
+                         summaryState, globalNumWells);
     }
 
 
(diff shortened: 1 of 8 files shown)


bska:
Together with OPM/opm-common#2031 this PR provides sufficient functionality to run a reduced test dataset with comparable CTF output to ECLIPSE.  Some example plots are provided below.  `0B_WELPI_MODEL2` was run with ECLIPSE while `0B_WELPI_MODEL2_FLOW` was run with this PR and OPM/opm-common#2031.

### CTF in Connection (10,4,4) of Producer Well PROD2 ###

PROD2_Completion_10_4_4_Connection_Transmissibility_Factor

### CTF in Connection (6,19,10) of Producer Well PROD3 ###

PROD3_Completion_6_19_10_Connection_Transmissibility_Factor

### CTF in Connection (2,13,11) of Injection Well INJ1 ###

INJ1_Completion_2_13_11_Connection_Transmissibility_Factor

### CTF in Connection (12,20,9) of Injection Well INJ2 ###

INJ2_Completion_12_20_9_Connection_Transmissibility_Factor

### Water Phase Well Level PI in Producer Well PROD1 ###

PROD1_Water_phase_PI

### Gas Phase Well Level PI in Producer Well PROD2 ###

PROD2_Gas_phase_PI

### Oil Phase Well Level PI in Producer Well PROD3 ###

PROD3_Oil_phase_PI

bska:
jenkins build this please

bska:
jenkins build this please opm-common=2031 please

bska:
jenkins build this opm-common=2064 please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

jalvestad:
I have tested this pullrequest together with pr2846 for a comprehensive set of WELPI data for both production and injection wells. The tests have covered the different ways of using the WELPI keyword. These two pull-requests work correctly for all tests performed and are therefore judged to work according to the specification in the manual - and as Eclipse works!

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
Marking this as ready for review now that the upstream PR #2846 has been merged into the master branch.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
I'm converting this back to draft status.  PR #2960 must be merged first.

bska:
Marking this as "ready for review" now that #2960 has been merged into master.

bska:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2539,6 +2555,127 @@ namespace Opm {
...
 
 
+
+
+    template<typename TypeTag>
+    void
+    BlackoilWellModel<TypeTag>::
+    runWellPIScaling(const int timeStepIdx, DeferredLogger& local_deferredLogger)
+    {
+        if (this->last_run_wellpi_.has_value() && (*this->last_run_wellpi_ == timeStepIdx)) {
```
GitPaean, line 2565: let me confirm here. If we use the keyword `WELPI`, do we do this for each following report steps?

The manual is not very clear to me, I got the impression it is a one-time thing when reading it. 

do we handle the situation that connections change (added new connections or connection tranmissibility changes with new input?)

bska: Static `WELPI`s&mdash;i.e., those that are listed as separate keywords and not triggered as the result of a dynamic action (`ACTIONX`)&mdash;act as follows

1. At the start of the report step for which `WELPI` is specified, compute the dynamic PI value for the pertinent well's "preferred" phase
2. Use this dynamic PI value to compute a scaling factor as `WPI(2) / dynamic_PI`
3. Apply that scaling factor to the transmissibility factor of every connection in the well, whether they are open or not.  Moreover, this scaling should persist until the the transmissibility factor is explicitly reset in COMPDAT.

The last operation is handled by [`Schedule::applyWellProdIndexScaling()`](https://github.com/OPM/opm-common/blob/0102d736acead804ee843852f62f32cf5d85333a/src/opm/parser/eclipse/EclipseState/Schedule/Schedule.cpp#L1545).  This PR is about handling steps 1 and 2.

> do we handle the situation that connections change (added new connections or connection tranmissibility changes with new input)?

Yes.  That's built into `Schedule::applyWellProdIndexScaling()`.

> Under these circumstances, the well PI will not be the one specified in WELPI.

That is correct.  `WELPI` is an instantaneous scaling based on the flow conditions of a single phase.  The dynamic PI values will typically drift away from the requested value fairly soon after the scaling has been applied, but will start from that new value.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2591,6 +2728,61 @@ namespace Opm {
...
+    assignWellGuideRates(data::Wells& wsrpt) const
+    {
+        for (const auto& well : this->wells_ecl_) {
+            auto xwPos = wsrpt.find(well.name());
+            if (xwPos == wsrpt.end()) { // No well results.  Unexpected.
+                continue;
+            }
+
+            auto& grval = xwPos->second.guide_rates;  grval.clear();
+            grval += this->getGuideRateValues(well);
```
GitPaean, line 2747: Is `+=` intentional here?

bska: > Is `+=` intentional here?

Yes.  We `.clear()` the value on the previous line.

atgeirr: That is not very visible, please put the clear() on a separate line.

GitPaean: I know it is probably equivalent to `=`.  Just wondering if there is extra reasons to use `+=`?

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2539,6 +2555,127 @@ namespace Opm {
...
+                    ? pi[pu.phase_pos[BlackoilPhases::PhaseIndex::Liquid]]
+                    : 0.0;
+
+            case Phase::GAS:
+                return pu.phase_used[BlackoilPhases::PhaseIndex::Vapour]
+                    ? pi[pu.phase_pos[BlackoilPhases::PhaseIndex::Vapour]]
+                    : 0.0;
+
+            default:
+                throw std::invalid_argument {
```
GitPaean, line 2603: I am not updated with this. Are we supposed to use `DeferredLogger`  facility for this? @atgeirr ?

GitPaean (approved):
This is very fine PR. From code side, I did not spot anything problematic. I assume the results are well tested and checked.

GitPaean:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
atgeirr: I do not think we must use it here, but the calling code should ideally try-catch, communicate failure etc. However it is very annoying to use such elaborate things for a simple "missing enum value" type check, perhaps we should rather return an `std::optional` here if it goes wrong instead?

bska: > perhaps we should rather return an `std::optional` here if it goes wrong instead?

Then the question becomes what to do if there is no defined value in the `optional`.  In that case, there really is no well-defined way to continue the simulation since the input model requests something that we cannot support.

opm/simulators/wells/BlackoilWellModel_impl.hpp
bska: I've pushed an update here.  I'd forgotten that [`GuideRateValue`](https://github.com/OPM/opm-common/blob/968aa79468ddef5e901538b8d09d2b46b011b379/opm/output/data/GuideRateValue.hpp#L31) supports direct assignment.

GitPaean:
jenkins build this please

GitPaean:
One question here, did we add a running test for `WELPI`? Is it a good time to add `0B_WELPI_MODEL2.DATA` to the running test? @bska

bska:
> Is it a good time to add `0B_WELPI_MODEL2.DATA` to the running test?

I have one more update coming in this area (supporting `WELPI` for shut/stopped wells).  When that's in I'll add a regression test.

GitPaean:
If I read correctly, the current implementation assumes the productivity index has the same formulation for StandardWell and MultisegmentWell, is it something verified with simulation?

bska:
> the current implementation assumes the productivity index has the same formulation for StandardWell and MultisegmentWell,

That is correct.

> is it something verified with simulation?

Not really&mdash;we don't currently have a test model that combines multi-segment wells wit WELPI.  On the other hand, the main distinction between standard wells and multisegmented wells is the way they model well-bore pressure drops and drawdowns.  Those factors don't go into the PI calculation.

GitPaean:
> Not really—we don't currently have a test model that combines multi-segment wells wit WELPI. On the other hand, the main distinction between standard wells and multisegmented wells is the way they model well-bore pressure drops and drawdowns. Those factors don't go into the PI calculation.

Thanks for clarifying.  I will think for multisegment wells, the more complicated interaction through the wellbores, the frictional forces and accelerational forces, and also all kinds of effects from ICDs, will make the productivity index formulation different, at least in theory.  It will not be a simple summation through the connections anymore.  But in practice, what is used in E100 is another different issue.  That is why I asked whether it is tested and verified.

bska:
> the more complicated interaction through the wellbores, the frictional forces and accelerational forces, and also all kinds of effects from ICDs, will make the productivity index formulation differen

That's decidedly possible.  I'm also going to suggest we deal with that when we get a model that demonstrates the difference.

GitPaean:
>  I'm also going to suggest we deal with that when we get a model that demonstrates the difference.

Yes. That surely will work and that is something I am really happy if we found out. At least, using the same formulation might give decent results for many situations, just like many cases, StandardWell and MSW gives basically the same results.

GitPaean:
Jenkins is happy. Merging as promised.

PR #2867, opened 2020-10-21
changed: run tests on 4 processors

bska:
more than 4 is problematic on some build machines


bska:
Self-merging in my capacity as OPM 2020.10 Release Manager.

PR #2871, opened 2020-10-23
changed: compile each flow simulator as a object library

akva2:
increases parallelism

@atgeirr check if this works better for you.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,9 +229,14 @@ set_property(TARGET moduleVersion PROPERTY POSITION_INDEPENDENT_CODE ON)
 # but this is not exposed in a super-build.
 add_dependencies(moduleVersion opmsimulators)
 
-add_library(flow_libblackoil OBJECT flow/flow_ebos_blackoil.cpp)
+set(FLOW_TGTS)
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

CMakeLists.txt
```diff
@@ -272,7 +267,7 @@ opm_add_test(flow_blackoil
   LIBRARIES opmsimulators
   SOURCES
   flow/flow.cpp
-  flow/flow_ebos_blackoil.cpp
+  $<TARGET_OBJECTS:flow_libblackoil>
```
atgeirr, line 270: Was not this target removed?

atgeirr (commented):
My compile time was reduced, although not quite as much as I hoped. I still think the change has merit on its own, as it simplifies the cmake code.

I have a question regarding the flow_libblackoil target.

CMakeLists.txt
akva2: Nope. It's added in the loop (now there's one such target for each flow simulator).

atgeirr: Thanks, now I got it!

PR #2874, opened 2020-10-26
 WIP Add flow_dyn

totto82:
Adds a dynamic flow version that only takes number of equations as parameter

The goal is to reduce the number of flow versions i.e reduce compile time

The drawback is a slight overhead in the run time. Initial testing indicates
1) Extra cost of using dynamic mappings and if evaluations < 1% for norne
2) Extra cost of adding enableSolvent unconditionally for norne. ~5%. The solvent code is never called and the number of equation is still 3, but the intensive quantities size is increased to carry solvent related stuff. 
3) Extra cost of using dynamic flow for oil-water olympus. ~2%..  The dynamic flow currently always allocating three phases in the fluid state. 

This should neither alter the results nor the runtime for flow.

Some optimization could potentially further reduce the overhead associated with the dynamic blackoil indices code.

Depends on OPM/opm-models#631

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ option(SIBLING_SEARCH "Search for other modules in sibling directories?" ON)
 set( USE_OPENMP_DEFAULT OFF ) # Use of OpenMP is considered experimental
 option(BUILD_FLOW "Build the production oriented flow simulator?" ON)
 option(BUILD_FLOW_BLACKOIL_ONLY "Build the production oriented flow simulator only supporting the blackoil model?" OFF)
+option(BUILD_FLOW_DYN "Build the flow simulator with dynamic indicies?" OFF)
 option(BUILD_FLOW_VARIANTS "Build the variants for flow by default?" OFF)
 option(BUILD_EBOS "Build the research oriented ebos simulator?" ON)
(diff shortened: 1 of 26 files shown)


joakim-hove:
I kind of like the current compile time?!

totto82:
> I kind of like the current compile time?!

It may not be worth the extra complication and overhead in the code, but the ever increasing number of flow version worried me enough to test an alternative.  I know some colleagues have some more versions in making and also at TNO they are planning more. 

The suggested approach allows for both compile time and dynamic versions so we can get the best from both words.

joakim-hove:
> The suggested approach allows for both compile time and dynamic versions so we can get the best from both words.

Jokes in writing is always difficult; I am positive to any suggestion which can reduce the current compile time!

alfbr:
Cool, I need to let it sink in a bit before fully realizing the implications, but definitely positive from the outset :-)

bska:
@totto82 : If this work is still relevant, how much effort would it be to rebase the patch onto the current master and make the changes necessary&mdash;if any&mdash;to make it run

totto82:
I will close the PR. It was for mostly discussion and testing. I think the bullet 1) above should be carried out. i.e. to avoid compile time determination if it is a gas-water, oil-water or a oil-gas case. The rest should probably not change.

PR #2875, opened 2020-10-26
Extract Correct Connection-Level Rates

bska:
The Polymer, Brine, and Solvent quantities would be extracted from elements `0..#perf-1` of their pertinent container rather than from the elements associated to the particular well.


bska:
Merging into release branch in my role as release manager.

PR #2878, opened 2020-10-28
[testing] clearing well_container_ at timeStepSucceeded()

GitPaean:
it should not be usable for next time step. It caused confusion related to other development. 

Not sure it is doable, so it is testing.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -508,6 +508,8 @@ namespace Opm {
 
         previous_well_state_ = well_state_;
 
+        this->well_container_.clear();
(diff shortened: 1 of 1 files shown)


GitPaean:
jenkins build this please

GitPaean:
The failures (if confirmed) are possibly because that well models are used for some purposes after `timeStepSucceeded` or `endTimeStep`, it can be group control, UDQ, THP ..... At the same time, many other tests are not affected. 

Most likely, we should be able to correct the failures, hopefully the right thing to do.

GitPaean:
jenkins build this please

GitPaean:
for testing.

GitPaean:
jenkins build this please

PR #2884, opened 2020-10-30
Solve the well model twice as a preprocessing step

totto82:
After the first solve the explicit quantities are updated and the well model solved again
This options was earlier only used at the initial step

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -81,8 +81,6 @@ namespace Opm {
 
         gravity_ = ebosSimulator_.problem().gravity()[2];
 
-        initial_step_ = true;
-
         // add the eWoms auxiliary module for the wells to the list
         ebosSimulator_.model().addAuxiliaryModule(this);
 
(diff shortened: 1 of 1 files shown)


blattms:
out of curiosity: Does this result in better convergence/performance, etc.?

totto82:
> out of curiosity: Does this result in better convergence/performance, etc.?

On norne the effect on runtime is small. For larger cases I don't know. The motivation for the change is to get more accurate results. But It should not hurt for the convergence either I think.

alfbr:
Thanks, will test!

alfbr:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.005    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.005    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.016    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.005    |



View result details @ https://www.ytelses.com/opm/?page=result&id=941

PR #2886, opened 2020-10-30
add scotch(par)metis to debian packaging

bska:


bska:
Merging in my capacity as release manager for the OPM 2020.10 release.

PR #2897, opened 2020-11-03
Switch default for restarting to read historic schedule

alfbr:
Unfortunately restarting without reading historic schedule is a long term effort. In order to avoid excessive problems for users, I suggest we switch the default until we are closer to support restarting without historic schedule available.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -138,7 +138,7 @@ struct EclStrictParsing<TypeTag, TTag::EclBaseVanguard> {
 };
 template<class TypeTag>
 struct SchedRestart<TypeTag, TTag::EclBaseVanguard> {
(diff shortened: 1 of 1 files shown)


alfbr:
jenkins build this please

joakim-hove:
Hmm; seems jenkins was a bit reluctant to take, I'll kick him again and then merge when he approves **but**:

As one of the people working on restart without the historical schedule section I can confirm that this is a complex task. Up until now the thinking has been that we have two alternatives:

1. Restart reading all schedule information from the historical schedule file.
2. Restart reading where the schedule information is assembled from the restart file.

The initial opm implementation has been of type 1, and we are currently working to reach alternative. I think maybe we will have to settle for a best effort middle ground.

joakim-hove:
jenkins build this please

PR #2898, opened 2020-11-03
Always accumulates data in AMG to 1 process on coarse levels.

### LINKED ISSUE #2894: Deadlock/ MPI message truncation when using CPR and AMG
running norne with the current rc of flow from the Ubuntu packages with 16 processes I am experiencing the following:
1. using `--linsolver=amg` an exception is thrown:
```
Restart file written for report step:   0/247.  Date: 1997/11/06
[galadriel:4255] *** An error occurred in MPI_Alltoallv
[galadriel:4255] *** reported by process [21692417,6]
[galadriel:4255] *** on communicator MPI_COMM_WORLD
[galadriel:4255] *** MPI_ERR_TRUNCATE: message truncated
[galadriel:4255] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[galadriel:4255] ***    and potentially your MPI job)
[galadriel:04239] 2 more processes have sent help message help-mpi-errors.txt / mpi_errors_are_fatal
[galadriel:04239] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```
2. using using `--linsolver=cpr` flow deadlocks (most probably in the setup phase of the first linear solve).

This should not hold back the release, but the parallel versions using CPR/AMG should be marked as experimental.

blattms:
In OPM the matrix graph might be unsymmetric as we do not store the full sparsity pattern for copy rows but only the diagonal.
Unfortunately, DUNE assumes that matrices from finite elements and finite volumes have a symmetric sparsity pattern for copy rows to and uses this assumption to create the graphs for PTScotch/ParMETIS more easily. But PTScotch/ParMetis assume a symmetric graph.

Closes #2894 for me.

There are other possible (and better) fixes mentioned in #2894. But this one seemed the quickest and should be good enough for the release (Reverting OPM/opm-common#2061 would not cut it for ParMETIS).

### DIFF
--- a/opm/simulators/linalg/PreconditionerFactory.hpp
+++ b/opm/simulators/linalg/PreconditionerFactory.hpp
@@ -134,6 +134,10 @@ private:
         criterion.setNoPreSmoothSteps(prm.get<int>("pre_smooth", 1));
         criterion.setNoPostSmoothSteps(prm.get<int>("post_smooth", 1));
         criterion.setDebugLevel(prm.get<int>("verbosity", 0));
+        // As the default we request to accumulate data to 1 process always as our matrix
+        // graph might be unsymmetric and hence not supported by the PTScotch/ParMetis
+        // calls in DUNE. Accumulating to 1 skips PTScotch/ParMetis
+        criterion.setAccumulate(static_cast<Dune::Amg::AccumulationMode>(prm.get<int>("accumulate", 1)));
         return criterion;
     }
 
--- a/opm/simulators/linalg/setupPropertyTree.cpp
+++ b/opm/simulators/linalg/setupPropertyTree.cpp
@@ -61,6 +61,10 @@ setupCPR(const std::string& conf, const FlowLinearSolverParameters& p)
     prm.put("preconditioner.coarsesolver.preconditioner.verbosity", 0);
     prm.put("preconditioner.coarsesolver.preconditioner.maxlevel", 15);
     prm.put("preconditioner.coarsesolver.preconditioner.skip_isolated", 0);
+    // We request to accumulate data to 1 process always as our matrix
+    // graph might be unsymmetric and hence not supported by the PTScotch/ParMetis
+    // calls in DUNE. Accumulating to 1 skips PTScotch/ParMetis
+    prm.put("preconditioner.coarsesolver.preconditioner.accumulate", 1);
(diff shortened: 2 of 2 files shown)


blattms:
jenkins build this please

opm/simulators/linalg/PreconditionerFactory.hpp
```diff
@@ -134,6 +134,10 @@ class PreconditionerFactory
         criterion.setNoPreSmoothSteps(prm.get<int>("pre_smooth", 1));
         criterion.setNoPostSmoothSteps(prm.get<int>("post_smooth", 1));
         criterion.setDebugLevel(prm.get<int>("verbosity", 0));
+        // As the default we request to accumulate data to 1 process always as our matrix
+        // graph might be unsymmetric and hence not supported by the PTScotch/ParMetis
+        // calls in DUNE. Accumulating to 1 skips PTScotch/ParMetis
+        criterion.setAccumulate(static_cast<Dune::Amg::AccumulationMode>(prm.get<int>("accumulate", 1)));
```
atgeirr, line 140: Which accumulation mode is "1" cast to? I assume it is atOnceAccum? If so, does that mean that we go from N to 1 process immediately on level 2, or that we coarsen several levels on N processes, and then go to 1 process? I will assume the second option, which seems the most reasonable. This would be the same as the "no PTScotch/ParMETIS at all" way, I think?

Perhaps we should use a string parameter instead of an integer?

atgeirr (approved):
Seems ok to me. We should however try to find a better solution that allows redistribution. How large would the changes be to accomplish this?

blattms:
Better alternative coming.

opm/simulators/linalg/PreconditionerFactory.hpp
blattms: It casts to atOnceAccu as the other would trigger the same problem (pass an unsymmetric graph to a software assuming a symmetric one).

totto82:
Thanks this fix me problem as well.

atgeirr:
Merged as per discussion in #2894

atgeirr:
Should be backported to the release.

bska:
> Should be backported to the release.

I'll backport.  And create RC-5.

PR #2899, opened 2020-11-04
Make AMG use the graph of the communication pattern for agglomeration.

### LINKED ISSUE #2894: Deadlock/ MPI message truncation when using CPR and AMG

blattms:
The default in DUNE is to use the graph of the matrix to compute a coarser partitioning when building the hierarchy of matrices.
(This done to keep the ratio of communication versus computation reasonable as the matrices get coarser and coarser).

The sparsity pattern of our matrix might be unsymmetric due to not storing offdiagonals for the ghost rows. DUNE assumes a
symmetric sparsity pattern to prepare the graph for PTScotch/ParMETIS (without too much communication/computation).
Hence our approach breaks the assumption and PTScotch/ParMETIS might deadlock or MPI will error out in some MPI_Allgatherv calls.

Fortunately, there is another approach implemented that uses the graph of the communication happening when applying the linear operator. In this graph there is one vertex for each MPI rank and an edge between v and w (and w and v) exists if these
exchange messages during the application of the operator. This graph is constructed within AMG and always symmetric.

With this commit our build system defines AMG_REPART_ON_COMM_GRAPH to instruct AMG to use the graph of the communication pattern for loadbalancing.

Closes #2894 
Might be a better alternative than #2898

### DIFF
--- a/opm-simulators-prereqs.cmake
+++ b/opm-simulators-prereqs.cmake
@@ -13,7 +13,20 @@ set (opm-simulators_CONFIG_VAR
   DUNE_ISTL_VERSION_MINOR
   DUNE_ISTL_VERSION_REVISION
   HAVE_SUITESPARSE_UMFPACK
+  AMG_REPART_ON_COMM_GRAPH
   )
+
+# the sparsity pattern of our matrix might be unsymmetric due to
+# not storing offdiagonals for the ghost rows. DUNE assumes a
+# symmetric sparsity pattern to prepare the graph for
+# PTScotch/ParMETIS. Hence our approach breaks the assumption and
+# PTScotch/ParMETIS might deadlock or MPI will error out
+# in some MPI_Allgatherv calls.
+# We define AMG_REPART_ON_COMM_GRAPH to instruct AMG to use
+# the graph of the communication patter (one vertex for each
+# MPI rank and edge between v and w exists if these exchange
+# messages)
+set(AMG_REPART_ON_COMM_GRAPH 1)
 
 # dependencies
 set (opm-simulators_DEPS


blattms:
Works at least on my system,

blattms:
jenkins build this please

atgeirr:
This sounds good, but I do not really understand what it does. If the graph used has just one vertex per MPI rank, how can it rebalance? Should we not have one vertex per coarse node/vertex/row (not sure what term I should use)?

blattms:
Let nparts be the number of parts to rebalance to, mpi_size the number of mpi ranks, and nv the number of vertices. Then the following needs to hold:

- nparts <= mpi_size (to be reasonable nparts < mpi_size, for AMG nparts will be aroung mpi_size/8
- nv >= nparts

atgeirr:
I'll hold off merging this until we have tested a bit more, and perhaps pinned down the cause of the lack of symmetry. But I consider it likely to be merged eventually.

totto82:
@atgeirr @blattms Is this something we want in the release?

blattms:
We have fixed the bug with the alternative approach #2898 (which is IMHO a bit less scalable). I will not fight for this. @atgeirr seemed to plan investigations. Maybe they have been performed?

If not maybe we maybe should not merge it just before the release as there might be unexpected flaws. Unless somebody will test on parallel runs.

atgeirr:
I have not been able to do extensive testing of this. But we should keep it alive, the stopgap fix of #2898, I agree, is less scalable.

> If not maybe we maybe should not merge it just before the release as there might be unexpected flaws. Unless somebody will test on parallel runs.

I agree with this as well, let's not merge this for the release.

PR #2901, opened 2020-11-05
Always accumulates data in AMG to 1 process on coarse levels.

bska:
In OPM the matrix graph might be unsymmetric as we do not store the full sparsity pattern for copy rows but only the diagonal. Unfortunately, DUNE assumes that matrices from finite elements and finite volumes have a symmetric sparsity pattern for copy rows to and uses this assumption to create the graphs for PTScotch/ParMETIS more easily. But PTScotch/ParMetis assume a symmetric graph.


bska:
Merging into release branch in my capacity as 2020.10 release manager.

PR #2902, opened 2020-11-05
Fixes to the wlift implementation 

totto82:
Small fixes for the well gas lift implementation.

### DIFF
--- a/opm/simulators/wells/GasLiftRuntime_impl.hpp
+++ b/opm/simulators/wells/GasLiftRuntime_impl.hpp
@@ -35,14 +35,13 @@ GasLiftRuntime(
     const Simulator &ebos_simulator,
     const SummaryState &summary_state,
     DeferredLogger &deferred_logger,
-    std::vector<double> &potentials,
     const WellState &well_state,
     const Well::ProductionControls &controls
 ) :
     controls_{controls},
     deferred_logger_{deferred_logger},
     ebos_simulator_{ebos_simulator},
-    potentials_{potentials},
+    potentials_(well_state.numPhases(),0.0),
     std_well_{std_well},
     summary_state_{summary_state},
     well_state_{well_state},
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

alfbr:
This greatly improves the accuracy of the gas lift optimization. There are still some differences, but they are minor compared to what this fixes. Great if we can get this in.

opm/simulators/wells/GasLiftRuntime_impl.hpp
```diff
@@ -354,8 +372,14 @@ runOptimizeLoop_(bool increase)
         // NOTE: if BHP is below limit, we set state.stop_iteration = true
         auto bhp = state.getBhpWithLimit();
         computeWellRates_(bhp, cur_potentials);
-        auto new_oil_rate = getOilRateWithLimit_(cur_potentials);
-        auto new_gas_rate = getGasRateWithLimit_(cur_potentials);
+        double new_oil_rate = 0.0;
+        bool oilIsLimited = getOilRateWithLimit_(new_oil_rate, cur_potentials);
```
hakonhagland, line 376: I would suggest using snake case style for the variables, since that has been used for all other variables in this file. So instead of `oilIsLimited` use `oil_is_limited`. And similar for `gasIsLimited` below.

hakonhagland:
@totto82  Thanks,  this looks good.

hakonhagland (changes requested):

totto82:
jenkins build this please

opm/simulators/wells/GasLiftRuntime_impl.hpp
totto82: done

totto82:
Jenkins, @alfbr @hakonhagland is happy. I will self-merge.

PR #2903, opened 2020-11-05
Two fixes in StandardWell_impl.hpp

plgbrts:
1) Corrected phaseIsActive with PhaseIdx argument instead of CompIdx argument in "subtraction of dissolved gas from oil phase and vapporized oil from gas phase".
2) Fix for well accumulation calculation in case oil is absent.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -269,12 +269,10 @@ namespace Opm
                 return primary_variables_evaluation_[SFrac];
             }
         }
-        else if (FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx) && compIdx == Indices::canonicalToActiveComponentIndex(FluidSystem::waterCompIdx)) {             
-
+        else if (FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx)) {             
             if (FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx) && compIdx == Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx)) {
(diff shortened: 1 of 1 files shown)


goncalvesmachadoc:
jenkins build this please

totto82 (approved):

totto82:
Thanks for the fix.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -269,12 +269,10 @@ namespace Opm
                 return primary_variables_evaluation_[SFrac];
             }
         }
-        else if (FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx) && compIdx == Indices::canonicalToActiveComponentIndex(FluidSystem::waterCompIdx)) {             
-
+        else if (FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx)) {             
             if (FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx) && compIdx == Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx)) {
                 return primary_variables_evaluation_[GFrac];
             }
```
atgeirr, line 275: I was about to comment on this: Should there not be a way to return `primary_variables_evaluation_[WFrac]` here? I.e. the line deleted below should have been kept? @totto82 please comment since you merged.

totto82: I thought so too. But after carefully reading I concluded that this was correct. The WFRAC is not used for gas-water simulations. The water fraction is 1-primary_variables_evaluation(gfrac) See below

atgeirr: Thanks. I see it now.

PR #2907, opened 2020-11-06
Improve error message for problematic NNCEDIT

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/2095


joakim-hove:
jenkins build this please

ebos/nncsorter.cpp
```diff
@@ -48,33 +53,49 @@ std::vector<Opm::NNCdata> sortNncAndApplyEditnnc(const std::vector<Opm::NNCdata>
...
+        auto j = cell % cartDims[1];
+        auto k = cell / cartDims[1];
+
+        return fmt::format("({},{},{})", i + 1,j + 1,k + 1);
+    };
+
+
+    auto make_warning = [&format_ijk] (const Opm::KeywordLocation& location, const Opm::NNCdata& nnc) -> std::string {
+        return fmt::format("Problem with EDITNNC keyword\n"
+                           "In {} line {} \n"
```
bska, line 68: Tiny nit: Why the trailing blank before the newline?

joakim-hove: No reason - removed

joakim-hove:
I will polish this - not the least git wise; but together with the upstream PR's it is "essentially" ready from my hand. Review comments appreciated.

ebos/eclwriter.hh
```diff
@@ -603,8 +593,11 @@ private:
 
         // Cartesian index mapper for the serial I/O grid
         const auto& equilCartMapper =  simulator_.vanguard().equilCartesianIndexMapper();
-
         const auto& cartDims = simulator_.vanguard().cartesianIndexMapper().cartesianDimensions();
+        auto nncCompare =  []( const Opm::NNCdata& nnc1, const Opm::NNCdata& nnc2){
+                               return nnc1.cell1 < nnc2.cell1 ||
+                                      ( nnc1.cell1 == nnc2.cell1 && nnc1.cell2 < nnc2.cell2);};
```
bska, line 599: Since we're moving this around anyway, maybe we could switch to implementing `operator<` in terms of `tuple/tie`?
```C++
auto nncCompare = [](const Opm::NNCData nnc1, const Opm::NNCData& nnc2)
{
    return std::tie(nnc1.cell1, nnc1.cell2) < std::tie(nnc2.cell1, nnc2.cell2);
};
```

bska (approved):
Looks good to me.  We could use `operator<` for `std::tuple` to implement `nncCompare` slightly more simply and robustly though.  Other than that, this is good to go once the commit history is cleaned up.

ebos/eclwriter.hh
joakim-hove: I thought of that - the reason I did not do it was that I was a bit uncertain about the expected semantics from the `trans` member.

alfbr:
benchmark please

ytelses:
Benchmark result overview: 




View result details @ https://www.ytelses.com/opm/?page=result&id=955

alfbr:
It did not build. Just wanted to double check that now performance penalty snuck in.

joakim-hove:
> It did not build

No - it requires two upstream PR's

PR #2910, opened 2020-11-09
[WIP] Numerical aquifer

GitPaean:
This is the downstream of the PR OPM/opm-common#2097 and https://github.com/OPM/opm-grid/pull/502. 

It did two things here, one is to use the cell depth specified from the numerical aquifer, the other is to initialize the water saturation in the numerical aquifer cells to be unity.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -2660,13 +2660,6 @@ private:
             auto& elemFluidState = initialFluidStates_[elemIdx];
             elemFluidState.assign(equilInitializer.initialFluidState(elemIdx));
         }
-        const auto& aquifer_cells = simulator.vanguard().eclState().aquifer().numericalAquifers().aquiferCells();
-
-        const auto& vanguard = simulator.vanguard();
-        size_t numCartDof = vanguard.cartesianSize();
-        std::vector<int> cartesianToCompressedElemIdx(numCartDof, -1);
-        for (unsigned elemIdx = 0; elemIdx < numElems; ++elemIdx)
-            cartesianToCompressedElemIdx[vanguard.cartesianIndex(elemIdx)] = elemIdx;
     }
 
     void readEclRestartSolution_()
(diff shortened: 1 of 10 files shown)


ebos/equil/initstateequil.hh
```diff
@@ -1881,24 +1884,44 @@ private:
...
             };
 
+            const size_t global_index = UgGridHelpers::globalCell(grid)[cell];
+
+            if (aquifer.hasCell(global_index)) {
+                const auto& aqu_cells = aquifer.aquiferCells();
+                const auto& aqu_cell = aqu_cells.at(global_index);
+                pos = CellPos {
+                    cell, aqu_cell.depth
+                };
```
GitPaean, line 1912: It looks like we do repeated works here related to cell_depth. Maybe it is good if we can use the results from `elementCenterDepth_` from `eclproblem`.

ebos/eclproblem.hh
```diff
@@ -2272,6 +2273,15 @@ private:
...
+            if (ecl_state.aquifer().hasNumericalAquifer()) {
+                const auto& num_aquifer = ecl_state.aquifer().numericalAquifers();
+                const unsigned int global_index = vanguard.cartesianIndex(elemIdx);
+                if (num_aquifer.hasCell(global_index)) {
+                    const auto& cell = num_aquifer.aquiferCells().at(global_index);
+                    elementCenterDepth_[elemIdx] = cell.depth;
+                }
+            }
+
         }
```
totto82, line 2285: I suggest we move the elementCenterDepth to the vanguard

See https://github.com/OPM/opm-simulators/pull/2975

ebos/equil/initstateequil.hh
totto82: I plat to get https://github.com/OPM/opm-simulators/pull/2941 merged soon. That may affect you work here. It should be trivial to rebase it but just so that you are aware.

bska:
@GitPaean : Is this PR still relevant or was it replaced by #3134 (and others)?

GitPaean:
It should not be relevant anymore.  I am closing it now.

bska:
> It should not be relevant anymore. I am closing it now.

Very good.  What about this PR's upstream companions OPM/opm-common#2097 and OPM/opm-grid#502?  I suppose those could be closed too or do they have changes that you'd still like to include?

PR #2912, opened 2020-11-10
Make a flow version that uses the polyhedreal grid

totto82:
Currently the simulator creates the polyhedreal grid from an eclGrid from opm-common

TODO (but not in this PR)
 - make it possible to create the grid directly from DGF or MRST format
 - fix issue on norne.
 - add a test case to make sure this continues to work

Depends on OPM/opm-grid#500


akva2:
Update https://github.com/OPM/opm-common/blob/master/jenkins/build-opm-module.sh#L6 to have it built on jenkins.

totto82:
jenkins build this opm-grid=500 please

akva2:
jenkins build this opm-common=2103 opm-grid=500 please

blattms:
This is very cool. Finally we might have a means to better test the grid interface.

blattms (approved):

totto82:
Thanks for the review. I will merge this later today if no further comments.

PR #2920, opened 2020-11-16
Pass region in place values to the summary writer

joakim-hove:
This is a step towards implementing support for the summary keyword `ROEW`, to calculate these values we need access to the original in place values for the different regions. These values must be passed from the simulator to the summary output code. 

The interface for passing non trivial values from the simulator to the summary output is not in a very good shape, this PR contains minimum steps in order to support the ROEW keyword.

Requires: https://github.com/OPM/opm-common/pull/2114

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -44,6 +44,7 @@
 #include <opm/output/data/Cells.hpp>
 #include <opm/output/eclipse/EclipseIO.hpp>
 #include <opm/common/OpmLog/OpmLog.hpp>
+#include <opm/output/eclipse/Inplace.hpp>
 
 #include <dune/common/fvector.hh>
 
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this opm-common=2114 please

blattms:
For dummies like me:
Why is unordered_map better/needed?
I could not find the answer in the description or commit message.

joakim-hove:
> Why is unordered_map better/needed?

The point is that it should be runtime configurable what quantities to calculate, and then a map is more convenient to work with than `std::array` - which is probably the least flexible data structure in STL. As mentioned in the PR description the ambition for this PR is the minumum amount of changes to support the `ROEW` summary keyword, so full runtime configurability will not be achieved with this PR.

joakim-hove:
Along with the accompanying PR in opm-common this should now be ready for review. The essence of this PR is to use the new class `Inplace` to manage the aggregated region properties. The PR work is done in a minimum-required-change way, the `eclblackoiloutputmodule` could probably benefit from further refactoring - hopefully the changes in this PR are at least in a good direction.

When this is merged the plan is to continue with a PR passing the intial inplace volumes/properties to the summary output and then implement the `ROEW` summary keyword - that should be quite simple when this is merged. 

The current code still uses region values as indices in a vector, that is in my opinion a prime index <-> ID misunderstanding and I have a real itch to change the code to use a map. I have waited with that to keep the amount of change down.

joakim-hove:
jenkins build this opm-common=2114 please

joakim-hove:
Comments to this - or the upstream:  OPM/opm-common#2114

joakim-hove:
OK - I'll give this a final Jenkins round and then I intend to merge this - unless .... ;-)

joakim-hove:
jenkins build this opm-common=2114 please

PR #2923, opened 2020-11-17
throw if well has no connection. 

totto82:
Wells without connection should be shut at this point
See OPM/opm-common#2115

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -730,6 +730,9 @@ namespace Opm {
                     wellIsStopped = true;
                 }
 
+                if ( well_perf_data_[w].size() == 0)
+                    OPM_THROW(std::runtime_error, "Well has no connections: " + well_ecl.name());
+
                 // Use the pvtRegionIdx from the top cell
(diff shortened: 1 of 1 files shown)


blattms:
Please do not merge this without further discussion. 

For distributed wells we will have exactly this situation not matter what the schedule looks like. This due to determining the distribution once using all possible connections (using getWellsatEnd()). Depending on the time step all local connections might
be closed (while non-local ones are open somewhere else(.

alfbr:
I can confirm that this fixes our segmentaion fault problem.

alfbr:
Actually, it seems I was too fast. As far as I can see the fix is in opm-common. The throw implemented here does not seem to come into effect. Sorry for the confusion.

totto82:
This only adds a guard against wells with empty connection sets. After discussing with @joakim-hove we decided to handle the empty connections set in the parsing. 

Exchanging the throw with 
```

        well_state_.shutWell(w);
                    continue;
```

will make the simulator handle the case by not including them in the well container. @blattms  is this how you plan to do the distribution?

totto82:
Code like 
```
        // pick the density in the top layer
        const double rho = perf_densities_[0];
```
in StandardWells_impl#3061 does not play well with wells without connections.

PR #2925, opened 2020-11-17
add python3-numpy to debian packaging

akva2:
required after newly introduced test


akva2:
shameless self-merge to (hopefully) fix nightly build.

PR #2926, opened 2020-11-17
Use consisten pvt region index for groups in parallel.

### LINKED ISSUE #2921: pvt region index in BlackoilWellModel::checkGroupHigherConstraints inconsistent in parallel run.
In parallel this is much worse. Not only might the pvt region computed have no association to the wells of the group, but may also be different from process to process as there might be no local wells and if there are the first one might not be the same well
[BlackoilWellModel_impl.hpp#L2390](https://github.com/OPM/opm-simulators/blob/0dcfc6036339f1b9fa10100db21666acea4d37d4/opm/simulators/wells/BlackoilWellModel_impl.hpp#L2390) 

~There is a very problematic data access at [BlackoilWellModel_impl.hpp#L2390](https://github.com/OPM/opm-simulators/blob/0dcfc6036339f1b9fa10100db21666acea4d37d4/opm/simulators/wells/BlackoilWellModel_impl.hpp#L2390) where we do
`pvt_region_idx_[well_perf_data_[0][0].cell_index]`. The problem is that well_perf_data has an entry for each global well, but if the
well at index 0 is not in this process's then `well_perf_data_[0]` is an empty vector.~

I would fix this, but it is very misty what the original code is trying to do. At least it is not clear to me, why anyone would query a well without even testing if that well belongs to the group we are computing for.

This really stalls me and I need help.

blattms:
In serial we use the first cell of the first well to determine the pvt region index for a group. Previously, we used the first cell of
the first local well in a parallel run. Unfortunately that may lead to different pvt region indices being used for the same goup on
different processes.

We fix this by using the same approach in parallel as we already use in serial. For this we usedWell::seqIndex() to determine the needed ordering.

Closes #2921 (at least the reported parallel inconsistency, not whether the index used makes sense).

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -2384,10 +2384,27 @@ namespace Opm {
         // TODO fix this!
         // This is only used for converting RESV rates.
         // What is the proper approach?
+        const auto& comm = ebosSimulator_.vanguard().grid().comm();
         const int fipnum = 0;
-        const int pvtreg = well_perf_data_.empty()
+        int pvtreg = well_perf_data_.empty()
             ? pvt_region_idx_[0]
             : pvt_region_idx_[well_perf_data_[0][0].cell_index];
+
+        if ( comm.size() > 1)
+        {
+            // Just like in the sequential case the pvtregion is determined
+            // by the first cell of the first well. What is the first well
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

blattms:
Would be cool if this was merged soon.

blattms:
jenkins build this please

atgeirr (approved):
Looks fine, can be merged when green.

blattms:
In it goes.

PR #2927, opened 2020-11-18
Disable python tests in debian packaging

akva2:
add cmake option to disable the python tests
    
Due to bugs in the openmpi on bionic, this test fails to execute properly in pbuilder environments. 
Instead of rebuilding openmpi without dynamic loading (which is the suggested fix) and potentially break users
systems, this is a non-intrusive workaround to be used for packaging.
    
Also add explicit option for python support to make it visible in cmake frontends.


akva2:
jenkins build this please

blattms:
Unrelated, but I am curious:
Have they fixed the other problems that we had in previous releases (that force us rebuilding openmpi)?

akva2:
yes, that's fixed in focal (or well, rather, it was fixed in debian 10).

akva2:
jenkins build this please

akva2:
can this be merged?

bska:
> can this be merged?

Yes.  I'll merge this now.

PR #2934, opened 2020-11-19
StandardWell only assing primary_variables_[WQTotal] once.

blattms:
There seems to be no need to assign the same value consecutively as often as there are phases. To the very least it is confusing.


blattms:
jenkins build this please

atgeirr:
Thanks for noticing. Merging.

PR #2935, opened 2020-11-20
Pass the initial inplace values to Summary::eval()

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/2122

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -1866,6 +1866,10 @@ public:
     const std::map<std::pair<std::string, int>, double>& getBlockData()
     { return blockData_; }
 
+    const Opm::Inplace& initialInplace() const {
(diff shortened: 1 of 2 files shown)


ebos/ecloutputblackoilmodule.hh
```diff
@@ -1866,6 +1866,10 @@ public:
     const std::map<std::pair<std::string, int>, double>& getBlockData()
     { return blockData_; }
 
+    const Opm::Inplace initialInplace() const {
```
bska, line 1869: Is this function supposed to return a reference to the contained object or a copy of that object?  In the latter case, it should typically not return a `const` object.

joakim-hove: Ahhh - `&` here and a `&` there; it should be a const ref. Will fix

PR #2939, opened 2020-11-23
Fix well mixture for single-phase

goncalvesmachadoc:


goncalvesmachadoc:
jenkins build this please

atgeirr:
Thanks for the patch, it should be merged eventually. However I have a few requests first.
You have structured the code like this:
```
...
            } else {
                 std::fill(mix.begin(), mix.end(), 0.0);
                 if (num_comp == 1){
                     mix[num_comp-1] = 1.0;
                 }else{
                  // No flow => use well specified fractions for mix.
...
           } // new brace
...
```
The part before and after are unchanged, but the part after has moved one indentation level in, obscuring the fact that there are no changes to that part. I suggest rather the following:
```
...
            } else if (num_comp == 1) {
                     mix[num_comp-1] = 1.0;
            } else {
                 std::fill(mix.begin(), mix.end(), 0.0);
                // No flow => use well specified fractions for mix.
               ...
```
That way there will be no change to the part after here, and it will be easier to see what the commit does.

Some nitpicks:
 - The title for the PR is misleading, please change it. Perhaps "Fix well mixture for single-phase"?
 - Adapt code style to the surrounding code. In general, prefer to have space around if, else, for, while.
 - Make sure you use 4-space indentation.

goncalvesmachadoc:
jenkins build this please

atgeirr:
Good, merging!

PR #2941, opened 2020-11-25
dunify and make the initialization parallel

totto82:
Tested on norne and it gives the same initial values in parallel and in serial. Should be tested on more models.

### DIFF
--- a/ebos/eclequilinitializer.hh
+++ b/ebos/eclequilinitializer.hh
@@ -100,7 +100,7 @@ public:
 
         EQUIL::DeckDependent::InitialStateComputer<TypeTag> initialState(materialLawManager,
                                                                          eclState,
-                                                                         vanguard,
+                                                                         vanguard.gridView(),
                                                                          simulator.problem().gravity()[dimWorld - 1]);
 
         // copy the result into the array of initial fluid states
--- a/ebos/eclequilinitializer.hh
+++ b/ebos/eclequilinitializer.hh
@@ -100,7 +100,7 @@ public:
 
         EQUIL::DeckDependent::InitialStateComputer<TypeTag> initialState(materialLawManager,
                                                                          eclState,
-                                                                         vanguard.grid(),
+                                                                         vanguard,
                                                                          simulator.problem().gravity()[dimWorld - 1]);
 
         // copy the result into the array of initial fluid states
--- a/ebos/equil/initstateequil.hh
+++ b/ebos/equil/initstateequil.hh
@@ -1410,89 +1410,29 @@ invertCapPress(const double   pc,
 
 // ===========================================================================
 
-template <typename Grid, typename CellRange>
-void verticalExtent(const Grid&           grid,
-                    const CellRange&      cells,
+template <typename CellRange, typename Comm>
+void verticalExtent(const CellRange&      cells,
+                    const std::vector<std::pair<double, double>>& cellZMinMax,
+                    const Comm& comm,
                     std::array<double,2>& span)
 {
-    // This code is only supported in three space dimensions
-    assert (Grid::dimensionworld == 3);
-
     span[0] = std::numeric_limits<double>::max();
     span[1] = std::numeric_limits<double>::lowest();
-
-    const int nd = Grid::dimensionworld;
 
     // Define vertical span as
     //
     //   [minimum(node depth(cells)), maximum(node depth(cells))]
     //
-    // Note: We use a sledgehammer approach--looping all
-    // the nodes of all the faces of all the 'cells'--to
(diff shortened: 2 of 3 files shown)


totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
The test failures on MOD4_GRP is expected since this PR uses the average of the four corners of the top/bottom cell and not the centroid of the face to determine the cells vertical span. I can go back to using the face centroid, but I think the average of the corners is more aligned with how we compute cell centers etc in the code. Any opinion?

totto82:
When I compare the initial results in resInsight they are exactly the same. I will rerun jenkins to make sure I don't run after ghosts.

totto82:
jenkins build this please

ebos/eclequilinitializer.hh
```diff
@@ -100,7 +100,7 @@ public:
 
         EQUIL::DeckDependent::InitialStateComputer<TypeTag> initialState(materialLawManager,
                                                                          eclState,
-                                                                         vanguard.grid(),
+                                                                         vanguard,
```
blattms, line 103: I would appreciate it if we would not pass the vanguard here, but the gridView. I seems like we only need to access the view itseld, the grid (accessible via `gridView.grid()` and the communicator (accessible via `gridView.grid().comm()`.

blattms (approved):
I generally approve of this but would prefer to not use the vanguard, if possible,

ebos/eclequilinitializer.hh
totto82: I agree. I will update.

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
The initial results of the test cases that fails are the same. i.e. the difference = 0 for all cells when compared in ResInsight. When comparing the simulations I notice a difference in number of newton iterations after a few time step that leads to slight change of timesteps that further leads to difference in when the simulator has problems to converge. A visual inspection of the results shows agreement. IMO it is safe to do a update_data and merge this.

totto82:
@atgeirr @bska Any comments? Great if we can get this in soon.

bska (commented):
Is there a reason you're copying all the `Geometry` objects?  Also, `inline` is superfluous for templates.  Those are implicitly inline.

totto82:
> Is there a reason you're copying all the Geometry objects? Also, inline is superfluous for templates. Those are implicitly inline.

Thanks for noticing. I have pushed a fix.

totto82:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

totto82:
This needs an update data. I have verified that the test failures are only numerical noise.

bska (approved):
Looks good to me.  Note that the requisite update to the reference solutions is going to conflict with the existing update for #2949 (i.e., OPM/opm-tests#455).  One of these has to go in first and then the other will likely have to be redone.

totto82:
Lets get #2949 in first. I will fix the conflict and update

totto82:
@bska Are https://github.com/OPM/opm-common/pull/2119 and https://github.com/OPM/opm-common/pull/2164 ready for merging? Or can I try to get this in first?

bska:
> can I try to get this in first?

I think we should merge this PR first.  It's been ready for a while and the other two aren't approved yet.  I'll generate updated reference solutions.

bska:

bska:
jenkins build this opm-tests=458 please

bska:
New reference data is installed on the CI system.  I'm merging this now to bring the system back to a consistent state.  Thanks a lot for the new feature and for your patience.

ebos/equil/initstateequil.hh
```diff
@@ -1773,6 +1770,28 @@ private:
...
+        cellCenterDepth_.resize(numElements);
+        cellZSpan_.resize(numElements);
+        cellZMinMax_.resize(numElements);
+
+        auto elemIt = gridView.template begin</*codim=*/0>();
+        const auto& elemEndIt = gridView.template end</*codim=*/0>();
+        for (; elemIt != elemEndIt; ++elemIt) {
+            const Element& element = *elemIt;
+            const unsigned int elemIdx = elemMapper.index(element);
+            cellCenterDepth_[elemIdx] = Details::cellCenterDepth(element);
```
GitPaean, line 1790: One question here, why not use the `cellCenterDepth_` from the `vanguard` here? Thanks. @totto82

GitPaean: Never mind. I thought without this can simply some work for me. It is not case.

PR #2919, opened 2020-11-16
Do not filter connections on the schedule of the loadbalanced grid.

blattms:
It would remove perforated cells from wells that cross the local domain's border. That would make it impossible to figure out the
first connection. In addition we would not be able to check that the connections exist (as rank 0 would have the complete information -> inconsistency).

This is cherry-picked from the work on distributed wells.

The filtering removes all connections from wells in the schedule that are not in the local domain of the process. It is only performed on processes with rank greater than zero (as zero needs all the information for output?). As currently a process either has all or no possible connections of a well in it's domain and the wells that have none are not treated anyway, this change should have no effect.

PR #2945, opened 2020-11-26
Preparations for distributed parallel standard wells.

blattms:
This is the status of that work. IMHO this should be merged if it does not drastically change the results. There are a lot of people working on the well part currently and rebasing has quite some toll now.

It does not yet activate the distributed wells in the sense that these may appear in a parallel run. But most of the infrastructure should be there now.

Surprisingly iteration counts and well curves change a little but this does not seem serious to me. It could still point to a bug though.

This include #2919 which has been hanging around for quite some while now. Hopefully this one will not.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -40,6 +40,7 @@ list (APPEND MAIN_SOURCE_FILES
   opm/simulators/utils/DeferredLogger.cpp
   opm/simulators/utils/gatherDeferredLogger.cpp
   opm/simulators/utils/ParallelRestart.cpp
+  opm/simulators/wells/ParallelWellInfo.cpp
   opm/simulators/wells/VFPProdProperties.cpp
   opm/simulators/wells/VFPInjProperties.cpp
   opm/simulators/wells/WellGroupHelpers.cpp
@@ -90,6 +91,7 @@ list (APPEND TEST_SOURCE_FILES
   tests/test_norne_pvt.cpp
   tests/test_wellprodindexcalculator.cpp
   tests/test_wellstatefullyimplicitblackoil.cpp
+  tests/test_parallelwellinfo.cpp
   )
 
 if(MPI_FOUND)
@@ -235,6 +237,7 @@ list (APPEND PUBLIC_HEADER_FILES
   opm/simulators/wells/MSWellHelpers.hpp
   opm/simulators/wells/BlackoilWellModel.hpp
   opm/simulators/wells/BlackoilWellModel_impl.hpp
+  opm/simulators/wells/ParallelWellInfo.hpp
   )
 
 list (APPEND EXAMPLE_SOURCE_FILES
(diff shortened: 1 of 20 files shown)


blattms:
jenkins build this please

bska:
> IMHO this should be merged if it does not drastically change the results. There are a lot of people working on the well part currently and rebasing has quite some toll now.

Are you then suggesting that we more or less blindly merge a 1500 line patch touching 20 files and let those other people deal with the fallout?

atgeirr:
> This include #2919 which has been hanging around for quite some while now.

I'll review that now at least!

blattms:
> Are you then suggesting that we more or less blindly merge a 1500 line patch touching 20 files and let those other people deal with the fallout?

Nope that is not what I wanted to suggest. Nevermind, I actually found the cause for the slightly different iteration counts and well curves (different rounding errors) and will provide another PR (first part of this one) with this and the failing tests fixed.

blattms:
Please merge #2947 first as it fixes the differences in the well curves experienced.

As I managed to get around sorting the wells the from ecl, rebasing this (and fixing the one test that needs an MPI fixture) should make the tests go through (unfortunately hard to check without jenkins) and should not change any timestepping (anymore). This PR does not activate distributing the wells and checking jenkins and the well curves should be enough to warrant that we did not break anything.

Be aware that there is some serious refactoring in the reporting of wellsin commit .7072280db We need to distinguish between the rates/stuff reported per well and the one reported per connection, as the latter might be distributed.

If requested, I can break this up in fewer PRs, but then the additional work needs to result in merges in a timely manner. Otherwise this becomes unmanageable. 

Currently, the commits are rather structured, only try to change one thing at a time and I tried to have explaining commit message tht explain why a change is needed. (Expception is the ParallelWellInfo interface which is changed rather late and has a lot of implications). Hence maybe reviewing be them is an alternative?

What makes this work complicated and cumbersome is that the actual testing of the changes has to be on a modified version of OPM (that skips ZOLTAN to produce distributed wells on 8 processes). Hence I need to test and maintain twice.

Anyway I need an answer here soon: Leave it as one PR or make multiple (e.g. nearly one per commit)? I can live with both, but the latter is considerably more work (and probably safer). Once I have an answer I will start the rebasing and continue the work.

atgeirr:
> Anyway I need an answer here soon: Leave it as one PR or make multiple (e.g. nearly one per commit)? I can live with both, but the latter is considerably more work (and probably safer). Once I have an answer I will start the rebasing and continue the work.

I think it is very good that the PR is split into commits that may be considered one by one instead of requiring a monolithic merge. I think a good way to leverage that is to split into smaller PRs. In order to get timely merges, you can nag by github and email! (As long as the order of merging etc. is clear.)

It is not clear to me if the current code gives identical results to current master when the wells are NOT distributed? If/when we are at that point it will be more acceptable to do a monolithic merge.

blattms:
> It is not clear to me if the current code gives identical results to current master when the wells are NOT distributed? If/when we are at that point it will be more acceptable to do a monolithic merge.

Once this is rebased on #2947 it should give identical results (I did a bisection for that yesterday).

blattms:
There is of course one downside of many small PRs. Reverting the change will be a lot more work.

If distributed wells do not pay off, we might want to do this to reduce complexity again.

blattms:
Closing as I factored out stuff from various commits. Please refer to #2947 and following

PR #2936, opened 2020-11-21
WIP: Open well halfways in report step

joakim-hove:
Testcase: https://github.com/OPM/opm-tests/pull/425

The ACTIONX keyword is evaluated at the end of every *timestep* - when an ACTIONX evaluates to true changing the status of a well is a quite common thing to do, but as I understand it the well structures in the simulator are mainly designed to change status for report steps.

This PR is just to encourage simulator/well specialists to think of the situation.

Failing tests: https://ci.opm-project.org/job/opm-simulators-PR-builder/1598/

PR #2947, opened 2020-11-26
Prepares for apply distributed standard wells.

blattms:
These are the first few commits of PR #2945 with a fix for the slight deviations in the well curves.

These were due to doing 
```C++
    Y temp(y);
    B.mv(x, temp);
    y -= temp;
```
instead of `B.mmv(x,y)`.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -416,6 +416,16 @@ public:
         int outputInterval = EWOMS_GET_PARAM(TypeTag, int, EclOutputInterval);
         if (outputInterval >= 0)
             schedule().restart().overrideRestartWriteInterval(outputInterval);
+
+        // Initialize parallelWells with all local wells
+        const auto& schedule_wells = schedule().getWellsatEnd();
+        parallelWells_.reserve(schedule_wells.size());
+
+        for (const auto& well: schedule_wells)
+        {
+            parallelWells_.emplace_back(well.name(), true);
+        }
+        std::sort(parallelWells_.begin(), parallelWells_.end());
     }
 
     /*!
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -217,11 +217,23 @@ namespace Opm {
     template<typename TypeTag>
     std::vector< Well >
     BlackoilWellModel<TypeTag>::
-    getLocalNonshutWells(const int timeStepIdx, int& globalNumWells) const
+    getLocalNonshutWells(const int timeStepIdx, int& globalNumWells)
     {
         auto w = schedule().getWells(timeStepIdx);
         globalNumWells = w.size();
         w.erase(std::remove_if(w.begin(), w.end(), is_shut_or_defunct_), w.end());
+        local_parallel_well_info_.clear();
+        local_parallel_well_info_.reserve(w.size());
+        for (const auto& well : w)
+        {
+            auto wellPair = std::make_pair(well.name(), true);
+            auto pwell = std::lower_bound(parallel_well_info_.begin(),
+                                          parallel_well_info_.end(),
+                                          wellPair);
+            assert(pwell != parallel_well_info_.end() &&
+                   *pwell == wellPair);
+            local_parallel_well_info_.push_back(&(*pwell));
+        }
         return w;
     }
 
@@ -802,6 +814,7 @@ namespace Opm {
         const auto& perf_data = this->well_perf_data_[wellID];
 
         return std::make_unique<WellType>(this->wells_ecl_[wellID],
+                                          *local_parallel_well_info_[wellID],
                                           time_step,
                                           this->param_,
                                           *this->rateConverter_,
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -54,16 +54,18 @@ namespace Opm {
                                     cartDims[0]*cartDims[1]*cartDims[2]);
         auto& parallel_wells = ebosSimulator.vanguard().parallelWells();
         parallel_well_info_.assign(parallel_wells.begin(), parallel_wells.end());
-        is_shut_or_defunct_ = [this, &ebosSimulator](const Well& well) {
+        const auto& pwell_info = parallel_well_info_;
+        std::size_t numProcs = ebosSimulator.gridView().comm().size();
+        is_shut_or_defunct_ = [&pwell_info, numProcs](const Well& well) {
                 if (well.getStatus() == Well::Status::SHUT)
                     return true;
-                if (ebosSimulator.gridView().comm().size() == 1)
+                if (numProcs == 1u)
                     return false;
                 std::pair<std::string, bool> value{well.name(), true}; // false indicate not active!
-                auto candidate = std::lower_bound(parallel_well_info_.begin(),
-                                                  parallel_well_info_.end(),
+                auto candidate = std::lower_bound(pwell_info.begin(),
+                                                  pwell_info.end(),
                                                   value);
-                return candidate == parallel_well_info_.end() || *candidate != value;
+                return candidate == pwell_info.end() || *candidate != value;
             };
 
         alternative_well_rate_init_ = EWOMS_GET_PARAM(TypeTag, bool, AlternativeWellRateInit);
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -217,14 +217,21 @@ namespace Opm {
     template<typename TypeTag>
     std::vector< Well >
     BlackoilWellModel<TypeTag>::
-    getLocalNonshutWells(const int timeStepIdx, int& globalNumWells)
+    getLocalNonshutWells(const int timeStepIdx, int& globalNumWells) const
     {
         auto w = schedule().getWells(timeStepIdx);
         globalNumWells = w.size();
         w.erase(std::remove_if(w.begin(), w.end(), is_shut_or_defunct_), w.end());
-        local_parallel_well_info_.clear();
-        local_parallel_well_info_.reserve(w.size());
-        for (const auto& well : w)
+        return w;
+    }
+
+    template<typename TypeTag>
+    std::vector< ParallelWellInfo* >
+    BlackoilWellModel<TypeTag>::createLocalParallelWellInfo(const std::vector<Well>& wells)
+    {
+        std::vector< ParallelWellInfo* > local_parallel_well_info;
+        local_parallel_well_info.reserve(wells.size());
+        for (const auto& well : wells)
         {
             auto wellPair = std::make_pair(well.name(), true);
             auto pwell = std::lower_bound(parallel_well_info_.begin(),
@@ -232,9 +239,9 @@ namespace Opm {
                                           wellPair);
             assert(pwell != parallel_well_info_.end() &&
                    *pwell == wellPair);
-            local_parallel_well_info_.push_back(&(*pwell));
+            local_parallel_well_info.push_back(&(*pwell));
         }
-        return w;
+        return local_parallel_well_info;
     }
 
     template<typename TypeTag>
@@ -250,6 +257,7 @@ namespace Opm {
         int globalNumWells = 0;
         // Make wells_ecl_ contain only this partition's non-shut wells.
         wells_ecl_ = getLocalNonshutWells(timeStepIdx, globalNumWells);
+        local_parallel_well_info_ = createLocalParallelWellInfo(wells_ecl_);
 
         this->initializeWellProdIndCalculators();
         initializeWellPerfData();
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -52,17 +52,18 @@ namespace Opm {
         const auto& cartDims = Opm::UgGridHelpers::cartDims(grid);
         setupCartesianToCompressed_(Opm::UgGridHelpers::globalCell(grid),
                                     cartDims[0]*cartDims[1]*cartDims[2]);
-
-        is_shut_or_defunct_ = [&ebosSimulator](const Well& well) {
+        auto& parallel_wells = ebosSimulator.vanguard().parallelWells();
+        parallel_well_info_.assign(parallel_wells.begin(), parallel_wells.end());
+        is_shut_or_defunct_ = [this, &ebosSimulator](const Well& well) {
                 if (well.getStatus() == Well::Status::SHUT)
                     return true;
                 if (ebosSimulator.gridView().comm().size() == 1)
                     return false;
                 std::pair<std::string, bool> value{well.name(), true}; // false indicate not active!
-                const auto& parallel_wells = ebosSimulator.vanguard().parallelWells();
-                auto candidate = std::lower_bound(parallel_wells.begin(), parallel_wells.end(),
-                                                 value);
-                return candidate == parallel_wells.end() || *candidate != value;
+                auto candidate = std::lower_bound(parallel_well_info_.begin(),
+                                                  parallel_well_info_.end(),
+                                                  value);
+                return candidate == parallel_well_info_.end() || *candidate != value;
             };
 
         alternative_well_rate_init_ = EWOMS_GET_PARAM(TypeTag, bool, AlternativeWellRateInit);
--- a/opm/simulators/wells/ParallelWellInfo.cpp
+++ b/opm/simulators/wells/ParallelWellInfo.cpp
@@ -45,7 +45,8 @@ void ParallelWellInfo::DestroyComm::operator()(Communication* comm)
 ParallelWellInfo::ParallelWellInfo(const std::string& name,
                                    bool hasLocalCells)
     : name_(name), hasLocalCells_ (hasLocalCells),
-      isOwner_(true), comm_(new Communication(Dune::MPIHelper::getLocalCommunicator()))
+      isOwner_(true), rankWithFirstPerf_(-1),
+      comm_(new Communication(Dune::MPIHelper::getLocalCommunicator()))
     {}
 
 ParallelWellInfo::ParallelWellInfo(const std::pair<std::string,bool>& well_info,
@@ -61,6 +62,16 @@ ParallelWellInfo::ParallelWellInfo(const std::pair<std::string,bool>& well_info,
     comm_.reset(new Communication(Dune::MPIHelper::getLocalCommunicator()));
 #endif
     isOwner_ = (comm_->rank() == 0);
+}
+
+void ParallelWellInfo::communicateFirstPerforation(bool hasFirst)
+{
+    int first = hasFirst;
+    std::vector<int> firstVec(comm_->size());
+    comm_->allgather(&first, 1, firstVec.data());
+    auto found = std::find_if(firstVec.begin(), firstVec.end(),
+                              [](int i) -> bool{ return i;});
+    rankWithFirstPerf_ = found - firstVec.begin();
 }
 
 bool operator<(const ParallelWellInfo& well1, const ParallelWellInfo& well2)
--- a/opm/simulators/wells/ParallelWellInfo.cpp
+++ b/opm/simulators/wells/ParallelWellInfo.cpp
@@ -18,6 +18,7 @@
 */
 #include <config.h>
 #include <opm/simulators/wells/ParallelWellInfo.hpp>
+#include <opm/common/ErrorMacros.hpp>
 
 namespace Opm
 {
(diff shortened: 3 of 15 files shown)


blattms:
Let's see whether this fixed the jenkins tests.

blattms:
jenkins build this please

blattms:
Yay, jenkins is green.

blattms:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.001    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1        |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.008    |



View result details @ https://www.ytelses.com/opm/?page=result&id=977

blattms:
jenkins build this please

atgeirr:
I will review this early next week. (Monday if time permits.)

blattms:
Highly appreciated as this needs to go in first before the rest. If you are too busy then I am sure you know some capable guys for taking over this task.

I did split some more commits to add some code here that only touches the new files. This also makes the new classes for complete and clean.

blattms:
jenkins build this please

bska (commented):
The parallel aspects of this are above my head, but I noticed a couple of minor issues in the more trivial parts.

opm/simulators/wells/ParallelWellInfo.cpp
```diff
@@ -0,0 +1,325 @@
...
+    bool missingCells = false;
+    auto start = foundConnections_.begin();
+    for(auto conFound = start; conFound != foundConnections_.end(); ++conFound)
+    {
+        if (*conFound == 0)
+        {
+            const auto& completion = well_.getConnections()[conFound - start];
+            msg = msg + " " + std::to_string(completion.getI()) + "," +
+                std::to_string(completion.getJ()) + ","
+                + std::to_string(completion.getK()) + " ";
```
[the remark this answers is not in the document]
blattms: I tend to agree, but I guess this needs to be decided for the whole module. Currently, we are not always doing that, see [BlackoilWellModel_impl.hpp:L625-L633](/OPM/opm-simulators/blob/ced3c99e09e6fd4fcfe0d3baabd4dc9644aa99f5/opm/simulators/wells/BlackoilWellModel_impl.hpp#L625-L633), which is consistent with this one.

opm/simulators/wells/ParallelWellInfo.cpp
```diff
@@ -0,0 +1,325 @@
...
+}
+
+bool operator!=(const ParallelWellInfo& well1, const ParallelWellInfo& well2)
+{
+    return ! (well1 == well2);
+}
+
+bool operator<(const std::pair<std::string, bool>& pair, const ParallelWellInfo& well)
+{
+    return pair.first < well.name() || ( !( well.name() < pair.first ) && pair.second < well.hasLocalCells() );
```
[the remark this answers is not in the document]
blattms: Interesting idea. To make this work, `hasLocalCells()` would need to return `const bool&`, and then the line would read:
```
return static_cast<const std::tuple<std::string, bool>>(pair) < std::tie(well.name(), well.hasLocalCells());
```
I am not sure whether this is easier to read for everybody.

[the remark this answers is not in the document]
blattms: I would only change if somebody really insists on it. To me this change would not improve readability.

opm/simulators/wells/ParallelWellInfo.cpp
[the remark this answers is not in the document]
blattms: Actually this is exactly what was done in BlackoilWellModel_impl.hpp:L625-L633. It just moved here.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -52,17 +52,18 @@ namespace Opm {
         const auto& cartDims = Opm::UgGridHelpers::cartDims(grid);
         setupCartesianToCompressed_(Opm::UgGridHelpers::globalCell(grid),
                                     cartDims[0]*cartDims[1]*cartDims[2]);
-
-        is_shut_or_defunct_ = [&ebosSimulator](const Well& well) {
+        auto& parallel_wells = ebosSimulator.vanguard().parallelWells();
+        parallel_well_info_.assign(parallel_wells.begin(), parallel_wells.end());
+        is_shut_or_defunct_ = [this, &ebosSimulator](const Well& well) {
```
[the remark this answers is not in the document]
blattms: How about this change (to even get rid off capturing `this`):
```diff
@@ -54,16 +54,21 @@ namespace Opm {
                                     cartDims[0]*cartDims[1]*cartDims[2]);
         auto& parallel_wells = ebosSimulator.vanguard().parallelWells();
         parallel_well_info_.assign(parallel_wells.begin(), parallel_wells.end());
-        is_shut_or_defunct_ = [this, &ebosSimulator](const Well& well) {
+        const auto& pwell_info = parallel_well_info_;
+        is_shut_or_defunct_ = [&pwell_info, &ebosSimulator](const Well& well) {
                 if (well.getStatus() == Well::Status::SHUT)
                     return true;
                 if (ebosSimulator.gridView().comm().size() == 1)
                     return false;
                 std::pair<std::string, bool> value{well.name(), true}; // false indicate not active!
-                auto candidate = std::lower_bound(parallel_well_info_.begin(),
-                                                  parallel_well_info_.end(),
+                auto candidate = std::lower_bound(pwell_info.begin(),
+                                                  pwell_info.end(),
                                                   value);
-                return candidate == parallel_well_info_.end() || *candidate != value;
+                return candidate == pwell_info.end() || *candidate != value;
             };
 
         alternative_well_rate_init_ = EWOMS_GET_PARAM(TypeTag, bool, AlternativeWellRateInit);
```

blattms:
jenkins build this please

opm/simulators/wells/ParallelWellInfo.cpp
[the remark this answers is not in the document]
blattms: If we want to change we should do it in another PR (which might mean never of course).

opm/simulators/wells/BlackoilWellModel_impl.hpp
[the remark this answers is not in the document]
blattms: done.

blattms:
there you go, should change code for minor things. One tests fails because of restart:
```
restart sequences 

    0 |     0
....
  156 |   156
  157 |   157
  158 |   158
  159 |      
   16 |    16
  160 |      
  161 |      
  162 |      
  163 |      
....
Program threw an exception: [/var/lib/jenkins/workspace/opm-simulators-PR-builder@2/deps/opm-common/test_util/EclRegressionTest.cpp:781] 
Restart files not having the same report steps: 
````

Not sure how that should be related to the last commit.

**Comment by bska** [comment]
> One tests fails because of restart:

Race condition.  OPM/opm-tests#450, which reduces the number of report steps in this regression test, was merged in the interim without an associate update to the reference solutions.  Master is broken until my request to update the reference solution from #2960 goes through.

**Comment by blattms** [comment]
jenkins build this please

**Comment by blattms** [comment]
Would have been nice to postpone b7fd9e42 a bit. Now I have to resolve merge conflicts again.

**Comment by blattms** [comment]
I have the feeling this will be fun (with my multiple PRs that touch quite some parts of the refactored code). 

It has been 7 days and there seem more to come for this (IMHO rather clean) PR.

**Comment by atgeirr** [comment]
I am reviewing this (started Tuesday, continuing now), I hope to have useful feedback before lunch.

**Comment by blattms** [comment]
rebased.

**Comment by blattms** [comment]
jenkins build this please

File: opm/simulators/wells/ParallelWellInfo.cpp [hunk 4]
```diff
@@ -0,0 +1,325 @@
...
+        remote_indices_.rebuild<true>();
+        using FromSet = Dune::EnumItem<Attribute,owner>;
+        using ToSet = Dune::AllSet<Attribute>;
+        interface_.build(remote_indices_, FromSet(), ToSet());
+        communicator_.build<double*>(interface_);
+    }
+#endif
+}
+
+std::vector<double> CommunicateAbove::communicate(double first_above,
```
atgeirr, line 93: I have some trouble understanding what is supposed to happen here. Could you give a simple example?

opm/simulators/wells/WellInterface.hpp
```diff
@@ -317,6 +319,8 @@ namespace Opm
 
         Well well_ecl_;
 
+        const ParallelWellInfo* parallel_well_info_;
```
atgeirr, line 322: Why a pointer and not a reference?

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -216,11 +217,23 @@ namespace Opm {
     template<typename TypeTag>
     std::vector< Well >
     BlackoilWellModel<TypeTag>::
-    getLocalNonshutWells(const int timeStepIdx, int& globalNumWells) const
+    getLocalNonshutWells(const int timeStepIdx, int& globalNumWells)
```
atgeirr, line 220: Suggest to make this a separate function, and keep the old function unchanged (and still const).

opm/simulators/wells/WellHelpers.hpp
```diff
@@ -33,6 +40,108 @@ namespace Opm {
...
+                if (this->pinfo_->communication().size() == 1)
+                {
+                    // Do the same thing as before. The else branch
+                    // produces different rounding errors and results
+                    // slightly different iteration counts / well curves
+                    B_->mmv(x, y);
+                }
+                else
+                {
+                Y temp(y);
```
atgeirr, line 135: Nitpick: indentation.

opm/simulators/wells/WellHelpers.hpp
```diff
@@ -33,6 +40,108 @@ namespace Opm {
...
+                mv(x, temp); // includes parallel reduction
+                y -= temp;
+                }
+            }
+        private:
+            const Matrix* B_;
+            const ParallelWellInfo* pinfo_;
```
atgeirr, line 142: The `pinfo_` name should be changed to `pwinfo_`, I confused it with the ParallelInfo objects we pass around in other parts of the code that are an entirely different class (not suggesting we go for full Hungarian notation though...)

opm/simulators/wells/WellHelpers.hpp
```diff
@@ -33,6 +40,108 @@ namespace Opm {
...
+                                          + " well should be " + pinfo_->name() + " but is "
+                                + name;
+                            OpmLog::debug(msg);
+                        }
+                        consistentWells = false;
+                        break;
+                    }
+                }
+                pinfo_->communication().barrier();
+                // As not all processes are involved here we need to use MPI_Abort and hope MPI kills them all
```
atgeirr, line 100: This should be dealt with in the BlackoilWellModel context (where all processes participate), using a DeferredLogger for the logging above, but as this is a pathological situation that (hopefully) is only arising during debugging, I think we can leave it as is for now.

ebos/eclbasevanguard.hh
```diff
@@ -416,6 +416,16 @@ public:
...
         if (outputInterval >= 0)
             schedule().restart().overrideRestartWriteInterval(outputInterval);
+
+        // Initialize parallelWells with all local wells
+        const auto& schedule_wells = schedule().getWellsatEnd();
+        parallelWells_.reserve(schedule_wells.size());
+
+        for (const auto& well: schedule_wells)
+        {
+            parallelWells_.emplace_back(well.name(), true);
```
atgeirr, line 426: Not sure why we create this? We will filter later in any case?

atgeirr (commented):
This looks quite good, but I struggle understanding the CommunicateAbove stuff. Also, the comments made in #2840 in October still apply, as far as I can tell.

blattms:
Sigh. Quite frankly: Isn't this review a little bit over the top?

opm/simulators/wells/ParallelWellInfo.cpp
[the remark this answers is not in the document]
blattms: And back to start it is...

We need to find wells by name that perforate local cells (ìt is `hasLocalCells()`). The load balancer returns pairs, and we need to compare.

> I still think it is better to store the info in a map by name instead of a sorted vector, which adds complexity.

I don't. this saves storage and should have better performance. Usage is rather confined, too.

atgeirr:
> Sigh. Quite frankly: Isn't this review a little bit over the top?

What is over the top about it? Please be concrete.

We are talking about lots of new code, and I have spent hours reading it and trying to understand it well enough to give useful feedback. When new classes are obscure and hard to understand, I think it is important to make things clearer, because other people than us may need to understand and change it in the future. I did already make some comments on an earlier version in October, and none of them have been considered as far as I can tell. It should not be surprising that these things then resurface.

opm/simulators/wells/ParallelWellInfo.hpp
```diff
@@ -0,0 +1,260 @@
...
+#include <dune/istl/owneroverlapcopy.hh>
+
+#include <opm/parser/eclipse/EclipseState/Schedule/Well/Well.hpp>
+
+#include <memory>
+
+namespace Opm
+{
+
+/// \brief Class to facilitate getting values associated with the above perforation
```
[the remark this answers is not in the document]
blattms: that is not my word, it is used in current master, see 042fb78

opm/simulators/wells/ParallelWellInfo.hpp
```diff
@@ -0,0 +1,260 @@
...
+    using MPIComm = typename Dune::MPIHelper::MPICommunicator;
+#if DUNE_VERSION_NEWER(DUNE_COMMON, 2, 7)
+    using Communication = Dune::Communication<MPIComm>;
+#else
+    using Communication = Dune::CollectiveCommunication<MPIComm>;
+#endif
+
+
+    /// \brief Constructs object using MPI_COMM_SELF
+    ParallelWellInfo(const std::string& name = {""},
```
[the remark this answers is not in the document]
blattms: It is needed.

opm/simulators/wells/WellHelpers.hpp
blattms: Sorry but that approach does not work anymore with distribributed wells if you want to have any chance of better performance.

Indeed the DeferredLogger will not work anymore as before as we are using different communicators. Maybe there is a fix, but certainly not at this point.

opm/simulators/wells/WellHelpers.hpp
blattms: :angry: this is what happens with constant rebasing.... Will result in another 3 merge conflicts....

opm/simulators/wells/WellHelpers.hpp
blattms: Alas more merge conflicts....

opm/simulators/wells/WellInterface.hpp
blattms: I think we need to be able to copy the wells? Anyhow I will not change this.

blattms (commented):
thanks for finally reviewing this. Given the time I have put in to make this reviewable, the review seems still a bit discerning.

I am considering to file a complaint about animal abuse with PETA. If you make a circus horse jump through higher and higher hoops you should at least not withhold the reward/sugar.

In the future I might try to mimic the approach of self-merging of code not touching the default behaviour just like [others](https://github.com/OPM/opm-simulators/pull/2946#issuecomment-734698791) already often do...

blattms:
jenkins build this please

atgeirr:
> the review seems still a bit discerning.

Not sure what you mean?

adjective: discerning
having or showing good judgement.
"the brasserie attracts discerning customers"

Did you mean the opposite?

In any case, I'll start on your responses.

opm/simulators/wells/ParallelWellInfo.hpp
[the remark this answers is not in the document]
atgeirr: > It is needed.

Why? Certainly you will not construct this in 5 different ways? Why the complexity?

opm/simulators/wells/ParallelWellInfo.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+    if ( del )
+    {
+        // Not 100% nice but safe as comm is deleted anyway
+        // We can only access a copy and no reference.
+        MPI_Comm mpi_comm = *comm;
+        MPI_Comm_free(&mpi_comm);
+    }
+#endif
+    delete comm;
```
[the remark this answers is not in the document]
atgeirr: Sure, but our previous discussion on this was not resolved. Of course I can change the code, but if my ideas on this are wrong that won't help. It's still not clear to me why we need this kludge instead of the simpler approach.

opm/simulators/wells/WellInterface.hpp
atgeirr: > I think we need to be able to copy the wells?

That is no reason to avoid using a reference. If it cannot be null and need not be reset it should be a reference.

> Anyhow I will not change this.

Huh?

opm/simulators/wells/BlackoilWellModel_impl.hpp
atgeirr: You have a conflict with #2936 over this, which will cause more wasted time. Conceptually, a getter function should remain const. The new function does two things, and the new addition has different semantics (mutating state). I think that is very good reason for splitting it.

opm/simulators/wells/WellHelpers.hpp
atgeirr: That is very strange, as this is a brand new code that has not been modified anywhere else.

opm/simulators/wells/WellHelpers.hpp
atgeirr: Again, I do not understand why this would cause merge conflicts.

opm/simulators/wells/WellHelpers.hpp
atgeirr: > Sorry but that approach does not work anymore with distribributed wells if you want to have any chance of better performance.

I completely disagree. Are you saying that we cannot do proper communication of error conditions anymore? As long as we do not communicate such things too often, it should pose no problem. I think that for debugging it is ok, but in general it is not fine to crash and burn like this with MPI_Abort!

> Indeed the DeferredLogger will not work anymore as before as we are using different communicators. Maybe there is a fix, but certainly not at this point.

It will work just fine, it will still do communication with all processes. No change needed. As long as you push a message regarding a well to a DeferredLogger from just a single process (the well-set rank 0 as used above seems fine) all should be good. Then they could be collected at the global scope as before. We may want to avoid it for errors inside the apply() method, but that is an issue of granularity and avoiding communication too often, not of the DeferredLogger being unapplicable.

ebos/eclbasevanguard.hh
atgeirr: Building local_parallel_well_info_ instead should be sufficient. Then you have one less member, and less code.

opm/simulators/wells/ParallelWellInfo.cpp
atgeirr: Thanks, that was clarifying, I now understand the purpose of the class. I think that it is correct for the single-process case. I have not checked the actually communicating case, but will assume that is ok.

blattms (commented):
Most comments should be adressed. I really hope this can be merged now.

blattms:
jenkins build this please

blattms:
Last commit was missing in action. I am loosing my sanity with all these branches in flight.

blattms:
jenkins build this please

opm/simulators/wells/ParallelWellInfo.hpp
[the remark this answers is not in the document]
atgeirr: Thanks for the usage notes. I do not see the need for the default constructor still, but I assume it will be needed.

> Seriously, how is using default arguments in two constructors increasing complexity. In my opinion it is the other way around.

Because it adds more possible signatures. More signatures means more complexity reading and understanding the code. I may not have complained about too many constructors much, but then it is not a very common problem!

I think default arguments should preferably only be used when adding a new argument and it is important to not have to rewrite existing calls.

opm/simulators/wells/ParallelWellInfo.cpp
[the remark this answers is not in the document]
atgeirr: I'll accept that for now, and revise it later instead, as suggested.

opm/simulators/wells/WellHelpers.hpp
atgeirr: The initializeWellPerfData() is also actually buggy, good example. I did not intend that you should remove this code, just tried to make sure we agree on the general approach to errors here.

ebos/eclbasevanguard.hh
atgeirr: Ok, keep as is.

atgeirr:
I get a compile error after merging this, which is strange since it passed on Jenkins:
```
/Users/atgeirr/opm/src/opm-simulators/opm/simulators/wells/ParallelWellInfo.cpp:102:70: error: 'CopyGatherScatter' is a protected member of
      'Dune::OwnerOverlapCopyCommunication<int, int>'
        using Handle = Dune::OwnerOverlapCopyCommunication<int,int>::CopyGatherScatter<double*>;
                                                                     ^
/Users/atgeirr/duneinstallhack/include/dune/istl/owneroverlapcopy.hh:201:12: note: declared protected here
    struct CopyGatherScatter
```
Looking at the code the error looks real, with no change to the protectedness in Dune 2.7 vs 2.6 either. Anyone can provide some insight?

I also get a second compile error, but that one is clearly a gcc vs clang issue, with clang being a little stricter on when to require `template` for calling templated member functions. (Will make a PR for that one.)

akva2:
i see the same when building with gcc 9. gcc 8 accepts the code. so this basically builds by accident/ due to compiler bugs.

atgeirr:
> i see the same when building with gcc 9. gcc 8 accepts the code. so this basically builds by accident/ due to compiler bugs.

Then an alternative must be found. I have temporarily changed protected to private in the particular header file to get around this, but a proper solution is needed. @blattms what do you propose?

bska:
> i see the same when building with gcc 9.

Me too.

opm/simulators/wells/BlackoilWellModel_impl.hpp
blattms: ... and this somehow broke #2954. Seems like calling two functions (`getLocalNonshutWells` followed by `createLocalParallelWellInfo`) in `initFromRestartFile` is to complex for me. 

 Quite surprising that jenkins on this PR did not segfault as in the other PR.

blattms: turns out that is no surprise as changes trickering it are still missing. puh, that was close.

PR #2949, opened 2020-11-27
Check operability for MSW

totto82:
This PR moves the operability checking to the wellInterface class to make it usable for both well models.

only
checkOperabilityUnderBHPLimitProducer 
checkOperabilityUnderTHPLimitProducer
updateIPR
are specialized. 

The two first are almost identical and could be moved to the wellInterface class with changes to some of the functions. But not in this PR since it will change unrelated code. 

The updateIPR needs to be specialized since the pressure gradient is calculated differently in the Standard well and the MSW well. 

The PR has been tested on some toy-cases and on a big model where this PR is sufficient and necessary for Flow to finish the simulation.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -2678,9 +2678,11 @@ namespace Opm {
             this->previous_well_state_ = this->well_state_;
 
             well_container_ = createWellContainer(timeStepIdx);
-
+            std::vector< Scalar > B_avg(numComponents(), Scalar() );
+            // we don't plan to iterate so just passing trivial B_avg
+            // for now
             for (auto& well : well_container_) {
-                well->init(&phase_usage_, depth_, gravity_, local_num_cells_);
+                well->init(&phase_usage_, depth_, gravity_, local_num_cells_, B_avg);
             }
 
             std::fill(is_cell_perforated_.begin(), is_cell_perforated_.end(), false);
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -343,8 +343,11 @@ namespace Opm {
             // do the initialization for all the wells
             // TODO: to see whether we can postpone of the intialization of the well containers to
             // optimize the usage of the following several member variables
+            std::vector< Scalar > B_avg(numComponents(), Scalar() );
+            computeAverageFormationFactor(B_avg);
+
             for (auto& well : well_container_) {
-                well->init(&phase_usage_, depth_, gravity_, local_num_cells_);
+                well->init(&phase_usage_, depth_, gravity_, local_num_cells_, B_avg);
             }
 
             // update the updated cell flag
@@ -435,7 +438,7 @@ namespace Opm {
                 WellInterfacePtr well = createWellForWellTest(well_name, timeStepIdx, deferred_logger);
 
                 // some preparation before the well can be used
-                well->init(&phase_usage_, depth_, gravity_, local_num_cells_);
+                well->init(&phase_usage_, depth_, gravity_, local_num_cells_, B_avg);
                 const Well& wellEcl = schedule().getWell(well_name, timeStepIdx);
                 double well_efficiency_factor = wellEcl.getEfficiencyFactor();
                 WellGroupHelpers::accumulateGroupEfficiencyFactor(schedule().getGroup(wellEcl.groupName(), timeStepIdx), schedule(), timeStepIdx, well_efficiency_factor);
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -2393,7 +2393,7 @@ namespace Opm
     template<typename TypeTag>
     void
     MultisegmentWell<TypeTag>::
-    checkOperabilityUnderBHPLimitProducer(const WellState& well_state, const Simulator& ebos_simulator, Opm::DeferredLogger& deferred_logger)
+    checkOperabilityUnderBHPLimitProducer(const WellState& /*well_state*/, const Simulator& ebos_simulator, Opm::DeferredLogger& deferred_logger)
     {
         const auto& summaryState = ebos_simulator.vanguard().summaryState();
         const double bhp_limit = Base::mostStrictBhpFromBhpLimits(summaryState);
@@ -2554,7 +2554,7 @@ namespace Opm
     template<typename TypeTag>
     void
     MultisegmentWell<TypeTag>::
-    checkOperabilityUnderTHPLimitProducer(const Simulator& ebos_simulator, const WellState& well_state, Opm::DeferredLogger& deferred_logger)
+    checkOperabilityUnderTHPLimitProducer(const Simulator& ebos_simulator, const WellState& /*well_state*/, Opm::DeferredLogger& deferred_logger)
     {
         const auto& summaryState = ebos_simulator.vanguard().summaryState();
         const auto obtain_bhp = computeBhpAtThpLimitProd(ebos_simulator, Base::B_avg_, summaryState, deferred_logger);
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -2574,11 +2574,10 @@ namespace Opm
                 deferred_logger.debug(msg);
             }
         } else {
-            // Shutting wells that can not find bhp value from thp may results in
-            // too early shutting of the well during startup of new wells.
-            // Lets try to keep them open.
-            //this->operability_status_.can_obtain_bhp_with_thp_limit = false;
-            //this->operability_status_.obey_bhp_limit_with_thp_limit = false;
+            // Shutting wells that can not find bhp value from thp
+            // when under THP control
+            this->operability_status_.can_obtain_bhp_with_thp_limit = false;
+            this->operability_status_.obey_bhp_limit_with_thp_limit = false;
             if (!this->wellIsStopped()) {
                 const double thp_limit = this->getTHPConstraint(summaryState);
                 deferred_logger.debug(" could not find bhp value at thp limit "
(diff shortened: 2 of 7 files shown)


totto82:
jenkins build this please

alfbr:
Thank you! I will test this.

GitPaean:
The following failures are related to shutting multisegment wells. 
```
mpi.compareECLFiles_flow+3D_MSW
mpi.compareECLFiles_flow+2D_H__
mpi.compareECLFiles_flow+1_MSW_MODEL6
mpi.compareRestartedSim_flow+3D_MSW
```

totto82:
> The following failures are related to shutting multisegment wells.

I will test them carefully. The code may need some adjustments.

alfbr:
I am about to test them now :-)

totto82:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -353,8 +353,11 @@ namespace Opm {
             // do the initialization for all the wells
             // TODO: to see whether we can postpone of the intialization of the well containers to
             // optimize the usage of the following several member variables
+            std::vector< Scalar > B_avg(numComponents(), Scalar() );
+            computeAverageFormationFactor(B_avg);
+
             for (auto& well : well_container_) {
-                well->init(&phase_usage_, depth_, gravity_, local_num_cells_);
+                well->init(&phase_usage_, depth_, gravity_, local_num_cells_, B_avg);
             }
```
totto82, line 361: @GitPaean @atgeirr What do you think about this? Computing the B_avg at every well iteration and everytime we compute the well potential is really not necessary IMO.  I would happily remove the passing of the B_avg around in the rest of the well code if this gets merged. We shouldn't compute global averages like this more often than we need.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -2365,33 +2381,202 @@ namespace Opm
...
+            // if the BHP limit is not defaulted or the well does not have a THP limit
+            // we need to check the BHP limit
+
+            double temp = 0;
+            for (int p = 0; p < number_of_phases_; ++p) {
+                temp += ipr_a_[p] - ipr_b_[p] * bhp_limit;
+            }
+            if (temp < 0.) {
+                this->operability_status_.operable_under_only_bhp_limit = false;
+            }
```
totto82, line 2405: @GitPaean For the standard well every phase is checked. For phases with very small rates this could lead to immature shutting of wells. What do you think? Should I change this for the standard wells as well?

akva2:
please mind new warnings.

totto82:
> please mind new warnings.

I know. Still trying to fix the last failure. But thanks for noticing. I often forget.

totto82:
The last failure is due to small changes in convergence and time step behavior. The results are the same in the eyeball norm.

akva2:
115/126 Test #100: compareECLFiles_flow+MOD4_UDQ_ACTIONX .............................   Passed  190.30 sec
116/126 Test  #99: compareECLFiles_flow+MOD4_GRP .....................................   Passed  314.72 sec

these use to pass in around 60-80 secs. i don't think it can be attributed to machine load only .. ?

totto82:
> these use to pass in around 60-80 secs. i don't think it can be attributed to machine load only .. ?

It runs fine locally. Lets try again.

totto82:
jenkins build this please

akva2:
same thing. every test with MSW is severely slowed down. i can reproduce locally as well, went from 53 secs to 186 secs. in particular
```
   Well assembly (seconds):      5.07 (Failed: 0.9; 17.4%)
```
is now
```
   Well assembly (seconds):    144.03 (Failed: 25.3; 17.6%)
```

opm/simulators/wells/MultisegmentWell_impl.hpp
GitPaean: The well operability formulation is really far from ideal. Both Flow and commercial simulator can shut well prematurely because of the simplified model.  For this sense, if algorithm can reduce the shutting of the wells in general, it should be okay. 

> For phases with very small rates this could lead to immature shutting of wells. 

I think here you mean it shuts well less. Ideally, check three phases individually and check the sum should be the same. I am okay if you want to check the sum to reduce the shutting of the wells because of the reason stated above. Maybe can be slightly better, if you check the weighted sum with the scaling factors (similar to the weighted total rate), since the purpose is to avoid small rate values causing difficulty in determining the sign. Weighting might help to avoid some other problems when gas rate is dominating. 

Maybe the change related to StandardWell can be done in a separate PR for better testing.

totto82:
The main reason for the slow down is that checking of the well operability for MSW is not trivial. The operability check is done every newton iteration. Most of the time is spent in computeWellRatesWithBhp(...) that again calls the iterateWellEqWithControl(...) that includes solving of the well equation using umfpack at every well iteration. If the well model does not convergence easily this leads to very many solves of the well equation. I think the best solution for speed up is to improve the initialization of the well solution when calling computeWellRatesWithBhp(...) This will need some thinking and restructuring of the code. i.e. should not be part of this pr. For larger models I have tested the extra simulation time from the operability check is not dominating the cost. Indeed it significantly reduce the time, since the operability check shuts problematic wells that cause severe convergence issues.

alfbr:
I have tested this on one full field case where it works perfectly. I am also interested in doing full ensemble testing to check robustness, but have we sorted out the test-case failures yet?

totto82:
> but have we sorted out the test-case failures yet?

I have manually checked it and it is the same in the eyeball norm. we should just do a update_data and merge it IMO.

atgeirr:
> we should just do a update_data and merge it IMO.

Will not that give a significant performance regression for all MSW cases? Or was there some change to the calculations that made them simpler and faster?

totto82:
> Will not that give a significant performance regression for all MSW cases? Or was there some change to the calculations that made them simpler and faster?

The problem is  the cost of computeWellRatesWithBhp(...) not this PR in itself. Too speed up the code we will need to improve the convergence of the well model or reducing the cost of an iteration. I think we need to do both, but not in this PR. For the larger models I have tested this PR significantly reduce the simulation time.

alfbr:
> Will not that give a significant performance regression for all MSW cases? Or was there some change to the calculations that made them simpler and faster?

I did not see the performance degradation, so I believe this is more of an issue on the small test cases.

totto82:

totto82:
jenkins build this opm-tests=455 please

alfbr:
Can this be merged?

bska:
> Can this be merged?

In principle, yes.  In practice we still need to decide what to do about the considerable extra cost of these checks.  The time to run the regression/unit tests goes up from about 17 minutes to 54 minutes on the CI system when we include this patch and that's not nothing.

alfbr:
> In principle, yes. In practice we still need to decide what to do about the considerable extra cost of these checks. The time to run the regression/unit tests goes up from about 17 minutes to 54 minutes on the CI system when we include this patch and that's not nothing.

That is considerable. Still, it seems this is a performance problem that got exposed here, but which has been present for a long time for multisegment wells. I cannot see that the burden of fixing the performance can be put on this PR, but rather needs to be addressed separately.

totto82:
> 17 minutes to 54 minutes

That's too much. I will try out some ideas to speed it up.

alfbr:
benchmark please

ytelses:
Benchmark result overview: 




View result details @ https://www.ytelses.com/opm/?page=result&id=988

alfbr:
Seems we need a rebase to get benchmark results. We do not expect any difference for the benchmark cases though, so no hurry.

totto82:
jenkins build this please

GitPaean:
Instead of changing the logic/code way above the purpose of this PR too much (then no one knows what is the effect of this PR anymore). If there is some idea which test case is having problem with this PR, we can adjust the test case. For example, making some unnecessary prediction case to be history matching case.  Operability checking is only for prediction case and to some extent, the THP checking will be expensive, even for the commercial simulator.

GitPaean:
I believe well operability checking for MSW is expensive is because we need to solve the well equations when `computeWellRatesWithBhp`. Not sure whether profiling supports this and whether there is something we can do about it.

totto82:
jenkins build this please

totto82:
This PR with latest modfication
```
118/129 Test #102: compareECLFiles_flow+MOD4_UDQ_ACTIONX .............................   Passed   91.56 sec
119/129 Test #101: compareECLFiles_flow+MOD4_GRP .....................................   Passed  122.51 sec

```
PR https://github.com/OPM/opm-simulators/pull/2968 (should be the same as master)
```
118/129 Test #102: compareECLFiles_flow+MOD4_UDQ_ACTIONX .............................   Passed  100.69 sec
119/129 Test #101: compareECLFiles_flow+MOD4_GRP .....................................   Passed  133.32 sec
```

totto82:
jenkins build this opm-tests=455 please

atgeirr:
@totto82 Is it correct to interpret your timing results that the tests are now faster with this PR instead of being a lot slower?

totto82:
> @totto82 Is it correct to interpret your timing results that the tests are now faster with this PR instead of being a lot slower?
Locally the speed is the same. The difference in timing is probably due to jenkins load. The message is that it is not significant slower anymore, but it still fixes the issue on the field model. 
@atgeirr The alternative fix of only checking failed wells during iterations did not work on the big model.

atgeirr:
> The alternative fix of only checking failed wells during iterations did not work on the big model.

Can you summarize the changes you made instead, to get acceptable speed?

alfbr:
> Can you summarize the changes you made instead, to get acceptable speed?

It is in the commit message I believe: 
Only check well operability for wells under THP and BHP limits

totto82:
> Can you summarize the changes you made instead, to get acceptable speed?

With this we only update the operability for thp and bhp controlled wells. I think that is sufficient since a well should first hit the bhp or thp limit before it stops being operable. 

See commit: https://github.com/OPM/opm-simulators/pull/2949/commits/5fb73c971440ac344983a1bbc5f413eaf8c5370

totto82:
@bska or @akva2 Do I need to run update data again since https://github.com/OPM/opm-tests/pull/457 is merged? I must admit I have never understood when I have to rerun the update_data.  https://github.com/OPM/opm-tests/pull/455 is mergable and does not change files that is affected by https://github.com/OPM/opm-tests/pull/457

akva2:
if there are no conflicts, it's all good.

akva2:
jenkins build this opm-tests=455 please

totto82:
Thanks @akva2 for the clarification.

bska:
New reference data has been installed on the CI system.  Merging this now to bring the system back into a consistent state.

PR #2950, opened 2020-11-27
Well PI: Use Appropriate Calculator for Each Well

bska:
The original code assumed that
```C++
well_container_.size() == numLocalWells()
```
This assumption does not hold when wells open/shut dynamically in the context of `WECON` and/or `WTEST`.

Switch to indexing into the `prod_index_calc_` vector using the well's own linear index instead of manually advancing iterators.

Pointy Hat: [at]bska

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1471,13 +1471,11 @@ namespace Opm {
             return;
         }
 
-        auto piCalc = this->prod_index_calc_.begin();
         for (const auto& wellPtr : this->well_container_) {
             wellPtr->updateProductivityIndex(this->ebosSimulator_,
-                                             *piCalc,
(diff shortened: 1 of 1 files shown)


bska:

bska:
This fixes a real problem.  I'll update the 3D_WECON test data and self-merge to bring the reference solutions back to a consistent state.

bska:

atgeirr:
This is a good change, and should be merged together with the data update when green. (Assuming no surprises, and that the data update only changes the WECON test case).

@joakim-hove I think that the change here is good also for the work you are doing on ACTIONX dynamic opening of wells.

bska:
jenkins build this serial opm-tests=448 please

bska:
New reference data is installed on the CI system.  Merging this now to bring the system back to a consistent state.

joakim-hove:
> I think that the change here is good also for the work you are doing on ACTIONX dynamic opening of wells.

Hehe - yes I barked up that tree for a while ;-)

PR #2953, opened 2020-11-30
Shift the report step argument used when calling UDQ eval

joakim-hove:


joakim-hove:
jenkins build this please

ebos/eclwriter.hh
```diff
@@ -328,8 +328,15 @@ public:
...
 
-            const auto& udq_config = schedule().getUDQConfig(reportStepNum);
-            udq_config.eval( reportStepNum, schedule().wellMatcher(reportStepNum), summaryState(), udqState() );
+            /*
+              Off-by-one-fun: The reportStepNum argument corresponds to the
+              report step these results will be written to, whereas the argument
+              to UDQ function evaluation corresponds to the report step we are
+              currently on.
+            */
+            auto udq_step = reportStepNum - 1;
```
bska, line 337: I guess there's a guarantee from the broader context that `reportStepNum > 0` here, but I think we should add a comment to that effect to assist future readers.

PR #2957, opened 2020-12-01
Wbp collect

joakim-hove:
I have written some MPI stuff:

image

Thing goes as follows:

1. There is a family of summary keywords `WBP, WBP4, WBP5` and `WBP9`. These are different ways to calculate an average well pressure based on the completed cells and their neighbours. 
2. In opm-common there is a class [`PAvgCalculatorCollection`](https://github.com/OPM/opm-common/blob/master/opm/parser/eclipse/EclipseState/Schedule/Well/PAvgCalculatorCollection.hpp) which holds on to the wells which want to calculate the `WBPx` keywords. The most important method for the current PR is the `::index_list()` which returns all the global indices where we need the pressure.
3. This PR will distribute the `index_list` to all PE's and then assemble the requested pressures and update `PAvgCalculatorCollection` instance and pass it to the summary writer.

The actual calculation of WBP in opm-common is not yet complete; but that should be possible to complete only in opm-common with no coupling to opm-simulators.

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -225,6 +225,11 @@ public:
             }
         }
 
+        for (const auto& global_index : wbp_index_list) {
+            if (collectToIORank.isCartIdxOnThisRank(global_index - 1))
+                this->wbpData_[global_index] = 0.0;
+        }
+
         forceDisableFipOutput_ = EWOMS_GET_PARAM(TypeTag, bool, ForceDisableFluidInPlaceOutput);
     }
 
@@ -837,6 +842,8 @@ public:
             if (gasConnectionSaturations_.count(cartesianIdx) > 0) {
                 gasConnectionSaturations_[cartesianIdx] = Opm::getValue(fs.saturation(gasPhaseIdx));
             }
+            if (this->wbpData_.count(cartesianIdx) > 0)
+                this->wbpData_[cartesianIdx] = Opm::getValue(fs.pressure(oilPhaseIdx));
 
             // tracers
             const auto& tracerModel = simulator_.problem().tracerModel();
@@ -1852,6 +1859,10 @@ public:
         return 0;
     }
 
+    const std::map<std::size_t, double>& getWBPData() const {
+        return this->wbpData_;
+    }
+
     const std::map<std::pair<std::string, int>, double>& getBlockData()
     { return blockData_; }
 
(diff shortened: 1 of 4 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this opm-common=2146 please

joakim-hove:
jenkins build this opm-common=2146 please

joakim-hove:
Cool if someone in the know when IT comes to mpi Will tale look at this: @blattms, @akva2, @atgeirr

ebos/ecloutputblackoilmodule.hh
```diff
@@ -837,6 +842,8 @@ public:
             if (gasConnectionSaturations_.count(cartesianIdx) > 0) {
                 gasConnectionSaturations_[cartesianIdx] = Opm::getValue(fs.saturation(gasPhaseIdx));
             }
+            if (this->wbpData_.count(cartesianIdx) > 0)
+                this->wbpData_[cartesianIdx] = Opm::getValue(fs.pressure(oilPhaseIdx));
```
blattms, line 846: using map.find would save one lookup.

blattms (commented):
As far as  I can jusge this seems sound. Not sure how much the calculator has to calculate. If it is not negligible, we might not want to do all of it on one processes.

joakim-hove:
Thanks for looking into this.

> Not sure how much the calculator has to calculate. If it is not negligible, we might not want to do all of it on one processes.

That is really neglible; essentially a sum over the total number of connected cells.

ebos/ecloutputblackoilmodule.hh
joakim-hove: I agree - however here I followed the pattern in the surrounding code.

joakim-hove:
jenkins build this please

ebos/eclwriter.hh
```diff
@@ -191,22 +191,37 @@ public:
...
+            wbp_index_list = wbp_calculators.index_list();
         }
+        if (collectToIORank_.isParallel()) {
+#ifdef HAVE_MPI
+            unsigned long size = wbp_index_list.size();
+            MPI_Bcast(&size, 1, MPI_UNSIGNED_LONG, collectToIORank_.ioRank, MPI_COMM_WORLD);
+            if (!collectToIORank_.isIORank())
+                wbp_index_list.resize( size );
 
+            MPI_Bcast(wbp_index_list.data(), size * sizeof(std::size_t), MPI_CHAR, collectToIORank_.ioRank, MPI_COMM_WORLD);
```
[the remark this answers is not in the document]
joakim-hove: Thank I have updated.

akva2:
no update pushed.

joakim-hove:
> no update pushed.

Thanks - now it is.

joakim-hove:
jenkins build this please

PR #2958, opened 2020-12-01
Use WellInterface::perf_data_ to calculate PI.

blattms:
This simplifies the code a bit and will work with distributed wells. Previously, we assumed that all non-shut perforations are stored locally. That does not hold any more.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1210,14 +1210,10 @@ namespace Opm
         setToZero(wellPI);
 
         const auto preferred_phase = this->well_ecl_.getPreferredPhase();
+        auto subsetPerfID   = 0;
 
-        const auto& allConn = this->well_ecl_.getConnections();
-        const auto  nPerf   = allConn.size();
-        auto subsetPerfID   = 0*nPerf;
-        for (auto allPerfID = 0*nPerf; allPerfID < nPerf; ++allPerfID) {
-            if (allConn[allPerfID].state() == Connection::State::SHUT) {
-                continue;
-            }
+        for ( const auto& perf : *this->perf_data_){
+            auto allPerfID = perf.ecl_index;
 
             auto connPICalc = [&wellPICalc, allPerfID](const double mobility) -> double
             {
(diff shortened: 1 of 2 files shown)


blattms:
BTW, the functions for the two wells seem identical to me.

blattms:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1210,14 +1210,10 @@ namespace Opm
         setToZero(wellPI);
 
         const auto preferred_phase = this->well_ecl_.getPreferredPhase();
+        auto subsetPerfID   = 0;
```
bska, line 1213: I'd prefer if you didn't introduce a type-conversion here.  The original expression was crafted to generate a zero of the appropriate type.

bska (commented):
Thanks.  This does indeed simplify the logic.

> the functions for the two wells seem identical to me.

They are indeed very close, but the syntactic similarity belies a semantic difference.  The `EvalWell` types are different in the two cases.  The `StandardWell` uses a [`DenseAd::DynamicEvaluation`](https://github.com/OPM/opm-simulators/blob/f34b28663608be72c2fffb1ccb4d1f44f8b5edb7/opm/simulators/wells/StandardWell.hpp#L152) while the `MultisegmentWell` uses a [`DenseAd::Evaluation`](https://github.com/OPM/opm-simulators/blob/f34b28663608be72c2fffb1ccb4d1f44f8b5edb7/opm/simulators/wells/MultisegmentWell.hpp#L99) and the two `getMobility()` functions know about this difference.

That said, I'm planning a refactoring here to pull much of the implementation up into the `WellInterface`.

opm/simulators/wells/MultisegmentWell_impl.hpp
blattms: What is the correct type? It seems like there are two usages below: 
`getMobility(ebosSimulator, static_cast<int>(subsetPerfID), mob);` and fluidState(subsetPerfID)

blattms: ... and both seem to take an `int` and not a `std::size_t`. Shouldn't my changes be at least as correct as before or does auto not result in an `int` here?

bska: > both seem to take an `int` and not a `std::size_t`. Shouldn't my changes be at least as correct as before or does auto not result in an `int` here?

Ehm, yes&mdash;you're right.  I was thinking of an earlier version of this code that never made it to GitHub and which used `subsetPerfID` differently.  I'll merge this now.

PR #2959, opened 2020-12-01
[bugfix] Do not iterate pass end of perf_data_ in initCompletions

blattms:
IMHO this might have happened if perf_data_ is empty or if the last connection is closed. (Discovered while working on distributed wells but might happen already before!)


blattms:
jenkins build this please

PR #2960, opened 2020-12-01
Well Model: Add Helper for Well Pointer Creation

bska:
This PR adds a new helper function,
```
WellInterfacePtr createWellPointer(wellID, reportStep) const
```
which is responsible for creating appropriately typed derived well pointers depending on well types (multi-segment vs. standard). This, in turn, allows us to centralise this logic and use the same factory function both when creating the `well_container_` and when forming the well-test objects.

Finally, this helper will become useful for calculating PI/II values of shut/stopped wells in the context of WELPI.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -31,8 +31,15 @@
 #include <opm/common/utility/platform_dependent/reenable_warnings.h>
 
 #include <cassert>
-#include <unordered_map>
+#include <functional>
+#include <map>
+#include <memory>
+#include <string>
 #include <tuple>
+#include <unordered_map>
+#include <vector>
+
+#include <stddef.h>
 
 #include <opm/parser/eclipse/EclipseState/Runspec.hpp>
 
@@ -288,6 +295,15 @@ namespace Opm {
 
             // create the well container
             std::vector<WellInterfacePtr > createWellContainer(const int time_step);
+
+            WellInterfacePtr
+            createWellPointer(const int wellID,
+                              const int time_step) const;
+
+            template <typename WellType>
+            std::unique_ptr<WellType>
+            createTypedWellPointer(const int wellID,
+                                   const int time_step) const;
 
             WellInterfacePtr createWellForWellTest(const std::string& well_name, const int report_step, Opm::DeferredLogger& deferred_logger) const;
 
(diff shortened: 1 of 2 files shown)


bska:
jenkins build this please

bska:
jenkins build this please

GitPaean:
This PR passes the Jenkins test.  Let me try whether we can use this one to help generate update for https://github.com/OPM/opm-tests/pull/450 .

GitPaean:
jenkins build this please opm-tests=450 update_data please

GitPaean:
jenkins build this opm-tests=450 update_data please

GitPaean (approved):
This PR looks good to me, although not totally following the intentional usage yet, but the PR itself makes this part more tidy and I consider it an improvement. Will merge after the `update_data` attempt.

bska:
> Let me try whether we can use this one to help generate update for OPM/opm-tests#450 .

I'm not completely sure that we can update the reference data through a PR in opm-tests (@akva2 would know).  If that's not possible, I suggest just merging OPM/opm-tests#450 and then updating the reference data from an existing (or new) PR in a different repository.

GitPaean:
> I suggest just merging OPM/opm-tests#450 and then updating the reference data from an existing (or new) PR in a different repository.

Yes. I guess that is the solution maybe it can  be done during some time when less people using jenkins since the master branch will be broken for a short period.

bska:
> not totally following the intentional usage yet

It turns out that we need to be able to calculate productivity/injectivity index values (PI/II) of shut/stopped wells.  We currently compute those values using [`WellInterface::updateProductivityIndex()`](https://github.com/OPM/opm-simulators/blob/ced3c99e09e6fd4fcfe0d3baabd4dc9644aa99f5/opm/simulators/wells/WellInterface.hpp#L225-L228), but "only" for the wells that exist in the `well_container_`.  With the introduction of `createWellPointer()` I can leverage the existing PI calculation routine by forming a `WellInterface` pointer on the fly for shut/stopped wells when needed.

GitPaean:
> With the introduction of createWellPointer() I can leverage the existing PI calculation routine by forming a WellInterface pointer on the fly for shut/stopped wells when needed.

Thanks for the explanation.

akva2:
generation works as such as long as there are no overlapping changes in the two PR's.  however, you have no way to test the changes, cause we cannot run with two opm-tests PRs.

GitPaean:
jenkins build this opm-tests=453 please

GitPaean:
jenkins build this opm-tests=453 please

bska:

GitPaean:
jenkins build this opm-tests=454 please

GitPaean:
merging as promised.

PR #2965, opened 2020-12-07
changed: use BOOST_CHECK instead of BOOST_TEST

akva2:
BOOST_TEST is not available in older versions of boost.
in particular this broke building on rh6/7.


akva2:
jenkins build this please

PR #2966, opened 2020-12-07
Use a custom handle for copying doubles for above.

blattms:
The one previously used is unfortunately protected upstream (not noticed with gcc <=8).

Fixes fallout from #2947


blattms:
jenkins build this please

PR #2967, opened 2020-12-07
Use constexpr members in BlackoilModelEbos to eliminate unnecessary code

akva2:
Small drive-by change. May as well use constexpr here to get some code elimination.


akva2:
jenkins build this please

atgeirr:
I think this is fine, merge when green. BTW, is there any difference between static constexpr members and non-static constexpr members?

akva2:
a non static member implies that *this* needs to be constexpr as well.

PR #2971, opened 2020-12-08
Simplify computation of connection pressure.

blattms:
Previously we did compute the delta between two perforations and then used a partial sum. That seemed to be a rather complex
algorithm to calculate the delta between a connection and the reference depth.
This commit now simply computes the latter without an intermediary step (difference between two perforations) and hopefully is more clear.

It is also more suitable for distributed wells.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -2201,28 +2201,14 @@ namespace Opm
         // the 'top' perforation is nearest to the surface topologically.
         // Our goal is to compute a pressure delta for each perforation.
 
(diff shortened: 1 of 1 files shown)


blattms:
Hopefully this can be merged fast (if it does not make a fool out of me, that is).

blattms:
jenkins build this please

blattms:
jenkins build this please

bska:
@GitPaean : Would you mind taking a look at this?  Unless I'm missing something, this PR appears to assume that the fluid/density distribution in the wellbore is constant from the reference depth down to all connections.  Is that a valid assumption?

blattms:
Indeed it does.

PR #2973, opened 2020-12-08
[bugfix] Include config.h in WellGroupHelpers.cpp

blattms:
as should be done in all cpp files.

Probably did not surface in master but it does in my branches.


blattms:
jenkins build this please

PR #2975, opened 2020-12-09
Dunify aquifers

totto82:
This PR modifies the aquifer model to use the dune interface. Similar to what is done for the initialization code in https://github.com/OPM/opm-simulators/pull/2941 

I have also removed the usage of the inputGrid since this will not be available in a partitioned grid.

### DIFF
--- a/opm/simulators/aquifers/AquiferCarterTracy.hpp
+++ b/opm/simulators/aquifers/AquiferCarterTracy.hpp
@@ -46,6 +46,7 @@ public:
     using typename Base::RateVector;
     using typename Base::Scalar;
     using typename Base::Simulator;
+    using typename Base::ElementMapper;
 
     using Base::waterCompIdx;
     using Base::waterPhaseIdx;
@@ -92,16 +93,12 @@ protected:
     inline void initializeConnections() override
     {
         const auto& eclState = this->ebos_simulator_.vanguard().eclState();
-        const auto& ugrid = this->ebos_simulator_.vanguard().grid();
         const auto& grid = eclState.getInputGrid();
 
         // We hack the cell depth values for now. We can actually get it from elementcontext pos
         this->cell_depth_.resize(this->size(), this->aquiferDepth());
         this->alphai_.resize(this->size(), 1.0);
         this->faceArea_connected_.resize(this->size(), 0.0);
-
-        auto cell2Faces = Opm::UgGridHelpers::cell2Faces(ugrid);
-        auto faceCells = Opm::UgGridHelpers::faceCells(ugrid);
 
         // Translate the C face tag into the enum used by opm-parser's TransMult class
         Opm::FaceDir::DirEnum faceDirection;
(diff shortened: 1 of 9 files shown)


totto82:
jenkins build this please

opm/simulators/aquifers/AquiferInterface.hpp
```diff
@@ -201,37 +200,20 @@ class AquiferInterface
...
-        const auto calculatedFaceArea
-            = (!this->connections_[idx].influx_coeff.first) ? defaultFaceArea : this->connections_[idx].influx_coeff.second;
-        faceArea = (cellNeighbour0 * cellNeighbour1 > 0) ? 0. : calculatedFaceArea;
-        if (cellNeighbour1 == 0) {
-            faceArea = (cellNeighbour0 < 0) ? faceArea : 0.;
-        } else if (cellNeighbour0 == 0) {
-            faceArea = (cellNeighbour1 < 0) ? faceArea : 0.;
-        }
-        return faceArea;
+        const auto& geometry = intersection.geometry();
```
blattms, line 207: I don't really understand the original comment, but it seems like we are not doing this checks anymore. Did they move elsewhere.

blattms (commented):
Just want to point out that the code might compile with any Dune grid. But depending on which on is used for the aquifer, it might not do the right thing. For the grids used previously it seems to do the same as before.

I think if the face tag info is needed then it needs to stored somewhere and in general cannot be deduced from the grid.

This change will increase the memory consumption a bit as e.g. the depth was previously computed on the fly and now seems to be computed and stored (until the end of the simulation).

totto82:
> I think if the face tag info is needed then it needs to stored somewhere and in general cannot be deduced from the grid.

I know. The face tags are used several places in the code. For instance in the ecltransmissibility.hh, so if we want to support all dune grids we still have a long way to go. My initial aim is to support cpgrid, polyhedralgrid and alugrid created from the inputGrid in eclState. For more general grids we will also need a better discretization. 

>This change will increase the memory consumption a bit as e.g. the depth was previously computed on the fly and now seems to be computed and stored (until the end of the simulation).

I just moved the storage from the problem to the vanguard. That makes it possible to implement the depth method differently. 
For now it is the same.

opm/simulators/aquifers/AquiferInterface.hpp
totto82: I changed it to this. I think it should do the same              
```
  // only deal with grid boundaries
                if (!intersection.boundary())
                    continue;
```

opm/simulators/aquifers/AquiferCarterTracy.hpp
```diff
@@ -91,37 +91,52 @@ class AquiferCarterTracy : public AquiferInterface<TypeTag>
...
+            for (; isIt != isEndIt; ++ isIt) {
+                // store intersection, this might be costly
+                const auto& intersection = *isIt;
+
+                // only deal with grid boundaries
+                if (!intersection.boundary())
+                    continue;
+
+                int insideFaceIdx  = intersection.indexInInside();
+                switch (insideFaceIdx) {
```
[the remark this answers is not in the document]
totto82: I don't think we support NNC + aquifers at the moment.

joakim-hove:
>I just moved the storage from the problem to the vanguard

Ehhh - I thought we wanted to reduce the Vanguard usage - not increase ut?

totto82:
> Ehhh - I thought we wanted to reduce the Vanguard usage - not increase ut?

As long as we insist on using a different cell depth than what the grid can give we need to compute and store the depth somewhere. I think it better belongs to the vanguard than the problem class. 

What about adding the cell depth to the well connection class and to the aquifer data on creation in opm-common? That will reduce the usage of the cell depths. I also want to store the "cell_size = grid.getCellDims" for each connection to avoid computing them from the grid in computeRepRadiusPerfLenght in the WellInterface_impl.hpp
What do you think?

totto82:
jenkins build this please

blattms:
Some parts of the grid will always be a bit different and we might have to live with vanguards working around it. But I agree wherever possible the grid should be used directly.

Concerning the face tag I would recommend to do at least a compile time check like this (untested):
```
constexpr bool supportsFaceTag(const CpGrid&){ return true;}
constexpr bool supportsFaceTag(const PolyHedralGrid&){ return true;}
constexpr bool supportsFaceTag(const ALUGrid<3, 3, Dune::cube, Dune::nonconforming>&){ return true;}
template<class G>
constexpr bool supportsFaceTag(const G&){ return false;}

....
static_assert(supportsFaceTag(grid), "Grid needs to support Facetag");
```

Beware my experience with constexpr is rather limited (and above code might be purely theoretical).

totto82:
jenkins build this please

totto82:
@blattms Something like this you had in mind?

opm/simulators/aquifers/BlackoilAquiferModel_impl.hpp
```diff
@@ -25,6 +25,9 @@ template <typename TypeTag>
 BlackoilAquiferModel<TypeTag>::BlackoilAquiferModel(Simulator& simulator)
     : simulator_(simulator)
 {
+    // Grid needs to support Facetag
+    assert(supportsFaceTag(simulator.vanguard().grid()));
```
akva2, line 29: static_assert - we want this error on compile time.

totto82: static_assert gives 

`non-constant condition for static assertion`

blattms: I am awfully sorry for suggesting something, that I have never tried. Seems we ran into a [gcc bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=77446). 

See my old-fashioned solution in PR totto82/opm-simulators#10

totto82:
jenkins build this please

totto82:
Jenkins fails on `BASE_MSW_WSEGAICD ` which is unrelated to aquifers. This PR moves the computation of cell depths from the problem to the vanguard. Could that be enough to trigger difference in results. If so the testing criteria is either too strict, or their is an underlying instability in the MSW implementation that makes the run sensitive to even changes in calculation order. Can I do a update data and get this in? I believe @akva2 build on this for the MPI support for the aquifers.

alfbr:
The BASE_MSW_WSEGAICD test use the strict tolerance with no command line parameters used to tighten the solution. It is known to be one of the unstable tests. The only significant deviation on the test is Rs, which is unstable for low saturations. Yes, you can do an update_data for this.

totto82:

opm/simulators/aquifers/AquiferCarterTracy.hpp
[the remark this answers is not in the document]
blattms: ok

blattms:
Looks good now from my point of view..

atgeirr:
> Jenkins fails on `BASE_MSW_WSEGAICD ` which is unrelated to aquifers. This PR moves the computation of cell depths from the problem to the vanguard. Could that be enough to trigger difference in results.

I think it is a bit weird that it triggers. The cell depths should be computed in the same way, right? Do you see bitwise identical results or not for other cases, such as Norne?

totto82:
> I think it is a bit weird that it triggers. The cell depths should be computed in the same way, right? Do you see bitwise identical results or not for other cases, such as Norne?

Yes.

totto82:
jenkins build this opm-tests=462 please

totto82:
@atgeirr Can I merge this?

PR #2978, opened 2020-12-10
Use enum Well::Status for well status in WellState

joakim-hove:
In order to fully support #2936 we need to differentiate between `SHUT` wells and `STOP` wells - this is a step on the way. 

Mel: Long and winding road

### DIFF
--- a/opm/simulators/wells/WellState.hpp
+++ b/opm/simulators/wells/WellState.hpp
@@ -225,6 +225,10 @@ namespace Opm
                 this->wellrates_[np * well_index + p] = 0;
         }
 
+        virtual void stopWell(int well_index) {
+            this->status_[well_index] = Well::Status::STOP;
+            this->thp_[well_index] = 0;
+        }
 
         virtual data::Wells report(const PhaseUsage& pu, const int* globalCellIdxMap) const
         {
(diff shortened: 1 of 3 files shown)


joakim-hove:
jenkins build this please

opm/simulators/wells/WellState.hpp
```diff
@@ -230,7 +233,7 @@ namespace Opm
             data::Wells dw;
             for( const auto& itr : this->wellMap_ ) {
                 const auto well_index = itr.second[ 0 ];
-                if (!this->open_for_output_[well_index])
+                if (this->status_[well_index] != Well::Status::OPEN)
```
atgeirr, line 236: I am not sure this is correct. For wells that are STOPPED you would also want output I think (at least bhp makes sense even if rates are 0).

joakim-hove: Well - that is probably right; but the current PR should not change anything; so then this problem has always been there.

atgeirr: I do not think it is a bug now, only that it may become a bug once we start actively using the STOPPED status as a possibility for this flag.

joakim-hove: Ok - then we fully agree, better differentiation between stop and shut is the Next pr.

joakim-hove:
jenkins build this please

joakim-hove:
Can this be merged?

atgeirr:
> Can this be merged?

Yes, since we agree that the lines discussed should be revisited when considering STOP I will merge.

PR #2981, opened 2020-12-10
Skip previously shut wells in WellState init

joakim-hove:
If the well was shut in the previous report step we should not use the previous `WellState` to initialize it for this report step. In the case of wells which have been shut dynamically (due to well testing or ACTIONX) we previously nevertheless initialized the well (with more or less stale values). With this PR we skip the initialization in this case. This will lead to a data update because result will not be identical due to different initialisation. 

When this has been merged - with the necessary data updates, #2936 should be ready directly green in Jenkins :crossed_fingers:

### DIFF
--- a/opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
+++ b/opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
@@ -188,8 +188,12 @@ namespace Opm
                     auto it = prevState->wellMap().find(well.name());
                     if ( it != end )
                     {
-                        const int oldIndex = (*it).second[ 0 ];
                         const int newIndex = w;
+                        const int oldIndex = it->second[ 0 ];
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
@tskille / @jalvestad : Cool if you could compare the IWEL vector from this PR with eclipse?

joakim-hove:

joakim-hove:
jenkins build this opm-tests=459 please

jalvestad:
I have checked this PR with respect to IWEL and it looks OK, as I would have expected. There is no changes to the output of IWEL[10] for a well that has been shut because of an actionx - after the shut operation has been done.

joakim-hove:
OK - since this has been verified in Equinor I will merge it later today; unless someone objects.

PR #2980, opened 2020-12-10
Only compute potentials at beginTimeStep after an event

totto82:
This will need more testing, but is looks promising.

PR #2982, opened 2020-12-11
move the try-catch inside the compute potential loop

totto82:
We still want to try to compute the well potentials for the rest of the wells. 

See https://github.com/OPM/opm-simulators/pull/2980


totto82:
jenkins build this please

atgeirr (approved):
Merge when green.

PR #2976, opened 2020-12-09
Make distributed standard well work on SPE9.

blattms:
This an extension of #2954 that produces correct. It is rebased on current master. though (and includes the other PR). Anyway, to be safe, this should only be merged after #2954 and I will probably need to rebase it before.

Relevant changes are only the last 4 commits.

PR #2983, opened 2020-12-11
use random access to simplify code at second occasion

blattms:
:flush: forgot to stash the second occasion in #2976


blattms:
jenkins build this please

atgeirr (approved):

PR #2984, opened 2020-12-11
Do not include config.h in header (not even in *_impl.hpp)

blattms:
I was getting warnings about redefined definitions (e.g. HAVE_PARMETIS).


blattms:
jenkins build this please

bska (approved):

PR #2987, opened 2020-12-15
Add supports for aquifers in parallel

akva2:
Downstream of https://github.com/OPM/opm-common/pull/2170


opm/simulators/aquifers/AquiferFetkovich.hpp
```diff
@@ -64,8 +64,8 @@ class AquiferFetkovich : public AquiferInterface<TypeTag>
     {
         for (const auto& q : this->Qai_) {
             this->W_flux_ += q * this->ebos_simulator_.timeStepSize();
-            aquifer_pressure_ = aquiferPressure();
         }
+        aquifer_pressure_ = aquiferPressure();
```
totto82, line 68: Is it more readable if this->W_flux_ is communicated here and not in aquiferPressure()

totto82 (approved):

totto82:
This can be merged if the test are green.

ebos/eclbasevanguard.hh
```diff
@@ -801,7 +803,7 @@ protected:
     /*! \brief Mapping between cartesian and compressed cells.
      *  It is initialized the first time it is called
      */
-    std::vector<int> cartesianToCompressed_;
+    std::map<int,int> cartesianToCompressed_;
```
totto82, line 807: why do you need to be a map?

akva2: To avoid a full cart size vector.. ie only store ids for local cells.

blattms:
benchmark opm-common=2170 please

akva2:
there is still something fishy with ct aquifers, so please hold merge.

ebos/eclbasevanguard.hh
blattms: Somebody should check whether using a map somewhere really saves anything. Let us do some back of the envelope calculation:

For each map entry you need to store key and value (for int that should be 8 bytes) + plus the overhead of red-black tree node (for gcc that is another 3 pointers (parent, left, right) + an enum, hence more than 12 bytes per entry).

A vector only needs 4 bytes per entry. Unless we usually have  less than 20% of active cells, std::map seems an odd choice even if performance does not matter at all. We probably can assume that if we have more than 4 processes computing, but below this threshold we are very generous with our memory.

IMHO at some time this needs to done properly using a sorted vector of std::pairs or an unordered_map. Personally, I would prefer the former (but I might be a bit lonely with this opinion), because once the map is set up no further entries will be added.

akva2: it's not just about active cells, it's about cells on this process vs global cartesian cells. as in, the vector is of full cartesian size (of the underlying grid). but anyways, i'll just drop this commit.

opm/simulators/aquifers/AquiferFetkovich.hpp
akva2: W_flux_ is an Evaluation so I cannot easily collect the global values. Thus I see no way to move the reduction.

akva2:
jenkins build this please

opm/simulators/aquifers/AquiferInterface.hpp
```diff
@@ -233,7 +233,99 @@ class AquiferInterface
...
+                    faceDirection = Opm::FaceDir::YPlus;
+                    break;
+                case 4:
+                    faceDirection = Opm::FaceDir::ZMinus;
+                    break;
+                case 5:
+                    faceDirection = Opm::FaceDir::ZPlus;
+                    break;
+                default:
+                    OPM_THROW(Opm::NumericalIssue,
```
atgeirr, line 307: While here anyway, we should make this a logic_error, as it should only arise through programmer error. Also, the error message is misleading, I suggest "Internal error in initialization of aquifer." or something like that.

opm/simulators/aquifers/AquiferCarterTracy.hpp
```diff
@@ -87,94 +90,7 @@ class AquiferCarterTracy : public AquiferInterface<TypeTag>
...
-
-        const double eps_sqrt = std::sqrt(std::numeric_limits<double>::epsilon());
-        for (size_t idx = 0; idx < this->size(); ++idx) {
-            this->alphai_.at(idx) = (denom_face_areas < eps_sqrt)
-                ? // Prevent no connection NaNs due to division by zero
-                0.
-                : (this->connections_[idx].influx_mult * this->faceArea_connected_.at(idx)) / denom_face_areas;
-        }
-    }
+    Scalar fluxValue; // value of flux
```
atgeirr, line 93: Please add trailing underscore in line with other class variables here. Also, the `this->` convention is used for other members, so best to apply it to fluxValue as well.

opm/simulators/aquifers/AquiferInterface.hpp
atgeirr, line 307: This should have been a logic_error all along i believe, and the error message is misleading. Maybe just "Internal error in aquifer setup." or similar.

ebos/eclwriter.hh
```diff
@@ -410,7 +413,8 @@ public:
                                      eclOutputModule_->getBlockData(),
                                      eclOutputModule_->getWBPData(),
                                      localWellData,
-                                     localGroupAndNetworkData);
+                                     localGroupAndNetworkData,
+                                     {});
```
atgeirr, line 417: I do not understand why you do not add the aquifer data here?

atgeirr (commented):
Only minor comments from me. I have read it quite carefully, except for the packunpack parts which I have only scanned superficially.

alfbr:
This has been tested and confirmed correct. Please consider merging as there are other PR's in flight.

alfbr:
jenkins build this please

totto82:
> there is still something fishy with ct aquifers, so please hold merge.

Is this fixed now?

akva2:
Yes

akva2:
That being said there are two unadressed review comments i want to take care of before merge, give me an hour

opm/simulators/aquifers/AquiferCarterTracy.hpp
akva2: done.

ebos/eclwriter.hh
akva2: The aquifer data is only used in evalSummary, it is not used in writeOutput. so there is no need to communicate it here.

opm/simulators/aquifers/AquiferInterface.hpp
akva2: done.

akva2:
jenkins build this please

ebos/eclwriter.hh
atgeirr: Thanks, I assume that is because we only output aquifer data in summary-type output?

bska: For now, yes. Adding aquifer restart output is on my list of things to do in the new year.

akva2:
green, approved, comments adressed.

ebos/collecttoiorank.hh
```diff
@@ -708,19 +708,80 @@ public:
...
+              buffer.read(key);
+              data::AquiferData data;
+              data.read(buffer);
+              auto& aq = globalAquiferData_[key];
+              aq.aquiferID = std::max(aq.aquiferID, data.aquiferID);
+              aq.pressure = std::max(aq.pressure, data.pressure);
+              aq.initPressure = std::max(aq.initPressure, data.initPressure);
+              aq.datumDepth = std::max(aq.datumDepth, data.datumDepth);
+              aq.fluxRate += data.fluxRate;
+              aq.volume += data.volume;
```
GitPaean, line 766: This will makes the parallel restarting more difficult.  Because with this, each process only have its own portion of `volume` (theoretically), when restarting, you only have the one for the whole aquifer, then it is not easy to recover the individual potions anymore. 

I suggest we handle the parallel summation within the simulation code (aquifer model code), then all the processes have the same `volume`. 

`fluxRate` is probably should be the same since we use `fluxRate` to update `volume`.

blattms: I am not sure you are aware that this not connected directly to restart.
Unpackinging will be done on the root process during file output only. It will unpack one result for each process. The result is only written to disk. Hence only the root process needs to have the same result as a serial run (I guess that is assured by taking the max or the sum).

GitPaean: > I am not sure you are aware that this not connected directly to restart.

It is not directly related to restart. But when considering restarting, how you can manage to get some meaningful `volume` for each process to begin with, because `volume` is an accumulative term.  Although at the end, it is the root process result that matters for output, but imagining you have some meaningful information on the root process, and other processes only have some half meaningful information after restarting. I do not think that is pretty.

bska: For what it's worth, the restart file will have certain pieces of information pertaining to each aquifer-reservoir connection in addition to values aggregated to the aquifer as a whole. I haven't fully characterised everything yet. It **may** happen that we need to maintain additional dynamic quantities at the connection level.

PR #2993, opened 2020-12-18
Final fixes to make distributed wells work for Norne.

blattms:
This is tested manually and finally produces the correct result. Currenty, it is still not possible to request that wells might be cut  by process boundaries. I will prepare some more PRs for that later.

Current functionaility should remain unchanged, but there is no an additional copy if we need to access values
from perforations above. If that turns out to slow things down, then we can remove that copy for sequential runs. That would be some work.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -2562,6 +2562,7 @@ namespace Opm
                 well_flux[ebosCompIdxToFlowCompIdx(p)] += cq_s[p].value();
             }
         }
+        this->parallel_well_info_.communication().sum(well_flux.data(), well_flux.size());
     }
 
 
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -1676,6 +1676,8 @@ namespace Opm
                 ipr_b_[ebosCompIdxToFlowCompIdx(p)] += ipr_b_perf[p];
             }
         }
+        this->parallel_well_info_.communication().sum(ipr_a_.data(), ipr_a_.size());
+        this->parallel_well_info_.communication().sum(ipr_b_.data(), ipr_b_.size());
     }
 
 
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -2001,23 +2001,36 @@ namespace Opm
         //    component) exiting up the wellbore from each perforation,
         //    taking into account flow from lower in the well, and
         //    in/out-flow at each perforation.
-        std::vector<double> q_out_perf(nperf*num_comp);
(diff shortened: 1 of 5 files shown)


blattms:
jenkins build this please

blattms:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.001    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.003    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 0.997    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1002

blattms:
@atgeirr In case that was not clear: This was the PR that I meant when saying "quid pro quo" before Christmas.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -2272,6 +2294,12 @@ namespace Opm
             connPI += np;
         }
 
+        //  Sum with communication in case of distributed well.
+        const auto& comm = this->parallel_well_info_.communication();
+        if (comm.size() > 1)
+        {
+            comm.sum(wellPI, np);
+        }
```
bska, line 2302: Is there a reason you did not add this step to the [MultisegmentWell](https://github.com/OPM/opm-simulators/blob/aa38a2264d319e735aade14742f9e4897c13abe8/opm/simulators/wells/MultisegmentWell_impl.hpp#L1238-L1244) case?

blattms: Multisegment well is not treated, yet. Apart from the places where it was needed to satisfy necessary interface changes.

blattms: But they sure look very similar.

atgeirr:
> @atgeirr In case that was not clear: This was the PR that I meant when saying "quid pro quo" before Christmas.

Thanks for the reminder, I'll look at it today afternoon.

atgeirr:
I have read through the code and saw no problems. The only thing I want to ask is about perforation ordering. It was not quite clear to me what the global ordering will be. It is stated to be in the order given from the Schedule, but will that take into account the various possible ordering algorithms? I guess this is as much a question for @joakim-hove as for @blattms!

After reading this, I think we should perhaps make a tiny MPI help library to expand on what we get from Dune, to avoid having to do the "size and displ" dance over and over again, what do you think? (Not for this PR of course.)

I will rerun jenkins just in case, since it was run some time ago.

atgeirr:
jenkins build this please

joakim-hove:
> It is stated to be in the order given from the Schedule, but will that take into account the various possible ordering algorithms? 

The `Schedule`ordering has taken the `COMPORD` keyword into account.

atgeirr:
> The `Schedule`ordering has taken the `COMPORD` keyword into account.

Then all should be fine, thanks.

PR #2995, opened 2020-12-21
don't shut stopped wells

totto82:
We don't want to automatically shut wells stopped by the user. 

Wells stopped due to issues with operability should be triggered by isOperable() if I read the code correctly.


totto82:
jenkins build this please

totto82:
This has been tested and is ready to be merged. @atgeirr Any comments?

atgeirr:
I agree, merging.

PR #2999, opened 2021-01-03
Serialize std::chrono::system_clock::time_point through std::time_t

joakim-hove:
Basic serialization of `std::chrono::system_clock::time_point`- the current PR serializes through `std::time_t`, for general `std::chrono` time_points this will lead to a loss of precision. But since time is currently treated with `std::time_t` in opm this should be sufficient for now.

This PR is required to get https://github.com/OPM/opm-common/pull/2178 through.

### DIFF
--- a/opm/simulators/utils/ParallelRestart.cpp
+++ b/opm/simulators/utils/ParallelRestart.cpp
@@ -22,6 +22,7 @@
 #endif
 
 #include "ParallelRestart.hpp"
+#include <ctime>
 #include <dune/common/parallel/mpitraits.hh>
 
 #define HANDLE_AS_POD(T) \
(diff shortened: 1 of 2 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
Can this be merged?

opm/simulators/utils/ParallelRestart.cpp
```diff
@@ -282,6 +283,13 @@ std::size_t packSize(const RestartValue& data, Dune::MPIHelper::MPICommunicator
         +  packSize(data.extra, comm);
 }
 
+std::size_t packSize(const std::chrono::system_clock::time_point&, Dune::MPIHelper::MPICommunicator comm)
+{
+    std::time_t tp;
```
atgeirr, line 288: Assuming that the size of a time_t does not depend on its contents, I'd say yes: can be merged.

PR #3001, opened 2021-01-05
Output typeid(T).name() for types without packing support

joakim-hove:
Simplify debugging

### DIFF
--- a/opm/simulators/utils/ParallelRestart.hpp
+++ b/opm/simulators/utils/ParallelRestart.hpp
@@ -34,7 +34,9 @@
 #include <optional>
 #include <map>
 #include <set>
+#include <string>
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

opm/simulators/utils/ParallelRestart.hpp
```diff
@@ -34,6 +34,7 @@
 #include <optional>
 #include <map>
 #include <set>
+#include <string>
```
bska, line 37: Technically you need `<typeinfo>` too, if you want to invoke member functions on the object returned from `typeid`.

joakim-hove:
jenkins build this please

PR #3002, opened 2021-01-05
move well operability check for STW

totto82:
We don't want to check the well operability in every inner iterations only in the outer iterations. This also makes the logic for the standard well code the same as the MSW code. Current default is to not use inner iterations for standard wells so it should not affect the simulations running with default options.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -547,6 +547,8 @@ namespace Opm
                    WellState& well_state,
                    Opm::DeferredLogger& deferred_logger)
     {
+        checkWellOperability(ebosSimulator, well_state, deferred_logger);
+
         const bool use_inner_iterations = param_.use_inner_iterations_wells_;
         if (use_inner_iterations) {
             this->iterateWellEquations(ebosSimulator, B_avg, dt, well_state, deferred_logger);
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

totto82:
@GitPaean This should be a trivial cleanup. Can you review/merge it?

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -574,9 +576,6 @@ namespace Opm
     {
         // TODO: only_wells should be put back to save some computation
         // for example, the matrices B C does not need to update if only_wells
-
-        checkWellOperability(ebosSimulator, well_state, deferred_logger);
-
         if (!this->isOperable() && !this->wellIsStopped()) return;
```
GitPaean, line 579: If there is nothing changed related to operability, you probably want to move this line together.

totto82: Since assembleWellEqWithoutIteration can be called directly and not only from assembleWellEq I think it makes sense to add the operability guard here.

GitPaean: Okay. That is sensible regarding to this aspect, to have this line to use function  `assembleWellEqWithoutIteration` by itself. 

Then how about adding this  line after `checkWellOperability` in function `assembleWellEq`. It is clear that if the well is not operable, then nothing afterwards is needed in the function `assebleWellEq`. 

No strong opinion here. Just let know whether you think it is sensible.  It is merge-able either way to me.

totto82: I dont think it is necessary since `assembleWellEq` just calls `assembleWellEqWithoutIteration` or `iterateWellEquations` which also starts with calling  `assembleWellEqWithoutIteration`.

GitPaean: Okay. I just thought code will be easier to understand/read. Anyway, I will merge this.

PR #3003, opened 2021-01-06
dynamically set zoltan imbalance tolerance

totto82:
Gives the user a hint and a way to fix the empty partition issue by using a stricter imbalance tolerance. 

Depends on OPM/opm-grid#507


totto82:
jenkins build this opm-grid=507 please

atgeirr:
I have a very minor nitpick (no reason to use Scalar instead of double for the new parameter), but not enough to hold back a merge. Thanks!

PR #3008, opened 2021-01-08
refactoring the pressure assemble equations for ICD

GitPaean:
to reduce some code duplication. It should not change the results in any way.


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

bska:
I like it.  Unless @atgeirr wants to say something about this I suggest we merge it as is.  Reducing code complexity is always welcome.

atgeirr:
> I like it. Unless @atgeirr wants to say something about this I suggest we merge it as is. Reducing code complexity is always welcome.

Sorry for not commenting earlier, I have no reason to go against this, refactoring is good. Thanks for the patch and the review, I'll merge now.

PR #3020, opened 2021-01-14
Vfp refactor

joakim-hove:
Have also simplified a bit here and there.

Downstream of: https://github.com/OPM/opm-common/pull/2210


joakim-hove:
jenkins build this please

PR #3022, opened 2021-01-20
Don't test serialization for isolated ScheduleState

joakim-hove:
The serialization of the `ScheduleState`objects is handled from the `Schedule` class - the isolated `ScheduleState` instances are not serialized completely - and can therefor not be tested in isolation.


joakim-hove:
jenkins build this please

PR #3023, opened 2021-01-20
Add diffusion in Flow

totto82:
This PR adds 
1) code to compute diffusivity in the eclTransmissibility class. Only called if diffusion is enabled. i.e. should not affect the standard Flow. 
2) Adds optional oil-gas-diffusion and bo-diffusion variants of Flow. 

Depends on OPM/opm-models#640

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -349,6 +349,16 @@ opm_add_test(flow_co2
   DEPENDS opmsimulators
   LIBRARIES opmsimulators)
 
+opm_add_test(flow_bo_diffusion
+  ONLY_COMPILE
+  DEFAULT_ENABLE_IF ${FLOW_VARIANTS_DEFAULT_ENABLE_IF}
+  SOURCES
+  flow/flow_bo_diffusion.cpp
+  $<TARGET_OBJECTS:moduleVersion>
+  EXE_NAME flow_bo_diffusion
+  DEPENDS opmsimulators
+  LIBRARIES opmsimulators)
+
 opm_add_test(flow_onephase_energy
   ONLY_COMPILE
   DEFAULT_ENABLE_IF ${FLOW_VARIANTS_DEFAULT_ENABLE_IF}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -339,13 +339,13 @@ opm_add_test(flow_onephase
   DEPENDS opmsimulators
   LIBRARIES opmsimulators)
 
-opm_add_test(flow_co2
+opm_add_test(flow_co2_diffusion
   ONLY_COMPILE
   DEFAULT_ENABLE_IF ${FLOW_VARIANTS_DEFAULT_ENABLE_IF}
   SOURCES
   flow/flow_co2_diffusion.cpp
   $<TARGET_OBJECTS:moduleVersion>
-  EXE_NAME flow_co2
+  EXE_NAME flow_co2_diffusion
   DEPENDS opmsimulators
   LIBRARIES opmsimulators)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -339,6 +339,16 @@ opm_add_test(flow_onephase
   DEPENDS opmsimulators
   LIBRARIES opmsimulators)
 
+opm_add_test(flow_co2
+  ONLY_COMPILE
+  DEFAULT_ENABLE_IF ${FLOW_VARIANTS_DEFAULT_ENABLE_IF}
+  SOURCES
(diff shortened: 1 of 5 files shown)


hnil (commented):
should one extract porosity here or add it to intensivequentities to enable  more couplings?
Is the computehalfDiffusivity_ compatible with transmissibilites. In principle the same code and logic should be  used?

For 1e-30 for avoiding diffusion is also hard code, is there some genneralmechanism for this. This values depend on particular problems allthough 1e-30 seems safe.

totto82:
> should one extract porosity here or add it to intensivequentities to enable more couplings?
For now (all) transmissibilities are computed as a pre-processing step. I think it could stay that way until we have a use case where we need more dynamic couplings. Note that tortuosity uses the porosity from the intensive quantities. 

> Is the computehalfDiffusivity_ compatible with transmissibilites. In principle the same code and logic should be used?
I tried to refactor the code so that I could use the same but the Tensor property of the Permeability needs some special treatment. In the end I concluded that the current code was the best.  Currently multipliers etc. are not supported for diffusion. If this is implemented in the future we may want to refactoring to avoid code duplication. 


> For 1e-30 for avoiding diffusion is also hard code, is there some genneralmechanism for this. This values depend on particular problems although 1e-30 seems safe.
This is a copy from the transmissibility code. I don't know if it is needed. Don't we define cells with porosity 0 to be inactive. But I added it to stay in sync with the other code and to avoid numerical issues from near zero porosity cells.

totto82:
jenkins build this opm-material=443 opm-models=640 please

bska:
> jenkins build this opm-material=443 opm-models=640 please

Just for the record, it is generally better to trigger the build check from the most upstream repository&mdash;in this case OPM-material.  The reason is that doing so will get the greatest test coverage.  Jenkins only runs the unit and regression tests in the repository from which we trigger the build and that repository's downstream repositories.

totto82:
> Just for the record, it is generally better to trigger the build check from the most upstream repository—in this case OPM-material. The reason is that doing so will get the greatest test coverage. Jenkins only runs the unit and regression tests in the repository from which we trigger the build and that repository's downstream repositories.

Good point.  For this I started an instance from all the repository to assure every test is triggered. Jenkins seems to be happy with the PRs.

CMakeLists.txt
```diff
@@ -339,6 +339,26 @@ opm_add_test(flow_onephase
...
+
+opm_add_test(flow_bo_diffusion
+  ONLY_COMPILE
+  DEFAULT_ENABLE_IF ${FLOW_VARIANTS_DEFAULT_ENABLE_IF}
+  SOURCES
+  flow/flow_bo_diffusion.cpp
+  $<TARGET_OBJECTS:moduleVersion>
+  EXE_NAME flow_bo_diffusion
+  DEPENDS opmsimulators
+  LIBRARIES opmsimulators)
```
bska, line 360: @akva2 : Is this the right way to enable new specialised simulators?  What, if anything, would we need change in the build system in order to incorporate these into the main `flow` executable if we wanted to?

ebos/eclproblem.hh
```diff
@@ -3072,6 +3085,8 @@ private:
     struct PffDofData_
     {
         Opm::ConditionalStorage<enableEnergy, Scalar> thermalHalfTrans;
+        Opm::ConditionalStorage<enableDiffusion, Scalar> diffusivity;
+        //Scalar diffusivity;
```
bska, line 3089: Is this comment supposed to be a hint/explanatory?  If not, maybe we can drop it altogether?

bska (commented):
Looks good to me.  I just had one question concerning the CMake changes and I've asked @akva2 to provide expert opinion on that.  Once that's resolved I'm in favour of merging this.

CMakeLists.txt
totto82: My reasoning for keeping them out of `flow ` at the moment is that I would like to test the performance drawback (if any) of always enabling diffusion. You will still need to provide the DIFFUSE keyword for it to have an effect, but compile time enabling flow may have computational overhead that needs investigation. If the overhead is small I would like to make a PR that enables diffusion by default. If the overhead is too large I would like to add the specialized version to  `flow ` in a later PR. The reason for adding them here as separate binaries is to simplify the testing.

ebos/eclproblem.hh
totto82: Just a leftover. I will remove it.

bska:
Build check triggered from upstream PR is clean.  I'll merge this now.  Thanks a lot!

PR #3024, opened 2021-01-20
Ensure Well Defined Initial Values for CT Data Members

bska:


bska:
jenkins build this please

totto82:
jenkins build this please

opm/simulators/aquifers/AquiferCarterTracy.hpp
```diff
@@ -89,8 +89,8 @@ class AquiferCarterTracy : public AquiferInterface<TypeTag>
     const AquiferCT::AQUCT_data aquct_data_;
     Scalar beta_; // Influx constant
     // TODO: it is possible it should be a AD variable
-    Scalar mu_w_; // water viscosity
-    Scalar fluxValue_; // value of flux
+    Scalar mu_w_{1}; // water viscosity
+    Scalar fluxValue_{0}; // value of flux
```
totto82, line 94: The problem seems to be that the fluxValue_ is used before it is set. This will not fix the restart case

GitPaean: Okay. That makes sense. 

`fluxValue_` is assigned in the function `endTimeStep()`, but it is used in `calculateEqnConstants`?

Is this the problem? 

I was looking at `mu_w_`, I did not think it was problematic.

totto82: https://github.com/OPM/opm-simulators/blob/9552d0cc8a62f243fe47c0b0879224b47f010835/opm/simulators/aquifers/AquiferInterface.hpp#L102

I think fluxValue_ = W_flux_ needs to be added here.

bska: > I think fluxValue_ = W_flux_ needs to be added here.

There is no `fluxValue_` member in the `AquiferInterface`.  That data member is specific to the Carter-Tracy implementation.

totto82: Ok then it should be added to the assignRestartData method

bska: > Ok then it should be added to the assignRestartData method

Possibly.  There is currently [no way](https://github.com/OPM/opm-simulators/blob/9552d0cc8a62f243fe47c0b0879224b47f010835/opm/simulators/aquifers/AquiferCarterTracy.hpp#L95-L99) to restart a simulation run using Carter-Tracy aquifers so we'll need to quite a bit more work in order to gain that ability.

totto82: Ok. Then I think this should be merged as is.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
I'm going to merge this now.  There may still be other issues in this code, but this does at least make the tests reliably succeed.

PR #3025, opened 2021-01-21
fix well model for gasoil thermal

totto82:
This PR contains two fixes
1) make sure the correct well  indices are used for gasoil+energy
2) fix 2 phase for WELLPI by making sure the fluidstate is accessed by the canonical phase index and not the active phase position provided by the phaseUsage.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -4170,7 +4170,8 @@ namespace Opm
             // Note: E100's notion of PI value phase mobility includes
             // the reciprocal FVF.
             const auto connMob =
-                mobility[ flowPhaseToEbosCompIdx(p) ].value() * fs.invB(p).value();
+                mobility[ flowPhaseToEbosCompIdx(p) ].value()
+                    * fs.invB(flowPhaseToEbosPhaseIdx(p)).value();
 
             connPI[p] = connPICalc(connMob);
         }
@@ -4227,6 +4228,6 @@ namespace Opm
 
         const auto zero   = EvalWell { this->numWellEq_ + this->numEq, 0.0 };
         const auto mt     = std::accumulate(mobility.begin(), mobility.end(), zero);
-        connII[phase_pos] = connIICalc(mt.value() * fs.invB(phase_pos).value());
+        connII[phase_pos] = connIICalc(mt.value() * fs.invB(flowPhaseToEbosPhaseIdx(phase_pos)).value());
     }
 } // namespace Opm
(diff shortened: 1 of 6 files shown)


totto82:
jenkins build this please

totto82:
I went for the simplest solution of using a mapping from the phase index from phaseUsage and the one used in opm-models. Synchronizing the indexes used by the wellState and opm-models should be a prioritized cleanup to avoid issues like this, IMO. @atgeirr Any chance for prioritizing the wellstate refactoring you started on some time ago?

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -4170,7 +4170,8 @@ namespace Opm
             // Note: E100's notion of PI value phase mobility includes
             // the reciprocal FVF.
             const auto connMob =
-                mobility[ flowPhaseToEbosCompIdx(p) ].value() * fs.invB(p).value();
+                mobility[ flowPhaseToEbosCompIdx(p) ].value()
+                    * fs.invB(flowPhaseToEbosPhaseIdx(p)).value();
```
bska, line 4174: Cool.  We probably need the same fix in the [MultisegmentWell](https://github.com/OPM/opm-simulators/blob/9552d0cc8a62f243fe47c0b0879224b47f010835/opm/simulators/wells/MultisegmentWell_impl.hpp#L4337-L4339) implementation too.

totto82: Thanks for spotting, I will update with a fix to the MSW.

totto82:
jenkins build this please

totto82:
The failed test runs fine locally. I will try again.

totto82:
jenkins build this please

bska:
> The failed test runs fine locally. I will try again.

I've seen that failure in other contexts recently.  You **may** want to try incorporating #3024 too.

totto82:
> I've seen that failure in other contexts recently. You may want to try incorporating #3024 too.
We should get in #3024 in asap.

bska:
> > I've seen that failure in other contexts recently. You may want to try incorporating #3024 too.
> 
> We should get in #3024 in asap.

Kind of.  That PR definitely makes the test go through, but I don't know if I'm just papering over some underlying issue since I don't know the aquifer implementation well enough.

bska:
I think this looks good, but maybe @GitPaean would like to have a look before we merge this?

GitPaean:
> I think this looks good, but maybe @GitPaean would like to have a look before we merge this?

It looks good to me.  Especially the more clear way to handle `numWellEq`.

bska:
> It looks good to me. Especially the more clear way to handle `numWellEq`.

Thanks a lot for looking at this.  Then we merge this.

PR #3028, opened 2021-01-25
get connection length and radius from the connection object

totto82:
Depends on OPM/opm-common#2244

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -394,7 +394,7 @@ namespace Opm {
 
             void setupCartesianToCompressed_(const int* global_cell, int local_num__cells);
 
-            void computeRepRadiusPerfLength(const Grid& grid, Opm::DeferredLogger& deferred_logger);
+            void setRepRadiusPerfLength(Opm::DeferredLogger& deferred_logger);
 
 
             void computeAverageFormationFactor(std::vector<Scalar>& B_avg) const;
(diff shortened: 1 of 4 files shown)


totto82:
jenkins build this opm-common=2244 please

GitPaean:
The changes introduced by this PR I can see mostly the following two: 
1. now the equivalent radius and connection length will be calculated always
2. these calculations are based on EclipseGrid instead of processed CpGrid. It might be different for some complicated situation, but we do not have such test case either. 

The above are just observation and comments. Not against this PR since we do not have a case to make the above difference.

totto82:
> now the equivalent radius and connection length will be calculated always

It was already calculated in the wellConnections so this set of PRs just store them in the connection class instead of recalculating them. 

>these calculations are based on EclipseGrid instead of processed CpGrid. It might be different for some complicated situation, but we do not have such test case either.

The only thing I could think of is MINPV-PINCH. What happens to a connection that perforate a cell that is pinched-out?

GitPaean:
> It was already calculated in the wellConnections so this set of PRs just store them in the connection class instead of recalculating them.

Yes. That is new in the PR from opm-common

> The only thing I could think of is MINPV-PINCH. What happens to a connection that perforate a cell that is pinched-out?

Currently, we remove that connection from the EclipseState.  I am not familiar with MINPV-PINCH, will it change the thickness of a cell during PINCH process? That is the tricky part. But I am not sure it should make in the way of the PR.  Just for commenting and discussion (if interested), possibly helpful if in future we have problem related to this(I highly doubt it.).

totto82:
> But I am not sure it should make in the way of the PR. 

@GitPaean  Does that mean that you approve this PR?

GitPaean:
> @GitPaean Does that mean that you approve this PR?

Sure. I can do that. We can revisit it if it causes problem.

GitPaean:
jenkins build this please

PR #3029, opened 2021-01-25
Activate distributed wells in the simulator

blattms:
We introduce a new parameter --enable-distributed-wells=<true|false> for this. During startup we check that the model either only has standard wells or that multisegement wells are actively interpreted as standard wells (by way of passing --enable-multisegment-wells=false as an option).

Note that the typetag property `UseMultisegmentWells` is now there twice once for `TypeTag` of `EclBaseVanguard` as template parameter and once for `TTag::FlowModelParameters` as template parameter. If there is a better way, please let me know. This drove me crazy and it seemed the only way to make all tests compile.

Needs OPM/opm-grid#511

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -175,6 +175,12 @@ template<class TypeTag>
 struct AllowDistributedWells<TypeTag, TTag::EclBaseVanguard> {
     static constexpr bool value = false;
 };
+
+// Same as in BlackoilModelParametersEbos.hpp but for here.
+template<class TypeTag>
+struct UseMultisegmentWell<TypeTag, TTag::EclBaseVanguard> {
+    static constexpr bool value = true;
+};
 } // namespace Opm::Properties
 
 namespace Opm {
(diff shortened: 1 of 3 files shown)


ebos/eclbasevanguard.hh
```diff
@@ -444,6 +466,45 @@ public:
...
+
+            if (hasMsWell)
+            {
+                if (myRank == 0)
+                {
+                    std::string message = std::string("Option --allow-distributed-wells=true is only allowed ")
+                        + "if model only has only standard wells."
+                        + "You need to provide option "
+                        + " with --enable-multisegement-wells=false to treat "
+                        + "existing multisegment wells as standard wells.";
```
joakim-hove, line 501: This will be a really long line in the log - add a newline somewhere?

blattms: Good idea. Changed.

blattms:
Triggered a build from OPM/opm-grid#511.
I would really like to merge this stuff and it should not affect flow with default parameters at all. Hence I might interpret @totto82's approval of #3032  as one for  this PR (as it is included there) and merge this myself tomorrow afternoon. Grant that the triggered test becomes green.

blattms:
Tested upstream

PR #3031, opened 2021-01-26
Removing explicit usage of the Grid in the relperm diagnostics

totto82:


opm/simulators/flow/FlowMainEbos.hpp
```diff
@@ -545,35 +544,6 @@ namespace Opm
...
-            if (FluidSystem::numActivePhases() > 1) {
-                RelpermDiagnostics diagnostic;
-                if (mpi_size_ > 1) {
-#if HAVE_MPI
-                    this->grid().switchToGlobalView();
-                    static_cast<ParallelEclipseState&>(this->eclState()).switchToGlobalProps();
-#endif
-                }
-                diagnostic.diagnosis(eclState(), this->grid());
-                if (mpi_size_ > 1) {
```
totto82, line 568: Why was the global grid passed to the diagnostics here? I have tested this PR both in serial and in parallel and it seems to work perfectly without a similar logic.

totto82:
jenkins build this please

opm/simulators/flow/FlowMainEbos.hpp
akva2: the diagnostics needs the entire grid to do the checks. if you only pass the local part, you get *different* diagnostics.

totto82: >  you get different diagnostics.

Can you point me where the difference occur? I though it just checked the tables and the endpoints.

akva2: The details alludes me, it was a year ago or so. But you got different output on parallel and serial norne. In particular you got some warnings in parallel that did not spew in serial.

totto82: ok, I will checkout norne, thanks

akva2: Note that this could also be related to the mapper. Maybe we didn't have the local mapper available or something. I just remember I did this change to yield the same results in parallel and serial (https://github.com/opm/opm-simulators/pull/2309 - not a lot of additional info there..)

totto82: Norne does not give any warnings from the relperm diagnostics neither in serial or in parallel. Maybe @bska has fixed the underlying issue  during his re-factoring?

akva2: that is certainly possible. if you get the same results and are confident then just go ahead. i know it fixed an issue back then, but if it's the same now..

totto82: The relperm diagnostics for Model 2 gives the same output for this PR and master both in serial and in parallel. 

I however notices that the simulator gives to many warnings like 

> Warning: Equilibration region 1 has no active cells

in parallel. Since it gives a warning if a partition has no active cells. I will provide a fix in another PR.

bska: > The relperm diagnostics for Model 2 gives the same output for this PR and master both in serial and in parallel.

Very good.  In that case I'm satisfied that this PR is good to go.  I'll merge into master.

PR #3033, opened 2021-01-27
Fix empty region output in parallel 

totto82:
only warn about empty regions when none of the partitions contains the region

### DIFF
--- a/ebos/equil/initstateequil.hh
+++ b/ebos/equil/initstateequil.hh
@@ -1819,14 +1819,14 @@ private:
         auto psat   = PhaseSat { materialLawManager, this->swatInit_ };
         auto vspan  = std::array<double, 2>{};
 
+        std::vector<int> regionIsEmpty(rec.size(), 0);
         for (size_t r = 0; r < rec.size(); ++r) {
             const auto& cells = reg.cells(r);
 
             Details::verticalExtent(cells, cellZMinMax_, comm, vspan);
 
             if (cells.empty()) {
-                Opm::OpmLog::warning("Equilibration region " + std::to_string(r + 1)
-                                     + " has no active cells");
+                regionIsEmpty[r] = 1;
                 continue;
             }
 
@@ -1850,6 +1850,14 @@ private:
                 // Horizontal subdivision
                 this->equilibrateHorizontal(cells, eqreg, -acc,
                                             ptable, psat);
+            }
+        }
+        comm.min(regionIsEmpty.data(),regionIsEmpty.size());
+        if (comm.rank() == 0) {
+            for (size_t r = 0; r < rec.size(); ++r) {
+                if (regionIsEmpty[r]) //region is empty on all partitions
+                    Opm::OpmLog::warning("Equilibration region " + std::to_string(r + 1)
+                                         + " has no active cells");
             }
         }
     }


totto82:
jenkins build this please

ebos/equil/initstateequil.hh
```diff
@@ -1819,14 +1819,14 @@ private:
         auto psat   = PhaseSat { materialLawManager, this->swatInit_ };
         auto vspan  = std::array<double, 2>{};
 
+        int regionIsEmpty[rec.size()] = {0};
```
bska, line 1822: Tiny nit: This should probably be a `std::vector<int>` instead.  Variable-length arrays are a GCC extension.

totto82: The ` comm.sum(regionIsEmpty,rec.size())` seems to prefer ` int[] `to `std::vector<int>`. At least it did not compile when I tried to use `std::vector<int>`

akva2:
```
comm.summ(&vec[0], vec.size());
```

ebos/equil/initstateequil.hh
```diff
@@ -1852,6 +1852,12 @@ private:
                                             ptable, psat);
             }
         }
+        comm.sum(regionIsEmpty,rec.size());
+        for (size_t r = 0; r < rec.size(); ++r) {
+            if (regionIsEmpty[r] > 0 )
+                Opm::OpmLog::warning("Equilibration region " + std::to_string(r + 1)
+                                     + " has no active cells");
```
bska, line 1859: Does this mean that every MPI process will emit all these messages now or is `OpmLog::warning()` aware of IO ranks?

bska: Another thing: The above will emit a diagnostic if a region is empty on at least one process.  Might it make sense to emit the diagnostic only if the region is empty on **all** processes?

ebos/equil/initstateequil.hh
blattms: Indeed Dune only provides very limited support for containers. It is a rather thin wrappper around the MPI C bindings (hence the vectors). For vector simply use
```
comm.sum(vec.data, vec.size());
```
which will also work (without valgrind complaining) if the vector is empty.

ebos/equil/initstateequil.hh
blattms: Yes and no.
Yes the message is printed on all processes. We might want to use  `if(comm.rank() == 0)`before the for-loop.
No, it will only emit the message if all processes have an empty region (note the `comm.sum(...)` above).

bska: > it will only emit the message if all processes have an empty region (note the `comm.sum(...)` above).

Then I'm missing something.  Suppose region 3 is empty on rank (process) 4 (i.e., `regionIsEmpty[3] = 1` on rank 4, but `= 0` on all other ranks).  Won't the `comm.sum()` then ensure that `regionIsEmpty[3] = 1` on all processes/ranks?

totto82: I think @bska is right. It should be `if(regionIsEmpty[r] == numRegions) `

bska: > It should be `if(regionIsEmpty[r] == numRegions)`

Or possibly
```C++
if (regionIsEmpty[r] == comm.size())
```
instead.

blattms: Indeed, you guys are right. I just assumed the value of the variable and its name would correlate.

Nitpicking:
Hence `regionIsEmpty` actually does `numProcsRegionIsEmpty`.
Maybe rename or change the computation to resemble what the name seems to indicate
```
comm.min(regionIsEmpty.data(), regionIsEmpty.size());
...
if(regionIsEmpty[r])
  ...
```

bska: > Maybe rename or change the computation to resemble what the name seems to indicate

That's a good idea.  I agree that `comm.min()` is the most suitable reduction here.

totto82:
Thanks for the comments. I tested this on Norne with 4 ranks and it no longer warns about empty regions in parallel.

totto82:
jenkins build this please

bska:

PR #3036, opened 2021-01-28
changed: this is known on compile time

akva2:
use info to disable code.

granted, extremely minor optimization but may as well apply it.


akva2:
jenkins build this please

bska (commented):
Cool&mdash;merge when green.  I especially like that we've removed a member that's not actually used at all.

bska:

PR #3037, opened 2021-01-29
reset the iteration count in the newtonSolver

totto82:
This hopefully fix seg-fault reported in #3000

PR #3038, opened 2021-01-29
Reset Non-Linear Iteration Count Before 'beginTimeStep'

bska:
While here, also fix a couple of misprints in a nearby comment.

### DIFF
--- a/opm/simulators/flow/BlackoilModelEbos.hpp
+++ b/opm/simulators/flow/BlackoilModelEbos.hpp
@@ -238,15 +238,15 @@ namespace Opm {
                 ebosSimulator_.model().advanceTimeLevel();
             }
 
-            // set the timestep size and episode index for ebos explicitly. ebos needs to
-            // know the report step/episode index because of timing dependend data
-            // despide the fact that flow uses its own time stepper. (The length of the
-            // episode does not matter, though.)
+            // Set the timestep size, episode index, and non-linear iteration index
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

totto82:
For me the test failures looks reasonable if it caused by #3037.  @blattms commented: 

> I was a bit too fast for jenkins. At least I checked the diff and it seem very ok.. Anyway fingers crossed.

Does that mean that Jenkins was not green when #3037 was merged.

akva2:
jenkins wasn't green because he merged before the job finished, but he was lucky in that there were no failures.

totto82:
Thanks for the clarification. Then we need to understand why this fails.

totto82:
I think it should be safe to do a reference update on this, but I will do a manual check on the last two failures to be sure that it does not have unwanted side effects.

totto82:
```
mpi.compareECLFiles_flow+9_4B_WINJ_VREP-W_STW
mpi.compareECLFiles_flow+MOD4_UDQ_ACTIONX
```
has visible changes in the results, but they are qualitatively the same. It shows that NUPCOL impacts the results. I will initiate an update to get this fix in.

totto82:

bska:
jenkins build this opm-tests=475 please

bska:
jenkins build this opm-tests=475 please

bska:
New reference solutions have been installed on the CI system.  Merging this now to bring the system back to a consistent state.

PR #3040, opened 2021-02-01
Avoid division by small rate in divisor in VFP table fraction WFR/GFR.

totto82:
A threshold of 1e-8 is used to "define" small rate

I have tested this on the norne prediction models where it significantly reduce problems, run-time and unwanted peaks in the solution. Setting the "correct" thresholds is not trivial and we may want adjust this value a bit if it has unwanted impact on some cases. i.e. this PR should be tested carefully.

### DIFF
--- a/opm/simulators/wells/VFPHelpers.hpp
+++ b/opm/simulators/wells/VFPHelpers.hpp
@@ -37,43 +37,15 @@
 namespace Opm {
 namespace detail {
 
+static double threshold = 1e-12;
 
 /**
- * Returns zero if input value is NaN of INF
+ * Returns zero if input value is negative
  */
-inline double zeroIfNanInf(const double& value, const std::string& rate, const VFPProdTable& table) {
-    const bool nan_or_inf = std::isnan(value) || std::isinf(value);
-
-    if (nan_or_inf) {
-        std::string msg = "Problem with VFPPROD keyword table " + std::to_string(table.getTableNum()) + "\n" +
-                          "In " + table.location().filename + " line " + std::to_string(table.location().lineno) + "\n"
-                          "Nan or inf encountered while calculating " + rate + " - using zero";
-        OpmLog::warning("NAN_OR_INF_VFP", msg);
-    }
-
-    return nan_or_inf ? 0.0 : value;
+template <typename T>
+static T chopNegativeValues(const T& value) {
+    return Opm::max(0.0, value);
 }
-
-
-/**
- * Returns zero if input value is NaN or INF
- */
-template <class EvalWell>
-inline EvalWell zeroIfNanInf(const EvalWell& value, const std::string& rate, const VFPProdTable& table) {
-    const bool nan_or_inf = std::isnan(value.value()) || std::isinf(value.value());
-
-    if (nan_or_inf) {
-        std::string msg = "Problem with VFPPROD keyword table " + std::to_string(table.getTableNum()) + "\n" +
-            "In " + table.location().filename + " line " + std::to_string(table.location().lineno) + "\n"
-            "Nan or inf encountered while calculating " + rate + " - using zero";
-        OpmLog::warning("NAN_OR_INF_VFP_EVAL", msg);
-    }
-
-    using Toolbox = MathToolbox<EvalWell>;
-
-    return nan_or_inf ? Toolbox::createConstantZero(value) : value;
-}
-
 
 /**
  * Computes the flo parameter according to the flo_type_
@@ -137,18 +109,16 @@ template <typename T>
 static T getWFR(const VFPProdTable& table, const T& aqua, const T& liquid, const T& vapour) {
     auto type = table.getWFRType();
     switch(type) {
-    case VFPProdTable::WFR_TYPE::WFR_WOR:
+    case VFPProdTable::WFR_TYPE::WFR_WOR: {
         //Water-oil ratio = water / oil
-        return zeroIfNanInf(aqua/liquid, "WOR", table);
-
+        return chopNegativeValues(-aqua) / Opm::max(threshold, chopNegativeValues(-liquid));
+    }
     case VFPProdTable::WFR_TYPE::WFR_WCT:
         //Water cut = water / (water + oil)
-        return zeroIfNanInf(aqua / (aqua + liquid), "WCT", table);
-
+        return chopNegativeValues(-aqua) / Opm::max(threshold, chopNegativeValues(-aqua - liquid));
     case VFPProdTable::WFR_TYPE::WFR_WGR:
         //Water-gas ratio = water / gas
-        return zeroIfNanInf(aqua / vapour, "WGR", table);
-
+        return chopNegativeValues(-aqua) / Opm::max(threshold, chopNegativeValues(-vapour));
     default:
         throw std::logic_error("Invalid WFR_TYPE");
     }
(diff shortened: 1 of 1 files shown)


opm/simulators/wells/VFPProdProperties.cpp
```diff
@@ -215,7 +215,7 @@ calculateBhpWithTHPTarget(const std::vector<double>& ipr_a,
 
     std::vector<detail::RateBhpPair> ratebhp_samples;
     for (size_t i = 0; i < flo_samples.size(); ++i) {
-        ratebhp_samples.push_back( detail::RateBhpPair{flo_samples[i], bhp_flo_samples[i]} );
+        ratebhp_samples.push_back( detail::RateBhpPair{-flo_samples[i], bhp_flo_samples[i]} );
     }
```
totto82, line 219: @atgeirr Can you double check this change? I think the current code is wrong, but I may have miss something.

totto82:
jenkins build this please

totto82:
benchmark please

ytelses:
Benchmark result overview: 




| Test    | Configuration                                 | Relative |
| ------- | --------------------------------------------- | -------- |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1    | 1.003    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8    | 1        |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1    | 0.989    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8    | 0.99     |
| opm-git | OPM Benchmark: flow_mpi_norne_4c - Threads: 1 | 1.024    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c - Threads: 8 | 0.704    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1056

GitPaean:
A question here, are there some THP/VFP related in the benchmarking cases?

alfbr:
> A question here, are there some THP/VFP related in the benchmarking cases?

Yes, for the Norne 4C case, you find it here:
https://github.com/OPM/opm-tests/blob/master/norne/NORNE_ATW2013_4C_MSW.DATA
In particular, see the SCHEDULE part here:
https://github.com/OPM/opm-tests/blob/master/norne/INCLUDE/PRED_WINJ_MIN_THP_4C_MSW.SCH

GitPaean:
> Yes, for the Norne 4C case, you find it here:

Thanks for the explanation.

opm/simulators/wells/VFPProdProperties.cpp
GitPaean: Not sure if I see it correctly.  It does not looks like this function is used anywhere now.

GitPaean: If I read correctly, with the iterative solver, this intersection algorithm is not used anymore. With the iterative solver, the good thing is that the IPR is not used directly. And accurate IPR was a problem there.

totto82: > Not sure if I see it correctly. It does not looks like this function calculateBhpWithTHPTarget is used anywhere now.¨

That may explain why what looks like a wrong sign could have survived in the code for some time, and explains why I didn't notice any affect of changing this.

totto82:
> opm-git | OPM Benchmark: flow_mpi_norne_4c - Threads: 8 | 0.704

This definitely needs more investigation. I will report back.

opm/simulators/wells/VFPProdProperties.cpp
GitPaean: Referring to the comment above,
```
    // kind of unncessarily following the tradation that producers should have negative rates
    // the key is here that it should be consistent with the function bhpwithflo
```

Not sure it is wrong here, or whether it matters here. It is a very old code and it must succeed with some VFP/THP control cases before. I will hesitate to say the sign is necessarily wrong or whether it matters, or other later changes cause inconsistency.  If it were wrong from the very beginning, it would cause big issues much earlier.  But I might oversee something and be wrong here.

totto82: This is how I read the code. 

When the production rate is computed it is positive
Then

```
   // kind of unncessarily following the tradation that producers should have negative rates
    // the key is here that it should be consistent with the function bhpwithflo
    for (double& value : flo_samples) {
        value = -value;
    }

```

which makes it negative since bhpwithflo expects negative production rates. 

Then it is stored as a pair in `ratebhp_samles`, which again is used in `findIntersectionForBhp`. 

But as I read it. `findIntersectionForBhp ` expects positive signed rates

@GitPaean I would very much like you to go through the same code and check if you agree with my understanding. The constant flipping of signs may have "flipped" my brain as well.

GitPaean: I am not saying you are wrong, excuse me if you feel that way. 
I was trying to put a breakpoint to check how things work there, then it never got reached. Then I saw it is not used anymore. 

Your reading is probably correct.  With a very quick go through, can you tell me why `findIntersectionForBhp` needs positive signed rates?

totto82: > I am not saying you are wrong, excuse me if you feel that way.

Not at all. I really appreciate your eyes on this. I am not 100% sure myself I would not like this to be merged before one more developer has looked carefully at this. 


> then it never got reached.

If this is stale code. I suggest we just delete it. 

The signs of the fluxes in `findIntersectionForBhp ` needs to be consistent. flo_bhp_limit and flo_bhp_middle is positive I think i.e. flow_samles should be the same.

totto82:
jenkins build this please

totto82:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1.002    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 1.004    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 1.022    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 0.744    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1112

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
@atgeirr What do you think about this? I think this gives a better treatment of cases where the divisor is zero. In particular I think it is unfortunate that we return zero. If the divisor goes to zero we should look at the other side of the table. I would not worry to much about the 4C benchmark. It still has serious unsolved problems that makes the performance oscillates a lot. I would very much like to get this in soon.

totto82:
jenkins build this please

alfbr:
Among the four test failures, the first two are noise. The second two are large enough to warrant a look at the results.

atgeirr:
I hesitate to comment on the test failures, as I do not know the test cases very well. I can partake in the discussion if wanted.

About the code, I think it is fine, and could be merged. (I am a little hesitant about the use of abs throughout, as while that may protect this particular part from sign errors, it also would hide them. Each quantity here should have a well-defined sign, and while things are complicated by some parts expecting negative production rates while others expect all positive numbers, I am not sure it helps in the long run to abs() it out of existence rather than apply a minus sign where appropriate.)

totto82:
> I am not sure it helps in the long run to abs() it out of existence rather than apply a minus sign where appropriate

I see your point. The problem is that injectors can produce and producers and inject epsilons of fluids from time to time. This was a simple way of fixing that. I will try to fix the back-flow issue independently and pass values with switched signs.

totto82:
jenkins build this please

totto82:
I have manually checked the failure. The results are qualitatively the same. The difference is caused by small changes that results in slightly changed timestepping. I will start the update process.

totto82:

bska:
jenkins build this opm-tests=539 please

bska:
This PR was approved by a maintainer and the updated reference solutions have been installed on the CI system.  I'll merge this into master now.

PR #3044, opened 2021-02-03
correcting the output of the number of threads

GitPaean:
Currently, when running flow, no matter what number you set to `--threads-per-process`, you always get two threads for output,  unless you have a single-core machine. 

```
Using 1 MPI processes with 2 OMP threads on each 
```

This PR is purely for correcting above output. 

At the same time, in the file `FlowMainEbos.hpp` and `threadmanager.hh`, there are multiple places dealing with the nubmer of threads, even with slightly different algorithm, which is not easy to follow, even it should be a simple thing. 

Some refactoring should be done regarding to this.  Maybe let `ThreadManager` deal with it only can be a solution,  then we need to set it up before we `printBanner`.

### DIFF
--- a/opm/simulators/flow/FlowMainEbos.hpp
+++ b/opm/simulators/flow/FlowMainEbos.hpp
@@ -307,6 +307,11 @@ namespace Opm
                 threads =  omp_get_max_threads();
             else
                 threads = std::min(2, omp_get_max_threads());
+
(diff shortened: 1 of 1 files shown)


GitPaean:
jenkins build this please

blattms:
Cool.Thanks.

PR #3045, opened 2021-02-04
Don't Initialize Well State for Shut Wells

bska:
This is a rebased and trimmed-down version of the earlier PR #2936.  In particular, this PR prepares for the case of having the well state incorporate all wells on a process, not just the open/active ones.

This supersedes and closes #2936.

### DIFF
--- a/opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
+++ b/opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
@@ -213,7 +213,6 @@ namespace Opm
                 auto end = prevState->wellMap().end();
                 for (int w = 0; w < nw; ++w) {
                     const Well& well = wells_ecl[w];
-                    const int num_perf_this_well = well_perf_data[w].size();
                     auto it = prevState->wellMap().find(well.name());
                     if ( it != end )
                     {
@@ -258,6 +257,17 @@ namespace Opm
                         // perfPhaseRates
                         const int oldPerf_idx_beg = (*it).second[ 1 ];
                         const int num_perf_old_well = (*it).second[ 2 ];
+                        const auto new_iter = this->wellMap().find(well.name());
+                        if (new_iter == this->wellMap().end()) {
+                            throw std::logic_error {
+                                well.name() + " is not in internal well map - "
+                                "Bug in WellStateFullyImplicitBlackoil"
+                            };
+                        }
+
+                        const int connpos = new_iter->second[1];
+                        const int num_perf_this_well = new_iter->second[2];
+
                         int num_perf_changed = (num_perf_old_well != num_perf_this_well) ? 1 : 0;
                         num_perf_changed = parallel_well_info[w]->communication().sum(num_perf_changed);
                         bool global_num_perf_same = (num_perf_changed == 0);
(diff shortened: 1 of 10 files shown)


bska:
jenkins build this please

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
```diff
@@ -229,17 +249,24 @@ namespace Opm
...
+                        const auto new_iter = this->wellMap().find(well.name());
+                        if (new_iter == this->wellMap().end()) {
+                            throw std::logic_error {
+                                well.name() + " is not in internal well map - "
+                                "Bug in WellStateFullyImplicitBlackoil"
+                            };
+                        }
+
+                        const int connpos = new_iter->second[1];
+                        const int num_perf_this_well = new_iter->second[2];
```
blattms, line 261: What do the number of connections in the wellMap reflect?
Are they aware of parallelism (i.e. that the number connection on the local grid might differ from the number of connections in a serial run? If not, then this will cause problems for distributed wells.

bska: > What do the number of connections in the `wellMap` reflect?

Neither more nor less than what's [implied](https://github.com/OPM/opm-simulators/blob/009feaa35df69aa10902388b082b2f42bf0b010d/opm/simulators/wells/WellState.hpp#L84-L92) by the [`well_perf_data_`](https://github.com/OPM/opm-simulators/blob/009feaa35df69aa10902388b082b2f42bf0b010d/opm/simulators/wells/BlackoilWellModel_impl.hpp#L734-L736).  This PR just switches to using the value we already calculated rather than having to rederive it from `well_perf_data_[w].size()`.

> Are they aware of parallelism (i.e. that the number connection on the local grid might differ from the number of connections in a serial run?

No idea.  Unless you took special care to handle that when you introduced the distributed wells then probably not.

blattms: Thanks, I checked it. According to [`BlackoilWellModel::initializeWellPerfData`](/OPM/opm-simulators/blob/33e3102983287345c62c1158f839351add02fab5/opm/simulators/wells/BlackoilWellModel_impl.hpp#L637-L669) this should be ok.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
I'm resetting the state of this PR back to "work in progress" since we need to merge #3093 first.

bska:
The upstream prerequisite #3093 was merged.  I'm marking this as "ready for review" an launching a build check.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

totto82 (approved):
The cleanups in the PR look good and the code should do what the title promise. I assume the motivation is related to output.

bska:
> I assume the motivation is related to output.

Only indirectly.  This work is a precursor to #3094 which enables using `WELPI` **before** the wells become active.  In order to do that I've chosen to include **all** wells in the well model's [`wells_ecl_`](https://github.com/OPM/opm-simulators/blob/0468b2eee6d8915006a56ce2c7ce097d77e4a3d0/opm/simulators/wells/BlackoilWellModel.hpp#L285) list and this PR will then essentially skip the `SHUT` wells when creating the well state object.

For output purposes, this PR does not do anything in particular.

totto82:
I see. Thanks for the contribution. I will merge.

PR #3049, opened 2021-02-05
Pass unscaled newWellPI to Schedule::applyWellProdIndexScaling()

joakim-hove:
See: https://github.com/OPM/opm-common/pull/2276


bska (approved):
Merging this now to chase upstream semantic change to API.

PR #3050, opened 2021-02-07
Support combination of ACTIONX and WELPI

joakim-hove:


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
```diff
@@ -1047,6 +1047,15 @@ namespace Opm
             return productivity_index_;
         }
 
+        double productivityIndex(const std::string& well) const {
+            auto well_iter = this->wellMap().find(well);
+            if (well_iter == this->wellMap().end())
+                throw std::logic_error("Tried to get PI for unknown well: " + well);
+
+            return this->productivity_index_[well_iter->second[0]];
```
bska, line 1055: Just for future reference as you develop this feature further: The `productivity_index_` vector has [`pu.num_phases`](https://github.com/OPM/opm-simulators/blob/0056d4e9c124da01123cd9b3905e7c0970a76423/opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp#L173) elements per well.  The one that's supposed to be used in the `WELPI` keyword is the one [pertaining](https://github.com/OPM/opm-simulators/blob/0056d4e9c124da01123cd9b3905e7c0970a76423/opm/simulators/wells/BlackoilWellModel_impl.hpp#L2572-L2601) to the well's preferred phase.

joakim-hove: Thanks!

joakim-hove:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -1329,8 +1377,16 @@ public:
...
+
+                const auto& wellpi = this->fetchWellPI(reportStep, *action, schedule, matching_wells);
+
+                schedule.applyAction(reportStep, std::chrono::system_clock::from_time_t(simTime), *action, actionResult, wellpi);
                 actionState.add_run(*action, simTime);
+
+                for ( const auto& [wname, _] : wellpi) {
+                    (void)_;
+                    this->wellModel_.updateEclWell(reportStep, wname);
+                }
```
bska, line 1389: This looks like it will call `updateEclWell()` even in the case that the actual action is not `WELPI`.  Am I missing something?

joakim-hove: I'll check

joakim-hove: Ehh - no the `wellpi` map should only contains wells which are mentioned in the ACTIONX clause in a ActionxX keyword; maybe a better name for `fetchWellPI` would be `fetchWellPI_for_ACTIONX_wells` - although that is a mouthfull. Or do you think I am in the wrong here?

bska: > the `wellpi` map should only contains wells which are mentioned in the ACTIONX clause in a ActionxX keyword;

Okay, I guess it comes down to the details.  Suppose we have
```
ACTIONX
A 1 /
WPI PROD01 > 10.0E+3 /
/

WELTARG
PROD01 LRAT 5000 /
/
ENDACTIO

ACTIONX
B 1
WOPR PROD2 < 150 /
/

WELPI
PROD3 400 /
/
ENDACTIO

ACTIONX
C 1 /
WPI PROD4 > 5000 /
/

WELPI
PROD4 4500 /
/
ENDACTIO
```
Which wells will be in `wellpi` if these actions trigger?

joakim-hove: You don't come up with simple examples do you :grinning:? Anyway I'll give it a shot:

**Action A:** Here the action statement does not contain the `WELPI` keyword and the `wellpi` map will be empty. The fact that the `WPI:PROD01` variable is accessed in the condition clause is irrelevant in this context, that has been evaluated in the summary code and made available in `SummaryState` prior to evaluating actions.

**Action B:** This action mentions the well `PROD3` in the `WELPI` keyword and we will have `wellpi = {"PROD3" : ???}`

**Action C:** This action mentions the well `PRO4` in the `WELPI` keyword and we will have `wellpi = {"PROD4" : xxxx}`

Modulo bugs.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

bska:
> jenkins build this please

This will fail.  You'll need to add the new member functions to the EBOS [`WellManager`](https://github.com/OPM/opm-simulators/blob/591826ea4e523a5678feb93be3c1a25f61459368/ebos/eclwellmanager.hh#L65) too.

joakim-hove:
>This will fail. You'll need to add the new member functions to the EBOS WellManager too.

Thanks - I guess I don't build the EBOS code locally.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

ebos/eclwellmanager.hh
```diff
@@ -620,6 +620,42 @@ public:
     }
 
 
+    void
+    updateEclWell(int timeStepIdx, int well_index)
+    {
+        const auto& schedule = this->simulator_.vanguard().schedule();
+        const auto& wname = this->wells_[well_index]->name();
+        this->wells_[well_index] = std::make_shared<Well>( schedule.getWell(wname, timeStepIdx) );
```
totto82, line 628: This does not look correct. Well from schedule is not
` typedef Opm::EclPeacemanWell<typeTag> Well;`

I quick fix is just to throw. And alternative is to add a method to the PeacemanWell that updated the eclWell. 

the `ebos_plain ` that uses this well model is probably not used by anybody and could IMO be removed.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

ebos/eclproblem.hh
joakim-hove: @bska : Are you reasonably happy with the answer here?

ebos/eclproblem.hh
```diff
@@ -1290,6 +1291,53 @@ public:
...
+                for (int rank=1; rank < comm.size(); rank++) {
+                    for (std::size_t well_index=0; well_index < num_wells; well_index++) {
+                        const auto global_index = rank*num_wells + well_index;
+                        const auto value = wellpi_buffer[global_index];
+                        if (value != 0)
+                            wellpi_vector[well_index] = value;
+                    }
+                }
+            }
+            comm.broadcast(wellpi_vector.data(), wellpi_vector.size(), 0);
```
bska, line 1329: @blattms : Would you mind taking a look at this communication pattern please?  The objective is to communicate/distribute `num_wells` scalar values between all processes (ranks), with the additional provision that each process owns a subset of the values and that that subset may be partially replicated on other processes in the case of distributed wells.

bska (commented):
I've read the PR carefully and with possible exception of the PI value synchronisation I am happy with this.  I've asked @blattms to take a closer look at the synchronisation and once that's approved we'll merge this.

joakim-hove:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2583,6 +2583,17 @@ namespace Opm {
...
     }
 
+    template<typename TypeTag>
+    void
+    BlackoilWellModel<TypeTag>::
+    updateEclWell(int timeStepIdx, const std::string& wname) {
+        auto well_iter = std::find_if( this->wells_ecl_.begin(), this->wells_ecl_.end(), [wname] (const auto& well) -> bool { return well.name() == wname;});
+        if (well_iter == this->wells_ecl_.end())
+            throw std::logic_error("Could not find well: " + wname);
+
```
blattms, line 2593: Same code is in this file at least another time. Maybe use add a functions for reuse?

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2619,6 +2630,16 @@ namespace Opm {
         }
     }
 
+    template<typename TypeTag>
+    double
+    BlackoilWellModel<TypeTag>::
+    wellPI(const std::string& well_name) const
+    {
+        auto well_iter = std::find_if(this->wells_ecl_.begin(), this->wells_ecl_.end(), [&well_name](const Well& well) { return well.name() == well_name; });
+        auto well_index = std::distance( this->wells_ecl_.begin(), well_iter );
```
blattms, line 2639: Another time, but this time even with a check whether this is at the end. See above,

joakim-hove:
jenkins build this please

ebos/eclproblem.hh
blattms: Communication seems correct, but seems a bit suboptimal (too much communication and too much space). Improvements should be made in a subsequent PR.

Seems like the goal here is that fetchWellPI returns a global vector (one entry for each well whether it is on the process or not) on each process. That is certainly safe, but we should ask whether that is needed (if actions on wells not on the process really influence the computation on the process.

blattms (changes requested):
I am not inisting on changes (besides the check), but I think it would be fairly reasonable to do them anyway

opm/simulators/wells/BlackoilWellModel_impl.hpp
joakim-hove: I have added the check here; have not created a separate function as that would need to expose iterator type in the API -  and I am not really comfortable with that (yet anyway) ...

joakim-hove:
> Seems like the goal here is that fetchWellPI returns a global vector (one entry for each well whether it is on the process or not) on each process. 

That is correct.

> [....] but we should ask whether that is needed (if actions on wells not on the process really influence the computation on the process.

First of all - this is not exactly inner loop; it is something applies in the case where people use the combination of WELLPI and ACTIONX - and when that actually happens it will be on the order O(10) times for a simulation - so I argue it is not very important for the total runtime. That said I agree that it might be possible to actually proceed without pushing identical value to all processors, but then we suddenly let go of the promise that the `Schedule`object is identical on all processes - and that I consider quite high risk. 

So: all in all I have responded to the review with one additional check, and some moaning. Have rebased and pushed an update.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

blattms (approved):

blattms:
Sine @bska already approved, I will merge.

PR #3051, opened 2021-02-09
Allow usage of non-gpu devices with OpenCL.

blattms:
This is needed to use the CPU backend of PoCL http://portablecl.org/ which seems handy for debugging and probably other things.


blattms:
@Tongdongq Can you please check this?

blattms:
jenkins build this please

Tongdongq:
So if I understand correctly there are 2 changes: all opencl devices on the chosen platform are being found, and only the chosen one is kept (previously all found devices were kept).
Looks good, but I didn't test it

blattms:
Previously the only platform allow was GPU preventing using the code e.g. on the GPU.
As before we only use one chosen device (here the change is only semantics.

Please test it by running.

PR #3056, opened 2021-02-10
Pass true to FieldPropertieManager::keys to include TRANS Modifiers.

blattms:
These seem not be listed anymore by default and opm-common has changes to reallow listing them if true is passed.
Needs OPM/opm-common#2289 and closes OPM/opm-common#2282


joakim-hove:
Not needed any longer.

PR #3059, opened 2021-02-11
Do not report confusing error

alfbr:
Removed a confusing error message from LOG/PRT. @tskille please see if this fixes your issue.

### DIFF
--- a/opm/simulators/wells/MSWellHelpers.hpp
+++ b/opm/simulators/wells/MSWellHelpers.hpp
@@ -27,6 +27,8 @@
 #include <opm/simulators/utils/DeferredLogger.hpp>
 #include <opm/common/ErrorMacros.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/MSW/SICD.hpp>
+#include <opm/common/OpmLog/OpmLog.hpp>
+#include <string>
 #include <dune/istl/solvers.hh>
 #if HAVE_UMFPACK
 #include <dune/istl/umfpack.hh>
@@ -64,7 +66,9 @@ namespace mswellhelpers
         for (size_t i_block = 0; i_block < y.size(); ++i_block) {
             for (size_t i_elem = 0; i_elem < y[i_block].size(); ++i_elem) {
                 if (std::isinf(y[i_block][i_elem]) || std::isnan(y[i_block][i_elem]) ) {
-                    OPM_THROW_NOLOG(Opm::NumericalIssue, "nan or inf value found after UMFPack solve due to singular matrix");
+                    const std::string msg{"nan or inf value found after UMFPack solve due to singular matrix"};
+                    OpmLog::debug(msg);
+                    OPM_THROW_NOLOG(Opm::NumericalIssue, msg);
                 }
             }
         }
--- a/opm/simulators/wells/MSWellHelpers.hpp
+++ b/opm/simulators/wells/MSWellHelpers.hpp
@@ -64,7 +64,7 @@ namespace mswellhelpers
         for (size_t i_block = 0; i_block < y.size(); ++i_block) {
             for (size_t i_elem = 0; i_elem < y[i_block].size(); ++i_elem) {
                 if (std::isinf(y[i_block][i_elem]) || std::isnan(y[i_block][i_elem]) ) {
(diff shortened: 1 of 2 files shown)


tskille:
Thanks @abir, confusing error messages gone.

alfbr:
Good, thanks for testing!

alfbr:
jenkins build this please

alfbr:
This is ready for review from my end.

GitPaean:
Hi, with this way, the error message will not show up anywhere, include .DBG file.  I think we should have this information in the .DBG file to help the developers. 

Maybe we should have a `OpmLog::debug()` before we `OPM_THROW_NOLOG` if we do not want develop new facility? Any suggestions @bska @joakim-hove ?

bska:
> Maybe we should have a `OpmLog::debug()` before we `OPM_THROW_NOLOG` if we do not want develop new facility?

Yes, that'd be the least effort alternative here.  That way we at least get a trace of the problem in one of the
```
CASE.N.DBG
```
files (or just `CASE.DBG` if rank zero ends up in this situation).

bska:
> That way we at least get a trace of the problem in one of the

We still have the problem that we don't terminate if one process/rank gets stuck after throwing an exception though

GitPaean:
> Yes, that'd be the least effort alternative here. That way we at least get a trace of the problem in one of the

Thanks for the input. 

> We still have the problem that we don't terminate if one process/rank gets stuck after throwing an exception though

Yes. It will be more involved and out of the scope of this PR.

GitPaean:
> We still have the problem that we don't terminate if one process/rank gets stuck after throwing an exception though

I can try to do it in a separate PR based on DeferredLogger.

opm/simulators/wells/MSWellHelpers.hpp
```diff
@@ -64,7 +64,7 @@ namespace mswellhelpers
         for (size_t i_block = 0; i_block < y.size(); ++i_block) {
             for (size_t i_elem = 0; i_elem < y[i_block].size(); ++i_elem) {
                 if (std::isinf(y[i_block][i_elem]) || std::isnan(y[i_block][i_elem]) ) {
-                    OPM_THROW(Opm::NumericalIssue, "nan or inf value found after UMFPack solve due to singular matrix");
+                    OPM_THROW_NOLOG(Opm::NumericalIssue, "nan or inf value found after UMFPack solve due to singular matrix");
```
GitPaean, line 67: Referring to the discussion in the comments, the following is suggested here, 

```c++
        const std::string msg{"nan or inf value found after UMFPack solve due to singular matrix"};
        OpmLog::debug(msg);
        OPM_THROW_NOLOG(Opm::NumericalIssue, msg);
```

You need to add the following header file
```
#include <opm/common/OpmLog/OpmLog.hpp>
#include <string>
```

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -3185,7 +3185,7 @@ namespace Opm
 
             const EvalWell d = 1.0 - rs * rv;
             if (d <= 0.0 || d > 1.0) {
-                OPM_THROW(Opm::NumericalIssue, "Problematic d value " << d << " obtained for well " << name()
```
GitPaean, line 3188: Similar to the previous comment, the following code is suggested, 
```c++
                std::ostringstream sstr;
                sstr << "Problematic d value " << d << " obtained for well " << name()
                     << " during conversion to surface volume with rs " << rs
                     << ", rv " << rv << " and pressure " << seg_pressure
                     << " obtaining d " << d;
                OpmLog::debug(sstr.str());
                OPM_THROW_NOLOG(Opm::NumericalIssue, sstr.str());
```

GitPaean (changes requested):
We need to keep some debugging output in the .DBG file. So `OpmLog::debug()` is suggested to call before the `OPM_THROW_NOLOG` call.

alfbr:
jenkins build this please

alfbr:
Review comments addressed and jenkins seems happy.

GitPaean:
> Review comments addressed and jenkins seems happy.

Thanks for the efforts. I am merging it now.

PR #3061, opened 2021-02-12
Better handling of non- or partially supported keywords

verveerpj:
**Aim**
The pull request will implement a better approach for handling missing keywords and options and provide better diagnostics.

**Current issue**
In the current implementation, while reading the deck, the keywords that are encountered are checked. If they are in a list of not-supported keywords, a warning is emitted, but the simulator will proceed.

For a few supported keywords, the value of some options is checked in the same manner. In the current implementation, if an option is checked, only a single valid value can be specified.

**Desired behaviour**
- Instead of only emitting a warning for all non-supported keywords, a distinction should be made between critical keywords and non-critical keywords. In the former case, the simulator should abort with an error, in the latter case a warning is sufficient.
- The same is true for options: if an option is not supported, it should be determined if it is critical or not before aborting or warning the user. For options, the supported values should be checked. There may be a sub-set of more than one value that is allowed, which is not supported in the current checker.
- Diagnostic messages should be improved. Maybe an optional message can be emiited to specific messages/options. 
  **Is there a use case for this or are a few generic messages sufficient?**
- Instead of just running the simulator and aborting/warning, an option should be added for "linting": instead of aborting or running the simulations, problems will be collected and reported, so that the user can inspect the issues in a user-friendly format.

**Ideas for further improvements**
- Are certain combinations of keywords and/or options disallowed?

**Implementation details**
- The list of unsupported keywords and options should be supplied in some declarative form. Two options:
  - Use a sufficiently simple structure that can be entered as literals in an include file.
  - Use the JSON format to provide the information, much like the parser seems to do to specify keywords.
- Currently the check is done only in the readDeck function, so presumably it should be sufficient to replace that function call.
- An command line option needs to be added to provide a "linting" mode.


joakim-hove:
I think this task sounds interesting and would like to offer some over engineering advice from the trenches - feel free to ignore :smile:: 

I would create a class `CheckDeck` which takes a `Deck` instance and has methods like `CheckDeck::lint()` and `CheckDeck::strict_check()` and so on. The `CheckDeck` class should not require any code level dependency on datastructures in opm-simulators; since the knowledge of the simulator capabilities is in the opm-simulators repository the eventual configuration/instantiation of the `CheckDeck` class must be there - but from a purely technical point of view it should be possible to implement it in opm-common.

One possibility would be to have a base class which in addition to the `::lint()` and `::check()` methods also have *configuration methods* like `::unsupported_keyword( const std::string& keyword, bool fatal)` and `::valid_option( keyword, item, set_of_values )` that way the main functionality (which will be quite closely tied to the `Deck`datastructure) can be in opm-common, and the configuration can be in opm-simulators.

I'll shut up now - feel free to reach out if you think these ideas are interesting.

verveerpj:
> I think this task sounds interesting and would like to offer some over engineering advice from the trenches - feel free to ignore :
> 
> I would create a class `CheckDeck` which takes a `Deck` instance and has methods like `CheckDeck::lint()` and `CheckDeck::strict_check()` and so on. The `CheckDeck` class should not require any code level dependency on datastructures in opm-simulators; since the knowledge of the simulator capabilities is in the opm-simulators repository the eventual configuration/instantiation of the `CheckDeck` class must be there - but from a purely technical point of view it should be possible to implement it in opm-common.
> 
> One possibility would be to have a base class which in addition to the `::lint()` and `::check()` methods also have _configuration methods_ like `::unsupported_keyword( const std::string& keyword, bool fatal)` and `::valid_option( keyword, item, set_of_values )` that way the main functionality (_which will be quite closely tied to the `Deck` datastructure_) can be in opm-common, and the configuration can be in opm-simulators.
> 
> I'll shut up now - feel free to reach out if you think these ideas are interesting.

Putting the main functionality in `opm-common` sounds like a good idea. I will try to structure it like that.

alfbr:
The plan looks great to me, it will bring us a large step further in making Flow useful to a broader audience. I have no strong opinions about whether the functionality should reside in opm-common or opm-simulators, but it seems we all agree that configuration should be in opm-simulators since it is simulator specific (we may actually have more than one configuration in the future).

verveerpj:
I see that there is a `checkDeck` function  in opm-common which is called in the `readDeck` function used by the simulators, right after the missing features are checked. So it may make sense to change that into a class and add functionality for missing features there.

However, I will start by just building the missing features functionality directly in `opm-simulators` and then we can see afterwards if we unify that with the `checkDeck` functionality in `opm-common`.

opm/simulators/flow/KeywordValidation.cpp
```diff
@@ -0,0 +1,310 @@
+/*
+  Copyright 2016 Statoil ASA.
```
joakim-hove, line 2: 2021 Equinor

opm/simulators/flow/KeywordValidation.cpp
```diff
@@ -0,0 +1,310 @@
...
+           {"WSEGMULT", {false, {}}}, {"WSEGPROP", {false, {}}},   {"WSEGPULL", {false, {}}},
+           {"WSEGSEP", {false, {}}},  {"WSEGSOLV", {false, {}}},   {"WSEGTABL", {false, {}}},
+           {"WSURFACT", {false, {}}}, {"WTADD", {false, {}}},      {"WTEMPQ", {false, {}}},
+           {"WTHPMAX", {false, {}}},  {"WTMULT", {false, {}}},     {"WVFPDP", {false, {}}},
+           {"WVFPEXP", {false, {}}},  {"WWPAVE", {false, {}}},     {"ZIPPY2", {false, {}}},
+           {"ZIPP2OFF", {false, {}}}};
+
+    void checkDeck(const Deck& deck, const ParseContext& parse_context, ErrorGuard& errorGuard)
+    {
+        for (size_t idx = 0; idx < deck.size(); idx++) {
```
joakim-hove, line 290: If you don't need the index in the loop this can be a range based for:
```
for (const auto& keyword : deck) {
    ...
}
```

opm/simulators/flow/KeywordValidation.hpp
```diff
@@ -0,0 +1,50 @@
...
+class ErrorGuard;
+class ParseContext;
+
+namespace KeywordValidation
+{
+
+    struct UnsupportedKeywordProperties {
+        bool critical;
+        std::optional<std::string> message;
+    };
```
joakim-hove, line 41: I guess this also needs some datastructures geared towards unsupported items; for an unsupported/partially supported item I guess you can choose between an allowed_values_list, illegal_values_list, only_default - or others. This might get quite low level (and ugly ...) in terms of opm-common/Deck fiddling.

opm/simulators/flow/KeywordValidation.cpp
```diff
@@ -0,0 +1,310 @@
...
+#include <opm/parser/eclipse/Parser/ParseContext.hpp>
+#include <opm/simulators/flow/KeywordValidation.hpp>
+
+namespace Opm
+{
+
+namespace KeywordValidation
+{
+
+    std::map<std::string, UnsupportedKeywordProperties> unsupported_keywords
```
joakim-hove, line 31: I realize this is still WIP - and there are of course many ways to go about this, but I would suggest creating a class:
```
class KeywordValidator {

...
private:
    std::map<std::string, UnsupportedKeywordProperties>unsupported_keywords;
};
```
And then build up that instance with unsupported keywords runtime (or have derived class `FlowKeywordValidator`which adds the default unsupported keywords in the constructor or ....). The advantages of this is that it will be simple to configure the validator runtime, it will be simple to have different validators for different simulators/situations and finally the validator can be tested that way.

verveerpj: Indeed, after solidifying the approach a bit, it will become a class, in case we would like this to be reusable.

verveerpj:
I have now implemented a class `KeywordValidator` that will check for missing features and checks partially supported options. It takes specifications for this in the form of a C++ map of appropriate structures. These are defined now as variables in separate include functions, that are passed to the validator. Thus, different validators may check for different sets of keywords/options if that is desired.

We can extend this with further functionality as desired, but some other things need to be done first:
- Unit tests should be written.
- Currently the validator is created and used in `readDeck `function. This is not optimal, since that means that every simulator that uses `readDeck`  gets the exact same validator. Instead, the `readDeck` function should acquire a new parameter to pass a validator. The validator can then be instantiated for `flow` in its main function file.
- The validator code itself is now defined in the `flow` source directory. That code should probably move to `opm-common`.

opm/simulators/utils/PartiallySupportedKeywordOptions.hpp
```diff
@@ -0,0 +1,104 @@
...
+PartiallySupportedOptions<std::string> partially_supported_string_options = {
+    {
+        "COMPORD",
+        {
+            {
+                "ORDER_TYPE",
+                {
+                    .critical = false,
+                    .permitted_values = {"INPUT"},
+                    .message = std::nullopt,
```
bska, line 38: I'm sorry, but you can't use this.  We don't have C++20 and therefore cannot use designated initialisers.

verveerpj: Ah, pity,, then we will use list-initialization for now. Can see later if we need something more friendly.

bska:
jenkins build this please

verveerpj:
It is now at a point that I am happy with the implementation and the tests. Now the following happens when you run flow:
- Keywords or options that are not supported are detected.
- Two reports are made:
  - One that reports all problems that are not critical. This is reported as a warning.
  - One that reports all problems that are critical. This is reported as an error.
- The warnings (if any) are displayed before the errors (if any). If there are errors, the program terminates.

Thus, the user now has a formatted report of all unsupported keywords and options. If there are not critical errors, the simulator will run. If there are, the program terminates, and the user can inspect the error to find all unsupported keywords.

alfbr:
Thank you! I will start testing this as soon as Easter vacation is over. Consider removing WIP from the pull request title.

I suggest one design change, and I hope everybody with an opinion on it chimes in. That is allowing the parser to finish the whole deck, also in case of missing critical features, i.e., errors in the sense of this pull request. This will allow a summary report at the end informing the user of all unsupported functionality found in the deck. I believe such an extension can be part of another pull request, so it should not hold this pull request back unless you prefer taking it in here.

bska (commented):
I haven't looked at the details of this, but I have one high-level question about the implementation.  Why did you decide to put the tables of partially supported and fully unsupported keywords into header files, even ones that are not protected against multiple inclusion?  That's very risky as there will be multiply defined symbols if we ever include tables in two translation units.

I would have thought that putting the tables in `.cpp` files and providing (read-only) access to them through a function interface would be a safer approach.

verveerpj:
> I haven't looked at the details of this, but I have one high-level question about the implementation. Why did you decide to put the tables of partially supported and fully unsupported keywords into header files, even ones that are not protected against multiple inclusion? That's very risky as there will be multiply defined symbols if we ever include tables in two translation units.
> 
> I would have thought that putting the tables in `.cpp` files and providing (read-only) access to them through a function interface would be a safer approach.

You are correct, that was some unfinished business. The inclusion protection was just forgotten, so I have added that now. 

Regarding multiple inclusion: The tables are constants and the include files are local files for use by `readDeck.cpp`. They are not intended for inclusion in other files and that is why they are not in an include directory. The tables are passed to the validation object as parameters. Putting them in the `KeywordValidation` namespace was not a good idea, so I moved them to a local namespace. Multiple tables for usage in other translation units can be defined with different names/namespaces.

verveerpj:
> I suggest one design change, and I hope everybody with an opinion on it chimes in. That is allowing the parser to finish the whole deck, also in case of missing critical features, i.e., errors in the sense of this pull request. This will allow a summary report at the end informing the user of all unsupported functionality found in the deck. I believe such an extension can be part of another pull request, so it should not hold this pull request back unless you prefer taking it in here.

This is already the current behavior. All errors will be reported, critical or not. If there are any critical errors, then after reporting the program will halt.

alfbr:
> This is already the current behavior. All errors will be reported, critical or not. If there are any critical errors, then after reporting the program will halt.

Thank you, yes I can see that now. I am in the process of testing, looks great so far! Did notice one fall-out when all warnings and errors are reported, the file path for where the keyword or option is found is only given for the first unsupported keyword found (same path repeated for warnings and errors). Give me a ping if you want sample datasets to see it.

verveerpj:
> Thank you, yes I can see that now. I am in the process of testing, looks great so far! Did notice one fall-out when all warnings and errors are reported, the file path for where the keyword or option is found is only given for the first unsupported keyword found (same path repeated for warnings and errors). Give me a ping if you want sample datasets to see it.

Ah, that was by design, but now I understand that may have been incorrect. I assumed that all parsed keywords reside in a single file, so I thought it would be sufficient to report the file location once. However, could it be that there are multiple files? Then it would be better to always print it.

alfbr:
> Ah, that was by design, but now I understand that may have been incorrect. I assumed that all parsed keywords reside in a single file, so I thought it would be sufficient to report the file location once. However, could it be that there are multiple files? Then it would be better to always print it.

Yes, they are typically scattered across multiple include files, even nested includes have bee seen in the wild. That is actually the main reason why listing file location is helpful.

verveerpj:
> > Ah, that was by design, but now I understand that may have been incorrect. I assumed that all parsed keywords reside in a single file, so I thought it would be sufficient to report the file location once. However, could it be that there are multiple files? Then it would be better to always print it.
> 
> Yes, they are typically scattered across multiple include files, even nested includes have bee seen in the wild. That is actually the main reason why listing file location is helpful.

Ok, will fix that and print the file at every warning/error.

goncalvesmachadoc:
jenkins build this please

bska (commented):
Thanks for the updates, especially the include guards.  Personally I'd probably still put the keyword tables in `.cpp` files with an accessor function, but I'm not going to insist.

I do have one question concerning the representation of the partially supported keywords though and I'd appreciate it if you could address that please.  If the current representation does not allow for that kind of value check, how much work would it be to enable it?

bska:
jenkins build this please

opm/simulators/flow/KeywordValidation.cpp
```diff
@@ -0,0 +1,153 @@
...
+            if (err.critical == critical) {
+                if (err.item_name && err.item_value) {
+                    std::string message = fmt::format(option_format, *(err.item_value), *(err.item_name));
+                    report.append(OpmInputError::format(message, err.location));
+                } else {
+                    report.append(OpmInputError::format(keyword_format, err.location));
+                }
+                report.append(OpmInputError::format(location_format, err.location));
+                if (err.user_message) {
+                    report.append("  NOTE: " + *(err.user_message) + "\n");
```
alfbr, line 54: Please remove 'NOTE: ' from the message, so that it reads:
`report.append("  " + *(err.user_message) + "\n");`

opm/simulators/flow/KeywordValidation.cpp
```diff
@@ -0,0 +1,153 @@
...
+namespace Opm
+{
+
+namespace KeywordValidation
+{
+
+    std::string get_error_report(const std::vector<ValidationError>& errors, const bool critical)
+    {
+        const std::string keyword_format = "  {keyword}: keyword not supported\n";
+        const std::string option_format = "  {{keyword}}: invalid value '{}' for option {}\n";
```
alfbr, line 40: Please replace 'for option' with 'item number'

opm/simulators/utils/PartiallySupportedFlowKeywords.hpp
```diff
@@ -0,0 +1,104 @@
...
+namespace Opm::FlowKeywordValidation
+{
+
+
+const Opm::KeywordValidation::PartiallySupportedKeywords<std::string> partially_supported_kewyords_strings = {
+    {
+        "COMPORD",
+        {
+            {
+                "ORDER_TYPE",
```
alfbr, line 55: Rather than item name, it is better to use item number here. That way we will also avoid having to rewrite the item names in the json schemes.

alfbr (changes requested):
I have now tested the functionality and discussed it with my colleagues. We are very happy with the current implementation! Just a few changes are wanted from our end. I have commented in line where natural. Short version is the following:
-rather than option name, it is better to use item number (this is easier to understand for the user, and also will avoid the whole renaming of options in opm-common)
-it would be great if we can also report record number of an unsupported item/option. See for instance COMPDAT, where there are typically multiple records. A message can then be like this:
`COMPDAT: Invalid value '2' in record 5 for item 12`

@bska great if you can do a review on the code.

opm/simulators/utils/UnsupportedFlowKeywords.hpp
```diff
@@ -0,0 +1,3114 @@
...
+        {false, std::nullopt},
+    },
+    {
+        "NOCASC",
+        {false, std::nullopt},
+    },
+    {
+        "NODEPROP",
+        {false, std::nullopt},
+    },
```
bska, line 1659: I think we support `NODEPROP` now.

bska (commented):
I commented a bit on the overall structure earlier.  I agree with @alfbr that any diagnostic message should probably include both the record and the item numbers (counted from 1). but I am not convinced that we should make item IDs the primary lookup key.

@joakim-hove : If there were a way of inferring the item ID from the compiled item names that would really help here.

I've also identified a couple of keywords that I think we support but which are classified as unsupported in this PR.  Please check those more closely.

opm/simulators/utils/PartiallySupportedFlowKeywords.hpp
joakim-hove: I guess this code is interesting for three different groups of people:

1. Pure developers who implement and maintain the solution.
2. Intermediate developers/users who might maintain the list of supported/unsupported.
3. End users.

At least for categories 1 & 2 I think string identifiers is better than integers, but we can of course present the user with an integer number even if we use strings as internal ID.

joakim-hove:
> @joakim-hove : If there were a way of inferring the item ID from the compiled item names that would really help here.

Not as we speak - but that should be simple to add.

opm/simulators/utils/UnsupportedFlowKeywords.hpp
joakim-hove: Yes

opm/simulators/utils/UnsupportedFlowKeywords.hpp
```diff
@@ -0,0 +1,3114 @@
...
+    {
+        "RPTSMRY",
+        {false, std::nullopt},
+    },
+    {
+        "RPTSOL",
+        {false, std::nullopt},
+    },
```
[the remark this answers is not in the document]
joakim-hove: These output keywords have a myriad of options - but I would say: Yes we support RPTSOL.

opm/simulators/utils/UnsupportedFlowKeywords.hpp
```diff
@@ -0,0 +1,3114 @@
...
+    {
+        "IMPLICIT",
+        {false, std::nullopt},
+    },
+    {
+        "IMPORT",
+        {false, std::nullopt},
+    },
```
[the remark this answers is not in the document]
joakim-hove: Yes - with some children sickness: https://github.com/OPM/opm-common/pull/2416

joakim-hove:
General comment: I *really* think this code should utilize the compiled keyword classes generated by the opm-common build process; then we have strings++ only one place. Furthermore I think we should add short (optional!) doc strings to the Json scheme, so that this code can fetch documentation from there instead of adding it here. In the example below I have added `"doc": ` attributes to the `IMPORT`keyword - but the whole keyword and the items:
```json
{
  "name": "IMPORT",
  "doc": "The IMPORT keyword can be used to load keywords from files in output format",
   "sections": [
    "GRID",
    "EDIT",
    "PROPS",
    "REGIONS",
    "SOLUTION"
  ],
  "size": 1,
  "items": [
    {
      "name": "FILE",
      "value_type": "STRING",
      "doc" : "The name of the file we are loading information from" 
   },
    {
      "name": "FORMATTED",
      "value_type": "STRING",
      "default": "UNFORMATTED",
      "Formatted or unformatted file. Default: {{default}}"
    }
  ]
}
```

opm/simulators/utils/UnsupportedFlowKeywords.hpp
```diff
@@ -0,0 +1,3114 @@
...
+    {
+        "HALFTRAN",
+        {false, std::nullopt},
+    },
+    {
+        "HAxxxxxx",
+        {false, std::nullopt},
+    },
```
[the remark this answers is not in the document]
joakim-hove: This is a "container" keyword: https://github.com/OPM/opm-common/blob/master/src/opm/parser/eclipse/share/keywords/000_Eclipse100/H/HAxxxxxx - so the unsupported are really: `[HASOWCR, HASOGCR, HASWL, HASWLPC, HASGLPC]` - I have no clue what they do.

opm/simulators/flow/KeywordValidation.hpp
alfbr: Yes, but I hope we can keep it simple at this point in time. The current implementation (before this PR) had to handle each keyword item manually in the code, and only had a binary approach (one value supported, one value not supported). This is already miles further.

opm/simulators/utils/PartiallySupportedFlowKeywords.hpp
alfbr: I believe it is group 2 we want to cater to. And currently they can all check the manual for item numbers, or just look at the data deck at hand. String identifiers is another beast, especially since they are different in the Json schemes and the manual.

verveerpj:
I have now made a modification that prints record and item numbers. 

Some comments:
- The approach I took to get record/item number is very simplistic, I assumed iterating through the records and items in the keyword object would do the right thing. Please check and let me know that if that line of thinking is valid...
- For now I am still using the item name in the table of partially supported options, since @joakim-hove  suggested we look into using the generated keyword classes. That seems indeed elegant to me, but also more work, so we should discuss.
- @bska asked how we handle `int` options, if I understand well, to have some constraints on those rather than just a list of supported values.  Right now, the code for the `int` options just does the same as the string options (its a template), but there is no reason why we could not have more sophisticated code for integer options. I would require a bit more work, so this is also something to discuss, to see what we really need for integer options.
- I have not yet implemented any of the suggestions for keywords that are supported/not supported. I suggest to compile a list of those at some point to get that right.

alfbr:
> I have now made a modification that prints record and item numbers.

Thank you, that is very helpful for our users!

>   Switching to using the item number is trivial, so if desired, that can be done quickly.

That is my preference, item names are inconsistent anyway, while item numbers can be inferred simply by looking at a data deck.

>     * I have not yet implemented any of the suggestions for keywords that are supported/not supported. I suggest to compile a list of those at some point to get that right.

It is not for this pull request. I promise to update the list of unsupported keywords in a new PR once this is merged.

alfbr:
I have now tested the latest changes, this is good from my end. How about you @bska , any blocking issues remaining?

bska:
> any blocking issues remaining?

None that I can think of.  This is a good start.  I expect we'll refactor this facility and make it more general, especially for the partially supported keywords, as time passes but for now I'm happy to merge this into the master (and hopefully release) branch.  I'll just run a build check first.

bska:
jenkins build this please

bska:
There are a few new warnings related to [shadowing data members](https://ci.opm-project.org/job/opm-simulators-PR-builder/2051/gcc/new/fileName.761161080/).  Once you fix those I'll merge this.

verveerpj:
jenkins build this please

verveerpj:
> There are a few new warnings related to [shadowing data members](https://ci.opm-project.org/job/opm-simulators-PR-builder/2051/gcc/new/fileName.761161080/). Once you fix those I'll merge this.

done

bska:
jenkins build this please

bska:
Thanks for the updates and the feature itself.  I'll merge this into master now.

PR #3070, opened 2021-02-17
[test] Not reducing time step to be below 3 days due to iterations counting

GitPaean:


GitPaean:
benchmark please

GitPaean:
jenkins build this please

GitPaean:
nonlinear iteration counting based damping of the time step contributes in getting extremely small time steps, those shortening might not be very helpful.

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.001    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.98     |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.009    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1074

alfbr:
Unfortunately, it seems this broke the ability to run norne-4c-msw in the opm-tests.

GitPaean:
> Unfortunately, it seems this broke the ability to run norne-4c-msw in the opm-tests.

Thanks for the information. This is just a very rough test to see how it works.  Will take a look at the test case norne-4c-msw to get some insight.

bska:
> Unfortunately, it seems this broke the ability to run [norne-4c-msw](https://github.com/OPM/opm-tests/blob/6e2f83b900d1c13617852e58054085d5f6b07066/norne/NORNE_ATW2013_4C_MSW.DATA) in the opm-tests.

This implementation will also break `TUNING` if a case happens to specify a maximum timestep size smaller than 3 days.  Finally, minimum three day timesteps will typically be very wrong for `LAB` unit simulations for which the time unit is hours.

GitPaean:
> This implementation will also break TUNING if a case happens to specify a maximum timestep size smaller than 3 days. Finally, minimum three day timesteps will typically be very wrong for LAB unit simulations for which the time unit is hours.

Yes. More detailed logic should join or begin with more conservative.  We have a case, the time step keeps going all the way to e-22 level even most of the previous time steps are successful while needs more newton iterations (like 14, 17). At the end, each newton iteration needs one linear iteration for linear solution.  For that case, the reducing of the time step is not helpful or desirable.  Anyway, it is just a rough idea and rough test.

GitPaean:
> This implementation will also break TUNING if a case happens to specify a maximum timestep size smaller than 3 days. 
> Finally, minimum three day timesteps will typically be very wrong for LAB unit simulations for which the time unit is hours.

A second thought is that it should not be that dangerous, because the final time step used are 
```c++
        return std::min(dtEstimatePID, dtEstimateIter);
```
We just do not use iteration number based damping to keep shortening the time step indefinitely.  `dtEstimatePID` still helping to decide the time step. 

But anyway, the code should be more sophisticated.

GitPaean:
If it is the deck `NORNE_ATW2013_4C_MSW.DATA` in the opm-tests, it runs through on my local machine. But it is a prediction case, it shuts wells at slightly different location.  But it does run through. 
[norne_4c_msw.pdf](https://github.com/OPM/opm-simulators/files/5999102/norne_4c_msw.pdf)

(In the result, the first one is master branch serial running, the second to the fourth one are this PR with 1, 4, 8 processes respectively)

Is the benchmark running with different tolerance? @alfbr

alfbr:
> If it is the deck `NORNE_ATW2013_4C_MSW.DATA` in the opm-tests,

Yes, that is the case.

 > it runs through on my local machine. But it is a prediction case, it shuts wells at slightly different location. But it does run through.

That is strange. We have seen that the MSW cases can behave differently on different machines, but I cannot recall seeing the difference cause failure to run the case.

> Is the benchmark running with different tolerance? @alfbr

Yes, all defaults. Only mpirun has some parameters to lock processes to cores and cache to obtain accuracy of run time within 1%.

On a related note, I did increase the default --time-step-control-decay-damping-factor to avoid the excessive chopping you describe here. Initially it was more aggressive causing at least one test case to not run through due to repeated chops.

alfbr:
> If it is the deck `NORNE_ATW2013_4C_MSW.DATA` in the opm-tests, it runs through on my local machine. 

There was a bug preventing the case to be run, thanks for testing and reporting! Should be fixed now.

alfbr:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 0.999    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 0.981    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 1.003    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 1.006    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 1.107    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 0.923    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1090

GitPaean:
> opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 0.999
> -- | -- | --
> opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.981
> opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.003
> opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.006

This change should not change anything for the above running.

GitPaean:
> There was a bug preventing the case to be run, thanks for testing and reporting! Should be fixed now.

Good. The case can be run through now. Thanks for fixing it.

GitPaean:
> 

opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 1.107
-- | -- | --
opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 0.923


For this case, the measure should not only be the performance. The key is still the results are reasonable before we look at the performance. But I guess it is not easy to do.

alfbr:
> > 
> 
> opm-git 	OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 	1.107
> opm-git 	OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 	0.923
> 
> For this case, the measure should not only be the performance. The key is still the results are reasonable before we look at the performance. But I guess it is not easy to do.

Yes, I am afraid that is a highly non-trivial thing to do. I agree though, having a balanced metric for correctness would be extremely valuable.

totto82:
> But anyway, the code should be more sophisticated.

The problem this PR ties to fix is real. I suggest the following improvements
1) make the 3 days limit a run-time parameter
2) Fix the UNIT issue @bska reported. Use X * Time, where Time is either Days or hours depending on the unitsystem and x is a non-dimensional time. For instance 3 as I suggested here. Or alternatively just make the default unit-system specific. I.e. 3 h for Lab else 3 days. I suggest you do whatever you think is easiest. 

> This implementation will also break TUNING if a case happens to specify a maximum timestep size smaller than 3 days.
It should be fine I think since the the max time step limit is handled like this here:

https://github.com/OPM/opm-simulators/blob/a2a9ed192da7a361abf33379d25818b91457b4b7/opm/simulators/timestepping/AdaptiveSimulatorTimer.cpp#L74

The timestep control code computes a suggestion for the next time step as also @GitPaean pointed out.

GitPaean:
jenkins build this please

GitPaean:
Made the changes based on @totto82 's comments. By default, it should not change any running results.

GitPaean:
jenkins build this please

GitPaean:
It did not touch anything related to linear iteration target, because there is no testing regarding to that and it is not the default time stepper. But with this PR, it is trivial to add something similar for the linear iteration based time step control.

alfbr:
I am afraid this will need more testing, I was not aware that you wanted this merged. I believe the data sets that currently benefits most from taking iteration count into consideration typically has time steps below three days. One day may work though. Do you want me to test?

GitPaean:
For testing, let us wait the jenkins build finish and reviewing is done so that the PR is ready first.

GitPaean:
For testing, let us wait the jenkins build finish and reviewing is done so that the PR is ready first.

alfbr:
> And also, the current default for this parameter is 0, it should not change the running results of any case by default.

Ah, I see. Then I have no reservations against merging.

PR #3073, opened 2021-02-22
Remove timemap member from Schedule

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/2310


bska:
I'm merging this into master now to chase the API update.

PR #3082, opened 2021-02-26
make sure wells has valid controls

totto82:
I don't understand why and when `Well::ProducerCMode::NONE` started appearing as a control. This is a band-aid type solution that fixes an urgent issue, but I would very much like to understand how a well production control can be NONE. I think it is related to recent changes in how we handle open/shut wells.


totto82:
jenkins build this please

blattms:
Does this fix:
```
opm/simulators/wells/StandardWell_impl.hpp:1575] Well control must be specified for well ....
```

totto82:
Yes, that was the intention.

joakim-hove:
Can we have a backtrace of the exception?

PR #3083, opened 2021-02-26
Fix MPI issues in the well model

totto82:
The first commit makes some throws in `getWellResiduals` thread safe. 
The second make sure that the group related quantities are communicated before the wellTest

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -326,9 +326,9 @@ namespace Opm {
 
         well_state_ = previous_well_state_;
         well_state_.disableGliftOptimization();
+        updateAndCommunicateGroupData();
         const int reportStepIdx = ebosSimulator_.episodeIndex();
         const double simulationTime = ebosSimulator_.time();
-
         int exception_thrown = 0;
         try {
             // test wells
@@ -403,7 +403,6 @@ namespace Opm {
         const auto& comm = ebosSimulator_.vanguard().grid().comm();
         WellGroupHelpers::updateGuideRatesForWells(schedule(), phase_usage_, reportStepIdx, simulationTime, well_state_, comm, guideRate_.get());
         try {
-            updateAndCommunicateGroupData();
             // Compute initial well solution for new wells
             for (auto& well : well_container_) {
                 const uint64_t effective_events_mask = ScheduleEvents::WELL_STATUS_CHANGE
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

totto82:
benchmark please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -326,9 +326,9 @@ namespace Opm {
 
         well_state_ = previous_well_state_;
         well_state_.disableGliftOptimization();
+        updateAndCommunicateGroupData();
```
blattms, line 329: Could the movement of this function to here, explain the failing tests?

blattms (commented):
Apart from the failing parallel tests this seems good. Maybe we need to communicate group at the old place?

Once this tests are fixed this is fine to be merged.

opm/simulators/wells/BlackoilWellModel_impl.hpp
totto82: yes. but I think moving it here is correct. The old code used old group data in the welltest and well potential code etc.

totto82:
> failing parallel tests

I don't see any of the parallel test failing. A update data should do here if you agree to the moving of `updateAndCommunicateGroupData()`

bska:
> > failing parallel tests
> 
> I don't see any of the parallel test failing.

I agree with @totto82 .  The [failing tests](https://ci.opm-project.org/job/opm-simulators-PR-builder/1878/testReport/) are sequential regression tests only.  I suppose the `mpi.` in the test name (e.g., [mpi.compareECLFiles_flow+MOD4_GRP](https://ci.opm-project.org/job/opm-simulators-PR-builder/1878/testReport/(root)/mpi/compareECLFiles_flow_MOD4_GRP/)) "just" means that the tests are being run with an executable targeting an MPI configuration.  The simulation run however is effectively launched as
```sh
mpirun -np 1
```
Details of the process are coded in [compareECLFiles.cmake](https://github.com/OPM/opm-simulators/blob/831f1db872150fb218fa205f6827c14e261c6e6b/compareECLFiles.cmake) and [tests/run-regressionTest.sh](https://github.com/OPM/opm-simulators/blob/831f1db872150fb218fa205f6827c14e261c6e6b/tests/run-regressionTest.sh).

blattms:
Indeed mpi indicates only compilation with mpi support. Sorry for the noise.

totto82:
I have manually looked at the test failures. It should be safe to merge this with updated data.

totto82:

bska:
jenkins build this opm-tests=491 please

bska:
New reference solutions have been installed on the CI system.  I'll merge this now.  Thanks a lot for identifying this issue.

PR #3084, opened 2021-02-26
only log changes of group controls on rank 0

totto82:
No need to spam the log with multiple messages with the same content.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -2333,10 +2333,7 @@ namespace Opm {
         }
 
         auto cc = Dune::MPIHelper::getCollectiveCommunication();
-        if (cc.size() > 1) {
-            ss << " on rank " << cc.rank();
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

bska:
Do control changes get communicated back to the I/O rank?  In other words, do we guarantee that if process 123 changes the control model of PROD-X from ORAT to GRUP then process 0 knows about that change?  If not, I think this PR is actually wrong.

totto82:
> Do control changes get communicated back to the I/O rank?

No, but all groups lives on all ranks and the group rates get communicated. I. E they should be violated at the same time.

bska:
> > Do control changes get communicated back to the I/O rank?
> 
> No, but all groups lives on all ranks and the group rates get communicated. I. E they should be violated at the same time.

Ah, yes&mdash;you're right.  I missed the key phrase "group control changes" and thought this was about wells too.  In that case we're safe and I'll merge this into master.  Sorry about the confusion.

PR #3085, opened 2021-02-26
Use fmt::format() for two messages

joakim-hove:
Mainly a dummy to provoke a `update_data`for: https://github.com/OPM/opm-tests/pull/487

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1376,7 +1376,7 @@ public:
                         wells_string += matching_wells[iw] + ", ";
                     wells_string += matching_wells.back();
                 }
-                std::string msg = "The action: " + action->name() + " evaluated to true at " + ts + " wells: " + wells_string;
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this opm-tests=487 please

joakim-hove:
jenkins build this opm-tests=487 update_data please

joakim-hove:
jenkins build this please

joakim-hove:
This was mainly introduced as a dummy for a update_data; but it is a small improvement anyway - merging.

blattms:
This will break modules using opm-simuiators as your are using fmtlib in a header!
You will need to put the logging into its own function and that needs to go into a cpp file

joakim-hove:
> This will break modules using opm-simuiators as your are using fmtlib in a header!

Ahhh ;-( Thanks for pointing out.

> You will need to put the logging into its own function and that needs to go into a cpp file

I'll just revert it.

Issue #3081, opened 2021-02-24
parallel results show obvious difference with using PORV to modify the pore volume directly


GitPaean:
It is part of the efforts to investigate the reason why the parallel results are different in parallel with PR https://github.com/OPM/opm-simulators/pull/3039 .  (https://github.com/OPM/opm-common/issues/2282 is also due to the same reason)

The test case used is attached here. 
[test_porv_parallel.tar.gz](https://github.com/OPM/opm-simulators/files/6039173/test_porv_parallel.tar.gz)

Without the following keywords 
```
148 EQUALS
149  'PORV' 0.25000000E+10   20 20  1 1  10 10 /
150  'PORV' 0.26400003E+10   20 20  2 2  10 10 /
151 /
```

The serial results and the results with 4 mpi processes are the same. 
[ftaq-0104.pdf](https://github.com/OPM/opm-simulators/files/6039196/ftaq-0104.pdf)

But with the above setup, the serial results and parallel results are different. 
[ftaq-0104.pdf](https://github.com/OPM/opm-simulators/files/6039189/ftaq-0104.pdf)

The running parameter is `--enable-tuning=true`.

blattms:
Actually if you access PORV via the `prov()` method in the (Parallel)FieldPropsManager, the values should be broadcasted from rank 0.  As along is the EQUALS operation is applied directly in in the parser (@joakim-hove should know) the values should be correct.

blattms:
Are we maybe summing pore values of a few cells without neglecting the ghosts and by coincidence the porv of the ghosts is zero without EQUALS?

joakim-hove:
Will investigate ...

joakim-hove:
Have added a testcase for this: https://github.com/OPM/opm-tests/pull/489 - will continue the investigations

blattms:
I did take another look at [`ParallelFieldPropsManger::porv()`](https://github.com/OPM/opm-simulators/blob/831f1db872150fb218fa205f6827c14e261c6e6b/opm/simulators/utils/ParallelEclipseState.cpp#L50-L60) and have to conclude that it will only work if called with argument `true` (to get a global field properties vector. Otherwise the resulting array should be bigger than the number of cells. It will be an array with the condensed field properties of the undistributed grid. That is not usable at all.

if called with `false` or no argument, we should still use `true` in the underlying call on the root process and use the CartesianIndexMapper to map the broadcasted result to a local array. Not very pretty but should work.

[...]

PR #3087, opened 2021-02-27
Use only active elements when updating reference porosity

joakim-hove:
Small drive-by improvements while working on: https://github.com/OPM/opm-simulators/issues/3081

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -2669,11 +2669,10 @@ private:
         referencePorosity_[/*timeIdx=*/0].resize(numDof);
 
         const auto& fp = eclState.fieldProps();
-        const std::vector<double> porvData = fp.porv(true);
+        const std::vector<double> porvData = fp.porv(false);
         const std::vector<int> actnumData = fp.actnum();
         for (size_t dofIdx = 0; dofIdx < numDof; ++ dofIdx) {
-            unsigned cartElemIdx = vanguard.cartesianIndex(dofIdx);
-            Scalar poreVolume = porvData[cartElemIdx];
+            Scalar poreVolume = porvData[dofIdx];
 
             // we define the porosity as the accumulated pore volume divided by the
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -2669,11 +2669,10 @@ private:
         referencePorosity_[/*timeIdx=*/0].resize(numDof);
 
         const auto& fp = eclState.fieldProps();
-        const std::vector<double> porvData = fp.porv(true);
+        const std::vector<double> porvData = fp.porv(false);
         const std::vector<int> actnumData = fp.actnum();
         for (size_t dofIdx = 0; dofIdx < numDof; ++ dofIdx) {
-            unsigned cartElemIdx = vanguard.cartesianIndex(dofIdx);
-            Scalar poreVolume = porvData[cartElemIdx];
+            Scalar poreVolume = porvData[dofIdx];
```
bska, line 2675: What does `FieldProps::porv(false)` return in a parallel simulation?  Does it return only those values that are local to the process or does it return a vector of values that is sized according to the number of active cells in the entire model?

blattms: @joakim-hove This will break flow if called on a distributed grid. As I tried to point out in the bug report #3081 fp.porv(false) returns wrong results. Only the approach taken here (before your change) yields correct results. (Actually I tried to explain that what was done here originally needs to be done there for the argument false).

Your change (without fixing ParallelEclipseState::porv) will have the following effect.
- Each process will have a porv array for the active cells of the undistributed grid.

Are you sure that this is what we want?

joakim-hove: Ehhh - I will close this now and try to digest what you write. Thanks.

joakim-hove:
OK - I have added an additional commit to this - and reopened it as #3088.

PR #3088, opened 2021-02-28
Use active index

joakim-hove:
New version of #3087


joakim-hove:
@bska / @blattms: OK the initial version of #3087 was plain wrong - I never really looked into/understood [this comment](https://github.com/OPM/opm-simulators/issues/3081#issuecomment-786717376) - but now I have added an additional commit and hope/think it should be correct.

joakim-hove:
jenkins build this please

blattms (approved):
Exactly what I meant.

PR #3093, opened 2021-03-02
Record dynamic wellstatus

bska:
This PR sets the `data::Well::dynamicStatus` based on the dynamically updated 'Schedule' object (i.e., from ACTIONX and similar) and the results of well/operability testing (`WECON` and/or `WTEST`).  If a well is closed due to economic limits (`WECON`) we still provide summary-style data at the timestep that closed the well, but omit this data at later steps until the well reopens.

We add a new parameter to `WellState::report()` to distinguish these situations.

This is in preparation of making the `BlackoilWellModel` manage both open and shut wells alike

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -217,7 +217,12 @@ namespace Opm {
 
             Opm::data::Wells wellData() const
             {
-                auto wsrpt = well_state_.report(phase_usage_, Opm::UgGridHelpers::globalCell(grid()));
+                auto wsrpt = well_state_
+                    .report(phase_usage_, Opm::UgGridHelpers::globalCell(grid()),
+                            [this](const int well_ndex) -> bool
+                    {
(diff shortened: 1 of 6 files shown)


bska:
This PR depends on upstream PR OPM/opm-common#2315 and must be merged int concert with that.

totto82:
Thanks both for the cleanup and the feature. The simulator part looks good IMO and can me merged. How will this work for restart? Is additional work needed there to support reading of dynamic well status?

bska:
> How will this work for restart? Is additional work needed there to support reading of dynamic well status?

It's been a while since I checked the details, but I believe we read the status field out of the restart file and then confer that value onto the pertinent `Opm::Well` instance managed by the `Schedule` object.  I don't think we do anything special to the simulator's well state object however, so we may need to do additional work there.

totto82:
Ok. I noticed the opm-common part was already approved. I can merge them.

bska:
jenkins build this opm-common=2315 please

PR #3094, opened 2021-03-02
Compute Well-Level PI For Shut Wells 

bska:
This commit includes the shut wells in `wells_ecl_` and expands the PI/II value calculation to apply to those shut wells too.  With this in place we are able to run cases that have a `WELPI` keyword before the well opens, even at the very first report step.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -65,9 +65,6 @@ namespace Opm {
 
         const auto numProcs = ebosSimulator.gridView().comm().size();
         this->not_on_process_ = [this, numProcs](const Well& well) {
-            if (well.getStatus() == Well::Status::SHUT)
-                return true;
-
             if (numProcs == decltype(numProcs){1})
                 return false;
 
@@ -923,6 +920,30 @@ namespace Opm {
 
 
     template <typename TypeTag>
+    void BlackoilWellModel<TypeTag>::
+    inferLocalShutWells()
+    {
+        this->local_shut_wells_.clear();
+
+        const auto nw = this->numLocalWells();
+
+        auto used = std::vector<bool>(nw, false);
+        for (const auto& wellPtr : this->well_container_) {
+            used[wellPtr->indexOfWell()] = true;
+        }
+
+        for (auto wellID = 0; wellID < nw; ++wellID) {
+            if (! used[wellID]) {
+                this->local_shut_wells_.push_back(wellID);
+            }
+        }
+    }
+
+
+
+
+
+    template <typename TypeTag>
     typename BlackoilWellModel<TypeTag>::WellInterfacePtr
     BlackoilWellModel<TypeTag>::
     createWellPointer(const int wellID, const int time_step) const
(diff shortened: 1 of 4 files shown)


bska:
This PR depends on #3045 and must not be merged before that is installed in the master branch.

tskille:
This has been tested on a major ffm in Equinor and simulation results are in very good agreement with result from the commercial simulator.

tskille:
@bska, I would very much like to test this together with fix in [#3095](https://github.com/OPM/opm-simulators/pull/3095) (which is already merged.) on one of our major full field models. It seems like this PR now need a re-base. Will you be able to do this for us ?

bska:
> It seems like this PR now need a re-base [in order to include #3095]. Will you be able to do this for us ?

I have done that locally and I'll push it once I've checked that I haven't broken anything obvious.

bska:
> > It seems like this PR now need a re-base [in order to include #3095]. Will you be able to do this for us ?
> 
> I have done that locally and I'll push it once I've checked that I haven't broken anything obvious.

@tskille: I have now pushed the rebased version of this PR so that it includes #3095.  Feel free to test it at your convenience.

bska:
jenkins build this opm-common=2315 please

bska:
jenkins build this opm-common=2315 please

bska:
I'm marking this as "ready for review" now.  In local testing there are three failing regression tests.  The differences are small and I believe it's safe to update the reference solutions for those but I'll leave that call to the reviewer.

bska:
jenkins build this please

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
```diff
@@ -1117,8 +1130,18 @@ namespace Opm
...
             WellState::shutWell(well_index);
             const int np = numPhases();
-            for (int p = 0; p < np; ++p)
-                this->well_reservoir_rates_[np * well_index + p] = 0;
+
+            auto* resv = &this->well_reservoir_rates_[np*well_index + 0];
+            auto* wpi  = &this->productivity_index_[np*well_index + 0];
+
+            for (int p = 0; p < np; ++p) {
+                resv[p] = wpi[p] = 0.0;
```
atgeirr, line 1138: Easy to miss that you are assigning two arrays, please use one line for each.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2792,17 +2862,21 @@ namespace Opm {
             auto saved_previous_well_state = this->previous_well_state_;
             this->previous_well_state_ = this->well_state_;
 
-            well_container_ = createWellContainer(timeStepIdx);
-            for (auto& well : well_container_) {
-                well->init(&phase_usage_, depth_, gravity_, local_num_cells_, B_avg_);
-            }
+            this->well_container_ = this->createWellContainer(timeStepIdx);
```
atgeirr, line 2865: Is it necessary to recreate the whole container here?

atgeirr (commented):
The code looks fine. I have a question below about the well container, and a minor nitpick, that is it for the code as such.

I am worried over any changes to cases that do not involve WELLPI: this should leave all bitwise the same. Please investigate this!

opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
bska: I've pushed an update to do this.

opm/simulators/wells/BlackoilWellModel_impl.hpp
bska: As of right now, it's not so much "recreate" as it is "create".  We (currently) run the enclosing function ([`runWellPIScaling`](https://github.com/OPM/opm-simulators/blob/0ff9e0c4bb861b93da73aae0053ebb4c34e9c147/opm/simulators/wells/BlackoilWellModel_impl.hpp#L2830)) as part of [`beginReportStep()`](https://github.com/OPM/opm-simulators/blob/0ff9e0c4bb861b93da73aae0053ebb4c34e9c147/opm/simulators/wells/BlackoilWellModel_impl.hpp#L296) at which point the `well_container_` does not yet have a well-defined value.

I didn't try too hard to call this function from `beginTimeStep()`&mdash;where `well_container_` **does** exist&mdash;because we need to guarantee that we don't scale connection transmissibility factors multiple times for the **same** `WELPI` keyword.

bska:
jenkins build this please

bska:
jenkins build this please

totto82:
I think this can be merged now. I will Jenkins it.

totto82:
jenkins build this please

bska:
jenkins build this please

totto82:
jenkins build this please

totto82:
@atgeirr @bska Is this ready noew? I will soon create release branches I would like to include this if it is ready.

atgeirr:
> @atgeirr @bska Is this ready noew? I will soon create release branches I would like to include this if it is ready.

From my POV yes, unless @bska added something new here?

bska:
> Is this ready now?

It Depends[tm].

If the criterion for inclusion in the release is "no regression failures", then no&mdash;it is not ready.  If the criterion is "does the right thing on important field cases albeit with slight slight regression failures in cases that use UDQ and/or ACTIONX", then yes it is ready for the release (once we update the reference solutions).

totto82:
I will create the release branch without this then, but keep the release tag. If it gets ready before the final release we include it.

totto82:
Any news on this.

bska:
> Any news on this.

No change since last week.  Unlikely to make it into the release so I've removed the 2021.04 milestone tag.

bska:
The requisite master update #3182 has been applied.  I've checked that the regression tests all pass locally.  I'll run a new official build check of this PR now.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
This PR finally passes all regression tests and I submit it for final review.  I'd also like to recall an earlier [comment](https://github.com/OPM/opm-simulators/pull/3094#issuecomment-789586164) by @tskille:

> This has been tested on a major ffm in Equinor and simulation results are in very good agreement with result from the commercial simulator.

I'd like to get this merged into master now and, ideally, also into the 2021.04 release. @totto82: Would that be possible or is too late by now?

totto82:
Thanks for your effort. We are in the final testing of the release I don't think this PR in itself is sufficient for another round. But if more bugs etc .are found that mandates an update I can include this as well.

PR #3095, opened 2021-03-02
Apply gefac also on topmost VREP controlled group

blattms:
We have a model that has a tree of groups (three levels of groups below the topmost group F(topmost)->G1->G2->G3, wells only below G3). When we have a VREP constraint target on G1 and a group efficiency factor g1<1 results are wrong (We actually sometimes inject more than we should and there is a lot of switching in the group control modes). If on the other hand g1 is applied to G2 and G1 has group efficiency factor 1, the results are (nearly) correct.

~This is my rather naive attempt to always apply g1 to the topmost group G1. It does not really work and even worse injection drops to zero at some stage and has problems coming up again.~

~1ce5c598 is how I understood @atgeirr's private comment. (I also tested it standalone and it lead to injection dropping way below production and even production becoming 0 at some time).~

~I need to go back to my desk and rethink what might be happening here (maybe at some occasions applying the group efficiency factor of the VREP controlled group is actually wrong). But if anybody has some hints then they are highly welcome.~

I missed on spot. Fixed in d1f6545 and now reproducing results.

### DIFF
--- a/opm/simulators/wells/WellGroupHelpers.cpp
+++ b/opm/simulators/wells/WellGroupHelpers.cpp
@@ -194,8 +194,7 @@ namespace WellGroupHelpers
         double rate = 0.0;
         for (const std::string& groupName : group.groups()) {
             const Group& groupTmp = schedule.getGroup(groupName, reportStepIdx);
-            rate += groupTmp.getGroupEfficiencyFactor()
-                * sumSolventRates(groupTmp, schedule, wellState, reportStepIdx, injector);
+            rate += sumSolventRates(groupTmp, schedule, wellState, reportStepIdx, injector);
         }
         const auto& end = wellState.wellMap().end();
         for (const std::string& wellName : group.wells()) {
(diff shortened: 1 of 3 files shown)


blattms:
there is a German(?) saying: "you should praise the day before it's end", but the current run looks really promising... First bump is already gone. Let' wait patiently (faster computer faster)

blattms:
jenkins build this please

blattms:
I should have a party tonight! Even jenkins agrees with my code! :love_letter:

joakim-hove:
Great news Markus!

blattms:
Thanks. And it was about time (nearly went nuts).

totto82:
Thanks. The code look good as well. Lets merge it!

> Even jenkins agrees with my code!

That means that the bug you fixed is not exposed in the tests. We should adapt one of the test case to make sure the topmost level is included in the GEFAC test coverage.

blattms:
It also means that I didn't break anything else (which was from obvious to me).

What test covers GEFAC? I might try to add one that fails before this PR (but might need help).

PR #3096, opened 2021-03-03
add support for POTN in WECON

totto82:
This adds support for POTN (default is RATE) in WECON.


totto82:
jenkins build this please

GitPaean:
The code looks fine. Have you have a small test case to demonstrate the running. You should add it as a regression test for demonstration purpose and also protect this feature.

totto82:
I have tested this locally based on a test from @tskille. This or something similar will be added to the test suite soon. I will merge this now at add the test when it is ready. Thanks for your review.

PR #3097, opened 2021-03-03
Implements gas lift optimization for groups.

hakonhagland:
Extends PR #2824 to include support for `GLIFTOPT` (item 2, maximum lift gas supply for a group) and group production constraints.

The optimization is split into two phases. First the wells are optimized separately (as in PR #2824). In this phase `LIFTOPT` and `WLIFTOPT` constraints (e.g. maxmimum lift gas injection for a well, minimum economic gradient) are
considered together with well production constraints.

Then, in the next phase the wells are optimized in groups. Here, the ALQ distribution from the first phase is used as a starting point. If a group has any production rate constraints, and/or a limit on its total rate of lift gas supply, lift gas is redistributed to the wells that gain the most benefit from it by considering which wells that currently has the largest weighted incremental gradient (i.e. increase in oil rate compared to increase in ALQ).


hakonhagland:
jenkins build this please

bska:
I haven't looked at this at all, but at minimum we need to do something about the 100 [new warnings](https://ci.opm-project.org/job/opm-simulators-PR-builder/1895/gcc/new/) being introduced in the current iteration of the PR.

totto82:
It seems like most new warning are relatert unused variable in assert statements. What is the best way to avoid these?

bska:
> unused variable in `assert` statements. What is the best way to avoid these?

```C++
[[maybe_unused]]
```

hakonhagland:
jenkins build this please

hakonhagland:
@bska Thanks for the comment, I have pushed a new commit. But I do not yet understand why [this](https://ci.opm-project.org/job/opm-simulators-PR-builder/1895/gcc/new/folder.1225362760/fileName.-1134880832/source.0dcb501b-e641-43f3-9808-9557281b1787/#44) line gives a warning. Any ideas?

joakim-hove:
> It seems like most new warning are relatert unused variable in assert statements. What is the best way to avoid these?

```C++

if (!condition)
    throw std::something("What the f...");
```

joakim-hove:
I have not looked at this - and it is not in my part of town so I probably will not either. But:

1. This is quite a lot of entirely new code - without knowing details I would assume that it would be relatively simpler to create tests for this than what is typically the case in opm-simulators.
2. This is a large Pull request - I would assume it should be possible to submit it in pieces. That simplifies the review process considerably.

opm/simulators/wells/GasLiftStage2_impl.hpp
```diff
@@ -0,0 +1,1027 @@
...
+    double gas_rate = 0.0;
+    double alq = 0.0;
+    // NOTE: A group can either contain wells or groups, but not both
+    if (group.wellgroup()) {
+        for (const std::string& well_name : group.wells()) {
+            auto [sw_oil_rate, sw_gas_rate, sw_alq] =
+                getCurrentWellRates_(well_name, group.name());
+            oil_rate += sw_oil_rate;
+            gas_rate += sw_gas_rate;
+            alq += sw_alq;
```
blattms, line 306: In the spirit of #3095, please double check the group efficiency is also applied for the top-most group.
If this is based on code before the merge, chances are that it is not

alfbr:
Thank you @hakonhagland  this is highly appreciated! I will start testing this now.

alfbr:
I am not able to run the test model https://github.com/OPM/opm-tests/blob/master/model5/3_GLIFT_MODEL5.DATA with this PR in. Am I missing something?

opm/simulators/wells/GasLiftSingleWell.hpp
```diff
@@ -0,0 +1,235 @@
...
+            DeferredLogger &deferred_logger,
+            WellState &well_state
+        );
+        struct GradInfo;
+        std::optional<GradInfo> calcIncOrDecGradient(
+            double oil_rate, double gas_rate, double alq, bool increase) const;
+        std::pair<double, double> getStage2Rates();
+        const WellInterface<TypeTag> &getStdWell() const { return std_well_; }
+        bool hasStage2Rates();
+        std::unique_ptr<GLiftWellState> runOptimize();
```
joakim-hove, line 88: My experience from a period of heavy `std::shared_ptr<T>` abuse is that the public api should only expose `GliftWellState`, `const GliftWellState&` and possibly `const GLiftWellState *` - by using `std::xxx_ptr<T>` in the API you impose restrictions on downstream which is really not necessary.

hakonhagland:
> I am not able to run the test model https://github.com/OPM/opm-tests/blob/master/model5/3_GLIFT_MODEL5.DATA with this PR in.

@alfbr Yes you are right it throws an exception while reading the input file.  I am not sure what the problem is yet, but I am able to run the case if I comment of the region keywords in [summary.inc](https://github.com/OPM/opm-tests/blob/master/model5/include/summary.inc). The case I tested the code with only had region keyword `RPR`, `ROIP`, and `RGIP`..

joakim-hove:
I'll tale a look

joakim-hove:
See: https://github.com/OPM/opm-common/pull/2322

alfbr:
Actually, the deck was wrong, it has never been run in any simulator it seems. Fix is here: https://github.com/OPM/opm-tests/pull/495

hakonhagland:
Sorry, I forgot to mention that I set `NUPCOL` to 4 in my tests, using the default value (12) of `NUPCOL` gives me convergence problems (which also seems to be the case for the test model [3_GLIFT_MODEL5.DATA](https://github.com/OPM/opm-tests/blob/master/model5/3_GLIFT_MODEL5.DATA)).  The value of `NUPCOL` only has an effect if item 4 of `LIFTOPT` is set to `YES`.

hakonhagland:
> I forgot to mention that I set `NUPCOL` to 4 in my tests

That is when running flow with `--enable-tuning=true`. When running with `--enable-tuning=false` it converges even if `NUPCOL` is 12

totto82:
> This is a large Pull request - I would assume it should be possible to submit it in pieces. That simplifies the review process considerably.

I have followed this work closely for some time. I would not recommend to spend time on submitting this in pieces. The code is complex as the different "stages" interfere closely. The part of LIFTOPT that looks at individual wells is already in place in master. i.e. it has already been submitted in pieces.

joakim-hove:
> I would not recommend to spend time on submitting this in pieces.

I am not arguing that it should be split now, but rather I guess that it would have been possible to submit pieces while it was under construction. That would have increased the review quality.

> The code is complex as the different "stages" interfere closely

Well - maybe publishing it in pieces would have reduced the coupling? With test driven development one of the main advantages is that the approach helps you create code with looser couplings - something which can be argued is only a side effect of the test first principle.

totto82 (commented):
In general I think the code looks ok. It is well commented and the structure is good. I hope we can get this merged soon. I will assume that it will need some iterations before all the functionality behave as expected, but I suggest to merge this more or less as is and improve later. To avoid an even larger PR. 

The code is not parallel as is I think. Again I suggest you fix this in another PR, but this PR should throw with a reasonable message to the user that MPI is not supported yet. 

> This is quite a lot of entirely new code - without knowing details I would assume that it would be relatively simpler to create tests for this than what is typically the case in opm-simulators

It don't think it would be relatively simple but lets give it a try. The test_equil.cc is a good start. A good test would also make it easier to fix or extend the functionality further.

alfbr:
I can confirm that the GLIFTOPT keyword works as advertised. From my end this is good enough to go in. Some issues have been discovered during testing, but those can be addressed later.

hakonhagland:
jenkins build this please

hakonhagland:
jenkins build this please

hakonhagland:
jenkins build this please

hakonhagland:
> The code is not parallel as is I think. Again I suggest you fix this in another PR, but this PR should throw with a reasonable message to the user that MPI is not supported yet.

@totto82 Added a check on `comm().size()` in the last commit. It gives a warning if the size is larger than one, and then skips gas lift optimization.

hakonhagland:
> It don't think it would be relatively simple but lets give it a try. The test_equil.cc is a good start. A good test would also make it easier to fix or extend the functionality further.

@totto82 I Agree. I have written a test case, it required some changes to the API in `BlackoilWellModel.hpp`. The test also reads a deck (which again reads some include files), so I decided to submit it as a separate PR that depends on the current one, see #3135.  By submitting it as a separate PR, it should simplify the review process.

totto82:
benchmark please

hakonhagland:
jenkins build this please

totto82:
I plan to merge this when benchmark finish to make sure it does not hit any performance for cases without gaslift.

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1.033    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 1.006    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 1.045    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 0.995    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 1.32     |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 0.909    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1129

totto82:
The benchmark results looks strange. I will rerun it to be sure.

totto82:
benchmark please

totto82:
Seems like the benchmark for not triggered. I will try again.

totto82:
benchmark please

ytelses:
Benchmark result overview: 




View result details @ https://www.ytelses.com/opm/?page=result&id=1132

ytelses:
Benchmark result overview: 




View result details @ https://www.ytelses.com/opm/?page=result&id=1134

ytelses:
Benchmark result overview: 




View result details @ https://www.ytelses.com/opm/?page=result&id=1134

alfbr:
Looks like we may need a rebase before running a new benchmark?

hakonhagland:
jenkins build this please

totto82:
benchmark please

alfbr:
Seems it did not pick it up, let's give it one more go.

alfbr:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1        |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 0.997    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 0.988    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.995    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 1.002    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1134

totto82:
Finally both Jenkins and Benchmark are happy. In it goes. Thanks @hakonhagland for the contribution.

PR #3106, opened 2021-03-09
quell uninitialized usage warning without mpi

akva2:


akva2:

blattms:
I am too dense to see where the warning came from. `data` seems to have been used...

akva2:
it's only used with MPI. without mpi it's not.

blattms:
Ahh. thanks. That's quite tricky.

blattms (approved):

PR #3109, opened 2021-03-11
fix ordering of first line in *INFOSTEP files

fgfuchs:
fix ordering of first line in *INFOSTEP files, more specifically switches LSolve and LSetup mark


fgfuchs:
jenkins build this please

PR #3058, opened 2021-02-11
Allow usage of system fmtlib and always use imported library.

blattms:
Needs OPM/opm-common#2291

PR #3110, opened 2021-03-11
Enable usage of fmtlib on the system.

blattms:
If fmtlib is present on the system we used that one in the normal mode (not header only). Otherwise we fallback to the embedded one header only. Searching for the library is done on opm-common.

Replacement of #3058. Needs OPM/opm-common#2348


blattms:
BTW jenkins run from opm-common was successful.

blattms:
@akva2 What do you think of this?

akva2:
seems sane.

Issue #2350, opened 2020-02-18
MULTPV not working when specified in the EDIT section


wouterjdb:
In a situation where `PORV` is set in the `EDIT` section and one wants to apply an additional multiplier in the form of `MULTPV`, Flow seems to ignore `MULTPV`. Running the same deck in a commercial simulator leads to the desired result (`PORV*MULTPV`) - which is in line with what the documentation says:

> If `MULTPV` is entered in the `EDIT` section, it acts as an additional multiplier on the pore volumes after any edited pore volumes have been read from the `EDIT` section.

joakim-hove:
There has been much back and forth on this - do you know exactly what is the version (when compiled) of flow you have used?

wouterjdb:
I'm using flowdaily internal....  `flow 2020.04-pre`

alfbr:
In others words, it is flow freshly compiled from master every day. We can also easily test pull-requests in the daily binary.

PR #3111, opened 2021-03-13
Upgrade pybind11

joakim-hove:
Fix: https://github.com/OPM/opm-common/issues/2350


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

blattms:
We need to be very careful with this and test it on all supported systems. Especially the different supported CMake versions might be troublesome.

There is a (IMHO rather high) maintenance cost for embedded sources.

The clean way of solving #2350 would be to support the pybind11 on the system and prefer that.

joakim-hove:
> There is a (IMHO rather high) maintenance cost for embedded sources.

I agree

> The clean way of solving #2350 would be to support the pybind11 on the system and prefer that.

I have no strong feeling on this - whatever works.

blattms:
Stale and superseded by #5644

PR #3112, opened 2021-03-15
Add support for explicit guiderate for injection groups

totto82:
With this commit the guiderate logic used for the production groups is also used for injectors
This allows for setting guiderates explicit at different group levels
Only RATE, NETV and VOID guiderate type is supported.

Needs to be merged together with OPM/opm-common#2355

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1319,6 +1319,7 @@ namespace Opm {
             well_state_nupcol_ = well_state_;
         }
 
+        const auto& summaryState = ebosSimulator_.vanguard().summaryState();
         // the group target reduction rates needs to be update since wells may have swicthed to/from GRUP control
         // Currently the group target reduction does not honor NUPCOL. TODO: is that true?
         std::vector<double> groupTargetReduction(numPhases(), 0.0);
@@ -1326,13 +1327,13 @@ namespace Opm {
         std::vector<double> groupTargetReductionInj(numPhases(), 0.0);
         WellGroupHelpers::updateGroupTargetReduction(fieldGroup, schedule(), reportStepIdx, /*isInjector*/ true, phase_usage_, *guideRate_, well_state_nupcol_, well_state_, groupTargetReductionInj);
 
+        // the guiderate update should not be part of this function
+        // remove in seperate PR since it affects existing functionality
         const double simulationTime = ebosSimulator_.time();
         std::vector<double> pot(numPhases(), 0.0);
-        WellGroupHelpers::updateGuideRateForGroups(fieldGroup, schedule(), phase_usage_, reportStepIdx, simulationTime, /*isInjector*/ false, well_state_, comm, guideRate_.get(), pot);
-        std::vector<double> potInj(numPhases(), 0.0);
-        WellGroupHelpers::updateGuideRateForGroups(fieldGroup, schedule(), phase_usage_, reportStepIdx, simulationTime, /*isInjector*/ true, well_state_, comm, guideRate_.get(), potInj);
+        WellGroupHelpers::updateGuideRateForProductionGroups(fieldGroup, schedule(), phase_usage_, reportStepIdx, simulationTime, well_state_, comm, guideRate_.get(), pot);
+        WellGroupHelpers::updateGuideRatesForInjectionGroups(fieldGroup, schedule(), summaryState, phase_usage_, reportStepIdx, well_state_, guideRate_.get());
 
-        const auto& summaryState = ebosSimulator_.vanguard().summaryState();
         WellGroupHelpers::updateREINForGroups(fieldGroup, schedule(), reportStepIdx, phase_usage_, summaryState, well_state_nupcol_, well_state_);
         WellGroupHelpers::updateVREPForGroups(fieldGroup, schedule(), reportStepIdx, well_state_nupcol_, well_state_);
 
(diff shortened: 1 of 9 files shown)


blattms:
Seems like I still don't get the efficiency factors. Somehow I would have expected that the group efficiency would influence the updating of the guide rates, too.

totto82:
> Seems like I still don't get the efficiency factors. Somehow I would have expected that the group efficiency would influence the updating of the guide rates, too.

Please tell me where  you miss the group efficiency factor. The code should follow the logic of the production group, which I assumed was correct.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2866,11 +2867,17 @@ namespace Opm {
             for (auto gi = 0*nu; gi < nu; ++gi) {
                 const auto& gname = up[start + gi];
                 const auto& group = sched.getGroup(gname, reportStepIdx);
+                const auto  grup_type = group.getGroupType();
```
bska, line 2870: This variable is not referenced later.  Is that intentional?  If so, we should just remove it.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -227,6 +227,60 @@ namespace WellGroupHelpers
...
+            case Group::GuideRateInjTarget::RESV:
+                OPM_THROW(std::runtime_error, "GUIDE PHASE RESV not implemented. Group " + group.name());
+            case Group::GuideRateInjTarget::POTN:
+                break;
+            case Group::GuideRateInjTarget::NO_GUIDE_RATE:
+                break;
+            default:
+                assert(false);
+            }
+            guideRate->injectionGroupCompute(group.name(), phase, reportStepIdx, guideRateValue);
```
blattms, line 280: The group efficiency for the production group is applied at [WellGroupHelpers.hpp#L156-L162](https://github.com/OPM/opm-simulators/blob/205c4839192d2aa5ced7d6cb88ccede78f246dcd/opm/simulators/wells/WellGroupHelpers.hpp#L156-L162) when we compute the guide rates for production. Shouldn't this also be done before this line. 
But I might be missing something.

totto82: From my understanding it should already be accounted for here https://github.com/OPM/opm-simulators/blob/c29b61ff999969c0964cd6e39131661c7c3ea123/opm/simulators/wells/WellGroupHelpers.cpp#L162

totto82:
Note that NETV guiderate type will not work in parallel yet. I temporary reverted the fix for this to avoid changes that effects results. PR #3115 or something similar need to be merged to make NETV work in parallel.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2949,12 +2955,34 @@ namespace Opm {
...
         return grval;
     }
+    template <typename TypeTag>
+    data::GuideRateValue
+    BlackoilWellModel<TypeTag>::
+    getGuideRateInjectionGroupValues(const Group& group) const
+    {
+        auto grval = data::GuideRateValue{};
+
+        assert (this->guideRate_ != nullptr);
```
joakim-hove, line 2971: According to Joakim's principle on assert: https://github.com/OPM/opm-simulators/issues/3117 - which is still undisputed, this should be a `std::logic_error()`

opm/simulators/wells/TargetCalculator.hpp
```diff
@@ -163,6 +163,126 @@ namespace WellGroupHelpers
...
+                target_ = GuideRateModel::Target::OIL;
+                break;
+            }
+            case Phase::GAS: {
+                pos_ = pu.phase_pos[BlackoilPhases::Vapour];
+                target_ = GuideRateModel::Target::GAS;
+                break;
+            }
+            default:
+                assert(false);
```
joakim-hove, line 207: See: https://github.com/OPM/opm-simulators/issues/3117

opm/simulators/wells/BlackoilWellModel_impl.hpp
totto82: I guess it is mainly for convenience. To avoid issues in parallel we currently need to pass around the `deferred_logger` which could be a pain sometimes.  I can give it a try for cases in this PR.

joakim-hove: OK; but the problem is only relevant when/if we want to log the exception? If we let the exception bubble all the way up and kill the process there should be no difference (i.e. entire application terminates) both for `assert()` and `throw`?

totto82:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2949,12 +2955,34 @@ namespace Opm {
...
+
+        const auto& gname = group.name();
+        if (this->guideRate_->has(gname, Opm::Phase::GAS)) {
+            // No guiderates exist for 'gname'.
```
bska, line 2975: Note: This comment directly contradicts the code.  I suspect you copied this from [`getGuideRateValue()`](https://github.com/OPM/opm-simulators/blob/3df75f5ab69d11e16120b824ddcc8c774591743d/opm/simulators/wells/BlackoilWellModel_impl.hpp#L2978-L2981), but that function uses the opposite criterion,
```C++
if (! this->guideRate_->has(...))
```

totto82:
jenkins build this opm-simulators=3112 opm-tests=504 please

totto82:
jenkins build this opm-common=2355 opm-tests=504 please

PR #3114, opened 2021-03-15
Support system pybind11

blattms:
Needs OPM/opm-common#2359

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -1,4 +1,12 @@
-if(NOT TARGET pybind11)
+find_package(pybind11 2.2 CONFIG)
+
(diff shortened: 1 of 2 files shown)


blattms:
@akva2 Would be call if you would glimpse at this and the upstream PR. Will merge tomorrow morning.

akva2:
they both look fine. of course there is potentially lots of devils between all those details but seems you have tested it broadly.

blattms:
Thanks for looking into it.

Concerning the devils there are of course two options:
- By default use the embedded stuff and provide a switch to prefer the system versions.
- Stay with preferring the system versions, but provide one common switch to always use the embedded stuff.

PR #3118, opened 2021-03-18
Throw "correct" exception type in deferred logger.

joakim-hove:
When the deferred logger is responsible for (rethrowing) an exception we can have the situation that a `std::logic_error` is caught and "transformed" to a `std::runtime_error` - and that again is caught in the simulator loop and the simulation continues with broken state.

Motivation: A `std::logic_error`is thrown in the well assembly, and then handled here: https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/BlackoilWellModel_impl.hpp#L969

I have hacked something up to achieve this:
~~A more elaborate implementation would probably rethrow the same exception type that had been thrown originally - was not immediately obvious to me how to achieve that.~~

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -892,7 +892,7 @@ private:
         static constexpr int zCoord = Element::dimension - 1;
         Scalar zz = 0.0;
 
-        const Geometry geometry = element.geometry();
+        const Geometry& geometry = element.geometry();
         const int corners = geometry.corners();
         for (int i=0; i < corners; ++i)
             zz += geometry.corner(i)[zCoord];
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -414,14 +414,13 @@ protected:
         Scalar zz1 = 0.0;
         Scalar zz2 = 0.0;
 
-        const Geometry geometry = element.geometry();
-        const int corners = geometry.corners();
+        const Geometry& geometry = element.geometry();
         // This code only works with CP-grid where the
         // number of corners are 8 and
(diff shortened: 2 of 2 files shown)


joakim-hove:
jenkins build this please

blattms:
Easiest would be an enum with values e.g. NONE=0, LOGIC, RUNTIME, ..., DEFAULT, communicate that and se a switch statement to throw.

joakim-hove:
> Easiest would be an enum with values e.g. NONE=0, LOGIC, RUNTIME, ..., DEFAULT, communicate that and use a switch statement to throw the correct type.

OK thank you - that I can implement.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -342,11 +342,17 @@ namespace Opm {
                     setRepRadiusPerfLength();
                 }
             }
+        } catch (std::runtime_error& e) {
+            exc_type = ExceptionType::RUNTIME_ERROR;
+        } catch (std::invalid_argument& e) {
+            exc_type = ExceptionType::INVALID_ARGUMENT;
+        } catch (std::logic_error& e) {
+            exc_type = ExceptionType::LOGIC_ERROR;
```
bska, line 350: We should typically catch objects as "reference-to-const", not as "reference-to-mutable".

joakim-hove:
Update: catch all exceptions by const ref + add original exceptions message too output.

PR #3125, opened 2021-03-22
Report Dimensionless Time and Pressure for CT Aquifers

bska:
This commit adds support for calculating and reporting the dimensionless time (simulator time divided by aquifer's time constant) and pressure (influence function evaluated at dimensionless time) values as part of the Carter-Tracy aquifer's `aquiferData()` reporting function.

These values are useful in their own right, e.g., for summary output through the keywords `AAQTD` and `AAQPD`, but they are also needed for ECLIPSE restart purposes.

### DIFF
--- a/opm/simulators/aquifers/AquiferCarterTracy.hpp
+++ b/opm/simulators/aquifers/AquiferCarterTracy.hpp
@@ -26,6 +26,7 @@
 #include <opm/output/data/Aquifer.hpp>
 
 #include <exception>
+#include <memory>
 #include <stdexcept>
 
 namespace Opm
@@ -80,7 +81,12 @@ public:
         }
         data.volume = this->W_flux_.value();
         data.initPressure = this->pa0_;
-        data.type = Opm::data::AquiferType::CarterTracey;
+        data.type = Opm::data::AquiferType::CarterTracy;
+
+        data.aquCT = std::make_shared<data::CarterTracyData>();
+        data.aquCT->dimensionless_time = this->dimensionless_time_;
+        data.aquCT->dimensionless_pressure = this->dimensionless_pressure_;
+
         return data;
     }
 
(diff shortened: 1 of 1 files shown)


bska:
Note: This PR depends on the expanded CT data transport protocol introduced in OPM/opm-common#2374 and must be merged along with that PR.

opm/simulators/aquifers/AquiferCarterTracy.hpp
```diff
@@ -92,17 +98,27 @@ class AquiferCarterTracy : public AquiferInterface<TypeTag>
     Scalar mu_w_{1}; // water viscosity
     Scalar fluxValue_{0}; // value of flux
 
+    Scalar dimensionless_time_{0};
```
joakim-hove, line 101: Can we just use `double`? Until we start doing `<float32>` and `<float128>` simulators on a regular basis I consider the templated scalar type to just be a unwanted complexity-

opm/simulators/aquifers/AquiferCarterTracy.hpp
```diff
@@ -92,17 +98,27 @@ class AquiferCarterTracy : public AquiferInterface<TypeTag>
...
+    Scalar dimensionless_pressure_{0};
+
     void assignRestartData(const data::AquiferData& /* xaq */) override
     {
         throw std::runtime_error {"Restart-based initialization not currently supported "
                                   "for Carter-Tracey analytic aquifers"};
     }
 
-    inline void getInfluenceTableValues(Scalar& pitd, Scalar& pitd_prime, const Scalar& td)
+    inline void getInfluenceTableValues(const Scalar td_plus_dt,
```
joakim-hove, line 110: Structured bindings gives a very elegant way of unpacking multiple values - I would suggest returning a `std::pair()` instead of passing mutable references.

opm/simulators/aquifers/AquiferCarterTracy.hpp
```diff
@@ -117,12 +133,16 @@ class AquiferCarterTracy : public AquiferInterface<TypeTag>
     inline void calculateEqnConstants(Scalar& a, Scalar& b, const int idx, const Simulator& simulator)
```
joakim-hove, line 133: `a` and `b` can be returned as a `std::pair()`

bska: > `a` and `b` can be returned as a `std::pair()`

I've pushed an update to make it so.

opm/simulators/aquifers/AquiferCarterTracy.hpp
bska: > would suggest returning a `std::pair()` instead of passing mutable references.

I've pushed an update to do this.

opm/simulators/aquifers/AquiferCarterTracy.hpp
bska: > Can we just use `double`?

I'm just trying to stay in the style of the rest of the module.

> Until we start doing `<float32>` and `<float128>`

There are experimental versions that do that.

bska:
jenkins build this opm-common=2374 please

PR #3133, opened 2021-03-24
only when both oil and gas exist, we handle drsdt and drvdt

GitPaean:


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
With this approach, we ignore the DRSDT, DRVDT, when not both oil and gas phase exist from the simulator side. 

Maybe we should check `DISGAS` and `VAPOIL`, but so far we have not used that way.

totto82 (approved):

bska:
Approved and build check green.  I'll merge into master.

PR #3135, opened 2021-03-26
Adds a simple test case for gas lift optimization

hakonhagland:
Note: this PR depends on #3097 which should be merged first.

Adds a simple test case for gas lift optimization. Currently this is very simplistic and only covers a fraction of the gas lift optimization code. The plan is to use this as a building block to add more tests in the future.


hakonhagland:
jenkins build this please

CMakeLists_files.cmake
```diff
@@ -138,6 +139,16 @@ list (APPEND TEST_DATA_FILES
   tests/rhs3rep.txt
   tests/options_flexiblesolver.json
   tests/options_flexiblesolver_simple.json
+  tests/GLIFT1.DATA
```
joakim-hove, line 142: Just a heads up about a utility. If you want to have everything in one file you can use `opmpack` which is built as part of opm-common:
```
opmpack GLIFT1.DATA > GLIFT1_ONEFILE.DATA
```

tests/test_glift1.cpp
```diff
@@ -0,0 +1,174 @@
...
+    well_model.beginReportStep(report_step_idx);
+    well_model.beginTimeStep();
+    well_model.updatePerforationIntensiveQuantities();
+    Opm::DeferredLogger deferred_logger;
+    well_model.calculateExplicitQuantities(deferred_logger);
+    well_model.prepareTimeStep(deferred_logger);
+    well_model.updateWellControls(deferred_logger, /* check group controls */ true);
+    well_model.initPrimaryVariablesEvaluation();
+    Opm::WellInterface<TypeTag> *well_ptr = well_model.getWell("B-1H").get();
+    StdWell *std_well = dynamic_cast<StdWell *>(well_ptr);
```
joakim-hove, line 146: Very very good that you have made this first effort - hopefully this can be a first step towards more testable code in opm-simulators!
image

hakonhagland:
jenkins build this please

bska (commented):
I did not expect to see changes to `BlackoilWellModel.hpp` in a PR that claims to do no more than add a test case.  Do you need to rebase this PR onto the most recent master sources?

hakonhagland:
> I did not expect to see changes to `BlackoilWellModel.hpp`

@bska Yes I agree this looks suspicious..It was a quick-and-dirty fix to make some methods in `BlackoilWellModel`  `public` (instead of `protected`) so they could be called from the test case.

totto82:
Thanks for the contribution. This test could also work as a starting point for adding more tests in opm-simulators.

Issue #3132, opened 2021-03-23
Geothermal case: heat flow balance error with varying DZ


MarkTNO:
Good afternoon,

We have a horizontally continuous model with only water present with two wells: one injector (injecting 30 degC water) and one producer. This creates a cold water bubble around the injector.
The 'simple' model (on the left in the picture below) has a continuous grid spacing. In the 'simple_1' model (on the right in the picture below) the over- and underburden consist of 1 layer with large DZ. The cold water bubble in the simple_1 model on the right is much smaller.
This 'simple_1' model was also run in Eclipse100 which gave a similar result to the simple model.

Additionally we compared the injected heat in the well (cold: 30 degC vs initial temp 84 degC) to the total cooling of all cells in the models (per grid cell: _delta_temp * bulk_volume * (1-por) * SPECROCK + deltaTemp * bulk_volume * por * SPECHEAT_) :
- injection well: _(30 degC - 84 degC) * 30 years * 365 days * 7200 m3/day * 3.8 kJ/kg/K * 1000kg/m3 = -1.62E13 kJ_ injected heat
- simple model OPM: _-1.6E13 kJ_ sum delta heat of all cells
- simple_1 model OPM: _-1.0E13 kJ_
- simple_1 model Eclipse100: _-1.59E13kJ_

The -1.0E13kJ for simple_1 model OPM is too small. This calculation confirms our visual conclusion based on the pictures below that the cold water bubble in the 'simple_1' model is too small. Dividing the over- and underburden into two z layers gave a similar result for simple_1 OPM, meaning that the bubble is still too small.

I can provide the OPM and Eclipse decks if needed. I discussed this with my colleague Cintia (@goncalvesmachadoc). We are both available for questions.

Kind regards,
Mark Vrijlandt

![simple](https://user-images.githubusercontent.com/25008935/112176197-2e4ed280-8bf8-11eb-9661-3914c4a8c92b.jpg)

totto82:
Looks like boundary issues. Are you using `flow_onephase_energy ` for your simulations? By default flow impose no thermal flow boundaries, but there is an compile time option to keep the geothermal gradient by allowing for thermal flux across the boundary.  Look for `enableThermalFluxBoundaries `in `eclproblem.hh`.

MarkTNO:
We are using flow with a shallow contact above the model, meaning only water in the model.

[...]

PR #3138, opened 2021-03-26
Fix thermal conductivity

totto82:
Depends on OPM/opm-models#645

Fix issue reported in #3132 
Change to compute the thermal conductivity the same way as for the permeability and diffusivity. It now used the cell center from eclipseState not the grid.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1469,13 +1469,26 @@ public:
      * \copydoc EclTransmissiblity::thermalHalfTransmissibility
      */
     template <class Context>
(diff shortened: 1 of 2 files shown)


bska (approved):

MarkTNO:
It looks much better now, see image below with old results on the left, new results on the right.
On the second image in the difference with eclipse after 50 years, max 1. 4 degC difference.

image
image

totto82:
Great. Please close your issue if you are happy. https://github.com/OPM/opm-simulators/issues/3132

PR #3116, opened 2021-03-17
Update ecl wells for all matching wells after actionx

joakim-hove:

PR #3140, opened 2021-03-26
Access well state variables through accessor functions

joakim-hove:
This is effectively no change - but for me it is easier to read; the motivation is https://github.com/OPM/opm-simulators/pull/3116 which needs an additional storing away of the previous well state. With this PR that can be implemented as:
```
well_model.commitWellState();
```
whereas the same functionality in current master must be implemented as:
```
well_model.prev_well_state_ = well_model.well_state_
```

That might be a bit into the future - but I would assume this was also a smell step in a direction to wrangle the mutable state object out of the model object.

Use of this PR is illustrated in the last commit of #3116

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -215,14 +215,113 @@ namespace Opm {
                 return grp_nwrk_values;
             }
 
+
+            /*
+              The dynamic state of the well model is maintained with an instance
+              of the WellStateFullyImplicitBlackoil class. Currently we have
+              three different wellstate instances:
+
+               1. The currently active wellstate is in the active_well_state_
+                  member. That is the state which is mutated by the simulator.
+
+               2. In the case timestep fails to converge and we must go back and
+                  try again with a smaller timestep we need to recover the last
+                  valid wellstate. This is maintained with the
+                  last_valid_well_state_ member and the functions
+                  commitWellState() and resetWellState().
+
+                3. For the NUPCOL functionality we should either use the
+                   currently active wellstate or a wellstate frozen at max
+                   nupcol iterations. This is handled with the member
+                   nupcol_well_state_ and the initNupcolWellState() function.
+            */
+
+
+            /*
+              Immutable version of the currently active wellstate.
+            */
+            const WellStateFullyImplicitBlackoil& wellState() const
+            {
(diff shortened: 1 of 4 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

opm/simulators/wells/BlackoilWellModel.hpp
```diff
@@ -215,14 +215,69 @@ namespace Opm {
                 return grp_nwrk_values;
             }
 
+            const WellStateFullyImplicitBlackoil& wellState() const
+            {
```
totto82, line 219: Can you add some explaining text before the accessor functions.

totto82 (approved):
After thinking a bit I agree with your changes. I also see some potential for further improvement by accessing the wellState via functions. Nice if you could add some text to the new accessor functions. I am probably to blame for the missing explanation in the first place....

opm/simulators/wells/BlackoilWellModel.hpp
joakim-hove: OK - I have tried to add some documentation. Hopefully it is clearer now.

joakim-hove:
jenkins build this please

opm/simulators/wells/BlackoilWellModel.hpp
totto82: Thanks!

opm/simulators/wells/BlackoilWellModel.hpp
```diff
@@ -215,14 +215,113 @@ namespace Opm {
...
+                this->active_well_state_ = this->last_valid_well_state_;
+            }
+
+            /*
+              Will store the current active wellstate in the nupcol_well_state_
+              member. This can then be subsequently retrieved with accessor
+              nupcolWellState().
+            */
+            void initNupcolWellState()
```
atgeirr, line 313: I suggest the name updateNupcolWellState() instead, since this will be called for each of the first NUPCOL iterations.

joakim-hove: OK

joakim-hove:
jenkins build this please

PR #3144, opened 2021-04-07
Use value object for endpoints

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/2412


bska:
Merging to restore build following OPM/opm-common#2412 and OPM/opm-material#450.

PR #3147, opened 2021-04-09
fixed: we have to "link" these to the fmt::fmt target

akva2:
if not, the required defines are not set when building
with external libfmt


akva2:
jenkins build this please

blattms:
Thanks. Probably did miss some magic switch during testing.

PR #3148, opened 2021-04-09
make gaslift parallel

totto82:
I am working on making the gaslift code parallel, but I bang my head into an issue I have not been able to solve. 

The code exits with `free(): invalid pointer. `
after the vector is communicated through `allgatherv` 

Cool if I could get some help on this....

### DIFF
--- a/opm/simulators/wells/GasLiftStage2_impl.hpp
+++ b/opm/simulators/wells/GasLiftStage2_impl.hpp
@@ -363,6 +363,14 @@ getCurrentWellRates_(const std::string &well_name, const std::string &group_name
         success = true;
         if ( this->debug_) debug_info = "(B)";
     }
+
+    if (well_ptr) {
+        // we only want rates from wells owned by the rank
+        if (!well_state_.wellIsOwned(well_ptr->indexOfWell(), well_name)) {
+            success = false;
+        }
+    }
+
     if (success) {
         assert(well_ptr);
         assert(well_ptr->isProducer());
@@ -802,17 +810,23 @@ localToGlobalGradVector_(const std::vector<GradPair> &grads_local) const
     if (comm.size() == 1)
         return grads_local;
 
+    using Pair = std::pair<int, double>;
+    std::vector<Pair> grads_local_tmp;
+    grads_local_tmp.reserve(grads_local.size());
+    for (size_t i = 0; i < grads_local.size(); ++i) {
+        if(!well_state_.wellIsOwned(grads_local[i].first))
+            continue;
+
+        grads_local_tmp.push_back(std::make_pair(well_state_.wellNameToGlobalIdx(grads_local[i].first), grads_local[i].second));
+    }
+
     std::vector<int> sizes_(comm.size());
     std::vector<int> displ_(comm.size() + 1, 0);
-    int mySize = grads_local.size();
+    int mySize = grads_local_tmp.size();
     comm.allgather(&mySize, 1, sizes_.data());
     std::partial_sum(sizes_.begin(), sizes_.end(), displ_.begin()+1);
-    using Pair = std::pair<int,double>;
     std::vector<Pair> grads_global(displ_.back());
-    std::vector<Pair> grads_local_tmp(grads_local.size());
-    for (size_t i = 0; i < grads_local.size(); ++i) {
-        grads_local_tmp[i] = std::make_pair(well_state_.wellNameToGlobalIdx(grads_local[i].first), grads_local[i].second);
-    }
+
     comm.allgatherv(grads_local_tmp.data(), grads_local_tmp.size(), grads_global.data(), sizes_.data(), displ_.data());
     std::vector<GradPair> grads(grads_global.size());
     for (size_t i = 0; i < grads_global.size(); ++i) {
@@ -871,7 +885,13 @@ bool
 GasLiftStage2<TypeTag>::OptimizeState::
 checkAtLeastTwoWells(std::vector<GasLiftSingleWell *> &wells)
 {
-    int numberOfwells = wells.size();
+    int numberOfwells = 0;
+    for (auto well : wells){
+        int index_of_wells = well->getStdWell().indexOfWell();
+        if (!this->parent.well_state_.wellIsOwned(index_of_wells, well->name()))
+            continue;
+        numberOfwells++;
+    }
     const auto& comm = this->parent.ebos_simulator_.vanguard().grid().comm();
     numberOfwells = comm.sum(numberOfwells);
     if (numberOfwells < 2) {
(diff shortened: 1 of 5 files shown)


akva2:
try
```
LD_PRELOAD=/usr/lib/x86_64-linux-gnu/valgrind/libmpiwrap-amd64-linux.so mpirun ...
```
(assuming debian derivate)

should give you the help you need.

akva2:
alternatively
```
mpirun -np x <your-favorite-terminal> -e gdb <binary>
```

to run in debugger (you'll get x debuggers, one for each process).

opm/simulators/wells/GasLiftStage2_impl.hpp
```diff
@@ -574,17 +573,41 @@ redistributeALQ_(std::vector<GasLiftSingleWell *> &wells,  const Opm::Group &gro
...
+    std::vector<int> displ2_(comm.size() + 1, 0);
+    int mySize2 = dec_grads_local.size();
+    comm.allgather(&mySize2, 1, sizes2_.data());
+    std::partial_sum(sizes2_.begin(), sizes2_.end(), displ2_.begin()+1);
+
+    std::vector<GradPair> inc_grads_global(displ_.back());
+    std::vector<GradPair> dec_grads_global(displ2_.back());
+
+    comm.allgatherv(inc_grads_local.data(), inc_grads_local.size(), inc_grads_global.data(), sizes_.data(), displ_.data());
+    comm.allgatherv(dec_grads_local.data(), dec_grads_local.size(), dec_grads_global.data(), sizes2_.data(), displ2_.data());
```
akva2, line 600: GradPair is not POD, it's a pair of a string and a double.  Unless the communicator is really magic, this can't possibly work?

totto82: Your probably right. I asked @blattms about this and got the impression that is should work, but I probably misunderstood. 
But I noticed that this code seems to work. 

https://github.com/OPM/opm-simulators/blob/3d738e83837f39bce036a22835203623fdb8dc95/opm/simulators/wells/ParallelWellInfo.cpp#L57-L75

I can try to communicate an index + double  pair instead and get the name from the index.

akva2: that should work just fine. but data with variable size / pointer members won't. you can't define a MPI data type for variable sized data.

totto82: Communicating a pair of index and double work. Thanks for help. It is not ready yet for other reasons. But the issue I had is fixed. I will continue on Monday....

totto82:
This or at least some fix that avoid spamming the output with GLIFT + MPI not supported should be part of the release. I will add the release tag an try to fix is as soon as possible.

totto82:
jenkins build this please

totto82:
I finally found the issue. It now reproduce serial results for 4_GLIFT_MODEL5.

bska:
> I finally found the issue. It now reproduce serial results for 4_GLIFT_MODEL5.

Would you mind addressing the [new warnings](https://ci.opm-project.org/job/opm-simulators-PR-builder/2063/gcc/new/fileName.1984391464/) that are generated by this PR?

opm/simulators/wells/GasLiftStage2_impl.hpp
```diff
@@ -665,11 +692,18 @@ removeSurplusALQ_(const Opm::Group &group,
...
-                dec_grad_itr, /*increase=*/false, dec_grads, inc_grads);
+                        dec_grad_itr, /*increase=*/false, dec_grads, inc_grads);
+
+            // The dec_grads and inc_grads needs to be syncronized across ranks
+            // We first restrict the vectors to only wells on own rank
+            // before we gather all wells from all ranks
+            auto dec_grads_local = globalToLocalGradVector_(dec_grads);
+            auto inc_grads_local = globalToLocalGradVector_(inc_grads);
+            dec_grads = localToGlobalGradVector_(dec_grads_local);
+            inc_grads = localToGlobalGradVector_(inc_grads_local);
```
blattms, line 705: Shouldn't we just do the extraction of the local grads in localToGlobalGradVector? Currently we are creating a vector (with pairs with a string as the first entry) in globalToLocalGradVector_, just to create another vector (with pairs with an int as the first entry).

totto82:
jenkins build this please

opm/simulators/wells/GasLiftStage2_impl.hpp
```diff
@@ -1010,17 +1092,35 @@ checkOilTarget()
...
+        delta_oil = factor * (gi.new_oil_rate - state.oilRate());
+        delta_gas = factor * (gi.new_gas_rate - state.gasRate());
+        delta_alq = factor * (gi.alq - state.alq());
+    }
+
+    // and communicate the results
+    const auto& comm = this->parent.ebos_simulator_.vanguard().grid().comm();
+    delta_oil = comm.sum(delta_oil);
+    delta_gas = comm.sum(delta_gas);
+    delta_alq = comm.sum(delta_alq);
```
blattms, line 1118: std::array and only one summation?
```
std::array<double, 4> deta{};
auto& [delta_oil, delta_gas, delta_alg] = delta;
....
comm.sum(delta.data(), delta.size());
```

opm/simulators/wells/GasLiftStage2_impl.hpp
blattms: We might want to make sure that these functions do not do anything for a serial run. There is a search involved!

blattms (commented):
Just some comments about improvements.

totto82:
@blattms Thanks for the review. I have followed your lead.

totto82:
jenkins build this please

blattms:
I think this Pr is ok for the case where no well is distributed, but it seems broken if not.

A distributed well it will show up on multiple processors in `gliftwells` and the sums will be wrong. If there are no computations on the perforations in the stages this might be easy to fix by filtering the wells using `BlackoilModel::local_parallel_well_info_[well_index].isOwner() well index is the index of the well in `BlackoilModel::ecl_wells_`

totto82:
> I think this Pr is ok for the case where no well is distributed, but it seems broken if not.

I think this should be easy to fix. I just need to add some ` if(!wellIsOwned)` here and there.

blattms:
You are probably right. Cool if you would do that.

totto82:
> You are probably right. Cool if you would do that.

Could you check my latest commit. It was a bit hard to test. For 2 cores with and without the last commit gives the same results.  For 4 cores there are differences, but they may be related to differences in convergence.

totto82:
jenkins build this please

totto82:
@blattms Cool if this can be merged today so I can include it in the next iteration of the release branch.

PR #3150, opened 2021-04-09
Always print CHOW_PATEL info if used

Tongdongq:
As CHOW_PATEL usage will be very rare, it's info is only printed when enabled.
Printing the number of colors found is printed more often, as that info is generally interesting.
`BILU0 reordering strategy: level_scheduling` is always printed, with the chosen reordering.


blattms:
jenkins build this please

PR #3153, opened 2021-04-12
Report cell volume removed due to MINPV

joakim-hove:
Report cell volume removed due to pinch.

### DIFF
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -41,8 +41,9 @@
 
 #include <dune/common/version.hh>
 
-#include <sstream>
 #include <functional>
+#include <numeric>
+#include <sstream>
 
 namespace Opm {
 template <class TypeTag>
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

ebos/eclcpgridvanguard.hh
```diff
@@ -351,16 +352,35 @@ protected:
...
+            OpmLog::info(fmt::format("Total number of active cells: {} / total volume: {:0.0f} {}", grid_->numCells(), total_volume , volume_unit));
+
+            double removed_volume = 0;
+            for (const auto& cell_index : removed_cells)
+                removed_volume += input_grid->getCellVolume(cell_index);
+
+            if (removed_volume > 0) {
+                removed_volume = unit_system.from_si( UnitSystem::measure::geometric_volume, removed_volume );
+                OpmLog::info(fmt::format("Removed {} cells with a total volume of {:0.0f} {} ({:5.3f} %) due to PINCH", removed_cells.size(), removed_volume, volume_unit, 100 * removed_volume / total_volume));
+            }
```
bska, line 381: Any chance we report the deactivated pore-volume too?

joakim-hove: Sure; I'll leave this open for @alfbr to comment - but we can report any combination geometric-volume and/or pore-volume

alfbr: Actually, I am a bit confused, so I think I at least need to understand better before making any comment. Is it not MINPV (or MINPVV) that actually removes the pore volume?

joakim-hove: Yes - you are right. Have upated.

joakim-hove:
OK; when running Norne you will now get the following message:
```
_______ Blank line here _______
Processing grid
Total number of active cells: 44431 / total pore volume: 673248728 SM3
Removed 467 cells with a total pore volume of 95962 SM3 (0.014 %) due to MINPV/MINPVV
```

alfbr:
Perfect! The second line is a bit long if we still aim to be within approximately 80 characters (the line is 85 characters with a three digit number of cells removed). If we remove the word "total" I think it is still clear that it actually is a total, and we are at 79 characters.

joakim-hove:
OK - "total" removed.

joakim-hove:
jenkins build this please

bska:
> ```
> Total number of active cells: 44431 / total pore volume: 673248728 SM3
> ```

Tiny nit: The `SM3` unit should **probably** be `RM3` instead.  These volumes are at reservoir condition.

joakim-hove:
> The SM3 unit should probably be RM3 instead. These volumes are at reservoir condition.

In my opinion this should be `SM3` - it is at reservoir conditions, but the volume we are talking about is a geometric voidage volume - unrelated to the phase(s) in the pores. Only reservoir condition to take into account should be rock compressibility?

bska:
> it is at reservoir conditions, but the volume we are talking about is a geometric voidage volume

On the other hand, we're talking about the volume available for fluids.  Moreover, the unit of measure for the ECLIPSE `PORV` array is `RM3` (METRIC, PVT-M), `RB` (FIELD) or `RCC` (LAB) so there's at least some prior art to suggest that we report this in reservoir volume units.

joakim-hove:
> [....] Moreover, the unit of measure for the ECLIPSE PORV array is RM3 (METRIC, PVT-M

OK - that is a good argument; I have updated.

joakim-hove:
jenkins build this please

atgeirr:
Looks nice! I have another minor request: 673248728 SM3 is hard to read, is it easy to add 1000-separators? I.e. 673,248,728 SM3? I think adding the L modifier to the formatting may do it, but I am not sure how it interacts with the rest of the formatting. The L modifier should ask for locale-dependent formatting, but we ensure Flow runs in the "C" locale so it should be consistent.

joakim-hove:
> 673248728 SM3 is hard to read

Agreed

> is it easy to add 1000-separators?

No. Or at least my attempt with `{:L}` did not produce the desired result. Searching for thousands separators seems like that is a tricky question. I suggest to merge as is - if someone knows how to enforce thousand separators it should be a simple fix, in the absolutely worst case we can of course implement something ourselves ....

bska:
> The L modifier should ask for locale-dependent formatting, but we ensure Flow runs in the "C" locale so it should be consistent.

Right, except in the C locale thousands separators are available for integral values only.  I **suppose** we could emit something like
```
673248728 (6.732e+8)
```
instead, but I don't know that it would be worth the extra characters.

bska:
I'll merge this now.  Additional enhancements to formatting can be made in follow-up PRs.

PR #3156, opened 2021-04-13
Well init restart bug

joakim-hove:
Update: after a review comment from @bska this code has been completely reworked.

If some connections have been closed there will be a difference between the number of connections in the well model and the number of connections loaded from the restart file. This should take that into account.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -630,7 +630,7 @@ namespace Opm {
             const size_t numCells = Opm::UgGridHelpers::numCells(grid());
             const bool handle_ms_well = (param_.use_multisegment_well_ && anyMSWellOpenLocal());
             this->wellState().resize(wells_ecl_, local_parallel_well_info_, schedule(), handle_ms_well, numCells, phaseUsage, well_perf_data_, summaryState, globalNumWells); // Resize for restart step
-            wellsToState(restartValues.wells, restartValues.grp_nwrk, phaseUsage, handle_ms_well, this->wellState());
+            loadRestartData(restartValues.wells, restartValues.grp_nwrk, phaseUsage, handle_ms_well, this->wellState());
         }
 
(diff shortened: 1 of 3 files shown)


joakim-hove:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1916,34 +1916,81 @@ namespace Opm {
...
+                        const auto& rst_connection = rst_well.connections[sched_connection_index];
+
+                        perf_pressure[local_connection_index] = rst_connection.pressure;
+                        perf_rates[local_connection_index] = rst_connection.reservoir_rate;
+                        for (int phase_index = 0; phase_index < np; ++phase_index)
+                            perf_phase_rates[local_connection_index*np + phase_index] = rst_connection.rates.get(phs[phase_index]);
+
+                        local_connection_index++;
+                    }
                 }
```
bska, line 1982: If the [`well_perf_data_`](https://github.com/OPM/opm-simulators/blob/061b5c2395e0928885c714c09dbea2447961dc54/opm/simulators/wells/BlackoilWellModel.hpp#L405) is set up by the time we get here, I think you can probably simplify the logic through [`PerforationData::ecl_index`](https://github.com/OPM/opm-simulators/blob/061b5c2395e0928885c714c09dbea2447961dc54/opm/simulators/wells/PerforationData.hpp#L32-L33)

joakim-hove: Thanks - that was a good suggestion; completely rewritten.

joakim-hove:
jenkins build this please

joakim-hove:
@totto82 : I have marked this with "Release 2021.04" as I think it is a quite important bug fix.

totto82:
Agree, I will back-port it when it is merged.

bska:
> I will back-port it when it is merged.

Feel free to do so at any time because I'm merging this now.

Issue #3154, opened 2021-04-13
Cannot restart Norne in parallel


atgeirr:
When running the NORNE_ATW2013_RESTART.DATA case in parallel (even with just 2 processes) I get a segfault on rank 0 in the  commitWellState() call on line 636 of BlackoilWellModel_impl.hpp. The actual segfaults looks to me like it happens inside copying of a std::map object (possibly `std::map<std::string, std::pair<bool, std::vector<double>>> well_rates`) which is very odd, and could perhaps point to a memory corruption bug.

I tested by first running the regular NORNE_ATW2013.DATA case to create the UNSRT, then running the _RESTART deck modified to use that UNRST file. In serial all is fine, running the restarted case in parallel segfaults. Running in debug mode gives no additional assertation failures. (Performance is no problem since the segfaults happens before the first iteration.)

I was unable to get any further on this, but I think it is a fairly serious bug. Maybe someone could valgrind this?

blattms:
Will take a look until noon.

joakim-hove:
I have also started looking at this; my findings are a slightly different from @atgeirr. I have started with the [norne restart case](https://github.com/OPM/opm-tests/blob/master/norne/NORNE_ATW2013_RESTART.DATA)[1,2]: 
```
flow NORNE_ATW2013_RESTART.DATA
```
and also that blows up hard; so to me it looks like this can be debugged & investigated in serial.

[1]: The input deck in this case is actually broken - fix here: https://github.com/OPM/opm-tests/pull/520
[2]: The restart case loads the restart file `restart/NORNE_ATW2013.UNRST` - so it must be run from the opm-tests directory.

blattms:
Cool, you probably more familiar with restart, anyway.

joakim-hove:
Current investigations seem to be related to the well/connection objects loaded from the deck.

atgeirr:
> [1]: The input deck in this case is actually broken - fix here: [OPM/opm-tests#520](https://github.com/OPM/opm-tests/pull/520)

I was a bit curious as to why there is a separate schedule for the restart case. I assume that it contains only the schedule from the restart point to the end, and that its purpose is to verify that this works, that you do not need the full original deck schedule?

> [2]: The restart case loads the restart file `restart/NORNE_ATW2013.UNRST` - so it must be run from the opm-tests directory.

[...]

PR #3157, opened 2021-04-13
Add test which loads a restart case for norne and runs one timestep

joakim-hove:
Verifies that norne can restart; the results are not checked in any way.

Will fail hard until: https://github.com/OPM/opm-simulators/pull/3156 is merged

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -938,6 +938,13 @@ add_test_compareECLFiles(CASENAME norne
                          PREFIX compareECLInitFiles
                          DIR_PREFIX /init)
 
+# This is not a proper regression test; the test will load a norne case prepared
+# for restart and run one single timestep - of length one day. The results are not
+# verified in any way.
+add_test(NAME NORNE_RESTART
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
Opinions on this?

alfbr:
The test itself looks good  and is a welcome addition. I am a bit in the dark as to how it works though. I assume it requires Norne to be run in order to have something to restart from, so is it triggered when Norne is triggered?

joakim-hove:
>I am a bit in the dark as to how it works though. I assume it requires Norne to be run in order to have something to restart from, so is it triggered when Norne is triggered?

The test is run as a normal integration test in opm-simulators, it is *not* related to running the full norne model at all. The test will load a restart file (which is already located in opm-tests: https://github.com/OPM/opm-tests/tree/master/norne/restart) and then run one simulation day on that model. That takes ~7.5 seconds. The results are not verified in any way - but exit status 0 would have been enough to catch this: https://github.com/OPM/opm-simulators/issues/3154 before it happened.

The not-so-nice thing with this is that when/if the norne restart file is significantly changed someone must create an updated restart locally and update the repo, there is no automated solution for that.

alfbr:
Ah, good. Then unless somebody objects, feel free to merge at your convenience.

PR #3159, opened 2021-04-14
get bhp from thp without solving

totto82:
This makes the code for update target for thp the same for standard wells (stw) producers as for stw injectors and msw injectors and producers. 

@atgeirr Do you remember (or know) why this is done differently for the stw producers.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -1547,17 +1547,12 @@ namespace Opm
             }
             case Well::ProducerCMode::THP:
             {
-                well_state.thp()[well_index] = this->getTHPConstraint(summaryState);
-                gliftDebug(
-                    "computing BHP from THP to update well state",
-                    deferred_logger);
-                auto bhp = computeBhpAtThpLimitProd(well_state, ebos_simulator, summaryState, deferred_logger);
-                if (bhp) {
-                    well_state.bhp()[well_index] = *bhp;
-                } else {
-                    deferred_logger.warning("FAILURE_GETTING_CONVERGED_BHP",
-                                            "Failed to find BHP when switching to THP control for well " + name());
+                std::vector<double> rates(3, 0.0);
+                for (int p = 0; p<np; ++p) {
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

totto82:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1.005    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 1.01     |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 1        |



View result details @ https://www.ytelses.com/opm/?page=result&id=1153

totto82:
I have manually investigated the test failure. Some like UDQ_M1 and 9_2B_DEPL_GCONPROD_2L_STW show noticeable  differences, partly due to different time-step length due to slight differences in the newton convergence.  But qualitatively the results are the same.

atgeirr (changes requested):
I think this is mostly ok. As long as we do not also modify the well rates here (but only bhp) it is better to keep the same algorithm as for the other cases. We could consider calculating proper rates here and modifying those, perhaps that could save iterations?

However, I think we should set the 'well_state.thp()' here, still. This should also be added to injectors, and MSW injectors/producers.

totto82:
> However, I think we should set the 'well_state.thp()' here, still. This should also be added to injectors, and MSW injectors/producers.

I don't think that should have any effect on the simulation. If I recall correctly `well_state.thp()`  is only used for output. In `calculateBhpFromThp,` `getTHPConstraint `is used directly. And `updateThp `is called during `updateWellState `before output.

atgeirr (approved):

totto82:

totto82:
jenkins build this opm-tests=523 please

totto82:
The PR is approved and Jenkins is green. Lets merge it.

bska:
> Lets merge it.

You'll need to bring in the new reference solutions too.

Issue #2889, opened 2020-11-01
2020-10 Brine Test Throws an Exception


OPMUSER:
This worked in 2020-04 but throws

`Program threw an exception: Could not initialize the problem: Tried to evaluate a tabulated function outside of its range`

Decks attached.
[SPE01-BRINE01.zip](https://github.com/OPM/opm-simulators/files/5470520/SPE01-BRINE01.zip)

joakim-hove:
I can confirm the error message in my master. It is not clear to me what the error is - might be trivial, might be a bit more work. I would not hold back the release for this.

bska:
I see this in the release build on my local machine.  The immediate cause is that the `SALTVD` table covers the range of 5000 to 5500 ft, but that the top of the formation is at 8325 ft and the `SALTVD` interpolant is not set up to extrapolate outside its domain of definition.  In this particular case extrapolation is trivial (the `SALTVD` is constant), but that is not always the case and we've not enabled extrapolation yet.

I am a little hesitant to enable the extrapolation option this close to the release without additional testing.  In the 2020.04 release the `SALTVD` data was simply not used&mdash;but recognised in the parser&mdash;and that's why it appeared to run.

OPMUSER:
In that case trap the error and issue an error message, so the user 
knows how to fix it, and of course back port it into the release.

OPM User
opmuser0@gmail.com

atgeirr:
In general, trying to evaluate outside tables is an error condition but not an unusual one. So perhaps it is not appropriate to use exceptions for such cases? We should at least consider flagging the error in some other way. Perhaps by returning an optional? Although that gives us only the success/fail state, not why it failed, so maybe returning a pair of result and error code is better? If you think this would be too much of a change, we should trap the error somewhere with enough context to tell the user that the SALTVD table was the failing one. Except for missing that crucial info, the existing error message is fine.

bska:
I had a look earlier today to see how easy it'd be to get something reasonable here.  I suppose we could put a hack in place by wrapping the [salt concentration derivation](https://github.com/OPM/opm-simulators/blob/0102287de0cb680784bfa4f93f505741efb4f79c/ebos/equil/initstateequil.hh#L1755-L1759) in a `try/catch` block in order to provide additional context to the diagnostic message, but that'd be **very** specific to `SALTVD`.

[...]

PR #3166, opened 2021-04-20
extrapolate saltvd table

totto82:
When reading the release 2020.10 announcement I came over issue #2889. 
This is a fix.

### DIFF
--- a/ebos/equil/initstateequil.hh
+++ b/ebos/equil/initstateequil.hh
@@ -1762,7 +1762,7 @@ private:
                 const auto& cells = reg.cells(i);
                 for (const auto& cell : cells) {
                     const double depth = cellCenterDepth_[cell];
-                    this->saltConcentration_[cell] = saltVdTable_[i].eval(depth);
+                    this->saltConcentration_[cell] = saltVdTable_[i].eval(depth, /*extrapolate=*/true);
                 }
             }
         }


totto82:
jenkins build this please

totto82:
I added a release tag. No need to let the issue survive two releases.

bska:
> This is a fix.

Right&mdash;I'd forgotten about that issue.  Did you test this on several models?

totto82:
Only on the case in the issue. 

Extrapolation is what we do for similar properties. So at-least the code becomes consistent. 
See for instance 

https://github.com/OPM/opm-simulators/blob/32a035f8a54eccffa7572eadc661aa40c71ff720/ebos/equil/initstateequil.hh#L183

in the same file.

bska:
>> Did you test this on several models?
>
> Only on the case in the issue.

I see.  In that case I'm still a little wary of enabling this.  As I said [earlier](https://github.com/OPM/opm-simulators/issues/2889#issuecomment-720163102), extrapolating that case's `SALTVD` table is trivial since the concentration is constant with respect to depth but in other models we probably will not be that lucky.  On the other hand, in a real simulation model the engineer will **hopefully** know better than to end the depth table a full 2825 ft above the top of the formation so it might not be a big problem in practice.

totto82:
>  In that case I'm still a little wary of enabling this

Extrapolation is how we handle all similar properties. Usually without warnings. Adding checks in the input layer in opm-common that produces a warning or throw if the depths is outside the range could be a way to warn the user. I still think this can be merged as is.

alfbr:
> I see. In that case I'm still a little wary of enabling this. As I said [earlier](https://github.com/OPM/opm-simulators/issues/2889#issuecomment-720163102), extrapolating that case's `SALTVD` table is trivial since the concentration is constant with respect to depth but in other models we probably will not be that lucky. On the other hand, in a real simulation model the engineer will **hopefully** know better than to end the depth table a full 2825 ft above the top of the formation so it might not be a big problem in practice.

Thanks for checking this thoroughly! At some point we should do a coordinated effort to provide good warning messages when tables are extrapolated. I agree with your assessment, in this case the engineer will hopefully know better, so from our end we do not need to handle this better now.

bska:
> in this case the engineer will hopefully know better, so from our end we do not need to handle this better now.

Noted.  I'll merge this then and just **hope** we don't get hard-to-debug problems later on.

PR #3170, opened 2021-04-21
fixing the serial running due to including of eclmpiserializer.hh

GitPaean:
reported in https://github.com/OPM/opm-common/pull/2389#issuecomment-824034013


GitPaean:

bska (approved):
Thanks for the quick turn-around.  I'll merge this into master to restore the build.

PR #3171, opened 2021-04-21
Use groupstate

joakim-hove:
On top of: #3165 

With this PR the `GroupState`object is partly exported from the `WellState` class - this is a preparation for pulling it completely out of the WellState - that will  be a PR or two down the road. The amount of churn in this PR is overwhelming - I think there is something to be read from that.

PR #3173, opened 2021-04-22
Wellmodel manage group state

joakim-hove:
On top of #3171

The WellState object no langer manages the GroupState.

This might not be the final destination (probaly not), but on the refactoring journey to Good Enough&trade; I see it as essential to have more explicit datastructures.


joakim-hove:
jenkins build this please

joakim-hove:
superceeded by #3174

Issue #3172, opened 2021-04-22
Official Debian Packages


blattms:
This is for people interested in this effort. The current state is on [salsa.debian.org](https://salsa.debian.org/blattms/opm-simulators). This is currently being finalized (mainly coming up with good package descriptions), and will then be submitted to debian.

blattms:
After preparing packages for the new release, we found a Sponsor (Anton Gladky), who so kind as to upload the packages to the [NEW queue of ftp master.](https://ftp-master.debian.org/new.html). There it will be manually checked (e.g. license, etc.) and either rejected  or added to official Debian. This process can take a while. It is a volunteer effort after all. There 372 packages in the queue (some for 8 month).

Anyway, interaction and support for Anton and Debian (Science) was amazing.

- [ITP bug for OPM at Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=987381)

joakim-hove:
> After preparing packages for the new release, we found a Sponsor

Great news - fantastic effort!

alfbr:
This is great news, thank you for a tedious and impressive effort! Crossing fingers for acceptance :-)

blattms:
Well, you can only do a good job if the base is good. Hence this is a team effort: Good software quality from all OPM devs, a very good starting point for the Debian packages  prepared by @akva2, and the amazing suppport and tools (git-buildpackage, pristine-tar, autopkgtest, linitian, ...) provided by Debian.

blattms:
:tada: :partying_face: Next step taken. Yesterday QA of ftpmaster have accepted opm-common and opm-material into Debian unstable. I assume the rest will follow in due time.

bska:
> Yesterday QA of ftpmaster have accepted opm-common and opm-material into Debian unstable.

Congratulations @blattms!  Great achievement.

atgeirr:
Nice! What does this mean for us in terms of keeping the Debian packages up to date?

blattms:
That we, as in @akva2 and me, mostly maintain the packages and try to find a workflow where we base Ubuntu packages on them to save work. Not sure whether that answer your question.

blattms:
opm-models got accepted, too.

[...]

PR #3175, opened 2021-04-22
debian/control: Improved Debian Package descriptions

blattms:
For the proposal of OPM as a debian package (see issue #3172). Comments/further improvements welcome.


blattms:
Closing as the plan should be to use the effort on https://salsa.debian.org/science-team/opm-simulators in the long run

PR #3177, opened 2021-04-23
improve update wellstate from target for wells where privious well state was trivial 

totto82:


totto82:
jenkins build this please

totto82:
Testing shows good results. I will close this for now, do some refactoring to avoid the MSW, STW code duplication and re-create the PR.

PR #3174, opened 2021-04-22
Use groupstate2

joakim-hove:
This is a second attempt at #3171 and #3173 combined.

This PR will remove the `GroupState` member from the `WellStateFullyIMplicitBlackOil` and create a microscopic class `WGState`which handles the well state and group state together as a pair. I am not saying that this very good, or the final destination for this refactoring, but the current code is more explicit, and that is in my opinion a step in the right direction.

The commits Step 1, Step 2, ... is just the way I organized the work. Will squash everything before merge. 

This quite banal refactoring was really hard work; when such conceptually simple refactorings are so hard I think that is the code trying to tell you something.

PR #3178, opened 2021-04-25
WellState will internalize PhaseUsage member

joakim-hove:
Small improvement in the #3174 effort.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -32,12 +32,12 @@
 namespace Opm {
     template<typename TypeTag>
     BlackoilWellModel<TypeTag>::
-    BlackoilWellModel(Simulator& ebosSimulator)
+    BlackoilWellModel(Simulator& ebosSimulator, const PhaseUsage& phase_usage)
         : ebosSimulator_(ebosSimulator),
-          phase_usage_(phaseUsageFromDeck(ebosSimulator_.vanguard().eclState())),
-          active_well_state_(phase_usage_.num_phases),
-          last_valid_well_state_(phase_usage_.num_phases),
-          nupcol_well_state_(phase_usage_.num_phases)
+          phase_usage_(phase_usage),
+          active_well_state_(phase_usage),
+          last_valid_well_state_(phase_usage),
+          nupcol_well_state_(phase_usage)
     {
         terminal_output_ = false;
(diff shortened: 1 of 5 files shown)


joakim-hove:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -74,6 +74,13 @@ namespace Opm {
         alternative_well_rate_init_ = EWOMS_GET_PARAM(TypeTag, bool, AlternativeWellRateInit);
     }
 
+    template<typename TypeTag>
+    BlackoilWellModel<TypeTag>::
+    BlackoilWellModel(Simulator& ebosSimulator) :
+        BlackoilWellModel(ebosSimulator, phaseUsageFromDeck(ebosSimulator_.vanguard().eclState()))
```
bska, line 80: Note: You can't use `ebosSimulator_` here since the data member is not initialised when we're here.  You'll need to use the `vanguard`from `ebosSimulator` (without the trailing underscore).

joakim-hove: Yes - thank you.

joakim-hove:
jenkins build this please

joakim-hove:
Objections to this?

joakim-hove:
jenkins build this please

opm/simulators/wells/WellState.hpp
```diff
@@ -354,13 +358,13 @@ namespace Opm
         WellState& operator=(const WellState& rhs) = default;
 
     private:
+        PhaseUsage phase_usage_;
```
bska, line 361: Hm, this is a bit dangerous.  I think you'll have to implement user-defined special member functions (i.e., assignment, copy constructor and their "move" counterparts) now.  The [`PhaseUsage`]() structure does not have proper value semantics since it contains two [built-in arrays](https://github.com/OPM/opm-simulators/blob/708645a2c07bb232609bff80f3967d1dc625a060/opm/core/props/BlackoilPhases.hpp#L46-L47) instead of proper objects.

joakim-hove: OK - I'll use `std::array<>` in `PhaseUsage`

joakim-hove:
jenkins build this please

bska:
Thanks a lot for the updates. I'll merge into master now.

PR #3179, opened 2021-04-26
Add json dump to GroupState

joakim-hove:
Add functionality to create a json dump of the GroupState object. Initially created to debug: #3174, but code on My machine indicates that this will also be valuable when debugging the restart capabilities.


joakim-hove:
jenkins build this please

PR #3180, opened 2021-04-26
Remove unused methods from wellmodel

joakim-hove:
These methods are not called - removing.

Maybe they *should* have been called?

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -199,11 +199,6 @@ namespace Opm {
                 timeStepSucceeded(ebosSimulator_.time(), ebosSimulator_.timeStepSize());
             }
 
-            void endEpisode()
-            {
-                endReportStep();
-            }
-
             template <class Context>
             void computeTotalRatesForDof(RateVector& rate,
                                          const Context& context,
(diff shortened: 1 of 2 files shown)


joakim-hove:
jenkins build this please

GitPaean:
maybe some ebos related called it? But if you find this out with `grep`, it is probably correct.

bska:
> maybe some ebos related called it?

Maybe at one time.  At present [`EclProblem<>::endEpisode()`](https://github.com/OPM/opm-simulators/blob/12880be916dc8568fccebb361b3109cb4136baf6/ebos/eclproblem.hh#L1253-L1268) does **not** call `wellModel_.endEpisode()`.  On the other hand, [`EclProblem<>::beginEpisode()`](https://github.com/OPM/opm-simulators/blob/12880be916dc8568fccebb361b3109cb4136baf6/ebos/eclproblem.hh#L1036) **does** call [`wellModel_.beginEpisode()`](https://github.com/OPM/opm-simulators/blob/12880be916dc8568fccebb361b3109cb4136baf6/ebos/eclproblem.hh#L1092-L1093) so from symmetry arguments alone `EclProblem::endEpisode()` should probably call `wellModel_.endEpisode()` too.

totto82:
> Maybe they should have been called?

They should be called from `endEpisode()` in` eclproblem.hh` to be consistent with the the rest i.e. `endTimeStep` etc. 
So please do not merge this. We should rather fix this by adding in to `eclproblem.hh`

joakim-hove:
https://github.com/OPM/opm-simulators/pull/3181

PR #3182, opened 2021-04-26
Forward Potential Rates to Next Report Step

bska:
These are used as part of calculating the BHP/THP when computing new potential rates in models that feature VFP tables and must therefore be properly initialised in the next report step's well state object.

### DIFF
--- a/opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
+++ b/opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
@@ -264,6 +264,12 @@ namespace Opm
                             wellReservoirRates()[ idx ] = prevState->wellReservoirRates()[ oldidx ];
                         }
 
+                        // Well potentials
+                        for( int i=0, idx=newIndex*np, oldidx=oldIndex*np; i<np; ++i, ++idx, ++oldidx )
+                        {
+                            wellPotentials()[ idx ] = prevState->wellPotentials()[ oldidx ];
+                        }
+
                         // perfPhaseRates
                         const int oldPerf_idx_beg = (*it).second[ 1 ];
                         const int num_perf_old_well = (*it).second[ 2 ];


bska:
Note: With this change I can make the integration tests for #3094 pass.  Although very late in the game, I think this is a bug fix that should be in the 2021.04 OPM release.

totto82:
Just a note a fix for this was added to PR #2980 but never merged.

totto82:
> I think this is a bug fix that should be in the 2021.04 OPM release.

I agree with the change, but I am not aware that it hit code in current master.

bska:
> a fix for this was added to PR #2980 but never merged.

I see.  Well, in that case I suggest we get this merged&mdash;it will require updating the reference solutions to at least `MOD4_UDQ_ACTIONX` and possibly `9_3D_GINJ_GAS_MAX_EXPORT_MSW` too&mdash;and then pick up the work on #2980 afterwards.

bska:
> > I think this is a bug fix that should be in the 2021.04 OPM release.
> 
> I agree with the change, but I am not aware that it hit code in current master.

It does.  Whilst debugging the regression failures for #3094 I got to a point where the potential rates for producer `OPL2` were **vastly** different in `MOD4_UDQ_ACTIONX` following a dynamic `WELOPEN SHUT` action.  I finally traced to root cause to looking up VFP values with a flowing gas-oil ratio of 0/0 (NaN, translated to zero for lookup purposes) in master while there was a well-defined (positive) GOR in #3094.  That, in turn, could be traced back to using the potential rates of a different well&mdash;the water injector WID1&mdash;instead of `OPL2`.

This PR fixes that problem.

totto82:
Thanks for the explanation.

joakim-hove:
> That, in turn, could be traced back to using the potential rates of a different well—the water injector WID1

And that in turn could be traced back to a quite stingy approach to metadata in the well state classes.

bska:
As expected, the cases MOD4_UDQ_ACTIONX and 9_3D_GINJ_GAS_MAX_EXPORT_MSW, but no other, [failed](https://ci.opm-project.org/job/opm-simulators-PR-builder/2119/testReport/) with this patch.  I suggest we update the reference solutions for these cases and then merge this PR.

joakim-hove:

bska:
jenkins build this opm-tests=524 please

bska:
jenkins build this opm-tests=524 please

bska:
jenkins build this opm-tests=524 please

bska:
The updated reference solutions have been merged.  I'll merge this now to bring the system back to a consistent state.

totto82:
I wanted to cherry-pick the update in the reference solution to the release. But get into a merge conflicts since some of the files are touched by other non-release updates. Any idea how to fix this? The potentially failed tests affected by this update reference are all test cases that always fails locally on my machine. i.e. number of test cases that fails in the release branch is not affected by cherry-picking the latest updates.

bska:
> I wanted to cherry-pick the update in the reference solution to the release. But get into a merge conflicts since some of the files are touched by other non-release updates. Any idea how to fix this?

It sort of depends on what you wanted to do with the updated reference solutions.  Did you want to run the release branch on the Jenkins CI system and compare to them or did you want to run on your local machine?

I can kind of imagine a messy, and manual, approach to applying the change

1. Convert the target files and **both versions** of the source files "formatted" output using `convertECL`
2. Generate a regular file difference between the "old" and "new" source files
3. Apply that file difference to the formatted target files
4. Convert back to "unformatted" output using `convertECL`
5. Commit as update

In addition to being a manual change, the primary downside is that the "apply file difference" step might fail and then you don't gain anything.  The other downside is that this, if it succeeds, will generate the CI system's notion of the difference and in my experience my local runs never agree with those from the CI system.  I've [outlined](https://github.com/OPM/opm-material/pull/447#issuecomment-800316909) my approach to regular development earlier, but I suspect that won't work if you're testing the release branches.

PR #3185, opened 2021-04-26
Use std::array<> in PhaseUsage

joakim-hove:


joakim-hove:
jenkins build this please

opm/core/props/phaseUsageFromDeck.hpp
```diff
@@ -34,7 +34,7 @@ namespace Opm
     /// Determine the active phases
     inline PhaseUsage phaseUsage(const Phases& phases) {
         PhaseUsage pu;
-        std::fill(pu.phase_used, pu.phase_used + BlackoilPhases::MaxNumPhases + BlackoilPhases::NumCryptoPhases, 0);
+        std::fill(pu.phase_used.begin(), pu.phase_used.end(), 0);
```
bska, line 37:
```C++
pu.phase_used.fill(0);
```

joakim-hove:
jenkins build this please

bska:

blattms:
A very simple value initialization should also do the trick for array:
```
PhaseUsage pu{}; // initialized to zeros
```

bska:
>```C++
> PhaseUsage pu{}; // initialized to zeros
> ```

It's not clear to me if that's sufficient.  The [`PhaseUsage` structure](https://github.com/OPM/opm-simulators/blob/2574634c5f5193f584d62ad6bade23e2b8a31d8a/opm/core/props/BlackoilPhases.hpp#L44) doesn't have initialisers for its members so while the above is okay for the array members, it might not be fully well defined for the others, particularly the `bool has*` members.  To be completely safe, we should amend the `PhaseUsage` structure as follows too
```C++
struct PhaseUsage : public BlackoilPhases
{
    std::array<int, MaxNumPhases + NumCryptoPhases> phase_used{};
    std::array<int, MaxNumPhases + NumCryptoPhases> phase_pos{};

    int num_phases{0};
    bool has_solvent{false};
    bool has_polymer{false};
    bool has_energy{false};
    // polymer molecular weight
    bool has_polymermw{false};
    bool has_foam{false};
    bool has_brine{false};
    bool has_zFraction{false};
};
```

PR #3188, opened 2021-04-27
Record the correct simulation time in ActionState

joakim-hove:
The Actionx machinery records the simulation time when an action has evaluated to true. This PR fixes the recording time - which was locked to report_step time in master, instead of actual simulation time.


joakim-hove:
jenkins build this please

PR #3195, opened 2021-04-29
mebos: hook up gas-water simulator

akva2:
This seems to have been forgotten when the gas-water model was added.

### DIFF
--- a/ebos/mebos_main.cc
+++ b/ebos/mebos_main.cc
@@ -171,10 +171,14 @@ int main(int argc, char **argv)
             return Opm::ebosGasOilMain(argc, argv);
         }
         else if (waterActive && gasActive) {
-            notSupportedErrorStream << "\n"
-                                    << "water-gas simulations are currently unsupported\n";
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

bska:
> This seems to have been forgotten when the gas-water model was added.

I don't think that's an accurate assessment.  Note that gas/water simulations are [disabled](https://github.com/OPM/opm-simulators/blob/dae6b61370747fb5f47ea63718f2feeddff668fc/opm/simulators/flow/Main.hpp#L223-L226) in the main `flow` binary too.  I seem to recall that the feature was not fully functional yet, but I may be mistaken.

@plgbrts: Would you like to comment on the current state of gas/water simulations?

goncalvesmachadoc:
Gas/water is not fully functional yet. We have now a working version locally, there are some things/fixes to be pushed to the master.

bska:
Thank you for your update, @goncalvesmachadoc. I think we should close this particular PR then and let your team enable the feature when you think it's ready.

Issue #3200, opened 2021-05-03
/usr/include/boost/bind.hpp:36:1: note: The practice of declaring the Bind placeholders (_1, _2, ...) in the global namespace is deprecated.


hakonhagland:
I am trying to get rid of some warnings when building `opm-simulators` on Ubuntu 21.04 with gcc 10.3.0, boost version 1.74.0 :
```
$ git clone git@github.com:OPM/opm-simulators.git
$ cd opm-simulators
$ mkdir build
$ cd build
$ cmake ..
$ make
-- The C compiler identification is GNU 10.3.0
-- The CXX compiler identification is GNU 10.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Performing Test HAVE_C99
-- Performing Test HAVE_C99 - Success
-- Found C99: -std=c99  
-- Found Boost: /usr/lib/x86_64-linux-gnu/cmake/Boost-1.74.0/BoostConfig.cmake (found suitable version "1.74.0", minimum required is "1.44.0") found components: system unit_test_framework 
CMake Warning (dev) at /usr/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:273 (message):
  The package name passed to `find_package_handle_standard_args` (CJSON) does
  not match the name of the calling package (cjson).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  /home/hakon/test/opm/opm-common/cmake/Modules/Findcjson.cmake:77 (find_package_handle_standard_args)
  /home/hakon/test/opm/opm-common/cmake/Modules/OpmFind.cmake:135 (find_package)
  /home/hakon/test/opm/opm-common/cmake/Modules/OpmPackage.cmake:189 (find_and_append_package_to)
  /home/hakon/test/opm/opm-common/opm-common-prereqs.cmake:25 (find_package_deps)
  /home/hakon/test/opm/opm-common/build/opm-common-config.cmake:127 (include)
  CMakeLists.txt:60 (find_package)
[...]

PR #3201, opened 2021-05-03
Fixes issue #3200 : boost 1.74 warning

hakonhagland:
Silences the warning

> The practice of declaring the Bind placeholders (`_1`, `_2`, ...) in the global namespace is deprecated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -256,7 +256,7 @@ endif()
 
 opm_set_test_driver(${CMAKE_CURRENT_SOURCE_DIR}/tests/run-parallel-unitTest.sh "")
 
-add_compile_definitions(BOOST_BIND_GLOBAL_PLACEHOLDERS)
+add_definitions(-DBOOST_BIND_GLOBAL_PLACEHOLDERS)
 opm_add_test(test_gatherconvergencereport
   DEPENDS "opmsimulators"
   LIBRARIES opmsimulators ${Boost_UNIT_TEST_FRAMEWORK_LIBRARY}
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

CMakeLists.txt
```diff
@@ -256,6 +256,7 @@ endif()
 
 opm_set_test_driver(${CMAKE_CURRENT_SOURCE_DIR}/tests/run-parallel-unitTest.sh "")
 
+add_compile_definitions(BOOST_BIND_GLOBAL_PLACEHOLDERS)
```
akva2, line 259: minimum cmake is 3.10, add_compile_options appeared in 3.12. use add_definitions(-DBOOST_BIND_GLOBAL_PLACEHOLDERS)

hakonhagland: Thanks, I have updated the PR.

bska:
jenkins build this please

bska:
@akva2, @blattms: Is there any reason for **not** merging this PR ahead of the 2023.10 release?

bska:
> Is there any reason for **not** merging this PR ahead of the 2023.10 release?

I suppose if we raise the minimum Boost requirement to 1.76 (released on the 16th of April 2021) we don't really have to do anything here, but that's probably not possible until the 2024.04 release when we can drop Ubuntu 20.04 LTS.

akva2:
yeah sorry. i have been out sick but i had previously found exactly what you have figured out on your own so i don't think we should address this.

blattms:
No problem.Equivalent rest is the most important thing. Thanks for clarifying then. Should we close this then?

bska:
> Should we close this then?

Yes, I think so.

PR #3202, opened 2021-05-03
add throw for unsupported positive item 9 in EQUIL

totto82:
Currently the simulator will return initial pressure 0 if item 9 is > 0 without any warning to the user other than non-physical result and convergence issues.

### DIFF
--- a/ebos/equil/initstateequil.hh
+++ b/ebos/equil/initstateequil.hh
@@ -1899,6 +1899,14 @@ private:
 
             Details::verticalExtent(cells, cellZMinMax_, comm, vspan);
 
+            const auto acc = rec[r].initializationTargetAccuracy();
+            if (acc > 0) {
+                throw std::runtime_error {
+                    "Cannot initialise model: Positive item 9 is not supported "
+                    "in EQUIL keyword, record " + std::to_string(r + 1)
+                };
+            }
+
             if (cells.empty()) {
                 regionIsEmpty[r] = 1;
                 continue;
@@ -1915,7 +1923,6 @@ private:
 
             ptable.equilibrate(eqreg, vspan);
 
-            const auto acc = eqreg.equilibrationAccuracy();
             if (acc == 0) {
                 // Centre-point method
                 this->equilibrateCellCentres(cells, eqreg, ptable, psat);
@@ -1924,6 +1931,11 @@ private:
                 // Horizontal subdivision
                 this->equilibrateHorizontal(cells, eqreg, -acc,
                                             ptable, psat);
+            } else {
+                // Horizontal subdivision with titled fault blocks
+                // the simulator throw a few line above for the acc > 0 case
+                // i.e. we should not reach here.
+                assert(false);
             }
         }
         comm.min(regionIsEmpty.data(),regionIsEmpty.size());


ebos/equil/initstateequil.hh
```diff
@@ -1924,6 +1924,9 @@ private:
                 // Horizontal subdivision
                 this->equilibrateHorizontal(cells, eqreg, -acc,
                                             ptable, psat);
+            } else {
+                // Horizontal subdivision with titled fault blocks
+                throw std::runtime_error("Cannot initialise: Positive item 9 is not supported");
```
joakim-hove, line 1929: I guess this should rather be handled with the https://github.com/OPM/opm-simulators/blob/master/opm/simulators/flow/KeywordValidation.hpp system?

Maybe this PR could be applied to the release?

totto82: I think it should be applied both places. Where it hits for the developer and with a nice message to the user in the KeywordValidation system.

bska: > guess this should rather be handled with the [KeywordValidation](https://github.com/OPM/opm-simulators/blob/aa1df1352ffad9263e3115d028a7ec09831680c0/opm/simulators/flow/KeywordValidation.hpp) system?

The current validation system isn't really equipped to handle this situation.  I specifically asked about this [item 9](https://github.com/OPM/opm-simulators/pull/3061#discussion_r608612895) case in my earlier review of the new implementation and didn't get an answer.  The validation system demands that we list out all unsupported values in a `std::vector` and in this case that could **in principle** be every positive integer in the range `1..INT_MAX`.  In practice we could probably get away with the positive integers `1..20`, but I still think we should make the validation system more capable in order to handle situations like this.

alfbr: > The current validation system isn't really equipped to handle this situation.

True, but the situation here is I believe that only the value "0" is supported for item 9. Luckily, that situation is supported by the system. I can create a pull request for it. Do we want the simulator to stop if non-zero value for item 9 is encountered?

bska: > but the situation here is I believe that only the value "0" is supported for item 9.

No, that is actually not correct.  We added support for `N<0` about a year ago in #2530.  The `N>0` case is unsupported though.

alfbr: > > but the situation here is I believe that only the value "0" is supported for item 9.
> 
> No, that is actually not correct. We added support for `N<0` about a year ago in #2530. The `N>0` case is unsupported though.

Ah, I see, I read the manual too fast. Then I guess this PR is it until we make "missing features" system even more advanced.

totto82: Does that mean I can start the jenkins and get this in. I really want it in the release.

joakim-hove:
jenkins build this please

ebos/equil/initstateequil.hh
joakim-hove: > Does that mean I can start the jenkins and get this in. I really want it in the release.

Yes - but the exception message should be updated to something like: "Cannot initialise: Positive item 9 is not supported in the EQUIL keyword"

bska: Do all processes get to that branch or do we risk some processes continuing while others throw an exception?

totto82: There is a `continue ` a few lines above. So we will get into trouble in parallel. I will fix.

totto82: I have moved the throw before the continue and verified that it throw nicely also in parallel.

ebos/equil/initstateequil.hh
```diff
@@ -1899,6 +1899,10 @@ private:
 
             Details::verticalExtent(cells, cellZMinMax_, comm, vspan);
 
+            const auto acc = rec[r].initializationTargetAccuracy();
+            if (acc > 0)
+                throw std::runtime_error("Cannot initialise: Positive item 9 is not supported");
```
bska, line 1904: Could you add a little more diagnostic information to this message, please?  Something along the lines of
```C++
throw std::runtime_error {
    "Cannot initialise model: Positive item 9 is not supported "
    "in EQUIL keyword, record " + std::to_string(r + 1)
};
```

bska (commented):
Looks good for the most part, but I agree with @joakim-hove&mdash;if we're going to provide diagnostic messages we should include a little more user-directed information in the message.

totto82:
jenkins build this please

bska (approved):
Thanks for the updates.  I'll merge this into the master branch now and defer release branch handling to you.

ebos/equil/initstateequil.hh
verveerpj: > > guess this should rather be handled with the [KeywordValidation](https://github.com/OPM/opm-simulators/blob/aa1df1352ffad9263e3115d028a7ec09831680c0/opm/simulators/flow/KeywordValidation.hpp) system?
> 
> The current validation system isn't really equipped to handle this situation. I specifically asked about this [item 9](https://github.com/OPM/opm-simulators/pull/3061#discussion_r608612895) case in my earlier review of the new implementation and didn't get an answer. The validation system demands that we list out all unsupported values in a `std::vector` and in this case that could **in principle** be every positive integer in the range `1..INT_MAX`. In practice we could probably get away with the positive integers `1..20`, but I still think we should make the validation system more capable in order to handle situations like this.

There was in fact an answer to that: I noted that having constraints on `int` items is quite doable, albeit a bit of work. It was not done for that pull request, it did not fit in the budgeted time anymore, but it is certainly a good idea for a follow up.

PR #3203, opened 2021-05-03
Ensure default value for item 9 in EQUIL.

alfbr:
Following #3202

### DIFF
--- a/opm/simulators/utils/PartiallySupportedFlowKeywords.hpp
+++ b/opm/simulators/utils/PartiallySupportedFlowKeywords.hpp
@@ -62,6 +62,12 @@ const Opm::KeywordValidation::PartiallySupportedKeywords<std::string> partially_
         },
     },
     {
+        "EQUIL",
+        {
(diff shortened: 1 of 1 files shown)


alfbr:
Too fast. We currently support more than the default.

opm/simulators/utils/PartiallySupportedFlowKeywords.hpp
```diff
@@ -61,6 +61,12 @@ const Opm::KeywordValidation::PartiallySupportedKeywords<std::string> partially_
             {2, {false, {"REVERS"}, std::nullopt}}, // IRREVERS
         },
     },
+    {
+        "EQUIL",
+        {
+            {9, {true, {0}, "Please set item 9 to 0 to ensure convergence."}}, // ORDER_TYPE
```
bska, line 67: As I mentioned [earlier](https://github.com/OPM/opm-simulators/pull/3202#discussion_r624944478) we technically do support `EQUIL(9) < 0` too.  Whether or not we converge in that case is another matter.

PR #3204, opened 2021-05-03
CollectDataToIORank: move to separate compile unit

akva2:
make it a template for grid types. this allows using
explicit template instantation and compile this code
only once per grid type, instead of once per simulator object.

this might be a more controversial change, opinions wanted.


akva2:
jenkins build this opm-grid=531 please

akva2:
jenkins build this opm-grid=531 please

akva2:
jenkins build this please

blattms:
I generally like this but am wondering:
- Shouldn't we make it possible for downstream users to use other grids?
- Can we deduce the grid view type?

akva2:
1) if this is a priority we need to reshuffle the code a little again.
2) possibly, it's a separate entity in the typetag so i kept it that way.

ebos/collecttoiorank.cc
```diff
@@ -0,0 +1,902 @@
...
+
+
+
+template class CollectDataToIORank<Dune::CpGrid,
+                                   Dune::CpGrid,
+                                   Dune::GridView<Dune::DefaultLeafGridViewTraits<Dune::CpGrid>>>;
+template class CollectDataToIORank<Dune::PolyhedralGrid<3,3,double>,
+                                   Dune::PolyhedralGrid<3,3,double>,
+                                   Dune::GridView<Dune::PolyhedralGridViewTraits<3,3,double,Dune::PartitionIteratorType(4)>>>;
+
```
totto82, line 901: I use the CollectDataToIORank in my adaptive gridding prototype simulator based on AluGrid.  Note the `needsReordering` I haven't read the code in details, but is it sufficient to add a AluGrid version here to make it compile and run with AluGrid?

akva2: yes.

totto82: >but in no way does it change the implementation
Good to know. It isn't always easy to tell from the `diff ` when it involves renaming.

akva2:
so, should we make it extendible or not? i would like to point out that this was not the case before either, as none of the headers in ebos/ are installed.

blattms:
If it was not the case before, we should not put in extra effort now.

blattms:
@totto82 should check the branch with his code, though.

totto82:
I fine with this. From my point of view this can be merged.

PR #3207, opened 2021-05-04
TargetCalculator: move to separate compile unit

akva2:
use explicit instantation to avoid some templates being
instanced a lot of places in a lot of objects.


akva2:
jenkins build this please

bska (approved):
Thanks.  I'll merge this into master now.

PR #3208, opened 2021-05-04
fixed: do not include headers into Opm namespace

akva2:
This gave me some headaches..

### DIFF
--- a/opm/simulators/wells/GasLiftStage2_impl.hpp
+++ b/opm/simulators/wells/GasLiftStage2_impl.hpp
@@ -31,6 +31,8 @@
 
 #include <fmt/format.h>
 
+namespace Opm {
+
 template<typename TypeTag>
 GasLiftStage2<TypeTag>::
 GasLiftStage2(
@@ -1140,4 +1142,4 @@ updateRates(const std::string &well_name)
     this->alq += delta_alq;
 }
 
-
+} // namespace Opm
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

opm/simulators/wells/GasLiftStage2_impl.hpp
```diff
@@ -1140,4 +1142,4 @@ updateRates(const std::string &well_name)
     this->alq += delta_alq;
 }
 
-
+}
```
bska, line 1145: Would you mind tagging this closing brace with `// namespace Opm`?

akva2: done.

akva2:
jenkins build this please

bska:

hakonhagland:
> This gave me some headaches

@akva2 Why? I thought I was being clever here :smile: By including the implementation header at that point I thought I avoided having the `namespace Opm { .. }` in the implementation file. But maybe it is not so readable if you are used to having the `namespace Opm {...}` in the implementation?

akva2:
Because you gor Opm::fmt namespace and Opm::Opm namespace for those headers you included on top of the _impl file. you were saved by header guards and the fact that all those were already included but i am reshuffling things so that was no longer the case.

hakonhagland:
> Because you got Opm::fmt namespace and Opm::Opm namespace for those headers you included on top of the _impl file

@akva2  Yes. So then we would get `Opm::Opm` instead of `Opm` since the header was already inside `Opm`.. Thanks for clarifying.

PR #3212, opened 2021-05-05
Store the number of connections in separate vector

joakim-hove:
To prepare for refactor

### DIFF
--- a/opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
+++ b/opm/simulators/wells/WellStateFullyImplicitBlackoil.hpp
@@ -144,7 +144,8 @@ namespace Opm
             perf_skin_pressure_.clear();
             perf_skin_pressure_.resize(nperf, 0.0);
 
-            first_perf_index_.resize(nw+1, 0);
+            num_perf_.resize(nw, 0);
+            first_perf_index_.resize(nw, 0);
             first_perf_index_[0] = 0;
             for (int w = 0; w < nw; ++w) {
                 // Initialize perfphaserates_ to well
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

blattms:
Please don't.
This is a very common and long established pattern in scientific computing to get performant code (due to less cache misses).

joakim-hove:
>Please don't.

OK - this is a multistep plan:

0. Store number of connections and connection offset in separate vectors (i.e. this PR)
1. Store the number of connections and connection offset in a `std::pair<int,int>`
2. Store the well state in a per well datastructure.

Don't know how far this will get; but the transition 0 &rightarrow; 1 should be quite simple in a followup PR. A WIP is here: https://github.com/OPM/opm-simulators/pull/3214

> This is a very common and long established pattern in scientific computing to get performant code (due to less cache misses).

I see that point; hopefully the step 1 described above will reduce the problem. That said I can not see that the code in question is exactly inner loop, and I think the current design has some sign of premature optimization. The WellState class is challenging to work with - and simplifications here are in my opinion valuable.

bska:
I hear the concern about cache misses, but I also think that this particular code could benefit from being organised as a structure of arrays and into an array of structures.  If measurements/profiling highlight this as a bottleneck I think we should investigate hybrid structures that store the actual data in larger arrays and loop over indices into those.  In the end, it all comes down to what we expect to be the primary access patterns.  Do we expect to access most/all quantities associated to a single well before moving on to the next well or do we expect to access a few quantities for all wells at a time?

In the meantime I'm going to merge this to enable the refactoring that @joakim-hove outlines.  My own biggest performance concern is moving to a data model that primarily does lookup in terms of names/strings rather than linear indices.

PR #3214, opened 2021-05-05
Wellcontainer

joakim-hove:
A small container class to hold well (or group) objects of type `T`. The class is introduced in order to facilitate safe and partial refactoring of the WellState class - if the WellContainer turns out be useful it can live on - otherwise it should be removed/replaced as part of the refactoring.

### DIFF
--- a/opm/simulators/wells/WellContainer.hpp
+++ b/opm/simulators/wells/WellContainer.hpp
@@ -1,0 +1,125 @@
+/*
+  Copyright 2021 Equinor ASA
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef OPM_WELL_CONTAINER_HEADER_INCLUDED
+#define OPM_WELL_CONTAINER_HEADER_INCLUDED
+
+#include <stdexcept>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace Opm {
+
+template <class T>
+class WellContainer {
+public:
+
+    std::size_t size() const {
+        return this->data.size();
+    }
+
+    void add(const std::string& name, T&& value) {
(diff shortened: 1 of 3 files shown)


opm/simulators/wells/WellContainer.hpp
```diff
@@ -0,0 +1,112 @@
...
+#define OPM_WELL_CONTAINER_HEADER_INCLUDED
+
+#include <stdexcept>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace Opm {
+
+template <class T>
```
bska, line 30: Is this type `T` expected to be "small" in the common case&mdash;e.g., an arithmetic type or an `enum`&mdash;or do you mostly expect to use this container with user-defined/class types that may be expensive to copy?

joakim-hove: Good question; I guess the plan is that `T` should reach up to *medium sized* objects - e.g. a  vector over all connections in a well. 

But maybe more interesting - this class might very well be short lived; it is introduced in order to facilitate safe and partial refactoring of the `WellState` class - if the `WellContainer`turns out be useful it can live on - otherwise it should be removed/replaced as part of the refactoring.

bska: > I guess the plan is that `T` should reach up to _medium sized_ objects - e.g. a vector over all connections in a well.

In that case it might be prudent to change the interfaces that take a `const T&` and copy the value into the internal `data` into taking forwarding references (`T&&`) and using `std::forward<T>(value)` to assign the pertinent `data` values.  That way we preserve the value categories of the arguments which in turn activates move semantics for temporaries and maintain the existing copy semantics for lvalues.

joakim-hove: OK - sounds smart; I must admit my experience with `std::forward<T>` is limited; but I have updated the PR as I think it should be. Cool if you take another look.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

blattms:
I know that the well code is not the best documented code in the OPM universe, but maybe refactoring would be a good time to add some documentation? As it is done now, it will add additional work to anybody working with the code (even and especially those who were kind of familiar with it).

Is the current code using an unordered_map, too? I recall that we relied on a given traversal order of the wells in parallel, But maybe I am wrong?

joakim-hove:
>  but maybe refactoring would be a good time to add some documentation

Will do.

> As it is done now, it will add additional work to anybody working with the code

Well - I *really* hope it will be simpler to work with - both for newcomers and developers already familiar with the code.

> Is the current code using an unordered_map, too?

I'd like to repeat/emphasize the pull request description - i.e. the class is created mainly to enable safer and piecewise refactoring. For the question about `unordered_map` the answer is that container preserves the order used when it is assembled (from the existing code - i.e. no change in order).

PR #3176, opened 2021-04-23
move solveWellEquation inside prepareTimeStep

totto82:
Due to the many test failures of #3158 I decided to chop in into three.This is the first part. 

This is almost a pure refactoring, but note that the  the `updateWellControls(local_deferredLogger, /* check group controls */ false);
` inside the `if (param_.solve_welleq_initially_ && iterationIdx == 0) {` is missing.

PR #3215, opened 2021-05-05
Use mutable WellGroupEvents instance to track events in WellState

joakim-hove:
Requires: https://github.com/OPM/opm-common/pull/2460

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -813,13 +813,14 @@ namespace Opm {
                 if (this->wellTestState_.hasWellClosed(well_name)) {
                     // TODO: more checking here, to make sure this standard more specific and complete
                     // maybe there is some WCON keywords will not open the well
-                    if (this->wellState().effectiveEventsOccurred(w)) {
+                    auto& events = this->wellState().events();
+                    if (events.hasEvent(well_name, WellStateFullyImplicitBlackoil::event_mask)) {
                         if (wellTestState_.lastTestTime(well_name) == ebosSimulator_.time()) {
                             // The well was shut this timestep, we are most likely retrying
                             // a timestep without the well in question, after it caused
                             // repeated timestep cuts. It should therefore not be opened,
                             // even if it was new or received new targets this report step.
-                            this->wellState().setEffectiveEventsOccurred(w, false);
+                            events.clearEvent(well_name, WellStateFullyImplicitBlackoil::event_mask);
                         } else {
                             wellTestState_.openWell(well_name);
                         }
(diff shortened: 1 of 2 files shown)


joakim-hove:
jenkins build this opm-common=2460 please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1686,18 +1687,17 @@ namespace Opm {
...
-                if (this->wellState().effectiveEventsOccurred(w) ) {
-                    this->wellState().setEffectiveEventsOccurred(w, false);
+                    // there is no new well control change input within a report step,
+                    // so next time step, the well does not consider to have effective events anymore
+                    // TODO: if we can know whether this is the first time step within the report step,
+                    // we do not need to set it to false
+                    // TODO: we should do this at the end of the time step in case we will need it within
+                    // this time step somewhere
+                    events.clearEvent(well->name(), WellStateFullyImplicitBlackoil::event_mask);
                 }
```
totto82, line 1701: We need to be a bit careful here. The events may be used by other part of the code.

joakim-hove: Well - og the tests pass the tests pass - or?

totto82: > Well - if the tests pass the tests pass - or?

I guess so. I like your change of making it mutable. That would make it possible to add an event when a well is opened during the simulations. For instance due to WTEST or some ACTION keyword. I have looked through the code and adding the clear logic here is ok. That would assure that updateWellStateWithTarget is only triggered once for each new event. 

I would suggest that we remove the TODO comments and move the clearEvent to just after the updateWellStateWithTarget to make it clear why we want to clear the event. I have a PR #3176 that touch the same code. That would need a rebase after this is merged. I can do the clean-up there.

joakim-hove: OK - I have removed the comment in a fixup. NB: This comment was not originally written by me - it was green in the diff only because of some minor reformatting.

totto82: >  NB: This comment was not originally written by me

I know. So thanks for the cleanup. I must admit I have never understood the logic of he setEffectiveEventsOccurred(w,false) before I read carefully through it today. I think the code is more readable after your fixup commit. You could consider adding a comment like
"We only want update the wellstate with the targets the first time the event occurs." 
before the events.clearEvent(...)
if you plan to do a squash anyway.

joakim-hove: OK - done.

joakim-hove:
jenkins build this opm-common=2460 please

joakim-hove:
jenkins build this opm-common=2460 please

joakim-hove:
jenkins build this opm-common=2460 please

PR #3219, opened 2021-05-06
Separate eclcpgridvanguard.hh into typetag dependent/independent code

akva2:
Stick independent parts in a separate compile unit and use explicit template instantation.

This way, we compile this code only once instead of for every simulator object.

One could argue that more of the template parameters could be removed, but I did not want to remove possible flexibility from the typetag.

This sit on top of https://github.com/OPM/opm-simulators/pull/3218 which should be processed first.
Downstream of OPM/opm-models#655
Downstream of OPM/opm-grid#534

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -24,6 +24,7 @@
 # find opm -name '*.c*' -printf '\t%p\n' | sort
 list (APPEND MAIN_SOURCE_FILES
   ebos/collecttoiorank.cc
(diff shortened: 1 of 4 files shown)


akva2:
jenkins build this opm-grid=534 opm-models=655 please

blattms:
Sorry @akva2, conflicts need to be resolved here.

blattms:
Thanks. merging. Nice work!

blattms:
Will do the dune-fem check as I woud be to blame anyway.

PR #3223, opened 2021-05-07
Separate eclctransmissibility.hh into typetag dependent/independent code

akva2:
This allows using explicit template instantation to only compile this code (a few times) per grid instead of in each simulator object.

This sits on top of https://github.com/OPM/opm-simulators/pull/3219 which must be processed first.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -26,6 +26,7 @@ list (APPEND MAIN_SOURCE_FILES
   ebos/collecttoiorank.cc
   ebos/eclgenericcpgridvanguard.cc
   ebos/eclgenericvanguard.cc
+  ebos/ecltransmissibility.cc
   opm/core/props/phaseUsageFromDeck.cpp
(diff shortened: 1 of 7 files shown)


akva2:
jenkins build this please

akva2:
rebased.

akva2:
jenkins build this please

blattms:
I have one question:
Why exactly do we need two classes `EclTransmissibility<TypeTag>` and EclGenericTransmissibility<GridType1,...>`?
Couldn't we do the same with one class by getting rid of the TypeTag and instead using either the list of types needed or the Type of the vanguard? As it is now `EclTransmissibility<TypeTag>` just has a constructor which might indicate that it is not needed. Where am I wrong?

akva2:
You are not really wrong. I did not know up front that there would be nothing left after I was done splitting. I can definitely change that.

akva2:
there u go.

akva2:
jenkins build this please

blattms:
Cool, thanks.

PR #3231, opened 2021-05-11
Fix function argument

joakim-hove:
Hmmm - green + green might not be green after all.


joakim-hove:
jenkins build this please

akva2:
as a wise old akva2 once said; when one leaves unused variables behind, never rely on somebody else not quelling the noise.

PR #3235, opened 2021-05-11
Small micro-optims in WellInterface

akva2:
Granted, this is more about my OCD than any real-world gains, but we may as well do this to get code elimination on compile time instead of runtime checks.


akva2:
jenkins build this please

PR #3238, opened 2021-05-11
fix regression: do not unconditionally dereference equilCartMapper

### LINKED ISSUE #3237: EclWriter Constructor Dereferences Nullpointer in Parallel When Constructing IO Rank Collection Member
Note: The underlying issue may have existed for a long time, but I came across it just now because I was debugging a parallel restart issue using a build that does **not** define `NDEBUG`.

When constructing the `collectToIORank_` data member of class `EclWriter` we make an unconditional call to [`Vanguard::equilCartesianIndexMapper()`]((https://github.com/OPM/opm-simulators/blob/a192ccb417fe0a0df9f51727a3c2b314f4b56d13/ebos/eclwriter.hh#L197)).  This call, in turn, typically resolves to [`EclCpGridVanguard::equilCartesianIndexMapper()`](https://github.com/OPM/opm-simulators/blob/a192ccb417fe0a0df9f51727a3c2b314f4b56d13/ebos/eclcpgridvanguard.hh#L321) which asserts that the member function is only called on the IO rank.  Indeed, the underlying pointer is null on ranks other than the IO rank whence the member function dereferences a null pointer in order to return a reference to the mapper object.

As I said, I don't know when we introduced this issue because I've not really run debugging builds in while.  We still need to fix the issue though.

akva2:
Closes #3237

### DIFF
--- a/ebos/collecttoiorank.cc
+++ b/ebos/collecttoiorank.cc
@@ -667,10 +667,10 @@ private:
 
 template <class Grid, class EquilGrid, class GridView>
 CollectDataToIORank<Grid,EquilGrid,GridView>::
-CollectDataToIORank(const Grid& grid, const EquilGrid& equilGrid,
+CollectDataToIORank(const Grid& grid, const EquilGrid* equilGrid,
                     const GridView& localGridView,
                     const Dune::CartesianIndexMapper<Grid>& cartMapper,
-                    const Dune::CartesianIndexMapper<EquilGrid>& equilCartMapper)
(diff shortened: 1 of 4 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this please

bska:
I've tested this PR locally and it fixes my debug configuration in parallel.  The change itself looks good so I'll merge this into master once the CI system's build check is green.

bska:
> I'll merge this into master once the CI system's build check is green.

Build check is green.  Merging as advertised.

PR #3239, opened 2021-05-11
Tracer summary curves, collect well rates.

osae:
There is a corresponding PR for opm-common. 

Upstream: https://github.com/OPM/opm-common/pull/2468

### DIFF
--- a/ebos/eclgenerictracermodel.cc
+++ b/ebos/eclgenerictracermodel.cc
@@ -133,14 +133,14 @@ doInit(bool enabled, size_t numGridDof,
 
 
         //TBLK keyword
-        if (!tracer.concentration.empty()){
-            int tblkDatasize = tracer.concentration.size();
+        if (!tracer.free_concentration.empty()){
+            int tblkDatasize = tracer.free_concentration.size();
             if (tblkDatasize < cartMapper_.cartesianSize()){
                 throw std::runtime_error("Wrong size of TBLK for" + tracer.name);
             }
             for (size_t globalDofIdx = 0; globalDofIdx < numGridDof; ++globalDofIdx){
                 int cartDofIdx = cartMapper_.cartesianIndex(globalDofIdx);
-                tracerConcentration_[tracerIdx][globalDofIdx] = tracer.concentration[cartDofIdx];
+                tracerConcentration_[tracerIdx][globalDofIdx] = tracer.free_concentration[cartDofIdx];
             }
         }
         //TVDPF keyword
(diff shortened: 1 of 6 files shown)


joakim-hove:
The PR in opm-common is ~ready for merge.

totto82:
Two general comments
1) This PR partly implements partitioning tracer. i.e. tracer that follow the gas when dissolving into the oil etc. I suggest you either finish the task, or even better I think, remove it from this PR and instead prepare a new PR for this extension in the future. You can add a comment if you want to motive your abstractions and naming of the pure parts.  
2)  I would also appreciate a short explanation about the approach. i.e. that we create one matrix pr phase and solve for multiple right-hand-sides for the individual tracers.

ebos/eclgenerictracermodel.cc
```diff
@@ -243,6 +243,43 @@ linearSolve_(const TracerMatrix& M, TracerVector& x, TracerVector& b)
...
+    TracerSolver solver (tracerOperator, tracerScalarProduct,
+                         tracerPreconditioner, tolerance, maxIter,
+                         verbosity);
+
+    bool converged = true;
+    for (int nrhs =0; nrhs < b.size(); ++nrhs) {
+        x[nrhs] = 0.0;
+        Dune::InverseOperatorResult result;
+        solver.apply(x[nrhs], b[nrhs], result);
+        converged = (converged && result.converged);
```
totto82, line 276: @blattms Can this be solved more efficiently using multiple right-hand-sides directly. i.e. utilizing vectorization.

osae: Thanks for your for comments. Then we agree that:
1. Then the Dual-stuff will be removed.
2. Will add some comments in the code.

Multiple rhs:  Currently we use std::vector as containers for the vectors. Migrating to a dune container is definitively on the agenda, but perhaps not for this PR?

totto82: > Then we agree that:
    1. Then the Dual-stuff will be removed.
    2. Will add some comments in the code.

Yes. I like the way you are planning to include the Dual-stuff. But lets do that in a separate PR later. 

> Multiple rhs: Currently we use std::vector as containers for the vectors. Migrating to a dune container is definitively on the agenda, but perhaps not for this PR? 

The current approach is fine for this PR.

totto82:
jenkins build this please

alfbr:
Is this ready for merging?

alfbr:
It is great to see this coming in, we have a case needing it now. It seems like only default unit is supported currently, on the relevant case unit is set to 'g' (default is sm3). This should not be a blocker for this PR, but we should consider supporting various units before we leave the topic.

osae:
As far as I understand, this is basically about getting correct units for output (internally, the simulator should not care about tracer scaling as long as it is consistent) and it might be a bit overkill to throw because of this. (If you default, and then substitute 'g' for 'sm3' as appropriate when you interpret your output, "everything" would probably be "ok".)   Of course, this needs  to be fixed, and the current throw can be seen as an "un-friendly" reminder.

The issues concerning the accompanying PR https://github.com/OPM/opm-common/pull/2468 seem to be resolved. and merging should be possible.

alfbr:
Thanks for the clarification. In the meantime, it seems the results are not correct. Will share the model with you.

alfbr:
I have now confirmed that the functionality works as advertised. I suggest this goes in, since it is already a functional implementation. There are some deviations, and we need to enable it by default at some point, but I suggest that does not hold back this PR from being merged.

bska:
> I have now confirmed that the functionality works as advertised. I suggest this goes in, since it is already a functional implementation.

Very good.  I'll merge this now.

PR #3241, opened 2021-05-12
Add initializer_list based WellContainer constructor

joakim-hove:


joakim-hove:
jenkins build this please

opm/simulators/wells/WellContainer.hpp
```diff
@@ -44,6 +45,15 @@ template <class T>
 class WellContainer {
 public:
 
+    WellContainer() = default;
+
+
+    WellContainer(const std::initializer_list<std::pair<std::string,T>>& init_list) {
```
bska, line 51: The typical way to pass an `initializer_list` to a function is by value as in
```C++
WellContainer(std::initializer_list<...> args)
```
There's nothing to gain by using a reference to `const` in this case.

joakim-hove: OK - value used

joakim-hove:
jenkins build this please

PR #3222, opened 2021-05-07
Perforations well index

joakim-hove:
Make sure that all loops over the perforations of a well use an index spanning just over that wells perforations ... clear ehh?? Main point is to pull everything related to the offset of the per-well first perforation outside of the loops.

This is a preparation to be able to use a per-well vector for perforation rates.

PR #3242, opened 2021-05-12
Restore Debug Builds

bska:
This was broken in commit b5580f39f (PR #3222) which removed the `start_perf_next_well` symbol but did not update every use location.


bska:
jenkins build this please

bska:
Emergency self-merging this to restore the debug builds.

PR #3243, opened 2021-05-12
Small cleanups in WellInterface

akva2:
The only commit I'm unsure about is https://github.com/OPM/opm-simulators/commit/ab4772aba8c5f62810a7da060520e75e73dedf1b as I'm getting mixed messages whether the intention is to hardcode the well code to double precision or not. If that is the intention, we should use double instead of Scalar to avoid confusion.

### DIFF
--- a/opm/simulators/wells/StandardWell.hpp
+++ b/opm/simulators/wells/StandardWell.hpp
@@ -27,6 +27,7 @@
 #include <opm/simulators/linalg/bda/WellContributions.hpp>
 #endif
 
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

joakim-hove:
> as I'm getting mixed messages whether the intention is to hardcode the well code to double precision or not

I'll give you a clear message: *Use double - for f.. sake*!

akva2:
jenkins build this please

joakim-hove:
Since this PR is using the conservative - stay with `Scalar`- approach it should be totally uncontroversial. Merging.

PR #3244, opened 2021-05-12
add collective communication to generic vanguard

akva2:
@blattms this what you had in mind right.
@jcbowden this should get you going.

### DIFF
--- a/ebos/eclgenericvanguard.hh
+++ b/ebos/eclgenericvanguard.hh
@@ -29,6 +29,10 @@
 
 #include <opm/grid/common/GridEnums.hpp>
 
+#include <dune/common/version.hh>
+#include <dune/common/parallel/collectivecommunication.hh>
+#include <dune/common/parallel/mpihelper.hh>
+
 #include <array>
 #include <memory>
 #include <optional>
@@ -56,6 +60,12 @@ class UDQState;
 class EclGenericVanguard {
 public:
     using ParallelWellStruct = std::vector<std::pair<std::string,bool>>;
+
+#if DUNE_VERSION_NEWER(DUNE_COMMON, 2, 7)
+    using CommunicationType = Dune::Communication<Dune::MPIHelper::MPICommunicator>;
+#else
+    using CommunicationType = Dune::CollectiveCommunication<Dune::MPIHelper::MPICommunicator>;
+#endif
 
     /*!
      * \brief Constructor.
@@ -260,6 +270,17 @@ public:
     const ParallelWellStruct& parallelWells() const
     { return parallelWells_; }
 
+    //! \brief Set global communication.
+    static void setCommunication(std::unique_ptr<CommunicationType> comm)
+    { comm_ = std::move(comm); }
+
+    //! \brief Obtain global communicator.
+    static CommunicationType& comm()
+    {
+        assert(comm_);
+        return *comm_;
+    }
+
 protected:
     void updateOutputDir_(std::string outputDir,
                           bool enableEclCompatFile);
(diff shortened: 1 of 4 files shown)


blattms:
That looks about right. Using a unique-ptr seems a bit complicated. Can't we simply skip the pointer and use the communication object by value? The object should by quite slim and copying the MPI_Comm inside is not an issue.

akva2:
Absolutely. I used unique_ptr to be consistent with other statics in vanguards, but I agree it give us absolutely nothing except complications here.

jcbowden:
Hi,
Can akva2 get me a diff between release https://github.com/OPM/opm-simulators/tree/release/2021.04 and the changes in the akva2:add_collective_comm branch?
I'd like to try to test it and I am working on 2021.04 as master would not compile/run.
Thanks,
Josh

jcbowden:
ok, hold that last request - I can find the changes in "Files changed" above!

jcbowden:
Ok, the three files:
ebos/eclgenericvanguard.cc 
ebos/eclgenericvanguard.hh
opm/simulators/flow/Main.hpp

the first two do not exist in release opm-simulators/2021.04.

Where did they arrive from?

atgeirr:
> Where did they arrive from?

There is a significant refactoring going on to reduce compilation times.

> master would not compile/run.

Let us rather solve this! You can post an issue, or write me an email.

atgeirr:
jenkins build this please

blattms:
As this is one commit @jcbowden can simply do:
```
git checkout release/2021.04
git checkout -b export-comm
git cherry-pick ed3a172a1e40f416bddae33f65d96a392277fbf9
```

jcbowden:
Thanks Markus, I'll give that a try.

blattms:
... and that will fail miserably.

Personally, I would advice against branching of the 2021.04 release as you will have to sort the issues twice...

atgeirr:
Lots of errors, probably all alike:
```
terminate called after throwing an instance of 'Dune::ParallelError'
  what():  ParallelError [CollectiveCommunication:/usr/include/dune/common/parallel/mpicollectivecommunication.hh:153]: You must call MPIHelper::instance(argc,argv) in your main() function before using the MPI CollectiveCommunication!
```

blattms:
MPI thinks that it was not initialized (`MPI_initialized` returns false)

atgeirr:
I think we should get this in, so we can use it as a building block when refactoring the MPI_WORLD_COMM issues. But I believe it must be a unique_ptr not a value object, to defer construction. I think the reason for the error I saw before is that the value object is constructed at static init time, before MPI_Init() is called, using a unique_ptr should solve that.

akva2:
rebased, using a unique_ptr again, tested "working".

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

ebos/eclgenericvanguard.hh
```diff
@@ -260,6 +270,17 @@ public:
...
     { return parallelWells_; }
 
+    //! \brief Set global communication.
+    static void setCommunication(std::unique_ptr<CommunicationType> comm)
+    { comm_ = std::move(comm); }
+
+    //! \brief Obtain global communicator.
+    static CommunicationType& comm()
+    {
+        assert(comm);
```
atgeirr, line 280: This assert looks like it asserts that the function exists.

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
okay, finally got all variants it seems. reasonably happy with this, the python simulators.hpp change is perhaps a little brittle but it is well defined so i don't think it's a biggie.

atgeirr:
Thanks, this is good! Merging.

PR #3260, opened 2021-05-14
Simplify the error message on invalid item values for keywords with only a single record

verveerpj:
This pull request implements a very small improvement in the reporting of invalid item values. The current behavior is to always print the record number, even if there is only a single record. This changes the behavior to only include the record number if there is more than one record.


bska:
jenkins build this please

bska (approved):
Very good.  I'll merge into the master branch.

PR #3261, opened 2021-05-15
Reorder variables and remove .reserve() calls

joakim-hove:
Veeery small trivial changes in MSW WellState code - just to start warming up there.

### DIFF
--- a/opm/simulators/wells/WellStateFullyImplicitBlackoil.cpp
+++ b/opm/simulators/wells/WellStateFullyImplicitBlackoil.cpp
@@ -537,21 +537,18 @@ void WellStateFullyImplicitBlackoil::initWellStateMSWell(const std::vector<Well>
 {
     // still using the order in wells
     const int nw = wells_ecl.size();
-    const auto& pu = this->phaseUsage();
-    const int np = pu.num_phases;
     if (nw == 0) {
         return;
     }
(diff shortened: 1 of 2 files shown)


joakim-hove:
jenkins build this please

blattms:
What is the rational behind removing reserve calls? It seems to be good practice to prevent unnecessary reallocations and copies if possible.

I even fail to see any reason for this PR. Why is it needed?

joakim-hove:
> I even fail to see any reason for this PR. Why is it needed?

It is certainly not needed - just a micro step in my process of getting aquainted with the code, and in my opinion a small *micro* improvement.

The code in itself already has a mix of different methods to set the size of vectors: `.resize(), .assign(), { }` and `push_back`, to me the code is slightly clearer without the use of `reserve()`. The code in question will be refactored further - so this will have essentially zero consequences in the not-so-long-term.

Reordering was just to be consistent with comment (I guess I would get more flack for removing the misleading comment :wink: ?)

blattms:
IMHO, if you know the size you should always use reserve to prevent reallocating, copying and wasting memory. We are already hit by our memory consumption at other place, so let us not make the situation more worse.  https://www.geeksforgeeks.org/using-stdvectorreserve-whenever-possible/

PR #3263, opened 2021-05-15
changed: move Deck parameter to BlackoilModelParametersEbos

akva2:
it is used there. now eclbasevanguard.hh includes
BlackoilModelParametersEbos, instead of the other way around.


akva2:
jenkins build this please

PR #3271, opened 2021-05-18
BUGFIX fix direction of the resv check

totto82:


totto82:
jenkins build this please

totto82:
It seems like we need a test for RESV in prediction mode since a bug like this has survived.

totto82:
A test case is added OPM/opm-tests#533

atgeirr (approved):

PR #3273, opened 2021-05-18
Quell various clang warnings

akva2:
Thank you clang for the two first commits, they point to real issues.
The last one however, is bending over backwards to a false clang diagnostic. Whether we want to do this or not I leave up to others to decide.


akva2:
jenkins build this please

PR #3274, opened 2021-05-18
fix building with dunefem

totto82:
We need a test build with dunefem!


totto82:
jenkins build this please

akva2:
sorry. This is all good. I asked for dune-fem test, markus said he'd do it but never heard back so thought it was all good and thus copy-pasted broken stuff all over..

As for dune-fem, last time i looked into it I struggled to find a release that works fine with dune 2.6 so I gave up. I'll try harder.

totto82:
I use dune 2.7 and always build with dune-fem. Maybe it is time to move to 2.7?

akva2:
move will happen with next debian release which is just around the corner.

totto82:
> This is all good.

I take that as a approval.

PR #3276, opened 2021-05-18
Fixes a bug in GasLiftStage2_impl.hpp

hakonhagland:
When making gas lift parallel, see PR #3148, `redistributeALQ()` did not reserve space for the decremental and incremental gradients. Later in the execution thread when push_back() was called to add elements to the vectors, the capacity of the vector could get exceeded and hence the internal representation of the vector could be reallocated. This seems to have caused access to undefined memory errors since the iterators into the vectors were long longer valid.


totto82:
Thanks for the fix!

totto82:
jenkins build this please

PR #3283, opened 2021-05-20
 Merge wellstate

joakim-hove:
Merge the `WellState`and `WellStateFullyImplcitBlackOil`classes. The final merged class will be named `WellState`


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
Comments?

totto82:
I like this. @atgeirr @bska are we still keeping the legacy branch that use the pure wellState for some of the simpler simulators.

atgeirr:
> are we still keeping the legacy branch that use the pure wellState for some of the simpler simulators

It exists, but is in no way actively maintained, it must be combined with old versions of the other modules in any case. So if there is nothing that uses it in the current code, we do not worry.

joakim-hove:
jenkins build this please

joakim-hove:
I'll merge this when green.

PR #3284, opened 2021-05-20
Some small cleanups in GasLiftWellState

akva2:

PR #3286, opened 2021-05-20
GasLiftSingleWell: split typetag independent code to separate class 

akva2:
Sits on top of https://github.com/OPM/opm-simulators/pull/3284


totto82:
@akva2 The gaslift code is still under development. I would appreciate if we can hold merging this until we finish the development.

akva2:
i can wait but that is rather backwards. It is precisely during development you want to avoid all the full rebuilds..

totto82:
Good point. I thought the PR was in its making, and wanted to avoid yet another rebase. But we found some remaining  issues that needs to be addressed first so maybe it make sense to merge this first.  What do you think @hakonhagland?

hakonhagland:
> It is precisely during development you want to avoid all the full rebuilds..

@akva2 I am not sure exactly what this means? What is the problem with the current code using the type tag? Is this PR making the compilation faster or is it something else? Thanks.

akva2:
by putting the code in a separate compile unit, when you change the code, only that cpp file is rebuilt, not the entire simulator.

hakonhagland:
> What do you think hakonhagland?

@totto82 and @akva2 Maybe we can merge this into master, and after that I rebase our current work on this?

totto82:
jenkins build this please

totto82:
Lets merge this when its green.

bska:
> Let's merge this when its green.

It's green now.  I'll merge into master.

PR #3290, opened 2021-05-24
Use WellContainer<double> for RS / RV

joakim-hove:


joakim-hove:
jenkins build this please

PR #3295, opened 2021-05-25
Assemble all non-wellcontainer WellState members

joakim-hove:
Minor minor - just assemble all the `WellState`members which are not/will not become `WellContainer`members in one place


joakim-hove:
jenkins build this please

PR #3298, opened 2021-05-25
fixed: do not use Opm:: prefix when inside namespace Opm

akva2:


akva2:
jenkins build this please

joakim-hove:
I'll merge this now. Would it make sense to take this further and move code out from the `ebos/` directory in the repository?

PR #3300, opened 2021-05-25
Added tests for cusparseSolver and openclSolver

Tongdongq:
Copied `test_preconditionerfactory.cpp` and replaced Dune with cusparseSolver and openclSolver.
It needs `getSparsityPattern()` like in the `BdaBridge`, but that is not a public method.
It also creates an empty `WellContributions`.
The values in the `expected` vector are different, but both cusparseSolver and openclSolver converge for NORNE.

### DIFF
--- a/tests/test_cusparseSolver.cpp
+++ b/tests/test_cusparseSolver.cpp
@@ -72,14 +72,18 @@ testCusparseSolver(const boost::property_tree::ptree& prm, const std::string& ma
     const int deviceID = 0;
     const std::string gpu_mode("cusparse");
     const std::string fpga_bitstream("empty");    // unused
+    Dune::InverseOperatorResult result;
 
     Vector x(rhs.size());
     Opm::WellContributions wellContribs("cusparse");
-    Opm::BdaBridge<Matrix, Vector, bz> bridge(gpu_mode, fpga_bitstream, linear_solver_verbosity, maxit, tolerance, platformID, deviceID, opencl_ilu_reorder);
-    Dune::InverseOperatorResult result;
+    try {
+        Opm::BdaBridge<Matrix, Vector, bz> bridge(gpu_mode, fpga_bitstream, linear_solver_verbosity, maxit, tolerance, platformID, deviceID, opencl_ilu_reorder);
 
-    bridge.solve_system(&matrix, rhs, wellContribs, result);
-    bridge.get_result(x);
+        bridge.solve_system(&matrix, rhs, wellContribs, result);
+        bridge.get_result(x);
+    } catch (const std::logic_error& error) {
+        BOOST_WARN_MESSAGE(true, error.what());
+    }
 
     return x;
 }
--- a/tests/test_cusparseSolver.cpp
+++ b/tests/test_cusparseSolver.cpp
@@ -1,4 +1,5 @@
 /*
+  Copyright 2019 SINTEF Digital, Mathematics and Cybernetics.
   Copyright 2021 Equinor
 
   This file is part of the Open Porous Media project (OPM).
--- a/tests/test_cusparseSolver.cpp
+++ b/tests/test_cusparseSolver.cpp
@@ -1,0 +1,127 @@
+/*
+  Copyright 2021 Equinor
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <config.h>
+
+#define BOOST_TEST_MODULE OPM_test_cusparseSolver
+#include <boost/test/unit_test.hpp>
+#include <boost/version.hpp>
+
+#include <dune/common/version.hh>
+
+#if DUNE_VERSION_NEWER(DUNE_ISTL, 2, 6) && \
+    BOOST_VERSION / 100 % 1000 > 48
+
+#include <opm/simulators/linalg/bda/BdaBridge.hpp>
+
+#include <dune/common/fvector.hh>
+#include <dune/istl/bvector.hh>
+#include <dune/istl/bcrsmatrix.hh>
+#include <dune/istl/matrixmarket.hh>
+#include <dune/istl/solver.hh>
+
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/ptree.hpp>
+
+
+template <int bz>
+Dune::BlockVector<Dune::FieldVector<double, bz>>
+testCusparseSolver(const boost::property_tree::ptree& prm, const std::string& matrix_filename, const std::string& rhs_filename)
+{
+    using Matrix = Dune::BCRSMatrix<Dune::FieldMatrix<double, bz, bz>>;
+    using Vector = Dune::BlockVector<Dune::FieldVector<double, bz>>;
+    Matrix matrix;
+    {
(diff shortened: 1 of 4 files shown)


blattms:
Please try to limit code duplication as much as possible. There seem to be at least two identical functions in two or three files now. Maybe you should factor the out to header ans simply include that?

If you are copying stuff, please also state the old copyright holders.

Tongdongq:
@blattms I think it's ready for review

blattms:
Looks great now and compiles on my systems. Just one jenkins run.

blattms:
jenkins build this please

blattms:
Just noted that when ninja Generator is used for CMake, test do not get built on my system. That's odd.
Just checked again and both test_openClSolver and test_cusparseSolver do not compile because of the number of arguments of the constructor of BDA bridge.

Tongdongq:
Since the FPGA PR was merged, BdaBridge takes `fpga_bitstream` as argument. I thought I had fixed that, but perhaps I didn't push it.

blattms:
jenkins build this please

blattms:
Sigh, I guess there is no opencl platform on the jenkins system:
```
Running 1 test case...
%%MatrixMarket
%%MatrixMarket
unknown location(0): [NON-XML-CHAR-0x1B][4;31;49mfatal error: in "TestDefaultPreconditionerFactory": std::logic_error: [/var/lib/jenkins/workspace/opm-simulators-PR-builder@2/opm/simulators/linalg/bda/openclSolverBackend.cpp:182] OpenCL Error: clGetPlatformIDs(-1001)
UNKNOWN_CL_CODE[NON-XML-CHAR-0x1B][0;39;49m
/var/lib/jenkins/workspace/opm-simulators-PR-builder@2/tests/test_openclSolver.cpp(104): [NON-XML-CHAR-0x1B][1;36;49mlast checkpoint: "TestDefaultPreconditionerFactory" test entry[NON-XML-CHAR-0x1B][0;39;49m

[NON-XML-CHAR-0x1B][1;31;49m*** 1 failure is detected in the test module "OPM_test_openclSolver"
[NON-XML-CHAR-0x1B][0;39;49m--------------------------------------------------------------------------
mpiexec detected that one or more processes exited with non-zero status, thus causing

  Process name: [[65344,1],0]
  Exit code:    201
```
I guess you will have to work around this one...

akva2:
maybe there is a cpu platform available on debian? we for sure do not have any gpu..

Tongdongq:
Ah I did not get that.
So maybe a try-catch in the test should suffice? But should the test then fail?

blattms:
Well if the test fails the build will never get green... I think the test should say something like "skipping because no platforms available".

akva2:
i tried with pocl, but atleast on my ubuntu box, the ocl solver is not working properly. it does work on the opencl nv/cuda backend so it's not a general system issue.

Tongdongq:
I'll catch the exception, print a warning and let the test pass.

Tongdongq:
So now I just added
```cpp
try {
} catch (const std::logic_error& error) {
    BOOST_WARN_MESSAGE(true, error.what());
}
```
Any OpenCL errors in openclSolverBackend.cpp are read and rethrown as `std::logic_error`.

Tongdongq:
So apparently the `mpi.openclSolver` test produces `1.79769e+308` (max value for doubles) instead of the expected values. I don't know how to fix this.

blattms:
The beauty of openCL and different platforms. I have noticed this before for platforms other than GPUs (and different GPUs even). Our hunch is that there are some guarantees (in execution order) for GPUs that do not hold on e.g. on the CPU.

These test could be a good start to investigate these issues again.

Tongdongq:
How can I reproduce this error? My `make test_openclSolver` passes, and I don't know how to run `mpi.openclSolver`.

akva2:
```
ctest -R openclSolver
```

the mpi. in front is a prefix for the build configuration on jenkins.

but it won't fail on your system because you have a opencl platform. to reproduce you have to remove all opencl icd's.

blattms:
To reproduce you need to install pocl (pocl-opencl-icd on Debian or http://portablecl.org/) and use the cpu platform (e.g. via adding a command line parameter to the test).

For now we could just query platforms that are known to be GPU's for the test.

Tongdongq:
I had problems with installing pocl on my CentOS machine, and most of its cuda tests failed.
I tried installing it in an Ubuntu container, but pocl does not seem to provide the C++ bindings that are used in openclSolver. I tried to install the NVIDIA drivers but that also failed.
I think I'll just try to query the platform type and only allow GPUs for now.

How is opencl installed on the Jenkins server? Intel? PoCL?

akva2:
there is no icd on the jenkins server.

Tongdongq:
I misunderstood the function of libOpenCL.so, which is found on Jenkins.
So that lib is found, and Flow compiles with HAVE_OPENCL, which enables the test to be run.
The test runs, does not find an icd and should fail.
But instead of triggering the 'no platform found' exception, it produces `1.79769e+308`.
It looks like the Boost test initializes some x vector, and the solver never overwrites it.
Maybe the exception does trigger, but I fail to capture it because it is not a `std::logic_error`?

tests/test_cusparseSolver.cpp
```diff
@@ -0,0 +1,132 @@
...
+    Vector x(rhs.size());
+    Opm::WellContributions wellContribs("cusparse");
+    try {
+        Opm::BdaBridge<Matrix, Vector, bz> bridge(gpu_mode, fpga_bitstream, linear_solver_verbosity, maxit, tolerance, platformID, deviceID, opencl_ilu_reorder);
+
+        bridge.solve_system(&matrix, rhs, wellContribs, result);
+        bridge.get_result(x);
+    } catch (const std::logic_error& error) {
+        BOOST_WARN_MESSAGE(true, error.what());
+    }
```
blattms, line 86: I think this is the wrong place to catch the exception or at least you need to rethrow it in the catch clause. Otherwise or now we return an invalid solution and the caller cannot notice that there was a problem

Tongdongq: But otherwise the test fails.
The test is meant to test the solver, if the solver cannot run, because things are not installed, should it fail?
Perhaps CMake should not add the test if there is no device.

Tongdongq:
Instead of getting devices with `CL_DEVICE_TYPE_ALL`, `CL_DEVICE_TYPE_GPU` can be used instead. But this was replaced to allow for CPU backends (like pocl) as well.
Can we change this just for the test?

blattms:
Isn't the real problem that there are no devices? I think @akva2 was pointing at that.

Attached is my kind of error handling proposal (probably does not compile and is for cusparse, but hopefully clarifies my line of thought.

[0001-Error-handling-proposal.patch.txt](https://github.com/OPM/opm-simulators/files/6662827/0001-Error-handling-proposal.patch.txt)

tests/test_cusparseSolver.cpp
blattms: CMake is probably not the correct place. You might compile on a system without devices but OpenCL and Cusparse installed but run on one that has devices.

blattms:
In it is. Two questions though:
- Don't we need the same error handling for CUDA? What will happen if there are no devices.
- What happens in flow if we request opencl and there are no devices?

Tongdongq:
Shouldn't I have added the same for CUDA before merging then? I think there are less problems with icds from different vendors, but you can still cross-compile on a machine without GPUs.
If no opencl devices are found, an exception should be thrown:
https://github.com/OPM/opm-simulators/blob/2260fde4782f5a9b9e277e63ac8e85eae59aefb2/opm/simulators/linalg/bda/openclSolverBackend.cpp#L93

blattms:
>Shouldn't I have added the same for CUDA before merging then?

Well the rest was good, and all changes risk further delay (on my side). Just do that in another PR.

Concerning the exception:
I am a aware that there will be an exception. What I was wondering is what happens then:

- Simulator thinks there is a linear solver problem and reduces the time step until it gives up.
- Or does it fallback to DUNE.

Will the user see a clear error message?

Cool if you could check by just unconditionally throwing and see what happens.

Tongdongq:
Actually neither happens, the exception is not caught and Flow exists.
The reason: if the user explicitly wants to use opencl, but nothing can execute the opencl, then it should not quietly go to Dune.

PR #3301, opened 2021-05-25
Communicate well rates by directly iterating over container

joakim-hove:
The purpose of this change is to simplify change of container used for wellrates (`std::map<std::string, T> -> WellContainer<T>`) in a subsequent PR.

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -238,7 +238,7 @@ void WellState::init(const std::vector<double>& cellPressures,
     this->global_well_info = std::make_optional<GlobalWellInfo>( schedule, report_step, wells_ecl );
     for (const auto& winfo: parallel_well_info)
     {
-        well_rates.insert({winfo->name(), std::make_pair(winfo->isOwner(), std::vector<double>())});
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
> Why not simply refactor the functors to prevent code duplication?

yes that would have been a possibility - and maybe a better choice. The reason I did it the way I did was that I need to take multiple baby steps in order to keep control; and this way of doing it enabled smaller steps.

blattms:
I think this broke distributed wells. At least I cannot see that rates for wells that are not owned are neglected. This happened in the else branch and you chose to remove the if-else branches for whatever reasons.

blattms:
The else branch disregards the calculated well rates and calls the functor with zero rates.

joakim-hove:
> I think this broke distributed wells.

Are you saying this because a case is now failing - or mainly from code inspection.

> At least I cannot see that rates for wells that are not owned are neglected.

That I disagree - in the second loop there is an `if (owner)` check, and in the case owner is false zero is used explicitly.

PR #3306, opened 2021-05-26
Change well/group controll checking logic. 

totto82:
Remove the switched wells/group logic and instead always check all constrains but from the top.
Also note that I have removed the intermediate communication steps. I.e. All checks are done based on the same condition.


totto82:
jenkins build this please

atgeirr:
I am positive to the change in principle, although every time we change the logic I am nervous... If testing indicates this works well we should merge.

totto82:
Replaced by #3366

PR #3308, opened 2021-05-26
fixed: do not take reference to a temporary

akva2:
getWellsatEnd returns a temporary vector. it goes out of
scope on the line, thus the reference taken is dangling.

@joakim-hove I think this randomly bit and is what broke the post-builder.

### DIFF
--- a/ebos/eclwellmanager.hh
+++ b/ebos/eclwellmanager.hh
@@ -105,9 +105,10 @@ public:
         const Schedule& deckSchedule = simulator_.vanguard().schedule();
         const auto& summaryState = simulator_.vanguard().summaryState();
         // create the wells which intersect with the current process' grid
+        const auto wellsatEnd = deckSchedule.getWellsatEnd();
         for (size_t deckWellIdx = 0; deckWellIdx < deckSchedule.numWells(); ++deckWellIdx)
         {
-            const auto& deckWell = deckSchedule.getWellsatEnd()[deckWellIdx];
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

blattms:
Shouldn't a const reference correctly capture a temporary on the stack and extend its lifetime? Or is that not on the stack?

akva2:
problem is we don't catch the temporary, we only catch an entry inside the temporary (which sits in the heap but that is not the point, the point is the entries dies with the temporary we do not catch).

joakim-hove:
> I think this randomly bit and is what broke the post-builder.

Cool - thanks!

akva2:
(run valgrind on test_ecl_output if u are still not convinced..)

bska:
> we only catch an entry inside the temporary (which sits in the heap but that is not the point, the point is the entries dies with the temporary we do not catch).

Correct.  `const auto& w = wellsAtEnd()[i]` does not do lifetime extension of the well object.  If you really want to extend the lifetime you need to write
```C++
const auto& allwells = wellsAtEnd();
const auto& w = allwells[i];
```
instead.  Another point here is that `wellsAtEnd()` gets expensive if you call it in a loop since it will create a copy of all well objects for each call.

bska (approved):

blattms:
I am convinced if you mean me. Was hoping we would never return 
references to temporaries. But hey, this is real life.

PR #3309, opened 2021-05-26
Make sure all wells are added to the well_rates map

joakim-hove:
`<Disclaimer>`
I might have misunderstood the intended content of the `WellState::well_rates` member
`</Disclaimer>`

The `well_rates` member of the `WellState`object is a map over all wells, where a bool element indicates whether the well is owned by the current process. In current master this is only explicitly initialized for the local wells, and then the structure is subsequently updated (by accident ??) when the rate is assigned with `operator[]` - however the bool part of the `std::pair<bool, std::vector<double>>` is not explicitly assigned this way.

With this PR the `well_rates`member is explicitly assigned for *all* wells, and then the rate update is done with `at( wname )` instead of `operator[](wname)`

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -236,9 +236,13 @@ void WellState::init(const std::vector<double>& cellPressures,
     // call init on base class
     this->base_init(cellPressures, wells_ecl, parallel_well_info, well_perf_data, summary_state);
     this->global_well_info = std::make_optional<GlobalWellInfo>( schedule, report_step, wells_ecl );
+    for (const auto& wname : schedule.wellNames(report_step))
+    {
+        well_rates.insert({wname, std::make_pair(false, std::vector<double>(this->numPhases()))});
+    }
     for (const auto& winfo: parallel_well_info)
     {
-        well_rates.insert({winfo->name(), std::make_pair(winfo->isOwner(), std::vector<double>(this->numPhases()))});
+        well_rates[winfo->name()].first = winfo->isOwner();
     }
 
     const int nw = wells_ecl.size();
--- a/opm/simulators/wells/WellState.hpp
+++ b/opm/simulators/wells/WellState.hpp
@@ -110,8 +110,10 @@ public:
     Well::ProducerCMode currentProductionControl(std::size_t well_index) const { return current_production_controls_[well_index]; }
     void currentProductionControl(std::size_t well_index, Well::ProducerCMode cmode) { current_production_controls_[well_index] = cmode; }
 
-    void setCurrentWellRates(const std::string& wellName, const std::vector<double>& rates ) {
-        well_rates.at(wellName).second = rates;
+    void setCurrentWellRates(const std::string& wellName, const std::vector<double>& new_rates ) {
+        auto& [owner, rates] = this->well_rates.at(wellName);
+        if (owner)
+            rates = new_rates;
     }
 
     const std::vector<double>& currentWellRates(const std::string& wellName) const;
--- a/opm/simulators/wells/WellState.hpp
+++ b/opm/simulators/wells/WellState.hpp
@@ -111,7 +111,7 @@ public:
     void currentProductionControl(std::size_t well_index, Well::ProducerCMode cmode) { current_production_controls_[well_index] = cmode; }
 
     void setCurrentWellRates(const std::string& wellName, const std::vector<double>& rates ) {
-        well_rates[wellName].second = rates;
+        well_rates.at(wellName).second = rates;
     }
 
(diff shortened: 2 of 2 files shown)


joakim-hove:
jenkins build this please

blattms (changes requested):
Yes, I think you misunderstood the code (because of my lousy documentation).

I try to explain this:

WellState should only know wells that are part of the process local domain (i.e.) has perforations in the interior of the DUNE grid. Hence if a well is not part of the process, it should not be in `well_rates`.  While having all the wells there will work, it will result in more computation when running in parallel. That should be prevented (otherwise it will some previous hard work for good scaling have no effect any more).

In the case where a well is part of the interior domain of two or more processes, one process is chosen that owns the well. This is indicated by the `bool`. Only the rates calculated on that process should be taken into account for output and some of the calculations. If you do not do that then the rates of some wells will be added multiple times.

I fear that this might be the case already due to the changes of calculations to the group rates, when wells are distributed. There might be a test missing for groups and distributed wells as we only test one model.

blattms:
The weired thing is:
Looking at the group rate calculation code either we rely on non-local wells being default initialized when accessed or the group communication code was/is buggy. The former is a bit weired, because I am pretty sure that we disregarded local wells historically.

blattms:
I am lost as the code in communicateGroupRates looked quite different before the refactoring. In what of the multiple PRs was that changed?

It seems like the well_rates is also a product from refactoring. Hence I cannot really answer your question. Please reinterpret everything I said as being for `ParallelWellInfo::isOwner`?

joakim-hove:
> Looking at the group rate calculation code either we rely on non-local wells being default initialized when accessed or the group communication code was/is buggy

That is my understanding of the situation; during the `WellGroupHelpers::updateWellRates()` we will call the `WellState::setCurrentWellRate()` for *all wells in the reservoir*. For wells not on the current process this will pass in zero rates and default construct a `std::pair<bool, std::vector<double>>` with the new zero rate.

I have updated this PR with a second commit which just makes sure that the `WellState::setCurrentWellRate()` is *only* called for wells on the current process. If you agree with that commit I will remove the first commit.

joakim-hove:
jenkins build this please

blattms:
I cannot judge this because of the changes unless I invest a considerable amount of time.

What exactly is `well_rates`? It must be something different from before, because it is now used in communicateGroupRates, which previously used group specific vectors:
```
auto forAllGroupData = [&](auto& func) {
                iterateContainer(injection_group_rein_rates, func);
                iterateContainer(production_group_reduction_rates, func);
                iterateContainer(injection_group_reduction_rates, func);
                iterateContainer(injection_group_reservoir_rates, func);
                iterateContainer(production_group_rates, func);
                iterateRatesContainer(well_rates, func);
            };
.....
```
There is probably some magic elsewhere that I am missing....

joakim-hove:
> I cannot judge this because of the changes unless I invest a considerable amount of time.

OK; then I think I will stick to the original - because I think was correct and actually needed. I *think* communicateGroupRates* needs *all the wells in the reservoir* - and that has previously worked because entries for wells on remote processors have (by accident ??) been default constructed when calling `WellState::setCurrentRates()`

> What exactly is well_rates?

`well_rates` is unchangeḑ, the change in `communicateGroupRates` is that *all the other* rates are communicated here: https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/GroupState.hpp#L80

joakim-hove:
jenkins build this please

atgeirr:
I believe the well_rates member was added to provide global well rates, so that we could do identical group calculations based on those rates on all processes. (I hope that cleared up more than it confused?)

joakim-hove:
> I believe the well_rates member was added to provide global well rates, so that we could do identical group calculations based on those rates on all processes

Yes I agree - and it is my claim that this has worked only because `std::map<,>::operator[]` default constructs for the wells on other processors - this has been made more explicit in this PR; I have discussed this offline with @blattms and have the impression that he agrees on that point.

bska:
> because `std::map<,>::operator[]` default constructs for the wells on other processors

True.  There's one tangential point that I'd like to make in this context.  `map<K,V>::operator[]()` **always** creates a new map element (i.e., an internal map "node" containing a `std::pair<const K, V>`), even if an element with the same key already exists.  It is just that this new element is discarded if an element with the same key already exists in the map.  Otherwise, the new node is linked into the internal tree-like structure.  This behaviour means we should be wary of `map<K,V>::operator[]()` in performance sensitive code.

joakim-hove:
@blattms : You requested changes on this PR, but after offline discussion I got the impression you were happy with this after all?

joakim-hove:
I'll merge this later today.

PR #3296, opened 2021-05-25
Segment use well index

joakim-hove:

PR #3311, opened 2021-05-27
Bug in wellstate handling for pressure drop in MSW wells?

joakim-hove:
I *think* there is a bug in the handling of the pressure drop terms in the `WellState` object for MSW wells. Typical code looks like this:
```C++
for (int seg=0; seg < numSegments(); seg++) {
     ...
     well_state.segPressDrop()[seg] = .....
}
```
the problem with this code is that the `seg` variable only runs over the number of segments for *this particular well*, whereas the `WellState::segment_pressure_drop_` vector expects an index in the range `[0, total_number_of_segments_in_reservoir>`. With this PR the segment index is offset by the index location of the top segment before the `WellState` array is accessed.

The code in question is only used for summary output, and should therefor not affect the calculations as such. I have stumbled upon this while working with #3296 - so it is not reported as a bug by anyone. I hope @GitPaean will have a look, and maybe also this can be tested with Eclipse. In case this indeed addresses a bug there will be updates to the reference solutions.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1921,19 +1921,20 @@ namespace Opm
 
         // for top segment, the well control equation will be used.
         EvalWell pressure_equation = getSegmentPressure(seg);
+        auto top_segment_index = well_state.topSegmentIndex(this->index_of_well_);
 
         // we need to handle the pressure difference between the two segments
         // we only consider the hydrostatic pressure loss first
         // TODO: we might be able to add member variables to store these values, then we update well state
         // after converged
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

GitPaean:
okay. I will have a look later today.

GitPaean:
This is a real fix. Let me double check the results then we should update the reference data.

GitPaean:

GitPaean (approved):

bska:
jenkins build this opm-tests=542 please

bska:
jenkins build this opm-tests=542 please

bska:
jenkins build this please

bska:
Merging this now as the updated reference solutions have been installed on the CI system.

PR #3316, opened 2021-05-28
Segment use well index2

joakim-hove:
Use well index for segment output variables.


joakim-hove:
jenkins build this please

akva2:
merged with a nasty warning...

PR #3318, opened 2021-05-28
changed: do not set the ebos well model as default type

akva2:
rather, only set it where we want to use it. this avoids including
eclwellmanager.hh and eclpeacemanwell.hh unnecessarily in
simulator objects (where BlackoilWellModel is used).


joakim-hove:
It is green, it is in a direction I am strongly in favor of, and it has lingered for some time. I'll merge!

PR #3319, opened 2021-05-28
changed: remove unused functions from VFPHelpers

akva2:


akva2:
jenkins build this please

bska:
jenkins build this please

akva2:
jenkins build this please

opm/simulators/wells/VFPHelpers.cpp
```diff
@@ -0,0 +1,656 @@
...
+
+/**
+ * Helper function that finds x for a given value of y for a line
+ * *NOTE ORDER OF ARGUMENTS*
+ */
+double findX(const double& x0,
+             const double& x1,
+             const double& y0,
+             const double& y1,
+             const double& y)
```
bska, line 44: Any reason why the input parameters are reference to `double` instead of plain `double`?

akva2: i wouldn't know, but i can't imagine there is a reason (i just shuffled the code around).

opm/simulators/wells/VFPHelpers.hpp
```diff
@@ -560,284 +167,16 @@ bool hasTable(const std::map<int, std::reference_wrapper<const T>> tables, int t
...
-    if (x >= std::min(x1,x2) && x <= std::max(x1,x2)) {
-        bhp = y;
-        return true;
-    } else {
-        return false;
-    }
-}
-
-// calculating the BHP from thp through the intersection of VFP curves and inflow performance relationship
-inline bool findIntersectionForBhp(const std::vector<RateBhpPair>& ratebhp_samples,
```
GitPaean, line 792: Just a note, this is the old implementation regarding to how to calculate the BHP from the intersection of the IPR and VFP curves, which is the standard way in textbook and manual. OPM is going to a different direct now, and looks like working.  In the future, we might come back to revisit this method with more information. It is not used now and okay to be removed for now.

opm/simulators/wells/VFPHelpers.cpp
GitPaean: There are multiple places used `const double&`, since you are here, maybe it is good to stop them from circulating.

GitPaean (approved):
It looks good to me.

opm/simulators/wells/VFPHelpers.cpp
akva2: done.

akva2:
jenkins build this please

bska:
I'll merge this into master now.

PR #3297, opened 2021-05-25
Add class SegmentState

joakim-hove:
Add small class to hold the segment state for one well.

PR #3323, opened 2021-05-30
Remove segment pressure drop as separate variable

joakim-hove:
The segment pressure drop is the sum of three terms - the pressure drop due to friction, acceleration and hydrostatic change. With this PR the segment pressure drop is implemented as a function instead of a seperate member.

This is an intermediate step on the way towards: #3297.


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

PR #3329, opened 2021-05-31
Segment state1

joakim-hove:
Use segment state for output pressure and rates. Final step before: #3297


joakim-hove:
jenkins build this please

joakim-hove:
Merged as #3335

PR #3333, opened 2021-06-01
Pass the temperature to the restart machinary if enableTemperature is true

totto82:
EnableTemperature is true as default, i.e. the temperature is now passed to the restart machinery as default.
It will however be ignored and not written to the restart file if the user does not specify --enable-opm-restart-file=true

### DIFF
--- a/ebos/eclgenericoutputblackoilmodule.cc
+++ b/ebos/eclgenericoutputblackoilmodule.cc
@@ -69,6 +69,7 @@ EclGenericOutputBlackoilModule(const EclipseState& eclState,
                            const SummaryConfig& summaryConfig,
                            const SummaryState& summaryState,
                            bool enableEnergy,
+                           bool enableTemperature,
                            bool enableSolvent,
                            bool enablePolymer,
                            bool enableFoam,
@@ -79,6 +80,7 @@ EclGenericOutputBlackoilModule(const EclipseState& eclState,
     , summaryConfig_(summaryConfig)
     , summaryState_(summaryState)
     , enableEnergy_(enableEnergy)
+    , enableTemperature_(enableTemperature)
     , enableSolvent_(enableSolvent)
     , enablePolymer_(enablePolymer)
     , enableFoam_(enableFoam)
@@ -554,7 +556,14 @@ assignToSolution(data::Solution& sol)
     }
 
     if (!temperature_.empty()) {
-        sol.insert("TEMP", UnitSystem::measure::temperature, std::move(temperature_), data::TargetType::RESTART_SOLUTION);
+        if (enableEnergy_)
+            sol.insert("TEMP", UnitSystem::measure::temperature, std::move(temperature_), data::TargetType::RESTART_SOLUTION);
+        else {
+            // Flow allows for initializing of non-constant initial temperature.
+            // For output of this temperature for visualization and restart set --enable-opm-restart=true
+            assert(enableTemperature_);
+            sol.insert("TEMP", UnitSystem::measure::temperature, std::move(temperature_), data::TargetType::RESTART_AUXILIARY);
+        }
     }
 
     if (FluidSystem::phaseIsActive(waterPhaseIdx) && !saturation_[waterPhaseIdx].empty()) {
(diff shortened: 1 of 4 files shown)


bska:
> [Temperature] will however be ignored and not written to the restart file if the user does not specify `--enable-opm-restart-file=true`

That's not entirely true in the current situation.  The [`TEMP`](https://github.com/OPM/opm-simulators/blob/fc47eb8b4f76e1b32eb2609173756b91438a7623/ebos/eclgenericoutputblackoilmodule.cc#L557) output vector is tagged as [`RESTART_SOLUTION`](https://github.com/OPM/opm-common/blob/e01e8e92251faf11f905f3c388ae32bf63490273/opm/output/data/Cells.hpp#L63) and those vectors are always [output to the restart file](https://github.com/OPM/opm-common/blob/e01e8e92251faf11f905f3c388ae32bf63490273/src/opm/output/eclipse/RestartIO.cpp#L574-L578) regardless of the `--enable-opm-rst-file` setting.

When creating ECLIPSE compatible output files (`--enable-opm-rst-file=false`) we should ensure that we do not output `TEMP` unless the run specifies the temperature or the thermal option (keyword `TEMP` or keyword `THERMAL` in `RUNSPEC`).

totto82:
I tested this PR on CO2STORE. For --enable-opm-rst-file=true. TEMP appears in Resinsight but if i use default i.e. --enable-opm-rst-file=false it does not. 
I also made a restart version of CO2STORE. If i first run it with --enable-opm-rst-file=true and then I can restart from it. But if I first run it with --enable-opm-rst-file=false I cannot restart it i.e. it will complain that temperature is missing.

totto82:
jenkins build this please

bska:
> I tested this PR on CO2STORE. For `--enable-opm-rst-file=true`. TEMP appears in ResInsight but if i use default i.e. `--enable-opm-rst-file=false` it does not.

Then there's some other intervening facility somewhere that prevents `TEMP` from reaching the I/O layer.  [`RestartIO::save()`](https://github.com/OPM/opm-common/blob/e01e8e92251faf11f905f3c388ae32bf63490273/src/opm/output/eclipse/RestartIO.cpp#L759) treats `RESTART_SOLUTION` as an unconditional request to output the corresponding vector to the restart file.

bska:
>>  tested this PR on CO2STORE. For `--enable-opm-rst-file=true`. `TEMP` appears in ResInsight but if i use default i.e. `--enable-opm-rst-file=false` it does not.
>
> Then there's some other intervening facility somewhere that prevents TEMP from reaching the I/O layer. `RestartIO::save()` treats `RESTART_SOLUTION` as an unconditional request to output the corresponding vector to the restart file.

No, there's not.  I tested this PR on [0_BASE_MODEL2](https://github.com/OPM/opm-tests/blob/a9ea3a92b1a30e6df2dd490912da0f2d59884a96/model2/0_BASE_MODEL2.DATA) from opm-tests and got `TEMP` output for both `--enable-opm-rst-file=true` and `--enable-opm-rst-file=false`.
```sh
cse=path/to/opm-tests/model2/0_BASE_MODEL2.DATA
$ ${flow} --ecl-deck-file-name=${cse} --enable-opm-rst-file=false --output-dir=./false
$ ${flow} --ecl-deck-file-name=${cse} --enable-opm-rst-file=true --output-dir=./true
```

---

```
$ /home/bska/work/opm/inst/opt/bin/kw_list.x ./false/0_BASE_MODEL2.UNRST |grep '^TEMP'
TEMP             2794:REAL 
TEMP             2794:REAL 
[...]
TEMP             2794:REAL 

$ /home/bska/work/opm/inst/opt/bin/kw_list.x ./true/0_BASE_MODEL2.UNRST |grep '^TEMP'
TEMP             2794:REAL 
TEMP             2794:REAL 
[...]
TEMP             2794:REAL 
```

totto82:
I guess all the test failures proves your point. Could a solution where temperature is passed as RESTART_SOLUTION iff enableThermal (THERMAL or TEMP in RUNSPEC) and as RESTART_AUXILIARY if not. I.e. only enableTemperature is true. We need to be careful not to remove the capability of ecl compatible restart with THERMAL.

bska:
> Could a solution where temperature is passed as RESTART_SOLUTION iff `enableThermal` (THERMAL or TEMP in RUNSPEC) and as RESTART_AUXILIARY if not.

Yes, that'd have the proper semantics.

totto82:
jenkins build this please

bska:
Thanks for the updates.  This looks better than the first edition.  I'll merge into master.

PR #3337, opened 2021-06-02
Don't Create Well Instance for Wells With No Connections

bska:
This PR protects against `Opm::Well` instances for which the sets of connections are empty.  In those cases, do not put entries in the well container as there is no influence on the systems of non-linear equations.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -587,6 +587,12 @@ namespace Opm {
 
             for (int w = 0; w < nw; ++w) {
                 const Well& well_ecl = wells_ecl_[w];
+
+                if (well_ecl.getConnections().empty()) {
+                    // No connections in this well.  Nothing to do.
+                    continue;
+                }
+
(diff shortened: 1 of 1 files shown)


bska:
@totto82: Would you mind testing if this PR fixes the `WELPI` hang on your model?  The underlying issue should arguably be handled in opm-common.  We noticed a possibly related symptom in OPM/opm-common#2442.

totto82:
I think this PR make sense and should be merged even though it didn't fix the issue with WELPI+Smeaheia+MPI.

bska:
> I think this PR make sense and should be merged even though it didn't fix the issue with WELPI+Smeaheia+MPI.

Yes, I kind of think it makes sense too.  I'll mark it as "ready for review" and launch a build test.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

totto82:
Please merge when green.

bska:
> Please merge when green.

Cool, thanks.  It's green now so I'll merge.

PR #3338, opened 2021-06-02
Use perfrates.size() instead of storing num_perf explicitly

joakim-hove:


joakim-hove:
jenkins build this please

PR #3336, opened 2021-06-01
Manage well potentials with WellContainer<>

joakim-hove:
Manage  the Well potential using a WellContainer<>

PR #3341, opened 2021-06-02
Explicitly zero well potentials at the start of report step

joakim-hove:
This is more of a question than an actual PR. In current master the well potentials can carry over to the next report step; the logic is quite complex with respect to previous step closed or not .... I have not gone through it in all detail.

Point is that with: #3336 the wellpotensials are explicitly cleared at the start of the report step (and then fetched from previous if there is a relevant previous!). If this PR is correct there will be a data update.

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -337,17 +337,13 @@ void WellState::init(const std::vector<double>& cellPressures,
         }
     }
 
-    perfRateSolvent_.clear();
-    perfRateSolvent_.resize(nperf, 0.0);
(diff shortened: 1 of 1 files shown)


bska:
> In current master the well potentials can carry over to the next report step

That's necessary.  See PR #3182.

joakim-hove:
jenkins build this please

bska:
Having reviewed the context more thoroughly I agree that this fixes a programming error.  I'll update the reference solutions and merge this.

bska:

bska:
New reference solutions have been installed on the CI system.  Merging this now to bring the overall system back to a consistent state.

PR #3346, opened 2021-06-03
Amgcl support

Tongdongq:
This PR allows to use the [amgcl library](https://github.com/ddemidov/amgcl) in Flow.
The option --matrix-add-well-contributions=true is required.
Different amgcl parameters (e.g. choosing preconditioners or solvers) can be done in amgclSolverBackend.cpp at compile time.
Some parameters are passed via Flow (linear-solver-reduction, linear-solver-max-iter, flow-linear-solver-verbosity).
amgcl uses the builtin backend by default, which runs on CPU. It can also use CUDA as backend, if found by CMake.
In that case the macro AMGCL_CUDA must be set to 1 in amgclSolverBackend.cpp:solve_system().
A different preconditioner or iterative solver can be chosen in solve_system() as well.
I've done some small NORNE tests:
Generally working: ilu0, ilu1, damped_jacobi, gauss_seidel, bicgstab, gmres, idrs
Generally not working: amg, spai0, chebyshev, dummy, richardson

Notes:
The BdaBridge has `use_gpu` and `use_fpga` to determine if it should use a backend or not, amgcl is neither.
To let amgcl use the CUDA backend, `set_source_files_properties(opm/simulators/linalg/bda/amgclSolverBackend.cpp PROPERTIES LANGUAGE CUDA)` in CMakeLists.txt is used to mark the file as a CUDA source, instead of a C++ source.
It might be possible to rename the file to .cu, but I'm not sure if a non-CUDA compiler can interpret this.
Using ViennaCL without amgcl converged Flow for NORNE, but with 117k linear iterations
Using vexcl(OpenCL) with amgcl converged Flow for NORNE, but with 155k linear iterations.
Using vexcl(CUDA) with amgcl converged Flow for NORNE, but with 154k linear iterations.
ViennaCL and vexcl did not have satisfying convergence/performance so they were left out, but implementations are available.

Benchmarks:
```
Notation: A (B+C+D+E), F, G , H
A: Total time
B: Assembly time
C: Linear solve time
D: Update time
E: Pre/post step
F: Overall Linearizations
G: Overall Newton Iterations
H: Overall Linear Iterations
CPU: unknown 16-core Haswell, 2.3GHz, GPU: NVIDIA V100:
Dune:                            708 (164+393+ 84+55), 1866, 1501, 22131
amgcl + cuda:                   1090 (166+775+ 83+53), 1868, 1504, 22204
cusparseSolver:                  447 (165+133+ 83+53), 1846, 1485, 21167
openclSolver (graph_coloring):   612 (210+223+108+58), 2371, 1984, 45230
openclSolver (level_scheduling): 637 (170+313+ 85+56), 1824, 1462, 22133
CPU: Intel Xeon E5-2620 v3 @ 2.4GHz, GPU: NVIDIA 2080Ti:
Dune:                            816 (322+413+111+65), 1866, 1501, 22131
amgcl + cuda:                    695 (207+292+116+65), 1866, 1501, 22333
cusparseSolver:                  543 (205+155+106+62), 1845, 1485, 21132
openclSolver (graph_coloring):   725 (267+229+142+71), 2335, 1943, 43688
CPU: AMD Ryzen 5 2400G, NVIDIA GTX 1050Ti:
Dune:                            414 (120+194+59+34), 1866, 1501, 22131
amgcl + cuda:                    968 (138+720+66+36), 1835, 1474, 22190
amgcl + vexcl:                  1420 (155+1154+76+27), 2042, 1664, 153288 | 329 fallbacks to Dune
```
```
Detail, amgcl + cuda, NVDIA 2080Ti, accumulated time (s):
solve_system: 225.2
get_result:     0.3
convert_data:   7.4
copy_rhs:       0.4
solve_system here also includes the ilu decomposition inside amgcl
```

Example `amgcl_options.json`:
```json
{
    "backend_type": "cuda",
    "precond": {
        "class": "relaxation",
	"type": "ilu0",
        "damping": "0.9"
    },
    "solver": {
        "type": "bicgstab",
        "maxiter": "200",
        "tol": "1e-2",
        "verbose": "true"
    }
}
```


atgeirr:
I welcome the ability to test with amgcl. I think we may need to refactor this a little though (at least if we want to use it in the long term).
 - Copying the linear system reduces performance for no reason. We already have a system for dealing with differences in how to build matrices and vectors, the "sparse matrix adapters", of which IstlSparseMatrixAdapter (see opm-models directory opm/simulators/linalg) is the one used to work with ISTL matrices. We could make a new one that directly builds the structures required by agmcl. (This was already done for PETSc, but I do not know if that code made it into the repo).
 - Amgcl supports flexible options at runtime, via JSON files. This can also be easily supported by us, instead of hardcoding options in the cpp file.
 - The options for linear solvers should generally be treated at the same level, rather than some variants being in a separate, lower-level subsystem like the BdaBridge (what does that mean BTW?).

I have not yet looked much at the code here, but I assume it is ok for testing. My concerns are more about long-term maintainability and user (+ developer) friendliness.

Tongdongq:
I agree it should be refactored for the long term, however the results are not very impressive right now.
The closed https://github.com/OPM/opm-simulators/pull/1466 should be a good example.
We (the BDA company) have always worked from ISTLSolverEbos.hpp and down, since that is where the call to the Dune bicgstab solver was made, which we where meant to replace with cusparseSolver.
The BdaBridge is a way to convert the `Dune::BCRSMatrix` to the primitive arrays that accelerators require, although it also 'converts' with `static_cast<double*>(&(((*mat)[0][0][0][0])))`. The Bridge is also responsible for choosing different acceleratorSolvers.
I'll look at the JSON options

atgeirr:
> The BdaBridge is a way to convert the `Dune::BCRSMatrix` to the primitive arrays that accelerators require

I think this could be built directly using a new "sparse matrix adapter" but maybe it is not worth it (yet).

blattms:
Why is the CPU unknown? Shouldn't it be possible to read the configuration from the system?

blattms:
Please also add the CPU runtime for Intel Xeon E5-2620 v3. I assume these are all serial runs?

Tongdongq:
I added the CPU run results and some more details on the unknown CPU.

CMakeLists.txt
```diff
@@ -213,6 +213,18 @@ if(OpenCL_FOUND)
...
+find_file(AMG_HPP amgcl/amg.hpp HINTS ${AMGCL_INCLUDE_DIRS})
+if(AMG_HPP)
+  set(HAVE_AMGCL 1)
+  set(AMGCL_FOUND 1)
+  include_directories(${AMGCL_INCLUDE_DIRS})
+  set_source_files_properties(opm/simulators/linalg/bda/amgclSolverBackend.cpp PROPERTIES LANGUAGE CUDA)
+else()
+  set(HAVE_AMGCL 0)
+  set(AMGCL_FOUND 0)
+endif()
```
blattms, line 226: It seems like amgcl install package configuration scripts [here](https://github.com/ddemidov/amgcl/blob/c781719e2f961c161a7ff0cf2c168ac6949ac510/CMakeLists.txt#L309-L311), shouldn't we just do
```
find_package(amgcl)
```
that should set `amgcl_FOUND`

opm/simulators/linalg/bda/BdaBridge.cpp
```diff
@@ -197,40 +205,30 @@ void BdaBridge<BridgeMatrix, BridgeVector, block_size>::solve_system(BridgeMatri
...
+            OpmLog::warning("BdaSolver only accepts blocksize = 3 at this time, will use Dune for the remainder of the program");
+            use_gpu = use_fpga = false;
             return;
         }
 
         if (h_rows.capacity() == 0) {
             h_rows.reserve(Nb+1);
             h_cols.reserve(nnzb);
-#if PRINT_TIMERS_BRIDGE
-            Dune::Timer t;
```
blattms, line 209: Why is the timer removed?

blattms (commented):
Premature review.

Will need to take a look at the Backend stuff somwhen else.

opm/simulators/linalg/bda/BdaBridge.cpp
Tongdongq: This function was called only once and takes like 1 ms, I expect this timing info to be less important.

CMakeLists.txt
Tongdongq: f0cd5cb

blattms:
I am currently testing this.
Did you test the last change in a clean build?

I am experiencing compile errors:
```
In file included from /home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.cpp:28:0:
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.hpp:31:10: fatal error: amgcl/amg.hpp: Datei oder Verzeichnis nicht gefunden
 #include <amgcl/amg.hpp>
```
with an installed amgcl. I guess you need to use the targets for linking, as it seems that AMGCL_INCLUDE_DIRS might not be set.

blattms:
Looking at amgcl-targets.cmake I am doubting that it can handle viennacl and vexcl into different locations...

Tongdongq:
I did. I also tested with `find_package(amgcl HINTS ${AMGCL_INCLUDE_DIRS})`. I'll look again tomorrow.
For viennacl and vexcl, I added more `include_directories(${LIB_INCLUDE_DIRS})` statements in there, with passing them during the `cmake ..`.

blattms:
I think there is a misunderstanding.
To tell cmake where to find amgcl one uses amcl_DIR that points to the directory where AmgclConfig.cmake is. There is simply no variable AMGCL_INCLUDE_DIRS and it cannot be used.

See  /Tongdongq/opm-simulators/pull/3. Unfortunatly, just using the `amgcl::amgcl` for linking fails because nvcc 9.2. does not know `-fopenmp` [posted a bug at amgc](https://github.com/ddemidov/amgcl/issues/202)

CMakeLists.txt
```diff
@@ -213,6 +213,13 @@ if(OpenCL_FOUND)
   endif()
 endif()
 
+find_package(amgcl)
+if(amgcl_FOUND)
+  set(HAVE_AMGCL 1)
+  include_directories(${AMGCL_INCLUDE_DIRS})
+  set_source_files_properties(opm/simulators/linalg/bda/amgclSolverBackend.cpp PROPERTIES LANGUAGE CUDA)
```
blattms, line 220: How is this working without CUDA? 
I am getting this error when activating only OpenCL, maybe that is related?
```
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_CUDA_COMPILE_WHOLE_COMPILATION
```

blattms:
Will come back to testing this next week. Have a nice weekend.

Tongdongq:
I have difficulties putting back my vexcl implementation.
Could you add vexcl to the include dirs and `#include <amgcl/backend/vexcl.hpp>` to `amgclSolverBackend.hpp`?
I get some errors involving `boost::proto::result_of::as_child<>` in `vexcl/vexcl/operations.hpp`

blattms:
Sorry, but this still doesn't compile without cuda (`cmake -DCMAKE_DISABLE_FIND_PACKAGE_CUDA=ON`) for me.
```
In file included from /home/mblatt/opt/amgcl/include/amgcl/relaxation/runtime.hpp:50:0,
                 from /home/mblatt/opt/amgcl/include/amgcl/preconditioner/runtime.hpp:43,
                 from /home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.hpp:38,
                 from /home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/BdaBridge.cpp:42:
/home/mblatt/opt/amgcl/include/amgcl/relaxation/chebyshev.hpp: In constructor ‘amgcl::relaxation::chebyshev<Backend>::chebyshev(const Matrix&, const amgcl::relaxation::chebyshev<Backend>::params&, const typename Backend::params&)’:
/home/mblatt/opt/amgcl/include/amgcl/relaxation/chebyshev.hpp:117:15: warning: declaration of ‘prm’ shadows a member of ‘amgcl::relaxation::chebyshev<Backend>’ [-Wshadow]
             ) : prm(prm),
               ^
/home/mblatt/opt/amgcl/include/amgcl/relaxation/chebyshev.hpp:110:11: note: shadowed declaration is here
         } prm;
           ^~~
In file included from /home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/BdaBridge.cpp:42:0:
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.hpp: At global scope:
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.hpp:67:29: error: ‘cuda’ in namespace ‘amgcl::backend’ does not name a template type
     typedef amgcl::backend::cuda<double> CUDA_Backend;
                             ^~~~
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.hpp:72:63: error: ‘CUDA_Backend’ was not declared in this scope
     typedef amgcl::make_solver<amgcl::runtime::preconditioner<CUDA_Backend>, amgcl::runtime::solver::wrapper<CUDA_Backend> > CUDA_Solver;
                                                               ^~~~~~~~~~~~
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.hpp:72:63: note: suggested alternative: ‘CPU_Backend’
     typedef amgcl::make_solver<amgcl::runtime::preconditioner<CUDA_Backend>, amgcl::runtime::solver::wrapper<CUDA_Backend> > CUDA_Solver;
                                                               ^~~~~~~~~~~~
                                                               CPU_Backend
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.hpp:72:75: error: template argument 1 is invalid
     typedef amgcl::make_solver<amgcl::runtime::preconditioner<CUDA_Backend>, amgcl::runtime::solver::wrapper<CUDA_Backend> > CUDA_Solver;
                                                                           ^
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.hpp:72:110: error: ‘CUDA_Backend’ was not declared in this scope
     typedef amgcl::make_solver<amgcl::runtime::preconditioner<CUDA_Backend>, amgcl::runtime::solver::wrapper<CUDA_Backend> > CUDA_Solver;
                                                                                                              ^~~~~~~~~~~~
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.hpp:72:110: note: suggested alternative: ‘CPU_Backend’
     typedef amgcl::make_solver<amgcl::runtime::preconditioner<CUDA_Backend>, amgcl::runtime::solver::wrapper<CUDA_Backend> > CUDA_Solver;
                                                                                                              ^~~~~~~~~~~~
                                                                                                              CPU_Backend
....
```

 Please always make sure that the PRs work without CUDA and without opendl (`cmake -DCMAKE_DISABLE_FIND_PACKAGE_OpenCL=ON`).
There are also quite some warnings about shadowing variables, etc (at least when using OpenCL only) that should be resolved.

Tongdongq:
I forgot to add it apparently. The warnings about shadowed variables are coming from amgcl itself.

blattms:
Indeed they are from amgcl. Warnings addressed in Tongdongq/opm-simulators#4

blattms:
Since I got hit by this a few times:
It would be great, if we would stop doing
```
    set_source_files_properties(opm/simulators/linalg/bda/amgclSolverBackend.cpp PROPERTIES LANGUAGE CUDA)
```
and instead factor out the cuda part to a proper *.cu file. There even do not seem to be any kernels in there, which makes we wonder why nvcc would be needed at all here. What am I missing?

Tongdongq:
I get a bunch of `/usr/local/cuda-11.1/include/thrust/detail/static_assert.h:51:44: error: static assertion failed: unimplemented for this system` errors when not having it compiled by nvcc.

blattms:
Sorry, not enough information... Please give us information to judge this without replicating your particular system/software version
(what is the assertion, stack trace, etc...)

Tongdongq:
```
In file included from /usr/local/cuda-11.1/include/thrust/system/detail/generic/for_each.h:27:0,
                 from /usr/local/cuda-11.1/include/thrust/detail/for_each.inl:26,
                 from /usr/local/cuda-11.1/include/thrust/for_each.h:279,
                 from /usr/local/cuda-11.1/include/thrust/system/detail/generic/transform.inl:19,
                 from /usr/local/cuda-11.1/include/thrust/system/detail/generic/transform.h:105,
                 from /usr/local/cuda-11.1/include/thrust/detail/transform.inl:25,
                 from /usr/local/cuda-11.1/include/thrust/transform.h:724,
                 from /usr/local/cuda-11.1/include/thrust/system/detail/generic/copy.inl:23,
                 from /usr/local/cuda-11.1/include/thrust/system/detail/generic/copy.h:58,
                 from /usr/local/cuda-11.1/include/thrust/detail/copy.inl:21,
                 from /usr/local/cuda-11.1/include/thrust/detail/copy.h:90,
                 from /usr/local/cuda-11.1/include/thrust/detail/allocator/copy_construct_range.inl:21,
                 from /usr/local/cuda-11.1/include/thrust/detail/allocator/copy_construct_range.h:46,
                 from /usr/local/cuda-11.1/include/thrust/detail/contiguous_storage.inl:23,
                 from /usr/local/cuda-11.1/include/thrust/detail/contiguous_storage.h:235,
                 from /usr/local/cuda-11.1/include/thrust/detail/vector_base.h:30,
                 from /usr/local/cuda-11.1/include/thrust/device_vector.h:25,
                 from /home/tqiu/data/opm/pr3300/prerequisites/amgcl/amgcl/backend/cuda.hpp:41,
                 from /home/tqiu/data/opm/pr3300/opm/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.hpp:42,
                 from /home/tqiu/data/opm/pr3300/opm/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.cpp:28:
/usr/local/cuda-11.1/include/thrust/system/detail/generic/for_each.h: In instantiation of ‘InputIterator thrust::system::detail::generic::for_each(thrust::execution_policy<Derived>&, InputIterator, InputIterator, UnaryFunction) [with DerivedPolicy = thrust::cuda_cub::tag; InputIterator = thrust::zip_iterator<thrust::tuple<thrust::detail::normal_iterator<thrust::device_ptr<const double> >, thrust::detail::normal_iterator<thrust::device_ptr<const double> >, thrust::detail::normal_iterator<thrust::device_ptr<double> >, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type> >; UnaryFunction = amgcl::backend::axpbypcz_impl<double, thrust::device_vector<double, thrust::device_allocator<double> >, double, thrust::device_vector<double, thrust::device_allocator<double> >, double, thrust::device_vector<double, thrust::device_allocator<double> >, void>::functor]’:
/usr/local/cuda-11.1/include/thrust/detail/for_each.inl:44:18:   required from ‘InputIterator thrust::for_each(const thrust::detail::execution_policy_base<DerivedPolicy>&, InputIterator, InputIterator, UnaryFunction) [with DerivedPolicy = thrust::cuda_cub::tag; InputIterator = thrust::zip_iterator<thrust::tuple<thrust::detail::normal_iterator<thrust::device_ptr<const double> >, thrust::detail::normal_iterator<thrust::device_ptr<const double> >, thrust::detail::normal_iterator<thrust::device_ptr<double> >, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type> >; UnaryFunction = amgcl::backend::axpbypcz_impl<double, thrust::device_vector<double, thrust::device_allocator<double> >, double, thrust::device_vector<double, thrust::device_allocator<double> >, double, thrust::device_vector<double, thrust::device_allocator<double> >, void>::functor]’
/usr/local/cuda-11.1/include/thrust/detail/for_each.inl:58:26:   required from ‘InputIterator thrust::for_each(InputIterator, InputIterator, UnaryFunction) [with InputIterator = thrust::zip_iterator<thrust::tuple<thrust::detail::normal_iterator<thrust::device_ptr<const double> >, thrust::detail::normal_iterator<thrust::device_ptr<const double> >, thrust::detail::normal_iterator<thrust::device_ptr<double> >, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type> >; UnaryFunction = amgcl::backend::axpbypcz_impl<double, thrust::device_vector<double, thrust::device_allocator<double> >, double, thrust::device_vector<double, thrust::device_allocator<double> >, double, thrust::device_vector<double, thrust::device_allocator<double> >, void>::functor]’
/home/tqiu/data/opm/pr3300/prerequisites/amgcl/amgcl/backend/cuda.hpp:659:25:   required from ‘static void amgcl::backend::axpbypcz_impl<A, thrust::device_vector<V>, B, thrust::device_vector<V>, C, thrust::device_vector<V> >::apply(A, const vector&, B, const vector&, C, amgcl::backend::axpbypcz_impl<A, thrust::device_vector<V>, B, thrust::device_vector<V>, C, thrust::device_vector<V> >::vector&) [with A = double; B = double; C = double; V = double; amgcl::backend::axpbypcz_impl<A, thrust::device_vector<V>, B, thrust::device_vector<V>, C, thrust::device_vector<V> >::vector = thrust::device_vector<double, thrust::device_allocator<double> >]’
/home/tqiu/data/opm/pr3300/prerequisites/amgcl/amgcl/backend/interface.hpp:379:61:   required from ‘void amgcl::backend::axpbypcz(A, const Vector1&, B, const Vector2&, C, Vector3&) [with A = double; Vector1 = thrust::device_vector<double, thrust::device_allocator<double> >; B = double; Vector2 = thrust::device_vector<double, thrust::device_allocator<double> >; C = double; Vector3 = thrust::device_vector<double, thrust::device_allocator<double> >]’
/home/tqiu/data/opm/pr3300/prerequisites/amgcl/amgcl/solver/bicgstab.hpp:206:38:   required from ‘std::tuple<long unsigned int, typename amgcl::math::scalar_of<typename Backend::value_type, void>::type> amgcl::solver::bicgstab<Backend, InnerProduct>::operator()(const Matrix&, const Precond&, const Vec1&, Vec2&&) const [with Matrix = amgcl::backend::cuda_matrix<double>; Precond = amgcl::runtime::preconditioner<amgcl::backend::cuda<double> >; Vec1 = thrust::device_vector<double, thrust::device_allocator<double> >; Vec2 = thrust::device_vector<double, thrust::device_allocator<double> >&; Backend = amgcl::backend::cuda<double>; InnerProduct = amgcl::solver::detail::default_inner_product; typename amgcl::math::scalar_of<typename Backend::value_type>::type = double]’
/home/tqiu/data/opm/pr3300/prerequisites/amgcl/amgcl/solver/runtime.hpp:207:13:   required from ‘std::tuple<long unsigned int, typename amgcl::math::scalar_of<typename Backend::value_type, void>::type> amgcl::runtime::solver::wrapper<Backend, InnerProduct>::operator()(const Matrix&, const Precond&, const Vec1&, Vec2&&) const [with Matrix = amgcl::backend::cuda_matrix<double>; Precond = amgcl::runtime::preconditioner<amgcl::backend::cuda<double> >; Vec1 = thrust::device_vector<double, thrust::device_allocator<double> >; Vec2 = thrust::device_vector<double, thrust::device_allocator<double> >&; Backend = amgcl::backend::cuda<double>; InnerProduct = amgcl::solver::detail::default_inner_product; typename amgcl::math::scalar_of<typename Backend::value_type>::type = double]’
/home/tqiu/data/opm/pr3300/prerequisites/amgcl/amgcl/solver/runtime.hpp:227:23:   required from ‘std::tuple<long unsigned int, typename amgcl::math::scalar_of<typename Backend::value_type, void>::type> amgcl::runtime::solver::wrapper<Backend, InnerProduct>::operator()(const Precond&, const Vec1&, Vec2&&) const [with Precond = amgcl::runtime::preconditioner<amgcl::backend::cuda<double> >; Vec1 = thrust::device_vector<double, thrust::device_allocator<double> >; Vec2 = thrust::device_vector<double, thrust::device_allocator<double> >&; Backend = amgcl::backend::cuda<double>; InnerProduct = amgcl::solver::detail::default_inner_product; typename amgcl::math::scalar_of<typename Backend::value_type>::type = double]’
/home/tqiu/data/opm/pr3300/prerequisites/amgcl/amgcl/make_solver.hpp:142:31:   required from ‘std::tuple<long unsigned int, typename amgcl::math::scalar_of<typename IterativeSolver::backend_type::value_type, void>::type> amgcl::make_solver<Precond, IterativeSolver>::operator()(const Vec1&, Vec2&&) const [with Vec1 = thrust::device_vector<double, thrust::device_allocator<double> >; Vec2 = thrust::device_vector<double, thrust::device_allocator<double> >&; Precond = amgcl::runtime::preconditioner<amgcl::backend::cuda<double> >; IterativeSolver = amgcl::runtime::solver::wrapper<amgcl::backend::cuda<double> >; typename amgcl::math::scalar_of<typename IterativeSolver::backend_type::value_type>::type = double]’
/home/tqiu/data/opm/pr3300/opm/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.cpp:191:43:   required from ‘void bda::amgclSolverBackend<block_size>::solve_system(double*, Opm::WellContributions&, bda::BdaResult&) [with unsigned int block_size = 4]’
/home/tqiu/data/opm/pr3300/opm/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.cpp:273:17:   required from ‘bda::SolverStatus bda::amgclSolverBackend<block_size>::solve_system(int, int, int, double*, int*, int*, double*, Opm::WellContributions&, bda::BdaResult&) [with unsigned int block_size = 4]’
/home/tqiu/data/opm/pr3300/opm/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.cpp:288:1:   required from here
/usr/local/cuda-11.1/include/thrust/detail/static_assert.h:51:44: error: static assertion failed: unimplemented for this system
 #  define THRUST_STATIC_ASSERT_MSG(B, msg) static_assert(B, msg)
                                            ^
/usr/local/cuda-11.1/include/thrust/detail/static_assert.h:51:44: note: in definition of macro ‘THRUST_STATIC_ASSERT_MSG’
 #  define THRUST_STATIC_ASSERT_MSG(B, msg) static_assert(B, msg)
```
I didn't actually find a `static_assert` being used.
This error only occurred for blocksize=4, whereas I'm used to seeing errors occur for all defined blocksizes (1-4).
My thrust version is 10.09.10 patch 1. CUDA is 11.1.
CMake also mentions `amgclSolverBackend.cpp.o` is viewed as a CXX object.

blattms:
Apparently (i.e. see amgcl cuda examples), usage of the CUDA backend are suppossed to be compiled with nvcc. sorry, that I was not aware about that, That is what you are doing setting the flag, it seems. But as NVCC might not support fully support the full CPP standard, we should eventually refactor to only compile the minimum needed with NVCC. 

For now please make this PR compile without CUDA and we merge. Then we attack the next problem.

Tongdongq:
Do you mean: remove CUDA support for amgcl or make sure it compiles if a user does not have CUDA installed? Because I think the latter is addressed in 58bd1d4.

Tongdongq:
@blattms I added VexCL, it works with disabling CUDA and/or OpenCL.
Performance is bad, see startpost.

blattms:
As a dumb user as I am, I get 
```
Error: [/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.cpp:101] Error unknown value for amgcl parameter 'backend_type'
Program threw an exception: [/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.cpp:101] Error unknown value for amgcl parameter 'backend_type'
```
when running `flow --accelerator-mode=amgcl --matrix-add-well-contributions=true <CASE>`

blattms:
if I try to be a tiny bit smarter and add a file amgcl_options.json with just
```
{
"backend_type": "cpu",
}
```
(Not even sure whether that is the correct syntax) then i get time step chopping for SPE1CASE2 until flow gives up with:
```
Program threw an exception: [/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp:518] Solver failed to converge after cutting timestep 10 times.
```
I guess we missed the CPU backend?

Same happens if I use `backend_type": "vexcl".

blattms:
If I activate both opencl and cuda the branch does not compile on my system:
```
[  2%] Building CUDA object CMakeFiles/opmsimulators.dir/opm/simulators/linalg/bda/amgclSolverBackend.cpp.o
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.cpp(183): error: "auto" is not allowed here

/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/amgclSolverBackend.cpp(273): error: expected a "("
          detected during:
            instantiation of "bda::SolverStatus bda::amgclSolverBackend<block_size>::solve_system(int, int, int, double *, int *, int *, double *, Opm::WellContributions &, bda::BdaResult &) [with block_size=1U]" 
(41): here
            instantiation of "bda::amgclSolverBackend<block_size>::amgclSolverBackend(int, int, double, unsigned int, unsigned int) [with block_size=1U]" 
(346): here

2 errors detected in the compilation of "/tmp/tmpxft_00007073_00000000-6_amgclSolverBackend.cpp1.ii".
make[2]: *** [CMakeFiles/opmsimulators.dir/build.make:1012: CMakeFiles/opmsimulators.dir/opm/simulators/linalg/bda/amgclSolverBackend.cpp.o] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:1584: CMakeFiles/opmsimulators.dir/all] Fehler 2
make: *** [Makefile:141: all] Fehler 2
```
For some reason nvcc is called with `-Xcom-Xcompiler=-fPIC   -std=c++11` in this case, which does not allow such auto usage. I will investigate why.

blattms:
I hate build systems :angry: 
See Tongdongq/opm-simulators#5 for a fix of the compilation error.

blattms:
Too fast.

blattms:
Sorry, I closed the wrong one.  Reopened.

Should have been  Tongdongq#5. Well l force pushed to that one to at least use C++ standard 14, Unfortunately, that is the highest one allowed with CMake 3.10., but not enough for auto support. Hence please fix that here.

Just another reason to use nvcc only where really needed.

Tongdongq:
I apparently missed the default version where both OpenCL and CUDA are active.
I factored out the CUDA part now. The `amgclSolverBackend.cu` is only compiled if amgcl and CUDA are found.
The timestep chopping usually happens because some kind of exception is thrown, but not caught (only caught higher up). The `boost::property_tree::get()` [can throw](https://www.boost.org/doc/libs/1_76_0/doc/html/property_tree/accessing.html), I changed it to the default value version, that one cannot throw. Although I'm not sure if that was the error in the first place.

blattms:
I am afraid that I still have compilation issues:
```
/home/mblatt/src/dune/opm-gpu/opm-simulators/tests/test_openclSolver.cpp: In instantiation of ‘Dune::BlockVector<Dune::FieldVector<double, bz> > testOpenclSolver(const ptree&, const string&, const string&) [with int bz = 3; boost::property_tree::ptree = boost::property_tree::basic_ptree<std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >; std::__cxx11::string = std::__cxx11::basic_string<char>]’:
/home/mblatt/src/dune/opm-gpu/opm-simulators/tests/test_openclSolver.cpp:101:66:   required from here
/home/mblatt/src/dune/opm-gpu/opm-simulators/tests/test_openclSolver.cpp:82:28: error: no matching function for call to ‘Opm::WellContributions::WellContributions(const char [7])’
     Opm::WellContributions wellContribs("opencl");
                            ^~~~~~~~~~~~
In file included from /home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/BdaSolver.hpp:25:0,
                 from /home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/BdaBridge.hpp:28,
                 from /home/mblatt/src/dune/opm-gpu/opm-simulators/tests/test_openclSolver.cpp:32:
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.hpp:178:5: note: candidate: Opm::WellContributions::WellContributions(std::__cxx11::string, bool)
     WellContributions(std::string accelerator_mode, bool useWellConn);
     ^~~~~~~~~~~~~~~~~
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.hpp:178:5: note:   candidate expects 2 arguments, 1 provided
make[2]: *** [CMakeFiles/test_openclSolver.dir/build.make:63: CMakeFiles/test_openclSolver.dir/tests/test_openclSolver.cpp.o] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:3834: CMakeFiles/test_openclSolver.dir/all] Fehler 2
make[1]: *** Es wird auf noch nicht beendete Prozesse gewartet....
/home/mblatt/src/dune/opm-gpu/opm-simulators/tests/test_cusparseSolver.cpp: In instantiation of ‘Dune::BlockVector<Dune::FieldVector<double, bz> > testCusparseSolver(const ptree&, const string&, const string&) [with int bz = 3; boost::property_tree::ptree = boost::property_tree::basic_ptree<std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >; std::__cxx11::string = std::__cxx11::basic_string<char>]’:
/home/mblatt/src/dune/opm-gpu/opm-simulators/tests/test_cusparseSolver.cpp:102:68:   required from here
/home/mblatt/src/dune/opm-gpu/opm-simulators/tests/test_cusparseSolver.cpp:83:28: error: no matching function for call to ‘Opm::WellContributions::WellContributions(const char [9])’
     Opm::WellContributions wellContribs("cusparse");
                            ^~~~~~~~~~~~
In file included from /home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/BdaSolver.hpp:25:0,
                 from /home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/BdaBridge.hpp:28,
                 from /home/mblatt/src/dune/opm-gpu/opm-simulators/tests/test_cusparseSolver.cpp:32:
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.hpp:178:5: note: candidate: Opm::WellContributions::WellContributions(std::__cxx11::string, bool)
     WellContributions(std::string accelerator_mode, bool useWellConn);
     ^~~~~~~~~~~~~~~~~
/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/linalg/bda/WellContributions.hpp:178:5: note:   candidate expects 2 arguments, 1 provided
```

Please note that I rebased your branch onto current master (to save me from finding correct versions of the other modules). Maybe that is the problem and you could do a rebase yourself?

blattms:
jenkins build this please

blattms:
Thanks. Now it compiles for my settings.

My dumb user test `flow --accelerator-mode=amgcl --matrix-add-well-contributions=true SPE1CASE2` and the infamous json file with
```
{
"backend_type": "cuda",
}
```
no leads to:
```
Problem: 
Timestep chopped to 0.000015 days


Starting time step 0, stepsize 1.53158e-05 days, at day 0/31, date = 01-Jan-2015
Initializing amgclSolverBackend, matrix size: 900 blocks, nnzb: 1780
Maxit: 200, tolerance: 1.000000e-02
DeviceID: 0


Program threw an exception: [/home/mblatt/src/dune/opm-gpu/opm-simulators/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp:518] Solver failed to converge after cutting timestep 10 times.
```
in the first time step. Is this really expected for SPE1CASE2? Seems a bit fishy.

If I remove the json files flow runs through (I guess with the CPU backend), but I see warnings from AMGCL:
```
AMGCL WARNING: unknown parameter verbose
AMGCL WARNING: unknown parameter verbose
AMGCL WARNING: unknown parameter verbose
```

blattms:
Using vexcl as a backend, the behavior is the same as for CUDA. Unfortunately,I cannot tell which backend is used. I see the same output for vexcl and cuda:
```
Initializing amgclSolverBackend, matrix size: 900 blocks, nnzb: 1780
Maxit: 200, tolerance: 1.000000e-02
DeviceID: 0
```
Not sure what the device id means in the cuda case.

Looking the *.DBG file I see a cryptic message "Caught Exception: <unspecified file>(3): expected key string" that might indicate that my json file is rubbish. This is of course opaque to any user, if he even finds it..

Tongdongq:
I have the same for that file, but if you remove the `,` at the end, it is a valid json, gets read, and then amgcl defaults to using amg.
I did not account for incomplete datafiles, but that should be easily fixed.
Without a json file, it should print the default settings used, and be verbose if the `--flow-linear-solver-verbosity` is 2 or higher. I don't see the unknown parameter verbose warnings.

blattms:
Just found that out, too. Thanks.

Now I see:
```
1. GeForce GTX 1060 6GB (NVIDIA CUDA)

1. GeForce GTX 1060 6GB (NVIDIA CUDA)
``
at least for every timestep. maybe outputting this only once would be enough?

It also seems that amgcl neglects the --opencl-platform-id, and always uses the first one. I can even use values higher than the number of available platforms. How am I supposed to tell amgcl which platform to use?

Of course we should abort with a message that the json parsing failed.

**Comment by blattms** [comment]
I probably need to double check my setup, but vexcl seems a lot too slow compared to cuda. It takes 20 times the time as with cuda and the iteration counts are the same.

**Comment by blattms** [comment]
What is the default solver used for amgcl? It seems  a bit like this differs depending on the backend. For CPU 3438 linear iterations are used, while for the others it is 340. Shouldn't it be same for all?

**Comment by Tongdongq** [comment]
> I probably need to double check my setup, but vexcl seems a lot too slow compared to cuda. It takes 20 times the time as with cuda and the iteration counts are the same.

I have a similar result.
`amgclSolver` indeed ignores both deviceIDs when using vexcl. I'll try to change it to `--opencl-device-id`.
I don't think the default preconditioner should depend on the backend, I think it always defaults to amg.

File: opm/simulators/linalg/bda/amgclSolverBackend.cpp [hunk 4]
```diff
@@ -0,0 +1,332 @@
...
+            auto b_ptr = reinterpret_cast<dvec_type*>(b);
+            auto x_ptr = reinterpret_cast<dvec_type*>(x.data());
+            auto B = amgcl::make_iterator_range(b_ptr, b_ptr + N / block_size);
+            auto X = amgcl::make_iterator_range(x_ptr, x_ptr + N / block_size);
+
+            // actually solve
+            std::tie(iters, error) = solve(B, X);
+        } else if (backend_type == Amgcl_backend_type::vexcl) {
+#if HAVE_VEXCL
+            vex::Context ctx(vex::Filter::Env && vex::Filter::Count(1));
```
ddemidov, line 249: The device filter should probably take `--opencl-device-id` into account. See https://vexcl.readthedocs.io/en/latest/initialize.html#device-filters for possible filters. Also, `Filter::Env` here allows controlling the device choice with environment variables.

The context initialization should probably be done once, as it is done for cuda in `cuda_solve`.

ddemidov:
Hi,

> I probably need to double check my setup, but vexcl seems a lot too slow compared to cuda. It takes 20 times the time as with cuda and the iteration counts are the same.

This should not really be the case. OpenCL should be slightly slower, but not as much. For example, here is a result of solving a 3D Poisson system on a 128^3 grid (both versions are compiled from [amgcl/examples/solver.cpp](https://github.com/ddemidov/amgcl/blob/master/examples/solver.cpp)):

**CUDA**
```
solver_cuda -n 128
NVIDIA GeForce GTX 1050 Ti

Solver
======
Type:             BiCGStab
Unknowns:         2097152
Memory footprint: 112.00 M

Preconditioner
==============
Number of levels:    4
Operator complexity: 1.62
Grid complexity:     1.13
Memory footprint:    567.74 M

level     unknowns       nonzeros      memory
---------------------------------------------
    0      2097152       14581760    422.42 M (61.61%)
    1       263552        7918340    127.63 M (33.46%)
    2        16128        1114704     15.07 M ( 4.71%)
    3          789          53055      2.62 M ( 0.22%)

Iterations: 10
Error:      2.50965e-09

[Profile:          2.177 s] (100.00%)
[ self:            0.254 s] ( 11.68%)
[  assembling:     0.155 s] (  7.13%)
[  setup:          1.205 s] ( 55.36%)
[  solve:          0.562 s] ( 25.83%)
```

**OpenCL**
```
solver_vexcl_cl -n 128
1. NVIDIA GeForce GTX 1050 Ti (NVIDIA CUDA)

Solver
======
Type:             BiCGStab
Unknowns:         2097152
Memory footprint: 112.00 M

Preconditioner
==============
Number of levels:    4
Operator complexity: 1.62
Grid complexity:     1.13
Memory footprint:    744.74 M

level     unknowns       nonzeros      memory
---------------------------------------------
    0      2097152       14581760    553.22 M (61.61%)
    1       263552        7918340    168.88 M (33.46%)
    2        16128        1114704     20.01 M ( 4.71%)
    3          789          53055      2.62 M ( 0.22%)

Iterations: 10
Error:      2.50965e-09

[Profile:          2.181 s] (100.00%)
[ self:            0.097 s] (  4.42%)
[  assembling:     0.115 s] (  5.29%)
[  setup:          1.406 s] ( 64.47%)
[  solve:          0.563 s] ( 25.81%)
```

ddemidov:
> I see warnings from AMGCL:
AMGCL WARNING: unknown parameter verbose

@blattms, the verbosity parameter was added after the last published version (1.4.0), and you are probably using the exact 1.4.0 version.

ddemidov:
> What is the default precinditioner used for amgcl? It seems a bit like this depends on the backend. For CPU 3438 linear iterations are used, while for the others it is 340. Shouldn't it be same for all?

The default preconditioner when runtime interface is used is smoothed aggregation amg, with spai0 as relaxation, but the defaults seem to be overridden here:

https://github.com/OPM/opm-simulators/pull/3346/files#diff-32af453351220dd31192c9cc7ff238f9eec3bee0d431a458454fd7b7b04717c3R84-R93

blattms:
Thanks @ddemidov for the insight. Much appreciated.

Indeed that means that without a json file (which I am doing for the cpu test), we use a different solver. That explains the different runtimes.

Thanks also for the other hints. We will look into them.

Tongdongq:
Indeed thanks!

I now get a platform and device based on the passed IDs, then set the environment variables with their names, and let vexcl use those variables to get the device. This is a bit cumbersome, but at least a user can choose now.

opm/simulators/linalg/bda/amgclSolverBackend.cpp
```diff
@@ -199,9 +199,39 @@ void amgclSolverBackend<block_size>::convert_data(double *vals, int *rows) {
...
+
+    std::vector<cl::Device> devices;
+    platforms[platformID].getDevices(CL_DEVICE_TYPE_ALL, &devices);
+
+    if (devices.size() <= deviceID){
+        OPM_THROW(std::logic_error, "Error chosen too high OpenCL device ID");
+    }
+
+    devices[deviceID].getInfo(CL_DEVICE_NAME, &opencl_info);
+    setenv("OCL_DEVICE", opencl_info.c_str(), true); // overwrites if OCL_DEVICE is already set
```
ddemidov, line 223: Using environment filter when you already have a `cl::Device` reference is a bit roundabout. `vex::Context` in vexcl is just a convenience wrapper, and all it is needed for is filter functionality for easy initialization. It is always implicitly converted to `std::vector<cl::CommandQueue>` for such things as `vex::vector` constructor, so I would use `std::vector<cl:CommandQueue>` directly instead of `vex::Context` here. All you need to do at this point is something like
```cpp
cl::Context c(devices[deviceID]);
cl::CommandQueue q(c, devices[deviceID]);
ctx.push_back(q); // ctx is now std::vector<cl::CommandQueue>

```

Tongdongq: I've overwritten it with 796d1b7

blattms:
Sigh, my AMD GPU setup with opencl seems to be broken again. Will test only with NVIDIA for now,

blattms:
jenkins build this please

blattms:
Got AMD running with rocm again. We are running into convergence issues on SPE10 when using the BiCGStab with ILU0 and the vexcl backend when using my gfx803 AMD GPU. If we use our opencl implementation it works reasonably well (not perfect, though).

Tongdongq:
Vexcl for Norne also has issues. Are you using our openclSolver with
level_scheduling?

ddemidov:
Just out of interest, could you perhaps save and post the failing matrices (and RHS) here in the matrix market format? [`amgcl::io::mm_write`](https://github.com/ddemidov/amgcl/blob/681bfc71edf6c3721109bd0f263b33c92ec3ba2e/amgcl/io/mm.hpp#L347-L383) from `<amgcl/io/mm.hpp>` takes the matrices in the same format solver constructor does, so that should be easy.

Tongdongq:
I'm on vacation now, perhaps Markus can, otherwise I'll do it later

ddemidov:
ILU0 for vexcl uses jacobi approximation for the upper and lower triangular solves, so you probably need to increase [`prm.precond.relax.solve.iters`](https://amgcl.readthedocs.io/en/latest/components/relaxation.html#_CPPv4N5amgcl10relaxation6detail9ilu_solve6params5itersE)

blattms:
Is there a reason why the cuda and vexcl backends differ in this regard? Seems a bit surprising.

ddemidov:
CUSPARSE library provides implementation of ILU0, which is used in amgcl when [`<amgcl/relaxation/cusparse_ilu0.hpp>`](https://github.com/ddemidov/amgcl/blob/master/amgcl/relaxation/cusparse_ilu0.hpp) is included. Otherwise, same ILU0 implementation is used in both the  CUDA and VexCL backends. Other ILU variants (iluk, ilut, ilup) are using Jacobi approximation in both backends. Here are some quick tests using 3D Poisson on a 64^3 grid:

**CUDA backend, (CUSPARSE ILU0)**
```
$ solver_cuda -n 64 -1 precond.type=ilu0
NVIDIA GeForce GTX 1050 Ti

Solver
======
Type:             BiCGStab
Unknowns:         262144
Memory footprint: 14.00 M

Preconditioner
==============
Relaxation as preconditioner
  Unknowns: 262144
  Nonzeros: 1810432
  Memory:   48.44 M

Iterations: 44
Error:      9.53726e-09

[Profile:          1.387 s] (100.00%)
[ self:            0.251 s] ( 18.06%)
[  assembling:     0.025 s] (  1.80%)
[  setup:          0.120 s] (  8.63%)
[  solve:          0.992 s] ( 71.50%)
```

**VexCL backend (using CUDA backend for VexCL, ILU0 with triangular solves approximated by 2 Jacobi iterations)**
```
solver_vexcl_cuda -n 64 -1 precond.type=ilu0
1. NVIDIA GeForce GTX 1050 Ti

Solver
======
Type:             BiCGStab
Unknowns:         262144
Memory footprint: 14.00 M

Preconditioner
==============
Relaxation as preconditioner
  Unknowns: 262144
  Nonzeros: 1810432
  Memory:   63.25 M

Iterations: 59
Error:      8.45782e-09

[Profile:          0.378 s] (100.00%)
[ self:            0.076 s] ( 20.23%)
[  assembling:     0.015 s] (  3.85%)
[  setup:          0.053 s] ( 13.95%)
[  solve:          0.234 s] ( 61.98%)
```

The number of iterations has increased, but since the Jacobi approximation is embarassingly parallel, the solution is 4 times faster (and the setup is 2 times cheaper).

The quality of Jacobi approximation to the triangular solves may be improved, but that of course make each iteration more expensive:
```
solver_vexcl_cuda -n 64 -1 precond.type=ilu0 precond.solve.iters=5
1. NVIDIA GeForce GTX 1050 Ti

Solver
======
Type:             BiCGStab
Unknowns:         262144
Memory footprint: 14.00 M

Preconditioner
==============
Relaxation as preconditioner
  Unknowns: 262144
  Nonzeros: 1810432
  Memory:   63.25 M

Iterations: 43
Error:      7.22054e-09

[Profile:          0.494 s] (100.00%)
[ self:            0.089 s] ( 18.01%)
[  assembling:     0.019 s] (  3.79%)
[  setup:          0.052 s] ( 10.59%)
[  solve:          0.334 s] ( 67.61%)
```

blattms:
@Tongdongq I used the default options (the log says graph coloring is used).

Tongdongq:
I did a test to vary the `precond.solve.iters` parameter, and added them to the startpost.
Short summary: by using 4-6 iterations, the time for the linear solving part is faster than the exact ILU0 application in cuda. The OPM reported pre/post time increased a lot between these and older tests.

Issue #3347, opened 2021-06-03
Group Level Injection RESV Control Produces Confusing Results


bska:
I came across some unexpected behaviour whilst analyzing an unrelated problem.  It **appears** that the simulator's implementation of `GCONINJE`'s `RESV` control applies unit conversion (or formation volume factor reduction) one too many times, but I've not been able to pinpoint the exact source of the issue.

I changed [SPE1CASE2](https://github.com/OPM/opm-tests/blob/d7a6d647214afe5b52dd3ecd254b17c9c9001edc/spe1/SPE1CASE2.DATA) as follows
```diff
RPTSCHED
@@ -411,8 +429,14 @@ WCONPROD
 
 WCONINJE
 -- Item #:1     2       3       4      5      6  7
-       'INJ'   'GAS'   'OPEN'  'RATE'  100000 1* 9014 /
+--     'INJ'   'GAS'   'OPEN'  'RATE'  100000 1* 9014 /
+  'INJ'  'GAS'  'OPEN'  'GRUP'  1* 1* 9014 /
+/
+
+GCONINJE
+  'G1' 'GAS' 'RATE' 100.0E3 60.0E3 /
 /
```
(in addition to outputting a few more group-level summary vectors).  The `60.0E3` RB/day limit (item 5; roughly 0.1104 m3/s) is a sort of average value of the well-level reservoir volume injection rate in the standard run.  I get the reservoir volume injection rate results in the picture below.  The blue curve (left-hand axis) is from a run **without** the RESV limit (i.e., with only the RATE limit).  The red curve (right-hand axis) is from a run that does include the RESV limit (60e3 RB/day).  While we should not expect the two curves to be equal since the controls differ, they should have comparable orders of magnitude.  Here, however, the run **without** RESV produces rates on the order of 50e3 RB/day to 85e3 RB/day while the run **with** the RESV constraint produces rates on the order of 200 RB/day to 620 RB/day.  As another experiment, I disabled the group constraints and placed the RESV constraint of 60e3 RB/day at the well level instead.  Doing so produces well (and group) level reservoir volume injection rates in the order of 60e3 RB/day for all times.  That suggests that the RESV problem is limited to the GCONINJE context rather than being a general problem with Flow's overall constraint implementation.

Like I said, I've tried to trace the logic here but now I'm at a point where I don't know enough about the system to know where to look anymore.  Any suggestions would be most welcome.

---

[...]

PR #3349, opened 2021-06-04
Scale target with resv conversion factor

totto82:
Fix for #3347

### DIFF
--- a/opm/simulators/wells/TargetCalculator.cpp
+++ b/opm/simulators/wells/TargetCalculator.cpp
@@ -184,7 +184,7 @@ double InjectionTargetCalculator::groupTarget(const Group::InjectionControls& ct
     case Group::InjectionCMode::RATE:
         return ctrl.surface_max_rate;
     case Group::InjectionCMode::RESV:
-        return ctrl.resv_max_rate;
+        return ctrl.resv_max_rate / resv_coeff_[pos_];
     case Group::InjectionCMode::REIN: {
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

totto82:
image

totto82:
@bska I assume this is closer to the results you expected.

bska:
> I assume this is closer to the results you expected.

Yes it is, thank you very much.  I'll try this PR on the model where I noticed this symptom originally.

bska:
> > assume this is closer to the results you expected.
> 
> Yes it is, thank you very much. I'll try this PR on the model where I noticed this symptom originally.

This PR is **much** better than current master, but we do not quite match ECLIPSE yet.  Below is a plot of the group-level reservoir volume injection rate when the model is constrained as
```
GCONINJE
-- nm         ph           cm         sir          rvir      rifr      vrfr     rhl
 'INJE'      'GAS'      'RATE'       1000000.       1000.     1*      1*       YES   /
/
```
and the reservoir volume injection rate constraint is active.  The value should be very close to (or, strictly speaking, exactly equal to) 1000, but we're a little off at roughly 976 even with this PR.

I can share this model with you offline if you'd like to have a look.  There's also a possibly related issue with `VREP`, but we can address that in a later PR.

---

INJE_Res_Volume_Injection_Rate

bska:
In any case, I'll merge this PR to get us closer to the target.

PR #3352, opened 2021-06-04
Perf data

joakim-hove:
Assemble all the per perforation data in one datastructure.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -67,6 +67,7 @@ list (APPEND MAIN_SOURCE_FILES
   opm/simulators/wells/MultisegmentWellEval.cpp
   opm/simulators/wells/MultisegmentWellGeneric.cpp
   opm/simulators/wells/ParallelWellInfo.cpp
+  opm/simulators/wells/PerfData.cpp
   opm/simulators/wells/SegmentState.cpp
   opm/simulators/wells/StandardWellEval.cpp
   opm/simulators/wells/StandardWellGeneric.cpp
@@ -286,6 +287,7 @@ list (APPEND PUBLIC_HEADER_FILES
   opm/simulators/utils/ParallelEclipseState.hpp
   opm/simulators/utils/ParallelRestart.hpp
   opm/simulators/utils/PropsCentroidsDataHandle.hpp
+  opm/simulators/wells/PerfData.hpp
   opm/simulators/wells/PerforationData.hpp
   opm/simulators/wells/RateConverter.hpp
   opm/simulators/utils/readDeck.hpp
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -188,13 +188,15 @@ loadRestartData(const data::Wells& rst_wells,
             well_state.wellRates(well_index)[i] = rst_well.rates.get(phs[i]);
         }
 
-        auto& perf_pressure = well_state.perfPress(well_index);
         auto& perf_rates = well_state.perfRates(well_index);
         auto& perf_phase_rates = well_state.perfPhaseRates(well_index);
-        const auto& perf_data = this->well_perf_data_[well_index];
 
-        for (std::size_t perf_index = 0; perf_index < perf_data.size(); perf_index++) {
-            const auto& pd = perf_data[perf_index];
+        auto& perf_data = well_state.perfData(well_index);
+        auto& perf_pressure = perf_data.pressure;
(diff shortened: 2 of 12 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
Comments to this?

blattms:
benchmark please

bska:
> Comments to this?

Please address the [new warnings](https://ci.opm-project.org/job/opm-simulators-PR-builder/2543/gcc/new/) first.

blattms:
I am probably not educated enough to comment in the part from OPM and don't have a clear picture what the final goal is. But here are my two cents, anyway:

As someone coming from high performance computing, I am usually fond of plain old data in big arrays (rather than arrays of classes with multiple small arrays), like it was before, as it usually minimizes cache misses. I am a bit unsure whether this plays a role here or not (might depend on the number of wells and the number of perforation for each well). But minimizing cache misses usually does not hurt.

I am not sure what problem you are solving here. If the problem is getting rid of the outrages numbers of indexing in the code,  one could have used a of view approach on the global arrays. PerfDataView, a view on the data of a single well, would have references to the global arrays and the well index and could be used like your PerData (without introducing additional allocations and cache misses).

bska:
> I am usually fond of plain old data in big arrays (rather than arrays of classes with multiple small arrays), like it was before, as it usually minimizes cache misses.

Whether to use "array of structures" or "structure of arrays" depends greatly on the expected/common access pattern in the performance sensitive inner loops.  Especially on whether you expect to touch "all" items associated to a single entity (e.g., well, group, connection) before moving on to the next entity or whether you're processing a single item (e.g., reservoir volume flow rates) for all entities before moving on to another item.  I haven't looked too closely at all of the well code, but my impression is that it's a bit of a mix between the two.

joakim-hove:
> I am usually fond of plain old data in big arrays (rather than arrays of classes with multiple small arrays)

It is of course in the mind of the beholder; but my preference is generally "array of structs" rather than the opposite. The reason I prefer "array of structs" is that I find it easier to work with (to test and reason about in particular, and also changes - e.g. adding a new field stay more localized). I realize that "struct of (large) arrays" might be more performant, but at least for the well code all[1] loops over perforations only iterate over the perforations of one well - i.e. the vectors in the new `PerfData` class - I therefor do not think decreased cache hit rate should be a problem.

> I am not sure what problem you are solving here.

Making the code easier to work with, more compact and testable, and in particular simpler to initialize the structures from a restart file.

> If the problem is getting rid of the outrages numbers of indexing in the code, 

Yes - that is certainly large part of it.

> one could have used a of view approach on the global arrays. 

Yes - that would have been possible - but more complex.


[1]: All that I am aware of at least ....

joakim-hove:
jenkins build this please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1.005    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 0.999    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 1.043    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.933    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 1.017    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1229

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

blattms:
What is more beautiful is indeed in the eye of the beholder. IMHO we might close the door for some optimizations in the future (once the rest of the well code is performant).

> loops over perforations only iterate over the perforations of one well

There always seems to be an outer loop over the wells if I am not mistaking. We never process only one well. That means that were actually reading the complete large arrays as the order of the wells seems to be static. It could be of course that the order of processing the wells was not from front to back...

blattms:
I am shutting up now... At least currently there is no penalty.

joakim-hove:
> IMHO we might close the door for some optimizations in the future (once the rest of the well code is performant).

I see that point - my thinking is that if/when we want to do optimizations of this kind, that will  be simpler if the code in question is easy to reason about (famous last words ....).

PR #3354, opened 2021-06-05
MSW: drop unused method

akva2:
Unused which is a good thing since it does not compile if you instance it. No such member in segmentSet()


akva2:
jenkins build this please

bska (approved):
Very good. I'll merge into master.

PR #3359, opened 2021-06-08
Use ecl well for total number of connections

joakim-hove:
The `Schedule::Well` objects will always have all connections on all processors - i.e. the `global_num_perf_this_well` can be initialized from here, without performing a collective sum. 

An important part of the motivation for the change is to check my own understanding?

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -298,10 +298,11 @@ void WellState::init(const std::vector<double>& cellPressures,
     for (int w = 0; w < nw; ++w) {
         // Initialize perfphaserates_ to well
         // rates divided by the number of perforations.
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
> I believe it is the number of open well connections at this time step.

I have checked - and that is correct. I will update.

joakim-hove:
OK - I have updated to use `ConnectionSet::num_open()`

joakim-hove:
jenkins build this please

joakim-hove:
> If the former holds this change is correct and better than the original one.

Well - the former was not correct; but after update to use `ConnectionSet::num_open()` that should be the case. Will merge later today.

PR #3361, opened 2021-06-08
Suggestion: Simplify check for number of perforations

joakim-hove:
I might very well have misunderstood something here, but my understanding goes as follows:

1. We copy the state of the perforation rates from the previous step - this is good for convergence, but not essential for correctness.
2. If the number of the perforations have changed - either because new perforations have been specified in the input, or alternatively due to loadbalancing - we do not copy the previous perforation rates but rather initialize with an average rate.

In current master the check for change in perforation numbers is global - taking all processers into account, whereas I with this PR suggest checking just the local process for changes in number of perforations. I realize that this *can* optionally lead to a situation where perforations on process 1 get initialized by copying from previous state, whereas perforations on process 2 are initialized with an average rate - but as long as this is "just for convergence" I would say that was totally OK?

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -396,11 +396,7 @@ void WellState::init(const std::vector<double>& cellPressures,
                 }
 
                 const int num_perf_this_well = new_iter->second[2];
-
-                const int num_perf_changed = parallel_well_info[w]->communication()
-                    .sum(static_cast<int>(num_perf_old_well != num_perf_this_well));
-                const bool global_num_perf_same = num_perf_changed == 0;
-
+                const bool global_num_perf_same = (num_perf_this_well == num_perf_old_well);
 
                 // copy perforation rates when the number of
                 // perforations is equal, otherwise initialize


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

blattms (commented):
Doesn't not seem worse than the previous one (but it will do something different in parallel than in in serial).

I am actually questioning the decision based on the number of perforations all together: Can't it happen that one perforation gets shut and a previous shut one opened? In that case we would take the old rates even though some of them are from perforation to other cells then now.

joakim-hove:
> Doesn't not seem worse than the previous one 

Well - it is simpler and a global communication has been removed; so in my book that is a minor improvement.

> (but it will do something different in parallel than in in serial).

Well - only in the case that a well is distributed among different processors, and a new connection is added runtime. In that case there will be a difference - but again only possibly affecting convergence as I understand it.

> [.... ] Can't it happen that one perforation gets shut and a previous shut one opened? In that case we would take the old rates even though some of them are from perforation to other cells then now.

Yes I agree - I also thought of that; to protect against that one would have to check the ids of all perforations; however since this is only for convergence and furthmore *quite seldom* I think we can ignore that possibility.

GitPaean:
Just some comments for information, not a reviewing comment. 

> We copy the state of the perforation rates from the previous step - this is good for convergence, but not essential for correctness.

In very rare situations, it might make some noticeable difference.  It will affect the explicit quantities (using the data from previous time steps) in the well calculation. And the case needs to be sensitive to small changes in the result. 

As discussed above, only checking the number of the perforations itself is not a perfect strategy anyway.

joakim-hove:
jenkins build this please

blattms:
> > Doesn't not seem worse than the previous one

> Well - it is simpler and a global communication has been removed; so in my book that is a minor improvement.

Just for clarification:
Please be aware the communicator is not MPI_COMM_WORLD, but only a xmall subset, most often there is even no communication at all. I would not call this global communication.

joakim-hove:
> Please be aware the communicator is not MPI_COMM_WORLD

Yes - @atgeirr told me; but I admittedly thought it was `MPI_COMM_WORLD` at play.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

PR #3363, opened 2021-06-08
removed dissolved gas and vapporized oil from RESV injection rates

totto82:
The standard RESV coefficients use the average RS and RV values from the reservoir. For injection we assume no dissolved gas or vaporized oil. This is supposed to fix the issue reported by @bska in #3349

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -373,6 +373,10 @@ namespace Opm {
                            const int pvtreg,
                            std::vector<double>& resv_coeff) override;
 
+            void calcInjRates(const int fipnum,
+                           const int pvtreg,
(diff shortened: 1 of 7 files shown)


totto82:
jenkins build this please

bska:
> This is supposed to fix the issue reported in #3349

Confirmed.  This PR does fix the problem I reported earlier.  I'll have a look at the details in this PR.

totto82:
We should add a test that expose this. (Jenkins should not been green if RESV + gas injection was in the test suite). Maybe the model2 version you shared with me.

bska (approved):
I agree with this change.  I think we always assume we inject single-phase fluids with no mixing.  I'll merge into master.

bska:
> We should add a test that expose this. Maybe the model2 version you shared with me.

Yes, I think that's a good idea.  I'll upload the datasets and try to add them to the regression tests.

PR #3364, opened 2021-06-08
Determine index of pressure from model used to increase usability

blattms:
Previously, the user had to specify it in the json file read from the FlexibleSolver or 1 was used. Unfortunately, the index depends on the model used and it seem rather opaque to a user what that index is.

With this commit we determine the pressure index from the model.

### DIFF
--- a/opm/simulators/linalg/FlexibleSolver.hpp
+++ b/opm/simulators/linalg/FlexibleSolver.hpp
@@ -50,14 +50,16 @@ public:
     /// Create a sequential solver.
     FlexibleSolver(AbstractOperatorType& op,
                    const Opm::PropertyTree& prm,
-                   const std::function<VectorType()>& weightsCalculator = std::function<VectorType()>());
+                   const std::function<VectorType()>& weightsCalculator,
+                   std::size_t pressureIndex);
 
     /// Create a parallel solver (if Comm is e.g. OwnerOverlapCommunication).
     template <class Comm>
     FlexibleSolver(AbstractOperatorType& op,
                    const Comm& comm,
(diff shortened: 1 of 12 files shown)


blattms:
jenkins build this please

blattms:
jenkins build this please

atgeirr:
Thanks for addressing this issue. I just wonder if this is the time when we implement a more general facility, where we can attach arbitrary info and pass to the preconditioners and solvers, including at least the pressure index and the weight functions. What about passing a std::map<std::string, std::any>? The code that needs something extra looks up by name ("weight_function", "pressure_variable_index" etc.), and if there is no object of the required name, or it has the wrong type, then it throws an exception.

blattms:
Thanks for the input as this forced me to think about this.

What kind of external information are thinking of?

I actually think the current design of determining the weights is not perfect an maybe a relic of not having the `pressure_var_index`. Weights are constructed using the matrix and pressure index, and the parameters. All this information is available in the Factory or Solver. Hence is could and IMHO should be constructed there and not be passed from the outside in.

Hence pressure_var_index would be the only paramter in the map currently.

A bit unrelated: It seems when updating CPR we will use the old weights, which might not be optimal....

blattms:
> What kind of external information are thinking of?

If we want to calculate the weights inside the preconditioner it needs the whole simulator, sigh. Because to calculate the the trueimpes weights we need the whole simulator to construct ElementContext. On the bright side that has all the information we need (even the pressure index).

blattms:
jenkins build this please

blattms:
Will merge this tomorrow unless somebody barks. Refactoring postponed to next time.

PR #3369, opened 2021-06-10
handle GRUP in wellState from target

totto82:
This seems to help significantly for some key models. 
The code for getting the well part of the group target is at least at a first glance ripe for refactoring. The similar logic is used 4 places. But since this fix is urgent, and the refactoring nontrivial, I suggest that happens another day. 

I have tested this on the norne family and on a proprietary model.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell.hpp
+++ b/opm/simulators/wells/MultisegmentWell.hpp
@@ -118,6 +118,7 @@ namespace Opm
 
         /// updating the well state based the current control mode
         virtual void updateWellStateWithTarget(const Simulator& ebos_simulator,
+                                               const GroupState& group_state,
                                                WellState& well_state,
                                                DeferredLogger& deferred_logger) const override;
 
--- a/opm/simulators/wells/WellInterfaceFluidSystem.cpp
+++ b/opm/simulators/wells/WellInterfaceFluidSystem.cpp
@@ -25,13 +25,15 @@
 #include <opm/material/fluidsystems/BlackOilFluidSystem.hpp>
 
 #include <opm/parser/eclipse/EclipseState/Schedule/Well/WellTestState.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
 
 #include <opm/simulators/utils/DeferredLogger.hpp>
 #include <opm/simulators/wells/RateConverter.hpp>
 #include <opm/simulators/wells/ParallelWellInfo.hpp>
 #include <opm/simulators/wells/WellGroupHelpers.hpp>
 #include <opm/simulators/wells/WellState.hpp>
-
+#include <opm/simulators/wells/GroupState.hpp>
+#include <opm/simulators/wells/TargetCalculator.hpp>
 #include <ebos/eclalternativeblackoilindices.hh>
 
 #include <cassert>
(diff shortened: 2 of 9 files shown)


totto82:
jenkins build this please

totto82:
benchmark please

ytelses:
Benchmark result overview: 




View result details @ https://www.ytelses.com/opm/?page=result&id=1232

totto82:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 0.993    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 0.99     |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 1.014    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 1.02     |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.762    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 0.772    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1233

totto82:
jenkins build this please

opm/simulators/wells/WellInterfaceFluidSystem.cpp
```diff
@@ -916,6 +918,181 @@ flowPhaseToEbosPhaseIdx(const int phaseIdx) const
...
+    default:
+        // Should not be here.
+        assert(false);
+    }
+
+    auto currentGroupControl = group_state.injection_control(group.name(), injectionPhase);
+    if (currentGroupControl == Group::InjectionCMode::FLD ||
+        currentGroupControl == Group::InjectionCMode::NONE) {
+        if (!group.injectionGroupControlAvailable(injectionPhase)) {
+            return 1.0;
```
atgeirr, line 961: This, and also line 974, will give a target of 1 cubic meter per second, would it not be better to return an optional double, and set the target only if relevant? It looks like this may be taken from the production variant, which returns a scaling rather than a target (and a 1.0 scaling would be the same as doing nothing).

atgeirr (changes requested):
I think the logic is ok, I agree it should be refactored!

The 1.0 target should be fixed before merging.

opm/simulators/wells/MultisegmentWell.hpp
```diff
@@ -115,6 +115,7 @@ namespace Opm
 
         /// updating the well state based the current control mode
         void updateWellStateWithTarget(const Simulator& ebos_simulator,
```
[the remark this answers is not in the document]
totto82: Thanks for noticing See #3385

opm/simulators/wells/WellInterfaceFluidSystem.cpp
totto82: Good catch. It was a bit back-and-forth if the injector rates should be scaled or just set like we do now. Or will update as you suggest.

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
I have addressed the issues pointed out by @atgeirr  and fixed the critical test failures. The rest of the test failures are due to expected changes in the convergence and thus differences in the time-stepping. From my side this can be merged. I will created updated reference data later today if no objections.

totto82:

totto82:
jenkins build this opm-tests=564 please

joakim-hove:
OK - we seem to have a data race going - **you go first**!

totto82:
Ok, thanks.

PR #3373, opened 2021-06-11
 Add keyword validation using a predicate function

verveerpj:
The recently added keyword validation code could only check values of items against a list of options. This pull request adds the option to call a predicate function to test the value of an item. This allows for a broad range of tests, such as checking the sign of a value, or for  a range of values. The cost in complications of the specification of the validations is minor: one field is added to the struct used to define a check.

A test has been added that demonstrates this approach using a lambda function as a predicate.

In addition to this improvement, checking of items with double value types has been added.

**EDIT:** Decided to go for the simpler solution. Everything is done via a validator function, and a small helper class was added to be able to easily define lists of permitted values. Small change to the keyword definition structures that the user needs to write, but they are also clearer.

### DIFF
--- a/opm/simulators/flow/KeywordValidation.hpp
+++ b/opm/simulators/flow/KeywordValidation.hpp
@@ -22,6 +22,8 @@
 
 #include <opm/common/OpmLog/KeywordLocation.hpp>
 
+#include <functional>
+#include <initializer_list>
 #include <map>
 #include <optional>
 #include <string>
@@ -47,7 +49,7 @@ namespace KeywordValidation
     template <typename T>
     struct PartiallySupportedKeywordProperties {
         bool critical; // Set to true if the unsupported item value should be an error
-        std::vector<T> permitted_values; // The list of permitted values
+        std::function<bool(T)> validator; // Predicate function to test values
         std::optional<std::string> message; // An optional message to show if an illegal item is encountered
     };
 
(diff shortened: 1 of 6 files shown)


joakim-hove:
> This pull request adds the option to call a predicate function to test the value of an item

Yes!

verveerpj:
Jenkins build this please

verveerpj:
Jenkins build this please

opm/simulators/flow/KeywordValidation.hpp
```diff
@@ -121,9 +125,30 @@ namespace KeywordValidation
...
+    class permit
+    {
+    public:
+        permit(const std::initializer_list<T>& permitted_values) {
+            for(auto item: permitted_values) {
+                m_permitted.push_back(item);
+            }
+        }
+
+        bool operator()(const T value) const
```
joakim-hove, line 143: I guess the argument could be a `const T&`?

joakim-hove:
I think this is a good change. Maybe `allow_values<T>` is a better name than `permit<T>`?

opm/simulators/flow/KeywordValidation.hpp
verveerpj: yes. done.

verveerpj:
> I think this is a good change. Maybe `allow_values<T>` is a better name than `permit<T>`?

Agreed. done.

verveerpj:
Jenkins build this please

joakim-hove:
Wow - seems jenkins is case insensitive!?

PR #3377, opened 2021-06-14
PerfData maintains producre / injector status

joakim-hove:
Check injector/producer status in when creating perfdata and save some bytes


joakim-hove:
jenkins build this please

PR #3378, opened 2021-06-14
Quell some warnings in static analyzers

akva2:


akva2:
jenkins build this please

opm/core/props/phaseUsageFromDeck.cpp
```diff
@@ -135,7 +135,7 @@ PhaseUsage phaseUsage(const Phases& phases)
     pu.has_zFraction = phases.active(Phase::ZFRACTION);
     if (pu.has_zFraction) {
         pu.phase_pos[BlackoilPhases::ZFraction] = numActivePhases;
-        ++ numActivePhases;
+        //++ numActivePhases;
```
joakim-hove, line 138: If it is unnecessary it is unnecessary - delete!

bska: > If it is unnecessary it is unnecessary - delete!

This one stays.  I don't care if it's unnecessary or if static analyzers warn about it.  It makes the code easier to maintain if we add new enumerators to `BlackoilPhases`.

akva2: so you want me to drop the commit?

bska: Just to be clear: I want this particular change reverted and the increment reinstated.

bska: Just to be clear: I want this particular change reverted and the increment reinstated.

akva2:
jenkins build this please

PR #3380, opened 2021-06-14
Use GuideRate value instead of std::unique_ptr<GuideRate>

joakim-hove:
Small drive-by change while fighting with restart


joakim-hove:
jenkins build this please

bska:
Not all models have guide rates.

joakim-hove:
> Not all models have guide rates.

That is obviously true, but current master did not utilize pointer semantics to differentiate between models with and without guide rates - so unless there is a plan to actually allow empty guide rate pointers in the well model I would say the the current PR is a (quite minor) simplification by removing the uneccessary pointer semantics.

joakim-hove:
jenkins build this please

PR #3382, opened 2021-06-14
not comparing guideRate_ with nullptr

GitPaean:
it is not a pointer anymore since OPM/opm-simulators#3380

trying to fix the compilation.


GitPaean:
jenkins build this please

PR #3384, opened 2021-06-16
Check group limits in gas lift stage 1

hakonhagland:
Check group limits in gas lift stage 1 to avoid adding too much ALQ which must anyway later be removed in stage 2. This should make the optimization more efficient for small ALQ increment values. Also adds MPI support.

### DIFF
--- a/opm/simulators/wells/WellInterface.hpp
+++ b/opm/simulators/wells/WellInterface.hpp
@@ -41,7 +41,6 @@
 namespace Opm {
     template<typename TypeTag> class GasLiftSingleWell;
     template<typename TypeTag> class BlackoilWellModel;
-    class GasLiftGroupInfo;
 }
 #include <opm/simulators/wells/GasLiftGroupInfo.hpp>
 #include <opm/simulators/wells/GasLiftSingleWell.hpp>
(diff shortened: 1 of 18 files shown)


totto82:
Thanks.

totto82:
jenkins build this please

totto82:
The code looks ok. I have tested this on the relevant models (model5/ and a field model). And Jenkins is green. This is ready to be merged IMO after the new warnings are addressed:
https://ci.opm-project.org/job/opm-simulators-PR-builder/2597/gcc/new/folder.1225362760/

opm/simulators/wells/WellInterface.hpp
```diff
@@ -41,8 +41,11 @@
 namespace Opm {
     template<typename TypeTag> class GasLiftSingleWell;
     template<typename TypeTag> class BlackoilWellModel;
+    class GasLiftGroupInfo;
```
akva2, line 44: why? you include it two line further down, nothing needs to know between those two lines.

hakonhagland:
jenkins build this please

opm/simulators/wells/WellInterface.hpp
hakonhagland: Thanks! I have removed the line.

PR #3388, opened 2021-06-17
update all the guide rates for all the wells at the same time

GitPaean:
Downstream PR of https://github.com/OPM/opm-common/pull/2509


joakim-hove:
As commented in OPM/opm-common#2509 I think the decision whether to recalculate or not should be handled deep inside the `GuideRate` class, that way the API can be unchanged and this PR can actually be closed.

GitPaean:
> As commented in OPM/opm-common#2509 I think the decision whether to recalculate or not should be handled deep inside the GuideRate class, that way the API can be unchanged and this PR can actually be closed.

Let us try to do so. At least, make this PR as tiny as possible.

PR #3371, opened 2021-06-11
[WIP] Fixing guide rates

GitPaean:
This is the downstream PR for https://github.com/OPM/opm-common/pull/2502 . The description is in https://github.com/OPM/opm-common/pull/2502.

PR #3389, opened 2021-06-17
scale guiderates from si

totto82:
When combining guiderates form potentials, WGRUPCON and GCONINJE/GCONPROD the guiderates comes with different units. 
Testing indicates that using SI units may not be the preferred option. I mark it draft since more testing is needed. @GitPaean Does this effect your test case?

### DIFF
--- a/opm/simulators/wells/WellGroupHelpers.cpp
+++ b/opm/simulators/wells/WellGroupHelpers.cpp
@@ -290,6 +290,9 @@ namespace WellGroupHelpers
                                  "Invalid GuideRateInjTarget in updateGuideRatesForInjectionGroups",
                                  deferred_logger);
             }
+
+            const UnitSystem& unit_system = schedule.getUnits();
+            guideRateValue = unit_system.from_si(UnitSystem::measure::rate, guideRateValue);
             guideRate->compute(group.name(), phase, reportStepIdx, guideRateValue);
         }
     }
@@ -1307,6 +1310,10 @@ namespace WellGroupHelpers
         oilPot = comm.sum(oilPot);
         gasPot = comm.sum(gasPot);
         waterPot = comm.sum(waterPot);
+        const UnitSystem& unit_system = schedule.getUnits();
+        oilPot = unit_system.from_si(UnitSystem::measure::liquid_surface_rate, oilPot);
+        waterPot = unit_system.from_si(UnitSystem::measure::liquid_surface_rate, waterPot);
+        gasPot = unit_system.from_si(UnitSystem::measure::gas_surface_rate, gasPot);
         guideRate->compute(group.name(), reportStepIdx, simTime, oilPot, gasPot, waterPot);
     }
 
@@ -1344,6 +1351,10 @@ namespace WellGroupHelpers
             oilpot = comm.sum(oilpot);
             gaspot = comm.sum(gaspot);
             waterpot = comm.sum(waterpot);
+            const UnitSystem& unit_system = schedule.getUnits();
+            oilpot = unit_system.from_si(UnitSystem::measure::liquid_surface_rate, oilpot);
+            waterpot = unit_system.from_si(UnitSystem::measure::liquid_surface_rate, waterpot);
+            gaspot = unit_system.from_si(UnitSystem::measure::gas_surface_rate, gaspot);
             guideRate->compute(well.name(), reportStepIdx, simTime, oilpot, gaspot, waterpot);
         }
     }


totto82:
jenkins build this please

GitPaean:
> Testing indicates that using SI units may not be the preferred option. I mark it draft since more testing is needed. @GitPaean Does this effect your test case?

I do not know yet, but it is good if you can give some information about how it affects other cases. It will probably be something in the similar way.

totto82:
I think `9_4E_WINJ_GINJ_GUIDERATE_MSW.DATA` should be effected by this since you use the potentials in the calculations of the potential ratio. The ratios will be different if you use different units. But there is so many "conversion function" in the guiderate code that this may be taken case of already there.

totto82:
This PR does not change the results of `MOD4_GRP` where `GUIDERATE` is used. Actually I have not noticed any significant change in any of the tests. To really get the effect of this you need to combine `WGRUPCON` and potential based guiderates. I have made a test case together with @tskille that expose this combination and the results looks fine.

GitPaean:
> I think 9_4E_WINJ_GINJ_GUIDERATE_MSW.DATA should be effected by this since you use the potentials in the calculations of the potential ratio. The ratios will be different if you use different units. But there is so many "conversion function" in the guiderate code that this may be taken case of already there.

For this case, the next thing after https://github.com/OPM/opm-common/pull/2509 from my side is to do the scaling when calculating the guide rate. I mean the scaling happens the same time when we update the guide rate based on the well rates or well potential rates during the update time. Then for the next period, (30days for this case), the guide rates for all the phases stay const for the producers.  

Then it is possible the content in this PR will improve the results further. 

The observation is posted in https://github.com/OPM/opm-simulators/pull/3371#issuecomment-860270495 . The deck was modified based on 9_4E_WINJ_GINJ_GUIDERATE_MSW further to make the problem exposed further.

totto82:
I have investigated the test failures and tested this on different field models. From my end this is merge ready. If no objection I will start the update process of the reference solution later today.

totto82:

totto82:
jenkins build this opm-tests=562 please

bska:
New reference solutions have been installed on the CI system.  Merging this now.

GitPaean:
Hi, this PR changed the scale of the outputted guide rates, I think the unit conversion needs to be taken care of when outputting. 

reference results for one case, 
```
-- Days   dd/mm/yyyy        WOPGR:PROD1      WOPGR:PROD2      WOPGR:PROD3      WOPGR:PROD4       WGIGR:INJ1       WWIGR:INJ2 
-----------------------------------------------------------------------------------------------------------------------------
   0.00   01/11/2018                  0                0                0                0                0                0 
   1.00   02/11/2018                  0                0          20.9511                0      3.01766e+08                0 
   3.00   04/11/2018                  0                0          20.9511                0      3.01663e+08                0 
   5.00   06/11/2018                  0                0          20.9511                0      3.02012e+08                0 
   7.00   08/11/2018                  0                0          20.9511                0      3.02473e+08                0 
   9.00   10/11/2018                  0                0          20.9511                0      3.02951e+08                0 
  11.00   12/11/2018                  0                0          20.9511                0      3.03413e+08                0 
  13.00   14/11/2018                  0                0          20.9511                0      3.03846e+08                0 
  15.00   16/11/2018                  0                0          20.9511                0      3.04246e+08                0 
  17.00   18/11/2018                  0                0          20.9511                0      3.04612e+08                0 
  19.00   20/11/2018                  0                0          20.9511                0      3.04945e+08                0 
  21.00   22/11/2018                  0                0          20.9511                0      3.05245e+08                0 
  23.00   24/11/2018                  0                0          20.9511                0      3.05515e+08                0 
  25.00   26/11/2018                  0                0          20.9511                0      3.05757e+08                0 
  27.00   28/11/2018                  0                0          20.9511                0      3.05968e+08                0 
  29.00   30/11/2018                  0                0          20.9511                0       3.0616e+08                0 
  30.50   01/12/2018                  0                0          20.9511                0      3.06324e+08                0 
  32.00   03/12/2018                  0                0          17.1954                0       3.0644e+08                0 
  34.00   05/12/2018                  0          22.1157          17.1954                0      3.06545e+08                0 
  36.00   07/12/2018                  0          22.1157          17.1954                0       3.0636e+08                0 
  38.00   09/12/2018                  0          22.1157          17.1954                0      3.06377e+08                0 
  40.00   11/12/2018                  0          22.1157          17.1954                0       3.0642e+08                0 
  42.00   13/12/2018                  0          22.1157          17.1954                0      3.06438e+08                0 
  44.00   15/12/2018                  0          22.1157          17.1954                0      3.06303e+08                0 
  46.00   17/12/2018                  0          22.1157          17.1954                0      3.06052e+08                0 
  48.00   19/12/2018                  0          22.1157          17.1954                0      3.05783e+08                0 
```

Results with this PR
```
-- Days   dd/mm/yyyy        WOPGR:PROD1      WOPGR:PROD2      WOPGR:PROD3      WOPGR:PROD4       WGIGR:INJ1       WWIGR:INJ2 
-----------------------------------------------------------------------------------------------------------------------------
   1.00   02/11/2018                  0                0      1.79607e+06                0                0                0 
   3.00   04/11/2018                  0                0      1.79607e+06                0                0                0 
   5.00   06/11/2018                  0                0      1.79607e+06                0                0                0 
   7.00   08/11/2018                  0                0      1.79607e+06                0                0                0 
   9.00   10/11/2018                  0                0      1.79607e+06                0                0                0 
  11.00   12/11/2018                  0                0      1.79607e+06                0                0                0 
  13.00   14/11/2018                  0                0      1.79607e+06                0                0                0 
  15.00   16/11/2018                  0                0      1.79607e+06                0                0                0 
  17.00   18/11/2018                  0                0      1.79607e+06                0                0                0 
  19.00   20/11/2018                  0                0      1.79607e+06                0                0                0 
  21.00   22/11/2018                  0                0      1.79607e+06                0                0                0 
  23.00   24/11/2018                  0                0      1.79607e+06                0                0                0 
  25.00   26/11/2018                  0                0      1.79607e+06                0                0                0 
  27.00   28/11/2018                  0                0      1.79607e+06                0                0                0 
  29.00   30/11/2018                  0                0      1.79607e+06                0                0                0 
  31.00   02/12/2018                  0                0      1.79607e+06                0                0                0 
  32.00   03/12/2018                  0                0      1.47404e+06                0                0                0 
  34.00   05/12/2018                  0      1.89365e+06      1.47404e+06                0                0                0 
  36.00   07/12/2018                  0      1.89365e+06      1.47404e+06                0                0                0 
  38.00   09/12/2018                  0      1.89365e+06      1.47404e+06                0                0                0 
  40.00   11/12/2018                  0      1.89365e+06      1.47404e+06                0                0                0 
  42.00   13/12/2018                  0      1.89365e+06      1.47404e+06                0                0                0 
  44.00   15/12/2018                  0      1.89365e+06      1.47404e+06                0                0                0 
  46.00   17/12/2018                  0      1.89365e+06      1.47404e+06                0                0                0 
  48.00   19/12/2018                  0      1.89365e+06      1.47404e+06                0                0                0 
```

I think an extra `86400.` was applied when outputting.

GitPaean:
I would like very much to get the output problems regarding to the guide rate output fixed. I think either we do not apply the unit conversion to the guide rates anymore when outputting, (because it is already in the proper output unit already if I understand correctly), otherwise we convert `WGRUPCON`  specified guide rate to be `SI` internally, then everything is in `SI`, which is the original way.  I prefer the latter way, because it looks like a more consistent manner to the output code (also considering the situation output well potential values), but it might oversee some problems, for example converting `ORAT` guide rate to `GRAT` guide rate, which might be different with different unit systems.  Please suggest if you find something missing here. 

Without fixing this, we have a different output from the reference simulator when coming to guide rates output . (`86400. factor different`. ).

totto82:
Can you check if

--- a/src/opm/output/eclipse/Summary.cpp
+++ b/src/opm/output/eclipse/Summary.cpp
@@ -1391,7 +1391,9 @@ inline quantity well_control_mode( const fn_args& args )
 template <Opm::data::GuideRateValue::Item i>
 quantity guiderate_value(const ::Opm::data::GuideRateValue& grvalue)
 {
-    return { !grvalue.has(i) ? 0.0 : grvalue.get(i), rate_unit<i>() };
+    const auto unit = Opm::UnitSystem::measure::identity;
+    // the guide rates already has the currect unit
+    return { !grvalue.has(i) ? 0.0 : grvalue.get(i), unit };
 }

does the trick. Due to the converting we may want to keep the non SI unit in the simulator. But I agree that it violates the standard assumption that everything is SI internally.

bska:
> Can you check if [no unit conversion] does the trick.

Doing that will probably output the correct **value**, but it will output the wrong unit of measurement to the `SMSPEC` file.  In other words, postprocessors won't have access to the appropriate unit string when plotting the guiderates as time series.

totto82:
Is there a simple way of passing the correct unit but still not scale the value or do we need to re-scale it back to SI before re-scaling it to the correct unit again

bska:
> Is there a simple way of passing the correct unit but still not scale the value

Not unless we want to essentially rewrite all of the summary output module.  That module assumes that all of its input values are strict SI.

GitPaean:
When calculating the guide rate, can we convert the unit when calculating, because we need the ratio correct during the calculating, then we convert the calculated guide rate back to SI. 

If the scale is not correct comparing with `WGRUPCON`, it means the `WGRUPCON` values need to be converted to SI values.  Is there anything missing here?

I think with this way, we do not need to touch the output module, which just takes care of so many different values, kind of hard to touch.

totto82:
@GitPaean I am fine with your solution. Do you have time to make a PR?

GitPaean:
> @GitPaean I am fine with your solution. Do you have time to make a PR?

Actually I am not the best person to do it, because I do not know what situation/case it breaks and I could not test it. You probably can do it much more efficiently than me since you are familiar with the matter and have the test case ready. @totto82 .

GitPaean:
@totto82 , if you are too busy, let me know. I can try to make the efforts to create a draft. I would like to get it fixed to ease the debugging process.

GitPaean:
It looks like that we do not have a case in opm-tests that uses WGRUPCON keyword.

ducbueno:
Hello.

Any updates on the usage of the WGRUPCON keyword?

I'm running a June 2022 build of Flow on a proprietary model which contains the WGRUPCON keyword and I'm getting a strange error message saying I'm passing it 6 items when its clearly only 5 being passed (as required in Flow's manual).

PR #3395, opened 2021-06-21
Partially Supported Keywords RUNSPEC and some Aquifer Keywords

OPMUSER:
Partially Supported all RUNSPEC keywords, plus AQUANCON, AQUCON and AQUCT

### DIFF
--- a/opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
+++ b/opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
@@ -1,81 +1,430 @@
-/*
-  Copyright 2021 Equinor.
-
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

akva2:
looks like this changes the entire file to typewriter compatibility...

OPMUSER:
I'm going to re-do this so I would like to delete this pull request. I 
actually checked the range values of items that we did support instead 
of only checking items that we do not support - apologies.

akva2:
sure. and just in case: don't take my pointing out issues as hostility, i'm only trying to help.

PR #3397, opened 2021-06-21
Report Pressure Dependent Pore Volume in PRT File

bska:
This PR distinguishes the reference condition pore volume from the dynamic, pressure (and/or temperature) dependent pore volume value.  Previously we would report the latter as the `PORV` value in the "Field Totals" and "FIPNUM region" reports, but this commit switches to reporting the former instead&mdash;mostly for compatibility. We still report the dynamic pore volume value, but now we report this on a line of its own, before the table, using one of the forms
```
Field total pressure dependent pore volume = 12345 RM3
FIPNUM report region 1 pressure dependent pore volume = 123 RM3
```

### DIFF
--- a/ebos/eclgenericoutputblackoilmodule.cc
+++ b/ebos/eclgenericoutputblackoilmodule.cc
@@ -38,6 +38,7 @@
 #include <opm/parser/eclipse/Units/Units.hpp>
 
 #include <cassert>
+#include <initializer_list>
 #include <iomanip>
 #include <sstream>
 #include <stdexcept>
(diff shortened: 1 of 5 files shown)


bska:
Note: This PR depends on the new `DynPoreVolume` slot introduced to the `Inplace` facility in OPM/opm-common#2522.  I will keep it in "draft" state until that PR is approved and installed in the master branch.

bska:
jenkins build this opm-common=2522 please

joakim-hove:
I originally looked briefly on this - and my guess was that the difference was due to minor differences in the cell volume calculations in opm-common and those used in the simulator (dune); whereas this solution seems to be based on dynamic porosity - i.e. rock compressibility?

bska:
> my guess was that the difference was due to minor differences in the cell volume calculations in opm-common and those used in the simulator

That was my initial thought too, but that did not turn out to be the case. I was finally able to trace the root cause back to using [dynamic porosity](https://github.com/OPM/opm-simulators/blob/57aa8d0384c70c09e60ee7cc95fd7f965a239b04/ebos/ecloutputblackoilmodule.hh#L628) values when deriving the `Inplace::Phase::PoreVolume`:
```C++
this->fip_[Inplace::Phase::PoreVolume][globalDofIdx] = pv;
````
This PR switches to using [reference porosities](https://github.com/OPM/opm-simulators/blob/42824765764fdfc0a29c1f18cd2d1e2c6ad1d0b2/ebos/ecloutputblackoilmodule.hh#L634-L635) instead.  Doing so makes us match ECLIPSE to the last digit.

**Comment by bska** [comment]
Upstream PR installed in master branch.  I'll mark this as "ready for review".

**Comment by bska** [comment]
jenkins build this please

File: ebos/eclgenericoutputblackoilmodule.cc [hunk 1]
```diff
@@ -1133,11 +1123,25 @@ outputRegionFluidInPlace_(std::unordered_map<Inplace::Phase, Scalar> oip,
         return;
 
     // don't output FIPNUM report if the region has no porv.
-    if (cip[Inplace::Phase::PoreVolume] == 0)
+    if (! (cip[Inplace::Phase::PoreVolume] > Scalar{0}))
```
joakim-hove, line 1126: I would *really* prefer the microsteps to go away from the `Scalar & ScalarBuffer`rather than the opposite.

joakim-hove:
Hmmm - looks like you will not join forces in the anti `Scalar` crusade - fair enough. I'll merge.

bska:
> looks like you will not join forces in the anti `Scalar` crusade

I'll join you, but we need to do it in a more holistic manner.  We can't just pick a single function in the middle of a call stack and enforce `Scalar == double` there.  Rather, we need to start at the leaves and work back towards the root.  At some point that'll entail making all data members `vector<double>` in one go.

PR #3398, opened 2021-06-21
[WIP] Add FIPRESV

goncalvesmachadoc:


ebos/ecloutputblackoilmodule.hh
```diff
@@ -155,6 +168,8 @@ public:
     {
         EWOMS_REGISTER_PARAM(TypeTag, bool, ForceDisableFluidInPlaceOutput,
                              "Do not print fluid-in-place values after each report step even if requested by the deck.");
+        EWOMS_REGISTER_PARAM(TypeTag, bool, ForceDisableResvFluidInPlaceOutput,
```
joakim-hove, line 171: I don't think we need this as a parameter

goncalvesmachadoc:
Started the draft to compute reservoir fluid-in-place from surface fluid-in-place multiplied by
formation volume factors. But Eclipse 2010.1 and forward changed to calculate it as pore volume multiplied by saturation. So, will modify accordingly.

goncalvesmachadoc:
Depends on https://github.com/OPM/opm-common/pull/2628

goncalvesmachadoc:
image

PR #3400, opened 2021-06-22
Add missing header file to GasLiftGroupInfo.cpp

hakonhagland:
`GasLiftGroupInfo.cpp` did not include `config.h` which caused the macro `HAVE_MPI` to be undefined which again caused the file to be compiled with `Dune::Communication<No_Comm>` instead of with `Dune::Communication<MPI_Comm>`. This later caused linking problems with files that used MPI.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -872,9 +872,10 @@ namespace Opm {
                 ebosSimulator_.model().newtonMethod().numIterations(),
                 phase_usage_,
                 deferred_logger,
-                this->wellState()
+                this->wellState(),
+                ebosSimulator_.vanguard().grid().comm()
             };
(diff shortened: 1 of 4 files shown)


hakonhagland:
jenkins build this please

bska:
Just for the record, this PR does more than just add `<config.h>`.  Is that intentional?

hakonhagland:
> this PR does more than just add <config.h>

@bska Yes, I forgot to mention that.  It also refactors back the dependency on the MPI communicator from the header file `GasLiftGroupInfo.hpp` to the .cpp file. The dependency was originally refactored out of the .cpp file to work around the linker errors (due to the missing `config.h` header). Now as the problem has been found it can be refactored back into the .cpp file.

bska:
> Now that the [linker] problem has been found it can be refactored back into the `.cpp` file.

I see.  In that case this looks good to me and I'll merge into master.

PR #3358, opened 2021-06-07
Chase Type Specific Aquifer Data API Change 

bska:
This PR switches to using the new `typeData` interface for representing type-specific aquifer data items introduced in OPM/opm-common#2496.  In particular we use the new `typeData.is<>()` and `typeData.get<>()` member functions to query and access the data that is specific to each aquifer type (e.g., Carter-Tracy or numerical).

While here, also expand the reported data items for numerical aquifers to one initial pressure value for each aquifer cell.  This is needed for restart purposes.

PR #3402, opened 2021-06-22
Load Analytic Aquifers From Restart File

bska:
This PR activates restart support for analytical aquifers. Analytic aquifer keywords do not exist in the input file for restarted runs since the 'SOLUTION' section is empty apart from from the `RESTART` specification.

We use the `RestartIO::Aquifer` (OPM/opm-common#2526) class to load the information from the restart file and then assign this information to the `AquiferConfig` object hosted by the `EclipseState`.  Loading `Schedule` information from the restart file is still contingent on the `--sched-restart=false` command line setting.  We support all aquifer types currently supported by the simulator, namely

* Carter&ndash;Tracy
* Fetkovich
* Numerical

Numerical aquifer restart support is limited to restoring the cumulative water production and initial aquifer cell pressures from the `RAQN` vector.  On the other hand, numerical aquifers are specified in the `GRID` section so we always have access to that information in a restarted run.  Analytical aquifers must always be constructed purely from the restart file information in restarted runs.

We take some care to account for multiple processes sharing a single aquifer when restarting a simulation run.  In the case of analytical aquifers we apportion the cumulative production based on the fraction of the total inflow area that is "owned" by each MPI process.  This is, at best, a very crude approximation.  In the **very** special case that all aquifer connections produce the same flow rates then this will nevertheless be an okay approximation.  Full fidelity apportionment will require tracking the cumulative production at the connection level which in turn will probably entail relinquishing complete ECLIPSE compatibility in the aquifer restart vectors.

### DIFF
--- a/opm/simulators/utils/readDeck.cpp
+++ b/opm/simulators/utils/readDeck.cpp
@@ -38,6 +38,7 @@
 #include <opm/output/eclipse/RestartIO.hpp>
 #include <opm/io/eclipse/ERst.hpp>
 #include <opm/io/eclipse/RestartFileView.hpp>
+#include <opm/io/eclipse/rst/aquifer.hpp>
 #include <opm/io/eclipse/rst/state.hpp>
 
 #include <opm/parser/eclipse/Deck/Deck.hpp>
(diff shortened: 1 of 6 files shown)


bska:
Note: This PR depends on PRs OPM/opm-common#2526 and #3358.  I will keep this in "draft" mode until those are merged into the master branch.

opm/simulators/utils/readDeck.cpp
```diff
@@ -225,31 +227,63 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
...
+                // not exist in the input deck in this case.  In other
+                // words, there's no way to check if there really are
+                // analytic aquifers in the run until we attempt to read the
+                // specifications from the restart file.  If the loader
+                // determines that there are no analytic aquifers, then
+                // 'getAnalyticAquifers()' does nothing.
+                const int report_step = init_config.getRestartStep();
+                const auto rst_filename = eclipseState->getIOConfig().getRestartFileName( init_config.getRestartRootName(), report_step, false );
+                auto rst_file = std::make_shared<Opm::EclIO::ERst>(rst_filename);
+
```
joakim-hove, line 244: I have of course not looked in detail on all of this - but I would really like some structural change:

1. I would prefer if the `RstState::load()` was brought out to the current scope and the `RstState` "class" gets the basic data-transfer support for Aquifers.
2. When it comes to the Aquifer configuration information the `EclipseState` class is the "main manager" - so rather than setting aquifer configuration on the `EclipseState` with:
```
aquifers.getAnalyticAquifers( eclipseState->mutableAquifers() );
```
I would prefer if this was implemented with:
```
EclipseState::load_rst_aquifers( RstState::aquifers );
```

bska:
jenkins build this opm-common=2526 please

bska:
jenkins build this opm-common=2526 please

bska:
jenkins build this opm-common=2526 please

bska:
jenkins build this opm-common=2526 please

bska:
jenkins build this opm-common=2526 please

bska:
jenkins build this opm-common=2526 please

bska:
jenkins build this opm-common=2526 please

PR #3404, opened 2021-06-23
PartiallySupportedKeywords for RUNSPEC and GRID Sections

OPMUSER:
RUNSPEC and GRID keywords for partially supported keywords

### DIFF
--- a/opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
+++ b/opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
@@ -17,7 +17,7 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 
   Generated : keyw-code.py
-  Date      : 2021-06-23 12:54:52
+  Date      : 2021-06-25 13:02:39
 */
 
 #include <opm/simulators/utils/PartiallySupportedFlowKeywords.hpp>
@@ -46,8 +46,8 @@ partiallySupported()
          {
             "EDITNNC",
             {
-               {12,{false, allow_values<std::string> {}, "FACE1 not supported use 1* - will continue"}}, // VE_FACE1
-               {13,{false, allow_values<std::string> {}, "FACE2 not supported use 1* - will continue"}}, // VE_FACE2
+               {12,{false, allow_values<std::string> {}, "FACE1 not supported use 1* - will continue"}}, // FACE_FLOW12
+               {13,{false, allow_values<std::string> {}, "FACE2 not supported use 1* - will continue"}}, // FACE_FLOW21
             },
          },
          {
@@ -93,19 +93,19 @@ partiallySupported()
          {
             "MULTIREG",
             {
-               {4,{false, allow_values<std::string> {"F", "M", "O"}, "REGION_NAME must equal to F/M or O"}}, // REGION_NAME
+               {4,{false, allow_values<std::string> {"F", "M", "O"}, "REGION_NAME must equal to F/M or O"}}, // NOT_DEFINED
             },
          },
          {
             "MULTREGP",
             {
-               {3,{false, allow_values<std::string> {"F", "M", "O"}, "REGION_NAME must equal to F/M or O"}}, // REGION_NAME
+               {3,{false, allow_values<std::string> {"F", "M", "O"}, "REGION_NAME must equal to F/M or O"}}, // REGION_TYPE
             },
          },
          {
             "MULTREGT",
             {
-               {6,{false, allow_values<std::string> {"F", "M", "O"}, "REGION_NAME must equal to F/M or O"}}, // REGION_NAME
+               {6,{false, allow_values<std::string> {"F", "M", "O"}, "REGION_NAME must equal to F/M or O"}}, // REGION_DEF
             },
          },
          {
@@ -139,6 +139,13 @@ partiallySupported()
             },
          },
          {
+            "TABDIMS",
+            {
+               {20,{false, allow_values<std::string> {}, "NOTUSED should be defaulted (1*) - ignored as not used"}}, // ITEM20_NOT_USED
+               {25,{false, allow_values<std::string> {}, "RESVED should be defaulted (1*) - ignored as not used"}}, // RESERVED
+            },
+         },
+         {
             "UDQDIMS",
             {
                {11,{false, allow_values<std::string> {"N"}, "RSEED option is not supported – value ignored"}}, // RESTART_NEW_SEED
@@ -155,19 +162,12 @@ partiallySupported()
 {
    static const KeywordValidation::PartiallySupportedKeywords<int>partially_supported_keywords_int = {
          {
-            "AQUCON",
-            {
-               {11,{false, allow_values<int> {1}, "VEOPT1 Vertical Equilibrium option one– not used"}}, // VEFRAC
-               {12,{false, allow_values<int> {1}, "VEOPI2 Vertical Equilibrium option two– not used"}}, // VEFRACP
-            },
-         },
-         {
             "EDITNNC",
             {
-               {8,{false, allow_values<int> {0}, "ISATNUM1 only default value of 0 supported – will continue"}}, // SIM_DEPENDENT1
-               {9,{false, allow_values<int> {0}, "ISATNUM2 only default value of 0 supported – will continue"}}, // SIM_DEPENDENT2
-               {10,{false, allow_values<int> {0}, "IPRSNUM1 only default value of 0 supported – will continue"}}, // PRESSURE_TABLE1
-               {11,{false, allow_values<int> {0}, "IPRSNUM2 only default value of 0 supported – will continue"}}, // PRESSURE_TABLE2
+               {8,{false, allow_values<int> {0}, "ISATNUM1 only default value of 0 supported – will continue"}}, // SAT_TABLE12
+               {9,{false, allow_values<int> {0}, "ISATNUM2 only default value of 0 supported – will continue"}}, // SAT_TABLE21
+               {10,{false, allow_values<int> {0}, "IPRSNUM1 only default value of 0 supported – will continue"}}, // PRESS_TABLE12
+               {11,{false, allow_values<int> {0}, "IPRSNUM2 only default value of 0 supported – will continue"}}, // PRESS_TABLE21
             },
          },
          {
@@ -202,7 +202,7 @@ partiallySupported()
          {
             "NETWORK",
             {
-               {3,{false, allow_values<int> {}, "NBCMAX option is not used – value ignored"}}, // NBCMAX
+               {3,{false, allow_values<int> {20}, "NBCMAX option is not used and should be defaulted– value ignored"}}, // NBCMAX
             },
          },
          {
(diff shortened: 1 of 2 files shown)


opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
```diff
@@ -27,55 +30,309 @@ template <>
 const PartiallySupportedKeywords<std::string>&
 partiallySupported()
 {
-    static const PartiallySupportedKeywords<std::string> partially_supported_keywords_strings = {
-        {
+   static const PartiallySupportedKeywords<std::string> partially_supported_keywords_strings = {
+         {
+            "BRINE",
+            {
+               {1,{false, allow_values<std::string> {}, "SALTS Multi-Component Brine model not supported"}}, // NO_NAME
```
joakim-hove, line 37: How should an empty list of allowed values be interpreted?

joakim-hove:
jenkins build this please

opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
OPMUSER: An empty list implies that no values are accepted.

OPMUSER: Note some of the default values in the jason definitions are incorrect based on the commercial simulator values (TABDIMS for example). I have used the "correct" values so there will be some superfluous warning message until we amend the jason definitions. I can do this if you like, but l would like to have a quick discussion first.

joakim-hove: > An empty list implies that no values are accepted.

OK - I understand; another way of putting is: "This item must be defaulted"? Instead of outputting an empty list in this situation I would say a better solution would be to create another datastructure with all the keywords & item combinations which *must be defaulted*; that will require a minor update to the framework though.

When it comes to the `BRINE` keyword specifically that is a story of its own - there are some issues with the json definition there.

joakim-hove: > Note some of the default values in the jason definitions are incorrect based on the commercial simulator values (TABDIMS for example).

We can discuss - but I see no reason not to just go along and update.

alfbr: > OK - I understand; another way of putting it is: "This item must be defaulted"? Instead of outputting an empty list in this situation I would say a better solution would be to create another datastructure with all the keywords & item combinations which _must be defaulted_; that will require a minor update to the framework though.

From my end (a user of the framework) this sounds more complicated, I favour the current approach. Please keep in mind that the framework is not for listing supported options, it is for listing unsupported options. As long as it is easy to find the  meaning of an empty list (e.g., explaining the syntax at the top of the file), I am fine with it. If not, then my preference would rather be to use a symbol. For instance a star * instead of an empty list to indicate that all options are unsupported. I believe we should thrive for supporting the default option in any supported keyword, so that it can be taken for granted (not sure we are there now, but maybe?).

OPMUSER: > OK - I understand; another way of putting it is: "This item must be defaulted"? Instead of outputting an empty list in this situation I would say a better solution would be to create another datastructure with all the keywords & item combinations which must be defaulted; that will require a minor update to the framework though.

What about putting this in the jason file instead and try to consolidate everything in these files eventually. At the moment we have PartiallySupportedFlowKeywords.cpp and UnsupportedFlowKeywords.cpp. What do you think?

Also is it okay to add another parameter to the jason files, item=1, etc. to keep count of the number of variables, I would only do this for the keywords with lots of variables like TABDIMS and REGDIMS..

joakim-hove: > What about putting this in the jason file instead and try to consolidate everything in these files eventually. At the moment we have PartiallySupportedFlowKeywords.cpp and UnsupportedFlowKeywords.cpp. What do you think?

In many ways I would have liked to sprinkle information like this through the json scheme, the problem with that is that the json scheme does not know what the simulator supports, so that would be layering violation as some guy used to say.

> Also is it okay to add another parameter to the jason files, item=1, etc. to keep count of the number of variables

Yes - that is OK. They must be entered with correct json syntax - but will otherwise be ignored.

alfbr: > What about putting this in the jason file instead and try to consolidate everything in these files eventually. 

I am afraid this breaks with the intent of opm-simulators. The parser should not need to care what the simulators support, the parser is actually designed to support all keywords and options. I also fail to see how moving everything to the parser will make life easier for those who will maintain PartiallySupportedFlowKeywords.cpp and UnsupportedFlowKeywords.cpp.

OPMUSER: > If not, then my preference would rather be to use a symbol. For instance a star * instead of an empty list to indicate that all options are unsupported. I believe we should thrive for supporting the default option in any supported keyword, so that it can be taken for granted (not sure we are there now, but maybe?)

I like `*` or `1*`to indicate that all options are unsupported.

alfbr: > I like `*` or `1*`to indicate that all options are unsupported.

Great, sounds like a plan! But please do not hold this PR for it, we can easily update these occurrences from empty to stars once we get the improvement in the framework.

joakim-hove: Ehhh no - this is not a viable plan within the scope of the current framework. The current framework is based on C++ *code* - and that limits our ability to choose syntax ourselves. In order to allow syntax like `*`or `1*` to indicate "only default allowed" we will have to change to use data (i.e. strings of some kind) to configure this.

alfbr: > Ehhh no - this is not a viable plan within the scope of the current framework. The current framework is based on C++ _code_ - and that limits our ability to choose syntax ourselves. In order to allow syntax like `*`or `1*` to indicate "only default allowed" we will have to change to use data (i.e. strings of some kind) to configure this.

Ah, you are right. We need to think a bit more about this then. But great if we can go with the current solution in the meantime.

OPMUSER: What about a numeric constant -999.999 or -1.0E-20, for example

alfbr: > What about a numeric constant -999.999 or -1.0E-20, for example

Not sure about others, but to me an empty list is actually more intuitive then.

opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
```diff
@@ -27,55 +30,309 @@ template <>
...
+const KeywordValidation::PartiallySupportedKeywords<int>&
 partiallySupported()
 {
-    static const PartiallySupportedKeywords<int> partially_supported_keywords_int = {
-        {
+   static const KeywordValidation::PartiallySupportedKeywords<int>partially_supported_keywords_int = {
+         {
+            "AQUCON",
+            {
+               {11,{false, allow_values<int> {1}, "VEOPT1 Vertical Equilibrium option one– not used"}}, // VEFRAC
```
joakim-hove, line 160: These should be items 12 and 13.

opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
bska: > The current framework is based on C++ _code_ - and that limits our ability to choose syntax ourselves. In order to allow syntax like `*`or `1*` to indicate "only default allowed" we will have to change to use data (i.e. strings of some kind) to configure this.

Don't we already support function-like predicates in this position?  How about
```C++
struct default_only
{
    bool operator()(const std::string& x) const
    {
        return x == "*";
    }
};
```
or whatever we need to do to match defaulted values.  Then the specification would be something along the lines of
```C++
{1, {false, default_only{}, "SALTS ..."}}
```

joakim-hove: Yes we support function predicates, but the desire is to have a literal `*` in the code - not compare string data with `*`

bska: > Yes we support function predicates, but the desire is to have a literal `*` in the code

Right, but since that's not possible we need a semantically equivalent alternative that can be expressed in C++ source code.

> - not compare string data with `*`

What is the appropriate criterion then?  I'm starting to lean towards passing the `DeckItem` as an argument to the predicate rather than `DeckItem::get<T>()`.  That way, `default_only` (or however we choose to name it) can return
`item.defaultApplied(0)` instead of trying to match the value in an intelligent way.  The predicates that **do** need to look at the values can call `DeckItem::get<T>()` internally.

joakim-hove: The current semi-good solution is to pass `allow_values<T>{}` as an indirect way of saying anything but `*` in input is illegal (mind that this will fail to validate even input where the default value is entered explicitly).

The two alternatives I have thought of are:

1. Add a new datastructure with {keyword,item} pairs which *must be defaulted*.
2. Pass in `DeckItem&` instead of `DeckItem.get<T>(0)`

Alternative 2 will certainly give the most flexibility for this and possible future scenarios; however that will require immediate significant refactoring to the framework. I can discuss and assist on this - but I will not drive it in any direction.

bska: > The current semi-good solution is to pass `allow_values<T>{}` as an indirect[1] way of saying anything

It's rather a way of saying **nothing**.  The `allow_values<>` [predicate](https://github.com/OPM/opm-simulators/blob/c101ed10f4f69127311e71f32c8c6e9e28ef1367/opm/simulators/flow/KeywordValidation.hpp#L144-L147) matches `DeckItem::get<>(0)` against a list of permissible values.  If that list is empty, then any value at all will fail to match and the input will be deemed invalid.

I still think that passing the `DeckItem` as an argument is the better solution.  Note that any specification that uses `allow_values<>` will be automatically updated once we change the function call operator of that type.  We could even just add a new function call operator of the form
```C++
bool operator()(const DeckItem& item) const
{
    return (*this)(item.get<T>(0));
}
```
The downside is that we would need to include `DeckItem.hpp` into the `KeywordValidation.hpp` header.

opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
```diff
@@ -15,6 +15,9 @@
 
   You should have received a copy of the GNU General Public License
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+
+  Generated : keyw-code.py
+  Date      : 2021-06-25 13:02:39
```
joakim-hove, line 20: The license & Copyright header should be "clean" - the
```
Generated : keyw-code.py
Date      : 2021-06-25 13:02:39
```
should be removed

joakim-hove:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
Thanks a lot for the updates.  I'll merge this now.

PR #3407, opened 2021-06-24
Potentially reduce gaslift when under well is under group control

totto82:
Current implemention only support ORAT and GRAT controlled groups

### DIFF
--- a/opm/simulators/wells/GasLiftGroupInfo.hpp
+++ b/opm/simulators/wells/GasLiftGroupInfo.hpp
@@ -171,7 +171,7 @@ private:
     GroupIdxMap group_idx_;
     int next_group_idx_ = 0;
     // Optimize only wells under THP control
-    bool optimize_only_thp_wells_ = true;
+    bool optimize_only_thp_wells_ = false;
 
 };
 
--- a/opm/simulators/wells/GasLiftSingleWellGeneric.cpp
+++ b/opm/simulators/wells/GasLiftSingleWellGeneric.cpp
@@ -26,6 +26,8 @@
 #include <opm/simulators/utils/DeferredLogger.hpp>
 #include <opm/simulators/wells/GasLiftWellState.hpp>
 #include <opm/simulators/wells/WellState.hpp>
+#include <opm/simulators/wells/GroupState.hpp>
+
 
 #include <fmt/format.h>
 
@@ -38,6 +40,7 @@ namespace Opm
 GasLiftSingleWellGeneric::GasLiftSingleWellGeneric(
     DeferredLogger& deferred_logger,
     WellState& well_state,
+    const GroupState& group_state,
     const Well& ecl_well,
     const SummaryState& summary_state,
     GasLiftGroupInfo &group_info,
@@ -47,6 +50,7 @@ GasLiftSingleWellGeneric::GasLiftSingleWellGeneric(
 ) :
     deferred_logger_{deferred_logger}
     , well_state_{well_state}
+    , group_state_{group_state}
     , ecl_well_{ecl_well}
     , summary_state_{summary_state}
     , group_info_{group_info}
@@ -93,10 +97,10 @@ calcIncOrDecGradient(double oil_rate, double gas_rate, double alq, bool increase
         return std::nullopt;
     double new_alq = *new_alq_opt;
     if (auto bhp = computeBhpAtThpLimit_(new_alq)) {
-        auto [new_bhp, bhp_is_limited] = getBhpWithLimit_(*bhp);
+        auto new_bhp = getBhpWithLimit_(*bhp);
         // TODO: What to do if BHP is limited?
         std::vector<double> potentials(this->num_phases_, 0.0);
-        computeWellRates_(new_bhp, potentials);
+        computeWellRates_(new_bhp.first, potentials);
         auto [new_oil_rate, oil_is_limited] = getOilRateWithLimit_(potentials);
         auto [new_gas_rate, gas_is_limited] = getGasRateWithLimit_(potentials);
         if (!increase && new_oil_rate < 0 ) {
@@ -125,7 +129,10 @@ runOptimize(const int iteration_idx)
             state = runOptimize2_();
         }
         if (state) {
-            if (state->increase()) {
+            // NOTE: that state->increase() returns a std::optional<bool>, if
+            //  this is std::nullopt it means that we was not able to change ALQ
+            //  (either increase or decrease)
+            if (state->increase()) {  // ALQ changed..
                 double alq = state->alq();
                 if (this->debug_)
                     logSuccess_(alq, iteration_idx);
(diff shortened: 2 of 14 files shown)


totto82:
jenkins build this please

opm/simulators/wells/GasLiftGroupInfo.hpp
```diff
@@ -171,7 +171,7 @@ class GasLiftGroupInfo
     GroupIdxMap group_idx_;
     int next_group_idx_ = 0;
     // Optimize only wells under THP control
-    bool optimize_only_thp_wells_ = true;
+    bool optimize_only_thp_wells_ = false;
```
totto82, line 175: This significantly slows down the well assembly time

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

opm/simulators/wells/GasLiftGroupInfo.hpp
totto82: I have made a small fix that reduce most of the extra time. Most of the time is spent in `computeWellRatesWithBhp` Currently the code passes evaluation objects of size 7 to this function. Since the derivatives are not needed for this call this leads to a significant uneccesary overhead in the computational time. Improving this shouldn't be too hard. I will put it on my TODO list. But probably not before summer.

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
Any comments. I plan to merge this later today.

bska:
> Any comments. I plan to merge this later today.

The code itself looks fine to me, but I'm concerned about merging something that's advertised as

>> _This significantly slows down the well assembly time_

I think we need to benchmark that before we merge it, especially if this is going to be the default setting for the summer.

bska:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1.001    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 1.02     |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.997    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 1        |



View result details @ https://www.ytelses.com/opm/?page=result&id=1255

totto82:
> I think we need to benchmark that before we merge it, especially if this is going to be the default setting for the summer.
It only hit models with gaslift. i.e. none of the models benchmarked. I did some more checks yesterday and notices some discrepancies I would like to check out first. So hold on merging....

totto82:
jenkins build this please

bska:
> It only hit models with gaslift.

Understood.  Please address the new [shadow](https://ci.opm-project.org/job/opm-simulators-PR-builder/2680/gcc/new/source.e2cd077e-ae87-4a9c-a374-c812c74422a0/#1186) [warnings](https://ci.opm-project.org/job/opm-simulators-PR-builder/2680/gcc/new/source.119f0771-3bef-4ba3-ad3b-79a45523cb7d/#1207) before merging.

totto82:
```
/home/thsa/workspace/opm/opm-simulators/opm/simulators/wells/GasLiftSingleWellGeneric.cpp:586:34: warning: unused variable ‘bhp_is_limited’ [-Wunused-variable]
         auto [bhp, bhp_is_limited] = getBhpWithLimit_(*bhp_opt);
```
Whats the best way to solve this unused issues. I notices a couple of them around in the gaslift code.

GitPaean:
> Whats the best way to solve this unused issues. I notices a couple of them around in the gaslift code.

Maybe you can try
```c++
[[maybe_unused]] auto...
```

bska:
> Whats the best way to solve this unused issues. I notices a couple of them around in the gaslift code.

If it really isn't used at all then I'd probably **not** use structured bindings here:
```C++
const auto bhp = this->getBhpWithLimit_(bhp_opt.value());
```
and then access `bhp.first` or `bhp.second` as needed.

totto82:
jenkins build this please

bska (approved):
Thanks for the updates.  I'll merge into master now.

hakonhagland:
Hmm.. seems like something went wrong with [the previous comment](https://github.com/OPM/opm-simulators/pull/3407#commitcomment-62833020). I wrote it by clicking the "+" sign to the left of line 1193 in `GasLiftSingleWellGeneric.cpp` in the [files view](https://github.com/OPM/opm-simulators/pull/3407/files) of the current PR. This has previously worked well to provide a context (file name and line number) for a comment, but here it seems like that information has been lost (the comment seems to have no context) ? I am not sure why this did not work or if I did something wrong. Should I try re-post the comment in the files view, or alternatively, edit the context information manually into the comment?

opm/simulators/wells/GasLiftSingleWellGeneric.cpp
```diff
@@ -748,6 +761,11 @@ runOptimizeLoop_(bool increase)
             state.stop_iteration = true;
         }
     }
+    // we only iterate if well is under thp control
+    if (!state.checkThpControl()) {
+        state.stop_iteration = true;
+    }
```
hakonhagland, line 767: I am probably missing something fundamental, but it is not clear to me why THP control is checked this late? For example, at line 719 at the beginning of this function, `runOptimizeLoop()_`, we call  `computeInitialWellRates_()` which will calculate the well rates given the current ALQ at the THP limit. The ALQ is also adjusted at 734 by the call to `maybeAdjustALQbeforeOptimizeLoop_()` which also use the THP limit pressure to calculate the well rates. Does it make sense to use the THP limit pressure like this when the well is not under THP control? Shouldn't this check be at the beginning of this function? Or even earlier, for example in `gasLiftOptimizationStage1()` in `BlackoilWellModel_impl.hpp` ([line 940](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/BlackoilWellModel_impl.hpp#L940)) : 
```
                    if (group_info.hasWell(well->name())) {
```
The `hasWell()` method of the group info object could pick out the wells with THP control. In fact it already has a check for THP control, see `GasLiftGroupInfo.cpp` [line 280](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/GasLiftGroupInfo.cpp#L280) : 
```
   if (this->optimize_only_thp_wells_) {
```

totto82: If I remember correctly the thp check was moved down the stack to allow for wells under group control to reduce its gaslift. A typical case for this could be that a new well opens up. Then gaslift can be reduced while still keeping the group target.

PR #3414, opened 2021-06-28
Add GRAVITY keyword to UnsupportedFlowKeywords

OPMUSER:
Add GRAVITY keyword to UnsupportedFlowKeywords with message to use DENSITY instead.


bska:
jenkins build this please

PR #3415, opened 2021-06-28
Unsupported Keywords: Fix RVCONST add ROCKOPTS

OPMUSER:
Added ROCKOPTS to unsupported keyword list, as there are no opm-test examples, is this correct?

If ROCKOPTS is supported what is supported?


alfbr:
Actually, I believe it is supported, but cannot remember details. Maybe @totto82 knows?

OPMUSER:
> 
> 
> Actually, I believe it is supported, but cannot remember details. Maybe @totto82 knows?

 @totto82 is it just the defaults implemented?

PR #3416, opened 2021-06-29
Chase Restart File Output Time Decision API Change

bska:
The `Schedule::write_rst_file()` function now takes the time of the previous restart output as a parameter in order to properly account for combinations of `RPTRST`'s `BASIC=4`/`BASIC=5` and `FREQ>1` settings.

### DIFF
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -302,7 +302,12 @@ public:
 
         const auto& gridView = simulator_.vanguard().gridView();
         unsigned numElements = gridView.size(/*codim=*/0);
-        eclOutputModule_->allocBuffers(numElements, restartStepIdx, /*isSubStep=*/false, /*log=*/false, /*isRestart*/ true);
+        eclOutputModule_->allocBuffers(numElements,
+                                       restartStepIdx,
+                                       restartStepIdx,
+                                       /*isSubStep=*/false,
+                                       /*log=*/false,
+                                       /*isRestart*/ true);
 
         {
             SummaryState& summaryState = simulator_.vanguard().summaryState();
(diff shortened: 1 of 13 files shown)


bska:
Note: This PR must be merged along with upstream PR OPM/opm-common#2551.

bska:
jenkins build this opm-common=2551 please

ebos/eclwriter.hh
```diff
@@ -333,6 +338,19 @@ public:
...
     {}
 
+    std::size_t previousRestartOutput() const
+    {
+        auto previous = std::size_t{0};
+        if (this->eclIO_ != nullptr) {
+            previous = this->eclIO_->previousRestartOutputStep();
+        }
+
+        this->simulator_.vanguard().gridView().comm()
```
joakim-hove, line 348: It is only misleading to broadcast this - it is 100% based on the input configuration and that is already globally distributed.

joakim-hove:
I really think the reformatting+++ should be in separate commits marked with e.g. "Whitespace changes".

joakim-hove (changes requested):
Unless there is a very good reason I am not familiar with - I think the broadcast of "last_output_step" should be removed; when it is superfluous it only becomes misleading to keep it. If the parallel distribution of the `EclipseState` were to become different a massive refactoring would be required, and then I doubt preemptive broadcast of "last_report_step" will have any real value.

joakim-hove:
Hmmm - I have thought more about this; I realize I was probably in the wrong ;-( Still not entirely happy with the API though. Right now it looks like the `last_report_step` is set both here in opm-simulators and in the `EclipseIO` class in opm-common; I would prefer if the latter could stay immutable - and the `last_output_step`bookeeping was entirely here in opm-simulator; it is form of a state variable after all - and they should be managed from here. Can discuss later

bska:
Closing this as no longer needed.  The revised implementation of OPM/opm-common#2551 does not change the API anymore.

PR #3421, opened 2021-06-30
compute potentials for newly opened wells due to testing

totto82:


totto82:
jenkins build this please

totto82:
The test failure is caused by a small difference in oil rate (1002 vs 997) that again cause one step earlier closing of the well due to WECON (with limit 1000). Since this fix a bug. i.e. you need the potential for the newly created well , I will start a ref-sol update.

totto82:

totto82:
jenkins build this opm-tests=569 please

PR #3422, opened 2021-06-30
Don't Access Phase Rates Unless Active

bska:
Fixes two-phase Oil/Water runs.  Probably needs more refinement later.


bska:
jenkins build this please

bska:
Self merging to fix undefined behaviour in two-phase cases.

PR #3423, opened 2021-07-01
Assemble UDQState external to the vanguard

joakim-hove:
With this PR the `UDQState` can be initialized external to the vanguard. This is necessary to enable restart from models with UDQ.

I will disavow the current code any day including Sundays, but I have tried to follow the existing pattern as close as possible - for all that is worth.


joakim-hove:
jenkins build this please

joakim-hove:
Comments to this?

bska (approved):
Good extension.  I'll merge into master.

PR #3425, opened 2021-07-01
changed: use std::nullopt

akva2:
the alternative causes some compilers to emit uninitialized
use warnings


akva2:
jenkins build this please

PR #3426, opened 2021-07-01
Unsupported Keywords BCAD

goncalvesmachadoc:


OPMUSER:
Cinta,

BCAD is not an E100/E300 keyword, if this is an OPM Flow keyword then 
may it should be in the 900 OPM directory instead.

OPM User

goncalvesmachadoc:
Hi David, that's an Eclipse 100 keyword to output the polymer adsorption block concentration.

OPMUSER:
Cinta,
Then it does not go in the UnsupportedKeyword list, SUMMARY variables are treated differently. I think if they are not defined in 

`../opm-common/src/opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.cpp` 

then they are automatically ignored - but @joakim-hove is Dr Parser on these topics.

goncalvesmachadoc:
This keyword is not currently ignored:
image
Will add to the parser list then. Thank you.

PR #3427, opened 2021-07-03
Rst load udq

joakim-hove:
With this basic UDQ restarting works - but ....
- There are some white space differences in UDQ expression.
- UDA is not supported yet
- The existing UDQ testcases do not restart - for other reasons ;-(


joakim-hove:
jenkins build this please

PR #3428, opened 2021-07-05
remove ebos_altidx

akva2:
building a whole simulator for this, and then not even
running a test for it, seems rather excessive. if a test for
index-conformance is wanted, a better approach should be taken.

at least in my opinion..


akva2:
jenkins build this please

joakim-hove:
I'll give you review!

image

totto82:
I think the purpose was to avoid error related to hard-coding of indices. From my point of view it can be removed. I really want to move to dynamic indices like what I started on in OPM/opm-models#631 that will also make this simulator absolute. Shouldn't be too much work to finish, but hasn't been on my agenda for a long time. Testing I did back than shows that using dynamic indices did not change the simulation time significantly, but will reduce the number of simulators to compile drastically.

akva2:
jenkins build this please

akva2:
It seems master is broken. MSW instabilities likely triggered by rebooting the machine. Gave us different hardware I guess..

akva2:

akva2:
jenkins build this opm-tests=573 please

akva2:
jenkins build this please

PR #3430, opened 2021-07-06
Some regularisation of tracer rate reporting.

osae:
This hopefully improves the behavior of tracer concentration summary curves, (F/W)T(IR)C.

### DIFF
--- a/ebos/ecltracermodel.hh
+++ b/ebos/ecltracermodel.hh
@@ -390,6 +390,11 @@ protected:
             if (well.getStatus() == Well::Status::SHUT)
                 continue;
 
+            if (!well.isProducer()) //Injection rates already reported during assembly
+                continue;
+
+            Scalar rateWellPos = 0.0;
+            Scalar rateWellNeg = 0.0;
             std::array<int, 3> cartesianCoordinate;
             for (auto& connection : well.getConnections()) {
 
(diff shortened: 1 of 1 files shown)


alfbr:
This is highly welcome! The results are now at the point where I believe it makes sense to mark Tracer supported and have it on by default. The only negative impact I can see is that the computational time of Norne goes up a little, but it really is about time we stop cheating there :-)

bska:
jenkins build this please

ebos/ecltracermodel.hh
```diff
@@ -402,11 +407,33 @@ protected:
...
+            // TODO: Some inconsistencies here that perhaps should be clarified. The "offical" rate as reported below is
+            //  occasionally significant different from the sum over connections (as calculated above). Only observed
+            //  for small values, neglible for the rate itself, but matters when used to calculate tracer concentrations.
+            std::size_t well_index = simulator_.problem().wellModel().wellState().wellIndex(well.name());
+            Scalar official_well_rate_total = simulator_.problem().wellModel().wellState().wellRates(well_index)[tr.phaseIdx_];
+
+            rateWellTotal = official_well_rate_total;
+
+            if (rateWellTotal > rateWellNeg) { // Cross flow
+                const Scalar bucketPrDay = 10.0/(1000.*3600.*24.); // ... keeps (some) trouble away
```
bska, line 432: Do we expect to solve tracers in "LAB" environments?  If so, this value might be too large.

bska (commented):
The code itself looks okay to me.  I just have a question about the `bucketPerDay` threshold.  It seems a little arbitrary and very much geared towards the reservoir simulation application.  Then, of course, if tracers aren't common elsewhere that may well be a reasonable compromise.

alfbr:
> The code itself looks okay to me. I just have a question about the `bucketPerDay` threshold. It seems a little arbitrary and very much geared towards the reservoir simulation application. Then, of course, if tracers aren't common elsewhere that may well be a reasonable compromise.

I cannot recall tracers being used in a lab setting, so I believe it is safe to assume we have a reservoir setting.

bska:
> I cannot recall tracers being used in a lab setting, so I believe it is safe to assume we have a reservoir setting.

Okay, in that case I'll merge this.

PR #3432, opened 2021-07-06
changed: replace code with data in restart output

akva2:
This may be a preference thing, but imo data should be data and not copy-pasted code.


akva2:
jenkins build this please

akva2:
jenkins build this please

bska (approved):
This is a clear improvement over the existing implementation.  I'll merge into master.

PR #3438, opened 2021-07-07
Don't Pass 'log' Argument to 'Schedule::rst_file'

bska:
The function does not use that argument.


bska:
jenkins build this please

bska:
Self-merging as trivial.

PR #3443, opened 2021-07-13
Address Static Code Analysis Warnings

bska:
In particular, don't print uninitialized memory (`Reorder.cpp:left`) and don't capture objects (`prm`) that aren't actually used.  While
here, refactor the initialization of the MT19937 random number generator.  Constructing this object is too expensive to do for each try, especially when we can just run the generator in place.


bska:
jenkins build this please

PR #3447, opened 2021-07-22
Changed default for sched-restart to false

alfbr:
Any simulation without historic schedule will crash if this default is not changed. Maybe it is time to change default, and rather use the command line switch when we fail to initialize wells/groups from binary files.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -125,7 +125,7 @@ struct EclStrictParsing<TypeTag, TTag::EclBaseVanguard> {
 };
 template<class TypeTag>
 struct SchedRestart<TypeTag, TTag::EclBaseVanguard> {
-    static constexpr bool value = true;
+    static constexpr bool value = false;
 };
 template<class TypeTag>
 struct EdgeWeightsMethod<TypeTag, TTag::EclBaseVanguard> {


alfbr:
jenkins build this please

bska:
jenkins build this please

alfbr:
I believe this one is now ready for review/merge. Keeping the current default will probably only create problems at this point.

alfbr:
jenkins build this please

bska:
Can you try to add `--sched-restart=true` to the parallel [numerical_aquifer_3d_2aqu](https://github.com/OPM/opm-simulators/blob/6e37b9ae0511b2f4908cd9c2807d6995be1848a4/compareECLFiles.cmake#L1085) restart test case?  The current test failure **may** be related to incomplete handling of `TUNING` in restarted cases that don't read the complete `SCHEDULE` section.  I [recently](https://github.com/OPM/opm-common/pull/2715#issuecomment-927882734) noticed differences between `--sched-restart=true` and `--sched-restart=false` when running with `--enable-tuning=true`.

I'd prefer to still have `--enable-tuning=true` for this test case lest the simulator take timesteps that are too large for the comparison to mean anything.

alfbr:
> Can you try to add `--sched-restart=true` to the parallel [numerical_aquifer_3d_2aqu](https://github.com/OPM/opm-simulators/blob/6e37b9ae0511b2f4908cd9c2807d6995be1848a4/compareECLFiles.cmake#L1085) restart test case? The current test failure **may** be related to incomplete handling of `TUNING` in restarted cases that don't read the complete `SCHEDULE` section. I [recently](https://github.com/OPM/opm-common/pull/2715#issuecomment-927882734) noticed differences between `--sched-restart=true` and `--sched-restart=false` when running with `--enable-tuning=true`.
> 
> I'd prefer to still have `--enable-tuning=true` for this test case lest the simulator take timesteps that are too large for the comparison to mean anything.

Yes, I have tested this case with and without historic schedule. The failure is real, we are not restarting this case correctly without historic schedule. I am very impressed that this is the only failure. From my end, the best solution is to fix loading this case without historic schedule.  See plot of first failing vector (restarted from 2019-03):
image

bska:
> From my end, the best solution is to fix loading this case without historic schedule.

Of course.  On the other hand if you want to merge this PR soon then the only short-terms solution I can offer is to add `--sched-restart=true` to that test case.

alfbr:
> > From my end, the best solution is to fix loading this case without historic schedule.
> 
> Of course. On the other hand if you want to merge this PR soon then the only short-term solution I can offer is to add `--sched-restart=true` to that test case.

I would prefer disabling the test actually (since it effectively will maintain a deprecated solution in that case). However, I have waited for over two months on this merge, so I can wait some more :-D

PR #3448, opened 2021-07-23
Add basic functionality for validating special case keywords

joakim-hove:
Just suggestion

### DIFF
--- a/opm/simulators/flow/KeywordValidation.cpp
+++ b/opm/simulators/flow/KeywordValidation.cpp
@@ -33,8 +33,11 @@
 namespace Opm
 {
 
+
+
 namespace KeywordValidation
 {
+
 
     std::string get_error_report(const std::vector<ValidationError>& errors, const bool critical)
     {
@@ -75,12 +78,21 @@ namespace KeywordValidation
     }
 
     void
-    KeywordValidator::validateDeck(const Deck& deck, const ParseContext& parse_context, ErrorGuard& error_guard) const
+    KeywordValidator::validateDeck(const Deck& deck,
+                                   const ParseContext& parse_context,
+                                   ErrorGuard& error_guard) const
     {
         // Make a vector with all problems encountered in the deck.
         std::vector<ValidationError> errors;
-        for (const auto& keyword : deck)
+        for (const auto& keyword : deck) {
             validateDeckKeyword(keyword, errors);
+
+            const auto& special_it = this->m_special_validation.find(keyword.name());
+            if (special_it != this->m_special_validation.end()) {
+                const auto& validator = special_it->second;
+                validator(keyword, errors);
+            }
+        }
 
         // First report non-critical problems as a warning.
         auto warning_report = get_error_report(errors, false);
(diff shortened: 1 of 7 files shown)


joakim-hove:
jenkins build this please

opm/simulators/flow/KeywordValidation.cpp
```diff
@@ -111,6 +128,17 @@ namespace KeywordValidation
...
+        std::unordered_map<std::string, std::function<void(const DeckKeyword& keyword, std::vector<ValidationError>& errors)>> special_validation {
+            {"BRINE", validateBRINE}
+        };
+
+        const auto& special_it = special_validation.find(keyword.name());
+        if (special_it != special_validation.end()) {
+            const auto& validator = special_it->second;
+            validator(keyword, errors);
+        }
+
```
verveerpj, line 141: This looks like a good approach to add some very generic tests to keywords. However, maybe it would be good to add/store the map of validation functions as an argument when the object is created. Then actual validation code remains separate from the more general validator object.

joakim-hove: > However, maybe it would be good to add/store the map of validation functions as an argument when the object is created.

Absolutely - the chosen implementation was just a demonstration which should be as little obtrusive as possible. I'll let it linger for now so others can comment - but if this is generally well received I will address your comment before merging.

joakim-hove:
@alfbr : comments to this?

alfbr:
I am sorry, I have not had any time to look at this after vacation. Please allow a few more days.

joakim-hove:
OK - I have updated as suggested by @verveerpj. Exactly at what scope the mapping of special functions should be defined can of course be discussed - this is my suggestion.

opm/simulators/flow/KeywordValidation.cpp
```diff
@@ -33,9 +33,26 @@
 namespace Opm
 {
 
+
+
 namespace KeywordValidation
 {
 
+
```
verveerpj, line 41: It would probably be better to have special validation files in a separate file, to separate it from the generic functionality of the validation class. But that can be done when the number of these functions grows.

opm/simulators/flow/KeywordValidation.cpp
```diff
@@ -75,13 +92,23 @@ namespace KeywordValidation
     }
 
     void
-    KeywordValidator::validateDeck(const Deck& deck, const ParseContext& parse_context, ErrorGuard& error_guard) const
+    KeywordValidator::validateDeck(const Deck& deck,
+                                   const std::unordered_map<std::string, std::function<void(const DeckKeyword& keyword, std::vector<ValidationError>& errors)>>& special_validation,
```
verveerpj, line 96: We could also pass the map of functions in the constructor of the validation object. That is also where the other validation info is passed, so it may be a bit more logical. But passing it here works also fine, of course. What do you think?

verveerpj (approved):
LGTM, two comments just to think about.

opm/simulators/flow/KeywordValidation.cpp
joakim-hove: That is a good idea - Will upfate.

opm/simulators/flow/KeywordValidation.cpp
joakim-hove: I'll Split as suggested

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

PR #3453, opened 2021-08-01
Adding command line argument for writing esmry file 

tskille:
Extra command line argument write-lodsmry for creating LODSMRY file
 - default is false
 - LODSMRY file is re-written for every time step
 - file LODSRMY optimized for fast loading of selected vectors
 - This PR is dependent on PR OPM/opm-common#2590

### DIFF
--- a/ebos/eclgenericwriter.cc
+++ b/ebos/eclgenericwriter.cc
@@ -163,7 +163,8 @@ EclGenericWriter(const Schedule& schedule,
                  const GridView& gridView,
                  const Dune::CartesianIndexMapper<Grid>& cartMapper,
                  const Dune::CartesianIndexMapper<EquilGrid>* equilCartMapper,
-                 bool enableAsyncOutput)
+                 bool enableAsyncOutput,
+                 bool enableEsmry )
     : collectToIORank_(grid,
                        equilGrid,
                        gridView,
@@ -182,7 +183,7 @@ EclGenericWriter(const Schedule& schedule,
         eclIO_.reset(new EclipseIO(eclState_,
                                    UgGridHelpers::createEclipseGrid(*equilGrid, eclState_.getInputGrid()),
                                    schedule_,
-                                   summaryConfig_));
+                                   summaryConfig_, "", enableEsmry));
 
         const auto& wbp_calculators = eclIO_->summary().wbp_calculators( schedule.size() - 1 );
         wbp_index_list_ = wbp_calculators.index_list();
--- a/ebos/eclgenericwriter.hh
+++ b/ebos/eclgenericwriter.hh
@@ -69,7 +69,8 @@ public:
                      const GridView& gridView,
                      const Dune::CartesianIndexMapper<Grid>& cartMapper,
                      const Dune::CartesianIndexMapper<EquilGrid>* equilCartMapper,
-                     bool enableAsyncOutput);
+                     bool enableAsyncOutput,
+                     bool enableEsmry);
 
     const EclipseIO& eclIO() const;
 
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -437,6 +437,11 @@ template<class TypeTag>
 struct EnableAsyncEclOutput<TypeTag, TTag::EclBaseProblem> {
     static constexpr bool value = true;
 };
+// Write ESMRY file for fast loading of summary data
+template<class TypeTag>
+struct EnableEsmry<TypeTag, TTag::EclBaseProblem> {
+    static constexpr bool value = false;
+};
 
 // By default, use single precision for the ECL formated results
 template<class TypeTag>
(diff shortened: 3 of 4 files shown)


bska:
jenkins build this opm-common=2590 please

bska (approved):
This looks good to me.  I'll merge once the upstream is approved.

OPMUSER:
I have two comments, don't like the name LODSMRY, especially if that is going to be the extension and also the command line option:

 `--write_lodsmry, default = false`

as it will list in alphabetic order and be on it's own. So What about:

` --enable-ecl-output-lodsmry, default = false`

Again I don't like the name  LODSMRY, so what about UNSMRY-FLOW, or UNLOD

The idea being the files will list next to the other standard unified binary files.

bska:
> I [...] don't like the name LODSMRY, especially if that is going to be the extension and also [name of] the command line option

@OPMUSER: I appreciate your concerns, but there's already some historic significance to the `LODSMRY` name.  As it happens, ResInsight recognises that extension (and file structure) and we should at least coordinate with them if we consider changing the extension.

I don't really have an opinion on the option name.  Maybe one of

* `--enable-lodsmry=true`
* `--enable-ecl-lodsmry=true`

could be a reasonable compromise?  We might also consider "hiding" the option from the `--help` output, meaning the option would be listed only as part of `--help-all`.

magnesj:
We have implemented beta support for import of `LODSMRY` files in ResInsight. As this is beta support, it will be a minor issue for us to change the naming of these files. Backwards compability is not important from our point of view.

We would like a more descriptive name than `LODSMRY`, and I prefer the extension `UNSMRY-FLOW` as suggested by @OPMUSER

tskille:
This is not a unified file in the sense of multiple files brought together in to one file. The UNSMRY file is a collation of PARAMS keywords (which holds summary data for one time step).  There will never a a "multiple" variant of this file. One file for each summary vector will never happen.  

This file is fundamentally different from the UNSMRY file as this file is storing the summary vector data together (rather than time step data). This is the reason why this file need a complete rewrite each time a new time step is added.  

I don't like to attach the name FLOW to the file since this is a general format that can be used by other simulators or applications. Notice that none of the files generated by Eclipse, have the name ECL (or Eclipse) in the file name.

The name LODSMRY consists of two part LOD and SMRY. SMRY tells that this is a summary file and LOD is an abbreviation for Load On Demand. This file is designed for effective loading of individual vectors. The format of the file makes is easy to calculate the location of each summary vector. 

The process of loading data from this file is as following
```
1. Open the file
2. seek to location of selected summary vector 
3. read the data as one continuous  stream
4. close the file. 
```
I'm not saying that LODSMRY is the perfect name, but in my opinion this is fare more descriptive than the extension UNSMRY-FLOW.

atgeirr:
> I'm not saying that LODSMRY is the perfect name, but in my opinion this is fare more descriptive than the extension UNSMRY-FLOW.

I agree with that, but I also see that LOD is not really informative to the regular user. What about FASTSMRY?

tskille:
> What about FASTSMRY?

if we use FASTSMRY, this will then be the only file extension starting with F, which is not a formatted file. All other files starting with F are formatted (FEGRID, FINIT, FSMSPEC, FUNSMRY, FUNRST, ..)

Would it be ok with **ESMRY** (enhanced summary file), or even simpler **SMRY**. 

It is not more informative but we are at least getting rid of LOD which everybody seems to dislike.

alfbr:
First up, I have no strong opinions either way on this. That said, my preference is a short suffix, like ESMRY. I also think we should avoid the simulator name in the suffix, e.g., SMRY-FLOW. This is after all a file format, and branding it with the simulator name is in my opinion not good practice. From a general point of view, I believe we should aim for file formats that potentially can be standardized and broadly used, rather than Flow specific.

OPMUSER:
Good we are having the discussion and appear to be converging on a consensus. Some suggestions based on the above:

- ESMRY
- SMRYE
- SMYR

I think any of the above will work fine, somebody decide :-)

For the command line, assuming we go with ESMYR

`--enable-ecl-output-esmry=true`
or
`--enable-ecl-esmry=true`

tskille:
@OPMUSER , I'm glad that you are happy with ESMRY as file extension.

I think both of your suggestions on command line options are too long.  I would preferer to have this as short as possible. This increases the the chance of getting this correct when typing these commands on the command line.

```
--enable-esmry=true
```
why would you like to include -ecl and/or -output to the command line. It is pretty clear that this is output esmry is what you get. Is the description of the option ok ?

```
flow -h 
...
--enable-esmry=BOOLEAN                       Write ESMRY file for fast loading of summary data. Default: false
```

Are you happy about

bska:
For what it's worth, I think @tskille's latest suggestion of an `ESMRY` file extension and `--enable-esmry` option name is a good way to go.

OPMUSER:
@tskille you mean you never typed `linear-solver-require-full-sparsity-pattern=false` or` min-time-step-before-shutting-problematic-wells-in-days=0.01` :-). I use OPMRUN so never have to type any commands.

The reason for the my suggestions is to try and group similar keywords together, that's all. But I'm okay with -`-enable-esmry=true
`

OPMUSER:
@tskille Forgot to confirm that the description for `--enable-esmry=true ` is good.

tskille:
@bska , I have now updated the PR with respect to discussions above

./bin/flow -h | grep ESMRY

show on my terminal 

```
    --enable-esmry=BOOLEAN                        Write ESMRY file for fast loading of summary data. Default: false
```
The upstream PR is also updated.

bska:
jenkins build this opm-common=2590 please

bska (approved):
Thanks for renaming the option.  I'll merge into master.

PR #3445, opened 2021-07-21
Rateconvfixed

plgbrts:
Fix for the situation that hydrocarbon consist only of gas. Relevant for water-gas twophase system

Issue #665, opened 2016-04-27
Non-Symmetric Matrix Structure in Two-Phase Simulation


bska:
I tried to test @babrodtk's changes OPM/opm-core#1003 and #661 on a two-phase simulation case that is currently **not** in [opm-data](https://github.com/OPM/opm-data).  I can provide additional details offline.  Towards the end of the simulation I get an exception from Dune ISTL's [`compressed_base_array_unmanaged<>`](https://gitlab.dune-project.org/core/dune-istl/blob/master/dune/istl/basearray.hh#L565-571) saying

```
index 2345 not in compressed array
```

This is just the symptom, of course.

The immediate calling code is method

``` C++
NewtonIterationBlackoilInterleavedImpl<>::formInterleavedSystem()
```

and especially the [loop](https://github.com/OPM/opm-simulators/blob/master/opm/autodiff/NewtonIterationBlackoilInterleaved.cpp#L349-L364) that inserts derivatives into the interleaved system matrix.

Just prior to encountering this problem all wells were switched from being constrained by `RESERVOIR_RATE` targets to being constrained by `BHP` targets

```
Iter  W-FLUX(W)  W-FLUX(O)
   0  1.050e-05  6.617e-04
well converged iter: 0

Iter   MB(Wat)    MB(Oil)     CNV(W)     CNV(O)   W-FLUX(W)  W-FLUX(O)
   0  6.382e-06  3.804e-07  3.212e-03  2.453e-03  1.050e-05  6.617e-04
Switching control mode for well BR-P-1 from RESERVOIR_RATE to BHP
Switching control mode for well BR-P-2 from RESERVOIR_RATE to BHP
Switching control mode for well BR-P-3 from RESERVOIR_RATE to BHP
Switching control mode for well BR-P-4 from RESERVOIR_RATE to BHP
Switching control mode for well BR-P-5 from RESERVOIR_RATE to BHP
Switching control mode for well BR-P-6 from RESERVOIR_RATE to BHP
Switching control mode for well BR-P-7 from RESERVOIR_RATE to BHP
Switching control mode for well BR-P-8 from RESERVOIR_RATE to BHP
Switching control mode for well BR-P-9 from RESERVOIR_RATE to BHP
Switching control mode for well BR-P-10 from RESERVOIR_RATE to BHP
Switching control mode for well BR-P-11 from RESERVOIR_RATE to BHP
Switching control mode for well BR-P-12 from RESERVOIR_RATE to BHP
[...]

Issue #446, opened 2015-08-31
[regression] flow: the norne deck does not work


andlaus:
I'm using debug mode and get the following:

```
and@heuristix:~/src/opm-autodiff|13262f6... > ./bin/flow deck_filename=~/src/opm-data/norne/NORNE_ATW2013.DATA
**********************************************************************
*                                                                    *
*                   This is Flow (version 2015.04)                   *
*                                                                    *
* Flow is a simulator for fully implicit three-phase black-oil flow, *
*            and is part of OPM. For more information see:           *
*                       http://opm-project.org                       *
*                                                                    *
**********************************************************************

---------------    Reading parameters     ---------------
[...]

================ Starting main simulation loop ===============
timestep.adaptive not found. Using default value 'true'.
solver.restartfactor not found. Using default value '0.1'.
solver.growthfactor not found. Using default value '1.25'.
timestep.max_timestep_in_days not found. Using default value '365'.
solver.restart not found. Using default value '10'.
solver.verbose not found. Using default value 'true'.
timestep.verbose not found. Using default value 'true'.
full_timestep_initially not found. Using default value 'false'.
timestep.control not found. Using default value 'pid+iteration'.
timestep.control.tol not found. Using default value '0.001'.
timestep.control.targetiteration not found. Using default value '30'.
timestep.control.maxgrowth not found. Using default value '3'.
restorefile not found. Using default value ''.


---------------    Simulation step number 0    ---------------
      Current time (days)     0
      Current stepsize (days) 8
      Total time (days)       3312

Substep( 0 ), try with stepsize 0.8 (days).

Iter  MB(WATER)   MB(OIL)    MB(GAS)       CNVW       CNVO       CNVG   W-FLUX(W)  W-FLUX(O)  W-FLUX(G)
   0  7.960e-10  1.805e-06  1.085e-05  8.519e-03  9.334e-03  1.977e-01  3.975e-06  1.154e-02  9.972e-02
   1  2.454e-09  5.599e-07  8.061e-08  3.149e-02  7.581e-03  4.554e-01  3.097e-07  5.314e-04  1.966e-04
   2  1.012e-08  3.840e-07  3.682e-07  4.418e-04  2.594e-03  2.145e-02  6.552e-11  1.819e-07  1.041e-07
   3  2.335e-11  2.287e-08  3.970e-08  2.480e-05  7.059e-04  7.574e-03  2.114e-14  2.394e-11  9.059e-12
Overall linear iterations used: 3
[...]

PR #3454, opened 2021-08-02
adjustments needed for gas-water system

plgbrts:
fixes to allow for gas-water system. This PR is related to https://github.com/OPM/opm-models/pull/665

### DIFF
--- a/opm/simulators/wells/StandardWellEval.cpp
+++ b/opm/simulators/wells/StandardWellEval.cpp
@@ -90,19 +90,19 @@ relaxationFactorFractionsProducer(const std::vector<double>& primary_variables,
     double relaxation_factor = 1.0;
 
     if (FluidSystem::numActivePhases() > 1) {
-        if (FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx)) {
+        if (has_wfrac_variable) { 
             const double relaxation_factor_w = StandardWellGeneric<Scalar>::
                                                relaxationFactorFraction(primary_variables[WFrac], dwells[0][WFrac]);
             relaxation_factor = std::min(relaxation_factor, relaxation_factor_w);
         }
 
-        if (FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx)) {
+        if (has_gfrac_variable) { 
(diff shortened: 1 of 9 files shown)


opm/simulators/wells/StandardWellEval.cpp
```diff
@@ -90,19 +90,19 @@ relaxationFactorFractionsProducer(const std::vector<double>& primary_variables,
     double relaxation_factor = 1.0;
 
     if (FluidSystem::numActivePhases() > 1) {
-        if (FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx)) {
+        if (has_wfrac_variable) {
```
akva2, line 93: please use if constexpr for purely compile time conditionals. micro-optim which allows for code elimination.

plgbrts: changed `if (...) {` to ` if constexpr (...) { `

totto82:
Thanks for the contribution. The code looks good to me. Are these PRs sufficient to run gas-water or is it a step on way.

plgbrts:
With this PR (#3454) and
-opm-models PR #665 (not merged yet)
-opm-simulators PR #3445 (merged)
-opm-material PR #446 (not merged yet)
a gas-water case can run. I will make a PR for a GAS-WATER test case.

totto82:
I suggest you add the newly merged 2p gas-water case in the jenkins test suite so we don't loose the functionality again.

goncalvesmachadoc:
He is about to push that. :)

OPMUSER:
Although the test case runs because the model users Enumeration Initialization to initialize the model. If you use the standard Equilibrium Initialization the model fails to initialize because it is looking for an oil phase:

```
Program threw an exception: Could not initialize the problem: Phase pressure function for "OIL" most not be null
Failed: 2882
```

Test deck attached.

[GASW-BASE01-R01.zip](https://github.com/OPM/opm-simulators/files/7119785/GASW-BASE01-R01.zip)

goncalvesmachadoc:
Yes, that's a known issue. Equilibration for initialization in OPM only works when there is oil in the model.

OPMUSER:
@goncalvesmachadoc thanks for the clarification.

PR #3460, opened 2021-08-04
WIP: SALTP initialization and output

plgbrts:


ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -675,6 +675,9 @@ assignToSolution(data::Solution& sol)
     if (!cSalt_.empty())
         sol.insert ("SALT", UnitSystem::measure::salinity, std::move(cSalt_), data::TargetType::RESTART_SOLUTION);
 
+    if (!pSalt_.empty())
+        sol.insert ("SALTP", UnitSystem::measure::identity, std::move(pSalt_), data::TargetType::RESTART_SOLUTION);
+
```
totto82, line 680: For homemade names like SALTP you will need to use  data::TargetType::RESTART_AUXILIARY and run with --enable-opm-rst-file=true. Isn't that right @bska?

bska: > For homemade names like `SALTP` you will need to use `data::TargetType::RESTART_AUXILIARY` and run with `--enable-opm-rst-file=true`

Yes, that's the current scheme.  Vectors tagged as `RESTART_SOLUTION` are treated as fully ECLIPSE compatible and if they're not we won't be able to restart ECLIPSE from a Flow run.

PR #3418, opened 2021-06-30
iterate to get more accurate potentials for stw wells

totto82:
This is adds iterations to the well potential calculation.

PR #3461, opened 2021-08-04
consindering the situation of single water phase for RateConverter

GitPaean:
an attempt to fix the failure https://ci.opm-project.org/job/opm-simulators-PR-builder/2766/testReport/junit/(root)/mpi/compareParallelSim_flow_onephase_energy_SPE1CASE2_THERMAL_ONEPHASE/ from https://github.com/OPM/opm-simulators/pull/3418#issuecomment-892716714

It is a bug introduced in https://github.com/OPM/opm-simulators/pull/3445 , not sure how it was missed from the jenkins test. 

And also, the current version is a quick attempt. I do not have the full knowledge about how the gas-water system is developed, the current version might introduce some inconsistency. 

Anyway, it is an attempt.

### DIFF
--- a/opm/simulators/wells/RateConverter.hpp
+++ b/opm/simulators/wells/RateConverter.hpp
@@ -521,10 +521,13 @@ namespace Opm {
                         if (Details::PhaseUsed::oil(pu)) {
                             attr.pressure += fs.pressure(FluidSystem::oilPhaseIdx).value() * pv_cell;
                             attr.temperature += fs.temperature(FluidSystem::oilPhaseIdx).value() * pv_cell;
-                        } else {
-                             assert(Details::PhaseUsed::gas(pu));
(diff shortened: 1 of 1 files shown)


GitPaean:
jekins build this please

akva2:
jenkins build this please

akva2:
undrafting so i can test https://github.com/OPM/opm-common/pull/2594 with it.

akva2:
this seems to do the job.

bska:
> this seems to do the job.

With **seems** being the operative word.  The proposed change does not take into account that the entire `RateConverter` facility is predicated on the existence of at least one hydrocarbon phase.

akva2:
okay, what would be a proper approach ?

bska:
Having mulled over the issue for some time I think I prefer this option to #3463.  It looks like it'll probably do the job so I'll merge into master.

PR #3462, opened 2021-08-05
fixed: add missing initialization call

akva2:
I think this is the correct fix. It certainly fixes the assertion (ref https://github.com/OPM/opm-common/pull/2594)


akva2:
@hakonhagland please confirm.

akva2:
ref https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/BlackoilWellModel_impl.hpp#L882

akva2:
jenkins build this please

totto82:
Thanks for the catch and the fix.

PR #3463, opened 2021-08-05
disable rateconverter without hydrocarbon phase

akva2:
Total hack, but atleast it does the job. Waiting for @bska to instruct on how to actually do it.

### DIFF
--- a/opm/simulators/wells/RateConverter.hpp
+++ b/opm/simulators/wells/RateConverter.hpp
@@ -418,6 +418,7 @@ namespace Opm {
                 , rmap_ (region)
                 , attr_ (rmap_, Attributes())
             {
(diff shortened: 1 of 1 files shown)


akva2:
as expected, this affects the regression tests, but the parallel tests pass, indicating that behavior is at least deterministic.

opm/simulators/wells/RateConverter.hpp
```diff
@@ -432,6 +433,8 @@ namespace Opm {
             template <typename ElementContext, class EbosSimulator>
             void defineState(const EbosSimulator& simulator)
             {
+                if (!active)
+                    return;
```
joakim-hove, line 437: Yaaah - the *Say no to superfluous braces* club has just accepted a new sturdy member!

akva2: weell. this was not really a grand political statement from me, more of a habitual mistake haha.

PR #3466, opened 2021-08-12
fix build

akva2:
need to include header for assert.

PR #3244 broke the build in serial, sorry about that..


akva2:
self-merging as this is trivial and post-builder is broken..

PR #3468, opened 2021-08-16
Improve parallel build

akva2:
Slightly more compile units, but we can then build in parallel with the main library. This improves build time, in particular with many build jobs.


akva2:
jenkins build this please

bska (approved):
Looks good to me and I appreciate the effort to improve build times for parallel builds.  I'll merge into master.

PR #3477, opened 2021-08-18
added: 'update_data' custom target

akva2:
this allows updating opm-tests reference data repository
with the current results.

this is useful for devs as some of the tests only pass on the
CI system due to slight differences. you can now run this
before applying changes, and thus only reveal real regressions
introduced by your changes.

Downstream of https://github.com/OPM/opm-common/pull/2613

@joakim-hove took a while but here you go. please test.

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -1152,3 +1152,10 @@ endif()
                                        DIR udq_actionx
                                        TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-6)
 endif()
+
+if(OPM_TESTS_ROOT)
+  add_custom_target(update_data
+                    COMMAND ${CMAKE_COMMAND} --build ${PROJECT_BINARY_DIR} --target check --parallel || exit 0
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this opm-common=2613 please

joakim-hove:
> took a while but here you go. please test.

Cool - will give it a spin tomorrow!

joakim-hove:
Thank you - this works as intended for me. There are some performance issues which you might address:

1. It does not seem to respect `-j N` for parallel builds:
```
hove@ws:~/work/OPM/opm-simulators/build$ make update_data -j 16
[100%] Updating reference data
make[4]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
....
```
then a serial build starts. 

2. It seems to build all sorts of ebos++ simulators which are not needed in the regression testing?

joakim-hove:
Comments about performance aside it is *extremely nice* to be able to test properly in opm-simulators!

akva2:
1 is a bit hard to address safely if you use Make. I cannot simply depend on the target as the target will fail. So I have to execute it as a separate build invocation and ignore failure. I opted for the über-safe now, but  I guess we have to go for the somewhat unsafe 'user better have enough memory for make -j'. With ninja (which I use locally) this is less of a problem since default number of build jobs at least is limited by the amount of cores on the system.

2 is beyond the scope of this PR. the check target always did this because there is no facility to say whether or not a simulator has a test or not, so it has to build em all.

PR #3479, opened 2021-08-18
Make sure that flow is not compiled with -DNDEBUG

joakim-hove:
Ensure that the `assert(....)` safety belt is always present.

flow is a complex piece of software, and the input format is quite complex - and not 100% understood. This implies that the distinction between pure programming errors which *should not happen* - and error conditions which can be traced to user input is not entirely clear. I have several times wasted quite some debugging time in a situation where I thought everything worked - to later discover that `assert( something )` has been broken when switching to debug mode.

Piggy back on: https://github.com/OPM/opm-common/pull/2594

### DIFF
--- a/opm/simulators/flow/Main.hpp
+++ b/opm/simulators/flow/Main.hpp
@@ -67,6 +67,15 @@
 #include <string>
 #include <type_traits>
 
+
+#ifdef NDEBUG
+static_assert(false, "In order to enforce assert(...) checks flow can *not* be compiled with -DNDEBUG");
+#endif
+
+
+
+
+
 namespace Opm::Properties {
 
 // this is a dummy type tag that is used to setup the parameters before the actual


joakim-hove:
jenkins build this please

akva2:
my .01c; no. this is worse than -Werror. if we want to run with asserts enabled by default, we can do that (see https://github.com/OPM/opm-common/pull/2594) but enforcing it is not how you treat your users.

joakim-hove:
> if we want to run with asserts enabled by default, we can do that (see OPM/opm-common#2594) but enforcing it is not how you treat your users.

Well - my main point is that we waste quite a lot of debugging time due to silenced `assert(...)` - the inability to merge OPM/opm-common#2594 illustrates the point. I was not expecting this to be merged - but see it as a strong nudge in the direction of building with `assert( )` enabled as default.

PR #3482, opened 2021-08-19
Store the action result - when evaluated to true

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/2618


bska:
Merging this now.

PR #3480, opened 2021-08-19
Single well state2

joakim-hove:
These commits were "forgotten" when merging #3458

PR #3485, opened 2021-08-23
Single well state3

joakim-hove:
Part of #3480

### DIFF
--- a/opm/simulators/wells/SingleWellState.hpp
+++ b/opm/simulators/wells/SingleWellState.hpp
@@ -20,6 +20,8 @@
 #ifndef OPM_SINGLE_WELL_STATE_HEADER_INCLUDED
 #define OPM_SINGLE_WELL_STATE_HEADER_INCLUDED
 
+#include <vector>
+
 #include <opm/parser/eclipse/EclipseState/Schedule/Well/Well.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/Events.hpp>
 
@@ -27,7 +29,7 @@ namespace Opm {
 
 class SingleWellState {
 public:
-    SingleWellState(bool is_producer, double temp);
+    SingleWellState(bool is_producer, std::size_t num_phases, double temp);
 
     Well::Status status{Well::Status::OPEN};
     bool producer;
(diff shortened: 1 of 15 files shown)


joakim-hove:
jenkins build this please

opm/simulators/wells/SingleWellState.hpp
```diff
@@ -20,20 +20,28 @@
...
 public:
-    SingleWellState(bool is_producer, double temp);
+    SingleWellState(bool is_producer, std::size_t num_phases, double temp);
 
     Well::Status status{Well::Status::OPEN};
     bool producer;
     double bhp{0};
     double thp{0};
-    double temperature{};
+    double temperature;
```
bska, line 39: This makes `temperature` undefined in the `SingleWellState` unless a call site explicitly assigns a value.  Please revert.

bska (changes requested):
Most of this looks to be a mechanical API change, but there's one piece of the PR that causes undefined behaviour.  Please fix.

opm/simulators/wells/SingleWellState.hpp
joakim-hove: I'll revert sure enough, not intentional in any way. That said I do think the fringe feature temperature gets way to much attention/focus in the well code.

joakim-hove: Ehhh - second thought; the change to uninitialized was actually semi-intentional - to highlight the fact that the temperature value is just garbage. With `double temperature{}` it is well defined from a C++ standpoint, but from a physics/reservoir it still just garbage - although garbage.

Anyway - I'll put that :horse: back in the stable and revert!

joakim-hove:
jenkins build this please

opm/simulators/wells/SingleWellState.hpp
bska: > the change to uninitialized was actually semi-intentional - to highlight the fact that the temperature value is just garbage.

I appreciate that, but by making the data member uninitialised you also create undefined behaviour in any operation that works on the structure as a whole.

> With `double temperature{}` it is well defined from a C++ standpoint

And that's needed.

bska (approved):

PR #3486, opened 2021-08-23
always return fraction 1.0 if only one well is under group control 

totto82:
This also applies to cases where the guide rate is zero


totto82:
@atgeirr  Any comments on this?

I want this merged asap.

totto82:
jenkins build this please

PR #3498, opened 2021-09-01
Support for gpmaint

totto82:
Adds support for gpmaint.

### DIFF
--- a/opm/simulators/wells/RateConverter.hpp
+++ b/opm/simulators/wells/RateConverter.hpp
@@ -197,6 +197,32 @@ namespace Opm {
                     return this->find(reg).cell_;
                 }
 
+                bool has(const RegionID reg) const
+                {
+                    const auto& i = attr_.find(reg);
+
+                    if (i == attr_.end()) {
+                        return false;
+                    }
+
+                    return true;
+                }
+
+                void insert(const RegionID r, const Attributes& attr)
+                {
+                    using VT = typename AttributeMap::value_type;
+
+                    auto v = std::make_unique<Value>(attr);
+
+                    const auto stat = attr_.insert(VT(r, std::move(v)));
+
+                    if (stat.second) {
+                        // Region's representative cell.
+                        stat.first->second->cell_ = -1.0;
+                    }
+                }
+
+
                 /**
                  * Request read-only access to region's attributes.
                  *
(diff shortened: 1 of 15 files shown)


totto82:
jenkins build this please

opm/simulators/wells/RateConverter.hpp
```diff
@@ -855,6 +855,191 @@ namespace Opm {
...
+         *
+         * \tparam Region Type of a forward region mapping.  Expected
+         * to provide indexed access through \code operator[]()
+         * \endcode as well as inner types \c value_type, \c
+         * size_type, and \c const_iterator.  Typically \code
+         * std::vector<int> \endcode.
+         */
+        template <class FluidSystem, class Region>
+        class AverageRegionalPressure {
+        public:
```
totto82, line 872: The file name (and type) RateConverter is now misleading since it used to do more stuff. I suggest renaming to AverageRegionalCalculator. Any objections?

opm/simulators/wells/RateConverter.hpp
```diff
@@ -855,6 +855,191 @@ namespace Opm {
...
+                             attr.pressure += fs.pressure(FluidSystem::gasPhaseIdx).value() * pv_cell;
+                        } else {
+                            assert(Details::PhaseUsed::water(pu));
+                            attr.pressure += fs.pressure(FluidSystem::waterPhaseIdx).value() * pv_cell;
+                        }
+                    }
+                }
+
+                for (const auto& reg : rmap_.activeRegions()) {
+                      auto& ra = attr_.attributes(reg);
```
totto82, line 976: The current RateConverter code is buggy. If a region does not contain any cells for a particular partition this comm.sum is never called and we get a dead-look.

totto82: I have fixed the mpi deadlook for gpmaint, but not yet for the rateconverter. Not the most elegant solution, but it is simple and works.

opm/simulators/wells/RateConverter.hpp
```diff
@@ -197,6 +197,32 @@ namespace Opm {
...
 
+                bool has(const RegionID reg) const
+                {
+                    const auto& i = attr_.find(reg);
+
+                    if (i == attr_.end()) {
+                        return false;
+                    }
+
+                    return true;
```
bska, line 208:
```C++
return this->attr_.find(reg) != this->attr_.end();
```

opm/simulators/wells/RateConverter.hpp
```diff
@@ -197,6 +197,32 @@ namespace Opm {
...
+                    using VT = typename AttributeMap::value_type;
+
+                    auto v = std::make_unique<Value>(attr);
+
+                    const auto stat = attr_.insert(VT(r, std::move(v)));
+
+                    if (stat.second) {
+                        // Region's representative cell.
+                        stat.first->second->cell_ = -1.0;
+                    }
```
bska, line 222:
```C++
auto [pos, inserted] = this->attr_.try_emplace(r, std::make_unique<Value>(attr));
if (inserted) {
    pos->second->cell_ = -1; // NOT -1.0 -- "cell_" is 'int'
}
```

bska (commented):
> Not the most elegant solution, but it is simple and works.

I agree.  I'll likely revise this entire facility when I get the time.  This really isn't how it's supposed to work.

opm/simulators/wells/RateConverter.hpp
bska: > I suggest renaming to AverageRegionalCalculator. Any objections?

I object.  That's not a good name as the primary purpose is **still** to convert flow rates from surface conditions to reservoir conditions (and back).

totto82: The AverageRegionalCalculator is a new class,  the old SurfaceToReservoirVoidage is still there to do the conversion of the flow rates. What I want is to rename the file.

atgeirr: I think we should split these classes into separate files. It is good to have one important class with unique function per file. Maybe also put common facilities in its own, separate file.

totto82: I would still argue for keeping them in the same file and instead rename the file. The implementation are almost the same. But since both @bska and @atgeirr thinks differently I will yield and make a new file called RegionalAverageCalculator.hpp. Can I put the facilities there? If not do you have a suggestion for a name for the file with the facilities? FacilitiesForRegionalAverageCalculators?

atgeirr (commented):
I think this code is fine, just a few minor issues. There are some corner cases I think are not handled, but we may not want to spend much time addressing them:
 - The most important is handling region 0 (which means all cells). I do not think this is handled currently. If so, I missed it!
 - I do not believe we handle combinations of WCONINJE and GPMAINT correctly (they might both apply I think), but without cases that test this, we cannot really know. The NONE option to GPMAINT is likely not handled correctly. I believe there are three states for a (non-defaulted) group control: Target set explicitly (GCONINJE or GCONPROD), target should be updated every timestep (GPMAINT) and target should remain as it is, but no longer updated (after GPMAINT with NONE).

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
Review is done and jenkins is happy. I will merge later today if no objections.

PR #3499, opened 2021-09-01
[bugfix] Prevent segmentation fault for unkown keywords in parallel.

blattms:
If there are unknown keywords and the parser throws an exception the we nevertheless broadcasted the eclipseState and schedule. Unfortunately, these might be null pointers in this case and the serializer will run into a segmentation fault.

Broadcasting is now only done if parsing was successful.

### DIFF
--- a/opm/simulators/utils/readDeck.cpp
+++ b/opm/simulators/utils/readDeck.cpp
@@ -324,9 +324,17 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
             eclipseState = std::make_unique<Opm::ParallelEclipseState>();
     }
 
+    // In case of parse errors eclipseState/schedule might be null
+    // and trigger segmentation faults in parallel during broadcast
+    // (e.g. when serializing the non-existent TableManager)
+    auto comm = Dune::MPIHelper::getCollectiveCommunication();
+    parseSuccess = comm.min(parseSuccess);
     try
     {
-        Opm::eclStateBroadcast(*eclipseState, *schedule, *summaryConfig);
+        if (parseSuccess)
+        {
+            Opm::eclStateBroadcast(*eclipseState, *schedule, *summaryConfig);
+        }
     }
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

opm/simulators/utils/readDeck.cpp
```diff
@@ -324,9 +324,16 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
             eclipseState = std::make_unique<Opm::ParallelEclipseState>();
     }
 
+    // In case of parse erorors eclipseState/schedule might be null
```
bska, line 327: Tiny nit: `erorors` should be `errors`.  Otherwise this looks good to me.

blattms: Sigh, my typing just sucks. Maybe this 4.5 finger system of mine does not work :smile: 
Done and force-pushed.

blattms:
jenkins build this please

blattms:
:flushed: Now?

bska:
> 😳 Now?

🚢 !

bska:
jenkins build this please

bska:

opm/simulators/utils/readDeck.cpp
```diff
@@ -344,8 +352,6 @@ void readDeck(int rank, std::string& deckFilename, std::unique_ptr<Opm::Deck>& d
         errorGuard->clear();
     }
 
-
-    auto comm = Dune::MPIHelper::getCollectiveCommunication();
     parseSuccess = comm.min(parseSuccess);
```
bska, line 355: My bad, I didn't see the implication here.  Moving the definition of `comm` into the `HAVE_MPI` section means that
```C++
parseSuccess = comm.min(parseSuccess);
```
becomes invalid in sequential builds.  We should probably protect the `comm.min()` call with `#if HAVE_MPI` too.

blattms:
broke sequential build. Will add another fix

PR #3500, opened 2021-09-01
Add GPMaint::State member to GroupState

joakim-hove:
Upstream: https://github.com/OPM/opm-common/pull/2643


joakim-hove:
jenkins build this opm-common=2643 please

PR #979, opened 2016-12-15
FlowMainEbos: fix the output of the initial condition

andlaus:
I probably did not see this warning when testing f7910af7d70 because I'm currently flooded by deprecation warnings stemming from Eigen. (these warnings are caused because I use Ubuntu 16.04 and an
old version of Eigen is cloned by the build system that is not system installed.)

@totto82: please merge if this fixes #976.

PR #3503, opened 2021-09-02
[bugfix] Fix sequential run after #979

blattms:
Defition of the communicator was moved inside a `#if HAVE_MPI` there. This commit queries the communicator again below to fix this.


blattms:
jenkins build this sequential please

PR #3507, opened 2021-09-04
Add Runspec argument to restart loader

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/2651


akva2:
upstream was merged. pulling this to restore master to buildable state.

PR #3508, opened 2021-09-06
drop aliases for Indices entries

akva2:
using Indices directly more clearly shows where the data comes
from without having to hop through hoops to do so.

This is a style thing more than anything else so personal preferences play a big role. Feel free to reject.


akva2:
jenkins build this please

PR #3511, opened 2021-09-07
Drop using statements

akva2:
Again this is mostly a style choice. Having huge lists of using statements is visual noise and such lists tend to be 'add only', meaning the lists keep growing over time and are seldom kept clean. Instead we qualify members using 'this' at call sites.


akva2:
jenkins build this please

blattms (approved):
Would be my preferred choice, too.

akva2:
jenkins build this please

bska:

PR #3514, opened 2021-09-09
Handle empty region in average calculator

totto82:
This should fix model 2 regression. 
A more elegant solution would be to make the RegionMapping.hpp parallel aware. But that is a task for another time and person. @bska or @blattms Cool if one of you could take a look at this some time.

### DIFF
--- a/opm/simulators/wells/RegionAverageCalculator.hpp
+++ b/opm/simulators/wells/RegionAverageCalculator.hpp
@@ -185,10 +185,11 @@ namespace Opm {
                           // using the pore volume to do the averaging
                           const auto& attri_pv = attributes_pv[reg];
                           const double pv_sum = comm.sum(attri_pv.pv);
-                          assert(pv_sum > 0.);
-                          const double p_pv_sum = comm.sum(attri_pv.pressure);
-                          ra.pressure = p_pv_sum / pv_sum;
-
+                          // pore volums can be zero if a fipnum region is empty
+                          if (pv_sum > 0) {
+                            const double p_pv_sum = comm.sum(attri_pv.pressure);
+                            ra.pressure = p_pv_sum / pv_sum;
+                          }
                       }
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

totto82:
I have tested this on model2 and it fixes the problem. Merge?

alfbr:
I can confirm that it fixes my crash on the CO2 model too. Since this is a bug making multiple models break, I support merging now.

blattms:
I am having some problems following. Looking at the patch and the comment, it seems like this bug should have also surfaced in a sequential run as the sum over all processes is already taken (and hence the porvolume should always be at as big in a parallel run as in a sequential one.

To me it seems that the function is already "parallel aware" (with lots of global communication, etc.), but is might not be very scalable as it is. Is that what you meant @totto82?

bska:
> To me it seems that the function is already "parallel aware"

Kind of.  The very foundational "region -> cell set" mapping is not aware of multiple processes.  Making this fully aware of parallel environments is a moderately big undertaking&mdash;my best guess is a that it will require at least a 3 full work days&mdash;but I expect the final outcome to be a lot better than the hacky facility we have at the moment.

blattms:
> The very foundational "region -> cell set" mapping is not aware of multiple processes

IMHO it shouldn't be aware of that. The only information that might help is what processes might have cells of a region. But maybe that is what you meant, anyway

alfbr:
> I am having some problems following. Looking at the patch and the comment, it seems like this bug should have also surfaced in a sequential run 

Yes, I can confirm that the bug was there in sequential runs, and that this PR fixes it for sequential runs.

totto82:
What I want is to be able to 
```
for (auto reg : regions)
 comm.sum(...)
```

without hanging since `regions ` means something different on the different ranks

PR #3519, opened 2021-09-10
Option for continue on solver convergence failure

totto82:
Sometimes you just want you're simulation to run through. 

Note 1. It is false pr default. 
Note 2. It is set to true if TUNING is enabled

### DIFF
--- a/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
+++ b/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
@@ -45,6 +45,10 @@ struct SolverMinTimeStep {
     using type = UndefinedProperty;
 };
 template<class TypeTag, class MyTypeTag>
+struct SolverContinueOnConvergenceFailure {
+    using type = UndefinedProperty;
+};
+template<class TypeTag, class MyTypeTag>
 struct SolverMaxRestarts {
     using type = UndefinedProperty;
 };
@@ -137,6 +141,10 @@ template<class TypeTag>
 struct SolverMinTimeStep<TypeTag, TTag::FlowTimeSteppingParameters> {
     using type = GetPropType<TypeTag, Scalar>;
     static constexpr type value = 1.0e-12;
+};
+template<class TypeTag>
+struct SolverContinueOnConvergenceFailure<TypeTag, TTag::FlowTimeSteppingParameters> {
+    static constexpr bool value = false;
 };
 template<class TypeTag>
 struct SolverMaxRestarts<TypeTag, TTag::FlowTimeSteppingParameters> {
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
Any comments on this?

opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
```diff
@@ -288,7 +297,8 @@ namespace Opm {
             , growthFactor_(tuning.TFDIFF)
             , maxGrowth_(tuning.TSFMAX)
             , maxTimeStep_(tuning.TSMAXZ) // 365.25
-            , minTimeStep_(unitSystem.to_si(UnitSystem::measure::time, EWOMS_GET_PARAM(TypeTag, double, SolverMinTimeStep))) // 1e-12;
+            , minTimeStep_(tuning.TSFMIN) // 0.1;
+            , ignoreConvergenceFailure_(true)
```
blattms, line 301: Looks strange. Shouldn't you query the parameter or use the default (`false`)

blattms:
Mathematicians might be a bit suspicious concerning this.

Can you explain what happens?
- Nonlinear solver takes  max number of iterations
- if it converged previously, timestep will become smaller
- What then? What is the effect? Might the errors sum up?

opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
totto82: The logic is as following. 
If --enable-tuning=true
we honor the TUNING keyword. 
where TSFMIN is the minimum step size for which all time steps are accepted regardless of convergence. 
That is at least my understanding. @OPMUSER Isn't that right?

blattms:
Just curious by the way...

opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
blattms: I see. Hence this will get set by the Tuning keyword somewhere? Maybe we should add a comment somewhere (for the case others are on my dumbness/smartness level)?

OPMUSER: Sorry for the late reply, been "enjoying: GitHub PRs :-(

@totto82 Yes that is correct, you get a message saying something like, None Convergence Time Step Accepted as Too Small to Chop`

OPMUSER:
Just to be clear:
2021-09-17 11_07_16-Clipboard
In case there is some confusion for the various parameters.

PR #3520, opened 2021-09-10
oil target may be limited by LRAT as well

totto82:
Fix assert


totto82:
jenkins build this please

PR #3521, opened 2021-09-10
only compute fip avg pressure when gpmaint

totto82:


totto82:
jenkins build this please

totto82:
This is uncontroversial, merging....

PR #3528, opened 2021-09-15
bugfix: Detect exception during beginReportStep in parallel & abort

blattms:
Previously, exceptions happening at this stage have deadlocked flow. E.g.  `UniformTabulated2DFunction` in opm-material throws a `NumericalIssue` if the values passed are outside the tabulated reason. This function is e.g. called in 2-phase CO2-storage cases during `BlackoilModel::initializeWellState`.

BTW: This is only the first step as it is not very user friendly that a simulation aborts at this stage.


blattms:
jenkins build this please

PR #3530, opened 2021-09-16
Wip: Added --throw-untabulated=Bool parameter to allow preventing aborts for tabulated values).

blattms:
For a CO2 model UniformTabulated2d would throw an exception during `BlackoilWellModel::initializeWellState` when a well was suddenly switched on.

To prevent that there is now a `--throw-untabulated=Bool` parameter for flow that allows the user to still get this behavior. The default value is false, which means a change to previous flow versions that would always throw.

There might be side effects of this (E.g. cases where throwing might be the correct default solution). If that is the case, the please yell.

Needs OPM/opm-material#470

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -131,6 +131,12 @@ struct EclBaseProblem {
 #endif
 }
 
+// Determine whether to throw for untabulated values
+template<class TypeTag, class MyTypeTag>
+struct ThrowUntabulated {
+    using type = UndefinedProperty;
+};
+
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this opm-material=470 please

blattms:
jenkins build this opm-material=470 please

blattms:
Seems ike this does not even fix the problem as for CO2 storage the setting code is never executed.

blattms:
Not needed anymore as the upstream PR was closed

PR #3531, opened 2021-09-17
support aquct and co2store

totto82:
Depends on OPM/opm-material#471 and OPM/opm-common#2694

If you want to run the test case in opm-tests you will also need OPM/opm-common#2691

### DIFF
--- a/opm/simulators/aquifers/AquiferCarterTracy.hpp
+++ b/opm/simulators/aquifers/AquiferCarterTracy.hpp
@@ -50,8 +50,6 @@ public:
     using typename Base::Simulator;
     using typename Base::ElementMapper;
 
-    using Base::waterCompIdx;
-    using Base::waterPhaseIdx;
     AquiferCarterTracy(const std::vector<Aquancon::AquancCell>& connections,
                        const Simulator& ebosSimulator,
                        const AquiferCT::AQUCT_data& aquct_data)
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

totto82:
Any comments on this? I plan to merge this later today if no objections.

opm/simulators/aquifers/AquiferCarterTracy.hpp
```diff
@@ -191,7 +215,20 @@ class AquiferCarterTracy : public AquiferInterface<TypeTag>
...
         this->pa0_ = this->aquct_data_.initial_pressure.value();
-        this->rhow_ = this->aquct_data_.waterDensity();
+        if(this->co2store_()) {
+             const auto press = this->aquct_data_.initial_pressure.value();
+
+             Scalar temp = FluidSystem::reservoirTemperature();
+             if (this->aquct_data_.initial_temperature.has_value())
+                 temp = this->aquct_data_.initial_temperature.value();
+
+             Scalar rs = 0.0; // no dissolved CO2
```
bska, line 225: The `rs` variable is unused.  Did you mean to pass this as an argument to `inverseFormationVolumeFactor()`?

bska (commented):
Why do we need to use the `oilPvt()` to calculate water properties in the CO2STORE case?  The aquifer is still primarily filled with water is it not?

totto82:
> Why do we need to use the oilPvt() to calculate water properties in the CO2STORE case? The aquifer is still primarily filled with water is it not?

Currently only the oil phase supports dissolved gas therefore CO2STORE "use" OIL for the water phase. This is the standard trick used for doing co2 storage simulations using a blackoil simulator.

opm/simulators/aquifers/AquiferCarterTracy.hpp
totto82: yes. thanks for noticing. I will update accordingly.

opm/simulators/aquifers/AquiferCarterTracy.hpp
```diff
@@ -159,6 +166,11 @@ class AquiferCarterTracy : public AquiferInterface<TypeTag>
         return std::make_pair(a, b);
     }
 
+    size_t pvtRegionIdx() const
```
bska, line 169: Please use `std::size_t` here instead.

totto82:
jenkins build this please

PR #3532, opened 2021-09-17
combine co2store and numerical aquifers

totto82:
This is a sufficient fix to make it possible to run CO2STORE and numerical aquifers together.


totto82:
jenkins build this please

bska (approved):
Looks good to me. I'll merge into master.

PR #3533, opened 2021-09-19
Create Action::State external to the vanguard

joakim-hove:
Create the `Action::State` external to the vanguard. This is a necessary step to load the action state variables from the restart file.


joakim-hove:
jenkins build this please

bska:
Did you intentionally postpone handling `actionState` in parallel?  This code looks like it will only create/initialise `actionState` on rank 0 while other ranks leave the object unchanged.  Come to think of it, the same restriction **seems** to apply to `udqState` too.

joakim-hove:
Hmmm - good point regarding parallell.

PR #3534, opened 2021-09-20
Use single well state

joakim-hove:
Use class `SingleWellState` instead of full `WellState` container in some locations. More to come.


joakim-hove:
jenkins build this please

PR #3535, opened 2021-09-21
add well_test_state to WGState

totto82:
If the simulator fails after a well opens due to well testing we want to also revert the wellTestState. Adding the wellTestState to the WGState class automatically includes it in the resetWGState and commitWGState logic.


totto82:
jenkins build this please

joakim-hove:
jenkins build this please

totto82:
jenkins build this please

joakim-hove:
Good that the logic can be reused and extended - merging!

PR #3537, opened 2021-09-21
update status after WTEST

totto82:
without this PR the well will open and be included in the well container, but the output will be trivial since the status is still SHUT.


totto82:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -369,7 +369,10 @@ namespace Opm {
 
                 const auto& events = schedule()[reportStepIdx].wellgroup_events();
                 const bool event = report_step_starts_ && events.hasEvent(well->name(), effective_events_mask);
-                if (event) {
+                const bool dyn_status_change = this->wellState().well(well->name()).status
```
joakim-hove, line 372: The `dyn_status_change` variable was a real mouthful - but I have no alternative suggestion.

joakim-hove:
Looks good - merging.

PR #3540, opened 2021-09-21
Initialize blackoil simulator from Opm::Schedule shared with Python.

hakonhagland:
Depends on https://github.com/OPM/opm-common/pull/2700

Adds a new constructor to `Main.hpp` that takes shared pointers to `Deck`, `EclipseState`, `Schedule`, and `SummaryConfig`. This makes it possible to share these variables with Python without worrying about lifetime issues of the underlying C++ objects.  For example, a Python script can first create an `opm.io.schedule.Schedule` object which is then modified from Python. Then, assume the same Python script creates an `opm.simulators.BlackOilSimulator` which is initialized with the same schedule object. Since the underlying C++ object is a shared pointer, the schedule object in Python may go out of scope (get deleted by Python) without having the C++ schedule object being deleted. And the Python `BlackOilSimulator` may continue to be used after the Python Schedule object has been deleted since it still has a valid C++ schedule object. Here is an example:
```Python
from opm.io.parser import Parser
from opm.io.ecl_state import EclipseState
from opm.io.schedule import Schedule
from opm.io.summary import SummaryConfig
from opm.simulators import BlackOilSimulator

deck  = Parser().parse('SPE1CASE1.DATA')
state = EclipseState(deck)
schedule = Schedule( deck, state )
summary_config = SummaryConfig(deck, state, schedule)
sim = BlackOilSimulator( deck, state, schedule, summary_config  )
del schedule     # the schedule object may be deleted now and the simulator still has a valid Opm::Schedule
sim.step_init()
sim.step()
# ... 
sim.step_cleanup()
```


hakonhagland:
jenkins build this opm-common=2700 please

bska (changes requested):
Since the `set*` function assume ownership of the resource and don't change it, we should pass the argument by value and move it into the global
```C++
void EclGenericVanguard::setExternalSchedule(std::shared_ptr<Schedule> schedule)
{
    externalSchedule_ = std::move(schedule);
}
```

hakonhagland:
> Since the `set*` function assume ownership of the resource and don't change it,we should pass the argument by value and move it into the global

@bska But the point of the change in this PR is that it should not assume ownership? I changed from `unique_ptr` to `shared_ptr` so the vanguard object can share the object reference with Python. As I understand it, the vanguard should just increase the use count of the shared pointer by one ?

bska:
> @bska But the point of the change in this PR is that it should not assume ownership?

Assigning a `shared_ptr<>` automatically makes the assignee part of the pool that owns the shared object therefore it's appropriate to say that `set*()` assumes ownership&mdash;just not exclusive ownership.  The right way to convey those semantics is to pass the `shared_ptr<>` by value, which increments the use-count, and then to move the new value into the final destination.  Passing a reference to a `shared_ptr<>` is only appropriate if the function makes a change to the `shared_ptr<>` itself such as resetting the target and if the caller is aware that the `shared_ptr<>` object might/will change.

> The vanguard should just increase the use count of the shared pointer by one

We achieve that goal by passing the `shared_ptr<>` by value.  If the caller then goes on to say
```C++
setExternalSchedule(std::move(my_schedule_ptr))
```
then that fully transfers the object and makes the ownership exclusive/unique instead of shared.  In both cases, taking the `shared_ptr<>` by value is the right parameter type.

hakonhagland:
> The right way to convey those semantics is to pass the shared_ptr<> by value

@bska I think I understand what you are saying, this makes the code more readable and less error-prone. Thanks for explaning. 

> If the caller then goes on to say
>
>            setExternalSchedule(std::move(my_schedule_ptr))
>
> then that fully transfers the object and makes the ownership exclusive/unique instead of shared

Ok, so even in `Main.hpp` we should use `std::move()` on the shared pointer when passing them to the vanguard? For example, on line 88:

https://github.com/OPM/opm-simulators/blob/master/opm/simulators/flow/Main.hpp#L88

I note that using `std::move()` here will make the internal `std::shared_ptr` stored as a class variable in `Main.hpp` invalid.
What I do not understand is why we transfer these pointer to the vanguard? It is not clear to me why they cannot be owned by `Main.hpp` (or have shared ownership with `Main.hpp`).

bska:
> > If the caller then goes on to say
> > ```C++
> > setExternalSchedule(std::move(my_schedule_ptr))
> > ```
> > then that fully transfers the object and makes the ownership exclusive/unique instead of shared
> 
> Ok, so even in `Main.hpp` we should use `std::move()` on the shared pointer when passing them to the vanguard?

Sorry for being unclear&mdash;that's not what I wanted to say.  Rather, I meant to convey that when you create a function `f` with the signature
```C++
Y f(std::shared_ptr<T> x)
```
for some types `T` and `Y` (e.g., `Schedule` and `void` respectively) in which the `shared_ptr<>` is explicitly passed by value then you're signalling to the caller that `f` is likely going to retain a copy of `x` in some internal structure and will continue to participate in the lifetime/ownership of the shared object.  This then allows the caller, at the call site, to decide whether to fully transfer ownership (by passing `std::move(ptr)`) or to continue participating in managing the lifetime of the shared object by just passing the `ptr` that they already must have at that point.

> I note that using `std::move()` here will make the internal `std::shared_ptr` stored as a class variable in `Main.hpp` invalid.

True.  If we need shared ownership then we should not pass the pointers by `std::move()` but rather as just as plain arguments.

> It is not clear to me why [the large objects] cannot be owned by `Main.hpp` (or have shared ownership with `Main.hpp`).

For better of worse `Main.hpp` is not the only way to launch a simulation process.  The expert user can choose to invoke a separate binary that is specific to their use case and in that case the other binary will be responsible for reading the input file.  On the other hand, `Main.hpp` offers a convenient wrapper that

1. Reads the input file
2. Determines which particular simulation function to run
3. Invokes that function with the partially formed core simulation objects.

This organisation means we don't read the input file twice in the common case of dynamic dispatch through `Main.hpp`.  The downside is that if we were to pass the simulation objects as function call parameters to the core simulation function instead of through the side channel/out-of-bound `set*()` functions then the separate binaries would need to duplicate the code to read the input file and form those core simulation objects.

hakonhagland:
jenkins build this please

hakonhagland:
The Jenkins build [reports](https://ci.opm-project.org/job/opm-simulators-PR-builder/2955/testReport/junit/(root)/mpi/python_schedule/) one failed test. 
```
from opm.io.parser import Parser
ModuleNotFoundError: No module named 'opm'
```
This is from the [`test_schedule.py`](https://github.com/OPM/opm-simulators/blob/4c5245196bc5033820a80651302c91fb7e1e6d01/python/test/test_schedule.py#L7) test. I have added the `opm-common` build directory to `PYTHONPATH` in [`CMakeLists.txt`](https://github.com/OPM/opm-simulators/blob/4c5245196bc5033820a80651302c91fb7e1e6d01/python/simulators/CMakeLists.txt#L40) so it should be able to find the module and it works fine on my computer. Any idea why it may fail on the test server?

akva2:
jenkins does not use sibling builds, rather it uses installed builds. you have to
1) make sure opm-common installs the python binding
2) modify the PYTHONPATH in the jenkins builder script

since i probably know the hows better, i'll give it a whirl as soon as my brain is operational (needs the first cup of coffee to boot).

akva2:
jenkins build this opm-common=2707 please

akva2:
jenkins build this opm-common=2707 please

akva2:
there u go.

hakonhagland:
@akva2 Thanks!

flow/flow_ebos_foam.cpp
```diff
@@ -43,19 +43,19 @@ struct EnableFoam<TypeTag, TTag::EclFlowFoamProblem> {
...
 
     Vanguard::setExternalSetupTime(setupTime);
-    Vanguard::setExternalDeck(std::move(deck));
-    Vanguard::setExternalEclState(std::move(eclState));
-    Vanguard::setExternalSchedule(std::move(schedule));
-    Vanguard::setExternalSummaryConfig(std::move(summaryConfig));
+    Vanguard::setExternalDeck(deck);
+    Vanguard::setExternalEclState(eclState);
+    Vanguard::setExternalSchedule(schedule);
+    Vanguard::setExternalSummaryConfig(summaryConfig);
```
bska, line 58: Technically, these could still be `std::move()`s.  The by-value parameter to `flowEbosFoatSetDeck()` will increase the `shared_ptr<>`'s use-count by one and that's all we really need here.  Passing the pointers on by value will then increase the use-count by one more&mdash;meaning the use-count is temporarily increased by a total of two&mdash;until `flowEbosSetDeck()` returns.

bska (commented):
Thanks, this is a clear improvement.  Some of the existing `std::move()` calls should still remain in my opinion just to avoid unnecessarily raising `shared_ptr<>` use-counts by two.

On a tangential note it's also not entirely clear to me what kind of changes you expect the Python layer to make to a `Schedule` object.  If those changes could potentially involve UDQs or action blocks then we may need to expose/share the UDQ and Action states too.  That could be follow-up work of course.

atgeirr:
Is it necessary (or much more convenient) to add the SPE1 case here, instead of getting it from opm-tests?

hakonhagland:
jenkins build this please

opm/simulators/flow/python/PyBlackOilSimulator.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include <opm/parser/eclipse/Deck/Deck.hpp>
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
+#include <opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp>
+
+namespace Opm::Pybind {
+class PyBlackOilSimulator
+{
+    using TypeTag = Opm::Properties::TTag::EclFlowProblem;
+private:
```
bska, line 37: The `TypeTag` is implicitly `private:` too.  Maybe you'd like to explicitly declare it as such?

python/simulators/PyBlackOilSimulator.cpp
```diff
@@ -0,0 +1,162 @@
...
+    , schedule_{std::move(schedule)}
+    , summary_config_{std::move(summary_config)}
+{
+}
+
+const Opm::FlowMainEbos<typename Opm::Pybind::PyBlackOilSimulator::TypeTag>&
+         PyBlackOilSimulator::getFlowMainEbos() const
+{
+    if (this->mainEbos_) {
+        return *(this->mainEbos_.get());
```
bska, line 59: Style: `std::shared_ptr<>` defines a dereference operation so you could just write this as
```C++
return *this->mainEbos_;
```
instead.

python/simulators/PyBlackOilSimulator.cpp
```diff
@@ -0,0 +1,162 @@
...
+{
+    std::size_t size_ = array.size();
+    const double *poro = array.data();
+    materialState_->setPorosity(poro, size_);
+}
+
+int PyBlackOilSimulator::step()
+{
+    if (!hasRunInit_) {
+        throw std::logic_error("step() called before step_init()");
```
bska, line 91: You should include `<stdexcept>` to explicitly bring a declaration of `std::logic_error` into scope.

bska (approved):
Thanks for the updates, this looks good to me now.  I just have a couple of stylistic comments in the implementation of `PyBlackOilSimulator` and then I'm ready to merge this into the master branch.

hakonhagland:
jenkins build this please

python/simulators/PyBlackOilSimulator.cpp
hakonhagland: Thanks for the comments, they look good to me. I have pushed a new commit to implement these.

hakonhagland:
jenkins build this please

bska (approved):
Thanks!  I'm satisfied with the updates and I'll merge into master.

PR #3545, opened 2021-09-24
[bugfix] Fix deadlocks because of parallel try-catch for some CO2 storage cases

blattms:
For a CO2 storage case excpeptions are thrown in opm-material because the passed parameters are unreasonable. Unfortunately, this made the models deadlock because there were problems in the exception handling:

- collective communication happened in the try-catch clause. As we also perform collective communication in/after the catch, processes were waiting at different collective communication calls forever
- Each try-catch had its own exceptions that were catched and only these. If there was an unknown exception it simply was not caught which also cause a deadlock if the exception only happened on a few processes. Those never reached the collective communication after the catch.

This PR introduces macros for the parallel-try-catch statements to make sure that the list of exceptions caught is consistent, it also
has a catch clause that will catch anything, and it makes the model with the problems abort (with signal 6 now) instead of running into a deadlock:
```
terminate called after throwing an instance of 'Opm::NumericalIssue'
  what():  Attempt to get tabulated value for (336.650000, 81706.071880) on a table of extend 280.000000 to 400.000000 times 100000.000000 to 100000000.000000
[smaug:90429] *** Process received signal ***
```
Interestingly this happens at an earlier report step than the deadlock before.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -68,6 +68,8 @@
 
 #include <opm/core/props/satfunc/RelpermDiagnostics.hpp>
 
+#include <opm/simulators/utils/DeferredLoggingErrorHelpers.hpp>
+
 #include <opm/models/utils/pffgridvector.hh>
 #include <opm/models/blackoil/blackoilmodel.hh>
 #include <opm/models/discretization/ecfv/ecfvdiscretization.hh>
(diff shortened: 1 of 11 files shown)


blattms:
jenkins build this please

totto82:
Thanks. With this the simulator will still abort when it hits an exception that was not cached before but no longer deadlock in parallel. i.e. this will not change the catching in serial. Or have I misunderstood something.

blattms:
Exactly. Serial behaviour should be the same as before. This PR only fixes the deadlock in parallel.

Actually, I should do another run first just to be sure it aborts at the same time as before
in serial.

blattms:
At least on tests for other cases this did not change things. Merging this after discussion with @totto82

PR #3546, opened 2021-09-24
updateWsolvent only after well container is created

totto82:


totto82:
jenkins build this please

bska (approved):
Thanks.  This makes sense to me and I'll merge into master.  I too agree that we should have hit this condition earlier.

PR #3547, opened 2021-09-24
Make wtest more robust

totto82:
We only want to open wells that can operate under both economical and physical reasons.

Depens on OPM/opm-common#2711

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -214,36 +214,13 @@ namespace Opm
     void
     WellInterface<TypeTag>::
     wellTesting(const Simulator& simulator,
-                const double simulation_time, const int report_step,
-                const WellTestConfig::Reason testing_reason,
+                const double simulation_time,
                 /* const */ WellState& well_state,
                 const GroupState& group_state,
                 WellTestState& well_test_state,
                 DeferredLogger& deferred_logger)
     {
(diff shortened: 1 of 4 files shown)


opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -261,13 +238,27 @@ namespace Opm
...
+            bool converged = solveWellForTesting(simulator, well_state_copy, group_state, deferred_logger);
+            if (!converged) {
+                const std::string msg = " well " + this->name() + " did not get converged during well testing";
+                deferred_logger.debug(msg);
+                return;
+            }
+
+            updateWellOperability(simulator, well_state_copy, deferred_logger);
+            if ( !this->isOperable() ) {
+                const std::string msg = " well " + this->name() + " is not operable during well testing for physical reason";
```
joakim-hove, line 250:
```C++
auto msg = fmt::format("Well {} is not operable during well testing for physical reason", this->name());
```

joakim-hove:
Upstream in opm-common is ready to be merged.

opm/simulators/wells/WellInterface_impl.hpp
bska: If we do update this, can we rephrase the message to be less ambiguous too?  E.g, something along the lines of

> WTEST: Well &lt;name&gt; is not operable (physical reasons)

bska:
jenkins build this please

bska (approved):

PR #3550, opened 2021-09-27
remove cartesian to compressed vector

totto82:
Remove an unnecessary global vector to reduce memory for the 20 million cell case.


totto82:
jenkins build this please

totto82:
jenkins build this please

blattms:
I had the same idea, just did not find the time. We just need to be sure that the the map in vanguard is actually updated correctly before. (At least in other places of the code there is also the problem, that sometimes we work on the unpartitioned grid and sometimes not. Not sure whether the vanguard always has the correct one.)

blattms:
Thanks. Well, actually I missed this one. There are to many names we are using for this (cartesianToCompressed, cartesian_to_compresses_, globalToLocal, ...).

PR #3551, opened 2021-09-27
Use SingleWellState to manage parallel well info

joakim-hove:
Simplification ... ?

### DIFF
--- a/opm/simulators/wells/SingleWellState.cpp
+++ b/opm/simulators/wells/SingleWellState.cpp
@@ -21,8 +21,9 @@
 
 namespace Opm {
 
-SingleWellState::SingleWellState(bool is_producer, std::size_t num_perf, std::size_t num_phases, double temp)
-    : producer(is_producer)
+SingleWellState::SingleWellState(const ParallelWellInfo* pinfo, bool is_producer, std::size_t num_perf, std::size_t num_phases, double temp)
+    : parallel_info(pinfo)
+    , producer(is_producer)
     , temperature(temp)
     , well_potentials(num_phases)
     , productivity_index(num_phases)
--- a/opm/simulators/wells/SingleWellState.cpp
+++ b/opm/simulators/wells/SingleWellState.cpp
@@ -67,6 +67,25 @@ void SingleWellState::open() {
     this->status = Well::Status::OPEN;
 }
 
+
+double SingleWellState::sum_connection_rates(const std::vector<double> connection_rates) const {
+    return this->parallel_info.get().sumPerfValues(connection_rates.begin(), connection_rates.end());
+}
+
+double SingleWellState::sum_brine_rates() const {
+    return this->sum_connection_rates(this->perf_data.brine_rates);
+}
+
+
+double SingleWellState::sum_polymer_rates() const {
+    return this->sum_connection_rates(this->perf_data.polymer_rates);
+}
+
+
+double SingleWellState::sum_solvent_rates() const {
+    return this->sum_connection_rates(this->perf_data.solvent_rates);
+}
+
 }
 
 
--- a/opm/simulators/wells/SingleWellState.cpp
+++ b/opm/simulators/wells/SingleWellState.cpp
@@ -21,7 +21,7 @@
 
 namespace Opm {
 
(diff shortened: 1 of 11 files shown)


joakim-hove:
jenkins build this please

bska:
I haven't looked at this at all, but please don't add new [shadowing variable warnings](https://ci.opm-project.org/job/opm-simulators-PR-builder/2974/gcc/fileName.1802511049/).

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

blattms:
At least there need to be BIG BIG BIG BIG warnings in the documentation about all methods that are calling methods using communication. Hiding communication like this (method name does not make it clear that there is communication, no documentation) will most probably lead to hard to debug problems/deadlocks.

joakim-hove:
> Hiding communication like this (method name does not make it clear that there is communication, no documentation) will most probably lead to hard to debug problems/deadlocks.

I agree on that - and I am happy to change names and add warnings for the communicating functions - but the communication was quite well hidden before as well - in my opinion.

blattms:
Indeed and agreed.

Should reference_wrapper really be used in this case, is it needed? Why? IMHO it makes the code more complicated and adds another level of indirection. Is there a recommendation of when one should use it (only found https://www.nextptr.com/tutorial/ta1441164581/stdref-and-stdreference_wrapper-common-use-cases) and really never thought about it before.

joakim-hove:
> Should reference_wrapper really be used in this case, is it needed? Why?

Well - no hard and absolute rules on this. My reason for changing to reference wrapper was to get reference semantics - i.e. the code signals that the `ParallelWellInfo` instance must be initialized to a valid value, whereas the pointer could be null (however in this particular case it could not be null - then things would come tumbling down). So all in all I think reference_wrapper makes sense in this situation - but I am happy to change it back to pointer.

blattms:
I see. I guess if code is safer it might be a bit more complicated, but:

Even with the change the reference may be dangling if `pwell` in `std::ref(*pwell)` is a nullptr. Hence maybe change everything to reference_wrapper or nothing? Going half way might complicate things (dangling references seem worse than null pointers to me).

joakim-hove:
I *think* the current PR is all the way to reference wrapper - in the expression `std::ref(*pwell)` `pwell` is an iterator and not a pointer.

joakim-hove:
OK; I have now looked more into this. In the case of convergence problems and eventual bail out [this](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp#L518) is the exception thrown. This is unhandled and will currently bubble all the way up to program termination. So with this PR output will be like:
```
Program threw an exception: [/home/hove/work/OPM/opm-simulators/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp:518] Solver failed to converge after cutting timestep 10 times.

  Process name: [[47497,1],1]
  Exit code:    1
```
i.e. an mpi abort message will be printed for the other failing processes. I agree that mpi abort message is not particularly pretty, but I still think the deadlock safetybelt in this PR is worth it; and most importantly *I want to be able to use exceptions when developing in opm-simulator*.

joakim-hove:
OK - I have renamed the communicating functions and added a comment about the fact.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

blattms:
Indeed. Thanks for the clarification.

opm/simulators/wells/SingleWellState.cpp
```diff
@@ -66,6 +67,25 @@ void SingleWellState::open() {
     this->status = Well::Status::OPEN;
 }
 
+
+double SingleWellState::sum_connection_rates(const std::vector<double> connection_rates) const {
```
bska, line 71: Was the parameter type here supposed to be a reference (`const vector<>&`) or did you in fact mean to create an independent copy of the input object.

joakim-hove: It should have been const ref - PR added.

PR #3552, opened 2021-09-27
cusparsesolver wrong size and type in cols,rows fixed

fgfuchs:
cusparsesolver: wrong size and type in cols,rows fixed

### DIFF
--- a/opm/simulators/linalg/bda/cusparseSolverBackend.cu
+++ b/opm/simulators/linalg/bda/cusparseSolverBackend.cu
@@ -227,8 +227,8 @@ void cusparseSolverBackend<block_size>::initialize(int N, int nnz, int dim) {
     cudaMalloc((void**)&d_t, sizeof(double) * N);
     cudaMalloc((void**)&d_v, sizeof(double) * N);
     cudaMalloc((void**)&d_bVals, sizeof(double) * nnz);
-    cudaMalloc((void**)&d_bCols, sizeof(double) * nnz);
-    cudaMalloc((void**)&d_bRows, sizeof(double) * (Nb + 1));
+    cudaMalloc((void**)&d_bCols, sizeof(int) * nnzb);
+    cudaMalloc((void**)&d_bRows, sizeof(int) * (Nb + 1));
     cudaMalloc((void**)&d_mVals, sizeof(double) * nnz);
     cudaCheckLastError("Could not allocate enough memory on GPU");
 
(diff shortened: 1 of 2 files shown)


fgfuchs:
jenkins build this please

opm/simulators/linalg/bda/cusparseSolverBackend.cu
```diff
@@ -227,8 +227,8 @@ void cusparseSolverBackend<block_size>::initialize(int N, int nnz, int dim) {
     cudaMalloc((void**)&d_t, sizeof(double) * N);
     cudaMalloc((void**)&d_v, sizeof(double) * N);
     cudaMalloc((void**)&d_bVals, sizeof(double) * nnz);
-    cudaMalloc((void**)&d_bCols, sizeof(double) * nnz);
-    cudaMalloc((void**)&d_bRows, sizeof(double) * (Nb + 1));
+    cudaMalloc((void**)&d_bCols, sizeof(int) * nnzb);
+    cudaMalloc((void**)&d_bRows, sizeof(int) * (Nb + 1));
```
bska, line 231: Would it be possible to write this as
```C++
cudaMalloc((void**)&d_bCols, nnzb * sizeof *d_bCols);
cudaMalloc((void**)&d_bRows, (Nb + 1) * sizeof *d_bRows);
```
instead?  Ignoring the fact that `(void**) &some_pointer` is a constraint violation unless `some_pointer` is the exact type `void*`, I always prefer to use `sizeof object` instead of `sizeof(type)` when I have to use a `malloc`-like interface.

fgfuchs: I agree, I plan a separate PR regarding this to clearly separate this issue. Otherwise, there are a lot of places where this needs to be changed. Would that be an OK solution for you?

bska: > I plan a separate PR regarding this to clearly separate this issue. [...] Would that be an OK solution for you?

Perfect!

blattms (commented):
Thanks. Good catch. @Tongdongq would you mind double checking (e.g. that there is no assumption in the code that needs bigger arrays, which I doubt?

Tongdongq:
The cusparse blocked functions use a blocked sparsity pattern, so this looks good.
I don't think the opencl counterparts have any similar issues.

atgeirr:
Thanks all, merging.

PR #3553, opened 2021-09-27
Communicators2

ElyesAhmed:
The opm simulators support a variable communicator instead of the fixed MPI_COMM_WORLD. A demo code has been added to main to test the use of non-global communicators. The compilation was done with and without USE_MPI and the tests with mpirun also.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -32,6 +32,12 @@
 
 #include <fmt/format.h>
 
+using MPIComm = typename Dune::MPIHelper::MPICommunicator;
+#if DUNE_VERSION_NEWER(DUNE_COMMON, 2, 7)
+    using Communication = Dune::Communication<MPIComm>;
+#else
+    using Communication = Dune::CollectiveCommunication<MPIComm>;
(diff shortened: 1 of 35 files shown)


blattms:
Unfortunately, you will need to rebase this to resolve the conflicts. Would be nice if you would do that.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -32,6 +32,12 @@
 
 #include <fmt/format.h>
 
+using MPIComm = typename Dune::MPIHelper::MPICommunicator;
+#if DUNE_VERSION_NEWER(DUNE_COMMON, 2, 7)
+    using Communication = Dune::Communication<MPIComm>;
+#else
+    using Communication = Dune::CollectiveCommunication<MPIComm>;
+#endif
```
blattms, line 40: Please don't do that in the global namespace. The name `Communication` is rather generic and could be used somewhere else (including 3rd party libraries). Maybe doing it in an Opm::Parallel namespace is better and would accomplish the same?

ElyesAhmed:
OK!
________________________________

atgeirr:
Superseded by #3571, closing.

PR #3554, opened 2021-09-27
Remove unused query/switch code for GLO - it is always on

joakim-hove:
The `WellState` class had a `bool` switch whether to perform gas lift optimization - it was always on. Just removed with this PR in the long time quest of simplifying the `WellState` class. If runtime switching of the GLO should be brought back in, I don't think the `WellState`is the correct home for such a future variable?


joakim-hove:
jenkins build this please

totto82:
Thanks. It used to be a debugging tool but I guess it is not needed anymore.

PR #3556, opened 2021-09-28
Call MPI_Abort() in main catch all clause

joakim-hove:
In current master it is difficult to use exceptions in the simulator code, because when only when process throws an exception the other processes will quite quickly deadlock in a subsequent collective communication.

With this PR `MPI_Abort(MPI_COMM_WORLD)`will be called in the catch all clause of flow execution and we can safely use arbitrary `std::exceptions` in the simulator. (The simulator will come tumbling down in flames - but I consider that far superior to a deadlock).

Exactly how this should be done is maybe outside my MPI competence - but I *really* hope something like this can be merged.

### DIFF
--- a/opm/simulators/flow/FlowMainEbos.hpp
+++ b/opm/simulators/flow/FlowMainEbos.hpp
@@ -431,7 +431,10 @@ namespace Opm
                         std::cout << message.str() << "\n";
                     }
                 }
-
+#if HAVE_MPI
+                if (this->mpi_size_ > 1)
+                    MPI_Abort(MPI_COMM_WORLD, EXIT_FAILURE);
+#endif
                 return EXIT_FAILURE;
             }
         }


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

atgeirr:
I agree that something like this should go in to protect users from our errors, but we must be aware that it addresses the symptom rather than the cause. Whenever a single process can fail, there should (eventually) be communication to check status between all processes. We have made significant improvements in this direction before, but there are still places where this is not done right.

joakim-hove:
I agree that the principle "Whenever one prosess can fail ....", but it is a long way to go - and unfortunately that principle can only be enforced with developer discipline/review - not very scalable.

As I see this PR - or something similar - provides an extra protection against deadlocks which can only be valuable?

joakim-hove:
The intention is that only otherwise unhandled exceptions should end up here; although I am not yet 100% certain og the exact way out in the case of multiple failed iterations.

joakim-hove:
OK; I have now looked more into this. In the case of convergence problems and eventual bail out this is the exception thrown. This is unhandled and will currently bubble all the way up to program termination. So with this PR output will be like:

Program threw an exception: [/home/hove/work/OPM/opm-simulators/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp:518] Solver failed to converge after cutting timestep 10 times.

  Process name: [[47497,1],1]
  Exit code:    1

i.e. an mpi abort message will be printed for the other failing processes. I agree that mpi abort message is not particularly pretty, but I still think the deadlock safetybelt in this PR is worth it; and most importantly I want to be able to use exceptions when developing in opm-simulator.

blattms:
doesn't d18414b introduce the possibility of deadlock again?

joakim-hove:
> doesn't d18414b introduce the possibility of deadlock again?

That is removed again.

blattms:
Thanks. So if there is no message from every process this can be merged.

blattms:
jenkins build this please

joakim-hove:
> So if there is no message from every process this can be merged.

Well; there is a message from every process in those cases which would otherwise deadlock - but I assume that is OK.

blattms:
> Well; there is a message from every process in those cases which would otherwise deadlock - but I assume that is OK.

that means 20 messages if the timestepping breaks down on a 20 process run. I do not think that is ok in  a production run.

joakim-hove:
> that means 20 messages if the timestepping breaks down on a 20 process run

No you misunderstand - in the case of timestep breakdown *all* processes will throw an exception - and they will remain silent. It is only the processes which have been forcefully shut down due to `MPI_Abort()` which will complain[1], and in that case deadlock is/was the alternative

[1]: Well - that is at least how it is on my machine.

blattms:
Seems like recent OpenMPI version show only one help message:
```
Program threw an exception: [/home/mblatt/src/dune/opm/opm-simulators/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp:518] Solver failed to converge after cutting timestep 10 times.
MPI_ABORT was invoked on rank 4 in communicator MPI_COMM_WORLD
with errorcode 1.

[smaug:11746] 31 more processes have sent help message help-mpi-api.txt / mpi-abort
[smaug:11746] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

Without this patch we get:
```
Program threw an exception: [/home/mblatt/src/dune/opm/opm-simulators/opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp:518] Solver failed to converge after cutting timestep 10 times.

  Process name: [[21000,1],30]
  Exit code:    1
```
Seems ot be of the same quality.
Still we have to be aware that deadlocks might still be happening depending on where the exception is thrown. E.g. this change
```
diff --git a/opm/simulators/flow/BlackoilModelEbos.hpp b/opm/simulators/flow/BlackoilModelEbos.hpp
index e68835608..23f4ac708 100644
--- a/opm/simulators/flow/BlackoilModelEbos.hpp
+++ b/opm/simulators/flow/BlackoilModelEbos.hpp
@@ -419,6 +419,8 @@ namespace Opm {
         SimulatorReportSingle assembleReservoir(const SimulatorTimerInterface& /* timer */,
                                                 const int iterationIdx)
         {
+            if(ebosSimulator_.gridView().grid().comm().rank()==4)
+            throw NumericalIssue("test");
             // -------- Mass balance equations --------
             ebosSimulator_.model().newtonMethod().setIterationIndex(iterationIdx);
             ebosSimulator_.problem().beginIteration();
```
will still result in a deadlock with 32 processes. Probably because all other proceses are in one of our Parallel catch clauses that do global communication and this one is in a different one.

Out of curiosity where did your problem/exception happen?

Maybe eventually we should implement some kind of debug mode where we use asynchronous collective communication with an implemented timeout mechanism. That migh actually prevent deadlocjs.

blattms (approved):
I think this is fine if it fixes some problems

joakim-hove:
> Out of curiosity where did your problem/exception happen?

My problems have typically been in well code; while developing & testing. And also when I have tried - and backed down - from replacing `assert( ... )` with `throw xxx`[1].

[1]: When we by default build with asserts enabled this is not so pressing any longer.

blattms:
> When we by default build with asserts enabled this is not so pressing any longer.
Don't we do that since the middle of August?

PR #3557, opened 2021-09-29
Use const reference

joakim-hove:
Fixup: https://github.com/OPM/opm-simulators/pull/3551#pullrequestreview-766401806


joakim-hove:
jenkins build this please

bska (approved):
Thanks for the rapid fix!

PR #3560, opened 2021-09-29
fix date for large time steps also for the interface functions

totto82:


totto82:
jenkins build this please

bska (approved):
At some point we should switch these to using the `TimeService` facilities and `format()`'s support for outputting time-points.  We do this when recording [restart output](https://github.com/OPM/opm-common/blob/05bac5f59a443c142a4bb5b2965a7f107aa0d9b2/src/opm/output/eclipse/RestartIO.cpp#L738) in the various log streams.  For now though, this does indeed fix very confusing output for simulation runs with long time horizons (e.g., CO2 sequestration) so I'll merge into master.

Issue #3555, opened 2021-09-28
2021-10-PRE WECON Issue


OPMUSER:
Again testing the ACTIONX facility with the intention of using the WECON keyword to shut-in connections when the well GOR is above a certain value and then use ACTIONX with WTEST to re-open the wells at a latter date. However, WECON kills all the wells prematurely due to numerical issue after the WECON keyword is processed. The part of the input deck of interest is:
```
--
--       ADVANCE SIMULATION BY REPORTING TIME
--
--       DAILY
TSTEP                                                                           
         30*10.0
/
--      
--       WELL ECONOMIC CRITERIA FOR PRODUCTION WELLS
-- WELL  MIN    MIN    MAX    MAX    MAX    CNTL    END   WELL                    
-- NAME  ORAT   GRAT   WCUT   GOR    WGR    MODE    RUN   NAME                    
WECON                                                                           
'OP*'    1*     1*     0.95   3.0    1*     CON    'NO'                        / 
/                                                                               
--
--       WELL PRODUCTION WELL CONTROLS (MAX OIL RATE SET 100 STB/D)                                                  
--                                                                              
-- WELL  OPEN/  CNTL   OIL    WAT    GAS   LIQ    RES    BHP   THP   VFP    VFP  
-- NAME  SHUT   MODE   RATE   RATE   RATE  RATE   RATE   PRES  PRES  TABLE  ALFQ 
WCONPROD                                                                    
'OP*'    OPEN   ORAT   100    1*     1*    1*     1*     1000.                 /
/
```
And here is some strange error messages:
```
Starting time step 0, stepsize 10 days, at day 360/370, date = 27-Dec-2021

Error: [/home/david/OPM/GitHub/opm-simulators/opm/simulators/flow/BlackoilModelEbos.hpp:319] NaN residual found!

Problem: Solver convergence failure - Numerical problem encountered
Timestep chopped to 3.300000 days
```
Since all the wells are shut-in then results are garbage compared to the commercial simulator. The test case is base on SPE09 and  the deck and results attached.

[ACTIONX_WTEST.zip](https://github.com/OPM/opm-simulators/files/7241062/ACTIONX_WTEST.zip)

totto82:
What happens is that the deck tries to open wells with WCONPROD where all connection of these wells are earlier shut due to WECON with "CON".  Wells without connections does naturally not converge. What behavior do we want here? Opening all earlier closed connections or don't open the well if all connections are closed.

[...]

PR #3563, opened 2021-09-30
open completions after well event

totto82:
depends on https://github.com/OPM/opm-common/pull/2723
solves https://github.com/OPM/opm-simulators/issues/3555


joakim-hove:
Well code is fine - whether this is the correct behavior is beyond me - but I'll take your word for it.

joakim-hove (approved):

PR #3564, opened 2021-09-30
communicate before usage not after

totto82:
Avoid deadlock in parallel when wells are shut.


totto82:
jenkins build this please

totto82:
This is a trivial bug fix. merging...

PR #3565, opened 2021-09-30
[refactor] Move cell centroid lookup magic to vanguard for reuse.

blattms:
In addition to transmissibilities We will need this for the tracers,
too, and should prevent code duplication.

### DIFF
--- a/ebos/eclalugridvanguard.hh
+++ b/ebos/eclalugridvanguard.hh
@@ -102,7 +102,6 @@ public:
 
     ~EclAluGridVanguard()
     {
-        delete cartesianIndexMapper_;
         delete equilCartesianIndexMapper_;
         delete grid_;
         delete equilGrid_;
@@ -178,6 +177,18 @@ public:
     const EquilCartesianIndexMapper& equilCartesianIndexMapper() const
     { return *equilCartesianIndexMapper_; }
 
+    /*!
+     * \brief Get function to query cell centroids for a distributed grid.
+     *
+     * Currently this only non-empty for a loadbalanced CpGrid.
+     * It is a function return the centroid for the given element
+     * index.
+     */
+    std::function<std::array<double,dimensionworld>(int)>
+    cellCentroids() const
+    {
+        return this->cellCentroids_(cartesianIndexMapper_.get());
+    }
 protected:
     void createGrids_()
     {
@@ -210,7 +221,8 @@ protected:
         grid_ = factory.convert(*equilGrid_, cartesianCellId_);
 
         cartesianIndexMapper_ =
-            new CartesianIndexMapper(*grid_, cartesianDimension_, cartesianCellId_);
+            std::make_unique<CartesianIndexMapper>(*grid_, cartesianDimension_,
+                                                   cartesianCellId_);
     }
 
     void filterConnections_()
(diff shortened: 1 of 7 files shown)


blattms:
jenkins build this please

bska:
On a tangential note: Is there a reason why we have to continue to store cell centres for all `Nx * Ny * Nz` Cartesian cells?

blattms:
> On a tangential note: Is there a reason why we have to continue to store cell centres for all Nx * Ny * Nz Cartesian cells?

I don't think that we do that outside of opm-common.

bska:
> > On a tangential note: Is there a reason why we have to continue to store cell centres for all Nx * Ny * Nz Cartesian cells?
> 
> I don't think that we do that outside of opm-common.

Maybe not, but [these expressions](https://github.com/OPM/opm-simulators/blob/6511d448d642230a186d89dee2dce00b0efeb2d7/ebos/eclbasevanguard.hh#L403-L404) certainly **seem** to index into an array that's sized according to the number of Cartesian cells.

blattms:
jenkins build this please

blattms:
would be cool of this was reviewed. Otherwise I might interpret the signs (e.g. @bska's comments) as approval and merge tonight.

totto82:
The code change looks good to me. 

> On a tangential note: Is there a reason why we have to continue to store cell centres for all Nx * Ny * Nz Cartesian c

`centroid = this->eclState().getInputGrid().getCellCenter(cartesianCellIdx);`

is only callable on rank==0 are has this changed?

blattms:
No that has not changed and it is guarded by `if(rank==0)` at [eclbasevanguard.hh#L401-L404](/OPM/opm-simulators/blob/457d270bdf6970dfe4e520fa4c72c38af756fb48/ebos/eclbasevanguard.hh#L401-L404)

bska:
>> ```c++
>> centroid = this->eclState().getInputGrid().getCellCenter(cartesianCellIdx);
>> ```
>>   
>> is only callable on rank==0 or has this changed?
> ```
>
> No that has not changed and it is guarded by `if(rank==0)`

But my question was really why do we use functions that require Cartesian indices at all?  Is there a reason we can't just base this lookup on the active cell index on all ranks?

akva2:
this is exactly what is done further down the pipe after grid has been distributed, see CellCentroidDataHandle. this is used to build that mapping.

blattms:
> But my question was really why do we use functions that require Cartesian indices at all

I guess the answer might:
Because either nobody found the time to do the full compression in the ECL layer, or nobody was brave enough to do that. There might be side effects... But if it works I am all for it. @joakim-hove might know why cell centers are excluded in particular.

joakim-hove:
> Because either nobody found the time to do the full compression in the ECL layer, or nobody was brave enough to do that.

The properties like `PERMX` and `SATNUM` are fully compressed in opm-common. The `EclipseGrid` stores the geometry internally with `COORD` and `ZCORN` - this representation is "before" the question of active and inactive cells is meaningful.

As for this particular example it is mainly a question of "overload space" - i.e. `EclipseGrid::getCellCenter( std::size_t )` has already taken the `std::size_t` overload, but if we define e.g. types `EclipseGrid::active_index` and `EclipseGrid::global_index` we can implement the `EclipseGrid::cetCellCenter( const EclipseGrid::active_index& index )` overload. I can do that.

PR #3566, opened 2021-09-30
1: Pack and unpack WellTestState in CollectIO class

joakim-hove:
Collect WellTestState on IO rank for restart output.

Upstream: https://github.com/OPM/opm-common/pull/2724

### DIFF
--- a/ebos/collecttoiorank.cc
+++ b/ebos/collecttoiorank.cc
@@ -576,6 +576,41 @@ public:
     }
 };
 
+class PackUnPackWellTestState : public P2PCommunicatorType::DataHandleInterface
+{
+public:
+    PackUnPackWellTestState(const WellTestState& localWTestState,
+                            WellTestState& globalWTestState,
+                            bool isIORank)
+        : local_(localWTestState)
+        , global_(globalWTestState)
+    {
+        if (isIORank) {
+            MessageBufferType buffer;
+            pack(0, buffer);
+
+            // pass a dummyLink to satisfy virtual class
+            int dummyLink = -1;
+            unpack(dummyLink, buffer);
+        }
+    }
+
+    void pack(int link, MessageBufferType& buffer) {
+        if (link != 0)
+            throw std::logic_error("link in method pack is not 0 as expected");
+        this->local_.pack(buffer);
+    }
+
+    void unpack(int, MessageBufferType& buffer) {
+        this->global_.unpack(buffer);
+    }
+
+private:
+    const WellTestState& local_;
+    WellTestState& global_;
+};
+
+
 class PackUnPackAquiferData : public P2PCommunicatorType::DataHandleInterface
 {
     const data::Aquifers& localAquiferData_;
@@ -842,7 +877,8 @@ collect(const data::Solution& localCellData,
         const std::map<std::size_t, double>& localWBPData,
         const data::Wells& localWellData,
         const data::GroupAndNetworkValues& localGroupAndNetworkData,
-        const data::Aquifers& localAquiferData)
+        const data::Aquifers& localAquiferData,
+        const WellTestState& localWellTestState)
 {
     globalCellData_ = {};
     globalBlockData_.clear();
@@ -850,6 +886,7 @@ collect(const data::Solution& localCellData,
     globalWellData_.clear();
     globalGroupAndNetworkData_.clear();
     globalAquiferData_.clear();
+    globalWellTestState_.clear();
 
     // index maps only have to be build when reordering is needed
     if(!needsReordering && !isParallel())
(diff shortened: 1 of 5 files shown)


joakim-hove:
OK - I need some typetag first aid here: Deep down the test `test_ecl_output` calls `CollectToIORank::collect()` - that function calls `WellManager<TypeTag>::wellTestState()`. The `test_ecl_output` is configured (by magic ...???) to use the ebos class `EclWellManager` - and there the `wellTestState()`methods look like:
```
WellTestState& wellTestState() {
        throw std::logic_error("wellTestState() method not implemented for class eclwellmanager");
}
```
and all in all the `test_ecl_output` test fails hard. I guess the good/correct solution would be to use the `BlackoilWellModel` instead of `EclWellManager` in the offending test?

bska:
> I need some typetag first aid here

I take it you've already run the test in a debugger locally with something along the lines of `catch throw` to identify the call site?

joakim-hove:
> I take it you've already run the test in a debugger locally with something along the lines of catch throw to identify the call site?

Ehhh - I had not thought of that; and I will try it - but my understanding of the problem is more how to control which wellmodel is instantiated. When it is run with the wrong wellmodel - i.e. `EclWellManager` in this case - the call to `CollectToIORank::collect()` in `eclwriter`*will fail*. Anyway - backtrace goes like this:
```
#0  0x00007ffff2aa5afd in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00005555558cd2bc in Opm::EclWellManager<Opm::Properties::TTag::TestEclOutputTypeTag>::wellTestState (this=0x555556c64e50)
    at /home/hove/work/OPM/opm-simulators/ebos/eclwellmanager.hh:650
#2  0x00005555558c0418 in Opm::EclWriter<Opm::Properties::TTag::TestEclOutputTypeTag>::evalSummaryState (this=0x555556bc0ef0, isSubStep=false)
    at /home/hove/work/OPM/opm-simulators/ebos/eclwriter.hh:185
#3  0x000055555587d26c in Summary::test_method (this=0x7fffffffc30f) at /home/hove/work/OPM/opm-simulators/tests/test_ecl_output.cc:179
#4  0x000055555587cd72 in Summary_invoker () at /home/hove/work/OPM/opm-simulators/tests/test_ecl_output.cc:151

```

bska:
> When it is run with the wrong wellmodel - i.e. `EclWellManager` in this case - the call to `CollectToIORank::collect()` in `eclwriter` _will fail_.

Okay&mdash;just one more question.  Do you need to meet a semantic requirement that `EclWellManager::wellTestState()` return a reference to an existing object or is it merely syntactical?  In other words, would defining the function as below meet the call-site requirements?
```C++
WellTestState EclWellManager::wellTestState()
{
    return {};
}
```

joakim-hove:
>  In other words, would defining the function as below meet the call-site requirements?

```
WellTestState EclWellManager::wellTestState()
{
    return {};
}
```
That would solve my immediate problem - and I might end up there - but it would introduce a silent failure in `EclWellManager`

joakim-hove:
jenkins build this opm-common=2724 please

joakim-hove:
jenkins build this opm-common=2724 please

joakim-hove:
jenkins build this opm-common=2724 please

joakim-hove:
I am working on restart support for the WTEST functionality - surprisingly it seems the greatest challenge has been to hook the `WellTestState` instance up to the IO framework. This PR is the first of two series with that target.

The second commit in this PR disables parts of the test `test_ecl_output`. To summarize the problem with the `test_ecl_output` test goes as follows:

1. With this PR a `WellTestState`member is added to the `CollectToIORank` class - and as part of the `::collect()` method the method `<WellModel>::wellTestState()` is called.
2. The method `<WellModel>::wellTestState()` is only implemented for `WellModel == BlackOilModel` which is the well model used by `flow` (and all other simulators?). For the model `EclWellModel` the method `wellTestSate()` is only a shim which throws an exception. 
3. The test in `test_ecl_output` instantiates a `EclWellModel` instance and during the test execution the shim which throws is invoked and the whole test fails.

I have made some half hearted attempts getting the test to instantiate and use `BlackoilWellModel` instead of `EclWellModel, but it fails with error messages from the language Ebos++ - quite frankly I am not all that interested in learning Ebos++ and would prefer to stick with C++. I am all for unit tests, but this particular test mainly tests our ability to create correct summary output - we have quite good test coverage on that already; both explicitly in opm-common and in the form of regression tests in opm-simulators.

I will let this linger for a not-very-long time and then I'll merge it.

blattms:
I am probably missing something, but can't you catch the exception and send a dummy pack a dummy test state in that case? And throw at the unpacking site to?

As the well manager type is known at compile time one could omit the methods for the ecl manger and check for their existence at compile time.

Seems like pure C++ to me, but I could be wrong of course.

joakim-hove:
> but can't you catch the exception and send a dummy pack a dummy test state in that case? 

Yes - that is correct; I can remove the exception entirely and return a dummy `WellTestState` from the `EclWellManager` class, but then the `EclWellmanager` returns invalid data - akin to a silent failure in my opinion. But for me that is of course the simplest way out of this - although I am reluctant to give the `EclWellManager` class "a scent of maintained" this way.

PR #3569, opened 2021-10-01
check operability injectors

totto82:
Address issue https://github.com/OPM/opm-simulators/issues/3555

Needs more testing on bigger cases. Will report back....


totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1300,13 +1292,21 @@ namespace Opm
             this->operability_status_.obey_bhp_limit_with_thp_limit = (*obtain_bhp >= bhp_limit);
 
             const double thp_limit = this->getTHPConstraint(summaryState);
-            if (*obtain_bhp < thp_limit) {
+            const double temp = *obtain_bhp < thp_limit;
+            if (this->isProducer() && temp < 0.) {
```
bska, line 1296: This looks very unexpected.  The variable `temp` is of type `double`, is initialised with an expression of type `bool`, and is then subsequently compared to a floating-point constant&mdash;the result of which is always `false` since Boolean values are never less than zero.  If this is in fact intended, I suggest you make an effort to think of a better name than `temp`.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -1047,8 +1041,8 @@ namespace Opm
 
                 const double thp = this->calculateThpFromBhp(well_state, well_rates_bhp_limit, bhp_limit, deferred_logger);
                 const double thp_limit = this->getTHPConstraint(summaryState);
-
-                if (thp < thp_limit) {
+                const double temp = thp < thp_limit;
+                if ( (this->isProducer() && temp < 0.) || (this->isInjector() && temp > 0.) ) {
```
bska, line 1045: This too is a very unexpected use of a floating-point variable.  Furthermore, this definition of `temp` shadows the [outer](https://github.com/OPM/opm-simulators/blob/8f2261b4a25f5f0b6475af6d70599e83b723c432/opm/simulators/wells/StandardWell_impl.hpp#L1027) object with the same name which further increases suspicion.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -1084,13 +1079,21 @@ namespace Opm
             this->operability_status_.obey_bhp_limit_with_thp_limit = (*obtain_bhp >= bhp_limit);
 
             const double thp_limit = this->getTHPConstraint(summaryState);
-            if (*obtain_bhp < thp_limit) {
+            const double temp = *obtain_bhp < thp_limit;
+            if (this->isProducer() && temp < 0.) {
```
bska, line 1083: Please clear this up.  We can't have these kinds of comparisons.

bska (changes requested):
I looked at the [new warnings](https://ci.opm-project.org/job/opm-simulators-PR-builder/3097/gcc/fileName.-1904334838/source.7ceabe81-cff1-40a6-8c06-ca66c25de8ec/#1160) introduced by this PR.  The "shadows" issue pointed to, I think, a deeper problem and I'd like to clear up the logic before we consider merging this.

Code like
```C++
const double temp = a < b;
if (temp < 0.0) {
    do_something();
}
```
is very confusing since `temp < 0.0` will never be true if `temp` is initialised with a boolean expression.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: Obs, this was not my intention. I will clean it up. Should have doubled checked before converting it from draft.....

bska: > Obs, this was not my intention. I will clean it up.

Cool.  I look forward to seeing the revised implementation.

> Should have doubled checked before converting it from draft...

Just in case you don't know, it's possible to go back to draft state by clicking the "convert to draft" link in the upper right-hand corner of this page.

---

convert-to-draft

bska (approved):
Thanks a lot for the updates.  This is much more readable than the earlier version of the patch.

totto82:
jenkins build this please

bska:
@totto82 : Do you expect any further changes here?  If not, please merge this into master.

PR #3572, opened 2021-10-04
Prune Analytic Aquifer Connections to Deactivated Cells

bska:
This PR broadcasts the `removed_cells` to all ranks and uses that array to delete any analytic aquifer connections that might have been previously formed to those deactivated cells.  This is needed to have a consistent view of which cells are connected to analytic aquifers.

We use the new&mdash;hopefully temporary&mdash;interface `EclipseState::pruneDeactivatedCells()` introduced in OPM/opm-common#2730 to affect the actual removal of analytic aquifer connections to deactivated cells.


bska:
jenkins build this opm-common=2730 opm-grid=551 please

ebos/eclgenericcpgridvanguard.cc
```diff
@@ -291,6 +298,19 @@ void EclGenericCpGridVanguard<ElementMapper,GridView,Scalar>::doCreateGrids_(Ecl
         auto &field_props = eclState.fieldProps();
         const_cast<FieldPropsManager&>(field_props).reset_actnum(actnum);
     }
+
+#if HAVE_MPI
+    {
+        EclMpiSerializer ser(Dune::MPIHelper::getCollectiveCommunication());
+        ser.broadcast(removed_cells);
```
blattms, line 305: Oooh, this serializer concept is spreading everywhere. I always assumed that this would only be used for EclipseState.

Wouldn't it be easier to just do `grid_.comm().broadcast(removed_cells.data(), removed_cells.size(), 0)`? Then you could also kill the pack/unpack code.

Additionally the PR removing MPI_COMM_WORLD would work more nicely with this.

atgeirr: > Wouldn't it be easier to just do `grid_.comm().broadcast(removed_cells.data(), removed_cells.size(), 0)`? Then you could also kill the pack/unpack code.

> Additionally the PR removing MPI_COMM_WORLD would work more nicely with this.

I strongly support this. Any new use of MPI_COMM_WORLD (directly or through using a default collective communication object) must be replaced anyway.

bska: > Wouldn't it be easier to just do
> ```C++
> grid_.comm().broadcast(removed_cells.data(), removed_cells.size(), 0)
> ```
> ?

Maybe.  Do I need to broadcast the size in a separate call first or does that happen implicitly.

> Then you could also kill the pack/unpack code.

That's an obvious benefit.

On a separate note, why do we have three different frameworks for communicating between ranks?

bska: I've pushed an update that switches to using the `comm().broadcast()` method.

joakim-hove:
The PR would be simpler to read if the reformatting was relegated to a separate commit.

bska:
jenkins build this please

PR #3573, opened 2021-10-05
remove assert and handle opposite rates in wecon related fractions. 

totto82:
First PR to remove wrong usage of asserts.

### DIFF
--- a/opm/simulators/wells/WellInterfaceFluidSystem.cpp
+++ b/opm/simulators/wells/WellInterfaceFluidSystem.cpp
@@ -508,19 +508,18 @@ checkMaxWaterCutLimit(const WellEconProductionLimits& econ_production_limits,
     // function to calculate water cut based on rates
     auto waterCut = [](const std::vector<double>& rates,
                        const PhaseUsage& pu) {
-
-        const double oil_rate = rates[pu.phase_pos[Oil]];
-        const double water_rate = rates[pu.phase_pos[Water]];
-
-        // both rate should be in the same direction
-        assert(oil_rate * water_rate >= 0.);
-
+        const double oil_rate = -rates[pu.phase_pos[Oil]];
+        const double water_rate = -rates[pu.phase_pos[Water]];
         const double liquid_rate = oil_rate + water_rate;
-        if (liquid_rate != 0.) {
-            return (water_rate / liquid_rate);
-        } else {
+        if (liquid_rate == 0.)
             return 0.;
-        }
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

totto82:
jenkins build this please

GitPaean:
jenkins build this please

opm/simulators/wells/WellInterfaceFluidSystem.cpp
```diff
@@ -508,19 +508,18 @@ checkMaxWaterCutLimit(const WellEconProductionLimits& econ_production_limits,
...
+        const double oil_rate = -rates[pu.phase_pos[Oil]];
+        const double water_rate = -rates[pu.phase_pos[Water]];
         const double liquid_rate = oil_rate + water_rate;
-        if (liquid_rate != 0.) {
-            return (water_rate / liquid_rate);
-        } else {
+        if (liquid_rate == 0.)
             return 0.;
-        }
+        else if (water_rate < 0)
```
GitPaean, line 516: A question here, the changes here implies that we have encountered different signs for oil_rate and water_rate, is that the case?

totto82: Yes, this assert triggers on a proprietary case. We may have situations where producers temporary injects epsilon during the iterations process.

opm/simulators/wells/WellInterfaceFluidSystem.cpp
```diff
@@ -594,25 +581,14 @@ checkMaxWGRLimit(const WellEconProductionLimits& econ_production_limits,
...
-            }
-        }
-
-        return water_gas_ratio;
+        const double water_rate = -rates[pu.phase_pos[Water]];
+        const double gas_rate = -rates[pu.phase_pos[Gas]];
+        if (water_rate <= 0.)
+            return 0.;
+        else if (gas_rate <= 0.)
+            return 1.e100; // big value to mark it as violated
```
GitPaean, line 589: Do we do anything regarding to these big values? Should we just throw here?

totto82: I don't think we should throw. We should handle cases with 0 or negative gas rate by closing the well. The return of the large values here is just to make sure we close the wells due to economic  reason.

GitPaean:
Thanks for the contribution. It is a simple PR serving its purpose. It will go in after Jenkins passes.

PR #3574, opened 2021-10-05
remove assert for isfinite

totto82:
Remove assert for infinite primary variables update.


totto82:
jenkins build this please

bska:
Why?  Do we expect that updates will be `NaN` or `Inf`?

totto82:
It assumed it may happen if the D matrix is singular, but while hunting for other asserts that are triggered for full field models, I noticed a   `/ newton_update` in relaxationFactorRate that looks very suspicious. I will try to dig deeper into this.

atgeirr:
> Why? Do we expect that updates will be `NaN` or `Inf`?

Yes, at least Inf. User input can create a physically impossible case, and it is not the fault of the programmer that the variables can become infinite. (Or at least, it is not the fault of this method or its immediate callers.) So the assert() is wrong. Could consider replacing with a throw, but I am ok with removal (there are inf-checks in the convergence checking code).

blattms:
The big question is, will we recover from an inf (without treating convergence issues as successful convergence)?

On a more general note:
I think most of these asserts are implemented under the assumption that a default productive binary is compiled with -DNDEBUG (which many, including Unix distributions thinks is bad). If we start removing these assert now then we should at least have a way to activate them for debugging, e.g. via a common `-DOPM_DEBUG` that developers can use turn on expensive or otherwise unpleasing checks. 

Just removing them might backfire/increase work later.

atgeirr:
> If we start removing these assert now then we should at least have a way to activate them for debugging, e.g. via a common `-DOPM_DEBUG` that developers can use turn on expensive or otherwise unpleasing checks.

Replacing the use of NDEBUG here (which is redunant since the assert()s go away with it anyway) with `#ifdef OPM_DEBUG` would also be fine with me. Does not really change the issue that assert()ing is not quite right (user input can cause the assert() to trigger), but buries it well enough.

totto82:
jenkins build this please

totto82:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -868,19 +868,19 @@ namespace Opm
...
         const double dBHPLimit = this->param_.dbhp_max_rel_;
         this->StdWellEval::updatePrimaryVariablesNewton(dwells, dFLimit, dBHPLimit);
 
         updateExtraPrimaryVariables(dwells);
 
-#ifndef NDEBUG
         for (double v : this->primary_variables_) {
-            assert(isfinite(v));
+            if(!isfinite(v))
+                OPM_DEFLOG_THROW(NumericalIssue, "Infinite primary variable after newton update well: " << this->name(),  deferred_logger);
```
atgeirr, line 882: Just to be sure this is what you actually want: you will now (as with the existing assert()) always, unconditionally, react to the condition, but the reaction will not be terminating the program but throwing an exception. (Which will be picked up higher up in the code, communicated properly,  and result in a timestep cut.)

It sounds good to me, so if I understood you correctly, this can be merged. (Although I would appreciate adding spaces after the `if`s and braces around the bodies of the `if`s. We write macros that are safe for this, but not all do.)

totto82:
This PR deadlock for `NORNE_ATW2013_2B_MSW`. Needs further investigations.

atgeirr:
> This PR deadlock for `NORNE_ATW2013_2B_MSW`. Needs further investigations.

Check that there is communication of throw status higher up in the call stack (in BlackoilWellModel), I expected there would be one since we have a DeferredLogger in the scope already, but perhaps that was wrong?

totto82:
In MSWellHelpers there is a `OPM_THROW_NOLOG(NumericalIssue, msg) ` which is now locally cached in iterateWellEquation. Either we need to pass the deferred_logger to applyUMFPack or I need to handle the catch differently. Any thoughts?

bska:
> now locally cached in `iterateWellEquation`.

Did you mean "cached" or "caught" here?

atgeirr:
> Any thoughts?

You must either: a) let the (or at least some) exception propagate out of the function, so it can be caught in the BlackoilWellModel method from which you are called and communicated,  b) avoid throwing and continue, or c) replace exception usage with returning a status flag. If b) is not possible (i.e. there is no reasonable way to continue) then either a) or c) must be done. If you do c), such a flag must be communicated, and you must remember to replace all throwing in that part of the code with such an error return. In general I find exceptions less intrusive than error flags.

Note: the DeferredLogger itself has nothing to do with the deadlock, it is just a mechanism to ensure that messages from a single process reach rank 0 eventually.

PR #3576, opened 2021-10-05
Regression added test BASE model6

jilocode:
Verified that flow simulation results are correct for the case.

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -962,6 +962,14 @@ add_test_compareECLFiles(CASENAME actionx_wefac
                          REL_TOL ${rel_tol}
                          DIR actionx)
 
+add_test_compareECLFiles(CASENAME 0-BASE-MODEL6
+                         FILENAME 0_BASE_MODEL6
+                         SIMULATOR flow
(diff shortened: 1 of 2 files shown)


bska:
Welcome to the project!

Thank you very much for doing this verification work.  I have launched a regression test run just to make sure that we don't accidentally break any existing tests.  When that's complete I'll generate reference solutions for the new test case and install this as part of our continuous integration workflow.

Thank you again for this very valuable contribution.

bska:
Would you be able to squash the two commits down to one?  If not that's fine, but if you are that would be perfect.

bska:
> > Would you be able to squash the two commits down to one?
> 
> We could make more active use of the "Squash and merge" alternative on the big green button.

This is true, but doing so will create local merge conflicts for @jilocode unless they're fully onboard with rebasing.  That's why I asked.

bska:
Thank you for the updates, @jilocode.  Do you consider this work ready for merging or are there additional changes incoming?

jilocode:
Ready to merge

bska:
jenkins build this please

bska:
jenkins build this please

PR #3578, opened 2021-10-05
Reset ACTNUM at EclipseState Level

bska:
This method avoids resorting to const_cast tricks.


bska:
Note: This PR depends on and contains #3572 and is therefore created in draft mode.  I will mark it as "ready for review" once the upstream PR has been merged.  We also use the new interface `EclipseState::reset_actnum` introduced in OPM/opm-common#2734 and therefore also depend on that PR.

blattms (approved):

bska:
jenkins build this please

PR #3579, opened 2021-10-06
WTEST: only count actually closed wells.

joakim-hove:
See upstream PR - there is a real change in semantics here. 

Upstream: https://github.com/OPM/opm-common/pull/2736


blattms:
If I read something like closedWells, I would rather expect to get identifiers of closed wells than the number of closed wells...

PR #3583, opened 2021-10-07
brine+thermal

totto82:
Allow for combining of brine and thermal using a dedicated exactable: `flow_brine_thermal`
A test case will be added soon.


totto82:
jenkins build this please

PR #3585, opened 2021-10-07
Main.hpp: Split Individual Cases Out to Helper Functions

bska:
This PR moves the bodies of the various `dynamicDispatch_<>()` cases out to separate helper functions.  Not only does this reduce the number of nested conditionals, it also helps reasoning about each case in isolation and aids future maintenance.

While here, switch to using `std::find_if()` to determine if the user passed `--version` and ensure that we honour the Posix standard that `argv[argc]` exist and be a nullpointer.

### DIFF
--- a/opm/simulators/flow/Main.hpp
+++ b/opm/simulators/flow/Main.hpp
@@ -185,12 +185,18 @@ namespace Opm
 
         void setArgvArgc_(const std::string& filename)
         {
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska:
jenkins build this please

joakim-hove:
A little clearer - good!

PR #3586, opened 2021-10-07
[refactor] Use well_container and PerforationData for tracers

blattms:
This saves some (expensive?) lookup (already done for us in the well model). We had to make the well_container accessible from the well model for this.

Using the perforation data will automatically make sure that the perforation are not shut and on this process in a parallel run.

The recent refactoring of the well code is quite handy :+1:

### DIFF
--- a/ebos/ecltracermodel.hh
+++ b/ebos/ecltracermodel.hh
@@ -291,38 +291,21 @@ protected:
         }
 
         // Wells  terms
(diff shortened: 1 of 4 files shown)


blattms:
jenkins build this please

blattms:
Will do later.

blattms:
Since this is approved I will shamelessly selfmerge now.

Issue #3587, opened 2021-10-08
Clean-up output of well switching


totto82:
At the moment we have this
```
Starting time step 0, stepsize 1 days, at day 2947/2948, date = 01-Dec-2005
    Switching control mode for well C-1H from RATE to GRUP
    Switching control mode for well C-2H from RATE to GRUP
    Switching control mode for well C-3H from RATE to GRUP
    Switching control mode for well F-1H from RATE to GRUP
    Switching control mode for well F-2H from RATE to GRUP
    Switching control mode for well F-3H from RATE to GRUP
    Switching control mode for well C-4AH from RATE to GRUP
    Switching control mode for well C-2H from GRUP to RATE
    Switching control mode for well C-3H from GRUP to RATE
    Switching control mode for well E-3CH from RESV to BHP
    Switching control mode for well C-1H from GRUP to BHP
    Switching control mode for well F-3H from GRUP to RATE
    Switching control mode for well C-1H from BHP to RATE
    Switching control mode for well C-1H from RATE to BHP
    Switching control mode for well C-4AH from GRUP to RATE
```
Is it possible to change to something like this instead:
```
Starting time step 0, stepsize 1 days, at day 2947/2948, date = 01-Dec-2005
    Well C-1H    control mode changed from RATE->GRUP->BHP ->RATE-> BHP
    Well C-2H    control mode changed from RATE->GRUP->RATE
    Well C-3H    control mode changed from RATE->GRUP->RATE  
    Well C-4AH   control mode changed from RATE->GRUP->RATE   
    Well E-3CH   control mode changed from RESV->BHP
    Well F-1H    control mode changed from RATE->GRUP 
    Well F-2H    control mode changed from RATE->GRUP
    Well F-3H    control mode changed from RATE->GRUP->RATE  
```
or something similar. So that the wells are in alphabetic order and all the changes for a well are on the same line. 

Just a suggestion to improve readability.

_Originally posted by @OPMUSER in https://github.com/OPM/opm-simulators/issues/3160#issuecomment-915171521_

PR #3588, opened 2021-10-08
make multiple well control switching output more readable

totto82:
See discussion in #3587

Output from report step 30 in SPE9 

```
    Well PRODU4 control mode changed from GRUP->ORAT
    Well PRODU5 control mode changed from GRUP->ORAT
    Well PRODU7 control mode changed from GRUP->ORAT->GRUP
    Well PRODU8 control mode changed from GRUP->ORAT
    Well PRODU10 control mode changed from GRUP->ORAT
    Well PRODU14 control mode changed from GRUP->ORAT
    Well PRODU15 control mode changed from GRUP->ORAT
    Well PRODU17 control mode changed from GRUP->ORAT
    Well PRODU18 control mode changed from GRUP->ORAT
    Well PRODU19 control mode changed from GRUP->ORAT
    Well PRODU24 control mode changed from GRUP->ORAT
```

### DIFF
--- a/opm/simulators/wells/WellInterfaceGeneric.cpp
+++ b/opm/simulators/wells/WellInterfaceGeneric.cpp
@@ -96,6 +96,8 @@ WellInterfaceGeneric::WellInterfaceGeneric(const Well& well,
     }
 
     wsolvent_ = 0.0;
+
+    well_control_log_.clear();
 }
 
 const std::string& WellInterfaceGeneric::name() const
@@ -376,5 +378,26 @@ double WellInterfaceGeneric::getALQ(const WellState& well_state) const
 {
     return well_state.getALQ(name());
 }
+
+void WellInterfaceGeneric::reportWellSwitching(const SingleWellState& ws, DeferredLogger& deferred_logger) const
+{
+    if (well_control_log_.empty())
+        return;
+
+    std::string msg = "    Well " + name()
+        + " control mode changed from ";
+    for (const std::string& from : well_control_log_) {
+        msg += from + "->";
+    }
+    std::string to;
+    if (isInjector()) {
+        to = Well::InjectorCMode2String(ws.injection_cmode);
+    } else {
+        to = Well::ProducerCMode2String(ws.production_cmode);
+    }
+    msg += to;
+    deferred_logger.info(msg);
+}
+
 
 } // namespace Opm
--- a/opm/simulators/wells/WellInterfaceGeneric.hpp
+++ b/opm/simulators/wells/WellInterfaceGeneric.hpp
@@ -169,6 +169,8 @@ public:
     // whether a well is specified with a non-zero and valid VFP table number
     bool isVFPActive(DeferredLogger& deferred_logger) const;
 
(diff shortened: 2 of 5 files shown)


totto82:
jenkins build this please

opm/simulators/wells/WellInterfaceGeneric.cpp
```diff
@@ -377,4 +379,26 @@ double WellInterfaceGeneric::getALQ(const WellState& well_state) const
     return well_state.getALQ(name());
 }
 
+void WellInterfaceGeneric::reportWellSwitching(const WellState& well_state, DeferredLogger& deferred_logger) const
```
joakim-hove, line 382: This could just take `const SingleWellState&` as argument.

opm/simulators/wells/WellInterfaceGeneric.hpp
```diff
@@ -301,6 +303,8 @@ class WellInterfaceGeneric {
     double well_efficiency_factor_;
     const VFPProperties* vfp_properties_;
     const GuideRate* guide_rate_;
+
+    std::vector< std::string> well_control_log_;
```
joakim-hove, line 307: Maybe use the control enum instead of a string here?

atgeirr:
Had not seen this PR when I made my comment in #3587, but I think it should be easy to eat our cake and have it too here?

opm/simulators/wells/WellInterfaceGeneric.hpp
bska: > Maybe use the control enum instead of a string here?

Agreed.  Instead of spending a lot of effort formatting strings we should collect the individual transition states and only format when creating the final log statements.

totto82: Then we will need a `well_control_log_inj_` and a `well_control_log_prod_ ` but that's fine with me.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -212,15 +209,7 @@ namespace Opm
...
             }
-            well_control_log_.push_back(from);
-            std::ostringstream ss;
-            ss << "    Switching control mode for well " << this->name()
-               << " from " << from
-               << " to " <<  to;
-            if (cc.size() > 1) {
-               ss << " on rank " << cc.rank();
-            }
-            deferred_logger.info(ss.str());
```
atgeirr, line 223: Please keep this, but using debug(), not info(), then we still get the appropriate switch messages on each iteration in the debug file (only).

opm/simulators/wells/WellInterfaceGeneric.hpp
totto82: I will argue for keeping it std::string 1) We now need the std::string format and thus the conversion for the .debug here anyway. 2) The `reportWellSwitching` becomes more complicated if we use enum since we need to distinguish between injectors and producers.

atgeirr: I think these are good arguments, ok with keeping the string.

PR #3592, opened 2021-10-08
remove assert for isfinite and correctly handle exceptions

blattms:
Remove assert for infinite primary variables update and handle exceptions that might be thrown in the linearize method of the well model correctly.

Superseeds #3574 and combines it with totto82/opm-simulators#11. Rebased on current master-


totto82:
I can confirm that this fixes the issue on `NORNE_2B_MSW`

PR #3593, opened 2021-10-10
Mark 'comm' As Potentially Unused

bska:
The communicator is only needed in parallel builds.


bska:

bska:
Self merging to fix a warning in sequential builds.

PR #3595, opened 2021-10-11
Add serialization testing for WellTestState

joakim-hove:
Upstream: https://github.com/OPM/opm-common/pull/2748


joakim-hove:
jenkins buld this opm-simulators=2748 please

PR #3601, opened 2021-10-12
Restart sched flag

joakim-hove:
Add test flag `RESTART_SCHED`when registering a restart test. The purpose of this switch is to be able to pass `--sched-restart=true` or `--sched-restart=false` to the restarted run. If the `RESTART_SCHED`argument is not supplied flow will restart with whatever is default behavior.

### DIFF
--- a/tests/run-regressionTest.sh
+++ b/tests/run-regressionTest.sh
@@ -12,8 +12,9 @@ REL_TOL="$6"
 COMPARE_ECL_COMMAND="$7"
 RST_DECK_COMMAND="$8"
 RESTART_STEP="${9}"
-EXE_NAME="${10}"
-shift 10
+RESTART_SCHED="${10}"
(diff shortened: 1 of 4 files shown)


tests/run-regressionTest.sh
```diff
@@ -12,8 +12,9 @@ REL_TOL="$6"
 COMPARE_ECL_COMMAND="$7"
 RST_DECK_COMMAND="$8"
 RESTART_STEP="${9}"
-EXE_NAME="${10}"
-shift 10
+RESTART_SCHED="${10}"
+EXE_NAME="${11}"
+shift 11
```
akva2, line 17: this is getting unmanageable. not caused by this addition alone though, but it is time to introduce getopt usage. i will handle that later. that being said there are other drivers which needs updating too, notably the porv and init ones.

PR #3607, opened 2021-10-14
Check operabiliy for wells under history

totto82:
Builds on #3569


bska (commented):
The code looks good to me, but I have a general question.  Suppose a well has non-zero observed rates and that these rates are entered in `WCONHIST` or `WCONINJH` as appropriate.  Why would we ever consider shutting that well?

totto82:
> The code looks good to me, but I have a general question. Suppose a well has non-zero observed rates and that these rates are entered in WCONHIST or WCONINJH as appropriate. Why would we ever consider shutting that well?

We may have rationalizations that are far from the truth that causes convergence issues and inoperable wells. The current behavior would be to throw and give up. This will help the simulator keep on simulating. This was asked for by @tskille and he is testing it. It shouldn't be merged before he approves it.

joakim-hove:
> Suppose a well has non-zero observed rates and that these rates are entered in WCONHIST or WCONINJH as appropriate. Why would we ever consider shutting that well?

Because the reservoir description is (always) wrong.

totto82:
rationalizations = realizations. (spell checking going wrong...)

totto82:
The testing I and @tskille has done is positive and the code change itself is already approved. -> I will merge.

PR #3611, opened 2021-10-15
Broadcast wtest state

joakim-hove:
Downstream: https://github.com/OPM/opm-common/pull/2758


tests/test_wellstate.cpp
```diff
@@ -440,6 +440,10 @@ BOOST_AUTO_TEST_CASE(TESTWellContainer) {
     BOOST_CHECK_EQUAL(wc.well_name(1), "W2");
     BOOST_CHECK_THROW(wc.well_name(10), std::exception);
 
+    const auto& wells = wc.wells();
+    BOOST_CHECK_EQUAL(wells.size() , wc.size());
+    BOOST_CHECK( (wells[0] == "W1" && wells[1] == "W2") ||
+                 (wells[1] == "W1" && wells[0] == "W2") );
```
bska, line 446: It's probably overkill for two-element sequences,  but in general you can use [`std::is_permutation()`](https://en.cppreference.com/w/cpp/algorithm/is_permutation)&mdash;preferably in the dual-range form&mdash;to answer these kinds of questions.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -279,7 +279,7 @@ initFromRestartFile(const RestartValue& restartValues,
     }
 
 
-    this->active_wgstate_.well_test_state = std::move(wtestState);
+    this->active_wgstate_.wtest_state(wtestState);
```
bska, line 282: Is there no benefit to using `std::move()` here?  In other words, do you not get any improvements if you say
```C++
this->active_wgstate_.wtest_state(std::move(wtestState));
```
in this context?

bska (commented):
Looks good to me for the most part.  I just have one question concerning whether or not to use `std::move()` on a `WTestState` object.

tests/test_wellstate.cpp
joakim-hove: Well overkill or not - `std::is_permutation()` was a good tool to learn about.

tests/test_wellstate.cpp
```diff
@@ -440,6 +441,9 @@ BOOST_AUTO_TEST_CASE(TESTWellContainer) {
     BOOST_CHECK_EQUAL(wc.well_name(1), "W2");
     BOOST_CHECK_THROW(wc.well_name(10), std::exception);
 
+    const auto& wells = wc.wells();
+    std::vector<std::string> expected = {"W1", "W2"};
+    BOOST_CHECK( std::is_permutation( wc.begin(), wc.end(), expected.begin(), expected.end());
```
bska, line 446: Looks like there's a right parenthesis missing in the call to `BOOST_CHECK` here.

bska (approved):
Thanks for the updates.  This looks good to me now.

PR #3612, opened 2021-10-15
Report Group Level Injection Guide Rates if Present

bska:
We used to tie reporting these quantities to whether or not a group had any guide rates for production ([`GuideRate::has()`](https://github.com/OPM/opm-common/blob/3b2d650f8d68d87cd41449bf796d0dfd7118cd52/src/opm/parser/eclipse/EclipseState/Schedule/Group/GuideRate.cpp#L193-L196)), but this is clearly insufficient for group-level injection guide rates.


atgeirr (approved):

bska:
jenkins build this please

PR #3613, opened 2021-10-15
Switch Guide Rate Accumulation to Post-Order Traversal

bska:
This guarantees, under the assumption that the group tree does not have cycles, that we do not accumulate group-level guide rate values until all of its children are fully evaluated.  We use an iterative depth-first post-order tree traversal with an explicit stack instead of a recursive implementation.

The previous implementation, which tried to do the same kind of child-to-parent accumulation, might visit a parent group multiple times which in turn might lead to losing updates.  This is a more formalised approach to the value accumulation than was originally employed.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -44,8 +44,10 @@
 
 #include <algorithm>
 #include <cassert>
+#include <stack>
 #include <stdexcept>
 #include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
 #include <fmt/format.h>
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska:
The [regression failures](https://ci.opm-project.org/job/opm-simulators-PR-builder/3192/testReport/) are all of the expected kind&mdash;new non-zero items added to `XGRP` as a result of a more accurate accumulation procedure.  I'm going to update the reference solutions of the two failing regression tests and merge per prior approval.

bska:

bska:
jenkins build this opm-tests=651 please

bska:
New reference solutions have been installed on the CI system.  Merging this now to bring the overall system back to a consistent state.

PR #3616, opened 2021-10-18
build flow_brine_energy as an object library

akva2:
improves parallel builds


akva2:
jenkins build this please

PR #3619, opened 2021-10-19
Opencl refactor

Tongdongq:
This PR refactors some opencl code.
The ChowPatel decomposition on CPU is moved to `ChowPatelIlu.cpp`, where the GPU version already was.
The opencl kernels were created by `openclSolver`, which would pass them to `BILU0`. This could become more complex when a CPR preconditioner is added.
Now the static `OpenclKernels` class compiles/creates them, and allows any other class to use them.
`openclSolver` still triggers the initialization of the opencl kernel creation.
Another option would be to have each kernel string and compilation in the class where it is used, however this could lead to duplication if multiple classes use the same kernel.

There is a problem with convergence for the openclSolver when the wellcontributions are not added to the matrix, which also occurs in the current master.

```
Benchmarks for NORNE_ATW2013, wellcontributions added to matrix, R5 2400G, NVIDIA GTX1050Ti:
Runtime omitted for CHOW_PATEL benchmarks.
Timed for masters of 2021-10-13.
Notation: A (B+C+D+E), F, G, I | J
A: Total time
B: Assembly time
C: Linear solve time
D: Update time
E: Pre/post step
F: Overall Linearizations
G: Overall New Iterations
I: Overall Linear Iterations
J: Number of times openclSolver failed to converge and Dune was used instead.
CHOW_PATEL on CPU:
LS: 1871, 1510, 54661 | 20
GC: 2351, 1960, 83141 | 12
CHOW_PATEL on GPU:
LS: 1854, 1491, 57390 | 15
GC: 2302, 1910, 44647 | 24
openclSolver this PR:
LS: 621 (127+308+68+100), 1806, 1444, 20648 | 0
GC: 710 (158+340+87+107), 2285, 1895, 43638 | 0
openclSolver previous:
LS: 786 (129+472+68+ 99), 1806, 1444, 20648 | 0
GC: 754 (162+381+88+105), 2246, 1862, 43091 | 0
```


Tongdongq:
I rebased this with the `cl2.hpp` header PR.

blattms:
Compiles on my machine. Will run a small test.

Both vexcl and amgcl use the old headers that are not available on newer system. I should make a Cmake check for that and
deactivate until it it is fixed upstream.

blattms:
jenkins build this please

blattms:
It has been a long time since I have run Norne with OpenCL. Running it with this branch seems to take a very long time due to small timesteps. At least with the default options.

Tongdongq:
> It has been a long time since I have run Norne with OpenCL. Running it with this branch seems to take a very long time due to small timesteps. At least with the default options.

Did you add the wellcontributions? There is a problem with openclSolver without them: https://github.com/OPM/opm-simulators/issues/3635

blattms:
No I have not.

blattms:
Yep, with that is works much better

blattms:
@Tongdongq The problem with vexcl should be fixed with upstream's latest master.

PR #3620, opened 2021-10-19
[refactor] Remove cartesianToCompressed Mapping from well model

blattms:
Rebased version of #3550 with fixes. 

Main fix is that we add and use a new method compressedIndexForInterior that return -1 for non-interior cells. For distributed wells this makes sure that each perforation is only present/computed on one MPI rank.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -296,6 +296,8 @@ public:
     /*!
      * \brief Return compressed index from cartesian index
      *
+     * \return compressed index of cell is in interior, -1 otherwise
+     *
      */
     int compressedIndex(int cartesianCellIdx) const
     {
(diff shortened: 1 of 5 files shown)


ebos/eclbasevanguard.hh
```diff
@@ -424,9 +445,22 @@ protected:
     void updateCartesianToCompressedMapping_()
     {
         size_t num_cells = asImp_().grid().leafGridView().size(0);
-        for (unsigned i = 0; i < num_cells; ++i) {
-            unsigned cartesianCellIdx = cartesianIndex(i);
-            cartesianToCompressed_[cartesianCellIdx] = i;
+        is_interior_.resize(num_cells);
```
bska, line 448: Just as an apropos/tangential remark, this `num_cells` is the total number of (active) cells in the model, right?  Not just the number of active cells on the current rank?

PR #3632, opened 2021-10-25
Separate MICP Case Out to Helper Function

bska:
In keeping with the other simulation cases.


bska:
jenkins build this please

bska:
Merging per prior approval.

PR #3633, opened 2021-10-25
Report Well Level Injection Guide Rates if Group Controlled

bska:
This PR introduces a new helper class `GroupTreeWalker`, currently private to the `BlackWellModelGeneric.cpp` implementation file, that is based on the code in #3613 for accumulating group-level injection and production guide rate values.  The user can add visitor operations for wells and groups, typically with side effects, and then choose whether to run a preorder walk (visit groups before their children) or a post-order walk (visit children&mdash;especially wells&mdash;before their parents).

We use this helper class to ensure that we always extract and report pertinent injection guide rates at the well level (i.e., `WxIGR`) if at least one of the groups in the well's upline to the FIELD group has an entry for the corresponding phase in `GCONINJE`.  This is for increased compatibility with ECLIPSE.

Prior to this change we would report zero-valued `WWIGR` vectors on a real field case which made analysing simulation results needlessly difficult.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -44,8 +44,11 @@
 
 #include <algorithm>
 #include <cassert>
+#include <functional>
 #include <stack>
 #include <stdexcept>
+#include <string_view>
+#include <tuple>
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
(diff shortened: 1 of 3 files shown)


bska:
The prerequisite PR #3613 has been merged.  Marking this as "ready for review" now.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
I have now verified that the regression failures in the previous [build check](https://ci.opm-project.org/job/opm-simulators-PR-builder/3197/testReport/) are as expected.  In other words, it is safe to update the reference solutions once this PR is reviewed, updated, and finally fully approved.  With additional assistance from @jalvestad we also identified a similar problem with the production guide rates and I have included a fix for that problem in the third commit here.

With these fixes in place we get non-trivial well-level injection and production guide rates (i.e., summary vectors `Wx{I,P}GR`) when we're supposed get non-trivial vectors.  As a demonstration, I've run [model2/9_3A_GINJ_REIN-G_STW.DATA](https://github.com/OPM/opm-tests/blob/c4ebb5bb2423f99a31ae47911fdf180bd0c058b2/model2/9_3A_GINJ_REIN-G_STW.DATA) from `opm-tests` and plotted a selection of the guide rate vectors below.

---
 
Vector | Comparison Plot
--- | ----
Gas Injection | WGIGR-INJ1
Gas Injection | WGIGR-INJ2
Gas Production | WGPGR-PROD1
Oil Production | WOPGR-PROD2
Water Production | WWPGR-PROD3

bska:
jenkins  build this please

bska:
jenkins build this please

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -82,6 +85,238 @@ namespace {
 
         return { 0.0, restart.get(item), target };
     }
+
+    struct RetrieveWellGuideRate
```
joakim-hove, line 89: Looks like this functionality could be tested without bringing in the full vanguard?

joakim-hove: Looks like the general group control code could benefit from something like this?

joakim-hove: > We do need a mostly complete Schedule object with a non-trivial group tree though

I think this should be brought out with a header file++ and then tested; could use e.g. `MOD_GRP.DATA` as test data?

bska: > > We do need a mostly complete Schedule object with a non-trivial group tree though
> 
> I think this should be brought out with a header file++ and then tested; could use e.g. `MOD_GRP.DATA` as test data?

On the other hand, there's just a single use-case at the moment and I'd rather not generalise from that single example.  For instance, we don't know how `GroupTreeWalker` will hold up if deployed to more general use-cases, especially ones that may involve communicating partial results between MPI ranks during group tree traversal.

In other words, I'd prefer to keep this completely hidden from other parts of the code until we can explore the problem more broadly.

joakim-hove: I beg to differ here:

1. We don't have any API stability in this project - at all; i.e. there is no real harm in exposing functionality which will probably change in the future. On the contrary it would be valuable to have *current* behavior and expectations codified in a test when it is at a later staged extended in new directions.
2. opm-simulators is notoriously difficult to test because many things are strongly coupled. This is an example of functionality which is not strongly coupled to other parts of the code (very good!) and therefor testing is relatively simpler, I really think we should use the opportunities to add tests to opm-simulators when the opportunity arises.

I have made my point; I'll back off now.

atgeirr: > In other words, I'd prefer to keep this completely hidden from other parts of the code until we can explore the problem more broadly.

I disagree with this, refactoring to move the code with generalizing potential out of its hiding place in a cpp file is good! Otherwise it becomes very unlikely that it will never be improved or reused, since any interested such user first must extract it and make sure that was ok. It is also not necessarily good to put all implementation details in a cpp file, putting classes in their own file is an aid to understanding.

atgeirr: I have another idea, that may be worth considering. We could implement the Visitor pattern for group and well hierarchies. I think that is well suited for our situation: a stable set of classes in the tree (groups and wells) and an ever-increasing set of operations we would like to apply/compute in the hierarchy. So virtual methods are a bad choice since we would need to add new virtual methods to the nodes all the time. New visitor classes do not require changes to the well and group classes.

bska: > > In other words, I'd prefer to keep this completely hidden from other parts of the code until we can explore the problem more broadly.
> 
> I disagree with this, refactoring to move the code with generalizing potential out of its hiding place in a cpp file is good! Otherwise it becomes very unlikely that it will never be improved or reused, since any interested such user first must extract it and make sure that was ok. It is also not necessarily good to put all implementation details in a cpp file, putting classes in their own file is an aid to understanding.

We need the overall feature of this PR (reporting the well-level injection and production guide rates to the summary file) rather soon, so I suggest we merge this PR in its current state.  Then I'll extract the `GroupTreeWalker` in a follow-up PR.  I'd still prefer not to do so until we agree on its (mostly) final location though, because I can envision using it in opm-common too.

> We could implement the Visitor pattern for group and well hierarchies

Yes, that's a good idea that would complement the current facility.  For accumulation and distribution purposes you still need to know if you visit the parent before the child (or children) or the other way around, but there could be multiple hierarchy traversal strategies.

atgeirr: > We need the overall feature of this PR (reporting the well-level injection and production guide rates to the summary file) rather soon, so I suggest we merge this PR in its current state. Then I'll extract the `GroupTreeWalker` in a follow-up PR. I'd still prefer not to do so until we agree on its (mostly) final location though, because I can envision using it in opm-common too.

Ok, that I can get behind.

> For accumulation and distribution purposes you still need to know if you visit the parent before the child (or children) or the other way around, but there could be multiple hierarchy traversal strategies.

Typically all we would need for any visitor is is preVisit(Well), postVisit(Well), preVisit(Group) and postVisit(Group). We could also (orthogonally) support depth first and breadth first visit order.

bska:

bska:
jenkins build this opm-tests=653 please

bska:
The updated reference solutions have been installed on the CI system.  Merging this now per earlier approval to bring the overall system back to a consistent state.

PR #3638, opened 2021-10-26
Exclude pv of numerical aquifer cells in ratio of cells violating CNV

### LINKED ISSUE #3627: Erroneous CNV convergence detection with Aquifers and --max-tolerance-cnv.
In a model there are about three aquifer cells, but these represent a considerable fraction of the total volume of all cells. When we check the `cnvErrorFraction` in `BlackoilModelEbos.hpp` lines [830&mdash;834](https://github.com/OPM/opm-simulators/blob/1ee8322a510ef219968e734ff94b02ea24ec4519/opm/simulators/flow/BlackoilModelEbos.hpp#L830-L834), the fraction might be quite smaller than without aquifer cells and the simulator might detect convergence where it should not.

blattms:
In some models there will be only a few cells with rather high pore volume representing the numerical aquifer. Including them (as non-violated cells) in the ratio will make the ratio lower than without the aquifer there and might lead to accepting vectors as converged when they should not.

Problem found by and hint of fix was given by @totto82.

Needs OPM/opm-grid#555
Closes #3627

### DIFF
--- a/opm/simulators/flow/BlackoilModelEbos.hpp
+++ b/opm/simulators/flow/BlackoilModelEbos.hpp
@@ -576,14 +576,16 @@ namespace Opm {
         }
 
         template <class CollectiveCommunication>
-        double convergenceReduction(const CollectiveCommunication& comm,
-                                    const double pvSumLocal,
-                                    std::vector< Scalar >& R_sum,
-                                    std::vector< Scalar >& maxCoeff,
-                                    std::vector< Scalar >& B_avg)
+        std::tuple<double,double> convergenceReduction(const CollectiveCommunication& comm,
+                                                       const double pvSumLocal,
+                                                       const double numAquiferPvSumLocal,
+                                                       std::vector< Scalar >& R_sum,
+                                                       std::vector< Scalar >& maxCoeff,
+                                                       std::vector< Scalar >& B_avg)
         {
             // Compute total pore volume (use only owned entries)
             double pvSum = pvSumLocal;
+            double numAquiferPvSum = numAquiferPvSumLocal;
 
             if( comm.size() > 1 )
             {
@@ -591,7 +593,7 @@ namespace Opm {
                 std::vector< Scalar > sumBuffer;
                 std::vector< Scalar > maxBuffer;
                 const int numComp = B_avg.size();
-                sumBuffer.reserve( 2*numComp + 1 ); // +1 for pvSum
+                sumBuffer.reserve( 2*numComp + 2 ); // +2 for (numAquifer)pvSum
                 maxBuffer.reserve( numComp );
                 for( int compIdx = 0; compIdx < numComp; ++compIdx )
                 {
@@ -602,6 +604,7 @@ namespace Opm {
 
                 // Compute total pore volume
                 sumBuffer.push_back( pvSum );
+                sumBuffer.push_back( numAquiferPvSum );
 
                 // compute global sum
                 comm.sum( sumBuffer.data(), sumBuffer.size() );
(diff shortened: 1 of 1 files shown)


bska:
@blattms : Is there more work incoming here or in the upstream PR OPM/opm-grid#555?  If not, would you like me to have a look at this change?

blattms:
There is some more work coming (once I find the time) will also address the comments in opm-grid.

bska:
>> Is there more work incoming here or in the upstream PR
>
> There is some more work coming

Very good!  Do please let me know once you consider this work ready for review.

totto82:
Can this be merged now? From the simulator side I think its fine. I tried to follow the comments on the related PR in opm-grid but wasn't sure if it was ready to be merged or nor. This is a critical bug for models with numerical aquifers.

GitPaean:
As @totto82 pointed out, if this PR works as intended, it should make sure the extra parameter `--relaxed-max-pv-fraction=0` for the numerical aquifer tests can be removed without changing the results in any significant way (assuming the reference results is correct), which is directly related to the motivation of this PR.  And a `update_data` will be required because there might be non-significant changes in the results of the numerical aquifer cases.

alfbr:
> As @totto82 pointed out, if this PR works as intended, it should make sure the extra parameter `--relaxed-max-pv-fraction=0` for the numerical aquifer tests can be removed without changing the results in any significant way (assuming the reference results is correct), which is directly related to the motivation of this PR. And a `update_data` will be required because there might be non-significant changes in the results of the numerical aquifer cases.

I agree that we should update the tests removing the special parameters. However, I do believe merging this PR should not be postponed significantly by that, since right now this is a really serious bug in master. Unaware users are basically running their simulations with dramatically relaxed convergence criteria if their model contain numerical aquifers.

GitPaean:
> . Unaware users are basically running their simulations with dramatically relaxed convergence criteria if their model contain numerical aquifers.

Can they run with `--relaxed-max-pv-fraction=0` instead of relaxing tolerances? Excuse me if I might miss something here. 

And the checking for the testing case is basically a correctness checking for this PR and I do not think it takes a lot of time. If you want, I can do it later today.

alfbr:
> Can they run with `--relaxed-max-pv-fraction=0` instead of relaxing tolerances? Excuse me if I might miss something here.

Short answer, no. Long answer, there is a combination of multiple convergence criteria you can replace it with, but then you need to go back to the PR that enabled it. Otherwise, you basically need to manually to what this PR does.

> And the checking for the testing case is basically a correctness checking for this PR and I do not think it takes a lot of time. If you want, I can do the checking and report it back later today.

Sure, please do!

GitPaean:
Only tested the serial running for the full running. The well curves look very good.  I think we probably can remove the parameter `--relaxed-max-pv-fraction=0` from the test with deck file `3D_2AQU_NUM.DATA` and `3D_1AQU_3CELLS.DATA`.  Other parameters also affect the results, which need more investigation.  Due to the limited working time today, only verified the removal `--relaxed-max-pv-fraction=0`. 

[ftaq-0104_1AQU.pdf](https://github.com/OPM/opm-simulators/files/7527027/ftaq-0104_1AQU.pdf)
[ftaq-0104_2AQU.pdf](https://github.com/OPM/opm-simulators/files/7527028/ftaq-0104_2AQU.pdf)

blattms:
Merged the grid PR.

In contrast to jenkins the tests fail if I run them on my old system.  Need to check other systems. But if @GitPaean thinks the results are good this should go in. Somebody probably has to do the update_data dance (not sure where). Mayby changing the tolerances should be another PR?

alfbr:
> In contrast to jenkins the tests fail if I run them on my old system. Need to check other systems. But if @GitPaean thinks the results are good this should go in. Somebody probably has to do the update_data dance (not sure where). Mayby changing the tolerances should be another PR?

All tests are green, updating tests can be done in a separate PR. @bska or @totto82 , maybe one of you feel compelled to press the big green button :-)

totto82:
Thanks. Merging. @GitPaean Can you make a Pr with the changes in the tests since you already have investigated this.

PR #3642, opened 2021-10-27
Remove debugging aid GroupState::dump

joakim-hove:
The ability to dump a json representation was useful while debugging - suggest to remove it now.


joakim-hove:
jenkins build this please

PR #3645, opened 2021-10-29
Adjust reduction rate when well control is considered for GRUP

totto82:
The reduction rate is computed differently for cases without wells under GRUP
For a well to check whether to switch to GRUP it needed to use the reduction rate
that would have been computed if this particular well was under GRUP control
and thus recompute the reduction rate without entering the no-grup path


atgeirr (approved):
A few typos... "hiearchy"

totto82:
jenkins build this please

totto82:

totto82:
jenkins build this opm-tests=654 please

PR #3648, opened 2021-10-31
Create PhaseUsage constructor

joakim-hove:
Add a constructor to enable creating `PhaseUsage` objects without a Deck. Good for testing at least.

### DIFF
--- a/opm/core/props/BlackoilPhases.cpp
+++ b/opm/core/props/BlackoilPhases.cpp
@@ -1,0 +1,53 @@
+/*
+  Copyright 2021 Equinor ASA
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#include <algorithm>
+#include <vector>
+
+#include <opm/core/props/BlackoilPhases.hpp>
+
+Opm::PhaseUsage::PhaseUsage(std::vector<BlackoilPhases::PhaseIndex> phases)
+{
+    std::sort(phases.begin(), phases.end());
+    this->phase_used.fill(0);
+    this->phase_pos.fill(-1);
+
+    std::size_t current_pos = 0;
+    for (const auto& phase : phases) {
+        this->phase_used[phase] = 1;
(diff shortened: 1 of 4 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

totto82:
Currently we have one set of phase indexes used by the simulator and most of the well model that is defined in opm-material in the blackoil fluidsystem and another in opm-simulators/opm/core that is used by the wellState and related code. I had hoped we either could exchange usage of phaseUsage with call to the FluidSystem or construct phaseUsage from the FluidSystem.

joakim-hove:
> I had hoped we either could exchange usage of phaseUsage with call to the FluidSystem or construct phaseUsage from the FluidSystem.

That sounds like a good plan - I am not in the know of these things. It would be convenient to have such a constructor for testing of the wellmodel (which I am doing a tiny bit of right now) - but this PR should *not* be seen as part of a larger plan for a hostile takeover ;-)

joakim-hove:
jenkins build this please

opm/core/props/BlackoilPhases.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+Opm::PhaseUsage::PhaseUsage(const std::vector<BlackoilPhases::PhaseIndex>& phases)
+{
+    this->phase_used.fill(0);
+    this->phase_pos.fill(-1);
+
+    std::size_t current_pos = 0;
+    for (const auto& phase : phases) {
+        this->phase_used[phase] = 1;
+        this->phase_pos[phase] = current_pos;
```
atgeirr, line 31: I believe this breaks an assumption: that the ordering of active "phases" is the same as the canonical ordering (as given by the line
```
enum PhaseIndex { Aqua = 0, Liquid = 1, Vapour = 2, Solvent = 3, Polymer = 4, Energy = 5, PolymerMW = 6, Foam = 7, Brine = 8, ZFraction = 9 };
```
Indeed the test explicitly test this by passing Polymer before Solvent. I think the resulting positions should *not* depend on (and be the same as) the input order, but be the same irrespective of input order, the test is therefore surprising to me.

Now, I cannot immediately come up with some code lines that break if this assumption is dropped. But just the fact that "this is surprising" makes me worry that there could be some, and I would prefer to avoid having to hunt for them...

Otherwise this looks good and should be merged if we can reach a consensus on the ordering issue.

joakim-hove: I do not know these details - I have tried to create a constructor based on my understanding of the `createFromDeck` function; I obviously missed a point here. Currently the usage order will follow the (random and arbitrary) input order. I will update the PR to ensure that the items come in the canonical order.

joakim-hove:
jenkins build this please

opm/core/props/BlackoilPhases.cpp
joakim-hove: Have added a `std::sort()` to ensure that canonical ordering is respected. Have also removed an unused utility class `PhasePresence`

joakim-hove:
Ping?

PR #3649, opened 2021-11-01
changed: build separate simulator binaries for flow

akva2:
this is very convenient during development.
we can then remove the FLOW_BLACKOIL_ONLY option,
as it is no longer needed - use the flow_blackoil binary instead.
however we need to keep this support in Main.hpp due to the python
bindings relying on it.

this is in preparation for the removal of the ebos binaries.


akva2:
jenkins build this please

akva2:
jenkins build this please

bska:
Just to be clear: The regression tests will still be defined in terms of the main `flow` binary, right?  In other words, if we want to actually run regression tests locally we still need to build that executable?

akva2:
yes, all tests still use the flow binary, which i think is the correct approach.

joakim-hove:
I think this is a very positive contribution; I encourage those with the greatest umph in this part of town to merge this and the following #3650 with no unnecessary delay!

PR #3647, opened 2021-10-30
Update rates after actionx

joakim-hove:
In the case an ACTIONX block has keywords affecting wellrates - e.g. `WCONPROD` we need to force an update of the wellrates used in the simulator. This PR is mainly a refactoring to extract the functionality to update rates out from the `WellState::init()` - so that it can be called halways through a report step. The actual change is in the last commit. Must be combined with: https://github.com/OPM/opm-common/pull/2801 to produce correct results.

The essence of the refactoring is:

1. Functionality is moved from `WellState` to `SingleWellState`
2. Code is split in an injector function and a producer function.
3. The pressure in the first connection which can be used as BHP limit is assigned to the `PerfData` structure at construction time.

The refactoring in the first N -1 commits can very well be squashed together to one commit.

The pure refactoring (i.e. the 12 first commits) have been transferred to a separate pr: https://github.com/OPM/opm-simulators/pull/3659

PR #3659, opened 2021-11-01
Refactor wellstate init

joakim-hove:
This is the refactor part of: #3647

### DIFF
--- a/opm/simulators/wells/SingleWellState.hpp
+++ b/opm/simulators/wells/SingleWellState.hpp
@@ -65,7 +65,6 @@ public:
     ///    PerforationData::connection_transmissibility_factor actually
     ///    used (overwrites existing internal values).
     void reset_connection_factors(const std::vector<PerforationData>& new_perf_data);
-
     void updateStatus(Well::Status status);
     void init_timestep(const SingleWellState& other);
     void shut();
--- a/opm/simulators/wells/SingleWellState.hpp
+++ b/opm/simulators/wells/SingleWellState.hpp
@@ -55,6 +55,7 @@ public:
     Well::ProducerCMode production_cmode{Well::ProducerCMode::CMODE_UNDEFINED};
 
 
+    void updateStatus(Well::Status status);
     void init_timestep(const SingleWellState& other);
     void shut();
     void stop();
--- a/opm/simulators/wells/SingleWellState.hpp
+++ b/opm/simulators/wells/SingleWellState.hpp
@@ -33,8 +33,9 @@ namespace Opm {
 
 class SingleWellState {
 public:
-    SingleWellState(const ParallelWellInfo& pinfo, bool is_producer, std::size_t num_perf, std::size_t num_phases, double temp);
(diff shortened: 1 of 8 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
Comments to this?

opm/simulators/wells/WellState.cpp
```diff
@@ -53,144 +54,70 @@ void WellState::base_init(const std::vector<double>& cellPressures,
...
 
 
+void WellState::initSingleWell(const std::vector<double>& cellPressures,
+                               const Well& well,
+                               const std::vector<PerforationData>& well_perf_data,
+                               const ParallelWellInfo& well_info,
+                               const SummaryState& summary_state)
+{
+    if (well.isInjector() == well.isProducer())
+        throw std::logic_error(fmt::format("Well must be either producer or injector - logic error for well: {}", well.name()));
```
bska, line 107: This condition will never trigger in the current implementation since it ultimately [reduces](https://github.com/OPM/opm-common/blob/dd54d2aad1eb5517a6c4aecbedf6ecc339bfa4d9/src/opm/parser/eclipse/EclipseState/Schedule/Well/Well.cpp#L742-L748) to
```C++
WellType::producer() && !WellType::producer()
```
given that `WellType::injector()` is [`! WellType::producer()`](https://github.com/OPM/opm-common/blob/dd54d2aad1eb5517a6c4aecbedf6ecc339bfa4d9/src/opm/parser/eclipse/EclipseState/Schedule/ScheduleTypes.cpp#L167-L173).  Unless we consider introducing new well types (e.g., "observer"), then I suggest to remove that check.

opm/simulators/wells/SingleWellState.hpp
```diff
@@ -55,6 +67,16 @@ class SingleWellState {
     Well::ProducerCMode production_cmode{Well::ProducerCMode::CMODE_UNDEFINED};
 
 
+    /// Special purpose method to support dynamically rescaling a well's
+    /// CTFs through WELPI.
+    ///
+    /// \param[in] well_perf_data New perforation data.  Only
```
bska, line 73: Tiny nit: The documented parameter name should match the actual parameter name.

bska (commented):
Looks good for the most part.  Just a couple of minor points that I'd like to address before we merge this.

opm/simulators/wells/WellState.cpp
joakim-hove: I fully agree. The current logic was just to reproduce the existing `assert(...)` based approach (might have got that wrong though?).

opm/simulators/wells/SingleWellState.hpp
joakim-hove: OK

opm/simulators/wells/WellState.cpp
joakim-hove: I'll push an update without the check.

joakim-hove:
jenkins build this please

bska (approved):
Thanks for the updates. This looks good to me now and I'll merge once the build check completes.

bska:

PR #3664, opened 2021-11-04
Remove unused restart (de)serialize methods in aquifer/tracer models.

blattms:
The methods either throw or do nothing. Apparently they are not used by any code and just add to some developer's confusion. Hence I propose to remove them.

I might be wrong here and there is a purpose that I fail to see. Maybe @GitPaean and @osae should double check.

### DIFF
--- a/ebos/ecltracermodel.hh
+++ b/ebos/ecltracermodel.hh
@@ -133,23 +133,6 @@ public:
         advanceTracerFields(gas_);
     }
 
-    /*!
-     * \brief This method writes the complete state of all tracer
-     *        to the hard disk.
-     */
-    template <class Restarter>
-    void serialize(Restarter&)
-    { /* not implemented */ }
-
-    /*!
-     * \brief This method restores the complete state of the tracer
-     *        from disk.
-     *
-     * It is the inverse of the serialize() method.
-     */
-    template <class Restarter>
-    void deserialize(Restarter&)
-    { /* not implemented */ }
 
 protected:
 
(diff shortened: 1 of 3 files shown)


blattms:
jenkins build this please

totto82:
I guess it's more like signpost that (de)serialize methods needs to be added if we are going to restart using the non-ecl restart mechanism. I don't now how much that is in use anymore, but it used to work. But of-course if most new implementations just adds throws etc instead of implementing the methods the restart mechanism does not work as expected anymore.

GitPaean:
They are used in the following, while the functionality is not implemented. Not sure how to define these.  I would not bother to remove them, it might give some light for the people who want to write the restart code. But I do not have strong opinions either way. 

https://github.com/OPM/opm-simulators/blob/c563f14affbb7d451f1dc39253aa7c3db0f5cd3b/ebos/eclproblem.hh#L981

https://github.com/OPM/opm-simulators/blob/c563f14affbb7d451f1dc39253aa7c3db0f5cd3b/ebos/eclproblem.hh#L997


``` c++
    template <class Restarter>
    void deserialize(Restarter& res)
    {
        // reload the current episode/report step from the deck
        beginEpisode();

        // deserialize the wells
        wellModel_.deserialize(res);

        if (enableAquifers_)
            // deserialize the aquifer
            aquiferModel_.deserialize(res);
    }

    /*!
     * \brief This method writes the complete state of the problem and its subobjects to
     *        disk.
     *
     * The file format used here is ad-hoc.
     */
    template <class Restarter>
    void serialize(Restarter& res)
    {
        wellModel_.serialize(res);

        if (enableAquifers_)
            aquiferModel_.serialize(res);
    }
```

PR #3665, opened 2021-11-04
amgcl to prefix path

akva2:
This is just a PR to allow me experimenting with getting amgcl setup on jenkins without bypassing the build job queue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,8 @@ option(OPM_ENABLE_PYTHON "Enable python bindings?" OFF)
 option(OPM_ENABLE_PYTHON_TESTS "Enable tests for the python bindings?" ON)
 option(ENABLE_FPGA "Enable FPGA kernels integration?" OFF)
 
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

blattms:
does that mean you are patching it (amgcl/vexcl) to work with cl2.hpp/opencl.hpp? It seemst to expect the old header (did not really investigate, just jusdging from the error messages that I saw on my system)

akva2:
well i am looking into how much patchery it takes. seems like more than just replacing the header is needed. likely some of them defines we have in our opencl.hpp. feels like it might be too much hassle right now.

akva2:
yeah this is not happening.

akva2:
jenkins build this please

akva2:
@blattms okay, it is success with vexcl master. i'll enable it in the builds now.

PR #3492, opened 2021-08-26
Fixes and removes deprecation warnings for Dune 2.8

blattms:
Fixes:
- The layout parameter of MultipleCodimMultipleGeomTypeMapper was deprecated for a long time and removed in DUNE 2.8
- bilu_backsolve is deprecated and one should use Dune::ILU::blockILU0Decomposition.
- Compilation was broken due to inclusion order problems with our firstmatrixelement method (that is now CamelCase).

needs OPM/opm-models#668

PR #3672, opened 2021-11-08
Add CPR preconditioner to openclSolver

Tongdongq:
This PR allows to also use a CPR preconditioner in openclSolver (combined with ilu0).
The functionality is similar to that of OPM.

Per linear system:
- extract pressure, transform 3x3 block to scalar using quasiimpes weights
- create AMG hierarchy if not reused
- upload AMG preconditioner to GPU

Per linear iteration:
- transform blockvector to scalar vector using quasiimpes weights
- apply AMG preconditioner
- copy vector back to CPU
- solve coarse system on CPU using UMFPack
- copy vector back to GPU
- add solution to pressurecomponents of blockvector
- continue with bicgstab

Only quasiimpes weights can be used. Their functionality is duplicated from `getQuasiImpesWeights.hpp`. Those functions only take `Dune::BlockVector`.
Each level is pre- and postsmoothed using a simple diagonal jacobi smoother, which is the default option in amgcl.

Plans for a later PR:
- solve coarse system on GPU as well, probably using a nested openclSolver
- extract pressure on GPU
```
R5 2400G, NVIDIA GTX1050Ti
OPM ilu0:           566 (134+224+77+111), 1851, 1483, 21898
opencl ilu0 LS:     649 (131+311+78+110), 1806, 1444, 20648 | 0
opencl ilu0 GC:     735 (163+337+99+116), 2255, 1869, 42629 | 0
OPM CPR quasiimpes: 519 (122+195+72+101), 1694, 1342, 3934
vary num_pre/post_steps jacobi smoothing during AMG cycle:
LS, jacobi_damping=0.72:
0: 453 (102+194+47+95), 1422, 1081, 5006 | 0
1: 434 (104+170+48+96), 1463, 1119, 2828 | 0
2: 443 (107+173+49+97), 1484, 1137, 2447 | 0
3: 442 (106+176+48+96), 1466, 1123, 2257 | 0
4: 454 (105+188+48+96), 1476, 1130, 2121 | 1
LS, jacobi_damping=0.65:
0: 451 (100+194+46+95), 1422, 1081, 5006 | 0
1: 433 (105+167+47+96), 1438, 1094, 2758 | 0
2: 434 (104+171+48+95), 1453, 1109, 2437 | 0
3: 458 (112+179+52+97), 1479, 1136, 2331 | 0
4: 458 (110+182+50+99), 1458, 1113, 2159 | 0

Intel Haswell Xeon, NVIDIA Tesla V100:
OPM CPR quasiimpes: 769 (179+311+97+151), 1694, 1342, 3934
jacobi_damping=0.72
opencl CPR LS:      623 (148+222+78+144), 1463, 1119, 2828 | 0
opencl CPR GC:      676 (161+251+86+146), 1567, 1217, 4196 | 0
jacobi_damping=0.65
opencl CPR LS:      616 (148+218+77+143), 1438, 1094, 2758 | 0
opencl CPR GC:      653 (156+239+84+144), 1528, 1183, 3773 | 0
```
The `level_scheduling` is faster than the `graph_coloring`, which was not the case with just ilu0. The coloring affects the amount of parallelism during the ilu0 decomposition and application, but not for the CPR preconditioner. The increased accuracy of `level_scheduling` could account for this performance shift.
The V100 is a bit slower for the linear solve part than my GTX1050Ti, which is not expected. Perhaps there is not enough parallelism available during the AMG cycle to utilize the V100 enough.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -100,6 +100,7 @@ if(OPENCL_FOUND)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/BILU0.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/Reorder.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/ChowPatelIlu.cpp)
+  list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/CPR.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/opencl.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/openclKernels.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/openclSolverBackend.cpp)
@@ -107,11 +108,11 @@ if(OPENCL_FOUND)
 endif()
 if(CUDA_FOUND OR OPENCL_FOUND OR HAVE_FPGA OR HAVE_AMGCL)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/WellContributions.cpp)
+  list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/Matrix.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/MultisegmentWellContribution.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/BdaBridge.cpp)
 endif()
 if(HAVE_FPGA)
-  list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/FPGAMatrix.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/FPGABILU0.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/FPGASolverBackend.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/FPGAUtils.cpp)
@@ -247,10 +248,10 @@ list (APPEND PUBLIC_HEADER_FILES
   opm/simulators/linalg/bda/BdaSolver.hpp
   opm/simulators/linalg/bda/BILU0.hpp
   opm/simulators/linalg/bda/BlockedMatrix.hpp
+  opm/simulators/linalg/bda/CPR.hpp
   opm/simulators/linalg/bda/cuda_header.hpp
   opm/simulators/linalg/bda/cusparseSolverBackend.hpp
   opm/simulators/linalg/bda/ChowPatelIlu.hpp
-  opm/simulators/linalg/bda/FPGAMatrix.hpp
   opm/simulators/linalg/bda/FPGABILU0.hpp
   opm/simulators/linalg/bda/FPGASolverBackend.hpp
   opm/simulators/linalg/bda/FPGAUtils.hpp
@@ -260,6 +261,7 @@ list (APPEND PUBLIC_HEADER_FILES
   opm/simulators/linalg/bda/openclKernels.hpp
   opm/simulators/linalg/bda/openclSolverBackend.hpp
   opm/simulators/linalg/bda/openclWellContributions.hpp
+  opm/simulators/linalg/bda/Matrix.hpp
   opm/simulators/linalg/bda/MultisegmentWellContribution.hpp
   opm/simulators/linalg/bda/WellContributions.hpp
   opm/simulators/linalg/amgcpr.hh
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -104,6 +104,7 @@ if(OPENCL_FOUND)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/opencl.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/openclKernels.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/OpenclMatrix.cpp)
+  list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/opencl/Preconditioner.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/openclSolverBackend.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/openclWellContributions.cpp)
 endif()
@@ -262,6 +263,7 @@ list (APPEND PUBLIC_HEADER_FILES
   opm/simulators/linalg/bda/opencl.hpp
   opm/simulators/linalg/bda/openclKernels.hpp
   opm/simulators/linalg/bda/OpenclMatrix.hpp
+  opm/simulators/linalg/bda/opencl/Preconditioner.hpp
   opm/simulators/linalg/bda/openclSolverBackend.hpp
   opm/simulators/linalg/bda/openclWellContributions.hpp
   opm/simulators/linalg/bda/Matrix.hpp
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -147,6 +147,7 @@ list (APPEND TEST_SOURCE_FILES
   tests/test_timer.cpp
   tests/test_invert.cpp
   tests/test_stoppedwells.cpp
+  tests/test_solvetransposed3x3.cpp
   tests/test_relpermdiagnostics.cpp
   tests/test_norne_pvt.cpp
   tests/test_wellprodindexcalculator.cpp
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -103,6 +103,7 @@ if(OPENCL_FOUND)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/CPR.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/opencl.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/openclKernels.cpp)
+  list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/OpenclMatrix.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/openclSolverBackend.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/openclWellContributions.cpp)
 endif()
@@ -259,6 +260,7 @@ list (APPEND PUBLIC_HEADER_FILES
   opm/simulators/linalg/bda/ILUReorder.hpp
   opm/simulators/linalg/bda/opencl.hpp
   opm/simulators/linalg/bda/openclKernels.hpp
+  opm/simulators/linalg/bda/OpenclMatrix.hpp
   opm/simulators/linalg/bda/openclSolverBackend.hpp
   opm/simulators/linalg/bda/openclWellContributions.hpp
   opm/simulators/linalg/bda/Matrix.hpp
--- a/opm/simulators/linalg/bda/BdaBridge.cpp
+++ b/opm/simulators/linalg/bda/BdaBridge.cpp
@@ -63,7 +63,8 @@ BdaBridge<BridgeMatrix, BridgeVector, block_size>::BdaBridge(std::string acceler
                                                              double tolerance,
                                                              [[maybe_unused]] unsigned int platformID,
                                                              unsigned int deviceID,
-                                                             [[maybe_unused]] std::string opencl_ilu_reorder)
+                                                             [[maybe_unused]] std::string opencl_ilu_reorder,
+                                                             [[maybe_unused]] std::string linsolver)
 : verbosity(linear_solver_verbosity), accelerator_mode(accelerator_mode_)
 {
     if (accelerator_mode.compare("cusparse") == 0) {
@@ -88,7 +89,7 @@ BdaBridge<BridgeMatrix, BridgeVector, block_size>::BdaBridge(std::string acceler
         } else {
             OPM_THROW(std::logic_error, "Error invalid argument for --opencl-ilu-reorder, usage: '--opencl-ilu-reorder=[level_scheduling|graph_coloring]'");
         }
-        backend.reset(new Opm::Accelerator::openclSolverBackend<block_size>(linear_solver_verbosity, maxit, tolerance, platformID, deviceID, ilu_reorder));
+        backend.reset(new Opm::Accelerator::openclSolverBackend<block_size>(linear_solver_verbosity, maxit, tolerance, platformID, deviceID, ilu_reorder, linsolver));
 #else
         OPM_THROW(std::logic_error, "Error openclSolver was chosen, but OpenCL was not found by CMake");
 #endif
@@ -168,27 +169,28 @@ int checkZeroDiagonal(BridgeMatrix& mat) {
 // iterate sparsity pattern from Matrix and put colIndices and rowPointers in arrays
 // sparsity pattern should stay the same
 // this could be removed if Dune::BCRSMatrix features an API call that returns colIndices and rowPointers
-template <class BridgeMatrix>
-void getSparsityPattern(BridgeMatrix& mat, std::vector<int> &h_rows, std::vector<int> &h_cols) {
+template <class BridgeMatrix, class BridgeVector, int block_size>
+void BdaBridge<BridgeMatrix, BridgeVector, block_size>::getSparsityPattern(const BridgeMatrix& mat, std::vector<int> &h_rows, std::vector<int> &h_cols) {
     int sum_nnzs = 0;
 
-    // convert colIndices and rowPointers
-    if (h_rows.empty()) {
-        h_rows.emplace_back(0);
-            for (typename BridgeMatrix::const_iterator r = mat.begin(); r != mat.end(); ++r) {
-                int size_row = 0;
-                for (auto c = r->begin(); c != r->end(); ++c) {
-                    h_cols.emplace_back(c.index());
-                    size_row++;
-                }
-                sum_nnzs += size_row;
-                h_rows.emplace_back(sum_nnzs);
-            }
+    h_rows.clear();
+    h_cols.clear();
 
-        // h_rows and h_cols could be changed to 'unsigned int', but cusparse expects 'int'
-        if (static_cast<unsigned int>(h_rows[mat.N()]) != mat.nonzeroes()) {
-            OPM_THROW(std::logic_error, "Error size of rows do not sum to number of nonzeroes in BdaBridge::getSparsityPattern()");
+    // convert colIndices and rowPointers
+    h_rows.emplace_back(0);
+    for (typename BridgeMatrix::const_iterator r = mat.begin(); r != mat.end(); ++r) {
+        int size_row = 0;
+        for (auto c = r->begin(); c != r->end(); ++c) {
+            h_cols.emplace_back(c.index());
+            size_row++;
         }
+        sum_nnzs += size_row;
+        h_rows.emplace_back(sum_nnzs);
+    }
+
+    // h_rows and h_cols could be changed to 'unsigned int', but cusparse expects 'int'
+    if (static_cast<unsigned int>(h_rows[mat.N()]) != mat.nonzeroes()) {
+        OPM_THROW(std::logic_error, "Error size of rows do not sum to number of nonzeroes in BdaBridge::getSparsityPattern()");
     }
 } // end getSparsityPattern()
 
--- a/opm/simulators/linalg/bda/BdaBridge.cpp
+++ b/opm/simulators/linalg/bda/BdaBridge.cpp
@@ -170,7 +170,7 @@ int checkZeroDiagonal(BridgeMatrix& mat) {
 // sparsity pattern should stay the same
 // this could be removed if Dune::BCRSMatrix features an API call that returns colIndices and rowPointers
 template <class BridgeMatrix, class BridgeVector, int block_size>
-void BdaBridge<BridgeMatrix, BridgeVector, block_size>::getSparsityPattern(const BridgeMatrix& mat, std::vector<int> &h_rows, std::vector<int> &h_cols) {
+void BdaBridge<BridgeMatrix, BridgeVector, block_size>::copySparsityPatternFromISTL(const BridgeMatrix& mat, std::vector<int> &h_rows, std::vector<int> &h_cols) {
 
     h_rows.clear();
     h_cols.clear();
@@ -186,9 +186,9 @@ void BdaBridge<BridgeMatrix, BridgeVector, block_size>::getSparsityPattern(const
 
     // h_rows and h_cols could be changed to 'unsigned int', but cusparse expects 'int'
     if (static_cast<unsigned int>(h_rows[mat.N()]) != mat.nonzeroes()) {
-        OPM_THROW(std::logic_error, "Error size of rows do not sum to number of nonzeroes in BdaBridge::getSparsityPattern()");
+        OPM_THROW(std::logic_error, "Error size of rows do not sum to number of nonzeroes in BdaBridge::copySparsityPatternFromISTL()");
     }
-} // end getSparsityPattern()
+}
 
 
 template <class BridgeMatrix, class BridgeVector, int block_size>
@@ -218,7 +218,7 @@ void BdaBridge<BridgeMatrix, BridgeVector, block_size>::solve_system([[maybe_unu
         if (h_rows.capacity() == 0) {
             h_rows.reserve(Nb+1);
             h_cols.reserve(nnzb);
-            getSparsityPattern(*mat, h_rows, h_cols);
+            copySparsityPatternFromISTL(*mat, h_rows, h_cols);
         }
 
         Dune::Timer t_zeros;
--- a/opm/simulators/linalg/bda/BdaBridge.cpp
+++ b/opm/simulators/linalg/bda/BdaBridge.cpp
@@ -171,7 +171,6 @@ int checkZeroDiagonal(BridgeMatrix& mat) {
 // this could be removed if Dune::BCRSMatrix features an API call that returns colIndices and rowPointers
 template <class BridgeMatrix, class BridgeVector, int block_size>
 void BdaBridge<BridgeMatrix, BridgeVector, block_size>::getSparsityPattern(const BridgeMatrix& mat, std::vector<int> &h_rows, std::vector<int> &h_cols) {
-    int sum_nnzs = 0;
 
     h_rows.clear();
     h_cols.clear();
@@ -179,13 +178,10 @@ void BdaBridge<BridgeMatrix, BridgeVector, block_size>::getSparsityPattern(const
     // convert colIndices and rowPointers
     h_rows.emplace_back(0);
     for (typename BridgeMatrix::const_iterator r = mat.begin(); r != mat.end(); ++r) {
-        int size_row = 0;
         for (auto c = r->begin(); c != r->end(); ++c) {
             h_cols.emplace_back(c.index());
-            size_row++;
         }
-        sum_nnzs += size_row;
-        h_rows.emplace_back(sum_nnzs);
+        h_rows.emplace_back(h_cols.size());
     }
 
     // h_rows and h_cols could be changed to 'unsigned int', but cusparse expects 'int'
--- a/opm/simulators/linalg/bda/CPR.cpp
+++ b/opm/simulators/linalg/bda/CPR.cpp
@@ -86,9 +86,21 @@ bool CPR<block_size>::analyse_matrix(BlockedMatrix *mat_) {
 
 template <unsigned int block_size>
 bool CPR<block_size>::create_preconditioner(BlockedMatrix *mat_) {
+    Dune::Timer t_bilu0;
     bool result = bilu0->create_preconditioner(mat_);
+    if (verbosity >= 3) {
+        std::ostringstream out;
+        out << "CPR create_preconditioner bilu0(): " << t_bilu0.stop() << " s";
+        OpmLog::info(out.str());
+    }
     if (use_amg) {
+        Dune::Timer t_amg;
         create_preconditioner_amg(mat); // already points to bilu0::rmat if needed
+        if (verbosity >= 3) {
+            std::ostringstream out;
+            out << "CPR create_preconditioner_amg(): " << t_amg.stop() << " s";
+            OpmLog::info(out.str());
+        }
     }
     return result;
 }
@@ -500,9 +512,22 @@ void CPR<block_size>::apply_amg(const cl::Buffer& y, cl::Buffer& x) {
 
 template <unsigned int block_size>
 void CPR<block_size>::apply(const cl::Buffer& y, cl::Buffer& x) {
+    Dune::Timer t_bilu0;
     bilu0->apply(y, x);
+    if (verbosity >= 4) {
+        std::ostringstream out;
+        out << "CPR apply bilu0(): " << t_bilu0.stop() << " s";
+        OpmLog::info(out.str());
+    }
+
     if (use_amg) {
+        Dune::Timer t_amg;
         apply_amg(y, x);
+        if (verbosity >= 4) {
+            std::ostringstream out;
+            out << "CPR apply amg(): " << t_amg.stop() << " s";
+            OpmLog::info(out.str());
+        }
     }
 }
 
--- a/opm/simulators/linalg/bda/CPR.cpp
+++ b/opm/simulators/linalg/bda/CPR.cpp
@@ -383,33 +383,32 @@ void CPR<block_size>::analyzeAggregateMaps() {
     for(int level = 0; level < num_levels - 1; ++mapIter, ++level) {
         DuneAmg::AggregatesMap *map = *mapIter;
 
+        Rmatrices.emplace_back(level_sizes[level+1], level_sizes[level], level_sizes[level]);
+        std::fill(Rmatrices.back().nnzValues.begin(), Rmatrices.back().nnzValues.end(), 1.0);
+
         // get indices for each row of P and R
+        std::vector<std::vector<unsigned> > indicesR(level_sizes[level+1]);
         PcolIndices[level].resize(level_sizes[level]);
-        std::vector<std::set<long int> > indicesR(level_sizes[level+1]);
+
         using AggregateIterator = DuneAmg::AggregatesMap::const_iterator;
         for(AggregateIterator ai = map->begin(); ai != map->end(); ++ai){
             if (*ai != DuneAmg::AggregatesMap::ISOLATED) {
                 const long int diff = ai - map->begin();
                 PcolIndices[level][diff] = *ai;
-                indicesR[*ai].insert(diff);
+                indicesR[*ai].emplace_back(diff);
             }
         }
 
-        Rmatrices.emplace_back(level_sizes[level+1], level_sizes[level], level_sizes[level]);
-        std::fill(Rmatrices.back().nnzValues.begin(), Rmatrices.back().nnzValues.end(), 1.0);
-
-        // set sparsity pattern of R
         int col_idx = 0;
+        // set sparsity pattern of R
         Rmatrices.back().rowPointers[0] = 0;
         for (unsigned int i = 0; i < indicesR.size(); ++i) {
             Rmatrices.back().rowPointers[i + 1] = Rmatrices.back().rowPointers[i] + indicesR[i].size();
             for (auto it = indicesR[i].begin(); it != indicesR[i].end(); ++it) {
-                Rmatrices.back().colIndices[col_idx] = *it;
-                col_idx++;
+                Rmatrices.back().colIndices[col_idx++] = *it;
             }
         }
     }
-
 }
 
 
--- a/opm/simulators/linalg/bda/CPR.cpp
+++ b/opm/simulators/linalg/bda/CPR.cpp
@@ -279,6 +279,8 @@ void CPR<block_size>::create_preconditioner_amg(BlockedMatrix *mat_) {
             using CriterionBase = Dune::Amg::AggregationCriterion<Dune::Amg::SymmetricDependency<DuneMat, Dune::Amg::FirstDiagonal>>;
             using Criterion = Dune::Amg::CoarsenCriterion<CriterionBase>;
             const Criterion c = Opm::PreconditionerFactory<MatrixOperator, Dune::Amg::SequentialInformation>::amgCriterion(property_tree);
+            num_pre_smooth_steps = c.getNoPreSmoothSteps();
+            num_post_smooth_steps = c.getNoPostSmoothSteps();
 
             dune_amg->build<OverlapFlags>(c);
 
@@ -451,9 +453,11 @@ void CPR<block_size>::amg_cycle_gpu(const int level, cl::Buffer &y, cl::Buffer &
     cl::Buffer& u = d_u[level]; // u was 0-initialized earlier
 
     // presmooth
-    double jacobi_damping = 0.72; // default value in amgcl: 0.72
-    OpenclKernels::residual(A->nnzValues, A->colIndices, A->rowPointers, x, y, t, Ncur, 1);
-    OpenclKernels::vmul(jacobi_damping, d_invDiags[level], t, x, Ncur);
+    double jacobi_damping = 0.65; // default value in amgcl: 0.72
+    for (unsigned i = 0; i < num_pre_smooth_steps; ++i){
+        OpenclKernels::residual(A->nnzValues, A->colIndices, A->rowPointers, x, y, t, Ncur, 1);
+        OpenclKernels::vmul(jacobi_damping, d_invDiags[level], t, x, Ncur);
+    }
 
     // move to coarser level
     OpenclKernels::residual(A->nnzValues, A->colIndices, A->rowPointers, x, y, t, Ncur, 1);
@@ -462,8 +466,10 @@ void CPR<block_size>::amg_cycle_gpu(const int level, cl::Buffer &y, cl::Buffer &
     OpenclKernels::prolongate_vector(u, x, d_PcolIndices[level], Ncur);
 
     // postsmooth
-    OpenclKernels::residual(A->nnzValues, A->colIndices, A->rowPointers, x, y, t, Ncur, 1);
-    OpenclKernels::vmul(jacobi_damping, d_invDiags[level], t, x, Ncur);
+    for (unsigned i = 0; i < num_post_smooth_steps; ++i){
+        OpenclKernels::residual(A->nnzValues, A->colIndices, A->rowPointers, x, y, t, Ncur, 1);
+        OpenclKernels::vmul(jacobi_damping, d_invDiags[level], t, x, Ncur);
+    }
 }
 
 
--- a/opm/simulators/linalg/bda/CPR.cpp
+++ b/opm/simulators/linalg/bda/CPR.cpp
@@ -142,21 +142,19 @@ void solve_transposed_3x3(const double *A, const double *b, double *x) {
 template <unsigned int block_size>
 void CPR<block_size>::init_opencl_buffers() {
     d_Amatrices.reserve(num_levels);
-    d_Pmatrices.reserve(num_levels - 1);
     d_Rmatrices.reserve(num_levels - 1);
     d_invDiags.reserve(num_levels - 1);
     for (Matrix& m : Amatrices) {
         d_Amatrices.emplace_back(context.get(), m.N, m.M, m.nnzs, 1);
     }
-    for (Matrix& m : Pmatrices) {
-        d_Pmatrices.emplace_back(context.get(), m.N, m.M, m.nnzs, 1);
-        d_invDiags.emplace_back(*context, CL_MEM_READ_WRITE, sizeof(double) * m.N); // create a cl::Buffer
-        d_t.emplace_back(*context, CL_MEM_READ_WRITE, sizeof(double) * m.N);
-    }
     for (Matrix& m : Rmatrices) {
         d_Rmatrices.emplace_back(context.get(), m.N, m.M, m.nnzs, 1);
         d_f.emplace_back(*context, CL_MEM_READ_WRITE, sizeof(double) * m.N);
         d_u.emplace_back(*context, CL_MEM_READ_WRITE, sizeof(double) * m.N);
+
+        d_PcolIndices.emplace_back(*context, CL_MEM_READ_WRITE, sizeof(int) * m.M);
+        d_invDiags.emplace_back(*context, CL_MEM_READ_WRITE, sizeof(double) * m.M); // create a cl::Buffer
+        d_t.emplace_back(*context, CL_MEM_READ_WRITE, sizeof(double) * m.M);
     }
     d_weights = std::make_unique<cl::Buffer>(*context, CL_MEM_READ_WRITE, sizeof(double) * N);
     d_rs = std::make_unique<cl::Buffer>(*context, CL_MEM_READ_WRITE, sizeof(double) * N);
@@ -171,21 +169,19 @@ void CPR<block_size>::opencl_upload() {
     d_mat->upload(queue.get(), mat);
 
     err = CL_SUCCESS;
-    events.resize(Pmatrices.size() + 1);
-    for (unsigned int i = 0; i < Pmatrices.size(); ++i) {
+    events.resize(2 * Rmatrices.size() + 1);
+    err |= queue->enqueueWriteBuffer(*d_weights, CL_FALSE, 0, sizeof(double) * N, weights.data(), nullptr, &events[0]);
+    for (unsigned int i = 0; i < Rmatrices.size(); ++i) {
         d_Amatrices[i].upload(queue.get(), &Amatrices[i]);
 
-        err |= queue->enqueueWriteBuffer(d_invDiags[i], CL_FALSE, 0, sizeof(double) * Amatrices[i].N, invDiags[i].data(), nullptr, &events[i]);
+        err |= queue->enqueueWriteBuffer(d_invDiags[i], CL_FALSE, 0, sizeof(double) * Amatrices[i].N, invDiags[i].data(), nullptr, &events[2*i+1]);
+        err |= queue->enqueueWriteBuffer(d_PcolIndices[i], CL_FALSE, 0, sizeof(int) * Amatrices[i].N, PcolIndices[i].data(), nullptr, &events[2*i+2]);
     }
-    err |= queue->enqueueWriteBuffer(*d_weights, CL_FALSE, 0, sizeof(double) * N, weights.data(), nullptr, &events[Pmatrices.size()]);
     cl::WaitForEvents(events);
     events.clear();
     if (err != CL_SUCCESS) {
         // enqueueWriteBuffer is C and does not throw exceptions like C++ OpenCL
         OPM_THROW(std::logic_error, "CPR OpenCL enqueueWriteBuffer error");
-    }
-    for (unsigned int i = 0; i < Pmatrices.size(); ++i) {
-        d_Pmatrices[i].upload(queue.get(), &Pmatrices[i]);
     }
     for (unsigned int i = 0; i < Rmatrices.size(); ++i) {
         d_Rmatrices[i].upload(queue.get(), &Rmatrices[i]);
@@ -333,8 +329,7 @@ void CPR<block_size>::analyzeHierarchy() {
     diagIndices.resize(num_levels);
 
     Amatrices.reserve(num_levels);
-    Pmatrices.reserve(num_levels - 1);  // coarsest level does not need one
-    Rmatrices.reserve(num_levels - 1);
+    Rmatrices.reserve(num_levels - 1);  // coarsest level does not need one
     invDiags.reserve(num_levels);
 
     Amatrices.clear();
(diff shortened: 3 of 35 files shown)


opm/simulators/linalg/bda/BdaBridge.cpp
```diff
@@ -216,7 +217,7 @@ void BdaBridge<BridgeMatrix, BridgeVector, block_size>::solve_system([[maybe_unu
         if (h_rows.capacity() == 0) {
             h_rows.reserve(Nb+1);
             h_cols.reserve(nnzb);
-            getSparsityPattern(*mat, h_rows, h_cols);
+            getSparsityPattern(*mat, h_rows.data(), h_cols.data());
```
bska, line 220: This is undefined behaviour as there are no objects in `h_rows.data()` or `h_cols.data()` if you just use `vector<>::reserve(N)`.  We will have `vector<>::size() == 0`, but `vector<>::capacity() >= N`.  It is of course **likely** that this won't blow up given that these are `vector<double>`s, but if you really want to use `vector<>::data()` instead of passing a reference to the `vector` itself, then you must ensure that the vector is properly initialised.

bska (commented):
I did not read this in any kind of detail, but I would like to know your reasoning for changing a parameter type from
```C++
std::vector<double>&
```
into the more low-level
```C++
double*
```
Especially, when you're not really passing a valid pointer to the function.

Tongdongq:
I need a `Matrix` from `Matrix.hpp` to be filled with the sparsity pattern of a `Dune::BCRSMatrix` for the AMG cycle. `Matrix` only has raw pointers to store its sparsity pattern.
I suppose it's better to just replace them with `std::vector`.

opm/simulators/linalg/bda/BdaBridge.cpp
```diff
@@ -164,27 +165,27 @@ int checkZeroDiagonal(BridgeMatrix& mat) {
...
+    for (typename BridgeMatrix::const_iterator r = mat.begin(); r != mat.end(); ++r) {
+        int size_row = 0;
+        for (auto c = r->begin(); c != r->end(); ++c) {
+            h_cols[nnz_idx] = c.index();
+            size_row++;
+            nnz_idx++;
         }
+        sum_nnzs += size_row;
+        h_rows[r.index() + 1] = sum_nnzs;
+    }
```
blattms, line 184: Probably will not find the time for a full review this week, maybe we will/shoudl discuss on Monday at our meeting.

Just a note here (in case I might forget):
I don't see why saw a need to change this. Please never resort to C if you can use C++ (especially if you are actually changing C++ containers). 
@bska is completely right with his concern. You are just reserving and assuming that for an empty vector with size 0 but reserved space `data()` will return the pointer to he reserved space. Unfortunately, the standard allows returning a nullpointer.

In addition just reserving space only worked before because you used `emplace_back` here as it will adapt the size of the container. With your change to C that is not the case anymore and the vectors will remain at size 0 which will be unexpected to fellow developer (including you in the future).

blattms:
You are using a template parameter block_size in some of your class, yet it seems to be only used when calculating buffer sizes.
Can you please double check.

In that case please do not use a template for this. It should be possible to make it a normal class member and save compile time and hassle (and you can easily move the implementations to a *.cpp file). In case the parameter is needed because a few methods take e.g. a `FieldMatrix<double, block_size...>` then you can just make those functions take a template parameter.

Tongdongq:
I reverted the change to raw pointer. It still uses a `std::vector<int>&` and the members of `Opm::Accelerator::Matrix` now use `std::vector<int>` to store the sparsity pattern.

opm/simulators/linalg/bda/BdaBridge.cpp
```diff
@@ -168,27 +169,28 @@ int checkZeroDiagonal(BridgeMatrix& mat) {
...
+    // convert colIndices and rowPointers
+    h_rows.emplace_back(0);
+    for (typename BridgeMatrix::const_iterator r = mat.begin(); r != mat.end(); ++r) {
+        int size_row = 0;
+        for (auto c = r->begin(); c != r->end(); ++c) {
+            h_cols.emplace_back(c.index());
+            size_row++;
         }
+        sum_nnzs += size_row;
+        h_rows.emplace_back(sum_nnzs);
```
blattms, line 188: Probably a matter of taste, but alas: One could skip the counters `size_row` (not really used) and `sum_nnz` (`==h_cols.size()`).

CMakeLists_files.cmake
```diff
@@ -267,6 +268,7 @@ list (APPEND PUBLIC_HEADER_FILES
   opm/simulators/linalg/bda/openclKernels.hpp
   opm/simulators/linalg/bda/openclSolverBackend.hpp
   opm/simulators/linalg/bda/openclWellContributions.hpp
+  opm/simulators/linalg/bda/Matrix.hpp
```
blattms, line 271: Aren't we missing CPR.hpp?

opm/simulators/linalg/bda/Matrix.hpp
```diff
@@ -17,41 +17,83 @@
...
 namespace Opm
 {
 namespace Accelerator
 {
 
+class Matrix;
+
+/// This struct resembles a csr matrix, only doubles are supported
+/// The matrix data is stored in OpenCL Buffers
+class OpenclMatrix {
```
blattms, line 37: Please change the filenames to reflect the class name

opm/simulators/linalg/bda/Matrix.cpp
```diff
@@ -17,17 +17,43 @@
...
+    }
+}
+
+void OpenclMatrix::upload(cl::CommandQueue *queue, Matrix *matrix) {
+    upload(queue, matrix->nnzValues.data(), matrix->colIndices.data(), matrix->rowPointers.data());
+}
+
+void OpenclMatrix::upload(cl::CommandQueue *queue, BlockedMatrix *matrix) {
+    upload(queue, matrix->nnzValues, matrix->colIndices, matrix->rowPointers);
+}
```
blattms, line 55: Should we add some checks here? It seems like e.g. uploading a `Matrix` if `blocksize>1` might have surprising effects.

opm/simulators/linalg/bda/Matrix.hpp
```diff
@@ -17,41 +17,83 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-#ifndef FPGA_MATRIX_HEADER_INCLUDED
-#define FPGA_MATRIX_HEADER_INCLUDED
+#ifndef MATRIX_HEADER_INCLUDED
+#define MATRIX_HEADER_INCLUDED
```
blattms, line 21: This is a rather generic header name and chances are that other use it, too and if that is the case compilation will break as it might already have been defined elsewhere. Please at least add a OPM_ prefix.

opm/simulators/linalg/bda/Matrix.hpp
```diff
@@ -17,41 +17,83 @@
...
     }
 
+    /// Allocate rectangular Matrix and data arrays with given sizes
+    /// \param[in] N               number of rows
+    /// \param[in] M               number of columns
+    /// \param[in] nnzs            number of nonzeros
+    Matrix(int N_, int M_, int nnzs_)
+    : N(N_),
+      M(M_),
+      nnzs(nnzs_)
```
blattms, line 89: I think you can call other constructors here and save some code:
```
Matrix(int N_, int M_, int nnzs_)
: Matrix(N_, nnzs_), M(M_)
{}
```

opm/simulators/linalg/bda/Matrix.hpp
```diff
@@ -17,41 +17,83 @@
...
-    /// Allocate Matrix and data arrays with given sizes
+    /// Allocate square Matrix and data arrays with given sizes
     /// \param[in] N               number of rows
     /// \param[in] nnzs            number of nonzeros
     Matrix(int N_, int nnzs_)
-    : nnzValues(new double[nnzs_]),
-      colIndices(new int[nnzs_]),
-      rowPointers(new int[N_+1]),
-      N(N_),
+    : N(N_),
```
blattms, line 74: Please initialize `M`, too.

opm/simulators/linalg/bda/Matrix.hpp
```diff
@@ -17,41 +17,83 @@
...
-
-    /// All constructors allocate new memory, so always delete here
-    ~Matrix(){
-        delete[] nnzValues;
-        delete[] colIndices;
-        delete[] rowPointers;
+    {
+        nnzValues.resize(nnzs);
+        colIndices.resize(nnzs);
+        rowPointers.resize(N+1);
```
blattms, line 79: Cool, more C++. :+1:

opm/simulators/linalg/bda/openclSolverBackend.cpp
```diff
@@ -191,6 +203,19 @@ openclSolverBackend<block_size>::openclSolverBackend(int verbosity_, int maxit_,
     }
 }
 
+template <unsigned int block_size>
+openclSolverBackend<block_size>::openclSolverBackend(int verbosity_, int maxit_, double tolerance_, ILUReorder opencl_ilu_reorder_) :
+    BdaSolver<block_size>(verbosity_, maxit_, tolerance_), use_cpr(false), opencl_ilu_reorder(opencl_ilu_reorder_)
+{
+    std::cout << "openclSolverBackend() 2\n";
```
blattms, line 210: Should we remove the debug output?

opm/simulators/linalg/bda/openclSolverBackend.cpp
```diff
@@ -251,13 +276,19 @@ void openclSolverBackend<block_size>::gpu_pbicgstab(WellContributions& wellContr
...
-        t_prec.stop();
+        t_bilu0.start();
+        bilu0->apply(d_p, d_pw);
+        t_bilu0.stop();
+
+        if (use_cpr) {
+            t_cpr.start();
+            cpr->apply(d_p, d_pw);
+            t_cpr.stop();
+        }
```
blattms, line 287: This looks strange to me. When I read CPR, I assume that it it the full preconditioner (iilu on the fine system, plus the correction on the pressure only system). Here it seems like the CPR is just rerstriction to pressure system, solve for correction, prolongate and add correction.

What is the reason, why we can't put the whole CPR in a class/struct and only have one application of the preconditioner here that does it all? (Why can either use virtual inheritance here, or type erasure with std::function)

Thus we would get rid of all the rather ugly if clauses introduced here.

blattms (changes requested):
Sorry for the long delay. Here is the first batch of my review.

Please be aware that I still need to review openCLKernels* and Cpr.cpp. Hope to get that done before our meeting.

opm/simulators/linalg/bda/Matrix.hpp
Tongdongq: I'll create new files for `OpenclMatrix`.

opm/simulators/linalg/bda/BdaBridge.cpp
```diff
@@ -168,27 +169,28 @@ int checkZeroDiagonal(BridgeMatrix& mat) {
...
+    h_cols.clear();
 
-        // h_rows and h_cols could be changed to 'unsigned int', but cusparse expects 'int'
-        if (static_cast<unsigned int>(h_rows[mat.N()]) != mat.nonzeroes()) {
-            OPM_THROW(std::logic_error, "Error size of rows do not sum to number of nonzeroes in BdaBridge::getSparsityPattern()");
+    // convert colIndices and rowPointers
+    h_rows.emplace_back(0);
+    for (typename BridgeMatrix::const_iterator r = mat.begin(); r != mat.end(); ++r) {
+        int size_row = 0;
+        for (auto c = r->begin(); c != r->end(); ++c) {
```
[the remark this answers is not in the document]
Tongdongq: The function would silently not do anything. The caller needed the empty the vector beforehand. Now it always (re)writes the sparsity pattern.

opm/simulators/linalg/bda/CPR.cpp
```diff
@@ -0,0 +1,515 @@
...
+        std::vector<std::set<long int> > indicesP(level_sizes[level]);
+        std::vector<std::set<long int> > indicesR(level_sizes[level+1]);
+        using AggregateIterator = DuneAmg::AggregatesMap::const_iterator;
+        for(AggregateIterator ai = map->begin(); ai != map->end(); ++ai){
+            if (*ai != DuneAmg::AggregatesMap::ISOLATED) {
+                const long int diff = ai - map->begin();
+                indicesP[diff].insert(*ai);
+                indicesR[*ai].insert(diff);
+            }
+        }
```
blattms, line 351: I think this is a bit of overkill (even if you insist on using matrices for both P and R). Unless a node isolated the corresponding row of P should only have one entry otherwise none. Hence the set will only have one entry or none. That could be determined withour set (and its complexity), using e.g. a vector with either 1 or 0.

For R you need to know the number of dofs per aggregate. Here you could simply count these (e.g. in a vector containing the size for each coarse grid dof) and then loop over the aggregates map again to determine the column index when creating the sparsity pattern.

Complexity of iterating a second time over the aggregates map should not be worse than over two vectors of sets. (Famous last words...)

opm/simulators/linalg/bda/CPR.cpp
```diff
@@ -0,0 +1,515 @@
...
+                col_idx++;
+            }
+        }
+    }
+
+}
+
+
+void solve_coarse_umfpack(const Matrix *A, std::vector<double> &y, std::vector<double> &x) {
+    const int N = A->N;
```
blattms, line 384: I do not understand why we are doing it the way we do it here. Here is my understanding of what we do:

- We use the coarsest matrix (not in BCRSMatrix format) despite having the BCRSMatrix lying around anyway
- Because UmfPack-bindings of dune-istl do not understand it we convert it back to BCRSMatrix
- I assume the UmfPack bindings now convert it back to a matrix of C-pointers (but using column compressed row storage).

I would recommend to either use the available BCRSMatrix or to use the C respresentation, use umfpack routines to transpose it to column compressed storage (User guide Section 12) and call umfpack directly.

In any case we should probably only setup umfpack (decompose) if the matrix changes and not in every coarse grid solve.

opm/simulators/linalg/bda/CPR.cpp
```diff
@@ -0,0 +1,515 @@
...
+    int Nnext = d_Amatrices[level+1].Nb;
+
+    cl::Buffer& t = d_t[level];
+    cl::Buffer& f = d_f[level];
+    cl::Buffer& u = d_u[level]; // u was 0-initialized earlier
+
+    // presmooth
+    double jacobi_damping = 0.72; // default value in amgcl: 0.72
+    OpenclKernels::residual(A->nnzValues, A->colIndices, A->rowPointers, x, y, t, Ncur, 1);
+    OpenclKernels::vmul(jacobi_damping, d_invDiags[level], t, x, Ncur);
```
blattms, line 460: Maybe we should allow for a bit more flexibiility (at least a few jacobi steps)?

Anyway, interesting that one Jacobi step seems to work.

blattms (changes requested):
Thanks for the contribution.

This completes my review.

opm/simulators/linalg/bda/openclSolverBackend.cpp
Tongdongq: Squashed this in

opm/simulators/linalg/bda/Matrix.hpp
Tongdongq: 845485ae

opm/simulators/linalg/bda/Matrix.hpp
Tongdongq: 845485ae99

opm/simulators/linalg/bda/Matrix.hpp
Tongdongq: 613cfd2e

opm/simulators/linalg/bda/Matrix.cpp
Tongdongq: 8c42e4fa

opm/simulators/linalg/bda/Matrix.hpp
Tongdongq: 3b8bb461

opm/simulators/linalg/bda/BdaBridge.cpp
Tongdongq: afb25e9

opm/simulators/linalg/bda/openclKernels.cpp
```diff
@@ -395,6 +507,27 @@ void OpenclKernels::apply_stdwells_no_reorder(cl::Buffer& d_Cnnzs_ocl, cl::Buffe
...
+            __global double const *in2,
+            __global double *out,
+            const int N)
+        {
+            unsigned int NUM_THREADS = get_global_size(0);
+            int idx = get_global_id(0);
+
+            while(idx < N){
+                out[idx] += alpha * in1[idx] * in2[idx];
+                idx += NUM_THREADS;
```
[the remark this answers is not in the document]
Tongdongq: The comment was outdated, I copied the comment from the header now.
```
/// multiply vector with another vector and a scalar, element-wise
/// add result to a third vector
```
This explains what the method does, although it could be hard to read.

opm/simulators/linalg/bda/openclKernels.cpp
```diff
@@ -506,15 +639,77 @@ void OpenclKernels::apply_stdwells_no_reorder(cl::Buffer& d_Cnnzs_ocl, cl::Buffe
         )";
     }
 
-    std::string OpenclKernels::get_spmv_blocked_string() {
+
+    // transform blocked vector to scalar vector using pressure-weights
+    // every workitem handles one blockrow
+    std::string OpenclKernels::get_move_to_coarse_string() {
```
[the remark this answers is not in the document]
Tongdongq: 4cf01847

opm/simulators/linalg/bda/openclKernels.cpp
```diff
@@ -506,15 +639,77 @@ void OpenclKernels::apply_stdwells_no_reorder(cl::Buffer& d_Cnnzs_ocl, cl::Buffe
...
+                coarse_y[target_block_row] = sum;
+                target_block_row += NUM_THREADS;
+            }
+        }
+        )";
+    }
+
+    // add the coarse pressure solution back to the finer, complete solution
+    // every workitem handles one blockrow
+    std::string OpenclKernels::get_move_to_fine_string() {
```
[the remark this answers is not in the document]
Tongdongq: 4cf01847

opm/simulators/linalg/bda/openclKernels.cpp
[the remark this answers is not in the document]
Tongdongq: 4cf01847

opm/simulators/linalg/bda/CPR.cpp
```diff
@@ -0,0 +1,515 @@
...
+    this->nnz = nnzb_ * block_size * block_size;
+
+    context = context_;
+    queue = queue_;
+
+    coarse_vals.resize(nnzb);
+    coarse_x.resize(Nb);
+    coarse_y.resize(Nb);
+    weights.resize(N);
+    diagIndices.resize(1);
```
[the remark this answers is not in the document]
Tongdongq: The `diagIndices` are used on every AMG level to calculate the jacobi smoother.
The first level is created before the number of level is known. First, the pressure is extracted and the scalar matrix is created. Then the AMG hierarchy is created. Then the jacobi values are calculated.
The first level indices are used to calculate the quasiimpes weights.

opm/simulators/linalg/bda/CPR.cpp
```diff
@@ -0,0 +1,515 @@
...
+                assert(candidate != mat->colIndices + end);
+                diagIndices[0][row] = candidate - mat->colIndices;
+            }
+        }
+
+        // calculate weights for each row
+        for (int row = 0; row < Nb; ++row) {
+            // solve to find weights
+            double *row_weights = weights.data() + block_size * row; // weights for this row
+            solve_transposed_3x3(mat->nnzValues + block_size * block_size * diagIndices[0][row], rhs, row_weights);
```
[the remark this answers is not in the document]
Tongdongq: For some reason it looks like the opposite:
``` c++
if (transpose) {
    diag_block.solve(bweights, rhs);
} else {
    auto diag_block_transpose = Details::transposeDenseMatrix(diag_block);
    diag_block_transpose.solve(bweights, rhs);
}
```
The standard `cpr` option takes the `else` branch.

opm/simulators/linalg/bda/CPR.cpp
```diff
@@ -0,0 +1,515 @@
...
+
+            // normalize weights for this row
+            double abs_max = get_absmax(row_weights, block_size);
+            for(unsigned int i = 0; i < block_size; i++){
+                row_weights[i] /= abs_max;
+            }
+        }
+
+        // transform blocks to scalars
```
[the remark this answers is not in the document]
Tongdongq: 21164574

opm/simulators/linalg/bda/CPR.cpp
```diff
@@ -0,0 +1,515 @@
...
+                d_Rmatrices.emplace_back(context.get(), m.N, m.M, m.nnzs, 1);
+                d_f.emplace_back(*context, CL_MEM_READ_WRITE, sizeof(double) * m.N);
+                d_u.emplace_back(*context, CL_MEM_READ_WRITE, sizeof(double) * m.N);
+            }
+            d_weights = std::make_unique<cl::Buffer>(*context, CL_MEM_READ_WRITE, sizeof(double) * N);
+            d_rs = std::make_unique<cl::Buffer>(*context, CL_MEM_READ_WRITE, sizeof(double) * N);
+            d_mat = std::make_unique<OpenclMatrix>(context.get(), Nb, Nb, nnzb, block_size);
+            d_coarse_y = std::make_unique<cl::Buffer>(*context, CL_MEM_READ_WRITE, sizeof(double) * Nb);
+            d_coarse_x = std::make_unique<cl::Buffer>(*context, CL_MEM_READ_WRITE, sizeof(double) * Nb);
+        });
```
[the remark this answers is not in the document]
Tongdongq: Please check a6315d6c

opm/simulators/linalg/bda/CPR.cpp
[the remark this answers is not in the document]
blattms: Thanks for the clarification

opm/simulators/linalg/bda/CPR.cpp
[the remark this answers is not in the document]
blattms: Much better, thanks

opm/simulators/linalg/bda/openclKernels.cpp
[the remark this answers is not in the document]
blattms: Thanks

opm/simulators/linalg/bda/CPR.cpp
Tongdongq: 6fa456ea
Now reuses the umfpack object. `setMatrix()` takes about 3.4ms, `apply()` only 0.25ms. This used to be 4.1ms each linear iteration due to the things you mentioned.
I also removed `solve_coarse_umfpack()`, it would now only consist of one line.

opm/simulators/linalg/bda/CPR.cpp
```diff
@@ -0,0 +1,515 @@
...
+            for (auto c = r->begin(); c != r->end(); ++c) {
+                Amatrices.back().nnzValues[nnz_idx] = A[r.index()][c.index()];
+                if (fillDiagIndices && r.index() == c.index()) {
+                    diagIndices[level].emplace_back(nnz_idx);
+                }
+                nnz_idx++;
+            }
+        }
+
+        Opm::BdaBridge<DuneMat, DuneVec, 1>::getSparsityPattern(A, Amatrices.back().rowPointers, Amatrices.back().colIndices);
```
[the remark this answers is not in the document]
Tongdongq: 19e03e68

opm/simulators/linalg/bda/CPR.cpp
Tongdongq: It also seems that 0 Jacobi steps works, although it takes more linear iterations and is a little slower. I updated benchmarks in the startpost.

Tongdongq: We could add more runtime parameters, perhaps
`--opencl-cpr-jacobi-damping-factor` and `--opencl-cpr-jacobi-num-steps`.

opm/simulators/linalg/bda/CPR.cpp
```diff
@@ -0,0 +1,515 @@
...
+                for (unsigned int i = 0; i < block_size; ++i) {
+                    value += block[block_size * i + pressure_idx] * row_weights[i];
+                }
+                coarse_vals[idx] = value;
+            }
+        }
+
+        using Communication = Dune::OwnerOverlapCopyCommunication<int, int>;
+        using OverlapFlags = Dune::NegateSet<Communication::OwnerSet>;
+        if (recalculate_aggregates) {
```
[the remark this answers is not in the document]
Tongdongq: I'm not sure what is a good long term plan. We should probably keep the interface with the `cusparseSolver` and `fpgaSolver` similar.

opm/simulators/linalg/bda/CPR.cpp
```diff
@@ -0,0 +1,515 @@
...
+            abs_max = std::fabs(data[i]);
+        }
+    }
+    return abs_max;
+}
+
+
+// solve A^T * x = b
+void solve_transposed_3x3(const double *A, const double *b, double *x) {
+    const int B = 3;
```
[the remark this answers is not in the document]
Tongdongq: Can I test this without turning it into a public member function?

opm/simulators/linalg/bda/CPR.cpp
blattms: Please don't introduce any opencl-cpr specific options. We can use what is there already.

> It also seems that 0 Jacobi steps works

You mean no pre- and no post-smoothing at all in the pressure solve?

Well, I guess that is probably rather caused by the fact that ILU0 is often good enough and the course grid direction does not make it worse too much without smoothing and we only solve very inaccurately. If we would increase the the accuracy, this would probably change. According multigrid theory and my experience smoothing is needed.

opm/simulators/linalg/bda/CPR.cpp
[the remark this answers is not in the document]
blattms: Well you need to declare the free function in a header.

opm/simulators/linalg/bda/CPR.cpp
```diff
@@ -0,0 +1,515 @@
...
+    double t10 = A[0+2*B] * A[1+0*B];
+    double t12 = A[0+1*B] * A[2+0*B];
+    double t14 = A[0+2*B] * A[2+0*B];
+
+    double d = (t4*A[2+2*B]-t6*A[2+1*B]-t8*A[2+2*B]+
+          t10*A[2+1*B]+t12*A[1+2*B]-t14*A[1+1*B]); //determinant
+
+    x[0] = (b[0]*A[1+1*B]*A[2+2*B] - b[0]*A[2+1*B]*A[1+2*B]
+          - b[1] *A[0+1*B]*A[2+2*B] + b[1]*A[2+1*B]*A[0+2*B]
+          + b[2] *A[0+1*B]*A[1+2*B] - b[2]*A[1+1*B]*A[0+2*B]) / d;
```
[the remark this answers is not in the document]
blattms: Well, I did take out a sheet of paper and tried to calculate this myself. Either I made a mistake (possible as I am not good with indices) or you. E.g. x_0 looks like this on paper
```
x_0 = 1/|A| *(|A^T_{00}| b_0 - |A^T_{10}| b_1 + |A_{20}| b_2) =
       = 1/|A|  * ( (a_{11} * a_{22}  - a_{21} * a_{12}) * b_0 
                  - (a_{10} * a_{22}  - a_{20} * a_{12}) * b_1
                  + (a_{10} * a_{21} - a_{20} * a_{11})*b[2])
```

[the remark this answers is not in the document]
blattms: Did send you my sheet of paper as pm.

opm/simulators/linalg/bda/CPR.cpp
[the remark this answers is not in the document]
blattms: Indeed. Sorry for the noise

opm/simulators/linalg/bda/CPR.cpp
```diff
@@ -0,0 +1,515 @@
...
+        // transform blocks to scalars
+        for (int row = 0; row < Nb; ++row) {
+            int start = mat->rowPointers[row];
+            int end = mat->rowPointers[row + 1];
+            for (int idx = start; idx < end; ++idx) {
+                double *block = mat->nnzValues + idx * block_size * block_size;
+                double *row_weights = weights.data() + block_size * row;
+                double value = 0.0;
+                for (unsigned int i = 0; i < block_size; ++i) {
+                    value += block[block_size * i + pressure_idx] * row_weights[i];
```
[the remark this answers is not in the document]
blattms: dito

opm/simulators/linalg/bda/openclSolverBackend.cpp
Tongdongq: I combined them, now `openclSolver` has a `cpr` object, which has a `bilu0` object.
I moved the `use_cpr` bool to the `CPR` class, and renamed it to `use_amg`, but I'm not sure if that is the best name. Perhaps `pressure_correction` is a better name.

opm/simulators/linalg/bda/CPR.cpp
Tongdongq: Yes, 0 Jacobi steps also converges better than only using ilu0.
I'll get a value from the `Criterion`.

opm/simulators/linalg/bda/CPR.cpp
Tongdongq: I managed to remove the P matrices, and replaced them with a vector of just colIndices and a new kernel.
R is also filled differently. I'm not sure what you mean by iterating the aggregates map twice, I only see one.

blattms:
After rebasing onto current master I am getting compile errors with DUNE 2.7 and g++-10. I guess this is still based on DUNE 2.6 (which we actually do not support anymore since we released 2021.10). For 2.7 the header to be included is matrixhierarchy.hh, and the constructor takes a shared pointer to the Operator. `Dune::stackobject_to_shared_ptr` from dune/common/shared_ptr.hh might help here.
While on it, please also test DUNE 2.8 and rebase this branch onto current master to get rid off the conflicts.

```
In file included from /usr/include/c++/10/memory:83,
                 from opm-common/opm/common/OpmLog/OpmLog.hpp:23,
                 from opm-simulators/opm/simulators/linalg/bda/CPR.cpp:22:
/usr/include/c++/10/bits/unique_ptr.h: In instantiation of ‘typename std::_MakeUniq<_Tp>::__single_object std::make_unique(_Args&& ...) [with _Tp = Dune::Amg::MatrixHierarchy<Dune::MatrixAdapter<Dune::BCRSMatrix<Dune::FieldMatrix<double, 1, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> > >, Dune::Amg::SequentialInformation>; _Args = {Dune::MatrixAdapter<Dune::BCRSMatrix<Dune::FieldMatrix<double, 1, 1>, std::allocator<Dune::FieldMatrix<double, 1, 1> > >, Dune::BlockVector<Dune::FieldVector<double, 1>, std::allocator<Dune::FieldVector<double, 1> > >, Dune::BlockVector<Dune::FieldVector<double, 1>, std::allocator<Dune::FieldVector<double, 1> > > >&, Dune::Amg::SequentialInformation&}; typename std::_MakeUniq<_Tp>::__single_object = std::unique_ptr<Dune::Amg::MatrixHierarchy<Dune::MatrixAdapter<Dune::BCRSMatrix<Dune::FieldMatrix<double, 1, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> > >, Dune::Amg::SequentialInformation> >]’:
opm-simulators/opm/simulators/linalg/bda/CPR.cpp:234:49:   required from ‘void Opm::Accelerator::CPR<block_size>::create_preconditioner(Opm::Accelerator::BlockedMatrix*) [with unsigned int block_size = 1]’
opm-simulators/opm/simulators/linalg/bda/CPR.cpp:478:1:   required from here
/usr/include/c++/10/bits/unique_ptr.h:962:30: error: no matching function for call to ‘Dune::Amg::MatrixHierarchy<Dune::MatrixAdapter<Dune::BCRSMatrix<Dune::FieldMatrix<double, 1, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> > >, Dune::Amg::SequentialInformation>::MatrixHierarchy(Dune::MatrixAdapter<Dune::BCRSMatrix<Dune::FieldMatrix<double, 1, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> > >&, Dune::Amg::SequentialInformation&)’
  962 |     { return unique_ptr<_Tp>(new _Tp(std::forward<_Args>(__args)...)); }
  |                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/dune/istl/paamg/amg.hh:10,
                 from opm-simulators/opm/simulators/linalg/twolevelmethodcpr.hh:14,
                 from opm-simulators/opm/simulators/linalg/PressureTransferPolicy.hpp:25,
                 from opm-simulators/opm/simulators/linalg/PressureSolverPolicy.hpp:23,
                 from opm-simulators/opm/simulators/linalg/OwningTwoLevelPreconditioner.hpp:24,
                 from opm-simulators/opm/simulators/linalg/PreconditionerFactory.hpp:26,
                 from opm-simulators/opm/simulators/linalg/bda/CPR.cpp:26:
/usr/include/dune/istl/paamg/matrixhierarchy.hh:395:5: note: candidate: ‘Dune::Amg::MatrixHierarchy<M, PI, A>::MatrixHierarchy(std::shared_ptr<_Tp>, std::shared_ptr<_Tp>) [with M = Dune::MatrixAdapter<Dune::BCRSMatrix<Dune::FieldMatrix<double, 1, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> > >; PI = Dune::Amg::SequentialInformation; A = std::allocator<Dune::MatrixAdapter<Dune::BCRSMatrix<Dune::FieldMatrix<double, 1, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> > > >]’
  395 |     MatrixHierarchy<M,IS,A>::MatrixHierarchy(std::shared_ptr<MatrixOperator> fineMatrix,
      |     ^~~~~~~~~~~~~~~~~~~~~~~
/usr/include/dune/istl/paamg/matrixhierarchy.hh:395:78: note:   no known conversion for argument 1 from ‘Dune::MatrixAdapter<Dune::BCRSMatrix<Dune::FieldMatrix<double, 1, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> > >’ to ‘std::shared_ptr<Dune::MatrixAdapter<Dune::BCRSMatrix<Dune::FieldMatrix<double, 1, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> > > >’
  395 |     MatrixHierarchy<M,IS,A>::MatrixHierarchy(std::shared_ptr<MatrixOperator> fineMatrix,
      |                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
/usr/include/dune/istl/paamg/matrixhierarchy.hh:58:11: note: candidate: ‘Dune::Amg::MatrixHierarchy<Dune::MatrixAdapter<Dune::BCRSMatrix<Dune::FieldMatrix<double, 1, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> > >, Dune::Amg::SequentialInformation>::MatrixHierarchy(const Dune::Amg::MatrixHierarchy<Dune::MatrixAdapter<Dune::BCRSMatrix<Dune::FieldMatrix<double, 1, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> >, Dune::BlockVector<Dune::FieldVector<double, 1> > >, Dune::Amg::SequentialInformation>&)’
   58 |     class MatrixHierarchy
      |           ^~~~~~~~~~~~~~~
/usr/include/dune/istl/paamg/matrixhierarchy.hh:58:11: note:   candidate expects 1 argument, 2 provided
make[2]: *** [CMakeFiles/opmsimulators.dir/build.make:992: CMakeFiles/opmsimulators.dir/opm/simulators/linalg/bda/CPR.cpp.o] Fehler 1
```

Tongdongq:
I still have to test Dune 2.8.

opm/simulators/linalg/bda/CPR.cpp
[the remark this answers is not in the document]
Tongdongq: I added two tests from NORNE, and one easy to verify-by-hand test.

blattms:
jenkins build this please

blattms:
That issue has been fixed a while ago via OPM/opm-models#668 and #3492. Please make sure that all models are up to date and rebase did not undo those changes.

blattms:
jenkins faills to build:
```
In file included from /usr/include/c++/10/memory:83,
                 from /usr/include/boost/smart_ptr/detail/sp_counted_impl.hpp:35,
                 from /usr/include/boost/smart_ptr/detail/shared_count.hpp:27,
                 from /usr/include/boost/smart_ptr/shared_ptr.hpp:17,
                 from /usr/include/boost/shared_ptr.hpp:17,
                 from /usr/include/boost/test/tools/assertion_result.hpp:21,
                 from /usr/include/boost/test/tools/old/impl.hpp:20,
                 from /usr/include/boost/test/test_tools.hpp:46,
                 from /usr/include/boost/test/unit_test.hpp:18,
                 from ../../tests/test_openclSolver.cpp:24:
/usr/include/c++/10/bits/unique_ptr.h: In instantiation of 'typename std::_MakeUniq<_Tp>::__single_object std::make_unique(_Args&& ...) [with _Tp = Opm::BdaBridge<Dune::BCRSMatrix<Dune::FieldMatrix<double, 3, 3>, std::allocator<Dune::FieldMatrix<double, 3, 3> > >, Dune::BlockVector<Dune::FieldVector<double, 3>, std::allocator<Dune::FieldVector<double, 3> > >, 3>; _Args = {const std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, const std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, const int&, const int&, const double&, const int&, const int&, const std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&}; typename std::_MakeUniq<_Tp>::__single_object = std::unique_ptr<Opm::BdaBridge<Dune::BCRSMatrix<Dune::FieldMatrix<double, 3, 3>, std::allocator<Dune::FieldMatrix<double, 3, 3> > >, Dune::BlockVector<Dune::FieldVector<double, 3>, std::allocator<Dune::FieldVector<double, 3> > >, 3>, std::default_delete<Opm::BdaBridge<Dune::BCRSMatrix<Dune::FieldMatrix<double, 3, 3>, std::allocator<Dune::FieldMatrix<double, 3, 3> > >, Dune::BlockVector<Dune::FieldVector<double, 3>, std::allocator<Dune::FieldVector<double, 3> > >, 3> > >]':
../../tests/test_openclSolver.cpp:86:71:   required from 'Dune::BlockVector<Dune::FieldVector<double, bz> > testOpenclSolver(const ptree&, const string&, const string&) [with int bz = 3; boost::property_tree::ptree = boost::property_tree::basic_ptree<std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >; std::string = std::__cxx11::basic_string<char>]'
../../tests/test_openclSolver.cpp:102:66:   required from here
/usr/include/c++/10/bits/unique_ptr.h:962:30: error: no matching function for call to 'Opm::BdaBridge<Dune::BCRSMatrix<Dune::FieldMatrix<double, 3, 3>, std::allocator<Dune::FieldMatrix<double, 3, 3> > >, Dune::BlockVector<Dune::FieldVector<double, 3>, std::allocator<Dune::FieldVector<double, 3> > >, 3>::BdaBridge(const std::__cxx11::basic_string<char>&, const std::__cxx11::basic_string<char>&, const int&, const int&, const double&, const int&, const int&, const std::__cxx11::basic_string<char>&)'
  962 |     { return unique_ptr<_Tp>(new _Tp(std::forward<_Args>(__args)...)); }
      |                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from ../../tests/test_openclSolver.cpp:32:
../../opm/simulators/linalg/bda/BdaBridge.hpp:58:5: note: candidate: 'Opm::BdaBridge<BridgeMatrix, BridgeVector, block_size>::BdaBridge(std::string, std::string, int, int, double, unsigned int, unsigned int, std::string, std::string) [with BridgeMatrix = Dune::BCRSMatrix<Dune::FieldMatrix<double, 3, 3>, std::allocator<Dune::FieldMatrix<double, 3, 3> > >; BridgeVector = Dune::BlockVector<Dune::FieldVector<double, 3>, std::allocator<Dune::FieldVector<double, 3> > >; int block_size = 3; std::string = std::__cxx11::basic_string<char>]'
   58 |     BdaBridge(std::string accelerator_mode, std::string fpga_bitstream, int linear_solver_verbosity, int maxit, double tolerance,
      |     ^~~~~~~~~
../../opm/simulators/linalg/bda/BdaBridge.hpp:58:5: note:   candidate expects 9 arguments, 8 provided
In file included from ../../tests/test_openclSolver.cpp:32:
../../opm/simulators/linalg/bda/BdaBridge.hpp:38:7: note: candidate: 'Opm::BdaBridge<Dune::BCRSMatrix<Dune::FieldMatrix<double, 3, 3>, std::allocator<Dune::FieldMatrix<double, 3, 3> > >, Dune::BlockVector<Dune::FieldVector<double, 3>, std::allocator<Dune::FieldVector<double, 3> > >, 3>::BdaBridge(Opm::BdaBridge<Dune::BCRSMatrix<Dune::FieldMatrix<double, 3, 3>, std::allocator<Dune::FieldMatrix<double, 3, 3> > >, Dune::BlockVector<Dune::FieldVector<double, 3>, std::allocator<Dune::FieldVector<double, 3> > >, 3>&&)'
   38 | class BdaBridge
      |       ^~~~~~~~~
../../opm/simulators/linalg/bda/BdaBridge.hpp:38:7: note:   candidate expects 1 argument, 8 provided
[34/263] Building CXX object CMakeFiles/flow_distribute_z.dir/flow/flow_distribute_z.cpp.o
[35/263] Building CXX object CMakeFiles/flow_libmicp.dir/flow/flow_ebos_micp.cpp.o
[36/263] Building CXX object CMakeFiles/flow_liboilwater_brine.dir/flow/flow_ebos_oilwater_brine.cpp.o
ninja: build stopped: subcommand failed.
Recording test results
ERROR: Step ?Publish JUnit test result report? failed: No test report files were found. Configuration error?
[GNU C Compiler (gcc)] Skipping execution of recorder since overall result is 'FAILURE'
Setting status of 374f8276dc62f865d8a9196148dabe41e67f4f0b to FAILURE with url https://ci.opm-project.org/job/opm-simulators-PR-builder/3350/ and message: 'Build finished. '
Finished: FAILURE
```

Tongdongq:
I have the right branches, from Dec 1. All 5 commits from those 2 PRs show up in `git log`. I also checked `ParallelOverlappingILU0.hpp` and it has that `Dune::ILU::blockILU0Decomposition` line.

blattms:
jenkins build this please

tests/test_solvetransposed3x3.cpp
```diff
@@ -0,0 +1,58 @@
...
+                                            0.125114, 1.24919e-11, 0,
+                                            -11.3854, 1.32995e-06, 0.065447},
+                                           {-0.00354818, 2.24354e-08, 55.9563,
+                                            0.284031, 2.83543e-11, 0,
+                                            -62.1156, 0.000392767, 6350.26},
+                                           {1, 0, 2,
+                                            3, 1, 0,
+                                            0, 0, 3}};
+
+    std::vector<double> b = {0, 1, 0};
```
blattms, line 45: I think you should use multiple right hand sides and make sure that each entry is at least for one rhs not zero. Currently you might not be testing the whole range of the inversion.

tests/test_solvetransposed3x3.cpp
```diff
@@ -0,0 +1,58 @@
...
+{
+    const unsigned numTests = 3;
+    const unsigned blockSize = 3;
+
+    std::vector<std::vector<double> > A = {{-0.0973322, 1.18758e-08, -0.000185231,
+                                            0.125114, 1.24919e-11, 0,
+                                            -11.3854, 1.32995e-06, 0.065447},
+                                           {-0.00354818, 2.24354e-08, 55.9563,
+                                            0.284031, 2.83543e-11, 0,
+                                            -62.1156, 0.000392767, 6350.26},
```
blattms, line 40: I am a bit unsure whether we really should use these test matrices. They are really ill-conditioned and we cannot assume that our computed result is correct. Yet we are assuming our algorithm (and test data) to be exact as we allow only 10^-6% difference. 

If I take e.g. the second matrix and use octave to solve it, the second entry of the solution has opposite sign.
```
C=[ -0.00354818, 2.24354e-08, 55.9563; 0.284031, 2.83543e-11, 0; 62.1156, 0.000392767, 6350.26];
linsolve(C', [0;1;0])
```
```
ans =

  -2.9083e+05
  -5.6408e+05
   2.5627e+03
```


maybe we should use artificially generated matrices (a bit less ill conditioned) and another program for computing the reference solution?

blattms (commented):
Another simple test would have been to test against the other approach of first transposing the matrix and then solving and compare results.

tests/test_solvetransposed3x3.cpp
Tongdongq: Do you mean generate randoms with Python or something like that? I'm not sure where to get them.

blattms: I meant you as a random number generator should think of a few 3x3 ones. They should not be to random, not symmetric, invertible, use values that are equal to integers or fractions.

Tongdongq: I changed the testcases, and also solve them via Dune.

blattms:
jenkins build this please

blattms:
Would you mind supporting "--linsolver=cpr", currently I get
```
Error: [/home/mblatt/src/opm/opm-simulators/opm/simulators/linalg/bda/openclSolverBackend.cpp:58] Error unknown value for argument --linsolver, cpr
using this needs expert knowledge. One has to provide ```--accelerator-mode=opencl --linsolver=cpr_quasiimpes```. Without accelerator ```--linsolver=cpr``` would work (but default to trueimpes weights).

Maybe we should change the error message for when passing cpr or cpr_trueimpes to include what actually works? Also indicate this in the help message?

**Comment by blattms** [comment]
The last comment can be addressed in a subsequent PR. I am testing this a bit and intend to merge it today (famous last words :wink:).

**Comment by Tongdongq** [comment]
I can turn it into
``` c++
} else if (linsolver.compare("cpr_trueimpes") == 0 || linsolver.compare("cpr") == 0) {
    OPM_THROW(std::logic_error, "Error openclSolver does not support --linsolver=cpr or cpr_trueimpes, only cpr_quasiimpes is supported");
} else {
    OPM_THROW(std::logic_error, "Error unknown value for argument --linsolver: " + linsolver);
}
```
in the next PR.

bska:
> I can turn it into

Sure, but any chance you could write it as
```C++
linsolver == "cpr_trueimpes" || linsolver == "cpr"
```
instead of using
```C++
linsolver.compare(/* ... */) == 0
```
The former is more idiomatic C++.

blattms:
On my system this seems to run very long. Just tried to understand your numbers but failed miserably as there some explanations missing. What does this line mean:
```
0: 453 (102+194+47+95), 1422, 1081, 5006 | 0
```

blattms:
I tried running this with norne like this on my system with a GeForce GTX 1060 6GB
```
flow --accelerator-mode=opencl --linsolver=cpr_quasiimpes  NORNE_ATW2013
```
and have experienced a lot of timestep chopping (down to 0.33 days) which should not happen. Finally the program aborted with
```

Starting time step 4, stepsize 1.63289 days, at day 1311.86/1320, date = 09-Jun-2001

Problem: Solver convergence failure - Iteration limit reached
Timestep chopped to 0.538855 days


Starting time step 4, stepsize 0.538855 days, at day 1311.86/1320, date = 09-Jun-2001
flow: opm/simulators/wells/StandardWellEval.cpp:119: static double Opm::StandardWellEval<FluidSystem, Indices, Scalar>::relaxationFactorFractionsProducer(const std::vector<double>&, const BVectorWell&) [with FluidSystem = Opm::BlackOilFluidSystem<double, Opm::BlackOilDefaultIndexTraits>; Indices = Opm::BlackOilIndices<0, 0, 0, 0, false, false, 0, 0>; Scalar = double; Opm::StandardWellEval<FluidSystem, Indices, Scalar>::BVectorWell = Dune::BlockVector<Dune::DynamicVector<double, std::allocator<double> >, std::allocator<Dune::DynamicVector<double, std::allocator<double> > > >]: Assertion `relaxation_factor >= 0.0 && relaxation_factor <= 1.0' failed.
[frau-mahlzahn:24983] *** Process received signal ***
[frau-mahlzahn:24983] Signal: Aborted (6)
[frau-mahlzahn:24983] Signal code:  (-6)
[frau-mahlzahn:24983] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x14140)[0x7f07f2165140]
[frau-mahlzahn:24983] [ 1] /lib/x86_64-linux-gnu/libc.so.6(gsignal+0x141)[0x7f07dcde6ce1]
[frau-mahlzahn:24983] [ 2] /lib/x86_64-linux-gnu/libc.so.6(abort+0x123)[0x7f07dcdd0537]
[frau-mahlzahn:24983] [ 3] /lib/x86_64-linux-gnu/libc.so.6(+0x2540f)[0x7f07dcdd040f]
[frau-mahlzahn:24983] [ 4] /lib/x86_64-linux-gnu/libc.so.6(+0x34662)[0x7f07dcddf662]
[frau-mahlzahn:24983] [ 5] ../../opm-simulators/opm-parallel/bin/flow(+0x1a3b407)[0x556018083407]
[frau-mahlzahn:24983] [ 6] ../../opm-simulators/opm-parallel/bin/flow(+0x1a3c96b)[0x55601808496b]
[frau-mahlzahn:24983] [ 7] ../../opm-simulators/opm-parallel/bin/flow(+0x340c90)[0x556016988c90]
[frau-mahlzahn:24983] [ 8] ../../opm-simulators/opm-parallel/bin/flow(+0x3b0751)[0x5560169f8751]
[frau-mahlzahn:24983] [ 9] ../../opm-simulators/opm-parallel/bin/flow(+0x38a282)[0x5560169d2282]
[frau-mahlzahn:24983] [10] ../../opm-simulators/opm-parallel/bin/flow(+0x3c41af)[0x556016a0c1af]
[frau-mahlzahn:24983] [11] ../../opm-simulators/opm-parallel/bin/flow(+0x3e8701)[0x556016a30701]
[frau-mahlzahn:24983] [12] ../../opm-simulators/opm-parallel/bin/flow(+0x43cfb7)[0x556016a84fb7]
[frau-mahlzahn:24983] [13] ../../opm-simulators/opm-parallel/bin/flow(+0x454e66)[0x556016a9ce66]
```

This assert should be fixed by #3712 and made me realize that this branch is not rebased. I will try another run but would assume that the time step chopping will persist. maybe the assert is a consequence of the solver not really working.

In the mean time, please tell me what properties of CPR will be used in my run (such as smoother, pre/post smother steps, damping factors etc.).

And during writing this I realized that maybe I should add `--matrix-add-well-contributions=true`. Will try that, too. Seems like we will not get the price for "User friendliest software 2022".

blattms:
`--matrix-add-well-contributions=true` indeed fixes the timestep chopping.

blattms:
Well should have tested a serial... Master is broken now (at least with OpenCL there). Will fix ASAP.

bska:
With latest master, including #3731, I'm getting a build failure on a system (Ubuntu 20.04 LTS, CMake 3.16.3, GCC 9.3.0) that does not have OpenCL in any form:
```
/home/bska/work/opm/src/opm/opm-simulators/opm/simulators/linalg/bda/opencl.hpp:31:10: fatal error: CL/cl2.hpp: No such file or directory
   31 | #include <CL/cl2.hpp>                   // supports up to OpenCL 1.2
      |          ^~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/test_solvetransposed3x3.dir/build.make:63: CMakeFiles/test_solvetransposed3x3.dir/tests/test_solvetransposed3x3.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1725: CMakeFiles/test_solvetransposed3x3.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```
Full compilation backtrace attached.  We presumably need to be a little more diligent in omitting OpenCL dependent components when OpenCL is unavailable.

---

[build.err.txt](https://github.com/OPM/opm-simulators/files/7680903/build.err.txt)

blattms:
@Tongdongq Eventually we will need to do an apple to apple comparision of OpenCL and CPU CPR. Please answer my questions about the parameters used for this. We need to set up the CPR on the CPU accrordingly using a json file (there is one output to *.PRT file that can serve as a template) for such a test.

blattms:
It will be interesting to see whether we can replicate the convergence issues on the CPU when using Jacobi within AMG,

Tongdongq:
I updated the startpost with some clarifications.
I haven't used a json file with the CPU implementation.
The AMG parameters in `opm/simulators/linalg/PreconditionerFactory.hpp:126` are probably used, as I don't populate the parameter `property_tree`, except the `alpha` is set to 0.333333333333, similar to `opm/simulators/linalg/setupPropertyTree_impl.hpp:91`.

The runtime parameters are `--linsolver=cpr_quasiimpes --matrix-add-well-contributions=true`, since seperate wellcontributions are bugged.

PR #3673, opened 2021-11-09
Add check for group production control for LRAT and WRAT in the gasslift code

totto82:
Testing on modified versions of tests from model5. We are currently working on adding more tests for gasslift that should also cover LRAT and WRAT controlled groups. 

Note that the first commit is unrelated and fixes a bug in runOptimize1 that is not used by default. It is only triggered if compiled if line https://github.com/OPM/opm-simulators/blob/ad44564f277c7923960c43bba1295375aad24bb8/opm/simulators/wells/GasLiftSingleWellGeneric.cpp#L61 is set to true.

### DIFF
--- a/opm/simulators/wells/GasLiftSingleWellGeneric.cpp
+++ b/opm/simulators/wells/GasLiftSingleWellGeneric.cpp
@@ -317,8 +317,9 @@ computeInitialWellRates_(std::vector<double>& potentials)
 
 void
 GasLiftSingleWellGeneric::
-debugCheckNegativeGradient_(double grad, double alq, double new_alq, double oil_rate,
-                            double new_oil_rate, double gas_rate, double new_gas_rate, bool increase) const
+debugCheckNegativeGradient_(double grad, double alq, double new_alq,
+                            double oil_rate, double new_oil_rate,
+                            double gas_rate, double new_gas_rate, bool increase) const
 {
     {
         const std::string msg = fmt::format("calculating gradient: "
@@ -471,6 +472,22 @@ getGasRateWithLimit_(const std::vector<double>& potentials) const
     return { new_rate, limit};
 }
 
+std::pair<double, bool>
+GasLiftSingleWellGeneric::
+getWaterRateWithLimit_(const std::vector<double>& potentials) const
+{
+    double new_rate = -potentials[this->water_pos_];
+    bool limit = false;
+    if (this->controls_.hasControl(Well::ProducerCMode::WRAT)) {
+        auto target = this->controls_.water_rate;
+        if (new_rate > target) {
+            new_rate = target;
+            limit = true;
+        }
+    }
+    return { new_rate, limit};
+}
+
 // NOTE: If the computed oil rate is larger than the target
 //   rate of the well, we reduce it to the target rate. This
 //   will make the economic gradient smaller than it would be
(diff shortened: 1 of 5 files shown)


opm/simulators/wells/GasLiftSingleWellGeneric.cpp
```diff
@@ -843,15 +843,15 @@ GasLiftSingleWellGeneric::
...
+    int dec_count = this->well_state_.gliftGetAlqDecreaseCount(this->well_name_);
     if (dec_count == 0 && inc_count == 0) {
         state = tryIncreaseLiftGas_();
-        if (!state && !state->alqChanged()) {
+        if (!state || !(state->alqChanged())) {
             state = tryDecreaseLiftGas_();
         }
     }
-    else if (dec_count == 0) {
+     else if (dec_count == 0) {
```
akva2: meh.

totto82: some leftovers from debugging. The space is fixed in the next commit.

akva2: ic

totto82:
jenkins build this please

akva2:
please process this before https://github.com/OPM/opm-simulators/pull/3674, https://github.com/OPM/opm-simulators/pull/3675 and https://github.com/OPM/opm-simulators/pull/3676

bska (approved):
I guess this patch follows the conventions in the rest of the gas lift implementation, but I really think we should try to avoid using `std::tuple<double, double, ..., double>` as function return types.  I do foresee refactoring this component in the future, but for now I'll merge into master.

opm/simulators/wells/GasLiftSingleWellGeneric.cpp
```diff
@@ -484,44 +501,152 @@ std::pair<double, bool>
...
+            double new_gr_water_rate = gr_water_rate + efficiency * delta_water;
+            double new_gr_oil_rate = gr_oil_rate + efficiency * delta_oil;
+            double new_gr_liquid_rate = new_gr_water_rate + new_gr_oil_rate;
+            if (new_gr_liquid_rate > *gr_liquid_target_opt) {
+                const std::string msg = fmt::format("limiting liquid rate to group target: "
+                    "computed group rate: {}, target: {}", new_gr_liquid_rate, *gr_liquid_target_opt);
+                displayDebugMessage_(msg);
+                double oil_fraction = new_gr_oil_rate / new_gr_liquid_rate;
+                double water_rate_limited = water_rate + (1.0 - oil_fraction) * (new_gr_liquid_rate - *gr_liquid_target_opt) / efficiency;
+                double oil_rate_limited = oil_rate + oil_fraction * (new_gr_liquid_rate - *gr_liquid_target_opt) / efficiency;
```
hakonhagland, line 635: This should probably be
```
double water_rate_limited = water_rate + (1.0 - oil_fraction) * (*gr_liquid_target_opt - gr_liquid_rate) / efficiency;
double oil_rate_limited = oil_rate + oil_fraction * (*gr_liquid_target_opt - gr_liquid_rate) / efficiency;
```
where `gr_liquid_rate = gr_water_rate + gr_oil_rate` since we are reducing the liquid rate down to the target.

PR #3674, opened 2021-11-09
added: utility to broadcast multiple variables in one operation

akva2:
Then we can micro-optimize thngs like https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/BlackoilWellModel_impl.hpp#L968

### DIFF
--- a/opm/simulators/utils/ParallelRestart.hpp
+++ b/opm/simulators/utils/ParallelRestart.hpp
@@ -361,6 +361,53 @@ ADD_PACK_PROTOTYPES(RestartValue)
 ADD_PACK_PROTOTYPES(std::string)
 ADD_PACK_PROTOTYPES(time_point)
 
+template<typename T, typename... Args>
+void variadic_packsize(size_t& size, Parallel::Communication comm, T& first, Args&&... args)
+{
+  size += packSize(first, comm);
+  if constexpr (sizeof...(args) > 0)
+      variadic_packsize(size, comm, std::forward<Args>(args)...);
+}
+
+template<typename T, typename... Args>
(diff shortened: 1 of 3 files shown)


opm/simulators/utils/ParallelRestart.hpp
```diff
@@ -361,6 +361,53 @@ ADD_PACK_PROTOTYPES(RestartValue)
 ADD_PACK_PROTOTYPES(std::string)
 ADD_PACK_PROTOTYPES(time_point)
 
+template<typename T, typename... Args>
+void variadic_packsize(size_t& size, Parallel::Communication comm, T& first, Args&... args)
+{
+  size += packSize(first, comm);
+  if constexpr (sizeof...(args) > 0)
+      variadic_packsize(size, comm, args...);
+}
```
bska, line 370: Any reason why you don't preserve the value categories of `Args` here?

akva2: probably ignorance. i don't quite know what you are referring to.

bska: > i don't quite know what you are referring to.

Taking the pack by reference, instead of forwarding reference, and not `std::forward<>`-ing it on to the callee:
```C++
template <typename T, typename... Args>
void variadic_packsize(..., Args&&... args)
{
    // ...
    variadic_packsize(size, comm, std::forward<Args>(args)...)
}
```

akva2: ah, i gotcha. that is of course better, no reason why i didn't, so changed.

akva2:
jenkins build this please

bska (approved):

PR #3678, opened 2021-11-09
Fix gasslift reduction when under LRAT, GRAT or WRAT

totto82:


totto82:
jenkins build this please

opm/simulators/wells/GasLiftSingleWellGeneric.cpp
```diff
@@ -805,30 +805,68 @@ reduceALQtoOilTarget_(double alq,
...
+        gas_rate = -potentials[this->gas_pos_];
+        if (oil_is_limited && this->controls_.hasControl(Well::ProducerCMode::ORAT) && oil_rate < this->controls_.oil_rate) {
+            limiting_control = Well::ProducerCMode::ORAT;
+            break;
+        }
+        if (oil_is_limited && this->controls_.hasControl(Well::ProducerCMode::LRAT) && (oil_rate + water_rate) < this->controls_.liquid_rate) {
+            limiting_control = Well::ProducerCMode::LRAT;
+            break;
+        }
+        if (gas_is_limited && gas_rate < this->controls_.gas_rate) {
```
hakonhagland, line 818: Shouldn't there be a check if the well has a GRAT control here? Similar to what is done above on line 810 and 814 for ORAT and LRAT..

totto82: Since gas is limited we now that it has a GRAT control. I could add a assert.

alfbr:
I have now tested, and confirm that this fixes the bug. Thank you!

PR #3680, opened 2021-11-10
Use Main::runStatic in PyBlackoilSimulator

akva2:
Motivation is getting rid of the FLOW_BLACKOIL_ONLY hack.

### DIFF
--- a/opm/simulators/flow/Main.hpp
+++ b/opm/simulators/flow/Main.hpp
@@ -24,21 +24,19 @@
 
 #include <flow/flow_ebos_blackoil.hpp>
 
-# ifndef FLOW_BLACKOIL_ONLY
-#  include <flow/flow_ebos_gasoil.hpp>
-#  include <flow/flow_ebos_oilwater.hpp>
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

opm/simulators/flow/Main.hpp
```diff
@@ -290,7 +288,6 @@ namespace Opm
 
             if (false) {}
 
-#ifndef FLOW_BLACKOIL_ONLY
```
bska, line 293: If we're really removing this check then we can also remove the `if (false) {}` check above.

akva2: ah yes.

bska (commented):
Very good.  Maybe we'll even be able to move some of the implementation to a `.cpp` file, especially the `run*()` functions.

akva2:
jenkins build this please

bska (approved):
Very good.  I'll merge once #3682 is in.

bska:
> I'll merge once #3682 is in.

That PR has been merged.  Merging this now as promised.

PR #3675, opened 2021-11-09
Use multivariable broadcast facility to broadcast rates

akva2:
This was the motivation for https://github.com/OPM/opm-simulators/pull/3674
Sits on top of that and should be processed after

PR #3682, opened 2021-11-10
fix serial build after PR #3675

akva2:
annoyingly, dune has a casting operator for mpicollectivecommunication
-> MPIComm but for some reason does not have a casting operator for
the dummy collectivecommunication -> No_Comm


akva2:

bska (approved):
Thanks.  I'll merge into master to restore the build.

PR #3684, opened 2021-11-10
cosmetics: do not indent for namespace

akva2:
also avoid an unnecessary close/open for namespace Opm


bska (approved):

PR #3686, opened 2021-11-10
Geo actionx

joakim-hove:
Support MULTXYZ in ACTIONX block. Register that MULTXYZ has been invoked in an ACTIONX block and recalculate transmissibilities.

Downstream of: https://github.com/OPM/opm-common/pull/2832


joakim-hove:
jenkins build this please

PR #3688, opened 2021-11-10
Fixes compilation of WellContributions with deactivated OpenCL.

blattms:
OpenCL header was included and OpenclKernels used unconditionally in WellContributions.cpp. As this file is also compiled if CUDA is found, it lead to undefined references for e.g. cl::CommandQueue if opencl headers were there and compile error if not.

The header is included (if needed) and appropriately guarded in WellContributions.hpp, already.


blattms:
jenkins build this please

akva2:
this is definitely a fix. merging.

PR #3690, opened 2021-11-11
Some simple cleanup in BDA WellContributions

akva2:
Just some obvious simplifications.

PR #3691, opened 2021-11-11
Refactor BDA well contributions

akva2:
split in API specific classes for Cuda/OpenCL
    
this to
1) it's cleaner
2) it avoids pulling in openCL code in cuda classes which leads to clashes between nvidia headers and opencl.hpp
    
there is still too much API specific things in interface between the
bda components to work through a virtual interface so we still have to cast
to the relevant implementation in various places.

sits on top of #3690 which should be processed first.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -49,6 +49,7 @@ list (APPEND MAIN_SOURCE_FILES
   opm/simulators/linalg/FlexibleSolver6.cpp
   opm/simulators/linalg/PropertyTree.cpp
   opm/simulators/linalg/setupPropertyTree.cpp
+  opm/simulators/linalg/bda/WellContributions.cpp
   opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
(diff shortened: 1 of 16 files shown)


akva2:
jenkins build this cuda please

Tongdongq:
I'm trying it now.
It looks like the first 3 commits are the same as in #3690.
Doesn't that mean that one is obsolete?

akva2:
As I noted in pr description that should be processed separately as it is the simple things. I'll move some more commits from here to there, so this PR only holds the refactoring eventually but I have to go get my kid in the kindergarden now.

Tongdongq:
I only tested the first few linear systems for now.
It runs normally with CUDA and OpenCL, and with just OpenCL for all reordering methods.
cusparseSolver also works like expected.

akva2:
great. then we need to figure out why the exception handling in test_cusparsesolver fails to do its job unlike the test_openclsolver and we can enable ci for the cuda code.

akva2:
jenkins build this cuda please

PR #3694, opened 2021-11-12
Use rst_deck capability to interpret commandline argument as file for restart testing

joakim-hove:
Small simplification using the rst_deck application possibility to treat argument as proper filename

PR #3698, opened 2021-11-14
Suggestion: Rename restart case to ${CASE}_RESTART in regression test

joakim-hove:
When the restart test option is invoked for the regression tests currently the equivalent of these shell commands are invoked:
```bash
mkdir $RESULT_PATH/restart
cp $RESULT_PATH/$CASE.UNRST $RESULT_PATH/restart
cd $RESULT_PATH/restart
flow $CASE --sched-restart ....
```

The important point is that the name `$CASE` is used both for the initial simulation and the restarted simulation, and in particular the result file `restart/$CASE.UNRST` will contain a mix of the base case simulation and the restarted simulation. In this PR I suggest renaming the restarted simulation `${CASE}_RESTART` - discovered while debugging the failure of #3694

Will require a data update.

### DIFF
--- a/tests/run-regressionTest.sh
+++ b/tests/run-regressionTest.sh
@@ -71,21 +71,21 @@ then
   echo "=== Executing restart run ==="
   mkdir -p ${RESULT_PATH}/restart
   cp -f ${RESULT_PATH}/${FILENAME}.UNRST ${RESULT_PATH}/restart
-  ${RST_DECK_COMMAND}  ${INPUT_DATA_PATH}/${FILENAME}.DATA ${FILENAME}:${RESTART_STEP} -m inline -s > ${RESULT_PATH}/restart/${FILENAME}.DATA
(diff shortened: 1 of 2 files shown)


joakim-hove:
jenkins build this please

joakim-hove:

joakim-hove:
jenkins build this opm-tests=670 please

joakim-hove:

joakim-hove:
jenkins build this opm-tests=671 please

akva2:
seems fine to me. you'll have to do some manual cleaning in opm-tests afterwards.

akva2 (approved):

joakim-hove:
>you'll have to do some manual cleaning in opm-tests afterwards.

Ehhh - that I did not understand?

akva2:
the reference data using the old filenames will still be in there.

joakim-hove:
>the reference data using the old filenames will still be in there.

Ok sure - I'll clean up that.

akva2:
jenkins run volve update_volve_data please

PR #3700, opened 2021-11-15
Include limits header where needed for g++-11

blattms:
Without this compilation fails on Debian unstable.


blattms:
jenkins build this please

PR #3703, opened 2021-11-17
Fixes compile error incomplete type WellContributions

blattms:
Happened with g++-8, deactivated opencl, deacivated cuda, but active amgcl.

```
    In file included from opm-simulators/opm/simulators/flow/BlackoilModelEbos.hpp:53,
                     from opm-simulators/opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.hpp:26,
                     from opm-simulators/flow/flow_ebos_gaswater.cpp:28:
    opm-simulators/opm/simulators/linalg/ISTLSolverEbos.hpp: In member function ‘bool Opm::ISTLSolverEbos<TypeTag>::solve(Opm::ISTLSolverEbos<TypeTag>::Vector&)’:
    opm-simulators/opm/simulators/linalg/ISTLSolverEbos.hpp:265:56: error: incomplete type ‘Opm::WellContributions’ used in nested name specifier
                     auto wellContribs = WellContributions::create(accelerator_mode, useWellConn_);
                                                            ^~~~~~
```


blattms:
don't know how jenkins works these days.

blattms:
jenkins build this please

akva2:
it will build with cuda, opencl and amgcl so no flags required. however, you are now unconditionally including a header that will lead to linker errors for symbols in MultisegmentWellContribution with shared libraries. The include needs to go under the ifdefs below.

blattms:
I doubt that. No symbols will be searched for that are not used. Of course, if we call a function defined in the header that calls another one in the .cc file that might happen.

Anyway will guard the include.

blattms:
jenkins build this please

akva2:
that's what triggered the trouble earlier, namely, the unique_ptr's destructor called in the wellcontrib's destructor caused linker issues as its definition is in the header. anyways, guarding the include is the correct thing to do.

Issue #3668, opened 2021-11-05
2021-10-RC2 Oil-Water Case Saturation Type Issue.


OPMUSER:
I don't think this should stop the release, as there is a simple work around.

For a dead oil PVT case using Format Type II saturation tables, as for example:

```
SWFN
0.22   0.0   0.48
0.3    0.07  0.27
0.4    0.15  0.21
0.5    0.24  0.17
0.6    0.33  0.14                                                                                                                                               
0.8    0.65  0.07
0.9    0.83  0.03
1.0    1.0   0.0   /

SOF2
0.04   0.0
0.1    0.022
0.2    0.1
0.3    0.24
0.4    0.34
0.5    0.42
0.6    0.5
0.7    0.8125
0.78   1.0    /
```
OPM Flow terminates with:

```
System:  Oil-Water system.
Relative permeability input format: Saturation Family II.
Number of saturation regions: 1

Program threw an exception: Could not initialize the problem: TableContainer does not have any table in the range 0...0
Failed: 14905
OPM Flow Process 14905 Has Failed
Process Complete (0) 

```
However, using Saturation Family I tables:
```
--                                                                                     
--       WATER-OIL RELATIVE PERMEABILITY TABLES (SWOF)                                                                  
-- 
SWOF
--       SWAT       KRW       KROW      PCOW                                                     
--       FRAC                           PSIA                                                  
--       --------   --------  -------   -------                                                
         0.200000   0.0000    0.9000    0.000000
         0.238616   0.0002    0.7664    0.000000
         0.245309   0.0004    0.7443    0.000000
         0.261989   0.0010    0.6907    0.000000
         0.303091   0.0044    0.5671    0.000000
         0.368269   0.0191    0.3962    0.000000
         0.435026   0.0519    0.2528    0.000000
         0.486387   0.0940    0.1643    0.000000
         0.522283   0.1339    0.1137    0.000000
         0.550683   0.1725    0.0803    0.000000
         0.575342   0.2115    0.0559    0.000000
         0.599076   0.2542    0.0367    0.000000
         0.621294   0.2991    0.0223    0.000000
         0.642171   0.3458    0.0120    0.000000
         0.658984   0.3868    0.0061    0.000000
         0.671123   0.4183    0.0030    0.000000
         0.679268   0.4403    0.0015    0.000000
         0.684963   0.4562    0.0008    0.000000
         0.688893   0.4674    0.0004    0.000000
         0.692025   0.4765    0.0002    0.000000
[...]

PR #3706, opened 2021-11-22
Extend VFP calculations to two-phase flow.

hakonhagland:
Make VFP calculations work for the two-phase oil-water case. The plan is to address the issue with gas lift optimization in  #3668. 

Currently the VFP calculations requires three-phase input data, see the documentation for keyword `VFPPROD` and its implementation in [VFPProdProperties.cpp](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/VFPProdProperties.cpp). To extend the VFP calculations to the two-phase oil-water case, it might be enough to set the gas flow rate to a dummy value in VFPPROD record 5 (GFR values) and supplying a dummy input value for the gas rate to the methods in `VFPProdProperties.cpp`? 

This change does at least make the test case `GASLIFT-TST02` in #3668 run to the end. The next step is to see how the summary output compares with Eclipse and if we need to make any changes to the gas lift optimization code also.

### DIFF
--- a/opm/simulators/wells/GasLiftSingleWell_impl.hpp
+++ b/opm/simulators/wells/GasLiftSingleWell_impl.hpp
@@ -156,11 +156,7 @@ setupPhaseVariables_()
 {
     const auto& pu = this->phase_usage_;
     bool num_phases_ok = (pu.num_phases == 3);
-    if (pu.num_phases == 2
-        && pu.phase_used[BlackoilPhases::Aqua] == 1
-        && pu.phase_used[BlackoilPhases::Liquid] == 1
-        && pu.phase_used[BlackoilPhases::Vapour] == 0)
-    {
+    if (pu.num_phases == 2) {
         // NOTE: We support two-phase oil-water flow, by setting the gas flow rate
         //   to zero. This is done by initializing the potential vector to zero:
         //
@@ -171,7 +167,16 @@ setupPhaseVariables_()
         //  has been adapted to the two-phase oil-water case, see the comment
         //  in WellInterfaceGeneric.cpp for the method adaptRatesForVFP() for
         //  more information.
-        num_phases_ok = true;  // two-phase oil-water is also supported
+        if (    pu.phase_used[BlackoilPhases::Aqua] == 1
+             && pu.phase_used[BlackoilPhases::Liquid] == 1
+             && pu.phase_used[BlackoilPhases::Vapour] == 0)
+        {
+            num_phases_ok = true;  // two-phase oil-water is also supported
+        }
+        else {
+            throw std::logic_error("Two-phase gas lift optimization only supported"
+                                  " for oil and water");
+        }
     }
     assert(num_phases_ok);
     this->oil_pos_ = pu.phase_pos[Oil];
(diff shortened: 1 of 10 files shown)


hakonhagland:
jenkins build this please

hakonhagland:
jenkins build this please

OPMUSER:
@hakonhagland I working on the gas lift test cases with @totto82 and have several test cases for this issue, as well as others, which will be eventually uploaded. 

I'm just waiting on the E100 results for comparison, will keep you in the loop.

hakonhagland:
jenkins build this please

hakonhagland:
jenkins build this please

hakonhagland:
This combined with https://github.com/OPM/opm-common/pull/2966 allows case [GASLIFT-02](https://github.com/OPM/opm-tests/blob/master/gaslift/GASLIFT-02.DATA) to be run up to 01-Aug-2021, see [line 990](https://github.com/OPM/opm-tests/blob/master/gaslift/GASLIFT-02.DATA#L990) where gas lift optimization is turned off. 

At this point the simulator gets convergence problems
```
Well OP-A01 will be shut because it cannot get converged.
Well OP-A02 will be shut because it cannot get converged.
Well OP-B01 will be shut because it cannot get converged.
Well OP-B02 will be shut because it cannot get converged.
Well OP-C01 will be shut because it cannot get converged.
Well OP-C02 will be shut because it cannot get converged.
Problem:
Problematic well(s) were shut: OP-A01 OP-A02 OP-B01 OP-B02 OP-C01 OP-C02 (retrying timestep)
Starting time step 0, stepsize 0.0289908 days, at day 213/244, date = 01-Aug-2021
```
the timestep is reduced from 7.4 days to 0.03 days and it eventually gives up and closes all the production wells. 

Now, the wells should be put into production again later due to well testing but this does not happen. Also gaslift should be reactivated at 01-Nov-2021 see [line 1003](https://github.com/OPM/opm-tests/blob/master/gaslift/GASLIFT-02.DATA#L1003), but it does not happen since the wells are shut. After discussion with @totto82 we believe fixing this part is not related to the two-phase implementation in this PR, and will be a subject for a following PR instead.

Si this PR should now be ready for review and merging.

opm/simulators/wells/GasLiftSingleWell_impl.hpp
```diff
@@ -152,6 +149,36 @@ computeBhpAtThpLimit_(double alq) const
...
+        //     std::vector<double> potentials(NUM_PHASES, 0.0);
+        //
+        //  see e.g. runOptimizeLoop_() in GasLiftSingleWellGeneric.cpp
+        //  In addition the VFP calculations, e.g. to calculate BHP from THP
+        //  has been adapted to the two-phase oil-water case, see the comment
+        //  in WellInterfaceGeneric.cpp for the method adaptRatesForVFP() for
+        //  more information.
+        num_phases_ok = true;  // two-phase oil-water is also supported
+    }
+    assert(num_phases_ok);
```
totto82, line 176: I suggest a `throw` here that tells the user that we don't support oil-gas or water-gas

totto82 (changes requested):
Thanks for the contribution. This can be merged when the comments has been addressed.

hakonhagland:
jenkins build this please

totto82:
Thanks for the update.

PR #3711, opened 2021-11-25
fix co2store + aqunum

totto82:
Make it possible to use CO2STORE in combination with numerical aquifers.

### DIFF
--- a/opm/simulators/aquifers/AquiferNumerical.hpp
+++ b/opm/simulators/aquifers/AquiferNumerical.hpp
@@ -161,7 +161,7 @@ private:
         return ebos_simulator_.vanguard().eclState().runspec().co2Storage();
     }
 
-    inline bool phaseIdx_() const
+    inline int phaseIdx_() const
     {
         if(co2store_())
             return FluidSystem::oilPhaseIdx;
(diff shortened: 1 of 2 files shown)


totto82:
jenkins build this please

opm/simulators/aquifers/AquiferNumerical.hpp
```diff
@@ -158,6 +156,19 @@ class AquiferNumerical
     // TODO: maybe unordered_map can also do the work to save memory?
     std::vector<int> cell_to_aquifer_cell_idx_;
 
+    inline bool co2store_() const
+    {
+        return ebos_simulator_.vanguard().eclState().runspec().co2Storage();
+    }
+
+    inline bool phaseIdx_() const
```
bska, line 164: That's a suspicious return type if this is supposed to represent a phase index.

totto82: Good catch. Its a copy paste from AquiferInterface.hpp. i.e. we have a bug in master. I suppose it works by an accident since the phase indices are 0 and 1. I will fix both here and there.

totto82:
jenkins build this please

bska (approved):

PR #3713, opened 2021-11-26
Fix in WTEST

totto82:
If a well is shut due to physical or economical reason the wellstate status is shut
so we can not check the wellstate we instead check the well from the schedule to
make sure we don't test wells that are shut by the user.


totto82:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -384,16 +384,15 @@ namespace Opm {
...
-                if (ws.status != Well::Status::OPEN)
-                    continue;
 
                 WellInterfacePtr well = createWellForWellTest(well_name, timeStepIdx, deferred_logger);
                 // some preparation before the well can be used
                 well->init(&phase_usage_, depth_, gravity_, local_num_cells_, B_avg_);
                 const Well& wellEcl = schedule().getWell(well_name, timeStepIdx);
+                if (wellEcl.getStatus() == Well::Status::SHUT)
+                    continue;
+
```
joakim-hove, line 395: Two comments/questions:

1. I guess the check could be moved to the top of the body of the loop? 
2. I guess the check could be `wellEcl.getStatus() != Well::Status::OPEN)`

totto82:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
totto82: 1) done. 2) I think stopped wells should be tested

PR #3716, opened 2021-11-30
Move getPerfCellPressure to wellInterface

totto82:
The cell pressure is independent of well model and belongs to the interface.
This should move the MSW model one step closer to supporting GasWater cases.


totto82:
jenkins build this please

PR #3717, opened 2021-12-01
Suggestion: Remove enable switch for tracer support

joakim-hove:
The tracer support is now enabled by default - I don't see many people running flow as:
```
flow --enable-tracer=false CASE.DATA
```
I therefor suggest just removing the switch.


joakim-hove:
jenkins build this please

alfbr:
Tracer is still in the early development stages, so I do believe there is an interest in benchmarking performance implications. However, that can easily be done by commenting out tracer related keywords in the deck , so I support this removal.

PR #3720, opened 2021-12-03
Only check operable wells for wecon

totto82:
Only relevant is non default parameters are used
```
--shut-unsolvable-wells = true
```
or
```
--enable-well-operabilty-check-iter=true
```
that shuts wells during time-step


totto82:
jenkins build this please

totto82:
Glad we have jenkins to avoid stupid mistakes. I will update the PR with a better fix.

totto82:
jenkins build this please

PR #3721, opened 2021-12-03
Respect NEXTSTEP in ecl genericproblem timestepper

joakim-hove:


joakim-hove:
jenkins build this please

totto82:
The change will only work if enable-tuning=true is that intended?

OPMUSER:
@totto82 I for one don't think so. NEXTSTEP should be independent from TUNNING., so one should not have to  `enable-tuning=true` for NEXTSTEP to work. It is a common keyword to get around time step chops when large operational changes occur in the SCHEDULE section.

totto82:
Then the code needs to change since currently the change is within if( &&enableTuning_&& )

joakim-hove:
> The change will only work if enable-tuning=true is that intended?

Well I don't know I should call intentional; but I knew what I was doing at least ;-)

> NEXTSTEP should be independent from TUNNING., so one should not have to enable-tuning=true for NEXTSTEP to work. It is a common keyword to get around time step chops when large operational changes occur in the SCHEDULE section.

It is simple to change the implementation; I do not have strong opinions on this. The `TUNING` keyword is one where we have not followed ECL (unless explicitly enabled) - so we must ourselves determine what is the best/correct way to proceed. The `NEXTSTEP` and `TUNING` keywords interact in such a way that it makes a certain sense to couple both to the   `--enable-tuning` parameter - but sure; everything is possible.

This is more a political than a technical question; I will sit quite still and see what the jury concludes.

totto82:
I think it should be included inside the enableTuning as you propose since it is typically related to manual tuning. Come to think of it enableTuning is really the wrong name for the parameter. Its more about enableTuningFromDeck or something like that.

OPMUSER:
I still strongly suggest that we de-couple the -`-enable-tuning` option with NEXTSTEP. We don't do this for NUPCOL so we should not do this for NEXTSTEP.

Comments from other RE's appreciated.

@totto82 I don't think changing -`-enable-tuning` is worthwhile, but if you are going to this maybe -`-enable-tuning-keyw` may be a better fit.

PR #3725, opened 2021-12-07
Don't Log Any Messages on Non-IO MPI Ranks

bska:
This is a proof of concept to demonstrate one possible way of suppressing log messages on MPI ranks other than the IO rank (0).

Not intended for production.

### DIFF
--- a/opm/simulators/utils/readDeck.cpp
+++ b/opm/simulators/utils/readDeck.cpp
@@ -30,6 +30,7 @@
 
 #include <opm/common/ErrorMacros.hpp>
 #include <opm/common/OpmLog/EclipsePRTLog.hpp>
+#include <opm/common/OpmLog/LogBackend.hpp>
 #include <opm/common/OpmLog/OpmLog.hpp>
 #include <opm/common/utility/OpmInputError.hpp>
 #include <opm/common/utility/String.hpp>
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

opm/simulators/utils/readDeck.cpp
```diff
@@ -405,16 +419,27 @@ Opm::setupLogging(const int          mpi_rank_,
...
-        prtLog->setMessageFormatter(std::make_shared<Opm::SimpleMessageFormatter>(false));
+        if (mpi_rank_ == 0) {
+            std::shared_ptr<Opm::EclipsePRTLog> prtLog = std::make_shared<Opm::EclipsePRTLog>(logFileStream.str(), Opm::Log::NoDebugMessageTypes, false, output_cout_);
+            Opm::OpmLog::addBackend("ECLIPSEPRTLOG", prtLog);
+            prtLog->setMessageLimiter(std::make_shared<Opm::MessageLimiter>());
+            prtLog->setMessageFormatter(std::make_shared<Opm::SimpleMessageFormatter>(false));
+        }
+        else {
+            auto prtLog = std::make_shared<NullLog>(Opm::Log::NoDebugMessageTypes);
+            Opm::OpmLog::addBackend("ECLIPSEPRTLOG", std::move(prtLog));
```
blattms, line 430: Would have used `move` here as I would assume it only saves an increment. Would rather skip the variable if there are performance concerns.

bska: I appreciate the review, but just to be clear: This particular code will never be merged into the master branch.  It is only intended as a proof of concept.

bska:
Closing this now as the proof-of-concept has served its purpose.

PR #3732, opened 2021-12-09
Only include test if opencl is found

Tongdongq:
The `test_solvetransposed3x3.cpp` was always included, even when no OpenCL was found, causing a compilation error. Sorry for that.
I'm still testing this.

@bska @blattms

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -148,7 +148,6 @@ list (APPEND TEST_SOURCE_FILES
   tests/test_timer.cpp
   tests/test_invert.cpp
   tests/test_stoppedwells.cpp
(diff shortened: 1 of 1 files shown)


akva2:
i've tested the exact same fix locally just now, it works fine.

akva2:
jenkins build this please

akva2:
merging to restore build.

bska:
Thank you for the quick turn-around.  I confirm that this fixes the build [reported earlier](https://github.com/OPM/opm-simulators/pull/3672#issuecomment-989284645) on my system without OpenCL.

blattms:
Thanks. I guess I need to learn that there realy are no shortcuts when it comes to testing...

PR #3733, opened 2021-12-09
fix build with dune 2.6

akva2:
While 2.6 isn't really supported as such, this is a simple fix


akva2:
jenkins build this please

blattms:
> While 2.6 isn't really supported as such, this is a simple fix

Well, seems like someone is still using it. I am merging this, but I really think that we should officially not support 2.6 anymore.

akva2:
yeah, the problem is ubuntu focal is still on 2.6. if necessary i will backport in the ppa but i would like to avoid that as long as possible.

PR #3729, opened 2021-12-08
allow for gas-water + brine

goncalvesmachadoc:

PR #3728, opened 2021-12-08
Fix when both well and group control restrict gaslift

totto82:
The unmodified rates need to be passed to reduceALQtoGroupTarget(...) to get the right change in group rates.

PR #3735, opened 2021-12-09
Small clean up in GasLiftSingleWellGeneric.cpp

hakonhagland:
As discussed in PR #3728, I think it is better to move the two methods `reduceALQtoGroupTarget()` and `checkGroupTargetsViolated()` from `OptimizeState` to the parent class, then it is not necessary to abuse
`OptimizeState` in `maybeAdjustALQbeforeOptimizeLoop_()` just to call `reduceALQtoGroupTarget()`.

Also fixes a typo (as discussed in PR #3729) in `reduceALQtoGrouptTarget()` where the water rate is updated with the gas flow rate instead of the water flow rate. Should be like this:

```
water_rate = -potentials[this->parent.water_pos_];
```
instead of
```
water_rate = -potentials[this->parent.gas_pos_];
```

### DIFF
--- a/opm/simulators/wells/GasLiftSingleWellGeneric.cpp
+++ b/opm/simulators/wells/GasLiftSingleWellGeneric.cpp
@@ -222,6 +222,76 @@ checkALQequal_(double alq1, double alq2) const
 
 bool
 GasLiftSingleWellGeneric::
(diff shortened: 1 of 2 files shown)


hakonhagland:
jenkins build this please

totto82:
Thanks.

PR #3739, opened 2021-12-10
Take into account group limits in stage2

totto82:
Fix motivated by 4_GLIFT_MODEL5
This PR takes into account group limits in the gradients used to redistribute the gaslift (stage2)
Which again reduce oscillation in the solution. There is still some oscillation left that needs further investigation but this PR is a step in the right direction. 

image


totto82:
jenkins build this please

hakonhagland:
Thanks for this. I am starting on a cleanup of `GasLiftSingleGeneric.cpp`, after that the plan is to address this PR.

totto82:
This is made obsolete by https://github.com/OPM/opm-simulators/pull/3780

PR #3747, opened 2021-12-15
Consider all groups when reducing oil rate to group limits

hakonhagland:
This commit is a first step in a plan to fix up repetitive code in `GasLiftSingleWellGeneric.cpp`. More specifically: Note that 
the four methods [`getOilRateWithGroupLimit_()`](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/GasLiftSingleWellGeneric.cpp#L607), `getGasRateWithGroupLimit_()`, `getWaterRateWithGroupLimit_()`, and `getLiquidRateWithGroupLimit_()` all has essentially the same structure. The plan is to write a single general method that can be used instead, or possibly be wrapped by the four. 

This first step just fixes up one of the existing methods `getOilRateWithGroupLimit_()`, the plan is to use this method as the basis for deriving a general method that can be wrapped by the four methods mentioned above.

There is also a plan to address #3739 when this has been done. See aslo #3673.

In this first commit the main change is that all groups are considered when reducing oil rate to group limits. The previous code just checked the first group limit in the set. But there might be groups later in the set with more restrictive limits, which could have caused the oil rate to be reduced more than the first limit.

### DIFF
--- a/opm/simulators/wells/GasLiftSingleWellGeneric.cpp
+++ b/opm/simulators/wells/GasLiftSingleWellGeneric.cpp
@@ -607,22 +607,47 @@ GasLiftSingleWellGeneric::
 getOilRateWithGroupLimit_(const double new_oil_rate, const double oil_rate) const
 {
     const double delta_oil = new_oil_rate - oil_rate;
-    const auto &pairs =
-        this->group_info_.getWellGroups(this->well_name_);
-    for (const auto &[group_name, efficiency] : pairs) {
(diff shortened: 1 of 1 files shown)


hakonhagland:
jenkins build this please

totto82:
Thanks. From my side this looks ok. Any reasons why its WIP.

hakonhagland:
@totto82 Thanks. The idea was to build more commits on this, but I think I instead will submit those as separate PRs. I have removed the `WIP` tag, and have submitted a new PR that builds on this, see #3748

PR #3748, opened 2021-12-17
Refactor some methods in `GasLiftSingleWellGeneric.cpp`

hakonhagland:
Note this PR builds on #3747 which should be merged first.

Continues the work in #3747 to fix up repetitive code in `GasLiftSingleWellGeneric.cpp`.  In this PR `getOilRateWithGroupLimit_()`, `getGasRateWithGroupLimit_()`, `getWaterRateWithGroupLimit_()`, and `getLiquidRateWithGroupLimit_()` is refactored into a single generic method called `getRateWithGroupLimit_()`. 

Note that this change also includes a bug fix, see [this comment](https://github.com/OPM/opm-simulators/pull/3673/files#r771027732) and I guess this could be a reason to why the solution for the case [`4_GLIFT_MODEL5`](https://github.com/OPM/opm-tests/blob/master/model5/4_GLIFT_MODEL5.DATA) changes:

glift1

here `4_GLIFT_MODEL5_F1` is with the code in the master branch, `4_GLIFT_MODEL5_F2` (which is identical to the curve for `4_GLIFT_MODEL5_F1`) is with the code in PR #3747,  and `4_GLIFT_MODEL5` is with the current PR.

### DIFF
--- a/opm/simulators/wells/GasLiftSingleWellGeneric.hpp
+++ b/opm/simulators/wells/GasLiftSingleWellGeneric.hpp
@@ -57,6 +57,7 @@ protected:
 
 public:
     using GLiftSyncGroups = std::set<int>;
+    using Rate = GasLiftGroupInfo::Rate;
     struct GradInfo
     {
         GradInfo() { }
@@ -180,14 +181,20 @@ protected:
     std::pair<double, bool> getGasRateWithLimit_(const std::vector<double>& potentials) const;
     std::tuple<double,double,double,bool,bool,bool>
     getInitialRatesWithLimit_(const std::vector<double>& potentials);
+    std::tuple<double, const std::string*, double> getRateWithGroupLimit_(
+        Rate rate_type, const double new_rate, const double old_rate) const;
     std::pair<double, bool> getOilRateWithLimit_(const std::vector<double>& potentials) const;
     std::pair<double, bool> getWaterRateWithLimit_(const std::vector<double>& potentials) const;
 
-    std::pair<double, bool> getOilRateWithGroupLimit_(const double new_oil_rate, const double oil_rate) const;
-    std::pair<double, bool> getGasRateWithGroupLimit_(const double new_gas_rate, const double gas_rate) const;
-    std::pair<double, bool> getWaterRateWithGroupLimit_(const double new_water_rate, const double water_rate) const;
-    std::tuple<double,double,bool,bool> getLiquidRateWithGroupLimit_(double new_oil_rate, const double oil_rate,
-                                                                     double new_water_rate, const double water_rate) const;
+    std::pair<double, bool> getOilRateWithGroupLimit_(
+                           double new_oil_rate, double oil_rate) const;
+    std::pair<double, bool> getGasRateWithGroupLimit_(
+                           double new_gas_rate, double gas_rate) const;
+    std::pair<double, bool> getWaterRateWithGroupLimit_(
+                           double new_water_rate, double water_rate) const;
+    std::tuple<double,double,bool,bool> getLiquidRateWithGroupLimit_(
+                           double new_oil_rate, double oil_rate,
+                           double new_water_rate, double water_rate) const;
 
     std::tuple<double,double,bool,bool,double>
     increaseALQtoPositiveOilRate_(double alq,
--- a/opm/simulators/wells/GasLiftSingleWellGeneric.hpp
+++ b/opm/simulators/wells/GasLiftSingleWellGeneric.hpp
@@ -182,9 +182,11 @@ protected:
     std::tuple<double,double,double,bool,bool,bool>
     getInitialRatesWithLimit_(const std::vector<double>& potentials);
     std::tuple<double, const std::string*, double> getRateWithGroupLimit_(
-        Rate rate_type, const double new_rate, const double old_rate) const;
-    std::pair<double, bool> getOilRateWithLimit_(const std::vector<double>& potentials) const;
-    std::pair<double, bool> getWaterRateWithLimit_(const std::vector<double>& potentials) const;
+                  Rate rate_type, const double new_rate, const double old_rate) const;
+    std::pair<double, bool> getOilRateWithLimit_(
+                  const std::vector<double>& potentials) const;
+    std::pair<double, bool> getWaterRateWithLimit_(
+                  const std::vector<double>& potentials) const;
 
     std::pair<double, bool> getOilRateWithGroupLimit_(
                            double new_oil_rate, double oil_rate) const;
@@ -193,8 +195,8 @@ protected:
     std::pair<double, bool> getWaterRateWithGroupLimit_(
                            double new_water_rate, double water_rate) const;
     std::tuple<double,double,bool,bool> getLiquidRateWithGroupLimit_(
-                           double new_oil_rate, double oil_rate,
-                           double new_water_rate, double water_rate) const;
+                           const double new_oil_rate, const double oil_rate,
+                           const double new_water_rate, const double water_rate) const;
 
     std::tuple<double,double,bool,bool,double>
     increaseALQtoPositiveOilRate_(double alq,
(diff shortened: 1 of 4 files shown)


hakonhagland:
jenkins build this please

opm/simulators/wells/GasLiftSingleWellGeneric.hpp
```diff
@@ -180,14 +181,20 @@ class GasLiftSingleWellGeneric
...
-                                                                     double new_water_rate, const double water_rate) const;
+    std::pair<double, bool> getOilRateWithGroupLimit_(
+                           double new_oil_rate, double oil_rate) const;
+    std::pair<double, bool> getGasRateWithGroupLimit_(
+                           double new_gas_rate, double gas_rate) const;
+    std::pair<double, bool> getWaterRateWithGroupLimit_(
+                           double new_water_rate, double water_rate) const;
+    std::tuple<double,double,bool,bool> getLiquidRateWithGroupLimit_(
+                           double new_oil_rate, double oil_rate,
+                           double new_water_rate, double water_rate) const;
```
totto82, line 198: Why are they made nonconst

totto82:
I have tested this. This can be merged when the `const `comment is addressed. Note that results depends on tuning. I usually test with --enable-tuning=true. Then you notice that current master and this PR gives the same results. I.e. the code with the bug that is fixed in this PR is never activated.  Nevertheless, thanks for noticing and fixing a bug! I think we are soon on a level where 4_GLIFT_MODEL5 can be added to automatic testing. 

image

opm/simulators/wells/GasLiftSingleWellGeneric.hpp
hakonhagland: I could not see any reason to that these should be constant since they are not reference parameters. Why would you make them `const` here?

bska: > I could not see any reason why these should be constant since they are not reference parameters. Why would you make them `const` here?

On the other hand, if there's no reason to have mutable state&mdash;i.e., if you're actually changing the state&mdash;then then object should be declared `const` since that will generate build errors if you inadvertently change  something.

hakonhagland: > then the object should be declared const

@bska Yes I agree, but here we are passing simple types `double` which I would not call "object" and they are not passed by reference. On the other hand, if I had called `getLiquidRateWithGroupLimit_()` with a reference to an object of class `Foo` (for example) and I did not intend to modify the state of that object, I could declare the parameter as `const Foo& foo` to notify my intent to the compiler and humans reading the code. The compiler would then as a debugging aid check if I kept my promise.

blattms: @hakonhagland the reasoning of @baska is not about preventing copies, but about preventing developers introducing bugs, see  [Cpp Core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-const). But I have to admit that I usually do not adhere to this to (too much typing). I guess I need to step my game...

blattms (commented):
:+1: Cool. Really appreciate this kind of refactoring. Makes the code a lot cleaner,

bska:
> the reasoning [...] is not about preventing copies, but about preventing developers introducing bugs

That is precisely it.  Built-in arithmetic types `X` (e.g., `int` or `double`) should indeed be passed by value in which case `const X` behaves exactly the same as `X` from the point of view of the caller.  There is however a big difference in how those parameter types behave in the **body** of the function.  The protection awarded by `const` pays dividends there.

hakonhagland:
> The protection awarded by const pays dividends there.

@bska But by defining it `const` you cannot modify the variable. I did not see any reason to prevent that in this case, and we do actually modify the input oil rates (for example) and return the updated value back to the caller. I thought it might be confusing to define these double values as `const`, because then another reading the code later might wonder "why can I not modify these rate variables ? It seems counterproductive that I need to first copy them to a new local variable before changing them."

atgeirr:
> I did not see any reason to prevent that in this case

To simplify for the next reader. If I see "const double x" I can rest assured this will not be modified. Otherwise, there may be some clever code that exploits modification of x to do its work, and clever code is better avoided, unless it provides great performance benefits of some kind.

hakonhagland:
> there may be some clever code that exploits modification of x to do its work

@atgeirr I do not see that as "clever".The double parameters act similarly as any other locally defined variable. For example
```
void foo() {
     const double bar;

    // some code
}
```
I see that as the same as (regarding the scope and life time of `bar`)
```
void foo(const double bar) {

    // some code
}
```
So if we need to modify `bar` we do not use `const` and the code is not "clever" in my opinion.

atgeirr:
> So if we need to modify `bar` we do not use `const` and the code is not "clever" in my opinion.

If `bar` attains different values during the function execution, that has higher complexity than if it has a constant value. Avoiding it is good, and making it `const` communicates it clearly and ensures you do not change `bar` by mistake. If you really need to, make it non-const, sure. But all else being the same, having it const lowers complexity and therefore improves quality.

hakonhagland:
@atgeirr Ok I think I see your point now. Thanks for clarifying. So it depends on the complexity of the function whether one should make the parameter `const` and then copy it to another local variable before modifying it, or if one should just make it non-constant (for example if the function is just a couple of lines long the complexity is small). For this specific PR, the function is more than a couple of lines long, so we might declare the parameter as `const` to decrease complexity (and increase readability)?

blattms:
> The protection awarded by const pays dividends there.

Maybe, maybe not. Personally, I think we should not put too much effort in discussing such minor issues. They seem a bit controversial (c++ headers of gcc seem not to use const in these cases), and a bit a  matter of "out of principal...". One side has to bend here for progress.

This PR is a great cleanup and should be merged (once const functions are const). If somebody insists on const for function parameters: Hey this is open source, you can still create a PR.

atgeirr:
> @atgeirr Ok I think I see your point now. Thanks for clarifying. So it depends on the complexity of the function whether one should make the parameter `const` and then copy it to another local variable before modifying it, or if one should just make it non-constant (for example if the function is just a couple of lines long the complexity is small). For this specific PR, the function is more than a couple of lines long, so we might declare the parameter as `const` to decrease complexity (and increase readability)?

That is how I see it, yes. Mind you that I have made general remarks here, I have not carefully read this particular PR and leave the judgement to you and the reviewers.

hakonhagland:
jenkins build this please

hakonhagland:
> This can be merged when the const comment is addressed.

@totto82 The parameters have now been changed to `const`, see last commit.

blattms:
I was just about to merge this, but then I checked what const we changed. That really made my day.

We have been nagging about adding const to passed by value function parameters, but totally neglected that there might be some functions added without being const that do not change the internal state (getRate(s), getTarget, rateToString...). But I might as well have missed something, or have wrong priorities.

Merging anyway as @totto82's concerns seem satisfied. @hakonhagland Cool if you could check whether I am right and fix in a follow PR.

hakonhagland:
> there might be some functions added without being const that do not change the internal state (getRate(s), getTarget, rateToString...).

@blattms Yes you are right! I have added a new PR to address this, see #3753

hakonhagland:
> That is how I see it, yes. 

@atgeirr That makes sense to me also. But after reading [this](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con1-by-default-make-objects-immutable) from the C++ Core Guidelines: 

> Function parameters passed by value are rarely mutated, but also rarely declared const. To avoid confusion and lots of false positives, don’t enforce this rule for function parameters.

I am starting to doubt that we should enforce `const` on function parameters..

atgeirr:
> I am starting to doubt that we should enforce `const` on function parameters..

We already discussed the reasons to do so, so I will not reiterate them. I do find the argument for making an exception weak, it is essentially because "it is rare to mutate it, and also rare to bother". Certainly it is just as important for the function arguments as for any other local variable with the same scope and duration (the whole function!), alternatively we could declare "never mutate function arguments" but that would eliminate some optimization opportunities. I just think const-ing by default is the better habit!

bska:
> I just think const-ing by default is the better habit!

Agreed.  I've been writing code according to "anything that **can** be `const` **should** be `const`" for several years now.

PR #3749, opened 2021-12-17
File reorganization: OpenCL kernels

ducbueno:
Hello everyone.

This PR adds a CMake script that embeds OpenCL kernels written as *.cl files. The objective with this is to give a better organization of the OpenCL kernels, which have been growing in number.

The CMake script currently has an option `DEBUG_KERNELS` which can be used to find errors in the OpenCL codes. However this option only works for Intel OpenCL (via the `ocloc` command). I believe this can generalize easily to work with NVidia.


opm/simulators/linalg/bda/clSources.cpp
```diff
@@ -0,0 +1,977 @@
+// This file is auto-generated. Do not edit!
```
blattms, line 1: Generated files should not be commited and be put into the build directory. Please remove

opencl-source-provider.cmake
```diff
@@ -0,0 +1,50 @@
...
+if(DEBUG_KERNELS)
+  set(DEBUG_DIR ${KERNELS_DIR}/.debug)
+
+  execute_process(
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${DEBUG_DIR}
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+  )
+endif()
+
+set(CL_SRC_FILE ${BDA_DIR}/clSources.cpp)
```
blattms, line 13: Please use `${PROJECT_BINARY_DIR}/${BDA_DIR}/clSources.cpp` to create this file in the build directory.

opm/simulators/linalg/bda/openclKernels.cpp
```diff
@@ -20,6 +20,8 @@
 #include <config.h>
 #include <cmath>
 #include <sstream>
+#include <fstream>
+#include <iostream>
```
blattms, line 24: Are your sure that you need these?

blattms (changes requested):
Nice. I really like this approach and the cleanup. Jus took a short glimpse (was just to tempting...)

Did you do this after the lastest opecl PR merges? If not you will need to rebase this shortly before a merge.

ducbueno:
This PR was made after the merges of the newer OpenCL PRs @blattms. I'll fix the things pointed in the comments.

opencl-source-provider.cmake
```diff
@@ -10,7 +10,7 @@ if(DEBUG_KERNELS)
   )
 endif()
 
-set(CL_SRC_FILE ${BDA_DIR}/clSources.cpp)
+set(CL_SRC_FILE ${PROJECT_SOURCE_DIR}/${BDA_DIR}/clSources.cpp)
```
blattms, line 13: Still not the build directory. Please use ${PROJECT_BINARY_DIR}.

ducbueno: Hello Markus. I'm not sure if I'm following what you want here. Are you saying I should put the `clSources.cpp` elsewhere? Where would that be?

ducbueno: I tried using (without thinking much, I confess) `${PROJECT_SOURCE_DIR}` as you suggested, but the value of `${PROJECT_SOURCE_DIR}` gets appended to the elements of the `MAIN_SOURCE_FILES` list, so the build process ends up looking for `clSources.cpp` in a directory that doesn't exist.

blattms: If you generate the file below `${PROJECT_BINARY_DIR}` then we need to append that source file with complete path to 
the variable opm_simulators_SOURCES. The convention in OPM is to do that (and the generation of the source file) in the macro sources_hook which then should look like:
```
macro (sources_hook)
  if(OPENCL_FOUND)
    include(opencl-source-provider)
    list(APPEND opm-simulators_SOURCES ${PROJECT_BINARY_DIR}/clSources.cpp)
  endif()
endmacro (sources_hook)
```
With this there will be no changes to CMakeLists_files.cmake in this PR.

Sorry for not being clear previously. (Admittedly, I always have to look this procedure up in the code).

ducbueno: Got it. Just fixed it.

blattms:
Thanks. Will test this locally. What kind of builds did you try?

blattms:

opencl-source-provider.cmake
```diff
@@ -0,0 +1,50 @@
...
+if(DEBUG_KERNELS)
+  set(DEBUG_DIR ${KERNELS_DIR}/.debug)
+
+  execute_process(
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${DEBUG_DIR}
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+  )
+endif()
+
+set(CL_SRC_FILE ${_leaf_dir_name}/clSources.cpp)
```
blattms, line 13: This looks like an internal variable that might change without prior heads up. Would it be possible to use an official one like `${PROJECT_BINARY_DIR}`?

blattms: Hopefully this will also fix the errors in jenkins:
```
CMake Error at opencl-source-provider.cmake:14 (file):
  file failed to open for writing (Permission denied):

    /clSources.cpp
  CMakeLists.txt:291 (include)
  mpi/install/share/opm/cmake/Modules/OpmLibMain.cmake:120 (sources_hook)
  CMakeLists.txt:306 (include)
```

blattms: reason might be deactivated SiblingSearch

ducbueno: Ah, I forgot to change this. As soon as I get home I'll fix it.

ducbueno: Fixed.

blattms:

blattms:
Thanks for fixing jenkins. Will test here.

I am a bit tempted to try the "Squash and merge" feature of the "Merge pull request" button, but feel a bit uneasy about it.
@ducbueno Any chance that you squash this branch to one commit with a good commit message? There is one merge commit in here, not sure if that turns out to be troublesome.

atgeirr:
> There is one merge commit in here, not sure if that turns out to be troublesome.

In my experience they are always troublesome, if you want to rebase it... I would not risk the squash and merge unless you are sure it will be ok. And I do not see a way to be sure without doing the squash anyway, so I guess that feature is not really for me!

ducbueno:
Hmm I tried rebasing but I didn't look as I expected... perhaps due to the merge commit... any ideas how I can fix this?

ducbueno:
Just realized I really messed up... already trying to solve it

blattms:
Nevermind, if it is too much trouble I will merge as is.

ducbueno:
I think I was able to squash every thing to a single commit. Apparently there are no conflicts with the master branch. If there is anything wrong or something else to be done to make things cleaner let me know.

atgeirr:
> Apparently there are no conflicts with the master branch.

This is not true, for the GasLiftGroupInfo.[ch]pp files. Also, this PR now seems to add (again) the recently added tests for split communicators, which should not be necessary (although git does not mark it as a conflict, perhaps because it just adds exactly the same change). Did you rebase upon the master branch, or perhaps squash the commits (including the merge) but not changing the base? In my experience, it is better to drop the merge commit, squash your own commits (harder if you have own commits both before and after merge commits), then rebase that on master (instead of dealing with a merge commit). Needless to say, make backup branches before playing around...!

Worst case, you can copy the files you have changed, check out a new branch based on master, and use diff to carefully copy back your actual changes, then commit that.

blattms:
Closing this in favor of #3754. Unfortunately, there is some stuff still missing there. Hope to add that. Let's discuss later at our meeting.

PR #3697, opened 2021-11-13
Add IncompleteSAI preconditioner to openclSolver

ducbueno:
This PR adds an implementation for a Block Incomplete Sparse Approximate preconditioner. It offers a way to find approximate inverses for the matrices L and U given by ILU0 and apply them as a preconditioner.

This current implementation was tested only in my system (OpenCL + Intel Integrated Graphics) and only works for SPE3 and SPE9 models, so there is still some investigation to be done.

Now that the PR is done, I'll do some further tests on different platforms.

PR #3750, opened 2021-12-18
Adds IncompleteSAI preconditioner to openclSolver

ducbueno:
Reopening of #3697 due to my mess there.


bska:
Please rebase this onto the current master sources to solve the merge conflicts.

blattms:
Happy new year :tada: Thanks for rebasing. Will try to look through this in the next few days.

blattms (commented):
First (?) review just from a glimpse (seen as ratio what I looked at versus total changes, not glimpse in terms of fast as it took quite some time). More to come.

In general (new year's resolutions?):
- would be good if we add some tests (at least for the mappings). 
- would help if you would use some more characters for some variable names where appropriate as it might save a lot of time for others.

blattms:
Turns out it was at least the second review. Some of the comments I made in a similar way the last time.

ducbueno:
Hello there @blattms. Happy new year! Thanks for the comments.

Yes, you have made practically the same comments on the previous PR. It took some time to make this refactor to solve the conflicts, but I will go through all of them and make the changes as soon as I can.

ducbueno:
@blattms I believe I addressed all of your observations. If there you find that there still is stuff to clarify let me know.

opm/simulators/linalg/bda/BILU0.hpp
```diff
@@ -115,11 +115,16 @@ class BILU0 : public Preconditioner<block_size>
         return rmat.get();
     }
 
-    void get_preconditioner_structure(std::vector<int>& pointers, std::vector<int>& indices, std::vector<int>& diagIndex_)
+    // void get_preconditioner_structure(std::vector<int>& pointers, std::vector<int>& indices, std::vector<int>& diagIndex_)
+    // {
+    //     pointers.assign(LUmat->rowPointers, LUmat->rowPointers + (Nb + 1));
+    //     indices.assign(LUmat->colIndices, LUmat->colIndices + nnzb);
+    //     diagIndex_ = diagIndex;
+    // }
```
blattms, line 123: I guess we should remove the commented out code?

blattms (commented):
Thanks. much better.

blattms:
Did you add a test for the sparse matrix pattern conversion?

ducbueno:
@blattms I didn't add the test yet, I got caught up in some parallel activities. I already started to take a look at the other test programs and I hope I'll be able to add the pattern conversion test today.

opm/simulators/linalg/bda/BILU0.hpp
ducbueno: Oops. Removed it already. I'll push soon.

blattms:
jenkins build this please

tests/test_csrToCscOffsetMap.cpp
```diff
@@ -0,0 +1,63 @@
...
+        if (!mfile) {
+            throw std::runtime_error("Could not read matrix file");
+        }
+        readMatrixMarket(matrix_transposed, mfile);
+    }
+
+    // has to make copy because the output of readMatrixMarket does not
+    // have contiguous non-zero values
+    Matrix matrix_copy(matrix);
+    Matrix matrix_transposed_copy(matrix);
```
blattms, line 41: that means that `matrix_copy` and matrix_transposed_copy` are indentical.

tests/test_csrToCscOffsetMap.cpp
```diff
@@ -0,0 +1,63 @@
...
+            colIndices.emplace_back(c.index());
+        }
+        rowPointers.emplace_back(colIndices.size());
+    }
+
+    map = Opm::Accelerator::buildCsrToCscOffsetMap(rowPointers, colIndices);
+
+    for (unsigned int i = 0; i < colIndices.size(); i++){
+        BOOST_CHECK_EQUAL(nnzValues[i], nnzValues_transposed[i]);
+    }
```
blattms, line 62: Sorry, for being dense.
If we test the map then shouldn't the map be used somewhere in the test?

The test succeeds because the transposed is not the transposed of the other matrix but the same.

blattms (commented):
I think there is a flaw in the test.

Might also be worth to use a smaller matrix. The test should be as meaningful.

tests/test_csrToCscOffsetMap.cpp
ducbueno: Fixed

tests/test_csrToCscOffsetMap.cpp
ducbueno: Fixed

ducbueno:
Sorry about the test. Forgot to commit some local changes.

blattms:
Thanks for the quick reaction.

Sorry for being annoying :wink::
There are now 8 commits here (and none of them has anything about ISAI in the commit). Would be really nice if
you would squash all of these commits into one (`git rebase -i 3e4e62bc4f`) and add a meaningful commit message  with what you are adding, status, what works.

blattms:
Cool,  much appreciated, thanks.

atgeirr:
We just discovered that this PR adds a routine to set up a parameter tree for the FlexibleSolver if --linsolver=isai is given on the command line. However, that routine only sets up the normal ILU0, and there is no ISAI preconditioner for the CPU in any case, so this is very misleading.

I will simply remove that code,  unless stopped.

blattms:
@ducbueno Can you please comment?

ducbueno:
I don't think it would be a problem to remove this code. At the time I thought I should set new entries in the PropertyTree even if the preconditioner is implemented only for GPU. And ISAI works on top of ILU, so setting the properties only for the latter should suffice for now.

atgeirr:
> And ISAI works on top of ILU,

I assume you mean that it is on top on the OpenCL ILU, not the dune-istl ILU, right?

ducbueno:
Yes. It works on top of the OpenCL ILU.

PR #3753, opened 2021-12-21
Made some methods in GasLiftGroupInfo const

hakonhagland:
This is in agreement with [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con2-by-default-make-member-functions-const): _A member function should be marked `const` unless it changes the object’s observable state. This gives a more precise statement of design intent, better readability, more errors caught by the compiler, and sometimes more optimization opportunities_.

### DIFF
--- a/opm/simulators/wells/GasLiftGroupInfo.cpp
+++ b/opm/simulators/wells/GasLiftGroupInfo.cpp
@@ -70,7 +70,7 @@ getGroupIdx(const std::string& group_name)
 
 double
 GasLiftGroupInfo::
-gasRate(const std::string& group_name)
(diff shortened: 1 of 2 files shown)


hakonhagland:
jenkins build this please

atgeirr:
Good! Usually having such members const would be more or less implicitly enforced by using the members functions with a const object, such as typically passed by const reference to some function. That the missing const qualifications did not cause compile errors could indicate that you are using mutable GasLiftGroupInfo objects in contexts where immutable(const) objects could be used?

PR #3754, opened 2021-12-22
Clean opencl kernels rebased (rebased #3749)

blattms:
Moves all OpenCL kernels to *.cl files and uses Cmake to combine.
    
Kernel files are located in opm/simulators/linalg/bda/opencl/kernels. CMake will combine them for usage in     ${PROJECT_BINARY_DIR}/clSources.cpp that becomes part of the library.

I renamed the debug variable to indicate that it is for intel,

@ducbueno Still had your branch lying around and just cherry-picked the relevant commits on top of master.

PR #3746, opened 2021-12-15
Fix stdwell_apply opencl kernel

Tongdongq:
This fixes the bug in the opencl kernel `stdwell_apply` and `stdwell_apply_no_reorder`.
Fixes https://github.com/OPM/opm-simulators/issues/3635.
Also removes unnecessary line in opencl and in cuda kernel.

```
Benchmarks for NORNE_ATW2013.DATA
Notation: A (B+C+D+E), F, G, I | J
A: Total time
B: Assembly time
C: Linear solve time
D: Update time
E: Pre/post step
F: Overall Linearizations
G: Overall New Iterations
I: Overall Linear Iterations
J: Number of times openclSolver failed to converge and Dune was used instead.
LS: level scheduling
GC: graph coloring

R5 2400G, NVIDIA GTX1050Ti:
--matrix-add-well-contributions=true, clear CPU:
LS: 620 (136+378+ 78+109), 1814, 1453, 21656 | 0
GC: 755 (170+345+103+117), 2384, 1996, 46638 | 0
--matrix-add-well-contributions=false, clear CPU:
LS: 656 (134+316+ 77+110), 1810, 1447, 20797 | 0
GC: 771 (172+359+101+118), 2309, 1920, 43975 | 0
```

PR #3755, opened 2021-12-22
Fix opencl apply stdwells rebased (replaces #3746)

blattms:
Took @Tongdongq's PR #3746 and adjusted it to the new opencl kernels structure (see #3754 )


blattms:
jenkins build this please

blattms:
@Tongdongq Would you double that I replicated your fixes here, please. I think I did but running Norne without  adding the well contributions stil results in rather small timesteps. Just try `flow --accelerator-mode=opencl --linsolver=cpr_quasiimpes  NORNE_ATW2013` and compare without accelerator.

Tongdongq:
I cannot reproduce the smaller than 1 days timesteps. With CPR, I get 379 timesteps for Dune, and 349 for opencl. With the masters of 2022-1-3.
I did not count the number of timesteps when using just iliu0, but the statistics at the end looked normal.

blattms:
Man, how hard can it be. :flushed: must have used flow compiled with the wrong branch. Indeed this works as advertised.  Good work and sorry for the noise.

PR #3758, opened 2022-01-07
Handle case where no completions have ratio above the limit while the well has

totto82:
The main objective of this PR is to avoid the asserting on some corner cases where the well rates gives ratios that are above the limit while none of the completions does. The only case where I have seen this is when a problematic well is opened up due to well testing and starts to produce with very very small rates. Ratios between very small numbers are sensitive to numerical noise.

### DIFF
--- a/opm/simulators/wells/WellInterfaceFluidSystem.cpp
+++ b/opm/simulators/wells/WellInterfaceFluidSystem.cpp
@@ -635,8 +635,20 @@ checkRatioEconLimits(const WellEconProductionLimits& econ_production_limits,
     }
 
     if (report.ratio_limit_violated) {
-        assert(report.worst_offending_completion != INVALIDCOMPLETION);
-        assert(report.violation_extent > 1.);
+        // No worst offending completion is found because all the completions are either injecting or
+        // have trivial rates.
+        if(report.worst_offending_completion == INVALIDCOMPLETION) {
+            std::string message = "The well ratio limit is violated but all the completion rates are trivial! " + this->name() + " is kept open";
+            deferred_logger.warning("WECON_INVALIDCOMPLETION", message);
+            report.ratio_limit_violated = false;
+        }
+        // Due to numerical instability there may exist corner cases where the well breaks
+        // the ratio limit but no completion does.
+        else if(report.violation_extent <= 1.) {
+            std::string message = "The well ratio limit is violated but no completion ratio limit is violated! " + this->name() + " is kept open";
+            deferred_logger.warning("WECON_INCONSISTANT_COMPLETION_WELL", message);
+            report.ratio_limit_violated = false;
+        }
     }
 }
 
@@ -836,10 +848,7 @@ checkMaxRatioLimitCompletions(const SingleWellState& ws,
         }
     } // end of for (const auto& completion : completions_)
 
-    assert(max_ratio_completion > max_ratio_limit);
-    assert(worst_offending_completion != INVALIDCOMPLETION);
     const double violation_extent = max_ratio_completion / max_ratio_limit;
-    assert(violation_extent > 1.0);
 
     if (violation_extent > report.violation_extent) {
         report.worst_offending_completion = worst_offending_completion;


totto82:
jenkins build this please

GitPaean:
Because it is a design problem, can you please just confirm this behavoir is in line with the reference behavior we want to match?

totto82:
> Because it is a design problem, can you please just confirm this behavoir is in line with the reference behavior we want to match?

I am not sure what you mean by  "reference behavior" here, but what I would argue is that the simulator should not assert due to minor numerical noise but continue here. The design follows the logic that if well and completion disagree the least intrusive measure is taken i.e. not shut the well. (It may be shut due to small rates, but that is another test)

GitPaean:
> I am not sure what you mean by "reference behavior" here,

Sorry for assuming this PR is for recovering the behavoir for a test case with reference simulator. 

I will comment on the code directly.

opm/simulators/wells/WellInterfaceFluidSystem.cpp
```diff
@@ -836,6 +836,17 @@ checkMaxRatioLimitCompletions(const SingleWellState& ws,
         }
     } // end of for (const auto& completion : completions_)
 
+    // The well got a ratio above the max_ratio, but none of the completions did
+    // This can happen for rare cases with very small rates.
+    // Note that the well rates are adjusted after they are accumulated up for convergence reasons
+    // and are therefore not exactly the sum of all its connection rates.
+    // For these cases we reset the violated flag for the well and continue instead of aborting
+    // due to missing worst_offending_completion
+    if (max_ratio_completion == 0) {
```
GitPaean, line 845: I might read the code wrong. 

With the code above, 
``` c++
        if (ratio_completion > max_ratio_completion) {
            worst_offending_completion = completion.first;
            max_ratio_completion = ratio_completion;
        }
```

We should always find some completion here, right. Unless there is no completions for this well at all?
Under what circumstances, `max_ratio_completion == 0)`?

totto82: I tried to explain it in the comment in the code:
```
   // The well got a ratio above the max_ratio, but none of the completions did
    // This can happen for rare cases with very small rates.
    // Note that the well rates are adjusted after they are accumulated up for convergence reasons
    // and are therefore not exactly the sum of all its connection rates.
    // For these cases we reset the violated flag for the well and continue instead of aborting
    // due to missing worst_offending_completion
```

So in a perfect world we should always find a completion but in the real world it may not due to minor differences between the sum of the perforation rates and the well rate.

GitPaean: Excuse me if am too slow about here, 
`max_ratio_completion` is initialized as `0`. 

```C++
double max_ratio_completion = 0;
```

If we go through all the completions with the code below, 
```c++
        if (ratio_completion > max_ratio_completion) {
            worst_offending_completion = completion.first;
            max_ratio_completion = ratio_completion;
        }
```

If at the end `(max_ratio_completion == 0) `, 
the only way I can think about is that for all the completions, `ratio_completion <=0` or there is no completion for this well. 

Is that  (for all the completions, `ratio_completion <=0`) the situation you are looking at?

totto82: Yes. That is correct. In particular `ratio_completion `will be zero for completions where flow is flowing out from not into the producer. In the particular case I am debugging the well has an epsilon positive production rate with a water cut above the limit (0.95) while the all the completion (only one in this case) is injecting and hence return ratio_completion =0.

totto82:
Can this be merged now?

GitPaean:
> Can this be merged now?

Busy with other things. Did not come back earlier. Will come with a suggestion soon.

opm/simulators/wells/WellInterfaceFluidSystem.cpp
GitPaean: This is my suggestion. 

Use the following if statement instead, 
```c++
if (worst_offending_completion == INVALIDCOMPLETION) 
```
Then add the comment to say that `no worst offending completion is found because all the completions are injecting` due to the reason you mentioned above. 

In my opinion, this if statement can be easier to understand, and also we say clearly that the reason is because all the completions are injecting for this producer.

totto82:
jenkins build this please

opm/simulators/wells/WellInterfaceFluidSystem.cpp
```diff
@@ -836,10 +836,21 @@ checkMaxRatioLimitCompletions(const SingleWellState& ws,
...
+    // No worst offending completion is found because all the completions are either injecting or
+    // have trivial rates.
+    if (worst_offending_completion == INVALIDCOMPLETION) {
+        report.ratio_limit_violated = false;
+        return;
+    }
+
+    // Due to numerical instability there may exist corner cases where the well breaks
+    // the ratio limit but no completion does for wells with very small rates.
+    if (max_ratio_completion <= max_ratio_limit) {
```
GitPaean, line 848: Thanks for addressing the comments.  

This line is also a special situation we do not see normally, that noisy small values make the general assumption does not apply anymore. It is a little different from the situation that with the only completion is injecting. 

And changing the value `report.ratio_limit_violated` at the this stage is pretty dramatic. It can be good to give some message `debug` or even` warning` to let people know that something extraordinary going on at these two occasions. 

I have no more further comments.

totto82:
jenkins build this please

opm/simulators/wells/WellInterfaceFluidSystem.cpp
totto82: I have added a warning as you suggested. We may want to move it to debug if this happens more often than we think. Thanks for the review.

totto82:
> I have no more further comments.

I interpret this as an approval and will merge as all checks have passed

PR #3765, opened 2022-01-10
Accumulate Dissolved Gas and Vaporized Oil Flow Rates Across Ranks

bska:
This is needed to get consistent estimates for the summary vectors

  * `{F,G,W}OP{R,T}{F,S}` &mdash; Free/Vaporized Oil Production Rate/Cumulative Volume
  * `{F,G,W}GP{R,T}{F,S}` &mdash; Free/Dissolved Gas Production Rate/Cumulative Volume

in the case of distributed wells.

Thanks to [at]blattms for the suggested fix.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -467,6 +467,14 @@ namespace Opm
         // Update the connection
         this->connectionRates_ = connectionRates;
 
+        // Accumulate dissolved gas and vaporized oil flow rates across all
+        // ranks sharing this well (this->index_of_well_).
+        {
+            const auto& comm = this->parallel_well_info_.communication();
+            ws.dissolved_gas_rate = comm.sum(ws.dissolved_gas_rate);
+            ws.vaporized_oil_rate = comm.sum(ws.vaporized_oil_rate);
+        }
+
         // accumulate resWell_ and invDuneD_ in parallel to get effects of all perforations (might be distributed)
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska:
With this change we get consistent summary vectors for the `flow_distribute_z` binary applied to [`spe9/SPE9_CP_SHORT`](https://github.com/OPM/opm-tests/blob/f701d41f8962d2155116a1dc1dc63d3c9ff7abb0/spe9/SPE9_CP_SHORT.DATA).  Below are a couple of example plots as a demonstration of the effect.  With this PR, the curves coincide with a sequential run while the current master branch underestimates the dissolved gas production rate in the case of four MPI processes.  Note that the summary output code calculates the ["free" production rate](https://github.com/OPM/opm-common/blob/f9ba16d72ea08e3c01aa18d7e9201dc31c1e6a27/src/opm/output/eclipse/Summary.cpp#L1666) as
```
free = total - dissolved
```
so if we underestimate the "dissolved" rate then we will overestimate the "free" rate if the "total" rate matches.

Moreover, this PR is sufficient to fix the issue that blocks OPM/opm-common#2871.

---

Quantity | Comparison
----- | -----
Field-Level Dissolved Gas Production Rate | FGPRS
Group-Level Free Gas Production Rate | GGPRF_P
Well-Level Free Gas Cumulative Production | WGPTF_PRODU17

blattms:
Thanks a lot for the fix. In it goes.

PR #3767, opened 2022-01-11
Include config.h in clSources.cpp to include correct openCL header.

blattms:
Otherwise the definitions determined by CMake are not available to opm/simulators/linalg/bda/opencl.hpp and choosing the correct header might fail.

On my system this at least fixes a warning.

Fallout from #3749


blattms:
jenkins build this please

blattms:
Shameless selfmerge

PR #3773, opened 2022-01-18
return zero thp value for extremely small injection rate

GitPaean:
it is a finding from one of the test cases. We do not have a situation to verify about the producers, so we do not include producers in this PR although probably should be the same. 

Basically, this is an injector with zero GRAT control. The reference results output zero THP values regardless the BHP values of the well.

### DIFF
--- a/opm/simulators/wells/VFPInjProperties.cpp
+++ b/opm/simulators/wells/VFPInjProperties.cpp
@@ -31,6 +31,8 @@
 
 #include <opm/simulators/wells/VFPHelpers.hpp>
 
+#include <limits>
+
 namespace Opm {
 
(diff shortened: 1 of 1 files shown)


GitPaean:
jenkins build this please

GitPaean:
The benefit for this is to reduce the confusion when checking the well curve comparison for THP controlled wells, although maybe only very few occasion.

atgeirr:
I think this should be merged.

I will wait a little for @tskille to weigh in on the question of what is the correct value for zero flow first, if he wants to.

atgeirr:
After discussion with @tskille we decide to merge this. It is also a very small change.

PR #3774, opened 2022-01-18
allow for explicit initialization of SALT and SALTP

plgbrts:
This PR depends on https://github.com/OPM/opm-common/pull/2931

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -2575,10 +2575,10 @@ private:
                                      " vaporized oil is enabled");
         if (enableBrine && !has_salt)
             throw std::runtime_error("The ECL input file requires the SALT keyword to be present if"
-                                     " brine is enabled");
+                                     " brine is enabled and the model is initialized explicitly");
         if (enableSaltPrecipitation && !has_saltp)
             throw std::runtime_error("The ECL input file requires the SALTP keyword to be present if"
-                                     " salt precipitation is enabled");
+                                     " salt precipitation is enabled and the model is initialized explicitly");
 
         size_t numDof = this->model().numGridDof();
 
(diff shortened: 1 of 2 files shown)


ebos/eclproblem.hh
```diff
@@ -2571,6 +2573,12 @@ private:
         if (FluidSystem::enableVaporizedOil() && !has_rv)
             throw std::runtime_error("The ECL input file requires the RV keyword to be present if"
                                      " vaporized oil is enabled");
+        if (enableBrine && !has_salt)
+            throw std::runtime_error("The ECL input file requires the SALT keyword to be present if"
+                                     " brine is enabled");
```
blattms, line 2578: Isn't there a change of semantics here?

It seems that now an error is thrown if you simulate brine and did not explicitly specify SALT (`has_salt==false`). I assume that previously the default would have been used.

blattms (commented):
As I am far from being an expert for eclipse file, I might be wrong and the change is the right thing, but

ebos/eclproblem.hh
plgbrts: The SALT keyword is only required if brine is enabled _and_ the model is initialized  explicitly. The modified part of the code deals with the explicit initialization.  I have improved the error message.

totto82:
jenkins build this please

blattms:
Thanks @goncalvesmachadoc. Will do the update_data drill (quite uncommon for me).

blattms:

blattms:
jenkins build this with opm-tests=3774

akva2:
yes look at your trigger...

akva2:
jenkins build this opm-tests=3774 please

akva2:
argh, and i copied your second screw-up. let's try again.

akva2:
jenkins build this opm-tests=727 please

akva2:
yes, missing please. we parse between jenkins *uild this and please. and you referenced the opm-simulators pr number, not the opm-tests one. the with would have been harmless.

blattms:
:flushed: Oh my...

PR #3751, opened 2021-12-19
Refactor some methods in `GasLiftSingleWellGeneric.cpp` (part 2)

hakonhagland:
Note this PR includes #3748 which should be merged first. 

Continues the work in #3748 of cleaning up some methods in `GasLiftSingleWellGeneric.cpp` to get rid of repetitive code and make it easier to maintain the code in the future.
    
This last commit refactors `getOilRateWithLimit_()`, `getGasRateWithLimit_()`, and  `getWaterRateWithLimit_()`. The common part of the methods is split out into a new method called `getRateWithLimit_()`.

I have tested this PR with the case [`4_GLIFT_MODEL5`](https://github.com/OPM/opm-tests/blob/master/model5/4_GLIFT_MODEL5.DATA) and there was no difference in the solution for FGLIR (field gas lift injection rate) compared to the solution for #3748 which indicates that the numerical solution is not affected by this commit.

PR #3780, opened 2022-01-20
Refactor code in GasLiftSingleWellGeneric.cpp (part 3)

hakonhagland:
Continues the work in #3751 of cleaning up code in `GasLiftSingleWellGeneric.cpp` to get rid of repetitive code and make it easier to maintain the code in the future.

Introduces two new data types `BasicRates` and `LimitedRates` to capture oil, gas, and water rates, and whether they have been limited by well or group targets. This reduces the number of variables that are passed to and returned from various methods and thus makes the code easier to read.

I tested this PR with the case [`4_GLIFT_MODEL5`](https://github.com/OPM/opm-tests/blob/master/model5/4_GLIFT_MODEL5.DATA) and there was no difference in the solution for GGLIR (group gas lift injection rate for group `PLAT-A`) compared to the solution with the code in the master branch, indicating that the numerical solution is not affected for this test case. However, I expect the solution could be different for other cases since this PR includes some minor bug fixes like [this](https://github.com/OPM/opm-simulators/pull/3407#commitcomment-62833020).

### DIFF
--- a/opm/simulators/wells/GasLiftGroupInfo.cpp
+++ b/opm/simulators/wells/GasLiftGroupInfo.cpp
@@ -102,7 +102,6 @@ getRate(Rate rate_type, const std::string& group_name) const
     }
 }
 
-
 std::tuple<double, double, double, double>
 GasLiftGroupInfo::
 getRates(const int group_idx) const
@@ -158,6 +157,14 @@ groupIdxToName(int group_idx) const
     //   cannot be nullptr here..
     assert(group_name);
     return *group_name;
+}
+
+bool
+GasLiftGroupInfo::
+hasAnyTarget(const std::string& group_name) const
+{
+    return oilTarget(group_name)   || gasTarget(group_name)
+        || waterTarget(group_name) || liquidTarget(group_name);
 }
 
 bool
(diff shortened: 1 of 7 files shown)


hakonhagland:
jenkins build this please

totto82:
Thanks for the massive cleanup. Removing the duplicated oil/gas/water code indeed improves the readability and maintainability of the code significantly. I did look through the code, and didn't find anything really suspicious, but since this cleanup is massive I think it would be good to also do a sanity check on some of newly made tests under opm-tests/gaslift. Could you try to run GASLIFT_05 to GASLIFT_08 with master and compare it with this PR?

hakonhagland:
jenkins build this please

hakonhagland:
rebased this on master

hakonhagland:
jenkins build this please

hakonhagland:
> Could you try to run GASLIFT_05 to GASLIFT_08 with master and compare it with this PR?

@totto82 Good idea. The first case `GASLIFT_05` shows an identical behavior with master for `GGLIR` for `PLAT-A`. Here `GASLIFT-05-M` is the results run with the master branch, and the blue curve `GASLIFT-05` for the current branch is covered by the red curve. :

glift1

however the second case, `GASLIFT_08` shows a completely different behavior for `GGLIR` : 

glift2

here `GASLIFT-08-M2` is the result with the master branch.

totto82:
Thanks for testing. I will do some testing myself.

totto82:
I have tested this PR on several models and my conclusion is that the differences we notice between master and this PR for GASLIFT-08 is only indirectly related to this PR.  Since the test models exposed to gaslift that we have included in our test-suite looks good and tests on a field model that uses gaslift looks good, I will merge this now to ease further development.

PR #3781, opened 2022-01-20
fixing the logic in function updateWellOperability

GitPaean:
The existing logic is to check only when the wells are under THP/BHP control, but THP/BHP are limits/constraints that have to be met for any operable wells. 

It means for Rate controlled wells, the operability checking basically do not check the THP/BHP constraints with the existing code, which needs to be corrected. 

It might take big testing efforts before it can go in.

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -586,19 +586,18 @@ namespace Opm
     {
         this->operability_status_.resetOperability();
 
-        bool thp_controled = this->isInjector() ? well_state.well(this->index_of_well_).injection_cmode == Well::InjectorCMode::THP:
-                                              well_state.well(this->index_of_well_).production_cmode == Well::ProducerCMode::THP;
-        bool bhp_controled = this->isInjector() ? well_state.well(this->index_of_well_).injection_cmode == Well::InjectorCMode::BHP:
-                                              well_state.well(this->index_of_well_).production_cmode == Well::ProducerCMode::BHP;
+        const auto& summary_state = ebos_simulator.vanguard().summaryState();
+        const bool has_thp_limit = this->isInjector() ?
+                                   this->well_ecl_.injectionControls(summary_state).hasControl(Well::InjectorCMode::THP) :
+                                   this->well_ecl_.productionControls(summary_state).hasControl(Well::ProducerCMode::THP);
 
-        // Operability checking is not free
-        // Only check wells under BHP and THP control
-        if(bhp_controled || thp_controled) {
-            updateIPR(ebos_simulator, deferred_logger);
-            checkOperabilityUnderBHPLimit(well_state, ebos_simulator, deferred_logger);
-        }
-        // we do some extra checking for wells under THP control.
-        if (thp_controled) {
+
+        // always check the BHP limit, which has to be honored for a operable well
+        updateIPR(ebos_simulator, deferred_logger);
+        checkOperabilityUnderBHPLimit(well_state, ebos_simulator, deferred_logger);
+
+        // we do some extra checking for wells has a THP limit
+        if (has_thp_limit) {
             checkOperabilityUnderTHPLimit(ebos_simulator, well_state, deferred_logger);
         }
     }


GitPaean:
Although still marked as draft because it is not tested yet, while it is good the people familiar with the matters can join the discussion now. @totto82 @atgeirr

GitPaean:
jenkins build this please

GitPaean:
Only one test case failed. Will check and report back.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -586,19 +586,18 @@ namespace Opm
...
-
-        // Operability checking is not free
-        // Only check wells under BHP and THP control
-        if(bhp_controled || thp_controled) {
-            updateIPR(ebos_simulator, deferred_logger);
-            checkOperabilityUnderBHPLimit(well_state, ebos_simulator, deferred_logger);
-        }
-        // we do some extra checking for wells under THP control.
-        if (thp_controled) {
+        const auto& summary_state = ebos_simulator.vanguard().summaryState();
```
atgeirr, line 589: Nitpick: this and the next can be moved down to line 599.

atgeirr (commented):
I agree this looks like a necessary change. Let's see how testing fares, if very costly we may look for other ways to reduce cost.

atgeirr:
benchmark please

opm/simulators/wells/WellInterface_impl.hpp
GitPaean: Yes. thanks for spotting.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -586,19 +586,18 @@ namespace Opm
...
+        const bool has_thp_limit = this->isInjector() ?
+                                   this->well_ecl_.injectionControls(summary_state).hasControl(Well::InjectorCMode::THP) :
```
GitPaean, line 591: It looks like we can use the function `wellHasTHPConstraints(const SummaryState& summaryState) const` for this purpose. Anyway, will change it after the benchmark test.

totto82:
> I agree this looks like a necessary change. Let's see how testing fares, if very costly we may look for other ways to reduce cost.

Does this change impact cases where well control is oscillating and the well ends up with a control other than thp even if it breaks the thp control. Could it be better to fix this by forcing the oscillating well to chose the thp/bhp limit. The original logic assumed that first a well reach its thp/bhp limit and then when it can no longer produce/inject under thp/bhp controls it shut down. With this all wells are tested all the time which ads extra simulation time.

GitPaean:
>  Could it be better to fix this by forcing the oscillating well to chose the thp/bhp limit. 

Is it guaranteed that the well will stay with the thp/bhp control  when entering the well testing?

totto82:
> Is it guaranteed that the well will stay with the thp/bhp control when entering the well testing?

Currently no. But maybe it make sense to force it to switch to the thp/bhp control if it oscillates between thp/bhp and grup/orat/etc.

GitPaean:
> Currently no. But maybe it make sense to force it to switch to the thp/bhp control if it oscillates between thp/bhp and grup/orat/etc.

I have some prototyping similar to that (if the pressure control (thp/bhp) can not get converged, do not jump to other RATE control) for the case I am debugging with and the result is improving.  The current efforts is trying to avoid re-opening the wells through well testing. When it re-opens, it takes a lot of efforts to shut it down again, especially if we keep re-opening it. 

But I really think these types of strategy really makes the well logic complicated. (but maybe not that bad)

The current PR will have problem if the operability updating code is not well established.

GitPaean:
Operability test part is very difficult to establish well. Maybe for the well testing, we should just try to solve under the BHP limit and also THP limit (if it exists) to see whether it is solvable to decide the operability.

totto82:
Lets see what `benchmark `says. If the overhead is small I suggest we merge this after the minor cleanups suggested by @atgeirr

atgeirr:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 0.993    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 0.992    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 0.992    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.972    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 1.001    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1491

GitPaean:
> Lets see what benchmark says. If the overhead is small I suggest we merge this after the minor cleanups suggested by @atgeirr

The key is still the results should not be worse. The thing I am worrying most that it shut wells that can be solved.

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 0.99     |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 0.996    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 1.003    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.978    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 1.019    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1492

totto82:
The performance penalty seems small so I suggest we merge this if @GitPaean agrees. The test failure is a bit suspicious, and should be double checked first.

GitPaean:
> The performance penalty seems small so I suggest we merge this if @GitPaean agrees. The test failure is a bit suspicious, and should be double checked first.

performance is one thing, we need to check whether it shuts wells unnecessarily. I think it is important to keep the wells open if they can solved and converged.

tskille:
I have started some testing locally on some full field models. I will report back when this simulations are finished

GitPaean:
> I have started some testing locally on some full field models. I will report back when this simulations are finished

@tskille thanks for doing that. Looking forward to your report back. 

My current testing shows it will shut some wells important. 

My current thinking, since the IPR based operability test difficult to make nice, we might evaluate whether we need it at all.  We will discuss.

tskille:
I have done some testing with this PR and there are some models that we need to look into before we can merge this PR. @GitPaean I will share these findings and the full field models with you with on email/one-drive.

GitPaean:
The logic fixing in this PR is valid, while due to not-perfect IPR calculation or poor initialization of the well values for multisegment wells, with this way, flow might close wells supposed to be open. We need to look for other way to tackle the problems we are tackling.

PR #3786, opened 2022-01-23
Small improvement of debugging tools in gaslift code

hakonhagland:
Introduces a gaslift debugging variable in `ALQState` in `WellState`. This variable will persist between timesteps in contrast to when debugging variables are defined in `GasLiftSingleWell`, `GasLiftGroupState`, or `GasLiftStage2`.

Currently only an integer variable `debug_counter` is added to `ALQState`, which can be used as follows: First debugging is switched on globally for `BlackOilWellModel`, `GasLiftSingleWell`, `GasLiftGroupState`, and `GasLiftStage2` by setting `glift_debug` to a true value in `BlackOilWellModelGeneric`. Then, the following debugging code can be added to e.g. one of
the `GasLiftSingleWell`, `GasLiftGroupState`, or `GasLiftStage2` source files:
```C++
auto count = debugUpdateGlobalCounter_();
if (count == some_integer) {
    displayDebugMessage_("stop here");
}
```
Here, the integer `some_integer` is to be determined typically by looking at the debugging output of a previous run. This can be done since the call to `debugUpdateGlobalCounter_()` above will print out the current value of the counter, then increment the counter by one, and finally return that value. And it should then be easy to recognize these values in the debug output. If you find a place in the output that looks suspect, just take a note of the counter value in the output around that point and insert that value for `some_integer` in the code above, then after recompiling with the desired value for `some_integer`, it is now easy to set a breakpoint in GDB at the line `displayDebugMessage_("stop here")` shown in the above snippet. 

This should improve the ability to quickly to set a breakpoint in GDB around at a given time and point in the simulation.


hakonhagland:
jenkins build this please

totto82:
Thanks I think this can be merged. In addition to the extra debug info, this PR introduce a common class which seems like a good idea. Since it is still early Monday morning I will let it linger a few more hours for other developers to have time to comment.

bska:
@hakonhagland: Please note that since `GasLiftCommon.hpp` was not added to `CMakeLists_files.cmake` as part of this PR, the header will not be included in `make install`.  Is that okay/intentional?

hakonhagland:
@bska Thanks for the information. I had forgotten about the `PUBLIC_HEADER_FILES` variable. I see it is documented at line 17 as a comment:

> List of public header files that should be distributed together with the library. The source files can of course include other files than these; you should only add to this list if the *user* of the library needs it.

There are no other gas lift header files currently in the list, and it does not seem likely that a user would need these headers since they are only used in internally by `BlackoilWellModel`. So I guess we should not include it in the list then?

bska:
> There are no other gas lift header files currently in the list,

Right, that's probably a buildsystem bug then.

> it does not seem likely that a user would need these headers since they are only used in internally by `BlackoilWellModel`

It's not quite that simple I'm afraid.  The headers  [`BlackoilWellModel.hpp`](https://github.com/OPM/opm-simulators/blob/90539a615de64b2384a22826e93a35686ec03c07/opm/simulators/wells/BlackoilWellModel.hpp#L52-L56), [`StandardWell.hpp`](https://github.com/OPM/opm-simulators/blob/90539a615de64b2384a22826e93a35686ec03c07/opm/simulators/wells/StandardWell.hpp#L34-L35),and [`WellInterface.hpp`](https://github.com/OPM/opm-simulators/blob/90539a615de64b2384a22826e93a35686ec03c07/opm/simulators/wells/WellInterface.hpp#L45-L47) all include various `GasLift*.hpp` headers directly.  That, in turn, means that any client code that uses those well-related headers will need to have the `GasLift*` headers available too.

hakonhagland:
@bska I have added PR to address, see #3789

bska:
> I have added a PR to address this,

Very good.  Thanks a lot for the quick turn-around.

akva2:
now while at it kill the warnings you introduced, https://ci.opm-project.org/job/opm-simulators/1180/gcc/new/

hakonhagland:
@akva2 Yes good idea, see #3790

PR #3789, opened 2022-01-25
Add gas lift headers to PUBLIC_HEADER_FILES

hakonhagland:
The headers `BlackoilWellModel.hpp`, `StandardWell.hpp`,and `WellInterface.hpp` all include various `GasLift*.hpp` headers directly. That means that any client code that uses those well-related headers will need to have the `GasLift*` headers available too.

Also sorts the headers under `opm/simulators/wells` to make the file easier to read.


hakonhagland:
jenkins build this please

bska (approved):
Thanks.  This looks good to me.  I'll merge into master.

PR #1708, opened 2019-01-14
Warn about parallel logging fallout only on request / Use uppercase logfiles

blattms:
One PR three changes:

1. Use always uppercase *.PRT and *.DBG. 
2. When construcing the logfile names always strip the extension.
3. Only warn about and append parallel logging fallout if explicitly requested using new option --enable-logging-fallout-warning

Coments:
1) Seems consistent as all other output files seem to use that convention, too.  There was a comment/Issue about this but I cannot find it.
2) Is needed as we can start flow like `flow MODEL.` (omitting DATA) and would get MODEL..DBG and MODEL..PRT files.
3) Only annoys users that are not developing flow using MPI

PR #3792, opened 2022-01-27
Prevent multiple *.?.dbg/prt files in production parallel runs.

### LINKED ISSUE #3709: During MPI runs multiple DBG and PRT files are written by Flow
On parallel runs, one DBG and one PRT file is written out for each process. This pollutes the folder, especially when running with many processes.

blattms:
There was a fallout when skipping concating these files to the default ones (PR #1708). We should also have deactivated creating these files at all. As a result these files appeared in all aborted parallel runs.

This change now prevents creating and logging to these files in parallel in a default run (--enable-parallel-logging-fallout-warning=false).

Developers can still activate logging and concating to see whether everything is only logged on the io process by passing --enable-parallel-logging-fallout-warning=true.

Alternative to #3725.
Closes #3709

### DIFF
--- a/opm/simulators/flow/Main.hpp
+++ b/opm/simulators/flow/Main.hpp
@@ -440,11 +440,14 @@ private:
             const bool init_from_restart_file = !EWOMS_GET_PARAM(PreTypeTag, bool, SchedRestart);
             if (outputDir.empty())
                 outputDir = EWOMS_GET_PARAM(PreTypeTag, std::string, OutputDir);
+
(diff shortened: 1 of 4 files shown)


blattms:
:flushed: Of course I introduced this problem myself to begin with.

blattms:
BTW there is logging fallout for injection/production report it seems.

blattms:
jenkins build this please

bska (approved):
Looks good to me and I'll merge into master.  Thanks a lot for the fix.

PR #3795, opened 2022-02-01
Limit well rates to honor group rates

totto82:
If a well is under a group that is limited by a target exceeding gaslift should be removed.
When gradients for the gaslift optimization are calculated the gradients need to take into account the group limit
Accumulate potentials instead of rates when comparing with group targets. We are only interested in the limiting case,
and since potentials >= rates this gives a less strict check. Testing indicates that using the potentials is more stable
since they stay the same during the iterations.


totto82:
jenkins build this please

bska:
> If a well is under a group that is limited by a target exceeding gaslift should be removed.

I don't understand what this sentence is trying to say.  Are there some words missing?  What "should be removed"?  Moreover I don't really understand what you mean by "limited by a target exceeding gaslift", but I'm probably missing something simple.

If not, please try to rephrase the commit message to make it more clear.

totto82:
@bska Sorry for the poor commit message 

Is this more clear:
"If a well is under a group that is limited by a target,  it should use as little gaslift as possible. The reduction algorithm will reduce the gaslift of the well as long as the groups potential is above the groups target. "

bska:
> Is this more clear:

Yes, much better&mdash;thank you for the clarification.  I'll review this PR now with that frame of reference in mind.

totto82:
> Yes, much better—thank you for the clarification. I'll review this PR now with that frame of reference in mind.

Thanks. I will update the commit message along with other issues you may discover in your review.

bska (approved):
I don't follow the complete logic in the gas lift code, but I think I understand most of what's happening in this PR.  I don't see any immediate problems so I think we should merge this as soon as you've updated the commit message.

totto82:
jenkins build this please

PR #3796, opened 2022-02-01
Accumulate Inter-Region Phase Flow Rate Contributions

bska:
This PR introduces a new helper class
```C++
Opm::EclInterRegFlowMapSingleFIP
```
that wraps an object of type
```C++
Opm::data::InterRegFlowMap
```
(see OPM/opm-common#2945) along with the MPI rank's notion of a FIP region array definition (e.g., the local `FIPNUM` array).  The new single-FIP flow map is responsible for accumulating local contributions to the inter-region flows defined by that FIP array.  In the case of connections between MPI ranks, the rank that owns the lowest region ID accumulates the associate flow rates.

We also add a new grid traversal that calculates and aggregates surface condition component flow rates across the bulk connections.  While presently restricted to accumulating inter-region flows, this pass can also be extended to supporting the 3D restart vectors for inter-cell flow rates&mdash;i.e., the `FLS*` and `FLR*` vectors requested by
```ecl
RPTRST
   FLOWS FLORES /
```

### DIFF
--- a/ebos/eclinterregflows.cc
+++ b/ebos/eclinterregflows.cc
@@ -50,9 +50,11 @@ addConnection(const Cell& source,
         };
     }
 
-    if (! (source.isInterior || destination.isInterior)) {
-        // Connection between two cells not on this process.  Unlikely, but
-        // nothing to do here.
+    if (! source.isInterior ||
+        (source.cartesianIndex > destination.cartesianIndex))
+    {
+        // Connection handled in different call.  Don't double-count
+        // contributions.
         return;
     }
 
@@ -64,14 +66,9 @@ addConnection(const Cell& source,
         return;
     }
 
-    if ((source.isInterior && destination.isInterior) ||
-        (source.isInterior && (r1 < r2)) ||
-        (destination.isInterior && (r2 < r1)))
-    {
-        // Inter-region connection internal to an MPI rank or this rank owns
-        // the flow rate across this connection.
-        this->iregFlow_.addConnection(r1, r2, rates);
-    }
+    // Inter-region connection internal to an MPI rank or this rank owns
+    // the flow rate across this connection.
+    this->iregFlow_.addConnection(r1, r2, rates);
 }
 
 void Opm::EclInterRegFlowMapSingleFIP::compress()
@@ -106,4 +103,110 @@ assignGlobalMaxRegionID(const std::size_t regID)
 
     this->maxGlobalRegionID_ = regID;
     return true;
+}
+
+// =====================================================================
+//
+// Implementation of EclInterRegFlowMap (wrapper for multiple arrays)
+//
+// =====================================================================
+
+Opm::EclInterRegFlowMap::
+EclInterRegFlowMap(const std::size_t                numCells,
+                   const std::vector<SingleRegion>& regions)
+{
+    this->regionMaps_.reserve(regions.size());
+    this->names_.reserve(regions.size());
+
+    this->numCells_ = numCells;
+
+    for (const auto& region : regions) {
+        this->regionMaps_.emplace_back(region.definition);
+        this->names_.push_back(region.name);
+    }
+}
+
+void
+Opm::EclInterRegFlowMap::
+addConnection(const Cell& source,
+              const Cell& destination,
+              const data::InterRegFlowMap::FlowRates& rates)
+{
+    for (auto& regionMap : this->regionMaps_) {
+        regionMap.addConnection(source, destination, rates);
+    }
+}
+
+void Opm::EclInterRegFlowMap::compress()
+{
+    for (auto& regionMap : this->regionMaps_) {
+        regionMap.compress();
+    }
+}
+
+void Opm::EclInterRegFlowMap::clear()
+{
+    for (auto& regionMap : this->regionMaps_) {
+        regionMap.clear();
+    }
+
+    this->readIsConsistent_ = true;
+}
+
+const std::vector<std::string>&
+Opm::EclInterRegFlowMap::names() const
+{
+    return this->names_;
+}
+
+std::vector<Opm::data::InterRegFlowMap>
+Opm::EclInterRegFlowMap::getInterRegFlows() const
+{
+    auto maps = std::vector<data::InterRegFlowMap>{};
+    maps.reserve(this->regionMaps_.size());
+
+    for (auto& regionMap : this->regionMaps_) {
+        maps.push_back(regionMap.getInterRegFlows());
+    }
+
+    return maps;
+}
+
+std::vector<std::size_t>
+Opm::EclInterRegFlowMap::getLocalMaxRegionID() const
+{
+    auto maxLocalRegionID = std::vector<std::size_t>{};
+    maxLocalRegionID.reserve(this->regionMaps_.size());
+
+    for (auto& regionMap : this->regionMaps_) {
+        maxLocalRegionID.push_back(regionMap.getLocalMaxRegionID());
+    }
+
+    return maxLocalRegionID;
+}
+
+bool
+Opm::EclInterRegFlowMap::
+assignGlobalMaxRegionID(const std::vector<std::size_t>& regID)
+{
+    if (regID.size() != this->regionMaps_.size()) {
+        return false;
+    }
+
+    auto assignmentOK = true;
+
+    const auto numReg = regID.size();
+    for (auto region = 0*numReg; region < numReg; ++region) {
+        const auto ok = this->regionMaps_[region]
+            .assignGlobalMaxRegionID(regID[region]);
+
+        assignmentOK = assignmentOK && ok;
+    }
+
+    return assignmentOK;
+}
+
+bool Opm::EclInterRegFlowMap::readIsConsistent() const
+{
+    return this->readIsConsistent_;
 }
--- a/ebos/eclinterregflows.cc
+++ b/ebos/eclinterregflows.cc
@@ -1,0 +1,109 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  Copyright 2022 Equinor AS
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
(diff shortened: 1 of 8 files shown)


bska:
jenkins build this opm-common=2945 please

bska:
jenkins build this opm-common=2945 please

bska:
jenkins build this opm-common=2945 please

ebos/eclwriter.hh
```diff
@@ -435,6 +436,41 @@ private:
         OPM_END_PARALLEL_TRY_CATCH("EclWriter::prepareLocalCellData() failed: ", simulator_.vanguard().grid().comm())
     }
 
+    void captureLocalFluxData()
+    {
+        const auto& gridView = this->simulator_.vanguard().gridView();
```
blattms, line 441: Not sure how costly this is. Maybe should check first whether something needs to be done (multiple regions).

tests/test_eclinterregflows.cpp
```diff
@@ -0,0 +1,3413 @@
...
+    const auto& iregFlows = flows.getInterRegFlows();
+    BOOST_CHECK_EQUAL(iregFlows.numRegions(), 5);
+
+    for (auto r1 = 0*iregFlows.numRegions(); r1 < iregFlows.numRegions(); ++r1) {
+        for (auto r2 = r1 + 1; r2 < iregFlows.numRegions(); ++r2) {
+            auto flow = iregFlows.getInterRegFlows(r1, r2);
+            BOOST_CHECK_MESSAGE(! flow.has_value(),
+                                "There must not be inter-regional flow "
+                                "between regions " << r1 << " and " << r2);
+        }
```
blattms: Shouldn't we also check that flow rate is correct?

blattms (commented):
This is kind of a premature review (still took quite some time).

What I could find neither here nor in the other PR is where the multiple connections for a region pair are added up. There seem to be a triple (r1, r2, rate) for every cell pair sharing a face with a region interface.

This adding up also needs to done after communicating the CRSs. But the parallel part seems to come later, I guess.

Did not take a look at the tests (besides of one comment). That is quite some testing. :clap:

ebos/eclwriter.hh
```diff
@@ -435,6 +436,41 @@ private:
...
+        auto const elemEndIt = gridView.template end</*codim=*/0>();
+
+        auto elemCtx = ElementContext { this->simulator_ };
+
+        const auto elemMapper = ElementMapper { gridView, Dune::mcmgElementLayout() };
+        const auto activeIndex = [&elemMapper](const Element& e)
+        {
+            return elemMapper.index(e);
+        };
```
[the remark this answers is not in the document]
blattms: Here I might have been wrong as you used the region indices to determine who owns the face. That should have worked. Sorry for the confusion. 
But with using the cartesian index it now definitely works

ebos/eclwriter.hh
```diff
@@ -435,6 +436,41 @@ private:
...
+             elemIt != elemEndIt; ++elemIt)
+        {
+            const auto& elem = *elemIt;
+
+            elemCtx.updateStencil(elem);
+            elemCtx.updateIntensiveQuantities(/*timeIdx=*/0);
+            elemCtx.updateExtensiveQuantities(/*timeIdx=*/0);
+
+            this->eclOutputModule_->processFluxes(elemCtx, activeIndex);
+        }
```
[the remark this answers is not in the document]
bska: > You would probably say that this needs to be `const elemIt` 😉

If that's the correct way of expressing a mutable iterator to constant `T`s, then yes that would be ideal.

> You may also use the newer more convenient way of iterating:
> ```C++
> for( const auto& elem = elements(gridViews))
> ```

Cool&mdash;thanks.  I've pushed an update to that effect.

ebos/eclinterregflows.cc
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    if (! (source.isInterior || destination.isInterior)) {
+        // Connection between two cells not on this process.  Unlikely, but
+        // nothing to do here.
+        return;
+    }
+
+    const auto r1 = this->region_[ source.activeIndex ];
+    const auto r2 = this->region_[ destination.activeIndex ];
```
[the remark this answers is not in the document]
bska: > Names like `r1` don't really ring a bell. Maybe something more descriptive like `sourceRegion`, `destRegion` would help reading the rest of the code (e.g. the logical expression below)?

Yeah maybe.  I've elected to keep `r1` and `r2` for now in light of having to rethink [the condition](https://github.com/OPM/opm-simulators/blob/c9ce4ba36815d469e30332fc7010b5dc675fc080/ebos/eclinterregflows.cc#L53-L59) for which rank owns the flux.  Now the only real condition on `r1` and `r2` is whether or not the two region values [are the same](https://github.com/OPM/opm-simulators/blob/c9ce4ba36815d469e30332fc7010b5dc675fc080/ebos/eclinterregflows.cc#L64-L67).  If you'd still prefer other names then I'll rename them.

ebos/eclwriter.hh
bska: > Not sure how costly this is. Maybe should check first whether something needs to be done (multiple regions).

I'll run some benchmarks to assess the computational impact of running this pass unconditionally.

tests/test_eclinterregflows.cpp
bska: > Shouldn't we also check that the flow rate is correct?

Not here.  This particular check verifies that there is no flow at all between the two regions.

bska:
benchmark opm-common=2945 please

ebos/eclwriter.hh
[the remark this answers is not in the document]
bska: As you say, we only use the active index to look up the region index of each local cell.  That said, I was unable to make my initial criterion work in parallel, so I've resorted to using global IDs instead.

bska:
The upstream prerequisite PR OPM/opm-common#2945 has been merged.  I'm marking this as "ready for review" now.

bska:
jenkins build this please

ebos/eclinterregflows.hh
```diff
@@ -0,0 +1,362 @@
...
+                auto map = EclInterRegFlowMapSingleFIP {
+                    std::vector<int>(this->numCells_, 1)
+                };
+
+                const auto numMaps = names.size();
+                for (auto n = 0*numMaps; n < numMaps; ++n) {
+                    map.read(buffer);
+                }
+
+                this->readIsConsistent_ = false;
```
blattms, line 322: Have not seen `readIsConsistent()` being queried somewhere besides the test? Is it used? What happens if we end up in the else branch? Aren't the names determined from either the EclState or the FieldProperties and should be the same on every process?

ebos/eclwriter.hh
blattms: This is executed for Norne, right. Hence according to ytelses it's not a real problem.

bska: > This is executed for Norne, right.

Kind of.  We definitely do run the grid traversal loop to calculate the flux values, but we don't really accumulate inter-region flow rates.  While the input file does nominally request [inter-region flow summary vectors](https://github.com/OPM/opm-tests/blob/20bf08cccb001df32479a5845ef9fd1b229acfd2/norne/INCLUDE/SUMMARY/summary.data#L74-L77) it fails to specify any region ID pairs.  We don't (currently) interpret missing region ID pairs as "all possible region ID pairs", but maybe we should.

blattms (commented):
Did take  a look at the tests. Really complicated stuff.

ebos/eclinterregflows.hh
bska: > Have not seen `readIsConsistent()` being queried somewhere besides the test? Is it used?

It's not, but that's (probably) a mistake.  I intended to use this flag to determine whether or not to finalise the global aggregation of the inter-region flows through a `compress()` call in `CollectToIORank`.  I may still do that.

> What happens if we end up in the else branch?

Right now, we "just" ignore the contributions from that partner MPI rank.  If we do check the state before running the global `compress()` we'll end up ignoring all contributions from all ranks.

> Aren't the names determined from either the `EclState` or the `FieldProperties` and should be the same on every process?

They're supposed to be which is why ending up in this branch is labelled as "unexpected".  We might consider stopping the simulation if this situation ever occurs.

bska:
> Did take a look at the tests. Really complicated stuff.

I'm trying to cover most conceivable cases of all cells being in a single region (no inter-region flow), multiple regions on a single rank, single region on multiple ranks, multiple regions on multiple ranks with and without connections.

ebos/eclinterregflows.hh
blattms: > > Have not seen `readIsConsistent()` being queried somewhere besides the test? Is it used?
> 
> It's not, but that's (probably) a mistake. I intended to use this flag to determine whether or not to finalise the global aggregation of the inter-region flows through a `compress()` call in `CollectToIORank`. I may still do that.
> 
> > What happens if we end up in the else branch?
> 
> Right now, we "just" ignore the contributions from that partner MPI rank. If we do check the state before running the global `compress()` we'll end up ignoring all contributions from all ranks.
> 

This will give us a rather hard time noticing this. Why not  throw and remove the rest?

bska: > Why not throw and remove the rest?

I suppose that would work, but we need to be mindful of the context here.

This particular code will run during `CollectToIORank::collect()`, and in that case only on the I/O rank.  Suppose that we've protected the [relevant region](https://github.com/OPM/opm-simulators/blob/b9a1270bbb90e0f40004c5fab6cbc7a1402e3e50/ebos/eclwriter.hh#L194-L205) with `OPM_BEGIN_PARALLEL_TRY_CATCH()` and `OPM_END_PARALLEL_TRY_CATCH()` and that the third of N (> 3) message happens to fail in this very unlikely way.  What will happen to the remaining messages that have not yet been received by the I/O rank and, more importantly, to their senders?  Will this deadlock the entire set of MPI processes?

bska: >> Will this [unlikely scenario] deadlock the entire set of MPI processes?
>
>Yep, it might deadlock as we might need to receive messages from other processes afterwards. We need to find a way throw after the [exchange](https://github.com/OPM/opm-simulators/blob/b9a1270bbb90e0f40004c5fab6cbc7a1402e3e50/ebos/collecttoiorank.cc#L1010) has completed, e.g. by querying a data member.

That's what I feared.  In that case I'm keeping the `readIsConsistent()` predicate for this purpose.  I've pushed an update in #3811 that checks `readIsConsistent()` on the I/O rank after the `CollectToIORank::collect()` call completes and communicates the failure condition to all ranks.

bska:
jenkins build this please

PR #3799, opened 2022-02-02
fixed: use the --porcelain option when parsing git status

akva2:
the formatting for the standard git status output is unstable,
and has changed in newer versions of the git client. this lead
to missing lines in untracked files.


bska (approved):
Yep, good idea.  We should prefer machine readable formats when parsing outputs of other utilities.  I'll merge into master.

PR #3800, opened 2022-02-03
Move gpu files

Tongdongq:
This PR moves some of the files used for GPU to either `opencl/` or `cuda/`.
Not all files are moved there, as some are used for both or neither.


blattms:
Seems like ChowPatel had quite some bitrod because of macro usage. Maybe we should do something about that eventually.

blattms:
jenkins build this please

blattms:
thanks. Always good to cleanup the repository structure. :+1:

PR #3801, opened 2022-02-03
Add more performance summary output

totto82:
This commit ads: NEWTON, MLINEARS, MSUMLINS, MSUMNEWT
and no also NLINEARS, NLINSMAX and NLINSMIN

### DIFF
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -39,6 +39,7 @@
 #include <ebos/eclgenericwriter.hh>
 
 #include <string>
+#include <limits>
 
 namespace Opm::Properties {
 
@@ -207,6 +208,27 @@ public:
         // Add TCPU
         if (totalCpuTime != 0.0) {
             miscSummaryData["TCPU"] = totalCpuTime;
+        }
+        if (this->sub_step_report_.total_newton_iterations != 0) {
+            miscSummaryData["NEWTON"] = this->sub_step_report_.total_newton_iterations;
+        }
+        if (this->sub_step_report_.total_linear_iterations != 0) {
+            miscSummaryData["MLINEARS"] = this->sub_step_report_.total_linear_iterations;
+        }
+        if (this->sub_step_report_.total_newton_iterations != 0) {
+            miscSummaryData["NLINEARS"] =  static_cast<float>(this->sub_step_report_.total_linear_iterations) / this->sub_step_report_.total_newton_iterations;
+        }
+        if (this->sub_step_report_.min_linear_iterations != std::numeric_limits<unsigned int>::max()) {
+            miscSummaryData["NLINSMIN"] = this->sub_step_report_.min_linear_iterations;
+        }
+        if (this->sub_step_report_.max_linear_iterations != 0) {
+            miscSummaryData["NLINSMAX"] = this->sub_step_report_.max_linear_iterations;
+        }
+        if (this->simulation_report_.success.total_newton_iterations != 0) {
+            miscSummaryData["MSUMLINS"] = this->simulation_report_.success.total_linear_iterations;
+        }
+        if (this->simulation_report_.success.total_newton_iterations != 0) {
+            miscSummaryData["MSUMNEWT"] = this->simulation_report_.success.total_newton_iterations;
         }
 
         this->evalSummary(reportStepNum, curTime,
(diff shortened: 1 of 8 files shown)


totto82:
jenkins build this please

akva2:
one would expect modifications to UnsupportedKeywords

totto82:
> one would expect modifications to UnsupportedKeywords

The added keywords where not in the list in the first place.

akva2:
yeah i checked. that is true except NLINEARS

totto82:
And we still don't support NLINEARS. 
We also miss NLINSMAX and NLINSMIN

I also this we should add some new keywords for the failed numbers both linear, newton and time.

akva2:
okay my bad, i guess your NLINEARS is something else then.

totto82:
> okay my bad, i guess your NLINEARS is something else then. aha, in code it's MLINEARS, in PR description it is NLINEARS

He,he, its was indeed "my bad". N and M are to close both on the keyboard and on the screen. I edited the PR description to avoid more confusion.

totto82:
> And we still don't support NLINEARS.
> We also miss NLINSMAX and NLINSMIN

They should be trivial to add, so I will add them also in this PR.

totto82:
This will need https://github.com/OPM/opm-common/pull/2953 to output NLINEARS

totto82:
jenkins build this opm-common=2953 please

ebos/eclwriter.hh
```diff
@@ -208,6 +209,27 @@ public:
...
             miscSummaryData["TCPU"] = totalCpuTime;
         }
+        if (this->sub_step_report_.total_newton_iterations != 0) {
+            miscSummaryData["NEWTON"] = this->sub_step_report_.total_newton_iterations;
+        }
+        if (this->sub_step_report_.total_linear_iterations != 0) {
+            miscSummaryData["MLINEARS"] = this->sub_step_report_.total_linear_iterations;
+        }
+        if (this->sub_step_report_.total_newton_iterations != 0) {
+            miscSummaryData["NLINEARS"] =  this->sub_step_report_.total_linear_iterations / this->sub_step_report_.total_newton_iterations;
```
bska, line 219: This is an integer.  Did you consider making it a floating-point value instead?

bska (commented):
I think this looks good.  I only have one question about how we calculate the average number of linear iterations per non-linear iteration, but otherwise I think we should merge this.

ebos/eclwriter.hh
totto82: It was supposed to be floating-point, but you are right its not. I though it should work since 
`std::map<std::string, double> miscSummaryData;` 
I guess I need use `static_cast`

totto82: That was better. Now NLINEARS is indeed the average. Thanks for noticing.

totto82:
jenkins build this opm-common=2953 please

ebos/eclwriter.hh
bska: > I though it should work since `std::map<std::string, double> miscSummaryData;`

You're right that the value will be **stored** as a floating-point value, but one of the pitfalls of C++ is that the expression
```
integer / integer
```
performs integer division.  This means that the value of that expression will be an integer too and that that integer will be converted to a floating-point value when it's stored in the `map<>`.

> I guess I need use `static_cast`.

Yes, that's the simplest workaround.

totto82:
> Before we merge this, I think we should also consider whether changes in any of these performance quantities should be enough to reject a new simulation candidate for the purpose of continuous integration testing

I think they should be tested. My experience is that if number of linear/newton steps change the results often change anyway. I suggest we include them first. If it gets annoying we can add them to the blacklist

bska:
> > Before we merge this, I think we should also consider whether changes in any of these performance quantities should be enough to reject a new simulation candidate for the purpose of continuous integration testing
> 
> I think they should be tested. My experience is that if number of linear/newton steps change the results often change anyway. I suggest we include them first. If it gets annoying we can add them to the blacklist

Very good.  I'll update the reference data from the upstream PR (OPM/opm-common#2953).

bska:
New reference solutions and upstream prerequisites have been installed on the CI system.  Merging this now to bring the overall system back to a consistent state.  Thanks a lot for the contribution.

PR #3807, opened 2022-02-08
update well potentials in wellState along with alq and update guide rates accordingly

totto82:
This makes sure the guide rates (i.e. the distribution between wells/groups) are updated when gaslift changes during iterations.
(LIFTOPT item 4 = YES) and as a result reduce oscillations in the solution.


totto82:
jenkins build this please

hakonhagland:
Thanks this looks good to me. I tested this PR with the case `4_GLIFT_MODEL5`:

pr3807

the blue curve is the results for this PR and oscillations are much less than for the current code (green curve)!

PR #3812, opened 2022-02-14
Adds gaslift optimization support for multisegmented wells (MSW)

hakonhagland:
Adds gaslift optimization support for multisegmented wells (MSW).

### DIFF
--- a/opm/simulators/wells/GasLiftSingleWell.hpp
+++ b/opm/simulators/wells/GasLiftSingleWell.hpp
@@ -42,7 +42,7 @@ namespace Opm
 
     public:
         GasLiftSingleWell(
-            const WellInterface<TypeTag> &std_well,
+            const WellInterface<TypeTag> &well,
             const Simulator &ebos_simulator,
             const SummaryState &summary_state,
             DeferredLogger &deferred_logger,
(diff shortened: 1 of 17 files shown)


hakonhagland:
jenkins build this please

opm/simulators/wells/GasLiftSingleWell.hpp
```diff
@@ -69,7 +61,7 @@ namespace Opm
         void setAlqMaxRate_(const GasLiftOpt::Well& well);
 
         const Simulator &ebos_simulator_;
-        const StdWell &std_well_;
+        const WellInterface<TypeTag> &std_well_;
     };
```
totto82, line 65: Please rename to well_  
and also:
const WellInterfaceGeneric &getStdWell() const override { return std_well_; }
to getWell()
to reflect the change from stdWell well to wellInterface. 
I

totto82 (changes requested):
Thanks. Both the extra debugging and the inclusion of MSW looks good. Can be merged when the well naming is fixed. It is a bit confusing to call it std_well_ when it can be both msw and stw.

hakonhagland:
jenkins build this please

opm/simulators/wells/GasLiftSingleWell.hpp
hakonhagland: Good idea! See added commit.

PR #3815, opened 2022-02-15
Subdomain matrix for opencl ILU preconditioner

Tongdongq:
Requires https://github.com/andrthu/opm-grid/tree/partition-without-scatter, works with masters of 2022-1-22 10:00.
Andreas has created branches that send 2 matrices to the openclSolver.
The original one is used for spmv, the other one is send to the ILU preconditioner and is dubbed jacobiMatrix. The jacobiMatrix contains less blocks (subset of sparsity pattern of original matrix) and therefore more parallelism can be extracted by the ILU preconditioner.
This jacobiMatrix is obtained by performed subdomain decomposition, similar to the partitioning that occurs when using multiple MPI processes. These partitions are then put back into this jacobiMatrix.
The jacobiMatrix can be configured using the runtime parameter `--num-jacobi-blocks`. Values 0 and 1 don't do anything, and the original matrix is used for both spmv and ILU.

The sparsity pattern for jacobiMatrix is calculated once, using the partition data. And the respective nonzero blocks are copied from the original matrix every linear solve.
The BILU0 object receives both the original and jacobiMatrix, and reorders then both. The reordered original is used in the spmv and AMG part of the CPR, and the jacobiMatrix stays private inside the BILU0 object.
The jacobiMatrix is exposed to the AMG part of the CPR in an unpublished commit for testing.
There are many merge conflicts because many opencl files where moved the `opencl/` folder.

@blattms @atgeirr @andrthu 
```
Benchmarks:
Notation:
X: A (B+C), D, E, F | G, H
X: Number of Jacobi blocks
A: Total time
B: Assembly time
C: Linear solve time
D: Overall Linearizations
E: Overall Newton Iterations
F: Overall Linear Iterations
G: Number of colors in BILU0
H: Number of fallbacks to Dune

AMD EPYC Milan 7763 + Tesla A100
opencl, level_scheduling, wellcontribs in matrix, no CPR
  0: 546 (142+215), 1810, 1447, 20797 | 194, 0
  1: 551 (144+216), 1810, 1447, 20797 | 194, 0
  4: 524 (145+189), 1826, 1466, 21392 | 132, 0
 32: 488 (142+157), 1787, 1430, 21867 |  92, 0
 64: 478 (143+148), 1796, 1440, 22353 |  75, 0
 80: 465 (141+137), 1783, 1426, 22536 |  62, 0
100: 467 (143+137), 1785, 1427, 22690 |  55, 0
128: 466 (142+136), 1777, 1420, 23215 |  55, 0
150: 454 (141+128), 1775, 1419, 24245 |  42, 0
200: 486 (150+141), 1841, 1476, 25094 |  55, 0

R5 2400G, GTX 1050Ti, using CPR, but AMG hierarchy uses original matrix
opencl, level_scheduling, wellcontribs in matrix
  0: 449 (104+166), 1442, 1097, 2768 | 194, 0
  4: 464 (107+177), 1447, 1104, 2787 | 132, 0
 16: 458 (105+172), 1443, 1100, 2885 |  91, 0
 32: 457 (104+174), 1435, 1093, 2952 |  92, 0
 64: 465 (106+178), 1456, 1111, 3139 |  75, 0
100: 462 (104+177), 1446, 1103, 3182 |  55, 0
150: 466 (106+179), 1447, 1104, 3311 |  42, 0
opencl, graph_coloring, wellcontribs in matrix
  0: 464 (107+175), 1470, 1126, 3555 | 32, 0
  4: 505 (112+208), 1505, 1160, 3914 | 31, 0
 16: 498 (109+205), 1497, 1151, 4009 | 30, 0
 32: 489 (108+198), 1487, 1144, 3761 | 31, 0
 64: 500 (110+205), 1514, 1167, 4048 | 29, 0
100: 535 (118+226), 1623, 1273, 4744 | 29, 0
150: 519 (114+218), 1587, 1238, 4577 | 31, 0

R5 2400G, GTX 1050Ti, using CPR, but AMG hierarchy uses jacobiMatrix
level_scheduling, wellcontribs in matrix, no runtimes reported:
  0: 1442, 1097,  2768 | 0
  4: 1501, 1155,  6030 | 0
 16: 1531, 1183,  6722 | 0
 64: 1596, 1247,  9394 | 0
100: 1703, 1345, 11729 | 0
150: 1917, 1551, 17316 | 0
200: 1988, 1617, 18961 | 0
256: 2044, 1669, 27859 | 1 fallback
ref: 1810, 1447, 20797 | 0, no CPR
```
Main takeaways:
- the jacobiMatrix speeds up computation a lot when using only the ILU preconditioner
- it does not when using the CPR preconditioner (although implementation error is always possible)
- the jacobiMatrix does not increase parallelism for graph_coloring
- the AMG part of the CPR should not use the jacobiMatrix

### DIFF
--- a/opm/simulators/linalg/ISTLSolverEbos.hpp
+++ b/opm/simulators/linalg/ISTLSolverEbos.hpp
@@ -161,7 +161,15 @@ namespace Opm
             // Set it up manually
             ElementMapper elemMapper(simulator_.vanguard().gridView(), Dune::mcmgElementLayout());
             detail::findOverlapAndInterior(simulator_.vanguard().grid(), elemMapper, overlapRows_, interiorRows_);
-
+	    numJacobiBlocks_ = EWOMS_GET_PARAM(TypeTag, int, NumJacobiBlocks);
+	    useWellConn_ = EWOMS_GET_PARAM(TypeTag, bool, MatrixAddWellContributions);
+	    if (numJacobiBlocks_ > 1) {
+		const auto wellsForConn = simulator_.vanguard().schedule().getWellsatEnd();
+		detail::setWellConnections(simulator_.vanguard().grid(), wellsForConn, useWellConn_,
+					   wellConnectionsGraph_, numJacobiBlocks_);
+		std::cout << "Create block-Jacobi pattern" << std::endl;
+		blockJacobiAdjacency();
+	    }
             useWellConn_ = EWOMS_GET_PARAM(TypeTag, bool, MatrixAddWellContributions);
 #if HAVE_FPGA
             // check usage of MatrixAddWellContributions: for FPGA they must be included
@@ -273,8 +281,15 @@ namespace Opm
                 }
 #endif
 
+		if (numJacobiBlocks_ > 1) {
+		    copyMatToBlockJac(getMatrix(), *blockJacobiForGPUILU0_);
                 // Const_cast needed since the CUDA stuff overwrites values for better matrix condition..
-                bdaBridge->solve_system(const_cast<Matrix*>(&getMatrix()), *rhs_, *wellContribs, result);
+		    bdaBridge->solve_system(const_cast<Matrix*>(&getMatrix()), &*blockJacobiForGPUILU0_,
+					    numJacobiBlocks_, *rhs_, *wellContribs, result);
+		}
+		else
+		    bdaBridge->solve_system(const_cast<Matrix*>(&getMatrix()), const_cast<Matrix*>(&getMatrix()),
+					    numJacobiBlocks_, *rhs_, *wellContribs, result);
                 if (result.converged) {
                     // get result vector x from non-Dune backend, iff solve was successful
                     bdaBridge->get_result(x);
@@ -497,6 +512,78 @@ namespace Opm
                 }
         }
 
+	/// Create sparsity pattern for block-Jacobi matrix based on partitioning of grid.
+	void blockJacobiAdjacency()
+	{
+	    const auto& grid = simulator_.vanguard().grid();
+	    std::vector<int> cell_part = simulator_.vanguard().cellPartition();
+
+	    typedef typename Matrix::size_type size_type;
+	    size_type numCells = grid.size( 0 );
+	    blockJacobiForGPUILU0_.reset(new Matrix(numCells, numCells, Matrix::random));
+
+	    std::vector<std::set<size_type>> pattern;
+	    pattern.resize(numCells);
+
+	    const auto& lid = grid.localIdSet();
+	    const auto& gridView = grid.leafGridView();
+	    auto elemIt = gridView.template begin<0>();
+	    const auto& elemEndIt = gridView.template end<0>();
+
+	    //Loop over cells
+	    for (; elemIt != elemEndIt; ++elemIt)
+            {
+
+		const auto& elem = *elemIt;
+		size_type idx = lid.id(elem);
+		pattern[idx].insert(idx);
+
+		// Add well non-zero connections
+		for (auto wc = wellConnectionsGraph_[idx].begin(); wc!=wellConnectionsGraph_[idx].end(); ++wc)
+		    pattern[idx].insert(*wc);
+
+		int locPart = cell_part[idx];
+
+		//Add neighbor if it is on the same part
+		auto isend = gridView.iend(elem);
+		for (auto is = gridView.ibegin(elem); is!=isend; ++is)
+                {
+		    //check if face has neighbor
+		    if (is->neighbor())
+                    {
+			size_type nid = lid.id(is->outside());
+			int nabPart = cell_part[nid];
+			if (locPart == nabPart) 
+			    pattern[idx].insert(nid);
+		    }
+
+		    blockJacobiForGPUILU0_->setrowsize(idx, pattern[idx].size());
+		}
+	    }
+	    blockJacobiForGPUILU0_->endrowsizes();
+	    for (size_type dofId = 0; dofId < numCells; ++dofId)
+	    {
+		auto nabIdx = pattern[dofId].begin();
+		auto endNab = pattern[dofId].end();
+		for (; nabIdx != endNab; ++nabIdx)
+		{
+		    blockJacobiForGPUILU0_->addindex(dofId, *nabIdx);
+		}
+	    }
+
+	    blockJacobiForGPUILU0_->endindices();
+	}
+
+	void copyMatToBlockJac(Matrix& mat, Matrix& blockJac)
+	{
+	    auto rbegin = blockJac.begin();
+	    auto rend = blockJac.end();
+	    for (auto row = rbegin; row != rend; ++row) {
+		for (auto col = (*row).begin(); col != (*row).end(); ++col) {
+		    blockJac[row.index()][col.index()] = mat[row.index()][col.index()];
+		}
+	    }
+	}
 
         Matrix& getMatrix()
         {
@@ -517,6 +604,8 @@ namespace Opm
         Matrix* matrix_;
         Vector *rhs_;
 
+        std::unique_ptr<Matrix> blockJacobiForGPUILU0_;
+
         std::unique_ptr<FlexibleSolverType> flexibleSolver_;
         std::unique_ptr<AbstractOperatorType> linearOperatorForFlexibleSolver_;
         std::unique_ptr<WellModelAsLinearOperator<WellModel, Vector, Vector>> wellOperator_;
@@ -530,6 +619,7 @@ namespace Opm
         FlowLinearSolverParameters parameters_;
         PropertyTree prm_;
         bool scale_variables_;
+	int numJacobiBlocks_;
 
         std::shared_ptr< CommunicationType > comm_;
     }; // end ISTLSolver
--- a/opm/simulators/linalg/ISTLSolverEbos.hpp
+++ b/opm/simulators/linalg/ISTLSolverEbos.hpp
@@ -272,7 +272,7 @@ namespace Opm
             if (use_gpu || use_fpga) {
                 const std::string accelerator_mode = EWOMS_GET_PARAM(TypeTag, std::string, AcceleratorMode);
                 auto wellContribs = WellContributions::create(accelerator_mode, useWellConn_);
-                bdaBridge->initWellContributions(*wellContribs);
+                bdaBridge->initWellContributions(*wellContribs, x.N() * x[0].N());
 
                 // the WellContributions can only be applied separately with CUDA or OpenCL, not with an FPGA or amgcl
 #if HAVE_CUDA || HAVE_OPENCL
--- a/opm/simulators/linalg/ISTLSolverEbos.hpp
+++ b/opm/simulators/linalg/ISTLSolverEbos.hpp
@@ -520,10 +520,7 @@ namespace Opm
 
             typedef typename Matrix::size_type size_type;
             size_type numCells = grid.size( 0 );
-            blockJacobiForGPUILU0_.reset(new Matrix(numCells, numCells, Matrix::random));
-
-            std::vector<std::set<size_type>> pattern;
-            pattern.resize(numCells);
+            blockJacobiForGPUILU0_.reset(new Matrix(numCells, numCells, 8, 0.4, Matrix::implicit));
 
             const auto& lid = grid.localIdSet();
             const auto& gridView = grid.leafGridView();
@@ -533,14 +530,14 @@ namespace Opm
             //Loop over cells
             for (; elemIt != elemEndIt; ++elemIt)
             {
-
                 const auto& elem = *elemIt;
                 size_type idx = lid.id(elem);
-                pattern[idx].insert(idx);
+                blockJacobiForGPUILU0_->entry(idx, idx) = 0.0;
 
                 // Add well non-zero connections
-                for (auto wc = wellConnectionsGraph_[idx].begin(); wc!=wellConnectionsGraph_[idx].end(); ++wc)
-                    pattern[idx].insert(*wc);
+                for (auto wc = wellConnectionsGraph_[idx].begin(); wc!=wellConnectionsGraph_[idx].end(); ++wc) {
+                    blockJacobiForGPUILU0_->entry(idx, *wc) = 0.0;
+                }
 
                 int locPart = cell_part[idx];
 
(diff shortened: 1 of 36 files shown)


alfbr:
Interesting results! I am still a bit in the dark as to how to understand the numbers though. I assume the first table is the ILU0 preconditioner? The first column, is that the scatter resulting from that number of MPI processes? While the second column is the total run time? In any case, very impressive gains on the linear solver for the first table!

blattms:
I have created a draft PR OPM/opm-grid#568 that should be a version of the work of @andrthu without merges from master. 

@Tongdongq Will you rebase this one on current master?

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -497,6 +512,78 @@ namespace Opm
...
+            for (; elemIt != elemEndIt; ++elemIt)
+            {
+
+                const auto& elem = *elemIt;
+                size_type idx = lid.id(elem);
+                pattern[idx].insert(idx);
+
+                // Add well non-zero connections
+                for (auto wc = wellConnectionsGraph_[idx].begin(); wc!=wellConnectionsGraph_[idx].end(); ++wc)
+                    pattern[idx].insert(*wc);
```
blattms, line 543: This is interesting. If there are wells the domains might still be connected. Was not clear to me.

@andrthu How are you doing the decomposition? Are the wells allowed to be cut?

blattms (commented):
We have similar solvers in dune-istl https://www.dune-project.org/doxygen/2.8.0/classDune_1_1SeqOverlappingSchwarz.html. In my phd thesis I constructed the subdomains using the matrix and the AMG aggregation algorithm to capture the some effects of the problem (similar to transmissibilities). That might be an alternative approach. But there we are sure that there are not connections between the domains. That is different here.

@andrthu Is it essential that the well connections are resembled by the sparsity pattern? In that case this here is something different as the domains might be connected. If they would not be, we could probably to coloring, scheduling and other things in parallel. Are we using the well connections for loadbalancing?

andrthu:
The wellConnectionsGraph_ is empty if MatrixAddWellContributions=False. wellConnectionsGraph_ is needed when MatrixAddWellContributions=True because we use the grid to create the sparsity pattern.

The partitioning was done with allowDistWells=False. This means that splitting of wells is unlikely, but also not impossible (I think the noScatter partitioner skips the manual moving of cells if wells are distributed). So if MatrixAddWellContributions=True, there is a possibility that subdomains will be reconnected through the well connections.

It seems that the best way of fixing this is to use the matrix+cell_part to create the sparsity pattern instead of the grid+cell_part+wellConnectionsGraph_, as already suggested.

atgeirr:
> In the linear solve time of the 1st set of benchmarks (Tesla A100), without CPR.

Ok, so it is comparing the 215s to 128s, right?

Tongdongq:
> > In the linear solve time of the 1st set of benchmarks (Tesla A100), without CPR.
> 
> Ok, so it is comparing the 215s to 128s, right?

Yes

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -161,7 +161,15 @@ namespace Opm
             // Set it up manually
             ElementMapper elemMapper(simulator_.vanguard().gridView(), Dune::mcmgElementLayout());
             detail::findOverlapAndInterior(simulator_.vanguard().grid(), elemMapper, overlapRows_, interiorRows_);
-
+            numJacobiBlocks_ = EWOMS_GET_PARAM(TypeTag, int, NumJacobiBlocks);
+            useWellConn_ = EWOMS_GET_PARAM(TypeTag, bool, MatrixAddWellContributions);
+            if (numJacobiBlocks_ > 1) {
```
blattms, line 166: Condition is not consistent with the one in the vanguard

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -273,8 +281,15 @@ namespace Opm
                 }
 #endif
 
+                if (numJacobiBlocks_ > 1) {
```
blattms, line 284: Same here. cell_part will only be meaningful for serial run.

opm/simulators/linalg/bda/BILU0.cpp
```diff
@@ -244,6 +356,102 @@ bool BILU0<block_size>::create_preconditioner(BlockedMatrix *mat)
     return true;
 } // end create_preconditioner()
 
+template <unsigned int block_size>
+bool BILU0<block_size>::create_preconditioner(BlockedMatrix *mat, BlockedMatrix *jacMat)
```
blattms, line 360: Can't we get along with one create_preconditioner method and some if statements? This seems to be a lot of duplicate code.

opm/simulators/linalg/bda/BILU0.cpp
```diff
@@ -150,6 +150,118 @@ bool BILU0<block_size>::analyze_matrix(BlockedMatrix *mat)
     return true;
 } // end init()
 
+template <unsigned int block_size>
+bool BILU0<block_size>::analyze_matrix(BlockedMatrix *mat, BlockedMatrix *jacMat)
```
blattms, line 154: Can't we get along with one analyze_matrix method and some if statements? This seems to be a lot of duplicate code.

atgeirr:
If it is easy for you to rerun the experiment, it would be interesting to see if changing the partitioning weights gives an improvement. In regular Flow this is done by changing the parameter `--edge-weights-method=INTEGER` [Choose edge-weighing strategy: 0=uniform, 1=trans, 2=log(trans). Default: 1]

opm/simulators/linalg/bda/BdaBridge.cpp
```diff
@@ -193,6 +193,8 @@ void BdaBridge<BridgeMatrix, BridgeVector, block_size>::copySparsityPatternFromI
 
 template <class BridgeMatrix, class BridgeVector, int block_size>
 void BdaBridge<BridgeMatrix, BridgeVector, block_size>::solve_system([[maybe_unused]] BridgeMatrix* mat,
+                                                                     [[maybe_unused]] BridgeMatrix* blockMat,
+                                                                     [[maybe_unused]] int numJacobiBlocks,
```
blattms, line 197: Use `size_t`? Why `[[maybe_unused]]`? Seems like both are used in the code (BTW: might hold for the remaining parameters, too).

opm/simulators/linalg/bda/amgclSolverBackend.cpp
```diff
@@ -359,6 +359,16 @@ void amgclSolverBackend<block_size>::get_result(double *x_) {
 } // end get_result()
 
 
+template <unsigned int block_size>
+SolverStatus amgclSolverBackend<block_size>::solve_system2(int N_, int nnz_, int dim,
+                                                           double *vals, int *rows, int *cols, double *b,
+                                                           int nnz2, double *vals2, int *rows2, int *cols2,
+                                                           WellContributions& wellContribs, BdaResult &res)
+{
+    (void) nnz2; (void) vals2; (void) rows2; (void) cols2;
```
blattms, line 368: Use `[[maybe_unused]]`?

blattms (commented):
Another pass.

We should try to reduce code duplication by using one method for both cases (1. only one matrix, 2. preconditioner and spmv matrix).

opm/simulators/linalg/bda/BdaBridge.cpp
Tongdongq: Do you mean the param `numJacobiBlocks`?
It is read by `EWOMS_GET_PARAM(TypeTag, int, NumJacobiBlocks)`, that's probably why it's an `int`.
`[[maybe_unused]]` is removed.

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -497,6 +512,78 @@ namespace Opm
...
+            blockJacobiForGPUILU0_->endindices();
+        }
+
+        void copyMatToBlockJac(Matrix& mat, Matrix& blockJac)
+        {
+            auto rbegin = blockJac.begin();
+            auto rend = blockJac.end();
+            for (auto row = rbegin; row != rend; ++row) {
+                for (auto col = (*row).begin(); col != (*row).end(); ++col) {
+                    blockJac[row.index()][col.index()] = mat[row.index()][col.index()];
```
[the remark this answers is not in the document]
Tongdongq: 3c9d036b
Copy goes from 2.9 to 2.6ms on my machine.

opm/simulators/linalg/ISTLSolverEbos.hpp
Tongdongq: 0ea1f86

opm/simulators/linalg/ISTLSolverEbos.hpp
Tongdongq: 0ea1f86

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -520,10 +520,7 @@ namespace Opm
 
             typedef typename Matrix::size_type size_type;
             size_type numCells = grid.size( 0 );
-            blockJacobiForGPUILU0_.reset(new Matrix(numCells, numCells, Matrix::random));
-
-            std::vector<std::set<size_type>> pattern;
-            pattern.resize(numCells);
+            blockJacobiForGPUILU0_.reset(new Matrix(numCells, numCells, 8, 0.4, Matrix::implicit));
```
Tongdongq, line 523: @blattms These number are arbitrarily chosen, based on NORNE and the example in the documentation.
How can we determine better numbers?

blattms: I must admit that I have previously also struggled with this build mode (in AMG).
After meditating over the documentation of the implicit build mode, I am assuming that the using 0.0 as the first parameter does not work, does it?

Maybe we can at least estimate the 4th factor. We now that no row can have more than 8 entries for 3D CpGrid. If we calculate the average row size of the original matrix as a=ceil((double(A.nnz()/A.N()), we could use cs= 1.05*(8-a). Does that make sense / work?

As a safety net we could catch the exception and use another build mode (row_wise) for the case the estimate was not good enough.

Anyway we should probably also benchmark implicit versus row_wise mode.

Tongdongq: For NORNE, row_wise is about 2 ms slower. More jacobi blocks means faster building.
For a bigger case, it varied wildly on my machine. Implicit between 1.5 and 12 s, row_wise between 1.2 and 8.5 s.
I think row_wise is still better overall.

blattms:
jenkins build this please

blattms:
Sorry, but it does not compile. See jenkins run.

akva2:
jenkins build this opm-grid=568 please

blattms:
:flushed: And my laziness was instantly discovered. Thanks @akva2 for actually looking.

blattms:
Please test all possibilities: No GPU, OpenCL only, Cuda only, both.

```
./../tests/test_cusparseSolver.cpp: In instantiation of 'Dune::BlockVector<Dune::FieldVector<double, bz> > testCusparseSolver(const ptree&, const string&, const string&) [with int bz = 3; boost::property_tree::ptree = boost::property_tree::basic_ptree<std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >; std::string = std::__cxx11::basic_string<char>]':
../../tests/test_cusparseSolver.cpp:108:68:   required from here
../../tests/test_cusparseSolver.cpp:90:29: error: no matching function for call to 'Opm::BdaBridge<Dune::BCRSMatrix<Dune::FieldMatrix<double, 3, 3>, std::allocator<Dune::FieldMatrix<double, 3, 3> > >, Dune::BlockVector<Dune::FieldVector<double, 3>, std::allocator<Dune::FieldVector<double, 3> > >, 3>::solve_system(Matrix*, Vector&, Opm::WellContributions&, Dune::InverseOperatorResult&)'
   90 |         bridge->solve_system(&matrix, rhs, *wellContribs, result);
      |         ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from ../../tests/test_cusparseSolver.cpp:32:
../../opm/simulators/linalg/bda/BdaBridge.hpp:75:10: note: candidate: 'void Opm::BdaBridge<BridgeMatrix, BridgeVector, block_size>::solve_system(BridgeMatrix*, BridgeMatrix*, int, BridgeVector&, Opm::WellContributions&, Opm::InverseOperatorResult&) [with BridgeMatrix = Dune::BCRSMatrix<Dune::FieldMatrix<double, 3, 3>, std::allocator<Dune::FieldMatrix<double, 3, 3> > >; BridgeVector = Dune::BlockVector<Dune::FieldVector<double, 3>, std::allocator<Dune::FieldVector<double, 3> > >; int block_size = 3; Opm::InverseOperatorResult = Dune::InverseOperatorResult]'
   75 |     void solve_system(BridgeMatrix *bridgeMat, BridgeMatrix *jacMat, int numJacobiBlocks, BridgeVector &b, WellContributions& wellContribs, InverseOperatorResult &result);
      |          ^~~~~~~~~~~~
../../opm/simulators/linalg/bda/BdaBridge.hpp:75:10: note:   candidate expects 6 arguments, 4 provided
```

blattms:
jenkins build this please

blattms:
:tada: Great. Thanks a lot. Really appreciated.

blattms:
jenkins build this please

blattms:
Probably, not really related to these changes, but asking here, anyway:

Is it expected that `flow --accelerator-mode=opencl --opencl-ilu-reorder=none NORNE_ATW2013` is much slower than when using cusparse? Or am I missing something again (like `martrix-add-wellcontributions=true`)?

blattms:
There you go. For people who can read properly the `opencl-ilu-reorder` name probably indicates that it has no effect when using cusparse. Seems to be clear now to the public that I am not one of these people :flushed:

blattms:
Sorry, but I really don't like your last commit. Using static variable to store data in a function is a bit scary and also really hard to
understand. Maybe the function should just get all needed state information and remove the part of the state needed for the next call?

Do we really need to check that the row itself is contiguous? It is always BCRS and hence it should hold.

blattms:
Shoot. Just wanted to cherry-pick the twio bugfixes ([9197877](https://github.com/OPM/opm-simulators/pull/3815/commits/9197877f9454df01bee07ac3d5b10fa86d4ef671), [9f3b8b6](https://github.com/OPM/opm-simulators/pull/3815/commits/9f3b8b6a7d9242a1bd95d323100b6722babf20da) from here to the release, but it seems this more complicated than I thought. 

@Tongdongq Could you maybe extract them to a sperate PR? I need to be able to test the release and failing tests are rather awful.

opm/simulators/linalg/bda/BdaBridge.cpp
```diff
@@ -191,6 +191,40 @@ void BdaBridge<BridgeMatrix, BridgeVector, block_size>::copySparsityPatternFromI
...
+                            OPM_THROW(std::logic_error, "Error memory of Matrix looks not contiguous");
+                        }
+                    }
+                }
+            }
+        } else {
+            break;
+        }
+    }
+}
```
blattms, line 225: I think we can assume that the entries of one row are contiguous. Wouldn't it suffice to only check
```
return mat.rows() < 2 || distance(&(*row[1]->begin())[0][0]), &(*(advance(row[0]->begin(), size_row0-1))[N-1][M-1]) == 1;
```

blattms:
:flushed: It is really embarrassing how long this took me. Really sorry about that. Finally tested this a bit and it looks good.

As always thanks for all the work and the persistence and patience.

PR #3803, opened 2022-02-04
[draft] handling THP constraints in solveWellForTesting

GitPaean:
The existing code just runs through the BHP constraints, it potentially open wells limited by THP constraints without checking.

PR #3816, opened 2022-02-17
Set well control to thp or bhp in well testing

totto82:
This is an alternative to #3803

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -418,6 +418,13 @@ namespace Opm
         // keep a copy of the original well state
         const WellState well_state0 = well_state;
         const double dt = ebosSimulator.timeStepSize();
+        const auto& summary_state = ebosSimulator.vanguard().summaryState();
+        const bool has_thp_limit = this->wellHasTHPConstraints(summary_state);
+        if (has_thp_limit)
+            well_state.well(this->indexOfWell()).production_cmode = Well::ProducerCMode::THP;
+        else
+            well_state.well(this->indexOfWell()).production_cmode = Well::ProducerCMode::BHP;
+
         const bool converged = iterateWellEquations(ebosSimulator, dt, well_state, group_state, deferred_logger);
         if (converged) {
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

totto82:
Based on discussion with @GitPaean we agreed to merge this before #3803 since it fixes an urgent issue. I have manually inspected the test failures and they are fine. Note however my comment on ACTIONX_UDQ above. I will soon start the update_data process.

blattms:
Please open a bug for UDQ_ACTIONX with a reference to the model.

totto82:

totto82:
jenkins build this opm-tests=733 please

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -418,6 +418,13 @@ namespace Opm
         // keep a copy of the original well state
         const WellState well_state0 = well_state;
         const double dt = ebosSimulator.timeStepSize();
+        const auto& summary_state = ebosSimulator.vanguard().summaryState();
+        const bool has_thp_limit = this->wellHasTHPConstraints(summary_state);
+        if (has_thp_limit)
+            well_state.well(this->indexOfWell()).production_cmode = Well::ProducerCMode::THP;
+        else
+            well_state.well(this->indexOfWell()).production_cmode = Well::ProducerCMode::BHP;
```
bska, line 426: It's not directly related to this PR, but why do we save a copy of the `WellState`, modify the original object, and then restore the input object instead of taking the `WellState` by reference-to-`const`, modifying a copy of it as needed, and then letting that copy go out of scope when it's no longer needed?  The latter would save one invocation of `WellState`'s assignment operator.

GitPaean: It is a good question. I do not remember how this part was designed originally, with its current way, it looks like the modified/updated `well_state` wants to be used if `converged`. 

But since the control is set forcefully to be `THP` or `BHP` for testing, maybe we should not reuse the updated/modified well state. . 

Then the way you mentioned is the correct way we should use. @bska

PR #3817, opened 2022-02-17
Remove Extraneous Semicolon

bska:
This generates warnings when using GCC's `-Wpedantic` option.


bska:
jenkins build this please

PR #3824, opened 2022-02-24
Avoid switching to thp control if production is decreasing or injection is increasing

totto82:
This means that we may temporary allow a producer to operate with thp below its limit if that gives a higher production. The operability of these wells will be tested along with wells under thp and bhp control.

### DIFF
--- a/opm/simulators/wells/WellInterfaceFluidSystem.cpp
+++ b/opm/simulators/wells/WellInterfaceFluidSystem.cpp
@@ -83,7 +83,8 @@ template <typename FluidSystem>
 Well::ProducerCMode
 WellInterfaceFluidSystem<FluidSystem>::
 activeProductionConstraint(const SingleWellState& ws,
-                           const SummaryState& summaryState) const
+                           const SummaryState& summaryState,
+                           DeferredLogger& deferred_logger) const
 {
     const PhaseUsage& pu = this->phaseUsage();
     const auto controls = this->well_ecl_.productionControls(summaryState);
@@ -162,8 +163,26 @@ activeProductionConstraint(const SingleWellState& ws,
     if (controls.hasControl(Well::ProducerCMode::THP) && currentControl != Well::ProducerCMode::THP) {
         const auto& thp = getTHPConstraint(summaryState);
         double current_thp = ws.thp;
-        if (thp > current_thp)
-            return Well::ProducerCMode::THP;
+        if (thp > current_thp) {
+            bool rate_less_than_potential = true;
+            for (int p = 0; p < number_of_phases_; ++p) {
+                // Currently we use the well potentials here computed before the iterations.
+                // We may need to recompute the well potentials to get a more
+                // accurate check here.
+                rate_less_than_potential = rate_less_than_potential && (-ws.surface_rates[p]) <= ws.well_potentials[p];
+            }
+            if(!rate_less_than_potential) {
+                this->operability_status_.thp_limit_violated_but_not_switched = false;
+                return Well::ProducerCMode::THP;
+            } else {
+                this->operability_status_.thp_limit_violated_but_not_switched = true;
+                deferred_logger.debug("NOT_SWITCHING_TO_THP",
+                "The THP limit is violated for producer " +
+                this->name() +
+                ". But the rate will increase if switched to THP. " +
+                "The well is therefore kept at " + Well::ProducerCMode2String(currentControl));
+            }
+        }
     }
 
     return currentControl;
@@ -174,7 +193,8 @@ template <typename FluidSystem>
 Well::InjectorCMode
 WellInterfaceFluidSystem<FluidSystem>::
 activeInjectionConstraint(const SingleWellState& ws,
-                          const SummaryState& summaryState) const
+                          const SummaryState& summaryState,
+                          DeferredLogger& deferred_logger) const
 {
     const PhaseUsage& pu = this->phaseUsage();
 
@@ -238,8 +258,26 @@ activeInjectionConstraint(const SingleWellState& ws,
     {
         const auto& thp = getTHPConstraint(summaryState);
         double current_thp = ws.thp;
-        if (thp < current_thp)
-            return Well::InjectorCMode::THP;
+        if (thp < current_thp) {
+            bool rate_less_than_potential = true;
+            for (int p = 0; p < number_of_phases_; ++p) {
+                // Currently we use the well potentials here computed before the iterations.
+                // We may need to recompute the well potentials to get a more
+                // accurate check here.
+                rate_less_than_potential = rate_less_than_potential && (ws.surface_rates[p]) <= ws.well_potentials[p];
+            }
+            if(!rate_less_than_potential) {
+                this->operability_status_.thp_limit_violated_but_not_switched = false;
+                return Well::InjectorCMode::THP;
+            } else {
+                this->operability_status_.thp_limit_violated_but_not_switched = true;
+                deferred_logger.debug("NOT_SWITCHING_TO_THP",
+                "The THP limit is violated for injector " +
+                this->name() +
+                ". But the rate will increase if switched to THP. " +
+                "The well is therefore kept at " + Well::InjectorCMode2String(currentControl));
+            }
+        }
     }
 
     return currentControl;
@@ -249,10 +287,11 @@ template <typename FluidSystem>
 bool
 WellInterfaceFluidSystem<FluidSystem>::
 checkIndividualConstraints(SingleWellState& ws,
-                           const SummaryState& summaryState) const
+                           const SummaryState& summaryState,
+                           DeferredLogger& deferred_logger) const
 {
     if (this->well_ecl_.isProducer()) {
-        auto new_cmode = this->activeProductionConstraint(ws, summaryState);
+        auto new_cmode = this->activeProductionConstraint(ws, summaryState, deferred_logger);
         if (new_cmode != ws.production_cmode) {
             ws.production_cmode = new_cmode;
             return true;
@@ -260,7 +299,7 @@ checkIndividualConstraints(SingleWellState& ws,
     }
 
     if (this->well_ecl_.isInjector()) {
-        auto new_cmode = this->activeInjectionConstraint(ws, summaryState);
+        auto new_cmode = this->activeInjectionConstraint(ws, summaryState, deferred_logger);
         if (new_cmode != ws.injection_cmode) {
             ws.injection_cmode = new_cmode;
             return true;
@@ -441,7 +480,7 @@ checkConstraints(WellState& well_state,
                  const SummaryState& summaryState,
                  DeferredLogger& deferred_logger) const
 {
-    const bool ind_broken = checkIndividualConstraints(well_state.well(this->index_of_well_), summaryState);
+    const bool ind_broken = checkIndividualConstraints(well_state.well(this->index_of_well_), summaryState, deferred_logger);
     if (ind_broken) {
         return true;
     } else {
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -262,7 +262,7 @@ namespace Opm
         }
         bool changed = false;
         if (iog == IndividualOrGroup::Individual) {
-            changed = this->checkIndividualConstraints(ws, summaryState);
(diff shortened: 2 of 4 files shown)


totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
@GitPaean I have updated the PR with some more debug and comments in the code following your recommendations via private communication. Plan to start the merge process later today.

totto82:
jenkins build this please

totto82:
jenkins build this please update_data please

opm/simulators/wells/WellInterfaceFluidSystem.cpp
```diff
@@ -162,8 +163,26 @@ activeProductionConstraint(const SingleWellState& ws,
...
         double current_thp = ws.thp;
-        if (thp > current_thp)
-            return Well::ProducerCMode::THP;
+        if (thp > current_thp) {
+            bool decrease = true;
+            for (int p = 0; p < number_of_phases_; ++p) {
+                // Currently we use the well potentials here computed before the iterations.
+                // We may need to recompute the well potentials to get a more
+                // accurate check here.
+                decrease = decrease && (-ws.surface_rates[p]) <= ws.well_potentials[p];
```
GitPaean, line 172: Is the `surface_rates` and `well_potentials` are of different signs here?

Anyway, the logic is that if switching to THP control will result in higher production, we do not switch to THP control. 

As a result, to make the logic easier and code easier to read, we should compare the absolute values and check whether the rate increases. And we discussed the variable should be named `increase` and the logic is the same for injectors and producers.

totto82: > Is the surface_rates and well_potentials are of different signs here?

Yes. From producers surface_rates are negative, while well_potentials are always positive. 


>Anyway, the logic is that if switching to THP control will result in higher production, we do not switch to THP control.
>As a result, to make the logic easier and code easier to read, we should compare the absolute values and check whether the rate increases. And we discussed the variable should be named increase and the logic is the same for injectors and producers.

I have though a bit more about it, and I prefer the way it is now. 1) I don't think we should check absolute values here. The direction of the flow matters. 2) I don't think reversing the logic and changing the variable name will increase the readability of the code significantly here. I instead added some more comments. I hope that is sufficient.

totto82:
jenkins build this opm-tests=738 please

opm/simulators/wells/WellInterfaceFluidSystem.cpp
```diff
@@ -236,10 +256,28 @@ activeInjectionConstraint(const SingleWellState& ws,
...
-            return Well::InjectorCMode::THP;
+        if (thp_limit < current_thp) {
+            bool increase = true;
+            for (int p = 0; p < number_of_phases_; ++p) {
+                // Currently we use the well potentials here computed before the iterations.
+                // We may need to recompute the well potentials to get a more
+                // accurate check here.
+                increase = increase && (ws.surface_rates[p]) >= ws.well_potentials[p];
+            }
+            if(!increase) {
```
GitPaean, line 269: `increase==true` means well injects less under THP control, we should switch to THP control. Can you double check this logic here. 

The reason we do not want to switch to THP control is because the current control is the more restrictive one than THP control, means the current inject rates is smaller than the rates under THP target (`increase == false`).

GitPaean: And also, for injectors, should we compare the injection phase in case there are some noise for the other near-zero values?

opm/simulators/wells/WellInterfaceFluidSystem.cpp
GitPaean: Sure. It is personal preference. My idea is that the code for injectors and producers can basically be almost same. Currently with two different logical way, it takes more time to understand the two sets of code and can be prune to mistakes. 

And here, you do not want some near-zero noisy rates play important roles in determining the simulation behavoir, maybe we should discard those comparisons, either  by comparing the rates of the most important phase we are looking at, or the weighted sum of all rates, or only discarding small values. I do not have an idea which is the best.

opm/simulators/wells/WellInterfaceFluidSystem.cpp
totto82: > increase==true means well injects less under THP control, we should switch to THP control. Can you double check this logic here.

The logic for an injector is the opposite of what you have for an producer. That is reflected in the code as it is. 

> And also, for injectors, should we compare the injection phase in case there are some noise for the other near-zero values?

Your point is that even if the rate of the injection phase is increasing there may be oscillation(production) in the other phases (i.e. ws.surface_rates[p] < 0) that could allow for switching to thp. That certainly is possible, but I have not seen it in the cases I have looked at. I can change the code accordingly, by adding switch(injectorType).... case.... around the check. Or we can cross that bridge when we come there. It is hard to guess what is the robust logic here without any cases where the feature is exposed.

opm/simulators/wells/WellInterfaceFluidSystem.cpp
totto82: >  I do not have an idea which is the best.

I don't either. And I don't think we will know before we find a case where the issue you point to is exposed.

opm/simulators/wells/WellInterfaceFluidSystem.cpp
GitPaean: > The logic for an injector is the opposite of what you have for an producer. That is reflected in the code as it is.

But the well control logic is that we stay at the more restrictive control mode (produce or inject less). If you agree with this logic, then the code for producers and injectors should be the same. And the code for the injectors is basically saying the opposite to the logic.

opm/simulators/wells/WellInterfaceFluidSystem.cpp
GitPaean: > I don't either. And I don't think we will know before we find a case where the issue you point to is exposed.

Yeah. That might cause confusion during debugging.  As a first step, maybe we can discard the very small rates first, or using the weighted sum (basically discard the effects of the very small rates), or only check the targeting phase?

The content in this PR should only be checking between the rate controls and THP control following the document of WVFPEXP, so there should be a targeting phase.

Sorry for being so detailed, but it is good the code can be more finely written based on the existing knowledge.  If you disagree, it is okay to agree to disagree.

totto82: My aim with this PR was not to implement item 4 = yes in WVFPEXP but rather fix an issue that happens when a group controlled well is controlled by group rate equal to zero. The case does not have WVFPEXP and OPM Flow does not support that feature yet. 

`    GR1     ORAT      0.0    1*	    1*    1*    RATE	  YES	  1*	'   '	1*	1*	1*	1*	1*	/
`

and C-1H is under GR1. 

image
 

With master control will switch between THP and GRUP since Rate THP(rate = 0) violates the thp control of 30. 

`Well C-1H control mode changed from THP->GRUP->THP->GRUP->THP`

image

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -600,12 +600,14 @@ namespace Opm
...
         // Only check wells under BHP and THP control
-        if(bhp_controled || thp_controled) {
+        bool check_thp = thp_controled || this->operability_status_.thp_limit_violated_but_not_switched;
+        if (check_thp || bhp_controled) {
             updateIPR(ebos_simulator, deferred_logger);
             checkOperabilityUnderBHPLimit(well_state, ebos_simulator, deferred_logger);
         }
         // we do some extra checking for wells under THP control.
-        if (thp_controled) {
+        const auto& summaryState = ebos_simulator.vanguard().summaryState();
```
bska, line 609: This `summaryState` object isn't used anywhere.  Is it supposed to be?  If not, we should not introduce it.

totto82: I did some final cleanup and forgot to remove it. Thanks for noticing. I will update the code accordingly.

bska: > I did some final cleanup and forgot to remove it.

Ah, I see.

> Thanks for noticing.

I had a lot of help from the [build check](https://ci.opm-project.org/job/opm-simulators-PR-builder/3540/gcc/).

opm/simulators/wells/WellInterfaceFluidSystem.cpp
GitPaean: > My aim with this PR was not to implement item 4 = yes in WVFPEXP but rather fix an issue that happens 

Yeah. As mentioned above, the logic for the injectors and producers are opposite in this PR, we really hesitate to say yes to that.  And also, for injectors, if we introduce the opposite logic with WVFPEXP, then we will face hug problem when we want to have support for WVFPEXP. And as we discussed, this keyword will take effects even without using it in the commercial simulator.  

> THP and GRUP since Rate THP(rate = 0) 

Then it comes to the question whether we should check THP values with `rate==0`, whether we should do extrapolation for `rate==0`, and also the rate is already zero, maybe we should not switch to any other controls because `rate==0` is the most restrictive control already if the well still operable. If the well is not operable, we should just shut it.  

Whether VFP table is still valid when `rate==0`, I am not expert, but I think it is not designed for those region. Maybe @atgeirr @tskille have some comments regarding that. 

But your fix for the producers part can temporarily fix the problem. My argument is that whether we should introduce an opposite logic for injectors as mentioned in the comments.  


And for the examples your posted figures with, what is the THP values from the reference results from the commercial simulator?

totto82:
jenkins build this please

opm/simulators/wells/WellInterfaceFluidSystem.cpp
totto82: I think I finally got your point. I have update the PR accordingly. Still doing some testing, to see the effect of the change for the injector. My main test case only expose the production part.

totto82:
jenkins build this please

totto82:
The model 2 failures are very small, the udq ones are larger and needs manual investigation 

image

image

The differences are still small enough to mandate an update.

totto82:
@atgeirr @GitPaean Can you confirm that our understanding of the issue and the implementation all are aligned now?

opm/simulators/wells/WellInterfaceFluidSystem.cpp
atgeirr: I was very confused now. I think the "increase" variable name is very unfortunate.

I think the code itself is correct now: If the current rate is less than the potential, then `increase` will be true (this is very weird... call it `rate_less_than_potential` to be clearer). If `increase` is true we will enter the `else` branch and not switch.

totto82: the logic with the naming from my side was that if rate < pot the rate will increase if we switch to thp. i.e. only switch if !increase. `rate_less_than_potential`  is a more descriptive name. I will update accordingly.

totto82:
jenkins build this please

totto82:
jenkins build this please

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -600,12 +600,13 @@ namespace Opm
...
+        bool check_thp = thp_controled || this->operability_status_.thp_limit_violated_but_not_switched;
+        if (check_thp || bhp_controled) {
             updateIPR(ebos_simulator, deferred_logger);
             checkOperabilityUnderBHPLimit(well_state, ebos_simulator, deferred_logger);
         }
         // we do some extra checking for wells under THP control.
-        if (thp_controled) {
+        if (check_thp) {
             checkOperabilityUnderTHPLimit(ebos_simulator, well_state, deferred_logger);
         }
```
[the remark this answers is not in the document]
totto82: Testing done and code updated.

totto82:
@GitPaean @atgeirr Can this be merged now? Thanks for persevering review @GitPaean I finally got your point and we avoided a bug creeping into the code.

atgeirr:
> @GitPaean @atgeirr Can this be merged now? Thanks for persevering review @GitPaean I finally got your point and we avoided a bug creeping into the code.

Yes it can be merged when green.

totto82:

totto82:
jenkins build this opm-tests=738 please

bska:
The updated reference solutions have been installed on the CI system.  Merging this now per prior approval and to bring the CI system back to a consistent state.

PR #3826, opened 2022-02-25
Don't Capture Local Flux Data Unless Needed

bska:
This is a potentially expensive operation and so we should not run it unless such values are actually needed for output purposes (e.g., inter-region flow summary vectors).

### DIFF
--- a/ebos/eclgenericoutputblackoilmodule.hh
+++ b/ebos/eclgenericoutputblackoilmodule.hh
@@ -186,6 +186,11 @@ public:
         return 0;
     }
 
+    bool needInterfaceFluxes([[maybe_unused]] const bool isSubStep) const
+    {
+        return this->interRegionFlows_.wantInterRegflowSummary();
+    }
+
     const std::map<std::size_t, double>& getWBPData() const
     {
         return this->wbpData_;
--- a/ebos/eclinterregflows.hh
+++ b/ebos/eclinterregflows.hh
@@ -274,6 +274,11 @@ namespace Opm {
         /// Whether or not previous read() operation succeeded.
         bool readIsConsistent() const;
 
+        bool wantInterRegflowSummary() const
+        {
+            return !this->regionMaps_.empty();
+        }
+
         /// Serialise internal representation to MPI message buffer
         ///
         /// \tparam MessageBufferType Linear MPI message buffer.  API should
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -187,7 +187,9 @@ public:
         const auto localWellTestState = simulator_.problem().wellModel().wellTestState();
         this->prepareLocalCellData(isSubStep, reportStepNum);
 
-        this->captureLocalFluxData();
+        if (this->eclOutputModule_->needInterfaceFluxes(isSubStep)) {
+            this->captureLocalFluxData();
+        }
 
         if (this->collectToIORank_.isParallel()) {
             OPM_BEGIN_PARALLEL_TRY_CATCH()


bska:
jenkins build this please

bska:
benchmark this please

atgeirr:
I did a run of Norne with and without the call to captureLocalFluxData() commented out (i.e. not exactly same as this PR), with total times of 453.99s (case A, without) vs. 470.85s (case B, with). The more detailed numbers were skewing slightly towards B being faster by a few seconds (within normal laptop variation I would say, but the "Pre/post step" times were 40.78s vs. 64.16s. So I would say it clearly has a large cost!

bska:
> the "Pre/post step" times were 40.78s vs. 64.16s

46.5 vs. 52 on my system.

atgeirr:
I ran again in parallel (previous runs were serial), and pre/post was now 17.17s vs 21.1s. However total times were almost identical due to countervariations in the other numbers. This was with CPR which is nondeterministic in parallel (i.e. get different iteration numbers etc.). I will rerun with ILU0 to have a more sure result.

atgeirr:
With ILU0 the runs are deterministic (identical iteration numbers etc.), still we see a significant cost to the captureLocalFluxData(). Without:
```
Number of MPI processes:         8
Threads per MPI process:         1
Total time (seconds):          211.15 
Solver time (seconds):         211.06 
 Assembly time (seconds):       56.99 (Failed: 0.0; 0.0%)
   Well assembly (seconds):      1.82 (Failed: 0.0; 0.0%)
 Linear solve time (seconds):  105.85 (Failed: 0.0; 0.0%)
   Linear setup (seconds):      15.28 (Failed: 0.0; 0.0%)
 Update time (seconds):         22.14 (Failed: 0.0; 0.0%)
 Pre/post step (seconds):       17.66 (Failed: 0.0; 0.0%)
 Output write time (seconds):    6.22
Overall Linearizations:       1809    (Failed:   0; 0.0%)
Overall Newton Iterations:    1443    (Failed:   0; 0.0%)
Overall Linear Iterations:   21985    (Failed:   0; 0.0%)
```
With:
```
Number of MPI processes:         8
Threads per MPI process:         1
Total time (seconds):          217.68 
Solver time (seconds):         217.60 
 Assembly time (seconds):       57.54 (Failed: 0.0; 0.0%)
   Well assembly (seconds):      1.86 (Failed: 0.0; 0.0%)
 Linear solve time (seconds):  106.83 (Failed: 0.0; 0.0%)
   Linear setup (seconds):      15.45 (Failed: 0.0; 0.0%)
 Update time (seconds):         22.26 (Failed: 0.0; 0.0%)
 Pre/post step (seconds):       22.39 (Failed: 0.0; 0.0%)
 Output write time (seconds):    6.34
Overall Linearizations:       1809    (Failed:   0; 0.0%)
Overall Newton Iterations:    1443    (Failed:   0; 0.0%)
Overall Linear Iterations:   21985    (Failed:   0; 0.0%)
```
All numbers are close, except the pre/post.

bska:
> All numbers are close, except the pre/post.

So that's a 26.8% increase in pre/post and a 3.1% in "total" time.  That's enough to justify this PR.  I'll nevertheless await the official benchmarking results before deciding to merge.

blattms:
Seems like this was not picked up.

blattms:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1.029    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 1.02     |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 1.058    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 1.014    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 1.038    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 1.024    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1536

bska:
Benchmark results are in.  There is greater benefit to avoiding this pass for sequential runs than for parallel runs, but non-negligible in all cases.  I'll mark this as "ready for review" with an intent to merge.

blattms:
Consistent with the run done in the offending PR: Seems like we need to find the real cause for the performance regression then. 
Have you benchmarked OPM/opm-common#2945. It was not obvious that it might have impact, but it falls into the time where the regression seems to occur (before 2022-02-18 1:00).

blattms:
And as it turns out, I might have interpreted the dates used on https://www.linuxbenchmarking.com/?open-porous-media-git wrong or misjudged elsewhere. This fixed the regression and I was wrong. Sorry for the noise any additional worked caused by it.

Seems like the impact differs quite a bit between the architectures (3-7%)

atgeirr:
> Seems like the impact differs quite a bit between the architectures (3-7%)

That is quite interesting, and it also points to the cost of assembly varying in the same manner, and taking significant time. Essentially, the code run is a bit like "doing the assembly, but not assembling any equations". We should investigate ways to reduce the cost of the ElementContext-centric approach.

PR #3828, opened 2022-03-01
Report Unique IJK Tuples for Failed Bubble/Dew Point Calculations

bska:
This PR switches the debug file's records of "failed" bubble and dew point pressure calculations from a non-unique list of linearised Cartesian indices to a unique list of (1-based) (I,J,K) tuples. This format is hopefully easier to read for humans.

Example messages from selected time steps in a fieldcase with Cartesian dimensions 137-by-236-by-58:

  - Original
  ```
  Finding the dew point pressure failed for 2 cells [1467066, 1467066]
  Finding the dew point pressure failed for 8 cells [1467063, 1467063, 1467066, 1467066, 1467066, 1467066, 1467066, 1467066]
  ```

  - This PR
  ```
  Finding the dew point pressure failed for 1 cell [(71,89,46)]
  Finding the dew point pressure failed for 2 cells [(68,89,46), (71,89,46)]
  ```


bska:
jenkins build this please

alfbr:
I support this change. Giving coordinates is an improvement.

ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -1440,64 +1441,112 @@ template<class FluidSystem,class Scalar>
 void EclGenericOutputBlackoilModule<FluidSystem,Scalar>::
 outputErrorLog(const Comm& comm) const
 {
-    const size_t maxNumCellsFaillog = 20;
+    const auto pbSize = static_cast<int>(this->failedCellsPb_.size());
+    const auto pdSize = static_cast<int>(this->failedCellsPd_.size());
```
blattms, line 1445: While changing this anyway, maybe we should resort to an utiility function that I implemented for doing this to reduce code:

```
#include<opm/grid/common/CommunicationUtils.hpp>
....
const auto int root = 0;
auto [ globalFailedCellsPb, displPb] = gatherv(this->failedCellsPb_, root);
auto [ globalFailedCellsPd, displPd] = gatherv(this->failedCellsPd_, root);
....
```

This functionality is of course too well hidden in opm-grid...

ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -1440,64 +1441,112 @@ template<class FluidSystem,class Scalar>
...
+               << ((numFailed != std::size_t{1}) ? "s" : "")
+               << " [" << ijkString(globalFailedCellsPb[0]);
+
+        const auto maxElems = std::min(maxNumCellsFaillog, numFailed);
+        for (auto i = 1 + 0*maxNumCellsFaillog; i < maxElems; ++i) {
+            errlog << ", " << ijkString(globalFailedCellsPb[i]);
         }
-        if (globalFailedCellsPb.size() > maxNumCellsFaillog) {
+
+        if (numFailed > maxNumCellsFaillog) {
```
blattms, line 1517: One could save one iteration over the vector and some memory copying  by first printing the cell indices (and simply skipping multiple) while keeping count of the number and then the total number.

ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -1440,64 +1441,112 @@ template<class FluidSystem,class Scalar>
...
+               << numFailed << " cell"
+               << ((numFailed != std::size_t{1}) ? "s" : "")
+               << " [" << ijkString(globalFailedCellsPd[0]);
+
+        const auto maxElems = std::min(maxNumCellsFaillog, numFailed);
+        for (auto i = 1 + 0*maxNumCellsFaillog; i < maxElems; ++i) {
+            errlog << ", " << ijkString(globalFailedCellsPd[i]);
         }
-        if (globalFailedCellsPd.size() > maxNumCellsFaillog) {
+
```
blattms, line 1543: Isn't this the very same algorithm as above for globalFailedCellsPb? Why not factor that out into a method and call that twice to reduce code duplication?

blattms (commented):
While at it, it might be good to reduce the code duplication here a bit.

ebos/eclgenericoutputblackoilmodule.cc
bska: That's a good tip.  It does have the same undefined behaviour of [forming](https://github.com/OPM/opm-grid/blob/fa96a96fa2a8bc89d74985ff5a2203a99841ef87/opm/grid/common/CommunicationUtils.hpp#L87) `displ.begin() + 1` on non-root processes as this code used to have, but I suppose we can fix it there too.

ebos/eclgenericoutputblackoilmodule.cc
bska: > One could save one iteration over the vector and some memory copying by first printing the cell indices (and simply skipping multiple) while keeping count of the number and then the total number.

It's a tradeoff.  These arrays **typically** don't have more than 300 elements including duplicates so I don't expect `std::unique()` to be a big cost, but I'll sketch out a proposal.  For what it's worth, the largest arrays I ever saw on the model I used for testing this was 350 elements with the common case being less than 100.

ebos/eclgenericoutputblackoilmodule.cc
bska: > Why not factor that out into a method and call that twice to reduce code duplication?

I'll do that but I won't add a new member function.

ebos/eclgenericoutputblackoilmodule.cc
blattms: Yep, you are right. It is undefined behavior, also the documentation is not 100% crystal clear (iterator must be in derefenceable state before increment.

bska: > Maybe use displ.begin() + (rank==root) or displ.begin() + (displ.begin() != displ.end()), then.

Or just not run the `partial_sum()` on non-root processes since by construction `displ` is empty on those.  I'd probably implement `gatherv()` along the lines of
```C++
// ...
comm.gather(&mySize, sizes.data(), 1, root);
std::vector<T,A> output;
if (isRoot) {
    std::partial_sum(sizes.begin(), sizes.end(), displ.begin() + 1);
    output.resize(displ.back(), T{});
}
comm.gatherv(...);
return { output, displ };
```

bska:
I've pushed an update that switches to using the `gatherv()` helper function to simplify the communication and that extracts the common logging parts out to a new helper function.  In the end I decided to keep the call to `std::unique()` because I like the ability to report the number of unique failing cells before outputting their cell IDs.  Doing so requires making a pass over the collection regardless and then I considered the extra cost of copying a typically small number of integers to be acceptable.

bska:
jenkins build this please

ebos/eclgenericoutputblackoilmodule.cc
blattms: I'll back off.

blattms:
Thanks a lot for the changes. Merging...

PR #3831, opened 2022-03-08
Add gaslift_opt_active flag to VFPProdTable constructor

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/2971


joakim-hove:
Replaced by #3833

PR #3834, opened 2022-03-09
Handle case where bhp target is violated

totto82:
This fixes a particular issue exposed while running a field model.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -975,7 +975,10 @@ namespace Opm
             // to taking into consideration the crossflow here.
             if ( (this->isProducer() && pressure_diff < 0.) || (this->isInjector() && pressure_diff > 0.) ) {
                 deferred_logger.debug("CROSSFLOW_IPR",
-                                "cross flow found when updateIPR for well " + name());
+                                "cross flow found when updateIPR for well " + name()
+                                + " . The connection is ignored in IPR calculations");
+                // we ignore these connections for now
+                continue;
             }
 
             // the well index associated with the connection
@@ -1871,7 +1874,19 @@ namespace Opm
         const auto& summaryState = ebosSimulator.vanguard().summaryState();
         if (!Base::wellHasTHPConstraints(summaryState) || bhp_controlled_well) {
             // get the bhp value based on the bhp constraints
-            const double bhp = this->mostStrictBhpFromBhpLimits(summaryState);
+            double bhp = this->mostStrictBhpFromBhpLimits(summaryState);
+
+            // In some very special cases the bhp pressure target are
+            // temporary violated. This may lead to too small or negative potentials
+            // that could lead to premature shutting of wells.
+            // As a remedy the bhp that gives the largest potential is used.
+            // For converged cases, ws.bhp <=bhp for injectors and ws.bhp >= bhp,
+            // and the potentials will be computed using the limit as expected.
+            if (this->isInjector())
+                bhp = std::max(ws.bhp, bhp);
+            else
+                bhp = std::min(ws.bhp, bhp);
+
             assert(std::abs(bhp) != std::numeric_limits<double>::max());
             computeWellRatesWithBhpIterations(ebosSimulator, bhp, well_potentials, deferred_logger);
         } else {


totto82:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -1871,7 +1873,16 @@ namespace Opm
         const auto& summaryState = ebosSimulator.vanguard().summaryState();
         if (!Base::wellHasTHPConstraints(summaryState) || bhp_controlled_well) {
             // get the bhp value based on the bhp constraints
-            const double bhp = this->mostStrictBhpFromBhpLimits(summaryState);
+            double bhp = this->mostStrictBhpFromBhpLimits(summaryState);
+
+            // in some very special cases the bhp pressure target are
+            // allowed to be violated. This may lead to negative potensials
+            // if not properly accounted for as done here.
```
akva2, line 1880: potentialS

totto82:
The test failures (all of them) are due to unexpected closing of well in current master. 
< well WI_1: has negative potentials and is not operable
< Well WI_1 will be shut as it can not operate under current reservoir conditions.
With this PR the well is kept open. 
image

The symptoms exposed by the test cases are the same as what is seen in the field model. 

I.e. from my point of view we should update the data and merge this. (I will fix the spelling mistake first..)

totto82:
jenkins build this please

GitPaean:
Sorry I have not seen this. I will review it soon (by tomorrow). If you have more pull requests waiting for me, please let me know.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -1872,7 +1874,16 @@ namespace Opm
         const auto& summaryState = ebosSimulator.vanguard().summaryState();
         if (!Base::wellHasTHPConstraints(summaryState) || bhp_controlled_well) {
             // get the bhp value based on the bhp constraints
-            const double bhp = this->mostStrictBhpFromBhpLimits(summaryState);
+            double bhp = this->mostStrictBhpFromBhpLimits(summaryState);
+
+            // in some very special cases the bhp pressure target are
```
GitPaean, line 1879: Can you explain more about under what special cases, the BHP constraint can be violated? I have been always thinking  BHP constraint is a pretty hard constraint. 

And also, have you seen the situations that negative well potentials are resulted?

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -976,6 +976,8 @@ namespace Opm
             if ( (this->isProducer() && pressure_diff < 0.) || (this->isInjector() && pressure_diff > 0.) ) {
                 deferred_logger.debug("CROSSFLOW_IPR",
                                 "cross flow found when updateIPR for well " + name());
+                // we ignore these connections for now
+                continue;
```
GitPaean, line 980: Can you explain a little more here also about the exact purpose?

And also, a message here (warning or debug) telling that the crossflow is ignored here.

opm/simulators/wells/StandardWell_impl.hpp
totto82: We compute the well potentials before we check the constrains and also we don't check the constrains in our getWellConverence method. Negative potentials can happen for new wells for instance. Say you open a injector that is supposed to give pressure support to your producer. The solution will need some newton iterations before the pressure increase due to injection is propagated to the producer.

opm/simulators/wells/StandardWell_impl.hpp
totto82: I think the IPR formulation handle crossflow at the moment. I guess that is the reason for giving a debug output here.. But since the formulation do not handle crossflow adding these connections may lead to premature closing of wells. 
I can update the debug message with " ... The connection is ignored." or something similar.

opm/simulators/wells/StandardWell_impl.hpp
GitPaean: Then there is a problem here. With the code, you are computing the well potentials based on the old bhp value fromthe well state instead of the BHP constraint. Will it be against the definition of the well potentials? or I misunderstood it?

totto82: Yes it will be against the definition, but it is only a temporary solution. When the simulator starts to iterate the bhp limit will be violated and ws.bhp = bhp and we are on track again. The point of this temporary solution is to be able to keep the well open even if it initially (i.e. before the newton loop) thinks it does not have potential to keep open it may get the potential it need to produce when the injector has done its job of increasing the pressure. I have tested this on several model and the results seems to be ok, also for the potentials. Even a slightly wrong solution for the potentials is better than shutting the well completely if it is supposed to stay open.

GitPaean: I am a little too slow on this and need a description of the scenario this fix is targeting. Maybe we will be able to discuss it in person tomorrow.

totto82:
jenkins build this please

GitPaean:
> @GitPaean I have updated the PR based on your feedback. I plan to start the merging process later today.

I still have difficulties in understanding why the changes help. But in the meeting @atgeirr  together, we agreed that it can go in as a temporary fix for the running of the target case.   Sure, go ahead.

totto82:
> Sure, go ahead. 

Thanks. As the test cases shows this fixes a real problem. I will start the merge process.

totto82:

totto82:
jenkins build this opm-tests=739 please

Issue #2029, opened 2019-09-29
Cannot Initialize Dry Gas Run


OPMUSER:
One cannot initialize a dry gas run because the the RVCONST and the RVCONSTT keywords are not supported, even though the dry gas PVDG table is recognized.

Is this a known issue?

Test deck attached.
[GASW-BASE01-OPM1904-R01.zip](https://github.com/OPM/opm-simulators/files/3666454/GASW-BASE01-OPM1904-R01.zip)

quagdoo:
I can verify that this is the case - I converted the Norne case and got an exception and very little info to check if I have made a mistake in the first place.

`Program threw an exception: Could not initialize the problem.`

In a Gas-Water system, the documentation says that RVCONST(T) are just for reporting so I suspect the fact that it doesn't initialise is a bit of bigger problem than these keywords. I've included my DATA deck. I guess the work-around is just use more physics and longer run times!

[NORNE_ATW2013_PVDG.DATA.gz](https://github.com/OPM/opm-simulators/files/4138236/NORNE_ATW2013_PVDG.DATA.gz)

quagdoo:
The manual says the following is green (implemented):

> 8.3.200 PVDO – OIL PVT PROPERTIES FOR DEAD OIL

Could you provide a working example of a case where this works? Just so I can test it.

bska:
### PVDO Example Models ###

The following models in [opm-tests](https://github.com/OPM/opm-tests/tree/ec31f391d6ca8ebf3f251084bc5cdaf89fbddd36) use `PVDO`:

* [`SPE1CASE2_2P.DATA`](https://github.com/OPM/opm-tests/blob/ec31f391d6ca8ebf3f251084bc5cdaf89fbddd36/spe1/SPE1CASE2.DATA)
* [`SPE10_MODEL1.DATA`](https://github.com/OPM/opm-tests/blob/ec31f391d6ca8ebf3f251084bc5cdaf89fbddd36/spe10model1/SPE10_MODEL1.DATA)
* [`SPE10_MODEL1_CP.DATA`](https://github.com/OPM/opm-tests/blob/ec31f391d6ca8ebf3f251084bc5cdaf89fbddd36/spe10model1/SPE10_MODEL1_CP.DATA)
[...]

PR #3838, opened 2022-03-11
fix initialization EQUIL + water and gas

totto82:
Fix for https://github.com/OPM/opm-simulators/issues/2029

### DIFF
--- a/ebos/equil/initstateequil.hh
+++ b/ebos/equil/initstateequil.hh
@@ -626,13 +626,20 @@ equil_WOG(const Region& reg, const VSpan& span)
         this->makeOilPressure(ic, reg, span);
     }
 
-    if (this->gasActive()) {
+    if (this->gasActive() && this->oilActive()) {
         // Pcgo = Pg - Po => Pg = Po + Pcgo
         const auto ic = typename GPress::InitCond {
             reg.zgoc(),
             this->oil(reg.zgoc()) + reg.pcgoGoc()
         };
 
+        this->makeGasPressure(ic, reg, span);
+    } else if (this->gasActive() && !this->oilActive()) {
+        // No oil phase set Pg = Pw + Pcgw
+        const auto ic = typename GPress::InitCond {
+            reg.zwoc(), // The WOC is really the GWC for gas/water cases
+            this->water(reg.zwoc()) + reg.pcowWoc() // Pcow(WOC) is really Pcgw(GWC) for gas/water cases
+        };
         this->makeGasPressure(ic, reg, span);
     }
 }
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

ebos/equil/initstateequil.hh
```diff
@@ -669,13 +669,20 @@ equil_GOW(const Region& reg, const VSpan& span)
...
+    if (this->waterActive() && this->oilActive()) {
         // Pcow = Po - Pw => Pw = Po - Pcow
         const auto ic = typename WPress::InitCond {
             reg.zwoc(),
             this->oil(reg.zwoc()) - reg.pcowWoc()
         };
 
+        this->makeWatPressure(ic, reg, span);
+    } else if (this->waterActive() && !this->oilActive()) {
+        // No oil phase set Pw = Pg???
```
bska, line 681: In this case we **probably** want the gas/water capillary pressure.  Since we define
* Pcgo = Pg - Po
* Pcow = Po - Pw

we can calculate
```
Pcgw = Pg - Pw = Pcgo + Pcow
```
from which we get
```
Pw = Pg - (Pcgo + Pcow)
```

totto82: Thanks. I will update. In the test case provided by @OPMUSER  pcgo and pcow was both 0.0.

totto82:
jenkins build this please

ebos/equil/initstateequil.hh
```diff
@@ -669,13 +669,20 @@ equil_GOW(const Region& reg, const VSpan& span)
...
             reg.zwoc(),
             this->oil(reg.zwoc()) - reg.pcowWoc()
         };
 
+        this->makeWatPressure(ic, reg, span);
+    } else if (this->waterActive() && !this->oilActive()) {
+        // No oil phase set Pw = Pg - (Pcgo + Pcow)
+        const auto ic = typename WPress::InitCond {
+            reg.zgoc(),
+            this->gas(reg.zgoc() - (reg.pcgoGoc() + reg.pcowWoc()))
```
bska, line 684: Oops, this should probably say
```C++
this->gas(reg.zgoc()) - (reg.pcGoc() + reg.pcowWoc());
```
instead.  Note closing parenthesis on the `this->gas()` call.  I guess there's also an implied assumption that the gas/oil and oil/water contacts are at the same depths, but I'll need to check the details.

totto82: Thanks for noticing. 

I will provide a similar fix for case when OWC/GOC < DATUM.

>I guess there's also an implied assumption that the gas/oil and oil/water contacts are at the same depths, but I'll need to check the details.

Cool. I have only testet with OWC=GOC.

bska: I checked the details and I was wrong before.

In the gas/water case then item 3 of the `EQUIL` keyword is the GWC and item 4 is the gas/water capillary pressure (Pcgw).  In all other cases item 3 is OWC and item 4 is the Pcow.  In other words we should really write
```C++
const auto ic = typename WPress::InitCond {
    reg.zwoc(), // The WOC is really the GWC for gas/water cases
    this->gas(reg.zwoc()) - reg.pcowWoc() // Pcow(WOC) is really Pcgw(GWC) for gas/water cases
};
```
here.  Sorry for adding to the confusion.

On the other hand we may need the `Pcwg = Pcgo + Pcow` relation for when we invert the capillary pressure relations to derive initial saturations.

totto82:
jenkins build this please

ebos/equil/initstateequil.hh
totto82: Thanks for digging in to this. Seems to work as expected for both  WGC < DATUM and WGC > DATUM. Note the test case is without capillary pressure.

bska (approved):

PR #3842, opened 2022-03-21
only output gaslift related output on rank 0

totto82:
The debugging gaslift output is not activated by default so this should not change any behavior of the simulator.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -2189,7 +2189,7 @@ BlackoilWellModelGeneric::
 gliftDebug(const std::string& msg,
            DeferredLogger& deferred_logger) const
 {
-    if (this->glift_debug) {
+    if (this->glift_debug && comm_.rank() == 0 ) {
         const std::string message = fmt::format(
             "  GLIFT (DEBUG) : BlackoilWellModel : {}", msg);
         deferred_logger.info(message);
--- a/opm/simulators/wells/GasLiftGroupInfo.cpp
+++ b/opm/simulators/wells/GasLiftGroupInfo.cpp
@@ -428,7 +428,7 @@ void
 GasLiftGroupInfo::
 displayDebugMessage_(const std::string &msg, const std::string &well_name)
 {
-    if (this->debug) {
+    if (this->debug && this->comm_.rank() == 0) {
         const std::string message = fmt::format(
              "  GLIFT (DEBUG) : Init group info : Well {} : {}",
              well_name, msg);
--- a/opm/simulators/wells/GasLiftStage2.cpp
+++ b/opm/simulators/wells/GasLiftStage2.cpp
@@ -251,7 +251,7 @@ void
 GasLiftStage2::
 displayDebugMessage_(const std::string &msg) const
 {
-    if (this->debug) {
+    if (this->debug  && this->comm_.rank() == 0) {
         const std::string message = fmt::format(
             "  GLIFT2 (DEBUG) : {}", msg);
         this->deferred_logger_.info(message);


totto82:
jenkins build this please

bska (commented):
Will limiting output exclusively to the IO rank imply that some messages get lost, especially for the well cases?

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -2189,7 +2189,7 @@ BlackoilWellModelGeneric::
 gliftDebug(const std::string& msg,
            DeferredLogger& deferred_logger) const
 {
-    if (this->glift_debug) {
+    if (this->glift_debug && comm_.rank() == 0 ) {
```
hakonhagland, line 2192: As I understand this, this happens because all ranks are executing the same `gliftDebug()` call. For example [line 802](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/BlackoilWellModel_impl.hpp#L802) in `BlackoilWellModel_impl.hpp` ? Also, I see that only rank 0 has `terminal_output_`, see [line 422](  https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/BlackoilWellModelGeneric.hpp#L422) in `BlackoilWellModelGeneric.hpp` and [line 46](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/BlackoilWellModel_impl.hpp#L46) in `BlackoilWellModel_impl.hpp`  set to `true`. So then this could alternatively be fixed on [line 802](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/BlackoilWellModel_impl.hpp#L802) like this:
```
        if (terminal_output_ && this->glift_debug) {
            const std::string msg = fmt::format(
                "assemble() : iteration {}" , iterationIdx);
            gliftDebug(msg, local_deferredLogger);
        }
```
That is: adding `terminal_output_` to the `if` condition?

bska: > As I understand, this happens because all ranks are executing the same `gliftDebug()` call.

That may be, but if the code in any way loops over the `well_container_` then the various ranks do not process the same data.  All ranks have the same group, but the `well_container_` only holds the wells that are active on the local rank.  I tried to trace the control flow here but was unable to conclusively ascertain whether or not the `glifDebug()` calls in any way look at the `well_container_`.

hakonhagland: @bska Yes good observation. Maybe we could split `gliftDebug()` into two method like this:
```
void
BlackoilWellModelGeneric::
gliftDebugRank0(const std::string& msg,
           DeferredLogger& deferred_logger) const
{
    if (terminal_output_) {
        gliftDebug(msg, deferred_logger);
    }
}

void
BlackoilWellModelGeneric::
gliftDebug(const std::string& msg,
           DeferredLogger& deferred_logger) const
{
    if (this->glift_debug) {
        const std::string message = fmt::format(
            "  GLIFT (DEBUG) : BlackoilWellModel : {}", msg);
        deferred_logger.info(message);
    }
}
```
then code that loops over `well_container_` should call `gliftDebug()` and code that is executed by all ranks should call `gliftDebugRank0()` ?

totto82: displayDebugMessage_ in GasLiftGroupInfo is called for individual wells, but then the wells are mostly given by group.wells(). I though group.wells() returned all wells under that group independent on if they exist on that rank or not?   
Also I there is code like this
```
    auto itr = this->ecl_wells_.find(well_name);
    if (itr == this->ecl_wells_.end()) {
        // well_name is not present in the well_model's well container
        displayDebugMessage_("Could find well ecl_wells. Skipping.", well_name);
        return false;
    }
```
That will currently with master give a lot of messages with missing wells if the well is not on the current rank. This will still give unexpected output for rank 0 and hide output of missing wells on other ranks.

hakonhagland: @totto82 I will try to locate the places in the code that calls the same debug output on all ranks. I think we should locate these places first, and then fix the code at those places (by adding a check for rank == 0 there) ?

bska: > `displayDebugMessage_` in `GasLiftGroupInfo` is called for individual wells, but then the wells are mostly given by `group.wells()`. I thought `group.wells()` returned all wells under that group independent on if they exist on that rank or not?

That's true.  `Group::wells()` does indeed return the names of all wells directly parented to the group, be they local to/active on the current rank or not.  My earlier comment was specifically about any code that uses `well_container_` as the data source for well information.  If neither `well_container_` nor [`BlackoilWellModelGeneric::wells_ecl_`](https://github.com/OPM/opm-simulators/blob/519b5dd8cc2c6e8a3efc65be6eb998d3c51263c8/opm/simulators/wells/BlackoilWellModelGeneric.hpp#L429) are used as the primary data source for the wells here then it's probably safe to assume that all calls to the debugging facility process the same well and group information.

hakonhagland: Here is another example of how the `displayDebugMessage_()` in `GasLiftGroupInfo.cpp` can be called for all ranks except the one that has the well in the well container:

First all ranks are calling `initGliftEclWellMap()` at [BlackoilWellModel_impl.hpp:890](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/BlackoilWellModel_impl.hpp#L890) but the well map will be different for each rank (since it is extracted from the well container, see [BlackoilWellModel_impl.hpp:1063](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/BlackoilWellModel_impl.hpp#L1063)). Then `initialize()` is called on the `group_info` object see [BlackoilWellModel_impl.hpp:903](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/BlackoilWellModel_impl.hpp#L903), then `initialize()` at [GasLiftGroupInfo.cpp:185](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/GasLiftGroupInfo.cpp#L185) calls `initializeWell2GroupMapRecursive_()` at [GasLiftGroupInfo.cpp:555](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/GasLiftGroupInfo.cpp#L555) which calls `checkDoGasLiftOptimization()` at [GasLiftGroupInfo.cpp:577](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/GasLiftGroupInfo.cpp#L577). So `checkDoGasLiftOptimization()` [GasLiftGroupInfo.cpp:287](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/GasLiftGroupInfo.cpp#L287) is called by all ranks for wells, but then `checkDoGasLiftOptimization()` checks the wells in the `ecl_wells_` well map that is specific for the current rank, see [GasLiftGroupInfo.cpp:294](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/GasLiftGroupInfo.cpp#L294). From this it looks like that the particular `displayDebugMessage_()` call at [GasLiftGroupInfo.cpp:297](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/GasLiftGroupInfo.cpp#L297) will be called for all ranks that does not have `well_name` in its well container?

Issue #3770, opened 2022-01-13
Master: Dune Processing Grid Error


OPMUSER:
Trying to run an actual model and got this error:

```
Processing grid
flow: /home/david/OPM/GitHub/opm-grid/opm/grid/common/GeometryHelpers.hpp:151: double Dune::GeometryHelpers::polygonCellVolume(const Vector<Point>&, const Point&, const Point&) [with Point = Dune::FieldVector<double, 3>; Vector = Dune::{anonymous}::IndirectArray]: Assertion `small_volume >= 0' failed.
Aborted (core dumped)
Process Complete (0) 
```
Unfortunately, I cannot share the model, but any pointers would be warmly appreciated.

atgeirr:
At first glance it looks like the grid may be sufficiently twisted to make the "normal" volume calculation fail. For Flow, this volume is ignored in favour of a different calculation for compatibility (that calculation in turn is based on some questionable assumptions, which is why it is not universally used), so maybe we could find a quick workaround for this, by optionally disabling this calculation.

Better would be to find out why it fails, but without the case that is very hard.

OPMUSER:
@atgeirr Your spot on - there are very distorted cells around the faults.

Instead of changing the calculation at this stage can we do the following:

- In the print file print an error message,  `Error Calculating PORV for some cells, see DBG file - will STOP`  

- Set the PORV to zero for the offending cells, so OPM Flow will continue, but at the end of parsing the deck and after the simulator has written EGRID and INIT files STOP. So one can visually check the model. The only reason for re-setting the PORV to zero is so that we can continue, and thus get all the offending cells, if you have a better idea (-999 etc.) that would be great.
-  In the the DBG list all the offending cells, in a nice tabular format, so with a bit of editing the user can manually set ACTNUM (or PORO etc.) to zero using the EQUALS keyword etc.
```
PORV Calculation Error for Cell  xxx  xxx xxx
PORV Calculation Error for Cell    1   10 100
PORV Calculation Error for Cell   10    1   1
etc.
```
Feel free to changes the test for the messages.

Some times this can just be one or two cells, and sometimes we need to rebuild the model, so to have a list of all the cells is very useful.

What do you think?

I cannot do this now due to my work load, but I could probably strip out a small portion of the grid and combined it with Norne data so that we can compare the calculation with you know who :-)

atgeirr:
> Instead of changing the calculation at this stage can we do the following:
> [...]

[...]

PR #3845, opened 2022-03-22
Don't Process Corner-Point Data If No Valid Cell Geometry

bska:
If no cell has a valid corner-point geometry, typically caused by using `GDFILE` to read non-finite data such as all `ZCORN = -1.0E+20`, then we must not attempt to generate a grid structure.  If we do, we will typically just fail somewhere deep down in the corner-point processing code and generate a diagnostic message that's hard to decipher.

With this commit we instead output a diagnostic message of the form
```
Error: No active cell in input grid has valid/finite cell geometry
Please check geometry keywords, especially if grid is imported through GDFILE
Program threw an exception: Allocating the simulation vanguard failed: No active cell in input grid has valid/finite cell geometry
Please check geometry keywords, especially if grid is imported through GDFILE
```
This may not be a lot better than the original diagnostic
```
Processing grid
flow: ${ROOT}/opm-grid/opm/grid/cpgpreprocess/preprocess.c:768: is_lefthanded: Assertion `! searching' failed.
Aborted (core dumped)
```
but does at least suggest that the grid data may be faulty.

### DIFF
--- a/opm/simulators/flow/Main.hpp
+++ b/opm/simulators/flow/Main.hpp
@@ -472,6 +472,8 @@ private:
                      summaryConfig_, nullptr, python, std::move(parseContext),
                      init_from_restart_file, outputCout_, outputInterval);
 
+            verifyValidCellGeometry(EclGenericVanguard::comm(), *this->eclipseState_);
+
             setupTime_ = externalSetupTimer.elapsed();
             outputFiles_ = (outputMode != FileOutputMode::OUTPUT_NONE);
         }
--- a/opm/simulators/utils/readDeck.cpp
+++ b/opm/simulators/utils/readDeck.cpp
@@ -54,6 +54,8 @@
 #include <opm/input/eclipse/Parser/ErrorGuard.hpp>
 #include <opm/input/eclipse/Parser/Parser.hpp>
 
+#include <opm/input/eclipse/Units/UnitSystem.hpp>
+
 #include <opm/simulators/flow/KeywordValidation.hpp>
 #include <opm/simulators/flow/ValidationFunctions.hpp>
 #include <opm/simulators/utils/ParallelEclipseState.hpp>
@@ -332,6 +334,44 @@ namespace {
         }
     }
 #endif
+
+    bool gridHasValidCellGeometry(const Opm::EclipseGrid& inputGrid,
+                                  const Opm::UnitSystem&  usys)
+    {
+        const auto numActive = inputGrid.getNumActive();
+
+        for (auto activeCell = 0*numActive; activeCell < numActive; ++activeCell) {
+            if (inputGrid.isValidCellGeomtry(inputGrid.getGlobalIndex(activeCell), usys)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    bool gridHasValidCellGeometry(Opm::Parallel::Communication comm,
+                                  const Opm::EclipseState&     eclipseState)
+    {
+        bool hasValidCells = false;
+
+        if (comm.rank() == 0) {
+            hasValidCells =
+                gridHasValidCellGeometry(eclipseState.getInputGrid(),
+                                         eclipseState.getDeckUnitSystem());
+        }
+
+#if HAVE_MPI
+        const auto status = comm.broadcast(&hasValidCells, 1, 0);
+
+        if (status != MPI_SUCCESS) {
+            throw std::invalid_argument {
+                "Unable to establish cell geomtry validity across MPI ranks"
+            };
+        }
+#endif  // HAVE_MPI
+
+        return hasValidCells;
+    }
 }
 
 // ---------------------------------------------------------------------------
@@ -517,4 +557,17 @@ void Opm::readDeck(Opm::Parallel::Communication    comm,
 
         std::exit(EXIT_FAILURE);
     }
+}
+
+void Opm::verifyValidCellGeometry(Parallel::Communication comm,
+                                  const EclipseState&     eclipseState)
+{
+    if (gridHasValidCellGeometry(comm, eclipseState)) {
+        return;
+    }
+
+    throw std::invalid_argument {
+        R"(No active cell in input grid has valid/finite cell geometry
+Please check geometry keywords, especially if grid is imported through GDFILE)"
+    };
 }
--- a/opm/simulators/utils/readDeck.hpp
+++ b/opm/simulators/utils/readDeck.hpp
@@ -86,6 +86,9 @@ void readDeck(Parallel::Communication         comm,
               bool                            initFromRestart,
               bool                            checkDeck,
               const std::optional<int>&       outputInterval);
+
+void verifyValidCellGeometry(Parallel::Communication comm,
+                             const EclipseState&     eclipseState);
 } // end namespace Opm
 
 #endif // OPM_READDECK_HEADER_INCLUDED


bska:
I am creating this PR in "draft" mode because it uses the new cell geometry validity predicate introduced in OPM/opm-common#2979 and must therefore not be merged until that PR is installed on the master branch.

I first encountered this issue in a model that did exactly what's outlined in the PR description:
```
GRID
-- ...
GDFILE
  'CASE_GEOMETRY.EGRID' /
```
and the `COORD` and `ZCORN` data stored in this `.EGRID` file were all invalid (values of `-1.0E+20`).

Note that the diagnostic messages in parallel are worse than what's shown in the description.  In a parallel run with `mpirun -np 6` I got the following diagnostics
```
Error: No active cell in input grid has valid/finite cell geometry
Please check geometry keywords, especially if grid is imported through GDFILE
Program threw an exception: Allocating the simulation vanguard failed: No active cell in input grid has valid/finite cell geometry
Please check geometry keywords, especially if grid is imported through GDFILE (on rank 0)
MPI_ABORT was invoked on rank 2 in communicator MPI_COMM_WORLD
with errorcode 1.

[...:1416714] 5 more processes have sent help message help-mpi-api.txt / mpi-abort
[...:1416714] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```
and any suggestion for how to make it better is greatly appreciated. I'll still argue that this is better than the alternative though
```
Processing grid
flow: ${ROOT}/opm-grid/opm/grid/cpgpreprocess/preprocess.c:768: is_lefthanded: Assertion `! searching' failed.
[...:1423782] *** Process received signal ***
[...:1423782] Signal: Aborted (6)
[...:1423782] Signal code:  (-6)
[...:1423782] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x143c0)[0x7f8a6dbfe3c0]
[...:1423782] [ 1] /lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb)[0x7f8a6d34403b]
[...:1423782] [ 2] /lib/x86_64-linux-gnu/libc.so.6(abort+0x12b)[0x7f8a6d323859]
[...:1423782] [ 3] /lib/x86_64-linux-gnu/libc.so.6(+0x22729)[0x7f8a6d323729]
[...:1423782] [ 4] /lib/x86_64-linux-gnu/libc.so.6(+0x34006)[0x7f8a6d335006]
[...:1423782] [ 5] ${BIN}/flow(+0x201c0e3)[0x560d4eda50e3]
[...:1423782] [ 6] ${BIN}/flow(+0x1ff7043)[0x560d4ed80043]
[...:1423782] [ 7] ${BIN}/flow(+0x1ffb8bb)[0x560d4ed848bb]
[...:1423782] [ 8] ${BIN}/flow(+0x1fec0e1)[0x560d4ed750e1]
[...:1423782] [ 9] ${BIN}/flow(+0x1fec1b2)[0x560d4ed751b2]
[...:1423782] [10] ${BIN}/flow(+0x138bbd9)[0x560d4e114bd9]
[...:1423782] [11] ${BIN}/flow(+0x43bb54)[0x560d4d1c4b54]
[...:1423782] [12] ${BIN}/flow(+0x43dcbe)[0x560d4d1c6cbe]
[...:1423782] [13] ${BIN}/flow(+0x43e1d9)[0x560d4d1c71d9]
[...:1423782] [14] ${BIN}/flow(+0x3227b3)[0x560d4d0ab7b3]
[...:1423782] [15] ${BIN}/flow(+0x300dfb)[0x560d4d089dfb]
[...:1423782] [16] ${BIN}/flow(+0x3018f8)[0x560d4d08a8f8]
[...:1423782] [17] ${BIN}/flow(+0x1ef509)[0x560d4cf78509]
[...:1423782] [18] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3)[0x7f8a6d3250b3]
[...:1423782] [19] ${BIN}/flow(+0x2f4e6e)[0x560d4d07de6e]
[...:1423782] *** End of error message ***
mpirun noticed that process rank 0 with PID 0 on node [...] exited on signal 6 (Aborted).
```

OPMUSER:
See #3770 for some ideas, as some times this can occur for just a few cells, normally around faults, so it would be good list say the first 100 in the debug file as per #3770.

bska:
> See https://github.com/OPM/opm-simulators/issues/3770 for some ideas, as some times this can occur for just a few cells, normally around faults, so it would be good list say the first 100 in the debug file as per https://github.com/OPM/opm-simulators/issues/3770.

That's a good point, but please be aware that the current PR addresses the case where this situation happens for **all** cells, not just some of them.

bska:
Upstream PR introducing validity predicate has been merged into master.  Marking this as "ready for review" now.

bska:
jenkins build this please

bska:
jenkins build this please

blattms:
As this is about EclState, I think this should be done for all grids. Hence we should do the check in another location.
I think only rank 0 will have a valid EclState, as only this rank reads the files.

Hence, maybe we should do this check in or after [readDeck]/OPM/opm-simulators/blob/2d6794f077ac81a6c8ef0c639ef715ddb5ac44d7/opm/simulators/flow/Main.hpp#L471)?

bska:
> As this is about `EclipseState`, I think this should be done for all grid types. Hence we should do the check in another location.

Good point.

> maybe we should do this check in or after [readDeck](/OPM/opm-simulators/blob/2d6794f077ac81a6c8ef0c639ef715ddb5ac44d7/opm/simulators/flow/Main.hpp#L471)?

That's a good idea.  I'll add a new helper function `verifyValidCellGeometry(const EclipseState& eclState)` or similar and call it immediately after `readDeck()`.

> I also think that only rank 0 will have a valid `EclipseState`, as only this rank reads the files.

That is true.  This PR tries to incorporate that fact by checking the cell geometry only on [rank 0](https://github.com/OPM/opm-simulators/pull/3845/files#diff-70d3ee73398c784858f50eb34f7c0eeb981fc46f1691afff4c0297b0aa2db832R233-R237) and then distributing the result of this check [to other ranks](https://github.com/OPM/opm-simulators/pull/3845/files#diff-70d3ee73398c784858f50eb34f7c0eeb981fc46f1691afff4c0297b0aa2db832R240-R252) when done.

bska:
> > As this is about `EclipseState`, I think this should be done for all grid types. Hence we should do the check in another location.
> 
> Good point.
> 
> > maybe we should do this check in or after [readDeck](/OPM/opm-simulators/blob/2d6794f077ac81a6c8ef0c639ef715ddb5ac44d7/opm/simulators/flow/Main.hpp#L471)?
> 
> That's a good idea. I'll add a new helper function `verifyValidCellGeometry(const EclipseState& eclState)` or similar and call it immediately after `readDeck()`.

I've pushed an update to implement the approach outlined above.  Thanks a lot for the suggestion, @blattms.

bska:
jenkins build this please

blattms:

blattms (approved):
Just running serial  jenkins for completeness,

bska:

bska:
Merging this now per prior approval.

blattms:
On a second look what we are doing seems strange. Is it really enough if there is only one cell with valid geometry. What do we do with all the invalid ones in that case? Ignore them?

Naively I would have assumed that already one cell with invalid cell geometry would create problems.

bska:
> On a second look what we are doing seems strange. Is it really enough if there is only one cell with valid geometry.

Every model I've ever seen either has "cell geometry valid" for all active cells or "cell geometry invalid" for all cells.  If we need to refine the current heuristic later we'll do so, but I'd rather not try to generalise from zero examples.

> What do we do with all the invalid ones in that case? Ignore them?

In the current case they'd all get geometric volume zero which means that they'd be deemed inactive unless explicitly assigned a positive pore-volume&mdash;e.g., through the `PORV` keyword.

PR #3847, opened 2022-03-23
improve initialization msw new wells

totto82:
Avoid singular matrix when initialized with zero rates.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -422,9 +422,15 @@ namespace Opm
 
         // initialized the well rates with the potentials i.e. the well rates based on bhp
         const int np = this->number_of_phases_;
-        const double sign = well_copy.well_ecl_.isInjector() ? 1.0 : -1.0;
+        bool trivial = true;
         for (int phase = 0; phase < np; ++phase){
-            ws.surface_rates[phase] = sign * ws.well_potentials[phase];
+            trivial = trivial && (ws.well_potentials[phase] == 0.0) ;
+        }
+        if (!trivial) {
+            const double sign = well_copy.well_ecl_.isInjector() ? 1.0 : -1.0;
+            for (int phase = 0; phase < np; ++phase) {
+                ws.surface_rates[phase] = sign * ws.well_potentials[phase];
+            }
         }
         well_copy.scaleSegmentRatesWithWellRates(well_state_copy);
 
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -1013,13 +1013,18 @@ namespace Opm
                 }
             }
         }
-        if (nonzero_rate_index == -1) {
-            // No nonzero rates.
-            return;
-        }
 
         // Calculate the rates that follow from the current primary variables.
         std::vector<double> well_q_s = computeCurrentWellRates(ebosSimulator, deferred_logger);
+
+        if (nonzero_rate_index == -1) {
+            // No nonzero rates.
+            // Use the computed rate directly
+            for (int p = 0; p < this->number_of_phases_; ++p) {
+               ws.surface_rates[p] = well_q_s[this->flowPhaseToEbosCompIdx(p)];
+            }
+            return;
+        }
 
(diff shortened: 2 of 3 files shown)


totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -422,9 +422,15 @@ namespace Opm
...
         for (int phase = 0; phase < np; ++phase){
-            ws.surface_rates[phase] = sign * ws.well_potentials[phase];
+            trivial = trivial && (ws.well_potentials[phase] == 0.0) ;
+        }
+        if (!trivial) {
+            const double sign = well_copy.well_ecl_.isInjector() ? 1.0 : -1.0;
+            for (int phase = 0; phase < np; ++phase) {
+                ws.surface_rates[phase] = sign * ws.well_potentials[phase];
+            }
         }
```
bska, line 434: So this will leave `ws.surface_rates` unchanged if the `trivial` condition is true.  Is there any guarantee that those `surface_rates` values will be reasonable/usable in that case?

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1013,14 +1013,22 @@ namespace Opm
...
+
         if (nonzero_rate_index == -1) {
             // No nonzero rates.
+            // Use the computed rate directly
+            const int comp_idx_nz = this->flowPhaseToEbosCompIdx(nonzero_rate_index);
+            for (int p = 0; p < this->number_of_phases_; ++p) {
+                const int comp_idx = this->flowPhaseToEbosCompIdx(p);
+                double& rate = ws.surface_rates[p];
+                rate = well_q_s[comp_idx];
+            }
```
bska, line 1028: The `comp_idx_nz` appears not to be used anywhere and the rest of the loop **could** be condensed to
```C++
for (int p = 0; p < this->number_of_phases; ++p) {
    ws.surface_rates[p] = well_q_s[this->flowPhaseToEbosCompIdx(p)];
}
```
if we wanted to.  I'm not suggesting that we write it like the above, I just want to make sure that this is the intended semantics.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: > Is there any guarantee that those surface_rates values will be reasonable/usable in that case?

The purpose of the rest of the commit is to assure usable initial condition for the case where the well start's up with zero rate and potentials.

bska: >> Is there any guarantee that those surface_rates values will be reasonable/usable in that case?
>
> The purpose of the rest of the commit is to assure usable initial condition for the case where the well start's up with zero rate and potentials.

Ah, right.  Sorry for not skimming though the complete PR before commenting.

totto82:
I am testing out the proposed automatic workflow for checking failed tests.

With course tolerances in the tests rel tol = 1e-2 and the -d option (only check restart steps) I get the following test failures locally 

```
	 73 - compareECLFiles_flow+UDQ_ACTIONX (Failed)
	 74 - compareECLFiles_flow+UDQ_WCONPROD (Failed)
	 75 - compareECLFiles_flow+ACTIONX_M1 (Failed)
	106 - compareECLFiles_flow+9_3C_GINJ_GAS_GCONSUMP_MSW (Failed)
	120 - compareECLFiles_flow+MOD4_UDQ_ACTIONX (Failed)
	128 - compareECLFiles_flow+NORNE_ATW2013_B1H_RE-PERF (Failed)
```


Plot for individual testing of these failures. 

image


image

image

image


image

totto82:
It would be very nice to automate this process! The difference can be explained by differences in time-stepping etc. A update of ref solutions should be save. I will fix the warning and @bska comment and start the update...

totto82:
jenkins build this please

alfbr:
> It would be very nice to automate this process! The difference can be explained by differences in time-stepping etc. A update of ref solutions should be save.

Thanks for sharing an overview of how it looks in practice. It looks like we have a good plan to make this process far more efficient.

atgeirr (approved):
Read through and think this is good.

totto82:

totto82:
jenkins build this opm-tests=741 please

PR #3848, opened 2022-03-23
Use empty string for buildtime if BUILD_TIMESTAMP is not set.

### LINKED ISSUE #3832: BUILD_TIMESTAMP (among others) prevents reproducible builds
The CMake scripts in opm-common are used to generate the header project-timestamp.h that contains a timestamp of the last cmake run. Unfortunately, this means that every build of the flow and the library differs even if there is no other change and makes build unreproducible. Does not seem nice. Maybe someone can comment on the purpose of this?

I fail to see why the time of the cmake run should matter at all. Maybe we should rather use the time the last source file was changed?
Might be challenging without git and just tarballs, though. Seems like we are relying on having git available at more places, which is probably also not the case for Debian package or anybody else using tarballs.

https://tests.reproducible-builds.org/debian/rb-pkg/bookworm/i386/opm-simulators.html

blattms:
In combination with the relevant changes in opm-common this prevent flow in binary Linux packages from having a timestamp in the executable that changes with every rebuild.

With the changes in opm-common  one can now set the variable OPM_BINARY_PACKAGE_VERSION to a meaningful version string (Debian 11.2: 2021.10-4). If that is done and flow is built from tarballs it will now not have a time stamp and print the package version to the PRT file. E.g.

```
Flow Version     =  2021.10 (Debian 11.2: 2021.10-1)
```

Note that without setting the variable or when building from a git repository there is no change.

Downstream of OPM/opm-common#2980
Closes #3832

### DIFF
--- a/opm/simulators/utils/moduleVersion.cpp
+++ b/opm/simulators/utils/moduleVersion.cpp
@@ -51,7 +51,11 @@ namespace Opm
     /// the binary was compiled.
     std::string compileTimestamp()
     {
+#ifdef BUILD_TIMESTAMP
         return BUILD_TIMESTAMP;
+#else
+        return "";
+#endif
     }
 
 } // namespace Opm


blattms:
jenkins build this opm-common=2980 please

PR #3849, opened 2022-03-24
Cleanup gaslift debugging output code

hakonhagland:
Some of the gas lift debugging output code outputs identical messages for all ranks. In some cases this is probably not desirable, in other cases it can be desirable if the output is prefixed by the rank number. Then it can be used as a debugging aid to verify that all ranks outputs identical messages.

This PR also cleans up the code by moving common parts of code to the common base class `GasLiftCommon`.

See also #3842.

### DIFF
--- a/opm/simulators/wells/GasLiftGroupInfo.cpp
+++ b/opm/simulators/wells/GasLiftGroupInfo.cpp
@@ -35,14 +35,13 @@ GasLiftGroupInfo(
     const Communication &comm,
     bool glift_debug
 ) :
-    GasLiftCommon(well_state, deferred_logger, glift_debug)
+    GasLiftCommon(well_state, deferred_logger, comm, glift_debug)
     , ecl_wells_{ecl_wells}
     , schedule_{schedule}
     , summary_state_{summary_state}
     , report_step_idx_{report_step_idx}
     , iteration_idx_{iteration_idx}
     , phase_usage_{phase_usage}
-    , comm_{comm}
     , glo_{schedule_.glo(report_step_idx_)}
 {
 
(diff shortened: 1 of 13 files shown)


hakonhagland:
jenkins build this please

totto82:
Thanks I will check and close by PR

opm/simulators/wells/GasLiftGroupInfo.cpp
```diff
@@ -294,7 +293,7 @@ checkDoGasLiftOptimization_(const std::string &well_name)
     auto itr = this->ecl_wells_.find(well_name);
     if (itr == this->ecl_wells_.end()) {
         // well_name is not present in the well_model's well container
-        displayDebugMessage_("Could find well ecl_wells. Skipping.", well_name);
+        displayDebugMessage_("Could not find well in ecl_wells. Skipping.", well_name);
         return false;
```
totto82, line 297: I suggest we remove this debug message altogether. The well container only contains local wells i.e. this will currently spam the output and I don't think that the value of this message is significant.

hakonhagland:
jenkins build this please

opm/simulators/wells/GasLiftGroupInfo.cpp
hakonhagland: Yes I agree. Have updated this PR accordingly.

hakonhagland: Yes I agree. Have updated this PR accordingly.

totto82:
Testing done. The output is reasonable now for my testing. Thanks.

PR #3851, opened 2022-03-28
Avoid premature closing of wells due to econ limits

totto82:
The motivation is to avoid premature closing of wells due to
1) inaccurate potentials
2) inaccurate ratios based on negative rates


totto82:
jenkins build this please

atgeirr:
Looks good to me. I had to think twice about the new logic: How is this different from just always using the rate? I eventually found the answer: If using potentials we can sometimes NOT close a well that would have been closed by using the rates. Merging.

PR #3853, opened 2022-03-30
Fix d issues 

totto82:
If d=1-rs*rv=0 we can not back-compute the surface volumes from the reservoir volumes. 

This PR consists of 2 commits. 
1) Avoid hard throw when d <=0 in MSW model and instead continue as if rs = rv = 0 for that connection. This fix is motivated by a field model and is sufficient for the simulator to finish the run. 
2) The second is still WIP in that it is not tested at all. When grabbing for "d = " I noticed that most places there was no guards for d = 0. The solution proposed for d<=0 is to continue as if rs = rv = 0.

### DIFF
--- a/opm/simulators/wells/MultisegmentWellEval.cpp
+++ b/opm/simulators/wells/MultisegmentWellEval.cpp
@@ -639,7 +639,8 @@ void
 MultisegmentWellEval<FluidSystem,Indices,Scalar>::
 computeSegmentFluidProperties(const EvalWell& temperature,
                               const EvalWell& saltConcentration,
-                              int pvt_region_index)
+                              int pvt_region_index,
+                              DeferredLogger& deferred_logger)
 {
     std::vector<double> surf_dens(baseif_.numComponents());
     // Surface density.
@@ -757,12 +758,22 @@ computeSegmentFluidProperties(const EvalWell& temperature,
             const unsigned oilCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::oilCompIdx);
 
             const EvalWell d = 1.0 - rs * rv;
-
-            if (rs > 0.0 && d > 0.0) {
-                mix[gasCompIdx] = (mix_s[gasCompIdx] - mix_s[oilCompIdx] * rs) / d;
-            }
-            if (rv > 0.0 && d > 0.0) {
-                mix[oilCompIdx] = (mix_s[oilCompIdx] - mix_s[gasCompIdx] * rv) / d;
+            if (d <= 0.0) {
+                std::ostringstream sstr;
+                sstr << "Problematic d value " << d << " obtained for well " << baseif_.name()
+                     << " during segment density calculations with rs " << rs
+                     << ", rv " << rv << " and pressure " << seg_pressure
+                     << " obtaining d " << d
+                     << " Continue as if no dissolution (rs = 0) and vaporization (rv = 0) "
+                     << " for this connection.";
+                deferred_logger.debug(sstr.str());
+            } else {
+                if (rs > 0.0) {
+                    mix[gasCompIdx] = (mix_s[gasCompIdx] - mix_s[oilCompIdx] * rs) / d;
+                }
+                if (rv > 0.0) {
+                    mix[oilCompIdx] = (mix_s[oilCompIdx] - mix_s[gasCompIdx] * rv) / d;
+                }
             }
         }
 
@@ -1195,16 +1206,16 @@ getSegmentSurfaceVolume(const EvalWell& temperature,
                  << " during conversion to surface volume with rs " << rs
                  << ", rv " << rv << " and pressure " << seg_pressure
                  << " obtaining d " << d
-                 << " Continue without dissolution (rs = 0) and vaporization (rv = 0) "
+                 << " Continue as if no dissolution (rs = 0) and vaporization (rv = 0) "
                  << " for this connection.";
             OpmLog::debug(sstr.str());
-        }
-
-        if (rs > 0.0) { // rs > 0.0?
-            mix[gasCompIdx] = (mix_s[gasCompIdx] - mix_s[oilCompIdx] * rs) / d;
-        }
-        if (rv > 0.0) { // rv > 0.0?
-            mix[oilCompIdx] = (mix_s[oilCompIdx] - mix_s[gasCompIdx] * rv) / d;
+        } else {
+            if (rs > 0.0) {
+                mix[gasCompIdx] = (mix_s[gasCompIdx] - mix_s[oilCompIdx] * rs) / d;
+            }
+            if (rv > 0.0) {
+                mix[oilCompIdx] = (mix_s[oilCompIdx] - mix_s[gasCompIdx] * rv) / d;
+            }
         }
     }
 
--- a/opm/simulators/wells/MultisegmentWellEval.cpp
+++ b/opm/simulators/wells/MultisegmentWellEval.cpp
@@ -1194,9 +1194,10 @@ getSegmentSurfaceVolume(const EvalWell& temperature,
             sstr << "Problematic d value " << d << " obtained for well " << baseif_.name()
                  << " during conversion to surface volume with rs " << rs
                  << ", rv " << rv << " and pressure " << seg_pressure
-                 << " obtaining d " << d;
+                 << " obtaining d " << d
+                 << " Continue without dissolution (rs = 0) and vaporization (rv = 0) "
+                 << " for this connection.";
             OpmLog::debug(sstr.str());
-            OPM_THROW_NOLOG(NumericalIssue, sstr.str());
         }
 
         if (rs > 0.0) { // rs > 0.0?
--- a/opm/simulators/wells/MultisegmentWellEval.cpp
+++ b/opm/simulators/wells/MultisegmentWellEval.cpp
@@ -758,10 +758,10 @@ computeSegmentFluidProperties(const EvalWell& temperature,
 
             const EvalWell d = 1.0 - rs * rv;
 
-            if (rs != 0.0) { // rs > 0.0?
+            if (rs > 0.0 && d > 0.0) {
                 mix[gasCompIdx] = (mix_s[gasCompIdx] - mix_s[oilCompIdx] * rs) / d;
             }
-            if (rv != 0.0) { // rv > 0.0?
+            if (rv > 0.0 && d > 0.0) {
                 mix[oilCompIdx] = (mix_s[oilCompIdx] - mix_s[gasCompIdx] * rv) / d;
             }
         }
--- a/opm/simulators/wells/StandardWellEval.cpp
+++ b/opm/simulators/wells/StandardWellEval.cpp
@@ -950,13 +950,14 @@ computeConnectionDensities(const std::vector<double>& perfComponentRates,
             if (!rvmax_perf.empty() && mix[gaspos] > 1e-12) {
                 rv = std::min(mix[oilpos]/mix[gaspos], rvmax_perf[perf]);
             }
-            if (rs != 0.0) {
+            double d = 1.0 - rs*rv;
+            if (rs > 0.0 && d > 0.0) {
                 // Subtract gas in oil from gas mixture
-                x[gaspos] = (mix[gaspos] - mix[oilpos]*rs)/(1.0 - rs*rv);
+                x[gaspos] = (mix[gaspos] - mix[oilpos]*rs)/d;
             }
-            if (rv != 0.0) {
+            if (rv > 0.0 && d > 0.0) {
                 // Subtract oil in gas from oil mixture
-                x[oilpos] = (mix[oilpos] - mix[gaspos]*rv)/(1.0 - rs*rv);;
+                x[oilpos] = (mix[oilpos] - mix[gaspos]*rv)/d;
             }
         }
         double volrat = 0.0;
--- a/opm/simulators/wells/StandardWellEval.cpp
+++ b/opm/simulators/wells/StandardWellEval.cpp
@@ -831,7 +831,8 @@ computeConnectionDensities(const std::vector<double>& perfComponentRates,
                            const std::vector<double>& b_perf,
                            const std::vector<double>& rsmax_perf,
                            const std::vector<double>& rvmax_perf,
-                           const std::vector<double>& surf_dens_perf)
+                           const std::vector<double>& surf_dens_perf,
+                           DeferredLogger& deferred_logger)
 {
     // Verify that we have consistent input.
     const int nperf = baseif_.numPerfs();
@@ -951,6 +952,18 @@ computeConnectionDensities(const std::vector<double>& perfComponentRates,
                 rv = std::min(mix[oilpos]/mix[gaspos], rvmax_perf[perf]);
             }
             double d = 1.0 - rs*rv;
+
+            if (d <= 0.0) {
+                std::ostringstream sstr;
+                sstr << "Problematic d value " << d << " obtained for well " << baseif_.name()
+                     << " during ccomputeConnectionDensities with rs " << rs
+                     << ", rv " << rv
+                     << " obtaining d " << d
+                     << " Continue as if no dissolution (rs = 0) and vaporization (rv = 0) "
+                     << " for this connection.";
+                deferred_logger.debug(sstr.str());
+            }
+
             if (rs > 0.0 && d > 0.0) {
                 // Subtract gas in oil from gas mixture
                 x[gaspos] = (mix[gaspos] - mix[oilpos]*rs)/d;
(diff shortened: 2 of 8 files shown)


totto82:
jenkins build this please

atgeirr:
> This fix is motivated by a field model

Does that mean the pressure is significantly out of range of the pvt tables when this happens? If so, this is related to the linear extrapolation of the tables that causes unphysical behaviour. Your fix may address the symptom, but we should also address the cause.

If not, then we should look for other errors!

totto82:
> Does that mean the pressure is significantly out of range of the pvt tables when this happens? If so, this is related to the linear extrapolation of the tables that causes unphysical behaviour. Your fix may address the symptom, but we should also address the cause.

The pressure was not significantly out of range for this case. Throwing here is anyway too strong I think. We don't solve our problems exactly and we should allow for some flexibility in the solution while iterating. Linear extrapolation is linear extrapolation I don't see how we could improve on that.

atgeirr:
> The pressure was not significantly out of range for this case. Throwing here is anyway too strong I think. We don't solve our problems exactly and we should allow for some flexibility in the solution while iterating. Linear extrapolation is linear extrapolation I don't see how we could improve on that.

I agree with most of this, except that we should not do linear extrapolation of the pvt tables but constant extrapolation. That is not for this PR to fix though, and I agree with the premise.

atgeirr (commented):
This is good, only a few nitpicks:
 - Use 0.0 not zero for double literal.
 - Braces for if-statements.

Can be merged after fixing. We still need to adress the cause of the d < 0.0 situation in another PR.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -573,16 +575,17 @@ namespace Opm
                 EvalWell cq_r_thermal(this->numWellEq_ + Indices::numEq, 0.);
                 if (FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx) && FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx)) {
 
-                    if(FluidSystem::waterPhaseIdx == phaseIdx)
+                    // d = 1.0 - rs * rv
```
GitPaean, line 578: I am not talking about the change in this PR, to me, the for loop `for (unsigned phaseIdx = 0; phaseIdx < FluidSystem::numPhases; ++phaseIdx) ` is a little problematic. Will get back to this very soon.

GitPaean: The things looking strange here is that, `if(FluidSystem::waterPhaseIdx == phaseIdx)`, we do not need to handle the things related to `rs`, `rv` and `d` at all. 

As a result, should not the `if` structure should be something like the following, 
```c++
            for (unsigned phaseIdx = 0; phaseIdx < FluidSystem::numPhases; ++phaseIdx) {
                if (!FluidSystem::phaseIsActive(phaseIdx)) {
                    continue;
                }

                EvalWell cq_r_thermal(this->numWellEq_ + Indices::numEq, 0.);
                const unsigned activeCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::solventComponentIndex(phaseIdx));
                const bool both_oil_gas = FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx) && FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx);
                if(FluidSystem::waterPhaseIdx == phaseIdx || ~both_oil_gas) {
                    cq_r_thermal = cq_s[activeCompIdx] / this->extendEval(fs.invB(phaseIdx));
                } else {
                    // remove dissolved gas and vapporized oil
                    const unsigned oilCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::oilCompIdx);
                    const unsigned gasCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx);
                    // q_os = q_or * b_o + rv * q_gr * b_g
                    // q_gs = q_gr * g_g + rs * q_or * b_o
                    // d = 1.0 - rs * rv
                    const EvalWell d = this->extendEval(1.0 - fs.Rv() * fs.Rs());
                    // q_gr = 1 / (b_g * d) * (q_gs - rs * q_os)
                    if(FluidSystem::gasPhaseIdx == phaseIdx)
                        cq_r_thermal = (cq_s[gasCompIdx] - this->extendEval(fs.Rs()) * cq_s[oilCompIdx]) / (d * this->extendEval(fs.invB(phaseIdx)) );
                    // q_or = 1 / (b_o * d) * (q_os - rv * q_gs)
                    if(FluidSystem::oilPhaseIdx == phaseIdx)
                        cq_r_thermal = (cq_s[oilCompIdx] - this->extendEval(fs.Rv()) * cq_s[gasCompIdx]) / (d * this->extendEval(fs.invB(phaseIdx)) );

                }
```

Or I miss something here? 

Sorry, I am talking about the master branch, not the change in the PR at the moment. However, the problem I think in the master branch is shown by the content in this PR `                    if(d <= 0 || FluidSystem::waterPhaseIdx == phaseIdx)` here.

totto82: I don't think the code your propose is any different than what is in the PR with the exception of how many times you compute `const EvalWell d = this->extendEval(1.0 - fs.Rv() * fs.Rs());` or did I misunderstand your comment. This calculations could be moved outside the loop over the phases if performance is an issue here.

totto82:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
GitPaean: In the original code, 
```c++
                if (FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx) && FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx)) {

                    if(FluidSystem::waterPhaseIdx == phaseIdx)
                        cq_r_thermal = cq_s[activeCompIdx] / this->extendEval(fs.invB(phaseIdx));

                    // remove dissolved gas and vapporized oil
                    const unsigned oilCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::oilCompIdx);
                    const unsigned gasCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx);
                    // q_os = q_or * b_o + rv * q_gr * b_g
                    // q_gs = q_gr * g_g + rs * q_or * b_o
                    // d = 1.0 - rs * rv
                    const EvalWell d = this->extendEval(1.0 - fs.Rv() * fs.Rs());
                    // q_gr = 1 / (b_g * d) * (q_gs - rs * q_os)
                    if(FluidSystem::gasPhaseIdx == phaseIdx)
                        cq_r_thermal = (cq_s[gasCompIdx] - this->extendEval(fs.Rs()) * cq_s[oilCompIdx]) / (d * this->extendEval(fs.invB(phaseIdx)) );
                    // q_or = 1 / (b_o * d) * (q_os - rv * q_gs)
                    if(FluidSystem::oilPhaseIdx == phaseIdx)
                        cq_r_thermal = (cq_s[oilCompIdx] - this->extendEval(fs.Rv()) * cq_s[gasCompIdx]) / (d * this->extendEval(fs.invB(phaseIdx)) );

                } else {
                    cq_r_thermal = cq_s[activeCompIdx] / this->extendEval(fs.invB(phaseIdx));
                }
```

Assuming black oil case, there are oil, gas and water. 

`if(FluidSystem::waterPhaseIdx == phaseIdx)`, 

after
```c++
                    if(FluidSystem::waterPhaseIdx == phaseIdx)
                        cq_r_thermal = cq_s[activeCompIdx] / this->extendEval(fs.invB(phaseIdx));
```

We still need do calculate `rs`, `rv`, `d`, and they will NOT be applied to `cq_r_thermal`, because there other two if guards there, 
```
                    if(FluidSystem::gasPhaseIdx == phaseIdx)
                        cq_r_thermal = (cq_s[gasCompIdx] - this->extendEval(fs.Rs()) * cq_s[oilCompIdx]) / (d * this->extendEval(fs.invB(phaseIdx)) );
                    // q_or = 1 / (b_o * d) * (q_os - rv * q_gs)
                    if(FluidSystem::oilPhaseIdx == phaseIdx)
                        cq_r_thermal = (cq_s[oilCompIdx] - this->extendEval(fs.Rv()) * cq_s[gasCompIdx]) / (d * this->extendEval(fs.invB(phaseIdx)) );
```

GitPaean: With 
```c++
                } else {
                    cq_r_thermal = cq_s[activeCompIdx] / this->extendEval(fs.invB(phaseIdx));
                }
```
The result is probably okay, while it is confusing `if` structure and there are non-necessary calculation of `rs`, rv`, and `d`.

totto82: I agree with you that this part of the code could be cleaned up a bit. Your suggestion looks good, but as you said this clean-up is unrelated to the content of this PR, and should be done in a following up PR IMO.

GitPaean: Yes. But it removes the necessity of the following change in this PR, because they do not need to be together.  I do not insist, a cleaning up PR can be okay. 

```
if(d <= 0.0 || FluidSystem::waterPhaseIdx == phaseIdx) {
```

opm/simulators/wells/StandardWellEval.cpp
```diff
@@ -950,13 +950,14 @@ computeConnectionDensities(const std::vector<double>& perfComponentRates,
...
                 // Subtract gas in oil from gas mixture
-                x[gaspos] = (mix[gaspos] - mix[oilpos]*rs)/(1.0 - rs*rv);
+                x[gaspos] = (mix[gaspos] - mix[oilpos]*rs)/d;
             }
-            if (rv != 0.0) {
+            if (rv > 0.0 && d > 0.0) {
                 // Subtract oil in gas from oil mixture
-                x[oilpos] = (mix[oilpos] - mix[gaspos]*rv)/(1.0 - rs*rv);;
+                x[oilpos] = (mix[oilpos] - mix[gaspos]*rv)/d;
             }
```
GitPaean, line 961: We should have debugging output here, for the abnormal d values (d <= 0.0 || d > 1.0). 

And I do not think we should use the `d` values that bigger than 1 either, which implies negative `rs` or `rv`.

opm/simulators/wells/MultisegmentWellEval.cpp
```diff
@@ -1194,9 +1194,10 @@ getSegmentSurfaceVolume(const EvalWell& temperature,
             sstr << "Problematic d value " << d << " obtained for well " << baseif_.name()
                  << " during conversion to surface volume with rs " << rs
                  << ", rv " << rv << " and pressure " << seg_pressure
-                 << " obtaining d " << d;
+                 << " obtaining d " << d
+                 << " Continue without dissolution (rs = 0) and vaporization (rv = 0) "
+                 << " for this connection.";
```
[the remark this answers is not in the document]
totto82: Thanks for noticing. I must have removed the "else" here without intention. I will need to double check to assure that this does not have any unintended effects. So we should hold of merging until I have tested a bit more.

opm/simulators/wells/StandardWellEval.cpp
totto82: I think the correct solution for d > 1 is to set it to 1. If rs > 0.0 and rv < 0.0 We should remove the effect of rs as if there is no rv. An alternative is to force rs/rv >= 0

opm/simulators/wells/StandardWell_impl.hpp
GitPaean: I created a cleanup PR based on the above discussion, 

https://github.com/OPM/opm-simulators/pull/3855

Maybe it is good to get that one first, so we can review the change in this part of the code?

totto82:
jenkins build this please

totto82:
@GitPaean @atgeirr Can this be merged now?

GitPaean:
>  @GitPaean @atgeirr Can this be merged now?

Thanks for the update, let me take a look at it now.  Are the testing results all good?

totto82:
> Are the testing results all good?

Yes, this is tested by @tskille and myself.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -580,14 +600,29 @@ namespace Opm
...
+                    if(FluidSystem::gasPhaseIdx == phaseIdx && d > 0.0) {
                         cq_r_thermal = (cq_s[gasCompIdx] - this->extendEval(fs.Rs()) * cq_s[oilCompIdx]) / (d * this->extendEval(fs.invB(phaseIdx)) );
                     // q_or = 1 / (b_o * d) * (q_os - rv * q_gs)
-                    if(FluidSystem::oilPhaseIdx == phaseIdx)
+                    } else if(FluidSystem::oilPhaseIdx == phaseIdx && d > 0.0) {
                         cq_r_thermal = (cq_s[oilCompIdx] - this->extendEval(fs.Rv()) * cq_s[gasCompIdx]) / (d * this->extendEval(fs.invB(phaseIdx)) );
+                    } else {
+                        assert(d <= 0.0);
+                        // for d <= 0.0 we continue as if rs = 0 and rv = 0
+                        cq_r_thermal = cq_s[activeCompIdx] / this->extendEval(fs.invB(phaseIdx));
```
GitPaean, line 624: Nitpicking here. 

This line can moved to below the line 614 `                        deferred_logger.debug(sstr.str());`, and for the gasPhase and oilPhase if conditions do not need to `d > 0.0`.  And also we do not need the `assert`.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -354,12 +359,27 @@ namespace Opm
...
+                         << ", rv " << rv << " and pressure " << pressure
+                         << " obtaining d " << d
+                         << " Continue as if no dissolution (rs = 0) and vaporization (rv = 0) "
+                         << " for this connection.";
+                    deferred_logger.debug(sstr.str());
+                    }
                     // vaporized oil into gas
                     // rv * q_gr * b_g = rv * (q_gs - rs * q_os) / d
-                    perf_vap_oil_rate = getValue(rv) * (getValue(cq_s[gasCompIdx]) - getValue(rs) * getValue(cq_s[oilCompIdx])) / d;
+                    if (d > 0.0) {
```
GitPaean, line 375: Same here, the two `if (d > 0.0)` can be in the `else` branch of `if (d <= 0.0) {`.

opm/simulators/wells/StandardWellEval.cpp
```diff
@@ -950,13 +951,26 @@ computeConnectionDensities(const std::vector<double>& perfComponentRates,
             if (!rvmax_perf.empty() && mix[gaspos] > 1e-12) {
                 rv = std::min(mix[oilpos]/mix[gaspos], rvmax_perf[perf]);
             }
-            if (rs != 0.0) {
+            double d = 1.0 - rs*rv;
```
GitPaean, line 954: `const double d` here please same with other `d` calculations.

opm/simulators/wells/StandardWellEval.cpp
```diff
@@ -950,13 +951,26 @@ computeConnectionDensities(const std::vector<double>& perfComponentRates,
...
+            if (d <= 0.0) {
+                std::ostringstream sstr;
+                sstr << "Problematic d value " << d << " obtained for well " << baseif_.name()
+                     << " during ccomputeConnectionDensities with rs " << rs
+                     << ", rv " << rv
+                     << " obtaining d " << d
+                     << " Continue as if no dissolution (rs = 0) and vaporization (rv = 0) "
+                     << " for this connection.";
+                deferred_logger.debug(sstr.str());
+            }
```
GitPaean, line 965: Same with other places, and also the version in `MultisegmentWellEval.cpp`, `d > 0.0` can be removed by putting them in the `else` branch.

GitPaean:
Small nitpicking comments, basically about coding style and using consistent code across different files. 

Other than that, the PR looks good to me.

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
Thanks @GitPaean for review. I have addressed your final remarks and Jenkins is green. Lets merge.

OPMUSER:
Okay I know what rs and rv but what is d (d for David) :-)

atgeirr:
> Okay I know what rs and rv are, but what is d (d for David) :-)

d = 1-rs*rv

OPMUSER:
Yep, I know but I don't understand the context for wring up the bug fix.

OPMUSER:
@atgeirr Still don't understand the context for writing up the bug fix.

totto82:
When you convert from surface to reservoir rates you need to solve a 2x2 system. For that you use Cramer's rule. 
So d is the determinant.  If d is zero the system is singular and the system unsolvable. This should not happen with physical sane rs and rv values, but unconverged solutions it may happen due to extrapolation of rs/rv tables for uphysical pressures. The fix removes a hard throw and instead ignore the dissolved/vaporized part and try to continue. Some times this is sufficient to get a more decent solution in the next newton iteration, other times not and the simulator need to chop the time etc. 
```
 // q_os = q_or * b_o + rv * q_gr * b_g
// q_gs = q_gr * b_g + rs * q_or * b_o
 // d = 1.0 - rs * rv
 // q_or = 1 / (b_o * d) * (q_os - rv * q_gs)
// q_gr = 1 / (b_g * d) * (q_gs - rs * q_os)
```

OPMUSER:
@totto82 Now I understand -Tusen takk

PR #3861, opened 2022-04-05
Retrieve Block Level Solvent Saturation When Requested

bska:
This PR adds support for the block level solvent saturation summary output keyword `BNSAT`.


bska:
When combined with OPM/opm-common#2986 this PR generates time series summary output for block level solvent saturation.

bska:
jenkins build this please

blattms:
any reason not to test with the upstream change?

PR #3863, opened 2022-04-05
fix to --shut-unsolvable-wells=true option

totto82:
Should not be affect default behavior. But fixes some fallout's in the shut-unsolvable-wells option.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1225,6 +1225,11 @@ namespace Opm {
         for (const auto& well : well_container_) {
             if (well->isOperableAndSolvable() ) {
                 local_report += well->getWellConvergence(this->wellState(), B_avg, local_deferredLogger, iterationIdx > param_.strict_outer_iter_wells_ );
+            } else {
+                ConvergenceReport report;
+                using CR = ConvergenceReport;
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

totto82:
Since this is a bug fix for an existing option (even though not default), I added a release flag. I plan to create a PR that makes the --shut-unsolvable-wells=true but that certainly would need more testing and should not be merged before the release.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1225,6 +1225,11 @@ namespace Opm {
         for (const auto& well : well_container_) {
             if (well->isOperableAndSolvable() ) {
                 local_report += well->getWellConvergence(this->wellState(), B_avg, local_deferredLogger, iterationIdx > param_.strict_outer_iter_wells_ );
+            } else {
+                ConvergenceReport report;
+                using CR = ConvergenceReport;
+                report.setWellFailed({CR::WellFailure::Type::Unsolvable, CR::Severity::Normal, -1, well->name()});
+                local_report += report;
```
blattms, line 1232: Why do we need to add a default report?

blattms: Nervermind, I think I got it now. It's for the status.

PR #3866, opened 2022-04-08
support item 4 wvfpexp

totto82:
Since default is NO for WVFPEXP item 4 this fixes unexpected behavior in Flow that was introduced in https://github.com/OPM/opm-simulators/pull/3824


totto82:
jenkins build this please

atgeirr (approved):

totto82:
The test failures are the same cases that needed an update with #3824. I will ask for new reference data that will take the state of this cases back to before #3824. None of these cases have WVFPEXP.

totto82:

totto82:
jenkins build this opm-tests=745 please

PR #3867, opened 2022-04-08
avoid throw for nan residuals instead return well as unconverged

totto82:
This should not change the behavior but avoid unexpected "errors"  counts in the error summary. 
Issue reported by @tskille 
 
image

### DIFF
--- a/opm/simulators/wells/MultisegmentWellEval.cpp
+++ b/opm/simulators/wells/MultisegmentWellEval.cpp
@@ -1619,10 +1619,10 @@ assemblePressureEq(const int seg,
 }
 
 template<typename FluidSystem, typename Indices, typename Scalar>
-std::vector<Scalar>
+std::pair<bool, std::vector<Scalar> >
 MultisegmentWellEval<FluidSystem,Indices,Scalar>::
-getWellResiduals(const std::vector<Scalar>& B_avg,
-                 DeferredLogger& deferred_logger) const
+getFiniteWellResiduals(const std::vector<Scalar>& B_avg,
+                       DeferredLogger& deferred_logger) const
 {
     assert(int(B_avg.size() ) == baseif_.numComponents());
     std::vector<Scalar> residuals(numWellEq + 1, 0.0);
@@ -1638,8 +1638,9 @@ getWellResiduals(const std::vector<Scalar>& B_avg,
                 }
             }
             if (std::isnan(residual) || std::isinf(residual)) {
-                OPM_DEFLOG_THROW(NumericalIssue, "nan or inf value for residal get for well " << baseif_.name()
-                                                << " segment " << seg << " eq_idx " << eq_idx, deferred_logger);
+                deferred_logger.debug("nan or inf value for residal get for well " + baseif_.name()
+                                      + " segment " + std::to_string(seg) + " eq_idx " + std::to_string(eq_idx));
+                return {false, residuals};
             }
 
             if (residual > residuals[eq_idx]) {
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1369,7 +1369,12 @@ namespace Opm
 
         const int max_iter_number = this->param_.max_inner_iter_ms_wells_;
         const WellState well_state0 = well_state;
-        const std::vector<Scalar> residuals0 = this->getWellResiduals(Base::B_avg_, deferred_logger);
+        std::vector<Scalar> residuals(numWellEq + 1, 0.0);
+
+        // getWellResiduals returns true for nan/inf residuals
+        if(this->getWellResiduals(Base::B_avg_, residuals, deferred_logger))
+            return false;
```
bska, line 1376: I know why it is this way, especially with the comment, but I'll still remark that
```C++
if (this->getX(...)) {
    return false
}
```
looks unexpected to me.  I would normally expect a function called `getSomething` to return `true` if successful and `false` otherwise.  Maybe we can rename the function to `getFiniteWellResiduals()`, make it return `true` if all residuals are finite and then write this conditional as
```C++
if (! this->getFiniteWellResiduals(...)) {
    return false;
}
```
instead?

Another option might be to extract the finite vs. non-finite residual checking out to a separate helper function and have `getWellResiduals()` just return the values as it already does.  That way we won't have to introduce an "out" parameter or add a deferred logger.  The downside to that is that we lose a little bit of context and it becomes harder to distinguish the control equation from the flow equations.

totto82: I liked your first option. Will update accordingly.

totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWellEval.cpp
```diff
@@ -1619,13 +1619,15 @@ assemblePressureEq(const int seg,
 }
 
 template<typename FluidSystem, typename Indices, typename Scalar>
-std::vector<Scalar>
+bool
```
blattms, line 1622: Just a thought:

std::pair<bool, std::vector<Scalar> > getFiniteWellResiduals(...){
   ....
   return {isFinite, residuals};
}
....
const auto& [isFinite, resilduals] = getFiniteWellResiduals(...);
```

Might be more explicit and easier to use.

**Comment by blattms** [comment]
benchmark please

**Comment by ytelses** [comment]
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 0.985    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 1.005    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 0.996    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 0.998    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.996    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 0.998    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1588

File: opm/simulators/wells/MultisegmentWellEval.cpp (hunk 2 above)
Comment (totto82) [reply opm/simulators/wells/MultisegmentWellEval.cpp:1622]: Agree, I will update accordingly.

**Comment by totto82** [comment]
jenkins build this please

File: opm/simulators/wells/MultisegmentWell_impl.hpp [hunk 3]
```diff
@@ -1394,7 +1399,12 @@ namespace Opm
                 break;
             }
 
-            residual_history.push_back(this->getWellResiduals(Base::B_avg_, deferred_logger));
+            // getFinteWellResiduals returns false for nan/inf residuals
+            const auto& [isFinite, residuals] = this->getFiniteWellResiduals(Base::B_avg_, deferred_logger);
```
blattms, line 1403: Unfortunately, this produces warnings:
```
MultisegmentWell_impl.hpp:1403:26: warning: declaration of ‘isFinite’ shadows a previous local [-Wshadow]
 1403 |             const auto& [isFinite, residuals] = this->getFiniteWellResiduals(Base::B_avg_, deferred_logger);
      |                          ^~~~~~~~
```
Maybe we should rename the variables here to avoid this? `const auto& [isFinite2, residuals2]` or use curly brackets to give each of these it's own scope?

totto82:
jenkins build this please

blattms:
Thanks for the improvements. Will merge once green and create release branches.

PR #3871, opened 2022-04-12
fix density in initialization for case without DISGAS

totto82:
Discovered when investigating unexpected test failures in https://github.com/OPM/opm-material/pull/498

### DIFF
--- a/ebos/equil/initstateequil.hh
+++ b/ebos/equil/initstateequil.hh
@@ -219,9 +219,12 @@ private:
     density(const double depth,
             const double press) const
     {
-        double rs = rs_(depth, press, temp_);
+        double rs = 0.0;
+        if(FluidSystem::enableDissolvedGas())
+            rs = rs_(depth, press, temp_);
+
         double bOil = 0.0;
-        if (!FluidSystem::enableDissolvedGas() || rs >= FluidSystem::oilPvt().saturatedGasDissolutionFactor(pvtRegionIdx_, temp_, press)) {
+        if (rs >= FluidSystem::oilPvt().saturatedGasDissolutionFactor(pvtRegionIdx_, temp_, press)) {
             bOil = FluidSystem::oilPvt().saturatedInverseFormationVolumeFactor(pvtRegionIdx_, temp_, press);
         }
         else {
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

totto82:
Seems we only have three test without DISGAS. The test failures are expected and we will need a ref update.

blattms (commented):
If there is no problem for both side being zero, then this looks good

ebos/equil/initstateequil.hh
```diff
@@ -219,9 +219,12 @@ private:
...
             const double press) const
     {
-        double rs = rs_(depth, press, temp_);
+        double rs = 0.0;
+        if(FluidSystem::enableDissolvedGas())
+            rs = rs_(depth, press, temp_);
+
         double bOil = 0.0;
-        if (!FluidSystem::enableDissolvedGas() || rs >= FluidSystem::oilPvt().saturatedGasDissolutionFactor(pvtRegionIdx_, temp_, press)) {
+        if (rs >= FluidSystem::oilPvt().saturatedGasDissolutionFactor(pvtRegionIdx_, temp_, press)) {
```
blattms, line 227: Might there be a problem or `if(0>=0)`, i.e if `FluidSystem::oilPvt().saturatedGasDissolutionFactor(pvtRegionIdx_, temp_, press)==0` holds? Same below.

totto82: No. Then both function Will return the same value.

blattms:
jenkins build this with opm-tests=747 please

ebos/equil/initstateequil.hh
atgeirr: Late to the party, just checking to see I understood the fix. The behaviour with dissolved gas enabled is the same as before. The behaviour with disabled dissolved gas is changed as follows:
 - Avoid calling rs_() for slightly better performance (the result never mattered).
 - Always use inverseFormationVolumeFactor() instead of saturatedInverseFormationVolumeFactor().
Is that correctly assessed? If so, it is not obvious to me that the change to call inverseFormationVolumeFactor() with a zero rs is the right thing to do, should not the saturated IFV be the same as the undersaturated IFV when there is no dissolved gas?

totto82: saturatedInverseFormationVolumeFactor calls inverseFormationVolumeFactor with rs = rsSat. We could add a fix for this in opm-material that rs = rsSat = 0 (if DISGAS is false)?

atgeirr: > saturatedInverseFormationVolumeFactor calls inverseFormationVolumeFactor with rs = rsSat.

I do not see this in the code? Looking in LiveOilPvt.hpp, I see:
```c++
    /*!                                                                                                                                                                                        
     * \brief Returns the formation volume factor [-] of the fluid phase.                                                                                                                      
     */
    template <class Evaluation>
    Evaluation saturatedInverseFormationVolumeFactor(unsigned regionIdx,
                                                     const Evaluation& /*temperature*/,
                                                     const Evaluation& pressure) const
    {
        // ATTENTION: Rs is represented by the _first_ axis!                                                                                                                                   
        return inverseSaturatedOilBTable_[regionIdx].eval(pressure, /*extrapolate=*/true);
    }
```

totto82: The problem is in brineCo2Pvt.hpp, which is not aware of the DISGAS keyword. That's taken case by the FluidSystem. For standard table version the pvt multiplexer will chose separate classes for DISGAS and not (i.e. for DISGAS you will use PVTO not PVDO etc.) for CO2STORE the call path is the same. I could add a disgas_ member to brineco2pvt.hpp and if(!disgas_) return 0.0 for the saturated functions.

PR #3859, opened 2022-04-04
use a relaxed (but not too relaxed) well tolerance when iter>6

totto82:
Motivation is to avoid premature closing of wells.

PR #3872, opened 2022-04-19
Fix STOP in combination with inoperable/unsolvable

totto82:
Some fixups for combination STOP + shut_unsolvable_wells=true (default is false) 

This also change default behavior for wells with STOP that do not converge when time step is less than threshold. 

Tested on the norne prediction tests in opm-tests/norne

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -2067,7 +2067,7 @@ forceShutWellByName(const std::string& wellname,
     // process that owns it.
     int well_was_shut = 0;
     for (const auto& well : well_container_generic_) {
-        if (well->name() == wellname && !well->wellIsStopped()) {
+        if (well->name() == wellname) {
             wellTestState().close_well(wellname, WellTestConfig::Reason::PHYSICAL, simulation_time);
             well_was_shut = 1;
             break;
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -449,7 +449,7 @@ namespace Opm
                       const GroupState& group_state,
                       DeferredLogger& deferred_logger)
     {
-        if (!this->isOperableAndSolvable())
+        if (!this->isOperableAndSolvable() && !this->wellIsStopped())
             return;
 
         // keep a copy of the original well state
@@ -510,6 +510,7 @@ namespace Opm
         }
         this->changed_to_open_this_step_ = false;
         const bool well_operable = this->operability_status_.isOperableAndSolvable();
+
         if (!well_operable && old_well_operable) {
             if (this->well_ecl_.getAutomaticShutIn()) {
                 deferred_logger.info(" well " + this->name() + " gets SHUT during iteration ");
@@ -539,7 +540,7 @@ namespace Opm
     void
     WellInterface<TypeTag>::addCellRates(RateVector& rates, int cellIdx) const
     {
-        if(!this->isOperableAndSolvable())
+        if(!this->isOperableAndSolvable() && !this->wellIsStopped())
             return;
 
         for (int perfIdx = 0; perfIdx < this->number_of_perforations_; ++perfIdx) {
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1223,7 +1223,7 @@ namespace Opm {
         ConvergenceReport local_report;
         const int iterationIdx = ebosSimulator_.model().newtonMethod().numIterations();
         for (const auto& well : well_container_) {
-            if (well->isOperableAndSolvable() ) {
+            if (well->isOperableAndSolvable() || well->wellIsStopped()) {
                 local_report += well->getWellConvergence(this->wellState(), B_avg, local_deferredLogger, iterationIdx > param_.strict_outer_iter_wells_ );
             } else {
                 ConvergenceReport report;


totto82:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1.001    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 1.001    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 1.009    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.996    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 0.996    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1598

totto82:
I added the milestone flag since this fixes a bug in the code for wells with STOP.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -507,7 +507,10 @@ namespace Opm
         }
         this->changed_to_open_this_step_ = false;
         const bool well_operable = this->operability_status_.isOperableAndSolvable();
-        if (!well_operable && old_well_operable) {
+
+        if (this->wellIsStopped() && !changed_to_stopped_this_step_) {
+            // we don't want to reopen stopped wells as they are already open
```
GitPaean, line 512: I do not have the background related to the targeting situation the PR is trying to solve. But an empty `if` branch does not look good in the code level.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -2067,7 +2067,7 @@ forceShutWellByName(const std::string& wellname,
     // process that owns it.
     int well_was_shut = 0;
     for (const auto& well : well_container_generic_) {
-        if (well->name() == wellname && !well->wellIsStopped()) {
+        if (well->name() == wellname) {
```
GitPaean, line 2070: this change is good to me. A stopped well should be able to be shut.

GitPaean:
This PR raises a relatively fundamental question.  Should we check the operability for a stopped well?

A well is stopped, is it possible the well has to be shut and can not just be stopped？


Basically, the main part of the PR is suggesting that a stopped well, the operability checking result can be discarded because it is stopped.  I think it might be easy just not to check the operability for stopped well. 

And the following part, 
```c++
-        if (well->name() == wellname && !well->wellIsStopped()) {
+        if (well->name() == wellname) {
```
is trying to shut a stopped well. 

It is a little difficult to tell what is the good way to go regarding stopped wells with false operability.

totto82:
>A well is stopped, is it possible the well has to be shut and can not just be stopped？

The problem happens when a stopped well do not converge. Cross-flowing wells can be hard to solve and lead to oscillatory solutions. This happens for some of the Norne prediction cases. With this PR also the stopped wells are shut if they cannot converge after repeated time-step chopping. I think this is more constant with how we treat other wells, and moreover it avoids failed simulations.

GitPaean:
> The problem happens when a stopped well do not converge. Cross-flowing wells can be hard to solve and lead to oscillatory solutions. This happens for some of the Norne prediction cases. With this PR also the stopped wells are shut if they cannot converge after repeated time-step chopping. I think this is more constant with how we treat other wells, and moreover it avoids failed simulations.

Thanks for the explanation.  Then the following is good.
 ``` c++
-        if (well->name() == wellname && !well->wellIsStopped()) {
+        if (well->name() == wellname) {
```

Then coming back to the other point, 
> it might be easy just not to check the operability for stopped well.

Do you think we can just skip checking the operability for the stopped wells and achieve the same purpose of the PR. By skipping, we save some calculations.

**Comment by totto82** [comment]
Could you explain what you mean by

> checking the operability for the stopped wells

Do you mean the convergence?

**Comment by GitPaean** [comment]
with this 
```c++
if(!this->isOperableAndSolvable() && !this->wellIsStopped())
```
Basically for the stopped wells, the operability does not matter. 

> Do you think we can just skip checking the operability for the stopped wells and achieve the same purpose of the PR. By skipping, we save some calculations.

I am suggesting not to do the operability checking for the stopped wells at all to save the computation cost.

totto82:
> I am suggesting not to do the operability checking for the stopped wells at all to save the computation cost.

But that is already the case in master. I don't change that in this PR. 

https://github.com/OPM/opm-simulators/blob/14866907384659cac8a239e855e6e16b4ed70555/opm/simulators/wells/WellInterface_impl.hpp#L584

What I changed by adding `!this->well_stopped() ` are 1) inner iterations for stopped wells 2) check convergence.

opm/simulators/wells/WellInterface_impl.hpp
totto82: I will fix.

atgeirr: I agree the parsing of this in my head is hard, assuming it will be improved. But I also do not understand the comment?

atgeirr (commented):
Maybe I understand your point: as a stopped well does not get updated solvable/operable information, you treat it as solvable/operable for some purposes, instead of using the state at the time when it was stopped. Is that correct?

Plus the change that stopped wells can now be shut due to convergence trouble, which interacts with the other changes (to have updated convergence for example).

Assuming I understood this correctly I am fine with this, except for a comment I do not understand.

opm/simulators/wells/WellInterface_impl.hpp
GitPaean:
```c++
if (this->wellIsStopped() && !changed_to_stopped_this_step_) 
```


For this `if` statement to be true, it needs `this->wellIsStopped() ` to be `true` and `changed_to_stopped_this_step_` to be `false`. 

from the code below that updates `changed_to_stopped_this_step_`, it looks like if `this->wellIsStopped() == true`, then `changed_to_stopped_this_step_` will always be `false`, 

Then I will question whether we need to use the `this->wellIsStopped() && !changed_to_stopped_this_step_` together. 

```c++
 512         const bool well_operable = this->operability_status_.isOperableAndSolvable();
 513         if (!well_operable && old_well_operable) {
 514             if (this->well_ecl_.getAutomaticShutIn()) {
 515                 deferred_logger.info(" well " + this->name() + " gets SHUT during iteration ");
 516             } else {
 517                 if (!this->wellIsStopped()) {
 518                     deferred_logger.info(" well " + this->name() + " gets STOPPED during iteration ");
 519                     this->stopWell();
 520                     changed_to_stopped_this_step_ = true;
 521                 }
 522             }
```

GitPaean:
> But that is already the case in master. I don't change that in this PR.

> [opm-simulators/opm/simulators/wells/WellInterface_impl.hpp](https://github.com/OPM/opm-simulators/blob/14866907384659cac8a239e855e6e16b4ed70555/opm/simulators/wells/WellInterface_impl.hpp#L584)

> Line 584 in [1486690](https://github.com/OPM/opm-simulators/commit/14866907384659cac8a239e855e6e16b4ed70555)

>  if (this->wellIsStopped() && !changed_to_stopped_this_step_) { 
> What I changed by adding !this->well_stopped() are 1) inner iterations for stopped wells 2) check convergence.


The comment https://github.com/OPM/opm-simulators/pull/3872#discussion_r855590439 also applies to the line above while it is not part of this PR.

GitPaean:
> But that is already the case in master. I don't change that in this PR.
> What I changed by adding !this->well_stopped() are 1) inner iterations for stopped wells 2) check convergence.

Sorry that I am slow to understand the details. 

```c++
-        if (!this->isOperableAndSolvable())
+        if (!this->isOperableAndSolvable() && !this->wellIsStopped())
```
For this change to make a difference, it needs `this->isOperableAndSolvable() == false` and `this->wellIsStopped() == true`. 

I want to know the exact scenario that this PR is targeting. If we are already not checking the operability for STOPPed wells (`this->wellIsStopped() == true`), how the well opearbility is false? (`this->isOperableAndSolvable() == false`)

totto82:
> For this change to make a difference, it needs this->isOperableAndSolvable() == false and this->wellIsStopped() == true.

A STOPed wells is always operable (since we don't check the operability) but it may not be solvable (if we set ShutUnsolvableWells = true (default is false but see PR https://github.com/OPM/opm-simulators/pull/3859/files and also my initial comment in the PR (Some fixups for combination STOP + shut_unsolvable_wells=true (default is false)) 

If wells are STOPed due to their unsolvability we still want their perforations rate to be added to the reservoir equations, and we also want to both check their convergence and do inner iterations on them. Hence the `&& wellIsStopped()`

opm/simulators/wells/WellInterface_impl.hpp
totto82: > from the code below that updates changed_to_stopped_this_step_, it looks like if this->wellIsStopped() == true, then changed_to_stopped_this_step_ will always be false,

I agree there seems to be something strange going on here. I will investigate further. The reason for the extra `if ` was to avoid duplicated "......gets REVIVED during iteration" output. But I agree that even without the extra `if ` the message should not be duplicated. I will remove and see if I am able to reproduce the issue again.

totto82: I re-checked without the `if ` and there is no difference regarding output. So I just removed it. Thanks for noticing @GitPaean. I guess it covered up some other issue in an earlier version of the code I had.

totto82:
jenkins build this please

totto82:
@GitPaean @atgeirr Can this be merged now?

GitPaean:
Sorry for not coming to this earlier. I see it is merged now, just leaving comments here, 

> A STOPed wells is always operable (since we don't check the operability) but it may not be solvable 

Thanks for the explanation. 

> If wells are STOPed due to their unsolvability we still want their perforations rate to be added to the reservoir equations, and we also want to both check their convergence and do inner iterations on them. Hence the && wellIsStopped()

1. not sure whether using solved perforation rates are good idea because they are generally unpredictable. 
2. instead of adding extra logical operations, you can change code to make sure for STOPPed wells always return `True` for `isOperableAndSolvable` function or not setting it to be unsolvable (only changing the code in one location and within one function).  Adding the extra logical operations in multiple places (with the negative operation also), will make code harder to read and maintain, and more error-prone.

atgeirr:
> 2\. instead of adding extra logical operations, you can change code to make sure for STOPPed wells always return `True` for `isOperableAndSolvable` function or not setting it to be unsolvable (only changing the code in one location and within one function).  Adding the extra logical operations in multiple places (with the negative operation also), will make code harder to read and maintain, and more error-prone.

I think these are good points, will you have a try at that refactoring?

totto82:
>  instead of adding extra logical operations, you can change code to make sure for STOPPed wells always return True for isOperableAndSolvable function or not setting it to be unsolvable (only changing the code in one location and within one function

Note that there are some places where `isOperableAndSolvable `should return False even for STOPed wells. But adding a version `isOperableAndSolvableAndNotStopped `or similar would indeed be an improvement. Note also that there may be places where the current code is buggy due to missing isSTOPed checks.

GitPaean:
> Note that there are some places where isOperableAndSolvable should return False even for STOPed wells. But adding a version isOperableAndSolvableAndNotStopped or similar would indeed be an improvement. Note also that there may be places where the current code is buggy due to missing isSTOPed checks.

Yeah. Then it is a design question whether so complicated logic is necessary.  When too complicated, then probably something can go wrong easily and it will be too difficult for other people touch. 

> Note that there are some places where isOperableAndSolvable should return False even for STOPed wells.

Do you what circumstances are these?

> Note also that there may be places where the current code is buggy due to missing isSTOPed checks.

Yes, that is exactly what the comments are talking about. Manually adding this in multiple places is error-prone.

PR #3874, opened 2022-04-19
only check if wells has negative potentials if check_well_operability=true(default)

totto82:


totto82:
jenkins build this please

totto82:
I think this also should go into the release. It would make it possible to disallow shutting of wells with negative potentials using running a run-time parameter.

atgeirr (commented):
I think this is good. Since you made the message more serious (info->warning): is it likely to trigger often and repeatedly for the same well? If so, consider adding a tag to silence repeats in the stdout, possibly including the well name in the tag.

totto82:
jenkins build this please

totto82:
TAG added. Will merge.

PR #3875, opened 2022-04-20
Radial and Spider Grid Keywords Not Supported

OPMUSER:
Add the DR, DTHETA and OUTRAD keywords as unsupported keywords with messages to enable the user to debug the input deck.

Back port candidate for release.


opm/simulators/utils/UnsupportedFlowKeywords.cpp
```diff
@@ -140,8 +140,10 @@ const KeywordValidation::UnsupportedKeywords& unsupportedKeywords()
         {"DPGRID", {false, std::nullopt}},
         {"DPKRMOD", {false, std::nullopt}},
         {"DPNUM", {false, std::nullopt}},
+        {"DR", {false, std::string{"Use the DRV keyword instead"}}},
```
bska, line 143: Is it really appropriate to use `critical = false` here?  There is currently no way we can construct a grid from (`DR`, `DTHETA`, `OUTRAD`) so that would suggest we should stop the input reading very quickly if we encounter any of those keywords.

OPMUSER: Yes, I think it is okay to use `critical = false`, because the errors are caught later on. For the DR and DTHETA keywords the current message is like this:
```
Error: 
An error occurred while creating the reservoir properties
Internal error: The current implementation *must* have theta values specified using the DTHETAV keyword 
```
But for OUTRAD we get this:

```
Error: 
An error occurred while creating the reservoir properties
Internal error: Not all keywords required for cylindrical grids present 
```
So I'm just trying to be consistent with having the validation for all three being performed via UnsupportedFlowKeywords.cpp, instead having one there another two there.  I elected to leave the existing messages as they are for the time being for DR and THETA.

blattms: Seems like this resolves the concerns. Thanks for the input.

PR #3868, opened 2022-04-08
Well testing of gas lift wells

hakonhagland:
Assigns a maximum ALQ value to each GLIFT producer when doing well testing in `beginTimeStep()`. This is done to allow the well to attain a positive potential and thereby be considered as open. Then, later in the timestep, when `assemble()` is called, the full gas lift optimization procedure can adjust the ALQ to its "correct" value.

It is also observed that in some cases when gas lift is switched off by setting ALQ to zero, and later in the schedule is switched back on again, it might not be possible to determine bhp from thp (the iteration fails to converge) for low small ALQ values. Instead of aborting the gas lift optimization, we should try increasing ALQ until we get convergence or until the maximum ALQ for the well is reached.

With this change to the code, test case [GASLIFT-02](https://github.com/OPM/opm-tests/blob/master/gaslift/GASLIFT-02.DATA) runs to completion and gives the following result for gas lift injection rate for group `PLAT-1`:

image

PR #3876, opened 2022-04-20
WIP: Fix assertion error for GASLIFT-02

hakonhagland:
Currently [GASLIFT-02](https://github.com/OPM/opm-tests/blob/master/gaslift/GASLIFT-02.DATA) fails with an assertion error:
```
Starting time step 0, stepsize 16 days, at day 32/60, date = 01-Feb-2021
Restart file written for report step  1/58, date = 01-Feb-2021 00:00:00
flow: /home/hakon/test/opm/opm-simulators/opm/simulators/wells/WellInterfaceGeneric.cpp:655: Opm::WellInterfaceGeneric::computeBhpAtThpLimitProdCommon(const std::function<std::vector<double>(double)>&, const Opm::SummaryState&, double, double, double, Opm::DeferredLogger&) const::<lambda(const std::vector<double>&)>: Assertion `rates.size() == 3' failed.
[hakon-Precision-7530:716293] *** Process received signal ***
[hakon-Precision-7530:716293] Signal: Aborted (6)
```
VFP calculations for two phase gas lift does not work unless the rates are adapted to include a zero gas rate.

I'll mark this as WIP until the issue with the failed test for `compareParallelSim_flow+SPE1CASE1_BRINE` on the jenkins server has been resolved, see #3868 for more information.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -2388,6 +2388,7 @@ namespace Opm
             // solution
             std::vector<double> rates(3);
             computeWellRatesWithBhpIterations(ebos_simulator, bhp, rates, deferred_logger);
+            this->adaptRatesForVFP(rates);
             return rates;
         };
 


hakonhagland:
jenkins build this please

hakonhagland:
Strange.. The jenkins build does not fail like in #3868 . 

@totto82 Should we just merge this then?

blattms:
> Strange.. The jenkins build does not fail like in https://github.com/OPM/opm-simulators/pull/3868 .

That's the thing with uninitialized values: You never know what they are initialized to as it depends on what the memory was for before. Hence might fail or might not.

hakonhagland:
> That's the thing with uninitialized values

@blattms Yes I guess this could be due to an uninitialized variable. But how can we best debug this? It seems difficult to me since we cannot reproduce the case consistently

totto82:
Thanks for the fix.

PR #3880, opened 2022-04-21
Fix GPU tests, use Dune as reference instead of hardcoded

Tongdongq:
The previous GPU output was tested against hardcoded values, which were incorrect anyway.
Now it first runs a solve in Dune to get the expected output.


blattms:
jenkins build this please

PR #3882, opened 2022-04-21
[backport] Fix GPU tests, use Dune as reference instead of hardcoded

blattms:
backport of #3880 to release.


blattms:
works on my system

PR #3886, opened 2022-04-25
Extract Block-Level Component-in-Place Summary Values

bska:
This PR calculates the various BxIP* summary output values such as `BOIP` (block-level oil-in-place) and `BGIPL` (block-level gas-in-place in liquid phase).  While here, also extract the block-level dissolved gas-oil-ratio (`BRS`) and vaporised oil-gas-ratio (`BRV`).

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -27,13 +27,6 @@
 #ifndef EWOMS_ECL_OUTPUT_BLACK_OIL_MODULE_HH
 #define EWOMS_ECL_OUTPUT_BLACK_OIL_MODULE_HH
 
-#include <array>
-#include <numeric>
-#include <optional>
-#include <stdexcept>
-#include <type_traits>
-#include <utility>
-
 #include <opm/models/blackoil/blackoilproperties.hh>
 
 #include <opm/models/utils/propertysystem.hh>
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

ebos/ecloutputblackoilmodule.hh
```diff
@@ -142,27 +145,32 @@ public:
...
-                    this->blockData_[key] = 0.0;
-                }
+        for (const auto& node : this->simulator_.vanguard().summaryConfig()) {
+            if ((node.category() == SummaryConfigNode::Category::Block) &&
+                collectToIORank.isCartIdxOnThisRank(node.number() - 1))
+            {
+                this->blockData_.emplace(std::piecewise_construct,
+                                         std::forward_as_tuple(node.keyword(),
+                                                               node.number()),
+                                         std::forward_as_tuple(0.0));
```
blattms, line 159: Shouldn't 
```
this->blockData_.try_emplace( {node.keyword(), node.number()),
                                                  0.0);
```
work since c++17?

ebos/ecloutputblackoilmodule.hh
```diff
@@ -579,14 +587,75 @@ public:
...
+                            val.second = getValue(fs.saturation(waterPhaseIdx));
+                        }
+                        else {
+                            val.second = getValue(fs.saturation(gasPhaseIdx));
+                        }
+
+                        // Include active pore-volume.
+                        val.second *= elemCtx.simulator().model().dofTotalVolume(globalDofIdx)
+                            * getValue(intQuants.porosity());
+                    }
```
blattms, line 611: Might be personal taste, but I don't see an improvement here. Just more branches. feel free to ignore, though

blattms (commented):
Looks good, Just would prefer the using the clearer c++17 `try_emplace`

ebos/ecloutputblackoilmodule.hh
bska: > Shouldn't [`try_emplace()`] work since c++17?

Yes, but that doesn't do the same thing as `emplace(piecewise_construct, ...)`.  `Try_emplace()` with an initialiser list will create an instance of the key type, in this case a `pair<string, int>`, and then create a "node" by copying the key and the value into it.  On the other hand, `emplace(piecewise_construct, ...)` will create the node in place and not copy anything.

ebos/ecloutputblackoilmodule.hh
bska: > I don't see an improvement here, just more branches.

That's fair, I guess.  I still consider it an improvement because I don't have to repeat the active pore-volume expression for all cases.  Doing it this way is less maintenance in my opinion.

blattms: Well, we disagree. One could also save some lines:
```
al.second = elemCtx.simulator().model().dofTotalVolume(globalDofIdx)
                            * getValue(intQuants.porosity());
if (key.first == "BOPV") {
    val.second *= getValue(fs.saturation(oilPhaseIdx));
}
else if (key.first == "BWPV") {
      val.second *= getValue(fs.saturation(waterPhaseIdx));
}
else {
      val.second *= getValue(fs.saturation(gasPhaseIdx));
}
```

blattms: and now that i look at it I see that the above statement is wrong.

PR #3887, opened 2022-04-25
use alternative init for wells with all rates less than eps

totto82:
This fixes (avoids premature shutting of a well) a corner case where a well is controlled by a group that has zero rate. Due to numerical instability the well still gets an epsilon rate and skip the improved initialization used for wells with trivial rate. Adding an epsilon in the check is sufficient to keep the well open and ready for production when the group gets a non-trivial target.

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -1053,8 +1053,9 @@ namespace Opm
         // if more than one nonzero rate.
         auto& ws = well_state.well(this->index_of_well_);
         int nonzero_rate_index = -1;
+        const double floating_point_error_epsilon = 1e-14;
         for (int p = 0; p < this->number_of_phases_; ++p) {
-            if (ws.surface_rates[p] != 0.0) {
+            if (std::abs(ws.surface_rates[p]) > floating_point_error_epsilon) {
                 if (nonzero_rate_index == -1) {
                     nonzero_rate_index = p;
                 } else {


totto82:
jenkins build this please

totto82:
Some difference in solution is expected since the initialization is changed. They are however minor and manual inspection points to difference in time-stepping. i.e. it should be save to update the reference data. 

image

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1053,7 +1053,7 @@ namespace Opm
         auto& ws = well_state.well(this->index_of_well_);
         int nonzero_rate_index = -1;
         for (int p = 0; p < this->number_of_phases_; ++p) {
-            if (ws.surface_rates[p] != 0.0) {
+            if (std::abs(ws.surface_rates[p]) > 1e-14) {
```
atgeirr, line 1056: That is a very low epsilon, I am surprised it had an observable effect! It also seems somewhat randomly chosen? As this is intended to account for calculation differences that do not depend on the magnitude of rates but on floating point errors etc. (at least that is how I interpret it), perhaps we should give it a name indicating this, such as `floating_point_error_epsilon`?

atgeirr (approved):

totto82:

opm/simulators/wells/WellInterface_impl.hpp
totto82: It is sufficiently large to fix the issue I encountered, but we may want to consider make it a bit larger if we get into convergence issues for wells with very small rates. Lets cross that bridge when we get there...

totto82:
jenkins build this opm-tests=752 please

PR #3888, opened 2022-04-25
fix trivial group target

totto82:
This fixes another fallout related to trivial group rates. See #3887 This time the well control starts oscillating between group (with target 0) and THP. This PR explicitly prevent switching to THP control if the target is zero. 

At the current stage only the case where the group target is zero is flagged, we may want to consider also flagging cases where the well has trivial target.


totto82:
jenkins build this please

PR #3895, opened 2022-04-27
Spelling fix geomtry -> geometry.

blattms:
Actually one should rename the function used from opm-common, too.
But that is for later...


blattms:
jenkins build this please

bska (approved):

PR #3906, opened 2022-05-02
change default reg factor for MSW to 100

totto82:
See discussion in  #3859

PR #3864, opened 2022-04-06
regularize MSW equation when iter > strict_inner_iter

totto82:
Attempt to improve convergence for problematic MSW wells. With default parameters i.e. 
regularization_factor_ms_wells_ = 1.0 this PR does not have any effect. 
The idea is to increase the default regularization factor to > 1.0 in later PR.

PR #3909, opened 2022-05-03
reset regularize_ before starting the inner iterations

totto82:
Should not affect the simulator with default setup, but see #3906


totto82:
jenkins build this please

totto82:
This PR fixes an issue introduced in https://github.com/OPM/opm-simulators/pull/3864
With this change the regularization terms is only applied after the strict number of inner iterations are done. Without this change, the regularization term is used in all subsequent linearizations until the next report step when reguarize_ is reset to false.

totto82:
I will self merge this in order to test it with #3906

Issue #3846, opened 2022-03-22
SKIPREST


loveduah1:
```
   61 Reading WELSPECS in PROJ1.DATA line 364
   62 Reading COMPDAT  in PROJ1.DATA line 374
   63 Reading WCONPROD in PROJ1.DATA line 385
   64 Reading WCONINJE in PROJ1.DATA line 394
   65 Reading TSTEP    in PROJ1.DATA line 405

Warning: Unsupported keywords or keyword items:

  NOECHO: keyword not supported
  In file: PROJ1.DATA, line 75
  Will continue

  ECHO: keyword not supported
  In file: PROJ1.DATA, line 107
  Will continue

Creating grid from keywords DX, DY, DZ and TOPS

Error: Problem with keyword RESTART
In PROJ1.DATA line 255
The restart file: PROJ1.X0002 does not exist

Error: Unrecoverable errors were encountered while loading input
```
[SKIPREST.zip](https://github.com/OPM/opm-simulators/files/8327453/SKIPREST.zip)

Please SKIPREST keyword returns the error above. attached is the file I ran.

bska:
> Please `SKIPREST` keyword returns the error above

In this case it's not actually `SKIPREST`.  Rather it's the `RESTART` keyword:

> ```
> Error: Problem with keyword RESTART
> In PROJ1.DATA line 255
> The restart file: PROJ1.X0002 does not exist
> ```

The `RESTART` keyword in your input file is
```
SOLUTION
-- -------------------------------------------------------------------------
--item #:      1        2         3          4    
RESTART
            PROJ1      2         1*         1*  /
```
Moreover, you've got `UNIFOUT` in your `RUNSPEC` section which means that the base run produces unified output files (`PROJ1.SMSPEC`, `PROJ1.UNSMRY`, and `PROJ.UNRST`).

Did you already produce a base run that you're trying to restart from?  If not, the `RESTART` keyword is not appropriate here.  If you **did** produce a base run, do you have the `PROJ1.UNRST` file somewhere?  If you do, then please insert the keyword `UNIFIN` next to the `UNIFOUT` keyword in the `RUNSPEC` section.  You will also use a more complete path if the `PROJ1.UNRST` file is in a different location than your `PROJ1.DATA` file.

[...]

PR #3912, opened 2022-05-10
Emit Actual Message on Input Failure

bska:
This gives more information to the user and hopefully aids them in resolving the underlying issue.

### DIFF
--- a/opm/simulators/utils/readDeck.cpp
+++ b/opm/simulators/utils/readDeck.cpp
@@ -548,7 +548,7 @@ void Opm::readDeck(Opm::Parallel::Communication    comm,
 
     if (! parseSuccess) {
         if (comm.rank() == 0) {
-            OpmLog::error("Unrecoverable errors were encountered while loading input");
+            OpmLog::error(fmt::format("Unrecoverable errors while loading input: {}", failureMessage));
         }
 
 #if HAVE_MPI


bska:
When combined with OPM/opm-common#3027, this PR changes the diagnostic message
```
Error: Unrecoverable errors were encountered while loading input
```
into
```
Error: Unrecoverable errors were encountered while loading input: Problem with keyword DATES
In PROJ2_RESTART2.DATA line 408
Keyword is not properly terminated
```
when faced with the underlying issue presented in [#3846](https://github.com/OPM/opm-simulators/issues/3846#issuecomment-1121927690)

bska:
jenkins build this please

OPMUSER:
@bska This is very good, this will fix a ton strange error messages. One thing with the first line of the message, that is:

`Error: Unrecoverable errors were encountered while loading input: Problem with keyword DATES`

Is 92 characters long, do you think we should shorten this to less than or equal to 80 characters for terminal output? For example:

`Error: Unrecoverable errors while loading input: Problem with keyword DATES`

Will this fix still work with multi-line input, for example:

```
--
--       DEFINE MICROBIALLY INDUCED CALCITE PRECIPITATION PARAMETERS
--
--       DEN     DEN     DET     CRI     FIT     HVE     HVE     MGR         
--       BIOF    CAL     RAT     POR     FAC     OXY     URE     RAT         
--       ------  ------  ------  ------  ------  ------  ------  ------  ------ 
--       MOX     MUR     MUR     MIA     MID     MIN     OXC     TOL     YGR
--       CON     CON     UTI     RAT     RAT     PER     FAC     CLO     COE
--       ------  ------  ------  ------  ------  ------  ------  ------  ------   
MICPPARA    
         35.     2710.   2.6E-13 0.1     3.      2E-5    21.3    3.6      
         0.04    60.     1391.   0.074   0.028  1.013E-5 0.5     1E-4    0.5    /

and SGOFLET, SWGFLET and SWOFLET keywords spanning multiple lines. 

```
Just asking because I cannot test.

bska:
> One thing with the first line of the message[...] is 92 characters long, do you think we should shorten this to less than or equal to 80 characters for terminal output?

I do&mdash;that's a really good suggestion.  I've pushed an update to match your suggested alternative.  Incidentally, this update makes the length of the first line 70 + number of characters in the keyword name which will be below 80 unless we have 10-character keyword names at some point.

> Will this fix still work with multi-line input,

Yes.  The detection is based on the terminating `/` character being present or absent.  We do not care about individual lines at this point.

bska:
jenkins build this please

blattms:
thanks a lot really appreciated. Looks great as always. just one more jenkins dance needed.

blattms:
jenkins build this please

bska:
jenkins build this please

PR #3914, opened 2022-05-18
removing unused well_state0 in iterateWellEqWithControl for MSW

GitPaean:
It is possible we actually need it to recover when the iteration not converged, but it is not used in the current form and not sure the usage of this variable was deleted intentionally or unintentionally.


GitPaean:
jenkins build this please

bska (approved):

PR #3910, opened 2022-05-05
Dynamically update Opm::Schedule from Python

hakonhagland:
Depends on upstream https://github.com/OPM/opm-common/pull/3020

Adds some methods that enables `Opm::Schedule` to be dynamically modified from Python. A test case, `test_schedule.py` illustrates the use case.

PR #3918, opened 2022-05-20
Dynamically update Opm::Schedule from Python (part II)

hakonhagland:
Depends on PR #3910 (which should be merged first) and upstream https://github.com/OPM/opm-common/pull/3031

Dynamically update schedule from Python (part II). Continues the work in #3910 (part I).

Adds some test cases that show how injection properties in `Opm::Schedule` can be dynamically modified from Python. 

In addition, some improvements to the build system (`CMakeLists.txt`) are introduced. The first, is to enable `make` and `make all` to pick up changes to the Python test files and test data and copy the modified files into the binary directory. The second, is to have `make install` install the Python simulators library alongside the Python bindings in `opm-common`.


hakonhagland:
jenkins build this opm-common=3031 please

hakonhagland:
I see that the jenkins build shows the cmake error (which I am not able to reproduce locally):
```
-- Found pybind11: /usr/include (found version "2.6.2" )
CMake Error at /usr/lib/cmake/pybind11/pybind11NewTools.cmake:196 (python3_add_library):
  Unknown CMake command "python3_add_library".
  python/simulators/CMakeLists.txt:13 (pybind11_add_module)
```
On my computer (Ubuntu 21.10, cmake version 3.18.4) I have the command `__Python3_add_library()` located in the file
`/usr/share/cmake-3.18/Modules/FindPython3.cmake`. It seems this is the command that is called locally. I wonder why it does not work on the Jenkins server?

blattms (commented):
Why so many changes to the build system? Are they all needed?

python/simulators/CMakeLists.txt
```diff
@@ -67,3 +60,22 @@ if(OPM_ENABLE_PYTHON_TESTS)
...
 endif()
+
+# NOTE: instead of using file( COPY ...) which copies the files at configure time (not at build time)
+#  we should add copying of the files at build time such that running "make" or "make all" will
+#  update the files if they have been modified. We use the install.py script in opm-common, see also
+#  CMakeLists.txt in opm-common
+add_custom_target(copy_python ALL
+    COMMAND ${PYTHON_EXECUTABLE} ${opm-common_DIR}/python/install.py
+      ${PROJECT_SOURCE_DIR}/python ${PROJECT_BINARY_DIR} 0)
+
```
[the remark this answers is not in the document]
hakonhagland: > Are you sure that this works for an installed opm-common? 

@blattms Thanks for the comment. I always use this folder structure when  building `opm-simulators`: The other opm-modules are located in an immediate parent folder. For example, the location of the opm-common git repository is `../opm-common` relative to `opm-simulators`. Could you share how do you build against an "installed opm-common" ? Do we need to consider other options for building `opm-simulators` ? 

>  instead of the obscure python script

My idea was to avoid duplicating code in `opm-common`, see [line 413](https://github.com/OPM/opm-common/blob/82a5e7192e4ff621ca3ee5098a7d5d3a2f9d54de/CMakeLists.txt#L413) It seemed like a good idea since it also avoided listing all the files to copy (in the future the number of files may increase) and the python script searches and creates sub folders as needed. What do you think?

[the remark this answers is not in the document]
akva2: Yes sibling builds is only one way of building, jenkins for instance does not use it. You just do like you do when you build any piece of software on linux, you build opm-common and call install from there, rinse repeat for every module.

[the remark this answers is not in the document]
blattms: @hakonhagland take a look at OPM/opm-common#3034 on how to do this with Cmake directly. I do not think doing that here is duplicating code

hakonhagland:
After rebasing this to include #3910, and also using the patch in #3923 I am able to reproduce the Jenkins error

```
-- Found pybind11: /usr/include (found version "2.6.2" )
CMake Error at /usr/lib/cmake/pybind11/pybind11NewTools.cmake:196 (python3_add_library):
  Unknown CMake command "python3_add_library".
  python/simulators/CMakeLists.txt:13 (pybind11_add_module)
```
locally. So this is progress!

hakonhagland:
jenkins build this please

hakonhagland:
According to the [documentation](https://pybind11.readthedocs.io/en/stable/cmake/index.html) we should add a `Development` component argument to `find_package(Python3)`, see latest commit. Still, this does not solve the issue locally, now I get another error:
```
CMake Error at /usr/share/cmake-3.18/Modules/FindPython/Support.cmake:3128 (add_library):
  Cannot find source file:

    SYSTEM

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx
  /usr/share/cmake-3.18/Modules/FindPython3.cmake:393 (__Python3_add_library)
  /lib/cmake/pybind11/pybind11NewTools.cmake:196 (python3_add_library)
  python/simulators/CMakeLists.txt:13 (pybind11_add_module)

```

hakonhagland:
jenkins build this please

hakonhagland:
It appears from the [documentation](https://pybind11.readthedocs.io/en/stable/compiling.html) and also by investigating [the source](https://github.com/pybind/pybind11/tree/master/tools) that the `SYSTEM` argument to `pybind11_add_module()` is deprectated and was removed after CMake version 3.12 and the introduction of the new [FindPython mode](https://pybind11.readthedocs.io/en/stable/compiling.html#findpython-mode)

With this change (see latest commit) cmake succeeds to build the makefile locally.

hakonhagland:
jenkins build this please

hakonhagland:
jenkins build this opm-common=3037 please

hakonhagland:
jenkins build this opm-common=3037 please

hakonhagland:
Added a new commit to adapt this PR to the last change to https://github.com/OPM/opm-common/pull/3037

hakonhagland:
jenkins build this opm-common=3037 please

python/simulators/CMakeLists.txt
```diff
@@ -71,3 +60,32 @@ if(OPM_ENABLE_PYTHON_TESTS)
...
+  set(_opm_common_python_script_dir "${opm-common_PYTHON_COMMON_DIR}")
+endif()
+
+# NOTE: instead of using file( COPY ...) which copies the files at configure time (not at build time)
+#  we should add copying of the files at build time such that running "make" or "make all" will
+#  update the files if they have been modified. We use the install.py script in opm-common, see also
+#  CMakeLists.txt in opm-common
+add_custom_target(copy_python ALL
+    COMMAND ${PYTHON_EXECUTABLE} "${_opm_common_python_script_dir}/install.py"
+      ${PROJECT_SOURCE_DIR}/python ${PROJECT_BINARY_DIR} 0)
```
blattms, line 76: Mhh, we already have problems on Debian's salsa-ci because our build artefacts are to big. With this change it seems we copy more to the binary directory than before (e.g. pybind). Is that really needed?

python/simulators/CMakeLists.txt
```diff
@@ -71,3 +60,32 @@ if(OPM_ENABLE_PYTHON_TESTS)
...
+# setup-install.sh - and completely bypass cmake in the installation phase.
+if (OPM_INSTALL_PYTHON)
+  include(PyInstallPrefix)  # from opm-common
+  install(TARGETS simulators DESTINATION ${DEST_PREFIX}${CMAKE_INSTALL_PREFIX}/${PYTHON_INSTALL_PREFIX}/opm)
+  install(
+    CODE "execute_process(
+      COMMAND ${PYTHON_EXECUTABLE}
+              ${_opm_common_python_script_dir}/install.py
+              ${PROJECT_BINARY_DIR}/python/opm2
+              ${DEST_PREFIX}${CMAKE_INSTALL_PREFIX}/${PYTHON_INSTALL_PREFIX} 1)")
```
blattms, line 90: Not a big fan of not using CMake (might have some strange implications if people try to reset some variables when calling make. What is the rational behind this and how was this done before?

blattms: opm2 will vanish again, right?

python/simulators/CMakeLists.txt
hakonhagland: Yes you are right, I believe the same thing is happening with opm-common, see [line 416](https://github.com/OPM/opm-common/blob/dbec8e4466b5b3293dd41067aaeb13566523bd34/CMakeLists.txt#L416) in `CMakeLists.txt`. I have submitted to new PRs, see  https://github.com/OPM/opm-common/pull/3043 and #3938  to address this.

python/simulators/CMakeLists.txt
hakonhagland: I suggest we use the `install.py` until with run into an explicit problem with it. To replace it with cmake, would require listing all the files explicitly in the `add_custom_target(copy_python ALL)`  with entries like `COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/python/some/dir/file1 ${PROJECT_BINARY_DIR}/python/some/dir` for each file, right?

>opm2 will vanish again, right?

@blattms Yes,  #3933 should remove `opm2` and replace it with `opm`

hakonhagland:
jenkins build this please

blattms:
The PR dependency chain for the Python stuff make me a bit dizzy. Let's hope I won't mess up merging :grimacing:

PR #3924, opened 2022-05-25
Steer bhp towards the solution with the highest production rate

totto82:


opm/simulators/wells/VFPProdProperties.cpp
```diff
@@ -166,7 +166,8 @@ EvalWell VFPProdProperties::bhp(const int table_id,
 
     detail::VFPEvaluation bhp_val = detail::interpolate(table, flo_i, thp_i, wfr_i, gfr_i, alq_i);
 
-    bhp = (bhp_val.dwfr * wfr) + (bhp_val.dgfr * gfr) - (bhp_val.dflo * flo);
+    bhp = (bhp_val.dwfr * wfr) + (bhp_val.dgfr * gfr) - (std::max(0.0, bhp_val.dflo) * flo);
```
atgeirr, line 169: I analysed this for a simple case, and think it is insufficient. I believe this will push the rate solution towards the point where the true inflow performance relation and the vfp curve have the same slope. This will be between the two solutions with low and high production rate.

You may be on the right track though, limiting this derivative may work, but the limit should probably depend on the IPR.

atgeirr: My analysis was not correct... The update will be in the correct direction. However, Newton acceleration is lost and oscillation may happen.

The corrected analysis points towards limiting the derivative to keep the Jacobian away from becoming singular. This condition can be written as (1 - W * f') > 0 where W is the well productivity index (more precisely the derivative with respect to bhp of the IPR), so f' > -1/W rather than f' > 0.

totto82: Cool if you can make a PR with your suggested solution so I could try it.

totto82:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1.002    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 1.012    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 0.995    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.96     |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 0.988    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1663

opm/simulators/wells/VFPProdProperties.cpp
atgeirr: Analysis indicates that while the f' > -1/W is more optimal than f' > 0, the latter condition still enlarges the Newton convergence region. Testing confirms that this works well. Should be merged when green.

totto82:

totto82:
jenkins build this opm-tests=766 please

PR #3926, opened 2022-05-31
Fix well operability

totto82:
The current check for operability using IPR is different between STW and MSW. STW return inoperable if at one of the components has negative(for producers) IPR. For MSW it checks if the sum is negative. This PR changes both so that only well where all components have negative IPRs are inoperable. I mark it as WIP and draft since I still don't understand how the well operability differs between the components.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1169,12 +1169,13 @@ namespace Opm
         if ( bhp_limit_not_defaulted || !this->wellHasTHPConstraints(summaryState) ) {
             // if the BHP limit is not defaulted or the well does not have a THP limit
             // we need to check the BHP limit
-
-            double ipr_rate = 0;
+            double total_ipr_mass_rate = 0.0;
             for (int p = 0; p < this->number_of_phases_; ++p) {
-                ipr_rate += this->ipr_a_[p] - this->ipr_b_[p] * bhp_limit;
+                const double ipr_rate = this->ipr_a_[p] - this->ipr_b_[p] * bhp_limit;
+                const double rho = FluidSystem::referenceDensity( p, Base::pvtRegionIdx() );
+                total_ipr_mass_rate += ipr_rate * rho;
             }
-            if ( (this->isProducer() && ipr_rate < 0.) || (this->isInjector() && ipr_rate > 0.) ) {
+            if ( (this->isProducer() && total_ipr_mass_rate < 0.) || (this->isInjector() && total_ipr_mass_rate > 0.) ) {
                 this->operability_status_.operable_under_only_bhp_limit = false;
             }
 
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1170,9 +1170,16 @@ namespace Opm
             // if the BHP limit is not defaulted or the well does not have a THP limit
             // we need to check the BHP limit
             double total_ipr_mass_rate = 0.0;
-            for (int p = 0; p < this->number_of_phases_; ++p) {
-                const double ipr_rate = this->ipr_a_[p] - this->ipr_b_[p] * bhp_limit;
-                const double rho = FluidSystem::referenceDensity( p, Base::pvtRegionIdx() );
+            for (unsigned phaseIdx = 0; phaseIdx < FluidSystem::numPhases; ++phaseIdx)
+            {
+                if (!FluidSystem::phaseIsActive(phaseIdx)) {
+                    continue;
+                }
+
+                const unsigned compIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::solventComponentIndex(phaseIdx));
+                const double ipr_rate = this->ipr_a_[compIdx] - this->ipr_b_[compIdx] * bhp_limit;
+
+                const double rho = FluidSystem::referenceDensity( phaseIdx, Base::pvtRegionIdx() );
                 total_ipr_mass_rate += ipr_rate * rho;
             }
             if ( (this->isProducer() && total_ipr_mass_rate < 0.) || (this->isInjector() && total_ipr_mass_rate > 0.) ) {
@@ -1267,15 +1274,12 @@ namespace Opm
             // the well index associated with the connection
             const double tw_perf = this->well_index_[perf]*ebos_simulator.problem().template rockCompTransMultiplier<double>(int_quantities, cell_idx);
 
-            // TODO: there might be some indices related problems here
-            // phases vs components
-            // ipr values for the perforation
             std::vector<double> ipr_a_perf(this->ipr_a_.size());
             std::vector<double> ipr_b_perf(this->ipr_b_.size());
-            for (int p = 0; p < this->number_of_phases_; ++p) {
-                const double tw_mob = tw_perf * mob[p] * b_perf[p];
-                ipr_a_perf[p] += tw_mob * pressure_diff;
-                ipr_b_perf[p] += tw_mob;
+            for (int comp_idx = 0; comp_idx < this->num_components_; ++comp_idx) {
+                const double tw_mob = tw_perf * mob[comp_idx] * b_perf[comp_idx];
+                ipr_a_perf[comp_idx] += tw_mob * pressure_diff;
+                ipr_b_perf[comp_idx] += tw_mob;
             }
 
             // we need to handle the rs and rv when both oil and gas are present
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -1065,14 +1065,15 @@ namespace Opm
...
             // if the BHP limit is not defaulted or the well does not have a THP limit
             // we need to check the BHP limit
-
+            bool operable_under_bhp = false;
             for (int p = 0; p < this->number_of_phases_; ++p) {
-                const double ipr_rate = this->ipr_a_[p] - this->ipr_b_[p] * bhp_limit;
-                if ( (this->isProducer() && ipr_rate < 0.) || (this->isInjector() && ipr_rate > 0.) ) {
-                    this->operability_status_.operable_under_only_bhp_limit = false;
-                    break;
+                double ipr_rate = this->ipr_a_[p] - this->ipr_b_[p] * bhp_limit;
```
atgeirr, line 1070: I did not understand why you removed the `const` here, but I think it was not intentional? Please also correct in the MSW implementation.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -1065,14 +1065,15 @@ namespace Opm
...
+                if ( (this->isProducer() && ipr_rate >= 0.) || (this->isInjector() && ipr_rate <= 0.) ) {
+                    operable_under_bhp = true;
                 }
             }
+            if (!operable_under_bhp)
```
atgeirr, line 1075: Nitpick: brackets.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -1065,14 +1065,15 @@ namespace Opm
...
             // we need to check the BHP limit
-
+            bool operable_under_bhp = false;
             for (int p = 0; p < this->number_of_phases_; ++p) {
-                const double ipr_rate = this->ipr_a_[p] - this->ipr_b_[p] * bhp_limit;
-                if ( (this->isProducer() && ipr_rate < 0.) || (this->isInjector() && ipr_rate > 0.) ) {
-                    this->operability_status_.operable_under_only_bhp_limit = false;
-                    break;
+                double ipr_rate = this->ipr_a_[p] - this->ipr_b_[p] * bhp_limit;
+                if ( (this->isProducer() && ipr_rate >= 0.) || (this->isInjector() && ipr_rate <= 0.) ) {
```
atgeirr, line 1071: If a well is at zero rate, I would expect that it is now more likely to be considered operable, because of numerical noise. Have you considered adding a low but nonzero threshold for this, such as 1e-8 m^3/s? Also, the same threshold could be applied to check that all phases with the wrong flow direction has a very low flow (just noise).

opm/simulators/wells/StandardWell_impl.hpp
atgeirr, line 1070: The sign of the ipr_rate is unfortunately following the opposite convention as our usual (negative flow is production). We should probably fix this (but not in this PR).

atgeirr (commented):
Overall a good change, should be merged when comments are addressed. The threshold question may need some more thought.

opm/simulators/wells/VFPProdProperties.cpp
```diff
@@ -166,7 +166,8 @@ EvalWell VFPProdProperties::bhp(const int table_id,
 
     detail::VFPEvaluation bhp_val = detail::interpolate(table, flo_i, thp_i, wfr_i, gfr_i, alq_i);
 
-    bhp = (bhp_val.dwfr * wfr) + (bhp_val.dgfr * gfr) - (bhp_val.dflo * flo);
+    bhp = (bhp_val.dwfr * wfr) + (bhp_val.dgfr * gfr) - (std::max(0.0, bhp_val.dflo) * flo);
```
atgeirr, line 169: Just to clarify: my review did not apply to this part, that discussion is separate.

opm/simulators/wells/StandardWell_impl.hpp
totto82: I think you point to some important issue here. I will try out your ideas and report back. Currently I think we keep open inoperable wells with the change suggested in this PR. So even thought it fixes an issue where wells are closed too early, we also need to make sure it shuts wells that really are inoperable.

totto82:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
totto82: I went back to a "sum" approach but this time weighted by the surface density. I.e. to make it more robust in terms of "small" values of gas flowing in the wrong direction. Seems to work fine both for field model and for the MIN_BHP_1 test.

atgeirr: That sounds reasonable. So is it exactly the reservoir rate you do then? Or a "close relative"?

totto82: It is the mass rate. There may be some issues with indexing in the code. I notices that the concept of components and phases are mixed up in the original code. So even though the code is 
`for (int p = 0; p < this->number_of_phases_; ++p) {`
i think it actually loop over the components. i.e. we may mix the water/oil ref density as the code is now. I will need to double check that before this can be undrafted.

atgeirr: I agree, it does not matter in terms of results until you add the density. Should preferably have been correct from day 1 of course...

totto82:
jenkins build this please

totto82:
jenkins build this please

bska:
> Is the "well is operable flag" handled by restart?

It is not.  We don't have anything like that in our restart files.  If we need it we

1. Must identify the item's location&mdash;probably somewhere in `IWEL`
2. Must communicate the appropriate value from the simulator to the output layer and write it out
3. Must reload it from the restart file
4. Must communicate the setting back to the simulator

totto82:
jenkins build this please

totto82:
I have manually checked the tests. 

[mpi.compareECLFiles_flow+0A_AQUCT_MODEL6](https://ci.opm-project.org/job/opm-simulators-PR-builder/3736/testReport/junit/(root)/mpi/compareECLFiles_flow_0A_AQUCT_MODEL6/)
[mpi.compareECLFiles_flow+MIN_BHP_1](https://ci.opm-project.org/job/opm-simulators-PR-builder/3736/testReport/junit/(root)/mpi/compareECLFiles_flow_MIN_BHP_1/)
[mpi.compareECLFiles_flow+MIN_BHP_2](https://ci.opm-project.org/job/opm-simulators-PR-builder/3736/testReport/junit/(root)/mpi/compareECLFiles_flow_MIN_BHP_2/)
[mpi.compareECLFiles_flow+MAX_GOR_1](https://ci.opm-project.org/job/opm-simulators-PR-builder/3736/testReport/junit/(root)/mpi/compareECLFiles_flow_MAX_GOR_1/)
[mpi.compareECLFiles_flow+MIN_QOIL_1](https://ci.opm-project.org/job/opm-simulators-PR-builder/3736/testReport/junit/(root)/mpi/compareECLFiles_flow_MIN_QOIL_1/)

For the first a well with very small rate is STOPed the last time step. 
The four next tests has larger differences, but they all seems to be improvements. i.e. wells are kept open longer are able to reopen (when they should) i.e. from my side this can be merged. Testing on field models also shows improvements. (i.e. avoids premature shutting of wells)

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1219,14 +1227,13 @@ namespace Opm
...
+        const double ref_depth = this->ref_depth_;
         for (int seg = 0; seg < nseg; ++seg) {
             // calculating the perforation rate for each perforation that belongs to this segment
             const double segment_depth = this->segmentSet()[seg].depth();
-            const double dp = wellhelpers::computeHydrostaticCorrection(ref_depth, segment_depth, this->segment_densities_[seg].value(), this->gravity_);
-            ref_depth = segment_depth;
-            seg_bhp_press_diff += dp;
+            const int outlet_segment_index = this->segmentNumberToIndex(this->segmentSet()[seg].outletSegment());
+            const double segment_depth_outlet = seg == 0? ref_depth : this->segmentSet()[outlet_segment_index].depth();
+            const double dp = wellhelpers::computeHydrostaticCorrection(segment_depth_outlet, segment_depth, this->segment_densities_[seg].value(), this->gravity_);
```
atgeirr, line 1236: At first I thought this was a nice bugfix, but now I think it maybe introduces a bug. The old code accumulates the dp values to form the hydrostatic pressure difference between the bhp reference point and the segment. I think it is mostly correct, except when encountering a fork in the well, there the new code correctly looks at the outlet segment rather than the previous segment.

However, when used further down on line 1264, the dp is used, rather than their sum accumulated up the segments. I believe that is wrong, that the pressure difference should be between segment and bhp ref, not between segment and outlet segment.

I think building a vector of segment-to-bhpref pressure differences is necessary, because to compute a lower one you need to add the difference of the outlet segment to your own. This should be a separate for loop to simplify reading of the code.

atgeirr (changes requested):
This is all good, except for what I think is a buggy bugfix, see comment.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: Seems reasonable. In `3D_MSW` all branches branch out from the top cell. i.e. the issue you point to never surface. I will fix asap.

atgeirr: I think it will surface then too, unless the number of segments in each branch is only one?

atgeirr: Yes I think that looks good. This will change the results, maybe significantly, so I fear a new round of tests might be required?

atgeirr (approved):

bska:
The updated reference solutions have been installed on the CI system.  Merging this now per prior approval to bring the overall system back to a consistent state.

PR #3929, opened 2022-06-01
Use Consistent Grid Views for Transmissibility Calculation

### LINKED ISSUE #3927: Possibly Conflicting Grid Views Employed in Load Balancing
Unless there is an externally defined load balancing routine, function `EclGenericCpGridVanguard<>::doLoadBalance_()` will [calculate face transmissibilities](https://github.com/OPM/opm-simulators/blob/26a9582ecec0b7d68a23a2bc81d07fff07ef6266/ebos/eclgenericcpgridvanguard.cc#L109) and pass those into [`CpGrid::loadBalance()`](https://github.com/OPM/opm-simulators/blob/26a9582ecec0b7d68a23a2bc81d07fff07ef6266/ebos/eclgenericcpgridvanguard.cc#L163-L165).  In this instance, the loop that calculates the transmissibilities uses an [`ElementMapper`](https://github.com/OPM/opm-simulators/blob/26a9582ecec0b7d68a23a2bc81d07fff07ef6266/ebos/eclgenericcpgridvanguard.cc#L111) defined in terms of a `GridView` that's passed as an [argument](https://github.com/OPM/opm-simulators/blob/26a9582ecec0b7d68a23a2bc81d07fff07ef6266/ebos/eclgenericcpgridvanguard.cc#L81) into `doLoadBalance_()`, but the loop itself explicitly uses [`CpGrid::leafGridView()`](https://github.com/OPM/opm-simulators/blob/26a9582ecec0b7d68a23a2bc81d07fff07ef6266/ebos/eclgenericcpgridvanguard.cc#L104).

Are these two `GridView`s ever expected/allowed to differ?  If so, isn't there a danger in using the mapper from one with the entities of the other?  If not, why do we take a `gridv` parameter in `doLoadBalance_()`?

bska:
Commit 5067ce2f27e64c05c04a634de93b91aee6685afa (initially OPM/opm-models@e22366d4a, PR https://github.com/OPM/opm-models/pull/103) implicitly assumed that the `gridv` parameter would coincide with `grid_->leafGridView()`.  Make this assumption explicit.

Resolves #3927.

### DIFF
--- a/ebos/eclcpgridvanguard.hh
+++ b/ebos/eclcpgridvanguard.hh
@@ -134,7 +134,7 @@ public:
 #if HAVE_MPI
         this->doLoadBalance_(this->edgeWeightsMethod(), this->ownersFirst(),
                              this->serialPartitioning(), this->enableDistributedWells(),
-                             this->zoltanImbalanceTol(), this->gridView(),
+                             this->zoltanImbalanceTol(),
                              this->schedule(), this->centroids_,
                              this->eclState(), this->parallelWells_, this->numJacobiBlocks());
 #endif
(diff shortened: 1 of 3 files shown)


bska:
jenkins build this please

blattms:
Thanks. Looks good. Merging.

PR #3934, opened 2022-06-07
stopped wells where cross flow is not allowed are not added to the container

totto82:
STOPPed wells without cross flow does not make sense (and does not converge...) handle them as SHUT instead.


totto82:
jenkins build this please

atgeirr (approved):
Should be merged when green.

PR #3935, opened 2022-06-08
Fix: zeros were not replaced correctly in the diagonal for GPU

Tongdongq:
The block-jacobi-ILU preconditioner did not work with `cusparseSolver`, due to zeros on the diagonal.
They were supposed to be removed by `checkZeroDiagonal()`, but this was bugged.
It used only one array for both `matrix` and `jacMatrix`, whereas they each need their own.
Also other minor changes were made.


blattms:
jenkins build this please

blattms:
Thanks for the fix.

Tongdongq:
For cusparse yes, because previously it kept falling back to Dune.
For opencl, it shouldn't matter much.

PR #3933, opened 2022-06-07
Make the Python `opm` package a namespace package

hakonhagland:
Depends on #3918 which should be merged first.
Also depends on upstream https://github.com/OPM/opm-common/pull/3040

Convert the Python `opm` package from a regular package to a namespace package such that `opm-simulators` and `opm-common` can contribute to the package from different filesystem paths. In this way, the two packages `opm.simulators` and `opm.io` (in `opm-common)`, for example, can have a different parent filesystem path.

PR #3938, opened 2022-06-10
Only copy the necessary files from the source python folder to the build folder

hakonhagland:
Depends on #3933 which should be merged first.

For background information, see: https://github.com/OPM/opm-simulators/pull/3918#pullrequestreview-997949357

Instead of copying the whole `python` folder to the build folder at build time, only copy the files that are considered to be necessary. This means the `opm`, `test`, and `test_data` directories.


hakonhagland:
jenkins build this opm-common=3043 please

blattms:
@hakonhagland Why do you have several forks (opm-simulators and opm-simulators-1)? Quite confusing, if one is looking for a branch but only set up one of them as remote.

PR #3939, opened 2022-06-13
Fix FOE definition

goncalvesmachadoc:
image


bska (approved):
Thanks @goncalvesmachadoc .  This clearly fixes an erroneous definition.

PR #3946, opened 2022-06-17
Fix: verbose timing for openclSolver

Tongdongq:
When `verbosity>=3`, the openclSolver will time certain components, like the preconditioner application, spmv, and well application, and print them.
To get accurate timings, the solver must wait for the opencl kernels to finish, which it did not before.
This uses `queue->finish()`, this works because `openclSolver` only ever uses one queue. If multiple queues were to be used, for example when overlapping copying and computation, or having two kernels simultaneously, changes have to be made.

I wanted to put `if(verbosity>=3)queue->finish();` into an inlined function before.
But I've also merged some of the timers, so that they only work when their results are actually used. The `Dune::Timer` is lightweight, but the `if(verbosity>=3)` should be even more lightweight.


blattms:
jenkins build this please

blattms:
Thanks a lot. In it goes.

PR #3953, opened 2022-06-22
Tweak opencl launch params

Tongdongq:
The ILU apply kernels launched `Nb/32` workgroups, where `Nb` is the number of blockrows in the matrix. This is wasteful, since due to level_scheduling, the number of blockrows processed in parallel is less than that. It now launches one workgroup for each blockrow.
Also changed the workgroup size to 64, this is slightly faster for a bigger case.

The ILU decomposition kernel always launched 1024 workgroups, it now also launches one workgroup for each blockrow.

Performance gains are marginal.

### DIFF
--- a/opm/simulators/linalg/bda/opencl/kernels/ILU_apply1_fm.cl
+++ b/opm/simulators/linalg/bda/opencl/kernels/ILU_apply1_fm.cl
@@ -14,7 +14,7 @@ __kernel void ILU_apply1(
     const unsigned int block_size,
     __local double *tmp)
 {
-    const unsigned int warpsize = 32;
+    const unsigned int warpsize = get_local_size(0);
     const unsigned int bs = block_size;
     const unsigned int idx_t = get_local_id(0);
     const unsigned int num_active_threads = (warpsize/bs/bs)*bs*bs;
--- a/opm/simulators/linalg/bda/opencl/openclKernels.cpp
+++ b/opm/simulators/linalg/bda/opencl/openclKernels.cpp
@@ -41,6 +41,7 @@ int OpenclKernels::verbosity;
 cl::CommandQueue *OpenclKernels::queue;
 std::vector<double> OpenclKernels::tmp;
 bool OpenclKernels::initialized = false;
+size_t OpenclKernels::preferred_workgroup_size = 0;
 
 std::unique_ptr<cl::KernelFunctor<cl::Buffer&, cl::Buffer&, cl::Buffer&, const unsigned int, cl::LocalSpaceArg> > OpenclKernels::dot_k;
 std::unique_ptr<cl::KernelFunctor<cl::Buffer&, cl::Buffer&, const unsigned int, cl::LocalSpaceArg> > OpenclKernels::norm_k;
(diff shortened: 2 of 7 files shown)


opm/simulators/linalg/bda/opencl/kernels/ILU_apply1_fm.cl
```diff
@@ -14,7 +14,7 @@ __kernel void ILU_apply1(
     const unsigned int block_size,
     __local double *tmp)
 {
-    const unsigned int warpsize = 32;
+    const unsigned int warpsize = 64;
```
blattms, line 17: Seems like for AMD this recommended, whereas INTEL (and NVIDIA) recommends 32,

Should we make this configurable or choose based on hardware vendor? Maybe use `CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE`?

opm/simulators/linalg/bda/opencl/openclKernels.cpp
```diff
@@ -373,10 +373,10 @@ void OpenclKernels::residual(cl::Buffer& vals, cl::Buffer& cols, cl::Buffer& row
     }
 }
 
-void OpenclKernels::ILU_apply1(cl::Buffer& vals, cl::Buffer& cols, cl::Buffer& rows, cl::Buffer& diagIndex, const cl::Buffer& y, cl::Buffer& x, cl::Buffer& rowsPerColor, int color, int Nb, unsigned int block_size)
+void OpenclKernels::ILU_apply1(cl::Buffer& vals, cl::Buffer& cols, cl::Buffer& rows, cl::Buffer& diagIndex, const cl::Buffer& y, cl::Buffer& x, cl::Buffer& rowsPerColor, int color, int rowsThisColor, unsigned int block_size)
```
blattms, line 376: Would be good to get rid of these long lines, if they are touched anyway.

blattms (commented):
May I ask you a favor?

For the functions definitions that you touch, anyway, can you please break the to something shorter? Like 80 characters or something that can be viewed in the github diff without scrolling?

blattms:
jenkins build this please

Tongdongq:
Ready for review, but still waiting for final tests after removing a variable.
I'm not sure how to consistently keep lines short, with many parameters or subsequent calls `a().b().c(d);`.
Is there a style guide somewhere?

blattms:
> Is there a style guide somewhere?

the answer probably depends on whom you ask. There is a clang-format file in this directory. The problem is that it AFAIK reformats whole files and might make spotting the relevant changes rather painful. There is also the (ancient) suggested [coding style in the wiki](https://wiki.opm-project.org/index.php?title=Suggested_coding_standard). I fear a bit that both are incompatible, though.

But any decent editor should be able to indent line-breaks for function declarations in a meaningful way.

Tongdongq:
Final tests passed for both AMD and NVIDIA GPUs.

I have the C/C++ extensions for VSCode, but it doesn't handle the long
```
cl::EnqueueArgs(*queue, cl::NDRange(total_work_items), cl::NDRange(work_group_size)
```
We could turn them into
```
    cl::Event event = (*ILU_apply2_k)(cl::EnqueueArgs(
                              *queue,
                              cl::NDRange(total_work_items),
                              cl::NDRange(work_group_size)
                          ),
                          vals, cols, rows, diagIndex, invDiagVals,
                          x, rowsPerColor, color, block_size,
                          cl::Local(lmem_per_work_group));
```
or try to put the `EnqueueArgs` into a separate function, but I think it'll always be meh.

opm/simulators/linalg/bda/opencl/kernels/ILU_apply1_fm.cl
Tongdongq: Now implemented

Tongdongq:
@blattms Can you restart Jenkins? I'd like to see the output.

blattms:
I don't think there was a jenkins error and I doubt that it will work as it is reverted in master.. You will need to create a new PR.

bska:
> [this PR] is reverted in master..

Correct.  I backed out this PR in #3973 since it broke the build.  One example of a build failure presented in an otherwise [unrelated build](https://ci.opm-project.org/job/opm-simulators-PR-builder/3801/console) of PR #3972.  Please look for `error:` in the console output.

PR #3958, opened 2022-06-24
Fix stdwell apply on GPU

Tongdongq:
Fixes bug introduced in 06a8b0ee in #3937.
Keeping the wells separate would cause convergence issues on GPU (both cusparse and opencl).
@blattms


blattms:
Thanks. 
This is why I think people should keep their changes as minimal as possible.

atgeirr:
Do we have any way to make Jenkins test this code?

blattms:
We can't Cloud gpu nodes are too expensive.

PR #3959, opened 2022-06-27
Warn about fault names with more than 8 characters

blattms:
This issues warning a la
```
  FAULTS: invalid value 'BlaBlubbLong_depth_copy' in record 1647 for item 1
  In file: /path/to/file.grdecl, line 4
  FAULTS(FAULTNAME): Only names of faults up to 8 characters are supported. Will ignore excess characters
```

This is of course abusing the partially supported keyword code a bit.
An alternative might be to a Deck type 'STRING8' in addition to `STRING` and `RAWSTRING` to do the same. Just not sure whether it is possible to issue a warning like here. (Would open up to handle well names the same...)

Needs OPM/opm-common#3062

### DIFF
--- a/opm/simulators/utils/ParallelRestart.cpp
+++ b/opm/simulators/utils/ParallelRestart.cpp
@@ -34,6 +34,7 @@
 #include <opm/output/data/GuideRateValue.hpp>
 #include <opm/output/data/Solution.hpp>
 #include <opm/output/data/Wells.hpp>
+#include <opm/input/eclipse/EclipseState/Util/OrderedMap.hpp>
 #include <opm/output/eclipse/EclipseIO.hpp>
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this opm-common=3062 please

blattms:
jenkins build this opm-common=3065 please

OPMUSER:
Can we change:

` FAULTS(FAULTNAME): Only names of faults up to 8 characters are supported. Will ignore excess characters`

to

 `FAULTS(FLTNAME): Only names of faults up to 8 characters are supported. Will ignore excess characters
`

In order to be consistent with the manual, as we only use a maximum of eight characters for parameter names- not a big deal.

blattms:
@OPMUSER Already noticed that myself and is part of this PR. Just used the wrong error message.

PR #3961, opened 2022-06-27
changed: use gcc pragma's to suppress strict-aliasing warning

akva2:
we don't care and this causes unwanted noise on CI.


akva2:
jenkins build this please

blattms:
I don't care. But maybe we should? Where is the warning?

akva2:
ie exactly the line surrounded by the pragmas..

blattms:
Thanks. I somehow assumed the problem was in readMatrixMarket.

PR #3964, opened 2022-06-29
handling shut valve

GitPaean:
When a valve is SHUT, the rate and pressure drop for that segment will be zero.


opm/simulators/wells/MultisegmentWellEval.cpp
```diff
@@ -1554,6 +1554,18 @@ assembleICDPressureEq(const int seg,
     // top segment can not be a spiral ICD device
     assert(seg != 0);
 
+    if(this->segmentSet()[seg].segmentType() == Segment::SegmentType::VALVE) {
+        const Valve& valve = this->segmentSet()[seg].valve();
+        if (valve.status() == Opm::ICDStatus::SHUT) { // we switch here to be a zero rate equation
+            resWell_[seg][SPres] = this->primary_variables_evaluation_[seg][GTotal].value();
+            duneD_[seg][seg][SPres][GTotal] = 1.;
+
+            auto& ws = well_state.well(baseif_.indexOfWell());
```
GitPaean, line 1563: If the valve is shut and there no active perforations for this shut branch, there might be single matrix here, similar to a closed well without active perforations.  There might need to be more development to handle that special situation.

opm/simulators/wells/MultisegmentWellEval.cpp
```diff
@@ -1554,6 +1554,18 @@ assembleICDPressureEq(const int seg,
     // top segment can not be a spiral ICD device
     assert(seg != 0);
 
+    if(this->segmentSet()[seg].segmentType() == Segment::SegmentType::VALVE) {
+        const Valve& valve = this->segmentSet()[seg].valve();
+        if (valve.status() == Opm::ICDStatus::SHUT) { // we switch here to be a zero rate equation
```
bska, line 1559: For what it's worth you can reduce the nesting here if you use C++17's `if` with an initialisation statement
```C++
if (const auto& segment = this->segmentSet()[seg];
    (segment.segmentType() == Segment::SegmentType::VALVE) &&
    (segment.valve().status() == Opm::ICDStatus::SHUT))
{
    // Handle shut valve
}
```

GitPaean: Will do that when the result is confirmed.

GitPaean:
jenkins build this please

totto82:
Thanks I will test it asap.

totto82:
Testing done. From my point of view this can be merged. Thanks. I will close the other PR

GitPaean:

tests/update_reference_data.sh
```diff
@@ -154,6 +154,7 @@ tests[udq_pyaction]="flow udq_actionx PYACTION_WCONPROD"
 tests[spe1_foam]="flow spe1_foam SPE1FOAM"
 tests[wsegsicd]="flow wsegsicd TEST_WSEGSICD"
 tests[wsegaicd]="flow wsegaicd BASE_MSW_WSEGAICD"
+tests[wsegaicd]="flow wsegvalv BASE_MSW_WSEGVALV"
```
bska, line 157: This should **probably** say `"wsegvalv"` in order to match the test name.

GitPaean: Thanks.

GitPaean:

opm/simulators/wells/MultisegmentWellEval.cpp
GitPaean: I see the reason now. I should have deleted the extra `}`.

GitPaean:
jenkins build this opm-tests=778 please

GitPaean:
I think it **can** be the proper way to handle `SHUT` valve, which is also giving desired results following the test report from @totto82 . It still has the limitation not being able to handle the `DEAD` shut branch without active connections, which will be more involved to tangle, but we can tangle it when we have the case for it. The idea is relatively clear, while might be more development.

The jenkins gives green now. Maybe @totto82  or @bska can press green if there is no more comments.

totto82:
Thanks @GitPaean for your help on this!

PR #3968, opened 2022-07-01
Adapt drift compensation code to the way Flow checks convergence. 

totto82:
Depends on https://github.com/OPM/opm-models/pull/699

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1840,7 +1840,13 @@ public:
 
         // if requested, compensate systematic mass loss for cells which were "well
         // behaved" in the last time step
-        if (enableDriftCompensation_) {
+        // Note that we don't allow for drift compensation if there is
+        // no source terms i.e. no wells and no aquifers.
+        const auto& schedule = this->simulator().vanguard().schedule();
+        int episodeIdx = this->simulator().episodeIndex();
+        const auto& aquifer = this->simulator().vanguard().eclState().aquifer();
+        bool compensateDrift =  schedule.numWells(episodeIdx) > 0 && aquifer.active();
+        if (enableDriftCompensation_ && compensateDrift) {
             const auto& simulator = this->simulator();
             const auto& model = this->model();
 
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -884,6 +884,7 @@ public:
             readEclRestartSolution_();
         else
             readInitialCondition_();
+
         tracerModel_.prepareTracerBatches();
 
         updatePffDofData_();
@@ -896,6 +897,9 @@ public:
         }
 
         readBoundaryConditions_();
+
+        // compute and set eq weights based on initial b values
+        computeAndSetEqWeights_();
 
         if (enableDriftCompensation_) {
             drift_.resize(this->model().numGridDof());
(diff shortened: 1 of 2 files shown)


totto82:
jenkins build this opm-models=699 please

bska:
> jenkins build this opm-models=699

I think I've said before that we really should trigger multi-repo build checks from the most upstream repository&mdash;in this case opm-models.  Otherwise, we don't run all unit tests in the upstream repositories.

akva2:
you have to rebase.

totto82:
> you have to rebase. i removed that function in my refactor as it was unused, you'll have to reintroduce it.

I found a better work-around. Just asked the eclState directly.

ebos/eclproblem.hh
```diff
@@ -2040,29 +2044,31 @@ public:
...
+        bool compensateDrift =  schedule.numWells(episodeIdx) > 0 && aquifer.active();
+        if (enableDriftCompensation_ && compensateDrift) {
             const auto& simulator = this->simulator();
             const auto& model = this->model();
 
             // we need a higher maxCompensation than the Newton tolerance because the
             // current time step might be shorter than the last one
-            Scalar maxCompensation = 10.0*model.newtonMethod().tolerance();
-
+            Scalar maxCompensation = model.newtonMethod().tolerance()/10;
```
blattms, line 2059: Maybe we should change the comment accordingly?

ebos/eclproblem.hh
```diff
@@ -3147,6 +3153,33 @@ private:
...
+            for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
+                if (!FluidSystem::phaseIsActive(phaseIdx))
+                    continue;
+
+                sumInvB[phaseIdx] += dofFluidState.invB(phaseIdx);
+            }
+        }
+        const auto& comm = this->simulator().vanguard().grid().comm();
+        comm.sum(sumInvB.data(),sumInvB.size());
+        Scalar numTotalDof = comm.sum(numDof);
```
blattms, line 3170: I assume that these are all dofs regardless whether they are at ghost cells or not. Then both sumInvB and numTotalDof will be too high in parallel as dofs that are in the ghost area on a process will be counted twice.

The easiest fix would be to iterate over the interior of the grid. Not sure whether there is a way to get the corresponding dof indices then.

blattms (commented):
Apart from the paralllel issues this looks good. They might not have such a big effect as long as the interior is big, but I think these need to fixed eventually. If we do it now we at least won't forget about it.

ebos/eclproblem.hh
totto82: This should be fixed now

ebos/eclproblem.hh
totto82: done

totto82:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -2943,6 +2950,44 @@ private:
...
+        std::vector<Scalar> sumInvB(numPhases, 0.0);
+        const auto& gridView = this->gridView();
+        ElementContext elemCtx(this->simulator());
+        auto elemIt = gridView.template begin</*codim=*/0>();
+        const auto& elemEndIt = gridView.template end</*codim=*/0>();
+        for (; elemIt != elemEndIt; ++elemIt) {
+            const auto& elem = *elemIt;
+            if (elem.partitionType() != Dune::InteriorEntity)
+                continue;
+
```
blattms, line 2963: BTW There is a shortcut for this in modern DUNE that simplifies code:
```
for(const auto& elem: elements(gridView, Dune::Partitions::interior))

blattms (approved):
Up to you whether to use the simpler modern way.

ebos/eclproblem.hh
totto82: Thanks. Changed. Who doesn't want to be modern!

blattms:
Thanks a lot. Will merge both of them.

blattms:
I guess you need to do an update_data first because of the test failing expectedly, first.

PR #3974, opened 2022-07-08
Wellconnections

ElyesAhmed:


bska (approved):
This entire function is very expensive if it's called more than once per simulation and I'm not really a fan of using `unordered_map<>::operator[]` since this will insert values into the map, but since this is only for CPR I suppose we can postpone fixing those issues until later.

I'll run a build check and merge if that's successful.

bska:
jenkins build this please

atgeirr:
There is no description here. What is the reason for the rewrite?

bska:
> I'll run a build check and merge if that's successful.

The build check was successful so I'll merge this now.

ElyesAhmed:
> > I'll run a build check and merge if that's successful.
> 
> The build check was successful so I'll merge this now.

Thanks!

PR #3975, opened 2022-07-11
Unsupported keywords

OPMUSER:
Remove MINPORV, keyword from unsupported keyword list, opm-common PR to follow to handle keyword. Remove RTEMPA and TEMPVD, as these are already supported.


bska:
jenkins build this please

bska (approved):
Looks good to me.  I'll merge this now albeit somewhat prematurely since OPM/opm-common#3071 is still not merged.  The upstream PR is nevertheless incoming shortly so the inconsistency does not matter too much.

PR #3976, opened 2022-07-13
[bugfix] Fix well transmissibility weights for distributed wells.

blattms:
If we compute a sum over all perforations than we need to cater for the case of distributed wells. That is we need to also sum up  over all processes involved after performing the local sum.

One of these global sums was missing in computeWellConnectionDensitesPressures for producers when we compute the weights based on well transmissibilities.


blattms:
jenkins build this please

blattms:
jenkins run norne serial please

blattms:
Shameless self-merge :grimacing:

PR #3980, opened 2022-07-27
add missing include

akva2:
Disclosed by removing include from header in https://github.com/OPM/opm-common/pull/3085


akva2:
jenkins build this please

PR #3989, opened 2022-08-09
Tweak opencl launch params v2

Tongdongq:
Same as https://github.com/OPM/opm-simulators/pull/3953, but without the compile error.
Also reduced the length of some more lines.

### DIFF
--- a/opm/simulators/linalg/bda/opencl/openclKernels.hpp
+++ b/opm/simulators/linalg/bda/opencl/openclKernels.hpp
@@ -64,6 +64,7 @@ private:
     static cl::CommandQueue *queue;
     static std::vector<double> tmp;     // used as tmp CPU buffer for dot() and norm()
     static bool initialized;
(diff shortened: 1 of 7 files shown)


bska:
jenkins build this please

opm/simulators/linalg/bda/opencl/openclKernels.hpp
```diff
@@ -64,6 +64,7 @@ class OpenclKernels
     static cl::CommandQueue *queue;
     static std::vector<double> tmp;     // used as tmp CPU buffer for dot() and norm()
     static bool initialized;
+    static size_t preferred_workgroup_size_multiple; // stores CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE
```
bska, line 67: You should include `<stddef.h>` if you have data members of type `size_t`.  Alternatively, you could/should make the data member into a `std::size_t` and include `<cstddef>` instead.

Tongdongq: 18d9f8c, uses `<cstddef>`.

bska:
jenkins build this please

bska (approved):

PR #3990, opened 2022-08-10
Fix Averaged Quantities for Non-HC Simulations

### LINKED ISSUE #3984: Curious error in the pressure solution may be a clue to finding a bug
[2022.04.0] The attached test model came about as I was looking into a subtle but consistent difference between the pressure solution in Flow compared to Eclipse. The attached model illustrates this and may help you to identify if there is an underlying bug of some sort. 

Model files
[Test_Water.zip](https://github.com/OPM/opm-simulators/files/9213252/Test_Water.zip)

Brief summary: The test model is a 1 x 1 x 1 single phase water cell and Flow seems to have some sensitivity to the formation compressibility (ROCK). The model is 1000 m x 1000 m area x 10 m thick, 1.0 ntg, 0.1 porosity, Bw 1.0, so pore volume is exactly 1 mln rm3. Initial pressure is 300 bara at datum 3000 m (i.e. initial FPR 300,5 bara). The schedule has 1000 rm3/d injection x 100 days with a maximum of 500 bara.  

Case 1 (Water1F, Water1E) has total compressibility 200E-6/bar, where Flow shows a slightly higher gradient of pressure increase and also plateaus at around 513 bara which is significantly higher than the maximum injection pressure 500 bara,  

image

Case 2 (Water2F, Water2E) has an artificially high total compressibility 5000E-6/bar. This is obviously on the high end but done as one way to to amplify this discrepancy. This may help can guide you to the root cause. Flow now shows a much higher gradient of pressure increase, rising to 350 bara compared to the correct final pressure of 320 bara (i.e. initial pressure 300 bara + injection 0.1 mln rm3 / pore volume 1 mln rm3 / total compressibility 5000E-6 = 320 bar). 

image

bska:
Commit 6d3da3d2e (PR #3397) introduced the notion of a "dynamic", pressure dependent pore-volume and switched the `PoreVolume` aggregates to reference condition evaluation for output to the `PORV` field in the .PRT file.  We did however fail to update all existing uses of `PoreVolume` which introduced an inconsistency.  In particular, for simulation models without hydrocarbons&mdash;e.g., a single-phase water run&mdash;the numerator of the volume weighted average would include
pressure effects through the rock compressibility but the denominator would not.

Thanks to Edmund Stephens for discovering the issue and providing an example model for analysing the underlying issue.

Resolves #3984

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -1191,6 +1191,13 @@ add_test_compareECLFiles(CASENAME bo_diffusion
                          REL_TOL ${rel_tol}
                          DIR diffusion )
 
+add_test_compareECLFiles(CASENAME fpr_nonhc
+                         FILENAME WATER2F
+                         SIMULATOR flow
+                         ABS_TOL ${abs_tol}
+                         REL_TOL ${rel_tol}
+                         DIR water-1ph)
+
 # Restart tests
 opm_set_test_driver(${PROJECT_SOURCE_DIR}/tests/run-restart-regressionTest.sh "")
 # Cruder tolerances for the restarted tests
--- a/ebos/eclgenericoutputblackoilmodule.cc
+++ b/ebos/eclgenericoutputblackoilmodule.cc
@@ -1544,7 +1544,7 @@ outputFipLogImpl(const Inplace& inplace) const
         Scalar fieldHydroCarbonPoreVolumeAveragedPressure = pressureAverage_(inplace.get(Inplace::Phase::PressureHydroCarbonPV),
                                                                              inplace.get(Inplace::Phase::HydroCarbonPV),
                                                                              inplace.get(Inplace::Phase::PressurePV),
-                                                                             inplace.get(Inplace::Phase::PoreVolume),
+                                                                             inplace.get(Inplace::Phase::DynamicPoreVolume),
                                                                              true);
 
         std::unordered_map<Inplace::Phase, Scalar> initial_values;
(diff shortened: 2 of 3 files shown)


bska:
I am keeping this in a draft state for now because I also want to add a regression test based on the model in OPM/opm-tests#784.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
I'm marking this as "ready for review" now.  The two [failing regression tests](https://ci.opm-project.org/job/opm-simulators-PR-builder/3831/testReport/) are **expected** because

1. The [`WATER2F`](https://ci.opm-project.org/job/opm-simulators-PR-builder/3831/testReport/(root)/mpi/compareECLFiles_flow_WATER2F/) test is new in this PR and therefore does not have an existing reference solution.  We must run `update_data` to create the first such solution.
2. The region pressures in regions 8, 12, and 16 of [NORNE_ATW2013_B1H_RE_PERF](https://ci.opm-project.org/job/opm-simulators-PR-builder/3831/testReport/(root)/mpi/compareECLFiles_flow_NORNE_ATW2013_B1H_RE_PERF/) actually match the situation corrected in this PR.  Those regions are completely water filled&mdash;i.e. $S_w=1$&mdash;at all times and therefore do not have hydrocarbons.  We must update the existing reference solution for this simulation model.

In summary, running `update_data` is the expected and safe way to make the tests go through.

akva2 (approved):
seems fine to me, minimal change. i can't really speak to correctness but i trust you on that.

akva2:

akva2:
jenkins build this opm-tests=786 please

PR #3994, opened 2022-08-11
BlackOilAquiferModel: use implementations through interface

akva2:
Refactor to have a common base for analytical and numerical aquifers, and then use them through this interface in the BlackOilAquiferModel.


akva2:
jenkins build this please

GitPaean:
It is a welcomed change. I assume there was no error introduced in the code moving process and I do not have further comment regarding the PR.  The jenkins test is green. I am merging this PR to the master.

PR #3995, opened 2022-08-11
remove include of Well header

akva2:
Downstream of https://github.com/OPM/opm-common/pull/3096


bska (approved):

PR #3996, opened 2022-08-12
Move action handling from eclproblem to separate class

akva2:
Clearer interface boundaries and a minor reduction in code built in all simulator objects.

Plus some drive-by cleanup / cosmetics.

### DIFF
--- a/ebos/equil/initstateequil.cc
+++ b/ebos/equil/initstateequil.cc
@@ -1793,7 +1793,7 @@ using GridView = Dune::Fem::GridPart2GridViewImpl<
                                      Dune::Fem::AdaptiveLeafGridPart<
                                          Dune::CpGrid,
                                          Dune::PartitionIteratorType(4),
-                                         false>;
+                                         false>>;
 #else
 using GridView = Dune::GridView<Dune::DefaultLeafGridViewTraits<Dune::CpGrid>>;
 #endif


akva2:
jenkins build this please

ebos/eclactionhandler.hh
```diff
@@ -0,0 +1,92 @@
...
+                   Action::State& actionState,
+                   SummaryState& summaryState,
+                   BlackoilWellModelGeneric& wellModel,
+                   Parallel::Communication comm);
+
+  void applyActions(int reportStep,
+                    double sim_time,
+                    const TransFunc& updateTrans);
+
+protected:
```
bska, line 65: Why are these marked as `protected` instead of `private`?  Do you envision inheriting from this class?

ebos/eclactionhandler.cc
```diff
@@ -0,0 +1,186 @@
...
+        if (actionResult) {
+            std::string wells_string;
+            const auto& matching_wells = actionResult.wells();
+            if (!matching_wells.empty()) {
+                for (std::size_t iw = 0; iw < matching_wells.size() - 1; iw++)
+                    wells_string += matching_wells[iw] + ", ";
+                wells_string += matching_wells.back();
+            }
+            std::string msg = "The action: " + action->name() + " evaluated to true at " + ts + " wells: " + wells_string;
+            OpmLog::info(msg);
```
bska, line 98: This block can be simplified quite considerably if we switch to using libfmt instead:
```C++
const auto msg =
    fmt::format("The action {} evaluated to true at {}. Well{}: {}",
                action->name(), ts,
                matching_wells.size() != 1 ? "s" : "",
                fmt::join(matching_wells, ", "));
OpmLog::info(msg)
```

bska (commented):
Simple enough, but I noticed a couple of potential improvements we could make while here.

ebos/eclactionhandler.hh
akva2: i guess not. i have a habit of protected unless i really don't want subclasses using stuff but i can make them private, sure.

ebos/eclactionhandler.cc
akva2: i guess i can do it. we'd get wells for an empty matching wells, not sure if the original if carried any meaning though.

akva2:
jenkins build this please

ebos/eclactionhandler.cc
bska: > we'd get wells for an empty matching wells

I'm not sure I completely understand what you're saying here.  The original version would print
```
" wells: "
```
while the libfmt-based version prints
```
". Wells: "
````
in that case (empty `matching_wells` array).  That's a minor difference in my opinion.

Comment (akva2) [reply ebos/eclactionhandler.cc]: yeah, sorry I was rather unclear. my point was that IF we can be here with matching_wells.empty(), maybe we should consider not printing the last part. the original code was just as broken in this aspect.

Comment (akva2) [reply ebos/eclactionhandler.cc]: ie something like
```
    std::string wmsg;
    if (!matching_wells.empty())
            wmsg = fmt::format("Well{}: {}",
                                             matching_wells.size() != 1 ? "s" : "",
                                             fmt::join(matching_wells, ", "));
            const auto msg =
                    fmt::format("The action {} evaluated to true at {}. {}",
                                action->name(), ts, wmsg);
```

Comment (bska) [reply ebos/eclactionhandler.cc]: > ie something like

Good!  I like it.  Let's do that.  The next step might be to use libfmt to format the `ts` too as doing so will (easily) allow us to output date strings of the form
```
15-Aug-2022
```
instead of the current format
```
2022/08/15
```
I can submit a follow-up PR to that effect unless you want to do it.

akva2: as this smells like a custom formatter i'm not entirely sure where is at, please do a follow-up.

akva2:
arf. i pushed the wrong branch in here and now github is confused. i have to open a new pr.

PR #4002, opened 2022-08-15
add Test Case for Salt Precipitation

goncalvesmachadoc:
Depends on https://github.com/OPM/opm-tests/pull/789


bska:
jenkins build this please

bska:

tests/update_reference_data.sh
```diff
@@ -71,6 +71,7 @@ tests[spe1_thermal_watvisc]="flow spe1 SPE1CASE2_THERMAL_WATVISC"
 tests[spe1_rockcomp]="flow spe1 SPE1CASE2_ROCK2DTR"
 tests[spe1_brine]="flow spe1_brine SPE1CASE1_BRINE"
 tests[spe1_precsalt]="flow spe1_precsalt SPE1CASE1_PRECSALT"
+tests[gas_precsalt]="flow spe1_precsalt GASWATER_VAPWAT_PRECSALT"
```
bska, line 74: @akva2: Why did this not create/update reference solutions (cf. OPM/opm-tests#792)?  As far as I can tell the directory, file, and case names are all correct.

akva2: because FILENAME containts .DATA

bska:

bska:
jenkins build this opm-tests=793 please

bska:
Hi @goncalvesmachadoc,

It does not appear that the Jenkins system is able to calculate a deterministic solution for this model.  We also see unexpected symptoms like negative water saturation values ($S_w < -210$ in one instance) so unless there's something I've missed this case may not be ready for inclusion into the regression test suite yet.

goncalvesmachadoc:
The case seems to work properly with the release 2022.04. Don't see negative water saturations:
image
 
Will investigate what has been broken in the master.

bska:

goncalvesmachadoc:
> That said, the `FSIP` (field-level salt-in-place) summary vector looks a little suspicious (below). Is this expected? 

The current implementation of FSIP only considers the SALT in the brine, so it would be expected to decrease when salt precipitates and/or water disappears.  

Our idea is in the near future adding a second keyword to output the amount of solid salt in the field.

bska:
> > That said, the `FSIP` (field-level salt-in-place) summary vector looks a little suspicious (below). Is this expected?
> 
> The current implementation of `FSIP` only considers the SALT in the brine, so it would be expected to decrease when salt precipitates and/or water disappears.

So, just for clarity, what you're saying is that the current `FSIP` curve (large :arrow_right: zero :arrow_right: large :arrow_right: zero :arrow_right: large) is the expected pattern?

> Our idea is in the near future adding a second keyword to output the amount of solid salt in the field.

Very good.  I look forward to seeing what that looks like.

goncalvesmachadoc:
> So, just for clarity, what you're saying is that the current `FSIP` curve (large ➡️ zero ➡️ large ➡️ zero ➡️ large) is the expected pattern?

That's weird indeed. I am getting something similar with release, but different:
image


But the PRESSURE, SALT, SALTP, SWAT and SGAS in the grid seem ok.  

I went ahead and run the https://github.com/OPM/opm-tests/blob/master/spe1_brine/SPE1CASE1_BRINE.DATA  with 
image

Found similar issue with FSIP:

Grid has salt from time 0:  
image
 
But FSIP:
image


Which makes us believe the issue is not related to salt precipitation. 

I have deleted FSIP from the test case https://github.com/OPM/opm-tests/blob/master/spe1_precsalt/GASWATER_VAPWAT_PRECSALT.DATA 


Could we go ahead and safe-guard the gas-water salt precipitation module as the simulation results looks as expected? Meanwhile we investigate what is going wrong with the summary output FSIP.

bska:
> Could we go ahead and safe-guard the gas-water salt precipitation module as the simulation results looks as expected? Meanwhile we investigate what is going wrong with the summary output FSIP.

Yep, let's do that.  I'll get OPM/opm-tests/797 into the master and then merge this.

bska:

bska:
jenkins build this opm-tests=798 please

bska (approved):
Thanks for your patience.  I'll merge this PR now.

PR #4003, opened 2022-08-15
Format Action Dates With Libfmt

bska:
Use libfmt's ability for format `std::tm` objects to output unambiguous dates.  While here, also apply 'const' to a number of objects and scalar parameters.

### Example output master (using [actionx/ACTIONX_WPIMULT](https://github.com/OPM/opm-tests/blob/62ceb6a16488067d0b860b21957cfeffeb7ff663/actionx/ACTIONX_WPIMULT.DATA)) ###
```
The action ACT-01 evaluated to true at 2018/11/02  report:0.
```

### Example output this PR ###
```
Action ACT-01 triggered at 02-Nov-2018 00:00:00 (report interval 0 to 1).
```

### DIFF
--- a/ebos/eclactionhandler.cc
+++ b/ebos/eclactionhandler.cc
@@ -26,20 +26,84 @@
 #include <ebos/eclactionhandler.hh>
 
 #include <opm/common/OpmLog/OpmLog.hpp>
+#include <opm/common/utility/TimeService.hpp>
 
 #include <opm/input/eclipse/EclipseState/EclipseState.hpp>
-#include <opm/input/eclipse/Schedule/Schedule.hpp>
 #include <opm/input/eclipse/Schedule/Action/ActionContext.hpp>
 #include <opm/input/eclipse/Schedule/Action/ActionX.hpp>
 #include <opm/input/eclipse/Schedule/Action/State.hpp>
+#include <opm/input/eclipse/Schedule/Schedule.hpp>
 
-#include <opm/simulators/utils/ParallelSerialization.hpp>
 #include <opm/simulators/wells/BlackoilWellModelGeneric.hpp>
+#include <opm/simulators/utils/ParallelSerialization.hpp>
 
-#include <iomanip>
-#include <sstream>
+#include <chrono>
+#include <cstddef>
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

atgeirr:
I think for *time points that coincide with report step time points* it is unambigous what we mean. I.e. "report time point" 0 is the start time, 1 is after the first report step etc. The problem would be when an action is evaluated in the middle, i.e. after a time step that does not end at a report step boundary. I suggest that with report time points t_i, i=0..N, and t_0 being the start time, we should report the index k if t is in [t_k, t_{k+1}). That is because the schedule has not yet seen anything from the deck's t_{k+1} data.

I wondered about the "report" though, what it really meant. Could we write "last report time point: k" instead of "report: k"?

bska:
> I suggest that with report time points t_i, i=0..N, and t_0 being the start time, we should report the index k if t is in [t_k, t_{k+1}). That is because the schedule has not yet seen anything from the deck's t_{k+1} data.

Good point.  Maybe, as a compromise, we can emit something like the following to identify the specific interval?

```
Action ACT-01 triggered at 02-Nov-2018 00:00:00 (report interval 0 to 1).
```

I also think we should stop outputting messages about the actions which do **not** trigger.  I suppose we could count the number of non-triggering actions and output the count, but even that might be too much.  The message limiters would be useful here though.

OPMUSER:
@bska see my comments on Slack https://app.slack.com/client/T02PFQ8JA/C02G325BB6V/thread/C02G325BB6V-1660621215.046899

blattms:
Well, the messages are mainly for engineers, just also useful for developers. Let the developers
do the thought work then and use reportstep+1.

Just my to cents

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

akva2 (approved):

bska:
jenkins build this please

bska:
PR approved and build check is green.  Merging this now.

PR #4008, opened 2022-08-17
Fix summing well rates for groups

totto82:
Don't include contribution from injecting producers and producing injectors when summing over groups


totto82:
jenkins build this please

PR #4010, opened 2022-08-18
changed: put ParallelIstlInformation in separate compile unit

akva2:
In itself this has not much value as such but it is part of a bigger unraveling.

### DIFF
--- a/opm/simulators/flow/BlackoilModelEbos.hpp
+++ b/opm/simulators/flow/BlackoilModelEbos.hpp
@@ -39,7 +39,6 @@
 
 #include <opm/grid/UnstructuredGrid.h>
 #include <opm/simulators/timestepping/SimulatorReport.hpp>
-#include <opm/simulators/linalg/ParallelIstlInformation.hpp>
 #include <opm/core/props/phaseUsageFromDeck.hpp>
 #include <opm/common/ErrorMacros.hpp>
 #include <opm/common/Exceptions.hpp>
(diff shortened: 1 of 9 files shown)


akva2:
jenkins build this please

opm/simulators/flow/BlackoilModelEbos.hpp
```diff
@@ -39,7 +39,6 @@
 
 #include <opm/grid/UnstructuredGrid.h>
 #include <opm/simulators/timestepping/SimulatorReport.hpp>
-#include <opm/simulators/linalg/ParallelIstlInformation.hpp>
```
blattms, line 42: Why? Are there other headers we do not install?
Maybe someone external wants to use the functionality. How would she proceed?

akva2: Why the include is removed? Because it is not used.

Why the separate compile unit?The goal is to pull less istl into compile units. If you think there may be external users, I can use an impl file in addition to the cpp file.

blattms: :flushed: Somehow thought this was a change to CMake_files.txt. No idea why was under that impression. Sorry for the noise.

blattms:
Thanks a lot.

bska:
This broke the sequential build.  The terminating namespace brace is [outside the `#if...#endif` region](https://github.com/OPM/opm-simulators/blob/f3c82fb1238a5d9934c55bc2a5e2ec6920c2280f/opm/simulators/linalg/ParallelIstlInformation.cpp#L349-L355) which depends on MPI availability, but the opening `namespace Opm {` statement is **inside** that region.

PR #4011, opened 2022-08-18
changed: split PreconditionerFactory in _impl file

akva2:
and put instances in separate cpp files.

Another step in unraveling linear solvers from simulators.


akva2:
jenkins build this please

blattms:
Personally I would prefer one PreconditionerFactory.cpp file. Am I missing any advantage of the proposed approach?

akva2:
two reasons;
1) less memory usage. if i build them all in one unit it uses 5GB (gcc9), splitting it limits this to about 1.5GB per compile unit. maybe 4GB machines aren't all that relevant these days, but still..
2) parallel build. if i build these in one unit it uses about 7 minutes on my box. that is the equivalent of almost 4 simulator objects. on the other hand i can build 6 instances in parallel in about 2 mins. i'd not be happy if i was working on the factory and had to wait 7 mins for it to build.

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

blattms:
Thanks for the explanation. Reducing hardware requirements when building is much appreciated.

bska:
The post-commit build check [failed](https://ci.opm-project.org/job/opm-simulators/1315/console).  Did this break non-MPI builds?  If so, please fix right away or revert this change.

blattms:
I never get notifications about this (but used to...) Who does get these?

blattms:
I will take a look.

blattms:
Yep this surely breaks sequential builds as it silently assumes that all parallel stuff is available for the macros. It seems to tough to fix for me. I will revert.

akva2:
Parallelized builds plus mem peaks at 4gb if all is built in one unit, more than any other unit in that case
________________________________

Issue #4012, opened 2022-08-18
Exception when adding wellcontributions to matrix


Tongdongq:
When running NORNE with `--matrix-add-well-contributions=true`, I get this error:
```
Program threw an exception: vector::_M_range_check: __n (which is 1574) >= this->size() (which is 0)
```

Probably `cartesian_to_compressed_` is not initialized, I did not find any other OPM files using it, other than the `BlackoilWellModel` files.
This leads to `BlackoilWellModel::addNeighbors()` throwing an exception.

Maybe `/opm-simulators/ebos/eclbasevanguard.hh:updateCartesianToCompressedMapping_()` is supposed to update it, but it writes to `cartesianToCompressed_`.

GDB snippet:
```
128	            for ( size_t c=0; c < connectionSet.size(); c++ )
(gdb) 
130	                const auto& connection = connectionSet.get(c);
(gdb) 
131	                int compressed_idx = cartesian_to_compressed_.at(connection.global_index());
(gdb) p connection.global_index()
$1 = 1574
(gdb) p cartesian_to_compressed_ 
$2 = {<std::_Vector_base<int, std::allocator<int> >> = {
    _M_impl = {<std::allocator<int>> = {<__gnu_cxx::new_allocator<int>> = {<No data fields>}, <No data fields>}, <std::_Vector_base<int, std::allocator<int> >::_Vector_impl_data> = {_M_start = 0x0, _M_finish = 0x0, 
        _M_end_of_storage = 0x0}, <No data fields>}}, <No data fields>}
```

bska:
Confirmed.  Seems to be fallout from the ALUGrid work in commit 5c2b60bcd (PR #3972).

PR #4013, opened 2022-08-18
simplifying the constructor for AquiferNumerical

GitPaean:
removing the unused global_cell and avoid the unnecessary construction of cartesianToCompressed . 

Inspired by https://github.com/OPM/opm-simulators/pull/4009 , but need to be cautious whether it is totally valid.

### DIFF
--- a/opm/simulators/aquifers/AquiferNumerical.hpp
+++ b/opm/simulators/aquifers/AquiferNumerical.hpp
@@ -61,13 +61,10 @@ public:
 
     // Constructor
     AquiferNumerical(const SingleNumericalAquifer& aquifer,
-                     const std::unordered_map<int, int>& cartesian_to_compressed,
-                     const Simulator& ebos_simulator,
-                     const int* global_cell)
+                     const Simulator& ebos_simulator)
         : AquiferInterface<TypeTag>(aquifer.id(), ebos_simulator)
         , flux_rate_      (0.0)
         , cumulative_flux_(0.0)
-        , global_cell_    (global_cell)
         , init_pressure_  (aquifer.numCells(), 0.0)
     {
         this->cell_to_aquifer_cell_idx_.resize(this->ebos_simulator_.gridView().size(/*codim=*/0), -1);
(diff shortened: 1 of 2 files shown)


GitPaean:
jenkins build this please

bska (commented):
I think this is a clear improvement, but I'm also concerned that this will conflict with the ALUGrid work.

PR #3972 changed a number of calls to
```
Vanguard::compressedIndexForInterior()
```
into accesses in an `unordered_map<int,int>` such as the one used in the original code here.  See, for instance, the change in to [`BlackoilWellModel_impl.hpp`](https://github.com/OPM/opm-simulators/pull/3972/files#diff-c0ac680ab7e373c6f15808a5ba1564b9d3bf32ed10ec73b8199ade39caa94b0aR131).  See also #4012.  It appears that ALUGrid (currently) expects to look up Cartesian-to-compressed index mapping one way but the rest of the code base expects a different approach.

We'll probably have to reconcile those two approaches rather soon if we expect to continue to carry the ALUGrid support.

atgeirr:
> We'll probably have to reconcile those two approaches rather soon if we expect to continue to carry the ALUGrid support.

I believe the best approach to this is to always use the vanguard's compressedIndex() or compressedIndexForInterior() methods, and make sure that this mapping always gets created correctly no matter which grid is used. To me that seems quite straightforward, but I may be missing something?

bska:
> > We'll probably have to reconcile those two approaches rather soon if we expect to continue to carry the ALUGrid support.
> 
> I believe the best approach to this is to always use the vanguard's `compressedIndex()` or `compressedIndexForInterior()` methods, and make sure that this mapping always gets created correctly no matter which grid is used.

I agree that that's the more attractive option

> To me that seems quite straightforward, but I may be missing something?

I don't know.  I just recall that we did not, at the time, succeed in establishing all of the requisite relations to build an appropriate mapper in the ALUGrid case.  I don't know which barrier prevented that though.

blattms:
> This will have to be fixed before the next release, then.

Please open an issue as we might forget otherwise.

atgeirr:
> Please open an issue as we might forget otherwise.

https://github.com/OPM/opm-simulators/issues/4021

PR #4019, opened 2022-08-22
put ParallelOverlappingILU0 in separate compile unit

akva2:
i chose to split in a separate _impl file because this code is so
generic that there may be downstream users who want to use on other
matrix types than what we use in opm-simulators.


akva2:

akva2:

PR #4022, opened 2022-08-23
changed: reduce some code duplication

akva2:
Reorganize the templates a little so we can don't have to carry more explicit code duplication than necessary, and remove code that is also in upstream.

Downstream of https://github.com/OPM/opm-models/pull/708


akva2:
jenkins build this please

bska (approved):

bska:
Sorry, I'm going to have to revert this since the Norne model no longer runs following this change.

PR #4027, opened 2022-08-25
Code modernization

aritorto:
Change from Functor class to functions.
Change white-spaces.

### DIFF
--- a/opm/simulators/linalg/MILU.cpp
+++ b/opm/simulators/linalg/MILU.cpp
@@ -1,5 +1,5 @@
 /*
-  Copyright 2015 Dr. Blatt - HPC-Simulation-Software & Services
+  Copyright 2015, 2022 Dr. Blatt - HPC-Simulation-Software & Services
   Copyright 2015 Statoil AS
 
   This file is part of the Open Porous Media project (OPM).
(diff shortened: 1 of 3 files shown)


blattms:
Shoot, we need to rebase this. Did not see that coming. This may take some time.

blattms:
Some more explanations about whitespace:
We accidentally did more witespace changes than expected, because the coding style (tabs versus spaces, indentation within namespace) was not consistent. Always not sure what the general rules are for OPM. But as we already change quite some otherwise untouched line, it might not hurt to use the un-agreed clang-format style her if that is consensus.

atgeirr:
> clang-format

Yes please!

atgeirr:
While here, I have a question regarding the ILU decomposition.

As part of this, we invert matrix blocks, and the need to do this in a robust, fast and predictable manner is why we introduced the MatrixBlock class (later duplicated, soon to be unduplicated) to have the block.invert() operation go to our 3x3 or 4x4 explicit inversions. Using just FieldMatrix would simplify the instantiation mess for all linear solver parts and maybe result in smaller executables as well, but then 4x4 inversion for example would use the Dune LU decomposition instead of the explicit formula.

So if we instead change the block ILU decomposition to explicitly invoke these inversions instead of using the member syntax block.invert() we could get the best of both worlds. The ghost_last_bilu0_decomposition() function we can easily change, in serial we are using Dune::ILU::blockILU0Decomposition, but maybe we should just use the ghost_last_bilu0_decomposition() in that case as well?

Dune::SeqILU and friends would of course not produce the same result in this scenario, but is that a problem? Will the AMG be impacted?

blattms:
>> clang-format
>
>Yes please!

Didn't know that you changed your last name to consensus :wink:

But I guess as this is seldomly touched, it might not hurt.
Is there a file where clang-format was used consistently, to see what it
should look like? 

Do I need to set anything in emacs to make it pickup the .clang-format in
the root directory?

atgeirr:
> Do I need to set anything in emacs to make it pickup the .clang-format in the root directory?

Not an emacs expert, but I have a clang-format.el from somewhere in .emacs.d/ and
```
(load "/Users/atgeirr/.emacs.d/clang-format.el")
```
in my .emacs file.

> Is there a file where clang-format was used consistently, to see what it should look like?

Not really sure, but it is easy to test it on any file to see what it does.

blattms:
jenkins build this please

PR #4028, opened 2022-08-25
changed: get rid of duplicate MatrixBlock headers/classes

akva2:
this has already led to some confusion. move some of the code
upstream to opm-models and remove the rest of the duplicated code.

the remainder of MatrixBlock.hpp is renamed to SmallDenseMatrixUtils.hpp

The utility function invertMatrix has changed behavior. On encountering a singular matrix in the 4x4 invert, it throws an exception. The continue-as-if-the-inverse-is-identity behavior is then applied upon catching this exception in StandardWell_impl.hpp

Downstream of https://github.com/OPM/opm-models/pull/710


atgeirr:
Sounds good. I think the new behaviour involving exceptions should be benchmarked for performance before we merge, I have a feeling it may be a frequent inner-loop-ish occurence so exceptions may not be appropriate. If so, we should consider other options such as returning error codes.

akva2:
okay, if you tell me how to find a timer with the appropriate precision ;)
this is exceptional. it happens for a handful of linear solves (107 calls) in a full norne run. as you know we only pay a cost upon catching in modern c++ implementations, and the cost is measured in nano seconds. but even if we go bonkers and say it's micros, we are talking 100 microsecs for a 1000sec simulation, a cost of 0.0000001 % of the run time).

i would also point out that the call was already in a try-catch block. the invert for 5 and 6 (which calls on to densematrix::invert already throws in the same circumstances. 1-3 does not however, there garbage would be returned without any way to detect it.

atgeirr:
> it happens for a handful of linear solves (107 calls) in a full norne run

Sounds not too unreasonable to keep it as an exception, then. Throwing cost in nanoseconds sounds incredible, do you have a reference? We will use the linuxbenchmarking tool for assessing performance.

atgeirr:
In any case, the new code is a better starting point even if it turns out to have a significant cost (against expectations). So I think this can be merged when green, no need to wait for the benchmark to finish.

PR #4035, opened 2022-08-29
fix analytical aquifers + energy

totto82:
Depends on https://github.com/OPM/opm-common/pull/3109

### DIFF
--- a/opm/simulators/aquifers/AquiferAnalytical.hpp
+++ b/opm/simulators/aquifers/AquiferAnalytical.hpp
@@ -158,6 +158,25 @@ public:
 
         rates[BlackoilIndices::conti0EqIdx + compIdx_()]
             += this->Qai_[idx] / model.dofTotalVolume(cellIdx);
+
+        if constexpr (enableEnergy) {
+            auto fs = intQuantsPtr->fluidState();
+            if (this->Ta0_.has_value() && this->Qai_[idx] > 0)
+            {
+                fs.setTemperature(this->Ta0_.value());
+                typedef typename std::decay<decltype(fs)>::type::Scalar FsScalar;
+                typename FluidSystem::template ParameterCache<FsScalar> paramCache;
+                const unsigned pvtRegionIdx = intQuantsPtr->pvtRegionIndex();
+                paramCache.setRegionIndex(pvtRegionIdx);
+                paramCache.setMaxOilSat(this->ebos_simulator_.problem().maxOilSaturation(cellIdx));
+                paramCache.updatePhase(fs, this->phaseIdx_());
+                const auto& h = FluidSystem::enthalpy(fs, paramCache, this->phaseIdx_());
+                fs.setEnthalpy(this->phaseIdx_(), h);
+            }
+            rates[BlackoilIndices::contiEnergyEqIdx]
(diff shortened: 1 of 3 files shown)


opm/simulators/aquifers/AquiferAnalytical.hpp
```diff
@@ -158,6 +158,27 @@ class AquiferAnalytical : public AquiferInterface<TypeTag>
 
         rates[BlackoilIndices::conti0EqIdx + compIdx_()]
             += this->Qai_[idx] / model.dofTotalVolume(cellIdx);
+
+        if (enableEnergy) {
```
akva2, line 162: if constexpr

totto82:
jenkins build this please

akva2:
jenkins build this opm-common=3109 please

opm/simulators/aquifers/AquiferAnalytical.hpp
```diff
@@ -158,6 +158,27 @@ class AquiferAnalytical : public AquiferInterface<TypeTag>
 
         rates[BlackoilIndices::conti0EqIdx + compIdx_()]
             += this->Qai_[idx] / model.dofTotalVolume(cellIdx);
+
+        if constexpr (enableEnergy) {
+            auto fs = intQuantsPtr->fluidState();
+            if (this->Ta0_.has_value() && this->Qai_[idx] > 0)
+            {
+                fs.setTemperature(this->Ta0_.value());
+                typedef typename std::decay<decltype(fs)>::type::Scalar FsScalar;
```
totto82, line 167: @GitPaean I would like to also set the aquifer pressure here for the enthalpy calculations (not the cell pressure as currently used in this PR) But how can I get the aquifer pressure for this IDX?

GitPaean: Not totally clear what you meant here.  For Carter-Tracy aquifer, there is one initial pressure, and also the the connection cells' pressures (they are cell pressures), for Fetkovich aquifer, there is an aquifer pressure (one), and connection cells, pressures. 

> I get the aquifer pressure for this IDX?

The aquifer pressure for an analytical pressure is one value, only there are multiple cell pressures for the connected reservoir. 

Excuse me if I misunderstand something here.

totto82: I hoped it was possible to compute the aquifer pressure for from some of the input parameters and values. I using the cell pressure for the enthalpy calculation is sufficiently accurate for now.

GitPaean: For Fetkovich aquifer, there is the aquifer pressure, maybe you can also try that one. Similar to the way you are using temperature, which assumes the pressure and temperature in the aquifer is uniform.   And also, do you need to consider the upwinding of the flow directions, since fluid can flow in/out of the aquifer?

For Carter-Tracy aquifer, there is only the initial aquifer pressure from the formulation. the initial pressure might be able to be used as the aquifer pressure (it is actually what is output for AAQP for Cater-Tracy aquifer).   Not totally clear whether it is possible to derive and update the pressure.

totto82: Then I think we should stick with the way its done here. I.e. use the adjacent cell pressure.

totto82:
@bska Can this and the corresponding pr in opm-common be merged?

bska (approved):
I think this is okay to merge.  While I personally prefer
```C++
using FsScalar = typename /* ... */
```
over
```C++
typedef typename /* ... */ FsScalar
```
I'm not going to insist we change that before merging.

If your discussion with @GitPaean is completed then I suggest we merge this.

PR #4037, opened 2022-08-29
scale energy residual to have the same relative order as the mass balance residuals

totto82:
We are using the same tolerances for the MB and CNV for energy and the various mass balance equations. The motivation of this change is to make the energy tolerance more comparable with the ones for mass.

### DIFF
--- a/opm/simulators/flow/BlackoilModelEbos.hpp
+++ b/opm/simulators/flow/BlackoilModelEbos.hpp
@@ -748,7 +748,7 @@ namespace Opm {
                 }
 
                 if constexpr (has_energy_) {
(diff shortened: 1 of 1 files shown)


atgeirr (approved):
Seems appropriate. As I understand it this makes error scale so that a temperature error of 1e-5 K is still significant, which seems quite safe.

totto82:
jenkins build this please

totto82:

totto82:
jenkins build this opm-tests=4037 please

bska:
jenkins build this opm-tests=825 please

PR #4040, opened 2022-09-01
consistently use ParallelCommunication.hpp for communication definition

akva2:


akva2:
jenkins build this please

ebos/eclmpiserializer.hh
```diff
@@ -21,7 +21,6 @@
 #ifndef ECL_MPI_SERIALIZER_HH
 #define ECL_MPI_SERIALIZER_HH
 
-#include <dune/common/version.hh>
 #include <opm/simulators/utils/ParallelRestart.hpp>
```
bska, line 24: Does this implicitly rely on `ParallelRestart.hpp` also bringing in `ParallelCommunication.hpp`?  If so, I think the requirement  should be made explicit by including `ParallelCommunication.hpp` directly.

opm/simulators/wells/WellState.cpp
```diff
@@ -825,8 +825,8 @@ WellState::parallelWellInfo(std::size_t well_index) const
     return ws.parallel_info;
 }
 
-template void WellState::updateGlobalIsGrup<ParallelWellInfo::Communication>(const ParallelWellInfo::Communication& comm);
-template void WellState::communicateGroupRates<ParallelWellInfo::Communication>(const ParallelWellInfo::Communication& comm);
+template void WellState::updateGlobalIsGrup<Parallel::Communication>(const Parallel::Communication& comm);
+template void WellState::communicateGroupRates<Parallel::Communication>(const Parallel::Communication& comm);
```
bska, line 829: This appears to rely on `ParallelCommunication.hpp` being brought in through transitive includes.  I'd recommend adding the header directly.

tests/test_parallelwellinfo.cpp
```diff
@@ -118,7 +118,7 @@ BOOST_AUTO_TEST_CASE(ParallelWellComparison)
     std::vector<Opm::ParallelWellInfo> well_info;
     
     for (const auto& wellinfo : pairs) {                   
-        well_info.emplace_back(wellinfo, Opm::ParallelWellInfo::Communication());         
+        well_info.emplace_back(wellinfo, Opm::Parallel::Communication());
```
bska, line 121: Should probably include `ParallelCommunication.hpp` directly then.

bska (commented):
This looks good other than a couple of places that appear to rely on `ParallelCommunication.hpp` being brought in through transitive includes.  Reducing the number of ways of defining our communicators can only be a good thing.

tests/test_parallelwellinfo.cpp
akva2: done

opm/simulators/wells/WellState.cpp
akva2: done

ebos/eclmpiserializer.hh
akva2: done

akva2:
jenkins build this please

bska (approved):

PR #4041, opened 2022-09-01
added: generalize variant serialization handling

akva2:
Replace the two hardcoded variant handlers with one generic implementation.


akva2:
jenkins build this please

PR #4045, opened 2022-09-01
Make one more template private in WellGroupHelper

akva2:
And clean up instantations.


bska (approved):
Looks good to me.  We can make the diagnostic messages more informative in a second pass.  I'll merge into master.

Issue #4042, opened 2022-09-01
Reduce User-Level Messages Concerning Action Handling


bska:
As of this writing, and PR #4003, we output a number of messages concerning actions which do not trigger to the console and the `.PRT` file. I would like to propose that we direct those messages exclusively to the `.DBG` file instead and to reserve the console and `.PRT` file to report actions which do trigger at a particular time. In running important real field cases we've experienced the "non-triggered" action messages exceeding 100,000 lines which greatly reduces the utility of the `.PRT` file. I will state unequivocally that the messages concerning non-triggered actions are very useful for developers in analysing underlying issues, but that's why I think they belong in the `.DBG` file instead.

We also report a block of keywords which get rerun as a result of an action triggering and I would like to direct this message block to the `.DBG` file too.

atgeirr:
I agree.

PR #4047, opened 2022-09-01
Direct All Non-Triggered Action Messages to Debug File

bska:
End users typically do not care that a particular action did NOT trigger at a particular time and these messages will typically reduce the utility of the `CASE.PRT` file.  On important field cases we've seen more than 100,000 such messages in a single run which is quite overwhelming.

Since we're now directing the messages to the `.DBG` file instead we're not taking away any diagnostic ability, merely altering the location of the information.

### DIFF
--- a/ebos/eclactionhandler.cc
+++ b/ebos/eclactionhandler.cc
@@ -101,7 +101,7 @@ namespace {
                         (numInactive != 1) ? "s" : "",
                         timeString);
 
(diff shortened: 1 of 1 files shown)


bska:
This PR partly addresses #4042 but needs a companion PR in the upstream OPM-Common repository to fully resolve that issue.

bska:
jenkins build this please

bska:
Merging per prior approval.

PR #4048, opened 2022-09-01
Add support for directional relative permeabilities

hakonhagland:


hakonhagland:
Depends on https://github.com/OPM/opm-models/pull/714, https://github.com/OPM/opm-material/pull/532, and https://github.com/OPM/opm-common/pull/3113

hakonhagland:
jenkins build this opm-models=714 opm-material=532 opm-common=3113 please

totto82:
jenkins build this opm-models=714 opm-material=532 opm-common=3113 please

totto82:
Thanks for the contribution. I think the opm-common/opm-material parts looks good. For opm-models my main concern is that this adds mobility[X,Y,Z] to the blackoilintensivequantities even in cases where KRNUMX, KRNUMY, KRNUMZ is not used. Since directional relative permeability is typically not very often used a solution with conditional storage of the extra mobiles is better IMO. For the opm-simulator part the NNC issue (https://ci.opm-project.org/job/opm-simulators-PR-builder/3916/testReport/junit/(root)/mpi/compareECLFiles_flow_3D_2AQU_NUM/) needs to be resolved. Maybe just using the relperm from the original satnum region will do for now. This implementation keeps the mobility call without the faceID. Is this a temporary fix to avoid changes in the output code etc.

hakonhagland:
> a solution with conditional storage of the extra mobiles is better IMO

@totto82 Good idea, I have updated this PR (and  https://github.com/OPM/opm-material/pull/532 and https://github.com/OPM/opm-models/pull/714) with a new approach that only has an empty shared pointer if directional permeabilities are not used. 

> the NNC issue ... needs to be resolved.

Yes indeed! In this update I try to check if directional relative permeabilties are enabled before trying to access the faceid. Then, only if the user enables directional relperms and also uses a grid with NNC faces the exception will be thrown. See also the corresponding update in https://github.com/OPM/opm-models/pull/714

hakonhagland:
jenkins build this opm-models=714 opm-material=532 opm-common=3113 please

atgeirr:
I have looked at this, and note that as written, this is not compatible with the TpfaLinearizer fast assembly. If directional relperms are needed on real fields we must do this in a way that is compatible. I think we may do this by adding a face direction argument to the local residual's computeFlux() method, so it should not be too hard. The current way depends on changes to the stencil inside the element context, which is not used by the TpfaLinearizer.

I think we should not merge until we have addressed this.

totto82:
@atgeirr You mean here https://github.com/OPM/opm-models/blob/2e2961939c181d87a9a0afaafac3e45eaffdb436/opm/models/blackoil/blackoillocalresidualtpfa.hh#L194
and https://github.com/OPM/opm-models/blob/2e2961939c181d87a9a0afaafac3e45eaffdb436/opm/models/blackoil/blackoillocalresidualtpfa.hh#L242

Why didn't Jenkins complain? I though we had tests for the TpfaLinearizer assembly.

atgeirr:
> Why didn't Jenkins complain? I though we had tests for the TpfaLinearizer assembly.

We do, but only for compilation at this point. And we have no tests that check that directional relative permeability works (of course...), that should be added. If this had been already merged, and we had an integration test for the functionality, then #4017 would have had that test failing.

hakonhagland:
> this is not compatible with the TpfaLinearizer fast assembly

@atgeirr Thanks for the comment! Yes I agree, we  should add support for directional relative permeability for the flux computations in the tpfa linearizer also. I have updated this PR to address this, see also the corresponding update in https://github.com/OPM/opm-models/pull/714 and https://github.com/OPM/opm-models/pull/714.

hakonhagland:
jenkins build this opm-models=714 opm-material=532 opm-common=3113 please

hakonhagland:
jenkins build this opm-models=714 opm-material=532 opm-common=3113 please

bska:
> opm-models=714 opm-material=532 opm-common=3113

Just for your information: Due to a quirk of the CI system, you should **generally** trigger multi-repository builds from the most upstream repository&mdash;in this case OPM-Common.  The reason is that unit tests in upstream repositories won't run otherwise.

hakonhagland:
>trigger multi-repository builds from the most upstream repository

@bska Thanks for the information. I have triggered a new jenkins build in https://github.com/OPM/opm-common/pull/3113 instead.

PR #4049, opened 2022-09-02
eclmpiserializer: improve pair handling

akva2:
use has_serializeOp trait

There is going to be a series of these PR's coming. I want to go slow to make sure I don't introduce bugs.


akva2:
jenkins build this please

bska (approved):
This looks good to me.  I guess we'll have a bit of an issue if one or both of `T1` and `T2` are `const` something (e.g., `std::pair<const std::string, std::size_t>`), but that's a problem I'm willing to postpone until it actually occurs.

I'll merge into master.

PR #4051, opened 2022-09-02
eclmpiserializer: improve array handler

akva2:
use has_serializeOp trait


bska (approved):

PR #4054, opened 2022-09-02
Allow for gaslift for wells that violate thp limit but are not allowed to switch

totto82:


totto82:
jenkins build this please

totto82:
@OPMUSER Any comments on this.

totto82:
jenkins build this please

OPMUSER:
@totto82 Very good question, unfortunately I don't know the answer at the top of my head.

What case and well are you trying to fix and I'll take a look and see what E100 does. If we don't have a test case send me the case control parameters you want to test and I'll try and put a test case together on Monday.

totto82:
> What case and well are you trying to fix and I'll take a look and see what E100 does. If we don't have a test case send me the case control parameters you want to test and I'll try and put a test case together on Monday.

The fix is not triggered by any case. I still think it make sense to allow for gaslift increase for wells the violate thp but are not allowed to switch.

OPMUSER:
Maybe, but we should not allocate lift gas if the well is not flowing:

image

totto82:
>Maybe, but we should not allocate lift gas if the well is not flowing:

This is a issue with the output. See https://github.com/OPM/opm-simulators/pull/4126
Also note that the bhp is missing since we treat the well as SHUT even though the well is tolled to STOP since crossflow is not allowed.

blattms:
This seems to have broken the build, ee <https://ci.opm-project.org/job/opm-simulators/1410/display/redirect?page=changes>

blattms:
Well, actually it might have been my PR #4157 that broke the build and not this one. I seem to have failed to initiate some functions. #

PR #4055, opened 2022-09-02
eclmpiserializer: improve map handler

akva2:
Sits on top of https://github.com/OPM/opm-simulators/pull/4052


bska (approved):

PR #4056, opened 2022-09-02
changed: rename macro

akva2:
to better reflect reason for needing two macros


akva2:
jenkins build this please

bska (approved):
Yep, this makes sense.  I'll merge into master.

PR #4059, opened 2022-09-05
Add data::GuideRateValue mpiserializer test

akva2:
Have to extend the primitive serializer with std::bitset support.

Downstream of https://github.com/OPM/opm-common/pull/3121


opm/simulators/utils/ParallelRestart.cpp
```diff
@@ -409,6 +409,27 @@ struct Packing<data::Segment>
     }
 };
 
+template <size_t Size>
```
bska, line 412: Is there a reason for using `size_t` instead of `std::size_t` here?

bska (commented):
Apart from `size_t` vs. `std::size_t` (`<cstddef>`) this looks good to me.

opm/simulators/utils/ParallelRestart.cpp
akva2: no reason, just a slip. fixed now.

bska (approved):

PR #4060, opened 2022-09-05
data::GroupGuideRates: mpiserializer test

akva2:
Sits on top of https://github.com/OPM/opm-simulators/pull/4059

Downstream of https://github.com/OPM/opm-common/pull/3122


bska (approved):

PR #4079, opened 2022-09-06
added: SummaryState mpiserializer test

akva2:
Have to extend the serializer with a map predicate to handle the nested maps.
Downstream of https://github.com/OPM/opm-common/pull/3140


ebos/eclmpiserializer.hh
```diff
@@ -464,6 +466,22 @@ protected:
...
+
+    template<class T1, class T2>
+    struct is_map<std::map<T1,T2>> {
+        constexpr static bool value = true;
+    };
+
+    template<class T1, class T2>
+    struct is_map<std::unordered_map<T1,T2>> {
+        constexpr static bool value = true;
+    };
```
bska, line 483: Do we gain anything from restricting these types to their default allocators, comparison predicates, and hash functions?

akva2: not really, just wanted to keep it simple but i can extend.

bska (approved):

PR #4082, opened 2022-09-06
fixed: do not include EclMpiSerializer without mpi

akva2:
This keeps popping up and I keep forgetting it.

While the code in the primitive serializer has dummy implementations without MPI, an annoying issue in dune keeps biting us.
CollectiveCommunication&lt;<MPI_Comm&gt;> has a casting operator to MPI_Comm.
However, CollectiveCommunication&lt;Dune::No_Comm&gt; has no casting operator to Dune::No_Comm.

We rely on the casting operator so we can use the collectivecommunication as a MPI_comm where appropriate. And then it all breaks in serial..

### DIFF
--- a/opm/simulators/utils/ParallelRestart.cpp
+++ b/opm/simulators/utils/ParallelRestart.cpp
@@ -29,7 +29,9 @@
 #include <memory>
 
 #include <dune/common/parallel/mpitraits.hh>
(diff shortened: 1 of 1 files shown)


akva2:

akva2:
merging to restore serial build in master.

PR #4083, opened 2022-09-06
drop primitive serializer support for data:: types

akva2:
Downstream of https://github.com/OPM/opm-common/pull/3142


opm/simulators/utils/ParallelRestart.cpp
```diff
@@ -31,35 +31,11 @@
 #include <dune/common/parallel/mpitraits.hh>
 #if HAVE_MPI
 #include <ebos/eclmpiserializer.hh>
-#endif
```
bska, line 34: I'm probably blind, but I didn't see this `#endif` reinstated somewhere else.  Is that intentional?

akva2: no, rebase meh. fixed.

bska (approved):
Thanks.  The one downside I see to removing the separate unit tests is that it is a lot harder to single-step into a macro in a debugger if the test fails.  On the other hand I suppose we can always insert a dedicated test case locally if that situation occurs.

I'll merge into master.

atgeirr:
Having written part of the code deleted here just weeks ago, I support this removal!

blattms:
Good for you. Not always easy, though. :+1:

PR #4087, opened 2022-09-08
Rename serialization files

akva2:
This is the start of a PR series where we end up with the current EclMpiSerializer (ultimately renamed to a generic Opm::Serializer) being templated over a primitive serializer (ie the code in the MPISerializer), to allow using it with other backends than MPI_Pack w/ friends. The goals are

- to generalize the serialization API
- to simplify the serialization API
- to allow moving the serialization class and its tests to opm-common

To keep things building at all stages there will be code that is moved around several times.


akva2:
jenkins build this please

opm/simulators/utils/MPISerializer.hpp
```diff
@@ -0,0 +1,373 @@
...
+  std::vector<char> buffer(size);
+  if (comm.rank() == root) {
+      int pos = 0;
+      variadic_pack(pos, buffer, comm, std::forward<Args>(args)...);
+  }
+  comm.broadcast(buffer.data(), size, root);
+  if (comm.rank() != root) {
+      int pos = 0;
+      variadic_unpack(pos, buffer, comm, std::forward<Args>(args)...);
+  }
```
bska, line 362: Quick note here: We are generally not allowed to `forward<>` arguments more than once.  I am guilty of having done that myself in the past, but now I try hard to avoid it.  Forwarding probably works in this particular case since I expect the arguments to `broadcast()` to be l-values instead of pr- or x-values in which case the reference collapsing process on the forwarding references does nothing at all.

atgeirr:
Overall I agree with the direction proposed (have not reviewed this in detail), but I wonder if MPISerializer is a good name. What about MPIPacking instead? That would more precisely indicate what its role is.

akva2:
That's a better name indeed.  Will change. There is really nothing to review here, it's a pure splitting and renaming PR.

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
any reason why this has limboed?

bska:
> any reason why this has limboed?

You didn't respond to my [earlier comment](https://github.com/OPM/opm-simulators/pull/4087#pullrequestreview-1100526623).

opm/simulators/utils/MPISerializer.hpp
akva2: You are probably correct. However, this is not new code, it has been working for a year or so. Not sure how how I can avoid it, as we need to go over the params twice, once to find the size and once to pack them.

akva2: I can probably remove the forward, it was added upon your request, https://github.com/OPM/opm-simulators/pull/3674#discussion_r745524730 :)

bska: > I can probably remove the forward

Cool.

> it was added upon your request

Right.  For what it's worth I still stand by that comment and it applies to the implementations of the individual `variadic_*()` operations.  Just not to implementation of the `broadcast()` operation.  You **can** `forward<>()` the arguments to the **last** part of the operation (i.e., the `variadic_unpack()`) however.

akva2: done.

akva2:
jenkins build this please

bska (approved):

PR #4088, opened 2022-09-08
removing unused setupCartesianToCompressed_ function for BlackoilWellModel

GitPaean:
it shows up when reading code.


GitPaean:
jenkins build this please

bska (approved):
Good catch.  I **suspect** that this is a remnant from an early version of the ALUGrid work.  We'll reintroduce the mapping if we need it.  For now I'll merge into master.

PR #4089, opened 2022-09-08
Optionals are handled in serializer

akva2:
No need for this in the packer.

Also fix predicate for vector to allow custom allocators.

Sits on top of https://github.com/OPM/opm-simulators/pull/4087


akva2:
jenkins build this please

PR #4097, opened 2022-09-12
remove computeBhpAtThpLimitProd() output from .DBG file

totto82:
The current .DBG file is unreadable for cases with many wells due to hundreds of computeBhpAtThpLimitProd() output to the .DBG files. This PR removes them by default.

### DIFF
--- a/opm/simulators/wells/WellInterfaceGeneric.cpp
+++ b/opm/simulators/wells/WellInterfaceGeneric.cpp
@@ -459,11 +459,13 @@ bhpMax(const std::function<double(const double)>& fflo,
     double high = maxPerfPress + 1.0 * unit::barsa;
     double f_low = fflo(low);
     double f_high = fflo(high);
-    deferred_logger.debug("computeBhpAtThpLimitProd(): well = " + this->name() +
-                           "  low = " + std::to_string(low) +
-                           "  high = " + std::to_string(high) +
-                           "  f(low) = " + std::to_string(f_low) +
-                           "  f(high) = " + std::to_string(f_high));
+    if constexpr (extraBhpAtThpLimitProdOutput) {
+        deferred_logger.debug("computeBhpAtThpLimitProd(): well = " + this->name() +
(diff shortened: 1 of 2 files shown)


totto82:
jenkins build this please

bska:
While I absolutely appreciate the desire to not output those messages by default, I'd rather prefer that we not make the option a compile-time constant.  Doing so means that re-enabling the messages is only available to those who compile the simulator from source.  Having a way to request additional diagnostics that does not require compiling the simulator is very valuable.

alfbr:
How about a command-line parameter to control verbosity of the DBG file, could possibly also be added to the PRT-file and terminal output. Both the DBG- and the PRT-file are reacing a point where the share amount of output makes them less useful.

OPMUSER:
I agree, which is why I have suggested a RPTDEBUG keyword, similar to the RPTSCHED keyword. So here perhaps:

```
RPTDEBUG
     WELLS=1 
/
```
@bska  and  @totto82  What do you think, or am I flogging a dead horse again :-)

OPMUSER:
I don't like the idea of yet more command line parameters. If we implement a RPTDEBUG keyword than it will give more flexibility on both the output type and timing. We can also use:

```
RPTDEBUG
  NONE
/ 
```
Switch the output off. This way we have some control so we can get debug at specific time intervals.

The horse came back to life for a brief moment :-)

totto82:
> I'd rather prefer that we not make the option a compile-time constant.

My first suggestion was to removed the output completely, but then i hesitated and added a compile time constant. Are they useful at all at the current stage of development? Should we just removed them?

atgeirr:
I think adding the compile time option here is a good stopgap to make the DBG useful. Then if desired, we can make it a runtime option later with not too much fuzz. So I am inclined to merge this.

atgeirr:
Merging now, if we decide to add more runtime verbosity control it will not be hard to make these messages configurable as well.

PR #4098, opened 2022-09-12
remove cutoff for extrapolation

totto82:
Needs more testing

### DIFF
--- a/opm/simulators/wells/VFPHelpers.cpp
+++ b/opm/simulators/wells/VFPHelpers.cpp
@@ -134,13 +134,6 @@ InterpData findInterpData(const double value_in, const std::vector<double>& valu
         }
     }
 
-    // Disallow extrapolation with higher factor than 3.0.
-    // The factor 3.0 has been chosen because it works well
-    // with certain testcases, and may not be optimal.
-    if (retval.factor_ > 3.0) {
-        retval.factor_ = 3.0;
-    }
-
     return retval;
 }
 
--- a/tests/test_glift1.cpp
+++ b/tests/test_glift1.cpp
@@ -192,7 +192,7 @@ BOOST_AUTO_TEST_CASE(G1)
     group_info.initialize();
     auto state = glift.runOptimize(iteration_idx);
     BOOST_CHECK_CLOSE(state->oilRate(), 0.01736111111111111, 1e-8);
-    BOOST_CHECK_CLOSE(state->gasRate(), 1.6464646999768586, 1e-8);
+    BOOST_CHECK_CLOSE(state->gasRate(), 1.6464, 1e-3);
     BOOST_CHECK(state->oilIsLimited());
     BOOST_CHECK(!state->gasIsLimited());
     BOOST_CHECK(!state->alqIsLimited());


totto82:
jenkins build this please

atgeirr:
Removing this means allowing extrapolation further out than 3 times the length of the last interval in the table. That does not seem unreasonable, but I think it makes sense to keep some limit. What about limiting the `value` used to compute the factor to no more than (say) `values.back() + 3.0 * (values.back() - values.front())`, that is using the length of the whole defined interval, rather than the length of the last piece of it?

blattms:
This indeed fixed the issue reported, but we should send the user a message if the interpolation limit is reached

totto82:
Thanks for the feedback. I will do the following. Remove the cutoff but add a warning if the table is extrapolated 10% more than its complete range.

atgeirr:
> Remove the cutoff but add a warning if the table is extrapolated 10% more than its complete range.

Sounds like a good plan.

totto82:
jenkins build this please

totto82:
jenkins build this please

blattms:
This has been hanging for a quite a while. Did you check the failures and should we proceed with the update_data drill?

totto82:
From my side this is ok. I will rerun jenkins.

totto82:
jenkins build this please

totto82:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.018    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.026    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.012    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 0.977    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1801

totto82:
More testing indicates that extrapolation of vfp table is not a rare event. Even thought removing the factor 3 limit seems to do no harm,  the output becomes very noise. I will need to investigate further. This PR is thus not yet ready.

bska:
> This PR is thus not yet ready.

Okay.  Do you want to set it back to "draft" state to avoid someone accidentally merging the change?

totto82:
jenkins build this please

totto82:
> Okay. Do you want to set it back to "draft" state to avoid someone accidentally merging the change?

No, I rather want to merge this without outputting the extra debug info. I haven't find a case where removing the cutoff of 3 leads to problems. For some/many? cases the maximum flow potential in the well is far above the maximum entry in the "FLO" in the vfp-table. If we want to warn for every time this happens it will be too much. We need to find a better way to warn the user than outputting every time the vfp table is extrapolated, but that should be in another PR.

bska:
> > Okay. Do you want to set it back to "draft" state to avoid someone accidentally merging the change?
> 
> No, I rather want to merge this without outputting the extra debug info.

I only responded to you [stating that this PR was not ready for merging](https://github.com/OPM/opm-simulators/pull/4098#issuecomment-1263435600).  I did not look at any of the details.

totto82:
Approved by @atgeirr in private communication so I will start the update of data process (including the glift case)

totto82:

totto82:
jenkins build this opm-tests=817 please

PR #4101, opened 2022-09-13
Fix wells active determination 

GitPaean:
With the master branch, it uses the `wells_ecl` to determine whether wells are active, then SHUT wells might be counted. 

This is an effort to make this part cleaned up a little bit. 

`wellsActive()` or `localWellsActive()` are used many places, more testing and investigation will be required, so it is working in progress.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -198,11 +198,6 @@ namespace Opm {
             this->initializeWellPerfData();
             this->initializeWellState(timeStepIdx, summaryState);
 
-            // Wells are active if they are active wells on at least
-            // one process.
-            wells_active_ = localWellsActive() ? 1 : 0;
-            wells_active_ = grid.comm().max(wells_active_);
-
             // handling MS well related
             if (param_.use_multisegment_well_&& anyMSWellOpenLocal()) { // if we use MultisegmentWell model
                 this->wellState().initWellStateMSWell(wells_ecl_, &this->prevWellState());
@@ -263,6 +258,11 @@ namespace Opm {
 
             // create the well container
             createWellContainer(reportStepIdx);
+
+            // Wells are active if they are active wells on at least one process.
+            const Grid& grid = ebosSimulator_.vanguard().grid();
+            wells_active_ = !this->well_container_.empty();
+            wells_active_ = grid.comm().max(wells_active_);
 
             // do the initialization for all the wells
             // TODO: to see whether we can postpone of the intialization of the well containers to
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1717,7 +1717,13 @@ namespace Opm {
     int
     BlackoilWellModel<TypeTag>::numComponents() const
     {
-        if (wellsActive()  && numPhases() < 3) {
+        // The numComponents here does not reflect the actual number of the components in the system.
+        // It more or less reflects the number of mass conservation equations for the well equations.
+        // For example, in the current formulation, we do not have the polymer conservation equation
+        // in the well equations. As a result, for an oil-water-polymer system, this function will return 2.
+        // In some way, it makes this function appear to be confusing from its name, and we need
+        // to revisit/revise this function again when extending the variants of system that flow can simulate.
+        if (numPhases() < 3) {
             return numPhases();
         }
         int numComp = FluidSystem::numComponents;
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1104,10 +1104,6 @@ namespace Opm {
     BlackoilWellModel<TypeTag>::
     apply( BVector& r) const
     {
-        if ( ! localWellsActive() ) {
-            return;
-        }
-
         for (auto& well : well_container_) {
             well->apply(r);
         }
@@ -1120,11 +1116,6 @@ namespace Opm {
     BlackoilWellModel<TypeTag>::
     apply(const BVector& x, BVector& Ax) const
     {
-        // TODO: do we still need localWellsActive()?
-        if ( ! localWellsActive() ) {
-            return;
-        }
-
         for (auto& well : well_container_) {
             well->apply(x, Ax);
         }
@@ -1176,7 +1167,7 @@ namespace Opm {
     BlackoilWellModel<TypeTag>::
     applyScaleAdd(const Scalar alpha, const BVector& x, BVector& Ax) const
     {
-        if ( ! localWellsActive() ) {
+        if (this->well_container_.empty()) {
             return;
         }
 
@@ -1304,10 +1295,8 @@ namespace Opm {
         DeferredLogger local_deferredLogger;
         OPM_BEGIN_PARALLEL_TRY_CATCH();
         {
-            if (localWellsActive()) {
-                for (auto& well : well_container_) {
-                    well->recoverWellSolutionAndUpdateWellState(x, this->wellState(), local_deferredLogger);
-                }
+            for (auto& well : well_container_) {
+                well->recoverWellSolutionAndUpdateWellState(x, this->wellState(), local_deferredLogger);
             }
 
         }
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -670,13 +670,6 @@ initFromRestartFile(const RestartValue& restartValues,
     initial_step_ = false;
 }
 
-void
-BlackoilWellModelGeneric::
-setWellsActive(const bool wells_active)
-{
-    wells_active_ = wells_active;
-}
-
 std::vector<Well>
 BlackoilWellModelGeneric::
 getLocalWells(const int timeStepIdx) const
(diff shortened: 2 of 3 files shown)


opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -200,7 +200,7 @@ namespace Opm {
 
             // Wells are active if they are active wells on at least
             // one process.
-            wells_active_ = localWellsActive() ? 1 : 0;
+            wells_active_ = this->well_container_.size() > 0 ? 1 : 0;
```
bska, line 203:
```C++
wells_active_ = !this->well_container_.empty()
```

GitPaean:
benchmark please

GitPaean:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
GitPaean: Thanks. The comments can also be updated, but I will wait for the jenkins and benchmark finishes.

GitPaean:
not very surprisingly, it causes many problems because of usage. Will investigate and fix it.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1717,7 +1717,7 @@ namespace Opm {
     int
     BlackoilWellModel<TypeTag>::numComponents() const
     {
-        if (wellsActive()  && numPhases() < 3) {
+        if (numPhases() < 3) {
```
GitPaean, line 1720: It looks problematic for some scenarios, like oil water polymer situation, while the original code does not look correct either. Will investigate.

GitPaean: Will add the following comments when jenkins returns. 

```
        // The numComponents here does reflect the actual number of the components in the system.
        // It more or less reflects the number of mass conservation equations for the well equations.
        // For example, in the current formulation, we do not have the polymer conservation equation
        // in the well equations. As a result, for an oil-water-polymer system, this function will return 2.
        // In some way, it makes this function appear to be confusing from its name, and we need
        // to revisit/revise this function again when extending the variants of system that flow can simulate.
```

GitPaean:
jenkins build this please

GitPaean:
With the approach in this PR, we ignored the underline problem related to 
```c++
int
BlackoilWellModelGeneric::
numLocalWells() const
{
    return wells_ecl_.size();
}
```
which will affect much more usage. 

Maybe it will be a good task for a later PR.

GitPaean:
This PR is helping the https://github.com/OPM/opm-simulators/pull/4102 to have more sensible determination whether there is any well running to determine whether to trigger drift compensation. 

As mentioned in the comment above,  some more refactoring regarding the function `localWellsActive()` and `numLocalWells()` to make more complete cleaning up.

GitPaean:
@bska , can you give some comments and suggestions here.

blattms:
I doubt that this will work like this. The reason is that with this change there is not relation between `localWellsActive()` and `wellsActive()` and that is  a problem. Just think of a serial where previously both always returned the same and now the do not any more.

Either you change the semantic of none or both. Anything else might be problematic.

It seems like `localWellsActive()` is used to determine whether to use the well when setting up the well matrices and vectors. It still uses `wells_ecl_` and I assume thatt it is essential to represent even SHUT wells there. In that case maybe we should introduce an additional switch for the drift cpmpensation, e.g. `nonShutWells()``?

GitPaean:
>  The reason is that with this change there is not relation between localWellsActive() and wellsActive() and that is a problem.

This is a valid point. I will address this. 

> It seems like localWellsActive() is used to determine whether to use the well when setting up the well matrices and vectors. It still uses wells_ecl_ and I assume that it is essential to represent even SHUT wells there. 

I will check here whether we really want to use `wells_ecl_` information here.  It looks like some usage is  problematic/unnecessary. 

> In that case maybe we should introduce an additional switch for the drift compensation, e.g. nonShutWells()?

I am trying to avoid to introduce new this types of functions to avoid making it more complicated. Let me double check whether we can reuse `wellsActive()` and also `localWellsActive()`.  And also, `active` and `SHUT` should not go together from wording.  If I found we really could not use `wellsActive()` for the same purpose, introducing `nonShutWells()` can be an alternative to go, and I did consider this option.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
At the end, I removed the function `localWellsActive()`.  And the meaning of the function `wellsActive()`  is basically whether there is any wells not `SHUT`.   It is a narrower concept than based on the information from `wells_ecl_`, which contains wells `SHUT`.

GitPaean:
And the function `numLocalWells()` is not touched in this PR.  From the code level. the PI scaling looks like use this function with its current meaning.  It will be a little more involved to investigate the usage of `numLocalWells()`. 

And there is no `active` in the function name `numLocalWells()`, so it is okay in its current form.

GitPaean:
@blattms @bska  please help to review and suggest what you think.

bska:
> please help to review and suggest what you think.

I haven't looked into the broader ramifications of this, but the PI scaling needs to run both for open/flowing and for shut/stopped wells.  In other words, we need to be able to run that code for all wells in `wells_ecl_` be they open or shut/stopped.

GitPaean:
> In other words, we need to be able to run that code for all wells in wells_ecl_ be they open or shut/stopped.

That is right, that is why the function `numLocalWells()` is untouched. I believe this PR does not affect the PR scaling simulation.

Do we have a regression test set up for the PI scaling so we are sure about that?

bska:
> Do we have a regression test set up for the PI scaling so we are sure about that?

We do not.  The [`model2/0B_WELPI_MODEL2`](https://github.com/OPM/opm-tests/blob/a1673423d127690c797828ce61cf7c87a8cc07ca/model2/0B_WELPI_MODEL2.DATA) test model in OPM-Tests is usable as an example, but does not exercise the "welpi for shut wells" path.  We should still nevertheless add that dataset to our regression test suite.

GitPaean:
> We do not. The [model2/0B_WELPI_MODEL2](https://github.com/OPM/opm-tests/blob/a1673423d127690c797828ce61cf7c87a8cc07ca/model2/0B_WELPI_MODEL2.DATA) test model in OPM-Tests is usable as an example, but does not exercise the "welpi for shut wells" path. We should still nevertheless add that dataset to our regression test suite.

I modified the case in the following way, it does not look like the PR is causing any difference in the running. But I am not sure whether the following modification matters for the running with `WELPI`. 


```
index 92410be1..1efe232c 100644
--- a/model2/include/hist_schedule_welpi_def-ro_compdat.sch
+++ b/model2/include/hist_schedule_welpi_def-ro_compdat.sch
@@ -33,7 +33,8 @@ COMPDAT
 /
 
 WCONHIST
-  PROD1 OPEN  RESV   2818.4    0.3   304523    5       1*      1*      1* /
+--   PROD1 OPEN  RESV   2818.4    0.3   304523 5       1*      1*      1* /
+  PROD1 SHUT  RESV   2818.4    0.3   304523    5       1*      1*      1* /
 /
 
 WCONINJH
@@ -56,7 +57,7 @@ DATES   -- 1
 /
  
 WCONHIST
-  PROD1  OPEN  RESV   2920.9    0.3   316286   5       1*      1*      1* /
+  PROD1  SHUT  RESV   2920.9    0.3   316286   5       1*      1*      1* /
 /
 
 WCONINJH
@@ -115,7 +116,7 @@ DATES   -- 4
 
 
 WCONHIST
-  PROD1  OPEN  RESV   2818.4    0.3   304523   5       1*      1*      1* /
+  PROD1  SHUT  RESV   2818.4    0.3   304523   5       1*      1*      1* /
   PROD2  OPEN  RESV   2141.6    1.1   230232   5       1*      1*      1* /
 /
```

ytelses:
Benchmark result overview: 




| Test    | Configuration                                                                                 | Relative |
| ------- | --------------------------------------------------------------------------------------------- | -------- |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1                                                    | 0.99     |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8                                                    | 0.989    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1                                                    | 1.002    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8                                                    | 0.98     |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 - FOPT (Total Oil Production At End Of Run) | 1        |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 - FOPT (Total Oil Production At End Of Run) | 1        |



View result details @ https://www.ytelses.com/opm/?page=result&id=1781

blattms (commented):
The `numComponets()` I don't understand. Can solvent only be there for 3 phases?
Apart from that this looks correct.

GitPaean:
> The numComponets() I don't understand. Can solvent only be there for 3 phases?

I think it is the case for now, theoretically probably no. This function will need to be rewritten/redesign. But I do not think we should handle it in this PR.

GitPaean:
Please suggest whether this PR can go in.  

We can do a refactoring for the function `numComponents()` (including potential name and usage change). It is indeed confusing to understand its current form and probably results future maintenance challenges.

bska (approved):
I'm still not entirely convinced that I understand all the interactions here, but as there are no regression tests and the change, in isolation, looks good I'm going to merge this PR into the master branch.

PR #4102, opened 2022-09-13
we only use drift compensation when wells are active

GitPaean:
The targeting situation is that for one case, the wells are not running for some report steps with only aquifers. Drift compensation really impacts the convergence for that running. 

The alternative can be to make `aquifer.active()` based on how much flux the aquifer is doing. 

This PR sits on top of https://github.com/OPM/opm-simulators/pull/4101 .

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -1396,7 +1396,7 @@ if(MPI_FOUND)
                                        SIMULATOR flow
                                        ABS_TOL ${abs_tol_parallel}
                                        REL_TOL ${rel_tol_parallel}
-                                       TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8)
+                                       TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8 --ecl-enable-drift-compensation=false)
 
   add_test_compare_parallel_simulation(CASENAME spe1_gaswater
                                        FILENAME SPE1CASE2_GASWATER
@@ -1411,7 +1411,7 @@ if(MPI_FOUND)
                                        SIMULATOR flow
                                        ABS_TOL ${abs_tol_parallel}
                                        REL_TOL ${rel_tol_parallel}
-                                       TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8)
+                                       TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8 --ecl-enable-drift-compensation=false)
 
   # A test for distributed standard wells. We load distribute only along the z-axis
   add_test_compare_parallel_simulation(CASENAME spe9
@@ -1419,7 +1419,7 @@ if(MPI_FOUND)
                                        SIMULATOR flow_distribute_z
                                        ABS_TOL ${abs_tol_parallel}
                                        REL_TOL ${rel_tol_parallel}
-                                       TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8)
+                                       TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8 --ecl-enable-drift-compensation=false)
 
   add_test_compare_parallel_simulation(CASENAME spe9group
                                        FILENAME SPE9_CP_GROUP
@@ -1540,7 +1540,7 @@ if(MPI_FOUND)
                                        ABS_TOL ${abs_tol_parallel}
                                        REL_TOL ${rel_tol_parallel}
 				       DIR model2
-                                       TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8)
+                                       TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8 --ecl-enable-drift-compensation=false)
 
   add_test_compare_parallel_simulation(CASENAME rxft
                                        FILENAME TEST_RXFT
@@ -1548,7 +1548,7 @@ if(MPI_FOUND)
                                        ABS_TOL ${abs_tol_parallel}
                                        REL_TOL 1.0e-3
                                        DIR rxft_smry
-                                       TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8)
+                                       TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8 --ecl-enable-drift-compensation=false)
 endif()
 
 if(OPM_TESTS_ROOT)
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1883,12 +1883,8 @@ public:
 
         // if requested, compensate systematic mass loss for cells which were "well
         // behaved" in the last time step
-        // Note that we don't allow for drift compensation if there is
-        // no source terms i.e. no wells and no aquifers.
-        const auto& schedule = this->simulator().vanguard().schedule();
-        int episodeIdx = this->simulator().episodeIndex();
-        const auto& aquifer = this->simulator().vanguard().eclState().aquifer();
-        bool compensateDrift =  schedule.numWells(episodeIdx) > 0 && aquifer.active();
+        // Note that we don't allow for drift compensation if there are no active wells.
+        const bool compensateDrift = wellModel_.wellsActive();
         if (enableDriftCompensation_ && compensateDrift) {
             const auto& simulator = this->simulator();
             const auto& model = this->model();


totto82:
Thanks. This can be merged when #4101 is merged.

GitPaean:
A note here is that a better criteria will be how much injection/production flux or aquifer flux is ongoing. Because wells and aquifer can be open while not running much, then drift compensation will impacting the convergence behavior.

GitPaean:
I will trigger a benchmark test to see the performance change with this PR.

GitPaean:
benchmark please

GitPaean:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 0.985    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.009    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 0.946    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1786

GitPaean:
I think there is some messing up of `benchmark` of this PR and https://github.com/OPM/opm-simulators/pull/4116 .  Let me trigger the benchmark again to see. The previous benchmark results are seriously different from the local testing.

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.021    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.009    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 0.982    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1788

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 0.996    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.022    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.015    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 0.973    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1789

GitPaean:
jenkins build this please

GitPaean:

GitPaean:
jenkins build this opm-tests=805 please

GitPaean:

GitPaean:
jenkins build this opm-tests=806 please

GitPaean:
jenkins build this opm-tests=806 please

GitPaean:
I do not understand here. The `update_data` created the update for the opm-tests, but the jenkins will not give green to the update from `update_data`. @bska @akva2 , do you have some suggestions? Thanks.

bska:
> jenkins will not give green to the update from `update_data`

It's a bit more subtle than that, unfortunately.  The [regression failures](https://ci.opm-project.org/job/opm-simulators-PR-builder/4017/testReport/) are all for `mpi.compareParallelSim_*`-type runs.  These do not compare to a "known good" reference solution, but rather to a sequential run (effectively `mpirun -np 1`) of the same binary on the same case.  There is no way to fix those with `update_data` so you'll either have to understand what the difference is between the sequential and parallel runs&mdash;and ideally fix it too&mdash;or possibly make the comparison tolerances less strict.

totto82:
There is some issue with the parallel support. Is wellsActive communicated?

GitPaean:
Let us try again, since after the master branch changes.

GitPaean:

GitPaean:
jenkins build this opm-tests=806 please

GitPaean:
benchmark please

GitPaean:
The test failure https://ci.opm-project.org/job/opm-simulators-PR-builder/4029/testReport/junit/(root)/mpi/compareParallelSim_flow_3_A_MPI_MULTFLT_SCHED_MODEL2/ 

is the one stopping the PR goes in. 

The comparison plot is in this pdf file. 
[3_A_MPI_MULTFLT_SCHED_MODEL2_fix.pdf](https://github.com/OPM/opm-simulators/files/9656156/3_A_MPI_MULTFLT_SCHED_MODEL2_fix.pdf)

It appears to be there is a little different time stepping between `01/01/2021` and `01/02/2021`. The serial running takes two extra shorter time steps there and it is where the visual difference shown in the plot. 

To me, it is not worrying for the comparison, but it is not easy for jenkins to approve it either.

GitPaean:
because the problem happens when there is relatively bigger step (31 days), I suggest we use moderate time steps to avoid these types of not easy to fix time stepping problems. 

Let us try with 10days.

GitPaean:
It looks like the case had been previously difficult to fix the comparison. 

The following arguments have been put there for the comparison, 

```
--linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8
```

Strict tolerance will also make the time stepping hard. Maybe tightening tolerances was not the optimal way of fixing the comparison.

GitPaean:
jenkins build this opm-tests=806 please

GitPaean:
Just for comparison, when adding `--solver-max-time-step-in-days=10` to the running of the case 3_A_MPI_MULTFLT_SCHED_MODEL2 , the parallel comparison is as follows, 

[3_A_MPI_MULTFLT_SCHED_MODEL2_fix.pdf](https://github.com/OPM/opm-simulators/files/9656459/3_A_MPI_MULTFLT_SCHED_MODEL2_fix.pdf)

basically visually identical.

blattms:
As jenkins fails whenever the time stepping is different (e.g. different number of time steps), we need to make sure that the time stepping is the same for the serial and parallel run (there is no guarantee that it is as the (non-)linear solvers might behave slightly different).

Hence the question is whether the time stepping is the same now with your additional parameter. If that is the case, then we should probably add that option to compareECLFiles.cmake. Unless @akva2 objects.

GitPaean:
> Hence the question is whether the time stepping is the same now with your additional parameter. If that is the case, then we should probably add that option to compareECLFiles.cmake.

Yes, for the example `compareParallelSim_flow_3_A_MPI_MULTFLT_SCHED_MODEL2`, adding `time-step-in-days=10` does make the time stepping the same between serial running and parallel running. As a result, it fixes the comparison.

GitPaean:
And also, one observation is that, if the serial running needs to chop time steps, it is very hard to guarantee the parallel running behaves exactly the same.  

I think limiting the time step sizes are an viable approach here.

GitPaean:
jenkins build this opm-tests=806 please

GitPaean:
For the remaining failure, 
https://ci.opm-project.org/job/opm-simulators-PR-builder/4035/testReport/(root)/mpi/compareParallelSim_flow_TEST_RXFT/

the well curve comparison is  good. 
[TEST_RXFT_fix.pdf](https://github.com/OPM/opm-simulators/files/9659896/TEST_RXFT_fix.pdf)

But the following are exceeding the tolerances, 
```
	RGFT-:1-4: Summary file
		Fails for 1 entries
		Largest absolute error: 5.3650879e+01
		Largest relative error: 1.6932280e-02
	RGFT:1-2: Summary file
		Fails for 2 entries
		Largest absolute error: 6.5027344e+02
		Largest relative error: 4.7599140e-01
	RGFT:1-4: Summary file
		Fails for 1 entries
		Largest absolute error: 5.3650879e+01
		Largest relative error: 1.6932280e-02
	RGFT:3-4: Summary file
		Fails for 6 entries
		Largest absolute error: 6.4686719e+02
		Largest relative error: 1.4048543e-02
	ROFT-:1-4: Summary file
		Fails for 1 entries
		Largest absolute error: 4.4709396e-01
		Largest relative error: 1.6932405e-02
	ROFT:1-2: Summary file
		Fails for 2 entries
		Largest absolute error: 5.9317017e+00
		Largest relative error: 1.3549482e-02
	ROFT:1-4: Summary file
		Fails for 1 entries
		Largest absolute error: 4.4709396e-01
		Largest relative error: 1.6932405e-02
	ROFT:3-4: Summary file
		Fails for 7 entries
		Largest absolute error: 5.4766541e+00
		Largest relative error: 1.6980304e-02
	RWFT-:1-4: Summary file
		Fails for 1 entries
		Largest absolute error: 5.8640862e-01
		Largest relative error: 1.9329853e-02
	RWFT:1-4: Summary file
		Fails for 1 entries
		Largest absolute error: 5.8640862e-01
		Largest relative error: 1.9329853e-02
```

I relax the relative tolerance further to be 2.e-2 from 1.e-2.

GitPaean:
jenkins build this opm-tests=806 please

GitPaean:
o, I oversaw the following comparison failure, 

```

Value index                 = 47
(first value, second value) = (-714.761, -1364.03)

Program threw an exception: [/var/lib/jenkins/workspace/opm-simulators-PR-builder/deps/opm-common/test_util/EclRegressionTest.cpp:202] Deviations exceed tolerances.
The absolute deviation is 649.265, and the tolerance limit is 0.02.
The relative deviation is 0.475991, and the tolerance limit is 0.02.
```

which is a little unfortunate.  

Will check this value on local running.

GitPaean:
The well curve comparison for `RGFT:1-2` is 

image

GitPaean:
jenkins build this please

GitPaean:
jenkins build this opm-tests=806 please

GitPaean:
From the local testing, the following two options are making the parallel regression tests harder, 

```
--ecl-enable-drift-compensation
```
and 
```
--relaxed-max-pv-fraction
```

When they are disabled, the parallel regression tests are much easier. 
Will update the PR.

GitPaean:
I can be wrong, for the specific case `TEST_RXFT`, the `--ecl-enable-drift-compensation` will affect the time stepping due to nonlinear convergence for the parallel regression test, and the option `--relaxed-max-pv-fraction` affects the digital comparison of the results in the parallel setting. 

maybe we can check whether `--relaxed-max-pv-fraction` is done well for parallel running?

GitPaean:
jenkins build this opm-tests=806 please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.021    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.023    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.005    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.03     |



View result details @ https://www.ytelses.com/opm/?page=result&id=1799

GitPaean:
jenkins build this opm-tests=806 please

GitPaean:
jenkins build this opm-tests=806 please

GitPaean:
At the end and I should tried this option earlier, it is the drift compensation activated by this PR makes the parallel regression testing difficult, which is the direct cause without saying it is the root problem. 

Although adjusting other options like `--relaxed-max-pv-fraction` and `--solver-max-time-step-in-days` can fix the regression test, but at the end, I decided to disable the drift compensation for these failed parallel regression tests. 

> There is some issue with the parallel support. Is wellsActive communicated?

I checked for two of the cases, it is not the cause.

GitPaean:
jenkins build this opm-tests=806 please

GitPaean:
The jenkins test is green now and the code self was approved by @totto82 . 

At the end, due to the parallel regression testing failures, I had to disable the drift compensation for those tests. I would like some approval regarding the adjustments regarding to those parallel tests. 

@totto82 @bska @akva2

totto82:
I am fine with disabling drift compensation for the failing tests. Thanks for the thorough testing.

GitPaean:
> I am fine with disabling drift compensation for the failing tests. Thanks for the thorough testing.

Then let us get it in. We will see how things will turn out since it will turn on the drift compensation for the majority of the cases now.

PR #4103, opened 2022-09-13
mpiserializer: handle all vectors

akva2:
Handle all vectors in eclmpiserializer and remove support in MPIPacker.

This contains some more changes, in particular we need to move the variadic broadcast support into the serializer since that previously relied on the vector support in the MPIPacker.

### DIFF
--- a/ebos/eclmpiserializer.hh
+++ b/ebos/eclmpiserializer.hh
@@ -369,30 +369,31 @@ public:
     //!
     //! \tparam T Type of class to broadcast
     //! \param data Class to broadcast
+    //! \param root Process to broadcast from
     template<class T>
-    void broadcast(T& data)
+    void broadcast(T& data, int root = 0)
     {
         if (m_comm.size() == 1)
             return;
 
-        if (m_comm.rank() == 0) {
+        if (m_comm.rank() == root) {
             try {
                 pack(data);
                 m_packSize = m_position;
-                m_comm.broadcast(&m_packSize, 1, 0);
-                m_comm.broadcast(m_buffer.data(), m_position, 0);
+                m_comm.broadcast(&m_packSize, 1, root);
+                m_comm.broadcast(m_buffer.data(), m_position, root);
             } catch (...) {
                 m_packSize = std::numeric_limits<size_t>::max();
-                m_comm.broadcast(&m_packSize, 1, 0);
+                m_comm.broadcast(&m_packSize, 1, root);
                 throw;
             }
         } else {
-            m_comm.broadcast(&m_packSize, 1, 0);
+            m_comm.broadcast(&m_packSize, 1, root);
             if (m_packSize == std::numeric_limits<size_t>::max()) {
                 throw std::runtime_error("Error detected in parallel serialization");
             }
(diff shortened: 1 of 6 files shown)


akva2:
jenkins build this please

ebos/eclmpiserializer.hh
```diff
@@ -95,12 +95,16 @@ public:
...
             optional(data);
+        } else if constexpr (is_vector<T>::value) {
+            vector(const_cast<T&>(data));
         } else if constexpr (is_map<T>::value) {
             map(const_cast<T&>(data));
         } else if constexpr (is_array<T>::value) {
             array(const_cast<T&>(data));
         } else if constexpr (is_set<T>::value) {
             set(const_cast<T&>(data));
+        } else if constexpr (has_serializeOp<detail::remove_cvr_t<T>>::value) {
```
blattms, line 106: Would have thought that this removes  `volatile` but it does not...

ebos/eclmpiserializer.hh
```diff
@@ -391,28 +413,66 @@ public:
             if (m_packSize == std::numeric_limits<size_t>::max()) {
                 throw std::runtime_error("Error detected in parallel serialization");
             }
+
             m_buffer.resize(m_packSize);
             m_comm.broadcast(m_buffer.data(), m_packSize, 0);
```
atgeirr, line 418: Er 0 her riktig?

akva2: nope, fixed.

blattms (approved):
One a side note:

I think we might loose something (speed memory) due this refactoring for contiguous arrays/vectors of pods. Those should actually be sized/packed/unpacked in one go by specifying the number of entries instead of using a loop over the  entries and make the operation on each entry. Hope hat can be changed.

ebos/eclmpiserializer.hh
akva2: indeed, i forgot to remove volatile.

akva2:
as vectors of pod; maybe, but this is not a regression. it was never actually utilized in the old MPIPacker support either. it was also a loop over the entries.

akva2:
jenkins build this please

ebos/eclmpiserializer.hh
```diff
@@ -64,7 +64,7 @@ decltype(auto) make_variant(std::size_t index)
 }
 
 template<class T>
-using remove_cvr_t = std::remove_const_t<std::remove_reference_t<T>>;
+using remove_cvr_t = std::remove_volatile_t<std::remove_const_t<std::remove_reference_t<T>>>;
```
bska, line 67: For what it's worth, `std::remove_cv_t<>` removes both `const` and `volatile`.

akva2: simplified.

akva2:
jenkins build this please

akva2:
merging with approval.

bska:
Did this [break](https://ci.opm-project.org/job/opm-simulators/1363/) the sequential build?

akva2:
it did. fix in https://github.com/OPM/opm-simulators/pull/4105

PR #4104, opened 2022-09-13
changed: handle all pairs in eclmpiserializer

akva2:
thus we can remove support in MPIPacker


akva2:
jenkins build this please

blattms (approved):
Approved if jenkins agrees

PR #4106, opened 2022-09-13
changed: handle tuples in eclmpiserializer

akva2:
thus we can remove support in MPIPacker

### DIFF
--- a/ebos/eclmpiserializer.hh
+++ b/ebos/eclmpiserializer.hh
@@ -89,12 +89,10 @@ public:
     {
         if constexpr (is_ptr<T>::value) {
             ptr(data);
-        } else if constexpr (is_pair<T>::value) {
-            pair(data);
+        } else if constexpr (is_pair_or_tuple<T>::value) {
+            tuple(data);
         } else if constexpr (is_variant<T>::value) {
             variant(data);
-        } else if constexpr (is_tuple<T>::value) {
-            tuple(data);
         } else if constexpr (is_optional<T>::value) {
             optional(data);
         } else if constexpr (is_vector<T>::value) {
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

ebos/eclmpiserializer.hh
```diff
@@ -498,6 +524,20 @@ protected:
           variadic_call(std::forward<Args>(args)...);
     }
 
+    template<std::size_t I = 0, typename Tuple>
+    typename std::enable_if<I == std::tuple_size<Tuple>::value, void>::type
```
blattms, line 528: Now I wonder why we don't treat pairs like tuple. :laughing: They should be accessible just like these.

blattms (approved):
Looks correct.

ebos/eclmpiserializer.hh
akva2: hmm, that's a good point. i'll see if i can make that work.

akva2:
jenkins build this please

akva2:
@blattms please re-review (i added tuple as pair)..

blattms (commented):
As in the other PR i now wonder whether we should have a specialization of `is_tuple` for pair and use that like
```
-  } else if constexpr (is_pair<T>::value) {
+ } else if constexpr (is_tuple<T>::value) {
-            pair(data);
+            tuple(data);
```
Other than that this looks great.

akva2:
okay, i have combined the predicates for pair and tuples.

akva2:
jenkins build this please

akva2:
i'll merge this now, comments are addressed.

PR #4107, opened 2022-09-13
Refactor MPIPacker

akva2:
to limit the overload party, we put packing details for
specific types in separate structs.
emit compiler error if unsupported type is given,
better to detect this on compile time rather than runtime.

also adds array packing for POD vector/arrays.

Sits on top of https://github.com/OPM/opm-simulators/pull/4106

### DIFF
--- a/ebos/eclmpiserializer.hh
+++ b/ebos/eclmpiserializer.hh
@@ -121,7 +121,7 @@ public:
     template <typename T>
     void vector(std::vector<T>& data)
     {
-        auto handle = [&](auto& d)
+        [[maybe_unused]] auto handle = [&](auto& d)
         {
             for (auto& it : d) {
               if constexpr (is_pair_or_tuple<T>::value)
@@ -139,15 +139,24 @@ public:
 
         if (m_op == Operation::PACKSIZE) {
             m_packSize += Mpi::Packer::packSize(data.size(), m_comm);
-            handle(data);
+            if constexpr (std::is_pod_v<T>)
+                m_packSize += Mpi::Packer::packSize(data.data(), data.size(), m_comm);
+            else
+                handle(data);
         } else if (m_op == Operation::PACK) {
-            Mpi::packer::pack(data.size(), m_buffer, m_position, m_comm);
-            handle(data);
+            Mpi::Packer::pack(data.size(), m_buffer, m_position, m_comm);
+            if constexpr (std::is_pod_v<T>)
+                Mpi::Packer::pack(data.data(), data.size(), m_buffer, m_position, m_comm);
+            else
+                handle(data);
         } else if (m_op == Operation::UNPACK) {
             size_t size;
             Mpi::Packer::unpack(size, m_buffer, m_position, m_comm);
             data.resize(size);
-            handle(data);
+            if constexpr (std::is_pod_v<T>)
+                Mpi::Packer::unpack(data.data(), size, m_buffer, m_position, m_comm);
+            else
+                handle(data);
         }
     }
 
@@ -184,7 +193,7 @@ public:
     {
         using T = typename Array::value_type;
 
-        auto handle = [&](auto& d) {
+        [[maybe_unused]] auto handle = [&](auto& d) {
             for (auto& it : d) {
                 if constexpr (is_pair_or_tuple<T>::value)
                     tuple(it);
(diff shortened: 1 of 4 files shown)


opm/simulators/utils/MPIPacker.hpp
```diff
@@ -19,179 +19,180 @@
...
+
+    static void unpack(T& data,
+                       std::vector<char>& buffer,
+                       int& position,
+                       Parallel::MPIComm comm)
+    {
+        unpack(&data, 1, buffer, position, comm);
+    }
+
+    static std::size_t packSize(const T*, std::size_t n, Parallel::MPIComm comm)
```
atgeirr, line 68: Are these members taking T* arguments intended to be public?

opm/simulators/utils/MPIPacker.hpp
```diff
@@ -19,179 +19,180 @@
...
+                       std::vector<char>& buffer,
+                       int& position,
+                       Parallel::MPIComm comm)
+    {
+        MPI_Unpack(buffer.data(), buffer.size(), &position, data, n,
+                   Dune::MPITraits<T>::getType(), comm);
+    }
+};
+
+//! \brief "Packging" for unsupported types.
```
atgeirr, line 96: Typo

opm/simulators/utils/MPIPacker.hpp
akva2: yes. it's the entry point for arrays.

opm/simulators/utils/MPIPacker.hpp
akva2: yes. i have a full revamp of the doxy coming up.

opm/simulators/utils/MPIPacker.hpp
```diff
@@ -19,179 +19,180 @@
...
+    template<> \
+    struct Packing<false,T> \
+    { \
+        static std::size_t packSize(const T&, Parallel::MPIComm); \
+        static void pack(const T&, std::vector<char>&, int&, Parallel::MPIComm); \
+        static void unpack(T&, std::vector<char>&, int&, Parallel::MPIComm); \
+    };
+
+ADD_PACK_SPECIALIZATION(std::string)
+ADD_PACK_SPECIALIZATION(time_point)
```
atgeirr, line 138: Is this for std::chrono::time_point? If so, how can you get away without namespace qualification?

akva2: because we are in the Opm namespace, and https://github.com/OPM/opm-common/blob/master/opm/common/utility/TimeService.hpp#L32

opm/simulators/utils/MPIPacker.hpp
akva2: see if it's better now..

atgeirr: Sure is!

ebos/eclmpiserializer.hh
```diff
@@ -90,9 +90,11 @@ public:
         if constexpr (is_ptr<T>::value) {
             ptr(data);
         } else if constexpr (is_pair<T>::value) {
-            pair(data);
+            tuple(data);
```
blattms, line 93: Maybe we should add a specialization of `is_tuple` for pair and use that instead of `is_pair` above.

ebos/eclmpiserializer.hh
```diff
@@ -177,15 +188,17 @@ public:
...
     void array(Array& data)
     {
         using T = typename Array::value_type;
 
-        auto handle = [&](auto& d) {
+        [[maybe_unused]] auto handle = [&](auto& d) {
             for (auto& it : d) {
                 if constexpr (is_pair<T>::value)
-                    pair(it);
+                    tuple(it);
```
blattms, line 201: same here: `is_tuple`

ebos/eclmpiserializer.hh
```diff
@@ -289,28 +314,28 @@ public:
         auto keyHandle = [&](auto& d)
         {
               if constexpr (is_pair<Key>::value)
-                  pair(d);
+                  tuple(d);
```
blattms, line 317: `is_tuple`?

opm/simulators/utils/MPIPacker.cpp
```diff
@@ -19,343 +19,118 @@
...
-}
-
-template<std::size_t I = 0, typename Tuple>
-typename std::enable_if<I == std::tuple_size<Tuple>::value, void>::type
-unpack_tuple_entry(Tuple&, std::vector<char>&, int&,
-                   Opm::Parallel::MPIComm)
-{
+             buffer.size(), &position, comm);
+    MPI_Pack(data.c_str(), length, MPI_CHAR, buffer.data(), buffer.size(),
+             &position, comm);
```
blattms, line 84: Aren't you make an assumption here that might be violated for some strings?

Previously we extracted the C string representation that is equivalent to the std::string (even for multibyte encodings). Now we are assuming that this is not a multibyte string (i.e. strleng(data.c_str()==data.size() holds).

blattms (commented):
I think we need to make string packing future proof.

ebos/eclmpiserializer.hh
akva2: changed in previous pr.

ebos/eclmpiserializer.hh
akva2: changed in previous PR.

ebos/eclmpiserializer.hh
akva2: done in previous pr.

bska:

bska:
This appears to [break](https://ci.opm-project.org/job/opm-simulators-PR-builder/3994/console) the sequential build.  Is the build failure the same issue as before?

akva2:
not really the same failure mode, but still me being stupid and easily fixable. thanks for checking that.

akva2:

blattms:
thanks.

PR #4108, opened 2022-09-14
Simplifications in serializer

akva2:
Simplify the user interface for the serializer. All ops can now go through operator().
Downstream of https://github.com/OPM/opm-common/pull/3148
Sits on top of https://github.com/OPM/opm-simulators/pull/4107

### DIFF
--- a/ebos/eclmpiserializer.hh
+++ b/ebos/eclmpiserializer.hh
@@ -94,13 +94,13 @@ public:
         } else if constexpr (is_optional<T>::value) {
             optional(data);
         } else if constexpr (is_vector<T>::value) {
-            vector(const_cast<T&>(data));
+            vector(data);
         } else if constexpr (is_map<T>::value) {
-            map(const_cast<T&>(data));
+            map(data);
         } else if constexpr (is_array<T>::value) {
-            array(const_cast<T&>(data));
+            array(data);
         } else if constexpr (is_set<T>::value) {
-            set(const_cast<T&>(data));
+            set(data);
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this opm-common=3148 please

ebos/eclmpiserializer.hh
```diff
@@ -515,6 +242,174 @@ public:
...
+        } else {
+            if (m_op == Operation::UNPACK) {
+                std::size_t size = 0;
+                (*this)(size);
+                auto& data_mut = const_cast<std::vector<T>&>(data);
+                data_mut.resize(size);
+                std::for_each(data_mut.begin(), data_mut.end(), std::reference_wrapper{*this});
+            } else {
+                (*this)(data.size());
+                std::for_each(data.begin(), data.end(), std::reference_wrapper{*this});
```
bska, line 274: Tiny nit: You must `#include <functional>` to use `std::reference_wrapper`.  Once you include that header you can also just use the [`std::ref()`](https://en.cppreference.com/w/cpp/utility/functional/ref) helper function to make your intent even more explicit.

bska (commented):
Looks good for the most part except for a small nit concerning `reference_wrapper`.  I like the simplification this work has brought.

ebos/eclmpiserializer.hh
akva2: include added, std:.ref used. i will trigger the new build from upstream.

bska (approved):
Thanks a lot for the update.  I'll merge this into master now.

PR #4109, opened 2022-09-14
Split up serializer and move main code to opm-common

akva2:
Downstream of https://github.com/OPM/opm-common/pull/3149
Sits on top of https://github.com/OPM/opm-simulators/pull/4108


bska (approved):
Thanks a lot for pushing through on this.  I'm really happy that we've gotten to this point and I'll merge into master.

PR #4110, opened 2022-09-14
avoiding wrong sign for QTotal primary variable for top segment

GitPaean:
for multisegment wells.

The scenario is that an injector is with zero rate control, and we get something like `-2.e-233` value for the primary variable and triggered assertion failure
```
MultisegmentWellEval.cpp:1783: void Opm::MultisegmentWellEval<FluidSystem, Indices, Scalar>::updateUpwindingSegments() [with FluidSystem = Opm::BlackOilFluidSystem<double, Opm::BlackOilDefaultIndexTraits>; Indices = Opm::BlackOilIndices<0, 0, 0, 0, false, false, 0, 0>; Scalar = double]: 
Assertion `! (baseif_.isInjector() && primary_variables_evaluation_[seg][GTotal] < 0.)' failed.
```

### DIFF
--- a/opm/simulators/wells/MultisegmentWellEval.cpp
+++ b/opm/simulators/wells/MultisegmentWellEval.cpp
@@ -463,6 +463,13 @@ updatePrimaryVariables(const WellState& well_state) const
             total_seg_rate += baseif_.scalingFactor(p) * segment_rates[baseif_.numPhases() * seg + p];
         }
 
+        if (seg == 0) {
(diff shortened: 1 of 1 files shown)


GitPaean:
A note for this PR is that, although the PR can avoid the assertion failure, but it looks like we can not solve the targeting well `an injector with zero injection target`.  Then the well got `SHUT`. But that is a separate problem.

GitPaean:
jenkins build this please

PR #4112, opened 2022-09-14
fixed: wrong root rank in variadic broadcast

akva2:
change the test to broadcast from rank 1 instead of 0 so this bug does not resurface.

i screwed up juggling all my branches and accidentially squashed the fix for this is the wrong branch. sorry for that.


akva2:
jenkins build this please

bska (approved):
Thanks.  I guess we should have caught this in review, but having an automated test is of course much better.  I'll merge into master.

PR #4115, opened 2022-09-16
fixed: remove default constructor declaration

akva2:
a class with a reference member cannot be default initialized, so constructor is implicitly deleted. quells a clang warning


akva2:
jenkins build this please

bska (approved):

PR #4117, opened 2022-09-16
added: command line option for zoltan customization

akva2:
modeled after the linsolver support. a few built-in presets, as well as reading options from a json file

Downstream of https://github.com/OPM/opm-grid/pull/586

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -108,6 +108,11 @@ struct ZoltanImbalanceTol {
 };
 
 template<class TypeTag, class MyTypeTag>
+struct ZoltanParams {
+    using type = UndefinedProperty;
+};
+
+template<class TypeTag, class MyTypeTag>
 struct AllowDistributedWells {
     using type = UndefinedProperty;
 };
@@ -160,6 +165,11 @@ struct SerialPartitioning<TypeTag, TTag::EclBaseVanguard> {
 template<class TypeTag>
 struct ZoltanImbalanceTol<TypeTag, TTag::EclBaseVanguard> {
     static constexpr double value = 1.1;
+};
+
+template<class TypeTag>
+struct ZoltanParams<TypeTag,TTag::EclBaseVanguard> {
+    static constexpr auto value = "graph";
 };
 
 template<class TypeTag>
@@ -235,10 +245,14 @@ public:
 
         EWOMS_REGISTER_PARAM(TypeTag, bool, OwnerCellsFirst,
                              "Order cells owned by rank before ghost/overlap cells.");
+#if HAVE_MPI
         EWOMS_REGISTER_PARAM(TypeTag, bool, SerialPartitioning,
                              "Perform partitioning for parallel runs on a single process.");
         EWOMS_REGISTER_PARAM(TypeTag, double, ZoltanImbalanceTol,
                              "Tolerable imbalance of the loadbalancing provided by Zoltan (default: 1.1).");
+        EWOMS_REGISTER_PARAM(TypeTag, std::string, ZoltanParams,
+                             "Configuration of Zoltan partitioner. Valid options are: graph (default)), hypergrah or scotch. Alternatively, you can request a configuration to be read from a JSON file by giving the filename here, ending with '.json.'");
+#endif
         EWOMS_REGISTER_PARAM(TypeTag, bool, AllowDistributedWells,
                              "Allow the perforations of a well to be distributed to interior of multiple processes");
         // register here for the use in the tests without BlackoildModelParametersEbos
@@ -263,8 +277,11 @@ public:
 #endif
 
         ownersFirst_ = EWOMS_GET_PARAM(TypeTag, bool, OwnerCellsFirst);
+#if HAVE_MPI
         serialPartitioning_ = EWOMS_GET_PARAM(TypeTag, bool, SerialPartitioning);
         zoltanImbalanceTol_ = EWOMS_GET_PARAM(TypeTag, double, ZoltanImbalanceTol);
+        zoltanParams_ = EWOMS_GET_PARAM(TypeTag, std::string, ZoltanParams);
+#endif
         enableDistributedWells_ = EWOMS_GET_PARAM(TypeTag, bool, AllowDistributedWells);
         ignoredKeywords_ = EWOMS_GET_PARAM(TypeTag, std::string, IgnoreKeywords);
         eclStrictParsing_ = EWOMS_GET_PARAM(TypeTag, bool, EclStrictParsing);
--- a/opm/simulators/utils/SetupZoltanParams.cpp
+++ b/opm/simulators/utils/SetupZoltanParams.cpp
@@ -1,0 +1,60 @@
+/*
+  Copyright 2022 SINTEF Digital, Mathematics and Cybernetics.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <config.h>
+#include <opm/simulators/utils/SetupZoltanParams.hpp>
+
+#include <opm/common/OpmLog/OpmLog.hpp>
+
+#include <boost/property_tree/json_parser.hpp>
+
+namespace Opm
+{
+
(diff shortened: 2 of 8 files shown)


opm/simulators/utils/SetupZoltanParams.cpp
```diff
@@ -0,0 +1,60 @@
...
+    std::map<std::string,std::string> result;
+
+    if (conf == "scotch") {
+        result.insert({"LB_METHOD", "GRAPH"});
+        result.insert({"GRAPH_PACKAGE", "Scotch"});
+    } else if (conf == "hypergraph") {
+        result.insert({"LB_METHOD", "HYPERGRAPH"});
+    } else if (conf == "graph") {
+        result.insert({"LB_METHOD", "GRAPH"});
+        result.insert({"GRAPH_PACKAGE", "PHG"});
```
bska, line 42: Any reason not to use
```C++
result.emplace(key, value);
```
instead of
```C++
result.insert({key, value});
```
here?  The former constructs the `value_type` in place while the latter is a construct+move.

opm/simulators/utils/SetupZoltanParams.cpp
```diff
@@ -0,0 +1,60 @@
...
+        boost::property_tree::ptree tree;
+        try {
+            boost::property_tree::read_json(conf, tree);
+        } catch (boost::property_tree::json_parser::json_parser_error& err) {
+            OpmLog::error(err.what());
+        }
+        for (const auto& node : tree) {
+            auto value = node.second.get_value_optional<std::string>();
+            if (value)
+              result[node.first] = *value;
```
bska, line 53: Any reason not to write this as
```C++
result.insert_or_assign(node.first, *value);
```
?

bska (commented):
Other than a couple of minor issues concerning how we construct the nodes in the options map this looks good to me.

opm/simulators/utils/SetupZoltanParams.cpp
akva2: no reason, changed.

opm/simulators/utils/SetupZoltanParams.cpp
akva2: no reason, changed.

ebos/eclbasevanguard.hh
```diff
@@ -235,10 +245,14 @@ public:
 
         EWOMS_REGISTER_PARAM(TypeTag, bool, OwnerCellsFirst,
                              "Order cells owned by rank before ghost/overlap cells.");
+#if HAVE_MPI
```
bska, line 248: Can you have a look at this please, @blattms?   I seem to recall that there is at least **one** workflow/use case that wants to leverage Zoltan support even if we don't have MPI.  OpenCL and/or Cuda, possibly?

akva2: i was actually looking into that. as is, there is no calls to that code, with or without mpi, i figured we can quell the options for now. but maybe that would break some unmerged code?

akva2: that being said, as we don't even link zoltan without mpi we would have more things to cover in that case.

akva2: aha, there's the call i could not findwhile grepping. but yeah, it's still under HAVE_MPI, https://github.com/OPM/opm-grid/blob/master/opm/grid/common/ZoltanPartition.hpp#L90

ebos/eclbasevanguard.hh
```diff
@@ -235,10 +245,14 @@ public:
...
         EWOMS_REGISTER_PARAM(TypeTag, bool, SerialPartitioning,
                              "Perform partitioning for parallel runs on a single process.");
         EWOMS_REGISTER_PARAM(TypeTag, double, ZoltanImbalanceTol,
                              "Tolerable imbalance of the loadbalancing provided by Zoltan (default: 1.1).");
+        EWOMS_REGISTER_PARAM(TypeTag, std::string, ZoltanParams,
+                             "Configuration of Zoltan partitioner. Valid options are: graph (default)), hypergrah or scotch. Alternatively, you can request a configuration to be read from a JSON file by giving the filename here, ending with '.json.'");
```
blattms, line 254: Can we somehow break this line?

Nobody will know the correct option names to use for the json file without researching the Zoltan user manual. Maybe provide a link?

I also wonder what the error message (if there is one at all) will look like if there are typos.

blattms (commented):
Shouldn't the new header appear somewhere in the CMake scripts?

Can we hide these options from normal users, somehow? Usage of them seems quite error prone and not for the faint of heart.

I really wonder what happens for mistakes like:
- typos in the json file
- requsting scotch if it is not there

blattms:
Frankly, you guys are a bit quick when it comes to adding new options which seem to be there mainly for testing purposes. :disappointed:

OPMUSER:
I agree with @blattms 200% here, we keep on adding command line options that the user is not expected to use, even worst there is absolutely no documentation on how to use it apart for a one line comment.

I would suggest the following, since we have two types of command line options:

1. `--help`  **New** user facing command line parameters that must be fully documented via the PR, so that the user at least can make an attempt to understand and use the feature. This needs to be more than a one line command that we currently have when using `--help.` The PR documentation will then be used in the written manual with additional commentary on how to use the feature, similar to what we do when we add specific OPM Flow keywords.
2. -`-help-all `**New** experimental command line parameters, that requires one line usage help, that is not intended for users to use.  I would also suggest that we put a time limit on these as well, that is, if the feature and command line parameter is not moved to the -`-help` category or incorporated into the input deck after a period of time, then the feature should be dropped and the command line parameter deleted. I would think one year testing period should be a suitable time period, but I would not object to alternative time periods. These features would not be documented in the manual.

Some background. 

I have been trying to write some guidance on the numerical controls for the manual, and it is just a mess from a user prospective. For example, we have`--solver-*` and `time-step-*`  with similar named features, but I guess different functionality. I mean how many growth factors does one need? There are a total of 127 user facing command line parameters for the  2022-04 release, most of them associated with the numeric control options. Now compare that with E100's TUNING keyword. I'm not saying that we should "dumb down" OPM Flow's features by any stretch, just that we need to have clarity on the features a user can and should use, and hide all the pedantic and experimental options, I mean, do really expect reservoir engineers to know what -`-ilu-reorder-spheres` option does? Perfectly okay for those that know what this means and does, but definitely not for reservoir engineers. 

 I have been to the Isles, so apologies for the rant.

akva2:
This is one option to set some 50 options, readily documented in publicly available documents for anyone who needs to understand the complexity. Exactly what you argue for with the magic number based tuning keyword. Damned if you do, damned if you don't...

ebos/eclbasevanguard.hh
notspiff: e.g.
```
Error: foo.json(2): expected ':'
```
and no scotch (actually parmetis in my case);
```
[0] Zoltan ERROR in Zoltan_Graph (line 103 of /build/trilinos-GM3Kol/trilinos-12.14.1/packages/zoltan/src/tpls/third_library.c):  ParMETIS partitioning was requested but Zoltan was compiled without ParMETIS.
```
followed by application abort.

akva2: ^ me with the wrong login.

atgeirr:
It is very important to get such experimental features merged to the master branch and made accessible. Even if we never change the default, this is how we can get new ideas and methods into the code in a way that can be experimented with, without breaking anyone's runs.

I can see that documenting and keeping track of an evolving set of command line parameters is tricky, however. I think separating "advanced" options from "regular" ones could be a way to go (similar to the `--help-all` suggestion from David), but not with any "make it a regular option or we kill it" mandatory approach. Some options are likely to be "advanced" for all eternity, and that is fine.

I think command line options like the one added here is showing the way forward. It allows easy selection of main method, and at the same time enourmous flexibility through the JSON option, similar to the approach for linear solvers. All this by adding a single new command line option. Comparing this to the situation pointed out by David for the options related to time step selection shows how much better it is (the `--solver-*` and `--time-step-*` options): for time steps we have many options with unclear interaction, and many options that are only used if a particular method variant is chosen via another option.

Making it "advanced", i.e. hiding it from the default `--help` text, is fine though.

ebos/eclbasevanguard.hh
blattms: > Error: foo.json(2): expected ':'

Nothing to add... :wink:

OPMUSER:
@atgeirr Well this all sounds reasonable, but without any form of documentation how is one suppose to use these experimental features. I not saying that they should be documented in the manual, because I don't think they should, but there needs to be something that another researcher can read in order to getting on using the feature without digging into the code. 

> modeled after the linsolver support. a few built-in presets, as well as reading options from a json file

Where is the documentation, web link etc., and why cannot this just be included in the PR? How hard is it write a paragraph with the PR to at least give people a head start. If you want researchers to use and test the feature, then a one line comment is not going to do this.

End of rant, hopefully :-)

atgeirr:
> Where is the documentation, web link etc., and why cannot this just be included in the PR? How hard is it write a paragraph with the PR to at least give people a head start. If you want researchers to use and test the feature, then a one line comment is not going to do this.

I believe that has been improved to the point of usefulness in the followup PR, for the zoltan options?

PR #4121, opened 2022-09-19
Damaris integ v3

ElyesAhmed:
This is a draft PR to integrate Damaris Library to OPM. The current test is only used to output "PRESSURE".


atgeirr:
@akva2 Can you take a look at the build system side of this?

ebos/eclwriter.hh
```diff
@@ -39,8 +39,22 @@
...
+#include <Damaris.h>
+#endif
+
+#if OPM_HAVE_DAMARIS
+#include <Damaris.h>
+#endif
+
+#if HAVE_DAMARIS
+#include <Damaris.h>
+#endif
```
akva2, line 48: i'd like to see this a fifth time so we are REALLY sure ;P

and don't remove the other includes.

opm-simulators-prereqs.cmake
```diff
@@ -17,7 +17,7 @@ set (opm-simulators_CONFIG_VAR
   DUNE_ISTL_VERSION_MAJOR
   DUNE_ISTL_VERSION_MINOR
   DUNE_ISTL_VERSION_REVISION
-  HAVE_SUITESPARSE_UMFPACK
+  HAVE_SUITESPARSE_UMFPACK
```
akva2, line 20: no.

CMakeLists.txt
```diff
@@ -542,6 +549,38 @@ if(HAVE_FPGA)
   ExternalProject_Get_Property(FPGA_library binary_dir)
   target_link_libraries(opmsimulators PUBLIC ${binary_dir}/fpga_lib_alveo_u280.a)
 endif()
+if (USE_DAMARIS_LIB) # User requested Damaris support
+        if (Damaris_FOUND)
```
akva2, line 554: please make a wrapper find module instead of putting all of this in the main cmakelists.

opm/simulators/flow/Main.hpp
```diff
@@ -197,18 +210,38 @@ class Main
 #endif
         EclGenericVanguard::setCommunication(std::make_unique<Parallel::Communication>());
 
-        handleTestSplitCommunicatorCmdLine_();
+/* Damaris I/O initialization has been moved to after the readDeck() call in Main::initalize_()
```
akva2: no new commented code. add it if/when it is required.

opm/simulators/utils/readDeck.cpp
```diff
@@ -266,9 +266,9 @@ namespace {
                                 *parseContext, errorGuard);
         }
 
-        if (eclipseState == nullptr) {
+        //if (eclipseState == nullptr) {
```
akva2: if we don't want the if, remove it.

opm/simulators/utils/readDeck.cpp
```diff
@@ -309,9 +309,9 @@ namespace {
                                        std::unique_ptr<Opm::WellTestState>& wtestState,
                                        std::shared_ptr<Opm::SummaryConfig>& summaryConfig)
     {
-        if (eclipseState == nullptr) {
+       // if (eclipseState == nullptr) {
```
akva2: ditto.

opm/simulators/utils/readDeck.cpp
```diff
@@ -477,7 +477,7 @@ Opm::setupLogging(const int          mpi_rank_,
     return output;
 }
 
-void Opm::readDeck(Opm::Parallel::Communication    comm,
+void Opm::readDeck(Opm::Parallel::Communication&    comm,
```
akva2: seems unrelated. these structs are very cheap to copy so that's been the preferred semantics.

akva2 (commented):
mostly commented on buildsystem, but also some other stuff i noticed.

alfbr:
I assume Damaris is a new dependency?

akva2:
it's an optional dependency.

alfbr:
> it's an optional dependency.

Sounds good. Has anybody considered the implications of deploying it? I am thinking about existence of rpms for Red Hat or debs for Ubuntu/Debian. In addition. is the code base well maintained?

bska:
> but also some other stuff i noticed.

@akva2: Those comments **mostly** appear to apply to `readDeck.cpp` which isn't touched at all in this PR.  Did GitHub bring in a stale ref before?

akva2:
I am seem to have accidentially clicked the commit listed in the commit message of the second commit.

ElyesAhmed:
> I am seem to have accidentially clicked the commit listed in the commit message of the second commit.

Yes Arne Morten, its two commits. However,  the second commit is still not a good set of commits with unclear messages. I wanted to preserve the authorship because Joshua Bowden from INRIA also contributed to this work! Maybe you have an idea on how to clean it under this constraint!

akva2:
you can safely split it, and use git commit --author to preserve authorship.

bska:
jenkins build this please

atgeirr:
The flaws mentioned previously are fixed, and I have made modifications to the code that I thought were necessary. I think this is ready to be merged, except for @akva2's request to put the Damaris build system stuff in its own function. I have not done this, as I am not very CMake-skilled, but I assume it is not so hard?

atgeirr:

akva2:
```
CMake Error at opm-common/cmake/Modules/OpmCompile.cmake:41 (add_library):
  Cannot find source file:

    /home/akva/kode/opm/opm-simulators/opm/simulators/utils/initDamarisXMLFile.cpp
```
XML -> Xml

ElyesAhmed:
I think we fixed all issues raised by reviewers now, including those related to the build-system

bska:

atgeirr:

atgeirr:
All green, merging. Thanks for contributing, reviewing and commenting!

ElyesAhmed:
> All green, merging. Thanks for contributing, reviewing and commenting!

thank you all!

blattms:
mmh, is this tested with DUNE 2.6?

atgeirr:
> mmh, is this tested with DUNE 2.6?

I assumed it was ok since Jenkins was green, but we may have to use a special trigger for that? In any case, do you see any problems? Would be a bit surprising, since this is not really touching on Dune things.

ebos/eclwriter.hh
```diff
@@ -281,10 +302,30 @@ public:
...
+            }
 
+            if (!isSubStep) {
+                data::Solution localCellData = {};
+                this->eclOutputModule_->assignToSolution(localCellData);
+                // Output the PRESSURE field
+                if (this->eclOutputModule_->getPRESSURE_ptr() != nullptr) {
+                    damaris_write("PRESSURE", (void*)this->eclOutputModule_->getPRESSURE_ptr());
+                    damaris_end_iteration();
+                }
```
bska, line 325: I didn't pick up on this before, but this sequence of statements risks reading from moved-from memory.  The reason is that [`assignToSolution()`](https://github.com/OPM/opm-simulators/blob/2064e8725f11a5ae00e0fd96a686809fd6eab983/ebos/eclgenericoutputblackoilmodule.cc#L645) currently [transfers](https://github.com/OPM/opm-simulators/blob/2064e8725f11a5ae00e0fd96a686809fd6eab983/ebos/eclgenericoutputblackoilmodule.cc#L655) ownership of its solution arrays into the `data::Solution` object, including that of the array that holds [`PRESSURE`](https://github.com/OPM/opm-simulators/blob/2064e8725f11a5ae00e0fd96a686809fd6eab983/ebos/eclgenericoutputblackoilmodule.cc#L683) values.  There is also no guarantee that `vector::data()` is `nullptr` after it has been moved from.

Furthermore, by calling `assignToSolution()` and just discarding the `localCellData`, you remove all the data arrays that would otherwise go to the ECLIPSE-compatible restart file.  Is that really intentional?  My suggestion here would be to remove the call to `assignToSolution()` here and just use the `PRESSURE_ptr()` functions exclusively.

blattms:
> > mmh, is this tested with DUNE 2.6?
> 
> I assumed it was ok since Jenkins was green, but we may have to use a special trigger for that? In any case, do you see any problems? Would be a bit surprising, since this is not really touching on Dune things.

Jenkins only checks 2.7. If you look at some of my current PRs (DUNE 2.9), you will notice that some of those also fix 2.6. I am just a bit anxious, this near to the release...

atgeirr:
> Jenkins only checks 2.7. If you look at some of my current PRs (DUNE 2.9), you will notice that some of those also fix 2.6. I am just a bit anxious, this near to the release...

Thanks for the info. I assume that we can rerun Jenkins on those PRs to know if we did something wrong here? If so, I will prioritize fixing it.

ebos/eclwriter.hh
atgeirr: Thanks for the explanation. I agree that the assignToSolution() is incorrect here. The PRESSURE_ptr() is a temporary API for testing, I think long term we should get the output arrays from the data::Solution object. Then it may be correct to reinstate it, or rather to make sure we only do assignToSolution() once by unifying with the parts further down in the code.

I will test and push a PR to remove the assignToSolution() call now.

PR #4124, opened 2022-09-20
- added spai0 and JacNew

hnil:
- added spai0
- removed synonyme of ILU0 and ParallelOverLappingILU0 and intrudced ILU for dune 2.8

This smoothers may be valuable for smothers in amg of the pressure solve in cpr. ILU may also have advantages over standard block ilu if cpr if resort is not used.


hnil:
Failing example:

hnil:
Closing — spai0/JacNew as AMG smoothers, opened 2022, no reviews in three years, and the linear-solver code has been reworked since. Branch kept if the smoothers are worth another look.

PR #4126, opened 2022-09-21
pass well state to output also for shut wells

totto82:
I think this is a better approach since it would not hide non-zero values in the wellState.

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -347,10 +347,6 @@ WellState::report(const int* globalCellIdxMap,
     data::Wells res;
     for( std::size_t well_index = 0; well_index < this->size(); well_index++) {
         const auto& ws = this->well(well_index);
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

bska (commented):
I don't know what you're trying to do, but this won't work without **a lot** more effort.  The entire output code basically depends on there not being dynamic data for shut wells.

totto82:
>I don't know what you're trying to do, but this won't work without a lot more effort. The entire output code basically depends on there not being dynamic data for shut wells.

I think my main concern is to handle cases where the wells is shut dynamically by the simulator. See https://github.com/OPM/opm-simulators/pull/4054

totto82:
Any idea why number of time-steps and convergence is changed by this PR?

OPMUSER:
> Need to check if the Wells are shut  or stopped, will check tomorrow

Sorry about the delay. For WVFPEXP cases 01-04 all wells have `WELSPECS(AUTO) = SHUT`, whereas, cases 05 and 06 have `STOP`

PR #4127, opened 2022-09-23
for the cells having zero thickness we use the cell center depth to do the equilibration

GitPaean:
In the targeting case, those cells happen to be numerical aquifer cells. 

The approach was along the direction suggested by @bska

### DIFF
--- a/ebos/equil/initstateequil.cc
+++ b/ebos/equil/initstateequil.cc
@@ -1864,8 +1864,18 @@ equilibrateHorizontal(const CellRange&  cells,
             totfrac += frac;
         }
 
-        saturations /= totfrac;
-        pressures   /= totfrac;
+        if (totfrac > 0.) {
+            saturations /= totfrac;
+            pressures /= totfrac;
+        } else {
+            // Fall back to centre point method for zero-thickness cells.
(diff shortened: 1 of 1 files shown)


bska (commented):
I think this makes sense.  I **suppose** we could bypass the accumulation loop altogether for zero-thickness cells using code like
```C++
if (cellZSpan_[cell].second > cellZSpan_[cell].first) {
    // do accumulation
}
else {
    // fall back to centre point approach
}
```
but this is a good compromise.  Moreover, zero-thickness cells are unlikely to be plentiful in most models so the above is probably not worth the additional complexity.

Would you mind adding a comment to the `else` branch to say something like
```C++
// Fall back to centre point method for zero-thickness cells.
```
to make it clear that we're dealing with a very special case?

bska:
jenkins build this please

GitPaean:
Thanks for the comments. I was also thinking the approach you mentioned.  Since you agree that the current way is acceptable, let us get in.

GitPaean:
jenkins build this please

bska (approved):
Thanks for the updates and for analysing the underlying issue.  I'll merge this into master now.

PR #4133, opened 2022-09-27
Fix PBUB and PDEW Export to Restart File

OPMUSER:
The saturation pressures written to the restart used SI units instead of the pressure units for the model. This PR fixes this.


bska:
jenkins build this please

bska (approved):
Thanks for identifying this incorrect unit tag.  I'll merge into master.

PR #4136, opened 2022-09-30
Remove reordering for opencl

Tongdongq:
This PR removes the explicit reordering of the rows of A.
This used to be done to improve memory performance on the FPGA (streaming data instead of random access).
A GPU does not need this, and not reordering reduces the amount of computation, and the number of variables.
Instead of reordering, the kernels receive the extracted parallelism information, and choose the correct rows accordingly.
The `opencl_ilu_reorder` parameter is renamed to `opencl_ilu_parallel` and changed to a bool.
The only supported way to extract parallelism is using level_scheduling, the graph_coloring option is too complex to implement without reordering, and with the addition of the jacobi matrix for ILU, it is no longer needed.
The FPGA is already broken since 68b604c8, merged in May 2022. Since there seems to be no interest in the FPGASolver, I'll just remove it and graph_coloring in a next PR.

I encountered a segfault on a simula node when using CPR on CPU, but I cannot reproduce this on my machine.
The segfault occurs during `PreconditionerFactory::create()` inside `FlexibleSolver::initOpPrecSp()`.

```
Notation:
X: A (B+C), D, E, F | G
X: Number of Jacobi blocks
A: Total time
B: Assembly time
C: Linear solve time
D: Overall Linearizations
E: Overall Newton Iterations
F: Overall Linear Iterations
G: Number of fallbacks to Dune
AMD EPYC Milan 7763, AMD MI100, openclSolver, vary jacobi blocks
2022-9-27:
no reordering (new), separate, 5c4d1a3f:
  0: 607 (176+226), 2015, 1638, 25015 | 0
  8: 535 (178+152), 2036, 1656, 24793 | 0
150: 469 (172+ 98), 1966, 1595, 27208 | 0
reordering (old), separate, 3434e9e3:
  0: 609 (178+228), 2027, 1652, 25420 | 0
  8: 528 (170+158), 1967, 1595, 25011 | 0
150: 470 (170+102), 1929, 1562, 26326 | 0
no reordering (new), coupled, 5c4d1a3f:
  0: 644 (171+271), 1967, 1593, 24496 | 0
  8: 568 (170+195), 1962, 1586, 26565 | 0
150: 461 (163+103), 1875, 1509, 27472 | 0
reordering (old), coupled, 3434e9e3:
  0: 646 (170+275), 1950, 1582, 23942 | 0
  8: 559 (166+195), 1928, 1560, 24136 | 0
150: 472 (167+111), 1895, 1529, 27215 | 0

2022-9-14:
no reordering, separate, 14f317a6 (no longer exists):
  0: 551 (149+219), 1972, 1597, 24690 | 0
  8: 487 (147+153), 1991, 1613, 24448 | 0
150: 415 (142+ 96), 1912, 1545, 26861 | 0
reordering, separate, e7d186b4:
  0: 562 (145+235), 1951, 1582, 24991 | 0
  8: 472 (141+154), 1933, 1562, 24542 | 0
150: 440 (149+108), 1998, 1625, 26996 | 0

```


Tongdongq:
Jenkins failed because the header `ILUReorder.hpp` was deleted in this PR, but not every reference to it.

bska:
jenkins build this please

opm/simulators/linalg/FlowLinearSolverParameters.hpp
```diff
@@ -323,7 +323,7 @@ namespace Opm
             EWOMS_REGISTER_PARAM(TypeTag, std::string, AcceleratorMode, "Use GPU (cusparseSolver or openclSolver) or FPGA (fpgaSolver) as the linear solver, usage: '--accelerator-mode=[none|cusparse|opencl|fpga|amgcl]'");
             EWOMS_REGISTER_PARAM(TypeTag, int, BdaDeviceId, "Choose device ID for cusparseSolver or openclSolver, use 'nvidia-smi' or 'clinfo' to determine valid IDs");
             EWOMS_REGISTER_PARAM(TypeTag, int, OpenclPlatformId, "Choose platform ID for openclSolver, use 'clinfo' to determine valid platform IDs");
-            EWOMS_REGISTER_PARAM(TypeTag, std::string, OpenclIluReorder, "Choose the reordering strategy for ILU for openclSolver and fpgaSolver, usage: '--opencl-ilu-reorder=[level_scheduling|graph_coloring], level_scheduling behaves like Dune and cusparse, graph_coloring is more aggressive and likely to be faster, but is random-based and generally increases the number of linear solves and linear iterations significantly.");
+            EWOMS_REGISTER_PARAM(TypeTag, bool, OpenclIluParallel, "Parallelize ILU decomposition and application on GPU. Default: true");
```
blattms, line 326: Please explain what this means. How is it parallelized?

opm/simulators/linalg/bda/Reorder.cpp
```diff
@@ -332,6 +332,7 @@ void findLevelScheduling(int *CSRColIndices, int *CSRRowPointers, int *CSCRowInd
                 nextActiveRowIndex++;
             }
         }
+        rowsToStart.clear();
```
blattms, line 335: Nice catch! :+1:

opm/simulators/linalg/bda/opencl/BILU0.cpp
```diff
@@ -156,15 +139,24 @@ bool BILU0<block_size>::analyze_matrix(BlockedMatrix *mat, BlockedMatrix *jacMat
...
+        rowsPerColorPrefix[i + 1] = rowsPerColorPrefix[i] + rowsPerColor[i];
     }
+
     err |= queue->enqueueWriteBuffer(s.rowsPerColor, CL_FALSE, 0, (numColors + 1) * sizeof(int), rowsPerColorPrefix.data(), nullptr, &events[1]);
 
+    if (opencl_ilu_parallel) {
+        err |= queue->enqueueWriteBuffer(s.rowIndices, CL_FALSE, 0, Nb * sizeof(unsigned), fromOrder.data(), nullptr, &events[2]);
+    } else {
+        // rowsPerColorPrefix is misused here
+        // s.rowIndices[i] == i must hold
```
blattms, line 156: A tiny bit more information please. For what is misused and where?

opm/simulators/linalg/bda/opencl/ChowPatelIlu.cpp
```diff
@@ -756,9 +756,7 @@ void ChowPatelIlu<block_size>::decomposition(
     for(int i = 0; i < Nb; ++i) {
         for(int k = Ut->rowPointers[i]; k < Ut->rowPointers[i+1]; ++k) {
             int j = Ut->colIndices[k];
-            if (j == i) {
-                inverter(Ut->nnzValues + k * bs * bs, invDiagVals + i * bs * bs);
-            } else {
```
blattms, line 761: Out of curiosity (or dumbness): What exactly does this change mean (why does this result in exact inversion)?

opm/simulators/linalg/bda/opencl/openclSolverBackend.cpp
```diff
@@ -497,23 +475,20 @@ void openclSolverBackend<block_size>::copy_system_to_gpu() {
 #if COPY_ROW_BY_ROW
     int sum = 0;
     for (int i = 0; i < Nb; ++i) {
-        int size_row = rmat->rowPointers[i + 1] - rmat->rowPointers[i];
-        memcpy(vals_contiguous.data() + sum, reinterpret_cast<double*>(rmat->nnzValues) + sum, size_row * sizeof(double) * block_size * block_size);
+        int size_row = mat->rowPointers[i + 1] - mat->rowPointers[i];
+        memcpy(vals_contiguous.data() + sum, reinterpret_cast<double*>(mat->nnzValues) + sum, size_row * sizeof(double) * block_size * block_size);
```
blattms, line 479: Was ther before, but anyway: Why does this need a reinterpret_cast? Shouldn't BlockedMatriy::nnzValues be double* anyway?

opm/simulators/linalg/bda/opencl/openclSolverBackend.cpp
```diff
@@ -537,17 +512,18 @@ void openclSolverBackend<block_size>::update_system_on_gpu() {
...
```
blattms, line 516: ditto

tests/test_cusparseSolver.cpp
```diff
@@ -113,7 +113,7 @@ testCusparseSolver(const boost::property_tree::ptree& prm, Matrix<bz>& matrix, V
     auto wellContribs = Opm::WellContributions::create("cusparse", false);
     std::unique_ptr<Opm::BdaBridge<Matrix<bz>, Vector<bz>, bz> > bridge;
     try {
-        bridge = std::make_unique<Opm::BdaBridge<Matrix<bz>, Vector<bz>, bz> >(accelerator_mode, fpga_bitstream, linear_solver_verbosity, maxit, tolerance, platformID, deviceID, opencl_ilu_reorder, linsolver);
+        bridge = std::make_unique<Opm::BdaBridge<Matrix<bz>, Vector<bz>, bz> >(accelerator_mode, fpga_bitstream, linear_solver_verbosity, maxit, tolerance, platformID, deviceID, opencl_ilu_parallel, linsolver);
```
blattms, line 116: Please break this looooooonnnnnnngggggg line.

tests/test_openclSolver.cpp
```diff
@@ -111,7 +111,7 @@ testOpenclSolver(const boost::property_tree::ptree& prm, Matrix<bz>& matrix, Vec
     auto wellContribs = Opm::WellContributions::create("opencl", false);
     std::unique_ptr<Opm::BdaBridge<Matrix<bz>, Vector<bz>, bz> > bridge;
     try {
-        bridge = std::make_unique<Opm::BdaBridge<Matrix<bz>, Vector<bz>, bz> >(accelerator_mode, fpga_bitstream, linear_solver_verbosity, maxit, tolerance, platformID, deviceID, opencl_ilu_reorder, linsolver);
+        bridge = std::make_unique<Opm::BdaBridge<Matrix<bz>, Vector<bz>, bz> >(accelerator_mode, fpga_bitstream, linear_solver_verbosity, maxit, tolerance, platformID, deviceID, opencl_ilu_parallel, linsolver);
```
blattms, line 114: ditto

blattms (commented):
Sorry for the long delay.
Code looks good AFAIK.  Just a few changes/comments requested. Most important is the parameter documentation.

I will test compilation.

blattms:
> I encountered a segfault on a simula node when using CPR, but I cannot reproduce this on my machine.
> The segfault occurs during PreconditionerFactory::create() inside FlexibleSolver::initOpPrecSp().

Is that for the CPR onthe CPU?

OPMUSER:
@Tongdongq Consider renaming the command line parameters such that all the GPU parameters start with `gpu-` so the commands all get grouped together. This will make life easier for one and all. :-)

Tongdongq:
> > I encountered a segfault on a simula node when using CPR, but I cannot reproduce this on my machine.
> 
> Is that for the CPR onthe CPU?

Yes

@OPMUSER That's a good idea

blattms:
But the parameter is not only gpu, is it? Also FPGA, but maybe we can neglect that (or use accelerator, which might further confuse engineers).
If we start such a change it should also be consistent for all parameters used only for the gpu/accelerators. Maybe that is better suited for a separate PR.

blattms:
> > > I encountered a segfault on a simula node when using CPR, but I cannot reproduce this on my machine.
> > 
> > Is that for the CPR on the CPU?
>
> Yes.

I am becoming haunted by these types of segfaults these days. :grimacing: At least it is unrelated to this PR.

Tongdongq:
The FPGA is already broken, and should be removed after this PR.
One issue with multiple backends is with amgcl/vexcl.

opm/simulators/linalg/FlowLinearSolverParameters.hpp
Tongdongq: Using level scheduling. When false, it performs serially and is extremely slow, this should only be used for debugging.
We might not actually need runtime access for this setting.

blattms: I am struggeling a bit.
It is parallel anyway, right just different (more parallel with level scheduling. Should we make that more obvious? E,g,
```
            EWOMS_REGISTER_PARAM(TypeTag, bool, GpuIluLevelScheduling, "Use level scheduling for ILU decomposition and application on GPU for more parallelism. Default: true");
```

Tongdongq: When false, it is not parallel, every row is done alone.
When true, it is parallelized using level scheduling, which functions the same as before. The difference is that graph coloring is removed.
Grouping GPU params together is fine, but cusparse also uses level scheduling, which cannot be turned off.

blattms:
Thank a lot. Finally this goes in. Sorry for the delay.

Issue #3932, opened 2022-06-03
Preliminary plans for experimental multi-GPU support


atgeirr:
We are looking at how to best support multi-GPU systems. This is relevant for instance for large cases run on modern HPC clusters that often have several GPUs per node (and multiple nodes), and we have access to such a system for testing that is equipped with multiple Nvidia GPU per node.

We think that the ISTL approach to separating the Krylov solver and preconditioner is sound, and can be used here. Our preliminary plan for making this proof of concept is:

1. Create an ISTL Preconditioner subclass that implements its actions using the GPU, initially using cuSparse, same as the existing GPU implementation. The communication will then be handled using MPI with no code changes, everything GPU related will be inside the preconditioner, that for each rank will communicate with separate GPUs. The solver will still be the ISTL Krylov solver running on the CPU, with data residing on the CPU as well.

2. The above will involve copying data between GPU and CPU for every linear iteration. To avoid this, in the next step we will use the ISTL Krylov solver with different matrix and vector classes, where the data now resides on the GPU. These classes would be taking on the roles of Dune::BCRSMatrix etc. within the limited scope of the Krylov solver. The Krylov solver will hopefully remain unchanged.

3. The last step currently envisioned is to modify the Krylov solver as necessary to optimize performance, by kernel fusion or other means.

An advantage of the above, ISTL-centric approach, is that it is relatively easy to support alternative preconditioners, such as AMGX, in a flexible way. While we initially plan the P.O.C. using existing Cuda implementations, we do not want to limit ourselves and the proposed approach explicitly avoids proprietary protocols for GPU-GPU communication.

Any comments welcome! @Tongdongq, @blattms?

Thanks for your consideration,
@atgeirr and @kjetilly

Tongdongq:
Sounds good!
There is ILU and CPR now, and SPAI is being worked on by @ducbueno.
Could these all be supported in the future?
By ISTL, do you mean dune-istl?
Would this work on a single node with multiple GPUs, or multiple nodes, each having multiple GPUs or both?
Could you do this in rocsparse or similar, instead of cuda? Or do you have a specific existing cuda implementation in mind?
How would this interact with the new block-jacobi-ILU preconditioner? Could each MPI process use it?

atgeirr:
> Sounds good!
> There is ILU and CPR now, and SPAI is being worked on by @ducbueno.
> Could these all be supported in the future?

Yes, we do not intend to limit ourselves.

[...]

PR #3945, opened 2022-06-17
Cuda ilu0

fgfuchs:
Start to implement #3932. Not ready for review or merging.

PR #4137, opened 2022-09-30
Multi-GPU for the linear solver support through a Dune::ISTL compatible Cusparse wrapper

kjetilly:
This addresses #3932. This pull requests provides the following:

1) Various small utilities for doing safe programming in CUDA (error checks, memory smart pointers, etc)

2) A Dune::ISTL compatible interface to cusparse and cublas, allowing us to run a large class of `Dune::IterativeSolver`s on the GPU without modifications to the solver (provided the solver only uses rudimentary linear operations, such as PBiCGStab).

3) **Multi-GPU support for the linear solver** through a CUDA compatible version of `Dune::OwnerOverlapCopyCommunication` .

The bulk of the heavy GPU lifting happens in `CuVector`, `CuSparseMatrix`, `CuSeqILU0` and `vector_operations.cu`, while the interface towards the rest of the OPM code is mostly handled through `PreconditionerAdapter` and `SolverAdapter`.

I believe (though have not tested) that this code should be `hipify`-able to HIP code, making it possible to run it also on AMD GPUs. My suggestion is that this is done at a later stage, as not to introduce another level of complexity or source of errors to the code at this point 

I furthermore believe that this design will make a good foundation to port a larger part (read: assembly) of the simulator to the GPU. 

To run the code on GPUs, one simply needs to set the linear solver to `cubicgstab` and the preconditioner to `CUILU0`, in eg. a JSON file as:

```json
{
    "preconditioner" : {"type":"CUILU0"},
    "solver" : "cubicgstab"
}
```
This version also supports running only the preconditioner on the GPU, while the rest of the solver does the linear operations on the CPU (this is mostly for testing and benchmarking purposes, though it does show the modularity of the design). To run this, specify a JSON solver file as

```json
{
    "preconditioner" : {"type":"CUILU0"},
    "solver" : "bicgstab"
}
```

Simple tests [should be tested by more people as well] on the Norne dataset indicate the added overhead of this approach compared to a more down-to-the-metals approach is negligible. I believe this difference will become even smaller on larger datasets.


bska:
I haven't been following this work at all so I'll just ask two **very** basic questions: Does this start to implement the plan laid out in #3932?  Is it in any way related to #3945?

kjetilly:
Yes. This is in effect a solution to #3932 and should be seen as a replacement of the PR #3945

bska:
> This is in effect a solution to #3932 and should be seen as a replacement of the PR #3945

Excellent!  Thanks a lot for information.

bska:
jenkins build this please

bska:
jenkins build this please

kjetilly:
So the warnings I should (mostly) be able to fix if they are important to remove. They are mostly related to the fact that Cusparse/cublas uses `int` for sizes, while dune and the rest of the STL use `size_t`$\approx$`unsigned long`. (I was hoping to postpone this until C++20's `cmp_less` etc became available, but maybe not...)

Now, the failed unit tests all seem to be because the Jenkins build server does not have a working CUDA device(is that correct?). There are several ways of fixing this. I can have a check in `main()` of the respective unit tests checking if there is a GPU, and successfully exiting if not. Personally, I do not like this option, because a silent failure like this could lead someone to believe their GPU is working fine for OPM, while in fact, it is not. Maybe there is a way to specify that a test file should only be run with a GPU available?

atgeirr:
> So the warnings I should (mostly) be able to fix if they are important to remove.

Generally, yes. There does not seem to be all that many unique ones anyway.



> Maybe there is a way to specify that a test file should only be run with a GPU available?

Sounds like the best option.

This PR is very large, and should be split for ease of reviewing. I suggest splitting along the following lines:
 - Generic CUDA utilities (perhaps the impl/ directory).
 - The CuVector, CuSparseMatrix and related general basic linear algebra functionality. 
 - The CuSeqILU0 class.
 - The wrappers and other classes that turn this preconditioner (and the solver) into a dune-istl compatible option.

If not split in different PRs, at least split in different commits. Then different people can more easily review the parts they are qualified to. For example, I can review the last part, but the first three are better reviewed by others.

I also suggest to rename impl/ to detail/, but others may disagree.

opm/simulators/linalg/cuistl/CuSeqILU0.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+
+namespace Opm::cuistl
+{
+
+template <class M, class X, class Y, int l>
+CuSeqILU0<M, X, Y, l>::CuSeqILU0(const M& A, scalar_field_type w)
+    : m_underlyingMatrix(A)
+    , m_w(w)
+    , m_LU(CuSparseMatrix<field_type>::fromMatrix(impl::makeMatrixWithNonzeroDiagonal(A)))
```
blattms, line 44: Will this change the original matrix? (I recall a function doing that with a const_cast. Not sure whether it was this) In that case do we really want this?

opm/simulators/linalg/cuistl/CuSeqILU0.cpp
```diff
@@ -0,0 +1,362 @@
...
+INSTANTIATE_CUSEQILU0_DUNE(double, 4);
+INSTANTIATE_CUSEQILU0_DUNE(double, 5);
+INSTANTIATE_CUSEQILU0_DUNE(double, 6);
+
+INSTANTIATE_CUSEQILU0_DUNE(float, 1);
+INSTANTIATE_CUSEQILU0_DUNE(float, 2);
+INSTANTIATE_CUSEQILU0_DUNE(float, 3);
+INSTANTIATE_CUSEQILU0_DUNE(float, 4);
+INSTANTIATE_CUSEQILU0_DUNE(float, 5);
+INSTANTIATE_CUSEQILU0_DUNE(float, 6);
```
blattms, line 362: Out of curiosity: Was this used in real model testing?

opm/simulators/linalg/cuistl/CuSparseMatrix.cpp
```diff
@@ -0,0 +1,319 @@
...
+    std::vector<int> columnIndices;
+    std::vector<int> rowIndices;
+
+    rowIndices.push_back(0);
+
+    const int blockSize = matrix[0][0].N();
+    const int numberOfRows = matrix.N();
+    const int numberOfNonzeroBlocks = matrix.nonzeroes();
+    const int numberOfNonzeroElements = blockSize * blockSize * numberOfNonzeroBlocks;
+#ifndef CUSPARSE_ASSUME_UNSAFE_SPARSITY
```
blattms, line 80: Please document what is meant here.

opm/simulators/linalg/cuistl/CuSparseMatrix.cpp
```diff
@@ -0,0 +1,319 @@
...
+template <typename T>
+void
+CuSparseMatrix<T>::umv(const CuVector<T>& x, CuVector<T>& y) const
+{
+    CHECK_SIZE(x)
+    CHECK_SIZE(y)
+    if (blockSize() < 2) {
+        OPM_THROW(
+            std::invalid_argument,
+            "CuSparseMatrix<T>::usmv and CuSparseMatrix<T>::mv are only implemented for block sizes greater than 1.");
```
blattms, line 222: Where does this restriction come from?

opm/simulators/linalg/cuistl/CuSparseMatrix.cpp
```diff
@@ -0,0 +1,319 @@
...
+                                               x.data(),
+                                               &beta,
+                                               y.data()));
+}
+
+template <typename T>
+void
+CuSparseMatrix<T>::umv(const CuVector<T>& x, CuVector<T>& y) const
+{
+    CHECK_SIZE(x)
```
blattms, line 217: Any reason not to call usmv with alpha=1.0?

opm/simulators/linalg/cuistl/CuSparseMatrix.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+    virtual ~CuSparseMatrix();
+
+    template <class MatrixType>
+    static CuSparseMatrix<T> fromMatrix(const MatrixType& matrix);
+
+    void setUpperTriangular();
+    void setLowerTriangular();
+    void setUnitDiagonal();
+    void setNonUnitDiagonal();
```
blattms, line 66: Some more documentation would really be useful for new interfaces.

blattms (commented):
Premature comments. Only looked at maybe 14/44 files. This is a quite massive PR to review.

Would really appreciate better documentation (at least on the interface parts). I know this is a general problem of many parts of OPM, but for helping colleagues this should really be improved in general.

PR #4139, opened 2022-09-30
fixing (partly) confusion naming in function initializeWellPerfData()

GitPaean:
When reading the function, because different names, `perforation`, `connection` and `completion` are all used in this function, which makes the code rather confusing to read. And it appears to me they are all referring to the same thing `Connection`.  

The modification is only limited in this function. There is not intention to convert further `perf` to `connection` in a larger scale, which might be a good thing to do in a later stage.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -709,53 +709,53 @@ initializeWellPerfData()
     well_perf_data_.resize(wells_ecl_.size());
     int well_index = 0;
     for (const auto& well : wells_ecl_) {
-        int completion_index = 0;
+        int connection_index = 0;
         // INVALID_ECL_INDEX marks no above perf available
-        int completion_index_above = ParallelWellInfo::INVALID_ECL_INDEX;
+        int connection_index_above = ParallelWellInfo::INVALID_ECL_INDEX;
         well_perf_data_[well_index].clear();
         well_perf_data_[well_index].reserve(well.getConnections().size());
         CheckDistributedWellConnections checker(well, local_parallel_well_info_[well_index].get());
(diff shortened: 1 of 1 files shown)


GitPaean:
At the same time, I might misunderstand this part of the code. Please voice out if that is the case.

GitPaean:
jenkins build this please

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -709,53 +709,53 @@ initializeWellPerfData()
...
-                completion_index_above = completion_index;
+                connection_index_above = connection_index;
             } else {
-                checker.connectionFound(completion_index);
-                if (completion.state() != Connection::State::SHUT) {
+                checker.connectionFound(connection_index);
+                if (connection.state() != Connection::State::SHUT) {
                     OPM_THROW(std::runtime_error,
-                              "Completion state: " << Connection::State2String(completion.state()) << " not handled");
+                              "Completion state: " << Connection::State2String(connection.state()) << " not handled");
```
bska, line 749: Maybe you can fix the diagnostic message too, i.e., instead of
```
Completion state: ... not handled
```
we output
```
Connection state: ... not handled
```

bska (commented):
Thanks.  This is a straightforward improvement over the current master sources and I'll be happy to merge it.  Just one minor nit that I think we should address first.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
GitPaean: Thanks for noticing. Updated.

GitPaean:
jenkins build this please

bska (approved):
Thanks for the updates.  I'll merge into master now.

PR #4140, opened 2022-09-30
Use CMake's find_file to find install.py

### LINKED ISSUE #4138: installl.py not found after failed builds
```
/usr/bin/python3.9: can't open file '/home/mblatt/src/dune/opm-2.9/opm-common//install.py': [Errno 2] No such file or directory
```
should be home/mblatt/src/dune/opm-2.9/opm-common/python/install.py.

blattms:
This seems to be a more stable approach as it does not produce
```
/usr/bin/python3.9: can't open file
'/home/mblatt/src/dune/opm-2.9/opm-common//install.py': [Errno 2] No such file or directory
```
for rebuilds after failed attempts.

Closes #4138

### DIFF
--- a/python/simulators/CMakeLists.txt
+++ b/python/simulators/CMakeLists.txt
@@ -56,22 +56,20 @@ if(OPM_ENABLE_PYTHON_TESTS)
       -m unittest test/test_schedule.py)
 endif()
 
-set(_opm_common_python_script_dir "${opm-common_DIR}/python")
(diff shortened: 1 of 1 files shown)


akva2 (approved):
While I don't quite understand the circumstances for the failure, using file_find is "always" a better approach so looks good.

blattms:
jenkins build this please

PR #4143, opened 2022-10-04
only output final change of well control

totto82:
Note that the full switching history (i.e. pr iteration) is still reported to the debug file. 

The change was suggested by @OPMUSER

### DIFF
--- a/opm/simulators/wells/WellInterfaceGeneric.cpp
+++ b/opm/simulators/wells/WellInterfaceGeneric.cpp
@@ -435,19 +435,19 @@ void WellInterfaceGeneric::reportWellSwitching(const SingleWellState& ws, Deferr
     if (well_control_log_.empty())
         return;
 
-    std::string msg = "    Well " + name()
-        + " control mode changed from ";
-    for (const std::string& from : well_control_log_) {
-        msg += from + "->";
-    }
+    std::string from = well_control_log_[0];
     std::string to;
     if (isInjector()) {
         to = Well::InjectorCMode2String(ws.injection_cmode);
     } else {
         to = Well::ProducerCMode2String(ws.production_cmode);
     }
-    msg += to;
-    deferred_logger.info(msg);
+    // only report the final switching
+    if (from != to) {
+        std::string msg = "    Well " + name()
+        + " control mode changed from " + from + " to " + to;
+        deferred_logger.info(msg);
+    }
 }
 
 std::optional<double>


totto82:
jenkins build this please

bska:
Does this change mean that there's no directly visible signal to the user in the case that the control oscillates between two states and happens to settle on the original control mode?

totto82:
>Does this change mean that there's no directly visible signal to the user in the case that the control oscillates between two states and happens to settle on the original control mode?

You will need to look into the .DBG file or if it oscillates more than 3 times it would give a warning that it does not switch anymore.

bska:
> > Does this change mean that there's no directly visible signal to the user in the case that the control oscillates between two states and happens to settle on the original control mode?
> 
> You will need to look into the .DBG file or if it oscillates more than 3 times it would give a warning that it does not switch anymore.

Right, but unless you get that warning you won't necessarily know that you need to look in the `.DBG` file to infer that there's a lot of control mode switching going on.

OPMUSER:
> Could you comment on this PR
> 
> https://github.com/OPM/opm-simulators/pull/4143
> 
> it is supposed to be inline with your suggestion. 
> BHP->THP->ORAT->THP
> is reduced to BHP->THP
> and BHP->THP->BHP
> is reduced to nothing. 

@totto82 I thought I did comment previously, but never mind. The new behavior is similar to the commercial simulator, in that it only reports the changes at the end of the time step, and not for each non-linear iteration, which is why it was suggested, so this is okay. 

> Right, but unless you get that warning you won't necessarily know that you need to look in the .DBG file to infer that there's a lot of control mode switching going on.

@bska If it is not an numeric issue, this type of behavior normally occurs as the well is dying, that is it cannot flow under the current operating conditions, so I don't have a problem with @totto82  implementation. Experience engineers will always check the debug file, if there a numeric issues. 

However, I do think we need to have some control over what gets printed to the debug file, as it seems to be the dumping ground for all things. Why is the WELSPECS and other standard reports dumped to the debug file for example. Getting rid of these reports from the debug file would make it more useful.

 I have suggested the RPTDEBUG keyword to control what gets written, but at last no takers :-)

bska:
> it [the `.DBG` file] seems to be the dumping ground for all things.

That's true, and it's intentional.  Whether or not it's a good idea is possibly debatable, but we did decide quite early on that the `.DBG` file gets **all** messages emitted through the logging system.

> If it is not an numeric issue, this type of behavior [oscillating control modes] normally occurs as the well is dying, that is it cannot flow under the current operating conditions,

Right.  I still think I'd like to see just a little hint on the console and/or in the `.PRT` file that there's control mode switching going on just to alert the user of the possible numeric issue.  Something along the lines of
```C++
if (from != to) {
    // Log control mode change
}
else if (number_of_changes > threshold) {
    // Log number of changes for well
}
```

OPMUSER:
> Right. I still think I'd like to see just a little hint on the console and/or in the .PRT file that there's control mode switching going on just to alert the user of the possible numeric issue. Something along the lines of

@bska I would not object either way, not a big deal from my point of view.

atgeirr:
I will merge this now. I think the suggestion for adding a little extra hint if the control oscillated during the step is good, it can be added in a separate PR. What would a good threshold be for that BTW? I guess maybe 5?

PR #4145, opened 2022-10-05
initialize THP wells with non-zero rates

totto82:
Zero rates is a bad start for wells set to THP control on startup as the vfp table often does not have entries for very low rates. This change fixes an issue exposed by #4098

### DIFF
--- a/opm/simulators/wells/SingleWellState.cpp
+++ b/opm/simulators/wells/SingleWellState.cpp
@@ -156,8 +156,14 @@ double SingleWellState::sum_solvent_rates() const {
 void SingleWellState::update_producer_targets(const Well& ecl_well, const SummaryState& st) {
     const double bhp_safety_factor = 0.99;
     const auto& prod_controls = ecl_well.productionControls(st);
-    if (prod_controls.hasControl(Well::ProducerCMode::THP))
+    if (prod_controls.hasControl(Well::ProducerCMode::THP)) {
         this->thp = prod_controls.thp_limit;
+        // For THP controlled wells we initialize with some non-zero rates
+        // to get a more resonable starting point for the lookup in the vpf table.
+        for (int p = 0; p < pu.num_phases; ++p) {
+            this->surface_rates[p] = -1.0;
+        }
+    }
 
 
     auto cmode_is_bhp = (prod_controls.cmode == Well::ProducerCMode::BHP);


totto82:
jenkins build this please

atgeirr (approved):

opm/simulators/wells/SingleWellState.cpp
```diff
@@ -156,8 +156,14 @@ double SingleWellState::sum_solvent_rates() const {
 void SingleWellState::update_producer_targets(const Well& ecl_well, const SummaryState& st) {
     const double bhp_safety_factor = 0.99;
     const auto& prod_controls = ecl_well.productionControls(st);
-    if (prod_controls.hasControl(Well::ProducerCMode::THP))
+    if (prod_controls.hasControl(Well::ProducerCMode::THP)) {
         this->thp = prod_controls.thp_limit;
+        // For THP controlled wells we initialize with some non-zero rates
+        // to get a more resonable starting point for the lookup in the vpf table.
+        for (int p = 0; p < pu.num_phases; ++p) {
+            this->surface_rates[p] = -1.0;
```
bska, line 164: So this is effectively 1 cubic metre per second, right?  If so, maybe you can add a comment to that effect?

totto82: The rate should be in the range of the vfp tables FLO entries. Since they are typically in SM3/day I think I would like to change it to 100 cubic meter pr day. It shouldn't be too sensitive to the actual number as long as it is not too close to zero. And yes it make sense to a comment.

totto82: more testing indicates that the convergence (or more the lack of convergence) is more sensitive to the choice of this more or less random rate that I anticipated. i.e. i need to go back and investigate more so I converted it to draft.

GitPaean:
We does have some results regarding zero rates calculation with VFP search. Are you interested in investigating a little bit?  Basically, it shows with zero rates, the VFPPROD table gives 12.65 bar pressure drop for this case. 

```
summary.x E100_NETBALAN_TEST02. GPR:FIELD GPR:PROD GOPR:PROD GWPR:PROD GGPR:PROD | more
-- Days   dd/mm/yyyy          GPR:FIELD         GPR:PROD        GOPR:PROD        GWPR:PROD        GGPR:PROD 
------------------------------------------------------------------------------------------------------------
   0.00   01/11/2018                100              100                0                0                0 
   0.10   01/11/2018                100           112.65                0                0                0 
   0.40   01/11/2018                100           112.65                0                0                0 
   1.30   02/11/2018                100           112.65                0                0                0 
   3.30   04/11/2018                100           112.65                0                0                0 
   5.30   06/11/2018                100           112.65                0                0                0 
   7.30   08/11/2018                100           112.65                0                0                0 
   9.30   10/11/2018                100           112.65                0                0                0 
  11.30   12/11/2018                100           112.65                0                0                0 
  13.30   14/11/2018                100           112.65                0                0                0 
  15.30   16/11/2018                100           112.65                0                0                0 
  17.30   18/11/2018                100           112.65                0                0                0 
  19.30   20/11/2018                100           112.65                0                0                0 
  21.30   22/11/2018                100           112.65                0                0                0 
  23.30   24/11/2018                100           112.65                0                0                0 
  25.30   26/11/2018                100           112.65                0                0                0 
  27.30   28/11/2018                100           112.65                0                0                0 
  28.65   29/11/2018                100           112.65                0                0                0 
  30.00   01/12/2018                100           112.65                0                0                0 
  32.00   03/12/2018                100          110.322          5434.22        0.0342927           589341 
  33.29   04/12/2018                100          110.193          5125.75        0.0316609           555946 
  34.83   05/12/2018                100          110.088          4872.72        0.0294522           528558 
```

opm/simulators/wells/SingleWellState.cpp
GitPaean: I think it makes sense because the ratios also matters a lot when coming to VFP table search.

PR #4146, opened 2022-10-05
Refactor property updating in EclProblem

akva2:
introduce a common function for looping over grid cells and updating something based on compressedDofIdx and the intensive quantities. do this by passing a std::function (ie lambdas).

update various functions having this structure to use the new common function. this to reduce code duplication. while at it const some more variables and reduce some long statements.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -2209,33 +2209,18 @@ private:
 
     bool updateMaxOilSaturation_()
     {
-        const auto& simulator = this->simulator();
         int episodeIdx = this->episodeIndex();
 
         // we use VAPPARS
         if (this->vapparsActive(episodeIdx)) {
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this please

bska:
I like it&mdash;this is a great improvement in usability.  On a side note, the standard library prefers to place functions and other callable entities as the **last** parameter to an algorithm, e.g.,
```C++
auto pos = std::find_if(first, last, predicate);
```
Did you consider that alternative?  We **might** also consider making the type of the per-element operation into a template parameter.  That way we (potentially) avoid the cost of virtual function calls through `std::function<>`.

akva2:
i did both. i doubt the 6 virtual function dispatches per time step would make any difference, but then again there is also no reason not to.

PR #4151, opened 2022-10-06
only enforcing production economic limit to producers.

GitPaean:
After consideration, decided not to enforce this through the parser level, since a well can switch back and forth between injector and producer.

At the end, it has to be confirmed through testing.

### DIFF
--- a/opm/simulators/wells/WellInterfaceFluidSystem.cpp
+++ b/opm/simulators/wells/WellInterfaceFluidSystem.cpp
@@ -725,6 +725,11 @@ updateWellTestStateEconomic(const SingleWellState& ws,
         return;
     }
 
+    if (this->isInjector()) {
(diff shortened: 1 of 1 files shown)


GitPaean:
jenkins build this please

GitPaean:
Although comparing with adding some logic to the parser, there might be some different consequence when a complicated `SCHEDULE` is used. But this PR itself should be good to go in since it is correct NOT to shut any injector due to economic limit, which is clear with the definition of `WECON`.

GitPaean:
jenkins build this please

bska (approved):

GitPaean:
> I guess the provided test case should also be added to the regression tests? Agreed @vkip?

Maybe let us modify an existing one to avoid increasing the numbers of the regression tests?

PR #4155, opened 2022-10-10
set alq before trying to solve the equation during well testing

totto82:


totto82:
jenkins build this please

totto82:
Note. This fixes issue with gaslift/GASLIFT-03 which is still not under testing.

PR #4156, opened 2022-10-11
start regression testing for SPE1CASE2_KRNUM

tskille:
Feature has been merged into master and simulation results have been verify. Hence we are ready to start regression testing on this model,

### DIFF
--- a/tests/update_reference_data.sh
+++ b/tests/update_reference_data.sh
@@ -194,6 +194,7 @@ tests[bo_diffusion]="flow diffusion BO_DIFFUSE_CASE1"
 tests[fpr_nonhc]="flow water-1ph WATER2F"
 tests[actionx_wpimult]="flow actionx ACTIONX_WPIMULT"
 tests[wvfpexp_02]="flow wvfpexp WVFPEXP-02"
+tests[spe1case2_krnum]="flow spe1 SPE1CASE2_KRNUM"
 
 changed_tests=""
 
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -1219,6 +1219,13 @@ add_test_compareECLFiles(CASENAME wvfpexp_02
                          REL_TOL ${rel_tol}
                          DIR wvfpexp)
 
+add_test_compareECLFiles(CASENAME spe1case2_krnum
+                         FILENAME SPE1CASE2_KRNUM
+                         SIMULATOR flow
+                         ABS_TOL ${abs_tol}
+                         REL_TOL ${rel_tol}
+                         DIR spe1)
+
 # Restart tests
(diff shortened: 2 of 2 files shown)


OPMUSER:
Can we move it to a KRNUM directory and name the file KRNUM-01 instead. Then  I'll add a a few test cases and document via markdown with plots etc.  The reason for the request, is that it is head to find specific tests if they are stored in the SPE1 or other SPE directories. I'll update the file when I get a minute, to reference the the SPE01 files via the Path keyword. . 

I'm currently working through the group control cases, so it will be a while.

alfbr:
> Can we move it to a KRNUM directory and name the file KRNUM-01 instead. Then I'll add a a few test cases and document via markdown with plots etc. The reason for the request, is that it is head to find specific tests if they are stored in the SPE1 or other SPE directories. I'll update the file when I get a minute, to reference the the SPE01 files via the Path keyword. .

I see the problem. We did have a discussion on folder re-org earlier, but did not conclude. I suggest we revive the discussion with the aim of organizing the folder structure better. I can call in a meeting with all four of us (present in this PR). However, please do not let that postpone this pull request. After all, there are already multiple variants of SPE1 under testing in this folder, so the discussion is bigger than the test in this PR.

tests/update_reference_data.sh
```diff
@@ -194,6 +194,7 @@ tests[bo_diffusion]="flow diffusion BO_DIFFUSE_CASE1"
 tests[fpr_nonhc]="flow water-1ph WATER2F"
 tests[actionx_wpimult]="flow actionx ACTIONX_WPIMULT"
 tests[wvfpexp_02]="flow wvfpexp WVFPEXP-02"
+spe1case2_krnum="flow spe1 SPE1CASE2_KRNUM"
```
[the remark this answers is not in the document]
tskille: Sorry my bad.

tests/update_reference_data.sh
```diff
@@ -194,6 +194,7 @@ tests[bo_diffusion]="flow diffusion BO_DIFFUSE_CASE1"
 tests[fpr_nonhc]="flow water-1ph WATER2F"
 tests[actionx_wpimult]="flow actionx ACTIONX_WPIMULT"
 tests[wvfpexp_02]="flow wvfpexp WVFPEXP-02"
+tests[spe1case2_krnum=]"flow spe1 SPE1CASE2_KRNUM"
```
bska, line 197: Oops.  It seems you've accidentally swapped the assignment and the right bracket (`]`) here.

tskille: this is getting embarrassing :-). Now this should be good. Again very sorry

bska:

bska (approved):
Thank you for the updates! This looks good to me now and we've successfully integrated the new test case into the regression test suite.  I'll install the new reference solutions in OPM/opm-tests#822 and then merge this PR into the master branch.

PR #4157, opened 2022-10-11
Fix compilation with amgcl/vexcl and without CUDA/OpenCL.

blattms:
There is no reason to omit WellContributions if neither CUDA nor OpenCL is found as the implementations/functions only use C++- To the contrary, the defines used for deactivating were not consistent with usage elsewhere and resulted in compiler erorrs for exotic configuration

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -253,10 +253,8 @@ namespace Opm {
             // subtract B*inv(D)*C * x from A*x
             void apply(const BVector& x, BVector& Ax) const;
 
-#if HAVE_CUDA || HAVE_OPENCL
             // accumulate the contributions of all Wells in the WellContributions object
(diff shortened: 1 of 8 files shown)


blattms:
jenkins build this please

bska:
This **may have** [broken the sequential build](https://ci.opm-project.org/job/opm-simulators/1410/console).  At least we get undefined references to
```
Opm::WellContributions::addMatrix(Opm::WellContributions::MatrixType, int*, double*, unsigned int)
```
now.  Please investigate and/or fix.  This might be a release blocker.

atgeirr:
Can confirm I see the same.

atgeirr:
Looks like the WellContributions.cpp is within
```
if(CUDA_FOUND OR OPENCL_FOUND OR HAVE_FPGA OR HAVE_AMGCL)
```
but that the usage inside the various addWellContribution() functions was not similarly guarded (guard removed).

Can you fix this @blattms?

PR #4161, opened 2022-10-12
Move include of config.h to cpp-file

blattms:
One should never include config.h in a header. To the very least it results in redefined macros.

Let's hope we don't rely on this elsewhere (have not tried compilation :grimacing: )

### DIFF
--- a/ebos/eclinterregflows.cc
+++ b/ebos/eclinterregflows.cc
@@ -19,6 +19,10 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif // HAVE_CONFIG_H
+
 #include <ebos/eclinterregflows.hh>
 
 #include <algorithm>
--- a/opm/core/props/BlackoilPhases.cpp
+++ b/opm/core/props/BlackoilPhases.cpp
@@ -16,6 +16,11 @@
   You should have received a copy of the GNU General Public License
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif // HAVE_CONFIG_H
+
 #include <algorithm>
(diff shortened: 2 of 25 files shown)


blattms:
jenkins build this please

atgeirr:
Good catch! When looking at this, I wonder what is "HAVE_CONFIG_H"? We have plenty of plain `#include "config.h"` without that.

akva2:
because this file is so ancient that it comes from an autotools age i presume.

akva2:
we always set HAVE_CONFIG_H so it will be included. always.

blattms:
It must be generated, otherwise the compiler would not have warned.

blattms:
Just noticed that there definitely missing includes somewhere. Need to fix these.

atgeirr:
> Just noticed that there definitely missing includes somewhere. Need to fix these.

Jenkins was green, I almost merged this. Curious where the missing ones are?

blattms:
jenkins build this please

ebos/eclinterregflows.cc
```diff
@@ -19,6 +19,10 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif // HAVE_CONFIG_H
```
bska, line 24: It obviously doesn't hurt and I appreciate the consistency.  It also doesn't help because there's nothing in this file that depends on configuration parameter.

atgeirr:
Thanks for taking the effort of identifying this. Should we remove the HAVE_CONFIG_H everywhere while at it? (Hopefully just a quick shell command away.)

blattms:
I don't care. But then this has to be done everywhere. I'm lacking the necessary magic to do it, anyway.

atgeirr:
> But then this has to be done everywhere. I'm lacking the necessary magic to do it, anyway.

I agree, and will merge this now in any case.

PR #4164, opened 2022-10-12
Make Reservoir Voidage Rate Calculator Independent

bska:
This commit adds a new overload of the `calcReservoirVoidageRates()` member function.  The new overload does not depend on the user calling `defineState()`, but instead takes state variables as direct input arguments.

Reimplement the state-dependent overload in terms of the new function.  The immediate use case is calculating additional dynamic quantities per segment in a multisegmented well model, mostly for reporting purposes.

### DIFF
--- a/opm/simulators/wells/RateConverter.hpp
+++ b/opm/simulators/wells/RateConverter.hpp
@@ -24,10 +24,12 @@
 
 #include <opm/core/props/BlackoilPhases.hpp>
 #include <opm/grid/utility/RegionMapping.hpp>
-#include <opm/simulators/wells/RegionAttributeHelpers.hpp>
 #include <opm/simulators/utils/DeferredLoggingErrorHelpers.hpp>
+#include <opm/simulators/wells/RegionAttributeHelpers.hpp>
+
 #include <dune/grid/common/gridenums.hh>
 #include <dune/grid/common/rangegenerators.hh>
+
 #include <algorithm>
 #include <cmath>
 #include <memory>
@@ -36,6 +38,7 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
+
 /**
  * \file
  * Facility for converting component rates at surface conditions to
@@ -56,13 +59,13 @@ namespace Opm {
          * The conversion uses fluid properties evaluated at average
          * hydrocarbon pressure in regions or field.
          *
-         * \tparam FluidSystem Fluid system class. Expected to be a BlackOilFluidSystem
+         * \tparam FluidSystem Fluid system class. Expected to be a
+         *    BlackOilFluidSystem
          *
-         * \tparam Region Type of a forward region mapping.  Expected
-         * to provide indexed access through \code operator[]()
-         * \endcode as well as inner types \c value_type, \c
-         * size_type, and \c const_iterator.  Typically \code
-         * std::vector<int> \endcode.
+         * \tparam Region Type of a forward region mapping.  Expected to
+         *    provide indexed access through \code operator[]() \endcode as
+         *    well as inner types \c value_type, \c size_type, and \c
+         *    const_iterator.  Typically \code std::vector<int> \endcode.
          */
         template <class FluidSystem, class Region>
         class SurfaceToReservoirVoidage {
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska:
I have pushed an update which reverts the strategy for calculating the flowing `Rs` and `Rv` values back to the approach used in the [current master branch](https://github.com/OPM/opm-simulators/blob/2064e8725f11a5ae00e0fd96a686809fd6eab983/opm/simulators/wells/RateConverter.hpp#L411-L426). My preferred approach to that calculation is still present in the file, but it is disabled behind a preprocessor macro (`BURN_RESV_BRIDGES`) which is explicitly defined to be zero (off) in this PR.

Using the original approach means that we expectedly do not see any regression test failures.  I do not see any such problems in local testing but will run build check using the official project CI system to verify.

bska:
jenkins build this please

atgeirr:
> BURN_RESV_BRIDGES

Love that macro name... But what bridges would that burn? I thought this was a refactoring that should not change any results?

bska:
> But what bridges would that burn? I thought this was a refactoring that should not change any results?

Sorry for not being explicit.  This is/was indeed intended as a pure refactoring to make the function callable without having to call `defineState()` previously.  I also, initially, took the opportunity to make the calculation of the `Rs` and `Rv` values derived from the surface flow rates more rigorous.   The [current master sources](https://github.com/OPM/opm-simulators/blob/2064e8725f11a5ae00e0fd96a686809fd6eab983/opm/simulators/wells/RateConverter.hpp#L411-L426) implicitly that all surface flow rates are non-negative in this context and I wanted to relax that restriction.  My initial approach therefore used `std::copysign()` and `std::clamp()` to guarantee that the flowing `Rs`/`Rv` values were in the range zero to a maximum value passed in by the caller&mdash;e.g.the average values for a FIP region.

That approach however ended up breaking a number of regression tests (see [CI build 4098](https://ci.opm-project.org/job/opm-simulators-PR-builder/4098/testReport/) for details) and so I decided to disable that aspect of this PR by a preprocessor macro and to reinstate the current master behaviour by default.  This makes those failing regression tests pass.  All of this is handled in the new private member function [`dissolvedVaporisedRatio()`](https://github.com/OPM/opm-simulators/blob/e962e067e69a74bf54d3d9afa44e4509d08269bd/opm/simulators/wells/RateConverter.hpp#L554).

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

opm/simulators/wells/RateConverter.hpp
```diff
@@ -552,6 +552,26 @@ namespace Opm {
                     return { rs, rv };
                 }
 
+#define BURN_RESV_BRIDGES 0
+
+#if !BURN_RESV_BRIDGES
+
+                auto b = rs;
+                if (io >= 0 && ig >= 0) {
+                    b = surface_rates[ig]/(surface_rates[io]+1.0e-15);
```
akva2, line 561: spaces around operators

opm/simulators/wells/RateConverter.hpp
```diff
@@ -552,6 +552,26 @@ namespace Opm {
...
+                }
+                const double Rs = std::min(rs, b);
+
+                b = rv;
+                if (io >= 0 && ig >= 0) {
+                    b = surface_rates[io]/(surface_rates[ig]+1.0e-15);
```
akva2, line 567: spaces around operators.

akva2 (approved):
Looks fine. Just a couple of cosmetics spotted.

bska:
> Looks fine. Just a couple of cosmetics spotted.

Thanks.  I've amended the PR accordingly.

bska:
jenkins build this please

bska:
Merging this per prior approval.

PR #4166, opened 2022-10-13
removing the usage of GTotal in the well code

GitPaean:
It was a mistake from the very beginning and partly replaced by WQTotal. 

This PR aiming at removing the usage of GTotal at all and replacing it with WQTotal.  WQTotal is a more proper name.


GitPaean:
jenkins build this please

bska (approved):
Thanks a lot!  This will greatly reduce the opportunity for confusion.  In reading through this part of the code recently I was worried that I might be missing some subtle difference between `GTotal` and `WQTotal`.  I'll merge into master.

PR #4168, opened 2022-10-14
Add Serialization Support for std::bitset<3>

bska:
Needed in upcoming work for communicating dynamic phase quantities per segment in a multisegmented well.


bska:
jenkins build this please

PR #4171, opened 2022-10-17
Make related BC data a single class member

akva2:
Pack all BC data in a single struct, instead of having 6 (12) separate vector members.

Cleaner organization imo.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1670,47 +1670,11 @@ public:
             unsigned interiorDofIdx = context.interiorScvIndex(spaceIdx, timeIdx);
             unsigned globalDofIdx = context.globalSpaceIndex(interiorDofIdx, timeIdx);
             unsigned pvtRegionIdx = pvtRegionIndex(context, spaceIdx, timeIdx);
-            switch (indexInInside) {
-            case 0:
-                if (freebcXMinus_[globalDofIdx])
-                    values.setFreeFlow(context, spaceIdx, timeIdx, initialFluidStates_[globalDofIdx]);
-                else
-                    values.setMassRate(massratebc_.data[1][globalDofIdx], pvtRegionIdx);
-                break;
-            case 1:
-                if (freebcX_[globalDofIdx])
-                    values.setFreeFlow(context, spaceIdx, timeIdx, initialFluidStates_[globalDofIdx]);
-                else
-                    values.setMassRate(massratebc_.data[0][globalDofIdx], pvtRegionIdx);
-                break;
-            case 2:
-                if (freebcYMinus_[globalDofIdx])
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -2874,22 +2863,22 @@ private:
...
+                        data = &freebc_.plus[1];
                         break;
                     case FaceDir::ZMinus:
-                        data = &freebcZMinus_;
+                        data = &freebc_.minus[2];
                         break;
                     case FaceDir::ZPlus:
-                        data = &freebcZ_;
+                        data = &freebc_.plus[2];
                         break;
```
blattms, line 2882: Why not get rid of this switch-case by using a container for both plus and minus that allows lookup with `bcface.dir`like `data = freebc.data[bcface_dir]`. But maybe that is not worth it as there are other swich-cases that won't be affected.

blattms (approved):
Just one minor/silly question

ebos/eclproblem.hh
akva2: i did consider it, but found it too much for too little gain. but i can do it if you insist.

blattms: Nope. I have the same impression now.

akva2:
okay, I pushed a little harder. I think this version is better.

akva2:
jenkins build this opm-common=3180 please

ebos/eclproblem.hh
```diff
@@ -3018,20 +2914,35 @@ private:
...
+        {
+            for (auto& d : data)
+                d.resize(size, defVal);
+        }
+
+        const std::vector<T>& operator()(FaceDir::DirEnum dir) const
+        {
+            if (dir == FaceDir::DirEnum::Unknown)
+                throw std::runtime_error("Tried to acess BC data for the 'Unknown' direction");
+            int idx = 31 - __builtin_clz(static_cast<int>(dir));
```
bska, line 2932: I **really** don't think we should use compiler intrinsics.  I'd prefer we use something more portable&mdash;e.g., our home-grown [`count_leading_zeros()`](https://github.com/OPM/opm-common/blob/f414cdd30e5792b1198cfe08082a7605b28668b8/src/opm/input/eclipse/Schedule/WriteRestartFileEvents.cpp#L61) function or its 32-bit equivalent.

akva2: if that's a concern i'd rather just use the 2 line while loop. okay?

akva2:
jenkins build this opm-common=3180 please

bska (approved):
Thanks.  This is a much welcome simplification.  I'll merge into master.

PR #4173, opened 2022-10-17
[bugfix] Unconditionally compile WellContributions.cpp et al

blattms:
Otherwise we get undefined references. Fixes problems introduced with #4157


blattms:

bska (approved):
Very good.  I'll merge this now to restore the sequential build.

PR #4176, opened 2022-10-18
Well primaryvariable scaling

hnil:
- make scaling of primary variables of wells
- scale control equation if it is bhp control
Idea is to be able to make scaling of the matrices more normal. It is a follow up on the scaling of the reservoir primary variables.

### DIFF
--- a/opm/simulators/wells/WellInterfaceEval.cpp
+++ b/opm/simulators/wells/WellInterfaceEval.cpp
@@ -342,11 +342,11 @@ assembleControlEqProd_(const WellState& well_state,
         break;
     }
     case Well::ProducerCMode::BHP: {
-        control_eq = bhp - controls.bhp_limit;
+        control_eq = (bhp - controls.bhp_limit)*this->bhpControlScaling();
         break;
     }
     case Well::ProducerCMode::THP: {
-        control_eq = bhp - bhp_from_thp();
+        control_eq = (bhp - bhp_from_thp())*this->bhpControlScaling();
         break;
     }
     case Well::ProducerCMode::GRUP: {
--- a/opm/simulators/wells/WellInterfaceEval.cpp
+++ b/opm/simulators/wells/WellInterfaceEval.cpp
@@ -47,7 +47,7 @@ namespace Opm
 template<class FluidSystem>
 WellInterfaceEval<FluidSystem>::
 WellInterfaceEval(const WellInterfaceFluidSystem<FluidSystem>& baseif)
-  : baseif_(baseif), bhp_scaling_(1.0), rate_scaling_(1.0)
+  : baseif_(baseif)
 {}
 
 template<class FluidSystem>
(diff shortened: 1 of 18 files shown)


GitPaean:
jenkins build this please

GitPaean:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1.004    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 0.999    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 0.991    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.997    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 0.996    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1830

GitPaean:
Do you have some ideas about what kind of number should be used to benefit a typical blackoil simulation running?

opm/simulators/wells/WellInterfaceEval.cpp
```diff
@@ -342,11 +342,11 @@ assembleControlEqProd_(const WellState& well_state,
         break;
     }
     case Well::ProducerCMode::BHP: {
-        control_eq = bhp - controls.bhp_limit;
+        control_eq = (bhp - controls.bhp_limit)*this->bhpControlScaling();
```
GitPaean, line 345: From here,  if the bhp_scale_ and bhp_control_scale are different values, how will it affects the tolerance checking?

GitPaean:
hei, one small thing is that, for the `bhp_scaling`, they are used to scale to reduce the values for the related primary variables through division, while for `bhp_control_scaling`, it is used to reduce the orders of the `control_eq` through multiplication. As a result, `bhp_scaling` will be a big value maybe in the order or bar level, while `bhp_control_scaling` will be a very small value. 

Do you think  it is easier to use if we can change this to make both `bhp_scaling` and `bhp_control_scaling` to be big values?

hnil:
Closing: five of the seventeen touched files no longer exist after the Ebos rename, and the well primary-variable code has moved on considerably since 2022. Well primary-variable scaling is better approached against the current StandardWellPrimaryVariables. Branch retained on my fork.

hnil:
Reopening as a draft to keep it alive as a marker rather than lose the topic. Well primary-variable scaling is still worth reviewing at some point.

To be clear about its state: this cannot be rebased as-is -- five of the seventeen touched files no longer exist after the Ebos rename. Treat it as a record of the idea, to be rewritten against the current StandardWellPrimaryVariables when someone picks it up.

hnil:
Superseded by #7286, which is the same idea done properly: the scaling is applied to the AD derivative only, so stored values stay physical and B/C are untouched, and it comes with measured conditioning numbers for D. Branch kept.

PR #4178, opened 2022-10-18
Add rocalutionSolver

Tongdongq:
This PR adds support for the [rocalution](https://github.com/ROCmSoftwarePlatform/rocALUTION) library from the [ROCm](https://en.wikipedia.org/wiki/ROCm) framework.

This solver might become obsolete quickly if rocsparse is supported, either via my branch or Kjetils hipification of the cuISTL branch. One advantage of rocalution over rocsparse is that rocalution can be easily compiled with g++, whereas rocsparse needs hipcc or some flags/configuration with g++ that I haven't been able to figure out.

The HAVE_ROCALUTION flag prevents rocalutionSolverBackend.cpp from being compiled, but we can also add the flag inside the files.

It allocates using `new[]` every linear solve, to use `SetDataPtrBCSR()`.
This function sets the `tmp_*` pointers to `nullptr` and frees those arrays after use. Valgrind shows no memory leaks here.

The rocalution files are not in `opencl`, because they don't include any opencl headers directly.

The rocalution solvers also feature an absolute tolerance, and a divergence tolerance, which Dune solvers don't. The hardcoded numbers are arbitrarily chosen.

It does not support using a jacobi matrix for the preconditioner, but the runtime parameter num-jacobi-blocks is only ignored, no error or warning is thrown.

The rocalution test will fail if rocalution is found on the machine to compile, but no supported devices are found during the test. This is fixed by initializing the rocalution library, and checking if the accelerator flag is set.
This results in the rocalution library being initialized twice, but that is not a problem.


CMakeLists_files.cmake
```diff
@@ -134,7 +134,10 @@ if(OPENCL_FOUND)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/opencl/openclSolverBackend.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/bda/opencl/openclWellContributions.cpp)
 endif()
-if(CUDA_FOUND OR OPENCL_FOUND OR HAVE_FPGA OR HAVE_AMGCL)
+if(HAVE_ROCALUTION)
```
akva2, line 137: this should be ROCALUTION_FOUND

Tongdongq: 5a5107b12
I left the `HAVE_FPGA` variable, because it will be removed in the near future.

Tongdongq:
I forgot to add some `HAVE_ROCALUTION` guards, but the test_rocalutionSolver still passed somehow.
I think combining them into some kind of `USE_BDA_BRIDGE` macro would be better.

akva2:
Please note my pr to your branch. Rocalution is now available in the ci mpi-enabled build. Use serial in trigger to test a build configuration without.

opm/simulators/linalg/bda/rocalutionSolverBackend.hpp
```diff
@@ -0,0 +1,108 @@
...
+    using Base::initialized;
+
+private:
+    std::vector<double> h_x; // store solution vector on host
+    int *tmp_rowpointers;    // store matrix on host, this pointer is given to and freed by rocalution
+    int *tmp_colindices;     // store matrix on host, this pointer is given to and freed by rocalution
+    double *tmp_nnzvalues;   // store matrix on host, this pointer is given to and freed by rocalution
+
+    std::unique_ptr<rocalution::BiCGStab<rocalution::LocalMatrix<double>, rocalution::LocalVector<double>, double> > roc_solver;
+    std::unique_ptr<rocalution::ILU<rocalution::LocalMatrix<double>, rocalution::LocalVector<double>, double> > roc_prec;
```
blattms, line 64: Please switch because of lifetime/references

opm/simulators/linalg/bda/rocalutionSolverBackend.cpp
```diff
@@ -0,0 +1,227 @@
...
+    this->N = Nb * block_size;
+    this->nnzb = matrix->nnzbs;
+    this->nnz = nnzb * block_size * block_size;
+
+    std::ostringstream out;
+    out << "Initializing rocalution, matrix size: " << Nb << " blockrows, nnzb: " << nnzb << "\n";
+    out << "Maxit: " << maxit << std::scientific << ", tolerance: " << tolerance << "\n";
+    OpmLog::info(out.str());
+    out.str("");
+    out.clear();
```
blattms, line 74: What is this for?

opm/simulators/linalg/bda/rocalutionSolverBackend.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+    /// Initialize sizes and allocate memory
+    /// \param[in] matrix     matrix A
+    void initialize(std::shared_ptr<BlockedMatrix> matrix);
+
+    /// Convert matrix to rocalution format
+    /// copy matrix to raw pointers, which are given to and freed by rocalution
+    /// \param[in] matrix     matrix A
+    void convert_matrix(std::shared_ptr<BlockedMatrix> matrix);
```
blattms, line 73: Is there a special reason to use shared_ptr here? Matrix is not stored inside of these methods and one could also use a referece to the object instead. This is probably nitpicking/personal taste.

tests/test_rocalutionSolver.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+#include <config.h>
+
+#define BOOST_TEST_MODULE OPM_test_rocalutionSolver
+#include <boost/test/unit_test.hpp>
+#include <boost/version.hpp>
+
+#include <dune/common/version.hh>
+
+#if DUNE_VERSION_NEWER(DUNE_ISTL, 2, 6) && \
```
blattms, line 29: At least the first part could be skipped as the minimal version always is 2.6

tests/test_rocalutionSolver.cpp
```diff
@@ -0,0 +1,180 @@
...
+                                                                               fpga_bitstream,
+                                                                               linear_solver_verbosity,
+                                                                               maxit,
+                                                                               tolerance,
+                                                                               platformID,
+                                                                               deviceID,
+                                                                               opencl_ilu_parallel,
+                                                                               linsolver);
+    } catch (const std::logic_error& error) {
+        BOOST_WARN_MESSAGE(true, error.what());
```
blattms, line 119: Mmh, I don't think we should continue in this case, should we? Maybe this rather calls for `BOOST_REQUIRE` or similar that stops the program?

blattms (commented):
Thanks a lot for the contribution. This looks promising. I added a few comments here and there. Hopefully, not to much nitpicking.

opm/simulators/linalg/bda/rocalutionSolverBackend.cpp
Tongdongq: There used to be more prints after that, this clears the stringstream contents.

tests/test_rocalutionSolver.cpp
Tongdongq: There are multiple tests that check for Dune 2.6 and Boost 1.48.
The [website](https://opm-project.org/?page_id=239) lists Boost 1.58 as minimum, so I'll remove those checks too.

tests/test_rocalutionSolver.cpp
Tongdongq: Good point. This is used in the other tests to catch 'device not found' exceptions. These occur when the machine has compile tools for GPUs but no actual GPU to run the test on.
For rocalution, the test checks with `rocalution_backend_descriptor->accelerator` if an accelerator is connected. If not, the test is skipped.
We could rethrow the error here, but since no error is ever handled in the catch, we could also just remove the whole `try` block.

opm/simulators/linalg/bda/rocalutionSolverBackend.hpp
Tongdongq: No reason, I simplified the parameters.

opm/simulators/linalg/bda/rocalutionSolverBackend.cpp
```diff
@@ -0,0 +1,227 @@
...
+}
+
+
+// copy result to host memory
+// caller must be sure that x is a valid array
+template <unsigned int block_size>
+void rocalutionSolverBackend<block_size>::get_result(double *x) {
+    Timer t;
+
+    std::copy(h_x.begin(), h_x.end(), x);
```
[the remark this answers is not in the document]
Tongdongq: Would you prefer `memcpy(x, h_x.data(), sizeof(double)*N)`?
I don't really understand your comment.

opm/simulators/linalg/bda/rocalutionSolverBackend.cpp
```diff
@@ -0,0 +1,227 @@
...
+    // roc_solver->GetSolverStatus() returns:
+    // 0, if no criteria has been reached yet
+    // 1, if absolute tolerance has been reached
+    // 2, if relative tolerance has been reached
+    // 3, if divergence tolerance has been reached
+    // 4, if maximum number of iteration has been reached
+
+    res.elapsed = t_solve.stop();
+    res.iterations = roc_solver->GetIterationCount();
+    res.reduction = roc_solver->GetCurrentResidual();
```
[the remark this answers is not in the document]
Tongdongq: It seems impossible to get the initial residual.
It is possible to record all residuals in a vector, but this can only be written to a file.
It is also possible to print the solver status via:
```c++
LOG_INFO("IterationControl RELATIVE criteria has been reached: "
                     << "res norm=" << std::abs(this->current_res_) << "; "
                     << "rel val=" << this->current_res_ / this->initial_residual_ << "; "
                     << "iter=" << this->iteration_);
```
[source file](https://github.com/ROCmSoftwarePlatform/rocALUTION/blob/4780aa22a6eda923e6f9f6ea2c72d478985d1834/src/solvers/iter_ctrl.cpp)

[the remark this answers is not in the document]
blattms: You could simply compute it as you should have everything: norm(b-Ax), where x is the initial guess.

opm/simulators/linalg/bda/rocalutionSolverBackend.hpp
blattms: Cool, thanks.

tests/test_rocalutionSolver.cpp
blattms: Ah, i see. The test should work without devices, I guess. You choose the proper solution.

opm/simulators/linalg/bda/rocalutionSolverBackend.cpp
```diff
@@ -0,0 +1,227 @@
...
+    for(int i = 0; i < Nb+1; ++i){
+        tmp_rowpointers[i] = matrix->rowPointers[i];
+    }
+    for(int i = 0; i < nnzb; ++i){
+        tmp_colindices[i] = matrix->colIndices[i];
+    }
+
+    // convert values inside block from row major to col major
+    // this is the same as transposing a block
+    // when compiling rocm from scratch, it is possible to choose the direction, making this transposing unnecessary
```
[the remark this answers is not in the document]
Tongdongq: Checking it now, but not actually tested for `BCSR_IND_BASE==1`.

opm/simulators/linalg/bda/rocalutionSolverBackend.cpp
```diff
@@ -0,0 +1,227 @@
...
+void rocalutionSolverBackend<block_size>::initialize(std::shared_ptr<BlockedMatrix> matrix) {
+    this->Nb = matrix->Nb;
+    this->N = Nb * block_size;
+    this->nnzb = matrix->nnzbs;
+    this->nnz = nnzb * block_size * block_size;
+
+    std::ostringstream out;
+    out << "Initializing rocalution, matrix size: " << Nb << " blockrows, nnzb: " << nnzb << "\n";
+    out << "Maxit: " << maxit << std::scientific << ", tolerance: " << tolerance << "\n";
+    OpmLog::info(out.str());
```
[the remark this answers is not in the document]
Tongdongq: Done, but it doesn't look like a big improvement. I didn't replace it for the other solvers.

opm/simulators/linalg/bda/rocalutionSolverBackend.cpp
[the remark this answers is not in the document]
Tongdongq: The initial guess is hardcoded to 0, so I just use `norm(b)` now.

blattms:
jenkins build this please

opm/simulators/linalg/bda/rocalutionSolverBackend.cpp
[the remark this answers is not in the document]
blattms: Seems to needed as rocalution otherwise has lifetime issues with the vectors.

blattms:

PR #4180, opened 2022-10-18
Convey More Dynamic State Values per Segment to Output Layer

bska:
In particular, report local condition segment flow velocities, segment holdup fractions, and segment phase viscosities.  To this end, expand the `SegmentState` to hold these additional quantities and update the values as part of [`updateWellStateFromPrimaryVariables()`](https://github.com/OPM/opm-simulators/blob/776180c08110d28790ff5cc783df2d19542ba3b9/opm/simulators/wells/MultisegmentWellEval.cpp#L1497).

We remark that the definition of the segment flow velocity is a little dubious and that we probably will have to revisit this definition (flow rate divided by cross-sectional area) at a later date.  Nevertheless, this PR is sufficiently ready to start testing segment-related RFT file output on more cases.

### DIFF
--- a/opm/simulators/wells/SegmentState.cpp
+++ b/opm/simulators/wells/SegmentState.cpp
@@ -16,44 +16,59 @@
   You should have received a copy of the GNU General Public License
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
+
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif // HAVE_CONFIG_H
 
-#include <algorithm>
-#include <stdexcept>
-
 #include <opm/simulators/wells/SegmentState.hpp>
 
-#include <opm/input/eclipse/Schedule/Well/WellConnections.hpp>
 #include <opm/input/eclipse/Schedule/MSW/WellSegments.hpp>
+
+#include <algorithm>
+#include <cstddef>
+#include <iterator>
+#include <stdexcept>
+#include <vector>
+
+namespace {
+
+std::vector<int> make_segment_number(const Opm::WellSegments& segments)
+{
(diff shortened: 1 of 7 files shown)


bska:
Note: This PR depends on, and includes, #4164 and must therefore not be merged into the master branch before that has been merged.  The PR also uses new interfaces introduced in OPM/opm-common#3179 which therefore becomes another upstream requirement.

bska:
The upstream prerequisites have been merged into the master branch.  I'm marking this as "ready for review".

bska:
jenkins build this please

opm/simulators/wells/SegmentState.cpp
```diff
@@ -16,45 +16,59 @@
...
+#include <vector>
 
 namespace {
-std::vector<int> make_segment_number( const WellSegments& segments ) {
-    std::vector<int> segment_number;
-    std::transform(segments.begin(), segments.end(), std::back_insert_iterator(segment_number), [](const Segment& segment) { return segment.segmentNumber(); });
-    return segment_number;
-}
 
+    std::vector<int> make_segment_number(const Opm::WellSegments& segments)
```
akva2, line 36: indent for namespace?

akva2 (approved):
Looks good

opm/simulators/wells/SegmentState.cpp
bska: > indent for namespace?

Nice catch.  I've pushed an update to remove that.

bska:
jenkins build this please

bska:
PR is approved and build is green.  I'll merge this into the master branch.

PR #4181, opened 2022-10-18
Some improvements for gaslift optimization stage 2

hakonhagland:
This PR provide some improvements to gaslift optimization in stage 2. First, it takes into account group limits when calculating incremental and decremental gradients (thanks to @totto82 for this contribution). Second, it takes into account LRAT group limits when removing surplus ALQ. We observe reduced oscillations in ALQ and oil production rate for [GASLIFT-03](https://github.com/OPM/opm-tests/blob/b76935322cd953adf914d2f88971add8f0defe7c/gaslift/GASLIFT-03.DATA) and also better agreement with the eclipse result.

Pasted image 20221019005856


hakonhagland:
jenkins build this please

opm/simulators/wells/GasLiftStage2.cpp
```diff
@@ -517,10 +543,9 @@ void
...
 optimizeGroup_(const Group &group)
 {
-    const auto &gl_group = this->glo_.group(group.name());
-    const auto &max_glift = gl_group.max_lift_gas();
-    const auto &max_total_gas = gl_group.max_total_gas();
-    if (group.has_control(Group::ProductionCMode::ORAT)
+    const auto max_glift = getGroupMaxALQ_(group);
+    const auto max_total_gas = getGroupMaxTotalGas_(group);
+    if (group.has_control(Group::ProductionCMode::ORAT) || group.has_control(Group::ProductionCMode::LRAT)
                        || max_glift || max_total_gas)
```
totto82, line 549: I think it would be better to check if the groups actual control is ORAT or LRAT not just that it has a control. 
i.e. `group_state.production_control(group.name()) == Group::ProductionCMode::ORAT`
I guess you then need to store const GroupState &group_state in GasLiftCommon.hpp

hakonhagland: Shouldn' t the logic in for example `removeSurplusALQ_()` also be done if the group  just has the control? That is, it does not need to be the actual control for the group. As an example, if the actual control of the group is WRAT but it also has a LRAT control, we should still remove surplus alq, see:

https://github.com/OPM/opm-simulators/blob/d31048d2987f6985ac1fef63ccd5527cfffa26aa/opm/simulators/wells/GasLiftStage2.cpp#L766

totto82: My understanding is that it only make sense to try to optimize the distribution of the gaslift if the amount of gaslift is constraint either by the maximum allowed gaslift or total gas or the group is under individual control i.e. so maybe a better logic will be to enter stage 2 if `group_state.production_control(group.name()) != Group::ProductionCMode::NONE && group_state.production_control(group.name()) != FLD`

hakonhagland:
jenkins build this please

opm/simulators/wells/GasLiftStage2.cpp
bska: > I have updated the PR to include this logic

The latest update leaves [`max_glift`](https://github.com/OPM/opm-simulators/blob/234750a563e3e353bf1ffd388f82e020841e84b8/opm/simulators/wells/GasLiftStage2.cpp#L547) and [`max_total_gas`](https://github.com/OPM/opm-simulators/blob/234750a563e3e353bf1ffd388f82e020841e84b8/opm/simulators/wells/GasLiftStage2.cpp#L548) [unused](https://ci.opm-project.org/job/opm-simulators-PR-builder/4159/gcc/new/).  Is that intentional because they will be used in a future update?  If not, you should probably remove the variables.

hakonhagland: > you should probably remove the variables

@bska Yes, I have updated the PR.

hakonhagland:
jenkins build this please

totto82:
I have tested this on a more complex model than GASLIFT3 and the models stagnates due to severe time-step chopping. I don't know why jet, but for now we should hold off merging this.

totto82:
I will close this, since it is replaced by #4214

PR #4182, opened 2022-10-19
implement dirichlet boundary conditions

totto82:
Depends on OPM/opm-common#3135 and OPM/opm-models#739


totto82:
I changed this to draft since it needs some modification to support default pressure and temperature as suggested by @OPMUSER i.e. use value form initialization if not provided.

totto82:
Note that this now depends on OPM/opm-common#3212

PR #4185, opened 2022-10-19
Rename option FlowNewtonMaxIterations to NewtonMaxIterations.

blattms:
Now it is consistent with other options (e.g. NewtonMinIterations.

Needs OPM/opm-models#741

### DIFF
--- a/opm/simulators/flow/NonlinearSolverEbos.hpp
+++ b/opm/simulators/flow/NonlinearSolverEbos.hpp
@@ -129,7 +129,7 @@ class WellState;
             static void registerParameters()
             {
                 EWOMS_REGISTER_PARAM(TypeTag, Scalar, NewtonMaxRelax, "The maximum relaxation factor of a Newton iteration used by flow");
-                EWOMS_REGISTER_PARAM(TypeTag, int, NewtonMaxIterations, "The maximum number of Newton iterations per time step used by flow");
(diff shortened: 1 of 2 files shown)


bska:
There are three regression test cases which explicitly explicitly set `FlowNewtonMaxIterations` in their `TEST_ARGS`.  You probably need to update those settings too

1. [`compareECLFiles_flow+SPE3CASE1`](https://github.com/OPM/opm-simulators/blob/3e680e41f56d39974403ec029be123589e10b138/compareECLFiles.cmake#L436)
2. [`compareECLFiles_flow+SPE5CASE1`](https://github.com/OPM/opm-simulators/blob/3e680e41f56d39974403ec029be123589e10b138/compareECLFiles.cmake#L497)
3. [`compareECLFiles_flow+SPE5CASE1_DYN`](https://github.com/OPM/opm-simulators/blob/3e680e41f56d39974403ec029be123589e10b138/compareECLFiles.cmake#L504)

bska (approved):
Thanks.  I'll merge this into the master branch now.

OPMUSER:
The option is currently hidden, that is it only printed with the CLI option `--help-all`, previously the old version was listed when` --help `was used. 

Can we revert to the same behavior for the release.

atgeirr:
> Can we revert to the same behavior for the release.

Yes we should unhide it.

PR #4186, opened 2022-10-19
Split up different tests in include files

alfbr:
Updating compareECLFiles.cmake has become a bit complicated. This is one suggestion in making it easier. Just a suggestion, please shoot if there are ways to do this better. In the mean time I intend to abuse this PR to tidy a bit in opm-tests (a couple of PRs there needs a data update).


alfbr:
jenkins build this please

alfbr:
jenkins build this please

alfbr:
jenkins build this opm-tests=571 update_data please

bska:
> I think this is a good idea, but we should not change any of the tests in this refactoring

We don't.

> I assume the actual usage here is just as a "dummy" PR to generate the opm-tests PR for another purpose.

True.  The data update is to enable merging OPM/opm-tests#571.

> we should make sure it goes through with a clean jenkins run, without any opm-tests updates.

We will.

bska:
jenkins build this please

bska:
I am in favour of this change and would like to see it go in.  Further refactoring can be made in follow-up PRs.

alfbr:
@blattms @akva2 great if one of you can have a quick look.

akva2:
i'd suggest you put the if() around the include statements instead of inside the include files. that way we save an indent all over those files

alfbr:
jenkins build this please

alfbr:
Looks like everything went through correctly.

PR #4193, opened 2022-10-21
- fix compilation of amgcl bindings in bda by not doing anything for …

hnil:
Compilation error for amgcl for blocksize 1.
This solves the it by disabling amgcl on blocksize 1.


bska:
jenkins build this please

bska (approved):
Thanks.  I'll merge this and then I'll make a follow-up PR to mark the other parameters as `[[maybe_unused]]` too.

PR #4194, opened 2022-10-21
- made heuristic scaling of trueimpes

hnil:
Make a normalization of the trueimpes weights. Maybe better exist but this fixes some issues.
It also highlight the need for the analytic weighs, which is in a seperate pullrequest


atgeirr:
jenkins build this please

atgeirr:
I have investigated the regression failure, and it is on a case that uses `--linear-solver=cpr` which defaults to the trueimpes variant. The aquifer rate error is different above threshold on 3 of the timesteps, differnces are in all cases less than 1, and the rates eventually reach ~200 and ~2000. I think this is fine, and will do a data update and merge eventually.

atgeirr:

atgeirr:
jenkins build this opm-tests=843 please

PR #4200, opened 2022-10-24
fix 2p case wvfpexp

totto82:


totto82:
jenkins build this please

opm/simulators/wells/VFPProperties.hpp
```diff
@@ -73,22 +73,20 @@ class VFPProperties {
...
     double getExplicitWFR(const int table_id, const size_t well_index) const {
         const auto& rates = well_state_.well(well_index).surface_rates;
-        assert(rates.size() == 3);
         const auto& pu = well_state_.phaseUsage();
-        const auto& aqua = rates[pu.phase_pos[BlackoilPhases::Aqua]];
-        const auto& liquid = rates[pu.phase_pos[BlackoilPhases::Liquid]];
-        const auto& vapour = rates[pu.phase_pos[BlackoilPhases::Vapour]];
+        const auto& aqua = pu.phase_used[BlackoilPhases::Aqua]? rates[pu.phase_pos[BlackoilPhases::Aqua]]:0.0;
+        const auto& liquid = pu.phase_used[BlackoilPhases::Liquid]? rates[pu.phase_pos[BlackoilPhases::Liquid]]:0.0;
+        const auto& vapour = pu.phase_used[BlackoilPhases::Vapour]? rates[pu.phase_pos[BlackoilPhases::Vapour]]:0.0;
```
bska, line 79: If you're changing this anyway now would be a good opportunity to remove the references and instead just write
```C++
const auto aqua = ...;
```
and so on.  [`SingleWellState::surface_rates`](https://github.com/OPM/opm-simulators/blob/cf6423bfce2fe7b482fee6a02ef5f5aa5329d382/opm/simulators/wells/SingleWellState.hpp#L62) is a vector of `double`s so there's no risk of copying large/expensive types here.  Removing the references will also remove an indirection in this case.  On the other hand, `pu` should remain a reference.

blattms:
Let's not split hairs, anyone can amend later. Merging.
@atgeirr Are you doing the backporting?

atgeirr:
> @atgeirr Are you doing the backporting?

Yes. I am keeping a list of PRs to backport, but I have not actually done the backporting yet.

PR #4201, opened 2022-10-24
Support Variable Length Connection Data in Distributed Wells

bska:
With the addition of tracer information in the `data::Rates` object we must relax the assumption that the object is POD.  Switch to using the `Point2PointCommunicator` protocol for gathering connection solutions on the well's owner before transferring well solution objects to the I/O rank.

While here, also supply user-defined message tags in P2P exchange calls during `CollectToIORank::collect()`.  That way we're not dependent on class-level static data to create new tags for each message.

Closes OPM/opm-common#2921.

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -27,10 +27,94 @@
 #include <opm/simulators/wells/ParallelWellInfo.hpp>
 
 #include <opm/simulators/utils/ParallelCommunication.hpp>
+#include <opm/grid/common/p2pcommunicator.hh>
+#include <opm/output/data/Wells.hpp>
 
 #include <algorithm>
 #include <cassert>
 #include <numeric>
+#include <set>
+#include <stdexcept>
+
+namespace {
+
+using P2PCommunicatorType = Dune::Point2PointCommunicator<Dune::SimpleMessageBuffer>;
+using MessageBufferType = P2PCommunicatorType::MessageBufferType;
+
+class PackUnpackXConn : public P2PCommunicatorType::DataHandleInterface
+{
+public:
+    using XConn = std::vector<Opm::data::Connection>;
+
+    explicit PackUnpackXConn(const bool   isOwner,
+                             const XConn& local,
+                             XConn&       global);
+
+    // Pack all data associated with link.
+    void pack(const int link, MessageBufferType& buffer);
+
+    // Unpack all data associated with link.
+    void unpack([[maybe_unused]] const int link,
+                MessageBufferType&         buffer);
+
+private:
+    const XConn& local_;
+    XConn& global_;
+};
+
+PackUnpackXConn::PackUnpackXConn(const bool   isOwner,
+                                 const XConn& local,
+                                 XConn&       global)
+    : local_ (local)
+    , global_(global)
+{
+    if (! isOwner) { return; }
+
+    this->global_.insert(this->global_.end(),
+                         this->local_.begin(),
+                         this->local_.end());
+}
+
+void PackUnpackXConn::pack(const int          link,
+                           MessageBufferType& buffer)
+{
+    // We should only get one link
+    if (link != 0) {
+        throw std::logic_error {
+            "link in pack() does not match expected value 0"
+        };
+    }
+
+    // Write all local connection results
+    {
+        const auto nconn = this->local_.size();
+        buffer.write(nconn);
+    }
+
+    for (const auto& conn : this->local_) {
+        conn.write(buffer);
+    }
+}
+
+void PackUnpackXConn::unpack([[maybe_unused]] const int link,
+                             MessageBufferType&         buffer)
+{
+    const auto nconn = [this, &buffer]()
+    {
+        auto nc = 0 * this->local_.size();
+        buffer.read(nc);
+
+        return nc;
+    }();
+
+    this->global_.reserve(this->global_.size() + nconn);
+
+    for (auto conn = 0*nconn; conn < nconn; ++conn) {
+        this->global_.emplace_back().read(buffer);
+    }
+}
+
+} // Anonymous namespace
 
 namespace Opm
 {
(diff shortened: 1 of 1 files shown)


bska:
This PR depends on upstream API change OPM/opm-grid#596 and must not be merged until that PR is available in the master branch.  I will keep this PR in "draft" state until such time as it is ready for review.

bska:
jenkins build this opm-grid=596 please

bska:
jenkins build this opm-grid=596 please

blattms:
Hopefully using native MPI calls for gathering is rather optimized. This change here now resorts to something really flexible (Point2PointCommunicator), that probably cannot be as performant.

My impression from the discussions on the other PRs is that the size of data is still fixed per well (we just do not happen to show that in the code). In addition the data to be gathered is simple pods due to that (masks, and maps don't need to gathered as they are the same). If that holds, then can't we accomplish gathering without using Point2PointCommunicator, that is set up every time we gather with a communication pattern that does not really change and is hence predetermined.

I know that this is tedious and there just occurred another though to me:
We do have these serializer interface that write stuff to a message buffer and broadcast. Couldn't we somehow reuse that or some of the code here. Each rank packs the stuff into a buffer and determines the size. Afterwards we use MPI_gather(v) as before to gather the message buffers size and the buffer on root and deserialize/unpack from the buffer. If that works then that would be a wonderful compromise between old and new.

Then in a later step we can actually make sure that sizes/masks are stored per well and not per connection and improve further.

bska:
> My impression from the discussions on the other PRs is that the size of data is still fixed per well 

Well, in this case it's fixed per connection but the number of connections differ per well, and typically also over time.

> Hopefully using native MPI calls for gathering is rather optimized. This change here now resorts to something really flexible (Point2PointCommunicator), that probably cannot be as performant.

This is true.  On the other hand the MPI call here cannot work as written because the mere presence of the `std::map` inside `data::Rates`, even if it's always empty, means the object is not a POD.  The easiest way to demonstrate that is to switch
```C++
res.insert({ wname, well });
```
([WellState.cpp:438](https://github.com/OPM/opm-simulators/blob/f373c2f90868718b479f2e68ab2c2fdaf4451fcd/opm/simulators/wells/WellState.cpp#L438)) into the nominally equivalent
```C++
res.emplace(wname, std::move(well));
```
and watch the `flow_distribute_z` regression test fall over.  On my machine this simple change generated all kinds of deadlocks in weird places, for instance in communicating well results to the IO rank.  Using `MPI_Gatherv` (or `comm.gatherv()`) is undefined behaviour right now so you may well see something else on your machine.

> We do have these serializer interface that write stuff to a message buffer and broadcast. Couldn't we somehow reuse that or some of the code here. Each rank packs the stuff into a buffer and determines the size. Afterwards we use MPI_gather(v) as before to gather the message buffers size and the buffer on root and deserialize/unpack from the buffer.

Sure.  It'll be something like this
```C++
auto packer = Mpi::Packer{ comm };
auto serializer = Serializer<Mpi::Packer>{ packer };
serializer.pack(connResults);
const auto& buff = serializer.buffer(); // New member function to gain read-only access to the internal buffer
const auto nconn = static_cast<int>(connResults.size());
const auto nchar = static_cast<int>(buff.size());
auto connSizes = std::vector<int>{};
auto charSizes = std::vector<int>{};
auto connDispl = std::vector<int>{};
auto charDispl = std::vector<int>{};
auto receiveBuffer = std::vector<char>{};

if (owner) {
    connSizes.resize(comm.size());
    charSizes.resize(comm.size());
}

comm.gather(&nconn, connSizes.data(), 1, 0);
comm.gather(&nchar, charSizes.data(), 1, 0);

if (owner) {
    connDispl.resize(comm.size() + 1, 0);
    std::partial_sum(connSize.begin(), connSizes.end(), connDispl.begin() + 1);

    charDispl.resize(comm.size() + 1, 0);
    std::partial_sum(charSize.begin(), charSizes.end(), charDispl.begin() + 1);
    receiveBuffer.resize(charDispl.back());
}

comm.gatherv(buff.data(), nchar, receiveBuffer.data(), charSizes.data(), charDispl.data(), 0);

if (owner) {
    finalConnResults.reserve(connDispl.back());
    for (auto rank = 0; rank < comm.size(); ++rank) {
        // New constructor for Serializer
        auto deserializer = Serializer { receiveBuff.begin() + charDispl[rank], receiveBuff.begin() + charDispl[rank + 1] };
        auto rankResults = std::vector<data::Connection>{};
        deserializer.unpack(rankResults);
        finalConnResults.insert(finalConnResults.end(), std::make_move_iterator(rankResults.begin()), std::make_move_iterator(rankResults.end()));
    }
}
```
I promise I'm not making this up to be difficult.  It really has to be this involved unless we do a lot of other work because our packer/serialization support is built to handle the case of one process has all information and must broadcast it to the world.  This use case, however, is the opposite.

bska:
Upstream PR OPM/opm-grid#599 was merged into the master branch.  Declaring this as "ready for review" now.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

blattms:
Well, it is probably me, but that looks nice. I just would pack the connSize at the beginning of the same buffer. and the you are only sending one charBuffer and the size of it. For the gatherv magic you can reuse [`allGatherv](https://github.com/OPM/opm-grid/blob/a617940397c2b898564acabdfe408ee822140fdb/opm/grid/common/CommunicationUtils.hpp#L45) to save code. On the other side you first unpack the connSize and then the connection data.

Given my history, I am probably missing something, though. 

I am not insisting though. But if we ever run into decreased performance for distributed wells, then this would be my first place to look.

Just let me know whether you will change or prefer the PR like this.

blattms:
To be honest: Reading this code I became a bit uneasy about my version even with varying degrees of freedom as the order of the connections will differ due to the numbering of the processors (connections on rank 0 first, the rank 1, ....)

I hope that is not a problem during output. Otherwise we should sort this by eclIndex of the connection after receiving.

bska:
> Well, it is probably me, but that looks nice.

Okay, then we disagree to some extent here.  I would really prefer not to manually piece out the individual components from a `vector<char>` if I don't have to.  I may revisit this later if I can get an appropriate set of operations into the serialiser API.  The [`gatherv()`](https://github.com/OPM/opm-grid/blob/a617940397c2b898564acabdfe408ee822140fdb/opm/grid/common/CommunicationUtils.hpp#L75) function&mdash;not `allGatherv()` since we don't need this information on every process&mdash;will simplify the code though and on more careful reflection we don't actually **need** the number of connection objects.  For preallocation (i.e., `finalConnResults.reserve()` on the "owner") we just need the total number of connection objects and that's essentially just `comm.sum(connResults.size())`.  On the other hand the deserialisation loop uses range insertion so preallocation is less critical.  Still useful, of course.

> the order of the connections will differ due to the numbering of the processors (connections on rank 0 first, the rank 1, ....)

This is true, and still true when using the point-to-point communication protocol.

> I hope that is not a problem during output.

It's not.  The output layer is robust in this sense and will, at some performance cost, search through the full array of connection results for each connection object and only use it if it's found.

> Otherwise we should sort this by eclIndex of the connection after receiving.

We could probably do that regardless in order to provide stronger guarantees.

> Just let me know whether you will change or prefer the PR like this.

I'm going to keep this as is for now.  Like I said, I'm not fully convinced that I have the best set of additional operations for the parallel serialiser.  It would be really convenient for this particular use case to have something like
```
finalConnResults = serializer.gatherv(connResults, root);
```
which could be specialised (through SFINEA or tagged dispatch or similar) on POD vs. non-POD but that may just be too specialised to add to the class.

blattms:
jenkins build this please

PR #4204, opened 2022-10-26
start regression testing for krnum cases

tskille:


alfbr:
jenkins build this please

bska:
Have you verified that the simulation results here agree with some other reference to a satisfactory degree?  If so, I think we should generate Flow's reference solutions and get these into the regression test suite.

tskille:
Yes this has been verified. The results have been compared with a commercial simulator.

bska:
>> Have you verified that the simulation results here agree with some other reference to a satisfactory degree?
>
> Yes this has been verified. The results have been compared with a commercial simulator.

Understood.  Thanks a lot for the confirmation.  In that case I'll generate reference solutions for these models and put them into the regression test suite.

bska:
Sorry, I hit the wrong button...

akva2:

akva2:
jenkins build this opm-tests=836 please

bska:
The reference solutions have been installed on the CI system.  I'm merging this now to bring the new cases into the test suite.

Issue #4202, opened 2022-10-25
2023-04-Pre: Norme Model Broken


OPMUSER:
I get this when attempting to run the Norne model.

```
Processing grid
Total number of active cells: 44431 / total pore volume: 673248728 RM3
Removed 496 cells with a pore volume of 100450 RM3 (0.015 %) due to MINPV/MINPVV
Program threw an exception: RangeError [get:/usr/include/dune/common/parametertree.hh:181]: Cannot parse value "true." for key ".OpenclIluParallel"RangeError [parse:/usr/include/dune/common/parametertree.hh:258]:  as a int
Failed: 3447
OPM Flow Process 3447 Has Failed
Process Complete (0) 
 

Queue Schedule Statistics
Number of Jobs Processed: 1
Number of Jobs Failed: 1
Number of Jobs Killed: 0
Number of Jobs Completed: 0

```

blattms:
I can't replicate. I notice that there is a dot after true that should not be there ("true." instead of "true"), but I am not 100% sure whether that is the cause. How do you run flow? Maybe share CMakeCache.txt to ease replication of the issue. This is master, right?

OPMUSER:
Not guilty. 

OPMRUN just runs the --help-all option to get all the command line parameters, so the error is in the code, not user input.

![image](https://user-images.githubusercontent.com/17958927/198189310-5150885f-5254-4097-bb70-a18e4a2380fe.png)
 and, just in case it was OPMRUN:

![image](https://user-images.githubusercontent.com/17958927/198189832-7f0d796c-e369-49d0-9638-87aaf9a060e7.png)

Notice the two Default entries. 

Which is why @Tongdongq found it as well.

@Tongdongq when you get a minute can you fix this, as it stops me running cases on the latest master

OPMUSER:
@blattms No.

Its is the syntax of the listing that is incorrect:

`Default: true. Default: true`

There should only be one `Default: true` and it has been like that for many years, and this is the first time for this type of error.

OPMUSER:
Oh, I see the fix has already been submitted, thank you @Tongdongq

PR #4206, opened 2022-10-27
Removed default declaration in wrong place

Tongdongq:
I didn't realize `registerParameters()` works like that.

Fixes https://github.com/OPM/opm-simulators/issues/4202.
Still testing it.

@blattms


Tongdongq:
This compiles on my end
```
    --num-jacobi-blocks=INTEGER                   Number of blocks to be created for the
                                                    Block-Jacobi preconditioner. Default: 0
    --opencl-ilu-parallel=BOOLEAN                 Parallelize ILU decomposition and
                                                    application on GPU. Default: true
    --opencl-platform-id=INTEGER                  Choose platform ID for openclSolver, use
                                                    'clinfo' to determine valid platform
                                                    IDs. Default: 0
```

blattms:
is the problem in the release, too? Or did I merge this after branch creation?

blattms:

Tongdongq:
This bug is not in the release/2022.10 branch, that still reorders explicitly.

PR #4212, opened 2022-10-27
Remove KRNUM from Unsupported Keyword List (Back-Port)

OPMUSER:
Remove KRNUM from Unsupported Keyword List, as it is now supported. This should be back-ported, to prevent a warning stating the keyword is not supported.


bska:
jenkins build this please

bska (approved):
Thanks.  I'll merge this into the master branch.  I've also tagged it with the 2022.10 release "milestone" to alert the release manager that this change should go into the release branch.

PR #4214, opened 2022-10-28
Fixes related to gaslift stage 2

totto82:
This PR extends #4181 by adding some necessary adjustments. The idea is that this PR should replace #4181.


totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

opm/simulators/wells/GasLiftSingleWellGeneric.cpp
```diff
@@ -247,7 +247,7 @@ GasLiftSingleWellGeneric::checkGroupTargetsViolated(const BasicRates& rates, con
             auto target_opt = this->group_info_.getTarget(rate_type, group_name);
             if (target_opt) {
                 auto delta_rate = new_rates[rate_type] - rates[rate_type];
-                auto new_group_rate = this->group_info_.getRate(rate_type, group_name) + efficiency * delta_rate;
+                auto new_group_rate = this->group_info_.getPotential(rate_type, group_name) + efficiency * delta_rate;
```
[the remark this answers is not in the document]
totto82: The potentials are only used to reduce alq before the update, so we don't need to update them. I can add a comment.

opm/simulators/wells/GasLiftStage2.cpp
```diff
@@ -1137,6 +1138,24 @@ checkOilTarget()
     return false;
 }
 
+bool
+GasLiftStage2::SurplusState::
+checkWaterTarget(double eco_grad)
+{
+    if (this->group.has_control(Group::ProductionCMode::WRAT)) {
+        if (this->water_target < (this->water_rate - eco_grad)  ) {
```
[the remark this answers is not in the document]
totto82: We would like to keep the gasrate as little as possible but still be able to produce the group target. Without the eco_grad the group target will drop slightly below the target.

opm/simulators/wells/GasLiftStage2.cpp
```diff
@@ -785,7 +785,7 @@ removeSurplusALQ_(const Group &group,
             state.updateRates(well_name);
             state.addOrRemoveALQincrement( this->dec_grads_, well_name, /*add=*/false);
             recalculateGradientAndUpdateData_(
-                        dec_grad_itr, /*increase=*/false, dec_grads, inc_grads);
+                        dec_grad_itr, group, /*increase=*/false, dec_grads, inc_grads);
```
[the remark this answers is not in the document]
totto82: Good point. I will rename the variable and/or add a comment.

totto82:
jenkins build this please

opm/simulators/wells/GasLiftStage2.cpp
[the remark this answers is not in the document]
hakonhagland: Hmm.. the `eco_grad` says how much the oil rate change for given change in alq, right? How is this related to the water rate?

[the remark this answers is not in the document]
totto82: It isn't. Now I see your point. So it only make sense to substract it in checkOilRate and checkLiquidRate. I will update and test.

totto82:
jenkins build this please

opm/simulators/wells/GasLiftStage2.cpp
```diff
@@ -1075,13 +963,35 @@ checkGasTarget()
...
 bool
 GasLiftStage2::SurplusState::
-checkOilTarget()
+checkOilTarget(double eco_grad)
 {
     if (this->group.has_control(Group::ProductionCMode::ORAT)) {
-        if (this->oil_target < this->oil_rate  ) {
+        // the eco_grad is subtracted from the rate to make sure the
+        // group still can produce its target
+        if (this->oil_target < (this->oil_rate - eco_grad)  ) {
```
hakonhagland, line 994: Thanks for the update! One final question: Should `eco_grad` be multiplied by the alq increment, to convert it into an oil rate? Something like: `auto delta_oil = eco_grad * this->parent.glo_.gaslift_increment()` and then the if statement: `if (this->oil_target < (this->oil_rate - delta_oil)  ) {`

totto82: Thanks for the comment. I though the eco_grad already was the delta_rate, but your're of-course right. Is it possible to get the delta_water etc from the gradient?

opm/simulators/wells/GasLiftStage2.cpp
```diff
@@ -1109,20 +1039,22 @@ updateRates(const std::string &well_name)
         if (this->parent.well_state_.wellIsOwned(well.indexOfWell(), well_name)) {
             const auto &well_ecl = well.wellEcl();
             double factor = well_ecl.getEfficiencyFactor();
-            auto& [delta_oil, delta_gas, delta_alq] = delta;
-                    delta_oil = factor * (gi.new_oil_rate - state.oilRate());
-                    delta_gas = factor * (gi.new_gas_rate - state.gasRate());
-                    delta_alq = factor * (gi.alq - state.alq());
+            auto& [delta_oil, delta_gas, delta_water, delta_alq] = delta;
+            delta_oil = factor * (gi.new_oil_rate - state.oilRate());
```
hakonhagland, line 1043: This might not be related to this PR, but I have a question to the above line: This `delta_oil` in the previous line is correct for the first call from removeSurplusALQ_(), see line 631. But if `updateRates()` is called a second time within the loop starting at line 610, then `delta_oil` should be calculated according to the previous oil rate and not `state.oilRate()`? Since `state.oilRate()` is fixed within the loop starting at line 610 (it comes from stage 1 and is not changed within stage 2). So we should rater use something like `delta_oil = factor * (gi.new_oil_rate - this->prev_oil_rate[well_name])` where we need to define a new variable `prev_oil_rate` ?

hakonhagland: I think I missed that we update the `GasLiftWellState` at line 117. So then this probably works correctly after all.

opm/simulators/wells/GasLiftStage2.cpp
hakonhagland: If you look at lines 1018-1021 in `updateRates()`, we compute `delta_oil`, `delta_water`, ... Maybe we could save these values or recompute them here? Instead of using the `eco_grad` to compute them.

totto82: I will try to split the method in two. First to compute the deltas = computeDeltas(..) , then later apply them in updateRates( deltas ) if remove = true.

totto82:
jenkins build this please

opm/simulators/wells/GasLiftStage2.cpp
```diff
@@ -627,8 +628,10 @@ removeSurplusALQ_(const Group &group,
         const auto well_name = dec_grad_itr->first;
         auto eco_grad = dec_grad_itr->second;
         bool remove = false;
-        if (state.checkOilTarget(eco_grad) || state.checkGasTarget()
-              || state.checkLiquidTarget(eco_grad) || state.checkWaterTarget() || state.checkALQlimit())
+        const auto delta = state.computeDelta(well_name);
+        if (state.checkOilTarget(delta[0]) || state.checkGasTarget(delta[1])
+              || state.checkLiquidTarget(delta[0] + delta[2]) || state.checkWaterTarget(delta[2])
```
hakonhagland, line 633: The last update looks good! A minor thing: In the above line, could we destructure the array `delta` into references like on line 1058: 
```
        const auto delta = state.computeDelta(well_name);
        auto& [delta_oil, delta_gas, delta_water, delta_alq] = delta;
        if (state.checkOilTarget(delta_oil) || state.checkGasTarget(delta_gas)
              || state.checkLiquidTarget(delta_oil + delta_gas) || state.checkWaterTarget(delta_water)
              || state.checkALQlimit(delta_alq, delta_gas))
        {
```
This could improve the readability of the last line.

totto82:
jenkins build this please

totto82:
@hakonhagland Can this be merged now?

hakonhagland:
> Can this be merged now?

@totto82 Yes it looks good!

PR #4215, opened 2022-10-28
changed: move test definitions to json files

akva2:
This is a suggestion for simplification of test definitions.

We here move the definitions of the tests to json files, thus splitting code and data.

There is a single json file for each test in this draft. That is mainly because it was the easiest while scripting the conversion to json files. The idea is that tests can be grouped in the json files. The code supports any number of array entries for each json file.This should make it easier to find a particular test. My suggestion would be that the json files reflect the directory structure in the opm-tests repo.

As an additional bonus, this allows me to automate the list of data needed in the data update script, making addition of new tests simpler.


akva2:
jenkins build this please

bska:
I applaud any effort to make adding/changing tests more automatic.  While I haven't looked at the details here, it does appear that neither of the new JSON files end in a newline character.  Is that intentional/required?

akva2:
that's just python being stupid, i can remedy that.

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

atgeirr:
I think this looks great! But there should be code comments/documentation that explains (to a developer) how it all fits together. This is also missing in the current code, it would be another important improvement to have docs! Especially:
- I want to add a test of type X, what must I do?
- How does the chain of things work, how will a test go from a JSON definition to a "test cmake knows about"?
- What is the role of each file and the code in it?

akva2:
sure, i can document it whenever we settle on a format, but any effort before that is quite the waste. right now it's a bare minimal dump of the parameters to json files, which is rather suboptimal since it repeats a lot of information, maybe not necessary. see the latest commit for an example.

atgeirr:
> sure, i can document it whenever we settle on a format, but any effort before that is quite the waste. right now it's a bare minimal dump of the parameters to json files, which is rather suboptimal since it repeats a lot of information, maybe not necessary. see the latest commit for an example.

Fine with me.

akva2:
one example of simplification (compare to latest commit)
```json
{
    "simulator" : "flow",
    "abs_tol" : 0.02,
    "rel_tol" : 1e-05,
    "dir" : "krnum",
    "test_cases" :
       [
        {
            "casename" : "krnum_02x",
            "filename" : "KRNUM-02X"
        },
        {
            "casename": "krnum_02y",
            "filename": "KRNUM-02Y"
        },
        {
            "casename": "krnum_02z",
            "filename": "KRNUM-02Z"
        },
        {
            "casename": "krnum_03x",
            "filename": "KRNUM-03X"
        },
        {
            "casename": "krnum_03y",
            "filename": "KRNUM-03Y"
        },
        {
            "casename": "krnum_03z",
            "filename": "KRNUM-03Z"
        }
    ]
}
```

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

PR #4216, opened 2022-10-28
fixed: some duplicate test case names

akva2:
while it causes no harm, it's still wrong


akva2:
jenkins build this please

bska (approved):
Thanks a lot for identifying this problem.  I'll merge into master.

PR #4217, opened 2022-10-31
Debian fixes (manpage and missing header include)

blattms:
Without the include compilation fails in experimental


blattms:
jenkins build this please

PR #4218, opened 2022-10-31
changed: use return value

akva2:
no reason to pass a reference here, just use return


akva2:
jenkins build this please

atgeirr (approved):

PR #4220, opened 2022-10-31
Allow for test definitions through json-files in opm-tests

akva2:
This is an alternative to / evolution of https://github.com/OPM/opm-simulators/pull/4215

It would be more messy to keep it in the same PR so I just opened a second one.

Here the test json definitions are actually stored in opm-tests. It gives some flexibility (multiple/alternative test repos comes to mind), and in general keeps the definitions closer to the source.

If we go with this alternative we need to rethink how to do update_data since the currently not supported situation of having an update in opm-tests that you want to generate data for, would now be the case every time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 
 project(opm-simulators C CXX)
 
-cmake_minimum_required (VERSION 3.10)
+cmake_minimum_required (VERSION 3.19)
 
 
 option(SIBLING_SEARCH "Search for other modules in sibling directories?" ON)
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -237,6 +237,51 @@ endfunction()
 
 ###########################################################################
 
+### Helper functions ###
+
+macro(parse_json_common_params)
+  string(JSON test_cases GET ${JSON_STRING} "test_cases")
+  string(JSON simulator GET ${JSON_STRING} "simulator")
+  string(JSON abs_tol GET ${JSON_STRING} "abs_tol")
+  string(JSON rel_tol GET ${JSON_STRING} "rel_tol")
+  string(JSON dir ERROR_VARIABLE err GET ${JSON_STRING} dir)
+  if (err)
+    # Common case - use directory hosting json file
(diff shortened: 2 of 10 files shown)


OPMUSER:
I like this as it it a cleaner implementation. Since we have both regression tests and integration tests (release tests), then perhaps we should have an additional entry that accounts for both types of tests:

```
    "test_cases" :
       [
        {
            "casename" : "krnum_02x",
            "filename" : "KRNUM-02X",
           "regression: true
        },
        {
            "casename": "krnum_02y",
            "filename": "KRNUM-02Y",
           "integration: true
        },
        {
            "casename": "krnum_02z",
            "filename": "KRNUM-02Z"
           "regression: true
           "integration: true
        }
    ]
```

I also think it would be a good idea to re-organize the directory structure for the individual tests, something along the lines;

``` 
                   testdirc                                                                     
                   |  Contains just DATA files and main Markdown file.
                      |---include                                                                  
                      |---results      
                           contains current / latest results here, with the following sub-directories                                                 
                             |---ecl
                                 contains commercial simulator results.
                             |---ref     
                                 contains OPM Flow reference results.                                                           
                      |--- plots       
                           contains model plots and case specific Markdown files here, with the following sub-directories.                                                   
                             |---ecl
                                 contains commercial simulator versus current plots for all cases used in the case Markdown files.
                             |--ref     
                                 contains OPM Flow reference versus current plots for all cases used in the case Markdown files.                                                    
                      |--- tests
                           contains JSON test definitions.                                                              
``` 
The idea behind this is that we automatically generate the plots via Jenkins, so that the plots are placed in the above structure. Take a look at the GASLIFT and GRUPCNTL test cases for examples. Notice that ECL cases all have DATA filenames ending in `-ECL` and the reference cases ending `-REF`. This is needed to distinguish between the three sets of runs (current, E100, and OPM Flow Reference) in the plots.

I can generate the plots semi-automatically via a Python script, so the intend is to make this fully automatic, in which a modified Python script is called by Jenkins and generates all the plots. Then it should be easy to check the results when Jenkins complains.

Comments.

PR #4225, opened 2022-11-01
removing unused number_segments from MultisegmentWell

GitPaean:
It does not looks like used anywhere and it caused failed debug running by using it.


GitPaean:
jenkins build this please

GitPaean:
Maybe a separate number might be needed if some segments are closed/shut, but in the form of the current master, the variable is not used. So removing it anyway for now.

bska (approved):

PR #4226, opened 2022-11-01
removed assert in cprw for strictly positive diagonal well coupling

hnil:
Remove assert. Not sure why it maybe triggered.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -814,14 +814,18 @@ namespace Opm
                     diag_ell -= matel;
                 }
             }
-        
+
+#define EXTRA_DEBUG_MSW 0
+#if EXTRA_DEBUG_MSW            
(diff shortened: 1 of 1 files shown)


atgeirr:
Removing the assert is fine, adding the `std::cout << ...` is not. If you want to add output it is unfortunately a bit complicated here. It must use the OpmLog system, and behave correctly in parallel, which for these well model call sites usually involves:
 - Create a DeferredLogger higher on the call stack, where all processes are involved. This will typically be in the BlackoilWellModel class.
 - Pass that object to the function where you want the output.
 - Use its logging function to do any logging you want.
 - Back up where you created the DeferredLogger, after the call returns, do a gatherDeferredLogger() etc.

hnil:
I can remove the warning to avoid adding the deferred logger. However it indicate something strange with well reservoir coupling i.e. total sum of effective transmissbilities maybe negative.

atgeirr:
jenkins build this please

PR #4235, opened 2022-11-03
fix group switch logging

totto82:
We want to keep the last group control for the terminal output i.e. use insert_or_assign instead of just insert.


totto82:
Maybe too late for the release, but since it only affects output it may still be in time?

totto82:
jenkins build this please

bska (approved):
This looks good to me, and I'll merge into master.  It's not immediately clear to me why we store strings instead of `enum` values in the `switched_*` maps, but that's a topic for another day.

bska:
I'll defer the decision on whether to include this in the release to the release manager.

PR #4237, opened 2022-11-04
changed: pass the actual update value to relaxationFactorRate

akva2:
this to avoid ussage of BVectorWell in interface. in preparation for moving BVectorWell out of the well hierarchy.


bska:
@GitPaean : Would you mind having a look at this, please?  Is there a reason for using the `BVectorWell` object?

bska:
> I have no objection to the PR.

Excellent.  Then I'll merge this into the master branch.

PR #4239, opened 2022-11-04
we do not copy SegmentState if the segment number changes

GitPaean:


GitPaean:
jenkins build this please

GitPaean:
copying will overwrite the properly initialized size.

bska (commented):
Is there any benefit at all to using the previous solution's segment state in the next episode?  If not, maybe we can just skip this block entirely and just use the default segment state values?

GitPaean:
> Is there any benefit at all to using the previous solution's segment state in the next episode? If not, maybe we can just skip this block entirely and just use the default segment state values?

In theory, it might save some solving time since it can be a good initial guess. And if time step is very small, then the storage effects of the wellbore might matter, while it might not matter for most of the reservoir simulation. 

But in practice, not sure how much it matters, which needs to be figured out through testing.

GitPaean:
I change a little bit my comment above. It might be important for some situations, under which we might struggle to have a good initial guess, for example, we might need some sensible composition information in the wellbore to do the VFP interpolation/search. Unless we can do a proper initialization (currently, it is very often only very trivial initialization is done), reusing the old SegmentState might be important.

bska:
> It might be important for some situations, under which we might struggle to have a good initial guess, for example, we might need some sensible composition information in the wellbore to do the VFP interpolation/search.

Right, yes that makes sense.  In that case we might need to revise this solution at a later time.  In the meantime I'll merge this into the master branch so that we at least fix the most obvious problem.  Thanks a lot!

GitPaean:
Thanks for merging. It fixes a real problem in a real case. That case is probably the first case we have that modifies the number of the segments of a multisegment well (doubles the number of the segments with SCHEDULE) during the simulation. Because of the bug fixed in this PR, the segment number in the SegmentState stays unchanged. Without bounding checking when visiting std::vector, it caused very confusing simulation behavoir until it crashes.

PR #4242, opened 2022-11-04
[bugfix] recover support for selecting GMRes as solver.

### LINKED ISSUE #4240: [2022.10-rc1] --use-gmres command line not honored
This parameter is read but neglected by the flexible solver interface and BiCGSTABSolver will be used anyway. For GMRES one needs use a json parameter file.

blattms:
Previously, passing --use-gmres=true had now effect. Probably was lost when introducing FlexibleSolver. With this change the FlexibleSolver will now use GMRes if the option is true.

Closes #4240


blattms:
jenkins build this please

bska (approved):

PR #4259, opened 2022-11-09
[bugfix] Do not unconditionally dereference std::optional.

blattms:
In WellGroupHelpers we unconditional referenced the region member of GPMaint. Unfortunately, that is a std:optional and doing this can create quite some havoc.

Now we stop processing if it is unset. Not sure whether this is what we want, though.

Drive-by fix. Unfortunately, not the issue that I am after.

### DIFF
--- a/opm/simulators/wells/WellGroupHelpers.cpp
+++ b/opm/simulators/wells/WellGroupHelpers.cpp
@@ -682,7 +682,11 @@ namespace WellGroupHelpers
         if (!gpm)
             return;
 
-        const auto [name, number] = *gpm->region();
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska (commented):
Sorry about the delayed response.  This is a good catch.  Function `region()` returning `nullopt` appears to be the case when the fluid-in-place region corresponds to the whole field (item 4 is zero).  We may need more logic in place to handle that case.

blattms:
Not sure whether that is approval or not. I'll just be optimistic and merge :grimacing:

Issue #4244, opened 2022-11-05
THERMAL mode run unstable solver performance 


EdmundStephens:
The attached project is a single phase WATER model in THERMAL mode and identical input I ran in E100 TEMP mode. 

[Bunter3E.zip](https://github.com/OPM/opm-simulators/files/9944099/Bunter3E.zip)
[Bunter3F.zip](https://github.com/OPM/opm-simulators/files/9944100/Bunter3F.zip)

In Flow the model shows very poor solver performance with repeated linear convergence problems and small time step sizes

```
Number of MPI processes:        30
Threads per MPI process:         2
Total time (seconds):         3837.62 
Solver time (seconds):        3828.22 
Assembly time (seconds):     2064.06 (Failed: 1497.6; 72.6%)
Well assembly (seconds):    305.76 (Failed: 227.7; 74.5%)
Linear solve time (seconds):  911.95 (Failed: 749.6; 82.2%)
Linear setup (seconds):     209.07 (Failed: 162.5; 77.7%)
Update time (seconds):        632.38 (Failed: 476.1; 75.3%)
Pre/post step (seconds):      143.50 (Failed: 7.9; 5.5%)
Output write time (seconds):   43.70
Overall Linearizations:     170041    (Failed: 123501; 72.6%)
Overall Newton Iterations:  158335    (Failed: 123501; 78.0%)
Overall Linear Iterations: 1447907    (Failed: 1222625; 84.4%)

Error summary:
Warnings          953
Info              47241
Errors            0
Bugs              0
Debug             0
Problems          5881
```

In E100 TEMP the run proceeds with good performance and timesteps. 

```
Error summary 
Comments               2
Warnings               1
Problems               0
Errors                 0
Bugs                   0
Final cpu      57.29 elapsed      57.29
Total number of time steps forced to be accepted          0
```

It would seem there is something going awry in the solver. But on the plus side, the results while not identical (as noted in your documentation) are comparable:

![image](https://user-images.githubusercontent.com/47630958/200126782-0efb5259-cf96-4077-90da-0b76cdee53c3.png)

totto82:
Thanks for reporting the issue and the test cases. I tried to run your cases and are able to reproduce the issue. There seems to be an issue with the injector. I will look into it.

totto82:
With #4262 the convergence is improved. I still see some oscilation in the temperature. 
Note also #4263. For now you will need to set WTEMP after WCONINJE in the input deck

PR #4262, opened 2022-11-10
remove derivative in thermal flux for injectors

totto82:
The well flux for injection perforations depends on the mobility in the cell. ie. we dont use the upstream value. this can lead to problems if the mobility depends strongly on temperature. The proposed change fixes the issue reported in #4244 

An alternative fix is to set the mobility of the injection perforation to Scalar but that will also affect non-isothermal cases. Initial testing indicates that this may have minor negative impact on convergence. Any comments?

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -681,10 +681,13 @@ namespace Opm
                     fs.setDensity(phaseIdx, rho);
                     const auto& h = FluidSystem::enthalpy(fs, paramCache, phaseIdx);
                     fs.setEnthalpy(phaseIdx, h);
+                    cq_r_thermal *= this->extendEval(fs.enthalpy(phaseIdx)) * this->extendEval(fs.density(phaseIdx));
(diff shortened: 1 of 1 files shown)


GitPaean:
sorry for not looking at this earlier due to absence from office recently.  At the same time, it is not that trivial for me to give a immediate comment. Will respond later this week when I think through it.

GitPaean:
sorry for the delay. Let me check this PR now.

GitPaean:
I understand that this will help the convergence by removing the undesirable derivatives. To me, it is okay to go in.

bska:
Ehm, this [appears to break](https://ci.opm-project.org/job/opm-simulators/1457/testReport/) the regression tests.  It looks like it was merged without a build check.

GitPaean:
> Ehm, this [appears to break](https://ci.opm-project.org/job/opm-simulators/1457/testReport/) the regression tests. It looks like it was merged without a build check.

Sorry. I did not notice that.

GitPaean:
@totto82 , please re-create the PR and I think we need to check and update the reference results.  Sorry for the trouble.

PR #4264, opened 2022-11-10
WIP CO2STORE + GAS + WATER

totto82:
Depends on OPM/opm-common#3217 and OPM/opm-material#541


totto82:
A better solution is provided in #4292

PR #4266, opened 2022-11-11
Alugrid Sfc Ordering

ElyesAhmed:
I fixed the ordering issue for flow_alugrid, so no need to set DISABLE_ALUGRID_SFC_ORDERING for flow_alugrid. @akva2 How much work is needed to let jenkins test flow_alugrid in the future?I think we will need that from now..

### DIFF
--- a/ebos/eclalugridvanguard.hh
+++ b/ebos/eclalugridvanguard.hh
@@ -314,7 +314,6 @@ protected:
 
         factory_ = std::make_unique<Factory>();
         grid_ = factory_->convert(*equilGrid_, cartesianCellId_, ordering_);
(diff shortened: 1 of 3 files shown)


akva2:
Last I looked it was not straight forwars to find a dune-alugrid that works with dune 2.7. With such a version it won't take much to add it.

blattms:
Doesn't branch releases/2.7 work?

ElyesAhmed:
> Last I looked it was not straight forwars to find a dune-alugrid that works with dune 2.7. With such a version it won't take much to add it.

But there is Dune AluGrid releases/2.7

PR #4238, opened 2022-11-04
added: StandardWellEquations and StandardWellAssemble

akva2:
the former is a container class for the equation system, as well as methods for applying to vectors.
the latter handles the assembly of the equation system.

in particular the equation system is now private to these two classes (the assembler can access equations members
through a friend declaration).

Sits on top of https://github.com/OPM/opm-simulators/pull/4236 and https://github.com/OPM/opm-simulators/pull/4237

PR #4268, opened 2022-11-14
added: StandardWellEquations

akva2:
This is a container class for the equation system, as well as methods for applying to vectors.

- better code organization
- less instances. instead of one instance per index instance there is now one instance per block size (1..6).

This is part of a process to refactor the equation system and assembly of this from in standard well, as drafted in https://github.com/OPM/opm-simulators/pull/4238. All data members are still public, but this will be remedied later.

Sits on top of https://github.com/OPM/opm-simulators/pull/4237


akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

bska (approved):
Very good.  I like this a lot and I appreciate the direction in which we're going.  I'm going to merge this into the master branch now.

PR #4269, opened 2022-11-14
ISTLSolverEbos: instance for ALUGrid

akva2:
To enable https://github.com/OPM/opm-common/pull/3219

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -499,23 +499,25 @@ opm_add_test(flow_distribute_z
   ${FLOW_TGTS}
   $<TARGET_OBJECTS:moduleVersion>
   )
-if (NOT BUILD_FLOW_ALU_GRID)
-  set(FLOW_ALUGRID_ONLY_DEFAULT_ENABLE_IF "FALSE")
-else()
-  set(FLOW_ALUGRID_ONLY_DEFAULT_ENABLE_IF "TRUE")
-endif()
 
-# the production oriented general-purpose ECL simulator
-opm_add_test(flow_alugrid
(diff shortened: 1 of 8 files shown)


akva2:

blattms:
I have the feeling, that this won't work if ALUGrid is not there, but a build with it is requested. Problem is that it will fail only when compiling and not already when configuring.

akva2:
HAVE flag is for availability, not what user requested. the USE flag is properly handled in https://github.com/OPM/opm-simulators/blob/master/ebos/eclproblem.hh#L31

akva2:
ah now i get what you mean. maybe we should just make eclproblem #error ?

akva2:
i opted to disable in the buildsystem if alugrid is not found. this is the normal way to handle these things.

akva2:
jenkins build this please

bska:
@blattms: Unless you have any reservations about the build system portion of this PR I think we should merge it.  Right now it's blocking all other PRs from progressing because it's impossible to get a clean build.

blattms:
Sorry, for the delay, but the information about blocking something was not clear to me from the available channels (such as PRs, lists, chats,...) Not really sure whether it was this PR that was blocking, but I simply merged all related ones and hope it is resolved now.

blattms:
@akva2 Is there a bot deleting your branches after merging in such a timely manner, or is there an option like in gitlab for requesting this that I fail to see?

akva2:
nah, i'm just very dilligent and if i am at a device at that time i immediately go and delete the branch because if not i'll surely forget about it..

PR #4273, opened 2022-11-17
Add test for cusparseSolver

Tongdongq:
This PR adds a test where `cusparseSolver` is used with multiple jacobi blocks.
And combines some `HAVE_XXX` variables into `COMPILE_BDA_BRIDGE` at some places.
The `HAVE_XXX` variables are still used in other places.


tests/test_cusparseSolver.cpp
```diff
@@ -129,24 +117,70 @@ testCusparseSolver(const boost::property_tree::ptree& prm, Matrix<bz>& matrix, V
     }
 }
 
+template <int bz>
+Dune::BlockVector<Dune::FieldVector<double, bz>>
+testCusparseSolver(std::unique_ptr<Opm::BdaBridge<Matrix<bz>, Vector<bz>, bz> >& bridge, Matrix<bz>& matrix, Vector<bz>& rhs)
```
bska, line 122: Just a general comment: Using `std::unique_ptr<T>&` as a parameter type **typically** announces that the function wants to participate in the lifetime management of the contained object and is going to change the `unique_ptr<>` object itself, not that we just want to use the underlying object.

Unless I'm missing something, this function does not appear to alter the actual `unique_ptr`, so in this case I'd probably implement the function as follows
```C++
template <int bz>
Dune::BlockVector<Dune::FieldVector<double, bz>>
testCusparseSolver(Opm::BdaBridge<...>& bridge, const Matrix<bz>& matrix, const Vector<bz>& rhs)
{
    // ...
    bridge.solve_system(...);
    bridge.get_result(x);

    return x;
}
```
and call it as
```
testCusparseSolver(*bridge, matrix, rhs);
```
instead.

Tongdongq: Good point, I'll change it

Tongdongq:
I put back the `roc_variable.release()` in the destructor. Reversing the declaration order in the header did not actually fix the segfaults.
It failed the test 3/20 times, and spe1 failed 11/20 times.
Now it doesn't fail anymore.

blattms:

Tongdongq:
The `test_cusparseSolver` fails when the test is compiled (because a CUDA compiler is found), but no device is found when running the test.
I can replicate, trying to fix it now.

Tongdongq:

bska:

blattms:
Thanks a lot for the fix.

PR #4275, opened 2022-11-17
FLOWS and FLORES support

daavid00:
Current implementation for compatibility with ResInsight for these two keywords. NNC flows are also written out. For now as a draft to check all tests.

FLOWS_FLORES

### DIFF
--- a/ebos/collecttoiorank.cc
+++ b/ebos/collecttoiorank.cc
@@ -773,7 +773,7 @@ public:
     PackUnpackInterRegFlows(const EclInterRegFlowMap& localInterRegFlows,
                             EclInterRegFlowMap&       globalInterRegFlows,
                             const bool                isIORank)
-        : localInterRegFlows_ (localInterRegFlows)
+        : localInterRegFlows_(localInterRegFlows)
         , globalInterRegFlows_(globalInterRegFlows)
     {
         if (! isIORank) { return; }
@@ -803,6 +803,68 @@ public:
     // unpack all data associated with link
     void unpack(int /*link*/, MessageBufferType& buffer)
     { this->globalInterRegFlows_.read(buffer); }
+};
+
+class PackUnpackFlows : public P2PCommunicatorType::DataHandleInterface
+{
+    const std::array<std::pair<std::string, std::pair<std::vector<int>, std::vector<double>>>, 3>& localFlows_;
+    std::array<std::pair<std::string, std::pair<std::vector<int>, std::vector<double>>>, 3>& globalFlows_;
+
+public:
+    PackUnpackFlows(const std::array<std::pair<std::string, std::pair<std::vector<int>, std::vector<double>>>, 3> & localFlows,
+                     std::array<std::pair<std::string, std::pair<std::vector<int>, std::vector<double>>>, 3>& globalFlows,
+                     const bool isIORank)
+        : localFlows_(localFlows)
+        , globalFlows_(globalFlows)
+    {
+        if (! isIORank) { return; }
+
+        MessageBufferType buffer;
+        this->pack(0, buffer);
+
+        // pass a dummy_link to satisfy virtual class
+        const int dummyLink = -1;
+        this->unpack(dummyLink, buffer);
+    }
+
+    void pack(int link, MessageBufferType& buffer)
+    {
+        if (link != 0)
+            throw std::logic_error("link in method pack is not 0 as expected");
+        for (int i = 0; i < 3; ++i) {
+            const auto& name = localFlows_[i].first;
+            buffer.write(name);
+            unsigned int size = localFlows_[i].second.first.size();
+            buffer.write(size);
+            for (unsigned int j = 0; j < size; ++j) {
+                const auto& nncIdx = localFlows_[i].second.first[j];
+                buffer.write(nncIdx);
+                const auto& flows = localFlows_[i].second.second[j];
+                buffer.write(flows);
+            }
+        }
+    }
+
+    void unpack(int /*link*/, MessageBufferType& buffer)
+    {
+        for (int i = 0; i < 3; ++i) {
+            std::string name;
+            buffer.read(name);
+            globalFlows_[i].first = name;
+            unsigned int size = 0;
+            buffer.read(size);
+            for (unsigned int j = 0; j < size; ++j) {
+                int nncIdx;
+                double data;
+                buffer.read(nncIdx);
+                buffer.read(data);
+                if (nncIdx < 0)
+                    continue;
+                // This array is initialized in the collect(...) method below   
+                globalFlows_[i].second.second[nncIdx] = data;
+            }
+        }
+    }
 };
 
 template <class Grid, class EquilGrid, class GridView>
@@ -937,7 +999,9 @@ collect(const data::Solution& localCellData,
         const data::GroupAndNetworkValues& localGroupAndNetworkData,
         const data::Aquifers& localAquiferData,
         const WellTestState& localWellTestState,
-        const EclInterRegFlowMap& localInterRegFlows)
+        const EclInterRegFlowMap& localInterRegFlows,
+        const std::array<std::pair<std::string, std::pair<std::vector<int>, std::vector<double>>>, 3>& localFlowsn,
+        const std::array<std::pair<std::string, std::pair<std::vector<int>, std::vector<double>>>, 3>& localFloresn)
 {
     globalCellData_ = {};
     globalBlockData_.clear();
@@ -947,6 +1011,8 @@ collect(const data::Solution& localCellData,
     globalAquiferData_.clear();
     globalWellTestState_.clear();
     this->globalInterRegFlows_.clear();
+    globalFlowsn_ = {};
+    globalFloresn_ = {};
 
     // index maps only have to be build when reordering is needed
     if(!needsReordering && !isParallel())
@@ -965,6 +1031,16 @@ collect(const data::Solution& localCellData,
     if (! isParallel()) {
         // no need to collect anything.
         return;
+    }
+
+    // Set the right sizes for Flowsn and Floresn
+    for (int i = 0; i < 3; ++i) {
+        unsigned int sizeFlr = localFloresn[i].second.first.size();
+        globalFloresn_[i].second.first.resize(sizeFlr, 0);
+        globalFloresn_[i].second.second.resize(sizeFlr, 0.0);
+        unsigned int sizeFlo = localFlowsn[i].second.first.size();
+        globalFlowsn_[i].second.first.resize(sizeFlo, 0);
+        globalFlowsn_[i].second.second.resize(sizeFlo, 0.0);
     }
 
     PackUnPackWellData packUnpackWellData {
@@ -1009,6 +1085,18 @@ collect(const data::Solution& localCellData,
         this->isIORank()
     };
 
+    PackUnpackFlows packUnpackFlowsn {
+        localFlowsn,
+        this->globalFlowsn_,
+        this->isIORank()
+    };
+
+    PackUnpackFlows packUnpackFloresn {
+        localFloresn,
+        this->globalFloresn_,
+        this->isIORank()
+    };
+
     toIORankComm_.exchange(packUnpackCellData);
     toIORankComm_.exchange(packUnpackWellData);
     toIORankComm_.exchange(packUnpackGroupAndNetworkData);
@@ -1017,6 +1105,8 @@ collect(const data::Solution& localCellData,
     toIORankComm_.exchange(packUnpackAquiferData);
     toIORankComm_.exchange(packUnpackWellTestState);
     toIORankComm_.exchange(packUnpackInterRegFlows);
+    toIORankComm_.exchange(packUnpackFlowsn);
+    toIORankComm_.exchange(packUnpackFloresn);
 
 #ifndef NDEBUG
     // make sure every process is on the same page
--- a/ebos/eclgenericoutputblackoilmodule.cc
+++ b/ebos/eclgenericoutputblackoilmodule.cc
@@ -139,6 +139,33 @@ EclGenericOutputBlackoilModule(const EclipseState& eclState,
         std::string key_pattern = "R" + EclString(phase) + "*";
         this->regionNodes_[phase] = summaryConfig_.keywords(key_pattern);
     }
+
+    // Check if FLORES/FLOWS is set in any RPTRST in the schedule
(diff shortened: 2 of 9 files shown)


akva2:
jenkins build this please

bska:
Is this supposed to work in parallel, or is that follow-up work?

daavid00:
> Is this supposed to work in parallel, or is that follow-up work?

It runs with mpirun but I have not checked if the written fluxes are different, I will check that, thanks.

bska:
> > Is this supposed to work in parallel, or is that follow-up work?
> 
> It runs with mpirun but I have not checked if the written fluxes are different, I will check that, thanks.

Yes please, and especially the non-neighbouring fluxes.

bska (commented):
There are good parts here that we should continue to use, and especially the `I+`, `J+`, and `K+` arrays do the right thing, with the exception that the surface flows have the wrong units.  However, the `N+` arrays require more work, possibly quite a lot more work before they'll do the right thing in parallel.

The primary reason is that the non-neighbouring connection structure (`outputNnc_`) is non-empty only on the I/O rank, and as far as I can tell there is no logic here to construct per-rank versions of that structure which match the ordering of the global one on the I/O rank.  What's more, the "extra" arrays, even if they were non-empty on ranks other than the I/O rank, are not communicated to the I/O rank for output.

alfbr:
We would really like to test this as soon as possible, does it need a rebase other PRs, or something else I should be aware of?

vkip:
Parallel remains an issue here (may be seen on the Norne case, e.g.).

bska (commented):
Any review I do here is going to just be superficial because I don't understand how this works.  There are some datastructure issues that we can address over time, but I am concerned by @vkip's remark that there are still issues in parallel.

As a stop-gap I think this can go in, but we may need to rethink this solution at a later time.

daavid00:
> Any review I do here is going to just be superficial because I don't understand how this works. There are some datastructure issues that we can address over time, but I am concerned by @vkip's remark that there are still issues in parallel.
> 
> As a stop-gap I think this can go in, but we may need to rethink this solution at a later time.

Many thanks @vkip and @bska for looking at this pull request. This is the serial (left) and parallel (right )output for Norne at the first time step for FLOGASN+ with default flags (up) and with a maximum time step for the solver (down). I have tried to fix the differences (e.g., adding comm.barrier() in different parts in the code) but I did not succeed. If someone has a solution for this, it would be very much appreciated.

norne_flows_flores

ebos/eclgenericwriter.cc
```diff
@@ -463,6 +473,27 @@ doWriteOutput(const int                     reportStepNum,
         restartValue.addExtra("OPMEXTRA", std::vector<double>(1, nextStepSize));
     }
 
+    // Add nnc flows and flores is existing.
+    if (isFlowsn) {
+        const auto flowsn_global = isParallel ? this->collectToIORank_.globalFlowsn() : std::move(flowsn);
+        for (const auto& flows : flowsn_global) {
+            if (flows.first.empty())
+                continue;
+            restartValue.addExtra(flows.first, UnitSystem::measure::rate, flows.second.second);
```
atgeirr, line 482: I think the unit here is wrong, the unit here is for reservoir volumes. The logic here and on lines 490--494 should be swapped. Testing with field units should show the problem.

ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -762,6 +795,47 @@ assignToSolution(data::Solution& sol)
         // reallocated)
         tracerConcentrations_.resize(0);
     }
+
+    // Flows
+    if (FluidSystem::phaseIsActive(gasPhaseIdx) && !flowsi_[gasCompIdx].empty())
```
atgeirr, line 800: It seems to me that the first condition is unneeded: there should be no way to get a nonempty `flowsi_[gasCompIdx]` if we have no active gas phase. So only the second part of the condition (emptyness) is needed, and then the generic mechanism used on lines 699--752 (line numbers from your branch) should be used. You may have to be careful about making sure the vectors all exist (I assume they do, not read as far), but are empty, in case of a nonactive phase.

ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -1075,6 +1150,104 @@ doAllocBuffers(unsigned bufferSize,
...
         invB_[gasPhaseIdx].resize(bufferSize, 0.0);
     }
+    
+    enableFlows_ = false;
+    enableFlowsn_ = false;
+    if (rstKeywords["FLOWS"] > 0) {
+        rstKeywords["FLOWS"] = 0;
+        enableFlows_ = true;
+        if (FluidSystem::phaseIsActive(gasPhaseIdx)){
+            if (this->eclState_.gridDims()[0] > 1)
```
atgeirr, line 1160: This test should be `> 0` not ` > 1`, as we should output a J+ value even if we have only one cell in the J direction. Postprocessors such as ResInsight expect this. You could test this with a 10 x 1 x 10 grid.

ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -1075,6 +1150,104 @@ doAllocBuffers(unsigned bufferSize,
         rstKeywords["BG"] = 0;
         invB_[gasPhaseIdx].resize(bufferSize, 0.0);
     }
+    
+    enableFlows_ = false;
+    enableFlowsn_ = false;
+    if (rstKeywords["FLOWS"] > 0) {
+        rstKeywords["FLOWS"] = 0;
+        enableFlows_ = true;
+        if (FluidSystem::phaseIsActive(gasPhaseIdx)){
```
atgeirr, line 1159: Instead of doing identical logic with three different indices you could:
```c++
auto phaseIdxs = std::array{ gasPhaseIdx, oilPhaseIdx, waterPhaseIdx }; // Using class template argument deduction to avoid saying the template arguments for array, neat!
auto compIdxs = std::array{gasCompIdx, oilCompIdx, waterCompIdx };
for (int ii = 0; ii < phaseIdxs.size(); ++i) {
    if (FluidSystem::phaseIsActive(phaseIdxs[ii]) {
    .... Use compIdx etc.
    }
}

File: ebos/ecloutputblackoilmodule.hh [hunk 5]
```diff
@@ -685,6 +688,98 @@ public:
...
                 }
             }
+
+            // flows
+            if (!problem.model().linearizer().getFlowsInfo().empty()){
+                const auto& flowsInf = problem.model().linearizer().getFlowsInfo();
+                auto flowsInfos = flowsInf[globalDofIdx];
+                for (auto& flowsInfo : flowsInfos) {
+                    if (flowsInfo.faceId == 1){
+                        if (FluidSystem::phaseIsActive(gasPhaseIdx) & !this->flowsi_[gasCompIdx].empty())
```
atgeirr, line 698: The `&` should probably be a `&&`. That is logical and, not bitwise and. Maybe one of the most easy to misuse operators in C++! Using it indicates that you are doing bit manipulation, and not pure boolean ops.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -685,6 +688,98 @@ public:
...
+                            this->flowsi_[gasCompIdx][globalDofIdx] = flowsInfo.flow[Indices::canonicalToActiveComponentIndex(gasCompIdx)];
```
atgeirr, line 699: Please break lines that get very long. Or store the Indices:: canonicalToActiveComponentIndex(gasCompIdx) et. numbers in variables to get a more readable line.

ebos/eclwriter.hh
```diff
@@ -370,7 +378,11 @@ public:
                                 this->summaryState(),
                                 simulator_.problem().thresholdPressure().data(),
                                 curTime, nextStepSize,
-                                EWOMS_GET_PARAM(TypeTag, bool, EclOutputDoublePrecision));
+                                EWOMS_GET_PARAM(TypeTag, bool, EclOutputDoublePrecision),
+                                isFlowsn,
+                                std::move(flowsn),
```
atgeirr, line 383: I don't think this will actually move anything, since the receiving function takes the argument by const reference and not move reference (i.e. `Object&&` rather than `const Object&`).

atgeirr:
I have now (with @bska) gone through all of this branch, **except** the collect to io rank parts, and if the comments given are adressed, I think those parts would be ok to merge.

However, the parallel collection of the data is known to be faulty, and must be debugged and reviewed. @blattms maybe you can assist with this?

daavid00:
> I have now (with @bska) gone through all of this branch, **except** the collect to io rank parts, and if the comments given are adressed, I think those parts would be ok to merge.
> 
> However, the parallel collection of the data is known to be faulty, and must be debugged and reviewed. @blattms maybe you can assist with this?

Many thanks @atgeirr and @bska for your comments, I have addressed them and pushed the commit.

ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -139,6 +139,39 @@ EclGenericOutputBlackoilModule(const EclipseState& eclState,
...
+        if(anyFlores_ == true)
+            continue;
+
+        std::map<std::string, int> rstKeywords = schedule_.rst_keywords(step);
+        for (auto& [keyword, should_write] : rstKeywords) {
+            if (keyword == "FLORES") {
+                anyFlores_ = true;
+            }
+        }
+    }
```
bska, line 174: You can probably simplify this section further, and avoid copying the [`RSTConfig::keywords`](https://github.com/OPM/opm-common/blob/b83f10d2d70d160e50a6df209eeb8f7cffd4c774/opm/input/eclipse/Schedule/RSTConfig.hpp#L217) map, if you do something along the lines of
```C++
for (const auto& block : this->schedule_) { // Uses Schedule::begin() and Schedule::end()
    const auto& rstkw = block.rst_config().keywords;

    if (! anyFlows_) {
        anyFlows_ = rstkw.find("FLOWS") != rstkw.end();
    }

    if (! anyFlores_) {
        anyFlores_ = rstkw.find("FLORES") != rstkw.end();
    }

    if (anyFlows_ && anyFlores_) {
        // Terminate report step loop early if both FLOWS and FLORES
        // have been set at some point as there's no need to search
        // any further in that case.
        break;
    }
}
```

ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -762,6 +795,47 @@ assignToSolution(data::Solution& sol)
         // reallocated)
         tracerConcentrations_.resize(0);
     }
+
+    // Flows
+    if (!flowsi_[gasCompIdx].empty())
+        sol.insert("FLOGASI+", UnitSystem::measure::gas_surface_rate, flowsi_[gasCompIdx], data::TargetType::RESTART_SOLUTION);
```
bska, line 801: I think @atgeirr mentioned this in the earlier review, but I didn't see a response.  Is there a reason you chose to use separate logic for the flow arrays here instead of integrating these into the output system using the [existing](https://github.com/OPM/opm-simulators/blob/38282e30f7d3fa8a97cce3f8bc5efe07bad5dbb9/ebos/eclgenericoutputblackoilmodule.cc#L666) table of arrays?

ebos/collecttoiorank.cc
```diff
@@ -805,6 +805,128 @@ class PackUnpackInterRegFlows : public P2PCommunicatorType::DataHandleInterface
...
+            unsigned int size = 0;
+            buffer.read(size);
+            for (unsigned int j = 0; j < size; ++j) {
+                int nncIdx;
+                double data;
+                buffer.read(nncIdx);
+                buffer.read(data);
+                if (nncIdx < 0)
+                    continue;
+                globalFlowsn_[i].second.second[nncIdx] = data;
```
bska, line 863: Is there a guarantee that the array element
```
globalFlowsn_[i].second.second[nncIdx]
```
exists, i.e., has been properly allocated, on the receiving process?  If so, I think it'd be appropriate to add a comment here stating what that guarantee is.

ebos/collecttoiorank.cc
```diff
@@ -805,6 +805,128 @@ class PackUnpackInterRegFlows : public P2PCommunicatorType::DataHandleInterface
...
+            }
+        }
+    }
+};
+
+class PackUnpackFloresn : public P2PCommunicatorType::DataHandleInterface
```
bska, line 869: I don't understand why you need both `PackUnpackFlowsn` and `PackUnpackFlowresn`.  Apart from the names of the data members, these appear to do exactly the same thing and could be fused into `PackUnpackFaceFlows` or something like that.

bska (commented):
I had another look here and there are still some opportunities to simplify the logic.

totto82:
@vkip Can you elaborate a bit. @daavid00  is not able to reproduce. There are differences but they seems to be more a result of differences in the solution. i.e. they are significantly reduced with smaller time-steps. See the attached file. 
The results are organized as following. 

serial                             |  parallel 
serial + small timestep |  parallel with small timestep 

https://user-images.githubusercontent.com/61784809/214692635-ab34de52-63b8-43a8-baa4-949365a2f4c3.png

vkip:
> @vkip Can you elaborate a bit. @daavid00 is not able to reproduce. There are differences but they seems to be more a result of differences in the solution. i.e. they are significantly reduced with smaller time-steps. See the attached file. The results are organized as following.
> 
> serial | parallel serial + small timestep | parallel with small timestep
> 
> https://user-images.githubusercontent.com/61784809/214692635-ab34de52-63b8-43a8-baa4-949365a2f4c3.png

Thanks for looking into this. The comment was based on using the results in ResInsight Well Allocation plots, which showed a (to me) surprisingly large difference depending on the number of CPUs. When not using this PR (i.e., no FLORES in the deck, causing RI to estimate the reservoir flows) the results were virtually identical (and equal to the reference simulator results), indicating that the other relevant output (well and connection rates) were not the issue.  

Having looked closer now, I see that there is some discrepancy in these results between serial and parallel also for the reference tool (so the allocation is likely quite sensitive to small solution differences..), and I can reproduce your finding above with shorter time-steps. Hence I'm now coming around to your conclusion, so please feel free to merge whenever the reviewers are happy. 

Attaching some plots for anyone interested. 
[PR4275 (FLORES).pdf](https://github.com/OPM/opm-simulators/files/10540264/PR4275.FLORES.pdf)

atgeirr:
I think this can be merged when green.

bska:
Thanks for the updates.  Would you mind rebasing this PR onto the top of the current master branch please?  That will remove the merge commit from this patch series.

PR #857, opened 2016-10-12
update redhat6 packaging

akva2:
build using devtoolset-3

PR #4278, opened 2022-11-21
Updates related to unsupported keywords and options

tskille:
Majority of unsupported keywords and unsupported keyword options are now critical (simulator will stop).


tskille:
As a starting point all unsupported keywords are defined as critical (simulator will stop).  Unsupported keywords which cannot change simulation results (saturations, pressures, rates, etc) are defined non-critical. Examples of keywords in this category are  **DEBUG, ECHO, NOECHO and RPTSOL**.  There are also a group of keywords that are used to modify simulator control parameters which Flow doesn’t support.  In the commercial simulator these will impact on the simulation results. However OPM Flow have alternative ways to set these via command line.  Examples of keywords in this category are  **TUNING, TUNINGDP, NSTACK, PARALLEL and CPR**.   For non-critical unsupported keywords OPM Flow will print a warning to the terminal (and the PRT file), then the simulation will carry on.

Notice that unsupported keyword in the summary section are handled separately. None of these are defined as critical.  

This PR should not be merged before  [PR #857 in opm-tests](https://github.com/OPM/opm-tests/pull/857)  is merged.

bska:
jenkins build this opm-tests=857 please

bska:
jenkins build this please

OPMUSER:
@tskille What abut the command line parameter `--ecl-strict-parsing`, is this still useful?

opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
```diff
@@ -230,85 +239,86 @@ partiallySupported()
...
-               {6,{false, allow_values<std::string> {}, "TRACER(ADSPHASE): partitioned tracer model not supported - ignored as not used"}}, // ADSORB_PHASE
+               {1,{true, [](const std::string& val){ return val.size()<=3;}, "TRACER(NAME): Only names of tracers up to 3 characters are supported."}},
+               {4,{true, allow_values<std::string> {}, "TRACER(SOLPHASE): partitioned tracer model not supported use default"}}, // SOLUTION_PHASE
+               {6,{true, allow_values<std::string> {}, "TRACER(ADSPHASE): partitioned tracer model not supported use default"}}, // ADSORB_PHASE
             },
          },
          {
             "TRACERS",
             {
                {5,{false, allow_values<std::string> {"NODIFF"}, "TRACERS(DIFFOPT): numerical diffusion control not implemented. - ignored as not used"}}, // NUMERIC_DIFF
```
vkip, line 250: Suggest to stop for anything else than NODIFF (until implemented), since the tracer solution will change depending on this setting.

opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
```diff
@@ -230,85 +239,86 @@ partiallySupported()
...
-               {8,{false, allow_values<std::string> {"NO"}, "TRACERS(NONLIN): only linear option NO supported – will continue"}}, // PASSIVE_NONLINEAR
+               {8,{false, allow_values<std::string> {"NO"}, "TRACERS(NONLIN): only linear option NO supported"}}, // PASSIVE_NONLINEAR
```
vkip, line 251: Might as well stop here also.

vkip:
LGTM - only have comments to the TRACERS keyword.

opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
tskille: > What abut the command line parameter `--ecl-strict-parsing`, is this still useful?

I thing this is something different. This is more related to  [ParseContext](https://github.com/OPM/opm-common/blob/a154a2a49f696d30e617285c8d9897ae8714d8ae/src/opm/input/eclipse/Parser/ParseContext.cpp#L73)

missing slashes and so on. Personally I haven't used this option ever as far as I can remember. 

When/If this PR goes in, I think it would be useful with a command option which will accept all keyword (ignore the critical variable) and just run until something crashes (or in some cases the simulator will actually run).

opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
tskille: > What abut the command line parameter --ecl-strict-parsing, is this still useful?

I thing this is something different. This is more related to [ParseContext](https://github.com/OPM/opm-common/blob/a154a2a49f696d30e617285c8d9897ae8714d8ae/src/opm/input/eclipse/Parser/ParseContext.cpp#L73)

missing slashes and so on. Personally I haven't used this option ever as far as I can remember.

When/If this PR goes in, I think it would be useful with a command option which will accept all keyword (ignore the critical variable) and just run until something crashes (or in some cases the simulator will actually run).

tskille:
> What abut the command line parameter `--ecl-strict-parsing`, is this still useful?

I thing this is something different. This is more related to ParseContext

missing slashes and so on. Personally I haven't used this option ever as far as I can remember.

When/If this PR goes in, I think it would be useful with a command option which will accept all keyword (ignore the critical variable) and just run until something crashes (or in some cases the simulator will actually run).

bska:
jenkins build this please

OPMUSER:
> When/If this PR goes in, I think it would be useful with a command option which will accept all keyword (ignore the critical variable) and just run until something crashes (or in some cases the simulator will actually run).

@tskille I think that is a good idea, then when stopping for an unsupported keyword, we can issue a message about the command line parameter. 

I noticed that you removed the `"- will stop"` comment from the message for when we invoke the STOP option,  is there an end of rerun message, stating  "`Simulation Terminated Due to Errors"` message?  Also we should be consistent with the message, so either all STOPS have `"- will stop"` or not.At the moment we have a mixture.

```
  "FOAMOPTS", // ?? not supported according to manual > should be added to UnsupportedFlowKeywords ??
            {
               {1,{true, allow_values<std::string> {"GAS"}, "FOAMOPTS(FOAMOPT1): only the default option of GAS is supported"}}, // TRANSPORT_PHASE
               {2,{true, allow_values<std::string> {"TAB"}, "FOAMOPTS(FOAMOPT2): only the default option of TAB is supported"}}, // MODEL
            },
```
The keyword is supported but only the default values.
```

             {2,{true, allow_values<std::string> {"NONE", "FLD", "ORAT", "WRAT", "GRAT", "LRAT", "RESV"}, "GCONPROD(TARGET): valid option should be NONE/FLD/ORAT/WRAT/GRAT/LRAT or RESV"}}, // CONTROL_MODE
               {11,{true, allow_values<std::string> {"NONE"}, "GCONPROD(ACTWAT): water violation procedure not implemented, item should be defaultd"}}, // WATER_EXCEED_PROCEDURE
               {12,{true, allow_values<std::string> {"NONE"}, "GCONPROD(ACTGAS): gas violation procedure not implemented, item should be defaultd"}}, // GAS_EXCEED_PROCEDURE
               {13,{true, allow_values<std::string> {"NONE"}, "GCONPROD(ACTLIQ): liquid violation procedure not implemented, item should be defaultd"}}, // LIQUID_EXCEED_PROCEDURE
               {21,{true, allow_values<std::string> {"NONE"}, "GCONPROD(COMBPROC): linearly combined procedure is not used and should be defaulted (1*)"}}, // LIN_TARGET_EXCEED_PROCEDURE
```

Typo defaultd > defaulted  in (11) to (21)

 ```
        {
            "GUIDERAT",
            {
               {2,{true, allow_values<std::string> {"OIL", "LIQ", "GAS", "RES", "NONE"}, "GUIDERAT(PHASE): unsupported option must be OIL LIQ GAS RES or NONE"}}, // NOMINATED_PHASE
               {9,{true, allow_values<std::string> {"YES"}, "GUIDERAT(GROPT01): only the default option of YES supported"}}, // ALLOW_INCREASE
            },
         },
 ``` 
Typo >  only the default option of YES **is** supported"}
```
            "PINCH",
            {
               {2,{true, allow_values<std::string> {"GAP"}, "PINCH with NOGAP in item 2 is not supported by Flow."}},
            },
```
I'm not sure about this but the manual states that both GAP and NOGAP are supported, may be @bska knows. In any case, I think the message should follow the same format as the other messages, that is: `PINCH(PINCHOPT) equal to NOGAP is not supported.` Also I don't think we need `by Flow`.

```
          "ROCKOPTS",
            {
               {1,{true, allow_values<std::string> {"PRESSURE"}, "ROCKOPTS: only the PRESSURE options are supported"}}, 
               {2,{true, allow_values<std::string> {"NOSTORE"}, "ROCKOPTS: only the NOSTORE options are supported"}}, 
               {3,{true, allow_values<std::string> {"PVTNUM", "ROCKNUM"}, "ROCKOPTS: only PVTNUM and ROCKNUM are supported"}}, 
            },
```
I have this as an unsupported keyword, whereas you are are saying the default values are supported. Is that because if the keyword is missing in the input deck then the defaults will be applied? If so I need to update the manual for ROCKOPTS(ROCKOP1) to ROCKOPTS(ROCKOPT3). I assume that the same is true for ROCKOPT(ROCKOPT4), that is we support the default value - correct?.  

```
           "TRACERS",
            {
               {5,{true, allow_values<std::string> {"NODIFF"}, "TRACERS(DIFFOPT): numerical diffusion control not implemented."}}, // NUMERIC_DIFF
```
Change message for better clarity: "TRACERS(DIFFOPT): numerical diffusion control for **tracers** not implemented."

```
           "WTEST",
            {
               {3,{true, allow_values<std::string> {"E", "P", "EP", "PE", ""}, "WTEST(TEST): only the E (economic) and P (physical) reason is currently supported"}}, // REASON
            },
         },
```

The manual only has the E option supported, are you sure that the P is supported as well?

 ```
"WVFPEXP",
            {
               {5,{false, allow_values<std::string> {"WG"}, "WVFPEXP(EXTRAPOLATION_CONTROL): only support linear extrapolation"}}, // EXTRAPOLATION_CONTROL
            },
         },
```
My understanding is that WVFPEXP(EXTRAP) is a E300 option only.,so it is not used by E100. I know the behavior in E100 is linear extrapolation. Perhaps we should change this to:

 ```
"WVFPEXP",
            {
               {5,{false, allow_values<std::string> {"NONE"}, "WVFPEXP(EXTRAP): only linear extrapolation is support "}}, // EXTRAPOLATION_CONTROL
            },
         },
```
That way we avoid any confusion.

I'll continue with the rest tomorrow.

bska:
jenkins build this please

tskille:
> is there an end of rerun message, stating "`Simulation Terminated Due to Errors"` message?

No I it is not for the time being.  Hoever, I believe that **-- will stop** and **-- will continue** is redundant. From the rest of the logfile it is very clear if the simulator has stopped or if it has continued. 

The commercial simulator Eclipse will continue with Warnings and Problems and stop with Errors and Bugs. I believe we should do the same. 

The unsupported non-critical keywords will be listed under warning (with blue font when printed on a terminal).  
```
Warning: Unsupported keywords or keyword items:

  RPTGRID: keyword not supported
  In file: TEST1.DATA, line 120

  NOECHO: keyword not supported
  In file: TEST1.DATA, line 132

  RPTRST: invalid value 'BASIC = 2' for item 1
  In file: TEST1.DATA, line 170
  RPTRST(RPTRST): invalid option or unsupported integer control format
```

Unsupported critical keywords and keyword items will be listed as Errors in read font. Below I have added SKIP and ENDSKIP to one of the models in opm-test.  

```
Error: Unsupported keywords or keyword items:

  SKIP: keyword not supported
  In file: TEST1.DATA, line 147

  ENDSKIP: keyword not supported
  In file: TEST1.DATA, line 149

In <memory string> line 0
Unsupported keywords or keyword items:

  SKIP: keyword not supported
  In file: TEST1.DATA, line 147

  ENDSKIP: keyword not supported
  In file: TEST1.DATA, line 149
```

In the current version of Flow there will not be any lines after this. There is no line explicitly stating that the simulator has stopped due to errors. I don't see any problems with having that. The above error message should any way be fixed. Keywords not supported are listed twice and the message "Unrecoverable errors while loading input" should in my opinion be removed. 

But this will not be done in this PR. This PR is focusing on critical and non-critical keywords and keyword options.

tskille:
> I'm not sure about this but the manual states that both GAP and NOGAP are supported, may be @bska knows.

We have a full field model internally in Equinor which demonstrates that PINCH and NOGAP is not supported.

tskille:
> The manual only has the E option supported, are you sure that the P is supported as well?

the test model [MIN_BHP_1.DATA](https://github.com/OPM/opm-tests/blob/8533f2a135ef86428d638e985479f804c3d82195/wtest/bhp_min/MIN_BHP_1.DATA#L326)

tskille:
@OPMUSER, suggestions not commented on are,  implemented in the updated commit.

tskille:
> I'll continue with the rest tomorrow.

Excellent, please let me know when you been through the PR.

OPMUSER:
> We have a full field model internally in Equinor which demonstrates that PINCH and NOGAP is not supported.

Noted and manual updated accordingly.

> The manual only has the E option supported, are you sure that the P is supported as well?
> 
> the test model [MIN_BHP_1.DATA](https://github.com/OPM/opm-tests/blob/8533f2a135ef86428d638e985479f804c3d82195/wtest/bhp_min/MIN_BHP_1.DATA#L326)

Noted and manual updated accordingly.

> No I it is not for the time being. Hoever, I believe that -- will stop and -- will continue is redundant. From the rest of the logfile it is very clear if the simulator has stopped or if it has continued.

I'm okay with this.

OPMUSER:
Pressed the wrong button - sorry about that.

tskille:
> I think we should treat these two parameters the same, rather having one as a warning and the other an error. suggest we change the messages to:

NTTRVD item 4 in EQLDIMS, we are supporting tracer depth tables, but only one table for each tracer (which is default). Hence your suggestion is misleading.

Flow is ignoring NSTRVD (item 5) in EQLDIMS since memory are allocated dynamically in Flow. Hence I think we should remove this from Partiallysupported.

tskille:
> Also consider EQLOPT4 and EQLOPT5 be warning messages as these are E300 compositional options.

If these are set to something different that defaults, this could be an indication that this actually is an input deck for Eclipse 300. If that is the case, we should really stop in my opinion.

tskille:
> NETBALAN(THPMXITE)

This is a numerical tuning parameter. There are most likely alternative ways to control this via command line parameters.

tskille:
> What about items (12) to (17) I have all these as unsupported, but I think (12) and (13) may be?

Notice that NNC is listed 3 times,  

12 and 13 std::string, not supported critical = true 
8,9,10,11 int, not supported critical = true
14,15,16,17 double, not supported critical = true

Item 8 -> 17 not supported

tskille:
> Need add items (4) and (6).

This are there, see section under std::string

OPMUSER:
>  I think we should treat these two parameters the same, rather having one as a warning and the other an error. suggest we change the messages to:

> NTTRVD item 4 in EQLDIMS, we are supporting tracer depth tables, but only one table for each tracer (which is default). Hence your suggestion is misleading.

Suggest changing message to:

"EQLDIMS(NTTRVD): only the default of one tracer depth table per tracer is allowed, as TNUM is unsupported ."

or something similar.  In the above I'm trying to give the impression that we support tracer versus depth tables, but only for TNUM =1. In many cases the default value implies that the feature is absent or not used.

> 
> Flow is ignoring NSTRVD (item 5) in EQLDIMS since memory are allocated dynamically in Flow. Hence I think we should remove this from Partiallysupported.

Agreed.

OPMUSER:
>  Also consider EQLOPT4 and EQLOPT5 be warning messages as these are E300 compositional options.

> If these are set to something different that defaults, this could be an indication that this actually is an input deck for Eclipse 300. If that is the case, we should really stop in my opinion.

I tend to be a bit more forgiving, as the deck may have originally come from an E300 case; but since we ignore the values and thus they don't effect the results, I would suggest a warning message. Since, if the deck is a E300 case it will stop once it hits a specific E300 keyword in the deck anyway. I also think this should be the general approach, that is E300 parameters result in warnings, but E300 keywords results in errors.

OPMUSER:
> NETBALAN(THPMXITE)

> This is a numerical tuning parameter. There are most likely alternative ways to control this via command line parameters.

I checked the command line parameters and there are no such control for manifold groups. 

Since this parameter will potentially change the results when compared to E100 and thus I would recommend an error instead of a warning. Most engineers will benchmark their runs against E100, so if we prevent decks running that have a potential to give "different" results due to numerical parameters not being applied, I think that is a positive outcome. They can always run E100 with the same configuration to verify against OPM Flow. 

This just a suggestion, I would object if this was not the case, but not too strongly.

OPMUSER:
> What about items (12) to (17) I have all these as unsupported, but I think (12) and (13) may be?

> Notice that NNC is listed 3 times,
> 
> 12 and 13 std::string, not supported critical = true 
> 8,9,10,11 int, not supported critical = true
> 14,15,16,17 double, not supported critical = true
> 
> Item 8 -> 17 not supported

Yep, you are right - my brain must have stopped working for a couple of minutes :-).

OPMUSER:
>  TRACER - Need add items (4) and (6).

> This are there, see section under std::string

Yep, you are right.

OPMUSER:
```
            "GCONINJE",
            {
               {13,{true, allow_values<double> {}, "GCONINJE(WGASRATE): wet gas rate is not used and should be defaulted (1*)"}}, // WETGAS_TARGET
            },
         },
```
Again, as this is an E300 parameter I think we should issue a warning not an error.

```
            "GCONPROD",
            {
               {15,{true, allow_values<double> {}, "GCONPROD(RESVFRAC): reservoir volume fraction is not supported and should be defaulted (1*)"}}, // RESERVOIR_VOLUME_BALANCE```

               {16,{true, allow_values<double> {}, "GCONPROD(WGASRATE): wet gas rate is not used and should be defaulted (1*)"}}, // WETGAS_TARGET
               {17,{true, allow_values<double> {}, "GCONPROD(CALRATE): calorific rate is not used and should be defaulted (1*)"}}, // CALORIFIC_TARGET
               {18,{true, allow_values<double> {}, "GCONPROD(GASFRAC): gas production fraction is not used and should be defaulted (1*)"}}, // SURFACE_GAS_FRACTION
               {19,{true, allow_values<double> {}, "GCONPROD(WATFRAC): water production fraction is not used and should be defaulted (1*)"}}, // SURFACE_WAT_FRACTION
               {20,{true, allow_values<double> {}, "GCONPROD(COMBRATE): linearly combined rate is not used and should be defaulted (1*)"}}, // LINEAR_COMBINED_TARGET
            },
         },
```
Items (16) to (19) are E300 parameters, so I think we should issue a warning not a stop. Item (20) is correct with stopping as an error, as we don't support the linearly combined production rate facility.

```

          "NETBALAN",
            {
               {1,{true, [](const double value) { return value <= 0.0; }, "NETBALAN(NSTEP): only negative values or 0 supported"}}, // TIME_INTERVAL
               {4,{false, allow_values<double> {0.01}, "NETBALAN(GRPCNV): not supported"}}, // THP_CONVERGENCE_LIMIT
               {6,{false, allow_values<double> {1e20}, "NETBALAN(NTRGERR): not supported"}}, // TARGET_BALANCE_ERROR
               {7,{false, allow_values<double> {1e20}, "NETBALAN(NMAXERR): not supported"}}, // MAX_BALANCE_ERROR
               {8,{false, allow_values<double> {}, "NETBALAN(NTSMIN): not supported"}}, // MIN_TIME_STEP
            },
         },
```
Upon reflection, I think items (4), (6) - (8) should be errors,  simply because they will effect the results.
```

 "NNC",
            {
               {14,{true, allow_values<double> {0}, "NNC(DIFFNNC): not supported"}}, // DIFFUSIVITY
               {15,{true, allow_values<double> {0}, "NNC(DISPNNC): not supported"}}, // SIM_DEPENDENT3
               {16,{true, allow_values<double> {0}, "NNC(AREANNC): not supported"}}, // VDFLOW_AREA
               {17,{true, allow_values<double> {0}, "NNC(PERMNNC): only default value of 0 supported"}}, // VDFLOW_PERM
            },
         },
```

For item (17) E100 has no default value, although OPM Flow's JSON file has a default value of 0.0. Thus, consider changing the message to   {17,{true, allow_values<double> {0}, "NNC(PERMNNC): not supported"}},


```           
"ROCKCOMP",
            {
               {4,{true, allow_values<double> {0}, "ROCKCOMP(CARKZEXP): transmissibility dependent on porosity model is not supported"}}, // CARKZEXP
            },
         },
```    
This should be item (5) not (4). The original was also wrong as well. The string item(4) entry is correct.

```
 "WCONHIST",
            {
               {11,{true, [](double x) { return x == 0; }, "WCONHIST(WGRA): wet gas rate is not supported use default"}}, // WGASRAT
               {12,{true, [](double x) { return x == 0; }, "WCONHIST(NGL): natural gas rate is not supported use default"}}, // NGLRAT
            },
         },
```

Again, as these are E300 parameters , so I think we should issue a warning not an error.

```
 "WCONPROD",
            {
               {13,{true, allow_values<double> {}, "WCONPROD(WGASRATE): wet gas rate is not used and should be defaulted (1*)"}}, // E300_ITEM13
               {14,{true, allow_values<double> {}, "WCONPROD(MOLARATE): molar rate not used and should be defaulted (1*)"}}, // E300_ITEM14
               {15,{true, allow_values<double> {}, "WCONPROD(STEAMRAT): steam rate is not used and should be defaulted (1*)"}}, // E300_ITEM15
               {16,{true, allow_values<double> {}, "WCONPROD(DELTAP): pressure offset not used and should be defaulted (1*)"}}, // E300_ITEM16
               {17,{true, allow_values<double> {}, "WCONPROD(DELTAT): temperature offset not used and should be defaulted (1*)"}}, // E300_ITEM17
               {18,{true, allow_values<double> {}, "WCONPROD(CALRATE): calorific rate not used and should be defaulted (1*)"}}, // E300_ITEM18
               {19,{true, allow_values<double> {}, "WCONPROD(COMBPROC): linearly combined rate not used and should be defaulted (1*)"}}, // E300_ITEM19
               {20,{true, allow_values<double> {}, "WCONPROD(NGL): natural gas liquid rate  is not used and should be defaulted (1*)"}}, // E300_ITEM20
            },
```
Items (13) to (18) and (20) are E300 parameters, so I think we should issue a warning not an error. Item (19) is correct with stopping as an error, as we don't support the linearly combined production rate facility.

tskille:
> Again, as this is an E300 parameter I think we should issue a warning not an error.

could you please argue why you think FLow should continue if Eclipse300 specific options are set to something different that default. For me this is a red flag and I really think we should stop when we se this.

vkip:
> > Again, as this is an E300 parameter I think we should issue a warning not an error.
> 
> could you please argue why you think FLow should continue if Eclipse300 specific options are set to something different that default. For me this is a red flag and I really think we should stop when we se this.

It is possible to set up E100 and E300 in the same deck using SKIP100/SKIP300 blocks, and allowing E300-specific options limits  the required number of such blocks. At present it does not seem these keywords are supported though, and even when implemented I would prefer the extra blocks to clearly indicate which options are relevant for which simulator. I'm therefore in favor of stopping here.

OPMUSER:
> could you please argue why you think FLow should continue if Eclipse300 specific options are set to something different that default. For me this is a red flag and I really think we should stop when we se this.

To be clear, I think that we should stop if the deck contains any E300 keyword, as we do not support compositional models.  Thus, if the deck does not include E300 keywords then the E300 specific parameters on the keywords can be safely ignored with just a warning message.

> It is possible to set up E100 and E300 in the same deck using SKIP100/SKIP300 blocks, and allowing E300-specific options limits the required number of such blocks. At present it does not seem these keywords are supported though, and even when implemented I would prefer the extra blocks to clearly indicate which options are relevant for which simulator. I'm therefore in favor of stopping here.

None of the SKIP keywords are supported, except for the SKIPREST. So it is only necessary to stop if there are E300 keywords in the deck.

tskille:
> NETBALAN; Upon reflection, I think items (4), (6) - (8) should be errors, simply because they will effect the results.

These are numerical tuning parameters and falls in to the same category as TUNING. We are not strictly following Eclipse on these parameters.

tskille:
> E300 specific parameters on the keywords can be safely ignored with just a warning message

All E300 specific options are changed to non-critical.

tskille:
@OPMUSER  have you been through all of the PR now ? I believe I have responded to all your comments.

OPMUSER:
> @OPMUSER have you been through all of the PR now ? I believe I have responded to all your comments.

Yes, looks good.

tskille:
Excellent, thanks.

bska:
jenkins build this please

bska (approved):
Build check is green and all comments have been addressed.  I'll merge into master.  Thanks a lot!

blattms:
will recover EclStrictParsing for this which is unused currently.

blattms:
Seems like there is an alternative: Setting environment variable OPM_ERRORS_IGNORE to "*"  prevents the simulator from stopping because of unsupported keywords. bash: `export OPM_ERRORS_IGNORE="*"` or in csh: `setenv OPM_ERRORS_IGNORE "*"`

PR #4283, opened 2022-11-22
StandardWellAssemble: simplify by using primary variables class

akva2:
- pass primary variables as parameter
- use the evaluation type

### DIFF
--- a/opm/simulators/wells/StandardWellAssemble.cpp
+++ b/opm/simulators/wells/StandardWellAssemble.cpp
@@ -32,6 +32,7 @@
 #include <opm/models/blackoil/blackoiltwophaseindices.hh>
 
 #include <opm/simulators/wells/StandardWellEquations.hpp>
+#include <opm/simulators/wells/StandardWellPrimaryVariables.hpp>
 #include <opm/simulators/wells/WellAssemble.hpp>
 #include <opm/simulators/wells/WellBhpThpCalculator.hpp>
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this please

opm/simulators/wells/StandardWellAssemble.cpp
```diff
@@ -264,48 +256,6 @@ assembleZFracEq(const EvalWell& cq_s_zfrac_effective,
 
 #define INSTANCE(Dim,...) \
 template class StandardWellAssemble<BlackOilFluidSystem<double,BlackOilDefaultIndexTraits>,__VA_ARGS__,double>; \
-template void \
-StandardWellAssemble<BlackOilFluidSystem<double,BlackOilDefaultIndexTraits>,__VA_ARGS__,double>:: \
```
bska, line 268: Hm, removing this leaves the previous line (`template class StandardWellAssemble<...>`) ending in a backslash character (`\`).  It is benign in this case since the following line is blank, but I think it might be preferable to remove the trailing backslash here just to be on the safe side.

bska (commented):
I like the reduction in the number of lines, but I noticed one potential issue which I think we should fix before merging.

opm/simulators/wells/StandardWellAssemble.cpp
akva2: indeed.

bska (approved):

PR #4284, opened 2022-11-23
changed: remove EclProblem::checkDeckCompatibility

akva2:
these checks are disabled and do not add anything new, as the checks of value here are also performed elsewhere


akva2:
jenkins build this please

PR #4285, opened 2022-11-23
Use internalized THPRESFT

akva2:
To avoid deck usage in the simulators.

Downstream of https://github.com/OPM/opm-common/pull/3225

PR #4286, opened 2022-11-23
Remove Deck member from simulators

akva2:
No reason holding on to the deck during simulations, just a waste of memory. Let's remedy that

Sits on top of
https://github.com/OPM/opm-simulators/pull/4284
https://github.com/OPM/opm-simulators/pull/4285

And is downstream of 
https://github.com/OPM/opm-common/pull/3225

### DIFF
--- a/opm/simulators/utils/readDeck.cpp
+++ b/opm/simulators/utils/readDeck.cpp
@@ -181,15 +181,14 @@ namespace {
         wtestState = std::make_unique<Opm::WellTestState>();
     }
 
-    std::shared_ptr<Opm::Deck>
+    Opm::Deck
     readDeckFile(const std::string&       deckFilename,
                  const bool               checkDeck,
                  const Opm::Parser&       parser,
                  const Opm::ParseContext& parseContext,
                  Opm::ErrorGuard&         errorGuard)
     {
-        auto deck = std::make_shared<Opm::Deck>
-            (parser.parseFile(deckFilename, parseContext, errorGuard));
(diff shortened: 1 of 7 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this opm-common=3225 please

akva2:
jenkins build this please

opm/simulators/utils/readDeck.cpp
```diff
@@ -235,33 +233,31 @@ namespace {
...
                       "or summaryConfig are not initialized");
         }
 
         auto parser = Opm::Parser{};
-        if (deck == nullptr) {
-            deck = readDeckFile(deckFilename, checkDeck, parser,
-                                *parseContext, errorGuard);
-        }
+        auto deck = readDeckFile(deckFilename, checkDeck, parser,
+                                 *parseContext, errorGuard);
```
bska, line 246: Can we mark this as `const`?  If so, I think we should.

bska (commented):
Great job!  This will hopefully be a noticeable reduction in the simulator's memory usage in normal operation.  Just one small nit and then I'm ready to merge this PR.

opm/simulators/utils/readDeck.cpp
akva2: indeed.

bska (approved):
Thanks a lot for the update.  I'm going to merge this into master now.

PR #4188, opened 2022-10-19
Put well BHP/THP calculations in separate class

akva2:
This splits out the Wellinterface/StandardWell/Multisegment bhp/thp related calculations to a separate class.

PR #4288, opened 2022-11-23
make sure getTHPConstratint() consistent between the WellInterfaceGeneric and WellBhpThpCalculator

GitPaean:
potentially a fallout introduced from the PR https://github.com/OPM/opm-simulators/pull/4188 , thanks @akva2 for spotting the potential problem. 

There is still some testing needs to clear out for now. So mark the PR as draft for now.

### DIFF
--- a/opm/simulators/wells/WellBhpThpCalculator.cpp
+++ b/opm/simulators/wells/WellBhpThpCalculator.cpp
@@ -66,18 +66,7 @@ WellBhpThpCalculator::wellHasTHPConstraints(const SummaryState& summaryState) co
 
 double WellBhpThpCalculator::getTHPConstraint(const SummaryState& summaryState) const
 {
-    const auto& well_ecl = well_.wellEcl();
-    if (well_ecl.isInjector()) {
-        const auto& controls = well_ecl.injectionControls(summaryState);
-        return controls.thp_limit;
-    }
-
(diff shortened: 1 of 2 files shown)


GitPaean:
jenkins build this please

akva2:
This does not address the regression. The regression is that https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/WellBhpThpCalculator.cpp#L171 calls the implementation in the WellBhpThpCalculator,
while it should call the one in the WellInterface, https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/WellInterfaceGeneric.cpp#L208

It thus misses the check for the dynamic THP limit. I think the proper fix is to pass thp_limit as a parameter to WellBhpThpCalculator::computeBhpAtThpLimitProd, and give the parameter the value from WellInterface::getTHPConstraint.

GitPaean:
Thanks for the comment while I am not totally following.  If you can draft a PR, it will be much appreciated. @akva2

GitPaean:
Because there is my confusion from my side during testing.  So it is good if you can draft it since you understand what is happening here better.

akva2:
https://github.com/OPM/opm-simulators/pull/4289

GitPaean:
closing this in favor of #4289.

PR #4289, opened 2022-11-23
fix regression in wells

akva2:
we need to call WellInterface::getTHPConstraint,
not WellBhpThpCalculator::getTHPConstraint

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1918,6 +1918,7 @@ namespace Opm
                                         this->maxPerfPress(ebos_simulator),
                                         this->getRefDensity(),
                                         alq_value,
+                                        this->getTHPConstraint(summary_state),
                                         deferred_logger);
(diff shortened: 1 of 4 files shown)


GitPaean:
thanks for the PR. If I did not see it wrong, it looks like for producers, we should check the dynamic_thp_limit from the network nodes. For injectors, we just need to check the thp limit from the EclipseState. 

Is this what this PR is trying to do? Thanks.

akva2:
yes, basically. that's what was done before.

GitPaean:
Thanks for the clarification. But I think #4288 should also fix it if we do not give dynamic_thp_limit for injectors (which should be case). Correct me if I am wrong there. This is somehow I got confused with.

Anyway, we will try to get this PR in after confirmation with testing.

GitPaean:
jenkins build this please

GitPaean:
The PR works as intended. Getting the PR in when jenkins is done.

GitPaean:
jenkins build this please

PR #4292, opened 2022-11-24
Add support for gas in water

totto82:
https://github.com/OPM/opm-material/pull/543

### DIFF
--- a/ebos/equil/initstateequil.cc
+++ b/ebos/equil/initstateequil.cc
@@ -280,7 +280,7 @@ density(const double depth,
 {
     // The initializing algorithm can give depths outside the range due to numerical noise i.e. we extrapolate
     double saltConcentration = saltVdTable_.eval(depth, /*extrapolate=*/true);
-    double rho = FluidSystem::waterPvt().inverseFormationVolumeFactor(pvtRegionIdx_, temp_, press, saltConcentration);
+    double rho = FluidSystem::waterPvt().inverseFormationVolumeFactor(pvtRegionIdx_, temp_, press, 0.0 /*=Rsw*/, saltConcentration);
     rho *= FluidSystem::referenceDensity(FluidSystem::waterPhaseIdx, pvtRegionIdx_);
     return rho;
 }
@@ -1583,8 +1583,9 @@ applyNumericalAquifers_(const GridView& gridView,
         const auto search = num_aqu_cells.find(cartIx);
         if (search != num_aqu_cells.end()) {
             // numerical aquifer cells are filled with water initially
-            // for co2store the oilphase is used for brine
-            const auto watPos = co2store? FluidSystem::oilPhaseIdx : FluidSystem::waterPhaseIdx;
+            // for co2store the oilphase may be used for the brine
+            bool co2store_oil_as_brine = co2store && FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx);
+            const auto watPos =  co2store_oil_as_brine? FluidSystem::oilPhaseIdx : FluidSystem::waterPhaseIdx;
             if (FluidSystem::phaseIsActive(watPos)) {
                 this->sat_[watPos][elemIdx] = 1.;
             } else {
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -84,6 +84,10 @@ namespace Opm
             OPM_THROW(std::runtime_error, "dissolved gas/ vapporized oil in injected oil/gas not supported by multisegment well yet."
             << " \n See  (WCONINJE item 10 / WCONHIST item 8)");
         }
+        if constexpr (!Indices::oilEnabled && Indices::numPhases > 1) {
+            OPM_THROW(std::runtime_error, "water + gas case not supported by multisegment well yet");
+        }
+
     }
 
 
@@ -1546,6 +1550,7 @@ namespace Opm
 
         auto& ws = well_state.well(this->index_of_well_);
         ws.dissolved_gas_rate = 0;
+        ws.dissolved_gas_rate_in_water = 0;
         ws.vaporized_oil_rate = 0;
         ws.vaporized_wat_rate = 0;
 
--- a/opm/simulators/wells/RateConverter.hpp
+++ b/opm/simulators/wells/RateConverter.hpp
@@ -104,6 +104,8 @@ namespace Opm {
                     ra.rv = 0.0;
                     ra.pv = 0.0;
                     ra.saltConcentration = 0.0;
+                    ra.rsw = 0.0;
+                    ra.rvw = 0.0;
 
                 }
 
@@ -162,6 +164,12 @@ namespace Opm {
                             attr.temperature += fs.temperature(FluidSystem::gasPhaseIdx).value() * hydrocarbonPV;
                         }
                         attr.saltConcentration += fs.saltConcentration().value() * hydrocarbonPV;
+                        if (FluidSystem::enableDissolvedGasInWater()) {
+                            attr.rsw += fs.Rsw().value() * hydrocarbonPV; // scale with total volume?
+                        }
+                        if (FluidSystem::enableVaporizedWater()) {
+                            attr.rvw += fs.Rvw().value() * hydrocarbonPV; // scale with total volume?
+                        }
                     }
 
                     if (pv_cell > 0.) {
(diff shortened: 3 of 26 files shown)


totto82:
Combined the PRs make it possible to run a CO2STORE case with WATER + GAS and get the same results as OIL + GAS.

image
image
image

totto82:
Note. The approach is not yet tested for MSW.

totto82:
Depends on https://github.com/OPM/opm-common/pull/3217 and https://github.com/OPM/opm-material/pull/543 and OPM/opm-models#763

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -841,6 +841,8 @@ namespace Opm
     computePerfRate(const Value& pressure_cell,
                     const Value& rs,
                     const Value& rv,
+                    const Value& rsw,
+                    const Value& rvw,
```
bska, line 845: These parameters appear to be unused.  Is that temporary, in which case they could be marked as `[[maybe_unused]]`, or permanent, in which case they could just be removed, or a programming error?

bska (commented):
I've not reviewed this in any detail, but I notice one potential issue.

opm/simulators/wells/MultisegmentWell_impl.hpp
plgbrts: To enable rvw for MSW had no priority so we did not go further into that and is not in our planning at the moment. The combination rvw and MSW will for now give a throw. The rvw is in particular relevant in combination with brine (to model salt precipitation) and brine is also not supported by MSW.

totto82: Ok, I have removed the MSW part of the PR. Adding support for MSW + WATER + GAS needs to be done first and in a separate PR. Since MSW + CO2STORE is probably not the most used combination, this PR should be merged first IMO, and the MSW part added later.

opm/simulators/wells/StandardWellConnections.cpp
```diff
@@ -232,34 +233,44 @@ computeDensities(const std::vector<Scalar>& perfComponentRates,
...
-               x[waterpos] = mix[waterpos] - mix[gaspos] * rvw;
+            Scalar rsw = 0.0;
+            if (!rswmax_perf.empty() && mix[waterpos] > 1e-12) {
+                rsw = std::min(mix[gaspos]/mix[waterpos], rswmax_perf[perf]);
+            }
+            const Scalar d = 1.0 - rsw*rvw;
+            if (d <= 0.0) {
+                std::ostringstream sstr;
+                sstr << "Problematic d value " << d << " obtained for well " << well_.name()
+                    << " during ccomputeConnectionDensities with rsw " << rsw
```
atgeirr, line 259: Typo, "ccompute...", also on line 220.

opm/simulators/wells/StandardWellConnections.cpp
```diff
@@ -330,8 +343,22 @@ computePropertiesForPressures(const WellState& well_state,
...
+                    const double gasrate = std::abs(ws.surface_rates[pu.phase_pos[Gas]]) - (Indices::enableSolvent ? ws.sum_solvent_rates() : 0.0);
+                    if (gasrate > 0) {
+                        rsw = waterrate / gasrate;
+                    }
+                    rsw = std::min(rsw, rswmax_perf[perf]);
+                }
+            }
             b_perf[ waterCompIdx + perf * well_.numComponents()] =
-            FluidSystem::waterPvt().inverseFormationVolumeFactor(region_idx, temperature, p_avg, saltConcentration);
+            FluidSystem::waterPvt().inverseFormationVolumeFactor(region_idx, temperature, p_avg, rsw, saltConcentration);
```
atgeirr, line 361: Misleading indentation.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -290,12 +301,20 @@ namespace Opm
...
+                const unsigned waterCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::waterCompIdx);
+                const unsigned gasCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx);
+                const Value cq_sWat = cq_s[waterCompIdx];
+                const Value cq_sGas = cq_s[gasCompIdx];
+                const Value vap_wat = rvw * cq_sGas;
+                const Value dis_gas_wat = rsw * cq_sWat;
+                cq_s[waterCompIdx] += vap_wat;
+                cq_s[gasCompIdx]   += dis_gas_wat;
+                if (this->isProducer()) {
+                    perf_vap_wat_rate = getValue(vap_wat);
```
atgeirr, line 316: I think it is good that you combine these two features in this one block. To be clear: this is a bugfix for vaporized water, right? (Moving it out of the block requiring oil to be an active phase.)

atgeirr (changes requested):
A few cases of using zero dissolution ratio that should be addressed, otherwise I think it looks good.

ebos/equil/initstateequil.cc
```diff
@@ -280,7 +280,7 @@ density(const double depth,
 {
     // The initializing algorithm can give depths outside the range due to numerical noise i.e. we extrapolate
     double saltConcentration = saltVdTable_.eval(depth, /*extrapolate=*/true);
-    double rho = FluidSystem::waterPvt().inverseFormationVolumeFactor(pvtRegionIdx_, temp_, press, saltConcentration);
+    double rho = FluidSystem::waterPvt().inverseFormationVolumeFactor(pvtRegionIdx_, temp_, press, 0.0 /*=Rsw*/, saltConcentration);
```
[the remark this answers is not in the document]
totto82: We don't support EQUIL initialization with GAS + WATER. I started here https://github.com/OPM/opm-simulators/pull/4316 But we also need to add RSW vs depth table. So this 0.0 will be replaced by values from RSW when that is ready.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -82,6 +82,10 @@ namespace Opm
             OPM_THROW(std::runtime_error, "dissolved gas/ vapporized oil in injected oil/gas not supported by multisegment well yet."
             << " \n See  (WCONINJE item 10 / WCONHIST item 8)");
         }
+        if constexpr (!Indices::oilEnabled && Indices::numPhases > 1) {
+            OPM_THROW(std::runtime_error, "water + gas case not supported by multisegment well yet");
```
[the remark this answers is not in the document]
totto82: I didn't know that either, so I naively started to add RSW, but stopped when I noticed. Adding GAS + WATER + MSW should definitely be added on some(ones) TODO list. But I will argue that it should be done outside this PR.

opm/simulators/wells/RateConverter.hpp
```diff
@@ -280,7 +280,7 @@ namespace Opm {
                 if (RegionAttributeHelpers::PhaseUsed::water(pu)) {
                     // q[w]_r = q[w]_s / bw
 
-                    const double bw = FluidSystem::waterPvt().inverseFormationVolumeFactor(pvtRegionIdx, T, p, saltConcentration);
+                    const double bw = FluidSystem::waterPvt().inverseFormationVolumeFactor(pvtRegionIdx, T, p, 0.0 /*=Rsw*/, saltConcentration);
```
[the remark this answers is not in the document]
totto82: Partly a temporary hack. The Rate converter is used for RESV controlled wells, but I was not  sure how this should be understood for a case with gas in water. Should it be the same as for gas in oil and oil in gas. Logically I think it should be included in RESV, but comments are welcome. I also noticed that rvw is missing from this file. I can add both if we agree that this is what we want.

opm/simulators/wells/StandardWell_impl.hpp
totto82: Yes I think so, but I have not checked. The perf_vap_wat_rate seems to be recomputed for producers at the end of this method. Maybe this whole block should be removed?

atgeirr: Seems correct, assuming you mean lines 317-319.

atgeirr (approved):
With the changes discussed and already made, this is good to be merged.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -407,10 +449,12 @@ namespace Opm
                         perf_vap_wat_rate = getValue(rvw) * (getValue(cq_s[gasCompIdx]) - getValue(rs) * getValue(cq_s[oilCompIdx])) / d;
                     }
                 }
-                else if (FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx) && FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx)) {
+                if (FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx) && FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx)) {
```
GitPaean, line 452: removing `else` here make it not exclude the oil phase anymore. Will it affect the correctness of the code? @totto82

totto82: I think @atgeirr and I agree that this change was correct even though it affect the results. Do you think otherwise?

PR #4293, opened 2022-11-25
added: StandardWellConnections

akva2:
One more class.


akva2:
jenkins build this please

bska (approved):
Very good.  Composition is usually preferable to inheritance, so I approve of this direction.  I'll merge into master.

PR #4301, opened 2022-11-30
MultisegmentWell: avoid mutable matrices/vectors

akva2:
No real reason for this, other than wrongly categorized member functions.


akva2:
jenkins build this please

PR #4304, opened 2022-12-01
using correct CPI indices for connections in WellState::reportConnections()

GitPaean:
Thanks @bska for the help. It caused a lot of confusion during the development of the keyword WINJMULT. 

@bska , I am pretty shallow when coming to this part of the code, if you have different ideas on how it should be done, go head to come up with a different suggestion.

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -569,7 +569,8 @@ void WellState::reportConnections(std::vector<data::Connection>& connections,
     }
     for( auto& comp : connections) {
         const auto * rates = &perf_data.phase_rates[np*local_comp_index];
-        const auto& connPI  = perf_data.prod_index;
+        const auto * connPI  = &perf_data.prod_index[np*local_comp_index];
+
 
         for( int i = 0; i < np; ++i ) {
             comp.rates.set( phs[ i ], rates[i] );
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -550,7 +550,6 @@ void WellState::reportConnections(std::vector<data::Connection>& connections,
     }
 
     const int np = pu.num_phases;
-    size_t local_comp_index = 0;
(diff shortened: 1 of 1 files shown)


opm/simulators/wells/WellState.cpp
```diff
@@ -569,7 +569,8 @@ void WellState::reportConnections(std::vector<data::Connection>& connections,
     }
     for( auto& comp : connections) {
         const auto * rates = &perf_data.phase_rates[np*local_comp_index];
```
GitPaean, line 571: It does looks like it is related to completion here, although we use `local_comp_index`.

GitPaean:
jenkins build this please

bska (approved):
Thanks Kai, this is how I would write it too.  Tangentially I agree that `local_comp_index` gives the wrong association.  It's really the local connection index here, so if you'd like fee free to rename it to `local_conn_index` or something similar.

bska:
> It's really the local connection index here, so if you'd like fee free to rename it to `local_conn_index` or something similar.

If you'd like to do that I'll wait.  Otherwise, I think we should merge this as soon as possible.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
> If you'd like to do that I'll wait. Otherwise, I think we should merge this as soon as possible.

Thanks, I am done updating.  And also, it looks like this is not well tested in the regression test, maybe that is why we did not see it before.   We should make sure it is tested in the regression test.

GitPaean:
> Agreed. Maybe we could add the summary vector

I think it is a really good idea. Let us do it with separate PRs to get this PR in quick. Since you have targeted the case you want to use, maybe you can do it?

bska:
> > Agreed. Maybe we could add the summary vector
> 
> I think it is a really good idea. Let us do it with separate PRs to get this PR in quick.

:+1: The build check is green here so I'll merge this PR now.

PR #4305, opened 2022-12-02
Enable Regression Testing of WELPI and CPI/CTFAC

bska:
Forgotten at the time we developed the WELPI support.


bska:
Note: I'm creating this in "draft" mode for two reasons

1. I'd really like to have the `CPI` summary vector added in OPM/opm-tests#865 be part of the regression test
2. I need to verify that this case compares reasonably well with a reference simulation

I'd also like to enable comparing restarted simulations, but those don't work in local testing and I need to figure out why first.

bska:
jenkins build this opm-tests=865 please

bska:
We have verified that the simulator results agree with a reference solution and the `CPI` summary vector has been included in the model.  I'm going to mark this as "ready for review" now.

bska:

bska:
jenkins build this opm-tests=870 please

bska:
The reference solution has been installed on the CI system.  Merging this now to activate regression tests for `WELPI` and `CPI`.

Issue #2770, opened 2020-09-07
SurfaceToReservoirVoidage might compute wrong results or deadlock in parallel.


blattms:
The [for-loop](/OPM/opm-simulators/blob/53005c477d981b4bdf2913a95ab0c876fbeeb738/opm/simulators/wells/RateConverter.hpp#L508-L529) in `SurfaceToReservoirVoidage::defineState`` in parallel silently assumes that the unordered map of all active regions (result of `RegionMapping::activeRegions` is the same on all processes. Unfortunately, as RegionMapping is currently only a local view this is only the case if each process has at least one cell of all regions in his domain. Otherwise the ordering might be different (and summation will be wrong) or the program will deadlock if the number of regions differs from process to process.

blattms:
I should add that it seems like this currently would only happen if we support parallel computations with domains that have no cells. At least the only way I see the RateConverter being initialized currently is via `RateConverter(phase_usage_, std::vector<int>(number_of_cells_, 0))` at [BlackoilWellModel_impl.hpp](/OPM/opm-simulators/blob/53005c477d981b4bdf2913a95ab0c876fbeeb738/opm/simulators/wells/BlackoilWellModel_impl.hpp#L300-L301), which means there is always only 1 region.

bska:
@blattms : Is this still a problem?

blattms:
Still an issue

PR #4308, opened 2022-12-06
Allow empty partitions

akva2:
For testing purposes (zoltan acting up), it was requested that flow can run with empty grid partitions. This is the start of that work.

Downstream of https://github.com/OPM/opm-grid/pull/606


blattms:
Great, then we will get a fix for #2770 if that still applies.

PR #4310, opened 2022-12-07
BUGFIX we now use group_info in stage 2 and need to make sure its communicated

totto82:


totto82:
jenkins build this please

PR #4311, opened 2022-12-08
Capture Component Names in Independent Datastructure

bska:
This is in preparation of enabling separate convergence output for diagnostic purposes.

### DIFF
--- a/opm/simulators/flow/BlackoilModelEbos.hpp
+++ b/opm/simulators/flow/BlackoilModelEbos.hpp
@@ -59,13 +59,14 @@
 #include <dune/common/timer.hh>
 #include <dune/common/unused.hh>
 
+#include <algorithm>
 #include <cassert>
 #include <cmath>
-#include <iostream>
 #include <iomanip>
+#include <iostream>
 #include <limits>
+#include <utility>
 #include <vector>
-#include <algorithm>
 
 namespace Opm::Properties {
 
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

opm/simulators/flow/BlackoilModelEbos.hpp
```diff
@@ -1065,6 +1092,8 @@ namespace Opm {
         BVector dx_old_;
 
         std::vector<StepReport> convergence_reports_;
+        std::optional<ComponentName> compNames_{};
```
akva2, line 1095: Optional to achieve delayed init? not sure optional is the right type to communicate intent..

bska: > Optional to achieve delayed init?

Yes, mostly because that matches the behaviour of the existing code (`static` vector, initialised first time through `getReservoirConvergence()`.  I can switch to putting this information in a standard object if you'd like.

akva2: I do think that is better, all info is known at construction time. Not insisting though if you are of a different opinion.

bska: > I do think that is better, all info is known at construction time.

Agreed.  I've pushed an update to that effect.

bska:
jenkins build this please

opm/simulators/flow/BlackoilModelEbos.hpp
akva2: Great. Finally, are we the people's front of judea or the judean people's front (ie should immutable class members be marked const or not).

akva2 (approved):
Looks good to me.

opm/simulators/flow/BlackoilModelEbos.hpp
bska: > should immutable class members be marked `const` or not

[Yes...](https://www.reddit.com/r/InclusiveOr/)  More seriously, It Depends[:tm:]  My personal default/preference is to **not** mark data members as `const`, even if the member is immutable.  That preference is guided by the fact that `const` data members inhibit move semantics, or at minimum generate throwing move operations.  If I don't have a reason for creating a non-movable type I favour having the move operations behave in the common way.  I do prefer to make member functions be `const` however.

bska:
jenkins build this please

bska:
Build check is green and PR is approved.  I'll merge into master.

PR #4312, opened 2022-12-08
Capture Timestep's Non-Linear Convergence History

bska:
This enables outputting a formatted record of the limiting MB and CNV quantities as time and non-linear iterations progress.  This, in turn, is intended for diagnostic and analysis purposes and will not be output unless specifically requested.

In particular, add a new type,
```
ConvergenceReport::ReservoirConvergenceMetric
```
which captures the convergence metric type (MB or CNV) along with the associate phase and numerical value of the convergence metric. We add a vector of these convergence metric objects as a new data member of the `ConvergenceReport`.

Finally, foreshadowing the intended use case, also store the report time in the `ConvergenceReport` object.

### DIFF
--- a/opm/simulators/flow/BlackoilModelEbos.hpp
+++ b/opm/simulators/flow/BlackoilModelEbos.hpp
@@ -888,7 +888,8 @@ namespace Opm {
             return grid_.comm().sum(errorPV);
         }
 
-        ConvergenceReport getReservoirConvergence(const double dt,
+        ConvergenceReport getReservoirConvergence(const double reportTime,
+                                                  const double dt,
                                                   const int iteration,
                                                   std::vector<Scalar>& B_avg,
                                                   std::vector<Scalar>& residual_norms)
@@ -927,7 +928,7 @@ namespace Opm {
             }
 
             // Create convergence report.
-            ConvergenceReport report;
+            ConvergenceReport report{reportTime};
             using CR = ConvergenceReport;
             for (int compIdx = 0; compIdx < numComp; ++compIdx) {
                 double res[2] = { mass_balance_residual[compIdx], CNV[compIdx] };
@@ -953,6 +954,7 @@ namespace Opm {
                     } else if (res[ii] > tol[ii]) {
                         report.setReservoirFailed({types[ii], CR::Severity::Normal, compIdx});
                     }
+                    report.setReservoirConvergenceMetric(types[ii], compIdx, res[ii]);
                 }
             }
 
@@ -1003,7 +1005,9 @@ namespace Opm {
         {
             // Get convergence reports for reservoir and wells.
             std::vector<Scalar> B_avg(numEq, 0.0);
-            auto report = getReservoirConvergence(timer.currentStepLength(), iteration, B_avg, residual_norms);
+            auto report = getReservoirConvergence(timer.simulationTimeElapsed(),
+                                                  timer.currentStepLength(),
+                                                  iteration, B_avg, residual_norms);
             report += wellModel().getWellConvergence(B_avg, /*checkWellGroupControls*/report.converged());
 
             return report;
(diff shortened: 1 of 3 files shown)


bska:
This PR depends on, and contains, #4311.  It must not be merged until the upstream prerequisite PR is merged into the master branch.  I am therefore creating the PR in "draft" mode and will keep it that way until it is ready for review.

bska:
jenkins build this please

bska:
The upstream PR has been merged.  I'm marking this as "ready for review".

bska:
jenkins build this please

akva2 (approved):

bska:
jenkins build this please

bska:
PR approved and build check is green.  I'll merge this into master.

OPMUSER:
@bska How does one request this additional diagnostic information? 

Secondly, how is this different to figures 2.40 and 2.41 in the 2022-10 Rev-A version of the manual? A simple screen capture would be suffice.

bska:
> How does one request this additional diagnostic information?

There will be a new command-line option, tentatively named `--extra-convergence-output`, which takes a string argument.  I am working on this right now.

> how is this different to figures 2.40 and 2.41 in the 2022-10 Rev-A version of the manual

It's the same information, but will the information will be output to a separate, machine-readable file (`CASE.INFOITER`).  For what it's worth, we expect that most users will not need this information&mdash;which is why the default setting will be to not output the `.INFOITER` file&mdash;but that developers working on for instance non-linear solvers might find it useful.

OPMUSER:
> It's the same information, but will the information will be output to a separate, machine-readable file (CASE.INFOITER). For what it's worth, we expect that most users will not need this information—which is why the default setting will be to not output the .INFOITER file—but that developers working on for instance non-linear solvers might find it useful

Well another separate file is not a good idea, as we are suppose to get rid of of CASE.INFOSTEP, it is on the To Do List.

If there are howls of objections to the new file, why not just dump it to a temporary file and concatenate it to the DBG file at the end of the run? Alternatively, you could write the data to the SUMMARY file instead, as an OPM Flow Simulation Performance variable (11.2.23) 

To be clear, I don't have a strong opinion on this, but I know others do.

bska:
> > It's the same information, but will the information will be output to a separate, machine-readable file (`CASE.INFOITER`).
> 
> Well another separate file is not a good idea, as we are suppose to get rid of of CASE.INFOSTEP, it is on the To Do List.

We're not getting rid of `.INFOSTEP`.  We'll switch to not outputting the file by default, but we're not dropping the file altogether.

> If there are howls of objections to the new file

We've discussed this topic with a number of parties and there are no objections to the new file, provided we do not output the file by default.

> why not just dump it to a temporary file and concatenate it to the DBG file at the end of the run?

That would lose the "machine readable" aspect of this information.

> Alternatively, you could write the data to the SUMMARY file instead, as an OPM Flow Simulation Performance variable (11.2.23)

We might consider doing that, but please keep in mind that we're capturing this information at each non-linear iteration, not just the final, converged solution at the end of a timestep.  The summary file isn't really set up to handle that kind of vector.

> I know others [have strong opinions on the topic of additional output files]

As I said, we've discussed this topic with a number of other parties, including those who run the simulator in production, and there's no objection to having an additional output file provided we do not create that file by default.

alfbr:
> We're not getting rid of `.INFOSTEP`. We'll switch to not outputting the file by default, but we're not dropping the file altogether.

Thanks for the information, this is very welcome, as the file today pollutes the folder for ordinary users. Ideally, the information would be output to the DBG file, as that file was supposed to be available for developer oriented output. I believe simply outputting the information in the DBG file, prefixing each line with INFOITER. would make it machine readable and always available with no hassle. However, outputting both as optional files is better than the current state where INFOSTEP is always outputted, so for me it is a compromise I live well with.

PR #4314, opened 2022-12-10
add missing include

akva2:
after upstream header removals.

Downstream of https://github.com/OPM/opm-common/pull/3244


bska (approved):

PR #4316, opened 2022-12-12
Suppot EQUIL + WATER + GAS

totto82:
Seems like most of the work was already done. @bska Did I miss something? Only tested on a simple case without capillary pressure. So definitely it needs more testing.

### DIFF
--- a/ebos/equil/initstateequil.cc
+++ b/ebos/equil/initstateequil.cc
@@ -515,6 +515,13 @@ typename PressureTable<FluidSystem,Region>::Strategy
 PressureTable<FluidSystem,Region>::
 selectEquilibrationStrategy(const Region& reg) const
 {
+    if (!this->oilActive()) {
+        if (reg.datum() > reg.zwoc()) { // Datum in water zone
+            return &PressureTable::equil_WOG;
+        }
+        return &PressureTable::equil_GOW;
+    }
+
     if (reg.datum() > reg.zwoc()) {      // Datum in water zone
         return &PressureTable::equil_WOG;
     }
@@ -636,12 +643,14 @@ void PhaseSaturations<MaterialLawManager, FluidSystem, Region, CellID>::deriveGa
     auto& sg = this->sat_.gas;
 
     const auto isIncr = true; // dPcgo/dSg >= 0 for all Sg.
+    const auto oilActive = this->evalPt_.ptable->oilActive();
 
     if (this->isConstCapPress(this->gasPos())) {
         // Sharp interface between phases.  Can derive phase saturation
         // directly from knowing where 'depth' of evaluation point is
         // relative to depth of O/G contact.
-        sg = this->fromDepthTable(this->evalPt_.region->zgoc(),
+        const auto gas_contact = oilActive? this->evalPt_.region->zgoc() : this->evalPt_.region->zwoc();
+        sg = this->fromDepthTable(gas_contact,
                                   this->gasPos(), isIncr);
     }
     else {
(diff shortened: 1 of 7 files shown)


totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
When testing the gas+water initialization I noticed that the results form the gas+oil initialization seemed wrong. (I think it used the wrong cap-pressure for the endpoints). 

To fix this I changed `accountForScaledSaturations` which means that this PR will change the behavior for more than the water + gas case. I may have missed something (or misunderstood somethings) so comments are welcome.

totto82:
jenkins build this please

bska:
jenkins build this please

totto82:
sorry for updating this again and again. I was editing the data file in the build directory... Now the equil test should not fail...

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
> To fix this I changed accountForScaledSaturations which means that this PR will change the behavior for more than the water + gas case. I may have missed something (or misunderstood somethings) so comments are welcome.

After fixing the fix that fixed the 2p case also the 3p case is working and all tests are fine.

plgbrts:
I ran a two phase gas-water case with nonzero capillary pressure with the EQUIL keyword and compared Flow with E100.
Flow:
image
E100:
image

The behavior is quite different but could not find an inconsistency in the code. Maybe good to have a similar comparison for an oil-gas case to see what is to be expected? 

The gas water case used for the comparison: 
[SPE1CASE2_GASWATER_EQUIL.zip](https://github.com/OPM/opm-simulators/files/10403738/SPE1CASE2_GASWATER_EQUIL.zip)

ebos/equil/initstateequil.cc
```diff
@@ -732,75 +741,106 @@ accountForScaledSaturations()
...
+        .oilWaterScaledEpsInfoDrainage(this->evalPt_.position->cell);
 
     const auto thresholdSat = 1.0e-6;
     if (watActive && ((sw + thresholdSat) > scaledDrainageInfo.Swu)) {
         // Water saturation exceeds maximum possible value.  Reset oil phase
         // pressure to that which corresponds to maximum possible water
         // saturation value.
         this->fluidState_.setSaturation(this->waterPos(), scaledDrainageInfo.Swu);
+        if (oilActive) {
+            this->fluidState_.setSaturation(this->oilPos(), so - sw + scaledDrainageInfo.Swu);
```
bska, line 764: How does this correctly adjust the oil saturation?  Suppose, for example, that inverting the capillary pressure curves gives
```
(Sw, So, Sg) = (0.4, 0.5, 0.1)
```
but that `Swu = 0.3`.  If we then reduce `Sw` to `Swu`, then I would **typically** expect the surplus 0.1 to be distributed among the other two phases, and if we assign the surplus entirely to the oil, then we **should** have
```
(Sw, So, Sg) = (0.3, 0.6, 0.1)
```
On the other hand, this expression gives
```
So = 0.5 - 0.4 + 0.3 = 0.4
```
from which we get the saturation distribution
```
(Sw, So, Sg) = (0.3, 0.4, 0.1)
```
which does not sum to one instead.  What am I missing?

bska (commented):
This looks good to me for the most part, except for the saturation updates in `accountForScaledSaturations()` which I don't fully comprehend.  I also think we should rename one of the intermediate variables to reflect the common case.

totto82:
>The behavior is quite different but could not find an inconsistency in the code. Maybe good to have a similar comparison for an oil-gas case to see what is to be expected?

Do we compare two different pressures? OPM Flow will output the water phase pressure for a gas-water case. The pressure shown for OPM seems to be 4700 - pc(sw). Have capillary pressure + gas + water, with explicit initialization been tested? This may not be related to the initialization.

ebos/equil/initstateequil.cc
totto82: > What am I missing?

Just that I forgot to flip signs. I am surprised that this bug in the PR did not affect any of the tests.

totto82:
jenkins build this please

totto82:
With #4372  The pressure is similar to the E100 results @plgbrts showed.

bska (approved):

ebos/equil/initstateequil.cc
bska: > I am surprised that this bug in the PR did not affect any of the tests.

I am too.  On the other hand, It's **possible** that the tests don't need this kind of adjustment at all, and that if they actually do then they mostly care about the water saturation and implicitly assume `So = 1 - Sw`.

PR #4320, opened 2022-12-13
Honor MULTZ- when determining TRANSZ over pinched cells (pinch all)

blattms:
Previously, a barrier between cells in Z-direction was only honored if it was done using MULTZ. We disregarded MULTZ- completely when calculating the TRANZ over pinched cells.

Now we also take into account MULTZ-.
Please note that this PR will always does this and not only with this in the deck:
```
GRIDOPTS
'YES'		0 / 
```
Not 100% sure whether this is what we want.

This is to fix an issue with tests in OPM/opm-tests#872.

Needs OPM/opm-common#3272 to work as expected.

### DIFF
--- a/ebos/ecltransmissibility.cc
+++ b/ebos/ecltransmissibility.cc
@@ -621,22 +621,24 @@ applyAllZMultipliers_(Scalar& trans,
         assert(insideFaceIdx==5); // as insideCartElemIdx < outsideCartElemIdx holds for the Z column
         assert(outsideCartElemIdx > insideCartElemIdx);
         auto lastCartElemIdx = outsideCartElemIdx - cartDims[0]*cartDims[1];
-        // Last multiplier
-        Scalar mult = transMult.getMultiplier(lastCartElemIdx , FaceDir::ZPlus);
+        // Last multiplier using (Z+)*(Z-)
+        Scalar mult = transMult.getMultiplier(lastCartElemIdx , FaceDir::ZPlus) *
+            transMult.getMultiplier(outsideCartElemIdx , FaceDir::ZMinus);
 
         if ( !pinchTop )
         {
-            // pick the smallest multiplier for Z+ while looking down the pillar until reaching the other end of the connection
-            // While Z- is not all used here.
-            for(auto cartElemIdx = insideCartElemIdx; cartElemIdx < lastCartElemIdx;
-                cartElemIdx += cartDims[0]*cartDims[1])
+            // pick the smallest multiplier using (Z+)*(Z-) while looking down
+            // the pillar until reaching the other end of the connection
+            for(auto cartElemIdx = insideCartElemIdx; cartElemIdx < lastCartElemIdx;)
             {
-                mult = std::min(mult, transMult.getMultiplier(cartElemIdx, FaceDir::ZPlus));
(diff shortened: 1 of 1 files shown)


OPMUSER:
> Not 100% sure whether this is what we want.

Computer says no:

```
 @--  ERROR  AT TIME        0.0   DAYS    ( 1-JLY-2018):
 @           GRID ARRAY MULTZ-   NOT ALLOWED WITH CURRENT                    
 @           RUNSPEC OPTIONS. SEE KEYWORD GRIDOPTS         
```

blattms:
If this is output from a run with flow, then I actually think this patch might be fine. It means one cannot specifiy MULTZ- in this case, hence it is defaulted to 1 and everything should work like it should. Or am I missing something?

OPMUSER:
> If this is output from a run with flow, then I actually think this patch might be fine. It means one cannot specifiy MULTZ- in this case, hence it is defaulted to 1 and everything should work like it should. Or am I missing something?

Your missing something.

The output is from E100 - so if GRIDOPTS is missing then it is an error, whereas OPM Flow runs. So I think this is an  issue,  as we should behave the same as E100. That is the frst item on GRIDOPTS must be set to YES to enable MULTZ- and the like.

```
--
--       NEG      MAX     MAX                                                    
--       MULTS    MULTNUM PINCHNUM                                               
GRIDOPTS                                                                        
         YES      9       1*                                                   /

```
What do you think @tskille?

tskille:
OPM flow ignores 
```
GRIDOPTS
'YES'		0 / 
```
which I think is fine.  OPM FLow is ignoring a significant part of input from the RUNSPEC sections. We do use some input which are needed for Eclipse compatible restart (restart files), but since Flow have dynamic allocation of memory much of the input in RUNSPEC is not needed. 

Flow runs with results as expected when MULTZ- is used without previously specifying GRIDOPTS in runspec section. 

> Please note that this PR will always does this and not only with this in the deck:
 
I think that it is good that Flow now will always look at both MULTZ and MULTZ-

blattms:
@OPMUSER Judging from the code OPM-flow has always used both multipliers for all directions with the slight exception for the z-direction in the case of pinch all. Here we used the minimum over all pinched faces for MULTZ but only one value for MULTZ-. That is (actually was) clearly inconsistent.

Our manual says something different, though.

OPMUSER:
> Our manual says something different, though.

Unfortunately, unlike my better half I cannot read minds - care to explain in a bit more detail.

I have added this text o the the GRIDOPTS, MULTX-, etc keywords:

`Note that OPM Flow does not require the GRIDOPTS(TRANMULT) parameter in the RUNSPEC section to be set to YES, in order to use this and other negative directional dependent multiplier keywords in the input deck. Whereas, the commercial simulator will terminate with an error if the keyword is present, and the GRIDOPTS(TRANMULT) parameter has not been set to YES.`

blattms:
I need to investigate this further. Somehow MULTZ- looks different to the simulator when using a debug build than whenusing a  default one (no zeros!). That is very unfortunate.

blattms:
:flushed: Well my debug build was with necessary changes to opm-common (see OPM/opm-common#3272) and optimized one was not. My bad, sorry. Undrafting.

bska (approved):

PR #4324, opened 2022-12-13
add missing include after upstream header removal

akva2:
Required by upstream https://github.com/OPM/opm-common/pull/3266 but can be processed before.


bska:
jenkins build this please

PR #4329, opened 2022-12-16
[bugfix,mswells] Throw for connections not attached to segments correctly.

blattms:
[bugfix] Throw for connections not attached to segments .

If there are connections of a multisegment well that are not connected to any segment, we throw with a meaningful error message instead of a silent segmentation fault. In that case an invalid segment index of zero will be returned. The corresponding storage index returned by segmentToNumberIndex becomes -1. This previously lead to segmentation faults when using it to index a container.

~There might be connections of a multisegment well that are not connected to any segment. In that case an invalid segment index of zero will be returned. The corresponding storage index returned by segmentToNumberIndex becomes -1. This will lead to segmentation faults when using it to index a container.~

~With this patch we check whether a connection is part of a segment and only add it to the list of perforations of a segment if it is.~

This is still untested but hopefully resolves an issue we are facing with a model. I will undraft once I have tested it.

### DIFF
--- a/opm/simulators/wells/MultisegmentWellGeneric.cpp
+++ b/opm/simulators/wells/MultisegmentWellGeneric.cpp
@@ -37,6 +37,8 @@
 #include <cmath>
 #include <stdexcept>
 
+#include <fmt/format.h>
+
 namespace Opm
 {
 
@@ -65,6 +67,13 @@ MultisegmentWellGeneric(WellInterfaceGeneric& baseif)
         const Connection& connection = completion_set.get(perf);
         if (connection.state() == Connection::State::OPEN) {
             const int segment_index = segmentNumberToIndex(connection.segment());
+            if ( segment_index == -1) {
+                OPM_THROW(std::logic_error,
+                          fmt::format("COMPSEGS: Well {} has connection in cell {}, {}, {} "
+                                      "without associated segment.", baseif_.wellEcl().name(),
+                                      connection.getI() + 1, connection.getJ() + 1,
+                                      connection.getK() + 1));
+            }
             segment_perforations_[segment_index].push_back(i_perf_wells);
             baseif.perfDepth()[i_perf_wells] = connection.depth();
             const double segment_depth = segmentSet()[segment_index].depth();
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -627,6 +627,14 @@ void WellState::initWellStateMSWell(const std::vector<Well>& wells_ecl,
                 const Connection& connection = completion_set.get(perf);
                 if (connection.state() == Connection::State::OPEN) {
                     const int segment_index = segment_set.segmentNumberToIndex(connection.segment());
+                    if ( segment_index == -1) {
+                        OPM_THROW(std::logic_error,
+                                  fmt::format("COMPSEGS: Well {} has connection in cell {}, {}, {} "
+                                              "without associated segment.", well_ecl.name(),
+                                              connection.getI() + 1 , connection.getJ() + 1,
+                                              connection.getK() + 1 ));
+                    }
+
                     segment_perforations[segment_index].push_back(n_activeperf);
                     n_activeperf++;
                 }


blattms:
jenkins build this please

GitPaean:
Maybe @OPMUSER also has comments regarding the proper way we handle this type of situation? 

Possible scenarios I guess it can be a connection is putting to a non-existing branch or a non-existing segment directly.

bska:
> There might be connections of a multisegment well that are not connected to any segment.

If that happens it's a bug.  As @GitPaean [says](https://github.com/OPM/opm-simulators/pull/4329#issuecomment-1354473246) `COMPSEGS` keyword is supposed to attach every single reservoir connection to a unique well segment.

OPMUSER:
@blattms when I run the multi-segment well regression tests using an old version of E100, I get the following messages for the `~/opm-tests/msw_3d_hfa/3D_MSW.DATA`

```
 @--WARNING  AT TIME        0.0   DAYS    ( 6-NOV-2011):
 @           SEGMENT   3 IN WELL PROD01   HAS A DEPTH CHANGE OF              
 @              26.0  METRES  WHEREAS ITS LENGTH IS ONLY    25.0  METRES .   
 @           THIS IS NOT PHYSICAL.                                           

 @--WARNING  AT TIME        0.0   DAYS    ( 6-NOV-2011):
 @           SEGMENT   4 IN WELL PROD01   HAS A DEPTH CHANGE OF              
 @              33.0  METRES  WHEREAS ITS LENGTH IS ONLY    25.0  METRES .   
 @           THIS IS NOT PHYSICAL.                                         
```
Is it possible to issue a similar warning?

@tskille The WSEGDIMS(MXLINKS) parameter is not used as OPM Flow does not support the WSEGLINK (MULTI-SEGMENT WELL LOOPED FLOW PATHS), should we trap this with a warning?

blattms:
It might be a problem with the model. COMPSEGS  for that well is commented (it would have an entry for that connection) but there is the COMPDAT keyword operating on that well and opening that connection.

OPMUSER:
@blattms 
`It might be a problem with the model. COMPSEGS for that well is commented (it would have an entry for that connection) but there is the COMPDAT keyword operating on that well and opening that connection.`

I don't think so, as the warning messages relates to the `WELSEGS` keyword. This is the input:

```
WELSEGS
    'PROD01' 2512.5 2512.5 1.0e-5 'ABS' 'HFA' 'HO' /
     2         2      1      1    2537.5 2534.5  0.3  0.00010 /
     3         3      1      2    2562.5 2560.5  0.3  0.00010 /
     4         4      1      3    2587.5 2593.5  0.3  0.00010 /
     5         5      1      4    2612.5 2614.5  0.3  0.00010 /
     6         6      1      5    2637.5 2635.5  0.3  0.00010 /    

```
And the computer says:

```
---------------------------------------------------------------------------------
:WELLNAME: SEG : BRN : MAIN  :       : SEGMENT :TOT LENGTH:  DEPTH   :T.V. DEPTH:
:  AND   : NO  : NO  : INLET :OUTLET : LENGTH  :  TO END  :  CHANGE  :  AT END  :
:SEG TYPE:     :     :SEGMENT:SEGMENT: METRES  :  METRES  :  METRES  :  METRES  :
---------------------------------------------------------------------------------
:PROD01  :   1 :   1 :    2  :    0  :  2512.5 :   2512.5 :   2512.5 :   2512.5 :
: HFA  HO:   2 :     :    3  :    1  :    25.0 :   2537.5 :     22.0 :   2534.5 :
:      HO:   3 :     :    4  :    2  :    25.0 :   2562.5 :     26.0 :   2560.5 :
:      HO:   4 :     :    5  :    3  :    25.0 :   2587.5 :     33.0 :   2593.5 :
:      HO:   5 :     :    6  :    4  :    25.0 :   2612.5 :     21.0 :   2614.5 :
:      HO:   6 :     :    0  :    5  :    25.0 :   2637.5 :     21.0 :   2635.5 :

```
So the length is 25m for each segment, so the maximum incremental depth cannot be greater than 25m. Note that the WELSEGS(TUBOPT) parameter is set to ABS. So maybe we just check for when this condition applies.

OPMUSER:
Ops looks like two different threads here, I'm talking about the length and depth inconsistencies on the WELSEGS keyword. Not the completion inconsistency.

GitPaean:
>  It might be a problem with the model. COMPSEGS for that well is commented (it would have an entry for that connection) but there is the COMPDAT keyword operating on that well and opening that connection.

If I understood correctly, you mean the well has WELSEGS defines the segment structure while there is no COMPSEGS to associate the connections (defined with COMPDAT keywords) to the segment.  Then it is not a well defined multi-segment well. We almost can just remove that well.  How to handle this situation, we can either check how the commercial simulator handles it or we come with our own design.  Since the segments (wellbore tube) are not connected to reservoir through connections, my honest opinion is that we can remove the well, or stop the simulation with error messages pointing to the problem.

GitPaean:
>  Ops looks like two different threads here, I'm talking about the length and depth inconsistencies on the WELSEGS keyword. Not the completion inconsistency.

Yes. The length/depth inconsistency is easy to fix. The MSW_3D model was manually written and it was easy to introduce something like that. If you can create a shortcut issue assign to either @GitPaean @blattms or @bska , we will handle it as a general maintenance task.

OPMUSER:
`Maybe @OPMUSER also has comments regarding the proper way we handle this type of situation?`

@GitPaean Hard to tell without having an example of the error. What model is causing the issue?

> Yes. The length/depth inconsistency is easy to handle. The MSW_3D model was manually written and it was easy to introduce something like that. If you can create a shortcut issue assign to either @GitPaean @blattms or @bska , we will handle it as a general maintenance task.

@GitPaean Yes, I know - my head always hurts when I have to deal with multi-segment wells :-) I'll create an issue.

GitPaean:
> @GitPaean Hard to tell without having an example of the error. What model is causing the issue?

It is a commercial model, maybe you can check with @blattms for access. I do not have all the overview.  At the end, the well is created with WELSPECS, WELSEGS, and COMPDAT, and there is no COMPSEGS  for that well. I think you should understand well the scenario now.  No definition how the connections should be associated with segments for that well at all. 

More generally, we specifies the connections through `COMPDAT`, but we never enter some connections in `COMPSEGS` to associate them to the segments. What we should do?

Or we fail to associate some connections to some segments even we input them in `COMPSEGS`, what we should do?

But maybe handle the targeting model first as the highest priority?

@OPMUSER

OPMUSER:
@GitPaean Okay. I'll use the `~/opm-tests/msw_3d_hfa/3D_MSW.DATA` model and create two test cases, one with COMPDAT connection missing and one with COMPSEGS connection missing.

OPMUSER:
**COMPDAT connections missing:**
```
 @--WARNING  AT TIME        0.0   DAYS    ( 6-NOV-2011):
 @           WELL PROD01   DOES NOT HAVE A CONNECTION IN                     
 @           THE GRID BLOCK     10    10     1 WHEN READING COMPSEGS.        
 @           THIS CONNECTION WILL BE IGNORED.                                

 @--WARNING  AT TIME        0.0   DAYS    ( 6-NOV-2011):
 @           WELL PROD01   DOES NOT HAVE A CONNECTION IN                     
 @           THE GRID BLOCK     10    10     2 WHEN READING COMPSEGS.        
 @           THIS CONNECTION WILL BE IGNORED.                                

 @--WARNING  AT TIME        0.0   DAYS    ( 6-NOV-2011):
 @           WELL PROD01   DOES NOT HAVE A CONNECTION IN                     
 @           THE GRID BLOCK     10    10     3 WHEN READING COMPSEGS.        
 @           THIS CONNECTION WILL BE IGNORED.                                

etc.
```
So the job runs in E100.

**COMPSEGS connections missing:**

```
 @--WARNING  AT TIME        0.0   DAYS    ( 6-NOV-2011):
 @           SEGMENT   3 IN WELL PROD01   HAS A DEPTH CHANGE OF              
 @              26.0  METRES  WHEREAS ITS LENGTH IS ONLY    25.0  METRES .   
 @           THIS IS NOT PHYSICAL.                                           

 @--WARNING  AT TIME        0.0   DAYS    ( 6-NOV-2011):
 @           SEGMENT   4 IN WELL PROD01   HAS A DEPTH CHANGE OF              
 @              33.0  METRES  WHEREAS ITS LENGTH IS ONLY    25.0  METRES .   
 @           THIS IS NOT PHYSICAL.                                           
    54 READING COMPSEGS

 @--  ERROR  AT TIME        0.0   DAYS    ( 6-NOV-2011):
 @           NOT ALL CONNECTIONS IN WELL PROD01   HAVE BEEN                  
 @           ASSIGNED WELL SEGMENTS IN KEYWORD COMPSEGS,                     
 @           INCLUDING CONNECTION     10    10     1    
```

So the **job fails** to run

Quick and dirty test decks attached for reference.

[MSW.zip](https://github.com/OPM/opm-simulators/files/10245236/MSW.zip)

GitPaean:
> So the job fails to run

Thanks @OPMUSER for the testing. Basically it means the targeting model will not run with the commercial simulator.  We can do the same to throw a error message similar to above and terminate the running.

blattms:
I will need ti double check whether this holds for this particular model. I silently assumed that it is running with the other simulator.

OPMUSER:
**All COMPSEGS connections missing:**
```
@--  ERROR  AT TIME        0.0   DAYS    ( 6-NOV-2011):
 @           NOT ALL CONNECTIONS IN WELL PROD01   HAVE BEEN                  
 @           ASSIGNED WELL SEGMENTS IN KEYWORD         ,                     
 @           INCLUDING CONNECTION   10  10   1               

```
Whereas, OPM Flow gives my favorite error message

`OPM Flow Process 8669 Segmentation Fault - Job Aborted`

**All COMPDATA connections missing:**
```
 @--WARNING  AT TIME        0.0   DAYS    ( 6-NOV-2011):
 @           KEYWORD COMPSEGS WILL HAVE NO EFFECT ON WELL PROD01             
 @           AS THE WELL HAS NO CONNECTIONS. CHECK THE COMPDAT,              
 @           COMPDATL OR COMPDATM DATA AS APPROPRIATE                        

 @--WARNING  AT TIME        0.0   DAYS    ( 6-NOV-2011):
 @           WELL PROD01   DOES NOT HAVE A CONNECTION IN                     
 @           THE GRID BLOCK     10    10     1 WHEN READING COMPSEGS.        
 @           THIS CONNECTION WILL BE IGNORED.      
........
 @--  ERROR  AT TIME        0.0   DAYS    ( 6-NOV-2011):
 @           CONNECTION DATA FOR WELL PROD01   IS MISSING.                   
 @           THE CONNECTIONS MUST BE DEFINED USING KEYWORD COMPDAT           
 @           BEFORE THE WELL CONTROL KEYWORD (E.G. WCONPROD). 
```

So the job fails to run                

Whereas, OPM Flow runs:

```
===============    End of simulation     ===============

Number of MPI processes:         1
Threads per MPI process:         2
Total time (seconds):            5.54
Solver time (seconds):           5.20
 Assembly time (seconds):        0.88 (Failed: 0.0; 0.0%)
   Well assembly (seconds):      0.00 (Failed: 0.0; -nan%)
 Linear solve time (seconds):    0.41 (Failed: 0.0; 0.0%)
   Linear setup (seconds):       0.08 (Failed: 0.0; 0.0%)
 Update time (seconds):          0.69 (Failed: 0.0; 0.0%)
 Pre/post step (seconds):        1.04 (Failed: 0.0; 0.0%)
 Output write time (seconds):    2.24
Overall Linearizations:        126    (Failed:   0; 0.0%)
Overall Newton Iterations:      84    (Failed:   0; 0.0%)
Overall Linear Iterations:     481    (Failed:   0; 0.0%)
```
Updated test cases attached.
[MSW.zip](https://github.com/OPM/opm-simulators/files/10245461/MSW.zip)

blattms:
@OPMUSER There is a slight difference, though. COMPSEGS is completely commented here and not just one cell.

OPMUSER:
I also ran with COMPDAT and then COMPSEGS fully commented out - see above.

The Isles are calling :-)

blattms:
> The Isles are calling :-)

Then have a nice weekend. :partying_face:

OPMUSER:
@blattms  There is a slight difference, though. COMPSEGS is completely commented here and not just one cell.

I have ran four cases as follows:

1. COMPDAT partially commented out case 01
2. COMPDAT fully commented out - case 03
3. COMPSEGS partially commented out - case 02
4. COMPSEGS fully commented out - case 04

The Isles are calling :-)

opm/simulators/wells/MultisegmentWellGeneric.cpp
```diff
@@ -63,8 +65,15 @@ MultisegmentWellGeneric(WellInterfaceGeneric& baseif)
...
         const Connection& connection = completion_set.get(perf);
-        if (connection.state() == Connection::State::OPEN) {
+        if (connection.state() == Connection::State::OPEN && connection.attachedToSegment()) {
             const int segment_index = segmentNumberToIndex(connection.segment());
+            if ( segment_index == -1) {
+                OPM_THROW(std::logic_error,
+                          fmt::format("COMPSEGS: Well {} has connection in cell {}, {}, {} "
+                                      "without associated segment.", baseif_.wellEcl().name(),
+                                      connection.getI(), connection.getJ(),
+                                      connection.getK()));
```
bska, line 75: Tiny nit: Could you output `getI() + 1` (etc.) instead, please?  The `Connection::getX()` functions return zero-based indices, and users are more accustomed to one-based indices in this context.

bska (commented):
We could possibly try to detect all instances of this problem instead of just the first one.  That might make it easier for the engineer to fix their input deck all in one go.

opm/simulators/wells/MultisegmentWellGeneric.cpp
```diff
@@ -63,8 +65,15 @@ MultisegmentWellGeneric(WellInterfaceGeneric& baseif)
     baseif.perfDepth().resize(baseif_.numPerfs(), 0.);
     for (size_t perf = 0; perf < completion_set.size(); ++perf) {
         const Connection& connection = completion_set.get(perf);
-        if (connection.state() == Connection::State::OPEN) {
+        if (connection.state() == Connection::State::OPEN && connection.attachedToSegment()) {
```
GitPaean, line 68: `connection.attachedToSegment()` will already return `return (segment_number > 0);`. So the condition `if ( segment_index == -1)` will never be reached?

GitPaean:
> We could possibly try to detect all instances of this problem instead of just the first one. That might make it easier for the engineer to fix their input deck all in one go.

Maybe we should do it in the parser, that checks whether all the connections are attached to segment properly, and also terminate the running from there after giving the diagnostic information.   The problem to address it in the simulator is that it might terminate the running after quite some time of running.

bska:
> > We could possibly try to detect all instances of this problem instead of just the first one. That might make it easier for the engineer to fix their input deck all in one go.
> 
> Maybe we should do it in the parser, that checks whether all the connections are attached to segment properly, and also terminate the running from there after giving the diagnostic information.

We can definitely check anything that's known **statically** at the parser level.  The picture becomes a little more muddied when we consider that dynamically triggered `ACTIONX` blocks might introduce new wells and/or connections.  On the other hand I don't think we currently support multisegmented wells in `ACTIONX` blocks.  There's also the wrinkle that if we "only" care about open connections then we potentially need to take `WTEST` and `WECON` into account too.

Still, a parser-level check is a lot better than nothing.

GitPaean:
> The picture becomes a little more muddied when we consider that dynamically triggered ACTIONX blocks might introduce new wells and/or connections

I am not familiar with `ACTIONX` that much, but I think the inconsistency introduced from `ACTIONX` needs to be handled when the simulation reaches the point actually.  It should not stop us from handling the inconsistency introduced by the static part of DATA file.  And also, the facility introduced in the `Well` from the parser can probably also be used for the detecting the problem introduced by `ACTIONX`, or `WTEST`, `WECON`? not totally sure for free though.

opm/simulators/wells/MultisegmentWellGeneric.cpp
blattms: Indeed, fixed. Classic copy & paste without thinking. :flushed:

opm/simulators/wells/MultisegmentWellGeneric.cpp
blattms: Don't even consider this a nit. Fixed.

blattms:
As this seems to be already quite an improvement over silent failure I am marking this ready for review, anyhow.

blattms:
jenkins build this please

bska (approved):
Yep, this is clearly better than silently indexing outside containers.  I'll merge into master.

OPMUSER:
Well I'm sorry to be a Christmas Grinch, but this does not catch an error in one of our regression tests, compared with the commercial simulator that taps the error.

If you run https://github.com/OPM/opm-tests/blob/master/msw_2d_h/2D_H__.DATA in the commercial simulator you get this error:
```
 @--WARNING  AT TIME     2960.0   DAYS    (14-DEC-2019):
 @           SEGMENT   5 IN WELL PROD01   HAS A DEPTH CHANGE OF              
 @           -2537.5  METRES  WHEREAS ITS LENGTH IS ONLY -2737.5  METRES .   
 @           THIS IS NOT PHYSICAL.                                           

 @--  ERROR  AT TIME     2960.0   DAYS    (14-DEC-2019):
 @           SEGMENT   5 IN WELL PROD01   HAS A LENGTH <OR= 0.0              

 @--  ERROR  AT TIME     2960.0   DAYS    (14-DEC-2019):
 @           SEGMENT   6 IN WELL PROD01   HAS A LENGTH <OR= 0.0              

 @--  ERROR  AT TIME     2960.0   DAYS    (14-DEC-2019):
 @           CONNECTIVITY ERROR IN WELSEGS  DATA FOR WELL PROD01  .          
 @              2 SEGMENTS ARE ISOLATED, INCLUDING SEGMENT    6              

```
The offending code is the last completion:

```
--
--       WELL CONNECTION DATA                           
--                                                                              
-- WELL  --- LOCATION ---  OPEN   SAT   CONN   WELL   KH    SKIN   D     DIR   
-- NAME   II  JJ  K1  K2   SHUT   TAB   FACT   DIA    FACT  FACT   FACT  PEN   
COMPDAT                                                                        
'INJE01'   1   1   4   5  'OPEN'  1*    200.   0.5  /
'PROD01'  20   1   1   1  'OPEN'  1*    200.   0.5  /
'PROD01'  20   1   2   2  'OPEN'  1*    200.   0.5  /
'PROD01'  20   1   3   3  'OPEN'  1*    200.   0.4  /
'PROD01'  19   1   2   2  'OPEN'  1*    200.   0.4  /
/
--
--       WELL SEGMENT SPECIFICATION DATA                                                     
--                                                                             
-- WELL  NODAL       LEN     WELL    DEPH    PRESS   FLOW       
-- NAME  DEPTH       TUBING  VOLM    OPTN    CALC    MODEL       
WELSEGS                                             
'PROD01' 2512.5      2512.5  1.0e-5  'ABS'   'H--'   'HO'  /
--                                                     
--       SEG   SEG   BRAN    SEG     TUBING  NODAL   TUBE   TUBE     XSEC  VOL
--       ISTR  IEND  NO      NO      LENGTH  DEPTH   ID     ROUGH    AREA  SEG
          2     2    1       1       2537.5  2537.5  0.3    0.00010           /
          3     3    1       2       2562.5  2562.5  0.2    0.00010           /
          4     4    2       2       2737.5  2537.5  0.2    0.00010           /
/
--
--       MULTISEGMENT WELL COMPLETION SEGMENT SPECIFICATION DATA                                                     
--                                                                             
-- WELL                                                         
-- NAME      
COMPSEGS
'PROD01' /
--                                                                             
--      --LOCATION--  BRAN  TUBING  NODAL   DIR  LOC    MID    COMP    ISEG              
--       II  JJ  K1   NO    LENGTH  DEPTH   PEN  I,J,K  PERFS  LENGTH  NO.              
         20   1   1   1     2512.5  2525.0                                     /
         20   1   2   1     2525.0  2550.0                                     /
         20   1   3   1     2550.0  2575.0                                     /
         19   1   2   2     2637.5  2837.5                                     /
/
```

Can somebody confirm this, as I don't fully understand what is going on here. As I said before, my head hurts when I have to deal with multi-segment wells.

For reference both simulators agree up to where E100 stops.

Field_Production_Comparison_Plot

GitPaean:
I am happy to help to improve this part and interested in doing that. But I can not really prioritize it. Can you create a shortcut story and assign it to me and I will do that after the New year?

PR #4294, opened 2022-11-25
StandardWell: avoid mutable primary variables

akva2:

PR #4331, opened 2022-12-16
MultisegmentWell: avoid mutable primary variables

akva2:
Same as #4294 for MSW.


akva2:
jenkins build this please

bska (approved):
Thanks.  This does better reflect the actual semantics of the class.  I'll merge into master.

PR #4323, opened 2022-12-13
Output Non-Linear Convergence to Separate File if Requested

bska:
This PR enables outputting non-linear convergence metrics, i.e., the MB and CNV values, per phase, for each non-linear iteration in each timestep.  If the user passes the option value `"iterations"` to the `--extra-convergence-output` command line option (PR #4322), this PR will create a new output file, `CASE.INFOITER`, that holds

 * report step
 * time step within that report step
 * elapsed time
 * MB and CNV values per phase
 * well convergence status

for each non-linear iteration.

We use an asynchronous file writing procedure and confer ownership of the report step's unprocessed convergence reports to this procedure just before the end of
```
SimulatorFullyImplicitBlackoilEbos::runStep()
```
At that point, the convergence reports are about to go out of scope. The asynchronous protocol uses a dedicated queue of output requests, class `ConvergenceReportQueue`, into which the producer&mdash;i.e., member function `runStep()`&mdash;inserts new convergence reports and from which the output thread, `ConvergenceOutputThread::writeASynchronous()`, retrieves those requests before writing the file data.

PR #4332, opened 2022-12-16
Rename Convergence Output Option

bska:
This PR is in response to an [earlier comment](https://github.com/OPM/opm-simulators/pull/4323#issuecomment-1354359405) which suggested we rename the previously introduced command line option
`ExtraConvergenceOutput` (`--extra-convergence-output`) into the more descriptive
```
OutputExtraConvergenceInfo (--output-extra-convergence-info)
```


bska:
Note that this PR depends on and includes #4323 and must, at minimum, not be merged before that PR is installed on the master branch.  I also don't really have an opinion on this particular name, so I'm going to leave it in a draft state until a decision has been made and we either keep the original option name or use this or some other name instead.

bska:
> I approve the renaming, should be merged.

Thanks.  I'll mark this as "ready for review" and run a build check.

bska:
jenkins build this please

PR #4335, opened 2022-12-20
clamp rs/rv in rate converter between 0.0 and rs/rv max

totto82:
Noticed while working on #4292

### DIFF
--- a/opm/simulators/wells/RateConverter.hpp
+++ b/opm/simulators/wells/RateConverter.hpp
@@ -562,27 +562,6 @@ namespace Opm {
                 if ((io < 0) || (ig < 0)) {
                     return { rs, rv };
                 }
-
-#define BURN_RESV_BRIDGES 0
-
-#if !BURN_RESV_BRIDGES
-
-                auto b = rs;
-                if (io >= 0 && ig >= 0) {
-                    b = surface_rates[ig] / (surface_rates[io] + 1.0e-15);
-                }
-                const double Rs = std::min(rs, b);
-
-                b = rv;
-                if (io >= 0 && ig >= 0) {
-                    b = surface_rates[io] / (surface_rates[ig] + 1.0e-15);
-                }
-                const double Rv = std::min(rv, b);
-
-                return { Rs, Rv };
-
-#else // BURN_RESV_BRIDGES
-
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

totto82:
> To be honest if we're going to do this anyway, I'd really prefer we go the [BURN_RESV_BRIDGES](https://github.com/OPM/opm-simulators/blob/f99b125156ef7fdeb0ea680943d6ba1ed0f4dee2/opm/simulators/wells/RateConverter.hpp#L586-L595) route instead.

Do you know the motivation of "BURN_RESV_BRIDGES" I haven't notices this before today. And yes the code within BURN_RESV_BRIDGES should do the trick, but why is this not the default behavior.

bska:
> > To be honest if we're going to do this anyway, I'd really prefer we go the [BURN_RESV_BRIDGES](https://github.com/OPM/opm-simulators/blob/f99b125156ef7fdeb0ea680943d6ba1ed0f4dee2/opm/simulators/wells/RateConverter.hpp#L586-L595) route instead.
> 
> Do you know the motivation of "BURN_RESV_BRIDGES" I haven't notices this before today

Yes.  I added that block in #4164 (commit f0e2a1efe0) because I considered it to be a better way of doing what the current master does.  I also made that code conditional on `BURN_RESV_BRIDGES` (commit d388ce40e) because it changes the simulation results and I didn't want to deal with that while I was working on something else (#4180).

> why is this not the default behavior.

Because I didn't have time to verify that every change generated by switching the algorithm was well within the acceptable limits.

totto82:
Thanks, I will update the PR so we start using the code inside BURN_RESV_BRIDGES instead

bska (approved):
This is fine provided the new simulation results are okay.

totto82:
I have manually checked the failures with most deviations. 
	 82 - compareECLFiles_flow+UDQ_M1 (Failed)
	125 - compareECLFiles_flow+MOD4_UDQ_ACTIONX (Failed)
	163 - compareECLFiles_flow+WVFPEXP-02 (Failed) 
The rest pass when only comparing restart steps and with looser tolerances.
I will start the updated process.

totto82:

totto82:
jenkins build this opm-tests=874 please

PR #4351, opened 2023-01-09
add missing include

akva2:
Downstream of https://github.com/OPM/opm-common/pull/3322


bska (approved):

PR #4363, opened 2023-01-11
add missing Balance.hpp includes

akva2:
And move more code to compile unit to avoid including Balance.hpp in BlackoilWellModel_impl.hpp

Downstream of https://github.com/OPM/opm-common/pull/3337


akva2:
jenkins build this please

PR #4368, opened 2023-01-12
Avoid WellConnections.hpp in simulator objects

akva2:
Downstream of https://github.com/OPM/opm-common/pull/3343

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -37,6 +37,7 @@
 #include <opm/input/eclipse/Schedule/Network/Balance.hpp>
 #include <opm/input/eclipse/Schedule/Network/ExtNetwork.hpp>
 #include <opm/input/eclipse/Schedule/Schedule.hpp>
+#include <opm/input/eclipse/Schedule/Well/WellConnections.hpp>
 #include <opm/input/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp>
 
 #include <opm/simulators/utils/DeferredLogger.hpp>
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1239,4 +1239,24 @@ shouldIterateNetwork(const int reportStepIdx,
            network_imbalance > balance.pressure_tolerance();
 }
 
+std::vector<int>
+BlackoilWellModelGeneric::
+getCellsForConnections(const Well& well) const
+{
+    std::vector<int> wellCells;
+    // All possible connections of the well
+    const auto& connectionSet = well.getConnections();
+    wellCells.reserve(connectionSet.size());
+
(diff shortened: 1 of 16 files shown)


opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1238,4 +1239,25 @@ shouldIterateNetwork(const int reportStepIdx,
...
+    wellCells.reserve(connectionSet.size());
+
+    for (std::size_t c = 0; c < connectionSet.size(); c++)
+    {
+        const auto& connection = connectionSet.get(c);
+        int compressed_idx = compressedIndexForInterior(connection.global_index());
+        if ( compressed_idx >= 0 ) { // Ignore connections in inactive/remote cells.
+            wellCells.push_back(compressed_idx);
+        }
+    }
```
bska, line 1258: You can use range-for here since the `WellConnections` class has the requisite `begin()` and `end()` member functions.

opm/simulators/wells/WellConnectionAuxiliaryModule.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    // initialize the additional cell connections introduced by wells.
+    for (const auto& well : schedule_wells)
+    {
+        std::vector<int> compressed_well_perforations;
+        // All possible completions of the well
+        const auto& completionSet = well.getConnections();
+        compressed_well_perforations.reserve(completionSet.size());
+
+        for (size_t c = 0; c < completionSet.size(); c++)
```
bska, line 63: You can use range-for here.

opm/simulators/wells/WellConnectionAuxiliaryModule.cpp
```diff
@@ -0,0 +1,88 @@
...
+        {
+            const auto& completion = completionSet.get(c);
+            int i = completion.getI();
+            int j = completion.getJ();
+            int k = completion.getK();
+            int cart_grid_idx = i + cartesianSize[0]*(j + cartesianSize[1]*k);
+            int compressed_idx = cartesianToCompressed[cart_grid_idx];
```
bska, line 70: You don't need the I,J,K indices here.  You can just use
```
int compressed_idx = cartesianToCompressed[completion.global_index()];
```
instead.

bska (commented):
Looks good for the most part, but since we're here anyway I think it'd be a good idea to simplify some of the logic too.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
akva2: done.

opm/simulators/wells/WellConnectionAuxiliaryModule.cpp
akva2: done.

opm/simulators/wells/WellConnectionAuxiliaryModule.cpp
akva2: done.

bska (approved):

PR #4370, opened 2023-01-12
compute CVIR and CVPR

totto82:
It seems like the connection reservoir rate was never computed.

### DIFF
--- a/opm/simulators/wells/WellInterfaceFluidSystem.cpp
+++ b/opm/simulators/wells/WellInterfaceFluidSystem.cpp
@@ -64,7 +64,7 @@ WellInterfaceFluidSystem(const Well& well,
 {
 }
 
-template<typename FluidSystem>
+template <typename FluidSystem>
 void
 WellInterfaceFluidSystem<FluidSystem>::
 calculateReservoirRates(SingleWellState& ws) const
@@ -72,14 +72,32 @@ calculateReservoirRates(SingleWellState& ws) const
     const int fipreg = 0; // not considering the region for now
     const int np = number_of_phases_;
 
-    std::vector<double> surface_rates(np, 0.0);
-    for (int p = 0; p < np; ++p) {
-        surface_rates[p] = ws.surface_rates[p];
-    }
+    this->rateConverter_
+        .calcReservoirVoidageRates(fipreg,
+                                   this->pvtRegionIdx_,
+                                   ws.surface_rates,
+                                   ws.reservoir_rates);
 
-    std::vector<double> voidage_rates(np, 0.0);
-    rateConverter_.calcReservoirVoidageRates(fipreg, pvtRegionIdx_, surface_rates, voidage_rates);
-    ws.reservoir_rates = voidage_rates;
+    // Compute total connection reservoir rate CVPR/CVIR
+    auto& perf_data = ws.perf_data;
+    const auto num_perf_well = perf_data.size();
+    const auto& surf_perf_rates = perf_data.phase_rates;
+    for (auto i = 0*num_perf_well; i < num_perf_well; ++i) {
+        const auto surface_rates_perf = std::vector<double>
+            { surf_perf_rates.begin() + (i + 0)*np ,
+              surf_perf_rates.begin() + (i + 1)*np };
+
+        std::vector<double> voidage_rates_perf(np, 0.0);
+        this->rateConverter_
+            .calcReservoirVoidageRates(fipreg,
+                                       this->pvtRegionIdx_,
+                                       surface_rates_perf,
+                                       voidage_rates_perf);
+
+        perf_data.rates[i] =
+            std::accumulate(voidage_rates_perf.begin(),
+                            voidage_rates_perf.end(), 0.0);
+    }
 }
 
 template <typename FluidSystem>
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

opm/simulators/wells/WellInterfaceFluidSystem.cpp
```diff
@@ -80,6 +80,23 @@ calculateReservoirRates(SingleWellState& ws) const
     std::vector<double> voidage_rates(np, 0.0);
     rateConverter_.calcReservoirVoidageRates(fipreg, pvtRegionIdx_, surface_rates, voidage_rates);
     ws.reservoir_rates = voidage_rates;
+
+    // Compute total connection reservoir rate CVPR/CVPI
```
bska, line 84: Tiny nit: The comment should refer to `CVIR` instead of `CVIP`.

totto82:
Why does this fail for the parallel and restart tests?

bska:
> Why does this fail for the parallel and restart tests?

I don't know, I need to run it locally to get a better overview.  Given the nature of the change I'm **guessing** that the failures are in the cumulatives for injected/produced reservoir voidage volumes at the connection level (summary vectors `CVPT` and `CVIT`), but I'd need to check to be sure.

bska:
> > Why does this fail for the parallel and restart tests?
> 
> I don't know, I need to run it locally to get a better overview.

Would you mind rebasing this PR onto the current master please?  I'm having trouble applying the change locally due to divergent base branches.

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
@bska Should we try to merge this?

bska:
> Should we try to merge this?

Yes I think so, but I'm not entirely comfortable merging something that we're not able to restart from or which produces different solutions in parallel vs. sequential.  I haven't gotten to the bottom of what's happening yet.

totto82:
jenkins build this please

totto82:
> Yes I think so, but I'm not entirely comfortable merging something that we're not able to restart from or which produces different solutions in parallel vs. sequential. I haven't gotten to the bottom of what's happening yet

Based on the known issue about the instability of the well model, how close can we expect the results to be?

opm/simulators/wells/WellInterfaceFluidSystem.cpp
```diff
@@ -80,6 +80,24 @@ calculateReservoirRates(SingleWellState& ws) const
...
+        perf_rates[i] = 0.0;
+        for (int p = 0; p < np; ++p) {
+            surface_rates_perf[p] = surf_perf_rates[np*i+p];
+        }
+        std::vector<double> voidage_rates_perf(np, 0.0);
+        rateConverter_.calcReservoirVoidageRates(fipreg, pvtRegionIdx_, surface_rates_perf, voidage_rates_perf);
+        for (int p = 0; p < np; ++p) {
+            perf_rates[i] += voidage_rates_perf[p];
+        }
+    }
```
bska, line 100: I've looked at this and I'm still not able to quite get to the bottom of what's happening with the test failures.  The differences between sequential and parallel runs are smaller than before, but still large enough to be a little concerning.  While looking into this, I rewrote the function to the one below which has the benefit of not unnecessarily copying rate data, and of not needing to update `perf_rates[i]` in a loop.  I would suggest you use this version too.

I will continue to look into this.

---

```C++
template <typename FluidSystem>
void
WellInterfaceFluidSystem<FluidSystem>::
calculateReservoirRates(SingleWellState& ws) const
{
    const int fipreg = 0; // not considering the region for now
    const int np = number_of_phases_;

    this->rateConverter_
        .calcReservoirVoidageRates(fipreg,
                                   this->pvtRegionIdx_,
                                   ws.surface_rates,
                                   ws.reservoir_rates);

    // Compute total connection reservoir rate CVPR/CVIR
    auto& perf_data = ws.perf_data;
    const auto num_perf_well = perf_data.size();
    const auto& surf_perf_rates = perf_data.phase_rates;
    for (auto i = 0*num_perf_well; i < num_perf_well; ++i) {
        const auto surface_rates_perf = std::vector<double>
            { surf_perf_rates.begin() + (i + 0)*np ,
              surf_perf_rates.begin() + (i + 1)*np };

        std::vector<double> voidage_rates_perf(np, 0.0);
        this->rateConverter_
            .calcReservoirVoidageRates(fipreg,
                                       this->pvtRegionIdx_,
                                       surface_rates_perf,
                                       voidage_rates_perf);

        perf_data.rates[i] =
            std::accumulate(voidage_rates_perf.begin(),
                            voidage_rates_perf.end(), 0.0);
    }
}
```

totto82:
jenkins build this please

opm/simulators/wells/WellInterfaceFluidSystem.cpp
totto82: Thanks, I will update the code.

totto82:
Stricter well tolerance helped for the parallel failures. So lets try jenkins again.

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
@bska The parallel and restart cases that failed pass with stricter well tolerances. Should we do an update date and get this in

bska (approved):
Thanks for the updates.  This looks good to me now and with the realisation that

> the parallel and restart cases that failed pass with stricter well tolerances

I think this is ready for merging.  I'll update the reference solutions and get the new version into the master branch.

bska:

bska:
jenkins build this opm-tests=908 please

PR #4372, opened 2023-01-13
output gas pressure for gas-water case

totto82:
Issue reported in #4316

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -32,8 +32,8 @@
 
 #include <opm/models/blackoil/blackoilproperties.hh>
 
-#include <opm/models/utils/propertysystem.hh>
 #include <opm/models/utils/parametersystem.hh>
+#include <opm/models/utils/propertysystem.hh>
 
 #include <opm/material/common/Valgrind.hpp>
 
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

ebos/ecloutputblackoilmodule.hh
```diff
@@ -234,11 +234,11 @@ public:
...
-                    // put pressure in oil pressure for output
-                    if (FluidSystem::phaseIsActive(waterPhaseIdx)) {
-                        this->oilPressure_[globalDofIdx] = getValue(fs.pressure(waterPhaseIdx));
-                    } else {
+                    // put pressure in oil pressure for output. Use gas if oil is not present
+                    if (FluidSystem::phaseIsActive(gasPhaseIdx)) {
                         this->oilPressure_[globalDofIdx] = getValue(fs.pressure(gasPhaseIdx));
+                    } else { // use water if neither oil nor gas is present
+                        this->oilPressure_[globalDofIdx] = getValue(fs.pressure(waterPhaseIdx));
                     }
```
bska, line 242: If we're changing this anyway, I think I'd prefer that we straighten out the conditionals too, e.g.,
```C++
if (FluidSystem::phaseIsActive(oilPhaseIdx)) {
    // ...
}
else if (FluidSystem::phaseIsActive(gasPhaseIdx)) {
    // ...
}
else {
    // ...
}
```
We might even consider writing this as a loop
```C++
for (const auto idx : { oilPhaseIdx, gasPhaseIdx, waterPhaseIdx }) {
    if (FluidSystem::phaseIsActive(idx)) {
        this->oilPressure_[globalDofIdx] = getValue(fs.pressure(idx));
        break;
    }
}
```
but that might on the other hand be too obscure.

bska (commented):
Looks good to me.  I just think we should take the opportunity to reduce the level of nesting here.   I'll merge once that's done.

ebos/ecloutputblackoilmodule.hh
totto82: I think we should also rename oilPressure_. What about referencePressure_?

bska: > I think we should also rename `oilPressure_`

I think so too.  Explicitly referring to this as oil pressure is likely to be more confusing as time passes.

> What about `referencePressure_`?

Yeah, either that or maybe `fluidPressure_`.

totto82: Good suggestion.

totto82:
jenkins build this please

bska (approved):

PR #4384, opened 2023-01-16
Adds support for directional relative permeabilities with hysteresis

hakonhagland:
See upstream https://github.com/OPM/opm-common/pull/3364 for more information

### DIFF
--- a/ebos/eclgenericproblem.cc
+++ b/ebos/eclgenericproblem.cc
@@ -368,6 +368,15 @@ updateKrnum_()
 }
 
 template<class GridView, class FluidSystem, class Scalar>
+void EclGenericProblem<GridView,FluidSystem,Scalar>::
+updateImbnum_()
+{
+    updateNum("IMBNUMX", imbnumx_);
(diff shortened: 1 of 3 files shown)


OPMUSER:
I'm working on tests for this, but it will take a while as we currently have six tests based on the ones @hakonhagland generated. Also, there are 16 relative permeability tables, but only four are used in the tests, but new sets are required for the hysteresis tests.

OPMUSER:
I get this:

```
Error: Unsupported keywords or keyword items:

  ENDSCALE: invalid value 'DIRECT' for item 1
  In file: KRNUM-02-IMBNUM.DATA, line 62
  ENDSCALE(DIRECT): only default value of NODIR supported

In <memory string> line 0
Unsupported keywords or keyword items:

  ENDSCALE: invalid value 'DIRECT' for item 1
  In file: KRNUM-02-IMBNUM.DATA, line 62
  ENDSCALE(DIRECT): only default value of NODIR supported
```

@hakonhagland would you like me to fix this by removing the check?

PR #4393, opened 2023-01-17
Noelelmctx update

hnil:
Update without element context need same branch name in opm-model.


bska:
> Update without element context need same branch name in opm-model.

That's OPM/opm-models#777, right?

hnil:
change the method and moved to opm-flowexperiments

PR #4395, opened 2023-01-18
Hack to fix for starup/restart

totto82:


totto82:
Depends on https://github.com/OPM/opm-models/pull/772

atgeirr:
This hack should no longer be needed.

PR #4396, opened 2023-01-18
Main: introduce a compile unit

akva2:
Reduce hotness of Python.hpp by 26% (27 files, 78 instead of 105)
Reduce hotness of UDQState.hpp by 60% (30 files, 20 instead of 50)
Reduce hotness of State.hpp by 48% (29 files, 31 instead of 60)

### DIFF
--- a/opm/simulators/flow/Main.cpp
+++ b/opm/simulators/flow/Main.cpp
@@ -24,6 +24,7 @@
 #include <opm/simulators/flow/Main.hpp>
 
 #include <opm/input/eclipse/Python/Python.hpp>
+#include <opm/input/eclipse/Schedule/UDQ/UDQState.hpp>
 
 #include <opm/simulators/utils/readDeck.hpp>
 
--- a/opm/simulators/flow/Main.cpp
+++ b/opm/simulators/flow/Main.cpp
@@ -23,6 +23,8 @@
 #include <config.h>
 #include <opm/simulators/flow/Main.hpp>
 
+#include <opm/simulators/utils/readDeck.hpp>
+
 namespace Opm {
 
 Main::Main(int argc, char** argv)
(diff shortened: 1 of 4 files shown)


akva2:
jenkins build this please

opm/simulators/flow/Main.cpp
```diff
@@ -0,0 +1,236 @@
...
+void Main::readDeck(const std::string& deckFilename,
+                    const std::string& outputDir,
+                    const std::string& outputMode,
+                    const bool init_from_restart_file,
+                    const bool allRanksDbgPrtLog,
+                    const bool strictParsing,
+                    const int mpiRank,
+                    const int output_param)
+{
+    auto python = std::make_shared<Python>();
```
bska, line 176: If you're changing this anyway, you might as well move the `make_shared<Python>()` call to the argument list of `Opm::readDeck()`.  That way we don't incur the extra shared-pointer copy when calling `readDeck()`.  Alternatively, you might use `std::move(python)` in the argument list.

bska (approved):
Very good.  Anything that reduces the effort of reading `Main.hpp` is most welcome.  Just one minor thing concerning the `Python` object, and then I'm ready to merge this.

opm/simulators/flow/Main.cpp
akva2: moved the make_shared call.

akva2:
jenkins build this please

akva2:
merging per approval.

opm/simulators/flow/Main.cpp
```diff
@@ -0,0 +1,244 @@
...
+    int is_client;
+    MPI_Comm new_comm;
+    int err = damaris_start(&is_client);
+    isSimulationRank_ = (is_client > 0);
+    if (isSimulationRank_ && err == DAMARIS_OK) {
+        damaris_client_comm_get(&new_comm);
+        EclGenericVanguard::setCommunication(std::make_unique<Parallel::Communication>(new_comm));
+    } else {
+        return false;
+    }
```
bska, line 240: Sorry, I didn't notice this during review.  Member function `Main::setupDamaris()` is declared as not returning anything, but the `else` branch here returns `false`.  That seems strange for two reasons,

1. The non-simulation rank, i.e., `is_client <= 0`, will return `false`
2. We'll return `false` also if `damaris_start()` returns something other than `DAMARIS_OK`.

What's the appropriate semantics here?  Granted most of the questionable logic was already there, but we need to decide what's the appropriate action for the `else` branch.

bska (commented):
I just noticed one potential issues post-commit.

opm/simulators/flow/Main.cpp
akva2: ai. i should have picked this up but i had forgotten to enable damaris in my build configuration :/

PR #4399, opened 2023-01-18
fix Damaris setup in Main

akva2:
Regression from https://github.com/OPM/opm-simulators/pull/4396

I think this should be fine, @bska


akva2:
jenkins build this please

bska (approved):
> I think this should be fine

Agreed, this looks good to me and a lot better than the original logic.  I'll merge this into master.

PR #4400, opened 2023-01-18
Use CPR if specified in RUNSPEC / Make cprw the default

blattms:
If there is a command line parameter for the linear solver via `--linear-solver=...` this has precedence.
For CPR in the RUNSPEC section and --linear-solver="cpr" the cprw variant will be use (as it does not need the well perforations to be added to the matrix)


blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
As expected the (aquifer) tests that have already been using cpr fail. That is because of the move to cprw as the default.

Either we split this PR in two, or we request "cpr_trueimpes"  (instead of  "cpr" which now selects cprw) in the jenkins scripts in opm-simulators.

akva2:
i'd say request cpr_trueimpes. changing the tests to use cprw is a separate concern.

blattms:
Done.

blattms:
jenkins build this please

OPMUSER:
"Houston we have a problem."  :-(

In the CPR JSON file definition we have this:
```
{
  "name": "CPR",
  "items": [
    {
      "name": "WELL",
      "value_type": "STRING"
    },
    {
      "name": "I",
      "value_type": "INT"
    },
    {
      "name": "J",
      "value_type": "INT"
    },
    {
      "name": "K",
      "value_type": "INT"
    }
  ],
  "sections": [
    "RUNSPEC",
    "SUMMARY"
  ],
  "comment": "in the SimulationConfig() constructor"
}
```

Which would appear to be the definition of the SUMMARY vector CPR well connection definition.

Should this be changed to:

```
{
  "name": "CPR",
  "items": [
    {
      "name": "CPROPTN",
      "value_type": "STRING",
      "default": "ORIGINAL"
    }
  ],
  "sections": [
    "RUNSPEC"
  ],
  "comment": "Equivalent to linear-solver=CPRW"
}
```

If so, I'll correct this in a separate PR.

OPMUSER:
Okay, this does not work:

OPMUSER:
```
 {
   "name": "CPR",
   size=1,
   "items": [
     {
       "name": "CPROPTN",
       "value_type": "STRING",
       "default": "ORIGINAL"
     }
   ],
   "sections": [
     "RUNSPEC"
   `],`
   "comment": "Equivalent to linear-solver=CPRW"
 } 
```
Okay, the above does not work, because there is a hard coded check requiring an empty record. If I change to:
```
{
  "name": "CPR",
  "sections": [
    "RUNSPEC"
  ],
  "comment": "Equivalent to linear-solver=CPRW"
}

```
Then this will run with 
```
CPR
 /
```
However, the commercial simulator allows for:
```
CPR
ORIGINAL
 /

CPR
ADAPTIVE
/
```
But then you get:
```
Error: Unknown keyword: ADAPTIVE

In <memory string> line 0
Unknown keyword: ADAPTIVE
```
The reason for this very long story is I'm trying to update the documentation for the CPR, with respect to OPM Flow and the commercial simulator. What we should allow to work is:

```
CPR
 /
```
And also:
```
CPR
ORIGINAL
 /
```
Then allow this to run as well, as this is a valid option in the commercial simulator and is equivalent to OPM Flow's implementation.

Finally, 
```
CPR
ADAPTIVE
 /
```
Should fail, and stop with an error, as this will influence the results.

The validation checking should be done in https://github.com/OPM/opm-simulators/blob/master/opm/simulators/utils/PartiallySupportedFlowKeywords.cpp. 

I have pushed a commit to remove CPR from the unsupported keyword list https://github.com/OPM/opm-simulators/pull/4404.

bska:
Okay, whatever we do to the `CPR` keyword description, we have to make sure that we still support that keyword with the "connection pressure" interpretation in the SUMMARY section. Adding the string CPR to the CONNECTION_PROBE meta-keyword **may** be sufficient, but we really have to test that.

blattms:
Well, I was aware that there is only one CPR json description but two ways of usage (CPR and connection pressure). That is not nice, but changing that would probably need a huge design change. I did not see a big problem in leaving it as it is.

I think all three versions  (default adaptive, and original) should work, (but probably also many other faulty ones). I think I did test that.

If we really want to do checks and issue warnings we should do that inside the Parser code manually. everything else is a PIA and lots of rewrite work for very little gain.

Also I don't see why adaptive should give an error. (non-adaptive is a subset of adaptive to me :stuck_out_tongue_winking_eye: )

OPMUSER:
Well I'm all for making life easy for one and all, but if @bska's suggestion:

> Adding the string CPR to the CONNECTION_PROBE meta-keyword may be sufficient, but we really have to test that.

Works without too much pain, then I think we should implement that, as it is more consistent with the existing code base.

Secondly, the reason for distinguishing between the `ORIGINAL` and `ADAPTIVE` options is the commercial simulator will give different results, and that may be an issue for users when comparing against OPM Flow. However, if @tskille is happy with not  distinguishing between the two options, then so am I.

Just my two cents.

atgeirr:
> Ideally the results (as in result curves) should not differ very much. I would assume that ADAPTIVE is mainly there for saving computation time for certain models.

I am not familiar with these variants, do you know what is the difference? To me ADAPTIVE sounds like it may be using CPR or the default linear solver depending on some heuristic, but that may be totally wrong...

OPMUSER:
> I am not familiar with these variants, do you know what is the difference? To me ADAPTIVE sounds like it may be using CPR or the default linear solver depending on some heuristic, but that may be totally wrong...

@atgeirr Yes you are correct. The Adaptive option means that the CPR preconditioner is only applied "during difficult part of the simulation". Basically it is an attempt to get the best of both types of solvers.

Based on this I think we should change the CPR check to only allow the `ORIGINAL` option. And I don't think we should allow  the CPR JSON file to be used for both CPR and the CPR SUMMARY vectors - but I'm not a C++ guy.

atgeirr:
> The Adaptive option means that the CPR preconditioner is only applied "during difficult part of the simulation". Basically it is an attempt to get the best of both types of solvers.

We have something similar in a PR (it times the linear solvers every 100 solves and chooses the fastest one for the next 100 solves), which needs a little work and tuning but can probably be merged reasonably soon. I think then supporting ADAPTIVE could be done that way, and then also add --linear-solver=adaptive which chooses the same variant.

OPMUSER:
> We have something similar in a PR (it times the linear solvers every 100 solves and chooses the fastest one for the next 100 solves), which needs a little work and tuning but can probably be merged reasonably soon. I think then supporting ADAPTIVE could be done that way, and then also add --linear-solver=adaptive which chooses the same variant.

I think that is a really excellent idea.

PR #4404, opened 2023-01-20
Remove CPR from UnsupportedFlowKeywords.cpp

OPMUSER:
CPR keyword is now supported.


blattms:
Thanks. I always forget about this part.

PR #4405, opened 2023-01-24
including <optional> header to WellContainer

GitPaean:
to remove the following complaint from CLION compiler, not sure why it did not show up before. 

```c++
/home/kaib/OPM-devel4/opm/opm-simulators/opm/simulators/wells/WellContainer.hpp:144:10: error: ‘optional’ in namespace ‘std’ does not name a template type
  144 |     std::optional<int> well_index(const std::string& wname) const {
```

### DIFF
--- a/opm/simulators/wells/WellContainer.hpp
+++ b/opm/simulators/wells/WellContainer.hpp
@@ -21,6 +21,7 @@
 #define OPM_WELL_CONTAINER_HEADER_INCLUDED
 
 #include <initializer_list>
+#include <optional>
 #include <stdexcept>
 #include <string>
 #include <unordered_map>


GitPaean:
jenkins build this please

bska (commented):
Looks good for the most part, but would you mind keeping the list of headers sorted alphabetically, please?

> not sure why it did not show up before.

The issue was **probably** hidden by a transitive include before.  Did you recently upgrade your toolchain or create a clean workspace?

GitPaean:
jenkins build this please

GitPaean:
> Looks good for the most part, but would you mind keeping the list of headers sorted alphabetically, please?

done! did not notice the order. 

> Did you recently upgrade your toolchain or create a clean workspace?

No, I did not. Sometimes I feel the change of the code might affect the compiler's interpretation like this, but it is not something I am familiar with and comfortable to give suggestions.

bska (approved):

blattms:
probably fallout of some refactoring. Different compilers/ stdlibs might include different headers transitively. If you are doing a cleanup of included headers then you might not test all of them...

PR #4406, opened 2023-01-25
Add rocsparseSolver

Tongdongq:
`rocsparseSolver` uses the low-level blas functions from rocblas and rocsparse to form an ilu0-bicgstab solver. The wellcontributions must be included in the matrix.
`rocalutionSolver` uses rocsparse underneath, but this solver allows to pass a second, more sparse matrix to the ilu0. This matrix is generated by passing `--num-jacobi-blocks=INT` via commandline.

There are quite a few configuration issues so far:
On the simula ex3 cluster, with rocm 5.2.3:
g++ 10.1 <s>fails because `-lCLANGRT_BUILTINS-NOTFOUND` makes its way into the linker command, same when rocalution is disabled</s>, fixed by passing `-DROCM_PATH` to cmake.
clang 14 and hipcc (clang 14) compile rocsparse (rocalution disabled), rocalution fails compilation
```
In file included from /cm/shared/apps/amd/rocm/5.2.3/include/rocalution/rocalution.hpp:68:
/cm/shared/apps/amd/rocm/5.2.3/include/rocalution/solvers/multigrid/ruge_stueben_amg.hpp:65:9: error: an attribute list cannot appear here
    [[deprecated("This function will be removed in a future release. Use "
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning and 1 error generated when compiling for gfx803.
```

On another machine with rocm 5.4.1:
g++ 9.4 compiles and runs both rocalution and rocsparse
clang 16 compiles and runs both rocalution and rocsparse
hipcc (clang 15) also compiles and runs both rocalution and rocsparse

It looks like that using `find_package(rocsparse)` in CMakeLists.txt does not matter.
Adding them to opm-simulators-prereqs.cmake is more important, otherwise this error occurs:
```
	/usr/bin/ld: lib/libopmsimulators.a(rocsparseSolverBackend.cpp.o): undefined reference to symbol 'hipStreamDestroy@@hip_4.2'
	/usr/bin/ld: /mnt/beegfs/software/rocm/rocm-5.4.1/lib/libamdhip64.so.5: error adding symbols: DSO missing from command line
```
The `rocsparse_dbsrmv()` are depricated in newer versions, and can be replaced by `rocsparse_dbsrmv_ex()`, in combination with `rocsparse_mat_info spmv_info` and analysis.

The 2 rocm solvers needed `#undef HAVE_CUDA` to prevent CUDA vector structs being defined, which would collide with the AMD vector structs.


blattms (commented):
Reviewed just the CMake part for now.
Please add an elaborate comment what works/is activated with which compilers/packages. Judging from the nested is rather hard

blattms:
This will need a rebase before merging.

CMakeLists.txt
```diff
@@ -205,6 +205,49 @@ else()
...
+      unset(HAVE_ROCSPARSE)
+    endif()
+    find_package(rocalution)
+    if(ROCALUTION_FOUND)
+      set(HAVE_ROCALUTION 1)
+      set(COMPILE_BDA_BRIDGE 1)
+    endif()
+  else()
+    find_package(rocblas)
+    find_package(rocsparse)
```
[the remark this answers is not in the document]
Tongdongq: `rocalutionSolver` and `rocsparseSolver` are independent of each other. (despite rocalution using rocsparse underneath, only the rocalution package needs to be found)
Only `rocsparseSolver` needs both rocblas and rocsparse.

blattms:
jenkins build this please

blattms:
Jenkins (with rocblas installed) fails compiling.

On my Debian testing without rocblas cmake succeeds, but compilation fails with:
```
/home/mblatt/src/opm/opm-simulators/opm/simulators/linalg/bda/rocsparseSolverBackend.hpp:30:10: fatal error: rocblas/rocblas.h: Datei oder Verzeichnis nicht gefunden
   30 | #include <rocblas/rocblas.h>
      |          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/opmsimulators.dir/build.make:1742: CMakeFiles/opmsimulators.dir/opm/simulators/linalg/bda/BdaBridge.cpp.o] Fehler 1
```

I guess we need to do something if rocblas is not there.

CMakeLists.txt
```diff
@@ -491,6 +528,11 @@ if(ROCALUTION_FOUND)
   target_include_directories(opmsimulators PUBLIC ${rocalution_INCLUDE_DIR}/rocalution)
 endif()
 
+if(rocsparse_FOUND AND rocblas_FOUND)
+  target_link_libraries( opmsimulators PUBLIC rocsparse )
+  target_link_libraries( opmsimulators PUBLIC rocblas )
```
blattms, line 533: Please use the targets roc::rocsparse and roc::rocblas instead here.

CMakeLists.txt
```diff
@@ -205,6 +205,49 @@ else()
...
+      set(HAVE_ROCALUTION 1)
+      set(COMPILE_BDA_BRIDGE 1)
+    endif()
+  else()
+    find_package(rocblas)
+    find_package(rocsparse)
+    if(rocsparse_FOUND AND rocblas_FOUND)
+      set(HAVE_ROCSPARSE 1)
+      set(COMPILE_BDA_BRIDGE 1)
+    endif()
```
blattms, line 238: Even without rocblas HAVE_ROCSPARSE is 1 on my system before the if clause. Hence we need an else branch that unsets this.

Generally, I am wondering whether rocsparse really needs rocblas. Are you sure about this? Or is it just that our code will use it.

On my system:
```
ldd /usr/lib/x86_64-linux-gnu/librocsparse.so
	linux-vdso.so.1 (0x00007ffcce948000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f45892d7000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f45891f8000)
	libamdhip64.so.5 => /lib/x86_64-linux-gnu/libamdhip64.so.5 (0x00007f453a200000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f4539e00000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f453a01f000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f4589315000)
	libhsa-runtime64.so.1 => /lib/x86_64-linux-gnu/libhsa-runtime64.so.1 (0x00007f4539a00000)
	libnuma.so.1 => /lib/x86_64-linux-gnu/libnuma.so.1 (0x00007f45891e9000)
	libhsakmt.so.1 => /lib/x86_64-linux-gnu/libhsakmt.so.1 (0x00007f453afd5000)
	libelf.so.1 => /lib/x86_64-linux-gnu/libelf.so.1 (0x00007f45891ce000)
	libdrm.so.2 => /lib/x86_64-linux-gnu/libdrm.so.2 (0x00007f453afbf000)
	libdrm_amdgpu.so.1 => /lib/x86_64-linux-gnu/libdrm_amdgpu.so.1 (0x00007f453afb3000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f4539de1000)
```

blattms (commented):
Reviewed CMake scripts.

CMakeLists.txt
```diff
@@ -205,6 +205,49 @@ else()
   endif()
 endif()
 
+message(INFO " HAVE_ROCSPARSE: ${HAVE_ROCSPARSE}")
```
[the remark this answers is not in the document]
blattms: It has by find_package.

CMakeLists.txt
blattms: This might solve the linker problem on Jenkins.

CMakeLists.txt
Tongdongq: It seems rocsparse does not depend on rocblas, but our solver does, due to the dense vector operations (axpy, etc.).

CMakeLists.txt
Tongdongq: 67ad9e6

opm/simulators/linalg/bda/BdaBridge.cpp
```diff
@@ -91,6 +95,13 @@ BdaBridge<BridgeMatrix, BridgeVector, block_size>::BdaBridge(std::string acceler
         backend.reset(new Opm::Accelerator::rocalutionSolverBackend<block_size>(linear_solver_verbosity, maxit, tolerance));
 #else
         OPM_THROW(std::logic_error, "Error rocalutionSolver was chosen, but rocalution was not found by CMake");
+#endif
+    } else if (accelerator_mode.compare("rocsparse") == 0) {
+#if HAVE_ROCSPARSE
+        use_gpu = true; // should be replaced by a 'use_bridge' boolean
+        backend.reset(new Opm::Accelerator::rocsparseSolverBackend<block_size>(linear_solver_verbosity, maxit, tolerance, platformID, deviceID));
+#else
+        OPM_THROW(std::logic_error, "Error openclSolver was chosen, but rocsparse was not found by CMake");
```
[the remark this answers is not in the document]
Tongdongq: 67ad9e6

blattms:
jenkins build this

bska:
jenkins build this please

Tongdongq:
The test failed, because it assumed the constructor fails when there is no device found.
Now indeed the constructor tries to find a device, and throw when not found.

I'll try to detect the rocm version and use the correct bsrmv().

blattms:
jenkins build this please

CMakeLists.txt
```diff
@@ -209,6 +209,41 @@ else()
...
 endif()
 
+find_package(hip)
+
+if(hip_FOUND)
+  get_filename_component(CXX_COMPILER ${CMAKE_CXX_COMPILER} NAME)
+  if(hip_VERSION VERSION_LESS "5.3")
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+      message(WARNING " Cannot use gcc for rocm < 5.3")
+      unset(HAVE_ROCSPARSE)
```
Tongdongq, line 219: This does not work, `HAVE_ROCSPARSE` is still 1. Setting it to 0 works, but it is still defined as `HAVE_ROCSPARSE 0` in `config.h`.

When testing with rocblas enabled, rocsparse is found, and a linker error occurs. This is because `rocsparseSolverBackend.cpp` does not get compiled, because `rocblas_FOUND` is false. But since `HAVE_ROCSPARSE` is 1, the `BdaBridge` includes `rocsparseSolverBackend.hpp`.

Tongdongq:
I disabled hipcc for both versions.
5.2.3 gives (also for master):
```
/opm/models/discretization/ecfv/ecfvstencil.hh:245:72: error: invalid operands to binary expression ('basic_string<char, std::char_traits<char>, std::allocator<char>>' and 'const int')
                    OPM_THROW(std::runtime_error, "Unexpected face id" << dirId_);
```
5.4 gives:
```
[100%] Linking CXX executable bin/flow_blackoil
ld.lld: error: duplicate symbol: guard variable for Opm::BlackOilFluidSystem<double, Opm::BlackOilDefaultIndexTraits>::gasPvt_
>>> defined at flow_ebos_blackoil.cpp
>>>            CMakeFiles/flow_libblackoil.dir/flow/flow_ebos_blackoil.cpp.o:(guard variable for Opm::BlackOilFluidSystem<double, Opm::BlackOilDefaultIndexTraits>::gasPvt_)
>>> defined at BlackOilFluidSystem.cpp
>>>            BlackOilFluidSystem.cpp.o:(.bss._ZGVN3Opm19BlackOilFluidSystemIdNS_26BlackOilDefaultIndexTraitsEE7gasPvt_E+0x0) in archive /opm-common/build/lib/libopmcommon.a
```

Tongdongq:
New test with 2023-3-16, 5c1d9644
5.4 compiles with g++, hipcc and clang++, when opm-simulators uses the same compiler as the other modules.
5.3 still fails for g++ because `-lCLANGRT_BUILTINS-NOTFOUND` is added to the linker command.
I can't compile opm-common (6b71f3d3) with clang++ 14:
```
[ 15%] Linking CXX executable bin/genkw
CMakeFiles/genkw.dir/src/opm/input/eclipse/Generator/KeywordGenerator.cpp.o: In function `Opm::KeywordGenerator::ensurePath(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
KeywordGenerator.cpp:(.text+0x68): undefined reference to `std::filesystem::__cxx11::path::parent_path() const'
KeywordGenerator.cpp:(.text+0x70): undefined reference to `std::filesystem::status(std::filesystem::__cxx11::path const&)'
```
Manually adding `-lstdc++fs` to the linker command works.

blattms:
Not nice but concerning stdc++fs you can ask the build system to link to it: `cmake -DOPM_CLANG_WITH_STDC++FS=ON`
We do that automatically for clang <=9, not sure why it is needed for 14 again.

Not sure where CLANGRT_BUILTINS comes from. Itseems this might be a library/package missing on the system (Hence the -NOT-FOUND extension). If you have the power to install it you might just install it. See "compiler-rt" runtime libraries on https://compiler-rt.llvm.org/
The usage of the variable probably comes from hip? Maybe check where the variable is in CMakeCache.txt?

Tongdongq:
After lots of debugging, I found that passing `-DROCM_PATH` fixes some intermediate variables.
More specifically in `rocm/5.2.3/lib/cmake/hip/hip-config.cmake`:
141: `set(ROCM_PATH /opt/rocm)`
154: `set(HIP_CLANG_ROOT "${ROCM_PATH}/llvm")`
183: `file(GLOB HIP_CLANG_INCLUDE_SEARCH_PATHS ${HIP_CLANG_ROOT}/lib/clang/*/include)`
The check on line 314 fails and `${CLANGRT_BUILTINS}` is added as a library, but contains the value `CLANGRT_BUILTINS-NOTFOUND`.

I think that's it concerning the errors.

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
```diff
@@ -0,0 +1,614 @@
...
+    ROCBLAS_CHECK(rocblas_set_stream(blas_handle, stream));
+}
+
+
+template <unsigned int block_size>
+rocsparseSolverBackend<block_size>::~rocsparseSolverBackend() {
+    HIP_CHECK(hipStreamSynchronize(stream));
+    HIP_CHECK(hipStreamDestroy(stream));
+    ROCSPARSE_CHECK(rocsparse_destroy_handle(handle));
+    ROCBLAS_CHECK(rocblas_destroy_handle(blas_handle));
```
bska, line 104: Since the `_CHECK` macros potentially call `OPM_THROW`, this is a destructor that will (potentially) throw an exception.  We should really try to avoid that.  Is there no other way of reporting failures here?

Tongdongq: We could write out the macro (or write a new one), replacing the `OPM_THROW` with something like `OpmLog::error()`.

The function `rocsparse_destroy_handle()` can perform some throwing and catching ([here](https://github.com/ROCmSoftwarePlatform/rocSPARSE/blob/82e7d8f09201ad55d08e62199b8ccc50251b1423/library/src/rocsparse_auxiliary.cpp#L79) and [here](https://github.com/ROCmSoftwarePlatform/rocSPARSE/blob/ba673954329423de932cc9b8d8365c0c277a4062/library/src/include/utility.h#L248)), but always returns a `rocsparse_status`.

Tongdongq: Changed to
```
    hipError_t hipstatus = hipStreamSynchronize(stream);
    if(hipstatus != hipSuccess){
        OpmLog::error("Could not synchronize with hipStream");
    }
```
could also write as
```
    if(hipStreamSynchronize(stream) != hipSuccess){
        OpmLog::error("Could not synchronize with hipStream");
    }
```

bska:
jenkins build this please

blattms:
I could not reproduce the clang problems. I don't version 14, but version 15 definitely works. Is the problem maybe in combination with hipcc?

Tongdongq:
That linker problem with `std::filesystem` was solved by `-DOPM_CLANG_WITH_STDC++FS=ON`.

I renabled hipcc, it works when other modules are compiled with `clang++`.
However, rocalution for hipcc/clang with rocm < 5.3 is disabled, because of this compile error:
```
In file included from /cm/shared/apps/amd/rocm/5.2.3/include/rocalution/rocalution.hpp:68:
/cm/shared/apps/amd/rocm/5.2.3/include/rocalution/solvers/multigrid/ruge_stueben_amg.hpp:65:9: error: an attribute list cannot appear here
    [[deprecated("This function will be removed in a future release. Use "
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning and 1 error generated when compiling for gfx803.
```

CMakeLists.txt
```diff
@@ -259,14 +259,12 @@ macro (config_hook)
...
     get_filename_component(CXX_COMPILER ${CMAKE_CXX_COMPILER} NAME)
-    if(${CXX_COMPILER} STREQUAL "hipcc")
-      message(FATAL_ERROR " hipcc does not work, use g++ or clang++")
-    endif()
     if(hip_VERSION VERSION_LESS "5.3")
-      if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-        message(WARNING " Cannot use g++ for rocsparse with rocm < 5.3\n Disabling rocsparseSolver")
-        unset(HAVE_ROCSPARSE)
-      endif()
+	    if(NOT CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
blattms, line 263: `.... AND ROCALUTION_FOUND)`?

Tongdongq: I added that to the check, and moved the rocalution adding outside of the if statement.

Tongdongq: Can you start jenkins again?

blattms:
jenkins build this please

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
```diff
@@ -0,0 +1,626 @@
...
+    }
+
+    initialized = true;
+} // end initialize()
+
+template <unsigned int block_size>
+void rocsparseSolverBackend<block_size>::copy_system_to_gpu(double *b) {
+    Timer t;
+
+#if COPY_ROW_BY_ROW
```
blattms, line 369: What is the default?

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
```diff
@@ -0,0 +1,626 @@
...
+    int sum = 0;
+    for (int i = 0; i < Nb; ++i) {
+        int size_row = rmat->rowPointers[i + 1] - rmat->rowPointers[i];
+        memcpy(vals_contiguous.data() + sum, reinterpret_cast<double*>(rmat->nnzValues) + sum, size_row * sizeof(double) * block_size * block_size);
```
blattms, line 373: Where do we make sure that vals_contiguous has the correct size? Can't find it. Is that done outside of the PR?
Maybe it should even be a local variable?

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
```diff
@@ -0,0 +1,626 @@
...
+        sum += size_row * block_size * block_size;
+    }
+    err = queue->enqueueWriteBuffer(d_Avals, CL_TRUE, 0, sizeof(double) * nnz, vals_contiguous.data(), nullptr, &events[0]);
+#else
+    // err = queue->enqueueWriteBuffer(d_Avals, CL_TRUE, 0, sizeof(double) * nnz, rmat->nnzValues, nullptr, &events[0]);
```
blattms, line 378: Is the empty else branch intentional?

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
```diff
@@ -0,0 +1,626 @@
...
```
blattms, line 416: empty else branch.

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
```diff
@@ -0,0 +1,626 @@
...
+    }
+} // end get_result()
+
+
+template <unsigned int block_size>
+SolverStatus rocsparseSolverBackend<block_size>::solve_system(std::shared_ptr<BlockedMatrix> matrix,
+                                                           double *b,
+                                                           std::shared_ptr<BlockedMatrix> jacMatrix,
+                                                           WellContributions& wellContribs,
+                                                           BdaResult &res)
```
blattms, line 587: Sorry for nitpicking: Indentation seems a bit off.

blattms (commented):
Hopefully final comments.

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
Tongdongq: Fixed in a squash

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
Tongdongq: Fixed in a squash

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
Tongdongq: Fixed in a squash

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
Tongdongq: Fixed in a squash

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
Tongdongq: Fixed in a squash.
The `BdaBridge` has `checkMemoryContiguous()`, this will error before copying row-by-row could be done.

blattms:
jenkins build this please

blattms:

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
```diff
@@ -0,0 +1,612 @@
...
+#undef HAVE_CUDA
+
+#include <opm/simulators/linalg/bda/rocsparseSolverBackend.hpp>
+
+#include <opm/simulators/linalg/bda/BdaResult.hpp>
+
+#include <hip/hip_runtime_api.h>
+#include <hip/hip_version.h>
+
+#ifdef HIP_HAVE_CUDA_DEFINED HAVE_CUDA
```
bska, line 46: `HAVE_CUDA` doesn't do anything here, and will just [emit a warning](https://ci.opm-project.org/job/opm-simulators-PR-builder/4769/gcc/new/folder.1616613304/fileName.396953383/).  What's the intended semantics of this directive?

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
```diff
@@ -0,0 +1,612 @@
...
+#define HAVE_CUDA HIP_HAVE_CUDA_DEFINED
+#undef HIP_HAVE_CUDA_DEFINED HAVE_CUDA
```
bska, line 48: `HAVE_CUDA` doesn't do anything here, and will just emit a warning.  What's the intended semantics of this directive?

bska (commented):
This version emits a couple of new warnings on the CI system, particularly for some unexpected preprocessor syntax.  It's not clear to me what the intended meaning is.  If you want to check if both symbols are defined you need to write
```C++
#if defined(HIP_HAVE_CUDA_DEFINED) && defined(HAVE_CUDA)
```
You also need two separate `#undef` directives to remove two separate macro definitions.

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
blattms: I think. you may test two defines with `#if defined(HIP_HAVE_CUDA_DEFINED) && defined(HAVE_CUDA)`

CMakeLists.txt
```diff
@@ -240,12 +240,6 @@ if(OpenCL_FOUND)
   endif()
 endif()
 
-find_package(rocalution)
-if(ROCALUTION_FOUND)
-  set(HAVE_ROCALUTION 1)
-  set(COMPILE_BDA_BRIDGE 1)
-endif()
-
```
blattms, line 248: Mmh, can't find the place where you search rocalution now. Seems like we might not do that anymore. Is that intentional?

Tongdongq: Should be lines 291-294.
We also removed the explicit `find_package(rocsparse)`, it should be done automatically because they are listed in `opm-simulators-prereqs.cmake`.

blattms: Ah, thanks.

OPMUSER:
@blattms Did this make it into 2023-04-RC1?

PR #4407, opened 2023-01-26
making sure to check the network nodal pressure constraint

GitPaean:
when checking the THP constraint for producers.

With the master branch, if we default the THP constraint in WCONPROD, the THP constraint will be skipped when checking well constraints because it does not exist in `WellProductionControls`.


GitPaean:
jenkins build this please

Issue #4341, opened 2023-01-02
OPM 2022.04 gives assertion error on simple setup


eivindjahren:
The following simple setup worked fine with opm 2021-10 but results in error
```
 flow: ./opm/simulators/timestepping/TimeStepControl.cpp:140: virtual double Opm::PIDTimeStepControl::computeTimeStepSize(double, int
, const Opm::RelativeChangeInterface&, double) const: Assertion `errors_[i]>0' failed.
```

when run with 2022.04

TEST.DATA
----------------
<details>

```
RUNSPEC
UNIFOUT

METRIC
DIMENS
 2 2 2 /
START
   1 'JAN' 2000 /
OIL
WATER
WELLDIMS
 0 /
EQLDIMS
 /
TABDIMS
 1* 1* 30 30 /
GRID
INIT
INCLUDE
 'TEST.GRDECL' /
PORO
 8*0.3 /
PERMX
 8*150 /
PERMY
 8*150 /
PERMZ
 8*150 /
PROPS
SWOF
 0.1 0.0    1.0 0.0
 0.5 2.0E-6 0.1 0.0
 1.0 1.0E-5 0.0 0.0 /
PVTW
 270.0 1.0 5e-05 0.3 0.0 /
DENSITY
 860.0
 1030.0
 0.8 /
PVTO
 0.001 15.0    1.0 0.1 /
 1.6   5000.0  1.8 0.5
                                    9000.0  1.7 0.6/
 /

SOLUTION
EQUIL
 1000
 4800
 1000
 0
 1000
 0
 0
 0
 0
 /
SUMMARY
FOPR
SCHEDULE
RPTSCHED
 'PRES'/
RPTRST
 'BASIC=1' /
TSTEP
 366 /
END
```
</details>

TEST.GRDECL
--------------------
<details>

```
SPECGRID
 2 2 2 1 F
 /
COORD
 0.0 0.0 0.0 0.0 0.0 3.0
 1.0 0.0 0.0 1.0 0.0 3.0
 2.0 0.0 0.0 2.0 0.0 3.0
 6.123233995736766e-17 1.0 0.0 6.123233995736766e-17 1.0 3.0
 1.0 1.0000000000000002 0.0 1.0 1.0000000000000002 3.0
 2.0 1.0000000000000002 0.0 2.0 1.0000000000000002 3.0
 1.2246467991473532e-16 2.0 0.0 1.2246467991473532e-16 2.0 3.0
[...]

PR #4415, opened 2023-01-31
fix the running when solution does not change between time steps

GitPaean:
The PID time step control will not pose any constraint on the time step size by returning a largest possible time step size.

I am not that familiar with this part, but hopefully this can be the start to fix the corner issues suggested in https://github.com/OPM/opm-simulators/issues/4341 .

### DIFF
--- a/opm/simulators/timestepping/TimeStepControl.cpp
+++ b/opm/simulators/timestepping/TimeStepControl.cpp
@@ -27,10 +27,14 @@
 #include <string>
 #include <fstream>
 #include <iostream>
+#include <limits>
 
 #include <opm/common/ErrorMacros.hpp>
+#include <opm/common/OpmLog/OpmLog.hpp>
 #include <opm/input/eclipse/Units/Units.hpp>
 #include <opm/simulators/timestepping/TimeStepControl.hpp>
+
+#include <fmt/format.h>
 
 namespace Opm
 {
@@ -141,16 +145,21 @@ namespace Opm
         errors_[ 2 ] = error;
         for( int i=0; i<2; ++i ) {
             assert(std::isfinite(errors_[i]));
-            assert(errors_[i]>0);
         }
 
-        if( error > tol_ )
+        if( errors_[2] > tol_ )
         {
             // adjust dt by given tolerance
             const double newDt = dt * tol_ / error;
-            if( verbose_ )
-                std::cout << "Computed step size (tol): " << unit::convert::to( newDt, unit::day ) << " (days)" << std::endl;
+            if ( verbose_ )
+                    OpmLog::info(fmt::format("Computed step size (tol): {} days", unit::convert::to( newDt, unit::day )));
             return newDt;
+        }
+        else if (errors_[1] == 0 || errors_[2] == 0.)
+        {
+            if ( verbose_ )
+                OpmLog::info("The solution between time steps does not change, there is no time step constraint from the PID time step control ");
+            return std::numeric_limits<double>::max();
         }
         else
         {
@@ -162,7 +171,7 @@ namespace Opm
                                  std::pow( tol_         / errors_[ 2 ], kI ) *
                                  std::pow( errors_[0]*errors_[0]/errors_[ 1 ]/errors_[ 2 ], kD ));
             if( verbose_ )
-                std::cout << "Computed step size (pow): " << unit::convert::to( newDt, unit::day ) << " (days)" << std::endl;
+                OpmLog::info(fmt::format("Computed step size (pow): {} days", unit::convert::to( newDt, unit::day )));
             return newDt;
         }
     }


GitPaean:
jenkins build this please

opm/simulators/timestepping/TimeStepControl.cpp
```diff
@@ -152,6 +153,12 @@ namespace Opm
...
         }
+        else if ( error == 0.)
+        {
+            if ( verbose_ )
+                std::cout << "The solution between time step does not change, there is no time step constraint from the PID time step control " << std::endl;
+            return std::numeric_limits<double>::max();
+        }
         else
         {
             // values taking from turek time stepping paper
```
GitPaean, line 164: For here, if `error==0`, `error_[2]` will be zero here. For more complicated scenarios, `errors_[1]` can be zero, the `else` branch will also fail.

GitPaean:
jenkins build this please

opm/simulators/timestepping/TimeStepControl.cpp
```diff
@@ -141,17 +142,23 @@ namespace Opm
...
             const double newDt = dt * tol_ / error;
             if( verbose_ )
                 std::cout << "Computed step size (tol): " << unit::convert::to( newDt, unit::day ) << " (days)" << std::endl;
             return newDt;
         }
+        else if (errors_[1] * errors_[2] == 0.)
+        {
+            if ( verbose_ )
+                std::cout << "The solution between time step does not change, there is no time step constraint from the PID time step control " << std::endl;
+            return std::numeric_limits<double>::max();
```
blattms, line 160: I am wondering whether the return value would ever be used directly or only as upper bound when using the minimum (like below or in AdaptiveTimeSteppingEbos.hpp). Assuming the latter this seems ok.

blattms (approved):
I should probably read the paper, but this looks sound to me as long the value returned by computeTimeStepSize is never used directly.
A second opinion would not hurt though.

Unrelated: maybe could change the usage of `std::cout` to OpmLog?

opm/simulators/timestepping/TimeStepControl.cpp
```diff
@@ -141,17 +142,23 @@ namespace Opm
...
             // adjust dt by given tolerance
             const double newDt = dt * tol_ / error;
             if( verbose_ )
                 std::cout << "Computed step size (tol): " << unit::convert::to( newDt, unit::day ) << " (days)" << std::endl;
             return newDt;
         }
+        else if (errors_[1] * errors_[2] == 0.)
+        {
+            if ( verbose_ )
+                std::cout << "The solution between time step does not change, there is no time step constraint from the PID time step control " << std::endl;
```
bska, line 159: It's tangential to the theme of the PR, but do you think we can switch these to using the logging system instead?  Messages output to `cout` will just vanish unless the user takes extra care to capture a log file.

opm/simulators/timestepping/TimeStepControl.cpp
```diff
@@ -141,17 +142,23 @@ namespace Opm
...
-        if( error > tol_ )
+        if( errors_[2] > tol_ )
         {
             // adjust dt by given tolerance
             const double newDt = dt * tol_ / error;
             if( verbose_ )
                 std::cout << "Computed step size (tol): " << unit::convert::to( newDt, unit::day ) << " (days)" << std::endl;
             return newDt;
         }
+        else if (errors_[1] * errors_[2] == 0.)
```
bska, line 156: Exact floating point equality comparisons, even with zero, always make me worry that we're missing something.  Is there no other way of phrasing this condition?

opm/simulators/timestepping/TimeStepControl.cpp
```diff
@@ -141,17 +142,23 @@ namespace Opm
         errors_[ 2 ] = error;
         for( int i=0; i<2; ++i ) {
             assert(std::isfinite(errors_[i]));
-            assert(errors_[i]>0);
+            assert(errors_[i]>=0);
```
bska, line 145: This `assert()` is basically tautologous now.  The `RelativeChangeInterface` [stipulates](https://github.com/OPM/opm-simulators/blob/5daf355a7b25b52850404bf3fa96aedad14cf648/opm/simulators/timestepping/TimeStepControlInterface.hpp#L36-L37) that this relation always holds.  Whether we choose to keep the `assert()` is a philosophical discussion I think.

bska (commented):
There are a couple of minor points here that I'd like to discuss, but overall I think this can go in more or less as is.

opm/simulators/timestepping/TimeStepControl.cpp
GitPaean: Yes, we can. I will do that.  I did not do it because it was not the initial attention for this PR. It will be addressed.

opm/simulators/timestepping/TimeStepControl.cpp
GitPaean: You are right, it is not needed anymore.

opm/simulators/timestepping/TimeStepControl.cpp
GitPaean: I know, it looks worrying. But technically, it really only causes problem when one of these two values are exactly 0.  If it is some extremely small values, the results will be the same with the current approach (most likely, while not excluding extraordinarily designed  situations).  

Alternatively, maybe we can use something like the following as the if condition here:
```c++ 
if (errors_[1] < std::numeric_limits<double>::min() || 
     errors_[2] < std::numeric_limits<double>::min()) 
``` 
Like me know what do you think.

GitPaean:
jenkins build this please

bska (approved):
Thanks for the updates.  This is better than the original version and I'll merge into master.

Issue #4410, opened 2023-01-30
Where are the command line options printed?


blattms:
I am probably making a fool out of myself, but anyway:

What file did we print all the parameters used to? Somehow I (and grep) can't find them anymore.

GitPaean:
I use the following, not sure it contains **all** the parameters. 

```
./flow_blackoil --help
Usage: ./flow_blackoil [OPTIONS] [ECL_DECK_FILENAME]
Flow, an advanced reservoir simulator for ECL-decks provided by the Open Porous Media project.

Recognized options:
    -h,--help                                     Print this help message and exit
    --help-all                                    Print all parameters, including obsolete, hidden and deprecated ones.
    --accelerator-mode=STRING                     Choose a linear solver, usage: '--accelerator-mode=[none|cusparse|opencl|amgcl|rocalution]'. Default: "none"
    --allow-distributed-wells=BOOLEAN             Allow the perforations of a well to be distributed to interior of multiple processes. Default: false
    --alternative-well-rate-init=BOOLEAN          Use alternative well rate initialization procedure. Default: true
    --bda-device-id=INTEGER                       Choose device ID for cusparseSolver or openclSolver, use 'nvidia-smi' or 'clinfo' to determine valid IDs. Default: 0
    --continue-on-convergence-error=BOOLEAN       Continue with a non-converged solution instead of giving up if we encounter a time step size smaller than the minimum time step size. Default: false
    --cpr-reuse-interval=INTEGER                  Reuse preconditioner interval. Used when CprReuseSetup is set to 4, then the preconditioner will be fully recreated instead of reused every N linear solve, where
                                                    N is this parameter. Default: 30
    --cpr-reuse-setup=INTEGER                     Reuse preconditioner setup. Valid options are 0: recreate the preconditioner for every linear solve, 1: recreate once every timestep, 2: recreate if last linear
                                                    solve took more than 10 iterations, 3: never recreate, 4: recreated every CprReuseInterval. Default: 4
    --dbhp-max-rel=SCALAR                         Maximum relative change of the bottom-hole pressure in a single iteration. Default: 1
    --dp-max-rel=SCALAR                           Maximum relative change of pressure in a single iteration. Default: 0.3
    --ds-max=SCALAR                               Maximum absolute change of any saturation in a single iteration. Default: 0.2
    --dwell-fraction-max=SCALAR                   Maximum absolute change of a well's volume fraction in a single iteration. Default: 0.2
[...]

PR #4418, opened 2023-01-31
fixed: regression from 82d2146c90ce74677bd8f04c745c31cd08b430de

akva2:
missing PRT header

### DIFF
--- a/opm/simulators/flow/FlowMainEbos.hpp
+++ b/opm/simulators/flow/FlowMainEbos.hpp
@@ -343,7 +343,6 @@ namespace Opm
         }
 
         // Print an ASCII-art header to the PRT and DEBUG files.
-        // \return Whether unkown keywords were seen during parsing.
         static void printPRTHeader(bool output_cout)
         {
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

bska (approved):
Thanks, this is a good work-around and I'll merge into master.  I suppose an additional step might be to move the `printPRTHeader()` function out of `FlowMainEbos.hpp`, and to have the new function take a callback to print the actual parameter values to a `std::ostream&` but I haven't thought through the implications yet.

For now we'll just use this to restore the feature.

blattms:
*sarcastic mode on*
What a wonderful declarative subject! So clear what this fixes in no time. :wink: 
*sarcastic mode off*

Really: Good subjects save *a lot* of valuable time in a collaborative projects.

Maybe close #4410 ?

akva2:
It tells you it fixes a regression, what commit caused the regression and what the regression was. Should I also give my grandmother's life story?

PR #4425, opened 2023-02-02
EclGenericProblem: add serialization support

akva2:
Sits on top of https://github.com/OPM/opm-simulators/pull/4422

For restart.

PR #4426, opened 2023-02-02
BlackoilAquiferModel: add serialization of dynamic state

akva2:
Sits on top of https://github.com/OPM/opm-simulators/pull/4425
Downstream of https://github.com/OPM/opm-common/pull/3388

### DIFF
--- a/opm/simulators/aquifers/BlackoilAquiferModel_impl.hpp
+++ b/opm/simulators/aquifers/BlackoilAquiferModel_impl.hpp
@@ -18,6 +18,9 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+#include <opm/common/ErrorMacros.hpp>
+#include <stdexcept>
+
 namespace Opm
 {
 
@@ -182,6 +185,27 @@ data::Aquifers BlackoilAquiferModel<TypeTag>::aquiferData() const
         data.insert_or_assign(aqu->aquiferID(), aqu->aquiferData());
 
     return data;
+}
+
+template<typename TypeTag>
+template<class Serializer>
+void BlackoilAquiferModel<TypeTag>::
+serializeOp(Serializer& serializer)
+{
+    for (auto& aiPtr : aquifers) {
+        auto* ct = dynamic_cast<AquiferCarterTracy<TypeTag>*>(aiPtr.get());
+        auto* fetp = dynamic_cast<AquiferFetkovich<TypeTag>*>(aiPtr.get());
+        auto* num = dynamic_cast<AquiferNumerical<TypeTag>*>(aiPtr.get());
+        if (ct) {
+            serializer(*ct);
+        } else if (fetp) {
+            serializer(*fetp);
+        } else if (num) {
+            serializer(*num);
+        } else {
+            OPM_THROW(std::logic_error, "Error serializing BlackoilAquiferModel: unknown aquifer type");
+        }
+    }
 }
 
 } // namespace Opm
--- a/tests/test_RestartSerialization.cpp
+++ b/tests/test_RestartSerialization.cpp
@@ -170,6 +170,24 @@ BOOST_AUTO_TEST_CASE(AquiferCarterTracy)
     BOOST_CHECK_MESSAGE(data_out == data_in, "Deserialized AquiferCarterTracy differ");
 }
 
+BOOST_AUTO_TEST_CASE(AquiferFetkovich)
+{
+    using TT = Opm::Properties::TTag::EbosTypeTag;
+    Opm::EclGenericVanguard::readDeck("GLIFT1.DATA");
+    using Simulator = Opm::GetPropType<TT, Opm::Properties::Simulator>;
+    Simulator sim;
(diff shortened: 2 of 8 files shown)


akva2:
jenkins build this opm-models=780 opm-common=3388 please

akva2:
note the test part is slightly suboptimal. i ran out of patience trying to grok the parametric macro to account for the different constructor signature for AquiferNumerical.

akva2:
jenkins build this please

tests/test_RestartSerialization.cpp
```diff
@@ -134,6 +134,64 @@ BOOST_AUTO_TEST_CASE(EclGenericProblem)
...
+
+struct AquiferFixture {
+    AquiferFixture() {
+        using TT = Opm::Properties::TTag::EbosTypeTag;
+        const char* argv[] = {
+            "test_RestartSerialization",
+            "--ecl-deck-file-name=GLIFT1.DATA"
+        };
+        Opm::setupParameters_<TT>(2, argv, /*registerParams=*/true);
+        Opm::EclGenericVanguard::setCommunication(std::make_unique<Opm::Parallel::Communication>());
```
bska, line 147: Does `Parallel::Communication` work in sequential mode?  If not, we should probably protect this with `HAVE_MPI` or similar.

opm/simulators/aquifers/BlackoilAquiferModel_impl.hpp
```diff
@@ -184,4 +184,23 @@ data::Aquifers BlackoilAquiferModel<TypeTag>::aquiferData() const
...
+        auto* ct = dynamic_cast<AquiferCarterTracy<TypeTag>*>(aiPtr.get());
+        auto* fetp = dynamic_cast<AquiferFetkovich<TypeTag>*>(aiPtr.get());
+        auto* num = dynamic_cast<AquiferNumerical<TypeTag>*>(aiPtr.get());
+        if (ct) {
+            serializer(*ct);
+        } else if (fetp) {
+            serializer(*fetp);
+        } else if (num) {
+            serializer(*num);
+        }
```
bska, line 202: This should probably blow up if neither of the `dynamic_cast<>`s succeed&mdash;e.g., if we get a new aquifer type and we forget to add support for it here.

bska (commented):
I think this is okay, but I've forgotten enough about `Parallel::Communication` to be completely sure that this works in sequential mode.  Please elaborate.

tests/test_RestartSerialization.cpp
akva2: It's handled by a dummy in dune so it works fine (mostly). The issue we have encountered sometimes is that the dummy communication is not castable to Dune::No_Comm, while the MPI communicator is castable to MPI_Comm. In this case this is not relevant so no ifdef needed.

opm/simulators/aquifers/BlackoilAquiferModel_impl.hpp
akva2: yeah, added an assertion.

akva2:
jenkins build this please

opm/simulators/aquifers/BlackoilAquiferModel_impl.hpp
```diff
@@ -184,4 +184,25 @@ data::Aquifers BlackoilAquiferModel<TypeTag>::aquiferData() const
...
+        auto* fetp = dynamic_cast<AquiferFetkovich<TypeTag>*>(aiPtr.get());
+        auto* num = dynamic_cast<AquiferNumerical<TypeTag>*>(aiPtr.get());
+        if (ct) {
+            serializer(*ct);
+        } else if (fetp) {
+            serializer(*fetp);
+        } else if (num) {
+            serializer(*num);
+        } else {
+            assert(0);
```
bska, line 203: While we don't do it right now, I believe there is ongoing work to re-enable `NDEBUG` in Release mode&mdash;i.e., to [disable all `assert()`s](https://github.com/OPM/opm-common/blob/63bfa8e47cbc1742b20a5b513ffcbe10658ca6f1/cmake/Modules/UseOptimization.cmake#L66-L69)  when building in Release mode.  In that case, we probably want more protection than just `assert(false)` here.

akva2:
that being said, you made me nervous, so let's just make sure ;)

akva2:

bska (commented):
Thanks for the updates.  With the possible caveat about `assert(false)`&mdash;and the sequential mode build check succeeding&mdash;I think this is ready for merging now.

opm/simulators/aquifers/BlackoilAquiferModel_impl.hpp
akva2: hmm. i'm not a fan of throwing due to logic/programming errors, but i see you point. exception it is then.

akva2:
serial build was green.

akva2:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  This looks good to me now.  Please merge when the current build check is green.

PR #4432, opened 2023-02-07
added: HDF5 serializer

akva2:
this takes the serialization data and stores it in a hdf5 file,  alternatively reads the data and deserializes.
will be used for restarting purposes.

Downstream of https://github.com/OPM/opm-common/pull/3383

### DIFF
--- a/ebos/hdf5serializer.hh
+++ b/ebos/hdf5serializer.hh
@@ -1,0 +1,132 @@
+/*
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+
+  Consult the COPYING file in the top-level source directory of this
+  module for the precise wording of the license and the list of
+  copyright holders.
+*/
+#ifndef ECL_HDF5_SERIALIZER_HH
+#define ECL_HDF5_SERIALIZER_HH
+
+#include <opm/common/utility/Serializer.hpp>
+#include <opm/common/utility/MemPacker.hpp>
+
+#include <opm/simulators/utils/HDF5File.hpp>
+#include <opm/simulators/utils/moduleVersion.hpp>
+
+#include <algorithm>
+#include <cctype>
+#include <cstdlib>
+
+namespace Opm {
+
+//! \brief Class for (de-)serializing using HDF5.
+class HDF5Serializer : public Serializer<Serialization::MemPacker> {
+public:
+    HDF5Serializer(const std::string& fileName, HDF5File::OpenMode mode)
+        : Serializer<Serialization::MemPacker>(m_packer_priv)
+        , m_h5file(fileName, mode)
+    {}
+
+    //! \brief Serialize and write data to restart file.
+    //! \tparam T Type of class to write
+    //! \param data Class to write restart data for
+    template<class T>
+    void write(T& data,
+               const std::string& group,
+               const std::string& dset)
+    {
+        try {
+            this->pack(data);
+        } catch (...) {
+            m_packSize = std::numeric_limits<size_t>::max();
+            throw;
+        }
+
+        m_h5file.write(group, dset, m_buffer);
+    }
+
+    //! \brief Writes a header to the file.
+    //! \param simulator_name Name of simulator used
+    //! \param module_version Version of simulator used
+    //! \param time_stamp Build time-stamp for simulator used
+    //! \param case_name Name of case file is associated with
+    //! \param params List of parameter values
+    //! \param num_procs Number of processes used
+    void writeHeader(const std::string& simulator_name,
+                     const std::string& module_version,
+                     const std::string& time_stamp,
+                     const std::string& case_name,
+                     const std::string& params,
+                     int num_procs)
+    {
+        try {
+            this->pack(simulator_name, module_version, time_stamp,
+                       case_name, params, num_procs);
+        } catch (...) {
+            m_packSize = std::numeric_limits<size_t>::max();
+            throw;
+        }
+        m_h5file.write("/", "simulator_info", m_buffer);
+    }
+
+    //! \brief Read data and deserialize from restart file.
+    //! \tparam T Type of class to read
+    //! \param data Class to read restart data for
+    template<class T>
+    void read(T& data,
+              const std::string& group,
+              const std::string& dset)
+    {
+        m_h5file.read(group, dset, m_buffer);
+        this->unpack(data);
+    }
+
+    //! \brief Returns the last report step stored in file.
+    int lastReportStep() const
+    {
+        const auto entries = m_h5file.list("/report_step");
+        int last = -1;
+        for (const auto& entry : entries) {
+            int num = std::atoi(entry.c_str());
+            last = std::max(last, num);
+        }
+
+        return last;
+    }
+
+    //! \brief Returns a list of report steps stored in restart file.
+    std::vector<int> reportSteps() const
+    {
+        const auto entries = m_h5file.list("/report_step");
+        std::vector<int> result(entries.size());
+        std::transform(entries.begin(), entries.end(), result.begin(),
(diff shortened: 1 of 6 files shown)


akva2:
jenkins build this opm-common=3383 please

ebos/hdf5serializer.hh
```diff
@@ -0,0 +1,103 @@
...
+            throw;
+        }
+
+        m_h5file.write(group, dset, m_buffer);
+    }
+
+    //! \brief Writes a header to the file.
+    //! \param simulator_name Name of simulator used
+    //! \param module_version Version of simulator used
+    //! \param time_stamp Built time-stap for simulator used
```
atgeirr, line 62: Typo: stap->stamp

ebos/hdf5serializer.hh
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    //! \brief Returns the last group (report step) stored in file.
+    int lastGroup() const
+    {
+        return m_h5file.lastGroup();
+    }
+
+private:
+    const Serialization::MemPacker m_packer; //!< Packer instance
```
atgeirr, line 97: As the Serializer base class already hold a reference to a packer using the same variable name, there is a risk of confusion. I suggest renaming this m_packer_private or something similar.

atgeirr (commented):
A few comments aside, looks good.

opm/simulators/utils/HDF5File.cpp
```diff
@@ -123,4 +123,23 @@ void HDF5File::read(const std::string& group,
     H5Dclose(dataset_id);
 }
 
+int HDF5File::lastGroup() const
```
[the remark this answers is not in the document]
akva2: 1) nothing bad, atoi would return -1 and it won't be > 0 so it's just a nop.  
still, i have revised this to add a generic listing function to the hdf5file class, so we can keep the specifics in the serializer class. also i've added some sanity stuff on the serializer side.
furthermore let's stick the data as /report_step/n/xxx to minimize the chance of this happening.

2) yeah, added

3) that's all available in the hdf5 c++ bindings if we really need it, no reason for us to implement it. i've stuck to the C api for so long since the c++ api was troublesome back in 2004 when I started using HDF5. I assume it's fine now, but old habits and all that jazz..

4) lambda's (and std::function) are designed to be 100% compatible with C function callbacks. the compiler does all the magic.

ebos/hdf5serializer.hh
akva2: fixed.

ebos/hdf5serializer.hh
akva2: done

akva2:
jenkins build this please

opm/simulators/utils/HDF5File.cpp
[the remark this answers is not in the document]
bska: > lambda's (and `std::function`) are designed to be 100% compatible with C function callbacks. the compiler does all the magic.

Only if there are no captures/internal state.  If there is internal state in the lambda, then converting to a C-style function pointer is undefined behaviour.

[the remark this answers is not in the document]
akva2: correct, i wasn't entirely precise.

ebos/hdf5serializer.hh
```diff
@@ -0,0 +1,114 @@
...
+    int lastReportStep() const
+    {
+        const auto entries = m_h5file.list("/report_step");
+        std::vector<std::string> filtered;
+        std::copy_if(entries.begin(), entries.end(), std::back_inserter(filtered),
+                     [](const std::string& s)
+                     {
+                         return std::all_of(s.begin(), s.end(), ::isdigit);
+                     });
+        std::sort(filtered.begin(), filtered.end());
```
atgeirr, line 103: This will sort "9" after "10" so that does not look quite correct to me. Suggest:
```
int last = -1;
for (const auto& entry : entries) {
    int num = std::atoi(entry.c_str()));
    last = std::max(last, num);
}
```

opm/simulators/utils/HDF5File.cpp
```diff
@@ -74,23 +76,40 @@ HDF5File::~HDF5File()
...
 void HDF5File::write(const std::string& group,
                      const std::string& dset,
-                     const std::vector<char>& buffer)
+                     const std::vector<char>& buffer) const
 {
     hid_t grp;
     if (groupExists(m_file, group)) {
         grp = H5Gopen2(m_file, group.c_str(), H5P_DEFAULT);
     } else {
+        auto grps = split_string(group, '/');
```
atgeirr, line 85: This is a bugfix for the write() function, to handle nested groups?

ebos/hdf5serializer.hh
akva2: arr yes, that was stupid. i'll do the loop.

opm/simulators/utils/HDF5File.cpp
akva2: well, yes, or, it's not so much that we couldn't write to nested group but to write to group a/b/ you have to create a before you create a/b.

akva2:
jenkins build this please

opm/simulators/utils/HDF5File.cpp
```diff
@@ -123,4 +142,23 @@ void HDF5File::read(const std::string& group,
...
+    auto&& list_group = [] (hid_t, const char* name, void* data) -> herr_t
+    {
+        auto& list = *static_cast<std::vector<std::string>*>(data);
+        list.push_back(name);
+        return 0;
+    };
+
+    int idx = 0;
+    std::vector<std::string> result;
+    if (H5Giterate(m_file, group.c_str(), &idx, list_group, &result) < 0) {
```
atgeirr, line 157: Reading the doc for H5Giterate() at https://docs.hdfgroup.org/hdf5/v1_12/group___h5_g.html#title18 I see that it is deprecated in favour of H5Literate1(), which in turn is deprecated in favour of H5Literate2(). Maybe we should use those?

Also, the newer functions take a parameter for the ordering. It is not clear to me what order will be used with the deprecated H5Giterate(): alphabetic, creation order, or undefined? In any case, the unit test assumes a particular order and I am not sure it can be assumed.

tests/test_HDF5File.cpp
```diff
@@ -94,3 +94,33 @@ BOOST_AUTO_TEST_CASE(WriteExistentDset)
...
+    {
+        Opm::HDF5File in_file(rwpath, Opm::HDF5File::OpenMode::READ);
+
+        auto res1 = in_file.list("/");
+        BOOST_CHECK_EQUAL(res1.size(), 1u);
+        BOOST_CHECK_EQUAL(res1[0], "test_data");
+
+        auto res2 = in_file.list("/test_data");
+        BOOST_CHECK_EQUAL(res2.size(), 3u);
+        BOOST_CHECK_EQUAL(res2[0], "d1");
```
atgeirr, line 118: This section of the test may fail unless one may make ordering assumptions for H5Giterate(), see previous comment.

opm/simulators/utils/HDF5File.hpp
```diff
@@ -53,7 +53,7 @@ class HDF5File {
     //! \details Throws exception on failure
     void write(const std::string& group,
                const std::string& dset,
-               const std::vector<char>& buffer);
+               const std::vector<char>& buffer) const;
```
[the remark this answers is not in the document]
akva2: no reason other than class itself not being mutated so i though may as well let the compiler know. but i agree that it's not really your typical immutable situation, so i'll change it back.

opm/simulators/utils/HDF5File.cpp
```diff
@@ -123,4 +142,23 @@ void HDF5File::read(const std::string& group,
     H5Dclose(dataset_id);
 }
 
+std::vector<std::string> HDF5File::list(const std::string& group) const
+{
+    // Lambda function pushing the group entries to
```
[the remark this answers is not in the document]
akva2: yup, fixed.

opm/simulators/utils/HDF5File.cpp
akva2: this used to be a bit tricky, I used to have to related to 1.6 which didn't have H5Literate. but i guess it's okay now. 

ubuntu 20.04 and rhel7 offers 1.8.x, ubuntu 22.04 and debian 11 offer 1.10.x and future distros are likely to offer 1.12.x or possibly even 1.15. But it seems H5Literate_by_name is available in 1.8, and is a macro in newer releases calling onto H5Literate_by_name1/2 in later releases, so I guess I can use it.

H5Giterate is always in lex order so the test is safe as such, but now it's obviously better since it's explicit.

tests/test_HDF5File.cpp
akva2: taken care of now.

akva2:
jenkins build this please

atgeirr:
Thanks, merging!

PR #4436, opened 2023-02-08
SegmentState: add serialization support

akva2:
Sits on top of https://github.com/OPM/opm-simulators/pull/4435

PR #4437, opened 2023-02-08
SingleWellState: add serialization support

akva2:
Sits on top of https://github.com/OPM/opm-simulators/pull/4436

### DIFF
--- a/opm/simulators/wells/SingleWellState.cpp
+++ b/opm/simulators/wells/SingleWellState.cpp
@@ -55,6 +55,12 @@ SingleWellState::SingleWellState(const std::string& name_,
     }
 }
 
+SingleWellState SingleWellState::serializationTestObject(const ParallelWellInfo& pinfo)
+{
+    SingleWellState result("testing", pinfo, true, 1.0, {}, PhaseUsage{}, 2.0);
+
+    return result;
+}
 
 void SingleWellState::init_timestep(const SingleWellState& other) {
     if (this->producer != other.producer)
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this opm-models=780 opm-common=3388 please

akva2:
Note: this is likely not needed. I'll leave it around for now.

akva2:
jenkins build this please

opm/simulators/wells/SingleWellState.cpp
```diff
@@ -276,9 +282,27 @@ void SingleWellState::update_targets(const Well& ecl_well, const SummaryState& s
         this->update_injector_targets(ecl_well, st);
 }
 
+bool SingleWellState::operator==(const SingleWellState& rhs) const
+{
+    return this->name == rhs.name &&
```
atgeirr, line 287: The `serializeOp()` does not serialize all members, but leaves out the `parallel_info`, `pu`, and `perf_data` members. All of these will be set in the (only) constructor, so one is guaranteed to have a valid object after deserialization in the sense of all members being assigned.

However, the `operator==()` is not comparing those three members. So one can have very different SingleWellState objects that compare equal, by for example passing in different `perf_input` arguments. I think we should also compare the three left-out members: `parallel_info` maybe by pointer (not 100% sure about that) and the others by their values.

I think the `pu` and `perf_data` should therefore be added to the serialization.

tests/test_RestartSerialization.cpp
```diff
@@ -96,6 +97,21 @@ namespace Opm {
...
 
+BOOST_AUTO_TEST_CASE(SingleWellState)
+{
+    Opm::ParallelWellInfo dummy;
+    auto data_out = Opm::SingleWellState::serializationTestObject(dummy);
+    Opm::Serialization::MemPacker packer;
+    Opm::Serializer ser(packer);
+    ser.pack(data_out);
+    const size_t pos1 = ser.position();
+    decltype(data_out) data_in("", dummy, false, 0.0, {}, Opm::PhaseUsage{}, 0.0);
```
atgeirr, line 108: Passing in a non-empty perf_input here would demonstrate my point above.

atgeirr (commented):
Some questions here, before it is ready.

opm/simulators/wells/SingleWellState.cpp
atgeirr: PerfData seems quite straightforward, and also PhaseUsage, and they are held by value here. I understand that parallel_info is trickier.

akva2: phase usage must be passed through the constructor and if those are not the same across runs we have bigger problems. the perf_data is recreated when a new report step is initialized so it's pointless to serialize it. the parallel well info is a reference so i cannot deserialize it. it is also recreated on the start of report steps so it would be pointless to store it.

tests/test_RestartSerialization.cpp
atgeirr: To sum up my initial reluctance here: I think it is a bit dangerous to create a serializeOp that is essentially only useful for that one particular use (restarting). However, if we ever have another use case here (passing the state to another MPI rank for example), we would need to refactor the class and construction anyway, so I think it is acceptable. Will merge.

PR #4438, opened 2023-02-08
ParallelRestart: use variadic broadcast

akva2:
micro optimization, there is no reason to do this in two operations


akva2:
jenkins build this please

bska (approved):

PR #4440, opened 2023-02-08
GroupState: add serialization support

akva2:
Sits on top of https://github.com/OPM/opm-simulators/pull/4437
Downstream of https://github.com/OPM/opm-common/pull/3392

### DIFF
--- a/opm/simulators/wells/WellContainer.hpp
+++ b/opm/simulators/wells/WellContainer.hpp
@@ -45,13 +45,21 @@ namespace Opm {
 template <class T>
 class WellContainer {
 public:
-
     WellContainer() = default;
-
 
     WellContainer(std::initializer_list<std::pair<std::string,T>> init_list) {
         for (const auto& [name, value] : init_list)
             this->add(name, value);
+    }
+
+    static WellContainer serializationTestObject(const T& data)
+    {
+        WellContainer<T> result;
+
+        result.m_data = {data};
+        result.index_map = {{"test1", 1}, {"test2", 4}};
+
+        return result;
     }
 
     bool empty() const {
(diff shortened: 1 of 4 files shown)


akva2:
jenkins build this opm-models=780 opm-common=3392 please

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

atgeirr (commented):
The changes look good. But I noticed that the equality operator is not comparing gpmaint_state, probably overlooked when that was added earlier!

akva2:
nice catch. addressed here and in https://github.com/OPM/opm-common/pull/3402

akva2:
jenkins build this opm-common=3402 please

opm/simulators/wells/WellContainer.hpp
```diff
@@ -45,15 +45,27 @@ namespace Opm {
 template <class T>
 class WellContainer {
 public:
-
     WellContainer() = default;
 
+    WellContainer(std::size_t size)
+        : m_data(size)
+    {}
```
bska, line 52: Would you mind marking this as `explicit` please?

bska (commented):
This looks good for the most part, but I think I'd prefer an `explicit` constructor if at all possible.

opm/simulators/wells/WellContainer.hpp
akva2: i'll do even better. i forgot to update my git on the laptop when i did the gpmaint_state fix and thus pushed an old version of this branch. this constructor is in fact not needed at all, and the groupstate one can be defaulted.

akva2:
jenkins build this please

bska (approved):
We might also consider adding a test for something like
```C++
struct X { std::vector<double> x1; std::array<float, 3> x2; };
WellContainer<X>
```
That's a little more representative of the kinds of types we typically stick in the `WellContainer`.  In the meantime I'll merge this into the master branch.

PR #4448, opened 2023-02-09
EclGenericTracerModel: remove unused tracerResidual_ member

akva2:


akva2:
jenkins build this please

bska (approved):
Good.  If we actually need this, or something like it, in the future we can reintroduce it then.  I'll merge into master.

akva2:
it's a refactoring artifact, it's stored in the TracerBatch struct, https://github.com/OPM/opm-simulators/blob/master/ebos/ecltracermodel.hh#L555

PR #4452, opened 2023-02-10
RestartSerialization: add parallel support

akva2:
This extends the restart serialization to support parallel simulations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,10 @@ endif()
 
 find_package(opm-common REQUIRED)
 
+if(USE_MPI)
+  set(HDF5_PREFER_PARALLEL TRUE)
+endif()
+
 include(OpmInit)
 OpmSetPolicies()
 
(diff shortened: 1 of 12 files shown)


akva2:
jenkins build this please

CMakeLists.txt
```diff
@@ -91,6 +91,10 @@ endif()
 
 find_package(opm-common REQUIRED)
 
+if(USE_MPI)
+  set(HDF5_PREFER_PARALLEL TRUE)
+endif()
```
bska, line 96: This is probably the right thing to do in general, but CMake has an [issue](https://gitlab.kitware.com/cmake/cmake/-/issues/19117) that's been open for three years stating that `PREFER_PARALLEL` doesn't really do what it's supposed to.

What's your experience?

opm/simulators/utils/HDF5File.cpp
```diff
@@ -76,63 +96,71 @@ HDF5File::~HDF5File()
...
-
-    hsize_t size = buffer.size();
-    hsize_t start = 0;
-
-    hid_t space = H5Screate_simple(1, &size, nullptr);
-    hid_t dataset_id = H5Dcreate2(grp, dset.c_str(), H5T_NATIVE_CHAR, space,
-                       H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
-    if (dataset_id == H5I_INVALID_HID) {
-        throw std::runtime_error("HDF5File: Trying to write already existing dataset '" + group + '/' + dset + "'");
+        throw std::runtime_error("Failed to create group '" + realGroup + "'");
```
bska, line 139: What happens if a reservoir engineer sees the diagnostic "Failed to create group ..."?  Are they going to associate that with HDF5 or will there be confusion?

bska (commented):
Just some high level comments, since I'm not familiar with the HDF5 API.

CMakeLists.txt
akva2: i haven't had any issues with it on debian derivates. that anaconda creates issues is business as usual.

opm/simulators/utils/HDF5File.cpp
akva2: i was a bit torn here. the "issue" is that this now (mostly) used in outer parallel exception handling macros which adds a prefix to the message. So in practice the exception would read "Error saving serialized state: Failed to create group '..'"

CMakeLists.txt
bska: > i haven't had any issues with it on debian derivates.

Understood.  Then I'm happy to have this.

opm/simulators/utils/HDF5File.cpp
bska: > in practice the exception would read "Error saving serialized state: Failed to create group '..'"

Ah, okay.  Then this is fine with me.  We'll merge this now and then adjust diagnostics and the overall implementation as need arises.

PR #4457, opened 2023-02-14
changes to make executable using original simulator

hnil:
- small changes to make the standard flow classes run with the pure simulator in opm-models.
-  this make an executable which is check can be used to check if the interface of the classes is consistent with opm-model design. 
-  This simulator is not checked as flow and probably have several issues
Known issues;
- match of
- some opening checks may be missing
- timestepping is not the same as in flow

Should the name be changed to not conflict with flow.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -914,7 +914,7 @@ public:
         // to the first "real" episode/report step
         // for restart the episode index and start is already set
         if (!initconfig.restartRequested()) {
-            simulator.startNextEpisode(schedule.seconds(0));
+            simulator.startNextEpisode(schedule.seconds(1));
             simulator.setEpisodeIndex(0);
         }
     }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -441,13 +441,13 @@ opm_add_test(flow_distribute_z
   $<TARGET_OBJECTS:moduleVersion>
   )
 
-opm_add_test(flow_blackoil_ebos
+opm_add_test(ebos
   ONLY_COMPILE
   ALWAYS_ENABLE
   DEPENDS opmsimulators
   LIBRARIES opmsimulators
   SOURCES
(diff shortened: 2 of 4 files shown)


akva2:
the target name will cause issues in certain generators as it is already taken for the simulator object for the blackoil simulator. can we use another name?

hnil:
Probably it should also have a different name to not be confused with being equal to flow at least for now.
What would a good name be?
ebos_eclblackoil?

akva2:
ebos i guess, 'ebos = eclipse blackoil simulator'. i'm curious why we are reintroducing it though.

hnil:
Resonse to reintroduce it:
- check that the original design still work with the problem
- simple data flow for tagbased standalone
- this simulator is checked for dynamic refinement and loadbalancing, for other problem
- maybe a step to remove a simulator with a simulator inside

The original design make it possible to change most things, without touching to much code. In the new case changing problem,simulator, model and newton loop can be done as one want to test tings. The current code is a bit more restrictive since things are a bit hard coded for the flow executable.

bska:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -914,7 +914,7 @@ public:
         // to the first "real" episode/report step
         // for restart the episode index and start is already set
         if (!initconfig.restartRequested()) {
-            simulator.startNextEpisode(schedule.seconds(0));
+            simulator.startNextEpisode(schedule.seconds(1));
```
bska, line 917: Why do you need this particular change?  Does the other simulator not run properly with a zero-sized step length passed into [`startNextEpisode()`](https://github.com/OPM/opm-models/blob/f1601ed1a833ef2f39109c96b1ac78473f08b4df/opm/models/utils/simulator.hh#L543-L548)?

hnil: Else dt=0 at start of simulator.

PR #4460, opened 2023-02-15
FIX VREF for groups with both gas and water injectors

totto82:


totto82:
jenkins build this please

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -351,32 +351,32 @@ namespace WellGroupHelpers
...
                 for (Phase phase : all) {
+                    int phase_pos = 0;
+                    switch (phase) {
+                    case(Phase::GAS):
+                        phase_pos = pu.phase_pos[BlackoilPhases::Vapour];
+                        break;
+                    case(Phase::OIL):
+                        phase_pos = pu.phase_pos[BlackoilPhases::Liquid];
+                        break;
+                    case(Phase::WATER):
```
bska, line 363: Tiny formatting nit: You don't need the parentheses around `Phase::x` here.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -351,32 +351,32 @@ namespace WellGroupHelpers
...
+                        phase_pos = pu.phase_pos[BlackoilPhases::Aqua];
+                        break;
+                    }
                     const Group::InjectionCMode& currentGroupControl
                             = group_state.injection_control(subGroup.name(), phase);
-                    individual_control = individual_control || (currentGroupControl != Group::InjectionCMode::FLD
+                    bool individual_control = (currentGroupControl != Group::InjectionCMode::FLD
```
bska, line 369: Please mark `individual_control` as `const` if it does not change later.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -351,32 +351,32 @@ namespace WellGroupHelpers
...
-                    }
-                } else {
-                    // The subgroup may participate in group control.
-                    bool has_guide_rate = false;
-                    for (Phase phase : all) {
-                        has_guide_rate = has_guide_rate || guide_rate.has(subGroupName, phase);
-                    }
-
-                    if (!has_guide_rate) {
+                    int num_group_controlled_wells
```
bska, line 371: Please mark `num_group_controlled_wells` as `const`.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -351,32 +351,32 @@ namespace WellGroupHelpers
...
+                            = groupControlledWells(schedule, wellState, group_state, reportStepIdx, subGroupName, "", !isInjector, phase);
+                    if (individual_control || num_group_controlled_wells == 0) {
+                        groupTargetReduction[phase_pos]
+                            += subGroupEfficiency * sumWellSurfaceRates(subGroup, schedule, wellState, reportStepIdx, phase_pos, isInjector);
+                    } else {
                         // Accumulate from this subgroup only if no group guide rate is set for it.
-                        for (int phase = 0; phase < np; phase++) {
-                            groupTargetReduction[phase] += subGroupEfficiency * subGroupTargetReduction[phase];
+                        if(!guide_rate.has(subGroupName, phase)) {
```
bska, line 378: Please add a space between `if` and the opening `parenthesis`.

bska (commented):
I don't immediately see what's wrong, but the [failing tests](https://ci.opm-project.org/job/opm-simulators-PR-builder/4540/testReport/) appear to point to some deeper issue and not just "fix vref" changing some reference solution.

Other than that, there are a couple of minor nits that I'd like to address.

totto82:
> I don't immediately see what's wrong

I only tested it for 3-phase. I will update with a fix later today.

totto82:
jenkins build this please

bska (approved):
Okay, I think I understand what's happening here.  Nevertheless, I think I would have preferred if the description included more detail than "fix vref..."  For instance,

1. What was the problem?
2. How does this change address that problem?
3. Any other approaches considered, if applicable

That said, I'll merge into master to fix the underlying problem

OPMUSER:
@totto82 I cannot document this bug fix without a clear description of what this fixes, I don't know even what the hell VREF is supposed to be.

In the future, I would appreciate that PRs are not merged unless there is a description of what the intent of PR is.

totto82:
> What was the problem?

We didn't support VREF group injection control if the group contains multiple subgroups with both water and gas injectors correctly. 
> How does this change address that problem?

The group reduction rate (i.e. the total rate of wells not under group control but in the group) is now correctly computed also for sub-groups.  
> Any other approaches considered, if applicable

This was a bug fix. So no.

OPMUSER:
Maybe my brain has stopped working for today, but what is VREF?

totto82:
Sorry VREP! It is my brain that hasn't started yet.

OPMUSER:
Now I understand, thanks for the additional commentary, will update the manual accordingly.

PR #4461, opened 2023-02-15
Fix default alq injector

totto82:
Use 0.0 as default alq for injector.

### DIFF
--- a/opm/simulators/wells/WellInterfaceGeneric.cpp
+++ b/opm/simulators/wells/WellInterfaceGeneric.cpp
@@ -410,6 +410,10 @@ bool WellInterfaceGeneric::thpLimitViolatedButNotSwitched() const
 
 double WellInterfaceGeneric::getALQ(const WellState& well_state) const
 {
+    // no alq for injectors.
+    if (isInjector())
+        return 0.0;
+
     return well_state.getALQ(name());
 }
 


totto82:
jenkins build this please

opm/simulators/wells/ALQState.cpp
```diff
@@ -38,7 +38,8 @@ double ALQState::get(const std::string& wname) const {
     if (default_iter != this->default_alq_.end())
         return default_iter->second;
 
-    throw std::logic_error("No ALQ value registered for well: " + wname);
+    // injectors does not have a default alq return 0.0
+    return 0.0;
```
bska, line 42: Could we solve this another way, e.g., by not calling the function if the well is an injector?  Blindly returning zero seems a little presumptuous and could end up masking real errors.

bska (commented):
I suppose as a quick fix this is probably okay, but I really think we need to solve this at a higher level, when we have more information about the well object.

opm/simulators/wells/ALQState.cpp
totto82: Ok, I will try to add isProducer guards to the call sites

totto82:
jenkins build this please

bska:
> jenkins build this please

I don't see any changes compared to the first version of the patch.  Did you forget to push?

totto82:
jenkins build this please

opm/simulators/wells/ALQState.cpp
totto82: We have some code that takes alq as input and is called fro both injectors and producers. For instance https://github.com/OPM/opm-simulators/blob/70152b67ff2d9d9fc9b9136e5c3bb444a51b4000/opm/simulators/wells/WellBhpThpCalculator.cpp#L101

It is possible to refactor those, but maybe the solution of returning 0 with getAlq is called from the well models is sufficient?

totto82: Also https://github.com/OPM/opm-simulators/blob/70152b67ff2d9d9fc9b9136e5c3bb444a51b4000/opm/simulators/wells/WellBhpThpCalculator.cpp#L226
takes alq as input for both injectors and producers

totto82:
@bska I think the current solution in the PR (i.e returning zero for injectors) is the simplest and best solution. If you agree, I think this PR should be merged since it fixes a bug in the code.

bska:
> think the current solution in the PR (i.e returning zero for injectors) is the simplest and best solution.

It's definitely the simplest option, and I it's probably good enough.  I haven't looked at how the current code uses the ALQ value and it may be that we'll need to distinguish "ALQ present" from "ALQ zero" at some point.  If that happens I think using an `optional<double>` to represent the ALQ value would be a reasonable strategy.

bska (approved):
I'll merge this into master now.

PR #4462, opened 2023-02-15
Add timing block

hnil:
Add timing blocks which can be used with tracy to get timings. Examples are in the attatchment.
It is same type of pullrequests in opm-common opm-models

 
image
image
image
image

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1198,14 +1198,15 @@ public:
      */
     void writeOutput(bool verbose = true)
     {
-        OPM_TIMEBLOCK(writeOutput);
+        OPM_TIMEBLOCK(problemWriteOutput);
         // use the generic code to prepare the output fields and to
         // write the desired VTK files.
         ParentType::writeOutput(verbose);
 
         bool isSubStep = !EWOMS_GET_PARAM(TypeTag, bool, EnableWriteAllSolutions) && !this->simulator().episodeWillBeOver();
-        if (enableEclOutput_)
+        if (enableEclOutput_){
             eclWriter_->writeOutput(isSubStep);
+        }
     }
 
     void finalizeOutput() {
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -972,6 +972,7 @@ public:
      */
     void beginEpisode()
     {
+        OPM_TIMEBLOCK(beginEpisode);
         // Proceed to the next report step
         auto& simulator = this->simulator();
         int episodeIdx = simulator.episodeIndex();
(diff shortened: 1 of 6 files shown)


bska:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -972,6 +972,7 @@ public:
      */
     void beginEpisode()
     {
+        OPM_TIMEBLOCK(beginEpisode);
```
atgeirr, line 975: I think this should indicate that it is the beginEpisode of EclProblem, since there are many classes that have a beginEpisode method (right?). I do not know if the Tracy macros allow arbitrary names, but `EclProblem::beginEpisode` would be my preferred version, `EclProblem_beginEpisode` if that is not allowed.

The same applies to begin/end Iteration, Timestep etc.

I see you already used a similar construction below: `eclProblemSource`, I would prefer either `EclProblem::source` or `EclProblem_source` to have a direct match between names.

opm/simulators/flow/BlackoilModelEbos.hpp
```diff
@@ -642,7 +643,10 @@ namespace Opm {
                                                     // residual
 
             // if the solution is updated, the intensive quantities need to be recalculated
-            ebosSimulator_.model().invalidateAndUpdateIntensiveQuantities(/*timeIdx=*/0);
+            {
+                OPM_TIMEBLOCK(invalidateAndUpdateIntensiveQuantities);
```
atgeirr, line 647: Better to have this block name in the opm-models implementation, as that may be called from other places than here.

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -237,6 +237,7 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
               converged_(false),
               matrix_()
         {
+            OPM_TIMEBLOCK(istlSolverEbos);
```
atgeirr, line 240: I think it is clearer to use `ISTLSolverEbos_constructor`.

atgeirr (approved):
I have a suggestion for a naming scheme that will make it easier to separate various function with the same name, but other than that I think this should go in.

ebos/eclproblem.hh
hnil: :: is not allowed. The other is ok. There is beginEpisode problems. In the original design it should only be in the problem, but our models (i.e. aquifers and wells are kind of a mix).

opm/simulators/linalg/ISTLSolverEbos.hpp
hnil: I can change it, but I think it is better to make it simple until more people have used it and see it is a problem.

atgeirr:
Closing, replaced by #4495.

PR #4465, opened 2023-02-16
changed: extract thread queue / worker handling to generic class

akva2:
for reuse purposes.

in particular i want to have similar functionality for the serialization if we are to enable compression.

this is very much a draft still, but atleast i can see what you think..


akva2:
not going this route.

atgeirr:
> not going this route.

What is the other route you want to take?

akva2:
no threading, compression level 1.

PR #4466, opened 2023-02-16
Fix Sequential Build

bska:
Include `<cassert>` for 'assert' macro.


bska:

bska:
Not sufficient to restore the build.  I'll close this now.  I'll also revert #4452 if there's no fix rather soon.

PR #4468, opened 2023-02-17
Fixing MSW connection pressure output

vkip:
Fixing connection pressure output for the MSW model (currently the grid block pressures corrected to perforation depth is output instead).

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -787,10 +787,15 @@ namespace Opm
         // pressure difference between the perforation and the grid cell
         const double cell_perf_press_diff = this->cell_perforation_pressure_diffs_[perf];
 
-        perf_press = pressure_cell - cell_perf_press_diff;
+        // perforation pressure is the wellbore pressure corrected to perforation depth
+        // (positive sign due to convention in segments_.perforation_depth_diff() )
+        perf_press = segment_pressure + perf_seg_press_diff;
+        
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

vkip:
To see the problem, add the following to the SUMMARY section in opm-tests/MSW_MODEL_1.DATA 

CPR 
  '*' /
/

Attached image demonstrates that the new version corresponds to the reference simulator. 
image

bska (commented):
I think this looks like the correct fix, but I'd like @GitPaean to have a look too.  The [regression failures](https://ci.opm-project.org/job/opm-simulators-PR-builder/4568/testReport/) are all because we store the connection pressure in the restart file's `XCON` array and they all look reasonable.  We'll have to update the reference solutions in order to merge this.  I will do that once the change itself is approved.

GitPaean (approved):
Thanks a lot for the contribution and it looks correct to me. With the verification against reference results, this is the way to go!

GitPaean:
We ran into the same question when implementing the WINJMULT for MSW, which requires the perforation pressure.  It looks like we adopted the same formula as in this PR, while we did not think about fix the output related to CPR back then. 

Very great it gets fixed now for the output.

GitPaean:
Jenkins build this update_data please

bska:
jenkins build this opm-tests=899 please

bska:
The new reference solutions have been installed on the CI system.  Merging this now to bring the overall system back to a consistent state.

PR #4470, opened 2023-02-19
Add Regression Test for NOPINCH Feature

bska:
Thanks a lot to [at]tskille for providing a test case (OPM/opm-tests#886)

According to [earlier comments](https://github.com/OPM/opm-tests/pull/886#issuecomment-1435968137) the simulator solves this case with sufficient accuracy to enable putting it into the suite of regression tests.


bska:

bska:
jenkins build this opm-tests=901 please

bska:
New reference solutions installed on the CI system.  Adding this case to the simulator's regression test suite.

PR #4473, opened 2023-02-20
[fix] For parallel build with serial hdf5 issue warning and skip hdf5.

blattms:
We rely on parallel hdf5 for a parallel build of OPM flow. Otherwise compilation will fail. Hence if this is a parallel build but only a serial hdf5 library is found, we now issue an informative warning and deactivate hdf5/parallel restart.

Closes OPM/opm-common#3405

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,10 +247,6 @@ if(ROCALUTION_FOUND)
 endif()
 
 
-# read the list of components from this file (in the project directory);
-# it should set various lists with the names of the files to include
-include (CMakeLists_files.cmake)
-
 macro (config_hook)
   opm_need_version_of ("dune-common")
   opm_need_version_of ("dune-istl")
@@ -282,6 +278,22 @@ macro (fortran_hook)
 endmacro (fortran_hook)
 
 macro (files_hook)
+  if(MPI_FOUND AND HDF5_FOUND AND NOT HDF5_IS_PARALLEL)
+    message(WARNING "When building parallel OPM flow we need a "
+      "parallel version of hdf5, but found only a serial one. "
+      "Please install a parallel hdf5 library for MPI "
+      "(e.g with apt-get install libhdf5-mpi-dev) and do a clean "
+      "rebuild (build after \"make clean\"). Continuing with "
+      "only normal restart without hdf5 file support.")
+    set(HDF5_FOUND OFF)
+    unset(HAVE_HDF5)
+  endif()
+  # read the list of components from this file (in the project directory);
+  # it should set various lists with the names of the files to include
+  # include needs to be here to make reset HDF5_FOUND available in
+  # (CMakeLists_files.cmake and because we need the created lists during
+  # then inclusion of OpmLibMain
+  include (CMakeLists_files.cmake)
 endmacro (files_hook)
 
 macro (tests_hook)


blattms:
jenkins build this please

blattms:
Can't persuade the build system to not define HAVE_HDF5 in config.h :angry:

CMakeLists.txt
```diff
@@ -291,6 +291,16 @@ endmacro (tests_hook)
 # all setup common to the OPM library modules is done here
 include (OpmLibMain)
 
+if(MPI_FOUND AND HDF5_FOUND AND NOT HDF5_IS_PARALLEL)
+  message(WARNING "When building parallel OPM flow we need a "
+    "parallel version of hdf5, but found only a serial one. "
+    "Please install a parallel hdf5 library for MPI "
+    "(e.g with apt-get install libhdf5-mpi-dev) and do a clean "
+    "rebuild (build after \"make clean\". Continuing without"
+    "hdf5 and parallel restart.")
```
bska, line 300: Technically we still have the original restart files, so maybe we can say
```
Disabling HDF5-based restart files
```
or something along those lines to avoid confusion?  In any case there should be a space after "without" and before "hdf5".  You should probably close the parentheses following "(build after..." too.

blattms: Changed to "Continuing with only normal restart (not hdf5 enabled one)"

blattms:
found it. Needs to be in file_hook to also prevent hdf5 files from being compiled.

Build systems: You hate them all, yet you can't live without them. :laughing:

blattms:
jenkins build this please

blattms:
Finally. Much harder work than expected. File lists needed to be modified in file_hook, too.

blattms:
jenkins build this please

CMakeLists.txt
```diff
@@ -282,6 +282,32 @@ macro (fortran_hook)
 endmacro (fortran_hook)
 
 macro (files_hook)
+  if(MPI_FOUND AND HDF5_FOUND AND NOT HDF5_IS_PARALLEL)
+    message(WARNING "When building parallel OPM flow we need a "
+      "parallel version of hdf5, but found only a serial one. "
+      "Please install a parallel hdf5 library for MPI "
+      "(e.g with apt-get install libhdf5-mpi-dev) and do a clean "
+      "rebuild (build after \"make clean\". Continuing with "
```
bska, line 290: Tiny nit, but you should close the parenthesis starting at "(build after".

CMakeLists.txt
```diff
@@ -282,6 +282,32 @@ macro (fortran_hook)
...
+    unset(HAVE_HDF5)
+  endif()
+
+  # CMakeLists_files.cmake is already processed before this hook.
+  # Hence we need to add all HDF5 specific files here to skip them
+  # if we do not find a parallel HDF5 in a parallel build.
+  if(HDF5_FOUND)
+    list(APPEND MAIN_SOURCE_FILES opm/simulators/utils/HDF5File.cpp)
+    list(APPEND TEST_SOURCE_FILES tests/test_HDF5File.cpp)
+    list(APPEND TEST_SOURCE_FILES tests/test_HDF5Serializer.cpp)
```
bska, line 302: Since you're moving these anyway, is there a reason not to coalesce the requests, i.e.,
```cmake
list(APPEND TEST_SOURCE_FILES
  tests/test_HDF5File.cpp
  tests/test_HDFSerializer.cpp
)
```
instead?

bska (commented):
One nit and a question, otherwise this looks good to me from a very brief look.  Could you
```cmake
set(HDF5_FOUND CACHE OFF)
```
in the MPI and no parallel HDF5 library case to avoid having to move the HDF5 support to `files_hook`?

blattms:
> One nit and a question, otherwise this looks good to me from a very brief look. Could you
> 
> ```cmake
> set(HDF5_FOUND CACHE OFF)
> ```
> 
> in the MPI and no parallel HDF5 library case to avoid having to move the HDF5 support to `files_hook`?

I don't see how caching is related to this or would help and it seems rather unconventional, but I think you have a point: We can move this code somewhere else as we do not use HDF5_FOUND in CMakeLists_files.cmake anymore. Hence having it in files_hook is just a result of my rather desperate trial-and-error fighting with the buildsystem. I still would like to find a way to change this setting of the find but before CMakeLists_files.cmake is read.

akva2:
files_hook is the intended hook for this so from my point of view this is spot on.

blattms:
But somehow I feel that we include CMakeLists_files.cmake to soon. Seems like that is done without any magic in CMakeLists.txt directly. I'll try to move it to a later point and keep CMakeLists_files.cmake as before.

akva2:
yeah, that might be, as that is included before any hooks are run. if you prefer, you can move the CMakeLists_files.txt include into file_hook.

blattms:
Thanks that was the missing link.

CMakeLists.txt
blattms: Not an issue anymore as it stays as it was before PR:

CMakeLists.txt
blattms: done

akva2 (approved):

PR #4477, opened 2023-02-22
WIP add WCD, GCDI, GCDM, WIPG, WIPL summary for F and R

totto82:
These are summary output that is used for CO2STORE simulations.

Depends on OPM/opm-common#3408

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -1032,6 +1032,7 @@ private:
         if (this->computeFip_) {
             Scalar fip[FluidSystem::numPhases];
             Scalar fipr[FluidSystem::numPhases]; // at reservoir condition
+
             for (unsigned phaseIdx = 0; phaseIdx < FluidSystem::numPhases; ++phaseIdx) {
                 fip[phaseIdx] = 0.0;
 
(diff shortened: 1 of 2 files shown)


ebos/ecloutputblackoilmodule.hh
```diff
@@ -1077,12 +1078,50 @@ private:
...
+                    this->fip_[Inplace::Phase::WaterInWaterPhase][globalDofIdx] = fip[waterPhaseIdx];
+
+                // Add dissolved gas and vaporized water to total Fip
+                if (!this->fip_[Inplace::Phase::WATER].empty())
+                    this->fip_[Inplace::Phase::WATER][globalDofIdx] += waterInPlaceGas;
+                if (!this->fip_[Inplace::Phase::GAS].empty())
+                    this->fip_[Inplace::Phase::GAS][globalDofIdx] += gasInPlaceWater;
+            }
+
+            Scalar criticalGas = 0.1; //TODO get this from the tables
```
totto82, line 1113: @bska What is the best way to get the critical gas saturation here? Make it a member in this class and initialize from eclState via tables and satprop code?

bska: > What is the best way to get the critical gas saturation here

Just to be sure: By "critical gas saturation" you mean the saturation at which the gas phase becomes mobile&mdash;i.e., exceeds the critical relative permeability tolerance ("TOLCRIT")?  If so, you typically need to care about end-point scaling and/or hysteresis which varies by cell and/or system state.  In that case, I think the most general approach is to go through
```
simulator_.problem().materialLawManager()
```
and effectively do the same that we do when computing the relative permeability.  Storing this as a data member will just duplicate what we already have.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1077,12 +1078,52 @@ private:
...
+                    this->fip_[Inplace::Phase::WaterInWaterPhase][globalDofIdx] = fip[waterPhaseIdx];
+
+                // Add dissolved gas and vaporized water to total Fip
+                if (!this->fip_[Inplace::Phase::WATER].empty())
+                    this->fip_[Inplace::Phase::WATER][globalDofIdx] += waterInPlaceGas;
+                if (!this->fip_[Inplace::Phase::GAS].empty())
+                    this->fip_[Inplace::Phase::GAS][globalDofIdx] += gasInPlaceWater;
+            }
+            const auto& scaledDrainageInfo
+                = simulator_.problem().materialLawManager()->oilWaterScaledEpsInfoDrainage(globalDofIdx);
```
bska, line 1113: Are we intentionally deferring imbibition/hysteresis?

bska (commented):
This looks good to me.  With the caveat that I don't know how/if this works for the imbibition case I'm prepared to merge this.  On a tangential note, I didn't notice that the immobile CO2 in the gas phase was tagged as `CO2InGasPhaseInMob` using the name `InMob`.  At some point we might consider renaming that to `CO2InGasPhaseImmob` or similar instead.

bska:

bska:
I need to merge this now to restore the build following OPM/opm-common#3408 being merged.  I'm still a little concerned about what to do with the imbibition curve, but we'll work on that as a refinement.  For now we'll use the current version.

PR #4478, opened 2023-02-22
fixed: output OPMRST file to output dir

akva2:
will look in both output dir and input file dir on load


akva2:
jenkins build this please

atgeirr:
Good fix, will merge.

> will look in both output dir and input file dir on load

I imagine a new output dir will typically be used for the restarted run, but any added convenience is good!

PR #4479, opened 2023-02-22
RestartSerialization: add missing PerfData

akva2:
Turns out that while perfdata is recreated for the current well state, it is not so for the previous ones. Thus we have to serialize this in order to have it available for the previous well state upon restart.

### DIFF
--- a/opm/simulators/wells/PerfData.cpp
+++ b/opm/simulators/wells/PerfData.cpp
@@ -24,9 +24,7 @@
 
 #include <opm/simulators/wells/PerfData.hpp>
 
-namespace Opm
-{
-
+namespace Opm {
 
 PerfData::PerfData(std::size_t num_perf, double pressure_first_connection_, bool injector_, std::size_t num_phases)
     : injector(injector_)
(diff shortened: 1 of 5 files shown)


akva2:
jenkins build this please

opm/simulators/wells/PerfData.cpp
```diff
@@ -51,6 +49,29 @@ PerfData::PerfData(std::size_t num_perf, double pressure_first_connection_, bool
...
+    result.brine_rates = {13.0};
+    result.prod_index = {14.0, 15.0};
+    result.micp_rates = {16.0};
+    result.cell_index = {17, 18, 19, 20};
+    result.connection_transmissibility_factor = {21.0};
+    result.satnum_id = {22, 23};
+    result.ecl_index = {24};
+    result.water_throughput = {25.0, 26.0};
+    result.skin_pressure = {27.0, 28.0};
+    result.water_velocity = {29.0, 30.0};
```
bska, line 70: I think this is okay as a serialisation test.  It's not immediately clear to me if it honours all of the implicit size requirements of the `PerfData` structure and we may need to revise the setup later if, for instance, it ends up breaking some kind of checked iterator build.

bska (approved):
Thanks a lot.  With the possible caveat that the `serializationTestObject()` **might** break some size assumptions I'll merge this into the master branch.  We'll revise if the need arises.

PR #4481, opened 2023-02-22
HDF5File: enable gzip compression

akva2:
I am keeping this in draft state, but I want to open so we can discuss a little.

Enabling gzip compression (level 9 - the highest compression effort) brings a serialized Norne state down from 43MB to 6.5MB, so there is certainly merit to this.

However, obviously compression eats cpu time, and doing this on the simulator thread will stall the simulator for seconds.
So I have 3 questions;

1) Should we enable this (i think yes).
2) Should we make it configurable (I think maybe not).
3) Should I put effort into moving the compression and storage onto a work thread, akin to the convergence output stuff?

### DIFF
--- a/opm/simulators/utils/HDF5File.cpp
+++ b/opm/simulators/utils/HDF5File.cpp
@@ -197,36 +197,41 @@ void HDF5File::writeSplit(hid_t grp,
                           const std::string& dset) const
 {
     std::vector<hsize_t> proc_sizes(comm_.size());
+    hid_t dxpl = H5P_DEFAULT;
     if (comm_.size() > 1) {
         hsize_t lsize = buffer.size();
         comm_.allgather(&lsize, 1, proc_sizes.data());
+        dxpl = H5Pcreate(H5P_DATASET_XFER);
+        H5Pset_dxpl_mpio(dxpl, H5FD_MPIO_COLLECTIVE);
     } else {
         proc_sizes[0] = buffer.size();
     }
 
     for (int i = 0; i < comm_.size(); ++i) {
         hid_t space = H5Screate_simple(1, &proc_sizes[i], nullptr);
+        hid_t dcpl = this->getCompression(proc_sizes[i]);
         hid_t dataset_id = H5Dcreate2(grp,
                                       std::to_string(i).c_str(),
                                       H5T_NATIVE_CHAR, space,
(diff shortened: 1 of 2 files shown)


alfbr:
Should it be parallel from day one? (I think yes)

akva2:
Not entirely sure what you mean with parallel? If you mean mpi, yes, it is. If you mean threaded, then vote noted.

alfbr:
I mean the gzipping, so that is relate to point 3 I guess. There are parallel compression/decompression algorithms out there, so I am simply asking if we should use a parallel algorithm for the compression/decompression.

akva2:
hdf5 supports gzip and szip. Latter is only really usable for floats and we store mixed data. So it is really just gzip or not, and its compression level.

alfbr:
Ah, I see, so parallel compression/decompression is basically not an option.

akva2:
Well you can implement custom filters so i am sure it is doable, but it would a bigger effort than just enabling the existing filter.

atgeirr:
If level 9 compression takes a lot of time, maybe we can find a sweet spot at lower compression levels, where it is still fast enough to not hold back the simulation, but still gives good space saving? Spinning off a separate thread would also somewhat address this, but that is a bigger job to do for a later PR.

akva2:
# Serialization times
|compression level|size|  runtime  |
|----------------:|----|----------:|
|                0|43M |0.345364808|
|                1|7.7M|0.787305235|
|                2|7.5M|0.798170751|
|                3|7.2M|0.857215990|
|                4|7.2M|0.950212850|
|                5|7.0M|1.064487231|
|                6|6.8M|1.393071489|
|                7|6.7M|1.540442490|
|                8|6.6M|2.952842035|
|                9|6.5M|6.592072060|

indeed some rather diminishing returns..

atgeirr:
> indeed some rather diminishing returns..

I think level 1 should be good enough.

akva2:
okay, set level to 1.

akva2:
jenkins build this please

atgeirr:
I think this looks good, although I am not sufficiently knowledgeable about the collective IO things in MPI to assess those details, so in my view that is experimental until proven by usage and testing. Merging!

Will this make Flow incompatible with existing .OPMRST files, or will they be read nicely, just not using the DEFLATE algorithm filter?

akva2:
Decompression is automagic so no breakage.

PR #4482, opened 2023-02-23
Add gaswater energy simulator

totto82:
Needed in order to run https://github.com/OPM/opm-tests/pull/903

The plan is to add it to the flow executable after some more testing, but the stand alone simulator should be merged first.


totto82:
jenkins build this please

akva2:
looks like some git adds are missing.

totto82:
jenkins build this please

atgeirr:
Merging this, but the FLOW_SUPPORT_AMG define is weird, as it seems to be unused anywhere else in the system, other than here and the other gas-water simulator instance definitions. @plgbrts can you comment on this? It looks like you originally added it.

Issue #4476, opened 2023-02-22
THP controlled producers can converged at injecting


GitPaean:
It looks like THP controlled production wells can end up with converged with injecting. 

If the production well is injecting, with the master branch, when calculating BHP from the THP limit, it will be cut off to be zero rate. The ratios can be calculated without issue.  Then the control equation might get converged when the bhp is the bhp value with zero rate from the VFP table. 

At the same time, the other operabiity checking strategy did not catch this situation either.  When group control and production network is involved, the subsequent effect is unpredictable.

GitPaean:
closed with #4483.

PR #4483, opened 2023-02-23
not considering wells with wrong flow direction as converged

GitPaean:
under THP constraint, well can get converged with wrong flow direction (e.g. a producer is injecting), becasue we cut off the wrong flow direction as zero rate in the VFP evaluation

### DIFF
--- a/opm/simulators/wells/MultisegmentWellEval.cpp
+++ b/opm/simulators/wells/MultisegmentWellEval.cpp
@@ -164,6 +164,19 @@ getWellConvergence(const WellState& well_state,
                                   report,
                                   deferred_logger);
 
+    // for stopped well, we do not enforce the following checking to avoid dealing with sign of near-zero values
+    // for BHP or THP controlled wells, we need to make sure the flow direction is correct
+    if (!baseif_.wellIsStopped() && baseif_.isPressureControlled(well_state)) {
+        // checking the flow direction
+        const double sign = baseif_.isProducer() ? -1. : 1.;
+        const auto weight_total_flux = this->primary_variables_.getWQTotal() * sign;
+        constexpr int dummy_phase = -1;
+        if (weight_total_flux < 0.) {
+            report.setWellFailed(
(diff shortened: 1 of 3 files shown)


GitPaean:
This way looks like the easiest way to enforce in term of coding. The original version uses the density to calculate the mass rate, that intervenes with the well code organization and indices appear still to be challenging.

GitPaean:
jenkins build this please

GitPaean:
As mentioned in https://github.com/OPM/opm-simulators/issues/4476, a wrong direction rate can have unpredictable effects when group control and network is involved.  This PR improve the case I am testing with production network.

opm/simulators/wells/MultisegmentWellEval.cpp
```diff
@@ -164,6 +164,15 @@ getWellConvergence(const WellState& well_state,
                                   report,
                                   deferred_logger);
 
+    // checking the flow direction
+    const double sign = baseif_.isProducer() ? -1. : 1.;
+    const auto weight_total_flux = this->primary_variables_.getWQTotal() * sign;
+    constexpr int dummy_phase = -1;
+    // TODO: maybe a small negative value if needed for robustness
```
atgeirr, line 171: Suggest to skip this test (also in StandardWellEval.cpp) if the well is STOPPED or has a zero rate control. Then we should not need to add an epsilon here.

atgeirr (commented):
I think this is fine, and should be merged after making the minor change for stopped wells requested.

In the future, if we make a more sophisticated way to shut wells flowing in the wrong direction, we may perhaps remove this again, but it is a good interim measure.

opm/simulators/wells/MultisegmentWellEval.cpp
GitPaean: zero rate control is a little troublesome to detect. What I can do here is that we only check BHP and THP controlled wells.

GitPaean:
jenkins build this please

GitPaean:
Thanks for comments and merging.

PR #4485, opened 2023-02-24
GSF WSF diagnostics

totto82:
Depends on https://github.com/OPM/opm-common/pull/3406

PR #4487, opened 2023-02-24
Revert "GSF WSF diagnostics"

bska:
Reverts OPM/opm-simulators#4485

This possibly broke the build.  I'll try to revert it as a test.


bska:

bska:
> This possibly broke the build.

It did not.  The problem is elsewhere&mdash;possibly #4477 which likely should have been merged along with OPM/opm-common#3408.

>  I'll try to revert it as a test.

No change.  I'll close this.

PR #4489, opened 2023-02-24
Inplace: add missing enum entries in formatting function

akva2:
This should restore build. I *think* I got the mapping right.


akva2:

bska:
PR #4477 which I merged just moments ago has the same mapping, so I'm a little surprised that this was mergable.

PR #4493, opened 2023-02-28
fixed: add missing HAVE_MPI ifs

akva2:
One never learns it seems..


akva2:

bska:
> jenkins build this serial please

Build check successful.  Merging this now to restore the sequential build.

atgeirr:
Should we simply require MPI going forward? What would be the drawbacks, and for who? @akva2 @bska @alf @blattms

bska:
> What would be the drawbacks, and for who?

For me, personally, the main drawback would be losing the ability to debug the main logic without the option of completely excluding MPI contributions.  I hardly ever run a case in sequential mode outside of a debugging context, but it's very useful to me in that specific situation.

PR #4494, opened 2023-02-28
FlowMainEbos: add a compile unit

akva2:
Small micro optimizations, but every little bit..

### DIFF
--- a/opm/simulators/flow/FlowMainEbos.cpp
+++ b/opm/simulators/flow/FlowMainEbos.cpp
@@ -1,0 +1,58 @@
+/*
+  Copyright 2013, 2014, 2015 SINTEF ICT, Applied Mathematics.
+  Copyright 2014 Dr. Blatt - HPC-Simulation-Software & Services
+  Copyright 2015 IRIS AS
+  Copyright 2014 STATOIL ASA.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <config.h>
+#include <opm/simulators/flow/FlowMainEbos.hpp>
+
+#include <opm/common/utility/String.hpp>
+
+#include <opm/simulators/utils/ParallelFileMerger.hpp>
+
+#include <algorithm>
+#include <string>
+
+namespace Opm {
+namespace detail {
+
+void mergeParallelLogFiles(std::string_view output_dir,
+                           std::string_view deckFilename,
+                           bool enableLoggingFalloutWarning)
+{
+    namespace fs = ::std::filesystem;
+    fs::path output_path(output_dir);
+    fs::path deck_filename(deckFilename);
+    std::string basename;
+    // Strip extension "." and ".DATA"
+    std::string extension = uppercase(deck_filename.extension().string());
+    if (extension == ".DATA" || extension == ".") {
+        basename = uppercase(deck_filename.stem().string());
+    } else {
+        basename = uppercase(deck_filename.filename().string());
+    }
+    std::for_each(fs::directory_iterator(output_path),
+                  fs::directory_iterator(),
+                  detail::ParallelFileMerger(output_path, basename,
+                                             enableLoggingFalloutWarning));
+}
+
+} // namespace detail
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

opm/simulators/flow/FlowMainEbos.cpp
```diff
@@ -0,0 +1,81 @@
...
+    std::for_each(fs::directory_iterator(output_path),
+                  fs::directory_iterator(),
+                  detail::ParallelFileMerger(output_path, basename,
+                                             enableLoggingFalloutWarning));
+}
+
+void handleExtraConvergenceOutput(SimulatorReport& report,
+                                  const std::string& option,
+                                  const std::string& output_dir,
+                                  const std::string& base_name)
```
bska, line 63: These can, and in my opinion should, all be `std::string_view` instead.  In that case you should `#include <string_view>` too.

opm/simulators/flow/FlowMainEbos.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+    const auto extraConvOutput = ConvergenceOutputConfiguration {
+        option,
+        R"(OutputExtraConvergenceInfo (--output-extra-convergence-info))"
```
bska, line 67: I think I'd prefer if the [`optionName`](https://github.com/OPM/opm-simulators/blob/b4d7ef85a031240f21c34f42bf199845bf4ee3f1/opm/simulators/flow/ConvergenceOutputConfiguration.hpp#L70) were passed in from the outside (as a `std::string_view`).  That way, there's a connection between the option and its name at the call site.  In the current version there's a (slight) risk that the option name string used here will get out of sync with its actual string.

Alternatively, you might choose to not use an option name at all.  That will just make the diagnostic message slightly less specific in case of failure.

opm/simulators/flow/FlowMainEbos.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+#include <opm/common/utility/String.hpp>
+
+#include <opm/simulators/flow/ConvergenceOutputConfiguration.hpp>
+
+#include <opm/simulators/utils/ParallelFileMerger.hpp>
+
+#include <algorithm>
+#include <filesystem>
+#include <fstream>
```
bska, line 34: You should probably `#include <string>` here instead of relying on transitive includes.

opm/simulators/flow/FlowMainEbos.hpp
```diff
@@ -388,30 +390,14 @@ namespace Opm
...
-                                                     EWOMS_GET_PARAM(TypeTag, bool, EnableLoggingFalloutWarning)));
+            detail::mergeParallelLogFiles(eclState().getIOConfig().getOutputDir(),
+                                          EWOMS_GET_PARAM(TypeTag, std::string, EclDeckFileName),
+                                          EWOMS_GET_PARAM(TypeTag, bool, EnableLoggingFalloutWarning));
         }
 
         void setupEbosSimulator()
         {
-            ebosSimulator_.reset(new EbosSimulator(EclGenericVanguard::comm(), /*verbose=*/false));
+            ebosSimulator_ = std::make_unique<EbosSimulator>(EclGenericVanguard::comm(), /*verbose=*/false);
```
bska, line 400: You should `#include <memory>` to bring a declaration of `std::make_unique<>()` into scope.

bska (commented):
It's not entirely clear to me how much we save in terms of compile time from doing this, but it does reduce the amount of code we have to read when trying to understand `FlowMainEbos` so that's clearly beneficial.  I'm in favour of merging this once the minor niggles have been addressed.

opm/simulators/flow/FlowMainEbos.cpp
akva2: done.

opm/simulators/flow/FlowMainEbos.cpp
akva2: done.

akva2:
my motivation was not really compile time reduction but header avoidance. now you can touch ParallelFileMerger.hpp without rebuilding simulator objects, and simulator objects do not pull in regex header.

akva2:
jenkins build this please

opm/simulators/flow/FlowMainEbos.hpp
```diff
@@ -500,20 +489,11 @@ namespace Opm
...
-
-                const auto infostep = fs::path { eclState().getIOConfig().getOutputDir() } /
-                    fs::path { eclState().getIOConfig().getBaseName() }.concat(".INFOSTEP");
-
-                std::ofstream os(infostep);
-                report.fullReports(os);
-            }
+            detail::handleExtraConvergenceOutput(report,
+                                                 R"(OutputExtraConvergenceInfo (--output-extra-convergence-info))",
+                                                 EWOMS_GET_PARAM(TypeTag, std::string, OutputExtraConvergenceInfo),
```
bska, line 494: Looks like you swapped those two lines, thereby interchanging the role of the option value and the option name.

akva2: ... cheeses, not my day.

akva2:
jenkins build this please

bska (approved):

PR #4496, opened 2023-03-01
changed: pass the version info as parameters in Banners.cpp/Main.cpp

akva2:
we want to avoid these symbols in the library.

fixes shared linking.

this might seem a bit backwards, but thing is, if we put these in the library, every time you reconfigure, every binary is rebuilt because of the compile timestamp / module revision changing.


akva2:
jenkins build this please

bska:
> if we put [version information functions] in the library, every time you reconfigure, every binary is rebuilt because of the compile timestamp / module revision changing.

That's a good point.  I'll merge this into master.

PR #4498, opened 2023-03-01
adjust to timing macro header split

akva2:
Downstream of https://github.com/OPM/opm-common/pull/3418


bska (approved):

PR #4502, opened 2023-03-02
- fixed wrong timing block

hnil:


bska:
jenkins build this please

atgeirr (approved):
This is inner loop stuff, so the LOCAL variant is preferred, agreed.

bska (approved):

PR #4503, opened 2023-03-02
fixed: output of thresholdpressure to restart file in parallel

akva2:
We need to grab this from the eclstate, not the configured data.

The configured data only holds values for the regions on the process.


akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
the joys of uninited member vars...

blattms:
I agree that this fixes the issue. But looking at [eclgenericthresholdpressure.cc#L150-L191)](https://github.com/OPM/opm-simulators/blob/153c4f486a9918c9ea3bb0af97831dcc5f89d58d/ebos/eclgenericthresholdpressure.cc#L150-L191), I am wondering whether we shouldn't take the opportunity to make sure that this code computes the correct values for all regions in parallel instead. It seem we could replace the loop with the same approach that you use and even save one grid traversal. The grid traversal seems to be there to only compute values for region interfaces that happen on this process, but I don't see why having correct values for others would pose a problem and it might save computation time.

Am I missing something?

ebos/eclgenericthresholdpressure.cc
```diff
@@ -220,6 +220,34 @@ configureThpresft_()
     }
 }
 
+template<class Grid, class GridView, class ElementMapper, class Scalar>
+std::vector<Scalar>
+EclGenericThresholdPressure<Grid,GridView,ElementMapper,Scalar>::
+getRestartVector() const
+{
+    if (!enableThresholdPressure_)
+        return {};
```
blattms, line 229: Might the size 0 of the vector be a problem?

akva2: no. the writer also checks if this is enabled, https://github.com/OPM/opm-simulators/blob/master/ebos/eclgenericwriter.cc#L466

blattms:
> the problem, and i tried that approach, is that we then miss out on values specified between regions that have no direct interaction. this is the case for NORNE, where flow between region 1 and 4 is specified.

I think we were talking passed each other. I was suggesting to use your code directly in applyExplicitThresholdPressures() as that code seems to be rather inefficienct in addition.

PR #4507, opened 2023-03-03
Mark SGWFLET as Unsupported

OPMUSER:
Although the SGOFLET and SWOFLET keywords are supposed (I'm in the process of testing these), the SGWFLET is not as there is no JSON definition. I have also corrected the manual as well.

This ready for review, as it a very minor change.

### DIFF
--- a/opm/simulators/utils/UnsupportedFlowKeywords.cpp
+++ b/opm/simulators/utils/UnsupportedFlowKeywords.cpp
@@ -551,6 +551,7 @@ const KeywordValidation::UnsupportedKeywords& unsupportedKeywords()
         {"RSSPEC", {true, std::nullopt}},
         {"SAMG", {true, std::nullopt}},
         {"SAVE", {false, std::nullopt}},
+        {"SGWFLET", {false, std::nullopt}},
         {"SKIP", {true, std::nullopt}},
         {"SKIP100", {true, std::nullopt}},
         {"SKIP300", {true, std::nullopt}},


atgeirr (approved):

atgeirr:
jenkins build this please

bska:
> the SGWFLET [keyword] is not as there is no JSON definition.

In that case the run will stop immediately at the parser stage because we encountered an unknown keyword.  That effectively means that this PR is redundant.

atgeirr:
> In that case the run will stop immediately at the parser stage because we encountered an unknown keyword. That effectively means that this PR is redundant.

I had the impression that it might be added, but then I guess also that we would implement the feature. This should do no harm though. Also, some run cases with more permissive parsing. I leave to you to decide merging or not.

bska:
> ```
> Error: Problem with keyword ROCK
> In SPE1CASE2_BRINE_GASWATER.DATA line 104.
> Internal error: Malformed floating point number 'SGWFLET'
> ```

That's a problem with `ROCK`, not a problem with `SGWFLET`.  I'm guessing you have defaulted records in your `ROCK` definition which Flow does not support.  It's a long-standing problem and we don't have a solution for it.  If you swap the order of `ROCK` and `SGWFLET` you'll absolutely get an "unknown keyword" diagnostic and the run will stop instantly.

I stand by my initial statement that this PR is redundant.

OPMUSER:
> I had the impression that it might be added, but then I guess also that we would implement the feature. This should do no harm though. Also, some run cases with more permissive parsing. I leave to you to decide merging or not.

Well it was supposed to be added with the original commit (https://github.com/OPM/opm-common/pull/2917) after some discussion, and I documented the keyword as supported.  The PR will add clarity to the error message for users.

OPMUSER:
> That's a problem with ROCK, not a problem with SGWFLET. I'm guessing you have defaulted records in your ROCK definition which Flow does not support. It's a long-standing problem and we don't have a solution for it. If you swap the order of ROCK and SGWFLET you'll absolutely get an "unknown keyword" diagnostic and the run will stop instantly.
> 
> I stand by my initial statement that this PR is redundant.

No it is not, this is the original code before any changes:
```
ROCK
-- Item 1: reference pressure (psia)
-- Item 2: rock compressibility (psi^{-1})

-- Using values from table 1 in Odeh:
	14.7 3E-6 /

-- Define here the capillairy pressures
SWFN
    .25    .0    0.0
    .7    1.0    0.0
/

-- Set the capillairy pressure to zero in this table
SGFN
    .3     .0    0.0
    .75   1.0    0.0
/
```
And this is the modified code:

```
ROCK
-- Item 1: reference pressure (psia)
-- Item 2: rock compressibility (psi^{-1})

-- Using values from table 1 in Odeh:
	14.7 3E-6 /

-- -- Define here the capillairy pressures
-- SWFN
--     .25    .0    0.0
--     .7    1.0    0.0
-- /
-- 
-- -- Set the capillairy pressure to zero in this table
-- SGFN
--     .3     .0    0.0
--     .75   1.0    0.0
-- /   
--                                                                                      
--       SGWFLET – GAS-WATER LET REL. PERMEABILITY FUNCTIONS (OPM FLOW KEYWORD)                                                                         
--                                                                                      
SGWFLET
--       SGL       SGCR     L-GAS     E-GAS     T-GAS     KRT-GAS                                                     
--       SORG      SOGCR    L-WAT     E-WAT     T-WAT     KRT-WAT                                                     -
--       L-PC      E-PC     T-PC      PCIR      PCIT                                                 
--       -------   ------   -------   -------   -------   -------                                             
         0.00000   0.0000   1.00000   1.00000   1.00000   1.00000       
         0.00000   0.0000   1.00000   1.00000   1.00000   1.00000     
         1*        1*       1*        1*        1*                 / TABLE NO. 01
```

bska:
> > That's a problem with ROCK, not a problem with SGWFLET.
> 
> No it is not

Yes it is.  If `SWGFLET` were preceded by (almost) any keyword other than `ROCK`, e.g. `DENSITY`, you'd get an "unknown keyword" diagnostic for `SWGFLET`.  The problem in this specific case is that Flow's implementation of `ROCK` has an unknown size.  If `ROCK` is followed by a **known** keyword then the parser is able to infer that all `ROCK` data must have been entered.  In this case, `ROCK` is followed by an **unknown** keyword and that's what causing the problems.

We are working to solve the "ROCK size" issue, but that will require altering some rather fundamental assumptions in the very lowest level of the parser implementation.

OPMUSER:
> We are working to solve the "ROCK size" issue, but that will require altering some rather fundamental assumptions in the very lowest level of the parser implementation.

Well until that is fixed I would suggest we merge this PR.

PR #4509, opened 2023-03-03
BlackoilModelEbos: introduce compile unit

akva2:
And various cleanups. Another micro.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -41,6 +41,7 @@ list (APPEND MAIN_SOURCE_FILES
   opm/core/props/satfunc/RelpermDiagnostics.cpp
   opm/simulators/timestepping/SimulatorReport.cpp
   opm/simulators/flow/Banners.cpp
+  opm/simulators/flow/BlackoilModelEbosHelpers.cpp
   opm/simulators/flow/countGlobalCells.cpp
   opm/simulators/flow/ConvergenceOutputConfiguration.cpp
   opm/simulators/flow/ExtraConvergenceOutputThread.cpp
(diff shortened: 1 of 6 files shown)


akva2:
jenkins build this please

atgeirr:
We have a branch touching on some of the same code, please wait a little bit before merging this.

hnil:
I think it is better to do this separate the code and the instantiation. i.e. if it really is template code then use xxx_impl.hpp and then make a xxx.cpp including impl. This make testing extensions not instantiated easier.

akva2:
i understand your point in general, but what's the chance of this using something other than double? (FY it's hard-coded to use double currently, I enable potentially using float as a byproduct of this PR..)

opm/simulators/flow/BlackoilModelEbos.cpp
```diff
@@ -0,0 +1,130 @@
...
+            }
+            for (int compIdx = 0; compIdx < numComp; ++compIdx) {
+                msg += "    CNV(";
+                msg += names[compIdx][0];
+                msg += ") ";
+            }
+            OpmLog::debug(msg);
+        }
+        std::ostringstream ss;
+        if (!terminal_header.empty()) {
```
akva2, line 97: i'm not sure if this difference between domain and global is required.

opm/simulators/flow/BlackoilModelEbos.cpp
```diff
@@ -0,0 +1,130 @@
...
+                }
+            } else if (res[ii] < 0.0) {
+                report.setReservoirFailed({types[ii], CR::Severity::Normal, compIdx});
+                if (terminal_output) {
+                    OpmLog::debug("Negative residual for " + names[compIdx] + " equation.");
+                }
+            } else if (res[ii] > tol[ii]) {
+                report.setReservoirFailed({types[ii], CR::Severity::Normal, compIdx});
+            }
+            report.setReservoirConvergenceMetric(types[ii], compIdx, res[ii]);
```
akva2, line 75: this was originally not done for subdomains but i think it was just an oversight.

akva2:
jenkins build this please

akva2:
jenkins build this please

PR #4510, opened 2023-03-03
Move two aquifer restart tests to "new" approach

akva2:
Reduce number of tests.

### DIFF
--- a/regressionTests.cmake
+++ b/regressionTests.cmake
@@ -161,7 +161,9 @@ add_test_compareECLFiles(CASENAME ctaquifer_2d_oilwater
                          SIMULATOR flow
                          ABS_TOL ${abs_tol}
                          REL_TOL ${rel_tol}
-                         DIR aquifer-oilwater)
+                         DIR aquifer-oilwater
+                         RESTART_STEP 15
+                         RESTART_SCHED true)
 
 add_test_compareECLFiles(CASENAME fetkovich_2d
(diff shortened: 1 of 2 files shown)


akva2:

akva2:
jenkins build this opm-tests=914 please

bska (commented):
I'm conflicted.  On the one hand, being able to reuse test models in a standardised way is clearly beneficial.  On the other it's harder to debug the tests when the restart model is generated on the fly.

akva2:
Right. I have nothing invested in this, but I do see a problem with having two different approaches. 

This was supposed to be the 'new' way when I added it some 2 years ago but then nothing much happened wrt the old tests. The fundamental difference, apart from the tool usage which is an implementation detail really, is of course that this way of doing restart tests compare to regression data, and not to a full run. 

While having no investment, I do want to note that in my opinion the new way is a workaround to avoid adressing the bugs/shortcoming of the restart support (yes I know, the restart format is a blackbox).

I'll close this then.

bska:
> The fundamental difference [...] is of course that this way of doing restart tests compare to regression data, and not to a full run.

Right, and that's partly what makes debugging the cases so difficult.  Our reference data is calculated by/on the CI system and it has been&mdash;and mostly continues to be&mdash;rather challenging to reproduce those solutions on local developer systems.

PR #4511, opened 2023-03-06
RateConverter: throw an exception, not a bare string

akva2:
Unless I'm missing something..

### DIFF
--- a/opm/simulators/wells/RateConverter.hpp
+++ b/opm/simulators/wells/RateConverter.hpp
@@ -328,9 +328,9 @@ namespace Opm {
 
                 // Currently we only support either gas in water or gas in oil
                 // not both
-                if(detR != 1 && detRw != 1 ) {
+                if (detR != 1 && detRw != 1) {
                     std::string msg = "only support " + std::to_string(detR) + " " + std::to_string(detR);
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this please

bska (approved):
This is fine as is.  I **might** consider using `std::range_error` instead of `std::runtime_error`, but that's possibly too finicky.  On the other hand, the actual message&mdash;introduced in #4292&mdash;is not helpful to anyone but the original author.  The part concerning "only support" **may** be useful, but the fact that both numbers refer to `detR` means we lose information that might have been useful.

I'll defer to @totto82 to fix the actual message.

akva2:
changed to range_error

bska (commented):
> changed to range_error

Thanks.  This looks good to me now.  I'll just run a build check to ensure I haven't overlooked anything.  I'll merge when that's successful.

bska:
jenkins build this please

bska:
> I'll just run a build check to ensure I haven't overlooked anything. I'll merge when that's successful.

Build check is successful.  I'll merge this into the master branch now.

PR #4513, opened 2023-03-06
fixed: remove opm-material from data update script

akva2:
Missed this when merging material into common.


bska (approved):

PR #4517, opened 2023-03-08
fixed: wrong order of OpmInit and use of the USE_MPI option

akva2:
Whadya know. I tested sibling builds and now I understand why people got in trouble with the hdf5 library selection.

Did not see this locally because I use the super-build. Did not see this on jenkins because it explicitly sets USE_MPI in its toolchain file.


akva2:
jenkins build this please

PR #4519, opened 2023-03-09
Refactor Construction of Analytic Aquifer Objects

bska:
In particular, split the 'static' aquifer object initialisation of member function `init()` into a new helper function
```
initializeStaticAquifers()
```
This is in preparation of adding a similar function to handle dynamic aquifer object initialisation from a restart file.  To that end, also add a new member function
```
createDynamicAquifers(episode_index)
```
containing the current implementation of `beginEpisode()`.  Creating the dynamic objects from a restart file then amounts to calling this function with a different 'episode_index'.  As another aid to maintainability, add a new templated member function
```
createAnalyticAquiferPointer()
```
which forms `unique_ptr<AquiferInterface>` objects for every known type of analytic aquifer.  This, in turn, requires reordering the parameters of the `AquiferConstantFlux` constructor to match those of the existing Fetkovich and Carter-Tracy types.

Finally, split the calculation of the constant flux aquifer's total flux rate out to a new helper function
```
AquiferConstantFlux::totalFluxRate()
```

PR #4520, opened 2023-03-09
Add Support for Initialising Constant Flux Aquifers from Restart

bska:
In particular, form constant flux aquifer objects from the restart step's collection if available and properly initialise their total produced volume.  To that end also calculate the local face area fraction for the aquifer's connections which handles the constant flux aquifer from the restart file in the case that the aquifer's connections are split across multiple MPI ranks.


bska:
This PR depends on, and contains, PR #4519.  It must not be merged into the master branch until the upstream PR has been merged.  I will keep this in "draft" mode until that time.

bska:
When combined with upstream PR OPM/opm-common#3437 I am able to restart the simple [`AQFLUX-01`](https://github.com/OPM/opm-tests/blob/ba3b8a2aa41796e4f48a47ec39643a0c8d04a1e9/aquifers/AQUFLUX-01.DATA).  Here are a couple of summary curves from a parallel base and restart run (restarted at [step 3](https://github.com/OPM/opm-tests/blob/ba3b8a2aa41796e4f48a47ec39643a0c8d04a1e9/aquifers/AQUFLUX-01.DATA#L272), in the middle of the [`flux = 0.12`](https://github.com/OPM/opm-tests/blob/ba3b8a2aa41796e4f48a47ec39643a0c8d04a1e9/aquifers/AQUFLUX-01.DATA#L265-L267) stage

---

Level | Curves
--- | -----
Field | Plot_1
Aquifer | Plot_2
Well  | PROD1

tskille:
This PR has been tested on a full field model in Equinor and results has been check.  Looking forward to see this PR merged into master.

opm/simulators/aquifers/AquiferConstantFlux.hpp
```diff
@@ -58,7 +58,8 @@ class AquiferConstantFlux : public AquiferInterface<TypeTag>
         , aquifer_data_            (aquifer)
         , connection_flux_         (connections_.size(), Eval{0})
     {
-        this->initializeConnections();
+        const auto connected_face_area = this->initializeConnections();
+        this->area_fraction_ = this->computeFaceAreaFraction(connected_face_area);
```
GitPaean, line 62: At least in one build, the `area_fraction_` caused problem for the parallel restarting. But I have not got clear understanding of the issue. Will post more later.

GitPaean: I need to check whether it is my building is problematic since the output is not very reasonable.

bska: > I need to check whether it is my building is problematic

Yes please.  For what it's worth, we need **something** like the area fraction for parallel restart if the set of aquifer connections happens to be split among multiple processes.

> since the output is not very reasonable.

What kind of symptoms do you see?

GitPaean: Using the following code to output inside function `computeFaceAreaFraction`, 
```c++
        const double fraction =(tot_face_area > 0.0)
            ? connected_face_area / tot_face_area
            : 0.0;

        const auto rank = this->ebos_simulator_.vanguard()
                .grid().comm().rank();
        std::cout << " id " << this->aquiferID() << " rank " << rank << " connected_face_area " << connected_face_area << " total_face_area " << tot_face_area  << " fraction " << fraction << std::endl;
```


For running with 1 process `computeFaceAreaFraction`,  (the first number is line number, you can ignore it.)
```
1234  id 2 rank 0 connected_face_area 25000 total_face_area 25000 fraction 1
1235  id 1 rank 0 connected_face_area 25000 total_face_area 25000 fraction 1
```

For running with 5 processes
```
 1252  id 2 rank 0 connected_face_area 20000 total_face_area 45000 fraction 0.444444
 1253  id 1 rank 4 connected_face_area 0 total_face_area 45000 fraction 0
 1254  id 1 rank 1 connected_face_area 0 total_face_area 45000 fraction 0
 1255  id 1 rank 2 connected_face_area 0 total_face_area 45000 fraction 0
 1256  id 1 rank 3 connected_face_area 25000 total_face_area 45000 fraction 0.555556
 1257  id 2 rank 2 connected_face_area 0 total_face_area 5000 fraction 0
 1258  id 1 rank 0 connected_face_area 0 total_face_area 5000 fraction 0
 1259  id 2 rank 3 connected_face_area 5000 total_face_area 5000 fraction 1
 1260  id 2 rank 4 connected_face_area 0 total_face_area 5000 fraction 0
 1261  id 2 rank 1 connected_face_area 0 total_face_area 5000 fraction 0
```

It is a build with clang, so I need to double check with other build. I will send you the case through slack.

GitPaean:
> Yes please. For what it's worth, we need something like the area fraction for parallel restart if the set of aquifer connections happens to be split among multiple processes.

If we always keep the accumulative values (AAQT) and total values (AAQR) the same across the processes, then we might not need this fraction.  Just need to remember to communicate when update them.  How do you think?

But I still think we should have a understanding how the tricky output was resulted.

GitPaean:
> I think that's more work than just having this fraction.

I think it is the same both in term of computation and code. We need to `sum` across the processes anyway for each time step.

bska:
> > I think that's more work than just having this fraction.
> 
> I think it is the same both in term of computation and code. We need to `sum` across the processes anyway for each time step.

Sure, but

1. Your suggestion would lead to doing that twice for each timestep whereas the area fraction is a single number updated once
2. We'd need to be very careful about updating the cumulative values correctly.

It's still not clear to me why the aquifer objects need to track their total produced volumes.  If we defer that work to the summary state (or similar), then there's no problem at all and we can dispense with all of this.

bska:
PR #4576 solves the "multiple aquifer area fraction" problem that was identified during the review and testing of this PR.  I have rebased this onto that PR and will keep it in draft mode until that other work is merged into the master branch.

bska:
jenkins build this please

atgeirr:
The upstream seems merged, can the PR be un-drafted?

bska:
> The upstream seems merged, can the PR be un-drafted?

Sadly not.  I'm getting a weird regression failures that I've not been able to analyse and resolve.

bska:
jenkins build this please

bska:
> > The upstream seems merged, can the PR be un-drafted?
> 
> Sadly not. I'm getting a weird regression failures that I've not been able to analyse and resolve.

I was finally able to drill down to the underlying issue.  It turned out to be some missing pieces and poorly sequenced operations in the upstream PR #4576.  Both that PR and this one now run through the CI system in standard mode.  I'll try a sequential build too, just to be sure.

bska:
jenkins build this please

bska:
The upstream PR #4576 has been merged.  I'll mark this as "ready for review" and run a build check.

bska:
jenkins build this please

GitPaean:
I think the previous problem is related to parallel restart. Is it possible to have parallel restart test, for example, with 4 processes?

bska:
> I think the previous problem is related to parallel restart. Is it possible to have parallel restart test, for example, with 4 processes?

I can absolutely add that, but just for clarity, my earlier [example run](https://github.com/OPM/opm-simulators/pull/4520#issuecomment-1462503507) demonstrated exactly that case running [`AQUFLUX-01`](https://github.com/OPM/opm-tests/blob/36142ea046ee6f39a50c12f209a897c465f0295d/aquifers/AQUFLUX-01.DATA) from OPM-Tests.

bska:
jenkins build this please

bska:
I've added sequential and parallel restart runs of `AQUFLUX-01` and they pass on the CI system.  That said, I'm a little concerned that I had to relax the tolerances quite a bit in order to make the tests run on my development machine.  I'm worried that I've made a mistake in the test definition, especially concerning the `TEST_ARGS`.

totto82:
I had some issues with AQUFLUX-01 in parallel. It pass when solved with higher accuracy. Could that work here as well?  See https://github.com/OPM/opm-simulators/pull/4666/files

bska:
> I had some issues with AQUFLUX-01 in parallel. It pass when solved with higher accuracy.

Yeah, possibly.  I'm also leaning towards the adaptive time-stepping procedure being [turned off](https://github.com/OPM/opm-simulators/blob/675ce03e52636ced1a7fa655e52cb0bc348c1209/tests/run-parallel-restart-regressionTest.sh#L52) in the test script for parallel restart cases possibly influencing this.  Thanks to @akva2 for the hint.

GitPaean:
> I can absolutely add that, but just for clarity, my earlier example run demonstrated exactly that case running AQUFLUX-01 from OPM-Tests.

Now I remember that the one caused the current PR fail is actually a different case with two aquifers. Let me check whether to update the case in opm-tests.  I think it is good that testing case can cover more feature and behavoir testing.

GitPaean:
I will try to add the AQUFLUX case I am referring to opm-tests tomorrow.

GitPaean:
I have open a PR to add the case mentioned above to opm-tests.  Maybe we should have the parallel restart tests with that one.

bska:
> I have open a PR to add the case mentioned above to opm-tests.

Thanks.  I'll update this PR to add that case for restart testing too.  In the meantime I'll reset this back to "draft" mode to prevent inadvertent merging.

bska:
jenkins build this please

GitPaean:

akva2:
if nothing is specified; 4.

GitPaean:

bska:
I consider this work complete now and I'm to merge this so I'm marking it "ready for review".

bska:
jenkins build this opm-tests=977 please

akva2:
Note https://github.com/OPM/opm-simulators/pull/4667 if you still want those timesteps reduced.

bska:
> Note #4667 if you still want those timesteps reduced.

I do, but I think I'll make that a follow-up to getting #4667 in later.

GitPaean:
The reference update OPM/opm-tests#977 is in,  I am getting this in now.

PR #4521, opened 2023-03-10
Tag Unused Parameter as Maybe_Unused

bska:
Noted by [at]akva2


bska:
jenkins build this please

akva2:
it's not maybe_unused, it's unused, so my personal preference would be to anonymize but i dunno what the policies on that are.

bska:
> my personal preference would be to anonymize [the parameter]

I prefer `[[maybe_unused]]` for the reason that it makes it easier to see what the value is in a debugger and, if we ever get there, to add Doxygen-style documentation for it.  The latter doesn't work if there's no parameter name.

bska:
I'll merge this now to remove the warning.

PR #4523, opened 2023-03-10
fix to broken built with clang

daavid00:
This commit fixes the current broken built with clang 
```
/Users/dmar/Github/opm/opm-simulators/opm/simulators/aquifers/BlackoilAquiferModel_impl.hpp:273:47: error: reference to local binding 'id' declared in enclosing function 'Opm::BlackoilAquiferModel<Opm::Properties::TTag::EclFlowProblemWaterOnlyEnergy>::createDynamicAquifers'
                return aquPtr->aquiferID() == id;
```

### DIFF
--- a/opm/simulators/aquifers/BlackoilAquiferModel_impl.hpp
+++ b/opm/simulators/aquifers/BlackoilAquiferModel_impl.hpp
@@ -268,7 +268,7 @@ void BlackoilAquiferModel<TypeTag>::createDynamicAquifers(const int episode_inde
         auto aquPos =
             std::find_if(std::begin(this->aquifers),
                          std::end(this->aquifers),
-                [id](const auto& aquPtr)
+                [id = id](const auto& aquPtr)
             {
                 return aquPtr->aquiferID() == id;
             });


opm/simulators/aquifers/BlackoilAquiferModel_impl.hpp
```diff
@@ -268,7 +268,7 @@ void BlackoilAquiferModel<TypeTag>::createDynamicAquifers(const int episode_inde
         auto aquPos =
             std::find_if(std::begin(this->aquifers),
                          std::end(this->aquifers),
-                [id](const auto& aquPtr)
+                [&id = id](const auto& aquPtr)
```
bska, line 271: Does
```C++
[id=id](const auto& aquPtr)
```
work instead?  I'd really prefer not to have to use a reference here.

daavid00:
I have updated the commit as suggested and it works, many thanks.

bska:
jenkins build this please

bska (approved):

PR #4524, opened 2023-03-13
allow for TEMP

goncalvesmachadoc:
Keyword TEMP is supported in OPM, it does exactly the same as THERMAL. Considering both THERMAL and TEMP do not follow  what is in the commercial simulator, don't see why to break the run if the user calls for TEMP.

### DIFF
--- a/opm/simulators/utils/UnsupportedFlowKeywords.cpp
+++ b/opm/simulators/utils/UnsupportedFlowKeywords.cpp
@@ -619,7 +619,6 @@ const KeywordValidation::UnsupportedKeywords& unsupportedKeywords()
         {"SURFWNUM", {true, std::nullopt}},
         {"SWF32D", {true, std::nullopt}},
         {"SWINGFAC", {true, std::nullopt}},
-        {"TEMP", {true, std::nullopt}},
         {"TEMPNODE", {true, std::nullopt}},
         {"TEMPTVD", {true, std::nullopt}},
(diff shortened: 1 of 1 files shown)


goncalvesmachadoc:
jenkins build this please

bska (approved):
Yes, this makes sense to me.  I'll merge into master.

OPMUSER:
@goncalvesmachadoc I don't understand this change, as the manual clearly states that TEMP is not supported and to activate  OPM Flow's thermal option one users the THERMAL keyword.

So now we allow the TEMP keyword, but what does it do? It does not activate the thermal option. If you just want the run to run instead of stopping, then there is a better way to do this by issuing a warning message saying the keyword is unsupported, rather than an error message.

@tskille would you care to comment on this as well.

Apologies for being late on the feedback, but I have been unavailable for a while and I'm just getting back into it.

goncalvesmachadoc:
If you use TEMP it does activate the thermal model, it's the same as using THERMAL. It was working ok before. If it's not doing that anymore, I can try to debug to see what is the problem.

OPMUSER:
@goncalvesmachadoc I disagree, TEMP has always been an unsupported keyword. I would like to hear from @tskille before we make any changes.

If do wish to make TEMP equal THERMAL then we could just make TEMP an alias for THERMAL, but I again I would like to see what @tskille thinks.

PR #4526, opened 2023-03-15
fix instatiation

ElyesAhmed:
Fixing Instantiation when having both Dune::ALUGrid and Dune::Fem


bska:
jenkins build this please

blattms:
You will have to check this manually with all supported DUNE versions.

ElyesAhmed:
> 

Ok!

atgeirr:
> You will have to check this manually with all supported DUNE versions.

I think as this is only affecting building with dune-fem, we should first clarify what versions of dune we should expect to be used with dune-fem. I still consider the dune-fem parts experimental, and requiring compatibility with the older dune versions is probably not worth it.

Dune 2.8 is fine with OPM using dune-fem (I know because that is what I use), but I do not know about other versions. I suggest requiring a minimum of 2.8 with dune-fem to simplify this.

hnil:
This should be not need with pullrequest: 4532. After discussion with we think this is a better solution.

PR #4530, opened 2023-03-15
Adjust Whitespace in Output Module

bska:
Mostly to split long lines and to add a few blank lines for readability.  While here, also apply 'const' in a few locations and flip the order of a conditional to (potentially) save a bit of work.

This is in preparation of revising the output destination categories of numerous 3D result vectors.

### DIFF
--- a/ebos/eclgenericoutputblackoilmodule.cc
+++ b/ebos/eclgenericoutputblackoilmodule.cc
@@ -56,26 +56,65 @@
 
 namespace {
 
-std::string EclString(Opm::Inplace::Phase phase) {
-    switch(phase) {
-    case Opm::Inplace::Phase::WATER: return "WIP";
-    case Opm::Inplace::Phase::OIL: return "OIP";
-    case Opm::Inplace::Phase::GAS: return "GIP";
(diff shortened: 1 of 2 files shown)


bska:
jenkins build this please

ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -943,14 +997,19 @@ doAllocBuffers(unsigned bufferSize,
         }
     }
 
+    if (auto& norst = rstKeywords["NORST"]; norst > 0) {
+        // Don't emit diagnostic messages about unsupported 'NORST' key.
+        norst = 0;
+    }
```
blattms, line 1003: This seem new and not whitespace. Intentional?

bska: > This seems new and not whitespace. Intentional?

It's intentional as I want to remove the misleading diagnostic
```
Keyword NORST is unsupported for output to file
```
but you're absolutely right that this is new code and not whitespace related.  I can split this off to a separate PR if you'd like.

blattms:
jenkins build this please

bska:
jenkins build this please

PR #4532, opened 2023-03-15
Refactoring to be able to use template class/methods even if they are…

hnil:
Refactoring to move template implementation to separate file from initiation. This make it easier to use the classes/functions for types not initiated in the cc file. It also may limit the number of initiations needed.
Current usecases is use of dune-fem with flow and code which avoid multiplexing in the fluid


bska:
jenkins build this please

blattms:
I must admit that I do not understand this reasoning. Of course having everything in a header files makes it easier, but it will increase compile time.

I also don't see why there should be less instantiations. I think it is the contrary the compiler will instantiated with the same types multiple times. Once for each *.cc file that includes it.

There has been a tremendous effort over the past to move as much stuff as possible into *.cc files to reduce compile time. If we change this back now, then there should be a very good reason for it.

blattms:
Of course I agree that it is cumbersome to do those instantiations needed. But as this is a limited set only it is doable.

akva2:
This is a 'best of both worlds' solution. It allows downstreams to instance for more types (in separate .cc files or however they want it) while still keeping the compile time advantages in this module. Notice the .hpp.files do not include the _impl.hpp files.

hnil:
- For the current code it do not change anything
- For other codes not in the main flow it do not need instantiation. This reduce the need of instantiation. But the other code may get larger compilation time, but also do not need to compile instantiations it do not need.

bska (approved):
I like it.  As others have said, this really achieves the "best of both worlds" goal of enabling easier experimentation with out-of-tree solvers/grid managers while retaining the separate explicit specialisations for the "known" combinations.  Unless anyone objects I intend to merge this into the master branch no later than noon CET tomorrow (Friday 17 March).

blattms:
thanks @akva2, that was not clear from the description. If hear *_impl.hpp that I usually think that it is included at the end of the header and don't carefully look at the code to find this one line is missing.

bska:
> Unless anyone objects I intend to merge this into the master branch no later than noon CET tomorrow (Friday 17 March).

A little delay, but I'll merge the PR now.

PR #4533, opened 2023-03-15
Split 3D Vectors into Base and Extended Categories

bska:
The `base` arrays are compatible in both name and interpretation to those emitted in the `SOLUTION` section from other simulators whereas the `extended` arrays are specific to Flow.  Tag extended arrays as `OPM_EXTENDED` instead of `AUXILIARY` as the latter is deprecated and will be removed at some point in the future.

While here, also re-tag the `TEMP` array as `OPM_EXTENDED` when not required for thermal simulations and add support for the per-phase fluid density `RPTRST` request keys (`DENx`, with `x` being `O`, `G`, or `W`) in addition to the collective key `DEN` that we already support.

### DIFF
--- a/ebos/eclgenericoutputblackoilmodule.cc
+++ b/ebos/eclgenericoutputblackoilmodule.cc
@@ -839,21 +839,39 @@ assignToSolution(data::Solution& sol)
     }
 
     if (eclState_.runspec().co2Storage() && !rsw_.empty()) {
-        std::vector<double> mfrac(rsw_.size(), 0.0);
-        const auto& pvtnum = eclState_.fieldProps().get_int("PVTNUM");
-        for (size_t i = 0; i < rsw_.size(); ++i) {
-            mfrac[i] = FluidSystem::convertXwGToxwG(FluidSystem::convertRswToXwG(rsw_[i], pvtnum[i]-1), pvtnum[i]-1);
-        }
-        sol.insert("XMFCO2", UnitSystem::measure::identity, std::move(mfrac), data::TargetType::RESTART_AUXILIARY);
+        auto mfrac = std::vector<double>(this->rsw_.size(), 0.0);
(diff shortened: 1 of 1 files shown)


bska:
Note as well, that while possibly not strictly needed, upstream PR OPM/opm-common#3447 was used for this work.

bska:
jenkins build this opm-common=3447 please

bska:
> Any reason for [not merging this PR now]?

Sadly, yes.  I get a regression failure in [compareParallelRestartedSim_flow+3D_2AQU_NUM](https://ci.opm-project.org/job/opm-simulators-PR-builder/4664/testReport/(root)/mpi/compareParallelRestartedSim_flow_3D_2AQU_NUM/) which I can't explain.  I need to analyse that failure before this is ready for merging.  Worst case we may have to disable that test case, but I would really prefer to understand what's happening here.

bska:
jenkins build this please

bska:
I'll rerun the build check just to reestablish the failure mode.  I'm unfortunately not able to reproduce that failure mode on my local machine.

bska:
jenkins build this opm-tests=931 please

totto82:
jenkins build this please

totto82:
From my point of view this PR should be ready after new reference cases are generated with the missing output. Thanks for the work.

bska:
> From my point of view this PR should be ready after new reference cases are generated with the missing output. Thanks for the work.

Cool.  I'll mark it as "ready for review" and **try** to generate a new set of reference solutions.

bska:

bska:
jenkins build this opm-tests=934 please

bska:
The new reference solutions have been installed on the CI system.  Merging this now, per prior approval, to bring the overall system back to a consistent state.

PR #4535, opened 2023-03-16
convert co2 output from volumes to moles

totto82:
Depends on https://github.com/OPM/opm-common/pull/3449

and https://github.com/OPM/opm-common/pull/3448


bska:
jenkins build this please

bska (approved):

PR #4539, opened 2023-03-20
Damaris vis data v3

jcbowden:
Hi OPM folks,

This is some work towards sending the OPM Flow mesh data through to Damaris. This will allow for 3D display of simulation results in Paraview and Visit. It has been tested as working using Ascent, which was accessible through the Damaris python interface.

It relies on a class that is packaged with Damaris in the include/util directory.

The mesh data is obtained from a derivative of the Dune VTKWriter.hh code in file GridDataOutput.hpp, class Opm::GridDataOutput::SimMeshDataAccessor.

This is a work in progress, as you might see.


jcbowden:
[damaris_config.xml.txt](https://github.com/OPM/opm-simulators/files/11023106/damaris_config.xml.txt)
[polygonal_mesh_conduit.py.txt](https://github.com/OPM/opm-simulators/files/11023107/polygonal_mesh_conduit.py.txt)

jcbowden:
damaris_ascent_opmflow_pressure_iter_56

I am not an Ascent expert, so getting the correct orientation is left for the user

bska:
I haven't looked at this in any kind of detail yet, but I notice that it appears to be using manual resource management through `new` and `delete`.  Does the Damaris software require this style?  If not, could we, at least at some point in the future, switch to using `std::unique_ptr<T[]>` or maybe even `vector<T>`, with `T` typically being `double`, instead?

blattms:
> The mesh data is obtained from a derivative of the Dune VTKWriter.hh code in file GridDataOutput.hpp, class Opm::GridDataOutput::SimMeshDataAccessor.

What does this mean? Are you deriving from class defined in VTKWriter.hh?

jcbowden:
> I haven't looked at this in any kind of detail yet, but I notice that it appears to be using manual resource management through `new` and `delete`. Does the Damaris software require this style? If not, could we, at least at some point in the future, switch to using `std::unique_ptr<T[]>` or maybe even `vector<T>`, with `T` typically being `double`, instead?

Damaris provides a raw pointer to shared memory, which is a block from a larger allocation, so probably cannot be managed by C++. I could be wrong with that though. Some of my use of new/delete is just from habit, and possibly not needed.

jcbowden:
Yes, derived from a class in VTKWriter.hh. The class also defines other internally defined helper classes that I also copy to get the required functionality.

bska:
> > I haven't looked at this in any kind of detail yet, but I notice that it appears to be using manual resource management through `new` and `delete`. Does the Damaris software require this style? If not, could we, at least at some point in the future, switch to using `std::unique_ptr<T[]>` or maybe even `vector<T>`, with `T` typically being `double`, instead?
> 
> Damaris provides a raw pointer to shared memory, which is a block from a larger allocation, so probably cannot be managed by C++. I could be wrong with that though. Some of my use of new/delete is just from habit, and possibly not needed.

Right.  But are you saying that Damaris participates in managing the application memory?  E.g., a memory block allocated by the application may be released by Damaris or the other way around?  If so, then we'll need to be a lot more careful than we currently are in the face of C++ exceptions.  If, on the other hand, the primary interaction between the application and Damaris is in terms of pointers to the start of application memory then I don't see why the application cannot use C++ techniques to manage its own memory resources.

jcbowden:
> Right. But are you saying that Damaris participates in managing the application memory? E.g., a memory block allocated by the application may be released by Damaris or the other way around? If so, then we'll need to be a lot more careful than we currently are in the face of C++ exceptions. If, on the other hand, the primary interaction between the application and Damaris is in terms of pointers to the start of application memory then I don't see why the application cannot use C++ techniques to manage its own memory resources.

No, application memory is managed on the application side, Damaris provides a 'data store' to copy the mesh data (or other data - pressure field data etc.) into. This is a shared memory area used for IPC.
Anyway, all C++ memory management options are there for use for memory allocated on the OPM flow side.

However, this is really a point of (lack of) optimisation of memory use, as currently, passing data through to Damaris requires a copy of the data. A better way would be to use the same Damaris API to allocate memory (perhaps as an allocator for a std::vector<>) and then the application can use the same memory as Damaris uses. It is a point for future development. Look to the end of the following page https://project.inria.fr/damaris/data-description/ to see the API I am speaking of. It is a bit confusing as the size of the requested memory is not explicit. This capability is not a priority, but something to keep in mind as maybe possible.

blattms:
> A better way would be to use the same Damaris API to allocate memory (perhaps as an allocator for a std::vector<>) and then the application can use the same memory as Damaris uses

This is a very strange idea. An output library forcing its users to use a custom allocator in their whole program. Sorry, but I doubt that this will fly in 99% of the cases. (last but not least because 80% aren't even correctly allocator aware to begin with)

jcbowden:
> This is a very strange idea. An output library forcing its users to use a custom allocator in their whole program. Sorry, but I doubt that this will fly in 99% of the cases. (last but not least because 80% aren't even correctly allocator aware to begin with)

Well, then the double up of memory for data being passed through to Damaris has to be accepted. I'm happy to look further into whether this is feasible. I would be trying to isolate the mesh data allocators and not replace all memory allocation. This is not a priority though, there is further work on the HDF5 output to get done for now.

bska:
> A better way would be to use the same Damaris API to allocate memory (perhaps as an allocator for a `std::vector<>`) and then the application can use the same memory as Damaris uses.

Right.  We're currently not allocator aware and making sufficiently large portions of the code base aware of traditional allocators is probably too much of an undertaking.  We could look into using C++17's [memory resources](https://en.cppreference.com/w/cpp/header/memory_resource) though, as those don't show up in the interfaces and vocabulary types.  Another option is to switch the relevant parts to using [`std::span<>`](https://en.cppreference.com/w/cpp/container/span) when we get C++20 or other library implementations such as [Boost.Range](https://www.boost.org/doc/libs/1_81_0/libs/range/doc/html/index.html).  That alternative depends on Damaris memory being contiguous for one variable for the cells on a given process and on being able to get a pointer/iterator to the start of the relevant Damaris memory block, as well as some way to demarcate the end of the range&mdash;either a number of elements or an actual end pointer/iterator.

The promise of not having to store the large solution arrays twice&mdash;or three times as it were, since we also effectively have copies inside the fluid state objects&mdash;is very enticing.

jcbowden:
-180 0_45 0_example_db_range_limited_3141933767
This version has Z height information that was missing in the last example image.

blattms:
Especially the last change (rewrite) is highly appreciated. Thanks a lot. Good that it makes the code simpler.

I have not looked at this in all detail. Hence just a few comments:

- Please add our standard copyright including license to all (new files). See the other already existing files.
- Not 100% sure whether this already is the case: All new headers need to be listed in CMakeLists_files.cmake on list PUBLIC_HEADER_FILES

jcbowden:
Okay, I think things are in order, thanks for indicating the few last tidy-up requirements.

atgeirr:
Hi, I am looking at this now, but noticed a ...-v4 version of this branch, but so far with no changes. Are there incoming things I should wait for? @jcbowden

jcbowden:
No, just -v3 is of interest. 

There are a set of additions to add command line arguments to flow, so as to support modifying the internal XML file, however that is part of damaris-improved-internal-hdf5. 

Did you want to review that? or should we leave it as a separate pull request?

atgeirr:
> Did you want to review that? or should we leave it as a separate pull request?

Separate is good, usually!

jcbowden:
@atgeirr You will probably have issues in Main.hpp. I sure did with a recent test compilation.

Also, there are some minor additions i'd like to add to eclwriter.hh - just the test for the return values of the mesh data writer functions. 

Can I push some changes to fix these issues?

jcbowden:
@atgeirr Currently the damaris-vis-data-v3 branch will need to be tested with damaris branch dask_pub_sub. I'll try to get the damaris changes available in master and release tag v1.8.3 sometime soon.

atgeirr:
> Can I push some changes to fix these issues?

Sure, I had to rebase and merge manually in Main.hpp (the master changed the line with ParsingStrictness recently, think that may be why).

jcbowden:
Yes, that recent merge made me mess up Main.hpp terribly. I started my changes from the current master Main.hpp. I'll push that in minute (maybe 5), I just want my test to run first.

jcbowden:
My test is failing, sorry, I might be another 1/2 hour or so.

jcbowden:
Sorry, I am making a moving target for you with these changes. I think everything is stable now.

PR #4540, opened 2023-03-21
output average pressure in all cells unconditionally with co2store

totto82:
An alternative is to change this to for all water + gas cases not only when CO2STORE is used.


totto82:
jenkins build this please

totto82:
jenkins build this please

bska (approved):

PR #4542, opened 2023-03-21
Optimisation of data processing for actionX and output

hnil:
* removed double evaluation of `processLocalData`
* separated fluid in place work
* removed double evaluation of `processLocalData` from other cell based data

### DIFF
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -175,6 +175,7 @@ public:
      */
     void evalSummaryState(bool isSubStep)
     {
+        OPM_TIMEBLOCK(evalSummaryState);
         const int reportStepNum = simulator_.episodeIndex() + 1;
         /*
           The summary data is not evaluated for timestep 0, that is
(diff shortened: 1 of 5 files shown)


bska:
jenkins build this please

ebos/eclwriter.hh
```diff
@@ -507,6 +515,7 @@ private:
     void prepareLocalCellData(const bool isSubStep,
                               const int  reportStepNum)
     {
+        if( !(eclOutputModule_->localDataValid()) ){
```
bska, line 518: You might alternatively write this as
```C++
if (eclOutputModule_->localDataValid()) {
    return;
}
```
which would save one level of nesting.

ebos/eclwriter.hh
```diff
@@ -470,7 +478,7 @@ public:
     void endRestart()
     {}
 
-    const EclOutputBlackOilModule<TypeTag>& eclOutputModule() const
+    EclOutputBlackOilModule<TypeTag>& eclOutputModule() const
```
bska, line 481: I think I'd prefer if this were left to return a reference to `const`, and then we'd add a new function, e.g.,
```C++
EclOutputBlackoilModule<TypeTag>& mutableEclOutputModule()
{ return *eclOutputModule_; }
```
instead.  At least in that way, it's clear at the call site that we're performing a potentially mutating operation on the output module object.  You'd be forced to write something like
```C++
ebosSimulator_.problem().eclWriter()->mutableEclOutputModule().invalidateLocalData();
```
instead.

bska (commented):
Apart from a couple of minor issues I like this a lot and I really look forward to seeing the benchmark results later.  I think we'll nevertheless merge this more or less as is.

ebos/eclwriter.hh
hnil: changed it.

ebos/eclwriter.hh
hnil: changed it.

hnil:
Test on norne. original code 60 s on post and pre + 20 on output. With pullrequest 40 + 1.
På SPE9_CP original code 1.56 + 0.7 til 1.5 + 0.14.

bska (approved):
Thanks a lot for the updates.  This looks good to me now.  I'll just run one more build check and then I'll merge this once the check is clear.

bska:
> I'll just run one more build check and then I'll merge this once the check is clear.

Build is successful.  I'll merge this into the master branch now.  Thanks a lot for improving the performance here!

PR #4545, opened 2023-03-23
Faster fip calculation

hnil:
-- separated out put in different functions
-- removed element context from the update fib loop.

### DIFF
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -541,13 +541,31 @@ private:
             eclOutputModule_->processElement(elemCtx);
         }
         }
+        if(!simulator_.model().linearizer().getFlowsInfo().empty()){
+            OPM_TIMEBLOCK(prepareFlowsData);     
+            for (const auto& elem : elements(gridView)) {
+                elemCtx.updatePrimaryStencil(elem);
+                elemCtx.updatePrimaryIntensiveQuantities(/*timeIdx=*/0);
+                eclOutputModule_->processElementFlows(elemCtx);
+            }
+        }
         {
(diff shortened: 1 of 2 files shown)


bska:
jenkins build this please

ebos/eclwriter.hh
```diff
@@ -541,13 +541,31 @@ private:
...
-            eclOutputModule_->updateFluidInPlace(elemCtx);
+            eclOutputModule_->processElementBlockData(elemCtx);
+        }
+        }
+        {
+        OPM_TIMEBLOCK(prepareFluidInPlace);
+#ifdef _OPENMP
+#pragma omp parallel for
+#endif
+        for (size_t dofIdx=0; dofIdx < numElements; ++dofIdx){
```
bska, line 565: For what it's worth, `numElements` is [declared](https://github.com/OPM/opm-simulators/blob/9450fc05962edc4679bf620ea5e5d1a0334b2ee1/ebos/eclwriter.hh#L527) earlier in the function as `const int`.  Using a `dofIdx` of type `size_t` means we [get a warning](https://ci.opm-project.org/job/opm-simulators-PR-builder/4691/gcc/new/) when comparing the unsigned `dofIdx` with the signed `numElments`.

Maybe we can write the loop condition as
```C++
for (size_t dofIdx=0; dofIdx < static_cast<size_t>(numElements); ++dofIdx) {
```
instead in order to suppress the "signed-vs-unsigned" warning?

bska (commented):
Overall I like this as it goes some way towards separating the concerns and providing timing measurements for the individual parts.  I'm in favour of merging this in more or less its current state.

bska:
jenkins build this please

PR #4546, opened 2023-03-23
Revert "Treat rate control with zero target same as stopped well."

GitPaean:
Reverts OPM/opm-simulators#4518


bska:
> Reverts #4518

Okay, but if we want to merge this we should probably revert OPM/opm-tests#918 too.

bska:
> There is no intention to merge this PR.

Noted.

> But technically for the PR in opm-tests, should not we just update it again with new PR instead of reverting the merged PR?

Yes, I think we'll pretty much be forced to generate new reference solutions instead.  If OPM/opm-tests#918 were the latest update to OPM-Tests, then reverting would probably work.  As it happens though, there have been a couple of updates since then which makes reverting the unformatted result files rather difficult.

PR #4547, opened 2023-03-23
using double for regularization_factor_wells_

GitPaean:
it should be a type double. With int, it does not looks like it can get proper value at all.


GitPaean:
jenkins build this please

bska:
Hm, this looks to have been an `int` since it was first introduced as `regularization_factor_ms_wells_` in commit 8a5c9e82d (PR #2611).  I guess no-one picked up on that during review.  Still, let's ask @totto82 about the intention here.

bska:
> it looks like [`regularization_factor_wells_`] should be a `double` value here.

Yes I agree.  If there's a good reason to make it be `int` then we can reset the type later.  For now I'll merge this into the master branch.

PR #4549, opened 2023-03-24
fix RESV for undersaturated wells

totto82:
Issue reported by @tskille 

A similar fix may be needed for groups controlled by RESV, but lets fix the wells first.

### DIFF
--- a/opm/simulators/wells/WellAssemble.cpp
+++ b/opm/simulators/wells/WellAssemble.cpp
@@ -129,7 +129,7 @@ assembleControlEqProd(const WellState& well_state,
         auto total_rate = rates[0]; // To get the correct type only.
         total_rate = 0.0;
         std::vector<double> convert_coeff(well_.numPhases(), 1.0);
-        well_.rateConverter().calcCoeff(/*fipreg*/ 0, well_.pvtRegionIdx(), convert_coeff);
+        well_.rateConverter().calcCoeff(/*fipreg*/ 0, well_.pvtRegionIdx(), well_state.well(well_.indexOfWell()).surface_rates, convert_coeff);
         for (int phase = 0; phase < 3; ++phase) {
             if (pu.phase_used[phase]) {
                 const int pos = pu.phase_pos[phase];
--- a/opm/simulators/wells/WellAssemble.cpp
+++ b/opm/simulators/wells/WellAssemble.cpp
@@ -179,9 +179,13 @@ assembleControlEqProd(const WellState& well_state,
                 active_rates[pu.phase_pos[canonical_phase]] = rates[canonical_phase];
             }
         }
-        auto rCoeff = [this](const RegionId id, const int region, std::vector<double>& coeff)
+        auto rCoeff = [this, &group_state](const RegionId id, const int region, const std::optional<std::string>& prod_gname, std::vector<double>& coeff)
         {
-            well_.rateConverter().calcCoeff(id, region, coeff);
+            if (prod_gname)
+                well_.rateConverter().calcCoeff(id, region, group_state.production_rates(*prod_gname), coeff);
+            else
+                well_.rateConverter().calcCoeff(id, region, coeff);
+
         };
         WellGroupControls(well_).getGroupProductionControl(group, well_state,
                                                              group_state,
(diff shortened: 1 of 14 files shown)


totto82:
jenkins build this please

opm/simulators/wells/WellAssemble.cpp
```diff
@@ -179,9 +179,13 @@ assembleControlEqProd(const WellState& well_state,
                 active_rates[pu.phase_pos[canonical_phase]] = rates[canonical_phase];
             }
         }
-        auto rCoeff = [this](const RegionId id, const int region, std::vector<double>& coeff)
+        auto rCoeff = [this, group_state](const int id, const int region, const std::optional<std::string>& prod_gname, std::vector<double>& coeff)
```
bska, line 182: Just to be clear: This will create a copy of the `group_state` inside the lambda.  Is that what you want?  If not, the capture should be `&group_state` instead.  Also, is there a reason for switching from `RegionId` to `int` here?  That potentially introduces needless type conversions.

totto82: Thanks, I will update the code. @tskille has tested the code and confirms that is indeed fixes the unexpected results for RESV controlled wells for under saturated oil. i.e. this should be merged when the code is fixed.

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
This commit build fine locally, so I am bit puzzled why Jenkins complains. It seems to complain about missing includes. I will try again with I am a bit lost since I dont understand the issue.

totto82:
jenkins build this please

totto82:
jenkins build this please

bska:
> This commit build fine locally, so I am bit puzzled why Jenkins complains.

If you're using GCC, then it's probably a different version of the standard library files.  

> It seems to complain about missing includes. I will try again with I am a bit lost since I don't understand the issue.

The [diagnostic message](https://ci.opm-project.org/job/opm-simulators-PR-builder/4720/consoleFull) is quite clear
> ```
> In file included from /var/lib/jenkins/workspace/opm-simulators-PR-builder@2/opm/simulators/wells/WellGroupControls.cpp:23:
> /var/lib/jenkins/workspace/opm-simulators-PR-builder@2/opm/simulators/wells/WellGroupControls.hpp:50:97: error: ‘string’ is not a member of ‘std’
 >   50 |     using RateConvFunc = std::function<void(const RegionId, const int, const std::optional<std::string>&, std::vector<double>&)>;
 >      |                                                                                                 ^~~~~~
> /var/lib/jenkins/workspace/opm-simulators-PR-builder@2/opm/simulators/wells/WellGroupControls.hpp:28:1: note: ‘std::string’ is defined in header ‘<string>’; did you forget to ‘#include <string>’?
>   27 | #include <functional>
>  +++ |+#include <string>
>   28 | #include <optional>
> ```

This PR [adds](https://github.com/OPM/opm-simulators/blob/0eb4410cdc07de43f47cf3e2f8ad9ee82fd97c1b/opm/simulators/wells/WellGroupControls.hpp#L51) a new parameter of type `optional<string>` to the definition of `RateConvFunc` in `WellGroupControls.hpp`, but does not include the `<string>` header in that file.  We can't have an `optional` of an incomplete type.

totto82:
jenkins build this please

totto82:
>I am a bit lost since I dont understand the issue.

Thanks @bska for the explanation. I didn't notice that the error message pointed to different files. i.e. WellGroupControls.hpp and WellGroupConstraints.hpp so when I thought I had fixed the problem it still complained. 

I am using "gcc version 9.4.0"  so lagging behind a bit.

totto82:
I have manually checked the failures. All expect SPE9_CP_GROUP_RESV are minor. 
Here "reference" is the results in opm-tests/spe9group/opm-simulation-reference

image

From 11 Apr 2015 to 20 Jly 2015 it is controlled by RESV = 1000. 
 
GCONPROD
--      oil   water  gas
   'F'    'RESV'  4* NONE NO 5* 1000 /
   'A'    'FLD'   4* 'RATE' 'YES' 1 'OIL' /
/

The PR is also tested by @tskille 

Since this is a bug fix the PR should be merged as soon as the code quality is sufficiently good.

bska (commented):
Yeah, this looks good to me and it does correct a programming error.  I'll just ask if you can fix the [diagnostic message](https://github.com/OPM/opm-simulators/blob/a73f4078ef61185cfca6320ad8790203cd161496/opm/simulators/wells/RateConverter.cpp#L197-L200) concerning detecting gas-in-oil and gas-in-water present at the same time.  Right now, both of the `std::string()` calls pass `detR` and I suspect that one of those should be `detRw` instead.  It's also not entirely clear to me what
```
only support <value> <value>
```
is supposed to mean to an end user who does not have the source code so maybe we can expand a little bit on that too?

totto82:
jenkins build this please

totto82:

bska:
jenkins build this opm-tests=929 please

PR #4551, opened 2023-03-24
output gas in water to F(R)GIPL for gas-water systems

totto82:
Only relevant for cases with that allows for dissolved gas in water

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -1150,6 +1150,10 @@ private:
                 if (!this->fip_[Inplace::Phase::WaterInWaterPhase].empty())
                     this->fip_[Inplace::Phase::WaterInWaterPhase][globalDofIdx] = fip[waterPhaseIdx];
 
+                // For water+gas cases the gas in water is added to the GIPL value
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1150,6 +1150,9 @@ private:
                 if (!this->fip_[Inplace::Phase::WaterInWaterPhase].empty())
                     this->fip_[Inplace::Phase::WaterInWaterPhase][globalDofIdx] = fip[waterPhaseIdx];
 
+                if (!this->fip_[Inplace::Phase::GasInLiquidPhase].empty())
+                    this->fip_[Inplace::Phase::GasInLiquidPhase][globalDofIdx] = gasInPlaceWater;
```
bska, line 1154: I think this may end up overwriting the value that's [already assigned](https://github.com/OPM/opm-simulators/blob/71ee88337e10a7f29205598a7454343077f7e460/ebos/ecloutputblackoilmodule.hh#L1130-L1131) in the O/G branch above, at least for the common case of a three-phase O/G/W system.

If this assignment is supposed to cover the gas/water case only, then we may need to additionally require
```C++
! FluidSystem::phaseIsActive(oilPhaseIdx)
```
here

bska:
jenkins build this please

bska (approved):

PR #4553, opened 2023-03-24
removing unused ToleranceWellControl command line argument

GitPaean:
found it not used accidentally. We might want to have something similar while different in place. But before then, we can remove this one since it is not used.


GitPaean:
The following one is not used either
```c++
        /// Tolerance for time step in seconds where single precision can be used
        /// for solving for the Jacobian
        double maxSinglePrecisionTimeStep_;
```

Maybe @hnil knows about it?

GitPaean:
okay. it is used by ebos.. closing this PR now.

GitPaean:
but at the same time, it will show with `flow --help`. It might be a little misleading since it will not affect the result for `flow`. Maybe better message/document should be in place.

PR #4557, opened 2023-03-27
Remove Unused Variables

bska:
While here, also include requisite headers.


bska:
jenkins build this please

akva2 (approved):

PR #4559, opened 2023-03-27
compute rock fraction

totto82:
See https://github.com/OPM/opm-models/pull/795


bska:
jenkins build this please

bska (approved):
Yep, this makes sense to me.  I'll merge into master.

PR #4560, opened 2023-03-28
- changed pointer to refernace for IQ for possibility of using rvalues

hnil:
small change to be able to return local variable for IQ


bska:
jenkins build this please

PR #4561, opened 2023-03-28
Minimal output handeling

hnil:
- steps to minimize output handeling on full vectors.

### DIFF
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -533,7 +533,7 @@ private:
         ElementContext elemCtx(simulator_);
         OPM_BEGIN_PARALLEL_TRY_CATCH();
         if(!isSubStep){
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska (approved):
This looks good to me.  Unless there's more work coming here I suggest we merge this as is.

blattms:
maybe fix indentation?

hnil:
Closing — both guards are on master already, in better form. `processElementBlockData` returns early when `blockExtractors_`, `extraBlockExtractors_` and `lgrBlockExtractors_` are all empty, and the extra block data is set up under `if (reportStepNum > 0 && !isSubStep)`. The OutputExtractor rework got there by another route. Thanks for the approval back in 2023 - it just took a while to become unnecessary.

PR #4562, opened 2023-03-28
add THERMAL boundary conditions which keeps constant temperature on the boundary

totto82:
See https://github.com/OPM/opm-common/pull/3457


totto82:
jenkins build this please

PR #4565, opened 2023-03-28
Reduced linearsolver option

hnil:
- added possibility to use reduced tolerance for linear solver

### DIFF
--- a/opm/simulators/linalg/ISTLSolverEbos.hpp
+++ b/opm/simulators/linalg/ISTLSolverEbos.hpp
@@ -434,13 +434,15 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
             // store number of iterations
             iterations_ = result.iterations;
             converged_ = result.converged;
-            if(result.reduction < parameters_.relaxed_linear_solver_reduction_){
-                std::stringstream ss;
-                ss<< "Full linear solver tolerance not achived. The reduction is:" << result.reduction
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

opm/simulators/linalg/ISTLSolverEbos.hpp
```diff
@@ -435,6 +435,10 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
             iterations_ = result.iterations;
             converged_ = result.converged;
             if(result.reduction < parameters_.relaxed_linear_solver_reduction_){
+                std::stringstream ss;
+                ss<< "Full linear solver tolerance not achived. The reduction is:" << result.reduction
+                  << " after " << result.iterations << " iteraions ";
```
akva2, line 440: iterations

hnil: The intention is just to accept a linear solve which after max iterations did not converged, but the solution is not that bad. This make it possible to set the max iteration quite low without triggering time step cuts due to none convergence and at the same time not continue if linear solves is clearly going in the wrong direction.

akva2: sure i got that. but we still don't want the typo ;)

hnil: done

akva2:
jenkins build this please

PR #4566, opened 2023-03-29
- changed some asserts to throw

hnil:
- changing asserts which may happen due to input to throw.

### DIFF
--- a/opm/simulators/flow/BlackoilModelEbos.hpp
+++ b/opm/simulators/flow/BlackoilModelEbos.hpp
@@ -581,15 +581,15 @@ namespace Opm {
                         Scalar tmpSat = saturationsNew[phaseIdx] - saturationsOld[phaseIdx];
                         resultDelta += tmpSat*tmpSat;
                         resultDenom += saturationsNew[phaseIdx]*saturationsNew[phaseIdx];
-                        assert(std::isfinite(resultDelta));
-                        assert(std::isfinite(resultDenom));
                     }
                 }
             }
 
             resultDelta = gridView.comm().sum(resultDelta);
             resultDenom = gridView.comm().sum(resultDenom);
-
+            if (!(std::isfinite(resultDelta) && std::isfinite(resultDenom))) {
+                OPM_THROW(std::runtime_error,"Not finite variable change");
+            }
             if (resultDenom > 0.0)
                 return resultDelta/resultDenom;
             return 0.0;
--- a/opm/simulators/timestepping/AdaptiveSimulatorTimer.cpp
+++ b/opm/simulators/timestepping/AdaptiveSimulatorTimer.cpp
@@ -24,10 +24,11 @@
 #include <cassert>
 #include <iostream>
 #include <vector>
-
+#include <stdexcept>
 #include <algorithm>
 #include <numeric>
-
+#include <opm/common/Exceptions.hpp>
+#include <opm/common/ErrorMacros.hpp>
 #include <opm/input/eclipse/Units/Units.hpp>
 #include <opm/simulators/timestepping/AdaptiveSimulatorTimer.hpp>
 
@@ -66,7 +67,10 @@ AdaptiveSimulatorTimer& AdaptiveSimulatorTimer::operator++ ()
     {
         ++current_step_;
         current_time_ += dt_;
-        assert(dt_ > 0);
+        if(!(dt_ > 0)){
+            OPM_THROW(std::runtime_error,
+                      "TimeStep is zero");
+        }
         // store used time step sizes
         steps_.push_back( dt_ );
         return *this;
@@ -78,7 +82,10 @@ AdaptiveSimulatorTimer& AdaptiveSimulatorTimer::operator++ ()
         double remaining = (total_time_ - current_time_);
         // apply max time step if it was set
         dt_ = std::min( dt_estimate, max_time_step_ );
-        assert(dt_ > 0);
+        if(!(dt_ > 0)){
+            OPM_THROW(std::runtime_error,
+                      "TimeStep is zero");
+        }        
         if( remaining > 0 ) {
 
             // set new time step (depending on remaining time)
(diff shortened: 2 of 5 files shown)


opm/simulators/flow/BlackoilModelEbos.hpp
```diff
@@ -581,15 +581,15 @@ namespace Opm {
...
-                        assert(std::isfinite(resultDelta));
-                        assert(std::isfinite(resultDenom));
                     }
                 }
             }
 
             resultDelta = gridView.comm().sum(resultDelta);
             resultDenom = gridView.comm().sum(resultDenom);
-
+            if (!(std::isfinite(resultDelta) && std::isfinite(resultDenom))) {
```
akva2, line 590: ! || ! is more readable imo.

opm/simulators/timestepping/AdaptiveSimulatorTimer.cpp
```diff
@@ -88,7 +95,10 @@ AdaptiveSimulatorTimer& AdaptiveSimulatorTimer::operator++ ()
                 if( dt_ > max_time_step_ ) {
                     dt_ = 0.5 * remaining;
                 }
-                assert(dt_ > 0);
+                if(!(dt_ > 0)){
```
akva2, line 98: <= 0 is more readable.

opm/simulators/timestepping/AdaptiveSimulatorTimer.cpp
```diff
@@ -97,7 +107,10 @@ AdaptiveSimulatorTimer& AdaptiveSimulatorTimer::operator++ ()
 
             if( 1.5 * dt_ > remaining ) {
                 dt_ = 0.5 * remaining;
-                assert(dt_ > 0);
+                if(!(dt_ > 0)){
```
akva2, line 110: <= 0 is more readable.

opm/simulators/timestepping/AdaptiveSimulatorTimer.cpp
```diff
@@ -111,7 +124,10 @@ AdaptiveSimulatorTimer& AdaptiveSimulatorTimer::operator++ ()
 
     double AdaptiveSimulatorTimer::currentStepLength () const
     {
-      assert(dt_ > 0);
+        if(!(dt_ > 0)){
```
akva2, line 127: <= 0 is more readable

opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
```diff
@@ -522,10 +522,14 @@ std::set<std::string> consistentlyFailingWells(const std::vector<StepReport>& sr
                     double dtEstimate = timeStepControl_->computeTimeStepSize(dt, iterations, relativeChange,
                                                                                substepTimer.simulationTimeElapsed());
 
-                    assert(dtEstimate > 0);
+                    if(!(dtEstimate > 0)){
```
akva2, line 525: <= 0 is more readable.

opm/simulators/timestepping/AdaptiveTimeSteppingEbos.hpp
```diff
@@ -522,10 +522,14 @@ std::set<std::string> consistentlyFailingWells(const std::vector<StepReport>& sr
...
+                        OPM_THROW(std::runtime_error,"Timestep estimate have become zero");
+                    }
                     // limit the growth of the timestep size by the growth factor
                     dtEstimate = std::min(dtEstimate, double(maxGrowth_ * dt));
-                    assert(dtEstimate > 0);
+                    if(!(dtEstimate > 0)){
```
akva2, line 530: <= 0 is more readable.

akva2 (commented):
Looks fine but I would prefer more explicit conditionals.

blattms:
Please, pretty please, with lots of sugar on top, use meaningful error messages that help people to see where the problem is. Those used here are 100% correct, but also 100% unable to tell the user in which part of the simulator the problem is.

opm/simulators/timestepping/AdaptiveSimulatorTimer.cpp
```diff
@@ -66,7 +67,10 @@ AdaptiveSimulatorTimer& AdaptiveSimulatorTimer::operator++ ()
     {
         ++current_step_;
         current_time_ += dt_;
-        assert(dt_ > 0);
+        if(!(dt_ > 0)){
```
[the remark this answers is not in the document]
bska: > <= 0 is more readable.

Arguably yes, but it's not completely equivalent.  If, through some unexpected calculation, `dt_` happens to be `NaN`, then `<= 0` is false while `! (dt_ > 0)` is true.  There's also the general advice of avoiding equality comparisons for floating-point numbers if possible.

hnil:
Closing: the patch targets BlackoilModelEbos.hpp and AdaptiveTimeSteppingEbos.hpp, which no longer exist after the Ebos rename.

The underlying issue is still live, so recording it rather than losing it: there are currently 140 assert( calls in opm/simulators/wells/*.cpp. Asserts compile out under NDEBUG, so in a release build those are not checks at all. This is worth revisiting as a fresh change against current master, particularly if opm-common #3427 (WITH_NDEBUG ON by default in release) is ever taken.

hnil:
Reopening as a draft to keep the issue alive.

The patch itself is dead -- BlackoilModelEbos.hpp and AdaptiveTimeSteppingEbos.hpp no longer exist -- but the problem is not: there are currently 140 assert( calls in opm/simulators/wells/*.cpp, and asserts compile out under NDEBUG, so in a release build they are not checks at all. Related to opm-common #3427, which would turn NDEBUG on for release builds; if that lands, converting the load-bearing asserts becomes more urgent.

hnil:
Superseded by #7355, which reapplies this to today's files - this branch predates the ebos rename and no longer merges. The asserts it targeted are all still on master, so the point stood; it just needed re-siting.

Issue #4548, opened 2023-03-23
Solvent model in 2 phase run does not run


EdmundStephens:
[2022.10] The attached run is modelling CCS in a depleted gas using the black oil SOLVENT method. (It is a 41 x 1 x 50 layer 2D secton based on an outcrop we are working with). I was able to test it in Eclipse and runs nicely albeit with tight tolerances and 1 day timesteps. But Flow immediately stops. Can you check what exactly is going wrong?

[Covesea1C_DG_CCS.zip](https://github.com/OPM/opm-simulators/files/11053692/Covesea1C_DG_CCS.zip)
[Covesea1C_DG_CCS-F.zip](https://github.com/OPM/opm-simulators/files/11053696/Covesea1C_DG_CCS-F.zip)

Thank you!

totto82:
Thanks for reporting the issue. The files you shared does not contain the include files which makes it hard to debug locally. Are you able to reproduce the issue on a model you can share with us?

EdmundStephens:
Also for your consideration, the same model with a dummy oil phase .... this again runs in Eclipse but not in Flow. 

[Covesea1C_DG_CCS3 (flat).zip](https://github.com/OPM/opm-simulators/files/11089408/Covesea1C_DG_CCS3.flat.zip)
[Covesea1C_DG_CCS3-F (flat).zip](https://github.com/OPM/opm-simulators/files/11089409/Covesea1C_DG_CCS3-F.flat.zip)

The output error:
```
===============Saturation Functions Diagnostics===============

System:  Solvent model.
Relative permeability input format: Saturation Family II.
Number of saturation regions: 5

Program threw an exception: TableContainer - invalid tableNumber

```

Quite likely here I have got one of the table sets wrongly enumerated, as there are 5 facies (NTSFUNC) ... but I cannot see what. Eclipse can often be more lenient with defaults etc. which in my opinion can be quite dangerous in making wrong input apparently work!

totto82:
Thanks for the cases. Flow does not support gas+water+solvent at the moment, but it has been on my TODO list for some time.  In principle it shouldn't be too hard to add, but will require some debugging to find places where 3p is hardcoded. I will report back when I have a working version. I haven't looked at the dummy-oil version yet. Maybe you could try to run a case with only 1 region first?

PR #4568, opened 2023-03-29
Add gaswater solvent

totto82:
Fix some places that assumed 3p + solvent

See https://github.com/OPM/opm-simulators/issues/4548

Depends on https://github.com/OPM/opm-models/pull/796

Compiles but does not converge.


totto82:
Gives same results as a test case with trivial oil phase. i.e. this is ready for review.

totto82:
jenkins build this please

opm/simulators/wells/SingleWellState.cpp
```diff
@@ -161,6 +161,7 @@ double SingleWellState::sum_polymer_rates() const {
 
 
 double SingleWellState::sum_solvent_rates() const {
+    //return this->surface_rates[pu.phase_pos[BlackoilPhases::Vapour]];
```
GitPaean, line 164: please remove this commented line.

opm/simulators/wells/StandardWellPrimaryVariables.cpp
```diff
@@ -455,6 +461,9 @@ volumeFraction(const unsigned compIdx) const
         if (has_gfrac_variable && compIdx == Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx)) {
             return evaluation_[GFrac];
         }
+        if (Indices::enableSolvent && compIdx == (unsigned)Indices::contiSolventEqIdx) {
+            return evaluation_[SFrac];
+        }
```
GitPaean, line 466: can we reduce the duplication of the following code? 

```c++
        if (has_gfrac_variable && compIdx == Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx)) {
            return evaluation_[GFrac];
        }
        if (Indices::enableSolvent && compIdx == (unsigned)Indices::contiSolventEqIdx) {
            return evaluation_[SFrac];
        }
```

opm/simulators/wells/StandardWellPrimaryVariables.cpp
```diff
@@ -475,6 +484,10 @@ volumeFraction(const unsigned compIdx) const
     else if (FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx) &&
              FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx)) {
         well_fraction -= evaluation_[GFrac];
+
+        if constexpr (Indices::enableSolvent) {
```
GitPaean, line 488: I think we also only need one the following code? 

```c++
        if constexpr (Indices::enableSolvent) {
            well_fraction -= evaluation_[SFrac];
        }
```

Maybe it might cause problem for single phase solvent case?

GitPaean (changes requested):
The code looks fine, but in general with all the phase combinations, the code in general (not only in this PR) is very hard to follow and easy to make mistakes now.  Maybe we should clear up the logic so it can be clearer to follow.

totto82:
Thanks for the review. I have followed your suggestions.

totto82:
jenkins build this please

totto82:
I plan to add the test case in  https://github.com/OPM/opm-tests/pull/973 when it gets merged.

opm/simulators/wells/StandardWellPrimaryVariables.cpp
```diff
@@ -438,45 +444,31 @@ volumeFraction(const unsigned compIdx) const
...
-        }
-
-        if (FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx)) {
-            well_fraction -= evaluation_[GFrac];
-        }
-
-        if constexpr (Indices::enableSolvent) {
-            well_fraction -= evaluation_[SFrac];
-        }
+    if (has_wfrac_variable && FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx)) {
```
GitPaean, line 463: can you modify the comment `// Oil or WATER fraction` to be more specific. 

Basically, it will be oil fraction when oil phase is active and water fraction when oil phase is not active.

GitPaean:
Thanks for the update. The PR looks fine to me now except the small request regarding comment (some comment is needed to make the code more maintainable). 

Then we should wait the testing case going in the opm-tests and create the regression test to protect the development.

totto82:
jenkins build this please

GitPaean:
https://github.com/OPM/opm-tests/pull/973 just got merged. Please add the regressing test, register the reference update, so we can update the reference and get this PR in.

GitPaean (approved):

totto82:
jenkins build this opm-tests=979 please

PR #4570, opened 2023-03-29
-- changed function to protected

hnil:
Models functions to protected. This make it possible to avoid global update of IQ with property changes, which is need for some models.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -2078,7 +2078,7 @@ public:
 private:
     Implementation& asImp_()
     { return *static_cast<Implementation *>(this); }
-
+protected:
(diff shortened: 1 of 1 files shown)


bska:
Just to be clear: Did you intend to make all of these functions `protected`?  This PR affects a number of member functions named `updateSomething_()` and a lot of other functions too.  It's probably fine to have these as `protected`, but I wanted to be sure that it was at least intentional.

hnil:
I thought I put private at the correct place. The important part is to have all of the local update function protected.

bska:
> The important part is to have all of the local update function protected.

Okay.  If we need to restrict the visibility of some of these functions we can do so later.  I'll merge into master.

PR #4571, opened 2023-03-30
Increase alq to its minimum value while checking that is does not violate its rate limits

totto82:
The old code stopped increasing if the well didn't give a valid rate. With this change we will continue to increase and hope that the well start flowing for larger alq values. This also fixes a assert trigged later due to alq < min_alq.


hakonhagland:
Thanks, this looks good to me.

bska:
> Thanks, this looks good to me.

Very good.  In that case I'll merge into the master branch

PR #4573, opened 2023-03-31
Fix bug when recalculating gaslift gradient

hakonhagland:
Fixes a bug in `recalculateGradientAndUpdateData_()` in gaslift stage 2. Also adds some minor improvement to the formatting of gaslift debugging output.

### DIFF
--- a/opm/simulators/wells/GasLiftSingleWellGeneric.cpp
+++ b/opm/simulators/wells/GasLiftSingleWellGeneric.cpp
@@ -1163,8 +1163,8 @@ GasLiftSingleWellGeneric::runOptimizeLoop_(bool increase)
     // if (this->debug) debugShowBhpAlqTable_();
     if (this->debug)
         debugShowAlqIncreaseDecreaseCounts_();
-    //if (this->debug)
-    //    debugShowTargets_();
+    if (this->debug)
+        debugShowTargets_();
     bool success = false; // did we succeed to increase alq?
     bool alq_is_limited = false;
     LimitedRates new_rates = *rates;
--- a/opm/simulators/wells/GasLiftSingleWellGeneric.cpp
+++ b/opm/simulators/wells/GasLiftSingleWellGeneric.cpp
@@ -85,8 +85,14 @@ GasLiftSingleWellGeneric::GasLiftSingleWellGeneric(DeferredLogger& deferred_logg
  ****************************************/
 // NOTE: Used from GasLiftStage2
 std::optional<GasLiftSingleWellGeneric::GradInfo>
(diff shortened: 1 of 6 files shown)


hakonhagland:
jenkins build this please

opm/simulators/wells/GasLiftSingleWellGeneric.cpp
```diff
@@ -1161,8 +1163,8 @@ GasLiftSingleWellGeneric::runOptimizeLoop_(bool increase)
     // if (this->debug) debugShowBhpAlqTable_();
     if (this->debug)
         debugShowAlqIncreaseDecreaseCounts_();
-    if (this->debug)
-        debugShowTargets_();
+    //if (this->debug)
+    //    debugShowTargets_();
     bool success = false; // did we succeed to increase alq?
```
totto82, line 1168: why is this commented out.

opm/simulators/wells/GasLiftStage2.cpp
```diff
@@ -457,26 +462,46 @@ recalculateGradientAndUpdateData_(
...
+            //   or the old decremental gradient becomes the new incremental gradient
+            // NOTE: The gradient itself (old_grad.grad) will be equal to the new decremental
+            //   gradient, but the other fields in old_grad cannot be used as they refer to
+            //   the incremental gradient. E.g. old_grad.alq is the alq after an increase in alq,
+            //   not a decrease, so we need to recalculate the decremental values here..
+            auto grad = calcIncOrDecGrad_(name, gs_well, gr_name_dont_limit, !increase);
+            if (grad) {
+                updateGrad_(name, *grad, !increase);
+                //updateGrad_(name, *old_grad, !increase);
+                // NOTE: This may invalidate any iterator into 'other_grads' since
```
totto82, line 490: Any reason to keep the old code here?

totto82 (approved):
Thanks for the fix. The code changes looks ok. I only have some cosmetic comments.

opm/simulators/wells/GasLiftSingleWellGeneric.cpp
hakonhagland: I think I commented it out because the amount of output it generated and almost the same information was output for each iteration of the well iterations. We can have it uncommented by default? Then it is easier to just comment it out (like I did) if the output is not interesting for the current debugging session.

opm/simulators/wells/GasLiftStage2.cpp
hakonhagland: No, no reason :)

hakonhagland:
jenkins build this please

blattms:
Comments are addressed. Merging.

PR #4574, opened 2023-03-31
add gas + water + energy to flow

totto82:


totto82:
jenkins build this please

GitPaean (approved):
The PR  looks fine to me and it is doing what is intended.  Jenkins approves. We get in now. 

At the same time, I did not work on `Thermal` or `Energy` related code much and I have to admit the combinations in `dispatchDynamic_` has some room to improve when coming to better readability.

PR #4576, opened 2023-03-31
Calculate Local Face Area Fraction for Aquifer's Connections

bska:
This is in preparation of reinitialising the total produced volume from the constant flux aquifer from the restart file in the case that the aquifer's connections are split across multiple MPI ranks.

In particular, we need a global view of face-area fractions if aquifer connections happen to be shared between processes.  To this end, add a new helper function,
```C++
BlackoilAquiferModel::computeConnectionAreaFraction()
```
that performs a collective operation to compute the total face areas and then defers to the local aquifer objects to compute their face area fractions.

### DIFF
--- a/opm/simulators/aquifers/AquiferAnalytical.hpp
+++ b/opm/simulators/aquifers/AquiferAnalytical.hpp
@@ -87,17 +87,45 @@ public:
                                           BlackoilIndices::numPhases>;
 
     // Constructor
-    AquiferAnalytical(int aqID,
-                     const std::vector<Aquancon::AquancCell>& connections,
-                     const Simulator& ebosSimulator)
+    AquiferAnalytical(const int aqID,
+                      const std::vector<Aquancon::AquancCell>& connections,
+                      const Simulator& ebosSimulator)
         : AquiferInterface<TypeTag>(aqID, ebosSimulator)
         , connections_(connections)
     {
+        this->initializeConnectionMappings();
     }
 
     // Destructor
     virtual ~AquiferAnalytical()
+    {}
+
+    void computeFaceAreaFraction(const std::vector<double>& total_face_area) override
     {
+        assert (total_face_area.size() >= static_cast<std::vector<double>::size_type>(this->aquiferID()));
+
+        const auto tfa = total_face_area[this->aquiferID() - 1];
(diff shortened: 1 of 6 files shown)


bska:
jenkins build this please

GitPaean:
I do not have the overview of the problem yet. Do you understand the failures in Jenins? Thanks. @bska

bska:
> Do you understand the failures in Jenins?

Not yet, and they're unexpected.  I'm looking into it.

bska:
jenkins build this please

blattms:
Unrelated to this and probably just a sign of my general ignorance of Aquifers:

There are some accumulation over alphai_ in the code when computing equilibrium in `calculateReservoirEquilibrium`. That code seems to assume that the whole alphai_/aquifer is on one process as there is no communication involved. If I am not missing soemething, then I would pose the question: Is that really guaranteed?

bska:
jenkins build this please

bska:

GitPaean:
Is it considered to be fixed and ready for review and potentially merging?

bska:
> Is it considered to be fixed and ready for review and potentially merging?

Yes.  This PR is very much ready for review/merging.

bska:
jenkins build this please

GitPaean:
jenkins build this please

blattms:
Since I did not get any reply to my comment (I assume that if it was relevant it is fixed) and we are seeing problems in parallel runs that might be related to bugs in the parallel treatment of parts of the simulator, it would be good to know:

Does this PR fix any parallelization issues?

bska:
> Does this PR fix any parallelization issues?

Not to my knowledge, no.  An early version of this introduced a couple of parallel issues, but we fixed those before merging.  I **think** we relaxed the assumption that `alphai_` is derived exclusively from contributions on a single rank ([revised calculation in `computeFaceAreaFraction()`](https://github.com/OPM/opm-simulators/blob/ed91f7528bbf110e219475e965a522275f3d44a6/opm/simulators/aquifers/AquiferAnalytical.hpp#L114-L120)), but I did not set out to analyse the complete communication patterns for all analytic aquifers.

We must do that separately.

blattms:
thanks for the valuable information.

PR #4579, opened 2023-04-03
Check all ALQ values when testing a gaslift well for reopening

hakonhagland:
When testing a gaslift well under THP control, if the well does not converge with maximum alq try to reduce alq in increments to check if the well equations converge with a smaller alq.

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -382,12 +382,16 @@ namespace Opm
         const double dt = ebosSimulator.timeStepSize();
         const auto& summary_state = ebosSimulator.vanguard().summaryState();
         const bool has_thp_limit = this->wellHasTHPConstraints(summary_state);
-        if (has_thp_limit)
+        bool converged;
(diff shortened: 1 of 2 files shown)


hakonhagland:
jenkins build this please

hakonhagland:
jenkins build this please

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -584,6 +588,42 @@ namespace Opm
...
+            bool converged;
+            while (alq > 0) {
+                well_state.setALQ(well_name, alq);
+                if ((converged =
+                      iterateWellEquations(ebos_simulator, dt, well_state, group_state, deferred_logger)))
+                {
+                    return converged;
+                }
+                alq -= increment;
+            }
```
totto82, line 619: Would it be better to start iterating from alq = 0.0 and increase it until either convergence or maximum alq rather than the opposite.

totto82: Thanks for the comment. Than the code is fine.

PR #4588, opened 2023-04-13
Geomech

hnil:
- branch with modification need for mechanics


ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -1138,9 +1164,69 @@ doAllocBuffers(unsigned bufferSize,
     rstKeywords["PRES"] = 0;
     rstKeywords["PRESSURE"] = 0;
 
-    // If TEMP is set in RPTRST we output temperature even if THERMAL
-    // is not activated
-    if (enableEnergy_ || rstKeywords["TEMP"] > 0) {
```
bska, line 1143: We need to be a little careful that we preserve this condition.  It was introduced in #4646, i.e., after the current PR was created, but it's mostly a matter of resolving a merge conflict I think.

ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -1138,9 +1164,69 @@ doAllocBuffers(unsigned bufferSize,
...
+    if (enableMech_ && eclState_.runspec().mech()){
+        this->mechPotentialForce_.resize(bufferSize,0.0);
+        rstKeywords["MECHPOTF"] = 0;
```
bska, line 1169: Unless we expect users to add these quantities to the `RPTRST` request, we don't really need to add the `rstKeyword["something"] = 0` statements here.

bska (commented):
This looks good for the most part.  I identified a couple of minor issues and we need to resolve the merge conflict, but other than that I think this should go in more or less "as is".

PR #4589, opened 2023-04-13
Remove two tests from regression testing

tskille:
There are issues with both of these that should be fixed before these are used for regression testing


tskille:
Two support tickets has been submitted to the Equinor M&S portal describing the issues with these regression tests.

bska:
jenkins build this please

bska (approved):
Okay, I'll merge this into the master branch.  We'll work hard to reenable these regression tests.

PR #4590, opened 2023-04-14
stop using saturated b-factors for case with zero rates

totto82:
I was debugging another issue and came pass this code. It seems wrong. Why are the saturated B factors used when rs/rv etc. = 0? 

I haven't tested this much, so I will make it a draft. I don't expect it to have huge impact on the solution.

### DIFF
--- a/opm/simulators/wells/StandardWellConnections.cpp
+++ b/opm/simulators/wells/StandardWellConnections.cpp
@@ -354,14 +354,11 @@ computePropertiesForPressures(const WellState& well_state,
         if (gasPresent) {
             const unsigned gasCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx);
             const int gaspos = gasCompIdx + perf * well_.numComponents();
-
-            if (oilPresent && waterPresent) {
+            double rvw = 0.0;
+            double rv = 0.0;
+            if (oilPresent) {
                 const double oilrate = std::abs(ws.surface_rates[pu.phase_pos[Oil]]); //in order to handle negative rates in producers
-                const double waterrate = std::abs(ws.surface_rates[pu.phase_pos[Water]]); //in order to handle negative rates in producers
                 props.rvmax_perf[perf] = FluidSystem::gasPvt().saturatedOilVaporizationFactor(region_idx, temperature, p_avg);
-                props.rvwmax_perf[perf] = FluidSystem::gasPvt().saturatedWaterVaporizationFactor(region_idx, temperature, p_avg);
-                double rv = 0.0;
-                double rvw = 0.0;
                 if (oilrate > 0) {
                     const double gasrate = std::abs(ws.surface_rates[pu.phase_pos[Gas]]) - (Indices::enableSolvent ? ws.sum_solvent_rates() : 0.0);
                     if (gasrate > 0) {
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

opm/simulators/wells/StandardWellConnections.cpp
```diff
@@ -364,94 +364,49 @@ computePropertiesForPressures(const WellState& well_state,
...
                     rvw = std::min(rvw, rvwmax_perf[perf]);
-
-                    b_perf[gaspos] = FluidSystem::gasPvt().inverseFormationVolumeFactor(region_idx, temperature, p_avg, 0.0 /*Rv*/, rvw);
                 }
-                else {
-                    b_perf[gaspos] = FluidSystem::gasPvt().saturatedInverseFormationVolumeFactor(region_idx, temperature, p_avg);
-                }
-
-            } else {
-                b_perf[gaspos] = FluidSystem::gasPvt().saturatedInverseFormationVolumeFactor(region_idx, temperature, p_avg);
```
GitPaean, line 431: Will we need this line anymore? I think it will only called when only `gas` exists?

GitPaean: or you think `rv==0.` and `rvw==0.0`, we only need `FluidSystem::gasPvt().inverseFormationVolumeFactor(region_idx, temperature, p_avg, rv, rvw);`?

GitPaean:
Generally, the PR itself is a very good cleaning up of the logic of the code. 

If the results confirm the correctness of the change,  then the PR should go in.

totto82:
benchmark please

GitPaean:
Do you see it is a pure cleaning up, or you think it is more like a correction, so some results might be changed? 

I am referring to the Jenkins failed cases.

totto82:
It is a bug fix and a clean-up. In the old code the saturated b factors where used when the pure values (rs = 0, etc) should have been used. It will give some differences in the bhp's and thus potentially also in the results. I will have to go through all the test failures to investigate them in detail.

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: drogon - Threads: 1         | 1.031    |
| opm-git | OPM Benchmark: drogon - Threads: 8         | 0.86     |
| opm-git | OPM Benchmark: smeaheia - Threads: 1       | 0.954    |
| opm-git | OPM Benchmark: smeaheia - Threads: 8       | 0.877    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 1  | 1.001    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 8  | 0.997    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.002    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.904    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 0.984    |



View result details @ https://www.ytelses.com/opm/?page=result&id=2045

totto82:
jenkins build this please

GitPaean:
The code does look cleaner with the PR, one question for here is whether the following code will give same results?

```c++
FluidSystem::oilPvt().inverseFormationVolumeFactor(region_idx, temperature, p_avg, 0 /*rs*/);
```
and
```c++
FluidSystem::oilPvt().saturatedInverseFormationVolumeFactor(region_idx, temperature, p_avg);
```

Do you have some ideas? @totto82 @bska

GitPaean:
jenkins build this please

totto82:
> The code does look cleaner with the PR, one question for here is whether the following two function calls will give same results?
> 
> FluidSystem::oilPvt().inverseFormationVolumeFactor(region_idx, temperature, p_avg, 0 /*rs*/);
> and
> 
> FluidSystem::oilPvt().saturatedInverseFormationVolumeFactor(region_idx, temperature, p_avg);
> Do you have some ideas? @totto82 @bska

No the dont. that is why this is a fix.
FluidSystem::oilPvt().saturatedInverseFormationVolumeFactor(region_idx, temperature, p_avg); gives the same as 

> FluidSystem::oilPvt().inverseFormationVolumeFactor(region_idx, temperature, p_avg, rsSat);

totto82:
I am going through the test failures as we speak. I dont expect large differences, but some differences is expected. I will report back soon.

GitPaean:
> > The code does look cleaner with the PR, one question for here is whether the following two function calls will give same results?
> > ```c++
> > FluidSystem::oilPvt().inverseFormationVolumeFactor(region_idx, temperature, p_avg, 0 /*rs*/);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and
> > ```c++
> > FluidSystem::oilPvt().saturatedInverseFormationVolumeFactor(region_idx, temperature, p_avg);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do you have some ideas? @totto82 @bska
> 
> No the dont. that is why this is a fix.
> 
> ```c++
> FluidSystem::oilPvt().saturatedInverseFormationVolumeFactor(region_idx, temperature, p_avg);
> ```
> 
> gives the same as
> 
> ```c++
> FluidSystem::oilPvt().inverseFormationVolumeFactor(region_idx, temperature, p_avg, rsSat);
> ```

Yeah. I just see the reply now. Github does not refresh the webpage. 

Sure. Then I think it is okay and approve the PR. But as mentioned above, I would like @bska to have a look at the PR.

GitPaean:
You can mark the PR as ready for review if you agree. @totto82

totto82:
> As a rule of thumb, if IWEL, IGRP, or SCON differ we're going to need more justification than "ok". 

 "ok" just means that I have done some manual checking. 

For instance MAX_WATERCUT_1


image


You clearly see differences between this PR and the reference solution. 

But if you run with --enable-tuning. It looks like this 


image

totto82:
Another example 9_4C_WINJ_GINJ_VREP-W_REIN-G_STW. Again the mismatch is gone when --enable-tuning=true

image

totto82:
Next is ACTIONX_WPIMULT

image

totto82:
And finally 0B_WELPI_MODEL2

image

totto82:
Can this be merged? From my point of view this fixes a bug in the code and should be merged. Using the saturated b values when rs/rv etc. is zero (component not present) can not be correct.

GitPaean:
> Next is ACTIONX_WPIMULT
> 
> image

Does the green part indicate a lot of oscillation?

totto82:
> Does the green part indicate a lot of oscillation?

Yes, but notice that the oscillation are both with and without this PR.

GitPaean:
jenkins build this please

GitPaean:
benchmark please

GitPaean:
For the case 0B_WELPI_MODEL2, can you check what happened with the well PROD3 with this PR?

image

GitPaean:
And also the case ACTIONX_M1?

image

GitPaean:
and the case MAX_WATERCUT_4?

image

totto82:
For  0B_WELPI_MODEL2 the situation is that if run with small time-steps you get oscillation. Both solutions are within the solution with maks 1 day timestep. 


image

GitPaean:
ACTIONX_WPIMULT shares the same problem with the case ACTIONX_M1

I only manually checked the cases from the comment https://github.com/OPM/opm-simulators/pull/4590#issuecomment-1549200527

And I believe most of the problem might disappear with the TUNING keyword on to use the same time stepping.  Time stepping is part of the reason some cases failed their regression test.  But it is good to understand the behavoir, which is costly to do. 

Should we re-design our regression test differently to limit the effect of different time steps? but the at the same time, our time-stepping strategy is something we also want to test.

totto82:
image

totto82:
> Thanks. I checked the WMCTL, it looks like RESV control. But do you understand why it oscillates so much? There might be bug while not related to this PR?

I haven't looked into why it oscillates when using small time-steps. I can do that, but as you said, it would't be related to this PR.

GitPaean:
> I haven't looked into why it oscillates when using small time-steps. I can do that, but as you said, it would't be related to this PR.

My point is that the oscillation with small time steps kind of reveals why the different results without tuning.  If we understand it, it might be bug in our code in master branch or problem of the case setup. From that, we can either fix the bug in the code or change the case setup to make future regression test for this specific case easier. It is very costly to check regression failures manually since we have so many of them now.

totto82:
> And I agree, this is not directly due to this PR, and not sure how we should handle this types of situation.

I think if we agree that the code in question is correct. We should just update the reference data and move on. 

As we have discussed earlier also in other settings I would really like a way to call the test suite with some options. LIke jenkins build this strict please. That will run and compare with a reference case computed with fixed and small timesteps and maybe also tighter tolerance for the linear solver. That is basically what we are doing manually now.

GitPaean:
> And also the case ACTIONX_M1?
> 
> image

For ACTIONX_M1, with TUNING on, the difference is still significant, it might be worth some understanding. It is possible the PR make the results **better**, but we do not know yet. 

image

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: drogon - Threads: 1         | 0.987    |
| opm-git | OPM Benchmark: drogon - Threads: 8         | 1.004    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 1  | 0.998    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 8  | 1.008    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 0.984    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.992    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 0.989    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.028    |



View result details @ https://www.ytelses.com/opm/?page=result&id=2102

totto82:
This is a comparison of this PR and the reference case. I think the results are close enough. Some deviations are expected since we actually change the density of the fluids in the well slightly. 

image

GitPaean:
> This is a comparison of this PR and the reference case. I think the results are close enough. Some deviations are expected since we actually change the density of the fluids in the well slightly.
> 
> image

Thanks for checking, but my local running is still the same with https://github.com/OPM/opm-simulators/pull/4590#issuecomment-1568135852

Can you double check? I am using DUNE 2.8.  And also, we just need to understand the cause of the difference. Sometimes, not necessarily which one is better.

GitPaean:
benchmark please

totto82:
@GitPaean Can we start the process of merging this now? This fixes a bug in the code. The current code uses the wrong b-factors. We have verified that it does not significantly change the performance and that the results does not change in an unexpected way.

GitPaean:
For the case `ACTIONX_M1`, I am confirm the result difference is from different time stepping. I am not familiar with the ACTIONX  keyword, but the following is my interpretation of the results.  Black line is the result of the master branch, the green line is the result of the PR. 

The following is the field pressure, 
image

And the following the WWIR of the well WI1
image

Around FPR at around 200 bar, the following ACTIONX is relevant related to well WI1, 

```
372 WCONINJE    --written by schedule script
373 --  Well_name   Type    Status  Ctrl    SRate1  Rrate   BHP THP VFP  
374     'WI1'   WATER   STOP    GRUP    2000     1* 325.0   1*  1*  /
375 /
376 
...................................
..................................
388 ACTIONX
389 INJECTION 10 /
390 WBHP P1 < 200.0 /
391 /
392 WELOPEN
393   'WI1' 'OPEN' 5* /
394 /
395 
```

GitPaean:
> @GitPaean Can we start the process of merging this now? This fixes a bug in the code. The current code uses the wrong b-factors. We have verified that it does not significantly change the performance and that the results does not change in an unexpected way.

I guess we can. 

At the same time, we are dependent on the jenkins test, it is not easy to just say okay to the discrepancy and go ahead. I agree totally with your point regarding the significant efforts involved in manually checking jenkins failures and we have done  many many times and it is a huge burden for development. Some more automatic process will largely favorable.

totto82:

totto82:
jenkins build this opm-tests=987 please

PR #4591, opened 2023-04-14
Fixes related to rsw/rvw handling in standard well

totto82:
I am debugging a case with rsw/rvw and found some suspicious code. I didn't fix my problem, but I think the fixes are correct.


totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
> Is it planned to make DISGASW generally applicable for a three-phase system?

Not that I am aware of. But adding an option to allow for DISGASW in a general two-phase system with gas+water shouldn't be too difficult. It would only require adding table input. A general case where gas is allowed to both dissolve in water and oil will need more work and a clear use case.

PR #4606, opened 2023-04-25
Serialized restart: Add some more checks

akva2:
One small fix, one added check for being more explicit about what is wrong, and one extra check to avoid misunderstandings.

### DIFF
--- a/opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.cpp
+++ b/opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.cpp
@@ -22,11 +22,14 @@
 #include <config.h>
 
 #include <opm/common/OpmLog/OpmLog.hpp>
+#include <opm/common/utility/String.hpp>
+
 #include <opm/input/eclipse/Units/Units.hpp>
 #include <opm/simulators/timestepping/SimulatorTimer.hpp>
 
 #include <boost/date_time.hpp>
 
+#include <algorithm>
 #include <sstream>
 #include <string>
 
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this please

opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.cpp
```diff
@@ -61,4 +65,63 @@ void outputTimestampFIP(const SimulatorTimer& timer,
...
+    {
+        std::vector<std::string> output;
+        output.reserve(input.size());
+        std::copy_if(input.begin(), input.end(), std::back_inserter(output),
+                     [](const std::string& line)
+                     {
+                        return line.compare(0, 11, "EclDeckFile") != 0 &&
+                               line.compare(0, 8, "LoadStep") != 0 &&
+                               line.compare(0, 8, "SaveFile") != 0 &&
+                               line.compare(0, 8, "SaveStep") != 0;
```
bska, line 81: Any reason not to use `line != ...` here?

opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.cpp
```diff
@@ -61,4 +65,63 @@ void outputTimestampFIP(const SimulatorTimer& timer,
...
+
+    const std::string curr_file = "/tmp/" + unique_path("current-%%%%");
+    const std::string stored_file = "/tmp/" + unique_path("stored-%%%%");
+
+    write_strings(stored_file, stored_strings);
+    write_strings(curr_file, curr_strings);
+
+    std::stringstream cmd;
+    cmd << "diff -u0 " << stored_file << " "  << curr_file << " | tail -n +4";
+    FILE* pipe = popen(cmd.str().c_str(), "r");
```
bska, line 109: According to Posix, you need `<stdio.h>` for `popen()` and `pclose()`.  Might also want to wrap this in a `unique_ptr` with a custom deleter.

opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.cpp
```diff
@@ -61,4 +65,63 @@ void outputTimestampFIP(const SimulatorTimer& timer,
...
+                         };
+    auto curr_strings = split_string(current, '\n');
+    auto stored_strings = split_string(stored, '\n');
+    std::sort(curr_strings.begin(), curr_strings.end());
+    std::sort(stored_strings.begin(), stored_strings.end());
+    curr_strings = filter_strings(curr_strings);
+    stored_strings = filter_strings(stored_strings);
+
+    const std::string curr_file = "/tmp/" + unique_path("current-%%%%");
+    const std::string stored_file = "/tmp/" + unique_path("stored-%%%%");
```
bska, line 102: Can we assume that `/tmp` is always available?  If not, we might want `filesystem::temp_directory_path()` instead.

opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.cpp
```diff
@@ -61,4 +65,63 @@ void outputTimestampFIP(const SimulatorTimer& timer,
...
+    cmd << "diff -u0 " << stored_file << " "  << curr_file << " | tail -n +4";
+    FILE* pipe = popen(cmd.str().c_str(), "r");
+    if (pipe) {
+      std::array<char,1024> buffer;
+      std::stringstream output;
+      while (fgets(buffer.data(), 1024, pipe) != nullptr) {
+          output << buffer.data();
+      }
+      pclose(pipe);
+      if (output.str().size() != 0) {
```
bska, line 117: `! empty()` instead?

bska (commented):
I made couple of comments regarding the implementation of `checkSerializedCmdLine()`, but then I thought that we might not need to invoke an external tool at all.  Is there a reason you can't use `std::set_difference()` and/or ` std::set_symmetric_difference()` to perform the comparison?

opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.cpp
akva2: yes. ideally it would be .startswith() but that's c++20. those params have values, i want to ignore these lines no matter their values.

akva2:
I originally used set_symmetric_difference but got annoyed by the sorted output. But I guess i'll just stomach that one. implementation changed.

opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.cpp
akva2: now irrelevant.

opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.cpp
akva2: since i changed to set_symmetric_difference this is irrelevant.

akva2:
jenkins build this please

bska:
> I originally used set_symmetric_difference but got annoyed by the sorted output.

Right.  If you **really** wanted to work around that you could do something like
```C++
struct StringListEntry
{
    std::size_t list{};
    std::vector<std::string>::size_type index{};
};

std::vector<StringListEntry>
sortList(std::size_t listID, const std::vector<std::string>& list)
{
    std::vector<StringListEntry> sortedList{};

    std::vector<std::vector<std::string>::size_type> index(list.size());
    std::iota(index.begin(), index.end(), 0);
    std::sort(index.begin(), index.end(),
        [&list](const auto i1, const auto i2)
    { return list[i1] < list[i2]; });

    sortedList.reserve(index.size());
    for (const auto& ix : index) {
        sortedList.push_back({ listID, ix });
    }

    return sortedList;
}

void checkSerializedCmdLine(const std::string& current,
                            const std::string& stored)
{
    // ...
    const auto strings = std::array {
        filter_strings(split_string(current)),
        filter_strings(split_string(stored))
    };
    const auto s1 = sortList(0, strings[0]);
    const auto s2 = sortList(1, strings[1]);

    auto diff = std::vector<StringListEntry>{};
    std::set_symmetric_difference(s1.begin(), s1.end(),
                                  s2.begin(), s2.end(),
                                  std::back_inserter(diff),
        [&strings](const auto& e1, const auto& e2)
    {
        return strings[e1.list][e1.index] < strings[e2.list][e2.index];
    });

    // Output 'diff' as needed
}
```
but I suppose that might be taking things too far.

> implementation changed.

Much appreciated.  I'll merge into master.

akva2:
don't merge! there are issues

bska:
> don't merge! there are issues

Okay, I'll hold.

akva2:
jenkins build this please

akva2:
there. previous implementation would go bad if there are parameters in just one of the sets. this happens if options are added or removed. now we detect this and output appropriate information.

akva2:
good to go?

bska:
> good to go?

Yes.  Thanks a lot for the updates.  I'll merge into master.

PR #4609, opened 2023-04-26
WellGroupHelpers: make sure we have a GuideRate value or potential for well

akva2:
this is not the case when WTEST executes at the beginning of (the first time step of) a report step if the well was created at that report step. this led to an exception being thrown trying to deref the potentials array

thanks to [at]bska for helping me pinpoint where to apply the fix


akva2:
jenkins build this please

bska:
We might consider back-porting this to the 2023.04 release.  It fixes a rather severe, and hard to diagnose, underlying problem.

akva2:
okay to merge ?

PR #4615, opened 2023-04-27
only prediction wells can have THP constraint

GitPaean:
This is reported and provided by @vkip . I think it can be one way of the solutions. 

An alternative way is to use the following code, 

```c++
    if (this->wellEcl().predictionMode() && dynamic_thp_limit_) {
        return true;
    }
```

It can make differences if we make mistakes from the parsing or other very special situations, but we have not seen it yet. The way in the PR should be the correct way.

### DIFF
--- a/opm/simulators/wells/WellInterfaceGeneric.cpp
+++ b/opm/simulators/wells/WellInterfaceGeneric.cpp
@@ -192,6 +192,11 @@ double WellInterfaceGeneric::rsRvInj() const
 
 bool WellInterfaceGeneric::wellHasTHPConstraints(const SummaryState& summaryState) const
 {
+    // only wells under prediction mode can have THP constraint
+    if (!this->wellEcl().predictionMode()) {
+        return false;
+    }
+
     if (dynamic_thp_limit_) {
         return true;
     }


GitPaean:
jenkins build this please

GitPaean:
For history matching wells, we can only have BHP and one type of rate constraint.

bska:
Any change we make to `wellHasTHPConstraints()` will affect, at least, the behaviour of our well-testing and our gas-lift optimisation.  Before we make any changes here, we need to ensure that there are no models using `WTEST` and/or `LIFTOPT` (particularly `WLIFTOPT`) on history-matching wells.  It's very unlikely, of course, but I think we need to take that possibility into consideration until we can prove that it does not happen.

GitPaean:
I can be wrong here, but my feeling is that WTEST and WLIFTOPT are for prediction-mode wells.  Maybe people having more experience in this can have some suggestions here. @OPMUSER @tskille 

At the same time, we have examples clearly showing that we have to avoid THP constraint for history matching wells (the constraint is from the network pressure with the master branch).  We need the content in this PR to make those examples straight.

bska (approved):
Having mulled over this a bit more, I think my initial concerns about `WTEST` and gas lift don't really apply here.  Those are decidedly more applicable in prediction cases.  If we come across simulation models which apply well testing and/or gas lift to history matching wells we'll have to find a better solution.

In any case, I'll merge the current state into master.

PR #4616, opened 2023-04-27
removing the unused BlackoilWellModel::hasWell()

GitPaean:
We have two `hasWell()` function in `BlackoilWellModel` and `BlackoilWellModelGeneric` , respectively, with different definition.  The one in `BlackoilWellModel` is hiding the one from `BlackoilWellModelGeneric`.  And the one in `BlackoilWellModel` is actually not used in the simulation code. 

We should remove one and I am removing the one not in use, while not knowing for sure which one should be kept from the function definition perspective.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -305,7 +305,6 @@ namespace Opm {
                                     const int iterationIdx);
 
             WellInterfacePtr getWell(const std::string& well_name) const;
-            bool hasWell(const std::string& well_name) const;
 
             using PressureMatrix = Dune::BCRSMatrix<Opm::MatrixBlock<double, 1, 1>>;
 
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1902,18 +1902,6 @@ namespace Opm {
         return *well;
     }
 
-    template<typename TypeTag>
-    bool
-    BlackoilWellModel<TypeTag>::
-    hasWell(const std::string& well_name) const
-    {
-        return std::any_of(well_container_.begin(), well_container_.end(),
-            [&well_name](const WellInterfacePtr& elem) -> bool
-        {
-            return elem->name() == well_name;
-        });
-    }
-
 
 
 


GitPaean:
jenkins build this please

akva2:
I'm a bit worried.

hasWell() was moved to BlackoilWellModelGeneric.cpp in my refactoring https://github.com/OPM/opm-simulators/commit/059140f2ee47776af471d5306889fccdeb67ffa9

Then the one in BlackoilWellModel_impl.hpp was later added by @totto82  in https://github.com/OPM/opm-simulators/commit/31ac5378cf499ecd3cfd5e17db0e0947ce5822da

They differ in one rather subtle but important aspect: The one in the generic checks the wells_ecl_ array. This is an array of all wells. The one in BlackoilWellModel_impl checks the well_container_ array. Unless I'm mistaken this only holds active local wells.

So I think they are both needed, but that it's unfortunately that they have the same name. Please comment @totto82

GitPaean:
> Then the one in BlackoilWellModel_impl.hpp was later added by @totto82 in https://github.com/OPM/opm-simulators/commit/31ac5378cf499ecd3cfd5e17db0e0947ce5822da

Thanks for pointing this out and I did not check ebos/ecltracermodel.hh .

bska:
> They differ in one rather subtle but important aspect: The one in the generic checks the wells_ecl_ array. This is an array of all wells. The one in BlackoilWellModel_impl checks the well_container_ array. Unless I'm mistaken this only holds active local wells.

You're right, `well_container_` holds those wells that are active on the current rank at the current report step whereas `wells_ecl_` holds all wells active in the model at the current report step.  Looking at the prospective call sites,

1. The one in [`loadRestartGuideRates()`](https://github.com/OPM/opm-simulators/blob/525cc76d623a568844b56c05669c949a336eacb4/opm/simulators/wells/BlackoilWellModelRestart.cpp#L190) unequivocally assumes that `hasWell()` applies to all active model wells
2. The one in [`fetchWellPI()`](https://github.com/OPM/opm-simulators/blob/525cc76d623a568844b56c05669c949a336eacb4/ebos/eclactionhandler.cc#L224) is less clear to me.  The code is written in a way that appears to use all active model wells, but it could **probably** use the on-rank active ones as well.

GitPaean:
It looks like `wells_ecl_` only handle the wells on the process also?  Not sure whether they need to be active?

https://github.com/OPM/opm-simulators/blob/525cc76d623a568844b56c05669c949a336eacb4/opm/simulators/wells/BlackoilWellModelGeneric.cpp#L180

and 

https://github.com/OPM/opm-simulators/blob/525cc76d623a568844b56c05669c949a336eacb4/opm/simulators/wells/BlackoilWellModelGeneric.cpp#L246

```c++
std::vector<Well>
BlackoilWellModelGeneric::
getLocalWells(const int timeStepIdx) const
{
    auto w = schedule().getWells(timeStepIdx);
    w.erase(std::remove_if(w.begin(), w.end(), not_on_process_), w.end());
    return w;
}
```

Sorry I misunderstood some comments above.

GitPaean:
Anyway, we probably should not simply delete one function. Most likely we need to refactor the code a little bit to remove the confusion and hiding of functions. I am converting this PR to be draft now.

bska:
> It looks like `wells_ecl_` only handle the wells on the process also? Not sure whether they need to be active?

Ehm, yes, you're right.  `Wells_ecl_` does hold only those wells that have at least one reservoir connection on the current rank, but does not care if the well is open/flowing or shut/stopped.  `Well_container_`, on the other hand, holds only those wells, on the current rank, that go into the system of (non-)linear equations.

atgeirr:
Since we seem to need both concepts:
 - all wells present on this process,
 - all non-shut wells on this process (i.e. wells that should be solved),
we should give the concept names and change the names of the accessor methods to match, probably also the underlying member variables. I suggest: allLocalWells() and activeLocalWells(), but this are just preliminary suggestions.

GitPaean:
benchmark please

akva2:
taken care of through https://github.com/OPM/opm-simulators/pull/5849

PR #4617, opened 2023-05-01
updating the interval when doing bisectBracket()

GitPaean:
to avoid getting stuck and iterate for no purpose. It might not affect the result much, while from the code, it looks like it should be updated iteratively.


GitPaean:
jenkins build this please

bska (approved):
Yes, this looks like a straightforward bug to me.  The bisection algorithm should keep track of the size of its search interval.  I'll let @atgeirr have a second look, but I'm in favour of merging this.

atgeirr:
Yes, this is a pure bugfix, min_interval would be ignored before. It is set in the code to 1 bar, so after merging this, we may consider modifying it if it is too large or small.

PR #4618, opened 2023-05-02
fixed: conditionally use Development.XXX in python find module

akva2:
the split into Development.Module and Development.Embed was is introduced in cmake 3.18


akva2:
jenkins build this please

PR #4620, opened 2023-05-03
Add support for WVFPDP

hakonhagland:
This PR together with https://github.com/OPM/opm-common/pull/3504 implements support for keyword WVFPDP. The implementation show good agreement with an Eclipse reference solution for test case [WVFPDP-01.DATA](https://github.com/OPM/opm-tests/blob/master/vfp/WVFPDP-01.DATA) :

image

### DIFF
--- a/opm/simulators/wells/WellBhpThpCalculator.cpp
+++ b/opm/simulators/wells/WellBhpThpCalculator.cpp
@@ -166,7 +166,6 @@ findThpFromBhpIteratively(
     bool do_iterate = true;
     int it = 1;
     int max_iterations = 50;
-    bool debug = true;
     while(do_iterate) {
         if (it > max_iterations) {
             break;
@@ -323,7 +322,7 @@ void WellBhpThpCalculator::updateThp(const double rho,
     }
     const std::optional<double> alq = this->well_.isProducer() ? std::optional<double>(alq_value()) : std::nullopt;
     const double thp_limit = well_.getTHPConstraint(summary_state);
-    ws.thp = this->calculateThpFromBhp(rates, ws.bhp, rho, alq_value(), thp_limit, deferred_logger);
+    ws.thp = this->calculateThpFromBhp(rates, ws.bhp, rho, alq, thp_limit, deferred_logger);
 }
 
 template<class EvalWell>
--- a/opm/simulators/wells/WellBhpThpCalculator.cpp
+++ b/opm/simulators/wells/WellBhpThpCalculator.cpp
@@ -28,8 +28,8 @@
 
 #include <opm/input/eclipse/Schedule/VFPInjTable.hpp>
 #include <opm/input/eclipse/Schedule/Well/Well.hpp>
-
 #include <opm/input/eclipse/Units/Units.hpp>
+#include <opm/input/eclipse/Schedule/Well/WVFPDP.hpp>
 
 #include <opm/material/densead/Evaluation.hpp>
 
@@ -181,8 +181,16 @@ computeBhpAtThpLimitProd(const std::function<std::vector<double>(const double)>&
         const auto& wfr =  well_.vfpProperties()->getExplicitWFR(controls.vfp_table_number, well_.indexOfWell());
         const auto& gfr = well_.vfpProperties()->getExplicitGFR(controls.vfp_table_number, well_.indexOfWell());
         const bool use_vfpexp = well_.useVfpExplicit();
-        return well_.vfpProperties()->getProd()
-        ->bhp(controls.vfp_table_number, rates[Water], rates[Oil], rates[Gas], thp_limit, alq_value, wfr, gfr, use_vfpexp) - dp;
+        const double bhp = well_.vfpProperties()->getProd()->bhp(controls.vfp_table_number,
+                                                     rates[Water],
+                                                     rates[Oil],
+                                                     rates[Gas],
+                                                     thp_limit,
+                                                     alq_value,
+                                                     wfr,
+                                                     gfr,
+                                                     use_vfpexp);
+        return bhp - dp + getVfpBhpAdjustment(bhp, thp_limit);
     };
 
     // Make the flo() function.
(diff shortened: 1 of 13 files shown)


opm/simulators/wells/WellBhpThpCalculator.cpp
```diff
@@ -288,32 +297,47 @@ calculateBhpFromThp(const WellState& well_state,
...
+        OPM_DEFLOG_THROW(std::logic_error, "Expected INJECTOR or PRODUCER for well " + well_.name(), deferred_logger);
+    }
+    double bhp_tab_double_value;
+    if constexpr (std::is_same_v<EvalWell, double>) {
+        bhp_tab_double_value = bhp_tab;
+    }
+    else {  // EvalWell and bhp_tab is of type DenseAd::Evaluation<double,...,...>
+        bhp_tab_double_value = bhp_tab.value();
+    }
+    const auto bhp_adjustment = getVfpBhpAdjustment(bhp_tab_double_value, thp_limit);
```
totto82, line 332: It is better to use bhp_tab_double_value = getValue(bhp_tab) then you don't need the std::is_same check.

totto82 (changes requested):
Thanks. Looks good, can be merged after my single comment has been addressed.

opm/simulators/wells/WellBhpThpCalculator.cpp
hakonhagland: Good idea. I assume the function `getValue()` is the one declared here: https://github.com/OPM/opm-material/blob/ee6c07fef239219d2470bc9be482ca894d2e60df/opm/material/common/MathToolbox.hpp#L330 ? If I try this the compiler says:

```
/home/hakon/test/opm11/opm-common/opm/material/common/MathToolbox.hpp:51:52: error: static assertion failed: This class expects floating point scalars! (specialization missing?)
   51 |     static_assert(std::is_floating_point<ScalarT>::value,
      |                                                    ^~~~~
/home/hakon/test/opm11/opm-common/opm/material/common/MathToolbox.hpp:51:52: note: ‘std::integral_constant<bool, false>::value’ evaluates to false
/home/hakon/test/opm11/opm-simulators/opm/simulators/wells/WellBhpThpCalculator.cpp: In instantiation of ‘EvalWell Opm::WellBhpThpCalculator::calculateBhpFromThp(const Opm::WellState&, const std::vector<EvalWell>&, const Opm::Well&, const Opm::SummaryState&, double, Opm::DeferredLogger&) const [with EvalWell = Opm::DenseAd::Evaluation<double, 3, 0>]’:
/home/hakon/test/opm11/opm-simulators/opm/simulators/wells/WellBhpThpCalculator.cpp:793:1:   required from here
/home/hakon/test/opm11/opm-simulators/opm/simulators/wells/WellBhpThpCalculator.cpp:326:66: error: cannot convert ‘Opm::MathToolbox<Opm::DenseAd::Evaluation<double, 3, 0> >::Scalar’ {aka ‘Opm::DenseAd::Evaluation<double, 3, 0>’} to ‘double’
  326 |     const auto bhp_adjustment = getVfpBhpAdjustment(Opm::getValue(bhp_tab), thp_limit);
      |                                                     ~~~~~~~~~~~~~^~~~~~~~~
      |                                                                  |
      |                                                                  Opm::MathToolbox<Opm::DenseAd::Evaluation<double, 3, 0> >::Scalar {aka Opm::DenseAd::Evaluation<double, 3, 0>}
/home/hakon/test/opm11/opm-simulators/opm/simulators/wells/WellBhpThpCalculator.cpp:332:63: note:   initializing argument 1 of ‘double Opm::WellBhpThpCalculator::getVfpBhpAdjustment(double, double) const’
  332 | double WellBhpThpCalculator::getVfpBhpAdjustment(const double bhp_tab, const double thp_limit) const
      |                                                  ~~~~~~~~~~~~~^~~~~~~
```
Do I need to give a template argument to `getValue()` or is it something else?

totto82: getValue will convert from Evaluation to Scalar. Seems like the compiler complains about conversion from Scalar to double 

Try 

`double bhp_tab_double_value = getValue(bhp_tab);`

and than use bhp_tab_double_value in the functions.

hakonhagland:
Note that @tskille has reported some issues with the THP in private conversation. This can be seen in the following plot for test case [WVFPDP-01](https://github.com/OPM/opm-tests/blob/master/vfp/WVFPDP-01.DATA) :

image

Notice the spike at 04.OCT.2018, and that the THP drops below the limit of 75 bars in July 2018. If I set item 2 at line 314 to zero

https://github.com/OPM/opm-tests/blob/5e9dff3417f9581022555cad0eb0fcb80dfe544d/vfp/WVFPDP-01.DATA#L314 

the THP looks more correct:

image

I am investigating this further.

hakonhagland:
> some issues with the THP

Think I found the problem, see last commit. The result for test case [WVFPDP-01](https://github.com/OPM/opm-tests/blob/master/vfp/WVFPDP-01.DATA) now looks like this:

image

hakonhagland:
There is still a small oscillation below the THP limit at 01.Aug.2018 for the second test case [WVFPDP-02](https://github.com/OPM/opm-tests/blob/master/vfp/WVFPDP-02.DATA):

image

I am investigating this further.

hakonhagland:
> There is still a small oscillation below the THP limit at 01.Aug.2018 for the second test case 

We need to solve for thp iteratively, see last commit. Thanks to @totto82 for this solution.

akva2:
jenkins build this opm-common=3504 please

hakonhagland:
> jenkins build this opm-common=3504 please

@akva2 Hmm, seems like the jenkins build from here (compared to the jenkins build from https://github.com/OPM/opm-common/pull/3504) passed.. any idea why?

akva2:
Yes. Rebase the opm-simulators pr.

PR #4621, opened 2023-05-03
Use mass balance limit (XXXMBE) from TUNING with --enable-tuning=true

vkip:
Requires https://github.com/OPM/opm-common/pull/3505


atgeirr:
I have looked at this and the companion in opm-common and it is fine. (It does show that perhaps we should not recreate the solver all the time, that is a different question). However, do you really want this to only apply to explicitly overridden TUNING records? Especially since the Flow default of 1e-6 is the same as the default for XXXMBE. It would be simpler to just use the TUNING record always (when --enable-tuning=true), and it would ensure that if we ever change the Flow default, with TUNING enabled one would get the same behaviour even then.

vkip:
The reason for only applying the change when explicitly overridden is that TUNING is often used to change specific parameters (e.g., maximum time step size) throughout a simulation, in which case most of the parameters will be defaulted. I think it would be quite cumbersome to always have to specify all parameters each time you only want to change one, and personally I also find it non-intuitive that unspecified parameters are implicitly reset. So yes, I really want this :)

bska:
> personally I also find it non-intuitive that unspecified parameters are implicitly reset

That's possibly a bug, then.  Looking at [the logic](https://github.com/OPM/opm-common/blob/644d371026b3ba8dcf9c4b6cd605ca55e1b0f1b7/src/opm/input/eclipse/Schedule/KeywordHandlers.cpp#L928) for how we internalise `TUNING` we always "read" every item of every record into our internal representation, and doing so will reset items not specifically assigned to their default values.

If the proper semantics is to only update those items that have user-defined values, then I would suggest to add `defaultApplied()` checks in `handleTUNING()` instead.

vkip:
What is 'proper semantics' can be discussed I guess, but this is what I would like to have at least. See the point about defaultApplied(), will try to update the PRs accordingly.

vkip:
Now depends on https://github.com/OPM/opm-common/pull/3522 to get the right semantics

bska:
jenkins build this please

vkip:
jenkins build this opm-tests=1006 please

bska:
So this doesn't require anything other than OPM/opm-common#3522 which we already merged last month, right?

vkip:
> So this doesn't require anything other than [OPM/opm-common#3522](https://github.com/OPM/opm-common/pull/3522) which we already merged last month, right?

Yes, that is correct.

bska:
> > So this doesn't require anything other than [OPM/opm-common#3522](https://github.com/OPM/opm-common/pull/3522) which we already merged last month, right?
> 
> Yes, that is correct.

Excellent.  Thanks for clarifying.

tests/test_tuning_XXXMBE.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+		// Return vector of double values values, invalid elements set to std::numeric_limits<int>::min()
+		std::vector<int> ivalues() const {
+				std::vector<int> vec;
+				vec.reserve(this->size());
+				
+				const auto& conv_func = [](const std::string& strval) {
+						int ival;
+						try {
+								ival = static_cast<int>(std::stod(strval));
```
bska, line 47: Is there a reason for not using [`std::stoi()`](https://en.cppreference.com/w/cpp/string/basic_string/stol) here?

tests/test_tuning_XXXMBE.cpp
```diff
@@ -0,0 +1,157 @@
...
+				// Read remaining lines into std::string vectors
+				int lineno = 1;
+				while (std::getline(ifs, line)) {
+						iss.str(line); iss.clear();
+						int i=0;
+						while (iss >> colname && i < num_columns) {
+								raw_columns[i].push_back(colname);
+								++i;
+						}
+						if (i > num_columns) {
```
bska, line 91: I don't think this is possible.  If you want to check for this condition you'll have to remove the `i < num_colums` condition in the `while()` loop.

tests/test_tuning_XXXMBE.cpp
```diff
@@ -0,0 +1,157 @@
...
+						}
+						++lineno;
+				}
+		}
+
+		// Get data vectors of different types
+		std::vector<double> get_dvector(const std::string& colname) { return columns[colname]->dvalues(); }
+		std::vector<int> get_ivector(const std::string& colname) { return columns[colname]->ivalues(); }
+		// Default is to return double values
+		std::vector<double> operator[](const std::string& colname) { return columns[colname]->dvalues(); }
```
bska, line 102: Strictly speaking, `columns[colname]` will always (attempt to) insert a new node into the `columns` map.  Maybe that's fine for your purposes here, but if it's not then we should probably write these as
```C++
columns.at(colname)->...
```
instead.  Then you could also declare the member functions as `const`.

tests/test_tuning_XXXMBE.cpp
```diff
@@ -0,0 +1,157 @@
...
+		std::map<std::string, Column*> columns;
+};
+
+BOOST_AUTO_TEST_CASE(CheckMassBalanceWithinXXXMBE)
+{
+		//std::string case_name(boost::unit_test::framework::master_test_suite().argv[1]);
+		std::string case_name("01_TUNING_XXXMBE");
+		std::string file_name = case_name + ".INFOITER";
+
+		ColumnData data(file_name);
```
bska, line 115: It's not obvious to me that these file references work in general, so I'll ask @akva2 to weigh in on the test setup parts of this.

tests/test_tuning_XXXMBE.cpp
```diff
@@ -0,0 +1,157 @@
...
+						rcur = rstep[i+1];
+				}			
+		}
+		max_mb.push_back( std::max({mbo.back(), mbw.back(), mbg.back(), max_mb_step}) );
+
+		std::cout << "---------------------------------------------------------------------------" << std::endl;
+		std::cout << "Found the following converged max mass balance error (per report step):" << std::endl;
+		for (auto& val : max_mb)
+				std::cout << val << std::endl;
+		std::cout << "---------------------------------------------------------------------------" << std::endl;
```
bska, line 149: Could you use [`BOOST_TEST_MESSAGE()`](https://www.boost.org/doc/libs/1_82_0/libs/test/doc/html/boost_test/test_output/test_tools_support_for_logging/test_output_macro_message.html) to output these messages instead?  The one drawback to `BOOST_TEST_MESSAGE()` is that you don't get the messages unless you pass
```
--log_level=message
```
to the test runner, but you **do** get the advantage of the thread-safe logger built into Boost.Test.

bska (commented):
The C++ parts of this PR look good to me.  I'll ask @akva2 to comment on the test setup parts and, particularly, on the filename paths.  Maybe there's a different way of accomplishing what you're trying to do?

compareECLFiles.cmake
```diff
@@ -37,9 +37,38 @@ function(add_test_runSimulator)
                            -f ${PARAM_FILENAME}
                            -n ${PARAM_PROCS}
                TEST_ARGS ${TEST_ARGS}
-               CONFIGURATION extra)
+               CONFIGURATION ${PARAM_CONFIGURATION})
 endfunction()
 
+###########################################################################
```
akva2, line 43: this macro seems to be unused, and not really necessary. looks like remnants from earlier attempts.

compareECLFiles.cmake
```diff
@@ -277,13 +306,52 @@ opm_set_test_driver(${PROJECT_SOURCE_DIR}/tests/run-test.sh "")
 add_test_runSimulator(CASENAME norne
                       FILENAME NORNE_ATW2013
                       SIMULATOR flow
-                      PROCS 1)
+                      PROCS 1
+											CONFIGURATION extra)
```
akva2, line 310: watch them tabs.

compareECLFiles.cmake
```diff
@@ -277,13 +306,52 @@ opm_set_test_driver(${PROJECT_SOURCE_DIR}/tests/run-test.sh "")
...
 
 add_test_runSimulator(CASENAME norne_parallel
                       FILENAME NORNE_ATW2013
                       SIMULATOR flow
                       DIR norne
-                      PROCS 4)
+                      PROCS 4
+											CONFIGURATION extra)
```
akva2, line 317: watch them tabs.

compareECLFiles.cmake
```diff
@@ -37,9 +37,38 @@ function(add_test_runSimulator)
                            -f ${PARAM_FILENAME}
                            -n ${PARAM_PROCS}
                TEST_ARGS ${TEST_ARGS}
-               CONFIGURATION extra)
+               CONFIGURATION ${PARAM_CONFIGURATION})
```
akva2, line 40: since we are using this for 'normal' tests now, please add
```
set_test_properties(runsimulator/${PARAM_CASENAME} PROPERTIES PROCESSORS ${PARAM_PROCS})
```

tests/test_tuning_XXXMBE.cpp
vkip: except ignorance, no.. 😳

tests/test_tuning_XXXMBE.cpp
vkip: Ah, trying again.

tests/test_tuning_XXXMBE.cpp
vkip: Good point, switching to at.

compareECLFiles.cmake
vkip: Ah, you're right 😳  - deleting.

vkip:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  I'll merge this into master now.

PR #4623, opened 2023-05-04
Add disgasw to gas water saltprec vapwat simulator

totto82:
Any reasons for not allowing dissolved gas in water here? Or should I rather add a separate simulator for this?

### DIFF
--- a/flow/flow_ebos_gaswater_saltprec_vapwat.cpp
+++ b/flow/flow_ebos_gaswater_saltprec_vapwat.cpp
@@ -47,6 +47,11 @@ struct EnableEvaporation<TypeTag, TTag::EclFlowGasWaterSaltprecVapwatProblem> {
     static constexpr bool value = true;
 };
 
+template<class TypeTag>
+struct EnableDisgasInWater<TypeTag, TTag::EclFlowGasWaterSaltprecVapwatProblem> {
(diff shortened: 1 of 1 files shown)


goncalvesmachadoc:
Do adding gas to water means that we have to introduce a new PVT table for water to account for RSW?

goncalvesmachadoc:
We didn't allow for it because this simulator was intended to model gas fields. Although water can completely dry-up in some places, the dissolution of methane in water is assumed neglectable.

totto82:
I want to simulate saltprec + vapwat + disgasw in a gas/brinesystem using CO2STORE. Currently we don't support adding table input for RSW so it will only be useful in combination with CO2STORE. I just though that it didn't make any difference to add disgasw to this executable. I think the overhead in terms of performance should be negligible and I think in general that we should try to add as few compiled versions as possible.

plgbrts:
Does this change mean that a user, to setup a gas-water case with evaporation and precipitation (and no gas dissolution) will need nevertheless to specify DISGAS? (Thus that the testcase GASWATER_VAPWAT_PRECSALT.DATA as is would break?)

plgbrts:
In that case it looks good to me

totto82:
jenkins build this please

PR #4624, opened 2023-05-04
Simplify interface for calculation perf rates

akva2:
Minor refactor, introduces a struct and passes an instance of this instead of several separate variables.

### DIFF
--- a/opm/simulators/wells/PerforationData.hpp
+++ b/opm/simulators/wells/PerforationData.hpp
@@ -35,6 +35,14 @@ struct PerforationData
     std::size_t ecl_index;
 };
 
+struct PerforationRates
+{
+    double dis_gas = 0.0;
+    double dis_gas_in_water = 0.0;
+    double vap_oil = 0.0;
+    double vap_wat = 0.0;
+};
+
 } // namespace Opm
 
 #endif // OPM_PERFORATIONDATA_HEADER_INCLUDED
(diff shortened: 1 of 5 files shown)


akva2:
jenkins build this please

opm/simulators/wells/PerforationData.hpp
```diff
@@ -35,6 +35,14 @@ struct PerforationData
     std::size_t ecl_index;
 };
 
+struct PerforationRates
+{
+    double dis_gas{};
```
atgeirr, line 40: Will this zero-initialize? Even if it does, would it not be preferable to write it as `member = 0.0` to make it clearer?

atgeirr (approved):
Looks fine, only one question about initializing.

opm/simulators/wells/PerforationData.hpp
akva2: it does zero initialize but sure, i can use more familiar syntax.

akva2:
jenkins build this please

opm/simulators/wells/PerforationData.hpp
bska: As @akva2 correctly states, [list initialisation](https://en.cppreference.com/w/cpp/language/list_initialization) will perform [value initialisation](https://en.cppreference.com/w/cpp/language/value_initialization) in this case which, for built-in arithmetic types, will assign the type's zero value.

bska (approved):
I like it and I'll merge into master.

PR #4625, opened 2023-05-04
changed: store rates in an array in SingleWellState

akva2:
introduce an enum for indexing into the array. this again allows us to coalesce 4 parallel reductions into one.

another micro optimization.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1560,10 +1560,7 @@ namespace Opm
         this->linSys_.clear();
 
         auto& ws = well_state.well(this->index_of_well_);
-        ws.dissolved_gas_rate = 0;
-        ws.dissolved_gas_rate_in_water = 0;
-        ws.vaporized_oil_rate = 0;
-        ws.vaporized_wat_rate = 0;
+        ws.phase_mixing_rates.fill(0.0);
 
         // for the black oil cases, there will be four equations,
         // the first three of them are the mass balance equations, the last one is the pressure equations.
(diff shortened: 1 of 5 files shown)


akva2:
jenkins build this please

opm/simulators/wells/SingleWellState.hpp
```diff
@@ -85,10 +82,15 @@ class SingleWellState {
     double bhp{0};
     double thp{0};
     double temperature{0};
-    double dissolved_gas_rate{0};
-    double dissolved_gas_rate_in_water{0};
-    double vaporized_oil_rate{0};
-    double vaporized_wat_rate{0};
+
+    std::array<double,4> rates{};
```
bska, line 86: I'm not sure `rates` is the best name for this data member.  There are multiple other quantities in `SingleWellState` that could potentially be named `rates` as well.  As written, this is a more specialised collection that exclusively holds dissolved and vaporised flow rates and I think the name should reflect that more restricted purpose.  That said, I don't have a good suggestion for another name.  Something like `dissolved_rates` would be close, but might suggest that the vaporised components are excluded.  `Phase_mixing_rates` or just `mixing_rates` might be an alternative, but we've not used that notion anywhere else.

bska (commented):
I think this is good save for some naming concerns.  I'm going to solicit opinions from @atgeirr, @GitPaean, and @totto82 on that matter.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1545,10 +1545,7 @@ namespace Opm
         this->linSys_.clear();
 
         auto& ws = well_state.well(this->index_of_well_);
-        ws.dissolved_gas_rate = 0;
-        ws.dissolved_gas_rate_in_water = 0;
-        ws.vaporized_oil_rate = 0;
-        ws.vaporized_wat_rate = 0;
+        ws.rates = {};
```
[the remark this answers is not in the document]
akva2: done.

opm/simulators/wells/SingleWellState.hpp
GitPaean: This is also my only concern. Maybe `phase_mixing_rates`  will be the most descriptive. Or `mixing_rates` with some comments around can also work to me.

akva2: let's just make it phase_mixing_rates then.

akva2:
jenkins build this please

PR #4626, opened 2023-05-04
Simplify WellconnectionPressures properties signatures

akva2:
Introduce a struct to group parameters, and use this to simplify method signatures.


akva2:
jenkins build this please

totto82:
Thanks for the clean-up. From my point of view this can be merged as is. I will let it linger a bit for other developers to be able to comment, but if no further comments this can be merged later today.

PR #4627, opened 2023-05-04
Import opm-models

akva2:
It seems it's never a good time for this, but I'll open this PR for testing purposes so I can nip any problems out up front for when it is.


akva2:
jenkins build this please

akva2:
jenkins build this please

bska:
Tangential note: Do we know how much of `doc/handbook` we (i.e., the OPM project) have written?  I would assume that at least some portion of it is from the Dumux project, but I have no idea how much.

blattms:
There might not be a better time than now... (I am aware this needs to be discussed)

Actually, that might open the door to moving the material law managers from opm-common to opm-simulators which might make supporting adaptivity a tad easier.

aritorto:
I;m currently working on supporting LGRs on EclThermalLawManager, namely, getting the field properties of type int and double for a leaf grid view with LGRs, and it would be more convenient if the files were located in opm-simulators.

akva2:
September is upon us. As previously agree I have updated this PR now. I hope this can be expedited quite fast so I don't have to update yet again.

akva2:

bska (commented):
I haven't reviewed the new files, other than `modelTests.cmake`, since I assume those are effectively copies of the corresponding files in the OPM-Models repository.  We could probably drop the `doc` directory at some point as well.

The integration changes look good to me and the CI system confirms the build is green. I do however have a couple of questions about the integration itself

1. What happens to [the open PRs in OPM-Models](https://github.com/OPM/opm-models/pulls)?
2. Do we not need the workaround for older compilers that don't have the floating-point overloads of `std::from_chars()`&ndash;i.e., PRs OPM/opm-models#926 and OPM/opm-models#927?

akva2:
1) They will have to be moved. There will always be PRs and Sep 1 was agreed upon back in April as the cut-off date to do this.
2) Good catch. That is definitely an oversight and I have remedied.

akva2:
jenkins build this please

akva2:
absolutely available to help, but it should simply be ```git format-patch``` and ```git am``` assuming it doesn't touch buildsystem etc.

bska (approved):
> absolutely available to help, but it should simply be git format-patch and git am assuming it doesn't touch buildsystem etc.

Okay, very good.  In that case I'll merge this to start the process of incorporating OPM-Models into OPM-Simulators.

PR #4630, opened 2023-05-08
changed: unify getMobility(Eval|Scalar)

akva2:
To avoid manual duplication of code.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell.hpp
+++ b/opm/simulators/wells/MultisegmentWell.hpp
@@ -238,7 +238,8 @@ namespace Opm
         template<class Value>
         void getMobility(const Simulator& ebosSimulator,
                          const int perf,
-                         std::vector<Value>& mob) const;
+                         std::vector<Value>& mob,
+                         DeferredLogger& deferred_logger) const;
 
         void computeWellRatesAtBhpLimit(const Simulator& ebosSimulator,
                                         std::vector<double>& well_flux,
--- a/opm/simulators/wells/MultisegmentWell.hpp
+++ b/opm/simulators/wells/MultisegmentWell.hpp
@@ -235,14 +235,10 @@ namespace Opm
                                            DeferredLogger& deferred_logger);
 
         // get the mobility for specific perforation
-        void getMobilityEval(const Simulator& ebosSimulator,
-                             const int perf,
-                             std::vector<EvalWell>& mob) const;
-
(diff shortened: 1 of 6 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
@totto82 @GitPaean  why the difference in msw for relperms, ref the relpermArray() lambda ? is the complication there for a reason and if so should we do the same for stw?

GitPaean:
>  @totto82 @GitPaean why the difference in msw for relperms, ref the relpermArray() lambda ? is the complication there for a reason and if so should we do the same for stw?

I think the same can be used for stw, it might be able to slightly faster to use Scalar when we only need the Scalar values.

totto82:
Cant getMobility be further unified between MSW and STW? The difference is related to missing solvent ++ in MSW, but shouldn't that be taken care of by `if constexpr (has_solvent)` Or did I miss something?

totto82:
> that was my plan yes, in a follow-up. but if you want it here, i'll go for it now.

Do what you think best. Thanks for the cleanup! From my point of view this can be merged.

akva2:
jenkins build this please

akva2:
any objections @bska @atgeirr @GitPaean  ?

GitPaean:
I did not check all the details, but I understand this PR is an effort to refactor and cleaning up of the code and does not change running results.  

I think it is good and have no objections to get it in.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this opm-tets=963 please

akva2:
jenkins build this opm-tests=963 please

akva2:
jenkins build this opm-tests=963 please

akva2:
jenkins build this opm-tests=963 please

GitPaean:
I think my purpose was to update_data, but somehow I forgot it. But I saw it is triggered in opm-tests now.

GitPaean:
jenkins build this opm-tests=963 update_data please

akva2:
it's not a good idea to use this PR as it changes code that might affect those values.

akva2:
jenkins build this please

GitPaean:
with the merging of https://github.com/OPM/opm-simulators/pull/4655 , there is a conflicts now. Please help to fix and update it.

akva2:
jenkins build this please

PR #4629, opened 2023-05-05
Refactor perforation rate calculations

akva2:
Split into separate methods to increase readability of calculatePerfRate

Needs https://github.com/OPM/opm-common/pull/3507

PR #4632, opened 2023-05-08
Move perforation rate calculations to separate class

akva2:
Sits on top of https://github.com/OPM/opm-simulators/pull/4629


akva2:
jenkins build this opm-common=3507 please

PR #4633, opened 2023-05-08
Path to multiGPU: Basic utilities for cuISTL

kjetilly:
We are creating a new linear solver GPU path compatible with the DUNE ISTL library. This first PR provides basic utilities for handling CUDA, CUSPARSE and CUBLAS resources, as well as safe call macros to ensure safe programming (checks for return codes and throws an exception with a meaningful error message).

[This is related to the issue 3932](https://github.com/OPM/opm-simulators/issues/3932)

[This PR partially replaces PR 4137](https://github.com/OPM/opm-simulators/pull/4137)

This PR is a prerequisite for [PR 4634](https://github.com/OPM/opm-simulators/pull/4634), [PR 4635](https://github.com/OPM/opm-simulators/pull/4635) and [PR4636](https://github.com/OPM/opm-simulators/pull/4636)

PR #4634, opened 2023-05-08
Path to multigpu: Cuistl vector and matrix classes

kjetilly:
We are creating a new linear solver GPU path compatible with the DUNE ISTL library. This first PR provides the vector and matrix wrapper classes for CuBlas and CuSparse respectively. 

Due to the fact that cublas and cusparse only work on signed integers, while Dune and friends operate on `size_t`, we have added some safe conversion routines to make sure we are always within bounds as well. These can be replaced or greatly simplified once C++20's safe cmp functions become available. 

[This is related to the issue 3932](https://github.com/OPM/opm-simulators/issues/3932)

[This PR partially replaces PR 4137](https://github.com/OPM/opm-simulators/pull/4137)

This PR builds upon [PR 4633](https://github.com/OPM/opm-simulators/pull/4633), and is a prerequisite for [PR 4635](https://github.com/OPM/opm-simulators/pull/4635) and [PR4636](https://github.com/OPM/opm-simulators/pull/4636)

### DIFF
--- a/opm/simulators/linalg/cuistl/CuSparseMatrix.cpp
+++ b/opm/simulators/linalg/cuistl/CuSparseMatrix.cpp
@@ -286,13 +286,13 @@ CuSparseMatrix<T>::usmv(T alpha, const CuVector<T>& x, CuVector<T>& y) const
 }
 
 template <class T>
-template <class M>
+template <class VectorType>
 void
-CuSparseMatrix<T>::assertSameSize(const M& x) const
+CuSparseMatrix<T>::assertSameSize(const VectorType& x) const
 {
     if (x.dim() != blockSize() * N()) {
         OPM_THROW(std::invalid_argument,
-                  fmt::format("Size mismatch. Input vector has {} elements, while we have {} elements.",
+                  fmt::format("Size mismatch. Input vector has {} elements, while we have {} rows.",
                               x.dim(),
                               blockSize() * N()));
     }
--- a/opm/simulators/linalg/cuistl/CuSparseMatrix.hpp
+++ b/opm/simulators/linalg/cuistl/CuSparseMatrix.hpp
@@ -295,8 +295,8 @@ private:
     detail::CuSparseMatrixDescriptionPtr m_matrixDescription;
     detail::CuSparseHandle& m_cusparseHandle;
 
-    template <class M>
-    void assertSameSize(const M& otherMatrix) const;
+    template <class VectorType>
+    void assertSameSize(const VectorType& vector) const;
 };
 } // namespace Opm::cuistl
 #endif


bska:
jenkins build this please

opm/simulators/linalg/cuistl/detail/cublas_wrapper.hpp
```diff
@@ -0,0 +1,168 @@
...
+}
+
+inline cublasStatus_t
+cublasScal([[maybe_unused]] cublasHandle_t handle,
+           [[maybe_unused]] int n,
+           [[maybe_unused]] const int* alpha, /* host or device pointer */
+           [[maybe_unused]] int* x,
+           [[maybe_unused]] int incx)
+{
+    OPM_THROW(std::runtime_error, "cublasScal multiplication for integer vectors is not implemented yet.");
```
atgeirr, line 70: Is there any value in including these integer overloads? The pointers don't convert in any case so there is no risk of accidentally passing arrays other than float or double.

opm/simulators/linalg/cuistl/detail/vector_operations.hpp
```diff
@@ -0,0 +1,58 @@
...
+ * @brief innerProductAtIndices computes the inner product between deviceA[indices] and deviceB[indices]
+ * @param deviceA data A
+ * @param deviceB data B
+ * @param buffer a buffer with number of elements equal to numberOfElements
+ * @param numberOfElements number of indices
+ * @param indices the indices to compute the inner product over
+ * @return the result of the inner product
+ *
+ * @note This is equivalent to projecting the vectors to the indices contained in indices, then doing the inner product
+ * of those indices.
```
atgeirr, line 53: indices->vectors

opm/simulators/linalg/cuistl/detail/vector_operations.hpp
```diff
@@ -0,0 +1,58 @@
...
+void setVectorValue(T* deviceData, size_t numberOfElements, const T& value);
+
+/**
+ * @brief setZeroAtIndexSet sets deviceData to zero in the indices of contained in indices
+ * @param deviceData the data to operate on
+ * @param numberOfElements number of indices
+ * @param indices the indices to use
+ */
+template <class T>
+void setZeroAtIndexSet(T* deviceData, size_t numberOfElements, const int* indices);
```
atgeirr, line 41: It seem to me that the indices also should be on the device, is that correct? Also, the buffer in innerProductAtIndices().

opm/simulators/linalg/cuistl/detail/vector_operations.cu
```diff
@@ -0,0 +1,112 @@
...
+        // TODO: [perf] Is it faster to just use a mask? Probably does not matter either way
+        //              This is hopefully not where we will spend most of our time.
+        if (globalIndex < numberOfElements) {
+            buffer[globalIndex] = a[indices[globalIndex]] * b[indices[globalIndex]];
+        }
+    }
+
+    constexpr inline size_t getThreads([[maybe_unused]] size_t numberOfElements)
+    {
+        return 1024;
```
atgeirr, line 63: I assume this is a value that should work well most of the time, but is there a way to ask the hardware to tell us the optimal size?

atgeirr (commented):
I have now reviewed the low-level parts of this, not yet looked at the vector, matrix and resource classes.

CMakeLists_files.cmake
```diff
@@ -155,6 +159,14 @@ if(CUDA_FOUND)
   list (APPEND PUBLIC_HEADER_FILES opm/simulators/linalg/cuistl/detail/cuda_check_last_error.hpp)
   list (APPEND PUBLIC_HEADER_FILES opm/simulators/linalg/cuistl/detail/CuBlasHandle.hpp)
   list (APPEND PUBLIC_HEADER_FILES opm/simulators/linalg/cuistl/detail/CuSparseHandle.hpp)
+  list (APPEND PUBLIC_HEADER_FILES opm/simulators/linalg/cuistl/CuVector.hpp)
+  list (APPEND PUBLIC_HEADER_FILES opm/simulators/linalg/cuistl/CuSparseMatrix.hpp)
+  list (APPEND PUBLIC_HEADER_FILES opm/simulators/linalg/cuistl/detail/CuMatrixDescription.hpp)
+  list (APPEND PUBLIC_HEADER_FILES opm/simulators/linalg/cuistl/detail/CuSparseResource.hpp)
+  list (APPEND PUBLIC_HEADER_FILES opm/simulators/linalg/cuistl/detail/CuSparseResource_impl.hpp)
+  list (APPEND PUBLIC_HEADER_FILES opm/simulators/linalg/cuistl/detail/safe_conversion.hpp)
```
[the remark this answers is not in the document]
kjetilly: They can be, but it is not a part of the interface we expose (they are burried in the `detail` namespace), and as far as I can tell, they are actually not included in any public header files, hence would not be needed for someone using OPM as a library. The files added here are all included in public header files. 

It's up to the convention of the OPM project, though. Technically there is no reason for them not to be included, subjectively you can say that they are not part of any exposed interface, hence do not need to be included.

opm/simulators/linalg/cuistl/detail/cublas_wrapper.hpp
kjetilly: We instantiate the `CuVector` with an `int` template argument, so there is a possibility of getting to this part of the code. So for compilation technical reasons, I think this needs to be there.

I think long term `int` multiplication can have a value for some algorithms, hence I am leaving it open now.

CMakeLists_files.cmake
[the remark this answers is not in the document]
atgeirr: I was told that including them improves the experience of using QCreator, that is good enough reason for me.

opm/simulators/linalg/cuistl/detail/cublas_wrapper.hpp
atgeirr: Do the corresponding integer algorithms exist in cublas?

kjetilly: No, I think it only supports vector spaces and not modules, unfortunately. We would have to write them ourselves, but the point is that once we write this (basic arithmetic operation kernels) ourselves, we would get them for whatever raw type we would want (since they can be templated as done in `vector_operations.cu`), hence the cublas dependency could be removed entirely.

I anyway found the current temporary solution easier/cleaner than:

a) overloading every arithmetic method in `CuVector<int>` to throw an error (I don't think `static_assert` would be possible in this case, since we need to instantiate the vector for `int`). There are more arithmetic operations in `CuVector<T>` than there are cublas-routines we have to overload.

b) adding a bunch of tests to the methods of `CuVector<T>` of the form

```C++
if (std::is_same_v<T, int>) { throw_some_exception(); }
```

A long-term solution I think is a mixture of:

1) Implement the corresponding algorithms for `int` (or rather `template<class T>`) akin to what is done already in `vector_operations.cu` if the need for these algorithms arises. If the hand-written algorithms prove to have equal or better performance for our use cases than CuBlas (maybe there is tuning that we can do, but I would not think so) then we can consider also using these algorithms for `float` and `double`. 

2) Add a new `CuBuffer<T>` class whose only job is to hold data (so `CuVector<T>` without the arithmetic expressions and conversions to `Dune::BlockVector`), and let `CuVector<T>` use this through composition). This would provide a data-holder (as in a GPU version of `std::vector`) on the GPU without any arithmetic operations (which is how `CuVector<int>` is used currently in the code), it would also facilitate having a data holder for raw types for which there are no suitable arithmetic operations (a `union` for instance). 

With such a buffer, the interfaces to the functions in `vector_operations.hpp` could also be simplified to only take a `CuBuffer<T>` (and one could add a corresponding `CuBufferView<T>` that would be CUDA kernel compatible (see for instance what I did in Alsvinn with [View.hpp](https://github.com/alsvinn/alsvinn/blob/master/alsfvm/include/alsfvm/memory/View.hpp), instead of a raw memory pointer and length.

Now, I think 2) should be done down the line anyway, but I also think that for the current state of the code, it would only obfuscate and make things more complicated (more classes to reason about) than it would clarify.

atgeirr: I agree with both: That 2) is a good solution, and also not required immediately.

opm/simulators/linalg/cuistl/CuSparseMatrix.hpp
```diff
@@ -0,0 +1,302 @@
...
+     * @brief fromMatrix creates a new matrix with the same block size and values as the given matrix
+     * @param matrix the matrix to copy from
+     * @param copyNonZeroElementsDirectly if true will do a memcpy from matrix[0][0][0][0], otherwise will build up the
+     * non-zero elements by looping over the matrix. Note that setting this to true will yield a performance increase,
+     * but might not always yield correct results depending on how the matrix matrix has been initialized. If unsure,
+     * leave it as false.
+     * @tparam MatrixType is assumed to be a Dune::BCRSMatrix compatible matrix.
+     */
+    template <class MatrixType>
+    static CuSparseMatrix<T> fromMatrix(const MatrixType& matrix, bool copyNonZeroElementsDirectly = false);
```
atgeirr, line 92: Why not make this a normal constructor instead of a static function?

opm/simulators/linalg/cuistl/CuSparseMatrix.hpp
```diff
@@ -0,0 +1,302 @@
...
+    void updateNonzeroValues(const MatrixType& matrix, bool copyNonZeroElementsDirectly = false);
+
+private:
+    CuVector<T> m_nonZeroElements;
+    CuVector<int> m_columnIndices;
+    CuVector<int> m_rowIndices;
+
+    // Notice that we store these three as int to make sure we are cusparse compatible.
+    //
+    // This gives the added benifit of checking the size constraints at construction of the matrix
```
atgeirr, line 289: Typo: benefit

opm/simulators/linalg/cuistl/CuSparseMatrix.cpp
```diff
@@ -0,0 +1,327 @@
...
+    : m_nonZeroElements(nonZeroElements, numberOfNonzeroBlocks * blockSize * blockSize)
+    , m_columnIndices(columnIndices, numberOfNonzeroBlocks)
+    , m_rowIndices(rowIndices, numberOfRows + 1)
+    , m_numberOfNonzeroBlocks(detail::to_int(numberOfNonzeroBlocks))
+    , m_numberOfRows(detail::to_int(numberOfRows))
+    , m_blockSize(detail::to_int(blockSize))
+    , m_matrixDescription(detail::createMatrixDescription())
+    , m_cusparseHandle(detail::CuSparseHandle::getInstance())
+{
+}
```
atgeirr, line 78: Maybe assert that m_rowIndices[numberOfRows] == numberOfNonzeroBlocks?

opm/simulators/linalg/cuistl/CuSparseMatrix.cpp
```diff
@@ -0,0 +1,327 @@
...
+                                                 blockSize(),
+                                                 x.data(),
+                                                 &beta,
+                                                 y.data()));
+}
+
+template <class T>
+template <class M>
+void
+CuSparseMatrix<T>::assertSameSize(const M& x) const
```
atgeirr, line 288: The typename M is potentially misleading here, it should be a V for vector instead perhaps?

opm/simulators/linalg/cuistl/CuSparseMatrix.cpp
```diff
@@ -0,0 +1,327 @@
...
+template <typename T>
+void
+CuSparseMatrix<T>::usmv(T alpha, const CuVector<T>& x, CuVector<T>& y) const
+{
+    assertSameSize(x);
+    assertSameSize(y);
+    if (blockSize() < 2) {
+        OPM_THROW(
+            std::invalid_argument,
+            "CuSparseMatrix<T>::usmv and CuSparseMatrix<T>::mv are only implemented for block sizes greater than 1.");
```
atgeirr, line 258: This may be needed for CPR for example, where we have a pressure matrix with 1x1 blocks, or scalars. I assume it is not a fundamental design and can be fixed when needed.

tests/cuistl/test_cusparsematrix.cpp
```diff
@@ -0,0 +1,181 @@
...
+    //
+    //     -2  1  0  0 ... 0  0
+    //      1 -2  1  0  0 ... 0
+    //      ....
+    //      0  0  0  ...1 -2  1
+    //      0  0  0  ...   1 -2
+
+    const int N = 5;
+    const int nonZeroes = N * 3 - 2;
+    using M = Dune::FieldMatrix<double, 1, 1>;
```
atgeirr, line 53: This seems to contradict a statement elsewhere that block sizes must be > 1. Seems to me that other statement is wrong, and we indeed support block size 1?

tests/cuistl/test_cusparsematrix.cpp
```diff
@@ -0,0 +1,181 @@
...
+    BOOST_CHECK_EQUAL(columnIndicesFromCUDA[0], 0);
+    BOOST_CHECK_EQUAL(columnIndicesFromCUDA[1], 1);
+    // TODO: Check rest
+}
+
+
+BOOST_AUTO_TEST_CASE(RandomSparsityMatrix)
+{
+    std::srand(0);
+    double nonzeroPercent = 0.2;
```
atgeirr, line 109: I think this is a fraction and not a percentage, since you draw from [0,1] not [0,100].

atgeirr (commented):
Very good! I have some comments though to be addressed before merging.

tests/cuistl/test_cusparsematrix.cpp
kjetilly: This only tests construction, where we don't have any limitations. Only arithmetic operations  (mv etc) require block size 1.

opm/simulators/linalg/cuistl/CuSparseMatrix.cpp
kjetilly: It would mean changing to a CSR and not BSR (block CSR) matrix. Can be done with some rewriting.

opm/simulators/linalg/cuistl/CuVector.hpp
```diff
@@ -0,0 +1,349 @@
...
+     * @brief returns the l2 norm of the vector
+     * @note this will call CuBlas in the background
+     * @note int is not supported
+     *
+     * @return the l2 norm
+     */
+    T two_norm() const;
+
+    //! Computes the dot product sum_i this[indexSet[i]] * other[indexSet[i]]
+    T dot(const CuVector<T>& other, const CuVector<int>& indexSet, CuVector<T>& buffer) const;
```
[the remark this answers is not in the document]
kjetilly: The algorithm needs a buffer to perform this task, but if the user does not supply it, we will allocate it temporarily. I will update the documentation to reflect this.

opm/simulators/linalg/cuistl/CuVector.hpp
```diff
@@ -0,0 +1,349 @@
...
+     * @note For now data.size() needs to be within the limits of int due to restrctions of CuBlas.
+     *
+     * @param data the vector to copy from
+     */
+    explicit CuVector(const std::vector<T>& data);
+
+    /**
+     * @brief operator= copies the content of the data vector to the memory of this vector.
+     *
+     * @note This requires the two vectors to be of the same size.
```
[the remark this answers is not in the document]
kjetilly: Sanity and peace of mind. When dealing with numerical vector, assigning one vector to a different size vector is in my experience usually the sign of a bug, and how say Julia's `somevector[:] = ` would behave.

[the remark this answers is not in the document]
atgeirr: Ok.

opm/simulators/linalg/cuistl/CuVector.hpp
```diff
@@ -0,0 +1,349 @@
...
+        const auto dataPointer = static_cast<const T*>(&(vector[0][0]));
+        copyFromHost(dataPointer, m_numberOfElements);
+    }
+
+    /**
+     * @brief copyToHost copies data to a Dune::BlockVector
+     * @param vector the vector to copy to
+     *
+     * @note This does synchronous transfer.
+     * @note This assumes that the size of this vector is equal to the dim of the input vector.
```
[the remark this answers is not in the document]
kjetilly: Added a todo for future optimization.

opm/simulators/linalg/cuistl/CuVector.hpp
```diff
@@ -0,0 +1,349 @@
...
+    T dot(const CuVector<T>& other, const CuVector<int>& indexSet) const;
+    //! Computes the norm sqrt(sum_i this[indexSet[i]] * this[indexSet[i]])
+    T two_norm(const CuVector<int>& indexSet) const;
+
+
+    /**
+     * @brief dim returns the dimension (number of T elements) in the vector
+     * @return number of elements
+     */
+    size_type dim() const;
```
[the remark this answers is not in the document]
kjetilly: Just to have the same interface as the dune vector.

opm/simulators/linalg/cuistl/CuVector.cpp
```diff
@@ -0,0 +1,285 @@
...
+                              numberOfElements));
+    }
+    OPM_CUDA_SAFE_CALL(cudaMemcpy(data(), dataPointer, numberOfElements * sizeof(T), cudaMemcpyHostToDevice));
+}
+
+template <class T>
+void
+CuVector<T>::copyToHost(T* dataPointer, size_t numberOfElements) const
+{
+    OPM_CUDA_SAFE_CALL(cudaMemcpy(dataPointer, data(), numberOfElements * sizeof(T), cudaMemcpyDeviceToHost));
```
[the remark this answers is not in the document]
kjetilly: Fixed.

tests/cuistl/test_cuvector.cpp
```diff
@@ -0,0 +1,295 @@
...
+    vectorOnGPU = 10.0;
+    vectorOnGPU.copyToHost(data.data(), data.size());
+
+    for (double x : data) {
+        BOOST_CHECK_EQUAL(10.0, x);
+    }
+}
+
+
+BOOST_AUTO_TEST_CASE(CopyConstructor)
```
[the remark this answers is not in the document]
kjetilly: Fixed

tests/cuistl/test_cuvector.cpp
```diff
@@ -0,0 +1,295 @@
...
+        auto aOutputScalar = aGPU.asStdVector();
+        for (size_t i = 0; i < N; ++i) {
+            BOOST_CHECK_EQUAL(aOutputScalar[i], scalar * a[i]);
+        }
+
+        aGPU = GVector(a);
+        aGPU.axpy(scalar, bGPU);
+        auto aOutputSaxypy = aGPU.asStdVector();
+        for (size_t i = 0; i < N; ++i) {
+            BOOST_CHECK_CLOSE(aOutputSaxypy[i], a[i] + scalar * b[i], 1e-7);
```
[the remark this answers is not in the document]
kjetilly: Made it stricter.

opm/simulators/linalg/cuistl/CuVector.hpp
```diff
@@ -0,0 +1,349 @@
...
+                                  vector.N(),
+                                  vector.dim()));
+        }
+        const auto dataPointer = static_cast<const T*>(&(vector[0][0]));
+        copyFromHost(dataPointer, m_numberOfElements);
+    }
+
+    /**
+     * @brief copyToHost copies data to a Dune::BlockVector
+     * @param vector the vector to copy to
```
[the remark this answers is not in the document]
kjetilly: Renamed.

opm/simulators/linalg/cuistl/CuSparseMatrix.hpp
```diff
@@ -0,0 +1,302 @@
...
+    CuSparseMatrix& operator=(const CuSparseMatrix&) = delete;
+
+    virtual ~CuSparseMatrix();
+
+    /**
+     * @brief fromMatrix creates a new matrix with the same block size and values as the given matrix
+     * @param matrix the matrix to copy from
+     * @param copyNonZeroElementsDirectly if true will do a memcpy from matrix[0][0][0][0], otherwise will build up the
+     * non-zero elements by looping over the matrix. Note that setting this to true will yield a performance increase,
+     * but might not always yield correct results depending on how the matrix matrix has been initialized. If unsure,
```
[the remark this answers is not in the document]
kjetilly: Fixed.

opm/simulators/linalg/cuistl/CuSparseMatrix.hpp
kjetilly: I would postpone refactoring when we get the cubuffer.

kjetilly:
jenkins build this please

atgeirr:
All green, merging!

PR #4635, opened 2023-05-08
Path to multigpu: Cuistl preconditioners

kjetilly:
We are creating a new linear solver GPU path compatible with the DUNE ISTL library. This first PR provides a wrapper to the CuSparse implementation of the sequential ILU0 algorithm. 

This new preconditioner is then exposed in the PreconditionerFactory, and can be run by specifying a `json` file of the form

```json
{
    "preconditioner" : {"type":"CUILU0"},
    "solver" : "bicgstab"
}
```

[This is related to the issue 3932](https://github.com/OPM/opm-simulators/issues/3932)

[This PR partially replaces PR 4137](https://github.com/OPM/opm-simulators/pull/4137)

This PR builds upon [PR 4634](https://github.com/OPM/opm-simulators/pull/4634), and is a prerequisite for [PR4636](https://github.com/OPM/opm-simulators/pull/4636)

### DIFF
--- a/opm/simulators/linalg/cuistl/CuSeqILU0.cpp
+++ b/opm/simulators/linalg/cuistl/CuSeqILU0.cpp
@@ -149,8 +149,6 @@ void
 CuSeqILU0<M, X, Y, l>::update()
 {
     m_LU.updateNonzeroValues(detail::makeMatrixWithNonzeroDiagonal(m_underlyingMatrix));
-    // updateILUConfiguration();
-
     createILU();
 }
 
--- a/opm/simulators/linalg/cuistl/CuSeqILU0.cpp
+++ b/opm/simulators/linalg/cuistl/CuSeqILU0.cpp
@@ -1,0 +1,364 @@
+/*
+  Copyright 2022-2023 SINTEF AS
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#include <cuda.h>
+#include <cuda_runtime.h>
+#include <cusparse.h>
+#include <dune/common/fmatrix.hh>
+#include <dune/common/fvector.hh>
+#include <dune/istl/bcrsmatrix.hh>
+#include <dune/istl/bvector.hh>
+#include <fmt/core.h>
+#include <opm/common/ErrorMacros.hpp>
+#include <opm/simulators/linalg/cuistl/CuSeqILU0.hpp>
+#include <opm/simulators/linalg/cuistl/detail/cusparse_constants.hpp>
+#include <opm/simulators/linalg/cuistl/detail/cusparse_safe_call.hpp>
+#include <opm/simulators/linalg/cuistl/detail/cusparse_wrapper.hpp>
+#include <opm/simulators/linalg/cuistl/detail/fix_zero_diagonal.hpp>
+#include <opm/simulators/linalg/cuistl/detail/safe_conversion.hpp>
+#include <opm/simulators/linalg/matrixblock.hh>
+
+// This file is based on the guide at https://docs.nvidia.com/cuda/cusparse/index.html#csrilu02_solve ,
+// it highly recommended to read that before proceeding.
+
+
+namespace Opm::cuistl
+{
+
+template <class M, class X, class Y, int l>
+CuSeqILU0<M, X, Y, l>::CuSeqILU0(const M& A, field_type w)
+    : m_underlyingMatrix(A)
+    , m_w(w)
+    , m_LU(CuSparseMatrix<field_type>::fromMatrix(detail::makeMatrixWithNonzeroDiagonal(A)))
+    , m_temporaryStorage(m_LU.N() * m_LU.blockSize())
+    , m_descriptionL(detail::createLowerDiagonalDescription())
+    , m_descriptionU(detail::createUpperDiagonalDescription())
+    , m_cuSparseHandle(detail::CuSparseHandle::getInstance())
+{
+    // Some sanity check
+    OPM_ERROR_IF(A.N() != m_LU.N(),
+                 fmt::format("CuSparse matrix not same size as DUNE matrix. {} vs {}.", m_LU.N(), A.N()));
+    OPM_ERROR_IF(
+        A[0][0].N() != m_LU.blockSize(),
+        fmt::format("CuSparse matrix not same blocksize as DUNE matrix. {} vs {}.", m_LU.blockSize(), A[0][0].N()));
+    OPM_ERROR_IF(
+        A.N() * A[0][0].N() != m_LU.dim(),
+        fmt::format("CuSparse matrix not same dimension as DUNE matrix. {} vs {}.", m_LU.dim(), A.N() * A[0][0].N()));
+    OPM_ERROR_IF(A.nonzeroes() != m_LU.nonzeroes(),
+                 fmt::format("CuSparse matrix not same number of non zeroes as DUNE matrix. {} vs {}. ",
+                             m_LU.nonzeroes(),
+                             A.nonzeroes()));
+
+
+    updateILUConfiguration();
+}
+
+template <class M, class X, class Y, int l>
+void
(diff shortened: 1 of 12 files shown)


opm/simulators/linalg/cuistl/detail/preconditioner_should_call_post_pre.hpp
```diff
@@ -0,0 +1,64 @@
...
+        return PreconditionerType::shouldCallPre();
+    } else {
+        // If the preconditioner type does not have a way of signalling the need
+        // to call pre(), we should always call it. This is because it could be an
+        // old design.
+        return true;
+    }
+}
+
+//! @brief Tests (compile time) if the preconditioner type needs to call pre
```
atgeirr, line 46: pre -> post

It may also be clearer to state "... if the calling code needs to call the post() method after calling apply() on the given preconditioner type" (similar with pre() above).

opm/simulators/linalg/cuistl/PreconditionerConvertFieldTypeAdapter.hpp
```diff
@@ -0,0 +1,240 @@
...
+//!     auto floatILU0 = std::make_shared<FloatILU0>(convertedMatrix, 0, 1.0, 0);
+//!
+//!     doubleToFloatConverter.setUnderlyingPreconditioner(floatILU0);
+//!
+//!     // This will convert x and y to float, thenn call floatILU0.apply on the converted arguments
+//!     doubleToFloatConverter.apply(x, y);
+//! }
+//!
+//! \endcode
+template <class CudaPreconditionerType, class M, class X, class Y, int l = 1>
```
atgeirr, line 84: The doc string above and the test program indicate that this class can be applied to any preconditioner. Then the template argument should be BasePreconditionerType or something like that, not using Cuda in the name. Also, it should be moved out of the cuistl subdir.

opm/simulators/linalg/cuistl/PreconditionerConvertFieldTypeAdapter.hpp
```diff
@@ -0,0 +1,240 @@
...
+//!     // Note that we do not need to make a new instance for every invocation, this
+//!     // is just done for this example
+//!     auto doubleToFloatConverter = DoubleToFloatConverter(matrix);
+//!     const auto& convertedMatrix = doubleToFloatConverter.getConvertedMatrix();
+//!
+//!     auto floatILU0 = std::make_shared<FloatILU0>(convertedMatrix, 0, 1.0, 0);
+//!
+//!     doubleToFloatConverter.setUnderlyingPreconditioner(floatILU0);
+//!
+//!     // This will convert x and y to float, thenn call floatILU0.apply on the converted arguments
```
atgeirr, line 79: thenn->then

opm/simulators/linalg/cuistl/PreconditionerConvertFieldTypeAdapter.hpp
```diff
@@ -0,0 +1,240 @@
...
+    using matrix_type_to =
+        typename Dune::BCRSMatrix<Dune::FieldMatrix<field_type_to, block_type::dimension, block_type::dimension>>;
+
+    //! \brief Constructor.
+    //!
+    //! \param A The matrix to operate on.
+    //!
+    //! \note After the PreconditionerConvertFieldTypeAdapter you can get the converted matrix
+    //! by calling getConvertedMatrix(), which in turn can be used to create the underlying preconditioner.
+    //! Once the underlying precondtioner has been called, this must be supplied to setUnderlyingPreconditioner.
```
atgeirr, line 119: precondtioner->preconditioner

opm/simulators/linalg/cuistl/PreconditionerConvertFieldTypeAdapter.hpp
```diff
@@ -0,0 +1,240 @@
...
+    explicit PreconditionerConvertFieldTypeAdapter(const M& matrix)
+        : m_matrix(matrix)
+        , m_convertedMatrix(createConvertedMatrix())
+    {
+    }
+
+    //! \brief Not used at the moment
+    virtual void pre([[maybe_unused]] X& x, [[maybe_unused]] Y& b) override
```
atgeirr, line 127: Should this not simply call the underlying preconditioner's pre() function (after the required conversions)? To avoid calling when not needed it should implement shouldCallPre() to forward the underlying preconditioner's behaviour. As is written here, I think this prevents me from testing arbitrary CPU preconditioners with float, which would be useful for experiments.

Similar for post(). Also the same applies to the PreconditionerAdapter class.

opm/simulators/linalg/cuistl/CuSeqILU0.hpp
```diff
@@ -0,0 +1,134 @@
...
+#include <opm/simulators/linalg/cuistl/CuSparseMatrix.hpp>
+#include <opm/simulators/linalg/cuistl/detail/CuMatrixDescription.hpp>
+#include <opm/simulators/linalg/cuistl/detail/CuSparseHandle.hpp>
+#include <opm/simulators/linalg/cuistl/detail/CuSparseResource.hpp>
+
+
+
+namespace Opm::cuistl
+{
+//! \brief Sequential ILU0 preconditioner on the GPU through the CuSparse library.
```
atgeirr, line 33: Sequential in the sense of not taking MPI into account I assume? I.e. similar to the SeqILU0 of dune-istl?

opm/simulators/linalg/cuistl/CuSeqILU0.hpp
```diff
@@ -0,0 +1,134 @@
...
+//!
+//! This implementation calls the CuSparse functions, which in turn essentially
+//! does a level decomposition to get some parallelism.
+//!
+//! \note This is not expected to be a fast preconditioner.
+//!
+//! \tparam M The matrix type to operate on
+//! \tparam X Type of the update
+//! \tparam Y Type of the defect
+//! \tparam l Ignored. Just there to have the same number of template arguments
```
atgeirr, line 43: That seems odd! Why was this needed?

opm/simulators/linalg/cuistl/CuSeqILU0.hpp
```diff
@@ -0,0 +1,134 @@
...
+    detail::CuSparseMatrixDescriptionPtr m_descriptionL;
+    detail::CuSparseMatrixDescriptionPtr m_descriptionU;
+    detail::CuSparseResource<bsrsv2Info_t> m_infoL;
+    detail::CuSparseResource<bsrsv2Info_t> m_infoU;
+    detail::CuSparseResource<bsrilu02Info_t> m_infoM;
+
+    std::unique_ptr<CuVector<field_type>> m_buffer;
+    detail::CuSparseHandle& m_cuSparseHandle;
+
+    bool m_analyzisDone = false;
```
atgeirr, line 123: Typo: z->s

opm/simulators/linalg/cuistl/PreconditionerConvertFieldTypeAdapter.hpp
kjetilly: It's  almost limited to cuistl, since  most other preconditioners use some functions that are not instantiated for float (either in OPM or Dune). It seemed non-trivial to fix. Therefore, I would I recommend keeping this as is, such that one does not expects this to work outside of cuistl.

opm/simulators/linalg/cuistl/PreconditionerConvertFieldTypeAdapter.hpp
kjetilly: It was to avoid adding more conversion code when the function is not at all used by ILU0, for which it was tested. I suggest adding this when the need arise.

opm/simulators/linalg/cuistl/CuSeqILU0.hpp
kjetilly: And Sequential in the sense that it should (up to rounding) give the same result as SeqILU0 on the CPU (ie. it does not approximation of the sequential algorithm)

kjetilly:
jenkins build this please.

opm/simulators/linalg/cuistl/CuSeqILU0.cpp
```diff
@@ -0,0 +1,362 @@
...
+                                                             blockSize,
+                                                             m_infoM.get(),
+                                                             CUSPARSE_SOLVE_POLICY_USE_LEVEL,
+                                                             m_buffer->data()));
+
+    // Make sure we can decompose the matrix.
+    int structuralZero;
+    auto statusPivot = cusparseXbsrilu02_zeroPivot(m_cuSparseHandle.get(), m_infoM.get(), &structuralZero);
+    OPM_ERROR_IF(statusPivot != CUSPARSE_STATUS_SUCCESS,
+                 fmt::format("Found a structucal zero at A({}, {}). Could not decompuse LU \\approx A.\n\n A has "
```
atgeirr, line 189: decompuse->decompose

atgeirr (commented):
This is fine, only minor quibbles.

kjetilly:
Jenkins build this please

atgeirr:
All green, all good!

PR #4638, opened 2023-05-10
Enable Zoltan-Based In-Process Subdomain Partitioning

bska:
Tested only for sequential running (`mpirun -np 1`) at the moment.  Intended to replace the original partitioning scheme based on a user-provided input file named `partition.txt`.

There are two new user-controlled runtime parameters

1. `NumLocalDomains` (`--num-local-domains`): Positive integer.  The partitioner will divide the domain into this many non-overlapping subdomains and run the non-linear domain decomposition method on those
2. `LocalDomainsPartitioningImbalance` (--local-domains-partitioning-imbalance`): Floating-point value $\ge 1$ which controls Zoltan's imbalance tolerance.  A value of $1.03$ corresponds to a tolerance of three percent domain imbalance

These parameters are not queried unless the run uses the NLDD non-linear solver.


bska:
I'm creating this in draft mode for two reasons

1. This PR depends on, and contains, the original NLDD PR (#3764)
2. The restriction of being limited to sequential runs is too severe and we should get to a point where the solver is usable in parallel rather soon.

PR #4640, opened 2023-05-10
Update of multisegment well pressure equations - testing

steink:
* rewrite segment hydrostatic p-drop as average of neighboring nodes
* slightly hacky inclusion of "skipped" derivatives in the friction-term for reverse-flow cases. Could be done similarly also for valve-models, but currently not included for these.

### DIFF
--- a/opm/simulators/wells/MultisegmentWellEval.cpp
+++ b/opm/simulators/wells/MultisegmentWellEval.cpp
@@ -214,7 +214,8 @@ template<typename FluidSystem, typename Indices, typename Scalar>
 void
 MultisegmentWellEval<FluidSystem,Indices,Scalar>::
 assembleDefaultPressureEq(const int seg,
-                          WellState& well_state)
+                          WellState& well_state, 
+                          const bool use_average_density)
 {
     assert(seg != 0); // not top segment
     const int seg_upwind = segments_.upwinding_segment(seg);
@@ -256,15 +257,22 @@ assembleDefaultPressureEq(const int seg,
         handleAccelerationPressureLoss(seg, well_state);
     }
 
-    // We approximate the hydrostatic pressure loss over the segment by using the mean of the densities
-    // at the current segment node and its outlet node. Since density derivatives are organized 
-    // differently than what is required for assemblePressureEq, this part needs to be assembled separately.  
-    const int seg_outlet = this->segmentNumberToIndex(this->segmentSet()[seg].outletSegment());
-    const auto hydro_pressure_drop_seg = segments_.getHalfHydroPressureLoss(seg, seg);
-    const auto hydro_pressure_drop_outlet = segments_.getHalfHydroPressureLoss(seg, seg_outlet);
-    MultisegmentWellAssemble<FluidSystem,Indices,Scalar>(baseif_).
-        assembleHydroPressureLoss(seg, seg_outlet, hydro_pressure_drop_seg, hydro_pressure_drop_outlet, linSys_);
-    segments.pressure_drop_hydrostatic[seg] = hydro_pressure_drop_seg.value() + hydro_pressure_drop_outlet.value();
+    // Since density derivatives are organized differently than what is required for assemblePressureEq, 
+    // this part needs to be assembled separately. Optionally use average density variant.  
+    const auto hydro_pressure_drop_seg = segments_.getHydroPressureLoss(seg, seg);
+    if (!use_average_density){
+        MultisegmentWellAssemble<FluidSystem,Indices,Scalar>(baseif_).
+            assembleHydroPressureLoss(seg, seg, hydro_pressure_drop_seg, linSys_);
+            segments.pressure_drop_hydrostatic[seg] = hydro_pressure_drop_seg.value();
+    } else {
+        const int seg_outlet = this->segmentNumberToIndex(this->segmentSet()[seg].outletSegment());
+        const auto hydro_pressure_drop_outlet = segments_.getHydroPressureLoss(seg, seg_outlet);
+        MultisegmentWellAssemble<FluidSystem,Indices,Scalar>(baseif_).
+            assembleHydroPressureLoss(seg, seg, 0.5*hydro_pressure_drop_seg, linSys_);
+        MultisegmentWellAssemble<FluidSystem,Indices,Scalar>(baseif_).
+            assembleHydroPressureLoss(seg, seg_outlet, 0.5*hydro_pressure_drop_outlet, linSys_);
+        segments.pressure_drop_hydrostatic[seg] = 0.5*hydro_pressure_drop_seg.value() + 0.5*hydro_pressure_drop_outlet.value();
+    }
 }
 
 template<typename FluidSystem, typename Indices, typename Scalar>
@@ -338,6 +346,7 @@ MultisegmentWellEval<FluidSystem,Indices,Scalar>::
 assemblePressureEq(const int seg,
                    const UnitSystem& unit_system,
                    WellState& well_state,
+                   const bool use_average_density,
                    DeferredLogger& deferred_logger)
 {
     switch(this->segmentSet()[seg].segmentType()) {
@@ -348,7 +357,7 @@ assemblePressureEq(const int seg,
             break;
         }
         default :
-            assembleDefaultPressureEq(seg, well_state);
+            assembleDefaultPressureEq(seg, well_state, use_average_density);
     }
 }
 
--- a/opm/simulators/wells/MultisegmentWellEval.cpp
+++ b/opm/simulators/wells/MultisegmentWellEval.cpp
@@ -261,10 +261,13 @@ assembleDefaultPressureEq(const int seg,
     }
 
     // Do separately since different organization of derivatives 
-    const auto hydro_pressure_drop = segments_.getHydroPressureLoss(seg);
+    const int seg_outlet = this->segmentNumberToIndex(this->segmentSet()[seg].outletSegment());
+    const auto hydro_pressure_drop1 = segments_.getHalfHydroPressureLoss(seg, seg);
+    const auto hydro_pressure_drop2 = segments_.getHalfHydroPressureLoss(seg, seg_outlet);
     MultisegmentWellAssemble<FluidSystem,Indices,Scalar>(baseif_).
-        assembleHydroPressureLoss(seg, seg_upwind, hydro_pressure_drop, linSys_);
-    segments.pressure_drop_hydrostatic[seg] = hydro_pressure_drop.value();
+        assembleHydroPressureLoss(seg, seg_outlet, hydro_pressure_drop1, hydro_pressure_drop2, linSys_);
+
+    segments.pressure_drop_hydrostatic[seg] = hydro_pressure_drop1.value() + hydro_pressure_drop2.value();
 }
 
 template<typename FluidSystem, typename Indices, typename Scalar>
--- a/opm/simulators/wells/MultisegmentWellEval.cpp
+++ b/opm/simulators/wells/MultisegmentWellEval.cpp
@@ -225,22 +225,12 @@ assembleDefaultPressureEq(const int seg,
     EvalWell extra_derivatives;
 
     // we need to handle the pressure difference between the two segments
-    // we only consider the hydrostatic pressure loss first
+    // hydrostatic pressure loss is assembled seperately at the end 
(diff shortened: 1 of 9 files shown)


GitPaean:
benchmark please

GitPaean:
jenkins build this please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: drogon - Threads: 1         | 1.059    |
| opm-git | OPM Benchmark: drogon - Threads: 8         | 1.393    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 1  | 0.995    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 8  | 1.007    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.022    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.028    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 0.984    |



View result details @ https://www.ytelses.com/opm/?page=result&id=2073

opm/simulators/wells/MultisegmentWellEval.cpp
```diff
@@ -217,22 +217,27 @@ assembleDefaultPressureEq(const int seg,
...
     auto& segments = ws.segments;
-    segments.pressure_drop_hydrostatic[seg] = hydro_pressure_drop.value();
-    pressure_equation -= hydro_pressure_drop;
+    //pressure_equation -= hydro_pressure_drop;
 
     if (this->frictionalPressureLossConsidered()) {
-        const auto friction_pressure_drop = segments_.getFrictionPressureLoss(seg);
+        const auto friction_pressure_drop = segments_.getFrictionPressureLoss(seg, false);
+        if (reverseFlow){
+            extra_derivatives = -segments_.getFrictionPressureLoss(seg, true);
```
GitPaean, line 239: The following code should go into this bracket? or at least inside the bracket (`if (this->frictionalPressureLossConsidered()) {`)
```c++
        MultisegmentWellAssemble<FluidSystem,Indices,Scalar>(baseif_).
            assemblePressureEqExtraDerivatives(seg, seg_upwind, extra_derivatives, linSys_);
```

GitPaean (changes requested):
This generally looks fine. The comments are very minor.  I need some more time to think about the frictional pressure loss part.

opm/simulators/wells/MultisegmentWellEval.cpp
GitPaean: In the current form, if there is no frictional pressure loss, calling the function `assemblePressureEqExtraDerivatives` can be dangerous?

opm/simulators/wells/MultisegmentWellSegments.cpp
```diff
@@ -484,19 +487,36 @@ getSurfaceVolume(const EvalWell& temperature,
...
+            const int WQTotal = Indices::numEq + PrimaryVariables::WQTotal;
+            const int SPres = Indices::numEq + PrimaryVariables::SPres;
+            density.setDerivative(WQTotal, 0.0);
+            density.setDerivative(SPres, 0.0);
+            visc.setDerivative(WQTotal, 0.0);
+            visc.setDerivative(SPres, 0.0);
+        } else {
+            const int WFrac = Indices::numEq + PrimaryVariables::WFrac;
+            const int GFrac = Indices::numEq + PrimaryVariables::GFrac;
+            // if has ...
```
GitPaean, line 511: Yeah, we need to have the guard in case there is no water or gas involved.

GitPaean:
And also, can you rebase to the latest branch? It is not required while recommended.

opm/simulators/wells/MultisegmentWellSegments.cpp
```diff
@@ -321,11 +321,14 @@ updateUpwindingSegments(const PrimaryVariables& primary_variables)
 template<class FluidSystem, class Indices, class Scalar>
 typename MultisegmentWellSegments<FluidSystem,Indices,Scalar>::EvalWell
 MultisegmentWellSegments<FluidSystem,Indices,Scalar>::
-getHydroPressureLoss(const int seg) const
-{
-    return densities_[seg] * well_.gravity() * depth_diffs_[seg];
+getHalfHydroPressureLoss(const int seg,
+                         const int seg_side) const
```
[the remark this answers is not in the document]
steink: Done

opm/simulators/wells/MultisegmentWellEval.cpp
```diff
@@ -217,22 +217,27 @@ assembleDefaultPressureEq(const int seg,
...
     // for top segment, the well control equation will be used.
     EvalWell pressure_equation = primary_variables_.getSegmentPressure(seg);
+    EvalWell extra_derivatives;
 
     // we need to handle the pressure difference between the two segments
     // we only consider the hydrostatic pressure loss first
     // TODO: we might be able to add member variables to store these values, then we update well state
     // after converged
-    const auto hydro_pressure_drop = segments_.getHydroPressureLoss(seg);
+
```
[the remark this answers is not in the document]
steink: Done

opm/simulators/wells/MultisegmentWellEval.cpp
steink: Yes, moved `assemblePressureEqExtraDerivatives` up into the `(if (this->frictionalPressureLossConsidered())  `-bracket

opm/simulators/wells/MultisegmentWellSegments.cpp
```diff
@@ -484,19 +487,36 @@ getSurfaceVolume(const EvalWell& temperature,
...
     EvalWell visc = viscosities_[seg_upwind];
     // WARNING
     // We disregard the derivatives from the upwind density to make sure derivatives
     // wrt. to different segments dont get mixed.
+    
     if (seg != seg_upwind) {
-        density.clearDerivatives();
-        visc.clearDerivatives();
+        if (!return_upwind_derivatives){
+            const int WQTotal = Indices::numEq + PrimaryVariables::WQTotal;
```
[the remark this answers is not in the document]
steink: Done

steink:
After the update 6dc90b2df77e4bdfc0af9d0e8d548f66bdf96bbf the behavior has slightly changed (one time-step cut versus zero for a simple test case). This is not really expected as the only functional change is the order of which the various terms of the MS pressure equations are assembled.

GitPaean:
> After the update [6dc90b2](https://github.com/OPM/opm-simulators/commit/6dc90b2df77e4bdfc0af9d0e8d548f66bdf96bbf) the behavior has slightly changed (one time-step cut versus zero for a simple test case). This is not really expected as the only functional change is the order of which the various terms of the MS pressure equations are assembled.

hei, at the end, it is the different orders of the assembly causing some deviations during the floating point arithmetic operations, which causes the simulations deviates a little bit. 

To recover the running results before commit https://github.com/OPM/opm-simulators/commit/6dc90b2df77e4bdfc0af9d0e8d548f66bdf96bbf , you can move the following code to the end of the function. 

```c++
    // We approximate the hydrostatic pressure loss over the segment by using the mean of the densities
    // at the current segment node and its outlet node. Since density derivatives are organized
    // differently than what is required for assemblePressureEq, this part needs to be assembled separately.
    const int seg_outlet = this->segmentNumberToIndex(this->segmentSet()[seg].outletSegment());
    const auto hydro_pressure_drop_seg = segments_.getHalfHydroPressureLoss(seg, seg);
    const auto hydro_pressure_drop_outlet = segments_.getHalfHydroPressureLoss(seg, seg_outlet);
    MultisegmentWellAssemble<FluidSystem,Indices,Scalar>(baseif_).
            assembleHydroPressureLoss(seg, seg_outlet, hydro_pressure_drop_seg, hydro_pressure_drop_outlet, linSys_);
    segments.pressure_drop_hydrostatic[seg] = hydro_pressure_drop_seg.value() + hydro_pressure_drop_outlet.value();
```

GitPaean:
jenkins build this please

steink:
> To recover the running results before commit https://github.com/OPM/opm-simulators/commit/6dc90b2df77e4bdfc0af9d0e8d548f66bdf96bbf , you can move the following code to the end of the function.

Yes, this recovers the results. Done.

> Personally, I think the following function should be called the first during the pressureEq assembly,

I kind of agree, but kept it as is for now for compactness. I guess we will want to return here at a later stage to check whether we should have special treatment of acceleration derivatives as well (possibly by reorganizing the assembling).

steink:
Reverted segment density calculations to the original formulation since this seemed to agree more closely with the reference. However, kept the average formulation as a hidden option (_--use-average-density-ms-wells_) since it appears to be more stable and more in line with the analogous formulation for the reservoir grid cells.

GitPaean:
jenkins build this please

GitPaean:
benchmark please

GitPaean:
I am checking the jenkins failures now before preparing to get the PR in.

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: drogon - Threads: 1         | 0.764    |
| opm-git | OPM Benchmark: drogon - Threads: 8         | 1.098    |
| opm-git | OPM Benchmark: smeaheia - Threads: 1       | 1.024    |
| opm-git | OPM Benchmark: smeaheia - Threads: 8       | 1.013    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 1  | 1.015    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 8  | 1.008    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.043    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.021    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 1.031    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.013    |



View result details @ https://www.ytelses.com/opm/?page=result&id=2091

GitPaean:
I reproduced the following benchmark results locally. 



Test | Configuration | Relative
-- | -- | --
opm-git | OPM Benchmark: drogon - Threads: 1 | 0.764
opm-git | OPM Benchmark: drogon - Threads: 8 | 1.098


Will discuss and evaluate.

steink:
A comment on **OPM Benchmark: drogon - Threads: 1** where there is a perfomance drop with the PR:

- Difference in run-time is mainly due to step 113 where there is a change in valve config. At this step, a reference run (with small time steps) shows an abrupt increase in WOPR (see fig).
- All configurations I ran (both master and PR) chops time-steps at this stage and captures the narrow peak. The only exception is master serial with default options which jumps over the peak. It also has convergence problems here, but accepts the step with relaxed tolerance.
- The behaviour of master serial at this step is very sensitive to perturbations. For instance, if it is restarted from HIST of some of  the other configurations, it also does the chopping at this stage.      

image

GitPaean:
jenkins build this please

bska:
> A comment on **OPM Benchmark: drogon - Threads: 1**

Thank you so much for the detailed analysis, @steink!  This is greatly appreciated.  Computing a more accurate numerical solution is of course vastly better than computing an incorrect solution so I think we should get this into the master branch rather quickly.

GitPaean:
This PR does not change (by default) the physical equations we are solving, but it updates/improves the derivatives and also the jacobian matrix we use, which affects the convergence behavoir. 

I checked the all the jenkins failures, most of the deviations are from different time stepping, which results from different convergence behavoir.  Some cases are more sensitive to the time stepping changes than other cases.

The performance degeneration (less than 25%) of the drogon case is also one of them.  And also, for the drogon case, the performance degeneration of the serial running are related to introduction of WSEGVALV keyword. It is possible when we improve the derivative regarding the inflow control devices, the situation might change. At the same time, in benchmarking/locall test, we also see 10% performance gain with the same code for drogon case in the parallel running. 

I think this PR is ready to update the reference data and prepare to be merged. 

I will try to do some refactoring after this PR get merged, hopefully for better readability and easier development.

GitPaean (approved):

GitPaean:
I am updating the reference data to prepare to get this PR in.

GitPaean:

GitPaean:
jenkins build this opm-tests=978 please

GitPaean:
The reference data is updated. I am getting the PR in now.

PR #4641, opened 2023-05-10
changed: use ctest mechanism for specifying number of processors required

akva2:
rather than forcing parallel tests to run sequentially. no change on jenkins but allows more efficient execution if more than 4 cpus are used for test execution.

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -130,7 +130,7 @@ endfunction()
 #   - This test class compares the output from a parallel simulation
 #     to the output from the serial instance of the same model.
 function(add_test_compare_parallel_simulation)
-  set(oneValueArgs CASENAME FILENAME SIMULATOR ABS_TOL REL_TOL DIR)
+  set(oneValueArgs CASENAME FILENAME SIMULATOR ABS_TOL REL_TOL DIR MPI_PROCS)
   set(multiValueArgs TEST_ARGS)
   cmake_parse_arguments(PARAM "$" "${oneValueArgs}" "${multiValueArgs}" ${ARGN} )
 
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

bska (commented):
The [FindMPI](https://cmake.org/cmake/help/v3.16/module/FindMPI.html) CMake module has the result variable `MPIEXEC_MAX_NUMPROCS` that might be useful here to avoid oversubscription.  Furthermore, the documentation suggests we also revise our run command to something resembling
```sh
${MPIEXEC_EXECUTABLE} \
${MPIEXEC_NUMPROC_FLAG} ${NUM_PROCS} \
${MPIEXEC_PREFLAGS} \
EXECUTABLE  \
${MPIEXEC_POSTFLAGS} \
EXE_ARGS
```
with `NUM_PROCS` being at most equal to `${MPIEXEC_MAX_NUMPROCS}`.  I don't know how much work it would be to update our run scripts accordingly.

akva2:
that is really a separate issue. 

i can't use the total number of procs as we are running up to 3 builds at the same time. we have 12 cores on jenkins and each of them are limited to 4 cpus. parallel tests runs with 4 procs by default for this reason. nothing changes there, this is machine specific and up to the user to control with the -jx param to ctest. as i said nothing changes on jenkins, but now i can run these tests in parallel on my machine with ctest -j32.

as for passing the mpi flags, sure, i can update the scripts for that but it is not related to the change here.

bska:
> we have 12 cores on jenkins and each of them are limited to 4 cpus.

Sorry, I don't understand what that means.  Do you mean we have a total of 48 ($= 4\times 12$) hardware threads on the CI system or do we have a set of 12 cores that are logically separated into three separate machines of 4 cores each?

> i can't use the total number of procs as we are running up to 3 builds at the same time.

I did not mean to give that impression.  I was merely pointing out that whatever value we use for the number of processors, it must not exceed `${MPIEXEC_MAX_NUMPROCS}`.  I gather that the value 4 is tuned for the CI system and is probably fine for current systems, but it absolutely was not on one of my earlier machines.

> now i can run these tests in parallel on my machine with `ctest -j32`.

Does that run 32 (paralle) tests at once or one test on 32 cores?

akva2:
sorry that was a bit of a unclear brain dump: 12 cores in total, each build job gets 4 of them.

the whole point here is exactly that ctest -j32 schedules tests so 32 cores can be used in parallel. so 8 of those parallel tests that specifies 4 processors for instance  or 32 single core tests, or any combination so the number or processes sum to at most 32.

back when i set this system up, ctest had no such support (2.8 days), so i was forced to run them sequentially to avoid oversubscribing the jenkins box.

atgeirr:
I regularly run tests with ctest -j 10 in, which seems to work (or at least I thought so until I tried in opm-models, where it does not for some reason). Will this PR change what I will experience (in opm-simulators)?

akva2:
yes it  will change what you experience. 

currently, all the parallel tests will run *sequentially*. With this PR two of them, along with 2 single-core tests can execute in parallel.

PR #4642, opened 2023-05-10
take into account trapped saturation from hysteresis

totto82:
Depends on https://github.com/OPM/opm-common/pull/3517


bska:
jenkins build this please

bska (approved):

PR #4646, opened 2023-05-11
output temperature if TEMP is set in RPTRST or THERMAL is set in RUNSPEC

totto82:
With this temperature is stored in the restart file if TEMP is used in RPTRST or THERMAL is set in RUNSPEC even if enable-opm-rst-file=false


totto82:
jenkins build this please

bska (approved):
I guess this is fine, but on the other hand don't think outputting a cell-sized array of unchanging and possibly defaulted values (60 F, 15.56 C) is going to be particularly useful to anyone.  I'll nevertheless merge this into master.

PR #4647, opened 2023-05-11
initialize trivial rvw if no table is given

totto82:
This is to make is easier to use VAPWAT for CO2STORE cases.

### DIFF
--- a/ebos/equil/initstateequil_impl.hh
+++ b/ebos/equil/initstateequil_impl.hh
@@ -1436,14 +1436,16 @@ InitialStateComputer(MaterialLawManager& materialLawManager,
             }
             else {
                 if (rec[i].gasOilContactDepth() != rec[i].datumDepth()) {
-                    throw std::runtime_error(
-                              "Cannot initialise: when no explicit RVWVD table is given, \n"
-                              "datum depth must be at the gas-oil-contact. "
-                              "In EQUIL region "+std::to_string(i + 1)+" (counting from 1), this does not hold.");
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

ebos/equil/initstateequil_impl.hh
```diff
@@ -1436,14 +1436,16 @@ InitialStateComputer(MaterialLawManager& materialLawManager,
             }
             else {
                 if (rec[i].gasOilContactDepth() != rec[i].datumDepth()) {
-                    throw std::runtime_error(
```
goncalvesmachadoc, line 1439: For a three-phase case it seems to be ok, but what happens for a WATER + GAS + VAPWAT case? Does this assume that a gas-oil contact (that is supposed to be ignored in water-gas cases) is read or would that be set as equal to the value from gas-water contact input in Equil keyword?

totto82: I am not sure if we have proper test cases for the combination of EQUIL + WATER + GAS + VAPWAT where table is not given but contact depth == datum depth.  But based on reading the code I agree that the current implementation seems wrong. It should have been waterOilContactDepth not gasOilContactDepth etc. I can do some testing, but I dont think a potential fix for this should be done in this PR. I would rather open a new one after this is merged.

plgbrts: @totto82 I could prepare some test cases and if you wish I can look further into the issue

totto82: Thanks. That would be great.

PR #4648, opened 2023-05-11
we should not request alq_value for injectors

GitPaean:
through the function { return well_.wellEcl().alq_value(); }.

It will throw a runtime_error which is not properly caught.

### DIFF
--- a/opm/simulators/wells/MultisegmentWellPrimaryVariables.cpp
+++ b/opm/simulators/wells/MultisegmentWellPrimaryVariables.cpp
@@ -393,6 +393,8 @@ copyToWellState(const MultisegmentWellGeneric<Scalar>& mswell,
                        [maxVel](const auto hf) { return (hf > 0.0) ? maxVel : 0.0; });
     }
 
+    // Note: for the ALQ value, in the StandardWell, WellInterfaceGeneric::getALQ(well_state) is used.
+    // We might want to unify the way regarding AQL value.
     WellBhpThpCalculator(well_)
         .updateThp(rho, stop_or_zero_rate_target, [this]() { return well_.wellEcl().alq_value(); },
                    {FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx),
--- a/opm/simulators/wells/WellBhpThpCalculator.cpp
+++ b/opm/simulators/wells/WellBhpThpCalculator.cpp
@@ -41,6 +41,7 @@
 
 #include <cassert>
 #include <cmath>
+#include <optional>
 
 static constexpr bool extraBhpAtThpLimitOutput = false;
 
@@ -101,7 +102,7 @@ double WellBhpThpCalculator::mostStrictBhpFromBhpLimits(const SummaryState& summ
 double WellBhpThpCalculator::calculateThpFromBhp(const std::vector<double>& rates,
                                                  const double bhp,
                                                  const double rho,
-                                                 const double alq,
+                                                 const std::optional<double>& alq,
                                                  DeferredLogger& deferred_logger) const
 {
     assert(int(rates.size()) == 3); // the vfp related only supports three phases now.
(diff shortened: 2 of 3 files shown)


GitPaean:
It might not be the best way, while it will be the starting point for discussion.

GitPaean:
What happened is that with the original code, 
```c++
ws.thp = this->calculateThpFromBhp(rates, ws.bhp, rho, alq_value(), deferred_logger);
```
it throws through the following code
```c++
double Well::alq_value() const {
    if (this->wtype.producer())
        return this->production->ALQValue.getSI();

    throw std::runtime_error("Can not ask for ALQ value in an injector");
}
```

But I did not see any message related to that (at least with parallel running (4 proc) what I have done) and my tracking of the function calls get lost there. And I think there is something related to exception handling might need some improvements.

akva2 (approved):
i think the optional is a better solution than the callback, should be unified to use optional.

bska:
We tried to address this in #4461.  Does that solution not work?  What part did we miss?

GitPaean:
> We tried to address this in https://github.com/OPM/opm-simulators/pull/4461. Does that solution not work? What part did we miss?

The following code in `updateThp` does not go through the modified function call in #4461. 
```c++
ws.thp = this->calculateThpFromBhp(rates, ws.bhp, rho, alq_value(), deferred_logger);
```

the code modifed in #4461 is 

```c++
double WellInterfaceGeneric::getALQ(const WellState& well_state) const
{
    // no alq for injectors.
    if (isInjector())
        return 0.0;

    return well_state.getALQ(name());
}
```

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -1160,6 +1160,7 @@ namespace Opm
                 const double thp = WellBhpThpCalculator(*this).calculateThpFromBhp(well_rates_bhp_limit,
                                                                                    bhp_limit,
                                                                                    this->connections_.rho(),
+                                                                                    // TODO: why not [this]() { return well_.wellEcl().alq_value(); } like WellBhpThpCalculator.updateThp()
```
GitPaean, line 1163: Maybe we should also use `this->getALQ(well_state)` for the WellBhpThpCalculator::upateThp()?

@bska , do you have some idea? We should unify the code to avoid confusion if we can.

GitPaean:
For the `updateThp` part, StandardWell and MultisegmentWell are doing different and called in different places. Maybe that is part of the reason why it got found at a so late time.

opm/simulators/wells/StandardWell_impl.hpp
atgeirr, line 1163: Remove this comment?

atgeirr (approved):
I agree with @akva2 that optional is good here.

opm/simulators/wells/StandardWell_impl.hpp
GitPaean: I would like to rephrase the comment, because it is pointing to a problem between the MSW and STDW well implementation. We should have a follow up PR to address that.

GitPaean: I moved and rephrased the comment to `WellBhpThpCalculator.cpp`. Since it is not causing problem in this file.

GitPaean:
I will unify the updateThp implementation between StandardWell and MultisegmentWell in a follow up PR.

GitPaean:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
bska: > Maybe we should also use `this->getALQ(well_state)` for the `WellBhpThpCalculator::upateThp()`?

Yes, I think so.

bska:
jenkins build this please

PR #4650, opened 2023-05-12
Add energy + precipitation of salt in brine-gas simulator

totto82:
The combination allows for modelling of salt precipitation in combination of thermal effects for a gas-brine system. Only tested in combination with CO2STORE.


totto82:
jenkins build this please

akva2 (approved):

PR #4651, opened 2023-05-12
fixed: need a separate directory for each restart test

akva2:
if not parallel execution of tests fail.


akva2:
jenkins build this please

bska (approved):

PR #4654, opened 2023-05-12
Add solvent + foam simulator

totto82:
Also allow for water as transport phase in the well model. 
Depends on https://github.com/OPM/opm-common/pull/3523 and https://github.com/OPM/opm-models/pull/805


GitPaean:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -744,8 +744,17 @@ namespace Opm
 
         if constexpr (has_foam) {
             // TODO: the application of well efficiency factor has not been tested with an example yet
-            const unsigned gasCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx);
-            EvalWell cq_s_foam = cq_s[gasCompIdx] * this->well_efficiency_factor_;
+            unsigned foamTransportIdx = 0;
+            if (FoamModule::transportPhase() == Phase::GAS)
```
GitPaean, line 748: Please use `switch` here, 

and also safety check for `has_solvent`. 

Otherwise, when `FoamModule::transportPhase() == Phase::SOLVENT` and `has_solvent == false`, the error message might not point to the real problem.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -744,8 +744,17 @@ namespace Opm
 
         if constexpr (has_foam) {
             // TODO: the application of well efficiency factor has not been tested with an example yet
-            const unsigned gasCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx);
-            EvalWell cq_s_foam = cq_s[gasCompIdx] * this->well_efficiency_factor_;
+            unsigned foamTransportIdx = 0;
```
GitPaean, line 747: `0` is a usable indices. One way to improve is to make the code below to be a lambda function with name e.g. `getComputeFoamTransportIdx();`, then we can use

`const unsigned foamTransportIdx = computeFoamTransportIdx()` to get the index.

GitPaean:
I think we should have the regression test added here. Otherwise, the development along this line will not be protected after getting merged in.

opm/simulators/wells/StandardWellPrimaryVariables.cpp
```diff
@@ -741,5 +741,7 @@ INSTANCE(BlackOilIndices<0u,0u,0u,0u,true,false,0u,0u>)
 INSTANCE(BlackOilIndices<0u,0u,0u,0u,false,true,0u,0u>)
 INSTANCE(BlackOilIndices<0u,0u,0u,1u,false,true,0u,0u>)
 INSTANCE(BlackOilIndices<0u,0u,0u,1u,false,false,1u,0u>)
+INSTANCE(BlackOilIndices<0u,1u,0u,0u,true,false,0u,0u>)
+INSTANCE(BlackOilIndices<1u,0u,0u,0u,true,false,0u,0u>)
```
GitPaean, line 745: @akva2 , can you have a look at the part related to `BlackOilIndices`? thanks.

opm/simulators/wells/VFPHelpers.cpp
```diff
@@ -648,14 +648,15 @@ INSTANCE(DenseAd::Evaluation<double, -1, 7u>)
...
+INSTANCE(DenseAd::Evaluation<double, -1, 11u>)
 INSTANCE(DenseAd::Evaluation<double, 3, 0u>)
 INSTANCE(DenseAd::Evaluation<double, 4, 0u>)
 INSTANCE(DenseAd::Evaluation<double, 5, 0u>)
 INSTANCE(DenseAd::Evaluation<double, 6, 0u>)
 INSTANCE(DenseAd::Evaluation<double, 7, 0u>)
 INSTANCE(DenseAd::Evaluation<double, 8, 0u>)
 INSTANCE(DenseAd::Evaluation<double, 9, 0u>)
-
+INSTANCE(DenseAd::Evaluation<double, 10, 0u>)
```
GitPaean, line 659: maybe also the part related to `INSTANCE(DenseAd::Evaluation`. I have to say that I am really not familiar with these. @akva2  thanks.

opm/simulators/wells/StandardWellPrimaryVariables.cpp
akva2: only thing I find surprising is that we need two new instances. It's just one new simulator?

opm/simulators/wells/VFPHelpers.cpp
akva2: it simply have evaluations with one more component than what was used before, so we need to instance for those evaluations. it's all good.

opm/simulators/wells/StandardWellPrimaryVariables.cpp
totto82: I used the code to also test the foam + the extended blackoil module (an improved solvent option), but dropped it. I will remove the redundant lines.

totto82:

GitPaean:
jenkins build this opm-tests=972 please

GitPaean:
Thanks for the update. I am getting the PR in now.

PR #4655, opened 2023-05-12
fixed: copy the mobility values into the temporary Eval vector

akva2:
The updateWaterMobilityWithPolymer method updates values, it does not rewrite them. Hence we have to feed it the scalar values not zeros.

I noticed this while doing is some other refactoring. It does not affect any current tests.


akva2:
jenkins build this please

GitPaean:
It looks like a bug, but I do not understand why it works with the original version. Let me test a little bit.

GitPaean:
> It works in the original because we do not make the temporary copy, we directly pass the mob vector.

Yes. I might affect some output values, while we do not check those.

GitPaean:
jenkins build this opm-tests=968 please

GitPaean:
Thanks for fixing the bug. jenkins approves and I am getting this PR in now.

PR #4660, opened 2023-05-22
TUNING/NEXT[STEP]: Use max next time step only when specified

vkip:
Use max time step from TUNING.TSINIT or NEXTSTEP only when specified.
Requires https://github.com/OPM/opm-common/pull/3528

### DIFF
--- a/tests/test_tuning_TSINIT_NEXTSTEP.cpp
+++ b/tests/test_tuning_TSINIT_NEXTSTEP.cpp
@@ -14,8 +14,9 @@ inline bool is_close(double a, double b, double tol = TOLERANCE) {
 }
 
 BOOST_AUTO_TEST_CASE(CheckTSINITAndNEXTSTEP)
-{    
-    //std::string case_name(boost::unit_test::framework::master_test_suite().argv[1]);        std::string case_name("02_TUNING_TSINIT_NEXTSTEP");
+{
+    std::string case_name("02_TUNING_TSINIT_NEXTSTEP");
+
     BOOST_TEST_MESSAGE("---------------------------------------------------------------------------");
     BOOST_TEST_MESSAGE("Checking TSINIT and NEXTSTEP, see file " + case_name + ".DATA");
     BOOST_TEST_MESSAGE("---------------------------------------------------------------------------");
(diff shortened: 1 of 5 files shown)


vkip:
jenkins build this opm-common=3528 please

vkip:
jenkins build this opm-common=3528 please

akva2:
I added you to trigger list now.

vkip:
jenkins build this opm-common=3528 please

vkip:
The test fails due to this change to DOUBHEAD (negative value now signals no limit from TUNING/NEXT[STEP] on time step size)

https://github.com/OPM/opm-common/pull/3528/files#diff-ea70b392e8e8baad69c2cd36eda25b9739c95beded04946f68b5c4b0ec21a061

The use of negative values is a new convention, so reference results need to be updated if the PR is accepted. 


The most confusing was why some tests passed at all, since most '1' values at this location now are replaced with '-1'. Turns out the reason is here: 

https://github.com/OPM/opm-common/blob/35264e7ab83ace38e7ef4f2727c20f336a6cbd3c/test_util/EclFilesComparator.cpp#L106-L117

Instead of returning 0 (abs and rel) deviation between '-1' and '1', could perhaps lines 107-108 be removed and replace line 113 with:

deviation.rel = deviation.abs/(std::max(std::abs(val1), std::abs(val2)));

so that relative deviation > 1 indicates different signs?

vkip:
New test depends on https://github.com/OPM/opm-tests/pull/1028

tests/test_tuning_TSINIT_NEXTSTEP.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <boost/test/unit_test.hpp>
+
+constexpr double TOLERANCE = 1.0e-10;
+
+inline bool is_close(double a, double b, double tol = TOLERANCE) {
+    return std::abs(a-b) <= tol;
+}
+
+BOOST_AUTO_TEST_CASE(CheckTSINITAndNEXTSTEP)
+{
```
akva2, line 17: indent is off.

vkip:
jenkins build this opm-common=3528 opm-tests=1028 please

vkip:
The one failing test, BC_LAB, appears to fail because the current master erroneously writes TSINIT=24 hours instead of 1 hour if there is no explicit setting of TSINIT in TUNING.

bska (approved):
Thanks a lot for the updates.  This looks good to me now and the new reference solutions have been installed on  the CI system.  I'll merge this into the master branch now.

PR #4661, opened 2023-05-22
enable disgasw and vapwat for thermal model

totto82:
I suggest we add it to the thermal module rather than adding yet another version with dissolution + thermal.


totto82:
jenkins build this please

akva2 (approved):
Code wise it looks fine. I cannot comment on the approach taken though, other than avoiding additional simulators seems like a good idea to me.

PR #4664, opened 2023-05-23
dont store history of inactive gpmaint rates

totto82:
in the old code negative gpmaint rates where calculated and then removed by a `max(0.0, rates)`, but this introduce the wrong history to gpmaint which results in wrong rates.

### DIFF
--- a/opm/simulators/wells/WellGroupHelpers.cpp
+++ b/opm/simulators/wells/WellGroupHelpers.cpp
@@ -707,12 +707,12 @@ namespace WellGroupHelpers
         const double error = gpm->pressure_target() - regional_values.at(name)->pressure(number);
         double current_rate = 0.0;
         const auto& pu = well_state.phaseUsage();
-        double sign = 1.0;
+        bool injection = true;
         switch (gpm->flow_target()) {
             case GPMaint::FlowTarget::RESV_PROD:
             {
                 current_rate = -group_state.injection_vrep_rate(group.name());
-                sign = -1.0;
+                injection = false;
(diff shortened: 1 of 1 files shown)


opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -760,8 +760,11 @@ namespace WellGroupHelpers
                 throw std::invalid_argument("Invalid Flow target type in GPMAINT");
         }
         auto& gpmaint_state = group_state.gpmaint(group.name());
-        double rate = gpm->rate(gpmaint_state, current_rate, error, dt);
-        group_state.update_gpmaint_target(group.name(), std::max(0.0, sign * rate));
+        // we only activate gpmaint if pressure is lower than the target regional pressure for injectors
+        // (i.e. error > 0) and higher for producers.
+        bool activate = (injection && error > 0) || (!injection && error < 0);
+        double rate = activate? gpm->rate(gpmaint_state, current_rate, error, dt) : 0.0;
```
atgeirr, line 766: I do not understand the "history" comment in the PR description. From what I can tell, rate() is a const function that does not store any history, and the update_gpmaint_target() below also seemingly overwrites a target number, not storing anything. What am I missing?

Also, will the rate() method return a negative number for producers? If not, then the old code was simply buggy, and if it does, the code here is buggy.

atgeirr: I see it now, the gpmaint_state is modified.

PR #4665, opened 2023-05-24
avoid unused variable warning with NDEBUG

akva2:
Since the assert is nop the reference will be unused.


akva2:
jenkins build this please

PR #4667, opened 2023-05-25
changed: stop forcing no adaptive time stepping in parallel restart test driver

akva2:
this should be controllable on a per-test level as other options.

keep current behavior for existing test by re-adding the option using TEST_ARGS.


akva2:
jenkins build this please

bska:
jenkins build this please

bska (approved):
Thanks.  I'll merge this now.

PR #4544, opened 2023-03-23
faster begintimestep

hnil:
- faster updateProperty
- refactored for making local updating in inhereted classes

PR #4668, opened 2023-05-26
remove duplicated code

totto82:
Something went wrong during a rebase???

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1120,7 +1120,6 @@ public:
         asImp_().updateCompositionChangeLimits_();
         {
         OPM_TIMEBLOCK(driftCompansation);
-        asImp_().updateCompositionChangeLimits_();
         if (enableDriftCompensation_) {
             const auto& residual = this->model().linearizer().residual();
             for (unsigned globalDofIdx = 0; globalDofIdx < residual.size(); globalDofIdx ++) {


bska:
Does this impact anything other than the computational performance?  In other words, does calling the function twice affect correctness?  Because if it does then we have insufficient test coverage or it would have been caught while reviewing PR #4544.

bska:
@hnil: Would you mind having a look at this, please?  I'm particularly thinking about the `TIMEBLOCK` tags here and how we want to attribute the cost of this function call.

totto82:
> Does this impact anything other than the computational performance

It shouldn't impact the results. There is already a TIMEBLOCK within the function call

bska:
> > Does this impact anything other than the computational performance
> 
> It shouldn't impact the results. There is already a TIMEBLOCK within the function call

Yes, but I was wondering **which** `TIMEBLOCK` to attribute the cost to.  Removing the inner function call will attribute the cost to the outer, whereas removing the outer will attribute the cost to the inner.  I wanted to ask what the intention was.

totto82:
> I was wondering which TIMEBLOCK to attribute the cost to.

updateCompositionChangeLimits_ has nothing to do with drift-compensation.

atgeirr:
> Does this impact anything other than the computational performance?

I think not, will merge this in any case.

PR #4670, opened 2023-05-26
Don't Provide Deck Accessors in FlowMainEbos

bska:
Following commits f4f8c033d and c7016854d (PR #4286), the "vanguard" no longer maintains an internal `Deck` data member.  Don't pretend that it does by providing member functions to access that data member.  If someone tries to actually call these member functions then they will get an unfriendly diagnostic message and a build failure.


bska:
jenkins build this please

akva2 (approved):
Yup, forgotten leftovers as they cause no build failures..

bska:
Merging this now following maintainer approval.

PR #4673, opened 2023-05-31
Added working PINCH test cases as regressions.

akva2:
This hopefully will prevent breaking things when further work for PINCH is done.

The following tests are still known to fail with current master: T2A_NOPINCH T2A_GAP T2A_NOGAP T4A1_NOGAP T4A1_GAP

Fixed version of https://github.com/OPM/opm-simulators/pull/4671

PR #4674, opened 2023-05-31
changed: no need to manually specify test data in update script

akva2:
rather specify it as properties on the tests themself and extract using some awk-ing.

let's streamline this a bit more. it's a bit nasty digging into the ctest file but i cannot get the information otherwise.

tested through https://github.com/OPM/opm-simulators/pull/4673 (removed from that PR again now).

### DIFF
--- a/tests/update_reference_data.sh
+++ b/tests/update_reference_data.sh
@@ -49,194 +49,6 @@ copyToReferenceDir () {
   return $DIFF
 }
 
-declare -A tests
-# The tests are listed in a dictionary mapping name of the test to the commands
-# used to run the test:
-#
-#    tests[cmake_testname]="binary dirname SIMULATION_CASE
-#
-# The key in the dictionary must agree with the name given to the test when
-# registering it with cmake in regressionTests.cmake. The SIMULATION_CASE should
-# be the basename of the .DATA file used for the simulation.
-
-tests[spe1]="flow spe1 SPE1CASE1"
-tests[spe12]="flow spe1 SPE1CASE2"
-tests[spe1_2p]="flow spe1 SPE1CASE2_2P"
-tests[spe1_oilgas]="flow spe1 SPE1CASE2_OILGAS"
-tests[spe1_gaswater]="flow spe1 SPE1CASE2_GASWATER"
-tests[spe1_nowells]="flow spe1 SPE1CASE2_NOWELLS"
-tests[spe1_thermal]="flow spe1 SPE1CASE2_THERMAL"
-tests[spe1_thermal_onephase]="flow spe1 SPE1CASE2_THERMAL_ONEPHASE"
-tests[spe1_thermal_watvisc]="flow spe1 SPE1CASE2_THERMAL_WATVISC"
-tests[spe1_rockcomp]="flow spe1 SPE1CASE2_ROCK2DTR"
-tests[spe1_brine]="flow spe1_brine SPE1CASE1_BRINE"
-tests[spe1_precsalt]="flow spe1_precsalt SPE1CASE1_PRECSALT"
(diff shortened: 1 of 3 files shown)


akva2:
can this be merged? it's a significant convenience improvement for the test suite.

tests/update_reference_data.sh
```diff
@@ -246,40 +58,38 @@ test -z "$FAILED_TESTS" && exit 5
...
-    then
-      copyToReferenceDir \
+  grep -q compareECLFiles <<< $failed_test
+  test $? -ne 0 && continue
+  failed_test=`echo $failed_test | sed -e 's/.*://g'`
+  # Extract test properties
+  binary=`cat $BUILD_DIR/CTestTestfile.cmake | grep set_tests_properties\($failed_test| awk -F ' ' '{for (i=0; i <= NF; ++i) { if ($i == "SIMULATOR") {print $(i+1); exit}}}' | sed -e 's/\"//g'`
+  dir_name=`cat $BUILD_DIR/CTestTestfile.cmake | grep set_tests_properties\($failed_test| awk -F ' ' '{for (i=0; i <= NF; ++i) { if ($i == "DIRNAME") {print $(i+1); exit}}}' | sed -e 's/\"//g'`
+  file_name=`cat $BUILD_DIR/CTestTestfile.cmake | grep set_tests_properties\($failed_test| awk -F ' ' '{for (i=0; i <= NF; ++i) { if ($i == "FILENAME") {print $(i+1); exit}}}' | sed -e 's/\"//g'`
+  test_name=`cat $BUILD_DIR/CTestTestfile.cmake | grep set_tests_properties\($failed_test| awk -F ' ' '{for (i=0; i <= NF; ++i) { if ($i == "TESTNAME") {print $(i+1); exit}}}' | sed -e 's/\"//g'`
```
bska, line 68: You can reduce the number of processes here by writing a slightly more targeted awk script leveraging awk's built-in pattern matching engine:
```sh
v=$(awk -v search="set_tests_properties\\\($failed_test" \
  '$1 ~ search { for (i = 3; i <= NF; ++i) { if ($i == "SOMETHING") { val = $(i + 1); gsub(/"/, "", val); print val; exit } } } \
  $BUILD_DIR/CTestTestfile.cmake)
```
Note that I'm starting at `i = 3`, because `$0` is the entire line, `$1` would be the `set_tests_properties` command, and `$2` would be the word `PROPERTIES`.  If you don't want to make the latter two assumptions you should nevertheless start at `i = 1`.

Building on that we could even have a separate awk script in a file named `getprop.awk` of the form
```awk
$1 ~ search {
    for (i = 3; i <= NF; ++i) {
        if ($i == prop) {
            val = $(i + 1)
            gsub(/"/, "", val)
            print val
            exit
        }
    }
}
```
and invoke this from `update_reference_data.sh` as
```sh
binary=$(awk -v search="set_tests_properties\\\($failed_test" -v prop="SIMULATOR" -F getprop.awk $BUILD_DIR/CTestTestfile.cmake)
dir_name=$(awk -v search="set_tests_properties\\\($failed_test" -v prop="DIRNAME" -F getprop.awk $BUILD_DIR/CTestTestfile.cmake)
file_name=$(awk -v search="set_tests_properties\\\($failed_test" -v prop="FILENAME" -F getprop.awk $BUILD_DIR/CTestTestfile.cmake)
test_name=$(awk -v search="set_tests_properties\\\($failed_test" -v prop="TESTNAME" -F getprop.awk $BUILD_DIR/CTestTestfile.cmake)
```

bska (commented):
I like the idea.  I do think the way of extracting the property values could be improved slightly though.

bska (approved):
Thanks.  I'll merge into master and then we'll runit in production in due time.

PR #4675, opened 2023-05-31
Fix rvw and rv initialization

totto82:
Why was abs used here?!?

### DIFF
--- a/ebos/equil/equilibrationhelpers_impl.hh
+++ b/ebos/equil/equilibrationhelpers_impl.hh
@@ -175,7 +175,12 @@ operator()(const double depth,
            const double temp,
            const double satOil) const
 {
-    if (std::abs(satOil) > 1e-16) {
+    if (satOil < - std::sqrt(std::numeric_limits<double>::epsilon())) {
+        throw std::logic_error {
+            "Must not pass negative oil saturation"
+        };
+    }
+    if (satOil > 1e-16) {
         return satRv(press, temp);
     }
     else {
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

bska (commented):
> Why was abs used here?

I don't know.  The current code appears to have originated in commit f65663ffd, but the commit message mentions back porting another commit.  In any case, `std::abs()`, even though it's a common way of identifying  non-zero values, isn't appropriate here.  We arguably should not call the function with a negative saturation value, but if we want to detect that situation we should do it at a higher level.

totto82:
> We arguably should not call the function with a negative saturation value

For the case I was looking into, So was slightly below -1e-16. And thus rvsat was used instead of the tabulated values. Since So = 1- sg- sw I assume it could be some round off errors. Since So is only used to check if it is larger than 0, I dont think a very small negative value will be problematic for this call as long as the std::abs is removed. I will argue that we should merge this change as is. That will also make it consistent with similar code for rs/rsw.

totto82:
@bska merge?

bska:
> merge?

Maybe.  I still think we should try to detect a caller passing a negative saturation value here, because that suggests that there would be logic errors elsewhere.

Perhaps we could add a check of the form
```C++
if (So < - std::sqrt(std::numeric_limits<double>::epsilon())) {
    throw std::logic_error {
        "Must not pass negative oil saturation"
    };
}
```
:question:

bska:
> > Maybe. I still think we should try to detect a caller passing a negative saturation value here, because that suggests that there would be logic errors elsewhere.
> 
> Your point is that we should investigate why some cells (typically water filled cells) has so `-1.0e-16` and not exactly zero before merging this and forget about it.

That's not what I said.

If the only case for which `So < 0` is when `So = -1.something * 1.0e-16`, then there's no problem.  My concern is if we get `So < -1.0e-6` or thereabouts.  Then there's a problem somewhere else and I would rather know about it than silently accepting the value as being something else.

totto82:
> That's not what I said.

Thanks for the clarification. If I had read the code you suggested carefully i would have noticed. I will add the check.

totto82:
jenkins build this please

bska (approved):

PR #4680, opened 2023-06-01
start regression testing of test case with VAPPARS keyword 

tskille:
OPM Flow support of keyword VAPPARS used in Solution section was achieve with pr

   https://github.com/OPM/opm-common/pull/3550 

This has now been tested and simulation results are excellent and with this PRI would like to start regression testing.

### DIFF
--- a/regressionTests.cmake
+++ b/regressionTests.cmake
@@ -1155,3 +1155,10 @@ add_test_compareECLFiles(CASENAME 02_wgrupcon
                          ABS_TOL ${abs_tol}
                          REL_TOL ${rel_tol}
                          DIR wgrupcon)
+add_test_compareECLFiles(CASENAME 01_vappars
+                         FILENAME VAPPARS-01
+                         SIMULATOR flow
+                         ABS_TOL ${abs_tol}
+                         REL_TOL ${rel_tol}
+                         DIR vappars)                         
+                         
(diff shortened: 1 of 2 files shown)


regressionTests.cmake
```diff
@@ -1155,3 +1155,10 @@ add_test_compareECLFiles(CASENAME 02_wgrupcon
                          ABS_TOL ${abs_tol}
                          REL_TOL ${rel_tol}
                          DIR wgrupcon)
+add_test_compareECLFiles(CASENAME 01_vappars
+                         FILENAME 01-VAPPARS
```
bska, line 1159: I don't see a case named `01-VAPPARS`, but we do have a case named [`VAPPARS-01`](https://github.com/OPM/opm-tests/blob/e29c0fa7cac292fedf0f3be49c606c210f5007c4/vappars/VAPPARS-01.DATA) in the `vappars` directory of OPM-Tests.  Did you mean that one?

tskille: thanks @bska, my bad. This is fixed now.

bska:

bska:
jenkins build this opm-tests=982 please

bska:
The new reference solutions have been installed on the CI system.  Merging this now to activate the regression test.

PR #4158, opened 2022-10-11
supporting WINJMULT

GitPaean:
This is the downstream PR of PR https://github.com/OPM/opm-common/pull/3161 .

PR #4686, opened 2023-06-05
Support  WINJMULT

GitPaean:
This is the downstream PR of PR https://github.com/OPM/opm-common/pull/3161 .

re-opening here potentially to replace the existing #4158 . Mostly due to the difficulties encountered during the rebase process.  Keeping the older one for investigation purpose.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1411,16 +1411,18 @@ void BlackoilWellModelGeneric::initInjMult() {
         if (well->isInjector() && well->wellEcl().getInjMultMode() != Well::InjMultMode::NONE) {
             const auto& ws = this->wellState().well(well->indexOfWell());
             const auto& perf_data = ws.perf_data;
-            well->initInjMult(perf_data.inj_multiplier);
+            if ((this->prev_inj_multipliers_.count(well->name())) == 0 ) {
+                this->prev_inj_multipliers_[well->name()] = std::vector<double>(perf_data.size(), 1.0);
+            }
+            well->initInjMult(this->prev_inj_multipliers_.at(well->name()));
         }
     }
 }
 
 void BlackoilWellModelGeneric::updateInjMult(DeferredLogger& deferred_logger) {
     for (const auto& well : this->well_container_generic_) {
-        if (well->wellEcl().getInjMultMode() != Well::InjMultMode::NONE && well->isInjector()) {
-            auto& perf_data = this->wellState().well(well->indexOfWell()).perf_data;
-            well->updateInjMult(perf_data.inj_multiplier, deferred_logger);
+        if (well->isInjector() && well->wellEcl().getInjMultMode() != Well::InjMultMode::NONE) {
+            well->updateInjMult(this->prev_inj_multipliers_[well->name()], deferred_logger);
         }
     }
 }
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1411,10 +1411,12 @@ void BlackoilWellModelGeneric::initInjMult() {
         if (well->isInjector() && well->wellEcl().getInjMultMode() != Well::InjMultMode::NONE) {
             const auto& ws = this->wellState().well(well->indexOfWell());
             const auto& perf_data = ws.perf_data;
-            if ((this->prev_inj_multipliers_.count(well->name())) == 0 ) {
-                this->prev_inj_multipliers_[well->name()] = std::vector<double>(perf_data.size(), 1.0);
+
+            auto &values = this->prev_inj_multipliers_[well->name()];
+            if (values.empty()) {
+                values.assign(perf_data.size(), 1.0);
             }
-            well->initInjMult(this->prev_inj_multipliers_.at(well->name()));
+            well->initInjMult(values);
         }
     }
 }
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1405,4 +1405,24 @@ int BlackoilWellModelGeneric::numLocalNonshutWells() const
     return well_container_generic_.size();
 }
 
+
+void BlackoilWellModelGeneric::initInjMult() {
+    for (auto& well : this->well_container_generic_) {
+        if (well->isInjector() && well->wellEcl().getInjMultMode() != Well::InjMultMode::NONE) {
+            const auto& ws = this->wellState().well(well->indexOfWell());
+            const auto& perf_data = ws.perf_data;
+            well->initInjMult(perf_data.inj_multiplier);
+        }
+    }
+}
+
+void BlackoilWellModelGeneric::updateInjMult(DeferredLogger& deferred_logger) {
+    for (const auto& well : this->well_container_generic_) {
+        if (well->wellEcl().getInjMultMode() != Well::InjMultMode::NONE && well->isInjector()) {
+            auto& perf_data = this->wellState().well(well->indexOfWell()).perf_data;
+            well->updateInjMult(perf_data.inj_multiplier, deferred_logger);
+        }
+    }
+}
+
 }
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -22,6 +22,7 @@
 #include <opm/common/OpmLog/OpmLog.hpp>
 
 #include <opm/input/eclipse/Schedule/MSW/Valve.hpp>
+#include <opm/input/eclipse/Schedule/Well/WellConnections.hpp>
 #include <opm/input/eclipse/Units/Units.hpp>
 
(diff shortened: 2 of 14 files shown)


GitPaean:
jenkins build this opm-common=3161 please

GitPaean:
jenkins build this opm-common=3577 please

bska:
> jenkins build this opm-common=3577 please

This is okay, but as a rule of thumb it is generally better to trigger builds affecting multiple repositories/PRs from the most upstream PR&mdash;in this case from OPM-Common.  The reason is that if you trigger the build from a downstream PR then you don't necessarily get a full test run of upstream PRs.

GitPaean:
> The reason is that if you trigger the build from a downstream PR then you don't necessarily get a full test run of upstream PRs.

Thanks for the comments. But what is the cause of it?  Upstream PRs need to be incorporated and recompiled I would think.

akva2:
simply to save time on the ci system. by triggering from the downstream you're saying 'no need to build test suite and run tests for the upstreams, they are not affected'. obviously the libraries are still built.

GitPaean:
>  simply to save time on the ci system. by triggering from the downstream you're saying 'no need to build test suite and run tests for the upstreams, they are not affected'. obviously the libraries are still built.

Thanks for the response. It is a design and not a system fallout.

GitPaean:
jenkins build this please

GitPaean:
TODO: will update the tolerance for the parallel testing to make it pass. I does not look there is any problem for the well curves comparison.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
It is not easy to know what is the exact cause of the failures for parallel test. The well curves are okay. 

```
1 keyword exhibit failures
	XCON: Restart, sequence 16
		Fails for 4 entries
		Largest absolute error: 4.8069973e+00
		Largest relative error: 2.0416380e-01

Program finished 
```

GitPaean:
jenkins build this please

GitPaean:
The attempt to have the inj_multi stored in PerfData is not successful, since when a wells is `SHUT`, the information in the PerfData will not be continued.  While for the keyword WINJMULT, even a well is `SHUT`, we need to keep the related to information for later usage in case the well is re-opened. 

Maybe in this PR, I have no intention to change how `PerfData` and `WellState` work in case a well is `SHUT`. As a result, I will change back to store the `inj_multi` related in the `BlackoilWellModelGeneric`.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
jenkins build this  opm-tests=999 please

GitPaean:
@bska , can you have a look at this PR? from my side, it is ready to update the reference now.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -996,6 +1001,20 @@ namespace Opm
...
         WellInterface<TypeTag>::getMobility(ebosSimulator, perf, mob, obtain, deferred_logger);
+	
+        if (this->isInjector() && this->well_ecl_.getInjMultMode() != Well::InjMultMode::NONE) {
+            // from the reference results, it looks like MSW uses segment pressure instead of BHP here
+            // Note: this is against the documented definition.
+            // we can change this depending on what we want
+            const double segment_pres = this->primary_variables_.getSegmentPressure(seg).value();
+            const double perf_seg_press_diff = this->gravity() * this->segments_.density(seg).value()
+                                                               * this->segments_.perforation_depth_diff(perf);
+            const double perf_press = this->primary_variables_.getSegmentPressure(seg).value() + perf_seg_press_diff;
```
bska, line 1012: Tiny nit, but isn't this the same as
```C++
perf_press = segment_press + perf_seg_press_diff;
```
Why do you need to look up the segment pressure again?

opm/simulators/wells/WellInterfaceGeneric.cpp
```diff
@@ -190,6 +190,68 @@ double WellInterfaceGeneric::rsRvInj() const
...
+
+
+
+double WellInterfaceGeneric::getInjMult(const int perf,
+                                        const double bhp,
+                                        const double perf_pres) const
+{
+    assert(!this->isProducer());
+
+    double multipler = 1.;
```
bska, line 222: Do you want to write this as "multiplier" instead?  Right now the word is missing an "i" between "l" and "e".

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -982,6 +986,7 @@ namespace Opm
     void
     MultisegmentWell<TypeTag>::
     getMobility(const Simulator& ebosSimulator,
+                const int seg,
```
bska, line 989: I think you could probably infer the segment from the perforation if you don't want to change the API.  On the other hand, that might just be an expensive way to recompute information you already have so I'm okay with adding the new parameter.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1405,4 +1405,26 @@ int BlackoilWellModelGeneric::numLocalNonshutWells() const
...
 }
 
+
+void BlackoilWellModelGeneric::initInjMult() {
+    for (auto& well : this->well_container_generic_) {
+        if (well->isInjector() && well->wellEcl().getInjMultMode() != Well::InjMultMode::NONE) {
+            const auto& ws = this->wellState().well(well->indexOfWell());
+            const auto& perf_data = ws.perf_data;
+            if ((this->prev_inj_multipliers_.count(well->name())) == 0 ) {
+                this->prev_inj_multipliers_[well->name()] = std::vector<double>(perf_data.size(), 1.0);
```
bska, line 1415: This is fractionally better written as
```C++
this->prev_inj_multipliers_[well->name()].assign(perf_data.size(), 1.0)
```
You can even go further and write all of this, including the check for existence, as
```C++
this->prev_inj_multipliers_
    .emplace(std::piecewise_construct,
             std::forward_as_tuple(well->name()),
             std::forward_as_tuple(perf_data.size(), 1.0));
```
This statements depends on the fact that `unordered_map<>::emplace()` does not insert into the map if the key already exists.  That way you don't have to look up `well->name()` twice.  Function [`forward_as_tuple()`](https://en.cppreference.com/w/cpp/utility/tuple/forward_as_tuple) is declared in the `<tuple>` header.

bska (commented):
I think this looks good for the most part.  Just a couple of minor comments and then I'm ready to merge this.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
GitPaean: At the end, I will use the following code. It is more readable to me and it only searches once. 

```c++
auto &values = this->prev_inj_multipliers_[well->name()];
if (values.empty()) {
    values.assign(perf_data.size(), 1.0);
}
```

opm/simulators/wells/MultisegmentWell_impl.hpp
GitPaean: Let us remove seg from the interface.

GitPaean:
jekins build this please update_data please

GitPaean:
jekins build this update_data please

akva2:

bska (commented):
Thanks for the updates.  There is still one redundant string-based lookup in `prev_inj_multiplier_` and `unordered_map<>::operator[]` will **always** (attempt to) insert a new node into the map, but `BlackoilWellModelGeneric::initInjMult()` is probably called infrequently enough that it does not matter much.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1405,4 +1405,28 @@ int BlackoilWellModelGeneric::numLocalNonshutWells() const
...
+    for (auto& well : this->well_container_generic_) {
+        if (well->isInjector() && well->wellEcl().getInjMultMode() != Well::InjMultMode::NONE) {
+            const auto& ws = this->wellState().well(well->indexOfWell());
+            const auto& perf_data = ws.perf_data;
+
+            auto &values = this->prev_inj_multipliers_[well->name()];
+            if (values.empty()) {
+                values.assign(perf_data.size(), 1.0);
+            }
+            well->initInjMult(this->prev_inj_multipliers_.at(well->name()));
```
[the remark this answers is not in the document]
GitPaean: This is a good point. Will update after the `update_data` is done.

GitPaean:
> Thanks for the updates. There is still one redundant string-based lookup in `prev_inj_multiplier_` and `unordered_map<>::operator[]` will **always** (attempt to) insert a new node into the map, but `BlackoilWellModelGeneric::initInjMult()` is probably called infrequently enough that it does not matter much.

Are you referring to the following code? 
```c++
auto &values = this->prev_inj_multipliers_[well->name()];
if (values.empty()) {
    values.assign(perf_data.size(), 1.0);
}
```

I think we always need (at least) one lookup, what is the one redundant string-based loookup about? Thanks.

GitPaean:
jenkins build this opm-tests=1001 please

bska (commented):
Thanks a lot for the updates.  This looks good to me now and as the new reference solutions have been installed on the CI system I'll merge this now to activate the feature.

PR #4687, opened 2023-06-05
correct reservoir rate constraint calculation

plgbrts:
The `RESV` constraint was not honored specifying e.g.

 ```
GCONPROD 
-- if item 14 (RESV) is non-default this constraint should be honored (irrespective of item 7)
--   1     2      ORATE  WRATE GRATE LRATE    7	     8   ...      RESV           
  'PROD' 'LRAT'   1000. 1*     1.E6  2500.  'RATE' 'YES' 1* 1*  3* 800 
```

With the proposed change OPM-flow behaves the same as the commercial simulator, see figure below where also the RESV volume rate (orange line) is shown using the OPM master that clearly violates the constraint.
image


plgbrts:
jenkins build this please

atgeirr:
This looks like a very straightforward bugfix, and should be merged. Please indicate if you want to add anything else before merging. (I have made this a regular PR not a draft since it looked good enough to me.)

plgbrts:
I don't want to add anything to this PR. It can be merged.

PR #4688, opened 2023-06-06
Fix water evaporation initialization for two-phase gas-water system

plgbrts:
This PR fixes the issue of using a wrong (`GOC`) contact depth for water evaporation (`RVW`) initialization in the special case of a two-phase gas-water system. This issue was raised in [initialize trivial rvw if no table is given #4647](https://github.com/OPM/opm-simulators/pull/4647)


plgbrts:
jenkins build this please

totto82:
Thanks for the fix.

PR #4690, opened 2023-06-06
Enable Zoltan-Based In-Process Subdomain Partitioning

bska:
Tested only for sequential running (`mpirun -np 1`) at the moment. Intended to replace the original partitioning scheme based on a user-provided input file named `partition.txt`.

There are two new user-controlled runtime parameters

1. `NumLocalDomains` (`--num-local-domains`): Positive integer. The partitioner will divide the domain into this many non-overlapping subdomains and run the non-linear domain decomposition method on those

2. `LocalDomainsPartitioningImbalance` (`--local-domains-partitioning-imbalance`): Floating-point value 
 which controls Zoltan's imbalance tolerance. A value of 
 corresponds to a tolerance of three percent domain imbalance

These parameters are not queried unless the run uses the NLDD non-linear solver.

This PR replaces #4638 which I accidentally closed earlier, and in such a way that it could not be reopened.


bska:
I'm closing this now.  The upstream PR #3764 contains the Zoltan patch and can be used for separate testing.

PR #4691, opened 2023-06-06
Prepare for Revised Implementation of WBPn

bska:
This initial commit changes the API of the `CollectDataToIORank` class' handling of WBPn values from collecting a set cell pressures into communicating pre-computed WBPn values through the new `WellBlockAveragePressures` container class.  This is in preparation of moving the WBPn calculation to the simulator side for greater parallelism.  For now we do not compute any of the actual WBPn values.  That will be the subject of follow-up commits.

While here, also split a number of very long lines for readability.


akva2 (approved):
Must admit I didn't review all the cosmetic changes, but I think I got the main point (use the new containers for communication values across processes) and that looks fine.

bska:
jenkins build this opm-common=3557 please

bska:
Merging this per prior approval.

PR #4692, opened 2023-06-06
Add Container for Dynamic WBPn Source Values

bska:
This commit adds a new container class,
```
ParallelPAvgDynamicSourceData
```
which inherits from `PAvgDynamicSourceData` and provides a parallel view of source contributions.  Member function
```
collectLocalSources()
```
will call the user-provided source term evaluation function for each source location in its purview--typically those locations owned by the current MPI rank.  Those values will be distributed to other MPI ranks through member function `synchroniseSources()` which will fill the base class' `src_` data member, and become available to clients through read-only item spans.


bska:
I am creating this PR in draft mode because

1. It depends on, and contains, #4691.
2. It depends on upstream PR OPM/opm-common#3558.

I will keep the PR in draft mode until such time as it is ready for review.

bska:
The upstream PRs have both been merged into the master branch.  I'll mark this as "ready for review" and run a build check.

bska:
jenkins build this please

akva2 (approved):
Surprised to see new code with no tests exercising the class but I guess it will come in due time.

bska:
> Surprised to see new code with no tests exercising the class but I guess it will come in due time.

Maybe.  This is an MPI tool, so in order to test it I need an MPI context/communicator and a way to reliably and deterministically assign source locations to different ranks in the communicator.  If you have a suggestion for how to do that then I'll be happy to create a set of unit tests.

akva2:
See e.g. https://github.com/OPM/opm-simulators/blob/master/tests/test_broadcast.cpp and https://github.com/OPM/opm-simulators/blob/master/CMakeLists.txt#L366

bska:
> See e.g. [`tests/test_broadcast.cpp`](https://github.com/OPM/opm-simulators/blob/f47cceda634bbb5ac210811fa9e95ec3b1a51a15/tests/test_broadcast.cpp) and [line 366 of `CMakeLists.txt`](https://github.com/OPM/opm-simulators/blob/f47cceda634bbb5ac210811fa9e95ec3b1a51a15/CMakeLists.txt#L366).

Thanks a lot for the pointers.  I'll update the PR accordingly.

bska:
> Thanks a lot for the pointers. I'll update the PR accordingly.

I've pushed an update to include a unit test.

bska:
jenkins build this please

CMakeLists.txt
```diff
@@ -363,6 +363,54 @@ opm_add_test(test_parallelwellinfo_mpi
   NO_COMPILE
     )
 
+opm_add_test(test_parallel_wbp_sourcevalues_np1
```
akva2, line 366: okay did not know we want multiple tests. see how test_parallelwellinfo_mpi vs test_parallelwellinfo is done (ie standard CMakeLists_files.txt entry and then an opm_add_test for the parallel incarnations).

bska: > okay did not know we want multiple tests.

Neither did I until I actually started creating the test and realized that multiple tests with different number of ranks/processes might actually be useful...

> see how `test_parallelwellinfo_mpi` vs `test_parallelwellinfo` is done

Thanks for the hint.  I've pushed an update to this PR that mimics that pattern.

bska:
jenkins build this please

CMakeLists.txt
akva2: looks good. still approved.

bska: > looks good. still approved.

:+1:

Tangentially, does a "bare" [`opm_add_test()`](https://github.com/OPM/opm-simulators/blob/5cddb59b95ea1d7acb0ffe314e164333a7b6eccc/CMakeLists.txt#L377) call specifying a [runtime number of processes](https://github.com/OPM/opm-simulators/blob/5cddb59b95ea1d7acb0ffe314e164333a7b6eccc/CMakeLists.txt#L383) play well with the CTest `PROCESSORS` property used in PR #4641 to enable more test parallelism?  Do we risk oversubscribing the CI system's computational capacity when we have multiple tests that run MPI-aware executables without telling CTest how many processors they actually use?

It is **probably** not a big problem for this PR per se since each test runs in well under one second, but it maybe it's problem for tests that take more time to run?

akva2: if you specify PROCESSORS it is [set as a propery on the test](https://github.com/OPM/opm-common/blob/master/cmake/Modules/OpmSatellites.cmake#L361) and all is well.

bska: >> does a "bare" [`opm_add_test()`](https://github.com/OPM/opm-simulators/blob/5cddb59b95ea1d7acb0ffe314e164333a7b6eccc/CMakeLists.txt#L377) call specifying a [runtime number of processes](https://github.com/OPM/opm-simulators/blob/5cddb59b95ea1d7acb0ffe314e164333a7b6eccc/CMakeLists.txt#L383) play well with the CTest `PROCESSORS` property used in PR #4641 to enable more test parallelism?
>
> if you specify `PROCESSORS` it is [set as a propery on the test](https://github.com/OPM/opm-common/blob/master/cmake/Modules/OpmSatellites.cmake#L361) and all is well.

Noted.  I've pushed an update to set `PROCESSORS` on the new tests as well.  If the build check goes through I will merge this PR in its current state.

I will make a separate, follow-up PR to set proper `PROCESSORS` values on the other MPI tests which use a bare `opm_add_test()` macro call.

bska:
jenkins build this please

CMakeLists.txt
```diff
@@ -363,6 +363,58 @@ opm_add_test(test_parallelwellinfo_mpi
...
```
akva2, line 366: sorry for missing this earlier. this test will be a duplicate. the one processor one is added as a normal test from the CMakeLists_files.txt entry.

bska: > sorry for missing this earlier. this test will be a duplicate.

Ah, yes.  You're right.  I'll amend.

bska:

bska:
Build check is green.  Merging this now per prior approval.

PR #4693, opened 2023-06-06
Add Parallel Calculation Support for WBPn/WPAVE

bska:
This PR adds a parallel calculation object derived from the serial `PAvgCalculator` class.  This parallel version is aware of MPI communicators and knows how to aggregate contributions from wells that might be distributed across ranks.

We also add a wrapper class, `ParallelWBPCalculation`, which knows how to exchange information from `PAvgCalculatorCollection` objects on different ranks and, especially, how to properly prune inactive cells/connections.

### DIFF
--- a/opm/simulators/wells/ParallelWBPCalculation.hpp
+++ b/opm/simulators/wells/ParallelWBPCalculation.hpp
@@ -1,0 +1,368 @@
+/*
+  Copyright 2023 Equinor ASA.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef PARALLEL_WBP_CALCULATION_HPP
+#define PARALLEL_WBP_CALCULATION_HPP
+
+#include <opm/simulators/wells/ParallelPAvgDynamicSourceData.hpp>
+
+#include <opm/simulators/utils/ParallelCommunication.hpp>
+
+#include <opm/simulators/wells/PerforationData.hpp>
+
+#include <opm/input/eclipse/Schedule/Well/PAvgCalculator.hpp>
+#include <opm/input/eclipse/Schedule/Well/PAvgCalculatorCollection.hpp>
+
+#include <cstddef>
+#include <functional>
+#include <memory>
+#include <vector>
+
+namespace Opm {
+    class GridDims;
+    class ParallelWellInfo;
+    class PAvg;
+    class Well;
+}
+
+namespace Opm {
+
+/// Parallel facility for managing the on-rank collection and global
+/// distribution of WBPn source values as well as local calculation and
+/// distributed reduction of the inferred WBPn report values.
+class ParallelWBPCalculation
+{
+public:
+    /// Callback for inferring the source locations which are active on the
+    /// current MPI rank.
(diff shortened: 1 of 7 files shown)


bska:
I am creating this PR in draft state for two reasons:

1. It depends on, and contains, PR #4692.
2. It requires the revised `PAvgCalculator` and `PAvgCalculatorCollection` classes from upstream PR OPM/opm-common#3559.

I will keep the PR in draft state until such time as it is ready for review.

bska:
There will be additional updates here because I'm in the process of adding more documentation.  However, in the interest of expediting the process I'm nevertheless going to mark this as "ready for review" now.

bska:
jenkins build this please

akva2:
maybe it's testable?

bska:
> maybe it's testable?

I think so, and I'll try to add a test in the next update.

bska:
> > maybe it's testable?
> 
> I think so, and I'll try to add a test in the next update.

I have pushed an update which includes a simple unit test.  It's a small excerpt from the unit test introduced in OPM/opm-common#3559, but now using two MPI ranks and splitting the $5 \times 5 \times 10$ grid into a top and bottom half of $5 \times 5 \times 5$ cells each.  This also demonstrates support for distributed wells since each half gets three connections each from the well.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
> jenkins build this please

The new unit test [fails the check for number of processes](https://github.com/OPM/opm-simulators/blob/4bdbf928a8b0e533820f29a4e7e48641be18cf30/tests/test_parallel_wbp_calculation.cpp#L554) due to being run in serial mode.  @akva2: Is there a way to just **compile** the test binary, when MPI support exists, but **not** run it in serial mode?

akva2:
Remove from CMakeLists_files.txt and use
```
opm_add_test(test_parallel_wbp_calculation
              SOURCES
                tests/test_parallel_wbp_calculation.cpp
              LIBRARIES
                opmsimulators ${Boost_UNIT_TEST_FRAMEWORK_LIBRARY}
              CONDITION
               MPI_FOUND AND Boost_UNIT_TEST_FRAMEWORK_FOUND
              ONLY_COMPILE)
```

bska:
> ```cmake
> opm_add_test(test_parallel_wbp_calculation
> ...
>               ONLY_COMPILE)
> ```
>
> this conditionally builds the test executable but does not add any test for it.

Ah, right.  Thanks a lot for the hint!  I'll update the PR accordingly.

bska:
jenkins build this please

opm/simulators/wells/ParallelWBPCalculation.hpp
```diff
@@ -0,0 +1,368 @@
...
+    ///   according to \code well.getConnections().size() \endcode, but with
+    ///   a non-negative entry only for those active connections that
+    ///   intersect the current rank.  If \code localConnIdx[i] == j
+    ///   \endcode, then the \c i-th global connection is the \c j-th active
+    ///   connection on the current rank.  Use a negative value to identify
+    ///   a connection that is either not flowing or which does not
+    ///   intersect the current MPI rank.
+    ///
+    /// \param[in] makeWellSourceEvaluator Factory function to support
+    ///   deferred creation of an evaulation function for the per-connection
```
akva2, line 107: evaluation

opm/simulators/wells/ParallelWBPCalculation.hpp
```diff
@@ -0,0 +1,368 @@
...
+        /// \param[in] eval Source term evaluation function.
+        ///
+        /// \return \code *this \endcode
+        SourceData& evaluator(Evaluator eval);
+
+        /// Assign evaluation function factory for computing the on-rank,
+        /// WBPn source terms.
+        ///
+        /// \param[in] evalFactory Factory function to support deferred
+        ///   creation of an evaulation function for the per-connection WBP
```
akva2, line 247: evaluation

akva2 (approved):
Two doxy typos. Looks good to me.

bska:
> Two doxy typos

Thanks.  I'll correct those.

> Looks good to me.

:+1:

bska:
jenkins build this please

bska:
jenkins build this please

bska:
Build check is green and PR approved.  Merging this into the master branch.

PR #4694, opened 2023-06-06
Hook New WBPn Calculation Up to Well Model

bska:
This PR activates the support for calculating WBPn summary result values per well in parallel.  To affect the calculation we add two new data members in BlackoilWellModelGeneric:

  - `conn_idx_map_`:  Maps well's connection index (`0..getConnections().size() - 1`) to connections on current rank.  Its `local()` connections are negative 1 (`-1`) if the connection is not on current rank, and a non-negative value otherwise.  The `global()` function maps well connections on current rank to global connection ID for each well.  Effectively the reverse of `local()`.  Finally, the `open()` function maps well connections on current rank to open/flowing connections on current rank. Negative 1 if connection is not flowing.

  - `wbpCalculationService`: Parallel collection of WBPn calculation objects that knows how to exchange source and result information between all ranks in a communicator.  Also handles distributed wells.

We furthermore need a way to compute connection-level fluid mixture density values.  For the standard well class we add a way to access the `StandardWellConnection`'s `perf_densities_` values.  However, since these are defined for open/flowing connections only, this means we're not able to fully meet the requirements of the
```
WELL/ALL
```
`WPAVE` depth correction procedure for standard wells.  The multi-segmented well type, on the other hand, uses the fluid mixture density in the associated well segment and is therefore well defined for `ALL` connections.  `OPEN` well connections are supported for both well types.

PR #4695, opened 2023-06-06
Declare Support for WPAVE/WBPn

bska:
We emit a warning if the model uses connection flag `ALL`, but continue the run.  This behaviour is still being debated and we may decide to halt the run in this situation.


bska:
I'm creating this PR in draft mode because it depends on, and contains, PR #4694.  I will keep this PR in draft state until such time as it is ready for review.

blattms (approved):

blattms:
jenkins build this please

bska:
Merging per prior approval.

PR #4697, opened 2023-06-08
Add support item 7 NONE in GCONPROD

totto82:
Note that this may depend on https://github.com/OPM/opm-common/pull/3562 to be useful.


totto82:
jenkins build this please

totto82:
Sorry @plgbrts I didn't notice https://github.com/OPM/opm-simulators/pull/4658

I guess we had the same issue... 

I had to do some more work to make it to work.

PR #4703, opened 2023-06-13
Chase Upstream API Update

bska:
The `eval_assign()` and `eval()` member functions of `UDQConfig` now take a `const Schedule&` parameter.


bska:
Merging this now to get consistent function calls.

PR #4705, opened 2023-06-13
Add guard against invalid satnum, pvtnum, imbnum, eqlnum 

totto82:


totto82:
jenkins build this please

totto82:
jenkins build this please

ebos/eclgenericproblem.cc
```diff
@@ -374,7 +374,11 @@ updateNum(const std::string& name, std::vector<T>& numbers)
     unsigned numElems = gridView_.size(/*codim=*/0);
     numbers.resize(numElems);
     for (unsigned elemIdx = 0; elemIdx < numElems; ++elemIdx) {
-        numbers[elemIdx] = static_cast<T>(numData[elemIdx]) - 1;
+        if (numData[elemIdx] > 0) {
+            numbers[elemIdx] = static_cast<T>(numData[elemIdx]) - 1;
+        } else {
+            throw std::runtime_error("zero or negative values provided for region array: " + name);
+        }
```
bska, line 381: Do we handle the case of one rank/process throwing an exception here, but other ranks continuing?  What if `numData[elemIdx] - 1` is greater than (or equal to) the size of whatever region table it's being used to index?

bska (commented):
Did you verify that this approach works in parallel?

ebos/eclgenericproblem.cc
totto82: > Do we handle the case of one rank/process throwing an exception here, but other ranks continuing?


I guess not. If handling means that you get a nice error message. 

You will get something like this. 

```
*** An error occurred in MPI_Allreduce
*** reported by process [1191968769,14]
*** on communicator MPI_COMM_WORLD
*** MPI_ERR_TRUNCATE: message truncated
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
```

I think this is align with many user input error that will make the simulator throw. Do you want to add a layer of communication?


> What if numData[elemIdx] - 1 is greater than (or equal to) the size of whatever region table it's being used to index?

I guess that make sense to add a check for that as well. Current code will seg fault if you put in a satnum > number of tables.

blattms: This error message will happen at a comletely other place in the simulator (because different processes are doing different communications that are interfering. It might also be dependent on the MPI implementation used. I can imagine other implementations deadlocking. So yes, we need to do something about this,

bska: > We need to make sure that if there is a problem for one element then all processes will throw.

Yeah, we need better error handling in parallel.  For that matter, we also really need a human readable diagnostic message.  Something referring to aspects of an MPI operation is not usable for the engineer.

blattms: Something similar to https://github.com/OPM/opm-simulators/blob/ed591239a7371d972f496b36b0e610929efe79aa/opm/simulators/wells/BlackoilWellModel_impl.hpp#L148

Put `OPM_BEGIN_PARALLEL_TRY_CATCH();` before and `OPM_END_PARALLEL_TRY_CATCH(message, ebosSimulator_.gridView().comm());` after loop. We need to make sure that there no other nested ones inside

totto82:
jenkins build this please

ebos/eclgenericproblem.cc
```diff
@@ -373,8 +373,15 @@ updateNum(const std::string& name, std::vector<T>& numbers)
 
     unsigned numElems = gridView_.size(/*codim=*/0);
     numbers.resize(numElems);
+    const auto num_regions = eclState_.getTableManager().getTabdims().getNumSatTables();
```
bska, line 376: Do we ever call `updateNum()` on anything other than `SATNUM` or `IMBNUM`?  Because if we do, this is the wrong upper bound.

bska (commented):
Either we have a really restricted set of region arrays, or you're missing essential parts of the problem.  If it were me I think I'd move this check to a location that provides more context in order to get better detection and diagnostic messages.

totto82:
jenkins build this please

ebos/eclgenericproblem.cc
```diff
@@ -374,54 +374,66 @@ updateNum(const std::string& name, std::vector<T>& numbers)
     unsigned numElems = gridView_.size(/*codim=*/0);
     numbers.resize(numElems);
     for (unsigned elemIdx = 0; elemIdx < numElems; ++elemIdx) {
-        numbers[elemIdx] = static_cast<T>(numData[elemIdx]) - 1;
+        if (numData[elemIdx] > num_regions) {
```
bska, line 377: Would you mind `static_cast<>`-ing `num_regions` into an `int`, please?  That way we don't get additional [signed vs. unsigned comparison](https://ci.opm-project.org/job/opm-simulators-PR-builder/5040/gcc/folder.3106625/source.93ca4d50-d5bf-4f86-adea-1c0ace1af6c9/#377) diagnostics.

ebos/equil/initstateequil_impl.hh
```diff
@@ -1278,6 +1279,15 @@ equilnum(const EclipseState& eclipseState,
         const auto& e = eclipseState.fieldProps().get_int("EQLNUM");
         std::transform(e.begin(), e.end(), eqlnum.begin(), [](int n){ return n - 1;});
     }
+    OPM_BEGIN_PARALLEL_TRY_CATCH();
+    const auto num_regions = eclipseState.getTableManager().getEqldims().getNumEquilRegions();
```
bska, line 1283: Would you mind `static_cast<>`-ing `num_regions` into an `int`, please?  That way we don't get any additional [signed vs. unsigned comparison](https://ci.opm-project.org/job/opm-simulators-PR-builder/5040/gcc/folder.1194549214/source.f8f54d44-4296-4b6f-93aa-9161ed4b2ff4/#1284) diagnostics.

bska (commented):
Thanks for the updates.  This is starting to look good to me.  I'm not **completely** convinced that this is the best place to put the check, but it's absolutely better than nothing.  Just a couple of minor lingering issues over "signed vs. unsigned comparison" warnings and then I'll merge.

totto82:
jenkins build this please

bska (approved):

PR #4707, opened 2023-06-14
Equilibration with PPCWMAX

svenn-t:
This implements changes accommodating PPCWMAX when applying SWATINIT in equilibration, in particular the re-calculation of initial water saturation.

Depends on https://github.com/OPM/opm-common/pull/3570

### DIFF
--- a/ebos/equil/initstateequil.hh
+++ b/ebos/equil/initstateequil.hh
@@ -544,7 +544,7 @@ private:
     /// \param[in] pcow O/W capillary pressure value (Po - Pw).
     ///
     /// \return Water saturation value.
-    std::tuple<double, bool> applySwatInit(const double pcow);
+    std::pair<double, bool> applySwatInit(const double pcow);
 
     /// Derive water saturation from SWATINIT data.
     ///
@@ -558,7 +558,7 @@ private:
     ///
     /// \return Water saturation value.  Input value, possibly mollified by
     ///    current set of material laws.
-    std::tuple<double, bool> applySwatInit(const double pc, const double sw);
+    std::pair<double, bool> applySwatInit(const double pc, const double sw);
 
     /// Invoke material law container's capillary pressure calculator on
     /// current fluid state.
--- a/ebos/equil/initstateequil.hh
+++ b/ebos/equil/initstateequil.hh
@@ -544,7 +544,7 @@ private:
     /// \param[in] pcow O/W capillary pressure value (Po - Pw).
     ///
     /// \return Water saturation value.
-    double applySwatInit(const double pcow);
+    std::tuple<double, bool> applySwatInit(const double pcow);
 
     /// Derive water saturation from SWATINIT data.
     ///
@@ -558,7 +558,7 @@ private:
     ///
     /// \return Water saturation value.  Input value, possibly mollified by
     ///    current set of material laws.
-    double applySwatInit(const double pc, const double sw);
+    std::tuple<double, bool> applySwatInit(const double pc, const double sw);
 
     /// Invoke material law container's capillary pressure calculator on
     /// current fluid state.
--- a/ebos/equil/initstateequil_impl.hh
+++ b/ebos/equil/initstateequil_impl.hh
@@ -694,9 +694,16 @@ void PhaseSaturations<MaterialLawManager, FluidSystem, Region, CellID>::deriveWa
         // cell.
         const auto pcow = this->press_.oil - this->press_.water;
 
-        sw = this->swatInit_.empty()
-            ? this->invertCapPress(pcow, this->waterPos(), isIncr)
-            : this->applySwatInit(pcow);
+        if (this->swatInit_.empty()) {
+            sw = this->invertCapPress(pcow, this->waterPos(), isIncr);
+        }
+        else {
+            auto [swout, newSwatInit] = this->applySwatInit(pcow);
+            if (newSwatInit == true)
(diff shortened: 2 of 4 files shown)


totto82:
Thanks. I suggest you add the ppcw test in opm-tests for regression in this PR.

ebos/equil/initstateequil_impl.hh
```diff
@@ -846,18 +859,21 @@ accountForScaledSaturations()
...
 
 template <class MaterialLawManager, class FluidSystem, class Region, typename CellID>
-double PhaseSaturations<MaterialLawManager, FluidSystem, Region, CellID>::
+std::tuple<double, bool>
+PhaseSaturations<MaterialLawManager, FluidSystem, Region, CellID>::
 applySwatInit(const double pcow)
 {
-    return this->applySwatInit(pcow, this->swatInit_[this->evalPt_.position->cell]);
+    auto [swout, newSwatInit] = this->applySwatInit(pcow, this->swatInit_[this->evalPt_.position->cell]);
+    return {swout, newSwatInit};
```
bska, line 867: Is there a reason you can't use the original implementation here?  You're just forwarding to a different function with the same return type.

ebos/equil/initstateequil_impl.hh
```diff
@@ -718,7 +725,13 @@ fixUnphysicalTransition()
...
         // seems consistent with ECLIPSE, but fails to honour SWATINIT in
         // case of non-trivial gas/oil capillary pressure.
-        sw = this->applySwatInit(pcgw, sw);
+        auto [swout, newSwatInit] = this->applySwatInit(pcgw, sw);
+        if (newSwatInit){
+            const auto isIncr = false; // dPcow/dSw <= 0 for all Sw.
+            sw = this->invertCapPress(pcgw, this->waterPos(), isIncr);
+        }
+        else
+            sw = swout;
```
bska, line 734: Please keep the braces, even on single-statement blocks.

ebos/equil/initstateequil_impl.hh
```diff
@@ -846,18 +859,21 @@ accountForScaledSaturations()
...
 template <class MaterialLawManager, class FluidSystem, class Region, typename CellID>
-double PhaseSaturations<MaterialLawManager, FluidSystem, Region, CellID>::
+std::tuple<double, bool>
+PhaseSaturations<MaterialLawManager, FluidSystem, Region, CellID>::
 applySwatInit(const double pcow, const double sw)
 {
-    return this->matLawMgr_
-        .applySwatinit(this->evalPt_.position->cell, pcow, sw);
+    auto [swout, newSwatInit] =this->matLawMgr_.applySwatinit(this->evalPt_.position->cell, pcow, sw);
+    return {swout, newSwatInit};
```
bska, line 876: Why do you need to change the implementation (body) here?  You're just calling a function with same return type (assuming `Scalar = double`).

ebos/equil/initstateequil_impl.hh
```diff
@@ -694,9 +694,16 @@ void PhaseSaturations<MaterialLawManager, FluidSystem, Region, CellID>::deriveWa
...
 
-        sw = this->swatInit_.empty()
-            ? this->invertCapPress(pcow, this->waterPos(), isIncr)
-            : this->applySwatInit(pcow);
+        if (this->swatInit_.empty()) {
+            sw = this->invertCapPress(pcow, this->waterPos(), isIncr);
+        }
+        else {
+            auto [swout, newSwatInit] = this->applySwatInit(pcow);
+            if (newSwatInit == true)
```
bska, line 702: Any reason not to write this as
```C++
if (newSwatInit) {
    // ...
}
else {
    // ...
}
```
instead?  Comparing a `bool` directly to `true` is not something we usually do in this code base.

ebos/equil/initstateequil.hh
```diff
@@ -544,7 +544,7 @@ private:
     /// \param[in] pcow O/W capillary pressure value (Po - Pw).
     ///
     /// \return Water saturation value.
-    double applySwatInit(const double pcow);
+    std::tuple<double, bool> applySwatInit(const double pcow);
```
bska, line 547: Why `tuple<>` instead of `pair<>`?

bska (commented):
I think the implementation here looks good for the most part, but there are a couple of minor issues that I would like you to explain.

ebos/equil/initstateequil_impl.hh
svenn-t: Think I got some errors here at some point and made the changes. But they are unnecessary I see now, so I reverted back to the original implementation.

ebos/equil/initstateequil_impl.hh
svenn-t: I have changed this.

ebos/equil/initstateequil_impl.hh
svenn-t: Same as the above, got some errors and made changes. Reverted back to the original implementation.

ebos/equil/initstateequil.hh
svenn-t: I saw some other code using `tuple` so I used it as well. Honestly didn't know about `pair`, but that's more logical! I have changed it here and in the opm-common PR.

bska:
Thanks a lot for the updates. I'll merge this into the master branch to activate the support for this feature.

PR #4710, opened 2023-06-15
Wag hysteresis

totto82:
This PR replaces. https://github.com/OPM/opm-simulators/pull/4583

Depends on https://github.com/OPM/opm-common/pull/3542


bska:
Merging this to activate the feature.  Thanks a lot for bringing this in.

PR #4713, opened 2023-06-16
start regression testing of MULTFLT-01

tskille:
Support of MULTFLT both in GRID and EDIT section was implemented correctly with 

https://github.com/OPM/opm-common/pull/3556

Simulation results has been verified today and they are good. Hence, I would like to start regression testing with this model.


bska:

bska:
jenkins build this opm-tests=996 please

PR #4720, opened 2023-06-23
Fix  rv initialization

totto82:
Round off errors in satOil leads to using saturated rv instead of rv from depth.


bska:
Is there more work coming here?

totto82:
I added similar guards to other parts of the code.

totto82:
jenkins build this please

totto82:
jenkins build this please

bska (approved):
Looks good to me and I'll merge into master.

PR #4724, opened 2023-06-27
Some janitoring

akva2:
Quell some warnings.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -887,6 +887,7 @@ namespace Opm
         auto obtain = [this](const Eval& value)
                       {
                           if constexpr (std::is_same_v<Value, Scalar>) {
+                              static_cast<void>(this); // suppress clang warning
                               return getValue(value);
                           } else {
                               return this->extendEval(value);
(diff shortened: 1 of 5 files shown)


akva2:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -887,6 +887,7 @@ namespace Opm
         auto obtain = [this](const Eval& value)
                       {
                           if constexpr (std::is_same_v<Value, Scalar>) {
+                              (void)this; // suppress clang warning
```
bska, line 890: Would you mind writing `static_cast<>()` please?  That stands out more clearly when reading the code.

bska (commented):
Much appreciated.  I'd only like to see us use C++ casts instead of C-style casts.  Both because they are more obvious when reading the code and because they're more well behaved.

opm/simulators/wells/MultisegmentWell_impl.hpp
akva2: sure.

akva2:
jenkins build this please

bska (approved):
Perfect!  Thanks a lot for the updates.  Please merge when green.

PR #4525, opened 2023-03-13
keywords co2store + gaswat

goncalvesmachadoc:
Preparation for https://github.com/OPM/opm-tests/pull/902

PR #4725, opened 2023-06-28
Add CO2STORE_GW and CO2STORE_GASWAT to regression tests

totto82:
This supersede an earlier PR that includes the CO2STORE_GASWAT test 
https://github.com/OPM/opm-simulators/pull/4525
that was never merged. The code change seems to have been merged in other PRs.
 
The two tests are similar but test different input setup. 
Especially the CO2STORE_GW test salinity which recently was broken and should be fixed by 

https://github.com/OPM/opm-common/pull/3574

My plan is to frist merge these tests. Then rerun with https://github.com/OPM/opm-common/pull/3574. The test should then fail and we can update the reference with bugfix.

### DIFF
--- a/regressionTests.cmake
+++ b/regressionTests.cmake
@@ -509,6 +509,20 @@ add_test_compareECLFiles(CASENAME co2store
                          REL_TOL ${rel_tol}
                          DIR co2store)
 
+add_test_compareECLFiles(CASENAME co2store_gw
+                         FILENAME CO2STORE_GW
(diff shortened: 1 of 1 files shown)


totto82:

totto82:
jenkins build this opm-tests=1004 please

akva2:
not the best of naming going on here (what distinguishes GW from GASWAT), but that's nothing holding this PR back. thanks for improving the coverage.

PR #4726, opened 2023-06-28
Increase supported equil regions from 255 to 65525

blattms:
Previously we stored the equil region idx as unsigned short which imposed this limit. Apparently there models that use more regions than allowed. Hence we now store the idx using unsigned short which allows 65535 regions.

Still there is a caveat in the code: The region interaction is stored as a dense matrix. This means that using the maximum allowed number of regions now would consume 32 GB of memory. In the future we actually store the interactions in a sparse matrix to get rid of this memory requirement. For now we just issue a warning if there are more than 2048 regions used.

### DIFF
--- a/ebos/eclgenericthresholdpressure.cc
+++ b/ebos/eclgenericthresholdpressure.cc
@@ -99,8 +99,8 @@ thresholdPressure(int elem1Idx, int elem2Idx) const
     }
 
     // threshold pressure accross EQUIL regions
-    unsigned short equilRegion1Idx = elemEquilRegion_[elem1Idx];
-    unsigned short equilRegion2Idx = elemEquilRegion_[elem2Idx];
+    auto equilRegion1Idx = elemEquilRegion_[elem1Idx];
+    auto equilRegion2Idx = elemEquilRegion_[elem2Idx];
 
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please

akva2:
jenkins build this please

ebos/eclgenericthresholdpressure.cc
```diff
@@ -133,6 +133,14 @@ finishInit()
                                maxRegions)));
     }
 
+    if (numEquilRegions_ > 2048) {
+        // warn about performance
+        OpmLog::warning(fmt::format("Number of equil regions is {}. This larger "
+                                 "than 2048. Note, that this might "
+                                 "might have a negative impact on performance "
+                                 "and memory consumption", numEquilRegions_));
+    }
```
akva2, line 142: read read and fix fix this message message please please. I'd just remove the 'This larger than 2048'. While you chose that as the warning limit, there is nothing significant about the particular value as such.

blattms: Thanks for the hint. Improved.

blattms:
jenkins build this please

akva2 (approved):

PR #4729, opened 2023-06-29
PartitionCells: move all related code to compile unit

akva2:
This moves some of the partitionCells code from BlackoilModelEbos into partitionCells.cpp. Now all this code is in one compile unit. 

Also removes the indent for namespace which afaik is what we want.

### DIFF
--- a/opm/simulators/flow/partitionCells.cpp
+++ b/opm/simulators/flow/partitionCells.cpp
@@ -44,140 +44,139 @@
 
 namespace {
 
-    std::pair<std::vector<int>, int>
-    countDomains(std::vector<int> partition_vector)
-    {
-        auto maxPos = std::max_element(partition_vector.begin(),
-                                       partition_vector.end());
+std::pair<std::vector<int>, int>
+countDomains(std::vector<int> partition_vector)
+{
+    auto maxPos = std::max_element(partition_vector.begin(),
+                                   partition_vector.end());
 
-        const auto num_domains = (maxPos == partition_vector.end())
-            ? 0 : *maxPos + 1;
+    const auto num_domains = (maxPos == partition_vector.end())
+        ? 0 : *maxPos + 1;
 
-        return { std::move(partition_vector), num_domains };
-    }
+    return { std::move(partition_vector), num_domains };
+}
 
-    template <typename, class = void>
-    struct HasZoltanPartitioning : public std::false_type {};
+template <typename, class = void>
+struct HasZoltanPartitioning : public std::false_type {};
 
-    template <typename GridType>
-    struct HasZoltanPartitioning<
-        GridType,
-        std::void_t<decltype(std::declval<const GridType&>().zoltanPartitionWithoutScatter
-                             (std::declval<const std::vector<Opm::Well>*>(),
-                              std::declval<const double*>(),
-                              std::declval<const int>(),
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

opm/simulators/flow/partitionCells.cpp
```diff
@@ -29,46 +41,141 @@
...
-        std::ifstream is(filename);
-        const std::vector<int> cellpart{std::istream_iterator<int>(is), std::istream_iterator<int>()};
-        if (cellpart.size() != size_t(num_cells)) {
-            auto msg = fmt::format("Partition file contains {} entries, but there are {} cells.",
-                                   cellpart.size(), num_cells);
-            throw std::runtime_error(msg);
+template <typename GridType>
+struct HasZoltanPartitioning<
+    GridType,
+    std::void_t<decltype(std::declval<GridType>().zoltanPartitionWithoutScatter
```
bska, line 64: You'll strictly speaking need `<type_traits>` for `std::void_t`.

bska (commented):
Except for one missing header I think this is good.  I'll leave final approval to @atgeirr, though.

akva2:
jenkins build this please

opm/simulators/flow/partitionCells.cpp
bska: > indeed, fixed.

Cool.  I did overlook one other issue, though.  The way we're using `zoltanPartitionWithoutScatter()`, we'll have to be a little be more careful here.  We're passing `GridType` directly, but the parameter type is `const GridType&`.  The two parameter types must match lest we get build failures that are hard to analyse in case the `GridType` only has a mutating `zoltanPartitionWithoutScatter()` member function.  The easiest fix is probably to write
```C++
std::void_t<decltype(std::declval<const GridType&>().zoltanPartitionWithoutScatter
```
instead.

akva2: sure, mutable "safety" added.

akva2:
jenkins build this please

opm/simulators/flow/partitionCells.hpp
```diff
@@ -26,14 +26,28 @@
...
-    /// Read a partitioning from file, assumed to contain one number per cell, its partition number.
-    /// \return pair containing a partition vector (partition number for each cell), and the number of partitions.
-    std::pair<std::vector<int>, int> partitionCellsFromFile(const std::string& filename, const int num_cells);
-
-    /// Trivially simple partitioner assigning partitions en bloc, consecutively by cell index.
-    /// \return pair containing a partition vector (partition number for each cell), and the number of partitions.
-    std::pair<std::vector<int>, int> partitionCellsSimple(const int num_cells, const int num_domains);
+class Well;
+
+template<class Grid>
```
atgeirr, line 31: Please add a little documentation for this function, as well as for partitionCellsZoltan(), now that they will have wider exposure.

atgeirr (approved):
Good to merge, but prefer a few lines of doc as indicated.

opm/simulators/flow/partitionCells.hpp
akva2: done.

akva2:
jenkins build this please

akva2:
merging per approval.

PR #4730, opened 2023-06-30
BlackoilModelEbos: some janitoring

akva2:
Extracted the hopefully non-controversial parts from https://github.com/OPM/opm-simulators/pull/4509


akva2:
jenkins build this please

opm/simulators/flow/BlackoilModelEbos.hpp
```diff
@@ -52,24 +51,21 @@
...
 
 #include <algorithm>
 #include <cassert>
 #include <cmath>
 #include <iomanip>
-#include <iostream>
 #include <limits>
 #include <type_traits>
-#include <utility>
+#include <tuple>
```
bska, line 68: To be honest I think you need `<tuple>` **in addition** to `<utility>` here.  The file uses `std::move()` in a number of places and that function (cast) is defined in `<utility>`.  Moreover, PR #3764 which was merged in the interim has explicit uses of `std::pair<>`, also from `<utility>`.

bska (commented):
Looks good to me, except for switching `<utility>` out in favour of `<tuple>` in `BlackoilModelEbos.hpp`.

opm/simulators/flow/BlackoilModelEbos.hpp
akva2: yeah, situation has changed. so i simply made it include tuple in addition.

akva2:
jenkins build this please

bska (approved):

PR #4732, opened 2023-06-30
-- added need functions for element context for thermal

hnil:
Functions need to make intensive quantities independent of element cache. Use is in opm-flowexperimental, for now but will be moved to simulators when full tpfa linearizing of energy is done.


blattms:
This needs a rebase.

bska:
jenkins build this please

hnil:
Closing: this targets eclproblem.hh, which no longer exists after the Ebos rename, so there is nothing left to rebase. If the thermal element-context functions are still wanted they would have to be written against FlowProblem from scratch. Branch retained on my fork.

PR #4733, opened 2023-06-30
fixed: parallel super builds

akva2:
commit e5e7ff7287 introduced a dependency on a generated header in a header file. this is problematic with super-builds as there is no explicit dependency for the simulator objects to these generated headers.

with the ninja-generator it is smart enough to figure out this across the subdirectories, but the make generator is not. hence we explicitly add a dependency on the opmcommon target in this case. ideally we would only depend on the generated header to allow compiling opmcommon and simulator objects in parallel. however there is as far as i know no way to depend on OUTPUT targets across subdirectories.


akva2:
jenkins build this please

bska (approved):

PR #4736, opened 2023-07-03
simplify cmake code for tuning tests

akva2:
add a post command to the runSimulator script and use this to execute the unit test at the end of the run


akva2:
jenkins build this please

PR #4737, opened 2023-07-03
BlackoilModeEbos: Reduce some code duplication

akva2:
Separate out code shared by newton and nldd into separate methods.


akva2:
jenkins build this please

bska (approved):
Great! Any simplification we can do here is most welcome.  I'll merge into master.

PR #4738, opened 2023-07-03
BlackoilModelEbos: improve domain ordering

akva2:
Put in separate method and add command line parameter to choose between pressure and residual.


akva2:
jenkins build this please

PR #4739, opened 2023-07-03
BlackoilModelEbos: improve nldd solver approach

akva2:
Separate out some more code into separate methods to improve higher-level readability.
Add enumeration of solver approach, and command line validation. In particular we get rid of an assert
on user provided data.

Sits on top of https://github.com/OPM/opm-simulators/pull/4738

### DIFF
--- a/opm/simulators/flow/BlackoilModelEbos.hpp
+++ b/opm/simulators/flow/BlackoilModelEbos.hpp
@@ -545,44 +545,8 @@ namespace Opm {
                     solveDomainJacobi(solution, locally_solved, local_report,
                                       iteration, timer, domain);
                 } else {
-                    assert(param_.local_solve_approach_ == "gauss-seidel");
-                    auto initial_local_well_primary_vars = wellModel().getPrimaryVarsDomain(domain);
-                    auto initial_local_solution = Details::extractVector(solution, domain.cells);
-                    auto res = solveDomain(domain, timer, iteration);
-                    local_report = res.first;
-                    if (!local_report.converged) {
-                        // We look at the detailed convergence report to evaluate
-                        // if we should accept the unconverged solution.
-                        const auto& convrep = res.second;
-                        // We do not accept a solution if the wells are unconverged.
-                        if (!convrep.wellFailed()) {
-                            // Calculare the sums of the mb and cnv failures.
-                            double mb_sum = 0.0;
-                            double cnv_sum = 0.0;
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this please

opm/simulators/flow/BlackoilModelEbos.hpp
```diff
@@ -1794,7 +1749,12 @@ namespace Opm {
             const auto& solution = ebosSimulator().model().solution(0);
 
             std::vector<int> domain_order(domains_.size());
-            if (param_.local_solve_approach_ == "gauss-seidel") {
+            switch (param_.local_solve_approach_) {
+            case DomainSolveApproach::Jacobi: {
+                std::iota(domain_order.begin(), domain_order.end(), 0);
+                break;
+            }
+            case DomainSolveApproach::GaussSeidel: {
```
bska, line 1757: So this isn't **completely** equivalent to the original code.  That would return `[0, 1, ..., n-1]` for anything other than Gauss&ndash;Seidel.  This revised logic will return `[0, 1, ..., n-1]` only for the Jacobi method, and will return `[0, 0, ..., 0]` for anything else.  Maybe we should `throw` for unknown `local_solve_approach_` values?

bska (commented):
Apart from a potential semantic change to the domain order function, this looks good to me.

opm/simulators/flow/BlackoilModelEbos.hpp
akva2: it's an extensive switch.

bska: And by "anything else" I mean anything else that's **not** Gauss&ndash;Seidel.  Sorry for the confusion.

bska: > it's an extensive switch.

It's not.  That's one of the surprises of `enum class`.  The actual enumerator values are all of the values of the underlying integer types.

akva2: well okay, but it's still the only two values that can be assigned to the variable, we check and throw earlier in the process.

akva2: added some default labels to retain original behavior (the other block had the opposite logic...)

akva2:
jenkins build this please

PR #4745, opened 2023-07-07
FilterCake: put code in separate class

akva2:
More separation of concerns.


akva2:
jenkins build this please

opm/simulators/wells/WellFilterCake.cpp
```diff
@@ -0,0 +1,129 @@
+/*
+  Copyright 2017 SINTEF Digital, Mathematics and Cybernetics.
+  Copyright 2017 Statoil ASA.
```
GitPaean, line 3: Maybe the copyright claim can stay  with `  Copyright 2023 Equinor.`

akva2: of course, didn't know what to put since this code is so new. changed.

GitPaean:
jenkins build this please

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1422,26 +1423,41 @@ void BlackoilWellModelGeneric::initInjMult() {
...
-            const auto& ws = this->wellState().well(well->indexOfWell());
-            if (values.empty()) {
-                values.assign(ws.perf_data.size(), 0.); // initializing to be zero
+            auto it = this->filter_cake_.find(well->name());
+            if (it == this->filter_cake_.end()) {
+                this->filter_cake_.insert({well->name(), WellFilterCake(*well)});
+                it = this->filter_cake_.find(well->name());
             }
-            well->updateFiltrationParticleVolume(dt, water_index, this->wellState(), values);
+            it->second.updateFiltrationParticleVolume(dt, water_index, this->wellState());
```
bska, line 1436: I think you could probably use `unordered_map<>::emplace()` here:
```C++
auto [it, inserted] = this->filter_cake_
    .emplace(std::piecewise_construct,
             std::forward_as_tuple(well->name()),
             std::forward_as_tuple(*well));
```
which will insert a new node (`it`), and return `inserted = true`, if `well->name()` does not exist, or return the existing node (`it`) and `inserted = false` otherwise.  That way you don't have to look up `well->name()` twice.

bska (commented):
This is good, but I'll repeat my earlier suggestion of using `unordered_map<>::emplace()` to transparently handle the case of wells existing or not.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
akva2: done.

akva2:
i had to reorganize a little as the WellInterface instances are recreated at each step.

akva2:
jenkins build this please

bska (approved):

PR #4747, opened 2023-07-12
Extract Well and Group Initialization

bska:
In preparation of adding support for opening/creating wells or groups in the middle of a report step.  This is needed if an `ACTIONX` block runs something like `WELOPEN` or `WELSPECS`/`COMPDAT`.


bska:
Build check is green.  I'll mark this as "ready for review" now.

bska:
jenkins build this please

blattms (approved):

blattms:
thanks a lot. Looks good to me ( and better than before :muscle: )

PR #4749, opened 2023-07-13
Update Well Topology if Triggered From ACTIONX

bska:
This commit adds a new flag data member,
```C++
bool wellStructureChangedDynamically_
```
to the generic black-oil well model.  This flag captures the
```C++
well_structure_changed
```
value from the `SimulatorUpdate` structure in the `updateEclWells()` member function.  Then, in `BlackoilWellModel::beginTimeStep()`, we key a well structure update off this flag when set.  This, in turn, enables creating or opening wells as a result of an `ACTIONX` block updating the structure in the middle of a report step.

### DIFF
--- a/opm/simulators/flow/EclActionHandler.cpp
+++ b/opm/simulators/flow/EclActionHandler.cpp
@@ -188,21 +188,24 @@ void EclActionHandler::applySimulatorUpdate(const int report_step,
                                             const SimulatorUpdate& sim_update,
                                             bool& commit_wellstate,
                                             const TransFunc& updateTrans)
-  {
-      OPM_TIMEBLOCK(applySimulatorUpdate);
-      this->wellModel_.updateEclWells(report_step, sim_update.affected_wells, summaryState_);
-      if (!sim_update.affected_wells.empty())
-          commit_wellstate = true;
+{
+    OPM_TIMEBLOCK(applySimulatorUpdate);
 
-      if (sim_update.tran_update) {
-          const auto& keywords = schedule_[report_step].geo_keywords();
-          ecl_state_.apply_schedule_keywords( keywords );
-          eclBroadcast(comm_, ecl_state_.getTransMult() );
+    this->wellModel_.updateEclWells(report_step, sim_update, this->summaryState_);
 
-          // re-compute transmissibility
-          updateTrans(true);
-      }
-  }
+    if (!sim_update.affected_wells.empty()) {
+        commit_wellstate = true;
+    }
+
+    if (sim_update.tran_update) {
+        const auto& keywords = schedule_[report_step].geo_keywords();
+        ecl_state_.apply_schedule_keywords( keywords );
+        eclBroadcast(comm_, ecl_state_.getTransMult() );
+
+        // re-compute transmissibility
+        updateTrans(true);
+    }
+}
 
 std::unordered_map<std::string, double>
 EclActionHandler::fetchWellPI(const int reportStep,
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -32,6 +32,9 @@
 #include <opm/output/eclipse/RestartValue.hpp>
 
 #include <opm/input/eclipse/EclipseState/EclipseState.hpp>
+#include <opm/input/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp>
+
+#include <opm/input/eclipse/Schedule/Action/SimulatorUpdate.hpp>
 #include <opm/input/eclipse/Schedule/Group/GConSale.hpp>
 #include <opm/input/eclipse/Schedule/Group/GroupEconProductionLimits.hpp>
 #include <opm/input/eclipse/Schedule/Group/GConSump.hpp>
@@ -42,7 +45,7 @@
 #include <opm/input/eclipse/Schedule/Schedule.hpp>
 #include <opm/input/eclipse/Schedule/Well/WellConnections.hpp>
 #include <opm/input/eclipse/Schedule/Well/WellTestConfig.hpp>
-#include <opm/input/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp>
+
 #include <opm/input/eclipse/Units/Units.hpp>
 
 #include <opm/simulators/utils/DeferredLogger.hpp>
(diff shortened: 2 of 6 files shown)


bska:
As an example of the effect, the figure below compares the field pressure (`FPR`)  and well level water injection rate (`WWIR`) of injector `WI01` from [ACTIONX_WCONINJE](https://github.com/OPM/opm-tests/blob/f8d7a3454e9c1111f77602e62c4550bb20654f60/actionx/ACTIONX_WCONINJE.DATA) in OPM-Tests when run with
```sh
--solver-max-time-step-in-days=1.0
```
using the current master sources (`_MASTER`) and this PR (`_DEV`).  Action [`ACT-02`](https://github.com/OPM/opm-tests/blob/f8d7a3454e9c1111f77602e62c4550bb20654f60/actionx/ACTIONX_WCONINJE.DATA#L467-L489) triggers on the 12th of July 2019 in both runs, as a result of the field pressure dropping below the threshold value of 230 bars, but only the current PR immediately opens the `WI01` injector.  In the current master sources the `WI01` injector is not opened until the start of the next report step on the 1st of August 2019.  Both of these runs were parallel, using `mpirun -np 5`, on my local development machine.

---

injection-rate-comparison

blattms (approved):
Looks good to me (assuming that we catch all those structure changes), I would have merged this as it seems to fit urgent issues, but it is marked as WIP and I assume that there is  a reason for that.

blattms:
what is the status here?

bska:
> what is the status here?

We're technically still waiting for a bit of downstream testing, but as far as I'm concerned we should probably just merge this now and deal with any fall-out later.

blattms:
Sounds like a plan.

bska:
>> > what is the status here?
>> 
>> We're technically still waiting for a bit of downstream testing, but as far as I'm concerned we should probably just merge this now and deal with any fall-out later.
>
> Sounds like a plan.

Cool.  I'll mark this as ready for review and run a build check then.

bska:
jenkins build this please

blattms (approved):
Looks perfect to me. Thanks.

blattms:
jenkins build this please

blattms:
jenkins build this please

bska:
> jenkins build this please

We're getting unexpected [regression failures](https://ci.opm-project.org/job/opm-simulators-PR-builder/5431/testReport/) from this work.  The `compareECLFiles_flow+<CASE>` failure is mostly benign and can be fixed by updating the reference solutions.  The failure in `compareParallelSim_flow+<CASE>` however is more troublesome.  I can't make that case pass on my own development machine even if I use **really** small time steps (e.g., `--solver-max-time-step-in-days=0.1`).  At the moment I'm at a loss for how to proceed here, and in any case it seems unlikely that we'l be able to get this into the October release.

bska:
jenkins build this please

blattms:
There is something strange (not due to this PR?). I did some printf-debugging to whether sim_update.affected_wells is consistent in a parallel run.

I see messages like: 
```
Action INJECTION triggered at 01-Aug-2020 00:00:00 (report interval 3 to 4). Well: P1
```
is seems like sim_update.affected_wells contains "WI1". Both in parallel and serial.

maybe the index is off by one somewhere?

bska:
> I see messages like:
> 
> ```
> Action INJECTION triggered at 01-Aug-2020 00:00:00 (report interval 3 to 4). Well: P1
> ```
> 
> but it seems like sim_update.affected_wells contains "WI1" instead. Both in parallel and serial.

Right.  It's a little bit subtle, and I welcome any suggestions for improvement here.  The log message is generated in [`EclActionHandler::applyActions()`](https://github.com/OPM/opm-simulators/blob/e9bb5d0539899fef4f425de0b368d79794699934/opm/simulators/flow/EclActionHandler.cpp#L161-L163) and the "matching wells" are those wells for which the `ACTIONX` condition triggered.  In the case of `ACTIONX_M1/INJECTION` this is [well `P1`](https://github.com/OPM/opm-tests/blob/72b2fd08d5902c949c5ff983bce61c2da6f48608/udq_actionx/ACTIONX_M1.DATA#L387) since the condition uses `WBHP` for that well.  The "affected wells" in the simulator update, however, are those which are updated in the actual action block.  In this case that is `WI1` since the action block [has a `WELOPEN` event](https://github.com/OPM/opm-tests/blob/72b2fd08d5902c949c5ff983bce61c2da6f48608/udq_actionx/ACTIONX_M1.DATA#L390) for that well.  In other words, the log message is "correct" in that the affected wells are not the same as the matching wells but as I said this isn't a particularly good log message.

On the other hand, I've finally found a set of parameters that enable running the test in parallel and having the results match closely enough to those produced by a sequential run that we can consider merging this PR.  It will still require updating the reference results for the sequential run so we'll have to discuss what to do about that since we're in the process of generating a new release.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

blattms:
Thanks for the clarification. This now triggered a thought on my side:

I have previously been wondering why this works in parallel when it seemed that all code works only on the local process without any communication. I concluded that is probably OK because the state that the ACTIONX evaluates is shared between the processes (and I assumed that the wells triggering the action are  a superset of the affected wells).

Now this is not really the case, when an action gets triggered for a set wells but it affects another set of wells. In a parallel run the set of affected wells might only contain wells that do not perforate local wells while the triggering set might have wells which perforate local cells in it, and vice versa.

Could it be that in parallel we correctly evaluate the actions for local wells, but if the affected well is on another process (no perforated cell on this process) then we do nothing? Maybe this did not happen before because at report steps we do the necessary computation implicitly?

bska:
jenkins build this please

bska:

bska:
jenkins build this opm-tests=1085 please

bska:
The new reference solutions have been installed on the CI system.  Merging per prior approval.

blattms:
In retrospect, I have questions about this (because I have similar problems):

The reason for the problem (changes occur to the wells only on the next report step) was, that these changes happen(ed) in the simulator due to Events signaled by the schedule. This happens at the next report step (because that is where the schedule is changed and that is where we query for events).
Is it correct that with this change the event about the well changing will still be triggered on the next report step? 

Probably not a real problem here, because it is just additional work, but does not change anything, but for other keywords (e.g. NEXTSTEP) it might be.

PR #4750, opened 2023-07-19
Do a graceful exit instead of MPI_Abort for expected exceptions.

blattms:
Instead of unconditionally issuing MPI_Abort if we encounter a fatal exception, we try to test whether all processes have experienced this exception and if this is the case just terminate nomally with a exit code that signals an error. We still use MPI_Abort if not all processes get an exception as this is the only way to make sure that the program aborts.

This approach also works around issues in some MPI implementations that might not correctly return the error.

Multiple messages like this are gone now:
```
MPI_ABORT was invoked on rank 1 in communicator MPI_COMM_WORLD
with errorcode 1.

[smaug.dr-blatt.de:129359] 1 more process has sent help message help-mpi-api.txt / mpi-abort
[smaug.dr-blatt.de:129359] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

Bu we still see something like this:
```

  Process name: [[35057,1],0]
  Exit code:1
```

### DIFF
--- a/opm/simulators/flow/FlowMainEbos.cpp
+++ b/opm/simulators/flow/FlowMainEbos.cpp
@@ -77,6 +77,84 @@ void handleExtraConvergenceOutput(SimulatorReport& report,
       report.fullReports(os);
     }
 }
+void checkAllMPIProcesses()
+{
+#if HAVE_MPI
+    const auto& comm = EclGenericVanguard::comm();
+    if (comm.size() > 1)
+    {
+        // we try to prevent the abort here.
+        // For that we need a signal that each process is here.
+        // Each process sends  a message to rank 0.
+        const int tag = 357912;
+        if (comm.rank() == 0)
+        {
+            // wait for a message from all processes.
+            std::vector<MPI_Request> requests(comm.size() - 1, MPI_REQUEST_NULL);
+            std::vector<int> data(comm.size()-1);
+
+            for(decltype(comm.size()) i = 1; i < comm.size(); ++i)
+            {
+                if (auto error = MPI_Irecv(data.data() + i, 1, MPI_INT, i, tag, comm, requests.data() + i - 1);
+                    error != MPI_SUCCESS) {
+                    OpmLog::error(fmt::format("Error: Could not set up MPI receive (error code : {})", error));
+                    MPI_Abort(MPI_COMM_WORLD, EXIT_FAILURE);
+                }
+            }
+            std::size_t msgs = comm.size() - 1;
+            for(std::size_t tries = 0; msgs >0 && tries < 3; ++tries)
+            {
+                sleep(3);
+                int flag, idx;
+                for(auto left_msgs = msgs; left_msgs > 0; --left_msgs)
+                {
+                    if( auto error = MPI_Testany(comm.size()-1, requests.data(), &idx, &flag, MPI_STATUS_IGNORE);
+                        error != MPI_SUCCESS) {
+                        OpmLog::error(fmt::format("Error: Could not test for MPI message (error code : {})", error));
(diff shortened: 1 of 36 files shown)


blattms:

blattms:

alfbr:
Thank you, this is a very welcome improvement to the MPI abort handling. The code changes looks good to me. Will test this on some full ensembles, in the mean time we should also let our internal benchmark do a quick test.

alfbr:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: drogon - Threads: 1                | 0.966    |
| opm-git | OPM Benchmark: drogon - Threads: 8                | 0.999    |
| opm-git | OPM Benchmark: punqs3 - Threads: 1                | 1.009    |
| opm-git | OPM Benchmark: punqs3 - Threads: 8                | 1.005    |
| opm-git | OPM Benchmark: smeaheia - Threads: 1              | 0.97     |
| opm-git | OPM Benchmark: smeaheia - Threads: 8              | 1.006    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 1         | 1.002    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 8         | 1.001    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 0.995    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 1.014    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 0.997    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 0.982    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.991    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 1.013    |



View result details @ https://www.ytelses.com/opm/?page=result&id=2168

alfbr:
I have now tested this, and I can confirm that on a full ensemble of two hundred realizations all are reported finished when the finish in the queueing system. In addition, the error message is far less confusing. I do still believe that the message is developer centric, and that we should to a larger degree shelter users from it. That is this one:

```

  Process name: [[35057,1],0]
  Exit code:1

```
I would prefer to have that in the DBG file or even PRT file rather than in the terminal window. Alternatively, only output it when a command line option triggers it. For a user I believe it is sufficient replacing it with something like:

```
the job to be terminated.

```

alfbr (approved):

blattms:
Unfortunately, this message seems to be beyond our control as it comes from mpirun. Let me try to see whether there is a way to prevent it via command line and envrionment variables, though.

opm/simulators/flow/FlowMainEbos.cpp
```diff
@@ -77,6 +77,84 @@ void handleExtraConvergenceOutput(SimulatorReport& report,
...
+        const int tag = 357912;
+        if (comm.rank() == 0)
+        {
+            // wait for a message from all processes.
+            std::vector<MPI_Request> requests(comm.size() - 1, MPI_REQUEST_NULL);
+            std::vector<int> data(comm.size()-1);
+
+            for(decltype(comm.size()) i = 1; i < comm.size(); ++i)
+            {
+                if (auto error = MPI_Irecv(data.data() + i, 1, MPI_INT, i, tag, comm, requests.data() + i - 1);
```
atgeirr, line 98: I believe there is a off-by-one bug here, if for example comm has size 2, then data will have size 1, but we will pass a pointer pointing to the (non-existing) element 1.

blattms: Indeed we write out of bounds here.

PR #4753, opened 2023-07-24
Don't write out of bounds (fixes fallout from PR #4750)

blattms:
While we never use the data we receive, we should still not write beyond arrays as this may create problems.


blattms:

PR #4758, opened 2023-07-26
moved fully implicit calls to opm-simulators

hnil:
-- changed spesicic nldd calls to use subDomain instead of gridview (-- clang formated FIBlackoil)
-- moved fully implict related updates to FIblackoil

This pullrequest should be followed up with a delete of similar functions in fvbasediscretization.

### DIFF
--- a/ebos/FIBlackOilModel.hpp
+++ b/ebos/FIBlackOilModel.hpp
@@ -35,28 +35,108 @@
 
 #include <stdexcept>
 
-namespace Opm{
-    template<typename TypeTag>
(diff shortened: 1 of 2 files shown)


atgeirr (approved):
I think this looks like a good idea, the `invalidateAndUpdateIntensiveQuantities` methods make more sense here, I agree. It will also make it easier to refactor this particular part to not iterate over the grid and not use ElementContexts.

atgeirr:
jenkins build this please

PR #4760, opened 2023-07-26
removed GRUPNET from unsupported keyword list

plgbrts:


plgbrts:
This PR depends on https://github.com/OPM/opm-common/pull/3609

GitPaean:
We should add a regression test with the test case from the PR `https://github.com/OPM/opm-tests/pull/1014`.  So we can update the reference and this part of the development can be protected by regression checking.

GitPaean (changes requested):
For this PR to go in, please add a regression test with the case `NETWORK-01_STANDARD.DATA` so the development will be protected.  Likely you need to use the option `--enable-tuning=true`, but it depends your own testing finding.

opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
```diff
@@ -116,6 +116,13 @@ partiallySupported()
                {1,{true, allow_values<std::string> {"NO", "YES"}, "GRIDOPTS(TRANMULT): should be set to either NO or YES"}}, // TRANMULT
             },
          },
+         {
+            "GRUPNET",
+            {
+               {5,{true, allow_values<std::string> {"NO"}, "GRUPNET(SUBSEAMANIFOLD): only option NO is supported"}}, // SUB_SEA_MANIFOLD
+               {6,{true, allow_values<std::string> {"NO"}, "GRUPNET(LIFTGAS): only option NO is supported"}}, // LIFT_GAS_FLOW_THROUGH
```
GitPaean, line 123: Please also add item 7.

PR #4761, opened 2023-07-27
Fix temp boundary

hnil:
-- fix a bug in boundary conditions.


akva2 (approved):
Obvious error. The index is translated inside boundaryCondition().

PR #4762, opened 2023-07-28
-- added tracy converage of nldd code

hnil:
Added near full coverage of nldd code with tracy.


hnil:
Closing: Tracy has never been adopted upstream, so the NLDD Tracy coverage has no home here. Branch retained on my fork.

atgeirr:
> Tracy has never been adopted upstream

That is inaccurate, as it is indeed supported...

PR #4765, opened 2023-08-01
removed use of hidden private defines for poly and alugrid

hnil:
- fixed plygrid
- renamed executables to be with blackoil


hnil:
Need to check compilation with dune fem.

blattms:
jenkins build this please

akva2:
do not merge as-is, it's broken with dune-fem.

hnil:
Hopefully fixed for different combinations of dune fem. Also now all ALUGRID, POLYGRID ++ instantiations can be removed and substituted with xxx_impl.hh where it is needed. This simplifies things in particular for compilations with DUNE_FEM.

akva2:
rebase please.

akva2:

blattms:
Ok, this introduced quite some work on our side. To really be able to fix our stuff I needed to understand what it actually does. So I'll try:

Up to now whenever there were *_impl.hh files their content was actually part of the header that was move to this file for clarity. As a result those *_impl.hh files were included at the end corresponding header (*.hh file).

In this PR this is somewhat different:
Here it seems like code was moved from *.cc files to *_impl.hh files and the new header is now included in the correcsponding *.cc file. In addition it is included in flow_*.cc files. I guess the later is for alugrid and fem. Those *_impl.hh are never installed

I think using *_impl.hh differently from the rest of OPM is very confusing. Are we really sure that there still is reduction in compile time compared to having everything in the header? I would assume that this PR should have increased compile time over the previous version. Did somebody check that.

Can we think of a way to make the different behavior more apparent?

akva2:
the _impl.hh files holds the implementations. in the .cc files we instance for cpgrid, and cpgrid only.
for alugrid and polyhedralgrid we build the code as part of the main binary, so in those .cpp files the _impl.hh files are included.

there is the same reduction for cpgrid, and as most users do not build polyhedral or alugrid binaries, there is a net save. however for polyhedralgrid and alugrid we are back to how it used to be in terms of compile time.

i thought i added all the _impl.hh files to the installation list, if i missed something please send a pr.

atgeirr:
Good point about this use of _impl.hpp being different from earlier pattern. A simple way to distinguish would be to adopt a different naming scheme. For example, _impl.cpp for the "new pattern" usage (keeping _impl.hpp for the old). That would mean renaming a few files from _impl.hpp to _impl.cpp, and it should be easy to remember that you never include a .cpp file from a .hpp file.

PR #4768, opened 2023-08-02
ebos/: header cleanups

akva2:
Not at all claiming this is the end of the story, but this is a contribution towards making these headers standalone.

### DIFF
--- a/ebos/ecltracermodel.hh
+++ b/ebos/ecltracermodel.hh
@@ -31,8 +31,13 @@
 #include <ebos/eclgenerictracermodel.hh>
 
 #include <opm/models/utils/propertysystem.hh>
+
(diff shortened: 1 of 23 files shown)


akva2:
jenkins build this please

ebos/ecltracermodel.hh
```diff
@@ -31,8 +31,13 @@
 #include <ebos/eclgenerictracermodel.hh>
 
 #include <opm/models/utils/propertysystem.hh>
+
 #include <opm/simulators/utils/VectorVectorDataHandle.hpp>
 
+#include <array>
+#include <iostream>
```
bska, line 38: I know it's needed here, but we really should work to avoid including `<iostream>` in our headers.

bska (approved):
Thanks for doing this work.  Every little bit helps here and I'll merge into master.

PR #4772, opened 2023-08-03
changing the bhp lower limit to be zero during newton update instead of 1 bar

GitPaean:
an attempt to fix a reported case with 1bar production bhp limit.

### DIFF
--- a/opm/simulators/wells/MultisegmentWellPrimaryVariables.cpp
+++ b/opm/simulators/wells/MultisegmentWellPrimaryVariables.cpp
@@ -180,7 +180,8 @@ updateNewton(const BVectorWell& dwells,
         {
             const int sign = dwells[seg][SPres] > 0.? 1 : -1;
             const double dx_limited = sign * std::min(std::abs(dwells[seg][SPres]) * relaxation_factor, max_pressure_change);
-            value_[seg][SPres] = std::max(old_primary_variables[seg][SPres] - dx_limited, 1e5);
+            // we avoid segment pressure to go below 0.
+            value_[seg][SPres] = std::max(old_primary_variables[seg][SPres] - dx_limited, 0.);
         }
 
         // update the total rate // TODO: should we have a limitation of the total rate change?
--- a/opm/simulators/wells/StandardWellPrimaryVariables.cpp
+++ b/opm/simulators/wells/StandardWellPrimaryVariables.cpp
@@ -291,8 +291,13 @@ updateNewton(const BVectorWell& dwells,
 
     // updating the bottom hole pressure
     const int sign1 = dwells[0][Bhp] > 0 ? 1: -1;
-    const double dx1_limited = sign1 * std::min(std::abs(dwells[0][Bhp]),
-                                                std::abs(value_[Bhp]) * dBHPLimit);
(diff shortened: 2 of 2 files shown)


GitPaean:
jenkins build this please

GitPaean:
benchmark please

GitPaean:
This one shares similar problems that it affects the cases involving UDQ, WECON and WTEST, 

image

GitPaean:
The core cause of the reported issue is that the DATA file has a 1bar bhp limit for one producer, at the same time, our code had-coded so that the bhp does not go below 1 bar. In the reported case, the well needs to switch to the bhp control, the hard-coded 1 bar value will prevent that. 

A more ideal way to do this is to make the lower limit of the newton update depending on the well's bhp limit. It just needs to be able to go below  the bhp limit so that it can switch to the bhp control. Fixing the running of the reported issue is very easy while at the same time, it looks like touching this part affect the running which determining well operations based on well results.

GitPaean:
For the cases with WECON and WTEST, the results with WECON is not much affected, while it looks like the WTEST process is the cause of the discrepancy.

GitPaean:
It looks like one related reason this PR affects the cases with re-opening much is that during the re-opening test, very often (not checking all the situation yet), the wells are initialized with zero bhp (also thp). Continue investigating.

GitPaean:
For those wells with zero bhp
```c++
    const double dx1_limited = sign1 * std::min(std::abs(dwells[0][Bhp]),
                                                std::abs(value_[Bhp]) * dBHPLimit);
```
The following code will make the bhp can not be updated at all. At the end, it is not solvable and will not be re-opened.

GitPaean:
At least for this PR. The problem is that zero bhp was used when entering a well testing. It is something we should avoid.

GitPaean:
Unfortunate, it still breaks a few tests that requires detailed analysis,which is a little disappointing. Detailed analysis can be quite time consuming which is also a little unfortunate.  

image

GitPaean:
jenkins build this please

GitPaean:
> For those wells with zero bhp
> 
> ```c++
>     const double dx1_limited = sign1 * std::min(std::abs(dwells[0][Bhp]),
>                                                 std::abs(value_[Bhp]) * dBHPLimit);
> ```
> 
> The following code will make the bhp can not be updated at all. At the end, it is not solvable and will not be re-opened.

At the end, the trouble is still whenever the bhp is get to be zero, then dx1_limited will be zero. 

https://github.com/OPM/opm-simulators/pull/4841 fixes partly the zero bhp from the initialization. 

[a4d8c15](https://github.com/OPM/opm-simulators/pull/4772/commits/a4d8c15e8b914fc4afa6e8fdbac96c16a65d3044) fixes the zero bhp during the newton update.

This PR and https://github.com/OPM/opm-simulators/pull/4858 both work when coming to this respective.  I *believe* (has not checked exhaustively) the previous troubles with some cases should be gone. 

Will come to think about a good/systematic manner to get in. Some previous discussion agreed that we should not handle negative bhp values.

GitPaean:
A different approach got adopted https://github.com/OPM/opm-simulators/pull/4877 . Closing this one now.

bska:
> Closing this one now.

Did you mean to hit the "close" button too?

PR #4774, opened 2023-08-04
changed default fvbasediscretization to not have dunefem even if dune…

hnil:
-- need merging of changes in opm-models.


bska:
> need merging of changes in opm-models.

Does this refer to the PR OPM/opm-models#815?

hnil:
Closed after discussion, since the instances of alle the classes for different tings have a sufficient extra cost. For reference this show how to turn of DUNE_FEM type itertors in code even if DUNE_FEM is present.

PR #4775, opened 2023-08-04
Add dedicated class for output of logs

akva2:
More separation of concerns.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -48,6 +48,7 @@ list (APPEND MAIN_SOURCE_FILES
   opm/simulators/flow/ExtraConvergenceOutputThread.cpp
   opm/simulators/flow/FlowMainEbos.cpp
   opm/simulators/flow/KeywordValidation.cpp
+  opm/simulators/flow/LogOutputHelper.cpp
   opm/simulators/flow/Main.cpp
   opm/simulators/flow/NonlinearSolverEbos.cpp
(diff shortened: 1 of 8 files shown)


akva2:
jenkins build this please

atgeirr:
A bit tricky to review large code moves, especially for inadvertent changes. I guess the tests is what we rely on to make it work, but I do not think we have a lot of tests on the log output. Any ideas for how to go about checking in a rational way?

akva2:
i guess running old binary and new binary and diffing the output files is the only way to go about it. is what i did while developing it.

akva2:
jenkins build this please

atgeirr:
I have not yet reviewed this throroughly, but I think we should avoid the ebos/ directory as well as `EclXYZ` class names for things that do not take a TypeTag parameter or in another way directly deals with the opm-models interface or approach.

Maybe `LogOutputHelper`?

akva2:
changed to LogOutputHelper.

akva2:
jenkins build this please

akva2:
I have added tests. The reference data was created before the refactoring.

akva2:
i put it in utils, but okay, i can move it to flow..

akva2:
jenkins build this please

akva2:
jenkins build this please

atgeirr:
jenkins build this please

atgeirr (approved):
Tested and got identical output (up to the usual differences from one run to another).

PR #4777, opened 2023-08-04
remove accidentially left-over member

akva2:
Was left when I put the nldd solver in a separate class.


akva2:
jenkins build this please

PR #4780, opened 2023-08-07
EclTracerModel: replace cout with OpmLog

akva2:
I doubt this alone makes much of a difference, but it moves us in the right direction.


bska (approved):
Looks good to me.  Please merge when green

PR #4783, opened 2023-08-08
adding perf_data comparison in equality operator for SingleWellState

GitPaean:
not sure whether it was intentionally absent. I think it should be there unless I miss something.

### DIFF
--- a/opm/simulators/wells/SingleWellState.cpp
+++ b/opm/simulators/wells/SingleWellState.cpp
@@ -299,6 +299,7 @@ bool SingleWellState::operator==(const SingleWellState& rhs) const
            this->surface_rates == rhs.surface_rates &&
            this->reservoir_rates == rhs.reservoir_rates &&
            this->prev_surface_rates == rhs.prev_surface_rates &&
+           this->perf_data == rhs.perf_data &&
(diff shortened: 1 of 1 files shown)


GitPaean:
jenkins build this please

atgeirr:
I believe this is actually intentional. The operator is used primarily to unit test the serialization, and the serialization skips the perf_data. That is done because it is not strictly required for perfect restarting, as perf_data will be created correctly from other read objects. If you want to add this to the equality operator you should also add it to the serialization (including ensuring a nontrivial serializationTestObject for the perf_data).

I think it is a bit of a dilemma. On one hand we want the save file to be as compact as possible, but it is very weird to have equality operators that do not really compare equality. I am leaning towards adding these "unnecessary for restart" members to serialization and equality, but it is not clear. @akva2 probably has a more informed opinion.

akva2:
in general you should indeed be very careful with changing what is serialized and not. however, perf_data is already serialized, so the missing entry in equality operator is an oversight here.

atgeirr:
> perf_data is already serialized,

You are correct, I missed it since it was done in different order in the two methods. So this should be merged then?

akva2:
yes

PR #4784, opened 2023-08-09
- moded all bda/gpu spesific tings to separete class

hnil:
advantages:
- one can instantiate solvers without bda part of code
- main code gets cleaner


akva2:
this is pure removal, missing some adds i assume.

akva2:
jenkins build this please

akva2 (approved):
Since I authored parts of it, wait for secondary review but lt all is good for me now.

atgeirr:
> Since I authored parts of it, wait for secondary review but lt all is good for me now.

I will take at least a high-level look at it now.

atgeirr:
Looks fine to me, just one word of warning: the COMPILE_BDA_BRIDGE macro works in a way similar to the dune-fem stuff we just refactored, if set to 1 it changes what you get in regular flow (setting the LinearSolverSplice type), rather than enabling a separately compiled variant. But this is actually what we want right now, since it makes for a minimal change compared to the current situation. We can change this later and then it will be much easier thanks to this refactoring.

Merging!

bska:
Following this PR&ndash;I think, I haven't tried to revert it yet or toggle the `USE_BDA_BRIDGE` option&ndash;I'm getting local build failures of the form
```
${ROOT}/opm/simulators/linalg/bda/BdaBridge.cpp:306:18: error: invalid use of incomplete type ‘class Opm::WellContributions’
```
which is because [`BdaBridge.cpp`](https://github.com/OPM/opm-simulators/blob/e6597b6b20b561327c075f7790ba6818b54d2515/opm/simulators/linalg/bda/BdaBridge.cpp#L306) calls member functions on the `WellContributions` object without a valid declaration in scope.  I don't have OpenCL on this machine so the [`openclWellContributions`](https://github.com/OPM/opm-simulators/blob/e6597b6b20b561327c075f7790ba6818b54d2515/opm/simulators/linalg/bda/BdaBridge.cpp#L37) do not get included.

I can work around the problem and restore my local build with the patch below, but I don't know if this is the correct solution.  Please advice.

---

```diff
diff --git a/opm/simulators/linalg/bda/BdaBridge.cpp b/opm/simulators/linalg/bda/BdaBridge.cpp
index e62d5212c..9df7be0bd 100644
--- a/opm/simulators/linalg/bda/BdaBridge.cpp
+++ b/opm/simulators/linalg/bda/BdaBridge.cpp
@@ -28,6 +28,8 @@
 #include <opm/simulators/linalg/bda/BdaBridge.hpp>
 #include <opm/simulators/linalg/bda/BdaResult.hpp>
 
+#include <opm/simulators/linalg/bda/WellContributions.hpp>
+
 #if HAVE_CUDA
 #include <opm/simulators/linalg/bda/cuda/cusparseSolverBackend.hpp>
 #endif
```

bska:
> That is a correct fix.

Cool.  I'll make a PR.

PR #4788, opened 2023-08-10
fixed: sanity check rock compaction table indices

akva2:
Check that rock compaction indices are within bounds.

Fixes crash on startup in a client case where there are is ROCKNUM keyword and only a single ROCKTAB.
The indices are then based on ROCKOPTS item 3 which defaults to PVTNUM.


akva2:
jenkins build this please

ebos/eclgenericproblem_impl.hh
```diff
@@ -169,6 +169,22 @@ readRockParameters_(const std::vector<Scalar>& cellCenterDepths)
         const auto& num = eclState_.fieldProps().get_int(rock_config.rocknum_property());
         for (size_t elemIdx = 0; elemIdx < numElem; ++ elemIdx) {
             rockTableIdx_[elemIdx] = num[elemIdx] - 1;
+            if (!rockCompPoroMult_.empty() &&
+                rockTableIdx_[elemIdx] >= rockCompPoroMult_.size()) {
+                throw std::runtime_error("Rock compaction table index for elem " +
+                                         std::to_string(elemIdx) +
+                                         " read from " + rock_config.rocknum_property() +
+                                         " is larger than number of tables " +
```
bska, line 177: I agree, this is the common case.  There is another one though and we should report that one too.  If the model erroneously specifies a `*NUM` value less than 1 in an active cell, then we're equally in trouble.  On another note, if this function (`readRockParameters_()`) runs on all MPI ranks, then the `elemIdx` is local to each rank and **probably** won't make much sense to the user.  How much work would it be to translate back to a global (I,J,K) tuple?

In any case I think we should report a message of the form

> Rock compaction table "num[elemIdx]" ... is out of bounds

to encompass the `NUM < 1` possibility too.

bska (commented):
Good find!

ebos/eclgenericproblem_impl.hh
akva2: done and done.

akva2:
jenkins build this please

ebos/eclgenericproblem_impl.hh
```diff
@@ -169,6 +173,24 @@ readRockParameters_(const std::vector<Scalar>& cellCenterDepths)
         const auto& num = eclState_.fieldProps().get_int(rock_config.rocknum_property());
         for (size_t elemIdx = 0; elemIdx < numElem; ++ elemIdx) {
             rockTableIdx_[elemIdx] = num[elemIdx] - 1;
+            auto fmtError =
+                [&num,elemIdx,&ijkIndex,&rock_config](const char* type, std::size_t size)
+                {
+                    return fmt::format("{} table index {} for elem {} read from {}"
+                                      " is larger than number of tables {}",
```
bska, line 180: Maybe we could say

> is out of bounds for the number of tables

instead of "is larger than"?

bska (commented):
Thanks a lot for the updates.  I'm happy that it turned out to be that simple.  If I'm reading the code correctly we'll get zero-based Cartesian coordinates.  Do you think you could make the `ijkIndex` return one-based coordinates instead&ndash;i.e., add 1 to each element?  I **believe** that users are more accustomed to one-based coordinates in these kinds of messages.

ebos/eclgenericproblem_impl.hh
akva2: done and done.

akva2:
jenkins build this please

bska (approved):
Excellent.  Thanks a lot for the updates and sorry about the multiple rounds of review.  Please merge once the build check is green.

PR #4789, opened 2023-08-10
Changed: Do not use dune-fem in default simulators even if it is available

akva2:
It is now typetag configurable. Should be enabled where used.

Downstream of https://github.com/OPM/opm-models/pull/821

### DIFF
--- a/ebos/collecttoiorank.cc
+++ b/ebos/collecttoiorank.cc
@@ -35,6 +35,10 @@
 
 namespace Opm {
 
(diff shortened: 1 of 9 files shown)


akva2:
Note: I have tested this locally with dune-fem enabled, jenkins does not currently exercise the code path, but that will change soonish.

atgeirr (approved):
Reviewed with "show whitespace changes" turned off, to make it possible to follow the significant changes, so I may have missed bad indenting. Good to go!

PR #4792, opened 2023-08-11
Bring WellContributions Declaration in Scope

bska:
This restores the build on machines which enable the BDA bridge, but which do not have OpenCL installed.


bska:
Self-merging this to restore the build on my development machine.

PR #4794, opened 2023-08-15
fixed: avoid segfault in cleanup if simulator has not been set up

akva2:
If an exception is thrown during simulator setup, the ebosSimulator_ object might not be instanced. We then deref a nullptr calling the eclState() member further down.


akva2:
jenkins build this please

bska (approved):
Sure.  However unlikely, it's better to be safe than sorry.  I'll merge into master.

PR #4795, opened 2023-08-15
EclActionHandler: move from ebos/ to opm/simulators/flow

akva2:
Since there is no opm-models usage, we should move this out of the ebos/ folder.


akva2:
jenkins build this please

bska (approved):
Good.  I'l merge into master.

PR #4797, opened 2023-08-15
changed: move EclInterRegFlows to opm/simulators/flow

akva2:
no opm-models usage


akva2:
jenkins build this please

bska (approved):
This is okay and I'll merge into master.  This does nevertheless increase the coupling because we get more stuff in `ebos/` which depends on `opm/simulators/` in addition to the bits in `opm/simulators/` which depend on stuff in `ebos/`.  I don't know if we want to address that at some point.

PR #4800, opened 2023-08-16
SimulatorFullyImplicitBlackoilEbos: remove unused member

akva2:
In principle the whole WellConnectionAuxiliaryModule could be nuked, but I'm not entirely sure what the deal here is so I played it safe.


akva2:
jenkins build this please

blattms:
thanks a lot.

bska:
> In principle the whole WellConnectionAuxiliaryModule could be nuked

I agree that this is currently unused, but I'd be a little hesitant to remove auxiliary modules.  Those could very well become useful in the near future.  On the other hand, we could probably resurrect them if needed.

PR #4802, opened 2023-08-17
Remove Unused Return Value

### LINKED ISSUE #4799: EclTransmissibility::applyNncToGridTrans_ Returns a Value Which is Never Used
I assume I'm missing something since the current code is effectively unchanged since commit 5e83a26908 and I'm not at all familiar with this code, but [`EclTransmissibility<>::applyNncToGridTrans_()`](https://github.com/OPM/opm-simulators/blob/207e57804d9713090004fe002efa401cc1804c74/ebos/ecltransmissibility.hh#L199-L200) [returns](https://github.com/OPM/opm-simulators/blob/207e57804d9713090004fe002efa401cc1804c74/ebos/ecltransmissibility_impl.hh#L875) a value&ndash;a `tuple<>` of `vector<NNCdata>` objects&ndash;which in turn is [just discarded](https://github.com/OPM/opm-simulators/blob/207e57804d9713090004fe002efa401cc1804c74/ebos/ecltransmissibility_impl.hh#L469) at the call site.  This means that the effort to [identify unrepresented NNCs](https://github.com/OPM/opm-simulators/blob/207e57804d9713090004fe002efa401cc1804c74/ebos/ecltransmissibility_impl.hh#L864-L866) is effectively wasted.

Is this intentional?  If so, should we make the return type be `void` instead to make it clear that the function does not need to return anything?

bska:
The return value from `applyNncToGridTrans_()` was never used and the author confirms that this is intentional.  Remove the entire return value to signal this intention.

Resolves #4799.

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -192,12 +192,9 @@ protected:
      *
      * \param cartesianToCompressed Vector containing the compressed index (or -1 for inactive
      *                              cells) as the element at the cartesian index.
-     * \return Two vector of NNCs (scaled by EDITNNC). The first one are the NNCs that have been applied
-     *         and the second the NNCs not resembled by faces of the grid. NNCs specified for
-     *         inactive cells are omitted in these vectors.
+     * \return Nothing.
      */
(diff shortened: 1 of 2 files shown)


bska:
jenkins build this please

PR #4803, opened 2023-08-17
added output writing for geomechanical module

akva2:
This is a rebased and processed version of https://github.com/OPM/opm-simulators/pull/4588

### DIFF
--- a/ebos/eclgenericoutputblackoilmodule.cc
+++ b/ebos/eclgenericoutputblackoilmodule.cc
@@ -151,6 +151,7 @@ EclGenericOutputBlackoilModule(const EclipseState& eclState,
                            const SummaryState& summaryState,
                            bool enableEnergy,
                            bool enableTemperature,
+                           bool enableMech,
                            bool enableSolvent,
                            bool enablePolymer,
                            bool enableFoam,
@@ -166,6 +167,7 @@ EclGenericOutputBlackoilModule(const EclipseState& eclState,
     , logOutput_(eclState, schedule, summaryState)
     , enableEnergy_(enableEnergy)
     , enableTemperature_(enableTemperature)
+    , enableMech_(enableMech)
     , enableSolvent_(enableSolvent)
     , enablePolymer_(enablePolymer)
     , enableFoam_(enableFoam)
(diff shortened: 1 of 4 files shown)


akva2:
jenkins build this please

ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -760,6 +786,66 @@ doAllocBuffers(unsigned bufferSize,
     rstKeywords["PRES"] = 0;
     rstKeywords["PRESSURE"] = 0;
 
+    if (enableMech_ && eclState_.runspec().mech()) {
+        this->mechPotentialForce_.resize(bufferSize,0.0);
+        rstKeywords["MECHPOTF"] = 0;
+        this->mechPotentialTempForce_.resize(bufferSize,0.0);
+        rstKeywords["TEMPPOTF"] = 0;
+        this->mechPotentialPressForce_.resize(bufferSize,0.0);
+        rstKeywords["PRESPOTF"] = 0;
```
bska, line 795: I'll repeat [my comment from earlier](https://github.com/OPM/opm-simulators/pull/4588#discussion_r1291260147):

> If `enableMech_` is sufficient to request that geomechanical result arrays be output and we do not expect that users will have something along the lines of
> ```
> RPTRST
>   MECHPOTF TEMPPOTF PRESPOTF ... /
> ```
> in their input decks, then we do not need these `rstKeywords["something"] = 0` statements in here.

The question then becomes: Do we expect that users will enable individual result arrays in `RPTRST`?  Because if we do, then we also need to [tell the output system about them](https://github.com/OPM/opm-common/blob/e22b0944ddce07c58515dcb9b127d504eba995d1/src/opm/input/eclipse/Schedule/RSTConfig.cpp#L162).

bska (commented):
I think this is good, with just a minor caveat about potentially doing a bit too much work with `rstKeywords` helper object.

ebos/eclgenericoutputblackoilmodule.cc
hnil: I think writing out everything is ok for now. In other cases I think negating output is better.

bska (approved):
This was left to linger for too long.  Sorry about that.  It looks good to me with the additional knowledge that we may get to a point where we selectively enable some 3D array output for geomechanical analyses.  I'll merge into master.

PR #4805, opened 2023-08-21
Mark TransMult Object as Const

bska:
That way it's obvious that the surrounding code does not change the object.


bska:
jenkins build this please

akva2 (approved):

PR #4806, opened 2023-08-22
initialize all injectors with a rate to avoid singularities

totto82:
This gives injectors that that is setup using BHP control some start help.

### DIFF
--- a/opm/simulators/wells/SingleWellState.cpp
+++ b/opm/simulators/wells/SingleWellState.cpp
@@ -256,31 +256,28 @@ void SingleWellState::update_injector_targets(const Well& ecl_well, const Summar
         return;
     }
 
-
-    if (inj_controls.cmode == Well::InjectorCMode::GRUP) {
-        this->bhp = this->perf_data.pressure_first_connection * bhp_safety_factor;
-        return;
+    // we initialize all open wells with a rate to avoid singularities
+    double inj_surf_rate = 10.0 * Opm::unit::cubic(Opm::unit::meter) / Opm::unit::day;
+    if (inj_controls.cmode == Well::InjectorCMode::RATE) {
+        inj_surf_rate = inj_controls.surface_rate;
     }
 
-    if (inj_controls.cmode == Well::InjectorCMode::RATE) {
-        auto inj_surf_rate = inj_controls.surface_rate;
-        switch (inj_controls.injector_type) {
-        case InjectorType::WATER:
-            assert(pu.phase_used[BlackoilPhases::Aqua]);
-            this->surface_rates[pu.phase_pos[BlackoilPhases::Aqua]] = inj_surf_rate;
-            break;
-        case InjectorType::GAS:
-            assert(pu.phase_used[BlackoilPhases::Vapour]);
-            this->surface_rates[pu.phase_pos[BlackoilPhases::Vapour]] = inj_surf_rate;
-            break;
-        case InjectorType::OIL:
-            assert(pu.phase_used[BlackoilPhases::Liquid]);
-            this->surface_rates[pu.phase_pos[BlackoilPhases::Liquid]] = inj_surf_rate;
-            break;
-        case InjectorType::MULTI:
-            // Not currently handled, keep zero init.
-            break;
-        }
+    switch (inj_controls.injector_type) {
+    case InjectorType::WATER:
+        assert(pu.phase_used[BlackoilPhases::Aqua]);
+        this->surface_rates[pu.phase_pos[BlackoilPhases::Aqua]] = inj_surf_rate;
+        break;
+    case InjectorType::GAS:
+        assert(pu.phase_used[BlackoilPhases::Vapour]);
+        this->surface_rates[pu.phase_pos[BlackoilPhases::Vapour]] = inj_surf_rate;
+        break;
+    case InjectorType::OIL:
+        assert(pu.phase_used[BlackoilPhases::Liquid]);
+        this->surface_rates[pu.phase_pos[BlackoilPhases::Liquid]] = inj_surf_rate;
+        break;
+    case InjectorType::MULTI:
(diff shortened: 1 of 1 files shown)


opm/simulators/wells/SingleWellState.cpp
```diff
@@ -245,31 +245,28 @@ void SingleWellState::update_injector_targets(const Well& ecl_well, const Summar
         return;
     }
 
-
-    if (inj_controls.cmode == Well::InjectorCMode::GRUP) {
-        this->bhp = this->perf_data.pressure_first_connection * bhp_safety_factor;
-        return;
+    // we initialize all open wells with some rate to avoid singularities
+    double inj_surf_rate = 0.0001;
```
bska, line 249: In this part of the code everything is SI so 1.0e-4 corresponds to 8.64 SM3/day (or 54.344 STB/day for that matter).  Is that a reasonable guess?

totto82: As long as it is >0 it does not seem to be very sensitive to this number. For producers 1000 SM3/day is used. We could use the same for injectors.

alfbr:
I have now tested this on the model with closing injector. I confirm that this PR does indeed solve the problem with initial convergence issues.

opm/simulators/wells/SingleWellState.cpp
totto82: @bska Can this be merged? Do you want me to change the initial guess to 1000 SM3/day?

totto82:
jenkins build this please

opm/simulators/wells/SingleWellState.cpp
totto82: > without hitting the procedure with a large hammer

I can hit it with a small hammer if you like. Any value  > epsilon should work. But 1000 SM3DAY is used for the producers. So what is done here make the code for the injectors consistent with producers. Since the well fraction are computed from the initial rates a fix like this is needed if you want to run a case where your injector is bhp or group controlled from the start. If I remember correctly we used to initialize with an epsilon many years ago in the old wellState code in opm-core.

GitPaean:
Do we have some information regarding the detailed setup/scenario that the PR is tackling? Similar to https://github.com/OPM/opm-simulators/pull/4809, let us gradually be more sophisticated when coming to well state initialization.

GitPaean:
> Since the well fraction are computed from the initial rates a fix like this is needed if you want to run a case where your injector is bhp or group controlled from the start.

Not suggesting in this PR. Maybe we should **in the future** have fraction information in the well state, similar to in the primary variables? In the well primary variables, fraction is used to handle zero rate situation naturally. 

Multiple PRs and techniques are trying to fix the fraction problems.

GitPaean:
The failures in https://ci.opm-project.org/job/opm-simulators-PR-builder/5250/testReport/ are sensitive when wells are re-opened. I do not think they should be in the way to getting PRs in if the changes are understandable, while it needs to be checked and analyzed. 

I do not think the change in the PR will change the behavoir significantly in wells are to be shut.

totto82:
I can do some more testing to check where the singularities happen.

totto82:
I have done some more testing, but I haven't found an alternative way of fixing this.

totto82:
Any hope for getting this merged? This fixes an urgent issue as verified by @alfbr?

GitPaean:
Now I read the code with more details. It looks like to give some initial rate for the wells under `GRUP` control. While I am not sure where the simulation gets stuck with?

GitPaean:
Please let me know if you want discuss about it with a short meeting.

totto82:
jenkins build this please

totto82:
@GitPaean Can this be merged? It is a necessary fix for running some CO2 cases.

GitPaean:
> @GitPaean Can this be merged? It is a necessary fix for running some CO2 cases.

The code itself is fine. The jenkins failures do not look worrying to me personally. The only thing I am not sure is whether it is wise to begin with a rather large rate. My intuition tells me that it might be safer to begin with a smaller rate, but it might have difficulty situations for either ways. 

With this PR, all the injectors with control modes other than RATE (GRUP, THP, BHP, RESV) will be affected by initializing with a rate (1000 m^3/day). Do you have some understanding which one actually fixes the running?

totto82:
The original PR had `double inj_surf_rate = 0.0001;` i.e. a smaller rate, I can revert it back to something smaller if you like. Any number larger than 0 seems to work.

atgeirr:
I believe we are in agreement that the new code is an improvement: the replaced code did not even guarantee the correct sign of the flow, if the reference depth was not the same as the first perforation, and I think that is why this was changed for producers previously.

The only worry I have is the rate of 1000 m^3/day, which is fairly large, and might lead to temporary changes in group targets during iterations, if the true rate of the well is significantly lower. So I suggest using a lower rate, for example 10 m^3/day. (But we should keep it as a clearly written m^3/day rate in the code as it is now.)

Other than that, we should update data and merge.

bska:
> So I suggest using a lower rate, for example 10 m^3/day. (But we should keep it as a clearly written m^3/day rate in the code as it is now.)

10 m^3/day could still be rather large if the user is trying to simulate a lab-scale situation with cm^3/hour as a reasonable order of magnitude. Of course, they're probably not using advanced group control in that case.

atgeirr:
> Of course, they're probably not using advanced group control in that case.

That would surprise me as well... I think we should not worry about possible lab scale problems at this point.

totto82:
jenkins build this please

bska:

atgeirr (approved):

bska:
jenkins build this opm-tests=1043 please

bska:
> Other than that, we should update data and merge.

I'm happy we agreed in the end.  The new reference solutions have been installed on the CI system, so I'll merge this to bring the overall system back to a consistent state.

PR #4808, opened 2023-08-22
Option for doing well control/status switching during local well solve

steink:
Add option _local-well-solve-control-switching_ that runs alternative local solve for well equations that potentially performs control/status switching such that converged well gets _correct_ control/status wrt current reservoir state.

- running with option true will likely have little effect on most test-cases (exept those with many or complex switchings), but wish to use this functionality subsequently in hopefully improved potential and implicit ipr calculations.
- the new functions `iterateWellEqWithSwitching` largely overlap with the existing `iterateWellEqWithControl`, but I think it's best to keep them separate not to clutter the existing versions


GitPaean:
jenkins build this please

GitPaean:
Sorry for the delay. I start looking this PR now.

steink:
Thanks. Not convinced all the logic is entirely correct yet, especially that involving stopping and re-opening. Started looking through the failed tests for this PR when option was set to true. So far I've not seen any considerable differences, but will need to look at all 95 cases more systematically.

opm/simulators/wells/MultisegmentWellEval.cpp
```diff
@@ -523,9 +523,6 @@ getResidualMeasureValue(const WellState& well_state,
         ++count;
     }
 
-    // if (count == 0), it should be converged.
```
GitPaean, line 526: Maybe we discuss before. Can you explain the circumstances that this assert will fail? Maybe we can go through different approach avoid breaching the design of the function.

steink: Yes, when `iterateWellEqWithSwitching` is called for operability-update, it will do one iteration even though equations already are converged. I had some issues avoiding this, but looking at it now, I'm not sure why. Will re-check.

steink: It's actually a bit tricky to make sure `getResidualMeasureValue` is never called for converged equations. We might have a situation where the the equations are converged but not all constraints are satified. In this case `iterateWellEqWithSwitching` will continue until all constraints are met. If we skip calling  `getResidualMeasureValue` for one of the iterations, this will break the stagnation/oscillations-checking.

GitPaean: Then from the code level without considering the exact scenario, there is something inconsistent between the `getWellConvergence` and the criteria used in the function `getResidualMeasureValue` including `getFiniteWellResiduals`. because if `getWellConvergence` is `true`, the function `getResidualMeasureValue` will not be reached.  It has been long time since the code got written, not totally sure whether they should be consistent. 

The following code can be related also, 
```c++
    // for stopped well, we do not enforce the following checking to avoid dealing with sign of near-zero values
    // for BHP or THP controlled wells, we need to make sure the flow direction is correct
    if (!baseif_.wellIsStopped() && baseif_.isPressureControlled(well_state)) {
        // checking the flow direction
        const double sign = baseif_.isProducer() ? -1. : 1.;
        const auto weight_total_flux = this->primary_variables_.getWQTotal() * sign;
        constexpr int dummy_phase = -1;
        if (weight_total_flux < 0.) {
            report.setWellFailed(
                    {CR::WellFailure::Type::WrongFlowDirection, CR::Severity::Normal, dummy_phase, baseif_.name()});
        }
    }
```


The above can be an isolated issue to be handled separately.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -727,6 +810,23 @@ namespace Opm
...
+                                    const WellState& well_state,
+                                    DeferredLogger& deferred_logger)
+    {
+        // only makes sense if we're using this parameter is true 
+        assert(this->param_.local_well_solver_control_switching_);
+        this->operability_status_.resetOperability();
+        WellState well_state_copy = well_state;
+        const auto& group_state = ebos_simulator.problem().wellModel().groupState();
+        const double dt = ebos_simulator.timeStepSize();
+        // equations should be converged at this stage, so only one it is needed
```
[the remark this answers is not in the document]
GitPaean: Okay. The problem I see how is that not initialization problem. It is more related to that the well_state at this stage is already problematic. 
```
well P bhp  is -496510.732701 bhp in the well state is -496510.732701
```

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -254,7 +254,80 @@ namespace Opm
...
+                this->stopWell();
+                return true;
+            } else {
+                bool changed = false;
+                auto& ws = well_state.well(this->index_of_well_);
+                const bool hasGroupControl = this->isInjector() ? inj_controls.hasControl(Well::InjectorCMode::GRUP) :
+                                                                  prod_controls.hasControl(Well::ProducerCMode::GRUP);
+
+                changed = this->checkIndividualConstraints(ws, summary_state, deferred_logger, inj_controls, prod_controls);
+                if (hasGroupControl) {
```
GitPaean, line 287: The logic here can be problematic. For example, if  `checkIndividualConstraints` is `true` and `checkGroupConstraints` is false. 

after the two calls, 
`changed` will false. 

Do you think we should change the line below to be 
`changed = changed  | this->checkGroupConstraints(well_state, group_state, schedule, summary_state, deferred_logger); 
`?
or we should rather use different if condition

`if (!changed  && hasGroupControl) ` to not call `checkGroupConstraints` ?

opm/simulators/wells/MultisegmentWellEval.cpp
GitPaean: Now I see, it is the following code that can break the assert. 

```c++
                if (switch_count > 0 && its_since_last_switch < min_its_after_switch) {
                    final_check = true;
                    its_since_last_switch = min_its_after_switch;
                } else {
                    break;
                }   
```
so it is good for now.

opm/simulators/wells/WellInterface_impl.hpp
steink: You're absolutely right. My intention was to check both, so I'd prefer 
`changed = changed  | this->checkGroupConstraints(...)`

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1412,7 +1412,174 @@ namespace Opm
...
+            // TODO: maybe we should have more sophisticated strategy to recover the relaxation factor,
+            // for example, to recover it to be bigger
+
+            if (is_oscillate || is_stagnate) {
+                // HACK!
+                std::ostringstream sstr;
+                if (relaxation_factor == min_relaxation_factor) {
+                    // Still stagnating, terminate iterations if 5 iterations pass.
+                    ++stagnate_count;
+                    if (false){//(stagnate_count == 6) {
```
GitPaean, line 1524: just a note here, with this, the relaxed tolerance will not be used.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1412,7 +1412,174 @@ namespace Opm
...
+
+            {
+                // getFinteWellResiduals returns false for nan/inf residuals
+                const auto& [isFinite, residuals] = this->getFiniteWellResiduals(Base::B_avg_, deferred_logger);
+                if (!isFinite)
+                    return false;
+
+                residual_history.push_back(residuals);
+                measure_history.push_back(this->getResidualMeasureValue(well_state,
+                                                                    residual_history[it],
```
GitPaean, line 1504: another note here. The function `detectOscillations` is not ready to handle `measure` of `0`.  So I will suggestion put this function call within a `if (!converged)` . And then the function `getResidualMeasureValue` keeps the final `assert`.

opm/simulators/wells/WellInterface_impl.hpp
bska: > ```C++
> changed = changed || this->checkGroupConstraints(...)
> ```

Just remember that `||` has a short-circuiting behaviour so if `changed == true` then
```C++
changed || this->checkGroupConstraints(...)
```
will be unconditionally `true` and `this->checkGroupConstraints()` will not be called.  Is that okay here?

GitPaean: Maybe we should do the following:

```c++
                const bool individual_changed = this->checkIndividualConstraints(ws, summary_state, deferred_logger, inj_controls, prod_controls);
               const bool group_changed = hasGroupControl && this->checkGroupConstraints(well_state, group_state, schedule, summary_state, deferred_logger);

              changed = individual_changed || group_changed;
```

GitPaean: The complication is that both `checkIndividualConstraints` and `checkGroupConstraints` have actions involved and not purely checking.

GitPaean:
closing this PR for https://github.com/OPM/opm-simulators/pull/4895 .

PR #4809, opened 2023-08-22
Ensure initial well solve is performed in corner case

steink:
In the case when a well was under zero-rate constraint in the previous step, an initial well solve will currently not be performed. It will however trigger _alternative-well-rate-init_ and this initialization can be very problematic if used as initial guess for balancing of networks. This suggest a rather hacky fix, happy to formulate it differently if there is a good suggestion.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -494,8 +494,11 @@ namespace Opm {
                 const bool event = report_step_starts_ && events.hasEvent(well->name(), effective_events_mask);
                 const bool dyn_status_change = this->wellState().well(well->name()).status
                         != this->prevWellState().well(well->name()).status;
-
-                if (event || dyn_status_change) {
+                bool prev_zero_rates = true;
+                for (const double& rate : this->prevWellState().well(well->name()).surface_rates){
+                    prev_zero_rates &= (rate == 0.0);
+                } 
+                if (event || dyn_status_change || prev_zero_rates){
                     try {
                         well->updateWellStateWithTarget(ebosSimulator_, this->groupState(), this->wellState(), local_deferredLogger);
                         well->calculateExplicitQuantities(ebosSimulator_, this->wellState(), local_deferredLogger);


GitPaean:
jenkins build this please

GitPaean:
Maybe we should include the situation you are trying to address in the event_mask, so the well will solve based on `event`, 
```c++
                const uint64_t effective_events_mask = ScheduleEvents::WELL_STATUS_CHANGE
                        + ScheduleEvents::INJECTION_TYPE_CHANGED
                        + ScheduleEvents::WELL_SWITCHED_INJECTOR_PRODUCER
                        + ScheduleEvents::NEW_WELL;
```

GitPaean:
We should try to get rid of the `alternative-well-rate-init `. For the situation `alternative-well-rate-init ` trying to fix, we better solve the well?

steink:
> We should try to get rid of the `alternative-well-rate-init `. For the situation `alternative-well-rate-init ` trying to fix, we better solve the well?

I think it (or something similar) needs to be there for the following reasons:
- provide initial guess for the well solve 
- provide non-zero explicit fractions that occasionally are needed for solving newly opened thp-wells

totto82:
>provide non-zero explicit fractions that occasionally are needed for solving newly opened thp-wells

Can this work? https://github.com/OPM/opm-simulators/pull/4806

steink:
> > provide non-zero explicit fractions that occasionally are needed for solving newly opened thp-wells
> 
> Can this work? #4806

Don't think so. I beleive `alternative-well-rate-init` does a decent job in providing a fall-back for fractions used in vfp. Particular issue here is that a zero-rate-controlled well (with very large rates from `alternative-well-rate-init`) ends up beeing used in the network balancing.

Maybe better than my original suggestion would be to just never do the `alternative-well-rate-init` for zero-rate-controlled wells. This would fix this particular issue (allthough it might be a good idea to always initially solve wells that are part of a network that needs balacing).

steink:
Seems that a better solution for this problem is to just skip `alternative-well-rate-init` for stopped/zero-target wells. Will suggest that in stead.

PR #4815, opened 2023-08-24
Add FLO as supported option of item 6 of GRUPNET

plgbrts:


bska (approved):
Thanks.  PR OPM/opm-common#3651 was merged into the master branch, so we can activate this now.

PR #4817, opened 2023-08-28
Skip initialization of rates for stopped/zero-rate-target wells

steink:
For stopped/zero-rate-target wells the _alternative_well_rate_init_-procedure currently returns unscaled well-rates resulting from a zero-bhp condition. I don't think this ever is helpful, and in particular it can lead to convergence failures for network balancing (since the initialized rates can off by orders of magnitude ). 

This pull request simply suggests to skip the _alternative_well_rate_init_-procedure for stopped/zero-rate-target wells.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -435,13 +435,15 @@ namespace Opm {
         // we need the inj_multiplier from the previous time step
         this->initInjMult();
 
+        const auto& summaryState = ebosSimulator_.vanguard().summaryState();
         if (alternative_well_rate_init_) {
             // Update the well rates of well_state_, if only single-phase rates, to
             // have proper multi-phase rates proportional to rates at bhp zero.
             // This is done only for producers, as injectors will only have a single
             // nonzero phase anyway.
             for (auto& well : well_container_) {
-                if (well->isProducer()) {
+                const bool zero_target = well->stopppedOrZeroRateTarget(summaryState, this->wellState());
+                if (well->isProducer() && !zero_target) {
                     well->updateWellStateRates(ebosSimulator_, this->wellState(), local_deferredLogger);
                 }
             }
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

GitPaean:
The PR is sensible. Mostly, it will avoid the using of the following code, 

```c++
1174         if (nonzero_rate_index == -1) {
1175             // No nonzero rates.
1176             // Use the computed rate directly
1177             for (int p = 0; p < this->number_of_phases_; ++p) {
1178                ws.surface_rates[p] = well_q_s[this->flowPhaseToEbosCompIdx(p)];
1179             }
1180             return;
1181         }
```

And the rates were calculated based on
```c++
1171         // Calculate the rates that follow from the current primary variables.
1172         std::vector<double> well_q_s = computeCurrentWellRates(ebosSimulator, deferred_logger);
```

Following the discussion, the function `computeCurrentWellRates` possibly (**will double check**) uses uninitialized primary variables, which is not good.

GitPaean:
At least from the testing results from https://github.com/OPM/opm-simulators/pull/4818, (I did not check all the scenarios), two scenarios are affected by alternative_well_rate_init are 1. new producers with roughly initialized single phase rate, and 2. wells with zero rate limits.  

For the second scenario, we should not use the rates from `computeCurrentWellRates()` at all, which uses **zero** bhp value if entering from the alternative_well_rate_init.  This makes the PR is really good to avoid this types of scenario. 

For the first scenario, it might not do much harm to do it, but we probably can do it more elegantly.

GitPaean:
@steink , please make the PR to be ready for review if you are done testing.  And also, can you check the three regression failures? From the jenkins output, they are not worrying. I think you can just check the well curves.

GitPaean:
> for NORNE_ATW2013_B1H_RE-PERF the bhps differ up to 0.5 bar (pic). Haven't quite figured out why this happen, except that master initializes surface rates to huge values while the well is under zero rate control. Think therefore this failure should also be OK.


My guess on this is that the pressure drop between connections are changed since different rates are used to calculate the density in the wellbore. But I do not know what should be closer to the ground truth either since it is explicit?

GitPaean:
I am preparing to get this PR in.

GitPaean:

GitPaean (approved):
This is a relatively non-controversial change. Rates calculated from zero BHP is not great to initialize a well with zero rate limits. Which might bring bad consequence when initial rates are important, like THP control or network modeling.

GitPaean:
jenkins build this opm-tests=1031 please

GitPaean:
The reference in opm-tests have been installed. I am merging this PR in now.

PR #4819, opened 2023-08-28
Implement support for GECON keyword

hakonhagland:
Implement support for item 2 and 3 of the GECON keyword. See also companion PR https://github.com/OPM/opm-common/pull/3657 in opm-common for more information.

### DIFF
--- a/opm/simulators/wells/GroupEconomicLimitsChecker.cpp
+++ b/opm/simulators/wells/GroupEconomicLimitsChecker.cpp
@@ -1,0 +1,311 @@
+/*
+  Copyright 2023 Equinor ASA.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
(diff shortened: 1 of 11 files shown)


opm/simulators/wells/GroupEconomicLimitsChecker.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+/****************************************
+ * Public methods in alphabetical order
+ ****************************************/
+
+void
+GroupEconomicLimitsChecker::
+activateEndRun()
+{
+
```
totto82, line 66: Since this functionality is missing. Please add to the list in https://github.com/OPM/opm-simulators/blob/c6e30fbf3c3a9677ec245b004ce55358b9021c40/opm/simulators/utils/PartiallySupportedFlowKeywords.cpp#L1

and a throw here. 

The same goes for workover etc. that is missing.

totto82 (commented):
Thanks for the implementation. This and the asscosiated PR in opm-common can be merged when the issue with partially supported keywords is fixed.

hakonhagland:
@totto82 Thanks for the review. I have added a commit to address the comments.

mbmfer:
Hi, I can see that this issue has been addressed in the past. and additional code has been posted in the  Can you please give me a hand with the process of adding the GECON functionality to the 2023.04 version of the Flow engine?. Thanks

PR #4822, opened 2023-08-29
Apply Regional Multipliers to Initial Transmissibilities

bska:
This commit adds a new member function,
```C++
EclTransmissibility<>::applyNncMultreg_()
```
which applies regional transmissibility multipliers such as those entered in the `MULTREGT` keyword to the explicit input NNCs.  We make the application conditional on a new parameter, default value `false`, and pass `true` as the argument from `finishInit()`.

Along with `TransMult::getRegionMultiplierNNC()`, this implements all known connection behaviours for inter-region connections. Multipliers internal to a region in `MULTREGT` are not yet supported.

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -119,15 +119,27 @@ public:
      * either but at least it seems to be much better.
      */
     void finishInit(const std::function<unsigned int(unsigned int)>& map = {})
-    { update(true,map); }
+    {
+        this->update(true, map, /*applyNncMultRegT = */ true);
+    }
 
     /*!
      * \brief Compute all transmissibilities
      *
-     * \param global If true, update is called on all processes
-     * Also, this updates the "thermal half transmissibilities" if energy is enabled.
+     * \param[in] global Whether or not to call \c update() on all
+     *   processes.  Also, this updates the "thermal half
+     *   transmissibilities" if energy is enabled.
+     *
+     * \param[in] map Undocumented.
+     *
+     * \param[in] applyNncMultRegT Whether or not to apply regional
+     *   multipliers to explicit NNCs.  Explicit NNCs are those entered
+     *   directly in the input data, e.g., through the NNC/EDITNNC/EDITNNCR
+     *   keywords, or the result of generating connections to or within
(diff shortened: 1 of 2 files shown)


bska:
This PR depends on, and contains, upstream PR #4821.  We also require the new function `TransMult::getRegionMultiplierNNC()` which is introduced in OPM/opm-common#3655.  I will keep this  PR in draft mode until both of those PRs have been merged into their respective master branches and this PR is ready for review.

bska:
jenkins build this opm-common=3655 please

bska:
jenkins build this opm-common=3655 please

bska:
> This PR depends on, and contains, upstream PR #4821. We also require the new function `TransMult::getRegionMultiplierNNC()` which is introduced in [OPM/opm-common#3655](https://github.com/OPM/opm-common/pull/3655). I will keep this PR in draft mode until both of those PRs have been merged into their respective master branches and this PR is ready for review.

The upstream PRs have been merged into the master branches.  I'm marking this as "ready for review"

bska:
jenkins build this please

blattms (approved):
Thanks, looks good.

blattms:
> Multipliers internal to a region in MULTREGT are not yet supported.

For the documentation: Can you clarify this please. Do you mean the case where the indices of the two regions passed to MULTREGT are the same?

bska:
>> Multipliers internal to a region in `MULTREGT` are not yet supported.
>
> For the documentation: Can you clarify this please. Do you mean the case where the indices of the two regions passed to `MULTREGT` are the same?

Yes, that is exactly it.  A specification of the form
```
MULTREGT
  42 42 0.123 ... /
/
```
requests that the transmissibility of connections within region 42, **and** between region 42 and any of that region's neighbouring regions, be multiplied by `0.123`, provided the connections match the selection defined by `...`.  If, for example, items 4 through 6 are defaulted, then all connections in region 42 of the `MULTNUM` region set and any connections on the boundary between region 42 and another region would be multiplied by `0.123`.

Flow does not support that case/specification as of this time.

blattms:
Perfect. thanks a lot.

PR #4823, opened 2023-08-30
changed: put ecl time stepping parameters in separate struct

akva2:
this to allow reuse in ebos simulators


bska (approved):
Thanks.  This looks good to me and I'll merge into master.

PR #4816, opened 2023-08-25
Temp diffusion

hnil:
-- set tpfa lineariser for gasoil cases with energy and diffusion
-- for now all energy cases also have enabled diffusion

This PR depends on https://github.com/OPM/opm-models/pull/824 .

PR #4825, opened 2023-08-31
add tpfa as lineariser for gasoil_energy (with diffusion) and gas oil diffusion

GitPaean:
-- set tpfa lineariser for gasoil cases with energy and diffusion
-- for now all energy cases also have enabled diffusion

This PR replaces https://github.com/OPM/opm-simulators/pull/4816 for convenience. 

And the PR depends on https://github.com/OPM/opm-models/pull/825


totto82 (approved):
Fine for me.

PR #4826, opened 2023-08-31
Compute resv rates based on pressure and temperature in the well for injectors in thermal cases

totto82:


totto82:
jenkins build this please

totto82:
jenkins build this please

GitPaean:
Do you have some example values/curves to explain the finding when developing this PR?

totto82:
It worries me a bit that these reservoir rates are written to the restart files. Wouldn't it be more sensible to write the actual reservoir rates computed the well model, rather than the rates converted using some average values of pressure etc? @bska Any comments on this?

bska:
The restart file is supposed to hold the well's [current `RESV` **target** (limit)](https://github.com/OPM/opm-common/blob/aaa8af923ceeaaff60e788c224b86c77ea685498/opm/output/eclipse/VectorItems/well.hpp#L287-L288) value in its `SWEL` array.  In prediction mode we just use the value entered in `WCONPROD`, but in a history matching well we have to compute the target value through some kind of conversion from surface rates.  If there's a better way to do that, then by all means use that.

totto82:
> The restart file is supposed to hold the well's [current RESV target (limit)](https://github.com/OPM/opm-common/blob/aaa8af923ceeaaff60e788c224b86c77ea685498/opm/output/eclipse/VectorItems/well.hpp#L287-L288) value in its SWEL array

That should be consistent with how we compute the rates know. From my point of view this PR can be merged after an update of data (i.e after the release)

alfbr:
A review/merge would be great here. This change is necessary for CCS cases.

totto82:
@bska Are we happy with this change? Should we start the update of the reference data?

bska (approved):
Yes, this looks good to me.  I'll update the reference solutions and merge this now.

bska:

bska:
jenkins build this opm-tests=1081 please

bska:
The updated reference solutions have all been installed on the CI system.  I'm merging this now to bring the overall system back to a consistent state.

PR #4827, opened 2023-08-31
Damaris fix for deadlock at end of simulation when run with Damaris 

jcbowden:
A smaller set of changes relating to Damaris and its initialization.

- Corrects the order for testing if a rank is a simulation rank (isSimulationRank_) after the initialize_() call.
- Adds a check that the number of simulation ranks is > 1 when run with Damaris enabled (command line option: --enable-damaris-output=true), if only 1 rank then Damaris is disabled.
- Places guard around code (using if (isSimulationRank_) )  that stops the Damaris server ranks running the simulation rank code when the Damaris server ranks return from the damaris_start() call. The Damaris ranks return from damaris_start() to the main code after simulation ranks call damaris_stop().

### DIFF
--- a/opm/simulators/flow/Main.hpp
+++ b/opm/simulators/flow/Main.hpp
@@ -3,7 +3,8 @@
   Copyright 2014 Dr. Blatt - HPC-Simulation-Software & Services
   Copyright 2015 IRIS AS
   Copyright 2014 STATOIL ASA.
-
+  Copyright 2023 Inria
+  
   This file is part of the Open Porous Media project (OPM).
 
   OPM is free software: you can redistribute it and/or modify
@@ -142,8 +143,8 @@ public:
     int runDynamic()
     {
         int exitCode = EXIT_SUCCESS;
-        if (isSimulationRank_) {
-            if (initialize_<Properties::TTag::FlowEarlyBird>(exitCode)) {
+        if (initialize_<Properties::TTag::FlowEarlyBird>(exitCode)) {
+            if (isSimulationRank_) {
                 return this->dispatchDynamic_();
             }
         }
@@ -155,8 +156,8 @@ public:
     int runStatic()
     {
         int exitCode = EXIT_SUCCESS;
-        if (isSimulationRank_) {
-            if (initialize_<TypeTag>(exitCode)) {
+        if (initialize_<TypeTag>(exitCode)) {
+            if (isSimulationRank_) {
                 return this->dispatchStatic_<TypeTag>();
             }
         }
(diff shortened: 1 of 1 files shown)


jcbowden:
This pull request is to replace part of the code being presented in draft PR https://github.com/OPM/opm-simulators/pull/4759.
  
Other code additions present in PR 4759 will also be added as separate PR to have the changes added in manageable sized chunks.

atgeirr:
jenkins build this please

opm/simulators/flow/Main.hpp
```diff
@@ -325,77 +326,89 @@ class Main
 
 #if HAVE_DAMARIS
         enableDamarisOutput_ = EWOMS_GET_PARAM(PreTypeTag, bool, EnableDamarisOutput);
+        // Reset to false as we cannot use Damaris if there is only one rank.
+        if ((enableDamarisOutput_ == true) && (EclGenericVanguard::comm().size() == 1)) {
+            std::string msg ;
+            msg = "\nUse of Damaris (command line argument --enable-damaris-output=true) has been dissabled for run with only one rank.\n" ;
```
atgeirr, line 332: Typo: "dissabled"

opm/simulators/flow/Main.hpp
```diff
@@ -325,77 +326,89 @@ class Main
...
+            OpmLog::info(msg);
```
atgeirr, line 333: Make this a warning() instead of info() message.

opm/simulators/flow/Main.hpp
```diff
@@ -325,77 +326,89 @@ class Main
...
         }
 #endif // HAVE_DAMARIS
 
-        int mpiRank = EclGenericVanguard::comm().rank();
-        outputCout_ = false;
-        if (mpiRank == 0)
-            outputCout_ = EWOMS_GET_PARAM(PreTypeTag, bool, EnableTerminalOutput);
+        // Guard for when the Damaris core(s) return from damaris_start() 
+        // which happens when damaris_stop() is called in main simulation
+        if (isSimulationRank_) {
```
atgeirr, line 345: Doing it in this way indents all the existing code, without changing it. I think it would be better to write
```
if (isSimulationRank_) {
    exitCode = EXIT_SUCCESS;
    return true;
}
```
and then leave the rest of the function unchanged.

atgeirr (commented):
This is good! Just requesting a few minor adjustments.

opm/simulators/flow/Main.hpp
jcbowden: I do not find that as clear, however functionality is the same and it is good to keep indentation under control, so done.

opm/simulators/flow/Main.hpp
jcbowden: This test and change of: *enableDamarisOutput_ = false* does not propagate through to eclwriter.hh, which again tests for the value of the command line argument (using *EWOMS_GET_PARAM(PreTypeTag, bool, EnableDamarisOutput)*).
  
This causes some calls to the Damaris API to be called via the EclWriter class and tests for failure blemish the output of the OPM-Flow run.
  
Is there a way to set the EWOMS parameter? something like a EWOMS_SET_PARAM() ?

atgeirr: I think there is no way to set it. We might add it, but I will try a different fix. For now I'll run Jenkins on this and merge if green. Thanks for making the changes I requested!

atgeirr:

opm/simulators/flow/Main.hpp
jcbowden: Ok for that. 

I'll try to get another pull request with the larger set of changes available soon - I am trying to split the work into two, however I am running into extra issues, so it may be a bigger, single request.

PR #4832, opened 2023-09-04
Implement Forchheimer term in wellIndex

totto82:
Depends on https://github.com/OPM/opm-common/pull/3661

The Forchheimer term is added as a additional skin term that depends on the rate (explicitly). If the rate changes significantly we may want to solve this implicitly. That will require more refactoring / coding to implement. 

We use the viscosity of the connection cell here. In the dake book (Fundamentals of Reservoir Engineering) he suggest using the viscosity evaluated at the bhp pressure.

### DIFF
--- a/opm/simulators/wells/SingleWellState.cpp
+++ b/opm/simulators/wells/SingleWellState.cpp
@@ -51,6 +51,7 @@ SingleWellState::SingleWellState(const std::string& name_,
     for (std::size_t perf = 0; perf < perf_input.size(); perf++) {
         this->perf_data.cell_index[perf] = perf_input[perf].cell_index;
         this->perf_data.connection_transmissibility_factor[perf] = perf_input[perf].connection_transmissibility_factor;
+        this->perf_data.connection_d_factor[perf] = perf_input[perf].connection_d_factor;
         this->perf_data.satnum_id[perf] = perf_input[perf].satnum_id;
         this->perf_data.ecl_index[perf] = perf_input[perf].ecl_index;
     }
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -22,6 +22,7 @@
 #include <opm/common/Exceptions.hpp>
 
 #include <opm/input/eclipse/Schedule/ScheduleTypes.hpp>
+#include <opm/input/eclipse/Schedule/Well/WDFAC.hpp>
 #include <opm/simulators/utils/DeferredLoggingErrorHelpers.hpp>
 #include <opm/simulators/wells/GroupState.hpp>
 #include <opm/simulators/wells/TargetCalculator.hpp>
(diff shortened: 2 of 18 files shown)


opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1202,6 +1203,35 @@ namespace Opm
         }
     }
 
+    template <typename TypeTag>
+    double
+    WellInterface<TypeTag>::
+    wellIndex(const int perf, const IntensiveQuantities& intQuants, const SingleWellState& ws) const {
```
bska, line 1209: This function does a lot of unnecessary work for the typical case of `D = 0`.  Is there no other way of structuring this work to avoid that complexity?

bska (commented):
This is a good start, just one small note really.  It is however not clear to me that the new dynamic behaviour will be reflected in the `CTFAC` summary vector.  Come to think of it, neither will the `rockCompTransMultiplier<>()`, but that's not a new problem.

If we do want to have the new values output as `CTFAC` then we'll need to somehow convey them to [`WellState::reportConnections()`](https://github.com/OPM/opm-simulators/blob/18f0d8fd27faf582fb561334aec71be3e3310279/opm/simulators/wells/WellState.cpp#L600).

totto82:
> It is however not clear to me that the new dynamic behaviour will be reflected in the CTFAC summary vector. Come to think of it, neither will the rockCompTransMultiplier<>(), but that's not a new problem.

I have added rockCompTrans to the implementation. For testing I also updated CTFAC with the multiplier while adding output of CDFAC. I added a seperate function for computing the output values  rather storing the value every time they are computed.

blattms:
> > It is however not clear to me that the new dynamic behaviour will be reflected in the CTFAC summary vector. Come to think of it, neither will the rockCompTransMultiplier<>(), but that's not a new problem.
> 
> I have added rockCompTrans to the implementation. For testing I also updated CTFAC with the multiplier while adding output of CDFAC. I added a seperate function for computing the output values rather storing the value every time they are computed.

Would it be possible to factor the multiplication with  rockCompTransMultiplier<>() for CTFAC to a separate PR? Then we would have that more quickly which would make some people happy.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1317,6 +1318,84 @@ namespace Opm
...
+            //double h = connection.connectionLength();
+            //double k = connection.Kh()/h * trans_mult;
+            double Kh = connection.Kh()* trans_mult;
+            double Ke = connection.Ke()* trans_mult;
+            double h = Kh / Ke;
+            double rw = connection.rw();
+            double d = wdfac.useConnectionDFactor()? connection.dFactor() : wdfac.getDFactor(rho, mu, Ke, phi, rw, h);
+            double scaling = 3.141592653589 * Kh;
+            perf_data.connection_d_factor[perf] = d;
+            perf_data.connection_transmissibility_factor[perf] = this->well_index_[perf] * trans_mult;
```
bska, line 1381: It's not clear to me if this breaks the `CTFAC` output of runs which have both `WDFAC` and `WELPI` for the same well.  I haven't managed to perform all of the requisite analysis, but I think it's possible you'll only get one of the updates (`WDFAC` or `WELPI`) but not both, and in that case it is not clear right now to me which update will "win".  I'd appreciate some comments here.

bska (commented):
I think there's one error here&ndash;removing `phase_mixing_rate` from `SingleWellState::operator==()`&ndash;and one potential problem for runs which combine `WDFAC`/`WDFACCOR` with `WELPI` for the same well.

I also don't quite understand the reasoning behind commit e8ed2c973648ccfa684196a24297812376f3ec35.  The well's BHP reference depth might be very different from that of its connections, e.g., if the pressure gauge is located several hundred metres above the formation. In that case there will typically be pressure differences of several 10s of bars between the BHP and the connection pressures.  Does using the BHP give a representative value for viscosity calculations in that case?

opm/simulators/wells/WellInterface_impl.hpp
totto82: Ops. Line 1381 should be removed here. I moved it to a separate function and forgot to remove it from here.

opm/simulators/wells/SingleWellState.cpp
```diff
@@ -301,7 +302,6 @@ bool SingleWellState::operator==(const SingleWellState& rhs) const
            this->bhp == rhs.bhp &&
            this->thp == rhs.thp &&
            this->temperature == rhs.temperature &&
-           this->phase_mixing_rates == rhs.phase_mixing_rates &&
```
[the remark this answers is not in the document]
totto82: No good reason. I will fix.

totto82:
>I also don't quite understand the reasoning behind commit https://github.com/OPM/opm-simulators/commit/e8ed2c973648ccfa684196a24297812376f3ec35.

That is how it is formulated in the the Dake book (Fundamentals of Reservoir Engineering). Yes it significantly effects the results. @tskille has done some testing and is happy with the current approach of evaluating the viscosity using bhp.

tskille:
> happy with the current approach of evaluating the viscosity using bhp

I think the comment from @bska is good. The model I used to evaluate this did not had datum depth defaulted. Defaulted datum depth is equal to cell depth for first connection in the well. 

I believe that the best approach would be to use "bottom hole pressure" at connection grid cell depth for each connection.

totto82:
> I believe that the best approach would be to use "bottom hole pressure" at connection grid cell depth for each connection.

I agree that it make more sense to use the connection pressure. I will update.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1317,6 +1318,84 @@ namespace Opm
...
+
+        for (int perf = 0; perf < this->number_of_perforations_; ++perf) {
+            const int cell_idx = this->well_cells_[perf];
+            const auto& intQuants = simulator.model().intensiveQuantities(cell_idx, /*timeIdx=*/ 0);
+            const double trans_mult = simulator.problem().template rockCompTransMultiplier<double>(intQuants, cell_idx);
+            // viscosity is evaluated at connection pressure
+            const double connection_pressure = ws.perf_data.pressure[perf];
+            const double mu = FluidSystem::gasPvt().viscosity(this->pvtRegionIdx(), ws.temperature, connection_pressure, getValue(intQuants.fluidState().Rv()), getValue(intQuants.fluidState().Rvw()));
+            const double phi = getValue(intQuants.porosity());
+            const auto& connection = this->well_ecl_.getConnections()[perf];
```
bska, line 1373: I didn't pay enough attention to this before, my apologies, but this should probably be
```C++
const auto& connection = this->well_ecl_.getConnections()[perf_data.ecl_index[perf]];
```
instead.  Otherwise there will be an inconsistency in the case of individually shut/stopped connections which are not represented in `perf_data`.  Or if we have distributed wells.

bska (commented):
Thank you for the updates, this looks good to me now.  I did notice one issue that I failed to see earlier.  The `perf` index is, in general, not directly usable as a subscript into `Well::getConnections()`.  The `SingleWellState` object holds values for the open/flowing connections on this rank meaning that If we have distributed wells or certain connections are individually shut/stopped then the `perf_data` will be a subset of the full connection set.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1317,6 +1318,82 @@ namespace Opm
...
+        const auto& wdfac = this->well_ecl_.getWDFAC();
+        if (!wdfac.useDFactor()) {
+            return this->well_index_[perf] * trans_mult;
+        }
+
+        if constexpr (! Indices::gasEnabled) {
+            return this->well_index_[perf] * trans_mult;
+        }
+        // for gas wells we may want to add a Forchheimer term if the WDFAC or WDFACCOR keyword is used
+        const auto& connection = this->well_ecl_.getConnections()[perf];
```
bska, line 1335: You'll need `getConnections()[perf_data.ecl_index[perf]]` here too.

bska (approved):

PR #1646, opened 2018-11-16
Remove legacy simulators

akva2:
Moved to a separate repo.

PR #818, opened 2016-09-12
Frankenstein merge master

andlaus:
this fixes some serious screw-ups, almost all of them were caused by recent changes in the master
branch:

- there were methods added which depend on the types `V` and
`DataBlock`. these do not make much sense in the context of the
frankenstein simulator. Also, these types are defined globally for the
whole Opm namespace in `BlackoilModelBase_impl.hpp` (which should be
prosecuted as a fellony IMO)! Besides this, their names are useless;
'V' is the letter which comes after `U` in the alphabet and when it
comes to computers basically everything can be seen as a chunk of data
(i.e., a `DataBlock`).
- it seems like the new and shiny dense-AD based well model was never
compiled with assertations enabled, at least some asserts referenced
non-existing variables. Also in the new well code, some asserts within eigen
trigger  if debugging options are enabled. This is rarely a good thing because 
it seems like some out of bounds objects are accessed in arrays.
@totto82: did you already fix this in your recent cleanup branch?  if no, could you please have a look? To reproduce, you just need to compile everything with debug options and run Norne.
- the recent output-related API changes were pretty unfortunate
because they had the effect of tying the (sub-optimal, IMO) internal
structure of the model even closer to the output code: as far as I can
see, `rq` does only make sense if the model works *exactly* like
BlackoilModelBase and friends. (for flow_ebos, this could be
replicated, but first it would be another unnecessary conversion step
and second, most of the quantities in `rq` are of type `ADB` and much
of the "frankenstein" excercise is devoted to getting rid of these.) I
thus reverted back to an old version of the output code and created a
`frankenstein` branch in my personal `opm-output` github fork.

PR #4837, opened 2023-09-05
Remove Unused Functions

bska:
These functions became unused in commits 952ccf8338 (#818) and f027262ec4 (#1646), but their definitions remained in place.  We no longer need them.


bska:
jenkins build this please

PR #4839, opened 2023-09-06
[testing] setting the bhp to be bhp_limit in updateWellStateWithTarget

GitPaean:
to avoid uninitialized bhp 0. to be used.

based on finding when investigating PR https://github.com/OPM/opm-simulators/pull/4772 .

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -758,6 +758,8 @@ namespace Opm
         {
             const auto& controls = well.injectionControls(summaryState);
 
+            ws.bhp = controls.bhp_limit;
+
             InjectorType injectorType = controls.injector_type;
             int phasePos;
             switch (injectorType) {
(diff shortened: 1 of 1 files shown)


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -880,6 +882,9 @@ namespace Opm
         {
             const auto current = ws.production_cmode;
             const auto& controls = well.productionControls(summaryState);
+
+            ws.bhp = controls.bhp_limit;
+
```
totto82, line 887: If your well is rate controlled the bhp_limit can be very far from the solution. What about just scaling it as we do with the rate. ws.bhp *= controls.oil_rate / current_rate. If the rates are close enough that is a good approximation.

GitPaean: >  ws.bhp *= controls.oil_rate / current_rate. 

That will also be far, because the one related (proportionally) to the rates are drawdown, which is related to the difference between the wellbore pressure and the reservoir pressure. 

But thanks for the tips.

GitPaean:
There is no capacity to check all the failures. A smaller change https://github.com/OPM/opm-simulators/pull/4841 will be used.

PR #4846, opened 2023-09-07
dont mark inoperable wells uncoverged

totto82:
Should only affect cases where --enable-well-operability-check-iter=true (default is false).

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1775,7 +1775,7 @@ namespace Opm {
                 local_report += well->getWellConvergence(
                         summary_state, this->wellState(), B_avg, local_deferredLogger,
                         iterationIdx > param_.strict_outer_iter_wells_);
-            } else {
+            } else if (!well->isSolvable()) {
                 ConvergenceReport report;
                 using CR = ConvergenceReport;
                 report.setWellFailed({CR::WellFailure::Type::Unsolvable, CR::Severity::Normal, -1, well->name()});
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1775,7 +1775,7 @@ namespace Opm {
                 local_report += well->getWellConvergence(
                         summary_state, this->wellState(), B_avg, local_deferredLogger,
                         iterationIdx > param_.strict_outer_iter_wells_);
-            } else {
```
GitPaean, line 1778: With this added condition, why should we distinguish the non-operable wells and non-solveable wells in this context?

totto82: For unsolvable wells we would like to cut the time step and retry it. For inoperable wells we would just like to close them and continue.

GitPaean: What if the operability can be updated during the iteration? So my suggestion here is that you change the condition to be

`(!well->param_.check_well_operability_iter_ && !well->isSolvable()) `.   Some accessing function might need to be added to make the code work. 

Or not relying on the negative logic (which is hard to catch), changing `!well->isSolvable()` to be `well->isOperable()`. 

Then you need to refactor to split the function `isOperableAndSolvable()` to be `isOperable()&&isSolvable()`.

totto82: > What if the operability can be updated during the iteration? 

The motivation for this PR to handle cases where the operability is changed during iterations without marking the well as uncoverged which again will lead to cutting of timestep. So with this change inoperable wells are allowed to close without cutting the timesteps while unsolvable wells are not. 

I think I prefer to not split isOperableAndSolvable because it should usually be used together. It easy to forget one of them.

PR #4847, opened 2023-09-07
Reformulate ipr to handle cross-flow in wells

totto82:
This is only for standard wells. A similar reformulation is needed for msw wells.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -775,28 +775,73 @@ namespace Opm
             const double h_perf = this->connections_.pressure_diff(perf);
             const double pressure_diff = p_r - h_perf;
 
-            // Let us add a check, since the pressure is calculated based on zero value BHP
-            // it should not be negative anyway. If it is negative, we might need to re-formulate
-            // to taking into consideration the crossflow here.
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

steink:
Just a comment: for ms wells I have a version in the pipeline that computes ipr using the Jacobian of the converged well-equations to obtain dq/dp. It can also be used for standard wells, but there it is less accurate due to the explicit treatment of dp along the well.

totto82:
> Just a comment: for ms wells I have a version in the pipeline that computes ipr using the Jacobian of the converged well-equations to obtain dq/dp. It can also be used for standard wells, but there it is less accurate due to the explicit treatment of dp along the well.

Sounds like a good solution. A problem with the current approach is that it basically need to duplicate the code that computes the rate in the well model. Looking forward to the PR. Do you think this should be merged first or should we wait for your solution. A draft PR from you would be nice. I am eager to test it on the model that motivated my to provide this fix.

steink:
> Do you think this should be merged first or should we wait for your solution. A draft PR from you would be nice. I am eager to test it on the model that motivated my to provide this fix.
Might take some time before this is ready for merging, but I can make a quick PR for testing on standard wells.

PR #4848, opened 2023-09-07
Avoid adding/substracting GCONSUMP values per MPI process..

vkip:


vkip:
jenkins build this please

atgeirr (approved):
I believe the logic here is correct: all the other functionalities do not involve getting a (global) value from the Schedule, that gets added or subtracted from the local sums. GCONSUMP is an exception in that respect. (At least I could not think of another.)

PR #4845, opened 2023-09-06
Updated pybind11

hakonhagland:
Updated pybind11 from version 2.3.0 to version 2.11.0

Basically just replaced the folder `python/pybind11`: https://github.com/OPM/opm-simulators/tree/master/python/pybind11 with this tar ball:
https://github.com/pybind/pybind11/archive/refs/tags/v2.11.0.tar.gz

See also: https://github.com/OPM/opm-common/pull/3666

PR #4849, opened 2023-09-07
changed: vendor pybind11 using FetchContent

akva2:
This is an alternative to https://github.com/OPM/opm-simulators/pull/4845

Rather than hosting the source in-tree, vendor it using the FetchContent mechanism .


bska (commented):
Pruning the bundled copy of `pybind11` is very welcome, so thank you very much!  I am not familiar with the workings of `FetchContent`, so I'll ask a couple of probably stupid questions.

1. Will `FetchContent_MakeAvailable(pybind11)` download to (a subdirectory of) `${CMAKE_SOURCE_DIR}` (or `${CMAKE_CURRENT_SOURCE_DIR}`) or does it go to `${CMAKE_BINARY_DIR}` instead?
2. Is there a way to control the location to which the downloaded package will be extracted?  For instance, I imagine that a lot of the developers have a setup of the form `${ROOT}/opm-common`, `${ROOT}/opm-simulators`, etc., and if I could convince the `FetchContent` module to look for and/or put the pybind sources in `${ROOT}/3rdparty/pybind11` or some such location I could probably get away with downloading it once instead of potentially once for every module.

Am I making any sense at all?

akva2:
i gotta run in barnehage now, but i'll quickly answer;
1) it downloads in build directory

all of this is configurable https://cmake.org/cmake/help/latest/module/FetchContent.html i don't know exactly the answer to your question but i imagine it is possible.

if you use the super-build it only downloads it once. if you build common and simulators separately it will by default be downloaded once per build-directory.

bska:
> if you use the super-build it only downloads it once. if you build common and simulators separately it will by default be downloaded once per build-directory.

Okay, understood.  In that case I think I'll take slightly different action by installing a copy to a directory on my `CMAKE_PREFIX_PATH` and deferring to `find_package()` instead of `FetchContent` for my setup.  I typically build several configurations (release/debug, sequential/MPI, ...) and I'd prefer not to have a copy of pybind11 for each of those.

python/CMakeLists.txt
```diff
@@ -1,9 +1,12 @@
...
 
 if(NOT pybind11_FOUND)
-  add_subdirectory( pybind11 )
-  # SYSTEM is supported for embedded pybind source (>= 2.3.0)
-  set(PYBIND11_SYSTEM "SYSTEM")
+  include(FetchContent)
+  FetchContent_Declare(pybind11
+                       URL https://github.com/pybind/pybind11/archive/refs/tags/v2.11.0.tar.gz
+                       URL_HASH SHA512=10edfdc499ed70234de0796f79b9f3a7a14d4baf01fb09fedbd9126e0a13d9cdeb213950c53a8f7a67cd6049b1db31efacfe2192098236b242ad3685967ea907
+                       DOWNLOAD_EXTRACT_TIMESTAMP ON)
```
hakonhagland, line 8: Same comment as for https://github.com/OPM/opm-common/pull/3671, it works if   `DOWNLOAD_EXTRACT_TIMESTAMP ON` is put before `URL` and `URL_HASH` properties (tested on Ubuntu 22.04, cmake version 3.22.1)

hakonhagland:
> if you use the super-build it only downloads once

@akva2 What is a super build? I would like to try this, is documented somewhere?

akva2:
use CMakeLists.txt from https://github.com/OPM/opm-utilities/tree/master/opm-super put that in the in the same folder as the opm modules, i.e.
```
ls /home/akva/kode/opm
CMakeLists.txt
opm-common
opm-tests
opm-grid
opm-models
opm-simulators
opm-upscaling
```

point cmake to that directory and voila, you have one buildsystem for all modules.

PR #4850, opened 2023-09-07
[testing] implicit ipr-calculations for standard wells

steink:
- this PR is just for testing purposes
- requires opm-common https://github.com/OPM/opm-common/pull/3669
- produces almost exactly the same results as master for the few cases I've checked, but situation may be different for cross-flowing wells
- possible pitfall is that method requires converged well-equations to guarantee good solution
- only standard-well version included here


totto82:
I have tested this on a challenging case based on the Olympus model that motivated #4847. Unfortunately it performed badly and gave up after some time (similar to master) . With #4846 and #4847 the models runs smoothly.

steink:
OK, thanks for testing. Will check the ms-version later, where I think it should provide a robust check for stability of convereged solutions (since there everything is fully implicit).

PR #4852, opened 2023-09-08
Cujac preconditioner

multitalentloes:
Implementation of a jacobi preconditioner using CUDA. Currently unoptimized but valid. Works for blocksizes 2 and 3.


atgeirr:
Thanks for submitting your first PR, welcome to the community!

tests/cuistl/test_cujac.cpp
```diff
@@ -0,0 +1,271 @@
...
+    std::vector<T> expected_vec{7.0,15.0,20.0,8.0};
+    std::vector<T> computed_vec = d_vecVector.asStdVector();
+
+    BOOST_REQUIRE_EQUAL(expected_vec.size(), computed_vec.size());
+    for (size_t i = 0; i < expected_vec.size(); i++){
+        BOOST_CHECK_CLOSE(expected_vec[i], computed_vec[i], 1e-7);
+    }
+}
+
+BOOST_AUTO_TEST_CASE_TEMPLATE(CUJACApplyIsEqualToDuneSeqJacApply, T, NumericTypes)
```
atgeirr, line 220: I think this file should be split in three, keeping this test in the current file but moving the others to `test_vector_operations.cpp` and `test_cusparse_matrix_operations.cpp`. Making the filenames match makes it easier to locate the tests.

opm/simulators/linalg/cuistl/detail/cusparse_matrix_operations.cu
```diff
@@ -0,0 +1,99 @@
...
+                nnzIdx++;
+            }
+
+            // pDiagBlock points to the start of where the diagonal block is stored
+            T* pDiagBlock = (matNonZeroValues+(blocksize*blocksize*nnzIdx));
+            // pVecBlock points to the start of the block element in the vector where the inverse of the diagonal block element should be stored
+            T* pVecBlock = (vec + (blocksize*blocksize*thrIndex));
+
+            // untested - based on Dune implementation
+            if (blocksize == 2){
```
atgeirr, line 46: Would it be better to have the 2x2 and 3x3 cases as separate kernels, and dispatch among them in the invertDiagonalAndFlatten() function instead of the `if` on every thread? Also, then we can better handle the cases that will fail (blocksize == 4). What if blocksize is 1 btw? We should support that.

Similar arguments can be made for the elementWiseMultiplyMVKernel.
Similar arguments can be made about the vector_operation

opm/simulators/linalg/cuistl/PreconditionerAdapter.hpp
```diff
@@ -80,6 +80,10 @@ class PreconditionerAdapter
             m_inputBuffer.reset(new CuVector<field_type>(v.dim()));
             m_outputBuffer.reset(new CuVector<field_type>(v.dim()));
         }
+        //TODO: to implement CuJac I added the moving of data from host to the output buffer
```
atgeirr, line 83: I do not understand why this was necessary?

opm/simulators/linalg/cuistl/CuJac.hpp
```diff
@@ -0,0 +1,118 @@
...
+    //! \returns false
+    static constexpr bool shouldCallPost()
+    {
+        return false;
+    }
+
+
+private:
+    //! \brief Reference to the underlying matrix
+    const M& m_underlyingMatrix;
```
atgeirr, line 102: Prefer the trailing underscore convention to the `m_` convention. I.e. `underlyingMatrix_`.

opm/simulators/linalg/cuistl/CuJac.cpp
```diff
@@ -0,0 +1,182 @@
...
+#include <opm/simulators/linalg/cuistl/detail/cublas_safe_call.hpp>
+#include <opm/simulators/linalg/cuistl/detail/cublas_wrapper.hpp>
+#include <opm/simulators/linalg/cuistl/detail/CuBlasHandle.hpp>
+#include <opm/simulators/linalg/cuistl/detail/fix_zero_diagonal.hpp>
+#include <opm/simulators/linalg/cuistl/detail/safe_conversion.hpp>
+#include <opm/simulators/linalg/cuistl/detail/vector_operations.hpp>
+#include <opm/simulators/linalg/cuistl/PreconditionerAdapter.hpp>
+#include <opm/simulators/linalg/matrixblock.hh>
+
+// This file is based on the guide at https://docs.nvidia.com/cuda/cusparse/index.html#csrilu02_solve ,
```
atgeirr, line 43: This is not correct, is it?

opm/simulators/linalg/cuistl/CuJac.cpp
```diff
@@ -0,0 +1,182 @@
...
+CuJac<M, X, Y, l>::category() const
+{
+    return Dune::SolverCategory::sequential;
+}
+
+template <class M, class X, class Y, int l>
+void
+CuJac<M, X, Y, l>::update()
+{
+    //TODO: this update of m is done twice when it is initialized
```
atgeirr, line 153: Suggest that you instead of creating `m` in the initializer list of the constructor build it here in update(), with an `if` to distinguish between the "need to create from scratch" and "only need to update nonzero values" cases.

opm/simulators/linalg/cuistl/CuJac.hpp
```diff
@@ -0,0 +1,118 @@
...
+    //! \brief The relaxation factor to use.
+    field_type m_w;
+
+    CuSparseMatrix<field_type> m;
```
atgeirr, line 106: A bit too short name, maybe simply "matrix"?

opm/simulators/linalg/cuistl/CuJac.cpp
```diff
@@ -0,0 +1,182 @@
...
+    m.updateNonzeroValues(detail::makeMatrixWithNonzeroDiagonal(m_underlyingMatrix));
+    // Compute the inverted diagonal of A and store it in a vector format in m_diagInvFlattened
+    detail::invertDiagonalAndFlatten(m.getNonZeroValues().data(), m.getRowIndices().data(), m.getColumnIndices().data(), detail::to_int(m.N()), detail::to_int(m.blockSize()), m_diagInvFlattened.data());
```
atgeirr, line 156: Lines should not get that long, split it. Using clang-format helps.

atgeirr (commented):
Some questions, but looks close to being mergable.

You should squish all the commits into one (or a few).

opm/simulators/linalg/cuistl/CuJac.cpp
multitalentloes: In this case the update is 2 lines and I only need one of them to be added to the constructor, so I will instead just omit the update() in the constructor and simply flatten the diagonal immediately. I think that will be clean.

opm/simulators/linalg/cuistl/PreconditionerAdapter.hpp
multitalentloes: I think the result depends on both v and d? When preconditioning 0 = b - Ax I need both the b and the x vector to correctly update the x.

tests/cuistl/test_cujac.cpp
multitalentloes: agree

opm/simulators/linalg/cuistl/detail/cusparse_matrix_operations.cu
multitalentloes: Not sure if the difference in practice really matters, but in case the compiler can do some neat trick with knowledge of the blocksize I have split it up now

multitalentloes: When writing the tests for the case handling blocksize=1 I realized why I had not implemented it already. CuSparse does not always support it, for instance not in the case of bsrmv, which is a requirement for jacobi preconditioning. We can of course implement our own sparse multiplication for this case if the simulator needs to support it.

CMakeLists_files.cmake
```diff
@@ -158,15 +158,18 @@ if(CUDA_FOUND)
...
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/cuistl/CuVector.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/cuistl/detail/vector_operations.cu)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/cuistl/CuSparseMatrix.cpp)
+  list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/cuistl/CuJac.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/cuistl/CuSeqILU0.cpp)
   list (APPEND MAIN_SOURCE_FILES opm/simulators/linalg/cuistl/set_device.cpp)
 
   # CUISTL HEADERS
   list (APPEND PUBLIC_HEADER_FILES opm/simulators/linalg/cuistl/detail/cuda_safe_call.hpp)
+  list (APPEND PUBLIC_HEADER_FILES opm/simulators/linalg/cuistl/detail/cusparse_matrix_operations.cu)
```
atgeirr, line 172: Should this be the .hpp file, not .cu?

Also, the CuJac.hpp header is not added anywhere.

atgeirr (changes requested):
Other than some missing headers in cmake I think this is fine.

opm/simulators/linalg/cuistl/PreconditionerAdapter.hpp
multitalentloes: This has been sorted out now. It was not necessary and the apply could also be dramatically simplified for the same reason. The final version will be ready very shortly.

multitalentloes:
Tests are passing and the implementation is quite optimized by fusing kernels. Norne on cpu using bicgstab with jacobi preconditioner takes 1985sec on my machine, whereas the cujac takes 519sec. 
I think all the feedback has been resolved, and the dubious extra cuda memory copy introduced in earlier commits has been successfully removed.

multitalentloes:
I think the issues addressed to far have been resolved. The implementation runs on the norne case in 518 sec when used with cubicgstab, whereas the dune cpu jacobi with bicgstab does it in 1985 sec. The tests pass and the implementation supports block sizes up to 3.

opm/simulators/linalg/cuistl/detail/cusparse_matrix_operations.cu
```diff
@@ -0,0 +1,155 @@
...
+    }
+} // namespace
+
+template <class T>
+void
+invertDiagonalAndFlatten(T* d_mat, int rowIndices[], int colIndices[], size_t numberOfRows, size_t blocksize, T* d_vec)
+{
+    switch (blocksize) {
+    case 1:
+        cuInvertDiagonalAndFlattenBlocksize1<<<getBlocks(numberOfRows), getThreads(numberOfRows)>>>(d_mat, rowIndices, colIndices, numberOfRows, d_vec);
```
kjetilly, line 137: All of these kernels could probably be replaced by a single templated kernel (templated on `int blocksize`), and then this kernel could use a templated function for inverting the matrices. However, I would just leave this as a `// TODO` comment for now, as the current implementation is fine for just three versions (but if we ever add a fourth...)

kjetilly (commented):
Overall this is well written, but please check if the `makeMatrixWithNonzeroDiagonal`-call is needed in practice in this case, I doubt it is, since it was only introduced to make ILU0 from cusparse work. In practice, the block diagonals should all be invertible, which should be enough for CuJac.

There are some other comments in the code, most of them are purely stylistic.

I have made some stylistic comments in the code, but they can mostly be summarized to:

1) Use `camelCase` and not `snake_case` (underscores) when naming variables. Types should have the first letter uppercased (`PascalCase`). 

2) Avoid Hungarian notation (eg. `pMatrix`), [as this goes against the core guidelines.](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-name-type)

3) The {`.cpp', `.hpp', '.cu'} files should probably be reformatted with `clang-format` to keep it consistent.

4) To keep it consistent with the rest of OPM and especially the rest of cuistl, the member variables of a class should probably be prefixed with `m_`

5) I'd use the `@param[out]` feature of doxygen to signify when a parameter is used as an out parameter of a function.

opm/simulators/linalg/cuistl/CuJac.hpp
kjetilly: `m_` is already the established convention within cuistl, so to keep it consistent, it's probably better to stick with `m_`

opm/simulators/linalg/cuistl/CuJac.hpp
```diff
@@ -0,0 +1,109 @@
...
+    //! \returns false
+    static constexpr bool shouldCallPost()
+    {
+        return false;
+    }
+
+
+private:
+    //! \brief Reference to the underlying matrix
+    const M& cpuMatrix;
```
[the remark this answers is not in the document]
atgeirr: I am more in favor of a single trailing underscore rather than prefixing with m_, as that is used in most of OPM. I guess we did not have time to change this when originally merging CuIstl, but it would be a minor change I hope.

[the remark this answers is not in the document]
kjetilly: But why? With `m_` it's really easy to auto-complete within an IDE. In "Clean Code: A Handbook of Agile Software Craftsmanship" it's recommended against any distinguishing of class members (also trailing underscores) on the basis that the class and function should be so small that it's not needed, but if one has to do the distinction, shouldn't it at least benefit IDE users?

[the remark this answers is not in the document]
atgeirr: I guess the IDE argument makes sense, and may back down on this if pressured... The assumption of small classes and functions is not really one that holds though.

opm/simulators/linalg/cuistl/detail/vector_operations.cu
```diff
@@ -45,6 +47,60 @@ namespace
...
+    {
+        const auto globalIndex = blockDim.x * blockIdx.x + threadIdx.x;
+        const int blocksize = 2;
+
+        if (globalIndex < numberOfElements) {
+            T* pMat = (squareBlockVector + (blocksize * blocksize * globalIndex));
+            const T* localSrcVec = (globalSrcVec + (blocksize * globalIndex));
+            T* localDstVec = (globalDstVec + (blocksize * globalIndex));
+
+            T v0 = localSrcVec[0], v1 = localSrcVec[1];
```
kjetilly, line 79: I'd break this out into two lines, see https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-stmt and https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl21-declare-one-name-only-per-declaration 

The same applies for line 97 n the blocksize=3-version.

multitalentloes:
The stylistic feedback, and the feedback regarding template arguments to combine cuda kernels have been accounted for in the latest commit.

atgeirr (approved):

atgeirr:
jenkins build this please

akva2:
jenkins build this please

atgeirr:
All green, merging!

Thanks for the code and the review.

PR #4853, opened 2023-09-08
Get cell volumes from Python

hakonhagland:
Adds a new method `get_cell_volumes()` to the `opm.simulators.BlackOilSimulator` Python module. This method returns a Python list of the cell volumes as represented in the black oil simulator. Example of usage is given in the test case. This method can be useful if the python script will do reaction calculation based on cell volumes.


python/simulators/simulators.cpp
```diff
@@ -1,125 +0,0 @@
-/*
```
hakonhagland, line 1: I thinks this file and the corresponding `simulators.hpp` file are not used anymore. Probably I forgot to delete them in a previous PR.

totto82:
jenkins build this please

totto82:
Thanks.

PR #4854, opened 2023-09-08
Set GECON closure reason to GROUP, to not open with WTEST option 'E'.

vkip:


vkip:
jenkins build this please

bska:
jenkins build this please

totto82:
Thanks for the fix. We should add a test case for where the G option in WTEST is activated. I am not sure if we support that correctly now.

PR #4855, opened 2023-09-08
Fixed indexing error in PID controller

erikhide:
The PID controller formula (see articles referred to in the comments in the OPM code, e.g. "D. Kuzmin and S. Turek, Numerical simulation of turbulent bubbly flows (2004)") as implemented in the current version of OPM contains an indexing error.

Comparing the implementation and PID controller formula, errors_[0] correspond to e_{n-2}, errors_[1] to e_{n-1}, and errors_[2] to e_n.

### DIFF
--- a/opm/simulators/timestepping/TimeStepControl.cpp
+++ b/opm/simulators/timestepping/TimeStepControl.cpp
@@ -206,7 +206,7 @@ namespace Opm
             const double kD = 0.01 ;
             const double newDt = (dt * std::pow( errors_[ 1 ] / errors_[ 2 ], kP ) *
                                  std::pow( tol_         / errors_[ 2 ], kI ) *
-                                 std::pow( errors_[0]*errors_[0]/errors_[ 1 ]/errors_[ 2 ], kD ));
+                                 std::pow( errors_[1]*errors_[1]/errors_[ 0 ]/errors_[ 2 ], kD ));
             if( verbose_ )
                 OpmLog::info(fmt::format("Computed step size (pow): {} days", unit::convert::to( newDt, unit::day )));
             return newDt;


erikhide:
After looking at it quite a few times, I'm more or less certain that the indexing was wrong in the original implementation. And yes, the PID controller in OPM has remained unchanged since at least 2015. The time stepping algorithms seems to have been adapted to the current PID controller implementation, so there might be a need to update other parts and/or default parameter values to make it work with the new implementation.

That said, there is nothing inherently wrong with using the existing version, but as it stands now it is not a correct implementation of the PID controller as derived in the literature.

totto82:
Note that the PID controller has in my opinion never worked as intended. It was therefore removed as the default time-stepper long time ago. Thanks for looking into this. If you can make it to work,  it would be great.

alfbr:
benchmark please

atgeirr:
> Note that the PID controller has in my opinion never worked as intended. It was therefore removed as the default time-stepper long time ago. Thanks for looking into this. If you can make it to work, it would be great.

Our current default is "pid+newtoniteration" which still includes the erroneous pid implementation. I am very much in favour of merging this to correct it, but we must deal with the fallout including failed tests etc. and potentially also change other things as suggested by @erikhide:

> so there might be a need to update other parts and/or default parameter values to make it work with the new implementation.

Just a hunch: I cannot remember to have ever seen the verbose messages from the PID function. It might of course be activated in some places, but is it possible that it never was, and only after the current change is it relevant? If so, we could try to make the default TimeStepControl just "newtoniteration" and see if that is easier to deal with.

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: drogon - Threads: 1                | 0.998    |
| opm-git | OPM Benchmark: drogon - Threads: 8                | 1.169    |
| opm-git | OPM Benchmark: punqs3 - Threads: 1                | 0.979    |
| opm-git | OPM Benchmark: punqs3 - Threads: 8                | 1.026    |
| opm-git | OPM Benchmark: smeaheia - Threads: 1              | 1.009    |
| opm-git | OPM Benchmark: smeaheia - Threads: 8              | 1.123    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 1         | 1.059    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 8         | 1        |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1.019    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 1.062    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 1.026    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 1.012    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 1.021    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 1.007    |



View result details @ https://www.ytelses.com/opm/?page=result&id=2228

totto82:
The benchmark results looks very promising. 

> Our current default is "pid+newtoniteration" which still includes the erroneous pid implementation

My memory is getting worse. I remember looking into this 5 years ago and concluded that PID didn't work as expected but I never noticed the bug. I agree with @atgeirr that this should be merged and with @erikhide that we need to test this and possibly also tune the timestepping differently. Have you done any testing on only using "pid". The "newton" part have some issues at the moment. For some cases it can get stuck on very small timesteps due to some unlucky combinations of NUPCOL and group/well switching.

erikhide:
I have tried to use only "pid" as input to `--time-step-control`, but so far I cannot say anything conclusive about what works best. My understanding is that the "target iteration approach" was added to ensure that the time steps remained low, because the PID controller tended to favour larger time steps than what might be preferable. The time stepping algorithm as a whole (as currently implemented) finds several different time step proposals, and always chooses the smallest one. Hence, it might become rather conservative in some cases and possibly also fail to increase the time steps when it is both possible and desirable.

I'm also working on some other alternatives (other heuristics than the PID controller, as well as other versions of the PID controller itself), and I'm trying to decide how best to compare different time stepping algorithms to ensure both fast computations and robust behaviour. This might take some time, but hopefully possible improvements will come sometime in the near future.

bska:
> I'm also working on some other alternatives (other heuristics than the PID controller, as well as other versions of the PID controller itself), and I'm trying to decide how best to compare different time stepping algorithms to ensure both fast computations and robust behaviour. This might take some time, but hopefully possible improvements will come sometime in the near future.

This is very exciting news! Thank you so much for attacking this area :+1:

opm/simulators/timestepping/TimeStepControl.cpp
```diff
@@ -206,7 +206,7 @@ namespace Opm
             const double kD = 0.01 ;
             const double newDt = (dt * std::pow( errors_[ 1 ] / errors_[ 2 ], kP ) *
                                  std::pow( tol_         / errors_[ 2 ], kI ) *
-                                 std::pow( errors_[0]*errors_[0]/errors_[ 1 ]/errors_[ 2 ], kD ));
+                                 std::pow( errors_[1]*errors_[1]/errors_[ 0 ]/errors_[ 2 ], kD ));
```
bska, line 209: I've looked a little bit closer at this.  Suppose `errors_[0] == 0`, which it theoretically could be since the conditional above only checks [`errors_[1]` and `errors_[2]`](https://github.com/OPM/opm-simulators/blob/c1604161fe3a845280121c7e19235247ceaa05b1/opm/simulators/timestepping/TimeStepControl.cpp#L195).  Then `newDt` will be non-finite, typically `inf`.  Is that reasonable here?

erikhide: I guess the errors_[1] and errors[2] were included in the conditional because they appeared in the denominators of the PID formula, and hence it was only relevant to check those. However, having changed indexing, errors_[0] (and not errors_[1]) appears in the denominator. It might therefore be reasonable to add errors[0] to (and possibly just remove errors_[1] from) the conditional.

bska: > It might therefore be reasonable to add `errors[0]` to [...] the conditional.

Yes, that's probably needed.

> and possibly just remove `errors_[1]` from the conditional

I'd be wary of doing that.  If `errors_[1] = 0` we'll get `newDt = 0` as well and that seems unexpected to me.  Maybe I'm just misunderstanding something.

erikhide: You're probably right. It appears that that might have been a problem earlier, though. Regardless, it certainly shouldn't hurt to add `errors_[0] == 0` to the conditional.

I added it now.

bska:
jenkins build this please

opm/simulators/timestepping/TimeStepControl.cpp
bska: > It appears that that (`errors_[0] = 0` giving `newDt = 0`) might have been a problem earlier, though

Agreed.  I only remarked on this possibility because we're in the area now so it's an opportunity to fix it.  For what it's worth, I tried this PR on a field case and ran into that exact problem.  Otherwise I might not have noticed it.

bska (commented):
Thank you for the update.  I think this is starting to look good so I'm running a build check to assess the impact of the change.

alfbr:
Any chance we can get this in?

bska:
> Any chance we can get this in?

There's absolutely a chance we can get it in.  We're a little constrained by the way we handle updates to reference solutions in the time leading up to a new official release, so it'll definitely be easier to get it in once the new release is finalised.

bska:
> jenkins build this please

Note to @erikhide : The above build request is just to refresh my memory of the impact of this PR.  We are going to merge this regardless.

GitPaean:
jenkins build this please

GitPaean:
@erikhide, can you please rebase the branch on top of the master branch?

akva2:
are we having the awkward rebase conversation again? :smile:

erikhide:
> @erikhide, can you please rebase the branch on top of the master branch?

Done!

GitPaean:
> are we having the awkward rebase conversation again? 😄

Let me know if we do not need it anymore. Just think "Merge OPM/master to master" does not look great, and the branch name can be something else.

akva2:
erik rebased after my comment :+1:

GitPaean:
I will go through the regression failures and report back. From the discussion, it looks like a bug fix. But if you have more comments, please do so.

GitPaean:
benchmark please

alfbr:
> I will go through the regression failures and report back. From the discussion, it looks like a bug fix. But if you have more comments, please do so.

Please look at https://github.com/OPM/opm-simulators/pull/4856 instead. It includes this bug fix in addition to one more bug fix.

bska:
> > I will go through the regression failures and report back. [...]
> 
> Please look at #4856 instead. It includes this bug fix in addition to one more bug fix.

Do be advised, though, that #4856 caused rather different failures than this.  There were multiple tests that hit the 1500 second (25 minute) timeout mark in the test build for #4856 whereas this "only" caused the "normal" kinds of regression test failures.

GitPaean:
Let us get this in first and checking what happens with https://github.com/OPM/opm-simulators/pull/4856 .

alfbr:
> > > I will go through the regression failures and report back. [...]
> > 
> > 
> > Please look at #4856 instead. It includes this bug fix in addition to one more bug fix.
> 
> Do be advised, though, that #4856 caused rather different failures than this. There were multiple tests that hit the 1500 second (25 minute) timeout mark in the test build for #4856 whereas this "only" caused the "normal" kinds of regression test failures.

Thanks for the information, that is interesting. Yes, @GitPaean , then it may make sense to do this one first.

GitPaean:
https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareECLFiles_flow_MICP/

this case does not produce any summary output  except `YEARS`. Maybe we should add some.

GitPaean:
I went through almost all the cases except the ones below, 
[mpi.compareECLFiles_flow_blackoil_alugrid+SPE1CASE2](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareECLFiles_flow_blackoil_alugrid_SPE1CASE2/) (did not run, jenkins looks fine)
[mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_fluidstate_variables/)
[mpi.python_primary_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_primary_variables/)
[mpi.compareDamarisFiles_flow+SPE1CASE1](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareDamarisFiles_flow_SPE1CASE1/)

We need to update this one to make the regression meaningful. 
https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareECLFiles_flow_MICP/

And for [mpi.compareECLFiles_flow+MOD4_GRP](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareECLFiles_flow_MOD4_GRP/) , the PR introduces much more chopping at the end. so we need to evaluate. This case has long been a volatile one, while we need to evaluate. 

And [mpi.compareECLFiles_flow+UDQ_WCONPROD](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareECLFiles_flow_UDQ_WCONPROD/) needs more investigation.


All the other cases are okay, [mpi.compareECLFiles_flow+PINCH_MULTZ_ALL](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareECLFiles_flow_PINCH_MULTZ_ALL/)  might be the case show the effects of this PR. 

Need to leave office and will post more detailed information for the cases mentioned above.

GitPaean:
For [mpi.compareECLFiles_flow+MOD4_GRP](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareECLFiles_flow_MOD4_GRP/) the well `WIL2` has some oscillation and causes a lot of time step chopping, more than doubled the total time steps

image

image

GitPaean:
for [mpi.compareECLFiles_flow+UDQ_WCONPROD](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareECLFiles_flow_UDQ_WCONPROD/), the well water rates for OPU02 is relatively small, while it also shows big relative difference. I think it should be fine. 

image

GitPaean:
I have not checked the following, 
[mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_fluidstate_variables/)
[mpi.python_primary_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_primary_variables/)
[mpi.compareDamarisFiles_flow+SPE1CASE1](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareDamarisFiles_flow_SPE1CASE1/)

For the ones I checked, we need to discuss a little bit about [MOD4_GRP](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareECLFiles_flow_MOD4_GRP/) to see whether it is okay to just update the reference.

alfbr:
> I have not checked the following, [mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_fluidstate_variables/) [mpi.python_primary_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_primary_variables/) [mpi.compareDamarisFiles_flow+SPE1CASE1](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareDamarisFiles_flow_SPE1CASE1/)
> 
> For the ones I checked, we need to discuss a little bit about [MOD4_GRP](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareECLFiles_flow_MOD4_GRP/) to see whether it is okay to just update the reference.

Great, thanks for the hard work! I am available tomorrow to discuss, and help look at the case.

GitPaean:
Maybe @bska @atgeirr @totto82 can comment on the [MOD4_GRP](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareECLFiles_flow_MOD4_GRP/) failures.

GitPaean:
For [mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_fluidstate_variables/)

```
Traceback (most recent call last):
  File "/var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/python/test/test_fluidstate_variables.py", line 23, in test_all
    self.assertAlmostEqual(oil_pressure[0], 41729978.837, places=2, msg='value of oil pressure')
AssertionError: 41726073.576932766 != 41729978.837 within 2 places (3905.260067231953 difference) : value of oil pressure
```

I think it is too strict to have a tolerance at 0.01 pascal level for such hard-coded reference value, especially if the time stepping changes.

alfbr:
> Maybe @bska @atgeirr @totto82 can comment on the [MOD4_GRP](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareECLFiles_flow_MOD4_GRP/) failures.

Please allow me time to look at the case first. Compiling as we speak :)

alfbr:
> For [mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_fluidstate_variables/)
> 
> ```
> Traceback (most recent call last):
>   File "/var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/python/test/test_fluidstate_variables.py", line 23, in test_all
>     self.assertAlmostEqual(oil_pressure[0], 41729978.837, places=2, msg='value of oil pressure')
> AssertionError: 41726073.576932766 != 41729978.837 within 2 places (3905.260067231953 difference) : value of oil pressure
> ```
> 
> I think it is too strict to have a tolerance at 0.01 pascal level for such hard-coded reference value, especially if the time stepping changes.
> 
> It is same for [mpi.python_primary_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_primary_variables/) .

I am not familiar with those tests, maybe somebody who has worked with them can comment?

alfbr:
> Maybe @bska @atgeirr @totto82 can comment on the [MOD4_GRP](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareECLFiles_flow_MOD4_GRP/) failures.

I have now looked at the instability seen in the vector WWIR:WIL2. This looks like a regression, where the group controls are oscillating causing excessive time step chopping. This behavior was introduced between December 15th 2021 and June 9th 2022. Unfortunately a time when we had five repos to juggle, making bisection demanding.

GitPaean:
> [mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_fluidstate_variables/)
> [mpi.python_primary_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_primary_variables/)
> [mpi.compareDamarisFiles_flow+SPE1CASE1](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareDamarisFiles_flow_SPE1CASE1/)

Sure. Then we can discuss the other three tests. 

[mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_fluidstate_variables/)
[mpi.python_primary_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_primary_variables/)
[mpi.compareDamarisFiles_flow+SPE1CASE1](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareDamarisFiles_flow_SPE1CASE1/)



For [mpi.compareDamarisFiles_flow+SPE1CASE1](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareDamarisFiles_flow_SPE1CASE1/), it is okay to upate_data. 

For 
> [mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_fluidstate_variables/)
> [mpi.python_primary_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_primary_variables/)

We need to discuss how to change the tests.  With different time stepping, there is some different pressures near the injecting wells, the difference is too big with the current way for regression test, 
```c++
self.assertAlmostEqual(oil_pressure[0], 41729978.837, places=2, msg='value of oil pressure')
```

`places=2`, if I understand correctly, basically it is almost equivalent to a comparison tolerance around 0.01 level.

alfbr:
> Sure. Then we can discuss the other three tests.
> 
> [mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_fluidstate_variables/) [mpi.python_primary_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_primary_variables/) [mpi.compareDamarisFiles_flow+SPE1CASE1](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareDamarisFiles_flow_SPE1CASE1/)
> 
> For [mpi.compareDamarisFiles_flow+SPE1CASE1](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/compareDamarisFiles_flow_SPE1CASE1/), it is okay to upate_data.
> 
> For
> 
> > [mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_fluidstate_variables/)
> > [mpi.python_primary_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_primary_variables/)
> 
> We need to discuss how to change the tests. With different time stepping, there is some different pressures near the injecting wells, the difference is too big with the current way for regression test,
> 
> ```c++
> self.assertAlmostEqual(oil_pressure[0], 41729978.837, places=2, msg='value of oil pressure')
> ```
> 
> `places=2`, if I understand correctly, basically it is almost equivalent to a comparison with absolute tolerance around 0.01 level .

Indeed, @hakonhagland can you comment on the python tests, e.g., is it fine to relax the tolerances? Generally speaking, those tests if they are hardcoded (i.e., not part of the regression testing framework that have automated routines for updating data), then they really should be resilient to the kind of changes we have here.

GitPaean:
> Indeed, @hakonhagland can you comment on the python tests, e.g., is it fine to relax the tolerances? Generally speaking, those tests if they are hardcoded (i.e., not part of the regression testing framework that have automated routines for updating data), then they really should be resilient to the kind of changes we have here.

I think we can. But since we are updating the tests through coding, maybe we should **also** update the reference with the new values.  The results are checked and okay with this PR.

alfbr:
benchmark please

alfbr:
> maybe we should **also** update the reference with the new values. The results are checked and okay with this PR.

To me that seems like a too big burden for developers. Doing it now is probably fine since you already did the digging work to figure out what to do and where. This is not efficient though, so I would like the test to be changed to avoid this manual updating. Your time is too valuable.

GitPaean:
after #5429, we need to update the following two tests to update the reference.

[mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_fluidstate_variables/)
[mpi.python_primary_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6315/testReport/junit/(root)/mpi/python_primary_variables/)

The hard-code long digit with absolute tolerance at `0.01` is very vulnerable for any time stepping change and even linear/non-linear iteration change.

GitPaean:
jenkins build this please

alfbr:
benchmark please

GitPaean:
I will suggest some way to adjust the following two cases, 

[mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6354/testReport/junit/(root)/mpi/python_fluidstate_variables/)
[mpi.python_primary_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6354/testReport/junit/(root)/mpi/python_primary_variables/)

GitPaean:
One possible way to update and fix the following failures is suggested in #5431 . 

[mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6354/testReport/junit/(root)/mpi/python_fluidstate_variables/)
[mpi.python_primary_variables](https://ci.opm-project.org/job/opm-simulators-PR-builder/6354/testReport/junit/(root)/mpi/python_primary_variables/)

GitPaean:
https://github.com/OPM/opm-simulators/pull/5431 is merged. Hopefully the PR is ready for update_data.

GitPaean:
jenkins build this please

GitPaean:
I did not review the PR itself and only check the regression failures. With the python tests updated, the PR is ready to update_data.  Please feel free to do so if you think the content the PR is ready.

totto82:
jenkins build this please

atgeirr:

GitPaean:
image

There is 690 MB text file here. Is it intentional this way?

akva2:
uhm, yes? that is the textual diff of the changes and the changes here are quite substantial.

atgeirr:
jenkins build this opm-tests=4855 please

GitPaean:
jenkins build this opm-tests=1230 please

GitPaean:
A parallel test failure needs to be fixed, 

https://ci.opm-project.org/job/opm-simulators-PR-builder/6681/testReport/(root)/mpi/compareParallelSim_flow_3_A_MPI_MULTFLT_SCHED_MODEL2/

Basically the time stepping changes, then the comparison does not work.

GitPaean:
Referring to the incident that wrong PR number is used with `opm-tests=4855`, can jenkins stop and point it out that the PR wanted do not exist? 

@akva2

GitPaean:
> A parallel test failure needs to be fixed,
> 
> https://ci.opm-project.org/job/opm-simulators-PR-builder/6681/testReport/(root)/mpi/compareParallelSim_flow_3_A_MPI_MULTFLT_SCHED_MODEL2/
> 
> Basically the time stepping changes, then the comparison does not work.

```Keyword: XGRP, origin Restart, sequence 39

Value index                 = 180
(first value, second value) = (7094.52, 7088.57)

Program threw an exception: [/var/lib/jenkins/workspace/opm-simulators-PR-builder/deps/opm-common/test_util/EclRegressionTest.cpp:213] Deviations exceed tolerances.
The absolute deviation is 5.9550098537583835, and the tolerance limit is 0.02.
The relative deviation is 0.0008393810962458617, and the tolerance limit is 8e-05. ```

I do not consider this parallel running is easy to fix without modifying the DATA file to avoid chopping. 

When different chopping happens  (serial running chopped twice and parallel running chopped once), it is difficult to give a comparison tolerance.

**Comment by GitPaean** [comment]
> > A parallel test failure needs to be fixed,
> > https://ci.opm-project.org/job/opm-simulators-PR-builder/6681/testReport/(root)/mpi/compareParallelSim_flow_3_A_MPI_MULTFLT_SCHED_MODEL2/
> > Basically the time stepping changes, then the comparison does not work.
> 
> ```
> 
> Value index                 = 180
> (first value, second value) = (7094.52, 7088.57)
> 
> Program threw an exception: [/var/lib/jenkins/workspace/opm-simulators-PR-builder/deps/opm-common/test_util/EclRegressionTest.cpp:213] Deviations exceed tolerances.
> The absolute deviation is 5.9550098537583835, and the tolerance limit is 0.02.
> The relative deviation is 0.0008393810962458617, and the tolerance limit is 8e-05. 
> ```
> 
> I do not consider this parallel running is easy to fix without modifying the DATA file to avoid chopping.
> 
> When different chopping happens (serial running chopped twice and parallel running chopped once), it is difficult to give a comparison tolerance.

One a second thinking, we can raise the relative tolerance to be `1.e-4. ` to make it pass. 

I was too quick to jump into the conclusion.

**Comment by GitPaean** [comment]
> > > A parallel test failure needs to be fixed,
> > > https://ci.opm-project.org/job/opm-simulators-PR-builder/6681/testReport/(root)/mpi/compareParallelSim_flow_3_A_MPI_MULTFLT_SCHED_MODEL2/

> 
> One a second thinking, we can raise the relative tolerance to be `1.e-4` to make it pass.


@erikhide , will you be able to do it?

it is the following line, 
https://github.com/OPM/opm-simulators/blob/5137f518484d2ce72300b1b2fddc37bea90e112b/parallelTests.cmake#L204

**Comment by erikhide** [comment]
> > One a second thinking, we can raise the relative tolerance to be `1.e-4` to make it pass.
> @erikhide , will you be able to do it?

I hope I managed to do it successfully now. Let me know if I need to change something.

**Comment by GitPaean** [comment]
> > > One a second thinking, we can raise the relative tolerance to be `1.e-4` to make it pass.
> > > @erikhide , will you be able to do it?
> 
> I hope I managed to do it successfully now. Let me know if I need to change something.

Sorry, it is a little awkward, it should be (1.e-3) since the relative difference is (0.0008393810962458617 ~ 8e-4). 

Please also rebase your PR to the current master.

**Comment by GitPaean** [comment]
I will trigger a jenkins run anyway, still the same failure expected though due to my miscounting of 0s.

**Comment by GitPaean** [comment]
jenkins build this opm-tests=1230 please

**Comment by GitPaean** [comment]
jenkins build this opm-tests=1230 please

**Comment by erikhide** [comment]
> Sorry, @erikhide, it is a little awkward, it should be (1.e-3) since the relative difference is (0.0008393810962458617 ~ 8e-4).
> 
> Please also rebase your PR to the current master.

I think it should be ok now.

File: parallelTests.cmake [hunk 2]
```diff
@@ -201,8 +201,8 @@ add_test_compare_parallel_simulation(CASENAME 3_a_mpi_multflt_mod2
                                      FILENAME 3_A_MPI_MULTFLT_SCHED_MODEL2
                                      SIMULATOR flow
                                      ABS_TOL ${abs_tol_parallel}
-                                     REL_TOL ${rel_tol_parallel}
-                               DIR model2
+                                     REL_TOL 1.0e-3
+                        			       DIR model2
```
GitPaean, line 205: There is some extra white spaces or tabs here. Bu the original alignment was not great either. 

But since you are here, can you fix the alignment for this line? No worry if you do not do it, we can fix it easily.

GitPaean:
Okay. At the end, it is harder to fix without changing the SCHEDULE in the DATA file, 

```
Value index                 = 181
(first value, second value) = (265.875, 269.609)

Program threw an exception: [/var/lib/jenkins/workspace/opm-simulators-PR-builder/deps/opm-common/test_util/EclRegressionTest.cpp:213] Deviations exceed tolerances.
The absolute deviation is 3.7344022249249633, and the tolerance limit is 0.02.
The relative deviation is 0.013851169861009862, and the tolerance limit is 0.001.

..... 

	OPMEXTRA: Restart, sequence 39
		Fails for 1 entries
		Largest absolute error: 1.0368000e+06
		Largest relative error: 1.5384615e-01
	OPM_XWEL: Restart, sequence 39
		Fails for 24 entries
		Largest absolute error: 8.2348517e+04
		Largest relative error: 2.1272244e-02
	PBUB: Restart, sequence 39
		Fails for 380 entries
		Largest absolute error: 4.3840332e+00
		Largest relative error: 1.7462628e-02
	PDEW: Restart, sequence 39
		Fails for 361 entries
		Largest absolute error: 9.7746124e+00
		Largest relative error: 8.9197796e-02
	PRESSURE: Restart, sequence 39
		Fails for 377 entries
		Largest absolute error: 8.8223267e-01
		Largest relative error: 3.5680844e-03
	RS: Restart, sequence 39
		Fails for 438 entries
		Largest absolute error: 2.3180008e+00
		Largest relative error: 2.1001179e-02
	SWAT: Restart, sequence 39
		Fails for 18 entries
		Largest absolute error: 3.3567309e-02
		Largest relative error: 1.1402052e-01
	XCON: Restart, sequence 39
		Fails for 135 entries
		Largest absolute error: 1.3746820e+06
		Largest relative error: 4.2813794e-02
	XGRP: Restart, sequence 39
		Fails for 8 entries
		Largest absolute error: 3.3155427e+02
		Largest relative error: 1.3851170e-02
	XWEL: Restart, sequence 39
		Fails for 32 entries
		Largest absolute error: 6.5161461e+05
		Largest relative error: 2.0969111e-02
```

We do not want to make relative tolerance 0.2 or something.  Since they might not show all the failures, 0.2 might not make the test pass either. 

My suggestion is to update the DATA file to avoid time step cutting, while others might have different suggestions. 
@totto82 @atgeirr @bska

atgeirr:
> My suggestion is to update the DATA file to avoid time step cutting, while others might have different suggestions.

Either change the schedule, or tighten some tolerances to make behaviour more equalized. I will look at this locally and see if I can provoke the same behaviour, then propose a fix.

alfbr:
Which test is this?

bska:
> Which test is this?

The parallel version of [`3_A_MPI_MULTFLT_SCHED_MODEL2`](https://github.com/OPM/opm-simulators/blob/5137f518484d2ce72300b1b2fddc37bea90e112b/parallelTests.cmake#L200) which compares the MPI results (four processes) to that of a sequential run.

atgeirr:
I have looked at this, and there are some timesteps where we are converging in 19 or 20 timesteps. This is close to the limit of 20, and minor differences can push the iterations over. There is also a step where the linear solver failes on Newton iteration 16 in parallel, yet the serial keep going to exceed the max Newtons.

Try to add the following argument: `--newton-max-iterations=30`. On my machine this lets all timesteps succeed, no cuts, and no wasted iterations.

Alternatively, if that fails, we can try instead reducing it: `--newton-max-iterations=14`. That should ensure the runs fail at the same time.

If this also does not work well enough, we could split (in the deck) the timesteps that take many iterations in two. But I prefer to not change the deck if the test can be made robust without it.

erikhide:
I have added `--newton-max-iterations=30` now. I also tried to fix the bad alignment. It looks good in my editor at least.

GitPaean:
jenkins build this opm-tests=1230 please

atgeirr:
> It looks good in my editor at least.

It does not... You should make sure your editor outputs spaces and not tabs.
But no worries, I won't stop for that now...

Will make a final sanity check before merging.

atgeirr:

atgeirr:
Failure report plots indicate no problematic changes (cases not running etc.) and some results look smoother than before. Will do a final jenkins build and merge.

atgeirr:
jenkins build this opm-tests=1230 please

atgeirr:
Two wrong indices leading to 90 comments over just over a year... It really opened a can of worms. But now it is merged!

erikhide:
I didn't expect my first contribution to cause this much trouble. Nice to finally get it merged.

bska:
> But now it is merged!

This **may** have broken the sequential ALUGrid test.  At least we're getting [regression test failures](https://ci.opm-project.org/job/opm-simulators/2374/testReport/junit/(root)/serial/compareECLFiles_flow_blackoil_alugrid_SPE1CASE2/) following this PR being merged.  I don't want to request that this be backed out, but we should try to understand what's happening.

GitPaean:
Oh, I see that the time has two hour difference from the local time.

atgeirr:
I will look at this, but will not be able to fix until tomorrow morning. Hope the serial jenkins is not critical until then.

bska:
> I will look at this, but will not be able to fix until tomorrow morning.

Much appreciated.  There is definitely **something** going now, because I get the same regression test failure when I run my (in-progress) PR #5589 [in sequential mode](https://ci.opm-project.org/job/opm-simulators-PR-builder/6699/).  That PR adds a couple of classes that are not used anywhere.

> Hope the serial jenkins is not critical until then.

Not a problem at all.

atgeirr:
ALUGrid changes grid cell ordering between MPI (space filling curve order) and non-MPI (logical cartesian order) configurations. @ElyesAhmed has found some bugs in our code related to this and fixed them, but they have not yet come to a PR. That will happen soon though, I hope that will fix this problem.

GitPaean:
> ALUGrid changes grid cell ordering between MPI (space filling curve order) and non-MPI (logical cartesian order) configurations. @ElyesAhmed has found some bugs in our code related to this and fixed them, but they have not yet come to a PR. That will happen soon though, I hope that will fix this problem.

but how is it related to this PR?

akva2:
we get different grids -> different problem solved -> reference data from mpi simulator not matching serial simulator -> jenkins serial build is currently broken.

bska:
> > ALUGrid changes grid cell ordering between MPI (space filling curve order) and non-MPI (logical cartesian order) configurations.
> 
> but how is it related to this PR?

It's probably more fair to say that this PR ended up triggering just the right set of conditions to bring the hidden ALUGrid issue to light.  This PR did not itself cause the problem.

PR #4859, opened 2023-09-10
Damaris - Moving Damaris functionality to it's own class

jcbowden:
Step one for moving Damaris calls out of EclWriter class and into its own DamarisWriter class; 
  
EclProblem now calls the writeOutput methods of DamarisWriter  and EclWriter and passes in the data::Solution object that is shared between them; 
  
Currently has an issue in the first iteration (iteration 0 in Damaris) there is no PRESSURE data - perhaps DamarisWriter does not initialize data::Solution correctly?
  
Moved setupDamarisWritingPars into the DamarisWriter class from DaamrisOutputModule (which now only has initialization code)

### DIFF
--- a/ebos/damariswriter.hh
+++ b/ebos/damariswriter.hh
@@ -1,0 +1,377 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  Copyright 2022 SINTEF Digital, Mathematics and Cybernetics.
+  Copyright 2023 Inria, Bretagne–Atlantique Research Center
+  
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+
+  Consult the COPYING file in the top-level source directory of this
+  module for the precise wording of the license and the list of
+  copyright holders.
+*/
+/*!
+ * \file
+ *
+ * \copydoc Opm::DamarisWriter
+ */
+#ifndef EWOMS_DAMARIS_WRITER_HH
+#define EWOMS_DAMARIS_WRITER_HH
+
+#include <dune/grid/common/partitionset.hh>
+
+#include <ebos/collecttoiorank.hh>
+#include <ebos/eclbasevanguard.hh>
+#include <ebos/eclgenericwriter.hh>
+#include <ebos/ecloutputblackoilmodule.hh>
+
+#include <opm/input/eclipse/Units/UnitSystem.hpp>
+
+#include <opm/output/eclipse/RestartValue.hpp>
+
+#include <opm/simulators/utils/DeferredLoggingErrorHelpers.hpp>
+#include <opm/simulators/utils/ParallelRestart.hpp>
+
+#include <opm/common/OpmLog/OpmLog.hpp>
+
+#include <limits>
+#include <stdexcept>
+#include <string>
+
+#include <fmt/format.h>
+
+#include <Damaris.h>
+
+namespace Opm::Properties {
+
+template<class TypeTag, class MyTypeTag>
+struct EnableDamarisOutput {
+    using type = UndefinedProperty;
+};
+template<class TypeTag, class MyTypeTag>
+struct EnableDamarisOutputCollective {
+    using type = UndefinedProperty;
+};
+} // namespace Opm::Properties
+
+namespace Opm {
+
+/*!
+ * \ingroup EclBlackOilSimulator
+ *
+ * \brief Collects necessary output values and pass them to Damaris server processes.
+ *
+ * Currently only passing through PRESSURE, GLOBAL_CELL_INDEX and MPI_RANK information.
+ * This clss will be enhanced to pass through the 3D mesh information to Damaris to enable
+ * in situ visualization via Paraview or Ascent. And developed so that variables specified 
+ * through the Eclipse input deck will be available to Damaris.
+ */
+ 
+ 
+template <class TypeTag>
+class DamarisWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::Grid>,
+                                          GetPropType<TypeTag, Properties::EquilGrid>,
+                                          GetPropType<TypeTag, Properties::GridView>,
+                                          GetPropType<TypeTag, Properties::ElementMapper>,
+                                          GetPropType<TypeTag, Properties::Scalar>>
+{
+    using Simulator = GetPropType<TypeTag, Properties::Simulator>;
+    using GridView = GetPropType<TypeTag, Properties::GridView>;
+    using Grid = GetPropType<TypeTag, Properties::Grid>;
+    using EquilGrid = GetPropType<TypeTag, Properties::EquilGrid>;
+    using Scalar = GetPropType<TypeTag, Properties::Scalar>;
+    using ElementContext = GetPropType<TypeTag, Properties::ElementContext>;
+    using Element = typename GridView::template Codim<0>::Entity;
+    using ElementMapper = GetPropType<TypeTag, Properties::ElementMapper>;
+    
+    using BaseType = EclGenericWriter<Grid,EquilGrid,GridView,ElementMapper,Scalar>;
+    
+public:
+    static void registerParameters()
+    {
+        EWOMS_REGISTER_PARAM(TypeTag, bool, EnableDamarisOutputCollective,
+                             "Write output via Damaris using parallel HDF5 to get single file per timestep instead of one per Damaris core.");
+    }
+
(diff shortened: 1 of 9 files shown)


jcbowden:
This code change will behave better using the latest Damaris branch rc-v1.9.1 (to be pushed to master shortly and as v1.9.1) due to better testing for ill formed XML configuration (like empty file paths in Python and Paraview XML elements)

jcbowden:
The fix of the last commit removes the need to upgrade to Damaris v.1.9.1.

ebos/eclwriter.hh
```diff
@@ -373,9 +337,12 @@ public:
         const bool isFloresn = this->eclOutputModule_->hasFloresn();
         auto floresn = this->eclOutputModule_->getFloresn();
 
-        data::Solution localCellData = {};
+        // data::Solution localCellData = {};
         if (! isSubStep) {
-            this->eclOutputModule_->assignToSolution(localCellData);
+            
+            if (localCellData.size() == 0) {
```
atgeirr, line 343: Use `localCellData.empty()` instead.

ebos/eclwriter.hh
```diff
@@ -420,7 +387,10 @@ public:
                                 isFlowsn, std::move(flowsn),
                                 isFloresn, std::move(floresn));
         }
+
```
atgeirr, line 390: Better to avoid whitespace changes in places not related to the current changes. Instead make such changes in a consistent manner in a separate commit and PR, if they are necessary. Also lines 392-393.

ebos/eclwriter.hh
```diff
@@ -665,12 +630,8 @@ private:
     }
 
     Simulator& simulator_;
-    std::unique_ptr<EclOutputBlackOilModule<TypeTag>> eclOutputModule_;
+    std::unique_ptr<EclOutputBlackOilModule<TypeTag> > eclOutputModule_;
```
atgeirr, line 633: This (adding a space between the `>>`) used to be necessary in ancient C++, but in modern C++ it is not and should be avoided.

ebos/damariswriter.hh
```diff
@@ -0,0 +1,384 @@
...
+                this->setupDamarisWritingPars(simulator_.vanguard().grid().comm(), numElements_, elements_rank_offsets_);
+                
+                // sets data for non-time-varying variables MPI_RANK and GLOBAL_CELL_INDEX
+                this->SetGlobalIndexForDamaris() ; 
+                
+                // Currently by default we assume static grid (unchanging through the simulation)
+                // Set damarisUpdate_ to true if we want to update the geometry to sent to Damaris 
+                this->damarisUpdate_ = false; 
+            }
+
```
atgeirr, line 182: Max single blank line inside function.

ebos/damariswriter.hh
```diff
@@ -0,0 +1,384 @@
...
+    std::unique_ptr<EclOutputBlackOilModule<TypeTag>> damarisOutputModule_;
+    std::vector<unsigned long long> elements_rank_offsets_ ;
+    bool damarisUpdate_ = false;  ///< Whenever this is true writeOutput() will set up Damaris mesh information and offsets of model fields
+
+    static bool enableDamarisOutput_()
+    { 
+        return EWOMS_GET_PARAM(TypeTag, bool, EnableDamarisOutput); 
+    }
+
+    void SetGlobalIndexForDamaris ()
```
atgeirr, line 225: In OPM, `lowerCamelCase()` should be used for functions, i.e. `SetGlobal...` -> `setGlobal...` .

ebos/damariswriter.hh
```diff
@@ -0,0 +1,384 @@
...
+        rank_ = simulator_.vanguard().grid().comm().rank() ;
+        nranks_ = simulator_.vanguard().grid().comm().size();
+        
+        const auto& gridView = simulator_.gridView();
+        const auto& interior_elements = elements(gridView, Dune::Partitions::interior);
+        // Get the size of the unique vector elements (excludes the shared 'ghost' elements)
+        numElements_ = std::distance(interior_elements.begin(), interior_elements.end());
+        
+        this->elements_rank_offsets_.resize(nranks_) ;
+        this->damarisOutputModule_ = std::make_unique<EclOutputBlackOilModule<TypeTag>>(simulator, this->collectToIORank_);
```
atgeirr, line 135: This means that this writer owns a complete set of the output data stored in the EclOutputBlackOilModule (or precisely: in its base class EclGenericOutputBlackoilModule). So that will be duplicated, since also the EclWriter owns one of these.

We should try to avoid this extra copy and the processing required to build it. Maybe the EclProblem class should own it instead, and pass it to the EclWriter and DamarisWriter? Also, the prepareLocalCellData() method is duplicated between the writers, maybe that could be moved to EclProblem instead of duplicating it? (If so, renaming it prepareLocalCellDataForOutput() would make its purpose clear in the new context.)

@bska I would value your input on this. I want to have the separate DamarisWriter, but not to duplicate storage, processing and code more than necessary.

atgeirr (changes requested):
I think the main point here (separating Damaris-related output in its own class,
 and making it capable of outputting all fields) is very good, and this should be merged eventually.

However, I have some concerns about duplication of storage etc., explained in one of the comments below. The other issues are mostly trivial: formatting etc.

ebos/damariswriter.hh
jcbowden: Hi, yes, I see that issue.  I thought that this was looked after by passing *data::Solution& localCellData* as a reference from EclProblem, first to DamarisWriter::writeOutput() and then to EclWriter::writeOutput(). However, I think that the allocBuffers() call is where the data is created/copied and it is done by both damarisOutputModule_ and eclOutputModule_.

I'll try again passing through a EclOutputBlackOilModule object from  EclProblem, like you suggest.

atgeirr: @bska and I discussed this, and it is clear that we should attempt some refactoring to reduce the complexity of the data flow in this part of the simulator. However, doing so without incurring any extra copying is a bit tricky.

I believe that for now maybe you should not refactor as I suggested above, but instead keep more or less the current design. (We refactor after.) One change we request is that then the call EclWriter::writeOutput() should take a rvalue reference (`&&`) instead of an lvalue reference (`&`) since it eats the data passed into it. Also you must then pass it with std::move(). This makes explicit in the code the comment you already added to EclProblem (keep the comment as well) at the call site.

jcbowden: Okay, I can leave it for now. I'm thinking that the DamarisWriter and EclWriter will have to be consolidated into a single class. Not so much of a problem, if it works.

ebos/damariswriter.hh
```diff
@@ -167,46 +167,42 @@ public:
             // On first call and if the mesh and variable size change then set damarisUpdate_ to true
             if (damarisUpdate_ == true) {
                 // Sets the damaris parameter values "n_elements_local" and "n_elements_total" 
-                // which define sizes of the Damaris variables, per-rank and globally (over all ranks).
+                // which define sizes of the Damaris variables, per-rank_ and globally (over all ranks).
```
atgeirr, line 170: Seem to be a few extra searched-and-replaced instances here? Also line 189.

ebos/eclwriter.hh
```diff
@@ -556,7 +520,7 @@ private:
         const auto& gridView = simulator_.vanguard().gridView();
         const int numElements = gridView.size(/*codim=*/0);
         const bool log = this->collectToIORank_.isIORank();
-
+
```
atgeirr, line 523: More stray whitespace...

atgeirr (approved):

atgeirr:

jcbowden:
This fails in the serial compilation as Damaris.h cannot be found.   
As a fix for this, I have tried:  
Guards in eclproblem.hh
#if HAVE_DAMARIS
#include <ebos/damariswriter.hh>
#endif
  
Guards for including ebos/damariswriter.hh in CMakeLists_files:
if (Damaris_FOUND AND MPI_FOUND)
  list (APPEND PUBLIC_HEADER_FILES ebos/damariswriter.hh)
endif()

Neither of these solution seem to protect the opm-simulatiors if Damaris.h is not present.

Originally there was a switch to include Damaris or not in CMakeLists.txt :
option(USE_DAMARIS_LIB "Use the Damaris library for asynchronous I/O?" OFF)

The switch seems to do nothing now.
  
Someone who has a better grasp of the OPM CMake system might be needed?

jcbowden:
ah, I think I see why - another guard is needd in ./ebos/eclproblem_properties.hh

akva2:

atgeirr:
All green, merging!

ebos/eclproblem.hh
```diff
@@ -1,6 +1,8 @@
 // -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
 // vi: set et ts=4 sw=4 sts=4:
 /*
+  Copyright 2023 INRIA
```
blattms, line 4: If you add one copyright holder then you should add the others, too. Otherwise this might leave a wrong impression.
Of course this is kind of hard to do, but we should try.
As inspiration: For the Debian packages this  (and other) file's copyright is documented as:
```
Copyright: 2009-2022 OPM development team
Comment: We use the name "OPM development team", to denote that parts
     of individual files might be copyright by
     Dr. Blatt - HPC-Simulation-Software & Services, Equinor ASA,
     Statoil ASA (renamed to Equinor ASA in 2018),
     IWS-LH2 at University of Stuttgart, NORCE,
     IRIS (merged into NORCE in 2018), OPM-OP AS, Poware (Andreas Lauser),
     SINTEF, SIMULA, or TNO.
```

jcbowden: I think that having a separate list of contributors and a generic header seems like a good idea. Then I don't have to remember to add INRIA to the ones I contribute to.

PR #4862, opened 2023-09-11
Avoid fraction relaxation factors outside [0,1] (standard wells)

vkip:
Reset values slightly outside the interval to the interval limits, while keeping the assertion to possibly pick up failure cases.

### DIFF
--- a/opm/simulators/wells/StandardWellPrimaryVariables.cpp
+++ b/opm/simulators/wells/StandardWellPrimaryVariables.cpp
@@ -45,9 +45,6 @@
 namespace {
 
     constexpr double EPSILON = 1.0e-14;
-    double safe_fraction(const double fraction) {
-        return std::min(std::max(fraction, 0.0), 1.0);
-    }
     
 //! \brief Relaxation factor considering only one fraction value.
 template<class Scalar>
@@ -55,7 +52,7 @@ Scalar relaxationFactorFraction(const Scalar old_value,
                                 const Scalar dx)
 {
     assert(old_value >= -EPSILON && old_value <= (1.0+EPSILON));
-    const Scalar safe_old_value = safe_fraction(old_value);
+    const Scalar& safe_old_value = std::clamp(old_value, 0.0, 1.0);
 
     Scalar relaxation_factor = 1.;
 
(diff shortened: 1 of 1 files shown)


opm/simulators/wells/StandardWellPrimaryVariables.cpp
bska, line 1: I haven't read all of this, but I noticed you're using
```C++
std::min(std::max(x, low), high)
```
as part of your solution.  For what it's worth, the standard library has [`std::clamp()`](https://en.cppreference.com/w/cpp/algorithm/clamp) through which the above can be written as
```C++
std::clamp(x, low, high)
```
instead.

vkip: Ah, wasn't aware of that one. Thanks!

vkip:
jenkins build this please

bska (commented):
Please note that PR #4860, which will shortly be merged into the master branch, removes the `relaxationFactorRate()` helper function in favour of an alternative approach to updating the primary variables.  You may want to talk to @GitPaean and @steink about possibly reconciling their insights here, also for the `relaxationFactorFraction()` helper function.

steink:
> You may want to talk to @GitPaean and @steink about possibly reconciling their insights here, also for the relaxationFactorFraction() helper function.

I think this looks fine. The case in PR #4860 is a bit different since there the situation we want to avoid is a non-operable well not reaching exactly zero rate (which is neccessary for it to get stopped and converged). For the relaxation of fraction updates this should not be an issue since they don't need to hit the interval limits exactly (only to residual tolerance).

bska (approved):
> > You may want to talk to @GitPaean and @steink about possibly reconciling their insights here, also for the relaxationFactorFraction() helper function.
> 
> I think this looks fine. The case in PR #4860 is a bit different since there the situation we want to avoid is a non-operable well not reaching exactly zero rate (which is necessary for it to get stopped and converged). For the relaxation of fraction updates this should not be an issue since they don't need to hit the interval limits exactly (only to residual tolerance).

Excellent!  In that case I'll merge this work into the master branch.  Thanks a lot for weighin in.

PR #4866, opened 2023-09-13
Adding .PRT output when GECON is activated

vkip:
Also contains https://github.com/OPM/opm-simulators/pull/4854 (cleaner to get this in first?)

Might consider removing debug print at a later stage.

### DIFF
--- a/opm/simulators/wells/GroupEconomicLimitsChecker.cpp
+++ b/opm/simulators/wells/GroupEconomicLimitsChecker.cpp
@@ -24,7 +24,23 @@
 #include <opm/input/eclipse/Schedule/Well/WellTestConfig.hpp>
 #include <fmt/format.h>
 
+#include <ctime>
+#include <chrono>
+#include <iostream>
+#include <iomanip>
+
 namespace Opm {
+
+std::string simTimeToString(const std::time_t start_time, const double sim_time) {
+    const auto start_timep = std::chrono::system_clock::from_time_t(start_time);
+    const auto sim_duration = std::chrono::duration_cast<std::chrono::system_clock::duration>(
+        std::chrono::duration<double>(sim_time)
+    );
+    const std::time_t cur_time = std::chrono::system_clock::to_time_t(start_timep + sim_duration);
+    std::ostringstream ss;
+    ss << std::put_time(std::localtime(&cur_time), "%d-%b-%Y");
+    return ss.str();
+}
 
 GroupEconomicLimitsChecker::
 GroupEconomicLimitsChecker(
@@ -40,6 +56,8 @@ GroupEconomicLimitsChecker(
   , simulation_time_{simulation_time}
   , report_step_idx_{report_step_idx}
   , deferred_logger_{deferred_logger}
+  , date_string_{simTimeToString(well_model.schedule().getStartTime(),simulation_time)}
+  , unit_system_{well_model.eclipseState().getUnits()}
   , well_state_{well_model.wellState()}
   , well_test_state_{well_test_state}
   , schedule_{well_model.schedule()}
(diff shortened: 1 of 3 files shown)


vkip:
Example .PRT output (on opm-tests/gecon/GECON-01.DATA): 
image

totto82:
jenkins build this please

bska:
PR #4854 has been merged into the master branch.  Would you mind rebasing this work on top of the current master sources, please?  That will simplify the review process.

bska:
jenkins build this please

opm/simulators/wells/GroupEconomicLimitsChecker.cpp
```diff
@@ -277,16 +306,45 @@ displayDebugMessage(const std::string &msg) const
 
 void
 GroupEconomicLimitsChecker::
-closeWellsRecursive(Group group)
+addPrintMessage(const std::string &msg, const double value, const double limit, const UnitSystem::measure measure)
+{
+    const std::string header = fmt::format(
+        "{}At time = {:.2f} {} (date = {}): Group {} will close because: \n", this->message_separator,
```
bska, line 312: Note: Since you're already using `fmt::format()` you don't really need to store the `message_separator`.  Rather, it can be generated on the fly using libfmt's dynamic field witdth support
```C++
const auto empty_string = std::string{};
const auto num_asterisks = ...;
const std::string header = fmt::format("{:*>{}}\nAt time...", empty_string, num_asterisks, ...);
```
The `empty_string` is admittedly a bit of a hack here, but it does achieve the desired effect.

opm/simulators/wells/GroupEconomicLimitsChecker.hpp
```diff
@@ -74,6 +80,11 @@ namespace Opm
             {1, BlackoilPhases::Vapour},
             {2, BlackoilPhases::Aqua}};
         std::map<BlackoilPhases::PhaseIndex, int> phase_idx_reverse_map_;
+        const std::string date_string_;
+        const UnitSystem& unit_system_;
```
bska, line 84: Note that this declaration order&ndash;after phase_idx_reverse_map_&ndash;is at odds with the apparent [initialisation order](https://github.com/OPM/opm-simulators/blob/c3d006e91c14c9878d352982f7d02cf0d4bc406b/opm/simulators/wells/GroupEconomicLimitsChecker.cpp#L60-L61) in `GroupEconomicsLimitsChecker.cpp`&ndash;between `deferred_logger_` and `well_state_`.  That, in turn, generates some additional warnings that I'd prefer we avoid.  I think the simplest approach would be to just move the declarations of `date_string_` and `unit_system_` up to the location implied by the textual initialisation order in the `.cpp` file.

opm/simulators/wells/GroupEconomicLimitsChecker.cpp
```diff
@@ -277,16 +306,45 @@ displayDebugMessage(const std::string &msg) const
...
         const std::string msg = fmt::format("closing wells recursive : group {} ", group.name());
         displayDebugMessage(msg);
     }
     for (const std::string& group_name : group.groups()) {
         auto next_group = this->schedule_.getGroup(group_name, this->report_step_idx_);
-        closeWellsRecursive(next_group);
+        wells_closed = wells_closed | closeWellsRecursive(next_group, level+1);
+    }
+    std::string indent("  ");
+    for (int i=0; i<level; ++i) indent += std::string("  ");
```
bska, line 342: You can use the constructor taking a single character and a count to form this string value in one go:
```C++
const auto indent = std::string(2 * level, ' '); // 2 for number of characters in std::string("  ").
```

bska (commented):
I think this looks really good and I look forward to getting it into the simulator.  I just have a couple of minor remarks about some details here.

opm/simulators/wells/GroupEconomicLimitsChecker.cpp
vkip: Nice. Went for a slightly different solution to avoid a bit of code duplication.

vkip:
jenkins build this please

bska (approved):
Thanks for the updates.  This looks good to me now and the build check is green.  I'll merge into master.

PR #4868, opened 2023-09-13
Match Region Sets by Unique Prefix

bska:
This commit switches the region set tag matching algorithm to using unique prefixes.  This enables the simulator to recognise that the region set name
```
FIPUNI
```
should match up with the user defined region set `FIPUNIT`.  In the current master sources, the above summary vector would produce a diagnostic message saying that the region set `FIPUNI` (without the final `T`) does not exist.

To this end, we instruct the `ParallelEclipseState` to always pass the six character substring beginning with `FIP` for FIP-like region arrays and defer to the rank-0 `EclipseState`/`FieldProps` mechanism to match this prefix with its canonical region set.

### DIFF
--- a/opm/simulators/utils/ParallelEclipseState.cpp
+++ b/opm/simulators/utils/ParallelEclipseState.cpp
@@ -24,6 +24,15 @@
 #include <opm/common/ErrorMacros.hpp>
 
 #include <cstddef>
+#include <regex>
+#include <string>
+
+namespace {
+    bool is_FIP(const std::string& keyword)
+    {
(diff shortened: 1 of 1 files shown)


bska:
I am creating this PR in draft mode because it depends on upstream PR OPM/opm-common#3680 which is currently marked as work-in-progress.  I will keep this PR in a draft state until such time as it is ready for review.

bska:
The upstream PR has been merged into the master branch.  I'm marking this as "ready to review" and I'm running a build check.

bska:
jenkins build this please

PR #4869, opened 2023-09-14
PARAOPTS and SOLVDIRS should only give warning

alfbr:
The keywords do not change the results (unless there is numerical issues), so Flow should not treat them as critical.


alfbr:
jenkins build this please

bska (approved):
Yes, this makes sense to me.  Thanks a lot for the contribution.  I'll merge into master.

atgeirr:
Thanks, indeed these options are irrelevant for OPM Flow. One could argue that a few of the PARAOPTS options could be interpreted, but it is better to ignore the keyword in my opinion.

Merging!

atgeirr:
Already merged I see... At least that makes two of us who agree!

PR #4873, opened 2023-09-15
Allow UDA for WSEGVALV item 4

vkip:
Depends on https://github.com/OPM/opm-common/pull/3684

### DIFF
--- a/opm/simulators/wells/MultisegmentWellSegments.cpp
+++ b/opm/simulators/wells/MultisegmentWellSegments.cpp
@@ -815,8 +815,7 @@ pressureDropValve(const int seg,
     const EvalWell friction_pressure_loss =
         mswellhelpers::frictionPressureLoss(additional_length, diameter, area, roughness, density, mass_rate, visc);
 
-    const double udq_default = 0.0; // TODO: Get this from UDQPARAMS
-    ValveUDAEval uda_eval {summary_state, this->well_.name(), seg, udq_default};
+    const ValveUDAEval uda_eval {summary_state, this->well_.name(), static_cast<size_t>(seg)};
     const double area_con = valve.conCrossArea(uda_eval);
     const double cv = valve.conFlowCoefficient();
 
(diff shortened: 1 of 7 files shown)


vkip:
jenkins build this opm-common=3684 please

vkip:
jenkins build this opm-common=3684 please

opm/simulators/wells/MultisegmentWellSegments.cpp
```diff
@@ -740,7 +740,8 @@ pressureDropValve(const int seg) const
     const EvalWell friction_pressure_loss =
         mswellhelpers::frictionPressureLoss(additional_length, diameter, area, roughness, density, mass_rate, visc);
 
-    const double area_con = valve.conCrossArea();
+    const ValveUDAEval uda_eval {summary_state, this->well_.name(), static_cast<size_t>(seg)};
```
bska, line 743: Looking at the [the parameter usage](https://github.com/OPM/opm-common/blob/1594680f357d3c31376d1588701a826a2d3b4468/src/opm/input/eclipse/Schedule/MSW/Valve.cpp#L42-L43), the third constructor parameter appears to be treated as a zero-based segment number.  However, `seg` is not that in this context.  Rather, `seg` is a number in the [range 0..#segments - 1](https://github.com/OPM/opm-simulators/blob/3f436501d2fe213cb9b6b325c420da6c915e2ff1/opm/simulators/wells/MultisegmentWell_impl.hpp#L1622-L1624), in a possibly reordered sequence of segments.  If you really want the segment number here, you can use
```C++
segment_set[seg].segmentNumber()
```
instead.  That will give you the one-based segment number directly.

bska (commented):
I think this is good.  There is a potential misunderstanding concerning the `seg` parameter here so I'd appreciate it if you'd elaborate a little on what you really mean.

opm/simulators/wells/MultisegmentWellSegments.cpp
vkip: It's the segment number I actually want, thanks!

vkip:
jenkins build this please

bska (approved):

PR #4874, opened 2023-09-18
Refactor Fluid-In-Place Calculations For Maintainability

bska:
This PR splits `updateFluidInPlace_()` into several smaller helper functions, each with a narrow purpose.  They're all just called from the original call site&ndash;the body of `updateFluidInPlace_()`&ndash;but this new version is, in my opinion, easier to reason about and there is less shared state.

In anticipation of adding support for summary vectors `FHPV` and `RHPV` (field and region levels of hydrocarbon pore-volumes), we also split the pore-volume updates out to a branch separate from that needed for average pressure calculations.

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -51,14 +51,17 @@
 
 #include <algorithm>
 #include <array>
+#include <cassert>
 #include <cstddef>
 #include <initializer_list>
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska:
jenkins build this please

totto82:
Thanks for the refactoring. Looks good.

PR #4876, opened 2023-09-18
Calculate Hydrocarbon Pore-Volumes if Requested

bska:
This commit separates the computation of hydrocarbon pore-volumes out from the context of average pressure values.  These pore-volumes go into the `xHPV` summary vectors and are, therefore, useful in their own right&ndash;not just as a means to computing average pressure values.


bska:
I'm creating this PR in draft mode because it depends on, and contains, PR #4874.  I will keep the PR in draft state until such time as it is ready for review and merging.

bska:
The upstream PR has been merged into the master branch.  I'm marking this as "ready for review" now and I'm running a build check.

bska:
jenkins build this please

akva2 (approved):
Looks good.

bska (commented):
Build check is green.  Merging per prior approval.

PR #4877, opened 2023-09-19
reducing the bhp lower limit slightly to handle 1 bar bhp limit

GitPaean:
it turns out that 1bar is the default bhp constraint for some keywords,  and some cases are depending on it.

It is one least invading way to update OPM-flow to handle those cases. (when coming to passing the tests)

### DIFF
--- a/opm/simulators/wells/MultisegmentWellPrimaryVariables.cpp
+++ b/opm/simulators/wells/MultisegmentWellPrimaryVariables.cpp
@@ -23,6 +23,7 @@
 #include <opm/simulators/wells/MultisegmentWellPrimaryVariables.hpp>
 
 #include <opm/input/eclipse/Schedule/MSW/WellSegments.hpp>
+#include <opm/input/eclipse/Units/Units.hpp>
 
 #include <opm/material/fluidsystems/BlackOilDefaultIndexTraits.hpp>
 #include <opm/material/fluidsystems/BlackOilFluidSystem.hpp>
@@ -180,7 +181,12 @@ updateNewton(const BVectorWell& dwells,
         {
             const int sign = dwells[seg][SPres] > 0.? 1 : -1;
             const double dx_limited = sign * std::min(std::abs(dwells[seg][SPres]) * relaxation_factor, max_pressure_change);
-            value_[seg][SPres] = std::max(old_primary_variables[seg][SPres] - dx_limited, 1e5);
+            // some cases might have defaulted bhp constraint of 1 bar, we use a slightly smaller value as the bhp lower limit for Newton update
+            // so that bhp constaint can be an active control when needed.
+            // TODO: we might need to distinguish SPres of the top segment from the SPres of the other segments, since only the top segment
+            // is involved in the well constraints/control checking
+            constexpr double spres_lower_limit = 1. * unit::barsa - 1. * unit::Pascal;
+            value_[seg][SPres] = std::max(old_primary_variables[seg][SPres] - dx_limited, spres_lower_limit);
         }
 
         // update the total rate // TODO: should we have a limitation of the total rate change?
--- a/opm/simulators/wells/StandardWellPrimaryVariables.cpp
+++ b/opm/simulators/wells/StandardWellPrimaryVariables.cpp
@@ -23,6 +23,7 @@
 #include <opm/simulators/wells/StandardWellPrimaryVariables.hpp>
 
 #include <opm/common/Exceptions.hpp>
+#include <opm/input/eclipse/Units/Units.hpp>
 
 #include <dune/common/dynvector.hh>
 #include <dune/istl/bvector.hh>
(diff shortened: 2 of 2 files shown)


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
The results for the failures look fine.  But at the same time, it is a little unfortunate that this PR can affect those  cases. 

[mpi.compareECLFiles_flow+3_MULTFLT_MODEL2](https://ci.opm-project.org/job/opm-simulators-PR-builder/5357/testReport/junit/(root)/mpi/compareECLFiles_flow_3_MULTFLT_MODEL2/)
[mpi.compareECLFiles_flow+4_MINPVV_MODEL2](https://ci.opm-project.org/job/opm-simulators-PR-builder/5357/testReport/junit/(root)/mpi/compareECLFiles_flow_4_MINPVV_MODEL2/)

totto82:
Thanks for the contribution. I think this can be merged. Could you just elaborate a bit on the test failures. Is default bhp used for these cases? Or is the differences due to differences in convergence stemming from the slight modification of the lower bound.

GitPaean:
> Or is the differences due to differences in convergence stemming from the slight modification of the lower bound.

I think the above is the cause.  Because master branch could not solve the situation that wells needs to switch to 1 bar limit.  But I did not dig in to figure out what exactly changes the convergence behavior.  For the case 3_MULTFLT_MODEL2, since the middle of the running, some convergence difference show up and it affects the time stepping afterwards due to different iteration numbers. (left is the master, right is the PR. )

image

bska:
> > Or is the differences due to differences in convergence stemming from the slight modification of the lower bound.
> 
> I think the above is the cause. Because master branch could not solve the situation that wells needs to switch to 1 bar limit. But I did not dig in to figure out what exactly changes the convergence behavior. For the case `3_MULTFLT_MODEL2`, since the middle of the running, some convergence difference show up and it affects the time stepping afterwards due to different iteration numbers. [...] And the well curve difference is clearly from the different time step sizes.

Would you mind rerunning the cases using something like `--solver-max-time-step-in-days=1` to (try to) eliminate the different timestep sizes, please?

GitPaean:
> Would you mind rerunning the cases using something like --solver-max-time-step-in-days=1 to (try to) eliminate the different timestep sizes, please?

Then it will obtain same results (see the figure below). But I just do not know what triggers the difference in the residual history from the picture above. It has the same residual history until the point suddenly some small difference shows up, then the iteration numbers and time step sizes changes as well afterwards.  

image

GitPaean:
A little more looking into it, it looks like the chopping of Bhp values happens during the running. 

The master branch chops the Bhp to be 1 bar and this PR chops the Bhp to be 99999 bar. 

This explains the difference in the residual history and the small difference in the results. 
@bska  @totto82

GitPaean:
It is a little worry thing  it is a Injector that has reached the bhp lower limit of 1 bar.  But it is not in the scope of this PR. 

```
well INJ1 has bhp smaller than 1 bar, is 99999

```

GitPaean:
So it is from the Newton update, 

```
657 Starting time step 0, stepsize 30 days, at day 304/334, date = 01-Sep-2019
658 Restart file written for report step 17/39, date = 01-Sep-2019 00:00:00
659  well INJ1 has bhp smaller than 1 bar, is 99999
660  old bhp is 240.743 bar dwells[0][Bhp] 1613.02 bar
661  Newton its=14, linearizations=15 (0.0sec), linear its=172 (0.1sec)
662 

```

without looking into the well assembling process, we will not be able to see what is try with it.  But it is not in the scope of this PR. 

And also, the difference in the residual history is well explained by chopping to the 1 bar or 99999 pascal. In my opinion, the PR should be fine compared with the master branch. It does not do things worse and also it will be able to handle the situation with 1 bar bhp constraint.

bska:
> The master branch chops the Bhp to be 1 bar and this PR chops the Bhp to be 99999 bar.

At what point?  The `INJ1` well starts as a (shut) [producer](https://github.com/OPM/opm-tests/blob/c00c148359c5faf9e61ce15303274b7ee2803ab2/model2/include/hist_schedule.sch#L41) (1st November 2018), but is turned into an injector [shortly thereafter](https://github.com/OPM/opm-tests/blob/c00c148359c5faf9e61ce15303274b7ee2803ab2/model2/include/hist_schedule.sch#L61) (2nd November 2018).

bska:
> It looks like `std::numeric_limits<double>::epsilon();` is too small for this purpose.

Yeah, `epsilon()` is defined to be the smallest, normal, positive number such that `1 + epsilon() > 1` in the machine arithmetic for the type.  You can't subtract that from a number on the order of 1.0e5 and hope to get something meaningful.

I do wish we'd be more explicit about where the `1.0e5` number comes from, though.  Any chance you can write it as `1*unit::barsa` instead?

GitPaean:
jenkins build this please

GitPaean:
I am done updating the PR. Please suggest.

bska:
> I am done updating the PR. Please suggest.

Since we're getting messages of the form
```
Timestep chopped to 10.230000 days
```
I suggest we explicitly set `TEST_ARGS --solver-max-time-step-in-days=10` (or `=5` if that's a better value) in `regressionTests.cmake` for both [`3_MULTFLT_MODEL2`](https://github.com/OPM/opm-simulators/blob/50fb7b38cf1620e31bf7e031498f9567e10970e8/regressionTests.cmake#L595) and [`4_MINPVV_MODEL2`](https://github.com/OPM/opm-simulators/blob/50fb7b38cf1620e31bf7e031498f9567e10970e8/regressionTests.cmake#L609).  Then we update the reference solutions for those two regression tests and merge this PR.

GitPaean:
Another finding along the line is that the two cases have identical results, 

[3_MULTFLT_MODEL2](https://github.com/OPM/opm-simulators/blob/50fb7b38cf1620e31bf7e031498f9567e10970e8/regressionTests.cmake#L595) and [4_MINPVV_MODEL2](https://github.com/OPM/opm-simulators/blob/50fb7b38cf1620e31bf7e031498f9567e10970e8/regressionTests.cmake#L609).

image

The difference of the DATA files are 

image


Someone needs to look into whether these keywords are changing anything for the simulation.

bska:
jenkins build this please

bska (approved):
This looks good to me now and the new reference solutions from #4880/OPM/opm-tests#1047 mean that we no longer run into the instabilities exhibitied earlier.  I'll merge into master.

PR #4878, opened 2023-09-19
spesialied version for wetoil wetgas blackoil simulation

hnil:
This will break several tests and need checking before being merged but provide:
-  pvto pvtg based evaluation which is much faster than original code
-  no use of elmctx in when calculating properties
-  less prepost time when hysteresis is used since it only update relevant cells

Relperm evaluation is not  modified and is next step for fast evaluation.


bska:
jenkins build this please

GitPaean:
jenkins build this please

hnil:
Closing: this was a prototype and has been conflicting since 2023, and as the description notes it would break several tests in this form. The underlying idea -- PVTO/PVTG-based property evaluation without ElementContext, which was much faster -- is worth revisiting, and the branch is retained on my fork for that.

PR #4879, opened 2023-09-19
only outputting some network info to the rank 0 process

GitPaean:


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

bska (approved):
Very good.  This makes sense to me.  There's no need to write identical messages on all ranks for this particular case.

PR #4856, opened 2023-09-08
Changed the relative change used in adaptive time stepping

erikhide:
The current implementation of the relative change used in the adaptive time stepping (mainly in the PID controller) added pressure and saturation in a way which practically meant that saturation was ignored (since the saturation values are so much lower than the pressure values), and hence that the relative change ended up only considering changes in pressure and not in saturation.

The new implementation separates the pressure and saturation into two different relative change values, and then returns the maximum of these two.

PR #4885, opened 2023-09-21
Safer fluid evaluation

hnil:
-- safer loop
-- removed asserts which may be triggered in by strange input.

### DIFF
--- a/ebos/FIBlackOilModel.hpp
+++ b/ebos/FIBlackOilModel.hpp
@@ -61,8 +61,9 @@ public:
 
     void invalidateAndUpdateIntensiveQuantities(unsigned timeIdx) const
     {
-        this->invalidateIntensiveQuantitiesCache(timeIdx);
 
+        this->invalidateIntensiveQuantitiesCache(timeIdx);
+        OPM_BEGIN_PARALLEL_TRY_CATCH()
         // loop over all elements...
         ThreadedEntityIterator<GridView, /*codim=*/0> threadedElemIt(this->gridView_);
 #ifdef _OPENMP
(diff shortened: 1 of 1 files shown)


atgeirr (approved):
I think this should be merged, but should we add exception throws to replace the asserts?

opm/simulators/flow/BlackoilModelEbos.hpp
```diff
@@ -558,8 +558,6 @@ namespace Opm {
                         Scalar tmpSat = saturationsNew[phaseIdx] - saturationsOld[phaseIdx];
                         resultDelta += tmpSat*tmpSat;
                         resultDenom += saturationsNew[phaseIdx]*saturationsNew[phaseIdx];
-                        assert(std::isfinite(resultDelta));
-                        assert(std::isfinite(resultDenom));
```
bska, line 562: I these are not finite then we have a problem somewhere and I'd rather be told about it.  The mechanism probably should not be `assert()` though.  On another note, PR #4856 changes this area too so we might want to coordinate this with that PR.

bska (commented):
Other than the fact that diagnostic messages referring to `InvalideAndUpdateIntensiveQuantitiesOverlap` or similar will be understood only by those who have direct access to the source code I think this looks good.  I agree with @atgeirr though, the check on `isfinite(result*)` should probably remain in place as an exception because if those are not finite then we have a serious problem somewhere.

opm/simulators/flow/BlackoilModelEbos.hpp
atgeirr: For now, let's leave these two in, and we'll consider it with #4856 instead.

bska:
jenkins build this please

PR #3074, opened 2021-02-22
Temperature output for producers

plgbrts:
This is PR for temperature output reporting for producers. This is especially relevant for geothermal reservoir simulations.

A comparison with eclipse is given below for a three phase system:
image

Applied expressions derived for average well temperature: 
image

PR #4888, opened 2023-09-22
Fixes computation of temperature for distributed wells.

blattms:
Previously, we did a global summation of the size of the well_perf_data vector to determine the number of perforations of a well. In the case of distributed wells this will try to access more perforations than stored for the well in well_perf_data and hence might use data from cells that actually are not perforated by this cell. Note that for well not distributed the code worked as the summation has no effect.

This commit changes this to only query peforations on the local process. This should be enough to fix this problem.

In addition it removes the computation of connpos which is never used.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -2456,21 +2456,16 @@ namespace Opm {
             if (well.isInjector())
                 continue;
 
-            int connpos = 0;
-            for (int i = 0; i < wellID; ++i) {
-                connpos += well_perf_data_[i].size();
-            }
-            connpos *= np;
             std::array<double,2> weighted{0.0,0.0};
             auto& [weighted_temperature, total_weight] = weighted;
 
             auto& well_info = local_parallel_well_info_[wellID].get();
-            const int num_perf_this_well = well_info.communication().sum(well_perf_data_[wellID].size());
             auto& ws = this->wellState().well(wellID);
             auto& perf_data = ws.perf_data;
             auto& perf_phase_rate = perf_data.phase_rates;
 
-            for (int perf = 0; perf < num_perf_this_well; ++perf) {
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

blattms:
I closed this by accident (by removing the branch). But I need to double check my change. Not sure it is correct.

blattms:
I closed this by accident (by removing the branch). But I need to double check my change. Not sure it is correct.

blattms:
Double checked and seems correct to me: `well_perf_data_` is local.
Anyway, it would be good for @plgbrts to aslo check why it was done like this. I am good at missing stuff. Maybe it was even ill advise be me? :flushed:

plgbrts:
> Double checked and seems correct to me: `well_perf_data_` is local. Anyway, it would be good for @plgbrts to aslo check why it was done like this. I am good at missing stuff. Maybe it was even ill advise be me? 😳

I checked the PR (https://github.com/OPM/opm-simulators/pull/3074) where the global summation was introduced but cannot recall the reason for this. It seems not an ill advice from you either 😊.  (You did  advice global summation on lines 2492/2492).

I have checked your implementation with a parallel thermal run showing a perfectly matching temperature output of the producer well.

plgbrts:
> Thanks a lot for testing. The global summation that I argued should be there was to compute the average correctly. It is something different and not changed/touched here.

 I understand.

> I guess we never noticed because there is no check for thermal with distributed wells yet (we need to force the loadbalancer to do that). Putting that on my list.

Such test would be very useful ( not sure btw what it entails to _force the loadbalancer_ ) 

> Maybe we should postpone merging until there is this breaking test?

It looks like there is only an upside of your fix even without having such test now so that there is no need to postpone merging? I think you can better judge this and proceed as you prefer.

akva2 (approved):
Yeah this seems to be a clear bug and new code is correct. Sum up the local parts, do a reduction across the processes to get the global, then do the weighting.

blattms:
thanks a lot for reviewing.

PR #4895, opened 2023-09-26
Option for doing well control/status switching during local well solve

GitPaean:
This PR is a lightly cleaned up version of PR https://github.com/OPM/opm-simulators/pull/4808 . 

In this PR, a new function `iterateWellEqWithSwitching()` is added so that when we do the local solve for well equations, control/status will be updated during the iteration process, such that the converged well gets correct control/status regarding to the current reservoir state. Various change in the other parts of the code were made to make the function work as intended.

A running option local-well-solve-control-switching is added to trigger the use of the new function `iterateWellEqWithSwitching()`. In this PR, it is defaulted to be `false` at the moment, so it should not affect the running of cases when running without the running option on. 

It is developed by @steink with discussion with others.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -2175,7 +2175,7 @@ namespace Opm
         const auto& summary_state = ebosSimulator.vanguard().summaryState();
 
         // Max status switch frequency should be 2 to avoid getting stuck in cycle 
-        const int min_its_after_switch = 2;
+        constexpr int min_its_after_switch = 2;
         int its_since_last_switch = min_its_after_switch;
         int switch_count= 0;
         const auto well_status = this->wellStatus_;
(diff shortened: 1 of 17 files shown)


GitPaean:
jenkins build this please

GitPaean:
We will squash many of the commits at a later stage before merging.   At the same time, some reviewing will be required before the PR gets in.  

I am marking the PR as ready for review now.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -2155,6 +2155,99 @@ namespace Opm
...
+                const bool is_stopped = this->wellIsStopped();
+                if (this->wellHasTHPConstraints(summary_state)){
+                    this->operability_status_.can_obtain_bhp_with_thp_limit = !is_stopped;
+                    this->operability_status_.obey_thp_limit_under_bhp_limit = !is_stopped;
+                } else {
+                    this->operability_status_.operable_under_only_bhp_limit = !is_stopped;
+                }
+                // We reset the well status to it's original state. Status is updated 
+                // on the outside based on operability status
+                // TODO: this looks strange, let us check
```
steink, line 2239: For the well to update it's status to stop/shut, the flag `changed_to_stopped_this_step_` in `prepareWellBeforeAssembling` needs to be set to `true`. For this to happen, `isOperableAndSolvable()` must change from `true` to `false`, and (until the most recent commit) the well needs to be open for this to trigger. Hence, the resetting of status.

bska (approved):
I'm now mostly happy with the current state of the PR.  Maybe we can address [@steink's comment](https://github.com/OPM/opm-simulators/pull/4895#discussion_r1337067442) and coalesce some of the commits before merging?

opm/simulators/wells/StandardWell_impl.hpp
GitPaean: Thanks. I will incorporate your comments to the comments in the code.

bska:
One more thing: If this is the final edition of the PR, would you mind adding a summary of the work in the description box and making the title more meaningful, please?  If need to look at this PR at some point in the future we'll likely need a bit of context and motivation to understand the problem being solved here.

GitPaean:
jenkins build this please

GitPaean:
I believe that I have done following the requests. Please let me know if there are more comments or requests.

bska (approved):
Thanks a lot for the updates.  I'll merge this as a milestone, well aware that there is more work incoming in this area.

PR #4897, opened 2023-09-26
fixed: reorganize opencl/amgcl/cuda

akva2:
- opencl and amgcl should be listed in prereqs
- cuda should not use FindCUDA on newer cmake, only the language support. we manually have to set CUDA_FOUND.

all of this is done to fix building of downstreams of opm-simulators where the previous approach did not carry over to the generated config files as these do not use exported targets.

The buildsystem is getting to be an unwieldly mixed of old and new. We really should put in the effort soon to transition to pure target usage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,6 +126,7 @@ include(Findopm-tests)
 # with the find module
 include ("${project}-prereqs")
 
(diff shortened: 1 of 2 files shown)


akva2:

akva2:

PR #4778, opened 2023-08-07
Refactor element centroids lookup

aritorto:
The new structure LookUpCellCentroid is added, replacing code for search centroids of elements. Associated with OPM/opm-grid#672 (merged).
Currently, this new structure is used only when the rank is different from zero (CpGrid). Refactoring the code for the case zero rank (other grids), breaks many tests and it's (temporary ) postponed. 

EclBaseVanguard::centroids_ is removed. 

ProsCentroidsDataHandle.hpp now is called PropsDataHandle.hpp (centroids is removed too).

PR #4898, opened 2023-09-28
Supporting centroids lookup for rank 0 and LGRs

aritorto:
This is a follow-up/completition of #4778 where we separate the case when centroids are searched for a grid with LGRs and rank zero.


aritorto:

aritorto:

blattms (approved):
Thanks.
Looks good now. Let's see what jenkins says.

aritorto:
Thanks for idea of adding an auxiliary boolean to differentiate grid types!

blattms:
You are welcome. I'm just doing my job. :smiley:

PR #4899, opened 2023-09-28
Add DILU preconditioner

jakobtorben:
This PR adds the Diagonal-ILU (DILU) preconditioner to OPM-Simulators.

 DILU is a simplified version of the ILU(0) where only the diagonal terms are factorised, which means that we only need to store the diagonol for applying the preconditioner. More details on the method can be found here: https://netlib.org/linalg/html_templates/node64.html#SECTION00842200000000000000


opm/simulators/linalg/DILU.hpp
```diff
@@ -0,0 +1,210 @@
...
+        // we build the inverseD matrix
+        Dinv_.resize(A_.N());
+
+        // is this the correct usage?
+        update();
+    }
+
+    virtual void update() override {
+
+        using block = typename M::block_type;
```
atgeirr, line 76: Use capital initial letter for type and class names, i.e. `Block`. (Unless following another interface such as Dune which expects a particular name such as domain_type etc.)

opm/simulators/linalg/DILU.hpp
```diff
@@ -0,0 +1,210 @@
...
+ virtual void apply(X& v, const Y& d)
+    {
+        
+    // M = (D + L_A) D^-1 (D + U_A)   (a LU decomposition of M)
+    // where L_A and U_A are the strictly lower and upper parts of A and M has the properties:
+    // diag(A) = diag(M)
+    // solving the product M^-1(b-Ax) using upper and lower triangular solve
+    // z = x_k+1 - x_k = M^-1(b - Ax) = (D + L_A)^-1 D (D + U_A)^-1 (b - Ax)
+
+    typedef typename Y::block_type dblock;
```
atgeirr, line 128: Capital letter in typename. Also prefer `using` to typedef.

opm/simulators/linalg/DILU.hpp
```diff
@@ -0,0 +1,210 @@
...
+
+        // is this the correct usage?
+        update();
+    }
+
+    virtual void update() override {
+
+        using block = typename M::block_type;
+        
+        for ( auto i = A_.begin(); i != A_.end(); ++i)
```
atgeirr, line 78: Avoid using `i` or `j` as loop variables, use `row` or `row_i` here for example. Also in apply().

opm/simulators/linalg/DILU.hpp
```diff
@@ -0,0 +1,210 @@
...
+        {
+            block Dinv_temp;
+            for (auto a_ij=i->begin(); a_ij != i->end(); ++a_ij) {
```
atgeirr, line 81: Use spaces around binary operators such as `=` here and also in apply().

opm/simulators/linalg/DILU.hpp
```diff
@@ -0,0 +1,210 @@
...
+    Y y;
+    z.resize(A_.N());
+    y.resize(A_.N());
+    
+    // lower triangular solve: (D + L) y = b - Ax
+    auto endi=A_.end();
+    for (auto i=A_.begin(); i != endi; ++i)
+      {
+        dblock rhsValue(d[i.index()]);
+        auto&& rhs = Impl::asVector(rhsValue);
```
atgeirr, line 142: I do not understand why you need the `Impl::asVector()` and `Impl::asMatrix()`? I may be missing something, is this to handle the case of using scalars as block type?

opm/simulators/linalg/DILU.hpp
```diff
@@ -0,0 +1,210 @@
...
+#endif
+
+    /*! \brief Constructor.
+       Constructor gets all parameters to operate the prec.
+       \param A The matrix to operate on.
+     */
+    SeqDilu(const M& A)
+        : A_(A)
+    {
+        CheckIfDiagonalPresent<M, 1>::check(A_);
```
atgeirr, line 66: I think this also checks the inside of the blocks (if compiled with DUNE_ISTL_WITH_CHECKING), and we do not really require all nonzeroes on the diagonal, just that the diagonal block is invertible. (I found it a little hard to understand CheckIfDiagonalPresent though.) If you decide to keep it, you should include matrixutils.hh in this file and not rely on transitive includes.

opm/simulators/linalg/DILU.hpp
```diff
@@ -0,0 +1,210 @@
...
+            block Dinv_temp;
+            for (auto a_ij=i->begin(); a_ij != i->end(); ++a_ij) {
+                auto a_ji = A_[a_ij.index()].find(i.index());
+                
+                if (a_ij.index() == i.index()) {
+                    Dinv_temp += A_[i.index()][i.index()];
+                }
+
+                // if A[i, j] != 0 and A[j, i] != 0:
+                else if (a_ji != A_[a_ij.index()].end() && (a_ij.index() < i.index())) {
```
atgeirr, line 89: I misread this, please put the `else if` after the brace on line 86. Also, here you may want to check that you have a diagonal, instead of using CheckIfDiagonalPresent, since you are iterating over the matrix anyway.

atgeirr (commented):
This looks good, and should be merged eventually. I have some questions below, but mostly stylistic nitpicks. I have not really delved deeply into the numerics and checked them yet, though.

Congratulations on making your first OPM PR!

opm/simulators/linalg/DILU.hpp
```diff
@@ -0,0 +1,210 @@
...
+    // z = x_k+1 - x_k = M^-1(b - Ax) = (D + L_A)^-1 D (D + U_A)^-1 (b - Ax)
+
+    typedef typename Y::block_type dblock;
+    
+    // copy current v
+    X x(v);
+    X z;
+    Y y;
+    z.resize(A_.N());
+    y.resize(A_.N());
```
multitalentloes, line 135: These three vectors have a known size and should only be allocated once, probably in the constructor.

opm/simulators/linalg/PreconditionerFactory_impl.hpp
```diff
@@ -157,6 +159,10 @@ struct StandardPreconditioners
         F::addCreator("ILUn", [](const O& op, const P& prm, const std::function<V()>&, std::size_t, const C& comm) {
           return createParILU(op, prm, comm, prm.get<int>("ilulevel", 0));
         });
+        F::addCreator("DILU", [](const O& op, const P& prm, const std::function<V()>&, std::size_t, const C& comm) {
+                        DUNE_UNUSED_PARAMETER(prm);
```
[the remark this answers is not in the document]
jakobtorben: I agree, fixed.

opm/simulators/linalg/DILU.hpp
```diff
@@ -0,0 +1,210 @@
...
+    {
+        CheckIfDiagonalPresent<M, 1>::check(A_);
+        // we build the inverseD matrix
+        Dinv_.resize(A_.N());
+
+        // is this the correct usage?
+        update();
+    }
+
+    virtual void update() override {
```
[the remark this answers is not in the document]
jakobtorben: I agree, fixed.

opm/simulators/linalg/DILU.hpp
jakobtorben: There's a good mix between class names that come from Dune and names that are just used locally. I followed your suggestion for the ones used locally now.

opm/simulators/linalg/DILU.hpp
jakobtorben: The use was inspired by code from Dune but I agree that we should use `using` instead. I changed this throughout the code.

opm/simulators/linalg/DILU.hpp
jakobtorben: fixed

opm/simulators/linalg/DILU.hpp
jakobtorben: fixed

opm/simulators/linalg/DILU.hpp
jakobtorben: This was also inspired by code from Dune. Yes it wraps a scalar as a matrix and vector to support both:

```cpp
  /** \brief Wrap a scalar as a 1-1-matrix */
  template<class T,
    std::enable_if_t<IsNumber<T>::value, int> = 0>
  auto asMatrix(T& t)
  {
    return ScalarMatrixView<T>{&t};
  }
  ```
I am not sure if we need this functionality for our case. I removed it for now and everything still works. I don't see it being used anywhere else in OPM-Simulators. I can add it back in if you want.

File: opm/simulators/linalg/DILU.hpp (hunk 6 above)
Comment (jakobtorben) [reply opm/simulators/linalg/DILU.hpp]: I decided to remove it to avoid any potential expensive checks.

File: opm/simulators/linalg/DILU.hpp (hunk 8 above)
Comment (jakobtorben) [reply opm/simulators/linalg/DILU.hpp]: Yes you are right. This was initially used to ease the implementation but in the optimised version I removed the need for storing temporarily vectors completely, since we are only calculating the update to x.

**Comment by jakobtorben** [comment]
@blattms That is cool to hear. 
We have gotten promising results with this preconditioner so far. Using DILU as a preconditioner for BiCgStab it as fast as ILU0 and using DILU as a smoother in CPR-AMG it is even faster than using ILU0 as a smoother. We are looking into adding a parallel multicolour-DILU implementation  as well, which we expect will make it even faster.

I can add this to dune-istl as well and contribute to a release publications if needed. Just let me know when the branch is set up and maybe give me a few pointers on how to integrate it. It is based on other preconditioners in DUNE, so I expect it will be quite easy to add.

**Comment by blattms** [comment]
You don't need to wait until the branch is set up. These are two indepent things. Your contribution should be a merge request to the master branch. In addition I will merge it to the opm branch (Of course I still have to set that up. (How can one be so slow as me :sweat_smile: )

Due to my slowness (still need to remove a clang problem in DUNE) 2.9.1 is still not out. So there might a very little chance to get that in even there. Depends on my charms and the release manager (that is waiting for my fix).

File: opm/simulators/linalg/ExtraSmoothers.hpp [hunk 11]
```diff
@@ -0,0 +1,42 @@
+#ifndef OPM_EXTRASMOOTHERS_HPP
+#define OPM_EXTRASMOOTHERS_HPP
+
+#include "DILU.hpp"
+
+namespace Dune
+{
+  template <class M, class X, class Y>
```
atgeirr, line 8: Indent same as next line. I see also that the next namespace is placed on the same level, which is (I think) part of our clang-format style, so it will be a little weird anyway, but less weird is better...

opm/simulators/linalg/ExtraSmoothers.hpp
```diff
@@ -0,0 +1,42 @@
...
+     * @brief Policy for the construction of the SeqDilu smoother
+     */
+    template <class M, class X, class Y>
+    struct ConstructionTraits<SeqDilu<M, X, Y>> {
+        typedef DefaultConstructionArgs<SeqDilu<M, X, Y>> Arguments;
+#if DUNE_VERSION_NEWER(DUNE_ISTL, 2, 7)
+        static inline std::shared_ptr<SeqDilu<M, X, Y>> construct(Arguments& args)
+        {
+            return std::make_shared<SeqDilu<M, X, Y>>(
+							  args.getMatrix());
```
atgeirr, line 23: Weird formatting, also on line 30. This is due to the presence of tabs in the file. Avoid tabs in files, make your editor always output as spaces only.

opm/simulators/linalg/setupPropertyTree.cpp
```diff
@@ -111,7 +115,7 @@ setupPropertyTree(FlowLinearSolverParameters p, // Note: copying the parameters
     // No valid configuration option found.
     OPM_THROW(std::invalid_argument,
               conf + " is not a valid setting for --linear-solver-configuration."
```
atgeirr, line 117: I see this comment is outdated, the name of the option was changed to `--linear-solver` a long time ago. Can you drive-by-fix this too?

opm/simulators/linalg/PreconditionerFactory_impl.hpp
```diff
@@ -186,12 +194,23 @@ struct StandardPreconditioners
         // with the AMG hierarchy construction.
         if constexpr (std::is_same_v<O, Dune::OverlappingSchwarzOperator<M, V, V, C>>) {
           F::addCreator("amg", [](const O& op, const P& prm, const std::function<V()>&, std::size_t, const C& comm) {
+            using PrecPtr = std::shared_ptr<Dune::PreconditionerWithUpdate<V, V>>;
```
atgeirr, line 197: I think PrecPtr is already defined in this class, no need to make this alias here?

opm/simulators/linalg/PreconditionerFactory_impl.hpp
```diff
@@ -186,12 +194,23 @@ struct StandardPreconditioners
...
               using Smoother = Opm::ParallelOverlappingILU0<M, V, V, C>;
               auto crit = AMGHelper<O,C,M,V>::criterion(prm);
               auto sargs = AMGSmootherArgsHelper<Smoother>::args(prm);
-              return std::make_shared<Dune::Amg::AMGCPR<O, V, Smoother, C>>(op, crit, sargs, comm);
+              PrecPtr prec = std::make_shared<Dune::Amg::AMGCPR<O, V, Smoother, C>>(op, crit, sargs, comm);
+              return prec;
+            }
+            else if (smoother == "DILU") {
+              using Smoother = Dune::SeqDilu<M, V, V>;
+              using ParSmoother = Dune::BlockPreconditioner<V, V, C, Smoother>;
```
atgeirr, line 208: Suggest changing these two typenames: `Smoother -> SeqSmoother` and `ParSmoother` -> `Smoother` (also remember to change on lines 209 and 212). Then the Smoother is the one actually used, the SeqSmoother is just a building block and not the other way around.

opm/simulators/linalg/DILU.hpp
```diff
@@ -0,0 +1,183 @@
...
+        auto endi = A_.end();
+        for ( auto row_i = A_.begin(); row_i != endi; ++row_i) {
+            Dinv_[row_i.index()] = A_[row_i.index()][row_i.index()];
+        }
+        
+        for ( auto row_i = A_.begin(); row_i != endi; ++row_i)
+        {
+           auto Dinv_temp = Dinv_[row_i.index()];
+            for (auto a_ij = row_i->begin(); a_ij.index() < row_i.index(); ++a_ij) {
+                auto a_ji = A_[a_ij.index()].find(row_i.index());
```
atgeirr, line 83: Maybe add a local variable `const auto col_j = a_ij.index()` or similar since that is used a lot, I think it would be easier to read.

Also, you could use `const auto a_ji = ...`.

opm/simulators/linalg/DILU.hpp
```diff
@@ -0,0 +1,183 @@
...
+        OPM_TIMEBLOCK(update);
+        
+        auto endi = A_.end();
+        for ( auto row_i = A_.begin(); row_i != endi; ++row_i) {
+            Dinv_[row_i.index()] = A_[row_i.index()][row_i.index()];
+        }
+        
+        for ( auto row_i = A_.begin(); row_i != endi; ++row_i)
+        {
+           auto Dinv_temp = Dinv_[row_i.index()];
```
atgeirr, line 81: Indent.

atgeirr (commented):
I may have some more comments, but I'll make a break now and post those I have so far.

opm/simulators/linalg/DILU.hpp
```diff
@@ -0,0 +1,183 @@
...
+    /*!
+       \brief Apply the preconditioner.
+       \copydoc Preconditioner::apply(X&,const Y&)
+    */
+    virtual void apply(X& v, const Y& d) override
+    {
+        // M = (D + L_A) D^-1 (D + U_A)   (a LU decomposition of M)
+        // where L_A and U_A are the strictly lower and upper parts of A and M has the properties:
+        // diag(A) = diag(M)
+        // solving the product M^-1(b-Ax) using upper and lower triangular solve
```
atgeirr, line 115: Referring to `b - Ax` here, maybe rewrite slightly the comments referring to `d` and `v` instead?

opm/simulators/linalg/DILU.hpp
```diff
@@ -0,0 +1,183 @@
...
+            // storing y_i in v_i
+            Dinv_[row_i.index()].mv(rhs, v[row_i.index()]);  // (D + L_A)_ii = D_i
+        }
+
+        // upper triangular solve: (D + U_A) z = Dy 
+        auto rendi = A_.beforeBegin();
+        for (auto row_i = A_.beforeEnd(); row_i != rendi; --row_i)
+        {
+            // rhs = 0
+            Xblock rhs;
```
atgeirr, line 141: Are we 100% sure this will initialize rhs to zero always?

opm/simulators/linalg/setupPropertyTree.cpp
blattms: this should be done in a separate PR and marked for the release.

atgeirr: Good point. I will create such a fixup PR.

opm/simulators/linalg/ExtraSmoothers.hpp
jakobtorben: For some reason VSCode used 2 spaces for this file only... Fixed

opm/simulators/linalg/PreconditionerFactory_impl.hpp
jakobtorben: I don't think it is defined in the StandardPreconditioners struct. At least it doesn't compile when I don't define this alias.

opm/simulators/linalg/PreconditionerFactory_impl.hpp
jakobtorben: Good suggestion, changed it now.

opm/simulators/linalg/DILU.hpp
jakobtorben: Good suggestion, this will increase the readability. I changed this now.

opm/simulators/linalg/DILU.hpp
jakobtorben: I was a bit confused with the use of v,d and b,x in OPM/Dune initially, but now that the algorithm is implemented to work with the defect `d` and update `v` it makes sense to change the comments.

opm/simulators/linalg/DILU.hpp
jakobtorben: If the block is an `OPM::MatrixBlock` this should always be the case:

https://github.com/OPM/opm-models/blob/e945446f48b9bb19ea72f4923f4465e24639625e/opm/simulators/linalg/matrixblock.hh#L236

But I changed it to `Xblock rhs(0.0);` to be on the safe side and increase the readability.

opm/simulators/linalg/ExtraSmoothers.hpp
jakobtorben: My editor uses spaces, but it seems like something weird happened here. Fixed it now.

opm/simulators/linalg/PreconditionerFactory_impl.hpp
atgeirr: You are correct, I forgot that this was moved to the StandardPreconditioner helper. No need to change this.

atgeirr:
jenkins build this please

atgeirr:
jenkins build this please

PR #4900, opened 2023-09-29
fixed error using NNC with thermal

hnil:
* fixed error using NNZ with thermal

also:
* fixed but not checked for diffusion
* bug for zero thermal diffusion (nan in derivatives)
* added checking of consistency between input and taged phases

### DIFF
--- a/ebos/ecltransmissibility_impl.hh
+++ b/ebos/ecltransmissibility_impl.hh
@@ -301,6 +301,14 @@ update(bool global, const std::function<unsigned int(unsigned int)>& map, const 
                 // *added to* by applyNncToGridTrans_() later.
                 assert(outsideFaceIdx == -1);
                 trans_[isId(elemIdx, outsideElemIdx)] = 0.0;
+                if (enableEnergy_){
+                    thermalHalfTrans_[directionalIsId(elemIdx, outsideElemIdx)] = 0.0;
+                    thermalHalfTrans_[directionalIsId(outsideElemIdx, elemIdx)] = 0.0;
+                }
+
+                if (updateDiffusivity) {
+                    diffusivity_[isId(elemIdx, outsideElemIdx)] = 0.0;
+                }
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


GitPaean:
jenkins build this please

ebos/ecltransmissibility_impl.hh
```diff
@@ -866,13 +874,40 @@ applyNncToGridTrans_(const std::unordered_map<std::size_t,int>& cartesianToCompr
...
+        // if (enableDiffusivity_) {
+        //     auto candidate = diffusivity_.find(isId(low, high));
+        //     if (candidate != trans_.end()) {
+        //         // NNC is represented by the grid and might be a neighboring connection
+        //         // In this case the transmissibilty is added to the value already
+        //         // set or computed.
+        //         candidate->second += nncEntry.transDiffusion;
+        //     }
+        // }
     }
```
blattms, line 911: maybe remove this commented code to clean up?

hnil:
Should we add the test case?

blattms (approved):
Looks good to me. but I'll leave the floor to the requested reviewers.

GitPaean:
> Should we add the test case?

Probably we can. But we should ask for permission.

ebos/ecltransmissibility_impl.hh
hnil: I just added it to show where to ad properties if one need it for diffusion and energy and it is added in input + nncEntry.

hnil:
To make a test one only need to add NNC between first and last cell in the CO2STORE test.

ebos/ecltransmissibility_impl.hh
blattms: I see.
Indeed, NNC's are treated differently in OPM depending on their direction. In Z-direction (connections over pinched out cells) NNCs are repesented as normal faces. In this case Energy and Diffusion will be computed for them. For all explicitly specified NNCs this is not the case and we would need an additional custom keyword.

blattms:
I am merging this. The fix is good and adding a test can be done in a separate PR.

I am 100% sure that we have permission to derive extra tests from existing ones in opm-tests. That is covered by the license.

PR #4901, opened 2023-09-29
No connection without pinch and inactive/collapsed layers inbetween.

hnil:
Set no connections in tranz if vertical cells has more layers in between and pinch is not set


bska:
jenkins build this please

blattms:
Please bear with my, I have one more question:

It seems like diffusivity might still be nonzero [ecltransmissibility_impl.hh#L444-L476](/OPM/opm-simulators/blob/049cc36727d48b91587be08fbad2fcb167495285/ebos/ecltransmissibility_impl.hh#L444-L476) for the connection. Does that mean that we still can get flow over this connection even with this PR?

tskille:
This PR has been tested on a full field model in Equinor and on several smaller test modes. Results are excellent and I'm looking forward to see this PR merged into maser.

blattms:
Thanks a lot for the PR and the testing.

PR #3158, opened 2021-04-14
shut/stop unsolvable wells

totto82:
This adds a check for solvability in the operability check. 
Since the checking for solvability involves a full inner iteration for each well, this PR also
makes the usage of inner iterations obligatory. Hence it removes the parameter that controls this. 

I have tested this on the Norne prediction family and also on some other prediction cases with good results in terms of avoiding oscillatory solutions and improved performance due to less convergence issues. 

Currently this feature only works for producers. Testing indicates that this also is beneficial for injectors, but it run into some MPI issues that I will need some more time to figure out.

PR #4902, opened 2023-10-01
Fix shut-in logic for wells with shut-intructions

steink:
This fix was originally part of a previous PR, but was removed to ease testing (it breaks a lot of tests), hence this separate PR.

Currently, a well with shut-in instruction "shut" that changes to non-operable during an iteration will give the message _well xxx gets SHUT during iteration_ , but no further actions are triggered. This is due to the flag _changed_to_stopped_this_step__ not being set.

The logic is a bit intricate, but for a well to get stopped/shut it needs to be found non-operable at the end of the time-step in
`timeStepSucceeded` -> `updateWellTestState` -> `checkWellOperability`
However, unless  _changed_to_stopped_this_step__ is true, it's operability is not updated, and the closing procedure in 
`updateWellTestState` -> `updateWellTestStatePhysical` -> `close_well`
is not triggered.

With this PR, a non-operable well will give the message _well xxx gets STOPPED during iteration_ (irrespective of it's shut-in instructions), and at the end of the time-step get closed according to it's shut/stop instruction (with accompanied message).

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -626,23 +626,9 @@ namespace Opm
         const bool well_operable = this->operability_status_.isOperableAndSolvable();
 
         if (!well_operable && old_well_operable) {
-            if (this->param_.local_well_solver_control_switching_) {
-                deferred_logger.info(" well " + this->name() + " gets STOPPED during iteration ");
-                this->stopWell();
-                changed_to_stopped_this_step_ = true;
-            } else {
-                // \Note: keep the old manner for now for testing checking.
-                // Will be investgiated and fixed in a later PR
-                if (this->well_ecl_.getAutomaticShutIn()) {
-                    deferred_logger.info(" well " + this->name() + " gets SHUT during iteration ");
-                } else {
-                    if (!this->wellIsStopped()) {
-                        deferred_logger.info(" well " + this->name() + " gets STOPPED during iteration ");
-                        this->stopWell();
-                        changed_to_stopped_this_step_ = true;
-                    }
-                }
-            }
+            deferred_logger.info(" well " + this->name() + " gets STOPPED during iteration ");
+            this->stopWell();
+            changed_to_stopped_this_step_ = true;
         } else if (well_operable && !old_well_operable) {
             deferred_logger.info(" well " + this->name() + " gets REVIVED during iteration ");
             this->openWell();


bska:
jenkins build this please

bska:
benchmark please

steink:
I've gone through most of the failures. Will look at the rest and report back.

steink:
I've gone through the tests:

image

Three are three main issues:
1.	Many of the tests under wtest are very sensitive time-stepping and minor changes have huge effect on output. By using  `--enable-tuning=true`, results are more consistent and agreement between master and this pr are good. Also the UDQ_ACTIONX is very sensitive but produce more consistent results with shorter time-steps.
2.	WVFPEXP-02: for this test the pr shuts in one of the producers although it should be able to produce (it is stopped and is not able to re-open, operability checks fail), see plot. Master also reports shutting here,  but rather continues, fails to converge and is able to recover after time-step chopping. 
image

3.	01-WGRUPCON: this test clearly shows the problem with the current missing logic in master. It reports shutting, but rather continues and runs into convergence problems and time-step chopping. For this case it also ends up in the un-stable region of the vfp-curve. Finally, it gets shut due to convergence issues. PR produces output in close agreement with other simulators for this case.
image

To sum up, I think the current logic in master is not as it was intended to be (there is no reason why a well with shut-in instruction STOP gets closed when it can't operate, while one with SHUT does not). There are certain cases, however, where convergence failures and subsequent time-step chopping (as in master) may help in keeping a difficult well open longer, but I don’t think this is a viable approach for dealing with larger cases with many wells (since it lieads to a large amount of chopping).

bska:
> I've gone through the tests:

Thank you so much for your diligent testing.  I agree with your assessment that the current master logic is untenable and that we should apply `--enable-tuning=true` or `--solver-max-time-step-in-days=N` to the sensitive tests in order to improve determinism.  The one test that worries me a bit, however, is [`compareParallelSim_flow_TEST_RXFT`](https://ci.opm-project.org/job/opm-simulators-PR-builder/5418/testReport/(root)/mpi/compareParallelSim_flow_TEST_RXFT/).  Our `compareParallelSim_flow_<CASE>` tests run the `<CASE>` in sequential mode and then in parallel mode (`mpirun -np 4`) and compare the results of the two simulations.  We cannot change the reference results to fix any mismatch here.  It might be okay/sufficient to relax the tolerances or maybe we need to change the timestep sizes here too.

steink:
> Our compareParallelSim_flow_<CASE> tests run the <CASE> in sequential mode and then in parallel mode (mpirun -np 4) and compare the results of the two simulations.

I misunderstood this, can have a second look what's happening here.

steink:
Yes, reporoduced this locally. Time-steps are slightly different. Maybe this is also a test where tuning should be enabled (then there is at least no time-stepping issue here)
image

GitPaean:
> Yes, reporoduced this locally. Time-steps are slightly different. Maybe this is also a test where tuning should be enabled (then there is at least no time-stepping issue here)


It looks like the small difference in the iteration numbers changes the time step a little bit? How does it looks like if you test the case with --solver-max-time-step-in-days=2 or 1, or --enable-tuning=true ?

steink:
> How does it looks like if you test the case with --solver-max-time-step-in-days=2 or 1, or --enable-tuning=true ?

Time steps are identical both for both the cases `--enable-tuning=true ` (1 day steps) and `--solver-max-time-step-in-days=2`.

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: drogon - Threads: 1                | 1.006    |
| opm-git | OPM Benchmark: drogon - Threads: 8                | 1.011    |
| opm-git | OPM Benchmark: punqs3 - Threads: 1                | 1.002    |
| opm-git | OPM Benchmark: punqs3 - Threads: 8                | 1        |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 1         | 1.007    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 8         | 1        |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 0.994    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 0.999    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 1.001    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 1.006    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 1.002    |

FAILED RUNS: 
OPM Benchmark: smeaheia - Threads: 1 - opm-simulators 4902: Fix shut-in logic for wells with shut-intructions: The test run did not produce a result. E: mpirun was unable to launch the specified application as it could not access
OPM Benchmark: smeaheia - Threads: 8 - opm-simulators 4902: Fix shut-in logic for wells with shut-intructions: The test run did not produce a result. E: mpirun was unable to launch the specified application as it could not access

View result details @ https://www.ytelses.com/opm/?page=result&id=2250

GitPaean:
> OPM Benchmark: smeaheia - Threads: 1 - opm-simulators 4902: Fix shut-in logic for wells with shut-intructions: The test run did not produce a result. E: mpirun was unable to launch the specified application as it could not access
OPM Benchmark: smeaheia - Threads: 8 - opm-simulators 4902: Fix shut-in logic for wells with shut-intructions: The test run did not produce a result. E: mpirun was unable to launch the specified application as it could not access

We need to assess the cause of the running of this case for this PR. @bska @alfbr do we have access to this specific case?

GitPaean:
This code change in this PR is easy. But we need to discuss and agree whether we want the consequence , mostly, the well can be solved and converged and it might be SHUT with this PR. @steink @hnil @atgeirr @totto82

steink:
In my opinion the current behavior in master is rather problematic, and unless fixed, I see no way of preventing non-operable wells causing convergence issues since they are not closed until deemed _constantly failing_ (it also seems quite random that this is the sitation only for _SHUT_-instructions and not for _STOP_ ).  

With the PR, we may run into cases like WVFPEXP-02, but these are in my opinion easier to attack/fix than the convergence issues above.

steink:
Just to illustrate: by changing the automatic shut-in (item 9 of WELSPECS) from SHUT to STOP we see the same behaviour also in master.
image

GitPaean:
> Just to illustrate: by changing the automatic shut-in (item 9 of WELSPECS) from SHUT to STOP we see the same behaviour also in master. image

This is indeed a very good argument and point.  So the question becomes how to keep this well (can be solved and get converged) continuing running after this PR.  In the master branch, it is more like a fallout while not (at least not totally) intentional.

bska:
As far as I can tell, the current logic in master was introduced in PR #3158 (commit b5e29684f) and has remained unchanged since then.  There isn't a lot of reasoning in the discussion of that PR for why the current logic is the way it is, but I agree that it does not make sense to issue a message that a well is getting "STOPPED" and then not actually stopping the well.

I am in favour of merging this PR unless someone (@totto82?), (@atgeirr?) can justify the current logic.

GitPaean:
Should we care about the benchmark failure?
```
FAILED RUNS:
OPM Benchmark: smeaheia - Threads: 1 - opm-simulators 4902: Fix shut-in logic for wells with shut-intructions: The test run did not produce a result. E: mpirun was unable to launch the specified application as it could not access
OPM Benchmark: smeaheia - Threads: 8 - opm-simulators 4902: Fix shut-in logic for wells with shut-intructions: The test run did not produce a result. E: mpirun was unable to launch the specified application as it could not access
```

blattms:
> I'm tempted to say no. The [ytelses.com/daily](https://ytelses.com/daily/) service has not been operational for a long time now.

There is currently a build failure for OPM on their system. I will try to replicate now.

totto82:
> There isn't a lot of reasoning in the discussion of that PR for why the current logic is the way it is, but I agree that it does not make sense to issue a message that a well is getting "STOPPED" and then not actually stopping the well.

Note that inoperable wells with "SHUT" are not included in the solve at all. So yes they are stopped but shut. Inoperable wells that are "STOPPED" are kept in the system with rate = 0.0. This PR will consider the inoperable wells stopped not shut. 

I don't understand why this PR is a fix?

steink:
> This PR will consider the inoperable wells stopped not shut.

Not really, it will treat the well as _temporarily_ stopped and if the well still is stopped when the time-step has completed, it will _permanently_ be stopped or shut according to it's shut-in instructions.

totto82:
> Not really, it will treat the well as temporarily stopped and if the well still is stopped when the time-step has completed, it will permanently be stopped or shut according to it's shut-in instructions.

Agree. But why do you want them to be temporary stopped? Is the issue related to the fact that inoperable wells that are shut are treated as unconverged? 

```
            if (well->isOperableAndSolvable() || well->wellIsStopped()) {
                const auto& summary_state = ebosSimulator_.vanguard().summaryState();
                local_report += well->getWellConvergence(
                        summary_state, this->wellState(), B_avg, local_deferredLogger,
                        iterationIdx > param_.strict_outer_iter_wells_);
            } else {
                ConvergenceReport report;
                using CR = ConvergenceReport;
                report.setWellFailed({CR::WellFailure::Type::Unsolvable, CR::Severity::Normal, -1, well->name()});
                local_report += report;
            }
```

 See PR https://github.com/OPM/opm-simulators/pull/4846/files

steink:
> But why do you want them to be temporary stopped?

We want to allow the wells to jump between operable/non-operable during Newton iterations. If an inoperable well is not stopped, it will typically not converge (since rates will be chopped to zero and convergence stagnates). I don't think the logic around this should depend on the shut-in instructions for a well, since shut-in is treated when the time-step is completed.   

> Is the issue related to the fact that inoperable wells that are shut are treated as unconverged?

I think it is related although maybe coming in from a different angle. The problem I was seeing is that inoperable wells with shut-in instructions "SHUT" were **not** shut, but rather kept open and leading to convergence issues and time-step chopping.

totto82:
Ok. I see your point. Setting the wells to stop (i.e. zero rate) improves convergence during the iterations. Then I am fine with this change.

steink:
@totto82  wouldn't this PR also help to resolve the issue of https://github.com/OPM/opm-simulators/pull/4846, since then  `well->wellIsStopped()` will be true also for wells with shut instructions?
```
            if (well->isOperableAndSolvable() || well->wellIsStopped()) {
                const auto& summary_state = ebosSimulator_.vanguard().summaryState();
                local_report += well->getWellConvergence(
                        summary_state, this->wellState(), B_avg, local_deferredLogger,
                        iterationIdx > param_.strict_outer_iter_wells_);
            } else {
                ConvergenceReport report;
                using CR = ConvergenceReport;
                report.setWellFailed({CR::WellFailure::Type::Unsolvable, CR::Severity::Normal, -1, well->name()});
                local_report += report;
            }
```

totto82:
> @totto82 wouldn't this PR also help to resolve the issue of https://github.com/OPM/opm-simulators/pull/4846, since then well->wellIsStopped() will be true also for wells with shut instructions?

Yes, I think so. I will test the model that triggered the PR.

bska:

bska (approved):
I like it.  Let's add `--enable-tuning=true` or `--solver-max-time-step-in-days=N` to the sensitive tests and then merge this.

atgeirr:
> > @totto82 wouldn't this PR also help to resolve the issue of #4846, since then well->wellIsStopped() will be true also for wells with shut instructions?
> 
> Yes, I think so. I will test the model that triggered the PR.

What did you find on this?

> So the question becomes how to keep this well (can be solved and get converged) continuing running after this PR.

I agree with Kai here. Our current behaviour allows us to keep some wells alive that would otherwise get shut, and that seems to me to be a mostly good thing. The counterargment is that these wells cause difficulties (convergence problems and timestep cuts) while remaining open. My question is this: is it wrong to keep these wells open? @steink says about the 01-WGRUPCON test that "For this case it also ends up in the un-stable region of the vfp-curve" which probably means it *should* have been closed. Does that apply in general? If so, then keeping the wells open longer is not a good thing, and we should *not* force them to stay open.

In any case, we obviously must have consistent behaviour between STOP and SHUT-configured wells. I therefore support merging this, but I would like to know if it is likely to require follow-on work to keep wells open when possible, or not.

totto82:
> What did you find on this?

It helps. From my point of view this change can be merged.

atgeirr:
> I like it. Let's add `--enable-tuning=true` or `--solver-max-time-step-in-days=N` to the sensitive tests and then merge this.

That seems to be the consensus, please go ahead!

blattms:
Too late for  release. I am removing the milestone.

bska:
jenkins build this please

PR #4903, opened 2023-10-02
[cmake] Always demand rocblas to be there when activating rocsparse

blattms:
This was somehow only done for hip version >=5.3, but Debian bookworm is missing rocblas and using hipcc 2.3.0 and therefore compilation failed there.


blattms:

blattms:
shameless selfmerge :grimacing:, sorry.

PR #4904, opened 2023-10-02
Always allocate buffers for storing flows if BFLOW[I|J|K] is requested

vkip:
Currently these buffers are not allocated unless restart output is requested, causing segfault initially if BFLOW[I|J|K] is requested without RPTRST in the SOLUTION section.


vkip:
jenkins build this please

bska (approved):
Yes, we need that precaution for now.  I do believe we should revise the implementation here at some point because it's rather expensive as of right now, but at least we should not generate out-of-bounds subscripts until we do.  I'll merge into master.

PR #4907, opened 2023-10-04
Add missing reverse flow derivatives for valve/icd/aicd

steink:
In the case when the flow direction through valves/icd's is reversed, some derivatives are currently omitted (there is no _room_ for them in the current implementation). This PR fixes this by performing a second call to the respective functions and by doing a seperate assembly of the _extra_ derivatives. This is the same approach that was taken for acceleration pressure loss in #4640 (which helped in preventing singular well systems). 

The changes here have been tested on larger models and although results are not very conclusive, some positive effects have been observed. I think it's not unlikely though, that we will run into cases where these derivatives really matter (convergence and/or singularities), so unless they turn out to make things worse I think they should be incuded.

### DIFF
--- a/opm/simulators/wells/MultisegmentWellSegments.cpp
+++ b/opm/simulators/wells/MultisegmentWellSegments.cpp
@@ -498,7 +498,8 @@ getSurfaceVolume(const EvalWell& temperature,
 template<class FluidSystem, class Indices, class Scalar>
 typename MultisegmentWellSegments<FluidSystem,Indices,Scalar>::EvalWell
 MultisegmentWellSegments<FluidSystem,Indices,Scalar>::
-getFrictionPressureLoss(const int seg, const bool return_extra_derivatives) const
+getFrictionPressureLoss(const int seg, 
+                        const bool extra_reverse_flow_derivatives /*false*/) const
 {
     EvalWell mass_rate = mass_rates_[seg];
     const int seg_upwind = upwinding_segments_[seg];
@@ -511,7 +512,7 @@ getFrictionPressureLoss(const int seg, const bool return_extra_derivatives) cons
     // at segment node while fraction derivatives are given at upwind node.
 
     if (seg != seg_upwind) {
-        if (!return_extra_derivatives){
+        if (!extra_reverse_flow_derivatives){
             constexpr int WQTotal = Indices::numEq + PrimaryVariables::WQTotal;
             constexpr int SPres = Indices::numEq + PrimaryVariables::SPres;
             density.setDerivative(WQTotal, 0.0);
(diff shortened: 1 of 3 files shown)


GitPaean:
jenkins build this please

GitPaean:
Let me know if you are done development. If yes, please click ready for review.

steink:
Have done some testing on a larger model, and observe mild positive effects on convergence. Might not be extremely important, but at least it's an insurance againts potential problems for the special cases for which this PR has an effect.

bska:
jenkins build this please

GitPaean:
Sorry for the delay. Will have a look at it today or tomorrow.

steink:
> Sorry for the delay. Will have a look at it today or tomorrow.

No problem, I've also requested review from @bska. Probably suffecient that only one of you looks at it since there are no failures.

opm/simulators/wells/MultisegmentWellSegments.cpp
```diff
@@ -583,17 +584,39 @@ pressureDropSpiralICD(const int seg) const
...
+        if (!return_extra_derivatives){
+            zero_mask[PrimaryVariables::WQTotal] = true;
+            zero_mask[PrimaryVariables::SPres] = true;
+        } else {
+            if (PrimaryVariables::has_water){
+                zero_mask[PrimaryVariables::WFrac] = true;
+            }
+            if (PrimaryVariables::has_gas){
+                zero_mask[PrimaryVariables::GFrac] = true;
+            }
```
bska, line 605: These could probably be
```C++
if constexpr (PrimaryVariables::has_X) {
    // ...
}
```
instead, but I don't know that marking them as such would give any particular benefit here.  The compiler should be able to infer that the conditions [are constant](https://github.com/OPM/opm-simulators/blob/b60eb25285cc2cf1a78376bb64249e6af7d0ed7c/opm/simulators/wells/MultisegmentWellPrimaryVariables.hpp#L59-L61) anyway.

opm/simulators/wells/MultisegmentWellSegments.hpp
```diff
@@ -74,15 +74,18 @@ class MultisegmentWellSegments
...
     EvalWell pressureDropAutoICD(const int seg,
-                                 const UnitSystem& unit_system) const;
+                                 const UnitSystem& unit_system,
+                                 const bool return_extra_derivatives) const;
 
     // pressure drop for sub-critical valve (WSEGVALV)
-    EvalWell pressureDropValve(const int seg) const;
-
+    EvalWell pressureDropValve(const int seg, 
+                               const bool return_extra_derivatives) const;
```
bska, line 87: Maybe we could have a default value of `const bool return_extra_derivatives = false` for these functions?  That way the existing calls could remain unchanged while the new function calls for the "extra" derivatives would stand out more.

bska (approved):
It wasn't immediately obvious to me why you would call `mass_rate.clearDerivatives()` only in the `! return_extra_derivatives` case, but I think I get it now.  I noticed a couple of small potential changes, but I'd be happy to merge the current version as is if you don't want to make any additional changes here.

steink:
Thanks @bska! 
I've updated the PR with your suggestions (did also the same for the frictional pressure drop term in attempt to keep things consistent).  Also changed the option-name to `extra_reverse_flow_derivatives` which might be slightly more descriptive.

bska (approved):
Thanks a lot for the updates.  This looks good to me now and I'll merge into master provided a new build check comes back green.

bska:
jenkins build this please

bska:
> This looks good to me now and I'll merge into master provided a new build check comes back green.

The build check is green.  I'll merge into master.  Thanks a lot for the fix!

PR #4909, opened 2023-10-04
[bugfix] Fix stopping network update in parallel at max steps.

blattms:
We did only break in the loop for rank 0 and not the other ones. Hence all other processes kept iterating beyond the maximum number of allowed iterations. This lead to hard to find crashes because of non-matching MPI communication.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1125,8 +1125,8 @@ namespace Opm {
 
             if (network_update_iteration >= max_iteration ) {
                 if (terminal_output_) {
-                    local_deferredLogger.info("maximum of " + std::to_string(max_iteration) + " iterations has been used, we stop the network update now, "
-                                              "the simulation will continue with unconvergeed network results");
+                    local_deferredLogger.info("maximum of " + std::to_string(max_iteration) + " iterations has been used, we stop the network update now. "
+                                              "The simulation will continue with unconverged network results");
                 }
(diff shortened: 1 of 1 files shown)


blattms:
I will probably selfmerge as this shoud be uncontroversial.

blattms:
jenkins build this please

GitPaean (approved):
Thanks for fixing this. It is a bug that I have to see is a little not that obvious.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1122,9 +1122,12 @@ namespace Opm {
...
                     updateWellControlsAndNetworkIteration(mandatory_network_balance, relax_network_balance, dt,local_deferredLogger);
             ++network_update_iteration;
-            if (terminal_output_ && (network_update_iteration >= max_iteration) ) {
-                local_deferredLogger.info("maximum of " + std::to_string(max_iteration) + " iterations has been used, we stop the network update now, "
-                                                                                          "the simulation will continue with unconvergeed network results");
+
+            if (network_update_iteration >= max_iteration ) {
+                if (terminal_output_) {
+                    local_deferredLogger.info("maximum of " + std::to_string(max_iteration) + " iterations has been used, we stop the network update now, "
+                                              "the simulation will continue with unconvergeed network results");
```
akva2, line 1129: fix the converged while at it?

blattms: done

PR #4911, opened 2023-10-05
mark variable maybe_unused

akva2:
quells a clang warning.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1441,7 +1441,7 @@ namespace Opm
         double relaxation_factor = 1.;
         const double min_relaxation_factor = 0.6;
         bool converged = false;
-        int stagnate_count = 0;
+        [[maybe_unused]] int stagnate_count = 0;
         bool relax_convergence = false;
         this->regularize_ = false;
 


opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1441,7 +1441,7 @@ namespace Opm
         double relaxation_factor = 1.;
         const double min_relaxation_factor = 0.6;
         bool converged = false;
-        int stagnate_count = 0;
+        [[maybe_unused]] int stagnate_count = 0;
```
bska, line 1444: I guess clang is smart enough to realise that nothing reads the [incremented count](https://github.com/OPM/opm-simulators/blob/900056e8d719d1569b1d6d6e4cbd7d624f2f95b6/opm/simulators/wells/MultisegmentWell_impl.hpp#L1523) once it's complete.  In any case, there's more work coming in this area&ndash;notably a refactoring to extract the common parts of `iterateWellEqWithSwitching()` and `iterateWellEqWithControl()`, so this is a good fix in the short term.

bska (approved):
Compilers are clearly getting better at analysing these kinds of data flow graphs.  I'll merge into master.

opm/simulators/wells/MultisegmentWell_impl.hpp
akva2: yeah, clang is getting annoyingly good at looking through if (false) shenanigans etc ;)

PR #4912, opened 2023-10-06
fix issue when BCPROP is not set initially

totto82:
See https://github.com/OPM/opm-models/pull/833

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1195,7 +1195,7 @@ public:
                 values.setThermalFlow(context, spaceIdx, timeIdx, boundaryFluidState(globalDofIdx, indexInInside));
             else if (type == BCType::FREE || type == BCType::DIRICHLET)
                 values.setFreeFlow(context, spaceIdx, timeIdx, boundaryFluidState(globalDofIdx, indexInInside));
-            else
+            else if (type == BCType::RATE)
                 values.setMassRate(massrate, pvtRegionIdx);
         }
     }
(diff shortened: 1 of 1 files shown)


ebos/eclproblem.hh
```diff
@@ -1193,8 +1193,10 @@ public:
                 values.setThermalFlow(context, spaceIdx, timeIdx, boundaryFluidState(globalDofIdx, indexInInside));
             else if (type == BCType::FREE || type == BCType::DIRICHLET)
                 values.setFreeFlow(context, spaceIdx, timeIdx, boundaryFluidState(globalDofIdx, indexInInside));
-            else
+            else if (type == BCType::RATE)
                 values.setMassRate(massrate, pvtRegionIdx);
+            else
```
akva2, line 1198: won't this overwrite the thermal condition from the if above this nest?

totto82: I dont think we use enableThermalFluxBoundaries much but youre right. I will fix. Thanks for noticing.

ebos/eclproblem.hh
```diff
@@ -1459,103 +1459,105 @@ public:
...
-            std::array<Scalar, numPhases> pc = {0};
-            const auto& matParams = materialLawParams(globalDofIdx);
-            MaterialLaw::capillaryPressures(pc, matParams, fluidState);
-            Valgrind::CheckDefined(pressure);
-            Valgrind::CheckDefined(pc);
-            for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
-                if (!FluidSystem::phaseIsActive(phaseIdx))
-                    continue;
+        const auto& bcprop = this->simulator().vanguard().schedule()[this->episodeIndex()].bcprop;
+        if (bcprop.size() > 0){
```
atgeirr, line 1463: If I understand correctly, unless we have a dirichlet condition we just return the initialFluidStates_[globalDofIdx], and that is not changed by this PR. The new if her is just to avoid indexing into an empty bcprop?

Nitpick: space before {

ebos/eclproblem.hh
```diff
@@ -1459,103 +1459,105 @@ public:
     const InitialFluidState boundaryFluidState(unsigned globalDofIdx, const int directionId) const
     {
         OPM_TIMEBLOCK_LOCAL(boundaryFluidState);
-        FaceDir::DirEnum dir = FaceDir::FromIntersectionIndex(directionId);
-        assert(bcindex_(dir)[globalDofIdx] > 0);
```
atgeirr, line 1463: Why remove the assert()? What are the requirements/range for this index?

atgeirr (commented):
Looks like a very minor change to me, that should be merged.

ebos/eclproblem.hh
totto82: Thanks for noticing. I just removed it during debugging and forgot to add it back.

atgeirr (approved):

totto82:
jenkins build this opm-models=833 update_data please

ebos/eclproblem.hh
totto82: Seems like adding the assert() triggered a test failure. I will investigate.

totto82: Since we now ask for the boundaryState at every boundary face (also the ones where no boundary conditions are set) we can not use assert here. I instead added a explicit check and return to avoid accessing the bcprops for  invalid indices.

PR #4916, opened 2023-10-09
Fix to write vtks on report steps

daavid00:
If `--enable-vtk-output=true` and `--enable-write-all-solutions=false` (it is false as default), then the VTK files are written for all solutions. This commit fixes that, i.e., the VTK files are written for the report steps (unless we set `--enable-write-all-solutions=true`).

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -737,7 +737,9 @@ public:
         OPM_TIMEBLOCK(problemWriteOutput);
         // use the generic code to prepare the output fields and to
         // write the desired VTK files.
-        ParentType::writeOutput(verbose);
+        if (EWOMS_GET_PARAM(TypeTag, bool, EnableWriteAllSolutions) || this->simulator().episodeWillBeOver()){
+            ParentType::writeOutput(verbose);
+        }
 
         bool isSubStep = !EWOMS_GET_PARAM(TypeTag, bool, EnableWriteAllSolutions) && !this->simulator().episodeWillBeOver();
         


blattms:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -735,12 +735,15 @@ public:
     void writeOutput(bool verbose = true)
     {
         OPM_TIMEBLOCK(problemWriteOutput);
+
+        bool isSubStep = !EWOMS_GET_PARAM(TypeTag, bool, EnableWriteAllSolutions) && !this->simulator().episodeWillBeOver();
```
blattms, line 739: I think we need to used a different variable for the decision.

Changing isSubStep_ here will also influence how results are written with Damaris and for Eclipse. See the lines below:
```
damarisWriter_->writeOutput(localCellData, isSubStep) ;
eclWriter_->writeOutput(std::move(localCellData), isSubStep);
``

Comment (akva2) [reply ebos/eclproblem.hh:739]: afaict there is no change to the variable, it is only moved further up right? i think the name is highly misleading however.

Comment (daavid00) [reply ebos/eclproblem.hh:739]: > afaict there is no change to the variable, it is only moved further up right? i think the name is highly misleading however.

That is right, there is no change to the `isSubStep` variable from the original code, it is only moved before the VTK output call.  We could change to:
```
if (EWOMS_GET_PARAM(TypeTag, bool, EnableWriteAllSolutions) || this->simulator().episodeWillBeOver()){
            ParentType::writeOutput(verbose);
        }
```
if this is preferred.

blattms: Sorry, then it was me being dense. No need to change.

daavid00: I updated the code to the above form (and right after I saw your response @blattms, sorry for that).

blattms:
jenkins build this please

ebos/eclproblem.hh
blattms: No worries and  thanks.

PR #4919, opened 2023-10-09
Multiply well trans due to perm reduction in salt prec

daavid00:
This commit includes the reduction on the well connection transmissibilities due to salt precipitation. This is expected to change the current solutions for the cases in opm-tests including salt precipitation, i.e., the  [SPE1CASE_PRECSALT.DATA](https://github.com/OPM/opm-tests/blob/master/spe1_precsalt/SPE1CASE1_PRECSALT.DATA).

SPE1CASE1_PRECSALT_salt_production_total

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1666,6 +1666,40 @@ public:
         return this->rockCompTransMultWc_[tableIdx].eval(effectiveOilPressure, SwDeltaMax, /*extrapolation=*/true);
     }
 
+    /*!
+     * \brief Calculate the transmissibility multiplier due to porosity reduction.
+     *
+     * TODO: The API of this is a bit ad-hoc, it would be better to use context objects.
+     */
+    template <class LhsEval>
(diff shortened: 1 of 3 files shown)


ebos/eclgenericproblem.hh
```diff
@@ -352,6 +353,7 @@ protected:
     std::vector<TabulatedTwoDFunction> rockCompTransMultWc_;
     std::vector<TabulatedFunction> rockCompPoroMult_;
     std::vector<TabulatedFunction> rockCompTransMult_;
+    std::vector<TabulatedFunction> permFactTransMult_;
```
totto82, line 357: The permFactTransMult table is already in the Brine Module. Better to use that instead of creating a copy.

totto82 (commented):
Thanks. This can be merged when comments are addressed.

daavid00:
Many thanks @totto82 for your comments. I have updated the code.

totto82:
jenkins build this please

totto82:
jenkins build this please

PR #4920, opened 2023-10-10
Casting to support {fmt} v10.

kjetilly:
Due to issues with anonymous enums, newer versions of {fmt} (v10) do not seem to handle the implicit conversion from enum value to `int`. Hence we need a cast for `fmt::format`-calls involving `Matrix::block_matrix::rows` (or `cols`) to compile. This seems to only be relevant for one part of the code. 

For an isolated example, see https://github.com/kjetilly/fmt_fails_with_enum


kjetilly:
Jenkins build this please

bska (approved):

PR #4921, opened 2023-10-10
Don't Apply Regional Multipliers Twice to EDITNNC Connections

bska:
The connections entered in the `EDITNNC` keyword are already fully accounted for in the main 'update()' loop followed by the helper function `applyEditNncToGridTrans_()`.  Therefore, it is wrong to include those connections when applying `MULTREGT` to NNCs.  This commit removes the pertinent connections from the main loop in helper function `applyNncMultreg_()`.

Commit ddcafa8a91 (PR #4822) got this wrong, so pointy hat to [at]bska.


bska:
jenkins build this please

blattms:
For a minute I thought that this was probably me. :sweat_smile: 
Thanks for finding this.

PR #4923, opened 2023-10-11
distinguishing report step and time step

GitPaean:
for the functions related to createWellContainer() and runWellPIScaling().

So it is not exhaustive, but we should do it gradually.

got confused when working with the function createWellContainer(). 

It should NOT change any running results.


GitPaean:
jenkins build this please

GitPaean:
I consider this is not controversial. But still good to have someone have a quick look.

PR #4928, opened 2023-10-13
[testing purpose] adding some output for testing defferred_logger

GitPaean:


GitPaean:
It is an old issue I am not working on, but it was to show that some messages from the deferred_logger gets lost.

image

The left is from the std::cout output, the right is the related message from the DBG file. (I grep "THE FUNCTION" to extract them. )

test case is NETWORK-01.DATA . 

Just posting here in case picking it up at a later stage.

blattms:
Maybe they are there but in  a  different order?

GitPaean:
> Maybe they are there but in a different order?

I hope not. because it was used to check the function calling sequence which is for debugging bugs.

PR #1064, opened 2017-02-17
when no THP control/limit, thp value will be set to zero.

GitPaean:
It provides a strategy to reset the zero thp value when no thp control is involved anymore.

It should not change any running results of all the current examples, while trying to avoid outputting distracting thp values. 

Compared with OPM/opm-core#1143, it provides a strategy to recover the zero THP output when no THP control/limit is involved anymore.  Maybe it should just go ahead to the end of `init()` function.

PR #4930, opened 2023-10-16
not setting THP to be zero based whether having THP constraint from the Deck input in the WellState

GitPaean:
instead, we set the thp to be zero based on whether having a valid VFP table. The following is comments/arguments for the change (not well organized.) Basically the current code only works for limited cases due to coincidence (THP and VFP are provided/absent at the same time), while might impact other scenarios.  The code in this PR although not 100% necessary, while it should not do any harm.  

1. in the current form, the WellState is not aware of possible THP constraint from the network, so even it is trying to set the THP to be zero based on the absence of THP constraint, it is not doing with the full information. setting the thp to be zero based on part information in the WellState turns out to be problematic for some network setup. 

2. when provided a VFP table, even there is no THP constraint specified, we need the thp value for output purpose. Setting the THP value to be zero might not do any harm for this situation, but it does not do any good either.
 
3. The current code in the master might impact the simulation when network involved.  Using whether having a valid VFP table is a better/safer criterion. It should not do any bad to any simulations. For example, when network setup is involved, a valid VFP table has to be provided.

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -397,17 +397,11 @@ void WellState::init(const std::vector<double>& cellPressures,
 
                 // Productivity index.
                 new_well.productivity_index = prev_well.productivity_index;
-            }
 
-            // If in the new step, there is no THP related
-            // target/limit anymore, its thp value should be set to
-            // zero.
-            const bool has_thp = well.isInjector()
-                ? well.injectionControls (summary_state).hasControl(Well::InjectorCMode::THP)
-                : well.productionControls(summary_state).hasControl(Well::ProducerCMode::THP);
-
-            if (!has_thp) {
-                new_well.thp = 0;
+                // if there is no valid VFP table associated, we set the THP value to be 0.
+                if (well.vfp_table_number() == 0) {
+                    new_well.thp = 0.;
+                }
             }
         }
     }


GitPaean:
jenkins build this please

GitPaean:
Even without network, the logic is wrong.  When there is no network involved, the logic should be if there is no available VFP table.  because even without THP constraint, we might want to have the THP values for output purpose.

GitPaean:
zero THP will be set during the simulation. Maybe there is no good reason to set zero THP in the WellState when copying at all.  jenkins does show any failures either.  

Maybe simply removing it will  a solution.

GitPaean:
jenkins build this please

bska:
I **think** I've traced the origin of this code to PR #1064, but there have been a number of refactorings, file renames and structure changes since then so I'm not completely confident.  I guess we need to ask if the reasoning behind that change still holds, or if there are other/more guarantees in place now which obviate the current logic.

GitPaean:
I updated the PR code, will update the description later. Will do some tests and report back also.

opm/simulators/wells/WellState.cpp
```diff
@@ -397,17 +397,11 @@ void WellState::init(const std::vector<double>& cellPressures,
...
-            // If in the new step, there is no THP related
-            // target/limit anymore, its thp value should be set to
-            // zero.
-            const bool has_thp = well.isInjector()
-                ? well.injectionControls (summary_state).hasControl(Well::InjectorCMode::THP)
-                : well.productionControls(summary_state).hasControl(Well::ProducerCMode::THP);
 
-            if (!has_thp) {
-                new_well.thp = 0;
+                // if there is VFP table associated we set the THP value to be 0.
```
bska, line 401: Maybe this comment should say

> if there is **no** VPF table ...

instead?

bska (commented):
I think this makes sense.  I only wonder if there's a word missing in the associated comment.

GitPaean:
jenkins build this please

opm/simulators/wells/WellState.cpp
GitPaean: Yes. fixed.

bska (approved):
Excellent.  Thanks a lot for the updates.  I'll merge into master now.

GitPaean:
Thanks for the merging. With more investigation it shows the real benefit of this PR. For the case we are looking at, it has THP constraint from the network, while not under THP control. When setting the thp to be 0 in the WellState, and also since it was not under THP control, the 0-value thp will enter the simulation. The effect will be the well will be switched to THP control and switching to THP control sometimes creating extra difficulties because of the difficulties in solving the well under THP target depending on the actual situation.

PR #4933, opened 2023-10-17
LookUp origin cell indices in eclgenericproblem

aritorto:
To be able to support LGRs, lookup of origin cell is included in EclGenericProblem and EclProblem. This means that some features of a leaf grid view cell are taken from its origin cell, which is either a parent cell or the equivalent (non-refined) cell from level zero (global grid).

### DIFF
--- a/ebos/eclgenericproblem_impl.hh
+++ b/ebos/eclgenericproblem_impl.hh
@@ -169,14 +169,14 @@ readRockParameters_(const std::vector<Scalar>& cellCenterDepths,
         rockTableIdx_.resize(numElem);
         const auto& num = eclState_.fieldProps().get_int(rock_config.rocknum_property());
         for (std::size_t elemIdx = 0; elemIdx < numElem; ++ elemIdx) {
-            rockTableIdx_[elemIdx] = num[this->lookUpData_.getOriginIndex(elemIdx)] - 1;
+            rockTableIdx_[elemIdx] = num[this->getLookUpData(elemIdx)] - 1;
             auto fmtError =
                 [&num,elemIdx,&ijkIndex,&rock_config, this](const char* type, std::size_t size)
                 {
                     return fmt::format("{} table index {} for elem {} read from {}"
                                       " is is out of bounds for number of tables {}",
-                                       type, num[this->lookUpData_.getOriginIndex(elemIdx)],
-                                       ijkIndex(this-> lookUpData_.getOriginIndex(elemIdx)),
+                                       type, num[this-> getLookUpData(elemIdx)],
+                                       ijkIndex(this-> getLookUpData(elemIdx)),
                                        rock_config.rocknum_property(), size);
                 };
             if (!rockCompPoroMult_.empty() &&
@@ -213,7 +213,7 @@ readRockParameters_(const std::vector<Scalar>& cellCenterDepths,
                 tableIdx = rockTableIdx_[elemIdx];
             }
             overburdenPressure_[elemIdx] =
-                overburdenTables[tableIdx].eval(cellCenterDepths[this->lookUpData_.getOriginIndex(elemIdx)], /*extrapolation=*/true);
+                overburdenTables[tableIdx].eval(cellCenterDepths[this->getLookUpData(elemIdx)], /*extrapolation=*/true);
         }
     }
 }
@@ -356,7 +356,7 @@ rockFraction(unsigned elementIdx, unsigned timeIdx) const
     // geometric volume of the element. Note that it can
     // be larger than 1.0 if porevolume multipliers are used
     // to for instance implement larger boundary cells
-    Scalar porosity = poro[this->lookUpData_.getOriginIndex(elementIdx)]; 
+    Scalar porosity = poro[this->getLookUpData(elementIdx)]; 
     return referencePorosity(elementIdx, timeIdx) / porosity * (1 - porosity);
 }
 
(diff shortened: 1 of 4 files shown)


aritorto:

aritorto:

ebos/eclgenericproblem_impl.hh
```diff
@@ -168,14 +169,15 @@ readRockParameters_(const std::vector<Scalar>& cellCenterDepths,
         rockTableIdx_.resize(numElem);
         const auto& num = eclState_.fieldProps().get_int(rock_config.rocknum_property());
         for (std::size_t elemIdx = 0; elemIdx < numElem; ++ elemIdx) {
-            rockTableIdx_[elemIdx] = num[elemIdx] - 1;
+            rockTableIdx_[elemIdx] = num[this->getLookUpData(elemIdx)] - 1;
             auto fmtError =
-                [&num,elemIdx,&ijkIndex,&rock_config](const char* type, std::size_t size)
+                [&num,elemIdx,&ijkIndex,&rock_config, this](const char* type, std::size_t size)
                 {
```
blattms, line 175: Would it be cleaner to introduce a variable with name e.g. `coarseElemIdx`, which holds the result of the lookup? Then that could be captured instead of this in the lambda.

ebos/eclgenericproblem_impl.hh
```diff
@@ -210,7 +212,8 @@ readRockParameters_(const std::vector<Scalar>& cellCenterDepths,
             if (!rockTableIdx_.empty()) {
                 tableIdx = rockTableIdx_[elemIdx];
             }
-            overburdenPressure_[elemIdx] = overburdenTables[tableIdx].eval(cellCenterDepths[elemIdx], /*extrapolation=*/true);
+            overburdenPressure_[elemIdx] =
+                overburdenTables[tableIdx].eval(cellCenterDepths[this->getLookUpData(elemIdx)], /*extrapolation=*/true);
```
blattms, line 216: I think this should not be changed as we probably want the cell center of the cell on the leaf grid.

ebos/eclgenericproblem_impl.hh
```diff
@@ -317,7 +320,7 @@ rockCompressibility(unsigned globalSpaceIdx) const
 
     unsigned tableIdx = 0;
     if (!this->rockTableIdx_.empty()) {
-        tableIdx = this->rockTableIdx_[globalSpaceIdx];
+        tableIdx = this->rockTableIdx_[globalSpaceIdx];
```
blattms, line 323: Please remove trailing whitespace

ebos/eclgenericproblem_impl.hh
```diff
@@ -331,7 +334,7 @@ rockReferencePressure(unsigned globalSpaceIdx) const
...
```
blattms, line 337: We should prevent introducing trailing whitespace. Also below line 359.

ebos/eclproblem.hh
```diff
@@ -2091,13 +2092,19 @@ protected:
...
+                 this->micp_.microbialConcentration[elemIdx] =
+                     eclWriter_->eclOutputModule().getMicrobialConcentration(getLookUpData(elemIdx));
+                 this->micp_.oxygenConcentration[elemIdx] =
+                     eclWriter_->eclOutputModule().getOxygenConcentration(getLookUpData(elemIdx));
+                 this->micp_.ureaConcentration[elemIdx] =
+                     eclWriter_->eclOutputModule().getUreaConcentration(getLookUpData(elemIdx));
+                 this->micp_.biofilmConcentration[elemIdx] =
+                     eclWriter_->eclOutputModule().getBiofilmConcentration(getLookUpData(elemIdx));
+                 this->micp_.calciteConcentration[elemIdx]
+                     = eclWriter_->eclOutputModule().getCalciteConcentration(getLookUpData(elemIdx));
```
blattms, line 2107: Changing this part here might be a bit early. This here is for restart. If you have run a simulation then you can restart at a later point again.  I don't think we have thought about this for LGRs yet. As these concentrations will be calculated on the leaf grid they probably should also be stored for the leaf grid to preserve the full details of the simulation.

Maybe we should  remove these changes and thow an exception (std::invalid_argumet) at the start of the function with a message that refined grids are not supported for restart, if the grid has more than 1 levels.

IMHO we should not change anything here.

blattms (changes requested):
Thanks a lot.

ebos/eclgenericproblem_impl.hh
aritorto: If I understand correctly, `overburdenPressure_` has size `numElem` which is total amount of cells in the leaf grid view. To define the value of `overburdenPressure_` in certain leaf grid view cell with index `elemIdx`,  you look at` overburdenTables[/*tableIdx*/0].eval(cellCenterDepths[/*certain cell index*/], /*extrapolation*/true)`. Now, if these tables store information of the unrefined grid, then shouldn't /*certain cell index*/ be the index of the origin cell of the leaf-grid-view cell? Am I wrong? I do not know (yet) what eval() does though. 

To phrase it differently, if `overburdenTables[0] `stores information of the unrefined grid, does `eval(cellCenterDepths(**leaf-grid-view-cell-index**), true)` give certain level-0-cell-index to access data in the tables?

ebos/eclgenericproblem_impl.hh
aritorto: Thanks, I forgot to do that.

ebos/eclproblem.hh
aritorto: Thanks for the explanation. I add the exception then.

ebos/eclgenericproblem_impl.hh
aritorto: Thanks!

blattms (approved):

blattms:
Perfect. Looks even better now.

PR #4935, opened 2023-10-18
WIP: Enabling automatic choke to model a sub-sea manifold in an extended network

plgbrts:
This PR is work in progress and not ready for merging. Its purpose for now is to share and test.
It is a first step to model a well group in an extended network that acts 
as a so-called sub-sea manifold: the wells of such group operate on a 
common `THP` (equal to the manifold pressure) and such that the 
total (e.g. oil) rate of the wells obeys an imposed constraint set via `GCONPROD`.

To define a network node to be a subsea-manifold, item 3 of `NODEPROP` is set to` YES`.

In the figure below the simulation results of a simple case is shown and compared to a reference simulator. 
In this test case, group `B1` a subsea manifold with two wells `B-1H` and `B-2H`. 
From the figure: 
-`B-1H` and `B-2H `operates on the same `THP` and 
-the group `B1` satisfies the imposed constraint of oil production rate equals 3500 sm3/day.
-results are in agreement with a reference simulator

image

The simple test case is attached:
[CHECK_COMMONTHP_NETWORK_OPM.zip](https://github.com/OPM/opm-simulators/files/13007283/CHECK_COMMONTHP_NETWORK_OPM.zip)

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -466,7 +466,7 @@ template<class Scalar> class WellContributions;
                                               const double dt,
                                               DeferredLogger& local_deferredLogger);
 
-           void computeWellGroupThp(const double dt, DeferredLogger& local_deferredLogger);
+            void computeWellGroupThp(const double dt, DeferredLogger& local_deferredLogger);
 
             /// Update rank's notion of intersecting wells and their
             /// associate solution variables.
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -413,6 +413,7 @@ template<class Scalar> class WellContributions;
             Scalar gravity_{};
             std::vector<Scalar> depth_{};
             bool alternative_well_rate_init_{};
+            double well_group_thp_prev_{0.0};
 
             std::unique_ptr<RateConverterType> rateConverter_{};
             std::map<std::string, std::unique_ptr<AverageRegionalPressureType>> regionalAveragePressureCalculator_{};
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -413,7 +413,7 @@ template<class Scalar> class WellContributions;
             Scalar gravity_{};
             std::vector<Scalar> depth_{};
             bool alternative_well_rate_init_{};
-            double well_group_thp_prev_{0.0};
+            std::optional<double> well_group_thp_calc_;
 
             std::unique_ptr<RateConverterType> rateConverter_{};
             std::map<std::string, std::unique_ptr<AverageRegionalPressureType>> regionalAveragePressureCalculator_{};
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -413,8 +413,7 @@ template<class Scalar> class WellContributions;
             Scalar gravity_{};
             std::vector<Scalar> depth_{};
             bool alternative_well_rate_init_{};
-            std::optional<double> well_group_thp_calc_;
-
+            std::map<std::string, Scalar> well_group_thp_calc_;
             std::unique_ptr<RateConverterType> rateConverter_{};
             std::map<std::string, std::unique_ptr<AverageRegionalPressureType>> regionalAveragePressureCalculator_{};
 
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -465,7 +465,7 @@ template<class Scalar> class WellContributions;
                                               const double dt,
                                               DeferredLogger& local_deferredLogger);
 
-           void computeWellGroupThp(DeferredLogger& local_deferredLogger);
+           void computeWellGroupThp(const double dt, DeferredLogger& local_deferredLogger);
 
             /// Update rank's notion of intersecting wells and their
             /// associate solution variables.
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -465,6 +465,8 @@ template<class Scalar> class WellContributions;
                                               const double dt,
                                               DeferredLogger& local_deferredLogger);
 
+           void computeWellGroupThp(DeferredLogger& local_deferredLogger);
+
             /// Update rank's notion of intersecting wells and their
             /// associate solution variables.
             ///
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1363,7 +1363,6 @@ updateNetworkPressures(const int reportStepIdx)
     }
 
     for (auto& well : well_container_generic_) {
-
         // Producers only, since we so far only support the
         // "extended" network model (properties defined by
         // BRANPROP and NODEPROP) which only applies to producers.
@@ -1380,6 +1379,9 @@ updateNetworkPressures(const int reportStepIdx)
                 if (thp_is_limit) {
                     ws.thp = well->getTHPConstraint(summaryState_);
                 }
+                //PJPE: Set thp of wells belonging to a subsea manifold equal to the node_pressure
+                if (network.node(well->wellEcl().groupName()).as_choke())
+                    ws.thp = new_limit;
             }
         }
     }
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1363,6 +1363,7 @@ updateNetworkPressures(const int reportStepIdx)
     }
 
     for (auto& well : well_container_generic_) {
+
         // Producers only, since we so far only support the
         // "extended" network model (properties defined by
         // BRANPROP and NODEPROP) which only applies to producers.
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1379,9 +1379,6 @@ updateNetworkPressures(const int reportStepIdx)
                 if (thp_is_limit) {
                     ws.thp = well->getTHPConstraint(summaryState_);
                 }
-                //PJPE: Set thp of wells belonging to a subsea manifold equal to the node_pressure
-                if (network.node(well->wellEcl().groupName()).as_choke())
-                    ws.thp = new_limit;
             }
         }
     }
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -49,9 +49,12 @@
 #include <opm/simulators/utils/MPIPacker.hpp>
 #include <opm/simulators/utils/phaseUsageFromDeck.hpp>
 
+<<<<<<< HEAD
 #if COMPILE_BDA_BRIDGE
 #include <opm/simulators/linalg/bda/WellContributions.hpp>
 #endif
+=======
+>>>>>>> f92db77d3 (moved common thp calculation to updateWellControls)
 
 #if HAVE_MPI
 #include <opm/simulators/utils/MPISerializer.hpp>
@@ -1166,8 +1169,6 @@ namespace Opm {
     BlackoilWellModel<TypeTag>::
     updateWellControlsAndNetwork(const bool mandatory_network_balance, const double dt, DeferredLogger& local_deferredLogger)
     {
-        // PJPE: calculate common THP for subsea manifold well group (item 3 of NODEPROP set to YES)
-        computeWellGroupThp(dt,local_deferredLogger);
         // not necessarily that we always need to update once of the network solutions
         bool do_network_update = true;
         bool well_group_control_changed = false;
@@ -1250,7 +1251,7 @@ namespace Opm {
         return {more_network_update, well_group_control_changed};
     }
 
-    //PJPE: This function is to be used for well groups in an extended network that act as a subsea manifold
+    // This function is to be used for well groups in an extended network that act as a subsea manifold
     // The wells of such group should have a common THP and total phase rate(s) obeying (if possible) 
     // the well group constraint set by GCONPROD
     template <typename TypeTag>
@@ -1276,8 +1277,9 @@ namespace Opm {
                 std::optional<Group::ProductionControls> ctrl = group.productionControls(summary_state);
                 const auto cmode = ctrl->cmode;
                 const auto pu = this->phase_usage_;
-                //PJPE: conversion factor res<-> surface rates TODO: to be calculated
+                //PJPE: conversion factor res<-> surface rates TODO: to be calculated 
                 std::vector<double> resv_coeff(pu.num_phases, 1.0);
+                // rateConverter(0, well_.pvtRegionIdx(), group.name(), resv_coeff); // FIPNUM region 0 here, should use FIPNUM from WELSPECS.
                 double gratTargetFromSales = 0.0; //PJPE: check this
                 WellGroupHelpers::TargetCalculator tcalc(cmode, pu, resv_coeff,
                                                          gratTargetFromSales, nodeName, group_state,
@@ -1286,92 +1288,116 @@ namespace Opm {
                 //PJPE: TODO: include efficiency factor
 
                 auto mismatch = [&] (auto well_group_thp) {
-                    // auto well_state_copy = this->wellState();
                     double well_group_rate(0.0);
+                    double rate(0.0);
                     for (auto& well : this->well_container_) {
                         std::string well_name = well->name();
+                        auto& ws = well_state.well(well_name);
                         if (group.hasWell(well_name)) {
+                            rate = -tcalc.calcModeRateFromRates(ws.surface_rates);
+                            std::cout << "(1) well: " << well_name << ", rate: " << rate << ", thp: " << ws.thp << ", Cmode: " <<  ws.production_cmode << std::endl;
                             well->setDynamicThpLimit(well_group_thp);
+
                             const Well& well_ecl = wells_ecl_[well->indexOfWell()];
                             const auto inj_controls = Well::InjectionControls(0);
                             const auto prod_controls = well_ecl.productionControls(summary_state);
-                            well->iterateWellEqWithSwitching(this->ebosSimulator_, dt, inj_controls, prod_controls, well_state, group_state, local_deferredLogger);
-                            // well->updateWellStateWithTHPTargetProd(this->ebosSimulator_, well_state, local_deferredLogger);
-                            auto& ws = well_state.well(well_name);
-                            double rate = -tcalc.calcModeRateFromRates(ws.surface_rates);
-                            std::cout << "well: " << well_name << "rate: " << rate << std::endl;
+                            well->iterateWellEqWithSwitching(this->ebosSimulator_, dt, inj_controls, prod_controls, well_state, group_state, local_deferredLogger,  false, false);
+                            // May be used instead if the wells of the sub-sea manifold has only THP constraints
+                            // well->updateWellStateWithTHPTargetProd(this->ebosSimulator_, well_state, local_deferredLogger); 
+
+                            rate = -tcalc.calcModeRateFromRates(ws.surface_rates);
+                            if (ws.production_cmode == Opm::WellProducerCMode::THP)
+                                ws.thp = well_group_thp;
+
+                            std::cout << "(2) well: " << well_name << ", rate: " << rate << ", thp: " << ws.thp << ", Cmode: " <<  ws.production_cmode << std::endl;
                             well_group_rate += rate;
                         }
                     }
-                    // this->wellState() = well_state;
-                    return well_group_rate - orig_target;
+                    double diff = well_group_rate - orig_target; // (well_group_rate - orig_target)/orig_target;
+                    std::cout << "well_group_thp: " <<  well_group_thp << " mismatch: " << diff << "\n" << std::endl;
+                    return diff;
                 };
+
+                double well_group_rate(0.0);
+                double rate(0.0);
+                double thp(0.0);
+                double min_thp(1.0E8);
+                double max_thp(0.0);
+                for (auto& well : this->well_container_) {
+                    std::string well_name = well->name();
+                    if (group.hasWell(well_name)) {
+                        auto& ws = well_state.well(well_name);
+                        rate = -tcalc.calcModeRateFromRates(ws.surface_rates);
+                        well_group_rate += rate;
+                        thp = ws.thp;
+                        min_thp = std::min(min_thp, thp);
+                        max_thp = std::max(max_thp, thp);
+                    }
+                }
 
                 const auto upbranch = network.uptree_branch(nodeName);
                 const auto it = node_pressures_.find((*upbranch).uptree_node());
                 const double nodal_pressure = it->second;
-                const std::array<double, 2> range {1.0E5, 150.0E5}; //PJPE what lower/upper bound to be taken here?
-                double low, high;
-        low  = 3527000.0;
-        high = 3576666.7;
-        std::cout << "low: " << low << " high: " << high << std::endl;
-        double low_v = mismatch(low);
-        std::cout << "1: low_value: " << low_v << std::endl;
-        low_v = mismatch(low);
-        high_v = mismatch(high);
-        std::cout << "2: low_value: " << low_v << " high_value: " << high_v << std::endl;
+                double well_group_thp = nodal_pressure;
 
+                if ((reportStepIdx <= 1) || (well_group_rate > 0.9*orig_target)) {
+                    std::array<double, 2> range; // = {1.0E5, 150.0E5}; //PJPE what lower/upper bound to be taken here?
+                    if (well_group_thp_prev_ == 0) {
+                        range = {0.9*min_thp, 1.1*max_thp};
+                    }
+                    else {
+                        range =  {0.9* this->well_group_thp_prev_, 1.1*this->well_group_thp_prev_}; 
+                    }
 
-                local_deferredLogger.debug(" Trying the brute force search to bracket the common THP ");
-                const bool finding_bracket = WellBhpThpCalculator::bruteForceBracket(mismatch, range, low, high, local_deferredLogger);
-                std::cout << "low: " << low << " high: " << high << std::endl;
-                double low_value = mismatch(low);
-                double high_value = mismatch(high);
-                std::cout << "low_value: " << mismatch(low) << " high_value: " << mismatch(high) << std::endl;
+                    double low, high;
+                    // For testing
+                    // low =  3.42767e+06;// 3.32833e+06;
+                    // // high =  3.42767e+06;// 3.32833e+06;
+                    // // std::cout << "low: " << low << " high: " << high << std::endl;
+                    // double low_value_test = mismatch(low);
+                    // low_value_test = mismatch(low);
+                    // low_value_test = mismatch(low);
+                    // std::cout << "low_value: " << low_value_test  << " high_value: " << high_value_test << std::endl;
 
-                double well_group_thp = 0.0;
-                if ((low_value * high_value >= 0.0)  && (low >= nodal_pressure) ) {
-                    // In case no bracket is found
-                    const double limit = 0.0003; // PJPE what limit to be used?
-                    double abs_low = std::fabs(low_value);
-                    double abs_high = std::fabs(high_value);
-                    if (std::min(abs_low, abs_high) < limit) {
-                        well_group_thp = abs_low < abs_high ? low : high; 
-                        local_deferredLogger.warning("Common THP not solved precisely");
+                    std::optional<double> approximate_solution;
+                    const double limit = 1.0E-3;// 1.0E-4; 
+                    local_deferredLogger.debug(" Using brute force search to bracket the common THP ");
+                    std::cout << " Using brute force search to bracket the common THP " << std::endl;
+                    const bool finding_bracket = WellBhpThpCalculator::bruteForceBracketCommonTHP(mismatch, range, low, high, approximate_solution, limit, local_deferredLogger);
+
+                    double low_value = mismatch(low);
+                    double high_value = mismatch(high);
+
+                    if (approximate_solution.has_value()) {
+                        well_group_thp = *approximate_solution;
+                        double check = mismatch(well_group_thp);
+                        local_deferredLogger.debug("Approximate value found: mismatch = " +  std::to_string(check) + ", well_group_thp = " + std::to_string(well_group_thp));
+                    } else if (finding_bracket) {
+                        // PJPE: TODO what settings to take here?
+                        std::cout << "Found bracket: [" << low << ", " << high << "]; mismatch: [" << low_value << ", "<< high_value << "]" << std::endl;
+                        const double tolerance = 1.0E-3;
+                        const int max_iteration_solve = 100;
+                        int iteration = 0;
+                        well_group_thp = RegulaFalsiBisection<ThrowOnError>::
+                                            solve(mismatch, low, high, max_iteration_solve, tolerance, iteration);
+                        local_deferredLogger.debug( " bracket = [" + std::to_string(low) + ", " + std::to_string(high) + "], " +
+                                                      "mismatch = [" + std::to_string(low_value) + ", " + std::to_string(high_value) + "], " +
+                                                      "iteration = " + std::to_string(iteration));
+                        double check = mismatch(well_group_thp);
+                        local_deferredLogger.debug(" mismatch = " +  std::to_string(check) + ", well_group_thp = " + std::to_string(well_group_thp));
                     } else {
                         local_deferredLogger.warning("Common THP solve failed due to bracketing failure");
                         well_group_thp = nodal_pressure;
                     }
-                } else if (high <= nodal_pressure) {
-                      local_deferredLogger.warning("Common THP set to nodal pressure");
-                      well_group_thp = nodal_pressure;
+
+                    std::cout << "well_group_thp = " << well_group_thp << ", nodal_pressure = " << nodal_pressure << std::endl;
+
                 } else {
-                    // PJPE: TODO what settings to take here?
-                    const double tolerance = 1E-4;
-                    const int max_iteration_solve = 20;
-                    int iteration = 0;
-                    well_group_thp = RegulaFalsiBisection<ThrowOnError>::
-                                        solve(mismatch, low, high, max_iteration_solve, tolerance, iteration);
-                    std::cout << "Bracket = [" << low << ", " << high <<"] " << ", iteration = " << iteration << std::endl;
-                    double check = mismatch(well_group_thp);
-                    std::cout << "mismatch = " << check << ", well_group_thp = " << well_group_thp << std::endl;
+                    well_group_thp = nodal_pressure;
                 }
 
-                // well_group_thp = std::max(well_group_thp, nodal_pressure);
-                std::cout << "well_group_thp = " << well_group_thp << ", nodal_pressure = " << nodal_pressure << std::endl;
+                well_group_thp = std::max(well_group_thp, nodal_pressure);
 
-                // if (well_group_thp < 2.5E6){
-                //     double min = 24.0E5;
-                //     double max = 34.0E5;
-                //     double pr(0);
-                //     for (int i = 0; i < 1000; ++i){
-                //         // pr = (i/10.0 + 1) * 1.0E5 ;
-                //         pr = min + i * (max-min)/1000.0;
-                //         // std::cout << " thp = " << pr[i] << ", mismatch = " << mismatch(pr[i]) << std::endl;
-                //         std::cout <<  pr << ", " << mismatch(pr) << std::endl;
-                //     }
-                //     return;
-                // }
                 for (auto& well : this->well_container_) {
                     std::string well_name = well->name();
                     if (group.hasWell(well_name)) {
@@ -1380,16 +1406,19 @@ namespace Opm {
                         if (ws.production_cmode == Opm::WellProducerCMode::THP){
                             well->setDynamicThpLimit(well_group_thp);
                             ws.thp = well_group_thp;
-                            const Well& well_ecl = wells_ecl_[well->indexOfWell()];
-                            const auto inj_controls = Well::InjectionControls(0);
-                            const auto prod_controls = well_ecl.productionControls(summary_state);
-                            // well->updateWellStateWithTHPTargetProd(this->ebosSimulator_, well_state, local_deferredLogger);
-                            well->iterateWellEqWithSwitching(this->ebosSimulator_, dt, inj_controls, prod_controls, well_state, group_state, local_deferredLogger);
                         }
+                        // const Well& well_ecl = wells_ecl_[well->indexOfWell()];
+                        // const auto inj_controls = Well::InjectionControls(0);
+                        // const auto prod_controls = well_ecl.productionControls(summary_state);
+                        // // well->updateWellStateWithTHPTargetProd(this->ebosSimulator_, well_state, local_deferredLogger);
+                        // well->iterateWellEqWithSwitching(this->ebosSimulator_, dt, inj_controls, prod_controls, well_state_copy, group_state, local_deferredLogger);
+                        if (ws.production_cmode == Opm::WellProducerCMode::THP)
+                                ws.thp = well_group_thp;
                     }
                 }
-               
-                //PJPE: use the common THP in computeNetworkPressures
+                well_group_thp_prev_ = well_group_thp;
+
+                // Use the common THP in computeNetworkPressures
                 group_state.update_well_group_thp(nodeName, well_group_thp);
             }
         }
@@ -2055,6 +2084,9 @@ namespace Opm {
         // network related
         bool more_network_update = false;
         if (this->shouldBalanceNetwork(episodeIdx, iterationIdx) || mandatory_network_balance) {
+            const double dt = this->ebosSimulator_.timeStepSize();
+            // Calculate common THP for subsea manifold well group (item 3 of NODEPROP set to YES)
+            computeWellGroupThp(dt, deferred_logger);
             const auto local_network_imbalance = this->updateNetworkPressures(episodeIdx);
             const Scalar network_imbalance = comm.max(local_network_imbalance);
             const auto& balance = this->schedule()[episodeIdx].network_balance();
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1253,7 +1253,6 @@ namespace Opm {
     //PJPE: This function is to be used for well groups in an extended network that act as a subsea manifold
     // The wells of such group should have a common THP and total phase rate(s) obeying (if possible) 
     // the well group constraint set by GCONPROD
-    // subsea manifolds 
     template <typename TypeTag>
     void
     BlackoilWellModel<TypeTag>::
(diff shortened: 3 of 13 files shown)


plgbrts:
jenkins build this please

GitPaean:
Thanks for the good result. I might leave some comments for discussion, not necessarily suggesting code change yet.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1124,6 +1123,9 @@ updateNetworkPressures(const int reportStepIdx)
                 if (thp_is_limit) {
                     ws.thp = well->getTHPConstraint(summaryState_);
                 }
+                //PJPE: Set thp of wells belonging to a subsea manifold equal to the node_pressure
```
GitPaean, line 1126: Here, I believe the `dynamic_thp_limit_` is already updated with `well->setDynamicThpLimit(new_limit);` , so this code should not be necessary.  But  as communicated in the private email,  because of choke pressure, the 
`dynamic_thp_limit_` might not be the same with the nodal pressure of the node anymore.   It might be the nodal_pressure + choke_pressure.

opm/simulators/wells/WellAssemble.cpp
```diff
@@ -161,7 +162,15 @@ assembleControlEqProd(const WellState& well_state,
                 well_.rateConverter().calcCoeff(id, region, coeff);
 
         };
-        WellGroupControls(well_).getGroupProductionControl(group, well_state,
+      
+        bool has_choke(false);
+        const std::size_t report_step = well_.currentStep();
```
GitPaean, line 167: As mentioned in the private message. We might not want to handle the control equation assembly here.  We might want to treat the well under choke pressure control as a normal THP control.

And instead, when the group node is a choke (the group that gives the rate target), we might not have the GRUP control from that group node. 

Some work when specifying the control types might need to be done. (not sure whether from the parser side yet. )

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -897,7 +900,12 @@ namespace WellGroupHelpers
 #endif
                 } else {
                     // Table number specified as 9999 in the deck, no pressure loss.
-                    node_pressures[node] = up_press;
+                    if (network.node(node).as_choke()){
+                        // PJPE: Node pressure is set to the common THP of the wells
+                        node_pressures[node] = group_state.well_group_thp(node);
```
GitPaean, line 905: as mentioned in the private communication, the nodal pressure might not be the common THP of the wells. The nodal pressure is still calculated from the branch VFP calculation. We **might** need a variable to store the choke pressure.  Or the difference between the nodal pressure and the common THP constraints will be the choke pressure, then we do not need to store it explicitly.

GitPaean:
Thanks for the good work. It helps us to understand the auto-choke much better.  Will come with concrete suggestion about the code.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1175,7 +1181,82 @@ namespace Opm {
...
+                            well->updateWellStateWithTHPTargetProd(this->ebosSimulator_, well_state, local_deferredLogger);
+                            auto& ws = well_state.well(well_name);
+                            well_group_rate += -tcalc.calcModeRateFromRates(ws.surface_rates);
+                        }
+                    }
+                    return well_group_rate - orig_target;
+                };
+
+                // PJPE: TODO what settings to take here?
+                const double tolerance = 1E-4;
```
GitPaean, line 1234: tolerance should probably reflect https://github.com/OPM/opm-common/blob/a5fb8be6835708c8de1b39d76e992e06a526b52b/opm/input/eclipse/Schedule/Network/Balance.hpp#L58C32-L58C32

which is defined as a percentage of the group rate target.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1106,6 +1109,9 @@ namespace Opm {
     BlackoilWellModel<TypeTag>::
     updateWellControlsAndNetwork(const bool mandatory_network_balance, const double dt, DeferredLogger& local_deferredLogger)
     {
+        // PJPE: calculate common THP for subsea manifold well group (item 3 of NODEPROP set to YES)
+        computeWellGroupThp(local_deferredLogger);
```
GitPaean, line 1113: I can be wrong in the long term. Ideally, this function should be called within the function `WellGroupHelpers::computeNetworkPressures()`. 

Can be the same location **near** where the following code lies, 

```c++
                    if (network.node(node).as_choke()){
                        // PJPE: Node pressure is set to the common THP of the wells
                        node_pressures[node] = group_state.well_group_thp(node);
```

opm/simulators/wells/WellAssemble.cpp
plgbrts: As you observed the `WMCTL`  (Mode of Control) was not in agreement with the reference simulation. I  changed WellInterface_impl.hpp so that a well under `GRUP`control and with auto-choke its `production_cmode` is set to `Well::ProducerCMode::THP` and the handling of the control equation  for this case (GRUP control and with auto-choke) removed from  WellAssemble.cpp as it is not needed anymore as the well is now treated with `THP` control.

The figure below sows that B-1H is under THP control (`WMCTL=6`)
image

plgbrts:
The test case [5_NETWORK_MODEL5_STDW_AUTOCHK_6000_CHKNODE_TEST1_OPM.zip](https://github.com/OPM/opm-simulators/files/14010376/5_NETWORK_MODEL5_STDW_AUTOCHK_6000_CHKNODE_TEST1_OPM.zip) considered below has a subsea manifold (auto choke group) with 3 wells (`B-1H, B-2H, B-3H`). The group constraint is `ORAT=6000 SM3/DAY` and the wells have individual well constraints of `ORAT=2500 SM3/DAY`. 

Simulation with Flow:
image

Simulation with reference simulator:
image

Some observations:
- Initial all 3 wells of the subsea manifold group run on common `THP` (control mode=6 for the 3 wells) and deliver the group oil rate target  of `6000 SM3/DAY`
- When well `B-2H` meets its individual oil rate constraint of `2500 SM3/DAY`) (control mode=1 for `B-2H`) it runs on this constraint and the `THP` is derived independently of the other two wells that remain on `THP` control with common `THP` value.
- As long as the common `THP` stays above the group nodal pressure the group target of `6000` can be met. When the common `THP` calculation gets below the nodal pressure value the well will operate on the nodal pressure and the target cannot be satisfied 
- Note that for the reference simulation the well `B-1H` is shut-in at some moment while in OPM-flow the well stays open.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
plgbrts: > so this code should not be necessary

Is removed

> `dynamic_thp_limit_` might not be the same with the nodal pressure of the node anymore

True, see my comment below.

opm/simulators/wells/BlackoilWellModel_impl.hpp
plgbrts: For now I moved it deeper and placed it in `updateWellControls` next to the call to `computeNetworkPressures()` which seems to improve performance.

GitPaean:
jenkins build this please

plgbrts:
jenkins build this please

plgbrts:
jenkins build this please

plgbrts:
jenkins build this please

GitPaean:
jenkins build this please

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1187,7 +1187,6 @@ updateNetworkPressures(const int reportStepIdx)
     }
 
     for (auto& well : well_container_generic_) {
-
```
GitPaean, line 1190: This line was deleted probably unintentional.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -43,6 +48,7 @@
 #include <algorithm>
 #include <iomanip>
 #include <utility>
+#include <iostream>
```
GitPaean, line 51: We should not use std::cout in the final version of the code. You can use OpmLog::info or OpmLog::debug, or you can 
use 
deferred_logger.info() or deferred_logger.debug().

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -708,8 +714,6 @@ namespace Opm {
 
         // check group sales limits at the end of the timestep
         const Group& fieldGroup = schedule_.getGroup("FIELD", reportStepIdx);
-        checkGEconLimits(
```
GitPaean, line 711: Is removing of this line intentional?

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1169,6 +1173,7 @@ namespace Opm {
             if (terminal_output_ && (network_update_iteration == iteration_to_relax) ) {
                 local_deferredLogger.info(" we begin using relaxed tolerance for network update now after " + std::to_string(iteration_to_relax) + " iterations ");
             }
+
```
GitPaean, line 1176: This is an extra line not related to this PR.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2331,10 +2510,10 @@ namespace Opm {
                     const std::string msg = "Compute initial well solution for " + well->name() + " initially failed. Continue with the previous rates";
                     deferred_logger.warning("WELL_INITIAL_SOLVE_FAILED", msg);
                 }
+                // If we're using local well solves that include control switches, they also update
+                // operability, so reset before main iterations begin
+                well->resetWellOperability();
```
GitPaean, line 2515: Is moving this to be inside the ` if (param_.solve_welleq_initially_ && well->isOperableAndSolvable()) {` intentional or it is something `git` misunderstood?

opm/simulators/wells/WellAssemble.cpp
```diff
@@ -162,14 +163,14 @@ assembleControlEqProd(const WellState& well_state,
...
         WellGroupControls(well_).getGroupProductionControl(group, well_state,
-                                                             group_state,
-                                                             schedule,
-                                                             summaryState,
-                                                             bhp, active_rates,
-                                                             rCoeff,
-                                                             efficiencyFactor,
-                                                             control_eq,
-                                                             deferred_logger);
+                                                        group_state,
```
GitPaean, line 166: This changes are from the extra indentations. I think they are not intentional.

opm/simulators/wells/WellBhpThpCalculator.hpp
```diff
@@ -149,19 +163,13 @@ class WellBhpThpCalculator {
     //! \brief Get pressure adjustment to the bhp calculated from VFP table
     double getVfpBhpAdjustment(const double bph_tab, const double thp_limit) const;
 
-    //! \brief Find limits using bisection.
```
GitPaean, line 152: This is an unintentional indentation change.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -40,6 +40,9 @@
 #include <opm/simulators/wells/TargetCalculator.hpp>
 #include <opm/simulators/wells/VFPProdProperties.hpp>
 #include <opm/simulators/wells/WellState.hpp>
+#include <opm/simulators/wells/GroupState.hpp>
+// #include <opm/simulators/wells/WellInterface.hpp>
```
GitPaean, line 44: Please remove the commented line if they are not needed.

opm/simulators/wells/WellInterface.hpp
```diff
@@ -432,13 +439,6 @@ class WellInterface : public WellInterfaceIndices<GetPropType<TypeTag, Propertie
                               const GroupState& group_state,
                               DeferredLogger& deferred_logger);
 
-    bool solveWellWithTHPConstraint(const Simulator& ebos_simulator,
```
GitPaean, line 435: not sure whether this needs to be put in `public`.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1363,12 +1363,12 @@ namespace Opm
                 const auto& group = schedule.getGroup(well.groupName(), this->currentStep());
                 const double efficiencyFactor = well.getEfficiencyFactor();
                 double scale = this->getGroupProductionTargetRate(group,
-                                                          well_state,
-                                                          group_state,
-                                                          schedule,
-                                                          summaryState,
-                                                          efficiencyFactor,
-                                                          deferred_logger);
+                                                        well_state,
```
GitPaean, line 1366: Unnecessary indentation change.  And also the few lines below.

GitPaean (changes requested):
I have a few comments about unnecessary changes.  They did not change the code, while due to various reasons, they are shown as committed changes. It is good to clean them up.
I just commented out for cleaning up. The reviewing of the actual content will also come with other comments.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -903,7 +906,20 @@ namespace WellGroupHelpers
...
                     } else {
                         // Table number specified as 9999 in the deck, no pressure loss.
-                        node_pressures[node] = up_press;
+                        if (network.node(node).as_choke()){
+                            // Node pressure is set to the common THP of the wells
+                            const auto up_branch = network.uptree_branch(node);
+                            assert(up_branch);
+                            const auto up_pressure = node_pressures[(*up_branch).uptree_node()];
+                            const auto group_thp = group_state.well_group_thp(node);
+                            if (group_thp >= up_pressure){
```
GitPaean, line 915: This can be simplified as 
```c++
node_pressures[node] = group_thp >= up_pressure ? group_thp : up_press;
```

A comment here to tell for a production network, the choke pressure can not be smaller than zero, so group_thp can not be lower than up_press. Something similar to that.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1888,6 +2064,9 @@ namespace Opm {
         // network related
         bool more_network_update = false;
         if (shouldBalanceNetwork(episodeIdx, iterationIdx) || mandatory_network_balance) {
+            const double dt = this->ebosSimulator_.timeStepSize();
+            // Calculate common THP for subsea manifold well group (item 3 of NODEPROP set to YES)
+            computeWellGroupThp(dt, deferred_logger);
```
GitPaean, line 2069: I believe `computeWellGroupThp()` should be called within the function. Is there some practical reasons we have to call them in this file? for example, the need to access the simulator?

GitPaean (changes requested):
I am in the middle of the reviewing process.  Now the part that I have not checked in details is the function `computeWellGroupThp`. Will provide reviewing comments in another run.

plgbrts:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
plgbrts: Removed all

opm/simulators/wells/BlackoilWellModel_impl.hpp
plgbrts: Nope, it is restored now.

opm/simulators/wells/BlackoilWellModel_impl.hpp
plgbrts: Done

opm/simulators/wells/BlackoilWellModel_impl.hpp
plgbrts: This part of the code I intend not to change at all. I assume it is me to blame rather than git :) . It is fixed now.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -903,7 +906,20 @@ namespace WellGroupHelpers
     #endif
                     } else {
                         // Table number specified as 9999 in the deck, no pressure loss.
-                        node_pressures[node] = up_press;
+                        if (network.node(node).as_choke()){
+                            // Node pressure is set to the common THP of the wells
+                            const auto up_branch = network.uptree_branch(node);
+                            assert(up_branch);
```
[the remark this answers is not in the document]
plgbrts: I will keep that in mind, thanks. I realized I could simply use an earlier  up_branch definition so the up_branch definition here is redundant as well as the  assert so it  is removed.

opm/simulators/wells/WellGroupHelpers.cpp
plgbrts: That is indeed much neater. I changed it according to your suggestion and added a comment.

GitPaean:
Thanks for the update. I am looking into the function `computeWellGroupThp` now.  Hopefully, we are close to get it in after that.

opm/simulators/wells/BlackoilWellModel_impl.hpp
plgbrts: I recall I had a look  at this before and run into issues as the one you mentioned. I can have second try and investigate it. Maybe we could have a chat about this?

GitPaean: Sure. I think some refactoring can be done here. Let me review it first.

GitPaean:
Hei, @plgbrts , I believe the tests you are testing is the case opm-tests/network/NETWORK_MODEL5_STDW_AUTOCHK.DATA
 in opm-tests.  

Can you add a regression test?

plgbrts:
> Hei, @plgbrts , I believe the tests you are testing is the case opm-tests/network/NETWORK_MODEL5_STDW_AUTOCHK.DATA in opm-tests.
> Can you add a regression test?
That is indeed the example case but not yet fit for regression test as the simulation takes too long time. I will have a look into this.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1229,7 +1234,144 @@ namespace Opm {
...
+
+                    std::optional<double> approximate_solution;
+                    const double limit = 1.0E-3;
+                    local_deferredLogger.debug("Using brute force search to bracket the common THP");
+                    const bool finding_bracket = WellBhpThpCalculator::bruteForceBracketCommonTHP(mismatch, range, low, high, approximate_solution, limit, local_deferredLogger);
+
+
+                    if (approximate_solution.has_value()) {
+                        well_group_thp = *approximate_solution;
+                        double check = mismatch(well_group_thp);
```
GitPaean, line 1334: function mismatch() involves a solution for the well and it is good to avoid run it again.  You can get the value from the previous call of function
`bruteForceBracketCommonTHP()` since it is calculated there already.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1229,7 +1234,144 @@ namespace Opm {
...
+        for (const std::string& nodeName : network.node_names()) {
+            const bool has_choke = network.node(nodeName).as_choke();
+            if (has_choke) {
+                const auto& summary_state = this->ebosSimulator_.vanguard().summaryState();
+                const Group& group = schedule().getGroup(nodeName, reportStepIdx);
+                std::optional<Group::ProductionControls> ctrl = group.productionControls(summary_state);
+                const auto cmode = ctrl->cmode;
+                const auto pu = this->phase_usage_;
+                std::vector<double> resv_coeff(pu.num_phases, 1.0);
+                double gratTargetFromSales = 0.0;
```
GitPaean, line 1264: This is something I am not familiar, can we do the following instead of keeping a 0 there?

```c++
451     if (group_state.has_grat_sales_target(group.name())) {
452         gratTargetFromSales = group_state.grat_sales_target(group.name());
          }
 
```

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1229,7 +1234,144 @@ namespace Opm {
...
+                        max_thp = std::max(max_thp, thp);
+                    }
+                }
 
+                const auto upbranch = network.uptree_branch(nodeName);
+                const auto it = node_pressures_.find((*upbranch).uptree_node());
+                const double nodal_pressure = it->second;
+                double well_group_thp = nodal_pressure;
+
+                if ((reportStepIdx <= 1) || (well_group_rate > 0.9*orig_target)) {
```
GitPaean, line 1315: I think using the `(reportStepIdx <= 1) || (well_group_rate > 0.9*orig_target)` to indicate whether the group can reach target rate is not very intuitive or correct.  It might fix the running of the specific case you are working on, while it might cause problem for other cases in general. 

I think `(reportStepIdx <= 1)` is associated with the calculation of `well_group_rate`, so it is the `(well_group_rate > 0.9*orig_target)` as the main condition here.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1229,7 +1234,144 @@ namespace Opm {
...
+                        const int max_iteration_solve = 100;
+                        int iteration = 0;
+                        well_group_thp = RegulaFalsiBisection<ThrowOnError>::
+                                         solve(mismatch, low, high, max_iteration_solve, tolerance, iteration);
+                        local_deferredLogger.debug(" bracket = [" + std::to_string(low) + ", " + std::to_string(high) + "], " +
+                                                   "iteration = " + std::to_string(iteration));
+                        double check = mismatch(well_group_thp);
+                        local_deferredLogger.debug(" mismatch = " +  std::to_string(check) + ", well_group_thp = " + std::to_string(well_group_thp));
+                    } else {
+                        local_deferredLogger.warning("Common THP solve failed due to bracketing failure");
```
GitPaean, line 1347: You can say here that the upstream nodal pressure is used for this nodal pressure for completeness.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1229,7 +1234,144 @@ namespace Opm {
...
+                }
 
+                const auto upbranch = network.uptree_branch(nodeName);
+                const auto it = node_pressures_.find((*upbranch).uptree_node());
+                const double nodal_pressure = it->second;
+                double well_group_thp = nodal_pressure;
+
+                if ((reportStepIdx <= 1) || (well_group_rate > 0.9*orig_target)) {
+                    std::array<double, 2> range;
+                    if (well_group_thp_prev_ == 0) {
```
GitPaean, line 1317: safe range for the THP is very important, we should discuss regarding the current way of determining the search range for THP.

GitPaean (changes requested):
Hei, I just went through the function `computeWellGroupThp`, and I realized that it is more complicated/tricky than I thought.  There are a few details that we need to discuss to settle. 

At the same time, I left some comments for you to address.

I think the most difficult part can be how to determine the search range for the nodal pressure/ common THP. 

And there are many variables should be `const` if they are not intended to be modified again.

plgbrts:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1229,7 +1234,144 @@ namespace Opm {
...
+
+        auto& well_state = this->wellState();
+        auto& group_state = this->groupState();
+
+        for (const std::string& nodeName : network.node_names()) {
+            const bool has_choke = network.node(nodeName).as_choke();
+            if (has_choke) {
+                const auto& summary_state = this->ebosSimulator_.vanguard().summaryState();
+                const Group& group = schedule().getGroup(nodeName, reportStepIdx);
+                std::optional<Group::ProductionControls> ctrl = group.productionControls(summary_state);
```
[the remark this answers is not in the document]
plgbrts: I think so too, removed it

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1229,7 +1234,144 @@ namespace Opm {
...
+                const auto cmode = ctrl->cmode;
+                const auto pu = this->phase_usage_;
+                std::vector<double> resv_coeff(pu.num_phases, 1.0);
```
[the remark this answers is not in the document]
plgbrts: `RESV` group control combined with auto-choke is indeed not supported at this moment. To use `rateConverter()` seems not so obvious as we have to deal with a group of wells and each well may be associated to a different PVT region. I propose to have a error check in the keyword handling in opm-common for the combination of `RESV` (in `GCONPROD`) and auto-choke option set to `YES` in `NODEPROP`.  (Although it is not yet obvious to me how to do such error handling properly.)

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1229,7 +1234,144 @@ namespace Opm {
...
+        }
+
+        auto& well_state = this->wellState();
+        auto& group_state = this->groupState();
+
+        for (const std::string& nodeName : network.node_names()) {
+            const bool has_choke = network.node(nodeName).as_choke();
+            if (has_choke) {
+                const auto& summary_state = this->ebosSimulator_.vanguard().summaryState();
+                const Group& group = schedule().getGroup(nodeName, reportStepIdx);
```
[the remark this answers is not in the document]
plgbrts: Should such check better be done in `handleNODEPROP`?

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1229,7 +1234,144 @@ namespace Opm {
...
+                double orig_target = tcalc.groupTarget(ctrl, local_deferredLogger);
+
+                auto mismatch = [&] (auto well_group_thp) {
+                    double well_group_rate(0.0);
+                    double rate(0.0);
+                    for (auto& well : this->well_container_) {
+                        std::string well_name = well->name();
+                        auto& ws = well_state.well(well_name);
+                        if (group.hasWell(well_name)) {
+                            rate = -tcalc.calcModeRateFromRates(ws.surface_rates);
```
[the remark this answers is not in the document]
plgbrts: Removed it.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1229,7 +1234,144 @@ namespace Opm {
...
+                const Group& group = schedule().getGroup(nodeName, reportStepIdx);
+                std::optional<Group::ProductionControls> ctrl = group.productionControls(summary_state);
+                const auto cmode = ctrl->cmode;
+                const auto pu = this->phase_usage_;
+                std::vector<double> resv_coeff(pu.num_phases, 1.0);
+                double gratTargetFromSales = 0.0; 
+                WellGroupHelpers::TargetCalculator tcalc(cmode, pu, resv_coeff,
+                                                         gratTargetFromSales, nodeName, group_state,
+                                                         group.has_gpmaint_control(cmode));
+                double orig_target = tcalc.groupTarget(ctrl, local_deferredLogger);
```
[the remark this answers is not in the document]
plgbrts: Done.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1229,7 +1234,144 @@ namespace Opm {
...
+                        auto& ws = well_state.well(well_name);
+                        if (group.hasWell(well_name)) {
+                            rate = -tcalc.calcModeRateFromRates(ws.surface_rates);
+                            well->setDynamicThpLimit(well_group_thp);
+                            const Well& well_ecl = wells_ecl_[well->indexOfWell()];
+                            const auto inj_controls = Well::InjectionControls(0);
+                            const auto prod_controls = well_ecl.productionControls(summary_state);
+                            well->iterateWellEqWithSwitching(this->ebosSimulator_, dt, inj_controls, prod_controls, well_state, group_state, local_deferredLogger,  false, false);
+                            rate = -tcalc.calcModeRateFromRates(ws.surface_rates);
+                            if (ws.production_cmode == Opm::WellProducerCMode::THP)
```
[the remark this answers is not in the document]
plgbrts: It looks indeed not needed and is removed.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1229,7 +1234,144 @@ namespace Opm {
...
+                        range = {0.9*min_thp, 1.1*max_thp};
+                    }
+                    else {
+                        range =  {0.9* this->well_group_thp_prev_, 1.1*this->well_group_thp_prev_}; 
+                    }
+
+                    double low, high;
+
+                    std::optional<double> approximate_solution;
+                    const double limit = 1.0E-3;
```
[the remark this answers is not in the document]
plgbrts: Changed the naming.

opm/simulators/wells/BlackoilWellModel_impl.hpp
plgbrts: The recalculation was used for debugging purposes so I removed it

opm/simulators/wells/BlackoilWellModel_impl.hpp
plgbrts: I think we can. So I put it there (but did not test it yet).

opm/simulators/wells/BlackoilWellModel_impl.hpp
plgbrts: The condition I have changed hopefully in something more sensible. Now checking if the calculated common `THP` is larger than the nodal pressure. (Not sure if this is robust enough in general though.)

opm/simulators/wells/BlackoilWellModel_impl.hpp
plgbrts: I am struggling with this. @GitPaean It would be good to have a chat about this.

plgbrts:
jenkins build this please

plgbrts:
jenkins build this please

plgbrts:
jenkins build this please

plgbrts:
jenkins build this please

GitPaean:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1241,7 +1245,141 @@ namespace Opm {
...
+                    max_thp = high1;
+                    range_initial = {min_thp, max_thp};
+                }
+
+                const auto upbranch = network.uptree_branch(nodeName);
+                const auto it = this->node_pressures_.find((*upbranch).uptree_node());
+                const Scalar nodal_pressure = it->second;
+                Scalar well_group_thp = nodal_pressure;
+
+                if (!this->well_group_thp_calc_.has_value() || this->well_group_thp_calc_ > nodal_pressure) {
```
GitPaean, line 1334: This function is trying to calculate `this->well_group_thp_calc_`. With the condition `this->well_group_thp_calc_ > nodal_pressure`, will it happens that, when ``this->well_group_thp_calc_ > nodal_pressure` happens, we will never update `this->well_group_thp_calc_` anymore? At least, from the code, it can happens. Not sure it is a desired behavoir.

opm/simulators/wells/BlackoilWellModel.hpp
```diff
@@ -398,6 +398,7 @@ class WellContributions;
             Scalar gravity_{};
             std::vector<Scalar> depth_{};
             bool alternative_well_rate_init_{};
+            std::optional<double> well_group_thp_calc_;
```
GitPaean, line 401: I see it is single value for the `BlackoilWellModel`. It might not work for multiple auto-choke nodes. And it might be wrong for parallel running, where each process has its own `well_group_thp_calc_`. 

Should we get this value from GroupState::group_thp?

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1241,7 +1245,141 @@ namespace Opm {
...
+                    // The bracket is based on the initial bracket or on a range based on a previous calculated common group thp
+                    std::array<Scalar, 2> range;
+                    this->well_group_thp_calc_.has_value() ? 
+                        range =  {0.9 * this->well_group_thp_calc_.value(), 1.1 * this->well_group_thp_calc_.value()} :
```
GitPaean, line 1338: This line be problematic for some scenarios. We can discuss to address this one.  Possibly, we can always use `range_initial`.  `range_initial` is already a bracket smaller than 1 bar. Let me know what you think.

opm/simulators/wells/WellBhpThpCalculator.hpp
```diff
@@ -117,7 +117,25 @@ class WellBhpThpCalculator {
   std::pair<Scalar, Scalar>
   getFloIPR(const WellState<Scalar>& well_state,
             const Well& well, 
-            const SummaryState& summary_state) const;                                                                                              
+            const SummaryState& summary_state) const;
+
+  //! \brief Find limits using brute-force solver.
+  static bool bruteForceBracket(const std::function<Scalar(const Scalar)>& eq,
```
GitPaean, line 123: any reason to move this function outside of `private`? I think it might be a fallout during the development process.

GitPaean (changes requested):
Hei, thanks for the update. I have some questions and comments required to be addressed.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -40,7 +41,10 @@
 #include <opm/simulators/wells/ParallelPAvgDynamicSourceData.hpp>
 #include <opm/simulators/wells/ParallelWBPCalculation.hpp>
 #include <opm/simulators/wells/VFPProperties.hpp>
+#include <opm/simulators/wells/WellGroupHelpers.hpp>
```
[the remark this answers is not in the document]
plgbrts: Done

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1241,7 +1245,141 @@ namespace Opm {
...
+                const auto ctrl = group.productionControls(summary_state);
+                const auto cmode = ctrl.cmode;
+                const auto pu = this->phase_usage_;
+                //TODO: Auto choke combined with RESV control is not supported
+                std::vector<Scalar> resv_coeff(pu.num_phases, 1.0);
+                Scalar gratTargetFromSales = 0.0; 
+                if (group_state.has_grat_sales_target(group.name()))
+                    gratTargetFromSales = group_state.grat_sales_target(group.name());
+
+                WGHelpers::TargetCalculator tcalc(cmode, pu, resv_coeff,
```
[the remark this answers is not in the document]
plgbrts: Done

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1241,7 +1245,141 @@ namespace Opm {
...
+                            const auto prod_controls = well_ecl.productionControls(summary_state);
+                            well->iterateWellEqWithSwitching(this->simulator_, dt, inj_controls, prod_controls, well_state, group_state, local_deferredLogger,  false, false);
+                            rate = -tcalc.calcModeRateFromRates(ws.surface_rates);
+                            well_group_rate += rate;
+                        }
+                    }
+                    return (well_group_rate - orig_target)/orig_target;
+                };
 
+                Scalar min_thp, max_thp;
```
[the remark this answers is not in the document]
plgbrts: Done

opm/simulators/wells/WellBhpThpCalculator.cpp
```diff
@@ -1002,6 +1002,70 @@ getFloIPR(const WellState<Scalar>& well_state,
...
+}
+
+template<class Scalar>
+bool
+WellBhpThpCalculator<Scalar>::
+bruteForceBracketCommonTHP(const std::function<Scalar(const Scalar)>& eq,
+                  Scalar& min_thp, Scalar& max_thp)
+{
+    bool bracket_found = false;
+    const int sample_number = 1000;
```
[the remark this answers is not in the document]
plgbrts: Done

opm/simulators/wells/BlackoilWellModel_impl.hpp
plgbrts: I have tried to remove this (`this->well_group_thp_calc_ > nodal_pressure`) condition which goes with additional calculation costs but would be more general. However oscillating group thp pressure results (figure 1).  Tightening the tolerance  (`--tolerance-wells=1E-6`) gives smoother results but shut-in of wells (figure 2).  Compare both figures with the one given earlier (see above). Can we have a chat about what to do here?

figure 1:
image

figure 2:
image

opm/simulators/wells/BlackoilWellModel.hpp
plgbrts: I have tried to use `GroupState::group_thp` as we have discussed but that I could not get to work. Your solution to  use `std::map<std::string, Scalar> well_group_thp_calc_` in `BlackoilWellModel.hpp` is working so I followed that. Many thanks for this!

opm/simulators/wells/BlackoilWellModel_impl.hpp
plgbrts: I understand your concern. However I am failing to make this robust. E.g.  using `range_initial`  did not work. Maybe we can have chat about this.

opm/simulators/wells/WellBhpThpCalculator.hpp
plgbrts: You are right. I placed it back

plgbrts:
jenkins build this please

plgbrts:
jenkins build this please

GitPaean:
Does the jenkins failure indicate there is some problem with the CUDA setup?
```
  #$ "/usr/lib/ccache"/g++-12 -D__CUDA_ARCH__=520 -D__CUDA_ARCH_LIST__=520 -E
  -x c++ -DCUDA_DOUBLE_MATH_FUNCTIONS -D__CUDACC__ -D__NVCC__
  -D__CUDACC_VER_MAJOR__=11 -D__CUDACC_VER_MINOR__=8
  -D__CUDACC_VER_BUILD__=89 -D__CUDA_API_VER_MAJOR__=11
  -D__CUDA_API_VER_MINOR__=8 -D__NVCC_DIAG_PRAGMA_SUPPORT__=1 -include
  "cuda_runtime.h" -m64 "CMakeCUDACompilerId.cu" -o
  "tmp/CMakeCUDACompilerId.cpp1.ii"

  In file included from /usr/include/cuda_runtime.h:83,

                   from <command-line>:

  /usr/include/crt/host_config.h:132:2: error: #error -- unsupported GNU
  version! gcc versions later than 11 are not supported! The nvcc flag
  '-allow-unsupported-compiler' can be used to override this version check;
  however, using an unsupported host compiler may cause compilation failure
  or incorrect run time execution.  Use at your own risk.

    132 | #error -- unsupported GNU version! gcc versions later than 11 are not supported! The nvcc flag '-allow-unsupported-compiler' can be used to override this version check; however, using an unsupported host compiler may cause compilation failure or incorrect run time execution. Use at your own risk.
        |  ^~~~~

  # --error 0x1 --
```
@akva2

akva2:
well, it indicates that people are triggering builds while i'm hard at work updating jenkins. i have not announced that jenkins is ready yet, because it is not ready yet.

GitPaean:
> well, it indicates that people are triggering builds while i'm hard at work updating jenkins. i have not announced that jenkins is ready yet, because it is not ready yet.

Thanks for the hard work.  There is no intention of pushing.

opm/simulators/wells/BlackoilWellModel_impl.hpp
GitPaean: I am just commenting for information.

When the search for `range_initial` works, `range_initial` indeed has proper value and smaller range, 
```
 range_initial 33.781000000000006 33.806666666666672
 range 30.397929434506676 37.153024864397047
 range_initial 33.781000000000006 33.806666666666672
 range 30.397929434506676 37.153024864397047
 range_initial 33.781000000000006 33.806666666666672
```

But sometimes the search for `range_initial` is not always working properly, that will cause problem. 
```
 range_initial 38.500000000000007 38.500000000000007
 range 30.397929434506676 37.153024864397047
 well B-1H gets STOPPED during iteration 
 well B-3H gets STOPPED during iteration 
 range_initial 1123.100000000000364 1123.100000000000364
 range 30.397929434506676 37.153024864397047
 range_initial 1123.100000000000364 1123.100000000000364
 range 30.397929434506676 37.153024864397047
```

opm/simulators/wells/BlackoilWellModel_impl.hpp
GitPaean: Still for information and potential discussion, I think if with tighter tolerance, only well B-1H got shut, which is the behavoir from the reference simulator, I think it is fine.

GitPaean: The oscillation looks like solution is not converged.

GitPaean: But let me check why the well gets SHUT.

GitPaean: The manifold THP can not be lower than the nodal pressure. Ideally, we can check whether we can determine whether we can reach the production target first. If not, then the wells will run under the nodal pressure, so we do not need to look for the manifold pressure. 

If we can reach the production target, then we need to calculate the manifold choke pressure. 

But determining whether we can reach the production target involving solving all the wells under THP constraints might fail for various causes. 

The current approach assumes that we use the existing autochoke_thp (solved previously or equal to the nodal pressure ) to determine whether the group production can be reached. It works when the situation will not sway back and force between being able and not being able to each the group target.  It is the situation for the case the PR has been tested with.  The situations for other cases can be very different.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1241,7 +1245,141 @@ namespace Opm {
...
+                if (!this->well_group_thp_calc_.has_value()){
+                    // Retrieve the terminal pressure of the associated root of the manifold group
+                    std::string node_name =  nodeName;
+                    while (!network.node(node_name).terminal_pressure().has_value()) {
+                        auto branch = network.uptree_branch(node_name).value();
+                        node_name = branch.uptree_node();
+                    }
+
+                    min_thp = network.node(node_name).terminal_pressure().value();
+                    WellBhpThpCalculator<Scalar>::bruteForceBracketCommonTHP(mismatch, min_thp, max_thp);
```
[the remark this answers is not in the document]
GitPaean: This argument is still valid. We should not need to narrow down the range further.

GitPaean:
Thanks for the update.  I tested the code, the results are following what was reported. The bracketing of the thp pressure range, the solve for the manifold THPs and so on, many operations in this PR involves a lot of solving of the VFP table related, which is also prune to failures, which makes the change  or doing the proper logic is difficult because of unexpected failures. 

The PR introduces one way solve the auto-choke pressure and it works for the case it provides.  We can always improve this part with more cases going on.

GitPaean:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1243,7 +1247,140 @@ namespace Opm {
...
+                Scalar gratTargetFromSales = 0.0; 
+                if (group_state.has_grat_sales_target(group.name()))
+                    gratTargetFromSales = group_state.grat_sales_target(group.name());
+
+                WGHelpers::TargetCalculator tcalc(cmode, pu, resv_coeff,
+                                                  gratTargetFromSales, nodeName, group_state,
+                                                  group.has_gpmaint_control(cmode));
+                const Scalar orig_target = tcalc.groupTarget(ctrl, local_deferredLogger);
+
+                auto mismatch = [&] (auto well_group_thp) {
```
GitPaean, line 1289: The name `well_group_thp` probably shadows another `well_group_thp` defined later.

opm/simulators/wells/WellInterfaceGeneric.cpp
```diff
@@ -487,6 +487,13 @@ WellInterfaceGeneric<Scalar>::getDynamicThpLimit() const
     return dynamic_thp_limit_;
 }
 
+template<class Scalar>
+void WellInterfaceGeneric<Scalar>::
+setDynamicThpLimit(const std::optional<Scalar> thp_limit)
```
GitPaean, line 492: This function is node needed, please remove.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -935,7 +936,15 @@ computeNetworkPressures(const Network::ExtNetwork& network,
 #endif
                 } else {
                     // Table number specified as 9999 in the deck, no pressure loss.
-                    node_pressures[node] = up_press;
+                    if (network.node(node).as_choke()){
+                        // Node pressure is set to the common group THP of the wells.
+                        // The choke pressure must be non-negative therefore the node pressure of 
+                        // the auto-choke node must be larger or equal to the pressure of the uptree node of its branch 
+                        const auto group_thp = group_state.well_group_thp(node);
+                        node_pressures[node] = group_thp >= up_press ? group_thp : up_press;
```
GitPaean, line 944: We had some similar logic in the following code, 
```c++
                 if (autochoke_thp.has_value()) {
                    well_group_thp_calc_[nodeName] = autochoke_thp.value();
                    well_group_thp = std::max(autochoke_thp.value(), nodal_pressure);
                }
```
Do we need to enforce this logic again?

GitPaean (changes requested):
Thanks for the update.  Please address the last few comments, and please squash all commits does not have a meaningful commit message, for example, `rebasing`.  Let us know if you need help with that. 

After the few comments gets addressed and jenkins approves, I am ready to get this PR in. 

Thanks for the efforts.

opm/simulators/wells/WellInterfaceGeneric.cpp
plgbrts: Removed it.

opm/simulators/wells/BlackoilWellModel_impl.hpp
plgbrts: It is renamed to `group_thp`

opm/simulators/wells/WellGroupHelpers.cpp
plgbrts: It is indeed redundant. I removed it.

plgbrts:
jenkins build this please

GitPaean (approved):

GitPaean:
Thanks for the update. It provides some design for the auto-choke solution and it works for some cases. We will benefit from this development. As we discussed previously, with more testing cases, it is a field that we need to improve continuously in the future. 

Jenkins approves, I am getting it in now.

plgbrts:
Thanks a lot @GitPaean for all comments, discussions and help!

PR #4936, opened 2023-10-18
removing VtkCompositional related params from FlowMainEbos

GitPaean:
fixing the compilation errors due to the upstream https://github.com/OPM/opm-models/pull/844 .


GitPaean:
I have never run simulation with ebos, so I do not under the potential problems regarding VtkCompostional from the FlowMainEbos.  Maybe people are more aware of the matter can give some comments? @totto82 

The original motivation is that we should not need a lot of compositional related functions in the BlackoilFluidState. And also it becomes problems when working with compositional development.  Hopefully, the compostional and blackoil related can be more separated in a good manner in the future.  But current, the main Flow do not handle compositional related, I think it might be okay. If Flow in the future handle compositional related, then it will be handled as a new issue through new development or refactoring.

PR #4937, opened 2023-10-20
Add output of the residual at the end of a time step

totto82:
To activte the output you will need to use
--enable-opm-rst-file=true
and
RPTRST
 CONV /

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -278,6 +278,7 @@ public:
             return;
 
         const auto& problem = elemCtx.simulator().problem();
+        const auto& ebosResid = elemCtx.simulator().model().linearizer().residual();
         for (unsigned dofIdx = 0; dofIdx < elemCtx.numPrimaryDof(/*timeIdx=*/0); ++dofIdx) {
             const auto& intQuants = elemCtx.intensiveQuantities(dofIdx, /*timeIdx=*/0);
             const auto& fs = intQuants.fluidState();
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

bska:
> ```
> RPTRST
> CONV /
> ```

The mnemonic `CONV` is **usually** reserved for something else&ndash;an integral number of times a cell has been among the "worst" `N` cells at the start of a non-linear/Newton iteration.  Would you mind considering a different mnemonic here, please?

totto82:
Could RESIDUAL work?

bska:
> Could `RESIDUAL` work?

Yes, that's perfect.  You'll probably have to add the mnemonic to list of [known `RPTRST` keys](https://github.com/OPM/opm-common/blob/b3177a95da0b8425b61f0acd9842549727b87909/src/opm/input/eclipse/Schedule/RSTConfig.cpp#L162) to get that through the input layer, but once that's done everything else should "just work" the way you've outlined here.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -130,6 +130,7 @@ class EclOutputBlackOilModule : public EclGenericOutputBlackoilModule<GetPropTyp
 
     enum { conti0EqIdx = Indices::conti0EqIdx };
     enum { numPhases = FluidSystem::numPhases };
+    enum { numComponents = FluidSystem::numComponents };
```
bska, line 133: This value is, as far as I can tell, not used anywhere in this file.  Moreover, we have the same value [in our base class](https://github.com/OPM/opm-simulators/blob/e9645ada241736af52ab5ab97a8a47a6867050e0/ebos/eclgenericoutputblackoilmodule.hh#L280).  Is there a reason you need to introduce it here?

bska (commented):
This looks good for the most part.  I don't immediately see a use for the new `numComponents` value, however, so I'd like you to elaborate a little on that particular topic.  Otherwise I'm prepared to merge this.

totto82:
jenkins build this please

PR #4882, opened 2023-09-20
Add Support for Reading MPI Partitioning From File

bska:
This commit introduces new, experimental support for loading a partitioning of the cells from a text file.  The name of the file is passed into the simulator using the new, hidden, command line option
```
--external-partition=filename
```
and we perform some basic checking that the number of elements in the partition matches the number of cells in the `CpGrid` object.

PR #4938, opened 2023-10-23
Enable Loading Parallel NLDD Partition From File

bska:
This PR adds support for loading a three-column NLDD partitioning scheme of the form
```
MPI_Rank  Cartesian_Index  NLDD_Domain_ID
```
from a text file.  In an MPI run it is assumed that the first column holds integers in the range `0..MPI_Size()-1`, and typically that each such integer is listed at least once.  In a sequential run, the MPI rank column is ignored.

This PR also adds support for outputting such text files in both sequential and MPI runs.  If the user activates the new, and hidden, debugging option
```
DebugEmitCellPartition (--debug-emit-cell-partition)
```
each rank will write one such three-column text file into the special-purpose subdirectory `partition/CaseName` of the run's output directory.  That file will be named according to the process' MPI rank, so the first column will be the same as the file name.

The option is primarily intended for debugging the NLDD partitioning scheme, so is mostly reserved for runs with low MPI sizes (e.g., less than 20).

While here, also make the `MPIPartitionFromFile` helper class of PR #4882 aware of this format so that we can use concatenated output files as an input to the MPI partitioning algorithm for repeatability.

### DIFF
--- a/opm/simulators/flow/partitionCells.cpp
+++ b/opm/simulators/flow/partitionCells.cpp
@@ -23,6 +23,8 @@
 
 #include <opm/simulators/flow/countGlobalCells.hpp>
 
+#include <opm/simulators/utils/compressPartition.hpp>
+
 #if HAVE_MPI && HAVE_ZOLTAN
 #include <opm/simulators/utils/DeferredLogger.hpp>
 #include <opm/simulators/utils/DeferredLoggingErrorHelpers.hpp>
@@ -42,15 +44,20 @@
 #include <dune/alugrid/grid.hh>
 #endif // HAVE_DUNE_ALUGRID
 
+#include <opm/common/utility/CSRGraphFromCoordinates.hpp>
+
 #include <fmt/format.h>
 
 #include <algorithm>
 #include <cassert>
+#include <cstddef>
+#include <filesystem>
 #include <fstream>
 #include <iterator>
 #include <numeric>
 #include <stdexcept>
 #include <string>
+#include <string_view>
 #include <tuple>
 #include <type_traits>
 #include <unordered_map>
@@ -320,7 +327,7 @@ void ZoltanPartitioner::connectWells(const Comm                          comm,
         const auto* pl = (distributedWells == 1) ? "" : "s";
 
         throw std::invalid_argument {
-            fmt::format(R"({} distributed well{} detected on rank {}."
+            fmt::format(R"({} distributed well{} detected on rank {}.
 This is not supported in the current NLDD implementation.)",
                         distributedWells, pl, comm.rank())
         };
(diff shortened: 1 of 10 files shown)


bska:
jenkins build this opm-common=3726 please

opm/simulators/utils/compressPartition.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+#ifndef OPM_UTIL_COMPRESS_PARTITION_HPP_INCLUDED
+#define OPM_UTIL_COMPRESS_PARTITION_HPP_INCLUDED
+
+#include <utility>
+#include <vector>
+
+namespace Opm { namespace util {
+    std::pair<std::vector<int>, int>
+    compressAndCountPartitionIDs(std::vector<int>&& parts0);
```
atgeirr, line 28: Perhaps early to comment as this is still a draft but anyway: I would prefer that these functions took an lvalue reference instead, as that would make it possible to use them in-place. As is, if I have a partition vector `p` I'd like to compress I will have to move it to this function, than afterwards move the result to the original `p`.

bska:
The requisite upstream PRs have both been merged into their respective master branches.  I'm marking this as  "ready for review" now.

bska:
jenkins build this please

opm/simulators/utils/compressPartition.hpp
bska: > I would prefer that these functions took an lvalue reference instead, as that would make it possible to use them in-place.

I don't think it makes sense to have this particular compression function work in-place.  In addition to compressing, this function also counts the number of blocks&ndash;i.e., the number of unique block IDs&ndash;and it returns a pair  of the compressed partition vector and that number of unique block IDs.  We could add an overload of the other function, `void compressPartitionIDs(std::vector<int>& parts)`, though.  I'd rather not have a function which both mutates an in/out parameter **and** returns an integer value.

opm/simulators/flow/partitionCells.cpp
```diff
@@ -357,6 +363,122 @@ partitionCellsZoltan(const int                                      num_domains,
...
+    std::pair<std::vector<int>, int>
+    partitionCellsFromFileMPI(const std::filesystem::path&                   fileName,
+                              const GridView&                                grid_view,
+                              const Opm::ZoltanPartitioningControl<Element>& zoltan_ctrl)
+    {
+        const auto input = (grid_view.comm().rank() == 0)
+            ? integerVectorFromFile(fileName)
+            : std::vector<int>{};
+
+        return Opm::util::compressAndCountPartitionIDs
```
atgeirr, line 475: There should be some more testing for invalid input. I think the `input.size()` should be checked in this scope, and then all individual MPI ranks' sizes verified against the size of the scattered data in scatterPartition().

atgeirr (changes requested):
Just two minor issues.

opm/simulators/flow/partitionCells.cpp
```diff
@@ -357,6 +363,122 @@ partitionCellsZoltan(const int                                      num_domains,
...
+            for (const auto& cell : elements(grid_view, Dune::Partitions::interior)) {
+                const auto globIx = zoltan_ctrl.local_to_global(zoltan_ctrl.index(cell));
+                g2l.insert_or_assign(globIx, localCell++);
+            }
+        }
+
+        const auto myPartitionInput = groupInputByProcessID(rawInput, grid_view.comm());
+
+        auto partition = std::vector<int>(g2l.size(), -1);
+        for (auto i = 0*myPartitionInput.size(); i < myPartitionInput.size(); i += 2) {
```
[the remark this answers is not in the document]
bska: > More readable to increment by 1 and do 2*i inside the loop.

Maybe, but then we can't use `myPartitionInput.size()` as the upper bound.  It would be something like `myPartitionInput.size() / 2` and at that point it's not clear to me which is better.

bska:
I've pushed an update here which adds more input checking.

bska:
jenkins build this please

atgeirr (approved):

atgeirr:
Thanks for the improvements, merging!

PR #4941, opened 2023-10-24
Remove Leftover Debug Printout

bska:
This information already goes to the `.DBG` file so there's no need to echo it to the main console as well.


bska:
jenkins build this please

PR #4943, opened 2023-10-24
Get simulator time step size from Python BlackOilSimulator module

hakonhagland:
Adds a `get_dt()` to the `opm.simulators.BlackOilSimulator` Python module. This will return the size of the previous simulator time step.

### DIFF
--- a/opm/simulators/flow/FlowMainEbos.hpp
+++ b/opm/simulators/flow/FlowMainEbos.hpp
@@ -321,6 +321,12 @@ void handleExtraConvergenceOutput(SimulatorReport& report,
             return simtimer_.get();
         }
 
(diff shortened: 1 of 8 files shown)


hakonhagland:
jenkins build this please

totto82:
Thanks. I think it looks good. I would have preferred if you made the refactoring /alphabetical re-ordering of methods in a separate commit. That would have made the review easier.

hakonhagland:
> I would have preferred if you made the refactoring /alphabetical re-ordering of methods in a separate commit. That would have made the review easier.

@totto82 Agree. This was bad practice by me.  I realized too late that I should have made it into separate commits. Thanks for the review.

PR #4944, opened 2023-10-25
Apply Full SWATINIT-Like PCOW Rescaling at Restart

bska:
We switch to using the `applyRestartSwatInit()` member function (PR OPM/opm-common#3731) instead of directly mutating a data member of the `EpsInfo` structure. This way we defer the rescaling to a context with a more complete view of the changes needed to convey the information to all components involved.

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -1090,9 +1090,8 @@ public:
         }
 
         if (simulator_.vanguard().eclState().fieldProps().has_double("SWATINIT")) {
-            const auto& oilWaterScaledEpsInfoDrainage
-                = simulator.problem().materialLawManager()->oilWaterScaledEpsInfoDrainage(elemIdx);
-            const_cast<EclEpsScalingPointsInfo<Scalar>&>(oilWaterScaledEpsInfoDrainage).maxPcow = this->ppcw_[elemIdx];
+            simulator.problem().materialLawManager()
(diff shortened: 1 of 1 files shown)


bska:
I am creating this PR in a draft mode because it depends on upstream PR OPM/opm-common#3731 for the new member function on the `EclMaterialLawManager` class.  With this change I am able to mostly reproduce the base run in a restarted run started at the end of the historic period of a real field case that happens to have rather strong SWATINIT scaling of the oil/water capillary pressure function.

bska:
jenkins build this opm-common=3731 please

bska:
> Looks good.

Thanks.  I merged the upstream PR so I'll just formally mark this as "ready for review" and run a build check for good measure before merging.

> So this header contains code for output and in case of restarting also for input?

Yeah...   It's gotten that way over time, and "output module" isn't a good name.  I guess you can view this as an interim staging area between the simulator and the I/O layer.  When writing output, this module&ndash;and its base class&ndash;provides linearised backing storage for the output vectors which are then collected on the I/O rank and shipped off to the output writing code.  When restarting the I/O layer reads the restart file and distributes those arrays to the linearised buffers on each rank.  From there, the actual fluid state objects will internalise the values before starting the actual simulation run.

bska:
jenkins build this please

bska:
Build check is green and PR approved.  Merging this now.

bska:
By the way, @blattms, this and its upstream companion PR might be a bugfix candidate for RC-2 of the 2023.10 OPM release.  If you think so too then I can tag it with the release milestone to ensure it does not get lost.

PR #4946, opened 2023-10-25
making a few assertion to throw in StandardWellPrimaryVariables

GitPaean:
to get more outputting information and also avoid harsh termination due to assert.

They are technically numerical problem instead of programming mistakes.

### DIFF
--- a/opm/simulators/wells/StandardWellPrimaryVariables.cpp
+++ b/opm/simulators/wells/StandardWellPrimaryVariables.cpp
@@ -40,19 +40,36 @@
 #include <opm/simulators/wells/WellInterfaceIndices.hpp>
 #include <opm/simulators/wells/WellState.hpp>
 
+#include <fmt/format.h>
+
 #include <algorithm>
 #include <cassert>
+#include <limits>
+#include <string>
 
 namespace {
 
-    constexpr double EPSILON = 1.0e-14;
-
-//! \brief Relaxation factor considering only one fraction value.
+/**
+    * @brief Relaxation factor considering only one fraction value.
+    * @param old_value The previous value before the change.
+    * @param dx The solved newton update for the value.
+    * @param well_name The name of the well that calls this function
+    * @param value_name The name/meaning of the old_value.
+    * @param deferred_logger The deferred logger used for logging results and messages.
+    * @return The calculated relaxation factor.
+ */
 template<class Scalar>
 Scalar relaxationFactorFraction(const Scalar old_value,
-                                const Scalar dx)
+                                const Scalar dx,
+                                const std::string& well_name,
+                                const std::string& value_name,
+                                Opm::DeferredLogger& deferred_logger)
 {
-    assert(old_value >= -EPSILON && old_value <= (1.0+EPSILON));
+    constexpr double epislon = std::numeric_limits<Scalar>::epsilon();
+    if (old_value < -epislon ||  old_value > 1.0 + epislon) {
+        const std::string msg = fmt::format(" illegal fraction value {} {} is found for well {}", value_name, old_value, well_name);
+        OPM_DEFLOG_THROW(Opm::NumericalProblem, msg, deferred_logger);
+    }
     const Scalar& safe_old_value = std::clamp(old_value, 0.0, 1.0);
 
     Scalar relaxation_factor = 1.;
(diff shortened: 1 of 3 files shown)


GitPaean:
jenkins build this please

opm/simulators/wells/StandardWellPrimaryVariables.cpp
```diff
@@ -680,7 +697,10 @@ relaxationFactorFractionsProducer(const BVectorWell& dwells) const
                 relaxation_factor *= further_relaxation_factor;
             }
         }
-        assert(relaxation_factor >= 0.0 && relaxation_factor <= 1.0);
+        if (relaxation_factor <= 0.0 && relaxation_factor > 1.0) {
```
GitPaean, line 700: `<= 0.0` is intentional to see whether it actually causes problem.

opm/simulators/wells/StandardWellPrimaryVariables.cpp
```diff
@@ -50,9 +53,15 @@ namespace {
 //! \brief Relaxation factor considering only one fraction value.
 template<class Scalar>
 Scalar relaxationFactorFraction(const Scalar old_value,
-                                const Scalar dx)
+                                const Scalar dx,
+                                const std::string& well_name,
+                                const std::string& pv_name,
+                                Opm::DeferredLogger& deffered_logger)
```
akva2, line 59: one f

GitPaean: thanks. will address it when jenkins is done.

opm/simulators/wells/StandardWellPrimaryVariables.cpp
bska: > `<= 0.0` is intentional to see whether it actually causes problem.

Okay, but maybe you meant `f <= 0.0 || f > 1.0` instead?  As written, this condition is equivalent to `if (false)`.

opm/simulators/wells/StandardWellPrimaryVariables.cpp
bska: > thanks. will address it when jenkins is done.

Cool.  Would you mind adding doxygen-style descriptions of the parameters too, please?  That might help future maintainers get a handle on the meainings of (e.g.) `pv_name` here.

bska (commented):
This looks good to me for the most part.  Just a couple of minor issue, including, I think, one misused logical operator.

GitPaean:
jenkins build this please

opm/simulators/wells/StandardWellPrimaryVariables.cpp
GitPaean: changed to f < 0.0 || f > 1.0 to reflect the original behavoir.

GitPaean:
jenkins build this please

GitPaean:
I am marking it as ready for review. And the comments are addressed.

totto82 (approved):
Thanks for the work.

opm/simulators/wells/StandardWellPrimaryVariables.cpp
```diff
@@ -680,7 +705,10 @@ relaxationFactorFractionsProducer(const BVectorWell& dwells) const
                 relaxation_factor *= further_relaxation_factor;
             }
         }
-        assert(relaxation_factor >= 0.0 && relaxation_factor <= 1.0);
+        if (relaxation_factor < 0.0 || relaxation_factor > 1.0) {
+            const std::string msg = fmt::format(" illegal relaxation_fractor {} is obtained for well {}", relaxation_factor, this->well_.name());
```
atgeirr, line 709: Typo: fractor -> factor. Also, avoid variable names in messages, prefer the term "relaxation factor" to the variable name "relaxation_factor".

atgeirr (approved):
Only minor typo fix, otherwise fine. I did think a while about the second assert in relaxationFactorFraction(), line 89, and concluded that it is correct.

blattms:
Guys, I think you missed some asserts. e,g. [StandardWellPrimaryVariables.cpp#L89](https://github.com/OPM/opm-simulators/blob/17111bf71c1a11160660a27aa9e6ca22e8c7a9ae/opm/simulators/wells/StandardWellPrimaryVariables.cpp#L89)

blattms:
While I agree that a negative factor will break the simulation, I am a bit unsure whether we should really abort for something larger than 1. Getting something slightly larger than 1 might happen due to rounding errors.
Somebody should try what happens, if we just use 1 in this case, issue a warning and continue.

GitPaean:
> While I agree that a negative factor will break the simulation, I am a bit unsure whether we should really abort for something larger than 1. Getting something slightly larger than 1 might happen due to rounding errors.
Somebody should try what happens, if we just use 1 in this case, issue a warning and continue.

We can do 1 + EPSILON, no problem.

GitPaean:
I will change EPISLON to be `std::numeric_limits<Scalar>::epsilon();`.

GitPaean:
> > While I agree that a negative factor will break the simulation, I am a bit unsure whether we should really abort for something larger than 1. Getting something slightly larger than 1 might happen due to rounding errors.
> > Somebody should try what happens, if we just use 1 in this case, issue a warning and continue.
> 
> We can do 1 + EPSILON, no problem.

with  [StandardWellPrimaryVariables.cpp#L89](https://github.com/OPM/opm-simulators/blob/17111bf71c1a11160660a27aa9e6ca22e8c7a9ae/opm/simulators/wells/StandardWellPrimaryVariables.cpp#L89), we should not do 1 + EPSILON.

GitPaean:
There might be more that we can do in this parts.  Will keep [StandardWellPrimaryVariables.cpp#L89](https://github.com/OPM/opm-simulators/blob/17111bf71c1a11160660a27aa9e6ca22e8c7a9ae/opm/simulators/wells/StandardWellPrimaryVariables.cpp#L89), as it is for the moment. 

The original motive is that we could not reproduce a failure regarding `assert(old_value >= -EPSILON && old_value <= (1.0+EPSILON));`. 

Maybe with this PR, the people that reported the issue can have a little more output.  And also, maybe a cutting time step is also a good measure instead of terminating the simulation harshly. 

There might be more discussion in this topic with testing evolving.

opm/simulators/wells/StandardWellPrimaryVariables.cpp
GitPaean: done addressing.

GitPaean:
jenkins build this please

GitPaean:
I am merging this with the approvals and addressed comments.  As mentioned above, this is not a done story. There might be more similar adjustments with more testing.

PR #4951, opened 2023-10-26
Assign Field Property to all leaf grid view elements

aritorto:
To hide implementation details when assigning a field property to elements on the leaf grid view, using LookUpData methods (OPM/opm-grid#683).

### DIFF
--- a/ebos/eclgenericproblem_impl.hh
+++ b/ebos/eclgenericproblem_impl.hh
@@ -166,31 +166,33 @@ readRockParameters_(const std::vector<Scalar>& cellCenterDepths,
 
     unsigned numElem = gridView_.size(0);
     if (eclState_.fieldProps().has_int(rock_config.rocknum_property())) {
-        rockTableIdx_.resize(numElem);
-        const auto& num = eclState_.fieldProps().get_int(rock_config.rocknum_property());
-        for (std::size_t elemIdx = 0; elemIdx < numElem; ++ elemIdx) {
-            auto coarseElemIdx = this->getLookUpData(elemIdx);
-            rockTableIdx_[elemIdx] = num[coarseElemIdx] - 1;
-            auto fmtError =
-                [&num,coarseElemIdx,&ijkIndex,&rock_config](const char* type, std::size_t size)
-                {
-                    return fmt::format("{} table index {} for elem {} read from {}"
-                                      " is is out of bounds for number of tables {}",
-                                       type, num[coarseElemIdx],
-                                       ijkIndex(coarseElemIdx),
-                                       rock_config.rocknum_property(), size);
-                };
+        // Auxiliary function to check rockTableIdx_ values belong to the right range. Otherwise, throws.
+        std::function<void(int, int)> valueCheck = [&ijkIndex,&rock_config,this](int fieldPropValue, int coarseElemIdx)
+        {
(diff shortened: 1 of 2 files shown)


aritorto:
Yes! I'm updating the code in ebos files with the new method assignFieldPropIntOnLeaf(). I'm doing this ("step-by-step") mainly in eclgenericproblem_impl.hh (current commit lines 370 ff), and upcoming changes in lines 167-194.

ebos/eclgenericproblem_impl.hh
```diff
@@ -367,19 +367,19 @@ updateNum(const std::string& name, std::vector<T>& numbers, std::size_t num_regi
...
-            throw std::runtime_error("Values larger than maximum number of regions " + std::to_string(num_regions) + " provided in " + name);
-        } else if (numData[this->getLookUpData(elemIdx)] > 0) {
-            numbers[elemIdx] = static_cast<T>(numData[this->getLookUpData(elemIdx)]) - 1;
-        } else {
+    std::function<void(T)> valueCheck = [num_regions,name](T number) {
+        if ( (number+1) > (int)num_regions) {
+            throw std::runtime_error("Values larger than maximum number of regions "
+                                     + std::to_string(num_regions) + " provided in " + name);
+        }
+        if ( (number+1) <= 0) {
```
blattms, line 376: my impression is that this is hard to read. The check is done with the translated (C++ index instead of FORTRAN/Eclipse index). I think there should be comment on that at least.

Maybe it might be clearer to call the functor with untranslated value?

ebos/eclgenericproblem_impl.hh
```diff
@@ -367,19 +367,19 @@ updateNum(const std::string& name, std::vector<T>& numbers, std::size_t num_regi
...
             throw std::runtime_error("zero or negative values provided for region array: " + name);
         }
-    }
+    };
+
+    numbers = this->lookUpData_.assignFieldPropsIntOnLeaf(eclState_.fieldProps(), name, numElems, 1 /*defaultValue*/,
```
blattms, line 381: I think 1 might not be correct here ( with the current version of the other PR), as the default value is not translated in the companion PR.  Of course all default values will be overwritten anyway...

Maybe we should do the translation in the other PR? If we also do the checks with the untranslated values above then this is consistent.

blattms (commented):
Looks good, just wondering which values to use for checking.

ebos/eclgenericproblem_impl.hh
aritorto: It seems I could get the default values with an unordered map (https://github.com/OPM/opm-common/blob/aaa8af923ceeaaff60e788c224b86c77ea685498/opm/input/eclipse/EclipseState/Grid/FieldProps.hpp#L241). I agree that it must be better to include that in the LookUpData code. Thanks!

blattms: I probably was the one suggesting that there should be a default value. Not sure what triggered me to do that. :see_no_evil: 
These field properties should always have a value for all the cells (here on the coarse grid). So it seems we do not need to pass a default value to that function as everything gets overwritten anyway.

The code you are pointing to, will make sure that if certain field properties are omitted then we can still get the array for the field property with the correct default value used for all cells.

PR #4957, opened 2023-10-30
Treat exceptions during computation of well potentials as warnings.

blattms:
Numerical problems during the calculation were reported as errors and counted as such in the PRT file. Yet, the simulation would carry on (maybe with a smaller timestep), anyway. At least some users are parsing the PRT file to extract the number errors to get an indication whether a run succeeded. For this having
```
Error:
[opm-simulators/opm/simulators/wells/StandardWellPrimaryVariables.cpp:722] Infinite primary variable after update from wellState, well: BLA
...
Error summary:
Warnings 1416
Info 980
Errors 1
Bugs 0
Debug 0
Problems 0
```
on an otherwise successful run seemed unreasonable.

Therefore we now make sure that those exceptions are treated as warnings when they are logged (Errors 0). We do this by adding a boolean flag to DeferredLogger::logMessage to indicate that we want this kind of convertion. Default is false.

### DIFF
--- a/opm/simulators/utils/DeferredLogger.cpp
+++ b/opm/simulators/utils/DeferredLogger.cpp
@@ -82,10 +82,18 @@ namespace Opm
         messages_.push_back({Log::MessageType::Note, "", message});
     }
 
-    void DeferredLogger::logMessages()
+    void DeferredLogger::logMessages(bool convert_error_to_warning)
     {
-        for (const auto& m : messages_) {
-            OpmLog::addTaggedMessage(m.flag, m.tag, m.text);
+        if(convert_error_to_warning) {
+            for (const auto& m : messages_) {
+                OpmLog::addTaggedMessage((m.flag==Log::MessageType::Error)?
+                                         Log::MessageType::Warning : m.flag, m.tag, m.text);
+            }
+        }
+        else {
+            for (const auto& m : messages_) {
+                OpmLog::addTaggedMessage(m.flag, m.tag, m.text);
+            }
         }
         messages_.clear();
     }
--- a/opm/simulators/utils/DeferredLogger.hpp
+++ b/opm/simulators/utils/DeferredLogger.hpp
@@ -82,7 +82,7 @@ enum ExcEnum {
 
         /// Log all messages to the OpmLog backends,
         /// and clear the message container.
-        void logMessages();
+        void logMessages(bool convert_eror_to_warning=false);
 
         /// Clear the message container without logging them.
         void clearMessages();
--- a/opm/simulators/utils/DeferredLoggingErrorHelpers.hpp
+++ b/opm/simulators/utils/DeferredLoggingErrorHelpers.hpp
@@ -95,12 +95,13 @@ inline void logAndCheckForExceptionsAndThrow(Opm::DeferredLogger& deferred_logge
                                              Opm::ExceptionType::ExcEnum exc_type,
                                              const std::string& message,
                                              const bool terminal_output,
-                                             Opm::Parallel::Communication comm)
+                                             Opm::Parallel::Communication comm,
+                                             const bool convert_error_to_warning = false)
 {
     Opm::DeferredLogger global_deferredLogger = gatherDeferredLogger(deferred_logger, comm);
 
     if (terminal_output) {
-        global_deferredLogger.logMessages(/* convert_error_to_warning= */ true);
+        global_deferredLogger.logMessages(convert_error_to_warning);
     }
     // Now that all messages have been logged, they are automatically
     // cleared from the global logger, but we must also clear them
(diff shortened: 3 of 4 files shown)


blattms:
jenkins build this please

GitPaean:
@bska probably has more saying regarding this. I am requesting a review from him.

GitPaean:
I agree this is not necessarily an error, and I support to reduce the severity.  Not totally sure whether it should be a Problem or Warning, since the definition is not totally clear. I am not opposing making it a Warning.

tskille:
Have you thought about what should be the exit code from OPM Flow when number of errors are > 0 ? 

My expectation would be that this should be different from 0 to flag that the simulation has errors. Notice that If I’m running the following bash script (using opm flow built from master on 2023-10-27)

```
#!/bin/bash 

/project/multiscale/OPM/tskille/flow_bin/flow_20231027_051243 $1
exit_status=$?

echo -e "\nprogram finished with exit code:  $exit_status \n"
```

using a data deck which is producting this well potential error I'm getting the following from the PRT file

```
Error summary:
Warnings          1416
Info              980
Errors            1
Bugs              0
Debug             0
Problems          0
```

The exit code returned from OPM Flow is still 0.

```
program finished with exit code:  0 
```

GitPaean:
> Have you thought about what should be the exit code from OPM Flow when number of errors are > 0 ?
> 
> My expectation would be that this should be different from 0 to flag that the simulation has errors. Notice that If I’m running the following bash script (using opm flow built from master on 2023-10-27)
> 
> ```shell
> #!/bin/bash 
> 
> /project/multiscale/OPM/tskille/flow_bin/flow_20231027_051243 $1
> exit_status=$?
> 
> echo -e "\nprogram finished with exit code:  $exit_status \n"
> ```
> 
> using a data deck which is producing this well potential error I'm getting the following from the PRT file
> 
> ```
> Error summary:
> Warnings          1416
> Info              980
> Errors            1
> Bugs              0
> Debug             0
> Problems          0
> ```
> 
> The exit code returned from OPM Flow is still 0.
> 
> ```
> program finished with exit code:  0 
> ```

I think this is a good idea if the running gets terminated by errors.

opm/simulators/utils/DeferredLogger.cpp
```diff
@@ -82,10 +82,18 @@ namespace Opm
...
     }
 
-    void DeferredLogger::logMessages()
+    void DeferredLogger::logMessages(bool convert_error_to_warning)
     {
-        for (const auto& m : messages_) {
-            OpmLog::addTaggedMessage(m.flag, m.tag, m.text);
+        if(convert_error_to_warning) {
+            for (const auto& m : messages_) {
+                OpmLog::addTaggedMessage((m.flag==Log::MessageType::Error)?
```
GitPaean, line 89: Instead of converting, why we do not treat those as `Log::MessageType::Error` in the first place? That means, we change the definition/usage of OPM_DEFLOG_THROW or related a little bit?

GitPaean:
I now see it is related to `DeferredLogger`, I am inviting @atgeirr to have a look.

opm/simulators/utils/DeferredLogger.cpp
blattms: Because these methods are used at various different places. For some of those they are an Error.

An alternative would be to use additional special Exceptions for errors that are supposed to be ignored (and not NumericalProblem as now) and treat those as warnings in the DeferredLogger.

blattms:
Indeed I might have missed one detail: There are two call sites of logAndCheckForExceptionsAndThrow
```shell
src/dune/opm/opm-simulators$ grep -r -n logAndCheckForExceptionsAndThrow ebos opm
opm/simulators/utils/DeferredLoggingErrorHelpers.hpp:94:inline void logAndCheckForExceptionsAndThrow(Opm::DeferredLogger& deferred_logger,
opm/simulators/utils/DeferredLoggingErrorHelpers.hpp:164:logAndCheckForExceptionsAndThrow(obptc_logger, obptc_exc_type, \
opm/simulators/wells/BlackoilWellModelGeneric.cpp:1310:    logAndCheckForExceptionsAndThrow(deferred_logger, exc_type,
opm/simulators/wells/BlackoilWellModel_impl.hpp:519:        logAndCheckForExceptionsAndThrow(local_deferredLogger,
src/dune/opm/opm-simulators$
```
the latter (BlackoilWellModel_impl.hpp) causes the time step to be cut and while the former BlackoilWellModelGeneric.cpp, what this fix is about) continues with the same time step but with zero potential.

Now only the failing well potential calculation will not be counted as an error (as the time step stays the same, but not other failing well computations which will cause a reduction of the timestep.

blattms:
It is still up for debate if cutting the timestep is an error or not. My impression is that for master it counts as an error only if there is an exception when computing the wells, but not if something else in the assembly fails or the nonlinear/linear solver does not converge. I have not recheck this and my feeling could be wrong.

blattms:
Also note that even without the last patch the exceptions are always rethrown via [_throw](https://github.com/OPM/opm-simulators/blob/895074eb3b88804af979e737400c165620763f28/opm/simulators/utils/DeferredLoggingErrorHelpers.hpp#L56). Otherwise tests would have failed.

blattms:
jenkins build this please

PR #4958, opened 2023-10-30
[testing] testing whether defaulted zero thp has ever been used.

GitPaean:


GitPaean:
jenkins build this please

GitPaean:
Connected with this, not sure whether we should always use

dynamic_thp_limit_ to cover the existing THP limit from the WCON setup, 

```c++
double WellInterfaceGeneric::getTHPConstraint(const SummaryState& summaryState) const
{
    if (dynamic_thp_limit_) {
        return *dynamic_thp_limit_;
    }

    return WellBhpThpCalculator(*this).getTHPConstraint(summaryState);
}
```

Possibly we need to use the more restrictive one between the two. While working on rewriting on this function, the question related to this PR shows up.

GitPaean:
The concern regarding this PR is that 0.0 thp limit does not mean no thp limit. So if not used properly, it might have some undesirable consequence.

GitPaean:
jenkins build this please

GitPaean:
From the jenkins failures, it looks like that we might use the controls.thp_limit while the constraint does not exist. 

```
flow: /var/lib/jenkins/workspace/opm-simulators-PR-builder@2/opm/simulators/wells/WellBhpThpCalculator.cpp:85: double Opm::WellBhpThpCalculator::getTHPConstraint(const Opm::SummaryState&) const: Assertion `controls.hasControl(Well::ProducerCMode::THP)' failed.
```

Will investigate this.

GitPaean:
The assertion failures are from the following code, 
```c++
    const double thp_limit = well_.getTHPConstraint(summary_state);
    ws.thp = this->calculateThpFromBhp(rates, ws.bhp, rho, alq, thp_limit, deferred_logger);
```

somehow, due to WVFPDP, we need the thp_limit to do something, while `calculateThpFromBhp` should not need thp_limit. Moreover, thp_limit does not exist sometime, a zero value will be used there.

GitPaean:
The intention was to make sure `double WellBhpThpCalculator::getTHPConstraint(const SummaryState& summaryState) const` does not return any undefined THP limit value. (not THP constraint defined and return some value (most likely zero from the default value as a member variable)). 

No intention to change any implementation and result. 

Will clean up and propose a PR.

GitPaean:
jenkins build this please

PR #4959, opened 2023-10-30
Use new linearizer for gaswater with thermal and diffusion

totto82:


totto82:
jenkins build this please

GitPaean:
Thanks for this. This is NOT something I am familiar with while if any case in the  opm-tests will use this PR?

totto82:
I think it should have been part of https://github.com/OPM/opm-simulators/pull/4814 but somehow was forgotten. We really want the speed-up also for the gas-water case with diffusion and/or thermal.

GitPaean:
> I think it should have been part of https://github.com/OPM/opm-simulators/pull/4814 but somehow was forgotten. We really want the speed-up also for the gas-water case with diffusion and/or thermal.

Sure. Have you tested with any cases with this PR?

totto82:
Yes. The test shows the factor 2 speed-up on assembly as expected.

GitPaean:
Thanks for the PR. I am merging this PR now as jenkins and testing approve.

PR #4960, opened 2023-10-30
Support for mechanical dispersion

daavid00:
With https://github.com/OPM/opm-common/pull/3737 and https://github.com/OPM/opm-models/pull/847, this PR adds support for the linear disperison model described in the SPE CSP11 benchmark document (Eq. 2.3) (https://github.com/Simulation-Benchmarks/11thSPE-CSP/blob/main/description/spe_csp11_description.pdf). Two cases are added in https://github.com/OPM/opm-tests/pull/1072 , where the effect of adding dispersion is assessed for a finger and injection problem.
Screenshot 2023-10-30 at 15 47 40

### DIFF
--- a/ebos/ecltransmissibility_impl.hh
+++ b/ebos/ecltransmissibility_impl.hh
@@ -93,7 +93,8 @@ EclTransmissibility(const EclipseState& eclState,
                     const Grid& grid,
                     std::function<std::array<double,dimWorld>(int)> centroids,
                     bool enableEnergy,
-                    bool enableDiffusivity)
+                    bool enableDiffusivity,
+                    bool enableDispersivity)
       : eclState_(eclState)
       , gridView_(gridView)
       , cartMapper_(cartMapper)
@@ -101,6 +102,7 @@ EclTransmissibility(const EclipseState& eclState,
       , centroids_(centroids)
       , enableEnergy_(enableEnergy)
       , enableDiffusivity_(enableDiffusivity)
+      , enableDispersivity_(enableDispersivity)
       , lookUpData_(gridView)
       , lookUpCartesianData_(gridView, cartMapper)
 {
@@ -148,6 +150,17 @@ diffusivity(unsigned elemIdx1, unsigned elemIdx2) const
 }
 
 template<class Grid, class GridView, class ElementMapper, class CartesianIndexMapper, class Scalar>
+Scalar EclTransmissibility<Grid,GridView,ElementMapper,CartesianIndexMapper,Scalar>::
+dispersivity(unsigned elemIdx1, unsigned elemIdx2) const
+{
+    if (dispersivity_.empty())
+        return 0.0;
+
+    return dispersivity_.at(isId(elemIdx1, elemIdx2));
+
+}
+
+template<class Grid, class GridView, class ElementMapper, class CartesianIndexMapper, class Scalar>
 void EclTransmissibility<Grid,GridView,ElementMapper,CartesianIndexMapper,Scalar>::
 update(bool global, const std::function<unsigned int(unsigned int)>& map, const bool applyNncMultregT)
 {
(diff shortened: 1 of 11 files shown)


ebos/ecltransmissibility_impl.hh
```diff
@@ -157,6 +170,7 @@ update(bool global, const std::function<unsigned int(unsigned int)>& map, const
     // get the ntg values, the ntg values are modified for the cells merged with minpv
     const std::vector<double>& ntg = eclState_.fieldProps().get_double("NTG");
     const bool updateDiffusivity = eclState_.getSimulationConfig().isDiffusive() && enableDiffusivity_;
+    const bool updateDispersivity = eclState_.getSimulationConfig().rock_config().dispersion() && enableDispersivity_;
     unsigned numElements = elemMapper.size();
```
hnil, line 174: Is enabledispersivity given in the deck or a compile time option. If it is a compile time option I think the run should fail.

ebos/ecltransmissibility_impl.hh
```diff
@@ -483,6 +507,42 @@ update(bool global, const std::function<unsigned int(unsigned int)>& map, const
...
+                                                        axisCentroids),
+                                        dispersion_[outsideElemIdx]);
+
+                applyNtg_(halfDispersivity1, insideFaceIdx, elemIdx, ntg);
+                applyNtg_(halfDispersivity2, outsideFaceIdx, outsideElemIdx, ntg);
+
+                //TODO Add support for multipliers
+                Scalar dispersivity;
+                if (std::abs(halfDispersivity1) < 1e-30 || std::abs(halfDispersivity2) < 1e-30)
+                    // avoid division by zero
```
hnil, line 538: Probably ok, but I think this eps should be moved to a common place. Also in similar code for thermalconductivity I think <= 0 is used.

ebos/ecltransmissibility_impl.hh
```diff
@@ -93,14 +93,16 @@ EclTransmissibility(const EclipseState& eclState,
...
       , gridView_(gridView)
       , cartMapper_(cartMapper)
       , grid_(grid)
       , centroids_(centroids)
       , enableEnergy_(enableEnergy)
       , enableDiffusivity_(enableDiffusivity)
+      , enableDispersivity_(enableDispersivity)
 {
     const UnitSystem& unitSystem = eclState_.getDeckUnitSystem();
     transmissibilityThreshold_  = unitSystem.parse("Transmissibility").getSIScaling() * 1e-6;
```
hnil, line 108: Again if I think this arbitrary constant should be moved to a common place. Maybe mark with a comment which is easy to search for when one agree on how to do this.

hnil (approved):
For me it is ok. Made some comments.

ebos/ecltransmissibility_impl.hh
daavid00: This is just copy from diffusion, I agree this gathering of all epsilons should be in a common place, but maybe this should be done in a separate PR for all parameters.

ebos/ecltransmissibility_impl.hh
daavid00: It is a compile time option. But since the dispersivity returns 0.0 if empty (as in diffusivity), it does not fail.

daavid00:
Many thanks @hnil for your comments, I have addressed them. Let me know if there are further issues.

ebos/ecltransmissibility_impl.hh
atgeirr: If the rock_config().dispersion() is true and enableDispersivity_ is false we should abort the run, as we are trying to run a deck with dispersion in a simulator instantiation that does not honor it. Same with diffusion.

atgeirr (approved):
Looks good to me, only minor comment about running a dispersion case with a no-dispersion simulator instance.

PR #4962, opened 2023-10-31
[bugfix] Deprecated getCollectiveComminication 

aritorto:
After DUNE version 2.7, the method getCollectiveCommunication() was changed to getCommunication(). The getCollectiveCommunication() issues the deprecation warning  and assumes all process take part in the communication. However, this might not be the case, namely, only a subset might be used. 

The fix consists in getting the communication object used be the grid as this uses only processes that are actually computing and not only outputting.


aritorto:
jenkins build this please

aritorto:
Thanks @blattms for the explanation on the deprecated warming!

bska (approved):
Very good&ndash;thanks for staying on top of this.  I'll merge into master.

aritorto:
Perfect! Thanks!

PR #4964, opened 2023-10-31
Output XMF and YMF for H2STORE

svenn-t:
Simple extension of the feature that is already available for CO2STORE


totto82:
jenkins build this please

totto82:
Thanks.

PR #4965, opened 2023-10-31
Use min timestep from TUNING also when updating (not only initially) 

vkip:
(+ using the right value, TSMINZ, initially). Together with https://github.com/OPM/opm-common/pull/3740, should eliminate side effects of --enable-tuning=true when there is no TUNING record or non-default TUNING record items.


vkip:
jenkins build this opm-common=3740 please

vkip:
Test failures apparently due to changing the default value of minimum time step to coincide with the Flow default (i.e., without --enable-tuning=true). [Value is written to DOUBHEAD in the .INIT file.]

vkip:
jenkins build this opm-common=3740 please

blattms:
@vkip Should this still be reviewed and merged or is this resolved in another way already?

vkip:
> @vkip Should this still be reviewed and merged or is this resolved in another way already?

To my knowledge this is not yet resolved.

vkip:
Or actually, using the min timestep you maybe fixed when working on NEXTSTEP? If so the only relevant thing here might be the changed message (getting the correct value in days).

PR #4966, opened 2023-10-31
Fix to linker error in building tests (clang, macOS)

daavid00:
With cmake flag -DUSE_MPI=0, executing `make` leads to linker errors in _test_tuning_tsinit_nextstep_ and _test_tuning_xxxMBE_. This commit fixes this. 
clang_linker_error

### DIFF
--- a/ebos/eclgenericvanguard.cc
+++ b/ebos/eclgenericvanguard.cc
@@ -53,6 +53,7 @@
 #include <opm/input/eclipse/Schedule/UDQ/UDQConfig.hpp>
 #include <opm/input/eclipse/Schedule/UDQ/UDQState.hpp>
 #include <opm/input/eclipse/Schedule/Well/NameOrder.hpp>
+#include <opm/input/eclipse/Schedule/Well/WDFAC.hpp>
 #include <opm/input/eclipse/Schedule/Well/Well.hpp>
 #include <opm/input/eclipse/Schedule/Well/WellBrineProperties.hpp>
 #include <opm/input/eclipse/Schedule/Well/WellConnections.hpp>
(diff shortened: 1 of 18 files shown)


tests/test_tuning_TSINIT_NEXTSTEP.cpp
```diff
@@ -5,7 +5,7 @@
 #include <iostream>
 
 #define BOOST_TEST_MODULE TestTuningTSINIT
-#include <boost/test/unit_test.hpp>
+#include <boost/test/included/unit_test.hpp>
```
blattms, line 8: On my system there is no boost/test/included directory.

Can you please try whether it works for when adding 
```
#include <config.h>
```

at the top instead? That file should have
```
#define HAVE_DYNAMIC_BOOST_TEST 1
```
in it which probably does the trick.

blattms:
Please also check whether the problem is in the release.
I am running jenkins now, but I assume it will not find the header in the new location.

blattms:
jenkins build this please

daavid00:
Thanks @blattms for your quick response. I tried to add `#include <config.h>` instead but that does not fix the linking issue. I will try to find a fix by modifying the _CMakeLists.txt_.

daavid00:
I just updated the commit by adding those tests to _CMAKELists_files.cmake_ and there are no more linker issues.

akva2:
However now there are faulty tests added. This is not a proper fix.

daavid00:
I have updated now the commit, adding conditionals to the .cpp tests for clang and macOS.

akva2:
Which also feels like the wrong fix. If adding it to the tests list made it work fine, it means that the code for building tests sets some props that is the correct fix, https://github.com/OPM/opm-common/blob/e2c8d5536ce76bf0e1834e1b9457e5c0ddf6163a/cmake/Modules/OpmSatellites.cmake#L58 it is most likely hose link flags or the dyn flag further down.

bska:
The top-level Boost.Test header is always [`<boost/test/unit_test.hpp>`](https://www.boost.org/doc/libs/1_83_0/libs/test/doc/html/header/boost/test/unit_test_hpp.html).  Any path other than that will be an implementation detail,  I agree with @akva2 that the problem must be elsewhere and I'm opposed to the change proposed in this PR at the time of writing.

PR #849, opened 2016-10-03
Allow crossflow if all perforations is crossflowing

totto82:
This implements the same hack that is used in flow. With this fix Model 2 results are ok until 1500 days when MULTFLT appears. Changes in geoprops in not yet supported in ebos.

PR #4973, opened 2023-11-07
Enabling disGas and watVap at compile time for gasWaterBrine systems

daavid00:
This allows to run decks with 
```
WATER
GAS
BRINE
CO2STORE
VAPWAT
DISGASW
```


goncalvesmachadoc:
I see I made  a poor choice for this [flag](https://github.com/OPM/opm-models/blob/818c4b97056b49f40346c070a25ee2f682f00cb5/opm/models/blackoil/blackoilproperties.hh#L66) name. Should we change it to something like _EnableVapwat_?

daavid00:
Hei, I have made the renaming as suggested (now it depends on [#849](https://github.com/OPM/opm-models/pull/849)).

PR #4976, opened 2023-11-08
Fix rsw computation of wellbore density for StW

### LINKED ISSUE #4972: Calculation of gas dissolved in water ratio
https://github.com/OPM/opm-simulators/blob/763d25541a7b1a0bbf885c7f0be57596b5f5e33f/opm/simulators/wells/StandardWellConnections.cpp#L348

I think this should be
`rsw = gasrate / waterate;`

totto82:
closes #4972


totto82:
jenkins build this please

bska (approved):

PR #4979, opened 2023-11-08
log problem if asked to continue with nonconverged nonlinear solver.

blattms:
The user has explicitly asked for this behavior using --solver-continue-on-convergence-failure=true in this rare case.

Completes #4977


atgeirr:
Thanks, will merge. This is in line with other recent adjustments: Issues that to not require termination of the simulators should be classified as Problem, not Error.

PR #4983, opened 2023-11-09
Lookup NTG on Leaf Grid View

aritorto:
To support LGRs, the lookup of NTG has been updated, namely, for a leaf grid view cell, the ntg value is looked up via its parent cell (if existent), or its equivalent cell in level 0.  

Needs OPM/opm-grid#689 .

### DIFF
--- a/ebos/ecltransmissibility_impl.hh
+++ b/ebos/ecltransmissibility_impl.hh
@@ -101,6 +101,8 @@ EclTransmissibility(const EclipseState& eclState,
       , centroids_(centroids)
       , enableEnergy_(enableEnergy)
       , enableDiffusivity_(enableDiffusivity)
+      , lookUpData_(gridView)
+      , lookUpCartesianData_(gridView, cartMapper)
 {
     const UnitSystem& unitSystem = eclState_.getDeckUnitSystem();
     transmissibilityThreshold_  = unitSystem.parse("Transmissibility").getSIScaling() * 1e-6;
@@ -350,8 +352,11 @@ update(bool global, const std::function<unsigned int(unsigned int)>& map, const 
                                               axisCentroids),
                               permeability_[outsideElemIdx]);
 
-            applyNtg_(halfTrans1, insideFaceIdx, elemIdx, ntg);
-            applyNtg_(halfTrans2, outsideFaceIdx, outsideElemIdx, ntg);
+            auto coarseElemIdx = lookUpData_.template getFieldPropIdx<Grid>(elemIdx);;
+            auto coarseOutsideElemIdx = lookUpData_.template getFieldPropIdx<Grid>(outsideElemIdx);
+            
+            applyNtg_(halfTrans1, insideFaceIdx, coarseElemIdx, ntg);
+            applyNtg_(halfTrans2, outsideFaceIdx, coarseOutsideElemIdx, ntg);
 
             // convert half transmissibilities to full face
             // transmissibilities using the harmonic mean
@@ -469,8 +474,11 @@ update(bool global, const std::function<unsigned int(unsigned int)>& map, const 
                                                         axisCentroids),
                                         porosity_[outsideElemIdx]);
 
-                applyNtg_(halfDiffusivity1, insideFaceIdx, elemIdx, ntg);
-                applyNtg_(halfDiffusivity2, outsideFaceIdx, outsideElemIdx, ntg);
+                auto coarseElemIdx = this->lookUpData_.template getFieldPropIdx<Grid>(elemIdx);
+                auto coarseOutsideElemIdx = this-> lookUpData_.template getFieldPropIdx<Grid>(outsideElemIdx);
+
+                applyNtg_(halfDiffusivity1, insideFaceIdx, coarseElemIdx, ntg);
+                applyNtg_(halfDiffusivity2, outsideFaceIdx, coarseOutsideElemIdx, ntg);
 
                 //TODO Add support for multipliers
                 Scalar diffusivity;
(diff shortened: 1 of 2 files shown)


blattms:
This seems to be done differently than for other field properties. Maybe we should handle it consistently and make sure that the arrays are always of the size of elements on the leaf grid view?

E.g. in #4951 we made sure that we have arrays that live on the leaf grid view using `LookUpData::assignFieldPropsIntOnLeaf`, etc. . In a similar way there could be a `getFieldPropsDoubleOnLeaf` that returns the same as `eclState_.fieldProps().get_double(property_name)` without LGR and with LGR does project the values to an array on the leaf grid. The we would need to change only one line [ecltransmissibility_impl.hh#L158](https://github.com/OPM/opm-simulators/blob/ea9bc396570b38948b170b0c78c0ffb65974d617/ebos/ecltransmissibility_impl.hh#L158).

blattms:
There are some warnings about shadowed variables: https://ci.opm-project.org/job/opm-grid-PR-builder/478/gcc/

aritorto:
Indeed, thanks for noticing. We can use the existing method in one line.

aritorto:
jenkins build this please

ebos/ecltransmissibility_impl.hh
```diff
@@ -469,8 +474,8 @@ update(bool global, const std::function<unsigned int(unsigned int)>& map, const
                                                         axisCentroids),
                                         porosity_[outsideElemIdx]);
 
-                applyNtg_(halfDiffusivity1, insideFaceIdx, elemIdx, ntg);
-                applyNtg_(halfDiffusivity2, outsideFaceIdx, outsideElemIdx, ntg);
+                applyNtg_(halfDiffusivity1, insideFaceIdx, coarseElemIdx, ntg);
+                applyNtg_(halfDiffusivity2, outsideFaceIdx, coarseOutsideElemIdx, ntg);
```
blattms, line 478: Does this still work after the change above? I thought it should be elemIdx outsideElemIdx as before.

ebos/ecltransmissibility_impl.hh
```diff
@@ -1038,7 +1043,7 @@ applyNncMultreg_(const std::unordered_map<std::size_t,int>& cartesianToCompresse
     const auto& inputNNC = this->eclState_.getInputNNC();
     const auto& transMult = this->eclState_.getTransMult();
 
-    auto compressedIdx = [&cartesianToCompressed](const std::size_t globIdx)
+    auto compressedIdx = [&cartesianToCompressed, this](const std::size_t globIdx)
```
blattms, line 1046: "this" does not seem to be used. Please don't capture it.

ebos/ecltransmissibility_impl.hh
```diff
@@ -495,7 +500,7 @@ update(bool global, const std::function<unsigned int(unsigned int)>& map, const
     // Loop over all elements (global grid) and store Cartesian index
     for (const auto& elem : elements(grid_.leafGridView())) {
         int elemIdx = elemMapper.index(elem);
-        int cartElemIdx = cartMapper_.cartesianIndex(elemIdx);
+        int cartElemIdx =  this->lookUpCartesianData_.template getFieldPropCartesianIdx<Grid>(elemIdx);
```
blattms, line 503: Can you remind me what the difference between the calls is?
I was under the impression that both should do the same. In that case we should probably use the old version.

blattms (changes requested):
I think there is a problem with the changes now. ntg_ has entries for each element of the leaf grid, but we use the indices of the parent to access its elements.

ebos/ecltransmissibility_impl.hh
aritorto: Thanks, it was a left over of an old-wrong change. I'll remove it

ebos/ecltransmissibility_impl.hh
aritorto: They do the same with and without LGRs, when all the field properties are assumed to be given in the level 0/coarse grid. The difference is when field properties for LGRs are given in their respected refined grids (levels>0). Then, the lookUpCartesianData_ method will get the "local/level" Cartesian index of the leaf grid view cell, on the level this cell was born, and this is not necessarily level 0.

With the current assumptions, we can of course leave the old line of code, and go back to change it when refined field properties will be supported.

ebos/ecltransmissibility_impl.hh
aritorto: Indeed, thanks for noticing it!

ebos/ecltransmissibility_impl.hh
aritorto: it's back to the old version

aritorto:

blattms (approved):

blattms:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: drogon - Threads: 1                | 0.998    |
| opm-git | OPM Benchmark: drogon - Threads: 8                | 1.006    |
| opm-git | OPM Benchmark: punqs3 - Threads: 1                | 0.995    |
| opm-git | OPM Benchmark: punqs3 - Threads: 8                | 1.006    |
| opm-git | OPM Benchmark: smeaheia - Threads: 1              | 1.01     |
| opm-git | OPM Benchmark: smeaheia - Threads: 8              | 1.002    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 1         | 1.012    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 8         | 1.007    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 0.994    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 1.002    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 0.999    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 1.002    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.999    |



View result details @ https://www.ytelses.com/opm/?page=result&id=2289

blattms:
Cool, no performance penalty with this.

PR #4987, opened 2023-11-09
Update manpage for release

blattms:
No need to run jenkins.


blattms:
shameless self-mergr :grin:

PR #4990, opened 2023-11-10
Avoid some dead stores

akva2:
As flagged by static analyzers.


akva2:
jenkins build this please

blattms:
About the dead stores in ddd0c5629dd50a69b2230f86943246be8b21c7ee : To me it seems like all stores in result in that function are dead, not only those two. Actually the local variable result could even be removed. It does not seem to be read anywhere.

What am I missing?

akva2:
It is returned from the function below.

blattms:
I see. Sorry missed that.

PR #4989, opened 2023-11-10
Chase Constructor API Change for Opm::Connection Class

bska:
Downstream follow-up to OPM/opm-common#3763.  This must be merged in concert with the upstream PR.

PR #4992, opened 2023-11-10
Calculate D-Factor in Terms of Connection Class

bska:
This PR follows the API change of `WDFAC::getDFactor()` (PR OPM/opm-common#3765) to take the dynamic viscosity and gas density values in addition to a `Connection` object.

While here, also restructure the logic a little bit in order to avoid computing viscosity values unless they're actually needed.

### DIFF
--- a/opm/simulators/wells/WellInterface.hpp
+++ b/opm/simulators/wells/WellInterface.hpp
@@ -20,9 +20,16 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-
 #ifndef OPM_WELLINTERFACE_HEADER_INCLUDED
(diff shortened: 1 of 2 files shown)


bska:
I am creating this PR in draft mode for two reasons

1. It depends on, and contains, the earlier PR #4989.
2. It depends on an upstream API change (OPM/opm-common#3765) for member function `WDFAC::getDFactor()`.

I will keep the PR in draft state until it is ready for review.

bska:
The upstream PR #4989 has been merged into the master branch.  Marking this PR as "ready for review" now.

bska:
jenkins build this opm-common=3765 please

PR #4993, opened 2023-11-10
Communicate Dynamic Transmissibility Multiplier to Output Layer

bska:
This PR captures the rock compaction transmissibility multiplier in the `PerfData` during the non-linear iterations and communicates the converged value back to the output layer through the new data member to the `data::Connection` structure,
```C++
double data::Connection::compact_mult
```
added in upstream PR OPM/opm-common#3766.

### DIFF
--- a/opm/simulators/wells/PerfData.cpp
+++ b/opm/simulators/wells/PerfData.cpp
@@ -27,7 +27,10 @@
 
 namespace Opm {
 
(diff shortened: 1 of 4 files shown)


bska:
I am creating this PR in draft mode because it depends on the upstream PR OPM/opm-common#3766 to provide the new data member.  I have also placed the PR on top of the earlier work #4992, although it could in principle be made indpendent of that change.

bska:
The upstream PR has been merged into the master branch.  I'm marking this as "ready for review".

PR #4994, opened 2023-11-13
Check connect only once, with elemIdx instead of cart

aritorto:
To process intersections only once, Cartesian indices were used. When LGRs, two neighbors might have the same Cartesian index. Now, checking that each intersection is considered only once is done via element indices.


aritorto:

blattms:
Perfect. Thanks a lot. In it goes.

PR #4995, opened 2023-11-13
Lgrs neighbors considered in removeSmallNonCart

aritorto:
All neighbors in each LGR (if existent) are Cartesian neighbors on the level grid representing the LGR. 
When elements on the leaf grid view have the same parent cell, then their Cartesian Index on the leaf grid View coincide. 
This case is also taken into account in removeSmallNonCartesianTransmissibilities_().


aritorto:

PR #4996, opened 2023-11-14
Fix wdfac and d factor in comdat

totto82:
Depends on https://github.com/OPM/opm-common/pull/3773


opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -868,9 +868,10 @@ namespace Opm
                 }
             }
             // injecting connections total volumerates at standard conditions
-            const Value factor = total_mob / volume_ratio;
             for (int componentIdx = 0; componentIdx < this->numComponents(); ++componentIdx) {
-                cq_s[componentIdx] = - Tw[componentIdx] * cmix_s[componentIdx] * factor * drawdown;
+                const Value cqt_i = - Tw[componentIdx] * (total_mob * drawdown);
+                Value cqt_is = cqt_i / volume_ratio;
+                cq_s[componentIdx] = cmix_s[componentIdx] * cqt_is;
             }
```
totto82, line 875: For some reason the slight difference in the order of the operations to avoid computing the total_mob/volume_ratio factor for every component resulted in test failures in almost all the tests. I thus went back to the original order of operations.

bska (approved):

mbmfer:
Hi, I can see that the issue with unrecognized rate dependent skin setting has been addressed in the past. Can you please let me know how I can incorporate this into the 2023.04 version of Flow?. Thanks a lot.

bska:
> I can see that the issue with unrecognized rate dependent skin setting has been addressed in the past.

Flow-dependent skin ("D-factors") is still very much in development, and we've merged new changes and bug-fixes for the feature to the development branches in just the past two hours as I'm typing this.  This feature is not available in any released/packaged version of the simulator.

> Can you please let me know how I can incorporate this into the 2023.04 version of Flow?

At this moment you essentially need to be running the very latest development version of the simulator to get support for `WDFAC`, `WDFACCOR`, and item 12 of `COMPDAT`.  Incorporating this into the 2023.04 version will, I suspect, require considerable effort and code changes and the ability to recompile the simulator from sources.  Unless you have experience doing such work, or are willing to put in the effort to learn this, then I would recommend against taking on the task.  I'm not the release manager, but I suspect the project won't create special-purpose binaries to preview this feature.

mbmfer:
Thanks Bard, Is version 2023.10 the very latest development version?

PR #4998, opened 2023-11-15
use stricter tolerances for WINJMULT_MSW/WINJDAM_MSW parallel tests

akva2:
so we can use standard rel_tol tolerances and not the very relaxed values currently in use

### DIFF
--- a/parallelTests.cmake
+++ b/parallelTests.cmake
@@ -177,17 +177,17 @@ add_test_compare_parallel_simulation(CASENAME winjmult_msw
                                      FILENAME WINJMULT_MSW
                                      SIMULATOR flow
                                      ABS_TOL ${abs_tol}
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

GitPaean:
It is good to be able to use ` standard rel_tol tolerances`.  Have you tried whether CPR linear solvers can help these cases?  (was inspired by https://github.com/OPM/opm-tests/pull/1092)

akva2:
not really and that would be a bit tangential to my current goal. we need to stabilize the tests (in particular WINJDAM_MSW) as they caused headache in https://github.com/OPM/opm-common/pull/3763

bska:
Closing this now as an equivalent change was merged as part of #4989.

PR #4999, opened 2023-11-15
Always check if wells on historic controls can be opened

vkip:
Ref https://github.com/OPM/opm-common/pull/3771


vkip:
jenkins build this please

GitPaean:
It might be something evolving and we keep adjusting later. The current form looks fine to me. We can update when more findings come later.

PR #5001, opened 2023-11-17
limiting the lower limit of 1 - rs*rv

GitPaean:
to do an urgent fix while refactoring the issue on a broader view, while needs much more investigation, discussion and testing.

It is likely we need more restrictive rs and rv than the following code can provide. 
```c++
            if (!props.rsmax_perf.empty() && mix[oilpos] > 1e-12) {
                rs = std::min(mix[gaspos] / mix[oilpos], props.rsmax_perf[perf]);
            }
            if (!props.rvmax_perf.empty() && mix[gaspos] > 1e-12) {
                rv = std::min(mix[oilpos] / mix[gaspos], props.rvmax_perf[perf]);
            }
```

And we should limit the rsmax or rvmax, and using mix to derive rs and rv can be non-physical.  It is possible `1. - rs*rv`  should not be so much smaller than 1., at least not in orders. 

This PR should not impact much results, except some extremely sensitive cases, but it should not make any running worse.

### DIFF
--- a/opm/simulators/wells/StandardWellConnections.cpp
+++ b/opm/simulators/wells/StandardWellConnections.cpp
@@ -212,7 +212,8 @@ computeDensities(const std::vector<Scalar>& perfComponentRates,
                 rv = std::min(mix[oilpos] / mix[gaspos], props.rvmax_perf[perf]);
             }
             const Scalar d = 1.0 - rs*rv;
-            if (d <= 0.0) {
+            constexpr Scalar epsilon = 1.e-5;
+            if (d <= epsilon) {
                 std::ostringstream sstr;
                 sstr << "Problematic d value " << d << " obtained for well " << well_.name()
                      << " during computeConnectionDensities with rs " << rs


GitPaean:
jenkins build this please

tskille:
This PR has been tested on a full field model in Equinor and results are good. Looking forward to getting this into master.

GitPaean:
> to do an urgent fix while refactoring the issue on a broader view, while needs much more investigation, discussion and testing.
> 
> It is likely we need more restrictive rs and rv than the following code can provide.
> 
> ```c++
>             if (!props.rsmax_perf.empty() && mix[oilpos] > 1e-12) {
>                 rs = std::min(mix[gaspos] / mix[oilpos], props.rsmax_perf[perf]);
>             }
>             if (!props.rvmax_perf.empty() && mix[gaspos] > 1e-12) {
>                 rv = std::min(mix[oilpos] / mix[gaspos], props.rvmax_perf[perf]);
>             }
> ```
> 
> And we should limit the rsmax or rvmax, and using mix to derive rs and rv can be non-physical. It is possible `1. - rs*rv` should not be so much smaller than 1., at least not in orders.
> 
> This PR should not impact much results, except some extremely sensitive cases, but it should not make any running worse.

It does not look like there is something straightforward to limit the `rs*rv` to be smaller than 1. in a physical meaningful way.

bska (commented):
> It does not look like there is something straightforward to limit the `rs*rv` to be smaller than 1. in a physical meaningful way.

Okay, but what's the point of this PR, then?  It replaces one (almost) arbitrary threshold value with another (almost) arbitrary threshold value.  Is there really not guiding principles here?

GitPaean:
>  Okay, but what's the point of this PR, then? It replaces one (almost) arbitrary threshold value with another (almost) arbitrary threshold value. Is there really not guiding principles here?

It fixes a running problem without fixing the root cause of the problem with a minimal code change. 

The original code is 

```c++
            const Scalar d = 1.0 - rs*rv;
            if (d <= 0.0) {
```

not sure whether it ever `d < 0.`, but it happens very often `d == 0.` for the well injection potential calculation.  When `d == 0.`, it means all the gas is dissolved in the oil and all the oil vaporize to the gas.  When `d == 0.`, the master branch basically continue with `rs == 0.` and `rv == 0.`, because it is nonphysical.  It happens often when calculating the well potentials of injectors with default BHP constraint (6812 barsa.)

The targeting scenario happens when `d` is on the level of 1.e-16 due to round error.  For that scenario, the connection gets really huge (e+16  - e+19) level fluid density. The well will be shut due to the huge connection pressure drop.  This PR basically make this scenario run without impacting the simulation result of master branch.  The PR should not make anything worse compared with the master branch while make the targeting case run. 

It indeed does not fix the root cause.  The root cause of the problem is the extrapolation of the PVT table to really high pressure resulting in very big rsmax and rvmax, or we should not update the connection density and pressure (explicit quantities) for the well potential calculations, or we should iterate multiple times to make sure the explicit quantities better quality (half implicit?). 

Let me know if you think the approach in the PR is helpful.  If not, I can look for other approaches.

GitPaean:
With some discussion, this PR is not favored. Closing it now.

PR #5003, opened 2023-11-17
Check if well was closed this step before re-opening during history

vkip:
Fix potential infinite loop introduced in https://github.com/OPM/opm-simulators/pull/4999 (sorry).


vkip:
jenkins build this please

GitPaean:
Thanks for the update.  I mistakenly assumed https://github.com/OPM/opm-simulators/pull/4999 was tested with the targeting case. It was also my fault. But I did realized there might be some adjustments coming based on testing findings. 

I am getting this PR in now.

PR #5005, opened 2023-11-20
dont scale d factor from wdfaccor with connection factor

totto82:
The d factor is already scaled by the rate for WDFACCOR. 

Depends on https://github.com/OPM/opm-common/pull/3781


totto82:
jenkins build this please

bska:
jenkins build this please

bska (approved):
Thanks.  This is a substantially improved API.  I'll merge once the build check goes through.

bska (approved):
This is a great improvement in the interface.  Thanks a lot!

PR #5007, opened 2023-11-20
fixed: we need to initialize logger before parsing

akva2:
parsing logs some errors which meant using uninitialized values if no logger has been set


akva2:
jenkins build this please

bska (approved):
Yup, that makes sense.  We have the same ordering in the [main simulator](https://github.com/OPM/opm-simulators/blob/4ccc5cddc78c45c722d4ac6e8c6fd63ff77c6393/opm/simulators/flow/Main.cpp#L190-L194) construction process.

PR #5008, opened 2023-11-20
compressedLevelZero used to read relative permeability

aritorto:
When reading relative permeability defined by the input file, (currently) done for a grid without LGRs, the compressed index used has to be the one for level 0 grid. Therefore, a method compressedLevelZeroSize() has been added in the CartesianIndexMapper interface, for AluGrid. For CpGrid and Polyhedral, it's done in OPM/opm-grid694.


blattms (approved):

blattms:
Perfect. This looks great!

PR #5009, opened 2023-11-20
add graph coloring

multitalentloes:
This pr adds several new graph coloring algorithms that can be applied to Dune matrices. The colorings will pick out rows in sparse matrices that can be computed simultaneously when performing an upper or lower triangular solve.

### DIFF
--- a/opm/simulators/linalg/GraphColoring.hpp
+++ b/opm/simulators/linalg/GraphColoring.hpp
@@ -19,6 +19,10 @@
 #ifndef OPM_GRAPHCOLORING_HEADER_INCLUDED
 #define OPM_GRAPHCOLORING_HEADER_INCLUDED
 
+#include <opm/common/TimingMacros.hpp>
+
+#include <opm/grid/utility/SparseTable.hpp>
+
 #include <algorithm>
 #include <cstddef>
 #include <deque>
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this please

opm/simulators/linalg/GraphColoring.hpp
```diff
@@ -218,5 +222,94 @@ reorderVerticesSpheres(const std::vector<int>& colors,
...
+{
+    OPM_TIMEBLOCK(createMatrix);
+
+    std::vector<std::size_t> color(matrix.N(), 0);
+    std::vector<std::size_t> rowIndices(matrix.N(), 0);
+
+#ifdef _OPENMP
+#pragma omp parallel for
+#endif
+    for (std::size_t i = 0; i < matrix.N(); i++) {
```
akva2, line 258: no point parallelizing this loop i don't think, and thus
 ```
std::iota(rowIndices.begin(), rowIndices.end(), 0);
```
is simpler and more expressive.

akva2 (approved):
One minor suggestion for some code simplification, otherwise this looks good.

opm/simulators/linalg/GraphColoring.hpp
multitalentloes: Why is there no point, will this not run faster with multiple threads?

akva2: maybe it would run slightly faster if threads end up on processors with separate memory lanes but it is really nothing you'd be able to measure. there is basically no computation here, only writing to memory.

akva2:
jenkins build this please

PR #5010, opened 2023-11-21
Add output of connection dissolved gas and vaporized oil

totto82:
Depends on https://github.com/OPM/opm-common/pull/3782


bska (approved):
This looks good to me now and I'll merge into master.

PR #5015, opened 2023-11-21
not updating connection pressure for StandardWell early in computeWellRatesWithBhpIterations. 

GitPaean:
In `computeWellRatesWithBhpIterations()`, the perforation rates are not updated with updated well rates, and explicit quantities are not sensible to update based on the inconsistent well rates and perforation rates.

Better to keep the original explicit quantities for better consistency.

Furthermore, it can be dangerous to update the explicit quantities based on the irrelevant perforation rates, since the ratios can be very undesirable due to cross flow.  

Connection pressure will be updated later before calculating the rates.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -1438,7 +1438,9 @@ namespace Opm
             well_state_copy.wellRates(this->index_of_well_)[phase]
                     = sign * ws.well_potentials[phase];
         }
-        well_copy.calculateExplicitQuantities(ebosSimulator, well_state_copy, deferred_logger);
+        well_copy.updatePrimaryVariables(summary_state, well_state_copy, deferred_logger);
+        well_copy.initPrimaryVariablesEvaluation();
+        well_copy.computeAccumWell();
 
         const double dt = ebosSimulator.timeStepSize();
         const bool converged = well_copy.iterateWellEqWithControl(ebosSimulator, dt, inj_controls, prod_controls, well_state_copy, group_state, deferred_logger);


GitPaean:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -1434,7 +1434,6 @@ namespace Opm
             well_state_copy.wellRates(this->index_of_well_)[phase]
                     = sign * ws.well_potentials[phase];
         }
-        well_copy.calculateExplicitQuantities(ebosSimulator, well_state_copy, deferred_logger);
```
GitPaean, line 1437: The alternative is that the primary variables can be updated, while the connection pressure drop should not be updated since there is no consistent perforation rate here.

GitPaean: updated along this line.

GitPaean:
jenkins build this please

GitPaean:
With this PR, at least for the target case, all the following warning are gone
```
Problematic d value -2.22045e-16 obtained for well xxxx during computeConnectionDensities 
```

GitPaean:
This is an alternative solution for unfavored PR https://github.com/OPM/opm-simulators/pull/5001 to fix the same issue.

GitPaean:
> With this PR, at least for the target case, all the following warning are gone
> 
> ```
> Problematic d value -2.22045e-16 obtained for well xxxx during computeConnectionDensities 
> ```
> 
> Now I see d can actually get negative small value due to rounding error, probably also small positive ones. It reveals that those are actually from the inconsistent well rates and wellbore pressure.

Will propose a PR to fix this as a general fix to deal with this type of small values due to rounding error.

GitPaean:
More digging in shows the problem related to update the connection pressure after set `bhp` and `well rates`, 
The bhp is updated and perf_press is still the old one, p_above will show some inconsistency. 

```
 perf p_above perf_press 
0  6891.2 399.648
1  399.648 399.908
2  399.908 400.169
3  400.169 400.7
4  400.7 401.051
5  401.051 401.487
6  401.487 402.006
7  402.006 401.578
8  401.578 401.975
9  401.975 402.567
10  402.567 402.864
11  402.864 403.162
12  403.162 403.462
```

Since the connection pressure will be updated later before calculating the rates.  One possible alternative here can be update the connection pressure before setting the bhp.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
I will check the regression tests and report back.

GitPaean:
The regression tests look fine, 
image

bska (approved):
This looks much more targeted than the earlier attempt #5001.  While it is not completely clear to me that this change is safe/correct for all callers of `computeWellRatesWithBhpIterations()` or if we might need a special case function just for the potential calculations, I nevertheless trust your judgement on the regression failures.  The code change itself looks fine.

I'll update the reference solutions and merge this.

bska:

GitPaean:
> safe/correct for all callers of computeWellRatesWithBhpIterations() or if we might need a special case function just for the potential calculations,

When we only update the bhp and well rates, without updating the perf_press and perf_rates, updating the connection properties (density mostly, while it affects the connection pressure drop) is problematic or at least prune to problems.

bska:
jenkins build this opm-tests=1096 please

bska:
jenkins build this opm-tests=1096 please

bska:
> > safe/correct for all callers of `computeWellRatesWithBhpIterations()` or if we might need a special case function just for the potential calculations,
> 
> When we only update the bhp and well rates, without updating the `perf_press` and `perf_rates`, updating the connection properties (density mostly, while it affects the connection pressure drop) is problematic or at least prune to problems.
> 
> Because we use both the bhp and `perf_press` and well rates and `perf_rates` during the calculation.

We're probably talking past each other, because I don't see how that response addresses my comment.  I realise that I'm missing some important context here, and that I should probably not review PRs that go deep into the well code.  In any case, the new reference solutions have been installed on the CI system so I'll merge this to switch to the new method.

GitPaean:
> We're probably talking past each other, because I don't see how that response addresses my comment.

I was saying the original way was wrong due to the inconsistent bhp and perf_press, and well rates and perf rates.  The way in this PR is not perfect while should be better.  There might be still space for incremental improvement to make it safer. 

> and that I should probably not review PRs that go deep into the well code.

I think the logic in the well code is more complicated that it should be and we are not  on the way to or it is hard to reduce/simplify the logic.  And I appreciate your efforts in reviewing and discussing.

PR #5016, opened 2023-11-22
fixed: also add dependency with ninja generator

akva2:
I swear last time I looked at this ninja managed to resolve the dependency on its own, but maybe I was just hallucinating or something. In any case, let us get it right now.

The simulator objects depends on the generated headers in opm-common. When doing a super build we have to explicitly declare this dependency or simulator objects might start building too early with many build threads. While optimally we'd declare the dependency on the generated headers, these are not grouped under a single target so we'd have to have a long list of headers. Instead we simply use the opmcommon library as the proxy.


akva2:
jenkins build this please

bska (approved):

PR #5018, opened 2023-11-22
[fix] Do parallel sums with int instead of bool.

blattms:
Using bool here is at least frowned upon. To be honest, I have no idea what happens underneath here if we pass a bool. In contrast to other pod types we do not associate it with a builtin type of MPI (not even sure what to use). Hence we probably create a custom type for sending and receiving. That should work. But I have no idea what will be used for summation. Hence this might have very unpredictable results.

BTW: I am debugging a case that previously crashed and now suddenly works. This seems to be the only relevant change I made in the meantime.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1924,7 +1924,7 @@ namespace Opm {
             }
         }
 
-        changed_well_to_group = comm.sum(changed_well_to_group);
+        changed_well_to_group = comm.sum(static_cast<int>(changed_well_to_group));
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

bska (approved):
I might probably consider writing
```C++
changed_x = comm.sum(static_cast<int>(...)) != 0;
```
instead to be explicit about conversion to `bool` when storing the result, but implicit conversions work too of course.

In regular code, `bool` follows the usual integral promotions so
```C++
int + bool + int
```
is well defined (`true` converts to `int(1)` and `false` converts to `int(0)`), but for MPI we do need to apply extra care.

I'll merge into master.

PR #5022, opened 2023-11-23
checking both components are there before using d = 1. - rs * rv

GitPaean:
trying to remove some not helpful debugging message, like the following. Basically, we should not try to calculate the `d` when there is only one component there. Then we do not run to this trouble. 

```
684516 Problematic d value -0.25268203386454324 obtained for well A-6H during gasOilVolumeRatio calculations with rs 744.8471490915831, rv 0.0016817974471572007 and pressure 59324439.42451076. Continue as if no dissolution (rs = 0) and vaporization (rv = 0)  for this connection.
 684517  mix_s[oil] 0 mix_s[gas] 1  b[oil]  0.2949977743744645  b[gas]  229.2656516489468 
```

Ideally, we should check whether it is two phase state also under reservoir condition when using the `d` formula, while that is more involved.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -52,7 +52,7 @@ auto dValueError(const dValue& d,
                        " during {} calculations with rs {}"
                        ", rv {} and pressure {}."
                        " Continue as if no dissolution (rs = 0) and vaporization (rv = 0) "
-                       " for this connection.", d, name, methodName, Rs, Rv, pressure);
+                       " for this connection.", Opm::getValue(d), name, methodName, Opm::getValue(Rs), Opm::getValue(Rv), Opm::getValue(pressure));
 }
 
 }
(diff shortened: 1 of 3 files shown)


GitPaean:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -2344,13 +2344,15 @@ namespace Opm
             // convert to reservoir conditions
             EvalWell cq_r_thermal(this->primary_variables_.numWellEq() + Indices::numEq, 0.);
             const unsigned activeCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::solventComponentIndex(phaseIdx));
-            const bool both_oil_gas = FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx) && FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx);
+            constexpr double epsilon = std::numeric_limits<double>::epsilon();
+            const unsigned oilCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::oilCompIdx);
+            const unsigned gasCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx);
+            const bool both_oil_gas = FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx) && FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx) &&
+                                       (std::abs(getValue(cq_s[oilCompIdx])) > epsilon && std::abs(getValue(cq_s[gasCompIdx])) > epsilon);
```
GitPaean, line 2351: This code can be problematic, will update.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

PR #5023, opened 2023-11-23
Adapt wdfac to wpimult

totto82:
depends on https://github.com/OPM/opm-common/pull/3788


bska (approved):

PR #5025, opened 2023-11-25
[bugfix] Expect singular matrices, terminate updateWellControls/nonlinear and cut timestep

blattms:
We are experiencing singular matrices when solving mulisegment wells sometimes. In that case (here during
`BlackoilModelEbos::assembleReservoir()` :arrow_left: `BlackoilModelEbos::initialLinerization()` :arrow_left:  `BlackoilModelEbos:::nonlinearIterationNewton()` an exception is thrown when updating the controls of a well.

The problem here is that this exception only happens on one process. That one goes to the catch block in `NonLinearSolverEbos::step()`, marks the nonlinear solve as failed and cuts the time step. The others move to the collective communication below. Somehow and somewhen all end up in a non-matching collective communication with different data types and we get an MPI Error that the message was truncated.

Now all processes will throw, terminate the nonlinear solver and cut the timestep to continue the simulation.

There are alternative solutions which might make (more or less?) sense:
We could in this case leave the well control as before and try to continue. Of course the matrix will remain singular and the next time we try to solve it, it fail again. So the problem might just be postponed.

~I am creating this in draft mode as I want to double check that this fixes the problems.~
Undrafted because it is tested and fixes the issue.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1916,12 +1916,19 @@ namespace Opm {
         }
         // Check wells' group constraints and communicate.
         bool changed_well_to_group = false;
-        for (const auto& well : well_container_) {
-            const auto mode = WellInterface<TypeTag>::IndividualOrGroup::Group;
-            const bool changed_well = well->updateWellControl(ebosSimulator_, mode, this->wellState(), this->groupState(), deferred_logger);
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
jenkins build this please

atgeirr:
Looks good, merging!

PR #5028, opened 2023-11-28
Update prec factory

multitalentloes:
Update the factory to be more consistent by adding smoothers/preconditioners that were only added some places.
TODO: remove synonyms? (e.i. multiple ILU versions in the factory need only have a single add creator)


opm/simulators/linalg/PreconditionerFactory_impl.hpp
```diff
@@ -162,6 +162,12 @@ struct StandardPreconditioners
         F::addCreator("ILUn", [](const O& op, const P& prm, const std::function<V()>&, std::size_t, const C& comm) {
           return createParILU(op, prm, comm, prm.get<int>("ilulevel", 0));
         });
+        F::addCreator("DuneILU", [](const O& op, const P& prm, const std::function<V()>&, std::size_t, const C& comm) {
+            const double w = prm.get<double>("relaxation", 1.0);
+            const int n = prm.get<int>("ilulevel", 0);
+            const bool resort = prm.get<bool>("resort", false);
+            return wrapBlockPreconditioner<DummyUpdatePreconditioner<Dune::SeqILU<M, V, V>>>(comm, op.getmat(), n, w, resort);
```
atgeirr, line 169: Dune::SeqILU should not be wrapped in the DummyUpdatePreconditioner, I think it needs an actual update(). I suggest creating a new wrapper class that reconstructs on update(), ideas welcome.

atgeirr: > should only the Dune SeqILU use it?

I think so, any preconditioner that stores anything other than a reference to the matrix will need a nontrivial update(). There is another, SeqILDL, that also seems to need this. I am not familiar with that one.

multitalentloes:
I have added some new preconditioners that were only partially added to the factory, run clang format, and created a new wrapper class for dummypreconditioners so that preconditioners needing to be rebuilt on update can made easily.

multitalentloes:
Jenkins build this please

opm/simulators/linalg/PreconditionerWithUpdate.hpp
```diff
@@ -77,14 +77,70 @@ class DummyUpdatePreconditioner : public PreconditionerWithUpdate<typename Origi
...
-
 template <class OriginalPreconditioner, class... Args>
 std::shared_ptr<DummyUpdatePreconditioner<OriginalPreconditioner>>
-wrapPreconditioner(Args&&... args)
+getDummyUpdateWrapper(Args&&... args)
 {
     return std::make_shared<DummyUpdatePreconditioner<OriginalPreconditioner>>(std::forward<Args>(args)...);
 }
 
+template <class OriginalPreconditioner, class MatrixPtr>
```
atgeirr, line 87: Why do you stora a matrix pointer, and not a const reference?

opm/simulators/linalg/PreconditionerFactory_impl.hpp
```diff
@@ -404,115 +471,129 @@ struct StandardPreconditioners<Operator,Dune::Amg::SequentialInformation>
...
                 }
             });
             F::addCreator("famg", [](const O& op, const P& prm, const std::function<V()>&, std::size_t) {
-                auto crit = AMGHelper<O,C,M,V>::criterion(prm);
+                auto crit = AMGHelper<O, C, M, V>::criterion(prm);
                 Dune::Amg::Parameters parms;
                 parms.setNoPreSmoothSteps(1);
                 parms.setNoPostSmoothSteps(1);
-                return wrapPreconditioner<Dune::Amg::FastAMG<O, V>>(op, crit, parms);
+                return getDummyUpdateWrapper<Dune::Amg::FastAMG<O, V>>(op, crit, parms);
```
atgeirr, line 526: The FastAMG should probably not use the dummy wrapper. Changing this to the new rebuild-on-update wrapper would be a bugfix I think. Although the interface is not the same, so it would probably not compile...

Postponing this fix until we have a more generic rebuild wrapper.

multitalentloes: Yes, this will be more easily fixed when we have the more generic rebuildOnUpdate class

opm/simulators/linalg/PreconditionerWithUpdate.hpp
multitalentloes: changed.

multitalentloes:
Jenkins build this please

opm/simulators/linalg/PreconditionerWithUpdate.hpp
atgeirr: Name of type is still `MatrixPtr`... (as well as variable name)

multitalentloes: Template name and variable name now updated.

atgeirr:
jenkins build this please

PR #5029, opened 2023-11-28
CartesianCoordinateLevel added for ALUGrid

aritorto:
For a cell on the leaf grid view, cartesianIndexMapper.cartesianCoordinate( cell idx, ...) computes the ijk index of the parent cell or the equivalent cell on level zero, for CpGrid with LGRs. For every level, included level 0, a new method to compute the ijk index, given the index cell on a certain level and its level, is included. See OPM/opm-grid#696. 

Such a method has been added also for ALUGrid and is needed in RelpermDiagnostics.cpp, to create the string of a cell index, from level zero.  To unify calls between the different grid types (CpGrid, Polyhedral, ALU).


blattms:
thanks a lot.

PR #5030, opened 2023-11-28
fixed : make (MSW|Std) well copyable again

akva2:
the reference member in the base classes would point to the original instance, not the new instance

### DIFF
--- a/opm/simulators/wells/MultisegmentWell.hpp
+++ b/opm/simulators/wells/MultisegmentWell.hpp
@@ -80,6 +80,8 @@ namespace Opm
                          const int index_of_well,
                          const std::vector<PerforationData>& perf_data);
 
+        MultisegmentWell(const MultisegmentWell& well);
+
         virtual void init(const PhaseUsage* phase_usage_arg,
                           const std::vector<double>& depth_arg,
                           const double gravity_arg,
--- a/opm/simulators/wells/MultisegmentWellEquations.cpp
+++ b/opm/simulators/wells/MultisegmentWellEquations.cpp
@@ -52,6 +52,18 @@ MultisegmentWellEquations(const MultisegmentWellGeneric<Scalar>& well)
 }
 
 template<class Scalar, int numWellEq, int numEq>
+MultisegmentWellEquations<Scalar,numWellEq,numEq>::
+MultisegmentWellEquations(const MultisegmentWellGeneric<Scalar>& well,
+                          const MultisegmentWellEquations& eqs)
+    : duneB_(eqs.duneB_)
+    , duneC_(eqs.duneC_)
+    , duneD_(eqs.duneD_)
+    , resWell_(eqs.resWell_)
+    , well_(well)
+{
+}
+
+template<class Scalar, int numWellEq, int numEq>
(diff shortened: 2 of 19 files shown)


akva2:
this seems to influence quite a few tests unfortunately.

atgeirr:
I think this is an important bugfix, so we should understand why the test cases fail. I am a bit surprised, I would expect only cases that include some dependency on well potentials (i.e. having group controlled wells) to be affected! @steink @GitPaean or @totto82 maybe one of you can aid @akva2 investigating?

totto82:
> I would expect only cases that include some dependency on well potentials (i.e. having group controlled wells) to be affected!

Most of the test failures are due to small changes in the output of well potentials. I haven't systematically looked at all cases, but the once I have looked at seems reasonable.

atgeirr:
> Most of the test failures are due to small changes in the output of well potentials. I haven't systematically looked at all cases, but the once I have looked at seems reasonable.

Then it is not so strange! I would expect the results to be more correct, do you see any cases with significant change to the potential output?

With the current bug, even adding *output of* well potentials (triggering their calculation where it was not done before) might change the simulation results if it mutates the incorrectly copied base object, I do not know if this actually happens though.

steink:
I have the same impression, I haven't seen anything worrying. I am a bit surprised that it affects this many cases though, but there's probably some effect I haven't thought of. I haven't testet, but with this, I think we will see less convergence failures for potential calculations for stopped wells, since in that case (in master) convergence tollerances for the (bhp/thp) control eq will be of type rate.

steink:
So probably no worries here, but I guess we should make sure we don't miss anything. @GitPaean and @totto82 , could we just devide the 60 failed tests between us to confirm (Kai 1-20, Tor Harald 21-40, Stein 41-60)?

GitPaean:
> So probably no worries here, but I guess we should make sure we don't miss anything. @GitPaean and @totto82 , could we just devide the 60 failed tests between us to confirm (Kai 1-20, Tor Harald 21-40, Stein 41-60)?

How about let me go through all of them and report back.  Might delegate you guys with ones need more detailed investigation. 

Probably tomorrow.

GitPaean:
As promised, reporting back the checking of the regression failures. 

image
image


Will provide some information regarding the ones need more detailed investigations.

GitPaean:
Checking regression failures with big scale is quite expensive.  It is good if we can have some more automatic way to help. 

And also, because it is quite some efforts to check one by one, and some update of the reference might be done without careful checking from time to time.

GitPaean:
> Looks like there are some issues that needs fixing here. Seems to be missing a calculateExplicitQuantities in the potential calculations for STW. Hence, with the PR, there is zero p-drop along the well for the well_copy. In master it's non-zero since this->connections_.pressure_diff(perf) corresponds to the p-drop of the original well.

Thanks for looking into this. It look reasonable since the majority of the failing tests (if not tall) are related to STW.

GitPaean:
The regression failures look fine. Thanks for updating the PR. I will have a look by tomorrow.

GitPaean:
With the new failure_report system in Jenkins, we should revisit the PR again.

akva2:
rebased.

GitPaean:

PR #5032, opened 2023-11-30
Catch exceptions from well testing 

vkip:
Numerical problems during well testing (often: inf/nan in MSW matrices) will currently be caught at a higher level, causing time step chops in serial mode and possibly MPI communication errors in parallel. 

With this change wells that experience such problems during testing will simply not open, and MPI errors (mainly experienced during testing of https://github.com/OPM/opm-simulators/pull/4986) are no longer present.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -586,7 +586,7 @@ namespace Opm {
 
             try {
                 well->wellTesting(ebosSimulator_, simulationTime, this->wellState(), this->groupState(), wellTestState(), deferred_logger);
-            } catch (std::runtime_error& e) {
+            } catch (std::exception& e) {
                 const std::string msg = fmt::format("Exception during testing of well: {}. The well will not open.\n Exception message: {}", wellEcl.name(), e.what());
                 deferred_logger.warning("WELL_TESTING_FAILED", msg);
             }
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -584,7 +584,12 @@ namespace Opm {
                 well->setPrevSurfaceRates(this->wellState(), this->prevWellState());
             }
 
-            well->wellTesting(ebosSimulator_, simulationTime, this->wellState(), this->groupState(), wellTestState(), deferred_logger);
+            try {
+                well->wellTesting(ebosSimulator_, simulationTime, this->wellState(), this->groupState(), wellTestState(), deferred_logger);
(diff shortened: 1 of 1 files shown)


opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -584,7 +584,12 @@ namespace Opm {
                 well->setPrevSurfaceRates(this->wellState(), this->prevWellState());
             }
 
-            well->wellTesting(ebosSimulator_, simulationTime, this->wellState(), this->groupState(), wellTestState(), deferred_logger);
+            try {
+                well->wellTesting(ebosSimulator_, simulationTime, this->wellState(), this->groupState(), wellTestState(), deferred_logger);
+            } catch (std::runtime_error& e) {
```
atgeirr, line 589: Any logic_error thrown will be uncaught, maybe change to catch std::exception?

vkip: Good point, will update!

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -586,7 +586,7 @@ namespace Opm {
 
             try {
                 well->wellTesting(ebosSimulator_, simulationTime, this->wellState(), this->groupState(), wellTestState(), deferred_logger);
-            } catch (std::runtime_error& e) {
+            } catch (std::exception& e) {
```
bska, line 589: Sorry for the delayed response.  This should probably be a reference to `const`, i.e.,
```C++
catch (const std::exception& e)
````
instead.

bska:
jenkins build this please

bska (approved):
Thank you for the updates.  I think this makes sense and it looks good to me now.  If it turns out to have some unforeseen effects we can revise the approach in follow-up work.  For now I'll merge into master.

blattms:
Question to the reviewers (might be academic):
Why should the simulator continue if there is a std::logic_error. I would normally assume that such error is unrecoverable and the simulator should abort. What is different here.

Not sure whether this can happen here, but the same holds for std::bad_alloc.

GitPaean:
I guess you might misunderstand the comments. I think the comment is about `catch (std::runtime_error& e)` can not catch `std::logic_error`?

atgeirr:
> Why should the simulator continue if there is a std::logic_error. I would normally assume that such error is unrecoverable and the simulator should abort. What is different here?

Your point is correct, but we still need to avoid a deadlock in parallel. The catching done here is not to recover from the error, but to propagate it to all processes. A more sophisticated handling that propagates the thrown exception types may be possible, but would require that we handle situations where different processes throw different exceptions, so it is not clear that it is worth it.

PR #5034, opened 2023-12-04
fixed: grab list of FIP regions from FieldPropsManager

akva2:
not from SummaryConfig. the latter only holds regions with summary keywords.

Downstream of https://github.com/OPM/opm-common/pull/3813


bska (approved):

PR #5036, opened 2023-12-04
Communicate exceptions during WellModel::updateAndCommunicate

blattms:
For multi segment well the underlying call to
MultisegmentWell::updateWellStateWithTarget (at least if updateWellStateWithTHPTargetProd is called for a producer under thp control) might throw as there might be a singular matrix during the solve needed in MultisegmentWell::iterateWellEqWithControl.

Previously, if that happened then the MPI process where it happened would stop the nonlinear iteration as failed and try with a chopped time step. The others might go one with the current time step and we would see MPI errors about truncated messages.

Now we communicate any exception happening during this part of WellModel::updateAndCommunicate and all processes will stop the nonlinear iteration as failed and chop the time step.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -2144,10 +2144,16 @@ namespace Opm {
                          DeferredLogger& deferred_logger)
     {
         updateAndCommunicateGroupData(reportStepIdx, iterationIdx);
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please

blattms:
This is tested now. It was confirmed that it fixes serious issues.

akva2 (approved):

PR #5041, opened 2023-12-05
Guard against nans from certain expressions in the valve models

steink:
Most power expressions in `pressureDropSpiralICD` and `pressureDropAutoICD` will return nan if given a negative base number. Even though these numbers should be positive (e.g., denisty, phase-fractions etc.), there is currently no guarantee for this (happened for a model I was looking at).  Although this PR just provides a quick-fix and does't attack the root of the problem (which seems to be during initialization), I think it doesn't hurt to have some safety-measures when dealing with general power expressions.

Hopefully, will not break any tests (_famous last words_).

### DIFF
--- a/opm/simulators/wells/MultisegmentWellSegments.cpp
+++ b/opm/simulators/wells/MultisegmentWellSegments.cpp
@@ -637,8 +637,9 @@ pressureDropSpiralICD(const int seg,
     using MathTool = MathToolbox<EvalWell>;
 
     const double density_cali = sicd.densityCalibration();
-    const EvalWell temp_value1 = MathTool::pow(density / density_cali, 0.75);
-    const EvalWell temp_value2 = MathTool::pow(mixture_viscosity / viscosity_cali, 0.25);
+    // make sure we don't pass negative base to powers
+    const EvalWell temp_value1 = density > 0.0 ? MathTool::pow(density / density_cali, 0.75) : 0.0;
+    const EvalWell temp_value2 = mixture_viscosity > 0.0 ? MathTool::pow(mixture_viscosity / viscosity_cali, 0.25) : 0.0;
 
     // formulation before 2016, base_strength is used
     // const double base_strength = sicd.strength() / density_cali;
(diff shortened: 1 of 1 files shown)


GitPaean:
jenkins build this please

GitPaean:
> Even though these numbers should be positive (e.g., denisty, phase-fractions etc.), 

Just for discussion, not necessarily done in this PR. Should we do something related to these types quantities (non-positive values for density, fraction, FVF or viscosity). Those values will likely cause problem elsewhere also. Maybe at least a warning message?

phase fractions maybe should be fixed, other quantities from the PVT table might be more involved.

GitPaean:
It only touches one AICD case, [mpi.compareECLFiles_flow+BASE_MSW_WSEGAICD](https://ci.opm-project.org/job/opm-simulators-PR-builder/5649/testReport/junit/(root)/mpi/compareECLFiles_flow_BASE_MSW_WSEGAICD/)

opm/simulators/wells/MultisegmentWellSegments.cpp
```diff
@@ -735,13 +736,18 @@ pressureDropAutoICD(const int seg,
     }
 
     using MathTool = MathToolbox<EvalWell>;
-    const EvalWell mixture_viscosity = MathTool::pow(water_fraction, aicd.waterViscExponent()) * water_viscosity
-                                     + MathTool::pow(oil_fraction, aicd.oilViscExponent()) * oil_viscosity
-                                     + MathTool::pow(gas_fraction, aicd.gasViscExponent()) * gas_viscosity;
+    // make sure we don't pass negative base to powers
+    auto safe_pow = [](const auto a, const double b) {
```
bska, line 740: Tiny nit: This definition will copy `a` into the lambda.  That's never a problem if `a` turns out to be a scalar type like `double` but **could** become a concern if the underlying type is `EvalWell` or similar.  If you want to, could probably write this as
```C++
auto safe_pow = [](const auto& a, const double b) { ... };
```
with a "reference-to-`const`" parameter instead.  On the other hand I don't know that it matters too much.

opm/simulators/wells/MultisegmentWellSegments.cpp
```diff
@@ -637,8 +637,9 @@ pressureDropSpiralICD(const int seg,
     using MathTool = MathToolbox<EvalWell>;
 
     const double density_cali = sicd.densityCalibration();
-    const EvalWell temp_value1 = MathTool::pow(density / density_cali, 0.75);
-    const EvalWell temp_value2 = MathTool::pow(mixture_viscosity / viscosity_cali, 0.25);
+    // make sure we don't pass negative base to powers
+    const EvalWell temp_value1 = density > 0.0 ? MathTool::pow(density / density_cali, 0.75) : 0.0;
+    const EvalWell temp_value2 = mixture_viscosity > 0.0 ? MathTool::pow(mixture_viscosity / viscosity_cali, 0.25) : 0.0;
```
bska, line 642: I take it we're guaranteed that the `*_cali` values are always strictly positive?  If not, is there a check somewhere else?  If not, should we add a check here?

bska (commented):
I agree that it makes sense to guard against NaNs here, although a more comprehensive solution might look different in the end.

opm/simulators/wells/MultisegmentWellSegments.cpp
steink: Thanks, done.

bska:
jenkins build this please

steink:
> It only touches one AICD case, [mpi.compareECLFiles_flow+BASE_MSW_WSEGAICD](https://ci.opm-project.org/job/opm-simulators-PR-builder/5649/testReport/junit/(root)/mpi/compareECLFiles_flow_BASE_MSW_WSEGAICD/)

Differences are very small so this OK. Interestingly, slightly negative numbers also enter the power-expressions for this test-case, but they don't cause problems in master since the corresponding exponents are defaulted to 1 (this also explains why there is a difference at all between master and pr).

bska:
>  slightly negative numbers also enter the power-expressions for this test-case, but they don't cause problems in master since the corresponding exponents are defaulted to 1

"Let's switch to using complex numbers ‼️ ❓ ‼️ "

Seriously though, thanks for looking into the differences.  Unless you have any further comments or changes here, I suggest we update the reference solution for this test case and then merge this PR.

bska:
jenkins build this opm-tests=1101 please

bska:
The new reference solutions have been installed on the CI system.  I'll merge this now to fix this problem.  Thanks a lot for the excellent fix.

PR #5042, opened 2023-12-06
changed: throw early in simulator serializer

### LINKED ISSUE #5037: Message about missing HDF5 support (with --load-step) should be issued during startup of the simulation
If I e.g. request restart for report step 770 then it will take a long time to notice that my flow build does not support --load-step

akva2:
if HDF5 input/output is requested but no support is available we throw early instead of just logging an error when save/load is to be performed

closes #5037


akva2:
jenkins build this please

atgeirr:
Good, merging.

PR #5043, opened 2023-12-06
Only output interior cells

totto82:
This fixes parallel output of FLOWS and FLORES

It should also give a small speed up of the output part in parallel.

### DIFF
--- a/ebos/damariswriter.hh
+++ b/ebos/damariswriter.hh
@@ -44,6 +44,7 @@
 
 #include <opm/simulators/utils/DeferredLoggingErrorHelpers.hpp>
 #include <opm/simulators/utils/ParallelRestart.hpp>
+#include <opm/simulators/flow/countGlobalCells.hpp>
 
 #include <opm/common/OpmLog/OpmLog.hpp>
 
@@ -516,7 +517,7 @@ private:
         }
 
         const auto& gridView = simulator_.vanguard().gridView();
-        const int num_interior = Opm::detail::
+        const int num_interior = detail::
             countLocalInteriorCellsGridView(gridView);
         const bool log = this->collectToIORank_.isIORank();
 
--- a/ebos/damariswriter.hh
+++ b/ebos/damariswriter.hh
@@ -516,17 +516,18 @@ private:
         }
 
         const auto& gridView = simulator_.vanguard().gridView();
-        const int numElements = gridView.size(/*codim=*/0);
+        const int num_interior = Opm::detail::
+            countLocalInteriorCellsGridView(gridView);
         const bool log = this->collectToIORank_.isIORank();
 
-        damarisOutputModule_->allocBuffers(numElements, reportStepNum,
+        damarisOutputModule_->allocBuffers(num_interior, reportStepNum,
                                       isSubStep, log, /*isRestart*/ false);
 
         ElementContext elemCtx(simulator_);
         OPM_BEGIN_PARALLEL_TRY_CATCH();
         {
         OPM_TIMEBLOCK(prepareCellBasedData);
-        for (const auto& elem : elements(gridView)) {
+        for (const auto& elem : elements(gridView, Dune::Partitions::interior)) {
             elemCtx.updatePrimaryStencil(elem);
             elemCtx.updatePrimaryIntensiveQuantities(/*timeIdx=*/0);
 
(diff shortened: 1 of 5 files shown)


totto82:
jenkins build this please

ebos/eclgenericoutputblackoilmodule.cc
```diff
@@ -715,8 +715,13 @@ regionSum(const ScalarBuffer& property,
         if (property.empty())
             return totals;
 
-        assert(regionId.size() == property.size());
-        for (std::size_t j = 0; j < regionId.size(); ++j) {
+        // the regionId contains the ghost cells
+        // the property does not contain the ghostcells
+        // This code assumes that that the ghostcells are
+        // added after the interior cells
+        // Is this a valid assumption?
```
bska, line 722: The assumption (interior cells stored before ghost cells) holds unless the user changes the `OwnerCellsFirst` runtime parameter from its default state of `true`.

bska (commented):
It looks like this does the right thing in the [default](https://github.com/OPM/opm-simulators/pull/2512) case of `OwnerCellsFirst = true`.  Whether or not we want to continue to support `OwnerCellsFirst = false` is maybe a matter of debate.

ebos/eclgenericoutputblackoilmodule.cc
atgeirr: I think testing and throwing if not the case may be in order then? Or even removing the OwnerCellsFirst runtime option.

blattms: Just remove the option.

blattms: I correct my statement: Maybe someone should first test what happens with grids apart from CpGrid.

totto82: The linear solver has
```
            const bool ownersFirst = EWOMS_GET_PARAM(TypeTag, bool, OwnerCellsFirst);
            if (!ownersFirst) {
                const std::string msg = "The linear solver no longer supports --owner-cells-first=false.";
                if (on_io_rank) {
                    OpmLog::error(msg);
                }
                OPM_THROW_NOLOG(std::runtime_error, msg);
            }
```
I can add something similar to the output code.

totto82:
jenkins build this please

totto82:
Can we merge this now?

ebos/eclwriter.hh
```diff
@@ -329,15 +331,16 @@ public:
         }
 
         const auto& gridView = simulator_.vanguard().gridView();
-        const int numElements = gridView.size(/*codim=*/0);
+        const int num_interior = Opm::detail::
+            countLocalInteriorCellsGridView(gridView);
```
bska, line 335: We should generally not use the `Opm::` namespace qualifier when we're inside that namespace.  We should either just drop that qualifier (i.e., `detail::countLocalInteriorCellsGridView()`, preferable) or use `::Opm::` instead (`::Opm::detail::count*()`).

ebos/ecloutputblackoilmodule.hh
```diff
@@ -173,6 +173,16 @@ public:
...
+
+        const bool ownersFirst = EWOMS_GET_PARAM(TypeTag, bool, OwnerCellsFirst);
+        const bool on_io_rank = (simulator.gridView().comm().rank() == 0);
+        if (!ownersFirst) {
+            const std::string msg = "The output code does not support --owner-cells-first=false.";
+            if (on_io_rank) {
+                OpmLog::error(msg);
+            }
+            OPM_THROW_NOLOG(std::runtime_error, msg);
+        }
```
bska, line 185: I think I'd probably write this as
```C++
if (! EWOMS_GET_PARAM(TypeTag, bool, OwnerCellsFirst)) {
    const std::string msg = ...;
    if (collectToIORank.isIORank()) {
        OpmLog::error(msg);
    }
    OPM_THROW_NOLOG(std::runtime_error, msg);
}
```
instead.

ebos/damariswriter.hh
```diff
@@ -323,17 +323,18 @@ private:
         }
 
         const auto& gridView = simulator_.vanguard().gridView();
-        const int numElements = gridView.size(/*codim=*/0);
+        const int num_interior = Opm::detail::
+            countLocalInteriorCellsGridView(gridView);
```
bska, line 327: Please avoid using the `Opm::` namespace qualifier when inside that namespace.

bska (commented):
> Can we merge this now?

Just a couple of minor nits remaining, I think.

totto82:
jenkins build this please

totto82:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  This looks good to me now and I'll merge into master.  At some point we should probably just remove the `OwnerCellsFirst`  option.

PR #5046, opened 2023-12-07
[fix,MPI] Handle exception during assembleWellEqWithoutIteration correctly

blattms:
For this particular model WetGasPVT::saturationPressure did throw because convergence in the newton solver is not reached in 20 iterations. Unfortunately, the exception was only seen on one MPI rank and the others continued.

With this commit we communicate the problem and throw on all MPI processes. Time step will be cut as a result.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1518,7 +1518,7 @@ namespace Opm {
     {
         // We make sure that all processes throw in case there is an exception
         // on one of them (WetGasPvt::saturationPressure might throw if not converged)
-        OPM_BEGIN_PARALLEL_TRY_CATCH()
+        OPM_BEGIN_PARALLEL_TRY_CATCH();
 
         for (auto& well: well_container_) {
             well->assembleWellEqWithoutIteration(ebosSimulator_, dt, this->wellState(), this->groupState(),


blattms:
There are  a few more assembleWellEq* methods. Not sure whether there exceptions for individual wells might be throw also.  Maybe we should parallel try-catch statements there too?

A tangential note: In the newton solver when computing the [*PVT::computeSaturation](https://github.com/OPM/opm-common/blob/master/opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp#L287-L336) we treat convergence issues quite differently:
- More than 20 iterations used: throw exception
- derivative 0: return initial guess
- negative pressure encountered twice: return 0

blattms:
jenkins build this please

blattms:
jenkins build this please

atgeirr:
Good, will merge. The almost-secret (at least obscure) existence of a Newton loop deep in the fluid system is not very obvious!

I agree it is unfortunate that the message is completely lost. Basically, the existing parallel macro helper assumes that any important information will be put in the DeferredLogger, so the easiest would be to insert at catch() at the lowest level where that is available (I really do not want to push it further down all the calls of the material system...!) and put the message in the logger there.

Alternatively, we can modify the existing macro to add the exception message to the logger before collecting, but that would give lots of double messages for errors we handle well now. Maybe check if an error has been added to the logger and only add if empty?

It is also clear to me upon reading them that the existing macros give the impression of reasonable forwarding of exceptions, but without doing it. If there is an exception with no corresponding logger entry, that message will get lost *unless on rank 0*, I think this is clear from the code. After the OPM_PARALLEL_CATCH_CLAUSE then obptc_exc_msg will only be nonempty on the rank that threw the exception, there is no gathering here.

So: if we want to fix this problem more generally, we need to add a facility for gathering of error strings, choosing the correct one based on some criterion (i.e. in the presence of multiple exceptions, which is the most serious we want to rethrow). Should not be too hard, but it comes down to (once again) writing code that collects strings of arbitrary length using MPI, which by now we should have put in a library... If you want to do this, I suggest starting by extracting some of the code from gatherDeferredLogger, or maybe we already have this somewhere else? @akva2 do you have suggestions?

PR #5050, opened 2023-12-08
support source term from deck (SOURCE)

totto82:
Depends on https://github.com/OPM/opm-common/pull/3835


totto82:
There seem to be some scaling issue for the thermal part that needs to fix before this can be merged.

atgeirr:
> There seem to be some scaling issue for the thermal part that needs to fix before this can be merged.

What is the issue? I think we should try to get this in soon.

totto82:
> What is the issue? I think we should try to get this in soon.

Sorry I forgot to comment that I solved the issue. So yes from my point of view this can be merged.

PR #5057, opened 2023-12-11
Function to lookup origin cell indices added in thermal field props.

aritorto:
Two member functions have been added in EclProblem, to assign field properties of type double or int, unsigened int, ..., on the leaf grid view. 

For CpGrid with local grid refinement, the field property is inherited from the origin cell (parent or equivalent cell) in level zero. 

These new members: getFieldPropDoubleAssigner_ and getFieldPropIntTypeAssigner_, are used here to get thermal field properties on the leaf grid view, and in OPM/opm-simulator#5060, to get material field properties on the leaf grid view.

### DIFF
--- a/ebos/eclgenericproblem.hh
+++ b/ebos/eclgenericproblem.hh
@@ -377,12 +377,6 @@ protected:
     using LookUpData = Opm::LookUpData<Grid,GridView>;
     const LookUpData lookUpData_;
 
-    auto getLookUpData(unsigned elemIdx) const
-    {
-        using GridType = std::remove_cv_t< typename std::remove_reference<decltype(gridView_.grid())>::type>;
-        return lookUpData_.template getFieldPropIdx<GridType>(elemIdx);
-    }
-
 private:
     template<class T>
     void updateNum(const std::string& name, std::vector<T>& numbers, std::size_t num_regions);
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1955,6 +1955,38 @@ protected:
         }
     }
 
+    // \brief Function to assign field properties of type double, on the leaf grid view.
+    //
+    // For CpGrid with local grid refinement, the field property of a cell on the leaf
+    // is inherited from its parent or equivalent (when has no parent) cell on level zero.
+    std::function<std::vector<double>(const FieldPropsManager&, const std::string&, const unsigned int&)>
+    getFieldPropDoubleOnLeafAssigner_()
+    {
+        const auto& lookup = this->lookUpData_;
+        return [&lookup](const FieldPropsManager& fieldPropManager, const std::string& propString,
+                         const unsigned int& numElems)
+        {
+            return lookup.assignFieldPropsDoubleOnLeaf(fieldPropManager, propString, numElems);
+        };
+    }
+
+    // \brief Function to assign field properties of type int, unsigned int, ..., on the leaf grid view.
+    //
+    // For CpGrid with local grid refinement, the field property of a cell on the leaf
+    // is inherited from its parent or equivalent (when has no parent) cell on level zero.
+    template<typename IntType>
+    std::function<std::vector<IntType>(const FieldPropsManager&, const std::string&, const unsigned int&, bool)>
+    getFieldPropIntTypeOnLeafAssigner_()
+    {
+        const auto& lookup = this->lookUpData_;
+        return [&lookup](const FieldPropsManager& fieldPropManager, const std::string& propString,
+                         const unsigned int& numElems, bool needsTranslation)
+        {
+            return lookup.template assignFieldPropsIntOnLeaf<IntType>(fieldPropManager, propString,
+                                                                      numElems, needsTranslation);
+        };
+    }
+
     void readMaterialParameters_()
     {
         OPM_TIMEBLOCK(readMaterialParameters);
@@ -1999,7 +2031,9 @@ protected:
 
             // fluid-matrix interactions (saturation functions; relperm/capillary pressure)
             thermalLawManager_ = std::make_shared<EclThermalLawManager>();
-            thermalLawManager_->initParamsForElements(eclState, this->model().numGridDof());
+            thermalLawManager_->initParamsForElements(eclState, this->model().numGridDof(),
+                                                      this-> getFieldPropDoubleOnLeafAssigner_(),
+                                                      this-> template getFieldPropIntTypeOnLeafAssigner_<unsigned int>());
         }
     }
 


blattms:
I am adding a reference to the other PR, to make handling this  a bit easier:

This is downstream of and needs a merge of OPM/opm-common#3840

ebos/eclgenericproblem.hh
```diff
@@ -382,8 +382,22 @@ protected:
...
+        return lookUpData_.assignFieldPropsDoubleOnLeaf(fieldPropManager, propString, numElems);
+    }
+
+    std::vector<unsigned int> getAssignIntOnLeaf(const FieldPropsManager& fieldPropManager,
+                                                 const std::string& propString,
+                                                 const unsigned int& numElems, bool needsTranslation) const
+    {
+        return lookUpData_.template assignFieldPropsIntOnLeaf<unsigned int>(fieldPropManager, propString, numElems,
+                                                                            needsTranslation);
+    }
```
blattms, line 398: Wouldn't it be possibel to skip these functions and do the necessary calls in the lambdas below (you are capturing this there), e.g.:
```c++
auto getDoubleOnLeaf = [this](const FieldPropsManager& fieldPropManager, const std::string& propString,
                              const unsigned int& numElems)
{
       return this->lookUpData_.assignFieldPropsDoubleOnLeaf(fieldPropManager, propString, numElems);
};
```
Or am I missing something (again)?

ebos/eclgenericproblem.hh
```diff
@@ -382,8 +382,22 @@ protected:
...
 
-private:
+    private:
```
blattms, line 400: Would be nice to get rid off this white space change.

ebos/eclproblem.hh
```diff
@@ -1999,7 +1999,21 @@ protected:
 
             // fluid-matrix interactions (saturation functions; relperm/capillary pressure)
             thermalLawManager_ = std::make_shared<EclThermalLawManager>();
-            thermalLawManager_->initParamsForElements(eclState, this->model().numGridDof());
+            std::function<std::vector<double>(const FieldPropsManager&, const std::string&, const unsigned int&)>
+            getDoubleOnLeaf = [this](const FieldPropsManager& fieldPropManager, const std::string& propString,
+                                  const unsigned int& numElems)
```
blattms, line 2004: Please fix indentation.

ebos/eclproblem.hh
```diff
@@ -1999,7 +1999,21 @@ protected:
...
+            {
+                return EclGenericProblem<GridView,FluidSystem,Scalar>::getAssignDoubleOnLeaf(fieldPropManager, propString, numElems);
+            };
+            std::function<std::vector<unsigned int>(const FieldPropsManager&, const std::string&, const unsigned int&, bool)>
+            getIntOnLeaf = [this](const FieldPropsManager& fieldPropManager, const std::string& propString,
+                                  const unsigned int& numElems, bool needsTranslation)
+            {
+                return EclGenericProblem<GridView,FluidSystem,Scalar>::getAssignIntOnLeaf(fieldPropManager, propString,
+                                                                                             numElems, needsTranslation);
+            };
```
blattms, line 2014: Indentation looks a bit strange. Would be nice if you would fix that,

blattms (commented):
This looks good. Just a few nitpicks.

ebos/eclproblem.hh
aritorto: It seems I forgot to correct indentation/delete-white spaces in the end. Thanks!

ebos/eclgenericproblem.hh
aritorto: EclGenericPloblem has the member lookupdata_ and it's a derived class of EclProblem. I tried to downcast "this( EclProblem)" to be able to get the lookupdata memeber, and it didn't work. (I was probably doing something wrong). 
I will check out your suggestion, thanks!

blattms: Then I really missed something. I always forget about this complicated inheritance that we have. I might need some time to think about this then.

blattms: There is multiple [inheritance](https://github.com/OPM/opm-simulators/blob/df0daa1a1656c73d9e71be4a80a027edfd907a80/ebos/eclproblem.hh#L110-L113) . Maybe that was the problem. Have you tried casting to [BaseType](https://github.com/OPM/opm-simulators/blob/df0daa1a1656c73d9e71be4a80a027edfd907a80/ebos/eclproblem.hh#L115-L118)?
```
static_cast<BaseType* const>(this)->lookUpData_....
```

blattms: I probably placed `const` wrong. :sweat_smile:

aritorto: It might work, but also need a friend declaration, or some extra method returning the data member lookupdata_, since it's declared protected. I'll check it out

blattms: Shouldn't it be accessible in the derived class if it is protected in the base class? 
We should try to prevent adding friend declarations here.

aritorto: Many thanks for your feedback, it seems to work. I'll run jenkins again

aritorto: If protected members of a base class (here EclGenericProblem) are static, then any friends or member functions of the derived class (here EclProblem) can access them. If it's not static, only through pointer, reference, or object of the derived class.  
So, you're right, it should work without that extra method getLookUpData(). I'll double-check.

In any case, here EclProblem is publically derived from EclGenericProblem, and the static_cast is "upcast" (I used the wrong term before: downcast). So, here, if I get it correctly, the

aritorto: I think I see the issue now. Once we static_cast<BaseType*> "this", we are actually upcasting, so now we are a base class thing wanting to access protected member of a derived class (and not the other way around). That could explain why we need the extra public member function returning lookUpData_ in the base class EclGenericProblem

blattms: :flushed: I think the problem was me giving advice that is nonsense. I should have thought this through. My apologies.

In the lambda, which has its own class with no connection to the EclGenericProblem class and its parent or derived classes, we cannot access anything of EclGenericProblem that is not public.

I have sent you an email with a possible way.

blattms:
Cool, this looks much better now. I am just wondering whether we really need the cast with new version.

blattms (approved):
Looks great now. Just waiting for jenkins to confirm.

ebos/eclgenericproblem.hh
aritorto: Sent for the fix-idea! Incorporated in the last version

PR #5058, opened 2023-12-12
BUGFIX. Use given temperature in the density calculations in the initialisation

totto82:


totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

bska (approved):
Thanks!  This is a good improvement and I'll merge into master once I've updated the reference solutions.  The `TEMPI` array does not really factor into this, since that array is for the "enumerated" initialisation method.

bska:

bska:
jenkins build this opm-tests=1105 please

bska:
> merged commit b9ecc2a into OPM:master 1 minute ago

Just FYI: This is a potential race condition and we may get a post-commit build failure here.  I was actually waiting for the CI system's queue to drain before merging.  When updating the reference solutions, it is important that the "opm-tests" build **runs** before the post-commit build, and there are restrictions on where those "opm-tests" builds can run.

PR #5059, opened 2023-12-12
GroupEconomicLimitsChecker: cleanup includes

akva2:
Pass a parameters as a const ref instead of value and use std::array.


akva2:
jenkins build this please

bska (approved):
Thanks.  This looks good to me and I'll merge into master.

PR #5060, opened 2023-12-12
Assign int on leaf added for material props

aritorto:
A function has been added to lookup the origin cell indices to populate material field properties on the leaf grid view.

This is downstream of and needs a merge of OPM/opm-common#3841.


blattms (approved):
Nice and small and good

PR #5068, opened 2023-12-16
PRT file updates

tskille:
Adding reservoir volume array custom FIP regions

### DIFF
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -42,7 +42,7 @@
 #include <opm/simulators/utils/DeferredLoggingErrorHelpers.hpp>
 #include <opm/simulators/utils/ParallelRestart.hpp>
 #include <opm/simulators/flow/countGlobalCells.hpp>
-
+#include <opm/simulators/timestepping/SimulatorTimer.hpp>
 
 #include <opm/common/OpmLog/OpmLog.hpp>
 
@@ -123,6 +123,7 @@ class EclWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::Grid>
     enum { enableSolvent = getPropValue<TypeTag, Properties::EnableSolvent>() };
 
 public:
+
     static void registerParameters()
     {
         EclOutputBlackOilModule<TypeTag>::registerParameters();
@@ -176,6 +177,7 @@ public:
     {
         OPM_TIMEBLOCK(evalSummaryState);
         const int reportStepNum = simulator_.episodeIndex() + 1;
+        
         /*
           The summary data is not evaluated for timestep 0, that is
           implemented with a:
@@ -193,6 +195,7 @@ public:
           "Correct" in this context means unchanged behavior, might very
           well be more correct to actually remove this if test.
         */
+        
         if (reportStepNum == 0)
             return;
 
@@ -245,18 +248,22 @@ public:
             OPM_END_PARALLEL_TRY_CATCH("Collect to I/O rank: ",
                                        this->simulator_.vanguard().grid().comm());
         }
+        
 
         std::map<std::string, double> miscSummaryData;
         std::map<std::string, std::vector<double>> regionData;
         Inplace inplace;
+        
         {
             OPM_TIMEBLOCK(outputFipLogAndFipresvLog);
-            inplace = eclOutputModule_->outputFipLog(miscSummaryData, regionData, reportStepNum,
-                                                     isSubStep, simulator_.gridView().comm());
-            eclOutputModule_->outputFipresvLog(miscSummaryData, regionData, reportStepNum,
-                                               isSubStep, simulator_.gridView().comm());
-        }
 
+            inplace = eclOutputModule_->calc_inplace(miscSummaryData, regionData, simulator_.gridView().comm());
+            
+            if (this->collectToIORank_.isIORank()){
+                inplace_ = inplace;
+            }
+        }
+        
         // Add TCPU
         if (totalCpuTime != 0.0) {
             miscSummaryData["TCPU"] = totalCpuTime;
@@ -309,16 +316,6 @@ public:
                               this->summaryState(),
                               this->udqState());
         }
-
-        if (! isSubStep) {
-            OPM_TIMEBLOCK(outputProdInjLogs);
-
-            eclOutputModule_->outputProdLog(reportStepNum);
-            eclOutputModule_->outputInjLog(reportStepNum);
-            eclOutputModule_->outputCumLog(reportStepNum);
-
-            OpmLog::note("");   // Blank line after all reports.
-        }
     }
 
     //! \brief Writes the initial FIP report as configured in RPTSOL.
@@ -352,14 +349,21 @@ public:
         Inplace inplace;
         {
             OPM_TIMEBLOCK(outputFipLogAndFipresvLog);
-            inplace = eclOutputModule_->outputFipLog(miscSummaryData, regionData, 0,
+            
+            boost::posix_time::ptime start_time = boost::posix_time::from_time_t(simulator_.vanguard().schedule().getStartTime());
+
+            inplace = eclOutputModule_->calc_inplace(miscSummaryData, regionData, simulator_.gridView().comm());
+
+            if (this->collectToIORank_.isIORank()){
+                inplace_ = inplace;
+                
+                eclOutputModule_->outputFipAndResvLog(inplace_, 0, 0.0, start_time, 
                                                      false, simulator_.gridView().comm());
-            eclOutputModule_->outputFipresvLog(miscSummaryData, regionData, 0,
-                                               false, simulator_.gridView().comm());
+            }
         }
     }
 
-    void writeOutput(data::Solution&& localCellData, bool isSubStep)
+    void writeOutput(data::Solution&& localCellData, const SimulatorTimer& timer, bool isSubStep)
     {
         OPM_TIMEBLOCK(writeOutput);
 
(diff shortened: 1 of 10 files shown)


tskille:
This PR has been thoroughly tested in Equinor with internal workflows that depend on using simulation results written to the PRT file from reservoir simulations. Results are good and I now believe that this PR is ready for review.

ebos/eclwriter.hh
```diff
@@ -243,18 +250,23 @@ public:
...
             OPM_TIMEBLOCK(outputFipLogAndFipresvLog);
-            inplace = eclOutputModule_->outputFipLog(miscSummaryData, regionData, reportStepNum,
-                                                     isSubStep, simulator_.gridView().comm());
-            eclOutputModule_->outputFipresvLog(miscSummaryData, regionData, reportStepNum,
-                                               isSubStep, simulator_.gridView().comm());
-        }
 
+            inplace = eclOutputModule_->calc_inplace(miscSummaryData, regionData, simulator_.gridView().comm());
+            
+            if (simulator_.gridView().comm().rank() == 0){
```
bska, line 265: I think I'd prefer we write this check as `collectToIORank_.isIORank()` instead.

ebos/eclwriter.hh
```diff
@@ -243,18 +250,23 @@ public:
             OPM_END_PARALLEL_TRY_CATCH("Collect to I/O rank: ",
                                        this->simulator_.vanguard().grid().comm());
         }
+        
 
         std::map<std::string, double> miscSummaryData;
         std::map<std::string, std::vector<double>> regionData;
         Inplace inplace;
+        
+        if (!isSubStep)
```
bska, line 259: I think you need to run this calculation even for substeps.  Otherwise, we'll risk breaking summary vectors like `FOIP` and `FOE`.

ebos/eclwriter.hh
```diff
@@ -349,14 +351,27 @@ public:
         Inplace inplace;
         {
             OPM_TIMEBLOCK(outputFipLogAndFipresvLog);
-            inplace = eclOutputModule_->outputFipLog(miscSummaryData, regionData, 0,
+            
+            const std::string version = moduleVersionName();
+
+            auto es_ = simulator_.vanguard().eclState();
```
bska, line 357: This `es_` object appears to be unused.  Maybe we can just remove it?

ebos/eclwriter.hh
```diff
@@ -349,14 +351,27 @@ public:
...
+            auto sched_ = simulator_.vanguard().schedule();
+
+            boost::posix_time::ptime start_time = boost::posix_time::from_time_t(sched_.getStartTime());
```
bska, line 360: Since we use the `sched_` "only" to access the start time, maybe we can write
```C++
boost::posix_time::ptime start_time = boost::posix_time::from_time_t(simulator_.vanguard(),schedule().getStartTime());
```
to avoid creating a full copy of the `Schedule` object?

ebos/eclwriter.hh
```diff
@@ -349,14 +351,27 @@ public:
...
+
+            inplace = eclOutputModule_->calc_inplace(miscSummaryData, regionData, simulator_.gridView().comm());
+
+            if (simulator_.gridView().comm().rank() == 0){
```
bska, line 364: I think I'd prefer `collectToIORank_.isIORank()` instead of the explicit `rank() == 0` check.  They do the same thing, of course, but the `isIORank()` check is more explicit about the context.

ebos/eclwriter.hh
```diff
@@ -121,6 +123,9 @@ class EclWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::Grid>
     enum { enableSolvent = getPropValue<TypeTag, Properties::EnableSolvent>() };
 
 public:
+
+    Inplace inplace_;
```
bska, line 127: The object **probably** does not need to be `public:`, so maybe we can put it alongside the writer's other [data members](https://github.com/OPM/opm-simulators/blob/eec2fbaaa13a65f33fbf19a4c563d8be63ea50ce/ebos/eclwriter.hh#L670-L673) instead?

opm/simulators/flow/LogOutputHelper.cpp
```diff
@@ -306,34 +308,77 @@ fip(const Inplace& inplace,
...
         std::unordered_map<Inplace::Phase, Scalar> current_values;
 
-        for (const auto& phase : Inplace::phases()) {
+        for (const auto& phase : Inplace::phases()) 
             current_values[phase] = inplace.get(phase);
-        }
+            
+        Scalar field_dyn_pv = 0.0;
+        
+        for (std::size_t reg = 1; reg <= inplace.max_region(name); ++reg)
```
bska, line 324: [`Inplace::max_region()`](https://github.com/OPM/opm-common/blob/301759536b1bdf935c56891df0e7d045028699e2/src/opm/output/eclipse/Inplace.cpp#L115) is somewhat expensive, so maybe we could write this as
```C++
for (auto nreg = inplace.max_region(name), reg = 0*nreg + 1; reg <= nreg; ++reg) {
    // ...
}
```
instead, just to ensure that we don't compute `max_region()` more than once?

opm/simulators/flow/LogOutputHelper.cpp
```diff
@@ -306,34 +308,77 @@ fip(const Inplace& inplace,
...
+        
+        current_values[Inplace::Phase::DynamicPoreVolume] = field_dyn_pv;
+        
         this->fipUnitConvert_(current_values);
         this->outputResvFluidInPlace_(current_values, 0);
     }
-
-    for (std::size_t reg = 1; reg <= inplace.max_region("FIPNUM"); ++reg) {
+    
+    for (std::size_t reg = 1; reg <= inplace.max_region(); ++reg) {
```
bska, line 333: Same here.  If we're changing this anyway, I'd prefer we compute `inplace.max_region()` only once.

bska (commented):
Thank you for taking the effort to improve report sheet compatibility.  I just noticed a few minor issues.  On a tangential note, we might conceivably defer outputting the word `AT` to the `LogOutputHelper::timeStamp()` member function.  That way there'd be less need to be very precise about the spacing in the `lbl` input argument to `EclGenericOutputBlackoilModule<>::outputTimeStamp()`.

bska:
jenkins build this please

tskille:
Perfect, thanks a lot for your comments. I believe that I have implemented all of them. I have also re-tested local workflows to make sure that these are still working. The PR is now rebased and from my end this is now ready to go in.

tskille:
Unfortunately, I did forget to include the tests when building flow with this PR (BUILD TESTING OFF). When including the tests, I got a compiler/linker error and I learned that I had to update one of the regression tests. 

I now have a solution for this, and I have updated the test for class LogOutputHelper.  Hence, I believe that this PR is now ready for a final review and hopefully this is now ready to go into master.

tskille:
jenkins build this please

bska:
jenkins build this please

akva2:
jenkins build this please

bska (commented):
Thanks a lot for the updates.  This looks good to me now except for one potential issue with the right-alignment of the pore-volume header in the `METRIC` and/or `FIELD` unit cases.  At some point we might look into finding a way to avoid calling `moduleVersionName()` directly from `LogOutputHelper.cpp`, but for now I'm okay with adding the new build-time dependencies here.

opm/simulators/flow/LogOutputHelper.cpp
```diff
@@ -882,37 +926,37 @@ void LogOutputHelper<Scalar>::
...
-    }
-
     const UnitSystem& units = eclState_.getUnits();
     std::ostringstream ss;
+
     if (reg == 0) {
         ss << "\n                                                     ===================================\n";
         if (units.getType() == UnitSystem::UnitType::UNIT_TYPE_METRIC) {
-            ss << "                                                     :  RESERVOIR VOLUMES      M3      :\n";
+            ss << "                                                     :  RESERVOIR VOLUMES      RM3      :\n";
```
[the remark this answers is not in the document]
tskille: thanks, this is fixed now.

opm/simulators/flow/LogOutputHelper.cpp
```diff
@@ -817,63 +862,62 @@ outputRegionFluidInPlace_(std::unordered_map<Inplace::Phase, Scalar> oip,
...
-    if (reg == 0) {
-        ss << "Field total";
-    } else {
-        ss << name << " report region " << reg;
-    }
-
-    ss << " pressure dependent pore volume = "
-       << std::fixed << std::setprecision(0)
-       << cip[Inplace::Phase::DynamicPoreVolume] << ' '
-       << units.name(UnitSystem::measure::volume) << "\n\n";
```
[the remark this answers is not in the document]
tskille: I thought this could be removed since this property is also reported in reservoir volumes array 

RPTSOL
 'FIPRESV' /

```
                                                     ===================================
                                                     :  RESERVOIR VOLUMES      RM3     :
 :---------:---------------:---------------:---------------:---------------:---------------:
 : REGION  :  TOTAL PORE   :  PORE VOLUME  :  PORE VOLUME  : PORE VOLUME   :  PORE VOLUME  :
 :         :   VOLUME      :  CONTAINING   :  CONTAINING   : CONTAINING    :  CONTAINING   :
 :         :               :     OIL       :    WATER      :    GAS        :  HYDRO-CARBON :
 :---------:---------------:---------------:---------------:---------------:---------------
 :   FIELD :      262665108:       87119681:      159998829:       15546597:      102666278:
 :       1 :      131333065:       37446737:       78339731:       15546597:       52993334:
 :       2 :      131332043:       49672944:       81659099:              0:       49672944:
 :       3 :              0:              0:              0:              0:              0:
 :       4 :              0:              0:              0:              0:              0:
 ===========================================================================================
```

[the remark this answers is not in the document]
bska: > I thought this could be removed since this property is also reported in reservoir volumes array

Ooops&ndash;yes, you're [absolutely right](https://github.com/OPM/opm-simulators/blob/3856421a8828413d524b263deeffc84444150a2a/opm/simulators/flow/LogOutputHelper.cpp#L908).  Thanks a lot for pointing that out!

bska:
jenkins build this please

bska (approved):
Thanks a lot for the updates and for taking the time to do this work.  I'll merge into master.

PR #5070, opened 2023-12-17
Fix PRT file output directory

tskille:
PRT and DBG file should go to same directory as the rest of the output files from the simulation. In current version of OPM Flow this goes wrong if flow is run from a directory which is not the same as the data file parent directory.

Example 

```
flowdaily_rh7 model/TEST3B_FLOW.DATA
```

PRT file and DBG file will show up in same directory as flow was started from 

```
bash-4.2$ ls
model  TEST3B_FLOW.DBG	TEST3B_FLOW.PRT
```

### DIFF
--- a/opm/simulators/utils/readDeck.cpp
+++ b/opm/simulators/utils/readDeck.cpp
@@ -439,8 +439,8 @@ Opm::setupLogging(const int          mpi_rank_,
             : std::filesystem::current_path().generic_string();
     }
 
-    logFileStream << output_dir << "/" << baseName;
-    debugFileStream << output_dir << "/" << baseName;
+    logFileStream << fpath.parent_path().string() << "/" << fpath.stem().string();
+    debugFileStream << fpath.parent_path().string() << "/" << fpath.stem().string();
 
     if (mpi_rank_ != 0) {
         // Added rank to log file for non-zero ranks.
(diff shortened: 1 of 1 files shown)


opm/simulators/utils/readDeck.cpp
```diff
@@ -439,8 +439,8 @@ Opm::setupLogging(const int          mpi_rank_,
             : std::filesystem::current_path().generic_string();
     }
 
-    logFileStream << output_dir << "/" << baseName;
-    debugFileStream << output_dir << "/" << baseName;
+    logFileStream << fpath.parent_path().string() << "/" << fpath.stem().string();
+    debugFileStream << fpath.parent_path().string() << "/" << fpath.stem().string();
```
atgeirr, line 443: I think this will write those streams to the deck location and not the output directory?

I can reproduce the problem, and will take a look at this.

bska: > I think this will write those streams to the deck location and not the output directory?

Yes, this could potentially break the `--output-dir` option.  Ideally we'd use [`IOConfig::getFullBasePath()`](https://github.com/OPM/opm-common/blob/e111bf0e6e63e92664d1d43eafc9774788e5547b/src/opm/input/eclipse/EclipseState/IOConfig/IOConfig.cpp#L273) here, but we typically configure the logging system prior to forming the `EclipseState`.

atgeirr:
I found the cause of the problem. This was recently added in Main.hpp, lines 331-333:
```
        if (outputDir.empty()) {
            outputDir = ".";
        }
```
I believe this was done to ensure that the damaris subsystem always gets a nonempty dir, I will remove this and do it in a different way. (Empty output dir means default behaviour, i.e. write to the dir in which the deck is located.)

PR #5071, opened 2023-12-18
Assert refactored to include same-parent-leaf-cells

aritorto:

This PR modifies an assert to take into account this feature/issue, when checking if there are inactive cells in between and setting the transmissibility.


aritorto:
jenkins build this please

PR #5074, opened 2023-12-18
Lookup for extractPorosity and extractDispersion_

aritorto:
For CpGrid with LGRs, the field properties of a cell on the leaf grid view is inherited from its parent or equivalent cell from level zero. Now, extractPorosity_() and extractDispersion_() are modified to support that.


aritorto:
jenkins build this please

blattms:
Thanks a lot.

PR #5075, opened 2023-12-19
Avoid use of problematic 'saturationPressure'-function

steink:
In `computeConnectionDFactor` the function `gasPvt().saturationPressure(...)` is called to obtain a pressure for which the gas phase becomes saturated for a given `rv`-value. For cases where saturated `rv` is non-monotonic in pressure, this is not well defined. Also non-monotonicity may lead to convergence failures for the local Newton solver in the `saturationPressure`-function.

This PR simply swithces to rather use saturated `rv`-value at `connection_pressure` to test whether the gas phase is saturated or not.


bska:
jenkins build this please

bska (approved):
Good catch!  The change looks good to me and I'll merge into master.

PR #5076, opened 2023-12-19
Support LGRs when creating/resetting transmissibilities

aritorto:

This PR modifies createTransmissibilyt_() and resetTransmissibility_() to take this into account.

### DIFF
--- a/ebos/ecltransmissibility_impl.hh
+++ b/ebos/ecltransmissibility_impl.hh
@@ -788,7 +788,7 @@ updateFromEclState_(bool global)
     for (auto it = trans.begin(); it != trans.end(); ++it, ++key, ++perform)
     {
         if(*perform){
-            if(grid_.maxLevel()>1) {
+            if(grid_.maxLevel()>0) {
                 OPM_THROW(std::invalid_argument, "Calculations on TRANX/TRANY/TRANZ arrays are not support with LGRS, yet.");
             }
             fp->apply_tran(*key, *it);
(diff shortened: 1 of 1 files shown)


aritorto:
jenkins build this please

ebos/ecltransmissibility_impl.hh
```diff
@@ -786,7 +788,10 @@ updateFromEclState_(bool global)
     for (auto it = trans.begin(); it != trans.end(); ++it, ++key, ++perform)
     {
         if(*perform)
-            fp->apply_tran(*key, *it);
+            if(grid_.maxLevel()>1) {
+                OPM_THROW(std::invalid_argument, "Calculations on TRANX/TRANY/TRANZ arrays are not support with LGRS, yet.");
+            }
+        fp->apply_tran(*key, *it);
```
blattms, line 794: There seem to be braces missing. Now we will always call the last line. Should only be done if `*perform` is true

aritorto: Thanks, it's corrected now

aritorto:
jenkins build this please

aritorto:
jenkins build this please

blattms:
Suddenly I am  a bit undecided whether we should merge this. I am aware that this is due to me.

We are throwing if there is TRANX/TRANY/TRANZ in combination with LGRs, because we are not 100% sure how to handle this yet. On the other hand we already change the code in a way that makes clear that we assume that TRANX/TRANY/TRANZ are inherited and all neighbors in the LGR are cartesian and have the same TRANX/TRANY/TRANZ as the parent.

I am not 100% sure whether this is what should be done (although it seems reasonable), Another possibility would be that no values for TRANX/TRANY/TRANZ are applied to the intersections between cells of the LGRs and they just hold for the boundary between LGR and the rest.

Maybe somebody else has some insight here? @bska @goncalvesmachadoc

Maybe splitting the PR into two (one for the exception and one for the rest) would be better?

aritorto:
Now commits are splitted into two PRs: OPM/opm-simulators#5078 and OPM/opm-simulators#5079

blattms:
Should we close this one then?

aritorto:
We can close it for now

PR #5077, opened 2023-12-19
Throw for LGR combined with MULTZ keyword

aritorto:
For now, LGRs do not support MULTZ. However, a minor change has been added in applyAllZMultipliers_() to take into account that, when a grid is a CpGrid with LGRs, then cells on the leaf grid view with the same parent cell on level same have the same Cartesian indices.


aritorto:
jenkins build this please

blattms:
Thanks a lot.

PR #5078, opened 2023-12-19
Throw for LGRs combine with TRANX/Y/Z

aritorto:
For now, LGRs do not support TRANX, TRANY, TRANZ, therefore a throw has been added.


aritorto:
jenkins build this please

PR #5079, opened 2023-12-19
Support LGRs when creating/resetting transmissibilities

aritorto:

This PR modifies `createTransmissibilty_()` and `resetTransmissibility_()` to take this into account.


aritorto:
jenkins build this please

blattms:
jenkins build this please

PR #5081, opened 2023-12-21
Use gas pressure for gas-water cases in rockcomp

totto82:


totto82:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -2091,7 +2081,7 @@ protected:
             if (!this->maxOilSaturation_.empty())
                 this->maxOilSaturation_[elemIdx] = std::max(this->maxOilSaturation_[elemIdx], fs.saturation(oilPhaseIdx));
             if (!this->minOilPressure_.empty())
-                this->minOilPressure_[elemIdx] = std::min(this->minOilPressure_[elemIdx], fs.pressure(oilPhaseIdx));
+                this->minOilPressure_[elemIdx] = std::min(this->minOilPressure_[elemIdx], fs.pressure(refPressurePhaseIdx_()));
```
bska, line 2084: The other quantities here, $S_w^{\mathrm{max}}$ and $S_o^{\mathrm{max}}$, both use explicit phase indices.  Why is it that we only vary the phase index for the pressure quantity?

bska (commented):
Why do `updateMaxOilSaturation_()` and `updateMinPressure_()` both use the new `refPressurePhaseIdx_()` function, while `readinitialCondition_()` continues to use `oilPhaseIdx` and `waterPhaseIdx` to initialise the maximum oil and water saturation arrays?

ebos/eclproblem.hh
totto82: Both maxWaterSaturation_ and maxOilSaturation_ are empty if the phase does not exist. Pressure is needed for all combinations of phases. I will rename minOilPressure_ to minRefPressure_ to avoid further confusion.

totto82:
jenkins build this please

bska (approved):
> Both `maxWaterSaturation_` and `maxOilSaturation_` are empty if the phase does not exist.

Ah, right.  Sorry, I didn't think of that.

> Pressure is needed for all combinations of phases. I will rename `minOilPressure_` to `minRefPressure_` to avoid further confusion.

That makes sense to me.  Thanks a lot for the updates.  I'll merge into master now.

PR #5082, opened 2023-12-21
Skip writing vtk output when LGRs

aritorto:
Writing output when a grid is a CpGrid with LGRs is not supported yet. This PR skips writing output in those cases. 

An anlernative to this changes could be to support --disable-vtk-output, or similar. I'm not deeply familiar with this part of the code yet.


aritorto:
jenkins build this please

aritorto:
It turns out that such command parameter exists. So this PR will be most likely closed soon.

PR #5085, opened 2024-01-02
Mark keywords SKIP, SKIP100, SKIP300, and ENDSKIP as supported.

blattms:
Follow up to OPM/opm-common#3817 which implements the support for these keywords. Currently the simulator will still stop if these keywords are used.

Whether to merge this or not is probably up to @vkip and @bska


blattms:
jenkins build this please

vkip:
> Whether to merge this or not is probably up to @vkip and @bska

I support merging! Thanks for fixing this. I suspect I was working with a model that required '--parsing-strictness=low' when adding this, should of course have added a basic test..

bska:
> > Whether to merge this or not is probably up to @vkip and @bska
> 
> I support merging! Thanks for fixing this.

We **might** want to list this as "partially supported" instead, because I don't think we fully handle the case of **nested** `SKIP`/`ENDSKIP` keywords.  This probably aren't particularly common in real field cases, but I can imagine the situation presenting if we're combining include files from multiple models.

blattms:
Generally, this sounds like a good idea.

The problem is that by design the validator (to be used  for partially supported keywords) will only be used to test values of the items in the keyword itself, see [KeywordValidator::validateKeywordItems](/OPM/opm-simulators/blob/0c01a3ac509bfdcc8bce4d740942e2d06a4dad6f/opm/simulators/flow/KeywordValidation.cpp#L150-L178). So you won't get any warning at all, anyway. 

But maybe I am missing something or you meant something completely different?

bska:
> > We **might** want to list this as "partially supported" instead, because I don't think we fully handle the case of **nested** `SKIP`/`ENDSKIP` keywords.
>
> Generally, this sounds like a good idea.
> 
> The problem is that by design the validator (to be used for partially supported keywords) will only be used to test values of the items in the keyword itself,

Oh, yeah, right.  That's a good point. We'd need to redesign partially supported keyword logic in order to support detecting that case.  For now I'll merge this PR.

gdfldm:
@blattms / @bska, as I understand it the SKIP100 and SKIP300 keywords allow the same deck to be run using either E100 or E300. Where the deck includes for example both blackoil fluid model data and the equivalent compositional fluid model data, something like this...

```
--data common to E100 and E300

SKIP100
--compositional only data skipped by E100
ENDSKIP

SKIP300
--blackoil only data skipped by E300
ENDSKIP
```

So always skipping everything between SKIP100/SKIP300 and ENDSKIP could cause problems. Perhaps Flow should by default behave like E100 and have a command line option to change the behaviour (say, ```--input-skip-mode=100|300|all```).

vkip:
> So always skipping everything between SKIP100/SKIP300 and ENDSKIP could cause problems. Perhaps Flow should by default behave like E100 and have a command line option to change the behaviour (say, `--input-skip-mode=100|300|all`).

Good point, I'll add such a parameter.

PR #5088, opened 2024-01-04
use arithmetic averaged compaction transmissibility multipliers

totto82:
Should be merged together with https://github.com/OPM/opm-models/pull/858


atgeirr (approved):
Looks correct, and also fixes the same bug as OPM/opm-models#858

PR #5089, opened 2024-01-04
add option for explicit pressure for rock compaction (only transmissibility)

totto82:
The option is false per default. i.e we use implicit as default.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -243,6 +243,10 @@ public:
         EWOMS_REGISTER_PARAM(TypeTag, int, NumPressurePointsEquil,
                              "Number of pressure points (in each direction) in tables used for equilibration");
         EWOMS_HIDE_PARAM(TypeTag, NumPressurePointsEquil); // Users will typically not need to modify this parameter..
+        EWOMS_REGISTER_PARAM(TypeTag, bool, ExplicitRockCompaction,
+                             "Use pressure from end of the last time step when evaluating rock compaction");
+        EWOMS_HIDE_PARAM(TypeTag, ExplicitRockCompaction); // Users will typically not need to modify this parameter..
+
     }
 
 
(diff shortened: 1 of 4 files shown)


totto82:
jenkins build this please

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1562,7 +1562,7 @@ namespace Opm
         for (int perf = 0; perf < this->number_of_perforations_; ++perf) {
             const int cell_idx = this->well_cells_[perf];
             const auto& intQuants = simulator.model().intensiveQuantities(cell_idx, /*timeIdx=*/ 0);
-            const double trans_mult = simulator.problem().template rockCompTransMultiplier<double>(intQuants, cell_idx);
+            const double trans_mult = simulator.problem().template wellTransMultiplier<double>(intQuants, cell_idx);
```
totto82, line 1565: Here I include the perm reduction due to salt contribution to the output of the connection trans factor. The test failure are from this change. @plgbrts Is this inline with what you think should be outputted?

ebos/eclproblem.hh
```diff
@@ -1657,37 +1661,11 @@ public:
...
-            return this->rockCompTransMult_[tableIdx].eval(effectiveOilPressure, /*extrapolation=*/true);
-
-        // water compaction
-        assert(!this->rockCompTransMultWc_.empty());
-        LhsEval SwMax = max(decay<LhsEval>(fs.saturation(waterPhaseIdx)), this->maxWaterSaturation_[elementIdx]);
-        LhsEval SwDeltaMax = SwMax - initialFluidStates_[elementIdx].saturation(waterPhaseIdx);
-
-        return this->rockCompTransMultWc_[tableIdx].eval(effectiveOilPressure, SwDeltaMax, /*extrapolation=*/true);
+        bool implicit = !EWOMS_GET_PARAM(TypeTag, bool, ExplicitRockCompaction);
+        return implicit? this->simulator().problem().template computeRockCompTransMultiplier_<LhsEval>(intQuants, elementIdx) :
```
atgeirr, line 1665: Formatting: space before `?`, put the ':' on the next line, aligned with the '?'. Same on line 1698.

Extra whitespace on line 1668.

ebos/eclproblem.hh
```diff
@@ -1716,7 +1694,9 @@ public:
     {
         OPM_TIMEBLOCK_LOCAL(wellTransMultiplier);
         
-        double trans_mult = this->simulator().problem().template rockCompTransMultiplier<double>(intQuants, elementIdx);
+        bool implicit = !EWOMS_GET_PARAM(TypeTag, bool, ExplicitRockCompaction);
+        double trans_mult = implicit? this->simulator().problem().template computeRockCompTransMultiplier_<double>(intQuants, elementIdx) :
+                            this->simulator().problem().getRockCompTransMultVal(elementIdx);
         trans_mult *= this->simulator().problem().template permFactTransMultiplier<double>(intQuants);
```
atgeirr, line 1700: This is still implicit, should this factor also be explicit if RockCompaction is?

If keeping as is, you could simplify with
```
double trans_mult = rockCompTransMultiplier<double>(intQuants, elementIdx);
```
which already contain the above logic.

atgeirr (commented):
Looks good, just have a question about the perm factor being explicit and a formatting nitpick.

ebos/eclproblem.hh
totto82: This is still implicit, should this factor also be explicit if RockCompaction is?

I dont think that is necessary.  The perfFactTransMultiplier is related to salt precipitation and are different from rock compaction.

totto82:
jenkins build this please

totto82:

totto82:
jenkins build this opm-tests=1119 please

PR #5091, opened 2024-01-05
Bugfix for a corner case in MPI

locture:
A corner case was hit when running a model in MPI.

The condition became something like:
elemIdx < outsideElemIdx
insideFaceIdx == 4
useSmallestMultiplier == true

Which ultimately leads to an assert failure in applyAllZMultipliers().

Unfortunately, the model itself is proprietary thus I cannot share its detailed information. But changing the line to a comparison between insideCartElemIdx and outsideCartElemIdx does serve as a workaround.

### DIFF
--- a/ebos/ecltransmissibility_impl.hh
+++ b/ebos/ecltransmissibility_impl.hh
@@ -318,13 +318,11 @@ update(bool global, const std::function<unsigned int(unsigned int)>& map, const 
 
             // we only need to calculate a face's transmissibility
             // once...
-
-            // This comparison can be problematic for corner cases in mpi
-            // Test case observed elemIdx <= outsideElemIdx AND insideFaceIdx==4
-            // This means an assert failure in applyAllZMultipliers_() while useSmallestMultiplier==true
-            // Converting to comparison between insideCartElemIdx and outsideCartElemIdx fixes the problem
-            // if (elemIdx > outsideElemIdx)
-            if (insideCartElemIdx > outsideCartElemIdx)
+            // In a parallel run insideCartElemIdx>outsideCartElemIdx does not imply elemIdx>outsideElemIdx for
+            // ghost cells and we need to use the cartesian index as this will be used when applying Z multipliers
+            // We still need to cover the case where both cells are part of an LGR and as a consequence might have
+            // the same cartesian index
+            if (std::tie(insideCartElemIdx, elemIdx) > std::tie(outsideCartElemIdx, outsideElemIdx))
                 continue;
 
             // local indices of the faces of the inside and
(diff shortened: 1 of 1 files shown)


ebos/ecltransmissibility_impl.hh
```diff
@@ -318,7 +318,13 @@ update(bool global, const std::function<unsigned int(unsigned int)>& map, const
...
             // we only need to calculate a face's transmissibility
             // once...
-            if (elemIdx > outsideElemIdx)
+
+            // This comparison can be problematic for corner cases in mpi
+            // Test case observed elemIdx <= outsideElemIdx AND insideFaceIdx==4
+            // This means an assert failure in applyAllZMultipliers_() while useSmallestMultiplier==true
+            // Converting to comparison between insideCartElemIdx and outsideCartElemIdx fixes the problem
+            // if (elemIdx > outsideElemIdx)
+            if (insideCartElemIdx > outsideCartElemIdx)
```
blattms, line 327: I see. For a parallel the indices of the cells in the ghost layer violate the assumption that insideCartElemIdx>outsideCartElemIdx implies elemIdx>outsideElemIdx as ghost cells come after those in the interior. Z-multipliers are still applied using cartesian Indices, hence the assert.

We changed this to support LGRs more easily. With them the above won't work as we currently have different cells with the same cartesian index.

Please try
```
// In a parallel run insideCartElemIdx>outsideCartElemIdx does not imply elemIdx>outsideElemIdx for
// ghost cells and we need to use the cartesian index as this will be used when applying Z multipliers
// We still need to cover the case where both cells are part of an LGR and as a consequence might have
// the same cartesian index
if (insideCartElemIdx > outsideCartElemIdx || (insideCartElemIdx == outsideCartElemIdx && elemIdx > outsideElemIdx)
```
instead. That should work for both cases.

blattms (commented):
Great catch. thanks a lot.

Would be cool if you would try my alternative suggestion.

Somehow this must mean that we do not  a regression test that applies z-multipliers or we do not use a sufficiently high number of processors for it. Makes me a bit worried.

blattms:
jenkins build this please

ebos/ecltransmissibility_impl.hh
bska: > ```c++
> if (insideCartElemIdx > outsideCartElemIdx || (insideCartElemIdx == outsideCartElemIdx && elemIdx > outsideElemIdx)
> ```

That's, technically, equivalent to
```C++
if (std::tie(insideCartElemIdx, elemIdx) > std::tie(outsideCartElemIdx, outsideElemIdx))
```
which defers the comparison to `std::tuple<>`'s `operator>()` and is (potentially) more robust during maintenance.

bska (commented):
> Updated the comparison based on suggestions [...] Also tested on the model and can confirm problem solved.

Excellent!  I'll merge into the master branch.  Thanks a lot for identifying and fixing the problem.  This is greatly appreciated.

PR #5092, opened 2024-01-05
Output pore volume in RPTRST

svenn-t:
Output pore volume at reservoir condition using the RPORV mnemonic in RPTRST.


totto82:
jenkins build this please

totto82:
Thanks for the contribution. Looks fine.

PR #5094, opened 2024-01-05
Fix minimum newton iteration

totto82:
This bug has been around for some time. In practice master will always use 2 newton iterations even though default is to use 1. A lot of the tests will fail so this need some testing before it can be merged. A faster merge would be to change the default to 2, then all test will pass, but I think default should be 1.


totto82:
jenkins build this please

totto82:
benchmark this please

totto82:
benchmark please

atgeirr:
Thanks, merging.

We should try to change the default to 1, but as that will trigger tons of test regressions, it is better to do this first (just to get the facts straight and not lie to the user).

PR #5098, opened 2024-01-09
Add MULTREGT regression test.

blattms:
Test case was added in OPM/opm-tests#1099 and works since merge of OPM/opm-common#3845


blattms:

bska:
jenkins build this opm-tests=1114 please

bska (approved):
Thanks.  The new reference solutions have been installed on the CI system so I'll merge this to activate the test.  At some point we should consider having a parallel version of the test too.

PR #5108, opened 2024-01-18
Compute enthalpy for thermal boundary

totto82:
The flux computation for the boundary assumes that the enthalpy is computed and set in the fluidstate.


totto82:
jenkins build this please

totto82:
This is a trivial bug fix for a not much used feature. Will self merge.

PR #5109, opened 2024-01-18
Fix for crash when Damaris is deallocating variables

jcbowden:
Fix for crash when cleaning up damaris mesh offset variable due to shared memory buffer being size off by one.


akva2:
jenkins build this please

PR #5110, opened 2024-01-19
Remove check for active wells for drift compensation

totto82:
Well is active condition for drift compensation was added some time ago due to convergence issues for a case without any dynamics. For CO2 injection simulation the drift compensation is still beneficial even when wells are no longer active. An alternative to this PR is to add a parameter like onlyEnableDriftCompensationWhenWellsAreActive, but I would rather just disable driftCompensation for these cases completely.

### DIFF
--- a/opm/simulators/flow/FlowProblem.hpp
+++ b/opm/simulators/flow/FlowProblem.hpp
@@ -1543,9 +1543,7 @@ public:
 
         // if requested, compensate systematic mass loss for cells which were "well
         // behaved" in the last time step
-        // Note that we don't allow for drift compensation if there are no active wells.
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

atgeirr:
I think we should check the original motivating case before merging this.

totto82:
> I think we should check the original motivating case before merging this.

I am working on it.

totto82:
jenkins build this please

totto82:
> I think we should check the original motivating case before merging this.

The original motivation is gone since drift is no longer default. I suggest merging this when green.

PR #5115, opened 2024-01-23
changed: move ebos/eclmpiserializer.hh to opm/simulators/utils/MPISerializer.hpp

akva2:
it does not use the typetag system


akva2:
jenkins build this please

bska (approved):
I haven't quite decided what I think abut the name `Parallel::MPISerializer` yet as there's potentially some redundancy in that name, but other than that I like this.  On the other hand, it's not clear to me that `Parallel::Serializer` is a lot better either.

I'll merge this in its current form and then we can consider renaming the class later, and if the need arises.

PR #5117, opened 2024-01-24
Support at Least NTFIP Distinict Regions for Inter-Region Flow

bska:
This commit ensures that we have backing support for region IDs up to and including NTFIP (TABDIMS(5), REGDIMS(1)).  The existing setup would fail (segmentation violation) in the case of a summary vector of the form

    ROFT
      36 31 /
    /

when the maximum FIPNUM value was 30.  We nevertheless support maximum FIPNUM values exceeding NTFIP.

We add a new optional parameter to the EclInterRegionFlowMap constructor.  The parameter allows client code to specifiy a "minimum maximum" region ID that all ranks must support.  This value will be enforced during parallel aggregation.

### DIFF
--- a/ebos/eclgenericoutputblackoilmodule.cc
+++ b/ebos/eclgenericoutputblackoilmodule.cc
@@ -31,12 +31,17 @@
 #include <opm/output/data/Solution.hpp>
 
 #include <opm/input/eclipse/EclipseState/EclipseState.hpp>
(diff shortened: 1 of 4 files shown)


bska:
jenkins build this please

blattms (approved):
Looks great great. Please merge when jenkins is green.

bska:
>  Please merge when jenkins is green.

Thanks.  I'll merge this now.

PR #5121, opened 2024-01-25
Tracy local change

hnil:
-- changed some global tracy blocks to local to get clear view
-- made it easy to have build with and without local tracy


hnil:
Closing: Tracy has never been adopted upstream, so this local-block change has no home here. The branch stays on my fork if it is ever wanted.

PR #5123, opened 2024-01-25
Only print flow version in banner PRT file not all reports.

### LINKED ISSUE #5122: undefined references from time to time when using shared libs
I have no idea what I am doing wrong, but sometimes I get undefined references for test when using shared libs:
```
[ 65%] Linking CXX executable bin/test_dilu
[ 65%] Built target test_dilu
[ 65%] Linking CXX executable bin/test_wellstate
[ 65%] Linking CXX executable bin/test_aquifergridutils
[ 65%] Built target test_aquifergridutils
/usr/bin/ld: lib/libopmsimulators.so.2024.04: undefined reference to `Opm::moduleVersionName[abi:cxx11]()'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/test_wellstate.dir/build.make:146: bin/test_wellstate] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:1513: CMakeFiles/test_wellstate.dir/all] Fehler 2
[ 65%] Linking CXX executable bin/test_ParallelSerialization
/usr/bin/ld: lib/libopmsimulators.so.2024.04: undefined reference to `Opm::moduleVersionName[abi:cxx11]()'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/test_ParallelSerialization.dir/build.make:146: bin/test_ParallelSerialization] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:1840: CMakeFiles/test_ParallelSerialization.dir/all] Fehler 2
make: *** [Makefile:160: all] Fehler 2
```

blattms:
With this patch it is only printed in the banner and not in all reports.

```
                              **************************************************************************
  WELLS    AT        31  DAYS *                 SPE1 - CASE 2                                          *
  REPORT    1    01 Feb 2015  *                                             Flow  version 2024.04-pre  *
                              **************************************************************************
```
becomes
```
                              **************************************************************************
  WELLS    AT        31  DAYS *                 SPE1 - CASE 2                                          *
  REPORT    1    01 Feb 2015  *                                                                        *
                              **************************************************************************
```

This also resolves linker issues with some tests when using shared libraries.

For discussion as the flow version was added to reports in #5068 .

Closes #5122

### DIFF
--- a/opm/simulators/flow/LogOutputHelper.cpp
+++ b/opm/simulators/flow/LogOutputHelper.cpp
@@ -89,9 +89,7 @@ LogOutputHelper<Scalar>::LogOutputHelper(const EclipseState& eclState,
     : eclState_(eclState)
     , schedule_(schedule)
     , summaryState_(summaryState)
-{ 
-    flowVersionName_ = moduleVersionName();
-}
+{}
 
 template<class Scalar>
 void LogOutputHelper<Scalar>::
@@ -366,7 +364,7 @@ timeStamp(const std::string& lbl, double elapsed, int rstep, boost::posix_time::
         << "  " << std::left << std::setw(9) << lbl << "AT" << std::right << std::setw(10) 
         << (double)unit::convert::to(elapsed, unit::day) << "  DAYS" << " *" << std::setw(30) << eclState_.getTitle() << "                                          *\n"
         << "  REPORT " << std::setw(4) << rstep << "    " << currentDate
-        << "  *                                             Flow  version " << std::setw(11) << flowVersionName_ << "  *\n"
+        << "  *                                                                        *\n"
         << "                              **************************************************************************\n";
 
     OpmLog::note(ss.str());
--- a/opm/simulators/flow/LogOutputHelper.hpp
+++ b/opm/simulators/flow/LogOutputHelper.hpp
@@ -167,7 +167,6 @@ private:
     const EclipseState& eclState_;
     const Schedule& schedule_;
     const SummaryState& summaryState_;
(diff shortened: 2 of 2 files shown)


bska:
> For discussion as the flow version was added to reports in #5068 .

I don't think that's a completely accurate statement.  We didn't use to print the `WELLS AT` banner before, but here's an excerpt from a [Norne](https://github.com/OPM/opm-tests/blob/3b6cb1b43c8e37bc6bdc1a6e165ea9608b7fe1dc/norne/NORNE_ATW2013.DATA) run on my machine from the 6th of April 2022
>```
>                               **************************************************************************
>  Balance  at      1242  Days *                                                                        *
>  Report  100    01 Apr 2001  *                                             Flow  version 2022.04-pre  *
>                              **************************************************************************
>```
which does include the version string.

PR #5068 did however switch from passing the version string [as an argument](https://github.com/tskille/opm-simulators/blob/f46b6c5a0162913700011804aa14cb3033a4d87d/opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.hpp#L458-L459) to storing the string value internally.

blattms:
Thanks. Then I will come up with an alternative fix.

blattms:
Closing

bska:
> I will come up with an alternative fix.

Cool.  I **think** the best option here would be to go back to passing the version string as an argument&ndash;i.e., to call `moduleVersion()` at the outer level and push its result into the inner levels&ndash;but that will probably require changing some of the constructors along the way.

PR #5124, opened 2024-01-25
Revert always open history wells

vkip:
Also tries to fix some accidental well opening with local well solve control switching. Pairs with https://github.com/OPM/opm-common/pull/3873.


GitPaean:
Hei, the reverting commits is rather noncontroversial, while the commit [ae376ba](https://github.com/OPM/opm-simulators/pull/5124/commits/ae376ba63c67516118db040d011188d2e5a69488) is much more subtle. I suggest (you also suggested) we split them.  So the reverting commit along with https://github.com/OPM/opm-common/pull/3873 can go in quickly and also we can investigate the second commit as a separate issue in a separate PR. 

And also, please mark the PR as ready for review when you think it is ready.

vkip:
Thanks, will take [ae376ba](https://github.com/OPM/opm-simulators/pull/5124/commits/ae376ba63c67516118db040d011188d2e5a69488) separately, reverted now.

vkip:
jenkins build this opm-common=3873 please

PR #5126, opened 2024-01-26
Write pc for watergas systems

daavid00:
This PR allows to write in the restart file the capillary pressure for gas water systems using `PCOG`.


daavid00:
jenkins build this please

bska:
I'm not opposed to this, in general, but maybe we could output the gas/water capillary pressure as `PCGW` instead?  Is there a reason not to?

daavid00:
> I'm not opposed to this, in general, but maybe we could output the gas/water capillary pressure as `PCGW` instead? Is there a reason not to?

The reason was that in ResInsight on the Relative Permeability curves, the gas/water capillary pressure is under the label `PCOG`. I can update the PR to output it on `PCGW` instead if that is preferred.

daavid00:
I have changed it to `PCGW` and the tests passed, please let me know if further changes are required before merging this @bska.

bska (approved):

PR #5127, opened 2024-01-26
changes to account for mixing in thermal simulations

hnil:
This need pullrequests in opm-common, opm-models and opm-simulators.


hnil:
I close this: Changes moved to opm commen ensure all places are consistent.

PR #5129, opened 2024-01-26
Avoid re-opening previously stopped wells

steink:
https://github.com/OPM/opm-simulators/pull/5124 revealed a bug in the local well control switching option where previously stopped wells will re-open during the local solve. This PR simply updates the `allow_switching`-flag to `false` whenever the well state status is *not* `OPEN`. 

For reference, it's the well status in the well interface that potentially gets updated during the local solve, not the well state status.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1621,9 +1621,12 @@ namespace Opm
         const auto well_status_orig = this->wellStatus_;
         const auto operability_orig = this->operability_status_;
         auto well_status_cur = well_status_orig;
-        // only allow switcing if well is not under zero-rate target and is open from schedule
-        bool allow_switching = !this->wellUnderZeroRateTarget(summary_state, well_state) && (this->well_ecl_.getStatus() == WellStatus::OPEN);
-        allow_switching = allow_switching && (!fixed_control || !fixed_status);
+        // don't allow opening wells that are stopped from schedule or has a stopped well state
+        const bool allow_open =  this->well_ecl_.getStatus() == WellStatus::OPEN &&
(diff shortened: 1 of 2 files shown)


GitPaean:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1621,9 +1621,12 @@ namespace Opm
         const auto well_status_orig = this->wellStatus_;
         const auto operability_orig = this->operability_status_;
         auto well_status_cur = well_status_orig;
-        // only allow switcing if well is not under zero-rate target and is open from schedule
-        bool allow_switching = !this->wellUnderZeroRateTarget(summary_state, well_state) && (this->well_ecl_.getStatus() == WellStatus::OPEN);
-        allow_switching = allow_switching && (!fixed_control || !fixed_status);
+        // don't allow opening wells that are stopped from schedule or has a stopped well state
+        const bool allow_open =  this->well_ecl_.getStatus() == WellStatus::OPEN &&
```
GitPaean, line 1625: Should we also consider
```c++
const auto well_status = this->schedule()
                    .getWell(well_name, report_step).getStatus();
```
?

We consider this in https://github.com/OPM/opm-simulators/blob/d69ba021a90b534ce72651e5f0b98e1f736b95d6/opm/simulators/wells/BlackoilWellModel_impl.hpp#L808

At the same time, we `                        this->wellState().shutWell(w);` there if `well_status          == Well::Status::SHUT`.  there. 

It might be okay with the current form. 

I do think this is complicated and difficult to keep them sync. Just leaving a note here which might be helpful for the future investigation.

GitPaean (approved):
The code looks sensible.  It fixes a practical problem during testing. I am getting it in now.

PR #5133, opened 2024-01-30
DamarisVar: Some cleanup

akva2:
- fix some formatting
- avoid stream members in class
- add translation unit
- remove unused base class

Only thing I'm a bit uncertain of is what the purpose of the base class was. As-is it serves no purpose.


akva2:
jenkins build this please

bska (approved):

PR #5134, opened 2024-01-30
Feature/NONNC

lisajulia:
@blattms Can you have a look?

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -153,7 +153,7 @@ public:
 protected:
     void updateFromEclState_(bool global);
 
-    void removeSmallNonCartesianTransmissibilities_();
+    void removeNonCartesianTransmissibilities_(bool removeAll);
 
     /// \brief Apply the Multipliers for the case PINCH(4)==TOPBOT
     ///
--- a/ebos/ecltransmissibility_impl.hh
+++ b/ebos/ecltransmissibility_impl.hh
@@ -175,6 +175,8 @@ update(bool global, const std::function<unsigned int(unsigned int)>& map, const 
     const bool updateDiffusivity = eclState_.getSimulationConfig().isDiffusive();
     const bool updateDispersivity = eclState_.getSimulationConfig().rock_config().dispersion();
 
+    const bool disableNNC = eclState_.getSimulationConfig().useNONNC();
+
     if (map)
         extractPermeability_(map);
     else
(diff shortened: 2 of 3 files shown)


blattms:
Looks perfect to me.

lisajulia:
jenkins build this please

lisajulia:
@atgeirr @blattms (or someone else :)): Can you have a look? Thanks!

atgeirr (approved):
This is good, merge when green. One thing that would improve it slightly would be to make the test also include NNCs geometrically generated across a fault, not just explicitly with the NNC keyword, to ensure that all those are also removed. But that is not urgent.

atgeirr:
jenkins build this please

tests/test_nonnc.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include <opm/input/eclipse/Deck/Deck.hpp>
+#include <opm/input/eclipse/EclipseState/Grid/EclipseGrid.hpp>
+#include <opm/input/eclipse/Parser/Parser.hpp>
+#include <opm/input/eclipse/EclipseState/EclipseState.hpp>
+
+#include <dune/grid/common/mcmgmapper.hh>
+
+#include <opm/grid/CpGrid.hpp>
+
+#include <ebos/ecltransmissibility.cc>
```
akva2, line 34: don't include cc files.

tests/test_nonnc.cpp
```diff
@@ -0,0 +1,141 @@
...
+// Include ecltransmissibility_impl.hh as well, because we need the isId and isIdReverse functions that manage the keys of transmissibility map
+#include <ebos/ecltransmissibility_impl.hh>
```
akva2, line 36: there's no reason to include the impl file here. just include ecltransmissibility.hh. The code is in the library.

lisajulia: I need the functions isId and isIdReverse that are defined as free functions in ecltransmissibility_impl.hh though... of course it is not nice to include the _impl.hh file - I'll glady accept any suggestions!

akva2: i see. then de-anonymize them, ie. put them in the Opm::details namespace and put prototypes in the .hh.

akva2:
jenkins build this please

akva2:
green and approved. thanks for the effort, merging.

blattms:
> This is good, merge when green. One thing that would improve it slightly would be to make the test also include NNCs geometrically generated across a fault, not just explicitly with the NNC keyword, to ensure that all those are also removed. But that is not urgent.

If I am not totally mistaking this is exactly what we have accomplished here. This is for all NNCs as it only uses the intersections from the grid and checks the cartesian indices of the neighbors to determine whether to set the transmissibilty to 0. The test case has a fault and that is a barrier with NONNC now.

atgeirr:
> If I am not totally mistaking this is exactly what we have accomplished here. This is for all NNCs as it only uses the intersections from the grid and checks the cartesian indices of the neighbors to determine whether to set the transmissibilty to 0. The test case has a fault and that is a barrier with NONNC now.

I agree that the code is doing this. But the test case does not have a fault? Or am I looking at the wrong one? The one at test_nonnc.cpp line 73++ does not, it is a cartesian grid and not corner point.

blattms:
Indeed, but the regression test added in PR #5155 has

lisajulia:
You two are right, if needed I will add a grid in test_nonnc.cpp with faults.

PR #5138, opened 2024-01-31
DamarisWriter: some additional cleanup

akva2:
Cosmetics and some code shuffling.

### DIFF
--- a/ebos/damariswriter.cc
+++ b/ebos/damariswriter.cc
@@ -27,10 +27,14 @@
 #include <config.h>
 #include <ebos/damariswriter.hh>
 
+#include <opm/common/ErrorMacros.hpp>
 #include <opm/common/OpmLog/OpmLog.hpp>
 
 #include <Damaris.h>
 #include <fmt/format.h>
+
+#include <limits>
+#include <stdexcept>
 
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this please

ebos/damariswriter.cc
```diff
@@ -81,4 +85,62 @@ int endIteration(int rank)
     return dam_err;
 }
 
+int setupWritingPars(Parallel::Communication comm,
+                     const int n_elements_local_grid,
+                     std::vector<unsigned long long>& elements_rank_offsets)
+{
+    // one for each rank -- to be gathered from each client rank
+    std::vector<unsigned long long> elements_rank_sizes(comm.size());
```
bska, line 93: I guess we get a declaration of `vector<>` from somewhere else, but we really should `#include <vector>` here too.

bska (approved):
Apart from a missing `#include` statement, this looks good to me.

ebos/damariswriter.cc
akva2: indeed. i thought it was in the header but it's not. added there.

akva2:
jenkins build this please

bska (approved):

PR #5139, opened 2024-01-31
shutting wells zero rate constraint and not allowing crossflow

GitPaean:
this is the downstream PR of https://github.com/OPM/opm-common/pull/3880 .

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -881,47 +881,14 @@ namespace Opm {
                     }
                 }
 
-                // If a production well disallows crossflow and its
-                // (prediction type) rate control is zero, then it is effectively shut.
-                if (!well_ecl.getAllowCrossFlow() && well_ecl.isProducer() && well_ecl.predictionMode()) {
-                    const auto& summaryState = ebosSimulator_.vanguard().summaryState();
-                    const auto prod_controls = well_ecl.productionControls(summaryState);
-
-                    auto is_zero = [](const double x)
-                    {
-                        return std::isfinite(x) && !std::isnormal(x);
-                    };
-
-                    bool zero_rate_control = false;
-                    switch (prod_controls.cmode) {
-                    case Well::ProducerCMode::ORAT:
-                        zero_rate_control = is_zero(prod_controls.oil_rate);
-                        break;
-
-                    case Well::ProducerCMode::WRAT:
-                        zero_rate_control = is_zero(prod_controls.water_rate);
-                        break;
-
-                    case Well::ProducerCMode::GRAT:
-                        zero_rate_control = is_zero(prod_controls.gas_rate);
-                        break;
-
-                    case Well::ProducerCMode::LRAT:
-                        zero_rate_control = is_zero(prod_controls.liquid_rate);
-                        break;
-
-                    case Well::ProducerCMode::RESV:
-                        zero_rate_control = is_zero(prod_controls.resv_rate);
-                        break;
-                    default:
-                        // Might still have zero rate controls, but is pressure controlled.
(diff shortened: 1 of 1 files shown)


opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -883,45 +883,19 @@ namespace Opm {
...
+                        zero_rate_control = prod_controls.zeroRateLimit();
+                    } else { // injectors
+                        const auto inj_controls = well_ecl.injectionControls(summaryState_);
+                        zero_rate_control = inj_controls.zeroRateLimit();
                     }
 
                     if (zero_rate_control) {
                         // Treat as shut, do not add to container.
-                        local_deferredLogger.info("  Well shut due to zero rate control and disallowing crossflow: " + well_ecl.name());
+                        local_deferredLogger.info(fmt::format("  Well {} gets shut due to zero rate constraint and disallowing crossflow ", well_ecl.name()) );
```
GitPaean, line 898: We probably does not want to see this message for each time steps for each well.

GitPaean:
jenkins build this opm-common=3880 please

GitPaean:
jenkins build this opm-common=3880 please

GitPaean:
Same with the upstream PR, I am marking this PR as ready for review.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -881,47 +881,20 @@ namespace Opm {
...
-                    default:
-                        // Might still have zero rate controls, but is pressure controlled.
-                        zero_rate_control = false;
-                        break;
+                    if (well_ecl.isProducer()) {
+                        const auto prod_controls = well_ecl.productionControls(summaryState_);
+                        zero_rate_control = prod_controls.zeroRateLimit();
+                    } else { // injectors
+                        const auto inj_controls = well_ecl.injectionControls(summaryState_);
+                        zero_rate_control = inj_controls.zeroRateLimit();
```
bska, line 892: When combined with OPM/opm-common#3880, this logic will shut the well if **any** rate constraint, not just the currently active one, is zero.  Is that the intended behaviour here?

bska (commented):
Apart from the semantic change, which I don't immediately see the complete impact of, this looks good to me.

opm/simulators/wells/BlackoilWellModel_impl.hpp
GitPaean: I think if there is any zero rate constraint, it will be the most restrictive constraint, it means it will be the constraint that decides how the well behave.  When combined with no cross-flow, this well will not able to solve.

At the same time, it is assuming that we should not get a zero rate constraint from the group control.

bska: > I think if there is any zero rate constraint, it will be the most restrictive constraint, it means it will be the constraint that decides how the well behave. When combined with no cross-flow, this well will not able to solve.

Okay, I guess that makes sense.  Would you mind renaming the function to `anyZeroRateLimit()` though?  That makes it a little bit more explicit at the call site what condition we're actually checking.

We might also switch to writing this as
```C++
const bool zero_rate_control = well_ecl.isProducer()
    ? well_ecl.productionControls(summaryState_).anyZeroRateLimit()
    : well_ecl.injectionControls(summaryState_).anyZeroRateLimit();
```
just to make it clear that we're (now) computing the `zero_rate_control` predicate exactly once and don't need to split its calculation into two parts.

GitPaean: updated. 

Because I use constraint in the warning message and comments, I used the function name `anyZeroRateConstraint()`. I think they are same.

bska (approved):

PR #5140, opened 2024-01-31
Fix WCONPROD GRAT control mode with solvent

svenn-t:
Solvent production rate is now added to gas production rate such that their total rate is limited by GRAT control mode in WCONPROD.

From a (poorly constructed) example:
With current master, WNPR gets very large, and only the gas rate shown by (WGPR - WNPR) plot is limited by GRAT = 100 000 SM3/DAY
wgpr_wnpr_master

With PR, the total gas rate (WGPR) is limited by GRAT = 100 000 SM3/DAY 
wgpr_wnpr_fix


totto82:
jenkins build this please

PR #5142, opened 2024-02-01
BlackoilModel: remove Ebos from class names

akva2:
I've been tasked with cleaning up the ebos branding of things. This is the first step - removal in the BlackoilModel classes.


akva2:
jenkins build this please

bska (approved):
This is an overdue change and I'll merge into master.  I also appreciate replacing `typedef` with `using` where appropriate.

PR #5143, opened 2024-02-01
init diffusion module

totto82:


totto82:
Depends on https://github.com/OPM/opm-common/pull/3878

atgeirr (approved):

PR #5145, opened 2024-02-01
Add GPU Direct

multitalentloes:
This PR adds support for MPI communication being performed directly between GPUs instead of being sent via the CPU when using a multi-gpu linear solver. This is done by checking if we have a CUDA aware MPI implementation available at runtime and compile time, and then use regular MPI calls with memory buffers residing on the GPU instead of the CPU.

The first commit on this branch is Georg Zitzlsberger's prototype for the functionaity.

### DIFF
--- a/opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
+++ b/opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
@@ -198,9 +198,9 @@ public:
             size_t i = 0;
             for(const_iterator info = m_messageInformation.begin(); info != end; ++info, ++i) {
                 processMap[i]=info->first;
-                if(info->second.second.size_) {
-                    MPI_Irecv(m_GPURecvBuf->data()+info->second.second.start_,
-                            info->second.second.size_,
+                if(info->second.second.m_size) {
+                    MPI_Irecv(m_GPURecvBuf->data()+info->second.second.m_start,
+                            info->second.second.m_size,
                             MPI_BYTE,
                             info->first,
                             m_commTag,
@@ -216,9 +216,9 @@ public:
         {
             size_t i = 0;
             for(const_iterator info = m_messageInformation.begin(); info != end; ++info, ++i) {
-                if(info->second.first.size_) {
-                    MPI_Issend(m_GPUSendBuf->data()+info->second.first.start_,
-                            info->second.first.size_,
+                if(info->second.first.m_size) {
+                    MPI_Issend(m_GPUSendBuf->data()+info->second.first.m_start,
+                            info->second.first.m_size,
                             MPI_BYTE,
                             info->first,
                             m_commTag,
@@ -258,10 +258,10 @@ private:
 
     struct MessageInformation
     {
-        MessageInformation() : start_(0), size_(0) {}
-        MessageInformation(size_t start, size_t size) : start_(start), size_(size) {}
-        size_t start_; // offset in elements of "field_type"
-        size_t size_;  // size in bytes
+        MessageInformation() : m_start(0), m_size(0) {}
+        MessageInformation(size_t start, size_t size) : m_start(start), m_size(size) {}
+        size_t m_start; // offset in elements of "field_type"
+        size_t m_size;  // size in bytes
     };
 
     using InformationMap = std::map<int,std::pair<MessageInformation,MessageInformation> >;
--- a/opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
+++ b/opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
@@ -200,7 +200,7 @@ public:
                 processMap[i]=info->first;
                 if(info->second.second.m_size) {
                     MPI_Irecv(m_GPURecvBuf->data()+info->second.second.m_start,
-                            info->second.second.m_size,
+                            detail::to_int(info->second.second.m_size),
                             MPI_BYTE,
                             info->first,
                             m_commTag,
@@ -218,7 +218,7 @@ public:
             for(const_iterator info = m_messageInformation.begin(); info != end; ++info, ++i) {
                 if(info->second.first.m_size) {
                     MPI_Issend(m_GPUSendBuf->data()+info->second.first.m_start,
-                            info->second.first.m_size,
+                            detail::to_int(info->second.first.m_size),
                             MPI_BYTE,
                             info->first,
                             m_commTag,
--- a/opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
+++ b/opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
@@ -22,32 +22,46 @@
 #include <memory>
 #include <mutex>
 #include <opm/simulators/linalg/cuistl/CuVector.hpp>
+#include <vector>
 
 namespace Opm::cuistl
 {
 /**
- * @brief CUDA compatiable variant of Dune::OwnerOverlapCopyCommunication
- *
- * This class can essentially be seen as an adapter around Dune::OwnerOverlapCopyCommunication, and should work as
- * a Dune::OwnerOverlapCopyCommunication on CuVectors
- *
- *
- * @note This currently only has the functionality to parallelize the linear solve.
- *
- * @tparam field_type should be a field_type supported by CuVector (double, float)
- * @tparam block_size the block size used (this is relevant for say figuring out the correct indices)
- * @tparam OwnerOverlapCopyCommunicationType should mimic Dune::OwnerOverlapCopyCommunication.
- */
-template <class field_type, int block_size, class OwnerOverlapCopyCommunicationType>
-class CuOwnerOverlapCopy
-{
+ * @brief GPUSender is a wrapper class for classes which will implement copOwnerToAll
+ * This is implemented with the intention of creating communicators with generic GPUSender
+ * To hide implementation that will either use GPU aware MPI or not
+ * @tparam field_type is float or double
+ * @tparam OwnerOverlapCopyCommunicationType is typically a Dune::LinearOperator::communication_type
+*/
+template<class field_type, class OwnerOverlapCopyCommunicationType>
+class GPUSender {
 public:
     using X = CuVector<field_type>;
 
-    CuOwnerOverlapCopy(const OwnerOverlapCopyCommunicationType& cpuOwnerOverlapCopy)
-        : m_cpuOwnerOverlapCopy(cpuOwnerOverlapCopy)
+    GPUSender(const OwnerOverlapCopyCommunicationType& cpuOwnerOverlapCopy) : m_cpuOwnerOverlapCopy(cpuOwnerOverlapCopy){}
+
+    /**
+     * @brief copyOwnerToAll will copy source to the CPU, then call OwnerOverlapCopyCommunicationType::copyOwnerToAll on
+     * the copied data, and copy the result back to the GPU
+     * @param[in] source
+     * @param[out] dest
+     */
+    virtual void copyOwnerToAll(const X& source, X& dest) const = 0;
+    virtual void initIndexSet() const = 0;
+
+    /**
+     * @brief project will project x to the owned subspace
+     *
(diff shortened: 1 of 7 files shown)


atgeirr:
Thanks for the draft PR, makes it easier to discuss and test the code. I think you must make a modification/rebase though: the initial code commit should have Georg as author, this is important to ensure correct crediting.

multitalentloes:
Georg is now properly credited for the first commit that provided the prototype of the CUDA aware MPI.

opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
```diff
@@ -119,6 +285,96 @@ class CuOwnerOverlapCopy
     mutable std::unique_ptr<CuVector<int>> m_indicesCopy;
     mutable std::unique_ptr<CuVector<int>> m_indicesOwner;
 
+    mutable std::unique_ptr<CuVector<int>> m_commpair_indicesCopy;
+    mutable std::unique_ptr<CuVector<int>> m_commpair_indicesOwner;
```
kjetilly, line 289: snake case and camel case mixed.

opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
```diff
@@ -119,6 +285,96 @@ class CuOwnerOverlapCopy
...
+    mutable InformationMap messageInformation_;
+    typedef std::map<int,std::pair<std::vector<int>,std::vector<int> > > IM;
+    mutable IM m_im;
+
+    constexpr static int commTag_ = 0; // So says DUNE
+
+    void buildCommPairIdxs() const
+    {
+        int rank;
+        MPI_Comm_rank(m_cpuOwnerOverlapCopy.communicator(), &rank);
```
kjetilly, line 311: see above comment about rank.

opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
```diff
@@ -119,6 +285,96 @@ class CuOwnerOverlapCopy
...
+        auto &ri = m_cpuOwnerOverlapCopy.remoteIndices();
+        auto end = ri.end();
+        std::vector<int> commpair_indicesCopyOnCPU;
+        std::vector<int> commpair_indicesOwnerCPU;
+
+        for(auto process = ri.begin(); process != end; ++process) {
```
kjetilly, line 317: why not a modern for-loop?

```
for(auto process : ri)
```

opm/simulators/linalg/cuistl/SolverAdapter.hpp
```diff
@@ -163,10 +161,30 @@ class SolverAdapter : public Dune::IterativeSolver<X, X>
             auto preconditionerReallyOnGPU = preconditionerAdapterAsHolder->getUnderlyingPreconditioner();
             const auto& communication = m_opOnCPUWithMatrix.getCommunication();
 
+            // Temporary solution use the GPU Direct communication solely based on these prepcrosessor statements
+            bool mpi_might_be_supported_during_compilation = true;
+            bool mpi_might_be_supported_during_runtime;
+
+            #if defined(MPIX_CUDA_AWARE_SUPPORT) && !MPIX_CUDA_AWARE_SUPPORT
+                mpi_might_be_supported_during_compilation = false;
```
kjetilly, line 169: file is mainly using `camelCase`, this is `snake_case`

opm/simulators/linalg/cuistl/SolverAdapter.hpp
```diff
@@ -163,10 +161,30 @@ class SolverAdapter : public Dune::IterativeSolver<X, X>
             auto preconditionerReallyOnGPU = preconditionerAdapterAsHolder->getUnderlyingPreconditioner();
             const auto& communication = m_opOnCPUWithMatrix.getCommunication();
 
+            // Temporary solution use the GPU Direct communication solely based on these prepcrosessor statements
+            bool mpi_might_be_supported_during_compilation = true;
+            bool mpi_might_be_supported_during_runtime;
```
kjetilly, line 166: `snake_case`

kjetilly (commented):
I did a short review now. Main change I would suggest is to move dot and norm either to the base class GPUSender or back to the CuOwnerOverlapCopy.

multitalentloes:
Kjetils recomendations of moving functions from derived classes to the base class GPUSender have been completed. Inconsistencies in variable naming and style have been sorted out.

kjetilly:
Jenkins build this please

multitalentloes:
The test has been updated and will hopefully pass the pipeline, it is ready to be rebuilt.

akva2:
added you to trigger list.

multitalentloes:
Jenkins build this please

multitalentloes:
I have cleaned up the commit history and will rerun the tests now so we can get this merged

multitalentloes:
Jenkins build this please

multitalentloes:
Is this PR mergable @kjetilly?

opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
```diff
@@ -74,57 +88,264 @@ class CuOwnerOverlapCopy
...
+            }
+        }
+        int finished = MPI_UNDEFINED;
+        MPI_Status status;
+        for(size_t i = 0; i < numberOfRealRecvRequests; i++) {
+            status.MPI_ERROR=MPI_SUCCESS;
+            MPI_Waitany(m_messageInformation.size(), recvRequests.data(), &finished, &status);
+
+            if(status.MPI_ERROR!=MPI_SUCCESS) {
+                std::cerr<< rank << ": MPI_Error occurred while receiving message from "<< processMap[finished] << std::endl;
```
kjetilly, line 239: Should not write directly to `std::cerr`. Use the logging framework, and I believe the `OPM_THROW` macro automatically outputs the error message. Variables can easily be included in a throw message using eg. `fmt::format`

opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
```diff
@@ -74,57 +88,264 @@ class CuOwnerOverlapCopy
...
+                OPM_THROW(std::runtime_error, "MPI_Error while receiving message");
+            }
+        }
+        MPI_Status recvStatus;
+        for(size_t i = 0; i < m_messageInformation.size(); i++) {
+            if(MPI_SUCCESS!=MPI_Wait(&sendRequests[i], &recvStatus)) {
+                std::cerr << rank << ": MPI_Error occurred while sending message to " << processMap[finished] << std::endl;
```
kjetilly, line 246: see above comment about `std::cerr`

opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
```diff
@@ -74,57 +88,264 @@ class CuOwnerOverlapCopy
...
-    // always, but they are not always used.
-    mutable std::once_flag m_initializedIndices;
-    mutable std::unique_ptr<CuVector<int>> m_indicesCopy;
-    mutable std::unique_ptr<CuVector<int>> m_indicesOwner;
+    struct MessageInformation
+    {
+        MessageInformation() : start_(0), size_(0) {}
+        MessageInformation(size_t start, size_t size) : start_(start), size_(size) {}
+        size_t start_; // offset in elements of "field_type"
+        size_t size_;  // size in bytes
```
kjetilly, line 266: nitpick: Rest of file uses `m_` for member variables.

opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
```diff
@@ -74,57 +88,264 @@ class CuOwnerOverlapCopy
...
+        // Start MPI stuff here...
+        // Note: This has been taken from DUNE's parallel/communicator.hh
+        std::vector<MPI_Request> sendRequests(m_messageInformation.size());
+        std::vector<MPI_Request> recvRequests(m_messageInformation.size());
+        std::vector<int> processMap(m_messageInformation.size());
+        size_t numberOfRealRecvRequests = 0;
+
+        using const_iterator =  typename InformationMap::const_iterator;
+        const const_iterator end = m_messageInformation.end();
+
```
kjetilly, line 196: This function deserves a comment describing on a high-level what it does (beyond the words "copy owner to all"). similar for the ` buildCommPairIdxs` function. One could probably look at the phrasing for [the corresponding dune class](https://gitlab.dune-project.org/core/dune-common/-/blob/master/dune/common/parallel/communicator.hh?ref_type=heads)

opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
```diff
@@ -74,57 +88,264 @@ class CuOwnerOverlapCopy
...
+        {
+            size_t i = 0;
+            for(const_iterator info = m_messageInformation.begin(); info != end; ++info, ++i) {
+                processMap[i]=info->first;
+                if(info->second.second.size_) {
+                    MPI_Irecv(m_GPURecvBuf->data()+info->second.second.start_,
+                            info->second.second.size_,
```
kjetilly, line 203: Am I missing something here? MPI_Irecv (and send) takes an `int` size argument, while `info->second.second.size_,` is a `size_t`, shouldn't Jenkins have caught this? If I am not misunderstanding, could you add a `to_int` here to catch any overflows? While not that likely today, notice that we are sending bytes, so with `int` we would run into the `int` limit with "only" 268 435 455 cells per process. 

This is probably nothing to worry about for the forseeable future, but one never knows...

kjetilly (commented):
There are some beautification fixes that could be done, but those could also be postponed to a later PR. I would say this can be merged now to not stall the review process too much, but please note some of the comments.

opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
```diff
@@ -141,9 +362,54 @@ class CuOwnerOverlapCopy
             }
         }
 
-        m_indicesCopy = std::make_unique<CuVector<int>>(indicesCopyOnCPU);
-        m_indicesOwner = std::make_unique<CuVector<int>>(indicesOwnerCPU);
+        this->m_indicesCopy = std::make_unique<CuVector<int>>(indicesCopyOnCPU);
+        this->m_indicesOwner = std::make_unique<CuVector<int>>(indicesOwnerCPU);
```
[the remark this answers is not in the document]
multitalentloes: I have removed them now from the base class, but for the derived classes it seems necessary when accessing variables from the base class. Not sure why.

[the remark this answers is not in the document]
akva2: members inherited from a template base class requires this->  usage.

multitalentloes:
Jenkins build this please

kjetilly (approved):

opm/simulators/linalg/cuistl/CuOwnerOverlapCopy.hpp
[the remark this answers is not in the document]
atgeirr: The general trend in OPM code is to write all member variable access as this->member, consensus seems to be that this is a good idea. It may reduce the need to mark members by underscore postfix etc. as well. I have personally been a bit sceptical about making this a rule or strong recommendation, not least because it is not easy to check that it is followed 100% in any given class, but I have no strong objections.

opm/simulators/linalg/cuistl/SolverAdapter.hpp
```diff
@@ -33,7 +33,11 @@
 #include <opm/simulators/linalg/cuistl/PreconditionerAdapter.hpp>
 #include <opm/simulators/linalg/cuistl/detail/has_function.hpp>
 
-
+#ifdef OPEN_MPI
+#if OPEN_MPI
+#include "mpi-ext.h"
+#endif
+#endif
```
blattms, line 40: I am a bit surprised why it is done this way (everything tailored towards OpenMPI). It seems like many MPI implementations support this.

Is there a special reason why we focus only on one?

PR #5148, opened 2024-02-01
SimulatorFullyImplicitBlackoilEbos: rename to SimulatorFullyImplicitBlackoil

akva2:
And yet another class

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -453,7 +453,7 @@ list (APPEND PUBLIC_HEADER_FILES
   opm/simulators/flow/Main.hpp
   opm/simulators/flow/NonlinearSolver.hpp
   opm/simulators/flow/RSTConv.hpp
-  opm/simulators/flow/SimulatorFullyImplicitBlackoilEbos.hpp
+  opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp
   opm/simulators/flow/SimulatorReportBanners.hpp
   opm/simulators/flow/SimulatorSerializer.hpp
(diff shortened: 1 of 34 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this please

bska (commented):
Weird.  The `SimulatorFullyImplicitBlackoil.cpp` file was renamed without changes.  Do we not include the associate header in that file?

bska:
> There is only a few free functions in there

Right, but aren't these then, technically, built without a declaration in scope?

> and if we include the header we would pull in everything and the kitchen sink unnecessarily.

I appreciate that.  Maybe we could have a separate header for just the two free functions?

akva2:
Lacking definition is not really a problem for the duck when it comes to free functions But we absolutely can rename and add a separate header, it just did not end up that way. Suggested file name?

bska:
> Lacking definition is not really a problem for the duck when it comes to free function

Sorry, I don't understand what you mean by "duck" here.

> But we absolutely can rename and add a separate header, it just did not end up that way. Suggested file name

Cool.  How about `SimulatorReportBanners.[hc]pp`?  Tangentially, function `outputTimestampFIP()` appears to be unused now.  The only references I can find to that function are in the header and `.cpp` file being renamed here.

akva2:
Hahaha. That's what happens when you write with one arme while playing with your 4 year old with the other. s/duck/linker/g

akva2:
jenkins build this please

bska (approved):
Excellent.  Thanks a lot for the updates.  This looks good to me now and I'll merge into master.

PR #5157, opened 2024-02-02
New defaults

totto82:
I think it is time to change some defaults in Flow related to accuracy and performance. 
Motivation: 
- $10^{-6}$ mb tol sometimes leads to too large accumulated mass balance errors 
- drift compensation sometimes leads to significant convergence issues (and are not needed if mb = $10^{-7}$)
- ilu0 leads to tool many iterations for mb $10^{-7}$
- for cases where we force small timesteps (for instance using TUNING) 1 newton iteration is often sufficient)

I have run with these parameters for many cases for a long time. 

This will need extensive testing before it can be merged.


totto82:
jenkins build this please

GitPaean:
not sure about the best way to do this.  Do you think it might be easier to get things in if you split them among multiple PRs?  The combined effects can be tricky to test.

totto82:
> not sure about the best way to do this. Do you think it might be easier to get things in if you split them among multiple PRs? The combined effects can be tricky to test.

I see your point, but (1) I would like a common discussion about the new defaults (2) They are all interlinked (the fourth one could be taken independently) (3) They all effect results, and it will be a large effort to check all the failed tests.

totto82:
benchmark please

alfbr:
From our end, this is a highly welcome initiative. We will probably need some time to do proper testing as this will impact all models we run.

totto82:
The MICP case seg-faults in extractCPRPressureMatrix with CPRW (works for the other variants) The MICP model is a one-phase (water) model with 5 components.

alfbr:
benchmark please

totto82:
> SPE1CASE1_BRINE and SPE1CASE1_PRECSALT starts oscillating takes forever to complete with mb=1e-7.

For this case the simplest solution is to introduce a relaxed_tol_mb (ala TUNING 2-7).

blattms:
Posting missing ytelses results:

```
| Test    | Configuration                                                                         | Relative |
| ------- | ------------------------------------------------------------------------------------- | -------- |
| opm-git | OPM Benchmark: drogon - Threads: 1                                                    | 0.998    |
| opm-git | OPM Benchmark: drogon - Threads: 8                                                    | 0.999    |
| opm-git | OPM Benchmark: punqs3 - Threads: 1                                                    | 0.982    |
| opm-git | OPM Benchmark: punqs3 - Threads: 8                                                    | 1.006    |
| opm-git | OPM Benchmark: smeaheia - Threads: 1                                                  | 0.963    |
| opm-git | OPM Benchmark: smeaheia - Threads: 8                                                  | 1        |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 1                                             | 1.003    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 8                                             | 1.003    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 - FOIT (Total Oil Injection At End Of Run) | 1        |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 - FOIT (Total Oil Injection At End Of Run) | 1        |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1                                            | 0.995    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8                                            | 1.009    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1                                     | 1.006    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8                                     | 0.999    |
```

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
maybe we should incorporate https://github.com/OPM/opm-simulators/pull/5280 to this PR. @totto82

totto82:
benchmark please

atgeirr:
Making a data update here, not intending to merge that without analysis, but this is to find test failures that are not "fixable" by new reference data, such as serial vs parallel comparison.

bska:
> Making a data update here, not intending to merge that without analysis, but this is to find test failures that are not "fixable" by new reference data, such as serial vs parallel comparison.

Okay, but this is a fairly expensive approach given that this work affects more or less every existing regression test.  For what it's worth, the new candidate solutions generated roughly 1.9 GB of binary files.  Please hold additional attempts until other in-flight updates, e.g., #5232, have been merged.

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: drogon - Threads: 1                | 1.004    |
| opm-git | OPM Benchmark: drogon - Threads: 8                | 0.819    |
| opm-git | OPM Benchmark: punqs3 - Threads: 1                | 1.013    |
| opm-git | OPM Benchmark: punqs3 - Threads: 8                | 0.961    |
| opm-git | OPM Benchmark: smeaheia - Threads: 1              | 1.003    |
| opm-git | OPM Benchmark: smeaheia - Threads: 8              | 0.876    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 1         | 0.991    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 8         | 0.993    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 0.999    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 0.897    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 1.008    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 0.952    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.99     |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 0.946    |



View result details @ https://www.ytelses.com/opm/?page=result&id=2486

totto82:
jenkins build this opm-tests=5157 please

GitPaean:
> jenkins build this opm-tests=5157 please

OPM/opm-tests#5157 does not exist and https://github.com/OPM/opm-tests/pull/1165 can not be used due to merging conflicts. You probably should update_data again to get a reference update PR.

bska:
> You probably should update_data again to get a reference update PR.

Maybe.  If we're "just" trying to determine which case will not be cured by updating the reference solutions, then we can get a lot of information from the [test failures](https://ci.opm-project.org/job/opm-simulators-PR-builder/6277/testReport/) of the latest jenkins run.  Regression test cases named `mpi.compareECLFiles_flow+<CASE>` are those that would typically be fine when we update the reference solutions.  The others are

* `mpi.compareDamarisFiles_flow+SPE1CASE1`
* `mpi.compareParallelSim_flow+AQUFLUX-01`
* `mpi.compareParallelSim_flow+SPE1CASE2_SOLVENT`
* `mpi.compareRestartedSim_flow+NETWORK-01-REROUTE-RESTART`
* `mpi.compareRestartedSim_flow+NETWORK-01-RESTART`
* `mpi.python_fluidstate_variables`
* `mpi.python_primary_variables`

The Damaris case may be okay as well if we update the reference solutions.  Finally, there is the case named `mpi.compareECLFiles_flow_blackoil_polyhedralgrid+SPE1CASE2` which I don't know enough about without looking into the details.

steink:
I've started to run the tests for seperate parts of the PR in attempt to get a better overview. I don't know whether the `--local-well-solve-control-switching`-option has ever been run on anything other than `flow_blackoil`, and I see for instance the potential calculations for this option does not handle solvent. Should be simple fix, but will investigate a bit further to check for possible other implications.

GitPaean:
> I understand the intention here (ameliorate errors due to explicit treatment of pressure-drop along well), but I did not include this second round of updates in computeWellPotentialsImplicit because I was uncertain on the behaviour of this fix when equations are non-linear (e.g., cross-flow). It could easily be included, and would lead to fewer failures, but I'm not sure it's neccessary. Any opinions on this @GitPaean , @totto82 or others?

For this part, how big is the difference of the well potential values. If the difference is only minor, I think you can decide. If the difference is more significant, maybe we should have a look.

totto82:
>I understand the intention here (ameliorate errors due to explicit treatment of pressure-drop along well), but I did not include this second round of updates in computeWellPotentialsImplicit because I was uncertain on the behaviour of this fix when equations are non-linear (e.g., cross-flow). 

If I remember correctly this improved the results on some particular wells on some models. But this was long time ago, and I dont remember the model. Maybe it was Norne or SPE9? No strong opinions, as the difference should be small anyway.

steink:
@GitPaean @totto82 thanks! From what I can see, differences are small compared to e.g., changing the linear solver ;). At some point we discussed having a seperate _well_state_ for the potentials which would avoid the copying and provide better initial guesses for the calculations (this would also give _explicit_ p-drop calculations at relevant pressure), but this is for later.

atgeirr:
jenkins build this please

steink:
Note that #5435 needs to be merged in order to prevent a solvent-related crash

totto82:

totto82:
I agreed with Atgeirr to run a data update and check to prepare for merging. I final confirmation on the test failures should be done before this is merged.

totto82:
jenkins build this opm-tests=1205 please

bska:
> final confirmation on the test failures should be done before this is merged.

We also need to understand why the [compareSeparateECLFiles_flow_actionx_compdat_8_procs](https://ci.opm-project.org/job/opm-simulators-PR-builder/6487/testReport/junit/(root)/mpi/compareSeparateECLFiles_flow_actionx_compdat_8_procs/) regression test times out&ndash;i.e., reach the 1500 second (25 minute)&ndash;limit on maximum runtime without really doing anything at all.

totto82:
> > final confirmation on the test failures should be done before this is merged.
> 
> We also need to understand why the [compareSeparateECLFiles_flow_actionx_compdat_8_procs](https://ci.opm-project.org/job/opm-simulators-PR-builder/6487/testReport/junit/(root)/mpi/compareSeparateECLFiles_flow_actionx_compdat_8_procs/) regression test times out–i.e., reach the 1500 second (25 minute)–limit on maximum runtime without really doing anything at all.

Is https://github.com/OPM/opm-simulators/pull/5510 relevant for this?

bska:
> > We also need to understand why the [compareSeparateECLFiles_flow_actionx_compdat_8_procs](https://ci.opm-project.org/job/opm-simulators-PR-builder/6487/testReport/junit/(root)/mpi/compareSeparateECLFiles_flow_actionx_compdat_8_procs/) regression test times out–i.e., reach the 1500 second (25 minute)–limit on maximum runtime without really doing anything at all.
> 
> Is #5510 relevant for this?

Maybe, but it's not completely clear to me.  For what it's worth the [regression test](https://github.com/OPM/opm-simulators/blob/eb3d4e92035a30db54c104c463a3597525923d00/parallelTests.cmake#L229-L238) in question **does not** use the NLDD solver.

totto82:
There are 9 failures that can not be solved with just an update of the solution. These needs to be addressed one by one.

lisajulia:
> > > final confirmation on the test failures should be done before this is merged.
> > 
> > 
> > We also need to understand why the [compareSeparateECLFiles_flow_actionx_compdat_8_procs](https://ci.opm-project.org/job/opm-simulators-PR-builder/6487/testReport/junit/(root)/mpi/compareSeparateECLFiles_flow_actionx_compdat_8_procs/) regression test times out–i.e., reach the 1500 second (25 minute)–limit on maximum runtime without really doing anything at all.
> 
> Is #5510 relevant for this?



> > > We also need to understand why the [compareSeparateECLFiles_flow_actionx_compdat_8_procs](https://ci.opm-project.org/job/opm-simulators-PR-builder/6487/testReport/junit/(root)/mpi/compareSeparateECLFiles_flow_actionx_compdat_8_procs/) regression test times out–i.e., reach the 1500 second (25 minute)–limit on maximum runtime without really doing anything at all.
> > 
> > 
> > Is #5510 relevant for this?
> 
> Maybe, but it's not completely clear to me. For what it's worth the [regression test](https://github.com/OPM/opm-simulators/blob/eb3d4e92035a30db54c104c463a3597525923d00/parallelTests.cmake#L229-L238) in question **does not** use the NLDD solver.

I think #5510 is not relevant for the failure of [this regression test](https://github.com/OPM/opm-simulators/blob/eb3d4e92035a30db54c104c463a3597525923d00/parallelTests.cmake#L229-L238), as @bska said.
I'll take a look at this.

GitPaean:
FYI, the regression failure of [mpi.compareECLFiles_flow+2D_POLYMER_INJECTIVITY](https://ci.opm-project.org/job/opm-simulators-PR-builder/6489/testReport/junit/(root)/mpi/compareECLFiles_flow_2D_POLYMER_INJECTIVITY/) is due to the use of `cprw` as the new default.

bska:
> FYI, the regression failure of [mpi.compareECLFiles_flow+2D_POLYMER_INJECTIVITY](https://ci.opm-project.org/job/opm-simulators-PR-builder/6489/testReport/junit/(root)/mpi/compareECLFiles_flow_2D_POLYMER_INJECTIVITY/) is due to the use of `cprw` as the new default.

I **believe** that this ("cprw" new default linear solver) is the direct cause of the [MICP regression failure](https://ci.opm-project.org/job/opm-simulators-PR-builder/6489/testReport/junit/(root)/mpi/compareECLFiles_flow_MICP/) too.  At least I can make the test run, albeit not successfully, on my local machine with the change below.  Do we expect the CPRW solver to work for the [`MICP`](https://github.com/OPM/opm-tests/blob/2c047c1d4c053fecaeba6e556c4dbe7d967a4058/micp/MICP.DATA) model?

---

```diff
diff --git a/regressionTests.cmake b/regressionTests.cmake
index f07056bf5..ed48cb56c 100644
--- a/regressionTests.cmake
+++ b/regressionTests.cmake
@@ -1119,7 +1119,8 @@ add_test_compareECLFiles(CASENAME micp
                          SIMULATOR flow
                          ABS_TOL ${abs_tol}
                          REL_TOL ${rel_tol}
-                         DIR micp)
+                         DIR micp
+                         TEST_ARGS --linear-solver=ilu0)
 
 add_test_compareECLFiles(CASENAME 0_base_model6
                          FILENAME 0_BASE_MODEL6
```

GitPaean:
> Do we expect the CPRW solver to work for the [MICP](https://github.com/OPM/opm-tests/blob/2c047c1d4c053fecaeba6e556c4dbe7d967a4058/micp/MICP.DATA) model?

From my test, it is also due to cprw. But I do not know much about the micp implementation.

bska:
In any case, the behaviour **cannot** be the simulator crashing due to a segmentation fault if faced with inputs it cannot handle in its default setting.  If we're going to make CPRW be the default solver then we either need to detect these situations and stop the run with a diagnostic along the lines of

> That combination is not currently supported, please use `--linear-solver=ilu0`

or we need to ensure that the CPRW solver is actually able to run these models.

If we don't do any of this, then I don't think we can make CPRW be the default linear solver.

atgeirr:
The failure for compareSeparateECLFiles_flow+actionx_compdat_1_proc (and the NLDD variant of that test) also seems to be CPRW triggered.

I suggest the following immediate workaround for the offending tests: set the tests to use ILU0 as @bska demonstrated above. This is not a permanent, or even a good temporary solution. We must at least add error messages as discussed above.

For the actionx-related test though, maybe a proper fix is within reach, as work has been done in this area very recently. @lisajulia has the "include all well stuff from ACTIONX" work only focused on partitioning? Or did you also take this into account in other places? In any case, I think we now have a way to know to complete potential set of wells and completions from the very start, right?

lisajulia:
> The failure for compareSeparateECLFiles_flow+actionx_compdat_1_proc (and the NLDD variant of that test) also seems to be CPRW triggered.
> 
> I suggest the following immediate workaround for the offending tests: set the tests to use ILU0 as @bska demonstrated above. This is not a permanent, or even a good temporary solution. We must at least add error messages as discussed above.
> 
> For the actionx-related test though, maybe a proper fix is within reach, as work has been done in this area very recently. @lisajulia has the "include all well stuff from ACTIONX" work only focused on partitioning? Or did you also take this into account in other places? In any case, I think we now have a way to know to complete potential set of wells and completions from the very start, right?

Yes, I'm looking at the three ACTIONX + COMPDAT tests now :).

atgeirr:
> I'm looking at the three ACTIONX + COMPDAT tests now.

Great!

bska:
> note that the tests you are discussing are also mentioned [elsewhere] (with similar conclusions):

Thanks a lot for the reference.  You've done a lot of investigation here, clearly, and we'll need to take that work into account before merging this PR.

bska:
I looked a little more closely into the [`AQUFLUX-01`]() regression test failure.  For what it's worth, I can make the test pass on my local machine with the change below, i.e., by tightening the CNV and material balance tolerances.  I'd prefer doing this instead of once again raising the test's relative tolerance (cf. PR #5224, commit d4569dba7).

---

```diff
diff --git a/parallelTests.cmake b/parallelTests.cmake
index ed498f4f3..3112913d3 100644
--- a/parallelTests.cmake
+++ b/parallelTests.cmake
@@ -139,7 +139,7 @@ add_test_compare_parallel_simulation(CASENAME aquflux_01
                                      ABS_TOL ${abs_tol}
                                      REL_TOL 0.04
                                      DIR aquifers
-                                     TEST_ARGS --enable-tuning=true --enable-drift-compensation=false --relaxed-max-pv-fraction=0.0)
+                                     TEST_ARGS --enable-tuning=true --enable-drift-compensation=false --relaxed-max-pv-fraction=0.0 --tolerance-mb=1.0e-7 --tolerance-cnv=1.0e-3)
 
 add_test_compare_parallel_simulation(CASENAME aquflux_02
                                      FILENAME AQUFLUX-02
```

totto82:
> Can you have a look at this https://github.com/totto82/opm-simulators/pull/12?

Do you want me to merge it into my PR? Or do you plan to add it as a separate PR. I am fine with both. From my point of view your PR is fine.

lisajulia:
> > Can you have a look at this [totto82#12](https://github.com/totto82/opm-simulators/pull/12)?
> 
> Do you want me to merge it into my PR? Or do you plan to add it as a separate PR. I am fine with both. From my point of view your PR is fine.

Thanks :) I'd prefer you merging https://github.com/totto82/opm-simulators/pull/12 into your branch. The the other failing tests can be resolved from there :).

totto82:
jenkins build this opm-tests=1205 please

atgeirr:
With #5532 merged, there should be only the python tests left to fix!

bska:
> With #5532 merged, there should be only the python tests left to fix!

Hopefully.  We'll need to rebase this PR, though, to resolve the merge conflict.

totto82:
jenkins build this opm-tests=1205 please

totto82:
jenkins build this opm-tests=1205 please

totto82:
Do we need to run a new update reference run for these?

[mpi.compareECLFiles_flowexp_blackoil+SPE1CASE2](https://ci.opm-project.org/job/opm-simulators-PR-builder/6507/testReport/junit/(root)/mpi/compareECLFiles_flowexp_blackoil_SPE1CASE2/)
[mpi.compareECLFiles_flow+2D_POLYMER_INJECTIVITY](https://ci.opm-project.org/job/opm-simulators-PR-builder/6507/testReport/junit/(root)/mpi/compareECLFiles_flow_2D_POLYMER_INJECTIVITY/)
[mpi.compareECLFiles_flow+MICP](https://ci.opm-project.org/job/opm-simulators-PR-builder/6507/testReport/junit/(root)/mpi/compareECLFiles_flow_MICP/)

bska:
> Do we need to run a new update reference run for these?

Probably.  The `POLYMER_INJECTIVITY` results in the updated reference solutions (OPM/opm-tests#1205) is at least known to be faulty.  There is only a single report step's worth of simulation results ("SEQNUM" = 0, i.e., at the `START` time), but the real case is supposed to produce results for all sequence numbers (report steps) 0..91.  I don't know if running another `update_data` here will do the right thing or if we need to close OPM/opm-tests#1205 and create a completely new PR for this.

atgeirr:
> I don't know if running another `update_data` here will do the right thing or if we need to close [OPM/opm-tests#1205](https://github.com/OPM/opm-tests/pull/1205) and create a completely new PR for this.

Neither do I, I will try just rerunning with update_data here.

atgeirr:

atgeirr:
jenkins build this opm-tests=1205 please

akva2:
jenkins build this opm-tests=1205 please

atgeirr:
Only the two python tests remaining! I think they are way too strict: they request a comparison to two decimal places (after the comma), so we need to be within 0.01 Pa for that test to succeed. The "places" argument is not the number of significant digits... I suggest replacing "places=2" with "delta=1e4" (i.e. setting a 0.1 bar limit) in those two tests, as well as replacing the reference value 35795160.67 for the pressures with the new value 35795416.094404034.

There are more values that need to be fixed probably in the test_fluidstate_variables.py program, which requires very many digits, which is not expected to be stable to tons of digits when changing numerical parameters. That is, unless we solve the nonlinear equation to similarly strict tolerances (which is not done in this test).

@totto82 or @lisajulia perhaps you can fix this? It is the only remaining hurdle I think!

lisajulia:
@atgeirr and @totto82 I‘m on vacation and back on the 20th, sorry 🫣- if the two remaining tests are still failing then I’ll take a look :)

totto82:
I will give it a try.

atgeirr:
> I will give it a try.

Great! I do not think any more data updates should be necessary, so I hope trial and error should not be too time-consuming.

atgeirr:
Finally, this has been merged! Thanks to all who contributed with testing and fixing!

bska:
> Finally, this has been merged!

Just for the record, we're **likely** to see O(200) regression test failures in the post-commit build...

bska:
> > Just for the record, we're **likely** to see O(200) regression test failures in the post-commit build...
> 
> Waited until the post-build for opm-tests was queued before I pressed the button here, so hopefully we will not

Waiting until the job is in the queue is **generally** not sufficient for this.  To be absolutely safe, we need to wait until the job actually **completes** before merging the associate code changes/PRs.  There are mitigating circumstances in this particular case, however, which means that it'll hopefully work out nonetheless.  In particular, the currently running post-commit build is for opm-simulators as well and, since that is rather time consuming, it is likely that the opm-tests job will complete in time for the next post-commit build of opm-simulators.  There is nevertheless no guarantee since the post-commit build of opm-tests may fail for unknown reasons&ndash;e.g., because of the size of OPM/opm-tests#1205.

PR #5158, opened 2024-02-02
fixed: make sure Damaris is enabled before executing test

akva2:
Not just that it has been found.

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -405,7 +405,7 @@ if(MPI_FOUND)
                       REL_TOL 0.0)
 
   # Single test for damaris
-  if(Damaris_FOUND)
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this hip please

bska:
jenkins build this please

akva2:
okay, this will do, the hip build failed due to deprecation warnings but the post builder has no warning threshold so it won't hamper normal operation.

bska (approved):
> okay, this will do, the hip build failed due to deprecation warnings but the post builder has no warning threshold so it won't hamper normal operation.

Okay, understood.  In that case I'll merge into master.

PR #5161, opened 2024-02-06
Ignore Coordsys and print a warning

totto82:
COORDSYS is often used for single reservoir systems. Lets ignore it. Note that if NUMRES > 1 the simulator will stop.


totto82:
jenkins build this please

blattms:
I think we could actually make a check for NUMRES being 1 in opm/simulators/utils/PartiallySupportedFlowKeywords.hpp

blattms (approved):
thanks a lot. This seems to do the job.

PR #5166, opened 2024-02-07
Drop ebos in index functions

akva2:
Replaces 'ebos' with 'model'.


akva2:
jenkins build this please

PR #5168, opened 2024-02-07
Feature/remove files on simulation startup

lisajulia:
@bska @blattms @akva2 
I've added a 'cleanDeck' function in Opm namespace that removes possible old simulation output files from the specified output directory. This function is called during 'setupLogging' before the actual logging starts.

Can you have a look at this and give me some improvement tips/feedback?
Thanks :)!


opm/simulators/utils/readDeck.cpp
```diff
@@ -439,15 +452,21 @@ Opm::setupLogging(const int          mpi_rank_,
             : std::filesystem::current_path().generic_string();
     }
 
+    cleanDeck(baseName, output_dir);
```
akva2, line 455: should this run on all processes? of course shouldn't harm but also a bit of a eyebrow-raiser.

akva2 (approved):
One small question otherwise it looks good to me.

opm/simulators/utils/readDeck.cpp
lisajulia: Yes, I was thinking about this as well.
I saw that the check if the output directory output_dir exists (and creating it if it does not exist) is also done on all processes though... imho things like this should be consistent. What do you think?

akva2: i agree on consistency. this is okay then, we can consider addressing both afterwards.

bska: Regardless of what other mistakes we've made in this area, this particular function (`cleanDeck()`, although I might suggest `removePreviousResultFiles()` or `prepareResultOutputDirectory()` instead) should only run on the I/O rank and other ranks should wait for its completion.

opm/simulators/utils/readDeck.cpp
```diff
@@ -402,15 +402,28 @@ void Opm::ensureOutputDirExists(const std::string& cmdline_output_dir)
     }
 }
 
+void Opm::cleanDeck(const std::string& baseName,
+                    const std::string& outputDir)
```
bska, line 406: The `outputDir` should be a `std:filesystem:path` instead of a `std::string`.

bska (commented):
I think we're on the right track, but the solution must be more comprehensive before it's ready for review.  We should be doing this work in terms of `std::filesystem` operations and directory iteration to ensure we match **all** of the pertinent output files.  A simple list of filenames is not going to be sufficient.

lisajulia:
> I think we're on the right track, but the solution must be more comprehensive before it's ready for review. We should be doing this work in terms of `std::filesystem` operations and directory iteration to ensure we match **all** of the pertinent output files. A simple list of filenames is not going to be sufficient.

@bska: Can you have another look and let me know what still needs to be changed? Thanks!

lisajulia:
jenkins build this please

lisajulia:
jenkins build this please

akva2:
just in case; jenkins setup is based on polling so it takes up to 5 min to trigger.

opm/simulators/utils/readDeck.cpp
```diff
@@ -402,15 +403,28 @@ void Opm::ensureOutputDirExists(const std::string& cmdline_output_dir)
...
+                                       const std::filesystem::path& outputDir)  
+{   
+    //Loop over all files in the output directory and subdirectories and delete them if their name is baseName + a correct extension
+    std::regex r(baseName + ".(F?(DBG|E?GRID|INIT|PRT|RFT|SMSPEC|UNSMRY|UNRST)|([ABCFGHSTUXYZ]\\d{4})|(INFOSTEP|INFOITER|OPMRST))");
+    for (auto& file : std::filesystem::recursive_directory_iterator(outputDir)) {
+        std::string fileName = file.path().filename();
+        if (std::regex_match(fileName, r)) {
+            std::filesystem::remove(file);
+        }
+    }
```
bska, line 416: One **slight** alternative could be to match the `baseName` by string equality instead of regular expressions, e.g., along the lines of
```C++
const std::regex extension(R"~(\.F?(DBG|E?GRID|INIT|....|INFOITER|OPMRST))~");
for (const auto& elem : std::filesystem::recursive_directory_iterator(outputDir)) {
    const auto& pth = elem.path();
    if ((pth.stem() == baseName) && std::regex_match(pth.extension(), extension)) {
        std::filesystem::remove(elem);
    }
}
```
but it's not clear to me that that would be better.  I would advise using a raw string here, however, since that would simplify entering metacharacters&ndash;i.e., not having to double every backslash character.  We might additionally want to check `elem.is_regular_file()` for some additional safety, although it's not clear to me what to do if we encounter a directory with a matching name.

bska (approved):
This looks good to me and I'd be happy to merge the current version, possibly after switching the string literal to a raw string.

lisajulia:
jenkins build this please

opm/simulators/utils/readDeck.cpp
```diff
@@ -402,15 +403,28 @@ void Opm::ensureOutputDirExists(const std::string& cmdline_output_dir)
     }
 }
 
+void Opm::prepareResultOutputDirectory(const std::string&           baseName,
+                                       const std::filesystem::path& outputDir)  
+{   
+    //Loop over all files in the output directory and subdirectories and delete them if their name is baseName + a correct extension
+    std::regex r(baseName + R"(.(F?(DBG|E?GRID|INIT|PRT|RFT|SMSPEC|UNSMRY|UNRST)|([ABCFGHSTUXYZ]\d{4})|(INFOSTEP|INFOITER|OPMRST)))");
```
bska, line 410: Tiny nit: The full stop character (`.`) should be escaped here&ndash;i.e., the start of the regular expression should be written as `baseName + R"(\.(F?`.  Otherwise it will match any character and that's not the semantics we need here.

lisajulia:
jenkins build this please

bska:
jenkins build this please

bska (approved):

PR #5170, opened 2024-02-07
Improve computeWellPotentials() exception output

akva2:
In a model I'm looking into the simulator quits with the rather uninformative message:
```
Simulation aborted as program threw an unexpected exception: computeWellPotentials() failed:
```

The problem is that the exception does not happen on the root process, and thus the exception message
only contains the prefix. Instead of adding new code to sync the exception message between processes,
I opted to use the existing functionality for the deferred logger and add the exception message as errors in there.

Furthermore since we are queuing up exceptions here, we should also concatenate the messages from all exceptions.
The output is now a lot more informative,

```
Error: [Exception on rank 2]: computeWellPotentials() failed: 
For well WELL1: [/home/akva/kode/opm/opm-simulators/opm/simulators/wells/VFPHelpers.cpp:692] Nonexistent VFP table 0 referenced.
For well WELL2: [/home/akva/kode/opm/opm-simulators/opm/simulators/wells/VFPHelpers.cpp:692] Nonexistent VFP table 0 referenced.
Simulation aborted as program threw an unexpected exception: computeWellPotentials() failed:
```
Note that you still get the same "empty" exception in the end but all information is now conveyed to the user.

### DIFF
--- a/opm/simulators/utils/DeferredLoggingErrorHelpers.hpp
+++ b/opm/simulators/utils/DeferredLoggingErrorHelpers.hpp
@@ -114,6 +114,10 @@ inline void logAndCheckForExceptionsAndThrow(Opm::DeferredLogger& deferred_logge
                                              const bool terminal_output,
                                              Opm::Parallel::Communication comm)
 {
+    // add exception message to logger in order to display message from all ranks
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this please

atgeirr:
Good, merging!

PR #5171, opened 2024-02-07
BlackoilModel: drop ebos in variable names

akva2:
Either ebos -> model or ebos -> nothing

### DIFF
--- a/opm/simulators/flow/BlackoilModel.hpp
+++ b/opm/simulators/flow/BlackoilModel.hpp
@@ -629,7 +629,7 @@ namespace Opm {
         void solveJacobianSystem(BVector& x)
         {
             auto& jacobian = simulator_.model().linearizer().jacobian().istlMatrix();
-            auto& ebosResid = simulator_.model().linearizer().residual();
+            auto& residual = simulator_.model().linearizer().residual();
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

blattms:
Just my two cents:
Does changing ebos to model really make sense? Now we have modelSimulator, modeResId, etc. How about stripping model and going with simulator, resId, etc,

akva2:
I am open to any changes, and I do agree it seems maybe a bit redundant.  But before I do a second round I do insist on consensus to avoid back and forth.

blattms:
To be honest, I don't really care and can live with everything. Anything is better than ebos.

atgeirr:
I think "model" is fine if not perfect. There is no perfect name change here perhaps... At least I do not have a better proposal! I think this is good, and as @blattms I can live with any reasonable alternative.

akva2:
since there is no strong opinion I went with my own preference which is to drop the prefixes.

akva2:
jenkins build this please

bska (approved):
I prefer dropping `ebos` too, so the current version is fine by me.  I'll merge into master.

PR #5172, opened 2024-02-07
Implement WELL as group limit action

totto82:
Depends on https://github.com/OPM/opm-common/pull/3894

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -2132,7 +2132,6 @@ namespace Opm {
             updateAndCommunicate(reportStepIdx, iterationIdx, deferred_logger);
         }
 
-        //if (iterationIdx > 1) {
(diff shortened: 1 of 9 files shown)


opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2129,19 +2131,24 @@ namespace Opm {
             changed = true;
             updateAndCommunicate(reportStepIdx, iterationIdx, deferred_logger);
         }
+
+        //if (iterationIdx > 1) {
```
bska, line 2135: This (`// if (...)`) looks like leftover debugging code.  If it is, then please remove it.  If not, please add a comment for why it is or will be needed.

bska (approved):
As far as I can tell this does do what it claims to do.  I hasten to add that this is a part of the simulator that I'm not too familiar with, so I'll accept blame if I've overlooked some subtle detail here.

In any case, I'll merge into master.

PR #5179, opened 2024-02-12
move ebos/eclbasevanguard.hh to opm/simulators/flow

akva2:
BaseVanguard is taken so name the class FlowBaseVanguard.


akva2:
jenkins build this please

PR #5183, opened 2024-02-12
move ebos/eclfluxmodule.hh to opm/simulators/flow

akva2:
I opted to call this NewTranFluxModule


akva2:
jenkins build this please

PR #5205, opened 2024-02-15
-fixed use of intensivequantites

hnil:
Remove use of changedIntensiveQuantites in case one would like to reduce memory on the cost of more expensive assembly and update.


bska:
jenkins build this please

hnil:
Closing: three of the four touched files (eclproblem.hh, eclwriter.hh, damariswriter.hh) no longer exist after the Ebos rename, so this cannot be rebased in any meaningful sense. Branch retained on my fork if the intensive-quantity caching question is revisited.

PR #5206, opened 2024-02-15
[bugfix?] Do terminal output on rank zero in time stepper even for restart

blattms:
Seeing this felt a bit strange. I assume the with restart there might have been some messages missing.

To me this PR results in more consistent behavior. I could very well be wrong, though.

### DIFF
--- a/opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp
+++ b/opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp
@@ -310,7 +310,7 @@ public:
             if (isRestart()) {
                 // For restarts the ebosSimulator may have gotten some information
                 // about the next timestep size from the OPMEXTRA field
-                adaptiveTimeStepping_->setSuggestedNextStep(ebosSimulator_.timeStepSize());
+                adaptiveTimeStepping_->setSuggestedNextStep(ebosSimulator_.timeStepSize(), terminalOutput_);
             }
         }
     }


blattms:
jenkins build this please

opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp
```diff
@@ -310,7 +310,7 @@ class SimulatorFullyImplicitBlackoil : private SerializableSim
             if (isRestart()) {
                 // For restarts the ebosSimulator may have gotten some information
                 // about the next timestep size from the OPMEXTRA field
-                adaptiveTimeStepping_->setSuggestedNextStep(ebosSimulator_.timeStepSize());
+                adaptiveTimeStepping_->setSuggestedNextStep(ebosSimulator_.timeStepSize(), terminalOutput_);
```
blattms, line 313: :flushed:  I probably should have tried to compile this. Now I am wondering what constructor was actually called by the original code.

blattms: As the constructors of UnitSystem taking one argument are all marked as explicit, It must have been the other [constructor](/OPM/opm-simulators/blob/780f282ed8686d6776a5c5ff713f577352a1b753/opm/simulators/timestepping/AdaptiveTimeStepping.hpp#L278), which means my fix is wrong and won't compile.

blattms: In that case it seems rather strange that we did not pass unitSystem but created one with the default constructor. I must be missing something here.

blattms:
Never mind. I was blind.

PR #5207, opened 2024-02-15
[fix] Use only EnableTuning consistently (and not also EclEnableTuning)

blattms:
In some part queried the former and in most parts the latter. So user had to actually use both `--enable-tuning=true --ecl-enable-tuning`) to achieve what they wanted.

PR #5208, opened 2024-02-15
Apply NEXTSTEP even without TUNING enabled.

blattms:
Previously, it was ignored if the user did not run OPM-FlOW explicitly using the command line parameter --enable-tuning=true. Now it will be applied regardless which parameter is used.

Includes #5207


blattms:
jenkins build this please

blattms:
This is trying to accomplish something similar to #4965. Would be good if @vkip would take a look.

blattms:
Maybe my approach was a little naive. This changes results for a lot of cases that do not have NEXTSTEP (only 02_TUNING_TSINIT_NEXTSTEP.DATA has it and that should probably not even change).

blattms:
Convert to draft does not work for me. Maybe I need to use Internet Explorer? :angry:

vkip:
> This is trying to accomplish something similar to #4965. Would be good if @vkip would take a look. (Sorry, I did not see the other PR when looking into this)

#4965 is only slightly related, but I did touch NEXTSTEP in https://github.com/OPM/opm-common/pull/3528, so can have a look.

vkip:
The tests that use '--enable-tuning=true' seem to fail. I _think_ the problem is that the EclEnableTuning parameter has not been in use for quite some time, so that the code block in beginEpisode_ was previously not used (both the option and the code blocks that only run with --ecl-enable-tuning=true should probably be removed). If you leave the EnableTuning flag in it's original location the remaining looks fine as far as I can see. Tests will tell..

blattms:
jenkins build this please

vkip:
I see I was wrong earlier, sorry about that. After another (more careful) look I think I recall how this works: max_next_tstep() will take the max of the next time step possibly specified in TUNING and NEXTSTEP, so without enableTUNING the former should be ignored. I think the simples might be to add enableTUNING as a parameter to https://github.com/OPM/opm-common/blob/1859da35d7511a48e335e173f469c8c84acbc352/src/opm/input/eclipse/Schedule/ScheduleState.cpp#L382-L386 and modify the condition for tuning_value accordingly. 

In addition, updateNEXTSTEP should only be called if there is an actual TUNING_CHANGE event (either from a NEXTSTEP record or persistent from a previous record). 

And finally a tiny nit: it is not necessary to call updateNEXTSTEP if updateTUNING in AdaptiveTimeStepping has already been called.

blattms:
Closing this one in favor of #5215

PR #5211, opened 2024-02-16
Flows: simplify data structure

akva2:
Replace pair with string and pair of intvec and doublevec with a struct with named entries.


akva2:
jenkins build this please

atgeirr:
All good, merging.

PR #5213, opened 2024-02-16
remove duplicate EclEnableTuning parameter

blattms:
In some part we queried the former and in most parts the latter. So user had to actually use both `--enable-tuning=true --ecl-enable-tuning`) to achieve what they wanted.

Fabulous work from @akva2


blattms:
jenkins build this please

blattms:
In it goes as this is not  a self-merge :grinning:

PR #5215, opened 2024-02-19
Ensure NEXTSTEP is respected also without --enable-tuning=true

vkip:
Previously, it was ignored if the user did not run OPM-FlOW explicitly using the command line parameter --enable-tuning=true. Now it will be applied regardless which parameter is used.

### DIFF
--- a/ebos/eclgenericproblem_impl.hh
+++ b/ebos/eclgenericproblem_impl.hh
@@ -473,7 +473,7 @@ beginEpisode_(bool enableExperiments,
     {
         const auto& sched_state = schedule_[episodeIdx];
         const auto& tuning = sched_state.tuning();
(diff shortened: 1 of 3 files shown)


vkip:
jenkins build this opm-common=3944 please

blattms:
Just referencing OPM/opm-common#3944 as upstream for convenience

blattms:
In contrast to my attempt in #5208, this one passes all the tests.

vkip:
> Looks good to me. I have some test cases from @tskille to check the implementation that I can run on this one. Just tell me when a good time is to do that and I will run them

Great, just had to check that my understanding was correct (this time), and since this seems ok in terms of passing the tests I can take them out of draft and you can proceed to run the cases from @tskille any time.

blattms:
Perfect, results look decent:
image
image

PR #5220, opened 2024-02-21
Respect low parsing strictness in actions 

vkip:
(when reprocessing the schedule section after action application.)

### DIFF
--- a/opm/simulators/utils/readDeck.cpp
+++ b/opm/simulators/utils/readDeck.cpp
@@ -563,10 +563,9 @@ void Opm::readDeck(Opm::Parallel::Communication    comm,
                          summaryConfig, std::move(python), initFromRestart,
                          checkDeck, treatCriticalAsNonCritical, outputInterval, *errorGuard);
 
-            if (treatCriticalAsNonCritical) { // Update schedule so that re-parsing after actions use same strictness
-                assert(schedule);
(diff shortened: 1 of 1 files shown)


opm/simulators/utils/readDeck.cpp
```diff
@@ -557,12 +557,16 @@ void Opm::readDeck(Opm::Parallel::Communication    comm,
...
             }
             readOnIORank(comm, deckFilename, parseContext.get(),
                          eclipseState, schedule, udqState, actionState, wtestState,
                          summaryConfig, std::move(python), initFromRestart,
                          checkDeck, treatCriticalAsNonCritical, outputInterval, *errorGuard);
+
+            if (treatCriticalAsNonCritical) { // Update schedule so that re-parsing after actions use same strictness
+                assert(schedule);
+                schedule->treat_critical_as_non_critical(true);
+            }
```
bska, line 569: I guess you could, strictly speaking, write this as
```C++
schedule->treat_critical_as_non_critical(treatCriticalAsNonCritical);
```
instead.  It's not clear to me if that's easier to read though.

vkip: Ah, I think that's slightly better. Thanks, will update.

vkip:
jenkins build this opm-common=3947 please

PR #5222, opened 2024-02-22
[Fix] Correct handling of NEXTSTEP in ACTIONX

blattms:
Previously, the step size would only have used for the first time step of the next report step (at least of --enable-tuning=true was used).

Needs OPM/opm-common#3949

### DIFF
--- a/opm/simulators/timestepping/AdaptiveSimulatorTimer.cpp
+++ b/opm/simulators/timestepping/AdaptiveSimulatorTimer.cpp
@@ -62,7 +62,7 @@ namespace Opm
         return ( report_step_ == 0 ) && ( current_step_ == 0 );
     }
 
-AdaptiveSimulatorTimer& AdaptiveSimulatorTimer::operator++ ()
+    AdaptiveSimulatorTimer& AdaptiveSimulatorTimer::operator++ ()
     {
         ++current_step_;
         current_time_ += dt_;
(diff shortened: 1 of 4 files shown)


blattms:
I was haunting ghosts. This actually works as is with and without tuning.

blattms:
There is one thing though that OPM-flow does different when it sees a TUNING keyword: we will not only apply to the next report step but to all of them.

opm/simulators/timestepping/AdaptiveSimulatorTimer.cpp
```diff
@@ -115,6 +115,12 @@ AdaptiveSimulatorTimer& AdaptiveSimulatorTimer::operator++ ()
         return dt_;
     }
 
+    void AdaptiveSimulatorTimer::setCurrentStepLength(double dt)
+    {
+        assert(dt > 0);
```
akva2, line 120: this dt can be user provided i believe?

opm/simulators/timestepping/AdaptiveTimeStepping.hpp
```diff
@@ -367,6 +373,7 @@ std::set<std::string> consistentlyFailingWells(const std::vector<StepReport>& sr
             auto& modelSimulator = solver.model().simulator();
             auto& ebosProblem = modelSimulator.problem();
 
+            std::cout << "Creating  suggestedNextTimestep_="<<suggestedNextTimestep_<<std::endl;
```
akva2, line 376: no cout'ing. two more below.

akva2 (commented):
The forementioned niggles.

opm/simulators/timestepping/AdaptiveSimulatorTimer.cpp
blattms: the user is the simulator/common code here. The calling site should actually make sure that dt is appropriate, see e.g. [ScheduleState.cpp#L385](https://github.com/OPM/opm-common/blob/e4b5665cd875171da1affda36136f0e32e4a9f0a/opm/input/eclipse/Schedule/ScheduleState.cpp#L385).

akva2:
jenkins build this please

blattms:
thanks a lot for the review.

PR #5225, opened 2024-02-26
increasing RelaxedPressureTolMsw default to be 2.e4

GitPaean:
there is one testing case gets stuck with pressure residual 1.557e4 for one of the segments of a well. 

This PR is for testing purpose to check the impact of the new default pressure tolerance.


GitPaean:
jenkins build this please

PR #5226, opened 2024-02-27
changed: put SupportsFaceTag in separate header

akva2:
this way we don't have to pull in alugrid and polyhedralgrid in simulator objects that does not use them


akva2:
jenkins build this please

akva2:
jenkins build this please

bska (approved):
Thanks.  This makes sense to me and I'll merge into master.

PR #5229, opened 2024-02-27
WellGroupHelpers: use type alias instead of template parameter for Comm

akva2:


akva2:
jenkins build this please

PR #5231, opened 2024-02-28
Add Utility for Region-Based Phase-Filled Averages

bska:
The initial use case is calculating the phase-filled pore-volume weighted average of the fluid mass densities per PVT region.  This value goes into calculating depth-corrected per-cell phase pressure values such as the `BPPO` and `BPPG` summary vectors.

This class manages a single linear array which separately tracks the averages' numerators and denominators as running sums per region and region set.  We pick this data structure to simplify the cross-rank reduction needed in MPI parallel runs.  Client code is expected to add individual per-cell and per-phase contributions using the `addCell()` member function and then call the `accumulateParallel()` member function to affect the cross-rank reduction.  The averages will then be available through the `fieldValue()` and `value()` member functions.

As a further view towards the initial use case, we track two different types of average per phase&ndash;one for the phase-filled volume and one for the pore-volume filled volume.  The latter is the average we would get for the case of the phase saturation being one throughout the region.  This alternative value is the fallback option for the case of the phase saturation being identically zero throughout the region.

### DIFF
--- a/opm/simulators/flow/RegionPhasePVAverage.hpp
+++ b/opm/simulators/flow/RegionPhasePVAverage.hpp
@@ -1,0 +1,353 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  Copyright 2024 Equinor AS
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef OPM_REGION_PHASE_POREVOL_AVERAGE_MODULE_HPP
+#define OPM_REGION_PHASE_POREVOL_AVERAGE_MODULE_HPP
+
+#include <opm/simulators/utils/ParallelCommunication.hpp>
+
+#include <cstddef>
+#include <functional>
+#include <string>
+#include <string_view>
+#include <vector>
+
+namespace Opm {
+
+    /// Facility for calculating volume-weighted average function values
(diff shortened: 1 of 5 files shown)


bska:
jenkins build this please

opm/simulators/flow/RegionPhasePVAverage.hpp
```diff
@@ -0,0 +1,352 @@
...
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef ECL_REGION_PHASE_POREVOL_AVERAGE_MODULE_HPP
```
akva2, line 22: maybe use opm in header guard?

opm/simulators/flow/RegionPhasePVAverage.hpp
```diff
@@ -0,0 +1,352 @@
...
+        };
+
+        /// Call-back function type for accessing region arrays--typically the FIP* arrays.
+        using RegionArrayAccessor = std::function<const std::vector<int>&(const std::string&)>;
+
+        /// Constructor
+        ///
+        /// \param[in] comm grid level global communicator.
+        ///
+        /// \param[in] numPhases Number phases for which to caluclate average values
```
akva2, line 78: calculate

opm/simulators/flow/RegionPhasePVAverage.hpp
```diff
@@ -0,0 +1,352 @@
...
+        /// Number of phases for which to accumulate function value averages.
+        std::size_t np_{};
+
+        /// Named region sets.
+        std::vector<std::string> rsetNames_{};
+
+        /// Call-back function for accessing region set arrays.
+        RegionArrayAccessor getRegionArray_;
+
+        /// Start pointers for average values of eached named region set.
```
akva2, line 168: each

akva2 (approved):
Just a couple of doxy typos, otherwise it looks good to me

opm/simulators/flow/RegionPhasePVAverage.hpp
bska: > maybe use opm in header guard?

Good idea.  I've pushed an update to that effect.

bska:
Thanks a lot for the review.  I've pushed an update to address the comments and the new shadowing warnings that I didn't pay enough attention to initially.  I'll run a new build check.

bska:
jenkins build this please

bska:
Merging per approval.

PR #4572, opened 2023-03-30
stricter tolerance for wells with zero rate target

GitPaean:
for StandardWell only at this moment.

It improves the results of a reported case with **zero RESV** control (black line referee result, green line master branch, red line this PR ). 

image

And more importantly, with the defaulted relative slack tolerance, there is some error when coming to the connection rates and the well rates (**0** for this circumstance), potentially affecting the mass balance of the whole system (it means the flow out from the reservoir is not equal the flow into the reservoir through the wellbore with **0** well rates).  

Ideally stricter tolerance should be applied to all the wells (https://github.com/OPM/opm-simulators/pull/4556), but not sure how we should test and proceed with that direction, there might be some substantial investigation efforts involved.  This PR is implemented as a temporary approach to fix the immediate issue from the reported case.

PR #4556, opened 2023-03-26
[testing] Tighten tolerance standard wells to 1.e-8

GitPaean:

PR #5232, opened 2024-02-28
When checking for zero target rates, also check wells under group control where required

steink:
Several places in the code we check whether a well is set to inject/produce at zero rate since this case requires special treatment. This is currently done by the function `wellUnderZeroRateTarget` (often called from `stopppedOrZeroRateTarget`). Wells under group control are currently not included in this check, although the _special treatments_ are equally required if a well is assigned zero rate from a group.

I've written draft versions of the two above mentioned functions that also take group-controlled wells into account. Would just like some feedback/discussion whether this seems reasonable before I go ahead and suggest updating all relevant function calls (the new functions have temporary names `*Version`).   

I've verified that this approach fixes an issue with `iterateWellEqWithSwitching` which currently does not behave nicely for wells that get assigned zero rate from group.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -162,11 +162,11 @@ namespace Opm
     template <typename TypeTag>
     void
     MultisegmentWell<TypeTag>::
-    updatePrimaryVariables(const SummaryState& summary_state,
+    updatePrimaryVariables(const Simulator& simulator,
                            const WellState<Scalar>& well_state,
-                           DeferredLogger& /* deferred_logger */)
+                           DeferredLogger& deferred_logger)
     {
-        const bool stop_or_zero_rate_target = this->stopppedOrZeroRateTarget(summary_state, well_state);
+        const bool stop_or_zero_rate_target = this->stoppedOrZeroRateTarget(simulator, well_state, deferred_logger);
         this->primary_variables_.update(well_state, stop_or_zero_rate_target);
     }
 
@@ -199,7 +199,7 @@ namespace Opm
     template <typename TypeTag>
     ConvergenceReport
     MultisegmentWell<TypeTag>::
-    getWellConvergence(const SummaryState& /* summary_state */,
+    getWellConvergence(const Simulator& /* simulator */,
                        const WellState<Scalar>& well_state,
                        const std::vector<Scalar>& B_avg,
                        DeferredLogger& deferred_logger,
@@ -260,7 +260,7 @@ namespace Opm
     template <typename TypeTag>
     void
     MultisegmentWell<TypeTag>::
-    recoverWellSolutionAndUpdateWellState(const SummaryState& summary_state,
+    recoverWellSolutionAndUpdateWellState(const Simulator& simulator,
                                           const BVector& x,
                                           WellState<Scalar>& well_state,
                                           DeferredLogger& deferred_logger)
@@ -271,7 +271,7 @@ namespace Opm
 
         BVectorWell xw(1);
         this->linSys_.recoverSolutionWell(x, xw);
-        updateWellState(summary_state, xw, well_state, deferred_logger);
+        updateWellState(simulator, xw, well_state, deferred_logger);
     }
 
 
@@ -578,7 +578,7 @@ namespace Opm
     template <typename TypeTag>
     void
     MultisegmentWell<TypeTag>::
-    solveEqAndUpdateWellState(const SummaryState& summary_state,
+    solveEqAndUpdateWellState(const Simulator& simulator,
                               WellState<Scalar>& well_state,
                               DeferredLogger& deferred_logger)
     {
@@ -589,7 +589,7 @@ namespace Opm
         try{
             const BVectorWell dx_well = this->linSys_.solve();
 
-            updateWellState(summary_state, dx_well, well_state, deferred_logger);
+            updateWellState(simulator, dx_well, well_state, deferred_logger);
         }
         catch(const NumericalProblem& exp) {
             // Add information about the well and log to deferred logger
@@ -681,7 +681,7 @@ namespace Opm
     template <typename TypeTag>
     void
     MultisegmentWell<TypeTag>::
-    updateWellState(const SummaryState& summary_state,
+    updateWellState(const Simulator& simulator,
                     const BVectorWell& dwells,
                     WellState<Scalar>& well_state,
                     DeferredLogger& deferred_logger,
@@ -691,15 +691,20 @@ namespace Opm
 
         const Scalar dFLimit = this->param_.dwell_fraction_max_;
         const Scalar max_pressure_change = this->param_.max_pressure_change_ms_wells_;
-        const bool stop_or_zero_rate_target = this->stopppedOrZeroRateTarget(summary_state, well_state);
+        const bool stop_or_zero_rate_target =
+            this->stoppedOrZeroRateTarget(simulator, well_state, deferred_logger);
         this->primary_variables_.updateNewton(dwells,
                                               relaxation_factor,
                                               dFLimit,
                                               stop_or_zero_rate_target,
                                               max_pressure_change);
 
-        this->primary_variables_.copyToWellState(*this, getRefDensity(), stop_or_zero_rate_target,
-                                                 well_state, summary_state, deferred_logger);
+        const auto& summary_state = simulator.vanguard().summaryState();
+        this->primary_variables_.copyToWellState(*this, getRefDensity(),
+                                                 stop_or_zero_rate_target,
+                                                 well_state,
+                                                 summary_state,
+                                                 deferred_logger);
 
         {
             auto& ws = well_state.well(this->index_of_well_);
@@ -720,8 +725,7 @@ namespace Opm
                                 const WellState<Scalar>& well_state,
                                 DeferredLogger& deferred_logger)
     {
-        const auto& summary_state = simulator.vanguard().summaryState();
-        updatePrimaryVariables(summary_state, well_state, deferred_logger);
+        updatePrimaryVariables(simulator, well_state, deferred_logger);
         initPrimaryVariablesEvaluation();
         computePerfCellPressDiffs(simulator);
         computeInitialSegmentFluids(simulator);
@@ -1489,8 +1493,7 @@ namespace Opm
                 this->regularize_ = true;
             }
 
-            const auto& summary_state = simulator.vanguard().summaryState();
-            const auto report = getWellConvergence(summary_state, well_state, Base::B_avg_, deferred_logger, relax_convergence);
+            const auto report = getWellConvergence(simulator, well_state, Base::B_avg_, deferred_logger, relax_convergence);
             if (report.converged()) {
                 converged = true;
                 break;
@@ -1526,7 +1529,7 @@ namespace Opm
                     ++stagnate_count;
                     if (stagnate_count == 6) {
                         sstr << " well " << this->name() << " observes severe stagnation and/or oscillation. We relax the tolerance and check for convergence. \n";
-                        const auto reportStag = getWellConvergence(summary_state, well_state, Base::B_avg_, deferred_logger, true);
+                        const auto reportStag = getWellConvergence(simulator, well_state, Base::B_avg_, deferred_logger, true);
                         if (reportStag.converged()) {
                             converged = true;
                             sstr << " well " << this->name() << " manages to get converged with relaxed tolerances in " << it << " inner iterations";
@@ -1553,7 +1556,7 @@ namespace Opm
                 this->regularize_ = true;
                 deferred_logger.debug(sstr.str());
             }
-            updateWellState(summary_state, dx_well, well_state, deferred_logger, relaxation_factor);
+            updateWellState(simulator, dx_well, well_state, deferred_logger, relaxation_factor);
             initPrimaryVariablesEvaluation();
         }
 
@@ -1635,7 +1638,7 @@ namespace Opm
         const bool allow_open =  this->well_ecl_.getStatus() == WellStatus::OPEN &&
                                  well_state.well(this->index_of_well_).status == WellStatus::OPEN;
         // don't allow switcing for wells under zero rate target or requested fixed status and control
-        const bool allow_switching = !this->wellUnderZeroRateTarget(summary_state, well_state) &&
+        const bool allow_switching = !this->wellUnderZeroRateTarget(simulator, well_state, deferred_logger) &&
                                      (!fixed_control || !fixed_status) && allow_open;
         bool changed = false;
         bool final_check = false;
@@ -1673,7 +1676,7 @@ namespace Opm
                 this->regularize_ = true;
             }
 
-            const auto report = getWellConvergence(summary_state, well_state, Base::B_avg_, deferred_logger, relax_convergence);
+            const auto report = getWellConvergence(simulator, well_state, Base::B_avg_, deferred_logger, relax_convergence);
             converged = report.converged();
             if (converged) {
                 // if equations are sufficiently linear they might converge in less than min_its_after_switch
@@ -1717,7 +1720,7 @@ namespace Opm
                         if (false) { // this disables the usage of the relaxed tolerance
                             fmt::format_to(std::back_inserter(message), " Well {} observes severe stagnation and/or oscillation."
                                                                         " We relax the tolerance and check for convergence. \n", this->name());
-                            const auto reportStag = getWellConvergence(summary_state, well_state, Base::B_avg_,
+                            const auto reportStag = getWellConvergence(simulator, well_state, Base::B_avg_,
                                                                        deferred_logger, true);
                             if (reportStag.converged()) {
                                 converged = true;
@@ -1745,7 +1748,7 @@ namespace Opm
                     deferred_logger.debug(message);
                 }
             }
-            updateWellState(summary_state, dx_well, well_state, deferred_logger, relaxation_factor);
+            updateWellState(simulator, dx_well, well_state, deferred_logger, relaxation_factor);
             initPrimaryVariablesEvaluation();
         }
 
@@ -1906,10 +1909,11 @@ namespace Opm
                 }
             }
 
-            // the fourth dequation, the pressure drop equation
+            // the fourth equation, the pressure drop equation
             if (seg == 0) { // top segment, pressure equation is the control equation
                 const auto& summaryState = simulator.vanguard().summaryState();
                 const Schedule& schedule = simulator.vanguard().schedule();
+                const bool stopped_or_zero_target = this->stoppedOrZeroRateTarget(simulator, well_state, deferred_logger);
                 MultisegmentWellAssemble(*this).
                         assembleControlEq(well_state,
                                         group_state,
@@ -1920,6 +1924,7 @@ namespace Opm
                                         getRefDensity(),
                                         this->primary_variables_,
                                         this->linSys_,
+                                        stopped_or_zero_target,
                                         deferred_logger);
             } else {
                 const UnitSystem& unit_system = simulator.vanguard().eclState().getDeckUnitSystem();
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -462,6 +462,7 @@ namespace Opm
 
         const auto& summaryState = simulator.vanguard().summaryState();
         const Schedule& schedule = simulator.vanguard().schedule();
+        const bool stopped_or_zero_target = this->stoppedOrZeroRateTarget(simulator, well_state, deferred_logger);
         StandardWellAssemble<FluidSystem,Indices>(*this).
             assembleControlEq(well_state, group_state,
                               schedule, summaryState,
@@ -469,6 +470,7 @@ namespace Opm
                               this->primary_variables_,
                               this->connections_.rho(),
                               this->linSys_,
+                              stopped_or_zero_target,
                               deferred_logger);
 
 
@@ -692,16 +694,17 @@ namespace Opm
     template<typename TypeTag>
     void
     StandardWell<TypeTag>::
-    updateWellState(const SummaryState& summary_state,
+    updateWellState(const Simulator& simulator,
                     const BVectorWell& dwells,
                     WellState<Scalar>& well_state,
                     DeferredLogger& deferred_logger)
     {
         if (!this->isOperableAndSolvable() && !this->wellIsStopped()) return;
 
-        const bool stop_or_zero_rate_target = this->stopppedOrZeroRateTarget(summary_state, well_state);
+        const bool stop_or_zero_rate_target = this->stoppedOrZeroRateTarget(simulator, well_state, deferred_logger);
         updatePrimaryVariablesNewton(dwells, stop_or_zero_rate_target, deferred_logger);
 
+        const auto& summary_state = simulator.vanguard().summaryState();
         updateWellStateFromPrimaryVariables(stop_or_zero_rate_target, well_state, summary_state, deferred_logger);
         Base::calculateReservoirRates(well_state.well(this->index_of_well_));
     }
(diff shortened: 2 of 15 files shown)


GitPaean:
jenkins build this please

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1396,6 +1396,70 @@ namespace Opm
         }
     }
 
+    template<typename TypeTag>
+    bool
+    WellInterface<TypeTag>::
+    wellUnderZeroRateTargetVersion(const Simulator& ebos_simulator,
```
GitPaean, line 1402: I think you can use `wellUnderZeroRateTargetVersion` to replace the original `wellUnderZeroRateTarget`

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1396,6 +1396,70 @@ namespace Opm
...
+        const auto& ws = well_state.well(this->index_of_well_);
+        const auto& summaryState = ebos_simulator.vanguard().summaryState();
+        const bool isGroupControlled = (this->isInjector() && ws.injection_cmode == Well::InjectorCMode::GRUP) ||
+                                       (this->isProducer() && ws.production_cmode == Well::ProducerCMode::GRUP);
+        if (!isGroupControlled) {
+            // well is not under group control, check "light-weight" version
+            return this->wellUnderZeroRateTarget(summaryState, well_state);
+        } else {
+            const auto& well = this->well_ecl_;
+            const auto& schedule = ebos_simulator.vanguard().schedule();
```
GitPaean, line 1416: I think the only reason we need `Simulator` is because of `schedule`, maybe also the `GroupState`. We can pass in the `schedule` as a parameter, so the function can be moved to `WellInterfaceGeneric`.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1396,6 +1396,70 @@ namespace Opm
...
+            const auto& group_state = ebos_simulator.problem().wellModel().groupState();
+            const auto& group = schedule.getGroup(well.groupName(), this->currentStep());
+            const double efficiencyFactor = well.getEfficiencyFactor();
+            if (this->isInjector()) {
```
GitPaean, line 1420: @atgeirr , do you have some comment regarding how to determine the group rate target is zero?

Maybe we should not compared directly with `0.0` to consider the rounding error. 

There is more code in the group rate target determination, maybe they can go in a separate function.

opm/simulators/wells/WellInterface_impl.hpp
steink: Thanks, good suggestion.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -159,7 +159,7 @@ namespace Opm
                            const WellState& well_state,
                            DeferredLogger& /* deferred_logger */)
     {
-        const bool stop_or_zero_rate_target = this->stopppedOrZeroRateTarget(summary_state, well_state);
+        const bool stop_or_zero_rate_target = this->stoppedOrZeroRateTargetIndividual(summary_state, well_state);
```
steink, line 162: I don't think setting the `WQTotal` of first segment explicitly to zero for zero-target wells is strictly necessary below, so have left out the checking for zero targets from groups here.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -684,7 +684,7 @@ namespace Opm
 
         const double dFLimit = this->param_.dwell_fraction_max_;
         const double max_pressure_change = this->param_.max_pressure_change_ms_wells_;
-        const bool stop_or_zero_rate_target = this->stopppedOrZeroRateTarget(summary_state, well_state);
+        const bool stop_or_zero_rate_target = this->stoppedOrZeroRateTargetIndividual(summary_state, well_state);
```
steink, line 687: Same comment as above.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -692,7 +694,7 @@ namespace Opm
     {
         if (!this->isOperableAndSolvable() && !this->wellIsStopped()) return;
 
-        const bool stop_or_zero_rate_target = this->stopppedOrZeroRateTarget(summary_state, well_state);
+        const bool stop_or_zero_rate_target = this->stoppedOrZeroRateTargetIndividual(summary_state, well_state);
```
steink, line 697: Same comment as for MS-wells

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -1183,7 +1185,7 @@ namespace Opm
         constexpr double stopped_factor = 1.e-4;
         // use stricter tolerance for dynamic thp to ameliorate network convergence
         constexpr double dynamic_thp_factor = 1.e-1;
-        if (this->stopppedOrZeroRateTarget(summary_state, well_state)) {
+        if (this->stoppedOrZeroRateTargetIndividual(summary_state, well_state)) {
             tol_wells = tol_wells*stopped_factor;
```
steink, line 1189: I think it's a bit strange that zero target wells should be solved with a much stricter tollerance than others. For convenience, I've also skipped checking for zero targets from groups here.

steink:
I've restructered everything here, and renamed the original functions `wellUnderZeroRateTarget` and `stoppedOrZeroRateTarget` to `wellUnderZeroRateTargetIndividual` and `stoppedOrZeroRateTargetIndividual`, respectively. The functions `wellUnderZeroRateTarget` and `stoppedOrZeroRateTarget` now also include checking for zero targets from groups. I've updated the usage of these functions where I see it necessary (like forming of control equations and in the `iterateWellEqWithSwitching`-versions). 
Since the checking for zero targets from groups is a bit more involved, I've skipped this where I don't think it is important (see comments). Hence, these places now call the `*Individual` versions and the behaviour is unchanged from master.

GitPaean:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
GitPaean: It can be debatable. If we want to be a little more systematical, we probably want to remove this `stop_or_zero_rate_target` by all if it is not necessary. 

I think it is not a problem when it is with the correct sign, and it might be a problem when the sign of the `WQTotal` switches, which is easy for the wells without a zero rate target. 

So far we do not know how to handle wells with the opposite production/injection direction.

opm/simulators/wells/StandardWell_impl.hpp
GitPaean: There might be different approaches to address this, but it improves the results for some problematic cases. 

https://github.com/OPM/opm-simulators/pull/4572
https://github.com/OPM/opm-simulators/pull/4556

Possibly we should use stricter tolerance by all.

GitPaean:
> I've restructered everything here, and renamed the original functions `wellUnderZeroRateTarget` and `stoppedOrZeroRateTarget` to `wellUnderZeroRateTargetIndividual` and `stoppedOrZeroRateTargetIndividual`, respectively. The functions `wellUnderZeroRateTarget` and `stoppedOrZeroRateTarget` now also include checking for zero targets from groups. I've updated the usage of these functions where I see it necessary (like forming of control equations and in the `iterateWellEqWithSwitching`-versions). Since the checking for zero targets from groups is a bit more involved, I've skipped this where I don't think it is important (see comments). Hence, these places now call the `*Individual` versions and the behaviour is unchanged from master.

Thanks for the update.  Will need to think through and get back with suggestions as a whole.

GitPaean:
At the same time, can you check the Jenkins failures to see whether the failures are relevant?  Especially the ones with related to `MOD4_GRP`.

steink:
> At the same time, can you check the Jenkins failures to see whether the failures are relevant? Especially the ones with related to `MOD4_GRP`.

Yes, will look into what's going on with these.

steink:
Have looked at the four failing tests here. _NETWORK-01-REROUTE__ and _NETWORK-01-REROUTE_STD_ match very closely. For _MOD4_GRP_GEFAC_ there is slightly different time-stepping with the PR, but otherwise close match. 

Looked a bit more closely into _MOD4_GRP_. This is a very problematic test where both MASTER and PR struggles a lot towards the end with control switching oscillations resulting in extremely short time steps. PR struggles a bit more than MASTER here, but since both show very erratic behavour, I can'r really faviour one above the other. Besides this last problematic part, PR and MASTER match closely.

It's probably worthwhile to look more closely into why _MOD4_GRP_ is so problematic, and try to figure out whether the problem is with flow or with the test itself. Currently, it's not well suited as a test (due to the problems, it's also quite computationally expensive).

GitPaean:
jenkins build this please

GitPaean (changes requested):
Sorry for the delay.  I left some comments regarding how to organize the code.  Let me know if you need some help or want some discussion. 

I do realize this PR is a incremental change that fixes things without breaking other things, while still trying to make the code usage more systematical or reduce the number of the functions whenever possible. 

Let me know.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1396,6 +1395,75 @@ namespace Opm
...
+                                                                                   summaryState, controls.injector_type,
+                                                                                   efficiencyFactor, deferred_logger);
+        if (target.has_value()) {
+            return target.value() == 0.0;
+        } else {
+            return false;
+        }
+        } else {
+            // Check producer under group control
+            const double scale = this->getGroupProductionTargetRate(group, well_state,
```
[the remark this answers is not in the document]
steink: Thanks, will change to `target_scale` (since the output is a scaling factor).

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1396,6 +1395,75 @@ namespace Opm
...
+        const auto& ws = well_state.well(this->index_of_well_);
+        const bool isGroupControlled = (this->isInjector() && ws.injection_cmode == Well::InjectorCMode::GRUP) ||
+                                       (this->isProducer() && ws.production_cmode == Well::ProducerCMode::GRUP);
+        if (!isGroupControlled) {
+            // well is not under group control, check "individual" version
+            const auto& summaryState = simulator.vanguard().summaryState();
+            return this->wellUnderZeroRateTargetIndividual(summaryState, well_state);
+        } else {
+            const auto& group_state = simulator.problem().wellModel().groupState();
+            return wellUnderZeroRateTargetGroup(simulator, well_state, group_state, deferred_logger);
```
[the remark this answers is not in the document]
steink: This was was my first intention, but was not able to do it without a lot of other code-changes. If you see some smarter way to move it there, I'll be happy to.

opm/simulators/wells/StandardWell_impl.hpp
steink: Thanks, would be happy to discuss stricter well-tolerances overall.  Especially for standard wells I think this is a good idea!

opm/simulators/wells/MultisegmentWell_impl.hpp
steink: > I think it is not a problem when it is with the correct sign, and it might be a problem when the sign of the WQTotal switches, which is easy for the wells without a zero rate target.

I beleive `WQTotal` is not allowed to switch sign in the Newton update, so it either has the right sign or is zero.

opm/simulators/wells/WellInterface_impl.hpp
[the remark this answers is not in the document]
GitPaean: Then it is the function name `getGroupProductionTargetRate` needs to be updated. You can keep the `scale` variable name here.

steink:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean (approved):

atgeirr (approved):
Looks good and can be merged assuming the test failures have been checked.

For later: Will look at improving some interfaces to possibly reduce the need to pass `const Simulator&` around, and maybe avoid using logging and exceptions in the TargetCalculator.

GitPaean:
The testing failures need to be double checked.

GitPaean:
Not remembering what happened with the regression test before the refactoring, will create a PR for testing purpose.

steink:
I've gone through the tests and think it looks OK. For the majority of the failures

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/steink/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/steink/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">



-- | -- | --
NETWORK-01_STANDARD | OK |  
NETWORK-01-REROUTE | OK |  
NETWORK-01-REROUTE_STD | OK |  
UDQ_WCONPROD | check | OK, different time-stepping
UDQ_ACTIONX | check | OK, different time-stepping
PYACTION_WCONPROD | check |  
9_3A_GINJ_REIN-G_MSW | tstep | OK, different time-stepping
9_3B_GINJ_GAS_EXPORT_MSW | tstep | OK, different time-stepping
9_3C_GINJ_GAS_GCONSUMP_MSW | check | OK, different time-stepping
9_3E_GAS_MIN_EXPORT_STW | OK |  
9_3D_GINJ_GAS_MAX_EXPORT_MSW | check | OK, different time-stepping
9_4A_WINJ_MAXWRATES_MAXBHP_GCONPROD_1L_STW | tstep | Quite different   results/time-stepping, time-steps much too long to catch dynamics accurately
9_4A_WINJ_MAXWRATES_MAXBHP_GCONPROD_1L_MSW | tstep | Quite different   results/time-stepping, time-steps much too long to catch dynamics accurately
9_4C_WINJ_GINJ_VREP-W_REIN-G_STW | check | OK, different time-stepping
9_4B_WINJ_VREP-W_MSW | tstep | OK, different time-stepping
9_4D_WINJ_GINJ_GAS_EXPORT_STW | tstep | OK, different time-stepping
9_4C_WINJ_GINJ_VREP-W_REIN-G_MSW | check | Different time-stepping, mostly   close match, except master violates group oil rate and pr violates group gas   rate
9_4D_WINJ_GINJ_GAS_EXPORT_MSW | check | Different time-stepping, mostly   close match, except master violates group oil rate
MOD4_UDQ_ACTIONX | tstep | OK,  slightly different time-stepping
ACTIONX_GCONPROD | rsm? | Not sure why this triggered   failure, results are almost identical
MOD4_GRP | tstep | OK, different   time-stepping,problematic case, but pr less problematic
MOD4_GRP_GEFAC | tstep | OK, different time-stepping
KRNUM-03X | check | OK, different time-stepping   (very long time-steps vs dynamics)
KRNUM-03Y | check | OK, different time-stepping   (very long time-steps vs dynamics)



</body>

</html>
[NETWORK-01](https://ci.opm-project.org/job/opm-simulators-PR-builder/6129/testReport/(root)/mpi/compareECLFiles_flow_NETWORK_01/)	OK	
NETWORK-01_STANDARD	OK	
NETWORK-01-REROUTE	OK	
NETWORK-01-REROUTE_STD	OK	
UDQ_WCONPROD	check	OK, different time-stepping
UDQ_ACTIONX	check	OK, different time-stepping
PYACTION_WCONPROD	check	
9_3A_GINJ_REIN-G_MSW	tstep	OK, different time-stepping
9_3B_GINJ_GAS_EXPORT_MSW	tstep	OK, different time-stepping
9_3C_GINJ_GAS_GCONSUMP_MSW	check	OK, different time-stepping
9_3E_GAS_MIN_EXPORT_STW	OK	
9_3D_GINJ_GAS_MAX_EXPORT_MSW	check	OK, different time-stepping
9_4A_WINJ_MAXWRATES_MAXBHP_GCONPROD_1L_STW	tstep	Quite different results/time-stepping, time-steps much too long to catch dynamics accurately
9_4A_WINJ_MAXWRATES_MAXBHP_GCONPROD_1L_MSW	tstep	Quite different results/time-stepping, time-steps much too long to catch dynamics accurately
9_4C_WINJ_GINJ_VREP-W_REIN-G_STW	check	OK, different time-stepping
9_4B_WINJ_VREP-W_MSW	tstep	OK, different time-stepping
9_4D_WINJ_GINJ_GAS_EXPORT_STW	tstep	OK, different time-stepping
9_4C_WINJ_GINJ_VREP-W_REIN-G_MSW	check	Different time-stepping, mostly close match, except master violates group oil rate and pr violates group gas rate
9_4D_WINJ_GINJ_GAS_EXPORT_MSW	check	Different time-stepping, mostly close match, except master violates group oil rate
MOD4_UDQ_ACTIONX	tstep	OK,  slightly different time-stepping
ACTIONX_GCONPROD	rsm?	Not sure why this triggered failure, results are almost identical
MOD4_GRP	tstep	OK, different time-stepping,problematic case, but pr less problematic
MOD4_GRP_GEFAC	tstep	OK, different time-stepping
KRNUM-03X	check	OK, different time-stepping (very long time-steps vs dynamics)
KRNUM-03Y	check	OK, different time-stepping (very long time-steps vs dynamics)

steink:
I've gone through the failures and think they look OK. For most of them it's just a result of different time-stepping/different number of time-steps which in many cases leads to quite different results since time-steps are so large relative to the dynamics. 
Did not check `PYACTION_WCONPROD`.



































































































<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/steink/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/steink/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">


  |   |  
-- | -- | --
NETWORK-01 | OK |  
NETWORK-01_STANDARD | OK |  
NETWORK-01-REROUTE | OK |  
NETWORK-01-REROUTE_STD | OK |  
UDQ_WCONPROD | check | OK, different time-stepping
UDQ_ACTIONX | check | OK, different time-stepping
PYACTION_WCONPROD | check |  
9_3A_GINJ_REIN-G_MSW | tstep | OK, different time-stepping
9_3B_GINJ_GAS_EXPORT_MSW | tstep | OK, different time-stepping
9_3C_GINJ_GAS_GCONSUMP_MSW | check | OK, different time-stepping
9_3E_GAS_MIN_EXPORT_STW | OK |  
9_3D_GINJ_GAS_MAX_EXPORT_MSW | check | OK, different time-stepping
9_4A_WINJ_MAXWRATES_MAXBHP_GCONPROD_1L_STW | tstep | Quite different   results/time-stepping, time-steps much too long to catch dynamics accurately
9_4A_WINJ_MAXWRATES_MAXBHP_GCONPROD_1L_MSW | tstep | Quite different   results/time-stepping, time-steps much too long to catch dynamics accurately
9_4C_WINJ_GINJ_VREP-W_REIN-G_STW | check | OK, different time-stepping
9_4B_WINJ_VREP-W_MSW | tstep | OK, different time-stepping
9_4D_WINJ_GINJ_GAS_EXPORT_STW | tstep | OK, different time-stepping
9_4C_WINJ_GINJ_VREP-W_REIN-G_MSW | check | Different time-stepping, mostly   close match, except master violates group oil rate and pr violates group gas   rate
9_4D_WINJ_GINJ_GAS_EXPORT_MSW | check | Different time-stepping, mostly   close match, except master violates group oil rate
MOD4_UDQ_ACTIONX | tstep | OK,  slightly different time-stepping
ACTIONX_GCONPROD | rsm? | Not sure why this triggered   failure, results are almost identical
MOD4_GRP | tstep | OK, different   time-stepping,problematic case, but pr less problematic
MOD4_GRP_GEFAC | tstep | OK, different time-stepping
KRNUM-03X | check | OK, different time-stepping   (very long time-steps vs dynamics)
KRNUM-03Y | check | OK, different time-stepping   (very long time-steps vs dynamics)



</body>

GitPaean:
The following is some notes for information. 


The thing is a little worrying is that, when we check the group target for a well, the group target might not be ready. 

I looked at the case KRNUM-03X, 

the SCHUEDULE setup for the well `F-1H` is as follows, with the current code, at the beginning, it will think F-1H is under zero injection group target, probably because the production rate is not ready yet so VREP get 0. 

```
WELSPECS
 'F-1H'  'S1'   3    3      1*   GAS     1*   1*   SHUT 1* 1* 1* /
/

GCONINJE
 'S1'   'GAS'	 'VREP'  3*	 1.020    'NO'  5* /
/

WCONINJE
-- Well_name    Type    Status  Ctrl    SRate1  Rrate   BHP     THP     VFP
  'F-1H'        GAS   OPEN    GRUP    8.0E6    1*      300.0	1*	1*     /
/
```

The VREP group control with above similar setup should be causes for many regression failures.  Because ideally, for many cases, there should be no wells under zero rate target at all and should no be affected.

GitPaean:
> Yes, but don't you think it's better in this case to treat it properly as a zero rate well than trying to solve it with an individual zero rate (as in master) which even might give singular system?

It took some thinking to understand your point. Basically, you meant that without the checking, we will solve it with zero rate target in the master branch.  I think your point is valid and I agree. 

My point was that `stopppedOrZeroRateTarget()` can be improved to reflect better. Or we should have different ways in handling the VREP injections, but that should be outside of the scope of this PR.

steink:
>It took some thinking to understand your point. Basically, you meant that without the checking, we will solve it with zero rate target in the master branch. I think your point is valid and I agree.

Sorry, was trying to finnish the sentence before a meeting ;)

>My point was that stopppedOrZeroRateTarget() can be improved to reflect better. Or we should have different ways in handling the VREP injections, but that should be outside of the scope of this PR.

I see you point, I agree.

GitPaean:
For the test failure, `ACTIONX_GCONPROD`, the message is not very clear. 

```
Loading rft file /var/lib/jenkins/workspace/opm-simulators-PR-builder/deps/opm-tests/actionx/opm-simulation-reference/flow/ACTIONX_GCONPROD.RFT  .... done
Loading rft file /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/build-opm-simulators/tests/results/flow+actionx_gconprod/ACTIONX_GCONPROD.RFT  .... done

Well: OP01 date: 2017/12/31
Comparing: TIME ...  done.
Comparing: DATE ...  done.
Comparing: WELLETC ...  done.
Comparing: CONIPOS ...  done.
Comparing: CONJPOS ...  done.
Comparing: CONKPOS ...  done.
Comparing: HOSTGRID ...  done.
Comparing: DEPTH ...  done.
Comparing: PRESSURE ...  done.
Comparing: SWAT ...  done.
Comparing: SGAS ...  done.

Program threw an exception: [/var/lib/jenkins/workspace/opm-simulators-PR-builder/deps/opm-common/test_util/compareECL.cpp:265] 1 errors encountered in comparisons.
```

GitPaean:
With some discussion and help from @bska , the regression failure for `ACTIONX_GCONPROD` is related to RUNSUM keyword.  

Since the simulation results look generally fine and for this failure, we can update reference.  But we need to wait for the release processing to unfreeze the reference result update.

GitPaean:
@steink , can you update the PR to fix the merging conflicts, then we can check how to adjust and update the reference so that we can get the PR in.

steink:
> @steink , can you update the PR to fix the merging conflicts, then we can check how to adjust and update the reference so that we can get the PR in.
Thought I was able to fix conflicts here, but rebase attempt ended up in big mess. Maybe you can have a look @GitPaean?

akva2:
please grab rebased branch from https://github.com/akva2/opm-simulators/tree/check_zero_target_from_group. note I squashed the previously botched rebasing, makes no sense to keep that history.

GitPaean:
jenkins build this please

steink:
> please grab rebased branch from https://github.com/akva2/opm-simulators/tree/check_zero_target_from_group. note I squashed the previously botched rebasing, makes no sense to keep that history.

Thanks @akva2 !

GitPaean:
jenkins build this please

GitPaean:
Some how the regression related to `PYACTION_WCONPROD` does not look severe anymore. 

https://ci.opm-project.org/job/opm-simulators-PR-builder/6269/testReport/junit/(root)/mpi/compareECLFiles_flow_PYACTION_WCONPROD/

Then most likely, we do not need to adjust to make the regression test to work.

GitPaean:
Will update_reference for investigation purpose.  If things look okay, will get prepared to get this PR in.

GitPaean:

GitPaean:
> Some how the regression related to `PYACTION_WCONPROD` does not look severe anymore.
> 
> https://ci.opm-project.org/job/opm-simulators-PR-builder/6269/testReport/junit/(root)/mpi/compareECLFiles_flow_PYACTION_WCONPROD/
> 
> Then most likely, we do not need to adjust to make the regression test to work.

I took this back. It still remains a problem with more investigation.

GitPaean:

bska:
jenkins build this opm-tests=1168 please

bska:
The new reference solutions have been installed on the CI system.  Merging this now to activate the new logic.  Thanks a lot for seeing this through!

PR #4950, opened 2023-10-26
Get primary variables and fluid state from Python

hakonhagland:
Added methods to the Python module `opm.simulators.BlackOilSimulator` to access primary variables and fluid state variables.

PR #5235, opened 2024-02-29
fixed: don't link python module to flow_libblackoil

akva2:
this leads to ODR violations which causes all sorts of havoc. this was done to reuse a trivial method, duplicate it in the python source instead.

This is the final fallout of the troubles observed in https://github.com/OPM/opm-simulators/pull/4950
After a long time we were finally able to track the crash down to building using ninja rather than make.
Using object libraries the order of the objects differ in the linker command; with ninja the object libraries are always placed first while with make they are not. This makes a difference as the linker choose the duplicated symbols based on which source come first, and by pure luck the order with make had the linker choose the 'correct' copy of the symbol, while the order with ninja had it choose the wrong symbol.

### DIFF
--- a/python/simulators/CMakeLists.txt
+++ b/python/simulators/CMakeLists.txt
@@ -9,16 +9,13 @@
 #  https://pybind11.readthedocs.io/en/stable/compiling.html
 #
 pybind11_add_module(simulators ${PYBIND11_SYSTEM}
+  $<TARGET_OBJECTS:moduleVersion>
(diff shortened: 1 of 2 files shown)


hakonhagland:
> This is the final fallout of the troubles observed in #4950 .. After a long time we were finally able to track the crash down

@akva2 Veldig bra! Du burde skrive en blog post om dette så andre kan lære av det :smile:

PR #5236, opened 2024-02-29
Avoid deleting derivative i.e. getting "wrong" matrix in case of zero…

hnil:
* avoid getting zero in matrix in zero pressure/zero flux case
* may make it easier to have lower/upper limits on pressures

This should be the correct, but it will may change many tests so it should be investigated carefully. This is same change as in model (OPM/opm-models#877) but will only affect legacy assembly.


ebos/eclfluxmodule.hh
```diff
@@ -278,10 +278,6 @@ public:
                                         J,
                                         distZ*g,
                                         thpres);
-            if (pressureDifferences[phaseIdx] == 0) {
-                volumeFlux[phaseIdx] = 0.0;
-                continue;
-            }
```
blattms, line 284: I don't really understand why this changes anything. Souldn't `volumeFlux[phaseindex]` still be zero as there is a multiplication with zero below. What am I missing?

GitPaean:
This PR should be rebased due to the removal of the `eclfluxmodule`.  I will create a new PR unless @hnil want to do it himself.

GitPaean (approved):
The change looks fine. Thanks for the contribution.

GitPaean:
This test needs manual update the test, 

[mpi.python_fluidstate_variables](https://ci.opm-project.org/job/opm-models-PR-builder/585/testReport/junit/(root)/mpi/python_fluidstate_variables/)

PR #5237, opened 2024-02-29
AdaptiveTimeStepping: use {fmt} to format messages

akva2:
std::to_string doesn't use generic format and thus small numbers are truncated to 0.0000000


akva2:
jenkins build this please

PR #5238, opened 2024-03-01
fixed: if we write the size as a size_t we better read it as a size_t

akva2:
fixes regression after 7dcc411397baa585dca1fb6b804d074d9c3203d3


akva2:
jenkins build this please

bska (approved):

PR #5239, opened 2024-03-01
changed: rename FlowMain.cpp to FlowUtils.cpp

akva2:
and add a dedicated header.

this way we don't need to pull in FlowMain.hpp for the prototypes, avoiding pulling in the entire simulator machinery just to build some simple utility functions.


akva2:
jenkins build this please

PR #5233, opened 2024-02-28
Add Support for Calculating Block Level Depth Corrected Pressures

bska:
This PR calculates the block-level depth-corrected phase pressures `BPPO`, `BPPG`, and `BPPW`.  We use the simulation configuration's `DatumDepth` feature to extract the reference depths for the `FIPNUM` region set per cell and accumulate average phase densities per PVT regions using the `RegionPhasePoreVolAverage` utility.  To that end, add a new `optional<>` data member of this type to the generic output module and initialise this if the input model requests such summary output.

PR #5240, opened 2024-03-01
Add Support for Calculating Block Level Depth Corrected Pressures

bska:
This PR calculates the block-level depth-corrected phase pressures `BPPO`, `BPPG`, and `BPPW`.  We use the simulation configuration's `DatumDepth` feature to extract the reference depths for the `FIPNUM` region set per cell and accumulate average phase densities per PVT regions using the `RegionPhasePoreVolAverage` utility.  To that end, add a new `optional<>` data member of this type to the generic output module and initialise this if the input model requests such summary output.


bska:
Replacement for #5233 which was lost in a rebase operation.

bska:
jenkins build this please

bska:
Merging this now since this was already approved in a different guise in #5233.

PR #5241, opened 2024-03-04
Permit DATUMR and DATUMRX Keywords in Input File

bska:
Following PRs

  * OPM/opm-common#3958
  * OPM/opm-simulators#5240

these keywords are nominally supported, at least for the block level summary keywords (`BPPx`).


bska:
PR approved and build check is green.  Merging this into the master branch

PR #5242, opened 2024-03-04
Add docstrings for Python bindings

hakonhagland:
This PR builds on #4950 which should be merged first.

Adds docstrings for the Python bindings. Currently, the docstrings can be used to get context help in editors or in IPython. However, the plan is to also use these docstrings to generate sphinx documentation for the Python bindings. This is the reason why we are generating the docstrings at build time using a JSON input file. Since the JSON input file can also be read by a sphinx extension to generate sphinx documentation without having to build the pybind11 opm-simulators module first (which would involve having to build the whole opm-simulators). The plan is to build the sphinx documentation and deploy it to GitHub pages using a GitHub Action workflow. To avoid large resource usage in the GitHub action workflow, we therefore  try to avoid having to build the whole opm-simulators (and potentially also opm-common since the opm-simulators python bindings commonly also use the python bindings from opm-common) just to create the documentation.

PR #5243, opened 2024-03-05
Add sphinx documentation for Python bindings and GitHub action workflow to push to GitHub pages

hakonhagland:
Builds on #5242 which should be merged first.

Added sphinx documentation for the `opm.simulators.BlackOilSimulator`  Python module. Also added GitHub action workflow to deploy the documentation to GitHub Pages.

For this to work, someone with admin rights for the opm-simulators GitHub repository has to go to https://github.com/OPM/opm-simulators and then to the menu `Settings->Pages` and set the branch name for deployment to `gh-pages`, see screenshot below:

Pasted image 20240303133458

See also: https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site


blattms:
Under what url would the documentation appear? Where would something from opm-common appear?

Will this always be the documentation for current master? Would we have release documentation on the side?

hakonhagland:
>Under what url would the documentation appear? 

@blattms I believe it will appear at something like https://opm.github.io/opm-simulators/

> Where would something from opm-common appear?

 I think: https://opm.github.io/opm-common/

python/sphinx_docs/docs/conf.py
```diff
@@ -0,0 +1,44 @@
+# Configuration file for the Sphinx documentation builder.
+
+project = "opm.simulators"
+copyright = "2024 Equinor ASA"
+author = "Håkon Hægland"
+release = "0.1"
```
akva2, line 6: shouldn't this be generated/derived from versioning information? seems a bit counter-productive to have a fixed version number completely separate from everything else.

hakonhagland:
Rebased

hakonhagland:
jenkins build this please

python/sphinx_docs/docs/conf.py
hakonhagland: Yes maybe. Are you referring to the opm-simulators version like "2024.04" ?

akva2: yes, which has a single source; dune.module in top of repository.

hakonhagland:
jenkins build this please

python/sphinx_docs/docs/conf.py
hakonhagland: Ok great. I have updated the script to read the version from dune.module

python/sphinx_docs/pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[tool.poetry]
+name = "opm-simulators-docs"
+version = "0.1.0"
```
akva2, line 3: not sure if this needs to be in sync?

hakonhagland: I think not? Since it is not going to be published. The python project is only used to build the sphinx documentation.

akva2: yeah i kinda guessed, just wasn't sure if something would get their panties in a bundle or not.

akva2 (approved):
I don't have settings access so I can't active it but this looks good to me at least.

hakonhagland:
> Will this always be the documentation for current master? Would we have release documentation on the side?

@blattms Good point, I did not think about versions. I think we can get versioning with GitHub pages,  see for example: https://github.com/devanshshukla99/sphinx-versioned-docs and https://www.codingwiththomas.com/blog/my-sphinx-best-practice-for-a-multiversion-documentation-in-different-languages. But it seems like it is not very commonly used? Maybe we should try [Read the Docs](https://docs.readthedocs.io/en/stable/) instead? Since they appear to have a more well established workflow for versioning? @vkip What do you think?

hakonhagland:
> Will this always be the documentation for current master? Would we have release documentation on the side?

@blattms I played a little bit with https://github.com/devanshshukla99/sphinx-versioned-docs  and I think we can use this to get versioned docs. We have to tag the commits we want to include, then the sphinx build will collect the versions from those commits (all included in the sphinx `_build` directory. When we upload this directory to GitHub pages there will be a version selector at the front page similar to the one used at readthedocs. The user can then select the version of the documentation they want to view from that selector.

vkip:
> ...  @vkip What do you think?
No real strong opinion here. As you mention RTD versioning is well established, but as long as it works this is fine by me.

hakonhagland:
@vkip, @akva2, and @blattms I added a new commit to address the versioning, see: https://github.com/OPM/opm-simulators/pull/5243/commits/69d3d6caf1fde4c02f6634ac0cc2de966bebc130

hakonhagland:
jenkins build this please

blattms:
Dumb questions from me:
- Is this action executed for each merge?
- Is there any version information on the generated pages (e.g. commit or date)?

I know it is surprising if you see how I behave in this project :sweat_smile:, but I do not have any admin power here. Seems like only @alfbr, @atgeirr, or @totto82 can change settings.

hakonhagland:
> Is this action executed for each merge?

@blattms It is executed for each merge/push to the master branch but only if anything inside the `python` folder was changed.

> Is there any version information on the generated pages (e.g. commit or date)?

We can choose to tag certain commits (e.g. releases) and have those included in the version selector in the lower left corner of the page. Here is an example from another GitHub page:

image

This page has included versions for git tags "v0.1", "v1.0", "v1.1", and "v1.2" and for the current head of the "main" branch. The currently "active" documentation among those versions is the "main" as indicated with green color.

Let me know if we need more information than this tag information.

totto82:
Can we merge this? Or will this interfere with the release process?

hakonhagland:
jenkins build this please

hakonhagland:
Rebased this PR on master to include the changes in docstrings introduced in #5325.

lisajulia:
jenkins build this opm-common=4018 please

lisajulia:
@atgeirr / @akva2: Can you have a look here and possibly merge? Thanks :)

akva2 (approved):
Can't say I have really reviewed this in detail, but nothing jumps out, so yolo.

lisajulia:
> Builds on #5242 which should be merged first.
> 
> Added sphinx documentation for the `opm.simulators.BlackOilSimulator` Python module. Also added GitHub action workflow to deploy the documentation to GitHub Pages.
> 
> For this to work, someone with admin rights for the opm-simulators GitHub repository has to go to https://github.com/OPM/opm-simulators and then to the menu `Settings->Pages` and set the branch name for deployment to `gh-pages`, see screenshot below:
> 
> Pasted image 20240303133458
> 
> See also: https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site

@akva2 or @blattms  or @atgeirr or @alfbr can someone set up the deployment to github pages (as described by Håkon)? Thanks!

PR #5244, opened 2024-03-06
Defer SummaryConfig Initialisation

bska:
In the current approach, the full list of summary vectors, especially those that are defined at the region level, is not known until we've processed all UDQ and/or ACTIONX definitions.  As a quick solution to this, switch to using the `SummaryConfig` object that's defined in the `EclipseIO` container instead of the object that gets constructed when reading the input files.

It is likely that we will have to rethink and refactor this construction process later.

### DIFF
--- a/opm/simulators/flow/DamarisWriter.hpp
+++ b/opm/simulators/flow/DamarisWriter.hpp
@@ -36,6 +36,9 @@
 
 #include <opm/common/OpmLog/OpmLog.hpp>
 
+#include <mpi.h>
+#include <opm/simulators/utils/MPISerializer.hpp>
+
 #include <opm/simulators/flow/countGlobalCells.hpp>
 #include <opm/simulators/flow/DamarisProperties.hpp>
 #include <opm/simulators/flow/EclGenericWriter.hpp>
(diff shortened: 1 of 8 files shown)


bska:
I am creating this PR in draft mode because it depends on the new member function
```
EclipseIO::finalSummaryConfig()
```
which is introduced in the upstream PR OPM/opm-common#3968.  I will keep the PR in draft state until such time as it is ready for review.

bska:
The upstream PR OPM/opm-common#3968 was merged into the master branch.  I'm marking this PR as "ready for review".

bska:
jenkins build this please

bska:
jenkins build this please

opm/simulators/flow/DamarisWriter.hpp
```diff
@@ -149,25 +152,56 @@ class DamarisWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::G
...
+        {
+            const auto& gridView = this->simulator_.gridView();
+            const auto& interior_elements = elements(gridView, Dune::Partitions::interior);
+
+            this->numElements_ = std::distance(interior_elements.begin(), interior_elements.end());
+        }
 
-        this->elements_rank_offsets_.resize(nranks_) ;
-        this->damarisOutputModule_ = std::make_unique<OutputBlackOilModule<TypeTag>>(simulator, this->collectOnIORank_);
+        if (this->nranks_ > 1) {
```
akva2, line 187: consider a helper in ParallelSerialization ala eclStateBroadcast.

akva2 (commented):
Looks fine but we can avoid some duplication.

opm/simulators/flow/DamarisWriter.hpp
bska: > consider helper in ParallelSerialization (eclBroadCast).

Good idea.  I've pushed an update to that effect.

bska:
jenkins build this please

bska:
jenkins build this please

blattms:
All comments addressed. Merging.

PR #5245, opened 2024-03-06
Drop ebos: Mega PR

akva2:
As requested by @GitPaean here is all the drop PR's merged into one. I cannot give any guarantee there are no issues introduced when moving this fast but it should at the very least build at each commit.


GitPaean:
jenkins build this please

akva2:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
```c++
[opm/simulators/flow/FlowUtils.cpp](https://github.com/OPM/opm-simulators/pull/5245/files#diff-21cca80f6ddf52773d9e8fc647dada02c9ec1a15dd413dc3c46f77b0b1d72529)
Viewed
@@ -28,67 +28,67 @@
#include <opm/simulators/flow/ConvergenceOutputConfiguration.hpp>
#include <opm/simulators/timestepping/SimulatorReport.hpp>
#include <opm/simulators/utils/ParallelFileMerger.hpp>
#if HAVE_MPI
#include <ebos/eclgenericvanguard.hh>
#endif
```

this `ebos` needs cleaning up.

GitPaean:
And an extra head file needs to be included in the same file, 
```diff
diff --git a/opm/simulators/flow/FlowUtils.cpp b/opm/simulators/flow/FlowUtils.cpp
index ef5b99107..1d9313896 100644
--- a/opm/simulators/flow/FlowUtils.cpp
+++ b/opm/simulators/flow/FlowUtils.cpp
@@ -27,13 +27,10 @@
 #include <opm/common/utility/String.hpp>
 
 #include <opm/simulators/flow/ConvergenceOutputConfiguration.hpp>
+#include <opm/simulators/flow/FlowGenericVanguard.hpp>
 #include <opm/simulators/timestepping/SimulatorReport.hpp>
 #include <opm/simulators/utils/ParallelFileMerger.hpp>
 ```

File: opm/simulators/flow/FlowGenericProblem_impl.hpp [hunk 1]
```diff
@@ -131,24 +129,14 @@ helpPreamble(int,
 
 template<class GridView, class FluidSystem, class Scalar>
 std::string
-EclGenericProblem<GridView,FluidSystem,Scalar>::
+FlowGenericProblem<GridView,FluidSystem,Scalar>::
 briefDescription()
 {
-    if (briefDescription_.empty())
```
GitPaean, line 137: why this? Any possible implications?

opm/simulators/flow/FlowMain.hpp
```diff
@@ -155,11 +155,6 @@ namespace Opm {
             EWOMS_HIDE_PARAM(TypeTag, NewtonTargetIterations);
             EWOMS_HIDE_PARAM(TypeTag, NewtonVerbose);
             EWOMS_HIDE_PARAM(TypeTag, NewtonWriteConvergence);
-            EWOMS_HIDE_PARAM(TypeTag, EclNewtonSumTolerance);
```
GitPaean, line 158: question, why removing them?

opm/simulators/flow/GenericCpGridVanguard.hpp
```diff
@@ -22,15 +22,15 @@
 */
 /*!
  * \file
- * \copydoc Opm::EclCpGridVanguard
+ * \copydoc Opm::CpGridVanguard
```
GitPaean, line 25: Should this reflect the class name?

opm/simulators/flow/GenericThresholdPressure.hpp
```diff
@@ -23,15 +23,14 @@
 /*!
  * \file
  *
- * \copydoc Opm::EclThresholdPressure
+ * \copydoc Opm::ThresholdPressure
```
GitPaean, line 26: Same here. But originally Generic is missing also. I do not have a strong opinion.

opm/simulators/flow/equil/InitStateEquil.cpp
```diff
@@ -22,15 +22,14 @@
 */
 
 #include <config.h>
-#include <ebos/equil/initstateequil.hh>
-#include <ebos/equil/initstateequil_impl.hh>
+#include <opm/simulators/flow/equil/InitStateEquil_impl.hpp>
```
GitPaean, line 25: _impl.hpp files includes the .hpp file. Since it happens more than one place, it looks the including of the original .hh or .hpp file is intentional. I do not have an opinion about this, maybe @bska can comment on this.

tests/test_RestartSerialization.cpp
```diff
@@ -440,14 +442,17 @@ namespace {
...
+        using TT = Properties::TTag::TestRestartTypeTag;
         const char* argv[] = {
             "test_RestartSerialization",
             "--ecl-deck-file-name=GLIFT1.DATA"
         };
-        Opm::AdaptiveTimeStepping<TT>::registerParameters();
-        Opm::setupParameters_<TT>(2, argv, /*registerParams=*/true);
-        Opm::EclGenericVanguard::setCommunication(std::make_unique<Opm::Parallel::Communication>());
+        AdaptiveTimeStepping<TT>::registerParameters();
+        BlackoilModelParameters<TT>::registerParameters();
```
GitPaean, line 452: why the following two extra lines?
```c++
        BlackoilModelParameters<TT>::registerParameters();
        EWOMS_REGISTER_PARAM(TT, bool, EnableTerminalOutput, "Do *NOT* use!");
```

opm/simulators/flow/FlowMain.hpp
GitPaean: Is it because in the TestTypeTag.hh, they are defined?

GitPaean (changes requested):
Very roughly walking through. No big issues spotted. Please fix the building so we can test.

tests/test_RestartSerialization.cpp
GitPaean: Is it also because the new "TestTypeTag.hh"?

opm/simulators/flow/FIBlackoilModel.hpp
```diff
@@ -33,6 +33,8 @@
 
 #include <opm/common/ErrorMacros.hpp>
 
+#include <opm/simulators/utils/DeferredLoggingErrorHelpers.hpp>
```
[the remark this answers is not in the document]
akva2: Because its symbols are used and header reorganizations disclosed the lack of the include in this header.

opm/simulators/flow/FlowGenericProblem_impl.hpp
akva2: This was the banner used by the 'ebos' binary. As there is no ebos binary...

opm/simulators/flow/FlowMain.hpp
akva2: Because the were defined in the eclnewtonmethod.hh which we are removing. These settings are not used anywhere in flow, they were only ever used by ebos.

opm/simulators/flow/GenericCpGridVanguard.hpp
akva2: It's used for copying the doxy from the Opm::CpGridVanguard class. The generic class always used the same doxy as the base class, ie. nothing changes.

opm/simulators/flow/GenericThresholdPressure.hpp
akva2: Same as above.

opm/simulators/flow/equil/InitStateEquil.cpp
akva2: It's a noop as such.The header guard is defined at this point. It's an incomplete attempt at what I do properly in https://github.com/OPM/opm-simulators/pull/5227 (though not for this class).

tests/test_RestartSerialization.cpp
akva2: They used to be done in the EbosProblem class but since that is no longer a thing, I have to do them here instead.

akva2:
jenkins build this please

GitPaean:
Thanks for the update. I am getting it in now.  It probably will make other PRs need to rebase.

opm/simulators/flow/equil/InitStateEquil.cpp
akva2: Uhm, this is a different pattern. You have a .hpp, a _impl.hpp and a .cpp file. The cpp file instances for CpGrid. So in the regular simulators we just include the .hpp file and the instance is in the library. For simulators using other grids, we instead include the _impl.hpp file and thus the instance for those grid types are compiled inline in the simulator object.

PR #5246, opened 2024-03-07
Switch to Using Declared Maximum Region IDs From Common

bska:
The `declaredMaxRegionID()` function is now available in OPM-Common (PR OPM/opm-common#3969).


bska:
I'm creating this PR in draft mode for two reasons

1. It depends on, and must be merged in concert with, the upstream PR OPM/opm-common#3969
2. It is currently part of a PR chain building up to enabling using region-level summary vectors in the defining expressions for field-level UDQs.

I will probably pull this PR out of the chain at a later time.

bska:
jenkins build this opm-common=3969 please

blattms:
Where is this "switch"? Could not find where the removed function was used.

bska:
> Where is this "switch"? Could not find where the removed function was used.

It's in the [call](https://github.com/OPM/opm-simulators/blob/1d758706df03dd896c404c03fdf01272c189dee6/opm/simulators/flow/GenericOutputBlackoilModule.cpp#L183) to the inter-region flow map constructor.

PR #5247, opened 2024-03-07
Add Procedure for Calculating Basic FIP Region Statistics

bska:
This PR implements the parallel version of
```
EclipseState::computeFipRegionStatistics()
```
which computes a `FIPRegionStatistics` object for the current run's fluid-in-place regions.  The object construction uses an MPI-aware reduction process to compute the maximum region IDs across all MPI ranks.

While here, also unconditionally form the statistics object as part of the `EclWriter`'s constructor to ensure that all ranks participate in the process.  The initial approach of constructing the object on first use is not robust in parallel.  We may however wish to compute these statistics only when needed.  If so, that will be the subject of follow-up work.

### DIFF
--- a/opm/simulators/flow/EclWriter.hpp
+++ b/opm/simulators/flow/EclWriter.hpp
@@ -170,6 +170,8 @@ public:
         }
 
         this->rank_ = this->simulator_.vanguard().grid().comm().rank();
+
+        this->simulator_.vanguard().eclState().computeFipRegionStatistics();
     }
 
     ~EclWriter()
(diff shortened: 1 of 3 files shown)


bska:
I am creating this PR in draft mode because it is part of a PR chain building up to enabling using region-level summary vectors in the defining expressions of field-level UDQs.  It also depends on, and must be merged in concert with the upstream PR OPM/opm-common#3971.  I will keep the PR up-to-date and in draft state until such time as it is ready for reivew.

bska:
jenkins build this opm-common=3971 please

akva2 (approved):

bska:

PR #5252, opened 2024-03-12
fixed: we need to call loadRestartData on all processes

akva2:
we need to read in group data even if there are no local wells


akva2:
jenkins build this please

blattms (approved):
Great catch

opm/simulators/wells/BlackoilWellModelGeneric.cpp
bska, line 1: What's the cost of constructing `BlackoilWellModelRestart` objects?  Would it be worthwhile to do this construction once instead of once for each call to `loadRestart*()`?

akva2: basically zero cost. all it does is pass a reference.

bska (approved):
Yes, very good catch.  I'll merge into master.

PR #5255, opened 2024-03-13
Bug fix: remove dependencies for rocm & cuda on opencl to run block Jacobi ILU

razvnane:
This PR fixes a bug introduced when we removed the rocm dependency on opencl (#4883), which disabled the possibility to run the rocsparse and cusparse backends with the optimized version of block Jacobi ILU.

### DIFF
--- a/opm/simulators/flow/FlowBaseVanguard.hpp
+++ b/opm/simulators/flow/FlowBaseVanguard.hpp
@@ -89,12 +89,12 @@ struct EdgeWeightsMethod {
     using type = UndefinedProperty;
 };
 
-#if HAVE_OPENCL || HAVE_ROCSPARSE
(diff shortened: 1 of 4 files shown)


blattms:
@akva2 I think jenkins needs to be called differently to really test this, here? Or maybe we have to do this manually on our system?

akva2:
jenkins build this hip please

akva2:
that will test with cuda, no rocm and rocm, no cuda. no configuration with just opencl is available though, i can add one tmrw.

akva2:
jenkins build this hip serial opencl please

akva2:
there we go, tested with hip, cuda, only opencl and without any of them. all good. still cannot test runtime but nothing should have changed there.

PR #5258, opened 2024-03-19
Allow for Sparse Source-Location Subsets

bska:
If a well is nominally connected in inactive cells, i.e., if its connections have not been filtered down to the active connections only, then `allIndices` may not be a permutation of
```
0 .. allIndices.size()-1
```
Handle this case by sizing `storageIndex_` according to the maximum index value instead of `allIndices.size()`.

While here, also switch to constructing `PerforationData` objects in place since this is (slightly) faster than "construct"+"copy".

### DIFF
--- a/opm/simulators/wells/ParallelPAvgDynamicSourceData.cpp
+++ b/opm/simulators/wells/ParallelPAvgDynamicSourceData.cpp
@@ -148,7 +148,19 @@ void Opm::ParallelPAvgDynamicSourceData::defineCommunication()
 
     // 4) Build translation mapping from source term element indices to
     //    storage indices.
-    this->storageIndex_.resize(allIndices.size());
+    //
+    //    Note that if the source terms aren't all active--e.g., if a well
+    //    is connected in deactivated cells--then allIndices is not a
+    //    permutation of 0..allIndices.size()-1 and the maximum source
+    //    location may exceed size()-1.  Resize the storageIndex_ according
+    //    to the largest source location ID.
+    if (auto maxIxPos = std::max_element(allIndices.begin(), allIndices.end());
+        maxIxPos != allIndices.end())
+    {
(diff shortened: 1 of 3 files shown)


bska:
jenkins build this please

opm/simulators/wells/ParallelPAvgDynamicSourceData.cpp
```diff
@@ -148,7 +148,19 @@ void Opm::ParallelPAvgDynamicSourceData::defineCommunication()
...
     //    storage indices.
-    this->storageIndex_.resize(allIndices.size());
+    //
+    //    Note that if the source terms aren't all active--e.g., if a well
+    //    is connected in deactivated cells--then allIndices is not a
+    //    permutation of 0..allIndices.size()-1 and the maximum source
+    //    location may exceed size()-1.  Resize the storageIndex_ according
+    //    to the largest source location ID.
+    if (auto maxIxPos = std::max_element(allIndices.begin(), allIndices.end());
+        maxIxPos != allIndices.end())
```
akva2, line 158: maybe a bit wordy guard against an empty vector.

akva2 (approved):
Does as announced.

opm/simulators/wells/ParallelPAvgDynamicSourceData.cpp
bska: > maybe a bit wordy guard against an empty vector.

I'm not entirely sure I understand that comment.  Yes, this guards against an empty vector, but I also need the maximum value (`*maxIxPos`) in the `.resize()` call below.

akva2: yeah, i just would personally have structured it a bit differently with an if (!empty)  around the resize nest. no need to change anything.

bska:

PR #5259, opened 2024-03-21
[bugfix] Use reasonable timestep if  NEXTSTEP in ACTIONX is used.

blattms:
The step size might still need to be smaller than suggested to not simulate beyond the end of the current report step. To ensure this we now use AdaptiveTimeSimulationTimer::provideTimeStepEstimate which will limit it and also make sure that subsequent time steps will not get to small either.

### DIFF
--- a/opm/simulators/timestepping/AdaptiveTimeStepping.hpp
+++ b/opm/simulators/timestepping/AdaptiveTimeStepping.hpp
@@ -387,7 +387,8 @@ std::set<std::string> consistentlyFailingWells(const std::vector<StepReport>& sr
                 // get current delta t
                 auto oldValue = suggestedNextTimestep_;
                 if (tuningUpdater()) {
-                    substepTimer.setCurrentStepLength(suggestedNextTimestep_);
+                    // Use provideTimeStepEstimate to make we sure don't simulate longer than the report step is.
+                    substepTimer.provideTimeStepEstimate(suggestedNextTimestep_);
                     suggestedNextTimestep_ = oldValue;
                 }
                 const double dt = substepTimer.currentStepLength();


blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
I am inclined to remove method `AdaptiveSimulatorTimer::setCurrentStepLength` as this is not really save to be called ourside of that class. At least we should make it protected.

bska (approved):
Looks good to me.  Unless there's more work coming here, we should merge this as is.

> I am inclined to remove method `AdaptiveSimulatorTimer::setCurrentStepLength()` as this is not really save to be called outside of that class. At least we should make it protected.

Given that it's a recent invention (#5222), I think that's a good idea.

blattms:
> Given that it's a recent invention (#5222), I think that's a good idea.

Oopsy :flushed:

blattms:
I guess I have to research  a bit why the regression test [PINCH10_NOPINCH](https://ci.opm-project.org/job/opm-simulators-PR-builder/6061/testReport/junit/(root)/mpi/compareECLFiles_flow_PINCH10_NOPINCH/) fails with:
```
Keyword: INTEHEAD, origin Restart, sequence 7

Value index                 = 64
(first value, second value) = (14, 16)

Program threw an exception: [/var/lib/jenkins/post-builder/workspace/opm-simulators-PR-builder/deps/opm-common/test_util/EclRegressionTest.cpp:171] Non floating point values not identical 
```

blattms:
Strange. I am pretty sure I only changed the line with comment. The [last jenkins run](https://ci.opm-project.org/job/opm-simulators-PR-builder/6054/) before that change was successful

bska:
> ```
> Keyword: INTEHEAD, origin Restart, sequence 7
> 
> Value index                 = 64
> (first value, second value) = (14, 16)
> ```

So just for clarity/completeness: [`INTEHEAD[64]`](https://github.com/OPM/opm-common/blob/6102da99f244b7ae0a00e05e4205e9566645a5dd/opm/output/eclipse/VectorItems/intehead.hpp#L89) is the "day" of the report step's date.  I would not expect that value to change as a result of this PR.

blattms:
On my system at least the md5sum of the UNRST file (master and this branch) are the same. Should I look at another file?

blattms:
jenkins build this please

bska (approved):
Build check is green and I'll merge into master.

PR #4337, opened 2022-12-21
Avoid use of sstream in deferred logger header / macros

akva2:
Instead use {fmt} at callsites where available or simple string::operator+(..).

Waiting for #4292

PR #5261, opened 2024-03-22
[bugfix] Correctly form debug message to avoid exceptions in libfmt.

blattms:


blattms:
jenkins build this please

akva2 (approved):
woopsie

opm/simulators/wells/MultisegmentWellSegments.cpp
```diff
@@ -262,7 +262,7 @@ computeFluidProperties(const EvalWell& temperature,
                                 "rv {} and pressure {}. "
                                 "Continue as if no dissolution (rs = 0) and "
                                 "vaporization (rv = 0) for this connection.",
-                                d, well_.name(), rv, seg_pressure);
+                                d, well_.name(), rs, rv, seg_pressure);
```
bska, line 265: Sorry!  I should have caught that whilst reviewing #4337...

bska (approved):
Good catch!

opm/simulators/wells/MultisegmentWellSegments.cpp
blattms: On the bright side, this branch is not executed very often. :sweat_smile:

PR #5262, opened 2024-03-27
removing the including of FlowProblem in BlackoilWellModel

GitPaean:


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

blattms (approved):
Looks like a good and innocent cleanup to me,

akva2:
sorry, was on vacation. this is not entirely correct, we do not need flowproblem but we need some of the properties to be defined for happy clangd. will submit a follow-up PR.

GitPaean:
> sorry, was on vacation. this is not entirely correct, we do not need flowproblem but we need some of the properties to be defined for happy clangd. will submit a follow-up PR.

Can you elaborate so that we can discuss a solution?

akva2:
we need stuff which are , indirectly defined in FlowProblem.hpp, e.g. ```Properties::EquilGrid```,  ```Properties::GlobalEqVector``` is undefined. It's correct we don't need FlowProblem, but we need stuff that was pulled in through that header. Let me just create the PR.

PR #5263, opened 2024-04-01
[fix] Prevent dangling refernces in Logger for test_LogOutputHelper

blattms:
In the test methods of the test suite we construct a new logger that logs to a local stringstream. This logger will be stored in a static map and hence live longer that the test method and the stringstream.

Some methods use `Opm::Log::MessageType::Note`, others `Opm::Log::MessageType::Warning` for adding the logging backend . There is one map per `MessageType` and sometimes (e.g. previous method used Note and this one uses Warning) we will log to the old logger with dangling references.

This problem materialized in a segmentation fault on ppc64el architecture.

### DIFF
--- a/tests/test_LogOutputHelper.cpp
+++ b/tests/test_LogOutputHelper.cpp
@@ -170,6 +170,8 @@ BOOST_AUTO_TEST_CASE(Cumulative)
     helper.cumulative(0);
     std::string data = trimStream(str);
     BOOST_CHECK_EQUAL(data, reference);
+    // Cleanup backends with references to local variables
+    Opm::OpmLog::removeBackend("stream");
 }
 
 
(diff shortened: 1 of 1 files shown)


blattms:
Still need to check whether this really fixes the problems I am seeing, but I am confident that it will. Famous last words :sweat_smile:

blattms:
jenkins build this please

blattms:
It fixes the problem. Sooner or later this will bite someone on amd64, too...

bska (commented):
When there are common teardown actions, it is usually better to defer those to the destructor of a per-case fixture-class.  I think it makes a lot of sense to create a fixture here, because we could put the setup action in the fixture's constructor too.

For instance, something along the lines of
```C++
struct LogTestFixture
{
    LogTestFixture() { Opm::OpmLog::addBackend("stream", ...); }
    ~LogTestFixture() { Opm::OpmLog::removeBackend("stream"); }

    std::stringstream str;
};

BOOST_FIXTURE_TEST_CASE(Cumulative, LogTestFixture)
{
    // ...
}
```

blattms:
Damn, that was quite an effort. Done.

blattms:
jenkins build this please.

bska:
> please.

Looks like the trigger does not recognise the trailing full stop (`.`).  I'll retrigger the build.

bska:
jenkins build this please

bska (approved):

PR #5267, opened 2024-04-03
fixed: segfault due to taking a reference to temporary

akva2:
The WBP calculation service holds on to a reference to the given parameter so we have to use a local and not a temporary.


akva2:
jenkins build this please

bska (approved):

PR #5269, opened 2024-04-03
adjust to EWOMS_END_PARAM_REGISTRATION macro removal

akva2:
Downstream of https://github.com/OPM/opm-models/pull/884


bska (approved):

PR #5270, opened 2024-04-03
adjust to EWOMS_PARAM_IS_SET macro removal

akva2:
Downstream of https://github.com/OPM/opm-models/pull/885


bska (approved):

PR #5271, opened 2024-04-04
adjust to EWOMS_HIDE_PARAM macro removal

akva2:
Downstream of https://github.com/OPM/opm-models/pull/887


bska (approved):
It's a little annoying that we have to pass string literals here, but on the other hand there's not really any difference compared to the original code.  That still required care that you passed the same parameter name both in the `_REGISTER_PARAM` and in the `_HIDE_PARAM` macros.

I'll merge into master.

akva2:
yeah. they will disappear at the end of the chain.

PR #5272, opened 2024-04-04
Adjust to EWOMS_REGISTER_PARAM macro removal

akva2:
Downstream of https://github.com/OPM/opm-models/pull/890


bska (approved):

PR #5275, opened 2024-04-05
adjust to Parameters::hideParam change

akva2:
Downstream of https://github.com/OPM/opm-models/pull/891


bska (approved):

PR #5276, opened 2024-04-05
adjust to removal of EWOMS_GET_PARAM

akva2:
Downstream of https://github.com/OPM/opm-models/pull/892


bska (approved):

PR #5281, opened 2024-04-12
Summary for immobile free-phase at saturations for which the nonwetting relative permeability equals zero

daavid00:
See https://github.com/OPM/opm-common/pull/4010


bska (approved):

PR #5284, opened 2024-04-15
Fix fast amg wrapper and generalize RebuildOnUpdatePreconditioner

multitalentloes:
Atgeirr suggested that the famg preconditioner should have an update that reconstructs the preconditioner, this is fixed in this PR.
While doing this the RebuildOnUpdatePreconditioner wrapper class was made more general using parameter packs to keep the factory as clean as possible.

### DIFF
--- a/opm/simulators/linalg/PreconditionerWithUpdate.hpp
+++ b/opm/simulators/linalg/PreconditionerWithUpdate.hpp
@@ -99,32 +99,34 @@ struct GeneralPreconditionerMaker {
 /// @tparam ...Args - All arguments needed to construct the preconditioner of choice
 template <class OriginalPreconditioner, class... Args>
 struct PreconditionerMaker : public GeneralPreconditionerMaker<OriginalPreconditioner> {
+    using GenericPreconditioner = Preconditioner<typename OriginalPreconditioner::domain_type, typename OriginalPreconditioner::range_type>;
+    
     PreconditionerMaker(Args&&... args)
         : args_(args...)
     {
     }
-    std::unique_ptr<
-        Preconditioner<typename OriginalPreconditioner::domain_type, typename OriginalPreconditioner::range_type>>
+    std::unique_ptr<GenericPreconditioner>
     make() override
     {
-        return std::unique_ptr<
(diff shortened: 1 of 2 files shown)


atgeirr:
jenkins build this please

opm/simulators/linalg/PreconditionerWithUpdate.hpp
```diff
@@ -84,18 +84,49 @@ getDummyUpdateWrapper(Args&&... args)
...
+struct PreconditionerMaker : public GeneralPreconditionerMaker<OriginalPreconditioner> {
+    PreconditionerMaker(Args&&... args)
+        : args_(args...)
+    {
+    }
+    std::unique_ptr<
+        Preconditioner<typename OriginalPreconditioner::domain_type, typename OriginalPreconditioner::range_type>>
+    make() override
+    {
+        return std::unique_ptr<
```
atgeirr, line 110: I suggest avoiding the explicit `new` by using `make_unique`:
```
return std::apply([](auto...args){
    return std::make_unique<OriginalPreconditioner>(args...);
}, args_);
```

bska: > ```c++
> return std::apply(
>     [](auto...args){
>         return std::make_unique<OriginalPreconditioner>(args...);
>     }, args_);
> ```

Better still:
```C++
return std::apply(
    [](auto&&... args) {
        return std::make_unique<OriginalPreconditioner>(std::forward<Args>(args)...);
    }, args_);
```
There's no need to copy the arguments in this case.

multitalentloes: I now use the latest suggestion

multitalentloes:
The latest commits rewrites the wrapper function that is templated with the arguments to the preconditioner constructor such that the wrapper itself does not need to be. This avoids adding extra templates to the factory, while still allowing the RebuildOnUpdatePreconditioner to be generic with respect to construction parameters.

atgeirr (approved):

atgeirr:
jenkins build this please

PR #5285, opened 2024-04-16
changed: make PreconditionerType an enum class

akva2:
this to avoid symbol clashes with the implementations. while at it rename it to Type as Preconditioner::PreconditionerType is redundant.

I ran into this while working in this area and found it a bit annoying so let's improve.

### DIFF
--- a/opm/simulators/linalg/bda/opencl/Preconditioner.cpp
+++ b/opm/simulators/linalg/bda/opencl/Preconditioner.cpp
@@ -44,13 +44,13 @@ void Preconditioner<block_size>::setOpencl(std::shared_ptr<cl::Context>& context
 
 template <unsigned int block_size>
 std::unique_ptr<Preconditioner<block_size>>
-Preconditioner<block_size>::create(Type type, int verbosity, bool opencl_ilu_parallel)
+Preconditioner<block_size>::create(Type type, bool opencl_ilu_parallel, int verbosity)
 {
     switch (type ) {
     case Type::BILU0:
         return std::make_unique<BILU0<block_size> >(opencl_ilu_parallel, verbosity);
     case Type::CPR:
-        return std::make_unique<CPR<block_size> >(verbosity, opencl_ilu_parallel);
+        return std::make_unique<CPR<block_size> >(opencl_ilu_parallel, verbosity);
     case Type::BISAI:
         return std::make_unique<BISAI<block_size> >(opencl_ilu_parallel, verbosity);
     }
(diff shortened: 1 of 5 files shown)


akva2:
jenkins build this please

opm/simulators/linalg/bda/opencl/Preconditioner.cpp
```diff
@@ -40,13 +40,15 @@ void Preconditioner<block_size>::setOpencl(std::shared_ptr<cl::Context>& context
...
+{
+    if (type == Type::BILU0) {
+        return std::make_unique<BILU0<block_size> >(opencl_ilu_parallel, verbosity);
+    } else if (type == Type::CPR) {
+        return std::make_unique<CPR<block_size> >(verbosity, opencl_ilu_parallel);
+    } else if (type == Type::BISAI) {
+        return std::make_unique<BISAI<block_size> >(opencl_ilu_parallel, verbosity);
     } else {
         OPM_THROW(std::logic_error, "Invalid PreconditionerType");
     }
```
bska, line 54: Would it make sense to replace this with a `switch` instead?  That way, the compiler will warn us if we forget one of the alternatives when introducing a new enumerator.  E.g.,
```C++
switch (type) {
case Type::BILU0:
    return ...
    
case Type::CPR:
    return ...

case Type::BISAI:
    return ...
}

OPM_THROW(std::logic_error, "Unsupported preconditioner type " + std::to_string(static_cast<int>(type)));
```

bska (approved):
I didn't notice before, but does the `CPR` preconditioner really reverse its constructor arguments to the other two?  The others are `(opencl_ilu_parallel, verbosity)`, but the CPR one is `(verbosity, opencl_ilu_parallel)`.  Weird.

In any case, this looks good to me.  The only thing I might consider doing differently is replacing the `if`&ndash;`else` chain with a `switch` on the preconditioner type.

opm/simulators/linalg/bda/opencl/Preconditioner.cpp
akva2: I feared i would fall in that rabbit hole haha. Working in this code I'm constantly fighting the urge to inflate the commits with all sorts of cosmetics and formatting things. Anyways since you called these out; yes the switch makes sense, done. The inconsistent order is indeed inconsistent. I've take care of that as well.

akva2:
jenkins build this please

opm/simulators/linalg/bda/opencl/Preconditioner.cpp
bska: > Working in this code I'm constantly fighting the urge to inflate the commits with all sorts of cosmetics and formatting things.

Right.  If you're doing multiple steps, then we can hold off on those kinds of things until everything else has landed.  Just let me know and then I'll not make those kinds of comments.

akva2: Oh, please do make suggestion, it only makes the code better.. Don't read it as annoyance with those. It's just the BDA code is really out of wack when it comes to formatting and general little niggles. Too few eyes..

bska (approved):

PR #5286, opened 2024-04-16
Add support for using float Scalars in simulators

akva2:
Downstream of https://github.com/OPM/opm-common/pull/4016
Downstream of https://github.com/OPM/opm-models/pull/896

These are the changes necessary to support float Scalars.

Some TODOs left:
- UMFPack cannot handle floats. This again means MSW is disabled with floats. SuperLU supports floats so that is a possible solution for this.

Runtime has been lightly tested with the BDA gpu solvers, opencl, cusparse and amgcl (vexcl and cuda backends) and dune cpu solvers. It seems to work although there is non-uniform results (opencl seems to do better than cusparse for some reason).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -736,6 +736,18 @@ if (BUILD_FLOW)
     PASS_REGULAR_EXPRESSION "${${project}_LABEL}")
 endif()
 
+if(BUILD_FLOW_FLOAT_VARIANTS)
+  opm_add_test(flow_blackoil_float
+    ONLY_COMPILE
+    ALWAYS_ENABLE
+    DEFAULT_ENABLE_IF ${FLOW_DEFAULT_ENABLE_IF}
+    DEPENDS opmsimulators
+    LIBRARIES opmsimulators
+    SOURCES
+    flow/flow_blackoil_float_main.cpp
+    $<TARGET_OBJECTS:moduleVersion>)
+endif()
+
 if (OPM_ENABLE_PYTHON)
   add_subdirectory(python)
 endif()
--- a/flow/flow_blackoil_float_main.cpp
+++ b/flow/flow_blackoil_float_main.cpp
@@ -1,0 +1,38 @@
+/*
+  Copyright 2020, NORCE AS
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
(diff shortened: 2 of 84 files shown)


bska (commented):
Other than a few instances of "construct + copy/convert" for `std::vector<>` which could **probably** use the converting constructor from a sequence directly, and similar for assignment this looks good to me.  The explicit specialisation macros are less than ideal, but I don't have a better suggestion.  I'd like to get this into the mainline to start experimenting with the feature.

akva2:
okay, I have pushed a bisectable version (ie it builds at each step) incorporating your suggestions. I guess this is ready for review now.

opm/simulators/flow/GenericOutputBlackoilModule.cpp
```diff
@@ -451,10 +451,19 @@ assignToSolution(data::Solution& sol)
...
+                       std::get<UnitSystem::measure>(entry),
+                       std::move(std::get<2>(entry)),
+                       target);
+        } else {
+            const auto& input = std::get<2>(entry);
+            std::vector<double> data(input.begin(), input.end());
+            sol.insert(std::get<std::string>(entry),
+                       std::get<UnitSystem::measure>(entry),
+                       std::move(data),
+                       target);
```
bska, line 465: Minor nit: [`Solution::insert()`]() takes a `vector<double>` by value.  Could you not call this as
```C++
const auto& input = std::get<2>(entry);
solution.insert(..., std::vector<double>(input.begin(), input.end()), ...);
```
instead?

opm/simulators/flow/GenericOutputBlackoilModule.cpp
```diff
@@ -599,22 +617,22 @@ assignToSolution(data::Solution& sol)
     if (! this->temperature_.empty())
     {
         sol.insert("TEMP", UnitSystem::measure::temperature,
-                   std::move(this->temperature_), data::TargetType::RESTART_SOLUTION);
+                   std::move(getArray(this->temperature_)), data::TargetType::RESTART_SOLUTION);
```
bska, line 620: Hm, `move(getArray(...))` looks wrong to me.  The output module (regrettably) depends on `x.empty()` being `true` once the `x` array has been inserted into the `Solution` container.  I think we may have to rewrite `getArray()` as
```C++
auto getArray(std::vector<Scalar>&& input)
{
    if constexpr (std::is_same_v<Scalar, double>) {
        return input;
    }
    else {
        return std::vector<double>{input.begin(), input.end()};
    }
}
```
and then write
```C++
sol.insert(..., getArray(std::move(this->...)), ...)
```
instead.

bska (commented):
Thanks a lot for the updates.  Other than the `move()` calls in the output module this looks good to me now.  I do think we should revise the ownership model for the large solution arrays and when to update those, but until we do so we should ensure that the member arrays are empty when the corresponding value has been inserted into the `data::Solution` object.

opm/simulators/flow/GenericOutputBlackoilModule.cpp
akva2: changed (a bit differently as discussed elsewhere).

opm/simulators/flow/GenericOutputBlackoilModule.cpp
akva2: sure.

blattms:
umfpack might support single precision in the future, see DrTimothyAldenDavis/SuiteSparse#446

PR #5289, opened 2024-04-17
WellState: remove template parameters for comm

akva2:
use Parallel::Communication directly instead


bska (approved):
Much appreciated.  I'll merge into master.

PR #5292, opened 2024-04-17
Update GMDS and GMGP definitions in GenericOutputBlackoilModule.cpp

gdfldm:
Should GMDS be CO2 dissolved in water phase and GMGP be CO2 in gas phase (rather than the other way around)?


totto82:
jenkins build this please

totto82:
Thanks for noticing. Will merge when green.

Issue #5279, opened 2024-04-11
Initialisation of GW cases from saturation table


rolf-myhr-prores:
![image](https://github.com/OPM/opm-simulators/assets/70645560/4728082f-12eb-42a9-b6bd-d38e3bf50d37)
When initializing a gas-water case from a saturation table (without SWATINIT) by using familiy 2 saturation keywords (SWFN/SGFN), no transition zone is generated, as seen in figure.
[gw_init.zip](https://github.com/OPM/opm-simulators/files/14947347/gw_init.zip)

totto82:
Thanks for reporting. I am able to reproduce the issue and have detected the culprit in the code. Will try to provide a fix before the release.

PR #5293, opened 2024-04-18
Fix gas-water initialization with transision zone

totto82:
I preliminary quick fix. Need to test a bit more before merging.

### DIFF
--- a/opm/simulators/flow/equil/EquilibrationHelpers_impl.hpp
+++ b/opm/simulators/flow/equil/EquilibrationHelpers_impl.hpp
@@ -446,10 +446,15 @@ operator()(double s) const
     fluidState.setSaturation(FluidSystem::waterPhaseIdx, 0.0);
     fluidState.setSaturation(FluidSystem::oilPhaseIdx, 0.0);
     fluidState.setSaturation(FluidSystem::gasPhaseIdx, 0.0);
-    fluidState.setSaturation(phase_, s);    
-    if (phase_ == FluidSystem::gasPhaseIdx)
-        fluidState.setSaturation(FluidSystem::waterPhaseIdx, 1.0 - s);
+    fluidState.setSaturation(phase_, s);
 
+    // This code should only be called for water or gas phase
+    assert(phase_ != FluidSystem::oilPhaseIdx);
+    // The capillaryPressure code only uses the wetting phase saturation
+    if (phase_ == FluidSystem::gasPhaseIdx) {
+        fluidState.setSaturation(FluidSystem::waterPhaseIdx, 1.0 - s);
+        fluidState.setSaturation(FluidSystem::oilPhaseIdx, 1.0 - s);
+    }
     std::array<double, FluidSystem::numPhases> pc{0.0};
     using MaterialLaw = typename MaterialLawManager::MaterialLaw;
     MaterialLaw::capillaryPressures(pc, matParams, fluidState);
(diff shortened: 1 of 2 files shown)


totto82:
jenkins build this please

totto82:
https://github.com/OPM/opm-simulators/issues/5279

totto82:
jenkins build this please

lisajulia:
I would like to include this into the release candidate 2, so I'll wait until this is merged to master until I create the new release candidate, ok?

lisajulia:
> image image
> 
> I have tested this PR on a simple box model with CO2/BRINE for both the oil-gas and the water-gas model.
> 
> With
> 
> ```
> EQUIL
> 0.0 200 1000 0 5 0.0 1 0 0 /
> /
> ```
> 
> for the gas-oil case and
> 
> ```
> EQUIL
> 0.0 200 5 0.0 1* 0 0 0 0 /
> /
> ```
> 
> for the gas-water case.
> 
> And they agree and gives results are expected.

Is is needed / beneficial to add this test to opm-tests? :eyes: If not, then nevermind :).

opm/simulators/flow/equil/EquilibrationHelpers_impl.hpp
```diff
@@ -448,6 +448,13 @@ operator()(double s) const
     fluidState.setSaturation(FluidSystem::gasPhaseIdx, 0.0);
     fluidState.setSaturation(phase_, s);
 
+    // This code should only be called for water or gas phase
+    assert(phase_ != FluidSystem::oilPhaseIdx);
+    // The capillaryPressure code only uses the wetting phase saturation
+    if (phase_ == FluidSystem::gasPhaseIdx) {
+        fluidState.setSaturation(FluidSystem::waterPhaseIdx, 1.0 - s);
```
atgeirr, line 455: Should this not be 0? Or is this a small trick to handle both gas-water and gas-oil systems?

opm/simulators/flow/equil/InitStateEquil_impl.hpp
```diff
@@ -578,9 +578,11 @@ deriveSaturations(const Position& x,
     this->setEvaluationPoint(x, reg, ptable);
     this->initializePhaseQuantities();
 
-    if (ptable.waterActive()) { this->deriveWaterSat(); }
     if (ptable.gasActive())   { this->deriveGasSat();   }
 
+    if (ptable.waterActive()) { this->deriveWaterSat(); }
```
atgeirr, line 583: Why is the order changed for the gasActive() and waterActive() related lines?

Nitpick: double blank line below.

atgeirr (commented):
Just two questions, otherwise ok.

opm/simulators/flow/equil/EquilibrationHelpers_impl.hpp
totto82: Yes. It the later code will only use the wetting phase saturation. So water for gas-water and oil-water systems and oil for oil-gas systems.

opm/simulators/flow/equil/InitStateEquil_impl.hpp
totto82: For 2p gas-water Sw = 1 - Sg. So we need to compute the gas saturation first.

lisajulia:
@totto82, @atgeirr: Sorry for asking you once more about that: By when do you think this will be merged?
I'd like to create the tags for the release candidate 2 by tomorrow, and it would be great if we could have this in as well :)!
Thanks!

totto82:
From my point of view this is ok. @atgeirr Please merge if you are happy with my comments.

PR #5295, opened 2024-04-19
WellGroupHelpers: template Scalar type

akva2:
Make WellGroupHelpers and some related classes templates over the Scalar type.
To avoid having to explicitly instance every single function (now template) in there I have changed WellGroupHelpers from a namespace to a class with static members.


akva2:
jenkins build this please

atgeirr:
Thanks, merging. A minor note is that we may want to unify what now becomes the WGHelpers namespace and the WellGroupHelpers class in the future.

PR #5296, opened 2024-04-19
Redistribute gaslift for groups with alq limit

totto82:
In master gaslift was only redistributed for group limited by rates

### DIFF
--- a/opm/simulators/wells/GasLiftStage2.cpp
+++ b/opm/simulators/wells/GasLiftStage2.cpp
@@ -411,7 +411,7 @@ optimizeGroup_(const Group &group)
 {
     const auto& group_name = group.name();
     const auto prod_control = this->group_state_.production_control(group_name);
-    if ((prod_control != Group::ProductionCMode::NONE) && (prod_control != Group::ProductionCMode::FLD))
+    if (this->glo_.has_group(group.name()) || ((prod_control != Group::ProductionCMode::NONE) && (prod_control != Group::ProductionCMode::FLD)))
     {
         if (this->debug) {
             const std::string msg = fmt::format("optimizing (control = {})", Group::ProductionCMode2String(prod_control));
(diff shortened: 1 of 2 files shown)


totto82:
jenkins build this please

opm/simulators/wells/GasLiftStage2.cpp
```diff
@@ -938,11 +938,11 @@ addOrRemoveALQincrement(GradMap &grad_map, const std::string& well_name, bool ad
 
 bool
 GasLiftStage2::SurplusState::
-checkALQlimit(double delta_alq, double delta_gas)
+checkALQlimit()
 {
     if (this->max_glift) {
         double max_alq = *(this->max_glift);
-        if ((max_alq) < (this->alq + delta_alq)  ) {
+        if ((max_alq) < (this->alq)  ) {
```
hakonhagland, line 945: It is not clear to me why this change will allow ALQ to be removed from groups limited by ALQ. Maybe add a comment here to clarify the reasoning?

totto82: This is just a bug fix. For group control rates we would like to make sure that we dont remove so much gaslift that the  group rate can not be reached (hence the delta_) in the check. For ALQ the limits are strict and we shouldn't add a delta_alq.

hakonhagland: Ok, I see, so it is not possible to remove too much ALQ in this case. Is this because the the increments will always be divisible by the group ALQ limit? What happens if a group has both ALQ and oil/gas/water limits? Can you add this as comment in the source code too? I think it could help us in the future to have some explanation.

hakonhagland: Thanks!

totto82:
jenkins build this please

PR #5297, opened 2024-04-19
fixed: do not send and recv from same buffer

akva2:
even though it's perfectly fine to do so in this case, mpich will error out with an error if send and recv buffer is the same.
fixes mpich based (rhel7/8) builds.


akva2:
jenkins build this please

bska (approved):

PR #5299, opened 2024-04-22
BlackoilWellModel: template Scalar type

akva2:

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -210,22 +210,27 @@ initFromRestartFile(const RestartValue& restartValues,
                              this->schedule(), handle_ms_well, numCells,
                              this->well_perf_data_, this->summaryState_);
 
-    BlackoilWellModelRestart(*this).loadRestartData(restartValues.wells,
-                                                    restartValues.grp_nwrk,
-                                                    handle_ms_well,
-                                                    this->wellState(),
-                                                    this->groupState());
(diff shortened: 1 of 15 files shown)


akva2:
jenkins build this please

opm/simulators/wells/BlackoilWellModelRestart.cpp
```diff
@@ -260,4 +267,6 @@ loadRestartData(const data::Wells&                 rst_wells,
     }
 }
 
+template class BlackoilWellModelRestart<double>;
```
atgeirr, line 270: Is there a way to find all the explicit instantiations like this? For when we need to add a float instantiation, for example.

akva2: there will be a define FLOW_INSTANTIATE_FLOAT that you can grep for.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -407,7 +414,7 @@ checkGEconLimits(
         return;
     }
 
-    GroupEconomicLimitsChecker<double> checker {
+    GroupEconomicLimitsChecker checker {
```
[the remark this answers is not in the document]
akva2: Okay I like CTAD but will revert to explicit then.

atgeirr (approved):
Looks good and can be merged as is (despite my minor comment about template parameters above).

akva2:
jenkins build this please

PR #5304, opened 2024-04-24
fixed: use the eclipse grid from the eclipse state in test_nonnc 

akva2:
if we use a separate instance the actnum arrays get out of sync. in
particular the two cells with PORV == 0 will not be marked as inactive
in the actnum array of the separate instance. this leads to a segfault
in lookUpData as the field props will have those cells filtered, while the grid
still has the cells active.

also convert to a boost unit test while at it.


akva2:
jenkins build this please

bska (approved):

PR #5307, opened 2024-04-24
Accumulate CNV Pore-Volume Sum for Interior Cells Only

bska:
The `interiorBorder` category is **probably** equivalent to the `interior` category for codimension zero elements, but it's better to be safe than sorry.  We don't want to accumulate pore-volume contributions twice.


bska:
jenkins build this please

bska:
For reference: There is no specific discussion of `interiorBorder` vs. `interior` in PR #2701 (commit e04048458) which first introduced this helper function.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
> Merge when green!

Build check is green.  Merging per prior approval.

PR #5309, opened 2024-04-24
Aquifer: use Scalar type

akva2:
use the Scalar type in Aquifer classes.


akva2:
jenkins build this please

bska (approved):

PR #5310, opened 2024-04-25
FlowExp: spring cleaning

akva2:
Various cosmetics to align with opm standards.


akva2:
jenkins build this please

GitPaean:
Thanks for this. One colleague raised the point that we should really enforce the clangformat throughout OPM repos. What do you think?

akva2:
i'm no big fan of clang format. there will always be exceptions where violating rules is an advantage for readability and then you get forever stuck fighting an obnoxious tool.

GitPaean:
Thanks for the efforts. Just taking this PR for example, most of the changes can be done automatically. I do think if we can avoid spending human time on this is nice.  I am getting the PR in now.

PR #5321, opened 2024-04-29
Adding a separate CNV tolerance for energy (default 1.0e-4)

vkip:


totto82:
I agree that it make sense to add an option for setting the tolerances for energy separately from mass. Should we add it for the L2 error as well? (ala mB, not only CNV).  I also suggest that we do this in two steps. First add the option of setting the tolerances, than changing the defaults

vkip:
> I agree that it make sense to add an option for setting the tolerances for energy separately from mass. Should we add it for the L2 error as well? (ala mB, not only CNV). I also suggest that we do this in two steps. First add the option of setting the tolerances, than changing the defaults

Good points! Done. (I think.)

vkip:
jenkins build this please

totto82:
Thanks for the update. Looks good to me and Jenkins is happy.

totto82:
I just changed the title to adapt it to its content

PR #5323, opened 2024-04-29
Apply well/group efficiency factors for connection energy rates. 

vkip:
Ensure well and accumulated group efficiency factors are applied to the connection energy rates.


totto82:
Thanks. This is trivial to merge. Sorry for the 3mnd waiting time.

PR #5325, opened 2024-05-01
Allow Python bindings to control MPI setup in the OPM::Main constructor

hakonhagland:
Allows the Python bindings to control whether `MPI_Init()` and `MPI_Finalize()` will be called when creating an `Opm::Main` object. This allows the Python unit test files to run more than a single unit test. Currently, only a single unit test could be run per test file since each test would create a new simulator object (calling the `Opm::Main` constructor) which would call `MPI_Init()` a second time (after `MPI_Finalize()` had been called for the first test case) which would give an error like:

```
*** The MPI_Init() function was called after MPI_FINALIZE was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
 ```

(This is because it is not possible to unload the Python module `opm.simulators.BlackOilSimulator` (which btw links dynamically with `libmpi.so`) within a single Python script)

An idea for a future PR could be to allow the Python bindings to switch off MPI completely.

### DIFF
--- a/python/docstrings.json
+++ b/python/docstrings.json
@@ -27,9 +27,9 @@
         "signature": "get_dt() -> float",
         "doc": "Get the timestep size of the last completed step.\n\n:return: Timestep size in days.\n:type return: float"
     },
-    "getFluidStateVariables": {
-        "signature": "get_fluid_state_variables(name: str) -> NDArray[float]",
-        "doc": "Retrieve a fluid state variable for the simulation grid.\n\n:para name: The name of the variable. Valid names are 'pw' (pressure water), 'pg' (pressure gas), 'po' (pressure oil), 'rho_w' (density water), 'rho_g' (density gas), 'rho_o' (density oil)'Rs' (soultion gas-oil ratio), 'Rv' (volatile gas-oil ratio), 'Sw' (water saturation), 'Sg' (gas saturation), 'So' (oil saturation), and 'T' (temperature).\n:type name: str\n\n:return: An array of fluid state variables.\n:type return: NDArray[float]"
+    "getFluidStateVariable": {
+        "signature": "get_fluid_state_variable(name: str) -> NDArray[float]",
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by hakonhagland** [comment]
jenkins build this please

File: python/docstrings.json [hunk 1]
```diff
@@ -27,25 +27,25 @@
         "signature": "get_dt() -> float",
         "doc": "Get the timestep size of the last completed step.\n\n:return: Timestep size in days.\n:type return: float"
     },
-    "getFluidStateVariables": {
-        "signature": "get_fluid_state_variables(name: str) -> NDArray[float]",
-        "doc": "Retrieve a fluid state variable for the simulation grid.\n\n:para name: The name of the variable. Valid names are 'pw' (pressure water), 'pg' (pressure gas), 'po' (pressure oil), 'rho_w' (density water), 'rho_g' (density gas), 'rho_o' (density oil)'Rs' (soultion gas-oil ratio), 'Rv' (volatile gas-oil ratio), 'Sw' (water saturation), 'Sg' (gas saturation), 'So' (oil saturation), and 'T' (temperature).\n:type name: str\n\n:return: An array of fluid state variables.\n:type return: NDArray[float]"
+    "getFluidStateVariable": {
```
akva2, line 30: maybe split out the docstring fixes in a separate commit?

hakonhagland:
jenkins build this please

python/docstrings.json
hakonhagland: Yes, I thought it would not be worth it but it might be a good ida to keep the commits and PRs strictly on topic. Btw, I think #5243 also includes these docstring fixes.

akva2: it will make it easier in two years if we look back in history because something is broken with mpi. so i would really prefer commits to be kept on topic. and the two commits here squashed.

hakonhagland:
jenkins build this please

python/docstrings.json
hakonhagland: Ok, noted! Should I revert docstring changes to `PyBlackOilSimulator.cpp` also?

akva2: ah missed those. yes please keep out of the mpi commit. if you want you can include all those changes in this PR, that is no problem, just keep it in separate commits.

hakonhagland:
jenkins build this please

python/docstrings.json
akva2: perfect, thanks, sorry to be a PITA ;)

hakonhagland:
jenkins build this please

PR #5326, opened 2024-05-02
fixed: properly avoid inclusion of BDA types if BDA is disabled

akva2:


akva2:
jenkins build this nobda please

bska (approved):

PR #5328, opened 2024-05-02
Added: Python test case with two simulators

hakonhagland:
Added a test case to demonstrate that two simulators can be run simultaneously from Python. This is useful for simulating coupled reservoirs.


hakonhagland:
jenkins build this please

bska:
jenkins build this please

blattms:
I think we need to test with different models as the goal is to support different physics (maybe same phases, but different properties/tables).

I have seen [comments in the python code](https://github.com/OPM/opm-simulators/blob/ae613610616775298f3ec30d46edcb2806daa7ce/python/test/test_basic.py#L10-L26) indicating that this is not supported (as in getting correct results).

PR #5330, opened 2024-05-03
Rerun UDQ Assignments After Action Processing

bska:
If an action block happens to run a UDQ ASSIGN operation and, furthermore, happens to run at the last time step of an episode then the ["clear pending assignments"](https://github.com/OPM/opm-common/blob/6af6865727aa4b513adf40b37f56656772fcf7f3/opm/input/eclipse/Schedule/ScheduleState.cpp#L160-L168) behaviour of the `ScheduleState` copy constructor leads to not performing the UDQ assignment at all.  This PR works around this problem by invoking the action handler's UDQ assignment function after processing all active action blocks.

The underlying problem has been present since at least PR OPM/opm-common#3587 which introduced the "clear pending assignments" behaviour of `ScheduleState`'s copy constructor.

While here, also make a few whitespace adjustments to `FlowProblem::endTimeStep()` to be more consistent with the rest of the implementation.

### DIFF
--- a/opm/simulators/flow/FlowProblem.hpp
+++ b/opm/simulators/flow/FlowProblem.hpp
@@ -707,6 +707,19 @@ public:
                         return vg.gridIdxToEquilGridIdx(i);
                     });
             });
+
+            // Rerun UDQ assignents following action processing to make sure
+            // that any UDQ ASSIGN operations triggered in action blocks
+            // take effect.  This is mainly to work around a shortcoming of
+            // the ScheduleState copy constructor which clears pending UDQ
+            // assignments under the assumption that all such assignments
+            // have been processed.  If an action block happens to trigger
+            // on the final time step of an episode and that action block
+            // runs a UDQ assignment, then that assignment would be dropped
+            // and the rest of the simulator will never see its effect
+            // without this hack.
+            this->actionHandler_
+                .evalUDQAssignments(episodeIdx, this->simulator().vanguard().udqState());
         }
 
         // deal with "clogging" for the MICP model
--- a/opm/simulators/flow/FlowProblem.hpp
+++ b/opm/simulators/flow/FlowProblem.hpp
@@ -632,80 +632,93 @@ public:
     void endTimeStep()
     {
         OPM_TIMEBLOCK(endTimeStep);
+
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

blattms (approved):
Looks good.

One question: The call `this->actionHandler_.evalUDQAssignments(...)` in beginTimeStep will only have any effect on the first time step of an episode, because for the rest the assignment already happened at endTimeStep of the last time step? Or are they run twice now?

bska:
> The call `this->actionHandler_.evalUDQAssignments(...)` in beginTimeStep will only have any effect on the first time step of an episode

You're **probably** thinking of [this statement](https://github.com/OPM/opm-simulators/blob/8a4e78e7c4aecdf42190782d526488f575999d8a/opm/simulators/flow/FlowProblem.hpp#L565-L567).  That's triggered from [`beginEpisode()`](https://github.com/OPM/opm-simulators/blob/8a4e78e7c4aecdf42190782d526488f575999d8a/opm/simulators/flow/FlowProblem.hpp#L513) I think, and not from [`beginTimeStep()`](https://github.com/OPM/opm-simulators/blob/8a4e78e7c4aecdf42190782d526488f575999d8a/opm/simulators/flow/FlowProblem.hpp#L582).  That said, 

> are they run twice now?

Even if they were, that's where we get a bit of protection from the [pending assignment structure](https://github.com/OPM/opm-common/blob/5c2ec635a92a2003f8e770a4ce76b3a6febdb6ae/opm/input/eclipse/Schedule/UDQ/UDQConfig.hpp#L189) in the `UDConfig` class.  Every time we complete a round of UDQ assignments, we [clear the list of pending assignments](https://github.com/OPM/opm-common/blob/5c2ec635a92a2003f8e770a4ce76b3a6febdb6ae/opm/input/eclipse/Schedule/UDQ/UDQConfig.cpp#L576-L578) meaning if there are no new UDQ ASSIGN operations triggered between the end of the previous timestep and the start of the next, then we won't actually trigger the assignment twice.

PR #5331, opened 2024-05-03
Bugfix: initialize the input temperature in restart run without temperature in the restart file

totto82:
The issue is related to restart of cases using CO2STORE

### DIFF
--- a/opm/simulators/flow/FlowProblem.hpp
+++ b/opm/simulators/flow/FlowProblem.hpp
@@ -2221,6 +2221,14 @@ protected:
                 }
             }
 
+            // For CO2STORE and H2STORE we need to set the initial temperature for isothermal simulations
+            bool isThermal = eclState.getSimulationConfig().isThermal();
+            bool needTemperature = (eclState.runspec().co2Storage() || eclState.runspec().h2Storage());
+            if (!isThermal && needTemperature) {
+                const auto& fp = simulator.vanguard().eclState().fieldProps();
+                elemFluidState.setTemperature(fp.get_double("TEMPI")[elemIdx]);
+            }
+
             this->mixControls_.updateLastValues(elemIdx, elemFluidState.Rs(), elemFluidState.Rv());
 
             if constexpr (enablePolymer)


totto82:
jenkins build this please

opm/simulators/flow/OutputBlackoilModule.hpp
```diff
@@ -1115,6 +1115,10 @@ class OutputBlackOilModule : public GenericOutputBlackoilModule<GetPropType<Type
 
         if (!this->temperature_.empty())
             fs.setTemperature(this->temperature_[elemIdx]);
+        else {
+            const auto& fp = simulator_.vanguard().eclState().fieldProps();
+             fs.setTemperature(fp.get_double("TEMPI")[elemIdx]);
+        }
```
bska, line 1121: I don't think this will work in general.  You're not supposed to have `TEMPI`-style information in restarted runs.

bska (commented):
I don't think I understand what problem this is trying to solve.  If we're restarting a run and there is no temperature information in the restart file, then does that not mean that it's an isothermal run?  If so, why do we need temperature data?

totto82:
> I don't think I understand what problem this is trying to solve. If we're restarting a run and there is no temperature information in the restart file, then does that not mean that it's an isothermal run? If so, why do we need temperature data?

For CO2STORE you will need the initial temperature in the reservoir. We used to store it in the restart files, but that was removed since it is constant in isothermal runs.

opm/simulators/flow/OutputBlackoilModule.hpp
totto82: Isnt TEMPI defaulted to a constant reservoir temperature.

bska:
> > I don't think I understand what problem this is trying to solve. If we're restarting a run and there is no temperature information in the restart file, then does that not mean that it's an isothermal run? If so, why do we need temperature data?
> 
> For CO2STORE you will need the initial temperature in the reservoir. We used to store it in the restart files, but that was removed since it is constant in isothermal runs.

Ah, okay.   Then I think

1. This should be contingent on CO2STORE being active
2. The temperature initialisation itself should be in a separate function which is only invoked for CO2STORE, and only if temperature information is not already loaded from the restart file.

As currently written, I think the risk of confusion is too great.

totto82:
jenkins build this please

opm/simulators/flow/OutputBlackoilModule.hpp
bska: > Isn't `TEMPI` defaulted to a constant reservoir temperature.

Kind of (commit OPM/opm-common@3c9f56a14, current version in [`TableManager.cpp`](https://github.com/OPM/opm-common/blob/95795867abc0a2a9a8af1a568387fd4799b0de10/opm/input/eclipse/EclipseState/Tables/TableManager.cpp#L145-L147)), but it's usually not a good idea to depend on that.  You'll [just get](https://github.com/OPM/opm-common/blob/95795867abc0a2a9a8af1a568387fd4799b0de10/opm/input/eclipse/EclipseState/Grid/FieldProps.cpp#L1208-L1223) whatever temperature value happens to be the default in the [`RTEMP` keyword](https://github.com/OPM/opm-common/blob/95795867abc0a2a9a8af1a568387fd4799b0de10/opm/input/eclipse/share/keywords/000_Eclipse100/R/RTEMP#L12).  That temperature value may or may not be suitable for CO2STORE.

bska (approved):
I would really have liked that the new logic be placed in a separate function, not least because the condition does not depend on the current cell, but I suppose we can refactor that at a later time.  I'll merge into master.

PR #5253, opened 2024-03-13
Add HIP support

multitalentloes:
This PR adds support for using hipify on the cuda code to allow for GPU support on both AMD and Nvidia architectures. The HIP code is hidden from the programmer in the build directory, and the cmake adds dependencies to only recompile the necessary files when making changes in the cuda source code.

This PR is still in draft mode as several temporary changes have been made to make a cmake file.

PR #5332, opened 2024-05-03
remove usage of thrust

multitalentloes:
To solve compilation issues with cuda&hip in PR #5253 this contribution removes the usage of thrust in OPM.
This only requires changing one part of the GPU dot product implementation which is only ever used in multi-gpu MPI code, which is not a prominent use case in OPM. The PR introduces slower sum reduction, but I do not expect it to be a problem, TODO added.


multitalentloes:
jenkins build this please

akva2:
jenkins build this please

PR #5333, opened 2024-05-05
removing unused updateImbnum_ function

GitPaean:
found the function unused when looking at HYSTER related code.

### DIFF
--- a/opm/simulators/flow/FlowGenericProblem.hpp
+++ b/opm/simulators/flow/FlowGenericProblem.hpp
@@ -334,7 +334,6 @@ protected:
     void updateSatnum_();
     void updateMiscnum_();
     void updatePlmixnum_();
-    void updateKrnum_();
 
     const EclipseState& eclState_;
     const Schedule& schedule_;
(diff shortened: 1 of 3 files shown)


GitPaean:
jenkins build this please

bska (commented):
For what it's worth, the function was introduced in #4384 and looks intentional.  Maybe @hakonhagland would like to have a word before we remove this.

hakonhagland:
@bska I will have a look at this later today. I do not remember exactly, but I think the intention was that there should be a call to  `this->updateImbnum_()`  in `readMaterialParameters_()` in `FlowProblem.hpp` see: 

https://github.com/OPM/opm-simulators/blob/66aacb5042b0de5b9af96cef17cf93abcec36663/opm/simulators/flow/FlowProblem.hpp#L2045

hakonhagland:
I checked this again, and it seems we can also remove the `updateKrnum_()` ? Instead of storing the vectors `KRNUM` and `IMBNUM` in class `FlowGenericProblem`, we are storing them in class `EclMaterialManager`, see

https://github.com/OPM/opm-common/blob/9e942d2d2c1c5056da3aaec30e2f699c17c67fb1/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp#L456

GitPaean:
> I checked this again, and it seems we can also remove the `updateKrnum_()` ? Instead of storing the vectors `KRNUM` and `IMBNUM` in class `FlowGenericProblem`, we are storing them in class `EclMaterialManager`, see
> 
> https://github.com/OPM/opm-common/blob/9e942d2d2c1c5056da3aaec30e2f699c17c67fb1/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp#L456
> 
> so the vectors in class `FlowGenericProblem` seems to be unused.

Thanks @hakonhagland . Let me update the PR and let it run through the tests.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
It looks like mostly fine. @bska can you have a look?

bska (approved):
Yep, this is good.  I'll merge into master.

PR #5335, opened 2024-05-06
Tag 'BASIC=0' as allowed in RPTRST

vkip:
Since it appears to be allowed and working: 

https://github.com/OPM/opm-common/blob/95795867abc0a2a9a8af1a568387fd4799b0de10/opm/input/eclipse/Schedule/RSTConfig.cpp#L647-L648


vkip:
jenkins build this please

bska (approved):

PR #5337, opened 2024-05-07
Enforce Relaxed Tolerances Not Stricter Than Strict

bska:
Relaxed tolerances being more strict than the strict tolerances does not make sense in our use case.


bska:
jenkins build this please

PR #5341, opened 2024-05-08
Remove the need for add well contributions to matrix for NLDD

jakobtorben:
The PR addresses the following changes:

1. Make well equations use only perforated cells. Previously, the B and C equations for the well equations had dimensions of the total number of cells but with only non-zero elements for the perforated cells. After this fix it only uses the perforated cells. This requires the global input vectors and matrices in the well operators to be mapped to a local vector for only the perforated cells.
2. Introduce a well operator for NLDD domains, such that when applying the well operator for NLDD, the necessary mapping from global cell indexes to the NLDD subdomain indexes is done and the well apply is only applied to wells on that is in the domain.

These two changes, has made these two new features possible:
1. Remove the need for using add well contributions to matrix for NLDD.
2. Add support for the CPRW preconditioner for NLDD.

### DIFF
--- a/opm/simulators/linalg/ISTLSolver.hpp
+++ b/opm/simulators/linalg/ISTLSolver.hpp
@@ -352,6 +352,8 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
                 // Outch! We need to be able to scale the linear system! Hence const_cast
                 matrix_ = const_cast<Matrix*>(&M);
 
+                is_nldd_local_solver_ = parameters_[activeSolverNum_].is_nldd_local_solver_;
+
                 useWellConn_ = Parameters::Get<Parameters::MatrixAddWellContributions>();
                 // setup sparsity pattern for jacobi matrix for preconditioner (only used for openclSolver)
             } else {
@@ -450,6 +452,11 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
 
         const CommunicationType* comm() const { return comm_.get(); }
 
+        void setDomainIndex(int index)
+        {
+            domainIndex_ = index;
+        }
+
     protected:
 #if HAVE_MPI
         using Comm = Dune::OwnerOverlapCopyCommunication<int, int>;
@@ -490,8 +497,15 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
             OPM_TIMEBLOCK(flexibleSolverPrepare);
             if (shouldCreateSolver()) {
                 if (!useWellConn_) {
-                    auto wellOp = std::make_unique<WellModelOperator>(simulator_.problem().wellModel());
-                    flexibleSolver_[activeSolverNum_].wellOperator_ = std::move(wellOp);
+                    if (is_nldd_local_solver_) {
+                        auto wellOp = std::make_unique<DomainWellModelAsLinearOperator<WellModel, Vector, Vector>>(simulator_.problem().wellModel());
+                        wellOp->setDomainIndex(domainIndex_);
+                        flexibleSolver_[activeSolverNum_].wellOperator_ = std::move(wellOp);
+                    }
+                    else {
+                        auto wellOp = std::make_unique<WellModelOperator>(simulator_.problem().wellModel());
+                        flexibleSolver_[activeSolverNum_].wellOperator_ = std::move(wellOp);
+                    }
                 }
                 std::function<Vector()> weightCalculator = this->getWeightsCalculator(prm_[activeSolverNum_], getMatrix(), pressureIndex);
                 OPM_TIMEBLOCK(flexibleSolverCreate);
@@ -650,6 +664,10 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
         std::vector<detail::FlexibleSolverInfo<Matrix,Vector,CommunicationType>> flexibleSolver_;
         std::vector<int> overlapRows_;
         std::vector<int> interiorRows_;
+
+        bool is_nldd_local_solver_;
+
+        int domainIndex_ = -1;
 
         bool useWellConn_;
 
--- a/opm/simulators/linalg/ISTLSolver.hpp
+++ b/opm/simulators/linalg/ISTLSolver.hpp
@@ -352,8 +352,6 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
                 // Outch! We need to be able to scale the linear system! Hence const_cast
                 matrix_ = const_cast<Matrix*>(&M);
 
-                is_nldd_local_solver_ = parameters_[activeSolverNum_].is_nldd_local_solver_;
-
                 useWellConn_ = Parameters::Get<Parameters::MatrixAddWellContributions>();
                 // setup sparsity pattern for jacobi matrix for preconditioner (only used for openclSolver)
             } else {
@@ -452,9 +450,14 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
 
         const CommunicationType* comm() const { return comm_.get(); }
 
-        void setDomainIndex(int index)
+        void setDomainIndex(const int index)
         {
             domainIndex_ = index;
+        }
+
+        bool isNlddLocalSolver() const
+        {
+            return parameters_[activeSolverNum_].is_nldd_local_solver_;
         }
 
     protected:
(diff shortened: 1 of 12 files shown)


opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -212,7 +214,19 @@ namespace Opm {
             }
             // Apply as Schur complement the well residual to reservoir residuals:
             // r = r - duneC_^T * invDuneD_ * resWell_
-            well->apply(res);
+            // Well equations B and C uses only the perforated cells, so need to apply on local residual
+            auto cells = well->cells();
```
atgeirr, line 218: Use `const auto&` here, current code is making a copy. Also on lines 256, 1575 and 1620. And 1581, and 1627 in a slightly different context. The copies made on 1575 and 1620 are potentially overwritten, if in an NLDD context, so it is not straightforward to make those const references, but require some refactoring.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -203,6 +203,8 @@ namespace Opm {
     BlackoilWellModel<TypeTag>::
     linearize(SparseMatrixAdapter& jacobian, GlobalEqVector& res)
     {
+        GlobalEqVector res_local;
```
atgeirr, line 206: Consider making this a member to avoid reallocation every linearize() call.

atgeirr (commented):
I think this is a good approach! While just a draft now, I did go through the code and made some remarks. A more thorough and nitpicking review will come later... I have no immediate idea of why you get changed results, this should be looked at carefully.

jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
jakobtorben: Fixed.

opm/simulators/wells/BlackoilWellModel_impl.hpp
jakobtorben: Made all references to the well cells, except the one for NLDD, where it is a reference to the pre-calculated values, that is updated at each timestep.

opm/simulators/linalg/WellOperators.hpp
```diff
@@ -123,10 +123,46 @@ class WellModelAsLinearOperator : public LinearOperatorExtra<X, Y>
...
     const WellModel& wellMod_;
 };
 
+
+template <class WellModel, class X, class Y>
+class DomainWellModelAsLinearOperator : public WellModelAsLinearOperator<WellModel, X, Y>
+{
+public:
+    using Base = WellModelAsLinearOperator<WellModel, X, Y>;
+    using Base::Base;
```
atgeirr, line 136: This seems both unnecessary and confusing.

jakobtorben:
The most recent version has been tested in serial and parallel with all variants of the following options:
nonlinear solver = [newton, nldd]
preconditioners = [ilu0, cpr_trueimpes, cprw]
matrix-add-well-contributions = [false, true]

On SPE1 and SPE9, the number of timesteps, Newton iterations and linear iterations matches exactly with master.

Here are the complete result on Norne, for all variants, comparing this PR to master (with the extra command `--local-well-solve-control-switching=false` to address some current issues with Norne).

Norne_comparison


On Norne, the results, residuals, matrices and RHS matches exactly (the matrix and RHS has been checked with a precision of 18 digits) up until report step 3. After this they start to slightly differ, initially with a difference of 1e-10, which gradually increases. At report step 3, a new well is added, where the perforated cells are not monotonically increasing with the cell number. As a result, with this new fix, the matrix-vector reduction operations will be done in a different order, which for numbers of fixed precision will give a small difference. Everything still works, we just get slightly different simulation behaviour.

To verify that, the difference we observed was only caused by the different order, I did a test locally where I ensured that the well operators used the exact same order in the additions in the matrix-vector operations. I then got exactly the same number of timesteps and iterations for Norne with ILU0 and cpr_trueimpes.

ILU0
image

cpr_trueimpes
image

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1815,6 +1893,25 @@ namespace Opm {
...
 
+    template<typename TypeTag>
+    void
+    BlackoilWellModel<TypeTag>::
+    addWellPressureEquationsDomain(PressureMatrix& jacobian, const BVector& weights,const bool use_well_weights, int domainIndex) const
+    {
+        int nw =  this->numLocalWellsEnd();
+        int rdofs = local_num_cells_;
+        for ( int i = 0; i < nw; i++ ){
+            int wdof = rdofs + i;
```
atgeirr, line 1904: I am not sure this is correct. The `rdofs` should be the size of the domain, and the `nw` should be the number of wells in the domain. Maybe better to drop this for now, and instead disallow CPRW as linear solver for subdomain nldd solves,  I assume that is the only use case for this function?

atgeirr (commented):
This is good, only minor issues except for the one about CPRW as local solver.

opm/simulators/linalg/WellOperators.hpp
jakobtorben: I agree that it seems confusing. The `using Base::Base;` is necessary for inheriting constructors from base class.
Open for other suggestions.

opm/simulators/wells/BlackoilWellModel_impl.hpp
jakobtorben: That is probably correct. I have not yet looked into making CPRW and CPR trueimpes work as local linear solvers yet. This is something that did not work before any of the changes and would result in a crash. This is still the case. 
The changes added in this PR should be a step towards supporting these preconditioners for local linear solvers.
If I have time before this is taken in, I will fix it or else I will do it in a follow up PR.

jakobtorben:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
jakobtorben: I commented out the code, left some comments about what needs to be fixed, and throw an error if this functions is attempted to be used. I will see if I can implement this in a later PR.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1815,6 +1893,25 @@ namespace Opm {
         }
     }
 
+    template<typename TypeTag>
+    void
+    BlackoilWellModel<TypeTag>::
+    addWellPressureEquationsDomain(PressureMatrix& jacobian, const BVector& weights,const bool use_well_weights, int domainIndex) const
+    {
+        int nw =  this->numLocalWellsEnd();
```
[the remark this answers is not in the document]
jakobtorben: This was copied from the original function. Fixed both now.

opm/simulators/wells/BlackoilWellModel.hpp
```diff
@@ -584,6 +593,11 @@ template<class Scalar> class WellContributions;
 
         private:
             BlackoilWellModel(Simulator& simulator, const PhaseUsage& pu);
+
+            mutable BVector x_local_;
```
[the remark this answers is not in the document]
jakobtorben: Fixed.

opm/simulators/wells/BlackoilWellModel.hpp
```diff
@@ -297,12 +299,11 @@ template<class Scalar> class WellContributions;
                 return this->computeWellBlockAveragePressures();
             }
 
-            // subtract Binv(D)rw from r;
-            void apply( BVector& r) const;
-
             // subtract B*inv(D)*C * x from A*x
             void apply(const BVector& x, BVector& Ax) const;
 
+            void applyDomain(const BVector& x, BVector& Ax, int domainIndex) const;
```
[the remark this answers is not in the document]
jakobtorben: Fixed.

opm/simulators/linalg/ISTLSolver.hpp
```diff
@@ -450,6 +452,11 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
 
         const CommunicationType* comm() const { return comm_.get(); }
 
+        void setDomainIndex(int index)
```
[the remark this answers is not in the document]
jakobtorben: Fixed.

opm/simulators/linalg/WellOperators.hpp
jakobtorben: Changed the naming and added a comment for more clarity.

atgeirr:
All green and in it goes! Merging.

bska:
> All green and in it goes! Merging.

Quick heads-up: It is **possible** that this PR causes a regression failure in the sequential run of [examples/powerinjection_forchheimer_fd.cpp](https://github.com/OPM/opm-simulators/blob/064a66e794ed72761f1516325f9c592ffb53ee1a/examples/powerinjection_forchheimer_fd.cpp).  For what it's worth, there's a post-commit [regression failure](https://ci.opm-project.org/job/opm-simulators/2401/testReport/junit/(root)/serial/powerinjection_forchheimer_fd/) for this combination on the CI system following this PR.  We'll know more once the post-commit build of #5659 (i.e., the rebased version of #5658) completes.

atgeirr:
I see the failure for the opm-simulators post-builder: https://ci.opm-project.org/job/opm-simulators/
failing for run 2401. I do not understand why! But run 2402 is back to being good, so it was probably a fluke?

PR #5348, opened 2024-05-14
Omit deprecated declaration warning

aritorto:


akva2:
excellent. but maybe you can change it to use the type from ParallelCommunication.hpp instead ?

aritorto:
Thanks @akva2. It seems that I was as bit slow and got solved : )

PR #5349, opened 2024-05-14
Sphinx gha

lisajulia:
Based on #5243, which should be merged first.
Furthermore includes:
- Add documentation for the Python bindings from opm-common
- Add documentation for opm_embedded


lisajulia:
jenkins buil this opm-common=4018 please

lisajulia:
I'm currently still working on these two issues:

lisajulia:
> I'm currently still working on these two issues:
> 
>     * The way of making the docstrings_common.json from opm-common available here is not ideal yet, since this file is copied around a lot:
>       https://github.com/OPM/opm-common/pull/4018/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR477
>       https://github.com/OPM/opm-simulators/pull/5349/files#diff-3764531759f1b50aee06cf2267f2ab7508c016a86d776664e4754694fd711bf9R24
> 
>     * I'm still working on the documentation structure, since now all methods of all classes from opm-common are listed under "common" but actually they should be grouped

For the first problem:
- I've added a github action in opm-common (https://github.com/OPM/opm-common/pull/4018/files#diff-3ef24fe0222521a806b534108bf828ff317f06b784dd258bb28b0f5c8f1aa722R8) that pushes the current version of docstrings_common.json from opm-common to simulators, also not ideal.
- So at the moment, there are two mechanisms active for taking care of the docstrings_common.json (one withe cmake and one with the github action)
- Overall I'd like to disucss this some more

For the second problem: The methods are grouped now, but there is still room for improvement

python/simulators/CMakeLists.txt
```diff
@@ -20,6 +20,13 @@ add_custom_command(
     DEPENDS ${PYTHON_DOCSTRINGS_FILE}
     COMMENT "Generating PyBlackOilSimulatorDoc.hpp from JSON file"
 )
+
+find_file(DOCSTRINGS_JSON docstrings_common.json
+  PATHS ${opm-common_DIR} ${opm-common_PYTHON_COMMON_DIR}
+  PATH_SUFFIXES python NO_DEFAULT_PATH REQUIRED)
+
+configure_file(${DOCSTRINGS_JSON} ${PROJECT_SOURCE_DIR}/python/docstrings_common.json COPYONLY)
```
hakonhagland, line 28: I guess this copy to the project source directory (instead of to the binary directory) is not ideal. And since the github action runs separately from the cmake build, this file would not appear there anyway. It looks like you added a GitHub action in https://github.com/OPM/opm-common/pull/4018 to transfer the file directly instead? It is an interesting idea, but it looks a little bit dangerous :) Maybe the GitHub action here in opm-simulators could instead checkout the docstring json file from opm-common and pass that to the sphinx build scripts?

lisajulia: Yes, the copy command would only be neccessary for building the documentation locally.

I've added a GitHub action to transfer the file directly from opm-common to here so a new build is triggered if that file in opm-common is changed. If I add a github action here, then a new build of the documentation would not be triggered if that file in opm-common is changed, or is that possible somehow?

hakonhagland: Yes good point, I did not think of that we also need to trigger the action in opm-simulators immediately such that the docs there are in sync with the change of the file in opm-common... But then, instead of having the workflow in opm-common copy the file to opm-simulators to trigger the sphinx build, can we instead have it trigger the opm-simulators workflow directly? (Which will copy the file from opm-common and build the docs). See https://stackoverflow.com/q/70151645/2173773 for an example.

lisajulia: I've removed the copy command for now.

python/simulators/CMakeLists.txt
```diff
@@ -7,11 +7,16 @@ set(PYTHON_DOCSTRINGS_GENERATED_HPP "${PROJECT_BINARY_DIR}/python/PyBlackOilSimu
 #    ${PYTHON_EXECUTABLE} is set there to ${Python3_EXECUTABLE}
 #
 # Command to run the Python script
+
+find_file(PYTHON_GENERATE_DOCSTRINGS_PY generate_docstring_hpp.py
```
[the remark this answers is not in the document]
lisajulia: This should be fine I think since find_file is searching in both directories...

lisajulia:
Almost done, now I still need to change the Poetry build scripts to take the docstrings_common.json from here and not trigger a commit in opm-simulators.

Current state:
A workflow here is triggered when docstrings_common.json is changed in opm-common, the docstrings_common.json then gets updated here and the documentation is rebuilt.

python/simulators/CMakeLists.txt
lisajulia: I've added the docstrings_common.json to the .gitignore file now, now the github action here copies the file from the opm-common repository and the documentation is rebuilt also when the file in opm-common changes.
For local builds, this file is be copied from opm-common to here.

lisajulia:
jenkins build this please

lisajulia:
jenkins buil this opm-common=4018 please

lisajulia:
jenkins build this opm-common=4018 please

lisajulia:
jenkins build this opm-common=4018 please

lisajulia:
@hakonhagland and @blattms this is also ready for review, finally!

hakonhagland:
This looks good to me. Great that you managed to use GitHub actions in opm-common to trigger the sphinx build workflow in opm-simulators, hope this can be activated and tested in practice soon :smile:

hakonhagland:
> hope this can be activated and tested in practice soon

@atgeirr Can you have a look at https://github.com/OPM/opm-simulators/pull/5243 (which is a prerequisite for this PR)

lisajulia:
jenkins build this please

lisajulia:
For the build here to go through, github actions need read and write permissions:
Can someone check how the permissions are set currently?
Screenshot from 2024-07-02 11-06-51

PR #5351, opened 2024-05-14
Prevent spurious warnings about NNCs when applying EDITNNC in parallel.

blattms:
If we use transmissibilities for loadbalancing, then we calculate transmissibilities twice. First on the global grid before loadbalancing and then on the local grid after that. This is the default. In this case all warnings will be shown correctly when calculating the global transmissibilities.

If the user requests the same weights for all faces (command line parameter --edge-weights-method=0) then the transmissibilities are only calculated on the loadbalanced grid. Unfortunately, in this case only rank 0 will issue warnings for his part including the false positives mentioned below.

Due to load balancing many NNCs might be stored on another process, but we still use all EDITNNC entries when computing transmissibilties locally. Hence when applying EDITNNC on the loadbalanced grid we will issue warnings for cases where there are no problems (e.g. NNC between two overlap cells.

With this PR we will only warn when computing the transmissibilities for the first time. For the default settings this will remove spurious and duplicate warnings.

Not that for --edge-weights-method=0 nothing changes and we will still see only warnings for the first rank including spurious one.

### DIFF
--- a/opm/simulators/flow/Transmissibility.hpp
+++ b/opm/simulators/flow/Transmissibility.hpp
@@ -226,17 +226,18 @@ protected:
     void applyNncToGridTrans_(const std::unordered_map<std::size_t,int>& cartesianToCompressed);
 
     /// \brief Multiplies the grid transmissibilities according to EDITNNC.
-    void applyEditNncToGridTrans_(const std::unordered_map<std::size_t,int>& globalToLocal);
+    void applyEditNncToGridTrans_(const std::unordered_map<std::size_t,int>& globalToLocal, const bool warn);
 
     /// \brief Resets the grid transmissibilities according to EDITNNCR.
-    void applyEditNncrToGridTrans_(const std::unordered_map<std::size_t,int>& globalToLocal);
+    void applyEditNncrToGridTrans_(const std::unordered_map<std::size_t,int>& globalToLocal, const bool warn);
 
     void applyNncMultreg_(const std::unordered_map<std::size_t,int>& globalToLocal);
 
     void applyEditNncToGridTransHelper_(const std::unordered_map<std::size_t,int>& globalToLocal,
                                         const std::string& keyword, const std::vector<NNCdata>& nncs,
                                         const std::function<KeywordLocation(const NNCdata&)>& getLocation,
-                                        const std::function<void(Scalar&, const Scalar&)>& apply);
+                                        const std::function<void(Scalar&, const Scalar&)>& apply,
+                                        const bool warn);
 
     void extractPermeability_();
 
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please

opm/simulators/flow/Transmissibility_impl.hpp
```diff
@@ -1163,9 +1163,12 @@ applyEditNncToGridTransHelper_(const std::unordered_map<std::size_t,int>& global
         auto highIt = globalToLocal.find(c2);
 
         if (lowIt == globalToLocal.end() || highIt == globalToLocal.end()) {
-            print_warning(*nnc);
-            ++nnc;
-            warning_count++;
+            // Pevent warnings for NNCs stored on other processes in parallel (both cells inactive)
```
akva2, line 1166: Prevent

blattms:
This is deadlocking on the case. will investigate,

opm/simulators/flow/Transmissibility_impl.hpp
```diff
@@ -1163,9 +1163,12 @@ applyEditNncToGridTransHelper_(const std::unordered_map<std::size_t,int>& global
...
+            if ( lowIt != highIt) {
+                print_warning(*nnc);
+                ++nnc;
```
blattms, line 1169: :flushed: Increment should be outside!

blattms:
Mmh, not all of those messages are gone, which means that we somehow cut NNCs when partitioning.

Keeping this in draft.

blattms:
This is now redone and prevents spurious warning for the default settings (when we calculate global transmissibilities before load balancing because we pass them to the loadbalancer.

For the case  where we do not use transmissibilities nothing is changed and we will still see spurious warnings.

See the edited description for more details.

blattms:
jenkins build this please

opm/simulators/flow/Transmissibility.hpp
```diff
@@ -226,17 +226,18 @@ class Transmissibility {
     void applyNncToGridTrans_(const std::unordered_map<std::size_t,int>& cartesianToCompressed);
 
     /// \brief Multiplies the grid transmissibilities according to EDITNNC.
-    void applyEditNncToGridTrans_(const std::unordered_map<std::size_t,int>& globalToLocal);
+    void applyEditNncToGridTrans_(const std::unordered_map<std::size_t,int>& globalToLocal, const bool warn);
 
     /// \brief Resets the grid transmissibilities according to EDITNNCR.
-    void applyEditNncrToGridTrans_(const std::unordered_map<std::size_t,int>& globalToLocal);
+    void applyEditNncrToGridTrans_(const std::unordered_map<std::size_t,int>& globalToLocal, const bool warn);
```
bska, line 232: Is it ever conceivable to pass something other than `this->warnEditNNC_` as the `warn` argument to these member functions?  If not, maybe we don't need the additional parameter?

bska (commented):
Much appreciated.  My only question is about the use of a data member as an explicit parameter to a member function.

blattms:
jenkins build this please

bska (approved):
Thank you for the updates.  This looks good to me now and I'll merge into master.

PR #5352, opened 2024-05-14
Add ability to pass multiple variables to Damaris using DamarisWriter class.

jcbowden:
The full list of available variables is hard coded and defined in initDamarisXmlFile.cpp ; 

Also adds ability to limit the variables to be passed through using the
```
 --damaris-limit-variables=<CSV list> 
```
command line option.


opm/simulators/flow/DamarisWriter.hpp
```diff
@@ -154,6 +157,11 @@ class DamarisWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::G
              "This name should be unique if multiple simulations are running on "
              "the same node/server as it is used for the Damaris shmem name and by "
              "the Python Dask library to locate sections of variables.");
+        Parameters::registerParam<TypeTag, Properties::DamarisLimitVariables>
+            ("A comma separated list of variable names that a user wants to pass"
+             "through via DamarisOutput::DamarisWriter::writeOutput)() to the "
+             "damaris_write() call. This can be used to limit the number of"
```
atgeirr, line 163: Missing space after "of", and typo on next line "Daamis".

opm/simulators/flow/DamarisWriter.hpp
```diff
@@ -259,15 +269,81 @@ class DamarisWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::G
...
+                                                      this->elements_rank_offsets_[rank_]);
+            }
+
+            // Call damaris_write() for all available variables
+            for ( auto damVar : localCellData ) 
+            {
+               // std::map<std::string, data::CellData>
+              const std::string& name = damVar.first ;
+              
+              std::unordered_set<std::string>::const_iterator is_in_set = wanted_vars_set_.find ( name );
```
atgeirr, line 299: Since this variable is only used to determine the presence in the set, I suggest instead changing the `if` on line 301 to `if (wanted_vars_set_.count(name)) {...`

opm/simulators/flow/DamarisWriter.hpp
```diff
@@ -259,15 +269,81 @@ class DamarisWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::G
...
+            }*/
+            
+            
+            // Call damaris_set_position() for all available variables
+            // There is an assumption that all variables are the same size, with the same offset.
+            // see initDamarisTemplateXmlFile.cpp for the Damaris XML descriptions.
+            for ( auto damVar : localCellData ) {
+                // std::map<std::string, data::CellData>
+                const std::string name = damVar.first ;
+                dam_err_ = DamarisOutput::setPosition(name.c_str(), rank_,
```
atgeirr, line 289: Should this be called uncontitionally for all (possible) variables? Should we not do this inside the `if` below?

opm/simulators/flow/DamarisWriter.hpp
```diff
@@ -259,15 +269,81 @@ class DamarisWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::G
...
+               // std::map<std::string, data::CellData>
+              const std::string name = std::get<0>(damVar.first) ;
+              const int part = std::get<1>(damVar.first) ;
+              double  dataCol = damVar.second ;
+              std::cout << "Name of Damaris Block Varaiable : (" << rank_ << ")  "  << name  << "  part : " << part << "  Value : "  << dataCol <<  std::endl ;  
+            } 
+            
+            dam_err_ =  DamarisOutput::endIteration(rank_);
+            */
+            if (this->damarisOutputModule_->getPRESSURE_ptr() != nullptr)
```
atgeirr, line 342: I do not understand what the logic is here? Also it seems a bit non-generic?

opm/simulators/flow/DamarisWriter.hpp
```diff
@@ -259,15 +269,81 @@ class DamarisWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::G
...
+            // Call damaris_write() for all available variables
+            for ( auto damVar : localCellData ) 
+            {
+               // std::map<std::string, data::CellData>
+              const std::string& name = damVar.first ;
+              
+              std::unordered_set<std::string>::const_iterator is_in_set = wanted_vars_set_.find ( name );
+              
+              if ((is_in_set != wanted_vars_set_.end() ) || (wanted_vars_set_.size() == 0)) {
+                  data::CellData  dataCol = damVar.second ;
```
atgeirr, line 302: Is this copy necessary?

opm/simulators/flow/DamarisWriter.hpp
```diff
@@ -259,15 +269,81 @@ class DamarisWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::G
...
+                  OpmLog::debug(fmt::format("Name of Damaris Variable       : ( rank:{})  name: {}  ",  rank_, name));
+                  
+                  // It does not seem I can test for what type of data is present (double or int)
+                  // in the std::variant within the data::CellData, so I will use a try catch block. 
+                  // Although, only MPI_RANK and GLOBAL_CELL_INDEX are set as integer types (in the 
+                  // XML file) so it is a moot point. 
+                  // We could use damaris_get_type() to check what the type is specified as
+                  // within the Damaris XML file
```
atgeirr, line 310: I see your point, usually std::holds_alternative could be used, but as the variant itself is not exposed that is not possible here. Maybe we should add an observer method to the data::CellData class? Alternatively, should we treat MPI_RANK and GLOBAL_CELL_INDEX specially to avoid this try-catch?

atgeirr (commented):
I think this is a good way to dramatically improve the applicability of the Damaris layer. Some cleanup needed, and I have some questions, but this should get merged when finished.

opm/simulators/flow/DamarisWriter.hpp
```diff
@@ -259,15 +269,81 @@ class DamarisWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::G
...
                                                 this->damarisOutputModule_->getPRESSURE_ptr());
 
                 dam_err_ =  DamarisOutput::endIteration(rank_);
+            }*/
+            
+            
+            // Call damaris_set_position() for all available variables
+            // There is an assumption that all variables are the same size, with the same offset.
+            // see initDamarisTemplateXmlFile.cpp for the Damaris XML descriptions.
+            for ( auto damVar : localCellData ) {
```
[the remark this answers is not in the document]
jcbowden: Yes, these are all read only, so can be constant, and references. My poor form really!

opm/simulators/flow/DamarisWriter.hpp
jcbowden: Will there never be <int> values? If so, then using a variant seems redundant?

opm/simulators/flow/DamarisWriter.hpp
jcbowden: Here, again, it's my lack of use of references.

opm/simulators/flow/DamarisWriter.hpp
jcbowden: Me either. However, if this is not present, then the first iteration (iteration 0 in Damaris) is missed, and the GLOBAL_CELL_INDEX variable is not written. There is probably a better way to do this test, however, this works, and is derived from the original PRESSURE only code.

opm/simulators/flow/DamarisWriter.hpp
jcbowden: Probably only needs to be done once (unless the MPI partitioning gets changed)

opm/simulators/flow/DamarisWriter.hpp
jcbowden: Yes, I agree, you have a much nicer test.

jcbowden:
I am still trying to test the Python interface better. Then on to Paraview.

bska:
>  is there a way to obtain the output directory?

The `EclGenericWriter` base class has a protected `eclState_` data member of type `EclipseState`, so you can use a statement of the form
```C++
const auto& outputDir = this->eclState_.cfg().io().getOutputDir();
```
to get a read-only reference to a `std::string` containing the name of the output directory.  It should **probably** be a `std::filesystem::path` instead, but the `IOConfig` object is fairly old code by now and has not been made aware for the filesystem library.

jcbowden:
Great, thanks. I am going to add the output path as a variable to the Damaris XML configuration file, so that Python can output data in the correct place. The Python interface does not have an 'outputdir' tag in its XML configuration, unlike HDF5, which has a 'FilesPath' tag.

jcbowden:
Good afternoon, THe output directory information is available through the methods provided, however, when I am not using --output-dir="some/explicit/path/" then the output directory is defaulting to the directory of the input data. Is that the desired behavior?

atgeirr:
> Good afternoon @atgeirr and @bska. The output directory information is available through the methods provided above, however, when I am not using --output-dir="some/explicit/path/" then the output directory is defaulting to the directory of the input data. Is that the desired behavior?

Yes it is. This is what the engineers and their workflows assume, so it is also our default.

opm/simulators/utils/DamarisKeywords.hpp
```diff
@@ -104,6 +112,30 @@ getDamarisKeywords(const Parallel::Communication& comm, const std::string& Outpu
...
+    std::unordered_set<std::string> resuset ;
+    std::string tstr;
+    // The --damaris-limit-variables command line option (defaults to empty string)
+    std::string damarisLimitVars = Parameters::get<TypeTag, Properties::DamarisLimitVariables>();
+    std::stringstream ss(damarisLimitVars); 
+
+    // Use while loop to check the getline() function condition.  
+    while (std::getline(ss, tstr, ',')) {
+        //remove whitespace
+        // tstr.erase(std::remove_if(tstr.begin(), tstr.end(), ::isspace), tstr.end());
```
atgeirr, line 128: Why did you change this to not use ::isspace?
If keeping current version, delete the commented out line.

jcbowden:
Okay, I can see some utility in that. Anyway, I am almost ready to set this PR as "Ready for review" - I still need to go through some Paraview tests, and this will happen sometime soon.

opm/simulators/flow/DamarisWriter.hpp
```diff
@@ -307,10 +390,18 @@ class DamarisWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::G
         // We will add the MPI rank value directly into shared memory using the DamarisVar 
         // wrapper of the C based Damaris API.
         // The shared memory is given back to Damaris when the DamarisVarInt goes out of scope.
-        DamarisVarInt mpi_rank_var_test(1, {"n_elements_local"},  "MPI_RANK", rank_);
-        mpi_rank_var_test.setDamarisParameterAndShmem( {this->numElements_ } ) ;
+        // DamarisVarInt mpi_rank_var_test(1, {"n_elements_local"},  "MPI_RANK", rank_);
```
atgeirr, line 393: Remove commented-out code.

opm/simulators/flow/DamarisWriter.hpp
atgeirr: We can leave it like this for now.

opm/simulators/flow/DamarisWriter.hpp
atgeirr: I do not understand the comment.

I suggest the following changes. First on line 297, add:
```
const bool data_is_integer = (name == "MPI_RANK" || name == "GLOBAL_CELL_INDEX");
```
Then instead of the try, use the `data_is_integer` variable to select which code block to run.

opm/simulators/flow/DamarisWriter.cpp
```diff
@@ -42,7 +42,8 @@ int setPosition(const char* field, int rank, int64_t pos)
 {
     int dam_err = damaris_set_position(field, &pos);
     if (dam_err != DAMARIS_OK) {
-        OpmLog::error(fmt::format("damariswriter::writeOutput()       : ( rank:{})"
+        // OpmLog::error only prints from rank 0
+        OpmLog::error(fmt::format("damariswriter::setPosition()       : ( rank:{}) "
```
atgeirr, line 46: I think we must use the DeferredLogger class here to ensure we get all error log messages. Then use gatherDeferredLogger() to gather them to rank 0 once we are in a context where every process will be. Tell me if you want some guidance on this.

Also: The `error` category should only be for things that stop the simulator. If this is not necessary (and I do not think it is, no exception is thrown and the process will keep running), use the `problem` (for numerical issues) or `warning` categories instead. Probably `warning` is the best here.

atgeirr (commented):
Did a slightly more thorough review, and things are improving. Noticed some things that should be addressed, and think this can be merged when done.

opm/simulators/utils/DamarisKeywords.hpp
jcbowden: It did not work/compile - I forget the issue exactly.

opm/simulators/flow/DamarisWriter.cpp
jcbowden: The test:
```
const bool data_is_integer = (name == "MPI_RANK" || name == "GLOBAL_CELL_INDEX");
```
 will not allow a future developer/user to add a variable that is of integer type. I assume that because a std::variant is used, that integer data vectors are possible.

jcbowden: hanged to warning. removed comments about only rank 0 getting the error/warning message - this still seems an issue as other ranks can have an error and the error message will be hidden.

opm/simulators/flow/DamarisWriter.hpp
jcbowden: do you mean this comment: 
"Will there never be values? If so, then using a variant seems redundant?" or another one?
I think it may read better as:
 "Will there never be integer values? If so, then using a variant seems redundant?"

jcbowden:
Setting this as ready for review - I am still testing the Damaris Paraview plugin, however this does not affect the opm-simulators.

bska:
jenkins build this please

jcbowden:
Hi, okay, so what is the issue? Just the compile time warnings?

akva2:
if you are asking why the build is red, then yes that is due to the warnings.

jcbowden:
Okay, i'll fix them. I have another update that removes the define compile time guards in DamarisKeywords.cpp. The guards affect the flexibility of changing the Damaris library capabilities without needing to re-compile opm-simulators. I will push this change with the warnings issues fix.

jcbowden:
This does not have the int vs long int compiler issue fixed... give me a moment to think about it.

bska:
jenkins build this please

jcbowden:
A requirement is use of Damaris v.11.0 or higher as the HDF5 and Paraview plugins have been modified to work on iterations where no data is written to a particular variable (this seems common in flow). The Python plugin already tested for this, so earlier versions of Damaris should be okay if only using the Damaris Python output.

opm/simulators/flow/DamarisWriter.cpp
```diff
@@ -42,7 +42,7 @@ int setPosition(const char* field, int rank, int64_t pos)
 {
     int dam_err = damaris_set_position(field, &pos);
     if (dam_err != DAMARIS_OK) {
-        OpmLog::error(fmt::format("damariswriter::writeOutput()       : ( rank:{})"
+        OpmLog::warning(fmt::format("damariswriter::setPosition()       : ( rank:{}) "
```
atgeirr, line 45: This will all be lost if rank is not 0. I can fix this in a followup, though.

atgeirr:
Build is green, merging! Thanks for the effort.

jcbowden:
Great! I had the the Norne case running with live Paraview output - the update now allows one to select from the (many) variables and have multiple variables all being viewed in 3D simultaneously. It worked quite well.

PR #5353, opened 2024-05-15
VFP classes: template Scalar type

akva2:
Another small piece of https://github.com/OPM/opm-simulators/pull/5286

### DIFF
--- a/opm/simulators/wells/VFPProdProperties.hpp
+++ b/opm/simulators/wells/VFPProdProperties.hpp
@@ -34,9 +34,9 @@ class VFPProdTable;
  * water fraction, gas fraction, and artificial lift for production VFP tables, and similarly
  * the BHP as a function of the rate and tubing head pressure.
  */
+template<class Scalar>
 class VFPProdProperties {
 public:
-    VFPProdProperties() = default;
     /**
      * Takes *no* ownership of data.
      */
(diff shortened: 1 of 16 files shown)


akva2:
jenkins build this please

opm/simulators/wells/VFPProdProperties.hpp
```diff
@@ -64,10 +64,10 @@ class VFPProdProperties {
...
                  const EvalWell& liquid,
                  const EvalWell& vapour,
-                 const double& thp,
-                 const double& alq,
-                 const double& explicit_wfr,
-                 const double& explicit_gfr,
+                 const Scalar& thp,
+                 const Scalar& alq,
+                 const Scalar& explicit_wfr,
+                 const Scalar& explicit_gfr,
```
bska, line 70: It didn't make a lot of sense before, but perhaps it's a little more useful now.  I'll nevertheless ask why we're passing these parameters as references instead of values here.

bska (commented):
This looks good to me and appears to have been a mostly automatic substitution from `double` to `Scalar`.  I noticed a couple of places that pass parameters as `const Scalar&` (i.e., as "reference to `const`") and **maybe** we should switch those to passing the `Scalar` by value instead.  I guess it depends on what we expect `Scalar` to be in the typical case.  Built-in arithmetic types like `float` or `double` should probably be passed by value, but some class type object could conceivably be passed more efficiently as a reference.  Any insight here?

opm/simulators/wells/VFPProdProperties.hpp
akva2: Digging in history it seems they were made const-refs in context of old ADB code, so it should be safe to change to values, hence done.

bska (approved):

PR #5355, opened 2024-05-15
Resolve warning delete-abstract-non-virtual-dtor

multitalentloes:
These lines makes sure that the derived preconditionerMaker class has a destructor to avoid the clang warning -

> Wdelete-abstract-non-virtual-dtor

### DIFF
--- a/opm/simulators/linalg/PreconditionerWithUpdate.hpp
+++ b/opm/simulators/linalg/PreconditionerWithUpdate.hpp
@@ -91,6 +91,7 @@ struct GeneralPreconditionerMaker {
     virtual std::unique_ptr<
         Preconditioner<typename OriginalPreconditioner::domain_type, typename OriginalPreconditioner::range_type>>
     make() = 0;
+    virtual ~GeneralPreconditionerMaker(){};
 };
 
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

opm/simulators/linalg/PreconditionerWithUpdate.hpp
```diff
@@ -91,6 +91,7 @@ struct GeneralPreconditionerMaker {
     virtual std::unique_ptr<
         Preconditioner<typename OriginalPreconditioner::domain_type, typename OriginalPreconditioner::range_type>>
     make() = 0;
+    virtual ~GeneralPreconditionerMaker(){};
```
bska, line 94: Tiny nit: The final semicolon here, on the closing brace, will emit diagnostics of the form
> ```
> Redundant ';' ...
> ```
on some compilers, at least with certain warning settings.  In the overall interest of removing warnings maybe we could drop that trailing semicolon too?

akva2: ay. missed that :/

PR #5356, opened 2024-05-15
InitStateEquil: remove template parameter from Comm

akva2:
use Parallel::Communication directly

in the process this quells a deprecation warning for Dune::CollectiveCommunication usage.

replace #5348 

Also takes care of niggle https://github.com/OPM/opm-simulators/pull/5355#discussion_r1601241011


akva2:
jenkins build this please

PR #5363, opened 2024-05-16
fixed: build with dune-fem 2.9

akva2:
Interface has changed in 2.9 and including compat headers causes an error to be emitted.


bska (approved):
I don't see any obvious problems with this, but on the other hand I'm no dune-fem user either.  If it turns out we need adjustments here to handle some edge case, then I guess we'll have to make those at a later time.

For now I'll merge into master.

PR #5368, opened 2024-05-21
fix typo that would cause crash cuda aware openmpi

multitalentloes:
Strange and inconsistent capitalization would cause problems if compiled using cuda aware mpi.
Changed
> mpiSUpportsCudaAwareAtCompileTime
to 
> mpiSupportsCudaAwareAtCompileTime


multitalentloes:
Jenkins build this please

blattms (approved):

PR #5374, opened 2024-05-22
PAvgCalculator: template Scalar type

akva2:
Downstream of https://github.com/OPM/opm-common/pull/4073


bska (approved):

PR #5375, opened 2024-05-22
MultisegmentWell: use Scalar type

akva2:


bska (approved):
If we're really serious about this kind of work then we might consider additionally making all numeric literals be type `Scalar` too.  As of now we have a mix of
```C++
Scalar x = 1.2e3
```
and
```C++
Scalar x = Scalar{1.2e3} // or static_cast<Scalar>(...)
```
The former will implicitly convert if `Scalar` is not `double` whereas the latter will explicitly convert in that case.

For now, though, I'm happy that we're moving in the right direction so I'll merge this into master as is.

akva2:
yeah. i've only taken care of those that cause compilation issues when instancing for float..

PR #5378, opened 2024-05-23
amgclSolverBackend: fix for block_size == 1

akva2:
amgcl cannot use the block matrices for block_size == 1 with the built-in backend so we need to specialize the CPU code.

While the BDA solver currently only supports block size 3, and thus we never hit this code path, it created problems for me while adding support for floats because I instance the whole class and not just the constructor. If the restriction of bs=3 is lifted at some time, we are at least ready in this aspect now.


akva2:
jenkins build this please

akva2:
jenkins build this please

bska (approved):

PR #5380, opened 2024-05-23
LinAlg classes: template Scalar type

akva2:

### DIFF
--- a/opm/simulators/linalg/bda/cuda/cuWellContributions.cu
+++ b/opm/simulators/linalg/bda/cuda/cuWellContributions.cu
@@ -33,18 +33,17 @@ namespace Opm
 {
 
 // apply WellContributions using y -= C^T * (D^-1 * (B * x))
-__global__ void apply_well_contributions(
-    const double * __restrict__ Cnnzs,
-    const double * __restrict__ Dnnzs,
-    const double * __restrict__ Bnnzs,
-    const int * __restrict__ Ccols,
-    const int * __restrict__ Bcols,
-    const double * __restrict__ x,
-    double * __restrict__ y,
-    const int dim,
-    const int dim_wells,
-    const unsigned int * __restrict__ val_pointers
-)
+template<class Scalar>
+__global__ void apply_well_contributions(const Scalar* __restrict__ Cnnzs,
+                                         const Scalar* __restrict__ Dnnzs,
+                                         const Scalar* __restrict__ Bnnzs,
+                                         const int* __restrict__ Ccols,
+                                         const int* __restrict__ Bcols,
+                                         const Scalar* __restrict__ x,
+                                         Scalar* __restrict__ y,
+                                         const int dim,
+                                         const int dim_wells,
+                                         const unsigned int * __restrict__ val_pointers)
 {
     const int idx_b = blockIdx.x;
     const int idx_t = threadIdx.x;
@@ -57,9 +56,9 @@ __global__ void apply_well_contributions(
     const int c = lane % dim;                           // col in block
     const int r = (lane / dim) % dim_wells;             // row in block
 
-    extern __shared__ double smem[];
-    double * __restrict__ z1 = smem;
-    double * __restrict__ z2 = z1 + dim_wells;
+    extern __shared__ unsigned char smem[];
+    Scalar* __restrict__ z1 = reinterpret_cast<Scalar*>(smem);
+    Scalar* __restrict__ z2 = z1 + dim_wells;
 
     if (idx_t < dim_wells) {
         z1[idx_t] = 0.0;
@@ -70,7 +69,7 @@ __global__ void apply_well_contributions(
     // z1 = B * x
     if (idx_t < num_active_threads) {
         // multiply all blocks with x
-        double temp = 0.0;
+        Scalar temp = 0.0;
         int b = idx_t / vals_per_block + val_pointers[idx_b];       // block id, val_size indicates number of blocks
         while (b < val_size + val_pointers[idx_b]) {
             int colIdx = Bcols[b];
@@ -106,7 +105,7 @@ __global__ void apply_well_contributions(
 
     // z2 = D^-1 * B * x = D^-1 * z1
     if (idx_t < dim_wells) {
-        double temp = 0.0;
+        Scalar temp = 0.0;
         for (int c = 0; c < dim_wells; ++c) {
             temp += Dnnzs[idx_b * dim_wells * dim_wells + idx_t * dim_wells + c] * z1[c];
         }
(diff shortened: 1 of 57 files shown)


akva2:
jenkins build this rocm hipify serial please

opm/simulators/linalg/bda/cuda/cuWellContributions.cu
```diff
@@ -57,9 +56,9 @@ __global__ void apply_well_contributions(
     const int c = lane % dim;                           // col in block
     const int r = (lane / dim) % dim_wells;             // row in block
 
-    extern __shared__ double smem[];
-    double * __restrict__ z1 = smem;
-    double * __restrict__ z2 = z1 + dim_wells;
+    extern __shared__ unsigned char smem[];
+    Scalar* __restrict__ z1 = reinterpret_cast<Scalar*>(smem);
+    Scalar* __restrict__ z2 = z1 + dim_wells;
```
bska, line 61: I'm demonstrating my lack of CUDA knowledge here, but what entity is responsible for acquiring and releasing the resource provided by `smem` in this case?

bska (commented):
With the exception of a `reinterpret_cast<>` which piqued my interest I don't really have a lot of comments here.  Maybe someone with more in-depth CUDA/accelerator knowledge would like to have a look.  Perhaps @kjetilly, @jakobtorben, or @multitalentloes?

opm/simulators/linalg/bda/cuda/cuWellContributions.cu
akva2: smem is a  static runtime (language) provided pointer to a memory area shared between cpu and accelerator. For .. reasons you have yo ask the leather jacket .. this is exposed as an unsigned char pointer and not void* as it ought to be hence the nasty casts.

multitalentloes: I think shared memory is not shared between cpu and gpu, that would be unified memory using cuda terminology. Shared memory in this context is statically allocated memory taken from the L1 cache that is visible for all warps/threads running in the same thread block.

akva2: i stand corrected on terminology and for not remembering what __shared__ indicates correctly. but in any case this is not user managed memory.

bska: > but in any case this is not user managed memory.

Thanks for clearing that up, both of you.  Greatly appreciated!  I assume the memory is properly aligned to access `Scalar` objects too.

On a tangential note, if this were regular C++ code you would typically "placement-new" the individual `Scalar` objects here, since&ndash;according to the C++ object model&ndash;there is no `Scalar` object before its constructor runs.  This holds even for built-in arithmetic types like `float` or `double`.  In that case we'd write something along the lines of
```C++
for (int w = 0; w < dim_wells; ++w) {
    new (z1 + w) Scalar{};
    new (z2 + w) Scalar{};
}
```
and we'd need similar logic at the end of this block to call the destructors
```C++
for (int w = dim_wells; w > 0; --w) {
    z2[w - 1].~Scalar();
    z1[w - 1].~Scalar();
}
```
but maybe the rules are different in `.cu` files.

kjetilly: The `__shared__` memory is a bit special. No allocation or deallocation is needed, as this is done by the CUDA system upon kernel launch/exit. Furthermore, in this case, we are using the `extern` keyword, meaning the size of the shared memory will be determined upon kernel launch (as opposed to being statically allocated with eg `__shared__ unsigned char  smem[64 * 42];`). 

Note that the actual size of the shared memory is given upon the launch of the kernel, done here:

```C++
        int smem_size = 2 * sizeof(Scalar) * this->dim_wells;
        apply_well_contributions <<< this->num_std_wells, 32, smem_size, stream>>>(d_Cnnzs,
                                                                                   d_Dnnzs,
                                                                                   d_Bnnzs,
                                                                                   d_Ccols,
                                                                                   d_Bcols,
                                                                                   d_x,
                                                                                   d_y,
                                                                                   this->dim,
                                                                                   this->dim_wells,
                                                                                   this->d_val_pointers);
```

where `smem_size`, given in bytes, is supplied to the kernel as the second last argument in the `<<<>>>` brackets.

bska: > No allocation or deallocation is needed, as this is done by the CUDA system upon kernel launch/exit.

Okay, understood.  The issue of "no object (lifetime) exists in C++ until the constructor runs" is nevertheless a problem, however hypothetical, even when we don't have to acquire any explicit memory resources for it.  My [earlier](https://github.com/OPM/opm-simulators/pull/5380#discussion_r1613073865) loop was obviously wrong because that's vulnerable to all kinds of race conditions but we might consider adding something along the lines of
```C++
new (z1 + idx_t) Scalar{};
new (z2 + idx_t) Scalar{};
```
where we're currently [setting values to zero](https://github.com/OPM/opm-simulators/blob/9dfe949e680f68e3a4d704746e542a8f0a367717/opm/simulators/linalg/bda/cuda/cuWellContributions.cu#L64) and similar calls to the destructors at the end of the kernel.

akva2:
jenkins build this rocm hipify serial please

bska:
Are there any other comments here @kjetilly, @multitalentloes?  If not, I suggest merging this as-si.

opm/simulators/linalg/bda/cuda/cuWellContributions.cu
kjetilly: I had to check that placement-new was actually supported by CUDA, and it seems it is, and calling destructors as outlined is also possible. 

Now, this is not a part of the CUDA-OPM path I've looked at in depth (written by BDA), but I think I would make sense add either a `static_assert` or `std::enable_if` on `std::is_fundamental_v<Scalar>` to avoid the whole unpleasantness. This code will anyway fail to compile on anything other than a subset of the fundamental types since it uses [__shfl_down_sync](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html?highlight=__shfl_down_sync#warp-shuffle-synopsis) on `Scalar`.

multitalentloes:
Looks good to me

opm/simulators/linalg/bda/rocsparseSolverBackend.cpp
```diff
@@ -390,66 +395,85 @@ void rocsparseSolverBackend<block_size>::initialize(std::shared_ptr<BlockedMatri
...
+                             hipMemcpyHostToDevice, stream));
+    HIP_CHECK(hipMemcpyAsync(d_Acols, mat->colIndices, 
+                             sizeof(rocsparse_int) * nnzb,
+                             hipMemcpyHostToDevice, stream));
+    HIP_CHECK(hipMemcpyAsync(d_Avals, mat->nnzValues,
+                             sizeof(Scalar) * nnz,
+                             hipMemcpyHostToDevice, stream));
+    HIP_CHECK(hipMemsetAsync(d_x, 0, N * sizeof(Scalar), stream));
+    HIP_CHECK(hipMemcpyAsync(d_b, b, N * sizeof(Scalar) * N,
+                             hipMemcpyHostToDevice, stream));
```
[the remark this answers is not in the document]
razvnane: Ah, it seems I have trouble spotting obvious typos :(... thanks for the fix :).

PR #5382, opened 2024-05-24
compositional

GitPaean:


GitPaean:
jenkins build this opm-common=4045 opm-models=901 please

GitPaean:
jenkins build this opm-common=4045 opm-models=901 please

bska:
> opm-common=4045 opm-models=901

For future reference: We usually prefer to trigger multi-repository builds from the most upstream repository&ndash;in this case from opm-common.  The reason is that the CI system, as a cost saving measure, will bypass unit tests in upstream repositories when the build is triggered from a downstream repository.  This is **generally** not what we want although there may of course be special concerns in any one individual situation.

GitPaean:
> For future reference: We usually prefer to trigger multi-repository builds from the most upstream repository–in this case from opm-common. The reason is that the CI system, as a cost saving measure, will bypass unit tests in upstream repositories when the build is triggered from a downstream repository. This is **generally** not what we want although there may of course be special concerns in any one individual situation.

Yes. I was just confirming the compilation error related to `flowexp_comp`.

GitPaean:
Okay. The failure is related to `DamarisWriterType`. Local compilation does not have `HAVE_DAMARIS` enabled.

GitPaean:
jenkins build this opm-common=4045 opm-models=901 please

GitPaean:
PR in opm-common is tested alone.  Triggering the test from here to reproduce the running regression failures.

GitPaean:
I will deal with the upstream PRs in opm-common individually. So the jenkins is triggered here only for this PR. (opm-model PR is tiny and will likely b revised)

GitPaean:
jenkins build this opm-common=4045 opm-models=901 please

GitPaean:
jenkins build this opm-common=4045 opm-models=901 please

PR #5395, opened 2024-05-29
Small fixes realted to thp-control/vfp-extrapolation

steink:
This PR contains three small fixes related to thp vfp/control:
1. When computing thp(bhp), make sure we're using the same explicit/implicit fraction option as when computing bhp(thp). This fixes some corner-cases where inconsistency lead to no feasible control.
2. When computing thp(bhp) with extreme vfp-extrapolation there may be any number of solutions. This fix will always return the solution corresponding to the highest/lowest (producers/injectors) thp in attempt to minimize the risk of switcing control to a garbage thp-value (that is, return the garbage that is least likely to cause damage). 
3. Disallow local well controll switching when the _upper level_ `-maximum-number-of-well-switches` has reached it's limit. This improves the overall logic and prevents some problems related to _competing_ controls.


opm/simulators/wells/WellBhpThpCalculator.cpp
```diff
@@ -143,11 +143,14 @@ calculateThpFromBhp(const std::vector<Scalar>& rates,
     else if (well_.isProducer()) {
         const Scalar vfp_ref_depth = well_.vfpProperties()->getProd()->getTable(table_id).getDatumDepth();
         const Scalar dp = wellhelpers::computeHydrostaticCorrection(well_.refDepth(), vfp_ref_depth, rho, well_.gravity());
+        const bool use_vfpexp = well_.useVfpExplicit();
+        const double wfr = well_.vfpProperties()->getExplicitWFR(table_id, well_.indexOfWell());
```
akva2, line 147: Scalar

opm/simulators/wells/WellBhpThpCalculator.cpp
```diff
@@ -143,11 +143,14 @@ calculateThpFromBhp(const std::vector<Scalar>& rates,
...
+        const double gfr = well_.vfpProperties()->getExplicitGFR(table_id, well_.indexOfWell());
```
akva2, line 148: Scalar

steink:
I made an error here, halt any reviewing until fixed ...

steink:
> I made an error here, halt any reviewing until fixed ...

Now hopefully fixed

atgeirr (approved):
This looks good, time to test it!

atgeirr:
jenkins build this please

steink:
Sorry, forgot to build everything, so missed the tests in `test_vfpproperties`. Fixed/updated now.

atgeirr:
jenkins build this please

steink:
I've looked at the failures and they all are related to output-values for THP (while wells are controlled by something else). There are three tests where these values differ significantly.

For tests **0B_WELPI_MODEL2** and **ACTIONX_WPIMULT** this happens during a period where PROD3 is oscillating greatly. Here there is hardly any water production so there is (problematic) extrapolation in wcut. I think the PR-extrapolation seem more reasonnable in this case, so I think this is fine (see pic).


For test **ACTIONX_GCONPROD**, the output THP (well OP02) for the PR makes a huge jump for one of the steps (see pic). Was a bit difficult to get to the bottom of this, but the well is using excplicit fractions for vfp-interpolation here (hence the difference), and it seems that after an action is triggered, the well is not initialized with rates from the previous well-state, so in the vfp-interpolation, fractions based on _defaulted_ rates are used (this is not seen in master since it always uses implicit fractions for `thp(bhp)` regardless). Will probably need to look more closely into this. Nevertheless, this potential problem is not related to this PR (it's revealed by the PR), so I don't think it should stop it from getting in.

atgeirr:
Thanks for your analysis! I agree with your assessment, and we should do a data update and merge here. @totto82 do you agree?

atgeirr:

totto82:
do you agree?> Will probably need to look more closely into this

I agre

bska:
jenkins build this opm-tests=1173 please

tskille:
This PR has now been testen on a Full field model in Equinor. Results are excellent and I'm very looking forward to see this go into master.

GitPaean:
jenkins build this opm-tests=1173 please

bska:
> The reference has been updated.

Not really, no, and the post-commit build is (probably) going to fail.  The CI system uses a shared copy of the reference solutions for all builds and that shared copy is updated only when the "opm-tests" post-commit build **runs**.  That, in turn, means that it's usually not safe to merge PRs which bring new reference solutions until you've verified that the corresponding update to the reference solutions has fully completed.  We can **often** get away with merging these at the same time, but when there is a high load on the CI system we're more vulnerable to race conditions.

bska:
> > The CI system uses a shared copy of the reference solutions for all builds and that shared copy is updated only when the "opm-tests" post-commit build **runs**. That, in turn, means that it's usually not safe to merge PRs which bring new reference solutions until you've verified that the corresponding update to the reference solutions has fully completed.
> 
> It is complicated. What is the good practice should be here?

The best practice is to not merge code-changing PRs until you've **verified** that the opm-tests post-commit build job has completed.  If the CI system load is high that often means waiting until the load subsides before merging the new reference solutions as well since that's the only safe way of reducing the window where the reference solutions are not consistent with the live development master branches.

GitPaean:
> The best practice is to not merge code-changing PRs until you've verified that the opm-tests post-commit build job has completed. If the CI system load is high that often means waiting until the load subsides before merging the new reference solutions as well since that's the only safe way of reducing the window where the reference solutions are not consistent with the live development master branches.

Thanks for the explanation. Where could I check the opm-tests post-commit build?

GitPaean:
Are you referring to https://ci.opm-project.org/view/Post%20builders/job/opm-tests/778/ . 

Sorry I seldom check here and I did not see complicated issues related to this before either. I do not handle this as often as many others.

GitPaean:
Now I see this one failed. https://ci.opm-project.org/job/opm-simulators/2216/ 

It is because opm-tests grabbed the old reference with commit a6f4ab346856e693b862b9fc3b5056cfb128f4eb . 

Sorry for the trouble, I will pay attention to this in the future.

PR #5396, opened 2024-05-29
Add Regression Test for Undefined UDQ Variables

bska:
Undefined UDQs get the value in `UDQPARAM(3)`.  The UDQ-01 test model triggers an action block [if and only if](https://github.com/OPM/opm-tests/blob/1925c493cc8e4a4bb905c5bfc09da731391c3123/actionx/UDQ-01.DATA#L206) we implement that behaviour.


bska:

bska:
jenkins build this opm-tests=1169 please

bska:
The new reference solutions have been installed on the CI system.  Merging this now to activate the regression test for undefined UDQs.

PR #5399, opened 2024-05-31
[FIX]: Build issue if OpenMP is not present

daavid00:
Fix to the error:
```
/Users/dmar/Github/opm/opm-simulators/opm/simulators/linalg/ISTLSolverBda.cpp:127:15: error: use of undeclared identifier 'copyThread'
              copyThread = std::make_shared<std::thread>([&](){this->copyMatToBlockJac(matrix, *blockJacobiForGPUILU0_);});
              ^
/Users/dmar/Github/opm/opm-simulators/opm/simulators/linalg/ISTLSolverBda.cpp:127:50: error: no member named 'thread' in namespace 'std'
              copyThread = std::make_shared<std::thread>([&](){this->copyMatToBlockJac(matrix, *blockJacobiForGPUILU0_);});
```

### DIFF
--- a/opm/simulators/linalg/ISTLSolverBda.cpp
+++ b/opm/simulators/linalg/ISTLSolverBda.cpp
@@ -124,7 +124,9 @@ apply(Vector& rhs,
 	      //NOTE: copyThread can safely write to jacMat because in solve_system both matrix and *blockJacobiForGPUILU0_ diagonal entries
 	      //are checked and potentially overwritten in replaceZeroDiagonal() by mainThread. However, no matter the thread writing sequence,
 	      //the final entry in jacMat is correct.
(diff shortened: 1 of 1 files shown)


daavid00:
jenkins build this please

akva2 (approved):

blattms:
Mmh, I wonder why this should depend on OpenMP at all. @razvnane Why is the variable only there for OpenMP in the first place?
```

#if HAVE_OPENMP
#include <thread>
#include <omp.h>

std::shared_ptr<std::thread> copyThread;
#endif // HAVE_OPENMP
```

Does std::thread really depend OpenMP?

akva2:
it was requested that openmp was used to decide the use of threading here, see original PR.

PR #5406, opened 2024-06-04
only allocate trans_ for globalTrans_

GitPaean:
others are not needed for globalTrans_ for now. It will help to flatten the peak memory usage for rank 0 during a parallel running.

### DIFF
--- a/opm/simulators/flow/CpGridVanguard.hpp
+++ b/opm/simulators/flow/CpGridVanguard.hpp
@@ -285,7 +285,7 @@ protected:
                                                     getPropValue<TypeTag, Properties::EnableEnergy>(),
                                                     getPropValue<TypeTag, Properties::EnableDiffusion>(),
                                                     getPropValue<TypeTag, Properties::EnableDispersion>()));
-        globalTrans_->update(false);
+        globalTrans_->update(false, true);
     }
 
     double getTransmissibility(unsigned I, unsigned J) const override
@@ -306,6 +306,10 @@ protected:
         this->doFilterConnections_(this->schedule());
     }
 
+    // \Note: this globalTrans_ is used for domain decomposition and INIT file output.
+    // It only contains trans_ due to permeability and does not contain thermalHalfTrans_,
+    // diffusivity_ abd dispersivity_. The main reason is to reduce the memory usage for rank 0
+    // during parallel running.
     std::unique_ptr<TransmissibilityType> globalTrans_;
 };
 
--- a/opm/simulators/flow/CpGridVanguard.hpp
+++ b/opm/simulators/flow/CpGridVanguard.hpp
@@ -285,7 +285,7 @@ protected:
                                                     getPropValue<TypeTag, Properties::EnableEnergy>(),
                                                     getPropValue<TypeTag, Properties::EnableDiffusion>(),
                                                     getPropValue<TypeTag, Properties::EnableDispersion>()));
-        globalTrans_->update(false, true);
+        globalTrans_->update(false, TransmissibilityType::TransUpdateQuantities::Trans);
     }
 
     double getTransmissibility(unsigned I, unsigned J) const override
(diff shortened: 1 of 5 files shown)


GitPaean:
jenkins build this please

GitPaean:
With this PR, for the 1 million case SPE11C, with 14 processes, rank 0, it reduces the peak memory usage from 2.26 GB to 2 GB. And the difference will scale following the number of the faces globally. 
Master:
image

PR:
image

GitPaean:
jenkins build this please

opm/simulators/flow/Transmissibility.hpp
```diff
@@ -147,7 +152,7 @@ class Transmissibility {
      *   numerical aquifers.  Default value: \c false, meaning do not apply
      *   regional multipliers to explicit NNCs.
      */
-    void update(bool global, const std::function<unsigned int(unsigned int)>& map = {}, bool applyNncMultRegT = false);
+    void update(bool global, bool onlyTrnas = false, const std::function<unsigned int(unsigned int)>& map = {}, bool applyNncMultRegT = false);
```
blattms, line 155: small typo, that does not really matter: onlyTrans

blattms (approved):
Looks good.

If I understand this correctly, then this relevant before loadbalancing. With this PR we will only create the gloabl transmissibility arrays needed to feed to the loadbalancer.

GitPaean:
jenkins build this please

GitPaean:
I do see this will introduced a `incomplete` `TransmissbilityType`, which might pose problem for certain usage.  Maybe we should introduce the `onlyTrans` as member to the `TransmissbilityType` or something similar, so we can indicate what is offered from the object.

GitPaean:
Adding some comments to emphasize that only trans_ in the globalTrans_, so hopefully people can follow the issue.

GitPaean:
jenkins build this please

opm/simulators/flow/CpGridVanguard.hpp
```diff
@@ -306,6 +306,10 @@ class CpGridVanguard : public FlowBaseVanguard<TypeTag>
         this->doFilterConnections_(this->schedule());
     }
 
+    // \Note: this globalTrans_ is used for domain decompostion and INIT file output.
```
bska, line 309: Tiny misprint: "decompostion" ➡️ "decomposition".

bska (commented):
This is okay, but from a readability point of view I think I'd prefer that we use named enumerators instead of naked `bool`s which cannot be fully understood in isolation.  Although borderline now with just two `bool` parameters we might consider something like
```C++
struct TransUpdate {
    enum class Ranks { Zero, All, };
    enum class Quantities { Trans, All, };

    Ranks ranks;
    Quantities quantities;
};
```
and then one'd call `update()` as
```C++
update({ TransUpdate::Ranks::All, TransUpdate::Quantities::Trans }, /* ... */);
```
which, at least to me, is easier to read/understand than
```C++
update(true, false, /* ... */)
```

GitPaean:
> This is okay, but from a readability point of view I think I'd prefer that we use named enumerators instead of naked `bool`s which cannot be fully understood in isolation. Although borderline now with just two `bool` parameters we might consider something like
> 
> ```c++
> struct TransUpdate {
>     enum class Ranks { Zero, All, };
>     enum class Quantities { Trans, All, };
> 
>     Ranks ranks;
>     Quantities quantities;
> };
> ```
> 
> and then one'd call `update()` as
> 
> ```c++
> update({ TransUpdate::Ranks::All, TransUpdate::Quantities::Trans }, /* ... */);
> ```
> 
> which, at least to me, is easier to read/understand than
> 
> ```c++
> update(true, false, /* ... */)
> ```

this is a good idea and will incorporate.

GitPaean:
At the end I decided to only use `enum class` for onlyTrans parameter. Did not feel it is improving for the `global` parameter. 

And also, there is default values for onlyTrans while not for global. It affects the usage a little bit.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

bska (approved):

PR #5409, opened 2024-06-04
Collect Well Level Control Limits/Targets for Summary Output

bska:
This PR adds a new stage to the `wellData()` report function, `assignWellTargets()`, which collects the numerical values of the currently active limits/targets for summary file output purposes. It is important that we collect these values strictly before we start mutating the `SummaryState` object lest user-defined arguments in keywords like `WCONPROD` be evaluated seemingly too soon. Furthermore, it is faster to collect these values on the simulator side and send them to the I/O rank for file output than to defer the full calculation to the I/O rank.


bska:
I am creating this PR in draft mode because it depends on the new `data::WellControlLimits` structure introduced in OPM/opm-common#4092.  I will keep the PR in a draft state until such time as that upstream PR has been merged thereby unblocking this work.

bska:
jenkins build this opm-common=4092 please

bska:
The upstream PR was merged into the master branch.  I'm marking this as "ready for review" now and I'm running a build check.

bska:
jenkins build this please

akva2 (approved):

bska:
PR approved and build check is green.  I'll merge this into the master branch now.

PR #5410, opened 2024-06-05
Support instantaneous flow rates in extended network (WEFAC and GEFAC item 3)

vkip:
Needs https://github.com/OPM/opm-common/pull/4094

### DIFF
--- a/opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
+++ b/opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
@@ -112,12 +112,6 @@ partiallySupported()
             },
          },
          {
-            "GEFAC",
-            {
-               {3,{true, allow_values<std::string> {"YES"}, "GEFAC(GRPNETWK): Extended Network Model efficiency NO option not implemented"}}, // TRANSFER_EXT_NET
-            },
-         },
-         {
             "GRIDOPTS",
             {
                {1,{true, allow_values<std::string> {"NO", "YES"}, "GRIDOPTS(TRANMULT): should be set to either NO or YES"}}, // TRANMULT
@@ -271,12 +265,6 @@ partiallySupported()
                {3,{true, allow_values<std::string> {"YES"}, "WAGHYSTR(GAS_MODEL): only the YES option is supported – will STOP"}}, // GAS_MODEL
                {4,{true, allow_values<std::string> {"NO"}, "WAGHYSTR(RES_OIL): only the NO option is supported – will STOP"}}, // RES_OIL
                {5,{true, allow_values<std::string> {"NO"}, "WAGHYSTR(WATER_MODEL): only the NO option is supported – will STOP"}}, // WATER_MODEL
-            },
-         },
-         {
-            "WEFAC",
-            {
-               {3,{true, allow_values<std::string> {"YES"}, "WEFAC(WELNETWK): only the YES option is supported"}}, // EXTENDED_NETWORK_OPT
             },
          },
          {
--- a/opm/simulators/wells/GroupState.cpp
+++ b/opm/simulators/wells/GroupState.cpp
@@ -41,7 +41,7 @@ GroupState<Scalar> GroupState<Scalar>::serializationTestObject()
 {
     GroupState result(3);
     result.m_production_rates = {{"test1", {1.0, 2.0}}};
-    result.m_network_production_rates={{"test1", {1.0, 20}}};
+    result.m_network_leaf_node_production_rates={{"test1", {1.0, 20}}};
     result.production_controls = {{"test2", Group::ProductionCMode::LRAT}};
     result.prod_red_rates = {{"test3", {3.0, 4.0, 5.0}}};
     result.inj_red_rates = {{"test4", {6.0, 7.0}}};
@@ -62,7 +62,7 @@ template<class Scalar>
 bool GroupState<Scalar>::operator==(const GroupState& other) const
 {
     return this->m_production_rates == other.m_production_rates &&
-           this->m_network_production_rates == other.m_network_production_rates &&
+           this->m_network_leaf_node_production_rates == other.m_network_leaf_node_production_rates &&
            this->production_controls == other.production_controls &&
            this->prod_red_rates == other.prod_red_rates &&
            this->inj_red_rates == other.inj_red_rates &&
@@ -86,10 +86,10 @@ bool GroupState<Scalar>::has_production_rates(const std::string& gname) const
 }
 
 template<class Scalar>
-bool GroupState<Scalar>::has_network_production_rates(const std::string& gname) const
+bool GroupState<Scalar>::has_network_leaf_node_production_rates(const std::string& gname) const
 {
-    auto group_iter = this->m_network_production_rates.find(gname);
-    return (group_iter != this->m_network_production_rates.end());
+    auto group_iter = this->m_network_leaf_node_production_rates.find(gname);
+    return (group_iter != this->m_network_leaf_node_production_rates.end());
 }
 
 template<class Scalar>
@@ -103,13 +103,13 @@ void GroupState<Scalar>::update_production_rates(const std::string& gname,
 }
 
 template<class Scalar>
-void GroupState<Scalar>::update_network_production_rates(const std::string& gname,
+void GroupState<Scalar>::update_network_leaf_node_production_rates(const std::string& gname,
                                                  const std::vector<Scalar>& rates)
 {
     if (rates.size() != this->num_phases)
         throw std::logic_error("Wrong number of phases");
 
-    this->m_network_production_rates[gname] = rates;
+    this->m_network_leaf_node_production_rates[gname] = rates;
 }
 
 template<class Scalar>
@@ -125,10 +125,10 @@ GroupState<Scalar>::production_rates(const std::string& gname) const
 
 template<class Scalar>
 const std::vector<Scalar>&
-GroupState<Scalar>::network_production_rates(const std::string& gname) const
+GroupState<Scalar>::network_leaf_node_production_rates(const std::string& gname) const
 {
-    auto group_iter = this->m_network_production_rates.find(gname);
-    if (group_iter == this->m_network_production_rates.end())
+    auto group_iter = this->m_network_leaf_node_production_rates.find(gname);
+    if (group_iter == this->m_network_leaf_node_production_rates.end())
         throw std::logic_error("No such group");
 
     return group_iter->second;
--- a/opm/simulators/wells/GroupState.cpp
+++ b/opm/simulators/wells/GroupState.cpp
@@ -41,6 +41,7 @@ GroupState<Scalar> GroupState<Scalar>::serializationTestObject()
 {
     GroupState result(3);
     result.m_production_rates = {{"test1", {1.0, 2.0}}};
+    result.m_network_production_rates={{"test1", {1.0, 20}}};
     result.production_controls = {{"test2", Group::ProductionCMode::LRAT}};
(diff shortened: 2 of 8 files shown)


vkip:
jenkins build this opm-common=4094 please

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -673,6 +674,17 @@ updateGroupProductionRates(const Group& group,
         rates[phase] = sumWellPhaseRates(false, group, schedule, wellState, reportStepIdx, phase, /*isInjector*/ false);
     }
     group_state.update_production_rates(group.name(), rates);
+
+    const auto& network = schedule[reportStepIdx].network();
+    if (network.active()) {
+        std::vector<Scalar> network_rates = rates;
+        if (network.needs_instantaneous_rates(schedule, reportStepIdx)) {
+            for (int phase = 0; phase < np; ++phase) {
+                network_rates[phase] = sumWellPhaseRates(false, group, schedule, wellState, reportStepIdx, phase, /*isInjector*/ false, /*network*/ true);
```
GitPaean, line 683: I can be wrong here. It looks like for the `network_production_rates`, we only the values for all the `leaf_nodes` in its current form. 

Maybe we should only need to do differently for the `leaf_nodes`.  It looks like we will deal with other non-leaf nodes in the function `computeNetworkPressures` based on the rates of the `leaf_nodes`. 

If that sounds correct to you, maybe we can deal with this in the function `computeNetworkPressures` where the `leaf_nodes` information are available?

GitPaean (commented):
hei, I struggled to understand the roles of `m_network_production_rates` and `node_inflows`.  It is `node_inflows` that is used to update the network pressure. There is some overlap between these two. 

I left some comments for consideration and discussion.

opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
```diff
@@ -284,7 +284,7 @@ partiallySupported()
          {
             "WEFAC",
             {
-               {3,{true, allow_values<std::string> {"YES"}, "WEFAC(WELNETWK): only the YES option is supported"}}, // EXTENDED_NETWORK_OPT
+               {3,{true, allow_values<std::string> {"YES","NO"}, "WEFAC(WELNETWK): value must be YES or NO"}}, // USE_WEFAC_IN_NETWORK
```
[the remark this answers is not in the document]
vkip: We could, and probably should, do that, but I think we give users a better error message by leaving it in there. 

By explicitly listing only YES and NO as supported we get, e.g.:

--
WEFAC: invalid value 'YESss' for item 3
In file: TEST.DATA, line NNN
WEFAC(WELNETWK): value must be YES or NO
--

whereas we get something like the following (and at a later stage) if we remove the keywords from PartiallySupported:
--
In TEST.DATA line NNN
Internal error: Could not convert string 'NOSE' to bool
--

So what I would actually like is better first-level checking of fully supported keywords, but maybe we can leave it as is until that is in place?

opm/simulators/wells/GroupState.cpp
```diff
@@ -79,6 +81,13 @@ bool GroupState<Scalar>::has_production_rates(const std::string& gname) const
     return (group_iter != this->m_production_rates.end());
 }
 
+template<class Scalar>
+bool GroupState<Scalar>::has_network_production_rates(const std::string& gname) const
+{
+    auto group_iter = this->m_network_production_rates.find(gname);
```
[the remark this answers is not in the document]
vkip: I don't disagree, but I just copied the style of the existing 'has_production_rates' function. Can update that as well then. I would like to keep the function, I expect it to be required for output purposes (see below).

opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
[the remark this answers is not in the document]
GitPaean: I share your view regarding what is a good error message. At the same time, for the item 3, it is only supposed to have `YES` or `NO` two options, then we are fully supporting item (keyword) when we can handle `YES` and `NO`, it is a little weird to have it in the file that indicating `PartiallySupported`. 

Not sure whether we should enforce good error message through this strategy.  Maybe @bska can comment?

opm/simulators/wells/GroupState.cpp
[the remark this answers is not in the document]
GitPaean: It is fine and partly personal preference. When we support c++20, we can use the member function `contains`.

opm/simulators/wells/WellGroupHelpers.cpp
vkip: You are right, only the leaf nodes are used in computeNetworkPressures, but this seemed to be the easies way to get those correctly. Also, to be able to report the rates I expect we will need to have all the rates actually used in the network available. 

I may have missed something, but to handle this entirely within the computeNetworkPressures it appears that large parts of the sumWellPhaseRates function would need to be duplicated (essentially everything below the recursive call), so I fail to see how that would be simpler (?)

opm/simulators/wells/GroupState.cpp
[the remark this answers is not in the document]
vkip: Ok, I'll leave this as is then 👍

opm/simulators/wells/WellGroupHelpers.cpp
GitPaean: One important thing missing from current `m_network_production_rates` in the current form is that, if a network node is not in the GRUPTREE, it will not be in the `m_network_production_rates`.  Because network nodes and group nodes can be different. 

And also, can we call `sumWellPhaseRates` through `computeNetworkPressures` in some way instead of duplicating them. I could not tell why we can not do that by looking at the code.

Ideally, we might need a function called `updateNetworkProductionRates()`. You do not have to do it unless you are really up to it. We can make a refactoring PR following this PR.

opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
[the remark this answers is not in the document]
GitPaean: My current view is that we should enforce the good error message in the parser side.  Unless we want to have a `FullySupportedFlowKeywords.cpp` to regulate from the simulator side in an systematic level. It is confusing to have it in the `PartiallySupportedFlowKeywords` file.

opm/simulators/wells/WellGroupHelpers.cpp
GitPaean: You are correct when coming to the parallel communication. I did think about it and my apologizes.  Will respond with more looking into it.

vkip: No worries, and thanks! I'll await further input then.

GitPaean: (without considering NEFAC, then basically, NEFAC is 1.0 for the nodes not in the GRUPTREE)

In the current way, `m_network_production_rates` holds the values for all the group nodes.  `node_inflows` holds a complete version of the rates for all the network nodes. 

When all the group nodes are in the network, for the nodes they both contain, they should have the same values. although calculated through different paths and NEFAC is 1.0. 

For the situations where some group nodes are not in the network, for example, they have different levels of hierarchy, how to calculate the values (using the efficiency factors) I think is something we can only test to find out, it is not totally clear to me about how things should be done. Likely, `m_network_production_rates` can have different values from the values in `node_inflows`, depending on the understanding we use at the end. 

In the current code, `computeNetworkPressures` is going through the network branches, and `updateGroupProductionRates` is going through the group tree.  We do not want to do global communication in the function `computeNetworkPressures`.  Ideally we can borrow the function `updateGroupProductionRates` to get leaf-nodes (only leaf-nodes), and handle the network-way of production rates in `computeNetworkPressures`. 

Now I agree with you that the current way is a simpler way to get things done. And also, I see that with NEFAC and nodes in the GRUPTREE and not in the network probably introduces re-working in these parts. We can get the results correct first and follow with re-factoring in these parts. 

Sorry for the previous misunderstandings.

vkip: No problem, I can see it is confusing with the member ```m_network_production_rates``` where only leaf rates are used, and will try to make it more clear with proper comments. Glad we can agree to move forward with the current overall approach for now. Thank you.

GitPaean: > When all the group nodes are in the network, for the nodes they both contain, they should have the same values. although calculated through different paths and NEFAC is 1.0.

Just for record in case it helps understanding and design or for later usage.

Not totally sure though. With extended network model, we might have different hierarchy layout between nodes, then for the non-leaf nodes, they will have  very different values between `m_network_production_rates` and `node_inflows`. Luckily, the leaf nodes in the `m_network_production_rates`  should still have the correct/desired value.  `node_inflows` will most likely be the one holding the correct/desired values that represent the network structure.

GitPaean:
Please give a hint when the update is finished with this PR. And maybe @bska can have a look at the changes related to KeywordValidition and FullySupportedFlowKeywords?

vkip:
> Please give a hint when the update is finished with this PR. And maybe @bska can have a look at the changes related to KeywordValidition and FullySupportedFlowKeywords?

Will do, placing this and the companion in opm-common in draft for now.

vkip:
Now contains 5866 for testing - will rebase once that is merged.

opm/simulators/wells/WellGroupHelpers.cpp
vkip: Updated now to support NEFAC and only compute and store the leaf node rates in GroupState.

vkip:
jenkins build this opm-common=4094 please

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -962,13 +977,14 @@ computeNetworkPressures(const Network::ExtNetwork& network,
                 // Add downbranch rates to upbranch.
                 std::vector<Scalar>& up = node_inflows[(*upbranch).uptree_node()];
                 const std::vector<Scalar>& down = node_inflows[node];
+                // @TODO@ Also support NEFAC (for nodes that do not correspond to groups)
+                const double efficiency = network.node(node).efficiency();
```
GitPaean, line 981: `Scalar` here.

GitPaean (changes requested):
It looks correct to me.  Assuming the results have been verified and I only left a few minor comments.

GitPaean:
With https://github.com/OPM/opm-tests/pull/1175 merged, I think we can have a regression test with it. It can go in with this PR or go in with a separate following up PR.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -962,13 +977,14 @@ computeNetworkPressures(const Network::ExtNetwork& network,
                 // Add downbranch rates to upbranch.
                 std::vector<Scalar>& up = node_inflows[(*upbranch).uptree_node()];
                 const std::vector<Scalar>& down = node_inflows[node];
+                // @TODO@ Also support NEFAC (for nodes that do not correspond to groups)
```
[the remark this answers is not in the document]
vkip: Actually this is now supported, updated comment.

vkip:
> It looks correct to me. Assuming the results have been verified and I only left a few minor comments.

Thank you for the review, will update now. I agree that we should add the test, but let's do that in a follow-up PR.

vkip:
jenkins build this opm-common=4094 please

GitPaean:
Thanks for the update. The upstream PR https://github.com/OPM/opm-common/pull/4094 has been merged. I am also getting this PR in now.

PR #5411, opened 2024-06-05
Enable WTMULT as an Input Operation

bska:
Basic support for this keyword was added in commit OPM/opm-common@5e3e20c552c10f69f0717d19fe88c13119ec3e71 (PR OPM/opm-common#2763) and this PR enables running models which use that basic support. Advanced uses, such as including user-defined arguments for the multipliers, will still be rejected at the input level.


bska:
jenkins build this please

bska:
jenkins build this please

bska:
This is a simple change that unblocks important functionality.  I'll self-merge to activate the support.

PR #5418, opened 2024-06-07
extending drsdtcon with regimes, and option for GAS/WATER

trinemykk:
Extended the DRSDTCON to include more parameters that controls the regime and rate of convective mixing.
Included a convective term in the transport flux for the brine.
Also works for GAS/WATER (not only GAS/OIL as before)

Has corresponding PR in opm-models (OPM/opm-models#902) and opm-common (OPM/opm-common#4097)

### DIFF
--- a/opm/simulators/flow/EquilInitializer.hpp
+++ b/opm/simulators/flow/EquilInitializer.hpp
@@ -174,6 +174,7 @@ public:
                     const auto& h = FluidSystem::enthalpy(fluidState, phaseIdx, regionIdx);
                     fluidState.setEnthalpy(phaseIdx, h);
                 }
+
             }
 
             // set the salt concentration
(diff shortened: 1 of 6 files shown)


opm/simulators/flow/EquilInitializer.hpp
```diff
@@ -174,6 +174,7 @@ class EquilInitializer
                     const auto& h = FluidSystem::enthalpy(fluidState, phaseIdx, regionIdx);
                     fluidState.setEnthalpy(phaseIdx, h);
                 }
+
             }
```
totto82, line 178: Nitpick. Please remove the added line.

opm/simulators/flow/FlowProblem.hpp
```diff
@@ -520,6 +524,8 @@ class FlowProblem : public GetPropType<TypeTag, Properties::BaseProblem>
         const auto& schedule = simulator.vanguard().schedule();
         const auto& events = schedule[episodeIdx].events();
 
+
+
         if (episodeIdx >= 0 && events.hasEvent(ScheduleEvents::GEO_MODIFIER)) {
```
totto82, line 529: Nitpick. Please remove the added line.

totto82 (commented):
Thanks for the contribution. I only have minor comments.

totto82:
I will close this since its content is part of #5491

PR #5419, opened 2024-06-07
trying to release global transmissibility earlier to flatten the peak memory usage in the function finishInit()

GitPaean:


GitPaean:
jenkins build this please

GitPaean:
With the PR in its current form, the rank 0 memory usage (1 million cell SPE11c, 14 processes) looks like the following, 

image

We should try to look into the peak in the next stage with this PR.

GitPaean:
master branch after https://github.com/OPM/opm-simulators/pull/5406 looks like the following, 

image

GitPaean:
jenkins build this please

GitPaean:
with the latest commit 112d218, the memory profiling looks like the following, 

image

not totally sure jenkins will approve though.

GitPaean:
removing https://github.com/OPM/opm-simulators/commit/112d2181e5574f205f81afa6600733c3b655126a, from this PR for now and marking this PR as ready for review.

GitPaean:
But https://github.com/OPM/opm-simulators/commit/112d2181e5574f205f81afa6600733c3b655126a can change running results (not only some difference in the output file) is a little concerning.

opm/simulators/flow/FlowProblem.hpp
```diff
@@ -343,8 +343,39 @@ class FlowProblem : public GetPropType<TypeTag, Properties::BaseProblem>
...
+        bool transmissiblity_updated = false;
+        auto processTransmissibilities = [&simulator, &transmissiblity_updated, this]() {
+            // Re-ordering in case of ALUGrid
+            auto gridToEquilGrid = [&simulator](unsigned int i) -> unsigned int {
+                return simulator.vanguard().gridIdxToEquilGridIdx(i);
+            };
+
+            transmissibilities_.finishInit(gridToEquilGrid);
+            transmissiblity_updated = true;
+        };
```
bska, line 357: I think I'd prefer to write this **slightly** differently in order to avoid the mutable `transmissibility_updated` object with a very large scope:
```C++
auto finishTransmissibilities = [updated = false, this]() mutable
{
    if (updated) { return; }
    this->transmissibilities_.finishInit([&vg = this->simulator().vanguard()](const unsigned int it)
        return vg.gridIdxToEquilGridIdx(i);
    }
    updated = true;
};
```
That way you can unconditionally call `finishTransmissibilities()` and the lambda will do the right thing.

opm/simulators/flow/EclGenericWriter_impl.hpp
```diff
@@ -241,19 +241,29 @@ eclIO() const
...
+    }
+}
+template<class Grid, class EquilGrid, class GridView, class ElementMapper, class Scalar>
+void
+EclGenericWriter<Grid,EquilGrid,GridView,ElementMapper,Scalar>::
+extractOutputTransAndNNC(const std::function<unsigned int(unsigned int)>& map)
+{
+    if (collectOnIORank_.isIORank()) {
+        auto cartMap = cartesianToCompressed(equilGrid_->size(0), UgGridHelpers::globalCell(*equilGrid_));
+        this->outputTrans_ = std::make_unique<data::Solution>(computeTrans_(cartMap, map));
```
bska, line 265: If `extractOutputTransAndNnc()` is the only function which calls `computeTrans_()` and assigns to `outputTrans_()`, then maybe we could change `computeTrans_()` to assign `outputTrans_()` directly and to return `void` instead?

bska (commented):
Other than a couple of small remarks this looks good to me.  I do think we should look at refactoring `FlowProblem<>::finishInit()` since that function is getting really long now, but that could be the subject of follow-up work.

opm/simulators/flow/FlowProblem.hpp
GitPaean: This is a very good suggestion.  Incorporated already.

opm/simulators/flow/EclGenericWriter_impl.hpp
GitPaean: done!

bska (approved):

PR #5422, opened 2024-06-11
Fix to error in Clang: use 'template' keyword to treat 'data' ...

daavid00:


daavid00:
jenkins build this please

atgeirr:
Thanks for the fix, I will merge. Please always include a description in the PR.

daavid00:
> Thanks for the fix, I will merge. Please always include a description in the PR.

Sorry for that, I thought for this small fixes it was enough to write this in the name of the commit.

PR #5423, opened 2024-06-12
Only indicate problems for exceptions whem updating well potentials,

blattms:
Nearly all exceptions throw when computing well potentoals will not abort the simulator but result in timestep chops. Hence those should not be counted as errors (e.g. by calling the OPM_*THROW* macros) and be reported in the PRT file.

This change will cause at least two more  occurences (in MSWellHelpers) to be treated as problems. For this we added a new helper function.Nearly all exceptions throw when computing wells will not abort the simulator but result in timestep chops. Hence those should not be counted as errors (e.g. by calling the OPM_*THROW* macros) and be reported in the PRT file.

### DIFF
--- a/opm/simulators/utils/DeferredLoggingErrorHelpers.hpp
+++ b/opm/simulators/utils/DeferredLoggingErrorHelpers.hpp
@@ -130,7 +130,27 @@ inline void logAndCheckForExceptionsAndThrow(Opm::DeferredLogger& deferred_logge
     _throw(exc_type, message, comm);
 }
 
+inline void logAndCheckForProblemsAndThrow(Opm::DeferredLogger& deferred_logger,
+                                          Opm::ExceptionType::ExcEnum exc_type,
+                                          const std::string& message,
+                                          const bool terminal_output,
+                                          Opm::Parallel::Communication comm)
+{
+    // add exception message to logger in order to display message from all ranks
+    if (exc_type != Opm::ExceptionType::NONE && comm.size() > 1) {
+        deferred_logger.problem("[Exception on rank " + std::to_string(comm.rank()) + "]: " + message);
+    }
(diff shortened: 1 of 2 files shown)


blattms:
Arggh,  needs to be on a higher level.

atgeirr:
It seems to me that this particular error should not result in a Problem and timestep chop but terminating the simulator, but in the current master it will produce an Error but then chop and continue. Do you agree?

blattms:
Before force-pushing there were two. [NumericalProblem](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/MSWellHelpers.cpp#L124) and [std::runtime_error](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/MSWellHelpers.cpp#L273). I guess you mean the latter?

I don't like the new force-pushed version very much, but it works. Maybe we should just always use the problem category whenever there is a NumericalProblem in the well code. Maybe even use special custom exceptions here. That might make the code easier and more expressive

blattms:
jenkins build this please

blattms:
> > It seems to me that this particular error should not result in a Problem and timestep chop but terminating the simulator, but in the current master it will produce an Error but then chop and continue. Do you agree?
> 
> Before force-pushing there were two. [NumericalProblem](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/MSWellHelpers.cpp#L124) and [std::runtime_error](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/MSWellHelpers.cpp#L273). I guess you mean the latter?
> 

Ping @atgeirr: Which one was it?

atgeirr:
> Ping @atgeirr: Which one was it?

I do not recall the previous state of the PR well enough to say, sorry! However the current state of the PR seems fine, and can be merged if you do not want to make further changes.

atgeirr (approved):

PR #5426, opened 2024-06-14
fix option for output every timestep

totto82:
See https://github.com/OPM/opm-common/pull/4104


bska (approved):

PR #5427, opened 2024-06-14
Do NOT delete files recursively

vkip:
Recursive deletion causes results in all subfolders to be deleted if cmdline-arg output-dir happens to be empty in a single instance..


atgeirr:
jenkins build this please

bska (approved):
Good point.  I should have picked up on that whilst reviewing #5168.

atgeirr (approved):

bska:
Build check is green and PR approved.  I'll merge into master.  Thanks a lot, again, for fixing this problem!

PR #5429, opened 2024-06-16
Remove the MOD4_GRP regression test

alfbr:
Removing the test as it is broken. More specifically, the last part of the simulation is unstable with oscillating results, making regression assessment futile. See #4855 for further discussion and details.

I did not see where to turn off data updating, maybe that is handled automatically these days?


bska:
> I did not see where to turn off data updating, maybe that is handled automatically these days?

Yes, that is indeed handled automatically.  PR #4674 made that change.

bska:
jenkins build this please

bska (approved):

PR #4748, opened 2023-07-12
Enable items, 11, 12 and 13 of GCONPROD

plgbrts:
This PR depends on https://github.com/OPM/opm-common/pull/3601

PR #5432, opened 2024-06-19
Allow GCONPROD item 11-13 to actually take effect.

vkip:
Support for GCONPROD item 11-13 was added in https://github.com/OPM/opm-common/pull/3601, https://github.com/OPM/opm-simulators/pull/4748 , but does not take effect (at least for a lot of cases). 

With this PR results are as expected on the attached tests. 
[gconprod_item11to13.tar.gz](https://github.com/user-attachments/files/15898172/gconprod_item11to13.tar.gz)

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelConstraints.cpp
+++ b/opm/simulators/wells/BlackoilWellModelConstraints.cpp
@@ -483,7 +483,7 @@ actionOnBrokenConstraints(const Group& group,
     std::string ss;
     switch (group_limit_action.allRates) {
     case Group::ExceedAction::NONE: {
-        if (oldControl != newControl && oldControl != Group::ProductionCMode::NONE) {
+        if (oldControl != newControl) {
             if ((group_limit_action.water == Group::ExceedAction::RATE &&
                  newControl == Group::ProductionCMode::WRAT) ||
                 (group_limit_action.gas == Group::ExceedAction::RATE &&


vkip:
jenkins build this please

vkip:
Results on above test cases w/wo PR:

glpr
gwpr
ggpr

GitPaean:
The content of the PR is probably correct. Please allow some time so that I can check the implementation of the related PRs before I confirm what we should do here.

opm/simulators/wells/BlackoilWellModelConstraints.cpp
```diff
@@ -483,7 +483,7 @@ actionOnBrokenConstraints(const Group& group,
     std::string ss;
     switch (group_limit_action.allRates) {
     case Group::ExceedAction::NONE: {
-        if (oldControl != newControl && oldControl != Group::ProductionCMode::NONE) {
+        if (oldControl != newControl) {
```
GitPaean, line 486: I think the change is positive. I do not understand why we need to check `oldControl` to decide whether we can switch to a group control.

GitPaean:
And it looks both this PR and master can not handle the following situation, 

```
GCONPROD
 'PLAT-A' 'WRAT'  1*  100.0  1* 1*  NONE 3*  NONE   /
```
It should be water rate control, with the master and this PR the simulation just run without any group constraint. It came as a surprise. 

image

GitPaean:
Both the master and PR can handle the following setup. I was looking at the wrong group. 

```
 GCONPROD
  'PLAT-A' 'WRAT'  1*  100.0  1* 1*  NONE 3*  NONE   /
 /
```

opm/simulators/wells/BlackoilWellModelConstraints.cpp
GitPaean: With more investigation, I think the fix is correct, while I am not convinced we are managing the actions exceeding group constraint properly as a whole. 

But removing `oldControl != Group::ProductionCMode::NONE` is sensible any way.

GitPaean (approved):

GitPaean:
jenkins build this please

opm/simulators/wells/BlackoilWellModelConstraints.cpp
vkip: Great, thanks for reviewing this!

GitPaean: For example,  we probably should not include all the actions inside this condition.  In the parsing stage, we should generate the proper actions for `oil, water, gas and liquid` properly, because the item 7 and item 11-13 and item 2 should have combined effects. We should handle it in the parsing stage.  The following code might have problems if we defined item 7 (not NONE), the definition in item 11-13 might be ignored (which should override item 7 actually).  For sure, the current way of working will have big problem to be extended to other actions. 
(It is complicated and I might read things wrong. But leaving some comments in case we will revisit this part in the future development)

```c++
    switch (group_limit_action.allRates) {
    case Group::ExceedAction::NONE: {
```

vkip: I think you are right in that this needs to be changed in to support other actions than 'RATE', but for now it appears that all combinations of NONE or RATE in items 7,11,12,13 work as expected (?)

GitPaean: Too occupied to go check the details, while not sure whether it will work if we specifies item 7 with RATE, while using `NONE` in some of the items 11-13 will work. 

For example, (item 2 is 'NONE`) if we use `RATE` for item 7, and use `NONE` for all of the items 11-13, then it should be RATE action for oil constraint (not considering the constraints after item 13) and NONE for water, gas and liquid.  

Furthermore, when we using `allRates` in the parser side, it is difficult to do something specifically for `oil`. 

I can be wrong though, testing might show otherwise.

GitPaean: It is not clear to me whether we can use `NONE` in item 11-13 to override non-NONE actions in item 7 either.

vkip: Did a quick test to confirm, and at least this seems to work as expected ([test.pdf](https://github.com/user-attachments/files/15956287/test.pdf)):
```
GCONPROD
  'PLAT-A' 'NONE' 3900.0 100.0 100.0 100.0 RATE 3* NONE NONE NONE /
/
```

GitPaean: Thanks for checking. It will be interesting to see the manner how the actions get enforced.  My hypothesis was wrong.

vkip: It seems the way it works is by never adding the constraints in the first place if overridden by 'NONE' in items 11-13 (but I could be wrong.. :) )

GitPaean: I think you are correct. That is the only way it can work, otherwise the simulator side code will not work for the specification you specified.

GitPaean: Last comment here is that we have `allRates` and the `water`, `gas` and `liquid` (they considered the item 7 and item 11-13). We use `allRates` for the `oil` and the others for themselves respectively.  Probably we can using `oil` instead of `allRates` here.  And in the simulator code, we can avoid the current way of using `allRates` since we will not have it there.

GitPaean:
Thanks for the contribution. Jenkins approves. I am getting the PR in.

PR #5434, opened 2024-06-20
[fix] Do not count experiencing nans as error in PRT file.

blattms:
The simulation will just chop the time step and continue. Note, that the error count in the PRT file is used by engineers to decide whether a simulation was successfull. Hence the error count should not be increased here.

Downstream of OPM/opm-common#4113


atgeirr (approved):

atgeirr:
jenkins build this please

blattms:
I have merged the upstream PR. Let's see whether jenkins picks that up.

PR #5435, opened 2024-06-20
Minor updates in preparation for New defaults #5157.

steink:
Minor update in preparation for New defaults #5157.
* Fix missing treatment of solvent for STW potential calculations
* Do update of connection pressures relative to bhp before calculating STW potentials. Probably not neccessary, but this give closer match with current master when potentials are computed at defaulted bhp

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -1624,6 +1624,11 @@ namespace Opm
         // prepare/modify well state and control
         well_copy.prepareForPotentialCalculations(summary_state, well_state_copy, inj_controls, prod_controls);
 
+       // update connection pressures relative to updated bhp to get better estimate of connection dp
+        const int num_perf = ws.perf_data.size();
+        for (int perf = 0; perf < num_perf; ++perf) {
+            ws.perf_data.pressure[perf] = ws.bhp + well_copy.connections_.pressure_diff(perf);
+        }
         // initialize rates from previous potentials
         const int np = this->number_of_phases_;
         bool trivial = true;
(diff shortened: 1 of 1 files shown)


steink:
jenkins build this please

steink:
The two failures are as expected since these are the only two that currently use ` --local-well-solve-control-switching=true` and potential computations for this option here is slightly adjusted.

atgeirr:
> The two failures are as expected

Seems to me that the match is close anyway, I assume you have looked at this, and that we should do a data update and merge this? (After review...)

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -1650,9 +1655,18 @@ namespace Opm
         // fetch potentials (sign is updated on the outside).
         well_potentials.clear();
         well_potentials.resize(np, 0.0);
-        for (int compIdx = 0; compIdx < this->num_components_; ++compIdx) {
-            const EvalWell rate = well_copy.primary_variables_.getQs(compIdx);
-            well_potentials[this->modelCompIdxToFlowCompIdx(compIdx)] = rate.value();
+        for (int phase = 0; phase < np; ++phase) {
```
GitPaean, line 1658: By definition, we use `comp_idx` for `getQs`. Do you have some concrete observation for the change in this part?

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -1650,9 +1655,18 @@ namespace Opm
...
+            const EvalWell rate = well_copy.primary_variables_.getQs(phase);
+            well_potentials[this->modelCompIdxToFlowCompIdx(phase)] = rate.value();
+        }
+
+        // the solvent contribution is added to the gas potentials
+        if constexpr (has_solvent) {
+            const auto& pu = this->phaseUsage();
```
GitPaean, line 1665: Does this affect any regression testing cases? It looks sensible to me.

opm/simulators/wells/StandardWell_impl.hpp
steink: Yes, that's what I though, but this is where the inclusion of solvent fails as `num_compenents_` is 4 and there's no `getQs` for that.

GitPaean: The function `modelCompIdxToFlowCompIdx` is calling for component indices. 

I think we can do this, 
```c++
        for (int comp_idx = 0; comp_idx < this->num_components_; ++comp_idx) {
            if (has_solvent && comp_idx == Indices::contiSolventEqIdx) continue; // we do not store the solvent in the well_potentials
            const EvalWell rate = well_copy.primary_variables_.getQs(comp_idx);
            well_potentials[this->modelCompIdxToFlowCompIdx(comp_idx)] = rate.value();
        }
```

But it also reveals that potentially in multiple occasions, we might do things related to solvent not totally correct.

GitPaean: Furthermore, if you want, you can incorporate the code below inside the if condition, `if (has_solvent && comp_idx == Indices::contiSolventEqIdx)` to add the solvent to the gas `well_potentials`.  We are basically merging two code blocks to be one and might benefit the future way of coding related to solvent.

steink: Thanks, your suggestion makes more sense. Have updated the PR.

steink: > Furthermore, if you want, you can incorporate the code below inside the if condition, `if (has_solvent && comp_idx == Indices::contiSolventEqIdx)` to add the solvent to the gas `well_potentials`. We are basically merging two code blocks to be one and might benefit the future way of coding related to solvent.

Ah, updated before I saw your comment

GitPaean:
thanks for the update. It looks good to me now. I am triggering the last jenkins benchmark.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this opm-tests=1187 please

GitPaean:
The reference update PR has been merged. I am merging this PR now.

PR #5436, opened 2024-06-20
Add support for SLAVES keyword

hakonhagland:
Depends on upstream https://github.com/OPM/opm-common/pull/4114.

This is a first step to implement reservoir coupling

Currently the `SLAVES` keywords does nothing. The plan is to activate the keyword after `GRUPMAST` and `GRUPSLAV` has been implemented. This can hopefully be done using [`MPI_Comm_spawn()`](https://www.open-mpi.org/doc/v4.1/man3/MPI_Comm_spawn.3.php) from the master, and then MPI communication between master and slaves.

### DIFF
--- a/opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
+++ b/opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
@@ -230,6 +230,13 @@ partiallySupported()
             },
          },
          {
+            "SLAVES",
+            {
+                {1,{true, [](const std::string& val){ return val.size()<=8;}, "SLAVES(SLAVE_RESERVOIR): Only names of slave reservoirs of up to 8 characters are supported."}},
+                {3,{true, allow_values<std::string> {}, "SLAVES(HOST_NAME): should be defaulted. A specific host name is not implemented yet"}}, // HOST_NAME
+            },
+         },
+         {
             "TABDIMS",
             {
                {20,{false, allow_values<std::string> {}, "TABDIMS(NOTUSED): should be defaulted (1*) - ignored as not used"}}, // ITEM20_NOT_USED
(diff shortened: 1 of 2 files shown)


totto82:
Since we SLAVES does not do anything at the moment, isn't it better to wait with merging this until we actually supports this feature. For testing you can always set the parsing-strictness=low

hakonhagland:
> Since we SLAVES does not do anything at the moment, isn't it better to wait with merging this until we actually supports this feature. 

@totto82 Good point.

hakonhagland:
I convert this to draft until it is ready for merging

hakonhagland:
Rebased

hakonhagland:
@blattms I have removed the draft state of this PR, in accordance with https://github.com/OPM/opm-simulators/pull/5620#issuecomment-2368042583. Can you have a look again at this PR and see if we can get it merged?

opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
```diff
@@ -229,6 +229,13 @@ partiallySupported()
                {5,{true, allow_values<std::string> {"F"}, "SPECGRID(TYPE): only option F (Cartesian grids supported) supported"}}, // COORD_TYPE
             },
          },
+         {
+            "SLAVES",
+            {
+                {1,{true, [](const std::string& val){ return val.size()<=8;}, "SLAVES(SLAVE_RESERVOIR): Only names of slave reservoirs of up to 8 characters are supported."}},
```
hakonhagland, line 235: > We should check with the engineers whether they really only use 8 or reky on truncation and use the rest for further documentation. I think the commercial simulator just truncates.

@blattms Referring to: https://github.com/OPM/opm-simulators/pull/5453#discussion_r1746986481 : then we should use "false" instead of "true" here and in the keyword handlers in opm-common, we should truncate the length if the slave reservoir if it is more than 8 characters long?

blattms:
I would like to run jennkins here. Would be cool if you would rebase this onto current master. Without that some updated tests might actually fail.

blattms:
Just to clarify: This PR is there to simply mark the SLAVE keyword as supported?

IMHO, then this PR should actually be the last one in the sequence to support reservoir coupling. Once this is a user can run simulations with coupled reservoirs with the default option. He might expect that such a simulation then works and be surprised that it does not, because the feature is still work in progress.

I don't think that having the SLAVE keyword marked as supported is needed for testing. I would simply start flow with `--parsing-strictness=low` to make it continue the simulation with keywords marked as unsupported.

hakonhagland:
@blattms this PR should actually be the last one in the sequence to support reservoir coupling.

Good point. I put this back into draft state then.

hakonhagland:
Closing this since it will be replaced by #5643.

PR #5438, opened 2024-06-24
Add New Platform for Saturation Function Consistency Checks

bska:
The intention is that this will ultimately replace the existing `RelpermDiagnostics` component which does not really work in parallel and which does not report enough context to help diagnose underlying issues.  For now, though, we just add the shell of a new set of checks and hook that up to the build.

Class `SatfuncConsistencyChecks<Scalar>` manages a configurable set of consistency checks, the implementations of which must publicly derive from `SatfuncConsistencyChecks<Scalar>::Check`.  Client code will configure a set of checks by first calling
```
SatfuncConsistencyChecks<Scalar>::resetCheckSet()
```
then register individual checks by calling
```
SatfuncConsistencyChecks<Scalar>::addCheck()
```
and finally build requisite internal structures by calling
```
SatfuncConsistencyChecks<Scalar>::finaliseCheckSet()
```
Client code will then run the checks by calling
```
SatfuncConsistencyChecks<Scalar>::checkEndpoints()
```
typically in a loop.  Class `SatfuncConsistencyChecks<Scalar>` will count consistency check failures and attribute these to each individual check as needed.  We also maintain separate counts for "Standard" and "Critical" failures.  The former will typically generate warnings while the latter will typically cause the simulation run to stop.  Individual checks get to decide which check is "Critical", and client code gets to decide how to respond to "Critical" failures.

Member function `SatfuncConsistencyChecks<Scalar>::reportFailures()` will generate a textual report of the known set of consistency check failures at a give severity level.

As an internal implementation detail, `SatfuncConsistencyChecks` uses "reservoir sampling" (https://en.wikipedia.org/wiki/Reservoir_sampling) to track details about individual failed checks.  We maintain at most a fixed number of individual points (constructor argument).

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -183,6 +183,11 @@ list (APPEND MAIN_SOURCE_FILES
   opm/simulators/wells/WGState.cpp
   )
 
+if (HAVE_ECL_INPUT)
+  list (APPEND MAIN_SOURCE_FILES
+    opm/simulators/utils/satfunc/SatfuncConsistencyChecks.cpp
+  )
+endif()
 
 if (Damaris_FOUND AND MPI_FOUND AND USE_DAMARIS_LIB)
   list (APPEND MAIN_SOURCE_FILES
@@ -333,7 +338,10 @@ list (APPEND TEST_SOURCE_FILES
   )
 
 if (HAVE_ECL_INPUT)
-  list(APPEND TEST_SOURCE_FILES tests/test_nonnc.cpp)
+  list(APPEND TEST_SOURCE_FILES
+    tests/test_nonnc.cpp
+    tests/test_SatfuncConsistencyChecks.cpp
+  )
 endif()
 
 if(MPI_FOUND)
@@ -672,6 +680,12 @@ if (USE_BDA_BRIDGE)
     opm/simulators/linalg/bda/rocsparseWellContributions.hpp
     opm/simulators/linalg/bda/WellContributions.hpp
     opm/simulators/linalg/ISTLSolverBda.hpp
+  )
+endif()
+
+if (HAVE_ECL_INPUT)
+  list (APPEND PUBLIC_HEADER_FILES
+    opm/simulators/utils/satfunc/SatfuncConsistencyChecks.hpp
   )
 endif()
 
--- a/opm/simulators/utils/satfunc/SatfuncConsistencyChecks.cpp
+++ b/opm/simulators/utils/satfunc/SatfuncConsistencyChecks.cpp
@@ -1,0 +1,560 @@
+/*
+  Copyright 2024 Equinor AS
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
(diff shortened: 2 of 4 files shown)


bska:
I am creating this PR in draft mode because I may need to alter some of the internal workings here to add proper MPI support.  The `Check` interface might also need adjustments.

Finally, this is currently limited to builds which support `EclipseState`-like objects, but might conceivably be extensible to any configuration able to form [`EclEpsScalingPointsInfo<>`](https://github.com/OPM/opm-common/blob/15d1fc6197a68ff90813d82b02444a76d27c7d6f/opm/material/fluidmatrixinteractions/EclEpsScalingPoints.hpp#L56) objects.

bska:
jenkins build this please

CMakeLists_files.cmake
```diff
@@ -183,6 +183,11 @@ list (APPEND MAIN_SOURCE_FILES
   opm/simulators/wells/WGState.cpp
   )
 
+if (HAVE_ECL_INPUT)
+  list (APPEND MAIN_SOURCE_FILES
+    opm/core/props/satfunc/SatfuncConsistencyChecks.cpp
```
akva2, line 188: Maybe we should take this opportunity to replace core with something more descriptive like 'diagnostics' or such

bska: > Maybe we should take this opportunity to replace core with something more descriptive like 'diagnostics' or such

You mean, we'd have files named
```
opm/diagnostics/props/satfunc/ConsistencyChecks.[hc]pp
```
?  In that case we **might** need to find a new home for the other files in [`opm/core/props`](https://github.com/OPM/opm-simulators/tree/54d303ae5d40bb397a214fa45f808e4b3b94c4d8/opm/core/props), but that's probably a good idea too.

akva2: that was the idea yeah. if the rest of the core directory suffers the same fate that  is also fine indeed.

opm/core/props/satfunc/SatfuncConsistencyChecks.cpp
```diff
@@ -0,0 +1,552 @@
...
+    return *this;
+}
+
+template <typename Scalar>
+void Opm::SatfuncConsistencyChecks<Scalar>::resetCheckSet()
+{
+    this->startCheckValues_.clear();
+    this->startCheckValues_.push_back(0);
+
+    for (auto& violation : this->violations_) {
```
akva2, line 87: consider ViolationSample::clear() / reset.

CMakeLists_files.cmake
bska: > Maybe we should take this opportunity to replace core with something more descriptive like 'diagnostics' or such

I have rebased this PR on top of #5440 which moves the contents of `opm/core/props` into `opm/simulators/utils` instead.

opm/core/props/satfunc/SatfuncConsistencyChecks.cpp
bska: > consider ViolationSample::clear() / reset.

Good idea.  I've pushed an update to that effect (using `clear()`).

bska:
The MPI work appears to not need any significant API or data structure changes, so I'm marking this as "ready for review" now.  I'll run a build check.

bska:
jenkins build this please

akva2 (approved):
I've gone through this twice now and all looks good to me.

bska:
> I've gone through this twice now and all looks good to me.

Thanks a lot for taking the time to review this!  Much appreciated.  I'll merge this now as the PR's approved and the build check is green.

PR #5439, opened 2024-06-25
clean up some comments

akva2:
- simulator note was copied from the typetag dependent class (EclWriter)
- opm-output is no longer a separate module


bska (approved):
Looks good to me.  Might also want to adjust the [other reference to `opm-output`](https://github.com/OPM/opm-simulators/blob/d7d74d3a3ebf55c446a61ef9603b5b7af8c74e24/opm/simulators/flow/EclWriter.hpp#L88) in `EclWriter.hpp` too.

akva2:
jenkins build this please

PR #5440, opened 2024-06-25
Move opm/core/props to opm/simulators/utils

bska:
The `opm/core` directory is no longer meaningful, and it contains only components which might collectively be described as simulator utilities.


bska:
I think I updated all references to `opm/core` headers, but on the off chance that I did not, I'll run a full build check of all configurations to verify.

bska:
jenkins build this serial rocm hipify please

akva2 (approved):

PR #5444, opened 2024-06-26
changed: remove hipification at configure time

akva2:
the outputs will be generated when needed by the custom_command

since these outputs are not intended to be edited, there is no reason to hipify them up front. in particular this removes the long sequential process at configure time, allowing all hipification to run in parallel (if using multiple build jobs, ie. ninja -jx or make -jx)


akva2:

blattms:
Will there be a second PR where hipification is added somewhere during make? I could not see when/how it would be run, but I might be missing something.

akva2:
It is/was already there. There is a custom_command with OUTPUT for each of the files and it works just fine as is evident from the green status.

PR #5445, opened 2024-06-27
Make CPR reuse setup option only apply to CPR

jakobtorben:
The cpr-reuse-setup option is currently not limited to CPR and is being used for all preconditioners. The default option is to recreate the linear solver every 30th linear iteration. The motivation behind this option is that the AMG hierarchy depends on the values of the matrix, such that as the values of the matrix changes, the preconditioner should be updated. For preconditioners where the construction of the preconditioner only depends on the matrix structure, it is never needed to recreate the preconditioner, as the matrix structure should be constant throughout the simulation. 

I therefore suggest that we make this option only apply to CPR. This will also make the effect of the option match the name.

### DIFF
--- a/opm/simulators/linalg/ISTLSolver.hpp
+++ b/opm/simulators/linalg/ISTLSolver.hpp
@@ -495,6 +495,13 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
             if (!flexibleSolver_[activeSolverNum_].solver_) {
                 return true;
             }
+
+            if (flexibleSolver_[activeSolverNum_].pre_->hasPerfectUpdate()) {
+                return false;
+            }
+
+            // For AMG based preconditioners, the hierarchy depends on the matrix values
+            // so it is recreated at certain intervals
             if (this->parameters_[activeSolverNum_].cpr_reuse_setup_ == 0) {
                 // Always recreate solver.
                 return true;
@@ -509,7 +516,7 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
                 return this->iterations() > 10;
             }
             if (this->parameters_[activeSolverNum_].cpr_reuse_setup_ == 3) {
-                // Recreate solver if the last solve used more than 10 iterations.
+                // Never recreate the solver
                 return false;
             }
             if (this->parameters_[activeSolverNum_].cpr_reuse_setup_ == 4) {
@@ -518,7 +525,6 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
                 const bool create = ((solveCount_ % step) == 0);
                 return create;
             }
-
             // If here, we have an invalid parameter.
             const bool on_io_rank = (simulator_.gridView().comm().rank() == 0);
             std::string msg = "Invalid value: " + std::to_string(this->parameters_[activeSolverNum_].cpr_reuse_setup_)
@@ -528,7 +534,6 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
             }
             throw std::runtime_error(msg);
 
-            // Never reached.
             return false;
         }
 
--- a/opm/simulators/linalg/DILU.hpp
+++ b/opm/simulators/linalg/DILU.hpp
@@ -162,6 +162,10 @@ public:
         return SolverCategory::sequential;
     }
 
+    virtual bool hasPerfectUpdate() const override {
+        return true;
+    }
+
 private:
     //! \brief The matrix we operate on.
     const M& A_;
--- a/opm/simulators/linalg/OwningBlockPreconditioner.hpp
+++ b/opm/simulators/linalg/OwningBlockPreconditioner.hpp
@@ -74,6 +74,10 @@ public:
         orig_precond_.update();
     }
 
+    virtual bool hasPerfectUpdate() const override {
+        return orig_precond_.hasPerfectUpdate();
+    }
+
 private:
     OriginalPreconditioner orig_precond_;
     BlockPreconditioner<X, Y, Comm, OriginalPreconditioner> block_precond_;
--- a/opm/simulators/linalg/OwningTwoLevelPreconditioner.hpp
+++ b/opm/simulators/linalg/OwningTwoLevelPreconditioner.hpp
@@ -173,6 +173,10 @@ public:
         return linear_operator_.category();
     }
 
+    virtual bool hasPerfectUpdate() const override {
+        return false;
+    }
+
 private:
     using CoarseOperator = typename LevelTransferPolicy::CoarseOperator;
     using CoarseSolverPolicy = Dune::Amg::PressureSolverPolicy<CoarseOperator,
--- a/opm/simulators/linalg/ParallelOverlappingILU0.hpp
+++ b/opm/simulators/linalg/ParallelOverlappingILU0.hpp
@@ -144,6 +144,10 @@ public:
     using block_type = typename matrix_type::block_type;
     using size_type = typename matrix_type::size_type;
 
+    virtual bool hasPerfectUpdate() const override {
+        return true;
+    }
+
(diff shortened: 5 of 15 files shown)


multitalentloes:
jenkins build this please

jakobtorben:
image
 Here is a comparison on Norne where I used the ILU0 preconditioner, that only depends on the matrix structure. To the left is the simulation result when never recreating the linear solver and to the right is the result when recreating the preconditioner every linear solve. As expected, the number of timesteps, Newton iterations and linear iterations is the same.

jakobtorben:
Note that this a quick fix to make the option accurate and avoid recreating the linear solver for other preconditioners.
In a later stage we should introduce a flag that determines if a preconditioner should be recreated and see what can be resued between linear solves.

opm/simulators/linalg/ISTLSolver.hpp
```diff
@@ -495,40 +495,45 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
...
-            std::string msg = "Invalid value: " + std::to_string(this->parameters_[activeSolverNum_].cpr_reuse_setup_)
-                + " for --cpr-reuse-setup parameter, run with --help to see allowed values.";
-            if (on_io_rank) {
-                OpmLog::error(msg);
+            // For CPR/AMG based preconditioners we have the option to recreate the linear solver
+            auto linsolver = this->parameters_[activeSolverNum_].linsolver_;
+            if (linsolver == "cpr" || linsolver == "cprw" || linsolver == "cpr"||
+                linsolver == "cpr_quasiimpes" || linsolver == "cpr_trueimpes" ||
+                linsolver == "cpr_trueimpesanalytic" || linsolver == "amg" ||
+                linsolver == "hybrid") {
```
atgeirr, line 503: I think this condition is not precise enough. If you use an explicit .json file to set up your linear solver then this will not be checked, and you will never recreate the solver. I suggest instead making the opposite check: if using linsolver equal to "ilu0" or "dilu" then skip this, else do this. That would reproduce results for existing runs with .json files.

Corner case: if you then pass in a .json file specifying a preconditioner with "perfect update" such as ilu0 or dilu, and nothing else, then by default it would be needlessly recreated. However, then we have an expert user that can be expected to turn this off manually if required.

Better but more involved solution: extend the PreconditionerWithUpdate interface to provide a new method `bool hasPerfectUpdate()`, then skip this part if that returns true.

multitalentloes: I think the more involved solution extending the preconditionerWithUpdate would also make the code more consistent when maintained. Adding a new preconditioner would force the developer to find out whether or not this particular preconditioner must be reconstructed. Right now one could add a new preconditioner and be unaware that one must also update this particular if sentence in another file.

bska:
@jakobtorben : Just a quick clarification on this particular point

> For preconditioners where the construction of the preconditioner only depends on the matrix structure, it is never needed to recreate the preconditioner, as the matrix structure should be constant throughout the simulation.

Do you consider upwind directions to be a part of the matrix structure?  Those directions could change during the simulation run.

jakobtorben:
@bska I am not familiar with how the upwind directions affect the matrix structure in OPM. 
What I meant here is that certain preconditioners only depend on the matrix sparsity pattern, such that they only need to be recreated if this changes. If the upwind directions changes the sparsity pattern of a matrix, then such preconditioners should be recreated.

atgeirr:
> @bska I am not familiar with how the upwind directions affect the matrix structure in OPM.

I think they don't. For any two neighbour cells i and j, the flux between them depends on both cells' pressures, so both (i, j) and (j, i) elements of the Jacobian are nonzero. Due to upwinding, there will be some zeroes inside one of these blocks and not in the other, but the blocks as a whole are always nonzero because of the pressure dependency.

So that should not be a hindrance. However, the PR must still be changed as I requested earlier before it can be merged.

bska:
> What I meant here is that certain preconditioners only depend on the matrix sparsity pattern, such that they only need to be recreated if this changes

Right, that's what I thought.  Then this discussion probably comes down to whether the sparsity pattern comprises the **potential** non-zero matrix elements or the **actual** non-zero matrix elements.  The potential non-zero elements are static and depend only on the grid and the well/reservoir connections.  The actual non-zero elements also depend on the dynamic state and from which side of a connection we pick the property values (i.e., the upwinding).

blattms:
> What I meant here is that certain preconditioners only depend on the matrix sparsity pattern, such that they only need to be recreated if this changes
> 

But these (Jacobi SOR SSOR) are only very seldom used. All the rest actually depend on the values (ILU*, AMG, etc.)

jakobtorben:
Ok, thanks for clarifying. I will make the necessary changes to get this in now. Going for the `bool hasPerfectUpdate()` solution, that defaults to False.
@blattms The preconditioners where the update depends on the values (such as ILU, AMG) will still be updated but not recreated.

atgeirr:
> Ok, thanks for clarifying. I will make the necessary changes to get this in now. Going for the `bool hasPerfectUpdate()` solution, that defaults to False.

Sounds good. There are some things to consider along that line though:
 - This should probably be a constexpr property for each preconditioner, since it only should depend on the class, not the concrete instance. However we may then need the `virtual bool hasPerfectUpdate()` in addition to actually use it in contexts where we only have a base class pointer or reference.
 - For the wrappers that take Dune preconditioners we must think carefully about how to define the property. The RebuildOnUpdatePreconditioner should be `true`, for the DummyUpdatePreconditioner I am not 100% sure: Probably `true` here as well, as we anyway leave it to the code that employs this wrapper to only use it when an empty update() is indeed correct.

blattms:
Just a note: There is a difference between ILU and AMG here. For the former the sparsity pattern is fixed. For the latter the pattern on coarse levels actually depends on the values of the fine level.

atgeirr:
> For the former the sparsity pattern is fixed. For the latter the pattern on coarse levels actually depends on the values of the fine level.

Good point! After all, this is why we have the feature to rebuild it in the first place.

jakobtorben:
Here is a proposed solution that adds a method to `preconditionersWithUpdate`, which defines if the preconditioner has a perfect update. I opted for forcing the authors of preconditioners to define this property. For all CPR/AMG and two-level preconditioners I set it to false. There is probably a possibility of reusing more of the preconditioner/smoother in AMG, but I did not address this complexity here. I am open to other name suggestions as `hasPerfectUpdate` is not so clear.


I tested all available preconditioners in OPM, here is the result of calling `hasPerfectUpdate()`:

ilu0 true
dilu true
DuneILU true
GS true
ILUn true
Jac true
ParOverILU0 true
SOR true
SSOR true
isai true

CUDILU true
CUDILUFloat true
CUILU0 true
CUILU0Float true
CUJac true
OPMCUILU0 true

cprw false
cpr false
cprt false
cpr_quasiimpes false
cpr_trueimpes false
cpr_trueimpesanalytic false

jakobtorben:
jenkins build this please

opm/simulators/linalg/ISTLSolver.hpp
jakobtorben: See the proposed solution, which implements this.

bska:
jenkins build this please

atgeirr (approved):

atgeirr:
> I tested all available preconditioners in OPM, here is the result of calling `hasPerfectUpdate()`:

AMG is missing from the list, but I think it should be false, and that this code would indeed return false.

atgeirr:
Thanks for the work and refactoring, merging!

I you find a better name for hasPerfectUpdate() I will listen. Also, better name for update() might make it clearer what is expected.

jakobtorben:
I was not able to get AMG working as an independent preconditioner in OPM. It is used in CPR preconditioners (twolevelmethodcpr) where it is set to false.

I did not come up with any better name, so I just left it as is.

Thanks for merging @atgeirr!

PR #5449, opened 2024-06-27
Add MPI Support to Saturation Function Consistency Checks

bska:
This PR adds a new public member function
```
SatfuncConsistencyChecks<>::collectFailures(root, comm)
```
which aggregates consistency check violations from all ranks in the MPI communication object `comm` onto rank `root` of `comm`.  This amounts to summing the total number of violations from all ranks and potentially resampling the failure points for reporting purposes.

To this end, extract the body of function `processViolation()` into a general helper which performs reservoir sampling and records point IDs and which uses a call-back function to populate the check values associated to a single failed check.  Re-implement the original function in terms of this helper by wrapping `exportCheckValues()` in a lambda function.  Extract similar helpers for `numPoints()` and `anyFailedChecks()`, and add a new helper function
```
SatfuncConsistencyChecks<>::incorporateRankViolations()
```
which brings sampled points from an MPI rank into the 'root's internal data structures.

One caveat applies here.  Our current approach to collecting check failures implies that calling member function `reportFailures()` is safe only on the `root` process in a parallel run.  On the other hand functions `anyFailedChecks()` and `anyFailedCriticalChecks()` are safe, and guaranteed to return the same answer, on all MPI ranks.

On a final note, the internal helper functions are at present mostly implemented in terms of non-owning pointers.  I intend to switch to using `std::span<>` once we enable C++20 mode.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -542,6 +542,45 @@ opm_add_test(test_parallel_region_phase_pvaverage_np4
     4
 )
 
+opm_add_test(test_parallel_satfunc_consistency_checks_np2
+  EXE_NAME
+    test_SatfuncConsistencyChecks_parallel
+  CONDITION
+    MPI_FOUND AND Boost_UNIT_TEST_FRAMEWORK_FOUND
+  DRIVER_ARGS
+    -n 2
+    -b ${PROJECT_BINARY_DIR}
+  NO_COMPILE
+  PROCESSORS
+    2
+)
+
+opm_add_test(test_parallel_satfunc_consistency_checks_np3
+  EXE_NAME
+    test_SatfuncConsistencyChecks_parallel
+  CONDITION
+    MPI_FOUND AND Boost_UNIT_TEST_FRAMEWORK_FOUND
+  DRIVER_ARGS
+    -n 3
+    -b ${PROJECT_BINARY_DIR}
+  NO_COMPILE
+  PROCESSORS
+    3
+)
+
+opm_add_test(test_parallel_satfunc_consistency_checks_np4
+  EXE_NAME
+    test_SatfuncConsistencyChecks_parallel
+  CONDITION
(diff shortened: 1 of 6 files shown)


opm/simulators/utils/satfunc/SatfuncConsistencyChecks.cpp
```diff
@@ -194,9 +214,77 @@ void Opm::SatfuncConsistencyChecks<Scalar>::ViolationSample::clear()
...
+        gatherv(violation.count, comm, root);
+
+    const auto& [rankPointID, startRankPointID] =
+        gatherv(violation.pointID, comm, root);
+
+    const auto& [rankCheckValues, startRankCheckValues] =
+        gatherv(violation.checkValues, comm, root);
+
+    if (comm.rank() == root) {
+        // Re-initialise this violation sample to prepare for incorating
```
akva2, line 259: incorporating

CMakeLists.txt
```diff
@@ -542,6 +542,45 @@ opm_add_test(test_parallel_region_phase_pvaverage_np4
     4
 )
 
+opm_add_test(test_parallel_satfunc_consistency_checks_np2
```
akva2, line 545:
```cmake
foreach(NPROC 2 3 4)
opm_add_test(test_parallel_satfunc_consistency_checks_np${NPROC}
... -n ${NPROC}
..
PROCESSORS
  ${NPROC}
endforeach()
```

akva2 (approved):
One typo, one suggestion for cmake simplification.

bska:
> One typo

Good catch&ndash;thank you very much!

> one suggestion for cmake simplification.

Much appreciated, that is indeed a lot simpler.  I've pushed an update which also uses this approach to atone for similar transgressions committed earlier.

bska:
jenkins build this please

bska:

PR #5454, opened 2024-06-28
Handle empty or invalid OMP_NUM_THREADS by Flow default

vkip:
Also warn early if valid value of OMP_NUM_THREADS overrides --threads-per-process.

Motivation: Situation where an empty (but defined) OMP_NUM_THREADS caused OMP to select number of threads based on hardware information, resulting in a poor choice. Appears safer to stick with Flow default in this case.

### DIFF
--- a/opm/simulators/flow/Main.hpp
+++ b/opm/simulators/flow/Main.hpp
@@ -20,7 +20,7 @@
   You should have received a copy of the GNU General Public License
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
-#ifndef OPM_MAIN_HEADER_INCLUDED
+#ifndef OPM_MAIN_HEADER_INCLUDEDp
 #define OPM_MAIN_HEADER_INCLUDED
 
 #include <flow/flow_blackoil.hpp>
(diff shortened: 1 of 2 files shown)


blattms:
Unfortunately, this needs a rebase. What that be possible?

vkip:
> Unfortunately, this needs a rebase. What that be possible?

Of course, rebased now. Thanks for looking at this.

bska:
jenkins build this please

vkip:
jenkins build this please

opm/simulators/flow/Main.hpp
```diff
@@ -20,7 +20,7 @@
   You should have received a copy of the GNU General Public License
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
-#ifndef OPM_MAIN_HEADER_INCLUDED
+#ifndef OPM_MAIN_HEADER_INCLUDEDp
```
bska, line 23: This&ndash;the trailing `p` on `_INCLUDED`&ndash;may break something, although I can't think of anywhere we would include `Main.hpp` twice.

blattms:
We actually do Main.hpp multiple times, says the failing jenkins build.

opm/simulators/flow/Main.hpp
vkip: Sorry, sloppy fingers.

vkip:
jenkins build this please

bska (approved):

PR #5460, opened 2024-07-01
NewtonMethod: split parameters and properties

akva2:
Downstream of https://github.com/OPM/opm-models/pull/905


bska (approved):

PR #5462, opened 2024-07-01
make sure a well exists in the process before closing it due to group action

GitPaean:
Otherwise, we might try to test a well that do not exist in the process, which will cause failure when trying to retrieve information from wells_ecl_.


GitPaean:
jenkins build this please

GitPaean:
I think the PR fixes a reported high-priority issue. It should be ready for going in after reviewing.

bska (approved):
Yes, this makes sense.  The logic for identifying and stopping/shutting the "worst offending well" (i.e., `GCONPROD` action `'WELL'`) introduced in PR #5172 will consider wells across all ranks, but we can't close off-rank wells.

bska:
Note: Even though this is a correct fix, I'll defer merging this until we have an "all clear" signal that the CI system is back up and running as normal.

bska:
CI system is back up and running.  I'll merge this into master now.

PR #5467, opened 2024-07-04
Activate WELPI and WPIMULT for use in PyAction

lisajulia:
For the reference manual: With this PR and https://github.com/OPM/opm-common/pull/4136, the keywords WELPI and WPIMULT can be used in the "insert keywords"-function in Pyaction.

### DIFF
--- a/opm/simulators/flow/ActionHandler.cpp
+++ b/opm/simulators/flow/ActionHandler.cpp
@@ -140,30 +140,23 @@ namespace {
     fetchWellPI(const int                                    reportStep,
                 const Opm::Schedule&                         schedule,
                 const WellModel&                             wellModel,
-                const Opm::Action::ActionX&                  action,
-                const Opm::Action::Result::MatchingEntities& matches,
                 const Opm::Parallel::Communication           comm)
     {
         auto wellpi = std::unordered_map<std::string, Scalar> {};
 
-        const auto wellpi_wells = action.wellpi_wells
-            (schedule.wellMatcher(reportStep), matches);
+        auto allWells = schedule[reportStep].well_order().names();
 
-        if (wellpi_wells.empty()) {
-            return wellpi;
-        }
-
-        auto wellPI = std::vector<Scalar>(wellpi_wells.size());
-        for (auto i = 0*wellpi_wells.size(); i < wellpi_wells.size(); ++i) {
-            if (wellModel.hasWell(wellpi_wells[i])) {
-                wellPI[i] = wellModel.wellPI(wellpi_wells[i]);
+        auto wellPI = std::vector<Scalar>(allWells.size());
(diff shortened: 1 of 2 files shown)


opm/simulators/flow/ActionHandler.cpp
```diff
@@ -146,6 +146,12 @@ applyActions(const int reportStep,
     const auto ts  = formatActionDate(now, reportStep);
 
     bool commit_wellstate = false;
+
+    if (actions.pending_python(actionState_).size() > 0) {
+        //Only fetch this map if we have pending pyactions
+        auto wellPIAll = this->fetchWellPIAll(reportStep);
```
blattms, line 152: Why do a copy here?

opm/simulators/flow/ActionHandler.cpp
```diff
@@ -146,6 +146,12 @@ applyActions(const int reportStep,
...
+        schedule_.wellPIMap = std::make_shared<std::unordered_map<std::string, double>>(std::move(wellPIAll));
```
blattms, line 153: Wouldn't hiding this in a method of Schedule be better?
```
shedule.resetWellPIMap( this->fetchWellPIAll(reportStep));
```

opm/simulators/flow/ActionHandler.hpp
```diff
@@ -81,10 +81,16 @@ class ActionHandler
                               const TransFunc& updateTrans);
 
     std::unordered_map<std::string, Scalar>
-    fetchWellPI(int reportStep,
+    fetchWellPIMatching(int reportStep,
```
blattms, line 84: Please fix indentation,

opm/simulators/flow/ActionHandler.hpp
```diff
@@ -81,10 +81,16 @@ class ActionHandler
...
                 const Action::ActionX& action,
                 const std::vector<std::string>& matching_wells) const;
 
+    std::unordered_map<std::string, Scalar>
+    fetchWellPIAll(int reportStep) const;
+
+    std::unordered_map<std::string, Scalar>
+    _fetchWellPI(int reportStep, const std::vector<std::string>& matching_wells) const;
```
blattms, line 92: Why the underline? Seems a bit confusing.

opm/simulators/flow/ActionHandler.cpp
blattms: Will this member be reset later?

blattms:
I am not sure how the performance implications of this are. It seems like we now always get all the WELPI of all wells even without any PYACTION. This is more information/computation then we need. Do you see a possibility to get only the needed ones if there is no PYACTION?

I think getting them at once is an improvement.

blattms:
Should we reset the welpimap in the schedule somewhen?

lisajulia:
Closing as the alternative was merged

PR #5469, opened 2024-07-04
using perf rates for b factor calculation

GitPaean:
when calculating the connection pressure drop.

It is still posted here for information.


GitPaean:
Now I think the idea for this PR is not logically plausible.  So I close it now.

PR #5473, opened 2024-07-09
Avoid buffers exceeding intmax in MPI calls

vkip:
Depends on https://github.com/OPM/opm-common/pull/4139

### DIFF
--- a/opm/simulators/utils/MPIPacker.cpp
+++ b/opm/simulators/utils/MPIPacker.cpp
@@ -70,7 +70,7 @@ packSize(const std::string& data, Parallel::MPIComm comm)
     MPI_Pack_size(1, Dune::MPITraits<std::size_t>::getType(), comm, &size);
     int totalSize = size;
     MPI_Pack_size(data.size(), MPI_CHAR, comm, &size);
-    return totalSize + size;
+    return static_cast<std::size_t>(totalSize + size);
 }
 
 void Packing<false,std::string>::
@@ -80,7 +80,7 @@ pack(const std::string& data,
      Parallel::MPIComm comm)
 {
     std::size_t length = data.size();
-    int int_position;
+    int int_position = 0;
     MPI_Pack(&length, 1, Dune::MPITraits<std::size_t>::getType(), buffer.data()+position,
              mpi_buffer_size(buffer.size(), position), &int_position, comm);
     MPI_Pack(data.data(), length, MPI_CHAR, buffer.data()+position, mpi_buffer_size(buffer.size(), position),
@@ -95,7 +95,7 @@ unpack(std::string& data,
        Opm::Parallel::MPIComm comm)
 {
     std::size_t length = 0;
-    int int_position;
+    int int_position = 0;
     MPI_Unpack(buffer.data()+position, mpi_buffer_size(buffer.size(), position), &int_position, &length, 1,
                Dune::MPITraits<std::size_t>::getType(), comm);
     std::vector<char> cStr(length+1, '\0');
--- a/opm/simulators/utils/MPIPacker.cpp
+++ b/opm/simulators/utils/MPIPacker.cpp
@@ -70,7 +70,7 @@ packSize(const std::string& data, Parallel::MPIComm comm)
     MPI_Pack_size(1, Dune::MPITraits<std::size_t>::getType(), comm, &size);
     int totalSize = size;
     MPI_Pack_size(data.size(), MPI_CHAR, comm, &size);
-    return static_cast<std::size_t>(totalSize + size);
+    return totalSize + size;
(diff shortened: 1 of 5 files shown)


opm/simulators/utils/MPIPacker.hpp
```diff
@@ -26,19 +26,27 @@
...
 
+
 namespace Opm {
 namespace Mpi {
 namespace detail {
 
+static std::size_t mpi_buffer_size(const std::size_t bufsize, const std::size_t position) {
+    assert (bufsize >= position);
+    return static_cast<int>(std::min(bufsize-position,
+                                     static_cast<std::size_t>(std::numeric_limits<int>::max())));
```
blattms, line 40: I think we should rather throw an exception if `bufsize-position > static_cast<std::size_t>(std::numeric_limits<int>::max())` holds.

opm/simulators/utils/MPIPacker.hpp
```diff
@@ -59,9 +67,14 @@ struct Packing<true,T>
     //! \param comm The communicator to use
     static std::size_t packSize(const T*, std::size_t n, Parallel::MPIComm comm)
     {
+        // For now we do not handle the situation where a a single call to packSize/pack/unpack
+        // is likely to require an MPI_Pack_size value larger than intmax
+        assert ( n*sizeof(T) <= std::numeric_limits<int>::max() );
```
blattms, line 72: Here we should probably throw an exception instead. Otherwise this will go undetected if we compile with `-DNDEBUG` (not default).

opm/simulators/utils/MPIPacker.hpp
```diff
@@ -59,9 +67,14 @@ struct Packing<true,T>
...
         int size = 0;
         MPI_Pack_size(n, Dune::MPITraits<T>::getType(), comm, &size);
-        return size;
+        assert (size >= 0);
+        assert (size < std::numeric_limits<int>::max() );
```
blattms, line 76: I think we can skip this assert as the type is an int.

opm/simulators/utils/MPISerializer.hpp
```diff
@@ -52,7 +52,15 @@ class MpiSerializer : public Serializer<Mpi::Packer> {
...
-                m_comm.broadcast(m_buffer.data(), m_packSize, root);
+                const int maxChunkSize = std::numeric_limits<int>::max();
+                std::size_t remainingSize = m_packSize;
+                std::size_t pos = 0;
+                while (remainingSize > maxChunkSize) {
+                    m_comm.broadcast(m_buffer.data()+pos, maxChunkSize, root);
+                    pos += maxChunkSize;
+                    remainingSize -= maxChunkSize;
+                }
+                m_comm.broadcast(m_buffer.data()+pos, static_cast<int>(remainingSize), root);
```
blattms, line 63: Maybe put this into a separate function for reuse bwlow?

opm/simulators/utils/MPIPacker.cpp
```diff
@@ -70,34 +70,38 @@ packSize(const std::string& data, Parallel::MPIComm comm)
     MPI_Pack_size(1, Dune::MPITraits<std::size_t>::getType(), comm, &size);
     int totalSize = size;
     MPI_Pack_size(data.size(), MPI_CHAR, comm, &size);
-    return totalSize + size;
+    return static_cast<std::size_t>(totalSize + size);
```
blattms, line 73: The explicit cast does not hurt, but there should be an implicit one for the original code anyway.

blattms (commented):
This looks mostly good and indeed should fix the problem.
I have just added a few comments that might improve the PR.

opm/simulators/utils/MPIPacker.hpp
vkip: Actually I believe we need to allow bufsize-position > intmax, we limit it to the latter only to avoid problems in MPI_Pack, but the buffer is actually larger. Of course, trying to pack more than intmax bytes in one go will fail, but this should be picked up in the packSize call (below). 

Will change the assert to throw though.

opm/simulators/utils/MPIPacker.hpp
vkip: Good point, will update.

opm/simulators/utils/MPIPacker.hpp
vkip: Certainly!

opm/simulators/utils/MPISerializer.hpp
vkip: Good point, will do.

vkip:
jenkins build this opm-common=4139 please

PR #5475, opened 2024-07-11
Move documentation

lisajulia:
Move documentation to https://github.com/opm/opm-python-documentation such that the build of the documentation does not interfere with the actual code, i.e. in case the documentation can't be build, the pipeline of the documentation repository will fail but not here.

This will then replace #5349, since I have moved the changes from #5349 to the new repository.

Closes #5349


lisajulia:
@blattms: can you have a look here?

blattms:
Brilliant idea. As far as I can tell (not far...) this looks good.

blattms:
What about #5243?

PR #5476, opened 2024-07-11
Set correct target iterations in SimpleIterationTimestepControl

jakobtorben:
For the simple iteration timestep controller, used for both the "iterationcount" and "newtoniterationcount" timestep controller, we are currently subtracting 1 from the target iterations. This is not correct and is misleading from the time-step-control-target-newton-iterations option. For example, if one set the target to 2 Newton iterations, which works well for NLDD, the timestep will not change.

### DIFF
--- a/opm/simulators/timestepping/TimeStepControl.cpp
+++ b/opm/simulators/timestepping/TimeStepControl.cpp
@@ -84,7 +84,7 @@ namespace Opm
             dtEstimate *= decayrate_;
         }
         // increase the time step size if we are below the number of target iterations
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

atgeirr:
This should be merged, but I do not understand the regression failure here. The DAY field should be no more than 20 since October 20 is the final day of the case, yet we see 28 in the summary. And the simulation output also ends on the 20th.

jakobtorben:
jenkins build this please

jakobtorben:
@atgeirr Now the regression tests pass

PR #5477, opened 2024-07-11
Add the metis partitioner as a command line option as well

lisajulia:
Depends on https://github.com/OPM/opm-grid/pull/736

### DIFF
--- a/opm/simulators/flow/FlowBaseVanguard.hpp
+++ b/opm/simulators/flow/FlowBaseVanguard.hpp
@@ -106,10 +106,21 @@ struct OwnerCellsFirst {
 };
 
 template<class TypeTag, class MyTypeTag>
+struct PartitionMethod {
+    using type = UndefinedProperty;
+};
+
+template<class TypeTag, class MyTypeTag>
 struct SerialPartitioning {
     using type = UndefinedProperty;
 };
 
+template<class TypeTag, class MyTypeTag>
+struct ImbalanceTol {
+    using type = UndefinedProperty;
+};
+
+// Remove this for release 2025.04
 template<class TypeTag, class MyTypeTag>
 struct ZoltanImbalanceTol {
     using type = UndefinedProperty;
@@ -117,6 +128,11 @@ struct ZoltanImbalanceTol {
 
 template<class TypeTag, class MyTypeTag>
 struct ZoltanParams {
+    using type = UndefinedProperty;
+};
+
+template<class TypeTag, class MyTypeTag>
+struct MetisParams {
     using type = UndefinedProperty;
 };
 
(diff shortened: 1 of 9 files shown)


opm/simulators/flow/FlowBaseVanguard.hpp
```diff
@@ -190,6 +211,20 @@ struct ZoltanParams<TypeTag,TTag::FlowBaseVanguard> {
     static constexpr auto value = "graph";
 };
 
+template<class TypeTag>
+struct MetisImbalanceTol<TypeTag, TTag::FlowBaseVanguard> {
+#if defined(IS_SCOTCH_METIS_HEADER) && IS_SCOTCH_METIS_HEADER
+    static constexpr double value = 0.1;
+#else
+    static constexpr double value = 1.1;
+#endif
```
blattms, line 220: Please add a comment why these are different.

opm/simulators/flow/FlowBaseVanguard.hpp
```diff
@@ -120,6 +125,16 @@ struct ZoltanParams {
     using type = UndefinedProperty;
 };
 
+template<class TypeTag, class MyTypeTag>
+struct MetisImbalanceTol {
+    using type = UndefinedProperty;
+};
+
```
blattms, line 132: Do we really need a different one? Maybe we can have on to rule them all and deprecate the zoltan one?

opm/simulators/flow/FlowBaseVanguard.hpp
```diff
@@ -292,6 +329,19 @@ class FlowBaseVanguard : public BaseVanguard<TypeTag>,
              "See https://sandialabs.github.io/Zoltan/ug_html/ug.html "
              "for available Zoltan options.");
         Parameters::hideParam<TypeTag, Properties::ZoltanParams>();
+        Parameters::registerParam<TypeTag, Properties::MetisImbalanceTol>
+#if defined(IS_SCOTCH_METIS_HEADER) && IS_SCOTCH_METIS_HEADER
+            ("Tolerable imbalance of the loadbalancing provided by Scotch-METIS (default: 0.1).");
+#else
+            ("Tolerable imbalance of the loadbalancing provided by Metis (default: 1.1).");
+#endif
```
blattms, line 337: Would be cool, if we could go without #if #else here specify both defaults in one help string

blattms (commented):
I think there should only be one tolarable imbalance param and we should make sure that the same value behaves the same for all loadbalancers (by computing and adjusted one for them).

Changing parameteres when switching partitioners seems rather cumbersome,

lisajulia:
jenkins build this opm-grid=736 please

lisajulia:
jenkins build this opm-grid=736 please

lisajulia:
jenkins build this opm-grid=736 please

lisajulia:
jenkins build this opm-grid=736 please

blattms:
BTW if we trigger jenkins in the upstream repo, then also the tests will be run there. I think that they are omitted if we do it here.

lisajulia:
> BTW if we trigger jenkins in the upstream repo, then also the tests will be run there. I think that they are omitted if we do it here.

ok thanks :) then I will trigger the next build in opm-grid

blattms (approved):

PR #5478, opened 2024-07-12
Use the failure flag of the tasklet runner to detect a failure while …

lisajulia:
…writing output and throw an error if that happens.


lisajulia:
jenkins build this opm-models =909 please

lisajulia:
@bska: Can you have a look here? Thanks!

lisajulia:
jenkins build this please

bska (approved):
This is good and I'll merge into master.  We do need to take the next step of guarding the call to `doWriteOutput()` at some point, but that can be follow-up work.

PR #3169, opened 2021-04-21
Initialize empty producers using mobility and transmissibility ratio 

totto82:
The approach was suggested by @jalvestad 
And initial testing on norne and model2 is promising.

PR #5479, opened 2024-07-12
Revise Mixture Density Method for No-Flow Producers

bska:
This PR switches the approach introduced in commit eeb1b7e36c9ae3c30ba5d262fa623dc0eea2e10b (PR #3169) to using a mobility weighted average of cell level densities for the connection level mixture densities in no-flow producing wells.  We also use the recent `stoppedOrZeroRateTarget()` predicate to identify those no-flow producing wells instead of inspecting the connection flow rates.

The mobility weighted average gives a more monotone pressure buildup for the stopped wells and this is usually what the engineer wants. This revised approach furthermore needs fewer cell-level dynamic properties so simplify the `computeProperties()` signature by introducing a structure for the property callback functions and update the callers accordingly.

While here, also refactor parts of the `StandardWellConnections` code to aid future maintenance, mostly by splitting out parts of the existing `computePropertiesForPressures()` member function to several other helper functions.

### DIFF
--- a/opm/simulators/wells/StandardWellConnections.cpp
+++ b/opm/simulators/wells/StandardWellConnections.cpp
@@ -419,6 +419,44 @@ initialiseConnectionMixture(const int                  num_comp,
     }
 }
 
+template <class FluidSystem, class Indices>
+void StandardWellConnections<FluidSystem, Indices>::
+computeDensitiesForStoppedProducer(const DensityPropertyFunctions& prop_func)
+{
+    const auto np = this->well_.numPhases();
+
+    const auto modPhIx = [this, np]() {
+        auto phIx = std::vector<int>(np);
+
+        for (auto p = 0*np; p < np; ++p) {
+            phIx[p] = this->well_.flowPhaseToModelPhaseIdx(p);
+        }
+
+        return phIx;
+    }();
+
+    auto mob = std::vector<Scalar>(np);
+    auto rho = std::vector<Scalar>(np);
+
+    const auto nperf = this->well_.numPerfs();
+    for (auto perf = 0*nperf; perf < nperf; ++perf) {
+        const auto cell_idx = this->well_.cells()[perf];
+
+        prop_func.mobility     (cell_idx, modPhIx, mob);
+        prop_func.densityInCell(cell_idx, modPhIx, rho);
+
+        auto& rho_i = this->perf_densities_[perf];
+
+        auto mt = rho_i = Scalar{0};
+        for (auto p = 0*np; p < np; ++p) {
+            rho_i += mob[p] * rho[p];
+            mt    += mob[p];
+        }
+
+        rho_i /= mt;
+    }
+}
+
 template<class FluidSystem, class Indices>
 typename StandardWellConnections<FluidSystem, Indices>::Properties
 StandardWellConnections<FluidSystem,Indices>::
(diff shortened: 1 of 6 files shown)


bska:
jenkins build this please

bska:
I've gone through the regression test failures and most appear to be triggered by (slightly) different time steps.  It's not clear to me, however, whether the solutions generated by the current master sources are "better" than those generated by this PR.  For what it's worth, this PR does seem to give more stable WBHP behaviour in the shut-in period [prior to the 1st of April 1999](https://github.com/OPM/opm-tests/blob/1e7654ccd7683fe2b9921a0d918edf87c8dc278f/norne/INCLUDE/B1H_RE-PERF.SCH#L106-L113) in `NORNE_ATW2013_B1H_REPERF` than the current master sources.  PR #4817 also [discussed this stability issue](https://github.com/OPM/opm-simulators/pull/4817#issuecomment-1697380989).

---

B_1H_WBHP

tskille:
The PR has been tested on the test model and also on a field model. The results are excellent and I'm looking forward to see this going into master.

bska:
> The PR has been tested on the test model and also on a field model. The results are excellent and I'm looking forward to see this going into master.

Thanks a lot for testing the PR&ndash;that is greatly appreciated.  We have a few work items that must be completed before we're ready to merge this, but we will prioritise getting this work into the master branch once those have been completed.

opm/simulators/wells/StandardWellConnections.cpp
```diff
@@ -95,212 +207,257 @@ computeDensities(const std::vector<Scalar>& perfComponentRates,
...
-                // This is the bottom perforation. No flow from below.
-                global_q_out_perf[index] = 0.0;
-            } else {
-                // Set equal to flow from below.
-                global_q_out_perf[index] = global_q_out_perf[index + num_comp];
-            }
+            const auto index = perf*num_comp + component;
+            auto& q_out = global_q_out_perf[index];
+
+            q_out = (perf == factory.numGlobalPerfs() - 1)
```
atgeirr, line 311: The comments are lost here, I find them useful in understanding what is going on so they should be reinstated.

opm/simulators/wells/StandardWellConnections.cpp
```diff
@@ -95,212 +207,257 @@ computeDensities(const std::vector<Scalar>& perfComponentRates,
...
-                    // Subtract gas in oil from gas mixture
-                    x[gaspos] = (mix[gaspos] - mix[oilpos]*rs)/d;
-                }
-                if (rv > 0.0) {
-                    // Subtract oil in gas from oil mixture
-                    x[oilpos] = (mix[oilpos] - mix[gaspos]*rv)/d;
-                }
+            else {
+                // For the rest of the perforations without flow we use the
+                // previousMixtureture from the perforation above.
```
atgeirr, line 401: Typo.

opm/simulators/wells/StandardWellConnections.cpp
```diff
@@ -95,212 +207,257 @@ computeDensities(const std::vector<Scalar>& perfComponentRates,
...
+                previousMixture = currentMixture;
```
atgeirr, line 402: I do not understand the previous/current naming. The original names mix and x were not good, but I do not see any relation to time or changing state here that the terms previous or current imply. What about surfVolMix and volMix for example, maybe that would be clearer?

opm/simulators/wells/StandardWellConnections.cpp
```diff
@@ -33,13 +33,125 @@
...
+                                    const typename std::vector<Scalar>::size_type waterpos,
+                                    const typename std::vector<Scalar>::size_type perf,
+                                    const Properties&                             props,
+                                    const std::vector<Scalar>&                    previousMixture,
+                                    std::vector<Scalar>&                          currentMixture,
+                                    Opm::DeferredLogger&                          deferred_logger)
+    {
+        // matrix system: (mix[oilpos] = q_os, x[oilpos] = bo*q_or, etc...)
+        // ┌             ┐   ┌                ┐  ┌           ┐
+        // │mix[oilpos]  │   | 1     Rv     0 |  |x[oilpos]  |
```
atgeirr, line 113: This comment is outdated with new names for the variables. Also, it applies to both this and the above function, maybe create an explanatory comment above both, explaining them in light of this equation?

atgeirr (commented):
A few issues, should be merged when addressed.

GitPaean:
It is great this PR fixes certain reported cases. 

My main comment is that, it is a little unfortunate that different formulation is used for exact zero-rate control wells. 

Is it possible that the same formula can actually be used for other normally running wells? If it is true, then it might means it does not change the results significantly enough so that is why we have been using different formula for normally running wells. 

As part of investigation, how about the wells with non-zero while very small rate? for example, 0.1 m^3 / day for oil production. It will use the original formula in the master branch, but theoretically, it might should be similar to the zero rate controlled wells?

opm/simulators/wells/StandardWellConnections.cpp
bska: Most of the original comments remain in the [`computeDensities()`](https://github.com/OPM/opm-simulators/blob/853fda9d4207e1dd36bf95b943490352141b105c/opm/simulators/wells/StandardWellConnections.cpp#L209) member function from which this function was extracted.  That said, I've pushed an update to annotate some of the statements here.

opm/simulators/wells/StandardWellConnections.cpp
bska: > I do not understand the previous/current naming

Current and previous were meant to refer to the the current and previous perforations in the index sequence `0..nperf-1`.  That said, I also wasn't particularly happy with that naming convention so I've pushed an update that uses `componentMixture` and `phaseMixture` instead.

opm/simulators/wells/StandardWellConnections.cpp
bska: I've pushed an update to move this comment block above `reapportion*Mixture()` helper functions, including to switch to the `componentMixture` and `phaseMixture` naming.

atgeirr (approved):
I am happy with the revision, and this can be merged.

I am curious about the answer to @GitPaean's question though, it could point to some future refinements here. If I understand his concern, the question is if we can have continuous behaviour between the zero and non-zero flux cases, and that seems like a good idea to me. I will not require it for this PR though.

bska:
jenkins build this please

bska:
jenkins build this please

atgeirr:
jenkins build this please

bska:
> jenkins

Yeah...  The current PR breaks a number regression tests.  I've been investigating the failures to understand whether or not this introduces any unforeseen consequences and there does seem to be some impact on the timestep selection algorithm.  If I impose a maximum timestep of one day, e.g., through `--solver-max-time-step-in-days=1` I can make this PR match the solutions generated by the current master branch.  On the other hand, with unrestricted timestep selection we do get differences.

I **think** I'd like to generate new reference solutions and merge this, but I'm open to other points of view.

atgeirr:
> If I impose a maximum timestep of one day, e.g., through `--solver-max-time-step-in-days=1` I can make this PR match the solutions generated by the current master branch

I think this treatment should be done on a case-by-case basis and not universally, as this would make test times grow significantly.

> I **think** I'd like to generate new reference solutions and merge this, but I'm open to other points of view.

I tentatively agree with this. The fact that with one-day timesteps you get a match with the master branch indicates that the changes here do no harm. Have you also verified that the behaviour of the failing tests is reasonable?

GitPaean:
> If I impose a maximum timestep of one day, e.g., through --solver-max-time-step-in-days=1 I can make this PR match the solutions generated by the current master branch. 

I think it is generally fine in my opinion. We had to do this for numerous occasions.  Main concern is whether this will increase the running time of the cases to be much longer.  If the regression only happens to the first time steps, maybe can enforce the 1-day (or other time step sizes that help) for the relevant time steps either by modifying DATA file or using tuning.  With this way, we might gradually stablize many of the regression test  cases. 

I am not opposed to updating the reference and get the PR merged either if the failures are clearly due to time stepping changes and no behavoir changes occurred.

bska:
> > If I impose a maximum timestep of one day, e.g., through `--solver-max-time-step-in-days=1` I can make this PR match the solutions generated by the current master branch
> 
> I think this treatment should be done on a case-by-case basis and not universally, as this would make test times grow significantly.

Sorry for not being clear.  I did not mean to suggest that we alter the test parameters.  Rather, I intended (just) to state the fact that if I restrict to the maximum timestep size to one day in both the master sources and this PR, then I can get both versions of the simulator to generate the same&ndash;in eyeball norm&ndash;numerical solutions.

> > I **think** I'd like to generate new reference solutions and merge this, but I'm open to other points of view.
> 
> [...] Have you also verified that the behaviour of the failing tests is reasonable?

Yes.  Both the master sources and the current PR capture the same "trends" with unrestricted timestep selection, albeit with a small caveat for the cases that involve dynamically triggered action blocks.

> I am not opposed to updating the reference and get the PR merged either if the failures are clearly due to time stepping changes and no behavior changes occurred.

Cool.  I do indeed think that the differences are related to timestepping differences.  That said, I would of course prefer that the timestep selection were unaffected by this work...

atgeirr:
It seems we agree on a data update here. I will start the Jenkins process.

atgeirr:

bska:
jenkins build this opm-tests=1215 please

bska:
jenkins build this opm-tests=1215 please

bska:
The PR's approved, the build check is green and the new reference solutions have been installed on the CI system.  I'm merging this now to activate the new procedure.

PR #5483, opened 2024-07-17
Output support for RSWSAT and RVWSAT

daavid00:
Depends on https://github.com/OPM/opm-common/pull/4142


bska (approved):

PR #5486, opened 2024-07-25
Fix/actionx compdat show error

lisajulia:

PR #5488, opened 2024-07-25
Fix/actionx compdat

lisajulia:
Fix error when using the keyword COMPDAT in ACTIONX: For parallel runs, the connections opened in ACTIONX are unknown to the grid partitioner, I've added a test comparing two runs (one with actionx and one without) which fails for the current master (https://github.com/OPM/opm-simulators/pull/5486) and also a fix which makes this test pass.

This PR depends on https://github.com/OPM/opm-tests/pull/1199, https://github.com/OPM/opm-grid/pull/742 and https://github.com/OPM/opm-common/pull/4150 and neither of these PRs affect the reference manual.

### DIFF
--- a/opm/simulators/linalg/ISTLSolverBda.cpp
+++ b/opm/simulators/linalg/ISTLSolverBda.cpp
@@ -78,12 +78,14 @@ void BdaSolverInfo<Matrix,Vector>::
 prepare(const Grid& grid,
         const Dune::CartesianIndexMapper<Grid>& cartMapper,
         const std::vector<Well>& wellsForConn,
+        const std::unordered_map<std::string, std::set<std::array<int,3>>>& possibleFutureConnections,
         const std::vector<int>& cellPartition,
         const std::size_t nonzeroes,
         const bool useWellConn)
 {
     if (numJacobiBlocks_ > 1) {
       detail::setWellConnections(grid, cartMapper, wellsForConn,
+                                 possibleFutureConnections,
                                  useWellConn,
                                  wellConnectionsGraph_,
                                  numJacobiBlocks_);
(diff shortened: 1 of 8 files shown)


lisajulia:
jenkins build this opm-tests=1199 opm-grid=742 opm-common=4150 please

lisajulia:
jenkins build this opm-tests=1199 opm-grid=742 opm-common=4150 please

lisajulia:
jenkins build this opm-tests=1199 opm-grid=742 opm-common=4150 please

opm/simulators/linalg/ISTLSolverBda.cpp
```diff
@@ -78,12 +78,15 @@ void BdaSolverInfo<Matrix,Vector>::
 prepare(const Grid& grid,
         const Dune::CartesianIndexMapper<Grid>& cartMapper,
         const std::vector<Well>& wellsForConn,
+        //const std::unordered_map<std::string, std::set<std::array<int,3>>>& possibleFutureConnections,
         const std::vector<int>& cellPartition,
         const std::size_t nonzeroes,
         const bool useWellConn)
 {
+    std::unordered_map<std::string, std::set<std::array<int,3>>> possibleFutureConnections;
```
blattms, line 86: I think we should fix this. Currently, we just silently assume that no extra connections will be defined via ACTIONX.

Hopefully, we can get the futureConnections at the [calling site ISTLSolverBda.hpp#L207](https://github.com/OPM/opm-simulators/blob/9f76155312d1c5f96c0be15bcd9205854c0251a0/opm/simulators/linalg/ISTLSolverBda.hpp#L207) and pass it as an argument.

blattms (changes requested):
Looks good. I have just two suggestions for improvement.

blattms:
Once this is fixed, please trigger a test in the opm-common PR. That will also run the tests there.

blattms:
According to @atgeirr we also need to take the future connections into account in [BlackoilModelNldd.hpp#L944](https://github.com/OPM/opm-simulators/blob/125ee003363454f1fa0e30bbd0a0b20c14cd503a/opm/simulators/flow/BlackoilModelNldd.hpp#L944-L958) and files opm/simulators/utils/ParallelNLDDPartitioningZoltan.*pp

lisajulia:
> According to @atgeirr we also need to take the future connections into account in [BlackoilModelNldd.hpp#L944](https://github.com/OPM/opm-simulators/blob/125ee003363454f1fa0e30bbd0a0b20c14cd503a/opm/simulators/flow/BlackoilModelNldd.hpp#L944-L958) and files opm/simulators/utils/ParallelNLDDPartitioningZoltan.*pp

Yes, I can do that, but can we still merge this one and https://github.com/OPM/opm-tests/pull/1199, https://github.com/OPM/opm-grid/pull/742 and https://github.com/OPM/opm-common/pull/4150  already? I had a brief look and I think the same approach words for NLDD and I would then open a new PR for that.

atgeirr:
> I had a brief look and I think the same approach words for NLDD and I would then open a new PR for that.

That sounds good! It is fine to do this in two steps.

blattms:
This definitely make #5494 disappear. Tested on  my system.

Hence, many thanks for finding and fixing this rather severe problem.

blattms (approved):

PR #5496, opened 2024-07-31
Write vtk (if enabled) for dry runs

daavid00:
With this PR, the .vtu is written when `--enable-vtk-output=1 --enable-dry-run=1`


daavid00:
jenkins build this please

PR #5501, opened 2024-08-02
Add flow executable with simpler MaterialLawManager

multitalentloes:
This PR adds a flow executable that will use the simple EclMaterialLawManager in opm-common to enable GPU support of property evaluation on simple flow cases


GitPaean:
I know it is still under draft, while a small suggestion, maybe a more descriptive name can be better.

GitPaean:
Thanks for changing the title, but I meant the name of the executable `flow_simple`.   Hopefully the name can give people some hints on what this executable is doing.

PR #5505, opened 2024-08-05
Fix the usage of the COMPDAT keyword in an ACTIONX for the NLDD solver for sequential runs

lisajulia:
This PR fixes a simulation runs with well connections that are added through an ACTIONX with the COMPDAT keyword and these connections need to be taken into account by the NLDD solver.
Without this fix, the simulator stops with an error (https://ci.opm-project.org/job/opm-simulators-PR-builder/6462/). This PR fixes sequential runs, there is a [separate PR](https://github.com/OPM/opm-simulators/pull/5510) for parallel runs.

This PR depends on https://github.com/OPM/opm-common/pull/4159 and https://github.com/OPM/opm-grid/pull/747.

I've triggered the tests in https://github.com/OPM/opm-common/pull/4159.

### DIFF
--- a/opm/simulators/flow/BlackoilModelNldd.hpp
+++ b/opm/simulators/flow/BlackoilModelNldd.hpp
@@ -953,9 +953,10 @@ private:
             ? param.num_local_domains_
             : num_cells / default_cells_per_domain;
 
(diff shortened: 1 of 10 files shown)


blattms (commented):
Looks good to me. I am just wondering whether we can get rid off the pointer usage,

PR #5508, opened 2024-08-06
Reveal error in g2l map in partitionCells for the NLDD solver

lisajulia:
I'm looking into a bug happening when using the NLDD solver.
The actual bug happens in situations where well connections are added through an ACTIONX with the COMPDAT keyword and need to be taken into account by the solver and the grid partitioner and I plan to fix it in PR #5505.

My current problem even occurs when running the simulator with a deck without an ACTIONX, i.e. with this deck (https://github.com/OPM/opm-tests/blob/master/actionx/COMPDAT_SHORT.DATA):
There is a global2local (g2l) map (created in the function connectElements of the ZoltanPartitioner in [partitionCells.cpp](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/flow/partitionCells.cpp#L250)) which does not seem to contain the correct elements when running the simulation in parallel, or the elements that are queried on each process are not right.
I've also added a failing test for this:
https://ci.opm-project.org/job/opm-simulators-PR-builder/6469/testReport/junit/(root)/mpi/compareParallelSim_flow_COMPDAT_SHORT/
Maybe the error went unnoticed until now, see this comment: https://github.com/OPM/opm-simulators/pull/5508/files#diff-8e157d0d44a0cf2ae336f54feae615c8007dfa70b7d065b0311a0f25d5971625R252

Am I doing something wrong here or why does that happen?

### DIFF
--- a/opm/simulators/flow/partitionCells.cpp
+++ b/opm/simulators/flow/partitionCells.cpp
@@ -249,7 +249,9 @@ ZoltanPartitioner::connectElements(const GridView&                              
 {
     auto g2l = std::unordered_map<int, int>{};
 
-    for (const auto& element : elements(grid_view, Dune::Partitions::interior)) {
+    // For Dune::Partitions::interior, even more global indices are not found in the connectWells functions,
+    // yet the std::invalid_argument at the end of connectWells is not thrown...
+    for (const auto& element : elements(grid_view, Dune::Partitions::all)) {
         {
             const auto c = zoltan_ctrl.index(element);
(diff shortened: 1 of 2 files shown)


lisajulia:
jenkins build this please

lisajulia:
closing this, since I think the error comes from somewhere else

lisajulia:
> closing this, since I think the error comes from somewhere else

I've found the error, the fix is here: https://github.com/OPM/opm-simulators/pull/5510. The g2l map is correct, the error was somewhere different.

PR #5510, opened 2024-08-06
Fix/compdat nldd parallel

lisajulia:
This PR fixes the NLDD solver for parallel runs with the COMPDAT keyword inside an ACTIONX.

### DIFF
--- a/opm/simulators/flow/BlackoilModelNldd.hpp
+++ b/opm/simulators/flow/BlackoilModelNldd.hpp
@@ -945,6 +945,9 @@ private:
         const auto wells = need_wells
             ? this->model_.simulator().vanguard().schedule().getWellsatEnd()
             : std::vector<Well>{};
+        const auto& possibleFutureConnectionSet = need_wells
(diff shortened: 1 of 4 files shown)


lisajulia:
jenkins build this please

lisajulia:
jenkins build this please

akva2 (approved):
There are a little too many deep details for me to investigate this right now but the usage of the serializer is certainly as intended.

Do we really need to add tests with 8 procs, what does that give that 4, the number of physical cores dedicated to a ci job, would not cover?

lisajulia:
> There are a little too many deep details for me to investigate this right now but the usage of the serializer is certainly as intended.
> 
> Do we really need to add tests with 8 procs, what does that give that 4, the number of physical cores dedicated to a ci job, would not cover?

Thanks for having a look at this :)! I'll check if a test with less than 8 processes fails without and passes with this fix.

lisajulia:
jenkins build this please

PR #5511, opened 2024-08-06
Support type-tag free parameter system

akva2:
Downstream of https://github.com/OPM/opm-models/pull/916


akva2:
note: I have added the test here and not in models as models do not provide boost::test. We are merging models into simulators in a months time, this little niggle will go away then.

bska (approved):

PR #5513, opened 2024-08-07
Quell some clang warnings

akva2:
Missing overrides and static non-inline function in header.


akva2:
jenkins build this please

bska (approved):

PR #5514, opened 2024-08-07
Remove support for the print-properties parameter

akva2:
Downstream of https://github.com/OPM/opm-models/pull/918


bska (approved):
Agreed.  I'll merge into master.

PR #5515, opened 2024-08-07
Avoid exception macro in device code

multitalentloes:
OPM cuISTL does not compile at the momet because OPM_THROW allocates a type of string not allowed in device code.

This PR fixes this for now by separating device and host code where necessary and using asserts instead. 

A TODO is also added because when https://github.com/OPM/opm-common/pull/4153 is merged we can use macros from opm-common.


kjetilly:
Jenkins build this please

kjetilly (approved):

PR #5504, opened 2024-08-05
avoiding exception from std::stoi("")

GitPaean:
when not setting `OMP_NUM_THREADS` in the environment.

it causes debugging hassle for the environment that not setting `OMP_NUM_THREADS`.

PR #5516, opened 2024-08-07
avoiding unnecessary exception from std::stoi("")

GitPaean:
in FlowMain.hpp

Same with https://github.com/OPM/opm-simulators/pull/5504, while it is in a different file `FlowMain.hpp` due to code duplication. 

The exception was caught also when I do debugging, while not digging into how everything works here.

### DIFF
--- a/opm/simulators/flow/FlowMain.hpp
+++ b/opm/simulators/flow/FlowMain.hpp
@@ -38,6 +38,7 @@
 #include <dune/common/parallel/mpihelper.hh>
 #endif
 
+#include <charconv>
 #include <cstddef>
 #include <memory>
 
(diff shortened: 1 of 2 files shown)


GitPaean:
jenkins build this please

blattms (commented):
The only thing I am wondering about has been there before:

Why do environment variables overwrite command line parameters? Does not seem to be intuitive.

blattms:
According to the manpage getenv returns `NULL` if the variable is not set. We could also just check that.
I'll leave the merging to @akva2

GitPaean:
> Why do environment variables overwrite command line parameters? Does not seem to be intuitive.

Agreed. while the PR and https://github.com/OPM/opm-simulators/pull/5504 are not intended to deal with this. 

They only wants to avoid to exception which is constantly thrown in an environment that does not set `OMP_NUM_THREADS`.

GitPaean:
Maybe `                if (env_var) { ` should be outside `try {`.

bska:
We generally should not be using `stoi()` at all.  There are [much better](https://en.cppreference.com/w/cpp/utility/from_chars) ways of converting character sequences to an integer.

bska:
>> We generally should not be using `stoi()` at all. There are [much better](https://en.cppreference.com/w/cpp/utility/from_chars) ways of converting character sequences to an integer.

>
> Even then we need to cater for `NULL`, which is the problem that we solve here.

That's part of the problem, sure, but the real problem is the exception handling.  `Std::stoi()` has no choice but to report failures as exceptions whereas `std::from_chars()` has a return code.

GitPaean:
this part of code for sure can be improved. but I would argue the real problem is the code uses `""` intentionally, then makes `std::stoi()` to throw (intentionally?).  Then someone have to go through this exception every time when we debug if the debug is set to catch all the exceptions.

GitPaean:
If someone set the environment variable `OMP_NUM_THREADS` to be something alien, then the code is up to improvements to tell what parsing problem we have.  The current code just ignore it except throwing an exception.

bska:
And all of that complexity comes from using `std::stoi()` in the first place.  `Std::from_chars()` has much cleaner semantics, both for failure reporting and actual conversion from a sequence of characters.

GitPaean:
Sure. Let me update since I have begun with the code here.

GitPaean:
jenkins build this please

bska (approved):

PR #5521, opened 2024-08-12
Reservoir coupling: Send start date from slaves to master

hakonhagland:
Builds on #5453 which should be merged first.

The master process needs to know when each slave starts to determine when to activate it. I am splitting the implementation of the reservoir coupling into smaller PRs like this to simplify the review process.


opm/simulators/flow/ReservoirCouplingMaster.cpp
```diff
@@ -50,6 +50,37 @@ ReservoirCouplingMaster::ReservoirCouplingMaster(
...
+    if (this->comm_.rank() == 0) {
+        // Ensure that std::time_t is of type long since we are sending it over MPI with MPI_LONG
+        static_assert(std::is_same<std::time_t, long>::value, "std::time_t is not of type long");
+        for (unsigned int i = 0; i < this->master_slave_comm_.size(); i++) {
+            std::time_t start_date;
+            int result = MPI_Recv(
+                &start_date,
+                /*count=*/1,
+                /*datatype=*/MPI_LONG,
+                /*source_rank=*/0,
```
blattms, line 64: I have never used intercommunicators. 

Why is the source 0?
The communicator consists of the processes of the initial communicator an the slave communicator. Who exactly is 0 here? 

Note we could use MPI_ANY_SOURCE if we do not know the rank and there is only one message for the tag.

opm/simulators/flow/ReservoirCouplingMaster.cpp
```diff
@@ -101,6 +132,7 @@ void ReservoirCouplingMaster::spawnSlaveProcesses(int argc, char **argv) {
         );
         this->master_slave_comm_.push_back(std::move(master_slave_comm));
         this->slave_names_.push_back(slave_name);
+        this->num_slaves_++;
```
blattms, line 135: DO we need this member? Might be implicitly given by `this->slave_names_.size()`

opm/simulators/flow/ReservoirCouplingMaster.hpp
```diff
@@ -61,9 +54,10 @@ class ReservoirCouplingMaster {
 
     const Parallel::Communication &comm_;
     const Schedule& schedule_;
-    // MPI communicators for the slave processes
-    std::vector<MPI_Comm_Ptr> master_slave_comm_;
+    std::size_t num_slaves_ = 0;  // Initially zero, will be updated in spawnSlaveProcesses()
+    std::vector<MPI_Comm_Ptr> master_slave_comm_; // MPI communicators for the slave processes
```
blattms, line 58: Is the documentation string correct? I thought this would be an intercommunicator and would inlude all ranks from the original one and the ones from the newly created one ("Intercommunicator between original group and the newly spawned group (handle)." says the documentation)

opm/simulators/flow/ReservoirCouplingMaster.cpp
hakonhagland: @blattms Thanks for reviewing this. (I plan to rebase and update this PR soon.) The source_rank of 0 means that we will be receiving messages from rank 0 in the remote group of the inter communicator. This corresponds to rank 0 in the slave process sending messages to the parent communicator. Yes, we could use message tag MPI_ANY_SOURCE but it might be clearer to use specific tag and document the type of message we will be receiving here.

opm/simulators/flow/ReservoirCouplingMaster.hpp
hakonhagland: We currently use these handles to communicate (MPI_Send(), MPI_Recv()) between a specific rank in the master and a specific rank in the slave process. So the comment should maybe use "to" instead of "for" ?  Like this: "MPI communicators to the slave processes" ?

hakonhagland:
Rebased

blattms:
This can be closed now.

hakonhagland:
> This can be closed now as #5620 was merged, right?

PR #5527, opened 2024-08-13
Support deactivation of individual DRSDT/DRVDT/DRSDTCON regions

totto82:
Depends on https://github.com/OPM/opm-common/pull/4161

### DIFF
--- a/opm/simulators/flow/FlowGenericProblem_impl.hpp
+++ b/opm/simulators/flow/FlowGenericProblem_impl.hpp
@@ -418,8 +418,7 @@ vapparsActive(int episodeIdx) const
 
 template<class GridView, class FluidSystem>
 bool FlowGenericProblem<GridView,FluidSystem>::
-beginEpisode_(std::size_t numDof,
-              bool enableExperiments,
+beginEpisode_(bool enableExperiments,
               int episodeIdx)
 {
     if (enableExperiments && gridView_.comm().rank() == 0 && episodeIdx >= 0) {
@@ -438,10 +437,6 @@ beginEpisode_(std::size_t numDof,
                   << "\n ";
         OpmLog::info(ss.str());
     }
-
-    this->mixControls_.init(numDof,
-                        episodeIdx,
-                        eclState_.runspec().tabdims().getNumPVTTables());
 
     const auto& events = schedule_[episodeIdx].events();
 
--- a/opm/simulators/flow/FlowGenericProblem_impl.hpp
+++ b/opm/simulators/flow/FlowGenericProblem_impl.hpp
@@ -418,7 +418,8 @@ vapparsActive(int episodeIdx) const
 
 template<class GridView, class FluidSystem>
 bool FlowGenericProblem<GridView,FluidSystem>::
-beginEpisode_(bool enableExperiments,
+beginEpisode_(std::size_t numDof,
+              bool enableExperiments,
               int episodeIdx)
 {
     if (enableExperiments && gridView_.comm().rank() == 0 && episodeIdx >= 0) {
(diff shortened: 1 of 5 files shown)


opm/simulators/flow/MixingRateControls.cpp
```diff
@@ -137,22 +137,49 @@ drsdtConvective(int episodeIdx) const
...
 template<class FluidSystem>
 void MixingRateControls<FluidSystem>::
 updateExplicitQuantities(const int episodeIdx,
                          const Scalar timeStepSize)
 {
     const auto& oilVaporizationControl = schedule_[episodeIdx].oilvap();
-    if (this->drsdtActive(episodeIdx)) {
-        // DRSDT is enabled
-        for (std::size_t pvtRegionIdx = 0; pvtRegionIdx < maxDRs_.size(); ++pvtRegionIdx)
+    // DRSDT is enabled
```
GitPaean, line 171: Here, we already have access to `oilVaporizationControl`, then `oilVaporizationControl.drsdtActive(pvtRegionIdx)` will be able to determine whether it is active. In my opinion, we do not need to go through the function `drvdtActive(episodeIdx, pvtRegionIdx)` .  Basically, it will create another reference to `oilVaporizationControl`.

GitPaean:
The PR mostly extends the existing implementation of DRSDTCON to multiple PVT regions. 

To protect this development, it can be a good idea to extend the regression test 'CO2STORE_DRSDTCON.DATA'  to be a multiple PVT region one.

opm/simulators/flow/MixingRateControls.cpp
```diff
@@ -218,12 +249,12 @@ maxGasDissolutionFactor(const unsigned timeIdx,
                         const int episodeIdx,
                         const int pvtRegionIdx) const
 {
-    if (!this->drsdtActive(episodeIdx) || maxDRs_[pvtRegionIdx] < 0.0) {
+    if (!this->drsdtActive(episodeIdx, pvtRegionIdx) || maxDRs_[pvtRegionIdx] < 0.0) {
```
GitPaean, line 252: from the code above, 
```c++
    for (std::size_t pvtRegionIdx = 0; pvtRegionIdx < maxDRs_.size(); ++pvtRegionIdx) {
        if (this->drsdtActive(episodeIdx, pvtRegionIdx)) {
            maxDRs_[pvtRegionIdx] = oilVaporizationControl.getMaxDRSDT(pvtRegionIdx) * timeStepSize;
        }
    }
```

Will it be possible to have `this->drsdtActive(episodeIdx, pvtRegionIdx)` is true while `maxDRs_[pvtRegionIdx] < 0.0`?

totto82:
> To protect this development, it can be a good idea to extend the regression test 'CO2STORE_DRSDTCON.DATA' to be a multiple PVT region one. (modifying the existing one with new features, while not adding the number of regression tests. )

Good point. I will do that.

opm/simulators/flow/MixingRateControls.cpp
totto82: No, it shouldn't. I will double check and remove it if possible.

opm/simulators/flow/MixingRateControls.cpp
totto82: Good point. I will update.

GitPaean (approved):
Thanks for the update.  I am happy with the PR now. I am not specifically familiar with this part, but I think we can keep improving it along the way.

opm/simulators/flow/FlowGenericProblem_impl.hpp
```diff
@@ -418,7 +418,8 @@ vapparsActive(int episodeIdx) const
 
 template<class GridView, class FluidSystem>
 bool FlowGenericProblem<GridView,FluidSystem>::
-beginEpisode_(bool enableExperiments,
+beginEpisode_(std::size_t numDof,
```
GitPaean, line 421: As notified by @akva2 , `numDof` introduces a warning here.

PR #5528, opened 2024-08-14
Pass gas mass rate to summary

totto82:
Depends on https://github.com/OPM/opm-common/pull/4162


bska (approved):
I might personally have chosen to pass the `Scalar` argument to `assignMassGasRate()` as a value instead of a reference, but that's a mostly a matter of taste I guess.  I'll merge into master.

PR #5533, opened 2024-08-14
Move parameters to typetag free parameter system: Last batch

akva2:
This is the final batch completing moving parameters to the typetag free system.

I'm marking this draft as it will create havoc for #5157, I'll handle the rebasing after that one has been merged.


akva2:
jenkins build this please

bska:
> This is the final batch completing moving parameters to the typetag free system.

Is there no companion PR in any upstream repositories?

bska:
> indeed, no companion.

Okay understood.

> there will be a cleanup PR afterwards removing the old system as well as one moving some more code to cpp files.

Very good.  I look forward to seeing this work becoming fully finalised.

akva2:
jenkins build this please

bska (approved):
I didn't verify that every single parameter default value was fully transferred to the new system, but the successful regression test suggests that it was.  This looks good to me and I'll merge into master.

PR #5534, opened 2024-08-15
Ensure default ALQ is set for all production wells on all processes

vkip:
Fixes an MPI bug for cases where a well switches from producer to injector at the restart step.

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -424,7 +424,9 @@ void WellState<Scalar>::init(const std::vector<Scalar>& cellPressures,
         }
     }
 
-    updateWellsDefaultALQ(wells_ecl, summary_state);
+    // ALQ state needs to be defined for all wells, otherwise alqstate.packSize() will be different
+    // on each process, and the comm.sum call in communicateGroupRates will fail.
+    updateWellsDefaultALQ(schedule.getWells(report_step), summary_state);
 }
 
 template<class Scalar>
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -424,9 +424,7 @@ void WellState<Scalar>::init(const std::vector<Scalar>& cellPressures,
         }
     }
 
-    // ALQ state needs to be defined for all wells, otherwise alqstate.packSize() will be different
-    // on each process, and the comm.sum call in communicateGroupRates will fail.
-    updateWellsDefaultALQ(schedule.getWells(report_step), summary_state);
+    updateWellsDefaultALQ(schedule, report_step, summary_state);
 }
 
 template<class Scalar>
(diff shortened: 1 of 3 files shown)


vkip:
jenkins build this please

opm/simulators/wells/WellState.cpp
```diff
@@ -424,7 +424,9 @@ void WellState<Scalar>::init(const std::vector<Scalar>& cellPressures,
         }
     }
 
-    updateWellsDefaultALQ(wells_ecl, summary_state);
+    // ALQ state needs to be defined for all wells, otherwise alqstate.packSize() will be different
+    // on each process, and the comm.sum call in communicateGroupRates will fail.
+    updateWellsDefaultALQ(schedule.getWells(report_step), summary_state);
```
bska, line 429: I appreciate the sentiment and we clearly need to fix this.  However [`Schedule::getWells()`](https://github.com/OPM/opm-common/blob/e85c8b8c8abcae34279de425093754a497b6a24f/opm/input/eclipse/Schedule/Schedule.cpp#L1047-L1057) is an expensive function, so if we could make this less expensive then that would be preferable.  Perhaps a new (private) overload of `WellState::updateWellsDefaultALQ()` of the form
```C++
template <typename Scalar>
void WellState<Scalar>::updateWellsDefaultALQ(const Schedule& schedule,
                                              const int report_step,
                                              const SummaryState& summary_state)
{
    const auto wells = schedule.wellNames(report_step);
    for (const auto& wname : wells) {
        const auto& well = schedule.getWell(wname, report_step);
        if (! well.isProducer()) {
            continue;
        }
        const auto alq = well.alq_value(summary_state);
        this->alq_state.update_default(wname, alq);
    }
}
```
since [`Schedule::getWell()`](https://github.com/OPM/opm-common/blob/e85c8b8c8abcae34279de425093754a497b6a24f/opm/input/eclipse/Schedule/Schedule.cpp#L1082-L1084) does not create copies of the `Well` objects?

bska (commented):
Thanks.  With the possible caveat about the cost of `Schedule::getWells()` this does indeed appear to solve a real problem.  That said, I think the `ALQState` object could use some revision too.  It's not clear to me that the current code&ndash;irrespective of the fix in the current PR&ndash;is able to handle the case of a well switching from an injector to a producer in an `ACTIONX` block.

Tangentially, I'm also not a huge fan of the way the `ALQState` uses [four separate](https://github.com/OPM/opm-simulators/blob/e548bbfd9d6b7d81a66503edf24b85b7084721dd/opm/simulators/wells/ALQState.hpp#L63-L66) `std::map<>` objects keyed on the well name to track the object data.  Addressing that would be the subject of separate work, though.

opm/simulators/wells/WellState.cpp
vkip: Good point, I didn't think of the cost. Maybe just change the signature to what you propose above and update the single other call to the method (in BlackoilWellModel_impl.hpp)? Might bring us one step closer to handling switching in ACTIONX also (maybe - didn't dig into it..)

vkip:
jenkins build this please

bska (approved):
Thanks a lot for the update.  We **may** want to reintroduce the `vector<Well>` overload at some point, and pass `wells_ecl_` instead of `schedule()` from `beginTimeStep()`, since `wells_ecl_` contains only those wells that are connected on the local rank while `schedule()` contains all wells.  That's an optimisation though, and only profiling will reveal if it makes any difference.

I'll merge the current version into the master branch.

PR #5537, opened 2024-08-16
Avoid updating max Rs first step after restart.

vkip:
Needed for https://github.com/OPM/opm-common/pull/4166


vkip:
jenkins build this opm-common=4166 please

bska (approved):
This looks good to me and I'll merge into master.  Thanks a lot for fixing this.  The one aspect that one might discuss is passing `first_step_ && (episodeIdx > 0)` as an argument to `updateExplicitQuantities_()`.  Since `first_step_` is a data member of the same class. member function `updateExplicitQuantities_()` could arguably load the same value directly from the data member.  That said, if we ever get more implementations of `updateExplicitQuantities_()` then it probably makes sense to be explicit about the meaning of the new parameter.

PR #5538, opened 2024-08-16
Move parameter registration to translation units

akva2:
To avoid compiling the code in every simulator object.

### DIFF
--- a/opm/simulators/flow/BlackoilModelParameters.cpp
+++ b/opm/simulators/flow/BlackoilModelParameters.cpp
@@ -1,0 +1,227 @@
+/*
+  Copyright 2015 SINTEF ICT, Applied Mathematics.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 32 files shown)


akva2:
jenkins build this please

opm/simulators/flow/BlackoilModelParameters.cpp
```diff
@@ -0,0 +1,226 @@
...
+    local_well_solver_control_switching_ = Parameters::Get<Parameters::LocalWellSolveControlSwitching>();
+    use_implicit_ipr_ = Parameters::Get<Parameters::UseImplicitIpr>();
+    nonlinear_solver_ = Parameters::Get<Parameters::NonlinearSolver>();
+    const auto approach = Parameters::Get<Parameters::LocalSolveApproach>();
+    if (approach == "jacobi") {
+        local_solve_approach_ = DomainSolveApproach::Jacobi;
+    } else if (approach == "gauss-seidel") {
+        local_solve_approach_ = DomainSolveApproach::GaussSeidel;
+    } else {
+        throw std::runtime_error("Invalid domain solver approach '" + approach + "' specified.");
```
bska, line 77: I'm sure we get it through some transitive include statement, but we should explicitly
```C++
#include <stdexcept>
```
to bring a declaration of the `std::runtime_error` type into scope.

bska (commented):
Save for one potential missing `#include` statement this looks good to me.

opm/simulators/flow/BlackoilModelParameters.cpp
akva2: sure, just an oversight..

akva2:
jenkins build this please

PR #5539, opened 2024-08-19
fixed: do not try to hide parameters that are not registered without MPI

akva2:
Whoopsie..

### DIFF
--- a/opm/simulators/flow/FlowGenericVanguard.cpp
+++ b/opm/simulators/flow/FlowGenericVanguard.cpp
@@ -447,15 +447,15 @@ void FlowGenericVanguard::registerParameters_()
          "distribution purposes. If empty, the built-in partitioning "
          "method will be employed.");
     Parameters::Hide<Parameters::ExternalPartition>();
+
+    Parameters::Hide<Parameters::ZoltanImbalanceTol<Scalar>>();
+    Parameters::Hide<Parameters::ZoltanParams>();
 #endif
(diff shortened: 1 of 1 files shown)


akva2:

bska (approved):
Very good&ndash;I was about to submit something similar.  Please merge when green.

bska:
> the test failure is caused by the new defaults PR, not the parameter stuff

Agreed.  We noticed this in [post-commit build](https://ci.opm-project.org/job/opm-simulators/2305/testReport/) of #5157.  For whatever reason, the [ALUGrid regression test](https://github.com/OPM/opm-simulators/blob/bee91d2fdd5586505eceb54f525ec8f905b1c483/regressionTests.cmake#L1493-L1498) appears to give different results when configured as a sequential build compared to a parallel build.

bska:
> For whatever reason, the [ALUGrid regression test](https://github.com/OPM/opm-simulators/blob/bee91d2fdd5586505eceb54f525ec8f905b1c483/regressionTests.cmake#L1493-L1498) appears to give different results when configured as a sequential build compared to a parallel build.

One potential fix might be to go back to using `--linear-solver=ilu0` for this particular test.

akva2:
have you confirmed that fixes it? if so that seems like a acceptable solution.

bska:
> have you confirmed that fixes it?

I have not because I don't currently have a build which includes ALUGrid...

akva2:
there is an approval so i'll merge this to simplify further investigation.

PR #5540, opened 2024-08-19
changed: use ilu0 for alugrid test

akva2:
Trying to see if this resolves the serial-differ-from-parallel fallout from the new defaults.


akva2:

akva2:
jenkins build this.opm-tests=1210 serial please

akva2:
so no, this did not help.

bska:
> so no, this did not help.

Yeah, that's a pity.  One other potential test would be to restore **all** of the default parameters from #5157 to their original states, i.e.,
```sh
--linear-solver=ilu0 --tolerance-mb=1.0e-6 --min-strict-cnv-iter=0 \
--local-well-solve-control-switching=false --use-implicit-ipr=false \
--enable-drift-compensation=true
```
If that does not help, then we may have to disable the ALUGrid test.

PR #5542, opened 2024-08-19
Add float support to simulators: Batch 1

akva2:
#5286  has bit-rotted a little, and it's time to get back on this horse.

These are parts that easily was cherry-picked and which do not rely on any upstreams. I'd appreciate if we can flush this to keep the amount to reduce the amount of code I have to juggle while rebasing.


akva2:

akva2:
jenkins build this please

bska (approved):
I see no obvious problem here&ndash;other than the fact that we're gradually getting more instantiation macros&ndash;and the build check is green.  I'll merge into master.

akva2:
yeah the macros are not pretty but there is also no better solution to not repeat oneself too much..

PR #5545, opened 2024-08-20
restore all flags for alugrid test

akva2:


akva2:

bska:

bska:
The build check is green without updating the reference solutions.  I'm very tempted to merge this as is and not worry too much about it right now, but this does suggest that there's **something** about the new defaults that makes the tests more sensitive.

akva2:
agreed, passing test beats not passing.

bska:
>> I'm very tempted to merge this as is and not worry too much about it right now,
>
> agreed, passing test beats not passing.

Very good then.  I'll merge this into the master branch as is.

PR #5546, opened 2024-08-20
BUGFIX Fix no wells case

totto82:
Cases with CO2STORE and was broken by https://github.com/OPM/opm-simulators/pull/5528. This is a fix.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1690,7 +1690,7 @@ assignMassGasRate(data::Wells& wsrpt,
     using rt = data::Rates::opt;
     for (auto& wrpt : wsrpt) {
         auto& well_rates = wrpt.second.rates;
-        const auto w_mass_rate = well_rates.get(rt::gas) * gasDensity;
+        const auto w_mass_rate = well_rates.get(rt::gas, 0.0) * gasDensity;
         well_rates.set(rt::mass_gas, w_mass_rate);
     }
 }


totto82:
jenkins build this please

blattms:
Mmh. This does not fix the deadlock on my system.

blattms:
the actual problem is that the gas rate seems to be uninitialized for some wells and we throw an exception on one process (but not on the other):
```
Thread 1 "flow" hit Catchpoint 4 (exception thrown), 0x00007ffff24a90a1 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) up
#1  0x00007ffff7138aba in Opm::data::Rates::get (
    this=this@entry=0x55555789f8b0, m=m@entry=Opm::data::Rates::opt::gas)
    at /home/mblatt/src/dune/opm-smeaheia-fail/opm-common/opm/output/data/Wells.hpp:1049
1049                throw std::invalid_argument( "Uninitialized value." );
(gdb) l
1044            return static_cast< opt >( mand ) == m;
1045        }
1046
1047        inline double Rates::get( opt m ) const {
1048            if( !this->has( m ) )
1049                throw std::invalid_argument( "Uninitialized value." );
1050
1051            return this->get_ref( m );
1052        }
1053
(gdb) 
```

blattms:
Not even a parallel problem. The exception is thrown in serial, too and the simulation is aborted, immediately with:
```
Simulation aborted as program threw an unexpected exception: Uninitialized value.
```

totto82:
Thanks for the update. Seems this fixes "a" problem but not "the" problem. I will dig a bit more.

totto82:
jenkins build this please

totto82:
This should work better. The first fix only fixed the case where all wells where shut. This should fix the case where only one well is shut.

blattms (approved):
That is the one that fixes stuff. Thanks

blattms:
jenkins build this please

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1690,6 +1690,8 @@ assignMassGasRate(data::Wells& wsrpt,
     using rt = data::Rates::opt;
     for (auto& wrpt : wsrpt) {
         auto& well_rates = wrpt.second.rates;
+        if (wrpt.second.dynamicStatus == Opm::Well::Status::SHUT)
+            continue;
```
bska, line 1694: That's fine I guess although it might be better to check `dynamicStatus != OPEN` instead.  Moreover, if we're really going to take action based on the status, then there's no need to intialise the `well_rates` reference before we do the checking.  Another option would be to use [the default value overload](https://github.com/OPM/opm-common/blob/81076046b018d278a50919f6efb3453a6d756a14/opm/output/data/Wells.hpp#L1054-L1058) instead and to write
```C++
const auto w_mass_rate = well_rates.get(rt::gas, 0.0) * gasDensity;
```
In that case we do not check the well status.  I don't immediately know which alternative makes the most sense here.

bska (commented):
I think there are a couple of alternative implementations here and I'd like to know what you think about those.

blattms:
I will assign the merging to @bska, then. Just be aware that this fixes an issue that breaks the daily benchmark builds. (since Thursday) Hence it is a bit urgent.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
totto82: > const auto w_mass_rate = well_rates.get(rt::gas, 0.0) * gasDensity;

I didnt know about this option. Seems like a more robust way of handling this.

totto82:
jenkins build this please

opm/simulators/wells/BlackoilWellModelGeneric.cpp
totto82: I have updated the PR accordingly and all my small tests pass. Lets get this in when Jenkins is happy.

bska (approved):

PR #5548, opened 2024-08-20
Support float simulators - batch 2

akva2:
Small batch to separate out the upstream dependency.

Downstream of https://github.com/OPM/opm-common/pull/4168


bska (approved):

PR #5549, opened 2024-08-21
ActionHandler: instance for float

akva2:
Another small one to isolate the upstream.

Downstream of https://github.com/OPM/opm-common/pull/4169


bska (approved):

PR #5550, opened 2024-08-21
Float support in simulators: Batch 3

akva2:
Another batch. No upstream for this.


bska (approved):
This is fine and I'll merge into master.  The one thing that was not fully clear to me was the change to `opencl-source-provider.cmake`, but there's nothing that jumps out to me about it either.

akva2:
basically it is replacing 'double' with 'float' in the opencl kernels (which are written using double).

PR #5551, opened 2024-08-21
Float support in simulators: Batch 4

akva2:
Downstream of https://github.com/OPM/opm-common/pull/4171


bska (approved):

PR #5559, opened 2024-08-26
do not capture compile time data in lambda

akva2:
quells clang warnings


akva2:
jenkins build this please

blattms:
Is that compile time because of boost (only marked const, it seems)?

akva2:
for PODs const is 'the same as' constexpr if they are known at compile time.

akva2:
jenkins build this please

bska (approved):

PR #5561, opened 2024-08-27
amgcpr: use override

akva2:
Since ```hasPerfectUpdate()``` was marked override, the rest of the methods needs to use modern style as well or compilers warn.


akva2:
jenkins build this please

bska (approved):

PR #5564, opened 2024-08-28
adding --tolerance-cnv-relaxed for H2STORE regression

GitPaean:
so that the solution is converged and results in more stable regression checking.

### DIFF
--- a/regressionTests.cmake
+++ b/regressionTests.cmake
@@ -689,35 +689,40 @@ add_test_compareECLFiles(CASENAME h2store
                          SIMULATOR flow
                          ABS_TOL ${abs_tol}
                          REL_TOL ${rel_tol}
-                         DIR h2store)
+                         DIR h2store
+                         TEST_ARGS --tolerance-cnv-relaxed=0.01)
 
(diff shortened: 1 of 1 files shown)


GitPaean:
When solution is not converged, we see the unnecessary zig-zag in one case, H2STORE_DISGAS

image

With this PR, 
image

For other cases, it improves the matching for the BGSAT curves, 
image


image



But the changes in https://github.com/OPM/opm-tests/pull/1218/commits/315f6efed92d29c471211b40de2f51516903002d is more significant in stabilizing the regression checking since it stablizes the initial time stepping. 

When initial time stepping is not stable, then any regression checking afterwards are not easy.

GitPaean:
Just noticed that H2STORE_DISGAS is not in the regression. But it should be able to join now.

GitPaean:
jenkins build this opm-tests=1218 please

GitPaean:
jenkins build this opm-tests=1218 please

GitPaean:
The reference has been updated. I am merging this PR now.

PR #5567, opened 2024-08-30
add python packaging fluff

akva2:
This is the fluff required to deploy the opm-simulators module to pypi.

It also builds the opm-common package to ensure that there are no version differences between the bindings.
Strictly speaking we could thus remove the packaging Docker from opm-common as it will not be used, but I keep it intact for now in case somebody wants to just build those.

In the process I was forced to abandon manylinux2014 and bump to manylinux_2_28. Basically this means dropping RH7 support. manylinux2014 is based on centos7 and all the repos are gone so it's .. very tedious to build for and frankly not worth it any longer. Those on those platforms can stick to the 2024.04 stable release.

### DIFF
--- a/python/build-deps.sh
+++ b/python/build-deps.sh
@@ -1,0 +1,83 @@
+#!/bin/bash
+
+set -e
+
+BUILD_JOBS=$1
+
+export CMAKE_GENERATOR=Ninja
+
+pushd /tmp/opm
+
+# Build boost
+git clone --depth 1 --branch boost-1.84.0 https://github.com/boostorg/boost
+pushd boost
+git submodule init
+git submodule update
+mkdir build
+cd build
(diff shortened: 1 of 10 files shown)


python/build-deps.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+
+set -e
+
+BUILD_JOBS=$1
+
+export CMAKE_GENERATOR=Ninja
```
atgeirr, line 7: Is this crucial, or just convenient?

atgeirr (approved):
I see no problems here, but I am not a Python or deployment expert.

python/build-deps.sh
akva2: convenience. it simply means i don't have to pass it to every cmake invocation. using ninja instead of make is a significant speedup.

akva2:
jenkins build this please

PR #5569, opened 2024-09-02
fixed: make python simulator object depend on copy_python

akva2:
if not, building just the python binding won't end up in a complete build


akva2:
jenkins build this please

bska:
> if not building just the python binding won't end up in a complete build

Sorry, I'm having a little bit of difficulty parsing that sentence.  Would you mind expanding a little on what situation you're addressing.

akva2:
while packaging the python bindings i only want to build the python objects. so i do ```ninja simulators```. problem is this does not copy the necessary python files into the build directory. now it does.

bska (commented):
> when packaging the python bindings i only want to build the python object. so i do `ninja simulators`. problem is this does not copy the necessary python files (in particular the `__init__.py` file) into the build directory. now it does.

Ah, okay&ndash;thanks.  Then I understand what you're trying to do.  Very good.  I'll merge into master.

PR #5573, opened 2024-09-03
Add Saturation Function Consistency Checks for SCALECRS

bska:
This commit introduces consistency checks for the scaled displacing saturation in the three point horizontal scaling method (`SCALECRS=YES`).  These plug into the framework introduced in commit c3939c54 (PR #5438).  We implement the following two checks

  - $S_{\mathrm{G},\mathrm{CR}} < 1-S_{\mathrm{OG},\mathrm{CR}}-S_{\mathrm{WL}} < S_{\mathrm{GU}}$
  - $S_{\mathrm{W},\mathrm{CR}} < 1-S_{\mathrm{OW},\mathrm{CR}}-S_{\mathrm{GL}} < S_{\mathrm{WU}}$

which collectively guarantee a mobile displacing oil saturation in the two phase gas/oil and oil/water systems.


bska:
jenkins build this please

akva2 (approved):

bska:

PR #5574, opened 2024-09-04
BlackoilXXParams: introduce translation units

akva2:
A small chip at compile times, allowing this code to only be compiled once instead of once per simulator object.


akva2:
jenkins build this please

opm/models/blackoil/blackoilpolymerparams.cpp
```diff
@@ -0,0 +1,398 @@
...
+#include <opm/input/eclipse/EclipseState/Tables/PlyviscTable.hpp>
+#endif
+
+#include <cassert>
+#include <cstddef>
+#include <stdexcept>
+#include <type_traits>
+
+namespace {
+std::vector<std::vector<float>> doubleVecsToFloat(const std::vector<std::vector<double>>& input)
```
bska, line 42: If I'm reading the code correctly, this helper function is unused unless we're building the `float` variants or if we don't `HAVE_ECL_INPUT` (unlikely).  Unused functions in anonymous namespaces may (or may not) generate a warning in some compilers so maybe we could guard the definition by `#if FLOW_INSTANTIATE_FLOAT && HAVE_ECL_INPUT`?

bska (commented):
Looks good except for a tiny nit concerning a potentially unused helper function.  I notice a couple of instances of vectors being constructed with a size and then filled in by copying values that could probably use range constructors instead, but we can address those in follow-up work.

opm/models/blackoil/blackoilpolymerparams.cpp
akva2: indeed. remedied. i also rewrote to use emplace_back.

akva2:
jenkins build this please

opm/models/blackoil/blackoilpolymerparams.cpp
bska: > indeed. remedied. i also rewrote to use emplace_back.

Cool&ndash;thanks. Another option might of course be to make it a template since unused templates are just whitespace:
```C++
template <typename To, typename From>
std::vector<std::vector<To>> convertVecVecTo(const std::vector<std::vector<From>>& src)
{
    std::vector<std::vector<To>> dst{};
    // ...
    return dst;
}
```
and call it as `convertVecVecTo<float>(src)`, but that may just be too much effort for too little gain.

akva2: if you prefer i can do that. just say which.

bska: > if you prefer i can [make it a template]. just say which.

Let's keep it as is.  The template option might become useful if `Scalar` can ever be anything other than `float` or `double`.

bska (approved):

opm/models/blackoil/blackoilpolymerparams.cpp
```diff
@@ -0,0 +1,403 @@
...
+#endif
+
+#include <cassert>
+#include <cstddef>
+#include <stdexcept>
+#include <type_traits>
+
+namespace {
+
+#if FLOW_INSTANTIATE_FLOAT && HAVE_ECL_INPUT
```
GitPaean, line 43: question here, if FLOW_INSTANTIATE_FLOAT is false, then the code will not compile, right?

akva2: yes, that's the point of the #if's.

GitPaean: Somehow locally, I got a config.h file without defining FLOW_INSTANTIATE_FLOAT through IDE cmake.  Let me try again.

akva2: that's likely because you have not turned on the cmake option to build the float simulators, ie, -DBUILD_FLOW_FLOAT_VARIANTS=1 ?

GitPaean: Yes. All the flow variants do not compile without using the cmake flag.

akva2: ah yes. I see the problem. That was very stupid of me. fix coming up.

PR #5576, opened 2024-09-04
Fix build without float simulators

akva2:
And without MPI.

I look silly in my double pointy hat. At least I'm one of the cool guys moving fast and breaking stuff...


akva2:

bska (approved):
Thanks.  Sorry for leading you astray on the preprocessor-based approach.  Please merge when green.

PR #5577, opened 2024-09-04
Correctly treat transmissibilities for PINCH option(4) ALL.

blattms:
In that case processEclipseFormat will create NNCs over the pinched out cells with correct transmissibilities. Those NNCs are marked as being created in PINCH and used to overwrite calculated transmissibilties for the horizontal intersections (transmissibilities of normal NNCs are usually added).

Note that the transmissibilities (without the multipliers) do not change later. Only the multipliers can change them in the schedule and that will still happen to some extend. Canges of MULTZ for pinched out cells will currently not be honored.

This is tested to work for all provided test cases.

Needs OPM/opm-common#4181 and OPM/opm-grid#756


atgeirr (approved):
Looks good to me.

PR #5579, opened 2024-09-05
Don't Count Cells Unless Needed

bska:
If the user chooses a specific number of NLDD domains, there is no need to compute the default number of domains based on a specific target number of cells per domain.  Defer the, potentially costly, calculation to when we know we need it.  I didn't pick up on this detail when reviewing PR #5021.

While here, also split/rejoin a couple of long lines.


bska:
jenkins build this please

akva2 (approved):

bska:

PR #5581, opened 2024-09-05
Fix Build on Older Compilers

bska:
The `strsignal()` function is declared in `<string.h>` on those.

### DIFF
--- a/opm/models/utils/terminal.cpp
+++ b/opm/models/utils/terminal.cpp
@@ -30,6 +30,7 @@
 
 #include <csignal>
 #include <iostream>
+#include <string.h>             // strsignal()
 #include <sys/ioctl.h>
 #include <unistd.h>
 


bska:
jenkins build this please

akva2:
hmm, okay, i see it errors on jenkins without mpi as well so i guess some transitive include hid the issue.

akva2 (approved):

GitPaean:
I also ran into this after rebasing. Any difference between string.h or cstring ? both will fix it.

akva2:
```<cstring>``` allows accessing symbols as std::foo instead of ::foo but no real difference as far as fixing this problem goes.

**Comment by bska** [comment]
> Any difference between string.h or cstring ?

Yes, there actually is.  For **standard** C library functions (e.g., `strlen()`), the primary difference is that the `<cX>` version of the `<X.h>` header will, or is at least intended to, export the symbols in the `std::` namespace&ndash;e.g., `std::strlen()`.  We should generally not expect the symbols to be visible outside the `std::`  namespace, i.e., in the global namespace, when including `<cX>`.  If they are, then that's actually namespace pollution.

The C++ standard expects you to do something like
```C++
#include <cX>
using namespace std;
```
if you wish to access standard C library functions without the `std::` namespace.  That's of course from the ideal/ivory tower point of view in which everyone creates an implementation from scratch and do not "just" build on top of an existing C implementation.

> both will fix it.

In general, no, not in this case.  `Strsignal()` is not a standard C library function, it's a Posix function, and for those we need the `<X.h>` headers.

PR #5582, opened 2024-09-05
Move more opm-models code to translation units

akva2:
Granted these are very minor optimizations, but small chips...


akva2:
jenkins build this please

bska (approved):

PR #5584, opened 2024-09-06
added: scripts for generating failure reports on jenkins

akva2:
This creates a pdf for each failed case with plots of summary curves.

I am in no way claiming this is a marvel of engineering but it's a start.

Requires https://github.com/OPM/opm-common/pull/4184

### DIFF
--- a/tests/make_failure_report.sh
+++ b/tests/make_failure_report.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# Generates a summary plot comparison for each failed test case
+
+OPM_TESTS_ROOT=$1
+BUILD_DIR=$2
+RESULT_DIR=$3
+SOURCE_DIR=`dirname "$0"`
+
+FAILED_TESTS=`cat $BUILD_DIR/Testing/Temporary/LastTestsFailed*.log`
+
+mkdir -p $BUILD_DIR/failure_report
+cd $BUILD_DIR/failure_report
+rm -f *
+
+for failed_test in $FAILED_TESTS
+do
+  grep -q -E "compareECLFiles" <<< $failed_test
+  test $? -ne 0 && continue
+
+  failed_test=`echo $failed_test | sed -e 's/.*://g' -e 's/\+/./g'`
+  # Extract test properties
+  binary=$(awk -v search="set_tests_properties\\\($failed_test\$" -v prop="SIMULATOR" -f ${SOURCE_DIR}/getprop.awk $RESULT_DIR/CTestTestfile.cmake)
+  dir_name=$(awk -v search="set_tests_properties\\\($failed_test\$" -v prop="DIRNAME" -f ${SOURCE_DIR}/getprop.awk $RESULT_DIR/CTestTestfile.cmake)
+  file_name=$(awk -v search="set_tests_properties\\\($failed_test\$" -v prop="FILENAME" -f ${SOURCE_DIR}/getprop.awk $RESULT_DIR/CTestTestfile.cmake)
+  test_name=$(awk -v search="set_tests_properties\\\($failed_test\$" -v prop="TESTNAME" -f ${SOURCE_DIR}/getprop.awk $RESULT_DIR/CTestTestfile.cmake)
+  echo "Processing ${test_name}"
+  $SOURCE_DIR/plot_well_comparison.py -r $OPM_TESTS_ROOT/$dir_name/opm-simulation-reference/$binary/$file_name -s $RESULT_DIR/tests/results/$binary+$test_name/$file_name -c $test_name -o plot
+done
+
+$SOURCE_DIR/plot_well_comparison.py  -o rename
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -435,6 +435,19 @@ include (${CMAKE_CURRENT_SOURCE_DIR}/modelTests.cmake)
 
 if (HAVE_OPM_TESTS)
   include (${CMAKE_CURRENT_SOURCE_DIR}/compareECLFiles.cmake)
+
+  if(OPM_ENABLE_PYTHON)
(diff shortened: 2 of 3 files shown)


GitPaean:
Maybe some others can comment whether they want pdf or just images on the webpage to avoid downloading it? @steink @atgeirr @hnil @totto82

akva2:
inlining 100's of images in the pull request would not be very .. nice .. and it would mean converting to bitmaps which is really suboptimal for plots.

akva2:
jenkins build this opm-common=4184 failure_report please

akva2:
jenkins build this opm-common=4184 failure_report please

akva2:
woop woop. only two attempts? maybe i'm getting better at stuff haha. generated document: https://ci.opm-project.org/job/opm-simulators-PR-builder/6607/artifact/mpi/build-opm-simulators/failure_report/0a3_grpctl_msw_model2.pdf

GitPaean:
It looks cool. Yeah, the makers can be bigger (:P). 

At the same time, maybe someone still want the text or summary output, for example for some file format or the restart file regression. Maybe @bska can comment?

akva2:
problem with increasing the mark size is that for cases with lots of steps you will basically only see the markers.

you can execute this yourself locally if you want to play around with cosmetics.
invocation is basically

```bash
$OPM_SIMULATORS_SOURCE_DIR/tests/make_failure_report.sh $OPM_TESTS_ROOT $OPM_SIMULATORS_BUILD_DIR
```
if you use the super-build you need to make a link manually
```
ln -sf $BUILD_DIR/Testing/Temporary/LastTestsFailed.og  $BUILD_DIR/opm-simulators/Testing/Temporary/LastTestsFailed.log
 ```

**Comment by akva2** [comment]
jenkins build this opm-common=4184 failure_report please

**Comment by steink** [comment]
This is very good and useful! With time it would be good if we also could sort the plots according to _badness_ (I know this is not necessarily straight-forward) and, say, only output the 10-20 worst. Think that would be sufficient for determining whether a failure is kind of OK or whether it needs further investigation.

**Comment by GitPaean** [comment]
> problem with increasing the mark size is that for cases with lots of steps you will basically only see the markers.

yes. I are aware of this and skeptical whether we can do it in a uniform manner.

**Comment by akva2** [comment]
okay have added a rather non-rigorous attempt at sorting by eyeball norm order. Specifically linearly resample the new simulation on reference grid and then look at the relative l2-norm deviation. It kinda works, there is no guarantee it will find *the* worst case and put it first but it will at least put things that obviously deviate in eyeball norm before those that do not.

**Comment by akva2** [comment]
jenkins build this opm-common=4184 failure_report please

**Comment by bska** [comment]
> okay have added a rather non-rigorous attempt at sorting by eyeball norm order. Specifically linearly resample the new simulation on reference grid and then look at the relative l2-norm deviation.

Cool.  I suppose a (much) harder problem is to rank between cases as opposed to ranking individual curves **within** a case.  In other words, if a PR generates O(30) regression test failures, which are 5 cases that have the "worst" failures?

**Comment by akva2** [comment]
well, fundamentally isn't that the the same problem? you'd just have to sum (or max or sum of square) whatever measure you use to rank within a case to rank between cases.  it does however mean i would to change the approach completely as currently the python script only see a single case at a time and then the bash script loops over cases and calls the python script.

**Comment by steink** [comment]
> okay have added a rather non-rigorous attempt at sorting by eyeball norm order. Specifically linearly resample the new simulation on reference grid and then look at the relative l2-norm deviation. It kinda works, there is no guarantee it will find _the_ worst case and put it first but it will at least put things that obviously deviate in eyeball norm before those that do not. I see no point in limiting the number of graphs output, ignoring those you don't care about is cheap.

I think this look really good.

**Comment by akva2** [comment]
jenkins build this opm-common=4184 failure_report please

**Comment by akva2** [comment]
okay I have added a stab at ranking badness between tests as well. It basically uses the max of each test to rank between the tests. This ranking is reflected in the filenames of the pdf's uploaded to jenkins. 

I see the usefulness but I also urge you not to blindly rely on any of this ranking as the measure is far from perfect.

**Comment by akva2** [comment]
jenkins build this opm-common=4184 failure_report please

**Comment by akva2** [comment]
added a better measure based on wasserstein distance. unless there are other comments i now consider this merge worthy.

**Comment by akva2** [comment]
absolutely

**Comment by akva2** [comment]
removed and marked ready for review.

File: tests/make_failure_report.sh [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+
+OPM_TESTS_ROOT=$1
+BUILD_DIR=$2
+RESULT_DIR=$3
+SOURCE_DIR=`dirname "$0"`
+
+FAILED_TESTS=`cat $BUILD_DIR/Testing/Temporary/LastTestsFailed*.log`
+
+mkdir -p $2/failure_report
+cd $2/failure_report
```
bska, line 13: Tiny nit: You already store `$2` in `BUILD_DIR`.  Might it perhaps be more descriptive to replace `$2` with `$BUILD_DIR` here?

tests/make_failure_report.sh
```diff
@@ -0,0 +1,32 @@
...
+rm -f *
+
+for failed_test in $FAILED_TESTS
+do
+  grep -q -E "compareECLFiles" <<< $failed_test
+  test $? -ne 0 && continue
+
+  failed_test=`echo $failed_test | sed -e 's/.*://g' -e 's/\+/./g'`
+  # Extract test properties
+  dir=`dirname "$0"`
```
bska, line 23: This `dir` variable appears to be unused.

bska (approved):
I like it.  Other than a couple of minor nits in `make_failure_report.sh` this looks good to me.

tests/make_failure_report.sh
akva2: done.

akva2:
jenkins build this please

PR #5500, opened 2024-08-02
bump minimum dune version to 2.9

akva2:
Downstream of https://github.com/OPM/opm-common/pull/4157

Closes https://github.com/OPM/opm-simulators/issues/5425

PR #5585, opened 2024-09-07
Prepare OPM for DUNE 2.10

blattms:
DUNE_VERSION_NEWER has been removed. Falling back to DUNE_VERSION_GTE which exists since at least DUNE 2.7.

Removed switches for older version than 2.7.

### DIFF
--- a/opm/models/discretization/common/fvbaseproblem.hh
+++ b/opm/models/discretization/common/fvbaseproblem.hh
@@ -288,7 +288,7 @@ public:
      */
     void gridChanged()
     {
-#if DUNE_VERSION_NEWER(DUNE_GRID, 2, 8)
+#if DUNE_VERSION_GTE(DUNE_GRID, 2, 8)
                 elementMapper_.update(gridView_);
                 vertexMapper_.update(gridView_);
 #else
--- a/opm/models/discretization/common/fvbaseproblem.hh
+++ b/opm/models/discretization/common/fvbaseproblem.hh
@@ -288,13 +288,8 @@ public:
      */
     void gridChanged()
     {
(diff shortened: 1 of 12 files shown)


akva2:
So this is basically the same thing i do in the bump to 2.9 required, except the other way around, ie, it keeps support for older dune.

blattms:
> So this is basically the same thing i do in the bump to 2.9 required, except the other way around, ie, it keeps support for older dune.

I am not sure I understand this comment. The macro DUNE_VERSION_GTE is there from at least 2.7. Hence there is no bumping here. We just make sure that it also compiles with 2,10 which dropped the DUNE_VERSION_NEWER macro.

Note that DUNE_VERSION_NEWER means the at least the specified version is required. SO NEWER is a bit misleading. If that is 2.7 then the version needs to be >=2.7

akva2:
The conditions you introduce is the same starting in dune 2.8, so you basically add stuff to handle 2.7. i just removed the 2.7 support thus we need none of the DUNE_VERSION_NEWER / DUNE_VERSION_GTE stuff.

blattms:
Making this draft until i have a compiling version

blattms:
I still don't get it. I test >=2.7 (GTE)and not >2.7 (GT).

Anyway, as it turns out we do not compile with 2.7 anyway because of #5595

blattms:
I see. Well, it now works with 2,7 -2.10 to make @totto82 happy.

blattms:
I added a commit that completes #5500

akva2 (approved):

akva2:
still fine.

PR #5587, opened 2024-09-09
fix python package versioning

akva2:
i forgot the package version tag in last iteration.

remove opm-models cloning as it is no longer required
optimize build a little by building both bindings in one build directory

### DIFF
--- a/python/generate-pypi-package.sh
+++ b/python/generate-pypi-package.sh
@@ -27,7 +27,6 @@ DIR=`pwd`
 # Setup opm modules
 git clone https://github.com/OPM/opm-common
 git clone https://github.com/OPM/opm-grid
-git clone https://github.com/OPM/opm-models
 git clone https://github.com/OPM/opm-simulators
 git clone https://github.com/OPM/opm-utilities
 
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this please

python/generate-pypi-package.sh
```diff
@@ -38,39 +37,29 @@ mkdir -p /tmp/opm/wheelhouse
...
       rm -rf $tag
     fi
-    mkdir $tag && cd $tag
-    cmake -DPYTHON_EXECUTABLE=${python_versions[$tag]} -DWITH_NATIVE=0 \
-      -DOPM_ENABLE_PYTHON=ON -DOPM_PYTHON_PACKAGE_VERSION_TAG=${VERSION_TAG} ..
+    mkdir $tag && pushd $tag
+    cmake -DPYTHON_EXECUTABLE=${python_versions[$tag]} -DWITH_NATIVE=0 -DBoost_USE_STATIC_LIBS=1 \
+    -DOPM_ENABLE_PYTHON=ON -DOPM_PYTHON_PACKAGE_VERSION_TAG=${VERSION_TAG} -DBLA_STATIC=1 -DBLAS_LIBRARIES=/usr/lib64/libblas.a -DSUITESPARSE_USE_STATIC=1 -DCMAKE_DISABLE_FIND_PACKAGE_QuadMath=1 ..
+
+    cmake --build . --target opmcommon_python simulators -- -j${BUILD_JOBS}
```
bska, line 48: If you're changing this anyway, you may as well use the cmake utility's `--parallel` option to write this as
```sh
cmake --build . --target ... --parallel ${BUILD_JOBS}
```
instead.  This will pass the appropriate job option to the underlying build tool instead of assuming that the underlying build tool has a `-j` option.

bska (commented):
No real input on the packaging aspects here, just one minor detail concerning how we invoke the cmake utility.

python/generate-pypi-package.sh
akva2: done.

bska (approved):
Thanks a lot for the updates.  Like I said, I don't know enough about the packaging to comment on that aspect, but I trust that this is what you need.  I'll merge into master.

PR #5588, opened 2024-09-09
Move python initialization function

akva2:
This saves us hundreds of MB per simulator object.

The entire blackoil simulator has been instanced (instanced, not built mind you) in every single simulator object due to the member type.


akva2:
jenkins build this please

bska (approved):

PR #5590, opened 2024-09-10
Convergence monitors

jakobtorben:
# Implement Convergence Monitoring

This PR introduces a new convergence monitoring feature to improve the robustness and efficiency of the simulator. It is based on the following publication:

Lie, K., Moyner, O., Klemetsdal, Ø., Skaflestad, B., Moncorgé, A., & Kippe, V. (2024). Enhancing Performance of Complex Reservoir Models via Convergence Monitors. ECMOR, 2024(1), 1-9. (https://doi.org/10.3997/2214-4609.202437057)

The convergence monitoring system tracks the convergence behaviour across iterations, applying penalties for non-convergence. If the total penalty count exceeds the specified cut-off limit, the simulator will cut the timestep.

This feature allows early-exiting for steps that are not converging, saving wasted iterations and assembly.

This is the first version that will be iterated on before considering to merge it.

### DIFF
--- a/opm/simulators/flow/BlackoilModel.hpp
+++ b/opm/simulators/flow/BlackoilModel.hpp
@@ -392,6 +392,7 @@ namespace Opm {
                 // For each iteration we store in a vector the norms of the residual of
                 // the mass balance for each active phase, the well flux and the well equations.
                 residual_norms_history_.clear();
+                total_penaltyCard_.reset();
                 current_relaxation_ = 1.0;
                 dx_old_ = 0.0;
                 convergence_reports_.push_back({timer.reportStepNum(), timer.currentStepNum(), {}});
--- a/opm/simulators/flow/BlackoilModel.hpp
+++ b/opm/simulators/flow/BlackoilModel.hpp
@@ -398,6 +398,7 @@ namespace Opm {
                 // the mass balance for each active phase, the well flux and the well equations.
                 residual_norms_history_.clear();
                 total_penaltyCard_.reset();
+                prev_above_tolerance_ = 0;
                 prev_distance_ = std::numeric_limits<double>::infinity();
                 current_relaxation_ = 1.0;
                 dx_old_ = 0.0;
@@ -1117,33 +1118,35 @@ namespace Opm {
 
     void checkCardPenalty(ConvergenceReport& report, int iteration)
     {
-        if (iteration > 0) {
-            const auto& current_metrics = report.reservoirConvergence();
-            auto distances = std::vector<double>(current_metrics.size(), 0.0);
-            int current_above_tolerance = 0;
 
-            for (size_t i = 0; i < current_metrics.size(); ++i) {
-                distances[i] = std::max(std::log(current_metrics[i].value()), 0.0);
-                    // Count number of metrics above tolerance
-                    if (current_metrics[i].value() > current_metrics[i].tolerance()) {
-                        current_above_tolerance++;
-                    }
+        const auto& current_metrics = report.reservoirConvergence();
+        auto distances = std::vector<double>(current_metrics.size(), 0.0);
+        int current_above_tolerance = 0;
+
+        for (size_t i = 0; i < current_metrics.size(); ++i) {
+            distances[i] = std::max(std::log10(current_metrics[i].value()/current_metrics[i].tolerance()), 0.0);
+                // Count number of metrics above tolerance
+                if (current_metrics[i].value() > current_metrics[i].tolerance()) {
+                    current_above_tolerance++;
                 }
+            }
 
+        // use L1 norm of the distances vector
+        double current_distance = std::accumulate(distances.begin(), distances.end(), 0.0);
+
+        if (iteration > 0) {
             // Add penalty if number of metrics above tolerance has increased
             if (current_above_tolerance > prev_above_tolerance_) {
                 report.addNonConvergedPenalty();
             }
 
-            // use L1 norm of the distances vector
-            double current_distance = std::accumulate(distances.begin(), distances.end(), 0.0);
-
             if (current_distance > param_.convergence_monitoring_decay_factor_ * prev_distance_) {
                 report.addDistanceDecayPenalty();
             }
-            prev_distance_ = current_distance;
-            prev_above_tolerance_ = current_above_tolerance;
         }
+
+        prev_distance_ = current_distance;
+        prev_above_tolerance_ = current_above_tolerance;
 
         if (report.wellFailures().size() > 0) {
             report.addLargeWellResidualsPenalty();
--- a/opm/simulators/flow/BlackoilModel.hpp
+++ b/opm/simulators/flow/BlackoilModel.hpp
@@ -1154,12 +1154,10 @@ namespace Opm {
 
         total_penaltyCard_ += report.getPenaltyCard();
 
-
-
         if (param_.convergence_monitoring_ && (total_penaltyCard_.total() > param_.convergence_monitoring_cutoff_)) {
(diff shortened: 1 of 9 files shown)


jakobtorben:
This first version implements the following convergence monitors:

- Distance decay: Define the distance from convergence as a vector of d_i = max(log(r_i ), 0), for each of the convergence metrics of the reservoir. Calculate the L1 norm of the distance vector, and add add a penalty card if the current distance norm is greater than the previous distance norm multiplied by some decay factor (default 0.75): d^k > σ d^k−1. 
- Degradation of reservoir metrics: Add a penalty for each of the metrics (here CNV and MB) that have increased from the previous iteration
- Unconverged wells: Add a penalty card if there are unconverged wells.

If the total penalty cards if above a given cut-off limit (default 30), cut the timestep.

I tested this on Norne, where we observe a slight decrease in nonlinear and linear iterations. But other cases are probably more suited since Norne does not fail a lot to begin with. (Ignore the zero wasted, I am not exiting the timestep cut gracefully yet).

image

opm/simulators/timestepping/ConvergenceReport.hpp
```diff
@@ -53,6 +53,31 @@ namespace Opm
...
+
+            PenaltyCard& operator+=(const PenaltyCard& other) {
+                nonConverged += other.nonConverged;
+                distanceDecay += other.distanceDecay;
+                largeWellResiduals += other.largeWellResiduals;
+                return *this;
+            }
+
+            template <typename Serializer>
+            void serializeOp(Serializer& serializer)
```
akva2, line 80: maybe this will change, but I don't see any situation where this needs to be serialized, obviously not in the initial eclipse state broadcast, and as restart-serialization happens at report steps, there is also no reason to serialize as afaict this is reset at each nonlinear iteration, right?

jakobtorben: You are perhaps correct. I just followed the class structure of the other classes but was unsure if I actually needed this part. Subject to change, I am planning that for each report there is a penalty card. Such that at the end of the simulation, I can add the counts to the INFORITER (or similar) for analysis. Do I need the serializer for that?

jakobtorben:
In the second version, the implementation should be the same as the paper, given that the tolerance for adding a card for too large well residual, is the same as OPM default such that the well is unconverged. The reporting has also been fixed, such that the wasted iterations coming from the convergence monitoring is counted. (This fix also involved making sure that failed iterations from NaN and too large residuals errors are counted).

When tested on Norne, the results are currently similar to without using convergence monitoring

image

jakobtorben:
Fixed some bugs and added the penalty counts to the INFOITER file for analysis. 

The INFOITER file was used to analyse the convergence behaviour and cut-off values, using a tool similar to the paper.

Norne_step_275

Using this tool we can also estimate the number of iterations saved if using convergence monitoring. Which can be used to find the optimal parameters to use for a specific case:
NORNE_fraction_of_iterations_remaining_after_early_exit

And number of incorrectly aborted timesteps:

NORNE_number_of_incorrectly_aborted_timesteps


Optimal parameters found at cut-off 14 and distance decay factor 0.60, which should give an estimated number of Newton iterations as a factor of 0.989. The small reduction is likely due to Norne not failing a lot to begin with for OPM.

Using these optimal parameters, we can run OPM with convergence monitoring to see if we get any improvements. Here run with relaxed CNV and MB tol equal to original tol to match the format used in the analysis tool.

image

From the results, we see a small reduction in Newton iterations and runtime, as expected from our analysis. Better results are likely achieved on cases with more failed timesteps.

opm/simulators/timestepping/ConvergenceReport.hpp
bska: Just picking up this point now, sorry for the delayed response.

> I don't see any situation where this needs to be serialized,

I switched to using the `serializeOp()` protocol for the `ConvergenceReport` object in commit 0b40277e0 (PR #5338).  As long as this information is intended for output to the `.INFOITER` file, it needs to have a `serializeOp()` member function.

jakobtorben:
Depends on https://github.com/OPM/opm-common/pull/4244.

jakobtorben:
Note that the well convergence metric is not used at the moment. But the plan is to also include well convergence metrics in the convergence monitoring. However, this requires two things:
- WellConvergenceMetric must to be extended to multi-segment wells. 
- Some logic needs to be added to deal with the fact that number of wells increases during the simulation, which needs to be dealt with when comparing the number of unconverged residuals to the previous iteration.

jakobtorben:
jenkins build this opm-common=4244 please

opm/simulators/timestepping/ConvergenceReport.hpp
```diff
@@ -51,6 +51,39 @@ namespace Opm
             Normal     = 1,
             TooLarge   = 2,
             NotANumber = 3,
+            ConvergenceMonitorFailure = 4,
```
atgeirr, line 54: The severityOfWorstFailure() function expects severity of error to be worse for higher numbers. How severe is a failure due to the monitoring saying enough is enough? I would say greater than Normal (since we continue iterating for Normal) but less than the exploding residuals or NaNs, so put it at 2, resort and renumber? And maybe insert the comment I should have long ago:
```
// More severe problems should have higher numbers
```
or something similar.

opm/simulators/timestepping/AdaptiveTimeStepping.hpp
```diff
@@ -247,6 +247,10 @@ void registerAdaptiveParameters();
                     logException_(e, solverVerbose_);
                     // since linearIterations is < 0 this will restart the solver
                 }
+                catch (const ConvergenceMonitorFailure& e) {
+                    causeOfFailure = "Convergence monitor failure";
```
atgeirr, line 251: All the others set the substepReport first, please flip order to avoid unnecessary differences.

opm/simulators/flow/BlackoilModel.hpp
```diff
@@ -1108,6 +1116,53 @@ namespace Opm {
...
+            report.addLargeWellResidualsPenalty();
+        }
+
+        total_penaltyCard_ += report.getPenaltyCard();
+
+
+
+        if (param_.convergence_monitoring_ && (total_penaltyCard_.total() > param_.convergence_monitoring_cutoff_)) {
+            report.setReservoirFailed({ConvergenceReport::ReservoirFailure::Type::ConvergenceMonitorFailure,
+                                        ConvergenceReport::Severity::ConvergenceMonitorFailure,
```
atgeirr, line 1161: Nitpick: indent.

opm/simulators/flow/BlackoilModel.hpp
```diff
@@ -1108,6 +1116,53 @@ namespace Opm {
...
+
+        prev_distance_ = current_distance;
+        prev_above_tolerance_ = current_above_tolerance;
+
+        if (report.wellFailures().size() > 0) {
+            report.addLargeWellResidualsPenalty();
+        }
+
+        total_penaltyCard_ += report.getPenaltyCard();
+
```
atgeirr, line 1156: Nitpick: Never more than one blank line inside a function.

atgeirr (commented):
Nice, just some small fry to fix.

jakobtorben:
jenkins build this please

opm/simulators/timestepping/ConvergenceReport.hpp
jakobtorben: Good point, I agree. Fixed.

atgeirr:
All good, all green! Merging.

PR #5594, opened 2024-09-10
Fix Consistency Check for Maximum Gas Saturation

bska:
The earlier condition
```math
0 \le S_{\mathrm{GU}} < 1
```
was not appropriate and would, for instance, fail the `NORNE_ATW2013` test case in which [`SGU` = 1](https://github.com/OPM/opm-data/blob/eaa2261683a97027e057c2bc49612ad1c86390b3/norne/INCLUDE/RELPERM/SCAL_NORNE.INC#L82) in the unscaled table for saturation region 1.  Revise the condition to be more in line with that of SWU, i.e., as
```math
0 < S_{\mathrm{GU}} \le 1
```
Pointy Hat: @bska


bska:
jenkins build this please

akva2 (approved):
I cannot comment much on the validity of the change, but I can confirm that it implements what is given as the intention.

bska:

PR #5596, opened 2024-09-10
Enable New Saturation Function Consistency Checks

bska:
Currently opt-in by the new parameter `CheckSatfuncConsistency` (command line option `--check-satfunc-consistency`), this PR hooks up the `SatfuncConsistencyCheckManager<>` pass to the `FlowProblem`.  I hope to make the parameter be "opt-out" at some point, but for now the new pass breaks one of the cases in our `benchmark` suite.

The new parameter `--num-satfunc-consistency-sample-points` allows the user to select the maximum number of reported failures for each individual consistency check.  By default, the simulator will report at most five failures for each check.

We check the unscaled curves if the run does not activate the end-point scaling option and the scaled curves otherwise.  At present we're limited to reversible and non-directional saturation functions for the drainage process, but those restrictions will be lifted in due time.

### DIFF
--- a/opm/simulators/flow/FlowProblem.hpp
+++ b/opm/simulators/flow/FlowProblem.hpp
@@ -247,9 +247,14 @@ public:
 
         this->explicitRockCompaction_ = Parameters::Get<Parameters::ExplicitRockCompaction>();
 
-        RelpermDiagnostics relpermDiagnostics{};
-        relpermDiagnostics.diagnosis(simulator.vanguard().eclState(),
-                                     simulator.vanguard().cartesianIndexMapper());
+        if (! Parameters::Get<Parameters::CheckSatfuncConsistency>()) {
+            // User did not enable the "new" saturation function consistency
+            // check module.  Run the original checker instead.  This is a
+            // temporary measure.
+            RelpermDiagnostics relpermDiagnostics{};
+            relpermDiagnostics.diagnosis(simulator.vanguard().eclState(),
+                                         simulator.vanguard().cartesianIndexMapper());
+        }
     }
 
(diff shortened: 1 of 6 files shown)


bska:
I'm creating this in draft mode for two reasons

1. The PR depends on, and contains, the earlier PR #5592
2. I'd like to refine the new parameter, ideally to not exist at all, but that will probably require adjustments to the underlying systems to avoid breaking the regression tests.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please opm-tests=1226

bska:
jenkins build this opm-tests=1226 please

bska:
jenkins build this please

tskille:
This functionality has been tested in Equinor and I'm looking forward to see this going into master.

bska:
The upstream PR was merged into the master branch.  I've pushed an update here that adds a second new command line parameter, `--num-satfunc-consistency-sample-points`, in addition to `--check-satfunc-consistency` that was here when the PR was first created.  The `--num-satfunc-consistency-sample-points` parameter allows the user to request more or fewer sample points than the default value of five and I've updated the PR description to reflect this fact.

I'm marking this PR as "ready for review" now.

bska:
jenkins build this please

bska:
jenkins build this please

bska:
I've pushed an update which adds a few more comments to explain the reasoning and I've moved the MPI barrier closer to where the simulation stops in the case of "critical" consistency failures.

atgeirr:
With the latest changes this is good to go. Merging!

PR #5597, opened 2024-09-11
add ad & TwoPhaseMaterial gpu instantiation tests

multitalentloes:
These test just verify that AD and PiecewiseLinearTwoPhaseMaterials can be used on GPU. This is a part of supporting property evaluation on GPU

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -455,11 +455,11 @@ if (HAVE_CUDA)
   if(NOT CONVERT_CUDA_TO_HIP)
     set(CU_FILES_NEEDING_RELAXED_CONSTEXPR
       tests/gpuistl/test_gpu_ad.cu
-      tests/gpuistl/gpu_linear_two_phase_material.cu
+      tests/gpuistl/test_gpu_linear_two_phase_material.cu
     )
 
-    foreach(file ${LIST_OF_FILES})
-      set_source_file_properties(${file} "--expt-relaxed-constexpr")
+    foreach(file ${CU_FILES_NEEDING_RELAXED_CONSTEXPR})
+        set_source_files_properties(${file} PROPERTIES COMPILE_FLAGS "--expt-relaxed-constexpr")
     endforeach()
   endif()
 endif()
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -448,6 +448,20 @@ if (HAVE_CUDA)
   ADD_CUDA_OR_HIP_FILE(TEST_SOURCE_FILES tests test_cuVector_operations.cpp)
   ADD_CUDA_OR_HIP_FILE(TEST_SOURCE_FILES tests test_safe_conversion.cpp)
   ADD_CUDA_OR_HIP_FILE(TEST_SOURCE_FILES tests test_solver_adapter.cpp)
+  ADD_CUDA_OR_HIP_FILE(TEST_SOURCE_FILES tests test_gpu_ad.cu)
+  ADD_CUDA_OR_HIP_FILE(TEST_SOURCE_FILES tests test_gpu_linear_two_phase_material.cu)
+
+  # for loop providing the flag --expt-relaxed-constexpr to fix some cuda issues with constexpr
+  if(NOT CONVERT_CUDA_TO_HIP)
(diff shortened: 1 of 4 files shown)


multitalentloes:
Jenkins build this please

multitalentloes:
Jenkins build this please

multitalentloes:
Jenkins build this please

CMakeLists_files.cmake
```diff
@@ -32,6 +32,11 @@ macro (ADD_CUDA_OR_HIP_FILE LIST DIR FILE)
 
   if(CUDA_FOUND AND NOT CONVERT_CUDA_TO_HIP)
     list (APPEND ${LIST} "${DIR}/gpuistl/${FILE}")
+
+    # .cu files might struggle with evaluating things at compile-time, this flag solves those issues
+    if(${FILE} MATCHES "\\.cu$")
+      set_source_files_properties("${DIR}/gpuistl/${FILE}" PROPERTIES COMPILE_FLAGS "--expt-relaxed-constexpr")
```
multitalentloes, line 38: This flag is the simplest way to allow cuda to compile code where std::arrays are used in code that will be run on the GPU, because for some reason constexpr indexing is not allowed otherwise... Is this the proper way to add compilation flags in OPM, and does this look okay @akva2

akva2: yes this is okay, although i'd like to see a more focused approach, ie, a list with files where it is needed rather than bulk for every .cu file (unless it is required for all but i don't think it it).

multitalentloes: I can add a boolean argument to the function such that the ones needing this flag can pass an extra true to it while the others use false, it will clutter the code a bit since there are many calls to the macro though...

multitalentloes:
Jenkins build this please

CMakeLists_files.cmake
akva2: you can set source file properties whenever. i don't see why this can't be
```cmake
set(LIST_OF_FILES foo1.cu foo2.cu)
if(NOT CONVERT_CUDA_TO_HIP)
  foreach(file ${LIST_OF_FILES})
    set_source_file_properties(${file} blablabla)
  endforeach()
endif()
```
?

multitalentloes:
Jenkins build this please

CMakeLists_files.cmake
akva2: the annoyance is actually a big part of my reasoning for not doing that but okay, i won't insist.

multitalentloes:
Jenkins build this please

CMakeLists_files.cmake
multitalentloes: I trust your judgement and have rewritten it to list the problematic .cu files and add the property explicitly without change the existing hipification macro

tests/gpuistl/test_gpu_linear_two_phase_material.cu
```diff
@@ -0,0 +1,102 @@
...
+
+    ValueVector testXs = {-1.0, 0, 0.1, 0.3, 0.5, 0.7, 0.9, 0.99, 1.0, 1.1};
+
+    for (Scalar x_i : testXs){
+        auto cpuMadeAd = NorneEvaluation(x_i, 0);
+        NorneEvaluation cpuInterpolatedEval = CPUTwoPhaseMaterialLaw::twoPhaseSatPcnw<NorneEvaluation>(cpuParams, cpuMadeAd);
+
+        NorneEvaluation* gpuAdInput;
+        NorneEvaluation* gpuAdResOnGPU;
+        NorneEvaluation gpuAdResOnCPU[1];
```
kjetilly, line 90: again no need to use the array syntax here.

kjetilly (changes requested):
Small changes.

multitalentloes:
Jenkins build this please

multitalentloes:
Jenkins build this please

multitalentloes:
All feedback should now have been addressed:)

multitalentloes:
Some warnings will not go away until https://github.com/OPM/opm-common/pull/4229 is merged, waiting for that before rerunning jenkins

kjetilly (approved):
Looks good, merging!

PR #1709, opened 2019-01-14
Deferred Logger: Support Older MPI Implementations

bska:
The MPI_Pack() function does not support pointers-to-const until MPI-3.

This is the same type of change as the earlier PR #1637.

PR #1637, opened 2018-11-15
Make MPI calls in gatherConvergenceReport with void*.

blattms:
Fixes compilation issues of PR #1612 with open-mpi 1.6.5 discovered by @bska.
Compiler complained about not being able to transform const void* to void*
as needed by MPI.

PR #5598, opened 2024-09-11
MPIPacker: mark buffer const in unpacker functions

akva2:
Same as https://github.com/OPM/opm-common/pull/4189 but for the mpi packer. Also contains a commit that is required by the former since it is extending the mempacker (hosted here as it is only used for restart serialization and we don't want to pull in boost date_time in opm-common in case you wonder).


bska:
Cool.  I suppose the (pertinent) MPI implementations have caught up enough that we don't need to care about #1637 or #1709 or similar.

akva2:
yeah, those PRs problems related to the data, not the buffer we read the data from. The buffer read from is already const void* in the MPI spec.

bska (approved):
This looks good to me and I'll merge into master.  Thanks a lot!

PR #5600, opened 2024-09-11
compositional cleaning up 

GitPaean:
This is a cleaning up version of the compositional implementation. 

It includes two commits now. 

The first commit is to split the black oil related from FlowProblem and FlowBaseProblem, so that some code can be used for compositional simulation. 

I do not consider this splitting has been done complete or perfectly due to the complexity of the components involved in the black oil simulator. More work will be done along the way and gradually one component/unit by one, which will reduce the difficulties and we can do better that way. 

The second commit is to add a compositional simulator to run from a DATA file.

I will foresee refactoring will be done so that more functionality can be handled by the compositional simulator. 

With the refactoring, the compositional simulator only uses hundreds lines of code compared to the thousands of lines of code duplication. 

The two commits are rather independent, please let me know whether you want me to split it two PRs.


GitPaean:
The compositional part requires upstream PR https://github.com/OPM/opm-common/pull/4045

akva2:
imho it's good to get this as soon as it builds and nothing breaks. then a more iterative approach can be taken for the rest of the refactoring which is good for all parties.

GitPaean:
> imho it's good to get this as soon as it builds and nothing breaks. then a more iterative approach can be taken for the rest of the refactoring which is good for all parties.

I agree.

GitPaean:
It is green from jenkins test triggered in opm-common

https://ci.opm-project.org/job/opm-common-PR-builder/7435/

CMakeLists_files.cmake
```diff
@@ -750,14 +750,16 @@ list (APPEND PUBLIC_HEADER_FILES
...
   opm/simulators/flow/FlowBaseVanguard.hpp
   opm/simulators/flow/FlowGenericProblem.hpp
   opm/simulators/flow/FlowGenericProblem_impl.hpp
   opm/simulators/flow/FlowGenericVanguard.hpp
   opm/simulators/flow/FlowMain.hpp
   opm/simulators/flow/FlowProblem.hpp
+  opm/simulators/flow/FlowProblemBlackoil.hpp
   opm/simulators/flow/FlowProblemParameters.hpp
-  opm/simulators/flow/FlowProblemProperties.hpp
+  opm/simulators/flow/FlowProblemBlackoilProperties.hpp
```
akva2, line 763: keep this in alpha order

flowexperimental/flowexp.hpp
```diff
@@ -126,9 +126,9 @@ struct LinearSolverBackend<TypeTag, TTag::FlowExpTypeTag> {
 namespace Opm {
 
 template <class TypeTag>
-class FlowExpProblem : public FlowProblem<TypeTag> //, public FvBaseProblem<TypeTag>
+class FlowExpProblem : public FlowProblemBlackoil<TypeTag> //, public FvBaseProblem<TypeTag>
 {
-    typedef FlowProblem<TypeTag> ParentType;
+    typedef FlowProblemBlackoil<TypeTag> ParentType;
```
akva2, line 131: since you are touching this line change to using

akva2 (approved):
Nothing particular bad spotted. Let's do it.

PR #5602, opened 2024-09-12
BUGFIX fix injection bhp pressure for CO2/H2 STORE

totto82:
The b factor was doubled up for the gas component if DISGASW and VAPWAT was activated. Should only affect the bhp calculations, not the reservoir results itself.

### DIFF
--- a/opm/simulators/wells/StandardWell_impl.hpp
+++ b/opm/simulators/wells/StandardWell_impl.hpp
@@ -279,33 +279,39 @@ namespace Opm
 
             // compute volume ratio between connection at standard conditions
             Value volumeRatio = bhp * 0.0; // initialize it with the correct type
-;
+
             if (FluidSystem::enableVaporizedWater() && FluidSystem::enableDissolvedGasInWater()) {
                 disOilVapWatVolumeRatio(volumeRatio, rvw, rsw, pressure,
                                         cmix_s, b_perfcells_dense, deferred_logger);
+                // DISGASW only supported for gas-water CO2STORE/H2STORE case
+                // and the simulator will throw long before it reach to this point in the code
+                // For blackoil support of DISGASW we need to add the oil component here
+                assert(FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx));
+                assert(FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx));
+                assert(!FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx));
             } else {
+
                 if (FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx)) {
                     const unsigned waterCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::waterCompIdx);
                     volumeRatio += cmix_s[waterCompIdx] / b_perfcells_dense[waterCompIdx];
                 }
-            }
 
-            if constexpr (Indices::enableSolvent) {
-                volumeRatio += cmix_s[Indices::contiSolventEqIdx] / b_perfcells_dense[Indices::contiSolventEqIdx];
-            }
-
-            if (FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx) && FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx)) {
-                gasOilVolumeRatio(volumeRatio, rv, rs, pressure,
-                                  cmix_s, b_perfcells_dense, deferred_logger);
-            }
-            else {
-                if (FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx)) {
-                    const unsigned oilCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::oilCompIdx);
-                    volumeRatio += cmix_s[oilCompIdx] / b_perfcells_dense[oilCompIdx];
+                if constexpr (Indices::enableSolvent) {
+                    volumeRatio += cmix_s[Indices::contiSolventEqIdx] / b_perfcells_dense[Indices::contiSolventEqIdx];
                 }
-                if (FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx)) {
-                    const unsigned gasCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx);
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -304,8 +304,10 @@ namespace Opm
                     volumeRatio += cmix_s[oilCompIdx] / b_perfcells_dense[oilCompIdx];
                 }
                 if (FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx)) {
-                    const unsigned gasCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx);
-                    volumeRatio += cmix_s[gasCompIdx] / b_perfcells_dense[gasCompIdx];
+                    if (! (FluidSystem::enableVaporizedWater() && FluidSystem::enableDissolvedGasInWater())) {
+                        const unsigned gasCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::gasCompIdx);
```
GitPaean, line 308: It is a good catch, but maybe you can put this in the else branch of this if condition above.  The logic here is rather complicated, that is why we had the fallout in the first place. 

```c++
            if (FluidSystem::enableVaporizedWater() && FluidSystem::enableDissolvedGasInWater()) {
                disOilVapWatVolumeRatio(volumeRatio, rvw, rsw, pressure,
                                        cmix_s, b_perfcells_dense, deferred_logger);
            } else {
```

totto82:
The test failure is expected since bhp is changed. i.e. this will need an update of data after review.

opm/simulators/wells/StandardWell_impl.hpp
totto82: Seems like a good idea. I will try to improve the code logic.

totto82:
jenkins build this please

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -279,36 +279,42 @@ namespace Opm
...
             } else {
+
                 if (FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx)) {
                     const unsigned waterCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::waterCompIdx);
                     volumeRatio += cmix_s[waterCompIdx] / b_perfcells_dense[waterCompIdx];
                 }
+
+                if (FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx) && FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx)) {
+                    gasOilVolumeRatio(volumeRatio, rv, rs, pressure,
+                                    cmix_s, b_perfcells_dense, deferred_logger);
```
akva2, line 299: meh

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -279,36 +279,42 @@ namespace Opm
...
+                } else {
+                    if (FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx)) {
+                        const unsigned oilCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::oilCompIdx);
+                        volumeRatio += cmix_s[oilCompIdx] / b_perfcells_dense[oilCompIdx];
+                    }
+                    if (FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx)) {
+                        if (! (FluidSystem::enableVaporizedWater() && FluidSystem::enableDissolvedGasInWater())) {
```
GitPaean, line 306: `if (! (FluidSystem::enableVaporizedWater() && FluidSystem::enableDissolvedGasInWater())) ` is not needed here.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -279,36 +279,42 @@ namespace Opm
 
             // compute volume ratio between connection at standard conditions
             Value volumeRatio = bhp * 0.0; // initialize it with the correct type
-;
+
             if (FluidSystem::enableVaporizedWater() && FluidSystem::enableDissolvedGasInWater()) {
                 disOilVapWatVolumeRatio(volumeRatio, rvw, rsw, pressure,
                                         cmix_s, b_perfcells_dense, deferred_logger);
+                // only supported for gas-water case
```
GitPaean, line 286: I leave @bska and @akva2  to comment whether we want to use `assert` here.  

It basically means that we enter a simulation not allowed at the moment, which is have water-gas mixing while we might not have water or gas, or we have oil active.  We do not support that.

akva2: even though we ideally would have failed earlier, using assert for things that can be triggered by user input (files) is never the correct solution.

totto82: The Simulator will throw in the setup face. I added the asserts for the programmer if/when we start thinking about adding disgasw for a blackoil run. I can remove them and add a throw instead.

akva2: if it throws earlier it's all good.

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
All comments are addressed. @GitPaean Can I start the process of updating the data and get this in?

GitPaean (approved):
Thanks for the update. The changes introduced in this PR look fine now.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -279,33 +279,39 @@ namespace Opm
 
             // compute volume ratio between connection at standard conditions
             Value volumeRatio = bhp * 0.0; // initialize it with the correct type
-;
+
             if (FluidSystem::enableVaporizedWater() && FluidSystem::enableDissolvedGasInWater()) {
```
GitPaean, line 283: This is not in this PR, so I approved the PR. 

A discussion here, should here be 
```c++
if (FluidSystem::enableVaporizedWater() || FluidSystem::enableDissolvedGasInWater()) {
```
??

GitPaean:

GitPaean:
The regression looks mostly fine. The well INJ looks like injecting with constant gas rate and there is some bhp/pressure change.

GitPaean:
jenkins build this opm-tests=1228 please

PR #5603, opened 2024-09-12
fixed: initialize members

akva2:
caused issues for serialization tests in the debug iterator build

### DIFF
--- a/opm/models/blackoil/blackoilsolventparams.hpp
+++ b/opm/models/blackoil/blackoilsolventparams.hpp
@@ -59,30 +59,30 @@ struct BlackOilSolventParams
     void initFromState(const EclipseState& eclState, const Schedule& schedule);
 #endif
 
-    BrineCo2Pvt brineCo2Pvt_;
-    BrineH2Pvt brineH2Pvt_;
-    Co2GasPvt co2GasPvt_;
-    H2GasPvt h2GasPvt_;
-    SolventPvt solventPvt_;
+    BrineCo2Pvt brineCo2Pvt_{};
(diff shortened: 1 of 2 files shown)


opm/models/blackoil/blackoilsolventparams.hpp
```diff
@@ -59,30 +59,30 @@ struct BlackOilSolventParams
...
+    std::vector<Scalar> tlMixParamViscosity_{}; // Todd-Longstaff mixing parameter for viscosity
+    std::vector<Scalar> tlMixParamDensity_{};   //  Todd-Longstaff mixing parameter for density
+    std::vector<TabulatedFunction> tlPMixTable_{}; // the tlpmixpa(Po) column of the TLPMIXPA table
+
+    bool isMiscible_{};
     bool rsSolw_active_ = false;
-    bool co2sol_;
-    bool h2sol_;
+    bool co2sol_{};
+    bool h2sol_{};
```
bska, line 85: Bools are kind of tricky and we should probably ensure that value intialisation does what we want in this instance.  On the other hand, it's probably no worse than what we already have so I guess it's not a big deal.

bska (approved):
With the possible exception of using value initialisation for `bool`s, this looks good to me and we should merge it.

opm/models/blackoil/blackoilsolventparams.hpp
akva2: sure i can write the = false ;)

akva2:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  Please merge when green.

PR #5604, opened 2024-09-12
fixed: make sure we have valid indices in readInitialCondition_

akva2:
quells warnings building flowexp_comp


akva2:
jenkins build this please

akva2:
jenkins build this please

GitPaean (approved):
I think it is okay. Maybe better to check the whether the phase is active when resize the vectors.  I do not think the current way will do any harm.

PR #5605, opened 2024-09-13
Fix issue with stopped injectors in thermal runs

totto82:


totto82:
jenkins build this please

hnil (commented):
This is probably of given the current temperature model for wells, but there maybe issures:
1)  For injecting perforations of injectors: l 2619 calculate enthalpy potentially with pressure and temperature from different plases i.e. temperature for injection (top well), but pressure from perforation.
2) For injecting perforations for produces (not(injectors) which is stopped) enthalpy is taken from reservoir at the cell... and derivatives is lost. This is probably a hack which is ok until temperature/enthalpy is added to wells. It also maybe "inconsistent" with output which set the "temperature" in the well to an average of producing perforations (but this would give dynamic large coplings in the model).

hnil:
The changes should make the code better! but is cq_r_thermal > 0.0 need?

totto82:
I think this should be merged as it fixes an issue with thermal runs and stopped injectors. As @hnil points out there are other issues that needs to be solved but they could be fixed in a different PR.

PR #5608, opened 2024-09-13
adjust to constructor being marked explicit

akva2:
Downstream of https://github.com/OPM/opm-common/pull/4197


bska (approved):

PR #5609, opened 2024-09-13
In partitioning, only account for wells that will be active at some point in time

vkip:


blattms:
Can we actually anticipate this? If there is an ACTIONX manipulating wells, it seems that we cannot as itwill get executed later and change the rest of the schedule.

We had to deal with this (connections not even represented in the Schedule at this stage) in #5488

vkip:
Requires https://github.com/OPM/opm-common/pull/4198

lisajulia:
jenkins build this please

lisajulia:
> Can we actually anticipate this? If there is an ACTIONX manipulating wells, it seems that we cannot as it will get executed later and change the rest of the schedule.
> 
> We had to deal with this (connections not even represented in the Schedule at this stage) in #5488. But maybe our fix there even fixes that and this PR is good. @lisajulia?

As far as I can tell, with this change, an inactive well can be split across multiple processes, yet if an ACTIONX or PYACTION will turn such a well into an active well, this will cause problems. Is there possibly an option of marking such wells as "permanently inactive"?

vkip:
> > Can we actually anticipate this? If there is an ACTIONX manipulating wells, it seems that we cannot as it will get executed later and change the rest of the schedule.
> > We had to deal with this (connections not even represented in the Schedule at this stage) in #5488. But maybe our fix there even fixes that and this PR is good. @lisajulia?
> 
> As far as I can tell, with this change, an inactive well can be split across multiple processes, yet if an ACTIONX or PYACTION will turn such a well into an active well, this will cause problems. Is there possibly an option of marking such wells as "permanently inactive"?

Not as far as I know, but the idea of [this part](https://github.com/OPM/opm-common/pull/4198/files#diff-2fcef755dddf6c909bcecf8d6451fe22f7582552876326b6722271a7038e04a0R676-R685) is to try to keep track of wells that could possibly be opened in an ACTIONX block (and then these will not be marked as inactive for the partitioning). Seems to work (at least on the test case in opm-tests where a well is opened in ACTIONX). With PYACTION I guess it is possible to break this (but then again I think it is possible to break the current master as well if you, e.g., start adding completions with indices that are determined dynamically?)

lisajulia:
> > > Can we actually anticipate this? If there is an ACTIONX manipulating wells, it seems that we cannot as it will get executed later and change the rest of the schedule.
> > > We had to deal with this (connections not even represented in the Schedule at this stage) in #5488. But maybe our fix there even fixes that and this PR is good. @lisajulia?
> > 
> > 
> > As far as I can tell, with this change, an inactive well can be split across multiple processes, yet if an ACTIONX or PYACTION will turn such a well into an active well, this will cause problems. Is there possibly an option of marking such wells as "permanently inactive"?
> 
> Not as far as I know, but the idea of [this part](https://github.com/OPM/opm-common/pull/4198/files#diff-2fcef755dddf6c909bcecf8d6451fe22f7582552876326b6722271a7038e04a0R676-R685) is to try to keep track of wells that could possibly be opened in an ACTIONX block (and then these will not be marked as inactive for the partitioning). Seems to work (at least on the test case in opm-tests where a well is opened in ACTIONX). With PYACTION I guess it is possible to break this (but then again I think it is possible to break the current master as well if you, e.g., start adding completions with indices that are determined dynamically?)

Yes, then I think it should be fine :) - great! :+1: 
To double check with Matthew as well that we're not missing any keywords: @gdfldm can you have a look at [these keywords than can open a well](https://github.com/OPM/opm-common/pull/4198/files#diff-2fcef755dddf6c909bcecf8d6451fe22f7582552876326b6722271a7038e04a0R676-R685) and let us know if we possibly missed a keyword here! Thanks!

gdfldm:
> can you have a look at [these keywords than can open a well](https://github.com/OPM/opm-common/pull/4198/files#diff-2fcef755dddf6c909bcecf8d6451fe22f7582552876326b6722271a7038e04a0R676-R685) and let us know if we possibly missed a keyword here!

WTEST could reopen a closed well.

There are some keywords not currently supported by Flow, e.g. WELOPENL for LGRs, WCONINJP for polymer injectors, deprecated keyword WELLOPEN, and drilling queues, that could open wells.

vkip:
> > can you have a look at [these keywords than can open a well](https://github.com/OPM/opm-common/pull/4198/files#diff-2fcef755dddf6c909bcecf8d6451fe22f7582552876326b6722271a7038e04a0R676-R685) and let us know if we possibly missed a keyword here!
> 
> WTEST could reopen a closed well.
> 
> There are some keywords not currently supported by Flow, e.g. WELOPENL for LGRs, WCONINJP for polymer injectors, deprecated keyword WELLOPEN, and drilling queues, that could open wells.

Thanks, I'll have a look at these as well. There are also some other issues with this (breaks WBP* calculations, at least), so I'll keep it in draft for now.

opm/simulators/flow/GenericCpGridVanguard.cpp
```diff
@@ -207,6 +208,34 @@ doLoadBalance_(const Dune::EdgeWeightMethod             edgeWeightsMethod,
...
+            const auto& well = schedule.getWell(well_name, schedule.size()-1);
+            std::vector<int> well_on_rank(nranks, 0);
+            for (const auto& conn: well.getConnections()) {
+                if (cellOnRank.count(conn.global_index()) > 0) {
+                    well_on_rank[rank] = 1;
+                    break;
+                }
+            }
+            std::vector<int> well_on_rank_global(nranks, 0);
+            MPI_Allreduce(&well_on_rank[0], &well_on_rank_global[0], nranks, MPI_INT, MPI_MAX, this->grid_->comm());
```
blattms, line 230: To support building without MPI, I would suggest to use`comm.max(&well_on_rank[0], nranks);`

vkip:
> Looks good.
> 
> Regarding your comment: "There has to be a better way". Yes, I think detecting where is inactive wells end up distributed, might be accomplished in CpGrid::loadBalance. We would need to pass in additional information, though.
> 
> Moving it there is probably a good last step, once we know that this works.

Thanks!  Will try to move it there once I have resolved a remaining issue with multi-segment wells that remain shut, but have open connections.

opm/simulators/flow/GenericCpGridVanguard.cpp
vkip: Good point, will update.

vkip:
jenkins build this opm-common=4198 please

vkip:
jenkins build this opm-common=4198 please

vkip:
> > Regarding your comment: "There has to be a better way". Yes, I think detecting where is inactive wells end up distributed, might be accomplished in CpGrid::loadBalance. We would need to pass in additional information, though.
> > Moving it there is probably a good last step, once we know that this works.
> 
> Thanks! Will try to move it there once I have resolved a remaining issue with multi-segment wells that remain shut, but have open connections.

Had a look at handling the inactive wells in CpGrid::loadBalance, and see your point about having to pass additional information. Seems to require quite a few interface changes, so maybe the approach here is just as simple after all? Will leave that up to you to decide. 

Taking it out of draft now since it seems to work (build check green + tested on some field models).

blattms:
jenkins build this opm-common=4198 serial please

opm/simulators/flow/GenericCpGridVanguard.cpp
```diff
@@ -207,6 +208,33 @@ doLoadBalance_(const Dune::EdgeWeightMethod             edgeWeightsMethod,
...
+            const auto& well = schedule.getWell(well_name, schedule.size()-1);
+            std::vector<int> well_on_rank(nranks, 0);
+            for (const auto& conn: well.getConnections()) {
+                if (cellOnRank.count(conn.global_index()) > 0) {
+                    well_on_rank[rank] = 1;
+                    break;
+                }
+            }
+            std::vector<int> well_on_rank_global(nranks, 0);
+            comm.max(&well_on_rank[0], nranks);
```
blattms, line 229:
```suggestion
            int well_on_rank = 0;
            for (const auto& conn: well.getConnections()) {
                if (cellOnRank.count(conn.global_index()) > 0) {
                    well_on_rank = 1;
                    break;
                }
            }
            std::vector<int> well_on_rank_global(nranks, 0);
            comm.allgather(&well_on_rank, 1, well_on_rank_global.data());
```
Your version is correct, too.

Now that I think of it, we could also do this in one communication step as those collective communications are rather expensive.
```c++
std::size_t num_wells = inactive_well_names.size();
std::vector<int> well_on_rank(num_wells, 0);
std::size_t well_idx = 0;
for (const auto& well_name : inactive_well_names) {
  const auto& well = schedule.getWell(well_name, schedule.size()-1);
  for (const auto& conn: well.getConnections()) {
      if (cellOnRank.count(conn.global_index()) > 0) {
          well_on_rank[well_idx] = 1;
          break;
      }
  }
  ++well_idx;
}
// values from rank i will be at indices i*num_wells, i*num_wells + 1, ..., (i+1) * num_wells -1
std::vector<int> global_well_on_rank(num_wells * nranks, 0);
comm.allgather(well_on_rank.data(), inactive_well_names.size(), well_on_rank_global.data());
well_idx = 0;
for (const auto& well_name : inactive_well_names) {
  for (int i=0; i<nranks; ++i) {
     if (well_on_rank[i*num_wells + well_idx]) {
        parallelWells.emplace_back(well_name, i);
     }
  }
  ++well_idx;
}

File: opm/simulators/flow/GenericOutputBlackoilModule.cpp [hunk 3]
```diff
@@ -391,6 +391,54 @@ outputFipAndResvLog(const Inplace& inplace,
...
+    }
+}
+
+template<class FluidSystem>
+void GenericOutputBlackoilModule<FluidSystem>::
+collectRftMapOnRoot(std::map<std::size_t, Scalar>& local_map, const Parallel::Communication& comm) {
+
+    const auto mapsize = local_map.size();
+    std::vector<int> keys;
+    std::vector<Scalar> values;
```
blattms, line 410: Dune can magically communicate pairs, too. Hence we could use 
```c++
std::vector<std::pair<int,Scalar>> pairs(local_map.begin(), local_map.end());
```
and pass that to `gatherv`

opm/simulators/flow/GenericOutputBlackoilModule.cpp
```diff
@@ -391,6 +391,54 @@ outputFipAndResvLog(const Inplace& inplace,
...
+    std::tie(all_values, std::ignore) = Opm::gatherv(values, comm, 0);
+    assert(all_keys.size() == all_values.size());
+
+    // Insert/update map values on root
+    if (comm.rank() == 0) {
+        for (auto i=static_cast<std::size_t>(offsets[1]); i<all_keys.size(); ++i) {
+            const auto index = all_keys[i];
+            if (local_map.count(index)>0) {
+                const Scalar prev_value = local_map[index];
+                local_map[index] = std::max(prev_value, all_values[i]);
```
blattms, line 432: Maybe prevent multiple searches?
```suggestion
            if (auto candidate = local_map.find(index); candidate != local_map.end() && candidate->first == index) {
                const Scalar prev_value = candidate->second;
                candidate->second = std::max(prev_value, all_values[i]);
```
Will max do the correct thing on all occasions?

blattms (commented):
This looks already really good. Just a few comments on possible improvements.

I am wondering a bit about the max usage for connection pressures. This is for wells, right. I would assume that we only consider perforations in the interior of the local grid. In that case that code might never be executed.

opm/simulators/flow/GenericCpGridVanguard.cpp
vkip: Thanks, this is better. Will update!

opm/simulators/flow/GenericOutputBlackoilModule.cpp
vkip: Nice, thanks!

opm/simulators/flow/GenericOutputBlackoilModule.cpp
vkip: I believe so - the gathered values are set via processElement via EclWriter, which iterates only over the interior, so if my understanding is correct the only non-zero value comes from the partition to which the connection truly belongs?

vkip:
Thanks for good suggestions! Updated accordingly now (I think..)

vkip:
jenkins build this opm-common=4198 please

blattms:
Perfect,thanks a lot for the changes and the nice improvement.

PR #5611, opened 2024-09-16
Print gpus used on simulator startup

multitalentloes:
Slight improvement in use of gpuISTL to show what device is used in the simulation
TODO: make sure we get the output from all the ranks

### DIFF
--- a/opm/simulators/linalg/gpuistl/set_device.cpp
+++ b/opm/simulators/linalg/gpuistl/set_device.cpp
@@ -17,33 +17,34 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
 #include <config.h>
+#include <sstream>
 #include <cuda_runtime.h>
 #include <opm/common/OpmLog/OpmLog.hpp>
+#include <opm/simulators/flow/FlowGenericVanguard.hpp>
 #include <opm/simulators/linalg/gpuistl/detail/gpu_safe_call.hpp>
 #include <opm/simulators/linalg/gpuistl/set_device.hpp>
+
 namespace Opm::gpuistl
 {
 void
 setDevice(int mpiRank, [[maybe_unused]] int numberOfMpiRanks)
 {
-
     int deviceCount = -1;
     [[maybe_unused]] auto cuError = cudaGetDeviceCount(&deviceCount);
 
     if (deviceCount <= 0) {
-        // If they have CUDA enabled (ie. using a component that needs CUDA, eg. gpubicgstab or CUILU0), this will fail
+        // If they have CUDA/HIP enabled (ie. using a component that needs CUDA, eg. gpubicgstab or CUILU0), this will fail
         // later down the line. At this point in the simulator, we can not determine if CUDA is enabled, so we can only
         // issue a warning.
-        OpmLog::warning("Could not find any CUDA devices.");
+        OpmLog::warning("Could not find any CUDA/HIP devices.");
         return;
     }
 
     // Now do a round robin kind of assignment
     // TODO: We need to be more sophistacted here. We have no guarantee this will pick the correct device.
     const auto deviceId = mpiRank % deviceCount;
-    OPM_GPU_SAFE_CALL(cudaDeviceReset());
-    OPM_GPU_SAFE_CALL(cudaSetDevice(deviceId));
-    OpmLog::info("Set CUDA device to " + std::to_string(deviceId) + " (out of " + std::to_string(deviceCount)
-                 + " devices).");
+    OPM_GPU_WARN_IF_ERROR(cudaDeviceReset());
+    OPM_GPU_WARN_IF_ERROR(cudaSetDevice(deviceId));
 }
+
 } // namespace Opm::gpuistl
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -244,6 +244,8 @@ endif()
 
 # add these files if we should compile the hip code
 if (HAVE_CUDA)
+  list(APPEND MAIN_SOURCE_FILES opm/simulators/linalg/gpuistl/device_management.hpp) # should not be hipified to make main independant of library
+  ADD_CUDA_OR_HIP_FILE(MAIN_SOURCE_FILES opm/simulators/linalg device_management.cpp)
   ADD_CUDA_OR_HIP_FILE(MAIN_SOURCE_FILES opm/simulators/linalg detail/CuBlasHandle.cpp)
   ADD_CUDA_OR_HIP_FILE(MAIN_SOURCE_FILES opm/simulators/linalg detail/gpusparse_matrix_operations.cu)
   ADD_CUDA_OR_HIP_FILE(MAIN_SOURCE_FILES opm/simulators/linalg detail/CuSparseHandle.cpp)
--- a/opm/simulators/flow/Main.cpp
+++ b/opm/simulators/flow/Main.cpp
@@ -35,10 +35,6 @@
 #include <opm/simulators/utils/DamarisOutputModule.hpp>
 #endif
 
-#if HAVE_CUDA
-#include <opm/simulators/linalg/gpuistl/set_device.hpp>
-#endif
-
 namespace Opm {
 
 Main::Main(int argc, char** argv, bool ownMPI)
@@ -163,7 +159,7 @@ void Main::initMPI()
     }
 
 #if HAVE_CUDA
-    Opm::gpuistl::setDevice(FlowGenericVanguard::comm().rank(), FlowGenericVanguard::comm().size());
+    Opm::gpuistl::setDevice();
 #endif
 
 #endif // HAVE_MPI
--- a/opm/simulators/flow/Main.hpp
+++ b/opm/simulators/flow/Main.hpp
@@ -72,6 +72,10 @@
 #include <opm/simulators/utils/ParallelEclipseState.hpp>
 #endif
 
+#if HAVE_CUDA
+#include <opm/simulators/linalg/gpuistl/device_management.hpp>
+#endif
+
 #if HAVE_DAMARIS
 #include <opm/simulators/utils/DamarisKeywords.hpp>
 #endif
@@ -425,6 +429,10 @@ protected:
             exitCode = EXIT_FAILURE;
             return false;
         }
+
+#if HAVE_CUDA
+    Opm::gpuistl::printDevice();
+#endif
 
         exitCode = EXIT_SUCCESS;
         return true;
--- a/opm/simulators/linalg/gpuistl/device_management.cpp
+++ b/opm/simulators/linalg/gpuistl/device_management.cpp
@@ -1,0 +1,77 @@
+/*
+  Copyright 2024 SINTEF AS
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <config.h>
+
+#include <opm/simulators/flow/FlowGenericVanguard.hpp>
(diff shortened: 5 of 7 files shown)


multitalentloes:
Jenkins build this hip please

multitalentloes:
Something about this PR causes the build system to not hipify anything, not sure why

opm/simulators/linalg/gpuistl/set_device.cpp
```diff
@@ -43,7 +44,28 @@ setDevice(int mpiRank, [[maybe_unused]] int numberOfMpiRanks)
...
-                 + " devices).");
 }
+
+void
+printDevice (int mpiRank, int numberOfMpiRanks)
+{
+    int deviceCount = -1;
+    OPM_GPU_SAFE_CALL(cudaGetDeviceCount(&deviceCount));
+
+    const auto deviceId = mpiRank % deviceCount;
```
blattms, line 55: This looks like there is an association of MPI ranks to GPU devices.
Out of curiosity: How are we enforcing this associating?

Basically, we are assuming that e.g. ranks 0,4,8,12,... are on the same node if we have for gpus in the calculation. There is of course some kind of control where to start MPI ranks. But it is actually really hard to do this. So the default would be that we need to detect which ranks are near the GPU (read on the same node) to reduce latency. Probably there is some machanism for that...

blattms (commented):
Looks good. I just have a question on the general assumptions here.

blattms:
Some of the headers are not found. Maybe they are missing from the PR?

akva2:
no, that is not the problem. i have explained to tobias in other channels. the problem is that we try to pull hip headers into simulator objects and simulator objects do not depend on the library, hence headers have not been hipified.

opm/simulators/linalg/gpuistl/set_device.cpp
multitalentloes: Having the rank modulo the local device count would mean mapping for instance ranks {0, 1, 2, 3} to the same node if we have four gpus that we want to use, not {0, 4, 8, 12}? In the latter case the result of the modulo operation would always be 0 and these four processes would just use device0.

multitalentloes:
Jenkins build this hip please

multitalentloes:
Jenkins build this please

opm/simulators/linalg/gpuistl/set_device.cpp
blattms: My bad. I should give back my degrees.

My point was  that I am doubting that one can assume any locations for the MPI ranks at all (unless enforcing it from the outside). Hence a rank associated with a GPU might live on a distant node, if if the rank is close. Depending on the interconnect that might make a difference. But I guess you are aware about this.

Seems like you can actually determine local ranks via creating communicators with MPI_COMM_TYPE_SHARED. The you can  get an "on-node" communicator, to do the association in a multi-node-safe manner. See https://medium.com/@jeffrey_91423/binding-to-the-right-gpu-in-mpi-cuda-programs-263ac753d232

multitalentloes: For GPU the only thing that makes sense I think is to use N processes, where N is the number of GPUs (Nodes * gpusPerNode), and then I think filling up the processes in a consecutive way (node1 has {p0, p1, p2, p3}, node2 has {p4, p4...}... will work just fine for now. This is probably not that hard to do with slurm, which I hope we will use in the multigpu case in the foreseeable future? Otherwise I can create an issue requesting more functionality to select gpus so we keep this in mind and handle it later.

kjetilly: Just to add to this: The current way we are assigning the GPUs is very rudimentary, and might not yield the best performance (especially for two socket/CPU NUMA server nodes with multiple GPUs), nor is it guaranteed that the MPI ranks will map that nicely to different nodes/GPUs unless you control the submission. This is also why I wrote:

```C++
   // Now do a round robin kind of assignment
   // TODO: We need to be more sophistacted here. We have no guarantee this will pick the correct device.
```
(typo is highly intentional!)

The simple reason why this hasn't been developed or investigated further because we are not really at a point where running over multiple nodes with multiple GPUs is relevant. However, we are applying for access to LUMI-G, where this will be looked into. The current solution "works well enough" for the current testing we are doing.

opm/simulators/linalg/gpuistl/set_device_indirection.cpp
```diff
@@ -0,0 +1,51 @@
+/*
```
kjetilly, line 1: I am not a fan of this name. First this file includes two functions, one of which has nothing to do with setting the device, secondly, what indirection refers to is a bit unclear (and for a user of said file, maybe not needed).

May I suggest renaming to `device_management.hpp` ? and then include this file in the main file.

The file now named `set_device.hpp` should probably be moved to `detail` (and renamed to something like `device_management`), as you don't want this file to be included from outside `gpuistl` (and the `detail` namespace signals this)

opm/simulators/linalg/gpuistl/set_device.cpp
```diff
@@ -17,33 +17,62 @@
...
-        // issue a warning.
-        OpmLog::warning("Could not find any CUDA devices.");
-        return;
+        // Now do a round robin kind of assignment
+        // TODO: We need to be more sophistacted here. We have no guarantee this will pick the correct device.
+        const auto deviceId = mpiRank % deviceCount;
+        OPM_GPU_SAFE_CALL(cudaDeviceReset());
+        OPM_GPU_SAFE_CALL(cudaSetDevice(deviceId));
     }
+    catch(...){}
```
kjetilly, line 50: I don't think it's a good idea to have a catch all here. At the very least just catch `std::runtime_error` which is what `OPM_GPU_SAFE_CALL` will call, but it is probably a better idea to just call `OPM_GPU_WARN_IF_ERROR` and remove the whole try/catch.

opm/simulators/linalg/gpuistl/set_device.cpp
```diff
@@ -17,33 +17,62 @@
...
+
+        std::ostringstream out;
+        if (numberOfMpiRanks > 0){
+          out << "GPU: " << props.name << ", Compute Capability: " << props.major << "." << props.minor;
+          out << " (device " << std::to_string(deviceId) << " out of " << std::to_string(deviceCount) << ")\n";
+        }
+        else{
+          out << "GPU: " << props.name << ", Compute Capability: " << props.major << "." << props.minor << std::endl;
+        }
+        OpmLog::info(out.str());
```
kjetilly, line 73: This is probably a lot more readable with `fmt::format` to avoid the stringstream.

kjetilly (changes requested):
Mostly stylistic comment, but the catch all is probably something that should be avoided.

opm/simulators/linalg/gpuistl/set_device.cpp
multitalentloes: done

opm/simulators/linalg/gpuistl/set_device.cpp
multitalentloes: updated

opm/simulators/linalg/gpuistl/set_device_indirection.cpp
multitalentloes: done

multitalentloes:
The PR is now updated and I have tried to use deferred logger, which I think is the right choice when wanting output from all MPI ranks, this is not tested on a machine with multiple GPUs or on nvidia cards, which should be tested before merging

multitalentloes:
jenkins build this please

kjetilly (approved):
This looks good now.

bska:
Heads up: Merging this may or may not have broken the build.  At least we're getting the following regression test failure following this PR: https://ci.opm-project.org/job/opm-simulators/2425/testReport/junit/(root)/mpi-hipify/outputdir/.

multitalentloes:
I have discussed this with @akva2, the issue is caused by compiling with HIP when you do not have an AMD gpu. He will look into as I have some other commitments today

kjetilly:
Maybe I'm missing something, but the error message from HIP is just printed as a warning?

akva2:
the problem is that it somehow leads to a div by zero.

kjetilly:
I'm a bit stumped, since Jenkins passed on this PR before I had accepted it, and the callstack looks like it's coming from somewhere completely different? But I'll wait for your investigation.

```
[NON-XML-CHAR-0x1B][0m
[opm-jenkins2:1864847] *** Process received signal ***
[opm-jenkins2:1864847] Signal: Floating point exception (8)
[opm-jenkins2:1864847] Signal code: Integer divide-by-zero (1)
[opm-jenkins2:1864847] Failing at address: 0x924491
[opm-jenkins2:1864847] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x3c050)[0x7fa87574d050]
[opm-jenkins2:1864847] [ 1] /var/lib/jenkins/post-builder/workspace/opm-simulators/mpi-hipify/build-opm-simulators/bin/test_outputdir[0x924491]
[opm-jenkins2:1864847] [ 2] /var/lib/jenkins/post-builder/workspace/opm-simulators/mpi-hipify/build-opm-simulators/bin/test_outputdir[0x721c29]
[opm-jenkins2:1864847] [ 3] /var/lib/jenkins/post-builder/workspace/opm-simulators/mpi-hipify/build-opm-simulators/bin/test_outputdir[0x7245c8]
[opm-jenkins2:1864847] [ 4] /var/lib/jenkins/post-builder/workspace/opm-simulators/mpi-hipify/build-opm-simulators/bin/test_outputdir[0x725d7b]
[opm-jenkins2:1864847] [ 5] /lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.74.0(+0x323fe)[0x7fa90e3fe3fe]
[opm-jenkins2:1864847] [ 6] /lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.74.0(_ZN5boost17execution_monitor13catch_signalsERKNS_8functionIFivEEE+0x14d)[0x7fa90e3fcb0d]
[opm-jenkins2:1864847] [ 7] /lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.74.0(_ZN5boost17execution_monitor7executeERKNS_8functionIFivEEE+0x51)[0x7fa90e3fcb91]
[opm-jenkins2:1864847] [ 8] /lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.74.0(_ZN5boost17execution_monitor8vexecuteERKNS_8functionIFvvEEE+0x31)[0x7fa90e3fcc61]
[opm-jenkins2:1864847] [ 9] /lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.74.0(_ZN5boost9unit_test19unit_test_monitor_t21execute_and_translateERKNS_8functionIFvvEEEm+0x121)[0x7fa90e42b1a1]
[opm-jenkins2:1864847] [10] /lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.74.0(+0x37ce0)[0x7fa90e403ce0]
[opm-jenkins2:1864847] [11] /lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.74.0(+0x37fda)[0x7fa90e403fda]
[opm-jenkins2:1864847] [12] /lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.74.0(_ZN5boost9unit_test9framework3runEmb+0x7f7)[0x7fa90e407f37]
[opm-jenkins2:1864847] [13] /lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.74.0(_ZN5boost9unit_test14unit_test_mainEPFbvEiPPc+0x22e)[0x7fa90e42a10e]
[opm-jenkins2:1864847] [14] /var/lib/jenkins/post-builder/workspace/opm-simulators/mpi-hipify/build-opm-simulators/bin/test_outputdir[0x72eb80]
[opm-jenkins2:1864847] [15] /lib/x86_64-linux-gnu/libc.so.6(+0x2724a)[0x7fa87573824a]
[opm-jenkins2:1864847] [16] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x85)[0x7fa875738305]
[opm-jenkins2:1864847] [17] /var/lib/jenkins/post-builder/workspace/opm-simulators/mpi-hipify/build-opm-simulators/bin/test_outputdir[0x71fe21]```

akva2:
in particular cudaGetDeviceCount doesn't set the result var to 0 if it fails to find any device, so we have the initial -1 value, and hence we did not get the div-by-zero. hipGetDeviceCount(..) however does set it to 0, hence we got the div-by-zero.
and since the regular trigger only builds with cuda, you did not see it before we ran with hip in the post-builder. stuff like this happens, the system caught it, we'll fix it and everybody is happy again.

PR #5612, opened 2024-09-16
Split out vtk module parameters to separate translation units

akva2:
This splits out the parameters for various modules to separate files and adds translation units for them. At the same time rename more .hh files to .hpp for consistency.

Unlikely to be a major reduction in compile time and/or memory usage but I still find it tidier. And since we can...


akva2:
jenkins build this please

opm/models/blackoil/blackoilnewtonmethodparams.cpp
```diff
@@ -0,0 +1,81 @@
...
+template<class Scalar>
+void BlackoilNewtonParams<Scalar>::registerParameters()
+{
+    Parameters::Register<Parameters::DpMaxRel<Scalar>>
+        ("Maximum relative change of pressure in a single iteration");
+    Parameters::Register<Parameters::DsMax<Scalar>>
+        ("Maximum absolute change of any saturation in a single iteration");
+    Parameters::Register<Parameters::PriVarOscilationThreshold<Scalar>>
+        ("The threshold value for the primary variable switching conditions "
+                                                                        "after its meaning has switched to hinder oscilations");
```
bska, line 40: Unexpected indentation here.  Is that an accident?  On a tangential note, maybe we could fix the misprint ("oscilation" ➡️ "oscillation") as well, possibly in follow-up work?

bska (commented):
I admit that I didn't read this particularly carefully, but it appears to be mostly automatic and straightforward.  I did notice one small whitespace nit, however, and an opportunity to fix a misprint in a parameter name.

opm/models/blackoil/blackoilnewtonmethodparams.cpp
akva2: not on purpose, just me not paying attention to IDE autoindent. fixed this and the typo.

akva2:
jenkins build this please

bska (approved):
Thanks a lot for the updates&ndash;this looks good to me now and I'll merge into master.

At some point we should probably fix the misprint in the parameter name too, i.e., `priVarOscilationThreshold_` ➡️ `priVarOscillationThreshold_`, but I expect that will take more testing and is, in fact, a user-visible change which might affect run scripts.

PR #5617, opened 2024-09-18
Make SolverAdapter explicitly depend on the communication object

kjetilly:
This fixes the MPI issues with CUDA/HIP/GPUISTL that were introduced with the `GhostLastMatrixAdapter` operator. The previous design of the SolverAdapter relied on the presence of the function `getCommunication` to determine if one should run the GPU solver with MPI. This updated version now decides this on the type of communication sent in. For a general communication type, it assumes it is an MPI run, while for `Dune::Amg::SequentialInformation`, it defaults to a serial (no MPI) run. 

The upside of this change is that any new, similar issues where the interface of the operator or communication changes will be detected at compile time and not runtime.

### DIFF
--- a/opm/simulators/linalg/gpuistl/SolverAdapter.hpp
+++ b/opm/simulators/linalg/gpuistl/SolverAdapter.hpp
@@ -23,6 +23,7 @@
 
 #include <dune/istl/bcrsmatrix.hh>
 #include <dune/istl/operators.hh>
+#include <dune/istl/paamg/pinfo.hh>
 #include <dune/istl/schwarz.hh>
 #include <dune/istl/solver.hh>
 #include <opm/common/ErrorMacros.hpp>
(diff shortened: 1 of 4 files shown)


kjetilly:
Jenkins build this please

multitalentloes:
This works on my AMD gpu:)

opm/simulators/linalg/gpuistl/SolverAdapter.hpp
```diff
@@ -122,116 +130,120 @@ class SolverAdapter : public Dune::IterativeSolver<X, X>
...
 
-            auto mpiPreconditioner = std::make_shared<GpuBlockPreconditioner<XGPU, XGPU, CudaCommunication>>(
-                preconditionerReallyOnGPU, cudaCommunication);
+        auto scalarProduct = std::make_shared<Dune::ParallelScalarProduct<XGPU, CudaCommunication>>(
+            cudaCommunication, m_opOnCPUWithMatrix.category());
 
-            auto scalarProduct = std::make_shared<Dune::ParallelScalarProduct<XGPU, CudaCommunication>>(
-                cudaCommunication, m_opOnCPUWithMatrix.category());
 
+        // NOTE: Ownsership of cudaCommunication is handled by mpiPreconditioner. However, just to make sure we
```
atgeirr, line 205: Typo: "OwnSership"

opm/simulators/linalg/gpuistl/SolverAdapter.hpp
```diff
@@ -122,116 +130,120 @@ class SolverAdapter : public Dune::IterativeSolver<X, X>
...
+        //       remember this, we add this check. So remember that we hold one count in this scope, and one in the
+        //       GpuBlockPreconditioner instance. We accomedate for the fact that it could be passed around in
```
atgeirr, line 207: Typo: "accomEdate"

atgeirr (commented):
Looks good from my perspective, should be merged when minor quibbles have been addressed.

kjetilly:
Jenkins build this please

bska:
Might it be possible to use the [`Parallel::Communication`](https://github.com/OPM/opm-simulators/blob/71ca4c35d7a64e37d24c5b02ad4d0f929886c5a3/opm/simulators/utils/ParallelCommunication.hpp#L29-L30) type instead of introducing a new template parameter?

kjetilly:
> Might it be possible to use the [`Parallel::Communication`](https://github.com/OPM/opm-simulators/blob/71ca4c35d7a64e37d24c5b02ad4d0f929886c5a3/opm/simulators/utils/ParallelCommunication.hpp#L29-L30) type instead of introducing a new template parameter?

I do not think so, but I might be wrong. Currently the communication object is in the end passed to say the [GPUObliviousSender](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/linalg/gpuistl/GpuOwnerOverlapCopy.hpp#L113) which needs [copyOwnerToAll](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/linalg/gpuistl/GpuOwnerOverlapCopy.hpp#L128) and other functions that, to my knowledge, are not found in `Parallel::Communication` (=`Dune::Communication`?). 

And to be consistent with the rest of the class, it is probably a good idea to keep the template argument for `initSolver`, as both the constructor [`FlexibleSolver`](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/linalg/FlexibleSolver.hpp#L60) and [`initOpPrecSp`](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/linalg/FlexibleSolver.hpp#L81C10) has it as a template argument, and the constructor will call `initSolver` with the communication (a template `class Comm`) as an argument.

kjetilly:
Jenkins build this please

multitalentloes:
How serial version is chosen is now a bit clearer with the new documentation. Looks good to me

atgeirr:
Thanks for the contribution, and for reviewing. Merging!

PR #5618, opened 2024-09-18
setting up logging so flowexp variants can report parsing errors

GitPaean:


GitPaean:
Ideally, that we can make flowexp variants to use the same parser with flow, but I do not have overview for this issue. 

Will discuss with someone potentially understand better in this part. 

But the current code is trying to make sure there is some parsing failure reported on screen for flowexp_blackoil and flowexp_comp.

GitPaean:
jenkins build this please

opm/simulators/utils/readDeck.cpp
```diff
@@ -545,6 +545,16 @@ void Opm::readDeck(Opm::Parallel::Communication    comm,
...
 {
+
+    if (comm.rank() == 0) {
+        std::shared_ptr<Opm::StreamLog> streamLog = std::make_shared<Opm::StreamLog>(std::cout, Opm::Log::StdoutMessageTypes);
+        Opm::OpmLog::addBackend("STDOUT_LOGGER", streamLog);
+        // Set a tag limit of 10 (no category limit).
+        streamLog->setMessageLimiter(std::make_shared<Opm::MessageLimiter>(10));
+        const bool use_color_coding = OpmLog::stdoutIsTerminal();
+        streamLog->setMessageFormatter(std::make_shared<Opm::SimpleMessageFormatter>(use_color_coding));
+    }
```
bska, line 556: At least create a helper function for this part so we don't have to duplicate the implementation here and in the "official" setup function.

bska: On second thought, we should not include defining the logging system as part of reading the input.  Please pull this out of `readDeck()`.

GitPaean: I do not know the history, but in flow, the logging system is setup up in the Main::readDeck() function. But I agree with you, it should be outside of it. Will update.  Just not very familiar with this part of the code. 

```c++
void Main::readDeck(const std::string& deckFilename,
                    const std::string& outputDir,
                    const std::string& outputMode,
                    const bool init_from_restart_file,
                    const bool allRanksDbgPrtLog,
                    const std::string& parsingStrictness,
                    const std::string& actionParsingStrictness,
                    const std::string& inputSkipMode,
                    const std::size_t numThreads,
                    const int output_param,
                    const std::string& parameters,
                    std::string_view moduleVersion,
                    std::string_view compileTimestamp)
{
    auto omode = setupLogging(FlowGenericVanguard::comm(),
                              deckFilename,
                              outputDir,
                              outputMode,
                              outputCout_, "STDOUT_LOGGER", allRanksDbgPrtLog);
```

opm/simulators/utils/readDeck.hpp
```diff
@@ -67,6 +67,9 @@ prepareResultOutputDirectory(const std::string& baseName, const std::filesystem:
 
 std::unique_ptr<ParseContext> setupParseContext(const bool exitOnAllErrors);
 
+// Setup the OpmLog backends for screen output
+void setupStreamLogging(const std::string& stdout_log_id);
```
GitPaean, line 71: I do not think `stdout_log_id` is necessary for this function and also setupLogging() function. but I might misunderstood something here.  Please suggest if there is some opinion here.

GitPaean:
I updated the PR, while not totally sure about where is the most appropriate place to do it.  Please suggest.

GitPaean:
jenkins build this please

bska (approved):
Having looked at the broader context of this, and the implementation of `OpmLog::addBackend()`, I've concluded that this is fine.  I'll merge into master.

PR #5624, opened 2024-09-23
fixed: do not dereference invalid iterator

akva2:
if a well has connections added in an action, the well perf data is not yet initialized in the call to updateEclWells. it will however be reinitialized before the next time step so simply adding a bounds check here avoid the invalid dereference and the connection factor will take effect through the reinitialization.

ref e.g. https://ci.opm-project.org/job/opm-simulators-static-analysis/383/testReport/junit/(root)/debug_iterator/compareSeparateECLFiles_flow_pyaction_gruptree_insert_kw/

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -752,7 +752,11 @@ updateEclWells(const int timeStepIdx,
             auto pdIter = pd.begin();
 
             for (const auto& conn : well.getConnections()) {
-                if (conn.state() != Connection::State::SHUT) {
+                // Iterator needs to be bounds checked in case connections were added
+                // in the triggering actions. In that case the well_perf_data is not
+                // yet resized. The wellStructureChangedDynamically_ below
+                // however will trigger reinitialization on start of next time step.
+                if (conn.state() != Connection::State::SHUT && pdIter != pd.end()) {
                     pdIter->connection_transmissibility_factor = conn.CF();
                     ++pdIter;
                 }


akva2:
jenkins build this please

bska (commented):
On the face of it this looks benign, but I need to review the wider context here.  Please hold any further action here.

lisajulia:
Looks fine :) @akva2: maybe you can add a comment about the newly introduced check and why it is needed?

akva2:
yes precisely, the bool wellStructureChangedDynamically be set to true in the call to updateEclWells() thus triggering that block of code. Have added a comment.

bska (commented):
There's more at play here than it initially appears.  Please hold any further development on this PR until I've had the time to do a thorough review.

bska:
I've looked into this more closely now and as far as I can tell, the proposed change here only papers over a deeper problem and doesn't really fix anything other than ensuring we don't dereference an `end()` iterator.  While mostly benign, I think I'd prefer that we fix the real problem instead of merging this.

blattms:
Maybe share what the real problem is?

bska:
> Maybe share what the real problem is?

Sure.  The real problem is that if there are dynamic changes to the a well's connections within an action block&ndash;e.g., if there are new connections introduced in `COMPDAT` or some connections open and/or some connections close through `WELOPEN`, then the data structure used in `updateWellsEcl()` is inconsistent and the invariants upon which the current logic depends do not hold.  Function `updateWellsEcl()` was created (PR #3116, commit 98f6a9a7a) to solve a **much** simpler problem than we're currently using it to solve&ndash;opening/closing a well in response to keywords like `WELOPEN` in an `ACTIONX` block.  The connection loop that's altered here was introduced as a copy of the `WELPI` handling that I added in PR #2866 (commit 319c24033).

The problem is that it's just not safe to run that loop at all if there are  new connections being added and/or connections being opened/closed.  A quick workaround that's probably safe in most cases is to completely remove that loop here and to make sure that we record a structural change in the [dynamic `WELPI` handler](https://github.com/OPM/opm-common/blob/12334dc673e68d09b94427e020632dfb7798b87d/opm/input/eclipse/Schedule/Well/WellPropertiesKeywordHandlers.cpp#L160), but the **real** solution will require reworking our machinery for imparting dynamic well or group changes from the `ACTIONX` handler to the simulator layer.  A list of "affected wells" is not sufficient anymore.

akva2:
superseeded by a better solution.

PR #5625, opened 2024-09-23
supporting ZMF explicit solution

GitPaean:
it needs https://github.com/OPM/opm-common/pull/4200 .

### DIFF
--- a/opm/simulators/flow/FlowProblemComp.hpp
+++ b/opm/simulators/flow/FlowProblemComp.hpp
@@ -350,54 +350,65 @@ protected:
         const auto& vanguard = simulator.vanguard();
         const auto& eclState = vanguard.eclState();
         const auto& fp = eclState.fieldProps();
-        // const bool has_xmf = fp.has_double("XMF");
-        assert(fp.has_double("XMF"));
-        // const bool has_ymf = fp.has_double("YMF");
-        assert(fp.has_double("YMF"));
-        const bool has_temp = fp.has_double("TEMPI");
         const bool has_pressure = fp.has_double("PRESSURE");
+        if (!has_pressure)
+            throw std::runtime_error("The ECL input file requires the presence of the PRESSURE "
+                                     "keyword if the model is initialized explicitly");
+
+        const bool has_xmf = fp.has_double("XMF");
+        const bool has_ymf = fp.has_double("YMF");
+        const bool has_zmf = fp.has_double("ZMF");
+        if (! (has_zmf || (has_xmf && has_ymf))) {
+            throw std::runtime_error("The ECL input file requires the presence of ZMF or XMF and YMF "
+                                     "keyword if the model is initialized explicitly");
+        }
+
+        if (has_zmf && (has_xmf || has_ymf)) {
+            throw std::runtime_error("The ECL input file can not handle explicit initialization "
+                                     "with both ZMF and XMF or YMF");
+        }
+
+        if (has_xmf != has_ymf) {
+            throw std::runtime_error("The ECL input file needs XMF and YMF combined to do the explicit "
+                                     "initializtion when using XMF or YMF");
+        }
+
+        const bool has_temp = fp.has_double("TEMPI");
 
         // const bool has_gas = fp.has_double("SGAS");
         assert(fp.has_double("SGAS"));
-        if (!has_pressure)
-            throw std::runtime_error("The ECL input file requires the presence of the PRESSURE "
-                                      "keyword if the model is initialized explicitly");
 
         std::size_t numDof = this->model().numGridDof();
 
         initialFluidStates_.resize(numDof);
 
-        std::vector<double> xmfData;
-        std::vector<double> ymfData;
         std::vector<double> waterSaturationData;
         std::vector<double> gasSaturationData;
         std::vector<double> soilData;
         std::vector<double> pressureData;
         std::vector<double> tempiData;
 
-        if (waterPhaseIdx > 0 && Indices::numPhases > 1)
+        const bool water_active = FluidSystem::phaseIsActive(waterPhaseIdx);
+        const bool gas_active = FluidSystem::phaseIsActive(gasPhaseIdx);
+        const bool oil_active = FluidSystem::phaseIsActive(oilPhaseIdx);
+
+        if (water_active && Indices::numPhases > 1)
             waterSaturationData = fp.get_double("SWAT");
         else
             waterSaturationData.resize(numDof);
 
         pressureData = fp.get_double("PRESSURE");
 
-        assert(waterPhaseIdx < 0);
-
-        xmfData = fp.get_double("XMF");
-        ymfData = fp.get_double("YMF");
         if (has_temp) {
             tempiData = fp.get_double("TEMPI");
         } else {
             ; // TODO: throw?
         }
 
-        if (gasPhaseIdx > 0) // && FluidSystem::phaseIsActive(oilPhaseIdx))
+        if (gas_active) // && FluidSystem::phaseIsActive(oilPhaseIdx))
             gasSaturationData = fp.get_double("SGAS");
         else
             gasSaturationData.resize(numDof);
-
-        // constexpr std::size_t num_comps = 3;
 
         for (std::size_t dofIdx = 0; dofIdx < numDof; ++dofIdx) {
             auto& dofFluidState = initialFluidStates_[dofIdx];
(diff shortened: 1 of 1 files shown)


GitPaean:
There will probably evolving update in the related code with new development. and I think it is ready for review for what the PR does.

GitPaean:
jenkins build this please

opm/simulators/flow/FlowProblemComp.hpp
```diff
@@ -438,13 +449,43 @@ class FlowProblemComp : public FlowProblem<TypeTag>
...
+                        }
+                    }
+                    if (oil_active) {
+                        dofFluidState.setMoleFraction(FluidSystem::oilPhaseIdx, compIdx, zmf);
+                        const auto& sat_oil = dofFluidState.saturation(oilPhaseIdx);
+                        if (sat_oil > 0.) {
+                            dofFluidState.setMoleFraction(FluidSystem::oilPhaseIdx, compIdx, zmf);
+                        } else {
+                            dofFluidState.setMoleFraction(FluidSystem::oilPhaseIdx, compIdx, 0.);
+                        }
```
bska, line 486: Does `setMoleFraction()` affect the return value of `saturation()`?  If not, the `sat_oil > 0` branch is redundant.  Maybe this could be rewritten as
```C++
dofFluidState.setMoleFraction(FluidSystem::oilPhaseIdx, compIdx, zmf);
if (! (dofFluidState.saturation(oilPhaseIdx) > 0)) {
    dofFluidState.setMoleFraction(FluidSystem::oilPhaseIdx, compIdx, Scalar{0});
}
```
or maybe even
```C++
const auto xmf = (dofFluidState.saturation(oilPhaseIdx) > 0) ? zmf : Scalar{0};
dofFluidState.setMoleFraction(FluidSystem::oilPhaseIdx, compIdx, xmf);
```
By the way, this block uses both the name `FluidSystem::oilPhaseIdx` and the unadorned name `oilPhaseIdx`.  Do those two the represent the same index value?

opm/simulators/flow/FlowProblemComp.hpp
```diff
@@ -438,13 +449,43 @@ class FlowProblemComp : public FlowProblem<TypeTag>
...
+                    const std::size_t data_idx = compIdx * numDof + dofIdx;
+                    const Scalar zmf = zmfData[data_idx];
+                    // TODO: I know we do this for co2_ptflash example, but maybe we should not do it this way?
+                    if (gas_active) {
+                        const auto& sat_gas = dofFluidState.saturation(gasPhaseIdx);
+                        if (sat_gas > 0.) {
+                            dofFluidState.setMoleFraction(FluidSystem::gasPhaseIdx, compIdx, zmf);
+                        } else {
+                            dofFluidState.setMoleFraction(FluidSystem::gasPhaseIdx, compIdx, 0.);
+                        }
```
bska, line 477: An alternative way to write this is as
```C++
const auto ymf = (dofFluidState.saturation(gasPhaseIdx) > 0) ? zmf : Scalar{0};
dofFluidState.setMoleFraction(FluidSystem::gasPhaseIdx, compIdx, ymf);
```
which may or may not be easier to read.

opm/simulators/flow/FlowProblemComp.hpp
```diff
@@ -438,13 +449,43 @@ class FlowProblemComp : public FlowProblem<TypeTag>
...
+                    const Scalar ymf = ymfData[data_idx];
+
+                    dofFluidState.setMoleFraction(FluidSystem::oilPhaseIdx, compIdx, xmf);
+                    dofFluidState.setMoleFraction(FluidSystem::gasPhaseIdx, compIdx, ymf);
+                }
+            }
 
-                dofFluidState.setMoleFraction(FluidSystem::oilPhaseIdx, compIdx, xmf);
-                dofFluidState.setMoleFraction(FluidSystem::gasPhaseIdx, compIdx, ymf);
+            if (has_zmf) {
```
bska, line 465: This block doesn't do anything unless at least one of `gas_active` or `oil_active` is `true`.  Could there ever be a case where both of those are `false` at the same time?

bska (commented):
Apart from a couple of smaller issues this looks okay to me.  That said, if I were writing this, I would probably split the meat of the function into two separate helper functions, one for the ZMF case and one for the XMF+YMF case.  Then I'd probably reverse the loop order too with the component index as the outer loop, i.e. for the ZMF case:
```C++
const auto& zmfData = fp.get_double("ZMF");
auto zIx = 0*zmfData.size();
for (auto compIdx = 0*numComponents; compIdx < numComponents; ++compIdx) {
    for (auto dofIdx = 0*numDof; dofIdx < numDof; ++dofIdx, ++zIx) {
        auto& fs = this->initialFluidStates_[dofIdx];
        if (gas_active) {
            const auto ymf = (fs.saturation(gasPhaseIdx) > 0) ? zmfData[zIx] : Scalar{0};
            fs.setMoleFraction(gasPhaseIdx, compIdx, ymf);
        }
        if (oil_active) {
            const auto xmf = (fs.saturation(oilPhaseIdx) > 0) ? zmfData[zIx] : Scalar{0};
            fs.setMoleFraction(oilPhaseIdx, compIdx, xmf);
        }
    }
}
```

opm/simulators/flow/FlowProblemComp.hpp
```diff
@@ -350,55 +350,66 @@ class FlowProblemComp : public FlowProblem<TypeTag>
...
-        const bool has_temp = fp.has_double("TEMPI");
         const bool has_pressure = fp.has_double("PRESSURE");
+        if (!has_pressure)
+            throw std::runtime_error("The ECL input file requires the presence of the PRESSURE "
+                                     "keyword if the model is initialized explicitly");
+
+        const bool has_xmf = fp.has_double("XMF");
+        const bool has_ymf = fp.has_double("YMF");
+        const bool has_zmf = fp.has_double("ZMF");
+        if (! (has_zmf || (has_xmf && has_ymf))) {
```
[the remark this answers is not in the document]
GitPaean: To me, they read equivalent. I will update with your suggestion.

opm/simulators/flow/FlowProblemComp.hpp
GitPaean: > Does setMoleFraction() affect the return value of saturation()? If not, the sat_oil > 0 branch is redundant. Maybe this could be rewritten as

No, they are different things. Here, I want to set mole fraction is 0 for the phase does not exist, for example, the cell does not have oil, I will set the mole fraction in oil 0.  So I should have removed the following line, 
```c++
                        dofFluidState.setMoleFraction(FluidSystem::oilPhaseIdx, compIdx, zmf);
```

GitPaean: > 
```
const auto xmf = (dofFluidState.saturation(oilPhaseIdx) > 0) ? zmf : Scalar{0};
dofFluidState.setMoleFraction(FluidSystem::oilPhaseIdx, compIdx, xmf);
```

I will use this one. 

`oilPhaseIdx` is the same with `FluidSystem::oilPhaseIdx`,  defined in FlowProblem.hpp with ` enum { oilPhaseIdx = FluidSystem::oilPhaseIdx };`

opm/simulators/flow/FlowProblemComp.hpp
GitPaean: done.

opm/simulators/flow/FlowProblemComp.hpp
GitPaean: Although I would think it will be very strange, but I do not know whether it is allowed at all. I would add more checks inside the if condition.

GitPaean:
> Apart from a couple of smaller issues this looks okay to me. That said, if I were writing this, I would probably split the meat of the function into two separate helper functions, one for the ZMF case and one for the XMF+YMF case. Then I'd probably reverse the loop order too with the component index as the outer loop, i.e. for the ZMF case:
> 
> ```c++
> const auto& zmfData = fp.get_double("ZMF");
> auto zIx = 0*zmfData.size();
> for (auto compIdx = 0*numComponents; compIdx < numComponents; ++compIdx) {
>     for (auto dofIdx = 0*numDof; dofIdx < numDof; ++dofIdx, ++zIx) {
>         auto& fs = this->initialFluidStates_[dofIdx];
>         if (gas_active) {
>             const auto ymf = (fs.saturation(gasPhaseIdx) > 0) ? zmfData[zIx] : Scalar{0};
>             fs.setMoleFraction(gasPhaseIdx, compIdx, ymf);
>         }
>         if (oil_active) {
>             const auto xmf = (fs.saturation(oilPhaseIdx) > 0) ? zmfData[zIx] : Scalar{0};
>             fs.setMoleFraction(oilPhaseIdx, compIdx, xmf);
>         }
>     }
> }
> ```

Thanks for the suggestion. Please suggest whether you insist this way or you think the current way is okay.  I am okay with either way, but I prefer less work if it is okay.

GitPaean:
Thanks for the comments. I think I have addressed them.

GitPaean:
jenkins build this please

bska (approved):

> Please suggest whether you insist [on separate helpers for the XMF+YMF vs. ZMF cases] or you think the current way is okay.

I don't insist on anything since I'm not involved enough with this work.

> I am okay with either way, but I prefer less work if it is okay.

Fine with me, although there's often a bit of tension between "less work right now" an "less work in the future".

GitPaean:
Thanks for the merging. The knowledge about the compositional simulation is still evolving, hopefully, the implementation form will improve along the way.

PR #5628, opened 2024-09-23
[bugfix] Always broadcast NNCs from input.

blattms:
This was only done if there were NNCs from aquifers, but it should always be done whenever there are NNCs.

Those NNCs also hold information of (additional) transmissibilities for a grid intersection. While the intersection is already represented by the grid, the transmissibility is not. This PR fixes this and makes the NNcs available on all processes.

If I missed where we broadcast NNCs, then please tell me.

### DIFF
--- a/opm/simulators/flow/GenericCpGridVanguard.cpp
+++ b/opm/simulators/flow/GenericCpGridVanguard.cpp
@@ -433,19 +433,26 @@ void GenericCpGridVanguard<ElementMapper,GridView,Scalar>::doCreateGrids_(Eclips
 
 #if HAVE_MPI
     {
-        const bool has_numerical_aquifer = eclState.aquifer().hasNumericalAquifer();
         int mpiSize = 1;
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

blattms:
We must be doing this somewhere else already. Just need to find it.

akva2:
These are already broadcast during the EclipseState broadcast, ie, https://github.com/OPM/opm-common/blob/master/opm/input/eclipse/EclipseState/EclipseState.hpp#L151

blattms:
And that seems to happen in readDeck.cpp, which is really early. Why might just want to move that to a later stage. It seem that we don't need it before we have a distributed grid.

atgeirr:
A somewhat common way to do arbitrary non-cornerpoint grids in OPM Flow is to make a disjointed grid with all cells in a row, then adding every connection as an explicit NNC. Will this work make that scale worse in parallel than before?

PR #5629, opened 2024-09-25
fixed: properly disable SFC ordering in alugrid

akva2:
This define seem to have fell out at some point.


akva2:

akva2:
note the build will still fail because reference data is sfc. i just want files to do manual comparison.

akva2:
okay, this define is only read when building alugrid.

PR #5630, opened 2024-09-25
add changes to fix the order

ElyesAhmed:
This PR enables SFC (Space-Filling Curve) reordering in AluGrid as the default choice. The changes has been tested both with and without MPI across multiple cases from the opm-tests and opm-data!

### DIFF
--- a/flow/flow_blackoil_alugrid.cpp
+++ b/flow/flow_blackoil_alugrid.cpp
@@ -17,9 +17,15 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
 #include <config.h>
-// The default in dune-ALUGrird is "DISABLE_ALUGRID_SFC_ORDERING is false"
-#undef DISABLE_ALUGRID_SFC_ORDERING
-#undef USE_ALUGRID_SFC_ORDERING
+// By default, dune-ALUGrid uses Space-Filling Curve (SFC) ordering
+// for cells to optimize data access patterns for adaptive mesh refinements/coarsening.
+// However, if you want to use Cartesian ordering, as is used in OPM, you
+// can switch to it by defining the macro #define DISABLE_SFC_ORDERING 1.
+// This will change the default cell order to Cartesian.
+// Note that this option is not available for pre-built or installed versions of dune-ALUGrid.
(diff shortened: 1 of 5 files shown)


akva2:
Note that flipping these defines in opm code does nothing. You are entirely at the mercy of however dune-alugrid itself was built as the defines are not in headers but in compile units.

akva2:
No there is no indication of this. It's just an #ifdef in its gridfactory.cc that exits early. What makes it even more fun is that gridfactory.cc *can* be built inline (ie the .cc file is included in the .hh file) if you build dune-alugrid with COMPILE_ALUGRID_INLINE define. But you can't just set this for a library that was built without it as you'd get duplicate symbols. So as I said, at the mercy of how the library is built.

atgeirr:

atgeirr:
jenkins build this serial failure_report please

ElyesAhmed:
can we try again!

akva2:
jenkins build this serial failure_report please

akva2:
jenkins build this serial opm-tests=1234 please

atgeirr:
jenkins build this serial failure_report update_data please

atgeirr:
jenkins build this serial opm-tests=1234 please

atgeirr:
Data update merged, will merge this once the data update Jenkins job has started.

bska:
> will merge this once the data update Jenkins job has started.

The new reference solutions were installed 30 minutes ago.  Did you accidentally forget to merge this?

atgeirr:
> Did you accidentally forget to merge this?

Yes, had to go to a meeting, sorry!

blattms:
Just a note. According to my failed tests (because of missing rebase). The changes to the well solutions are quite large: https://ci.opm-project.org/job/opm-common-PR-builder/7522/testReport/(root)/mpi/compareECLFiles_flow_blackoil_alugrid_SPE1CASE2/

atgeirr:
> The changes to the well solutions are quite large:

We are aware of this, it is because the previous solution was wrong. The bugfixes in this PR apparently had a big influence.

blattms:
Well, our code should not rely on internal ordering. If it does then we are doomed (not only for ALUGrid...).

Maybe somebody could describe what this is supposed to fix.

PR #5631, opened 2024-09-25
Add dampening to the injectivity multipliers (WINJMULT)

vkip:


vkip:

vkip:
Test failures: tiny differences on the WINJMULT test cases, not visible on the plots.

GitPaean:

GitPaean:

GitPaean:
It looks looks relocating the `param_` causes some compilation problem which can mostly be fixed by using `this->param_`. 

@vkip , can you have a look and fix the compilation?

vkip:
> It looks looks relocating the `param_` causes some compilation problem which can mostly be fixed by using `this->param_`.
> 
> @vkip , can you have a look and fix the compilation?

Sure, should be ok now.

GitPaean:

GitPaean (changes requested):
The PR generally looks fine.  The complication of this PR is that the well solving iteration code is used for very different purposes.  And `inj_multiplier_damp_factor_` is kept when transferring from one process to the other one, which is difficult to evaluate how it affects the simulation. 

One thing I observed that when we calculate the well potentials based on BHP limit. If a small damping factor is obtained, and if it is used in the well potential calculation based on BHP limit, it might cause more iterations. Ideally, when the well bhp is set, the inj_multiplier should also be set already, the damping might not help much. 

For that, I suggest that in the computeWellRatesWithBhpIterations() function, when we make a copy for the well with `MultisegmentWell<TypeTag> well_copy(*this);
` or `StandardWell<TypeTag> well_copy(*this);`, we reset the damping factor to be 1. 

For other parts, let us evaluate based on testing.

GitPaean:
A little more details about the observation, for the testing case we have, the inj_multiplier is about 26 for the normal simulation, and damping factor is on the 0.13 level. 

When it enters well potential calculation based on BHP limit. With BHP set, inj_muliplier is basically based on bhp or perforation pressure, the inj_multiplier is almost know, in that case, it can reach 13XXX level. With entering with inj_multiplier 26 and 0.13 level damping factor, it takes less then 10 iterations to reach the desired 13xxx level inj_multiplier.  That is what the mentioned observation is about. 

For most of cases, it might NOT be a problem. The case we have has a very big scaling factor or the injectivity multiplier at 100 level.

vkip:
> A little more details about the observation, for the testing case we have, the inj_multiplier is about 26 for the normal simulation, and damping factor is on the 0.13 level.
> 
> When it enters well potential calculation based on BHP limit. With BHP set, inj_muliplier is basically based on bhp or perforation pressure, the inj_multiplier is almost know, in that case, it can reach 13XXX level. With entering with inj_multiplier 26 and 0.13 level damping factor, it takes less then 10 iterations to reach the desired 13xxx level inj_multiplier. That is what the mentioned observation is about.
> 
> For most of cases, it might NOT be a problem. The case we have has a very big scaling factor or the injectivity multiplier at 100 level.

Thanks! It is a good point, and your suggestion does indeed help for the potential calculations. Pushing an update now.

GitPaean (approved):
Thanks for the update. It looks good to me now.  Let me prepare to get it in.

GitPaean:
jenkins build this update_data failure_report please

GitPaean:
jenkins build this opm-tests=1241 please

GitPaean:
The reference is updated. I am merging the PR in now.

PR #5633, opened 2024-09-26
Extend CartesianIndexMapper due to speciatilization of CpGrid

aritorto:
Only relevant after OPM/opm-grid#762 is reviewed/merged.


aritorto:
Replace/improve by OPM/opm-simulators#5649

PR #5635, opened 2024-09-27
changed: don't store schedule keywords unless required in simulators

akva2:
in the python simulator we cannot make this decision up front since the schedule can be modified from python code so we have to do the safe thing

Downstream of https://github.com/OPM/opm-common/pull/4234


bska (approved):

PR #5636, opened 2024-09-27
Gpu support co2gaspvt

multitalentloes:
Adds test verifying that new classes work on the GPU, classes that are now tested include:
The test depends on new functionality in https://github.com/OPM/opm-common/pull/4222

- Constants
- IdealGas
- Brine_CO2
- UniformTabulated2DFunction
- BrineDynamic
- CO2
- SimpleHuDuanH20
- Co2GasPvt

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -457,12 +457,14 @@ if (HAVE_CUDA)
   ADD_CUDA_OR_HIP_FILE(TEST_SOURCE_FILES tests test_solver_adapter.cpp)
   ADD_CUDA_OR_HIP_FILE(TEST_SOURCE_FILES tests test_gpu_ad.cu)
   ADD_CUDA_OR_HIP_FILE(TEST_SOURCE_FILES tests test_gpu_linear_two_phase_material.cu)
+  ADD_CUDA_OR_HIP_FILE(TEST_SOURCE_FILES tests test_gpuPvt.cu)
 
   # for loop providing the flag --expt-relaxed-constexpr to fix some cuda issues with constexpr
   if(NOT CONVERT_CUDA_TO_HIP)
     set(CU_FILES_NEEDING_RELAXED_CONSTEXPR
       tests/gpuistl/test_gpu_ad.cu
       tests/gpuistl/test_gpu_linear_two_phase_material.cu
+      tests/gpuistl/test_gpuPvt.cu
     )
 
     foreach(file ${CU_FILES_NEEDING_RELAXED_CONSTEXPR})
--- a/opm/simulators/linalg/gpuistl/GpuBuffer.cpp
+++ b/opm/simulators/linalg/gpuistl/GpuBuffer.cpp
@@ -84,22 +84,29 @@ GpuBuffer<T>::resize(size_t newSize)
     if (newSize < 1) {
         OPM_THROW(std::invalid_argument, "Setting a GpuBuffer size to a non-positive number is not allowed");
     }
-    // Allocate memory for the new buffer
-    T* tmpBuffer = nullptr;
-    OPM_GPU_SAFE_CALL(cudaMalloc(&tmpBuffer, sizeof(T) * newSize));
 
-    // Move the data from the old to the new buffer with truncation
-    size_t sizeOfMove = std::min({m_numberOfElements, newSize});
-    OPM_GPU_SAFE_CALL(cudaMemcpy(tmpBuffer,
-                                  m_dataOnDevice,
-                                  sizeOfMove * sizeof(T),
-                                  cudaMemcpyDeviceToDevice));
+    if (m_numberOfElements == 0) {
+        // We have no data, so we can just allocate new memory
+        OPM_GPU_SAFE_CALL(cudaMalloc(&m_dataOnDevice, sizeof(T) * newSize));
+    }
+    else {
+        // Allocate memory for temporary buffer
+        T* tmpBuffer = nullptr;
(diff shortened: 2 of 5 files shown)


opm/simulators/linalg/gpuistl/GpuBuffer.cpp
```diff
@@ -33,6 +33,14 @@ GpuBuffer<T>::GpuBuffer(const std::vector<T>& data)
 {
 }
 
+// some classes need a default constructor...
+// should we reintroduce the default constructor that allocates 0 bytes and initializes the handles?
```
atgeirr, line 37: There is already a declared default constructor, but it seems it is not actually implemented before this, which is weird.

That aside, I think it should not allocate anything. Default construct then resize should allocate only once. I suggest removing this implementation, defaulting `m_numberOfElements` to zero in the header file, and then just defaulting the constructor in the header:
```
    /**
     * @brief Default constructor that will initialize cublas and allocate 0 bytes of memory
     */
    GpuBuffer() = default;
```
That way you do not need to change the doc string either to say 1 instead 0...

Finally, note that it is currently declared explicit in the header, which is unnecessary and misleading.

CMakeLists_files.cmake
```diff
@@ -457,12 +457,14 @@ if (HAVE_CUDA)
...
   ADD_CUDA_OR_HIP_FILE(TEST_SOURCE_FILES tests test_gpu_ad.cu)
   ADD_CUDA_OR_HIP_FILE(TEST_SOURCE_FILES tests test_gpu_linear_two_phase_material.cu)
+  ADD_CUDA_OR_HIP_FILE(TEST_SOURCE_FILES tests test_gpuPvt.cu)
 
   # for loop providing the flag --expt-relaxed-constexpr to fix some cuda issues with constexpr
   if(NOT CONVERT_CUDA_TO_HIP)
     set(CU_FILES_NEEDING_RELAXED_CONSTEXPR
       tests/gpuistl/test_gpu_ad.cu
       tests/gpuistl/test_gpu_linear_two_phase_material.cu
+      tests/gpuistl/test_co2gaspvt.cu
```
atgeirr, line 467: Did you mean test_gpuPvt.cu?

tests/gpuistl/test_gpuPvt.cu
```diff
@@ -0,0 +1,271 @@
...
+
+using Evaluation = Opm::DenseAd::Evaluation<float, 3>;
+using GpuB = const Opm::gpuistl::GpuBuffer<double>;
+using GpuV = Opm::gpuistl::GpuView<const double>;
+
+using GpuTab = Opm::UniformTabulated2DFunction<double, GpuV>;
+
+using GpuBufCo2Tables = Opm::CO2Tables<double, GpuB>;
+using GpuViewCO2Tables = Opm::CO2Tables<double, GpuV>;
+template Opm::CO2Tables<double, GpuB>::CO2Tables(const Opm::UniformTabulated2DFunction<double, GpuB>&, const Opm::UniformTabulated2DFunction<double, GpuB>&);
```
atgeirr, line 35: What do these explicit instantiations do? We do not access this in a scope linking to this file.

tests/gpuistl/test_gpuPvt.cu
```diff
@@ -0,0 +1,271 @@
...
+#include <opm/simulators/linalg/gpuistl/detail/gpu_safe_call.hpp>
+#include <opm/simulators/linalg/gpuistl/GpuBuffer.hpp>
+#include <opm/simulators/linalg/gpuistl/GpuView.hpp>
+
+#include <cuda_runtime.h>
+#include <vector>
+#include <utility>
+#include <cmath>
+
+using Evaluation = Opm::DenseAd::Evaluation<float, 3>;
```
atgeirr, line 27: Is float used on purpose here? Everything else is doubles.

tests/gpuistl/test_gpuPvt.cu
```diff
@@ -0,0 +1,271 @@
...
+
+    // Launch the kernel
+    kernel<<<1, 1>>>(args..., resultOnGpu);
+
+    // Check for any errors in kernel launch
+    OPM_GPU_SAFE_CALL(cudaPeekAtLastError());
+    OPM_GPU_SAFE_CALL(cudaDeviceSynchronize());
+
+    // Retrieve the result from the GPU to the CPU
+    OPM_GPU_SAFE_CALL(cudaMemcpy(&gpuComputedResultOnCpu, resultOnGpu, sizeof(double), cudaMemcpyDeviceToHost));
```
atgeirr, line 132: What would `gpuComputedResultOnCpu = *resultOnGpu;` do?

atgeirr (commented):
Only a few minor issues to be resolved. Should be merged when fixed.

tests/gpuistl/test_gpuPvt.cu
multitalentloes: That would not yield anything meaningful, you cannot dereference a gpu pointer and get the value in CPU code

CMakeLists_files.cmake
multitalentloes: Good catch, I decided to change the name to `test_gpuPvt.cu` at some point

multitalentloes:
jenkins build this please

opm/simulators/linalg/gpuistl/GpuBuffer.cpp
multitalentloes: fixed now with updated doc string and logic in resize function accounting for original size being 0

atgeirr (approved):

PR #5640, opened 2024-10-02
Fix Lower Bound Conditions for Three Point Checks

bska:
These are supposed to verify that the displacing saturation is strictly between the critical and the maximum saturation values.

Introduced in #5573, found while adding unit tests to #5592.


bska:
jenkins build this please

bska:
jenkins build this please

akva2 (approved):
Consistent with comments.

bska:

PR #5641, opened 2024-10-02
Fix oscillations - continuation of #5607

steink:
Continuing #5607 here due to vacation.
Updates:

- Increased default `maximum-number-of-well-switches` from 3 to 10 (see #5607)
- Removed suggested update of control-log during local iterations. Keeping track of switches during local iterations is not important wrt global convergence + the suggested update would also add an entry for every status switch (which is not desirable).


steink:

PR #5644, opened 2024-10-03
Adjust packaging for pybind bump

akva2:
not supported by newer pybind11 and long out of support in any case

Companion to https://github.com/OPM/opm-common/pull/4247

### DIFF
--- a/python/generate-pypi-package.sh
+++ b/python/generate-pypi-package.sh
@@ -19,7 +19,7 @@ python_versions[cp312-cp312]=/opt/python/cp312-cp312/bin/python
 for python_bin in ${python_versions[*]}
 do
   ${python_bin} -m pip install pip --upgrade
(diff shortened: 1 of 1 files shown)


alfbr:
I am afraid we are still using 3.6 on Red Hat 7. I do believe we will phase out Red Hat 7 very soon, and then be on 3.8. Any chance we can postpone this a month or so?

akva2:
I asked @vkip and he said you were on 3.8.

It's really unfortunate if we do not get this into the release and that will mean broken python bindings for everyone on python > 3.8 for at least another 6 months.

alfbr:
He is ahead of the curve, on Red Hat 7 we are still on 3.6.

akva2:
i have increased the complexity of this to keep you afloat. while in theory some trouble maker may have built numpy2 for python < 3.8 i think the chance of finding this animal in the wild is small.

PR #5645, opened 2024-10-03
Support for 2 to 7 components in flowexp_comp

kjetilly:
This is a rather rudimentary way of supporting from 2 to 7 components in the flowexp_comp executable. I am marking this as draft primarily because I'm unsure if the current version, which first reads the deck file before instantiating, is the best course of action since it could involve reading the deck twice (?).


akva2:
that's exactly what the main flow binary does (reading the deck twice).

atgeirr:
> adding the ability to only read a single section of the deck would benefit all simulators.

I thought we already got rid of the need to read it twice! If false, this seems like a good thing to do.

bska:
> that's exactly what the main flow binary does (reading the deck twice).

No it doesn't.  The ["EarlyBird"](https://github.com/OPM/opm-simulators/blob/84e6ed64930095ab31eef4699bdb0b777a9a0f84/opm/simulators/flow/Main.hpp#L154) type tag [loads](https://github.com/OPM/opm-simulators/blob/84e6ed64930095ab31eef4699bdb0b777a9a0f84/opm/simulators/flow/Main.hpp#L400) the input deck and forms the `EclipseState` and `Schedule` (and other) objects, and then forwards those to the "real" simulator in a [call](https://github.com/OPM/opm-simulators/blob/84e6ed64930095ab31eef4699bdb0b777a9a0f84/opm/simulators/flow/Main.hpp#L191) to [`Main::setupVanguard()`](https://github.com/OPM/opm-simulators/blob/84e6ed64930095ab31eef4699bdb0b777a9a0f84/opm/simulators/flow/Main.cpp#L250)

akva2:
yeah, i realized. it's just the parameters that get double parsed now.

bska:
> adding the ability to only read a single section of the deck

We [have that](https://github.com/OPM/opm-common/blob/8e45fb7a1b13824959475bea70ca5674b3f97344/opm/input/eclipse/Parser/Parser.hpp#L75-L78).  A call of the form
```C++
const auto deck = Parser{}.parseFile("CASE.DATA", ParseContext{}, std::vector { Parser::SectionType::RUNSPEC });
```
would load just the `RUNSPEC` section of `CASE.DATA`.  Of course this isn't as "nicely" packaged as our `readDeck()` function, but if all you need is the ability to load `RUNSPEC`, then it's available in the box.

GitPaean:
For this PR, `numCompos()` from `Runspec` can also be used to determine the number of the components.

kjetilly:
Thanks for the feedback! 

I implemented @bska suggestion on reading only the `RUNSPEC` section, and cleaned up quite a bit.

Removing draft status.

GitPaean:
jenkins build this please

akva2:
i am a little concerned with instancing 6 simulators in one translation unit. while not as elegant code wise, i think it's better to do some variant of https://github.com/akva2/opm-simulators/tree/one_tu_per_components to keep memory usage acceptable. compiling this now requires 3.7GB and does not memory usage is less than 2GB and build time is still lower assuming tu's can build in parallel.

kjetilly:
@akva2 I absolutely agree that splitting over 6 translation units would be better. I see that your branch essentially implements this PR (except it uses the old version of deck reading). 

How should we go about this git wise? Should I just do the minimal change of splitting up into multiple TUs and keep this PR?

akva2:
Just pick my commit, adjust and amend this pr. I am on vacation until tuesday

kjetilly:
Jenkins build this please

akva2 (approved):

bska:
PR approved and build check is green.  I'll merge into the master branch.

GitPaean:
Hei, @kjetilly, can you go a little further so that people can compile only one of compositional variants when developing? For example, I am developing based on a testing case with four components, and I do not want to compile the variants with 2, 3, 5, 6, 7 all the time.   It is okay those targets only visible when a cmake FLAG is set. 

Similar to the black oil way, more than 90% percent of time I only compile `flow_blackoil`.

flowexperimental/comp/flowexp_comp.cpp
```diff
@@ -17,292 +17,48 @@
...
+        = Opm::FlowGenericVanguard::canonicalDeckPath(Opm::Parameters::Get<Opm::Parameters::EclDeckFileName>());
+
+    // Only read the RUNSPEC section of the deck
+    const auto deck = Opm::Parser{}.parseFile(inputFilename,
+                                              Opm::ParseContext{},
+                                              std::vector { Opm::Ecl::SectionType::RUNSPEC });
+    const auto runspec = Opm::Runspec(deck);
+    const auto numComps = runspec.numComps();
+
+    OPM_ERROR_IF(numComps < 2 || numComps < 7,
```
GitPaean, line 51: should be `numComps > 7`.

flowexperimental/comp/flowexp_comp.cpp
```diff
@@ -17,292 +17,48 @@
...
+                 "Deck has {} components, not supported. We support a maximum of 7 components, "
+                 "and a minimum of 2.");
```
GitPaean, line 53: The `numComps` is not in the outputting message.  I believe it is supposed to use fmt here.

GitPaean:
opened a pr to fix these small fallouts, https://github.com/OPM/opm-simulators/pull/5650

flowexperimental/comp/flowexp_comp.cpp
kjetilly: Hmm, I think something went wrong when I cherry picked the splitting commit, sorry!

kjetilly:
I'm working on having the number of components decideable by cmake, should be done by today/evening.

GitPaean:
> @GitPaean I'm working on having the number of components decideable by cmake, should be done by today/evening.

Great! I opened the PR because I have it in my opened file.

PR #5646, opened 2024-10-03
Actually use fixed ordering (to get reproducible results in parallel)

vkip:


vkip:
jenkins build this please

blattms:
Great catch. Thanks.

atgeirr:
I had to think about why this did not bite me: I have been using a patched older dune, with the old version (not runtime configurable) of the fix, so I never noticed this did not function.

PR #5648, opened 2024-10-06
Enable EQUALREG for Array Operations

bska:
The keyword's implementation has reached a point where we can justify its addition to the simulator's feature set.


bska:
jenkins build this please

blattms (approved):
Looks great.

Ok. Then my recollection was correct, that this should work code-wise.

Will this pick up and use my changes of #4238 or do I need to do more?

bska:
> Will this pick up and use my changes of #4238 or do I need to do more?

Do you mean OPM/opm-common#4238?  If so, then yes this PR will enable all of that.  The practical outcome of this PR is that we no longer reject `EQUALREG` as "unsupported" at the simulator layer.  Any `EQUALREG` features managed by the input layer will be correctly conferred onto the simulator layer now that it's enabled.

PR #5651, opened 2024-10-08
Compile only a select number of components.

kjetilly:
I am not sure if this is overkill, and I'll let the reviewers judge, but this PR enables one to select the components to compile through cmake.

To set in cmake, do 

```bash
cmake ... -DOPM_COMPILE_COMPONENTS="2;4;7"
```

this will then only compile for 2, 4, and 7 components. 

Though not supported in the current version, one could extend the support to an arbitrary number of components through some cmake magic to generate new compilation units, but I'll leave that for now.


kjetilly:
Jenkins build this please

kjetilly:
Note that I haven't tried breaking this by giving in weird cmake arguments to `OPM_COMPILE_COMPONENTS`, and while I taken some steps to make it robust, it can probably fail in some unforeseen ways if the user really makes an effort.

kjetilly:
Jenkins build this please

flowexperimental/comp/flowexp_comp.cpp
```diff
@@ -19,18 +19,58 @@
...
 
-int main(int argc, char** argv)
+template <int compileTimeComponent>
+std::tuple<bool, int>
+runComponent(int runtimeComponent, int argc, char** argv)
+{
+    if (runtimeComponent == compileTimeComponent) {
+        return std::make_tuple(true, Opm::dispatchFlowExpComp<compileTimeComponent>(argc, argv));
+    }
+    return std::make_tuple(false, EXIT_FAILURE);
```
bska, line 42: Tiny nit: You need to `#include <cstdlib>` to bring a definition of the `EXIT_FAILURE` macro into scope.

flowexperimental/comp/flowexp_comp.cpp
```diff
@@ -44,23 +84,20 @@ int main(int argc, char** argv)
...
-                             "and a minimum of 2.", numComps) );
-
-    switch (numComps) {
-    case 2: return Opm::dispatchFlowExpComp<2>(argc, argv);
-    case 3: return Opm::dispatchFlowExpComp<3>(argc, argv);
-    case 4: return Opm::dispatchFlowExpComp<4>(argc, argv);
-    case 5: return Opm::dispatchFlowExpComp<5>(argc, argv);
-    case 6: return Opm::dispatchFlowExpComp<6>(argc, argv);
-    case 7: return Opm::dispatchFlowExpComp<7>(argc, argv);
+    auto [componentSupported, executionSatus]
```
akva2, line 92: Status

akva2 (approved):
I can see simpler ways to implement these but you seem dead set on using recursive templates so okay ;)

GitPaean:
> I can see simpler ways to implement this but you seem dead set on using recursive templates so okay ;)

If it is not too much to do, you can always suggest before it gets updated and goes in.

kjetilly:
Jenksin build this please

kjetilly:
> I can see simpler ways to implement this but you seem dead set on using recursive templates so okay ;)

I honestly did not see any simpler ways, but maybe the saying about having a hammer and only seeing nails  rings true...

One thing I wanted to avoid was too much massaging of code through CMake, but granted, that is a personal preference, but I am very open to suggestions on how this could have been done simpler.

akva2:
yeah, codegen from cmake side was my idea, see https://github.com/akva2/opm-simulators/commits/only_one_component/

choose yer poison.

atgeirr:
I think I prefer the recursive template to cmake codegen, for my sake. Interesting alternative though!

kjetilly:
Jenkins build this please

GitPaean:
Is there a fmt version issue? I could not compile. 

```
error: ‘println’ is not a member of ‘fmt’
```

kjetilly:
Jenkins build this please

kjetilly:
> Is there a fmt version issue? I could not compile.
> 
> ```
> error: ‘println’ is not a member of ‘fmt’
> ```
> 
> I can use `print`.

I think this is available in {fmt} v10, but I switched to `print`

flowexperimental/comp/flowexp_comp.cpp
```diff
@@ -44,23 +86,20 @@ int main(int argc, char** argv)
...
-    case 5: return Opm::dispatchFlowExpComp<5>(argc, argv);
-    case 6: return Opm::dispatchFlowExpComp<6>(argc, argv);
-    case 7: return Opm::dispatchFlowExpComp<7>(argc, argv);
+    auto [componentSupported, executionStatus]
+        = runComponent<OPM_COMPILE_COMPONENTS_TEMPLATE_LIST>(numComps, argc, argv);
+
+    if (!componentSupported) {
+        fmt::print("Deck has {} components, not supported. In this build of the simulator, we support the "
+                     "following number of components:\n\n\t{}.\n\n Note that the supported components can be changed "
+                     "when configuring CMake through the OPM_COMPILE_COMPONENTS options. Exiting.\n",
```
GitPaean, line 100: options-> option?

maybe it is helpful to give an example to show the full grammar here for reminding? 
```
-DOPM_COMPILE_COMPONENTS="2;4;7"
```

kjetilly:
Jenkins build this please

GitPaean:
It looks good now. Thanks for the update.

PR #5626, opened 2024-09-23
Support RPTSCHED WELLS=N

vkip:

PR #5653, opened 2024-10-09
Initialise Fluid-in-Place Balance Sheets at Restart

bska:
Commit 6ba9dc086 (PR #5626) moved the balance sheet report output logic ahead of the timestep.  As a consequence, the output module's `inplace_` and `initialInplace_` data members must be fully initialised the first time we call `writeReports()` if the run requests balance sheet output.

This PR initialises those objects by calling `calc_inplace()` from `EclWriter<>::endRestart()`.  Doing so restores the behaviour from before PR #5626.  Note, however, that using this value for `initialInplace_` is incorrect as the simulator then, in a restarted run, will calculate recovery factors based on the fluid distribution at the restart time instead of at the start of the base run. Nevertheless this is, as alluded to earlier, how the simulator has always performed.  The behaviour is just a little more explicit now.


bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

atgeirr:
Thanks, merging.

PR #5655, opened 2024-10-09
Bug fix in rocsparsesolver

razvnane:


multitalentloes:
jenkins build this serial rocm please

blattms:
Would you please make the actual change more apparent. This is rather a bugfix than a  typo fix. 
If I read "this fixes a typo" my expectation is that results stay the same, but maybe the text in the output changes.

atgeirr:
Good to call the right precision method! With the more precise title and all green, it should go in. Merging.

PR #5656, opened 2024-10-09
Simplify Unit Tests for LogOutputHelper

bska:
In particular, put more of the common setup code into the test fixture and use more powerful constructors for the `EclipseState` object.  While here, also switch to using the `update_well_var()` and `update_group_var()` member functions on the `SummaryState` object where appropriate since the resulting statements become slightly easier to understand in isolation.


bska:
jenkins build this please

blattms (approved):
Looks good. Thanks for rducing the code duplication

PR #5658, opened 2024-10-10
Delay stopping wells after local well solves until global update agrees

steink:
As we've looked into several remedies for getting rid of annoying REVIVED/STOPPED oscillations (e.g., for NORNE_ATW2013), I think perhaps this one is the most promising and/or less intrusive. Basically, it will just re-open a well that was stopped during local well solve unless the well entered the well solve with a zero rate (meaning that the previous global Newton update was not able to provide a non-zero rate - i.e. it suggested a rate-update with wrong sign). 

Also added some more debug output.


steink:

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -827,11 +827,45 @@ namespace Opm
...
+                    this->openWell();
+                    deferred_logger.debug("    " + this->name() + " is re-opened after beeing stopped during local solve");
+                }
+                // Add debug info for switched controls
+                if (ws.production_cmode != pmode_orig || ws.injection_cmode != imode_orig) {
+                    std::string from,to;
+                    if (this->isInjector()) {
+                        from = WellInjectorCMode2String(imode_orig);
+                        to = WellInjectorCMode2String(ws.injection_cmode);
+
```
GitPaean, line 859: remove this blank line if not intentional please.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -827,11 +827,45 @@ namespace Opm
...
-            if (!converged) {
+            if (converged) {
+                const bool zero_target = this->wellUnderZeroRateTarget(simulator, well_state, deferred_logger);
+                if (this->wellIsStopped() && !zero_target && !(qtotal_orig == 0.0)) {
+                    // Well had non-zero rate, but was stopped during local well-solve. We re-open the well 
+                    // for the next global iteration, but if the zero rate persists, it will be stopped.
+                    // This logic is introduced to prevent/ameliorate stopped/revived oscillations  
+                    this->operability_status_.resetOperability();
+                    this->openWell();
+                    deferred_logger.debug("    " + this->name() + " is re-opened after beeing stopped during local solve");
```
atgeirr, line 851: Typo: "beeing"

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -831,7 +831,7 @@ namespace Opm
             const auto pmode_orig = ws.production_cmode;
             const auto imode_orig = ws.injection_cmode;
 
-            Scalar qtotal_orig = 0.0;
+            const Scalar qtotal_orig = 0.0;
```
GitPaean, line 834: This will not work. I was referring to the above

```c++
const auto& ws = ...
```

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -827,11 +827,45 @@ namespace Opm
...
+            }
+
             this->operability_status_.solvable = true;
             bool converged = this->iterateWellEquations(simulator, dt, well_state, group_state, deferred_logger);
 
-            // unsolvable wells are treated as not operable and will not be solved for in this iteration.
-            if (!converged) {
+            if (converged) {
+                const bool zero_target = this->wellUnderZeroRateTarget(simulator, well_state, deferred_logger);
+                if (this->wellIsStopped() && !zero_target && !(qtotal_orig == 0.0)) {
```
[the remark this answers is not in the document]
steink: The main reason for stop/revive oscillations is 
1. well gets stopped during local solve due to too high (low) reservoir pressure
2. global eqs solved with stopped well, Newton update decrease (increase) reservoir pressure around well (typically too much)
3. well gets reopened during local solve and get high rate due to low (high) reservoir pressure
4. global eqs solved with high rate well, Newton update increase (decrease) reservoir pressure (typically too much) 
5. ->1

The PR breaks this cycle by requiring that for a well to remained stopped after stopped in the local solve, then it should have had zero rate when it entered the local solve, basically meening that both the global and local solve agree it should be stopped.

For wells that should be stopped/shut, this approach potentially adds one iteration to the global Newton count. However for wells that should not be stopped/shut, it potentially removes two or more.

opm/simulators/wells/WellInterface_impl.hpp
steink: typo, now fixed

opm/simulators/wells/WellInterface_impl.hpp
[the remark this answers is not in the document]
GitPaean: Thanks for the explanation. Then let us add a little more comments for later reference.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -827,11 +827,44 @@ namespace Opm
         // only use inner well iterations for the first newton iterations.
         const int iteration_idx = simulator.model().newtonMethod().numIterations();
         if (iteration_idx < this->param_.max_niter_inner_well_iter_ || this->well_ecl_.isMultiSegment()) {
+            const auto& ws = well_state.well(this->indexOfWell());
+            const auto pmode_orig = ws.production_cmode;
+            const auto imode_orig = ws.injection_cmode;
+
+            Scalar qtotal_orig = 0.0;
+            const int np = well_state.numPhases();
```
GitPaean, line 835: This code can be written the following, 

```c++
const bool all_zero_rate_original  = std::all_of(ws.surface_rates.begin(), ws.surface_rates.begin() + well_state.numPhases(), [](Scalar rate) {
    return rate == Scalar(0.0);
});

We use `all_zero_rate_original` instead of `qtotal_orig == 0.` later. I think it will be easier to read. 

It needs the header file <algorithm>

opm/simulators/wells/WellInterface_impl.hpp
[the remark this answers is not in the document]
GitPaean: I suggested variable name change in another comment.

atgeirr:
Moved to #5659

bska:
Given that #5659 was merged into the master branch, do we still need to keep this PR open?

GitPaean:
> Given that #5659 was merged into the master branch, do we still need to keep this PR open?

This is not needed anymore. Closing it.

PR #5660, opened 2024-10-10
Remove unused num_cells argument from well equations

jakobtorben:
The `num_cells` argument is no longer used (and should not have been there to begin with as cells is already an argument and can be used to get the number of cells). This removes the warnings that #5341 introduced.


jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

jakobtorben:
Removing the argument cascaded into a lot more changes than I anticipated. But I still think we should consider merging this in. We are passing an argument around in 15 different files that is never used. Also there are members, local_num_cells_ and global_num_cells_, that has this information, so having a third num_cells can be confusing.

atgeirr:
Good cleanup! Merging.

PR #5664, opened 2024-10-11
in flowexp_blackoil, invalidate output data after solution is updated

GitPaean:
without invalidating the output data, many quantities will not be evaluated/updated and output to RESTART files.

### DIFF
--- a/opm/simulators/flow/BlackoilModel.hpp
+++ b/opm/simulators/flow/BlackoilModel.hpp
@@ -707,7 +707,6 @@ namespace Opm {
             {
                 OPM_TIMEBLOCK(invalidateAndUpdateIntensiveQuantities);
                 simulator_.model().invalidateAndUpdateIntensiveQuantities(/*timeIdx=*/0);
-                simulator_.problem().eclWriter()->mutableOutputModule().invalidateLocalData();
             }
         }
 
--- a/opm/simulators/flow/FlowProblemBlackoil.hpp
+++ b/opm/simulators/flow/FlowProblemBlackoil.hpp
@@ -436,6 +436,9 @@ public:
     {
         FlowProblemType::endTimeStep();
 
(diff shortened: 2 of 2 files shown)


GitPaean:
The comparison of the UNRST files are expected to fail.

GitPaean:
jenkins build this please

bska:
> The comparison of the UNRST files are expected to fail.

Okay.  In that case we either have to postpone updating the reference solutions and merging this PR until the 2024.10 release is final **or** this change has to go into the release branch as well.

GitPaean:
> > The comparison of the UNRST files are expected to fail.
> 
> Okay. In that case we either have to postpone updating the reference solutions and merging this PR until the 2024.10 release is final **or** this change has to go into the release branch as well.

This is the only case will be affected 
```
109/441 Test #144: compareECLFiles_flowexp_blackoil+SPE1CASE2 ........................***Failed    3.83 sec . 
```

I did not following the release progress, but if it is open to new update, I will suggest that we have it in the release, since it is a bug fix although not for the main simulator.

bska (commented):
I'm uneasy about this.  We already call `invalidateLocalData()` as part of [`updateSolution()`](https://github.com/OPM/opm-simulators/blob/c228f31287279bd7b93446efbbcb7c4ae72ef742/opm/simulators/flow/BlackoilModel.hpp#L710) and that really should be sufficient.

GitPaean:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  This looks good to me now.  I'll update the reference data and then merge this into both the master and the release branch for 2024.10.

bska:

bska:
jenkins build this opm-tests=1244 please

bska:
The new reference solutions have been installed on the CI system.  I'll merge this into the master branch and back-port it to the release branch.

PR #3708, opened 2021-11-23
Tracer rst init

joakim-hove:
Will load and initiialize tracer model from restart vectors. NB: This is code I am not veeeery familiar with, and the implementation is on a strict minimum change basis.

test update: https://github.com/OPM/opm-tests/pull/689

PR #5665, opened 2024-10-11
Use Rank's Local Cells for Tracer Concentration Restart

bska:
The `globalIdx` generated by `localIdxToGlobalIdx()` is in the range zero to number of active cells in global model, not zero to number of active cells on rank.  Using `globalIdx` as an argument to `set*TracerConcentration()` therefore ends up indexing out-of-bounds for the internal arrays in the tracer model object in parallel restarted simulation runs.

The problem has been present since the restart support for tracers was added in commit e9c45f4ca (PR #3708).

This commit switches to using the `elemIdx` instead, as we already do when calling the output module's `setRestart()` member function. Following this, we are able to restart the standard Norne benchmark case (NORNE_ATW2013.DATA) in parallel.  The numerical results for the tracer concentrations are, however, a little questionable and will warrant further investigation.  Nevertheless, the tracer values are now the same in a parallel restarted run as in a sequential restarted run.

### DIFF
--- a/opm/simulators/flow/EclWriter.hpp
+++ b/opm/simulators/flow/EclWriter.hpp
@@ -583,8 +583,8 @@ public:
 
                     for (auto elemIdx = 0*numElements; elemIdx < numElements; ++elemIdx) {
                         const auto globalIdx = this->collectOnIORank_.localIdxToGlobalIdx(elemIdx);
-                        tracer_model.setFreeTracerConcentration(tracer_index, globalIdx,
-                                                                free_tracer_solution[globalIdx]);
+                        tracer_model.setFreeTracerConcentration
+                            (tracer_index, elemIdx, free_tracer_solution[globalIdx]);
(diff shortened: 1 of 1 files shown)


bska:
Given the fact that this problem has apparently been in the code since its inception I am very surprised that we have not encountered this before.  @totto82: Would you, or one of your team, have a look at this please?

bska:
jenkins build this please

totto82:
Thanks for spotting this. I don't recall that parallel restart was tested for tracers so errors may have slipped through. From my point of view this can be merged.

bska:
> From my point of view this can be merged.

Much appreciated.  That said, I think there's still work to do in this area, since I get mismatch like the following between the base and restarted runs&ndash;be they sequential or parallel

---

field-level-SEA-tracer-comparison

bska:

PR #5668, opened 2024-10-14
FIX gas-water restart

totto82:
This removes and fixes some parts in the restart that asserts that there is an oil component to allow for restarts for gas-water cases. 

@bska I add the release tag since this is a bug fix.

### DIFF
--- a/opm/simulators/flow/EclWriter.hpp
+++ b/opm/simulators/flow/EclWriter.hpp
@@ -497,19 +497,18 @@ public:
         const auto gasActive = FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx);
         const auto waterActive = FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx);
         const auto enableSwatinit = simulator_.vanguard().eclState().fieldProps().has_double("SWATINIT");
-        const auto opm_rst_file = Parameters::Get<Parameters::EnableOpmRstFile>();
-        const auto read_temp = enableEnergy || (opm_rst_file && enableTemperature);
 
         std::vector<RestartKey> solutionKeys {
             {"PRESSURE", UnitSystem::measure::pressure},
             {"SWAT",     UnitSystem::measure::identity,    waterActive},
             {"SGAS",     UnitSystem::measure::identity,    gasActive},
-            {"TEMP",     UnitSystem::measure::temperature, read_temp},
+            {"TEMP",     UnitSystem::measure::temperature, enableEnergy},
             {"SSOLVENT", UnitSystem::measure::identity,    enableSolvent},
 
             {"RS",  UnitSystem::measure::gas_oil_ratio, FluidSystem::enableDissolvedGas()},
             {"RV",  UnitSystem::measure::oil_gas_ratio, FluidSystem::enableVaporizedOil()},
             {"RVW", UnitSystem::measure::oil_gas_ratio, FluidSystem::enableVaporizedWater()},
+            {"RSW", UnitSystem::measure::gas_oil_ratio, FluidSystem::enableDissolvedGasInWater()},
 
             {"SGMAX", UnitSystem::measure::identity, enableNonWettingHysteresis && oilActive && gasActive},
             {"SHMAX", UnitSystem::measure::identity, enableWettingHysteresis && oilActive && gasActive},
--- a/opm/simulators/flow/GenericOutputBlackoilModule.cpp
+++ b/opm/simulators/flow/GenericOutputBlackoilModule.cpp
@@ -802,9 +802,9 @@ setRestart(const data::Solution& sol,
             rswSol_[elemIdx] = sol.data<double>("RSWSOL")[globalDofIndex];
 
     }
-
-    assert(!saturation_[oilPhaseIdx].empty());
-    saturation_[oilPhaseIdx][elemIdx] = so;
+    if (!saturation_[oilPhaseIdx].empty())  {
+        saturation_[oilPhaseIdx][elemIdx] = so;
+    }
 
     auto assign = [elemIdx, globalDofIndex, &sol](const std::string& name,
                                                   ScalarBuffer& data)
--- a/opm/simulators/flow/OutputBlackoilModule.hpp
+++ b/opm/simulators/flow/OutputBlackoilModule.hpp
@@ -1146,12 +1146,18 @@ public:
             MaterialLaw::capillaryPressures(pc, matParams, fs);
             Valgrind::CheckDefined(this->fluidPressure_[elemIdx]);
             Valgrind::CheckDefined(pc);
-            assert(FluidSystem::phaseIsActive(oilPhaseIdx));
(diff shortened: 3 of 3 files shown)


totto82:
jenkins build this please

opm/simulators/flow/OutputBlackoilModule.hpp
```diff
@@ -1146,12 +1146,18 @@ class OutputBlackOilModule : public GenericOutputBlackoilModule<GetPropType<Type
...
+            const auto& pressure = this->fluidPressure_[elemIdx];
             for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
                 if (!FluidSystem::phaseIsActive(phaseIdx))
                     continue;
 
-                fs.setPressure(phaseIdx, this->fluidPressure_[elemIdx] + (pc[phaseIdx] - pc[oilPhaseIdx]));
+                if (Indices::oilEnabled)
+                    fs.setPressure(phaseIdx, pressure + (pc[phaseIdx] - pc[oilPhaseIdx]));
+                else if (Indices::gasEnabled)
+                    fs.setPressure(phaseIdx, pressure + (pc[phaseIdx] - pc[gasPhaseIdx]));
```
bska, line 1157: Did you verify that `pressure + (...)` matches the sign definition of the capillary pressure provided by `MaterialLaw::capillaryPressure()`?

totto82: Yes, as I understand the code, this is correct. 

This also aligns with what is done in https://github.com/OPM/opm-simulators/blob/c7b3e40bfae7db8fb964465bc89dc334d5f9ebfd/opm/simulators/flow/FlowProblemBlackoil.hpp#L1328

totto82:
I will do some more testing as further testing shows that there may be other issues when restarting gas-water cases.

totto82:
The restart does not work when --enable-opm-rst-file=true due to some missing well control input. For --enable-opm-rst-file=false(default), RSW and RVW is needed to get correct restart)

totto82:
jenkins build this please

opm/simulators/flow/GenericOutputBlackoilModule.cpp
```diff
@@ -493,8 +493,10 @@ assignToSolution(data::Solution& sol)
         DataEntry{"RPORV",    UnitSystem::measure::volume,                                rPorV_},
         DataEntry{"RS",       UnitSystem::measure::gas_oil_ratio,                         rs_},
         DataEntry{"RSSAT",    UnitSystem::measure::gas_oil_ratio,                         gasDissolutionFactor_},
+        DataEntry{"RSW",      UnitSystem::measure::gas_oil_ratio,                         rsw_},
         DataEntry{"RV",       UnitSystem::measure::oil_gas_ratio,                         rv_},
         DataEntry{"RVSAT",    UnitSystem::measure::oil_gas_ratio,                         oilVaporizationFactor_},
+        DataEntry{"RVW",      UnitSystem::measure::oil_gas_ratio,                         rvw_},
```
bska, line 499: The `RVW` array is an OPM extensions and must never be listed among the "base" solution arrays.  You must find another way to do this.

bska (commented):
> The restart does not work when `--enable-opm-rst-file=true` due to some missing well control input

And this is why `--enable-opm-rst-file=true` is an opt-in feature.  At **minimum** we need to remove the [`OPM_?WEL`](https://github.com/OPM/opm-common/blob/3a107a565f51d5ffd5cd0425ee4749b455bde9fa/opm/output/eclipse/RestartIO.cpp#L460-L470) arrays if we're even to think about restoring the well controls for the extended restart files.  Even then it's not clear to me that we'll have a consistent set of objects from the extended restart files.

Putting `RSW` into the base solution arrays may also break compatibility and should be tested **a lot** before merging.

opm/simulators/flow/GenericOutputBlackoilModule.cpp
totto82: I must admit I don't fully understand why we can not add these arrays to the "base" solution since they are only conditionally added if OPM specific keywords are used. Do you have a suggestion on how this can be done? I really think we should have a solution that makes it possible to restart CO2 runs. For now I have removed the last commit that adds the RSW/RVW arrays. For cases without VAPWAT and DISGASW this is sufficient for restart.

bska: > I must admit I don't fully understand why we can not add these arrays to the "base" solution since they are only conditionally added if OPM specific keywords are used.

We've already had far too much trouble with arrays being added to the solution set and breaking compatibility, notably the temperature array, that I don't want to tempt fate. We need to be really careful about these lest we break some production workflows.

> Do you have a suggestion on how this can be done?

Like I said, at minimum we need to remove the `OPM_?WEL` arrays.  Then, and only then, can we begin to explore more specific conditions for these arrays.

> I really think we should have a solution that makes it possible to restart CO2 runs.

Sure, but that has to be a lower priority than keeping existing workflows running.  If you want to do a quick local hack on your machine you can call `restore_wells_ecl()` unconditionally to form the well solution object `xw` in [`RestartIO::load()`](https://github.com/OPM/opm-common/blob/3a107a565f51d5ffd5cd0425ee4749b455bde9fa/opm/output/eclipse/LoadRestart.cpp#L1607-L1609).  If that's enough to enable CO2 simulation restart with `--enable-opm-rst-file=true`, then it should be fairly quick.  If not, then we have more work to do.

totto82: the hack (https://github.com/OPM/opm-common/pull/4261)  + the changes in the second commit in this PR is sufficient for restart.

bska:
jenkins build this please

opm/simulators/flow/EclWriter.hpp
```diff
@@ -498,7 +498,7 @@ class EclWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::Grid>
         const auto waterActive = FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx);
         const auto enableSwatinit = simulator_.vanguard().eclState().fieldProps().has_double("SWATINIT");
         const auto opm_rst_file = Parameters::Get<Parameters::EnableOpmRstFile>();
-        const auto read_temp = enableEnergy || (opm_rst_file && enableTemperature);
+        const auto read_temp = enableEnergy;
```
bska, line 501: This change means that the `opm_rst_file` object created on the line above becomes unused.  Is that intentional?  If so, please remove the `opm_rst_file` object.  If not, then please use it where it's needed.

bska (commented):
Thanks a lot for the updates.  This looks good now, for the most part, except for one object that's now unused.  Please resolve that issue.

bska:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  This looks good to me now and as the build check is green I'll merge into master.  I'll also back-port this to the release branch to get it into the second release candidate for OPM 2024.10.

PR #5669, opened 2024-10-14
using reference for the few states in EclWriteTasklet

GitPaean:


GitPaean:
jenkins build this please

akva2:
these copies are on purpose since write runs on a separate thread.

GitPaean:
Thanks for the information.  Then this PR is not valid.

PR #5675, opened 2024-10-15
Fix memory bug in setPrimaryVars for MSW

jakobtorben:
Sets the correct start of memcpy to 

```cpp
for (int ii = 0; ii < num_seg; ++ii) {
     const auto start = it + ii * num_eq;
```

such that it starts from `segment_index * num_eq` instead of `num_seg * num_eq`.

Note that this function is only used when using the NLDD nonlinear solver, where we have an unconverged local NLDD domain, which has a multisegment well on that domain. This was exposed after #5341 was merged which motivated to try NLDD with more complex cases.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -2237,7 +2237,7 @@ namespace Opm
         constexpr int num_eq = MSWEval::numWellEq;
         std::array<Scalar, num_eq> tmp;
         for (int ii = 0; ii < num_seg; ++ii) {
-            const auto start = it + num_seg * num_eq;
+            const auto start = it + ii * num_eq;
             std::copy(start, start + num_eq, tmp.begin());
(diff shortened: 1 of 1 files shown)


jakobtorben:
jenkins build this please

bska (approved):
Good catch!  Seems to have been present since the member function was introduced in commit bc9cfc8cd5 (PR #4704).  A slightly alternative phrasing might be something along the lines of
```C++
for (int ii = 0; i < num_seg; ++ii, it += num_eq) {
    std::copy(it, it + num_eq, tmp.begin());
    this->primary_variables(ii, tmp);
}
```
which (slightly) reduces the risk of this problem.  In C++20 I could imagine passing a `std::span<Scalar>` as an argument to `setValue()` instead and bypass the step of copying into a `std::array<>`.

bska:
Build is green.  I'll merge this into the master branch and port it to the release branch ahead of 2024.10-RC2.

PR #5682, opened 2024-10-17
Some lambda fixes

akva2:
- Remove an unused capture
- Restructure to avoid a capture that causes warnings in clang but is required by g++-14

### DIFF
--- a/opm/simulators/flow/FlowProblemBlackoil.hpp
+++ b/opm/simulators/flow/FlowProblemBlackoil.hpp
@@ -1524,7 +1524,7 @@ protected:
         using ViolationLevel = typename Satfunc::PhaseChecks::
             SatfuncConsistencyCheckManager<Scalar>::ViolationLevel;
 
-        auto reportFailures = [&sfuncConsistencyChecks, this]
+        auto reportFailures = [&sfuncConsistencyChecks]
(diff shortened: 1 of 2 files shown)


atgeirr (approved):

opm/simulators/flow/FlowProblemBlackoil.hpp
```diff
@@ -1524,7 +1524,7 @@ class FlowProblemBlackoil : public FlowProblem<TypeTag>
         using ViolationLevel = typename Satfunc::PhaseChecks::
             SatfuncConsistencyCheckManager<Scalar>::ViolationLevel;
 
-        auto reportFailures = [&sfuncConsistencyChecks, this]
+        auto reportFailures = [&sfuncConsistencyChecks]
```
bska, line 1527: Thanks!  I forgot to remove `this` in one of the later updates here.

bska (approved):
Looks good to me.  Please merge when green.

akva2:
The second commit here fixes a build error with g++-14 so I think it should be backported.

bska:
> The second commit here fixes a build error with g++-14 so I think it should be backported.

Very good.  I'll port it now.

PR #5686, opened 2024-10-21
Bugfixes #5609

vkip:
This intends to fix a couple of issues introduced with #5609:
* Properly add inactive wells to all ranks with connections (but not others) ( ParallelWells <string,bool> instead of <string,int> )
  * Also inform the users about these wells (and where they end up)
* Disallow splitting inactive wells for parallel restarts (currently causes issues, will look into it)
  * Also introduces a command-line option to disallow this for non-restarts.

Sorry for the mess!

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -695,9 +695,6 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
     for (int w = 0; w < nw; ++w) {
         const auto& well_ecl = wells_ecl[w];
         auto& ws = this->well(w);
-        // If the phase_rates has zero size this is an inactive well that will never be solved
-        if (ws.perf_data.phase_rates.size() == 0)
-            continue;
 
         if (well_ecl.isMultiSegment()) {
             const WellSegments& segment_set = well_ecl.getSegments();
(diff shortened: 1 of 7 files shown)


vkip:
jenkins build this please

vkip:
jenkins build this please

opm/simulators/wells/WellState.cpp
```diff
@@ -695,9 +695,6 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
     for (int w = 0; w < nw; ++w) {
         const auto& well_ecl = wells_ecl[w];
         auto& ws = this->well(w);
-        // If the phase_rates has zero size this is an inactive well that will never be solved
-        if (ws.perf_data.phase_rates.size() == 0)
-            continue;
```
blattms, line 700: Is this  a general bugfix independent of the changes in this and the earlier PR?

blattms (commented):
Thanks and I am sorry for the mediocre review. I should have catched some of this.

Do we have an idea why restart is problematic. That might hold for all distributed wells, then.

Looks good. Just one curious question about one bugfix

opm/simulators/wells/WellState.cpp
vkip: No, I unfortunately introduced this in the earlier PR to counteract the side-effects of my previous misunderstanding.. Sorry for the confusion, and thanks for the swift handling of this bugfix!

blattms: Yeah, doesn't really speak for me if you say 4 days is swift. :sweat_smile: 
Sorry. I was a little occupied with other stuff.

PR #5687, opened 2024-10-21
ecl style output for compositional simulation 

GitPaean:
it depends on the upstream PR https://github.com/OPM/opm-common/pull/4273 . 

as said there, it has been refined a lot, while still in the prototyping stage.  In the process of cleaning up before discussing several key design elements in the approach. 

The goal is trying to reuse a much as possible the existing facility.


opm/simulators/flow/GenericOutputBlackoilModule.cpp
```diff
@@ -659,6 +688,15 @@ assignToSolution(data::Solution& sol)
                    std::move(this->saturation_[gasPhaseIdx]),
                    data::TargetType::RESTART_SOLUTION);
     }
+    
+    if (FluidSystem::phaseIsActive(oilPhaseIdx) &&
```
GitPaean, line 692: Here, it will make OPM-flow output SOIL, which is not favored.

GitPaean:
jenkins build this opm-common=4273 please

GitPaean:
This PR works as intended for the compositional running. Now it outputs INIT, EGRID, SMSPEC, UNRST, and UNSMRY file as intended.  To avoid updating the reference before the release, I am not creating a regression tests in this PR yet.  

The PR itself still has more things to improve, for example, splitting the [GenericOutputBlackoilModule.cpp](https://github.com/OPM/opm-simulators/pull/5687/files#diff-a71bfa0107237a8f75121b64430fd2b936be082ce373f287503568e405270f1e) so that compostional and blackoil can have their own classes, especially to avoid the instantiation always for both compositional and blackoil .  After discussing with @bska , we agreed to have the refactoring in a separate PR with a regression checking, which will be pure refactoring PR. Difficulties is that there are many things in common between compositional and black oil simulations and it might be quite some work to distinguish them properly. 

And also, there are some code duplication in the  `OutputCompositionalModule.hpp`. 

I will mark this PR for review now so the discussion can start.

opm/simulators/flow/FlowProblemComp.hpp
```diff
@@ -220,6 +243,53 @@ class FlowProblemComp : public FlowProblem<TypeTag>
...
+    /*!
+     * \brief Called by the simulator after each time integration.
+     */
+    void endTimeStep() override
+    {
+        FlowProblemType::endTimeStep();
+
+        const bool isSubStep = !this->simulator().episodeWillBeOver();
+
+        // after the solution is updated, the values in output module needs also updated
```
akva2, line 255: also needs to be updated

opm/simulators/flow/GenericOutputBlackoilModule.hpp
```diff
@@ -539,6 +541,10 @@ class GenericOutputBlackoilModule {
     std::array<ScalarBuffer, numPhases> viscosity_;
     std::array<ScalarBuffer, numPhases> relativePermeability_;
 
+    // totoal mole fractions for each component
```
akva2, line 544: total

opm/simulators/flow/EclWriter.hpp
```diff
@@ -30,23 +30,30 @@
...
 #include <opm/simulators/flow/countGlobalCells.hpp>
 #include <opm/simulators/flow/EclGenericWriter.hpp>
 #include <opm/simulators/flow/FlowBaseVanguard.hpp>
-#include <opm/simulators/flow/OutputBlackoilModule.hpp>
 #include <opm/simulators/timestepping/SimulatorTimer.hpp>
 #include <opm/simulators/utils/DeferredLoggingErrorHelpers.hpp>
 #include <opm/simulators/utils/ParallelRestart.hpp>
 #include <opm/simulators/utils/ParallelSerialization.hpp>
 
+#include <boost/date_time/posix_time/posix_time.hpp>
```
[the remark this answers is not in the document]
GitPaean: It is needed in the following line in function `writeInitialFIPReport()`. 

```c++
            boost::posix_time::ptime start_time = boost::posix_time::from_time_t(simulator_.vanguard().schedule().getStartTime());
```

Somehow, when I made `OutputModule` as a template parameter, it created some compilation errors. This header was added along with the a few headers above.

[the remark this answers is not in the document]
akva2: right. did not see "old sins" from the diff.

bska:
Just a meta-question before I start looking at this in detail: Do you intend to coalesce (some of) the commits here before merging the PR?

examples/problems/co2ptflashproblem.hh
```diff
@@ -498,6 +498,24 @@ private:
...
+        // determine the component total fractions
+        // TODO: duplicated code here, while should be refactored out when we swithing
+        // to starting from total mole fractions
+        Dune::FieldVector<Scalar, numComponents> z(0.0);
+        Scalar sumMoles = 0.0;
+        for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
+            for (unsigned compIdx = 0; compIdx < numComponents; ++compIdx) {
+                Scalar tmp = Opm::getValue(fs.molarity(phaseIdx, compIdx) * fs.saturation(phaseIdx));
+                z[compIdx] += Opm::max(tmp, 1e-8);
+                sumMoles += tmp;
```
bska, line 510: If `tmp < 1.0e-8` then `sumMoles` is, strictly speaking, not the sum of the elements of `z`.  Is that okay?

GitPaean:
> Just a meta-question before I start looking at this in detail: Do you intend to coalesce (some of) the commits here before merging the PR?

was thinking to do after the review. but let me do it now and also including the upstream https://github.com/OPM/opm-common/pull/4273.

opm/simulators/flow/FlowProblemComp.hpp
```diff
@@ -492,8 +589,13 @@ class FlowProblemComp : public FlowProblem<TypeTag>
         throw std::logic_error("polymer is disabled for compositional modeling and you're trying to add polymer to BC");
     }
 
+    FlowThresholdPressure<TypeTag> thresholdPressures_;
+
     std::vector<InitialFluidState> initialFluidStates_;
 
+    bool enableEclOutput_;
```
bska, line 596: Please always initialise `bool`s&ndash;typically to `false`.

opm/simulators/flow/GenericOutputBlackoilModule.cpp
```diff
@@ -624,6 +628,38 @@ assignToSolution(data::Solution& sol)
...
 
+    // basically, for compositional, we can not use std::array for this.  We need to generate the ZMF1, ZMF2, and so on
+    // and also, we need to map these values.
+    // TODO: the following should go to a function
+    if (this->isCompositional_) {
+        auto compositionalEntries = std::vector<DataEntry>{};
+        {
+            // ZMF
+            for (int i = 0; i < numComponents; ++i) {
+                const std::string name = "ZMF" + std::to_string(i + 1);  // Generate ZMF1, ZMF2, ...
```
bska, line 639: You have `fmt::format()` in this file, so it's a lot better to write this as
```C++
const auto name = fmt::format("ZMF{}", i + 1);
```
instead.

examples/problems/co2ptflashproblem.hh
```diff
@@ -498,6 +498,24 @@ private:
...
 
+        // determine the component total fractions
+        // TODO: duplicated code here, while should be refactored out when we swithing
+        // to starting from total mole fractions
+        Dune::FieldVector<Scalar, numComponents> z(0.0);
+        Scalar sumMoles = 0.0;
+        for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
+            for (unsigned compIdx = 0; compIdx < numComponents; ++compIdx) {
+                Scalar tmp = Opm::getValue(fs.molarity(phaseIdx, compIdx) * fs.saturation(phaseIdx));
+                z[compIdx] += Opm::max(tmp, 1e-8);
```
bska, line 509: We're inside the `Opm` namespace, so you probably should not use the `Opm::` namespace prefix here.  Alternatively, you should write these as `::Opm::getValue(...)` and `... += ::Opm::max(...)` instead.

On a second note, `fs.saturation(phaseIdx)` is (hopefully) constant for each `compIdx`, so you could arguably write
```C++
for (phaseIdx) {
    const auto s = getValue(fs.saturation(phaseIdx))
    for (compidx) {
        const auto tmp = getValue(fs.molarity(phaseIdx, compIdx)) * s;
        ...
    }
}
```
instead.

GitPaean:
cleaning up of the commits is done for this PR and https://github.com/OPM/opm-common/pull/4273.

bska (commented):
I won't claim to have studied the details here, but it looks good for the most part.  I've identified a couple of minor points.

examples/problems/co2ptflashproblem.hh
GitPaean: It is originally through the commit OPM/opm-common@1d2db38172539e0dd722e884e647824db06ceacb 

It is used multiple places and it might be used to treat unexpected values.  We might not need it but I do not think it will affect the actual results. 

But I will make a follow up PR to check whether we can get rid of all of them (there are four usages, and three of them are in opm-common). 

It is possible that it was introduced in the early stage of the development to get things run through.

examples/problems/co2ptflashproblem.hh
GitPaean: good catch! done fixing.

opm/simulators/flow/GenericOutputBlackoilModule.cpp
GitPaean: done fixing.

opm/simulators/flow/FlowProblemComp.hpp
GitPaean: fixed.

examples/problems/co2ptflashproblem.hh
```diff
@@ -498,6 +498,25 @@ private:
...
 
+        // determine the component total fractions
+        // TODO: duplicated code here, while should be refactored out when we swithing
+        // to starting from total mole fractions
+        Dune::FieldVector<Scalar, numComponents> z(0.0);
+        Scalar sumMoles = 0.0;
+        for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
+            const auto saturation = getValue(fs.saturation(phaseIdx));
+            for (unsigned compIdx = 0; compIdx < numComponents; ++compIdx) {
+                Scalar tmp = getValue(fs.molarity(phaseIdx, compIdx) * saturation);
```
bska, line 509: If you write `getValue(fs.molarity(...)) * saturation` instead of `getValue(fs.molarity(...) * saturation)` then the resulting operation is never more expensive than what's here.  The former is `Scalar * Scalar` whereas the latter is `getValue(something * Scalar)` which **could** be more expensive.

opm/simulators/flow/FlowProblemComp.hpp
```diff
@@ -290,13 +361,29 @@ class FlowProblemComp : public FlowProblem<TypeTag>
             fs.setViscosity(FluidSystem::gasPhaseIdx, FluidSystem::viscosity(fs, paramCache, FluidSystem::gasPhaseIdx));
         }
 
+        // determine the component fractions
+        Dune::FieldVector<Scalar, numComponents> z(0.0);
+        Scalar sumMoles = 0.0;
+        for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
+            for (unsigned compIdx = 0; compIdx < numComponents; ++compIdx) {
+                Scalar tmp = Opm::getValue(fs.molarity(phaseIdx, compIdx) * fs.saturation(phaseIdx));
```
bska, line 369: Here's another instance of `fs.saturation()`  which could be hoisted outside the `compIdx` loop.

opm/simulators/flow/GenericOutputBlackoilModule.cpp
```diff
@@ -624,6 +628,38 @@ assignToSolution(data::Solution& sol)
...
 
+    // basically, for compositional, we can not use std::array for this.  We need to generate the ZMF1, ZMF2, and so on
+    // and also, we need to map these values.
+    // TODO: the following should go to a function
+    if (this->isCompositional_) {
+        auto compositionalEntries = std::vector<DataEntry>{};
+        {
+            // ZMF
+            for (int i = 0; i < numComponents; ++i) {
+                const auto name = fmt::format("ZMF{}", i + 1);;  // Generate ZMF1, ZMF2, ...
```
bska, line 639: Tiny nit: This statement ends in a doubled-up semicolon (`;;`).  You should remove one of those.

opm/simulators/flow/FlowProblemComp.hpp
bska: > fixed.

Thanks.  I think we should initialise `enableEclOutput_` as well, not just `enableVtkOutput_`.

bska (commented):
Thanks a lot for the updates.  This is starting to look good to me with a couple of exceptions.

examples/problems/co2ptflashproblem.hh
GitPaean: I see your point now. Let me think a little.

GitPaean: I changed the code to be as follows, 

```c++
                Scalar tmp = getValue(fs.molarity(phaseIdx, compIdx)) * saturation;
                tmp = max(tmp, 1.e-8);
                z[compIdx] += tmp;
                sumMoles += tmp;
```

opm/simulators/flow/FlowProblemComp.hpp
GitPaean: It was a mistake. I will initialize both now.

GitPaean:
Thanks for the comments. I have addressed them and requested a jenkins checking from https://github.com/OPM/opm-common/pull/4273.

akva2:
No, I don't think we can accept them, as they are actual logic errors. 

The ```waterCompIdx``` is -1 in the FluidSystem and we are thus deref'ing at size_t(-1). I see no other approach than splitting the tables by component and only instance if we have a valid component index. Sure, we are not using the values that are read but it really can't go in as-is.

GitPaean:
> No, I don't think we can accept them, as they are actual logic errors.
> 
> The `waterCompIdx` is -1 in the FluidSystem and we are thus deref'ing at size_t(-1). I see no other approach than splitting the tables by component and only instance if we have a valid component index. Sure, we are not using the values that are read but it really can't go in as-is.

Is using std::vector acceptable instead of std::array here. If using std::vector, we can choose to only add when the index are  valid.

GitPaean:
It looks like the warnings are gone. 

https://ci.opm-project.org/job/opm-common-PR-builder/7654/

bska:
> It looks like the warnings are gone.

Much appreciated!  Thanks a lot for the quick turn-around.

bska:
The build check triggered from the upstream PR is green and there are no new warnings.  I'll merge this into the master branch now.

PR #5688, opened 2024-10-23
Add mixed precision double diagonal gpu ilu0 variants

multitalentloes:
Adds another mixed precision variant to both DILU and ILU0 on the GPU.
This version stores the off-diagonals in float and the diagonal in double. 
Tested to give reasonable iteration counts and speedup on large cases of spe11c.

Introduces a new enum class that keeps track of which mixed precision scheme is being used as we now have 3 choices and a boolean no longer suffices.


kjetilly:
Jenkins build this please

opm/simulators/linalg/gpuistl/detail/kernelEnums.hpp
```diff
@@ -0,0 +1,46 @@
...
+#define OPM_GPUISTL_PRECONDITIONER_STORAGE_OPTION_HPP
+
+/*
+    This file is here to organize a growing amount of different mixed precision options for the preconditioners.
+*/
+
+namespace Opm::gpuistl {
+    enum MixedPrecisionScheme {
+        DEFAULT,
+        STORE_ENTIRE_FACTORIZATION_AS_FLOAT,
```
kjetilly, line 30: You probably want to give them explicit values here, just to be safe, makes it easier to reference them for the user as well. (you never know when you add or remove an option, and then suddenly the ordering gets messed up)

opm/simulators/linalg/gpuistl/detail/kernelEnums.hpp
```diff
@@ -0,0 +1,46 @@
...
+    inline bool isValidMixedPrecisionScheme(int scheme) {
+        switch (scheme) {
+            case DEFAULT:
+            case STORE_ENTIRE_FACTORIZATION_AS_FLOAT:
+            case STORE_ONLY_FACTORIZED_DIAGONAL_AS_DOUBLE:
+                return true;
+            default:
+                return false;
+        }
+    }
```
kjetilly, line 43: add the conversion function with error checking here.

kjetilly (commented):
Some changes suggested.

opm/simulators/linalg/gpuistl/detail/kernel_enums.hpp
```diff
@@ -0,0 +1,112 @@
...
+namespace Opm::gpuistl {
+    // Mixed precision schemes used for storing the matrix in GPU memory
+    enum MatrixStorageMPScheme {
+        DOUBLE_DIAG_DOUBLE_OFFDIAG = 0, // full precision should be default
+        FLOAT_DIAG_FLOAT_OFFDIAG = 1,
+        DOUBLE_DIAG_FLOAT_OFFDIAG = 2
+    };
+
+    namespace detail {
+        __host__ __device__ constexpr bool storeDiagonalAsFloat(MatrixStorageMPScheme scheme);
```
kjetilly, line 38: if you are defining the functions in the same file, there is no need to declare it separately. That is to say, you can remove these lines and just keep the ones below.

opm/simulators/linalg/gpuistl/detail/kernel_enums.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+#include <cuda_runtime.h>
+
+/*
+    This file organizes a growing amount of different mixed precision options for the preconditioners.
+*/
+
+namespace Opm::gpuistl {
+    // Mixed precision schemes used for storing the matrix in GPU memory
+    enum MatrixStorageMPScheme {
```
kjetilly, line 31: can be a made a `enum class Matrix...` for slightly better name spacing.

kjetilly (commented):
very minor adjustments

opm/simulators/linalg/gpuistl/detail/kernel_enums.hpp
multitalentloes: It works for the first three functions at least, so a bit cleaner. The `isValidMatrixStorageMPScheme` causes a compilation error if it is not declared before the `makeMatrixStorageMPScheme` which calls it.

kjetilly (approved):

kjetilly:
Jenkins build this please

PR #5690, opened 2024-10-25
Get WellMatcher Object Directly From Schedule

bska:
There's no need for us to invoke the `WellMatcher` constructor when the `Schedule` object already knows how to do that.

Missed opportunity in commit ff51c2d7e (PR #3050).


bska:
jenkins build this please

blattms:
Nice. Thanks.

PR #5691, opened 2024-10-25
enableThresholdPressure_ initialized to be false

GitPaean:
in GenericThresholdPressure

I saw some compiler dependent behavoir related to this with the flowexp_comp simulator.


GitPaean:
It can be because we did not call fnishInit() there. 

But I think it is good to initialize it.

GitPaean:
jenkins build this please

bska (approved):
> But I think it is good to initialize it.

True.  We should always initialise our data members.  Please merge when green.

bska:
> Please merge when green.

PR #5693, opened 2024-10-28
Make WELPI Infrastructure for ACTIONX Fully Private

bska:
This is in preparation of making the match set for `ACTIONX` more general.  Having a fully private `WELPI` framework means we can change the types more freely without having to rebuild all users of `ActionHandler`.

While here, also add Doxygen-style documentation to the `ActionHandler`.

### DIFF
--- a/opm/simulators/flow/ActionHandler.cpp
+++ b/opm/simulators/flow/ActionHandler.cpp
@@ -22,6 +22,7 @@
 */
 
 #include <config.h>
+
 #include <opm/simulators/flow/ActionHandler.hpp>
 
 #include <opm/common/OpmLog/OpmLog.hpp>
@@ -29,6 +30,7 @@
 #include <opm/common/TimingMacros.hpp>
 
 #include <opm/input/eclipse/EclipseState/EclipseState.hpp>
+
 #include <opm/input/eclipse/Schedule/Action/ActionContext.hpp>
 #include <opm/input/eclipse/Schedule/Action/Actions.hpp>
 #include <opm/input/eclipse/Schedule/Action/ActionX.hpp>
@@ -40,6 +42,8 @@
 #include <opm/input/eclipse/Schedule/Well/WellMatcher.hpp>
 
 #include <opm/simulators/wells/BlackoilWellModelGeneric.hpp>
+
+#include <opm/simulators/utils/ParallelCommunication.hpp>
 #include <opm/simulators/utils/ParallelSerialization.hpp>
 
 #include <chrono>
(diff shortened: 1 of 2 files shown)


bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

opm/simulators/flow/ActionHandler.cpp
```diff
@@ -128,6 +132,61 @@ namespace {
...
+        if (comm.size() > 1) {
+            std::vector<Scalar> wellpi_buffer(num_wells * comm.size());
+            comm.gather(wellpi_vector.data(), wellpi_buffer.data(), num_wells, 0);
+
+            if (comm.rank() == 0) {
+                for (int rank = 1; rank < comm.size(); ++rank) {
+                    for (std::size_t well_index = 0; well_index < num_wells; ++well_index) {
+                        const auto global_index = rank*num_wells + well_index;
+                        const auto value = wellpi_buffer[global_index];
+                        if (value != Scalar{0}) {
```
akva2, line 173: does this if really serve any purpose?

opm/simulators/flow/ActionHandler.hpp
```diff
@@ -31,19 +31,20 @@
...
+} // namespace Opm::Action
 
+namespace Opm {
 template<class Scalar> class BlackoilWellModelGeneric;
 class EclipseState;
 class Schedule;
 struct SimulatorUpdate;
 class SummaryState;
 class UDQState;
+} // maespace Opm
```
akva2, line 45: namespace

akva2 (approved):
Looks good. One typo, one question.

opm/simulators/flow/ActionHandler.cpp
bska: > does this if really serve any purpose?

I think **as written** this serves a purpose, because this is rank 0 accumulating contributions from all other ranks.  It is quite possible that some other rank does not have a value for the well identified by `well_index` and we need to ensure that we don't overwrite a valid non-zero value with a zero value.

That said, I **think** it's possible to make this logic more lean.  If (and only if) we guarantee that [`ActionX::wellpi_wells()`](https://github.com/OPM/opm-simulators/blob/5f54883d4cfa7c92d034b1ab44440abf1b322cb9/opm/simulators/flow/ActionHandler.cpp#L147-L148) returns the same result, and in the same order, on all ranks then we can simplify this to something along the lines of
```C++
auto wellPI = vector<Scalar>(wellpi_wells.size());
for (auto i = 0*wellpi_wells.size(); i < wellpi_wells.size(); ++i) {
    if (wellModel.has(wellpi_wells[i])) {
        wellPI[i] = wellModel.wellPI(wellpi_wells[i]);
    }
}
comm.max(wellPI.data(), wellPI.size());
for (auto i = 0*wellpi_wells.size(); i < wellpi_wells.size(); ++i) {
    wellpi.emplace(i, wellPI[i]);
}
```
Such an approach will, as an added benefit, also end up communicating fewer values than we currently do in our gather + accumulate +  broadcast algorithm.  I remember being a bit hesitant about our current algorithm when it was first introduced in PR #3050 (commit ff51c2d7e), but at the time I was not able to identify what the alternative would be.

opm/simulators/flow/ActionHandler.hpp
bska: > namespace

Good catch.  Wow&ndash;that's a truly messed up spelling :fearful:

opm/simulators/flow/ActionHandler.cpp
akva2: right. then i think it would be worth going for the simplification. isn't the order guaranteed by the seqIndex ordering so a max reduction on the wellpi_vector should be fine?

bska: > right. then i think it would be worth going for the simplification.

I agree.  I'll do that in follow-up work, however, because I need to verify a couple of more things.

> isn't the order guaranteed by the seqIndex ordering so a max reduction on the wellpi_vector should be fine?

Yes, I **believe** so, but I need convince myself that I've not overlooked something somewhere.

bska:
jenkins build this please

bska:
PR approved and build check is green.  I'll merge this into the master branch now with the provision that I'll simplify the communication logic in follow-up work.

opm/simulators/flow/ActionHandler.cpp
bska: Quick update here: `ActionX::wellpi_wells()` currently returns a `vector<string>` for which the element order depends on the [iterator order](https://github.com/OPM/opm-common/blob/a22d214b2ac878aa1b128eea07863c69a60040f6/opm/input/eclipse/Schedule/Action/ActionX.cpp#L352) of an [`unordered_set<string>`](https://github.com/OPM/opm-common/blob/a22d214b2ac878aa1b128eea07863c69a60040f6/opm/input/eclipse/Schedule/Action/ActionX.cpp#L335).  While it's **likely** that all ranks will return the same elements in the same order, we can't really guarantee it.  I am however working on a larger update of our ACTIONX code and I'll be sure to add the logic to provide the requisite guarantee.  At that point I'll simplify the communication logic.

PR #5695, opened 2024-10-29
dont switch from GRUP if check_group_constrains_well_iterations=false

totto82:
As the code is now switching to/from GRUP in the local iterations is problematic as the group reduction rates (the combined rates of all wells that are not on group rate) are not updated. With this PR, individual controls are not checked if the well is on GRUP if check-group-constrains-well-iterations=false. 

I would also suggested to set check-group-constrains-well-iterations=false as default while fixing the reduction rate.


totto82:
jenkins build this please

steink (approved):
I think this makes sense. I've tested on a couple of other models and did not see any unexpected side-effects. As discussed, will look further into a few options for getting rid of the `check_group_constraints_inner_well_iterations`-option.

totto82:
jenkins build this please

PR #5696, opened 2024-10-29
Dont iterate when computing bhp from thp for computing of the well potentials for gaslift optimization

totto82:
This gives a 10 times speed up of the gaslift optimization for many cases as we dont repeatedly try to compute bhp from thp for alq values that dont converge anyway.

### DIFF
--- a/opm/simulators/wells/GasLiftSingleWell_impl.hpp
+++ b/opm/simulators/wells/GasLiftSingleWell_impl.hpp
@@ -143,7 +143,8 @@ computeBhpAtThpLimit_(Scalar alq, bool debug_output) const
         this->simulator_,
         this->summary_state_,
         alq,
-        this->deferred_logger_);
+        this->deferred_logger_,
+        /*iterate_if_no_solution */ false);
     if (bhp_at_thp_limit) {
         if (*bhp_at_thp_limit < this->controls_.bhp_limit) {
             if (debug_output && this->debug) {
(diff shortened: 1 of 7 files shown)


totto82:
I have tested this for different cases and I cannot find any downside to it so I mark it as ready for review.

opm/simulators/wells/GasLiftSingleWell_impl.hpp
```diff
@@ -143,7 +143,7 @@ computeBhpAtThpLimit_(Scalar alq, bool debug_output) const
         this->simulator_,
         this->summary_state_,
         alq,
-        this->deferred_logger_);
+        this->deferred_logger_, /*iterate_if_no_solution */false);
```
GitPaean, line 146: Very tiny nitpicking here,  a space before `false` and this arguments goes to a separate line to be consistent with other arguments.

opm/simulators/wells/WellInterface.hpp
```diff
@@ -190,7 +190,7 @@ class WellInterface : public WellInterfaceIndices<GetPropType<TypeTag, Propertie
     computeBhpAtThpLimitProdWithAlq(const Simulator& ebos_simulator,
                                     const SummaryState& summary_state,
                                     const Scalar alq_value,
-                                    DeferredLogger& deferred_logger) const = 0;
+                                    DeferredLogger& deferred_logger, bool iterate_if_no_solution = true) const = 0;
```
GitPaean, line 193: I would avoid to have default arguments for virtual class and subclass, maybe other colleagues can weigh in here. @bska @akva2

totto82:
jenkins build this please

GitPaean:
Thanks for the update. It looks good to me now. I am getting it into the master.

PR #5697, opened 2024-10-29
Handle WELPI from ACTIONX Separately From Constraints

bska:
This commit switches to using the `welpi_wells` information from the `SimulatorUpdate` structure (OPM/opm-common#4292) as a basis to decide the wells for which it is safe/sufficient to update only the CTFs in response to an ACTIONX block running WELPI.  To this end, we split the actions of the existing member function `updateEclWells()` into two parts,

  1. `updateEclWellsConstraints()`
  2. `updateEclWellsCTFFromAction()`

in which the first handles well status and well control updates while the second deals with CTF updates in response to WELPI.  We do not run the second part if the well structure has changed&ndash;e.g., due to COMPDAT or WELOPEN&ndash;since the update loop depends on a static connection topology.

We add a new member function `wellUpdateLoop()` which will traverse a sequence of well names and invoke a loop body on those wells which exist in `wells_ecl_`.  This collects common operations needed for both the constraints and the CTF updates.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -75,6 +75,7 @@
 #include <tuple>
 #include <unordered_map>
 #include <unordered_set>
+#include <utility>
 #include <vector>
 
 #include <fmt/format.h>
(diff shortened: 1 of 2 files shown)


bska:
This PR must be reivewed/merged alongside the upstream PR OPM/opm-common#4292.  I am creating the PR in a draft state pending empirical verification that it solves the same problem as #5624.

akva2 (approved):

bska:
PR approved and build check triggered from upstream companion PR is green.  I'll merge into the master branch and back-port it to the release branch for OPM 2024.10 as well.

PR #5699, opened 2024-10-29
Add Logging for Python Actions

bska:
Similar to the logging facilities we have for the regular ACTIONX keyword, although without an explicit list of "matching" wells.


bska:
jenkins build this please

akva2 (approved):

bska:
PR approved and build check is green.  I'll merge into the master branch.

PR #5702, opened 2024-10-30
support matrix slitting for cpu dilu

multitalentloes:
Current version makes apply much faster, from 58, to 38ms in the apply, but the update gets more expensive (82->128ms). 
Many cases have many applies per update so this is good to provide.

Todo: how to make using matrix splitting a choice in the amg hierarchies?

### DIFF
--- a/opm/simulators/linalg/DILU.hpp
+++ b/opm/simulators/linalg/DILU.hpp
@@ -65,16 +65,16 @@ public:
     /*! \brief Constructor gets all parameters to operate the prec.
        \param A The matrix to operate on.
     */
-    MultithreadDILU(const M& A)
-        : A_(A)
+    MultithreadDILU(const M& A, bool split_matrix = true)
+        : A_(A), split_matrix_(split_matrix)
     {
         OPM_TIMEBLOCK(prec_construct);
         // TODO: rewrite so this value is set by an argument to the constructor
 #if HAVE_OPENMP
         use_multithreading = omp_get_max_threads() > 1;
 #endif
-
         if (use_multithreading) {
+            assert(!split_matrix_);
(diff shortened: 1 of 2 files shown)


opm/simulators/linalg/DILU.hpp
```diff
@@ -113,7 +134,11 @@ class MultithreadDILU : public PreconditionerWithUpdate<X, Y>
         if (use_multithreading) {
             parallelUpdate();
         } else {
-            serialUpdate();
+            if (split_matrix_) {
+                serialSplitUpdate();
```
atgeirr, line 138: The method names serial[Split]Update and parallelUpdate may be confusing since they both can be used in an MPI parallel setting, but just depend on the number of threads per process. I suggest taking the opportunity to rename them to for example singleThreaded[Split]Update and multiThreadedUpdate.

opm/simulators/linalg/DILU.hpp
```diff
@@ -65,16 +65,16 @@ class MultithreadDILU : public PreconditionerWithUpdate<X, Y>
     /*! \brief Constructor gets all parameters to operate the prec.
        \param A The matrix to operate on.
     */
-    MultithreadDILU(const M& A)
-        : A_(A)
+    MultithreadDILU(const M& A, bool split_matrix = true)
```
atgeirr, line 68: I think you can remove the default argument. Also, here it defaults to true, whereas the default is false both in the class member, and in the factory treatment, which is confusing.

atgeirr:
> Todo: how to make using matrix splitting a choice in the amg hierarchies?

Not really sure about the intricacies of this, but one hacky way could be to create a subclass DILUSplitPreconditioner that just inherits the regular DILU and sets the splitting parameter to true, but there is probably a better way.

PR #5704, opened 2024-10-31
Simplify Logic for Exchanging PI/II Values in ACTIONX/WELPI

bska:
This commit switches to using the new guarantees provided by `ActionX::wellpi_wells()` (PR OPM/opm-common#4296), namely that the well names are returned in sorted order on all ranks, to simplify how we exchange the PI/II values.  In particular, given this guarantee, we can calculate the maximum values across all ranks because zero values correspond to missing entries and non-zero values correspond to known entries.

### DIFF
--- a/opm/simulators/flow/ActionHandler.cpp
+++ b/opm/simulators/flow/ActionHandler.cpp
@@ -151,38 +151,17 @@ namespace {
             return wellpi;
         }
 
-        const auto num_wells = schedule[reportStep].well_order().size();
-
-        std::vector<Scalar> wellpi_vector(num_wells);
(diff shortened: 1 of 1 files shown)


bska:
I'm creating this PR in draft mode because it depends on the upstream work in PR OPM/opm-common#4296.  I will keep the PR in a draft state until such time as it is ready for review.

bska:
The upstream PR was merged into the master branch.  I'm marking this as "ready for review" and I'm running a  build check.

bska:

akva2 (approved):

bska:

PR #5708, opened 2024-11-04
Update flow manpage with help2man.

blattms:
Not ideal as it uses our help messages. Unfortunately, the descriptions in there are not the best.


blattms:
If there a typos than I would need to fix those in the code which might delay the release.

bska (approved):
Very good.  I'll merge into master and back-port to the release branch.

PR #5712, opened 2024-11-05
quell unused private member warning

akva2:
Clang gets chatty. As I'm not entirely sure what the plan is I have suppressed the warning using maybe_unused.
If this member is not required, we should perhaps consider removing it instead.


akva2:
jenkins build this please

bska (approved):
Cool.  I'll merge into master.

PR #5713, opened 2024-11-05
fixed: use constexpr formats and/or mark format strings runtime

akva2:
fmt wants to do compile time validation of format strings under c++-20. mark formatting strings constexpr where possible, and mark formats runtime to disable validation where not possible.


akva2:
jenkins build this please

bska (approved):

PR #5716, opened 2024-11-06
Quell c++-20 warnings

akva2:
- use constexpr instead of anonymous enums
- avoid use of std::is_pod_v
- mark comparison operator const
- add some casts to avoid enum arithmetics


akva2:
jenkins build this please

bska (approved):
This is fine and I'll merge into master although I'm a little surprised you did not see similar issues in the [initialisation](https://github.com/OPM/opm-simulators/blob/15b12eaa3e140f57a95b1c0fcf0dd7c0fb6d5091/opm/simulators/flow/RegionPhasePVAverage.cpp#L87) of `RegionPhasePoreVolAverage::x_`.

PR #5718, opened 2024-11-06
Don't Pass Action::Result By Value Into Add_Run()

bska:
Following OPM/opm-common#4322 this is no longer needed so we can go back to using `const` objects.


bska:
Although not **strictly** necessary, I'm choosing to mark this as a draft pending OPM/opm-common#4322.  I'll keep the PR in a draft state until such time as the upstream PR is merged.

bska:
jenkins build this opm-common=4322 please

bska:
The upstream PR was merged into the master branch.  I'm marking this PR as "ready for review" and I'm running a build check.

bska:
jenkins build this please

akva2 (approved):

PR #5719, opened 2024-11-07
fixed: building pypi release packages

akva2:
Of course you only notice things when you try to use them.

Found some small issues while deploying the new release packages to pypi.
Patched it while building but I rather not have to do the dance again in 6 months..


python/generate-pypi-package.sh
bska, line 1: There not being a default value for BUILD_JOBS when VERSION_TAG and VERSION both have default values looks a little suspicious to me.  Maybe it's better to reorder the parameters as
```
BUILD_JOBS VERSION_TAG VERSION
```
instead?

bska (commented):
Looks good for the most part, but I think I saw a **potential** issue in the package generation script.

python/generate-pypi-package.sh
akva2: yeah i was minimizing changes while patching but i will reorder. i'll also add a default value for BUILD_JOBS, they are usually defaulted from the docker file so the defaults are for manual running but it should be consistent.

bska (approved):

PR #5720, opened 2024-11-08
Fix index bug in gpubridge separate wells

razvnane:
Gpubridge separate wells implementation requires the global index. Fix needed after the update to using local indexes introduced in PR #5341


bska:
@jakobtorben : Would you mind having a look at this PR, please?

jakobtorben:
That should be correct! Sorry for not catching this when I updated the well equations, I had not activated the gpubridge when testing and assumed that all the tests passing was enough.

bska:
> That should be correct! Sorry for not catching this when I updated the well equations

Very good.  Thanks a lot for the quick response.  I'll run a build check and merge if green.

bska:
jenkins build this please

bska:
> I'll run a build check and merge if green.

Build check is green.  I'll merge into master.  Thanks a lot for the fix, @razvnane!

PR #5721, opened 2024-11-08
Removed GPL copyright header from licensed under Boost license.

blattms:
One cannot just add another license when there are no substantial creative changes and you are changing existing code only.

There is no need either as we can use code under Boost license next to GPL.


blattms:
jenkins build this please

alfbr:
The standard copyright header was most probably added without your valid point considered. Will merge when green unless somebody has new information.

PR #5722, opened 2024-11-08
Added the missing plus in gplv3+ for the Python packages.

blattms:


akva2 (approved):
Yup. Spec says there should be a +

PR #5723, opened 2024-11-08
Correctly mark data files with ODBL and DBCL license

blattms:
tests/wells_no_perforation.data and tests/msw.data were falsely marked as GPLv3, the rest was missing lincense marks.


blattms:
jenkins build this please

alfbr:
This is a welcome clean-up! It will go in, but we really should add a copyright notice too. Equinor is maybe correct, but maybe @atgeirr can help us clarify?

alfbr:
Jenkins came back green, and no help on the copyright. In it goes.

atgeirr:
A bit late to the party, but I agree that these are the correct licenses. As to copyright, we must check the git history, I do not think these all have Equinor copyright, although most probably do.

PR #5724, opened 2024-11-11
Changes in how the simulator handles NUPCOL

totto82:
1) Dont allow wells to change to GRUP when iter >NUPCOL
2) Update VREP even for iter > NUPCOL if it changes significantly. For now hardcoded to relative change > 0.01 (TODO make this a parameter. Update: DONE)

TODO

- [x] If this is a good idea for VREP it should be done for REIN etc.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1179,7 +1179,8 @@ groupAndNetworkData(const int reportStepIdx) const
 template<class Scalar>
 void BlackoilWellModelGeneric<Scalar>::
 updateAndCommunicateGroupData(const int reportStepIdx,
-                              const int iterationIdx)
+                              const int iterationIdx,
+                              DeferredLogger& deferred_logger)
 {
     const Group& fieldGroup = schedule().getGroup("FIELD", reportStepIdx);
     const int nupcol = schedule()[reportStepIdx].nupcol();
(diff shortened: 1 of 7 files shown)


totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:

steink:
I'm currently looking into this, but having some unrelated issues with recent updates, so not done testing

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2197,31 +2199,34 @@ namespace Opm {
         const int nupcol = this->schedule()[episodeIdx].nupcol();
         // don't switch group control when iterationIdx > nupcol
         // to avoid oscilations between group controls
+        // We also dont allow wells to change to group controls
+        // when iterationIdx > nupcol
```
steink, line 2203: Should the opposite also apply, e.g., that wells under group control should not be allowed to change to individual in the subsequent `updateWellControl` (individual mode)?

totto82: This change indeed reduce oscillations and improve convergence, but it has the potential to give very wrong results. I think we rather should remove the if statement completely i.e. always check the constrains. This will need some adjustment in the code to handle oscillatory controls for instance when a higher group has a constraint that is equal to the sum of the sub groups controls. I am testing out an option of detecting oscillatory controls like we do for the wells. This seems to work but need more testing. As a first move I plan to remove this part of this PR to be able to evaluate and test these things separately.

totto82:

opm/simulators/wells/BlackoilWellModel_impl.hpp
steink: OK, sounds good.

totto82:

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1188,8 +1190,50 @@ updateAndCommunicateGroupData(const int reportStepIdx,
...
+                                                    /*isInjector*/ false);
+                        }
+                        if ( std::abs( (gr_rate_nupcol - gr_rate) / (0.5*gr_rate_nupcol + 0.5*gr_rate)) > tol_nupcol) {
+                            this->updateNupcolWGState();
+                            const std::string control_str = is_vrep? "VREP" : "REIN";
+                            const std::string msg = fmt::format("Group prodution relative change larger than tolerance {} "
+                                                    "at iteration {}. Update {} for Group {} even if iteration is larger than {} given by NUPCOL." ,
+                                                    tol_nupcol, iterationIdx, control_str, gr_name, nupcol);
+                            deferred_logger.info(msg);
+                        }
```
totto82, line 1232: This should be changes to .debug before merging.

totto82:

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1188,8 +1190,51 @@ updateAndCommunicateGroupData(const int reportStepIdx,
...
+                        for (int phaseIdx = 0; phaseIdx < np; ++phaseIdx) {
+                            gr_rate += WellGroupHelpers<Scalar>::sumWellPhaseRates(is_vrep,
+                                                    group,
+                                                    schedule(),
+                                                    this->wellState(),
+                                                    reportStepIdx,
+                                                    phaseIdx,
+                                                    /*isInjector*/ false);
+                        }
+                        Scalar rel_change = std::abs( (gr_rate_nupcol - gr_rate) / (0.5*gr_rate_nupcol + 0.5*gr_rate));
```
steink, line 1225: Is there a possibility both group rates in denominator could be zero here? Or would that situation lead to failure at an earlier stage?

steink:
Larger testing looks fine on my side. Did some investigation on `02_WGRUPCON` which gets quite a bit worse with the PR. Main reason is that with master we get `Well C-1H will be shut because it fails to converge`, while this does not trigger with the PR which rather does extra `VREP`-iterations until convergence failure. Reason for C-1H not converging in the first place is that there exists no converged state it will rest (production with its thp is not feasible (so it should stop), but when global system converges with well stopped, reservoir pressure supports re-opening). Since our operability checks only considers _current_ reservoir pressure and we don't have any limits on how many times a well can try to re-open during a time-step, this well will always cause issues manifesting in one or the other way. The issue is not related to this PR however, and we should address it elsewhere.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
totto82: It is possible so I will add a check. Maybe we should have a "large epsilon" here since relative changes does not make much sense for very small numbers.

totto82:
> re-open during a time-step

Does it make sense to not allow re-opening wells when iter > nupcol? This could reduce oscillations in the VREP control, but also to premature closing of wells. Often the closing/opening of individual wells stabilize after some iterations so maybe worth a try to test out?

totto82:
When I only allow for well opening when iter < nupcol I get back the reference solution. I will do some more tests with this and report back. 

image

steink:
> > re-open during a time-step
> 
> Does it make sense to not allow re-opening wells when iter > nupcol? This could reduce oscillations in the VREP control, but also to premature closing of wells. Often the closing/opening of individual wells stabilize after some iterations so maybe worth a try to test out?

I agree that it should not be allowed to re-open at some stage, but not sure whether this should be related to nupcol (the problem is not neccessarlily related to groups).

totto82:
> I agree that it should not be allowed to re-open at some stage, but not sure whether this should be related to nupcol (the problem is not neccessarlily related to groups).

I see your point. It seems to stabilize some corner cases thought. Does this mean that we accept the results of well C-1H in 02_WGRUPCON and can start the process of getting this PR in?

steink:
> > I agree that it should not be allowed to re-open at some stage, but not sure whether this should be related to nupcol (the problem is not neccessarlily related to groups).
> 
> I see your point. It seems to stabilize some corner cases thought. Does this mean that we accept the results of well C-1H in 02_WGRUPCON and can start the process of getting this PR in?

It's OK from my side.

totto82:

totto82:
jenkins build this opm-tests=1259 please

PR #5725, opened 2024-11-11
fixed: RESTART_SCHED parameter never took effect

akva2:
in particular if(string_var) evaluates to false if it has value "false"


akva2:
jenkins build this please

bska:

PR #5727, opened 2024-11-11
Janitoring

akva2:
- add virtual dtor to class with virtual members
- remove unused variable


akva2:
jenkins build this please

bska:
@multitalentloes : Would you mind having a look at this, please?  In particular whether or not the `size` variable is genuinely unused or if it is supposed to be used elsewhere.

multitalentloes:
I dont see how this variable could have any effect. Looks like dead code which is safe to remove

bska:
>  Looks like dead code which is safe to remove

Okay-cool.  Thanks for looking into this.  I'll merge into master.

PR #5731, opened 2024-11-12
ValidationFunctions: some cleanup

akva2:
- introduce a type alias for the std::function type
- make implementation of specialValidations() private
- make validateBRINE private


akva2:
jenkins build this please

PR #5735, opened 2024-11-13
dont shut well with GRUP control that struggles to converge

totto82:
if one well on individual control (typical thp/bhp) in a group struggles to converge
it may lead to problems for the other wells in the group
we dont want to shut all the wells in a group only the one creating the problems.

### DIFF
--- a/opm/simulators/timestepping/AdaptiveTimeStepping.hpp
+++ b/opm/simulators/timestepping/AdaptiveTimeStepping.hpp
@@ -426,24 +426,34 @@ void registerAdaptiveParameters();
                             // Found no wells to close, chop the timestep as above.
                             chopTimestep();
                         } else {
-                            // Close all consistently failing wells.
-                            int num_shut_wells = 0;
+                            // Close all consistently failing wells that are not under group control
+                            std::vector<std::string> shut_wells;
                             for (const auto& well : failing_wells) {
-                                bool was_shut = solver.model().wellModel().forceShutWellByName(well, substepTimer.simulationTimeElapsed());
+                                bool was_shut = solver.model().wellModel().forceShutWellByName(
+                                                            well, substepTimer.simulationTimeElapsed(), /*dont_shut_grup_wells =*/ true);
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

totto82:

totto82:
jenkins build this please

opm/simulators/timestepping/AdaptiveTimeStepping.hpp
```diff
@@ -426,24 +426,37 @@ void registerAdaptiveParameters();
                             // Found no wells to close, chop the timestep as above.
                             chopTimestep();
                         } else {
-                            // Close all consistently failing wells.
+                            // Close all consistently failing wells that are not under group control
                             int num_shut_wells = 0;
+                            std::set<std::string> shut_wells;
```
GitPaean, line 431: since you only need `insert` and  loop over it, `std::vector` will be a better choice.

opm/simulators/timestepping/AdaptiveTimeStepping.hpp
```diff
@@ -426,24 +426,37 @@ void registerAdaptiveParameters();
                             // Found no wells to close, chop the timestep as above.
                             chopTimestep();
                         } else {
-                            // Close all consistently failing wells.
+                            // Close all consistently failing wells that are not under group control
                             int num_shut_wells = 0;
```
GitPaean, line 430: If I did not read it wrong, since you have a `shut_wells` keep tracking of things, `num_shut_wells` is not needed anymore.

GitPaean (approved):
I do have the concern that we are introducing more and more complicated logic. But this logic do look trying to keep more wells alive. I go with what testing indicates.

I only have two small comments regarding the code.

GitPaean:
jenkins build this please

totto82:
jenkins build this please

GitPaean:
Thanks for the update. Jenkins approves and I am getting the PR in now.

PR #5736, opened 2024-11-13
Chase API Update of Action::Result

bska:
Following OPM/opm-common#4340, the `Action::Result` type no longer uses operator overloading and the matching entity back-end has been revised to no longer expose a vector of strings for the well names directly.


akva2 (approved):

bska:
PR approved and build check triggered from upstream PR is green.  I'll merge into master.

PR #5737, opened 2024-11-13
Network sub iterations

totto82:
The network update is damped by a factor 0.1. This add sub-iterations to avoid updating gaslift and groupcontrols etc at every network update. 

@GitPaean  Why is the damping factor included? 0.1 seems very conservative. Wouldn't the update become smaller and smaller?


GitPaean:
> @GitPaean Why is the damping factor included? 0.1 seems very conservative. Wouldn't the update become smaller and smaller?

It was probably introduced to tackle a specific case back then, while it is possible the design is not always good and remains to be adjusted.

GitPaean:

totto82:
> > @GitPaean Why is the damping factor included? 0.1 seems very conservative. Wouldn't the update become smaller and smaller?
> 
> It was probably introduced to tackle a specific case back then, while it is possible the design is not always good and remains to be adjusted.

I think the current design is a bit unfortunate as the update is limited also when dp is small. I am also testing using no damping factor. Do you remember  the specific case? Could you share it with me?

GitPaean:
> I think the current design is a bit unfortunate as the update is limited also when dp is small. I am also testing using no damping factor. Do you remember the specific case? Could you share it with me?

It was long time ago and it will be difficult and helpful to check the old cases. and I think you can stick with the testing casein the opm-tests and the field case. 

> I think the current design is a bit unfortunate as the update is limited also when dp is small

There might be something to play here.


It looks your PR working well with the testing cases in opm-tests, while I have not looked at the details of the PR.

totto82:

totto82:
This gives a significant speed up on cases with gaslift and network. I have checked the test failures and they are all minor.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2184,12 +2184,18 @@ namespace Opm {
...
             // Calculate common THP for subsea manifold well group (item 3 of NODEPROP set to YES)
             computeWellGroupThp(dt, deferred_logger);
-            const auto local_network_imbalance = this->updateNetworkPressures(episodeIdx);
-            const Scalar network_imbalance = comm.max(local_network_imbalance);
-            const auto& balance = this->schedule()[episodeIdx].network_balance();
-            constexpr Scalar relaxation_factor = 10.0;
-            const Scalar tolerance = relax_network_tolerance ? relaxation_factor * balance.pressure_tolerance() : balance.pressure_tolerance();
-            more_network_update = this->networkActive() && network_imbalance > tolerance;
+            const int max_number_of_sub_iterations = 20;
+            const Scalar damping_factor = 0.1;
```
GitPaean, line 2188: both line can be `constexpr`.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2184,12 +2184,18 @@ namespace Opm {
             const double dt = this->simulator_.timeStepSize();
             // Calculate common THP for subsea manifold well group (item 3 of NODEPROP set to YES)
             computeWellGroupThp(dt, deferred_logger);
```
GitPaean, line 2186: this should be in the sub iteration loop.

totto82: Are you sure about that? I think it should be outside. This is potentially expensive to evaluate. I have done some testing with auto choke and this seems to speed up the calculations without affecting the results significantly. I may change my mind as this with further testing and as the auto choke support advances. Any comments @plgbrts?

opm/simulators/wells/BlackoilWellModel_impl.hpp
totto82: Ok, will update.

totto82:

opm/simulators/wells/BlackoilWellModel_impl.hpp
GitPaean: In theory, probably, since other node pressure change (upstream nodes above the auto-choke node) might cause the manifold group can not produce the target rate, but for that case, we do not need to go through the expensive process to adjust.  

For now, we can go with your testing.

totto82: I tested with https://github.com/OPM/opm-simulators/pull/5754 and needed to do some changes to get similar results. I added a check if the group_well_thp was updated. With this change the results looks similar, but the run time is reduced by a factor 3. 

image

totto82:

plgbrts:
I did some further testing. For the case attached to https://github.com/OPM/opm-simulators/pull/5754 the match is good:
image
image
The simulation improvement is about a factor or 5(!)
The simulation using PR5737 however  failed to converge around 2020-04 (simulation aborted due to excessive time chopping), Setting `--tolerance-wells=1E-9` the simulation could complete the full simulation period. This tolerance is used for both simulations for a fair performance comparison.

Another case also with an autochoke group but without group guide rate. (There was  no need for tightening the well tolerance for this case)

image
In this case PR5737 shows smoother behavior and wells are kept open until the end.
image

The simulation improvement is for this case a factor of 3.

totto82:
Thanks for the testing @plgbrts 
From my point this is ready to go in. Any further comments @GitPaean?

GitPaean:
> Thanks for the testing @plgbrts From my point this is ready to go in. Any further comments @GitPaean?

I go with the testing. It looks fine to me.

GitPaean:

GitPaean:

GitPaean:
The compilation error should be easy to fix. typo or similar.

totto82:
I removed the last commit to another PR since it depends on and its only relevant for autochoke (PR #5766)

PR #5733, opened 2024-11-12
Feature/ms wells part1

lisajulia:

PR #5680, opened 2024-10-16
Feature/ms wells

lisajulia:
Make it possible for an MSWell to be split across multiple processes when running a parallel simulation.

PR #5738, opened 2024-11-13
Move assemblePerforationEq to its own loop at the beginning

lisajulia:


lisajulia:
jenkins build this please

GitPaean:

GitPaean:
Other cases look fine, only https://ci.opm-project.org/job/opm-simulators-PR-builder/6918/artifact/mpi/build-opm-simulators/failure_report/03_9_3d_grpctl_msw_model2.pdf  shows some deviation at the later stage.

bska:
The PR description doesn't say, so I'll just ask: What goal are you trying to achieve by doing this?

lisajulia:
> The PR description doesn't say, so I'll just ask: What goal are you trying to achieve by doing this?

Sorry for not including that here, I wrote that in an email to Kai and Halvor since we had discussions about the  MSWells before:

I found (at least some of) the parts of the assembly after which communication is needed, i.e. after calling the function assemblePerforationEq (https://github.com/OPM/opm-simulators/blob/3188ac713a59105e240d84d8d5ca1367d451dcb1/opm/simulators/wells/MultisegmentWell_impl.hpp#L1908), yet this function is called together with other assembly functions inside a loop over the segments (https://github.com/OPM/opm-simulators/blob/3188ac713a59105e240d84d8d5ca1367d451dcb1/opm/simulators/wells/MultisegmentWell_impl.hpp#L1820).
I'd like to isolate the function in its own loop, communicate afterwards and then add all other contributions that do not need communication, like this: https://github.com/OPM/opm-simulators/pull/5733/commits/5fe80888073ddd45df599bc0a31a2db1c855f5f3.
Yet, this is causing errors, acutally just moving the assemblePerforationEq function to its own loop is causing errors, as can be seen in this PR.


I already got the answer that the failures here are likely due to the rounding off errors from the changing order of floating-point calculation.

GitPaean:
I think it is good  to use draft mode for this type of PRs?

lisajulia:
@GitPaean yes thanks :)

Yet, how do we proceed with this?
How about the small deviation of the case https://ci.opm-project.org/job/opm-simulators-PR-builder/6918/artifact/mpi/build-opm-simulators/failure_report/03_9_3d_grpctl_msw_model2.pdf, as you mentioned in https://github.com/OPM/opm-simulators/pull/5738#issuecomment-2475812339?

If we agree that we merge this PR: Should it be merged right away or should it be merged as soon as a certain set of functionality using this change for MSWells in parallel is ready? If so, how much should be ready? I'm thinking about the test file that I'm using now with one MSWell with one branch running through correctly.

GitPaean:
We can update the reference if we think the deviation is fine/acceptable. 

If you think the PR itself serves some good purpose, we can get it in independently.  For example, you are organizing this code so that it is easier or more clearer to the parallel implementation, you can have a PR to re-organize this part of the code with clear description of the purpose.

lisajulia:
> We can update the reference if we think the deviation is fine/acceptable.
> 
> If you think the PR itself serves some good purpose, we can get it in independently. For example, you are organizing this code so that it is easier or more clearer to the parallel implementation, you can have a PR to re-organize this part of the code with clear description of the purpose.

Ok :) I‘ll get back to you with regards to this then!

lisajulia:
jenkins built this opm-tests=1249 failure_report please

GitPaean:
because of the reference is still the old reference corresponding to the old DATA file, so you will see the regression failures.

OPM/opm-tests#1249

GitPaean:

lisajulia:
Closing this since #5680 has been merged.

PR #5739, opened 2024-11-14
Support GCONSUMP on multiple levels in the group hierarchy

vkip:


vkip:
jenkins build this please

alfbr:
This one is needed for new rpms, any chance we can get it in?

opm/simulators/wells/GroupState.cpp
```diff
@@ -419,6 +423,63 @@ has_gpmaint_target(const std::string& gname) const
...
+                rates.second += static_cast<Scalar>(it->second.second * parent_gefac);
+                return true;
+            }
+
+            // Accumulate from sub-groups and keep track of any updates in 'has_values'
+            bool has_values = false;
+            if (sched_state.groups.has(group_name)) {
+                for (const auto& child_gname : sched_state.groups(group_name).groups()) {
+                    const auto gefac = sched_state.groups(child_gname).getGroupEfficiencyFactor();
+                    has_values = has_values || self(self, child_gname, rates, gefac);
```
bska, line 451: Potential issue here: If `has_values` is true, e.g., from the first child, then we won't call `self()` for subsequent children because of short-circuiting.  Is that intentional?  If so, please add a comment for why that's the intended semantics.  If not, maybe we could rewrite it as
```C++
const auto child_has_values = self(...);
has_values = has_values || child_has_values;
```

opm/simulators/wells/GroupState.cpp
```diff
@@ -419,6 +423,63 @@ has_gpmaint_target(const std::string& gname) const
...
+            // Add consumption/import rates at current level
+            if (gconsump.has(group_name)) {
+                const auto& group_gc = gconsump.get(group_name, summary_state);
+                rates.first += static_cast<Scalar>(group_gc.consumption_rate);
+                rates.second += static_cast<Scalar>(group_gc.import_rate);
+                has_values = true;
+            }
+
+            // Update map if values are set
+            if (has_values) this->m_gconsump_rates[group_name] = rates;
```
bska, line 464: Micro-optimisation: It's (slightly) more efficient to use [`map<>::insert_or_assign()`](https://en.cppreference.com/w/cpp/container/map/insert_or_assign) instead of `map[] =` here.  I.e., as
```C++
this->m_gconsump_rates.insert_or_assign(group_name, rates);
```
instead of the assignment.

opm/simulators/wells/GroupState.cpp
```diff
@@ -419,6 +423,63 @@ has_gpmaint_target(const std::string& gname) const
...
+            return has_values;
+        };
+
+    std::pair<Scalar, Scalar> rates = {0.0, 0.0};
```
bska, line 468: This will generate a type conversion if `Scalar` is different from `double`.  I think I'd personally use class template argument deduction and write it as
```C++
auto rates = std::pair { Scalar{0}, Scalar{0} };
```

bska (commented):
This looks good for the most part.  I think it'd be possible to avoid the recursive lambda by slightly rewriting the `update_gconsump()` function if we wanted to, but the current version is fine too.

That said, I think there's at least a **potential** issue in the current implementation, so please have a look at the way we invoke the lambda recursively.

opm/simulators/wells/GroupState.cpp
vkip: Oops, you're right, that's a real issue.. thanks! Sufficient to change the order though, right? (I.e., `has_values = self(self, child_gname, rates, gefac) || has_values; `)

vkip:
jenkins build this please

bska (approved):
Thanks a lot for the update&ndash;this looks good to me now and as the build check is green I'll merge into master.  At some point in the future we may want to extract the recursive lambda out to a named helper function to simplify the recursive calls, but for now this is a considerable improvement in our capability.

PR #5740, opened 2024-11-16
Avoid potential segfault due to invalid vector.end() after erase()

vkip:


atgeirr (approved):
Erasing in loops is just too hard to get right, thanks for the fix!

atgeirr:
jenkins build this please

PR #5745, opened 2024-11-19
Cmake: Use upper case for COMPONENTS of FindSuiteSparse

blattms:
This is what DUNE uses since 2015 makes OPM modules usable as DUNE modules in all cases.

Downstream of OPM/opm-common#4349


blattms:
Maybe I should make this backwards compatible to prevent full rebuilds from being needed first.

blattms:
jenkins build this please

blattms:
Assuming approval of upstream PR extends to this one I am doing a shameless selfmerge

PR #5748, opened 2024-11-19
Only update target of group-controlled wells during updateAndCommunicate

steink:
The `updateWellStateWithTarget` for all wells can be problematic here. In particular, this is the case for thp-wells, since in this case `updateWellStateWithTarget` only provides an _initial guess_ well-state. This means that at this stage, we may have converged thp-wells that after going through `updateWellStateWithTarget` are no longer convereged, which again may prevent overall convergence. 

I can't see that `updateAndCommunicateGroupData` changes any well control-modes, so there should be no reason to update the target of wells not under group control here?

This PR just skips the `updateWellStateWithTarget` for wells not under group control


steink:

totto82:

totto82:
jenkins build this opm-tests=1255 please

PR #5753, opened 2024-11-20
Alternative way of avoiding group control oscillations

totto82:
Allow for checking group constrains also when iter > nupcol. To avoid oscillations we instead keep track of the number group control changes.


totto82:

totto82:

totto82:
Snap from GGPR for group PROD. Its maximum is 1.6e6, which the reference does violate in one time step. 

image

totto82:

totto82:

totto82:

totto82:

akva2:
do not be confused, the update in opm-tests have not yet landed.....

totto82:
> do not be confused, the update in opm-tests has not yet landed.
Thanks. I will rerun it after some time.

akva2:
it's there now. there is a 5 min window..

steink:
Sorry for beeing a bit late to comment here. I think the ability to _count_ group switching to detect oscillations is very useful. I'm a bit uncertain about the following items though (I might very well have misunderstood something here, so please correct me if I'm wrong)
- so you're basically suggesting to ignore `NUPCOL` since now you always do `updateGroupControls` in `updateWellControls`. I'm not sure I have the complete picture of the implications for this choice. There are still other places in the code where `NUPCOL` is respected (e.g., `updateAndCommunicateGroupData`), is there any chance this may lead to issues?
- there's a possible issue when counting number of switches (both for wells and groups) if we at the start of a step need do many iterations in `updateWellControlsAndNetworkIteration`. If a lot of switching happens there, I think we may end up _freezing_ controls before doing a single global update. Could it be an option to reset `switched_prod_groups_`   (and `well_control_log_`) e.g., after `doPreStepNetworkRebalance`?

totto82:
> so you're basically suggesting to ignore NUPCOL since now you always do updateGroupControls in updateWellControls. I'm not sure I have the complete picture of the implications for this choice. There are still other places in the code where NUPCOL is respected (e.g., updateAndCommunicateGroupData), is there any chance this may lead to issues?

As I see it the purpose of NUPCOL is to freeze the well/group targets after NUPCOL iterations. Using NUPCOL in updateAndCommunicateGroupData has that effect. The NUPCOL in updateGroupControls was added to avoid oscillations between group controls.  I dont think that is the correct usage of NUPCOL. It potentially leads to significantly wrong results. That's why I instead propose to count the number of switches. 

> there's a possible issue when counting number of switches (both for wells and groups) if we at the start of a step need do many iterations in updateWellControlsAndNetworkIteration. If a lot of switching happens there, I think we may end up freezing controls before doing a single global update. Could it be an option to reset switched_prod_groups_ (and well_control_log_) e.g., after doPreStepNetworkRebalance?

Good point. I will try to update accordingly.

steink:
> As I see it the purpose of NUPCOL is to freeze the well/group targets after NUPCOL iterations. Using NUPCOL in updateAndCommunicateGroupData has that effect. The NUPCOL in updateGroupControls was added to avoid oscillations between group controls. I dont think that is the correct usage of NUPCOL. It potentially leads to significantly wrong results. That's why I instead propose to count the number of switches.

Thanks for the explanation. Then this makes perfect sense to me

totto82:

totto82:
> there's a possible issue when counting number of switches (both for wells and groups) if we at the start of a step need do many iterations in updateWellControlsAndNetworkIteration. If a lot of switching happens there, I think we may end up freezing controls before doing a single global update. Could it be an option to reset switched_prod_groups_ (and well_control_log_) e.g., after doPreStepNetworkRebalance?

I went for a slightly different approach. Now the number switched controls is only checked when iter>nupol i.e. multiple switching are allowed while the network is iterating. I think I will test something similar for the wells. Keeping the controls to avoid oscillation is something we would like to avoid as much as possible.

totto82:
I interpret @steink thumbs up as approval. @bska or @atgeirr Do you want to check the c++ part? If not I will self-merge later this week.

bska:
> Do you want to check the c++ part?

Yes.  I'll do a more detailed review later&ndash;probably Thursday afternoon&ndash;but there are a few things here that could be improved.  In particular we should try to avoid using `map<pair<>,vector<>>` if at all possible.  I'll make a suggestion in my next update.

opm/simulators/wells/BlackoilWellModelConstraints.cpp
```diff
@@ -561,8 +561,9 @@ template<class Scalar>
 bool BlackoilWellModelConstraints<Scalar>::
 updateGroupIndividualControl(const Group& group,
                              const int reportStepIdx,
-                             std::map<std::pair<std::string,Phase>,std::string>& switched_inj,
-                             std::map<std::string, std::string>& switched_prod,
+                             const int max_number_of_group_switch,
+                             std::map<std::pair<std::string,Phase>,std::vector<std::string>>& switched_inj,
```
bska, line 565: `Map<pair, vector>` really isn't a good data structure for anything.  I suppose we'd create a separate type to handle some of the logic here, but barring that I'd probably either pull the `Phase` out of the pair and make it a `vector<map<string, vector>>`, or pull the group name out and make it a `map<string, vector<vector>>`.  The former would  be indexed, at least conceptually, as `switched_inj[phase][group]` while the latter would be indexed as `switched_inj[group][phase]`.  Given the current loop structure I think the latter may be preferable.

On a second note, since we now **know** that we're dealing with injection controls we should store the integer [`Group::InjectionCMode`](https://github.com/OPM/opm-common/blob/268f5b8b3e5cb89a7d304f4d31bc6fdf4dd96dc9/opm/input/eclipse/Schedule/Group/Group.hpp#L70) values instead of their string representation.  That way we use less memory and the comparisons are more efficient as well.

If we use the group name as the primary key and the `Phase` as the secondary key I'd probably write this block along the lines of
```C++
bool group_is_oscillating = false;
if (auto groupPos = switched_inj.find(group.name()); groupPos != switched_inj.end()) {
    auto& ctrls = groupPos->second[static_cast<std::underlying_type_t<Phase>>(phase)];
    for (const auto& ctrl : ctrls) {
        if (std::count(ctrls.begin(), ctrls.end(), ctrl) < max_number_of_group_switches) {
            continue;
        }

        if (ctrls.back() != *(ctrls.end() - 2)) {
            if (rank == 0) {
                // report, use InjectionCMode2String() for string value of 'ctrl'
            }
            ctrls.push_back(ctrl);
        }
        group_is_oscillating = true;
        break;
    }
}
```

bska (commented):
There is repeated logic here which suggests we should try to extract it to a helper class.  Maybe in follow-up work.  That said, even if we don't do that right now, I do think we should switch to a more linear data structure for simpler processing and we absolutely should reduce the number of calls to `map<>::operator[]()` since that has quite a bit of hidden cost.

totto82:

totto82:

totto82:
I have looked at the test failures. UDQ_WCONPROD shows the largest deviation. Current master allow for a solution that violate WOPR target (1500). With PR the results are ok. 

image

totto82:
From my point this is ready to be merged after an update of data. @bska Are you happy with the changes? I more or less followed your suggestions.

bska:
> Are you happy with the changes?

I'll have a look now.

bska (approved):
>> Are you happy with the changes?
>
> I'll have a look now.

Thanks a lot for the updates.  This looks a lot better now.  There are a couple of further optimisations that we may do here, but that's suitable for follow-up work, I think.  For now this is ready for merging from a C++ point of view.

totto82:

totto82:
Thanks for the reviews.

PR #5755, opened 2024-11-22
fixed: add missing include

akva2:
broke sequential build


akva2:
shameless self-merge to restore CI.

PR #5756, opened 2024-11-22
Make Floating Point Existence Flag Initially Unset

bska:
PR OPM/opm-common#4348 altered the OPM build system's handling of boolean flags and setting the flag to zero no longer works.


bska:
jenkins build this please

akva2:
in she goes

PR #5757, opened 2024-11-22
Well testing using gas lift facilities

svenn-t:
The current master version of well test (WTEST keyword) with gas lift typically resulted in max ALQ given to well after re-opening. With this PR we use gas lift facilities to increase ALQ until max. incremental gradient instead. Running the gas lift test cases in [opm-tests](https://github.com/OPM/opm-tests/tree/master/gaslift) show that large ALQ spikes after re-opening well(s) are avoided with the PR compared to master.

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -1041,17 +1041,19 @@ namespace Opm
                                                         deferred_logger);
         auto [max_alq, success] = glift->wellTestALQ();
         std::string msg;
+        const auto& unit_system = schedule.getUnits();
         if (success) {
             well_state.setALQ(well_name, max_alq);
             msg = fmt::format(
                 "GLIFT WTEST: Well {} : Setting ALQ to optimized value = {}",
-                well_name, max_alq);
+                well_name, unit_system.from_si(UnitSystem::measure::gas_surface_rate, max_alq));
         }
         else {
             if (!gl_well.use_glo()) {
                 msg = fmt::format(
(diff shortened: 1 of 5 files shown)


totto82:
jenkins build this please

svenn-t:
Hopefully the update version will compile with jenkins now. At least it compiles on my computer.

akva2:
it will. to build for float locally pass ```-DBUILD_FLOW_FLOAT_VARIANTS=1``` cmake param.

akva2:
jenkins build this please

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1016,28 +1024,44 @@ namespace Opm
...
+        // Use gas lift optimization to get ALQ for well test
+        std::unique_ptr<GasLiftSingleWell> glift =
+            initializeGliftWellTest_<GasLiftSingleWell>(simulator,
+                                                        well_state,
+                                                        group_state,
+                                                        phase_usage,
+                                                        ecl_well_map,
+                                                        deferred_logger);
+        auto [max_alq, success] = glift->wellTestALQ();
+        std::string msg;
```
totto82, line 1043: Is this the max? Isn't this the min alq that allows the well to open?

totto82 (approved):
Thanks. The code look good. Only one minor nitpick. I have also tested it and it indeed removes some spikes in the gaslift after a well opening.

opm/simulators/wells/WellInterface_impl.hpp
svenn-t: Indeed, forgot to change the old variable name. Fixed now.

PR #5760, opened 2024-11-25
added: ability to pass command line parameters from python

### LINKED ISSUE #5759: Adding a CASENAME.PARAM when running flow in Python
Hello there!

If I run flow in Python, is it possible to add a CASENAME.PARAM file somewhere to change the simulation parameters?

akva2:
This adds the ability to pass standard command line parameters in the python simulator.

Closes #5759

### DIFF
--- a/opm/simulators/flow/python/PyBlackOilSimulator.hpp
+++ b/opm/simulators/flow/python/PyBlackOilSimulator.hpp
@@ -40,7 +40,8 @@ private:
     using Simulator = Opm::GetPropType<TypeTag, Opm::Properties::Simulator>;
 
 public:
-    PyBlackOilSimulator( const std::string& deckFilename);
+    PyBlackOilSimulator(const std::string& deckFilename,
+                        const std::vector<std::string>& args);
     PyBlackOilSimulator(
         std::shared_ptr<Opm::Deck> deck,
         std::shared_ptr<Opm::EclipseState> state,
@@ -92,6 +93,7 @@ private:
     std::shared_ptr<Opm::EclipseState> eclipse_state_;
     std::shared_ptr<Opm::Schedule> schedule_;
     std::shared_ptr<Opm::SummaryConfig> summary_config_;
+    std::vector<std::string> args_;
 };
 
 } // namespace Opm::Pybind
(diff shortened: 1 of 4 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this please

opm/simulators/flow/python/PyBlackOilSimulator.hpp
```diff
@@ -40,7 +40,8 @@ class PyBlackOilSimulator
     using Simulator = Opm::GetPropType<TypeTag, Opm::Properties::Simulator>;
 
 public:
-    PyBlackOilSimulator( const std::string& deckFilename);
+    PyBlackOilSimulator(const std::string& deckFilename,
+                        const std::vector<std::string>& args = {});
```
atgeirr, line 44: Could we drop the default argument? If I understand this correctly the only call site is line 300 in PyBlackOilSimulator.cpp, so it should be fine.

python/simulators/PyMain.hpp
```diff
@@ -34,6 +34,22 @@ class PyMain : public Main
 public:
     using Main::Main;
 
+    void setArguments(const std::vector<std::string>& args)
+    {
+        if (args.empty()) {
+            return;
+        }
+
+        this->argc_ = 2 + args.size();
```
atgeirr, line 43: I think the logic here is correct, but it would be nice with a comment explaining it. It is hard to juggle `args`, `argv_` and `argv_python_` without getting confused.

atgeirr (commented):
Looks good to me, with only minor issues. Should be merged after fixing.

opm/simulators/flow/python/PyBlackOilSimulator.hpp
akva2: strictly speaking line 300 is not a call site, it is defining the python wrapping of the constructor, the call site is in the .py files. but okay, seems pybind does indeed allow default values on python side without having the same on c++ side so i have removed the default value.

python/simulators/PyMain.hpp
akva2: tried to add an explanation.

atgeirr (approved):

PR #5761, opened 2024-11-25
Move debug output from thp calcuations from .PRT to .DBG

totto82:
Especially fro cases with gaslift the .PRT file is flooded with THP related output. I think it make sense to move this to the debug file since is this is mostly of interest when debugging.


totto82:
jenkins build this please

bska (approved):
Yep, I think this makes a lot of sense.  Thanks a lot for tidying up this aspect of our diagnostic output.  I'll merge into master.

PR #5743, opened 2024-11-18
Restore threshold pressures from restart

vkip:

PR #5770, opened 2024-11-28
Write correct THRESHPR values also from restart

vkip:
Managed to leave this out of https://github.com/OPM/opm-simulators/pull/5743, sorry about that.

### DIFF
--- a/opm/simulators/flow/GenericThresholdPressure_impl.hpp
+++ b/opm/simulators/flow/GenericThresholdPressure_impl.hpp
@@ -227,6 +227,12 @@ getRestartVector() const
         return {};
 
     std::vector<Scalar> result(numEquilRegions_ * numEquilRegions_, 0.0);
+
+    if (this->restart_) {
+        result = this->thpres_;
+        return result;
+    }
+
     const auto& simConfig = eclState_.getSimulationConfig();
     const auto& thpres = simConfig.getThresholdPressure();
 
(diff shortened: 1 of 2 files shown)


vkip:
jenkins build this please

opm/simulators/flow/GenericThresholdPressure_impl.hpp
```diff
@@ -227,6 +227,12 @@ getRestartVector() const
         return {};
 
     std::vector<Scalar> result(numEquilRegions_ * numEquilRegions_, 0.0);
+
+    if (this->restart_) {
```
atgeirr, line 231: This certainly looks correct to me, even to the point where always returning `this->thpres_ ` should be correct, also for non-restart cases? After all, that array is initialized by code identical to the block below?

vkip: That makes sense! Updating..

vkip:
jenkins build this please

vkip:
jenkins build this please

opm/simulators/flow/GenericThresholdPressure_impl.hpp
vkip: Needed a small fix for parallel without restart, but seems to work.

atgeirr:
Thanks, merging.

PR #5772, opened 2024-11-29
Move addLgrs' call to support refinement on a distributed grid

aritorto:
Moving the call of addLgrsUpdateLeafView, to allow adding LGRs on a [level zero] distributed CpGrid. 

This PR does not affect the serial runs and paves the way to support the parallel ones. 
 

aritorto:

PR #5773, opened 2024-11-29
fixed: well model could deadlock

akva2:
if umfpack encounters a singular matrix on any process, the exception propagated up across communication boundaries leading to deadlock as the other processes tried to do global reductions.


akva2:
jenkins build this please

bska:
I wonder how many other such problems lie dormant in our code base...

atgeirr:
> I wonder how many other such problems lie dormant in our code base...

One less now! I think they are most likely to be located in relatively new parts of the code though, as we went through this very carefully when we added the DeferredLogger mechanism.

PR #5779, opened 2024-12-03
adjust output and parameters in gpu autotuner

multitalentloes:
This PR refines the autotuning output a bit.
Additionally it changes the thread-block size used for the lower solve when tuning the upper triangular solve. That fixes confusing output that could make it look like the tuning was making the triangular solve slower when it wasn't.

### DIFF
--- a/opm/simulators/linalg/gpuistl/GpuDILU.cpp
+++ b/opm/simulators/linalg/gpuistl/GpuDILU.cpp
@@ -270,12 +270,12 @@ GpuDILU<M, X, Y, l>::tuneThreadBlockSizes()
     auto tuneMoveThreadBlockSizeInUpdate = [this](int moveThreadBlockSize){
         this->update(moveThreadBlockSize, m_DILUFactorizationThreadBlockSize);
     };
-    m_moveThreadBlockSize = detail::tuneThreadBlockSize(tuneMoveThreadBlockSizeInUpdate, "Kernel moving data to reordered matrix");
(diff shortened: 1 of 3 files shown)


multitalentloes:
Jenkins build this please

bska:
> That fixes confusing output that could make it look like the tuning was making the triangular solve slower when it wasn't.

Right.  Do you have an example of before/after this change?

multitalentloes:
Here is an example from before and after, though the actual numbers are not comparable since they are run on different hardware

> Kernel computing a lower triangular solve for a level set: Tuned Blocksize: 64 (fastest runtime: 5.481018).
> Kernel computing an upper triangular solve for a level set: Tuned Blocksize: 64 (fastest runtime: 12.579492).

> [Kernel tuning completed] (in DILU apply) Triangular lower solve: Tuned Blocksize = 64, Fastest Runtime = 4.10624ms.
> [Kernel tuning completed] (in DILU apply) Triangular upper solve: Tuned Blocksize = 64, Fastest Runtime = 4.073472ms.

Before the fix, the upper solve did not use the best parameter for the lower solve, causing the runtime shown to be higher than what it will actually be when using the parameters we found. This makes it look like the tuning of the upper solve made it worse.

bska:
> Here is an example from before and after, though the actual numbers are not comparable since they are run on different hardware

Oooh, that's certainly an improvement.  From purely a user point of view I'd be very happy to have this in the master branch, but I'll defer to others (@kjetilly?) to review the technical details here.

PR #5782, opened 2024-12-04
changed: do not use boost.test in test_tasklets_failure

akva2:
the boost.test signal handlers do not play nicely with fork()

i am aware this is not optimal but better a stable test than a sometimes broken test. in particular it breaks with dynamic boost test and clang.


akva2:

atgeirr (approved):

PR #5784, opened 2024-12-05
fixed: always calculate initial FIP

akva2:
this can be used by various Schedule keywords even if FIP is not in RPTSOL

### DIFF
--- a/opm/simulators/flow/EclWriter.hpp
+++ b/opm/simulators/flow/EclWriter.hpp
@@ -350,12 +350,6 @@ public:
     //! \brief Writes the initial FIP report as configured in RPTSOL.
     void writeInitialFIPReport()
     {
-        const auto& fip = simulator_.vanguard().eclState().getEclipseConfig().fip();
-        if (!fip.output(FIPConfig::OutputField::FIELD) &&
-            !fip.output(FIPConfig::OutputField::RESV)) {
-            return;
-        }
-
         const auto& gridView = simulator_.vanguard().gridView();
         const int num_interior = detail::
             countLocalInteriorCellsGridView(gridView);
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

opm/simulators/flow/GenericOutputBlackoilModule.cpp
```diff
@@ -325,6 +325,13 @@ outputInjLog(std::size_t reportStepNum)
     this->logOutput_.injection(reportStepNum);
 }
 
+template<class FluidSystem>
+void GenericOutputBlackoilModule<FluidSystem>::
+calc_initial_inplace(const Parallel::Communication& comm)
+{
+    this->accumulateRegionSums(comm);
```
bska, line 332: Could you add a comment here, please, to remind the reader that `accumulateRegionSums()` as a side effect mutates the `GenericOutputModule::initialInplace_` data member?

bska (commented):
Looks good to me although I had to go read the implementation of `accumulateRegionSums()` to understand why the proposed implementation  of `calc_initial_inplace()` is sufficient.

opm/simulators/flow/GenericOutputBlackoilModule.cpp
akva2: done.

akva2:
jenkins build this please

opm/simulators/flow/EclWriter.hpp
```diff
@@ -350,19 +350,24 @@ class EclWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::Grid>
...
         const auto& gridView = simulator_.vanguard().gridView();
         const int num_interior = detail::
             countLocalInteriorCellsGridView(gridView);
 
         this->outputModule_->
             allocBuffers(num_interior, 0, false, false, /*isRestart*/ false);
 
+        // We always calculate the initial fip values as it may be used by various
+        // keywords in the Schedule, e.g. FIP=2 in RPTSCHED but no FIP in RPTSOL
+        outputModule_->calc_initial_inplace(simulator_.gridView().comm());
```
bska, line 362: Oh, it just occurred to me that this call might be happening a little too early.  If we don't populate the buffers allocated in `allocBuffers()`&ndash;by calling [`updateFluidInPlace()`](https://github.com/OPM/opm-simulators/blob/dea336a25daa52c7e6cee2920c53e2abac51aae1/opm/simulators/flow/EclWriter.hpp#L378)&ndash;then we risk that the initial fluid volumes are all zero.  Please verify that that does not happen now.

bska (commented):
Thanks a lot for the updates and sorry for not identifying the potential issue of all zero volumes before.  We need to check that before we merge this.

opm/simulators/flow/EclWriter.hpp
akva2: nice catch. indeed I need to move that block as well. and then I realized I can simplify this whole thing, see new commit.

bska: > indeed I need to move that block as well

Cool&ndash;I'm happy we caught that.

>. and then I realized I can simplify this whole thing, see new commit.

Nice!

bska:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  This looks really good to me now and I'll merge into master.

PR #5785, opened 2024-12-05
Use initial pressure if ROCKOPTS item 2 is STORE

svenn-t:
If ROCKOPTS item 2 is equal to STORE, initial pressure is used in ROCK and ROCKTAB. For ROCK, the reference pressure input is replaced by initial pressure of the cell, while for ROCKTAB, effective pressure  (i.e. P_eff = P - P_init) will be used in the ROCKTAB table. Note that OVERBURD is ignored if ROCKOPTS item 2 is equal to STORE.

Depends on https://github.com/OPM/opm-common/pull/4370

### DIFF
--- a/opm/simulators/flow/FlowProblem.hpp
+++ b/opm/simulators/flow/FlowProblem.hpp
@@ -172,7 +172,6 @@ public:
     using BaseType::shouldWriteOutput;
     using BaseType::shouldWriteRestartFile;
     using BaseType::rockCompressibility;
-    using BaseType::rockReferencePressure;
     using BaseType::porosity;
 
     /*!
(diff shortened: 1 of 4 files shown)


opm/simulators/flow/FlowProblem.hpp
```diff
@@ -719,7 +718,28 @@ class FlowProblem : public GetPropType<TypeTag, Properties::BaseProblem>
...
+    */
+    Scalar rockReferencePressure(unsigned globalSpaceIdx) const
+    {
+        const auto& rock_config = this->simulator().vanguard().eclState().getSimulationConfig().rock_config();
+        if (rock_config.store()) {
+            return asImp_().initialFluidState(globalSpaceIdx).pressure(refPressurePhaseIdx_());
+        }
+        else {
+            if (this->rockParams_.empty())
+                return 1e5;
```
bska, line 735: Can we please not use magic constants in our code?

svenn-t: Everything in the `else {}` section here is from the original code previously located in `FlowGenericProblem_impl.hpp`, so I just moved it wholesale. From what I can see, just deleting the `if (this->rockParams_.empty())` clause will be OK since `rockReferencePressure()` will not be called without `ROCK` being present in the deck but I may be wrong in that assumption.

totto82:
jenkins build this please

opm/simulators/flow/FlowProblem.hpp
totto82: I agree about reducing the number of magic constants in the code, but I would rather see that done in another PR.

totto82 (approved):
Thanks for the contribution. I have looked at the code here and in opm-common and it both looks fine.

PR #5791, opened 2024-12-10
Allow for the third group switch

totto82:


totto82:
jenkins build this please

bska (approved):
I think this makes sense.  Maybe we could add a comment here to explain why `<=` is the correct condition?

PR #5793, opened 2024-12-11
Make OpenCL optional

kjetilly:
The current version of  the CMake setup compiles OpenCL as long as OpenCL is found. This PR makes OpenCL an optional feature that can be turned off (by default it is on) even if OpenCL is present on the system.

Use case: Faster compilation times, or present but broken OpenCL setup on a computer.


kjetilly:
Jenkins build this please

CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ option(USE_GPU_BRIDGE "Enable the GPU bridge (GPU/AMGCL solvers)" ON)
 option(USE_TRACY_PROFILER "Enable tracy profiling" OFF)
 option(CONVERT_CUDA_TO_HIP "Convert CUDA code to HIP (to run on AMD cards)" OFF)
 set(OPM_COMPILE_COMPONENTS "2;3;4;5;6;7" CACHE STRING "The components to compile support for")
+option(OPM_ENABLE_OPENCL "Enable OpenCL support?" ON)
```
akva2, line 39: please use USE_OPENCL for consistency with other such options.

kjetilly: Fixed. I took inspiration from `OPM_ENABLE_PYTHON` I think.

kjetilly:
Jenkins build this please

CMakeLists.txt
akva2: sounds likely. i'd really like to change that as well but it has lived too long now so i'd break too much...

akva2 (approved):

blattms:
Shouldn't setting CMAKE_DISABLE_FIND_PACKAGE to ON do the exact same thing? Did this not work?

kjetilly:
> Shouldn't setting CMAKE_DISABLE_FIND_PACKAGE_OpenCL to ON do the exact same thing? Did this not work?

Potentially, yes, but if I am not mistaken, this is applied to the whole cmake system, so if any third party package also tries to find OpenCL, that would fail. It will also mess up if you have one `find(OpenCL REQUIRED)` somewhere else in the cmake system (say from an external package being found). It does not work well with caching, so if I first run cmake without this option OpenCL will already have been found, so I would need to clear the cmake cache to disable OpenCL again.

And as a matter of principle, I think it is nicer to expose optional features with an explicit option. It would probably be good in the medium term to also make OpenCL required if the `USE_OPENCL` option is enabled (and similarly for CUDA). Failing silently makes for a lot of headache when debugging is my experience.

PR #5794, opened 2024-12-11
fixed: remove HYPRE from prereqs

akva2:
this is controlled by USE_HYPRE and we should not unconditionally look for it.

pointy hat [at]akva2


akva2:

bska (approved):
Very good.  Please merge when green to restore the sequential build.

akva2:
jenkins build this serial hipify please

CMakeLists.txt
```diff
@@ -328,6 +328,22 @@ macro (config_hook)
   if(have_float_from_chars)
     set(HAVE_FLOATING_POINT_FROM_CHARS 1)
   endif()
+
+  if(USE_HYPRE)
```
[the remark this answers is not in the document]
akva2: sure.

akva2:
jenkins build this serial hipify please

CMakeLists.txt
```diff
@@ -328,6 +328,21 @@ macro (config_hook)
...
+    find_package(HYPRE)
+    if(HYPRE_FOUND)
+      set(HAVE_HYPRE 1)
+      list(APPEND opm-simulators_LIBRARIES HYPRE::HYPRE)
+    else()
+      message(WARNING "Hypre requested but not found. Continuing without Hypre support.")
+      set(USE_HYPRE OFF)
+    endif()
+  else()
+    message(WARNING "Hypre requested but MPI not activated. Continuing without Hypre support.")
```
bska, line 342: This message is possibly a little misleading since it will be issued unless both `USE_HYPRE` and `USE_MPI` are activated.  Should probably be guarded by `if(USE_HYPRE)` here.  The `set()` commands must be unconditional in this branch though.

akva2: yeah, time to actually focus on this and not what i was actually doing.. you are of course correct. the sets are unnecessary unless we hit the conditional.

and i always has to check the hook order cause they always confuse me. note to self: we run sources before config which is why we have all the find blocks before we invoke LibMain. this should be changed!

akva2:
jenkins build this serial hipify please

CMakeLists.txt
```diff
@@ -289,6 +289,20 @@ if(OpenCL_FOUND)
...
+if(USE_HYPRE AND USE_MPI)
+  find_package(HYPRE)
+  if(HYPRE_FOUND)
+    set(HAVE_HYPRE 1)
+    list(APPEND opm-simulators_LIBRARIES HYPRE::HYPRE)
+  else()
+    message(WARNING "Hypre requested but not found. Continuing without Hypre support.")
+    set(USE_HYPRE OFF)
+  endif()
+elseif(USE_MPI)
```
bska, line 301: This needs to be either `USE_HYPRE` or `NOT USE_MPI`.

akva2:
jenkins build this serial hipify please

CMakeLists.txt
akva2: aye

PR #5798, opened 2024-12-11
Regression test for the partitioner with GraphOfGrid

michal-toth:
Add a regression test for the new partitioner that uses graph representation of the grid in which a well is represented by a single cell. The test uses SPE3CASE1 grid which has wells with multiple cells (unlike SPE1 or SPE5).

### DIFF
--- a/regressionTests.cmake
+++ b/regressionTests.cmake
@@ -256,6 +256,13 @@ add_test_compareECLFiles(CASENAME spe3
                          REL_TOL ${coarse_rel_tol}
                          TEST_ARGS --tolerance-wells=1e-6 --newton-max-iterations=20)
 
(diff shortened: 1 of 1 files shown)


michal-toth:

michal-toth:
The regression test that is needed must be parallel in order to use the grid partitioner. Changes of this pull request add a regression test which is sequential. I would like to add a different test that also targets a SPE3 test but is parallel and the problem is that the macros for parallel tests do not have an option to add the prefix to the name. I have to find another way to resolve the name clash, and the pull request might be reopened afterwards.
However, if the partitioner with GraphOfGrid becomes the default option, this PR will become unnecessary. (We might need another one for other partitioner(s).)

PR #5799, opened 2024-12-13
Compute temperature for xflowing injectors as weighted average 

vkip:
Weighted average of fluids entering the well from both surface and reservoir.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -2875,15 +2875,10 @@ namespace Opm {
         for (auto wellID = 0*nw; wellID < nw; ++wellID) {
             const Well& well = this->wells_ecl_[wellID];
             auto& ws = this->wellState().well(wellID);
-            if (well.isInjector()){
-                if( !(ws.status == WellStatus::STOP)){
-                    this->wellState().well(wellID).temperature = well.inj_temperature();
-                    continue;
-                }
-            }
 
-            std::array<Scalar,2> weighted{0.0,0.0};
-            auto& [weighted_temperature, total_weight] = weighted;
+            std::array<Scalar,4> weighted{0.0,0.0,0.0,0.0};
+            auto& [total_weight_prod, total_weight_inj, weighted_temp_prod, weighted_temp_inj] = weighted;
+            const Scalar injectionTemperature = well.isInjector() ? well.inj_temperature() : 0.0;
 
             auto& well_info = this->local_parallel_well_info_[wellID].get();
             auto& perf_data = ws.perf_data;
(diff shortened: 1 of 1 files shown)


vkip:
jenkins build this please

GitPaean:
I am not familiar with the energy conservation equations. But ideally, should we have a primary variable to monitor the temperatures in the wellbore to handle the crossflow through heat conservation?

vkip:
> I am not familiar with the energy conservation equations. But ideally, should we have a primary variable to monitor the temperatures in the wellbore to handle the crossflow through heat conservation?

Maybe @hnil and/or @totto82 can comment on that. I'm just trying to get slightly more representative output with the current approach..

totto82:
If we are doing this only for output. I think a post processing like @vkip is suggestion is ok. But if we want to use this computed temperature for the energy equation I think we need to add a variable and solve for it as @GitPaean suggests. Maybe we should start investigating this for MSW? For STW we will only have one temperature for the whole well. For cross-flowing wells that may be a too coarse approximation.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -2908,14 +2902,20 @@ namespace Opm {
                     cellBinv = fs.invB(phaseIdx).value();
                     cellDensity = fs.density(phaseIdx).value();
                     perfPhaseRate = perf_phase_rate[ perf*np + phaseIdx ];
-                    weight_factor += cellDensity  * perfPhaseRate/cellBinv * cellInternalEnergy/cellTemperatures;
+                    const Scalar vol_heatcap = cellDensity/cellBinv * cellInternalEnergy/cellTemperature;
+                    const Scalar heatcap_rate_prod = -std::min(perfPhaseRate, 0.0)*vol_heatcap;
+                    const Scalar heatcap_rate_inj = perfPhaseRate*vol_heatcap;
+                    total_weight_prod += heatcap_rate_prod;
```
totto82, line 2908: Why is this not std::max(perfPhaseRate, 0.0)? (to get the inflow?)

vkip:
jenkins build this please

GitPaean:
I assume this PR is for improving the output of the SUMMARY keyword  WTICHEA, while from the summary output, it does not look it makes a difference for SPE1CASE2_THERMAL,  maybe I am mistaken here. 

```
-- Days   dd/mm/yyyy        WTICHEA:INJ     WTPCHEA:PROD 
---------------------------------------------------------
   1.00   02/01/2015                200          183.282 
   4.00   05/01/2015                200          183.187 
  11.43   12/01/2015                200          183.015 
  31.00   01/02/2015                200          182.593 
  59.00   01/03/2015                200          182.046 
  90.00   01/04/2015                200          181.505 
 120.00   01/05/2015                200          181.057 
 151.00   01/06/2015                200          180.639 
 181.00   01/07/2015                200           180.27 
 212.00   01/08/2015                200          179.916 
 243.00   01/09/2015                200           179.57 
 273.00   01/10/2015                200          179.265 
 304.00   01/11/2015                200          178.977 
 334.00   01/12/2015                200          178.709 
 365.00   01/01/2016                200          178.477 
 396.00   01/02/2016                200          178.243 
 424.00   29/02/2016                200          178.059 
```

PR #5802, opened 2024-12-16
Feature/ms wells fix restart file error

lisajulia:
Fix error in restart files for parallel runs and remove option to not compare them since this is not needed anymore now.
FYI: I used the undo debugger to trace this bug: https://docs.undo.io/GettingStartedWithUDB.html.

This bugfix repairs the known issues of PR #5746.


lisajulia:
jenkins build this please

bska (commented):
Very good.  Other than the fact that we might be able to get away with a single call to the cross-rank `sum()` function, this looks good to me.

lisajulia:
jenkins build this please

bska (approved):

PR #5803, opened 2024-12-16
remove unused fipnum argument

akva2:
also allows to drop the default value for the tuningUpdater callback


akva2:
jenkins build this please

bska (approved):
Excellent!  If it turns out that we actually need this functionality at a later date we can reintroduce it in a more comprehensive way.

I'll merge into master.

PR #5808, opened 2024-12-17
Amgx integration

jakobtorben:
## Description

This PR adds support for using the AMGX library in OPM. AMGX is a high-performance algebraic multigrid solver for NVIDIA GPUs.

For flexibility and being able to use AMGX with the other components in OPM, it has been implemented as a preconditioner, templated on vectors and matrices, such that it can be wrapped in the ISTL format used in OPM. The preconditioner handles the copying to and from the AMGX memory objects. To avoid this extra memory operations, we would have to assemble the matrix directly into the AMGX matrix, which is also possible to implement at a later stage. MPI parallelism and multi-GPU is currently not supported.

Since we are most interested in using the AMG as a coarsesolver for the two-stage CPR-AMG preconditioner in OPM, it only supports scalar matrices (1x1 blocks).

The integration supports only execution on the GPU (device), as the CPU (host) implementation in AMGX does not support all the operations we require.

### Installation
To use the AMGX preconditioner, the AMGX library must be installed from source, with the following commands:

```bash
git clone --recursive git@github.com:nvidia/amgx.git
mkdir amgx/build
cd amgx/build
cmake ../
make -j12 all
```
OPM must then be built with AMGX support using:

```bash
cmake .. -DUSE_AMGX=ON -DAMGX_DIR:PATH=/your_path_to_amgx
make -j 12
```

### Using the solver
The AMGX preconditioner can be selected with the usual OPM linear solver setup tree:

```json
{
    "maxiter": "20",
    "tol": "0.005",
    "verbosity": "0",
    "solver": "bicgstab",
    "preconditioner": {
        "type": "cprw",
        "use_well_weights": "false",
        "add_wells": "true",
        "weight_type": "trueimpes",
        "pre_smooth": "0",
        "post_smooth": "1",
        "finesmoother": {
            "type": "GPUDILU",
            "split_matrix": "true",
            "tune_gpu_kernels": "true"
        },
        "verbosity": "0",
        "coarsesolver": {
            "maxiter": "1",
            "tol": "0.1",
            "solver": "loopsolver",
            "verbosity": "0",
            "preconditioner": {
                "type": "amgx",
                "determinism_flag": "0",
                "print_grid_stats": "0",
                "print_solve_stats": "0",
                "solver": "AMG",
                "algorithm": "CLASSICAL",
                "interpolator": "D2",
                "selector": "PMIS",
                "smoother": "BLOCK_JACOBI",
                "presweeps": "3",
                "postsweeps": "3",
                "strength_threshold": "0.5",
                "max_iters": "1"
            }
        }
    }
}
```
Note that one does not need to set all the AMGX parameters, as they have defined defaults.

## Results
To test the AMG, we defined a variation of the SPE10 case with 1.1 million cells. Where we set the timesteps and maximum iterations such that it completes the simulation with no timestep cuts and maximum timestep sizes. This allows a direct comparison between the different solvers. Below are three different variants:
- OPM default: CPRW with Dune AMG on the CPU, and ILU0 second stage preconditioner on the CPU
- CPRW with Hypre AMG on the GPU (with default Hypre GPU parameters), and DILU second stage preconditioner on the GPU
- CPRW with AMGX on the GPU (with default AMGX parameters), and DILU second stage preconditioner on the GPU

Note that the other components of the linear solver (CPR, well operators and Krylov BiCGStab solver) is still performed on the CPU for all the configurations.

Note that Dune AMG and AMGX supports reusing the AMG hierarchy between nonlinear iterations but Hypre does not. In the above test, Dune AMG and AMGX rebuilds the hierarchy every 30th Newton iteration. For this reason we can see that the AMGX run has a lower linear setup compared to Hypre (GPU). All tests has been executed on an AMD Ryzen 9 5900X 12-Core Processor CPU and NVIDIA GeForce RTX 3090 Ti GPU. Two top consumer models from the same period. 

performance_comparison_parallel

iteration_numbers_baseline

### Tuned parameters
In OPM, the CPR-AMG preconditioners has a finesmoother that operators on the fine matrix (complete blocked system). The current default is to apply this once after the coarsesolver (AMG). For some cases it can be beneficial to also to apply this preconditioner more than once, and also before using the coarsesolver. Below are result from using this finesmoother 3 times before and 1 time after. 
performance_comparison_parallel_pre_3_post_1

iteration_numbers_pre_3_post_1


jakobtorben:
jenkins build this opm-common=4387 please

opm/simulators/linalg/AmgxPreconditioner.hpp
```diff
@@ -0,0 +1,348 @@
...
+    std::string algorithm = "CLASSICAL";
+    std::string interpolator = "D2";
+    std::string selector = "PMIS";
+    std::string smoother = "BLOCK_JACOBI";
+    int presweeps = 3;
+    int postsweeps = 3;
+    double strength_threshold = 0.5;
+    int max_iters = 1;
+
+    AmgxConfig(const Opm::PropertyTree& prm) {
```
atgeirr, line 57: Make constructor explicit.

opm/simulators/linalg/AmgxPreconditioner.hpp
```diff
@@ -0,0 +1,348 @@
...
+        if (x_amgx_) {
+            AMGX_SAFE_CALL(AMGX_vector_destroy(x_amgx_));
+        }
+        if (b_amgx_) {
+            AMGX_SAFE_CALL(AMGX_vector_destroy(b_amgx_));
+        }
+        if (A_amgx_) {
+            AMGX_SAFE_CALL(AMGX_matrix_destroy(A_amgx_));
+        }
+        // Destroying resrouces and config crashes when reinitializing
```
atgeirr, line 193: Is this a bug in AMGX or in our usage?

opm/simulators/linalg/AmgxPreconditioner.hpp
```diff
@@ -0,0 +1,348 @@
...
+     * @param d The defect vector.
+     */
+    void pre(X& /*v*/, Y& /*d*/) override {
+    }
+
+    /**
+     * @brief Applies the preconditioner to a vector.
+     *
+     * Performs one AMG cycle to solve the system.
+     * Involves uploading vectors to AMGX, applying the preconditioner, and downloading the result.
```
atgeirr, line 217: This means this class behaves like a CPU preconditioner outwardly, would it be simple to make it a pure GPU preconditioner instead and use the wrapper to make a CPU preconditioner of it (for combining with the current CPU CPRW etc.)?

opm/simulators/linalg/AmgxPreconditioner.hpp
```diff
@@ -0,0 +1,348 @@
...
+     * @brief Updates the preconditioner with the current matrix values.
+     *
+     * This method should be called whenever the matrix values change.
+     */
+    void update() override
+    {
+        OPM_TIMEBLOCK(prec_update);
+        copyMatrixToAmgx();
+        if (update_counter_ == 0) {
+            AMGX_SAFE_CALL(AMGX_solver_setup(solver_, A_amgx_));
```
atgeirr, line 257: Is it safe to do setup without destroying it first?

atgeirr (commented):
Looks good, just a few questions.

opm/simulators/linalg/AmgxPreconditioner.hpp
akva2: also typos

akva2:
jenkins build this opm-common=4387 serial please

opm/simulators/linalg/AmgxPreconditioner.hpp
jakobtorben: Yes exactly! I do believe it is possible to make it a pure GPU preconditioner, and maybe reusing the device memory, as both AMGX and gpuISTL uses CuSparse under the hood. This is something I don't have time to do now but I want to explore further next year. So suggesting that we merge it in as it is now and revisit it.

opm/simulators/linalg/AmgxPreconditioner.hpp
jakobtorben: Yes this should be safe. There is some functionality to reuse some memory and levels of the AMG in the setup function also. I have done some digging in the code but still not 100 % clear to me how much more is reused with resetup.

Only using the setup function is actually the way to do it with the official API and in the manual, as the resetup was marked deprecated in the initial AMGX commit 7 years ago... https://github.com/NVIDIA/AMGX/blob/69e625e99b8788ac7b259c6b217ad4453cc69464/examples/amgx_capi.h#L522
But since it still available 7 years later and important to us I decided to still use it.

opm/simulators/linalg/AmgxPreconditioner.hpp
jakobtorben: Fixed.

atgeirr (approved):

atgeirr:
I am fine with merging the code here when green, but I would appreciate someone else checking the Cmake code in the upstream PR. @akva2 ?

akva2 (approved):

PR #5809, opened 2024-12-18
Regression test for --partition-method=3

michal-toth:
This PR contains regression test for the `flow` option `--partition-method=3`. This partitioning method represents each well by one cell, therefore a well can not be split over several subdomains by the partitioner. The regression test uses SPE3CASE1 which has wells that have multiple cells (unlike e.g. spe1 or spe5).

Since there is already a test for SPE3CASE1, we got a name conflict with the new test. We extended the macro generating the test to accept the PREFIX keyword, which enabled us to distinguish two tests (with different partitioners) comparing to the same file.

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -194,12 +194,16 @@ endfunction()
 #   - This test class compares the output from a parallel simulation
 #     to the output from the serial instance of the same model.
 function(add_test_compare_parallel_simulation)
-  set(oneValueArgs CASENAME FILENAME SIMULATOR ABS_TOL REL_TOL DIR MPI_PROCS)
+  set(oneValueArgs CASENAME FILENAME SIMULATOR ABS_TOL REL_TOL DIR PREFIX MPI_PROCS)
   set(multiValueArgs TEST_ARGS)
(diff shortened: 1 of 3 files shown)


akva2:
no need to extend the macro.
use
```
add_test_compareECLFiles(CASENAME spe3_partitionmethod3
                         FILENAME SPE3CASE1
                         SIMULATOR flow
                         ABS_TOL ${abs_tol}
                         REL_TOL ${coarse_rel_tol}
                         TEST_ARGS --partition-method=3
                         DIR spe3)
```

michal-toth:
The problem is with `add_test_compare_parallel_simulation`, not `add_test_compareECLFiles` (which supported `PREFIX` before). However, extending `CASENAME` and specifying `DIR` gives me the compilation error:
CMake Error at .../opm-common/cmake/Modules/OpmSatellites.cmake:352 (add_test):
  add_test given test NAME "compareParallelSim_flow+SPE3CASE1" which already
  exists in this directory.

michal-toth:
I will remove the sequential test.

akva2:
yeah i totally understand. it's all built from a this-was-needed with some hidden uses here and there, and not from a coherent design.

michal-toth:
Thanks for suggestions. It seems almost every line of this PR has to be changed, so I will create a new PR with the changes we want.

PR #5810, opened 2024-12-18
Regression test for --partition-method=3

michal-toth:
This PR contains regression test for the flow option --partition-method=3. This partitioning method represents each well by one cell, therefore a well can not be split over several subdomains by the partitioner. The regression test uses SPE3CASE1 which has wells that have multiple cells (unlike e.g. spe1 or spe5).

Since there is already a test for SPE3CASE1, we got a name conflict with the new test. We extended the macro generating the test to accept the POSTFIX keyword, which enabled us to distinguish two tests (with different partitioners) having the same CASENAME and FILENAME.

### DIFF
--- a/compareECLFiles.cmake
+++ b/compareECLFiles.cmake
@@ -194,12 +194,16 @@ endfunction()
 #   - This test class compares the output from a parallel simulation
 #     to the output from the serial instance of the same model.
 function(add_test_compare_parallel_simulation)
-  set(oneValueArgs CASENAME FILENAME SIMULATOR ABS_TOL REL_TOL DIR MPI_PROCS)
+  set(oneValueArgs CASENAME FILENAME SIMULATOR ABS_TOL REL_TOL DIR POSTFIX MPI_PROCS)
   set(multiValueArgs TEST_ARGS)
   cmake_parse_arguments(PARAM "$" "${oneValueArgs}" "${multiValueArgs}" ${ARGN} )
 
   if(NOT PARAM_DIR)
     set(PARAM_DIR ${PARAM_CASENAME})
+  endif()
+
+  if(NOT PARAM_POSTFIX)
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this please

compareECLFiles.cmake
```diff
@@ -221,11 +225,11 @@ function(add_test_compare_parallel_simulation)
                   -n ${MPI_PROCS})
 
   # Add test that runs flow_mpi and outputs the results to file
-  opm_add_test(compareParallelSim_${PARAM_SIMULATOR}+${PARAM_FILENAME} NO_COMPILE
+  opm_add_test(compareParallelSim_${PARAM_SIMULATOR}+${PARAM_FILENAME}+${PARAM_POSTFIX} NO_COMPILE
```
akva2, line 228: this will add the + everywhere. you need to prepend it to PARAM_POSTFIX

parallelTests.cmake
```diff
@@ -78,6 +78,14 @@ add_test_compare_parallel_simulation(CASENAME spe3
                                      REL_TOL ${coarse_rel_tol_parallel}
                                      TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8 --tolerance-wells=1e-7)
 
+add_test_compare_parallel_simulation(CASENAME spe3
```
akva2, line 81: this needs a different casename because the casename is used for folder names and it now clashes with the other test. then you also need to use DIR.

michal-toth:
jenkins build this please

akva2:
i have white listed you now, but i can trigger

akva2:
jenkins build this please

michal-toth:
> Thanks for test and patience.

Thank you for the guidance. And for whitelisting!

PR #5812, opened 2024-12-18
Allow query for efficiency scaling factor in off-process wells where required

vkip:
Fixes an issue with MPI hang when well_state[name] throws for off-process wells.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1576,7 +1576,7 @@ calculateEfficiencyFactors(const int reportStepIdx)
     for (auto& well : well_container_generic_) {
         const Well& wellEcl = well->wellEcl();
         Scalar well_efficiency_factor = wellEcl.getEfficiencyFactor() *
-                                        wellState()[well->name()].efficiency_scaling_factor;
+                                        wellState().getGlobalEfficiencyScalingFactor(well->name());
         WellGroupHelpers<Scalar>::accumulateGroupEfficiencyFactor(schedule().getGroup(wellEcl.groupName(),
                                                                                       reportStepIdx),
                                                                   schedule(),
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1061,8 +1061,7 @@ namespace Opm {
                                                                           this->well_open_times_,
                                                                           schedule_open))
                 {
-                    if (this->wellState().has(wname))
-                        this->wellState()[wname].efficiency_scaling_factor = wscale;
+                    this->wellState()[wname].efficiency_scaling_factor = wscale;
                     this->schedule_.add_event(ScheduleEvents::WELLGROUP_EFFICIENCY_UPDATE, report_step);
                 }
             }
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -655,7 +655,7 @@ namespace Opm {
             well->init(&this->phase_usage_, depth_, gravity_, B_avg_, true);
 
             Scalar well_efficiency_factor = wellEcl.getEfficiencyFactor() *
-                                            this->wellState()[well_name].efficiency_scaling_factor;
+                                            this->wellState().getGlobalEfficiencyScalingFactor(well_name);
             WellGroupHelpers<Scalar>::accumulateGroupEfficiencyFactor(this->schedule().getGroup(wellEcl.groupName(),
                                                                                                 timeStepIdx),
                                                                       this->schedule(),
@@ -1061,7 +1061,8 @@ namespace Opm {
                                                                           this->well_open_times_,
                                                                           schedule_open))
                 {
-                    this->wellState()[wname].efficiency_scaling_factor = wscale;
+                    if (this->wellState().has(wname))
+                        this->wellState()[wname].efficiency_scaling_factor = wscale;
                     this->schedule_.add_event(ScheduleEvents::WELLGROUP_EFFICIENCY_UPDATE, report_step);
                 }
             }
--- a/opm/simulators/wells/GasLiftStage2.cpp
+++ b/opm/simulators/wells/GasLiftStage2.cpp
@@ -832,7 +832,7 @@ computeDelta(const std::string& well_name, bool add)
         if (this->well_state_.wellIsOwned(well.indexOfWell(), well_name)) {
             const auto& well_ecl = well.wellEcl();
             Scalar factor = well_ecl.getEfficiencyFactor() *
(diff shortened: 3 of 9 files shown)


vkip:
jenkins build this please

vkip:
Failing tests are due to minor numerical differences and disappear, locally here at least, if setting the relaxed cnv tolerance equal to the strict cnv tolerance (--tolerance-cnv-relaxed=0.01).

bska:
@akva2 : Would you mind having a look at this, please?

opm/simulators/wells/WellState.hpp
```diff
@@ -180,6 +180,11 @@ class WellState
         return this->global_well_info.value().is_open(name);
     }
 
+    bool getGlobalEfficiencyScalingFactor(const std::string& name) const
```
akva2, line 183: took me a while to track this one. not a bool!

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1576,7 +1576,7 @@ calculateEfficiencyFactors(const int reportStepIdx)
     for (auto& well : well_container_generic_) {
         const Well& wellEcl = well->wellEcl();
         Scalar well_efficiency_factor = wellEcl.getEfficiencyFactor() *
-                                        wellState()[well->name()].efficiency_scaling_factor;
+                                        wellState().getGlobalEfficiencyScalingFactor(well->name());
```
blattms, line 1579: At other places there seems to be a check that the well is owned around this instead
```
if (this->well_state_.wellIsOwned(well.indexOfWell(), well.name())) {
  ...
```
Wouldn't this work here and the other places, too?

opm/simulators/wells/GasLiftStage2.cpp
```diff
@@ -832,7 +832,7 @@ computeDelta(const std::string& well_name, bool add)
         if (this->well_state_.wellIsOwned(well.indexOfWell(), well_name)) {
             const auto& well_ecl = well.wellEcl();
             Scalar factor = well_ecl.getEfficiencyFactor() *
-                            this->well_state_[well_name].efficiency_scaling_factor;
+                            this->well_state_.getGlobalEfficiencyScalingFactor(well_name);
```
blattms, line 835: Are you sure there was a problem here? 
Line 832 should prevent processing for wells that are not owned by us via `if (this->well_state_.wellIsOwned(well.indexOfWell(), well_name))`

Maybe we could use this at the other places, too?

blattms (commented):
Without having looked much into this, I hope we can skip adding the global array.

At least for distributed wells the approach might count factors multiple times as is.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
blattms: Nope, it won't. (I was tricked by accumlate... below).
For distributed wells, we need to do this on all processes that have cells of a well.

opm/simulators/wells/WellState.hpp
vkip: Wow, the joys of copy-paste to save a few keystrokes.. Thanks!

opm/simulators/wells/GasLiftStage2.cpp
vkip: You're right - got a bit too eager here.. will revert.

akva2:
@vkip please have a look at https://github.com/OPM/opm-simulators/pull/5814

Maybe whatever you are trying to fix was caused by that? All the cases pass parallel regression tests (I only added one so we don't inflate the test suite too much).

opm/simulators/wells/GlobalWellInfo.hpp
```diff
@@ -64,6 +64,7 @@ class GlobalWellInfo {
         comm.sum( this->m_in_injecting_group.data(), size);
         comm.sum( this->m_in_producing_group.data(), size);
         comm.sum( this->m_is_open.data(), size);
+        comm.min( this->m_efficiency_scaling_factors.data(), size);
```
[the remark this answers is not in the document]
vkip: Factors as initialized to 1, but may be set to something lower than 1 on the process that owns the well. Min to the this value on all processes (required a few places, e.g. in the network pressure calculations).

vkip:
> @vkip please have a look at #5814
> 
> Maybe whatever you are trying to fix was caused by that? All the cases pass parallel regression tests (I only added one so we don't inflate the test suite too much).

Nope, both network and gas lift calculations need to get the efficiency factor for off process wells, in which case well_state[well_name] will throw..

vkip:
jenkins build this please

akva2:
okay, we definitely need both PRs then.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1061,7 +1061,8 @@ namespace Opm {
                                                                           this->well_open_times_,
                                                                           schedule_open))
                 {
-                    this->wellState()[wname].efficiency_scaling_factor = wscale;
+                    if (this->wellState().has(wname))
```
akva2, line 1064: we shouldn't be here unless we have the well so this if should be redundant?

vkip: ah, right - will revert

opm/simulators/wells/BlackoilWellModelGeneric.cpp
blattms: @akva2 I think this is where it broke and what @vkip is fixing.

vkip: Not sure I follow - as commented below, both network and gas lift calculations need to get the efficiency factor for off process wells.

akva2: I don't think this block is actually where the real fix is situated. This loops over the well_container_generic_ and that only has entries for local active wells. The real reason for the change are the two call sites in GasLiftGroupInfo.

vkip: and this one: https://github.com/OPM/opm-simulators/blob/ee972dcbe2187c1051773415dda47f37f072f2b0/opm/simulators/wells/WellGroupHelpers.cpp#L875-L876

akva2: indeed.

vkip:
jenkins build this please

akva2:
@vkip please pick and amend with https://github.com/akva2/opm-simulators/commit/c717f94a7a107eb136b001eb454221ac9007b8c1

vkip:
> @vkip please pick and amend with [akva2@c717f94](https://github.com/akva2/opm-simulators/commit/c717f94a7a107eb136b001eb454221ac9007b8c1)

Pretty sure I'll mess that up - probably safer if you just create a new PR, so closing this one.

PR #5813, opened 2024-12-18
Do not write initial FIP for CpGrid with LGRs for now

aritorto:
Unfortunately, for CpGrid with LGRs we need to skip - for now - the computation of initial FIP. This affects the "always calculate initial FIP" done in OPM/opm-simulators#5784 in the sense that CpGrid with LGRs is excluded for the moment.


aritorto:
jenkins build this please

aritorto:

PR #5822, opened 2024-12-20
Fixes related to zero guiderates from NETV

totto82:
This commit includes several fixes related to supporting NETV guide rates that are zero. Depends on https://github.com/OPM/opm-common/pull/4391

### DIFF
--- a/opm/simulators/wells/FractionCalculator.cpp
+++ b/opm/simulators/wells/FractionCalculator.cpp
@@ -77,11 +77,11 @@ Scalar FractionCalculator<Scalar>::
 localFraction(const std::string& name,
               const std::string& always_included_child)
 {
-    bool only_use_potentials = false;
-    const Scalar my_guide_rate = guideRate(name, always_included_child, only_use_potentials);
+    bool always_use_potentials = false;
+    const Scalar my_guide_rate = guideRate(name, always_included_child, always_use_potentials);
 
     const Group& parent_group = schedule_.getGroup(parent(name), report_step_);
-    const auto [total_guide_rate, num_active_groups] = guideRateSum(parent_group, always_included_child, only_use_potentials);
+    const auto [total_guide_rate, num_active_groups] = guideRateSum(parent_group, always_included_child, always_use_potentials);
 
     // the group/well "name" is the only active group/well we therefore return 1 as the fraction
     // even though my_guide_rate may be zero
@@ -91,9 +91,9 @@ localFraction(const std::string& name,
     if (total_guide_rate == 0 ) {
         // if the total guide rate is zero (for instance due to netv = 0) we use the potentials
         // to distribute the group rate
-        only_use_potentials = true;
-        const Scalar my_pot = guideRate(name, always_included_child, only_use_potentials);
-        const Scalar my_total_pot = guideRateSum(parent_group, always_included_child, only_use_potentials).first;
+        always_use_potentials = true;
+        const Scalar my_pot = guideRate(name, always_included_child, always_use_potentials);
+        const Scalar my_total_pot = guideRateSum(parent_group, always_included_child, always_use_potentials).first;
         return my_pot / my_total_pot;
     }
     return my_guide_rate / total_guide_rate;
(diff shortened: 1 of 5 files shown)


opm/simulators/wells/FractionCalculator.cpp
```diff
@@ -122,7 +131,10 @@ guideRateSum(const Group& group,
                         (ctrl == Group::InjectionCMode::NONE);
         }
         if (included) {
-            total_guide_rate += guideRate(child_group, always_included_child);
+            if(groupControlledWells(child_group, always_included_child) > 0) {
```
atgeirr, line 134: Nitpick: space after if.

opm/simulators/wells/FractionCalculator.cpp
```diff
@@ -104,11 +111,13 @@ parent(const std::string& name)
...
 template<class Scalar>
-Scalar FractionCalculator<Scalar>::
+std::pair<Scalar, int> FractionCalculator<Scalar>::
 guideRateSum(const Group& group,
-             const std::string& always_included_child)
+             const std::string& always_included_child,
+             const bool use_potentials)
 {
     Scalar total_guide_rate = 0.0;
+    int number_of_included_well_or_group = 0;
```
atgeirr, line 120: Suggest plural in the name: `number_of_included_wells_or_groups`.

opm/simulators/wells/FractionCalculator.hpp
```diff
@@ -56,10 +56,12 @@ class FractionCalculator
 
 private:
     std::string parent(const std::string& name);
-    Scalar guideRateSum(const Group& group,
-                        const std::string& always_included_child);
+    std::pair<Scalar,int> guideRateSum(const Group& group,
```
atgeirr, line 59: Add docs here and for guideRate(). The new return value and parameter must be documented at least.

Nitpick: space after comma.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -300,32 +300,31 @@ updateGuideRatesForInjectionGroups(const Group& group,
...
-        Scalar guideRateValue = 0.0;
+        std::optional<Scalar> guideRateValue;
         const auto& controls = group.injectionControls(phase, summaryState);
         switch (controls.guide_rate_def){
         case Group::GuideRateInjTarget::RATE:
             break;
         case Group::GuideRateInjTarget::VOID:
         {
-            guideRateValue = group_state.injection_vrep_rate(group.name());
+            guideRateValue = std::max(Scalar(0.0), group_state.injection_vrep_rate(group.name()));
```
atgeirr, line 310: Why is the max() needed here? Also added on line 324.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -300,32 +300,31 @@ updateGuideRatesForInjectionGroups(const Group& group,
...
+                guideRateValue = *guideRateValue - injRES[pu.phase_pos[BlackoilPhases::Vapour]];
             if (phase != Phase::WATER && pu.phase_used[BlackoilPhases::Aqua])
-                guideRateValue -= injRES[pu.phase_pos[BlackoilPhases::Aqua]];
+                guideRateValue = *guideRateValue - injRES[pu.phase_pos[BlackoilPhases::Aqua]];
+
+            guideRateValue = std::max(Scalar(0.0), *guideRateValue);
             break;
         }
         case Group::GuideRateInjTarget::RESV:
-            OPM_DEFLOG_THROW(std::runtime_error, "GUIDE PHASE RESV not implemented. Group " + group.name(), deferred_logger);
```
atgeirr, line 326: Should this line really be deleted, or was it a victim of the POTN removal?

atgeirr (commented):
A few minor issues, should be merged when addressed.

opm/simulators/wells/WellGroupHelpers.cpp
totto82: Thanks for noticing. I will add it back.

opm/simulators/wells/WellGroupHelpers.cpp
totto82: we assume that all guiderates are positive when we compute the fractions. In the extreme case of all wells have zero target I though that we may end of with negative rates due to lack of numerical precision. The max is just a precaution. We can probably remove it and nobody will notice in a 100 year.

opm/simulators/wells/WellGroupHelpers.cpp
totto82: Fixed

atgeirr (approved):

PR #5823, opened 2024-12-20
Move implementation of WellOperators out of BlackoilWellModel

akva2:
We already have adapters for this, the additional indirection is seems unnecessary to me. As only const access is required, the additional accessors seems like a small price to pay.

### DIFF
--- a/opm/simulators/linalg/WellOperators.hpp
+++ b/opm/simulators/linalg/WellOperators.hpp
@@ -85,22 +85,7 @@ public:
     {
         OPM_TIMEBLOCK(apply);
         for (const auto& well : this->wellMod_) {
-            // Well equations B and C uses only the perforated cells, so need to apply on local vectors
-            const auto& cells = well->cells();
-            x_local_.resize(cells.size());
-            Ax_local_.resize(cells.size());
-
-            for (size_t i = 0; i < cells.size(); ++i) {
-                x_local_[i] = x[cells[i]];
-                Ax_local_[i] = y[cells[i]];
-            }
-
-            well->apply(x_local_, Ax_local_);
-
-            for (size_t i = 0; i < cells.size(); ++i) {
-                // only need to update Ax
-                y[cells[i]] = Ax_local_[i];
-            }
+            this->applySingleWell(x, y, well, well->cells());
         }
     }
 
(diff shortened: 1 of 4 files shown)


akva2:
jenkins build this please

opm/simulators/linalg/WellOperators.hpp
```diff
@@ -84,14 +84,43 @@ class WellModelAsLinearOperator : public LinearOperatorExtra<X, Y>
     void apply(const X& x, Y& y) const override
     {
         OPM_TIMEBLOCK(apply);
-        wellMod_.apply(x, y);
+        for (const auto& well : this->wellMod_) {
+            // Well equations B and C uses only the perforated cells, so need to apply on local vectors
```
atgeirr, line 88: I think we can reduce duplication in a good way by moving the content of this loop to a separate function `applySingleWell()`, then the apply() function of both classes here would not duplicate each other.

opm/simulators/linalg/WellOperators.hpp
```diff
@@ -125,6 +154,10 @@ class WellModelAsLinearOperator : public LinearOperatorExtra<X, Y>
 
 protected:
     const WellModel& wellMod_;
+
+    mutable X x_local_{};
```
atgeirr, line 158: I think `mutable` variables should always be explained in a comment. (Can be lifted from the BlackoilWellModel class.)

atgeirr (commented):
I think this is a good idea, and it should be merged. I believe this is a useful step also if/when we move further toward separating the well-related linear algebra from the well modeling classes.

opm/simulators/linalg/WellOperators.hpp
akva2: good idea. i have added a commit that does this. furthermore, applyscaleadd is also identical so i decided to share them.

opm/simulators/linalg/WellOperators.hpp
akva2: alright, lifted the comment.

akva2:
jenkins build this please

atgeirr (approved):
Great simplifications, now the difference stands out very clearly between the Domain and global version!

PR #5824, opened 2024-12-20
Fix RFT output for shut distributed wells 

vkip:
and also for permanently inactive wells that are split across processors (partition method 3 issue only).

To see the issue for shut distributed wells, run attached example ([shut-distributed-wells-rft-issue.tar.gz](https://github.com/user-attachments/files/18209079/shut-distributed-wells-rft-issue.tar.gz)) as:
>```sh
> mpirun -np 4 flow \
>    --external-partiton=part4.txt \
>    --allow-distributed-wells=true \
>    TEST-RFT-SHUT-DISTRIBUTED-WELL.DATA
>```
and for inactive wells split across processors with partition method 3 (where root does not appear to contain all cells):
>```sh
> mpirun -np 3 flow \
>    --partition-method=3 \
>    --allow-distributed-wells=true \
>    TEST-RFT-INACTIVE-SPLIT-WELL-PARTMETHOD3.DATA
>```
In these cases current master outputs 0 for RFT data for some connections of well INJE01/INJE02 respectively, whereas this PR ensures that all connections get actual values.

The solution here is a quick fix within the current framework, and should at some point be revised to avoid some global communication and adding the well data multiple times for shut distributed and split permanently inactive wells.

### DIFF
--- a/opm/simulators/flow/GenericOutputBlackoilModule.cpp
+++ b/opm/simulators/flow/GenericOutputBlackoilModule.cpp
@@ -407,32 +407,30 @@ template<class FluidSystem>
 void GenericOutputBlackoilModule<FluidSystem>::
 accumulateRftDataParallel(const Parallel::Communication& comm) {
     if (comm.size() > 1) {
-        collectRftMapOnRoot(oilConnectionPressures_, comm);
-        collectRftMapOnRoot(waterConnectionSaturations_, comm);
-        collectRftMapOnRoot(gasConnectionSaturations_, comm);
+        gatherAndUpdateRftMap(oilConnectionPressures_, comm);
+        gatherAndUpdateRftMap(waterConnectionSaturations_, comm);
+        gatherAndUpdateRftMap(gasConnectionSaturations_, comm);
     }
 }
 
 template<class FluidSystem>
 void GenericOutputBlackoilModule<FluidSystem>::
-collectRftMapOnRoot(std::map<std::size_t, Scalar>& local_map, const Parallel::Communication& comm) {
+gatherAndUpdateRftMap(std::map<std::size_t, Scalar>& local_map, const Parallel::Communication& comm) {
 
     std::vector<std::pair<int, Scalar>> pairs(local_map.begin(), local_map.end());
     std::vector<std::pair<int, Scalar>> all_pairs;
     std::vector<int> offsets;
 
-    std::tie(all_pairs, offsets) = Opm::gatherv(pairs, comm, 0);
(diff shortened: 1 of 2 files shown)


vkip:
jenkins build this please

opm/simulators/flow/GenericOutputBlackoilModule.cpp
```diff
@@ -407,32 +407,30 @@ template<class FluidSystem>
...
-            if (auto candidate = local_map.find(key_value.first); candidate != local_map.end()) {
-                const Scalar prev_value = candidate->second;
-                candidate->second = std::max(prev_value, key_value.second);
-            } else {
-                local_map[key_value.first] = key_value.second;
-            }
+    std::tie(all_pairs, offsets) = Opm::allGatherv(pairs, comm);
+
+    // Update maps on all ranks
+    for (auto i=static_cast<std::size_t>(offsets[0]); i<all_pairs.size(); ++i) {
```
atgeirr, line 427: Would it be correct to just use 0 instead of offsets[0]?

atgeirr (commented):
I see no problems with this in itself, and believe it can be merged. However I am not sure I  understand how the problem it fixes arises in the first place, and is therefore a bit reluctant: I would like to understand the problem before accepting a fix. Could you give an explanation?

vkip:
> I see no problems with this in itself, and believe it can be merged. However I am not sure I understand how the problem it fixes arises in the first place, and is therefore a bit reluctant: I would like to understand the problem before accepting a fix. Could you give an explanation?

RFT data for shut wells is currently handled by explicitly adding connection values to the well data (https://github.com/OPM/opm-simulators/blob/119282bd6df8e7617bfa9430c9d736b51568661b/opm/simulators/flow/GenericOutputBlackoilModule.cpp#L444) .  For distributed wells each process will only have non-zero values for connections to owned cells, but when data are collected on the IO rank one copy of the data appears to be collected from each process. Without ensuring that all connections have values on all processors we end up with partially zero RFT data (zero for connection that are not present on the lowest numbered rank touched by the well, I believe).

opm/simulators/flow/GenericOutputBlackoilModule.cpp
vkip: Most likely, but left it at this since I couldn't find it specified (admittedly after a very brief search..)

atgeirr:
I think I have understood the issue now, and that it stems from processes adding connection data for cells they do not own but only have as overlap. Although ideally we should resolve this in a different way, by not adding those connections, the current solution also works, so I will merge.

PR #5826, opened 2024-12-20
Move well linearization code out of BlackoilWellModel to WellConnectionAuxillaryModule

akva2:
Separates code for well linearization to a separate class that implements the relevant interface.

We still model this using is-a in BlackoilWellModel but this can now be changed to an adapter approach if we want to.

Basically this is #5823 but for the nonlinear solver.


akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

atgeirr (approved):

atgeirr:
Looks good, looking forward to the continuation (especially getting rid of the member that is a reference to the derived class).

PR #5828, opened 2024-12-20
Remove nnecessary check isInterior

aritorto:
Iterate over the grid and evaluate initial condition skips non-interior cells, so it seems not necessary to double-check the Interior partition type before adding dofVolume to gridTotalVolume_

Not relevant for the Reference Manual


blattms:
Thanks. Nice catch/cleanup.

PR #5832, opened 2025-01-02
changed: move GasLift code in BlackoilWellModel to separate class

akva2:
move to has-a instead of is-a for modelling in BlackoilWellModel

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -29,7 +29,6 @@
 #include <cstddef>
 #include <map>
 #include <memory>
-#include <optional>
 #include <string>
 #include <vector>
 
@@ -48,6 +47,7 @@
 
 #include <opm/simulators/linalg/matrixblock.hh>
 
+#include <opm/simulators/wells/BlackoilWellModelGasLift.hpp>
 #include <opm/simulators/wells/BlackoilWellModelGeneric.hpp>
 #include <opm/simulators/wells/BlackoilWellModelGuideRates.hpp>
 #include <opm/simulators/wells/GasLiftGroupInfo.hpp>
@@ -104,14 +104,6 @@ template<class Scalar> class WellContributions;
             using RateVector = GetPropType<TypeTag, Properties::RateVector>;
             using GlobalEqVector = GetPropType<TypeTag, Properties::GlobalEqVector>;
             using SparseMatrixAdapter = GetPropType<TypeTag, Properties::SparseMatrixAdapter>;
-            using GasLiftSingleWell = typename WellInterface<TypeTag>::GasLiftSingleWell;
-            using GLiftOptWells = typename BlackoilWellModelGeneric<Scalar>::GLiftOptWells;
-            using GLiftProdWells = typename BlackoilWellModelGeneric<Scalar>::GLiftProdWells;
-            using GLiftWellStateMap =
-                typename BlackoilWellModelGeneric<Scalar>::GLiftWellStateMap;
-            using GLiftEclWells = typename GasLiftGroupInfo<Scalar>::GLiftEclWells;
-            using GLiftSyncGroups = typename GasLiftSingleWellGeneric<Scalar>::GLiftSyncGroups;
-
             using ModelParameters = BlackoilModelParameters<Scalar>;
 
             constexpr static std::size_t pressureVarIndex = GetPropType<TypeTag, Properties::Indices>::pressureSwitchIdx;
@@ -356,8 +348,6 @@ template<class Scalar> class WellContributions;
 
             void addWellPressureEquationsStruct(PressureMatrix& jacobian) const;
 
-            void initGliftEclWellMap(GLiftEclWells &ecl_well_map);
-
             /// \brief Get list of local nonshut wells
             const std::vector<WellInterfacePtr>& localNonshutWells() const
             {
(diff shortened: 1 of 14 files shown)


akva2:
jenkins build this please

opm/simulators/wells/WellInterface.hpp
```diff
@@ -82,12 +84,7 @@ class WellInterface : public WellInterfaceIndices<GetPropType<TypeTag, Propertie
     using SparseMatrixAdapter = GetPropType<TypeTag, Properties::SparseMatrixAdapter>;
     using RateVector = GetPropType<TypeTag, Properties::RateVector>;
     using GasLiftSingleWell = ::Opm::GasLiftSingleWell<TypeTag>;
-    using GLiftOptWells = typename BlackoilWellModel<TypeTag>::GLiftOptWells;
-    using GLiftProdWells = typename BlackoilWellModel<TypeTag>::GLiftProdWells;
     using GLiftEclWells = typename GasLiftGroupInfo<Scalar>::GLiftEclWells;
-    using GLiftWellStateMap = 
-        typename BlackoilWellModel<TypeTag>::GLiftWellStateMap;
-    using GLiftSyncGroups = typename GasLiftSingleWellGeneric<Scalar>::GLiftSyncGroups;
```
bska, line 90: Same here: Maybe we don't need the `GLiftEclWells` type alias here, and if we do, maybe we could retrieve it from the new type instead?

opm/simulators/wells/BlackoilWellModelGasLift_impl.hpp
```diff
@@ -0,0 +1,297 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+// Improve IDE experience
+#ifndef OPM_BLACKOILWELLMODEL_GASLIFT_HEADER_INCLUDED
+#define OPM_BLACKOILWELLMODEL_GASLIFT_IMPL_HEADER_INCLUDED
```
bska, line 25: Maybe we could use the traditional include guards, akin to PR #5835, here?

bska (commented):
This is greatly appreciated.  Other than a couple of "leaky" type aliases this looks good to me.

opm/simulators/wells/BlackoilWellModelGasLift_impl.hpp
akva2: yes, already done locally, was waiting for review before push.

opm/simulators/wells/BlackoilWellModel.hpp
```diff
@@ -110,13 +110,7 @@ template<class Scalar> class WellContributions;
             using RateVector = GetPropType<TypeTag, Properties::RateVector>;
             using GlobalEqVector = GetPropType<TypeTag, Properties::GlobalEqVector>;
             using SparseMatrixAdapter = GetPropType<TypeTag, Properties::SparseMatrixAdapter>;
-            using GasLiftSingleWell = typename WellInterface<TypeTag>::GasLiftSingleWell;
-            using GLiftOptWells = typename BlackoilWellModelGeneric<Scalar>::GLiftOptWells;
-            using GLiftProdWells = typename BlackoilWellModelGeneric<Scalar>::GLiftProdWells;
-            using GLiftWellStateMap =
-                typename BlackoilWellModelGeneric<Scalar>::GLiftWellStateMap;
             using GLiftEclWells = typename GasLiftGroupInfo<Scalar>::GLiftEclWells;
-            using GLiftSyncGroups = typename GasLiftSingleWellGeneric<Scalar>::GLiftSyncGroups;
```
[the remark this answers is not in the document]
akva2: it's needed in the impl file. But as it is a single call site I moved to a local type alias at call site instead.

opm/simulators/wells/WellInterface.hpp
akva2: We do need the type as it's used in various well testing methods etc and I don't want to make WellInterface rely on BlackoilWellModel level stuff.

akva2:
jenkins build this please

opm/simulators/wells/WellInterface.hpp
bska: > it's used in various well testing methods etc

Hm, pity.  I guess we'll just have to be really vigilant that the two aliases don't get out of synch.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1235,11 +1244,12 @@ namespace Opm {
     {
 
         DeferredLogger local_deferredLogger;
-        if (this->glift_debug) {
+/*        if (this->glift_debug) {
             const std::string msg = fmt::format(
                 "assemble() : iteration {}" , iterationIdx);
             this->gliftDebug(msg, local_deferredLogger);
         }
+*/
```
bska, line 1252: Just out of curiosity: Why do we disable this debugging facility?

akva2: nice catch. that was supposed to be a temporary while refactoring but things got a little muddy. I had some follow-ups for this but they require some somewhat unrelated reorganizations. I now realize I need to do it here tough, so there's an additional commit in the PR now, moving some things around.

akva2:
jenkins build this please

bska (approved):

PR #5833, opened 2025-01-02
Some janitoring

akva2:
- use standard algorithm instead of loop
- format some code as everything else around it is formatted (braces, spaces)..

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1166,21 +1166,19 @@ namespace Opm {
                           DeferredLogger& deferred_logger) const
     {
         // Finding the location of the well in wells_ecl
-        const int nw_wells_ecl = this->wells_ecl_.size();
-        int index_well_ecl = 0;
-        for (; index_well_ecl < nw_wells_ecl; ++index_well_ecl) {
-            if (well_name == this->wells_ecl_[index_well_ecl].name()) {
-                break;
-            }
-        }
+        const auto it = std::find_if(this->wells_ecl_.begin(),
+                                     this->wells_ecl_.end(),
+                                     [&well_name](const auto& w)
+                                     { return well_name == w.name(); });
         // It should be able to find in wells_ecl.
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1166,21 +1166,18 @@ namespace Opm {
...
         // It should be able to find in wells_ecl.
-        if (index_well_ecl == nw_wells_ecl) {
+        if (it == this->wells_ecl_.end()) {
             OPM_DEFLOG_THROW(std::logic_error,
                              fmt::format("Could not find well {} in wells_ecl ", well_name),
                              deferred_logger);
         }
 
-        return this->createWellPointer(index_well_ecl, report_step);
+        return this->createWellPointer(std::distance(this->wells_ecl_.begin(), it), report_step);
```
bska, line 1180: Tiny nit: `std::distance()` returns an `iterator_traits<>::difference_type`, which for iterators into `std::vector<>` is almost always `std::ptrdiff_t` but `createWellPointer()` [takes an `int`](https://github.com/OPM/opm-simulators/blob/df0820f67ac326b0130877e9767a3af2ba39c22c/opm/simulators/wells/BlackoilWellModel_impl.hpp#L1113).  While this call won't trigger "signed vs. unsigned" comparison warnings, there could nevertheless be a type conversion here (e.g., `long` ➡️ `int`).  Might it be better to write this as
```C++
const auto idx = static_cast<int>(std::distance(begin, it));
return this->createWellPointer(idx, report_step);
```
instead in order to be explicit about the type conversion?

bska (commented):
Looks good to me, except for the implicit type conversion in the call to `createWellPointer()`.

opm/simulators/wells/BlackoilWellModel_impl.hpp
akva2: sure.

akva2:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  This looks good to me and I'll merge into master.

PR #5839, opened 2025-01-03
add test for gpu brineco2pvt

multitalentloes:


atgeirr (approved):
Looks fine, can be merged when upstream is done.

atgeirr:
jenkins build this please

tests/gpuistl/test_gpuPvt.cu
```diff
@@ -110,6 +115,11 @@ __global__ void co2GasPvtInternalEnergy(GpuViewCo2Pvt gpuViewCo2Pvt, Evaluation*
     *result = gpuViewCo2Pvt.internalEnergy<Evaluation>(1, *temp, *pressure, Evaluation(0.4), Evaluation(0.0)).value();
 }
 
+// Kernel using a BrineCo2Pvt object on a GPU
+__global__ void brineCo2PvtInternalEnergy(GpuViewBrineCo2Pvt gpuViewBrineCo2Pvt, Evaluation* temp, Evaluation* pressure, Evaluation* rs, Evaluation* saltConcentration, double* result) {
+    *result = gpuViewBrineCo2Pvt.internalEnergy<Evaluation>(1, *temp, *pressure, *rs, *saltConcentration).value();
+}
```
bska, line 121: I realise that the rest of the code does this as well and that you're "just" following the established pattern, but I would have thought that the compiler would be able to deduce the `Evaulation` type from the parameter list without you having to pass an explicit template argument in the call to `internalEnergy<>()`.  Does the following statement not work/compile?
```C++
 *result = gpuViewBrineCo2Pvt.internalEnergy(1, *temp, *pressure, *rs, *saltConcentration).value();
```

multitalentloes: You are right, I could actually remove every `<Evaluation>` in the file... Things are a bit cleaner now, good catch

multitalentloes:
Jenkins build this please

PR #5843, opened 2025-01-07
Use new linearizer for blackoil + thermal

totto82:


totto82:
jenkins build this please

totto82:
I dont see any reasons for not merging this. It gives a decent speedup.

PR #5849, opened 2025-01-08
Fix BlackoilWellModel::hasWell mixups

akva2:
this to avoid confusion as there is already a hasWell in BlackoilWellModelGeneric which queries the schedule.
1) the one in BlackoilWellModelGeneric is renamed to hasLocalWell, as this queries list of all wells having connections on rank
2) the one in BlackoilWellModel is renamed to hasOpenWell, as this queries list of all open wells having connection on rank

Also moves the definition of hasOpenWell to the generic class

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.hpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.hpp
@@ -105,7 +105,7 @@ public:
     bool wellsActive() const;
 
     //! \brief Returns true if well is defined and has connections on current rank.
-    bool hasWell(const std::string& wname) const;
+    bool hasLocalWell(const std::string& wname) const;
 
     //! \brief Returns true if well is defined, open and has connections on current rank.
     bool hasOpenLocalWell(const std::string& well_name) const;
--- a/opm/simulators/wells/BlackoilWellModelGeneric.hpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.hpp
@@ -103,7 +103,12 @@ public:
 
     /// return true if wells are available in the reservoir
     bool wellsActive() const;
+
+    //! \brief Returns true if well is defined and has connections on current rank.
     bool hasWell(const std::string& wname) const;
+
+    //! \brief Returns true if well is defined, open and has connections on current rank.
+    bool hasOpenLocalWell(const std::string& well_name) const;
 
     /// return true if network is active (at least one network well in prediction mode)
     bool networkActive() const;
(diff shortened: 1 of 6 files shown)


akva2:
jenkins build this please

akva2:
hmm. there is more confusion. some methods use the 'local' term to refer to schedule wells, some use local to refer to wells active on current process in report step.. i'll amend this with more work, drafting for now.

GitPaean:
Did not check in details, but is https://github.com/OPM/opm-simulators/pull/4616 relevant to this in any aspect?

GitPaean:
> yes it's the same thing.

good to know, and I am glad it is taken care of.

atgeirr:
> some methods use the 'local' term to refer to schedule wells, some use local to refer to wells active on current process in report step.

Not exactly sure what you mean by the first here (schedule wells)? But I agree there is some unfortunate overuse of the terms local/global:
 - Local meaning "on this MPI process/rank" (this would be your second)
 - Local meaning "on this NLDD domain" (although I think we renamed many of those to something with Domain instead)

Not relevant for the current setting relating to wells:
 - Global index meaning linearized logical cartesian index of a cell, i.e. for cell (i, j, k) "global index" = i + nx * j + nx * ny * k

I think I would prefer for global/local to ALWAYS be interpreted with regards to MPI process/rank, does that sound good?

akva2:
schedule wells = all wells defined at some point in the schedule.

atgeirr:
Just to add to the global/local usage examples: In the models part of the code, you have globalSpaceIndex which is the cell index typically (on this MPI partition, so local by the criteria I would like to use...) and local indices which are local numberings of the cells' neighbors (i.e. the order of intersection in the dune-grid sense). Have to keep the context in mind when coding these things!

akva2:
Should be cleared up now - including my misunderstanding of what is stored in ecl_wells_.

akva2:
jenkins build this please

opm/simulators/wells/BlackoilWellModelGeneric.hpp
```diff
@@ -103,7 +103,12 @@ class BlackoilWellModelGeneric
 
     /// return true if wells are available in the reservoir
     bool wellsActive() const;
-    bool hasWell(const std::string& wname) const;
+
+    //! \brief Returns true if well is defined and has connections on current rank.
+    bool hasLocalWell(const std::string& wname) const;
+
+    //! \brief Returns true if well is defined, open and has connections on current rank.
+    bool hasOpenWell(const std::string& wname) const;
```
atgeirr, line 111: Maybe call this `hasOpenLocalWell()`.

akva2: sure

akva2:
jenkins build this please

atgeirr:
Looks good to me now, merging!

PR #5851, opened 2025-01-08
Compositional simulator expanded to three phases

svenn-t:
Added a "dummy"/immiscible water phase to the compositional simulator using the three-phase fluid system in https://github.com/OPM/opm-common/pull/4410. An additional equation for the water phase has been added with water saturation as primary variable. Water phase properties are set using standard keywords as in black oil simulations.


GitPaean:
Sorry for the delay. But I will review the PR now.  Due to the change in master branch, there is some update needed in `opm-simulators/opm/simulators/flow/FIPContainer.cpp`.

svenn-t:
> Sorry for the delay. But I will review the PR now. Due to the change in master branch, there is some update needed in `opm-simulators/opm/simulators/flow/FIPContainer.cpp`.

I've pushed a new update with changes in the mentioned file.

opm/simulators/flow/FIPContainer.cpp
```diff
@@ -438,16 +438,27 @@ INSTANTIATE_TYPE(double)
...
+#define INSTANTIATE_COMP_TWOPHASE(NUM) \
+    template<class T> using GFS##NUM = GenericOilGasWaterFluidSystem<T, NUM, false>; \
+    template class FIPContainer<GFS##NUM<double>>;
+
+INSTANTIATE_COMP_TWOPHASE(2)
+INSTANTIATE_COMP_TWOPHASE(3)
+INSTANTIATE_COMP_TWOPHASE(4)
+INSTANTIATE_COMP_TWOPHASE(5)
+INSTANTIATE_COMP_TWOPHASE(6)
+INSTANTIATE_COMP_TWOPHASE(7)
```
akva2, line 462: make a new macro
```
 #define INSTANTIATE_COMP(NUM) \
    INSTANTIATE_COMP_THREEPHASE(NUM) \
   INSTANTIATE_COMP_TWOPHASE(NUM)
```
less duped lines, easier maintenance (you can still call the threephase one for 0).

svenn-t: Done!

opm/models/ptflash/flashintensivequantities.hh
```diff
@@ -216,19 +219,24 @@ public:
...
         Evaluation L = fluidState_.L();
-        Evaluation So = Opm::max((L * Z_L / ( L * Z_L + (1 - L) * Z_V)), 0.0);
-        Evaluation Sg = Opm::max(1 - So, 0.0);
-        Scalar sumS = Opm::getValue(So) + Opm::getValue(Sg);
+        Evaluation So = Opm::max((1 - Sw) * (L * Z_L / ( L * Z_L + (1 - L) * Z_V)), 0.0);
+        Evaluation Sg = Opm::max(1 - So - Sw, 0.0);
+        Scalar sumS = Opm::getValue(So) + Opm::getValue(Sg) + Opm::getValue(Sw);
         So /= sumS;
         Sg /= sumS;
+        Sw /= sumS;
```
GitPaean, line 233: Just leaving some notes for later,

should we update the water saturation in the fluidState_ again here?

GitPaean: Please fix here so that if we need to re-normalize the saturation, then Sw should be also updated.  Maybe it should be updated both for `fluidState_` and `priVars`, please test to figure out. 

And also, since `Sw` is the primary variable, make sure the derivative for it is correct.

svenn-t: I moved the update of water saturation in `fluidState_` after the normalization. I looked at the derivatives and they seem to be correct. I'm unsure if anything needs to be corrected wrt to `priVars`.

GitPaean:
The upstream PR https://github.com/OPM/opm-common/pull/4410 has been merged. I am getting this PR in now also.

PR #5854, opened 2025-01-08
Add option for solving the energy equation sequentially 

totto82:
- adds flow_blackoil_temp simulator that solves energy equation sequentially 
- activated by the TEMP keyword (THERMAL is still used for fully implicit)


opm/simulators/flow/TemperatureModel.hpp
```diff
@@ -0,0 +1,503 @@
...
+            }
+        }
+
+#warning fix this
+        // Comunicate the residual
+        std::vector<EnergyVector> tmp;
+        tmp.resize(1);
+        tmp[0] = this->energyVector_;
+        auto handle = VectorVectorDataHandle<GridView, std::vector<EnergyVector>>(tmp,simulator_.gridView());
+        simulator_.gridView().communicate(handle, Dune::InteriorBorder_All_Interface,Dune::ForwardCommunication);
```
totto82, line 405: @blattms I need a VectorDataHandle not a VectorVectorDataHandle. Do I need to make my own Handle for this or can I use an existing one.

blattms: there seems to be `GridCommHandleGhostSync` in opm/models/parallel/gridcommhandles.hh that does this, hopefully.

totto82: Thanks. I did the trick.

totto82:
jenkins build this opm-common=4411 please

totto82:
@akva2 I dont build with HDF5 on the machine I am working on so I have being misusing jenkins for some testing without any sucsess. It seems to complain about some missing serializeOp(). I have tried to add them accordingly based on my understanding, but it keeps complaining. 1) Any hints of what I am missing? 2) Would adding a test for the temperatureModel serialization in test_RestartSerialization help detecting the issue that causes the failure related to HDF5?

akva2:
Well, you are on to it, but fundamentally you need to add serializeOp to the types your are serializing, in particular the EnergyVectors in storage1_ and the intQuants. But first a fundamental question:

I'm surprised that you need to serialize any of this. The only things you need to serialize are things that will not be recreated when we start a new report step. It looks to me like both of these are recreated from simulator state in beginTimeStep(), hence none of the serialization should be required, and the "fix" is to remove the serialization of the TemperatureModel in FlowProblem

totto82:
> and the "fix" is to remove the serialization of the TemperatureModel in FlowProblem

Does that mean that I need to remove these
>     template <class Restarter>
>     void serialize(Restarter&)
>     { /* not implemented */ }

>    template <class Restarter>
>   void deserialize(Restarter&)
 >  { /* not implemented */ }

The serializeOp is already commented
```
    //template<class Serializer>
    //void serializeOp(Serializer& serializer)
    //{
    //    serializer(intQuants_);
    //    serializer(storage1_);
    //}
```

akva2:
only serializeOp is used with flow. the other serialize/deserialize combo is an opm-models mechanism that is not in use in flow, but you need the functions there for the code to compile. 

you can just remove the implementation of serializeOp. but as I said you have to remove the serializer call for temperature model in FlowProblem. that's the call site where you are telling it to call serializeOp.

totto82:
Arg.. I didn't notice that I had added it in FlowProblem. Thanks!

totto82:
jenkins build this opm-common=4411 please

totto82:
jenkins build this opm-common=4411 please

totto82:
jenkins build this opm-common=4411 please

opm/models/blackoil/blackoilenergymodules.hh
```diff
@@ -458,62 +458,128 @@ class BlackOilEnergyIntensiveQuantities<TypeTag, false>
...
     }
 
-    void updateEnergyQuantities_(const ElementContext&,
-                                 unsigned,
-                                 unsigned,
-                                 const typename FluidSystem::template ParameterCache<Evaluation>&)
-    {}
 
-    const Evaluation& rockInternalEnergy() const
+    void updateEnergyQuantities_(const ElementContext& elemCtx,
```
atgeirr, line 558: Make this function call the other one instead of duplicating.

opm/models/discretization/common/fvbaseadlocallinearizer.hh
```diff
@@ -68,12 +68,13 @@ template<class TypeTag>
 struct Evaluation<TypeTag, TTag::AutoDiffLocalLinearizer>
 {
 private:
-    static constexpr unsigned numEq = getPropValue<TypeTag, Properties::NumEq>();
+    static const unsigned numEq = getPropValue<TypeTag, Properties::NumEq>();
```
atgeirr: Should be constexpr?

opm/models/blackoil/blackoilvariableandequationindices.hh
```diff
@@ -93,6 +94,8 @@ struct BlackOilVariableAndEquationIndices
     static constexpr int numEq = numPhases + numSolvents + numExtbos + numPolymers +
                                  numEnergy + numFoam + numBrine + numMICPs;
 
+    static constexpr int numDerivatives = numEq + enableTemperature;
```
atgeirr: Add a comment explaining the idea (you can evaluate intensive quantities with a larger number of derivatives than you have equations).

opm/models/blackoil/blackoilenergymodules.hh
```diff
@@ -712,18 +778,60 @@ class BlackOilEnergyExtensiveQuantities<TypeTag, false>
...
-                             const unsigned& /*exIdx*/,
-                             const IntensiveQuantities& /*inIq*/,
-                             const IntensiveQuantities& /*exIq*/,
-                             const FluidState& /*inFs*/,
-                             const FluidState& /*exFs*/,
-                             const Scalar& /*inAlpha*/,
-                             const Scalar& /*outAlpha*/,
-                             const Scalar& /*faceArea*/)
-    {}
+    static void updateEnergy(Evaluation& energyFlux,
```
atgeirr, line 897: This function is identical to the one in the `<TypeTag, true>` specialization. Use either a free function or call the other one.

opm/models/blackoil/blackoiltwophaseindices.hh
```diff
@@ -98,6 +98,7 @@ struct BlackOilTwoPhaseIndices
     static constexpr int numEq = numPhases + numSolvents + numExtbos + numPolymers +
                                  numEnergy + numFoam + numBrine + numMICPs;
 
+    static constexpr int numDerivatives = numEq; //TODO + enableTemperature;
```
atgeirr: Why not add enableTemperature here? Does the TEMP option require 3-phase?

opm/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -1070,14 +1070,51 @@ private:
...
-        else if constexpr (enableTemperature) {
-            return problem.temperature(globalDofIdx, /*timeIdx*/ 0);
-        }
         else {
-            return FluidSystem::reservoirTemperature();
+            return problem.temperature(globalDofIdx, /*timeIdx*/ 0);
         }
     }
 
+    Scalar microbialConcentration_() const
```
atgeirr: These functions were removed because they were never called (same-name functions in IQ class used instead). Do not add them back...

opm/simulators/flow/GenericTemperatureModel_impl.hpp
```diff
@@ -0,0 +1,237 @@
...
+template<class Grid, class GridView, class DofMapper, class Stencil, class FluidSystem, class Scalar>
+void GenericTemperatureModel<Grid,GridView,DofMapper,Stencil,FluidSystem,Scalar>::
+doInit(std::size_t numGridDof)
+{
+    doTemp_ = eclState_.getSimulationConfig().isTemp();
+
+    temperature_.resize(numGridDof);
+    energyVector_.resize(numGridDof);
+    // allocate matrix for storing the Jacobian of the temperature residual
+    energyMatrix_ = std::make_unique<EnergyMatrix>(numGridDof, numGridDof, EnergyMatrix::random);
```
atgeirr, line 120: If the nonzero structure is assumed to be the same as in the flow problem, it is simpler and faster to reuse it as in PressureTransferPolicy::createCoarseLevelSystem().

opm/simulators/flow/GenericTemperatureModel_impl.hpp
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+    energyMatrix_->endindices();
+}
+
+template<class Grid, class GridView, class DofMapper, class Stencil, class FluidSystem, class Scalar>
+bool GenericTemperatureModel<Grid,GridView,DofMapper,Stencil,FluidSystem,Scalar>::
+linearSolve_(const EnergyMatrix& M, EnergyVector& x, EnergyVector& b)
+{
+    x = 0.0;
```
atgeirr, line 165: Introduce a parameter for the temperature linear solve setup (LinearSolverSequentialTemperature) and unify serial and parallel solver setup. Also, store solver as member to avoid recreating every timestep, taking advantage of update(). I can help with this!

opm/simulators/flow/GenericTemperatureModel_impl.hpp
```diff
@@ -0,0 +1,237 @@
...
+#endif
+}
+
+template<class Grid, class GridView, class DofMapper, class Stencil, class FluidSystem, class Scalar>
+void GenericTemperatureModel<Grid,GridView,DofMapper,Stencil,FluidSystem,Scalar>::
+syncOverlap_()
+{
+#if HAVE_MPI
+    // syncronize the solution on the ghost and overlap elements
+    using GhostSyncHandle = GridCommHandleGhostSync<Dune::FieldVector<Scalar, 1>,
```
atgeirr: We have no ghost elements, only overlap. Is this an attempt to future-proof?

opm/simulators/flow/GenericTemperatureModel_impl.hpp
```diff
@@ -0,0 +1,237 @@
...
+        // return the result of the solver
+        return result.converged;
+#if HAVE_MPI
+    }
+#endif
+}
+
+template<class Grid, class GridView, class DofMapper, class Stencil, class FluidSystem, class Scalar>
+void GenericTemperatureModel<Grid,GridView,DofMapper,Stencil,FluidSystem,Scalar>::
+syncOverlap_()
```
atgeirr: Please add a comment explaining the need for this method, and how it is implemented.

opm/simulators/flow/TemperatureModel.hpp
```diff
@@ -0,0 +1,480 @@
...
+
+    bool converged(const int iter) {
+        const unsigned int numCells = simulator_.model().numTotalDof();
+        Scalar maxNorm = 0.0;
+        Scalar sumNorm = 0.0;
+        for (unsigned globI = 0; globI < numCells; ++globI) {
+            maxNorm = max(maxNorm, std::abs(this->energyVector_[globI]));
+            sumNorm += std::abs(this->energyVector_[globI]);
+        }
+        Scalar scaling = 1.0/4.182e1; //// converting J -> RM3 (entalpy / (cp * deltaK * rho) assuming change of 1e-5K of water
```
atgeirr: Too much magic... Use named constexpr Scalars and use the unit system where appropriate.

opm/simulators/flow/FlowBaseProblemProperties.hpp
```diff
@@ -95,6 +96,12 @@ template <class TypeTag>
 struct TracerModel<TypeTag, TTag::FlowBaseProblem>
 { using type =  ::Opm::TracerModel<TypeTag>; };
 
+template<class TypeTag, class MyTypeTag>
+struct TemperatureModel {  using type = UndefinedProperty; };
+template <class TypeTag>
+struct TemperatureModel<TypeTag, TTag::FlowBaseProblem>
+{ using type =  ::Opm::TemperatureModel<TypeTag>; };
```
atgeirr, line 113: Avoid ::Opm::, already inside Opm. (Also above)

opm/simulators/flow/GenericTemperatureModel_impl.hpp
blattms: It is as the name seems to be misleading. Ghost only appears in the name. The handle does not distinguish where it scatters and gathers. That is governed by the argument line 230 (`Dune::InteriorBorder_All_Interface`)

opm/simulators/flow/GenericTemperatureModel_impl.hpp
totto82: This is only done initially

totto82:
Rebased and updated. Now depends on https://github.com/OPM/opm-common/pull/4825

opm/simulators/flow/FlowBaseProblemProperties.hpp
totto82: I think we need ::Opm:: here

opm/simulators/flow/GenericTemperatureModel_impl.hpp
totto82: I kept this as is. Perhaps you could create a PR afterward?

opm/models/blackoil/blackoilenergymodules.hh
totto82: Comment deplicated

opm/simulators/flow/TemperatureModel.hpp
totto82: The same hard-coding is used for the standard energy equations. Moved both to the parameter

opm/models/blackoil/blackoilenergymodules.hh
```diff
@@ -506,6 +510,108 @@ protected:
...
+    }
+
+    void updateEnergyQuantities_(const Problem& problem,
+                                 const unsigned globalSpaceIdx,
+                                 const unsigned timeIdx)
+    {
+        auto& fs = asImp_().fluidState_;
+
+        // compute the specific enthalpy of the fluids, the specific enthalpy of the rock
+        // and the thermal condictivity coefficients
```
atgeirr: Typo: condictivity->conductivity

opm/models/blackoil/blackoilenergymodules.hh
atgeirr: I do not understand what you mean. My comment is still true as far as I can tell?

opm/models/blackoil/blackoilenergymodules.hh
```diff
@@ -779,6 +885,97 @@ public:
...
+            // depends on the solution.
+            H = 1.0/(1.0/inH + 1.0/exH);
+        }
+        else
+            H = 0.0;
+
+        energyFlux = deltaT * (-H/faceArea);
+    }
+
+    void updateEnergy(const ElementContext&,
```
atgeirr, line 923: These should also be implemented by calling the same function from all the overloads (as well as from the fully implicit version), assuming they are supposed to be identical.

opm/models/blackoil/blackoilvariableandequationindices.hh
```diff
@@ -92,6 +93,10 @@ struct BlackOilVariableAndEquationIndices
     static constexpr int numEq = numPhases + numSolvents + numExtbos + numPolymers +
                                  numEnergy + numFoam + numBrine;
 
+    //! For seqential implicit approches we evaluate the intensive quantities with a larger number of derivaties
```
atgeirr: Typo derivaties->derivatives. Also, append "...than equations" to the sentence.

Also in the twophase and onephase classes.

opm/simulators/flow/GenericTemperatureModel_impl.hpp
```diff
@@ -0,0 +1,237 @@
...
+#include <dune/istl/schwarz.hh>
+#include <dune/istl/preconditioners.hh>
+#include <dune/istl/schwarz.hh>
+
+#include <opm/common/OpmLog/OpmLog.hpp>
+
+#include <opm/grid/CpGrid.hpp>
+
+#include <opm/input/eclipse/EclipseState/EclipseState.hpp>
+//#include <opm/input/eclipse/EclipseState/Tables/TemperatureVdTable.hpp>
```
atgeirr: Cleanup commented lines.

opm/simulators/flow/GenericTemperatureModel_impl.hpp
atgeirr: Ok we can do that afterwards.

opm/simulators/flow/GenericTemperatureModel_impl.hpp
atgeirr: Still relevant.

opm/simulators/flow/FlowBaseProblemProperties.hpp
atgeirr: Is it because you use the same name for the property as the class name?

opm/simulators/flow/TemperatureModel.hpp
```diff
@@ -0,0 +1,480 @@
...
+            solveAndUpdate();
+        }
+    }
+
+    void solveAndUpdate() {
+        const unsigned int numCells = simulator_.model().numTotalDof();
+        EnergyVector dx(numCells);
+        bool converged = this->linearSolve_(*this->energyMatrix_, dx, this->energyVector_);
+        if (!converged) {
+            OpmLog::warning("### Temp model: Linear solver did not converge. ###");
```
atgeirr: Remove the `###`s, and make sure this is only logged from rank 0.

opm/simulators/flow/TemperatureModel.hpp
```diff
@@ -0,0 +1,480 @@
...
+        } else {
+            for (unsigned globI = 0; globI < numCells; ++globI) {
+                this->temperature_[globI] -= std::clamp(dx[globI][0], Scalar(-10.0), Scalar(10.0));
```
atgeirr: This 10 degree limit seems a bit arbitrary. Why is it necessary to clamp? If it must be kept, put the 10.0 in a named constant. (Later we can make it a runtime option if necessary.)

opm/simulators/flow/GenericTemperatureModel_impl.hpp
totto82: What do you want me to write? Isn't this standard Dune parallel linear solver stuff?

opm/simulators/flow/FlowBaseProblemProperties.hpp
totto82: Do you want me to rename either the property or the class name? I guess this goes for other "models" as well.

atgeirr: No keep it for now. I think it is unfortunate to have the same, but we should not rename randomly. Let's agree on a naming scheme and apply it consistently later.

opm/simulators/flow/GenericTemperatureModel_impl.hpp
atgeirr: I would like a comment stating which variables are communicated, on which entities. (I think the one I commented on originally was a bit more convoluted?)

opm/models/blackoil/blackoilenergymodules.hh
```diff
@@ -779,6 +885,68 @@ public:
...
+                     exFs,
+                     inAlpha,
+                     outAlpha,
+                     faceArea);
+    }
+
+    void updateEnergy(const ElementContext&,
+                      unsigned,
+                      unsigned)
+    { } // Old interface still used output code for fluxes. But energy flux is not used. i.e. do nothing
```
atgeirr, line 926: What is meant here? It is not obvious to me this and the following functions should be empty.

opm/models/blackoil/blackoilenergymodules.hh
```diff
@@ -789,17 +957,17 @@ class BlackOilEnergyExtensiveQuantities<TypeTag, EnergyModules::NoTemperature>
 
 public:
     template<class FluidState>
-    static void updateEnergy(Evaluation& /*energyFlux*/,
```
atgeirr, line 792: This will cause compilation warnings, please fix.

opm/models/utils/basicproperties.hh
```diff
@@ -79,6 +79,11 @@ struct Scalar { using type = UndefinedProperty; };
 template<class TypeTag, class MyTypeTag>
 struct NumEq { using type = UndefinedProperty; };
 
+//! Number of equations in the system of PDEs
```
atgeirr: Wrong comment.

opm/simulators/flow/BlackoilModel_impl.hpp
```diff
@@ -1161,7 +1161,7 @@ getMaxCoeff(const unsigned cell_idx,
     }
 
     if constexpr (has_energy_) {
-        B_avg[contiEnergyEqIdx] += 1.0 / (4.182e1); // converting J -> RM3 (entalpy / (cp * deltaK * rho) assuming change of 1e-5K of water
+        B_avg[contiEnergyEqIdx] += 1.0;
```
atgeirr, line 1164: What is the intention behind this? Or maybe rather, what was the original idea, that you abandon now. I assume this is related to the question of what the tolerance means. Maybe a good idea with a comment.

opm/simulators/flow/BlackoilModelParameters.hpp
```diff
@@ -55,11 +55,13 @@ struct ToleranceMb { static constexpr Scalar value = 1e-7; };
 template<class Scalar>
 struct ToleranceMbRelaxed { static constexpr Scalar value = 1e-6; };
 
+//TODO change to a simpler number with fewer digits
+//converting J -> RM3 (entalpy / (cp * deltaK * rho) assuming change of 1e-5K of water 
 template<class Scalar>
-struct ToleranceEnergyBalance { static constexpr Scalar value = 1e-7; };
+struct ToleranceEnergyBalance { static constexpr Scalar value = 1e-7*41.82; };
```
atgeirr, line 61: Make this number a named constexpr, also it is not clear to me that it belongs here rather than in the scaling of the equation?

opm/simulators/flow/GenericOutputBlackoilModule.cpp
```diff
@@ -128,7 +128,7 @@ GenericOutputBlackoilModule(const EclipseState& eclState,
                             RSTConv::LocalToGlobalCellFunc globalCell,
                             const Parallel::Communication& comm,
                             bool enableEnergy,
-                            bool enableTemperature,
+                            bool constantTemperature,
```
atgeirr, line 131: Is `constantTemperature` really what you want here? What about the sequential implicit case? Also in the hpp file.

opm/simulators/flow/MainDispatchDynamic.cpp
```diff
@@ -131,6 +133,11 @@ int Opm::Main::dispatchDynamic_()
         return this->runThermal(phases);
     }
 
+    // Blackoil case with temperature
+    else if (phases.size() == 4 && temp) {
```
atgeirr, line 137: I would assume that phases.size() would be 3 not 4? Or would the energy bit be set in the Runspec Phases object? If so that is a bit counterintuitive.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -2728,7 +2728,6 @@ namespace Opm
                 result += Base::restrictEval(cq_r_thermal);
             }
         }
-
```
atgeirr, line 2731: No other change in this file, better to leave out.

atgeirr (commented):
I have now reviewed the complete PR, and while there are a few issues they are relatively minor. Good work!

opm/models/blackoil/blackoilenergymodules.hh
totto82: ` void captureLocalFluxData()` still uses the old linearizer. This should be changed at some point, as we already have computed and can store the fluxes, so there is no need to recompute them.  This code does not output the energy flux, so no update is needed.

opm/simulators/flow/BlackoilModel_impl.hpp
totto82: Some time ago, we used the same tolerance on all equations. Since mass and energy are not the same, we scaled them to make them more comparable.  We now set the tolerance individually and therefore do not need to scale them. I instead scale the default tolerance as a temporary solution to avoid test failures. I plan to make a new PR that changes the default to a more reasonable number. See comment below.

opm/simulators/flow/BlackoilModelParameters.hpp
totto82: As written above, I think we should change the default to 1e-6 (or 1e-5). However, since that would change the results, I would rather do that in another PR. I will make it immediately after this PR is made.

opm/simulators/flow/GenericOutputBlackoilModule.cpp
totto82: Is it correct? But note that enableEnergy here is `enable_energy = sequential implicit || fullyImplicit`

opm/simulators/flow/MainDispatchDynamic.cpp
totto82: Yes, agree it is counterintuitive. I did == 3 first and the code didn't behave as I expected. But I dont think I would like to fix this in this PR.

opm/simulators/flow/GenericOutputBlackoilModule.cpp
atgeirr: I see, it was a bit confusing to have enable_energy mean either sequential or fully implicit energy, whereas it used to mean only fully implicit, but it is ok. I now think it is correct.

PR #5855, opened 2025-01-08
Avoid setWsolvent call for wells shut by WCYCLE 

vkip:


opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -987,24 +987,25 @@ updateWsolvent(const Group& group,
...
     for (const std::string& groupName : group.groups()) {
         const Group& groupTmp = schedule_.getGroup(groupName, reportStepIdx);
-        setWsolvent(groupTmp, reportStepIdx, wsolvent);
+        setWsolvent(groupTmp, reportStepIdx, wellState, wsolvent);
     }
 
     for (const std::string& wellName : group.wells()) {
         const auto& wellTmp = schedule_.getWell(wellName, reportStepIdx);
-        if (wellTmp.getStatus() == Well::Status::SHUT)
+        if (!wellState.has(wellName) || wellState.well(wellName).status == WellStatus::SHUT)
```
akva2, line 1008: i think it should be enough to check if wellName is in generic_wells_

vkip: yeah, you're right, I'm just too lazy to look up values in vectors.. Maybe well_container_generic_ should be a WellContainer?

vkip: thanks, will do

vkip:
jenkins build this please

opm/simulators/wells/BlackoilWellModelGeneric.cpp
akva2: https://github.com/OPM/opm-simulators/pull/5849 got around to it, so this will be a simple 
```c++
if (!hasOpenWell(foo)) {
    continue;
}
``` 

which makes intent very readable.

opm/simulators/wells/BlackoilWellModelGeneric.hpp
```diff
@@ -105,6 +105,9 @@ class BlackoilWellModelGeneric
     bool wellsActive() const;
     bool hasWell(const std::string& wname) const;
 
+    // return true if well exists and is not shut
```
atgeirr, line 108: This should be more specific, in light of the discussion in https://github.com/OPM/opm-simulators/pull/5849, I assume this is just on this process? If so, we should call the method something with Local. But I am not sure exactly what role this plays compared to the other has...Wells() methods?

vkip: I'll use `hasOpenLocalWell` from https://github.com/OPM/opm-simulators/pull/5849 once that is merged.

vkip:
jenkins build this please

akva2 (approved):

PR #5856, opened 2025-01-09
Bugfix/thread spawn

michal-toth:
Move the function setting the maximum number of threads to an earlier place, before the Deck is read which uses `#pragma parallel for`. This allows us to simplify the getNumThreads() in Main.hpp to just return the current maximum number of threads, so it no longer needs to guess the future setting.

In addition, `--threads-per-process=-1` is fixed to let OMP set the number of threads automatically, it is no longer rewritten to 2. The default number of threads is changed to 2 explicitly (it was -1 that got rewritten to 2). The change of defaults will require a change in the reference manual.

### DIFF
--- a/opm/simulators/flow/FlowMain.hpp
+++ b/opm/simulators/flow/FlowMain.hpp
@@ -310,10 +310,10 @@ namespace Opm {
                 }
             }
 
-            // We are not limiting this to the number of processes
-            // reported by OpenMP as on some hardware (and some OpenMPI
-            // versions) this will be 1 when run with mpirun
-            omp_set_num_threads(threads);
+            // Requesting -1 thread will let OMP automatically deduce the number
+            if (requested_threads != -1) {
+                omp_set_num_threads(threads);
+            }
 #endif
 
             using TM = GetPropType<TypeTag, Properties::ThreadManager>;
--- a/opm/simulators/flow/FlowMain.hpp
+++ b/opm/simulators/flow/FlowMain.hpp
@@ -156,6 +156,9 @@ namespace Opm {
                 }
             }
 
+            // set the maximum limit on OMP threads
+            setMaxThreads();
+
             return status;
         }
 
@@ -278,6 +281,11 @@ namespace Opm {
             mpi_rank_ = comm.rank();
             mpi_size_ = comm.size();
 
+            setMaxThreads();
+        }
+
+        static void setMaxThreads()
+        {
 #if _OPENMP
             // If openMP is available, default to 2 threads per process unless
             // OMP_NUM_THREADS is set or command line --threads-per-process used.
--- a/opm/simulators/flow/FlowMain.hpp
+++ b/opm/simulators/flow/FlowMain.hpp
@@ -302,10 +302,9 @@ namespace Opm {
                 if (result.ec == std::errc() && omp_num_threads > 0) {
                     // Set threads to omp_num_threads if it was successfully parsed and is positive
                     threads = omp_num_threads;
-                    // Warning in 'Main.hpp', where this code is duplicated
-                    // if (requested_threads > 0) {
-                    //     OpmLog::warning("Environment variable OMP_NUM_THREADS takes precedence over the --threads-per-process cmdline argument.");
-                    // }
+                    if (requested_threads > 0) {
+                        OpmLog::warning("Environment variable OMP_NUM_THREADS takes precedence over the --threads-per-process cmdline argument.");
+                    }
                 } else {
                     OpmLog::warning("Invalid value for OMP_NUM_THREADS environment variable.");
                 }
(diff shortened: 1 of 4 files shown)


michal-toth:
The code structure for setting the `Parameters::ThreadsPerProcess` turned out to be more complicated than expected. Therefore, I make this hotfix mergable and will work iteratively on the clean solution.

michal-toth:
The bug is now properly resolved. The maximum limit on threads is set after parameters are parsed (command line argument `--threads-per-process` is known), and before the Deck is read and its parallel for is used.
The function `getNumThreads()` from Main.hpp no longer holds the hotfix and is simplified to just return `omp_get_max_threads()`.

michal-toth:
jenkins build this please

blattms:
Wow. that is a nice diff. A fix that removes a lot of boiler plate code. I love it. :clap: :+1: 
Gotta try it though!

blattms:
Please update the description and make sure that title indicates what it fixed.

atgeirr:
> The function `getNumThreads()` from Main.hpp no longer holds the hotfix

But that function is still called from two places (printFlowBanner() and readDeck() calls), what will it return now? 16 for a 16-core processor etc.? Or will it be the correct number (the one set in setMaxThreads() now).

michal-toth:
getNumThreads prints the current maximum of threads. That maximum is set before either of those calls.

blattms:
@michal-toth You will need to uncomment the warning that OMP_NUM_THREADS takes precedence)in setMaxThreads as the other occurence is removed.

Other than that this look great

atgeirr:
> @michal-toth You will need to uncomment the warning that OMP_NUM_THREADS takes precedence)in setMaxThreads as the other occurence is removed.

Good point Markus, just one further minor potential issue: Has logging been set up when setMaxThreads() is called? If so, using OpmLog::warning() is good, otherwise it will be (almost) a no-op. (We should make sure logging is set up early enough, but there may be some snags I do not see.)

bska:
> > Has logging been set up when setMaxThreads() is called?
> 
> Yes, there is an OpmLog::error call just before setMaxThreads().

We construct the run's logging facilities as part of [`Main::readDeck()`](https://github.com/OPM/opm-simulators/blob/48454762bf38f3abc1e250ca2ed87a7e5836842c/opm/simulators/flow/Main.cpp#L237-L241).  Any calls to `OpmLog::*` prior to this moment will be lost.

michal-toth:
> > > Has logging been set up when setMaxThreads() is called?
> > 
> > 
> > Yes, there is an OpmLog::error call just before setMaxThreads().
> 
> We construct the run's logging facilities as part of [`Main::readDeck()`](https://github.com/OPM/opm-simulators/blob/48454762bf38f3abc1e250ca2ed87a7e5836842c/opm/simulators/flow/Main.cpp#L237-L241). Any calls to `OpmLog::*` prior to this moment will be lost.

readDeck happens after we set max threads, so the message would be lost in runs with static dispatch. Running `flow`, that has dynamic dispatch, parses parameters once more after the deck is read and the warning would be recorded...

bska:
> ```sh
> export OMP_NUM_THREADS=bla
> ./bin/flow path/to/opm-tests/spe1/SPE1CASE2 &> log
> ```

What happens if you invoke `bin/flow_blackoil` directly?

blattms:
> > ```shell
> > export OMP_NUM_THREADS=bla
> > ./bin/flow path/to/opm-tests/spe1/SPE1CASE2 &> log
> > ```
> 
> What happens if you invoke `bin/flow_blackoil` directly?

> > If run `flow_blackoil" instead that line is only there once.

Like I said above. There is output, just not twice.

michal-toth:

blattms:
Interesting unrelated timeouts...

blattms:

akva2:
all those sims use the parameter definition that is now changed so i don't think it's unrelated.

michal-toth:
That could be caused by the change in `fvbaseparameters.hh` from 1 to 2.
`BlackoilModelParameters.cpp` used -1 but the threadmanager reinterpreted it to 2, so using 2 as default just made it explicit.

akva2:
yes, this is a bit obtuse.

for the 'other' sims the constexpr value is used. for flow, this value is overriden by the call to SetDefault. so your change in fvbaseparameters.hh only applies to the other sims, it has no effect for flow.

michal-toth:
> so, in effect, you have unconditionally enabled multi-threading. in places where it should not be enabled. since you changed the default to 2.

I will revert the `fvbaseparameters.hh` to its original state.

michal-toth:

michal-toth:
Reverting the change in `fvbaseparameters.hh` solved the problem.

michal-toth:
jenkins build this please

opm/simulators/flow/FlowMain.hpp
```diff
@@ -101,12 +101,11 @@ namespace Opm {
                 ("Developer option to see whether logging was on non-root processors. "
                  "In that case it will be appended to the *.DBG or *.PRT files");
 
-            ThreadManager::registerParameters();
```
blattms, line 104: Where/How is this called now?

PR #5861, opened 2025-01-10
Bugfix: Upon restart, do not try to restore segment info for inactive wells

vkip:
Fallout from https://github.com/OPM/opm-simulators/pull/5751  (causes segfault if any well present in the restart is found to be inactive in the restarted run).

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelRestart.cpp
+++ b/opm/simulators/wells/BlackoilWellModelRestart.cpp
@@ -256,7 +256,7 @@ loadRestartData(const data::Wells&                 rst_wells,
     {
         const auto& well_name = well_state.name(well_index);
 
-        this->loadRestartWellData(well_name, handle_ms_well, phs,
+        this->loadRestartWellData(well_name, handle_ms_well & !well_state.is_permanently_inactive_well(well_name), phs,
                                   rst_wells.at(well_name),
                                   wellModel_.perfData(well_index),
                                   well_state.well(well_index));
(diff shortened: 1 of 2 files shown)


vkip:
jenkins build this please

GitPaean (approved):
This is a small change and bug fixing,  and I go with what the test indicates.   Potentially, we might need to revisit here again if more complicated logic joins (ACTIONX) and it will still go with what test indicates.

opm/simulators/wells/BlackoilWellModelRestart.cpp
```diff
@@ -256,7 +256,7 @@ loadRestartData(const data::Wells&                 rst_wells,
     {
         const auto& well_name = well_state.name(well_index);
 
-        this->loadRestartWellData(well_name, handle_ms_well, phs,
+        this->loadRestartWellData(well_name, handle_ms_well & !well_state.is_permanently_inactive_well(well_name), phs,
```
bska, line 259: Sorry for not responding in a timely manner.  This single ampersand (`&`, bit-wise "and") should probably be a double ampersand (`&&`, logical "and") here.

GitPaean: Yeah. You are right. I did not notice it.  And I will create a PR for this.

vkip: Thanks!

PR #5862, opened 2025-01-13
adapting to the interface change for PTFlash::solve()

GitPaean:
depending on https://github.com/OPM/opm-common/pull/4419

### DIFF
--- a/opm/models/ptflash/flashmodel.hh
+++ b/opm/models/ptflash/flashmodel.hh
@@ -231,7 +231,7 @@ public:
             ("Method for solving vapor-liquid composition. Available options include: "
              "ssi, newton, ssi+newton");
 
-        Parameters::SetDefault<Parameters::FlashTolerance<Scalar>>(1e-12);
+        Parameters::SetDefault<Parameters::FlashTolerance<Scalar>>(1.e-8);
         Parameters::SetDefault<Parameters::EnableIntensiveQuantityCache>(true);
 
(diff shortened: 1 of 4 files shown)


opm/models/ptflash/flashmodel.hh
```diff
@@ -231,7 +231,7 @@ public:
             ("Method for solving vapor-liquid composition. Available options include: "
              "ssi, newton, ssi+newton");
 
-        Parameters::SetDefault<Parameters::FlashTolerance<Scalar>>(1e-12);
+        Parameters::SetDefault<Parameters::FlashTolerance<Scalar>>(1.e-8);
```
atgeirr, line 234: It is not obvious to me what this tolerance is used to compare, a change of 4 orders of magnitude is a lot though! Was this change indicated by testing?

GitPaean: >         Parameters::SetDefault<Parameters::FlashTolerance<Scalar>>(1e-12);

This one was set by mistake.  From OPM/opm-common#4419, you will see the value is never used in the master branch.  There might be a reason to use 1.e-12 due to some sources, while I could not tell now. And 1.e-8  and 1.e-6  has been hard-coded in opm-common for Newton and SSI flash calculation as the actual flash tolerance. 

Here, we just use 1.e-8 and the test results are good. It might be some values we tune in the future, but 1.e-8 flash tolerance looks good for now.

PR #5863, opened 2025-01-13
refactor move_to_gpu

multitalentloes:
To avoid confusion with move semantics in C++ `move_to_gpu` is renamed to `copy_to_gpu` as the movement of data to the GPU does not invalidate the existing CPU copy.

Should be updated in tandem with https://github.com/OPM/opm-common/pull/4420


kjetilly (approved):
Good semantic change to avoid confusion with move semantics

PR #5867, opened 2025-01-13
BlackoilModel: add impl file

akva2:
This class has grown large enough to warrant splitting into an impl file.

Since this would basically touch every line anyways, I also opted to re-indent the existing header to drop the indent for namespace.


akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

bska (approved):
It looks like the actual changes are minor (reformatting &c) so I'll merge into master given that the build check is green.  For future reference, though, these kinds of changes would preferably be piecemeal. E.g., introduce the `_imp.hpp` file first and then move one or a few functions at a time in separate PRs.

akva2:
yeah, that's fair enough

PR #5868, opened 2025-01-13
regression test for gas water and multisegment wells

GitPaean:


GitPaean:
jenkins build this opm-tests=1280 please

akva2:
while this is fine as such, i want to wait for https://github.com/OPM/opm-common/pull/4423 (which also means we can and should drop the egrid change)

GitPaean:
> which also means we can and should drop the egrid change

That was not intentional. I recovered EGRID file from the master but git can not compare binary file still thinks it is changed.  Now I have removed it.  Let jenkins check.

GitPaean:
jenkins build this opm-tests=1280 please

akva2:
excellent. thanks.

PR #5869, opened 2025-01-13
BlackoilModel: make wasSwitched_ private

akva2:
No idea why this was public in the first place, but in any case there does not seem to be a reason any longer.


akva2:
jenkins build this please

bska (approved):

PR #5873, opened 2025-01-14
Change return type of _init_ of the BlackOilSimulator class

lisajulia:


lisajulia:
jenkins build this please

PR #5874, opened 2025-01-14
SimulatorConvergenceOutput: move to separate class

akva2:
model using has-a in SimulatorFullyImplicitBlackoil

### DIFF
--- a/opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp
+++ b/opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp
@@ -34,6 +34,7 @@
 #include <opm/simulators/flow/ConvergenceOutputConfiguration.hpp>
 #include <opm/simulators/flow/ExtraConvergenceOutputThread.hpp>
 #include <opm/simulators/flow/NonlinearSolver.hpp>
+#include <opm/simulators/flow/SimulatorConvergenceOutput.hpp>
 #include <opm/simulators/flow/SimulatorReportBanners.hpp>
 #include <opm/simulators/flow/SimulatorSerializer.hpp>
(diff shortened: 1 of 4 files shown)


akva2:
jenkins build this please

opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp
```diff
@@ -420,14 +426,7 @@ class SimulatorFullyImplicitBlackoil : private SerializableSim
...
             const auto& reps = this->solver_->model().stepReports();
-
-            auto reports = std::vector<StepReport> {
-                reps.begin() + this->already_reported_steps_, reps.end()
-            };
-
-            this->writeConvergenceOutput(std::move(reports));
-
-            this->already_reported_steps_ = reps.size();
+            convergence_output_.write(std::move(reps));
```
bska, line 429: Tiny nit: This should be `write(reps)` instead of `write(move(reps))`.  Given that `reps` is a reference-to-`const` and that `SimulatorConvergenceOutput::write()` takes a reference-to-`const`, the extra `move()` here is misleading.

bska (commented):
Looks good to me except for one tiny nit.

opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp
akva2: yeah. leftover, as i changed to do the range extract on the inside..

akva2:
jenkins build this please

bska (approved):
Thanks a lot for the speedy response.  This looks good to me now and I'll merge into master.

PR #5877, opened 2025-01-14
Fix WCYCLE for multiple blocks

akva2:
This fixes WCYCLE when there are multiple WCYCLE blocks, multiple cycling well and these are combined with manual WELOPEN statements. Downstream of https://github.com/OPM/opm-common/pull/4426 and requires https://github.com/OPM/opm-tests/pull/1282


bska (approved):

PR #5878, opened 2025-01-14
FlowBaseVanguard: move some more code to FlowGenericVanguard

akva2:


akva2:
jenkins build this please

PR #5882, opened 2025-01-15
OutputBlackoilModule: move some more code to GenericOutputBlackoilModule

akva2:

### DIFF
--- a/opm/simulators/flow/GenericOutputBlackoilModule.hpp
+++ b/opm/simulators/flow/GenericOutputBlackoilModule.hpp
@@ -48,6 +48,13 @@
 #include <utility>
 #include <vector>
 
+namespace Opm::Parameters {
+
+struct ForceDisableFluidInPlaceOutput { static constexpr bool value = false; };
+struct ForceDisableResvFluidInPlaceOutput { static constexpr bool value = false; };
+
+} // namespace Opm::Parameters
+
 namespace Opm {
 
 namespace data { class Solution; }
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

opm/simulators/flow/GenericOutputBlackoilModule.hpp
```diff
@@ -65,14 +72,21 @@ class GenericOutputBlackoilModule {
...
+    Scalar* getPRESSURE_ptr(void)
+    {
         return (this->fluidPressure_.data()) ;
     };
 
-    int  getPRESSURE_size( void ) {
+    int getPRESSURE_size( void )
+    {
         return (this->fluidPressure_.size()) ;
     };
```
bska, line 83: If you're doing these kinds of changes here, maybe we could remove the `void` parameter type and the unnecessary semicolon on the functions' closing braces as well?

bska (commented):
Looks good to me and I'd be happy to merge this as is.  That said, I think this is an opportunity to make one other aesthetic change here.

opm/simulators/flow/GenericOutputBlackoilModule.hpp
akva2: sure. i know it's unrelated but i can't resist doing some cleanups while passing through..

akva2:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  This looks good to me and I'll merge into master.

PR #5883, opened 2025-01-15
FlowProblem: move some more code to FlowGenericProblem

akva2:


akva2:
jenkins build this please

bska (approved):

PR #5887, opened 2025-01-16
fixing compilation related to conversion from std::optional

GitPaean:
The following compilation error did not show up with the master branch, but it showed up when I tried to compile https://github.com/OPM/opm-simulators/pull/5851 and it caused some hassle. 

Not sure what is the exact reason, but just propose a solution here for discussion. 

```
/home/kaib/OPM-devel3/opm/opm-common/opm/input/eclipse/Schedule/Group/GuideRate.hpp:125:10: note: candidate: ‘void Opm::GuideRate::compute(const std::string&, const Opm::Phase&, std::size_t, double)’
  125 |     void compute(const std::string& wgname,
      |          ^~~~~~~
/home/kaib/OPM-devel3/opm/opm-common/opm/input/eclipse/Schedule/Group/GuideRate.hpp:128:37: note:   no known conversion for argument 4 from ‘std::optional<double>’ to ‘double’
  128 |                  const double       guide_rate);
      |                  ~~~~~~~~~~~~~~~~~~~^~~~~~~~~
```

### DIFF
--- a/opm/simulators/wells/WellGroupHelpers.cpp
+++ b/opm/simulators/wells/WellGroupHelpers.cpp
@@ -334,11 +334,11 @@ updateGuideRatesForInjectionGroups(const Group& group,
                              deferred_logger);
         }
 
-        const UnitSystem& unit_system = schedule.getUnits();
         if (guideRateValue) {
+            const UnitSystem& unit_system = schedule.getUnits();
             guideRateValue = unit_system.from_si(UnitSystem::measure::rate, *guideRateValue);
         }
-        guideRate->compute(group.name(), phase, reportStepIdx, guideRateValue);
+        guideRate->compute(group.name(), phase, reportStepIdx, guideRateValue.value_or(0.0));
     }
 }
 


GitPaean:
I do not know what the correct thing to do if `guideRateValue` is not set with some value, the fix appears as a fix from the grammar level to fix the compilation for the circumstance.

bska:
> ```
> ‘void Opm::GuideRate::compute(const std::string&, const Opm::Phase&, std::size_t, double)’
> ```

That looks weird.  The four-parameter version of `GuideRate::compute()` takes an [`optional<double>`](https://github.com/OPM/opm-common/blob/37c468b16e203f68d971ce26c5e98edb36f3cd01/opm/input/eclipse/Schedule/Group/GuideRate.hpp#L125-L128) directly.  Do you have any local changes to the `GuideRate` class?

GitPaean:
image
image


the branch is clean and it does not look it touch anything related to GuideRate.

bska:
> the branch is clean and it does not look it touch anything related to `GuideRate`.

Right, but that's not what I'm asking.  I'm asking if your copy of `GuideRate` is exactly the same as [the current master version](https://github.com/OPM/opm-common/blob/37c468b16e203f68d971ce26c5e98edb36f3cd01/opm/input/eclipse/Schedule/Group/GuideRate.hpp).

GitPaean:
> Right, but that's not what I'm asking. I'm asking if your copy of `GuideRate` is exactly the same as [the current master version](https://github.com/OPM/opm-common/blob/37c468b16e203f68d971ce26c5e98edb36f3cd01/opm/input/eclipse/Schedule/Group/GuideRate.hpp).

Sure, that is the problem. 

> > the branch is clean and it does not look it touch anything related to `GuideRate`.
> 
> Right, but that's not what I'm asking. I'm asking if your copy of `GuideRate` is exactly the same as [the current master version](https://github.com/OPM/opm-common/blob/37c468b16e203f68d971ce26c5e98edb36f3cd01/opm/input/eclipse/Schedule/Group/GuideRate.hpp).

Sure. I see now. Somehow the updating through the folders were not complete, it misses the latest update from https://github.com/OPM/opm-common/pull/4391. 

Thanks for pointing out and I will close the PR.

PR #5888, opened 2025-01-16
Transmissibility: add a cache for centroids and use this in update()

akva2:
The loops repeatedly calls distanceVector which recalculates the centroid for a cell. This is not a trivial calculation

### DIFF
--- a/opm/simulators/flow/Transmissibility.hpp
+++ b/opm/simulators/flow/Transmissibility.hpp
@@ -284,6 +284,7 @@ protected:
     const CartesianIndexMapper& cartMapper_;
     const Grid& grid_;
     std::function<std::array<double,dimWorld>(int)> centroids_;
+    std::vector<std::array<double,dimWorld>> centroids_cache_;
     Scalar transmissibilityThreshold_;
     std::map<std::pair<unsigned, unsigned>, Scalar> transBoundary_;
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this please

atgeirr:
> This is not a trivial calculation

Indeed, and in the current master there are two (or more, with thermal etc.) such calculations for every connected cell pair, which is very excessive!

Limited testing indicate that the memory usage increase comes well before the peak, so it should be fine in that perspective, since it is cleared when no longer needed.

blattms:
So no other performance changes allowed until tomorrow. Unless you make your private benchmarking available. :sweat_smile:

akva2:
What exactly do you want ? This would not show up on a regular benchmark at all. Do millions of cells and it does.

e.g. on a 5M cell model with diffuse I got a reduction from 21s to 15 ish seconds. On a 50M model atgeirr saw 51s to 35s, I don't know exactly what kind of model that was.

atgeirr:
> On a 50M model atgeirr saw 51s to 35s, I don't know exactly what kind of model that was.

That was a model with just basic transmissibilities, no thermal, diffusion or others. Would probably be a greater difference if that was included.

On our automated benchmarks I doubt this would show up as significant. The main advantage of this is that it becomes easier to experiment with huge models to push the limits: if you are waiting 20 minutes or more for the setup stage to see if something worked, then reducing that significantly is valuable.

PR #5891, opened 2025-01-16
Chase UDQConfig::Eval() API Change

bska:
Following PR OPM/opm-common#4433, we must now pass a `GroupOrder` object into `eval()` in order to support group name pattern matching in group level UDQ assignments.


blattms (approved):
I think we should change the title to make the additional functionality immediately apparent. (Helps with release notes)

This is not only chasing stuff, right?

tskille:
This PR has now been tested in Equinor and results are good. Looking forward to see this going into master

bska:
I'm marking this PR ready for review now.  As for the PR title, it's somewhat debatable.  The actual change here is "just" to add another argument to a function call and as such is really just chasing an API change.  The justification for that API change is explained in the upstream PR.

akva2 (approved):

PR #5892, opened 2025-01-16
Feature/py action insert kw welpi wpimult alternative

lisajulia:
Alternative to https://github.com/OPM/opm-simulators/pull/5467 as suggested by @bska in https://github.com/OPM/opm-common/pull/4136#issuecomment-2583291034.

For the reference manual: With this PR and https://github.com/OPM/opm-common/pull/4434, the keywords WELPI and WPIMULT can be used in the "insert keywords"-function in Pyaction.

### DIFF
--- a/opm/simulators/flow/ActionHandler.cpp
+++ b/opm/simulators/flow/ActionHandler.cpp
@@ -137,17 +137,11 @@ namespace {
 
     template <typename Scalar, class WellModel>
     std::unordered_map<std::string, Scalar>
-    fetchWellPI(const int                                    reportStep,
-                const Opm::Schedule&                         schedule,
-                const WellModel&                             wellModel,
-                const Opm::Action::ActionX&                  action,
-                const Opm::Action::Result::MatchingEntities& matches,
+    fetchWellPI(const WellModel&                             wellModel,
+                const std::vector<std::string>&              wellpi_wells,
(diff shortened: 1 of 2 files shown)


opm/simulators/flow/ActionHandler.cpp
```diff
@@ -237,9 +237,12 @@ applyActions(const int reportStep,
 
         logActiveAction(action->name(), matches.wells(), ts);
 
-        const auto wellpi = fetchWellPI<Scalar>
-            (reportStep, this->schedule_, this->wellModel_,
-             *action, matches, this->comm_);
+        // The std::unordered_map<std::string, Scalar> wellpi contains the well production indices from the last
+        // timestep. This map is needed for the keywords WELPI and WPIMULT.
```
bska, line 241: Strictly speaking, the current PI values aren't really needed for WPIMULT.  WPIMULT applies a constant CTF multiplier to (a subset of) the well's connections and the multiplier is given explicitly in the WPIMULT input data.  On the other hand, the WELPI keyword calculates the multiplier as a ratio of the input value and the current dynamic PI value.

bska (commented):
This looks good to me save for a slightly misleading comment concerning the `WPIMULT` keyword.

bska (approved):

PR #5894, opened 2025-01-17
Implement gsatprod

totto82:
Depends on
https://github.com/OPM/opm-common/pull/4188

Sufficient to run https://github.com/OPM/opm-tests/pull/1207 where GSATPROD only interact with output and new test case https://github.com/OPM/opm-tests/pull/1286

### DIFF
--- a/opm/simulators/wells/WellGroupHelpers.cpp
+++ b/opm/simulators/wells/WellGroupHelpers.cpp
@@ -103,14 +103,15 @@ namespace Opm {
             if (wellState.isRank0() && !injector && gsatprod.has(groupName)) {
                 const auto& gsatprod_rates = gsatprod.get(groupName);
                 const auto& pu = wellState.phaseUsage();
+                using Rate = GSatProd::GSatProdGroup::Rate;
                 if (pu.phase_used[BlackoilPhases::Aqua] && pu.phase_pos[BlackoilPhases::Aqua] == phasePos) {
-                    rate += gsatprod_rates.water_rate;
+                    rate += gsatprod_rates.rate[Rate::Water];
                 }
                 if (pu.phase_used[BlackoilPhases::Liquid] && pu.phase_pos[BlackoilPhases::Liquid] == phasePos) {
-                    rate += gsatprod_rates.oil_rate;
+                    rate += gsatprod_rates.rate[Rate::Oil];
                 }
                 if (pu.phase_used[BlackoilPhases::Vapour] && pu.phase_pos[BlackoilPhases::Vapour] == phasePos) {
-                    rate += gsatprod_rates.gas_rate;
+                    rate += gsatprod_rates.rate[Rate::Gas];
                 }
             }
         }
--- a/opm/simulators/wells/WellGroupHelpers.cpp
+++ b/opm/simulators/wells/WellGroupHelpers.cpp
@@ -98,21 +98,13 @@ namespace Opm {
             const auto& groupTmp = schedule.getGroup(groupName, reportStepIdx);
             const auto& gefac = groupTmp.getGroupEfficiencyFactor();
             rate += gefac * sumWellPhaseRates(res_rates, groupTmp, schedule, wellState, reportStepIdx, phasePos, injector);
-            const auto& gsatprod = schedule[reportStepIdx].gsatprod.get();
-            // only sum once
-            if (wellState.isRank0() && !injector && gsatprod.has(groupName)) {
-                const auto& gsatprod_rates = gsatprod.get(groupName);
-                const auto& pu = wellState.phaseUsage();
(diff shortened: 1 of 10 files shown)


totto82:
Now also with interaction with group controls.

totto82:
Simple test case added. https://github.com/OPM/opm-tests/pull/1286

totto82:
The only difference between GSATPROD1 and GSATPROD2 is that GSATPROD2 includes GCONPROD with 2500 ORAT. The well PROD's ORAT is correctly limited to meet 2500 ORAT target for group FIELD in GSATPROD2. 

image

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -97,6 +98,22 @@ namespace Opm {
...
+                if (pu.phase_used[BlackoilPhases::Aqua] && pu.phase_pos[BlackoilPhases::Aqua] == phasePos) {
+                    rate += gsatprod_rates.rate[Rate::Water];
+                }
+                if (pu.phase_used[BlackoilPhases::Liquid] && pu.phase_pos[BlackoilPhases::Liquid] == phasePos) {
+                    rate += gsatprod_rates.rate[Rate::Oil];
+                }
+                if (pu.phase_used[BlackoilPhases::Vapour] && pu.phase_pos[BlackoilPhases::Vapour] == phasePos) {
+                    rate += gsatprod_rates.rate[Rate::Gas];
+                }
+            }
```
bska, line 116: I think I'd prefer if this logic were separated out to a distinct helper function, or at least a separate loop.  Something along the lines of
```C++
template <typename Scalar>
Scalar satelliteProduction(const ScheduleState& sched,
                           const std::vector<std::string>& groups,
                           const GSatProd::GSatProdGroup::Rate rateComp)
{
    auto satProdRate = Scalar{};

    const auto& gsatProd = sched.gsatprod();
    for (const auto& group : groups) {
        if (! gsatProd.has(group)) {
            continue;
        }

        gsatProdRate += gsatProd.get(group).rate[rateComp];
    }

    return gsatProdRate;
}

std::optional<GSatProd::GSatProdGroup::Rate>
selectRateComponent(const PhaseUsage& pu, const int phasePos)
{
    using Rate = GSatProd::GSatProdGroup::Rate;

    for (const auto& [phase, rateComp] : std::array {
        std::pair { BlackoilPhases::Aqua, Rate::Water },
        std::pair { BlackoilPhases::Liquid, Rate::Oil },
        std::pair { BlackoilPhases::Vapour, Rate::Gas } })
    {
        if (pu.phase_used[phase] && (pu.phase_pos[phase] == phasePos)) {
            return rateComp;
        }
    }

    return std::nullopt;
}
```
which you'd call, either as part of `sumWellPhaseRates()` or separately, as
```C++
if (wellState.isRank0() && !injector) {
    const auto rateComp = selecteRateComponent(wellState.phaseUsage(), phasePos);
    if (rateComp.has_value()) {
        rate += satelliteProduction(schedule[reportStepIdx], group.groups(), *rateComp);
    }
}
```
This would reduce the amount of nested logic.

bska (commented):
This looks okay to me, but I really think we should try to limit the amount of nested logic we add when making new features, and, generally try to simplify the functions if at all possible.  The `WellGroupHelpers` are already quite complicated and the fact that we can't understand any single function in isolation because everything gets accumulated into `updateAndCommunicateGroupData()`.

opm/simulators/wells/WellGroupHelpers.cpp
totto82: Thanks for the suggestions.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -97,6 +98,13 @@ namespace Opm {
             const auto& groupTmp = schedule.getGroup(groupName, reportStepIdx);
             const auto& gefac = groupTmp.getGroupEfficiencyFactor();
             rate += gefac * sumWellPhaseRates(res_rates, groupTmp, schedule, wellState, reportStepIdx, phasePos, injector);
+            // only sum satelite production once
+            if (wellState.isRank0() && !injector) {
+                const auto rateComp = selectRateComponent(wellState.phaseUsage(), phasePos);
+                if (rateComp.has_value()) {
+                    rate += satelliteProduction(schedule[reportStepIdx], group.groups(), *rateComp);
+                }
+            }
```
bska, line 107: This section needs to be outside of the `groupName` loop lest we include the contributions multiple times in the case that more than one of `group.groups()` is a satellite.

bska (commented):
Thank you for the updates.  I think this is starting to look good.  Just one small issue remaining, I think.

bska (approved):

PR #5895, opened 2025-01-17
Transmissibility: some janitoring

akva2:
- fix cosmetics
- simplify method signatures by introducing a struct
- reduce code duplication

### DIFF
--- a/opm/simulators/flow/Transmissibility_impl.hpp
+++ b/opm/simulators/flow/Transmissibility_impl.hpp
@@ -1319,31 +1319,11 @@ applyMultipliers_(Scalar& trans,
                   unsigned cartElemIdx,
                   const TransMult& transMult) const
 {
-    // apply multiplyer for the transmissibility of the face. (the
+    // apply multiplier for the transmissibility of the face. (the
     // face index is the index of the reference-element face which
     // contains the intersection of interest.)
-    switch (faceIdx) {
-    case 0: // left
-        trans *= transMult.getMultiplier(cartElemIdx, FaceDir::XMinus);
-        break;
-    case 1: // right
-        trans *= transMult.getMultiplier(cartElemIdx, FaceDir::XPlus);
-        break;
-
-    case 2: // front
-        trans *= transMult.getMultiplier(cartElemIdx, FaceDir::YMinus);
-        break;
-    case 3: // back
-        trans *= transMult.getMultiplier(cartElemIdx, FaceDir::YPlus);
-        break;
-
-    case 4: // bottom
-        trans *= transMult.getMultiplier(cartElemIdx, FaceDir::ZMinus);
-        break;
-    case 5: // top
-        trans *= transMult.getMultiplier(cartElemIdx, FaceDir::ZPlus);
-        break;
-    }
+    trans *= transMult.getMultiplier(cartElemIdx,
+                                     FaceDir::FromIntersectionIndex(faceIdx));
 }
 
 template<class Grid, class GridView, class ElementMapper, class CartesianIndexMapper, class Scalar>
--- a/opm/simulators/flow/Transmissibility_impl.hpp
+++ b/opm/simulators/flow/Transmissibility_impl.hpp
@@ -414,9 +414,7 @@ update(bool global, const TransUpdateQuantities update_quantities,
                 //  PINCH(4) == TOPBOT is assumed here as we set useSmallestMultipliers
                 // to false if  PINCH(4) == ALL holds
                 // In contrast to the name this will also apply
-                applyAllZMultipliers_(trans, inside.faceIdx,
-                                      outside.faceIdx, inside.cartElemIdx,
-                                      outside.cartElemIdx, transMult, cartDims);
+                applyAllZMultipliers_(trans, inside, outside, transMult, cartDims);
             }
             else {
                 applyMultipliers_(trans, inside.faceIdx, inside.cartElemIdx, transMult);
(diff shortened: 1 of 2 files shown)


atgeirr:
I'll review this.

opm/simulators/flow/Transmissibility_impl.hpp
```diff
@@ -238,301 +239,240 @@ update(bool global, const TransUpdateQuantities update_quantities,
...
-            unsigned outsideElemIdx = elemMapper.index(outsideElem);  
+            outside.elemIdx = elemMapper.index(outsideElem);
 
-            // Get the Cartesian indices of the origen cells (parent or equivalent cell on level zero), for CpGrid with LGRs.
-            // For genral grids and no LGRs, get the usual Cartesian Index.
-            unsigned insideCartElemIdx = this-> lookUpCartesianData_.template getFieldPropCartesianIdx<Grid>(elemIdx);
-            unsigned outsideCartElemIdx =  this-> lookUpCartesianData_.template getFieldPropCartesianIdx<Grid>(outsideElemIdx);
+            // Get the Cartesian indices of the origin cells (parent or equivalent cell on level zero),
+            // for CpGrid with LGRs. For general grids and no LGRs, get the usual Cartesian Index.
+            inside.cartElemIdx = this->lookUpCartesianData_.
```
atgeirr, line 368: Maybe we should set all the properties of `inside` outside the intersection loop, instead of setting them over and over again for every intersection.

opm/simulators/flow/Transmissibility_impl.hpp
```diff
@@ -1346,59 +1309,24 @@ applyMultipliers_(Scalar& trans,
...
-        break;
-
-    case 2: // front
-        trans *= ntg[elemIdx];
-        break;
-    case 3: // back
-        trans *= ntg[elemIdx];
-        break;
-
-        // NTG does not apply to top and bottom faces
```
atgeirr, line 1401: This comment should be included still, otherwise it is very magical to see trans modified just for a seemingly random range of face indices.

atgeirr (commented):
Looks good to me, just a few issues.

opm/simulators/flow/Transmissibility_impl.hpp
```diff
@@ -238,301 +239,234 @@ update(bool global, const TransUpdateQuantities update_quantities,
 
     // fill the centroids cache to avoid repeated calculations in loops below
     centroids_cache_.resize(gridView_.size(0));
-    for (const auto& elem : elements(gridView_)) {
+    for (const auto& elem : Dune::elements(gridView_)) {
```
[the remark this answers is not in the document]
akva2: it works as such, it's just my head that doesn't immediately follow the same parsing rules. dropped the prefix.

opm/simulators/flow/Transmissibility_impl.hpp
akva2: it's good idea but there is some reuse of variables etc that i have not fully analyzed yet.

opm/simulators/flow/Transmissibility_impl.hpp
akva2: re-added.

opm/simulators/flow/Transmissibility_impl.hpp
atgeirr: Ok.

opm/simulators/flow/Transmissibility_impl.hpp
```diff
@@ -243,296 +244,235 @@ update(bool global, const TransUpdateQuantities update_quantities,
...
+                                      distanceVector_(inside.faceCenter, inside.elemIdx),
+                                      permeability_[inside.elemIdx]);
 
                 // normally there would be two half-transmissibilities that would be
                 // averaged. on the grid boundary there only is the half
                 // transmissibility of the interior element.
-                unsigned insideCartElemIdx = cartMapper_.cartesianIndex(elemIdx);
-                applyMultipliers_(transBoundaryIs, intersection.indexInInside(), insideCartElemIdx, transMult);
-                transBoundary_[std::make_pair(elemIdx, boundaryIsIdx)] = transBoundaryIs;
+                inside.cartElemIdx = cartMapper_.cartesianIndex(inside.elemIdx);
```
akva2, line 337: @aritorto can you explain the reason for this lookup not going through lookupdata?

aritorto: If I remember correctly, we do not support multipliers - yet - for CpGrid with lgrs. I'll have a deeper look though. That would be my first answer!

opm/simulators/flow/Transmissibility_impl.hpp
```diff
@@ -725,35 +674,32 @@ removeNonCartesianTransmissibilities_(bool removeAll)
...
-                      unsigned outsideCartElemIdx,
+                      const FaceInfo& inside,
+                      const FaceInfo& outside,
                       const TransMult& transMult,
                       const std::array<int, dimWorld>& cartDims)
 {
-    if(grid_.maxLevel()> 0) {
-                OPM_THROW(std::invalid_argument, "MULTZ not support with LGRS, yet.");
+    if (grid_.maxLevel() > 0) {
+        OPM_THROW(std::invalid_argument, "MULTZ not support with LGRS, yet.");
```
aritorto, line 683: @akva2 so what we do not support yet is MULTZ (for CpGrid with LGRs), throwing when calling applyAllZMultipliers(...). applyMultipliers does not throw for CpGrid with Lgrs. Then it can/must be a bug.

akva2: As far as I can see, the only difference here is that we are on the boundary.

opm/simulators/flow/Transmissibility_impl.hpp
akva2: As far as I can see, the difference here is that the intersection is on the boundary. We will continue the intersection loop inside this if, so we will never hit the multiplier code in this code path.

aritorto: Just double-checked that, for a CpGrid with Lgrs, cartMapper_.cartesianIndex(inside.elemidx) returns the Cartesian index of the parent cell. So, it's correct in the end. 

When we only need the Cartesian Index of an element on the leaf grid, then we can just use its Cartesian mapper. If we want to know the value of a field property of a leaf element, then it's more convenient to use the lookup via lookupCartesianData. 

I hope this helps!

akva2: What I am trying to understand is whether or not I can move the lookup for the inside cell outside the intersection loop so that the lookup is done only once and not once per intersection. If it is safe to replace this call by a call through lookupdata then that is the case, and I can move https://github.com/OPM/opm-simulators/blob/master/opm/simulators/flow/Transmissibility_impl.hpp#L310 outside the loop and remove this line.

aritorto: I see your point. Indeed, it's much better and safe to replace this call by a call through lookup**Cartesian**Data.

akva2: yeah sorry for my shorting, I of course mean lookupCartesianData. thx!

opm/simulators/flow/Transmissibility_impl.hpp
akva2: Addressed.

atgeirr (approved):

akva2:
green and approved, merging.

PR #5900, opened 2025-01-22
Improve well shutting logic

totto82:
In particular allow for failed wells to shut even though they haven't failed consistently in the last 3 attempts if the time step is reduce enough already.

### DIFF
--- a/opm/simulators/timestepping/AdaptiveTimeStepping.cpp
+++ b/opm/simulators/timestepping/AdaptiveTimeStepping.cpp
@@ -33,7 +33,7 @@ void logTimer(const AdaptiveSimulatorTimer& substepTimer)
 }
 
 std::set<std::string>
-consistentlyFailingWells(const std::vector<StepReport>& sr)
+consistentlyFailingWells(const std::vector<StepReport>& sr, bool requireRepeatedFailures)
 {
     // If there are wells that cause repeated failures, we
     // close them, and restart the un-chopped timestep.
@@ -60,10 +60,10 @@ consistentlyFailingWells(const std::vector<StepReport>& sr)
     const int rep_step = sr.back().report_step;
     const int sub_step = sr.back().current_step;
     const int sr_size = sr.size();
-    if (sr_size >= num_steps) {
-        for (const auto& wf : wfs) {
-            failing_wells.insert(wf.wellName());
-        }
+    for (const auto& wf : wfs) {
+        failing_wells.insert(wf.wellName());
+    }
+    if (requireRepeatedFailures && sr_size >= num_steps) {
         for (int s = 1; s < num_steps; ++s) {
             const auto& srep = sr[sr_size - 1 - s];
             // Report must be from same report step and substep, otherwise we have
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

opm/simulators/timestepping/AdaptiveTimeStepping.cpp
```diff
@@ -33,7 +33,7 @@ void logTimer(const AdaptiveSimulatorTimer& substepTimer)
 }
 
 std::set<std::string>
-consistentlyFailingWells(const std::vector<StepReport>& sr)
+consistentlyFailingWells(const std::vector<StepReport>& sr, bool checkRepeatedFailures)
```
atgeirr, line 36: I misunderstood the intention at first, maybe call the argument `requireRepeatedFailuresToFlag`? Also, make it const.

opm/simulators/timestepping/AdaptiveTimeStepping_impl.hpp
```diff
@@ -902,10 +902,14 @@ AdaptiveTimeStepping<TypeTag>::SubStepIteration<Solver>::
 chopTimeStepOrCloseFailingWells_(const int new_time_step)
 {
     bool wells_shut = false;
-    // We are below the threshold, and will check if there are any wells we should close
-    // rather than chopping again.
-    std::set<std::string> failing_wells = detail::consistentlyFailingWells(
-                                         solver_().model().stepReports());
+    // We are below the threshold, and will check if there are any
```
atgeirr, line 905: Typos: failes->fails, several->several.

opm/simulators/timestepping/AdaptiveTimeStepping_impl.hpp
```diff
@@ -902,10 +902,14 @@ AdaptiveTimeStepping<TypeTag>::SubStepIteration<Solver>::
...
+    // wells that failes repeatedly (that means that it fails in the last three steps)
+    // we should close rather than chopping again.
+    // If we already have chopped the timestep serveral times due to failed wells we also shut wells that
+    // fails only on this step.
+    bool checkRepeatedFailures = new_time_step > ( minTimeStepBeforeClosingWells_()*restartFactor_()*restartFactor_());
```
atgeirr, line 910: Why do you multiply twice with the restartFactor_()? It may be ok, but then it should be easy to understand why from the comments in the code.

atgeirr (commented):
Looks good, I think this should be merged. Just a few minor issues.

opm/simulators/timestepping/AdaptiveTimeStepping_impl.hpp
totto82: It mimics the checking of the three last timesteps in "repeated failures". 
The idea is that if you already have reduced the timestep 2 times due to well A before it gets shut, you dont want to have to reduce it again for well B. As sometimes well B does not fail until well A is shut. I will add a comment.

totto82:
jenkins build this please

atgeirr:
Thanks, merging.

PR #5901, opened 2025-01-22
using logical AND instead of bit-wise AND

GitPaean:
overlooking when reviewing PR #5861


GitPaean:
jenkins build this please

bska (approved):
Thanks a lot for the quick turn-around.  I'll merge into master.

PR #5902, opened 2025-01-22
Dont reset subtimestep from tuning/wcycle if time step is chopped due  to convergence issues

totto82:
Needs a bit more testing before it is ready


akva2:
This should be fine for WCYCLE. WCYCLE wants to adjust the time step to precisely match the cycling period for some well. A restart would cut the time step to be smaller than whichever adjustment WCYCLE wanted to do, and once we have a converged step, restarts is set back to 0, allowing WCYCLE to (potentially) do a new adjustment to achieve its goal.

hakonhagland (approved):

PR #5904, opened 2025-01-22
xxxPrimaryVariables: remove ctor from Scalar

akva2:
these were only in use for assignments to a constant. 

Instead of detouring all of these assignments through a temporary, use the base field vector assignment operator to do inplace assignments.


akva2:
jenkins build this please

akva2:
rather
```
x = 0.0; 
```
Those that were explicit already had assignments operators.

bska (approved):
I really like the simplification here.  The changes look good to me and I'll merge into master.

PR #5905, opened 2025-01-22
fixed: condition on MPI_FOUND, not USE_MPI

akva2:
USE_XXX variables are user provided and should not be used to check whether or not a component is enabled (at least not those backed by some required library).

In particular the build system will configure a build without MPI if no mpi was found -> trouble.

Should fix nightly python packaging again, ref https://ci.opm-project.org/job/opm-nightly-pypi/1495/console


akva2:
jenkins build this please

PR #5906, opened 2025-01-22
Janitoring: const correctness

akva2:
Address more static analyzer issues


akva2:
jenkins build this please

opm/simulators/linalg/gpuistl/GpuOwnerOverlapCopy.hpp
```diff
@@ -275,24 +285,25 @@ class GPUAwareMPISender : public GPUSender<field_type, OwnerOverlapCopyCommunica
 
     void buildCommPairIdxs() const
     {
-        auto &ri = this->m_cpuOwnerOverlapCopy.remoteIndices();
+        const auto& ri = this->m_cpuOwnerOverlapCopy.remoteIndices();
         std::vector<int> commpairIndicesCopyOnCPU;
         std::vector<int> commpairIndicesOwnerCPU;
 
-        for(auto process : ri) {
+        for (auto process : ri) {
```
bska, line 292: I don't off hand know the size of `ri` here, but this range-`for` will copy the full sequence.  If you're changing this anyway, maybe we could rewrite it as
```C++
for (const auto& process : ri) {
    // ...
}
```
instead?

bska (commented):
Greatly appreciated.  Just one small thing that we may want to do while here.

opm/simulators/linalg/gpuistl/GpuOwnerOverlapCopy.hpp
akva2: good idea.

akva2:
jenkins build this please

PR #5908, opened 2025-01-23
Janitoring: address various static analyzer issues

akva2:
https://github.com/OPM/opm-simulators/commit/ec6cba9e0d33bb311180a359b51aecae1144bc39 is a bit eyebrow-raising but in any case it makes no sense as-is.

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -919,48 +919,44 @@ template<class Scalar>
 void WellState<Scalar>::communicateGroupRates(const Parallel::Communication& comm)
 {
     // Compute the size of the data.
-    std::size_t sz = 0;
-    for (const auto& [_, owner_rates] : this->well_rates) {
-        (void)_;
-        const auto& [__, rates] = owner_rates;
-        (void)__;
-        sz += rates.size();
-    }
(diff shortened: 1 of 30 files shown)


akva2:
jenkins build this please

opm/simulators/wells/WellState.cpp
```diff
@@ -919,48 +919,44 @@ template<class Scalar>
...
-        }
+    auto pos = data.begin();
+    std::for_each(this->well_rates.begin(), this->well_rates.end(),
+                  [&pos](const auto& input)
+                  {
+                      const auto& [owner, rates] = input.second;
+                      if (owner) {
+                          std::copy(rates.begin(), rates.end(), pos);
+                      }
+                      pos += input.second.second.size();
```
bska, line 937: You've already decomposed `input.second` into `owner` and `rates`, so maybe we could write this as
```C++
pos += rates.size();
```
instead?

bska (commented):
Much appreciated.  Just one small thing that stood out to me.

opm/simulators/wells/WellState.cpp
akva2: yup.

akva2:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  This looks good to me now and I'll merge into master.  I really appreciate getting rid of those (unused) objects named `_` and `__`...

PR #5914, opened 2025-01-24
Lgr wells

hnil:
- draft pull request for using WTRAJ wells with LGR depends on opm-common
Currently one need to run with the options
--parsing-strictness=low   : to avoid error in parsing
 --linear-solver=ilu0  : to avoid problems with changing linear system size maybe all but cprw is ok (wells sizes change)
 --enable-ecl-output=false : to avoid error in output


bska:
jenkins build this please

hnil:
Closing — this is the older WTRAJ-with-LGR approach, superseded by the current LGR series. Branch kept.

PR #5916, opened 2025-01-24
Fix gpmaint for region 0

totto82:
region 0 means the whole field

Depends on https://github.com/OPM/opm-common/pull/4447

### DIFF
--- a/opm/simulators/wells/RegionAverageCalculator.hpp
+++ b/opm/simulators/wells/RegionAverageCalculator.hpp
@@ -89,17 +89,10 @@ namespace Opm {
                     numRegions = std::max(numRegions, reg);
                 }
                 numRegions = comm.max(numRegions);
-                for (int reg = 1; reg <= numRegions ; ++ reg) {
+                // reg = 0 is used for field
+                for (int reg = 0; reg <= numRegions ; ++ reg) {
                     if (!attr_.has(reg))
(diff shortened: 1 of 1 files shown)


bska (commented):
This is fine, but as far as I can tell the `defineState()` function isn't called anywhere so I don't really know what problem we're addressing.

totto82:
> This is fine, but as far as I can tell the defineState() function isn't called anywhere so I don't really know what problem we're addressing.

What do you mean? Isn't called several places in the well model like here

https://github.com/OPM/opm-simulators/blob/c7c0aeb4e10509d3dce8357f32fdcdc9c5f405c1/opm/simulators/wells/BlackoilWellModel_impl.hpp#L218C12-L218C90

bska:
> > This is fine, but as far as I can tell the defineState() function isn't called anywhere so I don't really know what problem we're addressing.
> 
> What do you mean? Isn't called several places in the well model [like here](https://github.com/OPM/opm-simulators/blob/c7c0aeb4e10509d3dce8357f32fdcdc9c5f405c1/opm/simulators/wells/BlackoilWellModel_impl.hpp#L218C12-L218C90).

Yeah, you're right.  My grep(1) expression didn't account for the explicit template argument...

bska:
jenkins build this please

PR #5917, opened 2025-01-27
Remove Solver template parameter for AdaptiveTimeStepping::step()

hakonhagland:
Removes the template parameter `Solver` used with the `AdaptiveTimeStepping::solve()` method. Try to replace the template parameter with a equivalent `Solver` value from the `Opm::Properties::` namespace. I am currently debugging this, so I will put it in draft mode for now.

### DIFF
--- a/opm/simulators/timestepping/AdaptiveTimeStepping.hpp
+++ b/opm/simulators/timestepping/AdaptiveTimeStepping.hpp
@@ -13,6 +13,7 @@
 #include <opm/models/utils/basicproperties.hh>
 #include <opm/models/utils/propertysystem.hh>
 
+#include <opm/simulators/flow/NonlinearSolver.hpp>
 #include <opm/simulators/timestepping/AdaptiveSimulatorTimer.hpp>
 #include <opm/simulators/timestepping/SimulatorReport.hpp>
 #include <opm/simulators/timestepping/SimulatorTimer.hpp>
@@ -79,8 +80,11 @@ template<class TypeTag>
 class AdaptiveTimeStepping
 {
 private:
+    using Model = GetPropType<TypeTag, Properties::Model>;
     using Scalar = GetPropType<TypeTag, Properties::Scalar>;
-    template <class Solver>
+    using Solver = NonlinearSolver<TypeTag, Model>;
+    using Simulator = GetPropType<TypeTag, Properties::Simulator>;
+
     class SolutionTimeErrorSolverWrapper : public RelativeChangeInterface
     {
     public:
@@ -91,25 +95,21 @@ private:
         const Solver& solver_;
     };
 
-    // Forward declaration of SubStepIteration
-    // TODO: This forward declaration is not enough for GCC version 9.4.0 (released June 2021),
-    //  but it works fine with GCC version 13.2.0 (released July 2023).
-    template <class Solver> class SubStepIteration;
+    class SubStepIteration;
 
-    template <class Solver>
     class SubStepper {
     public:
         SubStepper(
             AdaptiveTimeStepping<TypeTag>& adaptive_time_stepping,
+            Solver &solver,
             const SimulatorTimer& simulator_timer,
-            Solver& solver,
             const bool is_event,
             const std::function<bool(const double, const double, const int)>& tuning_updater
         );
 
         AdaptiveTimeStepping<TypeTag>& getAdaptiveTimerStepper();
         SimulatorReport run();
-        friend class AdaptiveTimeStepping<TypeTag>::template SubStepIteration<Solver>;
+        friend class AdaptiveTimeStepping<TypeTag>::SubStepIteration;
 
     private:
         bool isReservoirCouplingMaster_() const;
@@ -127,17 +127,17 @@ private:
         double suggestedNextTimestep_() const;
 
         AdaptiveTimeStepping<TypeTag>& adaptive_time_stepping_;
-        const SimulatorTimer& simulator_timer_;
         Solver& solver_;
+        const SimulatorTimer& simulator_timer_;
         const bool is_event_;
         const std::function<bool(double elapsed, double dt, int sub_step_number)>& tuning_updater_;
+        Simulator& simulator_;
     };
 
-    template <class Solver>
     class SubStepIteration {
     public:
         SubStepIteration(
-            SubStepper<Solver>& substepper,
+            SubStepper& substepper,
             AdaptiveSimulatorTimer& substep_timer,
             const double original_time_step,
             const bool final_step
@@ -179,7 +179,7 @@ private:
         bool useNewtonIteration_() const;
         double writeOutput_() const;
 
-        SubStepper<Solver>& substepper_;
+        SubStepper& substepper_;
         AdaptiveSimulatorTimer& substep_timer_;
         const double original_time_step_;
         const bool final_step_;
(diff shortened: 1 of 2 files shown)


hakonhagland:
jenkins build this please

hakonhagland:
In `FlowProblemBlackoilProperties.hpp`, the Model type is defined as a `FIBlackOilModel<TypeTag>` :

https://github.com/OPM/opm-simulators/blob/6bfb60ded033450d1df31ae79674f429b26a4422/opm/simulators/flow/FlowProblemBlackoilProperties.hpp#L66

but I cannot find the actual definition of `FIBlackOilModel<TypeTag>` ?

akva2:
The Model in the typetag is the 'ewoms' (opm-models) type model, while the BlackoilModel.hpp is the opm model which is not the same thing. Welcome to the domain of frankenstein. There is no typetag entry for the latter.

hakonhagland:
> Welcome to the domain of frankenstein.

Yes, indeed :sweat_smile: a scary place to enter

> There is no typetag entry for the latter.

Then as a next attempt I would simply try to do the following in `AdaptiveTimeStepping.hpp` (and avoid using the `Model = GetPropType<TypeTag, Properties::Model>` type):

```
#include <opm/simulators/flow/BlackoilModel.hpp>
using Model = BlackoilModel<TypeTag>;
using Solver = NonlinearSolver<TypeTag, Model>;
// ...
```
Although this actually works for most opm `.cpp` files, there are cases like `flowexp_comp.cpp` which includes `FlowProblemComp.hpp`

https://github.com/OPM/opm-simulators/blob/d2b272b5f54779a72923d090b0a6282ae56e241a/flowexperimental/comp/flowexp_comp.hpp#L28

which includes `FlowProblem.hpp`

https://github.com/OPM/opm-simulators/blob/d2b272b5f54779a72923d090b0a6282ae56e241a/opm/simulators/flow/FlowProblemComp.hpp#L34

which includes `AdaptiveTimeStepping.hpp`

https://github.com/OPM/opm-simulators/blob/d2b272b5f54779a72923d090b0a6282ae56e241a/opm/simulators/flow/FlowProblem.hpp#L65

If now `AdaptiveTimeStepping.hpp` includes `BlackoilModel.hpp` as proposed above, then it again will include `BlackoilModelProperties.hpp`

https://github.com/OPM/opm-simulators/blob/d2b272b5f54779a72923d090b0a6282ae56e241a/opm/simulators/flow/BlackoilModel.hpp#L31

which will include `FlowProblemBlackoilProperties.hpp`

https://github.com/OPM/opm-simulators/blob/d2b272b5f54779a72923d090b0a6282ae56e241a/opm/simulators/flow/BlackoilModelProperties.hpp#L27

which defines the `MaterialLaw` property with given traits

https://github.com/OPM/opm-simulators/blob/d2b272b5f54779a72923d090b0a6282ae56e241a/opm/simulators/flow/FlowProblemBlackoilProperties.hpp#L76

which will use `FluidSystem::waterPhaseIdx` to create three phase material traits, however this is not using the expected fluid system for a blackoil problem but rather the fluid system included here

https://github.com/OPM/opm-simulators/blob/d2b272b5f54779a72923d090b0a6282ae56e241a/flowexperimental/comp/flowexp_comp.hpp#L23

which uses a `waterPhaseIdx = -1`

https://github.com/OPM/opm-common/blob/efbebfe159d41477aa2acffce9c6648a6b587294/opm/material/fluidsystems/GenericOilGasFluidSystem.hpp#L69

which leads to the following compilation error

```
[...]
/home/hakon/test/opm2/opm/opm-simulators/flowexperimental/comp/flowexp_comp.cpp:83:35:   required from here
/home/hakon/test/opm2/opm/opm-common/opm/material/fluidmatrixinteractions/MaterialTraits.hpp:105:21: error: static assertion failed: wettingPhaseIdx is out of range
  105 |     static_assert(0 <= wettingPhaseIdx && wettingPhaseIdx < numPhases,
      |                   ~~^~~~~~~~~~~~~~~~~~
/home/hakon/test/opm2/opm/opm-common/opm/material/fluidmatrixinteractions/MaterialTraits.hpp:105:21: note: the comparison reduces to ‘(0 <= -1)’
```

bska:
> Removes the template parameter Solver used with the `AdaptiveTimeStepping::solve()` method. Try to replace the template parameter with a equivalent Solver value from the `Opm::Properties::` namespace.

If you don't mind me asking, what problem are you trying to solve by this.  You're taking a specific direction, but I guess I'm missing the bigger picture/context for this change.

hakonhagland:
> what problem are you trying to solve by this? 

@bska Good question. In the beginning I just wanted to be able to declare the solver as a class variable of AdaptiveTimeStepping. However, I realized later that that would be difficult due to the serialization of AdaptiveTimeStepper. So currently I think of this as a merely cleanup or simplification of the current code. I was also curious why or if this template parameter was necessary. Now as as the complexity of the problem is clearer to me, we might just leave it there? Though, the difficulty of removing it might indicate some deeper issues with the type/property system that could be worth investigating. What do you think?

hakonhagland:
>  while the BlackoilModel.hpp is the opm model

@akva2 There is also path of include files such that if one includes the opm model in `BlackoilModel.hpp` like what is done in `SimulatorFullyImplicitBlackoil.hpp` the ewoms model `blackoilmodel.hh` will be automatically included:

https://github.com/OPM/opm-simulators/blob/6bfb60ded033450d1df31ae79674f429b26a4422/opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp#L44

and `BlackoilModel.hpp` includes `BlackoilModelNldd.hpp`

https://github.com/OPM/opm-simulators/blob/6bfb60ded033450d1df31ae79674f429b26a4422/opm/simulators/flow/BlackoilModel.hpp#L30

and `BlackoilModelNldd.hpp` includes `ISTLSolverGpuBridge.hpp`

https://github.com/OPM/opm-simulators/blob/6bfb60ded033450d1df31ae79674f429b26a4422/opm/simulators/flow/BlackoilModelNldd.hpp#L42

and `ISTLSolverGpuBridge.hpp` includes `ISTLSolver.hpp`

https://github.com/OPM/opm-simulators/blob/6bfb60ded033450d1df31ae79674f429b26a4422/opm/simulators/linalg/ISTLSolverGpuBridge.hpp#L25

and `ISTLSolver.hpp` includes `FlowBaseProblemProperties.hpp`:

https://github.com/OPM/opm-simulators/blob/6bfb60ded033450d1df31ae79674f429b26a4422/opm/simulators/linalg/ISTLSolver.hpp#L38

and ` FlowBaseProblemProperties.hpp` includes `TracerModel.hpp`

https://github.com/OPM/opm-simulators/blob/6bfb60ded033450d1df31ae79674f429b26a4422/opm/simulators/flow/FlowBaseProblemProperties.hpp#L42

and `TracerModel.hpp` includes `GenericTracerModel.hpp `

https://github.com/OPM/opm-simulators/blob/6bfb60ded033450d1df31ae79674f429b26a4422/opm/simulators/flow/TracerModel.hpp#L35

and  `GenericTracerModel.hpp ` includes `blackoilmodel.hh`

https://github.com/OPM/opm-simulators/blob/6bfb60ded033450d1df31ae79674f429b26a4422/opm/simulators/flow/GenericTracerModel.hpp#L35

I think this is what causes the redefinition of the `FluidSystem` property for cases like `flowexp_comp.cpp` since `blackoilmodel.hh` (re)defines the fluid system property to by of type `BlackOilFluidSystem`:

https://github.com/OPM/opm-simulators/blob/aa5c52ee5694f3911aa974fd72a749834b93fc25/opm/models/blackoil/blackoilmodel.hh#L141

PR #5919, opened 2025-01-27
Allow empty leaf nodes that do not exist in GRUPTREE

vkip:


vkip:
jenkins build this please

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -920,7 +920,14 @@ computeNetworkPressures(const Network::ExtNetwork& network,
         // Starting with the leaf nodes of the network, get the flow rates
         // from the corresponding groups.
         std::map<std::string, std::vector<Scalar>> node_inflows;
+        const std::vector<Scalar> zero_rates(3, 0.0);
```
GitPaean, line 923: it should be fine. ideally, we should use `PhaseUsage::num_phases` here, but I think there are many places in the file we are assuming 3 phases.  those should be refactored as a whole.

GitPaean:
Thanks for the contribution. I am getting it in now.

PR #5921, opened 2025-01-28
Document and Test PropertyTree Class

bska:
In particular, add Doxygen-style documentation to the header file and add a simple unit test for the `PropertyTree` class interface.

While here, also add missing headers and remove those explicit template arguments that are not needed.


bska:
jenkins build this please

opm/simulators/linalg/PropertyTree.cpp
```diff
@@ -23,8 +23,14 @@
...
 #include <boost/property_tree/json_parser.hpp>
 
-namespace Opm
-{
+#include <memory>
+#include <optional>
+#include <ostream>
+#include <string>
+
+#include <stddef.h>
```
akva2, line 31: we should probably change this to cstddef and instance for std::size_t. but okay to do later.

opm/simulators/linalg/PropertyTree.hpp
```diff
@@ -20,51 +20,136 @@
 #ifndef OPM_PROPERTYTREE_HEADER_INCLUDED
 #define OPM_PROPERTYTREE_HEADER_INCLUDED
 
+#include <functional>
```
akva2, line 23: probably missing something but i don't see where this is required?

akva2 (approved):
Two small comments.

akva2:
oh, and there are more TAD to be had (all of them can be deduced).

opm/simulators/linalg/PropertyTree.hpp
bska: > probably missing something but i don't see where this is required?

We use [`std::less<>`](https://en.cppreference.com/w/cpp/utility/functional/less) in the [`ptree` definition](https://github.com/OPM/opm-simulators/blob/f73a027edc116a407224b1d911d2b7caf7f55550/opm/simulators/linalg/PropertyTree.hpp#L31).

opm/simulators/linalg/PropertyTree.cpp
bska: > we should probably change this to cstddef and instance for std::size_t. but okay to do later.

I'll make a follow-up PR for that.

bska:
> oh, and there are more TAD to be had (all of them can be deduced).

Right.  I'll update the PR accordingly.

bska:
> > oh, and there are more TAD to be had (all of them can be deduced).
> 
> Right. I'll update the PR accordingly.

I've pushed an update to that effect now and I'll run a build test.

bska:
jenkins build this please

bska:
jenkins build this please

PR #5923, opened 2025-01-28
Correct a keyword in a message

michal-toth:
When `--allowDistributedWells=true` is set in a problem with multisegment wells, program outputs a message where the command line argument `--use-multisegment-well` is misnamed.

### DIFF
--- a/opm/simulators/flow/FlowGenericVanguard.cpp
+++ b/opm/simulators/flow/FlowGenericVanguard.cpp
@@ -351,7 +351,7 @@ void FlowGenericVanguard::init()
                 std::string message =
                         std::string("Option --allow-distributed-wells=true in a model with\n")
                         + "multisegment wells. This feature is still experimental. You can\n"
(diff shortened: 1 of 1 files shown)


michal-toth:
This commit only changes the text in a message. Testing if it builds is unnecessary.

atgeirr:
> Testing if it builds is unnecessary.

Yet I still do it... (at least some of the time)

One thing is that silly mistakes happen, but it also confirms to me that the system is working.

akva2:
yes. always run jenkins before merge. simple and explicit rule to relate to.

atgeirr (approved):

atgeirr:
jenkins build this please

bska:
Another point in favour of build testing every PR is that doing so will also run the unit and regression tests which is very valuable. In general our tests should not depend on the exact text of a diagnostic message, but the system is getting big enough that something might have slipped through...

PR #5924, opened 2025-01-29
Update group target reduction during local solve

totto82:
Just for testing.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -520,7 +520,7 @@ namespace Opm
 
         // store a copy of the well state, we don't want to update the real well state
         WellState<Scalar> well_state_copy = well_state;
-        const auto& group_state = simulator.problem().wellModel().groupState();
+        auto group_state = simulator.problem().wellModel().groupState();
         auto& ws = well_state_copy.well(this->index_of_well_);
         
         // get current controls        
@@ -1610,7 +1610,7 @@ namespace Opm
                              const Well::InjectionControls& inj_controls,
                              const Well::ProductionControls& prod_controls,
                              WellState<Scalar>& well_state,
-                             const GroupState<Scalar>& group_state,
+                             GroupState<Scalar>& group_state,
                              DeferredLogger& deferred_logger, 
                              const bool fixed_control /*false*/, 
                              const bool fixed_status /*false*/)
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -520,7 +520,7 @@ namespace Opm
 
         // store a copy of the well state, we don't want to update the real well state
         WellState<Scalar> well_state_copy = well_state;
-        auto group_state = simulator.problem().wellModel().groupState();
+        auto group_state_copy = simulator.problem().wellModel().groupState();
         auto& ws = well_state_copy.well(this->index_of_well_);
         
         // get current controls        
@@ -558,9 +558,9 @@ namespace Opm
         // solve equations
         bool converged = false;
         if (this->well_ecl_.isProducer() && this->wellHasTHPConstraints(summary_state)) {
-            converged = well_copy.solveWellWithTHPConstraint(simulator, dt, inj_controls, prod_controls, well_state_copy, group_state, deferred_logger);
+            converged = well_copy.solveWellWithTHPConstraint(simulator, dt, inj_controls, prod_controls, well_state_copy, group_state_copy, deferred_logger);
         } else {
-            converged = well_copy.iterateWellEqWithSwitching(simulator, dt, inj_controls, prod_controls, well_state_copy, group_state, deferred_logger);
+            converged = well_copy.iterateWellEqWithSwitching(simulator, dt, inj_controls, prod_controls, well_state_copy, group_state_copy, deferred_logger);
         }
 
         // fetch potentials (sign is updated on the outside).
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -314,6 +314,10 @@ namespace Opm
                     if (! (isGroupControl && !this->param_.check_group_constraints_inner_well_iterations_)) {
                         changed = this->checkIndividualConstraints(ws, summary_state, deferred_logger, inj_controls, prod_controls);
                     }
+                    if(changed) {
+                        updateWellStateWithTarget(simulator, group_state, well_state, deferred_logger);
+                        updatePrimaryVariables(simulator, well_state, deferred_logger);
+                    }
                     const auto well_state_orig = well_state; // Ugly - take a copy for now...
                     // Check for group controls if present and switch not disabled, but only if we did not already change to individual
                     if (hasGroupControl && this->param_.check_group_constraints_inner_well_iterations_) {
@@ -360,7 +364,11 @@ namespace Opm
                             if (!thp_controlled){
                             // don't call for thp since this might trigger additional local solve
                                 updateWellStateWithTarget(simulator, group_state, well_state, deferred_logger);
-                            } else if (isGroupControl) { // we switched from group
+                            } else {
+                                updateWellStateWithTarget(simulator, group_state, well_state, deferred_logger);
+                                updatePrimaryVariables(simulator, well_state, deferred_logger);
+                            }
+                            if (isGroupControl) { // we switched from group
                                 well_state.updateGlobalIsGrup(this->index_of_well_);
                                 int reportStepIdx = simulator.episodeIndex();
                                 const Group& fieldGroup = schedule.getGroup("FIELD", reportStepIdx);
@@ -375,10 +383,9 @@ namespace Opm
                                                                                     summary_state,
                                                                                     group_state,
                                                                                     groupTargetReduction);
-                            } else {
-                                ws.thp = this->getTHPConstraint(summary_state);
+                                //updateWellStateWithTarget(simulator, group_state, well_state, deferred_logger);
+                                //updatePrimaryVariables(simulator, well_state, deferred_logger);
                             }
-                            updatePrimaryVariables(simulator, well_state, deferred_logger);
                         }
                     }
                 }
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -308,15 +308,32 @@ namespace Opm
                     
                     const bool hasGroupControl = this->isInjector() ? inj_controls.hasControl(Well::InjectorCMode::GRUP) :
                                                                       prod_controls.hasControl(Well::ProducerCMode::GRUP);
-                    const bool isGroupControl = ws.production_cmode == Well::ProducerCMode::GRUP || ws.injection_cmode == Well::InjectorCMode::GRUP;
+                    const bool wasGroupControl = ws.production_cmode == Well::ProducerCMode::GRUP || ws.injection_cmode == Well::InjectorCMode::GRUP;
+
+                    int reportStepIdx = simulator.episodeIndex();
+                    const Group& fieldGroup = schedule.getGroup("FIELD", reportStepIdx);
 
                     // We first check for individual controls (unless on group control and switch disabled)
-                    if (! (isGroupControl && !this->param_.check_group_constraints_inner_well_iterations_)) {
+                    if (! (wasGroupControl && !this->param_.check_group_constraints_inner_well_iterations_)) {
                         changed = this->checkIndividualConstraints(ws, summary_state, deferred_logger, inj_controls, prod_controls);
                     }
                     if(changed) {
                         updateWellStateWithTarget(simulator, group_state, well_state, deferred_logger);
                         updatePrimaryVariables(simulator, well_state, deferred_logger);
+                        if (wasGroupControl) {
+                            well_state.updateGlobalIsGrup(this->index_of_well_);
+                            std::vector<Scalar> groupTargetReduction(well_state.numPhases(), 0.0);
+                            WellGroupHelpers<Scalar>::updateGroupTargetReduction(fieldGroup,
+                                                                                schedule,
+                                                                                reportStepIdx,
+                                                                                this->isInjector(),
+                                                                                well_state.phaseUsage(),
+                                                                                *this->guideRate(),
+                                                                                well_state,
+                                                                                summary_state,
+                                                                                group_state,
+                                                                                groupTargetReduction);
+                        }
                     }
                     const auto well_state_orig = well_state; // Ugly - take a copy for now...
                     // Check for group controls if present and switch not disabled, but only if we did not already change to individual
@@ -337,8 +354,6 @@ namespace Opm
                             // This may give some discrepencies in non-linear behaviour between MPI runs,
                             // but the solution should convergece to the same solution.
                             well_state.updateGlobalIsGrup(this->index_of_well_);
-                            int reportStepIdx = simulator.episodeIndex();
-                            const Group& fieldGroup = schedule.getGroup("FIELD", reportStepIdx);
                             std::vector<Scalar> groupTargetReduction(well_state.numPhases(), 0.0);
                             WellGroupHelpers<Scalar>::updateGroupTargetReduction(fieldGroup,
                                                                                 schedule,
@@ -357,35 +372,6 @@ namespace Opm
                             // New group target violates an individual constraint - reset
                             well_state = well_state_orig;
                             changed = false;
-                        }
-                    } else { // We did not switch to group, update individual target if present
-                        if (changed) {
-                            const bool thp_controlled = this->isInjector() ? ws.injection_cmode == Well::InjectorCMode::THP : ws.production_cmode == Well::ProducerCMode::THP;
-                            if (!thp_controlled){
-                            // don't call for thp since this might trigger additional local solve
-                                updateWellStateWithTarget(simulator, group_state, well_state, deferred_logger);
-                            } else {
-                                updateWellStateWithTarget(simulator, group_state, well_state, deferred_logger);
-                                updatePrimaryVariables(simulator, well_state, deferred_logger);
-                            }
-                            if (isGroupControl) { // we switched from group
-                                well_state.updateGlobalIsGrup(this->index_of_well_);
-                                int reportStepIdx = simulator.episodeIndex();
-                                const Group& fieldGroup = schedule.getGroup("FIELD", reportStepIdx);
-                                std::vector<Scalar> groupTargetReduction(well_state.numPhases(), 0.0);
-                                WellGroupHelpers<Scalar>::updateGroupTargetReduction(fieldGroup,
-                                                                                    schedule,
-                                                                                    reportStepIdx,
-                                                                                    this->isInjector(),
-                                                                                    well_state.phaseUsage(),
-                                                                                    *this->guideRate(),
-                                                                                    well_state,
-                                                                                    summary_state,
-                                                                                    group_state,
-                                                                                    groupTargetReduction);
-                                //updateWellStateWithTarget(simulator, group_state, well_state, deferred_logger);
-                                //updatePrimaryVariables(simulator, well_state, deferred_logger);
-                            }
                         }
                     }
                 }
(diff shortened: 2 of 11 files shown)


totto82:

totto82:

totto82:

totto82:

totto82:

totto82:

totto82:
Most test failures result from differences in time-stepping. Approximately half are minor differences. There are also some failures where the PR gives the correct solution while the master violates some group controls, like GGPR:PROD for 16_9_4a_grpctl_stw_model2.pdf, where the limit is violated at one timestep.

totto82:

opm/simulators/wells/WellState.cpp
```diff
@@ -948,6 +948,20 @@ void WellState<Scalar>::updateGlobalIsGrup(const Parallel::Communication& comm)
     this->global_well_info.value().communicate(comm);
 }
 
+template<class Scalar>
+void WellState<Scalar>::updateGlobalIsGrup()
+{
+    this->global_well_info.value().clear();
```
GitPaean, line 954: Do we need to communicate this with all the other processes, locally, there might be only some of the wells?

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -313,12 +313,16 @@ namespace Opm
...
+                    bool isGroupControl = ws.production_cmode == Well::ProducerCMode::GRUP || ws.injection_cmode == Well::InjectorCMode::GRUP;
+
+                    if (this->param_.check_group_constraints_inner_well_iterations_) {
+                        if (!isGroupControl && hasGroupControl) {
+                            changed = changed || this->checkGroupConstraints(well_state, group_state, schedule, summary_state,deferred_logger);
+                        }
+                        changed = changed || this->checkIndividualConstraints(ws, summary_state, deferred_logger, inj_controls, prod_controls);
+                    } else {
+                        if (!isGroupControl)
+                            changed = changed || this->checkIndividualConstraints(ws, summary_state, deferred_logger, inj_controls, prod_controls);
```
GitPaean, line 325: Just for discussion, should not we always check individual constraints? Here, if `check_group_constraints_inner_well_iterations_` is false, and well is under group, then it will not do any constraint checking. It looks strange from the code level.

totto82: In current master switching to and from GRUP during local switching is problematic since the reduction rate is not updated. I added "check_group_constraints_inner_well_iterations_" as a fast work-around. The goal of this PR is to fix this so we can switch to and from GRUP during local iterations. I plan to keep the parameter for some time, but if we don't see more issues, we should remove.

opm/simulators/wells/WellState.cpp
totto82: Yes. All processes need to know how many wells are under GRUP control.

GitPaean: but the function `void WellState<Scalar>::updateGlobalIsGrup()` does not look involve any communication? That was my main concern.

opm/simulators/wells/WellInterface_impl.hpp
GitPaean: Thanks for the explanation. 

I meant for the else branch (check_group_constraints_inner_well_iterations_ is false)
```c++
else {
                        if (!isGroupControl)
                            changed = changed || this->checkIndividualConstraints(ws, summary_state, deferred_logger, inj_controls, prod_controls);
```

if the well is under group control, this branch will not check any constraints, including the individual constraints.

Is this intentional?

opm/simulators/wells/WellState.cpp
totto82: Yes, you are right. During local solves, we are mostly concerned about ourselves. I.e., we only update the updateGlobalIsGrup locally. It would have been better to communicate this to all processes, but that is not possible during local solves without significant restructuring of the code, which could potentially reduce the scalability of the simulator as all processors would need to be aware of all the wells.  That is why I added an updateGlobalIsGrup without communication for the local solves. I will refactor this part and add some comments to make this clear.

totto82: I have added more comments and refactored the code. I hope this is more clear now.

opm/simulators/wells/WellInterface_impl.hpp
totto82: Yes. If check_group_constraints_inner_well_iterations_=false. We neither want to switch to GRUP or from GRUP. This is the same logic that we have in master.

opm/simulators/wells/WellState.cpp
GitPaean: for the case I am looking at, there are 9 wells, when running with 8 processes, most of the processes have 1 or 2 wells. 

In the function call, first,
```c++
this->global_well_info.value().clear();
```

All the information will be 0. 

The following, 
```c++
    for (std::size_t well_index = 0; well_index < this->size(); well_index++) {
        const auto& ws = this->well(well_index);
        this->global_well_info.value().update_efficiency_scaling_factor(well_index, ws.efficiency_scaling_factor);
        if (ws.producer)
            this->global_well_info.value().update_producer(well_index, ws.status, ws.production_cmode);
        else
            this->global_well_info.value().update_injector(well_index, ws.status, ws.injection_cmode);
    }
```
You will be able to update one or two wells in the `global_well_info`. The data for other most of the wells will stay 0. 

My question, can we use this type of partial information to `updateGroupTargetReduction()`. This is my big question here.

totto82: That was not my intention. Thanks for noticing. I have updated the PR accordingly. Surprisingly, it didn't change much in the small test cases I have. I will do some testing and re-run Jenkins.

totto82:

totto82:

totto82:

totto82:

opm/simulators/wells/WellState.cpp
totto82: I have tested this on relevant models in parallel and the results are good.

GitPaean: Thanks. I am looking into it.

totto82:
Testing done. From my perspective, this can be merged. We will still have the option of not doing local switching to fall back on, but I think this PR improves on the current state in the master. Even though it could possibly be improved even further.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -307,12 +307,16 @@ namespace Opm
                     
                     const bool hasGroupControl = this->isInjector() ? inj_controls.hasControl(Well::InjectorCMode::GRUP) :
                                                                       prod_controls.hasControl(Well::ProducerCMode::GRUP);
-                    bool isGroupControl = ws.production_cmode == Well::ProducerCMode::GRUP || ws.injection_cmode == Well::InjectorCMode::GRUP; 
-                    if (! (isGroupControl && !this->param_.check_group_constraints_inner_well_iterations_)) {
-                        changed = this->checkIndividualConstraints(ws, summary_state, deferred_logger, inj_controls, prod_controls);
-                    }
-                    if (hasGroupControl && this->param_.check_group_constraints_inner_well_iterations_) {
-                        changed = changed || this->checkGroupConstraints(well_state, group_state, schedule, summary_state,deferred_logger);
+                    bool isGroupControl = ws.production_cmode == Well::ProducerCMode::GRUP || ws.injection_cmode == Well::InjectorCMode::GRUP;
```
GitPaean, line 310: `const` before `bool`.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -326,6 +330,34 @@ namespace Opm
                         }
                         updatePrimaryVariables(simulator, well_state, deferred_logger);
                     }
+
+                    bool isNowGroupControl = ws.production_cmode == Well::ProducerCMode::GRUP || ws.injection_cmode == Well::InjectorCMode::GRUP;
```
GitPaean, line 334: `const` before `bool`

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -326,6 +330,34 @@ namespace Opm
...
+                    if (isGroupControl != isNowGroupControl) {
+                        // We need to update the globalIsGRUP vector and the group target reduction.
+                        // But we can not communicate it to the other ranks during the local iterations.
+                        // The globalIsGrup vector is thus only communicated after the local iterations are done.
+                        // This may give some discrepencies in non-linear behaviour between MPI runs,
+                        // but the solution should convergece to the same solution.
+                        well_state.updateGlobalIsGrup();
```
GitPaean, line 341: Here, you do not need to update the whole `well_state`. I believe you should only update the info related to the current well, because other wells stay unchanged.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -326,6 +330,34 @@ namespace Opm
...
+                        int reportStepIdx = simulator.episodeIndex();
```
GitPaean, line 342: `const` before `int`.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -520,7 +520,7 @@ namespace Opm
 
         // store a copy of the well state, we don't want to update the real well state
         WellState<Scalar> well_state_copy = well_state;
-        const auto& group_state = simulator.problem().wellModel().groupState();
+        auto group_state = simulator.problem().wellModel().groupState();
```
GitPaean, line 523: you might want to use `group_state_copy` to let people know it is a copy.  Same with `StandardWell_impl.hpp`.

GitPaean (changes requested):
The testing results look good, while it is difficult to understand all the chain consequence from the changes.  Most of the comments are small, the main comment is that you are solving one well, it is probably okay only to update the info related to one well without going through the well state. I approve the PR with minor comments.

multitalentloes:
This does not seem to make it within the (1-day-delayed) RC1 deadline, and subsequently https://github.com/OPM/opm-simulators/pull/6098 as well. If this is considered to an important contribution for this release then I can probably backport it if necessary ahead of the end of RC1, or possibly for the RC2, but no later than that. Thoughts?

totto82:
> I do not deal with release much, but I think you can backport this PR later, it should come in soon after the minor comments addressed.

I did some final testing wrt. robustness for running with different number of processes and I noticed some results I didn't like. So we should not merge this before I understand these new results.

bska:
> I do not deal with release much, but I think you can backport this PR later

I mean technically, yes, but at some point the release manager's judgement/decision will prevent such backporting as those will incur undue testing costs.  I think it's a good rule of thumb to assume that PRs which add new features or change simulator behaviour, even if nominally small, will not be considered eligible for backporting beyond RC2.

GitPaean:
> > I do not deal with release much, but I think you can backport this PR later, it should come in soon after the minor comments addressed.
> 
> I did some final testing wrt. robustness for running with different number of processes and I noticed some results I didn't like. So we should not merge this before I understand these new results.

I appreciate your testing finding. Just let me know when you figure out what to do with this PR. Updating the group related during a single well's solve itself is a little troublesome/conflicting, while we go with testing and hoping the global solve can fix the inconsistency.

totto82:

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -307,12 +307,16 @@ namespace Opm
...
-                    if (! (isGroupControl && !this->param_.check_group_constraints_inner_well_iterations_)) {
-                        changed = this->checkIndividualConstraints(ws, summary_state, deferred_logger, inj_controls, prod_controls);
-                    }
-                    if (hasGroupControl && this->param_.check_group_constraints_inner_well_iterations_) {
-                        changed = changed || this->checkGroupConstraints(well_state, group_state, schedule, summary_state,deferred_logger);
+                    const bool isGroupControl = ws.production_cmode == Well::ProducerCMode::GRUP || ws.injection_cmode == Well::InjectorCMode::GRUP;
+
+                    if (this->param_.check_group_constraints_inner_well_iterations_) {
+                        if (!isGroupControl && hasGroupControl) {
+                            changed = changed || this->checkGroupConstraints(well_state, group_state, schedule, summary_state,deferred_logger);
```
vkip, line 314: Not a big deal, but since changed here is always false this could be simplified (remove 'changed ||' on the rhs)

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -307,12 +307,16 @@ namespace Opm
...
+                        }
+                        changed = changed || this->checkIndividualConstraints(ws, summary_state, deferred_logger, inj_controls, prod_controls);
```
vkip, line 316: This one seems problematic though, since individual constraints will not be checked if 'changed' is true, i.e., if we violate a group constraint. Not sure if it is intentional? In practice it appears to cause issues with wells not respecting individual constraints while on group control...

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -326,6 +330,34 @@ namespace Opm
                         }
                         updatePrimaryVariables(simulator, well_state, deferred_logger);
                     }
+
+                    const bool isNowGroupControl = ws.production_cmode == Well::ProducerCMode::GRUP || ws.injection_cmode == Well::InjectorCMode::GRUP;
```
GitPaean, line 334: Just leaving a note for others here, from my testing, it is this part of the code of change fixes the issue that wells under group control not respecting the nodal pressure, while it was not easy for me to understand how. 

Or it is possible this part of the change is necessary for the whole things to work together to fix the issue.

opm/simulators/wells/WellInterface_impl.hpp
totto82: I have updated it accordingly. I am not sure what the best solution is. Only allow for one switch per iteration or always respect individual constraints.

opm/simulators/wells/WellInterface_impl.hpp
totto82: For my testing, changing this part didn't make any difference in the results.

GitPaean: really? I meant the whole part, 

```c++
const bool isNowGroupControl = ws.production_cmode == Well::ProducerCMode::GRUP || ws.injection_cmode == Well::InjectorCMode::GRUP;
                    if (isGroupControl != isNowGroupControl) {
                        // We need to update the globalIsGRUP vector and the group target reduction.
                        // But we can not communicate it to the other ranks during the local iterations.
                        // The globalIsGrup vector is thus only communicated after the local iterations are done.
                        // This may give some discrepencies in non-linear behaviour between MPI runs,
                        // but the solution should convergece to the same solution.
                        well_state.updateGlobalIsGrup(this->index_of_well_);
                        const int reportStepIdx = simulator.episodeIndex();
                        const Group& fieldGroup = schedule.getGroup("FIELD", reportStepIdx);
                        std::vector<Scalar> groupTargetReduction(well_state.numPhases(), 0.0);
                        WellGroupHelpers<Scalar>::updateGroupTargetReduction(fieldGroup,
                                                                             schedule,
                                                                             reportStepIdx,
                                                                             this->isInjector(),
                                                                             well_state.phaseUsage(),
                                                                             *this->guideRate(),
                                                                             well_state,
                                                                             summary_state,
                                                                             group_state,
                                                                             groupTargetReduction);

                        if (isNowGroupControl) {
                            updateWellStateWithTarget(simulator, group_state, well_state, deferred_logger);
                            updatePrimaryVariables(simulator, well_state, deferred_logger);
                        }
                    }
```

totto82: With "this"  I meant the changes that @vkip suggested.

vkip: No, misunderstanding on my part, sorry about that. Have another suggestion, initial results on a complex model seems promising, but takes a long time to finish. For potential discussion, what I did was basically to replace the block https://github.com/OPM/opm-simulators/blob/d201ecfc84e480cc6273553d8ba3325ba43b0253/opm/simulators/wells/WellInterface_impl.hpp#L303  with the following (highly unpolished) variant:

```c++

                    // Changing to group controls here may lead to inconsistencies in the group handling which in turn 
                    // may result in excessive back and forth switching. However, we currently allow this by default.
                    // The switch check_group_constraints_inner_well_iterations_ is a temporary solution.
                    
                    const bool hasGroupControl = this->isInjector() ? inj_controls.hasControl(Well::InjectorCMode::GRUP) :
                                                                      prod_controls.hasControl(Well::ProducerCMode::GRUP);
                    bool isGroupControl = ws.production_cmode == Well::ProducerCMode::GRUP || ws.injection_cmode == Well::InjectorCMode::GRUP;

                    // We first check for individual controls (unless on group control and switch disabled)
                    if (! (isGroupControl && !this->param_.check_group_constraints_inner_well_iterations_)) {
                        changed = this->checkIndividualConstraints(ws, summary_state, deferred_logger, inj_controls, prod_controls);
                    }
                    const auto well_state_orig = well_state; // Ugly - take a copy for now...
                    // Check for group controls if present and switch not disabled, but only if we did not already change to individual
                    if (hasGroupControl && this->param_.check_group_constraints_inner_well_iterations_) {
                        switched_to_group = this->checkGroupConstraints(well_state, group_state, schedule, summary_state,deferred_logger);
                        changed = changed || switched_to_group;
                    }
                    // Need to update well_state with group target rates and check for violation of the individual rates
                    if (switched_to_group) {
                        updateWellStateWithTarget(simulator, group_state, well_state, deferred_logger);
                        updatePrimaryVariables(simulator, well_state, deferred_logger);

                        const bool changed_individual = this->checkIndividualConstraints(ws, summary_state, deferred_logger, inj_controls, prod_controls);
                        if (! changed_individual) {
                            // We need to update the globalIsGRUP vector and the group target reduction.
                            // But we can not communicate it to the other ranks during the local iterations.
                            // The globalIsGrup vector is thus only communicated after the local iterations are done.
                            // This may give some discrepencies in non-linear behaviour between MPI runs,
                            // but the solution should convergece to the same solution.
                            well_state.updateGlobalIsGrup();
                            int reportStepIdx = simulator.episodeIndex();
                            const Group& fieldGroup = schedule.getGroup("FIELD", reportStepIdx);
                            std::vector<Scalar> groupTargetReduction(well_state.numPhases(), 0.0);
                            WellGroupHelpers<Scalar>::updateGroupTargetReduction(fieldGroup,
                                                                                schedule,
                                                                                reportStepIdx,
                                                                                this->isInjector(),
                                                                                well_state.phaseUsage(),
                                                                                *this->guideRate(),
                                                                                well_state,
                                                                                summary_state,
                                                                                group_state,
                                                                                groupTargetReduction);
                            // No need to update target again here?
                            // updateWellStateWithTarget(simulator, group_state, well_state, deferred_logger);
                            // updatePrimaryVariables(simulator, well_state, deferred_logger);
                        } else {
                            // New group target violates an individual constraint - reset
                            well_state = well_state_orig;
                            changed = false;
                        }
                    } else { // We did not switch to group, update individual target if present
                        if (changed) {
                            const bool thp_controlled = this->isInjector() ? ws.injection_cmode == Well::InjectorCMode::THP : ws.production_cmode == Well::ProducerCMode::THP;
                            if (!thp_controlled){
                            // don't call for thp since this might trigger additional local solve
                            updateWellStateWithTarget(simulator, group_state, well_state, deferred_logger);
                            } else {
                            ws.thp = this->getTHPConstraint(summary_state);
                            }
                            updatePrimaryVariables(simulator, well_state, deferred_logger);
                        }
                    }
                }
                return changed;


```

totto82: Thanks for the code. I assume this will disallow switching to GRUP if it violates individual constraints. If you could send me the patch by email (or make a PR to my PR branch), I can add the commit to this PR.

vkip: That's the intention at least...  Patch incoming!

vkip: Aaargh, hold on - I see now that I forgot to update reduction rates for wells that switch _from_ group...

totto82: The patch didn't apply directly, so I have copied its content and fixed your "aargh" in the new commit.

totto82:

totto82:

totto82:

totto82:

totto82:
Most of the testing is done, and this can be merged. But since it impacts almost all cases, mostly minor impacts though, I don't think we should include it in the release. Users still have the option of setting-- check-group-constraints-inner-well-iterations=false if they stumble on rare(?) cases where local group constraint checking is inconsistent and could lead to oscillations.

totto82:

PR #5925, opened 2025-01-29
janitoring: use string concat instead of operator+

akva2:
and avoid unnecessary temporary


akva2:
jenkins build this please

opm/simulators/flow/FlowGenericVanguard.cpp
```diff
@@ -349,12 +349,10 @@ void FlowGenericVanguard::init()
...
             {
-                std::string message =
-                        std::string("Option --allow-distributed-wells=true in a model with\n")
-                        + "multisegment wells. This feature is still experimental. You can\n"
-                        + "set --use-multisegment-well=false to treat the existing \n"
-                        + "multisegment wells as standard wells.";
-                OpmLog::info(message);
+                OpmLog::info("Option --allow-distributed-wells=true in a model with\n"
+                             "multisegment wells. This feature is still experimental. You can\n"
+                             "set --use-multisegment-well=false to treat the existing \n"
```
bska, line 354: If you're changing this anyway, maybe you could remove the trailing blank here.  In other words, change `...existing<SPACE>\n` into `...existing\n` instead?

akva2: sure. didn't notice.

akva2:
jenkins build this please

bska (approved):

PR #5926, opened 2025-01-29
hacking to register the parameter ForceDisableFluidInPlaceOutput

GitPaean:
Instantiating FluidSystem<0> at this stage due to how it is implemented. 

After https://github.com/OPM/opm-simulators/pull/5882, the flowexp_comp can not be run anymore. 

Not totally sure this is the proper fix, but just bring it up here to start the conversation.

### DIFF
--- a/opm/simulators/flow/GenericOutputBlackoilModule.cpp
+++ b/opm/simulators/flow/GenericOutputBlackoilModule.cpp
@@ -1857,6 +1857,8 @@ INSTANTIATE_TYPE(float)
     template<class T> using FS##NUM = GenericOilGasFluidSystem<T, NUM>; \
     template class GenericOutputBlackoilModule<FS##NUM<double>>;
 
+INSTANTIATE_COMP(0) // \Note: to register the parameter ForceDisableFluidInPlaceOutput
+
 INSTANTIATE_COMP(2)
 INSTANTIATE_COMP(3)
 INSTANTIATE_COMP(4)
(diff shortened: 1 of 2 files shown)


GitPaean:
jenkins build this please

akva2:
i see no hack here. also don't see why you override if you're just going to call the base class implementation.

GitPaean:
> i see no hack here. also don't see why you override if you're just going to call the base class implementation.

It is a good point. There was an empty function there before and I did not think about it.

akva2:
i get that but that was also the problem in the first place :)

GitPaean:
> > i see no hack here. also don't see why you override if you're just going to call the base class implementation.
> 
> It is a good point. There was an empty function there before and I did not think about it.

The empty function might indicate they wanted some local parameters for this module, but there was not yet.

GitPaean:
because we use the following code in flowexp_comp.cpp, 

```c++
int
main(int argc, char** argv)
{
    using TypeTag = Opm::Properties::TTag::FlowExpCompProblem<0>;
    Opm::registerEclTimeSteppingParameters<double>();
```

we need `INSTANTIATE_COMP(0)` in `GenericOutputBlackoilModule.cpp`.

GitPaean:
jenkins build this please

GitPaean:
Can this go in or there is suggestion for alternative solutions?

GitPaean:
jenkins build this please

akva2:
sorry, i forgot to answer here. yes it looks fine. the registering for 0 is unfortunate but the interfaces are as they are so this is fine for now.

GitPaean:
> sorry, i forgot to answer here. yes it looks fine. the registering for 0 is unfortunate but the interfaces are as they are so this is fine for now.

thanks for the comment. I am getting it in now so I do not need to keep rebase my local master branches.

PR #5927, opened 2025-01-29
Implement Property Tree Put/Get for std::size_t

bska:
This replaces the original specialisations for "bare" `size_t` from the `<stddef.h>` header.

While here, also reorder the specialisations to match declaration order in the header.


bska:
This is the follow-up to #5921 that was suggested in the [discussion](https://github.com/OPM/opm-simulators/pull/5921#discussion_r1933487848) of that PR.

bska:
jenkins build this please

akva2 (approved):

bska:

PR #5801, opened 2024-12-16
Add Infrastructure for Geomechanical Analysis

bska:
In particular, this lays the foundation for adding new well connections as a result of fracturing processes.

PR #5928, opened 2025-01-29
Move Dynamic Dispatch Function to Separate TU

bska:
This moves about 300 lines of code out of `Main.hpp` and, especially, moves the `<flow/flow_*.hpp>` include statements as well.  This, in turn, makes `Main::runStatic<>()` usable for out-of-tree consumers.


bska:
I am not sure that this is the best way to go about solving my problem, but my main goal here&ndash;in support of #5801&ndash;is to make `Main::runStatic<>()` usable for out-of-tree consumers with custom `TypeTag` implementations.  Having the `#include <flow/flow_*.hpp>` statements directly in `Main.hpp` makes that goal more difficult and effectively leads to having to install those headers despite them not having any publicly accessible symbols.

I welcome all suggestions for alternative strategies here.

bska:
jenkins build this please

akva2:
This entirely makes sense to me, but we need to make the FlowProblemTPFA available in some contexts where it is now missing (build with python). While that particular problem is solvable by including flow_blackoil.hpp in PyMain.hpp, for ..reasons (some machine learning extensions for flow that i'm maintaining), I'd really like if we could put it in a separate (installed) header.

bska:
> we need to make the `FlowProblemTPFA` available in some contexts where it is now missing (build with python)

Right.  I made this PR based on a build configuration without `EMBEDDED_PYTHON` so I didn't run into the problem until I tried a build check.

> While that particular problem is solvable by including flow_blackoil.hpp in PyMain.hpp

PR #5930 does essentially this.

> I'd really like if we could put it in a separate (installed) header.

The `FlowProblemTPFA` you mean?

bska:
> yes please. while it is very few lines of code, I'd really like to not duplicate them in my downstream.

:+1: 

Any preference for the location?  Do we still keep it in opm-simulators/flow, or is there some other location that might be better?

akva2:
i'd say put it in opm/simulators/flow to keep the opm-simulators/flow folder strictly no-install.

bska:
> i'd say put it in opm/simulators/flow to keep the opm-simulators/flow folder strictly no-install.

Good idea.  Please have a look at PR #5931.

bska:
The earlier, preparatory, PRs have been merged.  I'm marking this as "ready for review" and I welcome any comments you may have.

bska:
jenkins build this please

opm/simulators/flow/Main.hpp
```diff
@@ -189,84 +158,12 @@ class Main
     }
 
 private:
```
akva2, line 160: we can maybe move this block below so we don't need two private: ?

bska: > we can maybe move this block below so we don't need two private: ?

Good idea.  I've pushed an update to that effect and, while here, also added some skeletal documentation to `(run|dispatch)(Dynamic|Static)`.

bska:
jenkins build this please

akva2 (approved):

bska:

PR #5933, opened 2025-01-30
Add optional iterations in the group switching algorithm

totto82:
Default is 1 i.e. it will only check once.

### DIFF
--- a/opm/simulators/flow/BlackoilModelParameters.cpp
+++ b/opm/simulators/flow/BlackoilModelParameters.cpp
@@ -108,6 +108,9 @@ BlackoilModelParameters<Scalar>::BlackoilModelParameters()
     monitor_params_.decay_factor_ = Parameters::Get<Parameters::ConvergenceMonitoringDecayFactor<Scalar>>();
 
     nupcol_group_rate_tolerance_ = Parameters::Get<Parameters::NupcolGroupRateTolerance<Scalar>>();
+    group_control_max_iterations_ = Parameters::Get<Parameters::GroupMaxIterations>();
+
+
 }
 
 template<class Scalar>
@@ -269,6 +272,9 @@ void BlackoilModelParameters<Scalar>::registerParameters()
         ("Tolerance for acceptable changes in VREP/RAIN group rates");
 
     Parameters::Hide<Parameters::DebugEmitCellPartition>();
+
+    Parameters::Register<Parameters::GroupMaxIterations>
+    ("Maximum number of iterations in the group solver algorithm");
 
     // if openMP is available, use two threads per mpi rank by default
 #if _OPENMP
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1936,56 +1936,62 @@ namespace Opm {
             more_network_update = more_network_sub_update || well_group_thp_updated;
         }
 
+        int iter = 0;
         bool changed_well_group = false;
-        // Check group individual constraints.
         const Group& fieldGroup = this->schedule().getGroup("FIELD", episodeIdx);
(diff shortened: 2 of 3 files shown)


totto82:
This if first part of https://github.com/OPM/opm-simulators/pull/5607

totto82:
jenkins build this please

akva2:
mind the warning(s).

totto82:
jenkins build this please

totto82:
This can be merged as is. But we should also factor out the network part and remove any well solve from updateWellStateWithTarget before considering using this feature. I will make another PR where the network part is refactored to a separate call asap.

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
Since default is 1 it should be safe to merge this as is. More testing is needed if we want to change the default.

GitPaean:
Hei, the code change looks straightforward, I think this is towards the good directions, which hopefully make sure the well controls are settled for each iteration before we go ahead to assemble. 

From the testing side, do you find this helps any specific case and what is your observation, thanks.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1808,56 +1808,62 @@ namespace Opm {
...
-        changed_well_individual = comm.sum(static_cast<int>(changed_well_individual));
-        if (changed_well_individual) {
-            updateAndCommunicate(episodeIdx, iterationIdx, deferred_logger);
-            changed_well_group = true;
+            changed_well_individual = comm.sum(static_cast<int>(changed_well_individual));
+            if (changed_well_individual) {
+                updateAndCommunicate(episodeIdx, iterationIdx, deferred_logger);
+                changed_well_group = true;
+            }
+            iter++;
```
GitPaean, line 1866: What should the practice for this type of loop, do we need to use comm.barrier() here at the end of the loop?

@blattms @akva2

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1808,56 +1808,62 @@ namespace Opm {
             more_network_update = more_network_sub_update || well_group_thp_updated;
         }
 
+        size_t iter = 0;
         bool changed_well_group = false;
-        // Check group individual constraints.
         const Group& fieldGroup = this->schedule().getGroup("FIELD", episodeIdx);
-        changed_well_group = updateGroupControls(fieldGroup, deferred_logger, episodeIdx, iterationIdx);
+        // Check group individual constraints.
+        // iterate a few times to make sure all constrains are kept
```
GitPaean, line 1815: `are kept` should be `are honored`?

opm/simulators/flow/BlackoilModelParameters.cpp
```diff
@@ -276,6 +279,9 @@ void BlackoilModelParameters<Scalar>::registerParameters()
 
     Parameters::Hide<Parameters::DebugEmitCellPartition>();
 
+    Parameters::Register<Parameters::GroupMaxIterations>
+    ("Maximum number of iterations in the group solver algorithm");
```
GitPaean, line 283: The iteration not only handles the group control, when a well has multiple constraints itself, an iteration process might also be performed (also it is not that of a problem compared with group or network related).  We need to get a better  or more correct name?

totto82: What about WellGroupConstraintsMaxIterations?

GitPaean: It is better and usable. We can use it before we find something else.

totto82:
jenkins build this please

GitPaean:
Thanks for the update. It looks good to me. I am merging it to make easier for people to test.

PR #5934, opened 2025-01-30
Install More Headers

bska:
Needed for `Main::runStatic<>()` to be usable out-of-tree.


bska:
This is extracted from the original version of PR #5801.

bska:
jenkins build this please

akva2 (approved):

bska:

bska:
> Are these all headers now?

blattms:
My two cents:
I think it would be much easier to maintain the list if we always install all headers and use ifdef to deactivate code inside of them.

akva2:
and mine: it's much better to take on the complexity in the build system than to pay for it with spaghetti in the code.

PR #5939, opened 2025-01-31
OutputBlackoilModule: add a container for the FIP data

akva2:
Try to clean up this code a little. Still some mysteries I haven't gotten my head around.


akva2:
Okay, this is now ready for review. The code is still somewhat finicky and confusing, but I think this puts us in a better place to do cleanup.

akva2:
jenkins build this please

bska (approved):
Greatly appreciated.  Packaging those `! empty()` calls into a nicer interface is a very welcome change.  I do foresee some potential future improvements in this area.  We almost always use these values aggregated to at least the region level, and often the field level, so we might not always need to allocate full per-cell storage.  As I said, though, that's potential follow-up work so for now I'll merge this into the master branch.

PR #5940, opened 2025-02-03
Remove Unused FMT Header

bska:
Makes using the header from an installed location harder than it needs to be.


bska:
Extracted from #5801.

bska:
jenkins build this please

akva2 (approved):

bska:

PR #5943, opened 2025-02-03
modelTests: use separate output directories

akva2:
Diff is annoyingly big but the simulators throw on non-existent output directories.


akva2:
jenkins build this please

PR #5945, opened 2025-02-03
Limit Haaland friction factor to laminar value

vkip:


vkip:
jenkins build this please

vkip:

vkip:

vkip:
jenkins build this please

opm/simulators/wells/MSWellHelpers.cpp
```diff
@@ -54,12 +54,14 @@ ValueType haalandFormular(const ValueType& re,
...
-
-    // sqrt(1/f) should be non-positive
-    assert(value >= 0.0);
-
-    return 1. / (value * value);
+    // Since we currently do not guard for high roughness values in input we can end up with unrealistic friction factors.
+    // In particular, there could be a singularity in f(Re) if 1 \in image(X) in log10(X) below. In calculations we therefore
+    // limit the relative roughness to ensure the singularity never occurs in the relevant range (Re >= 4000).
+    assert( re >= 4000. );
+    constexpr Scalar MAX_REL_ROUGHNESS = 3.7 * std::pow((1.0 - 1.0e-3) - 6.9/4000.0, 9. / 10.);
```
GitPaean, line 61: It looks fine.

It is also a value can be evaluated in the parsing stage, should we enforce this when parsing the keyword WELSEGS so that we can give a warning message pointing to the DATA file?

The main difference is that we will also enforce this value to other Reynolds number range. 

For the 3D_MSW.DATA in the opm-tests, the `MAX_REL_ROUGHNESS` calculated is `3.69` and the absolute roughness will be 3.69 * 0.3 > 1.

I am okay with the current approach while discussing a little bit whether we do this in the parsing stage to give the users some info when too big value is specified on purpose or by accident.

vkip: I see your point, probably would be good with a warning and limit roughness while parsing, but maybe still have an assertion here that the relative roughness is in range (in case of some unforeseen future use-case, like setting the roughness via Python..)?

I can create a companion in opm-common and update this one.

akva2: you said it yourself - this is potentially user provided data -> throw, not assert.

vkip:
https://github.com/OPM/opm-common/pull/4461

GitPaean (approved):
It looks good to me. Awaiting the merging of the upstream PR.

GitPaean:
Upstream PR is merged. I am getting this one in also as promised.

PR #5952, opened 2025-02-04
Consistent primary variables

hnil:
Draft to see if make primary variables more consistent is good.

### DIFF
--- a/opm/simulators/wells/MultisegmentWellPrimaryVariables.cpp
+++ b/opm/simulators/wells/MultisegmentWellPrimaryVariables.cpp
@@ -152,6 +152,7 @@ update(const WellState<Scalar>& well_state,
             }
         }
     }
+    init();
 }
 
 template<class FluidSystem, class Indices>
@@ -208,6 +209,7 @@ updateNewton(const BVectorWell& dwells,
     if (stop_or_zero_rate_target) {
         value_[0][WQTotal] = 0.;
     }
+    init();
 }
 
 template<class FluidSystem, class Indices>
--- a/opm/simulators/wells/StandardWellPrimaryVariables.cpp
+++ b/opm/simulators/wells/StandardWellPrimaryVariables.cpp
@@ -223,6 +223,7 @@ update(const WellState<Scalar>& well_state,
 
     // BHP
     value_[Bhp] = ws.bhp;
+    init();
 }
 
 template<class FluidSystem, class Indices>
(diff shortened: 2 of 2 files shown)


GitPaean:
jenkins build this please

GitPaean:

atgeirr:
I think this should be merged, but with the assert()s removed. Afterwards we should remove all `initPrimaryVariablesEvaluation()` functions and calls, that should be a pure refactoring at that point.

I will rerun jenkins to look at the failures.

atgeirr:

hnil:
Maybe also merge the well copy pullrequest. This seems to remove may of the troubling cases.

GitPaean:

GitPaean:
It is the following cases having visible summary curve deviations.  from pdf file names.

01_spe1_metric_vfp1
03_9_3a_grpctl_msw_model2
04_9_3b_grpctl_msw_model2
07_9_4d_grpctl_msw_model2
05_9_2a_grpctl_msw_model2
12_9_4c_grpctl_msw_model2
06_udq_wconprod
08_9_3d_grpctl_msw_model2

GitPaean:
> I think this should be merged, but with the assert()s removed. Afterwards we should remove all initPrimaryVariablesEvaluation() functions and calls, that should be a pure refactoring at that point.

Let us do it in this PR. 

@hnil , maybe you can try to remove all the function calls related to `initPrimaryVariablesEvaluation()` including the functions themselves?

atgeirr:
> 01_spe1_metric_vfp1

This case just has a big relative change in some gas saturations that are around 1e-16, so not something we should worry about.

> 03_9_3a_grpctl_msw_model2
> 04_9_3b_grpctl_msw_model2
> 07_9_4d_grpctl_msw_model2
> 05_9_2a_grpctl_msw_model2
> 12_9_4c_grpctl_msw_model2
> 06_udq_wconprod
> 08_9_3d_grpctl_msw_model2

These cases all have more significant changes, after some point where the timestep is different. I think they are probably JUST due to the timestep changes, and will investigate one or two of them to see this more clearly.



> > I think this should be merged, but with the assert()s removed. Afterwards we should remove all initPrimaryVariablesEvaluation() functions and calls, that should be a pure refactoring at that point.
> 
> Let us do it in this PR.

I think it is easier to just get this in first.

atgeirr:
The affected test cases are very sensitive to time steps. Long-term they should be stabilized, for now it is enough to do a data update.

atgeirr:
jenkins build this opm-tests=1291 please

atgeirr:
jenkins build this opm-tests=1291 please

PR #5954, opened 2025-02-04
Add a container for FLOWS data output

akva2:
Sits on top of https://github.com/OPM/opm-simulators/pull/5950

### DIFF
--- a/opm/simulators/flow/FlowsContainer.cpp
+++ b/opm/simulators/flow/FlowsContainer.cpp
@@ -24,12 +24,34 @@
 
 #include <opm/input/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp>
 #include <opm/input/eclipse/Schedule/Schedule.hpp>
+#include <opm/input/eclipse/Units/UnitSystem.hpp>
 
 #include <opm/material/fluidsystems/BlackOilDefaultIndexTraits.hpp>
 #include <opm/material/fluidsystems/BlackOilFluidSystem.hpp>
 #include <opm/material/fluidsystems/GenericOilGasWaterFluidSystem.hpp>
 
+#include <opm/output/data/Solution.hpp>
+
 #include <algorithm>
+#include <tuple>
+
+namespace {
+
+    template<class Scalar>
+    using DataEntry = std::tuple<std::string, Opm::UnitSystem::measure, std::vector<Scalar>&>;
+
+    template<int idx, class Array, class Scalar>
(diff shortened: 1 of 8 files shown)


akva2:
jenkins build this opm-common=4460 please

bska:
> Sits on top of #5950

That PR was merged into master.  Is this work ready for review now, @akva2 ?

akva2:
as such, yes. however, there is incoming work in https://github.com/OPM/opm-common/pull/4410 that will affect this. in particular I can simplify this as there is no longer a -1 index for the water phase in the generic fluid system, which means i remove some additional template stuff I had to do to avoid compiler warnings. I think we should wait for that and then I can redo.

bska:
> there is incoming work in [OPM/opm-common#4410](https://github.com/OPM/opm-common/pull/4410) that will affect this. [...] I think we should wait for that and then I can redo.

Noted.  Thanks a lot for the information.

akva2:
turns out the component index is still -1 for water, only the phase index changes, so i couldn't really simplify things. but in any case, this is now ready for review.

akva2:
jenkins build this please

akva2:
jenkins build this please

opm/simulators/flow/FlowsContainer.cpp
```diff
@@ -0,0 +1,335 @@
...
+FlowsContainer<FluidSystem>::
+FlowsContainer(const Schedule& schedule,
+               const SummaryConfig& summaryConfig)
+{
+    // Check for any BFLOW[I|J|K] summary keys
+    blockFlows_ = summaryConfig.keywords("BFLOW*").size() > 0;
+
+    // Check if FLORES/FLOWS is set in any RPTRST in the schedule
+    anyFlores_ = false;     // Used for the initialization of the sparse table
+    anyFlows_ = blockFlows_;
```
bska, line 96: These assignments will be overwritten by the `any_of()` calls below.  Static analysis may or may not flag that as unnecessary writes.

bska (commented):
Excellent.  I think this is ready for merging once the final tiny issue has been addressed.

opm/simulators/flow/FlowsContainer.cpp
akva2: indeed. inherited from original code structure.

akva2:
jenkins build this please

akva2:
jenkins build this please

bska (approved):

PR #5960, opened 2025-02-06
Add MechContainer

akva2:
This is a container for the data output related to mechanics.


akva2:
jenkins build this please

opm/simulators/flow/MechContainer.cpp
```diff
@@ -0,0 +1,216 @@
...
+    this->potentialPressForce_.resize(bufferSize, 0.0);
+    rstKeywords["PRESPOTF"] = 0;
+
+    std::string disp = "DISPX";
+    std::for_each(disp_.begin(), disp_.end(),
+                  [bufferSize,&rstKeywords,&disp](auto& v)
+                  {
+                      v.resize(bufferSize, 0.0);
+                      rstKeywords[disp] = 0;
+                      ++disp.back();
```
bska, line 55: This&ndash;i.e., `++disp.back()`&ndash;makes assumptions about the basic execution character set that are not, strictly speaking, guaranteed by the standard.  The assumption does hold for ASCII and, more importantly, UTF-8, but we can't **really** assume that the integer values satisfy`'Y' == 'X' + 1` or that `'Z' == 'Y' + 1`.  Very strictly speaking we only know that those integer values are non-negative and distinct from each other.  On the other hand, the standard **does** guarantee that such arithmetic progression holds for the integer values of the character literals `'0'` through `'9'` in the basic execution character set.

One guaranteed safe option is to write something like
```C++
[suffix = "XYZ", is = 0, bufferSize, &rstKeywords](auto &v) mutable
{
    v.resize(bufferSize, 0.0);
    rstKeywords[std::string{"DISP"} + suffix[is++]] = 0;
}
```
(or variations thereof) instead.

bska (commented):
Sweet!  I really like the simplification and automation this brings.  I just spotted one, very small, potential issue that we might want to address before merging this, but otherwise this is a **really** good contribution.  Thanks a lot!

opm/simulators/flow/MechContainer.cpp
akva2: yeah correct you are. done exactly like this. i just re-learned the mutable keyword,.

akva2:
jenkins build this please

bska (approved):
Very good, thanks a lot for the updates.  This looks good to me now and I'll merge into master.

PR #5962, opened 2025-02-06
Added tests for error macros on GPU.

kjetilly:
Some simple tests for the new error macros defined in [the opm-common PR 4465](https://github.com/OPM/opm-common/pull/4465).

### DIFF
--- a/tests/gpuistl/test_throw_macros_on_gpu.cu
+++ b/tests/gpuistl/test_throw_macros_on_gpu.cu
@@ -1,0 +1,109 @@
+/*
+  Copyright 2025 Equinor ASA
+
+  This file is part of the Open Porous Media project (OPM).
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#include <boost/test/tools/old/interface.hpp>
+#include <config.h>
+#include <stdexcept>
+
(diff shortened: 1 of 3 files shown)


tests/gpuistl/test_throw_macros_on_gpu.cu
```diff
@@ -0,0 +1,109 @@
...
+#include <cuda.h>
+#include <cuda_runtime.h>
+#include <boost/test/unit_test.hpp>
+#include <opm/common/ErrorMacros.hpp>
+#include <opm/simulators/linalg/gpuistl/detail/gpu_safe_call.hpp>
+
+namespace {
+__global__ void codeThatContainsMacros(bool call) {
+    if (call) {
+        OPM_THROW(std::logic_error, "Something went wrong");
```
multitalentloes, line 31: I think something has gone wrong with the macro that assess whether we are inside a GPU kernel or not, as I get an error on an AMD gpu as it expands the wrong macro from `ErrorMacros.h`. It should have expanded the macro on line 110 in `ErrorMacros.hpp`
> opm-simulators/tests/gpuistl_hip/test_throw_macros_on_gpu.hip:32:9: error: no matching constructor for initialization of 'std::string' (aka 'basic_string<char>')
>    32 |         OPM_THROW(std::logic_error, "Something went wrong");
>       |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> opm-common/opm/common/ErrorMacros.hpp:56:28: note: expanded from macro 'OPM_THROW'
>    56 |         std::string oss_ = std::string{"["} + __FILE__ + ":" + \
>

multitalentloes: minimal example that can not be compiled on HIP
> #include <hip/hip_runtime.h>
>#include <cstdio>
>
>__global__ void test_kernel() {
>    #if  defined(__CUDA_ARCH__) || (defined(__HIP_DEVICE_COMPILE__) && __HIP_DEVICE_COMPILE__ > 0)
>        printf("Hello from HIP kernel!\n");
>    #else
>        static_assert(false, "This should not be compiled for host");
>    #endif
>}
>
>int main(int, char**) {
>    test_kernel<<<1, 1>>>();
>    hipDeviceSynchronize();
>    return 0;
>}

kjetilly:
I think this should now compile on HIPCC, @multitalentloes

PR #5964, opened 2025-02-07
Add a container for MICP data output

akva2:
Continuing this encapsulation and separation of concerns journey.

### DIFF
--- a/opm/simulators/flow/MICPContainer.cpp
+++ b/opm/simulators/flow/MICPContainer.cpp
@@ -27,10 +27,10 @@
 
 #include <opm/output/data/Solution.hpp>
 
+#include <algorithm>
 #include <array>
 #include <string>
 #include <tuple>
-
 
 namespace Opm {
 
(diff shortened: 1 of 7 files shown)


opm/simulators/flow/MICPContainer.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    using DataEntry =
+        std::tuple<std::string, UnitSystem::measure, const std::vector<Scalar>&>;
+
+    auto doInsert = [&sol](const DataEntry& entry)
+    {
+        if (!std::get<2>(entry).empty()) {
+            sol.insert(std::get<std::string>(entry),
+                       std::get<UnitSystem::measure>(entry),
+                       std::move(std::get<2>(entry)),
```
bska, line 98: We're in a `const` member function and `get<2>(entry)` is a `const` object.  Applying `std::move()` to that will just copy and leave the original intact.  In the current master sources, we actually move the contents of the source array into the `Solution` object and that's why the `.empty()` calls are "everywhere".  It's not entirely clear to me that we preserve the full semantics by marking `outputRestart()` as `const` in this case.

bska (commented):
This looks good for the most part.  The one aspect that I'm not entirely sure about is marking `outputRestart()` as `const` and what semantic implications come from doing so.

opm/simulators/flow/MICPContainer.cpp
akva2: yeah that's an aspect of this i did not consider and compiler did not tell me about.

first, from a conceptual pov it is not really intuitive that an output function destroys the container. 

furthermore [the tuple entry](https://github.com/OPM/opm-simulators/blob/master/opm/simulators/flow/GenericOutputBlackoilModule.cpp#L412)  where I copied the code from always was a const vector<Scalar>&, the move has always been moot.

bska: > the move has always been moot.

Maybe not "always", but at least for a very long time.  In any case, we do have logic in place that depends on the `move()` actually moving, but as long as all output vectors continue to output once they've been written once that latent error would have been hidden.

akva2: yeah, i just wanted to point out the problem. If these are the expected semantics I will implement them of course, but then we should fix the problem in the generic code as well.

akva2: done

akva2:
jenkins build this please

bska (approved):
Very good&ndash;I appreciate the increased abstraction level here.  I'll merge into master.

PR #5967, opened 2025-02-07
Glift fixes

hnil:
Draft pullrequest on using simple solves for gaslift.
It was tested agains the original code and gave the same for all cases on gaslift-03, but made some wells be open for low rates and large alq.
Currently it is not much faster due to to much copying of wellstate which should not be need.


hnil:
Closing as stale — the gas-lift code has moved on considerably since this was written in February 2025, and the branch no longer merges. The branch stays on hnil if the simple-solve idea is worth revisiting.

PR #5969, opened 2025-02-08
Using the CriticalException in PreconditionerFactory and ISTLSolver

kjetilly:
This pull request uses the new `CriticalError` exception class introduced the [opm-common PR 4467]( https://github.com/OPM/opm-common/pull/4467) to rethrow some exceptions happening in the initialisation of the linear solver as critical.

I am open to discussion if this is the correct level to introduce this exception. The reasoning behind the `prepare` choice was simply that in `initPrepare`, one access ```"preconditioner.type"``` which can lead to exceptions with ill-formed JSON input. It is also doubtful that any error occurring in `prepare` can be fixed by retrying with a smaller timestep.

### DIFF
--- a/opm/simulators/linalg/ISTLSolver.hpp
+++ b/opm/simulators/linalg/ISTLSolver.hpp
@@ -378,11 +378,11 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
         void prepare(const Matrix& M, Vector& b)
         {
             OPM_TIMEBLOCK(istlSolverPrepare);
-            OPM_BEGIN_TRY_CATCH_RETHROW_AS_CRITICAL_ERROR();
+            try {
                 initPrepare(M,b);
 
                 prepareFlexibleSolver();
-            OPM_END_TRY_CATCH_RETHROW_AS_CRITICAL_ERROR();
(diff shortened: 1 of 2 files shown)


opm/simulators/linalg/ISTLSolver.hpp
```diff
@@ -377,10 +378,11 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
         void prepare(const Matrix& M, Vector& b)
         {
             OPM_TIMEBLOCK(istlSolverPrepare);
+            OPM_BEGIN_TRY_CATCH_RETHROW_AS_CRITICAL_ERROR();
```
atgeirr, line 381: I agree this is the correct scope. I would like to avoid the macros, see the upstream PR review.

The message should include the "LinearSolver" configuration asked for (most likely something.json if we end up throwing here), you can get that from `parameters_[activeSolverNum_].linsolver_`.

kjetilly: I do not agree that this is the correct place to mention anything about the JSON file, since we can not guarantee that an exception caught here actually comes from a bad JSON file, it is simply one possibility. The point here is to in a certain sense say "Any error occurring in this function can not be fixed by lowering the timestep", and it just so happens that one error is a malformatted JSON file.

I would suggest that I afterwards make a PR where we check if the property_tree has the element and if not, throw a CriticalError with a more descriptive error message. The current macro in place would still protect us from other errors occurring in said function, though.

PR #1049, opened 2017-02-09
Import adaptive time stepping and associated classes

akva2:
As requested, this imports the AdaptiveTimeStepping class from opm-core.
Do not merge this branch as-is, it's a bit interleaved with other fixes so I will have to rewrite it to reflect reality as seen in upstream before it is merged, but I open early so nobody else redo the same work (it's just a matter of rerunning my script).

PR #5974, opened 2025-02-11
General third order time step controller

erikhide:
I've implemented an alternative time step controller which can be used instead of the PID controller (with or without target iterations). Hopefully, this can be merged without any issues, as I've not changed the default in any way. I had to change the input to computeTimeStepSize() since I needed to access the substepTimer object.


erikhide:
> Is input values for the pid-controller changed and scaled better.

I assume you mean the relative change? No, not in this PR. Changing that would mean changing the default numerical algorithm, which seems to require a lot more work before it can be merged (due to test failures). I have a different PR (#4856) with the relative change fix, but that one isn't up to date with the current master.

This PB only includes an alternative controller (with the same relative change as input). It can be used by adding `--time-step-control="general3rdorder"` when running Flow.

akva2:
please add a test, see https://github.com/OPM/opm-simulators/blob/master/tests/test_RestartSerialization.cpp#L122

erikhide:
> please add a test

Done!

bska:
jenkins build this please

opm/simulators/timestepping/TimeStepControl.cpp
```diff
@@ -73,7 +74,7 @@ namespace Opm
     }
 
     double SimpleIterationCountTimeStepControl::
-    computeTimeStepSize( const double dt, const int iterations, const RelativeChangeInterface& /* relativeChange */, const double /*simulationTimeElapsed */) const
+    computeTimeStepSize( const double dt, const int iterations, const RelativeChangeInterface& /* relativeChange */, const AdaptiveSimulatorTimer& substepTimer) const
```
bska, line 77: Note: The `subStepTimer` is unused in this function (`SimpleIterationCountTimeStepControl::computeTimeStepSize()`) so I suggest you either remove the name or tag the parameter as `[[maybe_unused]]`.

opm/simulators/timestepping/TimeStepControl.cpp
```diff
@@ -134,10 +135,10 @@ namespace Opm
     }
 
     double HardcodedTimeStepControl::
-    computeTimeStepSize( const double /*dt */, const int /*iterations */, const RelativeChangeInterface& /* relativeChange */ , const double simulationTimeElapsed) const
+    computeTimeStepSize( const double /*dt */, const int /*iterations */, const RelativeChangeInterface& /* relativeChange */ , const AdaptiveSimulatorTimer& substepTimer) const
     {
-        auto nextTime = std::upper_bound(subStepTime_.begin(), subStepTime_.end(), simulationTimeElapsed);
-        return (*nextTime - simulationTimeElapsed);
+        auto nextTime = std::upper_bound(subStepTime_.begin(), subStepTime_.end(), substepTimer.simulationTimeElapsed());
+        return (*nextTime - substepTimer.simulationTimeElapsed());
```
bska, line 141: Is there ever a risk that `nextTime == subStepTime_.end()`?  If so, then the `return` statement will dereference `end()`.

opm/simulators/timestepping/TimeStepControl.cpp
```diff
@@ -279,4 +280,91 @@ namespace Opm
...
+
+    double General3rdOrderController::
+    computeTimeStepSize(const double dt, const int /*iterations */, const RelativeChangeInterface& relChange, const AdaptiveSimulatorTimer& substepTimer) const
+    {
+        // Shift errors and time steps
+        for( int i = 0; i < 2; ++i )
+        {
+            errors_[i] = errors_[i+1];
+            timeSteps_[i] = timeSteps_[i+1];
+        }
```
bska, line 318: I was going to say "That's a [`rotate()`](https://en.cppreference.com/w/cpp/algorithm/rotate)" :tm:, but given that the index range is tiny that would be more trouble than it's worth.

opm/simulators/timestepping/TimeStepControl.hpp
```diff
@@ -32,7 +33,8 @@ namespace Opm
       SimpleIterationCount,
       PID,
       PIDAndIterationCount,
-      HardCodedTimeStep
+      HardCodedTimeStep,
+      General3rdOrder
```
bska, line 37: I suggest adding a trailing comma here (`General3rdOrder`) to simplify future maintenance.

opm/simulators/timestepping/TimeStepControl.cpp
```diff
@@ -279,4 +280,91 @@ namespace Opm
...
+            if (counterSinceFailure_ > 1)
+                counterSinceFailure_ = 0;
+            const double newDt = dt * std::pow(safetyFactor_ * tolerance_ / errors_[2], 0.35);
+            return newDt;
+        }
+        // Use the general third order controller for all other time steps
+        else
+        {
+            const std::vector<double> beta = { 0.125, 0.25, 0.125 };
+            const std::vector<double> alpha = { 0.375, 0.125 };
```
bska, line 349: These can probably be `std::array`s instead to avoid memory allocation.

bska (commented):
I've just looked at the high-level aspects of this work and not reviewed the actual meat of the algorithm.

opm/simulators/timestepping/TimeStepControl.cpp
multitalentloes: I think this would have been a rotation if the for loop when up to and including 2, and then used the index[(i+1)%3].

Now errors_[2] after the operation is unaffected, whereas it would have been equal to errors[0] if this was a proper rotation.

This is only arguing semantics and that replacing this with a rotate call will change the functionality of this code, not arguing what the code should be doing.

opm/simulators/timestepping/TimeStepControl.cpp
erikhide: I'm not sure. I only changed this part of the code to account for the change from `simulationTimeElapsed` to `substepTimer` as input to `computeTimeStepSize()`, so I haven't really looked much into the hardcoded time stepping option. If anyone thinks this should be changed, it might make sense to do so in a separate PR?

opm/simulators/timestepping/TimeStepControl.cpp
erikhide: If you think it's useful to change this, I can do so. Just let me know!

opm/simulators/timestepping/TimeStepControl.cpp
bska: > I'm not sure. I only changed this part of the code to account for [an API change]

I know; you're in the clear here.

My comment was directed more towards those who originally worked in this area, i.e., @totto82 and @atgeirr.  Tracing the history, the feature was introduced in PR #1049 (commit 380fe6e2f, originally OPM/opm-core#1064, commit OPM/opm-core@6e9bb4cf) and has remained essentially unchanged since then.

opm/simulators/timestepping/TimeStepControl.cpp
bska: > If you think it's useful to change this, I can do so. Just let me know!

No need.  As I said, the index range is tiny here so any gain we'd get from using a standard algorithm&ndash;other than possibly having a standard name which might aid in communicating to the reader&ndash;would be lost by function overhead.

bska:
jenkins build this please

opm/simulators/timestepping/TimeStepControl.cpp
```diff
@@ -279,4 +280,93 @@ namespace Opm
...
+            const double newDt = dt * std::pow(safetyFactor_ * tolerance_ / errors_[2], 0.35);
+            if( verbose_ )
+                OpmLog::info(fmt::format("Computed step size (pow): {} days", unit::convert::to( newDt, unit::day )));
+            return newDt;
+        }
+        // Use the general third order controller for all other time steps
+        else
+        {
+            const std::array<double, 3> beta = { 0.125, 0.25, 0.125 };
+            const std::array<double, 2> alpha = { 0.375, 0.125 };
```
bska, line 351: Thanks for making these be `std::array` instead of `std::vector`.  Please `#include <array>` to bring in a declaration of the `std::array` type.

erikhide: Done!

bska:
jenkins build this please

bska (approved):

PR #5978, opened 2025-02-13
Add CompositionalContainer, a container for compositional data output

akva2:
This adds a container for the compositional output. Furthermore it refactors the code a little such that the compositional output is completely contained within the compositional output module.

There is one additional unrelated commit that replaces some typedefs. I can separate that out if you want but it happend to sit in the tree on my laptop so..

### DIFF
--- a/opm/simulators/flow/CompositionalContainer.cpp
+++ b/opm/simulators/flow/CompositionalContainer.cpp
@@ -1,0 +1,29 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+  Consult the COPYING file in the top-level source directory of this
+  module for the precise wording of the license and the list of
+  copyright holders.
+*/
+
+#include <config.h>
+#include <opm/simulators/flow/CompositionalContainer.hpp>
+
+
+namespace Opm {
+
+} // namespace Opm
--- a/opm/simulators/flow/CompositionalContainer.cpp
+++ b/opm/simulators/flow/CompositionalContainer.cpp
@@ -23,8 +23,6 @@
 #include <config.h>
 #include <opm/simulators/flow/CompositionalContainer.hpp>
 
-#include <opm/material/fluidsystems/BlackOilDefaultIndexTraits.hpp>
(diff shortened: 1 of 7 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this opm-common=4481 please

opm/simulators/flow/OutputCompositionalModule.hpp
```diff
@@ -155,11 +158,17 @@ class OutputCompositionalModule : public GenericOutputBlackoilModule<GetPropType
...
-        this->doAllocBuffers(bufferSize,
-                              reportStepNum,
-                              substep,
-                              log,
-                              isRestart);
+        auto rstKeywords = this->schedule_.rst_keywords(reportStepNum);
+        this->compC_.allocate(bufferSize, rstKeywords);
+
+        this->doAllocBuffers(bufferSize, reportStepNum, substep, log, isRestart,
+                             false, false, false, false, 0, {}, 0, rstKeywords);
```
bska, line 165: It's a matter of taste, obviously, but I getting to a point now where sequences of `bool` arguments passed as just `true` or `false` is getting harder to read in isolation.  Did you consider adding tags of the form
```C++
/* vapparsActive = */ false,
/* enablePCHysteresis = */ false, // ...
```
here?

I think Qt uses (the moral equivalent of) `enum class` for these kinds of parameters so you have calls of the form
```
doAllocBuffers(..., Vappars::Inactive, PCHysteresis::Disable, ...)
```
instead.  I don't know if we want to go that route.

On a different note, though, you could probably `move()` the `rstKeywords` into the `doAllocBuffers()` argument to avoid copying.  Especially, since the changes made to `rstKeywords` within `doAllocBuffers()` aren't returned back out to `allocBuffers()`.

opm/simulators/flow/CompositionalContainer.cpp
```diff
@@ -0,0 +1,183 @@
...
+namespace Opm {
+
+template<class FluidSystem>
+void CompositionalContainer<FluidSystem>::
+allocate(const unsigned bufferSize,
+         std::map<std::string, int>& rstKeywords)
+{
+    if (rstKeywords["ZMF"] > 0) {
+        this->allocated_ = true;
+        rstKeywords["ZMF"] = 0;
```
bska, line 44: If you're changing this anyway, we could write it as
```C++
if (auto& zmf = rstKeywords["ZMF"]; zmf > 0) {
    this->allocated = true;
    zmf = 0;
    // ...
}
```
to avoid attempting to insert `"ZMF"` into the map twice.

bska (commented):
I like the simplification here.  Just a couple of smaller points that you might want to have a look at.

opm/simulators/flow/OutputCompositionalModule.hpp
akva2: I should definitely move the map.

I am the first in the line to get a pitch fork when it comes to tons of boolean arguments. I do have bigger plans in the area which would make these arguments go away, but that might take a bit longer. So for now I'll add those comments.

opm/simulators/flow/CompositionalContainer.cpp
akva2: sure thing.

akva2:
jenkins build this opm-common=4481 please

opm/simulators/flow/OutputCompositionalModule.hpp
bska: > I do have bigger plans in the area which would make these arguments go away,

:tada: 

> or now I'll add those comments.

Much appreciated!

bska (approved):

akva2:
note the required upstream

PR #5979, opened 2025-02-13
fix: instantiation of CompositionalContainer

akva2:
This fixes the instantiation for CompositionalContainer. Things were tested, then this merged in the mean time, then the other PRs got merged.

### DIFF
--- a/opm/simulators/flow/CompositionalContainer.cpp
+++ b/opm/simulators/flow/CompositionalContainer.cpp
@@ -23,7 +23,7 @@
 #include <config.h>
 #include <opm/simulators/flow/CompositionalContainer.hpp>
 
-#include <opm/material/fluidsystems/GenericOilGasFluidSystem.hpp>
+#include <opm/material/fluidsystems/GenericOilGasWaterFluidSystem.hpp>
 
 #include <opm/output/data/Solution.hpp>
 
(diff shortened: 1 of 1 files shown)


opm/simulators/flow/CompositionalContainer.cpp
```diff
@@ -168,16 +168,23 @@ outputRestart(data::Solution& sol,
...
+    INSTANTIATE_COMP_THREEPHASE(NUM) \
+    INSTANTIATE_COMP_TWOPHASE(NUM)
+
+INSTANTIATE_COMP_THREEPHASE(0)  // \Note: to register the parameter ForceDisableFluidInPlaceOutput
 INSTANTIATE_COMP(2)
 INSTANTIATE_COMP(3)
 INSTANTIATE_COMP(4)
 INSTANTIATE_COMP(5)
 INSTANTIATE_COMP(6)
-INSTANTIATE_COMP(7)
```
GitPaean, line 181: Is this intentional or accidental? If this is intentional, we need to adjust some statement regarding the number of the components that are supported by us.

akva2: okay, turns out i was just me being sloppy. it's needed and it is in fipcontainer, i somehow missed a line.

akva2:
jenkins build this please

PR #5981, opened 2025-02-13
Extract addLgrs, invoke it before loadBalance

aritorto:
This PR extracts "adding lgrs" from GenericCpGridVanguard::doCreateGrids_() and moves this into the Simulator constructor body, before load balancing, model and problem initialization. 

### DIFF
--- a/opm/models/io/basevanguard.hh
+++ b/opm/models/io/basevanguard.hh
@@ -110,6 +110,17 @@ public:
         updateGridView_();
     }
 
+    /*!
+     * \brief Add LGRs to the grid, if any. Only supported for CpGrid - for now.
+     */
+    void addLgrs()
+    {
+        if(&asImp_() != this) { // this check prevents an infinite-recursion warning
+            asImp_().addLgrs();
+            updateGridView_();
+        }
+    }
+
 protected:
     // this method should be called after the grid has been allocated
     void finalizeInit_()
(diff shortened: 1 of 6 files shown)


aritorto:
jenkins build this please

aritorto:
jenkins build this please

opm/models/io/basevanguard.hh
```diff
@@ -110,6 +110,17 @@ public:
         updateGridView_();
     }
 
+    /*!
+     * \brief Add LGRs to the grid, if any. Only supported for CpGrid - for now.
+     */
+    void addLgrs()
+    {
+        if(&asImp_() != this) {
```
blattms, line 118: Please explain the if statement to me. When is it true? Why do we need it?

opm/models/io/basevanguard.hh
```diff
@@ -110,6 +110,17 @@ public:
...
+            asImp_().addLgrs();
+            updateGridView_();
```
blattms, line 120: Should we move this call into addLgrs?

opm/models/utils/simulator.hh
```diff
@@ -172,6 +172,18 @@ public:
         checkParallelException("Allocating the simulation vanguard failed: ",
                                exceptionThrown, what);
 
+        // Only relevant for CpGrid
+        if (verbose_)
+            std::cout << "Adding LGRs, if any\n" << std::flush;
```
blattms, line 177: Please do not use cout, but logging.
Also it might make sense to move the message into addLgrs

opm/models/io/basevanguard.hh
aritorto: I added a comment

aritorto:
jenkins build this please

PR #5985, opened 2025-02-13
added other intialization in single wellstate if alternative initiali…

hnil:
…zation is usedx


steink:
Something very strange going on with the network-tests that needs sorting out. Will have a closer look.

hnil:
Closing: fixed upstream, and more thoroughly than this PR proposed.

Master now zero-fills surface_rates and explicitly leaves THP/BHP/GRUP producers at zero, deferring to WellInterface::initializeProducerWellState() to derive a sensible initial value from the reservoir state. The hard-coded -1000 sm3/day liquid/water and -100000 sm3/day gas guesses this PR was working around no longer exist, so the AlternativeWellRateInit switch is no longer needed.

PR #5990, opened 2025-02-14
Safer well state

hnil:
Very much draft. Idea is to keep all information in primary variables and use it for:
- better initial guess on solutions
- better fractions for vfp
- safer restarts from failed steps
- maybe more exact restart/load
The most problematic cases in current code is probably std injectors with crossflow and zero rate wells. More changes is need to use this in the code where solution at other conditions is need.


GitPaean:
jenkins build this please

GitPaean:

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -1625,53 +1625,100 @@ namespace Opm
...
+        //             return;
+        //         }
+        //     }
+        // }
         // Calculate the rates that follow from the current primary variables.
-        std::vector<Scalar> well_q_s = computeCurrentWellRates(simulator, deferred_logger);
-
-        if (nonzero_rate_index == -1) {
+        std::vector<Scalar> well_q_s;
+        Scalar bhp_tmp = 0.0;
```
GitPaean, line 1651: With this design, the function `initializeWellState()` can only be used with producers. I think it is used for both types of wells.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -436,10 +436,10 @@ namespace Opm {
             // This is done only for producers, as injectors will only have a single
             // nonzero phase anyway.
             for (const auto& well : well_container_) {
-                const bool zero_target = well->stoppedOrZeroRateTarget(simulator_, this->wellState(), local_deferredLogger);
-                if (well->isProducer() && !zero_target) {
-                    well->updateWellStateRates(simulator_, this->wellState(), local_deferredLogger);
-                }
+                //const bool zero_target = well->stoppedOrZeroRateTarget(simulator_, this->wellState(), local_deferredLogger);
+                //if (well->isProducer()){// && !zero_target) {
+                    well->initializeWellState(simulator_, this->groupState(), this->wellState(), local_deferredLogger);
```
GitPaean, line 441: Here, the function `initializeWellState()` will be used for both producers and injectors.  Needs to be careful.

hnil:
Closing. The idea — keep everything in the primary variables for a better initial guess, VFP fractions and restarts — is still live, but it is being carried by the current well-model work rather than this branch. Branch kept.

PR #5996, opened 2025-02-17
Add RFTContainer, a container for RFT data output

akva2:
More encapsulation work

### DIFF
--- a/opm/simulators/flow/RFTContainer.cpp
+++ b/opm/simulators/flow/RFTContainer.cpp
@@ -100,15 +100,10 @@ addToWells(data::Wells& wellDatas,
             std::transform(well.getConnections().begin(),
                            well.getConnections().end(),
                            std::back_inserter(wellData.connections),
-                           [this](const auto& connection)
+                           [](const auto& connection)
                            {
-                               const std::size_t i = std::size_t(connection.getI());
-                               const std::size_t j = std::size_t(connection.getJ());
-                               const std::size_t k = std::size_t(connection.getK());
-
-                               const std::size_t index = eclState_.gridDims().getGlobalIndex(i, j, k);
                                data::Connection res;
-                               res.index = index;
+                               res.index = connection.global_index();
                                return res;
                            });
             wellDatas.emplace(std::make_pair(well.name(), wellData));
@@ -160,10 +155,7 @@ allocate(const std::size_t reportStepNum)
         }
 
         for (const auto& connection: well.getConnections()) {
-            const std::size_t i = std::size_t(connection.getI());
-            const std::size_t j = std::size_t(connection.getJ());
-            const std::size_t k = std::size_t(connection.getK());
-            const std::size_t index = eclState_.gridDims().getGlobalIndex(i, j, k);
+            const std::size_t index = connection.global_index();
 
             if (FluidSystem::phaseIsActive(oilPhaseIdx)) {
                 oilConnectionPressures_.emplace(std::make_pair(index, 0.0));
--- a/opm/simulators/flow/RFTContainer.cpp
+++ b/opm/simulators/flow/RFTContainer.cpp
@@ -129,7 +129,6 @@ addToWells(data::Wells& wellDatas,
     gasConnectionSaturations_.clear();
 }
 
-
 template<class FluidSystem>
 void RFTContainer<FluidSystem>::
 allocate(const std::size_t reportStepNum)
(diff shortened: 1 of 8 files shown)


akva2:
jenkins build this please

opm/simulators/flow/RFTContainer.cpp
```diff
@@ -0,0 +1,234 @@
...
+            std::transform(well.getConnections().begin(),
+                           well.getConnections().end(),
+                           std::back_inserter(wellData.connections),
+                           [this](const auto& connection)
+                           {
+                               const std::size_t i = std::size_t(connection.getI());
+                               const std::size_t j = std::size_t(connection.getJ());
+                               const std::size_t k = std::size_t(connection.getK());
+
+                               const std::size_t index = eclState_.gridDims().getGlobalIndex(i, j, k);
```
bska, line 109: This is equivalent to
```C++
const std::size_t index = connection.global_index();
```
which is (potentially) less costly.  As follow-up work we could possibly switch to using that method instead?

opm/simulators/flow/RFTContainer.cpp
```diff
@@ -0,0 +1,234 @@
...
+                               data::Connection res;
+                               res.index = index;
+                               return res;
+                           });
+            wellDatas.emplace(std::make_pair(well.name(), wellData));
```
bska, line 114: Might as well write
```C++
wellDatas.emplace(well.name(), std::move(wellData));
```
here, although we could decide to write the overall section as
``C++
auto& wellData = wellDatas[well.name()];
wellData.connections.reserve(...);
std::transform(...);
```
instead.

File: opm/simulators/flow/RFTContainer.cpp [hunk 3]
```diff
@@ -0,0 +1,234 @@
...
+            continue;
+        }
+
+        //add data infrastructure for shut wells
+        if (!wellDatas.count(well.name())) {
+            data::Well wellData;
+
+            if (!rft_config.active()) {
+                continue;
+            }
```
bska, line 97: I suppose we could move that check ahead of the loop, and **possibly** ahead of the `comm.size()` check as well.

bska (commented):
Looks good to me.  We might embrace the opportunity to fix a couple of long-standing issues as well, although those could equally be addressed in follow-up work too.

opm/simulators/flow/RFTContainer.cpp
akva2: did the latter.

opm/simulators/flow/RFTContainer.cpp
akva2: indeed

akva2:
jenkins build this please

opm/simulators/flow/RFTContainer.cpp
bska: > did the latter.

Right, but then you don't need the `wellDatas.emplace()` call since `operator[]()` will already have inserted the object into the map.

opm/simulators/flow/RFTContainer.cpp
```diff
@@ -0,0 +1,225 @@
...
+    for (const auto& well: schedule_.getWells(reportStepNum)) {
+
+        // don't bother with wells not on this process
+        if (!wellQuery_(well.name())) {
+            continue;
+        }
+
+        if (!rft_config.active()) {
+            continue;
+        }
```
bska, line 154: This `!active` check could just go ahead of the loop too.

opm/simulators/flow/RFTContainer.cpp
```diff
@@ -0,0 +1,225 @@
...
+                oilConnectionPressures_.emplace(std::make_pair(index, 0.0));
+            }
+
+            if (FluidSystem::phaseIsActive(waterPhaseIdx)) {
+                waterConnectionSaturations_.emplace(std::make_pair(index, 0.0));
+            }
+
+            if (FluidSystem::phaseIsActive(gasPhaseIdx)) {
+                gasConnectionSaturations_.emplace(std::make_pair(index, 0.0));
+            }
```
bska, line 169: `Emplace(make_pair(i, 0))` is a bit wasteful when you can just do `emplace(i, 0)` instead.

bska (commented):
Thanks a lot for the updates.  This is starting to look good to me.  Just a couple of minor issues remaining.

opm/simulators/flow/RFTContainer.cpp
akva2: done and done. sorry for being sloppy..

opm/simulators/flow/RFTContainer.cpp
akva2: yup.

bska (approved):
Thanks a lot for the updates.  This looks good to me now.  I think I see one more possibility to reduce the cost here, by replacing loops of the form
```C++
for (const auto& well : schedule_.getWells(reportStep)) {
    // ...
}
```
with something along the lines of (untested)
```C++
for (const auto& [wname, wellPtr] : schedule_[reportStep].wells()) {
    // ...
}
```
since that will avoid creating a `vector<Well>` in `Schedule::getWells()`.

However, that's better done in follow-up work so I'll merge this into the master branch now.

PR #6000, opened 2025-02-18
move output of conv data into RSTConv class

akva2:
move member from model to output module. this way we don't have to shuffle data around

PR #6001, opened 2025-02-18
FlowProblem(Blackoil|Comp): Return a reference to EclWriter

akva2:
the pointer is allocated in the ctor, so the fact that it is a unique_ptr
can be kept as an implementation detail. furthermore, the previous
accessor did not properly enforce constness, as a const unique_ptr<EclWriter>
only applies const to the pointer class, not the data the pointer points to.

Sits on top of https://github.com/OPM/opm-simulators/pull/6000


akva2:
jenkins build this please

bska (commented):
Looks good.  I just saw one small issue.

akva2:
jenkins build this please

bska (approved):

PR #6002, opened 2025-02-18
Fix to write residuals

daavid00:
This fixes the writing of residuals (e.g., for a gas water system the current master only writes the residuals for the gas phase).


daavid00:
jenkins build this please

akva2:
woopsie. maybe we can add the residual output to some reg test?

akva2:
you could skip the for loop inside the if with None to avoid some unncessary work (no reason to update map entry multiple times).

akva2:
this fixes a real problem. i'll do the suggested change in follow-up work once we have regression test coverage. thanks for cleaning up my mess.

daavid00:
Ok, today the internet is supper slow so getting opm-tests was taking too long. I was planning to add the RESIDUAL to https://github.com/OPM/opm-tests/blob/master/editnnc/NNC_AND_EDITNNC.DATA, but then I will let you to decide and add the regression test.

PR #5811, opened 2024-12-18
Report Connection Level Fracturing Statistics to I/O Layer

bska:
Populates the new `data::Connection::fract` data member from PR OPM/opm-common#4388.

To this end, introduce two helper classes to compute statistics of a running sample:

  * `RunningStatistics` calculates basic statics measures of a running sample
  * `ConnFracStatistics` aggregates those measures for connection level fracture pressure, fracture width, and flow rate

PR #6004, opened 2025-02-18
Refactor Connection Value Report Function

bska:
In particular, split this function into distinct steps:

  1. Computing the linearised IJK index of each connection.
  2. Reporting multiplicative factors such as the CTF, the D factor, and the transmissibility multiplier due to rock compaction.
  3. Reporting pressures, flow rates, and productivity indices.
  4. Reporting filter cake values for injectors

While here, also reverse a condition in init() to enable "early continue" and reduce levels of nesting and to reduce scope of some objects.

---

Extracted from #5811.


bska:
jenkins build this please

bska:

PR #6006, opened 2025-02-18
double instead of int

daavid00:
Issue found by running spe11a.


daavid00:
jenkins build this please

PR #5317, opened 2024-04-26
Include rejection of time step if tolerance test fails

erikhide:
Most of the work published on adaptive time stepping includes a test for deciding whether or not to reject the most recent time step. This is usually done by taking the relative change or some error estimate and comparing this to one or more threshold value(s). If the change/error is too large, the time step is rejected, and we try again with a shorter time step. In other words, we discard the computations and move back to the previous point in time. Currently, however, this is not done in OPM, and my belief (still only a hypothesis) is that this has contributed to the lack of robustness in the "pure" PID controller implementation.

Note that the OPM Flow Reference Manual includes a footnote reference to "Algebraic Flux Correction III. Incompressible Flow Problems. Uni Dortmund, Turek and Kuzmin, January, 2006 (DOI: 10.1007/3-540-27206-2_8)" in which the method clearly states that a failure to satisfy the tolerance test should result in a rejection of the time step.

I have implemented such a rejection of time steps now. It seems to be working as it should, but please let me know if I should change something.

PR #6009, opened 2025-02-18
Time step rejection if tolerance test fails

erikhide:
I've implemented rejection of a completed time step if the relative change is larger than the time step control tolerance (only for the general 3rd order controller). The code seems to work as it should, but I might have missed something, so please let me know if you find anything wrong.

An earlier version of this can be found in #5317, but the current one is up to date with the current master.

### DIFF
--- a/opm/simulators/flow/NonlinearSolver.hpp
+++ b/opm/simulators/flow/NonlinearSolver.hpp
@@ -35,6 +35,7 @@
 
 #include <opm/simulators/timestepping/SimulatorReport.hpp>
 #include <opm/simulators/timestepping/SimulatorTimerInterface.hpp>
+#include <opm/simulators/timestepping/TimeStepControl.hpp>
 
 #include <memory>
 
@@ -128,7 +129,7 @@ struct NonlinearSolverParameters
         }
 
 
-        SimulatorReportSingle step(const SimulatorTimerInterface& timer)
+        SimulatorReportSingle step(const SimulatorTimerInterface& timer, const TimeStepControlInterface& timeStepControl)
         {
             SimulatorReportSingle report;
             report.global_time = timer.simulationTimeElapsed();
@@ -173,6 +174,13 @@ struct NonlinearSolverParameters
 
                 std::string msg = "Solver convergence failure - Failed to complete a time step within " + std::to_string(maxIter()) + " iterations.";
                 OPM_THROW_NOLOG(TooManyIterations, msg);
+            }
+            if (!timeStepControl.timeStepAccepted(model_->relativeChange())) {
+                failureReport_ = report;
+                report.converged = false;
+
+                std::string msg = "Time step was too large. Tolerance test failed.";
+                OPM_THROW_NOLOG(TimeSteppingBreakdown, msg);
             }
 
             // Do model-specific post-step actions.
--- a/opm/simulators/flow/NonlinearSolver.hpp
+++ b/opm/simulators/flow/NonlinearSolver.hpp
@@ -33,7 +33,6 @@
 #include <opm/models/utils/propertysystem.hh>
 #include <opm/models/utils/basicproperties.hh>
 
-#include <opm/simulators/timestepping/AdaptiveTimeStepping.hpp>
 #include <opm/simulators/timestepping/SimulatorReport.hpp>
 #include <opm/simulators/timestepping/SimulatorTimerInterface.hpp>
 #include <opm/simulators/timestepping/TimeStepControl.hpp>
@@ -101,7 +100,6 @@ struct NonlinearSolverParameters
         using Scalar = GetPropType<TypeTag, Properties::Scalar>;
 
     public:
-        using TimeStepper = AdaptiveTimeStepping<TypeTag>;
         using SolverParameters = NonlinearSolverParameters<Scalar>;
 
         // ---------  Public methods  ---------
@@ -124,7 +122,6 @@ struct NonlinearSolverParameters
             , nonlinearIterationsLast_(0)
             , linearIterationsLast_(0)
             , wellIterationsLast_(0)
-            , enableAdaptiveTimeStepping_(true)
         {
             if (!model_) {
                 OPM_THROW(std::logic_error, "Must provide a non-null model argument for NonlinearSolver.");
@@ -132,7 +129,7 @@ struct NonlinearSolverParameters
         }
 
 
-        SimulatorReportSingle step(const SimulatorTimerInterface& timer, const TimeStepper& adaptiveTimeStepping)
+        SimulatorReportSingle step(const SimulatorTimerInterface& timer, const TimeStepControlInterface *timeStepControl)
         {
             SimulatorReportSingle report;
             report.global_time = timer.simulationTimeElapsed();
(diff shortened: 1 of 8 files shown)


opm/simulators/timestepping/AdaptiveTimeStepping_impl.hpp
```diff
@@ -1139,6 +1147,9 @@ runSubStep_()
     catch (const TooManyIterations& e) {
         handleFailure("Solver convergence failure - Iteration limit reached", e);
     }
+    catch (const TimeSteppingBreakdown& e) {
+        handleFailure("Time step was too large", e);
```
atgeirr, line 1151: This message is misleading, see the other comment. Maybe pass e.what() as the first string, if the message is modified as suggested above (it will be repeated in the DBG file, but show up once on the terminal).

opm/simulators/flow/NonlinearSolver.hpp
```diff
@@ -174,6 +175,13 @@ struct NonlinearSolverParameters
                 std::string msg = "Solver convergence failure - Failed to complete a time step within " + std::to_string(maxIter()) + " iterations.";
                 OPM_THROW_NOLOG(TooManyIterations, msg);
             }
+            if (!timeStepControl.timeStepAccepted(model_->relativeChange())) {
```
atgeirr, line 178: The relativeChange() method still needs to be changed to not add saturations and pressures unscaled... I would really like to get in that, is there a PR for it already?

opm/simulators/flow/NonlinearSolver.hpp
```diff
@@ -174,6 +175,13 @@ struct NonlinearSolverParameters
...
+                report.converged = false;
+                failureReport_ = report;
+
+                std::string msg = "Time step was too large. Tolerance test failed.";
```
atgeirr, line 182: I think this message is misleading, better would be "Relative change in solution for time step was {}, which is larger than the tolerance accepted by the timestepping algorithm." (and then insert the relative change in the output). I.e. it is the change and not the step that is too large.

atgeirr (commented):
I think this is mostly fine, sorry for the long wait for a review.

opm/simulators/flow/NonlinearSolver.hpp
erikhide: Agreed! The PR #6026 aims to fix this. There is one issue I noticed some time after creating this PR: The relative change in saturation (at least for some models) is very high after the initial time step, since the saturation values for some reason jumps quite a lot. This does not seem to happen later in the simulation. One fix is to just use pressure for the first time step, and then both pressure and saturation after that. I've implemented this fix locally. I do not know what to do if we're only using saturation, however. Maybe we should remove that option altogether?

atgeirr: > the saturation values for some reason jumps quite a lot 

Even in average, not just for some cells? Sounds strange, I suggest you investigate that a little to understand why. Does this apply to any public cases?

erikhide: More specifically, what seems to happen is that e.g. the oil saturation and water saturation values for one cell "switch places"  with one going from something like 0.95 to 0.02, and the other to the opposite. This has happened for the Norne prediction cases, although I haven't checked all of them.

opm/simulators/flow/NonlinearSolver.hpp
erikhide: I agree! The proposed text is much better. I've updated the error message now.

atgeirr:
jenkins build this please

atgeirr:
jenkins build this please

opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp
```diff
@@ -468,7 +468,7 @@ class SimulatorFullyImplicitBlackoil : private SerializableSim
             simulator_.problem().setSimulationReport(report_);
         } else {
             // solve for complete report step
-            auto stepReport = solver_->step(timer);
+            auto stepReport = solver_->step(timer, adaptiveTimeStepping_->timeStepControl());
```
atgeirr, line 471: I believe this line is the problem. It assumes there always is a non-null adaptiveTimeStepping_ member, which is false: if you run with `--enable-adaptive-time-stepping=false` it is indeed a null pointer. So a slightly different approach is needed.

bska: >  So a slightly different approach is needed.

We should obviously always check for null, but there's likely to be a deeper problem here too.  If it were "only" `--enable-adaptive-time-stepping=false` that got us into this situation, then the problem would not consistently show up for restarted runs, and mostly in parallel.

akva2: the crash set is exactly those tests that use --enable-adaptive-time-stepping=false.

opm/simulators/flow/NonlinearSolver.hpp
```diff
@@ -122,14 +124,15 @@ struct NonlinearSolverParameters
             , nonlinearIterationsLast_(0)
             , linearIterationsLast_(0)
             , wellIterationsLast_(0)
+            , enableAdaptiveTimeStepping_(true)
```
akva2, line 127: this feels like a layer violation. it would, imo, be better to use the old approach, but pass the TimeStepControlInterface by ptr so you can pass nullptr from the non-adaptive case (and check for null on the inside).

atgeirr: I agree.

atgeirr:
jenkins build this please

opm/simulators/timestepping/TimeStepControl.hpp
```diff
@@ -65,6 +65,8 @@ namespace Opm
                                    const int iterations,
                                    const RelativeChangeInterface& /* relativeChange */,
                                    const AdaptiveSimulatorTimer& /* substepTimer */ ) const override;
+        
+        bool timeStepAccepted(const double error) const override { return true; }
```
bska, line 69: Please tag the `error` parameter as `[[maybe_unused]]`&ndash;or remove the name altogether&ndash;so as to ensure we do not get [additional warnings](https://ci.opm-project.org/job/opm-simulators-PR-builder/8108/gcc/) when this PR gets merged.

opm/simulators/timestepping/TimeStepControl.hpp
```diff
@@ -117,6 +119,8 @@ namespace Opm
                                    const RelativeChangeInterface& relativeChange,
                                    const AdaptiveSimulatorTimer& /* substepTimer */ ) const override;
 
+        bool timeStepAccepted(const double error) const override { return true; }
```
bska, line 122: Please tag the `error` parameter as `[[maybe_unused]]`&ndash;or remove the name altogether&ndash;so as to ensure we do not get [additional warnings](https://ci.opm-project.org/job/opm-simulators-PR-builder/8108/gcc/) when this PR gets merged.

opm/simulators/timestepping/TimeStepControl.hpp
```diff
@@ -166,6 +170,8 @@ namespace Opm
...
```
bska, line 173: Please tag the `error` parameter as `[[maybe_unused]]`&ndash;or remove the name altogether&ndash;so as to ensure we do not get [additional warnings](https://ci.opm-project.org/job/opm-simulators-PR-builder/8108/gcc/) when this PR gets merged.

bska (commented):
I defer full review to those who have already started reviewing the PR, but I noticed that the patch currently introduces some new warnings.  Please try to fix those.

atgeirr:
jenkins build this please

akva2:
this contains a merge commit. please rebase on master.

erikhide:
> please rebase on master.

Done!

akva2:
jenkins build this please

atgeirr:
Thanks, merging!

Please edit the PR description to add a few sentences of description for the manual.

PR #6010, opened 2025-02-18
Allow empty non-group leaf nodes

vkip:


vkip:
jenkins build this please

GitPaean (approved):
The code looks fine.  But what is the usage of this type of node? There is no flow flowing in or out through the node.

vkip:
> The code looks fine. But what is the usage of this type of node? There is no flow flowing in or out through the node.

No obvious use, but encountered in actual models. Since it is allowed at the input stage it would be good not to crash during simulation.

GitPaean:
> > The code looks fine. But what is the usage of this type of node? There is no flow flowing in or out through the node.
> 
> No obvious use, but encountered in actual models. Since it is allowed at the input stage it would be good not to crash during simulation.

Thanks for the response. I am getting the PR in now.

PR #6012, opened 2025-02-19
clean up terminal output

totto82:
Output of stopping / reviving of wells has a tendency of filling the terminal output. Move it to .DBG . If the wells end up stopping/shutting after the timestep. This will still be reported to the terminal/PRT.


totto82:
jenkins build this please

atgeirr:
Agreed, merging.

PR #6013, opened 2025-02-19
Fix Summary BFLOW*

daavid00:
Fix Summary BFLOW*

### DIFF
--- a/opm/simulators/flow/FlowsContainer.cpp
+++ b/opm/simulators/flow/FlowsContainer.cpp
@@ -272,12 +272,16 @@ outputRestart(data::Solution& sol)
 
     using Dir = FaceDir::DirEnum;
     std::vector<DataEntry<Scalar>> entries;
-    addEntry<gasCompIdx>  (entries, "FLRGAS", UnitSystem::measure::rate,                flores_);
-    addEntry<oilCompIdx>  (entries, "FLROIL", UnitSystem::measure::rate,                flores_);
-    addEntry<waterCompIdx>(entries, "FLRWAT", UnitSystem::measure::rate,                flores_);
-    addEntry<gasCompIdx>  (entries, "FLOGAS", UnitSystem::measure::gas_surface_rate,    flows_);
-    addEntry<oilCompIdx>  (entries, "FLOOIL", UnitSystem::measure::liquid_surface_rate, flows_);
-    addEntry<waterCompIdx>(entries, "FLOWAT", UnitSystem::measure::liquid_surface_rate, flows_);
+    if (this->enableFlores_) {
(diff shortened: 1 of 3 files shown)


daavid00:
jenkins build this please

akva2:
doh. another missing regression test then (not an excuse for my sloppyness, more proof it's needed).

daavid00:
Should I then add RESIDUAL and BFLOW to a regression test or you prefer to do it? If so, any particular test where to add these?

daavid00:
Ok, then I will add this today in opm-tests

daavid00:
By adding it to NNC_AND_EDITNNC it seems there are further issues, then in draft mode for now.

daavid00:
This also fixes writing only flores or flows when enable in the restart.

Requieres https://github.com/OPM/opm-tests/pull/1296

opm/simulators/flow/FlowsContainer.cpp
```diff
@@ -271,32 +271,56 @@ outputRestart(data::Solution& sol)
...
-                          std::pair{FaceDir::ToIntersectionIndex(Dir::ZPlus), "K+"},
-                      };
-                      const auto& name = std::get<0>(array);
-                      const auto& measure = std::get<1>(array);
-                      auto& value = std::get<2>(array);
-                      for (const auto& [index, postfix] : dirs) {
-                          doInsert(value[index], name + postfix, measure);
-                      }
-                  });
+    if (this->enableFlores_){
```
akva2, line 274: sorry but this fixes nothing. if the arrays are empty nothing is written and arrays are empty if output is not enabled. if that is not the case, the bug is that arrays have been resized. at the very least you could just if just the addEntry calls.

daavid00: In the test I added then they are written to the restart since BFLOW is in the summary. With this then they are not.

akva2: right. that was stupid of me. we cannot just use array existence since they are dual purpose of course. but then do as i said, just if the addEntry calls, don't duplicate all the code.

daavid00: Sorry for that and thanks for the elegant fix.

daavid00:
jenkins build this opm-tests=1296 update_data please

daavid00:
jenkins build this opm-tests=1296 please

akva2:
you have to merge https://github.com/OPM/opm-tests/pull/1297 into https://github.com/OPM/opm-tests/pull/1296

daavid00:
jenkins build this opm-tests=1296 please

akva2 (approved):
Thanks for cleaning up my mess

daavid00:
> Thanks for cleaning up my mess

Thanks for all your opm code work 🙂.

PR #6014, opened 2025-02-19
Remove pre balance of network

totto82:
I dont think the prebalance step is needed anymore after subiterations are added. This need to be tested more. Testing needs to be done together with https://github.com/OPM/opm-simulators/pull/6011

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -268,7 +268,7 @@ template<class Scalar> class WellContributions;
             void prepareTimeStep(DeferredLogger& deferred_logger);
 
             std::pair<bool, bool>
-            updateWellControls(const bool mandatory_network_balance, DeferredLogger& deferred_logger, const bool relax_network_tolerance = false);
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

GitPaean:

atgeirr:
What about doing a simplified solve (using just implicit IPRs) for the wells in this pre-balance step to have a cheap way to get close to the true solution before the proper balancing loop starts? Also in that case: we should probably run this code more often than we do, currently we require a status change event for some well whereas it could be done for all timesteps to save iterations.

Finally, if we instead continue with deleting this method, we should also remove BlackoilWellModelGeneric::needPreStepNetworkRebalance().

totto82:
>What about doing a simplified solve (using just implicit IPRs) for the wells in this pre-balance step to have a cheap way to get close to the true solution before the proper balancing loop starts?

I did a lot of testing with simplified solves like IPR / STW etc. I still have the code so maybe we can revisit this idea later. But it didn't give much effect. The simplest and best solution seems to be to just remove it. I made it a draft since I would like to know the motivation for this code, but as it is now it adds a huge cost for some models. If you have an example of a model where this has a positive effect we could add it as an option via a parameter. This also make it easier to consider improvements later.

PR #6016, opened 2025-02-19
Clean up: remove code in if(false) branch

lisajulia:
This logic was added in commit https://github.com/OPM/opm-simulators/commit/035d2166414d97c7f45e5e42496c162cc20bccb4 (PR https://github.com/OPM/opm-simulators/pull/4895).
@steink: Can you have a look at this?

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1644,7 +1644,6 @@ namespace Opm
         Scalar relaxation_factor = 1.;
         const Scalar min_relaxation_factor = 0.6;
         bool converged = false;
-        [[maybe_unused]] int stagnate_count = 0;
         bool relax_convergence = false;
         this->regularize_ = false;
         const auto& summary_state = simulator.vanguard().summaryState();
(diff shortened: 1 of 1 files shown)


lisajulia:
jenkins build this please

totto82:
I am fine with removing this if @steink agrees. But I have lately tested an approach that stops iterating (i.e. just give up) if it stagnates multiple times. This potentially gives some speedup but more testing is needed. Is is hard to add a communication step here to fix the parallel issue? You dont need to do it, I just want to know in case I pick this idea up at a later stage. And what about the similar code in iterateWellEqWithControl(...) does that create problems with MPI

steink:
I have to admit I don't recall this was part of https://github.com/OPM/opm-simulators/pull/4895, but I'm fine with deleting it. In general I'm sceptical to accepting wells with very loose convergence criteria because well convergence problems is typically a symptom of something else than just that the well is difficult to solve.

lisajulia:
> I have to admit I don't recall this was part of #4895, but I'm fine with deleting it. In general I'm sceptical to accepting wells with very loose convergence criteria because well convergence problems is typically a symptom of something else than just that the well is difficult to solve.

@steink Thanks :) then this can be merged safely?

totto82:
I just added a draft PR that start using this part of the code https://github.com/OPM/opm-simulators/pull/6020 I think some code that detects stagnation / oscillation is a good idea.

lisajulia:
> I just added a draft PR that start using this part of the code #6020 I think some code that detects stagnation / oscillation is a good idea.

Thanks!

PR #6019, opened 2025-02-19
set minimum pressure in segments

totto82:
For discussion. What is the motivation for allowing <1atm pressure in segments and not use the same minimum as for the top segment?


totto82:

vkip:
https://github.com/OPM/opm-simulators/pull/5234

totto82:
Thanks for the link. The problem is  that for 0 pressures the pvt data typically need to extrapolate which can lead to strange results.

vkip:
> Thanks for the link. The problem is that for 0 pressures the pvt data typically need to extrapolate which can lead to strange results.

True, but I believe the limit in the PR above was added to overcome issues in another field case. Maybe the extrapolation is less of an issues after merging https://github.com/OPM/opm-common/pull/3779 ?

totto82:
Thanks for the discussion. I will close this now. It seems to reduce the number for convergence failures for the wells slightly for some models but ruin others.

PR #6020, opened 2025-02-19
modify stagnation and oscillation code for MSW

totto82:
If the well stagnates 5 times with full damping we give up. (and hope it convergence at the next newton iteration).

### DIFF
--- a/opm/simulators/wells/MultisegmentWellGeneric.cpp
+++ b/opm/simulators/wells/MultisegmentWellGeneric.cpp
@@ -142,18 +142,14 @@ segmentNumberToIndex(const int segment_number) const
 }
 
 template<typename Scalar>
-void
+bool
 MultisegmentWellGeneric<Scalar>::
-detectOscillations(const std::vector<Scalar>& measure_history, bool& oscillate, bool& stagnate) const
+update_relaxation_factor(const std::vector<Scalar>& measure_history, Scalar& relaxation_factor, bool& regularize, DeferredLogger& deferred_logger) const
 {
     const auto it = measure_history.size() - 1;
     if ( it < 2 ) {
-        oscillate = false;
-        stagnate = false;
-        return;
+        return false;
     }
-
-    stagnate = true;
     const Scalar F0 = measure_history[it];
     const Scalar F1 = measure_history[it - 1];
     const Scalar F2 = measure_history[it - 2];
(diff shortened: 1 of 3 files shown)


totto82:

totto82:
For discussion. Does damping give a positive effect for stagnating solution or only for oscillating solution? I didn't check all the details here, so the change here is only based on the observation that no stagnating wells seems to recover after repeated stagnations so no need to waste iterations.

steink:
> For discussion. Does damping give a positive effect for stagnating solution or only for oscillating solution? I didn't check all the details here, so the change here is only based on the observation that no stagnating wells seems to recover after repeated stagnations so no need to waste iterations.

In my experience, stagnation only happens when a well is forced to use a non-feasible control (there is no valid solution). E.g. for a producer, too high bhp/thp resulting in chopping rates at zero, or too high rate resulting in chopping bhp/segment pressures at minimal value. For these cases, there is of course no remedy (except allowing a control switch).  

For oscillations, damping should help. For the version with switching, especially for bad initial guesses and bad explicit lookup rates, one may end up in stop/open oscillations, but I dont think these are captured in the current oscillation detection.

totto82:
> In my experience, stagnation only happens when a well is forced to use a non-feasible control (there is no valid solution). E.g. for a producer, too high bhp/thp resulting in chopping rates at zero, or too high rate resulting in chopping bhp/segment pressures at minimal value. For these cases, there is of course no remedy (except allowing a control switch).
> 
> For oscillations, damping should help. For the version with switching, especially for bad initial guesses and bad explicit lookup rates, one may end up in stop/open oscillations, but I dont think these are captured in the current oscillation detection.

This agrees with by experience too. I will update with the PR with code that distinguishes between oscillation and stagnation.  For oscillations damping is applied. For stagnation we give up if it stagnate repeated times.

totto82:

totto82:
I tried to write something aligned with what @steink suggested.

totto82:

totto82:
I have dome preliminary testing that looks promising. I will undraft and ask for review

steink:
This looks fine to me, but I don't really understand why the PR uses more time-steps than master for a couple of the tests. The PR should only influence the logic around stagnation and not the relaxation procedure, right?

totto82:

totto82:
>This looks fine to me, but I don't really understand why the PR uses more time-steps than master for a couple of the tests. The PR should only influence the logic around stagnation and not the relaxation procedure, right?

Almost. It removes the usage of relaxed tolerances.  This is based on earlier comments from you that this may cause more problems than it solves. I can put it back if you want.

steink:
> Almost. It removes the usage of relaxed tolerances. This is based on earlier comments from you that this may cause more problems than it solves. I can put it back if you want.

I see. So that has an effect on some of the tests, but it's not clear to me whether it is an improvement or not. Not sure, maybe it's best to just put the relaxed tollerances back for now, and disregard my previous statement. In that case, I expect tests should be all fine.

totto82:

totto82:

totto82:
Adding back the relaxed tolerance fall-back didn't change much on the test failures. i.e. with this PR we still use more time-steps after some well event for 01_9_4b_grpctl_msw_model2. We have had problems with instabilities for this test before. I will do some more testing. @steink Are the other test failures ok?

totto82:
Is there anything more I need to test / fix before we can start the process of getting this in?

steink:
Looks all good to me

totto82:

totto82:
jenkins build this opm-tests=1302 please

PR #6024, opened 2025-02-20
fix tracy support

akva2:
link to targets to get include paths set up.

Discovered these now as I do not have tracy in system paths.


akva2:
jenkins build this please

bska (approved):

PR #6025, opened 2025-02-20
move template to avoid clang warning

akva2:
Clangd complains the template is called before it has been defined and that it is a c++-20 extension.
Kinda weird, but don't want to fight it...


akva2:
jenkins build this please

bska (approved):

PR #5937, opened 2025-01-31
Add missing communication for parallel runs with MSWells

lisajulia:
This PR adds communication to three spots where it was missing before (commits adb7223eeb97fa475e50b0721e1e460350d4216b, d454d03d38f809295ab22a7b2b8faf9f1c465d50 and  f0fbe62056c0dfeab59b8c978d00778026b3a522) and renames some variables to make it obvious if they contain local or global quantities.
With these changes, the test added in commit 3b68c4eb114ed63225065542ce4f96e5e5ec8174 passes. The test does not pass without these changes, see #5999, there the parallel simulation runs into a deadlock due to the missing communication.

This PR is not relevant for the reference manual.

PR #6026, opened 2025-02-21
Fix for relative change (used in time step controllers)

erikhide:
This is a new and improved version of #4856. It adds the choice between using just pressure or saturation, or both, when calculating the relative change used in the time step controllers.

The current master adds pressure and saturation together in such a way that the saturation effectively doesn't do anything. Here, I've chosen to use only pressure as default, and I think this will give results more or less identical to what the current master gives.

Question: Might it make more sense to return the square root of the relative change (to get something resembling an L2 norm)?

### DIFF
--- a/opm/simulators/flow/BlackoilModel.hpp
+++ b/opm/simulators/flow/BlackoilModel.hpp
@@ -28,7 +28,6 @@
 
 #include <opm/simulators/flow/BlackoilModelConvergenceMonitor.hpp>
 #include <opm/simulators/flow/BlackoilModelNldd.hpp>
-#include <opm/simulators/flow/BlackoilModelParameters.hpp>
 #include <opm/simulators/flow/BlackoilModelProperties.hpp>
 #include <opm/simulators/flow/FlowProblemBlackoilProperties.hpp>
 #include <opm/simulators/flow/RSTConv.hpp>
--- a/opm/simulators/flow/BlackoilModel_impl.hpp
+++ b/opm/simulators/flow/BlackoilModel_impl.hpp
@@ -34,6 +34,7 @@
 #include <opm/common/ErrorMacros.hpp>
 #include <opm/common/OpmLog/OpmLog.hpp>
 
+#include <opm/simulators/flow/BlackoilModelParameters.hpp>
 #include <opm/simulators/flow/countGlobalCells.hpp>
 
 #include <algorithm>
@@ -373,14 +374,11 @@ relativeChange() const
     std::array<Scalar,4> r{};
     auto& [resultDeltaPressure, resultDenomPressure, resultDeltaSaturation, resultDenomSaturation] = r;
 
-    const auto version = this->param_.relative_change_version_;
-
     const auto& elemMapper = simulator_.model().elementMapper();
     const auto& gridView = simulator_.gridView();
     for (const auto& elem : elements(gridView, Dune::Partitions::interior)) {
         unsigned globalElemIdx = elemMapper.index(elem);
         const auto& priVarsNew = simulator_.model().solution(/*timeIdx=*/0)[globalElemIdx];
-        const double elemVolume = elem.geometry().volume();
 
         Scalar pressureNew;
         pressureNew = priVarsNew[Indices::pressureSwitchIdx];
@@ -409,6 +407,8 @@ relativeChange() const
         }
 
         const auto& priVarsOld = simulator_.model().solution(/*timeIdx=*/1)[globalElemIdx];
+
+        const double elemVolume = elem.geometry().volume();
 
         Scalar pressureOld;
         pressureOld = priVarsOld[Indices::pressureSwitchIdx];
@@ -450,7 +450,9 @@ relativeChange() const
 
     gridView.comm().sum(r.data(), r.size());
 
-    switch(version) {
+    const auto relative_change_version = this->param_.relative_change_version_;
+
+    switch(relative_change_version) {
         case RelativeChangeApproaches::Pressure:
             if (resultDenomPressure > 0.0)
                 return std::sqrt(resultDeltaPressure / resultDenomPressure);


bska:

akva2:
note: CI is in broken state.

bska:
> CI is in broken state.

Okay.  Is there a way to unbreak it within a reasonable timeframe?

akva2:
merge the missing opm-tests PR. i pointed it out but no response yet, https://github.com/OPM/opm-simulators/pull/5937

bska:

erikhide:
Concerning this PR: It might make more sense to take the square root before returning the relative change. Also, would it be better to weight the contributions from each cell by their volume (this is easy to implement)? Performing these changes will, however, change the default much more than the current PR does. What do you think, @atgeirr?

atgeirr:
> What do you think, @atgeirr?

I think you should make the changes you think are correct.

If that changes behaviour too much for existing cases (for example resulting in much shorter average timesteps than the current code does) we may try to address that by also changing the default parameters used for the method, i.e. tuning to get similar timesteps as currently. If that fails, we may need to keep the existing calculations (or something similar) as the default, and require a command line argument to override, but I really hope we do not need to.

erikhide:
> I think you should make the changes you think are correct.

Very well! I'll make the necessary changes soon. When taking the square root, the `time-step-control-tolerance` parameter will most likely have to be adjusted to avoid getting shorter time steps.

erikhide:
I've now added square root and weighting by volume.

Further: When using "pressure+saturation", only pressure is considered for the first time step, due to issues with very large change in saturation after the initial time step. I've not done any such adjustments when only "saturation" is being used.

opm/simulators/flow/BlackoilModel_impl.hpp
```diff
@@ -444,18 +446,37 @@ relativeChange() const
...
             }
         }
     }
 
-    resultDelta = gridView.comm().sum(resultDelta);
-    resultDenom = gridView.comm().sum(resultDenom);
+    gridView.comm().sum(r.data(), r.size());
 
-    return resultDenom > 0.0 ? resultDelta / resultDenom : 0.0;
+    if (version == "pressure") {
```
blattms, line 459: Could we use an internal representation other than string that always for a switch-case here instead of many string comparisons?

blattms:
jenkins build this please

opm/simulators/flow/BlackoilModel_impl.hpp
erikhide: I've made an attempt to do that now. Let me know if this looks good or if I should change something.

atgeirr: Use an `enum class` instead of `int`, so instead of `case 1:` you get something easier to read and understand.

erikhide: Done!

opm/simulators/flow/BlackoilModel_impl.hpp
```diff
@@ -444,18 +446,36 @@ relativeChange() const
...
+            if (resultDenomPressure > 0.0)
+                return std::sqrt(resultDeltaPressure / resultDenomPressure);
+            break;
+        case RelativeChangeApproach::Saturation:
+            if (resultDenomSaturation > 0.0)
+                return std::sqrt(resultDeltaSaturation / resultDenomSaturation);
+            break;
+        case RelativeChangeApproach::PressureSaturation:
+            if (resultDenomPressure > 0.0 && resultDenomSaturation > 0.0) {
+                if (simulator_.timeStepIndex() == 0) {
```
atgeirr, line 470: Add comment: "Only using pressure for time step zero is a pragmatic choice to avoid excessive cutting for the first step".

atgeirr (commented):
The code is fine, with just a few nitpicks.

We must investigate the tests though.

GitPaean:

GitPaean:
After discussion, I think @erikhide can try to fix failures in the previous comment first. Other regression failures are essentially due to different time stepping, although it triggers bheavoir changes. Still thinking how to address those. 

But fixing the previously mentioned failures can be a good start.

GitPaean:
The following tests the parallel running did not run through

[mpi.compareParallelSim_flow+SPE1CASE2](https://ci.opm-project.org/job/opm-simulators-PR-builder/8268/testReport/junit/(root)/mpi/compareParallelSim_flow_SPE1CASE2/)
[mpi.compareParallelSim_flow+SPE1CASE2_THERMAL](https://ci.opm-project.org/job/opm-simulators-PR-builder/8268/testReport/junit/(root)/mpi/compareParallelSim_flow_SPE1CASE2_THERMAL/)

Is it the case for your local running?
@erikhide

GitPaean:

erikhide:
> Is it the case for your local running?

Yes, it turns out that they crash for `time-step-control-tolerance=0.12`, but not for `0.11`, which I think is a bit strange; I would expect it to run through when being a bit slacker on the tolerances. I've changed to `0.2` now, which works for both models when I run them locally.

GitPaean:
With the latest branch, 
https://ci.opm-project.org/job/opm-simulators-PR-builder/8270/
 it looks like the only parallel failure is the following one, 

https://ci.opm-project.org/job/opm-simulators-PR-builder/8270/testReport/junit/(root)/mpi/compareParallelSim_flow_3_A_MPI_MULTFLT_SCHED_MODEL2/

Most likely we can relax the comparison tolerance if the well curve comparison looks good. 

```
Value index                 = 67
(first value, second value) = (9.12757e+08, 9.14413e+08)

Program threw an exception: [/var/lib/jenkins/workspace/opm-simulators-PR-builder/deps/opm-common/test_util/EclRegressionTest.cpp:227] Deviations exceed tolerances.
The absolute deviation is 1655232, and the tolerance limit is 0.02.
The relative deviation is 0.0018101587680645803, and the tolerance limit is 0.001.
```


I can check other regressions later today or tomorrow.  those are regression failures can be potentially fixed through updating the reference if agreed with.

GitPaean:

opm/simulators/flow/BlackoilModel.hpp
```diff
@@ -28,6 +28,7 @@
 
 #include <opm/simulators/flow/BlackoilModelConvergenceMonitor.hpp>
 #include <opm/simulators/flow/BlackoilModelNldd.hpp>
+#include <opm/simulators/flow/BlackoilModelParameters.hpp>
```
GitPaean: This can be moved to the _impl file, where it is used.

opm/simulators/flow/BlackoilModel_impl.hpp
```diff
@@ -376,14 +376,17 @@ typename BlackoilModel<TypeTag>::Scalar
 BlackoilModel<TypeTag>::
 relativeChange() const
 {
-    Scalar resultDelta = 0.0;
-    Scalar resultDenom = 0.0;
+    std::array<Scalar,4> r{};
+    auto& [resultDeltaPressure, resultDenomPressure, resultDeltaSaturation, resultDenomSaturation] = r;
+
+    const auto version = this->param_.relative_change_version_;
```
GitPaean: `version` can be some more informative variable name, can be `relative_change_approach` or something like that. 

And `version` and `elemVolume` can be moved to be closer to where they are actually used.

GitPaean:
I did not go through all the cases, but for the beginning parts I have been looking at, quite a few cases did not run through with the current version of the PR. 

The following is the ones I looked at just now, (based on the order in the [Build Artifacts](https://ci.opm-project.org/job/opm-simulators-PR-builder/8275/artifact/))

```
01_waghyst1 not finish
02_spe12 not finish
03_spe12_polyhedralgrid
04_spe1case2_thermal_watvisc
05_spe1case2_thermal
08_spe1_float
```
potentially more, while I did not go through.

opm/simulators/flow/BlackoilModel.hpp
erikhide: Done!

opm/simulators/flow/BlackoilModel_impl.hpp
erikhide: I'm sorry this took so long. I've made the changes you requested now.

PR #6028, opened 2025-02-21
Avoid warnings in cublas_wrapper

kjetilly:
Return value at end of function since nvcc seems to complain about missing return value even if we throw.


kjetilly:
Jenkins build this please

multitalentloes:
Looks good on HIP too

PR #6035, opened 2025-02-25
add some profiling blocks to tracer

akva2:


bska (approved):

PR #6037, opened 2025-02-26
Tracer: Multithreading

akva2:
Opening early to have somewhere to put markdown tables. Not ready for review/merge/whatever.


akva2:
don't look at the noise....

it's likely because things are not compute dominated but bandwidth dominated and whatever else i'm doing on my pc affects that even though the computations have their own cores. in particular update is basically shuffling data around, no computations going on there.

bska:
The alternative implementation in #6206 was merged into the master branch moments ago. Is there more work coming to leverage that PR?

akva2:
I have some other stuff but it needs rebasing anyways so I'll reopen with a more on-context PR if I ever get to finish it.

PR #6040, opened 2025-02-26
Work in progress: Primary variables on GPU

kjetilly:
This is a very incomplete attempt at the moment. To get the DUNE vectors with GPU decorators, those classes have in essence been copied over to the repo and then been modified to fit on a GPU. They are in a different namespace than Dune as well.


atgeirr:
Be aware that due to Dune having a different license than OPM (GPL 2 + exception vs GPL 3) it is not at all straightforward to move code from Dune to OPM. It is best avoided. A better solution is to make the necessary additions in a dune fork instead. That also makes it much easier to get them into upstream Dune, even if that might not be something that is likely to happen it is a much better situation.

PR #6041, opened 2025-02-27
Fix tracercontainer in flowexp_comp

akva2:
I cannot explain this, but this fixes the problem. 

For some reason, we get calls to tracerC_ where the eclState_ reference member is nullptr. By not holding this as a member this is avoided. I cannot figure out how tracerC_ gets in such a state.


akva2:
jenkins build this please

GitPaean:
Thanks for the fixing. With some efforts, we probably can find the problematic change to understand the issue, but with the limited resource, I am happy it gets fixed.

bska (approved):
Looks good and I'll merge into master.

The **usual** reason for reference data members to become invalid is if the containing type gets copied or moved, but I haven't traced all uses here.

akva2:
yeah i know, i disabled both copy and move ctor / assignment but non of them triggered.

PR #6045, opened 2025-03-04
TracerModel: avoid setting up ILU if no solves will be performed

akva2:
Mostly irrelevant, but why not?

# Timings for NORNE_ATW2013 [5 runs]
|    Revision    |tracerAdvance|tracerSolve |Pre/post step|p-value(tracerSolve)|
|----------------|-------------|------------|-------------|-------------------:|
|tracer_avoid_ilu|32.55 +- 0.13|5.16 +- 0.00|55.03 +- 0.13|                   0|
|master          |32.17 +- 0.23|5.34 +- 0.00|54.89 +- 0.51|                    |

# Timings for NORNE_ATW2013_1A_MSW [5 runs]
|    Revision    |tracerAdvance|tracerSolve |Pre/post step|p-value(tracerSolve)|
|----------------|-------------|------------|-------------|-------------------:|
|tracer_avoid_ilu|32.76 +- 0.48|0.13 +- 0.00|59.05 +- 0.81|                   0|
|master          |35.99 +- 0.42|3.10 +- 0.00|61.63 +- 0.66|                    |

### DIFF
--- a/opm/simulators/flow/GenericTracerModel_impl.hpp
+++ b/opm/simulators/flow/GenericTracerModel_impl.hpp
@@ -441,10 +441,9 @@ bool GenericTracerModel<Grid,GridView,DofMapper,Stencil,FluidSystem,Scalar>::
 linearSolveBatchwise_(const TracerMatrix& M, std::vector<TracerVector>& x, std::vector<TracerVector>& b)
 {
     OPM_TIMEBLOCK(tracerSolve);
-    Scalar tolerance = 1e-2;
-    int maxIter = 100;
-
-    int verbosity = 0;
+    const Scalar tolerance = 1e-2;
+    const int maxIter = 100;
+    const int verbosity = 0;
 
 #if HAVE_MPI
     if (gridView_.grid().comm().size() > 1)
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

opm/simulators/flow/GenericTracerModel_impl.hpp
```diff
@@ -488,7 +493,7 @@ linearSolveBatchwise_(const TracerMatrix& M, std::vector<TracerVector>& x, std::
             x[nrhs] = 0.0;
             Dune::InverseOperatorResult result;
             solver.apply(x[nrhs], b[nrhs], result);
-            converged = (converged && result.converged);
+            converged &= result.converged;
```
bska, line 496: This will fail if `result.converged` is ever changed to something other than `bool`, so I'd prefer if we kept the original statement even if it's more to type.

bska (commented):
Please don't use bitwise operations for `bool`s.  Other than that, this is fine.

opm/simulators/flow/GenericTracerModel_impl.hpp
akva2: i find it less expressive but sure.

akva2:
jenkins build this please

opm/simulators/flow/GenericTracerModel_impl.hpp
bska: > i find it less expressive but sure.

Maybe, but it's definitely more correct.

bska (approved):

PR #6046, opened 2025-03-04
make nonlinear system possible to change

hnil:
Currently nonlinear system is named BlackoilModel form BlackoilModel.hpp. This make it possible to use other nonlinear systems and or modify the behavoir of BlackoilModel in separate executables. Uses maybe:
- adding of mechanics and or fractures
- other method to solve system including the linear/nonlinear system.

### DIFF
--- a/opm/simulators/flow/FlowBaseProblemProperties.hpp
+++ b/opm/simulators/flow/FlowBaseProblemProperties.hpp
@@ -57,6 +57,11 @@ struct FlowBaseProblem {
 
 }
 
+template<class TypeTag, class MyTypeTag>
+struct NonlinearSystem { using type = UndefinedProperty; };
+
+
+
 // The class which deals with ECL aquifers
 template<class TypeTag, class MyTypeTag>
 struct AquiferModel { using type = UndefinedProperty; };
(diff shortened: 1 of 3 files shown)


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

opm/simulators/flow/BlackoilModelProperties.hpp
```diff
@@ -25,7 +25,6 @@
 #define OPM_BLACKOILMODEL_PROPERTIES_HEADER_INCLUDED
 
 #include <opm/simulators/flow/FlowProblemBlackoilProperties.hpp>
-
```
GitPaean, line 28: If you update, you can avoid deleting this line. It is not related to this PR.

GitPaean:
It looks generally fine to me. But I am not sure about the naming, 

The Property name is `NonLinearSystem`,  and we assign `BlackoilModel` to it, and use it as `Model`. 

Should we name `NonLinearSystem` to be `PhysicalModel` or something related?

@bska @akva2 @atgeirr

GitPaean:
It does not look like an interesting discussion.  I am getting the PR in now since it is easy to adjust if needed later.

PR #6048, opened 2025-03-05
Make pre-solve network optional

totto82:
This replaces https://github.com/OPM/opm-simulators/pull/6014 make make the pre-solve optional.


totto82:
jenkins build this please

atgeirr:
The new option is made hidden. I am not sure if I agree with that, but it is a minor thing that is easy to change later, and we expect to see changes in this area anyway so I will merge as is.

totto82:
>The new option is made hidden. I am not sure if I agree with that, but it is a minor thing that is easy to change later, and we expect to see changes in this area anyway so I will merge as is.

It should only be hidden for flow_exp

PR #6050, opened 2025-03-06
Support opening of following-on well

totto82:
Add support for item 9 in WECON

Depends on https://github.com/OPM/opm-common/pull/4507

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -528,7 +528,7 @@ template<class Scalar> class WellContributions;
             void extractLegacyDepth_();
 
             /// upate the wellTestState related to economic limits
-            void updateWellTestState(const double& simulationTime, WellTestState& wellTestState) const;
+            void updateWellTestState(const double& simulationTime, WellTestState& wellTestState);
 
             void wellTesting(const int timeStepIdx, const double simulationTime, DeferredLogger& deferred_logger);
 
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -528,7 +528,7 @@ template<class Scalar> class WellContributions;
             void extractLegacyDepth_();
 
             /// upate the wellTestState related to economic limits
-            void updateWellTestState(const double& simulationTime, WellTestState& wellTestState);
+            void updateWellTestState(const double simulationTime, WellTestState& wellTestState);
 
             void wellTesting(const int timeStepIdx, const double simulationTime, DeferredLogger& deferred_logger);
 
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1970,7 +1970,7 @@ namespace Opm {
     template<typename TypeTag>
     void
     BlackoilWellModel<TypeTag>::
-    updateWellTestState(const double& simulationTime, WellTestState& wellTestState)
+    updateWellTestState(const double simulationTime, WellTestState& wellTestState)
     {
         OPM_TIMEFUNCTION();
         DeferredLogger local_deferredLogger;
@@ -1997,11 +1997,22 @@ namespace Opm {
                 // maybe open a new well
                 const WellEconProductionLimits& econ_production_limits = well->wellEcl().getEconLimits();
                 if (econ_production_limits.validFollowonWell()) {
-                    auto& ws = this->wellState().well(econ_production_limits.followonWell());
-                    bool success = ws.updateStatus(WellStatus::OPEN);
+                    const auto episode_idx = simulator_.episodeIndex();
+                    const auto follow_on_well = econ_production_limits.followonWell();
+                    if (!this->schedule().hasWell(follow_on_well, episode_idx)) {
+                        const auto msg = fmt::format("Well {} was closed. But the given follow on well {} does not exist." 
+                                                     "The simulator continues without opening a follow on well.",
+                                                     wname, follow_on_well);
+                        local_deferredLogger.warning(msg);
+                    }
+                    auto& ws = this->wellState().well(follow_on_well);
+                    const bool success = ws.updateStatus(WellStatus::OPEN);
                     if (success) {
-                        const std::string msg = std::string("well ") + econ_production_limits.followonWell() + std::string(" opens instead");
+                        const auto msg = fmt::format("Well {} was closed. The follow on well {} opens instead.", wname, follow_on_well);
                         local_deferredLogger.info(msg);
+                    } else {
+                        const auto msg = fmt::format("Well {} was closed. The follow on well {} is already open.", wname, follow_on_well);
+                        local_deferredLogger.warning(msg);
                     }
                 }
 
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -37,6 +37,7 @@
 #include <opm/input/eclipse/Schedule/Well/PAvgDynamicSourceData.hpp>
 #include <opm/input/eclipse/Schedule/Well/WellMatcher.hpp>
 #include <opm/input/eclipse/Schedule/Well/WellTestConfig.hpp>
+#include <opm/input/eclipse/Schedule/Well/WellEconProductionLimits.hpp>
 
 #include <opm/input/eclipse/Units/UnitSystem.hpp>
 
@@ -815,7 +816,8 @@ namespace Opm {
                               }
 
                               constexpr auto events_mask = ScheduleEvents::WELL_STATUS_CHANGE |
-                                                           ScheduleEvents::REQUEST_OPEN_WELL;
+                                                           ScheduleEvents::REQUEST_OPEN_WELL |
+                                                           ScheduleEvents::REQUEST_SHUT_WELL;
                               const bool well_status_change =
                                   this->report_step_starts_ &&
                                   wg_events.hasEvent(well_ecl.name(), events_mask);
@@ -850,18 +852,15 @@ namespace Opm {
                 const auto well_status = this->schedule()
                     .getWell(well_name, report_step).getStatus();
 
-                if ((well_ecl.getStatus() == Well::Status::SHUT) ||
-                    (well_status          == Well::Status::SHUT))
-                {
-                    // Due to ACTIONX the well might have been closed behind our back.
-                    if (well_ecl.getStatus() != Well::Status::SHUT) {
-                        this->closed_this_step_.insert(well_name);
-                        this->wellState().shutWell(w);
-                    }
+                const bool shut_event = this->wellState().well(w).events.hasEvent(ScheduleEvents::REQUEST_SHUT_WELL);
+                const bool open_event = this->wellState().well(w).events.hasEvent(ScheduleEvents::REQUEST_OPEN_WELL);
+                const auto& ws = this->wellState().well(well_name);
 
-                    this->well_open_times_.erase(well_name);
-                    this->well_close_times_.erase(well_name);
-                    continue;
+                if (shut_event && ws.status != Well::Status::SHUT) {
+                    this->closed_this_step_.insert(well_name);
+                    this->wellState().shutWell(w);
+                } else if (open_event && ws.status != Well::Status::OPEN) {
+                    this->wellState().openWell(w);
                 }
 
                 // A new WCON keywords can re-open a well that was closed/shut due to Physical limit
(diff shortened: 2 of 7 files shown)


totto82:
In the current code we sometimes check the wellstate and sometimes the well objects in schedule for the well status. With this (and maybe already with the introduction of WCYCLE) they are not necessary the same. If the approach suggested here is acceptable, we need to replace status checks based on the well objects in the schedule with the ones from the well state. Before doing that I would like some feedback. (Especially from @bska).

totto82:
I think we should merge #6051 first. I will rebase that one.

opm/simulators/wells/WellTest.cpp
```diff
@@ -377,6 +373,13 @@ updateWellTestStateEconomic(const SingleWellState<Scalar>& ws,
                 deferred_logger.info(msg);
             }
         }
+        if (econ_production_limits.validFollowonWell()) {
```
GitPaean: This code is within the if condition ` if (rate_limit_violated) `, which should not be correct.  For other causes even physical causes, we can open the follow on well.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -425,6 +425,10 @@ namespace Opm
 
         WellTestState welltest_state_temp;
 
+        for (const auto& well: well_test_state.get_follow_on_wells()) {
```
GitPaean: similar to the comment in the upstreaming PR, we lost the mapping relationship between the two wells if we do this.

opm/simulators/wells/WellTest.cpp
GitPaean: If we want to also handle the physical cause as intended, it should go to higher level in the function `updateWellTestState` or handle it both in the function `updateWellTestStateEconomic` and `updateWellTestStatePhysical` separately.  The message can be more specific like `Well A will be opened following the shutting of the well B`.

opm/simulators/wells/WellInterface_impl.hpp
GitPaean: I am changing the comments above to be:

What is the purpose to add the follow-on well to `welltest_state_temp`?

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -870,9 +870,13 @@ namespace Opm {
...
 
-                    this->well_open_times_.erase(well_name);
-                    this->well_close_times_.erase(well_name);
-                    continue;
+                    if (!this->wellTestState().is_follow_on_well(well_name)) {
+                        this->well_open_times_.erase(well_name);
+                        this->well_close_times_.erase(well_name);
+                        continue;
+                    } else {
+                        this->wellState().openWell(w);
```
GitPaean: If we follow the current implementation, do we need to remove the well `well_name` from the follow_on_well list after we open it?

opm/simulators/wells/WellTest.cpp
totto82: > physical causes, we can open the follow on well

Are you sure that we also want to open it due to physical reason? It is an item in WECON?

opm/simulators/wells/BlackoilWellModel_impl.hpp
totto82: No, since the well_status given by the schedule will continue to be shut, we need to keep it in the list to keep it open.

GitPaean: Then it can be worrying. In my understanding, we probably only want to do the action of opening the follow-on well ONCE  when the specific well is closed automatically. 

Then this well can operate on its own (also free to close due to various conditions) no matter what happens to the specific well. 

We probably do not want keep re-opening this well in my view.

totto82: It will still not open if it is closed for other reasons. See the code below, which again checks the wellTestState.

GitPaean: But that is less direct and complicates the logic and code. I  think it is wise to make this opening only one-time action, then it will not intervene with potential later SCHEDULE setup.

totto82: So, your question is, what happens if there is a SHUT in the SCHEDULE setup later? I think the current code will still keep it open. That is not the intention, so I will need to fix that. Is your suggestion to modify the SCHEDULE object instead? I thought the schedule was supposed to be `const `.

GitPaean: My suggestion is that when a well is opened as a follow-on well (the follow-on action from WECON is finished), we should remove it from `follow_on_wells` in the well test state. The well is free to run following the whatever comes in the later SCHEDULE setup and simulator. 

It is possible the WECON well (item 1) can be closed/shut again in a later stage, I have not seen text about whether we should should re-open it again. I think we can do it following the situation at that time. 

By only handling follow-on opening action as a one-time action, it is a simpler and easier logic.

opm/simulators/wells/BlackoilWellModel.hpp
```diff
@@ -528,7 +528,7 @@ template<class Scalar> class WellContributions;
             void extractLegacyDepth_();
 
             /// upate the wellTestState related to economic limits
-            void updateWellTestState(const double& simulationTime, WellTestState& wellTestState) const;
+            void updateWellTestState(const double& simulationTime, WellTestState& wellTestState);
```
GitPaean: since you are here, we can remove the `&` in `const double&`.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -815,7 +816,8 @@ namespace Opm {
                               }
 
                               constexpr auto events_mask = ScheduleEvents::WELL_STATUS_CHANGE |
-                                                           ScheduleEvents::REQUEST_OPEN_WELL;
+                                                           ScheduleEvents::REQUEST_OPEN_WELL |
+                                                           ScheduleEvents::REQUEST_SHUT_WELL;
```
GitPaean, line 820: Need to check whether the following statement is still sensible related to `well_status_change`. 

```c++
const bool well_status_change =
                                  this->report_step_starts_ &&
                                  wg_events.hasEvent(well_ecl.name(), events_mask);
```

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1993,6 +1993,18 @@ namespace Opm {
 
             if (!wasClosed && wellTestState.well_is_closed(wname)) {
                 this->closed_this_step_.insert(wname);
+
+                // maybe open a new well
+                const WellEconProductionLimits& econ_production_limits = well->wellEcl().getEconLimits();
+                if (econ_production_limits.validFollowonWell()) {
+                    auto& ws = this->wellState().well(econ_production_limits.followonWell());
```
GitPaean: We should check whether the well does exist for safety.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1993,6 +1993,18 @@ namespace Opm {
...
+                    bool success = ws.updateStatus(WellStatus::OPEN);
```
GitPaean: `bool` should be `const`. And also, variable name `success` can be something like `updated` or `opened`.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1993,6 +1993,18 @@ namespace Opm {
...
+                    if (success) {
+                        const std::string msg = std::string("well ") + econ_production_limits.followonWell() + std::string(" opens instead");
```
GitPaean: using fmt is more the standard approach now.  The message can also be made more clear that the well is opened as a follow on well of well `wname`.

opm/simulators/wells/SingleWellState.cpp
```diff
@@ -122,8 +122,11 @@ void SingleWellState<Scalar, IndexTraits>::open()
 }
 
 template<typename Scalar, typename IndexTraits>
-void SingleWellState<Scalar, IndexTraits>::updateStatus(Well::Status new_status)
+bool SingleWellState<Scalar, IndexTraits>::updateStatus(Well::Status new_status)
 {
+    if (this->status == new_status)
```
GitPaean: using {} unless you write them in one single line.

opm/simulators/wells/BlackoilWellModel_impl.hpp
GitPaean: It looks like the variable name needs to change to reflect the code change. 

and the content inside the `if (well_status_change)` might not be correct anymore.

totto82: I'm not sure if I understand what you mean here. I think the logic should be fine.

GitPaean:
```c++
                              constexpr auto events_mask = ScheduleEvents::WELL_STATUS_CHANGE |
                                                           ScheduleEvents::REQUEST_OPEN_WELL |
                                                           ScheduleEvents::REQUEST_SHUT_WELL;
```

The variable name `well_status_change` does not mean the well status change.  For example, if the the event triggered is `ScheduleEvents::REQUEST_OPEN_WELL` . If the well has been `OPEN` at this point, there is no status change (variable name is `well_status_change` at the same time), Then the question if whether we still need to go through the code in the `if` condition. 

It is also possible I miss something from the design side.

GitPaean: Please let me know whether you agree or do something here. 

And also, since we have a DATA file for this functionality in opm-tests, a regression test will be good.

totto82: I think I understand what you mean, but I am not sure what the "correct" thing to do here is. I think the Pr does not change the existing logic. I could try to re-name well_status_change to well_status_event and add a ws.status != well_ecl.getStatus() to check if there really was a change of status.

totto82: And yes, I can add the regression test.

GitPaean: > I think I understand what you mean, but I am not sure what the "correct" thing to do here is. I think the Pr does not change the existing logic. I could try to re-name well_status_change to well_status_event and add a ws.status != well_ecl.getStatus() to check if there really was a change of status.

I like this while it looks like it intervenes with the WCYCLE setup. If the fix is not easy, you can keep the variable name change and add some comments to explain some following up work need to do here for later reference. I did realize it is not introduced in this PR while since you are here, I thought it might be good to make it straight.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -815,11 +816,12 @@ namespace Opm {
...
                               constexpr auto events_mask = ScheduleEvents::WELL_STATUS_CHANGE |
-                                                           ScheduleEvents::REQUEST_OPEN_WELL;
-                              const bool well_status_change =
+                                                           ScheduleEvents::REQUEST_OPEN_WELL |
+                                                           ScheduleEvents::REQUEST_SHUT_WELL;
+                              const bool well_event =
                                   this->report_step_starts_ &&
                                   wg_events.hasEvent(well_ecl.name(), events_mask);
-                              if (well_status_change) {
+                              if (well_event) {
```
GitPaean, line 831: I am in the process of getting the PR in, please add a comment within the `if (well_event)` condition to say there are more to investigate here since it is not necessary the OPEN or SHUT action actually opened or shut a well here, we keep the code this way to avoid intervening with the WCYCLE implementation. 

So that we can pick it up when we have capacity.

GitPaean: @totto82 I would like to have a comment in the code since you are here. Basically the code itself needs some attention for potential follow-up PRs.

totto82: What do you want me to write? I don't know too much about WCYCLE and how it is supposed to work when combined with OPEN/SHUT events from the schedule.

GitPaean: I am thinking something like 
```c++
    // Note: There is more to investigate here. An OPEN or SHUT event does not
    // necessarily mean the well was actually opened or shut at this point. 
    // We keep the code in this form to avoid interfering with the WCYCLE implementation.
```

akva2: WCYCLE is suspended by explicit events, ie, if there is a shut event it stops cycling until there is an open event, at which point cycling restarts. This behavior is in the tests, so if they pass it is at least as correct as it used to be.

totto82: Thanks. I added a line or two so we don't forget.

akva2:
This or upstream broke norne in parallel, https://ci.opm-project.org/job/opm-simulators/2917/

bska:
Post-commit build is broken now.  Either fix **immediately** or back this out.

GitPaean:
> This or upstream broke norne in serial, https://ci.opm-project.org/job/opm-simulators/2917/
> and likely https://ci.opm-project.org/job/opm-nightly-pypi-test/732/

please provide some education, those are tests separately from the regression tests, right?

akva2:
yes, the post builder runs norne in serial and parallel.

akva2:
it takes longer on a shared ci system. 
```
jenkixs run norne norne_parallel pleasx
```

has always been available

PR #6051, opened 2025-03-06
Check status from wellstate not the well object from Schedule

totto82:
Related to https://github.com/OPM/opm-simulators/pull/6050

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1129,27 +1129,20 @@ assignDynamicWellStatus(data::Wells& wsrpt,
                         const int reportStepIndex) const
 {
     this->loopOwnedWells([this, reportStepIndex, &wsrpt]
-                         (const auto wellID, const Well& well)
+                         ([[maybe_unused]] const auto wellID,
+                          const Well& well)
     {
+
         auto& xwel = wsrpt[well.name()]; // data::Wells is a std::map<>
+        xwel.dynamicStatus = this->wellState().well(well.name()).status;
 
-        xwel.dynamicStatus = this->schedule()[reportStepIndex]
-            .wells(well.name()).getStatus();
-
-        if ((xwel.dynamicStatus == Well::Status::OPEN) &&
-            this->wellTestState().well_is_closed(well.name()) &&
-            !this->wasDynamicallyShutThisTimeStep(wellID))
-        {
-            // Well is supposed to be flowing according to the run
-            // specification (Schedule object), but it is not operable or
-            // cannot meet its economic limits (well testing).
-            //
-            // Assign status based on the well's defined automatic shut-in
-            // procedure.
-            xwel.dynamicStatus = well.getAutomaticShutIn()
-                ? Well::Status::SHUT : Well::Status::STOP;
+        // Well was shut this time step we keep it open one last time
+        // for output
+        if (wasDynamicallyShutThisTimeStep(well.name())) {
+            xwel.dynamicStatus = Well::Status::OPEN;
         }
     });
+
 }
 
 template<class Scalar>
@@ -1161,17 +1154,17 @@ assignShutConnections(data::Wells& wsrpt,
                          ([[maybe_unused]] const auto wellID,
                           const Well&                 well)
     {
-        auto& xwel = wsrpt[well.name()]; // data::Wells is a std::map<>
-        xwel.dynamicStatus  = this->wellState().well(well.name()).status;
-        if (wasDynamicallyShutThisTimeStep(well.name())) {
-            xwel.dynamicStatus = Well::Status::OPEN; // we will output
-        }
-        const auto wellStatusInSchedule = this->schedule().getWell(well.name(), reportStepIndex).getStatus();
-        const auto wellIsOpenInSchedule = wellStatusInSchedule == Well::Status::OPEN;
-        auto skip = [wellIsOpenInSchedule](const Connection& conn)
+        const auto wellIsOpen =
+            this->schedule()[reportStepIndex].wells(well.name())
+            .getStatus() == Well::Status::OPEN;
+
+        auto skip = [wellIsOpen](const Connection& conn)
         {
-            return wellIsOpenInSchedule && (conn.state() != Connection::State::SHUT);
+            return wellIsOpen && (conn.state() != Connection::State::SHUT);
         };
+
+        auto& xwel = wsrpt[well.name()]; // data::Wells is a std::map<>
+
         auto& xcon = xwel.connections;
         for (const auto& conn : well.getConnections()) {
             if (skip(conn)) {
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1129,20 +1129,18 @@ assignDynamicWellStatus(data::Wells& wsrpt,
                         const int reportStepIndex) const
 {
     this->loopOwnedWells([this, reportStepIndex, &wsrpt]
-                         ([[maybe_unused]] const auto wellID,
+                         (const auto wellID,
                           const Well& well)
     {
-
         auto& xwel = wsrpt[well.name()]; // data::Wells is a std::map<>
         xwel.dynamicStatus = this->wellState().well(well.name()).status;
 
         // Well was shut this time step we keep it open one last time
         // for output
-        if (wasDynamicallyShutThisTimeStep(well.name())) {
+        if (wasDynamicallyShutThisTimeStep(wellID)) {
             xwel.dynamicStatus = Well::Status::OPEN;
         }
     });
-
 }
 
 template<class Scalar>
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1161,21 +1161,17 @@ assignShutConnections(data::Wells& wsrpt,
                          ([[maybe_unused]] const auto wellID,
                           const Well&                 well)
     {
-        const auto wellIsOpen =
-            this->schedule()[reportStepIndex].wells(well.name())
-            .getStatus() == Well::Status::OPEN;
-
-    for (const auto& well : this->wells_ecl_) {
         auto& xwel = wsrpt[well.name()]; // data::Wells is a std::map<>
         xwel.dynamicStatus  = this->wellState().well(well.name()).status;
-        const auto wellIsOpen = xwel.dynamicStatus == Well::Status::OPEN;
-        auto skip = [wellIsOpen](const Connection& conn)
+        if (wasDynamicallyShutThisTimeStep(well.name())) {
+            xwel.dynamicStatus = Well::Status::OPEN; // we will output
+        }
+        const auto wellStatusInSchedule = this->schedule().getWell(well.name(), reportStepIndex).getStatus();
+        const auto wellIsOpenInSchedule = wellStatusInSchedule == Well::Status::OPEN;
+        auto skip = [wellIsOpenInSchedule](const Connection& conn)
         {
-            return wellIsOpen && (conn.state() != Connection::State::SHUT);
+            return wellIsOpenInSchedule && (conn.state() != Connection::State::SHUT);
         };
-
-        auto& xwel = wsrpt[well.name()]; // data::Wells is a std::map<>
-
         auto& xcon = xwel.connections;
         for (const auto& conn : well.getConnections()) {
             if (skip(conn)) {
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1165,6 +1165,10 @@ assignShutConnections(data::Wells& wsrpt,
             this->schedule()[reportStepIndex].wells(well.name())
             .getStatus() == Well::Status::OPEN;
 
+    for (const auto& well : this->wells_ecl_) {
+        auto& xwel = wsrpt[well.name()]; // data::Wells is a std::map<>
+        xwel.dynamicStatus  = this->wellState().well(well.name()).status;
+        const auto wellIsOpen = xwel.dynamicStatus == Well::Status::OPEN;
         auto skip = [wellIsOpen](const Connection& conn)
         {
             return wellIsOpen && (conn.state() != Connection::State::SHUT);
--- a/opm/simulators/wells/GuideRateHandler.cpp
+++ b/opm/simulators/wells/GuideRateHandler.cpp
@@ -668,10 +668,9 @@ updateGuideRatesForWells_()
         const auto well_index = this->well_state_.index(wname);
 
         if (well_index.has_value() && this->well_state_.wellIsOwned(*well_index, wname)) {
-            const auto& ws = this->well_state_.well(well_index.value());
-
             // the well is found and owned
-            const auto& wpot = ws.well_potentials;
+            const auto& wpot = this->well_state_.well(*well_index).well_potentials;
+
             if (o_pos >= 0) { well_pot[o_ix] = static_cast<double>(wpot[o_pos]); }
             if (g_pos >= 0) { well_pot[g_ix] = static_cast<double>(wpot[g_pos]); }
             if (w_pos >= 0) { well_pot[w_ix] = static_cast<double>(wpot[w_pos]); }
@@ -753,9 +752,6 @@ updateProductionGroupPotentialFromSubGroups(const Group& group, std::vector<Scal
         if (well_tmp.isInjector())
             continue;
 
-        if (well_tmp.getStatus() == Well::Status::SHUT)
-            continue;
-
         const auto& well_index = this->well_state_.index(well_name);
         if (!well_index.has_value()) // the well is not found
             continue;
@@ -764,6 +760,7 @@ updateProductionGroupPotentialFromSubGroups(const Group& group, std::vector<Scal
         {
             continue;
         }
+
         const auto& ws = this->well_state_.well(well_index.value());
         if (ws.status == Well::Status::SHUT)
             continue;
--- a/opm/simulators/wells/GuideRateHandler.cpp
+++ b/opm/simulators/wells/GuideRateHandler.cpp
@@ -668,9 +668,10 @@ updateGuideRatesForWells_()
         const auto well_index = this->well_state_.index(wname);
 
         if (well_index.has_value() && this->well_state_.wellIsOwned(*well_index, wname)) {
-            // the well is found and owned
-            const auto& wpot = this->well_state_.well(*well_index).well_potentials;
+            const auto& ws = this->well_state_.well(well_index.value());
 
+            // the well is found and owned
+            const auto& wpot = ws.well_potentials;
             if (o_pos >= 0) { well_pot[o_ix] = static_cast<double>(wpot[o_pos]); }
             if (g_pos >= 0) { well_pot[g_ix] = static_cast<double>(wpot[g_pos]); }
             if (w_pos >= 0) { well_pot[w_ix] = static_cast<double>(wpot[w_pos]); }
@@ -754,6 +755,7 @@ updateProductionGroupPotentialFromSubGroups(const Group& group, std::vector<Scal
 
         if (well_tmp.getStatus() == Well::Status::SHUT)
             continue;
+
         const auto& well_index = this->well_state_.index(well_name);
         if (!well_index.has_value()) // the well is not found
             continue;
@@ -762,8 +764,10 @@ updateProductionGroupPotentialFromSubGroups(const Group& group, std::vector<Scal
         {
             continue;
         }
-
         const auto& ws = this->well_state_.well(well_index.value());
+        if (ws.status == Well::Status::SHUT)
+            continue;
+
         for (int phase = 0; phase < this->num_phases_; phase++) {
             pot[phase] += wefac * ws.well_potentials[phase];
         }
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1639,8 +1639,7 @@ namespace Opm
         const auto well_status_orig = this->wellStatus_;
         const auto operability_orig = this->operability_status_;
         // don't allow opening wells that are stopped from schedule or has a stopped well state
-        const bool allow_open =  this->well_ecl_.getStatus() == WellStatus::OPEN &&
-                                 well_state.well(this->index_of_well_).status == WellStatus::OPEN;
+        const bool allow_open = well_state.well(this->index_of_well_).status == WellStatus::OPEN;
         // don't allow switcing for wells under zero rate target or requested fixed status and control
         const bool allow_switching = !this->wellUnderZeroRateTarget(simulator, well_state, deferred_logger) &&
                                      (!fixed_control || !fixed_status) && allow_open;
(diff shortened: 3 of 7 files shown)


totto82:

totto82:

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1103,10 +1103,7 @@ assignShutConnections(data::Wells& wsrpt,
 
     for (const auto& well : this->wells_ecl_) {
         auto& xwel = wsrpt[well.name()]; // data::Wells is a std::map<>
-
-        xwel.dynamicStatus = this->schedule()
-            .getWell(well.name(), reportStepIndex).getStatus();
-
+        xwel.dynamicStatus  = this->wellState().well(well.name()).status;
```
bska, line 1106: This breaks the `CTFAC` summary vector, especially in cases using `WELPI`.  Please see PR #2866, commit f6130861d.

totto82:

opm/simulators/wells/BlackoilWellModelGeneric.cpp
bska: > So the correct behavior is that CTFAC should be zero for dynamically shut wells.

Yes.  We altered this behaviour in b982ad0fd (PR #3093).

totto82: I think the test failure for the THP case is an improvement as the well is only temporary opened for well testing but still it gives output.

totto82:
I think the failures for WCYCLE is expected. As I anticipated in https://github.com/OPM/opm-simulators/pull/6050 the current code in master may not by 100 correct for WCYCLE. @akva2 I would appreciate if you could take a look and see if you agree.

akva2:
Firstly, the wcycle failures here are entirely trivial and indistinguishable to the eyeball norm. The min_bhp one is more severe (but in the correct direction) as some spikes disappeared there. As far as correctness of WCYCLE, I mostly did was monkey saw others monkey do, so it is entirely possible I have checked the wrong status variable. I assume you are not on the monkey level, monkey will trust big fire god.

totto82:
@bska Do you have any further comments on this PR? If we agree that this is more consistent I would like to merge this to as a preparation for  https://github.com/OPM/opm-simulators/pull/6050

bska (commented):
I don't understand what you're trying to do in `assignShutConnections()`.  What situation are you addressing?

The function was created to handle the case of a well which is open in the input, but which stops/shuts in as a result of dynamic condition&ndash;e.g., due to operability concerns. In that case we report some static information (and pressures/rates of zero) for the **first** step after shut-in and then zero thereafter.

totto82:
We are checking the "dynamicStatus" of the well in the summary output code. Currently, this is updated in assignShutConnections() 
` xwel.dynamicStatus  = this->wellState().well(well.name()).status;`
I would like to use this feature when handling item 9 in WECON. See https://github.com/OPM/opm-simulators/pull/6050
Does it make sense to move this out from this function (maybe to a separate assignDynamicWellStatus()) since it is not only relevant for assignShutConnections()

bska:
> I would like to use this feature when handling item 9 in WECON. See #6050.

Right.  Function `assignShutConnections()` is supposed to leave (the connection information of) flowing wells untouched, so if the well is currently flowing as determined by [`SingleWellState::status`](https://github.com/OPM/opm-simulators/blob/66449c6b26b286439d30b1b534594abaf5167574/opm/simulators/wells/SingleWellState.hpp#L86) or some other authoritative source, then this function should do nothing.

So far so good.  The problem comes for wells which are **not** flowing.  Most of the logic in `assignShutConnections()` exists to handle the case of a well which is dynamically shut in, but which is supposed to be flowing according to the simulation input, i.e., in the `SCHEDULE` section.  When we added this function, the only case in which a well would be recorded as non-flowing was in the context of well testing (`WTEST` keyword).  That's why it assigns the `dynamicStatus` based on `wellTestState()`.  Furthermore, the condition concerning `wasDynamicallyShutThisTimeStep()` is because we're supposed to be reporting the values as if the well is flowing at the end of the time step at which the well is shut in.

Finally, the `skip()` helper function is intended to ignore open connections in flowing wells while ensuring that the $Kh$ product, the CTF, and the connection $D$ factor are reported as their input values for shut connections in flowing wells, and all connections in shut-in wells.  I'd really like to keep that behaviour.

totto82:
Thank you for the explanation. I think the implementation in this PR should align with your explanation, but maybe I still misunderstood something. The test failures are related to WCYCLE, for which I think this PR gives an improvement.

totto82:
@bska Thanks. I didn't notice. I assume I need to move my stuff to `assignDynamicWellStatus`

totto82:

bska:
> Did I get it right this time?

It Depends[tm].  Specifically, it depends on subtleties of definition.  The `OP_1` production well of the [`MIN_THP_1`](https://github.com/OPM/opm-tests/blob/a2f5b9998de6282d7642527d18d06d46b5ed5b3f/wtest/thp_min/MIN_THP_1.DATA) is inoperable and closes (shuts in) due to "physical reasons" fairly early in the simulation run.  However, it gets re-opened in [`WTEST`](https://github.com/OPM/opm-tests/blob/a2f5b9998de6282d7642527d18d06d46b5ed5b3f/wtest/thp_min/MIN_THP_1.DATA#L343-L345) 30 days after its initial shut-in.  The well is still not operable and immediately shuts in, but the current master sources still reports the well flowing at its test point.  This PR changes that behaviour and reports the well being shut from the initial shut-in point and through to the end of the simulation.  As for what kind of reporting behaviour we actually want in this situation I don't really know.  Perhaps @GitPaean could give some guidance?

GitPaean:
> The well is still not operable and immediately shuts in, but the current master sources still reports the well flowing at its test point. 

There has been some development/verification along this. Personally I think not reporting is fine/good if the re-opening is not successful, while there might be some observation that supports to output the values before it gets SHUT. Let me look into it. @steink , do you remember anything regarding this?

GitPaean:
When looking at what exactly happens, it shows something not directly related to this PR, while show how the wells are tested. 

The testing efforts failed because 
```c++
        while (testWell) {
            const std::size_t original_number_closed_completions = welltest_state_temp.num_closed_completions();
            bool converged = solveWellForTesting(simulator, well_state_copy, group_state, deferred_logger);
            if (!converged) {
                const auto msg = fmt::format("WTEST: Well {} is not solvable (physical)", this->name());
                deferred_logger.debug(msg);
                return;
            }
```
did not manage to get converged. 

And in the function `solveWellForTesting()` because `well_state.well(this->index_of_well_).status` is `SHUT`, it enters
the `else` condition branch. 
```c++
                if (this->param_.use_implicit_ipr_ && this->well_ecl_.isProducer() && this->wellHasTHPConstraints(summary_state) && (well_state.well(this->index_of_well_).status == WellStatus::OPEN)) {
                    converged = solveWellWithTHPConstraint(simulator, dt, inj_controls, prod_controls, well_state, group_state, deferred_logger);
                } else {
                    converged = this->iterateWellEqWithSwitching(simulator, dt, inj_controls, prod_controls, well_state, group_state, deferred_logger);
                }
```
And in the function `iterateWellEqWithSwitching()`, `allow_switching` will be `false` due to the following code, 
```c++
        const bool allow_open = well_state.well(this->index_of_well_).status == WellStatus::OPEN;
        // don't allow switcing for wells under zero rate target or requested fixed status and control
        const bool allow_switching =
            !this->wellUnderZeroRateTarget(simulator, well_state, deferred_logger) &&
            (!fixed_control || !fixed_status) && allow_open;
```

Then it tries to solve with `allow_switching` to be `false` and did not manage to get converged. 

I can be wrong, but it looks like very worrying, since I believe `solveWellForTesting()` is probably to check whether the well is solvable at all with the least restrictive condition (e.g. BHP constraint). 

I can be wrong.   @steink , any comments regarding this part?

GitPaean:
> @GitPaean Thanks for noticing. I think we should add a ws.open() to the well_state_copy before we do the well testing. I will update the PR accordingly to test.

Yeah. Now I see it is some difference introduced from this PR.

GitPaean:
> > well_state.well(this->index_of_well_).status is SHUT
> 
> Are you sure about this? When I just tested, the status was OPEN at the point you refer to.

I think so, with this PR. And make sure you enter there through the well testing. And also --enable-tuning=true .

GitPaean:
And in the master branch, the well testing attempt (re-opening) was successful for that time step.  And a side-line comment is the well testing part of code is too complicated (arguably most complicated code that I can tell) and probably involving repeated process and remaining to be cleaned up some.

totto82:
@GitPaean  Could you test https://github.com/OPM/opm-simulators/pull/6319 and see if you see the same instability wrt. the well re-opening or not.

GitPaean:
> @GitPaean Could you test #6319 and see if you see the same instability wrt. the well re-opening or not.

If you are referring to the difference between debug build and release build, it is not the case. I accidentally ran with the wrong build. Sorry for removing the message without clarifying.

totto82:

totto82:

totto82:
In master for UDT-1D-03.DATA: 
```
Starting time step 22, stepsize 1 days, at day 23/45, date = 24-Jan-2023
Well PROD1 will be stopped as it can not operate under current reservoir conditions.
 Newton its= 2, linearizations= 3 (0.0sec), linear its=  1 (0.0sec)

Starting time step 23, stepsize 1 days, at day 24/45, date = 25-Jan-2023
    Well PROD1 control mode changed from BHP to ORAT
    Well PROD2 control mode changed from ORAT to BHP
 Newton its= 3, linearizations= 4 (0.0sec), linear its=  0 (0.0sec)
```
With PR
```
Starting time step 22, stepsize 1 days, at day 23/45, date = 24-Jan-2023
Well PROD1 will be stopped as it can not operate under current reservoir conditions.
 Newton its= 2, linearizations= 3 (0.0sec), linear its=  1 (0.0sec)

Starting time step 23, stepsize 1 days, at day 24/45, date = 25-Jan-2023
    Well PROD2 control mode changed from ORAT to BHP
 Newton its= 2, linearizations= 3 (0.0sec), linear its=  2 (0.0sec)
```
I don't think the behaviour in the master is correct. We should not change control (re-open) a stopped well without WTEST.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1636,8 +1636,7 @@ namespace Opm
         const auto well_status_orig = this->wellStatus_;
         const auto operability_orig = this->operability_status_;
         // don't allow opening wells that are stopped from schedule or has a stopped well state
-        const bool allow_open =  this->well_ecl_.getStatus() == WellStatus::OPEN &&
-                                 well_state.well(this->index_of_well_).status == WellStatus::OPEN;
+        const bool allow_open = well_state.well(this->index_of_well_).status == WellStatus::OPEN;
```
GitPaean, line 1639: I think it is this line change the behavoir in the case `UDT-1D-03.DATA`.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -292,7 +292,8 @@ namespace Opm
             from = WellProducerCMode2String(ws.production_cmode);
         }
         const bool oscillating = std::count(this->well_control_log_.begin(), this->well_control_log_.end(), from) >= this->param_.max_number_of_well_switches_;
-        if (oscillating || this->wellUnderZeroRateTarget(simulator, well_state, deferred_logger) || !(this->well_ecl_.getStatus() == WellStatus::OPEN)) {
+
+        if (oscillating || this->wellUnderZeroRateTarget(simulator, well_state, deferred_logger) || !(well_state.well(this->index_of_well_).status == WellStatus::OPEN)) {
            return false;
```
GitPaean, line 297: I was wrong, it is this line that changes the simulation results for UDT-1D-03.DATA.

GitPaean: We have a few different ways to stop a well for different occasions. `this->wellUnderZeroRateTarget` should probably be `this->stoppedOrZeroRateTarget` here. 

Not totally sure about the purpose of `!(this->well_ecl_.getStatus() == WellStatus::OPEN)`, which is probably for some corner cases. 

@steink , do you have any comment regarding this.

totto82:

totto82:
@GitPaean I just rebased this. I would appreciate it if you could take a look at this again.

GitPaean:
> @GitPaean I just rebased this. I would appreciate it if you could take a look at this again.

just saw this.  Will have a look.

opm/simulators/wells/WellState.cpp
```diff
@@ -377,9 +377,6 @@ void WellState<Scalar>::init(const std::vector<Scalar>& cellPressures,
     // order may change so the mapping is based on the well name
     if ((prevState != nullptr) && (prevState->size() > 0)) {
         for (int w = 0; w < nw; ++w) {
-            if (wells_ecl[w].getStatus() == Well::Status::SHUT) {
```
GitPaean: Is removing this intentional?  If the SCHEDULE specifies this well as SHUT, do we have some way to make it OPEN and run?

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -1098,7 +1119,17 @@ getGuideRateInj(const std::string& name,
         if (!wellTmp.isInjector())
             continue;
 
-        if (wellTmp.getStatus() == Well::Status::SHUT)
+        const auto& well_index = wellState.index(wellName);
+        if (!well_index.has_value())
+            continue;
+
+        if (! wellState.wellIsOwned(well_index.value(), wellName) ) // Only sum once
+        {
```
GitPaean, line 1127: It is good that you have `{` and `}` for this, while it looks better if you keep the same style with the surrounding code.  There are multiple occasions of this. I do not consider this so critical though.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -1044,7 +1055,17 @@ getGuideRate(const std::string& name,
         if (wellTmp.isInjector())
             continue;
 
-        if (wellTmp.getStatus() == Well::Status::SHUT)
+        const auto& well_index = wellState.index(wellName);
```
GitPaean: It does not hurt much to use reference, I do not think it is very useful either.

opm/simulators/wells/StandardWell_impl.hpp
```diff
@@ -2393,8 +2393,7 @@ namespace Opm
         auto well_status_cur = well_status_orig;
         int status_switch_count = 0;
         // don't allow opening wells that are stopped from schedule or has a stopped well state
-        const bool allow_open =  this->well_ecl_.getStatus() == WellStatus::OPEN &&
-                                 well_state.well(this->index_of_well_).status == WellStatus::OPEN;
+        const bool allow_open = well_state.well(this->index_of_well_).status == WellStatus::OPEN;
```
GitPaean, line 2396: the comments above requires update following the code change.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1639,8 +1639,7 @@ namespace Opm
...
```
GitPaean, line 1642: the comments above requires update following the code change.

opm/simulators/wells/GuideRateHandler.cpp
```diff
@@ -754,6 +755,7 @@ updateProductionGroupPotentialFromSubGroups(const Group& group, std::vector<Scal
 
         if (well_tmp.getStatus() == Well::Status::SHUT)
```
GitPaean: Do you want to remove this checking also, following code changes in the other occasion?

opm/simulators/wells/GuideRateHandler.cpp
```diff
@@ -668,9 +668,10 @@ updateGuideRatesForWells_()
         const auto well_index = this->well_state_.index(wname);
 
         if (well_index.has_value() && this->well_state_.wellIsOwned(*well_index, wname)) {
-            // the well is found and owned
-            const auto& wpot = this->well_state_.well(*well_index).well_potentials;
+            const auto& ws = this->well_state_.well(well_index.value());
 
+            // the well is found and owned
+            const auto& wpot = ws.well_potentials;
```
GitPaean: this is a pure code rewriting, is this intentional?

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1129,27 +1129,20 @@ assignDynamicWellStatus(data::Wells& wsrpt,
...
-            // procedure.
-            xwel.dynamicStatus = well.getAutomaticShutIn()
-                ? Well::Status::SHUT : Well::Status::STOP;
+        // Well was shut this time step we keep it open one last time
+        // for output
+        if (wasDynamicallyShutThisTimeStep(well.name())) {
+            xwel.dynamicStatus = Well::Status::OPEN;
         }
     });
+
```
GitPaean: adding a blank line is most likely not intentional.

opm/simulators/wells/GuideRateHandler.cpp
```diff
@@ -754,6 +755,7 @@ updateProductionGroupPotentialFromSubGroups(const Group& group, std::vector<Scal
 
         if (well_tmp.getStatus() == Well::Status::SHUT)
             continue;
+
```
GitPaean: Is this blank line intentional?

GitPaean (changes requested):
I have gone through the PR, there are only small comments about the form of the code changes.  The changes in the regression test looks fine. 

From my side, after addressing the comments raised, we can update_data and merge.

opm/simulators/wells/WellState.cpp
totto82: If I recall correctly, I removed it to support item 9 in WECON.

GitPaean:

opm/simulators/wells/WellState.cpp
GitPaean: Let us deal with it in the corresponding PR. If it is not related to this PR, let us not do it here.

opm/simulators/wells/WellGroupHelpers.cpp
GitPaean: maybe my language is not clear. I do not think we should reference here. It causes confusion without bringing any benefit.

totto82: Sure, and I agree. I can change this. I just copied these lines from other parts of the code.

opm/simulators/wells/WellState.cpp
totto82: OK

opm/simulators/wells/GuideRateHandler.cpp
bska: > is this intentional?

Intentional or not, there's no reason to change this.  Please revert to the original code.

opm/simulators/wells/GuideRateHandler.cpp
```diff
@@ -752,18 +753,17 @@ updateProductionGroupPotentialFromSubGroups(const Group& group, std::vector<Scal
...
-        if (well_tmp.getStatus() == Well::Status::SHUT)
-            continue;
         const auto& well_index = this->well_state_.index(well_name);
         if (!well_index.has_value()) // the well is not found
             continue;
 
         if (!this->well_state_.wellIsOwned(well_index.value(), well_name) ) // Only sum once
-        {
             continue;
-        }
```
bska: Please keep the braces.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -117,20 +117,18 @@ namespace Opm {
                 continue;
 
             if (! wellState.wellIsOwned(well_index.value(), wellName) ) // Only sum once
-            {
                 continue;
-            }
```
bska: Please keep the braces.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -302,19 +300,18 @@ sumSolventRates(const Group& group,
             continue;
 
         if (! wellState.wellIsOwned(well_index.value(), wellName) ) // Only sum once
-        {
             continue;
-        }
```
bska: Please keep the braces.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1129,25 +1129,17 @@ assignDynamicWellStatus(data::Wells& wsrpt,
                         const int reportStepIndex) const
 {
     this->loopOwnedWells([this, reportStepIndex, &wsrpt]
-                         (const auto wellID, const Well& well)
+                         ([[maybe_unused]] const auto wellID,
+                          const Well& well)
     {
-        auto& xwel = wsrpt[well.name()]; // data::Wells is a std::map<>
```
bska, line 1134: Why did you move this statement down?

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1129,25 +1129,17 @@ assignDynamicWellStatus(data::Wells& wsrpt,
...
-            // specification (Schedule object), but it is not operable or
-            // cannot meet its economic limits (well testing).
-            //
-            // Assign status based on the well's defined automatic shut-in
-            // procedure.
-            xwel.dynamicStatus = well.getAutomaticShutIn()
-                ? Well::Status::SHUT : Well::Status::STOP;
+        // Well was shut this time step we keep it open one last time
+        // for output
+        if (wasDynamicallyShutThisTimeStep(well.name())) {
```
bska: Why did you switch to name-based lookup here?  Is there a reason not to continue using the ID-based lookup?

bska (commented):
For the most part, this looks okay to me, except for a couple of minor issues.

opm/simulators/wells/GuideRateHandler.cpp
totto82: @GitPaean Just asked me to remove them. I am fine with both.

GitPaean: Yeah. My comments are about that other if conditions added in this PR do not use braces and this one is the only one (multiple occasions, probably from copy-and-paste) that uses `braces`, I suggested to remove this one for consistency. 

If later we want to add braces across the whole file, we can add them all together. 

If we gonna keep the braces, then the suggestion be the other way around,  to be 

please add braces for all the if conditions in this PR. 

I am okay with either direction also.

totto82: I followed @bska's comments and added the braces back. I don't plan to add or remove more braces in this PR. I hope that is okay. If we want to clean up this, we should agree whether we should have or not have braces for one-liners and then do it for the whole code.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1129,25 +1129,16 @@ assignDynamicWellStatus(data::Wells& wsrpt,
                         const int reportStepIndex) const
 {
     this->loopOwnedWells([this, reportStepIndex, &wsrpt]
-                         (const auto wellID, const Well& well)
+                         ([[maybe_unused]] const auto wellID,
```
bska: On more thing here.  Since we're now back to unconditionally using the `wellID`, I'd prefer if we remove the `[[maybe_unused]]` attribute on this parameter.

totto82:

GitPaean:
> jenkins build this failure_report please

Jenkins is experiencing difficulties.  I think we need to wait.

totto82:

bska:

GitPaean (approved):
Thanks for the update. I approve the PR now and update the reference for merging.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
GitPaean: @totto82 , if you have time you can recover this line before it gets merged.

GitPaean:
The reference has been updated. I am merging the PR now.

PR #6044, opened 2025-03-04
Tracer: add some structs for well rate output and avoid complex keys

akva2:
Not really inner-loopy so not much gained in terms of efficiency. But seems to yield, if anything, a small speedup

# Timings for NORNE_ATW2013 [5 runs]
|     Revision      |tracerAdvance|tracerAssemble|Pre/post step|p-value(tracerAssemble)|
|-------------------|-------------|--------------|-------------|----------------------:|
|tracer_rate_structs|31.91 +- 0.27|25.72 +- 0.22 |53.72 +- 0.35|                   0.04|
|master             |32.68 +- 0.23|26.31 +- 0.20 |54.69 +- 0.33|                       |

# Timings for NORNE_ATW2013_1A_MSW [5 runs]
|     Revision      |tracerAdvance|tracerAssemble|Pre/post step|p-value(tracerAssemble)|
|-------------------|-------------|--------------|-------------|----------------------:|
|tracer_rate_structs|35.27 +- 0.11|31.19 +- 0.10 |61.14 +- 0.20|                   0.09|
|master             |35.80 +- 0.26|31.61 +- 0.23 |62.16 +- 0.37|                       |

PR #6052, opened 2025-03-06
Fix parallel tracers

akva2:
I mixed up some indices in #6044 

Also add parallel exception handling to tracer equation assembly (which would have caught the issue in a more readable way..)

### DIFF
--- a/opm/simulators/flow/TracerModel.hpp
+++ b/opm/simulators/flow/TracerModel.hpp
@@ -500,8 +500,7 @@ protected:
         }
 
         const Scalar dt = simulator_.timeStepSize();
-        const std::size_t well_index = eclWell.seqIndex();
-        const auto& ws = simulator_.problem().wellModel().wellState().well(well_index);
+        const auto& ws = simulator_.problem().wellModel().wellState().well(well.name());
         for (std::size_t i = 0; i < ws.perf_data.size(); ++i) {
             const auto I = ws.perf_data.cell_index[i];
             const Scalar rate = well.volumetricSurfaceRateForConnection(I, tr.phaseIdx_);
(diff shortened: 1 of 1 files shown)


akva2:
jenkins run norne please

opm/simulators/flow/TracerModel.hpp
```diff
@@ -499,7 +500,7 @@ class TracerModel : public GenericTracerModel<GetPropType<TypeTag, Properties::G
         }
 
         const Scalar dt = simulator_.timeStepSize();
-        const std::size_t well_index = eclWell.seqIndex();
+        const std::size_t well_index = simulator_.problem().wellModel().wellState().index(well.name()).value();
         const auto& ws = simulator_.problem().wellModel().wellState().well(well_index);
```
bska, line 504: The `well_index` isn't really used anywhere else, so you might as well use the [name based lookup](https://github.com/OPM/opm-simulators/blob/c7c0aeb4e10509d3dce8357f32fdcdc9c5f405c1/opm/simulators/wells/WellState.hpp#L332) for `ws`:
```C++
const auto& ws = simulator_.problem().wellModel().wellState.well(well.name());
```
That will bypass this problem completely.

bska (commented):
This looks correct to me, although I'm not yet too familiar with the inner workings of this structure.  Better diagnostics are of course always beneficial.  I just have one minor suggestion that **might** help avoiding this particular problem in the fututre.

opm/simulators/flow/TracerModel.hpp
akva2: Indeed, if that is available.. This is basically reverting to the old code, I mixed up the well state index and the seq index (which will match in serial only). Have applied this improvement.

akva2:

opm/simulators/flow/TracerModel.hpp
akva2: The new structure is a change from ```map<pair<well_name, tracer_name>, Scalar>``` to ```map<well_index, struct_with_tracer_name_and_scalar>```

bska: > The new structure is a change from `map<pair<well_name, tracer_name>, Scalar>` to `map<well_index, struct_with_tracer_name_and_scalar>`

Sure, I appreciate that and it's a clear improvement, but the `WellState` uses numbering relative to `wells_ecl_` which is local to each process and there's no direct way to compute that index if you "just" have a `Well` object.  If on the other hand you have a `WellInterfacePtr`, then you can use `indexOfWell()` as the lookup index.

akva2: yes, as I now learned the hard way.

bska (approved):

opm/simulators/flow/TracerModel.hpp
bska: > yes, as I now learned the hard way.

Right, and my point is that if you're not going to use that index for anything else, you might as well just use the name-based lookup.

PR #6053, opened 2025-03-06
Prevent zero division

michal-toth:
This bug concerns cprw linear solver (the default one) in the situation when a well has no active perforations on some rank. Other linear solvers are fine.

The diagonal element of the coarse (pressure) system's Jacobian for the well equation is normally calculated from the average cell weight over well's perforations. If there are no perforations we end up dividing zero by zero. This fix prevents the zero division and also adds a positive number to the Jacobian's diagonal. Other entries of that row are zero, so this regularizes the matrix. Since the well has no active perforations, the value of the variable has no effect on the solution. (Any nonzero value on the diagonal would work - tested with 1 and 1e-100).

### DIFF
--- a/opm/simulators/wells/StandardWellEquations.cpp
+++ b/opm/simulators/wells/StandardWellEquations.cpp
@@ -336,7 +336,7 @@ extractCPRPressureMatrix(PressureMatrix& jacobian,
     if (nperf != 0)
         cell_weights /= nperf;
     else {
-        // Edge case: the well has no active perforations on this rank (duneC_.size==0).
+        // duneC_.size==0, which can happen e.g. if a well has no active perforations on this rank.
         // Add positive weight to diagonal to regularize Jacobian (other row entries are 0).
         // Row's variable has no observable effect, since there are no perforations.
         cell_weights = 1.;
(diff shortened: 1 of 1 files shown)


blattms:
This actually just hit me on run a serial run. I activated floating pointing exceptions to find other problems and it signals here.

Great catch, IMHO.

opm/simulators/wells/StandardWellEquations.cpp
```diff
@@ -333,7 +333,14 @@ extractCPRPressureMatrix(PressureMatrix& jacobian,
             nperf += 1;
         }
     }
-    cell_weights /= nperf;
+    if (nperf != 0)
+        cell_weights /= nperf;
+    else {
+        // Edge case: the well has no active perforations on this rank (duneC_.size==0).
```
blattms, line 339: I am not 100% sure whether this comment is correct. The if statement above might be false for other reasons.

michal-toth: True. I adjusted the comment.

blattms:
jenkins build this please

PR #5499, opened 2024-08-02
Write vtk (if enabled) for dry runs

daavid00:
To write the .vtu when --enable-vtk-output=1 --enable-dry-run=1

PR #6057, opened 2025-03-06
Fix to write vtk (if enabled) for dry runs in cases without oilPhase

daavid00:
In #5499 this was tested for cases with oil. For a case without it, then some assertions fail (e.g., [updateRelpermAndPressures](https://github.com/OPM/opm-simulators/blob/96ec6ad143b4106ac601546e196ecb318faaca3f/opm/models/blackoil/blackoilintensivequantities.hh#L297)).

This PR fixes this.


daavid00:
Jenkins build this please

totto82:
Thanks for the fix.

PR #6059, opened 2025-03-07
Improving terminal/dbg output

daavid00:
Removing/adding some line breaks and spaces for cosmetics.

### DIFF
--- a/opm/simulators/linalg/ISTLSolver.hpp
+++ b/opm/simulators/linalg/ISTLSolver.hpp
@@ -312,7 +312,7 @@ std::unique_ptr<Matrix> blockJacobiAdjacency(const Grid& grid,
             // Print parameters to PRT/DBG logs.
             if (on_io_rank && parameters_[activeSolverNum_].linear_solver_print_json_definition_) {
                 std::ostringstream os;
-                os << "Property tree for linear solvers:\n";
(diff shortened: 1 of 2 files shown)


daavid00:
jenkins build this please

opm/simulators/flow/GenericCpGridVanguard.cpp
```diff
@@ -438,7 +438,7 @@ void GenericCpGridVanguard<ElementMapper,GridView,Scalar>::doCreateGrids_(Eclips
     if (isRoot) {
         input_grid = &eclState.getInputGrid();
         global_porv = eclState.fieldProps().porv(true);
-        OpmLog::info("\nProcessing grid");
+        OpmLog::info("Processing grid");
```
bska, line 441: I think I'd like to keep this newline, please.  It makes the 'Processing' stage stand out a bit.

bska (commented):
Fair enough.  The only thing I'd prefer is if the grid processing message be kept in its current form.

opm/simulators/flow/GenericCpGridVanguard.cpp
daavid00: Ok, this is now as before.

bska:
jenkins build this please

bska (approved):

PR #6060, opened 2025-03-07
Fix to write vtk (if enabled) for dry runs in cases with drsdt

daavid00:
This showed up when dry running the SPE1CASE1.DATA with vtk output, which has DRSDT in the SCHEDULE.


daavid00:
jenkins build this please

daavid00:
jenkins build this please

totto82:
Thanks for the fix and the update.

PR #6063, opened 2025-03-10
Cleanup: simplify standard well equations

michal-toth:
I just stumbled upon this while studying cprw solver. The code can be simplified in two ways:

1. The vector `rhs` has exactly one nonzero element equal to 1. The `for` loop accumulating elements multiplied by `rhs` can be thus simplified to an assignment.
2. The transposed `inv_diag_block` matrix does not have to be constructed. We access its elements, so swapping indices suffices.

### DIFF
--- a/opm/simulators/wells/StandardWellEquations.cpp
+++ b/opm/simulators/wells/StandardWellEquations.cpp
@@ -345,8 +345,9 @@ extractCPRPressureMatrix(PressureMatrix& jacobian,
         //NB! use this options without treating pressure controlled separated
         //NB! calculate quasiimpes well weights NB do not work well with trueimpes reservoir weights
         Scalar abs_max = 0;
-        DiagMatrixBlockWellType inv_diag_block = invDuneD_[0][0];
+        const DiagMatrixBlockWellType& inv_diag_block = invDuneD_[0][0];
         for (std::size_t i = 0; i < blockSz; ++i) {
+            // calculate (D^-T*rhs)[i] where rhs is unit vector e_{bhp_var_index}
             bweights[0][i] = inv_diag_block[bhp_var_index][i];
             abs_max = std::max(abs_max, std::fabs(bweights[0][i]));
         }
--- a/opm/simulators/wells/StandardWellEquations.cpp
+++ b/opm/simulators/wells/StandardWellEquations.cpp
@@ -346,10 +346,8 @@ extractCPRPressureMatrix(PressureMatrix& jacobian,
         //NB! calculate quasiimpes well weights NB do not work well with trueimpes reservoir weights
         Scalar abs_max = 0;
         DiagMatrixBlockWellType inv_diag_block = invDuneD_[0][0];
-        DiagMatrixBlockWellType inv_diag_block_transpose =
-            Opm::wellhelpers::transposeDenseDynMatrix(inv_diag_block);
         for (std::size_t i = 0; i < blockSz; ++i) {
-            bweights[0][i] = inv_diag_block_transpose[i][bhp_var_index];
+            bweights[0][i] = inv_diag_block[bhp_var_index][i];
             abs_max = std::max(abs_max, std::fabs(bweights[0][i]));
         }
         assert(abs_max > 0.0);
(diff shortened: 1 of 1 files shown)


michal-toth:
jenkins build this please

blattms:
benchmark please

ytelses:
Benchmark result overview: 




| Test    | Configuration                                                                                 | Relative |
| ------- | --------------------------------------------------------------------------------------------- | -------- |
| opm-git | OPM Benchmark: drogon - Threads: 1                                                            | 1.001    |
| opm-git | OPM Benchmark: drogon - Threads: 8                                                            | 1.053    |
| opm-git | OPM Benchmark: punqs3 - Threads: 1                                                            | 0.998    |
| opm-git | OPM Benchmark: punqs3 - Threads: 8                                                            | 0.991    |
| opm-git | OPM Benchmark: smeaheia - Threads: 1                                                          | 1.04     |
| opm-git | OPM Benchmark: smeaheia - Threads: 8                                                          | 1.121    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 1                                                     | 1.004    |
| opm-git | OPM Benchmark: spe10_model_1 - Threads: 8                                                     | 0.984    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1                                                    | 1.004    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8                                                    | 1.173    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1                                                    | 1.001    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8                                                    | 1.085    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 - FOPT (Total Oil Production At End Of Run) | 1        |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 - FOPT (Total Oil Production At End Of Run) | 1        |



View result details @ https://www.ytelses.com/opm/?page=result&id=2712

atgeirr (commented):
Code looks correct to me.

However, I think it is not executed at all in the normal case, since the line
```C++
prm.put("preconditioner.use_well_weights", "false"s);
```
in [setupPropertyTree.cpp](https://github.com/OPM/opm-simulators/blob/029c4ffe8a0fc4851ec5b7d0b6c817f21e3a1f04/opm/simulators/linalg/setupPropertyTree.cpp#L143) means we use the other branch, unless you pass a custom JSON definition for the linear solver.

Therefore it would be prudent to do some testing with a custom definition setting this parameter to true, before merging.

blattms:
> However, I think it is not executed at all in the normal case,

Whether this is called is subject to `add_well` properties of the CPR preconditioner (see [PressureBhpTransferPolicy.hpp#L203](https://github.com/OPM/opm-simulators/blob/9f3f8897d371c7b38812c33b1b49965c800a50d1/opm/simulators/linalg/PressureBhpTransferPolicy.hpp#L203) and that seem to be true for our default, see [setupPropertyTree.cpp#L144](https://github.com/OPM/opm-simulators/blob/9f3f8897d371c7b38812c33b1b49965c800a50d1/opm/simulators/linalg/setupPropertyTree.cpp#L144).

blattms:
I take back my statement above as it does not seem to hold.

hnil:
The use_well_weights has to be considered experimental. It is a test to try to make quasiimpes type weights for the well system i.e. get a uniform method to get weights. However the work on it was abandoned since it is not obviouse how to do it for multisegment wells. Also in the case of multi segment wells the cprw is a bit different than nomral, since it combine the full well into one node i.e. it both "coarsen" in dofs( only use one degree of freedom for the full well) and in variables(only keep bhp/pressure for the full well). The branch may be removed, but it also maybe usefull for extending the cprw.

michal-toth:
> The use_well_weights has to be considered experimental. It is a test to try to make quasiimpes type weights for the well system i.e. get a uniform method to get weights. However the work on it was abandoned since it is not obviouse how to do it for multisegment wells. Also in the case of multi segment wells the cprw is a bit different than nomral, since it combine the full well into one node i.e. it both "coarsen" in dofs( only use one degree of freedom for the full well) and in variables(only keep bhp/pressure for the full well). The branch may be removed, but it also maybe usefull for extending the cprw.

Thank you for the clarification. I wanted to make a regression test for this case and was surprised how horrible the performance with `use_well_weights` is in parallel (sequential is fine). Since it is experimental, unfinished, and no longer in development we should consider removing this branch of code, e.g. by `constexpr if(false) in case we want to keep the code around.

PR #6067, opened 2025-03-12
Dont stop force shut wells

totto82:
If a well is shut due to repeated convergence issues we actually shut it and not just stopped. This fixes an issue reported by @erikhide on NORNE_ATW2013_3B_STDW.DATA where the simulator just retries the timestep forever.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1477,6 +1477,7 @@ forceShutWellByName(const std::string& wellname,
             }
             wellTestState().close_well(wellname, WellTestConfig::Reason::PHYSICAL, simulation_time);
             well_was_shut = 1;
+            force_shut_wells_.push_back(well->indexOfWell());
             break;
         }
     }
@@ -1516,6 +1517,20 @@ inferLocalShutWells()
             this->local_shut_wells_.push_back(wellID);
         }
     }
+}
+
+template<class Scalar>
+bool BlackoilWellModelGeneric<Scalar>::
+isForceShutWell(int wellId) const
+{
+    return std::count(force_shut_wells_.begin(), force_shut_wells_.end(), wellId) > 0;
+}
+
+template<class Scalar>
+void BlackoilWellModelGeneric<Scalar>::
+removeFromForceShutWell(int wellId)
+{
+    std::remove(force_shut_wells_.begin(), force_shut_wells_.end(), wellId);
 }
 
 template<class Scalar>
@@ -2077,6 +2092,7 @@ operator==(const BlackoilWellModelGeneric& rhs) const
         && this->report_step_starts_ == rhs.report_step_starts_
         && this->last_run_wellpi_ == rhs.last_run_wellpi_
         && this->local_shut_wells_ == rhs.local_shut_wells_
+        && this->force_shut_wells_ == rhs.force_shut_wells_
         && this->closed_this_step_ == rhs.closed_this_step_
         && this->node_pressures_ == rhs.node_pressures_
         && this->prev_inj_multipliers_ == rhs.prev_inj_multipliers_
--- a/opm/simulators/wells/BlackoilWellModelGeneric.hpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.hpp
@@ -213,6 +213,9 @@ public:
                              const double simulation_time,
                              const bool dont_shut_grup_wells);
 
+    bool isForceShutWell(int wellId) const;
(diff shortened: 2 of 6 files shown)


totto82:
jenkins build this please

erikhide:
I've tested the PB on NORNE_ATW2013_3B_STDW.DATA (as well as another model for which the same problem occurred), and the issue seems to be resolved.

GitPaean:
Because there is no WTEST keyword in `NORNE_ATW2013_3B_STDW`, what this PR does is that if a well is `shut` due to `convergence` issue, we will not open it again. 

In the master branch, the well `D-3AH` manages to produce for some while, while it gets stuck with cycle of re-opening and closing due to convergence issue. 

Is my observation correct?

GitPaean:
This is how it gets stuck without the PR. I am inviting @atgeirr and @steink to give some insights also.  It is a little harsh to not open a well can do some work, while we need to have a solution to get out of shutting-due-to-unconvergence----reopen----shutting-due-to-noconverged cyle. 

```
2440 Starting time step 5, stepsize 0.0106601 days, at day 1279.14/1303, date = 08-May-2001
2441  well D-3AH is added to the well container 
2442  well D-3AH was shut down due to convergence problems
2443 Well D-3AH will be shut because it fails to converge.
2444 
2445 Problem: 
2446 Problematic well(s) were shut: D-3AH(retrying timestep)
2447 
2448 
2449 Starting time step 5, stepsize 0.0106601 days, at day 1279.14/1303, date = 08-May-2001
2450  well D-3AH is added to the well container 
2451  well D-3AH was shut down due to convergence problems
2452 Well D-3AH will be shut because it fails to converge.
2453 
2454 Problem: 
2455 Problematic well(s) were shut: D-3AH(retrying timestep)
2456 
2457 
2458 Starting time step 5, stepsize 0.0106601 days, at day 1279.14/1303, date = 08-May-2001
2459  well D-3AH is added to the well container 
```

totto82:
> It is a little harsh to not open a well can do some work

We could have a default WTEST for the force-shut wells. Forcing wells that don't converge is a final attempt to get the simulation finished, but I agree that it has its drawbacks.

GitPaean:
> We could have a default WTEST for the force-shut wells. Forcing wells that don't converge is a final attempt to get the simulation finished, but I agree that it has its drawbacks.

That can be one way.  We should provide some strategy to let this well revive.

totto82:
> That can be one way. We should provide some strategy to let this well revive.

I can draft a PR in which force-shut wells are always tested at the beginning of the timestep and test it.

GitPaean:
> I can draft a PR in which force-shut wells are always tested at the beginning of the timestep and test it.

Sure. In master branch we always re-open the force-shut wells without testing at the beginning of the next time step. We can try to test it before we re-open it.  Then we can evaluate how it goes.

GitPaean:
```
2449 Starting time step 5, stepsize 0.0106601 days, at day 1279.14/1303, date = 08-May-2001
2450  well D-3AH is added to the well container 
2451  well D-3AH was shut down due to convergence problems
2452 Well D-3AH will be shut because it fails to converge.
2453 
2454 Problem: 
2455 Problematic well(s) were shut: D-3AH(retrying timestep)
2456 
2457 
2458 Starting time step 5, stepsize 0.0106601 days, at day 1279.14/1303, date = 08-May-2001
2459  well D-3AH is added to the well container 
```

Look at here, when shut the well in the time step, but when we retry the same time step, we still add the well, so I will say there is a bug to fix here. 

We shut the well because it causes the time step could not get converged due to this well, when we retry the same time step, we should not have this well back in this same time step.

totto82:
> Look at here, when shut the well in the time step, but when we retry the same time step, we still add the well, so I will say there is a bug to fix here.
> 
> We shut the well because it causes the time step could not get converged due to this well, when we retry the same time step, we should not have this well back in this same time step.

This is what this PR should fix. I assume your report is from master.

GitPaean:
> This is what this PR should fix. I assume your report is from master.

The report is from the master. I think the original design of shutting wells due to convergence cause and retrying new time steps, we should not have incorporated the forcefully shut well in the **retrying of the same** time step.  @atgeirr wrote the original one, he might have something to say.  It might be some fallout of later development.

totto82:
The fallout came when we started "Stopping" the wells instead of "Shutting" them. But as I said I have a good idea of how to solve this. I will try to provide a solution later this week.

totto82:
The above-referenced PRs solve the reported issue. It magically reopens the problematic well and squeezes more oil out of the reservoir.

totto82:
I will close this since the other PRs are suppose to give a more robust solution to the problem.

PR #6068, opened 2025-03-12
Remove assigning to self

kjetilly:
Removes a warning from a `.cpp` file. The assignment should not have been there to begin with (it probably came about during debugging)


kjetilly:
Jenkins build this please

PR #6069, opened 2025-03-12
Dont check stagnation / oscillation for stopped wells

totto82:
PR https://github.com/OPM/opm-simulators/pull/6020 gives problem for wells that are STOPed. This PR suggest a fix and restore pre #6020 runtime for NORNE_ATW2013_1A_MSW


totto82:
jenkins build this please

totto82:

totto82:

steink (approved):
Very good you found the culprit. I tried to understand why the `repeatedStagnation` would trigger for non-converged stopped wells, but was not able to see any direct reason. Maybe it's just that for stopped wells there is more cross-flow/switching flow-directions going on and that residuals may not behave _nicely_ until flow-directions have converged. This is just guessing, but may be something to investigate further if we encounter unwanted triggers of  `repeatedStagnation` also for non-stopped wells.

totto82:
Could it be possible that the stagnation_rel_tol = 1e-2 is too high? I just used what was already in the code. I can test with a higher tolerance for stopped wells.

totto82:

totto82:
@steink Could you look at the PR again. I did some more digging and I think I found the real issue that caused the problems. At some point we moved from shutting non-convergent wells during iterations to stopping them. (i.e. still solve for them but with 0 target). In the current code we stop already stopped wells which leads to the stagnation. This PR should fix this and also recover the run time for  NORNE_ATW2013_1A_MSW

totto82:

steink:
So the situation where this change triggers: a well is stopped (either before or during local iterations), but it was not able to converge with zero rate. In this case you _don't_ want to set the `solvable` to false, so some subsequent `isOperableAndSolvable` will _not_ trigger. I'm probably missing something, but I'm not able to see why this prevents stagnation?

totto82:
changed_to_stopped_this_step_ will no longer be true -> we will check well operability again -> wells oscillate between being stopped and revived during iterations (it also starts using explicit VFPs)  -> problems with Newton convergence. (I have not checked why this causes problems with the Newton convergence. But if a well is stopped I don't see the point in trying to stop it again due to convergence issues. But I may be wrong here, so lets please continue the discussion.

steink:
Sure, I'm not yet able to follow the logic entirely, but maybe related:

- A few lines further up in this function we added (https://github.com/OPM/opm-simulators/pull/5659/) a check to delay stopping wells to prevent stopped - revived - stopped oscillations, which helped a lot for NORNE_ATW2013_1A_MSW. Your situation, however, seems to be slightly different: stopped (and not converged) - revived (maybe converged?) - stopped (and not converged)? 
- Since you also mention explicit VFP: problem could also have root in bad explicit fractions due to initialization issue.

totto82:

totto82:

GitPaean:

totto82:

PR #6070, opened 2025-03-12
Apply simulator update once

vkip:


bska:
>  Still think it makes sense to only apply the update once

I've thought about this now, and I agree with you&ndash;it does indeed make sense to gather all applicable updates and apply them in bulk.  The canonical example would be if there were three separate actions triggered at one time step, one of which applied MULTX and WELPI to a set of producers, another which applied MULTY, and the third which applied MULTZ- and WELOPEN'ed a set of injectors.  In that case the current approach would end up recalculating transmissibility values three times whereas the proposed approach would do it once.

I suggest marking this PR as "ready for review".

vkip:
> > Still think it makes sense to only apply the update once
> 
> I've thought about this now, and I agree with you–it does indeed make sense to gather all applicable updates and apply them in bulk. The canonical example would be if there were three separate actions triggered at one time step, one of which applied MULTX and WELPI to a set of producers, another which applied MULTY, and the third which applied MULTZ- and WELOPEN'ed a set of injectors. In that case the current approach would end up recalculating transmissibility values three times whereas the proposed approach would do it once.
> 
> I suggest marking this PR as "ready for review".

Thanks!

bska (approved):
Very good.  This looks good to me and I'll merge into master.

PR #6075, opened 2025-03-13
Python gas-water simulator

svenn-t:
Added a Python gas-water simulator to be able to run f.ex. CO2-/H2STORE data files. Also added the option to input command-line arguments in constructor so both ways of instantiating the simulator classes have that ability.

I open this PR in draft mode because most of the code is duplicate of the PyBlackOilSimulator implementation, which is not optimal. I've made futile attempts at making a base simulator that the other Python simulators can inherit. If anybody have suggestions on how to make a base simulator (or want to take up the mantle?), I'm happy to hear them!

### DIFF
--- a/opm/simulators/flow/TTagFlowProblemGasWater.hpp
+++ b/opm/simulators/flow/TTagFlowProblemGasWater.hpp
@@ -1,4 +1,6 @@
 /*
+  Copyright 2025 Equinor ASA.
+
   This file is part of the Open Porous Media project (OPM).
 
   OPM is free software: you can redistribute it and/or modify
@@ -14,25 +16,29 @@
   You should have received a copy of the GNU General Public License
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
-#ifndef FLOW_GASWATER_HPP
-#define FLOW_GASWATER_HPP
 
-#include <opm/simulators/flow/TTagFlowProblemGasWater.hpp>
+#ifndef TTAG_FLOW_PROBLEM_GASWATER_HPP
+#define TTAG_FLOW_PROBLEM_GASWATER_HPP
 
-#include <memory>
+#include <tuple>
 
 namespace Opm {
+namespace Properties {
+namespace TTag {
 
-//! \brief Main function used in flow binary.
-int flowGasWaterMain(int argc, char** argv, bool outputCout, bool outputFiles);
+    struct FlowProblem;
 
-//! \brief Main function used in flow_gaswater binary.
-int flowGasWaterMainStandalone(int argc, char** argv);
-
-template<class TypeTag> class FlowMain;
-
-std::unique_ptr<FlowMain<Properties::TTag::FlowGasWaterProblem>>
-flowGasWaterMainInit(int argc, char** argv, bool outputCout, bool outputFiles);
+    /// Specialised type tag for simulations that can use the customised
+    /// assembly process for TPFA discretisation schemes.
+    ///
+    /// All properties are otherwise the same as for the regular
+    /// FlowProblem.
+    struct FlowGasWaterProblem {
+      using InheritsFrom = std::tuple<FlowProblem>;
+  };
 }
 
-#endif // FLOW_GASWATER_HPP
+
+} // namespace Opm::Properties::TTag
+}
+#endif // TTAG_FLOW_PROBLEM_TPFA_HPP
--- a/opm/simulators/flow/python/PyBlackOilSimulator.hpp
+++ b/opm/simulators/flow/python/PyBlackOilSimulator.hpp
@@ -55,7 +55,8 @@ public:
         std::shared_ptr<Opm::Deck> deck,
         std::shared_ptr<Opm::EclipseState> state,
         std::shared_ptr<Opm::Schedule> schedule,
-        std::shared_ptr<Opm::SummaryConfig> summary_config);
+        std::shared_ptr<Opm::SummaryConfig> summary_config,
+        const std::vector<std::string>& args);
     void advance(int report_step);
     bool checkSimulationFinished();
     int currentStep();
--- a/opm/simulators/flow/python/PyGasWaterSimulator.hpp
+++ b/opm/simulators/flow/python/PyGasWaterSimulator.hpp
@@ -1,0 +1,109 @@
+/*
+  Copyright 2020 Equinor ASA.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef OPM_PY_GAS_WATER_SIMULATOR_HEADER_INCLUDED
+#define OPM_PY_GAS_WATER_SIMULATOR_HEADER_INCLUDED
+
+#include <python/simulators/PyMainGW.hpp>
+
+#include <opm/models/utils/parametersystem.hpp>
+#include <opm/models/utils/propertysystem.hh>
+
+#include <opm/simulators/flow/FlowMain.hpp>
+#include <opm/simulators/flow/TTagFlowProblemGasWater.hpp>
+#include <opm/simulators/flow/python/PyFluidState.hpp>
+#include <opm/simulators/flow/python/PyMaterialState.hpp>
+#include <opm/simulators/flow/python/Pybind11Exporter.hpp>
+
+#include <opm/input/eclipse/EclipseState/EclipseState.hpp>
+#include <opm/input/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp>
+#include <opm/input/eclipse/Schedule/Schedule.hpp>
+
+#include <map>
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace Opm::Pybind {
+
+class PyGasWaterSimulator
+{
+private:
+    using TypeTag = Opm::Properties::TTag::FlowGasWaterProblem;
+    using Simulator = Opm::GetPropType<TypeTag, Opm::Properties::Simulator>;
+
+public:
+    PyGasWaterSimulator(const std::string& deckFilename,
+                        const std::vector<std::string>& args);
+    PyGasWaterSimulator(
+        std::shared_ptr<Opm::Deck> deck,
+        std::shared_ptr<Opm::EclipseState> state,
+        std::shared_ptr<Opm::Schedule> schedule,
+        std::shared_ptr<Opm::SummaryConfig> summary_config);
+    void advance(int report_step);
+    bool checkSimulationFinished();
+    int currentStep();
+    py::array_t<double> getFluidStateVariable(const std::string &name) const;
+    py::array_t<double> getCellVolumes();
+    double getDT();
+    py::array_t<double> getPorosity();
+    py::array_t<double> getPrimaryVariable(const std::string &variable) const;
+    py::array_t<int> getPrimaryVarMeaning(const std::string &variable) const;
+    std::map<std::string, int> getPrimaryVarMeaningMap(const std::string &variable) const;
+    int run();
+    void setPorosity(
+         py::array_t<double, py::array::c_style | py::array::forcecast> array);
+    void setPrimaryVariable(
+        const std::string &idx_name,
+        py::array_t<double,
+        py::array::c_style | py::array::forcecast> array);
+    void setupMpi(bool init_mpi, bool finalize_mpi);
+    int step();
+    int stepCleanup();
+    int stepInit();
+
+private:
+    Opm::FlowMain<TypeTag>& getFlowMain() const;
+    PyFluidState<TypeTag>& getFluidState() const;
+    PyMaterialState<TypeTag>& getMaterialState() const;
+
+    const std::string deck_filename_;
+    bool has_run_init_ = false;
+    bool has_run_cleanup_ = false;
+    bool mpi_init_ = true;
+    bool mpi_finalize_ = true;
+    //bool debug_ = false;
+    // This *must* be declared before other pointers
+    // to simulator objects. This in order to deinitialize
+    // MPI at the correct time (ie after the other objects).
+    std::unique_ptr<Opm::PyMain> main_;
+
+    std::unique_ptr<Opm::FlowMain<TypeTag>> flow_main_;
+    Simulator* simulator_;
+    std::unique_ptr<PyFluidState<TypeTag>> fluid_state_;
+    std::unique_ptr<PyMaterialState<TypeTag>> material_state_;
+    std::shared_ptr<Opm::Deck> deck_;
+    std::shared_ptr<Opm::EclipseState> eclipse_state_;
+    std::shared_ptr<Opm::Schedule> schedule_;
+    std::shared_ptr<Opm::SummaryConfig> summary_config_;
+    std::vector<std::string> args_;
+};
+
+} // namespace Opm::Pybind
+#endif // OPM_PY_BLACKOIL_SIMULATOR_HEADER_INCLUDED
--- a/opm/simulators/flow/python/PyGasWaterSimulator.hpp
+++ b/opm/simulators/flow/python/PyGasWaterSimulator.hpp
@@ -55,7 +55,8 @@ public:
         std::shared_ptr<Opm::Deck> deck,
         std::shared_ptr<Opm::EclipseState> state,
         std::shared_ptr<Opm::Schedule> schedule,
-        std::shared_ptr<Opm::SummaryConfig> summary_config);
+        std::shared_ptr<Opm::SummaryConfig> summary_config,
+        const std::vector<std::string>& args);
     void advance(int report_step);
     bool checkSimulationFinished();
     int currentStep();
@@ -92,7 +93,7 @@ private:
     // This *must* be declared before other pointers
     // to simulator objects. This in order to deinitialize
     // MPI at the correct time (ie after the other objects).
-    std::unique_ptr<Opm::PyMain> main_;
+    std::unique_ptr<Opm::PyMainGW> main_;
 
     std::unique_ptr<Opm::FlowMain<TypeTag>> flow_main_;
     Simulator* simulator_;
--- a/python/simulators/PyBlackOilSimulator.cpp
+++ b/python/simulators/PyBlackOilSimulator.cpp
@@ -63,12 +63,14 @@ PyBlackOilSimulator::PyBlackOilSimulator(
     std::shared_ptr<Opm::Deck> deck,
     std::shared_ptr<Opm::EclipseState> state,
     std::shared_ptr<Opm::Schedule> schedule,
-    std::shared_ptr<Opm::SummaryConfig> summary_config
+    std::shared_ptr<Opm::SummaryConfig> summary_config,
+    const std::vector<std::string>& args
 )
     : deck_{std::move(deck)}
     , eclipse_state_{std::move(state)}
     , schedule_{std::move(schedule)}
     , summary_config_{std::move(summary_config)}
+    , args_{args}
 {
 }
 
@@ -303,8 +305,11 @@ void export_PyBlackOilSimulator(py::module& m)
              std::shared_ptr<Opm::Deck>,
              std::shared_ptr<Opm::EclipseState>,
              std::shared_ptr<Opm::Schedule>,
-             std::shared_ptr<Opm::SummaryConfig>>(),
-             PyBlackOilSimulator_objects_constructor_docstring)
+             std::shared_ptr<Opm::SummaryConfig>,
+             const std::vector<std::string>&>(),
+             PyBlackOilSimulator_objects_constructor_docstring,
+             py::arg("Deck"), py::arg("EclipseState"), py::arg("Schedule"), py::arg("SummaryConfig"),
+             py::arg("args") = std::vector<std::string>{})
         .def("advance", &PyBlackOilSimulator::advance, advance_docstring, py::arg("report_step"))
         .def("check_simulation_finished", &PyBlackOilSimulator::checkSimulationFinished,
              checkSimulationFinished_docstring)
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -888,6 +888,7 @@ list (APPEND PUBLIC_HEADER_FILES
   opm/simulators/flow/SolutionContainers.hpp
   opm/simulators/flow/SubDomain.hpp
   opm/simulators/flow/TTagFlowProblemTPFA.hpp
+  opm/simulators/flow/TTagFlowProblemGasWater.hpp
   opm/simulators/flow/TracerContainer.hpp
   opm/simulators/flow/TracerModel.hpp
   opm/simulators/flow/Transmissibility.hpp
--- a/flow/flow_gaswater.cpp
+++ b/flow/flow_gaswater.cpp
@@ -33,11 +33,6 @@
 
 namespace Opm {
 namespace Properties {
-namespace TTag {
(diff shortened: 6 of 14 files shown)


hakonhagland:
> most of the code is duplicate of the PyBlackOilSimulator implementation, which is not optimal.  I've made futile attempts at making a base simulator that the other Python simulators can inherit.

@svenn-t Interesting. It is not immediately clear to me what is the difficulty with making a base simulator class. Can you elaborate on that?

svenn-t:
I made an attempt here: 45be9b86ed7dd094278c384239993f3816df41c7 which compiled but gave "undefined symbol" errors when I imported any of the "BlackOilSimulator" or "GasWaterSimulator" in Python. It might have something to do with how I made the base simulator but I also have next to no experience with pybind and it could be some issues with how I exported the classes.

hakonhagland:
> which compiled but gave "undefined symbol" errors

@svenn-t I think we need to instantiate the templated base class explicitly for the two template values we are using, see this commit: https://github.com/hakonhagland/opm-simulators-1/commit/b551864f330c91972dacb6004e2b0355e5d8418b. This fixed the undefined symbol error for me.

svenn-t:
> > which compiled but gave "undefined symbol" errors
> 
> @svenn-t I think we need to instantiate the templated base class explicitly for the two template values we are using, see this commit: [hakonhagland@b551864](https://github.com/hakonhagland/opm-simulators-1/commit/b551864f330c91972dacb6004e2b0355e5d8418b). This fixed the undefined symbol error for me.

You are correct, this fixed the "undefined symbol" errors for me as well. I did get some other issues when running examples using the python classes that I need to look into before updating this PR. But thanks for fixing the immediate errors!

hakonhagland:
> I did get some other issues when running examples using the python classes that I need to look into before updating this PR

@svenn-t Yes, I am getting this error when I run the unit tests: `The MPI_Comm_free() function was called after MPI_FINALIZE was invoked.`. I will have a look at why this happens.

hakonhagland:
> The MPI_Comm_free() function was called after MPI_FINALIZE was invoked.

Seems like this happens because the derived class `PyBlackOilSimulator` constructor is called before the base class `PyBaseSimulator` destructor, and the derived class destructor implicitly frees its `Opm::Main` pointer which will cause `MPI_Finalize()` to be called, then the base class destructor is called which will implicitly free its `Opm::FlowMain` pointer, which will implicitly destroy the communicator, see:
https://github.com/OPM/opm-simulators/blob/abfe4223d2539bfacd6e7174136db3991752042d/opm/simulators/wells/ParallelWellInfo.hpp#L339

hakonhagland:
> I couldn't figure out how to have main_ in the base class

@svenn-t The following seems to work for me: https://github.com/hakonhagland/opm-simulators-1/commit/36ccf4ac8298225f32422b2c3144193eacb0777c

svenn-t:
> @svenn-t The following seems to work for me: https://github.com/hakonhagland/opm-simulators-1/commit/36ccf4ac8298225f32422b2c3144193eacb0777c

Indeed, this worked for me as well. Thanks a lot for the fixes! Unfortunately, I get a lot of "violates the C++ One Definition Rule" errors when I compile, and running cases with the GasWaterSimulator crashes. Since there are so many of the ODR errors I suspect there are some fundamental issue like, e.g. the BlackOilTwoPhaseIndices for the gas-water simulator have not been properly included or some header file is missing. I'll look into these issues.

hakonhagland:
> running cases with the GasWaterSimulator  crashes

@svenn-t Do you have any test cases for the gas water simulator? Then I can also have a look at it

svenn-t:
>  @svenn-t Do you have any test cases for the gas water simulator? Then I can also have a look at it

Yes, I have a modified SPE1CASE1 file here: 
[SPE1CASE1_GW.zip](https://github.com/user-attachments/files/19318203/SPE1CASE1_GW.zip)

svenn-t:
> I do not get any error when I compile, can you give more details? 

I see now that I get warnings and not errors. I have the build output here: 
[Build_log.zip](https://github.com/user-attachments/files/19321052/Build_log.zip)
if you are interested, but essentially there are a lot of statements like these:
```text
[build] /home/AD.NORCERESEARCH.NO/svtv/simulators/mupsi/opm-simulators/opm/simulators/flow/BlackoilModel.hpp:61:7: warning: type ‘struct BlackoilModel’ violates the C++ One Definition Rule [-Wodr]
[build]    61 | class BlackoilModel
[build]       |       ^
[build] /home/AD.NORCERESEARCH.NO/svtv/simulators/mupsi/opm-simulators/opm/simulators/flow/BlackoilModel.hpp:61: note: a different type is defined in another translation unit
[build]    61 | class BlackoilModel
[build]       | 
```
and
```text
[build] /home/AD.NORCERESEARCH.NO/svtv/simulators/mupsi/opm-simulators/opm/simulators/wells/StandardWell_impl.hpp:2493: warning: type of ‘computeCurrentWellRates’ does not match original declaration [-Wlto-type-mismatch]
[build]  2493 |     StandardWell<TypeTag>::
[build]       | 
[build] /home/AD.NORCERESEARCH.NO/svtv/simulators/mupsi/opm-simulators/opm/simulators/wells/StandardWell_impl.hpp:2493:5: note: ‘computeCurrentWellRates’ was previously declared here
[build]  2493 |     StandardWell<TypeTag>::
[build]       |     ^
[build] /home/AD.NORCERESEARCH.NO/svtv/simulators/mupsi/opm-simulators/opm/simulators/wells/StandardWell_impl.hpp:2493:5: note: code may be misoptimized unless ‘-fno-strict-aliasing’ is used
```
I just do a simple `make all` (with VS Code cmake extension, but nevertheless) and `make install` afterwards.

> The step_init() above runs fine, but then sim.step() crashes with:

I get the same crash report as you when I run the case as well.

hakonhagland:
> I see now that I get warnings

@svenn-t Interesting. I do not get any warnings. Here is my build log [build.zip](https://github.com/user-attachments/files/19323675/build.zip). I compiled like this from the terminal:
```
$ cmake -DCMAKE_BUILD_TYPE=Debug -DOPM_ENABLE_PYTHON=ON -DOPM_ENABLE_EMBEDDED_PYTHON=ON  ..
$ make -j4
```

> Assertion `this->num_components_ == numWellConservationEq' failed.

This is from https://github.com/OPM/opm-simulators/blob/78a2265717a79460ccce3e9f5dfe2bc31a20061e/opm/simulators/wells/StandardWell_impl.hpp#L66 and in gdb I can see that `this->num_components_` is 2, whereas `numWellConservationEq` is 3.

svenn-t:
> @svenn-t Interesting. I do not get any warnings. Here is my build log [build.zip](https://github.com/user-attachments/files/19323675/build.zip). I compiled like this from the terminal

When i build with Debug variant, I also get no compilation warnings. The build log I attached was with Release variant.

>  and in gdb I can see that this->num_components_ is 2, whereas numWellConservationEq is 3.

I suspect that the part in PyGasWaterSimulator.cpp where we set BlackOilTwoPhaseIndices is not done properly or even set at all. I tried to move that part to PyBaseSimulator.cpp, but I get the same crash when running the test case.

hakonhagland:
> The build log I attached was with Release variant.

@svenn-t I see, then I am able to reproduce the warnings also. There are many, the first one I get is:

```
/home/hakon/test/opm8/opm/opm-simulators/opm/simulators/wells/WellInterface.hpp:75:7: warning: type ‘struct WellInterface’ violates the C++ One Definition Rule [-Wodr]
   75 | class WellInterface : public WellInterfaceIndices<GetPropType<TypeTag, Properties::FluidSystem>,
      |       ^
/home/hakon/test/opm8/opm/opm-simulators/opm/simulators/wells/WellInterface.hpp:75:7: note: a type with the same name but different base type is defined in another translation unit
   75 | class WellInterface : public WellInterfaceIndices<GetPropType<TypeTag, Properties::FluidSystem>,
      |       ^
/home/hakon/test/opm8/opm/opm-simulators/opm/simulators/wells/WellInterfaceIndices.hpp:33:7: note: type name ‘Opm::WellInterfaceIndices<Opm::BlackOilFluidSystem<double, Opm::BlackOilDefaultIndexTraits, Opm::VectorWithDefaultAllocator, std::shared_ptr>, Opm::BlackOilTwoPhaseIndices<0u, 0u, 0u, 0u, false, false, 0u, 0u, 0u> >’ should match type name ‘Opm::WellInterfaceIndices<Opm::BlackOilFluidSystem<double, Opm::BlackOilDefaultIndexTraits, Opm::VectorWithDefaultAllocator, std::shared_ptr>, Opm::BlackOilIndices<0u, 0u, 0u, 0u, false, false, 0u, 0u> >’
   33 | class WellInterfaceIndices : public WellInterfaceFluidSystem<FluidSystem>
      |       ^
```
I believe this happens when building the Python shared library `simulators.cpython-312-x86_64-linux-gnu.so` since we are linking it with both `PyBlackOilSimulator.cpp.o` and `PyGasWaterSimulator.cpp.o` and the first uses `Opm::BlackOilIndices` and the latter uses `Opm::BlackOilTwoPhaseIndices`. So both variants of `WellInterface` gets linked into the shared object. An idea could be to create two separate shared objects, but there might be other solutions that lets us keep everything in a single shared object.

hakonhagland:
> An idea could be to create two separate shared objects

@svenn-t Here is an example of a splitting into two separate modules: https://github.com/hakonhagland/opm-simulators-1/commit/1ff143a8cf78a2a4412b4e623ee1f768848280cd. This compiles without ODR violations for me.

hakonhagland:
If we are going to split the python module into two, one issue is that each of the two python modules will duplicate much of opm-simulators and opm-common static libraries. To avoid this, we can compile opm-common, opm-grid and opm-simulators as shared libraries (option `-DBUILD_SHARED_LIBS=ON`) then the code will not be duplicated in the python libraries.

hakonhagland:
> Here is an example of a splitting into two separate modules: hakonhagland@1ff143a.

To also make the unit tests pass I needed to make some further changes to the pybind11 module names and python import statements, see: https://github.com/hakonhagland/opm-simulators-1/commit/6d24fe1091988824c05471e5d35017c5f2dd3d10

svenn-t:
> @svenn-t Here is an example of a splitting into two separate modules: https://github.com/hakonhagland/opm-simulators-1/commit/1ff143a8cf78a2a4412b4e623ee1f768848280cd. This compiles without ODR violations for me.

> To also make the unit tests pass I needed to make some further changes to the pybind11 module names and python import statements, see: https://github.com/hakonhagland/opm-simulators-1/commit/6d24fe1091988824c05471e5d35017c5f2dd3d10

With the new commits, I also do not get the ODR warnings, and running both the gas-water and blackoil versions of the SPE1CASE1 cases, I do not get any crashes. Thanks a lot for making these fixes and getting the simulators to work!

> If we are going to split the python module into two, one issue is that each of the two python modules will duplicate much of opm-simulators and opm-common static libraries. To avoid this, we can compile opm-common, opm-grid and opm-simulators as shared libraries (option -DBUILD_SHARED_LIBS=ON) then the code will not be duplicated in the python libraries.

I admit this is beyond my knowledge with pybind, cmake, etc. Are there any issues with doing this?

I will update this PR with the new commits since it seems to be compiling and running working correctly.

akva2:
relying on dynamic linking to opm libraries is problematic for pip deployment. i much prefer if you track down the odr violation instead. something is instanced in the python module that should not be instanced in the python module.

hakonhagland:
> relying on dynamic linking to opm libraries is problematic for pip deployment.

@akva2 Is it the auditwheel that is the problem? In principle we should be able to install all dependencies in a manylinux container, right? However, I assume it can be difficult find the correct versions of those that fit together?

 > i much prefer if you track down the odr violation instead

@akva2 It is quite a lot of the ODR violations, I think it can become a difficult task. As a workaround we could build the two python modules with the static opm libraries (i.e. include them into the shared objects). The drawback is then only the code duplication in the two modules (that is the size of the modules will be larger than strictly necessary)

akva2:
while we can in principle install shared dependencies with the python module, it is opening a can of worms i'm very reluctant of taking aboard. the problem is that in the wild west of python package management, there are ways users can screw themself over and end up loading wrong versions of the shared libraries from system paths and not the shipped libs, in particular if the don't use venvs and such.

while i'm usually a fan of shared linking, in this case i'll take larger modules any day to avoid the problems so if there is no other way that is what i'd do.

hakonhagland:
> there are ways users can screw themself over and end up loading wrong versions of the shared libraries from system paths and not the shipped libs, in particular if the don't use venvs and such.

@akva2 Interesting. By looking at the [source](https://github.com/pypa/auditwheel/blob/fff47fdb64567ab88374896edfaf3e6d842c946c/src/auditwheel/patcher.py#L65) it seems `auditwheel` is adjusting the rpath so the wheel's internally shipped shared objects should then take precedence over any equivalent libraries from the user's system path. Also, it looks like it renames the libraries, see [repair.py](https://github.com/pypa/auditwheel/blob/fff47fdb64567ab88374896edfaf3e6d842c946c/src/auditwheel/repair.py#L138), so a conflict in names should then be even less likely, right?

It would be interesting to test this, can I just run `docker build .` from the `python` directory?

akva2:
right, so there are new hacks in the tools to try to fix the situation then. my trust in python packaging tools is still exactly zero though, so i'm sure there are some tool that screws it up somehow, somewhere in the world..
anyways,

basically yes, invocation on jenkins is
```
DOCKER_BUILDKIT=1 docker build -t manylinux_2_28_opm:built --build-arg build_jobs=16 --build-arg version_tag="$2" --build-arg version="$3" . -f Dockerfile --output $1/wheelhouse
```

hakonhagland:
> DOCKER_BUILDKIT=1 docker build -t manylinux_2_28_opm:built --build-arg build_jobs=16 --build-arg version_tag="$2" --build-arg version="$3" . -f Dockerfile --output $1/wheelhouse

@akva2 Yes this worked for static opm libraries after a small change to the `generate-pypi-package.sh` script, see https://github.com/hakonhagland/opm-simulators-1/commit/1b690f91cb24ee08a1a2887f70f4f97b02622d73. However, adding `-DBUILD_SHARED_LIBS=ON` to the cmake flags and trying to build shared opm libraries failed. It seems like the first problem occurs when trying to build a shared `libopmgrid.so` due to only static versions of the `lapack.a` is available in the manylinux container, see https://github.com/OPM/opm-simulators/blob/44eb826766f39b1928e34abcb82c91db467a61c4/python/setup-docker-image.sh#L11

I will try to replace these packages with ones including the shared versions and see how far I get.

hakonhagland:
> I will try to replace these packages with ones including the shared versions and see how far I get.

It seems to work after these changes: https://github.com/hakonhagland/opm-simulators-1/commit/84a9041b024914c1b31f1ee4e8ac95a107bfac3b. I will try to install the generated wheels in some containers to check if they also installs successfully.

hakonhagland:
I am no longer able to build in the container since it now uses cmake 4.0, see: https://github.com/OPM/opm-simulators/issues/6122

python/test/test_primary_variables.py
```diff
@@ -41,3 +50,33 @@ def test_all(self):
...
+            gas_meaning = sim.get_primary_variable_meaning(
+                variable='gas')
+            gas_meaning_map = sim.get_primary_variable_meaning_map(
+                variable='gas')
+            self.assertEqual(gas_meaning[0], gas_meaning_map["Disabled"])
+            brine_meaning = sim.get_primary_variable_meaning(
+                variable='brine')
+            brine_meaning_map = sim.get_primary_variable_meaning_map(
+                variable='brine')
+            self.assertEqual(brine_meaning[0], brine_meaning_map["Disabled"])
```
akva2: eol at eof

opm/simulators/flow/TTagFlowProblemGasWater.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include <tuple>
+
+namespace Opm {
+namespace Properties {
+namespace TTag {
+
+    struct FlowProblem;
+
+    /// Specialised type tag for simulations that can use the customised
```
akva2: this doxy makes no sense, copy-pasta.

opm/simulators/flow/TTagFlowProblemGasWater.hpp
```diff
@@ -0,0 +1,44 @@
...
+    ///
+    /// All properties are otherwise the same as for the regular
+    /// FlowProblem.
+    struct FlowGasWaterProblem {
+      using InheritsFrom = std::tuple<FlowProblem>;
+  };
+}
+
+
+} // namespace Opm::Properties::TTag
```
akva2, line 38: one line before first brace, one line before endif.

opm/simulators/flow/python/PyBaseSimulator.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+template<class TypeTag>
+class PyBaseSimulator
+{
+private:
+    using Simulator = Opm::GetPropType<TypeTag, Opm::Properties::Simulator>;
+
+public:
+    PyBaseSimulator(const std::string& deckFilename,
+        const std::vector<std::string>& args);
```
akva2: wrong indent

opm/simulators/flow/python/PyBaseSimulator.hpp
```diff
@@ -0,0 +1,115 @@
...
+    PyBaseSimulator(
```
akva2: clang-format has barfed all over this code I see. I'll not comment on the others..

opm/simulators/flow/python/PyBaseSimulator.hpp
```diff
@@ -0,0 +1,115 @@
...
+    void setPrimaryVariable(
+        const std::string &idx_name,
+        py::array_t<double,
+        py::array::c_style | py::array::forcecast> array);
+    void setupMpi(bool init_mpi, bool finalize_mpi);
+    int step();
+    int stepCleanup();
+    int stepInit();
+
+    // void initMain();
```
akva2: no new commented lines

opm/simulators/flow/python/PyBaseSimulator_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+{
+    return getFlowMain().getSimTimer()->currentStepNum();
+    // NOTE: this->simulator_->episodeIndex() would also return the current
+    // report step number, but this number is always delayed by 1 step relative
+    // to this->flow_main_->getSimTimer()->currentStepNum()
+    // See details in runStep() in file SimulatorFullyImplicitBlackoilEbos.hpp
+}
+
+template<class TypeTag>
+py::array_t<double> PyBaseSimulator<TypeTag>::getCellVolumes() {
```
akva2: brace down

opm/simulators/flow/python/PyBaseSimulator_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+    auto vector = getMaterialState().getCellVolumes();
+    return py::array(vector.size(), vector.data());
+}
+
+template<class TypeTag>
+double PyBaseSimulator<TypeTag>::getDT() {
```
akva2: brace down

opm/simulators/flow/python/PyBaseSimulator_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+template<class TypeTag>
+std::map<std::string, int>
+PyBaseSimulator<TypeTag>::
+getPrimaryVarMeaningMap(const std::string &variable) const
+{
+
+    return getFluidState().getPrimaryVarMeaningMap(variable);
+}
+
+// template<class TypeTag>
```
akva2: no new commented code

opm/simulators/flow/python/PyBaseSimulator_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+    if (!this->has_run_init_) {
+        throw std::logic_error("step() called before step_init()");
+    }
+    if (this->has_run_cleanup_) {
+        throw std::logic_error("step() called after step_cleanup().");
+    }
+    if(checkSimulationFinished()) {
+        throw std::logic_error("step() called, but simulation is done");
+    }
+    //if (this->debug_)
```
akva2: ...

opm/simulators/flow/python/PyBaseSimulator_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+int PyBaseSimulator<TypeTag>::stepCleanup()
+{
+    this->has_run_cleanup_ = true;
+    return getFlowMain().executeStepsCleanup();
+}
+
+template<class TypeTag>
+int PyBaseSimulator<TypeTag>::stepInit()
+{
+
```
akva2: drop newline

opm/simulators/flow/python/PyBlackOilSimulator.hpp
```diff
@@ -20,89 +20,33 @@
...
-    bool mpi_init_ = true;
-    bool mpi_finalize_ = true;
-    //bool debug_ = false;
-    // This *must* be declared before other pointers
-    // to simulator objects. This in order to deinitialize
-    // MPI at the correct time (ie after the other objects).
-    std::unique_ptr<Opm::PyMain> main_;
+    PyBlackOilSimulator(const std::string& deck_filename,
+                        const std::vector<std::string>& args)
+    : BaseType(deck_filename, args)
```
akva2: indent

opm/simulators/flow/python/PyBlackOilSimulator.hpp
```diff
@@ -20,89 +20,33 @@
...
+    {}
+
+    PyBlackOilSimulator(std::shared_ptr<Opm::Deck> deck,
+                        std::shared_ptr<Opm::EclipseState> state,
+                        std::shared_ptr<Opm::Schedule> schedule,
+                        std::shared_ptr<Opm::SummaryConfig> summary_config,
+                        const std::vector<std::string>& args)
+    : BaseType(deck, state, schedule, summary_config, args)
```
akva2: indent

opm/simulators/flow/python/PyGasWaterSimulator.hpp
```diff
@@ -0,0 +1,52 @@
...
+                        std::shared_ptr<Opm::EclipseState> state,
+                        std::shared_ptr<Opm::Schedule> schedule,
+                        std::shared_ptr<Opm::SummaryConfig> summary_config,
+                        const std::vector<std::string>& args)
+    : BaseType(deck, state, schedule, summary_config, args)
+    {}
+
+};
+
+} // namespace Opm::Pybind
```
akva2, line 51: newline before endif

python/generate-pypi-package.sh
```diff
@@ -56,7 +56,7 @@ do
     cmake -DPYTHON_EXECUTABLE=${python_versions[$tag]} -DWITH_NATIVE=0 -DBoost_USE_STATIC_LIBS=$static \
     -DOPM_ENABLE_PYTHON=ON -DOPM_PYTHON_PACKAGE_VERSION_TAG=${VERSION_TAG} -DBLA_STATIC=$static -DBLAS_LIBRARIES=/usr/lib64/libblas.$libext -DSUITESPARSE_USE_STATIC=$static -DCMAKE_DISABLE_FIND_PACKAGE_QuadMath=1 -DBUILD_SHARED_LIBS=$shared ..
 
-    cmake --build . --target opmcommon_python simulators --parallel ${BUILD_JOBS}
+    cmake --build . --target opmcommon_python BlackOil GasWater --parallel ${BUILD_JOBS}
```
akva2: This file is in conflict, but anyways. I think we should learn from prior mistakes and make it easier to identify python targets. Add a target called python_simulator_modules or such, and make it depend on BlackOil and GasWater instead. Then if new targets are added in the future we don't have to update all over the place.

python/simulators/PyBlackOilSimulator.cpp
```diff
@@ -34,267 +27,41 @@
...
-        return exit_code;
-    }
-}
+namespace Opm {
 
-// Private methods alphabetically sorted
-// ------------------------------------
+template class PyMain<Opm::Properties::TTag::FlowProblemTPFA>;
+template std::unique_ptr<FlowMain<Opm::Properties::TTag::FlowProblemTPFA>>
+  flowMainInit<Opm::Properties::TTag::FlowProblemTPFA>(
```
akva2: <-

akva2 (commented):
Please rebase and address my comments so we can get this in.

akva2:
jenkins build this please

akva2:
jenkins build this please

PR #6077, opened 2025-03-13
Prevent some nan creations during output generation.

blattms:
If rate that is the nominator is zero just store zero instead of dividing it by another rater that might also be zero.

Makes searching for other nans more easy.


bska (approved):
Thanks a lot!  I'll merge into master.

Issue #6073, opened 2025-03-12
Initialization by Enumeration not working properly for some combinations of SGAS and PRESSURE


stimyd:
Hello,

I have observed a few cases in my past use of Flow where initialization by "Enumeration" (i.e., initializing the model using the keywords PRESSURE, SGAS, SWAT, RV and RS) can sometimes result in a different initial state of the model than specified. 

I've attached two file decks. Same input files, with one deck having the output files from an ECL100 2012.1 run, and the other from a Flow 2023.04 run. 

In summary, OPM Flow initializes the model with SOIL = 0.4, when in fact, it should be 0.0. ECL100 initializes the model correctly. The wrong initialization subsequently causes convergence issues immediately. If PRESSURE is reduced to 6400 psia, then everything runs fine, but the initialization is still slightly wrong with SOIL close to, but not quite equal to 0.0. You can also observe that BPR of cell 1 1 1 seems to be increasing with time in Flow. A material-balance calculation issue perhaps?

Any thoughts on why this might be happening, and/or suggested solutions to the problem? 

[ecl100.zip](https://github.com/user-attachments/files/19212151/ecl100.zip)
[opm-flow.zip](https://github.com/user-attachments/files/19212154/opm-flow.zip)

blattms:
I have not looked into this. It seems like you are using OPM flow 2023.04, which is rather old. The current version is 2024.10. Is there for a reason for using that version? 

I am not saying that this is fixed now but there at least is a chance. We usually do quite a few bug fixes during 1.5 years.

daavid00:
I have run the case with master from today and the initialisation for SOIL looks fine, i.e., if run with the flag `--enable-vtk-output=1` (to look directly to the arrays values in the .vtu using a text editor) and adding: 

> RPTRST
> 'BASIC=2'/

in the `SOLUTION` section, then the initial `SOIL` is practically 0 (1.11022e-16).

stimyd:
Hi

Thanks for getting back so quickly! 

I've have now tried running the case with Flow 2024.10 and the issue is still there (see attached .zip -> BSOIL is still 0.4 in .RSM, and the simulation continues to fail if I run it longer than a few seconds).

If it works on master, is there a potential fix to this coming with 2025.04?

![Image](https://github.com/user-attachments/assets/d2d80591-f721-40f5-b6c6-8dd8325c8a47)

[flow-202410.zip](https://github.com/user-attachments/files/19227020/flow-202410.zip)

[...]

PR #6079, opened 2025-03-13
Set SOIL to 0 when readExplicitInitialCondition and 1-SGAS-SWAT=0

daavid00:
This PR aims to fix the issue raised in https://github.com/OPM/opm-simulators/issues/6073 (with current master SOIL>0 for the deck in the issue given explicit the initial condition (SGAS=0.7535, SWAT=0.2465)).

### DIFF
--- a/opm/simulators/flow/FlowProblemBlackoil.hpp
+++ b/opm/simulators/flow/FlowProblemBlackoil.hpp
@@ -1354,11 +1354,15 @@ protected:
                     dofFluidState.setSaturation(FluidSystem::gasPhaseIdx,
                                                 gasSaturationData[dofIdx]);
             }
-            if (FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx))
-                dofFluidState.setSaturation(FluidSystem::oilPhaseIdx,
-                                            1.0
-                                            - waterSaturationData[dofIdx]
-                                            - gasSaturationData[dofIdx]);
+            if (FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx)) {
+                const Scalar soil = 1.0 - waterSaturationData[dofIdx] - gasSaturationData[dofIdx];
+                if (soil < smallSaturationTolerance_) {
+                    dofFluidState.setSaturation(FluidSystem::oilPhaseIdx, 0.0);
+                }
+                else {
+                    dofFluidState.setSaturation(FluidSystem::oilPhaseIdx, soil);
+                }
+            }
 
             //////
             // set phase pressures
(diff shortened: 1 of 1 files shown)


GitPaean:
I did not check #6073 in details, but I could not tell what is the problem with the following code with (SGAS=0.7535, SWAT=0.2465)). It will also set SOIL to be 0.? Can you explain some? Thanks. 

```c++
            if (FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx))
                dofFluidState.setSaturation(FluidSystem::oilPhaseIdx,
                                            1.0
                                            - waterSaturationData[dofIdx]
                                            - gasSaturationData[dofIdx]);
```

daavid00:
Without this PR and for this case the SOIL is set to 1.11E-16 which results in [oilPresent=True](https://github.com/OPM/opm-simulators/blob/master/opm/models/blackoil/blackoilprimaryvariables.hh#L392) which leads to different setting up of primary variables (e.g.,  [primaryVarsMeaningGas_ = GasMeaning::Sg](https://github.com/OPM/opm-simulators/blob/master/opm/models/blackoil/blackoilprimaryvariables.hh#L438)). With this PR, then the primary variables are different (e.g., [primaryVarsMeaningGas_ = GasMeaning::Rv](https://github.com/OPM/opm-simulators/blob/master/opm/models/blackoil/blackoilprimaryvariables.hh#L442)) which prevents the SOIL saturation to appear as reported in the issue.

opm/simulators/flow/FlowProblemBlackoil.hpp
```diff
@@ -1354,11 +1354,15 @@ class FlowProblemBlackoil : public FlowProblem<TypeTag>
...
             }
-            if (FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx))
-                dofFluidState.setSaturation(FluidSystem::oilPhaseIdx,
-                                            1.0
-                                            - waterSaturationData[dofIdx]
-                                            - gasSaturationData[dofIdx]);
+            if (FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx)) {
+                const Scalar soil = 1.0 - waterSaturationData[dofIdx] - gasSaturationData[dofIdx];
+                if (soil < smallSaturationTolerance_) {
+                    dofFluidState.setSaturation(FluidSystem::oilPhaseIdx, 0.0);
```
totto82, line 1360: Instead for adding a tolerance check for soil here you could consider checking if rv < rvSat? In the end I dont think it should lead to a big differences from what you suggest.

PR #6078, opened 2025-03-13
fixed: clear out events before processing actions

akva2:
these have already been processed at start of report step and should not be executed again

Downstream of https://github.com/OPM/opm-common/pull/4520

PR #6080, opened 2025-03-14
WellModel: make a copy of events at start of time step

akva2:
these can be modified during sub-stepping, and the wcycle code needs the original events for decision making

Noticed while working on https://github.com/OPM/opm-simulators/pull/6078 
This should be processed first.


akva2:
jenkins build this please

bska (approved):
Right, this makes sense to me.  Thanks a lot for the fix!  I'll merge into master.

PR #6081, opened 2025-03-14
Dont resolve well equation when computing initial guess for thp in  updateWellStateWithTarget

totto82:
Resolving the equation in updateWellStateWithTarget may lead to a long simulation time for problematic wells. I will put this in draft mode as it would need more testing.


totto82:

totto82:

PR #6084, opened 2025-03-14
Don't Store EclipseState Reference in Convergence Output Object

bska:
The EclipseState reference is used only for `startThread()`, so we might as well request that the caller pass an object when calling the `startThread()` function instead of keeping a reference to the object throughout the simulation run.

While here, also rename the `already_reported_steps_` to camel case.

### DIFF
--- a/opm/simulators/flow/SimulatorConvergenceOutput.cpp
+++ b/opm/simulators/flow/SimulatorConvergenceOutput.cpp
@@ -18,7 +18,9 @@
   You should have received a copy of the GNU General Public License
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
+
 #include <config.h>
+
 #include <opm/simulators/flow/SimulatorConvergenceOutput.hpp>
 
 #include <opm/input/eclipse/EclipseState/EclipseState.hpp>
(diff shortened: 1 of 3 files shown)


bska:
jenkins build this please

opm/simulators/flow/SimulatorConvergenceOutput.cpp
```diff
@@ -42,14 +50,14 @@ startThread(std::string_view convOutputOptions,
     }
 
     auto convertTime = ConvergenceOutputThread::ConvertToTimeUnits {
-        [usys = eclState_.getUnits()](const double time)
+        [usys = eclState.getUnits()](const double time)
```
akva2, line 53: copy necessary?

akva2 (approved):

opm/simulators/flow/SimulatorConvergenceOutput.cpp
bska: > copy necessary?

We do indeed need a copy here, because the ownership of the `convertTime()` utility function is [transferred](https://github.com/OPM/opm-simulators/blob/c800c89674d996346e2066313bf3d2ff6383cca6/opm/simulators/flow/SimulatorConvergenceOutput.cpp#L54) into the convergence output thread and therefore outlives the `startThread()` function.

If you'd like me to I can write this more explicitly as
```C++
auto usysCopy = eclState.getUnits();
auto convertTime = [usys = std::move(usysCopy)](...);
```
instead.

akva2: not at all, using capture to copy is fine, just wanted to confirm.

bska:
jenkins build this please

PR #6086, opened 2025-03-14
Enqueue Convergence Output Only if Needed

bska:
While here, also add Doxygen style documentation to the data members and member functions of class `SimulatorConvergenceOutput`.

### DIFF
--- a/opm/simulators/flow/SimulatorConvergenceOutput.hpp
+++ b/opm/simulators/flow/SimulatorConvergenceOutput.hpp
@@ -32,19 +32,20 @@
 
 namespace Opm {
 
-class EclipseState;
-struct StepReport;
+    class EclipseState;
+    struct StepReport;
+
+} // namespace Opm
+
+namespace Opm {
 
 /// Class handling convergence history output for a simulator.
 class SimulatorConvergenceOutput
 {
 public:
-    explicit SimulatorConvergenceOutput(const EclipseState& eclState)
-        : eclState_(eclState)
-    {}
-
-    void startThread(std::string_view convOutputOptions,
-                     std::string_view optionName,
+    void startThread(const EclipseState& eclState,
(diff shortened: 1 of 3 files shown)


bska:
I'm creating this PR in draft mode because it depends on, and contains, the earlier PR #6085.  I will keep the PR in a draft state until such time as it is ready for review.

bska:
> I'm creating this PR in draft mode because it depends on, and contains, the earlier PR #6085.

The earlier PR was merged into the master branch.  I'm marking this PR as "ready for review" and I'm running a build check.

bska:
jenkins build this please

bska:
jenkins build this please

akva2 (approved):
Output as a verb reads a little awkward but it is of course correct.

bska:
> Output as a verb reads a little awkward

I suppose we can say "we've already written..." instead.  Would that be an improvement?

akva2:
yes i think using written is an improvement.

bska:
> i think using "written" is an improvement.

Cool&ndash;I've pushed an update to that effect.

bska:
jenkins build this please

opm/simulators/flow/SimulatorConvergenceOutput.hpp
```diff
@@ -43,20 +43,69 @@ namespace Opm {
...
                      std::string_view    convOutputOptions,
                      std::string_view    optionName,
                      ConvergenceOutputThread::ComponentToPhaseName getPhaseName);
 
+    /// Create convergence output requests.
+    ///
+    /// Must not be called before startThread() or after endThread().
+    ///
+    /// Only those reports which have not previously been emitted will be
+    /// output, and only if the run actually requests convergence output at
```
akva2, line 71: here as well

bska:

PR #6088, opened 2025-03-18
Added set -e to hipify_file.sh

kjetilly:
If something fails to hipify (eg. because `hipify_perl` is misisng), the `hipify_file.sh` script will not fail but simply carry on. This PR adds the `set -e` command at the start of the script to make it return a non-zero return value if any command fails.


kjetilly:

kjetilly:

PR #6090, opened 2025-03-18
PRT: Use {fmt} to format injection log

akva2:
I'm working on adding support for RPTSCHED WELLS=2.

This appends connection level data to the log. In the process I realized two issues with current code;
1) the table format is dependent on group/well names being left aligned. This to properly fit the connection data and giving a sense of 'this is under this entry'.
2) the ijk-indices were misaligned

As fixing things like this with non-programmed table layouts is a major pain in the behind, I decided to fix the problem properly. In the process I also found various inconsistencies in the existing code, wrong centerings and such.

This is the first part of this work, starting with the injection log.

Before
```
:    INJ:   1,    1    :
:  BLOCK:   1,  1,  3 <-- this is the new data
```
After
```
:INJ   :   1,  1     :
: BLOCK:   1,  1,  3 :
```


akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

opm/simulators/flow/LogOutputHelper.cpp
```diff
@@ -617,30 +718,32 @@ outputInjectionReportRecord_(const std::vector<Scalar>& wellInj,
...
     if (! isWellRecord) {
-        ss << std::setw(11) << "" << ':';
+        ss << fmt::format("{:11}:", "");
     } else {
-        ss << std::setw( 5) << wellInj[WellInjDataType::WellLocationi] << ','
-           << std::setw( 5) << wellInj[WellInjDataType::WellLocationj] << ':';
+        ss << fmt::format("{:>3},{:>3} {:>3}:",
+                          wellInj[WellInjDataType::WellLocationi],
+                          wellInj[WellInjDataType::WellLocationj],
+                          "");
```
bska, line 729: So if we need three digits for both I and J, then we'll get output of the form
```
123,456
```
Is that intentional?

opm/simulators/flow/LogOutputHelper.cpp
```diff
@@ -580,32 +606,107 @@ outputCumulativeReportRecord_(const std::vector<Scalar>& wellCum,
...
 void LogOutputHelper<Scalar>::beginInjectionReport_() const
 {
-    const auto unitType = this->eclState_.getUnits().getType();
+    const auto& units = this->eclState_.getUnits();
+    const auto widths = std::array{8, 11, 6, 6, 6, 11, 11, 11, 11, 8, 8};
+
+    using namespace std::string_view_literals;
+
+    auto oil_rate =
+        [&units]()
```
bska, line 615: Might as well capture `units.getType()` here, e.g.,
```C++
auto oil_rate = [utype = units.getType()]()
{
    switch (utype) {
    // ...
    }
};
```
since you're not really using a lot of the `units` object internally.

bska (approved):
Looks good to me.  Thanks a lot for the clean-up.

opm/simulators/flow/LogOutputHelper.cpp
akva2: yes, that is the intent. it is dictated by column width (assuming i'm gonna stay "compatible").

opm/simulators/flow/LogOutputHelper.cpp
akva2: sure.

akva2:
jenkins build this please

PR #6092, opened 2025-03-18
Support WCONHIST/WCONINJ in ACTIONX

blattms:
Downstream of OPM/opm-common#4527

Comparison with other simulator looks good for the test cases actionx/ACTIONX_WCONINJH.DATA actionx/ACTIONX_WCONHIST.DATA

I needed to make the container set up for producers/injector to work for the other types.  This here seems to work if the ACTIONX is executed at the end of a report step. I am bit skeptical whether this also works during the time stepping. Maybe there is  a better way?

### DIFF
--- a/opm/simulators/wells/SingleWellState.cpp
+++ b/opm/simulators/wells/SingleWellState.cpp
@@ -323,13 +323,40 @@ update_injector_targets(const Well& ecl_well, const SummaryState& st)
 }
 
 template<class Scalar>
-void SingleWellState<Scalar>::
-update_targets(const Well& ecl_well, const SummaryState& st)
+bool SingleWellState<Scalar>::
+update_type_and_targets(const Well& ecl_well, const SummaryState& st)
 {
+    bool switchedToProducer = false;
+    if (this->producer != ecl_well.isProducer()) {
+        // type has changed due to ACTIONX
+        // Make sure that we consistent with the ecl_well
+        switchedToProducer = this->producer = ecl_well.isProducer();
+        if (switchedToProducer) {
+            this->production_cmode = ecl_well.productionControls(st).cmode;
+            // clear injection rates
+            for(auto&& val: this->surface_rates) {
+                if (val > 0.0) {
+                    val = 0.0;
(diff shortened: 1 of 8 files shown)


opm/simulators/wells/SingleWellState.cpp
```diff
@@ -323,13 +323,40 @@ update_injector_targets(const Well& ecl_well, const SummaryState& st)
...
+        // type has changed due to ACTIONX
+        // Make sure that we consistent with the ecl_well
+        switchedToProducer = this->producer = ecl_well.isProducer();
+        if (switchedToProducer) {
+            this->production_cmode = ecl_well.productionControls(st).cmode;
+            // clear injection rates
+            for(auto&& val: this->surface_rates) {
+                if (val > 0.0) {
+                    val = 0.0;
+                }
```
blattms, line 340: Without this I get an exception because wfac is larger than 1.

blattms:
This is now tested by the client and works perfectly for the test cases. Ready for review.

GitPaean:
Thanks for the request, but it it is okay I think I will suggest @bska to review this for better efficiency. I can review it while I realized it will kill some time for me to get familiar with the implementation of the actions first.

opm/simulators/wells/SingleWellState.cpp
bska: > Without this I get an exception because wfac is larger than 1.

Would you mind elaborating a little on that, please?  What is `wfac`?

bska (commented):
I like that we're now able to handle the case of a well switching from a producer to an injector, or the other way around, in the context of an action block&ndash;that's a great improvement.  Other than that, I only have minor comments here.

opm/simulators/wells/SingleWellState.cpp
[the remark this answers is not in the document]
blattms: I think we used to need them due to compiler bugs.
Anyway, I changed to your preference (under slight protest...).

bska:
jenkins build this please

PR #6093, opened 2025-03-19
PRT: Use {fmt} to format production log

akva2:
Continuing. Same reasoning as in https://github.com/OPM/opm-simulators/pull/6090


akva2:
jenkins build this please

PR #6096, opened 2025-03-19
PRT: Use {fmt} to format fip logs

akva2:
This one is a bit more messy as I can't really decode the actual formatting of this table. At least I have codified it as much as possible.


bska (approved):
With the **possible** caveat that we may have to restore the leading blank for these reports, this looks good to me and I'll merge into master.

PR #6098, opened 2025-03-20
Also solve std for every inner iterations as default

totto82:
In master, we always solve inner iterations for MSwells but not StdWells. The lack of inner iterations for StdWells can lead to convergence issues for Newton.  This changes the default value to 99 and removes the check for MSWells, thus allowing not to solve MSWells during some Newton iterations.

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -851,7 +851,7 @@ namespace Opm
 
         // only use inner well iterations for the first newton iterations.
         const int iteration_idx = simulator.model().newtonMethod().numIterations();
-        if (iteration_idx < this->param_.max_niter_inner_well_iter_ || this->well_ecl_.isMultiSegment()) {
+        if (iteration_idx < this->param_.max_niter_inner_well_iter_) {
             const auto& ws = well_state.well(this->indexOfWell());
             const auto pmode_orig = ws.production_cmode;
             const auto imode_orig = ws.injection_cmode;
--- a/opm/simulators/flow/BlackoilModelParameters.hpp
+++ b/opm/simulators/flow/BlackoilModelParameters.hpp
@@ -100,7 +100,7 @@ struct TolerancePressureMsWells { static constexpr Scalar value = 0.01*1e5; };
 template<class Scalar>
 struct MaxPressureChangeMsWells { static constexpr Scalar value = 10*1e5; };
 
-struct MaxNewtonIterationsWithInnerWellIterations { static constexpr int value = 8; };
+struct MaxNewtonIterationsWithInnerWellIterations { static constexpr int value = 99; };
 struct MaxInnerIterMsWells { static constexpr int value = 100; };
 struct MaxInnerIterWells { static constexpr int value = 50; };
 struct ShutUnsolvableWells { static constexpr bool value = true; };


totto82:

totto82:
Note that this fixes the issue of repeatedly shutting off NORNE_ATW2013_3B_STDW.DATA as reported by @erikhide in https://github.com/OPM/opm-simulators/pull/6067

GitPaean:
vow, then my guess is that it improves the well solve for the well under THP target?

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -851,7 +851,7 @@ namespace Opm
 
         // only use inner well iterations for the first newton iterations.
         const int iteration_idx = simulator.model().newtonMethod().numIterations();
-        if (iteration_idx < this->param_.max_niter_inner_well_iter_ || this->well_ecl_.isMultiSegment()) {
```
GitPaean, line 854: If we go this direction, we probably want to do it differently here. 
`iteration_idx` by default will be maximum 20. We probably can remove this if condition or using some other if condition if I did not read the code wrong.

totto82: I think it could be useful to keep the option of not solving with inner iterations for all Newton iterations.

GitPaean:
> In some cases, Newton's convergence gets worse.

Norne had this problem from many years ago, not sure whether it is still the case though.

erikhide:
> I have started testing the Norne prediction cases with STW, and the results are not conclusive. In some cases, Newton's convergence gets worse. I will have to dig deeper into these cases to understand what happens.

Note that some of the Norne prediction cases fail when using the current master. I've tested alternative time stepping approaches which seem to work better (i.e. the simulation doesn't fail). An early version is available using `--time-step-control="general3rdorder"`.

totto82:
benchmark this please

totto82:

totto82:
I think we should try to get this in before the release, as not solving the well with inner equations may lead to the stagnation of simulations.

totto82:
Some spikes in the results are the most worrying differences. I have tried to reproduce them locally:
udq_wconprod: WOPR:OPL02. Both master and PR give the same results locally (no spike)
editnnc_model2: WGIP:INJ1. Both master and PR give the same results locally (with spike)
I.e., the spikes are probably caused by something else. 

From my point of view, this should be merged. We would like to use the new local solve approach in the simulator at every iteration, as well as for standard wells.

totto82:
@GitPaean, what do you think? Should we merge this before the release? I think so, as it improves the convergence in most cases. If it doesn't, it points to other issues that must be solved. If this leads to a problem for particular models, we have a simple workaround of passing the old default parameters.

totto82:

GitPaean:
let me have a look. 

It was mentioned that it made some running worse, was there any update regarding that?

GitPaean:
I confirm that `editnnc_model2: WGIP:INJ1.` produces a spike both on the master and this PR locally. 

But it is a weird case, that INJ1 is either GAS or WATER injection history matching mode, while the most of the time (or all the time of reference results), WGIP for INJ1 is zero, which is rather strange.  DBG and PRT file does not mention anything related to this.

GitPaean:
Did you check the results related to UDQ_WCONPROD?

[02_udq_wconprod.pdf](https://github.com/user-attachments/files/19601559/02_udq_wconprod.pdf)

I confirmed the jenkins comparison locally.

GitPaean:
And also 
[01_udq_pyaction.pdf](https://github.com/user-attachments/files/19602802/01_udq_pyaction.pdf)
I confirmed this one in locally. 

did not check all of the regression tests.

totto82:
> Did you check the results related to UDQ_WCONPROD?

It seems super-sensitive to small perturbations in the time-stepping. I will dig deeper to see if I can solve the underlying issue.

vkip:
Using --check-group-constraints-inner-well-iterations=false (or applying https://github.com/OPM/opm-simulators/pull/5924) seems to fix 02_udq_wconprod (only time-stepping differences remain, visually identical results with shorter time steps).

totto82:
I think the issue is related to control switching. https://github.com/OPM/opm-simulators/pull/6145 seems to fix it. However, so does restricting the timestep and reducing the maximum number of well switching per step. So I could just be random, but I don't think it is directly related to this PR. i.e., I think we can merge this and update the data. I also tagged #6145 for the release. It fixes an output bug, but we may need to change the default maximum value of switching as we expect more switching during local solves.

totto82:
@vkip comment strengthens my suspicion that it is related to well-switching.

totto82:
Let's merge #5924 first.

vkip:
> Let's merge #5924 first.

Sounds good, hopefully many of the test failures here disappear after that.

totto82:

totto82:

totto82:

totto82:

totto82:
I will make a new attempt for this release. The way the inner solve code is written now, we really want to solve the STD wells for all Newton iterations. Yes, the run time increases for 1B, 2A, and 2B, but I will argue that that is better than the unfinished simulation of 4B. I have looked through the test failures, and I didn't spot anything that makes me nervous. This PR affects the convergence and thus the time-stepping, which again affects the results. The problematic cases we looked at the last time we investigated this, i.e., the UDQ tests, behave much better now.

GitPaean:
It has been en efforts crossing many years. I am not familiar with the significance of the NORNE cases, while it looks like systematically slower with the iteration for standard wells. 

For Norne, the difference is within the fluctuation, the PR increases the iteration number a little bit, which is not a concern. 

How does it behave in other field cases?

I think I can not make the decision alone, so I invited @atgeirr @bska also join the discussion.

GitPaean:
benchmark please

totto82:
>  I am not familiar with the significance of the NORNE cases,

Conceptually, there is no good reason not to solve the standard wells in every inner iteration, as we already do for MSW. Especially after @steink significantly improved the inner iteration algorithms. The only reason I know of for not updating this long ago is that it caused severe problems for the NORNE cases. This is related to the fact that for NORNE cases WELSPECS item 9 is "STOP" not "SHUT" as is more common.  This has led to some issues with the inner iteration logic, but most, if not all, are now fixed. Or at least 1A-4B, which all ran through with this PR, which was not the case earlier. 4C and 4D are still problematic for the master and this PR (and MSW).

totto82:

steink:
> maybe @steink also want to comment on this?

I think that for the typical cases (quite complex with groups/network) we are targeting, it makes sense to always solve STW as well, so to me it seems it should be the default. 

But I do think that there can be situations where it's not the best option. For instance the wgrupcon/wvfpexp-testcases are rather special as they are quite small (pressure travels fast) and have VREP-control while producers are becomming inoperable. In this case, I think the well problem becomes very non-local, so might give rise to situations where the local well solve does more harm than good for the global solve. For the NORNE-cases, I'm not sure, but they are also very special with the stop-instructions that make the well solves very sensitive to small variations in reservoir pressure and give rise to constantly changing directions of cross-flow.

totto82:
> But I do think that there can be situations where it's not the best option

I think this also applies to MSW. That is why I would also like to make it possible to avoid always using inner iterations for MSW (See the changes proposed in this PR).

GitPaean:
> For some of the NORNE-cases, I don't think the issues are related to logic since when wells are stopped, there is no control/status switching (always solved with zero rate). I think it's more related to small rates with constantly changing flow-directions/cross-flow, so the local well-equations do not behave smoothly with changing reservoir pressure. Also I think the cross-flow treatment with the standard wells is rather crude by design, so if cross-flow is an important ingredient, it should rather be modelled with MSW, even though tiny time-steps might be required. So I'm a bit sceptical to the relevance of all the different versions of the NORNE-model.

Thanks for the comments. Cross-flow in the STDW might be some rough, that is true. But conceptually, we should only apply the well information back to reservoir when the well solution is converged. Non-converged solution can be random in theory. 

Before we go ahead, what is the benefit from observation what this PR can make for the cases that are concerned with? Maybe @totto82 has some comments?

totto82:
> Before we go ahead, what is the benefit from observation what this PR can make for the cases that are concerned with?
See NORNE_ATW2013_4B_STDW in the report above.

GitPaean:

totto82:
It is time to merge this. I have reviewed the test failures, and I don't see any merge stoppers. Based on the comments, I believe we all agree that this is a positive change that should be merged.

atgeirr:
> It is time to merge this. I have reviewed the test failures, and I don't see any merge stoppers. Based on the comments, I believe we all agree that this is a positive change that should be merged.

Then go ahead and do the data update.

PR #6095, opened 2025-03-19
refactor the network iteration code

totto82:
This decouples the group/well constraint check and the network update. The motivation was to get better output to user and .DBG file.

PR #6100, opened 2025-03-21
Actually respect NUPCOL (not NUPCOL+1) and allow network to converge with GLO

vkip:
Also: 

 * Some more informative output in case network does not converge
 * Change order of return arguments in updateWellControlsAndNetworkIteration to be consistent with updateWellControls

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1192,6 +1192,7 @@ namespace Opm {
         // not necessarily that we always need to update once of the network solutions
         bool do_network_update = true;
         bool well_group_control_changed = false;
+        Scalar network_imbalance = 0.0;
         // after certain number of the iterations, we use relaxed tolerance for the network update
         const std::size_t iteration_to_relax = param_.network_max_strict_outer_iterations_;
         // after certain number of the iterations, we terminate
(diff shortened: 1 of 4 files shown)


totto82:
Note. I just refactored the code. https://github.com/OPM/opm-simulators/pull/6095

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1743,7 +1751,7 @@ namespace Opm {
 
 
     template<typename TypeTag>
-    std::pair<bool, bool>
+    std::tuple<bool, bool, typename BlackoilWellModel<TypeTag>::Scalar>
     BlackoilWellModel<TypeTag>::
```
totto82, line 1755: Cant you just use Scalar here?

totto82 (approved):
Thank you for the contribution. I just have one minor question.

totto82:

opm/simulators/wells/BlackoilWellModel_impl.hpp
vkip: I believe I tried that first, and it didn't compile at my end (GCC13)

akva2: Scalar has no meaning in the context as the return value is defined "outside" the class context. typename is required.

totto82: Thanks. I see that now. @vkip can you undraft this if you think it is ready for merging.

vkip:
Test failures are mainly due to different time-stepping. I've been through all the failed tests with shorter time-steps and find results to be (visually) identical to master, except for two tests that differ in a single time-step - these become identical to master when using NUPCOL=11.

totto82:
Thanks for looking into the test failures. I will start the process of updating the reference results and merging this.

totto82:

totto82:
jenkins build this opm-tests=1322 please

PR #6103, opened 2025-03-25
Add support for merging well cells for NLDD partitioning

jakobtorben:
# NLDD Improvements for Partitioning and Better Handling of Wells and Isolated Cells

This PR introduces several improvements to the Nonlinear Domain Decomposition (NLDD) implementation, with a main focus on better handling cells perforated by wells during subdomain partitioning.

### Well Cell Merging
- Cells perforated by the same well are now merged into a single entity before partitioning
- This ensures that all perforated cells of a well stay in the same subdomain
- Prevents wells from being split across NLDD subdomains, which is a requirement for the NLDD implementation
- Creates naturally better partitions than the previous fix to manually move the distributed cells after partitioning
- Uses the new vertex merging functionality from https://github.com/OPM/opm-common/pull/4537 in opm-common

### Well Neighborhood Control
- Added a new method to grow partitions around wells
- Controlled with the parameter `local_domains_partition_well_neighbor_levels_` which specifies the number of levels of neighbours to merge into one entity:
  - A value of 0 keeps only the merged well cells in the subdomain
  - A value of 1 (default) adds immediate neighbours of cells perforated by the well
  - A value of 2 adds all the neighbours of the cells in the previous levels, etc.
- Improves numerical behaviour by creating more buffer cells around wells to avoid large changes on the boundary of domains and automatically places wells in the same NLDD subdomain if placed too close to each other

### Enhanced Partition Visualization
- Added generation of a ResInsight-compatible partition file for easy visualisation and debugging
- Follows expected format for input properties: https://resinsight.org/import/appendingadditionalproperties/
- The file includes all cells (active and inactive) with the NLDD global domainID

#### Code Consistency
- Renamed internal parameters for consistency (from `local_domain_*` to `local_domains_*`)

## Testing
- Added  a large suite of tests for NLDD partitioning
- The changes have been thoroughly tested on multiple complex field cases with various well configurations
- Tests show improved convergence behaviour with wells properly contained within subdomains
- Verified that isolated cells are correctly identified and handled in problematic cases
- Tested with various values of the well neighbour level parameter to confirm proper behaviour
- The ResInsight-compatible output files have been verified to work correctly with visualisation tools

### DIFF
--- a/opm/simulators/flow/BlackoilModelNldd.hpp
+++ b/opm/simulators/flow/BlackoilModelNldd.hpp
@@ -817,7 +817,7 @@ private:
         } else if (model_.param().local_solve_approach_ == DomainSolveApproach::GaussSeidel) {
             // Calculate the measure used to order the domains.
             std::vector<Scalar> measure_per_domain(domains_.size());
-            switch (model_.param().local_domain_ordering_) {
+            switch (model_.param().local_domains_ordering_) {
             case DomainOrderingMeasure::AveragePressure: {
                 // Use average pressures to order domains.
                 for (const auto& domain : domains_) {
@@ -856,7 +856,7 @@ private:
                 }
                 break;
             }
-            } // end of switch (model_.param().local_domain_ordering_)
+            } // end of switch (model_.param().local_domains_ordering_)
 
             // Sort by largest measure, keeping index order if equal.
             const auto& m = measure_per_domain;
@@ -991,7 +991,7 @@ private:
 
         auto zoltan_ctrl = ZoltanPartitioningControl<Element>{};
 
-        zoltan_ctrl.domain_imbalance = param.local_domain_partition_imbalance_;
+        zoltan_ctrl.domain_imbalance = param.local_domains_partition_imbalance_;
 
         zoltan_ctrl.index =
             [elementMapper = &this->model_.simulator().model().elementMapper()]
@@ -1008,7 +1008,7 @@ private:
         };
 
         // Forming the list of wells is expensive, so do this only if needed.
-        const auto need_wells = param.local_domain_partition_method_ == "zoltan";
+        const auto need_wells = param.local_domains_partition_method_ == "zoltan";
 
         const auto wells = need_wells
             ? this->model_.simulator().vanguard().schedule().getWellsatEnd()
@@ -1023,12 +1023,10 @@ private:
         const int num_domains = (param.num_local_domains_ > 0)
             ? param.num_local_domains_
             : detail::countGlobalCells(grid) / default_cells_per_domain;
-        // TODO: Make this a parameter
-        const int num_neighbor_levels = 1;
-        return ::Opm::partitionCells(param.local_domain_partition_method_,
+        return ::Opm::partitionCells(param.local_domains_partition_method_,
                                      num_domains, grid.leafGridView(), wells,
                                      possibleFutureConnectionSet, zoltan_ctrl,
-                                     num_neighbor_levels);
+                                     param.local_domains_partition_well_neighbor_levels_);
     }
 
     std::vector<int> reconstitutePartitionVector() const
--- a/opm/simulators/flow/BlackoilModelNldd.hpp
+++ b/opm/simulators/flow/BlackoilModelNldd.hpp
@@ -107,7 +107,26 @@ public:
         , rank_(model_.simulator().vanguard().grid().comm().rank())
     {
         // Create partitions.
-        const auto& [partition_vector, num_domains] = this->partitionCells();
+        const auto& [partition_vector_initial, num_domains_initial] = this->partitionCells();
+
+        int num_domains = num_domains_initial;
+        std::vector<int> partition_vector = partition_vector_initial;
+
+        // Fix-up for an extreme case: Interior cells who do not have any on-rank
+        // neighbours. Move all such cells into a single domain on this rank,
+        // and mark the domain for skipping. For what it's worth, we've seen this
+        // case occur in practice when testing on field cases.
+        bool isolated_cells = false;
+        for (auto& domainId : partition_vector) {
+            if (domainId < 0) {
+                domainId = num_domains;
+                domains_to_skip_.push_back(domainId);
+                isolated_cells = true;
+            }
+        }
+        if (isolated_cells) {
+            num_domains++;
+        }
 
         // Set nldd handler in main well model
         model.wellModel().setNlddAdapter(&wellModel_);
@@ -222,6 +241,9 @@ public:
         for (const int domain_index : domain_order) {
             const auto& domain = domains_[domain_index];
             SimulatorReportSingle local_report;
+            if (std::find(domains_to_skip_.begin(), domains_to_skip_.end(), domain.index) != domains_to_skip_.end()) {
+                continue;
+            }
             try {
                 switch (model_.param().local_solve_approach_) {
                 case DomainSolveApproach::Jacobi:
@@ -1008,6 +1030,7 @@ private:
     std::vector<Domain> domains_; //!< Vector of subdomains
     std::vector<std::unique_ptr<Mat>> domain_matrices_; //!< Vector of matrix operator for each subdomain
     std::vector<ISTLSolverType> domain_linsolvers_; //!< Vector of linear solvers for each domain
+    std::vector<int> domains_to_skip_; //!< Vector of domains to skip
     SimulatorReportSingle local_reports_accumulated_; //!< Accumulated convergence report for subdomain solvers
     int rank_ = 0; //!< MPI rank of this process
 };
--- a/opm/simulators/flow/BlackoilModelNldd.hpp
+++ b/opm/simulators/flow/BlackoilModelNldd.hpp
@@ -362,23 +362,55 @@ public:
         return local_reports_accumulated_;
     }
 
+    /// Write the partition vector to a file in ResInsight compatible format for inspection
+    /// and a partition file for each rank, that can be used as input for OPM.
     void writePartitions(const std::filesystem::path& odir) const
     {
         const auto& elementMapper = this->model_.simulator().model().elementMapper();
         const auto& cartMapper = this->model_.simulator().vanguard().cartesianIndexMapper();
+        const auto& dims = cartMapper.cartesianDimensions();
+        const auto total_size = dims[0] * dims[1] * dims[2];
 
         const auto& grid = this->model_.simulator().vanguard().grid();
         const auto& comm = grid.comm();
-        const auto nDigit = 1 + static_cast<int>(std::floor(std::log10(comm.size())));
 
-        std::ofstream pfile { odir / fmt::format("{1:0>{0}}", nDigit, comm.rank()) };
+        // Create a full-sized vector initialized with -1 (indicating inactive cells)
+        auto full_partition = std::vector<int>(total_size, -1);
 
+        // Fill active cell values for this rank
         const auto p = this->reconstitutePartitionVector();
         auto i = 0;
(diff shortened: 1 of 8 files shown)


bska:
jenkins build this opm-common=4537 please

opm/simulators/flow/partitionCells.cpp
```diff
@@ -109,11 +109,26 @@ class ZoltanPartitioner
...
+                         const int                                             num_neighbor_levels);
+
+    /// Connect neighbors of cells up to a specified level.
+    ///
+    /// \tparam GridView DUNE grid view type
+    ///
+    /// \param[in] cells Initial set of cells.
+    /// \param[in] num_levels Number of neighbor levels to include.
+    ///
+    /// \return Set of all connected cells including neighbors up to specified level.
```
bska, line 129: Tiny nit: This function doesn't (directly) return anything.  Is this return value comment a left-over from earlier development?

opm/simulators/flow/partitionCells.cpp
```diff
@@ -109,11 +109,26 @@ class ZoltanPartitioner
...
     void buildLocalGraph(const GridView&                                       grid_view,
                          const std::vector<Opm::Well>&                         wells,
                          const std::unordered_map<std::string, std::set<int>>& possibleFutureConnections,
-                         const Opm::ZoltanPartitioningControl<Element>&        zoltan_ctrl);
+                         const Opm::ZoltanPartitioningControl<Element>&        zoltan_ctrl,
+                         const int                                             num_neighbor_levels);
+
+    /// Connect neighbors of cells up to a specified level.
+    ///
+    /// \tparam GridView DUNE grid view type
```
bska, line 124: There is no template parameter here so maybe we can remove this parameter description?

opm/simulators/flow/partitionCells.cpp
```diff
@@ -233,33 +290,27 @@ ZoltanPartitioner::partition(const int                                      num_
...
+                                   const Opm::ZoltanPartitioningControl<Element>& zoltan_ctrl,
+                                   const bool                                     build_neighbor_map)
 {
     auto g2l = std::unordered_map<int, int>{};
+    if (build_neighbor_map) {
+        neighbor_map_.clear();
+    }
 
     for (const auto& element : elements(grid_view, Dune::Partitions::interior)) {
-        {
```
bska, line 260: Tiny nit: By removing this block, the `c` object becomes visible to the entire `for` loop.  It's not a **huge** problem, obviously, especially since it is `const`, but I originally want to keep the object scope small to avoid inadvertent shadowing warnings.

opm/simulators/flow/partitionCells.cpp
```diff
@@ -600,12 +666,13 @@ Opm::partitionCells(const std::string& method,
                     const GridView&    grid_view,
                     [[maybe_unused]] const std::vector<Well>&                              wells,
                     [[maybe_unused]] const std::unordered_map<std::string, std::set<int>>& possibleFutureConnections,
-                    [[maybe_unused]] const ZoltanPartitioningControl<Element>&             zoltan_ctrl)
+                    [[maybe_unused]] const ZoltanPartitioningControl<Element>&             zoltan_ctrl,
+                    const int                                             num_neighbor_levels)
```
bska, line 670: I think this new parameter should be tagged as `[[maybe_unusued]]` as well since it's only referenced if we `HAVE_MPI && HAVE_ZOLTAN`.  In other words, it's unused in sequential builds.

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -107,7 +107,30 @@ class BlackoilModelNldd
...
+        std::vector<int> partition_vector = partition_vector_initial;
+
+        // Fix-up for an extreme case: Interior cells whose neighbours are all
+        // on a different rank--i.e., interior cells which do not have on-rank
+        // neighbours.  Make each such cell be a separate domain on this rank.
+        // Don't remove this code unless you've taken remedial action elsewhere
+        // to ensure the situation doesn't happen.  For what it's worth, we've
+        // seen this case occur in practice when testing on field cases.
+        // The non-reachable cells are marked with -1 in the partition vector.
+        // Here we collect them into  a single domain, per rank.
```
bska, line 122: This comment appears to be a little inconsistent now.  The first part says that we make each isolated cell be a separate single domain, while the last part says that we put all of these cells into one combined domain.  Perhaps you can clarify the comment a little?

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -340,23 +366,62 @@ class BlackoilModelNldd
...
+            for (const auto& val : full_partition) {
+                resInsightFile << val << '\n';
+            }
+            resInsightFile << "/" << '\n';
+        }
+
+        const auto nDigit = 1 + static_cast<int>(std::floor(std::log10(comm.size())));
+        auto partition_fname = odir / fmt::format("{1:0>{0}}", nDigit, comm.rank());
+        std::ofstream pfile { partition_fname };
+        if (!pfile) {
```
bska, line 415: You're not supposed to need this check.  The `std::ofstream` constructor already throws an exception if the file cannot be opened.  I suppose there's something to be said for making this code robust in parallel where we might risk that some processes/ranks are able to open the file while others aren't.  In that case we'd get weird hangs.  Handling that, however, will require introducing parallel exception handling.

opm/simulators/flow/partitionCells.cpp
```diff
@@ -203,9 +226,43 @@ template <class GridView, class Element>
...
+    // Initialize search structures
+    std::unordered_set<int> visited(cells.begin(), cells.end());
+    std::vector<int> frontier;
+    std::vector<int> new_frontier;
+    frontier.swap(cells);
+    frontier.reserve(frontier.size() * 2);
+
+    // Breadth-first search for neighbor cells up to num_levels away
+    for (int level = 0; level < num_levels && !frontier.empty(); ++level) {
+        new_frontier.reserve(frontier.size() * 2);
```
bska, line 249: There's a subtle problem here.  Since you `move()` the `new_frontier` object into `frontier` below, it will be in a "valid but unspecified state" if this loop runs (at least) twice.  I think I might personally write this loop using `vector<>::swap()` instead, i.e., something like
```C++
for (level) {
    new_frontier.clear();
    new_frontier.reserve(...);

    for (cell) {
        // ...
    }

    frontier.swap(new_frontier);
}
```
which avoids that semantic pitfall.

bska (commented):
This looks really good.  I mostly have minor comments, although there's one part that I'd write differently.

opm/simulators/flow/partitionCells.cpp
jakobtorben: Good catch! This is indeed a remedy for some code that has been optimised. Removed it now.

opm/simulators/flow/partitionCells.cpp
jakobtorben: Again something I forgot to remove once this was optimsed away. Removed it now.

opm/simulators/flow/partitionCells.cpp
jakobtorben: That makes sense! I added the block back in.

opm/simulators/flow/partitionCells.cpp
jakobtorben: Good point, tagged it correctly now.

opm/simulators/flow/BlackoilModelNldd.hpp
jakobtorben: It was misleading, indeed. It should be clearer now.

Note that before, we did not skip this domain. I mostly added this functionality as I experienced getting around 7 times as many domains as expected on a field case where there were many single isolated cells, which will be safe to skip. From the earlier comment, it sounded like the isolated cells could have neighbours on different ranks. But it should be safe to skip anyways, as NLDD would not update any solutions outside of this rank anyways. Right?

opm/simulators/flow/partitionCells.cpp
jakobtorben: Good point. Adopted `swap` instead now.

jakobtorben:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  This looks good to me now and you've addressed all my remarks.  Unless @atgeirr disagrees I think this should be merged into the master branch now.

jakobtorben:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  This looks good to me now and as the build check is green I'll merge into master.

opm/simulators/flow/partitionCells.cpp
```diff
@@ -182,11 +199,15 @@ class ZoltanPartitioner
     ///
     /// \param[in] g2l Mapping from globally unique cell IDs to local,
     ///   on-rank active cell IDs.  Return value from \c connectElements().
+    ///
+    /// \param[in] num_neighbor_levels Number of neighbor levels to include when connecting well cells.
+    ///   Default is 0, which means only direct well connections are considered.
```
atgeirr, line 204: No default here.

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -222,6 +241,9 @@ class BlackoilModelNldd
         for (const int domain_index : domain_order) {
             const auto& domain = domains_[domain_index];
             SimulatorReportSingle local_report;
+            if (std::find(domains_to_skip_.begin(), domains_to_skip_.end(), domain.index) != domains_to_skip_.end()) {
```
atgeirr, line 244: It is not a priori clear to me that it is a good idea to skip these domains of isolated cells, as a local solve might reduce newton iterations. However, given that this code has been tested thoroughly, I assume the skipping approach has been validated?

Could store a flag in the domain itself rather than doing a lookup here, to save a little time.

atgeirr (commented):
Only a few minor issues. Should be merged when addressed.

atgeirr:
Did not notice it was merged while I looked at it... Anyway not problematic, but I would like the comments to be addressed in a small follow-up PR.

opm/simulators/flow/partitionCells.hpp
```diff
@@ -94,7 +96,8 @@ partitionCells(const std::string&                                    method,
                const GridView&                                       grid_view,
                const std::vector<Well>&                              wells,
                const std::unordered_map<std::string, std::set<int>>& possibleFutureConnections,
-               const ZoltanPartitioningControl<Element>&             zoltan_ctrl);
+               const ZoltanPartitioningControl<Element>&             zoltan_ctrl,
+               const int                                             num_neighbor_levels = 0);
```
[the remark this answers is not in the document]
jakobtorben: I agree, no important reason for keeping this now. Removed it in https://github.com/OPM/opm-simulators/pull/6117.

opm/simulators/flow/partitionCells.cpp
```diff
@@ -109,11 +109,24 @@ class ZoltanPartitioner
...
     void buildLocalGraph(const GridView&                                       grid_view,
                          const std::vector<Opm::Well>&                         wells,
                          const std::unordered_map<std::string, std::set<int>>& possibleFutureConnections,
-                         const Opm::ZoltanPartitioningControl<Element>&        zoltan_ctrl);
+                         const Opm::ZoltanPartitioningControl<Element>&        zoltan_ctrl,
+                         const int                                             num_neighbor_levels);
+
+    /// Connect neighbors of cells up to a specified level.
+    ///
+    /// \param[in,out] cells Initial cells to find neighbors of. Updated to include all neighbors
```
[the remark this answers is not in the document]
jakobtorben: I updated the comment in https://github.com/OPM/opm-simulators/pull/6117 to make it clearer; check if it makes more sense now.

PR #6105, opened 2025-03-25
Updating MICP

daavid00:
Depends on https://github.com/OPM/opm-common/pull/4538

### DIFF
--- a/flow/flow_micp.cpp
+++ b/flow/flow_micp.cpp
@@ -18,14 +18,15 @@
 
 #include <flow/flow_micp.hpp>
 
+#include <opm/grid/CpGrid.hpp>
+
 #include <opm/material/common/ResetLocale.hpp>
 #include <opm/models/blackoil/blackoilonephaseindices.hh>
-
-#include <opm/grid/CpGrid.hpp>
-#include <opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp>
-#include <opm/simulators/flow/Main.hpp>
 #include <opm/models/blackoil/blackoillocalresidualtpfa.hh>
 #include <opm/models/discretization/common/tpfalinearizer.hh>
+
+#include <opm/simulators/flow/Main.hpp>
+#include <opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp>
 
 namespace Opm {
 namespace Properties {
--- a/flow/flow_micp.cpp
+++ b/flow/flow_micp.cpp
@@ -24,6 +24,8 @@
 #include <opm/grid/CpGrid.hpp>
 #include <opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp>
 #include <opm/simulators/flow/Main.hpp>
+#include <opm/models/blackoil/blackoillocalresidualtpfa.hh>
+#include <opm/models/discretization/common/tpfalinearizer.hh>
 
 namespace Opm {
 namespace Properties {
@@ -36,6 +38,10 @@ template<class TypeTag>
 struct EnableMICP<TypeTag, TTag::FlowMICPProblem> {
     static constexpr bool value = true;
 };
+template<class TypeTag>
+struct Linearizer<TypeTag, TTag::FlowMICPProblem> { using type = TpfaLinearizer<TypeTag>; };
+template<class TypeTag>
+struct LocalResidual<TypeTag, TTag::FlowMICPProblem> { using type = BlackOilLocalResidualTPFA<TypeTag>; };
 //! The indices required by the model
 template<class TypeTag>
 struct Indices<TypeTag, TTag::FlowMICPProblem>
@@ -58,6 +64,13 @@ public:
                                          /*enabledCompIdx=*/FluidSystem::waterCompIdx,
                                          5>; //Five MICP components
 };
+
+template<class TypeTag>
+struct EnableDiffusion<TypeTag, TTag::FlowMICPProblem> { static constexpr bool value = true; };
+
+template<class TypeTag>
+struct EnableDispersion<TypeTag, TTag::FlowMICPProblem> { static constexpr bool value = true; };
+
 }}
 
 namespace Opm {
--- a/flowexperimental/BlackOilIntensiveQuantitiesGlobalIndex.hpp
+++ b/flowexperimental/BlackOilIntensiveQuantitiesGlobalIndex.hpp
@@ -53,6 +53,7 @@
 
 #include <opm/utility/CopyablePtr.hpp>
 
+#include <stdexcept>
 #include <utility>
 
 #include <fmt/format.h>
--- a/flowexperimental/BlackOilIntensiveQuantitiesGlobalIndex.hpp
+++ b/flowexperimental/BlackOilIntensiveQuantitiesGlobalIndex.hpp
@@ -512,6 +512,11 @@ public:
     Scalar referencePorosity() const
     { return referencePorosity_; }
 
+    const Evaluation& permFactor() const
+    {
+        throw std::logic_error("permFactor() is not yet implemented for compositional modeling"); 
+    }
+
 private:
     friend BlackOilSolventIntensiveQuantities<TypeTag>;
     friend BlackOilExtboIntensiveQuantities<TypeTag>;
--- a/opm/models/blackoil/blackoildispersionmodule.hh
+++ b/opm/models/blackoil/blackoildispersionmodule.hh
@@ -208,79 +208,78 @@ public:
                  * dispersivity
                  * diffR;
             diffR = inIq.ureaConcentration()- Toolbox::value(exIq.ureaConcentration());
-            // dividing by scaling factor 10 (see WellInterfaceGeneric.cpp)
             flux[contiUreaEqIdx] +=
                  bAvg
                  * normVelocityAvg[waterPhaseIdx]
                  * dispersivity
-                 * diffR / 10;
+                 * diffR;
+            return;
         }
-        else {
-            unsigned pvtRegionIndex = inFs.pvtRegionIndex();
-            for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
-                if (!FluidSystem::phaseIsActive(phaseIdx)) {
-                    continue;
-                }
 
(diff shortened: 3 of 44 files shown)


flow/flow_micp.cpp
```diff
@@ -24,6 +24,8 @@
 #include <opm/grid/CpGrid.hpp>
 #include <opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp>
 #include <opm/simulators/flow/Main.hpp>
+#include <opm/models/blackoil/blackoillocalresidualtpfa.hh>
```
akva2, line 27: keep alpha order

flowexperimental/BlackOilIntensiveQuantitiesGlobalIndex.hpp
```diff
@@ -512,6 +512,11 @@ class BlackOilIntensiveQuantitiesGlobalIndex
     Scalar referencePorosity() const
     { return referencePorosity_; }
 
+    const Evaluation& permFactor() const
+    {
+        throw std::logic_error("permFactor() is not yet implemented for compositional modeling");
```
akva2, line 517: need ```<stdexcept>```

opm/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -692,6 +694,15 @@ public:
     Scalar referencePorosity() const
     { return referencePorosity_; }
 
+    const Evaluation& permFactor() const
+    {
+        if constexpr (enableMICP)
+            return MICPIntQua::permFactor();
+        if constexpr (enableSaltPrecipitation)
```
akva2, line 701: else if constexpr, and final else.

opm/models/blackoil/blackoilmicpparams.cpp
```diff
@@ -26,7 +26,11 @@
 
 #if HAVE_ECL_INPUT
 #include <opm/input/eclipse/EclipseState/EclipseState.hpp>
-#include <opm/input/eclipse/EclipseState/MICPpara.hpp>
+#include <opm/input/eclipse/EclipseState/Tables/BiofilmTable.hpp>
+#include <opm/input/eclipse/EclipseState/Tables/DiffMICPTable.hpp>
+#include <opm/input/eclipse/EclipseState/Tables/PermfactTable.hpp>
+#include <opm/input/eclipse/EclipseState/Tables/TableManager.hpp>
+#include <opm/input/eclipse/EclipseState/Tables/SimpleTable.hpp>
```
akva2, line 33: s before t

opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -569,7 +570,7 @@ private:
         if (anyFlores) {
             floresInfo_.reserve(numCells, 6 * numCells);
         }
-        if (enableDispersion) {
+        if (enableDispersion || enableMICP) {
```
akva2, line 573: if constexpr

opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -616,7 +617,7 @@ private:
                 if (anyFlores) {
                     floresInfo_.appendRow(loc_flinfo.begin(), loc_flinfo.end());
                 }
-                if (enableDispersion) {
+                if (enableDispersion || enableMICP) {
```
akva2, line 620: if constexpr

opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -757,7 +758,7 @@ private:
                 const IntensiveQuantities& intQuantsEx = model_().intensiveQuantities(globJ, /*timeIdx*/ 0);
                 LocalResidual::computeFlux(adres,darcyFlux, globI, globJ, intQuantsIn, intQuantsEx, nbInfo.res_nbinfo,  problem_().moduleParams());
                 adres *= nbInfo.res_nbinfo.faceArea;
-                if (enableDispersion) {
+                if (enableDispersion || enableMICP) {
```
akva2, line 761: if constexpr

opm/models/io/vtkblackoilmicpmodule.hpp
```diff
@@ -141,8 +141,9 @@ class VtkBlackOilMICPModule : public BaseOutputModule<TypeTag>
                 }
 
                 if (params_.ureaConcentrationOutput_) {
+                    // multypliging by scaling factor 10 (see WellInterfaceGeneric.cpp)
```
akva2, line 144: multiplying

opm/simulators/utils/phaseUsageFromDeck.hpp
```diff
@@ -39,6 +39,9 @@ namespace Opm
     /// Looks at presence of WATER, OIL and GAS keywords in deck
     /// to determine active phases.
     PhaseUsage phaseUsageFromDeck(const Deck& deck);
+
+    /// Handels MICP
```
akva2, line 43: handles

akva2 (commented):
Nitpics

blattms:
Please provide a more speaking title and a description if necessary. This will help with the manual.

daavid00:
Since this is quite technical and involves a lot of things that need to be documented, I am planning on doing myself a PR to the manual with all details after it is merged (as mentioned [here](https://github.com/OPM/opm-common/pull/4538#issue-2946041154)).

opm/models/discretization/common/tpfalinearizer.hh
daavid00: Many thanks @akva2 for your comments, the latest commit attempts to address them. Regarding this comment, only `enableMICP` can be checked with constexpr, since `enableDispersion` depends on if the deck has the `DISPERC` keyword. Please let me know if the way this is handle in the latest commit is ok, or if this needs to be modified.

akva2: yeah, in that case revert to prior solution. you cannot combine if and if constexpr in one statement.
since we need to do a runtime evaluation anyways for the one case, manually duplicating code so we can if constexpr it out for the other makes no sense.

daavid00: Thanks, now this is set back, I will start a jenkins then.

opm/models/blackoil/blackoildispersionmodule.hh
```diff
@@ -175,86 +182,105 @@ public:
...
+                 * dispersivity
+                 * diffR;
+            diffR = inIq.ureaConcentration()- Toolbox::value(exIq.ureaConcentration());
+            // dividing by scaling factor 10 (see WellInterfaceGeneric.cpp)
+            flux[contiUreaEqIdx] +=
+                 bAvg
+                 * normVelocityAvg[waterPhaseIdx]
+                 * dispersivity
+                 * diffR / 10;
+        }
```
totto82, line 217: add return and remove else

opm/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -506,11 +508,11 @@ public:
...
         porosity_ *= problem.template rockCompPoroMultiplier<Evaluation>(*this, globalSpaceIdx);
 
-        // the MICP processes change the porosity
+        // deal with MICP
         if constexpr (enableMICP){
           Evaluation biofilm_ = priVars.makeEvaluation(Indices::biofilmConcentrationIdx, timeIdx, linearizationType);
           Evaluation calcite_ = priVars.makeEvaluation(Indices::calciteConcentrationIdx, timeIdx, linearizationType);
-          porosity_ += - biofilm_ - calcite_;
+          porosity_ += max(- biofilm_ - calcite_, -referencePorosity_+1e-8);
         }
```
totto82, line 516: Could you explain the max(-refPoro+1e-8) part. Why is it needed?

opm/models/blackoil/blackoilnewtonmethod.hpp
```diff
@@ -373,25 +373,24 @@ class BlackOilNewtonMethod : public GetPropType<TypeTag, Properties::DiscNewtonM
...
+                }
+                if (pvIdx == Indices::oxygenConcentrationIdx) {
+                    nextValue[pvIdx] = std::max(nextValue[pvIdx], Scalar{0.0});
+                }
+                if (pvIdx == Indices::ureaConcentrationIdx) {
+                    nextValue[pvIdx] = std::max(nextValue[pvIdx], Scalar{0.0});
+                }
+                if (pvIdx == Indices::biofilmConcentrationIdx) {
+                    nextValue[pvIdx] = std::clamp(nextValue[pvIdx], Scalar{0.0}, this->problem().referencePorosity(globalDofIdx, 0) - 1.e-8);
+                }
```
totto82, line 390: Same as above. Please add an explanation.

totto82 (changes requested):
Thanks for updating and improving the MICP implementation. I have some suggestions for code improvements that I would like you to consider.

daavid00:
Many thanks @totto82 for your comments, I think I have addressed them. Please let me know if these PR's needs more modifications.

totto82:
Thanks for the update. I think what you did with the scaling factor is correct. Now, you only scale the equation for urea with a factor of 10. And not the variable so you don't need to scale back? Did you see any difference in non-linear behaviors (and results)? The only comment not addressed, I believe, is the first one, i.e., the alphabetic ordering of included files. After that, this can be merged IMO.

daavid00:
> Thanks for the update. I think what you did with the scaling factor is correct. Now, you only scale the equation for urea with a factor of 10. And not the variable so you don't need to scale back? Did you see any difference in non-linear behaviors (and results)? The only comment not addressed, I believe, is the first one, i.e., the alphabetic ordering of included files. After that, this can be merged IMO.

I think the alphabetic order mentioned by @akva2 has been addressed (somehow it does not showed as Outdated in those comments). The variable does not need to be scaled back, since the scaling factor is applied to the flux, source, and storage terms. For the added test, without this weighting the Overall Linear Iterations is 128, while by weighting this it results in 125 (both of them 82 Newton Iterations). For this case the cell results look basically the same, but it is expected this differences are larger for tighter weights,  which could be analogue to tightening the simulator cnv tolerances.

totto82:
Thanks for the testing. From my point of view this is good to go. I will just rerun jenkins to make sure everything is fine.

PR #6107, opened 2025-03-25
Trigger WELSPECS Report From Simulator Side

bska:
This achieves two goals

  1. It ensures proper sequencing with respect to other PRT file reports.
  2. It simplifies generating WELSPECS reports in restarted simulation runs.


tskille:
Tested in Equinor. Results are good and looking forward to seeing this go into master.

bska:
jenkins build this opm-common=4539 please

opm/simulators/flow/EclWriter.hpp
```diff
@@ -389,31 +389,47 @@ class EclWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::Grid>
...
+        // *previous* report step/episode.  We therefore need special case
+        // handling for reportStepNum() == 0 in base runs and
+        // reportStepNum() <= restart step in restarted runs.
+        const auto firstStep = this->initialStep();
+        const auto simStep =
+            std::max(timer.reportStepNum() - 1, firstStep);
+
+        const auto& rpt = this->schedule_[simStep].rpt_config();
+
+        if (rpt.contains("WELSPECS") && (rpt.at("WELSPECS") > 0)) {
```
akva2, line 410: maybe add a temp to add dual lookup

akva2 (approved):

opm/simulators/flow/EclWriter.hpp
bska: > maybe add a temp to add dual lookup

Do you mean something like the typical pattern
```C++
auto pos = map.find(kw);
if (pos != map.end() && pos->second > 0) {
    // do something
}
```
If so, that's not (currently) possible.  The [`RPTConfig`](https://github.com/OPM/opm-common/blob/3d86149501a42e02daadafd5e29a995544c18527/opm/input/eclipse/Schedule/RPTConfig.hpp#L39) class does not (currently) have a `find()` member function so we're (currently) forced to use `contains()` + `at()` to look up elements.

I can add a `RPTConfig::find()` function if you'd like.

bska: > ```
> const bool has_welspec = rpt.contains("WELSPECS") && rpt.at("WELSPECS") > 0;
> ```

I don't understand what you mean.  The second lookup is for `"WELLS"`, not `"WELSPECS"`.

akva2: doh. sorry for the noise.

bska:
jenkins build this opm-common=4539 please

PR #6113, opened 2025-03-27
Make wellModelMatrixAdapter serial only

jakobtorben:
# Remove parallel functionality from WellModelMatrixAdapter

## Description
This PR simplifies the `WellModelMatrixAdapter` by removing its parallel/MPI-specific functionality, since parallel operations are now handled by `WellModelGhostLastMatrixAdapter`.

## Motivation
The parallel functionality in `WellModelMatrixAdapter` was redundant since `WellModelGhostLastMatrixAdapter` is now used for all parallel operations. This change simplifies the codebase by removing unused functionality and makes the intended usage of each adapter class clearer. This is also motivated by the ongoing work to get well operators into the GPUIstl framework, where this hybrid serial/parallel class was causing issues. Another option for keeping this functionality is to split this into a serial and a parallel version, but I did not want to introduce a new class that is not actually used.

### Test update
When testing this locally I noticed a test that is only run if a GPU is available, which was failing for me but has not been executed and failed on jenkins.

### DIFF
--- a/opm/simulators/linalg/FlexibleSolver_impl.hpp
+++ b/opm/simulators/linalg/FlexibleSolver_impl.hpp
@@ -277,7 +277,7 @@ using OBM = Dune::BCRSMatrix<Opm::MatrixBlock<Scalar, N, N>>;
 template<class Scalar, int N>
 using SeqOpM = Dune::MatrixAdapter<OBM<Scalar,N>, BV<Scalar,N>, BV<Scalar,N>>;
 template<class Scalar, int N>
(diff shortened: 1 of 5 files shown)


jakobtorben:
jenkins build this please

atgeirr (approved):

jakobtorben:
Good point, I did not think of that. But none of the other grids were using this in the current state. I don't have anything against adding this back in, but it needs to be one separate parallel version so we do not depend on the comm object in the serial version.

If it is this version you are referring to, then the attic would be the git history. Not sure if there have been other similar adapters or if there is an OPM attic I don't know about 😅

PR #6118, opened 2025-03-28
fix mech extractor

akva2:
Using .emplace() on vector is not a good idea. Compiler also doesn't want to use braced init for the struct for some reason, so resort to push_back().

### DIFF
--- a/opm/simulators/flow/OutputBlackoilModule.hpp
+++ b/opm/simulators/flow/OutputBlackoilModule.hpp
@@ -1544,35 +1544,36 @@ private:
 
         if constexpr (getPropValue<TypeTag, Properties::EnableMech>()) {
             if (this->mech_.allocated()) {
-                this->extractors_.emplace(
-                    [&mech = this->mech_,
-                     &model = simulator_.problem().geoMechModel()](const Context& ectx)
-                    {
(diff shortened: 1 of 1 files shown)


bska:
> Compiler also doesn't want to use braced init for the struct for some reason,

It will in C++20.  The [`Entry`](https://github.com/OPM/opm-simulators/blob/2150481ed62e808e7cd62642fa74117ad579eac3/opm/simulators/flow/OutputExtractor.hpp#L112) is a POD and, before C++20, `vector<>::emplace()`&ndash;did you mean `emplace_back()`&ndash;doesn't work with PODs.  It's really annoying, and it's the reason I sometimes add a banal constructor to my PODs.  The rules for aggregate initialisation change in C++20 and, as a side effect, these kinds of in-place construction operations now work.

akva2:
yeah, that's what i mean. there were two problem, first i had used emplace, which obviously was wrong. the second problem is what you're describing much better than me, ie, that it is a c++20-ism (i knew it had worked in other code i did!). i could add the ctor, but since it's only a single entry and only a one-per-timestep thing I just ate the push_back.

bska (approved):
I finally had the time to test this and it does what it says on the tin.  I'll merge into master.

PR #6119, opened 2025-03-28
Calculate original FIP by reading restart data for step 0

akva2:
Currently the FIP table data in the restart file is not understood, and this meant that the original FIP was had zero values in a restart run. This adds a workaround for this; On restart we first try to load data for step 0. If there, we can use this to generate the OIP.

Downstream of https://github.com/OPM/opm-common/pull/4545

### DIFF
--- a/opm/simulators/flow/EclWriter.hpp
+++ b/opm/simulators/flow/EclWriter.hpp
@@ -368,7 +368,7 @@ public:
 
         // We always calculate the initial fip values as it may be used by various
         // keywords in the Schedule, e.g. FIP=2 in RPTSCHED but no FIP in RPTSOL
-        const Inplace inplace = outputModule_->calc_initial_inplace(simulator_.gridView().comm());
+        outputModule_->calc_initial_inplace(simulator_.gridView().comm());
 
         // check if RPTSOL entry has FIP output
         const auto& fip = simulator_.vanguard().eclState().getEclipseConfig().fip();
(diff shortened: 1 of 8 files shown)


akva2:
jenkins build this opm-common=4545 please

akva2:
jenkins build this opm-common=4545 please

akva2:
jenkins build this opm-common=4545 please

opm/simulators/flow/EclWriter.hpp
```diff
@@ -649,26 +680,10 @@ class EclWriter : public EclGenericWriter<GetPropType<TypeTag, Properties::Grid>
...
-        // very wrong, as the run's intial volumes then correspond to the
-        // volumes at the restart time instead of the start of the base run.
-        // Nevertheless, this is how Flow has "always" done it.
-        //
-        // See GenericOutputBlackoilModule::accumulateRegionSums() for
-        // additional comments.
-        auto inplace = this->outputModule_
-            ->calc_inplace(miscSummaryData, regionData,
-                           this->simulator_.gridView().comm());
+        this->outputModule_->calc_initial_inplace(this->simulator_.gridView().comm());
```
bska, line 686: Maybe add a comment here to remind the reader that this call does nothing if the initial in-place volumes have already been calculated&ndash;i.e., from the T=0 values.

bska (commented):
Looks good to me.  I'd just like a little more context for the `calc_initial_inplace()` function call in `EclWriter<>::endRestart()`.

opm/simulators/flow/EclWriter.hpp
akva2: comment added.

akva2:
jenkins build this please

PR #6121, opened 2025-03-28
Disambiguate two identical tests.

michal-toth:
The changed test used `partition-method=1` before the `partition-method=3` became the default option. To test the other partitioning method I added a test almost identical to the one touched by this PR with the only difference being the partitioning method. (That test starts at line 90, it is just below this one.) Changing the default partitioning method made these two tests identical. This PR makes them different again.


michal-toth:
jenkins build this please

PR #6123, opened 2025-03-30
Fix getName phase index for the RPTRST output

daavid00:
While testing `BG` and `KRG` in `RPTRST` for a `GASWATER` deck, then this issue came out (e.g., `KRG` would write the `WATKR`).  This commit fixes this.

### DIFF
--- a/opm/simulators/flow/GenericOutputBlackoilModule.cpp
+++ b/opm/simulators/flow/GenericOutputBlackoilModule.cpp
@@ -905,14 +905,15 @@ doAllocBuffers(const unsigned bufferSize,
                                                }
                                            }
                                        }
-
+                                       unsigned phaseIdx = 0;
                                        for (int phase : phases) {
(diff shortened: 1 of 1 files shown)


daavid00:
jenkins build this please

akva2:
I don't understand how this fixes anything. The phases run over an array that is 
```
(gasPhaseIdx, waterPhaseIdx, oilPhaseIdx), 
```
which is (0,1,2). This does the same, just with a non-typed index ?

daavid00:
When writing for all phases then no problem, but when we only would like to write for one phase like described in the first comment, then the issue appears (to reproduce this then you can try to run [this](https://github.com/OPM/opm-tests/blob/master/h2store/H2STORE_GW.DATA) with RPTRST 'BASIC=2' KRG /

akva2:
right. that was stupid. better fix in https://github.com/OPM/opm-simulators/pull/6134

daavid00:
Thanks, then I am closing this.

PR #6124, opened 2025-03-31
Fix and improve nldd timing and output reporting

jakobtorben:
# Improving reporting of NLDD

## Changes
This PR fixes some bugs that were present in the Non-Linear Domain Decomposition (NLDD) reporting and improves how statistics from the NLDD solver are collected, displayed, and analyzed. 

The key improvements are:

1. Fixed a bug that the simulation summary reported the NLDD summary for rank 0 only. Now, it only reports the total local solve time in the simulation summary, and more detailed reports are shown in the DBG file. Removing the breakdown from the final summary report is motivated by the fact that this does not make sense for NLDD in parallel, where all the ranks have to wait for the slowest rank. Instead, we now output detailed reports per domain and per rank in the DBG file, which makes more sense.
2. There were also some inconsistencies in the NLDD assembly reporting, which have been fixed.
3. Better reporting structure: Changed `BlackoilModel::localAccumulatedReports()` from returning a simple `SimulatorReportSingle` to a more comprehensive `SimulatorReport` that can track both successful and failed solves.
4. Per-domain statistics: Added a new method `BlackoilModel::domainAccumulatedReports()` that provides detailed statistics for each individual domain on the current rank.
5. Cell-level visualization: Implemented `BlackoilModel::writeNonlinearIterationsPerCell()` to generate ResInsight-compatible files showing nonlinear iteration counts at the cell level.
6. Added a helper method `BlackoilModel::hasNlddSolver()` that makes it cleaner to check if an NLDD solver exists before attempting to use it.
7. Created an NLDD-specific report writer with more details, used for rank- and domain-level statistics.

### 1. Domain Distribution Summary
The solver now provides a summary of how cells, wells and domains are distributed across ranks:

```
NLDD domain distribution summary:
  rank   owned cells   overlap cells   total cells   wells   domains
--------------------------------------------------------------------
     0         9682           2227         11909       5       12
     1        11809           1777         13586      14       12
     2        12131            536         12667      12       14
     3        10809            590         11399       5       13
--------------------------------------------------------------------
   sum        44431           5130         49561      36       51
```

### 2. Per-Domain Performance Reports
Each domain's performance is now reported with detailed statistics in the DBG file:

```
======  Accumulated local solve data for domain 0 on rank 1 ======
Owned + overlap cells:          1086
Number of wells:                   2
Number of domains:                 1
-------------------------------------------------------
Total time:                        0.00 s
  Pre/post/wait time:              0.00 s
  Solver time:                     2.87 s (Wasted: 0.1 s; 1.9%)
    Assembly time:                 1.39 s (Wasted: 0.0 s; 1.5%)
      Well assembly:               0.06 s (Wasted: 0.0 s; 1.3%)
    Linear solve time:             0.45 s (Wasted: 0.0 s; 2.3%)
      Linear setup:                0.19 s (Wasted: 0.0 s; 2.3%)
    Props/update time:             1.02 s (Wasted: 0.0 s; 2.2%)
Overall Linearizations:          1250   (Wasted:    22; 1.8%)
Overall Nonlinear Iterations:     787   (Wasted:    21; 2.7%)
Overall Linear Iterations:       1253   (Wasted:    12; 1.0%)
-------------------------------------------------------
Converged domain solves:         462
  Accepted with relaxed tol:       0
Unconverged domain solves:         1
```

### 3. Per rank statistics
Detailed NLDD statistics accumulated per rank is now shown per rank in the DBG file (what was former shown in the simulation summary).

```
======  Accumulated local solve data for rank 1 ======
Owned + overlap cells:         13586
Number of wells:                  14
Number of domains:                12
-------------------------------------------------------
Total time:                       31.76 s
  Pre/post/wait time:              2.10 s
  Solver time:                    29.66 s (Wasted: 0.1 s; 0.2%)
    Assembly time:                13.91 s (Wasted: 0.0 s; 0.1%)
      Well assembly:               0.30 s (Wasted: 0.0 s; 0.2%)
    Linear solve time:             5.35 s (Wasted: 0.0 s; 0.2%)
      Linear setup:                1.92 s (Wasted: 0.0 s; 0.2%)
    Props/update time:            10.22 s (Wasted: 0.0 s; 0.2%)
Overall Linearizations:         14517   (Wasted:    22; 0.2%)
Overall Nonlinear Iterations:    8961   (Wasted:    21; 0.2%)
Overall Linear Iterations:      13370   (Wasted:    12; 0.1%)
-------------------------------------------------------
Converged domain solves:        5555
  Accepted with relaxed tol:       0
Unconverged domain solves:         1
```

### 4. Visualisation Capabilities
The PR adds the ability to visualise nonlinear iteration counts:

- **Cell-level NLDD iteration count**: A ResInsight-compatible output file that allows visualisation of nonlinear iteration counts per cell, helping identify problematic regions in the model.

ResInsight_nonlinear_iterations

- **Domain-level performance**: The new detailed domain statistics make it possible to analyse the NLDD method in more detail and create useful plots like the one below.

NORNE_ATW2013_nonlinear

## Implementation Details
- The implementation extends the existing `SimulatorReport` structures for domain-specific performance reporting. A dedicated one can be created for NLDD to avoid having some unused fields during the current default Newton method. However, there is quite a lot of functionality built around it for things like adding reports and etc. So, I kept it as one report to avoid too much duplication.
- To produce the NLDD statistics there is a significant amount of processing that needs to be done. However, all the processing-intensive and global calls are only done once per simulation, and I argue that the value it provides justifies the cost.
- In the NLDD domain statistics report, there are the three following fields: 
- Total time: Total time spent in NLDD for this domain
- Pre/post/wait time: Total time spent preparing for NLDD solve and waiting for other ranks to finish
- Solver time: Total time spent doing an actual solve for this domain
- The total solve time will therefore be around the same for all domains and about the same as the local solve time in the simulations summary. Whereas the pre/post/wait time will differ based on the workload on that rank.

### DIFF
--- a/opm/simulators/flow/BlackoilModelNldd.hpp
+++ b/opm/simulators/flow/BlackoilModelNldd.hpp
@@ -380,6 +380,17 @@ public:
         return report;
     }
 
+    void logWells()
+    {
+        for (size_t i = 0; i < domain_reports_accumulated_.size(); ++i) {
+            auto& domain_report = domain_reports_accumulated_[i];
+            domain_report.success.num_wells = 0;
+        }
+        for (const auto& [wname, domain] : wellModel_.well_domain()) {
+            domain_reports_accumulated_[domain].success.num_wells++;
+        }
+    }
+
     /// return the statistics of local solves accumulated for this rank
     const SimulatorReport& localAccumulatedReports() const
     {
@@ -387,8 +398,9 @@ public:
     }
 
     /// return the statistics of local solves accumulated for each domain on this rank
-    const std::vector<SimulatorReport>& domainAccumulatedReports() const
+    std::vector<SimulatorReport>& domainAccumulatedReports()
     {
+        logWells();
         return domain_reports_accumulated_;
     }
 
@@ -469,6 +481,14 @@ private:
         local_reports_accumulated_.success.num_domains = num_domains;
         local_reports_accumulated_.success.num_overlap_cells = overlap_cells;
         local_reports_accumulated_.success.num_owned_cells = owned_cells;
+
+        // Set statistics for each domain report
+        for (size_t i = 0; i < domain_reports_accumulated_.size(); ++i) {
+            auto& domain_report = domain_reports_accumulated_[i];
+            domain_report.success.num_domains = 1;
+            domain_report.success.num_overlap_cells = 0;
+            domain_report.success.num_owned_cells = domains_[i].cells.size();
+        }
 
         // Gather data from all ranks
         std::vector<int> all_owned(comm.size());
--- a/opm/simulators/flow/BlackoilModelNldd.hpp
+++ b/opm/simulators/flow/BlackoilModelNldd.hpp
@@ -204,6 +204,8 @@ public:
 
         assert(int(domains_.size()) == num_domains);
 
+        domain_reports_accumulated_.resize(num_domains);
+
         // Print domain distribution summary
         printDomainDistributionSummary(partition_vector);
     }
@@ -299,7 +301,8 @@ public:
         int& num_domains = counts[3];
         {
             auto step_newtons = 0;
-            for (const auto& dr : domain_reports) {
+            for (size_t i = 0; i < domain_reports.size(); ++i) {
+                const auto& dr = domain_reports[i];
                 if (dr.converged) {
                     ++num_converged;
                     if (dr.total_newton_iterations == 0) {
@@ -307,6 +310,9 @@ public:
                     }
                 }
                 step_newtons += dr.total_newton_iterations;
+                // Accumulate local reports per domain
+                domain_reports_accumulated_[i] += dr;
+                // Accumulate local reports per rank
                 local_reports_accumulated_ += dr;
             }
             num_local_newtons = step_newtons;
@@ -378,6 +384,12 @@ public:
     const SimulatorReport& localAccumulatedReports() const
     {
         return local_reports_accumulated_;
+    }
+
+    /// return the statistics of local solves accumulated for each domain on this rank
+    const std::vector<SimulatorReport>& domainAccumulatedReports() const
+    {
+        return domain_reports_accumulated_;
     }
 
     /// Write the partition vector to a file in ResInsight compatible format for inspection
@@ -1165,6 +1177,7 @@ private:
     std::vector<std::unique_ptr<Mat>> domain_matrices_; //!< Vector of matrix operator for each subdomain
     std::vector<ISTLSolverType> domain_linsolvers_; //!< Vector of linear solvers for each domain
     SimulatorReport local_reports_accumulated_; //!< Accumulated convergence report for subdomain solvers per rank
+    std::vector<SimulatorReport> domain_reports_accumulated_; //!< Accumulated convergence reports per domain
     int rank_ = 0; //!< MPI rank of this process
 };
 
--- a/opm/simulators/flow/BlackoilModelNldd.hpp
+++ b/opm/simulators/flow/BlackoilModelNldd.hpp
@@ -451,6 +451,13 @@ private:
                 ++overlap_cells;
             }
         }
+
+        // Store data for summary output
+        local_reports_accumulated_.success.num_wells = num_wells;
+        local_reports_accumulated_.success.num_domains = num_domains;
+        local_reports_accumulated_.success.num_overlap_cells = overlap_cells;
+        local_reports_accumulated_.success.num_owned_cells = owned_cells;
+
         // Gather data from all ranks
         std::vector<int> all_owned(comm.size());
         std::vector<int> all_overlap(comm.size());
@@ -1019,6 +1026,7 @@ private:
                 const Scalar acceptable_local_cnv_sum = 1.0;
                 if (mb_sum < acceptable_local_mb_sum && cnv_sum < acceptable_local_cnv_sum) {
                     local_report.converged = true;
+                    local_report.accepted_unconverged_domains += 1;
                     logger.debug(fmt::format("Accepting solution in unconverged domain {} on rank {}.", domain.index, rank_));
                     logger.debug(fmt::format("Value of mb_sum: {}  cnv_sum: {}", mb_sum, cnv_sum));
                 } else {
@@ -1032,9 +1040,11 @@ private:
             }
         }
         if (local_report.converged) {
+            local_report.converged_domains += 1;
             auto local_solution = Details::extractVector(solution, domain.cells);
             Details::setGlobal(local_solution, domain.cells, locally_solved);
         } else {
+            local_report.unconverged_domains += 1;
             wellModel_.setPrimaryVarsDomain(domain.index, initial_local_well_primary_vars);
             Details::setGlobal(initial_local_solution, domain.cells, solution);
             model_.simulator().model().invalidateAndUpdateIntensiveQuantities(/*timeIdx=*/0, domain);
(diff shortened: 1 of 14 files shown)


jakobtorben:
jenkins build this please

opm/simulators/flow/FlowMain.hpp
```diff
@@ -391,6 +391,50 @@ namespace Opm {
...
+                    simulator_->model().localAccumulatedReports().reportNLDD(ss);
+                    // Use rank number as tag to ensure correct ordering
+                    local_log.debug(fmt::format("{:05d}", mpi_rank_), ss.str());
+
+                    // Gather all logs and output them in sorted order
+                    auto global_log = gatherDeferredLogger(local_log, FlowGenericVanguard::comm());
+                    if (this->output_cout_) {
+                        global_log.logMessages();
+                    }
+                }
```
bska, line 435: Maybe we could push this logic to a separate helper function?  That way, `runSimulatorAfterSim_()` wouldn't be dominated by NLDD-related output.

bska (commented):
To me this looks good for the most part.  I haven't looked very hard into the details, but the parts that I have checked seem sound to me.  I personally prefer using `std::size_t` from `<cstddef>` over "naked" `size_t`s (from `<stddef.h>`).  You may consider switching to the `std::` variant.

Other than that, I have only minor comments.

bska:
One more general question: If a model has many local domains (e.g. more than 5000), either because the operator requests a very fine-grained partition or because the model "just" has a lot of active cells, how do the statistics plots look then?  Do they become very crowded?

jakobtorben:
The use of `size_t` should be removed. 

For the NLDD_ITER plot in ResInsight, I don't think this would be an issue. Here we could also add a filter to only show the cells above a certain value. For the other plot I included, this is just something I made myself based on the domain statistics so here it is up to the user how they want to plot the statistics.

What I was slightly worried about is if we write many domain reports to the DBG file, this would clutter the file or the file would be too large. This is already starting to become a problem with NLDD, which has a lot more output. But I ended up adding it to the DBG file as the logger in OPM conveniently solved all the problems I had for logging and sorting this info. So for now, I think this is fine, but this is something we can consider fixing later.

jakobtorben:
jenkins build this please

opm/simulators/flow/NlddReporting.hpp
```diff
@@ -0,0 +1,374 @@
...
+    comm.max(full_iterations.data(), full_iterations.size());
+
+    // Only rank 0 writes the file
+    if (rank == 0) {
+        auto fname = odir / "ResInsight_nonlinear_iterations.txt";
+        std::ofstream resInsightFile { fname };
+        if (!resInsightFile) {
+            OPM_THROW(std::runtime_error,
+                      "Failed to open NLDD nonlinear iterations output file: " + fname.string());
+        }
```
bska, line 160: I don't think you really need this check.  The `std::ofstream` constructor will itself throw an exception if the it fails to open the output file.

opm/simulators/flow/NlddReporting.hpp
bska, line 1: You should probably add this new file (`NlddReporting.hpp`) to the set of build files too (top-level `CMakeLists_files.cmake` file).  If the file is supposed to be installed and usable from outside of opm-simulators, and I think it probably should be, then it should go into the [`PUBLIC_HEADER_FILES`](https://github.com/OPM/opm-simulators/blob/2fa0ad149e715f377604cc1a9cca684956790b6c/CMakeLists_files.cmake#L616).  Otherwise, I think it should go into the [`MAIN_SOURCE_FILES`](https://github.com/OPM/opm-simulators/blob/2fa0ad149e715f377604cc1a9cca684956790b6c/CMakeLists_files.cmake#L57).  It is, admittedly, not **quite** as critical in that case, but it does nevertheless help the build system to generate file listings for those who use IDEs.

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -203,6 +199,17 @@ class BlackoilModelNldd
         }
 
         assert(int(domains_.size()) == num_domains);
+
+        domain_reports_accumulated_.resize(num_domains);
+
+        // Print domain distribution summary
+        Opm::printDomainDistributionSummary(
```
bska, line 206: We're already in the Opm namespace, so this `Opm::` prefix is a little misleading.  If you want it, then `Opm::` itself should be prefixed by the global namespace, meaning this function call becomes
```C++
::Opm::printDomainDistributionSummary(...);
```
instead.

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -360,78 +383,69 @@ class BlackoilModelNldd
...
-    /// return the statistics if the nonlinearIteration() method failed
-    const SimulatorReportSingle& localAccumulatedReports() const
+    /// return the statistics of local solves accumulated for this rank
+    const SimulatorReport& localAccumulatedReports() const
     {
         return local_reports_accumulated_;
     }
 
+    /// return the statistics of local solves accumulated for each domain on this rank
+    std::vector<SimulatorReport>& domainAccumulatedReports()
```
bska, line 393: Did you mean to return a reference here?  Unless I'm reading it incorrectly, the header file declares this function as returning an object (a `vector<SimulatorReport>`) rather than a reference to a mutable vector.

bska (commented):
Thanks a lot for the improvements here.  This is starting to look good to me.  I (think I) have identified a couple of minor issues, the most important of which **may** be a differing function type for the `domainAccumulatedReports()` function.  Is that function called at all?

Other than that, we should use `::Opm::` as the namespace prefix when we need to disambiguate `Opm` from within the Opm namespace.

jakobtorben:
jenkins build this please

opm/simulators/flow/NlddReporting.hpp
jakobtorben: You are right. And apologies for not fixing this since you pointed this about in the last PR. This code was written before that...

opm/simulators/flow/NlddReporting.hpp
jakobtorben: Thanks for clarifying what should go into this file! Added it to `PUBLIC_HEADER_FILES` now.

opm/simulators/flow/BlackoilModelNldd.hpp
jakobtorben: Ahh yes I was following the pattern in the partitionCells call but hastly overlooked the prefixed `::`, added this prefix now.

opm/simulators/flow/BlackoilModelNldd.hpp
jakobtorben: Yes, this was intentional, as I don't want to make a copy of the reports, given that the object can be quite large.
There was indeed a mismatch in the function definitions in `BlackOilModel.hpp` and `BlackOilModelNldd.hpp`. 

The complication here was that `domainAccumulatedReports()` needed to count the wells for each domain using the mapping found in the well model for all the wells seen so far. So, we need to mutate this internal variable whenever we use this function to ensure this is updated. We could, of course, always keep this count updated, but I didn't want to do extra work since this is only used once in the end. Using the wells at the end from the schedule like we do for the per-rank reports is not possible since we do not have the well to domain mapping available here.

To overcome this issue, I made the functions consistently a const functions but made the `domain_reports_accumulated_` member  `mutable`. To also ensure that I returned a reference, I removed returning an empty object if not used by NLDD, and instead raised an error.

jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

jakobtorben:
Thanks for your comments @bska @akva2 ! This should be ready from my side now.

bska (commented):
> This should be ready from my side now.

PR #6126, opened 2025-03-31
Removing extra saturation computing BOIPG

daavid00:
Removing extra saturation computing BOIPG


daavid00:
jenkins build this please

totto82:
Thanks. This looks like an obvious bug fix. Will merge.

PR #6128, opened 2025-04-01
Use well rates from well_state_nupcol instead of previous_well_state in updateWellRates

totto82:
The previous_well_state was used to avoid oscillations long time ago. Using the nupcol rates aligns more with what we do elsewhere in the code and will be more accurate in particular at startup of wells. Test failures are expected. A qualitative look at the failed tests indicates that results looks more plausible.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1384,11 +1384,10 @@ updateAndCommunicateGroupData(const int reportStepIdx,
                                                          well_state_nupcol,
                                                          this->groupState());
 
-    // We use the rates from the previous time-step to reduce oscillations
     WellGroupHelpers<Scalar>::updateWellRates(fieldGroup,
                                               schedule(),
                                               reportStepIdx,
-                                              this->prevWellState(),
+                                              well_state_nupcol,
                                               well_state);
 
     // Set ALQ for off-process wells to zero


totto82:

vkip:
For what it's worth: I had a quick look at one of the more severe test failures (0A4_GRCTRL_LRAT_LRAT_GGR_BASE_MODEL2_MSW.DATA), and the behavior with this PR is much more reasonable than with master. Specifically, the early time results are closer to what is obtained with a very short initial time-step.

totto82:
> the behavior with this PR is much more reasonable than with master

I agree. I implemented this some years ago. Back then, it was necessary to avoid severe oscillations, but I don't see that issue anymore. In any case, setting NUPCOL to something reasonable, like 3-5, is a better workaround IMO. Should we try to get this in before the release?

GitPaean (approved):
With more learning and thinking, this is probably fine and should follow what with testing indicates. Not working on this area much and not familiar with the exact comparison between the two different well states. It is difficult for me to understand deeply how it affects the output exactly.  Will wait for @steink 's comments, otherwise, it should be able to go in.

steink:
This PR seems very reasonnable to me. Will just do a test-run, then report back.

steink (approved):
@GitPaean I'm good with this and think it should go in

GitPaean:

GitPaean:
jenkins build this opm-tests=1321 please

GitPaean:
the reference has been updated. I am getting the PR in now.

PR #6130, opened 2025-04-01
Add network+wtest test

totto82:


totto82:

totto82:
jenkins build this opm-tests=1320 please

regressionTests.cmake
```diff
@@ -1619,6 +1619,12 @@ add_test_compareECLFiles(CASENAME gsatprod
                          ABS_TOL ${abs_tol}
                          REL_TOL ${rel_tol}
                          DIR satellite)
+add_test_compareECLFiles(CASENAME network_01_wtest
```
GitPaean, line 1622: all the other NETWORK related is using TUNING, maybe this one can also. Unless you checked the results that without TUNING, it shows good comparison with and without the recent development related to NETWORK + WTEST. 

I only checked the results with TUNING, it looks good for regression.

totto82: I suggest we add the test without TUNING. If it creates problems due to instabilites etc. we can add the TUNING later.

totto82: Often when we add TUNING for the testing it is because the solution is instable wrt. small unrelated changes in the simulator. I assumed that was the reason for adding tuning for the network tests.

GitPaean:
Now I check the results without tuning, the comparison looks fine. I ma getting the PR into master now. 

image

GitPaean:
The reference has been installed. I am merging the PR in now.

PR #6132, opened 2025-04-01
Reapply "Compositional well"

GitPaean:
This reverts commit c782a9f5129b062ada21608dce0aef9a9ba7ec65.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -55,6 +55,27 @@ endmacro()
 # originally generated with the command:
 # find opm -name '*.c*' -printf '\t%p\n' | sort
 list (APPEND MAIN_SOURCE_FILES
+  flowexperimental/BlackOilEnergyIntensiveQuantitiesGlobalIndex.hpp
+  flowexperimental/BlackOilIntensiveQuantitiesGlobalIndex.hpp
+  flowexperimental/comp/EmptyModel.hpp
+  flowexperimental/comp/flowexp_comp.hpp
+  flowexperimental/comp/wells/CompositionalWellModel.hpp
+  flowexperimental/comp/wells/CompositionalWellModel_impl.hpp
+  flowexperimental/comp/wells/CompWellEquations.hpp
+  flowexperimental/comp/wells/CompWellEquations_impl.hpp
+  flowexperimental/comp/wells/CompWell.hpp
+  flowexperimental/comp/wells/CompWell_impl.hpp
+  flowexperimental/comp/wells/CompWellInterface.hpp
+  flowexperimental/comp/wells/CompWellInterface_impl.hpp
+  flowexperimental/comp/wells/CompWellPrimaryVariables.hpp
+  flowexperimental/comp/wells/CompWellPrimaryVariables_impl.hpp
+  flowexperimental/comp/wells/CompWellState.hpp
+  flowexperimental/comp/wells/CompWellState_impl.hpp
+  flowexperimental/comp/wells/SingleCompWellState.hpp
+  flowexperimental/comp/wells/SingleCompWellState_impl.hpp
+  flowexperimental/FIBlackOilModelNoCache.hpp
+  flowexperimental/flowexp.hpp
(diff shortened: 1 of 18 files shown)


GitPaean:

GitPaean:
The failure is due to the invalid option access from the following code, 
```c++
            this->evalSummary(reportStepNum,
                              curTime,
                              localWellData,
                              localWBP,
                              localGroupAndNetworkData,
                              localAquiferData,
                              blockData,
                              miscSummaryData,
                              regionData,
                              inplace,
                              this->outputModule_->initialInplace(),
                              interRegFlows,
                              this->summaryState(),
                              this->udqState());
        }
```

```c++
    const Inplace& initialInplace() const
    {
        return this->initialInplace_.value();
    }
```

GitPaean:
Since `FlowProblemComp` does not calculate the `initialInplace_` yet, so maybe the **easier** solution is to let evalSummary function either use `std::optional` or `pointer` as argument instead of `                    const Inplace&                                       initialInPlace,`. 

I would like some comments from @akva2 before I go ahead.

akva2:
right. yes, I think initialInPlace should return the optional, and the function should take an optional, and deal accordingly on the inside.

opm/simulators/flow/GenericOutputBlackoilModule.cpp
```diff
@@ -263,57 +263,6 @@ calc_inplace(std::map<std::string, double>& miscSummaryData,
     return inplace;
 }
 
-template<class FluidSystem>
-void GenericOutputBlackoilModule<FluidSystem>::
-outputFipAndResvLog(const Inplace& inplace,
```
GitPaean, line 268: this function assume that `this->initialInplace()` always have value, so it is moved to outputBlackoilModule.hpp.

GitPaean:
jenkins build this please

GitPaean:
I am not deep in understanding the usage of the reference to optional `std::optinal<T>&` , any comments regarding this. 

In the middle of mixing, was thinking to turn to pointer.

bska:
> I am not deep in understanding the usage of the reference to optional `std::optinal<T>&`

Any chance you could split that part out to a separate PR?  That would be easier to review/test than when it's embedded in the work that introduces compositional well models.

GitPaean:
> > I am not deep in understanding the usage of the reference to optional `std::optinal<T>&`
> 
> Any chance you could split that part out to a separate PR? That would be easier to review/test than when it's embedded in the work that introduces compositional well models.

I can do that.

GitPaean:
This PR requires https://github.com/OPM/opm-common/pull/4552 and https://github.com/OPM/opm-simulators/pull/6137 to be able to run. So I am making as draft for now.

GitPaean:
The dependent PRs have been merged. I am marking this PR as ready for review.

GitPaean:
> I notice that none of the new header files are listed in `CMakeLists_files.cmake`. Is that intentional?

I think it is just something we did not do it. I am adding all the .cpp and .hpp under flowexperimental/ to `CMakeLists_files.cmake` now.

GitPaean:
> I notice that none of the new header files are listed in `CMakeLists_files.cmake`. Is that intentional?

I just added them, let me know if you want some different organization.

GitPaean:
jenkins build this please

bska (commented):
>> I notice that none of the new header files are listed in CMakeLists_files.cmake. Is that intentional?
>
> I just added them, let me know if you want some different organization.

Thanks!  Maybe we could add these files to the [`MAIN_SOURCE_FILES`](https://github.com/OPM/opm-simulators/blob/292f571d2207668642e4841b9727a93628461c88/CMakeLists_files.cmake#L57) list instead?  We might not want to install them at this time, but it's nevertheless a good idea, I think, to have them in the file list as that makes it easier for those who use IDEs.

GitPaean:
Sure. I have done that.

bska (approved):
Thanks a lot for the update and for your patience in getting this through PR review.  I'll merge into master now.

PR #6135, opened 2025-04-02
BUGIFIX: Only respect item 7 in GCONPROD for individual constrains,

totto82:
i.e. allow for switching to FLD if item 8 is yes

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelConstraints.cpp
+++ b/opm/simulators/wells/BlackoilWellModelConstraints.cpp
@@ -463,10 +463,25 @@ actionOnBrokenConstraints(const Group& group,
                           DeferredLogger& deferred_logger) const
 {
 
-    bool changed = false;
     const Group::ProductionCMode oldControl =
         wellModel_.groupState().production_control(group.name());
 
+    // We switch to higher groups independently of the given group limit action in GCONPROD item 7
+    if (newControl == Group::ProductionCMode::FLD && oldControl != Group::ProductionCMode::FLD) {
+        // If newControl is FLD, the group should be subject to higher order controls
+        assert(group.productionGroupControlAvailable());
+        group_state.production_control(group.name(), newControl);
+        if (wellModel_.comm().rank() == 0) {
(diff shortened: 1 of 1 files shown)


opm/simulators/wells/BlackoilWellModelConstraints.cpp
```diff
@@ -476,11 +476,21 @@ actionOnBrokenConstraints(const Group& group,
...
         wellModel_.groupState().production_control(group.name());
-
     std::string ss;
+    if (newControl == Group::ProductionCMode::FLD && oldControl != Group::ProductionCMode::FLD) {
+        group_state.production_control(group.name(), newControl);
+        ss = fmt::format("Switching production control mode for group {} from {} to {}",
+        group.name(),
+        Group::ProductionCMode2String(oldControl),
+        Group::ProductionCMode2String(newControl));
+        if (!ss.empty() && wellModel_.comm().rank() == 0)
```
bska, line 488: When would `ss` ever be `.empty()` in this context?  As far as I can tell, this condition is equivalent to `rank() == 0`, so you might as well write this as
```C++
if (newControl ...) {
    if (rank == 0) {
        deferred_logger.debug(fmt::format(...));
    }

    return true;
}
```
in which case the `ss` object need not be introduced until the "exceed action" switch below.

I also think I'd add a comment to this block to explain why the `FLD` handling is a special case.  This, I think, will aid future maintainers.

bska (commented):
As far as the logic goes, this is fine.  I think I'd prefer if the `FLD` handling context provided a little explanatory documentation though.

opm/simulators/wells/BlackoilWellModelConstraints.cpp
totto82: Thanks for the review. I have updated the code accordingly.

totto82:
jenkins build this please

bska:
jenkins build this please

bska (approved):
Thanks a lot for the updates. This looks good to me now and I'll merge into master.

PR #6137, opened 2025-04-02
using std::optional for function initialInplace()

GitPaean:


GitPaean:
it depends on https://github.com/OPM/opm-common/pull/4552

bska:
jenkins build this opm-common=4552 please

bska (commented):
This looks fine, but just out of curiosity: Why do you move the `outputFipAndResvLog()` function from the generic to the black-oil specific output module?  It's been there since it was renamed from `outputFipLog()`, and made more general, in #5068.  That PR also, incidentally, made the "initial in-place" volume calculation separately callable.

GitPaean:
> This looks fine, but just out of curiosity: Why do you move the `outputFipAndResvLog()` function from the generic to the black-oil specific output module? It's been there since it was renamed from `outputFipLog()`, and made more general, in #5068. 

Because that function assumes that `this->initialInplace()` is valid or has value, I do not know how to modify the function for the situation that we did not have `this->initialInplace()` available.

GitPaean:
> That PR also, incidentally, made the "initial in-place" volume calculation separately callable.

I am not following this comment.

bska:
>>  Why do you move the `outputFipAndResvLog()` function from the generic to the black-oil specific output module?
>
> Because that function assumes that `this->initialInplace()` is valid or has value, I do not know how to modify the function for the situation that we did not have `this->initialInplace()` available.

Right, okay, that makes a bit of sense.  Default-constructed `Inplace` objects aren't particularly useful.  I suppose you could return early and not write anything if the initial in-place volumes don't exist, but that is more or less an equally poor hack.

>> That PR also, incidentally, made the "initial in-place" volume calculation separately callable.
>
> I am not following this comment.

I was merely pointing out, as an aside, that PR #5068 split the in-place volume calculations out to a separately callable function, mostly for the purpose of computing the initial in-place volumes.  Prior to that PR, the log-output function would itself compute the in-place volumes and return those values to the caller.

bska (approved):
With justification provided in the PR discussion I think this is okay.  We may have to push the balance sheet reporting function back into the generic output module at some point in the future, especially when we have a way of computing in-place volumes for compositional runs, but for now I accept that this function is black-oil only.

I'll merge into master.

PR #6143, opened 2025-04-03
Fix compositional field output names

akva2:
And avoid spurious water outputs


GitPaean:
jenkins build this please

GitPaean:
The testing results look good. But it is not in the area I understand well, so I will request @bska to review the PR.

PR #6145, opened 2025-04-04
Log local switching

totto82:
In the master, we don't log control switching during local solves, which could lead to changing controls without or with the wrong output in the terminal.  The log also detects oscillations, and controls may oscillate significantly during the local solves. I added a safeguard to restrict only changing controls for iter>nupcol and the local solves, but we may need a more robust solution. I tagged this with release since it is a bug fix, but I also made it a draft as it requires more testing. 

UPDATE: After some thinking and testing, I suggest starting to count oscillations only after NUPCOL iterations. After NUPCOL iterations, the control target is fixed, and we should, in principle, experience fewer oscillations. With this change, we less frequently end up with the "wrong" control, while the "cost" of stagnating simulations due to infinite oscillations is still avoided. The current issue in master that we don't log control changes during local switching (leading to missing log information to terminal and .PRT) is still fixed in the PR.

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -209,21 +209,20 @@ namespace Opm
         } else {
             from = WellProducerCMode2String(ws.production_cmode);
         }
-        bool oscillating = std::count(this->well_control_log_.begin(), this->well_control_log_.end(), from) >= this->param_.max_number_of_well_switches_;
+
         const int episodeIdx = simulator.episodeIndex();
         const int iterationIdx = simulator.model().newtonMethod().numIterations();
         const int nupcol = schedule[episodeIdx].nupcol();
-        if (oscillating && iterationIdx > nupcol) {
-            // only output frist time
-            bool output = std::count(this->well_control_log_.begin(), this->well_control_log_.end(), from) == this->param_.max_number_of_well_switches_;
+        const bool oscillating = std::count(this->well_control_log_.begin(), this->well_control_log_.end(), from) >= this->param_.max_number_of_well_switches_;
+        if (oscillating) {
+            // only output first time
+            const bool output = std::count(this->well_control_log_.begin(), this->well_control_log_.end(), from) == this->param_.max_number_of_well_switches_;
             if (output) {
-                std::ostringstream ss;
-                ss << "    The control mode for well " << this->name()
-                   << " is oscillating\n"
-                   << "    We don't allow for more than "
-                   << this->param_.max_number_of_well_switches_
-                   << " switches. The control is kept at " << from;
-                deferred_logger.info(ss.str());
+                const auto msg = fmt::format("    The control mode for well {} is oscillating. \n"
+                    "We don't allow for more than {} switches after NUPCOL iterations. (NUPCOL = {}) \n"
+                    "The control is kept at {}.",
+                    this->name(), this->param_.max_number_of_well_switches_, nupcol, from);
+                deferred_logger.info(msg);
                 // add one more to avoid outputting the same info again
                 this->well_control_log_.push_back(from);
             }
@@ -256,7 +255,12 @@ namespace Opm
             }
             deferred_logger.debug(ss.str());
 
-            this->well_control_log_.push_back(from);
+            // We always store the current control as it is used for output
+            // and only after iteration >= nupcol
+            // we log all switches to check if the well controls oscillates
+            if (iterationIdx >= nupcol || this->well_control_log_.empty()) {
+                this->well_control_log_.push_back(from);
+            }
             updateWellStateWithTarget(simulator, group_state, well_state, deferred_logger);
             updatePrimaryVariables(simulator, well_state, deferred_logger);
         }
@@ -288,7 +292,6 @@ namespace Opm
             from = WellProducerCMode2String(ws.production_cmode);
         }
         const bool oscillating = std::count(this->well_control_log_.begin(), this->well_control_log_.end(), from) >= this->param_.max_number_of_well_switches_;
-
         if (oscillating || this->wellUnderZeroRateTarget(simulator, well_state, deferred_logger) || !(this->well_ecl_.getStatus() == WellStatus::OPEN)) {
            return false;
         }
@@ -508,6 +511,9 @@ namespace Opm
         const auto& summary_state = simulator.vanguard().summaryState();
         const auto inj_controls = this->well_ecl_.isInjector() ? this->well_ecl_.injectionControls(summary_state) : Well::InjectionControls(0);
         const auto prod_controls = this->well_ecl_.isProducer() ? this->well_ecl_.productionControls(summary_state) : Well::ProductionControls(0);
+        const auto& ws = well_state.well(this->indexOfWell());
+        const auto pmode_orig = ws.production_cmode;
+        const auto imode_orig = ws.injection_cmode;
         bool converged = false;
         try {
             // TODO: the following two functions will be refactored to be one to reduce the code duplication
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -288,8 +288,10 @@ namespace Opm
             from = WellProducerCMode2String(ws.production_cmode);
         }
         const bool oscillating = std::count(this->well_control_log_.begin(), this->well_control_log_.end(), from) >= this->param_.max_number_of_well_switches_;
-
-        if (oscillating || this->wellUnderZeroRateTarget(simulator, well_state, deferred_logger) || !(this->well_ecl_.getStatus() == WellStatus::OPEN)) {
+        const int iterationIdx = simulator.model().newtonMethod().numIterations();
+        const int episodeIdx = simulator.episodeIndex();
+        const int nupcol = schedule[episodeIdx].nupcol();
+        if ( (oscillating && iterationIdx > nupcol)|| this->wellUnderZeroRateTarget(simulator, well_state, deferred_logger) || !(this->well_ecl_.getStatus() == WellStatus::OPEN)) {
```
vkip, line 294: >= nupcol?

totto82: In updateWellControl(..) it is iterationIdx >nupcol. This makes it consistent. But I don't know if waiting one iteration is on purpose or not.

vkip: You mean what was introduced in https://github.com/OPM/opm-simulators/pull/5753? Since we start counting iterations from 0 I believe we should use \>= unless there is a clear (and documented) reason to do so.

totto82: Since I made #5773, I assume > is a bug, I will update both instances to >= and retest.

vkip: Thanks. Can be discussed if it is a bug, but prior to https://github.com/OPM/opm-simulators/pull/6100 it was a mix between \>= and \>, and better to be consistent. Apparently I missed the one in updateWellControls in that PR..

totto82:

totto82:

totto82:
I don't think the logging issue fixed in this PR is severe enough to delay the release, so I removed the release flag.

totto82:

totto82:

GitPaean:
It looks like [mpi.compareECLFiles_flow+WVFPEXP-02](https://ci.opm-project.org/job/opm-simulators-PR-builder/7935/testReport/junit/(root)/mpi/compareECLFiles_flow_WVFPEXP_02/) needs some attention and investigation. 

image

totto82:
> It looks like [mpi.compareECLFiles_flow+WVFPEXP-02](https://ci.opm-project.org/job/opm-simulators-PR-builder/7935/testReport/junit/(root)/mpi/compareECLFiles_flow_WVFPEXP_02/) needs some attention and investigation

OP-1 switches to GRUP in master the 1 Jly 2025. This is due to oscillation in the master and looks wrong.  

image


.

GitPaean:

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -287,8 +293,15 @@ namespace Opm
         } else {
             from = WellProducerCMode2String(ws.production_cmode);
         }
+        // Only keep first and last control until iterationIdx >= nupcol (These are used for output)
+        // After iteration >= nupcol we log all switches to check if the well controls oscillates
+        const int episodeIdx = simulator.episodeIndex();
+        const int iterationIdx = simulator.model().newtonMethod().numIterations();
+        const int nupcol = schedule[episodeIdx].nupcol();
+        if (iterationIdx < nupcol && this->well_control_log_.size() > 2) {
```
GitPaean, line 301: From the code, during the local solve, we only update the `well_controls_log_`  after the local solve getting converged. Do we need to do this `erase` procedure during local solve here, will it conflict with the `erase` from the outer iteration?  Let me know if it happens that I read the code wrong.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -209,11 +209,17 @@ namespace Opm
...
         }
-        bool oscillating = std::count(this->well_control_log_.begin(), this->well_control_log_.end(), from) >= this->param_.max_number_of_well_switches_;
+        // Only keep first and last control until iterationIdx >= nupcol (These are used for output)
+        // After iteration >= nupcol we log all switches to check if the well controls oscillates
         const int episodeIdx = simulator.episodeIndex();
         const int iterationIdx = simulator.model().newtonMethod().numIterations();
         const int nupcol = schedule[episodeIdx].nupcol();
-        if (oscillating && iterationIdx > nupcol) {
+        if (iterationIdx < nupcol && this->well_control_log_.size() > 2) {
+            this->well_control_log_.erase (this->well_control_log_.begin()+1,this->well_control_log_.end()-1);
```
GitPaean, line 218: space between arguments.

GitPaean (commented):
I am still trying to understand the effects for this change.  Also leaving some comments for discussion/addressing.

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -209,11 +209,17 @@ namespace Opm
...
+        }
+
+        const bool oscillating = std::count(this->well_control_log_.begin(), this->well_control_log_.end(), from) >= this->param_.max_number_of_well_switches_;
+        if (oscillating) {
             // only output frist time
```
GitPaean, line 223: not part of this PR, but since you are editing here.  `frist` should be first.  `bool output` can be `const`.

GitPaean:
If I understand the PR correctly, it is mostly to postpone the oscillation checking after the iteration reached NUPCOL, potentially we might get more correct results for certain scenarios, that is fine, while whether it is done in a good way remains to be discussed. 

For example, 
```c++
        if (iterationIdx < nupcol && this->well_control_log_.size() > 2) {
            this->well_control_log_.erase (this->well_control_log_.begin()+1,this->well_control_log_.end()-1);
        }
```
basically, we keep discarding the elements in the `well_control_log_` until the iteration reaches `NUPCOL`, what is the difference if we do not add element to the `well_control_log_` before the iteration reaches `NUPCOL`. 

maybe for the local solve, we can passing the oscillating flag in instead of the `erase` and determination of the `oscillating` flag there also. 

I might read it wrong, but let us have some discussion here.

totto82:
Thanks for the review. I agree with most of them and will update the code accordingly. But what is it that you want to discuss?

> but let us have some discussion here.

The main issue with the code in master that this PR addresses is that the well control can change during local switching without the user getting any message. If we naively add the local switches to the logging, we will increase the number of switches considerably, and the current way we check for oscillations does not work very well anymore. Erasing part of the log (just keeping the first and last for output) for iter < NUPCOL is a way to minimize cases where we end up with the wrong control due to heavy oscillations during the first iterations.

GitPaean:
Thanks for the explanation. 

> The main issue with the code in master that this PR addresses is that the well control can change during local switching without the user getting any message. 

This PR indeed adding one extra element to the `well_control_log_` **after** the `iterateWellEqWithSwitching` get converged.  Previously we only output a DEUBG message.  It does not add anything during the local solve iterations if I read the content correctly.  Do we need to             `this->well_control_log_.erase(this->well_control_log_.begin() + 1, this->well_control_log_.end() - 1);` during this local iterations? since we are doing it in the global iteration with `updateWellControl` anyway?

Further more, 
```c++
        if (iterationIdx < nupcol && this->well_control_log_.size() > 2) {
            this->well_control_log_.erase (this->well_control_log_.begin()+1,this->well_control_log_.end()-1);
        }

        const bool oscillating = std::count(this->well_control_log_.begin(), this->well_control_log_.end(), from) >= this->param_.max_number_of_well_switches_;
```

Instead of keeping erasing, should we avoid adding more when `iterationIdx < nupcol ` ?

GitPaean:
> > Instead of keeping erasing, should we avoid adding more when iterationIdx < nupcol  ?
> 
> That is possible. If we add the current control at the beginning of the timestep, we can avoid the erase. I can update the PR accordingly if you like.

not totally sure about how you gonna do here, while it sounds right. The main point is that keeping appending and erasing is a more indirect way to do things, I believe some more explicit/direct code  can achieve the same and result in more readable code.  Referring to the comments above, it is **possible** that during local solve switching, we might not need to do the same with the outer iteration, which will reduce complexity, while you are the designer of the content of the PR, you are more familiar with what you want to achieve with all the code and possibly I can miss something from the code level.

totto82:

totto82:
@GitPaean I have now addressed all your comments.

GitPaean:
> @GitPaean I have now addressed all your comments.

Thanks for the update, the results looks good and less regressions. 

Sorry I did not understand your previous comments, with the change you made by passing in the initial control, do you think the following code can achieve the same by much simpler manner?

```c++
            if (iterationIdx >= nupcol || this->well_control_log_.empty()) {
                this->well_control_log_.push_back(from);
            }
```

totto82:
> Sorry I did not understand your previous comments, with the change you made by passing in the initial control, do you think the following code can achieve the same by much simpler manner?

Smart. I will update accordingly.

totto82:

GitPaean:
> > Sorry I did not understand your previous comments, with the change you made by passing in the initial control, do you think the following code can achieve the same by much simpler manner?
> 
> Smart. I will update accordingly.

If we go back to your earlier version, you might want to record the last control change when `iterationIdx < nupcol `, 

```c++
            // when iterationIdx < nupcol, keep only the initial and last control mode
           // when iterationIdx >= nupcol, keep all the control switching
            if (iterationIdx >= nupcol || this->well_control_log_.size() < 2) {
                this->well_control_log_.push_back(from);
            } else if (this->well_control_log_.size() == 2) {
                this->well_control_log_[1] = from;
           }
```
Sorry it was not trivial to follow what the PR is exactly about.  and it is up to you to decide.

GitPaean:
> > you might want to record the last control change
> 
> We don't need that since we can ask the wellstate what control we are currently on.

Oh, yeah, then we do not need it.

GitPaean (approved):
Thanks for the update the patience. It looks good to me now and I approve it. We will wait when we can update the reference to get the PR in.

GitPaean:

GitPaean:
jenkins build this opm-tests=1336 please

PR #6120, opened 2025-03-28
dont call beginTimeStep when restarting simulation after failing

totto82:
We do not want to redo the well testing, the pre-solving of wells, gas-lift optimization, network balancing, etc, after failed simulations

For now, this PR adds an if around the outer beginTimeStep. We may need a more fine-grained approach. However, initial testing on selected models is promising.

PR #6146, opened 2025-04-04
 Regularize wells where linear system solve fails

totto82:
Fixes issues related to MSW_2D_VERT_02 and https://github.com/OPM/opm-simulators/pull/6120
But may have positive effects on other problematic cases. 
Needs more testing.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -596,6 +596,7 @@ namespace Opm
             // this is the process with rank zero)
             deferred_logger.problem("In MultisegmentWell::solveEqAndUpdateWellState for well "
                                     + this->name() +": "+exp.what());
+            this->regularize_ = true;
             throw;
         }
     }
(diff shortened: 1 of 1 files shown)


totto82:

GitPaean:
And do you understand why the issue was caused in https://github.com/OPM/opm-simulators/pull/6120 ?

totto82:
> And do you understand why the issue was caused in https://github.com/OPM/opm-simulators/pull/6120 ?

The issue is related to singular matrices and MSW, so it is not directly related to 6120, but I noticed it while working on the PR.

totto82:
The test failure is related to MSW_2D_VERT_02. By regularizing, we can simulate this model without cutting the timestep.

totto82:

totto82:
> Should we target that symptom also/instead?
Sure. But I think we will not be able to fix all cases with a quick fix any time soon. The code will still throw, so the issue will not be hidden. But at least with this, we have a better chance of recovering.

GitPaean:
> > Should we target that symptom also/instead?
> 
> Sure. But I think we will not be able to fix all cases with a quick fix any time soon. The code will still throw, so the issue will not be hidden. But at least with this, we have a better chance of recovering.

I meant, should we concern that if one well_state (scenario) causing numerical problem, should we reset it somehow instead of using the same problematic well_state again?

PR #6150, opened 2025-04-08
[testing] using bigger epislon for episodeWillBeOver() function

GitPaean:


GitPaean:
jenkins build this please

GitPaean:
There are many ways to adjust this part, the current problem is that for double, (1. - eps) is too small to make the assumption 
```c++
        return this->time() + timeStepSize()
            >=  (episodeStartTime_ - startTime() + episodeLength())*(1 - eps);
```
correct. 

For example, the reference `this->time() + timeStepSize()` is `86399.999`, the `(episodeStartTime_ - startTime() + episodeLength()` is `86399.9990000725`, 
and the comparison value `((episodeStartTime_ - startTime() + episodeLength() )*(1 - eps)` is `86399.9990000533`. 

The intention is probably that the episode should be over, but ` (episodeStartTime_ - startTime() + episodeLength())*(1 - eps)` involves multiple operations, the rounding error is bigger than the `(1 - eps)` can eliminate.

GitPaean:
The original finding to motivate looking at this direction is because the regression test with `GASCONDENSATE_VAPWAT_PRECSALT_REGRESSION` only output summary and UNRST file for one report step despite that there are five report steps. 

The jenkins testing for this PR only reveal that we have the same issue with the `[BO_DIFFUSE_CASE1](https://ci.opm-project.org/job/opm-simulators-PR-builder/7846/testReport/junit/(root)/mpi/compareECLFiles_flow_BO_DIFFUSE_CASE1/)`  The report step 2 missing the output in the reference in the opm-tests.

GitPaean:
and the following function looks wrong, 
```c++
    bool finished() const
    {
        assert(timeStepSize_ >= 0.0);
        Scalar eps =
            std::max(Scalar(std::abs(this->time())), timeStepSize())
            *std::numeric_limits<Scalar>::epsilon()*1e3;
        return finished_ || (this->time()*(1.0 + eps) >= endTime());
    }
```

the `eps` here is related to `std::max(Scalar(std::abs(this->time())), timeStepSize())` and at a later stage it is used to scale the `this->time()`.

GitPaean:
https://github.com/OPM/opm-simulators/pull/6151 has been merged. We do not need this PR anymore.

PR #6153, opened 2025-04-08
Use Correct Array Name for RVSAT Data

bska:
The copy-pasted RSSAT name prevented outputting RVSAT data in runs requesting both of these arrays.

This is a bugfix which should go into the 2025.04 release, so I have tagged the PR with that milestone.


akva2 (approved):

bska:

PR #6156, opened 2025-04-09
Fix to 2025.04-r1 build candidate in clang

daavid00:
While testing 2025.04-r1, this issue appears, and this PR fixes this.


daavid00:
jenkins build this please

multitalentloes:
Maybe I am misreading your PR description, but I do not catch what issue you had to begin with that this resolves?

daavid00:
I am doing the testing that you requested in the email, and I use macOS (clang), then while building the 2025.04-r1 I got an error regarding the definition of `reservoir_coupling_master_`, which this PR fixes (by including the check of `RESERVOIR_COUPLING_ENABLED` as is done in other parts in the code).

PR #6159, opened 2025-04-10
Backport of pr 6104

multitalentloes:
backporting ahead of RC2


multitalentloes:

PR #6160, opened 2025-04-10
removing something not used or not defined

GitPaean:


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

atgeirr:
Looks good to me, but since this removes a somewhat large function related to gas lift, perhaps @totto82 could clarify if that function is still useful to keep around, for future needs? If not, then merge.

totto82:
The gas lift code was used for the well test; now, we do a more proper gas lift optimization during the well test. I.e., there is no reason to keep the code.

bska (approved):
> there is no reason to keep the [unused GLO function]

In that case I'll merge this into the master branch.

PR #6162, opened 2025-04-10
Fix to 2025.04-r1 build candidate in clang

multitalentloes:
Backporting ahead of RC2


multitalentloes:

PR #6166, opened 2025-04-10
Backport of pr 6151

multitalentloes:
Backporting ahead of rc2


multitalentloes:

PR #6167, opened 2025-04-10
moving flowPhaseToModelPhaseIdx to WellInterfaceIndices

GitPaean:
so it will be together with other index conversion functions. it is not a significant deal, while it makes easier to find them. 

originally, I thought there are only two of them since WellInterfaceIndices class only has two of them.


GitPaean:
Since this is a trival change, I am self-merging it to the master to ease the current investigation.

PR #6170, opened 2025-04-11
Reservoir coupling: Communicate group potentials from slaves and compute guide rates

hakonhagland:
Depends on https://github.com/OPM/opm-common/pull/4567

Sends group potentials from the slave processes to the master process, then use those potentials in the master process to calculate guide rates.

A future PR will calculate group target in the master based on the guide rates and send those targets to the slave groups.


hakonhagland:
I will run jenkins from the opm-common PR later

opm/simulators/flow/ReservoirCouplingSlave.cpp
```diff
@@ -268,9 +300,64 @@ saveMasterGroupNamesAsMap_(const std::vector<char>& group_names) {
...
     }
 }
 
+void
+ReservoirCouplingSlave::
+sendActivationDateToMasterProcess_() const
+{
+    if (this->comm_.rank() == 0) {
+        // NOTE: The master process will use this date to check that no slave starts before the master
+        double activation_date = this->getGrupSlavActivationDate_();
```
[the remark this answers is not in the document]
hakonhagland: This was intentionally named GrupSlav (and not GrupSlave) because the keyword name is `GRUPSLAV`.

opm/simulators/flow/ReservoirCouplingSlave.cpp
```diff
@@ -256,6 +209,85 @@ maybeActivate(int report_step) {
...
+        );
+        OpmLog::info("Received master group names from master process rank 0");
+    }
+    this->comm_.broadcast(&size, /*count=*/1, /*emitter_rank=*/0);
+    if (this->comm_.rank() != 0) {
+        group_names.resize(size);
+    }
+    this->comm_.broadcast(group_names.data(), /*count=*/size, /*emitter_rank=*/0);
+    this->saveMasterGroupNamesAsMap_(group_names);
+    this->checkGrupSlavGroupNames_();
```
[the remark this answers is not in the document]
hakonhagland: Same reason as for the previous comment.

blattms (commented):
Thanks a lot for the contribution. A few comments for now.

Note that this does not build without MPI.

blattms:
Unfortunately, this needs to be rebased on master it seems.

hakonhagland:
@blattms Thanks for the review with useful suggestions. I have added some commits to the PR to address the comments. I also added one commit to dump all guide rates to the debug file.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -457,23 +462,21 @@ namespace Opm {
...
-                                                   pot,
-                                                   local_deferredLogger);
+        this->guide_rate_handler_.updateGuideRates(
+            reportStepIdx, simulationTime, this->wellState(), this->groupState()
+        );
+#ifdef RESERVOIR_COUPLING_ENABLED
+        if (this->isReservoirCouplingSlave()) {
+            this->guide_rate_handler_.sendSlaveGroupPotentialsToMaster();
+        }
+#endif
```
blattms: This is probably a silly question: Why don't we do the receive and sends in GuideRateHandler::updateGuideRates?

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1271,22 +1277,9 @@ namespace Opm {
...
-                                                       simulationTime,
-                                                       this->wellState(),
-                                                       this->groupState(),
-                                                       comm,
-                                                       &this->guideRate_,
-                                                       pot,
-                                                       local_deferredLogger);
+            this->guide_rate_handler_.updateGuideRates(
+                reportStepIdx, simulationTime, this->wellState(), this->groupState()
+            );
```
blattms, line 1293: Maybe connected to the other comment:
There is no communication here. Why is this not needed? Please add a comment about this.

opm/simulators/flow/ReservoirCoupling.hpp
```diff
@@ -19,27 +19,76 @@
...
-    MasterGroupNames,
-    MasterGroupNamesSize,
 };
 
+// Used to communicate potentials for oil, gas, and water rates between slave and master processes
+struct Potentials {
+    // Define a num_fields constant to be used in MPI communication to simplify serialization
+    // and deserialization of the struct. Note that sizeof(Potentials)/sizeof(double) is not
+    // guaranteed to be equal to num_fields due to padding.
+    static constexpr std::size_t num_fields = 3;
```
blattms: This does not seem to be used. I do not really get the comment. What is it for?

opm/simulators/timestepping/AdaptiveTimeStepping.hpp
```diff
@@ -125,6 +127,7 @@ class AdaptiveTimeStepping
         Solver& solver_;
         const bool is_event_;
         const TuningUpdateCallback& tuning_updater_;
+        Simulator& simulator_;
```
blattms: Do we really need a full reference to the simulator? It only seems to be used to check a few states (are we master, slave or active). I fear that this will get abused by fellow developers.

Do you think it is possible to get rid of the reference? By maybe using few bools or a slim class that has only the needed methods (simulator could inherit from that).

blattms (commented):
Thanks  a lot for rebasing.
This is looks quite good already.

My main reservation is the introduction of a reference to the simulator in the time stepping. I would hope that there is another way.

blattms:

blattms:
:see_no_evil: jenkins will fail of course. Handled by opm-common PR, already

opm/simulators/timestepping/AdaptiveTimeStepping.hpp
hakonhagland: I did not think about that. Since the user can anyway abuse simulator by going through `solver.model().simulator()` and we need the solver reference anyway, storing the simulator is just a convenience for writing `this->simulator_` instead of `this->solver_.model().simulator()`, right? See, the added commit.

opm/simulators/flow/ReservoirCoupling.hpp
hakonhagland: Yes the comment was not clear. It was used by the `ReservoirCouplingMpiTraits.hpp` to determine the block lengths for the MPI trait for the `Potentials` struct. However, to avoid duplication of code when developing MPI traits for other structs in the future, I decided to make the code more generic. Luckily, as a by-product of that change, the need for the `num_fields` variable was also eliminated. See the added commit for details.

opm/simulators/wells/BlackoilWellModel_impl.hpp
hakonhagland: Good question. Since `updateGuideRates()` is not directly dependent on sending slave group potentials and updating master group potentials. Rather, it will compute the potentials from the well potentials and store those values in the `GuideRate` object (for non-master groups), and for master groups it will retrieve the group potentials stored in the `ReservoirCouplingMaster` object to compute guide rates. See `updateGuideRates()` in `updateWellControlsAndNetworkIteration()` for an example where sending slave group potentials/receiving master group potentials is not done. So the slave group potentials only need to be communicated at the start of the timestep and no update of those will be done during the nonlinear iterations.

opm/simulators/wells/BlackoilWellModel_impl.hpp
hakonhagland: Good observation. I tried to explain this in the [other comment](https://github.com/OPM/opm-simulators/pull/6170#discussion_r2141803109). Also added a comment in the source code, see latest commit.

hakonhagland:
> My main reservation is the introduction of a reference to the simulator in the time stepping.

@blattms Thanks for the detailed review. I have added some commits to address the comments you posed. I am also planning to add a commit to address [the comment](https://github.com/OPM/opm-common/pull/4567#issuecomment-2963363992) from @bska in the companion PR in opm-common. I think we can avoid getting the slave group potentials from the `GuideRate` object by storing them in the `GroupState` object instead. This will for now cause duplication of group potentials in `GuideRate` and `GroupState`. A future PR might try to remove that duplication by removing the potentials from the `GuideRate` object's internal state. What do you think? Then we can simply close/delete https://github.com/OPM/opm-common/pull/4567.

hakonhagland:
jenkins build this please

hakonhagland:
jenkins build this please

hakonhagland:
jenkins build this please

hakonhagland:
> The last commit tries to address OPM/opm-common#4567 (comment). Then OPM/opm-common#4629 will replace OPM/opm-common#4567.

@blattms and @bska After review by @bska, I think the companion PR https://github.com/OPM/opm-common/pull/4629 can also be eliminated. I have added a [commit](https://github.com/OPM/opm-simulators/pull/6170/commits/9823ebc6ffe52795afc1cf79b970c359e1a2dde9) here as a replacement, and plan to close both opm-common PRs as they are no longer needed. Jenkins builds with and without serial succeed. Let me know if anything else needs to be added to this PR.

blattms:
Thanks a lot. I'll take another look.

opm/simulators/wells/GroupState.cpp
```diff
@@ -509,6 +509,23 @@ gconsump_rates(const std::string& gname) const {
...
 }
 
+template<class Scalar>
+void GroupState<Scalar>::
+update_group_production_potential(const std::string& gname, Scalar oil_rate,
+                                  Scalar gas_rate, Scalar water_rate)
+{
+    this->production_group_potentials.emplace(
+        gname, GroupPotential{oil_rate, gas_rate, water_rate}
+    );
```
blattms: This confuses me a bit. The method name includes "update" which seems to hint that existing value should be overwritten. `std::map::emplace` on the other hand will only insert values if the key does not yet exist in the map.

opm/simulators/timestepping/AdaptiveTimeStepping.hpp
```diff
@@ -78,6 +78,8 @@ class AdaptiveTimeStepping
 
 private:
     using Scalar = GetPropType<TypeTag, Properties::Scalar>;
+    using Simulator = GetPropType<TypeTag, Properties::Simulator>;
```
blattms: Is this used?

opm/simulators/flow/ReservoirCouplingMpiTraits.hpp
```diff
@@ -0,0 +1,154 @@
...
+
+        MPI_Get_address(&(d.*Member), &disp[idx]);
+        disp[idx] -= base;
+
+        if constexpr (std::is_array_v<MemberT>) {
+            // C array  T[N]
+            using Elem = std::remove_extent_t<MemberT>;
+            blk  [idx] = std::extent_v<MemberT>;
+            types[idx] = MPITraits<Elem>::getType();
+        }
```
blattms, line 128: This should also work if T is not an array and give the same results as the else branch (you still need to handle num). Can we simplify this?

What branches are actually used? This looks quite general, but the surrouding class is not. Is this meant for further usage in the future?

blattms (commented):
Overall this looks good. Just a few curious questions.

Personally, I would appreciate a bit more doxygen documentation for new classes.

opm/simulators/flow/ReservoirCouplingMpiTraits.hpp
hakonhagland: Yes this is meant for to be used for other structs in future PRs. Currently only the `is_std_array<MemberT>::value` branch is used. While we could technically collapse the scalar/enum and C-array cases using more conditional logic, I believe the current structure is more readable.

opm/simulators/timestepping/AdaptiveTimeStepping.hpp
hakonhagland: Good catch. I added a commit that removes it.

opm/simulators/wells/GroupState.cpp
hakonhagland: Good catch! I have added a commit that uses `try_emplace()` instead of `emplace()` here

PR #6176, opened 2025-04-14
Add two tests for distributed MSWells with only shut perforations on one rank

lisajulia:

PR #6177, opened 2025-04-14
Fix runs with distributed MSWells with only shut perforations on one process

lisajulia:
Fixes the tests added in https://github.com/OPM/opm-tests/pull/1330.
Without the fix from this PR, the tests fail (see #6176).


lisajulia:
jenkins build this opm-tests=1330 please

PR #6178, opened 2025-04-14
Cut timestep if network imbalance is larger than a parameter

totto82:
Accepting an unconverged network solution could lead to premature shutting of wells and it is therefore sometimes better to re-try the timestep. It is still unclear what a good default value is (maybe 1 bar?) As a starting point I suggest to add the parameter with a very large default value to make the option available.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -1209,6 +1209,15 @@ namespace Opm {
                         max_iteration, network_imbalance*1.0e-5, well_group_control_changed);
                     local_deferredLogger.debug(msg);
                 } else {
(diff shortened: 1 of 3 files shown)


totto82:

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -1209,6 +1209,16 @@ namespace Opm {
                         max_iteration, network_imbalance*1.0e-5, well_group_control_changed);
                     local_deferredLogger.debug(msg);
                 } else {
+                    const auto& balance = this->schedule()[episodeIdx].network_balance();
+                    const Scalar maximum_imbalance = param_.network_max_pressure_imbalance_in_bars_;
+                    if (network_imbalance > maximum_imbalance) {
```
totto82, line 1214: Ops. I forgot to add the *Bars. Will update later.

totto82:

GitPaean:

PR #6181, opened 2025-04-14
Summary vectors for water mass

daavid00:
Depends on https://github.com/OPM/opm-common/pull/4569

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -1764,9 +1764,11 @@ namespace Opm
         const bool allow_cf = this->getAllowCrossFlow() || openCrossFlowAvoidSingularity(simulator);
 
         const int nseg = this->numberOfSegments();
-
-        const Scalar rhow = FluidSystem::referenceDensity( FluidSystem::waterPhaseIdx, Base::pvtRegionIdx() );
-        const unsigned watCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::waterCompIdx);
+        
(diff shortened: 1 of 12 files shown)


totto82 (approved):
Thanks. Looks good.

bska (commented):
I don't remember the exact data flow here so I'll just ask: Seeing as you make changes to make the standard well aware of the new rate, does this work for multi-segmented wells too?

daavid00:
> I don't remember the exact data flow here so I'll just ask: Seeing as you make changes to make the standard well aware of the new rate, does this work for multi-segmented wells too?

It did not, but after the last commit now it works for multi-segmented wells, thanks for pointing this.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1802,6 +1805,9 @@ namespace Opm
                 }
                 perf_press_state[local_perf_index] = perf_press.value();
 
+                // mass rates, for now only water
+                perf_data.wat_mass_rates[local_perf_index] = cq_s[watCompIdx].value() * rhow;
```
bska, line 1809: What happens here if water is not an active phase&ndash;e.g., in an oil/gas run?

bska (commented):
Thanks a lot for adding MSW support, that's greatly appreciated.  I only have one remaining question now.

opm/simulators/wells/MultisegmentWell_impl.hpp
daavid00: Now a check for water has been added, thanks again.

bska (approved):

PR #6182, opened 2025-04-14
adding PLYSHEAR to UnsupportedFlowKeywords

GitPaean:
SHRATE we do support, not sure why we have it there.


GitPaean:
PLYSHEAR caused some testing checking confusion.  Then it was found that we are not supporting PLYSHEAR while we do not see any output regarding it.

GitPaean:
PLYSHLOG we kind of only partially support, we can not do PLYSHLOG for multiple PVT region. Not sure how we should put in the keyword support system.

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
It looks like there can be a RC3, if it is the case, I will suggest to add this to the release.  @multitalentloes

PR #6184, opened 2025-04-16
Fix runs with distributed MSWells with only shut perforations on one process

lisajulia:
Fixes the tests added in https://github.com/OPM/opm-tests/pull/1330.
Without the fix from this PR, the tests fail (see https://github.com/OPM/opm-simulators/pull/6176).

### DIFF
--- a/opm/simulators/wells/MSWellHelpers.cpp
+++ b/opm/simulators/wells/MSWellHelpers.cpp
@@ -119,6 +119,8 @@ mv (const X& x, Y& y) const
     // Then all contributions come from the communication below.
     if (x.size() > 0) {
         B_.mv(x, y);
+    } else {
+        y = 0;
     }
 
     if (this->parallel_well_info_.communication().size() > 1)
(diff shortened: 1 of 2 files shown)


lisajulia:
jenkins build this opm-tests=1330 please

lisajulia:
jenkins build this opm-tests=1330 please

blattms:
@multitalentloes Ideally we would have the first commit in the release. The second one needs changes in opm-tests, which might be problematic.

multitalentloes:
I can cherry-pick that commit for RC3. If I understand correctly here https://github.com/OPM/opm-tests/pull/1330 has already been merged, so adding the second commit as well will probably not cause any problems?

lisajulia:
> I can cherry-pick that commit for RC3. If I understand correctly here [OPM/opm-tests#1330](https://github.com/OPM/opm-tests/pull/1330) has already been merged, so adding the second commit as well will probably not cause any problems?

Thanks!
Yes, https://github.com/OPM/opm-tests/pull/1330 has been merged to master, so adding the second commit should not cause any problems, if the master of opm-tests is taken for the release.

PR #6187, opened 2025-04-22
intersectWithIPR - add extrapolation for last interval whenever it applies

steink:
`intersectWithIPR` was mainly intended to find stable operating points for close to inoperable thp-wells. However, we may also end up here for other reasons, e.g., wells that don't converge because they are forced to operate at too high rate. In this case, if (flo, bhp) ipr-line lies above all points from table  (and hence is _very_ operable) , current code will return an empty intersection.
To prevent this, this PR simpy adds extrapolation based on the two largest flo-values such that an intersection is found also for this case.


steink:

atgeirr:
Spent embarrassingly long time reading and checking the rather simple formulae... Looks good, merging!

steink:
> Spent embarrassingly long time reading and checking the rather simple formulae... 

Same for me, and I wrote it ...
> Looks good, merging!

Thanks :)

PR #6188, opened 2025-04-22
removing the first beginReporStep() for well model

GitPaean:
it will be called in a later stage.

creating the test to evaluate whether it impacts any situations.

### DIFF
--- a/opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp
+++ b/opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp
@@ -349,7 +349,6 @@ public:
             simulator_.setEpisodeIndex(-1);
             simulator_.setEpisodeLength(0.0);
             simulator_.setTimeStepSize(0.0);
-            wellModel_().beginReportStep(timer.currentStepNum());
             simulator_.problem().writeOutput(true);
 
             report_.success.output_write_time += perfTimer.stop();


GitPaean:
jenkins build this please

GitPaean:
To the extent that jekins test can cover, it does not look like it causes any regression.  I am marking the review to be ready for review and potentially to get the PR in.

GitPaean:
For the information, the WellModel::beginReportStep() will be called in a later stage at

https://github.com/OPM/opm-simulators/blob/d58fcf0cbeb05626c2fe407efb4331600ab33f22/opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp#L377

bska (commented):
I **think** this is okay, but do note that the very next statement (`simulator_.problem().writeOutput(true)`) will **typically** end up in `EclWriter<>::writeOutput()` which [accesses](https://github.com/OPM/opm-simulators/blob/1a24fcaa527dcf277edcdaa87d11bb457639ef67/opm/simulators/flow/EclWriter.hpp#L443) the `wellModel()` object without any additional logic.

Before removing this, we either need to verify that it's safe to access the `wellModel()` object before `wellModel_().beginReportStep()`, or we need to add some special case logic to `EclWriter<>::writeOutput()` to handle that.

GitPaean:
jenkins build this please

PR #6191, opened 2025-04-23
Feature/25.04/rc2/ms wells shut perforations

lisajulia:


lisajulia:
This was only for comparison purposes

PR #6192, opened 2025-04-23
AdaptiveTimeStepping: some janitoring

akva2:
Quell some static analyzer issues, add some convenience, fix some headers and rectify some clang-format mess.

### DIFF
--- a/opm/simulators/timestepping/AdaptiveTimeStepping_impl.hpp
+++ b/opm/simulators/timestepping/AdaptiveTimeStepping_impl.hpp
@@ -647,14 +647,13 @@ SimulatorReport
 AdaptiveTimeStepping<TypeTag>::SubStepper<Solver>::
 runStepReservoirCouplingMaster_()
 {
-    bool substep_done = false;
     int iteration = 0;
     const double original_time_step = this->simulator_timer_.currentStepLength();
     double current_time{this->simulator_timer_.simulationTimeElapsed()};
     double step_end_time = current_time + original_time_step;
     auto current_step_length = original_time_step;
     SimulatorReport report;
-    while(!substep_done) {
+    while (true) {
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this please

opm/simulators/timestepping/AdaptiveTimeStepping_impl.hpp
```diff
@@ -720,7 +716,6 @@ runStepReservoirCouplingSlave_()
         report += sub_steps_report;
         current_time += timestep;
         if (final_step) {
-            substep_done = true;
```
bska, line 723: If you're removing this you might as well write the loop as
```C++
while (true) {
    // ...
}
```
since the condition `!substep_done` never changes.  Using `while (true)` also conditions the reader to look for a `break` statement.

bska (commented):
Looks good for the most part, but I suppose we could mark more objects as `const` if we were really tidying this part up.

akva2:
sadly it's more broken than that ;)

that parameter is used by the opm-models simulators and is registered in shared code. the "solution" to this is that we hide the irrelevant options, but with --help-all hidden options show up.

opm/simulators/timestepping/AdaptiveTimeStepping_impl.hpp
akva2: right you are. i only followed the 'unused assignment' warning, but didn't look close enough to realize it is completely unused otherwise. Have remedied, now in a separate commit. Also sprinkled some const.

akva2:
jenkins build this please

jakobtorben:
I see. It sounds like some more changes are needed, also to remove it from opm-models. We can leave it for later.

This parameter is still mentioned in the manual, so it can be confusing.

image

bska (approved):

PR #6194, opened 2025-04-23
parametersystem: add missing throw's

akva2:
These exceptions shall be thrown.


akva2:
jenkins build this please

PR #6196, opened 2025-04-23
Janitoring: quell static analyzer issues

akva2:
Mostly about using standard algorithms

### DIFF
--- a/opm/simulators/flow/FlowGenericVanguard.cpp
+++ b/opm/simulators/flow/FlowGenericVanguard.cpp
@@ -93,6 +93,7 @@
 #include <mpi.h>
 #endif // HAVE_MPI
 
+#include <algorithm>
 #include <filesystem>
 #include <stdexcept>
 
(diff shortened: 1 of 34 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this please

opm/simulators/flow/FlowGenericVanguard.cpp
```diff
@@ -325,11 +326,9 @@ void FlowGenericVanguard::init()
...
     const auto& schedule_wells = schedule().getWellsatEnd();
     parallelWells_.reserve(schedule_wells.size());
-
-    for (const auto& well: schedule_wells)
-    {
-        parallelWells_.emplace_back(well.name(), true);
-    }
+    std::transform(schedule_wells.begin(), schedule_wells.end(),
+                   std::back_inserter(parallelWells_),
+                   [](const auto& well) { return std::make_pair(well.name(), true); });
```
bska, line 331: This replaces in-place construction with construct+move.  Is that really what we want to do?

bska (commented):
I think this looks good for the most part, but I don't quite understand why it's an improvement to switch from in-place construction (`.emplace_back()` &c) to a more elaborate "construct+move" approach.

opm/simulators/flow/FlowGenericVanguard.cpp
atgeirr: Also, the old code is more readable. If static analysis is complaining about this, maybe we could tune the analysis a bit?

akva2: That is in the eye of the beholder. We've had this discussion before..

I was on the fence whether or not to sacrifice in-place, I can remove this one.

bska (approved):

PR #6198, opened 2025-04-24
fixed: gas and water rates swapped

akva2:
Thank you coverity.


GitPaean (approved):
it looks correct. And we seldom put the water at the last.

PR #6208, opened 2025-04-25
Regression tests: more multi registration

akva2:
More use of multi registration macros to reduce amount of duplicated code.


akva2:
jenkins build this please

bska (approved):
Greatly appreciated.  Eliminating duplication is worthwhile here, so I'll merge into master.

PR #6209, opened 2025-04-25
Mark pid+newtonteration as default time step control in help message.

blattms:
This must have been the case for quite some time. It is what the PRT files says when we run without options.


erikhide:
You might consider adding `general3rdorder` to the list. I realised that I forgot to do that earlier.

PR #6215, opened 2025-04-28
Remove mpiutil

akva2:
The mpiutil was added because we did not have access to the standard macros when opm-models was an upstream of opm-simulators. Since this is no longer the case, we can simplify.


akva2:
jenkins build this please

opm/models/utils/simulator.hh
```diff
@@ -145,109 +145,66 @@ public:
...
-                assert(!all_what.empty());
-                throw std::runtime_error(prefix + all_what.front());
-            }
-        };
-
-        try
-        { vanguard_.reset(new Vanguard(*this)); }
-        catch (const std::exception& e) {
-            catchAction(e, verbose_);
+          OPM_BEGIN_PARALLEL_TRY_CATCH();
```
atgeirr, line 149: Indent.

atgeirr (approved):
Only a tiny nitpick, otherwise good to go.

opm/models/utils/simulator.hh
akva2: fixed.

akva2:
jenkins build this please

akva2:
green and approved.

PR #6220, opened 2025-04-29
EnergyModule: some janitoring

akva2:
Some tidy ups


akva2:
jenkins build this please

bska (approved):
Very good and I'll merge into master.  This component isn't completely clean with respect to `Scalar` being something other than `double`, but I think we can address that in follow-up work.

PR #6223, opened 2025-04-29
TransFluxModule: some janitoring

akva2:
Some tidy ups. Also contains a small cleanup in QuantityCallbacks.

### DIFF
--- a/opm/models/common/transfluxmodule.hh
+++ b/opm/models/common/transfluxmodule.hh
@@ -68,11 +68,12 @@ struct TransFluxModule
     using FluxIntensiveQuantities = TransIntensiveQuantities<TypeTag>;
     using FluxExtensiveQuantities = TransExtensiveQuantities<TypeTag>;
     using FluxBaseProblem = TransBaseProblem<TypeTag>;
+
     /*!
      * \brief Register all run-time parameters for the flux module.
      */
     static void registerParameters()
-    { }
+    {}
 };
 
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

opm/models/common/quantitycallbacks.hh
```diff
@@ -53,7 +54,7 @@ class TemperatureCallback
     using ResultRawType = decltype(std::declval<IQFluidState>().temperature(0));
 
 public:
-    using ResultType = typename std::remove_const<typename std::remove_reference<ResultRawType>::type>::type;
+    using ResultType = typename std::remove_const_t<typename std::remove_reference_t<ResultRawType>>;
```
bska, line 57: Part of the benefit of `remove_const_t` and `remove_reference_t` is that they're already defined in terms of `typename`, meaning you can omit the `typename` keyword at the call site.  In other words, you can write this as
```C++
using ResultType = std::remove_const_t<std::remove_reference_t<ResultRawType>>;
```
instead.

bska (commented):
Very good.  Just one slight adjustment and then I think it's ready to go in.

opm/models/common/quantitycallbacks.hh
akva2: indeed. remedied.

akva2:
jenkins build this please

PR #2954, opened 2020-11-30
2nd part distributed stdwells: Write reports only for wells owned.

blattms:
This is done by first examining the local perforations. For this this part is factored out into  a virtual method reportConnections, whose results are later gathered on the process that owns the well.

#2947 is included and needs to be merged first. This one is the next urgent ones to reduce the (already quite high) rebase efforts. There are several people working on the well code.

PR #6231, opened 2025-04-30
Fix output files for distributed wells

lisajulia:
Without the changes in this PR, the newly added test is failing: https://github.com/OPM/opm-simulators/pull/6232

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -38,6 +38,10 @@
 
 #include <opm/simulators/utils/ParallelCommunication.hpp>
 
+#if HAVE_MPI
+#include <opm/simulators/utils/MPISerializer.hpp>
+#endif
+
 #include <opm/grid/common/p2pcommunicator.hh>
 
 #include <algorithm>
@@ -620,6 +624,11 @@ WellState<Scalar>::report(const int* globalCellIdxMap,
             const auto seg_no = ws.segments.segment_number()[seg_ix];
             well.segments[seg_no] = this->reportSegmentResults(well_index, seg_ix, seg_no);
         }
+    #if HAVE_MPI
+        Parallel::MpiSerializer ser(ws.parallel_info.get().communication());
+        ser.broadcast(Parallel::RootRank{0}, well);
+    #endif
+
     }
 
     return res;
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -533,8 +533,7 @@ WellState<Scalar>::report(const int* globalCellIdxMap,
         const auto& wv = ws.surface_rates;
         const auto& wname = this->name(well_index);
 
-        auto dummyWell = data::Well{};
-        auto& well = ws.parallel_info.get().isOwner() ? res[wname] : dummyWell;
+        auto& well = res[wname];
 
         well.bhp = ws.bhp;
         well.thp = ws.thp;
--- a/parallelTests.cmake
+++ b/parallelTests.cmake
@@ -36,6 +36,17 @@ add_test_compare_parallel_simulation(CASENAME spe9_dist_z
                                      REL_TOL ${rel_tol_parallel}
                                      TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8 --enable-drift-compensation=false)
 
+# A test for distributed standard wells with 8 processes. We load distribute only along the z-axis
+add_test_compare_parallel_simulation(CASENAME spe9_dist_z_8
+                                     FILENAME SPE9_CP_SHORT
+                                     POSTFIX 8
+                                     DIR spe9
+                                     SIMULATOR flow_distribute_z
+                                     ABS_TOL ${abs_tol_parallel}
+                                     REL_TOL ${rel_tol_parallel}
+                                     MPI_PROCS 8
+                                     TEST_ARGS --linear-solver-reduction=1e-7 --tolerance-cnv=5e-6 --tolerance-mb=1e-8 --enable-drift-compensation=false)
+
 # A test for distributed multisegment wells. We load distribute only along the z-axis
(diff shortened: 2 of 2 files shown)


lisajulia:
jenkins build this please

opm/simulators/wells/WellState.cpp
```diff
@@ -533,8 +533,7 @@ WellState<Scalar>::report(const int* globalCellIdxMap,
         const auto& wv = ws.surface_rates;
         const auto& wname = this->name(well_index);
 
-        auto dummyWell = data::Well{};
-        auto& well = ws.parallel_info.get().isOwner() ? res[wname] : dummyWell;
+        auto& well = res[wname];
```
bska, line 536: I don't see how this is a fix.  If anything, you've introduced a source of inconsistency.

Before, `WellState<>::report()` would return well data only for the wells which were owned by the local process, now it returns well data for all wells that are intersected on the local process.  That's a potential problem for [`CollectDataOnIORank::collect()`](https://github.com/OPM/opm-simulators/blob/ca92b6906c797278278feab3591b354cb1c1a01b/opm/simulators/flow/EclWriter.hpp#L251) since it now gets multiple sources for the same well.

What problem are you trying to solve?

bska (commented):
I don't understand what this is trying to do.  Please elaborate.

lisajulia:
jenkins build this please

bska:
> Without the changes in this PR, the newly added test is failing: #6232

Okay, but that doesn't really tell me anything.  What, specific, part fails if you don't have this.  There has to be a better way than to just remove a core piece of infrastructure.  Just to be clear, the `dummyWell` construction was introduced in PR #2954 (commit 472623d4a) as part of adding support for distributed wells.  It was an intentional change that has served us well for years.

opm/simulators/wells/WellState.cpp
lisajulia: @bska I have updated the description, also I've added this broadcast call to make sure the well info is consistent across all ranks of the well: https://github.com/OPM/opm-simulators/pull/6231/commits/9377386c6bd0055e7838f883ddde184b8d928aad

bska: > also I've added this broadcast call to make sure the well info is consistent across all ranks of the well: [9377386](https://github.com/OPM/opm-simulators/commit/9377386c6bd0055e7838f883ddde184b8d928aad)

Sure, but as far as I can tell you're addressing some surface level symptoms instead of fixing the real problem.  The `dummyWell` is intentional here.

lisajulia:
> > Without the changes in this PR, the newly added test is failing: #6232
> 
> Okay, but that doesn't really tell me anything. What, specific, part fails if you don't have this? There has to be a better way than to just remove a core piece of infrastructure. Just to be clear, the `dummyWell` construction was introduced in PR #2954 (commit [472623d](https://github.com/OPM/opm-simulators/commit/472623d4a3dd13d0dbe36de56bd39c389a295bd4)) as part of adding support for distributed wells. It was an intentional change that has served us well for years.

Ok, the problem I saw was the following:
When I have a distributed well that is distributed across several ranks (which are not zero, i.e. the "output rank"), then the well data that gets collected via the functions pack and unpack of https://github.com/OPM/opm-simulators/blob/master/opm/simulators/flow/CollectDataOnIORank_impl.hpp#L385 end up being wrong, because pack is called by all processes of the well, but only the owner-rank has the correct well data.
Another approach (that I have not checked yet) would be to extend the PackUnPackWellData such that it knows which process is the owner of the well and only this process calls pack on his well data (which then is hopefully the correct one).

I did not know that this dummy well approach was so critical, sorry.

lisajulia:
> > When I have a distributed well that is distributed across several ranks (but not rank 0, i.e. not on the "output rank"), then the well data that gets collected [...] ends up being wrong,
> 
> Okay, understood. What well data? All of it? Well-level quantities such as BHP/THP and well-level flow rates? Connection level quantities? Segment level quantities?

Lots of quantities actually:
- FGOR
- FGPR
- FOPR
- WBHP:PRODU7
- WGPR:PRODU7
- WGPT:PRODU7
- WOPR:PRODU7
- WOPT:PRODU7

Here is the comparison PDF for the failing test https://github.com/user-attachments/files/18832336/SPE9_CP_SHORT.pdf

lisajulia:
jenkins build this please

bska:
> > Okay, understood. What well data? All of it? Well-level quantities such as BHP/THP and well-level flow rates? Connection level quantities? Segment level quantities?
> 
> Lots of quantities actually:

Right.  That suggests that there's a deeper problem here, and I think I know what it is.  PR #6238 is an alternative fix that makes the handling of shut connections in `wellData()` aware of distributed wells.  Please feel free to give it a spin.

For what it's worth, I was able to reproduce the symptoms shown in your attachment when running `mpirun -np 8`, but not for smaller number of processes.

lisajulia:
> > > Okay, understood. What well data? All of it? Well-level quantities such as BHP/THP and well-level flow rates? Connection level quantities? Segment level quantities?
> > 
> > 
> > Lots of quantities actually:
> 
> Right. That suggests that there's a deeper problem here, and I think I know what it is. PR #6238 is an alternative fix that makes the handling of shut connections in `wellData()` aware of distributed wells. Please feel free to give it a spin.
> 
> For what it's worth, I was able to reproduce the symptoms shown in your attachment when running `mpirun -np 8`, but not for smaller number of processes.

Thanks, I've tested your PR and it solves the problem also :), so I'm happy with either solution. I'd like to keep the test added in 0d9e1c2e0f824cac09dac675821f5870227eefc3 though, to make sure the output will not break in the future.

Concerning the number of processes: When using less than 8 processes, the owner rank of all the production wells is global rank 0, whereas with 8 processes, that is not the case. Then, global rank 0 is treated differently in collectDataOnIORank, since that's also the IORank. That is the reason why this problem does not occur for less than 8 processes.

lisajulia:
Closing this since https://github.com/OPM/opm-simulators/pull/6238 has been merged

PR #6235, opened 2025-05-02
Reduce discontinuities in serial simulation for CpGrid with LGRs

aritorto:
By updating the grid view, cell depth and thickness. 

### DIFF
--- a/opm/models/utils/simulator.hh
+++ b/opm/models/utils/simulator.hh
@@ -129,17 +129,6 @@ public:
             OPM_END_PARALLEL_TRY_CATCH("Allocating the simulation vanguard failed: ", comm);
         }
 
-        // Only relevant for CpGrid
-        if (verbose_) {
-            std::cout << "Adding LGRs, if any\n" << std::flush;
-        }
-
-        {
-            OPM_BEGIN_PARALLEL_TRY_CATCH();
-            vanguard_->addLgrs();
-            OPM_END_PARALLEL_TRY_CATCH("Adding LGRs to the simulation vanguard failed: ", comm);
-        }
-
         if (verbose_) {
             std::cout << "Distributing the vanguard's data\n" << std::flush;
         }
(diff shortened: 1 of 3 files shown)


aritorto:

aritorto:

akva2:
FYI the mpi build will fail, i'm currently working on fully restoring CI.

aritorto:
> FYI the mpi build will fail, i'm currently working on fully restoring CI.

Thanks! 🙌🏾

aritorto:

aritorto:

aritorto:

opm/models/utils/simulator.hh
```diff
@@ -129,25 +129,27 @@ public:
...
         if (verbose_) {
-            std::cout << "Distributing the vanguard's data\n" << std::flush;
+            std::cout << "Adding LGRs, if any, in serial run\n" << std::flush;
         }
 
         {
-            OPM_BEGIN_PARALLEL_TRY_CATCH();
-            vanguard_->loadBalance();
-            OPM_END_PARALLEL_TRY_CATCH("Could not distribute the vanguard data: ", comm);
+            if ( comm.size() == 1 ) { // serial run
```
blattms, line 148: Should we log a warning if not?

blattms (commented):
Apart from the potentially missing warning this looks good.

opm/models/utils/simulator.hh
aritorto: Done! Not here though. I moved this if-serial run into CpGridVanguard::addLgrs(), and warn there if it's a parallel run.

aritorto:

opm/simulators/flow/CpGridVanguard.hpp
```diff
@@ -255,9 +255,16 @@ class CpGridVanguard : public FlowBaseVanguard<TypeTag>
...
+        if (const auto& lgrs = this->eclState().getLgrs(); (lgrs.size() > 0) && (this->grid_->comm().size() == 1)) {
             OpmLog::info("\nAdding LGRs to the grid and updating its leaf grid view");
             this->addLgrsUpdateLeafView(lgrs, lgrs.size(), *this->grid_);
+
+            this->updateGridView_();
+            this->updateCellDepths_();
+            this->updateCellThickness_();
+        }
+        else {
+            Opm::OpmLog::warning("Adding LGRs in parallel run is not supported yet.\n");
```
bska, line 267: This will log a warning about adding LGRs in parallel if `lgrs.size() == 0` and that's probably not what we want.  We probably want something like
```C++
if (const auto& lgrs = ...) {
    // ...
}
else if (this->grid_->comm().size() > 1) {
    // ...
}
```
instead.  On a second note, we're already in the `Opm::` namespace here, so the call to `warning()` here should just be
```C++
OpmLog::warning(...);
```
instead of
```C++
Opm::OpmLog::warning(...);
```

aritorto: Good point, we don't want that warning if there are no LGRs. I'll fix it! Thanks!

aritorto: @bska, thanks again. Your suggestion is incorporated in OPM/opm-simulators#6248

PR #6237, opened 2025-05-02
Reorder Well Data Report Function

bska:
The shut connection handling should be the final stage of this reporting.  While here, also split the tracer rate assingments out to a separate helper function.


bska:
jenkins build this please

akva2 (approved):
Looks good

PR #6238, opened 2025-05-02
Assign Shut Connection Data for Owned Wells Only

bska:
Otherwise, `BlackoilWellModel<>::wellData()` will generate reports for non-owned wells if the wells are distributed.  This, in turn, opens the door to race conditions when collecting contributions on the I/O rank.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModel.hpp
+++ b/opm/simulators/wells/BlackoilWellModel.hpp
@@ -220,6 +220,8 @@ template<class Scalar> class WellContributions;
 
                 this->assignWellTargets(wsrpt);
 
+                this->assignDynamicWellStatus(wsrpt, this->reportStepIndex());
+
                 // Assigning (a subset of the) property values in shut
                 // connections should be the last step of wellData().
                 this->assignShutConnections(wsrpt, this->reportStepIndex());
--- a/opm/simulators/flow/GenericOutputBlackoilModule.cpp
+++ b/opm/simulators/flow/GenericOutputBlackoilModule.cpp
@@ -157,7 +157,7 @@ GenericOutputBlackoilModule(const EclipseState& eclState,
     , flowsC_(schedule, summaryConfig)
     , rftC_(eclState_, schedule_,
             [this](const std::string& wname)
-            { return !isDefunctParallelWell(wname); })
+            { return this->isOwnedByCurrentRank(wname); })
     , rst_conv_(std::move(globalCell), comm)
     , local_data_valid_(false)
 {
--- a/opm/simulators/flow/GenericOutputBlackoilModule.hpp
+++ b/opm/simulators/flow/GenericOutputBlackoilModule.hpp
@@ -330,7 +330,8 @@ protected:
     void setupExtraBlockData(const std::size_t        reportStepNum,
                              std::function<bool(int)> isCartIdxOnThisRank);
 
-    virtual bool isDefunctParallelWell(std::string wname) const = 0;
+    virtual bool isDefunctParallelWell(const std::string& wname) const = 0;
+    virtual bool isOwnedByCurrentRank(const std::string& wname) const = 0;
 
     const EclipseState& eclState_;
     const Schedule& schedule_;
--- a/opm/simulators/flow/OutputBlackoilModule.hpp
+++ b/opm/simulators/flow/OutputBlackoilModule.hpp
@@ -597,7 +597,7 @@ private:
         Problem, std::void_t<decltype(std::declval<Problem>().geoMechModel())>
     > : public std::true_type {};
 
-    bool isDefunctParallelWell(std::string wname) const override
+    bool isDefunctParallelWell(const std::string& wname) const override
     {
         if (simulator_.gridView().comm().size() == 1)
             return false;
@@ -605,6 +605,11 @@ private:
         std::pair<std::string, bool> value {wname, true};
         auto candidate = std::lower_bound(parallelWells.begin(), parallelWells.end(), value);
         return candidate == parallelWells.end() || *candidate != value;
+    }
+
+    bool isOwnedByCurrentRank(const std::string& wname) const override
+    {
+        return this->simulator_.problem().wellModel().isOwner(wname);
     }
 
     void updateFluidInPlace_(const ElementContext& elemCtx, const unsigned dofIdx)
--- a/opm/simulators/flow/OutputCompositionalModule.hpp
+++ b/opm/simulators/flow/OutputCompositionalModule.hpp
@@ -354,7 +354,7 @@ public:
     }
 
 private:
-    bool isDefunctParallelWell(std::string wname) const override
+    bool isDefunctParallelWell(const std::string& wname) const override
     {
         if (simulator_.gridView().comm().size() == 1)
             return false;
@@ -362,6 +362,14 @@ private:
         std::pair<std::string, bool> value {wname, true};
         auto candidate = std::lower_bound(parallelWells.begin(), parallelWells.end(), value);
         return candidate == parallelWells.end() || *candidate != value;
+    }
+
+    bool isOwnedByCurrentRank(const std::string& wname) const override
+    {
+        // Note: This statement is not correct for distributed wells and
+        // will need additional logic once those are supported for
+        // compositional flows.
+        return ! this->isDefunctParallelWell(wname);
     }
 
     void createLocalRegion_(std::vector<int>& region)
--- a/opm/simulators/flow/RFTContainer.cpp
+++ b/opm/simulators/flow/RFTContainer.cpp
@@ -85,19 +85,21 @@ addToWells(data::Wells& wellDatas,
         gatherAndUpdateMap(gasConnectionSaturations_, comm);
     }
 
-    for (const auto& well: schedule_.getWells(reportStepNum)) {
-
-        // don't bother with wells not on this process
-        if (!wellQuery_(well.name())) {
+    for (const auto& wname : this->schedule_.wellNames(reportStepNum)) {
+        // Don't bother with wells not on this process.
+        if (!wellQuery_(wname)) {
             continue;
         }
 
-        //add data infrastructure for shut wells
-        if (!wellDatas.count(well.name())) {
-            auto& wellData = wellDatas[well.name()];
-            wellData.connections.reserve(well.getConnections().size());
-            std::transform(well.getConnections().begin(),
-                           well.getConnections().end(),
+        // Add data infrastructure for shut wells.
+        if (!wellDatas.count(wname)) {
+            const auto& conns = this->schedule_[reportStepNum]
+                .wells(wname).getConnections();
+
+            auto& wellData = wellDatas[wname];
+            wellData.connections.reserve(conns.size());
+
+            std::transform(conns.begin(), conns.end(),
                            std::back_inserter(wellData.connections),
                            [](const auto& connection)
                            {
@@ -107,7 +109,7 @@ addToWells(data::Wells& wellDatas,
                            });
         }
 
-        data::Well& wellData = wellDatas.at(well.name());
+        data::Well& wellData = wellDatas.at(wname);
 
         auto cond_assign = [](double& dest, unsigned idx, const auto& map)
         {
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1025,18 +1025,31 @@ setWsolvent(const Group& group,
     }
 }
 
+template <class Scalar>
+template <typename LoopBody>
+void BlackoilWellModelGeneric<Scalar>::
+loopOwnedWells(LoopBody&& loopBody) const
+{
+    auto wellIndex = 0 * this->wells_ecl_.size();
+
+    for (const auto& pwInfo : this->local_parallel_well_info_) {
+        if (pwInfo.get().isOwner()) {
+            loopBody(wellIndex, this->wells_ecl_[wellIndex]);
+        }
+
+        ++wellIndex;
+    }
+}
+
 template<class Scalar>
 void BlackoilWellModelGeneric<Scalar>::
 assignWellTargets(data::Wells& wsrpt) const
 {
-    auto pwInfo = this->local_parallel_well_info_.begin();
-
-    for (const auto& well : this->wells_ecl_) {
-        if (! pwInfo++->get().isOwner()) {
-            continue;
-        }
-
-        // data::Wells is a std::map<>
+    this->loopOwnedWells([this, &wsrpt]
+                         ([[maybe_unused]] const auto wellIndex,
+                          const Well&                 well)
+    {
+        // data::Wells (i.e., 'wsrpt') is a std::map<>
         auto& limits = wsrpt[well.name()].limits;
 
         if (well.isProducer()) {
@@ -1045,7 +1058,7 @@ assignWellTargets(data::Wells& wsrpt) const
         else {
             this->assignInjectionWellTargets(well, limits);
         }
-    }
+    });
 }
 
 template<class Scalar>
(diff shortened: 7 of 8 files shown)


bska:
I'm creating this PR in draft mode for two reasons

1. It depends on, and contains, the earlier PR #6237.
2. It needs a bit of testing. My local runs suggest that this fixes the underlying problem addressed by #6231, but additional confirmation is needed.

lisajulia:
jenkins build this please

opm/simulators/wells/BlackoilWellModel.hpp
```diff
@@ -198,39 +197,33 @@ template<class Scalar> class WellContributions;
...
+                    // for simplicity, we use region zero (0) here.
+
                     this->assignMassGasRate(wsrpt, FluidSystem::referenceDensity(FluidSystem::gasPhaseIdx, 0));
                 }
+
+                this->assignWellTargets(wsrpt);
+
+                // Assigning (a subset of the) property values in shut
+                // connections should be the last step of wellData().
+                this->assignShutConnections(wsrpt, this->reportStepIndex());
```
blattms, line 225: Unfortunately, there is no documentation for that function. 
The name of the function seems to suggest, that it will assign values to shut connection even for open wells. Is that correct?

blattms (commented):
Thanks for the hint. This explains why we can see the same well in data::Wells  on multiple processes.

It might still not be mergeable as the connections might be local, only and we might miss some on the root process.

blattms:
@akva2 @kjetilly CMake is for broken for CUDA on jenkins...

akva2:
yes which is why i just poked you in https://github.com/OPM/opm-common/pull/4597

bska:
jenkins build this please

opm/simulators/wells/BlackoilWellModel.hpp
bska: > Unfortunately, there is no documentation for that function.

I've added Doxygen-style function descriptions in the latest update.

> The name of the function seems to suggest, that it will assign values to shut connection even for open wells. Is that correct?

That is true, and intentional.  Connections that are [`SHUT`](https://github.com/OPM/opm-common/blob/8d214e843bb9e2a5f52b1a28dfe4238f7baf6d4b/opm/input/eclipse/Schedule/Well/Connection.hpp#L57) in the input never enter into the system of non-linear equations and therefore don't have any result quantities.  Function `assignShutConnections()` generates a few simple result quantities for these connections and ensures that summary vectors like `CTFAC` and `CDFAC` will produce non-zero values even for shut connections.

bska:
I've pushed an update here that makes the solution (slightly) more general and I have also extracted the dynamic status assignment to a separate helper function in anticipation of #6051.  To this end I have created a helper function that loops all wells on the current rank and invokes a call-back function, typically a lambda, for each well that's **owned** by the current rank.

I now consider this to be feature complete, so I'll mark the PR as "ready for review".

bska:
> It might still not be mergeable as the connections might be local, only and we might miss some on the root process.

Note that `assignShutConnections()` loops over [all `WellConnections`](https://github.com/OPM/opm-simulators/blob/146dbdb2e2c027655b3e7d24a3cc1fbfcbb6a23e/opm/simulators/wells/BlackoilWellModelGeneric.cpp#L1159) defined in the `Schedule` object.  Those are the same on all ranks, whether the rank owns the well or not.

bska:
jenkins build this please

lisajulia:
@bska: Thanks a lot for this solution!
I've added a warning if more than one process tries to write the same well to the output data: https://github.com/OPM/opm-common/pull/4602.
Without this PR, this occurs in 16 of our test cases (https://github.com/OPM/opm-common/pull/4602#issuecomment-2854799005), with this PR, still in *1*, in https://ci.opm-project.org/job/opm-common-PR-builder/8369/testReport/junit/(root)/mpi/compareSeparateECLFiles_flow_pyaction_gconsump_insert_kw_4_procs/, I don't know how critical that is...

bska:
> with this PR, still in **1**, in https://ci.opm-project.org/job/opm-common-PR-builder/8369/testReport/junit/(root)/mpi/compareSeparateECLFiles_flow_pyaction_gconsump_insert_kw_4_procs/, I don't know if this is critical...

Neither do I and I don't have a Python-enabled build configuration at the minute.  I can set one up, it's not a big deal, but I notice the following diagnostic output for the contentious well `GI01`:
> ```
> Processing grid
> [...]
> Load balancing distributed the wells as follows:
> [...]
> Well GI01 is inactive, with perforations on ranks: 0 3
> [...]
> ================ Starting main simulation loop ===============
> 
>
> Report step  0/49 at day 0/1461, date = 31-Dec-2017
> Writing well GI01
> Insertion of well GI01 successful.
> Writing well WI01
> Writing well GI01
> [...]
> Error:[...] Insertion of well GI01 unsuccessful, the well already existed in the well map.
> ```
We should at least verify that `GI01` has just a single "owner" in this case and that the the parallel well information does not return `isOwner() == true` on **both** ranks 0 and 3.

lisajulia:
> > with this PR, still in **1**, in https://ci.opm-project.org/job/opm-common-PR-builder/8369/testReport/junit/(root)/mpi/compareSeparateECLFiles_flow_pyaction_gconsump_insert_kw_4_procs/, I don't know if this is critical...
> 
> Neither do I and I don't have a Python-enabled build configuration at the minute. I can set one up, it's not a big deal, but I notice the following diagnostic output for the contentious well `GI01`:
> 
> > ```
> > Processing grid
> > [...]
> > Load balancing distributed the wells as follows:
> > [...]
> > Well GI01 is inactive, with perforations on ranks: 0 3
> > [...]
> > ================ Starting main simulation loop ===============
> > 
> > 
> > Report step  0/49 at day 0/1461, date = 31-Dec-2017
> > Writing well GI01
> > Insertion of well GI01 successful.
> > Writing well WI01
> > Writing well GI01
> > [...]
> > Error:[...] Insertion of well GI01 unsuccessful, the well already existed in the well map.
> > ```
> 
> We should at least verify that `GI01` has just a single "owner" in this case and that the the parallel well information does not return `isOwner() == true` on **both** ranks 0 and 3.

Ok, I'll check that - I'm off for today though, but tomorrow in the morning

bska:
> > We should at least verify that `GI01` has just a single "owner" in this case and that the the parallel well information does not return `isOwner() == true` on **both** ranks 0 and 3.
> 
> I think this directly points to my comment above. If there are shut connections on a process that does not own the well, then your changes will now add that well to `data::Wells` on that rank in addition.

That's not correct.  This PR ensures that we add result contributions to shut connections only for owned wells.  The only case where this fails is if multiple ranks claim to own the well.

> The beauty of operator[] on maps, that just adds a new entry if there is none, yet.

Yes, and we explicitly leverage that feature here.

opm/simulators/wells/BlackoilWellModel.hpp
bska: > I am wondering whether this will really work.

It does.

> There is one process that kind of owns the well for output and passes all the connection information to the output layer. This information is gathered directly after `reportConnections` in [WellState.cpp#L617](https://github.com/OPM/opm-simulators/blob/27f53657ef084cab095dfe0fefaa3cb4d2e76a38/opm/simulators/wells/WellState.cpp#L617) during gatherVectorOnRoot.

Right.

> Any modifications of values on connections that are not on the root process will be lost after that stage.

We don't modify anything as there are no results for shut connections.  We add new `data::Connection` structures for each shut connection.

> Hence I think this code will only add shut connections that the root process knows about to the output.

The root process, and all other processes sharing a well for that matter, know about all connections, shut or otherwise, since we explicitly loop [all connections](https://github.com/OPM/opm-simulators/blob/22f89067f49fc6f0d33b2a688e33ffc6a27bb322/opm/simulators/wells/BlackoilWellModelGeneric.cpp#L1159) of the well, whether on the current rank or not.  We do not, however, create a new `data::Connection` object if the connection is not SHUT, since the assumption is that those results have already been provided by `WellState::report()`.

lisajulia:
> > > with this PR, still in **1**, in https://ci.opm-project.org/job/opm-common-PR-builder/8369/testReport/junit/(root)/mpi/compareSeparateECLFiles_flow_pyaction_gconsump_insert_kw_4_procs/, I don't know if this is critical...
> > 
> > 
> > Neither do I and I don't have a Python-enabled build configuration at the minute. I can set one up, it's not a big deal, but I notice the following diagnostic output for the contentious well `GI01`:
> > > ```
> > > Processing grid
> > > [...]
> > > Load balancing distributed the wells as follows:
> > > [...]
> > > Well GI01 is inactive, with perforations on ranks: 0 3
> > > [...]
> > > ================ Starting main simulation loop ===============
> > > 
> > > 
> > > Report step  0/49 at day 0/1461, date = 31-Dec-2017
> > > Writing well GI01
> > > Insertion of well GI01 successful.
> > > Writing well WI01
> > > Writing well GI01
> > > [...]
> > > Error:[...] Insertion of well GI01 unsuccessful, the well already existed in the well map.
> > > ```
> > 
> > 
> > We should at least verify that `GI01` has just a single "owner" in this case and that the the parallel well information does not return `isOwner() == true` on **both** ranks 0 and 3.
> 
> Ok, I'll check that - I'm off for today though, but tomorrow in the morning

I have checked this, my findings:

-  The issue has nothing to do with PYACTION, it also occurs for: `mpirun -np 4 ./bin/flow ../../opm-tests/actionx/ACTIONX_GCONSUMP.DATA ` (hence you can also reproduce this with an installation that does not have python enabled).
- I believe it also has nothing to do with the ACTIONX, that is tested in this file, the warning occurs right at the beginning of the simulation, before any ACTIONX was triggered:
  ```
  ================ Starting main simulation loop ===============
  
  
  Report step  0/49 at day 0/1461, date = 31-Dec-2017
  
  Warning: Received output data for well GI01 several times - this might be incorrect!
  ```
- The well G01 is inactive, could that be the cause?
  ```
  Load balancing distributed the wells as follows:
  well name            ranks with perforated cells
---------------------------------------------------
            OP01       1
            OP02       1
            OP03       1

Well GI01 is inactive, with perforations on ranks: 0 3 
Well WI01 is inactive, with perforations on ranks: 2 

  ```

bska:
> > > We should at least verify that `GI01` has just a single "owner" in this case and that the the parallel well information does not return `isOwner() == true` on **both** ranks 0 and 3.
> > 
> > Ok, I'll check that - I'm off for today though, but tomorrow in the morning
> 
> I have checked this, my findings:
> 
> * The issue has nothing to do with PYACTION, it also occurs for: `mpirun -np 4 ./bin/flow ../../opm-tests/actionx/ACTIONX_GCONSUMP.DATA ` (hence you can also reproduce this with an installation that does not have python enabled).

Thanks a lot for the report.  It took a minute to find the underlying cause, but in the end it turned out to be another instance of the original problem, this time in the [RFT data collection](https://github.com/OPM/opm-simulators/blob/1648de24f8fbc9eeb0f6a8345535fe662149f8c5/opm/simulators/flow/EclWriter.hpp#L463-L466) in `EclWriter<>::writeOutput()`.  The latest update here fixes that problem too.

> * I believe it also has nothing to do with the ACTIONX, that is tested in this file, the warning occurs right at the beginning of the simulation, before any ACTIONX was triggered:
>   ```
>   ================ Starting main simulation loop ===============
>   
>   
>   Report step  0/49 at day 0/1461, date = 31-Dec-2017
>   
>   Warning: Received output data for well GI01 several times - this might be incorrect!
>   ```

You're right, this is unrelated to `ACTIONX`.

> * I've verified: GI01 has just a single owner: isOwner() == true only on rank 0, not on rank 3

Very good.

> * Still, both ranks 0 and 3 try to write data for GI01.

Yes, you're right.

> * As you and Markus already pointed out, the well GI01 is inactive, could that be the cause, i.e., this: [Assign Shut Connection Data for Owned Wells Only #6238 (comment)](https://github.com/OPM/opm-simulators/pull/6238#issuecomment-2857321026)?

The well being inactive was partly related, because inactive wells are distributed by default (#5609, #5686).  To be fair though, that was really only the direct trigger which exposed the underlying problem.

bska:
jenkins build this please

bska:
jenkins build this please

blattms:
@bska Unrelated:

While looking at all of this, I noticed that we treat some shut wells differently for reporting at various places:

- In assignShutConnections we do not report any connections if the well es marked as SHUT 
- In contrast to this some shut wells seem to be special in `WellState::report`, namely those that got shut due to convergence issues. There is an
```
const auto& ws = this->well(well_index);
        if ((ws.status == Well::Status::SHUT) && !wasDynamicallyClosed(well_index))
        {
            continue;
        }
```
to skip only a subset of the shut wells. This means that connections are reported for some shut wells.

Is this how it should be or should we report shut connections also for wells that got shut dynamically?

bska:
> * In assignShutConnections we do not report any connections if the well is marked as SHUT

If the well is marked as `SHUT` then `assignShutConnections()` will generate `data::Connection` objects for all connections.  The [`wellIsOpen`](https://github.com/OPM/opm-simulators/blob/cb3d29ef799c91ee26cfb141eb68f282fa7778a8/opm/simulators/wells/BlackoilWellModelGeneric.cpp#L1109) flag will be false whence `skip()` will always return false and we [add new `data::Connection` objects](https://github.com/OPM/opm-simulators/blob/cb3d29ef799c91ee26cfb141eb68f282fa7778a8/opm/simulators/wells/BlackoilWellModelGeneric.cpp#L1128) for each element in `Well::getConnections()`.

> * In contrast to this some shut wells seem to be special in `WellState::report`, namely those that got shut due to convergence issues.

That is indeed a special case.  It is supposed to generate the dynamic well and connection results for the last time step that a well was operable.

lisajulia:
> > > > We should at least verify that `GI01` has just a single "owner" in this case and that the the parallel well information does not return `isOwner() == true` on **both** ranks 0 and 3.
> > > 
> > > 
> > > Ok, I'll check that - I'm off for today though, but tomorrow in the morning
> > 
> > 
> > I have checked this, my findings:
> > 
> > * The issue has nothing to do with PYACTION, it also occurs for: `mpirun -np 4 ./bin/flow ../../opm-tests/actionx/ACTIONX_GCONSUMP.DATA ` (hence you can also reproduce this with an installation that does not have python enabled).
> 
> Thanks a lot for the report. It took a minute to find the underlying cause, but in the end it turned out to be another instance of the original problem, this time in the [RFT data collection](https://github.com/OPM/opm-simulators/blob/1648de24f8fbc9eeb0f6a8345535fe662149f8c5/opm/simulators/flow/EclWriter.hpp#L463-L466) in `EclWriter<>::writeOutput()`. The latest update here fixes that problem too.
> 
> > * I believe it also has nothing to do with the ACTIONX, that is tested in this file, the warning occurs right at the beginning of the simulation, before any ACTIONX was triggered:
> >   ```
> >   ================ Starting main simulation loop ===============
> >   
> >   
> >   Report step  0/49 at day 0/1461, date = 31-Dec-2017
> >   
> >   Warning: Received output data for well GI01 several times - this might be incorrect!
> >   ```
> 
> You're right, this is unrelated to `ACTIONX`.
> 
> > * I've verified: GI01 has just a single owner: isOwner() == true only on rank 0, not on rank 3
> 
> Very good.
> 
> > * Still, both ranks 0 and 3 try to write data for GI01.
> 
> Yes, you're right.
> 
> > * As you and Markus already pointed out, the well GI01 is inactive, could that be the cause, i.e., this: [Assign Shut Connection Data for Owned Wells Only #6238 (comment)](https://github.com/OPM/opm-simulators/pull/6238#issuecomment-2857321026)?
> 
> The well being inactive was partly related, because inactive wells are distributed by default (#5609, #5686). To be fair though, that was really only the direct trigger which exposed the underlying problem.

Ok perfect, thanks a lot! Then I think this can be merged now, right? :smiley:

bska:
> > The well being inactive was partly related, because inactive wells are distributed by default (#5609, #5686). To be fair though, that was really only the direct trigger which exposed the underlying problem.
> 
> Ok perfect, thanks a lot! Then this can be merged now, right? 😃

Hopefully, yes.  I'll run a final build check.

bska:
jenkins build this please

lisajulia:
One more check to make sure:

lisajulia:

bska:
> One more check to make sure:

Cool&ndash;thanks.  Once complete, assuming the check is successful, we "only" need maintainer approval before merging this.

bska:

atgeirr:
@lisajulia I interpret your comments above as approving of the PR, but you have not approved it using the review mechanism, so not 100% sure.

Please merge if that is what you meant!

bska:

lisajulia:
Merging now after @atgeirr's approval.

PR #6241, opened 2025-05-05
Instatiate preconditioners directly with GPU matrices

kjetilly:
This pull request makes the GPU preconditioners hold the GPU matrix as a reference (instead of owning it as a value). The motivation for this change is three-fold:

1) Reduce the memory usage on the GPU when more than just the preconditioner is using the matrix on the GPU, for instance when the full linear solver loop runs on the GPU, which is the case now with `gpubicgstab`.
2) Reduce host-to-device copy overhead.
3) Reduce the responsibility of the GPU Preconditioners. 

### Introducing the `PreconditionerCPUMatrixToGPUMatrix` adapter
We still need something to copy the matrix from the CPU to the GPU in the setting where we assemble the matrix on the CPU (which is always at the time of writing). This is done through the added class `Opm::gpuistl::PreconditionerCPUMatrixToGPUMatrix` which responsibility is owning the GPU matrix and copying the CPU matrix to the GPU matrix on calls to `update`.  

The GPU preconditioners have been augmented in such a way that they now all store a reference to their GPU matrix. 

### Dependency on CPU matrix
In the case of `GPUDILU` and `OPMGPUILU0`, we still need the matrix on the CPU for the construction. This necessitates the following, temporary fixes (temporary until we figure out if and how we should create it from a GPU matrix): 

1)  `GPUDILU` and `OPMGPUILU0` both take the CPU **and** GPU matrix as a constructor parameter. Note that it is only holding a reference to the GPU matrix after its creation. To keep the compilation simple, we keep the CPUMatrix template argument for these preconditioners for now.

2) Since we need the CPU matrix to create the aforementioned GPU conditioners, we create the CPU matrix on the fly for the preconditioner factory creating GPU preconditioners.

### Smaller changes

To accommodate said changes, we have made some smaller adjustments:

1) `GpuSparseMatrix` is now copyable
2)  `GpuSparseMatrix` is can now copy nonzero values from `GpuSparseMatrix`
3) `GpuVector` can now be constructor from a DUNE BlockVector
4) `GpuVector` can now transfer values from `GpuVector`
5) Some const-correctness changes in the kernels being called from the GPU preconditioners now that the GPU matrix is a const reference.

### Unit test for correctness

To the test the implementation, a unit test comparing the wrapped and non-wrapped preconditioners for the GPU has been added. This is in addition to the already existing unit tests for the GPU preconditioners testing them against the CPU variants.

## To manual
Reorganisational changes to the code structure of gpu-ISTL to accommodate less memory consumption and fewer host-to-device copies in the future.

### DIFF
--- a/opm/simulators/linalg/StandardPreconditioners_gpu_serial.hpp
+++ b/opm/simulators/linalg/StandardPreconditioners_gpu_serial.hpp
@@ -116,14 +116,14 @@ private:
         Dune::BCRSMatrix<BlockType> matrix(N, N, numberOfNonZeroes, Dune::BCRSMatrix<BlockType>::row_wise);
         for (auto row = matrix.createbegin(); row != matrix.createend(); ++row) {
 
-            for (std::size_t j = rowIndices[row.index()]; j != rowIndices[row.index() + 1]; ++j) {
+            for (auto j = rowIndices[row.index()]; j != rowIndices[row.index() + 1]; ++j) {
                 const auto columnIndex = columnIndices[j];
                 row.insert(columnIndex);
             }
         }
 
         for (std::size_t i = 0; i < N; ++i) {
-            for (std::size_t j = rowIndices[i]; j != rowIndices[i + 1]; ++j) {
+            for (auto j = rowIndices[i]; j != rowIndices[i + 1]; ++j) {
                 const auto columnIndex = columnIndices[j];
                 // Now it gets a bit tricky, first we need to fetch the block matrix
                 BlockType blockMatrix;
--- a/opm/simulators/linalg/StandardPreconditioners_gpu_serial.hpp
+++ b/opm/simulators/linalg/StandardPreconditioners_gpu_serial.hpp
@@ -18,19 +18,155 @@
 #define OPM_STANDARDPRECONDITIONERS_GPU_SERIAL_HEADER
 
 
-namespace Opm {
+#include <dune/istl/bcrsmatrix.hh>
+#include <type_traits>
+
+
+namespace Opm
+{
 
 
 template <class Operator>
-struct StandardPreconditioners<Operator, Dune::Amg::SequentialInformation, typename std::enable_if_t<Opm::is_gpu_operator_v<Operator>>> 
+struct StandardPreconditioners<Operator,
+                               Dune::Amg::SequentialInformation,
+                               typename std::enable_if_t<Opm::is_gpu_operator_v<Operator>>> 
 {
+
+    using O = Operator;
+    using C = Dune::Amg::SequentialInformation;
+    using F = PreconditionerFactory<O, C>;
+    using M = typename F::Matrix;
+    using V = typename F::Vector;
+    using P = PropertyTree;
+    using PrecPtr = typename F::PrecPtr;
+
+    using field_type = typename V::field_type;
+
+    static constexpr int maxblocksize = 6;
     static void add()
     {
-        // No standard preconditioners for this type of operator.
+        F::addCreator("ILU0", [](const O& op, const P& prm, const std::function<V()>&, std::size_t) {
+            const double w = prm.get<double>("relaxation", 1.0);
+            using GpuILU0 =
+                typename gpuistl::GpuSeqILU0<M, gpuistl::GpuVector<field_type>, gpuistl::GpuVector<field_type>>;
+            return std::make_shared<GpuILU0>(op.getmat(), w);
+        });
+
+        F::addCreator("JAC", [](const O& op, const P& prm, const std::function<V()>&, std::size_t) {
+            const double w = prm.get<double>("relaxation", 1.0);
+            using GPUJac = typename gpuistl::GpuJac<M, gpuistl::GpuVector<field_type>, gpuistl::GpuVector<field_type>>;
+            return std::make_shared<GPUJac>(op.getmat(), w);
+        });
+
+        // The next two (OPMILU0 and DILU) are the GPU preconditioners that need CPU matrices.
+        // Since we are not storing the block size compile time for the GPU matrices 
+        // **and** we need the CPU matrix for the creation, we need to
+        // dispatch the creation of the preconditioner based on the block size. 
+        //
+        // Note that this dispatching is not needed in the future, since we will have a constructor taking GPU matrices directly.
+        F::addCreator("OPMILU0", [](const O& op, [[maybe_unused]] const P& prm, const std::function<V()>&, std::size_t) {
+            return dispatchOnBlockSize<maxblocksize>(op, [&](const auto& cpuMatrix) {
+                const bool split_matrix = prm.get<bool>("split_matrix", true);
+                const bool tune_gpu_kernels = prm.get<bool>("tune_gpu_kernels", true);
+                const int mixed_precision_scheme = prm.get<int>("mixed_precision_scheme", 0);
+                using CPUMatrixType = std::remove_const_t<std::remove_reference_t<decltype(cpuMatrix)>>;
+                using GPUILU0 =
+                    typename gpuistl::OpmGpuILU0<CPUMatrixType, gpuistl::GpuVector<field_type>, gpuistl::GpuVector<field_type>>;
+
+                return std::make_shared<GPUILU0>(op.getmat(), cpuMatrix, split_matrix, tune_gpu_kernels, mixed_precision_scheme);
+            });
+        });
+
+        F::addCreator("DILU", [](const O& op, [[maybe_unused]] const P& prm, const std::function<V()>&, std::size_t) {
+            return dispatchOnBlockSize<maxblocksize>(op, [&](const auto& cpuMatrix) {
+                const bool split_matrix = prm.get<bool>("split_matrix", true);
+                const bool tune_gpu_kernels = prm.get<bool>("tune_gpu_kernels", true);
+                const int mixed_precision_scheme = prm.get<int>("mixed_precision_scheme", 0);
+                using CPUMatrixType = std::remove_const_t<std::remove_reference_t<decltype(cpuMatrix)>>;
+                using GPUDILU =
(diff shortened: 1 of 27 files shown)


kjetilly:
Jenkins build this please

kjetilly:
Jenkins build this please

kjetilly:
Jenkins build this please

kjetilly:

kjetilly:

opm/simulators/linalg/StandardPreconditioners_mpi.hpp
```diff
@@ -21,6 +21,7 @@
 #ifndef OPM_STANDARDPRECONDITIONERS_MPI_HEADER
 #define OPM_STANDARDPRECONDITIONERS_MPI_HEADER
 
+#include <opm/simulators/linalg/gpuistl/PreconditionerCPUMatrixToGPUMatrix.hpp>
```
atgeirr, line 24: Should be inside HAVE_CUDA

opm/simulators/linalg/StandardPreconditioners_mpi.hpp
```diff
@@ -333,9 +338,14 @@ struct StandardPreconditioners
             const int mixed_precision_scheme = prm.get<int>("mixed_precision_scheme", 0);
             using field_type = typename V::field_type;
             using GpuDILU = typename gpuistl::GpuDILU<M, gpuistl::GpuVector<field_type>, gpuistl::GpuVector<field_type>>;
-            auto gpuDILU = std::make_shared<GpuDILU>(op.getmat(), split_matrix, tune_gpu_kernels, mixed_precision_scheme);
-
-            auto adapted = std::make_shared<gpuistl::PreconditionerAdapter<V, V, GpuDILU>>(gpuDILU);
+            using MatrixWrapper = Opm::gpuistl::PreconditionerCPUMatrixToGPUMatrix<gpuistl::GpuVector<field_type>, 
+                gpuistl::GpuVector<field_type>, GpuDILU, M>;
+        
+            // Note: op.getmat() is passed twice, because the Jacobi needs both the CPU and GPU matrix.
```
atgeirr, line 344: Jacobi -> GpuDILU

opm/simulators/linalg/StandardPreconditioners_gpu_serial.hpp
```diff
@@ -18,19 +18,157 @@
...
+        // **and** we need the CPU matrix for the creation, we need to
+        // dispatch the creation of the preconditioner based on the block size. 
+        //
+        // Note that this dispatching is not needed in the future, since we will have a constructor taking GPU matrices directly.
+        F::addCreator("OPMILU0", [](const O& op, [[maybe_unused]] const P& prm, const std::function<V()>&, std::size_t) {
+            return dispatchOnBlockSize<maxblocksize>(op, [&](const auto& cpuMatrix) {
+                const bool split_matrix = prm.get<bool>("split_matrix", true);
+                const bool tune_gpu_kernels = prm.get<bool>("tune_gpu_kernels", true);
+                const int mixed_precision_scheme = prm.get<int>("mixed_precision_scheme", 0);
+
```
atgeirr, line 72: Delete extra newlines.

atgeirr (commented):
Reviewed the parts of this NOT in gpuistl, and it looks good, with only minor quibbles.

kjetilly:

opm/simulators/linalg/gpuistl/PreconditionerCPUMatrixToGPUMatrix.hpp
```diff
@@ -0,0 +1,127 @@
...
+        return detail::shouldCallPreconditionerPre<CudaPreconditionerType>();
+    }
+
+    virtual bool hasPerfectUpdate() const override
+    {
+        return m_underlyingPreconditioner.hasPerfectUpdate();
+    }
+
+private:
+    CPUMatrixType m_cpuMatrix;
```
multitalentloes, line 119: Is it not sufficient for this to hold a reference to avoid allocating another CPU matrix?

tests/gpuistl/test_GpuDILU.cpp
```diff
@@ -235,7 +238,7 @@ BOOST_AUTO_TEST_CASE(TestDiluApplyBlocked)
 
     // init matrix with 2x2 blocks
     Sp2x2BlockMatrix matA = get2x2BlockTestMatrix();
-    auto gpudilu = GpuDilu2x2(matA, true, true, false);
+    auto gpudilu = Wrapped2x2(matA, matA, true, true, false);
```
multitalentloes, line 241: initially thought it looked strange to send in the cpumatrix twice to the `PreconditionerCPUMatrixToGPUMAtrix`, but then realized this is because the second `matA` is actually the first argument to the DILU which needs a CPU matrix for coloring, this is the correct understanding?

jakobtorben:
Tried testing this locally with the following command:

```sh
opm-simulators/bin/flow /opm-data/norne/NORNE_ATW2013.DATA --linear-solver=linear_solver_gpubicgstab_GPUDILU_config.json --matrix-add-well-contributions=1
```

where `linear_solver_gpubicgstab_GPUDILU_config.json` has the following contents:

```json
{
    "tol": "0.01",
    "maxiter": "200",
    "verbosity": "0",
    "solver": "gpubicgstab",
    "preconditioner": {
        "type": "GPUDILU",
        "split_matrix": "true",
        "tune_gpu_kernels": "true"
    }
}
```
This fails almost immediately on Norne (but not on master):

```sh
Starting time step 8, stepsize 3.75975e-06 days, at day 2.00513/8, date = 08-Nov-1997

Simulation aborted: Solver failed to converge after cutting timestep 10 times.
```

Are you seeing the same behaviour on system?

opm/simulators/linalg/gpuistl/GpuVector.cpp
```diff
@@ -300,6 +300,19 @@ GpuVector<T>::copyToHost(std::vector<T>& data) const
     copyToHost(data.data(), data.size());
 }
 
+template <class T>
+void
+GpuVector<T>::copyFromDevice(const GpuVector<T>& data) const
```
multitalentloes, line 305: maybe we should consider another taxonomy like having `copyDtoD` and similar so origin and destination is obvious for the person reading the code, in this case `copyFromDevice` is not obvious that will be to the device as well. Not a hard requirement for this PR, but I realized this now and wanted to have it written

kjetilly:
Jenkins build this please

kjetilly:
Jenkins build this please

jakobtorben:
> Tried testing this locally with the following command:
> 
> ```shell
> opm-simulators/bin/flow /opm-data/norne/NORNE_ATW2013.DATA --linear-solver=linear_solver_gpubicgstab_GPUDILU_config.json --matrix-add-well-contributions=1
> ```
> 
> where `linear_solver_gpubicgstab_GPUDILU_config.json` has the following contents:
> 
> ```json
> {
>     "tol": "0.01",
>     "maxiter": "200",
>     "verbosity": "0",
>     "solver": "gpubicgstab",
>     "preconditioner": {
>         "type": "GPUDILU",
>         "split_matrix": "true",
>         "tune_gpu_kernels": "true"
>     }
> }
> ```
> 
> This fails almost immediately on Norne (but not on master):
> 
> ```shell
> Starting time step 8, stepsize 3.75975e-06 days, at day 2.00513/8, date = 08-Nov-1997
> 
> Error: Solver failed to converge after cutting timestep 10 times.
> Simulation aborted: Solver failed to converge after cutting timestep 10 times.
> ```
> 
> Are you seeing the same behaviour on system?

After your recent changes this is now working again and matching behaviour on master.

opm/simulators/linalg/StandardPreconditioners_gpu_serial.hpp
```diff
@@ -18,19 +18,155 @@
...
+                const bool split_matrix = prm.get<bool>("split_matrix", true);
+                const bool tune_gpu_kernels = prm.get<bool>("tune_gpu_kernels", true);
+                const int mixed_precision_scheme = prm.get<int>("mixed_precision_scheme", 0);
+                using CPUMatrixType = std::remove_const_t<std::remove_reference_t<decltype(cpuMatrix)>>;
+                using GPUDILU =
+                    typename gpuistl::GpuDILU<CPUMatrixType, gpuistl::GpuVector<field_type>, gpuistl::GpuVector<field_type>>;
+                return std::make_shared<GPUDILU>(op.getmat(), cpuMatrix, split_matrix, tune_gpu_kernels, mixed_precision_scheme);
+            });
+        });
+    }
```
jakobtorben, line 91: Just so I am sure that I understand correctly here, these GPU preconditioners creators are called "JAC", "OPMILU0" and "DILU" (without a GPU prefix), since they are only added when the whole linear solver is on the GPU, so it is not necessary to specify GPU?

multitalentloes:
I have looked over the code and made my comments now, I have also run norne locally as a sanity check, and everything looked normal. I think this is a good development of the codebase

tests/gpuistl/test_GpuDILU.cpp
kjetilly: This is the correct undertanding, yes.

opm/simulators/linalg/StandardPreconditioners_gpu_serial.hpp
kjetilly: that is correct.

kjetilly:
Jenkins build this please

kjetilly:
> Before we merge this, would you mind having a look at the list of [new warnings](https://ci.opm-project.org/job/opm-simulators-PR-builder/8084/gcc/folder.-979637784/) introduced here and see if there's a way to trim that list, please?

I think the latest commit should remove the new warnings, let's see if Jenkins agrees.

kjetilly:
The new warnings are gone at least.

kjetilly:
Jenkins build this please

jakobtorben (approved):

PR #6242, opened 2025-05-05
fix namespace in linearsolverreport.hh

multitalentloes:
we are already in the `Opm::linear` namespace.


multitalentloes:
Jenkins build this please

bska (approved):
Yep, good catch&ndash;I'll merge into master.  If at any point in the future we need to disambiguate the `Timer` name, we can use `::Opm::Timer` (or similar) to do it.

PR #6244, opened 2025-05-06
Move throws inside the loop over all wells to after the end of the loop

lisajulia:
Instead of throwing inside the loop over all wells, collect the findings and throw at the end. Throwing inside of the loop over all wells leads to a deadlock in some cases:

Without these changes, the following command locks with the current master:
`mpirun -np 4 ./bin/flow_distribute_z ../../opm-tests/msw/MSW-SIMPLE-5-SHUT-PERFORATIONS.DATA`
This PR also solves the deadlock that happens when adding more MS Wells to the above case, like:
```
Load balancing distributed the wells as follows:
  well name            ranks with perforated cells
---------------------------------------------------
          PROD1       0       1
          PROD2       0       1       2
          PROD3       0       1       2

```
as explained in https://github.com/OPM/opm-simulators/pull/6244#discussion_r2079120699.

Not relevant for the reference manual.

### DIFF
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -671,6 +671,7 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
     const int np = pu.num_phases;
 
     std::optional<std::string> distributedMSWellLocalErrorMessage;
+    std::optional<std::string> connectionWithoutAssociatedSegmentsLocalErrorMessage;
 
     // in the init function, the well rates and perforation rates have been initialized or copied from prevState
     // what we do here, is to set the segment rates and perforation rates
@@ -701,11 +702,16 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
                 if (connection.state() == Connection::State::OPEN) {
                     const int segment_index = segment_set.segmentNumberToIndex(connection.segment());
                     if (segment_index == -1) {
-                        OPM_THROW(std::logic_error,
-                                  fmt::format("COMPSEGS: Well {} has connection in cell {}, {}, {} "
-                                              "without associated segment.", well_ecl.name(),
-                                              connection.getI() + 1 , connection.getJ() + 1,
-                                              connection.getK() + 1 ));
+                        if (!connectionWithoutAssociatedSegmentsLocalErrorMessage) {
+                            connectionWithoutAssociatedSegmentsLocalErrorMessage = ""; // First well with error: initialize the string
+                        }
+
+                        // Append to the existing error message for errors in further wells
+                        *connectionWithoutAssociatedSegmentsLocalErrorMessage +=
+                            fmt::format("COMPSEGS: Well {} has connection in cell {}, {}, {} "
+                            "without associated segment.\n", well_ecl.name(),
+                            connection.getI() + 1 , connection.getJ() + 1,
+                            connection.getK() + 1 );
                     }
 
                     segment_perforations[segment_index].push_back(n_activeperf);
@@ -830,11 +836,24 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
         }
     }
 
-    // Instead of throwing inside the loop over all wells, collect the findings and throw at the end.
-    // Throwing inside of the loop over all wells leads to a deadlock in some cases.
+    {
+        // Instead of throwing inside the loop over all wells, collect the findings and throw at the end.
+        // Throwing inside of the loop over all wells leads to a deadlock in some cases.
 
-    if (!this->enableDistributedWells_ && distributedMSWellLocalErrorMessage) {
-        throw std::logic_error(*distributedMSWellLocalErrorMessage);
+        std::string fullErrorMessage = {};
+
+        if (!this->enableDistributedWells_ && distributedMSWellLocalErrorMessage) {
+            *distributedMSWellLocalErrorMessage += "Either adjust the partitioning or set the flag --allow-distributed-wells to true.\n";
+            fullErrorMessage += *distributedMSWellLocalErrorMessage;
+        }
+
+        if (connectionWithoutAssociatedSegmentsLocalErrorMessage) {
+            fullErrorMessage += *connectionWithoutAssociatedSegmentsLocalErrorMessage;
+        }
+
+        if (!fullErrorMessage.empty()) {
+            throw std::logic_error(fullErrorMessage);
+        }
     }
 
     if (prev_well_state) {
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -44,6 +44,7 @@
 #include <cassert>
 #include <initializer_list>
 #include <numeric>
+#include <optional>
 #include <set>
 #include <stdexcept>
 #include <type_traits>
@@ -669,6 +670,8 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
     const auto& pu = this->phaseUsage();
     const int np = pu.num_phases;
 
+    std::optional<std::string> distributedMSWellLocalErrorMessage;
+
     // in the init function, the well rates and perforation rates have been initialized or copied from prevState
     // what we do here, is to set the segment rates and perforation rates
     for (int w = 0; w < nw; ++w) {
@@ -716,8 +719,26 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
             }
             ws.parallel_info.get().setActiveToLocalMap(active_to_local);
 
-            if (!this->enableDistributedWells_ && static_cast<int>(ws.perf_data.size()) != n_activeperf)
-                throw std::logic_error("Distributed multi-segment wells cannot be initialized properly yet.");
+            // Check if the multi-segment well is distributed across several processes by comparing the local number
+            // of active perforations (ws.perf_data.size()) with the total number of active perforations (n_activeperf).
+            // If they differ on any rank, the well is considered distributed.
+
+            if (static_cast<int>(ws.perf_data.size()) != n_activeperf) {
+                if (!distributedMSWellLocalErrorMessage) {
+                    distributedMSWellLocalErrorMessage = ""; // First well with error: initialize the string
+                }
+
+                // Append to the existing error message for errors in further wells
+                *distributedMSWellLocalErrorMessage += fmt::format(
+                    "Distributed multi-segment well {} detected on rank {}.\n"
+                    "The number of active perforations on rank {} is {}, but the total number of active perforations is {}.\n\n",
+                    well_ecl.name(),
+                    ws.parallel_info.get().communication().rank(),
+                    ws.parallel_info.get().communication().rank(),
+                    ws.perf_data.size(),
+                    n_activeperf
+                );
+            }
 
 
             std::vector<std::vector<int>> segment_inlets(well_nseg);
@@ -809,6 +830,12 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
         }
     }
 
+    // Instead of throwing inside the loop over all wells, collect the findings and throw at the end.
+    // Throwing inside of the loop over all wells leads to a deadlock in some cases.
+
+    if (!this->enableDistributedWells_ && distributedMSWellLocalErrorMessage) {
+        throw std::logic_error(*distributedMSWellLocalErrorMessage);
+    }
 
     if (prev_well_state) {
         for (int w = 0; w < nw; ++w) {
--- a/opm/simulators/wells/WellState.cpp
+++ b/opm/simulators/wells/WellState.cpp
@@ -672,6 +672,7 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
 
     std::optional<std::string> distributedMSWellLocalErrorMessage;
     std::optional<std::string> connectionWithoutAssociatedSegmentsLocalErrorMessage;
+    std::optional<std::string> initializePerforationMismatchLocalErrorMessage;
 
     // in the init function, the well rates and perforation rates have been initialized or copied from prevState
     // what we do here, is to set the segment rates and perforation rates
@@ -799,7 +800,12 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
                         perforation_rates[global_active_perf_index * np + i] = perf_rates[perf * np + i];
                     }
                 } else {
-                    OPM_THROW(std::logic_error,fmt::format("Error when initializing MS Well state, there is no active perforation index for the local index {}", perf));
+                    if (!initializePerforationMismatchLocalErrorMessage) {
+                        initializePerforationMismatchLocalErrorMessage = ""; // First well with error: initialize the string
+                    }
+
+                    // Append to the existing error message for errors in further wells
+                    *initializePerforationMismatchLocalErrorMessage += fmt::format("Error when initializing MS Well state of well {}, there is no active perforation index for the local index {}.\n", well_ecl.name(), perf);
                 }
             }
             if (ws.parallel_info.get().communication().size() > 1) {
@@ -849,6 +855,10 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
 
         if (connectionWithoutAssociatedSegmentsLocalErrorMessage) {
             fullErrorMessage += *connectionWithoutAssociatedSegmentsLocalErrorMessage;
+        }
+
+        if (initializePerforationMismatchLocalErrorMessage) {
+            fullErrorMessage += *initializePerforationMismatchLocalErrorMessage;
         }
 
         if (!fullErrorMessage.empty()) {


lisajulia:
jenkins build this please

opm/simulators/wells/WellState.cpp
```diff
@@ -716,8 +716,14 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
             }
             ws.parallel_info.get().setActiveToLocalMap(active_to_local);
 
-            if (!this->enableDistributedWells_ && static_cast<int>(ws.perf_data.size()) != n_activeperf)
-                throw std::logic_error("Distributed multi-segment wells cannot be initialized properly yet.");
+            // Check if the multi-segment well is distributed across by comparing the local number of active perforations
```
blattms, line 719: I think that there are words missing after across

blattms (changes requested):
Only one nitpick about the comment.

One question: Can this happen in a default flow run, e.g. when using PYACTION or something else?

lisajulia:
> Only one nitpick about the comment.
> 
> One question: Can this happen in a default flow run, e.g. when using PYACTION or something else?

This is independent of PYACTION; that totally depends on if there are distributed MSWells in the code.

lisajulia:
> Do distributed MS wells "only" distribute the connections, or could the segment objects conceivably be distributed too? I ask because this PR exclusively checks whether the connections are distributed.

To my knowledge, if the connections are distributed, then the segments are distributed as well, so this check is valid here.
@vkip: Can you possibly answer the question from @bska? I saw that this check was introduced by you in https://github.com/OPM/opm-simulators/commit/c42f02e80262167b1f335747bf6feb49215b0b38. Thanks!

bska:
> > Do distributed MS wells "only" distribute the connections, or could the segment objects conceivably be distributed too? I ask because this PR exclusively checks whether the connections are distributed.
> 
> To my knowledge, if the connections are distributed, then the segments are distributed as well, so this check is valid here.

I more concerned about the (very) special case where all connections happen to go on one rank, but the segment objects happen to be distributed among multiple ranks.

lisajulia:
jenkins build this please

vkip:
> > Do distributed MS wells "only" distribute the connections, or could the segment objects conceivably be distributed too? I ask because this PR exclusively checks whether the connections are distributed.
> 
> To my knowledge, if the connections are distributed, then the segments are distributed as well, so this check is valid here. @vkip: Can you possibly answer the question from @bska? I saw that this check was introduced by you in [c42f02e](https://github.com/OPM/opm-simulators/commit/c42f02e80262167b1f335747bf6feb49215b0b38). Thanks!

I don't know - in this commit I was only concerned about inactive wells, actually distributed wells was left for someone else to handle (as indicated by the comment).

lisajulia:
> > > Do distributed MS wells "only" distribute the connections, or could the segment objects conceivably be distributed too? I ask because this PR exclusively checks whether the connections are distributed.
> > 
> > 
> > To my knowledge, if the connections are distributed, then the segments are distributed as well, so this check is valid here. @vkip: Can you possibly answer the question from @bska? I saw that this check was introduced by you in [c42f02e](https://github.com/OPM/opm-simulators/commit/c42f02e80262167b1f335747bf6feb49215b0b38). Thanks!
> 
> I don't know - in this commit I was only concerned about inactive wells, actually distributed wells was left for someone else to handle (as indicated by the comment).

Thanks - then, as far as I understand now, the question is rather: Should the check "Do we have distributed MSWells and how should we make the user aware of that?" possibly be done earlier / somewhere else in the code or differently?
I'd suggest opening an issue for this question and merging this PR now, because it is an improvement of the status quo.

GitPaean:
> To my knowledge, if the connections are distributed, then the segments are distributed as well, 

Not directly related to the PR, and sorry I did not follow the parallelization of the multisegment wells very well, but it is good that someone can answer this question.  connections are physically attached to the grid, while segments are kind of abstract concepts not attached to the grid. Some segments not necessarily has connections associated with them.

lisajulia:
> > To my knowledge, if the connections are distributed, then the segments are distributed as well,
> 
> Not directly related to the PR, and sorry I did not follow the parallelization of the multisegment wells very well, but it is good that someone can answer this question. connections are physically attached to the grid, while segments are kind of abstract concepts not attached to the grid. Some segments not necessarily has connections associated with them.

Yes, you're right, I meant segments that have connections associated with them.

In my interpretation segments without associated connections cannot be distributed in a "grid sense", since they are not connected to the grid; also, in the matrices of the Schur complement and the well residual, these segments appear on all ranks the well perforates (like in https://github.com/OPM/opm-tests/blob/master/msw/MSW-SIMPLE.DATA, which has 25 segments, only one branch, segments 0-12 are not connected to the grid, segments 13-24 are connected to the grid).

lisajulia:
@blattms: Can you have another look here?
@bska and @GitPaean: I'd suggest opening an issue for the question "Should the check "Do we have distributed MSWells and how should we make the user aware of that?" possibly be done earlier / somewhere else in the code or differently?" and merging this PR now, because it is an improvement of the status quo - what do you say?
Thanks!

opm/simulators/wells/WellState.cpp
```diff
@@ -716,8 +716,14 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
             }
             ws.parallel_info.get().setActiveToLocalMap(active_to_local);
 
-            if (!this->enableDistributedWells_ && static_cast<int>(ws.perf_data.size()) != n_activeperf)
-                throw std::logic_error("Distributed multi-segment wells cannot be initialized properly yet.");
+            // Check if the multi-segment well is distributed across several processes by comparing the local number
+            // of active perforations (ws.perf_data.size()) with the total number of active perforations (n_activeperf).
+            // If they differ on any rank, the well is considered distributed.
+            int haveDistributedMSWellLocalIndicator = (static_cast<int>(ws.perf_data.size()) != n_activeperf) ? 1 : 0;
+            bool haveDistributedMSWellGlobalFlag = ws.parallel_info.get().communication().max(haveDistributedMSWellLocalIndicator);
```
GitPaean, line 723: Do you consider it is too heavy to call the global communication within a for loop?

opm/simulators/wells/WellState.cpp
```diff
@@ -716,8 +716,14 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
...
+
+            if (!this->enableDistributedWells_ && haveDistributedMSWellGlobalFlag)
+                throw std::logic_error("Distributed multi-segment well detected (" + well_ecl.name() + ") but the flag --allow-distributed-wells is set to false.");
```
GitPaean, line 726: any comments regarding `throw` in a parallel running, I guess all the processes will throw in the current code.

opm/simulators/wells/WellState.cpp
atgeirr: I think this might lock if the ranks have the wells in different order, so maybe a bit fragile? Although that may not be an actual problem (if all have in same order).

opm/simulators/wells/WellState.cpp
GitPaean, line 726: The well name `well_ecl.name()` might not be the `well` that causes the condition triggered in a parallel running or I am mistaken here?  Since the flag  `            bool haveDistributedMSWellGlobalFlag = ws.parallel_info.get().communication().max(haveDistributedMSWellLocalIndicator); ` is a globally reduced value?

opm/simulators/wells/WellState.cpp
lisajulia: No, I think this will not lock since the communicator I'm using is the communicator of the well, not the global one.

lisajulia: I've acutally introduced this change, because the following locks with the current master:
mpirun -np 4 ./bin/flow_distribute_z ../../opm-tests/msw/MSW-SIMPLE-5-SHUT-PERFORATIONS.DATA

opm/simulators/wells/WellState.cpp
lisajulia: No, the following locks with the current master:
`mpirun -np 4 ./bin/flow_distribute_z ../../opm-tests/msw/MSW-SIMPLE-5-SHUT-PERFORATIONS.DATA`

opm/simulators/wells/WellState.cpp
atgeirr: It requires that all ranks that share the communicator are at this spot in the code at the same time to be successful, which requires that they are all in the same order: if ranks 0 and 1 have wells (A, B) and (B, A) then rank 0 will be waiting to communicate about A while rank 1 is waiting on well B. So I think it could lock, but will be ok (maybe with some waiting though) as long as we have same order everywhere, i.e. (A, B) for both rank 0 and 1 in the example.

opm/simulators/wells/WellState.cpp
lisajulia: I'll check that

opm/simulators/wells/WellState.cpp
lisajulia: Ah yes, you're right, ok - yet do we know that wells are in the same order? Now as you brought this up here, I believe all the rest of the parallel well code assumes that :thinking:.

atgeirr: If so, maybe that shows that they are in the same order... Otherwise it would lock all the time!

GitPaean:
Sorry, it is not an issue I followed very closely, so my comments will be shallow. 

In the original checking introduced by https://github.com/OPM/opm-simulators/commit/c42f02e80262167b1f335747bf6feb49215b0b38  . 

In the original comments, 
```c++
            // If perf_data does not contain all active perfs, the well is distributed or inactive
            // For inactive wells there is no need for calculating an initial solution
            // \todo{ Update the procedure below to work for actually distributed wells. }
            if (static_cast<int>(ws.perf_data.size()) != n_activeperf)
```

It mentioned the two possible causes that cause the checking to fail, one is distributed and the other one is `inactive wells`.  For the `inactive wells` scenario, not sure it is totally accurate, maybe it means some `connections` are not active? @vkip . 

Anyway, how this PR handles the second scenario of `inactive wells`, which does not involve parallel running. 


And furthermore, I would like the error message to be more specific to add `by detecting the connection/perforation number differ with the number it should be`, something like that, to ease the above concern regarding whether the segments are distributed.  Then it becomes a sanity checking of connection numbers of the running.

opm/simulators/wells/WellState.cpp
GitPaean: If the `ws.parallel_info` is a smaller communicator, will some processes throw causes other processes hang? Sorry if I am mistaken without checking the implementation details.

opm/simulators/wells/WellState.cpp
lisajulia: Yes, I think so!
This here is just one spot (there are many more):
https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/BlackoilWellModel_impl.hpp#L1521
There is communication in the function `assembleWellEqWithoutIteration` for Standard wells (https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/StandardWell_impl.hpp#L446) and also for Multisegment wells (https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/MultisegmentWell_impl.hpp#L1834).

lisajulia:
> > > Do distributed MS wells "only" distribute the connections, or could the segment objects conceivably be distributed too? I ask because this PR exclusively checks whether the connections are distributed.
> > 
> > 
> > To my knowledge, if the connections are distributed, then the segments are distributed as well, so this check is valid here.
> 
> I more concerned about the (very) special case where all connections happen to go on one rank, but the segment objects happen to be distributed among multiple ranks.

I've seen a deadlock with master in a similar case:
`mpirun -np 4 ./bin/flow_distribute_z ../../opm-tests/msw/MSW-SIMPLE-5-SHUT-PERFORATIONS.DATA`,
There the well is distributed across ranks 0 and 1 but rank 0 only contains shut perforations
Yet with the communication added in this PR, the error gets thrown properly and the simulation does not lock. I believe for your case it could be the same, no?

lisajulia:
> > It mentioned the two possible causes that cause the checking to fail, one is distributed and the other one is `inactive wells`. For the `inactive wells` scenario, not sure it is totally accurate, maybe it means some `connections` are not active? @vkip .
> 
> `inactive` wells here are wells that are not open (and do not need to be solved) for the entirety of the simulation, e.g., wells used only to collect RFT data. By default these wells can be split across processes even for the case '--allow-distributed-wells=false'.

I think these inactive wells handled by this
https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/WellState.cpp#L676, no?

opm/simulators/wells/WellState.cpp
lisajulia: I've adjusted the code now, can you have another look?

lisajulia:
jenkins build this please

lisajulia:
jenkins build this please

blattms (approved):

opm/simulators/wells/WellState.cpp
```diff
@@ -716,8 +718,17 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
...
             ws.parallel_info.get().setActiveToLocalMap(active_to_local);
 
-            if (!this->enableDistributedWells_ && static_cast<int>(ws.perf_data.size()) != n_activeperf)
-                throw std::logic_error("Distributed multi-segment wells cannot be initialized properly yet.");
+            // Check if the multi-segment well is distributed across several processes by comparing the local number
+            // of active perforations (ws.perf_data.size()) with the total number of active perforations (n_activeperf).
+            // If they differ on any rank, the well is considered distributed.
+
+            if (static_cast<int>(ws.perf_data.size()) != n_activeperf) {
+                haveDistributedMSWellLocalIndicator = haveDistributedMSWellLocalIndicator || true;
```
blattms, line 726: Why is the right hand side so complicated? It is always true, isn't it?

lisajulia: Yes :sweat_smile:

opm/simulators/wells/WellState.cpp
```diff
@@ -716,8 +718,17 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
...
+            // Check if the multi-segment well is distributed across several processes by comparing the local number
+            // of active perforations (ws.perf_data.size()) with the total number of active perforations (n_activeperf).
+            // If they differ on any rank, the well is considered distributed.
+
+            if (static_cast<int>(ws.perf_data.size()) != n_activeperf) {
+                haveDistributedMSWellLocalIndicator = true;
+                distributedMSWellLocalErrorMessage += "Distributed multi-segment well " + well_ecl.name() + " detected on rank " +
+                    std::to_string(ws.parallel_info.get().communication().rank()) + ".\nThe number of active perforations on rank " +
+                    std::to_string(ws.parallel_info.get().communication().rank()) + " is " + std::to_string(ws.perf_data.size()) +
+                    ", but the total number of active perforations is " + std::to_string(n_activeperf) + ".\n\n";
```
bska, line 730: We have [libfmt](https://github.com/OPM/opm-simulators/blob/6eaaa62ef3b46d62c1e1d4f0209b8851b26b17e9/opm/simulators/wells/WellState.cpp#L53) here, so there's no need to string together multiple `to_string()` calls in this case.

lisajulia: Thanks, yes!

lisajulia:
jenkins build this please

opm/simulators/wells/WellState.cpp
```diff
@@ -809,6 +825,11 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
         }
     }
 
+    if (!this->enableDistributedWells_ && haveDistributedMSWellLocalIndicator) {
```
GitPaean, line 828: I am not working with parallelization much, excuse me if I read things wrong. In the new version, `haveDistributedMSWellLocalIndicator` is not communicated through the processes (looks like), will it be the same across processes and all the processes will throw at the same time if needed?

GitPaean: I am not aware of the usage `OPM_PARALLEL_TRY_CATCH` here.  It sounds good and thanks for the explanation.

lisajulia:
jenkins build this please

opm/simulators/wells/WellState.cpp
lisajulia: Ok, also FYI: Now I'm looking for other spots where there is a throw statement inside such a loop.

atgeirr: That sounds like a very good idea! I think we did a thorough job back in 2019 or so(?) when we added the DeferredLogger and gatherDeferredLogger() to deal with things like this, but it is likely that we have not managed to entirely avoid such mistakes in the time since (although there have been plenty of fixes as well).

lisajulia:
jenkins build this please

opm/simulators/wells/WellState.cpp
```diff
@@ -669,6 +669,13 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
     const auto& pu = this->phaseUsage();
     const int np = pu.num_phases;
 
+    bool distributedMSWellLocalIndicator = false;
+    std::string distributedMSWellLocalErrorMessage = {};
+    bool connectionWithoutAssociatedSegmentsLocalIndicator = false;
+    std::string connectionWithoutAssociatedSegmentsLocalErrorMessage = {};
+    bool initializePerforationMismatchLocalIndicator = false;
+    std::string initializePerforationMismatchLocalErrorMessage = {};
```
bska, line 677: Could these pairs maybe be collected into `optional<string>` objects instead?  That way `nullopt` could signify what's now `Indicator=false`?  Either that, or some dedicated `struct`?  At least something that's less error prone than updating two objects consistently?

lisajulia:
jenkins build this please

opm/simulators/wells/WellState.cpp
```diff
@@ -809,6 +832,22 @@ void WellState<Scalar>::initWellStateMSWell(const std::vector<Well>& wells_ecl,
...
+        throw std::logic_error(distributedMSWellLocalErrorMessage);
+    }
+
+    if (connectionWithoutAssociatedSegmentsLocalIndicator) {
+        throw std::logic_error(connectionWithoutAssociatedSegmentsLocalErrorMessage);
+    }
+
+    if (initializePerforationMismatchLocalIndicator) {
+        throw std::logic_error(initializePerforationMismatchLocalErrorMessage);
+    }
```
bska, line 849: This structure means that if there's more than one error condition, then we'll throw an exception for the first of them (in code order).  Maybe there's a way to collect all conditions and just throw one exception for the full collection?

opm/simulators/wells/WellState.cpp
lisajulia: Yes, thanks!

opm/simulators/wells/WellState.cpp
lisajulia: Yes, also changed that now.

bska (approved):

opm/simulators/wells/WellState.cpp
vkip: Maybe I misunderstand something, but isn't the well/group handling littered with loops over wells that potentially contain well solves? Mostly in https://github.com/OPM/opm-simulators/blob/master/opm/simulators/wells/BlackoilWellModel_impl.hpp, I think. In particular, note that  ```well->updateWellStateWithTarget()``` potentially contains well solves in case of producers which can potentially be THP controlled. My guess is that at least these need to be handled:

* https://github.com/OPM/opm-simulators/blob/c57e93bf2d4ee8a75cdda550ccc68750c4186fa0/opm/simulators/wells/BlackoilWellModel_impl.hpp#L1104
* https://github.com/OPM/opm-simulators/blob/c57e93bf2d4ee8a75cdda550ccc68750c4186fa0/opm/simulators/wells/BlackoilWellModel_impl.hpp#L1156
* https://github.com/OPM/opm-simulators/blob/c57e93bf2d4ee8a75cdda550ccc68750c4186fa0/opm/simulators/wells/BlackoilWellModel_impl.hpp#L1256
* https://github.com/OPM/opm-simulators/blob/c57e93bf2d4ee8a75cdda550ccc68750c4186fa0/opm/simulators/wells/BlackoilWellModel_impl.hpp#L1519
* https://github.com/OPM/opm-simulators/blob/c57e93bf2d4ee8a75cdda550ccc68750c4186fa0/opm/simulators/wells/BlackoilWellModel_impl.hpp#L1781
* https://github.com/OPM/opm-simulators/blob/c57e93bf2d4ee8a75cdda550ccc68750c4186fa0/opm/simulators/wells/BlackoilWellModel_impl.hpp#L1804
* https://github.com/OPM/opm-simulators/blob/c57e93bf2d4ee8a75cdda550ccc68750c4186fa0/opm/simulators/wells/BlackoilWellModel_impl.hpp#L1908

lisajulia: Hey Vegard,
thank, yes you're absolutely right! I've added this list to our internal task for this!

atgeirr: I think that the point from Lisa from above: "Never throw inside a loop that loops over the wells and contains communication" stands. It is fine to throw in a loop over wells, as long as there is no other communication there.

The problem here (if I understand it correctly) is that with distributed wells, we do communication during the well solves themselves, thus breaking Lisa's rule. I think the suggestion given, "always finish the loop and throw afterwards" will require each of these loops to do try-catch inside the loop body, setting some error variable and continuing to treat every well. This may be cumbersome and error-prone, so we should make some support functionality for this, perhaps even with new macros similar to the ones we already have.

PR #6246, opened 2025-05-07
Throw if bhp is NaN in findTHP

totto82:
The canary died.

```
//Canary in a coal mine: shouldn't really be required
            assert(found == true);
            static_cast<void>(found); //Silence compiler warning
```
If bhp_arg is NaN we trigger this assert.

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -262,9 +262,20 @@ namespace Opm
             return;
         }
 
-        BVectorWell xw(1);
-        this->linSys_.recoverSolutionWell(x, xw);
-        updateWellState(simulator, xw, well_state, deferred_logger);
+        try {
+            BVectorWell xw(1);
+            this->linSys_.recoverSolutionWell(x, xw);
+
+            updateWellState(simulator, xw, well_state, deferred_logger);
+        }
+        catch (const NumericalProblem& exp) {
+            // Add information about the well and log to deferred logger
+            // (Logging done inside of recoverSolutionWell() (i.e. by UMFpack) will only be seen if
+            // this is the process with rank zero)
+            deferred_logger.problem("In MultisegmentWell::recoverWellSolutionAndUpdateWellState for well "
+                                    + this->name() +": "+exp.what());
+            throw;
+        }
     }
 
 
--- a/opm/simulators/wells/VFPHelpers.cpp
+++ b/opm/simulators/wells/VFPHelpers.cpp
@@ -21,6 +21,7 @@
 #include <opm/simulators/wells/VFPHelpers.hpp>
 
 #include <opm/common/ErrorMacros.hpp>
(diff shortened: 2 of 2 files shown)


totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -262,9 +262,21 @@ namespace Opm
...
         }
 
-        BVectorWell xw(1);
-        this->linSys_.recoverSolutionWell(x, xw);
-        updateWellState(simulator, xw, well_state, deferred_logger);
+
+        try{
+            BVectorWell xw(1);
+            this->linSys_.recoverSolutionWell(x, xw);
+
```
atgeirr, line 269: Nitpick: Extra blank. (Also space before brace at `try` and parenthesis at `catch`.)

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -262,9 +262,21 @@ namespace Opm
...
+            updateWellState(simulator, xw, well_state, deferred_logger);
+        }
+        catch(const NumericalProblem& exp) {
+            // Add information about the well and log to deferred logger
+            // (Logging done inside of solve() method will only be seen if
```
atgeirr, line 274: This confuses me for two reasons:
 - We are not calling solve() here?
 - If we have something in the deferred logger here, it will be seen for any rank, there is a OPM_END_PARALLEL_TRY_CATCH_LOG in BlackoilWellModel<>::recoverWellSolutionAndUpdateWellState().

Please explain, or change comment if wrong.

atgeirr (commented):
Should be merged, but have a question and some nitpicks.

opm/simulators/wells/MultisegmentWell_impl.hpp
totto82: Sorry. The comment is wrong, it was copied from elsewhere without thinking. I will update. Thanks for noticing.

totto82: Is the comment okay now? As I understand it, we need to catch and rethrow here to handle the throw inside the UMFpack.

totto82:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
atgeirr: That is better. Maybe not 100% precise still, as we do not get any logging from UMFPack afaik, just dumped output to standard output? (So it goes to the terminal but not the log files.) But your point is that without the catch and throw here, we might miss an exception thrown due to UMFPack failure?

totto82: I think so. It is @blattms that added this some time ago for similar calls, so maybe he would like to comment.

atgeirr: We can clarify this later, I'll merge as is.

PR #6251, opened 2025-05-08
dont solve if converged (MSW)

totto82:
In some places in the code, we solve before we check for convergence. 

The algorithm is
while(not converged)
assemble
check for convergence
solve and update if not converged

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -599,7 +599,6 @@ namespace Opm
         // which is why we do not put the assembleWellEq here.
         try{
             const BVectorWell dx_well = this->linSys_.solve();
-
             updateWellState(simulator, dx_well, well_state, deferred_logger);
         }
         catch(const NumericalProblem& exp) {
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

totto82:

totto82:
Test failures are expected as this PR change the order of when regularization is used (before or after assembly) and for network it adds one extra solve in the pre_step

totto82:

totto82:

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1712,6 +1711,7 @@ namespace Opm
                     break;
                 }
             }
+            const BVectorWell dx_well = this->linSys_.solve();
```
atgeirr: This could fail, so I think it would be appropriate to add the try-catch here as well, as in iterateWellEqWithSwitching() lines 1550-1561.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1506,26 +1506,13 @@ namespace Opm
...
-                                        + this->name() +": "+exp.what());
-                throw;
-            }
-
             if (it > this->param_.strict_inner_iter_wells_) {
                 relax_convergence = true;
                 this->regularize_ = true;
             }
 
+            assembleWellEqWithoutIteration(simulator, dt, inj_controls, prod_controls, well_state, group_state, deferred_logger);
```
atgeirr: Nitpick: too long line.

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1506,26 +1506,13 @@ namespace Opm
...
+
```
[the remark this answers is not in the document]
atgeirr: I agree that this is fine.

It did make me think though: it is not obvious that we should stop regularizing at the same iteration we relax convergence, why is it so? Also having regularize_ as a member is not very nice, better to be a function argument for assembleWellEqWithoutIteration() and a local variable here, i think.

totto82:

totto82:

totto82:
jenkins build this opm-tests=1345 please

PR #6254, opened 2025-05-09
fixed: pass by pointer instead of reference

akva2:
while harmless, as the result is unused, we currently deref a null ptr in serial builds. this causes issues for builds enabling cxxlib assertions


akva2:

PR #6258, opened 2025-05-09
fixing compilation with older gcc compiler (GCC 9.X)

GitPaean:
changing  friend class AdaptiveTimeStepping<TypeTag>::template SubStepIteration<Solver>;
to be     friend class SubStepIteration<Solver>;

It was reported by @ElyesAhmed and solution proposed by @bska .

### DIFF
--- a/opm/simulators/timestepping/AdaptiveTimeStepping.hpp
+++ b/opm/simulators/timestepping/AdaptiveTimeStepping.hpp
@@ -89,8 +89,6 @@ private:
     };
 
     // Forward declaration of SubStepIteration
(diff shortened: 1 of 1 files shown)


GitPaean:
jenkins build this please

akva2 (approved):
Fix is fine as such but for the record I'd rather see friend usage go altogether. That is a bigger refactor.

GitPaean:
> Fix is fine as such but for the record I'd rather see friend usage go altogether. That is a bigger refactor.

Thanks for the approval and jenkins is green. I am getting it in now.

PR #6259, opened 2025-05-09
Create Pull Request template

atgeirr:
Intended to remind developers to tag the PRs and give good descriptions.

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,0 +1,6 @@
+Remember to tag the Pull Request with one of the manual: labels, this is required to merge a PR.
+
+Delete this template text and make a description of the PR that explains its user relevance,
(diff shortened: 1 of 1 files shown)


multitalentloes:
It may be a bit excessive, but we can also add checklists and similar as a part of the template.

- [X] Example
- [ ] of
- [X] checklist

A few standard checklists might give reviewers a very quick overview over what kind of pr it is, for instance what other OPM repositories do they depend on, or create a list of which other PRs it depends on etc. Checklists can also nudge people to have better programming habits if they for instance have a checklist of whether they have remembered to apply the clang formatter, is the git history rebased into something reasonable, if new functionality is added, have you also added tests etc... Not sure if most contributers and reviewers consider this useful or just bloat.

PR #6260, opened 2025-05-09
FvBaseAdLocalLinearizer: some janitoring

akva2:
Some janitoring in this class. Includes, avoid raw pointers, cosmetics, ...


bska (approved):

PR #6265, opened 2025-05-12
Add Regression Tests for UDA Compatibility

bska:
Following PR OPM/opm-common#4439, we can activate most of the aspects of the model in PR OPM/opm-tests#1299.

### DIFF
--- a/regressionTests.cmake
+++ b/regressionTests.cmake
@@ -1072,6 +1072,16 @@ add_test_compareECLFiles(CASENAME 01_vappars
                          REL_TOL ${rel_tol}
                          DIR vappars)
 
+add_test_compareECLFiles(CASENAME 6_uda_model5_stdw
+  FILENAME 6_UDA_MODEL5_STDW
+  SIMULATOR flow
+  ABS_TOL ${abs_tol}
+  REL_TOL ${rel_tol}
+  DIR model5
+  RESTART_STEP 1
+  TEST_ARGS --enable-tuning=true
+)
+
 set(_mult_cases
   MULTFLT-01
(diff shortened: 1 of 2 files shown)


bska:
I'm creating this PR in draft mode because I want to run a few checks on the CI system before enabling the regression test.

bska:
jenkins build this please

bska:
I think I'm ready to introduce this regression test now so I'm marking the PR as "ready for review".  We will need to calculate reference solutions before merging the test.

regressionTests.cmake
```diff
@@ -1072,6 +1072,16 @@ add_test_compareECLFiles(CASENAME 01_vappars
...
 
+add_test_compareECLFiles(CASENAME 6_uda_model5_stdw
+  FILENAME 6_UDA_MODEL5_STDW
+  SIMULATOR flow
+  ABS_TOL ${abs_tol}
+  REL_TOL ${rel_tol}
+  DIR model5
+  RESTART_STEP 1
+  TEST_ARGS --enable-tuning=true
+  )
```
akva2, line 1083: <-

akva2 (approved):
One cosmetic

regressionTests.cmake
bska: > :arrow_left: 

Good catch.  I've pushed an update to address that.

bska:

bska:
jenkins build this opm-tests=1341 please

bska:
PR approved, build check is green, and the new reference solutions have been installed on the CI system.  I'll merge into master.

PR #6197, opened 2025-04-23
Fixing index single comp well state and many others

GitPaean:
The PR begins with fixing the  out-of-bounds writing to surface_phase_rates in the well state, thanks for @akva2 's reporting.

Then it removes the usage of the PhaseUsage in the compositional code as a trial for doing the same in the main flow code. 

At the end, it rename the CompositionalWellModel to be CompWellModel, so it is the same with other files and classes.

PR #5835, opened 2025-01-03
changed: use standard header guard structure for _impl.hpp clangd helpers

akva2:
@bska here you go

PR #6266, opened 2025-05-13
CompWellInterface/SingleCompWellState: some janitoring

akva2:
I happened to open these in my IDE and it was not happy. Remedy this.


akva2:
jenkins build this please

flowexperimental/comp/wells/CompWellInterface.hpp
```diff
@@ -98,12 +108,12 @@ class CompWellInterface
     std::vector<int> saturation_table_number_;
 
     // std::string name_;
-
 };
 
-
 } // end of namespace Opm
 
+#ifndef OPM_COMP_WELLINTERFACE_IMPL_HPP
```
GitPaean, line 115: This introduced extra lines, do we need it since the `#ifndef OPM_COMP_WELLINTERFACE_IMPL_HPP` defined in `CompWellInterface_impl.hpp`

akva2: yes. this is for IDEs only. it needs to include the CompWellInterface.hpp to have the symbols defined. in regular compilation the _impl.hpp file is included from the .hpp file. For IDEs it's the opposite, the _impl.hpp includes the .hpp file. These lines are needed to avoid circular includes in that case.

flowexperimental/comp/wells/SingleCompWellState_impl.hpp
```diff
@@ -17,6 +17,24 @@
...
+// Improve IDE experience
+#ifndef OPM_SINGLE_COMP_WELL_STATE_HPP
+#include <config.h>
+#include <flowexperimental/comp/wells/SingleCompWellState.hpp>
+#endif
+
+#include <opm/input/eclipse/EclipseState/Compositional/CompositionalConfig.hpp>
+#include <opm/input/eclipse/Schedule/Well/Well.hpp>
+
+#include <opm/simulators/utils/BlackoilPhases.hpp>
```
GitPaean, line 32: I thought part of the purpose of https://github.com/OPM/opm-simulators/pull/6197 was to remove the usage of  `BlackoilPhases.hpp`. Do we still need it?

akva2: uhm, probably not, i did the changes before I rebased the PR. i'll remedy.

flowexperimental/comp/wells/CompWellInterface.hpp
GitPaean: What IDE are you using and do you need to set some specifications to see this type of warning?

akva2: I'm using qtcreator, but anything using clangd or intellisense or whatever will show the issue. It creates the AST of the _impl.hpp file to analyze, and when it does, it has no definition of the class available.

akva2:
jenkins build this please

flowexperimental/comp/wells/CompWellInterface.hpp
bska: > This introduced extra lines, do we need it since the `#ifndef OPM_COMP_WELLINTERFACE_IMPL_HPP` defined in `CompWellInterface_impl.hpp`

For what it's worth, this is the structure we (ultimately) ended up using following the discussion in [#5834](https://github.com/OPM/opm-simulators/pull/5834#discussion_r1900899194) and also implemented in #5835.

flowexperimental/comp/wells/SingleCompWellState.hpp
```diff
@@ -99,6 +102,8 @@ class SingleCompWellState
 
 } // namespace Opm
 
+#ifndef OPM_SINGLE_COMP_WELL_STATE_IMPL_HPP
 #include "SingleCompWellState_impl.hpp"
+#endif
```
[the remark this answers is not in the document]
akva2: It's to avoid circular includes when processed in the IDE. It works without but you get a warning. I can remove it of course, but I don't really see the harm.

impl includes .hpp, and without the if .hpp includes the impl..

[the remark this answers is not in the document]
bska: > I can remove it of course, but I don't really see the harm.

The harm, in my opinion, is that it makes the "main" (i.e., public) header aware of a symbol that's strictly speaking needed only in the implementation.  That introduces unnecessary coupling between the two since the _impl header already needs to be aware of the preprocessing symbols from the main header in addition to its own include guards.

> It's to avoid circular includes when processed in the IDE. It works without but you get a warning.

I'm tempted to call that a deficiency in the IDE's parser since there is no circular inclusion here.  I know that the _impl header includes the main header in this particular use case and that the main header without these guards goes on to include the _impl header.  This clearly looks circular on the face of it, but it's also why we have separate include guards in the _impl header.

PR #6269, opened 2025-05-13
Add communication for calculating the max_pressure in maxPerfPress 

lisajulia:
... such that the maximal pressure over **all** perforations is taken, across all processes

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -2168,6 +2168,7 @@ namespace Opm
                 max_pressure = std::max(max_pressure, pressure_cell);
             }
         }
+        max_pressure = this->parallel_well_info_.communication().max(max_pressure);
         return max_pressure;
     }
 


lisajulia:
jenkins build this please

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1678,8 +1678,17 @@ namespace Opm
             const auto report = getWellConvergence(simulator, well_state, Base::B_avg_, deferred_logger, relax_convergence);
             converged = report.converged();
             if (this->parallel_well_info_.communication().size() > 1 && converged != this->parallel_well_info_.communication().min(converged)) {
-                OPM_THROW(std::runtime_error, fmt::format("Misalignment of the parallel simulation run in iterateWellEqWithSwitching - the well calculation succeeded on rank {} but failed on other ranks.", this->parallel_well_info_.communication().rank()));
+                std::cout << fmt::format("Misalignment of the parallel simulation run in iterateWellEqWithSwitching - the well calculation for well {} succeeded on rank {} but failed on other ranks.", this->name(), this->parallel_well_info_.communication().rank()) << std::endl;
```
blattms, line 1681: Why do we need cout and logging?

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1678,8 +1678,17 @@ namespace Opm
...
+                OpmLog::warning(fmt::format("Misalignment of the parallel simulation run in iterateWellEqWithSwitching - the well calculation  for well {} succeeded on rank {} but failed on other ranks.", this->name(), this->parallel_well_info_.communication().rank()));
             }
+            if (this->parallel_well_info_.communication().size() > 1 && converged != this->parallel_well_info_.communication().max(converged)) {
+                std::cout << fmt::format("Misalignment of the parallel simulation run in iterateWellEqWithSwitching - the well calculation  for well {} failed on rank {} but succeeded on other ranks.", this->name(), this->parallel_well_info_.communication().rank()) << std::endl;
```
blattms, line 1685: Why do we need cout and logging?

opm/simulators/wells/MultisegmentWell_impl.hpp
```diff
@@ -1678,8 +1678,17 @@ namespace Opm
...
+                OpmLog::warning(fmt::format("Misalignment of the parallel simulation run in iterateWellEqWithSwitching - the well calculation  for well {} failed on rank {} but succeeded on other ranks.", this->name(), this->parallel_well_info_.communication().rank()));
+            }
+            if (this->parallel_well_info_.communication().size() > 1 && (converged != this->parallel_well_info_.communication().min(converged) || converged != this->parallel_well_info_.communication().max(converged))) {
```
blattms, line 1688: I'm not sure that this will work. The second comparison might yield different results on different processes. In that case the latest communication will not be triggered on all processes.

So maybe move the communication outside of the logical expression and compare `this->parallel_well_info_.communication().min(converged)` with `this->parallel_well_info_.communication().max(converged)`?

blattms (changes requested):

lisajulia:
@blattms: Can you have another look? Thanks!

lisajulia:
@blattms: Can you have another look? Thanks!

lisajulia:
jenkins build this please

lisajulia:
> Looks good.
> 
> Please add a proper description about the bugfix if you think this should be part of the manual. Currently it is not clear that this only concerns distributed wells (not the default) and that it will throw if there are other bugs in our code.

I don't think this should be part of the manual, if all goes right, these error messages will never be seen my users but only by developers.

PR #6270, opened 2025-05-13
Making a GPUISTL aware ISTLSolver

kjetilly:
This pull request creates a GPUISTL version of the ISTLSolver class. The motivation for doing this is:

1) Performance: We will centralise GPU matrix allocation, meaning less memory requirements and fewer host to device copies.
2) Code cleanup: The current implementation of using a SolverAdapter works, and it keeps the rest of the code isolated, but it adds a lot of adapters that makes it harder to reason about the code when a larger part is on the GPU.
3) This will accommodate CPR on the GPU better.

To implement this change, I first added an `AbstractISTLSolver` interface, and then a `ISTLSolverRuntimeOptionProxy` which sole purpose is to allow the selection of ISTLSolver runtime. Due to the way the the current ISTLSolver is selected through TypeTags, the path of least resistance was to create a proxy that reads the `.json` file a selects the correct ISTLSolver subclass.

To keep things simple, I have added a new `backend` keyword to the JSON file, which defaults to `cpu`. To specify a GPU run with this PR, one would do

```JSON
{
    "solver" : "bicgstab",
    "backend" : "gpu",
    "preconditioner": {
        "type": "DILU"
    }
}
```

(naming up for debate).

while a CPU version would be 

```JSON
{
    "solver" : "bicgstab",
    "backend" : "cpu",
    "preconditioner": {
        "type": "DILU"
    }
}
```

or leaving out the `backend` argument since it defaults to `cpu`:

```JSON
{
    "solver" : "bicgstab",
    "preconditioner": {
        "type": "DILU"
    }
}
```

## Why draft?

The major things missing are:

1) Also support the GPUBridge ISTLSolver when needed.
2) Compilation in separate compilation units. Right now, the new ISTLSolver includes both the `PreconditionerFactory_impl.hpp` and `FlexibleSolver_impl.hpp` header files.
2) Suport for MPI.

## SolverAdapter is still supported

To keep things backwards compatible, I suggest we keep the `SolverAdapter` for now.


jakobtorben:
Very nice addition! I will have a look and test this out lcoally.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+#include <dune/istl/operators.hh>
+#include <opm/simulators/linalg/AbstractISTLSolver.hpp>
+#include <opm/simulators/linalg/ISTLSolver.hpp>
+
+#include <opm/simulators/linalg/FlexibleSolver.hpp>
+// TODO: These should be removed so that we avoid the compilation burden.
+//       That is, we should instantiate them in the cpp file instead.
+#include <opm/simulators/linalg/PreconditionerFactory_impl.hpp>
+#include <opm/simulators/linalg/FlexibleSolver_impl.hpp>
```
jakobtorben, line 28: I tried building it with AMGX support but get this error:

```sh
In file included from /home/jakob/code/opm-clean2/src/opm-simulators/opm/simulators/linalg/PreconditionerFactory_impl.hpp:58,
                 from /home/jakob/code/opm-clean2/src/opm-simulators/opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp:27,
                 from /home/jakob/code/opm-clean2/src/opm-simulators/opm/simulators/linalg/ISTLSolverRuntimeOptionProxy.hpp:23,
                 from /home/jakob/code/opm-clean2/src/opm-simulators/opm/simulators/flow/BlackoilModelNldd.hpp:47,
                 from /home/jakob/code/opm-clean2/src/opm-simulators/opm/simulators/flow/BlackoilModel.hpp:30,
                 from /home/jakob/code/opm-clean2/src/opm-simulators/opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp:44,
                 from /home/jakob/code/opm-clean2/src/opm-simulators/flow/flow_blackoil.cpp:23:
/home/jakob/code/opm-clean2/src/opm-simulators/opm/simulators/linalg/AmgxPreconditioner.hpp:32:10: fatal error: amgx_c.h: No such file or directory
   32 | #include <amgx_c.h>
      |          ^~~~~~~~~~
```
       
I think this is related to the include chain associated with the precondition factory.

kjetilly:
> There seems to be an increase in setup time for the linear solver that we need to look into.

We should look into it, but it seems like the overall simulation time has decreased, so this could be due to how things are measured, and maybe some of the setup time for the SolverAdapter is not actually measured.

kjetilly:
@jakobtorben I think the increase in setup time is, at least partially, due to `prepare` being included in the setup time. The old solver adapter transfers the GPU matrices and vectors in the call to `solve`, but the current version does the same in `prepare`.

jakobtorben:
Yes that makes sense. I think including the transfers in the linear solver setup, like in the current version, makes the most sense anyway.

Nevertheless, good to see a reduction in overall simulation time, while maintaining exactly the same number of iterations

kjetilly:
Jenkins build this please

kjetilly:

kjetilly:
I think the HIP build failure is due to some targets not seen the hipification as a dependency when they should, will look into it. It "works on my machine" with HIP, so that makes the dependency explanation more likely. On CUDA it builds fine.

kjetilly:

kjetilly:
Jenkins build this please

kjetilly:

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
kjetilly: This should work now.

kjetilly:

kjetilly:

kjetilly:
Jenkins build this please

jakobtorben:
I have tested the implementation in this PR extensively in https://github.com/OPM/opm-simulators/pull/6308, and it works as it should and fulfills all of the requirements a CPR-AMG two-level preconditioner requires.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,243 @@
...
+                OpmLog::warning(ss.str());
+                return true;
+            }
+        }
+        // Check for failure of linear solver.
+        if (!m_parameters.ignoreConvergenceFailure_ && !result.converged) {
+            const std::string msg("Convergence failure for linear solver.");
+            OPM_THROW_NOLOG(NumericalProblem, msg);
+        }
+
```
jakobtorben, line 182: Is it possible to move this out to ` checkConvergence(result);`  to clean it up and match the ISTLSolver implementation?

kjetilly:
Jenkins build this please

kjetilly:
Jenkins build this please

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
kjetilly: This has been refactored, though without using the state machine of checkConvergence of ISTLSolver.

CMakeLists.txt
```diff
@@ -544,6 +544,10 @@ set_property(TARGET moduleVersion PROPERTY POSITION_INDEPENDENT_CODE ON)
 add_library(MainDispatchDynamic OBJECT opm/simulators/flow/MainDispatchDynamic.cpp)
 set_property(TARGET MainDispatchDynamic PROPERTY POSITION_INDEPENDENT_CODE ON)
 
+if (CONVERT_CUDA_TO_HIP)
```
atgeirr, line 551: Why is this required for HIP but not CUDA? (Also similar in other places.)

opm/simulators/linalg/FlowLinearSolverParameters.cpp
```diff
@@ -150,6 +159,9 @@ void FlowLinearSolverParameters::registerParameters()
          "to determine valid platform IDs");
     Parameters::Register<Parameters::OpenclIluParallel>
         ("Parallelize ILU decomposition and application on GPU");
+    Parameters::Register<Parameters::LinearSolverAccelerator>
+        ("Choose the backend for the linear solver, usage: "
+         "'--linear-solver-accelerator=[cpu|gpu]'. The default is 'cpu'. ");
```
atgeirr: The default will automatically get printed on the help line, remove the last sentence.

opm/simulators/linalg/FlexibleSolver_impl.hpp
```diff
@@ -163,62 +166,74 @@ namespace Dune
         const int maxiter = prm.get<int>("maxiter", 200);
         const int verbosity = is_iorank ? prm.get<int>("verbosity", 0) : 0;
         const std::string solver_type = prm.get<std::string>("solver", "bicgstab");
-        if (solver_type == "bicgstab") {
-            linsolver_ = std::make_shared<Dune::BiCGSTABSolver<VectorType>>(*linearoperator_for_solver_,
+
+        if constexpr (Opm::is_gpu_operator_v<Operator>) {
+            if (solver_type == "bicgstab") {
```
atgeirr: Would instantiating LoopSolver (for example) with a gpu vector type fail? If not, the code could be identical for gpu and cpu all the way down to the umfpack and SolverAdapter-wrapped versions, so the `if constexpr` part could be a lot smaller.

opm/simulators/linalg/AbstractISTLSolver.hpp
```diff
@@ -0,0 +1,81 @@
...
+    using CommunicationType = Dune::Communication<int>;
+#endif
+
+    using SparseMatrixAdapter = GetPropType<TypeTag, Properties::SparseMatrixAdapter>;
+    using Vector = GetPropType<TypeTag, Properties::GlobalEqVector>;
+    using Matrix = typename SparseMatrixAdapter::IstlMatrix;
+
+    virtual ~AbstractISTLSolver() = default;
+    
+    virtual void eraseMatrix() = 0;
```
atgeirr, line 65: I think we should take a critical look at this interface, I think it has too many functions and the relations between them is not 100% clear.

This class in some way formalizes an interface that was formerly only indirectly defined as a concept (in the template sense), which is good I think, but part of why it is a little murky.

Which functions are we not using in practice from Flow?

opm/simulators/linalg/FlowLinearSolverParameters.hpp
```diff
@@ -52,9 +56,9 @@ template<class TypeTag>
 struct LinearSolverBackend<TypeTag, TTag::FlowIstlSolverParams>
 {
 #if COMPILE_GPU_BRIDGE
```
atgeirr: No need for this #if anymore.

opm/simulators/linalg/gpuistl/detail/FlexibleSolverWrapper.hpp
```diff
@@ -0,0 +1,65 @@
...
+public:
+    using AbstractSolverType = Dune::InverseOperator<Vector, Vector>;
+    using AbstractPreconditionerType = Dune::PreconditionerWithUpdate<Vector, Vector>;
+    using AbstractOperatorType = Dune::AssembledLinearOperator<Matrix, Vector, Vector>;
+
+    using AbstractOperatorPtrType = std::unique_ptr<AbstractOperatorType>;
+    using AbstractSolverPtrType = std::unique_ptr<AbstractSolverType>;
+
+    FlexibleSolverWrapper(const Matrix& matrix,
+                          bool parallel,
```
atgeirr, line 52: Make all `const` that can be.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,248 @@
...
+#ifndef OPM_ISTLSOLVERGPUISTL_HEADER_INCLUDED
+#define OPM_ISTLSOLVERGPUISTL_HEADER_INCLUDED
+
+#include <dune/istl/operators.hh>
+#include <memory>
+#include <opm/simulators/linalg/AbstractISTLSolver.hpp>
+#include <opm/simulators/linalg/ISTLSolver.hpp>
+
+#if USE_HIP
```
atgeirr, line 25: Similar to the question about the CONVERT_... macro, is it not possible to make GpuVector agnostic to cuda/hip?

opm/simulators/linalg/gpuistl/detail/FlexibleSolverWrapper.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include <dune/istl/operators.hh>
+#include <dune/istl/solver.hh>
+#include <opm/simulators/linalg/PreconditionerWithUpdate.hpp>
+#include <opm/simulators/linalg/PropertyTree.hpp>
+
+namespace Opm::gpuistl::detail
+{
+template <class Matrix, class Vector, class Comm>
+class FlexibleSolverWrapper
```
atgeirr, line 41: The need for this wrapper is not clear, please explain. (And make it comment here.)

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,248 @@
...
+    using SolverType = Opm::gpuistl::detail::FlexibleSolverWrapper<GPUMatrix, GPUVector, CommunicationType>;
+
+    /// Construct a system solver.
+    /// \param[in] simulator   The opm-models simulator object
+    /// \param[in] parameters  Explicit parameters for solver setup, do not
+    ///                        read them from command line parameters.
+    /// \param[in] forceSerial If true, will set up a serial linear solver only,
+    ///                        local to the current rank, instead of creating a
+    ///                        parallel (MPI distributed) linear solver.
+    ISTLSolverGPUISTL(const Simulator& simulator,
```
jakobtorben, line 83: If you doing another round here, can you also add the print to json functionality, matching ISTLSolver?

CMakeLists.txt
kjetilly: We only wait for the hipification process if we are compiling HIP. That is, we need to wait for the HIP header files to be generated before the rest of the code can compile. HIP header file includes anything in gpuistl, say.

opm/simulators/linalg/AbstractISTLSolver.hpp
kjetilly: So I think every method is used (list of usage is not complete):

`eraseMatrix()`:
- `opm/models/discretization/common/fvbasediscretization.hh:1777:        linearizer_->eraseMatrix();`
- `opm/models/nonlinear/newtonmethod.hh:463:    { linearSolver_.eraseMatrix(); }`

`setActiveSolver(...)`:
-`opm/simulators/flow/BlackoilModel_impl.hpp:484:            linSolver.setActiveSolver(solver);`

`numAvailableSolvers()`:
- `opm/simulators/flow/BlackoilModel_impl.hpp:470:    const int numSolvers = linSolver.numAvailableSolvers();
`

`prepare(...)`:
- `opm/simulators/flow/BlackoilModel_impl.hpp:486:            linSolver.prepare(jacobian, residual);`

`setResidual(...)`
- `opm/simulators/flow/BlackoilModelNldd.hpp:628:        linsolver.setResidual(res);`
- `opm/simulators/flow/BlackoilModel_impl.hpp:489:            linSolver.setResidual(residual);`
- `opm/simulators/flow/BlackoilModel_impl.hpp:514:        linSolver.setResidual(residual);`
- `opm/models/nonlinear/newtonmethod.hh:278:                linearSolver_.setResidual(residual);`

`getResidual(...)`
- `opm/models/nonlinear/newtonmethod.hh:279:                linearSolver_.getResidual(residual);`

`setMatrix(...)`
- `opm/models/nonlinear/newtonmethod.hh:312:                linearSolver_.setMatrix(jacobian);`

`solve(...)` (probably has a lot more uses, just took one)
- `opm/models/nonlinear/newtonmethod.hh:314:                bool converged = linearSolver_.solve(solutionUpdate);`

`iterations()`
- `opm/simulators/flow/BlackoilModel.hpp:176:    { return simulator_.model().newtonMethod().linearSolver().iterations (); }`

`comm()` (usage breaks Law of Demeter)
- `opm/simulators/flow/BlackoilModelNldd.hpp:341:            const auto* ccomm = model_.simulator().model().newtonMethod().linearSolver().comm();` 

`getSolveCount`
- `opm/simulators/flow/BlackoilModel_impl.hpp:471:    if (numSolvers > 1 && (linSolver.getSolveCount() % 100 == 0)) {`


Additionally, some of the methods that are actually not needed by the ISTLSolver (`setResidual`, `setMatrix`) are needed by other implementations such as a the `SuperLUBackend` class in `opm/simulators/linalg/superlubackend.hh`

So while I agree that there are more functions than strictly needed, it is a bit more complicated to tidy this up, and probably outside the scope of this PR.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
kjetilly: This is a future todo, but it is a bit complicated than it looks. GPUVector in itself could be possible to have an agnostic interface (one immediate issue is the `cudaStream_t` argument in some functions that needs hipifications, granted, this can be handled, but...), but we want to fix a larger issue first:

Essentially we could hipify the files to say `build/opm/simulators/linalg/gpuistl` instead of the current `build/opm/simulators/linalg/gpuistl_hip`, *but* we would have to make sure the `build` include gets precedence over the `opm-simulators`include everywhere, and make some extra sanity checks that no non-hipified files gets included by accident (some will give a compilation error, but others will give very difficult linker error, but in the long run, this is really the solution we want.

opm/simulators/linalg/gpuistl/detail/FlexibleSolverWrapper.hpp
kjetilly:
```C++
/**
 * @brief FlexibleSolverWrapper is compilational trick to reduce compile time overhead
 * 
 * Essentially this class acts as a direct interface towards the FlexibleSolver class, but
 * we want to have a clear wall between the ISTLSolverGPUISTL and the FlexibleSolver to avoid
 * compilational overhead (the former is based on TypeTag the latter on operators). 
 *
 * This wrapper handles the mapping from matrices to concrete operator types.
 */
 ```

File: opm/simulators/linalg/FlexibleSolver_impl.hpp (hunk 5 above)
Comment (kjetilly) [reply opm/simulators/linalg/FlexibleSolver_impl.hpp]: instantiated loop solver and gmres. Note that the error checking gets a bit more involved now, but I think the selected path is probably the cleanest (essentially checking if `linsolver_` is a `nullptr` at the end)

File: opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp (hunk 11 above)
Comment (kjetilly) [reply opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp:83]: Done.

File: opm/simulators/linalg/setupPropertyTree.cpp [hunk 12]
```diff
@@ -270,7 +270,12 @@ setupILU([[maybe_unused]] const std::string& conf, const FlowLinearSolverParamet
     prm.put("maxiter", p.linear_solver_maxiter_);
     prm.put("verbosity", p.linear_solver_verbosity_);
     prm.put("solver", getSolverString(p));
-    prm.put("preconditioner.type", "ParOverILU0"s);
+    if (p.linear_solver_accelerator_ == "gpu") {
+        // TODO: We could add ParOverILU0 as an alias in the GPU path to simplify this.
+        prm.put("preconditioner.type", "OPMILU0"s);
+    } else {
+        prm.put("preconditioner.type", "ParOverILU0"s);
+    }
```
jakobtorben: For this to work out of the box with NLDD, you need to add this extra statement in the check:

```suggestion
    if (p.linear_solver_accelerator_ == "gpu" && !p.is_nldd_local_solver_) {
        // TODO: We could add ParOverILU0 as an alias in the GPU path to simplify this.
        prm.put("preconditioner.type", "OPMILU0"s);
    } else {
        prm.put("preconditioner.type", "ParOverILU0"s);
    }
```

kjetilly:
Jenkins build this please

opm/simulators/linalg/FlexibleSolver_impl.hpp
```diff
@@ -180,47 +188,54 @@ namespace Dune
...
-        } else if (solver_type == "flexgmres") {
-            int restart = prm.get<int>("restart", 15);
-            linsolver_ = std::make_shared<Dune::RestartedFlexibleGMResSolver<VectorType>>(*linearoperator_for_solver_,
-                                                                                          *scalarproduct_,
-                                                                                          *preconditioner_,
-                                                                                          tol,// desired residual reduction factor
-                                                                                          restart,
-                                                                                          maxiter, // maximum number of iterations
-                                                                                          verbosity);
+                                                                            *scalarproduct_,
```
atgeirr: Looks like indentation mismatch with the line above.

opm/simulators/linalg/FlexibleSolver_impl.hpp
```diff
@@ -180,47 +188,54 @@ namespace Dune
...
+                        verbosity,
+                        comm));
+                }
+            }
+        }
+        #endif
+        if (!linsolver_) {
             OPM_THROW(std::invalid_argument,
-                      "Properties: Solver " + solver_type + " not known.");
+                    "Properties: Solver " + solver_type + " not known.");
```
atgeirr: Looks like your editor or setup makes strange indentation changes for no reason?

atgeirr (approved):
Just indentation quibbles now, as well as one "resolved" comment unresolved (const-ifying some function parameters).

Can be merged when these minor things are fixed.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,253 @@
...
+            m_x = std::make_unique<GpuVector<real_type>>(x);
+        } else {
+            m_x->copyFromHost(x);
+        }
+        m_gpuSolver->apply(*m_x, *m_rhs, result);
+    
+        m_x->copyToHost(x);
+
+        ++m_solveCount;
+
```
atgeirr, line 242: Too generous with the newlines... Avoid having more blanks inside functions that between functions.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,253 @@
...
+        // Check for failure of linear solver.
+        if (!m_parameters.ignoreConvergenceFailure_ && !result.converged) {
+            const std::string msg("Convergence failure for linear solver.");
+            OPM_THROW_NOLOG(NumericalProblem, msg);
+        }
+
+        return result.converged;
+    }
+
+    void updateMatrix(const Matrix& M) {
```
atgeirr: Brace at start of next line for functions and classes.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,253 @@
...
+        return nullptr;
+    }
+
+    int getSolveCount() const override
+    {
+        return m_solveCount;
+    }
+
+private:
+    bool checkConvergence(const Dune::InverseOperatorResult& result) const
```
atgeirr, line 275: This approach, avoiding having a converged_ member is a good improvement, please also make the identical change in the other ISTLSolver variants! Could be done by putting this implementation as a protected helper in the base class.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,253 @@
...
+            m_matrix.reset(new auto(GPUMatrix::fromMatrix(M)));
+            std::function<GPUVector()> weightsCalculator = {};
+            const bool parallel = false;
+            m_gpuSolver = std::make_unique<SolverType>(
+                *m_matrix, parallel, m_propertyTree, pressureIndex, weightsCalculator, m_forceSerial, nullptr);
+        } else {
+            m_matrix->updateNonzeroValues(M);
+        }
+    
+        m_gpuSolver->update();
```
atgeirr: Is it necessary to call update() when we construct it the first time?

atgeirr (commented):
Now I am done... Looks very good, some minor nitpicks and one request for refactoring.

opm/simulators/linalg/AbstractISTLSolver.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+    virtual void setActiveSolver(int num) = 0;
```
blattms, line 74: Please add some documentation for fellow develpers to understand this. Holds for other places, too.

opm/simulators/linalg/FlexibleSolver_impl.hpp
```diff
@@ -163,6 +167,10 @@ namespace Dune
         const int maxiter = prm.get<int>("maxiter", 200);
         const int verbosity = is_iorank ? prm.get<int>("verbosity", 0) : 0;
         const std::string solver_type = prm.get<std::string>("solver", "bicgstab");
+
+        
+        // make sure it is nullptr at the start (used for error checking in the end)
+        linsolver_.reset();
```
blattms: Out of curiosity: Why should this not be null to begin with?

opm/simulators/linalg/FlexibleSolver_impl.hpp
```diff
@@ -180,47 +188,54 @@ namespace Dune
...
-                                                                                          verbosity);
+                                                                            *scalarproduct_,
+                                                                            *preconditioner_,
+                                                                            tol,// desired residual reduction factor
+                                                                            restart,
+                                                                            maxiter, // maximum number of iterations
+                                                                            verbosity);
+         
+        } else {
+            if constexpr (!Opm::is_gpu_operator_v<Operator>) {
```
blattms, line 202: I cannot find the else branch to this. Is that not needed? What happens if the condition is false?

opm/simulators/linalg/FlexibleSolver_impl.hpp
```diff
@@ -234,11 +249,13 @@ namespace Dune
     recreateDirectSolver()
     {
 #if HAVE_SUITESPARSE_UMFPACK
-        if constexpr (std::is_same_v<typename VectorType::field_type, float>) {
-            OPM_THROW(std::invalid_argument, "UMFPack cannot be used with floats");
-        } else {
-        using MatrixType = std::remove_const_t<std::remove_reference_t<decltype(linearoperator_for_solver_->getmat())>>;
-        linsolver_ = std::make_shared<Dune::UMFPack<MatrixType>>(linearoperator_for_solver_->getmat(), 0, false);
+        if constexpr (!Opm::is_gpu_operator_v<Operator>) {
```
blattms, line 255: Again: No else branch, why?

opm/simulators/linalg/FlowLinearSolverParameters.hpp
```diff
@@ -115,6 +115,7 @@ struct FlowLinearSolverParameters
     int gpu_device_id_;
     int opencl_platform_id_;
     bool opencl_ilu_parallel_;
+    std::string linear_solver_accelerator_;
```
blattms: Should we rather use an enum for this and save some string comparisons?

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,253 @@
...
+
+#include <opm/simulators/linalg/gpuistl/detail/FlexibleSolverWrapper.hpp>
+#include <opm/simulators/linalg/printlinearsolverparameter.hpp>
+
+namespace Opm::gpuistl
+{
+
+
+template <class TypeTag>
+class ISTLSolverGPUISTL : public AbstractISTLSolver<TypeTag>
```
blattms, line 52: MAybe add a comment like for the other one?

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,253 @@
...
+    explicit ISTLSolverGPUISTL(const Simulator& simulator)
+        : ISTLSolverGPUISTL(simulator, FlowLinearSolverParameters(), false)
+    {
+    }
+
+
+
+    void eraseMatrix() override
+    {
+        // Nothing, this is the same as the ISTLSolver
```
blattms, line 111: I don't understand this explanation. what does it mean?

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,253 @@
...
+            updateMatrix(M);
+            updateRhs(b);
+        }
+        OPM_CATCH_AND_RETHROW_AS_CRITICAL_ERROR("This is likely due to a faulty linear solver JSON specification. "
+                                                "Check for errors related to missing nodes.");
+    }
+
+    void setResidual(Vector&) override
+    {
+        // Should be handled in prepare() instead.
```
blattms, line 177: Should we add some checks to prevent misusage (Calling it with another vector)?

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,253 @@
...
+    {
+        if (!m_rhs) {
+            OPM_THROW(std::runtime_error, "m_rhs not initialized, prepare(matrix, rhs); needs to be called");
+        }
+        m_rhs->copyToHost(b);
+    }
+
+    void setMatrix(const SparseMatrixAdapter&) override
+    {
+        // Should be handled in prepare() instead.
```
blattms, line 203: Should we add some checks to prevent misusage (Calling it with another matrix)?

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,253 @@
...
+
+    int iterations() const override
+    {
+        return m_lastSeenIterations;
+    }
+
+    const CommunicationType* comm() const override
+    {
+        // TODO: Implement this if needed
+        return nullptr;
```
blattms: Seems dangerous. Why not throw?

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -0,0 +1,253 @@
...
+            OPM_THROW_NOLOG(NumericalProblem, msg);
+        }
+
+        return result.converged;
+    }
+
+    void updateMatrix(const Matrix& M) {
+        if (!m_matrix) {
+
+            m_matrix.reset(new auto(GPUMatrix::fromMatrix(M)));
```
blattms, line 284: What exactly will this do? Call the copy constructor or move constructor (if the object has one)?
It seems like copy assignment is deleted from GPUMatrix, which might prevent generating a move constructor. I am not an expert here, though.
Hence the question might be how smart the compiler is here.

I know that this and similar approaches have been used at various places elsewhere. Even better to double check, though

blattms (commented):
Just my two cents.

One major question:
Do we really have to mix file naming schemes (FlexibleSolver_gpu_instantiate.cpp instead of CamelCase FlexibleSolverGpuInstantiate.cpp). Seems rather inconsistent.

opm/simulators/linalg/AbstractISTLSolver.hpp
kjetilly: Done

opm/simulators/linalg/FlexibleSolver_impl.hpp
kjetilly: The main reason is that we have slightly changed the logic of initSolver to check the state of `linsolver_` at the end of the function. We are now supporting GPU operators directly, but because the set of solvers supported directly on GPU operators is more limited than on the CPU, we needed to have some extra `if constexpr`s that do not have easy `else` cases.

opm/simulators/linalg/FlexibleSolver_impl.hpp
kjetilly: If the condition is false, the `linsolver_` will not be set, and the end of the function will throw an exception.

opm/simulators/linalg/FlexibleSolver_impl.hpp
kjetilly: See above

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
kjetilly: Added documentation if that is what you meant?

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
kjetilly: Some of the functions used in the solverbackend concept ("implemented" by `ISTLSolver` and say `SuperLUBackend`) are blank in some of the implementation. In this case, `Opm::ISTLSolver` does nothing in a call to `eraseMatrix()`. I have added documentation in the `AbstractISTLSolver` class describing the fact that this might not do anything.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
kjetilly: I am not sure. At the moment, this is the exact same behavior as we have had in `ISTLSolver` for a very long time. I would rather remove the calls to `setResidual` and `setMatrix` and simply rely on `prepare`. However, in the actual code, both functions are called, and other backends (`SuperLUBackend` for instance) uses this call. Anyway, I suggest we leave it as is for the current PR and rather to a subsequent PR where we clean up the interface a bit.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
kjetilly: See above. I would prefer removing them entirely to be honest.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
kjetilly: At the very least this will be something akin to RVO. It is a non-named variable and the only thing that happens here is that a stack allocated object (consisting of 2 64 bit integers) is transferred to the heap. 

[A simple test program confirms that no copy logic or move logic is invoked:](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:2,endLineNumber:30,positionColumn:1,positionLineNumber:1,selectionStartColumn:2,selectionStartLineNumber:30,startColumn:1,startLineNumber:1),source:'%23include+%3Cprint%3E%0A%0A%0Astruct+Car+%7B%0A++++static+Car+create()+%7B%0A++++++++return+Car%7B%7D%3B%0A++++%7D%0A%0A++++Car(const+Car%26)+%7B%0A++++++++std::println(%22copy+ctor%22)%3B%0A++++%7D%0A%0A++++Car(Car%26%26)+%7B%0A++++++++std::println(%22move+ctor%22)%3B%0A++++%7D%0A%0A++++Car%26+operator%3D(const+Car%26)+%7B%0A++++++++std::println(%22operator%3D%22)%3B%0A++++++++return+*this%3B%0A++++%7D%0A%0A++++Car()+%3D+default%3B%0A%7D%3B%0A%0Aint+main(int,+char**)+%7B%0A++++std::println(%22Before%22)%3B%0A++++auto+car+%3D+new+auto(Car::create())%3B%0A++++std::println(%22After%22)%3B%0A++++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:51.3246733021026,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:g151,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B23',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+15.1+(Editor+%231)',t:'0')),header:(),k:48.67532669789741,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+14.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+15.1+(Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:48.67532669789741,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

```C++
// Needs -std=c++23
#include <print>


struct Car {
    static Car create() {
        return Car{};
    }

    Car(const Car&) {
        std::println("copy ctor");
    }

    Car(Car&&) {
        std::println("move ctor");
    }

    Car& operator=(const Car&) {
        std::println("operator=");
        return *this;
    }

    Car() = default;
};

int main(int, char**) {
    std::println("Before");
    auto car = new auto(Car::create());
    std::println("After");
    return 0;
}
```

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
kjetilly: Agree. It now throws. The next PR in this series will implement this properly, though (ie. MPI support).

kjetilly:
Jenkins build this please.

opm/simulators/linalg/FlowLinearSolverParameters.hpp
kjetilly: Done.

kjetilly:
Jenkins build this please

kjetilly:
Jenkins build this

bska:
jenkins build this please

kjetilly:

atgeirr:
Looking over the most recent remarks, it seems they all are resolved, except for a few open questions about the setMatrix() etc. methods, if they should be removed or not. I merge this now as is. We should discuss how to improve the interfaces afterwards.

PR #6272, opened 2025-05-14
unifying instantiation of variants related to Indices

GitPaean:
There are some different orders and numbers, it causes confusion when modifying. 

With the first commit of this PR, they are all made the same, so hopefully, it is easier. 

In the second commit, I put the macro definition in a file, while I think the approach might need refinement.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -181,6 +181,7 @@ list (APPEND MAIN_SOURCE_FILES
   opm/simulators/utils/ComponentName.cpp
   opm/simulators/utils/DeferredLogger.cpp
   opm/simulators/utils/FullySupportedFlowKeywords.cpp
+  opm/simulators/utils/InstantiationIndicesMacros.hpp
   opm/simulators/utils/ParallelFileMerger.cpp
   opm/simulators/utils/ParallelRestart.cpp
   opm/simulators/utils/PartiallySupportedFlowKeywords.cpp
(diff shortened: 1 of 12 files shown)


GitPaean:
jenkins build this please

CMakeLists_files.cmake
```diff
@@ -181,6 +181,7 @@ list (APPEND MAIN_SOURCE_FILES
   opm/simulators/utils/ComponentName.cpp
   opm/simulators/utils/DeferredLogger.cpp
   opm/simulators/utils/FullySupportedFlowKeywords.cpp
+  opm/simulators/utils/InstantiationMarcros.hpp
```
akva2, line 184: Macros. Also I think the name is a bit too generic, and/or the macros themself is too generic. maybe INSTANTIATE_TYPE_INDICES or the likes.

GitPaean: Thanks for the comments, it is indeed too generic and automatic name changing changed a lot of macro usage. I had to roll back to do it manually. 

I updated it now and please have a look.

GitPaean:
jenkins build this please

opm/simulators/utils/InstantiationIndicesMacros.hpp
```diff
@@ -0,0 +1,63 @@
+/*
```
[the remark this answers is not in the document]
GitPaean: Good suggestion. A simple documentation was added.

GitPaean:
jenkins build this please

GitPaean:
Let me know if there is more you would like me to do. @akva2

akva2 (approved):

GitPaean:
Thanks for the approval. I am getting it in now.

PR #6274, opened 2025-05-14
FvBase(ExtensiveQuantites|FdLocalLinearizer): some janitoring

akva2:
Some tidy ups

### DIFF
--- a/opm/models/discretization/common/fvbasefdlocallinearizer.hh
+++ b/opm/models/discretization/common/fvbasefdlocallinearizer.hh
@@ -163,13 +163,11 @@ private:
 
     using LocalEvalBlockVector = typename LocalResidual::LocalEvalBlockVector;
 
+public:
+    FvBaseFdLocalLinearizer() = default;
+
     // copying local residual objects around is a very bad idea
     FvBaseFdLocalLinearizer(const FvBaseFdLocalLinearizer&) = delete;
-
-public:
-    FvBaseFdLocalLinearizer()
-        : internalElemContext_{}
-    { }
 
     /*!
      * \brief Register all run-time parameters for the local jacobian.
@@ -498,15 +496,15 @@ protected:
         }
     }
 
-    Simulator *simulatorPtr_;
+    Simulator* simulatorPtr_{};
 
-    std::unique_ptr<ElementContext> internalElemContext_;
+    std::unique_ptr<ElementContext> internalElemContext_{};
 
-    LocalEvalBlockVector residual_;
-    LocalEvalBlockVector derivResidual_;
-    ScalarLocalBlockMatrix jacobian_;
+    LocalEvalBlockVector residual_{};
+    LocalEvalBlockVector derivResidual_{};
+    ScalarLocalBlockMatrix jacobian_{};
 
-    LocalResidual localResidual_;
+    LocalResidual localResidual_{};
 };
 
 } // namespace Opm
--- a/opm/models/discretization/common/fvbasefdlocallinearizer.hh
+++ b/opm/models/discretization/common/fvbasefdlocallinearizer.hh
@@ -50,6 +50,7 @@
 #include <memory>
 
 namespace Opm {
+
 // forward declaration
 template<class TypeTag>
 class FvBaseFdLocalLinearizer;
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this please

opm/models/discretization/common/fvbasefdlocallinearizer.hh
```diff
@@ -75,7 +82,8 @@ template<class TypeTag>
 struct BaseEpsilon<TypeTag, TTag::FiniteDifferenceLocalLinearizer>
 {
     using type = GetPropType<TypeTag, Properties::Scalar>;
-    static constexpr type value = std::max<type>(0.9123e-10, std::numeric_limits<type>::epsilon()*1.23e3);
+    static constexpr type value = std::max<type>(0.9123e-10,
+                                                 std::numeric_limits<type>::epsilon() * 1.23e3);
```
bska, line 86: I usually dislike explicit template arguments, but here the alternative is **possibly** worse, e.g.,
```C++
std::max(static_cast<type>(0.9123e-10),
         std::numeric_limits<type>::epsilon() * static_cast<type>(1.23e3));
```
What do you think?

opm/models/discretization/common/fvbasefdlocallinearizer.hh
```diff
@@ -157,26 +165,11 @@ private:
...
-#endif
-public:
-    FvBaseFdLocalLinearizer()
-        : internalElemContext_(0)
-    { }
+    FvBaseFdLocalLinearizer() = default;
 
-    ~FvBaseFdLocalLinearizer()
-    { delete internalElemContext_; }
+    // copying local residual objects around is a very bad idea
```
bska, line 171: I suppose this comment (now) could be something along the lines of
```C++
// Disable copying for efficiency reasons
```
since, at least in principle, we could implement an actual copy constructor with value semantics.

opm/models/discretization/common/fvbasefdlocallinearizer.hh
```diff
@@ -247,9 +239,9 @@ public:
         localResidual_.eval(residual_, elemCtx);
 
         // calculate the local jacobian matrix
-        size_t numPrimaryDof = elemCtx.numPrimaryDof(/*timeIdx=*/0);
-        for (unsigned dofIdx = 0; dofIdx < numPrimaryDof; dofIdx++) {
-            for (unsigned pvIdx = 0; pvIdx < numEq; pvIdx++) {
+        const std::size_t numPrimaryDof = elemCtx.numPrimaryDof(/*timeIdx=*/0);
+        for (unsigned dofIdx = 0; dofIdx < numPrimaryDof; ++dofIdx) {
+            for (unsigned pvIdx = 0; pvIdx < numEq; ++pvIdx) {
```
bska, line 244: If you're changing these lines anyway, maybe we could write it as
```C++
for (auto dofIdx = 0*numPrimaryDof; ...) {
    for (auto pvIdx = 0*numEq; ...) {
```
instead?  That way, we don't incur type conversion on every comparison with the upper bounds.

bska (commented):
Much appreciated.  Just some ancillary comments about potential issues I noticed whilst looking at this.

opm/models/discretization/common/fvbasefdlocallinearizer.hh
akva2: Yeah, I've been here before, but decided to just leave it.

If the explicit template argument is not given, it  for some reason gives the error 
```

/home/akva/kode/opm/opm-simulators/opm/models/discretization/common/fvbasefdlocallinearizer.hh:85:43: error: call to non-‘constexpr’ function ‘quad std::max(quad, quad)’
   85 |     static constexpr type value = std::max(type{0.9123e-10},
      |                                   ~~~~~~~~^~~~~~~~~~~~~~~~~~
   86 |                                            std::numeric_limits<type>::epsilon() * type{1.23e3});
      |                                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/akva/kode/opm/opm-simulators/examples/co2injection_flash_vcfv.cpp:32:
/home/akva/kode/opm/opm-common/opm/material/common/quad.hpp:289:13: note: ‘quad std::max(quad, quad)’ declared here
  289 | inline quad max(quad a, quad b)
```
while the old code somehow does not complain. I don't understand why it fails, it should basically be identical.

If I update quad.hpp and mark the function constexpr it works.

opm/models/discretization/common/fvbasefdlocallinearizer.hh
akva2: problem is you are just moving the type conversions into the function calls, where both arguments are unsigned. for the first it would do ```std::size_t -> unsigned```, for the second it would do ```int -> unsigned```. So it makes no difference basically. Anyways, I have applied the requested change.

bska (approved):

PR #6275, opened 2025-05-14
move alq_state to singleWellState

totto82:
First step to facilitate #6245

### DIFF
--- a/opm/simulators/wells/ALQState.cpp
+++ b/opm/simulators/wells/ALQState.cpp
@@ -66,12 +66,7 @@ void ALQState<Scalar>::set(Scalar value)
 template<class Scalar>
 bool ALQState<Scalar>::oscillation() const
 {
-    auto inc_count = this->alq_increase_count_;
-    if (inc_count == 0)
-        return false;
-
-    auto dec_count = this->alq_decrease_count_;
-    return dec_count >= 1;
+    return this->alq_increase_count_ > 0 && this->alq_decrease_count_ >= 1;
 }
 
 template<class Scalar>
--- a/opm/simulators/wells/ALQState.cpp
+++ b/opm/simulators/wells/ALQState.cpp
@@ -32,146 +32,74 @@ template<class Scalar>
 ALQState<Scalar> ALQState<Scalar>::serializationTestObject()
 {
     ALQState result;
-    result.current_alq_ = {{"test1", 1.0}};
-    result.default_alq_ = {{"test2", 2.0}, {"test3", 3.0}};
-    result.alq_increase_count_= {{"test4", 4}};
-    result.alq_decrease_count_= {{"test5", 5}};
-    result.debug_counter_ = 6;
-
+    result.current_alq_ = {1.0};
+    result.default_alq_ = {2.0};
+    result.alq_increase_count_= {4};
+    result.alq_decrease_count_= {5};
     return result;
 }
 
 template<class Scalar>
-void ALQState<Scalar>::insert(const std::string& wname)
+Scalar ALQState<Scalar>::get() const
 {
-    this->current_alq_.emplace(wname, 0.0);
+    if (this->current_alq_)
+        return *this->current_alq_;
+
+    return this->default_alq_;
 }
 
 template<class Scalar>
-Scalar ALQState<Scalar>::get(const std::string& wname) const
+void ALQState<Scalar>::update_default(Scalar value)
 {
-    auto iter = this->current_alq_.find(wname);
-    if (iter != this->current_alq_.end())
-        return iter->second;
-
-    auto default_iter = this->default_alq_.find(wname);
-    if (default_iter != this->default_alq_.end())
-        return default_iter->second;
-
-    throw std::logic_error("No ALQ value registered for well: " + wname);
-}
-
-template<class Scalar>
-void ALQState<Scalar>::update_default(const std::string& wname, Scalar value)
-{
-    auto default_iter = this->default_alq_.find(wname);
-    if (default_iter == this->default_alq_.end() || default_iter->second != value) {
-        this->default_alq_.insert_or_assign(wname, value);
-        this->current_alq_.insert_or_assign(wname, value);
+    if (default_alq_ != value) {
+        this->default_alq_ = value;
+        this->current_alq_ = value;
     }
 }
 
 template<class Scalar>
-void ALQState<Scalar>::set(const std::string& wname, Scalar value)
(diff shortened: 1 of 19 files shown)


opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -963,7 +963,7 @@ computeNetworkPressures(const Network::ExtNetwork& network,
 
                     if (well.isInjector() || !well_state.isOpen(wellname)) continue;
                     const Scalar efficiency = well.getEfficiencyFactor(/*network*/ true) * well_state.getGlobalEfficiencyScalingFactor(wellname);
-                    node_inflows[node][BlackoilPhases::Vapour] += well_state.getALQ(wellname) * efficiency;
+                    node_inflows[node][BlackoilPhases::Vapour] += well_state.well(wellname).alq_state.get() * efficiency;
                 }
```
totto82, line 967: I assumed this would not work in parallel, but Jenkins seems to be happy. Do we have a test case that exposes this feature? I.e. gaslift in network.

GitPaean: It does not look that we have in the regression tests or parallel tests, but we might have some examples in the opm-tests. Maybe it is a good time if you can add one based on one of them.

totto82: Do you know which test is relevant?

GitPaean: I thought you work with gas lift more, you might know more? I have not tested much. If it is not relevant cases in opm-tests, maybe you can add one from your development?

totto82: I just tested with NETWORK-01_STANDARD where I added 'FLO' for item 6. I.e. with the latest change it should also work in parallel for item6 = FLO.

totto82:

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -963,7 +964,13 @@ computeNetworkPressures(const Network::ExtNetwork& network,
 
                     if (well.isInjector() || !well_state.isOpen(wellname)) continue;
                     const Scalar efficiency = well.getEfficiencyFactor(/*network*/ true) * well_state.getGlobalEfficiencyScalingFactor(wellname);
-                    node_inflows[node][BlackoilPhases::Vapour] += well_state.getALQ(wellname) * efficiency;
+                    Scalar alq = 0.0;
+                    const auto& well_index = well_state.index(well.name());
+                    if (well_index.has_value() && well_state.wellIsOwned(well_index.value(), well.name())) {
+                        alq += well_state.well(wellname).alq_state.get();
+                    }
+                    alq = comm.sum(alq);
```
GitPaean, line 972: I am not totally familiar how the parallel working here and did not test it, but from the code, it looks worrying. 

Very likely, you should `comm.sum()` outside the for loop `for (const std::string& wellname : group.wells())` to avoid the extra communications. 

And then update the `node_inflows[node][BlackoilPhases::Vapour]` outside the for loop with the summation. 

Please let me know what you think.

opm/simulators/wells/ALQState.cpp
```diff
@@ -32,146 +32,74 @@ template<class Scalar>
...
-int ALQState<Scalar>::get_debug_counter()
+void ALQState<Scalar>::set(Scalar value)
 {
-    return this->debug_counter_;
+    this->current_alq_ = value;
 }
 
 template<class Scalar>
-int ALQState<Scalar>::update_debug_counter()
+bool ALQState<Scalar>::oscillation() const
```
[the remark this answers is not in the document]
totto82: Again. I just copied the old code. But the changes you suggest are both trivial and a good improvement.  So I updated the code accordingly.

opm/simulators/wells/ALQState.hpp
```diff
@@ -22,6 +22,8 @@
 
 #include <map>
 #include <string>
+#include <optional>
```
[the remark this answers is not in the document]
totto82: done

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -863,10 +863,7 @@ updateEclWellsConstraints(const int              timeStepIdx,
         auto& ws = this->wellState().well(wellIdx);
 
         ws.updateStatus(well.getStatus());
-        auto switchToProducer = ws.update_type_and_targets(well, st);
-        if (switchToProducer) {
```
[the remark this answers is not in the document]
totto82: I have updated update_type_and_targets accordingly.

opm/simulators/wells/GasLiftSingleWellGeneric.cpp
```diff
@@ -1423,7 +1423,7 @@ runOptimizeLoop_(bool increase)
     }
     std::optional<bool> increase_opt;
     if (success) {
-        this->well_state_.gliftUpdateAlqIncreaseCount(this->well_name_, increase);
+        this->well_state_.well(this->well_name_).alq_state.update_count(increase);
```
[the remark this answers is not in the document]
totto82: This will be cleaner when the complete move from WellState to SingleWellState is done. All the difficult parts are done so it shouldn't take too long.

opm/simulators/wells/WellGroupHelpers.cpp
totto82: Thanks! I moved the communication outside the well loop.

opm/simulators/wells/ALQState.cpp
```diff
@@ -32,146 +32,74 @@ template<class Scalar>
...
-    if (default_iter != this->default_alq_.end())
-        return default_iter->second;
-
-    throw std::logic_error("No ALQ value registered for well: " + wname);
+    return this->default_alq_;
 }
 
 template<class Scalar>
-void ALQState<Scalar>::update_default(const std::string& wname, Scalar value)
+void ALQState<Scalar>::update_default(Scalar value)
```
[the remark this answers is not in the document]
GitPaean: Yes, please come with a following-up PR. This function name is misleading and dangerous, which can be reflecting some design issue.  It is good we improve it when we are there.

GitPaean (approved):
Thanks for the update. Assuming the results are checked and tested. I am getting it in. 

It was suggested to add a regression test for this development and also some refactoring should be done in a following-up PR to improve the function design.

PR #6280, opened 2025-05-15
Further updates and fixes to general controller

erikhide:
This PR includes a different reduction in time step size if tolerance test fails (than if we get numerical difficulties), the possibility to set controller parameters for the general 3rd order controller, the option to use control error filtering as tolerance test instead of comparing relative change directly to the tolerance, as well as adjustments/fixes to the general 3rd order controller code.

### DIFF
--- a/opm/simulators/timestepping/TimeStepControl.cpp
+++ b/opm/simulators/timestepping/TimeStepControl.cpp
@@ -333,19 +333,23 @@ namespace Opm
         // Use an I controller after report time steps
         if (substepTimer.currentStepNum() < 3)
         {
-            controllerVersion_ = "I-controller";
+            controllerVersion_ = 0; // "I-controller"
             const double newDt = dt * timeStepFactor(errors_, timeSteps_);
-            if( verbose_ )
+            if (verbose_)
+            {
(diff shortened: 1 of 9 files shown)


akva2:
jenkins build this please

opm/simulators/timestepping/TimeStepControl.cpp
```diff
@@ -292,83 +293,140 @@ namespace Opm
...
+                   std::pow(timeSteps[2] / timeSteps[1], -alpha_[0]) *
+                   std::pow(timeSteps[1] / timeSteps[0], -alpha_[1]);
+        }
+        // "I-controller"
+        return std::pow(safetyFactor_ * tolerance_ / errors[2], 0.35);
+    }
+
     bool General3rdOrderController::
-    timeStepAccepted(const double error) const
+    timeStepAccepted(const double error, const double timeStep) const
```
atgeirr, line 374: The change to making the storing and shifting of errors here is a significant change to the API, which makes it required to call timeStepAccepted() every step. Please document in header.

atgeirr:
jenkins build this please

atgeirr:
jenkins build this please

PR #6293, opened 2025-05-19
Allow For Regex Metacharacters in Case Name During File Removal

bska:
This PR pulls the `baseName` out of the file removal regular expression since it's a user-controlled string that might happen to contain regular expression metacharacters like `'+'` or `'-'`.  In rare corner cases, those characters might lead to matching too many or not enough files, thus making `prepareResultOutputDirectory()` too fragile.

This PR therefore switches to using regular string matching for the `baseName`, and only for regular files&ndash;i.e., symbolic links and directories are omitted from consideration.  We also, as an extra special case, exempt grid files from file removal since it's not entirely impossible for `CASE.DATA` to use `CASE.EGRID` as an input file (`GDFILE` keyword).

### DIFF
--- a/opm/simulators/utils/readDeck.cpp
+++ b/opm/simulators/utils/readDeck.cpp
@@ -75,14 +75,17 @@
 
 #include <fmt/format.h>
 
-#include <cstdlib>
+#include <algorithm>
 #include <cstdint>
+#include <cstdlib>
 #include <filesystem>
 #include <memory>
 #include <regex>
 #include <sstream>
 #include <stdexcept>
+#include <system_error>
 #include <utility>
+#include <vector>
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


opm/simulators/utils/readDeck.cpp
```diff
@@ -448,14 +451,68 @@ void Opm::ensureOutputDirExists(const std::string& cmdline_output_dir)
...
-    std::regex r(baseName + R"(\.(F?(DBG|E?GRID|INIT|PRT|RFT|SMSPEC|UNSMRY|UNRST)|([ABCFGHSTUXYZ]\d{4})|(INFOSTEP|INFOITER|OPMRST)))");
-    for (auto& file : std::filesystem::directory_iterator(outputDir)) {
-        std::string fileName = file.path().filename();
-        if (std::regex_match(fileName, r)) {
-            std::filesystem::remove(file);
-        }
-    }
+    namespace fs = std::filesystem;
+
+    // Delete every regeular file in the output directory whose name is
```
akva2, line 456: regular

bska: > regular

Thanks.  I've pushed an update to correct the spelling.  I've also switched to using `std::for_each()` for the directory traversal as well.

bska:
jenkins build this please

akva2 (approved):
Looks good to me now.

bska:
> Looks good to me now.

Much appreciated.  Seeing as the build check is green, I'll merge into master.

PR #6308, opened 2025-05-23
CPR-AMG fully on the GPU with GpuIstl

jakobtorben:
# CPR-AMG fully on the GPU with GpuIstl

This PR introduces a full implementation of the Constrained Pressure Residual (CPR) with Algebraic Multigrid (AMG) preconditioner (CPR-AMG) on the GPU using the GpuIstl framework.

The PR builds on the recent changes in https://github.com/OPM/opm-simulators/pull/6270, which allows creating a linear solver and preconditioner where all the types live on the GPU, and uses the same GPU matrix. This PR makes it possible to run the identical CPR-AMG preconditioner setup on both CPU and GPU (with different hardware specific AMG solvers) by using the option `linear-solver-accelerator=cpu/gpu`. Having the possibilty to run the same configuration on the CPU and GPU is a good flexibility for development, testing, and deployment on different hardware. The CPR-AMG preconditioner is also the default preconditioner on the CPU in OPM, so getting a good GPU implementation here will be essential for good GPU performance.

#### Note
Note that that the PR has been split into many smaller PRs to get all the required functionality merged in step by step. This final PR adds the CPR specific operations and registers the preconditioner in the GPU factory. 

### CPR operations on the GPU
*   With this PR, the linear solver in the GpuIstl framework now have support for core CPR operations such as `moveToCoarseLevel`, `moveToFineLevel`, and `calculateCoarseEntries` from fast, native GPU implementations.
*   CPR weights calculation for `quasiimpes`have a fast, GPU-native implementation.
*   While `trueimpes` and `trueimpesanalytic` weights still needs to be computed on the CPU (due to dependencies on linearization and derivatives not yet ported to the GPU), they can still be used here through copying the result to a `GpuVector` that is used in the GPU linear solver.
* From experiments, using `trueimpes/trueimpesanalytic` which takes longer to run, has still been the best option, due to the significant reduction in linear iterations.

### GPU Resource Management
*   This setup uses a single GPU matrix for the linear solver, preconditioner, and associated matrix operations, minimizing memory use and transfer overhead.
*   This PR adds support for:
    *   A full GPU enabled linear solver for both AMGX and Hypre BoomerAMG
    *   Direct GPU-to-GPU memory transfers when used with a GPU-type linear solver.
    *   While retaining support for using AMGX and Hypre on the GPU even when the rest of the linear solver is on the CPU. In this scenario, input data is uploaded from CPU to GPU, processed by AMG GPU solver, and results are downloaded back to the CPU.

### Leveraging GpuISTL framework for CPR
*   This PR enables the CPR-AMG preconditioner to use the same CPR configuration for CPU and GPU, which means that you can use the same input json or linear solver specification for the linear solver for both CPU and GPU acceleration.
*   The PR enables using the same preconditioner codes for both CPU and GPU (`twoLevelMethod` and `owningTwoLevelPreconditioner`).

### Well Operators
* Well operators are not yet integrated into this specific GPU CPR-AMG setup, which means that it requires `--matrix-add-well-contributions=true`. I have done some prototyping for standard wells earlier in a similar setup, so this should be possible to add. This will enable also supporting CPRW on the GPU, such that we have a GPU version of the current CPU default.

### Testing
*   The CPR implementation has dedicated tests in `test_GpuPressureTransferPolicy`, which tests that the CPR operations and quasiimpes calculation matches the CPU implementation.
* I have tested it on a range of different cases, with convergence typically being as good or better than the current CPU CPR-AMG solver.

### CPU vs. GPU Consistency
* When comparing a CPU run of CPR but with only the AMG offloaded to the GPU (with AMGX), I got almost exactly the same number of iterations as if I did the whole linear solver CPR-AMG (with AMGX) on the GPU, which indicates that the CPR operations are doing the same thing.

### Hardware Support
* The implementation has been tested with both the AMGX and Hypre BoomerAMG implementation, with AMGX being tested on NVIDIA GPUs and Hypre being tested with both NVIDIA and AMD GPUs. The AMD GPU implementation is made possible through the HIP interface used in OPM.

## Benchmark Results:
Here are some quick benchmark results when testing the implementation together with the Jacobi and DILU finesmoother (operating on the fine, blocked system). The following config was used for the GPU:

```json
{
    "maxiter": "20",
    "tol": "0.0050000000000000001",
    "verbosity": "0",
    "solver": "bicgstab",
    "preconditioner": {
        "type": "cpr",
        "weight_type": "trueimpes",
        "pre_smooth": "0/3",
        "post_smooth": "1/3",
        "finesmoother": {
            "type": "dilu/jac"
                },
        "verbosity": "0",
        "coarsesolver": {
            "maxiter": "1",
            "tol": "0.10000000000000001",
            "solver": "loopsolver",
            "verbosity": "0",
            "preconditioner": {
                "type": "amgx/hypre"
            }
        }
    }
}
```

With 1 MPI proccess and 12 threads per process and the options `linear-solver-accelerator=gpu`, `--full-time-step-initially=1` and `--matrix-add-well-contributions=true`.

While for the CPU 12 MPI processes were used with 1 thread per process. The default options for `--linear-solver=cpr_trueimpes` was used, with the option `--full-time-step-initially=1` and `--matrix-add-well-contributions=true`.

All tests has been executed on an AMD Ryzen 9 5900X 12-Core Processor CPU and NVIDIA GeForce RTX 3090 Ti GPU.

### **Norne Case:**

performance_comparison_norne

### **SPE10 Case:**

performance_comparison_spe10

### **Sleipner CO2 case:**

performance_comparison_sleipner

## How to test the new GPU functionality
* The PR sets the default linear solver to `cpr_trueimpes` when specifying the `--linear-solver-accelerator=gpu`. If you have not specified the full linear solver setup with an input json file, it will search for available AMG solvers with GPU backend, selecting AMGX first if available and Hypre if AMGX is not available.
* For further control, an input json file needs to be specified, where you can change things like the finesmoother used, and the AMG solver used (e.g. amgx or hypre). OPM provides an interface to the most important parameters for setting up the AMG solver, these can be specified in the json input. Here are the current defaults for both solvers. To get the most updated parameters inspect the source code or run a simulation with just specifying the AMG solver and inspecting the linear solver specification in the output DBG file.

#### AMGX
```json
            "preconditioner": {
                "type": "amgx",
                "determinism_flag": "0",
                "print_grid_stats": "0",
                "print_solve_stats": "0",
                "solver": "AMG",
                "algorithm": "CLASSICAL",
                "interpolator": "D2",
                "selector": "PMIS",
                "smoother": "BLOCK_JACOBI",
                "presweeps": "3",
                "postsweeps": "3",
                "strength_threshold": "0.5",
                "max_iters": "1",
                "setup_frequency": "30"
            }
```

#### Hypre CPU

```json 
            "preconditioner": {
                "type": "hypre",
                "print_level": "0",
                "max_iter": "1",
                "strong_threshold": "0.5",
                "agg_trunc_factor": "0.3",
                "interp_type": "6",
                "max_levels": "15",
                "tolerance": "0",
                "use_gpu": "false",
                "relax_type": "13",
                "coarsen_type": "10",
                "agg_num_levels": "1",
                "agg_interp_type": "4"
            }
```

#### Hypre GPU

```json 
            "preconditioner": {
                "type": "hypre",
                "print_level": "0",
                "max_iter": "1",
                "strong_threshold": "0.5",
                "agg_trunc_factor": "0.3",
                "interp_type": "6",
                "max_levels": "15",
                "tolerance": "0",
                "use_gpu": "true",
                "relax_type": "16",
                "coarsen_type": "8",
                "agg_num_levels": "0",
                "agg_interp_type": "6"
            }
```


jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this serial rocm hipify please

opm/simulators/linalg/OwningTwoLevelPreconditioner.hpp
```diff
@@ -110,7 +110,8 @@ class OwningTwoLevelPreconditioner : public Dune::PreconditionerWithUpdate<Vecto
                 OPM_THROW(std::ofstream::failure,
                           "Could not write weights to file " + filename + ".");
             }
-            Dune::writeMatrixMarket(weights_, outfile);
+            // Unqualified on purpose to enable ADL for GPU/CPU types
```
multitalentloes, line 113: what is ADL?

opm/simulators/linalg/getQuasiImpesWeights.hpp
```diff
@@ -96,6 +105,47 @@ namespace Amg
...
+                             const int pressureVarIndex,
+                             const bool transpose,
+                             gpuistl::GpuVector<T>& weights,
+                             const gpuistl::GpuVector<int>& diagonalIndices)
+    {
+        gpuistl::detail::getQuasiImpesWeights(matrix, pressureVarIndex, transpose, weights, diagonalIndices);
+    }
+
+    template <typename T>
+    gpuistl::GpuVector<T> getQuasiImpesWeights(const gpuistl::GpuSparseMatrix<T>& matrix,
```
multitalentloes, line 141: Is there good justification for having this function, which calls the one above as far as I can tell, be in a different namespace? Similar functions doing GPU stuff in the same file but only one in `gpustil::`

opm/simulators/linalg/gpuistl/detail/cpr_amg_operations.cu
```diff
@@ -0,0 +1,332 @@
...
+                                           const T* weights,
+                                           const int numberOfBlocks,
+                                           const int blockSize,
+                                           const int pressureVarIndex,
+                                           const bool transpose)
+    {
+        const auto blockIndex = blockDim.x * blockIdx.x + threadIdx.x;
+
+        if (blockIndex < numberOfBlocks) {
+            if (transpose) {
```
multitalentloes: Two things about this function:

1. The transpose bool can also be set as a template argument, the add `constexpr` here to evaluate the if at compile time instead.
2. The naming seems a bit weird to me (maybe standard in AMG terms?). If transpose is false the we extract the coarse values and update component. If transpose is true, then we use some weights to fill in the fine vector with values from the coarse one. Neither case uses a transposed matrix?

opm/simulators/linalg/gpuistl/detail/cpr_amg_operations.cu
```diff
@@ -0,0 +1,332 @@
...
+                                         const int blockSize,
+                                         const int pressureVarIndex,
+                                         const bool transpose)
+    {
+        const auto blockIndex = blockDim.x * blockIdx.x + threadIdx.x;
+
+        if (blockIndex < numberOfBlocks) {
+            T rhsEl = 0.0;
+
+            if (transpose) {
```
multitalentloes: Same comments as the prolongation kernel: transpose argument can be evaluated at compile time, and what exactly is transposed?

opm/simulators/linalg/gpuistl/detail/cpr_amg_operations.cu
```diff
@@ -0,0 +1,332 @@
...
+            const int start = rowIndices[row];
+            const int end = rowIndices[row + 1];
+
+            // Process all non-zeros in this row
+            for (int i = start; i < end; i++) {
+                const int col = colIndices[i];
+                const int blockOffset = i * blockSize * blockSize;
+                T matrixEl = 0.0;
+
+                if (transpose) {
```
multitalentloes: make transponse compile time

opm/simulators/linalg/gpuistl/detail/cpr_amg_operations.cu
```diff
@@ -0,0 +1,332 @@
...
+            // Storage for solution
+            T bweights[3] = {0}; // Maximum block size is 3
+
+            // Solve the system
+            if (transpose) {
+                // Solve using original matrix
+                solveBlock(block, rhs, bweights, blockSize);
+            } else {
+                // Create transposed block for solving
+                T transposed[9]; // Maximum block size squared is 3x3=9
```
multitalentloes: Blocksize is `{1, 2, 3}`, so we could make it a compile time argument and avoid having excessively large arrays here (presumable good for cache).

multitalentloes: That goes for the entire function if it is implemented

opm/simulators/linalg/gpuistl/detail/cpr_amg_operations.cu
```diff
@@ -0,0 +1,332 @@
...
+                                 + ". Only block sizes 1-3 are supported.");
+    }
+
+    // Ensure weights vector has the right size
+    if (weights.dim() != numberOfRows * blockSize) {
+        throw std::runtime_error("Weights vector has incorrect size");
+    }
+
+    // Initialize weights to zero
+    weights = 0.0;
```
multitalentloes: We will overwrite the entire weights vector while not reading from it, I think this operation can be skipped?

opm/simulators/linalg/gpuistl/detail/cpr_amg_operations.cu
```diff
@@ -0,0 +1,332 @@
...
+
+template <typename T>
+void
+restrictVector(const GpuVector<T>& fine,
+               GpuVector<T>& coarse,
+               const GpuVector<T>& weights,
+               std::size_t pressureVarIndex,
+               bool transpose)
+{
+    const int blockSize = fine.dim() / coarse.dim();
```
multitalentloes, line 260: A question revealing my ignorance: Given that this is how `blockSize` is computed, the restrictVector is only used to extract pressure system from blocked system. What function is used to restrict between AMG levels?

opm/simulators/linalg/gpuistl/detail/deviceBlockOperations.hpp
```diff
@@ -281,6 +281,91 @@ mmvMixedGeneral(const MatrixScalar* A, const VectorScalar* b, ResultScalar* c)
         }
     }
 }
+
+// Checks if a value is close to zero based on a precision limit
+template <typename T>
+__device__ __forceinline__ bool
+isCloseToZero(const T value, const T limit = T(1e-40))
```
multitalentloes, line 291: Very strict tolerance, is this particular number from anywhere?

opm/simulators/linalg/gpuistl/detail/deviceBlockOperations.hpp
```diff
@@ -281,6 +281,91 @@ mmvMixedGeneral(const MatrixScalar* A, const VectorScalar* b, ResultScalar* c)
...
+{
+    return abs(value) < limit;
+}
+
+// Solve a linear system Ax=b for block sizes 1-3
+template <typename T>
```
multitalentloes: add blocksize as compile time argument two avoid skipping a few extra ifs in the most common 3x3 case?

opm/simulators/linalg/setupPropertyTree.cpp
```diff
@@ -198,34 +329,28 @@ setupCPR(const std::string& conf, const FlowLinearSolverParameters& p)
     }
     prm.put("preconditioner.pre_smooth", 0);
     prm.put("preconditioner.post_smooth", 1);
-    prm.put("preconditioner.finesmoother.type", "paroverilu0"s);
+    // Choose finesmoother based on accelerator backend
+    if (p.linear_solver_accelerator_ == Parameters::LinearSolverAcceleratorType::GPU) {
+        prm.put("preconditioner.finesmoother.type", "dilu"s);
```
multitalentloes, line 320: If I am interpreting this correct then the GPU defaults to DILU finesmoother and the CPU to ILU0, is this discrepancy intended? If so why?

tests/gpuistl/test_GpuPressureTransferPolicy.cpp
```diff
@@ -0,0 +1,252 @@
...
+            cpuFineResultData[i * blockSize + j] = cpuFineResult[i][j];
+        }
+    }
+
+    std::vector<double> gpuFineResultData = gpuFineResult.asStdVector();
+
+    // Check that prolongation results match
+    BOOST_CHECK_EQUAL(cpuFineResultData.size(), gpuFineResultData.size());
+    for (size_t i = 0; i < cpuFineResultData.size(); ++i) {
+        BOOST_CHECK_CLOSE(cpuFineResultData[i], gpuFineResultData[i], 1e-10);
```
multitalentloes, line 378: I think this test tests very many things at once. Maybe do the preparation of the matrix in a fixture, and then test fewer things at once to make it more apparent where a bug is introduced based on what test fails?

opm/simulators/linalg/setupPropertyTree.cpp
jakobtorben: Yes this is correct. I was unsure whether I should set ILU0 or DILU as the default finesmoother when `--linear-solver--accelerator=gpu`. It was my impression that the DILU implementation had more features and optimisations enabled. You are probably the best judge what is best here. So keen to hear what you think.

opm/simulators/linalg/OwningTwoLevelPreconditioner.hpp
jakobtorben: Argument-dependent lookup see here https://en.wikipedia.org/wiki/Argument-dependent_name_lookup.

This was a solution to make it select the GpuVector implementation when called with a GpuVector type and the standard Dune::writeMatrixMarket for Dune input vectors. I am open to other suggestions that handles this but preferably doesn't add a lot of type specific code into the preconditioner.

opm/simulators/linalg/gpuistl/detail/cpr_amg_operations.cu
jakobtorben: That is correct. Restricting between different AMG levels is handled internally in the AMG solver being used (Hypre/AMGX). The challenge with this function is that the matrix is not available and the GpuVectors are block size unaware, so I had to calculate it from the vectors. One could instead pass in the block size or make GpuVector block size aware, but I preferred this solution.

opm/simulators/linalg/gpuistl/detail/deviceBlockOperations.hpp
jakobtorben: Yes this is to match what is being used when inverting small matrices on the CPU in 

https://github.com/OPM/opm-simulators/blob/fb815768d3bfb73c5dfa3040447b2c3ca14bdb38/opm/simulators/linalg/matrixblock.hh#L188

which I must admit sounds a bit random, and could be changed.

jakobtorben: When I implemented this I tried to keep the CPU and GPU code as close as possible to ensure the same behaviour. I went a bit back and forth between whether to add the safety checks for GPU as well. I ended up adding the functions and the checks but not really doing anything with the information, and never got around to revisit this part. I am considering if I should delete this function and the checks.

opm/simulators/linalg/gpuistl/detail/cpr_amg_operations.cu
jakobtorben: Good point, I will remove this.

opm/simulators/linalg/getQuasiImpesWeights.hpp
jakobtorben: They should all be in the same `Amg` namespace right? Perhaps you misread the version that returns a GpuVector? Having two versions on the GPU is to match the CPU interface.

multitalentloes: You are right, I misread the return type

opm/simulators/linalg/setupPropertyTree.cpp
multitalentloes: As of now it does feature the option to not reorder, which is there to optimize the update-call, more than the apply-call. This is not enabled by default, so making DILU by default like this will probably yield very similar performance to ILU0, I do not think we have tested enough to favor one over the other, so I think we can keep it the same as CPU for now.

opm/simulators/linalg/gpuistl/detail/deviceBlockOperations.hpp
multitalentloes: Ahh so it is used somewhere else as well, that is good to know. Since the priority is to produce similar behavior, I think it is wise to move this value into a header with constants and include it so that we update both at the same time - otherwise the fact that these two should have the same value will be hard to maintain.

opm/simulators/linalg/setupPropertyTree.cpp
jakobtorben: That makes sense to me. In that case we need to set it to `"ilu0"`, since `"paroverilu0"` is not registered on the gpu.

tests/gpuistl/test_GpuPressureTransferPolicy.cpp
jakobtorben: Good suggestion! Implemented separate tests for quasiimpes weight calculation, calculate coarse entries, prolongation and restriction. I  also made the block size a test parameter to test the supported range.

opm/simulators/linalg/setupPropertyTree.cpp
jakobtorben: Ended up setting this to "dilu" as I observed best performance with this preconditioner on some select cases like Norne. We can change it back to "ilu0" once the performance are on par.

opm/simulators/linalg/gpuistl/detail/cpr_amg_operations.cu
jakobtorben: 1. Setting transpose and block size to template parameters is a good suggestion. Note that I did not initially optimise this function too much as I noticed best performance with using the `trueimpes` weight on the CPU, as this would yield significantly fewer iterations. But it is always good to have a optimised GPU implementation and follow the pattern used in other places in GpuIstl. So I added this now.

2. The naming shadows the names used in the CPU implementation to be consistent and make it easy to verify that they are doing the same thing. The transpose actually refers to the `"CPRT"` preconditioner which uses the transposed representation that is needed for things like adjoints. Although this is not currently used, I decided to add it to the GPU implementation to match the CPU implementation.

opm/simulators/linalg/gpuistl/detail/cpr_amg_operations.cu
jakobtorben: Made it compile time. See comment above for explanation.

opm/simulators/linalg/gpuistl/detail/cpr_amg_operations.cu
jakobtorben: Fixed.

opm/simulators/linalg/gpuistl/detail/cpr_amg_operations.cu
jakobtorben: Good suggestion. I made block size a compile time argument now.

opm/simulators/linalg/gpuistl/detail/deviceBlockOperations.hpp
jakobtorben: Good suggestion, added it now.

opm/simulators/linalg/gpuistl/detail/deviceBlockOperations.hpp
jakobtorben: I agree but since this value seem a bit arbitrary I don't want to add an constant into a new file that others might be tempted to use. Maybe it should be changed but don't want to introduce anything that can change CPU behaviour into this PR. So for now I added a comment explaining where this value comes from and a note in the CPU version to update both places if someone changes it.

opm/simulators/linalg/gpuistl/detail/cpr_amg_operations.cu
multitalentloes: Okay justification for the name. The justification is also suitable as a comment around the restriction/prolongation kernels as it can be confusing to read the code without this external context.

opm/simulators/linalg/gpuistl/detail/deviceBlockOperations.hpp
```diff
@@ -281,6 +281,88 @@ mmvMixedGeneral(const MatrixScalar* A, const VectorScalar* b, ResultScalar* c)
...
+    else {
+        // Unsupported block size
+        return false;
+    }
+}
+
+// Transpose a block matrix (row-major)
+template <class T, int blockSize>
+__device__ __forceinline__ void
+transposeBlock(const T* srcBlock, T* dstBlock)
```
atgeirr, line 362: A user error here would be to pass the same pointer twice hoping to do an in-place transpose. Would it be better to only have an in-place transpose? Is it reasonable to add a check here that the pointers do not overlap? Also a test should be added.

opm/simulators/linalg/gpuistl/detail/deviceBlockOperations.hpp
```diff
@@ -281,6 +281,88 @@ mmvMixedGeneral(const MatrixScalar* A, const VectorScalar* b, ResultScalar* c)
...
+__device__ __forceinline__ bool
+isCloseToZero(const T value, const T limit = T(1e-40))
+{
+    return abs(value) < limit;
+}
+
+// Solve a linear system Ax=b for block sizes 1-3
+template <typename T, int blockSize>
+__device__ __forceinline__ bool
+solveBlock(const T* A, const T* b, T* x)
```
atgeirr, line 299: Did you copy the implementations from somewhere else? In any case, tests should be added for this low-level critical and easily testable functionality.

opm/simulators/linalg/gpuistl/GpuVector.hpp
```diff
@@ -515,5 +529,16 @@ class GpuVector
...
 
+} // namespace Opm::gpuistl
+
+// ADL bridge: convert GPU vector to Dune BlockVector and delegate to Dune's writer
+namespace Opm::gpuistl
+{
+template <typename T>
+inline void writeMatrixMarket(const GpuVector<T>& vectorOnDevice, std::ostream& ostr)
+{
+    const auto hostBlockVector = vectorOnDevice.template asDuneBlockVector<1>();
```
atgeirr, line 540: This code looks fine, but I noticed while checking it that the asDuneBlockVector() Doxygen doc claims to create an std::vector, would be good to fix while here.

jakobtorben:

opm/simulators/linalg/twolevelmethodcpr.hh
```diff
@@ -574,6 +580,10 @@ private:
   std::size_t preSteps_;
   /** @brief The number of postsmoothing steps to apply. */
   std::size_t postSteps_;
+  /** @brief Temporary vector for the left-hand side. */
+  FineDomainType u_;
```
atgeirr: I assume this is just a performance optimization (also for the CPU case). Maybe these should be made `mutable`?

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -87,13 +94,27 @@ class ISTLSolverGPUISTL : public AbstractISTLSolver<TypeTag>
...
+        , m_simulator(simulator)
     {
         m_parameters.init(simulator.vanguard().eclState().getSimulationConfig().useCPR());
         m_propertyTree = setupPropertyTree(m_parameters,
                                            Parameters::IsSet<Parameters::LinearSolverMaxIter>(),
                                            Parameters::IsSet<Parameters::LinearSolverReduction>());
 
+        useWellConn_ = Parameters::Get<Parameters::MatrixAddWellContributions>();
+
+        if (!useWellConn_) {
```
atgeirr: No need to have a member for this, unless you plan on using it soon in a follow-up?

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -284,16 +305,103 @@ class ISTLSolverGPUISTL : public AbstractISTLSolver<TypeTag>
...
+                m_cpuWeights.resize(m_matrix->N());
+                m_pinnedWeightsMemory = std::make_unique<PinnedMemoryHolder<real_type>>(
+                    const_cast<real_type*>(&m_cpuWeights[0][0]),
+                    m_cpuWeights.dim()
+                );
+                m_weights = std::make_unique<GPUVector>(m_cpuWeights);
+
+                // CPU implementation wrapped for GPU
+                weightsCalculator = [this]() -> GPUVector& {
+
```
atgeirr: Avoid excessive blanks inside the lambdas.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
```diff
@@ -333,8 +444,14 @@ class ISTLSolverGPUISTL : public AbstractISTLSolver<TypeTag>
     std::unique_ptr<PinnedMemoryHolder<real_type>> m_pinnedMatrixMemory;
     std::unique_ptr<PinnedMemoryHolder<real_type>> m_pinnedRhsMemory;
     std::unique_ptr<PinnedMemoryHolder<real_type>> m_pinnedXMemory;
+    std::unique_ptr<PinnedMemoryHolder<real_type>> m_pinnedWeightsMemory;
```
atgeirr, line 470: Why are all of these unique_ptr? Is it not possible or wanted to default-construct and empty PinnedMemoryHolder for some reason?

For m_weights, m_diagonalIndices and m_element_chunks I can see no obvious reason.

opm/simulators/linalg/gpuistl/GpuPressureTransferPolicy.hpp
```diff
@@ -0,0 +1,149 @@
...
+class GpuPressureTransferPolicy
+    : public Dune::Amg::LevelTransferPolicyCpr<FineOperator, Details::GpuCoarseOperatorType<Scalar, Communication>>
+{
+public:
+    using CoarseOperator = typename Details::GpuCoarseOperatorType<Scalar, Communication>;
+    using ParentType = Dune::Amg::LevelTransferPolicyCpr<FineOperator, CoarseOperator>;
+    using ParallelInformation = Communication;
+
+    using FineVectorType = typename FineOperator::domain_type;
+    using CoarseRangeType = typename CoarseOperator::range_type;
```
atgeirr: The CoarseRangeType and CoarseDomainType are unused.

opm/simulators/linalg/gpuistl/GpuPressureTransferPolicy.hpp
```diff
@@ -0,0 +1,149 @@
...
+    std::size_t getPressureIndex() const
+    {
+        return pressure_var_index_;
+    }
+
+private:
+    Communication* communication_;
+    const FineVectorType& weights_;
+    const std::size_t pressure_var_index_;
+    std::shared_ptr<Communication> coarseLevelCommunication_;
```
atgeirr, line 141: I think unique_ptr is more appropriate here. (Also in the existing PressureTransferPolicy for that matter.)

tests/gpuistl/test_GpuPressureTransferPolicy.cpp
```diff
@@ -0,0 +1,468 @@
...
+        for (auto col = row->begin(); col != row->end(); ++col) {
+            cpuCoarseMatrixData.push_back((*col)[0][0]);
+        }
+    }
+
+    // Get GPU coarse matrix values
+    std::vector<double> gpuCoarseMatrixData = gpuCoarseMatrix.getNonZeroValues().asStdVector();
+
+    // Check that coarse matrix entries match
+    BOOST_CHECK_EQUAL(cpuCoarseMatrixData.size(), gpuCoarseMatrixData.size());
```
atgeirr: BOOST_REQUIRE_EQUAL, since the parts below may crash if not true. Similar on lines 147 and 367.

opm/simulators/linalg/setupPropertyTree.cpp
```diff
@@ -198,34 +329,29 @@ setupCPR(const std::string& conf, const FlowLinearSolverParameters& p)
...
-    prm.put("preconditioner.coarsesolver.preconditioner.accumulate", 1);
-    prm.put("preconditioner.coarsesolver.preconditioner.prolongationdamping", 1.0);
-    prm.put("preconditioner.coarsesolver.preconditioner.maxdistance", 2);
-    prm.put("preconditioner.coarsesolver.preconditioner.maxconnectivity", 15);
-    prm.put("preconditioner.coarsesolver.preconditioner.maxaggsize", 6);
-    prm.put("preconditioner.coarsesolver.preconditioner.minaggsize", 4);
+    // Choose coarsesolver AMG backend based on accelerator backend and available AMG backends
+    if (p.linear_solver_accelerator_ == Parameters::LinearSolverAcceleratorType::GPU) {
+        setGpuAmgBackend(prm,
+                                "preconditioner.coarsesolver.preconditioner.type",
```
atgeirr: Weird formatting.

opm/simulators/linalg/setupPropertyTree.cpp
```diff
@@ -50,7 +180,27 @@ setupPropertyTree(FlowLinearSolverParameters p, // Note: copying the parameters
...
+
+            if (preconditionerType == "cpr" || preconditionerType == "cprt" ||
+                preconditionerType == "cprw" || preconditionerType == "cprwt") {
+                auto coarsesolverType = prm.get("preconditioner.coarsesolver.preconditioner.type", "amg"s);
+                std::transform(coarsesolverType.begin(), coarsesolverType.end(), coarsesolverType.begin(), ::tolower);
+
+                // Choose coarsesolver AMG backend based on accelerator backend and available AMG backends
+                if (p.linear_solver_accelerator_ == Parameters::LinearSolverAcceleratorType::GPU ||
+                    coarsesolverType == "amgx" || coarsesolverType == "hypre") {
+                    setGpuAmgBackend(prm,
```
atgeirr: Weird formatting.

opm/simulators/linalg/setupPropertyTree.cpp
```diff
@@ -50,7 +180,27 @@ setupPropertyTree(FlowLinearSolverParameters p, // Note: copying the parameters
             OPM_THROW(std::invalid_argument, "JSON file " + conf + " does not exist.");
         }
         try {
-            return PropertyTree(conf);
+            auto prm = PropertyTree(conf);
```
atgeirr: I think this section of changes should be removed. If the user actually specifies a json file we should not tamper with it in any way, that is very surprising behaviour to me. Unlike the novice-friendly "cpr" etc. options the json option is intended to give an expert user full control.

It would mean you cannot use the same JSON definition from file with both CPU and GPU runs, I do not think that is a critical loss? One can always use the simulator's user friendly options to generate a starting point for CPU or GPU JSON.

opm/simulators/linalg/setupPropertyTree.cpp
```diff
@@ -239,27 +365,20 @@ setupAMG([[maybe_unused]] const std::string& conf, const FlowLinearSolverParamet
...
     prm.put("solver", getSolverString(p));
-    prm.put("preconditioner.type", "amg"s);
-    prm.put("preconditioner.alpha", 0.333333333333);
-    prm.put("preconditioner.relaxation", 1.0);
+
+    // Choose AMG backend based on accelerator backend and available AMG backends
+    if (p.linear_solver_accelerator_ == Parameters::LinearSolverAcceleratorType::GPU ||
+        prm.get("preconditioner.type", "amg"s) == "amgx"s ||
+        prm.get("preconditioner.type", "amg"s) == "hypre"s) {
+        setGpuAmgBackend(prm, "preconditioner.type", "preconditioner.");
```
atgeirr: All similar functions are named `setupSomething()` not `setSomething()`. The `set` name implies that there is some underlying state that is changed, which is not really the case here.

opm/simulators/linalg/setupPropertyTree.cpp
```diff
@@ -31,6 +31,136 @@
...
+    prm.put(root + "solver", prm.get<std::string>(root + "solver", "AMG"s));
+    prm.put(root + "algorithm", prm.get<std::string>(root + "algorithm", "CLASSICAL"s));
+    prm.put(root + "interpolator", prm.get<std::string>(root + "interpolator", "D2"s));
+    prm.put(root + "selector", prm.get<std::string>(root + "selector", "PMIS"s));
+    prm.put(root + "smoother", prm.get<std::string>(root + "smoother", "BLOCK_JACOBI"s));
+    prm.put(root + "presweeps", prm.get<int>(root + "presweeps", 3));
+    prm.put(root + "postsweeps", prm.get<int>(root + "postsweeps", 3));
+    prm.put(root + "strength_threshold", prm.get<double>(root + "strength_threshold", 0.5));
+    prm.put(root + "max_iters", prm.get<int>(root + "max_iters", 1));
+    prm.put(root + "setup_frequency", prm.get<int>(root + "setup_frequency", Opm::Parameters::Get<Opm::Parameters::CprReuseInterval>()));
```
atgeirr: Too long line.

opm/simulators/linalg/setupPropertyTree.cpp
```diff
@@ -31,6 +31,136 @@
...
+    return;
+#elif HAVE_HYPRE
+#  if HYPRE_USING_CUDA || HYPRE_USING_HIP
+    prm.put(typeKey, "hypre"s);
+    setupHypreAMG(prm, root, true);
+    return;
+#  else
+    OPM_THROW(std::invalid_argument,
+              "GPU accelerator selected, but Hypre is built without GPU support. "
+              "Enable AMGX or build Hypre with CUDA/HIP, or use a different preconditioner (e.g., DILU by setting --linear-solver-configuration=dilu).");
```
atgeirr: Split line here and below. Correct name of option is `--linear-solver`.

atgeirr (commented):
A lot of nice work here, should be merged when issues are addressed.

opm/simulators/linalg/gpuistl/GpuPressureTransferPolicy.hpp
jakobtorben: I think so as well, but I kept it as a `shared_ptr` to maintain a consistent ownership pattern between this class and the other transfer policies. Changing it to a `unique_ptr` would not work with the current `clone()` implementation, as we then would be required to define a copy constructor. Changing the ownership patter should be done simultaneously across all the transfer policies, which is not something I want to introduce in this PR.

opm/simulators/linalg/gpuistl/detail/deviceBlockOperations.hpp
jakobtorben: For the specific use case considered here, adding an inplace version would still require allocating a transposed matrix, copying the elements and then do the in place transpose. Since the `quasiImpesWeightsKernel` is  the only user of this function I would like to keep it as it is. Checking that the pointers overlap is a good idea but it is not possible to assert or something like that in a device function. Some alternatives are returning early which would silently fail or add an error error flag but this would require some more functionality outside this function to handle it. So I think I would prefer to just keep it as it is now. This is more just a helper function, so not something a lot of users would typically use.

I will add a test now.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
jakobtorben: Fixed.

opm/simulators/linalg/gpuistl/GpuVector.hpp
jakobtorben: Fixed.

opm/simulators/linalg/twolevelmethodcpr.hh
jakobtorben: That is correct. There is no need for it to be `mutable` with its current use but there might be in the future, so no harm in making it `mutable`.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
jakobtorben: The member was to match `ISTLSolver` but it is really not needed for anything else so lets keep it simple and remove it.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
jakobtorben: The `PinnedMemoryHolder` does not support default construction. This is intentional as it only makes sense when pinned to some memory, that it can register and unregister.

`m_element chunks` has no need to be a `unique_ptr`. It does not have a default constructor but we can just construct it in the constructor. I changed `m_weights` and `m_diagonalIndices` to optional to make it clearer that these are not always needed.

opm/simulators/linalg/gpuistl/GpuPressureTransferPolicy.hpp
jakobtorben: Good catch, removed them.

tests/gpuistl/test_GpuPressureTransferPolicy.cpp
jakobtorben: Good point, fixed it now.

opm/simulators/linalg/setupPropertyTree.cpp
jakobtorben: Fixed.

opm/simulators/linalg/setupPropertyTree.cpp
jakobtorben: Fixed.

opm/simulators/linalg/setupPropertyTree.cpp
jakobtorben: Good point! Changed it to `setupGpuAmgBackend` now.

opm/simulators/linalg/setupPropertyTree.cpp
jakobtorben: Fixed.

opm/simulators/linalg/setupPropertyTree.cpp
jakobtorben: Fixed.

opm/simulators/linalg/gpuistl/detail/deviceBlockOperations.hpp
jakobtorben: No but it is using the same method as the equivalent Dune function and is rewritten to match the input data and GPU types used here. I added a test in the new test file `test_deviceBlockOperations.cu` which tests the two new functions added here.

opm/simulators/linalg/gpuistl/detail/deviceBlockOperations.hpp
jakobtorben: For the way this function is used here, an inplace version would sill require allocating a block, copying all the elements and then taking the inplace transpose. Since this is the only use of this function I would like to keep it as it is now. Passing the same ptrs to attempt using this as an inplace method would indeed yield the incorrect behaviour. I can add an assert here to check that they are not the same but this will only be active when compiling CUDA in debug mode. I also updated the comment to make it clear that it will not work when `srcBlock == dstBlock`.

A test has been added.

jakobtorben:

opm/simulators/linalg/setupPropertyTree.cpp
jakobtorben: Getting this to work for all the desired behaviour was tricky so I was a bit back and forth between different solutions. I did like the opportunity to use the same json for both CPU and GPU backends when testing. But I do see your point that the input json should leave the user in complete control. So I removed that part now. 

When using novice-friendly options like `cpr_trueimpes` with the `--linear-solver-accelerator=gpu` option, I also added some printing to show what backend was found and selected.

opm/simulators/linalg/gpuistl/ISTLSolverGPUISTL.hpp
atgeirr: I would prefer just empty vectors for unused weights and diagonalIndices, but that is just a matter of taste. The std::optional communicates intent well so your solution is fine.

opm/simulators/linalg/gpuistl/GpuPressureTransferPolicy.hpp
atgeirr: > Changing the ownership patter should be done simultaneously across all the transfer policies, which is not something I want to introduce in this PR.

Ok.

PR #6309, opened 2025-05-28
correct misleading macro name

multitalentloes:
Just noticed this badly named macro I made earlier, adjusted the name so it matches what preconditioner it actually is for...


multitalentloes:
jenkins build this please

blattms:
Thanks a lot. Highly appreciated.

PR #6310, opened 2025-05-29
Optimize GPU dilu/ilu0 update calls

multitalentloes:
Optimize update calls - at the cost of apply times to tailor the implementation more towards CPR-type preconditioners with smaller apply/update ratio.

Progress:

- [X] make reordering optional for DILU
- [ ] make reordering optional for ILU0
- [ ] precompute diagonal indices for DILU update variants
- [ ] precompute diagonal indices for ILU0 update variants
- [ ] run formatter and clean up code

Will add some further explanation of what options and now supported for GPU ilu and dilu after this PR.
Might add some timings to show difference in update and apply times

depends on https://github.com/OPM/opm-grid/pull/877


opm/simulators/linalg/gpuistl/detail/preconditionerKernels/DILUKernels.cu
```diff
@@ -215,6 +271,75 @@ namespace
         }
     }
 
+    template <class T, int blocksize>
+    __global__ void cuComputeDiluDiagonalNoReorder(const T* mat,
```
multitalentloes, line 275: Not sure whether or not to avoid this function duplication of two very similar functions for the reordered and noreorder version... Will probably add some const-expr with templates later to ensure no extra checks in this inner-loop computation

opm/simulators/linalg/gpuistl/detail/preconditionerKernels/DILUKernels.cu
```diff
@@ -215,6 +268,69 @@ namespace
...
+                    if (col == realRowIdx) {
+                        break;
+                    } else if (col < realRowIdx) {
+                        left = mid + 1;
+                    } else {
+                        right = mid - 1;
+                    }
+                }
+
+                const int symOpposite = mid;
```
jakobtorben, line 316: Since we are working with constant sparsity structures in the matrix, shouldn't it also be possible to precompute this symmetric opposite index? Following a similar methodology as this newly introduced `diagIndices`? This would of course require storing an extra array of indicies but we would replace a loop, comparisons, and multiple memory accesses to `colIndices` with a single memory lookup.

opm/simulators/linalg/gpuistl/detail/preconditionerKernels/DILUKernels.hpp
```diff
@@ -162,11 +187,36 @@ void computeDiluDiagonal(T* reorderedMat,
...
+                         size_t* diagIndices,
                          int startIdx,
                          int rowsInLevelSet,
                          T* dInv,
                          int threadBlockSize);
 
+/**
+ * @brief Computes the DILU of a BCSR matrix and stores it in a vector containing the diagonal blocks
+ * @param mat pointer to GPU memory containing nonzerovalues of the sparse matrix. The matrix reordered such
+ * that rows in the same level sets are contiguous
```
jakobtorben: Update comment for `noReorder` function version.

opm/simulators/linalg/gpuistl/detail/preconditionerKernels/ILU_variants_helper_kernels.hpp
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * @brief Computes indices of diagonal elements for reordered GPU preconditioner.
+ *
+ * @tparam T Matrix element type
+ * @param[in] mat The matrix in BCSR format
+ * @param[in] rowIndices Row indices of the BCSR matrix
+ * @param[in] colIndices Column indices of the BCSR matrix
+ * @param[in] indexConversion Index conversion mapping
+ * @param[in] rows Number of rows in the matrix
+ * @param[out] diagIndices Array to store the indices of diagonal elements
```
jakobtorben: Also here, check the comments to make sure they are accurate for the `noReorder` version.

jakobtorben:
Nice addition! I have tested the functionality introduced here with this PR https://github.com/OPM/opm-simulators/pull/6308 for the CPR-AMG preconditioner, where it gave a performance improvement for most cases I tested it on.

opm/simulators/linalg/gpuistl/detail/preconditionerKernels/DILUKernels.cu
multitalentloes: That is a very good observation, you are right! I think this would give a small performance improvement. Probably slightly better than the 2-3% improvement we saw from precomputing the diagaonal indices as this traversal of memory jumps more and probably has more adversarial cache effects.

multitalentloes:
@jakobtorben as you stated supporting the `noreorder` actually gives DILU the advantage of using less memory. I think the time has come to try and re-study how large cases we can fit on different GPUs.

opm/simulators/linalg/gpuistl/detail/preconditionerKernels/DILUKernels.cu
multitalentloes: preliminary tests of this did not show any speedup, might be worth trying to understand later though

opm/simulators/linalg/gpuistl/detail/preconditionerKernels/ILU_variants_helper_kernels.hpp
multitalentloes: done

opm/simulators/linalg/gpuistl/detail/preconditionerKernels/DILUKernels.hpp
multitalentloes: good catch, now updated

PR #6311, opened 2025-05-30
Store wellgroup targets

totto82:
The purpose is 
1. To make the group target consistent for local switching during well solving
2. To avoid the usage of global vectors with name mapping from Wellstate
3. Make it possible to use only singleWellState in the well switching algorithms
4. Avoid recomputing wellgroup targets.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -1288,8 +1288,8 @@ void BlackoilWellModelGeneric<Scalar>::
 updateAndCommunicateGroupData(const int reportStepIdx,
                               const int iterationIdx,
                               const Scalar tol_nupcol,
-                              DeferredLogger& deferred_logger,
-                              bool update_wellgrouptarget)
+                              const bool update_wellgrouptarget,
+                              DeferredLogger& deferred_logger)
 {
     OPM_TIMEFUNCTION();
     const Group& fieldGroup = schedule().getGroup("FIELD", reportStepIdx);
@@ -1475,7 +1475,7 @@ updateAndCommunicateGroupData(const int reportStepIdx,
                     break;
                 }
                 default:
-                    throw("Expected WATER, OIL or GAS as type for injector " + well->name());
+                    assert(false); //programming error
                 }
                 group_target = WellGroupHelpers<Scalar>::getWellGroupTargetInjector(well->name(),
                                             well->wellEcl().groupName(),
--- a/opm/simulators/wells/BlackoilWellModelGeneric.hpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.hpp
@@ -438,8 +438,8 @@ protected:
     void updateAndCommunicateGroupData(const int reportStepIdx,
                                        const int iterationIdx,
                                        const Scalar tol_nupcol,
-                                       DeferredLogger& deferred_logger,
-                                       bool update_wellgrouptarget = true);
+                                       const bool update_wellgrouptarget, // we only want to update the wellgrouptarget after the groups have found their controls
+                                       DeferredLogger& deferred_logger);
 
     void inferLocalShutWells();
 
--- a/opm/simulators/wells/BlackoilWellModel_impl.hpp
+++ b/opm/simulators/wells/BlackoilWellModel_impl.hpp
@@ -370,8 +370,8 @@ namespace Opm {
         const int reportStepIdx = simulator_.episodeIndex();
         this->updateAndCommunicateGroupData(reportStepIdx,
                                             simulator_.model().newtonMethod().numIterations(),
-                                            param_.nupcol_group_rate_tolerance_,
-                                            local_deferredLogger, false);
+                                            param_.nupcol_group_rate_tolerance_, /*update_wellgrouptarget*/ false,
+                                            local_deferredLogger);
 
         this->wellState().updateWellsDefaultALQ(this->schedule(), reportStepIdx, this->summaryState());
         this->wellState().gliftTimeStepInit();
@@ -495,6 +495,7 @@ namespace Opm {
         this->updateAndCommunicateGroupData(reportStepIdx,
                                     simulator_.model().newtonMethod().numIterations(),
                                     param_.nupcol_group_rate_tolerance_,
+                                    /*update_wellgrouptarget*/ true,
                                     local_deferredLogger);
         try {
             // Compute initial well solution for new wells and injectors that change injection type i.e. WAG.
@@ -1250,7 +1251,8 @@ namespace Opm {
         OPM_TIMEFUNCTION();
         const int iterationIdx = simulator_.model().newtonMethod().numIterations();
         const int reportStepIdx = simulator_.episodeIndex();
-        this->updateAndCommunicateGroupData(reportStepIdx, iterationIdx, param_.nupcol_group_rate_tolerance_, local_deferredLogger);
+        this->updateAndCommunicateGroupData(reportStepIdx, iterationIdx, 
+            param_.nupcol_group_rate_tolerance_, /*update_wellgrouptarget*/ true, local_deferredLogger);
         const auto [more_inner_network_update, network_imbalance] =
                 updateNetworks(mandatory_network_balance,
                                local_deferredLogger,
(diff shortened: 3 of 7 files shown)


totto82:

totto82:

totto82:

totto82:

totto82:
I can not find any significant test failures anymore. Most of the remaining test failures are related to changes in time-step sizes and not the well/group switching logic. Test failures are expected as we change how a well target under group controls is computed. Most importantly, the group target with-- check-group-constrains-inner-well-iterations=true (default) is consistent during local iterations. 

For the two most prominent test failures:

[01_wvfpexp_02.pdf](https://ci.opm-project.org/job/opm-simulators-PR-builder/8210/artifact/mpi/build-opm-simulators/failure_report/01_wvfpexp_02.pdf) 
Here, I get the same results if I run with --enable-tuning=true 

[02_gconprod_t1l.pdf](https://ci.opm-project.org/job/opm-simulators-PR-builder/8210/artifact/mpi/build-opm-simulators/failure_report/02_gconprod_t1l.pdf)
Here, I get similar results when I compare the master and this PR locally.

totto82:

totto82:

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -644,6 +644,8 @@ namespace Opm
                                             static_cast<Scalar>(prod_controls.bhp_limit));
                 solveWellWithBhp(simulator, dt, bhp, well_state, deferred_logger);
                 ws.thp = this->getTHPConstraint(summary_state);
+                const auto msg = fmt::format("Well {} try with explicit fractions, re-solving", this->name());
```
atgeirr: A little unclear, suggest "Well {} did not converge, resolving with explicit fractions for VFP calculations."

opm/simulators/wells/BlackoilWellModelGeneric.hpp
```diff
@@ -438,7 +438,8 @@ class BlackoilWellModelGeneric
     void updateAndCommunicateGroupData(const int reportStepIdx,
                                        const int iterationIdx,
                                        const Scalar tol_nupcol,
-                                       DeferredLogger& deferred_logger);
+                                       DeferredLogger& deferred_logger,
+                                       bool update_wellgrouptarget = true);
```
atgeirr: I think it would be better to remove the default argument, and then also make the argument `const bool` and move it before the `deferred_logger`.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1423,6 +1424,77 @@ updateAndCommunicateGroupData(const int reportStepIdx,
...
+                    injectionPhase = Phase::OIL;
+                    break;
+                }
+                case InjectorType::GAS:
+                {
+                    injectionPhase = Phase::GAS;
+                    break;
+                }
+                default:
+                    throw("Expected WATER, OIL or GAS as type for injector " + well->name());
```
atgeirr: This throw is a potential deadlock, since `well_container_generic_` contains local wells, i.e. it will not throw on all processes. Also, we should throw a logic_error with the string as argument instead of the string itself. Consider using the macros we already have for it:
```
OPM_BEGIN_PARALLEL_TRY_CATCH();
OPM_DEFLOG_THROW(std::logic_error, "...", deferred_logger);
OPM_END_PARALLEL_TRY_CATCH("...", comm);
```
However, this may be a bit of overkill for a situation which should not arise at all (it would be a programming error). I think an assert() might be appropriate.

opm/simulators/wells/BlackoilWellModel_impl.hpp
```diff
@@ -371,7 +371,7 @@ namespace Opm {
         this->updateAndCommunicateGroupData(reportStepIdx,
                                             simulator_.model().newtonMethod().numIterations(),
                                             param_.nupcol_group_rate_tolerance_,
-                                            local_deferredLogger);
+                                            local_deferredLogger, false);
```
atgeirr: Please add comments (here and at line 495) explaining the logic here, why is this called here (with false) and then again later (with true)?

opm/simulators/wells/WellGroupControls.cpp
```diff
@@ -124,74 +123,14 @@ getGroupInjectionControl(const Group& group,
...
-    for (std::size_t ii = 0; ii < num_ancestors; ++ii) {
-        if ((ii == 0) || well_.guideRate()->has(chain[ii], injectionPhase)) {
-            // Apply local reductions only at the control level
-            // (top) and for levels where we have a specified
-            // group guide rate.
-            target -= localReduction(chain[ii]);
-        }
-        target *= localFraction(chain[ii+1]);
+    Scalar target_rate = well_state.well(well_.indexOfWell()).group_target;
+    if (target_rate < std::numeric_limits<Scalar>::max()) {
```
atgeirr: I think std::optional would be better than using the max as a signal value.

Alternatively, we could store another value in the state, that we could check. Then we could make sanity checks such as verifying that we have the group target we think we have.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -1604,6 +1763,15 @@ checkGroupConstraintsInj(const std::string& name,
                                                group.has_gpmaint_control(injectionPhase,
                                                                          currentGroupControl),
                                                deferred_logger);
+    if (schedule.hasWell(name)) {  // for wells we already have computed the target
+        Scalar scale = 1.0;
+        Scalar target_rate_available = wellState.well(name).group_target;
```
atgeirr: Is this guaranteed to be set already?

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -1604,6 +1763,15 @@ checkGroupConstraintsInj(const std::string& name,
...
+            const Scalar current_rate_available
```
atgeirr: Weird indent, please add bracers around line 1772.

atgeirr (commented):
I think this should be merged, after addressing issues and questions raised.

totto82:

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -1467,6 +1469,36 @@ checkGroupConstraintsProd(const std::string& name,
...
+                if (currentRateFraction > (guiderateFraction * factor)) {
+                    return std::make_pair(true, guiderateFraction / currentRateFraction);
+                }
+            }
+        }
+    }
+
+    if (schedule.hasWell(name) && wellState.well(name).group_target) { // for wells we already have computed the target
+        const Scalar current_well_rate_available = -tcalc.calcModeRateFromRates(rates); // Switch sign since 'rates' are negative for producers.
+        Scalar group_target_rate_available = *wellState.well(name).group_target;
```
atgeirr: This should be const, also on lines 1538, 1847 and 1886. This used to be const, I assume the change was unintentional.

opm/simulators/wells/BlackoilWellModel_impl.hpp
atgeirr: Even though a comment saying `/*update_wellgrouptarget*/` has been added, there is still no explanation for the logic which is what I would like to have.

totto82:

opm/simulators/wells/BlackoilWellModel_impl.hpp
totto82: I added a line in the hpp file.

atgeirr: A bit minimal, but ok... From the code point of view I think this can be merged when tests are under control.

totto82:

totto82:
jenkins build this opm-tests=1350 please

GitPaean:
Hei, this PR made the regression test WVFPEXP-02 shut a well OP-1 earlier, 

image

Is it addressed?

totto82:
I get the same results with --enable-tuning=true (See comment above)

GitPaean:
> I get the same results with --enable-tuning=true (See comment above)

Thanks for the response.  It is not convincing to assume it is safe because tuning fixes the issue. because it is quite some period earlier. 

Just noticing the difference in the running results and figured out it was due to this PR.  Just reporting here for reference.

PR #6316, opened 2025-06-03
adding enableBrine and enableFoam in the Indices

GitPaean:
at the moment, it is relying on PhaseUsage. It is revealed during the efforts to remove the usage of PhaseUsage.


GitPaean:
I am okay this goes in either itself or it goes with the later PR targeting the removal of the PhaseUsage.

GitPaean:
It does not introduce any regressions, while abusing this PR for reference generation for https://github.com/OPM/opm-tests/pull/1344 .

PR #6317, opened 2025-06-04
Also check operability for wells without thp control

totto82:
In NORNE_ATW2013_1A_MSW and NORNE_ATW2013_1A_STW
there is no THP control, and the operability check does not work. It will instead cut the time step until it closes the wells due to repeated convergence issues. 

This PR adds the new operability check introduced for the controlled wells to wells without the THP controls. 

The issue reported on slow runs on the above-mentioned models is fixed with this PR. See also https://github.com/OPM/opm-simulators/pull/6069, which is no longer needed. 

image

### DIFF
--- a/opm/simulators/wells/MultisegmentWell_impl.hpp
+++ b/opm/simulators/wells/MultisegmentWell_impl.hpp
@@ -569,8 +569,8 @@ namespace Opm
         const double dt = simulator.timeStepSize();
         // solve equations
         bool converged = false;
-        if (this->well_ecl_.isProducer() && this->wellHasTHPConstraints(summary_state)) {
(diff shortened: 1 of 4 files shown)


totto82:
> All seem benign to me except the second
Could you explain? For me, the results you link to (except the watercut) seem very similar.

atgeirr:
> Could you explain? For me, the results you link to (except the watercut) seem very similar.

I misread the plot, thinking the well closed, but that was wrong. I still think it is strange that the watercut is suddenly different though, but as it is at a very low level it is not very significant. I withdraw my objection.

totto82:
>  I withdraw my objection.

Is this an approval? Or do you have other comments?

atgeirr:
> Is this an approval? Or do you have other comments?

Sorry for being dense, yes this is an approval.

atgeirr (approved):

totto82:

PR #6318, opened 2025-06-04
change default minimum newton iterations to 1

totto82:
Time to revisit this again?


totto82:

jakobtorben:
I think it is! When using the `nldd` nonlinear-solver, it is critical for performance to set this parameter to 1. I don't see any good reason why this should be 2 for `newton` either.

totto82:

totto82:
Wow. This time, after all the running through and a brief look at the worst failures, it does not look bad. @atgeirr @GitPaean @vkip, what do you think? Should we try to get this in? We will have to test this on bigger models and also look more carefully at the test failures. For some models, this change gives a substantial speedup.

vkip:
> Wow. This time, all the models are running through. I took a brief look at the worst failures, and it does not look bad. @atgeirr @GitPaean @vkip, what do you think? Should we try to get this in? We will have to test this on bigger models and also look more carefully at the test failures. For some models, this change gives a substantial speedup.

I think we should try. I'll be happy to test on some larger models, but maybe after some of you guys have had a look at the regression test failures?

jakobtorben:
Is this something we want to prioritise to get in before the upcoming release?

totto82:

totto82:
I added a fix for the mixing rate. Should be merged separately, but I added it here for testing.

totto82:

totto82:
> There are some values are negative, maybe we should look into it.

Better now, I think. 

PR #7415 still needs some work, but my hope is that with PR #7415 in, we can finally merge this one! 


image

totto82:

totto82:
Now we understand the issue related to negative "F" rates. See #7415.  I think it is safe to merge this as is, now independent of #7415.  @vkip do you want to do some tests?

vkip:
> @vkip do you want to do some tests?

Yes, I'll do a few and report back tomorrow.

vkip:
> > @vkip do you want to do some tests?
> 
> Yes, I'll do a few and report back tomorrow.

Looks good, I support merging this.

totto82:
> Looks good, I support merging this.

Thanks for testing. I will start the process of getting this in. The failures in the restart/parallel tests need some care, not just a ref data update.

PR #6319, opened 2025-06-04
open wells at well testing

totto82:
For testing. See https://github.com/OPM/opm-simulators/pull/6051/files

### DIFF
--- a/opm/simulators/wells/WellInterface_impl.hpp
+++ b/opm/simulators/wells/WellInterface_impl.hpp
@@ -402,6 +402,8 @@ namespace Opm
         else {
             ws.production_cmode = Well::ProducerCMode::BHP;
         }
+        // We will try to open the well
+        ws.open();
 
         updateWellStateWithTarget(simulator, group_state, well_state_copy, deferred_logger);
         calculateExplicitQuantities(simulator, well_state_copy, deferred_logger);
@@ -488,8 +490,6 @@ namespace Opm
                 if (!welltest_state_temp.completion_is_closed(this->name(), completion.first))
                     well_test_state.open_completion(this->name(), completion.first);
             }
-            // set the status of the well_state to open
-            ws.open();
             well_state = well_state_copy;
             open_times.try_emplace(this->name(), well_test_state.lastTestTime(this->name()));
         }


totto82:

GitPaean:
I just check the first regression failure, interestingly the different behavoir is because we get different solutions and during the `wellTesting` function. 

The case is set up with `WECON` with max GOR at 4500. 

The first two well testing results with the **master** branch looks like the following, 
```
443 Starting time step 4, stepsize 5 days, at day 324/549, date = 21-May-2019
444  well surface rates for well OP_1 are: -0 -2357.39 -1.37455e+07 GOR 5830.8 well bhp 1e+07 well thp 0 well control mode BHP
445  well OP_1 is being tested
```
```
463 Starting time step 10, stepsize 5 days, at day 354/549, date = 20-Jun-2019
464  well surface rates for well OP_1 are: -0 -5564.36 -1.67932e+07 GOR 3017.99 well bhp 1e+07 well thp 0 well control mode BHP
465  well OP_1 is being tested
466 well OP_1 is re-opened
```

with this **PR**
```
463 Starting time step 10, stepsize 5 days, at day 354/549, date = 20-Jun-2019
464  well surface rates for well OP_1 are: -0 -692.507 -4e+06 GOR 5776.12 well bhp 1.10832e+07 well thp 0 well control mode GRAT
465  well OP_1 is being tested
```
```
483 Starting time step 16, stepsize 5 days, at day 384/549, date = 20-Jul-2019
484  well surface rates for well OP_1 are: -0 -1200 -3.18903e+06 GOR 2657.52 well bhp 1.16576e+07 well thp 0 well control mode ORAT
485  well OP_1 is being tested
486 well OP_1 is re-opened
```

Both times the well testing returns different rates and different control mode. The first well testing did not impact the re-opening results, while the second one results in different re-opening results. 

Basically, this PR should not change the results we obtain from the well testing, and also even with different control mode, which means the well testing goes through different function calls. 
Furthermore, the different GOR is also a concern. Even the well is under different control mode, the ratios between the rates should be similar.

GitPaean:
For sure the following line in `iterateWellEqWithSwitching()` is potentially the main cause of the difference, 

```c++
        const bool allow_open =  this->well_ecl_.getStatus() == WellStatus::OPEN &&
                                 well_state.well(this->index_of_well_).status == WellStatus::OPEN;
```

maybe @steink can comment on this.

totto82:
@GitPaean I was looking at the same code. I think it is better to allow the well to switch controls during well testing. i.e., we should set ws.open() before we do the well testing not after as in master.

totto82:
@GitPaean What do you think? Should we get this in?

GitPaean:
> > @GitPaean What do you think? Should we get this in?
> 
> I would say this PR reveals some potential design flaws of the well testing and iterating well with switching. We can not update the reference without fixing those.
> 
> The bottom line is that this PR should not change the well testing results, or and it should potentially improves the results.

I meant the PR itself is potentially good while some investigation needs to go in wellTesting and iterateWellEquations to cope the change in this PR.

totto82:
> I meant the PR itself is potentially good while some investigation needs to go in wellTesting and iterateWellEquations to cope the change in this PR.

I am not sure I follow you here. IMO, this fixes a bug in the wellTesting, and we should expect it to affect the results.

GitPaean:
> I am not sure I follow you here. IMO, this fixes a bug in the wellTesting, 

Partly yes, while not sure other parts of the `well testing` and `iterateWellEquations` in the master branch  cope well with the change (potentially fix) in this PR. 

At least the regression failures needs to investigated in my opinion.

GitPaean:
> Master
> 
> ```
> 463 Starting time step 10, stepsize 5 days, at day 354/549, date = 20-Jun-2019
> 464  well surface rates for well OP_1 are: -0 -5564.36 -1.67932e+07 GOR 3017.99 well bhp 1e+07 well thp 0 well control mode BHP
> 465  well OP_1 is being tested
> 466 well OP_1 is re-opened
> ```
> 
> If you look at grat in master you see it is above the limit (4e6) and since it allows for grat > grat_limit it also gives smaller GOR.
> 
> With this PR
> 
> ```
> 463 Starting time step 10, stepsize 5 days, at day 354/549, date = 20-Jun-2019
> 464  well surface rates for well OP_1 are: -0 -692.507 -4e+06 GOR 5776.12 well bhp 1.10832e+07 well thp 0 well control mode GRAT
> 465  well OP_1 is being tested
> ```
> 
> the grat is less or equal to the limit at the GOR becomes too large to open it at this point.
> 
> In this case, solving the coupled wells+reservoir equation leads to a solution that honors the grat limit anyway for master, but I think it makes more sense not to open the well at this point based on a solution that violates the grat_limit. Note also that if we simulate without tuning, both master and PR agree on when to open the well. I haven't tested this enough to conclude on this but I am afraid that allowing wells to open that violate some constraints during testing could lead to wells that open and shut on the same iterations.


This is a WECON setup, I think we should consider the well constraints (GRAT) (as you mentioned here) when determining whether to open the well. In the result below, the well manages to re-open and run some period before it gets SHUT again. I take this as the well can be re-opened.   But the well testing function itself and also the iterate with switching function is so complicated, I did not managed to understand all the implications that this PR can introduce to the well testing function and why the GOR can be so different. 

image

GitPaean:
This PR is in the way of getting https://github.com/OPM/opm-simulators/pull/6051 in. I can check the details of the well testing procedure to evaluate the issue. The well can be opened and run for some period, ideally, well testing should be able to re-open it. We need to check whether there is something can be done to improve this part.

totto82:
> I can check the details of the well testing procedure to evaluate the issue 

That is great. I would really like to get #6051 in asap.

GitPaean:

GitPaean:
I am done evaluating the well testing regressions, will check the too cases regarding WCYCLE , which I am not very familiar.  Will provide my final comment soon.

GitPaean:
I have just checked the regression failure of case WCYCLE-06. At the end, it becomes a well testing issue, more specially, it is well testing issue with injector `OP1`.  I have not looked into the well testing for injectors that much and we have not tested it much either. 

Basically, the well testing is not written/designed for injectors much.  With the master branch, the well `OP` enters the well testing procedure with GAS RATE injection control, and it solves with this control and converged with this control while BHP is above the BHP limit. It is this way is because that the control is not allowed to switch, since it is not allowed to switch, is not checked. 

For producers, there is something special (not sure correct or necessary) to the set the well under BHP or THP control first, while the same did not do for injectors. 

There are many things can be improved for the well testing and the function `iterateWellEquations` is rather complicated, which the well testing procedure relies a lot on it. 

Regarding this regression, I now think it is okay to have this PR in, while some more refactoring and improvement regarding well testing can be done in this area.

GitPaean:

GitPaean:
The downstream reference has been updated. I am getting the PR in now.

PR #6320, opened 2025-06-04
Cleaned up and fixed verbosity for time step controllers

erikhide:
The verbosity argument for the time step controllers wasn't used. I've cleaned up the code a bit and improved the feedback when verbosity is set to true. I also removed some default arguments, as well as comments about default values.

Question: Should `time-step-verbosity` be set to false by default, or should we keep it set to true?

### DIFF
--- a/opm/simulators/timestepping/AdaptiveTimeStepping.cpp
+++ b/opm/simulators/timestepping/AdaptiveTimeStepping.cpp
@@ -160,38 +160,42 @@ void registerAdaptiveParameters()
 std::tuple<TimeStepControlType, std::unique_ptr<TimeStepControlInterface>, bool>
 createController(const UnitSystem& unitSystem)
 {
-    const double tol =  Parameters::Get<Parameters::TimeStepControlTolerance>(); // 1e-1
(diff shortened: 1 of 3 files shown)


blattms:
jenkins build this please

blattms:
I think the parameter's default should be set such that the output for default options stays the same. My feeling is that this would be false, but you are the expert here.

erikhide:
> I think the parameter's default should be set such that the output for default options stays the same. My feeling is that this would be false, but you are the expert here.

I've tested it now, and it turns out that setting this to false would remove some info from the DBG file, as well as the output written to terminal. In particular, lines like
`Starting time step 0, stepsize 1 days, at day 0/1, date = 01-Jan-2015`
will not be printed, which seems like a bad idea for default behaviour.

Keeping the parameter set to true will, for the default time stepping, add the lines
```
Suggested next time step size (from PID controller): 6.345541066667916 days
Suggested next time step size (from iteration target): 3.400000000000001 days
```
both to the DBG file and the terminal output.

If we want to keep current behaviour, we would need two different time step verbosity parameters.

PR #6321, opened 2025-06-04
removing the unused and mistakenly defined function

GitPaean:
assignPrimaryVars from blackoilenergymodules.

It looks that the function is intended to assign a value, while the definition is wrong and not used. 

We can always add a new and correct one when needed.


GitPaean:
jenkins build this please

PR #5818, opened 2024-12-19
gas water for multisegment well

GitPaean:
just started, nothing essential got done yet.

PR #6322, opened 2025-06-05
reusing primary variable indices and flags defined

GitPaean:
MultisegmentWellAssemble reuses the indices and flags defined in MultisegmentWellPrimaryVariable.

It triggered problems when developing https://github.com/OPM/opm-simulators/pull/5818 , while did not think to reuse the indices and flags back then.

### DIFF
--- a/opm/simulators/wells/MultisegmentWellAssemble.hpp
+++ b/opm/simulators/wells/MultisegmentWellAssemble.hpp
@@ -42,18 +42,18 @@ template<class Scalar> class WellState;
 template<class FluidSystem, class Indices>
 class MultisegmentWellAssemble
 {
-    static constexpr int WQTotal = 0;
-    static constexpr bool has_wfrac_variable = Indices::waterEnabled && Indices::oilEnabled;
-    static constexpr bool has_gfrac_variable = Indices::gasEnabled && Indices::numPhases > 1;
-    static constexpr int WFrac = has_wfrac_variable ? 1 : -1000;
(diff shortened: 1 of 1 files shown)


opm/simulators/wells/MultisegmentWellAssemble.hpp
```diff
@@ -42,12 +42,13 @@ template<class Scalar> class WellState;
 template<class FluidSystem, class Indices>
 class MultisegmentWellAssemble
 {
-    static constexpr int WQTotal = 0;
-    static constexpr bool has_wfrac_variable = Indices::waterEnabled && Indices::oilEnabled;
-    static constexpr bool has_gfrac_variable = Indices::gasEnabled && Indices::numPhases > 1;
-    static constexpr int WFrac = has_wfrac_variable ? 1 : -1000;
-    static constexpr int GFrac = has_gfrac_variable ? has_wfrac_variable + 1 : -1000;
-    static constexpr int SPres = has_wfrac_variable + has_gfrac_variable + 1;
+    using PrimarVariables = MultisegmentWellPrimaryVariables<FluidSystem, Indices>;
```
akva2: Unless there is a reason this should probably be PrimaryVariables

GitPaean: good catch! updating it now.

akva2 (approved):
Changes the type alias to be private but it is apparently fine

GitPaean:
> Changes the type alias to be private but it is apparently fine

Thanks for the approval. We can always make it public when needed.

PR #6323, opened 2025-06-05
Give a distributed well the same ID on different ranks

michal-toth:
A distributed well was given different IDs in the coarse solver. As a result, the coarse system had a separate row for each part on a new rank. This PR gives the well an ID which is derived from its position in the `parallel_well_info_` which is consistent across ranks.

The change has no effect on the result of the simulation, the calculation of the residual is untouched - this bug did not affect it. However, the coarse level solver will no longer treat a distributed well as a collection of smaller wells (disconnecting parts on different ranks). We hope to get a performance boost.

Relevant edge case: When a well has no active perforations on one rank, we regularize the local coarse matrix by putting 1 on the diagonal (row is zero elsewhere). Accumulating across ranks thus can put a wrong weight on the diagonal of the coarse system's matrix. The well pressure correction coming from the coarse solve is not used, so I do not know how big issue it will be.

### DIFF
--- a/opm/simulators/linalg/PressureBhpTransferPolicy.hpp
+++ b/opm/simulators/linalg/PressureBhpTransferPolicy.hpp
@@ -33,10 +33,21 @@
 
 namespace Opm
 {
+    /** \brief Add wells to the communicator
+     *
+     * @param comm:       Communicator to be extended
+     * @param commRw:     Created communicator. Communicates whatever comm does plus wells.
+     * @param localWells: Stores for each well on this rank the well ID and ownership status.
+     *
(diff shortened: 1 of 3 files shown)


michal-toth:

michal-toth:
The change must address the diagonal term before it can be added to the master. The diagonal is calculated from the fine matrix D (diagonal part for the well subsystem) by adding terms corresponding to perforations and weighing them by the average weight. The weight calculation is incorrect, each rank uses its own average.
If the weights are similar, no harm is done. However, if one rank has only inactive perforations we add 1 on the diagonal to regularize the matrix. This does not mix well with properly calculated D.

PR #6327, opened 2025-06-06
testing network implementation

GitPaean:


GitPaean:
jenkins build this please

GitPaean:
It is interesting that commenting out the `if` condition does not affect any running results, mostly for the few network cases.

PR #6330, opened 2025-06-11
(Partial) Support for parallel runs with CpGrid with LGRs

aritorto:
In OPM/opm-grid, distributing the level zero grid of a CpGrid with LGRs is supported. This PR allows adding LGRs in the simulation grid. 

The strategy used is to add (the same) LGRs in both global and distributed view, and synchronize the cell ids.

Main changes:
1. Replace cartesianIndexMapper_ for levelCartesianIndexMapper_  when distributing field properties. This allows access to the Cartesian indices of the distributed level zero grid, where the field properties are given - for now.
2.  Skip - for now - in parallel runs for CpGrid with LGRs building index maps in CollectDataInIORank. 
3.  Resize 'region' since, for CpGrid with LGRs, where level zero grid has been distributed, the total amount of elements - per process - in level zero grid and leaf grid do not coincide, in general.

To run a case, use the flags --parsing-strictness=low --enable-ecl-output=false --enable-vtk-output=true.

### DIFF
--- a/opm/simulators/flow/CpGridVanguard.hpp
+++ b/opm/simulators/flow/CpGridVanguard.hpp
@@ -272,12 +272,11 @@ public:
             this->updateGridView_();
             this->updateCellDepths_();
             this->updateCellThickness_();
-        }
-        if (this->grid_->comm().size()>1) {
-            // Add LGRs and update the leaf grid view in the global (undistributed) simulation grid.
-            // Purpose: To enable synchronization of cell ids in 'serial mode',
-            //          we rely on the "parent-to-children" cell id mapping.
-            if (const auto& lgrs = this->eclState().getLgrs(); lgrs.size() > 0) {
+
+            if (this->grid_->comm().size()>1) {
+                // Add LGRs and update the leaf grid view in the global (undistributed) simulation grid.
+                // Purpose: To enable synchronization of cell ids in 'serial mode',
+                //          we rely on the "parent-to-children" cell id mapping.
                 OpmLog::info("\nAdding LGRs to the global view and updating its leaf grid view");
                 this->grid_->switchToGlobalView();
                 this->addLgrsUpdateLeafView(lgrs, lgrs.size(), *this->grid_);
--- a/opm/simulators/flow/CpGridVanguard.hpp
+++ b/opm/simulators/flow/CpGridVanguard.hpp
@@ -263,10 +263,9 @@ public:
      */
     void addLgrs()
     {
-        // Check if input file contains Lgrs.
-        //
-        // If there are lgrs, create the grid with them, and update the leaf grid view.
-        if (const auto& lgrs = this->eclState().getLgrs(); (lgrs.size() > 0) && (this->grid_->comm().size() == 1)) {
+        // Check if input file contains Lgrs. Add them, if any.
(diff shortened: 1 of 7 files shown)


aritorto:

opm/simulators/flow/CpGridVanguard.hpp
```diff
@@ -263,19 +263,28 @@ class CpGridVanguard : public FlowBaseVanguard<TypeTag>
...
+            // Purpose: To enable synchronization of cell ids in 'serial mode',
+            //          we rely on the "parent-to-children" cell id mapping.
+            this->grid_->switchToGlobalView();
+            if (const auto& lgrs = this->eclState().getLgrs(); lgrs.size() > 0) {
+                OpmLog::info("\nAdding LGRs to the global view and updating its leaf grid view");
+                this->addLgrsUpdateLeafView(lgrs, lgrs.size(), *this->grid_);
+            }
+            this->grid_->switchToDistributedView();
+
+            this->grid_->syncDistributedGlobalCellIds();
```
blattms: Shouldn't the siwtching and syncing be inside the if?
```suggestion
            if (const auto& lgrs = this->eclState().getLgrs(); lgrs.size() > 0) {
                OpmLog::info("\nAdding LGRs to the global view and updating its leaf grid view");
                this->grid_->switchToGlobalView();
                this->addLgrsUpdateLeafView(lgrs, lgrs.size(), *this->grid_);
                this->grid_->switchToDistributedView();
                this->grid_->syncDistributedGlobalCellIds();
            }
```

opm/simulators/utils/ParallelEclipseState.hpp
```diff
@@ -102,9 +102,8 @@ class ParallelFieldPropsManager : public FieldPropsManager {
     template<class T>
     void resetCartesianMapper(const T* mapper)
     {
-        m_activeSize = std::bind(&T::compressedSize, mapper);
-        m_local2Global = std::bind(&T::cartesianIndex, mapper,
-                                   std::placeholders::_1);
+        m_activeSize = [mapper]() { return mapper->compressedSize(0); };
```
blattms: We should add a comment why we do this.
Maybe also in ParallelEclipseState

opm/simulators/flow/GenericCpGridVanguard.hpp
```diff
@@ -233,6 +233,8 @@ class GenericCpGridVanguard {
     std::unique_ptr<Dune::CpGrid> equilGrid_;
     std::unique_ptr<CartesianIndexMapper> cartesianIndexMapper_;
     std::unique_ptr<CartesianIndexMapper> equilCartesianIndexMapper_;
+    std::unique_ptr<LevelCartesianIndexMapper> levelCartesianIndexMapper_;
+    std::unique_ptr<LevelCartesianIndexMapper> equilLevelCartesianIndexMapper_;
```
blattms: Is this really used?

aritorto: the one for the simulation grid grid_ is definitely used (in distributeFieldProps_()), replacing the cartesianIndexMapper_ 

the one for equilGrid_, it's not used actually. I added it and initialized it (together with the one for the grid_) in doCreateGrids_()

it's removed now!

PR #6332, opened 2025-06-17
[BUG] TSMINZ instead of TSFMIN

daavid00:
`TSMINZ` instead of `TSFMIN`

### DIFF
--- a/opm/simulators/timestepping/AdaptiveTimeStepping_impl.hpp
+++ b/opm/simulators/timestepping/AdaptiveTimeStepping_impl.hpp
@@ -111,7 +111,7 @@ AdaptiveTimeStepping(double max_next_tstep,
     , growth_factor_{tuning.TFDIFF}
     , max_growth_{tuning.TSFMAX}
     , max_time_step_{tuning.TSMAXZ} // 365.25
-    , min_time_step_{tuning.TSFMIN} // 0.1;
+    , min_time_step_{tuning.TSMINZ} // 0.1;
     , ignore_convergence_failure_{true}
     , solver_restart_max_{Parameters::Get<Parameters::SolverMaxRestarts>()} // 10
     , solver_verbose_{Parameters::Get<Parameters::SolverVerbosity>() > 0 && terminal_output} // 2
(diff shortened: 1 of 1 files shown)


daavid00:
jenkins build this please

daavid00:
jenkins build this opm-tests=1351 please

daavid00:
While this PR also varnishes the output terminal message, this should be done also for other variables, but in a different PR. Then I will update the reference solution (the fails are expected since the values in INIT for the time step are changed in this PR).

daavid00:
jenkins build this opm-tests=1351 update_data please

daavid00:
jenkins build this opm-tests=1351 please

opm/simulators/timestepping/AdaptiveTimeStepping_impl.hpp
```diff
@@ -111,7 +111,7 @@ AdaptiveTimeStepping(double max_next_tstep,
     , growth_factor_{tuning.TFDIFF}
     , max_growth_{tuning.TSFMAX}
     , max_time_step_{tuning.TSMAXZ} // 365.25
-    , min_time_step_{tuning.TSFMIN} // 0.1;
+    , min_time_step_{tuning.TSMINZ} // 0.1;
```
bska, line 114: Good catch.  We've had this problem since #3519.

opm/simulators/timestepping/AdaptiveTimeStepping_impl.hpp
```diff
@@ -58,7 +58,7 @@ namespace Opm {
 /*********************************************
  * Public methods of AdaptiveTimeStepping
  * ******************************************/
-
+using Meas = UnitSystem::measure;
```
bska: I think I'd prefer if this type alias were restricted only to the scope in which it's used, i.e., in member function `checkTimeStepMinLimit_()`.  When placed here, it's essentially available to all clients of the Opm namespace and I think that's needlessly generous.

opm/simulators/timestepping/AdaptiveTimeStepping_impl.hpp
```diff
@@ -907,12 +907,22 @@ checkTimeStepMinLimit_(const double new_time_step) const
 {
     // If we have restarted (i.e. cut the timestep) too
     // much, we have failed and throw an exception.
+    const UnitSystem& unit_system = solver_().model().simulator().vanguard().eclState().getDeckUnitSystem();
```
bska: Very strictly speaking, you don't need this object unless `next_time_step < minTimeStep_()` and the run enables tuning.  Moreover, since getting it requires a little bit of pointer chasing, and since this function is called moderately often, I think it would be better to not access the `UnitSystem` object unless you actually need it.

daavid00:
> Thanks. This does indeed fix a problem we've had for a long time. Just a couple of minor details, the most important of which is to not expose the `Meas` type alias to all users of the `Opm` namespace.

Thanks for the review, I have addressed your comments. Let me know if this needs further changes; otherwise, please run Jenkins.

bska (approved):
Thanks a lot for the updates.  This looks good to me now.

The regression failure we got is a little disconcerting. PR #6293 was supposed to address that issue, so maybe there's something deeper going on.  In any case, it's not caused by this work and since the new `WGRUPCON-*` reference solutions have been installed on the CI system, I'll merge into master.

PR #6333, opened 2025-06-17
using  group_limit_action.oil instead of allRates for better code readability and potentially better flexibility

GitPaean:
it depends on PR https://github.com/OPM/opm-common/pull/4609

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelConstraints.cpp
+++ b/opm/simulators/wells/BlackoilWellModelConstraints.cpp
@@ -482,7 +482,7 @@ actionOnBrokenConstraints(const Group& group,
 
     const auto action = [&group_limit_action](Group::ProductionCMode control) {
         switch (control) {
-            case Group::ProductionCMode::ORAT: return group_limit_action.allRates;
+            case Group::ProductionCMode::ORAT: return group_limit_action.oil;
             case Group::ProductionCMode::WRAT: return group_limit_action.water;
             case Group::ProductionCMode::GRAT: return group_limit_action.gas;
             case Group::ProductionCMode::LRAT: return group_limit_action.liquid;


GitPaean:
> Fair enough, but it's very likely that we'll have to reintroduce something like allRates in the future for simulation restart purposes. If so, I'll try to encapsulate the logic in a better way.

`allRates` is one of the items (only item for now) in the RESTART file, we should cope with the evolving understanding, and it can be handled by [GroupLimitAction](https://github.com/OPM/opm-common/blob/099d79e17f238dd0745b1e969fce1f26996ebe9e/opm/input/eclipse/Schedule/Group/Group.hpp#L175) when restarting as you hinted in a different PR. 

And also, with the fact that `allRates` is the only item written and read in the RESTART file, it is very likely there is some cases that we can not restart properly.

GitPaean:
> And that's why we'll probably have to reintroduce allRates. The restart file cares a lot about whether an item is explicitly defined or just implicitly derived from a default setting and we need to be able to distinguish those cases.

We do still keep `allRates`, while use the derived values in the simulation.  I do not see that we have disagreements here.  But things can change based on the evolving understanding.

GitPaean:
> > > And that's why we'll probably have to reintroduce allRates. The restart file cares a lot about whether an item is explicitly defined or just implicitly derived from a default setting and we need to be able to distinguish those cases.
> > 
> > 
> > We do still keep `allRates`, while use the derived values in the simulation. I do not see that we have disagreements here.
> 
> The disagreement is that the restart file needs to be able to distinguish between the cases of
> 
> ```
> GCONPROD G ORAT 123 456 789 101 RATE /
> ```
> 
> and the (unlikely)
> 
> ```
> GCONPROD G ORAT 123 456 789 101 RATE YES 123.4 FORM RATE /
> ```
> 
> in which the overriding procedure for water (item 11) is explicitly set equal to the procedure for all rates (item 7). The current approach does not distinguish between these cases.

I understand this. And my best understanding is that we do not have item 11 written in or read from the RESTART file yet.  If item 11 is something like `WELL`, restarting will have this information, right?

GitPaean:
> That's true, but we're in the process of bridging that gap.

Thanks for clarification. With new development in the RESTART support and new/current targeting cases, group limit action related is an evolving process and there are open PRs and new PRs going in.

PR #6334, opened 2025-06-18
Don't Copy Well Objects When Well Names are Sufficient

bska:
In particular, favour calling member functions on `ScheduleState::wells` to retrieve references to individual well objects given a name instead of using `Schedule::getWells()` or `Schedule::getWellsAtEnd()`.  While convenient, these functions introduce needless copies of all well objects stored in the Schedule object.  This commit instead, where possible, uses the somewhat less convenient `ScheduleState::well_order()` accessor and then accesses individual wells by name.

To this end, change `BlackoilWellModelGeneric::not_on_process_` from a function taking a well and returning a bool into a function taking a well name and returning a bool.


bska:
> Looks good to me.

Thanks.  In that case, I think I'll mark it as "ready for review" given that I did not get any statistically significant performance measurements from this change.

> looking forward to std::erase_if...

Me too.

bska:

PR #6336, opened 2025-06-18
Making possible to simulate 100 Million years

daavid00:
With master, it is possible to simulate up to the date `31-Dec-9999`.
This PR makes possible to simulate for longer time (tested for 100 Million years); however, we keep in mind that 1 billion years results in a different error:

> Error: An error occurred while creating the reservoir schedule
> Internal error:  Report step 1 has start time after end time,
>    * Start time = 01-Jan-2025 00:00:00
>    * End time   = 17-May--292275000 16:47:05.
>  Possibly due to inconsistent RESTART/SKIPREST settings.
> 
> Error: Unrecoverable errors while loading input:
>  Report step 1 has start time after end time,
>    * Start time = 01-Jan-2025 00:00:00
>    * End time   = 17-May--292275000 16:47:05.
>  Possibly due to inconsistent RESTART/SKIPREST settings.


opm/simulators/flow/SimulatorReportBanners.cpp
```diff
@@ -38,11 +38,20 @@ void outputReportStep(const SimulatorTimer& timer)
...
-            << " at day " << unit::convert::to(timer.simulationTimeElapsed(), unit::day)
-            << "/" << unit::convert::to(timer.totalTime(), unit::day)
-            << ", date = " << timer.currentDateTime();
+    if ((timer.currentDateTime() <= boost::posix_time::ptime(boost::posix_time::max_date_time))
+         && (boost::posix_time::ptime(boost::posix_time::min_date_time) <= timer.currentDateTime())) {
+        stepMsg << "\nReport step " << std::setw(2) << timer.currentStepNum()
+                << "/" << timer.numSteps()
+                << " at day " << unit::convert::to(timer.simulationTimeElapsed(), unit::day)
+                << "/" << unit::convert::to(timer.totalTime(), unit::day)
+                << ", date = " << timer.currentDateTime();
```
akva2: please just make what's changed conditional and not the entire message so we don't have to deal with double changes later (same below)

daavid00: Thanks, addressed.

daavid00:
jenkins build this please

daavid00:
jenkins build this please

daavid00:
jenkins build this please

daavid00:
@akva2 could this be merged or needs further corrections?

opm/simulators/timestepping/SimulatorTimer.cpp
```diff
@@ -76,7 +76,10 @@ namespace Opm
         }
 
         setCurrentStepNum(report_step);
-        start_date_ = boost::posix_time::from_time_t(schedule.getStartTime()).date();
+        if ((boost::posix_time::from_time_t(schedule.getStartTime()) <= boost::posix_time::ptime(boost::posix_time::max_date_time)) &&
+            (boost::posix_time::ptime(boost::posix_time::min_date_time) <= boost::posix_time::from_time_t(schedule.getStartTime()))) {
```
bska: I appreciate the sentiment, but this doesn't do anything useful.  If the start time is outside the representable range of a `boost::posix_time`, then the `start_date_` will be "not_a_date" and that will cause all kinds of problems.  If we really want this check, then the simulation should stop if the start time is outside the representable range.

bska (commented):
This is good for the most part, but the `start_date_` handling is insufficient.

opm/simulators/timestepping/SimulatorTimer.cpp
daavid00: Thanks for looking at this, I have removed the last commit that added those lines and updated the PR description. Let me know if further changes are needed; otherwise, please run Jenkins.

bska:
jenkins build this please

PR #6337, opened 2025-06-19
Reduce Number of Guide Rate Well Potential Messages

bska:
This PR switches `updateGuideRatesForWells()` away from exchanging potential rates for individual wells through one collective message per well.  Instead, we gather well level potential rates for the current rank into a large vector, and exchange those values with all ranks in the current communicator in a single message.  While the size of that message is obviously larger (`3 * #wells`) and therefore uses more memory, we potentially gain runtime performance by having less latency and synchronisation cost.

### DIFF
--- a/opm/simulators/wells/GuideRateHandler.cpp
+++ b/opm/simulators/wells/GuideRateHandler.cpp
@@ -16,19 +16,32 @@
   You should have received a copy of the GNU General Public License
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
+
 #include <config.h>
-#include <opm/input/eclipse/EclipseState/Phase.hpp>
-#include <opm/input/eclipse/Schedule/Schedule.hpp>
-#include <opm/output/data/GuideRateValue.hpp>
-#include <opm/simulators/utils/DeferredLoggingErrorHelpers.hpp>
-#include <opm/simulators/utils/ParallelCommunication.hpp>
+
 #include <opm/simulators/wells/GuideRateHandler.hpp>
+
 #include <opm/common/TimingMacros.hpp>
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

opm/simulators/wells/GuideRateHandler.cpp
```diff
@@ -631,33 +643,57 @@ GuideRateHandler<Scalar>::UpdateGuideRates::
...
-                gaspot = wpot[pu.phase_pos[BlackoilPhases::Vapour]];
+    const auto g_pos = (pu.phase_used[BlackoilPhases::Vapour] > 0)
+        ? pu.phase_pos[BlackoilPhases::Vapour] : -1;
 
-            if (pu.phase_used[BlackoilPhases::Aqua] > 0)
-                waterpot = wpot[pu.phase_pos[BlackoilPhases::Aqua]];
+    const auto w_pos = (pu.phase_used[BlackoilPhases::Aqua] > 0)
+        ? pu.phase_pos[BlackoilPhases::Aqua] : -1;
+
+    const auto o_ix = std::size_t{0};
```
akva2: these can be constexpr

akva2 (approved):
Looks like an improvement to me

opm/simulators/wells/GuideRateHandler.cpp
bska: > these can be constexpr

Yep, you're right.  I've pushed an update to that effect now.

PR #6342, opened 2025-06-23
Fixed brace locations when compiling without CUDA

kjetilly:
An error was introduced in the previous PR where the build would fail without CUDA. This PR should fix that.


akva2:
jenkins build this rocm please

bska:
Build check is green.  Merging this now to restore the build.  Thanks a lot for the quick fix.

PR #6344, opened 2025-06-24
Some janitoring

akva2:
Quell shadowing warnings, https://ci.opm-project.org/job/opm-simulators/2827/gcc/folder.-238066453/


opm/simulators/flow/Transmissibility_impl.hpp
```diff
@@ -262,11 +262,11 @@ update(bool global, const TransUpdateQuantities update_quantities,
         centroids_cache_[elemIdx] = centroids_(elemIdx);
     }
 
-    auto harmonicMean = [](const Scalar half1, const Scalar half2)
+    auto harmonicMean = [](const Scalar half1, const Scalar Half2)
```
bska: I might consider calling these just `x1` and `x2` instead of mixing upper and lower case.

opm/simulators/flow/Transmissibility_impl.hpp
```diff
@@ -344,12 +344,12 @@ update(bool global, const TransUpdateQuantities update_quantities,
             auto computeHalfMean = [&inside, &outside, &computeHalf, &ntg, &harmonicMean]
                                    (const auto& halfComputer, const auto& prop)
             {
-                auto half = computeHalf(halfComputer, prop[inside.elemIdx], prop[outside.elemIdx]);
```
bska, line 347: If we're changing this anyway, maybe we could rename the result to `onesideded` or similar.  "Half" really isn't a good name for this.

bska (commented):
This obviously fixes the immediate problem, but I think we should take this opportunity to get better names instead.

opm/simulators/flow/Transmissibility_impl.hpp
akva2: ok.

opm/simulators/flow/Transmissibility_impl.hpp
akva2: ok.

bska (approved):

PR #6346, opened 2025-06-25
Add mobility thresholding to NLDD

jakobtorben:
# NLDD Solver: Skip Unchanged Subdomains Based on Mobility Changes

This PR introduces an a new feature to the NLDD (Nonlinear Domain Decomposition) solver that reduces the time spent on local solves by skipping subdomains that haven't changed significantly between iterations.

Mobility in reservoir simulation represents the ease with which a fluid phase can flow through porous media. It is calculated as the ratio of relative permeability to viscosity for each phase $\alpha$:
```math
\lambda_\alpha = \mathrm{kr}_\alpha / \mu_\alpha
```
Mobility changes indicate significant alterations in fluid flow characteristics, making it an effective indicator for determining when subdomain re-solving is necessary.

#### Adaptive Domain Solving
- To determine if we need to solve a domain the relative change in mobility is compared to a previous value and if it is above a given tolerance, the domain is marked for needing solving.
- **Relative change calculation**: 
    `|current_mobility - previous_mobility| / average_cell_mobility > threshold`
- If we do an initial global Newton iteration (default), the change in mobility between before and after this global iteration will be used for the next NLDD iteration.
- If we start with an NLDD iteration all domains will be solved in the first iteration and the change in mobility between before and after this complete NLDD iteration will be used for the next NLDD iteration.
- Domains that required solving in the previous iteration are automatically marked for solving in the next iteration.

####  Reporting
- Added tracking of skipped domains in simulation reports

#### New Parameter
- `--nldd-relative-mobility-change-tol`: Threshold for relative mobility change detection (default: 0.1). By setting this parameter to 0, the mobility thresholding will be disabled.

### Performance Impact
- It has been tested on multiple real reservoir assets, where it reduced the local solve time.
- The reduction in local solve time is most significant for serial runs, for obvious reasons.

##### Baseline

```
Number of MPI processes:         1
Threads per MPI process:        12
Setup time:                      1.88 s
  Deck input:                    1.25 s
Number of timesteps:           247
Simulation time:               143.50 s
  Assembly time:                12.50 s (Wasted: 0.0 s; 0.0%)
    Well assembly:               1.15 s (Wasted: 0.0 s; 0.0%)
  Linear solve time:            50.67 s (Wasted: 0.0 s; 0.0%)
    Linear setup:               19.21 s (Wasted: 0.0 s; 0.0%)
  Local solve time:             39.29 s (Wasted: 0.0 s; 0.0%)
  Props/update time:             9.04 s (Wasted: 0.0 s; 0.0%)
  Pre/post step:                29.03 s (Wasted: 0.0 s; 0.0%)
  Output write time:             0.38 s
Overall Linearizations:       1257      (Wasted:     0; 0.0%)
Overall Newton Iterations:     590      (Wasted:     0; 0.0%)
Overall Linear Iterations:    1531      (Wasted:     0; 0.0%)`
```
NORNE_ATW2013_nonlinear_sorted_baseline_NT12

##### Mobility thresholding

```
Number of MPI processes:         1
Threads per MPI process:        12
Setup time:                      1.88 s
  Deck input:                    1.26 s
Number of timesteps:           247
Simulation time:               141.01 s
  Assembly time:                12.79 s (Wasted: 0.0 s; 0.0%)
    Well assembly:               1.15 s (Wasted: 0.0 s; 0.0%)
  Linear solve time:            51.64 s (Wasted: 0.0 s; 0.0%)
    Linear setup:               19.70 s (Wasted: 0.0 s; 0.0%)
  Local solve time:             34.92 s (Wasted: 0.0 s; 0.0%)
  Props/update time:             9.14 s (Wasted: 0.0 s; 0.0%)
  Pre/post step:                29.56 s (Wasted: 0.0 s; 0.0%)
  Output write time:             0.38 s
Overall Linearizations:       1273      (Wasted:     0; 0.0%)
Overall Newton Iterations:     605      (Wasted:     0; 0.0%)
Overall Linear Iterations:    1547      (Wasted:     0; 0.0%)
```

NORNE_ATW2013_nonlinear_sorted_mob_thres_NT12

##### Baseline

```
Number of MPI processes:        12
Threads per MPI process:         1
Setup time:                      2.63 s
  Deck input:                    1.64 s
Number of timesteps:           247
Simulation time:                91.04 s
  Assembly time:                15.04 s (Wasted: 0.0 s; 0.0%)
    Well assembly:               0.84 s (Wasted: 0.0 s; 0.0%)
  Linear solve time:            27.88 s (Wasted: 0.0 s; 0.0%)
    Linear setup:               12.02 s (Wasted: 0.0 s; 0.0%)
  Local solve time:             17.32 s (Wasted: 0.0 s; 0.0%)
  Props/update time:             7.04 s (Wasted: 0.0 s; 0.0%)
  Pre/post step:                16.99 s (Wasted: 0.0 s; 0.0%)
  Output write time:             4.19 s
Overall Linearizations:       1409      (Wasted:     0; 0.0%)
Overall Newton Iterations:     675      (Wasted:     0; 0.0%)
Overall Linear Iterations:    2154      (Wasted:     0; 0.0%)
```
NORNE_ATW2013_nonlinear_sorted_baseline_NP12

##### Mobility thresholding

```
Number of MPI processes:        12
Threads per MPI process:         1
Setup time:                      2.61 s
  Deck input:                    1.57 s
Number of timesteps:           247
Simulation time:                90.13 s
  Assembly time:                15.43 s (Wasted: 0.0 s; 0.0%)
    Well assembly:               0.83 s (Wasted: 0.0 s; 0.0%)
  Linear solve time:            27.62 s (Wasted: 0.0 s; 0.0%)
    Linear setup:               11.83 s (Wasted: 0.0 s; 0.0%)
  Local solve time:             16.04 s (Wasted: 0.0 s; 0.0%)
  Props/update time:             7.13 s (Wasted: 0.0 s; 0.0%)
  Pre/post step:                17.19 s (Wasted: 0.0 s; 0.0%)
  Output write time:             4.14 s
Overall Linearizations:       1377      (Wasted:     0; 0.0%)
Overall Newton Iterations:     660      (Wasted:     0; 0.0%)
Overall Linear Iterations:    2115      (Wasted:     0; 0.0%)
```
NORNE_ATW2013_nonlinear_sorted_mob_thres_NP12

### DIFF
--- a/opm/simulators/flow/BlackoilModelNldd.hpp
+++ b/opm/simulators/flow/BlackoilModelNldd.hpp
@@ -1100,13 +1100,9 @@ private:
 
     bool checkSubdomainChangeRelative(const Domain& domain)
     {
-        // Threshold for domain-wide average relative change
-        const Scalar averageRelativeMobilityChangeTol = 0.005;
-
         int numCells = domain.cells.size();
         auto& prevMobilities = domainPreviousMobilities_[domain.index];
         bool needsSolving = false;
-        Scalar totalRelativeChange = 0.0;
 
         // Check mobility changes for all cells in the domain
         for (int cellIdx = 0; cellIdx < numCells; ++cellIdx) {
@@ -1123,7 +1119,6 @@ private:
             }
 
             // Check relative changes for each phase
-            Scalar cellMaxRelDiff = 0.0;
             for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
                 if (!FluidSystem::phaseIsActive(phaseIdx)) {
                     continue;
@@ -1134,26 +1129,10 @@ private:
 
                 // Store current value for next iteration
                 prevMobilities[mobIdx] = mobility;
-                // Check if this change exceeds single-cell tolerance
                 if (relDiff > model_.param().nldd_relative_mobility_change_tol_) {
                     needsSolving = true;
                 }
-                cellMaxRelDiff = std::max(cellMaxRelDiff, relDiff);
             }
-            totalRelativeChange += cellMaxRelDiff;
-        }
-        if (!needsSolving) {
-            OpmLog::debug(fmt::format("Domain {} skipped: as no cell has a relative change larger than {}",
-                                     domain.index, model_.param().nldd_relative_mobility_change_tol_));
-        }
-        // Check if average change exceeds domain-wide tolerance
-        const Scalar avgRelativeChange = totalRelativeChange / numCells;
-        if (avgRelativeChange > averageRelativeMobilityChangeTol) {
-            needsSolving = true;
-        }
-        if (avgRelativeChange < averageRelativeMobilityChangeTol) {
-            OpmLog::debug(fmt::format("Domain {} skipped: avg relative change {} is below threshold {}",
-                                     domain.index, avgRelativeChange, averageRelativeMobilityChangeTol));
         }
         return needsSolving;
     }
--- a/opm/simulators/flow/BlackoilModelNldd.hpp
+++ b/opm/simulators/flow/BlackoilModelNldd.hpp
@@ -95,6 +95,8 @@ public:
 
     static constexpr int numEq = Indices::numEq;
 
+    enum { numPhases = getPropValue<TypeTag, Properties::NumPhases>() };
+
     //! \brief The constructor sets up the subdomains.
     //! \param model BlackOil model to solve for
     //! \param param param Model parameters
@@ -178,6 +180,13 @@ public:
                                         skip);
         }
 
+        // Initialize storage for previous mobilities per domain
+        domainPreviousMobilities_.resize(domains_.size());
+        for (size_t domIdx = 0; domIdx < domains_.size(); ++domIdx) {
+            const auto& domain = domains_[domIdx];
+            domainPreviousMobilities_[domIdx].resize(domain.cells.size() * numPhases, 0.0);
+        }
+
         // Set up container for the local system matrices.
         domain_matrices_.resize(num_domains);
 
@@ -257,7 +266,10 @@ public:
             SimulatorReportSingle local_report;
             detailTimer.reset();
             detailTimer.start();
-            if (domain.skip) {
+
+            bool needsSolving = checkIfSubdomainNeedsSolving(domain);
+
+            if (domain.skip || !needsSolving) {
                 local_report.converged = true;
                 domain_reports[domain.index] = local_report;
                 continue;
@@ -571,7 +583,6 @@ private:
             if (!convreport.converged() && !convreport.wellFailed()) {
                 bool oscillate = false;
                 bool stagnate = false;
-                const int numPhases = convergence_history.front().size();
                 detail::detectOscillations(convergence_history, iter, numPhases,
                                            Scalar{0.2}, 1, oscillate, stagnate);
                 if (oscillate) {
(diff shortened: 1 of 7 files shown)


jakobtorben:
jenkins build this please

atgeirr:
> * `nldd_relative_mobility_change_tol_`: Threshold for relative mobility change detection (default: 0.1). By setting this parameter to 0, the mobility thresholding will be disabled.

For the manual: the name of the command line option is `--nldd-relative-mobility-change-tol`

jakobtorben:
Good clarification Atgeirr, I updated the comment to the actual CLI name to make this clearer.

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -253,11 +270,17 @@ class BlackoilModelNldd
         DeferredLogger logger;
         std::vector<SimulatorReportSingle> domain_reports(domains_.size());
         for (const int domain_index : domain_order) {
-            const auto& domain = domains_[domain_index];
+            auto& domain = domains_[domain_index];
```
atgeirr: Prefer this to remain const. Maybe we should put the flags you modify in this class instead? Or maybe just return them from checkIfSundomainNeedsSolving() etc.?

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -1065,6 +1097,88 @@ class BlackoilModelNldd
...
 
+    void updateMobilities(const Domain& domain)
+    {
+        if (domain.skip || model_.param().nldd_relative_mobility_change_tol_ == 0.0) {
+            return;
+        }
+
+        for (int cellIdx = 0; cellIdx < domain.cells.size(); ++cellIdx) {
+            unsigned globalDofIdx = domain.cells[cellIdx];
+            const auto& intQuants = *model_.simulator().model().cachedIntensiveQuantities(globalDofIdx, /*timeIdx=*/0);
```
atgeirr: Use `model->intensiveQuantities(globalDofIdx, /* time_idx = */ 0)` instead, this is a bit more robust wrt. cache availability  (see FIBlackoilModel.hpp for the implementation used in most Flow instantiations, it throws if no cache, but one could use a variant that calculated instead of throwing).

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -95,6 +95,8 @@ class BlackoilModelNldd
 
     static constexpr int numEq = Indices::numEq;
 
+    enum { numPhases = getPropValue<TypeTag, Properties::NumPhases>() };
```
atgeirr: Prefer FluidSystem::numPhases, since the FluidSystem gives the context in which it will be used (lines 187, 1110+). May be better to just drop it here, and only get it near the above-mentioned lines.

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -571,7 +604,6 @@ class BlackoilModelNldd
             if (!convreport.converged() && !convreport.wellFailed()) {
                 bool oscillate = false;
                 bool stagnate = false;
-                const int numPhases = convergence_history.front().size();
```
atgeirr, line 574: This is wrong, as in this version of the code numPhases should be equal to FluidSystem::numPhases (which is 3 for the blackoil fluid system, irrespective of the number of active phases), due to lines 1110++. However here we should have a 2 if we are in a two-phase run. It was maybe ambiguous to call this variable numPhases anyway, as that might indicate that it is FluidSystem::numPhases, I suggest just renaming the variable to num_residuals for example to avoid the mixup.

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -178,6 +180,14 @@ class BlackoilModelNldd
                                         skip);
         }
 
+        // Initialize storage and values for previous mobilities per domain
+        domainPreviousMobilities_.resize(domains_.size());
+        for (size_t domIdx = 0; domIdx < domains_.size(); ++domIdx) {
+            const auto& domain = domains_[domIdx];
+            domainPreviousMobilities_[domIdx].resize(domain.cells.size() * numPhases, 0.0);
```
atgeirr: This will store 3 mobilities per cell also for two-phase cases. Consider indexing by active phase instead of canonical phase.

Also: consider storing in a single vector (by global cell index) instead of one vector per domain.

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -1065,6 +1097,88 @@ class BlackoilModelNldd
...
+        if (domain.skip || model_.param().nldd_relative_mobility_change_tol_ == 0.0) {
+            return;
+        }
+
+        for (int cellIdx = 0; cellIdx < domain.cells.size(); ++cellIdx) {
+            unsigned globalDofIdx = domain.cells[cellIdx];
+            const auto& intQuants = *model_.simulator().model().cachedIntensiveQuantities(globalDofIdx, /*timeIdx=*/0);
+
+            for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
+                if (!FluidSystem::phaseIsActive(phaseIdx)) {
```
atgeirr: Consider iterating over (and storing) just active phases in the first place, using `FluidSystem::activeToCanonicalPhaseIdx(activePhaseIdx)` to get the correct index for the intQuants.mobility() call.

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -1065,6 +1097,88 @@ class BlackoilModelNldd
                                      param.local_domains_partition_well_neighbor_levels_);
     }
 
+    void updateMobilities(const Domain& domain)
+    {
+        if (domain.skip || model_.param().nldd_relative_mobility_change_tol_ == 0.0) {
+            return;
+        }
+
+        for (int cellIdx = 0; cellIdx < domain.cells.size(); ++cellIdx) {
```
atgeirr: If storing in single vector instead of vector of vectors as suggested above, this can become a range-for over domain.cells.

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -1065,6 +1097,88 @@ class BlackoilModelNldd
...
+            domain.needsSolving = true;
+        } else {
+            domain.needsSolving = checkSubdomainChangeRelative(domain);
+        }
+    }
+
+    bool checkSubdomainChangeRelative(const Domain& domain)
+    {
+        int numCells = domain.cells.size();
+        auto& prevMobilities = domainPreviousMobilities_[domain.index];
```
atgeirr: Const!

opm/simulators/flow/BlackoilModelNldd.hpp
```diff
@@ -1065,6 +1097,88 @@ class BlackoilModelNldd
...
+
+        // Check mobility changes for all cells in the domain
+        for (int cellIdx = 0; cellIdx < numCells; ++cellIdx) {
```
atgeirr: Several suggestions made earlier must be matched here if adopted: storing in one vector, only active phases, avoiding explicit cachedIntensiveQuantities(), making this loop a range-for over domain.cells.

jakobtorben:
jenkins build this please

opm/simulators/flow/BlackoilModelNldd.hpp
jakobtorben: I was contemplating where this flag should live, as it is updated in different places in `BlackOilModelNldd`, and we could extend it to be updated to even more places later. Storing it as a member in `BlackOilModelNldd` works as well, so I changed to this approach now, such that this can be `const` again.

atgeirr:
Looks very good now, all is resolved and Jenkins is happy. Merging!

PR #6347, opened 2025-06-26
Reservoir coupling: Send master group targets to slaves

hakonhagland:
Calculates the master group targets based on the slave group potentials, then sends those targets to the slave groups. The application of the received slave group targets to calculate the flow of the slave groups will be handled by a future PR.

### DIFF
--- a/opm/simulators/wells/BlackoilWellModelGeneric.cpp
+++ b/opm/simulators/wells/BlackoilWellModelGeneric.cpp
@@ -601,8 +601,15 @@ getGroupFipnumAndPvtreg() const
     // Use the pvtRegionIdx from the top cell of the first well.
     // TODO fix this!
     // This is only used for converting RESV rates.
-    // What is the proper approach?
-    // See discussion in: https://github.com/OPM/opm-simulators/issues/2921 for more details.
+    //
+    // Background (from opm-simulators issue #2921):
+    // The fundamental problem is selecting a single PVT region for a group that may
+    // include multiple wells, each potentially perforated in different PVT regions.
+    // WELSPECS items 11 and 13 define well-level region mappings, but there's no
+    // perfect solution for groups spanning multiple regions. The current approach
+    // uses the first perforation cell of the globally first well (by Well::seqIndex())
+    // for consistency across serial and parallel runs (see PR #2926).
+    // See: https://github.com/OPM/opm-simulators/issues/2921
     const int fipnum = 0;
     int pvtreg = well_perf_data_.empty() || well_perf_data_[0].empty()
         ? pvt_region_idx_[0]


hakonhagland:
jenkins build this please

hakonhagland:
jenkins build this please

hakonhagland:
jenkins build this please

hakonhagland:

blattms:

opm/simulators/flow/ReservoirCouplingMaster.cpp
```diff
@@ -61,24 +62,46 @@ ReservoirCouplingMaster(
...
            this->master_group_slave_names_.end();
 }
 
+template <class Scalar>
+const std::vector<std::string>&
+ReservoirCouplingMaster<Scalar>::
+getMasterGroupNamesForSlave(std::size_t slave_idx) const
+{
+    const auto& slave_name = this->getSlaveName(slave_idx);
+    auto it = this->slave_name_to_master_groups_map_.find(slave_name);
```
blattms: If this is called multiple times then in the long run it might be a good idea to also store whether the slave has a group in a vector and do the lookup by index.

opm/simulators/flow/ReservoirCouplingMaster.cpp
```diff
@@ -254,22 +284,111 @@ receivePotentialsFromSlaves()
...
-ReservoirCouplingMaster::
-numSlaveGroups(unsigned int index)
+template <class Scalar>
+void
+ReservoirCouplingMaster<Scalar>::
+sendInjectionTargetsToSlave(std::size_t slave_idx,
+                            const std::vector<InjectionGroupTarget>& injection_targets) const
 {
-    return this->master_group_name_order_[this->slave_names_[index]].size();
+    if (this->comm_.rank() == 0) {
```
blattms: I this there should be a comment about this if statement with any else. Other we might wonder whether we need a throwing else branch or at least an assert. The same holds for the functions below. I will not comment on those again.

opm/simulators/flow/ReservoirCouplingMaster.hpp
```diff
@@ -58,6 +62,7 @@ class ReservoirCouplingMaster {
     char *getArgv(int index) const { return this->argv_[index]; }
     char **getArgv() const { return this->argv_; }
     const Parallel::Communication &getComm() const { return this->comm_; }
+    const std::vector<std::string>& getMasterGroupNamesForSlave(std::size_t slave_idx) const;
```
blattms, line 78: Please add a bit of documentation

opm/simulators/flow/ReservoirCouplingSlave.cpp
```diff
@@ -59,17 +60,21 @@ ReservoirCouplingSlave(
...
-ReservoirCouplingSlave::
-sendAndReceiveInitialData() {
-    this->sendActivationDateToMasterProcess_();
-    this->sendSimulationStartDateToMasterProcess_();
-    this->receiveSlaveNameFromMasterProcess_();
-    this->receiveMasterGroupNamesFromMasterProcess_();
+ReservoirCouplingSlave<Scalar>::
+maybeActivate(int report_step) {
+    if (!this->activated()) {
+        auto rescoup = this->schedule_[report_step].rescoup();
```
blattms: Is this copy intended? Might it hurt performance?

opm/simulators/flow/ReservoirCouplingSpawnSlaves.cpp
```diff
@@ -168,12 +193,14 @@ getSlaveArgv_(
...
 
+template <class Scalar>
 std::pair<std::vector<char>, std::size_t>
-ReservoirCouplingSpawnSlaves::
+ReservoirCouplingSpawnSlaves<Scalar>::
 getMasterGroupNamesForSlave_(const std::string &slave_name) const
 {
     // For the given slave name, get all pairs of master group names and slave group names
     // Serialize the data such that it can be sent over MPI in one chunk
+    // NOTE: The order of the names in the vector is important, as the slaves will
```
blattms, line 202: Please complete the sentence.
Is there a similar note in the class documentation of the type of rescoup_? That is where people using the class would look.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -605,23 +605,21 @@ checkGconsaleLimits(const Group& group,
...
+std::pair<int, int>
+BlackoilWellModelGeneric<Scalar>::
+getGroupFipnumAndPvtreg() const
 {
     // Set up coefficients for RESV <-> surface rate conversion.
     // Use the pvtRegionIdx from the top cell of the first well.
     // TODO fix this!
     // This is only used for converting RESV rates.
     // What is the proper approach?
+    // See discussion in: https://github.com/OPM/opm-simulators/issues/2921 for more details.
```
blattms: In the event that this project is ever moved that link might not be valid or people might be offline Maybe include a short summary?

tests/rescoup/test_chopstep.cpp
```diff
@@ -165,7 +165,7 @@ struct SimulatorFixture
     OpmSimulatorTestCase simulator_wrapper;
     Simulator *simulator;
     Opm::Schedule& schedule;
-    Opm::ReservoirCouplingMaster rc_master;
+    Opm::ReservoirCouplingMaster<double> rc_master;
```
blattms: Shouldn't there be a test with float, too?

blattms (commented):
First (easy part) of my review. I have not looked at the new files, yet.

Is  there a test for the new functionality or is it implicitly covered by the existing ones?

opm/simulators/wells/GroupTargetCalculator.hpp
```diff
@@ -0,0 +1,198 @@
...
+    };
+    struct InjectionTargetInfo {
+        Scalar target;
+        Group::InjectionCMode cmode;
+    };
+
+    // Generalizes calculation for both injectors and producers to avoid code duplication.
+    class GeneralCalculator {
+    public:
+        GeneralCalculator(
```
blattms, line 106: Some documentation might not hurt if something is not self-explanatory.

E.g. what exactly is a GeneralCaclulator, what is it used for and what is the calculator in it.

opm/simulators/flow/ReservoirCouplingMaster.cpp
```diff
@@ -76,6 +76,12 @@ const std::vector<std::string>&
 ReservoirCouplingMaster<Scalar>::
 getMasterGroupNamesForSlave(std::size_t slave_idx) const
 {
+    // Fast path: O(1) direct vector access for performance-critical usage
+    if (slave_idx < this->slave_idx_to_master_groups_.size()) {
+        return this->slave_idx_to_master_groups_[slave_idx];
+    }
+
+    // Fallback to existing implementation for bounds checking and error handling
```
blattms: Maybe I am missing something. I do not totally understand the fallback.

If there is an additional group and that name is in the map, then aren't the iterators (and the other indices) invalidated?

And if nothing changed then we could just throw here.

opm/simulators/flow/SimulatorFullyImplicitBlackoil.hpp
```diff
@@ -650,13 +651,13 @@ class SimulatorFullyImplicitBlackoil : private SerializableSim
...
+    std::unique_ptr<ReservoirCouplingSlave<Scalar>> reservoirCouplingSlave_{nullptr};
 #endif
 
     SimulatorSerializer serializer_;
 };
 
 } // namespace Opm
 
-#endif // OPM_SIMULATOR_FULLY_IMPLICIT_BLACKOIL_HEADER_INCLUDED
+#endif // OPM_SIMULATOR_FULLY_IMPLICIT_BLACKOIL_HEADER_INCLUDED
```
blattms: Not 100% sure. But it might be good to add a linebreak at the end of files.

tests/rescoup/test_chopstep.cpp
hakonhagland: Thanks for the suggestion. I investigated adding float type testing, but discovered it would require extensive refactoring beyond this PR's scope.

The main blocker is that `AdaptiveTimeStepping` (where the `maybeChopSubStep()` will be used) is hardcoded to use `double` throughout. Templating it would be a major architectural change affecting core OPM components.

I've added a TODO comment in the test file documenting this limitation.

This should be tracked as technical debt for a future dedicated refactoring effort rather than expanding the current PR's scope.

akva2: The use of doubles for time stepping is on purpose. Floats cannot represent time values with sufficient precision.

hakonhagland: Makes sense. What are we trying to achieve with the Scalar=float? Save memory?

akva2: Well, my original intention was to enable gpu experimentation with full float simulators.

hakonhagland:
jenkins build this please

hakonhagland:
jenkins build this please

hakonhagland:

PR #6348, opened 2025-06-26
distinguishing injector or producter when setting well to be under pressure control during well testing.

GitPaean:
This is a minimum that we do to follow up the https://github.com/OPM/opm-simulators/pull/6319 . 

potentially, we should use `solveWellWithOperabilityCheck` to solve for testing for the injectors also, while let us do this one first. 

When it is smaller, it is easier to check.


GitPaean:
jenkins build this please failure_report please

opm/simulators/wells/WellInterface_impl.hpp
```diff
@@ -396,13 +396,13 @@ namespace Opm
 
         const auto& summary_state = simulator.vanguard().summaryState();
         const bool has_thp_limit = this->wellHasTHPConstraints(summary_state);
-        if (has_thp_limit) {
-            ws.production_cmode = Well::ProducerCMode::THP;
-        }
-        else {
-            ws.production_cmode = Well::ProducerCMode::BHP;
+        const bool is_producer = this->isProducer();
+        if (is_producer) {
```
bska: Since you appear not to use the new `is_producer` flag for anything other than this check, you might as well write it as
```C++
if (this->isProducer()) {
    // ...
}
else {
    // ...
}
```
instead.  That way we don't have to introduce a new name/object into the scope which is already somewhat busy.

GitPaean: You are correct. I updated the PR now.

GitPaean:
jenkins build this please failure_report please

bska (approved):
Thanks a lot for the updates.  This looks good to me now and as the build check is green, I'll merge into master.

PR #6341, opened 2025-06-23
Add support for debug levels in DeferredLogger

vkip:
Defaulting level parameter to 0, so output should not change for now.

PR #6350, opened 2025-06-26
Emit Converged in Zero Iteration Messages at Elevated Verbosity Level

bska:
That way, the .DBG file won't have significant number of

> Well P-1 converged in 0 inner iterations

messages unless the user specifically requests such messages.

This commit leverages the debug level support introduced in PRs OPM/opm-common#4643 and #6341.


bska:
jenkins build this please

atgeirr (approved):

PR #6353, opened 2025-06-26
Using partial specialization to save memory

atgeirr:
For discussion now.


atgeirr:
I approve this too, merge when Jenkins is ok.

PR #6305, opened 2025-05-22
Replace call to groupControlledWells with precomputed values

totto82:
This is one part of #6287

PR #6355, opened 2025-06-26
Replace group controls wells and fix

totto82:
This is #6305 + adding phase to injector group switching output + a fix that ensures we only count group-controlled wells per phase.


totto82:

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1372,10 +1372,26 @@ updateAndCommunicateGroupData(const int reportStepIdx,
...
         }
     }
-
-
     auto& well_state = this->wellState();
     const auto& well_state_nupcol = this->nupcolWellState();
+
+
+
+    std::vector<bool> is_production_group = {true, false, false, false};
```
atgeirr: This part is vulnerable to inconsistent changes. I suggest:
```C++
constexpr int num_configs = 4;
constexpr std::array<bool, num_configs> is_production_group = {true, false, false, false};
constexpr std::array<Phase, num_configs> phases = { Phase::OIL, Phase::WATER, Phase::OIL, Phase::GAS };
for (int i = 0; i < num_configs; i++) {
...
```
This guarantees that the number of configs remains consistent, and also avoids unnecessary memory allocation.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -1372,10 +1372,26 @@ updateAndCommunicateGroupData(const int reportStepIdx,
             }
         }
     }
-
-
     auto& well_state = this->wellState();
     const auto& well_state_nupcol = this->nupcolWellState();
+
```
atgeirr, line 1377: Use just a single blank line inside a function. Add a comment to explain the next code block if it needs to stand out more.

opm/simulators/wells/BlackoilWellModelGeneric.cpp
```diff
@@ -2186,9 +2202,12 @@ reportGroupSwitching(DeferredLogger& local_deferredLogger) const
             const Group::InjectionCMode& oldControl =
                 this->prevWGState().group_state.injection_control(grname, phase);
             if (ctrls.back() != oldControl) {
+                std::ostringstream ss;
+                ss << phase;
                 const std::string msg =
-                    fmt::format("    Injection Group {} control model changed from {} to {}",
+                    fmt::format("    Injection Group {} (phase = {}) control model changed from {} to {}",
                                 grname,
+                                ss.str(),
```
atgeirr, line 2208: We should consider adding a formatter specialization instead of using a stringstream here, see https://fmt.dev/latest/api/#ostream-api

This should be a one-liner, but I cannot decide where it should be defined. Preferably in Phase.hpp, but then we do not want to include <fmt/format.h> in headers, so there is a dilemma. Because of that I do not require this to be changed, but we should keep this in mind for the future.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -1228,6 +1238,35 @@ groupControlledWells(const Schedule& schedule,
...
+groupControlledWells(const Schedule& schedule,
+                     const WellState<Scalar>& well_state,
+                     const GroupState<Scalar>& group_state,
+                     const int report_step,
+                     const std::string& group_name,
+                     const std::string& always_included_child,
+                     const bool is_production_group,
+                     const Phase injection_phase)
+{
+    auto num_wells = is_production_group? group_state.number_of_wells_under_this_control(group_name): group_state.number_of_wells_under_this_inj_control(group_name, injection_phase);
```
atgeirr: Use `const`. Should have space before `?` and `:` if on a single line, however this line is far too long so it should be split.

All of this also for line 1265.

opm/simulators/wells/WellGroupHelpers.hpp
```diff
@@ -262,16 +262,24 @@ class WellGroupHelpers
...
-                                    const GroupState<Scalar>& group_state,
+                                    GroupState<Scalar>& group_state,
                                     const SummaryState& summary_state,
                                     const GuideRate* guideRate,
                                     const int report_step,
                                     const std::string& group_name,
                                     const std::string& always_included_child,
                                     const bool is_production_group,
                                     const Phase injection_phase);
+    static int groupControlledWells(const Schedule& schedule,
```
atgeirr, line 278: Please add comments explaining all three functions added/changed here.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -1262,6 +1301,35 @@ groupChainTopBot(const std::string& bottom,
...
+    // Get initial parent, 'bottom' can be a well or a group.
+    std::string parent;
+    if (schedule.hasWell(bottom, report_step)) {
+        parent = schedule.getWell(bottom, report_step).groupName();
+    } else {
+        parent = schedule.getGroup(bottom, report_step).parent();
+    }
+
+    // Build the chain from bottom to top.
+    std::vector<std::string> chain;
```
atgeirr: It is not necessary to build the chain just to determine a bool. Instead of
```
    chain.push_back(parent);
    if (parent == "FIELD")
        break;
```
do the following:
```
    if (parent == top) {
        return true;
    } else if (parent == "FIELD")
        return false;
    }
```
The order is important to ensure it returns true if `top` is FIELD.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -1262,6 +1301,35 @@ groupChainTopBot(const std::string& bottom,
     return chain;
 }
 
+template<class Scalar>
+bool
+WellGroupHelpers<Scalar>::
+isInGroupChainTopBot(const std::string& bottom,
+                 const std::string& top,
```
atgeirr: Indentation. Also above for updateGroupControlledWells().

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -1122,9 +1122,9 @@ getGuideRateInj(const std::string& name,
 
 template<class Scalar>
 int WellGroupHelpers<Scalar>::
-groupControlledWells(const Schedule& schedule,
+updateGroupControlledWells(const Schedule& schedule,
```
atgeirr, line 1115: Are you sure that the always_included_child should be an argument to this function? It seems to me it should not, since the new groupControlledWells() function adds 1 if appropriate.

opm/simulators/wells/GroupState.hpp
```diff
@@ -104,6 +104,15 @@ class GroupState {
     void injection_control(const std::string& gname, Phase phase, Group::InjectionCMode cmode);
     Group::InjectionCMode injection_control(const std::string& gname, Phase phase) const;
 
+    bool has_number_of_wells_under_this_control(const std::string& gname) const;
```
atgeirr: These functions are not used at all and could be removed, or maybe instead they should be called in some places? In any case, throwing as done in number_of_wells_under_this_control() etc. is probably the most reasonable way to go, there is not much point in adding a call to has_number_of...() just to throw if it is false, as that will not behave any different.

Alternatively, depending on if and where you would want to ask has_number_of..() you could return std::optional<int> instead from number_of_wells...().

opm/simulators/wells/GroupState.hpp
```diff
@@ -104,6 +104,15 @@ class GroupState {
...
+    void update_number_of_wells_under_this_control(const std::string& gname, int number);
+    int number_of_wells_under_this_control(const std::string& gname) const;
```
atgeirr: The function naming is a bit strange. I suggest `number_of_wells_under_group_control(const std::string group_name)` for example.

opm/simulators/wells/WellGroupHelpers.cpp
totto82: At least we need it if always_included_child is a group. I checked very carefully that the new and old code returned the same numbers for groupControlledWells, but you may be right that we double-count wells if they are always included. Could this also have been the case before? I will do some checking. For now, we only use ut to check >0, so it shouldn't matter for the results, but it would be a hard bug to find if we start using the actual number of wells here.

totto82: You're right; we can remove it. It is always empty, as it is used now. I will update accordingly.

opm/simulators/wells/WellGroupHelpers.cpp
```diff
@@ -1218,6 +1227,37 @@ groupControlledWells(const Schedule& schedule,
...
+                     const std::string& always_included_child,
+                     const bool is_production_group,
+                     const Phase injection_phase)
+{
+    auto num_wells = is_production_group ? group_state.number_of_wells_under_group_control(group_name)
+        : group_state.number_of_wells_under_inj_group_control(group_name, injection_phase);
+    if (schedule.hasWell(always_included_child, report_step)) {
+        const bool isInGroup = isInGroupChainTopBot(always_included_child, group_name, schedule, report_step);
+        const bool already_included = is_production_group ? well_state.isProductionGrup(always_included_child)
+        : well_state.isInjectionGrup(always_included_child);
```
atgeirr: Indent.

opm/simulators/wells/WellGroupHelpers.hpp
```diff
@@ -262,11 +262,22 @@ class WellGroupHelpers
                                   const Phase& injectionPhase,
                                   const PhaseUsage& pu);
 
+    /// update the number of wells that are activly under group control for a given group with name given by group_name
```
atgeirr: Typo: activly -> actively here and on line 277.

opm/simulators/wells/WellGroupHelpers.hpp
```diff
@@ -314,6 +325,12 @@ class WellGroupHelpers
                      const Schedule& schedule,
                      const int report_step);
 
+    static bool
```
atgeirr, line 330: Please document the function.

totto82:

totto82:

PR #6356, opened 2025-06-26
Janitoring: remove returns after throw

akva2:
Causes SCA warnings


akva2:
jenkins build this please

bska (approved):
Very good.  Please merge when the build check is green.

PR #6357, opened 2025-06-26
Remove intensive quantities cache for time idx 1

jakobtorben:
The `intensiveQuantitiesCache` is an important optimisation measure that avoids recalculating the properties several times. However, it takes up a significant amount of memory, which was measured to be around 20 % of the peak memory allocations for the entire simulation.  So it is important to use strictly what is necessary.

With our current implementation, having an `intensiveQuantitiesCache` for both timeIdx 0 (the current timestep) and 1 (the previous timestep) is not needed, as we only need it for timeIdx 0. Removing the cache for timeIdx 1 halves the memory use of this cache. This amounts to **reducing the peak memory use of the entire simulator by 10 %**, without any performance degradation.

Setting this as a draft for now, as we need to verify that the updating of the `storageCache` for timeIdx 1 is done elsewhere in a safe way that such that developers that want to add new models do not have to find out if this is updated or not.

### DIFF
--- a/opm/models/discretization/common/fvbasediscretization.hh
+++ b/opm/models/discretization/common/fvbasediscretization.hh
@@ -839,6 +839,7 @@ public:
     const EqVector& cachedStorage(unsigned globalIdx, unsigned timeIdx) const
     {
         assert(enableStorageCache_);
+        assert(storageCacheUpToDate_[timeIdx][globalIdx] != 0);
         return storageCache_[timeIdx][globalIdx];
     }
 
@@ -857,6 +858,48 @@ public:
     {
         assert(enableStorageCache_);
         storageCache_[timeIdx][globalIdx] = value;
+        storageCacheUpToDate_[timeIdx][globalIdx] = 1;
+    }
+
+    /*!
+     * \brief Returns true if the storage cache entry for a given DOF and time index is up to date.
+     *
+     * \param globalIdx The global space index for the entity
+     * \param timeIdx The index used by the time discretization
+     */
+    bool storageCacheIsUpToDate(unsigned globalIdx, unsigned timeIdx) const
+    {
+        if (!enableStorageCache_) {
+            return false;
+        }
+        return storageCacheUpToDate_[timeIdx][globalIdx] != 0;
+    }
+
+    /*!
+     * \brief Invalidate the storage cache for a given DOF and time index.
+     *
+     * \param globalIdx The global space index for the entity
+     * \param timeIdx The index used by the time discretization
+     */
+    void invalidateStorageCacheEntry(unsigned globalIdx, unsigned timeIdx) const
+    {
+        if (enableStorageCache_) {
+            storageCacheUpToDate_[timeIdx][globalIdx] = 0;
+        }
+    }
+
+    /*!
+     * \brief Invalidate the whole storage cache for a given time index.
+     *
+     * \param timeIdx The index used by the time discretization
+     */
+    void invalidateStorageCache(unsigned timeIdx) const
+    {
+        if (enableStorageCache_) {
+            std::fill(storageCacheUpToDate_[timeIdx].begin(),
+                      storageCacheUpToDate_[timeIdx].end(),
+                      /*value=*/0);
+        }
     }
 
     /*!
@@ -1840,6 +1883,7 @@ protected:
             const std::size_t numDof = asImp_().numGridDof();
             for (unsigned timeIdx = 0; timeIdx < historySize; ++timeIdx) {
                 storageCache_[timeIdx].resize(numDof);
+                storageCacheUpToDate_[timeIdx].resize(numDof, /*value=*/0);
             }
         }
 
@@ -1935,6 +1979,9 @@ protected:
     std::vector<bool> isLocalDof_;
 
     mutable std::array<GlobalEqVector, historySize> storageCache_;
+
+    // while these are logically bools, concurrent writes to vector<bool> are not thread safe.
+    mutable std::array<std::vector<unsigned char>, historySize> storageCacheUpToDate_;
 
     bool enableGridAdaptation_;
     bool enableIntensiveQuantityCache_;
--- a/opm/models/discretization/common/fvbasediscretization.hh
+++ b/opm/models/discretization/common/fvbasediscretization.hh
@@ -330,6 +330,7 @@ class FvBaseDiscretization
     enum {
         numEq = getPropValue<TypeTag, Properties::NumEq>(),
         historySize = getPropValue<TypeTag, Properties::TimeDiscHistorySize>(),
+        intensiveCacheSize = 1,
     };
 
     using IntensiveQuantitiesVector = std::vector<IntensiveQuantities,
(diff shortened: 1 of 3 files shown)


jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this opm-common=4655 please

jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

opm/models/discretization/common/fvbasediscretization.hh
```diff
@@ -412,19 +414,7 @@ public:
...
-                intensiveQuantityCacheUpToDate_[timeIdx].resize(numDof, /*value=*/false);
-            }
-
-            if (enableStorageCache_) {
-                storageCache_[timeIdx].resize(numDof);
-            }
-        }
-
-        resizeAndResetIntensiveQuantitiesCache_();
+        // Setting up the intensive quantities cache and storage cache is done in finishInit()
```
atgeirr, line 416: Seems to me it is done in applyInitialSolution(), not finishInit().

opm/models/discretization/common/fvbasediscretization.hh
```diff
@@ -658,6 +647,8 @@ public:
             return nullptr;
         }
 
+        assert(timeIdx < cachedIntensiveQuantityHistorySize_);
```
atgeirr: I think instead of assert(), you could rewrite the `if` below to
```
if (timeIdx >= cachedIntensiveQuantityHistorySize_) { ...
```
(Also rewrite the comment)

opm/models/discretization/common/fvbasediscretization.hh
```diff
@@ -686,8 +677,10 @@ public:
             return;
         }
 
+        assert(timeIdx < cachedIntensiveQuantityHistorySize_);
```
atgeirr: I suggest moving this condition into the above `if`, so we just return if we call this with too high timeIdx.

Similar on line 717, make it an early return instead.

opm/models/discretization/common/fvbasediscretization.hh
```diff
@@ -782,6 +783,7 @@ public:
      */
     void shiftIntensiveQuantityCache(unsigned numSlots = 1)
     {
+        // TODO: Consider removing this method as it i not needed when we only have one time step in the cache.
```
atgeirr: But in that case it does nothing? And it is still needed for the case when we have size 2?

opm/models/discretization/common/fvbasediscretization.hh
```diff
@@ -839,6 +841,8 @@ public:
     const EqVector& cachedStorage(unsigned globalIdx, unsigned timeIdx) const
     {
         assert(enableStorageCache_);
+        assert(timeIdx < historySize);
```
atgeirr: If I understand this correctly, historySize is still used but only for the storage cache history size? Should we then rename it accordingly?

opm/models/discretization/common/fvbasediscretization.hh
```diff
@@ -839,6 +841,8 @@ public:
...
+        assert(storageCacheUpToDate_[timeIdx][globalIdx] != 0);
```
atgeirr: This assert is in an inner loop, so it might be expensive (we still include assert() by default in release mode I think). However, we already have similar logic for cached intensive quantities, checking for validity for each globalIdx, returning a nullptr if not valid so it can't be that bad. However: Do we want to make the storage cache have the same logic? (I.e. allowing calling this always but possibly returning nullptr.)

opm/models/discretization/common/fvbaseelementcontext.hh
```diff
@@ -552,24 +552,33 @@ protected:
...
             }
             else {
-                updateSingleIntQuants_(dofSol, dofIdx, timeIdx);
-                model().updateCachedIntensiveQuantities(dofVars_[dofIdx].intensiveQuantities[timeIdx],
-                                                        globalIdx,
-                                                        timeIdx);
+                // If we have a cache for this time index, we can fetch the intensive quantities from the cache.
+                // We can then also fetch the thermodynamic hint from the cache.
+                dofVars_[dofIdx].thermodynamicHint[timeIdx] = model().thermodynamicHint(globalIdx, timeIdx);
+                const auto *cachedIntQuants = model().cachedIntensiveQuantities(globalIdx, timeIdx);
```
atgeirr: Nitpick: the `*` should be considered part of the type, so place it next to the auto:
 `const auto* cachedIntQuants = ...`

jakobtorben:
jenkins build this please

jakobtorben:
jenkins build this please

opm/models/discretization/common/fvbasediscretization.hh
jakobtorben: It is actually done in both places (it was already being done in finishInit() before this PR). I updated the comment to reflect this. I tried to remove if from finishInit() and only do it in applyInitialSolution() but then I got some test failures (perhaps from restart?). I didn't look too much into it as it is not that bad to set this up twice.

opm/models/discretization/common/fvbasediscretization.hh
jakobtorben: I was unsure if explicitly asserting this or just returning a nullptr was the best option here. When debugging, it was useful for it to not fail silently, but now that everything is working I am ok with returning a nulltpr instead which I now added.

opm/models/discretization/common/fvbasediscretization.hh
jakobtorben: I instead moved it to the if check above, to avoid getting a segmentation fault in this if checks' final clause. For this reason I also left the comment unchanged. However, I am not sure how valid the last check and comment is with the changes introduced in this PR.

opm/models/discretization/common/fvbasediscretization.hh
jakobtorben: Thanks for catching this. The comment was from an earlier iteration where I only considered the option to always have one intensive quantity cache. Now that we set it to 1 or 2, depending on the problem, this commented is outdated and was therefore deleted.

opm/models/discretization/common/fvbasediscretization.hh
jakobtorben: historySize is still used for the solution vectors and updating. So I think we can keep the current name.

opm/models/discretization/common/fvbaseelementcontext.hh
jakobtorben: done

opm/models/discretization/common/fvbasediscretization.hh
jakobtorben: I agree, we currently use two different methods for intensive quanitites (returning nullptr) and storage cache (assert), so I refactored storage cache to also return early or throw error if an invalid input was used.

atgeirr: Actually, they are different now as you made cachedStorage() throw if invalid, but now I think that is the best course of action. We should consider changing the IQ cache to behave in the same way, but not in this PR.

PR #6360, opened 2025-06-27
clean up memory and fix typo

multitalentloes:
Fix memory leak in autotuner

### DIFF
--- a/opm/simulators/linalg/gpuistl/detail/autotuner.hpp
+++ b/opm/simulators/linalg/gpuistl/detail/autotuner.hpp
@@ -44,7 +44,7 @@ tuneThreadBlockSize(func& f, std::string descriptionOfFunction)
 
     // TODO: figure out a more rigorous way of deciding how many runs will suffice?
     constexpr const int runs = 2;
-    GPUEvent events[runs + 1];
+    std::array<GPUEvent, runs+1> events;
 
     // Initialize helper variables
     float bestTime = std::numeric_limits<float>::max();


opm/simulators/linalg/gpuistl/detail/autotuner.hpp
```diff
@@ -82,6 +82,10 @@ tuneThreadBlockSize(func& f, std::string descriptionOfFunction)
         }
     }
 
+    for (int i = 0; i < runs + 1; ++i) {
+        OPM_GPU_SAFE_CALL(cudaEventDestroy(events[i]));
```
kjetilly: why not use the smart pointers for events?

multitalentloes: Yeah its more clean to just use the GPUEvents class, updated now.

opm/simulators/linalg/gpuistl/detail/autotuner.hpp
```diff
@@ -43,12 +44,7 @@ tuneThreadBlockSize(func& f, std::string descriptionOfFunction)
 
     // TODO: figure out a more rigorous way of deciding how many runs will suffice?
     constexpr const int runs = 2;
-    cudaEvent_t events[runs + 1];
-
-    // create the events
-    for (int i = 0; i < runs + 1; ++i) {
-        OPM_GPU_SAFE_CALL(cudaEventCreate(&events[i]));
-    }
+    GPUEvent events[runs + 1];
```
kjetilly: Should be an `std::array`. While it may not be paramount for the code as it stands, it will make our lifes easier in the future should we decide to extend the code (will have a `.size()`, we can add in bounds check, etc)

multitalentloes: fixed

PR #6366, opened 2025-07-04
Return geometry and their positions by value

SoilRos:
This is to start following the DUNE grid interface. Otherwise, these references will start to dangle.


blattms:
jenkins build this please

PR #6369, opened 2025-07-10
Remove the if that discoordinates values on different ranks

michal-toth:
When a well is distributed, we can run into a deadlock in _MultisegmentWellGeneric.cpp::scaleSegmentRatesWithWellRates_ when `segment_rates` differ between ranks and one rank picks an if-branch with a communication and the other without. The difference originates in the function _WellGroupHelpers.cpp:updateWellRatesFromGroupTargetScale_ that changes `surface_rates` only on the rank owning the well.

This commit makes all ranks change the `surface_rates`. It fixes a deadlock I encountered but needs further testing to see if it breaks something else. Alternative solution would be adding a communication copying the well from the owner to other ranks.

### DIFF
--- a/opm/simulators/wells/WellGroupHelpers.cpp
+++ b/opm/simulators/wells/WellGroupHelpers.cpp
@@ -522,11 +522,6 @@ updateWellRatesFromGroupTargetScale(const Scalar scale,
         if (!well_index.has_value())
             continue;
 
-        if (! wellState.wellIsOwned(well_index.value(), wellName) ) // Only sum once
-        {
-            continue;
-        }
-
         // scale rates
         auto& ws = wellState.well(well_index.value());
         if (isInjector) {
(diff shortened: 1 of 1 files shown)


michal-toth:
jenkins build this please

bska:
You might want to discuss this with the author of #2954 and, especially, commit fd4db9b933907a5814e38632ce63449dfe3865da.

blattms:
One could argue that this is a bugfix. But of course few people use distributed MS-wells, yet.

michal-toth:
> There is one more function in this file that could have the same bug: _updateGroupTargetReduction_. It is trickier though. It is also void but takes a vector by mutable reference. However, that vector is not used after this function is called (there is only one place in opm-simulators), so only changes to the `groupState` matter.

I removed the second early return for non-owner ranks because it looks to be a similar copy-paste mistake as the previous one. The function is more complicated, but here is the summary: it is _void_ and takes two mutable references (it is not _const_ but its class does not have any variables). There is no communication in the function or among the methods it calls.
One of the parameters taken by the reference is not use afterwards (the function is called only once in opm-simulators).
The second is `groupState` and it is the only thing affected by the commit. I believe that having all ranks perform the operations and have the same production- and injection reduction rates is better than leaving non-owner ranks with old values. However, I have not seen any differences in results yet after changing _updateGroupTargetReduction_.

PR #6370, opened 2025-07-10
Relax Upper Bound Check for Mobile Displacing Oil

bska:
This PR switches from a strict inequality for the upper bound on the mobile displacing oil saturation in the three-point scaling method to allowing equality as well.  Some recent experience demonstrates that equality is okay in this case.


tskille:
This has been tested on test models and full field models in Equinor. Results are excellent and I'm looking forward to see this go into master.

bska:
> This has been tested on test models and full field models in Equinor. Results are excellent and I'm looking forward to see this go into master.

Thank you for testing&ndash;that is greatly appreciated.  As the PR's approved and the build check is green, I'll merge into master.

PR #6371, opened 2025-07-14
Expand Support for Geomechanical Workflows With Fracturing

bska:
In particular,

* Create dedicated linear solvers, based on the Hypre package, for mechanical systems
* Add tools to amend grid views at runtime
* Assorted bug fixes


bska:
This work is superseded by #6593. Closing the PR as it is no longer needed.

PR #6372, opened 2025-07-14
cprw solver for distributed wells

michal-toth:
Intro:
The linear system uses Schur complement to remove DoFs of wells from the system. Nevertheless, cprw solver uses the information about the wells in the coarse correction. The pressure (sub-) system has includes wells. The correction to DoFs of wells is not used but having the entries in the matrix helps.

This commit has two parts:
-  The DoF of a distributed well will get the same ID on all ranks. Before this, each rank gave a unique ID to each DoF, so a distributed well was treated as a collection of smaller wells - each limited to its rank. As a result, contributions from different ranks get collected to the same row of the coarse system matrix. (Residual for the coarse well equation does not have to be collected, it is zero.)
- The calculation of the diagonal term averages the weights. We add the missing sum across ranks of the weights and of the number of perforations.

The changes heavily rely on the design of the `parallelWellInfo`, in particular the fact that on each rank this class holds the well-ID for each well and these IDs are the same. This is used to determine the IDs (we even save one communication). Moreover, the fixed ordering of wells prevents a deadlock. Each distributed well now needs to communicate, so progressing in a uniform pattern (from the smallest well-ID) is crucial.

### DIFF
--- a/opm/simulators/wells/StandardWellEquations.cpp
+++ b/opm/simulators/wells/StandardWellEquations.cpp
@@ -31,6 +31,7 @@
 #include <opm/simulators/linalg/istlsparsematrixadapter.hh>
 #include <opm/simulators/linalg/matrixblock.hh>
 #include <opm/simulators/linalg/SmallDenseMatrixUtils.hpp>
+#include <opm/simulators/wells/ParallelWellInfo.hpp>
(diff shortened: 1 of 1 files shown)


michal-toth:
jenkins build this please

PR #6373, opened 2025-07-16
VoigtArray and SymmTensor moved to opm-common

akva2:
Downstream of https://github.com/OPM/opm-common/pull/4661


bska (approved):
Very good.  I'll merge this into the master branch now

